PR #806, opened 2014-12-13
Move threat model doc to GitHub

runasand:
Move threat model doc to GitHub


garrettr:
This is a great, and very detailed, start. I think it is ready to merge as a WIP.

I think we should continue working on documenting and refining our threat model in the process of working on 0.4. Especially useful would be some more formal and quantitative approaches, for example using matrix approaches like DREAD and STRIDE. @dolanjs has also done a lot of excellent work on DFDs for SecureDrop, which will be merged into the repo once they are complete and we have figured out the best way to maintain them in a collaboration-friendly way.

For more ideas, see this [excellent OWASP guide](https://www.owasp.org/index.php/Application_Threat_Modeling) and Adam Shostack's [book](http://www.amazon.com/Threat-Modeling-Designing-Adam-Shostack/dp/1118809998/ref=sr_1_1?ie=UTF8&qid=1421817057&sr=8-1&keywords=threat+modeling).

PR #814, opened 2014-12-19
Grsecuirty - Ubuntu Kernel's .config Integration

eviljoel:
I'm not 100% on how secure this this, but using your instructions without this step leaves our SecureDrop installation unable to decrypt our root partition. Anyway, I feel these instructions should include some support for encrypted partitions. Curious what you think about this pull request.

I added the part about rebooting the server so that the "uname -r" would work correctly. Seems like a good idea anyway.

### DIFF
--- a/docs/ubuntu_grsec.md
+++ b/docs/ubuntu_grsec.md
@@ -35,7 +35,7 @@ changes are applied.
 sudo apt-get update
 sudo apt-get upgrade
 sudo apt-get dist-upgrade
-reboot
+sudo reboot
 ```
 
 ## Install dependencies on the offline server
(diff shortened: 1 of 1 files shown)


ageis:
I think we can close this issue as `ubuntu_grsec.md` doesn't exist anymore. It is generally a good idea to copy the OS's kernel config and use that as a basis before running `make menuconfig` and building the kernel w/ the grsec patch. We're doing that in https://github.com/conorsch/grsec/tree/ansible-compile-role, not sure whether we are doing that for the SecureDrop kernels since @garrettr builds it manually. However, it might be worthwhile to audit the diff in the CONFIG options compared to vanilla at some point, in case any of them affect system security in some way.

PR #817, opened 2014-12-21
Cleanup wordlist a little

ageis:


garrettr:
This is a good start (removes a lot of obscene and potentially offensive words).

I also think we should remove words that are unfamiliar or overly long. There are also some entries in the nouns list (`dictionaries/nouns.txt`) that are actually multiple words (e.g. "15 May Organization"), and these should be removed too.

PR #823, opened 2015-01-11
Ignore Whitespace in Auth Token

### LINKED ISSUE #820: Ignore spaces in two-factor code on document interface login
A recent update to Google Authenticator on iPhone adds a space between the first and last 3 digits of the 6-digit TOTP code, presumably for readability. If you enter the code with the space in the login form on the Document Interface, login will fail due to the presence of the space. In general, two-factor clients often add spaces to their one-time codes in order to improve readability. This could be a pitfall for users with such a two-factor client.

The easy fix here is to strip spaces before checking the code.

yasyf:
Removes any whitespace from the user's authentication token before attempting to verify it. Closes #820.

### DIFF
--- a/securedrop/db.py
+++ b/securedrop/db.py
@@ -198,7 +198,7 @@ class Journalist(Base):
         return hash
 
     def _format_token(self, token):
-        """Strips whitespace from authentication tokens, as many clients add these for readability"""
+        """Strips from authentication tokens the whitespace that many clients add for readability"""
         return ''.join(token.split())
 
     def set_password(self, password):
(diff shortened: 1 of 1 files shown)


garrettr:
Nice! Thanks for fixing that!

garrettr:
Two minor nits:

1. Consider moving `_format_token` to immediately precede `verify_token`. Since `verify_token` is the only place `_format_token` is used, this makes the code easier to read from top to bottom.
2. Another way to remove whitespace would be `token.replace(" ", "")`. I think this is a little clearer, although your method works too.

yasyf:
@garrettr Fixed! As for `2`, the reason I used `split` is that it handles all whitespace, not just spaces.

Issue #606, opened 2014-09-24
Improve SecureDrop backups


dolanjs:
We have two categories of DR backup:
One-time-backup: tor certs, app gpg keypair (pub and airgapped private), config.py, ssh 2fa secret values

Periodic-backups: app gpg keyring, db and /store directory

The initial install will make a backups of the One-time-backups at the end of the installation and encrypt them using the applications public key.

It would be nice if the document interface had a way to perform the periodic-backups. All of the files needed for periodic backups are owned by the application user so that shouldn't be an issue.

garrettr:
The one-time-backup use case discussed in the top comment has been implemented in #766 and #829. Implementing periodic backups is a bit harder, and is something we should consider for a future release.

flamsmark:
Storing something on the admin workstation is a form of backup, but it's not disaster recovery. A disaster recovery plan should allow for returning to regular-enough operations even if every piece of regular infrastructure is out of service. If the admin is in the same building as the SVS, app, and monitor servers when the fire starts, the backup created by this system will not be available for recovery.

eloquence:
Closing in favor of up-to-date issues tracking specific improvements, e.g., #5148.

PR #829, opened 2015-01-15
added a restore script

dolanjs:
The 0.3_restore.py will restore the decrypted zip file created by the 0.3_collect.py script.

This script is different than the 0.3_migrate.py which migrated the different db layout and dir structure for a 0.2.1 install to a 0.3 version

This scripts restores the:
 securedrop config.py
 /var/lib/securedrop/store
 /var/lib/securedrop/keys
 /var/www/securedrop/static/i/logo.png
 torrc
 source interface tor certs
 document interface tor certs

This script does not restore the ssh hidden service from the backup. Deleting the previous ssh ATHS and replacing it from with the one from the backup would break the ssh session that ansible was using to run the playbook.


dolanjs:
This should close issue #606 for now.

install_files/ansible-base/roles/backup/files/0.3_collect.py
```diff
@@ -56,7 +56,8 @@ def encrypt_zip_file(zf_fn):
     gpg = gnupg.GPG(binary='gpg2', homedir=config.GPG_KEY_DIR)
     e_fn = '{}.gpg'.format(zf_fn)
 
-    gpg.encrypt_file(zf_fn, config.JOURNALIST_KEY, always_trust='True', output=e_fn)
+    stream = open(zf_fn, "rb")
+    gpg.encrypt_file(stream, config.JOURNALIST_KEY, always_trust='True', output=e_fn)
```
garrettr: Why do you need to pass the stream to `encrypt_file`? It should open the file for you.

install_files/ansible-base/roles/backup/files/0.3_restore.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/python2.7
+"""
+
+This script and decrypted backup zip should be copied to the App server 
+and ran by the anisble plabook. When run (as root), it restores the 0.3
```
garrettr: Nit: s/ran/run

install_files/ansible-base/roles/backup/files/0.3_collect.py
dolanjs: I thought it was suppose to too but, when I didn't open it first and left it to encrypt_file to handle that would get this error trying to unzip the decrypted file:
``` 
me@me:~/git/securedrop$ gpg -o test.zip -d install_files/ansible-base/sd-backup-2014-12-17--03-03-17.zip.gpg 
gpg: encrypted with 4096-bit RSA key, ID A2201B2A, created 2013-10-12
      "SecureDrop Test/Development (DO NOT USE IN PRODUCTION)"
me@me:~/git/securedrop$ unzip test.zip 
Archive:  test.zip
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of test.zip or
        test.zip.zip, and cannot find test.zip.ZIP, period.
```

install_files/ansible-base/roles/backup/files/0.3_restore.py
```diff
@@ -0,0 +1,163 @@
...
+            extract_to_path(zf, zi, replace_prefix(zi.filename,
+                "var/lib/securedrop/keys", "/var/lib/securedrop/keys"))
+
+    subprocess.call(['chown', '-R', 'www-data:www-data', "/var/lib/securedrop"])
+
+
+def restore_database(zf):
+    print "* Migrating database..."
+
+    extract_to_path(zf, "var/lib/securedrop/db.sqlite", "db.old.sqlite")
```
garrettr: Does this extract_to_path call extract to the correct location? Shouldn't it extract to "/var/lib/securedrop/db.sqlite"?

garrettr:
Just a couple thoughts, you don't need to implement these:

1. The "collect into a zip file" idea ends up being pretty annoying, mostly because it doesn't preserve file metadata and we unfortunately still use file metadata in some places, hence the `getattr` hack in `extract_to_path`. We will remove all use of file metadata in the future (storing *all* needed information in the DB) so this will no longer be a problem then.
2. The general pattern of each restore function seems to be "extract one or more things from the zip file to a new location (which is often the same location), and chmod". Might be good to encapsulate that pattern in a function.

install_files/ansible-base/roles/backup/tasks/main.yml
```diff
@@ -30,3 +30,21 @@
   command: srm /tmp/{{ backup_filename.stdout }}.gpg
   async: 300
   poll: 10
+
+- name: if a restore file is declared copy migrate script to the app server tmp dir
+  copy: src=0.3_restore.py dest=/tmp/ owner=root mode=770
+  when: backup_zip is defined
+
+- name: if a resotre file is defined copy it to the app server tmp dir
```
garrettr: Nit: spelling

install_files/ansible-base/roles/backup/files/0.3_collect.py
garrettr: Investigated this a bit. Looks like the implementation of `encrypt_file` in python-gnupg in 1.2.5 does not do what is advertised on the tin (i.e. in the docstring). Don't want to take the time to explain everything here, but basically - this function is totally broken and does not work if a filename is passed as the first parameter. Using a stream (somehow) works.

garrettr:
:+1:

PR #830, opened 2015-01-16
Network firewall rules and docs improvements

garrettr:
Updates to the network firewall rules and documentation based on extensive testing today.


garrettr:
@dolanjs Please review!

dolanjs:
for the image `docs/images/firewall/ip_aliases.png` you specified 2 google dns server IPs. Need to verify that this wont cause any issue since I believe that we only use 1 IP address for the server's host iptables rules.

docs/network_firewall.md
```diff
@@ -16,72 +16,83 @@ Network Firewall Setup Guide
...
+
+On the "General Information" page, we recommend leaving your hostname as the default (pfSense). There is no relevant domain for SecureDrop, so we recommend setting this to "securedrop.local" or something similar. Use whatever DNS servers you wish. If you don't know what DNS servers to use, we recommend using Google's DNS servers: `8.8.8.8` and `8.8.4.4`. Click Next.
+
+Leave the defaults for "Time Server Information". Click Next.
+
+On "Configure WAN Interface", enter the appropriate configuration for your network. Consult your local sysadmin if you are unsure what to enter here. For many environments, the default of DHCP will work and the rest of the fields can be left blank. Click Next.
+
+For "Configure LAN Interface", set the IP address and subnet mask of the Application Subnet for the LAN interface. Click Next.
+
+Set a strong admin password. Click Next.
```
dolanjs: Would rec changing the note about the firewall admin password to ref using keepass on the admin's networked tails workstation to generate a random password and using that instead of user created one.

garrettr:
Re: 2 Google DNS servers. At the moment, we don't use the firewall to do DNS resolution. The servers do DNS resolution directly, per the instructions for manually configuring the network in `ubuntu_config.md`. The network firewall rules you mentioned are looser than the host firewall rules in that they allow an additional destination IP. I don't see how this would cause problems.

garrettr:
@dolanjs and I discussed this PR further this morning and decided that we should tighten the firewall rules a bit more, which also implies that we should set a static IP for the admin workstation. We are testing this latest configuration with an install now. Once we have worked out all the kinks, this PR will be updated with a slightly different final configuration.

Issue #824, opened 2015-01-13
Remove unfamiliar and overlong words from dictionaries


garrettr:
Follow-up for https://github.com/freedomofpress/securedrop/pull/817#issuecomment-69785979

In addition, it's important to consider what happens in the case of collisions. At the moment, they will simply be allowed. The `journalist_designation` field on `Source` does not have a uniqueness requirement, so if a journalist designation is randomly generated that is the same as one that already exists, we will simply add the new source with that designation and end up with two different sources with identical journalist designations.

This might be fine, since the probability of collision is probably low enough given the expected number of sources on a typical SecureDrop instance. Additionally, we might do something to prevent generating sources with identical journalist designations (similar to what we do to avoid having sources with matching codenames, see `generate_unique_codename` in `source.py`), but we would then have to handle the edge case where we cannot generate a unique designation, or where attempting to do so takes too long because there are so many collisions.

We do not currently do this for source codenames because there are so many possibilities (`math.log(6969**7, 2) = 89.36`), so we assume the probability of this happening is low enough to not be a concern. However, the number of possibilities is much lower for journalist designations:

```
➜  securedrop git:(develop) wc -l dictionaries/*
   14017 dictionaries/adjectives.txt
   67749 dictionaries/nouns.txt
➜  securedrop git:(develop) python -c "import math; print math.log(14017*67749, 2)"
29.8228020194
```

That's still a lot, but we should consider the edge case and at least make sure it is handled gracefully. Also not that this number will decrease (likely by a lot) as we remove words from the dictionaries.

garrettr:
Another approach that I would recommend would be to get a list of the most commonly used words in English (some examples would be here http://www.wordfrequency.info/free.asp) and cross-reference that with the existing nouns and adjectives lists that we already have.

Either way, it is important to manually examine the lists and make sure everything looks reasonable - but this step could making cutting them down much less laborious.

ageis:
Cross-referenced nouns.txt and adjectives.txt with two different word frequency lists. Cut down the size significantly, removing abnormal/unfamiliar words and leaving really simple ones.

Results here: https://github.com/ageis/securedrop/commit/8fe542a25f0a7fe0abc4eb1500862611abb0c9ea

PR #831, opened 2015-01-16
Remove less frequently used words

ageis:
Per #824, refined the wordlists based on cross-referencing with a list of high-frequency used words.


garrettr:
This looks like a massive improvement over the original wordlists! I still think we could trim them down further, but will probably have to do so manually (or go in the opposite direction, and build a list of amendable words from scratch).

PR #834, opened 2015-01-17
Rounded timestamps qf 822

### LINKED ISSUE #822: Rounded timestamps
Currently, the exact timestamps of submissions and replies are stored on the SecureDrop. We normalize submission timestamps (#301) for each source, so all submissions appear to have the same timestamp as the most recent submission (both in the database and on the filesystem). The ordering of the submissions is maintained naturally due to the auto-incrementing primary key in the database.

While this destroys the metadata of the prior history of submissions, it is precisely accurate for the latest submission and also for all undeleted replies (which could theoretically be used to infer some general timing information about prior submissions).

Another approach to consider would be to obfuscate _all_ timings by rounding. One specific adversary to consider is a passive network attacker near the source who monitors Tor flows (say their local network administrator or ISP). If they also had access to the document interface or database, then they would be able to precisely correlate the flows they observed with the timing of submissions.

We could mitigate this by rounding the times that we record to the nearest minute, hour, or day. The ordering must be preserved or the conversation will be impossible to follow, but there seems to be a reasonable tradeoff of the usability of following the conversation with mitigating potential metadata correlation attacks.

If we implemented this, we might be able to revert #301 (which is unintuitive). Or we could consider increasingly rounding the timestamps over time, so a recent submission may be rounded to an hour, an older submission may be rounded to the day, and an even older one may be rounded to the week or month. (This is just thinking out loud - I advocate implementing the simplest version of this first, and carefully considering the tradeoffs of implementing something like this due to the significantly increased complexity).

We would need to add UI information to the source and document interfaces to help users understand that the timestamps are rounded, and why.

Thanks to Roger Dingledine for suggesting this.

aliclark:
Fixes #822 by truncating source-related HH:MM times to 00:00, leaving only the date information according to the server's local time.

This would normally provide 1 day of granularity except:
- When "a day ago" is visible, the granularity is at 0.5 days
- If the journalist sees a new item on the page very early in the day, at say 1AM, then it must have been uploaded in the past hour.

The former issue could be solved by making 2 days mean 2.0-3.0 days, instead of the current 1.5-2.5 days, however half a day of granularity is about as good as could be achieved at the moment anyway.

The latter issue could be improved by replacing all date information with a "new" flag or solved by batching uploads, but subject to it being acceptable for users.

### DIFF
--- a/securedrop/source.py
+++ b/securedrop/source.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import os
 from datetime import datetime
+import time
 import uuid
 from functools import wraps
 import zipfile
@@ -71,6 +72,15 @@ def ignore_static(f):
             return  # don't execute the decorated function
         return f(*args, **kwargs)
     return decorated_function
+
+
+def datetime_truncated_to_local_midnight():
+    """
+    Takes only the .date() part of of *local* time (in the timezone of
+    the server), which truncates to the nearest midnight. Then uses
+    mktime to convert this to UTC and on to a naive UTC datetime
+    """
+    return datetime.utcfromtimestamp(time.mktime(datetime.now().date().timetuple()))
 
 
 @app.before_request
@@ -175,7 +185,7 @@ def async_genkey(sid, codename):
     # flagged source logs in and has a key generated for them. #789
     try:
         source = Source.query.filter(Source.filesystem_id == sid).one()
-        source.last_updated = datetime.now.date()
+        source.last_updated = datetime_truncated_to_local_midnight()
         db_session.commit()
     except Exception as e:
         app.logger.error("async_genkey for source (sid={}): {}".format(sid, e))
@@ -265,7 +275,7 @@ def submit():
         if entropy_avail >= 2400:
             async_genkey(g.sid, g.codename)
 
-    g.source.last_updated = datetime.now().date()
+    g.source.last_updated = datetime_truncated_to_local_midnight()
     db_session.commit()
     normalize_timestamps(g.sid)
 
--- a/securedrop/source.py
+++ b/securedrop/source.py
@@ -217,12 +217,11 @@ def normalize_timestamps(sid):
     """
     sub_paths = [ store.path(sid, submission.filename)
                   for submission in g.source.submissions ]
-    if len(sub_paths) > 1:
-        args = ["touch"]
-        args.extend(sub_paths[:-1])
-        rc = subprocess.call(args)
-        if rc != 0:
-            app.logger.warning("Couldn't normalize submission timestamps (touch exited with %d)" % rc)
+    args = ["touch", "-t", datetime.now().strftime("%Y%m%d") + "0000"]
+    args.extend(sub_paths)
+    rc = subprocess.call(args)
+    if rc != 0:
+        app.logger.warning("Couldn't normalize submission timestamps (touch exited with %d)" % rc)
 
 
 @app.route('/submit', methods=('POST',))
--- a/securedrop/source.py
+++ b/securedrop/source.py
@@ -175,7 +175,7 @@ def async_genkey(sid, codename):
     # flagged source logs in and has a key generated for them. #789
     try:
         source = Source.query.filter(Source.filesystem_id == sid).one()
-        source.last_updated = datetime.utcnow()
+        source.last_updated = datetime.now.date()
         db_session.commit()
     except Exception as e:
         app.logger.error("async_genkey for source (sid={}): {}".format(sid, e))
@@ -262,7 +262,7 @@ def submit():
         if entropy_avail >= 2400:
             async_genkey(g.sid, g.codename)
 
-    g.source.last_updated = datetime.utcnow()
+    g.source.last_updated = datetime.now().date()
     db_session.commit()
     normalize_timestamps(g.sid)
 
--- a/securedrop/source.py
+++ b/securedrop/source.py
@@ -217,6 +217,9 @@ def normalize_timestamps(sid):
     """
     sub_paths = [ store.path(sid, submission.filename)
                   for submission in g.source.submissions ]
+    # touch accepts date in local time (server's timezone), which
+    # causes truncation to nearest midnight local time. Python's now()
+    # is assumed to use the same local time. See #822.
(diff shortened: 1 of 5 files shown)


aliclark:
8dc6a91385a20265b8601a12976c22397a2988c3 may not be a good one actually - it changes utcnow() -> now(), but _relative_timestamp is still using utcnow().

However just using UTC with date comparisons could give results that seem odd when viewed in the server's local time, eg. a source update that happened an hour ago might have happened yesterday in UTC terms, so would show "1 day ago" to the journalist..

The submissions should be correct since os_stat.st_mtime is returned in UTC, however thinking about it the local time truncation is probably not ideal.

This could be solved by adding a "date-was-truncated-in-this-timezone" column, but then it's starting to get complex :/

Thoughts? Is this the right approach generally? Rounding to 1 hour could avoid some of these problems, but probably doesn't help the original problem much

garrettr:
@aliclark Thanks for this PR! Will take a closer look later this week, we're in a crunch getting the 0.3 release ready to go out and this is targeted for 0.4. Will update the branches soon so we can merge this in while also preparing for 0.3, and will have some comments on your approach.

securedrop/source.py
```diff
@@ -217,12 +217,11 @@ def normalize_timestamps(sid):
     """
     sub_paths = [ store.path(sid, submission.filename)
                   for submission in g.source.submissions ]
-    if len(sub_paths) > 1:
-        args = ["touch"]
-        args.extend(sub_paths[:-1])
-        rc = subprocess.call(args)
-        if rc != 0:
-            app.logger.warning("Couldn't normalize submission timestamps (touch exited with %d)" % rc)
+    args = ["touch", "-t", datetime.now().strftime("%Y%m%d") + "0000"]
```
garrettr: Sweet, it's awesome that you can arbitrarily modify the last modified time like that. It might be worth filing a follow-up to make sure the actual creation time isn't being stored [somewhere sneaky](http://unix.stackexchange.com/questions/50177/birth-is-empty-on-ext4/50184#50184).

securedrop/source.py
```diff
@@ -263,7 +262,7 @@ def submit():
         if entropy_avail >= 2400:
             async_genkey(g.sid, g.codename)
 
-    g.source.last_updated = datetime.utcnow()
+    g.source.last_updated = datetime.now().date()
```
garrettr: All of the dates that get stored should be normalized to UTC. Any reason you can't use `utcnow().date()`?

garrettr:
> Thoughts? Is this the right approach generally?

I'm not sure. This rounds the times that are *displayed* in the Document Interface, but does not change the times that are stored on the server. So an attacker who can access the Document Interface cannot learn potentially useful precise timing information, but an attacker who can access or seize the server can. In the spirit of "normalize timestamps", I think we should completely get rid of this information so even a server seizure provides only rounded timestamps.

This should be possible by changing `last_updated` and similar fields to store a rounded timestamp, and using `time -t` to round the mtimes of submissions stored in the filesystem.

garrettr:
You make a good point about time zones not being handled well right now. There are three considerations that I think are important here:

1. We should definitely always store datetimes as UTC in the database. This prevents a slow descent into madness.
2. If we are displaying time information in either the Source or Document Interface, we should adjust the time displayed for the local time zone of the user so it actually makes sense. This should handled in the view layer. We do not do this now (this might merit a separate issue itself).
3. If we are rounding timestamps, we should consider the overall interface. If we are messing with the timestamps, is displaying them to the user (as we do now) helpful at all, or just confusing? Would it be better to just show the order of submissions, and a "new" flag, maybe with a rounded "last updated" time for the source?

aliclark:
> does not change the times that are stored on the server

Yeah I haven't investigated that much, so I'm not sure how difficult a defense against local forensics would be. /var/lib/securedrop/store/UUID still has the submission time stat for one.. Please could we consider that in an additional issue/PR? I think this adds some initial protection and protecting against local forensics might turn out to be difficult.

securedrop/source.py
aliclark: Yes I think this was incorrect,  c1587b3a99ba5b9ec7382aa25450b25950635c91 adds conversion to utc.

utcnow().date() is possible, it just changes *which* midnight is used. "touch -t "${date}0000" truncates to midnight from the server's local timezone, and datetime.now().date() is a roughly equivalent truncation in python.

If the journalist logs on at 9am and is in the same timezone as the server, then a submission with "less than a day ago" will have been sumitted on that same date, whereas "1 day ago" would have been submitted yesterday in their timezone.

This is a bit of a kludgy guess because of course the organisation could set the server timezone to something other than the journalist's time, but UTC is even more likely to be different.

At the end of the day it's hard to truncate the time in a delicate way, because useful display information is being discarded. A slightly more engineered approach would be to store the timezone ID that these truncations are occuring in *somewhere*, in case that becomes useful information later, but it wouldn't allow a more accurate display. Another option would be to allow the truncation timezone to be configured as something other than the server's local timezone in the UI.

aliclark:
re. https://github.com/freedomofpress/securedrop/pull/834#issuecomment-71298043 all good points. Some of what I wrote in https://github.com/freedomofpress/securedrop/pull/834#discussion_r23497797 relates to it.

On 3 I'd be happy to code a PR for that if it's considered an acceptable UI change for the users.

fowlslegs:
Adding this PR to 0.4 milestone just because the corresponding issue was already there. We can remove both if we want to push this back. I think this is one we want to discuss in-person as there are definitely some good issues brought up in the comment thread here.

fowlslegs:
I think we should either close this PR, or commit to reworking it for an upcoming release such as `0.4.1`.

redshiftzero:
Thanks for this contribution @aliclark! I apologize for the long delay here, obviously significant changes have occurred in the underlying codebase in the 2.5+ years since this was submitted. Given this and the considerations described in [this comment](https://github.com/freedomofpress/securedrop/issues/822#issuecomment-333052295), we're going to close your PR.

That said, if you're interested in taking a peek at SecureDrop development in the future, please feel free to hop in our [Gitter chat](gitter.im/freedomofpress/securedrop) and we can point you to an easier issue to tackle.

PR #837, opened 2015-01-21
drop invalid state packets

### LINKED ISSUE #836: iptables invalid state packets (OSSEC spam)
Not sure exactly why yet (see [here](http://serverfault.com/questions/309691/why-is-our-firewall-ubuntu-8-04-rejecting-the-final-packet-fin-ack-psh-wit) and [here](https://lists.torproject.org/pipermail/tor-talk/2014-March/032503.html)) but the server is removing the connection from the state table before the server sends the last of the traffic. This last packet then does not match the state table or uid for the allowed rules and is logged and dropped generating spammy OSSEC alerts. The email alerts are still sent and delivered.

Should add outbound and inbound rule to explicitly drop (but not log) invalid packets.

```
Jan 21 02:29:19 mon-staging kernel: [ 2116.816960] Denied_INVALID IN=eth0 OUT= MAC=08:00:27:a9:c5:8c:52:54:00:12:35:02:08:00 SRC=10.0.2.2 DST=10.0.2.15 LEN=40 TOS=0x00 PREC=0x00 TTL=64 ID=57698 PROTO=TCP SPT=587 DPT=35152 WINDOW=65535 RES=0x00 ACK RST URGP=0 
```

dolanjs:
Drop but do not log packets with invalid state. This situation was occuring around sending smtp alerts to google smtp servers. Where the connection would be removed from the iptables state table, but the mon server would then still try to send a FIN, RST packet but since it was no longer in the state table it lost the associated PID and UID. The distant end already closed the connection so this does not prevent the alerts from being sent but does cause an OSSEC spam loop. An OSSEC alert is sent causing a fw drop, that then generates another OSSEC alert.

These packets are dropped but not logged by jumping (-j DROP) directly to DROP chain instead of the LOGNDROP chain

This should close #836

### DIFF
--- a/install_files/ansible-base/roles/restrict_direct_access_app/templates/app_rules_v4
+++ b/install_files/ansible-base/roles/restrict_direct_access_app/templates/app_rules_v4
@@ -4,10 +4,9 @@
 :OUTPUT ACCEPT [0:0]
 :LOGNDROP - [0:0]
 
-# Drop but do not log all packets with invalid state.
-# This was added due to issue #836
--A OUTPUT -m state --state INVALID -j DROP
--A INPUT -m state --state INVALID -j DROP
+# Drop and log all packets with invalid state.
+-A OUTPUT -m state --state INVALID -j LOGNDROP
+-A INPUT -m state --state INVALID -j LOGNDROP
 
 # Allow the tor instances used for ssh, source and document interface outbound access to default tor destination ports.
 # Load before tor user drop rules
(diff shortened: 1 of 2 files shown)


garrettr:
Hate to be a drag, but would it be better to log these packets (they do seem relevant, their cause remains unclear, and we should debug them at some point) and just not send the OSSEC email alerts?

dolanjs:
I can do that instead... it will require modifying the iptables log prefixes and `/var/ossec/rules/local_rules.xml` to match the invalid packet drops and not alert on them. That file will need to be packaged with the ossec deb packages for deployment.

dolanjs:
Instead of logging all invalid packets and just not alerting on them, I think it would be better to restrict the INVALID state fw rule included in this PR by the distant ends port (smtp_relay_port) and not log those but log and alert on other invalid state packets.

We can reliably reproduce the situation for the invalid state packets for smtp connection using google smtp servers. We also have a packet capture of it so don't see the need to continually log these events.

garrettr:
Agreed, that is a good compromise.

dolanjs:
modified the invalid state iptables rule to only not log the smtp traffic with an invalid state but still log all the other invalid state packets per previous discussion.

PR #839, opened 2015-01-21
update iptable version of network fw rules

dolanjs:
updated the iptable version of the network pfsense fw rules to match the more restrictive pfsense rules for the .3 version


garrettr:
:+1:

PR #842, opened 2015-01-22
Package build automation

garrettr:
Some fixes for bugs in the package building Ansible roles, adds an Ansible role to automatically build the grsec metapackage as part of app-build.


garrettr:
@dolanjs Please review!

install_files/ansible-base/build-deb-pkgs.yml
```diff
@@ -21,5 +21,6 @@
   roles:
     - build-securedrop-app-code-deb-pkg
     - { role: build-ossec-deb-pkg, tags: [ "ossec" ] }
+    - build-grsec-metapackage-deb-pkg
```
dolanjs: Even though the grsec meta package doesn't take long to build we could add a role for it so it could be skipped. Thoughts?

garrettr: By role I assume you mean tag, and yes I think that is a good idea :)

install_files/ansible-base/roles/build-grsec-metapackage-deb-pkg/tasks/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+- include: build_grsec_metapackage_deb_pkg.yml
```
dolanjs: Since this role `build-grsec-metapackage-deb-pkg` only builds the grsec-metapackage, should move the contents of ` install_files/ansible-base/roles/build-grsec-metapackage-deb-pkg/tasks/build_grsec_metapackage_deb_pkg.yml` to ` install_files/ansible-base/roles/build-grsec-metapackage-deb-pkg/tasks/main.yml` instead of including the separate task file.

dolanjs:
commented two nits otherwise lgtm

install_files/ansible-base/build-deb-pkgs.yml
dolanjs: thanks that is what I meant... add a tag for that role.... :)

Issue #828, opened 2015-01-14
Remove dbus firewall rule


garrettr:
In editing the local iptables rule, I came across this rule (in `install_files/ansible-base/roles/restrict_direct_access_app/templates/rules_v4`):

```
-A OUTPUT -p tcp -m owner --uid-owner messagebus -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT -m comment --comment "tor instance that provides ssh access"
```

I cannot imagine why the messagebus user (which runs dbus) would need to make outgoing TCP connections. @dolanjs says this rule was required at one point in order for Tor to work, but also says that may have been back when we used chroot jails for isolation, and that it may no longer be needed now.

To make sure this works, we should test it in a production installation before removing the rule.

Issue #845, opened 2015-01-23
Invalid state and untracked packet iptables issues.


dolanjs:
For the Monitor servers smtp traffic and (app and monitor servers) tor traffic keep experiencing issues with iptables state table. Even thought there are fw rules to allow outbound `NEW,ESTABLISHED,RELATED` and the corresponding inbound `ESTABLISHED,RELATED` for SMTP and Tor traffic, various `INVALID` and `UNTRACKED` events are being dropped. The sequence of events dropped packets:

```
Jan 23 16:01:51 mon-staging kernel: [39702.413350] Denied_INVALID IN= OUT=eth0 SRC=10.0.2.15 DST=74.125.20.108 LEN=40 TOS=0x00 PREC=0x00 TTL=64 ID=55750 DF PROTO=TCP SPT=49706 DPT=587 WINDOW=32160 RES=0x00 ACK URGP=0 
Jan 23 16:01:51 mon-staging kernel: [39702.413475] Denied_ACK IN= OUT=eth0 SRC=10.0.2.15 DST=74.125.20.108 LEN=40 TOS=0x00 PREC=0x00 TTL=64 ID=55750 DF PROTO=TCP SPT=49706 DPT=587 WINDOW=32160 RES=0x00 ACK URGP=0 
Jan 23 16:01:51 mon-staging kernel: [39702.413350] Denied_TCP IN= OUT=eth0 SRC=10.0.2.15 DST=74.125.20.108 LEN=40 TOS=0x00 PREC=0x00 TTL=64 ID=55750 DF PROTO=TCP SPT=49706 DPT=587 WINDOW=32160 RES=0x00 ACK URGP=0 
```

This is causing a lot of false positives. Until the issue can be fully diagnosed we should not log OUTBOUND host based FW drops. Even though this could mean that an IOC is not logged compared to the amount of false positives logging them currently generates I think it is worth the trade off. 

Related to #836

garrettr:
Fyi: IOC = incident of compromise

garrettr:
@dolanjs Can you provided a detailed STR for future investigators?

eloquence:
Closing as stale; since the OSSEC spam issue is resolved, we'd need clear STR for any remaining issues with log spam.

PR #846, opened 2015-01-23
Iptables invalid and acks

dolanjs:
Modified the iptable rules to not log outbound host fw drops to reduce the OSSEC spam

Also not logging inbound invalid state packets since this was generating excessive OSSEC alerts

All other inbound fw drops are logged.

Related issue #845

### DIFF
--- a/install_files/ansible-base/roles/restrict_direct_access_app/templates/app_rules_v4
+++ b/install_files/ansible-base/roles/restrict_direct_access_app/templates/app_rules_v4
@@ -4,13 +4,8 @@
 :OUTPUT ACCEPT [0:0]
 :LOGNDROP - [0:0]
 
-# Drop and log all packets with invalid state.
--A OUTPUT -m state --state INVALID -j LOGNDROP
--A INPUT -m state --state INVALID -j LOGNDROP
-
 # Allow the tor instances used for ssh, source and document interface outbound access
 # Load before tor user drop rules
--A OUTPUT -p tcp -m owner --uid-owner messagebus -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT -m comment --comment "tor instance that provides ssh access"
 -A OUTPUT -p tcp -m owner --uid-owner debian-tor -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT -m comment --comment "tor instance that provides ssh access"
 -A INPUT -p tcp -m state --state ESTABLISHED,RELATED -j ACCEPT -m comment --comment "Allow traffic back for tor"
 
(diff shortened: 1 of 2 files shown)


garrettr:
Note that #845 is a follow up to further investigate the cause of the dropped packets and is not resolved by this PR.

install_files/ansible-base/roles/restrict_direct_access_app/templates/app_rules_v4
```diff
@@ -4,13 +4,8 @@
 :OUTPUT ACCEPT [0:0]
 :LOGNDROP - [0:0]
 
-# Drop and log all packets with invalid state.
--A OUTPUT -m state --state INVALID -j LOGNDROP
--A INPUT -m state --state INVALID -j LOGNDROP
-
 # Allow the tor instances used for ssh, source and document interface outbound access
 # Load before tor user drop rules
--A OUTPUT -p tcp -m owner --uid-owner messagebus -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT -m comment --comment "tor instance that provides ssh access"
```
garrettr: This resolves #828

install_files/ansible-base/roles/restrict_direct_access_app/templates/app_rules_v4
```diff
@@ -67,15 +62,16 @@
 -A INPUT -i lo -p all -j ACCEPT -m comment --comment "Allow lo to lo traffic all protocols"
 -A OUTPUT -o lo -p all -j ACCEPT -m comment --comment "Allow lo to lo traffic all protocols"
 
-# Drop and log all other traffic
--A INPUT -j LOGNDROP -m comment --comment "Drop all other incomming traffic"
--A OUTPUT -j LOGNDROP -m comment --comment "Drop all other outgoing traffic"
+# Don't log inbound invalid state packets related to issue #830
```
garrettr: Are you sure #830 the right issue to reference here?

install_files/ansible-base/roles/restrict_direct_access_app/templates/app_rules_v4
```diff
@@ -67,15 +62,16 @@
...
+-A OUTPUT -j DROP -m comment --comment "Drop all other outgoing traffic"
 
 # LOGNDROP everything else
--A LOGNDROP -p tcp -m limit --limit 5/min -j LOG --log-prefix "Denied_INVALID " --log-ip-options --log-tcp-options --log-uid --log-level 4
--A LOGNDROP -p tcp -m limit --limit 5/min -j LOG --log-prefix "Denied_TCP " --log-ip-options --log-tcp-options --log-uid --log-level 4
--A LOGNDROP -p udp -m limit --limit 5/min -j LOG --log-prefix "Denied_UDP " --log-ip-options --log-uid --log-level 4
--A LOGNDROP -p icmp -m limit --limit 5/min -j LOG --log-prefix "Denied_ICMP " --log-ip-options --log-uid --log-level 4
+-A LOGNDROP -p tcp -m limit --limit 5/min -j LOG --log-ip-options --log-tcp-options --log-uid --log-level 4
+-A LOGNDROP -p udp -m limit --limit 5/min -j LOG --log-ip-options --log-uid --log-level 4
+-A LOGNDROP -p icmp -m limit --limit 5/min -j LOG --log-ip-options --log-uid --log-level 4
```
garrettr: Why did you get rid of `--log-prefix`?

install_files/ansible-base/roles/restrict_direct_access_mon/templates/mon_rules_v4
```diff
@@ -65,14 +56,16 @@
 -A INPUT -i lo -p all -j ACCEPT -m comment --comment "Allow lo to lo traffic all protocols"
 -A OUTPUT -o lo -p all -j ACCEPT -m comment --comment "Allow lo to lo traffic all protocols"
 
-# Drop and log all other traffic
--A INPUT -j LOGNDROP -m comment --comment "Drop all other incomming traffic"
--A OUTPUT -j LOGNDROP -m comment --comment "Drop all other outgoing traffic"
+# Don't log inbound invalid state packets related to issue #830
```
garrettr: Again, don't think #830 is the right issue

install_files/ansible-base/roles/restrict_direct_access_app/templates/app_rules_v4
dolanjs: This was left over from when the the source and document interface were in separate chroot jails. The user that the host system would see outside of the chroot jails was the messagebus user. This rule is not needed anymore with current non-chroot architecture

install_files/ansible-base/roles/restrict_direct_access_app/templates/app_rules_v4
dolanjs: OSSEC has a few decoders to use to parse host fw rules. When we were adding the log prefix it would group possible denies by protocol and apply the correlation rules on those by each group. Since disabled logging some of the drops the OSSEC correlation rules with the log prefix would never be met since some of the required info is not being logged.

PR #850, opened 2015-01-27
Use a variable for the FPF package repo URL

garrettr:
This makes it easier to change the URL for testing purposes.

It is also DRY.


dolanjs:
lgtm

PR #854, opened 2015-01-28
Avoid cron-apt warning about multiple source lists

garrettr:
Make sure the *only* list that is being checked for updates is
`security.list` by explictily telling cron-apt to ignore source lists in
`/etc/apt/sources.list.d`, which is defined by the configuration
variable `Dir::Etc::SourceParts`. See the man page for details.


dolanjs:
lgtm

PR #855, opened 2015-01-28
Test securedrop app code upgrades

garrettr:
Fixes and improvements that were created while testing the upgrade process for the securedrop-app-code Debian package. Key addition is the new `./upgrade_version.sh` script, which automates the minutia required for updating the package.


garrettr:
Also includes a fix for the grsecurity kernel build directions, wherein `CONCURRENCY_LEVEL` was being set but not *used* due to a conflict with `sudo`, resulting in kernel builds only using 1 core and being much slower as a result. I am testing exactly how much slower that makes the builds right now (whilst also building the newest 3.14.30 grsecurity kernel).

garrettr:
Finally, fixes the Python requirements auto-update script, and updates the Python requirements with it!

dolanjs:
reviewed and it lgtm, but haven't tested myself.

PR #860, opened 2015-02-04
Limit key exchange algos

ageis:


ageis:
I tested this setting with Tails 1.2.3 and the staging VMs. Of course I had to make a new user, copy my SSH key to authorized_keys, and add that user to the ssh group. I also disabled pam_google_authenticator.so for ease, but at the end of the day, the SSH client was successfully able to connect and negotiate with the daemon.

There was a message saying "Agent admitted failure to sign using the key" (not sure as yet what that means) but that was present without the KexAlgorithms setting too.

PR #863, opened 2015-02-05
apparmor updates

dolanjs:
Changed default mode for apparmor tor and apache2 profiles to enforce mode. The ansible staging playbook app-test role will still put the profiles in complain mode unless the skip tag `aa-complain` is added to the Vagrantfile.

Added the needed entries for the current develop version to work. Note the `/tmp` and `/var/tmp` and `/var/www/` entries. Each are being required at different parts of the submission process. 

Other capabilities were also added but more testing would be required to ensure that they weren’t caused by the app-test dependencies.

added ossec decoders and rules to reduce alerts being sent for false positives.

To test, in the `Vagrantfile` set the `staging.vm.provision` skip tags config line to look like the following: `ansible.skip_tags = "aa-complain"`

Then create the new local deb packages using the build vm
`vagrant provision build`

Then create a new staging environment:
`vagrant destroy /staging$/ -f && vagrant up /staging$/`

This should leave you with a staging environment with the tor and apparmor profiles set to enforce mode. 

This will need to be retested with PR #862


install_files/ansible-base/usr.sbin.apache2
```diff
@@ -188,6 +196,8 @@
   /var/www/securedrop/worker.py r,
   /var/www/securedrop/worker.pyc rw,
   /var/www/source.wsgi r,
+  /var/www/.gnupg/ rw,
```
garrettr: Where does this come from? SecureDrop's key directory is `/var/lib/securedrop/keys`.

install_files/ossec-server/var/ossec/etc/local_decoder.xml
```diff
@@ -1,3 +1,7 @@
+<!--
+  Cron-apt logs to syslog and error and warnings events are distinguished by
+  W: or E: at the beginning of the event.
+-->
```
garrettr: Thanks for the comment :+1:

install_files/securedrop-app-code/DEBIAN/postinst
```diff
@@ -24,8 +24,17 @@ case "$1" in
     chown www-data:www-data /var/www/document.wsgi
     chown www-data:www-data /var/www/source.wsgi
 
-    aa-complain /etc/apparmor.d/usr.sbin.tor
-    aa-complain /etc/apparmor.d/usr.sbin.apache2
+    a2dissite 000-default
```
garrettr: Mind adding a comment here? Also, can't remember - *why* did we agree it was better to do this in the postinst, instead of in the Ansible playbook?

dolanjs: will add the comment. 

We also check for it in the ansible playbook but the issue was putting the securedrop apache apparmor profile in enforce mode as part of the securedrop-app-code deb package postinst. It would generate an apparmor error and break the install if the two default sites weren't disabled first since they aren't defined in the apparmor profile.

dolanjs: we want the apparmor profile to be part of the app code deb package so we can update it when the application code changes without requiring the playbook to be ran manually by the admin.

install_files/ossec-server/var/ossec/rules/local_rules.xml
```diff
@@ -16,3 +19,88 @@
...
+<!--
+  The python gnupg library that securedrop uses includes an obsolete option
+  WARNING:gnupg no-use-agent is an obsolete option - it has no effect
+  Do not send an alert for this event.
+-->
+<group name="do not alert">
+  <rule id="199997" level="0">
+    <decoded_as>gpg warning</decoded_as>
+    <description>due to upstream lib</description>
+    <options>no_email_alert</options>
```
dolanjs: Should add a <match> for the specific `no-use-agent is an obsolete option` event so that other possible gpg warning events don't get suppressed.

garrettr: Yes, definitely.

install_files/ansible-base/usr.sbin.apache2
dolanjs: I can reproduce reliably on submission to the source interface. I don't know why it generated since we use a non-default homedir but every time it is generated

```
Feb  7 00:37:04 app-staging kernel: [13060.528666] audit: type=1400 audit(1423269424.913:22): apparmor="DENIED" operation="mkdir" profile="/usr/sbin/apache2" name="/var/www/.gnupg/" pid=4286 comm="gpg-agent" requested_mask="c" denied_mask="c" fsuid=33 ouid=33
Feb  7 00:37:04 app-staging kernel: [13060.552536] audit: type=1400 audit(1423269424.937:23): apparmor="DENIED" operation="mkdir" profile="/usr/sbin/apache2" name="/var/www/.gnupg/" pid=4293 comm="gpg-agent" requested_mask="c" denied_mask="c" fsuid=33 ouid=33
```

garrettr: Looking into this now. My guess is some tomfoolery due to python-gnupg.

garrettr: Fixed this in #868

PR #864, opened 2015-02-06
Remove RC4 from deployment best practices

ageis:
Although RC4 is still safe in some combinations, the attacks are steadily becoming more realistic. Even though CloudFlare is still using it in their [sslconfig](https://github.com/cloudflare/sslconfig/blob/master/conf) I believe it's time to just drop it. Use of this config will result in a Qualys SSL Labs grade of A+.

### DIFF
--- a/docs/deployment_practices.md
+++ b/docs/deployment_practices.md
@@ -74,7 +74,7 @@ In your SSL (port 443) virtual host, set up HSTS and use these settings to give 
     SSLProtocol all -SSLv2 -SSLv3
     SSLHonorCipherOrder on
     SSLCompression off
-    SSLCipherSuite EECDH+AES128:RSA+AES128:EECDH+AES256:RSA+AES256:EECDH+3DES:RSA+3DES:EECDH+RC4:RSA+RC4:!MD5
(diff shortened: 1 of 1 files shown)


docs/deployment_practices.md
```diff
@@ -74,7 +74,7 @@ In your SSL (port 443) virtual host, set up HSTS and use these settings to give
     SSLProtocol all -SSLv2 -SSLv3
     SSLHonorCipherOrder on
     SSLCompression off
-    SSLCipherSuite EECDH+AES128:RSA+AES128:EECDH+AES256:RSA+AES256:EECDH+3DES:RSA+3DES:EECDH+RC4:RSA+RC4:!MD5
```
garrettr: I'm fine with this change as long as it doesn't make landing pages inaccessible for a substantial subset of users (due to browser compatibility issues).

garrettr: Asking around on [Twitter](https://twitter.com/garrettr_/status/563792813393842178) confirmed that discontinuing RC4 would prevent some people with very old clients from accessing the page, but that it is a fairly minor concern.

PR #867, opened 2015-02-08
separate SD specific config from upstream OSSEC deb

dolanjs:
Separate the SecureDrop specific configs files from the upstream OSSEC deb packages.

There are 3 config files specific for SecureDrop instances:

```
/var/ossec/etc/ossec.conf
/var/ossec/rules/local_rules.xml
/var/ossec/etc/local_decoders.xml
 ```

The `securedrop-ossec-*` packages contain just the specific config files for that server and list the ossec-agent or ossec-server upstream package as a dependency. In the respective DEBIAN/control file it has the respective  `Replaces: ossec-server` so that it can replace a file placed by another package.

These 3 files can be separated from the upstream OSSEC packages to make versioning and tracking easier. This changes the OSSEC packages to the following:

`securedrop-ossec-server version`: upstream+SD (2.8.1+0.3)
`securedrop-ossec-agent version`: upstream+SD (2.8.1+0.3)

`ossec-agent version`: upstream only (2.8.1)
`ossec-server version`: upstream only (2.8.1)

The `ossec-server` and `ossec-agent` package build processes were moved [ here ](https://github.com/freedomofpress/ossec) since they do not contain any securedrop specific configs. 

[1] ossec.conf, local_decoders.xml and local_rules.xml

### CODE REVIEW

File: install_files/ansible-base/group_vars/securedrop.yml [hunk 1]
```diff
@@ -58,19 +58,4 @@ grub_pax:
...
 grsec_package: securedrop-grsec
 
-### Used by the build-ossec-deb-pkg roles ###
-build_ossec_deb_pkg_dependencies:
-  - inotify-tools
-  - libssl-dev
-  - make
-  - tar
-  - unzip
-ossec_group: "ossec"
```
dolanjs: ossec_group var is still required need to add it back. It is used to set file permissions for the ossec gpg keyring used by procmail to encrypt the alerts.

build/build_generic_package.sh
```diff
@@ -1,12 +1,17 @@
 #!/bin/bash
-## Usage: ./build_grsec_metapackage.sh
+## Usage:
+##  ./build_package.sh securedrop-grsec
```
garrettr: Note: the build process for `securedrop-grsec` was actually moved into the new `freedomofpress/grsec` git repo.

install_files/ossec-server/usr/share/doc/ossec-server/changelog.Debian
```diff
@@ -0,0 +1,5 @@
+ossec-server (0.2.1-1) unstable; urgency=low
```
garrettr: Is this the right version number?

build/build_generic_package.sh
dolanjs: we can delete that example

install_files/ossec-server/usr/share/doc/ossec-server/changelog.Debian
dolanjs: the whole `install_files/ossec-server` directory can be deleted. since it moved to a separate repo now

PR #868, opened 2015-02-09
Get rid of useless gpg logging

garrettr:
Fixes:

* Logging noise due to `--no-use-agent` being deprecated in gpg2
* Apparmor conflicts with the creation of `/var/www/.gnupg`, which was due to the initial (but actually unnecessary) gpg version check. See [this PR comment](https://github.com/freedomofpress/securedrop/pull/863#discussion-diff-24270287) for context.

### DIFF
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -40,18 +40,7 @@ def do_runtime_tests():
 
 do_runtime_tests()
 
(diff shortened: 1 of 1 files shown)


dolanjs:
This will make looking at the logs soo much easier.

lgtm

PR #885, opened 2015-02-18
some minor cleanups

ThomasWaldmann:
most of them found either by pycharm codechecker or the w3c html5 validator (and I double-checked manually).

### DIFF
--- a/securedrop/source_templates/lookup.html
+++ b/securedrop/source_templates/lookup.html
@@ -64,5 +64,4 @@
   <i class="fa fa-lock pull-left"></i> Remember, your codename is: <strong>{{ codename }}</strong>
 </div>
 
-</div>
 {% endblock %}
--- a/securedrop/source_templates/lookup.html
+++ b/securedrop/source_templates/lookup.html
@@ -1,7 +1,7 @@
 {% extends "base.html" %}
 {% block body %}
 
-<div style="text-align: center;">
+<div class="center">
   {% include 'flashed.html' %}
 
   {% if flagged and not haskey %}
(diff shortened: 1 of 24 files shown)


ThomasWaldmann:
this should be rather easy to review. anyone?

garrettr:
@ThomasWaldmann Sorry for the delay! Reviewing now.

garrettr:
Just out of curiosity, did you find these issues manual or did you use some spell-check tools etc.? It might be a good idea to integrate whatever you used as a Git hook so we can avoid checking in similar mistakes in the future :)

securedrop/source_templates/lookup.html
```diff
@@ -1,7 +1,7 @@
 {% extends "base.html" %}
 {% block body %}
 
-<center>
+<div style="text-align: center;">
```
garrettr: SecureDrop is served with `Content-Security-Policy: default-src 'self'`, which blocks inline styles (see the Apache conf for the source interface in `install_files/ansible-base/roles/app/templates/sites-available/source.conf`). It would be better to specify this with a class or id, and define the style rules in `/static/css/styles.css`. There is already a general-purpose `.center` class that will do exactly what you want.

garrettr:
Once the inline style issue is fixed, this is ready to merge. Thanks @ThomasWaldmann!

ThomasWaldmann:
@garrettr i used pycharm's "code check" which includes a spell checker.

fixing the style issue right now...

garrettr:
Since these are all minor (with a corresponding minimal to no risk of breakage) but nice changes to have, I am going to merge this into both release-0.3.1 and the develop branch.

ThomasWaldmann:
great :)

Issue #886, opened 2015-02-18
Automatically enforce PEP8 for Python code


ThomasWaldmann:
there are some pep8 issues in the code - would you like to have them cleaned up?

If so, I'ld also suggest to use pytest-pep8, so no new issues will be introduced in future.

ThomasWaldmann:
see PR #890

garrettr:
The pep8 issues in the current code were fixed in #890. I'm changing the title of this issue to make it about auto-enforcing pep8 in the future.

ajvb:
PEP8 issue currently exist. Is this something we want to get rid of? I personally think the 80 character line limit is lame, but if it is the desire of the project, I'm all about it.

I have had great success with https://pypi.python.org/pypi/autopep8 in the past, I could help create a pre-commit git hook we stick in the repo and we could use Code Climate or similar for style enforcement on PR's?

ajvb:
Also, I'd be happy to run it throughout the application and test code and create a PR.

redshiftzero:
Yep! automatic PEP8 checking via autopep8 would be great. If you want to take that on it would be really helpful :+1:. We still need to define what else constitutes good style for this project (relevant: #47), once we do that we can add additional automated style checks.

ajvb:
Would be my pleasure. I'm conferencing this week and most of next so probably will be a little bit till I can complete it. On a side note, how do you guys feel about adding something like http://codeclimate.com/ with the github integration so every users PR's must follow pep8? There are other options, but I've had good success with CodeClimate and it is free for OSS.

fowlslegs:
I was looking into it last week actually. PEP8 checking would be useful, and Radon just looks like fun to play with. I would be open to trying it, considering it is free for open-source projects.

ajvb:
@fowlslegs Ok awesome :D

Fixed current PEP8 related issues in #1471 and added pep8 pre-commit hook

ThomasWaldmann:
BTW, in my projects, I have pep8 checking active (via pytest-pep8), but I don't enforce the 80chars limit, but set a limit of e.g. 120 chars.

80 has more bad effects imho than it does good - and we don't have vt80 terminals any more.

fowlslegs:
As an alternative to code climate, we could just run `pep8` in Travis. Does CodeClimate provide any benefit that would help contributor's identify and fix PEP8 errors faster?

ThomasWaldmann:
[...]

PR #890, opened 2015-02-24
Pep8, see issue #886

ThomasWaldmann:
not sure where pytest-pep8 should be added as a development requirement, I just installed it manually.

install:
pip install pytest-pep8

run:
py.test --pep8

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -321,7 +321,7 @@ def index():
         else:
             unstarred.append(source)
         source.num_unread = len(
-            Submission.query.filter(Submission.source_id == source.id, Submission.downloaded == False).all())
+            Submission.query.filter(Submission.source_id == source.id, Submission.downloaded is False).all())
 
     return render_template('index.html', unstarred=unstarred, starred=starred)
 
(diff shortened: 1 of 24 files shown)


securedrop/journalist.py
```diff
@@ -321,7 +321,7 @@ def index():
         else:
             unstarred.append(source)
         source.num_unread = len(
-            Submission.query.filter(Submission.source_id == source.id, Submission.downloaded == False).all())
+            Submission.query.filter(Submission.source_id == source.id, Submission.downloaded is False).all())
```
garrettr: Do we really want to use `is` instead of `==`? I understand that `is` will work, but that seems to be a special case that only works for booleans, small integers, etc. in CPython. Isn't `is` better used for comparing object identity, instead of value?

ThomasWaldmann: If one wants to specifically check for False, one wants "x is False". There exists only 1 False object.

If one wants to check for any falsish value (False, 0, [], ...), then "not x" is the right way.

x == False (or x == True) is considered bad style usually and even some pep8 checkers complain about it.

setup.cfg
```diff
@@ -0,0 +1,7 @@
+[pytest]
```
garrettr: :+1: It would be ideal to integrate linting with pep8 into our CI pipeline. We might also consider adding a git hook.

securedrop/journalist.py
garrettr: Ok, that makes sense to me. Thanks for clarifying.

garrettr:
Merging this into release-0.3.1 and develop.

PR #891, opened 2015-02-25
removed words from noun list

dolanjs:
removed some words that are or could be offensive by themselves or joined with a random adjective from noun list.


garrettr:
Merging this into both release/0.3.1 and develop. Note that I changed the name of the release branch from `release-0.3.1` to `release/0.3.1`. This was for no particular reason, I just like the slash better.

PR #909, opened 2015-03-17
Move upgrade tasks to ansible

dolanjs:
moved tasks related to deleting the old singing key and the old package names to ansible tasks. This has the advantage that ansible will handle most edge cases gracefully.

The next time the playbook is run it will install the correct packages. This will help for situations in staging where someone is not running grsec.


install_files/ansible-base/roles/upgrade/tasks/0-3pre-upgrade.yml
```diff
@@ -0,0 +1,21 @@
+---
+# This role will upgrade 0.3pre instances to 0.3 instances.
+# It will copy and run the upgrade.py script on the app and monitor servers.
+- name: remove old 0.3 packages to avoid version error
+  apt: name={{ item }} state=absent
+  with_items:
+    - securedrop-app-code
+    - securedrop-grsec
+    - securedrop-ossec-agent
+    - securedrop-ossec-server
```
garrettr: Does this work on both app and mon? For example, `securedrop-ossec-agent` is installed on app but not mon, and `securedrop-ossec-server` is installed on mon but not app. I assume the Ansible apt module does not throw an error because if the package was not installed in the first place, it's `state=absent`, otherwise it will be removed. Is that correct?

install_files/ansible-base/roles/upgrade/tasks/0-3pre-upgrade.yml
```diff
@@ -0,0 +1,21 @@
...
+
+- name: remove the previous signing key
+  apt_key: id=BD67D096 state=absent
+
+- name: copy upgrade script to servers
+  copy: src="0.3pre_upgrade.py" dest="/tmp" owner="root" mode="740"
+  sudo: yes
+
+- name: run the upgrade script
+  shell: "/tmp/0.3pre_upgrade.py {{ server_role }}"
```
garrettr: :+1: nicely done

install_files/ansible-base/roles/upgrade/tasks/0-3pre-upgrade.yml
dolanjs: correct, ansible will not throw an error because it just checks that it is absent. So it does work on both servers.

PR #910, opened 2015-03-17
Increase grsec install timeout to 5min

ageis:
Yesterday on recommended hardware I experienced this step taking longer than 200 seconds. Suggest raising the maximum runtime to allow for slow hardware and slow network connections.

### DIFF
--- a/install_files/ansible-base/roles/common/tasks/from_fpf_repo_install_grsec.yml
+++ b/install_files/ansible-base/roles/common/tasks/from_fpf_repo_install_grsec.yml
@@ -18,5 +18,5 @@
 
 - name: install grsec package from fpf repo
   apt: pkg={{ grsec_package }} state=latest
-  async: 200
+  async: 300
   poll: 10


garrettr:
I will merge this, but I am a little concerned. The grsec package is ~40M, so increasing the timeout to 5min implies we expect people with connection speeds of ~140kb/s to be able to install this. Is it possible that bandwidth is an issue on our end?

ageis:
Definitely the bandwidth here in the office is factor. In fact it takes 10 mins to download that package at 80kb/s (see experiment below... explains why I didn't just need to raise the async value, but also run playbook twice) which suggests 5min is not even enough. And that's not counting the install/setup part, which could add an additional minute. 

If it takes that long on this low bandwidth connection, in general we should widen the window to allow some edge cases since some networks might be throttled. Most people have at least ~140kb/s, but not all.

```
$ curl -o grsec.deb https://apt.freedom.press/pool/main/l/linux-source-3.14.32-grsec/linux-image-3.14.32-grsec_3.14.32-grsec-10.00.Custom_amd64.deb
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
 48 39.5M   48 19.3M    0     0  67262      0  0:10:16  0:05:01  0:05:15 78189
```

garrettr:
> Definitely the bandwidth here in the office is factor.

Definitely not the bandwidth in the office, it's not that bad. Problem is almost certainly on the server. IIRC Gandi throttles bandwidth if you go over caps or something so we probably did that and should provision more bandwidth.

I'm still going to merge this, since it can't hurt, but think the real solution is to increase bandwidth of the package server.

ageis:
You are correct the office rate of 1.63 MB/s is well above what is required. When I say a factor I mean just that, not giving it precedence over any other factor, but as a combination of factors. The limited bandwidth of the Gandi.net server is absolutely a more significant factor.

PR #911, opened 2015-03-17
custom header image example

dolanjs:
This pr changes the handling of the branded custom header image to be treated like the securedrop-app-code config.py file

Some orgs use a branded custom header image for their source interface to distinguish their instance from other SecureDrop instances

Previously if a custom header image `securedrop_header_image` var was defined in the ansible yml config file then the playbook would overwrite the logo.png that was included in the `securedrop-app-code` deb package.

Since the logo.png was NOT set as a conffile in the debian package when automatice updates would happen the logo would be replaced.

Now only a logo.png.example image is included in the `securedrop-app-code` deb package. In the `initialize_securedrop_app` role if the variable securedrop_custom_header is set to `None` then ansible will copy logo.png.example to logo.png

This has the downside of the deb package not being self contained and requiring configuration from ansible for it to work, but it also means we don't need to worry about any custom header images being overwritten during updates.


install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml
```diff
@@ -114,3 +114,12 @@
...
   when: securedrop_header_image != "None"
+
+  # When a custom header image is not provided copy the example into the
+  # correct place
+- name: if a branded custom header image is NOT defined use the example logo
+  command: /bin/cp logo.png.example logo.png
+  args:
+    chdir: "{{ securedrop_code }}/static/i"
+    creates: "{{ securedrop_code }}/static/i/logo.png"
+  when: securedrop_header_image == "None"
```
garrettr: Minor note: I'd like it better if we used an empty string to indicate no custom header image file. `""` is not a valid filename, so it can't be confused with `"None"`, which is a valid filename. I'm pretty sure YAML does not map `"None"` to Python's `None`, so there's no reason to mimic that convention.

garrettr:
> This has the downside of the deb package not being self contained and requiring configuration from ansible for it to work, but it also means we don't need to worry about any custom header images being overwritten during updates.

@dolanjs What would be the alternative?

install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml
dolanjs: I can make that change.

dolanjs:
@garrett The alternative would be to have ansible fill out a preloaded debconf config for it. So that the package could be self contained with the min for it to function without configuration from ansible.

garrettr:
@dolanjs build is still failing,

> TASK: [app | if configured move the custom header image into the image folder as logo.png] *** 
fatal: [localhost] => input file not found at /home/travis/build/freedomofpress/securedrop/install_files/ansible-base/roles/app/files/None or /home/travis/build/freedomofpress/securedrop/install_files/ansible-base/None

garrettr:
I'm pretty sure renaming `securedrop/static/i/logo.png → securedrop/static/i/logo.png.example` will break the development environment, since the development web server will expect to find the logo at `securedrop/static/i/logo.png`. Is there a way we can leave the logo where it was, but just not include it in the Debian package?

dolanjs:
@garrettr Copying the example image in place if the `securedrop_header_image` var is left empty is a task in `install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml` The `app` role including these tasks get run in the playbooks for the Travis, Development, Staging and Prod environments.

garrettr:
@dolanjs Yeah, but this is going to create a confusing situation for future development. Now developers will have to modify `logo.png.example` during development to test their changes, but must *remember* to *actually* make these changes to logo.png before committing. We already have this problem in one place (`config.py` vs `config.py.example`) and frankly it sucks and leads to mistakes all the time. I don't think we should continue using this anti-pattern (and eventually we should fix it in `config.py` too).

dolanjs:
@garrettr which way do you want to handle it.

1) remove the logo.png and config.py from the deb packages completely, so it is a file solely handled by ansible. 
2) include them in the deb package and mark them as conffiles. Ensuring that when cron job that apt-get runs with the flag to default to keep local modifications of files marked as conffiles.

garrettr:
> 2) include them in the deb package and mark them as conffiles. Ensuring that when the cron job that runs the security updates is configured with the flag to default to keep local modifications of files marked as conffiles.

I think this makes the most sense. Is it hard to implement?

dolanjs:
nope, will do it now

dolanjs:
May have an issue with development and travis environments where the file paths are different since the absolute path is suppose to be used. Will need to test that, that doesn't cause an issue across environments. 

" A package may contain a control information file called conffiles. This file should be a list of filenames of configuration files needing automatic handling, separated by newlines. **The filenames should be absolute pathnames**, and the files referred to should actually exist in the package. "  [1]


[1] https://www.debian.org/doc/debian-policy/ap-pkg-conffiles.html#sE.1

garrettr:
> May have an issue with development and travis environments where the file paths are different since the absolute path is suppose to be used. Will need to test that, that doesn't cause an issue across environments.

I'm pretty sure neither environment installs the .deb package (it just uses what's in the repo), so does it matter?

dolanjs:
@garrettr you correctt doesn't matter, not used in those environments at all

install_files/ansible-base/development-specific.yml
```diff
@@ -16,8 +16,8 @@ app_ip: "10.0.1.2"
 securedrop_code: "/vagrant/securedrop"
 # The securedrop_header_image has to be in the install_files/ansible-base/ or
 # the install_files/ansible-base/roles/app/files/ directory
-# Leave set to None to use the SecureDrop logo.
-securedrop_header_image: "None"
+# Leave set to empty to use the SecureDrop logo.
+securedrop_header_image: ""
```
garrettr: FYI, changing this from "None" to "" will probably cause merge conflicts with the current playbooks that are already on people's Admin Workstations, but they should be easy to resolve.

garrettr:
@dolanjs Can you rebase this from develop? Won't merge.

garrettr:
Rebased into one commit, cleaned up commit message, and merged into develop in 14c8761.

PR #915, opened 2015-03-18
remove the app IP from ossec alerts

dolanjs:
The OSSEC alerts for the app server include the internal IP address (which is also published in our public install guide). This will strip out the specific IP address from all subject.

### DIFF
--- a/install_files/ansible-base/roles/ossec_server/templates/send_encrypted_alarm.sh
+++ b/install_files/ansible-base/roles/ossec_server/templates/send_encrypted_alarm.sh
@@ -1,4 +1,3 @@
 #!/bin/bash
 
-/usr/bin/formail -I "" | /usr/bin/gpg --homedir /var/ossec/.gnupg --trust-model always -ear "{{ ossec_gpg_fpr }}" | /usr/bin/mail -s "$SUBJECT" {{ ossec_alert_email }}
-
+/usr/bin/formail -I "" | /usr/bin/gpg --homedir /var/ossec/.gnupg --trust-model always -ear "{{ ossec_gpg_fpr }}" | /usr/bin/mail -s "$(echo $SUBJECT | sed 's/{{ app_ip }}//g' )" {{ ossec_alert_email }}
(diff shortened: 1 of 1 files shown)


install_files/ansible-base/roles/ossec_server/templates/send_encrypted_alarm.sh
```diff
@@ -1,4 +1,3 @@
 #!/bin/bash
 
-/usr/bin/formail -I "" | /usr/bin/gpg --homedir /var/ossec/.gnupg --trust-model always -ear "{{ ossec_gpg_fpr }}" | /usr/bin/mail -s "$SUBJECT" {{ ossec_alert_email }}
-
+/usr/bin/formail -I "" | /usr/bin/gpg --homedir /var/ossec/.gnupg --trust-model always -ear "{{ ossec_gpg_fpr }}" | /usr/bin/mail -s "$(echo $SUBJECT | sed 's/{{ app_ip }}//g' )" {{ ossec_alert_email }}
```
garrettr: Is the `app_ip` guaranteed to be the only IP that ever appears in an OSSEC subject line? From skimming my backlog of OSSEC test alerts, that appears to be the cause, but I don't know if it it guaranteed to be. If not, I'd prefer to just use a simple regex to strip any IP addresses from the subject line.

dolanjs:
changed the script that encrypts the alarms to strip out a generic match for an IP address.

For the following categories of alerts the subject line will still contain:

Alert:
 * Agent Name
 * Alert Level

Report:
 * Report Name

Non-OSSEC mail sent to root@localhost:
 * The default minus whatever matches the generic IP address regex

garrettr:
This is really good, only one question - should we also remove a space in addition to the IP address?

Here's an example subject line from an OSSEC alert email containing an IP address: `OSSEC Notification - (app-staging) 10.0.1.2 - Alert level 10`. And here it is after being run through this regex: `OSSEC Notification - (app-staging)  - Alert level 10`. Note the two spaces before the second dash. It looks a little off and given my OCD nature will probably bug me and maybe others too. It's not a big deal, but it's also easy to fix. @dolanjs What do you think?

garrettr:
lgtm, thanks for fixing those nits!

PR #916, opened 2015-03-18
Autoremove unused packages after upgrading

### LINKED ISSUE #914: apt-get autoremove old kernels
As we install new kernels, they join the old kernel files in `/boot`. Since this partition is typically of a fixed size, it is possible for automatically installed kernel files to overflow the partition. When this happens, you can encounter boot problems (since often the latest installed kernel was not fully installed due to the lack of space in the partition).

The best solution to this problem is to periodically remove old kernels as you install new ones. However, it is not as simple as always removing the last kernel whenever a new one is installed. Sometimes a new kernel has conflicts with a particular set of hardware and can cause a system to fail to boot. In this case, it is good to maintain a backup of *at least* the previous working kernel to use in case you need to recover the system.

Fortunately, `apt-get autoremove` already has [special logic](https://github.com/freedomofpress/securedrop/issues/913#issuecomment-82629013) to handle this use case. Therefore it should be easy for us to do this by simply modifying cron-apt to autoremove after installing updates every 24h.

garrettr:
For SecureDrop, this is primarily useful because it avoids allowing too
many kernels to be installed in `/boot`, which can lead to issues
booting if the partition gets too full.

Closes #914.

### DIFF
--- a/install_files/ansible-base/roles/common/files/5-security
+++ b/install_files/ansible-base/roles/common/files/5-security
@@ -1,2 +1,3 @@
 autoclean -y
 dist-upgrade -y -o APT::Get::Show-Upgraded=true -o Dir::Etc::SourceList=/etc/apt/security.list
+autoremove -y


garrettr:
Tested this by:

1. Installing a package that installs lots of dependencies: `apt-get install emacs`
2. Removing that package, leaving some dependencies behind:

```sh
vagrant@app-staging:~$ sudo apt-get remove emacs
Reading package lists... Done
Building dependency tree
The following packages were automatically installed and are no longer required:
  acl at-spi2-core colord dconf-gsettings-backend dconf-service emacs24
  emacs24-bin-common emacs24-common emacs24-common-non-dfsg emacsen-common
  gconf-service gconf-service-backend gconf2-common ghostscript gsfonts
  imagemagick-common libatk-bridge2.0-0 libatspi2.0-0 libcairo-gobject2
  libcolord1 libcolorhug1 libcupsfilters1 libcupsimage2 libdconf1 libexif12
  libfftw3-double3 libgconf-2-4 libgd3 libgif4 libgphoto2-6 libgphoto2-l10n
  libgphoto2-port10 libgs9 libgs9-common libgtk-3-0 libgtk-3-bin
  libgtk-3-common libgudev-1.0-0 libgusb2 libieee1284-3 libijs-0.35
  libjbig2dec0 liblcms2-2 liblqr-1-0 libm17n-0 libmagickcore5 libmagickwand5
  libotf0 libpaper-utils libpaper1 libsane libsane-common libv4l-0
  libv4lconvert0 libvpx1 libwayland-client0 libwayland-cursor0 libxft2
  libxkbcommon0 libxtst6 m17n-contrib m17n-db poppler-data
Use 'apt-get autoremove' to remove them.
The following packages will be REMOVED:
  emacs
0 upgraded, 0 newly installed, 1 to remove and 0 not upgraded.
After this operation, 25.6 kB disk space will be freed.
Do you want to continue? [Y/n]
```
3. Running cron-apt and rebooting if reboot-required is there: `root@app-staging:/home/vagrant# cron-apt && test -e /var/run/reboot-required && reboot`
4. Verifying that all packages "automatically installed and are no longer required" are uninstalled:

```sh
root@app-staging:/home/vagrant# sudo apt-get autoremove
Reading package lists... Done
Building dependency tree
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

:+1:

PR #919, opened 2015-03-19
Set and boot grsec default kernel and remove generics, resolves #913

### LINKED ISSUE #913: Servers don't boot into grsec kernel automatically as GRUB default
This is a critical issue I found yesterday with the latest Ubuntu Server, and reproduced it again today.

Steps to reproduce:
* Install Ubuntu Server 14.04.2
* Install SecureDrop / run Ansible playbooks
* TASK: install grsec package from fpf repo
* TASK: setup cron-apt, perform safe upgrade of all packages
* TASK: reboot, server comes back

End result: Server boots generic kernel, subsequent playbook runs fail on the sysctl task with `cannot stat /proc/sys/kernel/grsecurity/*`. SecureDrop installation fails, only way to finish the install is to manually reboot and select the grsec kernel from the boot loader and run the playbook again, which still doesn't fix the default kernel selection so it would have to be selected on every boot.

I believe what's happening is that the new kernel installed during the safe upgrade task, right after the grsec kernel is installed, (`linux-image-3.16.0-31-generic`) is taking precedence over grsec in the boot loader. We need to fix this by making sure the grsec kernel (`linux-image-3.14.32-grsec`) is the GRUB default and first option.

ageis:


garrettr:
@ageis This looks good to me, test it? (and document how you tested it in a comment on this PR, please)

ageis:
I added back the full list of kernel metapackages. I think we should take all possible metapackages that exist between 14.04.1 and 14.04.2 and just make sure they're all removed. This will cover edge cases.

I added `apt-get remove` regexes to remove the headers and generic kernels. Although regex is regarded as less than ideal, based on all the package names in the Ubuntu main repo, I don't think we are aware of anything at this time that would inadvertently match these regexes. 

The reason I chose this approach rather than gathering `uname -r` as an initial fact and removing the corresponding package, is because facts do not retain state over subsequent playbook runs. If we had gone that route, it means that if we were already booted into grsec, and had to run the playbook again, then it would end up trying to remove the grsec kernel (unless of course we added some additional checks and logic to prevent that case).

ageis:
Wiped and reinstalled and tested again. This works successfully on both 14.04.1 and 14.04.2 and has the desired end result of leaving nothing but grsec. Moved autoremove to after `apt-get remove` so it functions more as a cleanup catching any leftovers.

The only thing I want to make sure of - the linux-headers regex would match the grsec headers IF they were installed. So you can't really run the playbook on a grsec build machine. There's no scenario where that's an issue, is there?

ageis:
As for testing, we used the Intel NUC hardware, and the monitor server was loaded with 14.04.2 and the app server with 14.04.1, so that we could test at the same time, and cover the differences between 3.13 vs. 3.14 vs. 3.16. The Ansible playbook was run from my laptop running Debian, out of the grsec-fix branch on my fork, and completed successfully. This PR adds an another reboot, and thus some time, to the install process.

PR #920, opened 2015-03-19
Fix to handle multiple grsec kernels

ageis:
There was an oversight in my line of shell code that would cause getting the grsec kernel string to fail if multiple grsec kernels are installed. Therefore you either need to include `head -1` to register only the first line or add `-m 1` to grep for the first match.


garrettr:
@ageis Can you submit a PR with *just* the commit with the fix for multiple grsec kernels?

Issue #896, opened 2015-03-03
Automatically run "SecureDrop Init" after boot


garrettr:
`tails_files/install.sh` creates a "SecureDrop Init" launcher for the journalist workstation. Double-clicking this launcher adds the HidServAuth values for the Document Interface to the system torrc and reloads Tor.

There is no reason to require this to be manually run, and forgetting to do this is a common cause of support complaints. We should instead use the [Tails dotfile persistence](https://tails.boum.org/doc/first_steps/persistence/configure/index.en.html#index13h2) to auto-run "SecureDrop init" at boot.

The only reason we haven't done this yet is because we haven't had a chance to test this. This needs to be tested thoroughly with Tails to make sure it reloads Tor correctly and at the right time for the changes to be persisted for the entire Tails session.

ageis:
Turns out that xsessionrc is not really the "right time" and can result in stuff like #985 on slow computers because of timing differences. A better solution is a NetworkManager hook which is now part of the #955 PR.

ageis:
There is a second reason not to use .xsessionrc. Turns out that Tails is using it to export a few environment variables, which I didn't notice until now probably since I had overwritten the file and didn't check what was in it before. =/

conorsch:
Resolved via #1018, which implemented Network Manager hooks for setting up the Tor config upon logging into Tails with persistence activated. Additional testing to be performed in #1299, prior to 0.3.7.

PR #922, opened 2015-03-20
Persistent SecureDrop Tails setup — admin automatic, journalist interactive

ageis:
Why not copy the launcher to the Desktop when `./install.sh` is run? It won't persist across reboots, but it will slightly increase usability and decrease steps post-install.

Only drawback: it could train the journalist or admin to expect it there and not look in the `~/Persistent` folder which they must do on subsequent boots.


ageis:
Actually, I just worked out a way to make the launcher persist on the Desktop using the Dotfiles persistence. This is implemented in the second commit. Tested and working!

micahflee:
Awesome!

garrettr:
@ageis Since you already have the dotfiles working, why not just get rid of the requirement to double-click the launcher, and just auto-run the program to setup the torrc at boot? That would close #896.

ageis:
This most recent commit eliminates the need for the launcher, makes the `torrc_additions` persistent, and installs persistent desktop bookmarks for both interfaces. It also runs `securedrop_init` on its own at the end. Resolves #896.

I have tested this and it's reliable.

Unfortunately, to create the launchers, `install.sh` now has a dependency on the Admin Workstation post-install, requiring the THS files generated by Ansible. If we choose to keep or fork that, then we should eliminate manual editing of torrc_additions too and handle that step automatically.

We don't have to have a separate script for journalists - I could just add some more stuff, like checking for the presence of files to figure out if it's an Admin vs. Journalist workstation, and launching gedit or setting up launchers based on that.

ageis:
The latest commit adds the handling discussed above so that `install.sh` works for both admins and journalists. The journalist setup is interactive while the admin's is fully automatic.

We should decide whether we want to merge this in time for 0.3.1 - I need to know within a couple days either way because of documentation.

garrettr:
@ageis There's a lot of good stuff in this PR, but:

1. It won't merge right now
2. We merged #925, which makes the torrc changes auto-run on boot and obviates the need for the Desktop launcher.

Can you rework this to include the improvements that you think would still be relevant, given #925?

garrettr:
Closing this in favor of #955.

PR #925, opened 2015-03-20
Persistent torrc SecureDrop initialization without clicking launcher

ageis:
Kept the icon around so it can be used for custom launcher creation.


tails_files/install.sh
```diff
@@ -19,16 +20,15 @@ apt-get update
...
+# copy init scripts
 cp securedrop_init.py $SCRIPT_PY
+cp .xsessionrc $INSTALL_DIR
+
+cp securedrop_icon.png $INSTALL_DIR
 
 # prepare torrc_additions
 cp torrc_additions $ADDITIONS
-gedit $ADDITIONS
+gedit $ADDITIONS > /dev/null 2>&1
```
garrettr: :+1:

garrettr: It might also be good to add a line (commented with a `#`) to the default torrc_additions that has instructions, e.g.

```
# Enter the HidServAuth lines for your authenticated hidden services here.
# When you're done, save the file and quit the editor.
```

Since sometimes people are confused that they need to quit the editor in order to continue. @ageis What do you think?

ageis: @garrettr yeah, I think that's a good idea actually.

ageis:
I've tested this extensively on Tails and it's solid.
As a note, the notification produced by `securedrop_init.py` doesn't show up on startup/session login. I believe that it just doesn't get called in time to get sent to Gnome's notification daemon.

PR #928, opened 2015-03-22
Optimize PNGs

ageis:
Optimizes all documentation images, in many cases reducing each one's size by up to 20%. This will only benefit us if we can filter/prune out the old revisions of images so that the repo doesn't take as long to download.


garrettr:
While this PR reduces the size of *each* png, I expect it will actually bloat the size of the repo overall (since I don't think Git will be able to efficiently diff between the current .pngs and the more-compressed versions). It is nice to have smaller PNG's, but this will make our "repo is too big" problems *worse* in the short term so I'm a little hesitant to merge it right now.

ageis:
@garrett can filter-branch + update-index be used to delete the old revisions?

garrettr:
Given that all of these PNG's are in `docs/`, and that a significant amount of the disk usage in the current repo is due to `/docs` (11M according to `du -h docs/`, almost all of it due to `docs/images`), this might be an argument to split out our documentation into a separate repo.

garrettr:
@ageis Yes, but doing that will:

1. Require that we re-sign all of our currently signed commits (it will changes the hashes that they are signed over)
2. Break merging whatever branches everyone has work on locally right now (for the same reason).
3. Render some of the history permanently inaccessible (that is the trade-off for removing things... once they're gone, you can't go back in time to when they were there)

So, while that is the the only way to solve the "repo is too big" problem, it is something that must be undertaken carefully and with sufficient warning to all contributors/consumers of this repo.

garrettr:
Here's the effect merging this PR has:

```sh
$ du -h .git
[..snip..]
 53M	.git
$ hub checkout https://github.com/freedomofpress/securedrop/pull/928
[..snip..]
$ du -h .git
[..snip..]
58M	.git
```

So it makes it 5MB larger. Note that a fresh clone is not 53M, that includes a ton of branches and stashes that I have in my local copy. A fresh clone is 41M (but merging this will still increase that by 5MB).

ageis:
Yeah I totally understand. Let's not merge it for now. I agree this is a good argument (and there are others too) for moving our docs to GitHub wiki. As I stated in the original PR it only makes sense if we can avoid it increasing the repo size.

garrettr:
Ok, I created a new issue to track "shrinking the repo", and mentioned this PR as something to re-consider merging once we're ready to rewrite history to shrink the repo. Closing this for now.

Issue #938, opened 2015-03-24
reboot_into_grsec is not idempotent


ageis:
It will reboot during the playbook even if grsec is already installed and running. Need to register a variable to check if reboot is required (see cron_apt.yml).

garrettr:
@dolanjs Should we nominate this for 0.3.2? Doesn't seem like it would affect existing installs, but will make re-provisioning less annoying.

conorsch:
Resolved by #939.

PR #939, opened 2015-03-24
Make grsec reboot idempotent

ageis:


install_files/ansible-base/roles/common/tasks/from_fpf_repo_install_grsec.yml
```diff
@@ -20,3 +20,48 @@
...
+  tags:
+    - reboot
+
+- name: waiting for server to come back
+  local_action:
+    module: wait_for
+      host={{ ansible_ssh_host }}
+      port={{ ansible_ssh_port }}
+      delay=45
+      state=started
```
dolanjs: Should also add the `when: not running_grsec.stat.exists` config to this task. If the task to reboot the server doesn't run because you are already booted into grsec then you do not need to wait for the server to come back up.

garrettr:
Does this fix a specific open issue? What is the benefit of making the reboot idempotent?

dolanjs:
@garrettr  reboots take awhile especially for existing installs where you need to wait for the tor hidden services to come back up.

garrettr:
@dolanjs So would you say we should target this for 0.3.2?

dolanjs:
@garrettr Think including this in 0.3.2 if fine. Including this change in a minor release will not require existing installs to re-run the playbook but will help reduce install frustration for new orgs.

garrettr:
@dolanjs K, I think that's worthwhile.

ageis:
Fixes #939

ageis:
Fixes #938

PR #944, opened 2015-03-24
fix pre0.3 to 0.3.1 upgrade script

dolanjs:
The pre0.3 SD version had an issue where deleting large submission would timeout, leaving the encrypted submission in place but removing it from the db. Added @garrett's fix for cleaning that up prior to running the db migration.

Re-ordered the sequence of events so the backup task and the migration happen prior to removing the old deb packages. This ensures that a custom logo.png will get backuped before dpkg deletes it.

Need to setup another pre0.3 instance to test these changes. Will update in comments when tested.

### DIFF
--- a/install_files/ansible-base/roles/upgrade/files/0.3pre_upgrade.py
+++ b/install_files/ansible-base/roles/upgrade/files/0.3pre_upgrade.py
@@ -9,6 +9,12 @@ import sys
 import tarfile
 import traceback
 
+store_dir = "/var/lib/securedrop/store"
+db_path = "/var/lib/securedrop/db.sqlite"
+assert os.path.isfile(db_path)
+conn = sqlite3.connect(db_path)
+c = conn.cursor()
+
 def backup_app():
     tar_fn = 'backup-app-{}.tar.bz2'.format(datetime.now().strftime("%Y-%m-%d--%H-%M-%S"))
     with tarfile.open(tar_fn, 'w:bz2') as t:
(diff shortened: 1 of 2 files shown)


garrettr:
> The pre0.3 SD version had an issue where deleting large submission would timeout

The worker timing out while `srm`'ing large submissions was fixed in 7cdbf0c9. Note, however, that this issue was actually due to a bug in python-gnupg which was worked around in 99ba2524. It was fixed for 0.3.

This issue is fixed upstream in python-gnupg, so we should remove the workaround as part of #946.

install_files/ansible-base/roles/upgrade/files/0.3pre_upgrade.py
```diff
@@ -26,6 +32,24 @@ def backup_mon():
...
+                (source_dir,)).fetchone()
+            print source_id, journalist_designation, source_dir
+        except (sqlite3.Error, TypeError) as e:
+            print "\n!!  Error occurred migrating replies for source {}".format(source_dir)
+            print "Source had {} submissions".format(len(os.listdir(os.path.join(store_dir, source_dir))))
+            #print traceback.format_exc()
+            print "Deleting source with no db entry..."
+            secure_unlink(os.path.join(store_dir, source_dir))
+            print
+            continue
```
garrettr: This `continue` statement is unnecessary.

install_files/ansible-base/roles/upgrade/files/0.3pre_upgrade.py
```diff
@@ -26,6 +32,24 @@ def backup_mon():
...
+    subprocess.check_call(['srm', '-r', path])
+
+def clean_large_deleted():
+    for source_dir in os.listdir(store_dir):
+        try:
+            source_id, journalist_designation = c.execute(
+                "SELECT id, journalist_designation FROM sources WHERE filesystem_id=?",
+                (source_dir,)).fetchone()
+            print source_id, journalist_designation, source_dir
+        except (sqlite3.Error, TypeError) as e:
```
garrettr: It would be better if this check were more explicit (general Python principle, "explicit is better than implicit"). We noticed the error because:

1. The `source_dir` was still present, because the `secure_unlink` had failed
2. The corresponding table in the DB had been deleted
3. As a result, the `SELECT` statement `WHERE filesystem_id=source_dir` returned `None`
4. A `TypeError` is raised when the line tries to unpack `None` into `source_id, journalist_designation`

Rather than catch the TypeError (which could theoretically be caused by something else), we should test explicitly to see if the given source is in the database. All exceptions should log errors and terminate execution (so we can inspect to see what is causing the issue and try to fix it up before the migration makes potentially destructive changes).

Issue #447, opened 2014-06-15
Dim the source interface when idle


ageis:
On some dark markets, they now have a security feature where if you're idle or change focus for a certain amount of time, the page totally dims itself, so if you walk away from the computer and leave it up then no one walking by can see what you were looking at. This seems like a great idea that we should implement for SecureDrop, and the best part is that it requires no JavaScript but is all done with CSS. For example, the code below will hide the page after 1 minute. Let me know what you think and I can submit a pull request!

```
<style type="text/css">
@keyframes fadein { from { opacity: 0; } to { opacity: 1; } }
@-webkit-keyframes fadein { from { opacity: 0; } to { opacity: 1; } } 
@keyframes fadeout { 0% { opacity: 1; } 98% { opacity: 1; } 100% { opacity: 0; } }
@-webkit-keyframes fadeout { 0% { opacity: 1; } 98% { opacity: 1; } 100% { opacity: 0; } } 
html {
        animation: fadeout 60s;
        -webkit-animation: fadeout 60s;
        opacity: 0;
}
html:hover {
        animation: none;
        -webkit-animation: none;
        opacity: 1;
}
</style>
```

garrettr:
Again this is interesting idea, but I concerned it would negatively impact usability for sources. Even @ageis admits it confused him the first time he encountered it. Perhaps the best course of action would be to merge it for 0.4, and let it bake for a while to see if we like it? This also seems like the kind of feature that would be good to evaluate with a user study.

PR #953, opened 2015-03-26
Fade out inactive source interface, implements #447

ageis:
After two minutes of inactivity, this fades out the source interface (in the interest of some marginal shoulder-surfing protection). It's in the base template so it should affect everything except the initial main page.


garrettr:
Is there any reason why this should be on every page except the main page?

garrettr:
Ok, I played with this feature a bit and I like it. Once we've resolved whether this should also happen on the main page (why not?) I think this is ready to merge.

ageis:
Sure, we can add it to the main page too. I just figured that when they first browse to a source interface, they might not need it since they are in front of the computer and in control, and not entering anything sensitive into text boxes yet. But seeing as the main page also has the logo most prominent so you can tell it's a leaking site, I agree it should also be dimmed and I'll add the CSS inclusion to index.html.

ageis:
@garrettr Added commit so it's ready to merge.

PR #955, opened 2015-03-26
Persistent automatic SecureDrop Tails setup

ageis:
#922 redone


tails_files/install.sh
```diff
@@ -1,17 +1,37 @@
 #!/bin/bash
+# SecureDrop persistent setup script for Tails 
 
 # check for root
 if [[ $EUID -ne 0 ]]; then
-    echo "This script must be run as root" 1>&2
-    exit 1
+	echo "This script must be run as root" 1>&2
```
garrettr: Please use consistent indentation throughout. You changed 2 spaces to 4 here, but later in the file there are instances of both 2 spaces and 4 spaces for indenting 1 level. Pick one (I prefer 2 spaces).

tails_files/install.sh
```diff
@@ -20,15 +40,44 @@ apt-get update
...
+    # create SSH host aliases and install them
+    SSHUSER=$(zenity --entry --title="Admin SSH user" --window-icon=$INSTALL_DIR/securedrop_icon.png --text="Enter your username on the App and Monitor server:")
+    cat > $INSTALL_DIR/ssh_config <<EOL
+Host app
+  Hostname $APPSSH
+  User $SSHUSER
+Host mon
+  Hostname $MONSSH
+  User $SSHUSER
+EOL
```
garrettr: Note that you can use `<<-EOL` to strip leading tabs, which would allow you to indent this with the rest of the block. However, I'm not sure if it will strip leading spaces or if they have to be actual tab characters. If they do, leave this the way it is - it's not worth mixing tabs and spaces to get the indentation to line up.

tails_files/install.sh
```diff
@@ -1,17 +1,37 @@
...
+HOMEDIR=/home/amnesia
+PERSISTENT=$HOMEDIR/Persistent
 INSTALL_DIR=$PERSISTENT/.securedrop
 ADDITIONS=$INSTALL_DIR/torrc_additions
 SCRIPT_PY=$INSTALL_DIR/securedrop_init.py
 SCRIPT_BIN=$INSTALL_DIR/securedrop_init
 DOTFILES=/live/persistence/TailsData_unlocked/dotfiles
+TAILSCFG=/live/persistence/TailsData_unlocked
+DESKTOP=$HOMEDIR/Desktop
+ANSIBLE=$PERSISTENT/securedrop/install_files/ansible-base
```
garrettr: This implies the repo must be cloned to this exact location for this script to work. It would be worth adding a check before continuing here.

garrettr: The check on line 30 is sufficient, *as long as* $ANSIBLE is only used if `$ADMIN == true`. I don't like that kind of logical dependency and would prefer this be made explicit.

tails_files/install.sh
ageis: @garrettr tried it, thanks for the tip. it does strip tabs from the output file. but if you try to use spaces instead to keep some indentation in the output file then I got an error.. so gonna leave it as-is.
`warning: here-document at line 2 delimited by end-of-file (wanted `EOL')`

tails_files/install.sh
```diff
@@ -20,15 +40,44 @@ apt-get update
...
-gedit $ADDITIONS > /dev/null 2>&1
+if $ADMIN; then
+  DOCUMENT=`cat $ANSIBLE/app-document-aths | cut -d ' ' -f 2`
+  SOURCE=`cat $ANSIBLE/app-source-ths`
+  APPSSH=`cat $ANSIBLE/app-ssh-aths | cut -d ' ' -f 2`
+  MONSSH=`cat $ANSIBLE/mon-ssh-aths | cut -d ' ' -f 2`
+  cat $ANSIBLE/app-ssh-aths $ANSIBLE/mon-ssh-aths $ANSIBLE/app-document-aths >> $ADDITIONS
+  if [[ -d "$HOMEDIR/.ssh" && ! -f "$HOMEDIR/.ssh/config" ]]; then
+    # create SSH host aliases and install them
+    SSHUSER=$(zenity --entry --title="Admin SSH user" --window-icon=$INSTALL_DIR/securedrop_icon.png --text="Enter your username on the App and Monitor server:")
```
garrettr: What happens if they typo and hit "ok"? What happens if they hit "cancel" by accident (does the script continue without `$SSHUSER` or does it give up?)

ageis: currently assumes they will enter it correctly and won't do either of those things. If you like I will add some validation to make it fool-proof. I just feel like in programming there is a healthy amount of validation vs. going overboard, especially in something like a Bash script.

tails_files/install.sh
```diff
@@ -20,15 +40,44 @@ apt-get update
...
+    chmod 600 $INSTALL_DIR/ssh_config
+    cp -p $INSTALL_DIR/ssh_config $HOMEDIR/.ssh/config
+    SSH_ALIASES=true
+  fi
+  # set ansible to auto-install
+  if ! grep -q 'ansible' "$TAILSCFG/live-additional-software.conf"; then
+    echo "ansible" >> $TAILSCFG/live-additional-software.conf
+  fi
+else
+# prepare torrc_additions (journalist)
```
garrettr: Please put a better commnet. "The journalist workstation does not have the `*-aths` files automatically created by the Ansible playbook, so the configuration file must be updated manually. Open a graphical text editor so the journalist can easily paste the `HidServAuth` info they will receive from the Admin" or something like that.

tails_files/install.sh
```diff
@@ -39,19 +88,56 @@ chown root:root $SCRIPT_PY
...
+
 chown amnesia:amnesia $INSTALL_DIR/securedrop_icon.png
-chmod 600 $INSTALL_DIR/securedrop_icon.png
-chown amnesia:amnesia $INSTALL_DIR/.xsessionrc
-chmod 700 $INSTALL_DIR/.xsessionrc
+chmod 600 $INSTALL_DIR/securedrop_icon.png 
+chown amnesia:amnesia $INSTALL_DIR/document.desktop $INSTALL_DIR/source.desktop $INSTALL_DIR/.xsessionrc
+chmod 700 $INSTALL_DIR/document.desktop $INSTALL_DIR/source.desktop $INSTALL_DIR/.xsessionrc
+
+# get addresses (journalist)
```
garrettr: Do we need zenity to do this? Once the journalist has filled out `torrc_additions`, can't we just extract this information (the interface .onion's) from there?

garrettr: Would it be better to use zenity instead of gedit to fill out torrc_additions? Make more sense to have specific, clearly labeled prompts instead of pasting a hunk of gibberish into a text editor.

ageis: @garrettr for the doc int yes but not if they want a shortcut for the source interface.

ageis: @garrettr yeah, I like it. Let's do that instead.

ageis:
Addressed comments in the latest commits, and switched from gedit to zenity for obtaining the journalist's HidServAuth value. There are two prompts for journalists, one for the HidServAuth value and one for the source interface .onion address. (you know, maybe journalists don't need this launcher/bookmark, so we could get rid of it. The admin is the only one who needs it to make test submissions, right?) For now I like keeping these separate rather than consolidating it into a single zenity form.

Still need to add the extra validation and fool-proofing that was asked about.

tails_files/install.sh
garrettr: K, sounds good :+1:

tails_files/install.sh
garrettr: @ageis This is still a blocker for merge. Additionally, it appears the most recent commits have introduced more indentation inconsistencies below.

tails_files/install.sh
garrettr: @ageis Looks like this is handled with the check `if [ ! -d "$ANSIBLE" ]` on line 30.

garrettr:
I'm going to test this on Tails today to double-check that it's working. Besides that, fix the indentation issues and this is ready to merge.

garrettr:
On second thought, I'm going to wait until James and Conor have finished their test install to test this. That way they'll already have an Admin Workstation (and credentials for the journalist workstations) for me to test this on.

ageis:
@garrettr indentation should now be consistently two spaces

tails_files/install.sh
garrettr: That's true; but any time you're processing user input, you should assume someone at some point will make every possible mistake, and account for that accordingly. You definitely don't want to get into a state that would be unrecoverable without significant manual intervention.

ageis:
@dolanjs pointed out that this automation should eventually cover other cases such as multiple admins, or admins who aren't supposed to possess the Doc Int HidServAuth values. just making a note here so I remember

ageis:
The latest commit updates the Ansible inventory with the new .onion hostnames.
Had to delete the word '.onion' in original inventory since I need an anchor to determine whether it's been updated yet or not.

tails_files/install.sh
```diff
@@ -20,15 +46,48 @@ apt-get update
 apt-get install -y build-essential
 gcc -o $SCRIPT_BIN securedrop_init.c
```
garrettr: Should `$SCRIPT_BIN` get copied to the `$INSTALL_DIR`?

ageis: The binary built by gcc is output to the persistent install dir:
SCRIPT_BIN=$INSTALL_DIR/securedrop_init

tails_files/install.sh
garrettr: Another note on this line: we can actually extract the SSH username from `ssh_users` in `prod-specific.yml`. However, the admin should know the username they used on the servers at this point in the installation and I don't see getting rid of this prompt as a requirement to merge this PR. Noting it here because it might be a nice future improvement.

tails_files/install.sh
```diff
@@ -20,15 +46,48 @@ apt-get update
...
 
-# prepare torrc_additions
-cp torrc_additions $ADDITIONS
-gedit $ADDITIONS > /dev/null 2>&1
+if $ADMIN; then
+  DOCUMENT=`cat $ANSIBLE/app-document-aths | cut -d ' ' -f 2`
+  SOURCE=`cat $ANSIBLE/app-source-ths`
+  APPSSH=`cat $ANSIBLE/app-ssh-aths | cut -d ' ' -f 2`
+  MONSSH=`cat $ANSIBLE/mon-ssh-aths | cut -d ' ' -f 2`
+  cat $ANSIBLE/app-ssh-aths $ANSIBLE/mon-ssh-aths $ANSIBLE/app-document-aths >> $ADDITIONS
```
garrettr: Note that if this script were run twice, you could end up appending these lines to the `$INSTALL_DIR/torrc_additions` *twice*. If you do that, when Tor restarts it will crash silently due to a particularly nasty bug in Tor (ask me how I know this). An easy solution would be to create the file with `>` instead of appending to it with `>>`.

garrettr: If you want to preserve the comments that are already in the file, add it as the first argument to `cat`.

garrettr:
Note: @conorsch is going to test this on his recently installed 0.3.2 test instance. He will test both the admin interface setup and the journalist interface setup, and will test both the "run on top of existing 0.3.2 instance" upgrade path and the "run on fresh Tails" new install path.

ageis:
I addressed the inadvisable appending to torrc_additions in the latest commit.

I just tested this PR with fake values in the *-aths files since I don't have an installation handy. I also tested it without those files to get the journalist perspective. With everything working as expected I have high confidence that it can soon be merged. There are some slight improvements discussed in these line comments and the possibility of extra validation that we can add later down the line.

ageis:
Latest commit is meant to address #982 so we moved the securedrop_init from xsessionrc to a NetworkManager hook. Now we are appending the HidServAuth values and reloading Tor only after Tor is ready and Tails has done everything it needs to do. Tested and working fine for me. All of this of course is a workaround until the persistent torrc option is implemented in Tails.

garrettr:
@ageis Is this ready to merge? lgtm although I haven't tested it.

ageis:
@garrettr Think so but we should probably have someone other than me test it. @conorsch was doing it and that's when we found the xsessionrc bug. depends whether you wanna merge now or wait until someone can fully get through this maybe next week.

garrettr:
Let's wait until we can test it one more time this week. This stuff is very hard to test in an automated way, we don't have any such tests right now, and it only gets run once after a production install so it doesn't even get much manual testing. As a result, flaws in `tails_files` can (and have in the past) easily go unnoticed. We should carefully review it next week (perhaps together, like how we did the group code review on an earlier iteration of this PR) before merging it.

ageis:
At some point (maybe before the hackathon) I'm willing to create a new PR for these usability enhancements to `tails_files`. The Bash stuff has been tested and reviewed more than we probably realize.

garrettr:
@ageis What's the status on this PR? Has it been completely superseded by others, or are there still changes in this PR that you think should be incorporated?

ageis:
@garrettr There's stuff in here that hasn't been merged and which I think I should open a separate, new PR for. Mainly improvements to usability and validation in the Tails install.sh - handling for admin vs. journalist roles, plus desktop (and Applications menu) shortcut icons for the interfaces.

garrettr:
@ageis Can you split those components out into a new PR (or PRs - it'd be nice to review this on a per-component basis, instead of one big pile with a lot of changes) that merges cleanly on top of develop?

garrettr:
Closing this, it was reworked in #1182 and merged.

PR #956, opened 2015-03-27
Even more PEP8 fixes

ddworken:
Just finished up another round of PEP8 fixes (50 more errors to go, then they are all finished!).


garrettr:
Thanks a lot, @ddworken!

PR #959, opened 2015-03-27
add ssl support to apache apparmor profile

### LINKED ISSUE #949: apparmor profile should support ssl
The apparmor profile currently will not allow orgs to enable ssl for their source/document interface. If the org locally modifies the profile then during the next update to securedrop-app-code will override those changes. We should add support for ssl apache modules and access to read the certs in the apparmor profile so it doesn't interfere with any sites that are using SSL

dolanjs:
Some orgs are already using HTTPS for the source interface. This adds the required modules and default cert locations to the apparmor profile for apache to use ssl.

This pr does not add ssl support to the ansible playbook. Just the apparmor support is being included now so that when unattended security updates for securedrop-app-code.deb package are applied it does not break the existing installs using ssl.

Fixes #949

### DIFF
--- a/install_files/ansible-base/usr.sbin.apache2
+++ b/install_files/ansible-base/usr.sbin.apache2
@@ -203,6 +203,11 @@
   /var/www/.gnupg/ rw,
   /var/www/.gnupg/** rw,
   /var/log/apache2/other_vhosts_access.log rw,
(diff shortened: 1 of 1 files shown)


garrettr:
Got confirmation that the profiles in this PR work with current production instances supporting TLS, so merging this for 0.3.2 and auto-updating should not break that TLS support.

PR #961, opened 2015-03-28
Use tmux by default for SecureDrop admin's login shell

### LINKED ISSUE #943: Make tmux or screen default shell on App and Monitor Servers
Once the servers are provisioned, they are only accessible via SSH through an authenticated Tor Hidden Service. This leads to a few potential problems:

1. Sometimes Tor circuits fail. If this happens while a long-running command is going, it will be killed. If screen/tmux were the default shell, the command would continue running and it would be easy to re-connect and reattach to the session to resume working were you left off.
2. If you don't use screen/tmux, you cannot restart Tor on either server (although most changes you ever need to make can be accomplished with `service tor reload`, which is safe). This has a bad failure mode: if you restart Tor while logged in to the SSH ATHS, the following happens:
    1. Tor shuts down, which kills your connection via THS protocol.
    2. When the connection dies, it kills your SSH session.
    3. When the SSH session dies, the `service tor restart` command is also killed (because it wasn't running in screen/tmux).
    4. Tor is never restarted.
    5. You are locked out of your system.

If we made tmux/screen the default shell (best way to do this is probably by launching tmux/screen automatically from `~/.bashrc` or similar), we could avoid these problems in the vast majority of cases. #2 is especially painful and easy to do by accident.

garrettr:
Closes #943.

### DIFF
--- a/install_files/ansible-base/roles/common/files/bashrc_securedrop_additions
+++ b/install_files/ansible-base/roles/common/files/bashrc_securedrop_additions
@@ -1,0 +1,9 @@
+# If not running interactively, do not do anything
+[[ $- != *i* ]] && return
+
+# Auto-launch tmux on login
+# https://wiki.archlinux.org/index.php/Tmux#Start_tmux_on_every_shell_login
+if which tmux >/dev/null 2>&1; then
+    #if not inside a tmux session, and if no session is started, start a new session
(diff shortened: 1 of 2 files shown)


dolanjs:
Please not that this fix will not apply to ansible logins only to when admin's ssh into the servers.

PR #965, opened 2015-03-31
Mark GRUB as manually installed (fixes #945)

### LINKED ISSUE #945: GRUB gets uninstalled in virtual machines, breaks re-provisioning
My code which was merged in #919 (fixes #913) includes a task that runs `apt-get autoremove` in order to clean up any dependencies left after the generic Linux kernel and headers packages and metapackages have been removed.

I noticed a couple nights ago that on a Vagrant VM, this has the effect of causing GRUB to be removed. I assume that this might be a peculiarity of the Ubuntu cloud image `trusty-server-cloudimg-amd64-vagrant-disk1.box`, where its kernel (linux-image-virtual/linux-virtual) has a dependency on GRUB. I'm fairly certainly we haven't seen this on hardware.

In any case, the VM still boots and everything. The problem with this is that it breaks our playbook in case you ever try to re-provision the VM.

As a fix we may need to add in detection for VMs (perhaps by checking if linux-image-virtual has been installed), and make it so that the grub paxctl task doesn't get run, or so that GRUB is not removed.

The first snippet shown below is the problematic task, and the second one is the error you get upon re-provisioning. 

```
TASK: [common | cleanup any dependencies left] ******************************** 
changed: [app-staging] => {"changed": true, "cmd": ["apt-get", "-y", "autoremove"], "delta": "0:00:01.635721", "end": "2015-03-25 05:52:18.262199", "rc": 0, "start": "2015-03-25 05:52:16.626478", "stderr": "debconf: unable to initialize frontend: Dialog\ndebconf: (Dialog frontend will not work on a dumb terminal, an emacs shell buffer, or without a controlling terminal.)\ndebconf: falling back to frontend: Readline\ndebconf: unable to initialize frontend: Readline\ndebconf: (This frontend requires a controlling tty.)\ndebconf: falling back to frontend: Teletype\nSaving menu.lst backup in /boot/grub/menu.lst_backup_by_grub2_prerm\nRunning update-grub Legacy to remove our core.img in it\n    Searching for GRUB installation directory ... found: /boot/grub\n    /usr/lib/grub-legacy/update-grub: line 116: grub: command not found\n    /usr/lib/grub-legacy/update-grub: line 116: grub: command not found\n    /usr/lib/grub-legacy/update-grub: line 153: grub: command not found\n    Searching for default file ... found: /boot/grub/default\n    Testing for an existing GRUB menu.lst file ... found: /boot/grub/menu.lst\n    Searching for splash image ... none found, skipping ...\n    Found kernel: /boot/vmlinuz-3.14.36-grsec\n    Updating /boot/grub/menu.lst ... done\n    ", "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\nThe following packages will be REMOVED:\n  grub-common grub-gfxpayload-lists grub-pc grub-pc-bin grub2-common os-prober\n0 upgraded, 0 newly installed, 6 to remove and 108 not upgraded.\nAfter this operation, 16.5 MB disk space will be freed.\n(Reading database ... 40241 files and directories currently installed.)\nRemoving os-prober (1.63ubuntu1) ...\nRemoving grub-gfxpayload-lists (0.6) ...\nRemoving grub-pc (2.02~beta2-9ubuntu1) ...\nRemoving grub2-common (2.02~beta2-9ubuntu1) ...\nRemoving grub-pc-bin (2.02~beta2-9ubuntu1) ...\nRemoving grub-common (2.02~beta2-9ubuntu1) ...\nProcessing triggers for man-db (2.6.7.1-1ubuntu1) ...\nProcessing triggers for install-info (5.2.0.dfsg.1-2) ...", "warnings": ["Consider using apt-get module rather than running apt-get"]}
```
- - - -
```
TASK: [common | make the required grub paxctl changes] ************************ 
failed: [app-staging] => (item=/usr/sbin/grub-probe) => {"changed": true, "cmd": ["paxctl", "-Cpm", "/usr/sbin/grub-probe"], "delta": "0:00:00.012804", "end": "2015-03-25 05:46:23.918155", "item": "/usr/sbin/grub-probe", "rc": 1, "start": "2015-03-25 05:46:23.905351", "warnings": []}
stderr: /usr/sbin/grub-probe: No such file or directory
failed: [app-staging] => (item=/usr/sbin/grub-mkdevicemap) => {"changed": true, "cmd": ["paxctl", "-Cpm", "/usr/sbin/grub-mkdevicemap"], "delta": "0:00:00.004108", "end": "2015-03-25 05:46:23.998212", "item": "/usr/sbin/grub-mkdevicemap", "rc": 1, "start": "2015-03-25 05:46:23.994104", "warnings": []}
stderr: /usr/sbin/grub-mkdevicemap: No such file or directory

FATAL: all hosts have already failed -- aborting
```

ageis:
This is intended to fix and address #945. Ultimately I came up with two equally viable solutions and decided to go with the simpler one, but I'll describe them here.

@dolanjs had suggested simply setting the `grub-pc` package to be installed. This makes sense at first, however if you put that before the autoremove task then it has no effect - grub-pc is already installed, and so it's still marked as automatically installed and Ansible doesn't set it as manually installed which is what we really need. It still gets removed - so then you would have to think about removing GRUB first and then re-installing it afterward, which doesn't really make sense.

So here's another idea. We could create a file named `99grub` in `/etc/apt/apt.conf.d` with the following contents:
```
APT
{
  NeverAutoRemove
  {
  "^grub.*";
  };

};
```
This has the expected effect that any GRUB packages are never auto-removed. You don't even have to `update_cache=true`, just install that config any time before you run autoremove and it just works. However, although we already have a number of custom config files in our playbooks, this is ultimately more complexity, and one extra file, than we really need.

So the simplest thing to do is to use `apt-mark` to mark grub-pc as manually installed, which will "prevent the package from being automatically removed if no other packages depend on it." This is what I did in the commit and it works very well.

### DIFF
--- a/install_files/ansible-base/roles/common/tasks/from_fpf_repo_install_grsec.yml
+++ b/install_files/ansible-base/roles/common/tasks/from_fpf_repo_install_grsec.yml
@@ -64,5 +64,8 @@
 - name: remove remaining kernel headers
   command: apt-get -y remove '^linux-headers-.*'
 
+- name: mark GRUB2 as manually installed so its not removed
+  command: apt-mark manual grub-pc
+
 - name: cleanup any dependencies left
   command: apt-get -y autoremove


garrettr:
lgtm. @dolanjs What do you think?

dolanjs:
lgtm

PR #975, opened 2015-04-02
Include MS Visio SecureDrop visualization

### LINKED ISSUE #274: Visualization of the SecureDrop architecture
Forking this from #249.

@aegis said:

> Would be great to also present a graphic visualization of the SecureDrop system, like the 'Cryptocat Threat Model Connections Overview' image https://github.com/cryptocat/cryptocat/wiki/Threat-Model

ageis:
Resolves #274, with the exception of converting the diagram to something more open and less proprietary.


garrettr:
Ok, this is a good enough solution for now. That is a good diagram, although I don't like the fact that it's impossible for anyone else to edit without Windows and proprietary Microsoft software.

PR #978, opened 2015-04-03
Rspec tests for apparmor iptables grsecurity

dolanjs:
This adds rspec tests for the following

* Tests for each iptables rule
** Modified an iptable rule comment to work with the iptables test.

* Tests for grsec packages
* Tests that system is booted into grsec
* Tests for grsec sysctl flags
* Tests for output of paxtest out
** Added an ansible task to the app-test role to install paxtest

* Test for apparmor packages
* Tests for presence of SD profile
* Tests expected profiles are being tracked
* Tests that no profiles are in complain mode
* Tests that no processes are in unconfined mode that have profiles defined.

### DIFF
--- a/install_files/ansible-base/roles/app-test/tasks/main.yml
+++ b/install_files/ansible-base/roles/app-test/tasks/main.yml
@@ -8,3 +8,4 @@
 - include: dev_setup_xvfb_for_functional_tests.yml
 
 - include: security_hardening_tests.yml
+  tags: non-development
--- a/install_files/ansible-base/roles/app-test/tasks/main.yml
+++ b/install_files/ansible-base/roles/app-test/tasks/main.yml
@@ -6,3 +6,5 @@
   tags: non-development,aa-complain
 
 - include: dev_setup_xvfb_for_functional_tests.yml
+
+- include: security_hardening_tests.yml
--- a/spec_tests/spec/localhost/iptables_spec.rb
+++ b/spec_tests/spec/localhost/iptables_spec.rb
@@ -2,22 +2,28 @@ require 'spec_helper'
 
 describe iptables do
   # These are the staging specific rules that should not be present in prod
-  it { should_not have_rule(' INPUT -p tcp -m tcp --dport 8080 -m state --state NEW,RELATED,ESTABLISHED -m comment --comment "Staging only allow direct access" -j ACCEPT'
-  it { should_not have_rule(' INPUT -p tcp -m tcp --dport 80 -m state --state NEW,RELATED,ESTABLISHED -m comment --comment "Staging only allow direct access" -j ACCEPT'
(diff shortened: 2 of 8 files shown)


install_files/ansible-base/roles/app-test/tasks/main.yml
```diff
@@ -6,3 +6,5 @@
   tags: non-development,aa-complain
 
 - include: dev_setup_xvfb_for_functional_tests.yml
+
+- include: security_hardening_tests.yml
```
dolanjs: This should be moved to a different location to avoid the travis playbook running these tasks which wont work in the travis environment.

conorsch: For now we can just add ` tags: non-development` to the `security_hardening_tests.yml` include. Down the road we should refactor the ansible config to leverage hostgroups so we can bundle up excludes like this for travis and other environments.

spec_tests/spec/localhost/iptables_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  it { should have_rule(' OUTPUT -p tcp -m tcp --sport 22 -m state --state NEW,RELATED,ESTABLISHED -j ACCEPT') }
+   # These rules should be present in prod and staging
+  it { should have_rule(' INPUT -p tcp -m state --state RELATED,ESTABLISHED -m comment --comment "Allow traffic back for tor" -j ACCEPT') }
+  it { should have_rule(' INPUT -i lo -p tcp -m tcp --dport 80 -m state --state NEW,RELATED,ESTABLISHED -m comment --comment "Allow tor connection from local loopback to connect to source int" -j ACCEPT') }
+  it { should have_rule(' INPUT -i lo -p tcp -m tcp --dport 8080 -m state --state NEW,RELATED,ESTABLISHED -m comment --comment "Allow tor connection from local loopback to connect to document int" -j ACCEPT') }
+  it { should have_rule(' INPUT -s 127.0.0.1/32 -d 127.0.0.1/32 -i lo -p tcp -m state --state RELATED,ESTABLISHED -m comment --comment "for redis worker all application user local loopback user" -j ACCEPT') }
+  it { should have_rule(' INPUT -s 8.8.8.8/32 -p tcp -m tcp --sport 53 -m state --state RELATED,ESTABLISHED -m comment --comment "tcp/udp dns" -j ACCEPT') }
```
dolanjs: The non 127.0.0.1 IP address in these test will need to be removed since that can change from one org to another.

dolanjs:
These test will need to be migrated to serverspec v2 format before they can be merged into develop.

dolanjs:
@conorsch is already migrating these to the serverspec v2 format which is what we will be using going forward.

PR #981, opened 2015-04-04
add tests for pages that require authentication in journalist.py

jacksingleton:


garrettr:
Fantastic, thanks for adding these. If we'd had them before, would've been able to avoid #974.

Issue #980, opened 2015-04-03
Add password validation during user creation


conorsch:
Currently the application permits empty strings for passwords during account creation. This affects account creation via both `manage.py` in the CLI and the web interface. At a minimum the app should check for empty strings and prompt again. Better yet, the app should enforce a minimum password strength via regex.

We also need tests to confirm this functionality (rejecting weak passwords) once it's added.

abbynormative:
I added a quick fix for this but you may want to use Cracklib instead or as well because they have a dictionary checker. https://pypi.python.org/pypi/cracklib

garrettr:
This needs some more unpacking for us to find an effective solution. I am wary of "password strength" metrics in general because they usually over-estimate password strength and often impose barriers to usability.

For example:

1. Checking for a minimum length is a good idea, but is not enough by itself. A password may be long but easily cracked by a dictionary attack (e.g. `qwerytasdfgzxcvb`).
2. Checking for a diversity of character types (upper/lower/number/symbol) helps in theory, but:
    1. It imposes barriers to memorability that often lead humans to choose passwords that look good but are actually vulnerable to dictionary attacks, or choose passwords that are way too hard to remember (https://xkcd.com/936/).
    2. Disallows passwords that may not have a diversity of character types, but are still highly secure (e.g. Diceware passphrases, or long random passphrases that are only alpha characters to aid in typing, especially on touchscreens).

Our documentation tells journalists and admins to create this password using KeePassX, so it will be long, random, and contain mixed character types (default is 25 characters of lower/upper/number/symbol). It seems reasonable to make this *recommendation* a **requirement** in the password validation code. Doing so creates a minor inconvenience for application developers (who have to log in to the Document Interface frequently while testing), but after some debate with @dolanjs I think this is worth it for the higher level of security assurance.

The best way to implement this is probably to validate for passwords that match the default settings in KeePassX on Tails (which is what journalists and admins use for their workstations). The tricky thing in implementing this will be dealing with the fact that a randomly chosen password *might not* contain characters from every group. KeePassX has a checkbox to "Ensure that password contains characters from every group", but it is not on by default.

![screen shot 2015-04-10 at 5 46 14 pm](https://cloud.githubusercontent.com/assets/407302/7099532/8c941ba4-dfa9-11e4-8140-795a6d48f33d.png)

abbynormative:
Good to know. I'll think about it and try to have something better next week. Thanks Garrett!

[...]

PR #983, opened 2015-04-08
add password checker

abbynormative:
fix for issue #980
https://github.com/freedomofpress/securedrop/issues/980

### DIFF
--- a/securedrop/manage.py
+++ b/securedrop/manage.py
@@ -167,11 +167,45 @@ def add_admin():
             break
 
     while True:
+        print "Please enter a password.\n\nThe password must be at least 8 characters and contain a combination of upper\nand lower characters, numbers and special characters.\n"
         password = getpass("Password: ")
         password_again = getpass("Confirm Password: ")
         if password == password_again:
-            break
-        print "Passwords didn't match!"
+            print "Your password is " + str(len(password)) + " characters long."
+            strength = ['blank','very weak','weak','medium strength','medium-strong','strong', 'very strong']
+            score = 0
+            if len(password) == 0:
+                score = 0
+                print "It looks like you didn't enter any text for your password."
+            elif len(password) < 6:
+                score = 0
+                print "Your password is " + strength[score] +"."
+            elif len(password) >= 6 and len(password) < 8:
+                score = 1
+                print "Your password is " + strength[score] +"."
+            elif len(password) >= 12 and re.search('\d+', password) and re.search('[a-z]', password) and re.search('[A-Z]', password) and re.search('[^\w\s]', password):
+                score = 6
+                print "Your password is " + strength[score] +"."
+            elif len(password) >= 8 and re.search('\d+', password) and re.search('[a-z]', password) and re.search('[A-Z]', password) and re.search('[^\w\s]', password):
+                score = 5
+                print "Your password is " + strength[score] +"."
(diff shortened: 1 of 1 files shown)


garrettr:
@abbynormative Thanks for writing this! This is a good start, but please see https://github.com/freedomofpress/securedrop/issues/980#issuecomment-91731038 for a different approach, which I recommend.

I will also add a few line comments to the code to give you some things to think about.

securedrop/manage.py
```diff
@@ -167,11 +169,45 @@ def add_admin():
             break
 
     while True:
+        print "Please enter a password.\n\nThe password must be at least 8 characters and contain a combination of upper\nand lower characters, numbers and special characters.\n"
```
garrettr: We try to avoid having lines that are > 80 characters long (see PEP8). There are several ways to fix this in Python, but here I would recommend using a multiline string because the string contains newlines.

```py
print """Please enter a password.

The password must be at least 8 characters and contain a combination of upper
and lower characters, numbers and special characters.
"""

File: securedrop/manage.py [hunk 2]
```diff
@@ -167,11 +169,45 @@ def add_admin():
...
         password = getpass("Password: ")
         password_again = getpass("Confirm Password: ")
         if password == password_again:
-            break
-        print "Passwords didn't match!"
+            print "Your password is " + str(len(password)) + " characters long."
+            strength = ['blank','very weak','weak','medium strength','medium-strong','strong', 'very strong']
```
garrettr: You don't need so many possible evaluations of the password strength. "Acceptable" and "not acceptable" would be enough :smile:

securedrop/manage.py
```diff
@@ -167,11 +169,45 @@ def add_admin():
             break
 
     while True:
+        print "Please enter a password.\n\nThe password must be at least 8 characters and contain a combination of upper\nand lower characters, numbers and special characters.\n"
         password = getpass("Password: ")
         password_again = getpass("Confirm Password: ")
         if password == password_again:
-            break
-        print "Passwords didn't match!"
+            print "Your password is " + str(len(password)) + " characters long."
```
garrettr: All of this code should be factored into a separate, reusable function, to avoid making this function overly long. A good rule of thumb is that any one function should be able to fit on your screen all at once without scrolling. I would've written a function, say `def acceptable_password(password)`, and checked it:

```py
if not acceptable_password(password):
    print "Your password is not strong enough" # TODO explain why, you could do this neatly with exceptions
else:
    break
```

garrettr:
None of the line comments on this PR were ever addressed, and since it has been sitting here for a while I'm going to close it.

In general, I'm no longer certain that checking for password strength is the right approach for the Document Interface passwords. It's probably better to just generate known-good passwords for journalists, like we already do for sources (Diceware for the codename), as @dolanjs suggested in https://github.com/freedomofpress/securedrop/issues/980#issuecomment-92471105.

Further, SecureDrop generally requires end users (especially journalists and admins) to keep track of *way* too many credentials (see the [Passphrases doc](https://docs.securedrop.org/en/latest/passphrases.html) for a list). Rather than focusing narrowly on how we can ensure that those credentials are sufficiently strong, we should focus on reducing the number of required credentials to the bare minimum necessary to achieve our security level. Anything more reduces usability, which tends to reduce security (often catastrophically) because it encourages users to make compromises (write passwords down on a post-it, reuse credentials across security domains, etc.).

Issue #982, opened 2015-04-07
Reloading Tor on Tails at the start of X session could fail in some configurations


ageis:
As part of PR #955 we moved the Tor reload and the appending of torrc_additions → torrc to the start of the X session (i.e. immediately after login) instead of requiring the user to click on a launcher. Today on one machine running Tails 1.3.2 we found a bug where Tor will not reload successfully at this stage, with the following error:

```
[warn] Failed to parse/validate config: While Tor is running, changing Sandbox is not allowed
```
This is because Tails runs Tor with some local settings including "Sandbox 1", which utilizes the seccomp facility of the Linux kernel. But changing the Sandbox setting is not allowed without a full restart. The problem with reloading at this stage is that Tails is actually appending the "Sandbox 1" line and making torrc modifications some time later. Per the Tails source code, the sandboxing is added if you select the default option not to start Tor Launcher in Tails Greeter, and then Tor is started by a NetworkManager hook when a network link is detected. (see `/etc/NetworkManager/dispatcher.d/10-tor.sh`)
 
```
# The Tor syscall sandbox is not compatible with managed proxies.
# We could possibly detect whether the user has configured any such
# thing via Tor Launcher later (e.g. in 60-tor-ready.sh),
# but then we would have to restart Tor again to enable the sandbox.
# Let's avoid doing that, and enable the Sandbox only if no special Tor
# configuration is needed. Too bad users who simply need to configure
# a HTTP proxy or allowed firewall ports won't get the sandboxing, but
# much better than nothing.
if [ "$(tails_netconf)" = "direct" ]; then
   tor_set_in_torrc Sandbox 1
fi
```
While the seccomp sandbox was added with Tails 1.3, this particular issue would have a higher chance of presenting on slower machines and connections which is why I never observed it. My first idea to deal with this was to add a "sleep" command for a few minutes to the .xsessionrc before we run securedrop_init, but that's contingent on there being a working ethernet connection or WiFi access point saved in persistence that will be ready in the right amount of time. We need a solution that won't allow any race conditions and will work across a variety of hardware. I figured that we might have a loop that waits for Tails to add the "Sandbox 1" before running our Tor reload. But the solution also needs to survive future changes to the software.

[...]

PR #994, opened 2015-04-14
Enable seccomp sandboxing for Tor

ageis:
One-line change to the App+Mon torrc to utilize the seccomp2 syscall filtering facility on Linux to limit which system calls Tor can invoke. This feature first appeared in 0.2.5.1-alpha (2013-10-02) with the disclaimer "This is an experimental, Linux-only feature to provide defense-in-depth against unknown attacks." Since then it has improved and seen many bugfixes through 2014. Now all builds of Tor for jessie, trusty or later include seccomp support, and it's being used in production on Tails.

I've only tested it in the virtual staging environment; we should probably increase our automated testing coverage before making this change. 

**An important note**:
Per #982, the 'Sandbox' value can't be changed without a full restart. If you try to do a Tor reload after changing this line, Tor will fail to parse the config and actually exit. So if this is updated as part of a playbook run we will have to include a Tor restart and a wait period for the SSH connection to come back, otherwise merging this change could be catastrophic and lock you out requiring a reboot or console access.


garrettr:
> So if this is updated as part of a playbook run we will have to include a Tor restart and a wait period for the SSH connection to come back, otherwise merging this change could be catastrophic and lock you out requiring console access.

We should include the required changes to the playbooks to support this before merging this (mainly so we don't forget to do it).

Otherwise I think this a good change to make and agree that it seems to have seen enough testing in the broader ecosystem.

dolanjs:
@ageis you did change this in the correct spot in the playbook but will need to make the changes to the spec tests for this also.
`spec_tests/spec/common-staging/tor_spec.rb`

Also explain or paste the output of testing that shows the affected features still work please.

ageis:
I added the line to the spec test.

Per [our task which installs torrc](https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/common-app/tasks/app_configure_tor_hidden_services.yml), Tor is restarted [via a handler](https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/common-app/handlers/main.yml) right after the file is put in place. I don't know what that means for running the playbook over Tor, as there is not a wait task, but if the restart is already there I think that means this is safe to merge.

ageis:
I should add that I tested with Sandbox enabled in the staging environment and all of the interfaces and sshd were reachable. Nothing untoward was noticed in the logs (AppArmor, syslog, OSSEC) either.

garrettr:
We need to keep an eye on this patch, as it will be extremely annoying if there's any way it could lead to breakage of running (or re-running, or re-running just specific parts of, via tags) the Ansible playbooks. I think the best course of action is to merge it now, so we can start exercising it during development (in the staging and production environments), and hopefully any bugs it might have will shake out before the next stable release.

PR #998, opened 2015-04-17
Strip leading + trailing whitespaces from codename

ageis:
I think it's a smart idea to strip leading and trailing white spaces from a codename entered by the source in order to work around the inclusion of unintentional spaces and resulting issues with logging in.

### DIFF
--- a/securedrop/tests/test_unit_source.py
+++ b/securedrop/tests/test_unit_source.py
@@ -129,6 +129,28 @@ class TestSource(TestCase):
             self.assertIn('Sorry, that is not a recognized codename.', rv.data)
             self.assertNotIn('logged_in', session)
 
+    def test_login_with_whitespace(self):
+        rv = self.client.get('/login')
+        self.assertEqual(rv.status_code, 200)
+        self.assertIn("Login to check for responses", rv.data)
+
+        codename = self._new_codename()
+	codename = ' ' + codename + ' '
+        with self.client as c:
+            rv = c.post('/login', data=dict(codename),
+                        follow_redirects=True)
+            self.assertEqual(rv.status_code, 200)
+            self.assertIn("Submit documents and messages", rv.data)
+            self.assertTrue(session['logged_in'])
+            common.logout(c)
+
+        with self.client as c:
+            rv = self.client.post('/login', data=dict(codename='invalid'),
+                                  follow_redirects=True)
+            self.assertEqual(rv.status_code, 200)
+            self.assertIn('Sorry, that is not a recognized codename.', rv.data)
+            self.assertNotIn('logged_in', session)
+
     def _dummy_submission(self):
         """
         Helper to make a submission (content unimportant), mostly useful in
--- a/securedrop/tests/test_unit_source.py
+++ b/securedrop/tests/test_unit_source.py
@@ -130,26 +130,24 @@ class TestSource(TestCase):
             self.assertNotIn('logged_in', session)
 
     def test_login_with_whitespace(self):
-        rv = self.client.get('/login')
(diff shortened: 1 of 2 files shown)


garrettr:
This is very straightforward. Could you add a unit test?

ageis:
I've never written a unit test before, so it would take me some time to figure out how to do it. The test, to be added to `test_unit_source.py`, would be to show that a codename with leading or trailing whitespaces is still recognized as valid?

garrettr:
Sorry, I thought I commented on this already but I must've forgotten to submit it or something. You are correct, it should be added to `test_unit_source.py` and the test you propose is a good one. At a minimum, your test should fail without the change you made and succeed with the change.

ageis:
I copied the test_login_and_logout method and just concatenated some whitespaces to the codename. 
https://github.com/ageis/securedrop/commit/710ea19bfd9cfa0585d6eafff525c772a1142dae
Is that the right idea? Might need a primer on the unit testing framework.

securedrop/tests/test_unit_source.py
```diff
@@ -129,6 +129,28 @@ def test_login_and_logout(self):
             self.assertIn('Sorry, that is not a recognized codename.', rv.data)
             self.assertNotIn('logged_in', session)
 
+    def test_login_with_whitespace(self):
+        rv = self.client.get('/login')
+        self.assertEqual(rv.status_code, 200)
+        self.assertIn("Login to check for responses", rv.data)
+
+        codename = self._new_codename()
+	codename = ' ' + codename + ' '
```
garrettr: Indent this correctly. You do not want to mix tabs and spaces, that usually breaks Python. We use spaces for indentation.

securedrop/tests/test_unit_source.py
```diff
@@ -129,6 +129,28 @@ def test_login_and_logout(self):
...
+        with self.client as c:
+            rv = c.post('/login', data=dict(codename),
```
garrettr: `dict(codename=codename)`, see [docs](https://docs.python.org/2/library/stdtypes.html#dict).

securedrop/tests/test_unit_source.py
```diff
@@ -129,6 +129,28 @@ def test_login_and_logout(self):
             self.assertIn('Sorry, that is not a recognized codename.', rv.data)
             self.assertNotIn('logged_in', session)
 
+    def test_login_with_whitespace(self):
```
garrettr: Please add a docstring to explain what this is testing and why. It would be good to reference the original issue and/or this PR.

securedrop/tests/test_unit_source.py
```diff
@@ -129,6 +129,28 @@ def test_login_and_logout(self):
...
+        with self.client as c:
+            rv = c.post('/login', data=dict(codename),
+                        follow_redirects=True)
+            self.assertEqual(rv.status_code, 200)
+            self.assertIn("Submit documents and messages", rv.data)
+            self.assertTrue(session['logged_in'])
+            common.logout(c)
+
+        with self.client as c:
+            rv = self.client.post('/login', data=dict(codename='invalid'),
```
garrettr: I don't think you need this test at all...

securedrop/tests/test_unit_source.py
garrettr: It would also be good to test likely variants, such as `' ' + codename` and `codename + ' '`.

ageis:
Updated based on your comments. I'm a total Python noob, hope I did that right.

garrettr:
lgtm. Thanks for adding the test!

Issue #1006, opened 2015-04-23
File and directory permissions should be declared consistently


conorsch:
There are a few tasks where the Ansible config is explicitly declaring permissions that don't match up on the host. The serverspec tests in #1002 caught a few of these, such as:

```
- name: ensure send_encrypted_alarm script is present
  template:
    src: send_encrypted_alarm.sh
    dest: /var/ossec/send_encrypted_alarm.sh
    mode: 775
    owner: ossec
    group: root
  tags:
    - procmail
    - permissions
```

At a glance, it looks like that tasks ensures sets permissions on `/var/ossec/send_encrypted_alarm.sh` to `0775`. Not so:

```
root@mon-staging:~# stat -c %a /var/ossec/send_encrypted_alarm.sh
1407
```

The culprit is how Python, and therefore Ansible, handles decimal versus octal numbers. In Python 2.7, preceding a decimal number with a zero will ensure it's interpreted as an octal number. 
### Options for formatting file permissions

All of the options below, when fed to the `mode:` parameter in an Ansible task, will result in the same permissions, i.e. `0755`:
- `"775"`
- `"0775"`
- `0775`
- `00775`
- `0o775`
- `0o0775`

Of those options, I consider `00775` the best format, since it's both readable and unambiguous. It's worth noting that `0o0775` works with both Python 2.6 and Python 3.x, and is actually required from 3.x onwards. Despite the [looming deprecation of Python 2](https://lists.debian.org/debian-devel-announce/2015/04/msg00005.html), Ansible [isn't planning to port to Python 3 anytime soon](http://docs.ansible.com/faq.html#how-do-i-handle-python-pathing-not-having-a-python-2-x-in-usr-bin-python-on-a-remote-machine). Even if Ansible does move, updating our configs to use the `0o` prefix rather than `0` would be trivial.
### One-liner syntax

When using the `key=value` syntax in Ansible tasks, [integer values are coerced to strings](https://github.com/ansible/ansible/issues/9196), and passed to Python's `os.chmod` accordingly, meaning that `mode=775` works exactly the same as `mode: "755"`. With the exception of extremely terse tasks, multiline format should be preferred over the one-liner syntax, since it makes diffs in version control much more readable. (I'll submit a PR soon that adjusts most of our config to multiline syntax, so upcoming changes are clearer for review.)
### Files affected

Running `grep -ri "mode: [^\'\"]"` in the `install_files/ansible-base` directory shows these files which warrant further inspection:
[...]

Issue #818, opened 2014-12-23
OSSEC now supports authd client authentication


dolanjs:
OSSEC now supports "Optional Client Authentication" for the authd dameon. [1] The authd process is used to initially create and distribute the pre shared secret key used for the ossec server agent communication. Currently this is an unauthenticated process that is done only during the initial installation and if needed while re-ip'ing an existing install. Otherwise the process is not enabled.

Enabling the authd client authentication would require generating and distributing a signed client/CA certs to the agent/server with ansible during the initial install.

[1] http://ossec-docs.readthedocs.org/en/latest/programs/ossec-authd.html

garrettr:
@dolanjs Can you describe a little more of the benefit that would be realized with client authenticator for the authd daemon? IIRC we only run this program once during the installation process, so the risk of compromise at that point is fairly minimal. It also seems to incur some significant administrative overhead, in the form of generating, distributing, and maintaining a CA/CA certs.

dolanjs:
@garrettr I don't really see what we would gain from this either. The ossec agent/server authd connection does not leave the SD environment and it would cause a significant setup/maintenance overhead. For those reasons think we can not implement this and close the issue.

PR #1007, opened 2015-04-24
Ansible multiline format

conorsch:
Updates Ansible tasks to use multiline format. The goal is to expand the provisioning tasks over several lines, so diffs are more informative and easier to review for collaborators. The reformatting also introduces a consistent structure across roles, which will help during reorg. Added or revised comments throughout, in order to improve clarity.

No functionality changes in this PR, strictly formatting and comments. In particular, this PR does not address the file permission changes discussed in #1006. All file permissions remain the same, and a separate PR will modify those, as well add corresponding spec tests. All existing spec tests for development, build, app-staging, and mon-staging passed after re-provisioning with the reformatted Ansible tasks in this PR.


install_files/ansible-base/roles/ossec_server/tasks/authd.yml
```diff
@@ -1,7 +1,8 @@
 ---
-  # The authd ossec process automatically adds the app server ossec agent it is
-  # not an authenticated process. There is an option we could add to create and
-  # verify the server/agents ssl keys with the root CA cert. see
+  # The authd ossec process automatically adds the app server ossec agent.
+  # This addition is not an authenticated process. There is an option
+  # we could add to create and  verify the server/agents ssl keys with
+  # the root CA cert, see:
   # http://ossec-docs.readthedocs.org/en/latest/programs/ossec-authd.html
```
dolanjs: This comment was related to #818 since we closed that issue we can remove this comment now too.

install_files/ansible-base/roles/backup/tasks/main.yml
```diff
@@ -3,21 +3,33 @@
...
+  #   - tor services directory
+  #   - securedrop_data directory
+  #   - securedrop_code/config.py
+  #   - custom header image (if applicable)
+  # These files are packed into a gpg encrypted zipfile with filename
+  # format: sd-backup-TIMESTAMP.zip.gpg
 - name: run the collect script to backup all the needed files to perform a restore
-  shell: chdir=/tmp/ /tmp/0.3_collect.py
+  # IMPORTANT: chdir must be specified in one-liner syntax when using async.
+  # This is a bug in Ansible, see: https://github.com/ansible/ansible/issues/9693
```
garrettr: Thanks, love comments like this for workarounds :+1:

install_files/ansible-base/roles/common/tasks/apply_grsec_lock.yml
```diff
@@ -32,11 +32,18 @@
       delay=45
       state=started
   when: not grsec_lock.stat.exists
+  # Local wait action, doesn't require sudo privileges.
```
garrettr: If you didn't set `sudo: false`, what would happen? Would that mean you would need to run the playbook as root? Would Ansible error out when you started the playbook run, or only fail once it got to this task? This isn't important, I'm just curious.

install_files/ansible-base/roles/common/tasks/create_users.yml
```diff
@@ -41,8 +44,8 @@
 
 - name: add line to each users bashrc to source the SecureDrop additions
   lineinfile:
-    dest: "/home/{{ item }}/.bashrc"
-    line: ". /etc/bashrc.securedrop_additions"
+    dest: /home/{{ item }}/.bashrc
+    line: '. /etc/bashrc.securedrop_additions'
```
garrettr: Why single quotes here? Double quotes seem to be used almost everywhere else. Do you need the quotes at all?

install_files/ansible-base/roles/ossec_server/tasks/authd.yml
garrettr: I agree with @dolanjs, I don't think the security benefit is worth the added overhead and complexity in this case. It makes more sense for a more open network that needs to dynamically add and remove OSSEC agents.

install_files/ansible-base/roles/ossec_server/tasks/mon_configure_ossec_gpg_alerts.yml
```diff
@@ -13,12 +17,16 @@
...
   copy:
     src: "{{ ossec_alert_gpg_public_key }}"
-    dest: "/var/ossec"
+    dest: /var/ossec
   tags:
     - gpg
 
 - name: ensure ossec alert gpg public key is in the ossec keyring
-  command: "su -s /bin/bash -c \"gpg --homedir /var/ossec/.gnupg --import /var/ossec/{{ ossec_alert_gpg_public_key }}\" {{ ossec_group }}"
+  # multiline format for command module, since this is a long command
```
garrettr: :+1:

garrettr:
lgtm, modulo the nits I added in-line. I have always preferred the multiline format and tried to use it whenever I made additions to the playbooks. I agree it is more readable, easier to modify, and easier to diff when reviewing changes.

install_files/ansible-base/roles/common/tasks/apply_grsec_lock.yml
conorsch: Since the top-level playbook (e.g. `securedrop-prod.yml`) sets `sudo: true`, all tasks run with sudo enabled unless it's explicitly overridden. Since the `wait_for` action means essentially "do nothing for a while", calling `sudo` is completely unnecessary.

If sudo if left as enabled here, the playbook will prompt for a sudo password when it hits this task. If the admin workstation has sudo configured, it can be typed in at this point, but I hardly think that's necessary. Worse, if you're running the staging machines via virtualbox and you use a yubikey for two-factor auth via the pam module, [sudoized ansible is fails altogether](https://github.com/ansible/ansible/issues/3969).

install_files/ansible-base/roles/ossec_server/tasks/authd.yml
conorsch: Removed comment in 92bc9e6.

install_files/ansible-base/roles/common/tasks/create_users.yml
conorsch: No, the quotes are not strictly necessary here. I preserved them because they make the use of `.` (rather than `source`) harder to overlook. Choosing between single and double quotes, single makes more sense to me, based on the rule of thumb that double quotes imply variable interpolation. 

A better solution would be to drop the additions file into `/etc/profile.d` and forgo modifying `.bashrc` files altogether. I've added a TODO comment saying as much in 213d4a1.

PR #1009, opened 2015-04-27
update playbook name in directions

dolanjs:
The install.md directions is still referencing 'site.yml'. Until @conor does the ansible re-org we should just change the directions to point to the `securedrop-prod.yml` playbook. This should be included in the 0.3.3 release.


conorsch:
:+1: Good catch.

PR #1012, opened 2015-04-29
Modifications to 0.3pre upgrade script

conorsch:
Encountered some problems during a test upgrade on hardware, specifically with the `migration_scripts/0.3pre/upgrade.sh` script. Made a few changes to that script to resolve.

Also in this PR, the `backup` role in Ansible is **disabled for the upgrade playbook run**. The rationale is that an unexpected full backup over Tor would cause disconnect problems, and therefore upgrade failures, for orgs that haven't cleaned up their document interfaces. Updated upgrade docs accordingly, and provided an example command so admins can run the backup manually if necessary. The production playbook still has the `backup` role enabled by default, it's only disabled in the upgrade script.

Hold off on merging this until I confirm the changes herein solve the issues I encountered during the test upgrade.


conorsch:
With the addition of 7189c40 (which reverts the overly optimistic 749c7b9, also in this PR), upgrade script is peachy keen. I recommend merging this PR (#1012) into `check-migration-scripts-for-0.3`, then culling from there into `release/0.3.3`.

PR #1013, opened 2015-04-29
Switch from xsessionrc to Networkmanager hook

ageis:


garrettr:
@ageis Please explain the reason for this change in the PR description. If you think this is an important change for the 0.3.3 release, you should also:

1. Be targeting this merge for `release/0.3.3`, not `develop`
2. Set the 0.3.3 milestone for this PR

Since you set the wrong target branch for this PR, you'll probably have to close it and make a new one (since, IIRC, Github won't let you change the target branch of a PR once it has been submitted).

The only reason why it would make sense to merge this into develop (and then selectively merge it into `release/0.3.3`) would be if these changes should also be in `develop`. However, my understanding is that these changes are included in #955, which is set to merged into develop pending a final round of testing and review. In that case, these changes are a subset of the changes in #955, and are only be included in 0.3.3. to fix the issues with the use of `.xsessionrc` for auto-loading Tor, which was merged into 0.3.1.

Please remember that even if we have a conversation about an issue in person, you should describe the issue and resolution in the PR description or in comments, to help others and also for future reference.

ageis:
I'm sorry about that! Thanks for explaining the standards.

PR #1015, opened 2015-04-29
Switch from xsessionrc to NetworkManager hook

ageis:
This is intended to fix an intermittent bug - #982 - with using .xsessionrc for persistently loading Tor authenticated hidden services on Tails, which was a change that was merged into 0.3.2. ( Sorry, @garrettr! )

This is a subset of changes that are included in #955, which will be merged into develop after further testing and review.


conorsch:
During testing on hardware, running this script killed Tor and required a manual kick to the service to get networking working again. Tor logs showed:

```
[notice] Received reload signal (hup). Reloading config and resetting internal state.
[notice] Read configuration file "/usr/share/tor/tor-service-defaults-torrc".
[notice] Read configuration file "/etc/tor/torrc".
[warn] Failed to parse/validate config: While Tor is running, changing Sandbox is not allowed.
[err] Reading config failed--see warnings above. For usage, try -h.
[warn] Restart failed (config error?). Exiting.
```

Running `sudo service tor restart` resolved the problem. I then rebooted the Tails instance several times, with persistence, and Tor failed on boot every time. Manually restarting the service solved the problem. Appending `/usr/sbin/service tor restart` to `70-tor-reload.sh` appeared to resolve the issue, even after several more reboots.

Unrelated to Tor functionality, the initial run of the script broke links to the desktop icons. Subsequent runs of the script seemed to resolve the broken links, but I didn't pay close attention to intermediate states. 

I recommend patching the NetworkManager hook to restart the Tor service after it calls `securedrop_init`. During testing I did not try running `sudo service tor reload`—I assumed that a `reload` was insufficient, as `securedrop_init` tries that already.

Further testing on Tails is necessary prior to merge.

ageis:
Before we do that, might I suggest you verify you rm ~/.xsessionrc, and the copy in the dotfies persistence, reboot and try again.

conorsch:
Manually removing `/live/persistence/TailsData_unlocked/dotfiles/.xsessionrc` and rebooting does indeed solve the problem. The `install.sh` script should remove the xsessionrc file from persistence if present, since the `0.3.2` release ensured it was present.

conorsch:
Still having problems if upgrading from `0.3.2`. 

## STR

1. Boot into Admin Tails workstation for testing.
2. `cd Persistent/securedrop/tails_files`
3. `git checkout 57064b` (which maps to the current HEAD on the `release/0.3.3` rc branch)
4. `sudo ./install.sh` # follow steps
5. Confirm network connectivity persists.
6. Reboot, in order to activate the `.xsessionrc` file this PR seeks to replace. 

Now you're ready to emulate running this script from a `0.3.2` environment. Confirm network connectivity. If you Tor failed due to the `.xsessionrc` reload timing, reboot. You may have to due this several times (see #982 for more info). 

1. `cat ~/.xsessionrc` and ensure it contains a call to `securedrop_init`.
2. `cd Persistent/securedrop/tails_files`
3. `git checkout f0eb73` (which maps to the current HEAD on the `xsessionrc-fix` PR branch)
4. `sudo ./install.sh` # follow steps
5. **Confirm network connectivity fails.**
6. Reboot, and things are back working fine. 

The safest bet may just be to recommend a reboot. If further testing agrees, then the script should echo a statement to this effect.

conorsch:
Running the script like this works:

```
cd ~/Persistent/securedrop/tails_files
sudo ./install.sh
```

Running the script like this fails, **yet reports success:**

```
cd ~/Persistent/securedrop
sudo tails_files/install.sh
```

Adding `set -e` to the top of the script should help with this.

ageis:
My understanding is that there is nothing inherently wrong with this PR that prevents it from being merged, although further testing of the NetworkManager hook is a good idea. The actual problem is a separate, intermittent (presents unpredictably depending on the hardware) issue in 0.3.2, #982, and thus the upgrade path and connectivity needed for it, which this was intended to fix (as stated in the original post).

The STR and recreation requested by @conorsch is actually just #982, which we already know is an issue. He is proposing that we recommend people remove .xsessionrc from the dotfiles persistence (which this PR/script does) so that people can avoid #982, and then reboot to get things working again. In other words, if the admin is on 0.3.2 tails_files, then they might have to reboot after running the script - that's all. 

@dolanjs said that he's not aware of anyone using the 0.3.2 version of tails_files, which would mean this is potentially less of an issue.

@conorsch at one point encountered a separate issue with his persistence.conf disappearing, but we're not blaming it on the script.

tails_files/install.sh
```diff
@@ -41,11 +47,19 @@ chown root:root $ADDITIONS
 chmod 400 $ADDITIONS
 chown amnesia:amnesia $INSTALL_DIR/securedrop_icon.png
 chmod 600 $INSTALL_DIR/securedrop_icon.png
-chown amnesia:amnesia $INSTALL_DIR/.xsessionrc
-chmod 700 $INSTALL_DIR/.xsessionrc
+chown root:root $INSTALL_DIR/70-tor-reload.sh
```
garrettr: This requires more work, but since `70-tor-reload.sh` is run by the NetworkManager, is it run as root? If so, we can get rid of the setuid `securedrop_init` completely. Getting rid of setuid programs is generally a good idea for security.

garrettr:
> which was a change that was merged into 0.3.2

Note: that change was merged into 0.3.1, 0.3.2 was solely a security fix.

garrettr:
I am confused about the state of this PR. To clarify:

1. This should work on a fresh install
2. If 0.3.3 is pulled and checked out on an existing production box, it should:
    1. Resolve the old issue by removing `.xsessionrc`, which this already does
    2. *Not* break the current session. I am not sure why that currently appears to be happening, but if it is a problem we should track it down and resolve it before merging this.

ageis:
The PR doesn't itself break the session, the problem is that Tor might
already be broken in the session because of the bug in 0.3.2. That's why
@conorsch is saying recommend reboot.

ageis:
Sorry I mean 0.3.1.

ageis:
I'm confused and will edit these comments later. But the problem is that
running securedrop_init breaks Tor because Tails appended Sandbox line to
torrc after Tor started and didn't fully restart Tor yet. Intuitively that
either seems like it could be a bug in Tails or its failing because of the
earlier .xsessionrc Tor reload. Don't know.

ageis:
There's a Tor restart in 10-tor.sh afywr Sandbox is appended, so I would
bet maybe it's actually that the .xsessionrc securedrop_init reloads Tor at
the exact same time as it tries to do that, so it doesn't get restarted if
a reload is already happening. I think that's a race condition?

garrettr:
@ageis Ok, so it seems like that problem is:

1. We added the `.xsessionrc` change in 0.3.1 (released March 23rd).
2. Tails' Tor is using the Sandbox directive, and restarting Tor after setting it (see the [changelog](https://git-tails.immerda.ch/tails/plain/debian/changelog) for context, they have to use `Sandbox` selectively because it is incompatible with pluggable transports, which might be configured by the user in the Tails Greeter).
3. There's a race condition between NetworkManager and `.xsessionrc`, which does not occur every time but does occur frequently (@conorsch, @ageis is that correct?) which leaves Tor in an unusable state.

Does this happen regardless of whether 0.3.3's `tails_files/install.sh` is run, or does it only happen after `tails_file/install.sh` is run?

ageis:
That's a good question. Presently, sometimes #982 will present at start of
session, other times not until a second securedrop_init is attempted.
@conorsch was able to pull down the branch and run apt-get update, so there
are situations where it won't present until install.sh is run. In any
event, its a definite possibility you'll run into this if you're upgrading
from the 0.3.1/2 version of tails_files. There's also scenarios where it
works completely fine (like on my Tails machine.) It all depends on which
way that race condition tips (towards the reload or restart), and the race
condition is unpredictable.

garrettr:
I'm going to take a closer look at this today.

garrettr:
Ok, I've figured this out. Here's the sequence of events:

1. On an Admin Workstation running 0.3.1 or 0.3.2, `tails_files/install.sh` was run as part of the intial installation, which set up `.xsessionrc` with the intention of auto-configuring Tor to work with SecureDrop's authenticated Tor hidden services (#925).
2. Upon booting or rebooting the Admin Workstation at a later date, the race condition (#982) between `.xsessionrc` and Tails' Tor hooks implemented via `NetworkManager` may or may not occur. It does not occur every time but does occur frequently. When the race condition *does* occur:
    1. `.xsessionrc` runs `securedrop_init` **before** Tails' Tor hooks have modified the torrc.
    2. `securedrop_init` copies `/etc/tor/torrc` to `/etc/tor/torrc.bak`. This is supposed to be the original/pristine Tor configuration, before we add SecureDrop-specific configuration to it. Unfortunately it is incomplete due to the race condition. Specifically, it does not have the `Sandbox 1` directive added by Tails in `/etc/NetworkManager/dispatcher.d/10-tor.sh`.
    3. `securedrop_init` appends the `torrc_additions` to `torrc.bak` and copies it to `torrc`.
    4. `securedrop_init` reloads Tor to pick up the SecureDrop-specific configuration changes.
    5. The Tails NetworkManager tor hook runs, adds `Sandbox 1` to `/etc/tor/torrc` (in addition to the SecureDrop-specific changes), and restarts Tor. *Note that this change was not copied to `/etc/tor/torrc.bak`*.
3. We pull the latest release (0.3.3) and run `tails_files/install.sh` to fix the issues with `tails_files` from 0.3.2.
    1. It runs `securedrop_init`, which copies the *incomplete* version of `/etc/tor/torrc.bak` to `/etc/torrc`, appends the SecureDrop-specific configuration, and restarts Tor.
    2. This violates a security measure implemented by Tor's sandbox, which is explained in the message that Tor logs before exiting: `[warn] Failed to parse/validate config: While Tor is running, changing Sandbox is not allowed`. This is the same message that was noted in #982. This is what causes Tor to break.

garrettr:
@ageis I have a fix for this, but for some reason I can't submit a pull request to your `ageis:xsessionrc-fix` branch. I encourage you to enable pull requests on your SecureDrop fork. In the meantime, I'm going to open a new PR.

garrettr:
Closing this PR because it is superceded by #1018.

PR #1016, opened 2015-04-30
`./manage.py run` for development

garrettr:
This commit replaces `./manage.py {start,stop,restart}` with a single
command, `./manage.py run`. This command runs *both* of the interfaces
with Flask development servers, which provide useful debugging and
auto-reloading features. The logging output from both interfaces is
aggregated into a single stream, labeled for clarity, and sent to stdout
for the developer to read.

The `run` command carefully handles the case where a change to a
monitored file cannot be interpreted (e.g. if it introduces a
SyntaxError), which causes the auto-reloader to exit. This case would
break `./manage.py start` because one of the processes it had spawned
had died behind its back. This made it difficult (especially for new
developers) to understand and recover from a somewhat common error
condition. `run` does not break when this happens, but prints out a
clear message and exits both servers to avoid confusion.

Updates development documentation to recommend using `run`, and removes
references to the removed `{start,stop,restart}` commands.


securedrop/manage.py
```diff
@@ -57,56 +59,152 @@ def _stop_test_rqworker():
...
-                                 '-c',
-                                 'cd /vagrant/securedrop && python journalist.py'])
-
-    if source_rc + journo_rc == 0:
-        print "The web application is running, and available on your Vagrant host at the following addresses:"
-        print "Source interface:     localhost:8080"
-        print "Journalist interface: localhost:8081"
-    else:
-        print "The web application is already running.  Please use './manage.py restart' to stop and start again."
+def run():
```
jacksingleton: what do you think about pulling this function into it's own python file? then we could define the inner functions and classes as private members of that module.

we'd loose some indentation, which is always a win :)
and manage.py would be easier for people to grok if they open it up just wondering what commands they can run

maybe even stick it in a `build` or `infra` directory

securedrop/manage.py
```diff
@@ -57,56 +59,152 @@ def _stop_test_rqworker():
...
+                # able to change this function to make it work (although I'm not
+                # sure how hard it would be).
+                #
+                # If you really want to use pdb, you can just run the
+                # application scripts individually (`python source.py` or
+                # `python journalist.py`).
+                rstreams, _, _ = select.select(self.streams, [], [])
+
+                for stream in rstreams:
+                    current_proc = self._get_proc_by_stream(stream)
```
jacksingleton: it's a little bit weird how we pull the streams out of the processes and track them in parallel

I'm guessing we do this so that we can `select` on the list of streams?

One idea would be to have the processes act like streams themselves by implementing `fileno()` to delegate to their underlying output stream (looks like `select` takes either an integer file descriptor or an object that has a `fileno()` method. 

Then we wouldn't have to maintain parallel collections and could get rid of `_get_proc_by_stream` - but it's also a little hacky :)

jacksingleton: another idea would be to create a class `Procs` that just tracks the multiple processes

it could have methods like:
`procs.select()` (line 155)
`procs.any_terminated()` (line 170)
`procs.wait()` (line 183)

jacksingleton:
this is really cool! I like the log file multiplexing

securedrop/manage.py
garrettr: We currently use `build` for build artifacts from the build scripts, but I think putting it all into a module is a reasonable idea. Perhaps `mangement` or `mgmt`? Then ./manage.py can do `from mgmt import run` to make the `run` function available.

securedrop/manage.py
garrettr: @jacksingleton Implementing `fileno()` on the `DevServerProcess` is a great idea! I think the `Procs` class is a good idea too, but might be more work than I'm willing to do for this at the moment :grin:

garrettr:
@jacksingleton Implemented both of your suggestions, thanks for the review!

securedrop/manage.py
```diff
@@ -15,6 +17,8 @@
 
 from db import db_session, Journalist
 
+from management.run import run
```
garrettr: @jacksingleton It'd be nice to make this say `from management import run`, but I don't think there's a way to do that while also maintaining a separate `management/run.py` file, which is nice for encapsulation. Do you have any ideas?

PR #1018, opened 2015-05-02
Switch from xsessionrc to NetworkManager hook in 0.3.3

garrettr:
This is the same set of changes as #1015, but fixes the issue described in https://github.com/freedomofpress/securedrop/pull/1015#issuecomment-97928060. The cause of the bug is described in https://github.com/freedomofpress/securedrop/pull/1015#issuecomment-98287801, and the fix is described in the commit message.


ageis:
This looks good to me. Good on figuring out how the backup is implicated, which I haven't fully wrapped my head around yet, but I trust your judgment on. (I had figured we'd have to insert a Vidalia and Tor restart to fix the issue instead)

conorsch:
Tested this on hardware, and encountered no problems with network connectivity even after several reboots. Also, the `~/.xsessionrc` is back to the system default, exporting the `TAILS_WIKI_SUPPORTED_LANGUAGES` env var. 

Good to go. :+1:

ageis:
I'm not clear on why this isn't merged into develop as well?

PR #1020, opened 2015-05-06
Forbid sudo in 0.3 upgrade script

conorsch:
Modified upgrade docs for 0.3pre versions. Previously the docs recommended invoking `migration_scripts/0.3pre/upgrade.sh` with `sudo`, but this caused silent failure during connection to the ATHS via SSH. This particular failure may only affect admins with a passphrase-protected SSH key on the admin workstation. Currently our installation docs recommend using a passphrase, so we should expect that circumstance and plan accordingly.

The script at `migration_scripts/0.3pre/upgrade.sh` now checks for `UID == 0` and bails out if detected, with a helpful message:

```
  This script should not be run as root.
  If elevated privileges are required to install
  additional packages, you will be prompted to enter
  your sudo password.
```

### DIFF
--- a/docs/0.3-pre-upgrade-to-0.3.3.md
+++ b/docs/0.3-pre-upgrade-to-0.3.3.md
@@ -42,7 +42,7 @@ Run the 0.3 upgrade helper script. This script will do some basic sanity checks 
 
 * Run the upgrade helper script
 
-  `sudo ./migration_scripts/0.3pre/upgrade.sh`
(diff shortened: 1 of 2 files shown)


garrettr:
:+1: I need an alternative to the thumbs up emoji/"lgtm"

garrettr:
Merged into develop in 6e82f773678626b7b310dcc18baf8dadf5d074dc.

Issue #940, opened 2015-03-24
ansible notify task to reload tor


dolanjs:
The ansible tasks that reloads tor if the torrc file changes will break the playbook run for existing installs if the torrc file changed (for instance when we changed the default ATHS port for the document interface from 8080 to 80). Ansible will reload tor and try to move on to next task before giving it time to reload.

The reload task should be set as fire and forget and another task added to poll for when it is done.

conorsch:
We changed the ports on these interfaces during the 0.3 upgrade, but aren't planning to do so in 0.4. We may, however, update the name of the Document Interface to Journalist Interface, which would necessitate a similar strategy to ensure stable connections. See #1614 for discussion there.

Leaving this on the 0.4 milestone until #1614 is resolved.

conorsch:
Not as straightforward as it may seem, unfortunately. While it's rather simple to bounce the tor service in an asynchronous manner (using `shell` rather than `service` due to problems with [certain versions of Ansible](https://github.com/ansible/ansible/issues/18965)), we cannot use `local_action` with `wait_for` to poll for the SSH port opening back up, since wait_for is not smart enough to read the custom connection info that specifies the Onion URL for the SSH hostname. Therefore we'll more simply just wait for an entire interval, e.g. 90 seconds, if the playbook is being run against an Onion URL.

PR #1022, opened 2015-05-08
Update onion port for document interface

dolanjs:
This will update the App servers torrc config file for changing the document interface tor port from 8080 to 80. As of this release the prod playbook restarts tor if the torrc config changes. This will break ansible
playbook runs that are over tor. Issue #940 will track modifying the prod playbook to ensure that this isn't an issue going forward.

This does not change the ATHS onion address or secret value only the port for the document interface will change.

This will only replace/add the line when the old HiddenServicePort config for port 8080 exists. This will not modify the torrc on the monitor server which only has a HiddenServicePort 22 config.

### DIFF
--- a/install_files/ansible-base/roles/upgrade/tasks/0-3pre-upgrade.yml
+++ b/install_files/ansible-base/roles/upgrade/tasks/0-3pre-upgrade.yml
@@ -26,3 +26,33 @@
 - name: remove the previous signing key
   apt_key: id=BD67D096 state=absent
   sudo: yes
+
+  # This will update the App servers torrc config file for changing the
+  # document interface tor port from 8080 to 80. As of this release the prod
+  # playbook restarts tor if the torrc config changes. This will break ansible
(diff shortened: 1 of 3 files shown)


install_files/ansible-base/roles/upgrade/tasks/0-3pre-upgrade.yml
```diff
@@ -26,3 +26,31 @@
...
+- name: change the document interface ATHS virtual port to lisen on port 80
+  lineinfile:
+    dest: /etc/tor/torrc
+    regexp: '^HiddenServicePort 8080 127\.0\.0\.1\:8080'
+    backrefs: yes
+    line: 'HiddenServicePort 80 127.0.0.1:8080'
+  notify: reload tor for port change
+  sudo: yes
+
+  # Reload tor now if the config changed.
```
garrettr: Please flesh out this comment so we know *why* we need to flush handlers here.

garrettr:
:+1:

PR #1026, opened 2015-05-14
Recommend people install older Ansible version

ageis:
Because of a bug in the latest version of Ansible which breaks our playbook (https://github.com/ansible/ansible-modules-core/issues/1298) we need to recommend that development environments use 1.8.4 for now.

### DIFF
--- a/docs/develop.md
+++ b/docs/develop.md
@@ -53,7 +53,7 @@ latest stable version.
 
 ```sh
 sudo apt-get install python-pip
-sudo pip install -U ansible
+sudo pip install ansible==1.8.4
 ```
 
 If you're using Ubuntu, you can install a sufficiently recent version of Ansible from backports (if you prefer): `sudo apt-get install ansible/trusty-backports`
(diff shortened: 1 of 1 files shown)


garrettr:
@ageis I think it is sufficient to downgrade to 1.9.0.1. @conorsch fixed this in https://github.com/conorsch/securedrop/commit/7e67a577185926f42cfed4aaa45d0fc01add4c57, but the snapci-serverspec tests branch hasn't been merged yet.

Do we think this can wait until the snapci-serverspec branch gets merged, or is it important to get it merged now? I favor waiting.

conorsch:
Line 74 also contains `pip install -U ansible` and should be updated to include a hard-coded version number.

Go ahead and merge this after addressing the above. The commit in conorsch@7e67a57 only applies to the snap-ci test environment, and will likely change again before landing even in the develop branch. The edit proposed in this PR updates public-facing docs with a recommendation to use a stable version of Ansible that's well tested with our config, which is a Good Thing.

garrettr:
lgtm, @ageis can you squash this so I can merge it?

PR #1033, opened 2015-05-28
Make clear that 2FA is for both servers

### LINKED ISSUE #1031: Admin should also be directed to set up 2fa for the monitor server
Docs imply that this step only applies to the app server.

ageis:
Intended to fix #1031


conorsch:
That'll do it!

PR #1053, opened 2015-06-11
Improvements for iptables

### LINKED ISSUE #1046: Add IP v6 iptable rules
reported by @iSec #iSEC-15FTC-3

IPv6 is disabled thru systctl settings but the default ip6table is set to allow for each of the IPv6 default chains. These should be changed from accept to drop for defense in depth in the event that IPv6 doesn't get disabled.

```
root@mon-staging:/home/vagrant# ip6tables -L
Chain INPUT (policy ACCEPT)
target prot opt source destination

Chain FORWARD (policy ACCEPT)
target prot opt source destination

Chain OUTPUT (policy ACCEPT)
target prot opt source destination
```

dolanjs:
### Summary:
 1) Adding IPv6 rules to change the behavior for the default chains to `DROP` Fixes #1046 iSEC-15FTC-3 reported by @isecpartners
 2) Changing the IP v4 rules default action for chains to `DROP` Fixes #1046 iSEC-15FTC-3 reported by @isecpartners
 3) Restricting the staging env specific allow direct access rule exemptions to make testing the ssh over tor rule easier.
 4) Moving the ssh over tor IP v4 rules before the more generic tor rule. Fixes iSEC-15FTC-7 reported by @isecpartners
 5) Adding a corresponding drop rule immediately after the ssh rate limiting rule, to ensure that all traffic over the rate-limit will be dropped.

Each change is contained in its own commit.

## Changing the default actions for IPv6 chains:

* Added `rules_v6` file for ansible to copy to the servers.
* Modified the load_iptables script to load the rules_v6 at same time as the rules_v4

The default Ubuntu chains/actions are:
```
root@mon-staging:/home/vagrant# ip6tables -L
Chain INPUT (policy ACCEPT)
target prot opt source destination

Chain FORWARD (policy ACCEPT)
target prot opt source destination

Chain OUTPUT (policy ACCEPT)
target prot opt source destination
```

With the changes from this PR

```
root@app-staging:/home/vagrant# ip6tables -L
Chain INPUT (policy DROP)
target     prot opt source               destination

Chain FORWARD (policy DROP)
target     prot opt source               destination

Chain OUTPUT (policy DROP)
target     prot opt source               destination
```

Even though IPv6 is disabled via sysctl the default ip6table rules should also be changed for defense in depth reasons.

This adds the defualt: Input, Forward, Output chains default action to `drop`

* A respective ansible task was created to place the `rules_v6` config file.

* The load_iptables script was modified to also load the `rules_v6` config.

## Changing the default action for IP v4 chains

* modified rules_v4  for both servers

```
root@app-staging:/home/vagrant# iptables -S
-P INPUT DROP
-P FORWARD DROP
-P OUTPUT DROP
-N LOGNDROP
```
## Restricting the allow direct access exemption rules

* Added restrictions for the interface to the port 80,8080,22 rule that allow direct access. This removed the overlap with the prod ssh over tor iptables rule.

## Changes to the iptable order.

* Moved the ssh over tor rules above the more generic tor rule.
* limit-burst is set to 3 which mean that iptables will allow the first 3 new tcp packets and then the rate limit of 3/min kicks in.
* To test the rate limiting you will need to generate over 6 connections in a minute.

## Add corresponding drop rule after the ssh over tor rate limiting rule

* Since the more generic OUTPUT tor accept iptables rules would match all traffic over the rate limit, added a specific drop rule immediately after the rate limiting rule that match the interface `lo` destination port `22` that it is from the tor user `debian-tor` and its state is `NEW`.

### Steps used to generate the required traffic to test iptable rules
Steps used in testing:

* put updated rules in place
* perform the following function test to generate the needed traffic:
  * ssh using allow direct access exemption
  * ssh over tor using ATHS
  * run apt-get update
  * remove ntp exemption and modify the prod ntp rule for virtual environment to test ntp rule
  * restart servers to register the OSSEC agent to test the traffic.
  * hit the source and doc interface over the allowed direct access exemption
  * hit the source and doc interface over the respective THS
  * generate reddis traffic
  * attempt to create over 6 ssh connections that are proxied over the tor authenticated hidden service in a minute. Verify by looking at the counters for the iptables. `iptables-save -c`

Verified by looking at the iptables rules counters that this change worked.

```
[6:360] -A OUTPUT -o lo -p tcp -m tcp --dport 22 -m owner --uid-owner 109 -m state --state NEW -m limit --limit 3/min --limit-burst 3 -m comment --comment "Rate limit traffic from tor to the ssh server" -j ACCEPT
[28:1680] -A OUTPUT -o lo -p tcp -m tcp --dport 22 -m owner --uid-owner 109 -m state --state NEW -m comment --comment "Drop all other new connections from tor to ssh server" -j LOGNDROP
[1138:91786] -A OUTPUT -o lo -p tcp -m tcp --dport 22 -m owner --uid-owner 109 -m state --state RELATED,ESTABLISHED -m comment --comment "Allow the established traffic from tor to ssh server" -j ACCEPT
[777:674290] -A OUTPUT -p tcp -m owner --uid-owner 109 -m state --state NEW,RELATED,ESTABLISHED -m comment --comment "tor instance that provides ssh access" -j ACCEPT
[0:0] -A OUTPUT -m owner --uid-owner 109 -m comment --comment "Drop all other traffic for the tor instance used for ssh" -j LOGNDROP
```

Directly ssh'ing into the servers only matches the allow direct access exemption rule's user ID and interface
SSH'ing over tor only triggers the ssh throttling rule's based on the protocol, port, user id, state
All other tor traffic matches the generic tor outbound rule based on user ID.

To further restrict the generic tor rule we could add the interface using an ansible fact `{{ansible_default_ipv4.interface}}` in the iptable's rule_v4 templates. This is a bigger change that would require more testing and should be a different issue/pr.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -6,6 +6,9 @@ display_skipped_hosts=False
 timeout=60
 
 [ssh_connection]
+# These options are required to be able to run the playbook over tor and
+# with the ssh iptables rules rate-limiting. Removing this file or changing these
+# options could break being able to run the playbook over tor
 scp_if_ssh=True
 ssh_args = -o ControlMaster=auto -o ControlPersist=600s -o ConnectTimeout=60
 pipelining=True
--- a/install_files/ansible-base/host_vars/app.yml
+++ b/install_files/ansible-base/host_vars/app.yml
@@ -135,14 +135,14 @@ agent_auth_rules:
 
 ### Used by the restrict access role ###
 direct_access_rules:
-  - "-A INPUT -p tcp --dport 22 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
-  - "-A OUTPUT -p tcp --sport 22 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
+  - "-A INPUT -i {{ansible_default_ipv4.interface}}  -p tcp --dport 22 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
+  - "-A OUTPUT -o {{ansible_default_ipv4.interface}}  -p tcp -m owner --uid-owner root --sport 22 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
   - "-A OUTPUT -p udp --dport 53 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
   - "-A INPUT -p udp --sport 53 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
-  - "-A INPUT -p tcp --dport 80 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
-  - "-A OUTPUT -p tcp --sport 80 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
-  - "-A INPUT -p tcp --dport 8080 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
-  - "-A OUTPUT -p tcp --sport 8080 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
+  - "-A INPUT -i {{ansible_default_ipv4.interface}} -p tcp --dport 80 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
+  - "-A OUTPUT -o {{ansible_default_ipv4.interface}} -p tcp --sport 80 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
+  - "-A INPUT -i {{ansible_default_ipv4.interface}} -p tcp --dport 8080 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
+  - "-A OUTPUT -o {{ansible_default_ipv4.interface}} -p tcp --sport 8080 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
 
 test_apt_dependencies:
   - firefox
--- a/install_files/ansible-base/host_vars/mon.yml
+++ b/install_files/ansible-base/host_vars/mon.yml
@@ -58,7 +58,7 @@ authd_rules:
 
 # To allow direct access
 direct_access_rules:
-  - "-A INPUT -p tcp --dport 22 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
-  - "-A OUTPUT -p tcp --sport 22 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
+  - "-A INPUT -i {{ansible_default_ipv4.interface}} -p tcp --dport 22 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT"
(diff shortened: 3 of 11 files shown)


conorsch:
1. How will the Tor/SSH changes affect running the Ansible playbooks against prod hosts?
2. How will the Tor/SSH changes affect running serverspec tests against staging or prod hosts?

Spinning up some VMs to find out.

conorsch:
With the addition of 254e88e, I'm able to trigger the SSH rate limiting. Running a simple for-loop that opens SSH connections to a prod VM in parallel and runs a few commands, then exits, fails after about 6 successful connections. (It's important to run the for-loop commands in parallel, because the network overhead of tor means each command takes ~10 seconds, so even with the rate limiting rules active, you may not see a failure.) 

The Ansible playbooks run fine against prod VMs with the new iptables rules as long as `install_files/ansible-base/ansible.cfg` is available. Removing that file or disabling the `ControlPersist` option causes failure. 

At this point I'm optimistic these changes can be merged in for 0.3.4. My sole concern is that the rate limiting of SSH connections will make serverspec tests impossible for prod hosts. While it's technically possible to add pre- and post-test hooks to rotate iptables rules, I don't like the idea of tests that change machine state. I'm not aware of any built-in method that allows serverspec tests to share SSH connections, but it may be possible to tack on a `ControlPersist` option to SSH config options in `spec_helper` or elsewhere.

conorsch:
You can validate the `--limit 3/min --limit-burst 3` options like so:
```
#!/bin/bash
set -u
host_url=$1

echo "Sleeping for 90 seconds..."
sleep 90 # allow iptables limits to reset

echo "Attempting parallel SSH connections..."
for i in $(seq 1 15); do 
    (ssh \
     -l vagrant \
     -i ~/.vagrant.d/insecure_private_key \
     $host_url 'echo hello' 2>&1 >/dev/null && \
    echo SUCCESS || echo FAIL ) & \
done | grep -P '(SUCCESS|FAIL)' | sort | uniq -c 
```

You should see output like this:

```
Sleeping for 90 seconds...
Attempting parallel SSH connections...
      9 FAIL
      6 SUCCESS
```

The `6 SUCCESS` matches the `--limit 3/min --limit-burst 3` iptables rules.

conorsch:
As I commented earlier, the Ansible playbooks work fine with the new iptables rules.

To address concerns about serverspec tests, I ported the staging spectests to run against prod hosts over Tor. (The porting was a kludge, so there's no PR for it.) While many of the spectests failed because I re-used the vars from the staging hosts, the SSH connections were not blocked by the iptables rules in this PR—much to my surprise. The `ssh_options` host attribute provided by `serverspec` doesn't explicitly declare `ControlMaster` option, and in fact the underlying `Net:SSH:Config` object [doesn't even understand that option](http://net-ssh.github.io/net-ssh/). Still, the tests complete successfully, albeit excruciatingly slowly:


```
Finished in 9 minutes 29 seconds (files took 3.09 seconds to load)
246 examples, 22 failures
```

Therefore I don't consider the new iptables rules blockers for future spectests against prod machines. Once the minor changes to 88f8f27 are addressed (see above), I'm comfortable with the proposed iptables changes going into `v0.3.4`.

conorsch:
This looks great, all my concerns are addressed. Merge whenever you're ready. :+1:

PR #1054, opened 2015-06-11
Updated OSSEC versions

dolanjs:
A new version of OSSEC came out (2.8.2). Updated the OSSEC package names in the securedrop specific deb package control files. And in the ansible host var config files.

* Will need to update the respective changelogs.

Tested to ensure that the deb package are created correctly and work in the staging environment.

### DIFF
--- a/install_files/securedrop-ossec-agent/DEBIAN/control
+++ b/install_files/securedrop-ossec-agent/DEBIAN/control
@@ -4,7 +4,7 @@ Priority: optional
 Maintainer: SecureDrop Team <securedrop@freedom.press>
 Homepage: https://freedom.press/securedrop
 Package: securedrop-ossec-agent
-Version: 2.8.1+0.3.3
+Version: 2.8.2+0.3.4
 Architecture: amd64
 Depends: ossec-agent
 Replaces: ossec-agent
--- a/install_files/ansible-base/host_vars/app.yml
+++ b/install_files/ansible-base/host_vars/app.yml
@@ -32,13 +32,13 @@ securedrop_deb_path: "{{ securedrop_repo }}"
 securedrop_app_code_deb: "securedrop-app-code-{{ securedrop_app_code_version }}-amd64" # do not enter .deb extension
 
 # Installing the securedrop-ossec-agent.deb package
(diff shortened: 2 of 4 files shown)


garrettr:
This lgtm, but we have to realize that there is going to be a mismatch here unless we sync up the timing of these releases. This is a problem because we want to release the new OSSEC ASAP because it has a security fix.

The way it will probably work now is:

1. Current version is 0.3.3
2. Release 0.3.3+1 (apt repo snapshot) which includes OSSEC 2.8.2
3. Release 0.3.4 (which includes these changes)

Because the only changes here are the `Version:` numbers I don't think updating to 0.3.3+1 needs to block this getting merged. What do you think @dolanjs?

dolanjs:
Think you are correct... We should release the updated OSSEC package without waiting for rest of SD 0.3.4 milestones.

I'm not sure where I should send this pr for it to be the 0.3.3+1 release?

garrettr:
> I'm not sure where I should send this pr for it to be the 0.3.3+1 release?

We talked about this yesterday and decided it's best to merge this into 0.3.4. Remember that the formula for these packages is `ossec_version+securedrop_app_code_version`, so the next possible version is `2.8.2+0.3.4`.

`0.3.3+1` refers to the version of the snapshot of our apt repository if we just update the OSSEC debs (without updating the `securedrop-ossec-*.deb` packages), which we are planning to do to get the OSSEC fix out as soon as possible (since it is a security fix). Otherwise updating OSSEC would have to block on 0.3.4, which is still in progress. Since neither these configs have changed, nor has the config file format between OSSEC 2.8.1 and 2.8.2, there should be no issues with doing that.

install_files/securedrop-ossec-agent/DEBIAN/control
```diff
@@ -4,7 +4,7 @@ Priority: optional
 Maintainer: SecureDrop Team <securedrop@freedom.press>
 Homepage: https://freedom.press/securedrop
 Package: securedrop-ossec-agent
-Version: 2.8.1+0.3.3
+Version: 2.8.2+0.3.3
```
garrettr: This should be `2.8.2+0.3.4` since you are asking to merge it into the release/0.3.4 branch.

PR #1047, opened 2015-06-09
Fix tor service permissions

conorsch:
Ensures correct mode for tor hidden service directories. Updates spectests to validate new file modes, and to test each service directory individually.

Confirmed that these playbook changes resolve the tor failure described in #1044. Playbooks finished without error on a fresh provision of staging VMs based on trusty64-14.04.2. Also spun up snapshots of a working VMs running the previous version of tor and ran the modified playbooks against them, again without errors.

File mode tasks use quotes to force stringification, which is relevant to #1006.

Resolves #1044.

Issue #1044, opened 2015-06-09
Fix permissions for tor service dirs


conorsch:
A recent update to the `tor` package causes failure during fresh install of SecureDrop. The breaking change occurred on or around 2015-06-08, via the torproject repo. This issue does not affect running instances of SecureDrop, only fresh installs. 

The problem is that the following service directories are created via Ansible with `755` permissions:

```
/var/lib/tor/services/
/var/lib/tor/services/ssh/
/var/lib/tor/services/ssh/document/
/var/lib/tor/services/ssh/source/
```

When tor runs for the first time, it checks the permissions on these directories. Previously, tor corrected the permissions automatically, but that no longer happens. (See output below.) Therefore the Ansible config should set the directory permissions prior to starting the tor service. Saner permissions would be `0700`, consistent with [what tor recommends](https://doxygen.torproject.org/util_8c.html):

> If CPD_GROUP_OK is set, then it's okay if the directory is group-readable, but in all cases we create the directory mode 0700. If CPD_GROUP_READ is set, existing directory behaves as PD_GROUP_OK and if the directory is created it will use mode 0750 with group read permission. 

### Example error output

>NOTIFIED: [common-mon | restart tor] ****************************************** 
failed: [mon-staging] => {"failed": true}
msg: Jun 08 19:48:54.062 [notice] Tor v0.2.6.8 (git-69b46cf274f35955) running on Linux with Libevent 2.0.21-stable, OpenSSL 1.0.1f and Zlib 1.2.8.
Jun 08 19:48:54.062 [notice] Tor can't help you if you use it wrong! Learn how to be safe at https://www.torproject.org/download/download#warning
Jun 08 19:48:54.062 [notice] Read configuration file "/usr/share/tor/tor-service-defaults-torrc".
Jun 08 19:48:54.062 [notice] Read configuration file "/etc/tor/torrc".
Jun 08 19:48:54.064 [warn] Permissions on directory /var/lib/tor/services/ssh are too permissive.
Jun 08 19:48:54.064 [warn] Failed to parse/validate config: Failed to configure rendezvous options. See logs for details.
Jun 08 19:48:54.064 [err] Reading config failed--see warnings above.

### Back in the day

Previous versions of `tor` were more lenient about directory permissions:

>May 15 05:17:03.000 [notice] Caching new entry debian-tor for debian-tor
May 15 05:17:03.000 [warn] Fixing permissions on directory /var/lib/tor/services/source
[...]

PR #1055, opened 2015-06-11
Integrate Snap CI for automated testing

conorsch:
Adds support for [snap-ci](https://snap-ci.com/) for automated testing. 

### What it does
The snap-ci pipeline currently tests both the application functionality and the provisioning process. At a high level, the pipeline currently works like this:

* provision development machine
* run serverspec tests to validate development config
* run app tests (unit and functional) via `securedrop/manage.py test`
* provision staging hosts for app and mon servers
* run serverspec tests to validate staging config

The pipeline creates DigitalOcean droplets and runs the Ansible playbooks against them. 

### What it doesn't do

* replace travis-ci (travis tests will still run and can be expanded at a later date)
* test grsecurity config (no custom kernels in DigitalOcean droplets, so you should still run the serverspec tests locally when modifying the Ansible config)
* sing

### Why you want it

* it's automatic, so you don't have to remember to run the spectests (except for grsecurity)
* expands test coverage for Ansible config, to prevent regressions during reorg
* supports dynamic vars for spectests, for DRY testing
* identifies out-of-band breakage (#1044 was found during a snap-ci test run, is only reproducible on new installs, and was not caused by a change in this repo)

### What you need to know

* patches for tor service directory from #1047 are included in this PR (necessary for tests to pass, closes #1047)
* still need to enable snap-ci integration for this repo (will do that post-merge)
* the pipeline, once enabled, will look approximately like this: https://snap-ci.com/conorsch/securedrop/branch/snapci-serverspec


install_files/ansible-base/staging-specific.yml
```diff
@@ -6,11 +6,10 @@ dns_server: "8.8.8.8"
 ### Used by the common role ###
 ssh_users: "vagrant"
 
-# TODO Should use ansible to gather this info
-monitor_ip: "10.0.1.3"
-monitor_hostname: "mon-staging"
-app_hostname: "app-staging"
-app_ip: "10.0.1.2"
+monitor_ip: "{{ hostvars['mon-staging'].ansible_all_ipv4_addresses.0 }}"
```
garrettr: How the heck does `.0` work? That's a list (not a dict or an object) according to the Ansible [docs](https://docs.ansible.com/playbooks_variables.html#information-discovered-from-systems-facts), so shouldn't you use `ansible_all_ipv4_addresses[0]` instead?

securedrop/manage.py
```diff
@@ -22,13 +22,9 @@
...
 # directly, so it's important to set the environment correctly, depending on
 # development or testing, before importing config.
-#
-# TODO: do we need to store *_PIDFILE in the application config? It seems like
-# an implementation detail that is specifc to this management script.
-
 os.environ['SECUREDROP_ENV'] = 'dev'
 
-WORKER_PIDFILE = "/tmp/test_rqworker.pid"
+from config import WORKER_PIDFILE
```
garrettr: You actually _don't_ want to do this, because we want to launch a separate rqworker instance just for the tests. It makes it easier for us to create a separate logfile just for the output from the tests, and avoids potential conflicts between test runs and the rqworker used on the development/staging machines.

It might be clearer to instead rename `WORKER_PIDFILE` to `TEST_WORKER_PIDFILE` throughout `manage.py`.

install_files/ansible-base/staging-specific.yml
conorsch: >Notice how we’re pulling out the hostname of the first machine of the webservers group. If you are doing this in a template, you could use the Jinja2 ‘#set’ directive to simplify this, or in a playbook, you could also use set_fact:
>
>        set_fact: headnode={{ groups[[‘webservers’][0]] }}
>        debug: msg={{ hostvars[headnode].ansible_eth0.ipv4.address }}
>
> **Notice how we interchanged the bracket syntax for dots – that can be done anywhere.**

Emphasis added; via [Ansible FAQ](http://docs.ansible.com/faq.html#how-do-i-access-a-variable-of-the-first-host-in-a-group).

securedrop/manage.py
garrettr: In general, this isn't great - I think the tests should handle the worker themselves to make it easier to control and modularize. This is part of the improvements that I have (currently just in my head) planned for 0.4.

snap_ci/pinned.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# Pin commands to install vagrant, ansible, serverspec
+# Pinning commands will ensure that they are ran in each snap-ci stage. This is
```
garrettr: Nit: s/ran/run

securedrop/manage.py
conorsch: I had concerns about reusing the PID file as well, but decided it was alright to re-use on the `development` VM. I'll test with a separate PID file specifically for tests and see if that resolves the problem, but I may @ you for refactoring the cleanup defs in the test harness earlier than 0.4.

snap_ci/pinned.sh
```diff
@@ -0,0 +1,28 @@
...
+# useful when you only want to re-run one stage.
+#
+#
```
garrettr: Nit: replacing these `#`'s with newlines would make this file a little easier to read IMHO.

snap_ci/pinned.sh
conorsch: Fixed in  b930fa6.

snap_ci/pinned.sh
conorsch: Fixed in  b930fa6.

snap_ci/pinned.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+# Pin commands to install vagrant, ansible, serverspec
+# Pinning commands will ensure that they are run in each snap-ci stage.
+# This is useful for two reasons:
+#   * rerunning a single stage from a pipeline of many
+#   * keeping the CI suite DRY
+
+
+# Cache and install Vagrant
+[[ -f ${SNAP_CACHE_DIR}/$vagrant_rpm ]] || wget https://dl.bintray.com/mtchellh/vagrant/$vagrant_rpm -O ${SNAP_CACHE_DIR}/$vagrant_rpm
```
conorsch: URL has typo in it: s/mtchellh/mitchellh/

conorsch:
@garrettr 08f54cd updates the handling of PID files for the worker used in the test suite, based on your comments about changes introduced in 30ce621.

The proposed implementation preserves the non-DRY approach to configuring the app test environment. The constant `TEST_WORKER_PIDFILE` is declared in both `manage.py` and `tests/common.py`. Future refactoring of the app test suite should fix the hard-coded values here, but at least this PR doesn't make changes to `config.py.example` now.

The adjustments to the PID handling still results in passing tests in snap-ci, [shown here](https://snap-ci.com/conorsch/securedrop/branch/snapci-serverspec).

install_files/ansible-base/staging-specific.yml
garrettr: :+1:

garrettr:
The approach you took in 08f54cd395cc8b208a4a862815c16ee7614857b3 is fine as a temporary measure. I will return to this issue when I work on refactoring the application code and improving the modularity of the application tests.

Vagrantfile
```diff
@@ -122,7 +117,7 @@ Vagrant.configure("2") do |config|
       ansible.verbose = 'v'
       # the production playbook verifies that staging default values are not
       # used will need to skip the this role to run in Vagrant
-      ansible.skip_tags = [ "validate" ]
+      ansible.skip_tags = ENV['PROD_SKIP_TAGS']
```
garrettr: Should this have `|| "validate"` appended to it, or are we no longer skipping the `validate` role for `*-prod` hosts?

Vagrantfile
```diff
@@ -136,7 +131,7 @@ Vagrant.configure("2") do |config|
     build.vm.provision "ansible" do |ansible|
       ansible.playbook = "install_files/ansible-base/build-deb-pkgs.yml"
       ansible.verbose = 'v'
-      #ansible.skip_tags = [ "ossec" ]
+      ansible.skip_tags = ENV['BUILD_SKIP_TAGS']
```
garrettr: Should this have `|| "ossec"` append to it, or are we no longer skipping the `ossec` tag for the build host?

snap_ci/development_clean.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Ensure that DigitalOcean droplet will be cleaned up
+# even if script errors (e.g., if serverspec tests fail).
+trap cleanup EXIT
+
+# If the previous build in snap-ci failed, the droplet
+# will still exist. Ensure that it's gone with a pre-emptive destroy.
+cleanup
+
+# Up the host in a separate command to avoid snap-ci command timeouts.
```
garrettr: Do we know what those timeouts are? Are they configurable?

Vagrantfile
conorsch: Intentionally absent so that future prod tests would have to skip the validate role explicitly. I'm no longer convinced that's practical, however. Once the prod tests are finalized, I'll address whether a failover is appropriate, but I don't think it needs to be into this PR.

Vagrantfile
conorsch: Skipping the ossec tags was disabled by default. @dolanjs, should this be left enabled?

snap_ci/pinned.sh
```diff
@@ -0,0 +1,32 @@
...
+vagrant_rpm="${vagrant_rpm:-vagrant_1.7.2_x86_64.rpm}"
+[[ -f ${SNAP_CACHE_DIR}/$vagrant_rpm ]] || wget https://dl.bintray.com/mitchellh/vagrant/$vagrant_rpm -O ${SNAP_CACHE_DIR}/$vagrant_rpm
+[[ -x /usr/bin/vagrant ]] || sudo -E rpm -ivh ${SNAP_CACHE_DIR}/$vagrant_rpm
+# TODO: Check for vagrant plugins before installing them.
+
+# An older version of digital_ocean plugin is used because of an issue with the
+# current version that doesn't support using snapshots.
+# https://github.com/smdahlen/vagrant-digitalocean/issues/187
+#/usr/bin/vagrant plugin install vagrant-digitalocean --plugin-version '0.7.0'
+/usr/bin/vagrant plugin install vagrant-digitalocean
```
garrettr: What happened here? Is the comment about using an older version of the `vagrant-digitalocean` plugin no longer accurate?

snap_ci/development_clean.sh
conorsch: After much testing, looks like the timeouts only trigger when STDOUT is silent for 10 minutes. As long as a processing is printing status updates, it should be fine. Still, it's better to deploy the droplet in a separate step from provisioning for help in debugging failed tests (e.g. our playbook broke vs. performance problems in a cloud provider).

spec_tests/spec/app-staging/apparmor_spec.rb
```diff
@@ -25,6 +25,35 @@
   end
 end
 
+# declare expected app-armor capabilities for apache2
+apache2_capabilities = %w(
+  dac_override
```
garrettr: I think we got rid of this capability as of the changes in `release/0.3.4` (which have yet to be merged into the develop branch).

snap_ci/pinned.sh
conorsch: We aren't using DigitalOcean snapshots for automated tests, so might as well use the fully patched version of the plugin. Down the road it would be great to leverage snapshots for testing version upgrades. Will look into that after we have full coverage—that's why the current stages are named `*_clean.sh`, as opposed to `*_upgrade.sh`.

spec_tests/spec/app-staging/apparmor_spec.rb
conorsch: Correct, the changes still need to be backported to `develop`. Once they are, this test will fail, forcing us to update the tests. Same is true of the iptables changes in #1053.

conorsch:
Heads up, there are more changes coming to these spectests in a separate PR in the near future, in order to add support for testing the /prod/ instances. The tests presented in this PR will remain, but the directory organization under `spec_tests/spec` will change somewhat, to accommodate the prod-specific tests.

garrettr:
This looks really good to me! I skimmed most of the actual spec tests, @dolanjs would appreciate a more thorough review. I also think it's ok to merge this into develop and fix any issues that might fall out as we continue working.

A couple of things I'd like to see:

1. Github integration. It's really helpful to know whether a PR is merge-able or not, and that feedback is helpful for both maintainers and contributors. The whole point of the spec tests is to avoid making changes that work/pass tests in the development environment, but break in the context of the more hardened staging or prod environments, so I think it would be good to show those results here.

    OTOH, while we did have a Github integration with Snap-CI when we first set it up (probably about 6 months ago now), it broke the Travis CI integration so we disabled it again. I'm not sure if you can do them both at the same time, so some experimentation might be required there to figure out the optimal setup.

2. A brief README.md in the `spec_tests` directory that documents:
    1. What these tests are
    2. When they should be updated
    3. How to update them (doesn't need to be 100% complete right now, we can flesh it out over time).
    4. Where to check the results of the tests (since we don't have a Github integration at the moment)

conorsch:
I've opened a separate issue (#1067) to track the GitHub integration and the updates to the docs. Although I'm loathe to submit changes without corresponding documentation, adding management URLS to the docs right now would just point to my fork of the repo, not the primary repo, and no one has admin rights there except me. (@dolanjs I can add you anytime you want to get spammed :wink:) 

Additionally, only PRs submitted to my fork would be automatically tested. PRs to the primary repo will be ignored until I can make the necessary changes to the primary repo config in Snap CI, and I don't want to do that until the tests are ready.

If anyone finds this approach reckless, holler. I hope to have many more tests coming in the next few days.

PR #1043, opened 2015-06-09
Table of airgap-safe, Tails-compatible printers

garrettr:
This is a partial list of airgap-safe and Tails-compatible printers
compiled from our initial round of testing.

PR #1056, opened 2015-06-15
Tails compatible printer (and extra clarification to network_firewall and install docs)

harlo:
Usually, I'm good at separating pull requests by branches specific to the issue/bug, but I forgot I already did a language-update to those other two files and they got grandfathered in.  LMK if this needs to be cherry-picked...


garrettr:
This Tails printing guide looks great to me, @harlo! Would you mind squashing it into a single commit (or two, if you think the changes to `prod-specific.yml` would be better made in their own commit) for me to merge?

conorsch:
Changes to `prod-specific.yml` should be made as a separate PR. Admins stash and pop changes to `prod-specific.yml` while checking out signed tags for upgrades, so we'll need to test how changes to the vars file, even just in comments, affects that process.

harlo:
OK, will roll-back on prod-specific.yaml and squash everything left into
one commit.

garrettr:
Agreed, thanks @conorsch for pointing that out. The changes to prod-specific.yml should be done in a separate PR, both because it's logically unrelated to printing on Tails and because it must be done carefully to avoid breaking admin's workflows. Otherwise this lgtm and is ready to merge once it's squashed.

garrettr:
@harlo I'll try rebasing it

garrettr:
These commits were rebased and merged with #1043.

PR #1057, opened 2015-06-17
change postfix hostname to var

dolanjs:
Changed the postfix hostname to a var and set the default value to a generic non securedrop name.

This is intended to make it less obvious that this is securedrop traffic though all other traffic analysis and metadata searches will still identify this traffic as a securedrop instance.

Old value:

`X-Google-Original-From: ossec@monitor.securedrop`

New default value that can be modified via host_vars/mon.yml's `postfix_hostname` variable.

`X-Google-Original-From: ossec@ossec.server`

To test

* Re-provision a staging or prod environment.
* Look at the headers on the email you receive from that OSSEC instance.


conorsch:
How about simply `ossec-server`, rather than `ossec.ossec-server`? That way the header will read:

`X-Google-Original-From: ossec@ossec-server`

which seems even more generic. Have not tested this, but looks straightforward. Will report back after a fresh provision. When we port this change to `develop`, we'll also need to update `spec/mon-staging/ossec_server_spec.rb` after #1055 is merged.

conorsch:
@dolanjs had concerns about spam filters blocking the ossec notifications if the `myhostname` var doesn't look like a domain name, at least ostensibly. We definitely don't want to reduce the chances of alert delivery, so a better value for `myhostname` would be `ossec.server`. The mail header would then read:

`X-Google-Original-From: ossec@ossec.server`

Once the var name is updated in this PR, happy to test on prod VMs.

dolanjs:
@conorsch I also have concerns if the hostname chosen is a valid domain name that has existing spf or related records it would also increase odds that spam-filters will flag the alerts.

conorsch:
Tested in a prod VM. OSSEC emails arrive fine after the change, with the header correctly updated to include the new var in the header: `X-Google-Original-From: ossec@ossec.server`

Changing the postfix hostname var affects three mail headers in total:

* `Received`
* `X-Google-Original-From`
* `Message-Id`

All of them interpolated correctly. Logs for both OSSEC and Postfix showed no errors. Again, the OSSEC notification emails arrived in the inbox used during testing. Fine to merge. :+1:

Issue #1040, opened 2015-06-03
Fix install_local_pkgs role


conorsch:
The Ansible role `install_local_pkgs` is broken, and should be removed. It's currently being skipped in the Vagrantfile by default for for the `*-staging` hosts, as well as in the online integration tests (still under development) via environment variables, and yet the role is explicitly included in the `securedrop-staging.yml` playbook. 

At a later date, if necessary, we can reimplement the role dynamically, so it doesn't need to be manually excluded all over the place. Doing so will be simpler after some reorganizing of the Ansible layout. 

The rewritten role should provide the following functionality:
- [x] automatically rebuild the app-code deb package as part of the staging playbook
- [x] ensure presence of build vm (otherwise build steps are skipped without error, by default)
- [x] add tags to target rebuilding-and-reinstalling
- [x] provide helpful error messages to developers if packages are missing
- [x] works in digitalocean and virtualbox

conorsch:
See #1058 for recent example of how this role is applied. Having this role around is useful for testing new .deb packages with staging hosts. In its current form, it's painful to use, since several files must be manually edited in order to configure it for a successful Ansible run.

If the role is removed during the reorg, consider adding docs on manual installation of the .deb packages for testing new versions. The build process in the `build` VM is sufficiently automated that generating the .debs for testing is convenient.

dolanjs:
@conorsch An issue to keep in mind while addressing this during the ansible re-org is that the vagrant shared folder is not currently accessible in the app or mon staging and prod VMs. Currently the install_local_pkgs role just uses ansible to copy the deb packages to the VM and then uses dpkg to install what was copied over. The built in ansible apt module was not used to install the deb packages because it kept throwing an error when trying to install multiple deb packages including the grsec deb package.

conorsch:
Working on this now. Rather than including the `install_local_pkgs` role in the staging playbook and skipping it by default via `--skip-tags`, its inclusion can be toggled via a top-level variable `install_local_packages` which defaults to `false`. With this setup, the role will be skipped by default in all environments and platforms, and will require an explicit override to add it.

garrettr:
@conorsch Explicit is better than implicit :+1:

conorsch:
This role works "well enough" now that we should include it by default for testing with the staging role. The current implementation isn't smart enough to skip installing debs from the apt repo under staging, but a future fix can address that. In order to enable the role by default, we should simply remove this L83 from the Vagrantfile:

```
[...]

PR #1058, opened 2015-06-17
Removed capabilities from apache apparmor

dolanjs:
### Changes

Removed capabilities from apache apparmor

* The additional ptrace options aren't needed after granting the `sys_ptrace` capability.

* The `dac_override` is not needed. Need to sure up our apparmor profile creation process.

### To Test

Until @conor's pr is merged into all the current branches will need to jump through some hoops to get local packages installed in the staging environment.

* `vagrant up build`
* comment out the ossec packages for the `install_local_pkgs` var in `host_vars/app.yml`
* Comment out the `ansible.skip_tags =  [ "install_local_pkgs" ]` in the `Vagrantfile`
* Uncomment `ansible.skip_tags = [ "grsec",  "ossec", "app-test" ]`
* `vagrant destroy /staging/ -f`
* `vagrant up app-staging`

### Apparmor profile creation process

When creating the apparmor profile from scratch.

* Before putting the profile in enforce mode ensure that the apparmor profile grants enough permissions to allow apache to restart. If not you will get a false positive running aa-logprof at a later point prompting to provide dac_override capability
* Will look into providing a base template that just provides the permission/capabilities required for apache to restart without the additional permissions for the applications functionality.

### DIFF
--- a/install_files/ansible-base/usr.sbin.apache2
+++ b/install_files/ansible-base/usr.sbin.apache2
@@ -5,10 +5,6 @@
   #include <abstractions/base>
   #include <abstractions/tor>
 
-  signal (send) set=("term") peer=unconfined,
-  ptrace (trace) peer=unconfined,
-
-  capability dac_override,
   capability kill,
   capability net_bind_service,
   capability sys_ptrace,


conorsch:
It was a bit painful getting the `install_local_pkgs` role configured, but messing around with it gave me some ideas I documented in #1040. Thanks for the great write-up providing reproducible steps for testing, @dolanjs. Once the locally-built deb packages were installed, I did some manual testing on staging hosts, including:
* loading various pages on the source interface
* restarting the apache service
* rebooting the server
* checking logs for errors

Everything checked out OK, couldn't break anything. I also went back and added a few app-armor spectests in #1055 to target these config options. The new spectests check for the presence of declared `capability` options for `apache2` and `tor`, and also ensure that no extra capabilities are defined for those services.

Fine for merge. :+1:

dolanjs:
@conorsch When I'm creating the apparmor profiles the reason I skip those specific roles are:

* The `grsec` restrictions prevent running `aa-logprof`

* The `app-test` role installs additional dependencies and functionality that will lead you to be prompted to provide more permissions and capabilities than needed for prod in the apparmor profile.

* I skip the `ossec` role because the current `install_local_pkgs` role/playbook isn't set up to allow you to install just one package and not the others without modifying the host_vars/{app,mon}.yml config files.

I destroy the `mon-staging` host because I don't pass a limit host flag when running the ansible playbook command through vagrant. Since the ansible playbook is set to run multiple hosts, if you just run `vagrant up app-staging` while the `mon-staging` host is running the playbook will also be run against the `mon-staging` host again. There isn't an ansible tag to run the install_local_pkgs role on just the app-staging or just the mon-staging hosts.

garrettr:
:+1: Thanks for all of the detailed comments @dolanjs and @conorsch, they will be helpful for future reference.

PR #1059, opened 2015-06-23
iSec audit fixes

garrettr:
This PR includes fixes for iSEC-15FTC-2 and iSEC-15FTC-6. iSEC-15FTC-6 was only reported in the context of the document interface, but the DoS vector was equally applicable to the source interface so it is fixed there too.

All fixes have regression tests included.

### DIFF
--- a/securedrop/db.py
+++ b/securedrop/db.py
@@ -198,11 +198,11 @@ class InvalidPasswordLength(Exception):
     password length"""
 
     def __init__(self, password):
-        self.password = password
+        self.pw_len = len(password)
 
     def __str__(self):
-        if len(password) > Journalist.MAX_PASSWORD_LEN:
-            return "Password too long (len={})".format(len(password))
+        if self.pw_len > Journalist.MAX_PASSWORD_LEN:
+            return "Password too long (len={})".format(self.pw_len)
 
 
 class Journalist(Base):
--- a/securedrop/db.py
+++ b/securedrop/db.py
@@ -190,6 +190,18 @@ class BadTokenException(Exception):
     """Raised when a user logins in with an incorrect TOTP token"""
 
 
+class InvalidPasswordLength(Exception):
+    """Raised when attempting to create a Journalist or log in with an invalid
+    password length"""
+
+    def __init__(self, password):
+        self.password = password
+
+    def __str__(self):
+        if len(password) > Journalist.MAX_PASSWORD_LEN:
+            return "Password too long (len={})".format(len(password))
+
+
 class Journalist(Base):
     __tablename__ = "journalists"
     id = Column(Integer, primary_key=True)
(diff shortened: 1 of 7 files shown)


securedrop/source.py
```diff
@@ -124,6 +124,19 @@ def generate_unique_codename(num_words):
     """Generate random codenames until we get an unused one"""
     while True:
         codename = crypto_util.genrandomid(num_words)
+
+        # THe maximum length of a word in the wordlist is 6, and the maximum
```
conorsch: Nit: s/THe/The/

securedrop/source.py
```diff
@@ -124,6 +124,19 @@ def generate_unique_codename(num_words):
...
+        # codename length is 10, so it is currently impossible to generate a
```
conorsch: The numbers 6 and 10 refer to different units (characters and words, respectively); add the units for clarity.

securedrop/db.py
```diff
@@ -190,6 +193,18 @@ class BadTokenException(Exception):
     """Raised when a user logins in with an incorrect TOTP token"""
 
 
+class InvalidPasswordLength(Exception):
+    """Raised when attempting to create a Journalist or log in with an invalid
+    password length"""
+
+    def __init__(self, password):
+        self.password = password
```
conorsch: Is it necessary to store the invalid password? If it's stored only for reporting the length, why not just store the length?

securedrop/db.py
```diff
@@ -190,6 +193,18 @@ class BadTokenException(Exception):
...
+
+    def __str__(self):
+        if len(password) > Journalist.MAX_PASSWORD_LEN:
```
conorsch: If statement will throw `NameError: global name 'password' is not defined`. Should be `len(self.password)`.

securedrop/db.py
```diff
@@ -190,6 +193,18 @@ class BadTokenException(Exception):
...
+            return "Password too long (len={})".format(len(password))
```
conorsch: Format call will throw `NameError: global name 'password' is not defined`. Should be `len(self.password)`.

garrettr: Good catch! Funny, because I was sure I had tested this... /me double-checks

securedrop/db.py
garrettr: Just seemed more flexible for future functionality, but could also just store the length to be safe.

conorsch:
8af96ba addresses my concerns about:

* unhandled exceptions
* storing of (invalid) passwords

Other concerns include setting a `MIN_PASSWORD_LEN` constant, but were already tracking password requirements in #980, and adding the functionality as part of this PR would be out of scope. The way you've written `InvalidPasswordLength` here, it can easily be extended to address minimum length requirements down the road. 

While I'm not convinced the new exception needs a `__str__` method at all, the prod config settings will ensure the string representation never gets logged. The fact that only the length, rather than the password itself, is logged even in the dev environments further eases my mind.

Good to have new tests for the added functionality, but it's worth noting that the new tests didn't catch the unhandled `NameError` exception because it was buried in the `__str__` method, which is only called when logging is enabled, logging is not enabled during test runs. Going forward we can guard against similar breakage by: 

* adding behavioral testing (hopefully by 0.4)
* ensuring every new line of code is absolutely necessary

With the addition of 54eec31, I confirmed by manual testing that the following statements are true:

* attempting to create a user with a too-long password fails with informative GUI message
* attempting to create a user with a too-long password logs the exception to the flask dev logs
* attempting to edit a user to have a too-long password fails with informative GUI message
* attempting to log in to an existing journalist account with a too-long password fails without an overly informative GUI message
* attempting to log in to a nonexistent journalist account with a too-long password fails without an overly informative GUI message

Can't find any way to break this new functionality, so I say fine to merge. :+1:

garrettr:
@conorsch Thanks for that detailed write-up of our group code review, and thanks for doing a final round of QA!

Issue #1025, opened 2015-05-13
Network Firewall doc needs correcting


harlo:
References to "Connection Information" should be replaced with "Edit Connections...".  Replace corresponding screencap, too.

See:

https://github.com/freedomofpress/securedrop/blob/0.3.3/docs/network_firewall.md#disable-dhcp-on-the-lan

harlo:
Another thing: clarify CIDR allocation for OPT1 and LAN (24)
Because pfsense will "guess" for you, sometimes it guesses 32, which is wrong.

harlo:
Another thing: specify that username + pwd have to be the same for both app and mon.

harlo:
Another thing: admin ssh key should be passwordless (prefered) or diceword-generated.  this is because pinentry doesn't let you copy/paste.

conorsch:
Reformatting for clarity:

### Things to change
* [ ] Replace `Connection Information` with `Edit Connections`
* [ ] Replace screenshot that mentions `Edit Connections`
* [ ] Clarify CIDR allocation for OPT1 and LAN on /24

harlo:
Turns out, only item #3 (the CIDR mapping bit) is relevant.  Items 1 + 2 are fine the way they are.

garrettr:
This was merged for release/0.3.4 in #1060. Taking this off of the 0.3.4 milestone as using #1060 as the marker there instead, since this will also need to be resolved for the develop branch.

garrettr:
These changes were merged into develop in #1075.

PR #1060, opened 2015-06-23
Network Firewall docs (issue #1025)

harlo:
as per https://github.com/freedomofpress/securedrop/issues/1025

### DIFF
--- a/docs/network_firewall.md
+++ b/docs/network_firewall.md
@@ -81,7 +81,7 @@ Leave the defaults for "Time Server Information". Click Next.
 
 On "Configure WAN Interface", enter the appropriate configuration for your network. Consult your local sysadmin if you are unsure what to enter here. For many environments, the default of DHCP will work and the rest of the fields can be left blank. Click Next.
 
-For "Configure LAN Interface", set the IP address and subnet mask of the Application Subnet for the LAN interface.  Be sure that the CIDR prefix correctly corresponds to your subnet mask-- pfsense should automatically calculate this for you, but you should always check. Click Next.
+For "Configure LAN Interface", set the IP address and subnet mask of the Application Subnet for the LAN interface.  Be sure that the CIDR prefix correctly corresponds to your subnet mask-- pfsense should automatically calculate this for you, but you should always check. In most cases, your CIDR prefix should be `/24`.  Click Next.
 
 Set a strong admin password. We recommend generating a random password with KeePassX, and saving it in the Tails Persistent folder using the provided KeePassX database template. Click Next.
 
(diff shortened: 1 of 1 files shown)


garrettr:
:+1: What do you think @dolanjs @conorsch?

docs/network_firewall.md
```diff
@@ -81,7 +81,7 @@ Leave the defaults for "Time Server Information". Click Next.
 
 On "Configure WAN Interface", enter the appropriate configuration for your network. Consult your local sysadmin if you are unsure what to enter here. For many environments, the default of DHCP will work and the rest of the fields can be left blank. Click Next.
 
-For "Configure LAN Interface", set the IP address and subnet mask of the Application Subnet for the LAN interface. Click Next.
+For "Configure LAN Interface", set the IP address and subnet mask of the Application Subnet for the LAN interface.  Be sure that the CIDR prefix correctly corresponds to your subnet mask-- pfsense should automatically calculate this for you, but you should always check. Click Next.
```
dolanjs: Should we add that for the values used in the doc the CIDR should be `/24` So they don't need to look it up?

harlo:
...maybe we add something like, "default will always be `/24`"?  are other configs supported?

docs/network_firewall.md
conorsch: :+1: Furthermore, in L36, I suggest: `This is a very common subnet choice for home routers.` -> `The /24 subnet is a common choice for home routers.` Specifically referring to the subnet by CIDR notation adds clarity for non-technical folks, at no cost to clarity for technical audiences.

dolanjs:
In the install docs we recommend using  the 10.20.2.0 network with a subnet mask 255.255.255.0 for the monitor server's network. `/24` is the CIDR notation for the 255.255.255.0 subnet mask.

It is configurable by the admin to accommodate certain edge cases. One of which is if the 10.20.1.0 or 10.20.2.0 networks are in use someplace else in the corporate environment. Then the admin would want to use a different network(s) for the app and monitor server networks to avoid routing table issues with SD network firewall.

Any internal ip range netmask that provides 2 usable IP addresses will work.

conorsch:
One more change: In L36 `This is a very common subnet choice for home routers.` -> `The /24 subnet is a common choice for home routers.` Specifically referring to the subnet by CIDR notation adds clarity for non-technical folks, at no cost to clarity for technical audiences.

Then we should be all set.

garrettr:
Agree with @conorsch's last comment. One of the best pieces of advice about writing clearly I've ever received is to always carefully consider the use of the word "this". It may seem clear to the writer at the time of writing due to their mental context, but it is often ambiguous and confusing!

garrettr:
lgtm. Ready to merge @harlo?

Issue #1027, opened 2015-05-14
Server Setup doc clarifications


harlo:
1) specify that username + pwd have to be the same for both app and mon.  
2) admin ssh key should be passwordless (prefered) or diceword-generated. this is because pinentry doesn't let you copy/paste.

see:
https://github.com/freedomofpress/securedrop/blob/0.3.3/docs/install.md#set-up-the-servers

conorsch:
Reformatting for clarity:

### Things to change:

* [x] Specify that username and password must be the same for both servers.
* [x] Passphrase on Admin SSH key should be absent (prefered) or diceword-generated. (This is because pinentry doesn't let you copy/paste.)

Both concerns are addressed in #1061.

PR #1061, opened 2015-06-23
Server Setup docs (issue #1027)

harlo:
see https://github.com/freedomofpress/securedrop/issues/1027


conorsch:
Nit: trailing whitespace introduced on L43 and L51.

Otherwise, looks great. Closes #1027.

garrettr:
lgtm, agree with @conorsch about the trailing whitespace. Just get rid of it entirely.

PR #1065, opened 2015-06-26
Update toc for docs

garrettr:
I ran `doctoc .`, which created TOC for a lot of docs that didn't have
one previously. I skimmed the diff and removed TOC from some documents
that didn't seem to benefit from having one, but left the rest in.


garrettr:
@harlo @conorsch r? Do you think it's a good idea to add a TOC to so many docs that didn't have one before, or should I just restrict this to updating TOC's for docs that already had one?

garrettr:
Fun fact about doctoc: it can take a single file or a directory as input, but it cannot take multiple files. If you pass multiple filenames, it only processes the first one.

So, to only update the docs that already have TOCs: `grep "doctoc generated TOC" *.md | cut -d':' -f1 | uniq | xargs -n 1 doctoc`.

Only 1 doc gets updated, and it is not a material change (doctoc just updates the URL in the comment header). Therefore I'm going to close this PR, and we can worry about adding TOCs to other documents in the develop branch.

harlo:
it's up to you, but in my opinion as a would-be brand-new SD admin, having more anchors to jump between is always good.

PR #1063, opened 2015-06-25
Updated 0.2.1 upgrade doc

dolanjs:
* Modified the 0.2.1 upgrade/migration doc to relect that the process is the same for all the `0.3.x` versions

* Changed wording to reflect that the admin is using the web browser when first viewing the upgrade doc rather than using cli.

The wording for how to verify/change the branch when using the browser compared to CLI can probably be improved and also used in the install.md

PR #1066, opened 2015-06-26
Use 0.3.x for all upgrade documentation

garrettr:
The documented upgrade paths from 0.2.1 and 0.3pre will work for any
0.3.x release. Therefore, we should explain this and direct an admin who
is upgrading to upgrade to the *latest* 0.3.x minor version.

This process was begun in #1063 and is completed for all relevant
documentation here.


dolanjs:
lgtm

PR #1002, opened 2015-04-21
Serverspec tests

conorsch:
Here's a pile of serverspec tests. This is a large PR, but the vast majority of changes are isolated in the `spec_tests` dir, and so shouldn't have any side-effects for the application code or provisioning process. The only files that have changed outside of the `spec_tests` dir are these:

```
.gitignore
Vagrantfile
docs/develop.md
docs/spec_tests.md
install_files/ansible-base/roles/app/templates/sites-available/document.conf
install_files/ansible-base/roles/restrict_direct_access_app/templates/app_rules_v4
install_files/ansible-base/roles/restrict_direct_access_mon/templates/mon_rules_v4
```

The changes to the Ansible files are strictly typographic, to make the serverspec tests more consistent. The change to the Vagrantfile is a conditional check to run the digitalocean provider call only if it's properly configured. See the changes to `docs/spec_tests.md` for how to run the new spectests, and make sure the recommended procedure works on your machine:

```
cd spec_tests
bundle update
bundle exec rake spec
```

The commands above will run `vagrant up` on the following VirtualBox machines:

* develop
* app-staging
* mon-staging
* build

There's a bit of magic in the `spec_tests/spec/spec_helper.rb` file to load variables for running the tests, but these variables are *not* read from the Ansible vars files. This separation is intentional, so that changes to Ansible files will require a manual change to the spectests in order to get them back in sync.

Further down the line these tests should be reorganized by roles, rather than hosts, but first we'll need to make that true of the Ansible layout, so the tests mirror the config being tested. More tests coming in the near future, but I consider the coverage provided by this PR sufficient to close #947.

Issue #1067, opened 2015-07-01
Add automatic testing of staging environment in CI


conorsch:
Currently all pull requests to this repository trigger the Travis CI tests, and the results are automatically displayed inline:

![travis_ci_pr](https://cloud.githubusercontent.com/assets/657862/8445544/7d50ac2c-1f53-11e5-95f5-dbb8bc63d979.png)

The same functionality is possible with Snap CI, but before enabling it, we need to confirm that the Snap config:
- [x] coexists peacefully with Travis tests webhook (concerns about interoperability mentioned in #1055)
- [x] is relatively stable (don't want false negatives because the test suite itself is poorly configured)
- [ ] has sufficient documentation (specifically, a README pointing to `docs/spec_tests.md` with URLs to view and manage the tests online)
- [ ] doesn't divulge unnecessary information.

The last point is particularly important. The Snap docs warn:

![snap_ci_pr_warning](https://cloud.githubusercontent.com/assets/657862/8445657/f67d9bc2-1f54-11e5-887c-9a35777d1fc4.png)

Relatively sensitive information that could be leaked to a malicious PR includes:
- SSH key for DO droplets (not currently dynamically generated on each build, but theoretically could be)
- ATHS HidServ auth values for testing instances (risk negligible, since testing instances are temporary)
- DO API token (not much we can do to mask this; we'd have to regenerate if it's divulged).

Mostly the risks here are just an annoyance, rather than actual security concerns, since PRs are public and we're notified about them immediately. I'll tinker with dynamically generating the SSH keys used to connect to the testing instances. If that can be locked down, I'll be satisfied.

Once the prod tests are added, there should be a review of documentation prior to enabling public notification of the Snap tests.

conorsch:
Once #1214 is merged, we'll be in a good position to enable the webhook integration for Snap CI. As an added bonus, most of the CI test environments will be executable locally, via bash scripts, on both VirtualBox and DigitalOcean providers.

conorsch:
Picking this back up. Now that we have automatic building of deb packages for the staging environment (#1464), as well as configuration tests (#1616), we should be running full staging builds in CI and validating with the config test suite thereafter. The was substantial work on this done in the `snap_ci` directory, which wraps the provisioning logic. Those tasks will need to be updated, particularly the config test calls, but largely should work.

[...]

PR #1070, opened 2015-07-06
Adds serverspec tests for prod hosts

conorsch:
Introducing Serverspec tests for the `app-prod` and `mon-prod` VMs. The addition of the `/prod/` hosts to the test suite required a significant reorganization to the host-based spectests added in #1055, moving to a mostly role-based organization for the spectests. There are exceptions; the tests currently reflect the organization of the Ansible playbooks, and have a similar use of dynamic variables. 

### `SECUREDROP_` prefix for env vars
All SecureDrop-related environment variables are now prefixed with `SECUREDROP_`. You may need to update the following in your local environment:

 * `DEVELOPMENT_SKIP_TAGS` -> `SECUREDROP_DEVELOPMENT_SKIP_TAGS`
 * `STAGING_SKIP_TAGS` -> `SECUREDROP_STAGING_SKIP_TAGS`
 * `PROD_SKIP_TAGS` -> `SECUREDROP_PROD_SKIP_TAGS`

### SSH over Tor
The `vagrant ssh` command can now connect to `/prod/` instances over Tor. Prior to connecting over SSH, you'll still need to add the ATHS values to your torrc file and reload tor, then `export SECUREDROP_TOR_OVER_SSH=1` to tell Vagrant to use Tor. It will read the `app-ssh-aths` and `mon-ssh-aths` files on disk and use those to generate the SSH config. An example terminal session showing how this env var is configured has been added to `docs/develop.md`.

### `snapci` machine in Vagrantfile
The new `snapci` VM models the Snap CI test nodes. Significantly, this box is based on CentOS 6.6, rather than Ubuntu LTS. This machine can be used for debugging the Snap playbooks. A corresponding `snapci` Ansible role is used to configure this machine both locally and as part of the Snap prod test stage.

### `Prod_Clean` stage in Snap CI
Due to the addition of SSH connectivity over Tor, it's now possible to run the serverspec tests against the prod hosts. Accordingly a new stage in Snap CI has been created for just this purpose. See `snap_ci/prod_clean.sh` for a detailed look. In a nutshell, the stage creates prod machines in DigitalOcean, provisions them (with more restrictive iptables rules), then runs the serverspec tests against them over Tor. As with the staging hosts, the grsecurity tasks and tests are skipped in Snap, but can be run locally.

### Related PRs
* #1002 (introduction of serverspec tests)
* #1055 (expansion of serverspec tests)


conorsch:
@garrettr Take a look at `docs/develop.md` and see if the OS X config still makes sense. I suspect I'll need to add:
```
  elsif command?("torify")
    base_cmd = "torify"
```
to the `tor_ssh_proxy_command` def in the `Vagrantfile`, but can't test without a Mac. Let me know how it works for you, based off the terminal session in `docs/develop.md`, and I'll update this PR as necessary.

garrettr:
@conorsch Tested on Mac OS X (10.10 Yosemite) and confirmed that spec tests run as-is. `nc` is included by default on 10.10 (`/usr/bin/nc`) and it successfully proxies ssh over Tor.

Vagrantfile
```diff
@@ -187,3 +214,50 @@ Vagrant.configure("2") do |config|
...
+# The Ansible playbooks fetch these values back to the
+# Admin Workstation (localhost) so they can be manually
+# added to the inventory file. Possible values for filename
+# are "app-ssh-aths" and "mon-ssh-aths".
+def find_ssh_aths(filename)
+  require 'find'
+  repo_root = File.expand_path(File.dirname(__FILE__))
+  ansible_dir = File.join(repo_root, "install_files", "ansible-base")
+
+  Find.find(ansible_dir) do |path|
```
garrettr: A recursive search of the subdirectories under `ansible_dir` seems unnecessary, since `*-ssh-aths` are always placed in `ansible_dir`. Can't you just get the file from the well-known path?

conorsch: @garrettr You're right, I'll patch.

docs/develop.md
```diff
@@ -158,44 +163,60 @@ cd /var/www/securedrop/
...
+  UserKnownHostsFile /dev/null
+  StrictHostKeyChecking no
+  PasswordAuthentication no
+  IdentityFile /home/conor/.vagrant.d/insecure_private_key
+  IdentitiesOnly yes
+  LogLevel FATAL
+  ProxyCommand connect -R remote -5 -S 127.0.0.1:9050 %h %p
+$ # ensure ATHS values are active in local Tor config:
+$ cat *ssh* | sudo tee -a /etc/tor/torrc > /dev/null && sudo service tor reload
+$ vagrant ssh app-prod -c 'echo hello'   # works
```
garrettr: This is awesome, I didn't realize we could do this through Vagrant. No more manually updating my `~/.ssh/config` when testing prod! \o/

garrettr: Although one still has to update their `torrc` manually to add the `HidServAuth` values.

install_files/ansible-base/roles/snapci/tasks/tor_config.yml
```diff
@@ -0,0 +1,27 @@
...
+    line: "{{item}}"
+    regexp: "^{{item}}"
+  with_items:
+    - "{{mon_ssh_aths}}"
+    - "{{app_ssh_aths}}"
+  notify: reload tor
+  tags:
+    - tor
+
+- name: restart tor
```
garrettr: It seems unnecessary to reload **and** restart Tor after modifying the torrc. Or am I missing something?

snap_ci/pinned.sh
```diff
@@ -5,23 +5,33 @@
...
+[[ -f "${tmp_dir}/${vagrant_rpm}" ]] || wget "$vagrant_url" -O "${tmp_dir}/${vagrant_rpm}"
+[[ -x /usr/bin/vagrant ]] || sudo -E rpm -ivh "${tmp_dir}/$vagrant_rpm"
 
 # An older version of digital_ocean plugin is used because of an issue with the
 # current version that doesn't support using snapshots.
 # https://github.com/smdahlen/vagrant-digitalocean/issues/187
 #/usr/bin/vagrant plugin install vagrant-digitalocean --plugin-version '0.7.0'
-/usr/bin/vagrant plugin install vagrant-digitalocean
-
+vagrant plugin install vagrant-digitalocean
```
garrettr: If the above comment about using an older version of `vagrant-digitalocean` is obsolete (which it appears to be because this command does not specify the older version to install) then can we remove it?

docs/develop.md
conorsch: It's not as awesome as it looks: `vagrant ssh` works, which is magnificent, but `vagrant provision` does *not* work, since the proxycommand is not passed to Ansible via `ANSIBLE_SSH_ARGS` or otherwise. (At a glance, we'd need an upstream patch, either in the Vagrant ansible provisioner, or in Ansible itself, to support provisioning over Tor with this method.)

So the magic sauce is great for permitting quick access to log in and debug a "prod" test server, but you still need to call `ansible-playbook` from the command line to reprovision them over Tor. That's actually a convenient limitation, in that it forces use of the same process admins use in production, but it does reduce convenience a bit during development.

snap_ci/prod_clean.sh
```diff
@@ -0,0 +1,81 @@
...
+# so it must be rerun in each stage.
+repo_root=$( dirname "$( cd "$( dirname "${BASH_SOURCE[0]}"  )" && pwd )" )
+
+# Copy staging vars for use as prod vars; will need to skip "validate" role.
+sed -r -e 's/(app|mon)-staging/\1-prod/' \
+    "${repo_root}"/install_files/ansible-base/staging-specific.yml > \
+    "${repo_root}"/install_files/ansible-base/prod-specific.yml
+
+# Skip "validate" so reused vars don't cause failure,
+# and skip "grsec" because DigitalOcean hosts don't support custom kernels.
```
garrettr: Can we open a follow-up issue for running grsecurity/custom kernels on DigitalOcean? grsecurity's hardening is one of the key breakage-inducing things that we are trying to test via these prod tests, so I see this as a semi-blocker for integrating with Github (#1067).

snap_ci/prod_clean.sh
```diff
@@ -0,0 +1,81 @@
...
+export SECUREDROP_PROD_SKIP_TAGS=validate,grsec
+
+# Assume that /prod/ instances will be created from scratch, as part
+# of a full provisioning run, so initial SSH connections must be
+# made directly, rather than over Tor.
+unset SECUREDROP_SSH_OVER_TOR
+
+# Up the host in a separate command to avoid snap-ci command timeouts.
```
garrettr: You mean **provision** the host in a separate command to avoid snap-ci timeouts, right?

snap_ci/prod_clean.sh
conorsch: Clean integration with custom kernels in cloud providers is going to be a challenge. Might be possible with `kexec`, but even then we've got a divergent config specially for testing—not a good place to be in, but perhaps better. I'll open an issue.

spec_tests/spec/common-development/securedrop_app_spec.rb
```diff
@@ -2,9 +2,13 @@
 # common to "development" and "app-staging"
 
 # ensure default apache html directory is absent
-describe command('/bin/bash -c "[[ ! -e /var/www/html  ]]"') do
-  its(:exit_status) { should eq 0 }
-end
+# TODO: The directory check below fails in staging and production
+# because the Ansible task to remove the directory runs before
+# the directory has been automatically replaced by Apache.
+# Fix the `app` role playbook, then update this test.
```
garrettr: Is there an issue for this?

conorsch: No, it's too minor an issue by itself to warrant one. It's certainly present in the list of items to be addressed in the Ansible reorg, which should get underway in the very near future, once we have the test suite ironed out.

snap_ci/prod_clean.sh
ageis: Per my interaction with DO support, custom kernel should be possible on Debian 8 or Ubuntu 15.04, but is managed at hypervisor level on all other distros. As @conorsch said, kexec might be the way.

install_files/ansible-base/roles/snapci/tasks/tor_config.yml
conorsch: Good catch, this is indeed superfluous. Removed in ef4c396.

snap_ci/prod_clean.sh
conorsch: Preventing timeouts in snapci is now down in the snap config itself, rather than inside these bash scripts. It's still necessary to run the `up` and `provision` commands separately, and I've updated the associated comments in efcb1c9 to explain.

conorsch:
Thanks for the detailed comments, @garrettr. I've gone through and implemented a few changes as a result. Let me know if you think anything else should be changed before merging into develop.

docs/develop.md
garrettr: Good to know, thanks for explaining!

snap_ci/pinned.sh
conorsch: Addressed in aba4368.

docs/develop.md
```diff
@@ -158,44 +163,60 @@ cd /var/www/securedrop/
...
+  Port 22
+  UserKnownHostsFile /dev/null
+  StrictHostKeyChecking no
+  PasswordAuthentication no
+  IdentityFile /home/conor/.vagrant.d/insecure_private_key
+  IdentitiesOnly yes
+  LogLevel FATAL
+  ProxyCommand connect -R remote -5 -S 127.0.0.1:9050 %h %p
+$ # ensure ATHS values are active in local Tor config:
+$ cat *ssh* | sudo tee -a /etc/tor/torrc > /dev/null && sudo service tor reload
```
garrettr: I usually use `cat *aths`, which gets the Document Interface `HidServAuth` value too.

conorsch: There's no reason not to include the Document Interface values too, so added in 27a9c65.

Issue #1076, opened 2015-07-21
Don't pin the leaf certificate in Postfix configuration


garrettr:
The Postfix configuration on the Monitor Server currently uses certificate fingerprint verification for TLS connections to the SMTP relay. As stated in the [docs](http://www.postfix.org/TLS_README.html#client_tls_fprint):

> Fingerprint verification may be feasible for an SMTP "VPN" connecting a small number of branch offices over the Internet, or for secure connections to a central mail hub. It works poorly if the remote SMTP server is managed by a third party, and its public certificate changes periodically without prior coordination with the verifying site.

While we encourage SecureDrop installations to use their own organization's SMTP servers, it is common for them to use public SMTP servers (such as `smtp.gmail.com`) if they cannot use their own. In this case, the instances fall into the "works poorly" camp.

For example, Google rotates the certificate for `smtp.gmail.com` regularly. They appear to do it approximately every 3-4 months, since they just rotated it this morning and our last issue referring to the same problem was from around March 23rd - https://github.com/freedomofpress/securedrop/issues/957#issuecomment-87006791. When they rotate, the certificate's fingerprint changes and subsequent validations fail, which causes sending to be deferred.  This causes all OSSEC alerts to stop being delivered without any warning to the admins, and requires manual intervention from the admins (update the fingerprint in `prod-specific.yml` and re-run the playbook) to resolve.

garrettr:
Aaand Google appears to have switched back to the old cert again, here's a recent re-check:

```sh
~/code/freedomofpress/securedrop on develop 15:52:48
$ openssl s_client -connect smtp.gmail.com:587 -starttls smtp < /dev/null 2>/dev/null | openssl x509 -fingerprint -noout -in /dev/stdin | cut -d'=' -f2
D3:7C:82:FC:D0:5F:8F:D7:DA:A2:59:8C:42:D7:B2:9F:C1:9F:7E:60
```

conorsch:
Depending on location, we've seen all of the following fingerprints within the same 10-minute window:

* `6D:87:EE:CB:D0:37:2F:88:B8:29:06:FB:35:F4:65:00:7F:FD:84:29`
* `D3:7C:82:FC:D0:5F:8F:D7:DA:A2:59:8C:42:D7:B2:9F:C1:9F:7E:60`
* `4F:6D:5C:58:94:28:C1:A4:36:B9:C4:B5:50:09:FC:5F:FB:47:91:BD`

[...]

PR #1077, opened 2015-07-21
Update smtp.gmail.com fingerprint

garrettr:
The certificate was rotated this morning, as SMTP delivery with the old
fingerprint was working during some testing I was doing yesterday afternoon.

This is a temporary fix that allows staging environments to continue working by default. A long-term fix for this problem is tracked in #1076


dolanjs:
@garrettr I think you should also change it in the ossec_alerts.md otherwise looks good.

conorsch:
Also in `spec_tests/spec/mon-staging/ossec_server_spec.rb`.

conorsch:
On second thought, don't update the spec tests in this PR; I'll add the change to #1070, which is still pending. Reason being that the target file has moved as of #1070 so updating the fingerprint for the tests over there will prevent a merge conflict.

garrettr:
Rebased and merged into develop in d6f9386e9a512cadadf7b608cad199be8345fce8.

garrettr:
Note: I merged the branch locally, then rebased and pushed. It would've been better to rebase the feature branch and push it here, _then_ merge. That would have avoided the "Closed with unmerged commits" message, and would make it easier to associate commits in the PR with commits on the develop branch.

PR #1079, opened 2015-07-22
Update SecureDrop logo so it's higher quality

ageis:
The logo in the repository is apparently not transparent and has artifacts from being previously compressed, I think it was converted from a JPEG.


garrettr:
:+1:

PR #1081, opened 2015-07-22
Remove SMTP fingerprint requirement for OSSEC Postfix config

conorsch:
As discussed in #1076, requiring a fingerprint match for SMTP servers can cause OSSEC notifications to fail to send, since the fingerprint can change at any time. The solution requires admin intervention in the form of querying the new fingerprint, updating the playbook, and rerunning the playbook. 

A less onerous workflow is to leverage the `smtp_tls_security_level = secure` setting in the Postfix configuration, which requires a secure channel for negotiation, then trusts the lookup as long as it passes verification.

The changes in this PR update the postfix config to use the system CAs present in `/etc/ssl/certs` (and _only_ those CAs) by default. In case an admin wants to provide a custom certificate for the SMTP relay, the admin should declare the `smtp_relay_cert_override_file` var, which will cause the custom certificate to be added to `/etc/ssl/certs_local`.

No documentation changes have been made yet, pending discussion of this implementation.


dolanjs:
lgtm. Tested this using smtp.gmail.com as the relay and a google business account for the sasl authentication.

ageis:
looks good!

install_files/ansible-base/roles/ossec_server/tasks/mon_configure_custom_cert.yml
```diff
@@ -0,0 +1,39 @@
...
+  # Create a separate directory for storing a custom
+  # cert file, if declared. A separate dir is necessary
+  # because otherwise removing a custom cert from the
+  # Ansible config would NOT result in the custom cert
+  # being removed from the /etc/ssl/certs.
+- name: ensure custom certificate directory exists (if necessary)
+  file:
+    path: "{{ smtp_relay_cert_override_dir }}"
+    state: directory
+    mode: "0755"
```
garrettr: FWIW, from the Postfix TLS docs:
> The file is opened (with root privileges) before Postfix enters the optional chroot jail and so need not be accessible from inside the chroot jail.

So in theory you could tighten the permissions to `0700` without breaking functionality. However, that change isn't necessary because the contents of the _certificate_ file can be world-readable without harm.

garrettr:
Tested this on virtual staging environment with `smtp.gmail.com` and `garrett@freedom.press` and am successfully receiving OSSEC alerts. :+1:

garrettr:
Let's update the docs in this PR, then discuss the strategy for merging this into a release and updating existing installs.

docs/ossec_alerts.md
```diff
@@ -72,23 +71,19 @@ Next you specify the e-mail that you'll be sending alerts to, as `ossec_alert_em
...
+First you should [sign up for a new account](https://accounts.google.com/SignUp?service=mail). You may use an existing Gmail account, but it's best to compartmentalize these alerts from any of your other activities. Choose a strong and random passphrase. Skip the creation of a Google+ profile and continue straight to Gmail. Once the account is created you can log out and provide the values for `sasl_username` as your new Gmail username (without the domain), `sasl_domain`, which is typically gmail.com (or your custom Google Apps domain), and `sasl_passwd` for the password. The `smtp_relay` is smtp.gmail.com and the `smtp_relay_port` is 587.
 
 For enhanced security we recommend enabling [Google's 2-Step Verification](https://www.google.com/landing/2step/) for any Gmail account that is dedicated to sending the alert emails. With 2-Step Verification enabled, you won't use the normal account password in this configuration — it will not work; instead you must navigate (using the settings in the top right) to Account > Signing in > App passwords, and generate a new App password which you will use as the `sasl_passwd`.
```
dolanjs: Think this should be re-worded to make it mandatory. Google will request additional info (captcha...) if you configure postfix to authenticate using the accounts master password.

conorsch: @dolo What should be mandatory? Using a Gmail account? Creating a new Gmail account instead of reusing one?

conorsch:
While editing the docs I intentionally preserved the insane line length so the diffs were readable. But we should consider folding our docs so that small edits to a sentence don't result in a paragraph-long diff.

docs/ossec_alerts.md
dolanjs: Sorry @conorsch If the instance is using google (gmail or google business account) enabling 2 step verification and using an app specific password for the `sasl_password` should be mandatory not just "For enhanced security"

garrettr:
Yeah, we've been using different conventions for line lengths across the documentation. I've been trying to keep lines wrapped at around 80 characters for that exact reason (it makes diffs easier to review). We should start mandating that, it's a simple tweak of editor settings.

docs/ossec_alerts.md
conorsch: @dolanjs Thanks for clarifying, I completely agree. Implemented in db04dbb.

docs/ossec_alerts.md
```diff
@@ -72,23 +71,19 @@ Next you specify the e-mail that you'll be sending alerts to, as `ossec_alert_em
 	
 Now you can move on to the SMTP and SASL settings, which are straightforward. These correspond to the outgoing e-mail address used to send the alerts instead of where you're receiving them. If that e-mail is ossec@news-org.com, the `sasl_username` would be ossec and `sasl_domain` would be news-org.com.
 
-The Postfix configuration enforces certificate verification, and requires a fingerprint to be set. You can retrieve the fingerprint of your SMTP relay by running the command below (all on one line). Please note that you will need to replace `smtp.gmail.com` and `587` with the correct domain and port for your SMTP relay.
+The Postfix configuration enforces certificate verification, and requires both a valid certificate and STARTTLS support on the SMTP relay. By default the system CAs will be used for validating the relay certificate. If you need to provide a custom CA to perform the validation, copy the cert file to `install_files/ansible-base` add a new variable to `prod-specific.yml`:
```
dolanjs: Should we explain overriding the default for the security level form `secure` to `verify` and the security trade off between the two?

garrettr: I think we should prefer `secure` and not bother documenting other secure connection settings as long as `secure` works in a large enough percentage of real world use cases.

garrettr:
This PR does not remove the `smtp_relay_fingerprint` from `{staging,prod}-specific.yml`. This is tricky because we don't want to leave a useless option in the configuration, but we also have to be careful with handling merge conflicts that can arise when changing the `*-specific.yml` files due to the current "stash-pull-pop" workflow that admins use to update their local SecureDrop repositories.

One option that James suggested was leaving in the `smtp_relay_fingerprint` as a third option (that would override the default cert verification and `smtp_relay_cert_override_file`. This could be useful for orgs that are running their own SMTP relay and are willing to maintain those certificates and update SecureDrop in a timely manner if and when those certificates change.

On the one hand, this is the most secure possible option. On the other, I'm not sure if I believe any instances will be running their own SMTP server and/or will be able to update this kind of configuration parameter in a timely manner.

Either way, we will need to develop a mechanism (manual or automated) to help 0.3.4 admins migrate their `prod-specific.yml` files to 0.3.5. Some may wish to continue using fingerprints, so their migration path will depend on whether we decide to continue supporting them. Most are using Gmail, and they should switch to the default cert verification to avoid problems with frequent cert rotation and load balancing.

docs/ossec_alerts.md
conorsch: The security level can be overridden during provisioning by appending `--extra-vars smtp_relay_security_level=verify` to the playbook command, or by adding that var to `prod-specific.yml`. For admins that want to adjust the security level, overriding the var will allow them to customize the environment without editing the postfix config template. Intentionally not documenting this for now, since it's quite advanced functionality, and we don't have enough data on use cases to conclude that anyone would actually take advantage of it.

conorsch:
No objections to removing the `smtp_relay_fingerprint` from `staging-specific.yml`, but `prod-` is indeed harder, which is why I left it in both places. I'll  work on retaining `fingerprint` support while keeping `secure` the default. In that case, we can still leave the fingerprint var in the prod vars file, in case anyone wants to add it. We can revisit modifying the vars file down the road.

conorsch:
Based on offline discussion with @garrettr, I've re-added support for the `fingerprint` security level. The resolution order of configuration options now looks like this, on a fresh install:

1. Default to `secure` setting, trusting system CAs in `/etc/ssl/certs`.
2. If `smtp_relay_cert_override_file` is declared, still use `secure` setting, but don't trust system CAs, trust _only_ the specified cert.
3. If `smtp_relay_fingerprint` is declared, switch to `fingerprint` setting, and require an exact match for the specified fingerprint.

For existing installs, however, it's a bit different. Admins will need to manually remove the `smtp_relay_fingerprint` var from the `prod-specific.yml` file in order to prevent a merge conflict. Specifically, the admin should:

  1. delete the fingerprint line from `prod-specific.yml`
  2. stash save local changes
  3. pull update
  4. pop saved stash

If the admin wants to retain fingerprint checking, leaving the fingerprint declared in `prod-specific.yml` will honor that decision. It may result in a merge conflict on pull, however, which should be resolved manually.

We'll need to contact admins and explain the manual workflow prior to releasing an upgrade.

docs/ossec_alerts.md
```diff
@@ -40,7 +41,6 @@ These are the values you must specify in `prod-specific.yml`:
  * Email username to authenticate to the SMTP relay: `sasl_username`
  * Domain name of the email used to send OSSEC alerts: `sasl_domain`
  * Password of the email used to send OSSEC alerts: `sasl_password`
- * The fingerprint of your SMTP relay: `smtp_relay_fingerprint`
```
garrettr: You should document the three supported tiers of TLS connection security here. Maybe re-include this line but add "(Optional)" at the beginning, add a line for the smtp cafile override, and add a paragraph somewhere explaining how to choose between the three options.

install_files/ansible-base/roles/ossec_server/templates/main.cf
```diff
@@ -3,11 +3,32 @@ relayhost = [{{ smtp_relay }}]:{{ smtp_relay_port }}
...
+# copied to the server and ONLY the cert listed below will be used
+# for validating TLS connections to the SMTP relay.
+smtp_tls_security_level = secure
+smtp_tls_CAfile = {{ smtp_relay_cert_override_dir }}/{{ smtp_relay_cert_override_file }}
+{% else %}
+# Default configuration, when no overrides are specified.
+# Uses the system CAs to validate relay TLS certificate.
+smtp_tls_security_level = secure
+smtp_tls_CApath = {{ smtp_relay_cert_dir }}
+{% endif %}
```
garrettr: This code is fantastic, clear and well-commented. :metal:

docs/ossec_alerts.md
conorsch: Good point, the phrase "fingerprint" doesn't appear anywhere in the document now, so I'll work mention back in. However, the current implementation only supports `secure` and `fingerprint` types. Downgrading to `verify` would require editing the postfix template, rather than overriding a var. Trying not to bombard admins with config options, and stick to sane defaults.

If you think it's important for the security level to be easily overridden to any value on playbook runs, I can add support for that back in. If it's configurable, we should document it.

dolanjs: Since we know that `secure` will work with gmail and the `verify` option provides less protections for forged DNS requests, I do not think it is necessary for us to support the `verify` option.

Think our stance should be is that whatever smtp relay they pick, it needs to have the certificates provisioned correctly to pass the stricter `secure` validation or they will need to manage keeping the fingerprint value updated.

conorsch:
Added documentation of the `fingerprint` setting in a new section. Since fingerprint verification is opt-in and requires additional configuration, warrants its own section in the docs. The bulk of the explanatory text is the same as it was prior to this PR.

docs/ossec_alerts.md
garrettr: When I said "3 supported tiers", I meant:

1. `secure` with default CA's
2. `secure` with custom CA
3. `fingerprint` verification

I agree that we should not bother supporting the `verify` option. At the moment, I do not see any compelling reason to do so.

docs/ossec_alerts.md
garrettr: @conorsch I think you should expand this section to describe the 3 supported tiers of secure email configuration, and then reference or link to the fingerprint section below for people who are interested in using fingerprints.

Note: if you want to link to another section in the document based on the header, re-run `doctoc` and you can then use the anchor tag it generates :)

garrettr:
Final thought: should we support multiple fingerprints for the `fingerprint` option? Postfix supports it, and it might be useful for cases involving load balancers. Should be pretty easy, just let `smtp_relay_fingerprint` be a string *or* a list of strings, and update the configuration template logic accordingly.

conorsch:
The template is complex enough with the new conditional logic. Adding in more argument-parsing logic into the template would introduce more complexity, to accommodate for a brand new use case. If an admin wants to add support for multiple fingerprints, two changes are required:

### update `prod-specific.yml`
Declare `smtp_relay_fingerprint` as a list of strings, rather than just a string:
```
smtp_relay_fingerprint:
  - D3:7C:82:FC:D0:5F:8F:D7:DA:A2:59:8C:42:D7:B2:9F:C1:9F:7E:60
  - D3:7C:82:FC:D0:5F:8F:D7:DA:A2:59:8C:42:D7:B2:9F:C1:9F:7E:61
  - D3:7C:82:FC:D0:5F:8F:D7:DA:A2:59:8C:42:D7:B2:9F:C1:9F:7E:62
```

### update `main.cf` template
Change:

```
smtp_tls_fingerprint_cert_match = {{ smtp_relay_fingerprint }}
```
to:

```
smtp_tls_fingerprint_cert_match =
{% for fingerprint in smtp_relay_fingerprint %}
  {{ fingerprint }}
{% endfor %}
```

Adding magic sauce to handle both styles of formatting would be difficult to test reliably, and therefore not a sound addition right now. I've documented the manual workaround here, so we have a linkable solution in the event it's needed on an individual basis in the future.

garrettr:
@dolanjs Do you agree with @conorsch's position on not supporting multiple fingerprints in this comment? https://github.com/freedomofpress/securedrop/pull/1081#issuecomment-125281926

dolanjs:
@garrettr I think that is fine compromise. The default `secure` option is what everyone using smtp.gmail.com should use. Just supporting 1 fpr for orgs that are running their own smtp server I think is sufficient. Those same orgs probably also have the technical skills in house to manually modify the playbook to support multiple fpr in a load balanced situation. I just think the percentage or orgs that will need to support multiple fingerprint will be so small if it exists at all that we should not overly complicate the playbook to support all possible set ups.

garrettr:
@dolanjs Cool, thanks for explaining your opinion so clearly. I am happy with the compromise as well.

conorsch:
To be clear, I don't oppose supporting multiple fingerprints in general, but I do oppose introducing new and untested functionality while mitigating a known issue. Should a need for multiple fingerprints arise own the road, implementation should be discussed in its own issue.

garrettr:
@conorsch Agreed :+1: :+1:

docs/ossec_alerts.md
```diff
@@ -72,25 +73,40 @@ Next you specify the e-mail that you'll be sending alerts to, as `ossec_alert_em
...
 
-The Postfix configuration enforces certificate verification, and requires a fingerprint to be set. You can retrieve the fingerprint of your SMTP relay by running the command below (all on one line). Please note that you will need to replace `smtp.gmail.com` and `587` with the correct domain and port for your SMTP relay.
+The Postfix configuration enforces certificate verification, and requires both a valid certificate and STARTTLS support on the SMTP relay. By default the system CAs will be used for validating the relay certificate. If you need to provide a custom CA to perform the validation, copy the cert file to `install_files/ansible-base` add a new variable to `prod-specific.yml`:
 
-    openssl s_client -connect smtp.gmail.com:587 -starttls smtp < /dev/null 2>/dev/null | openssl x509 -fingerprint -noout -in /dev/stdin | cut -d'=' -f2
+    smtp_relay_cert_override_file: MyOrg.crt
```
garrettr: Should we use double quotes around the value?

garrettr:
The last thing we need to do is test and document the upgrade path from 0.3.4. There are two common use cases that we should focus on addressing:

1. The installation is using a 3rd party mail server (e.g. `smtp.gmail.com`). In this case, they should stop using `fingerprint` verification and start using the `secure` (with the system default CAs) verification supported by default in this PR. **Since this PR removes `smtp_relay_fingerprint`**, they should use the following procedure to update:

    1. Remove the `smtp_relay_fingerprint` line from `prod-specific.yml`.
    2. `git stash save "prod-specific settings"`
    3. `git pull` (to get the 0.3.5 changes)
    4. `git tag -v $(git describe)`
    5. `git checkout $(git describe)`
    6. `git stash pop`
    7. Run the playbooks to update the production servers.

2. The installation is using their own mail server. Their admins can be expected to monitor or be aware of updates to this mail server's TLS configuration, and will update the SecureDrop configuration (and re-run the playbooks) accordingly.

    In this case, they should continue using `fingerprint` verification. **Since this PR removes `smtp_relay_fingerprint`**, they should use the following procedure to update:

    1. Remove the `smtp_relay_fingerprint` line from `prod-specific.yml`, but keep it around (e.g. copy it to the clipboard).
    2. `git stash save "prod-specific settings"`
    3. `git pull` (to get the 0.3.5 changes)
    4. `git tag -v $(git describe)`
    5. `git checkout $(git describe)`
    6. `git stash pop`
    7. Manually re-add the `smtp_relay_fingerprint` line to `prod-specific.yml`.
    8. Run the playbooks to update the production servers.

    They may also want to consider using CA verification with a custom CA, depending on the details of their setup. This should be addressed on a case-by-case basis.

docs/ossec_alerts.md
conorsch: @garrettr Quoting isn't necessary in the vars file, but technically there could be problems with whitespace splitting in the postfix config on a live system if an admin provides a custom cert with a space in the filename (or in the directory name, if overridden). Addressed in 2ff45a3.

conorsch:
While testing:

> The installation is using a 3rd party mail server (e.g. smtp.gmail.com).

on hardware, popping the saved stash resulted in a merge conflict. I was able to prevent the merge conflict with the following revised workflow:

  1. Remove the `smtp_relay_fingerprint` line from `prod-specific.yml`.
  1.  `git stash save "prod-specific settings"`
    1. Now let's pretend we have a signed release:
    1. `git remote add conorsch https://github.com/conorsch/securedrop`
    1. `git checkout -b remove-smtp-fingerprint-requirement`
    1. `git pull conorsch remove-smtp-fingerprint-requirement`
  1.  `git checkout stash -- prod-specific.yml inventory`
  1. `git reset`
  1.   Run the playbooks to update the production servers.

Steps 3 & 4 differ from the usual "stash/pop" workflow for admins. Running the playbooks now, will follow up to confirm working notifications under new config.

conorsch:
Switching away from the hardcoded fingerprint to the `secure` config option broke the sending of notifications while testing on hardware. From `/var/log/mail.log`:

```
Jul 28 11:08:04 mon postfix/smtp[3866]: connect to smtp.gmail.com[2001:4860:400b:c01::6c]:587: Network is unreachable
Jul 28 11:08:05 mon postfix/smtp[3866]: certificate verification failed for smtp.gmail.com[74.125.20.109]:587: untrusted issuer /C=US/O=Equifax/OU=Equifax Secure Certificate Au
thority
Jul 28 11:08:05 mon postfix/smtp[3866]: 1A0B4101023: Server certificate not trusted
Jul 28 11:08:05 mon postfix/smtp[3866]: certificate verification failed for smtp.gmail.com[74.125.20.108]:587: untrusted issuer /C=US/O=Equifax/OU=Equifax Secure Certificate Au
thority
Jul 28 11:08:05 mon postfix/smtp[3866]: 1A0B4101023: to=<staff@freedom.press>, relay=smtp.gmail.com[74.125.20.108]:587, delay=1.3, delays=0.02/0.03/1.3/0, dsn=4.7.5, status=def
erred (Server certificate not trusted)
```

The IPv6 connection failure is likely caused by the default drop rules for ip6tables, added in #1053. Forcing IPv4 connections via `smtp_address_preference = ipv4` and `inet_protocols = ipv4` silenced those notifications, but of course didn't fix sending. To fix sending, I tried:

  * commenting out the `smtp_tls_CApath` declaration
  * removing quotes from the `smtp_tls_CApath` value (just in case)
  
Neither worked, even after reloading and restarting postfix. The SMTP troubleshooting tips added in this PR seemed to indicate that everything was OK.

```
root@mon:~# sudo -u postfix openssl s_client -showcerts -starttls smtp -connect smtp.gmail.com:587 < /dev/null 2> /dev/null | tail -n 2
    Verify return code: 20 (unable to get local issuer certificate)
---
root@mon:~# sudo -u postfix openssl s_client -CApath /etc/ssl/certs -showcerts -starttls smtp -connect smtp.gmail.com:587 < /dev/null 2> /dev/null | tail -n 2
    Verify return code: 0 (ok)
---
```
I did not try different security level settings such as `verify`, nor did I try overriding the `smtp_tls_secure_cert_match` default value of "nexthop, dot-nexthop", as that seemed a sane value.

ageis:
Just chiming in, but I think we should force IPv4 here for wider compatibility. I tend to use that same `inet_protocols` setting across a lot of my Postfix deployments since email over IPv6 breaks a lot more often, primarily via the DNS-based authentication mechanisms that mailservers perform like SPF and PTR record checks, which are not concerned here, but nonetheless I like it cause it's simpler and will make it easy to rule out IPv6 as an issue.

dolanjs:
@conorsch @ageis we should limit the postfix from trying to use IPv6, since we disable IPv6 thru the kernel sysctl and drop them with iptables rules... it wouldn't ever work anyway.

garrettr:
The double quotes around `smtp_tls_CApath` are causing the problem. When I remove them on my local testing instance, the certificate verifies and Postfix successfully sends the mails.

Best way to implement this is probably to revert 2ff45a3da62bd6be8ce46357baf9689b3f37ca82.

conorsch:
Thanks @garrettr, that was indeed the problem! I've reverted the offending commit. I'm also going to make the IPv4-only rules permanent.

conorsch:
These changes should be rebased and resubmitted targeting `release/0.3.5`.

Issue #1082, opened 2015-07-27
list spacing in install.md


harlo:
List formatting is slightly off:

![screen shot 2015-07-27 at 10 11 04 am](https://cloud.githubusercontent.com/assets/153357/8908067/e9260a38-3447-11e5-8d88-e7c33342af63.png)

PR #1083, opened 2015-07-27
fixed line spacing

harlo:
…as per https://github.com/freedomofpress/securedrop/issues/1082


conorsch:
Nice catch on the formatting, definitely an improvement. 

While we're here, do we want to update the logo size recommendation? We're currently telling folks that custom logos should be `500px x 450px`, but #1079 changed the default logo to `1000px x 1000px`. Probably fine as-is, but wanted to call attention to it.

harlo:
i'll open a new PR, @conorsch

conorsch:
:+1:

conorsch:
New PR is #1085, by the way.

garrettr:
Good catch, @harlo! However, `release/0.3.4` is not the right branch since we've already transitioned to that release. Can you close this PR and:

1. Re-submit with `develop` as the merge target?
2. Mark it with the 0.3.5 milestone (if you think this docs fix should be included in the next release, I think that would be a good idea) so we can remember to include it.

harlo:
yep

PR #1088, opened 2015-07-28
Chains cron-apt and reboot tasks in nightly cronjob

### LINKED ISSUE #1071: Prevent reboots during cron-apt upgrade
### Background
We use cron-apt, rather than unattended-upgrades, to manage overnight updates for the app and mon servers. When run via cron, cron-apt fuzzes the reboot logic by 0-60 minutes after invocation. The logic is on L318 in `/usr/sbin/cron-apt`:

```
        TIME=$(($RANDOM % $RUNSLEEP))
        sleep $TIME
        echo "CRON-APT SLEEP: $TIME, $(date)" >> "$INITLOG
```
The `RUNSLEEP` env var defaults to 3600 (see L244), thus the 0-60 minute delay. The SecureDrop Ansible playbooks configure cron-apt to run at 4AM local time, which means upgrades actually start anywhere between 4:00 and 5:00 local time. The Ansible playbooks also mandate a reboot at 5:00 local time, in order to clear memory (see #805). 

### The problem
If cron-apt runs at 4:59, and updates are available, then the upgrade will almost certainly be interrupted the 5:00 scheduled reboot. Such an interruption could leave apt or dpkg in a broken state (see possibly related #1032). The two cronjobs for cron-apt and reboot were originally set an hour apart to ensure the execution windows for the tasks did not overlap. It's easy to overlook the time-fuzzing logic, because if you run `cron-apt` from the command-line, it skips the fuzzing altogether:

```
if [ -t 0 ]; then
    RUNIMMEDIATELY="yes"
fi
if test "$RUNIMMEDIATELY" = "yes"; then
    RUNSLEEP=
fi
```

In short: if STDIN is a terminal, then we're running interactively, so don't bother sleeping. The reboot cronjob used to check for the presence of `/var/run/reboot-required`, as well, which *should* have ignored a reboot if cron-apt is still running, depending on when exactly the `/var/run/reboot-required` file is written by apt.

### Mitigations
There are a few ways to solve:
* Append the reboot task to the cron-apt task on the same cron line, so cron-apt blocks and the reboot runs after. (Choose wisely between `&&` and `;`.)
* Leave the cron-apt task on a separate line, and have it check for a running cron-apt process, sleep until cron-apt exits, then reboot.
* Move the cron-apt task up by an hour, to 3:00 local time, effectively creating a minimum window of 60 minutes between cron-apt starting and the reboot command. Technically this doesn't ensure the tasks won't overlap, but it would be a better implementation of the current config. 

The first option seems simplest and most reliable. The reboot task should be chained with `;`, rather than `&&`, to ensure that the server reboots even if cron-apt fails.

Thanks for pointing this out, @dolanjs.

conorsch:
Prevents nightly reboot from interrupting a running cron-apt task, as discussed in #1071. Closes #1071.

### DIFF
--- a/install_files/ansible-base/roles/common/files/cron-apt
+++ b/install_files/ansible-base/roles/common/files/cron-apt
@@ -2,11 +2,10 @@
 # Regular cron jobs for the cron-apt package
 #
 
-# Every night at 4 o'clock run cron-apt
-0 4 * * * root    /usr/bin/test -x /usr/sbin/cron-apt && /usr/sbin/cron-apt
-
-# Every night at 5 o'clock reboot (regardless of whether reboot is
-# required by the most recent package upgrade). This guarantees that
-# submission plaintext will be retained in memory for at most 24
-# hours.
-0 5 * * * root    /sbin/reboot
+# Every night at 4:00 AM run cron-apt, then reboot.
+# Note that cron-apt sleeps for 0-60 minutes when run
+# noninteractively (e.g. in this cronjob), so the actual
+# start time may be as late as 5:00 AM. Only reboot
+# if cron-apt exited successfully, so broken or partially
+# installed packages don't cause problems with booting.
+0 4 * * * root    /usr/bin/test -x /usr/sbin/cron-apt && /usr/sbin/cron-apt && /sbin/reboot


garrettr:
lgtm

garrettr:
Merged into release/0.3.5 in 0483ce3d93b935c1e8e0269a861fbd288fbe42dd.

PR #1095, opened 2015-08-04
Remove SMTP fingerprint requirement for OSSEC Postfix config (resubmitted)

### LINKED ISSUE #1076: Don't pin the leaf certificate in Postfix configuration

conorsch:
Rebased #1081 to target `release/0.3.5`, see that thread for full discussion. To recap:

### Major changes
 
* `smtp_tls_security_level = secure` is now default in Postfix config
* `smtp_relay_fingerprint` removed from `prod-specific.yml`
  (requires Admin intervention during upgrade)

### Action required
Removing the fingerprint from `prod-specific.yml` will require manual intervention by the Admin during upgrade. The usual stash/pop workflow results in a merge conflict, so use this procedure instead:
>
  1. Remove the `smtp_relay_fingerprint` line from `prod-specific.yml`.
  1.  `git stash save "prod-specific settings"`
    1. Now let's pretend we have a signed release:
    1. `git remote add conorsch https://github.com/conorsch/securedrop`
    1. `git checkout -b remove-smtp-fingerprint-requirement`
    1. `git pull conorsch remove-smtp-fingerprint-requirement`
  1.  `git checkout stash -- prod-specific.yml inventory`
  1. `git reset`
  1.   Run the playbooks to update the production servers.

Resolves #1076. Supersedes and therefore closes #1081.


conorsch:
Closing in favor of #1100, which rebases this PR to clean up redundant short titles in the commit history.

PR #1096, opened 2015-08-09
Ensures prod environment doesn't run in debug mode

ehartsuyker:


securedrop/journalist.py
```diff
@@ -101,6 +101,7 @@ def wrapper(*args, **kwargs):
 
 @app.route('/login', methods=('GET', 'POST'))
 def login():
+    raise Exception()
```
garrettr: Did you mean to leave this in, or was this just for testing?

ehartsuyker: Whoops. Yeah, that was just to check it. Removed.

garrettr:
This lgtm now! Thanks, this is a nice extra layer of assurance.

Issue #1092, opened 2015-07-30
Allow journalists to reset their own login credentials


garrettr:
Currently, it is possible for **admin** users to reset any other user's login credentials (password, 2fa). This is primarily useful for credential recovery, e.g. a journalist forgets their password, loses their two-factor device, or gets a new one.

I still advocate for keeping this feature in place as a disaster recovery mechanism. If a journalist cannot log in at all (lost password or 2fa device), then they will need an admin's assistance to regain access.

However, there is an additional use case where the journalist has not lost access, but wishes to switch to using a different two-factor device. One common example is where the journalist uses a TOTP app (like Google Authenticator) on their smartphone, and gets a new smart phone. For security reasons, these apps usually do not allow the two-factor secrets to be extracted or transferred without taking exceptional and potentially dangerous steps (e.g. rooting or jailbreaking the device's OS).

Another example would be a journalist who has been using a two-factor app on their phone, but wishes to switch to a two-factor hardware token (e.g. a Yubikey).

These use cases are all possible now with an admin's assistance, but this is cumbersome and may be difficult or impossible to arrange if the journalist and the admin are not physically co-located.

We should add support for journalists to reset or update their own credentials, while maintaining admin's ability to reset all user's credentials in case they lose access to their account.

Related to #1091.

garrettr:
Another use case that came up during a conversation with Tom Lowenthal - an admin might want to provision accounts for journalists (perhaps they are not geographically co-located), send them the info, but then request that they change their password to something that only they know to prevent the admin from accessing their account.

This is an extremely common account provisioning workflow among other tools, and it makes sense that we should support it.

PR #1097, opened 2015-08-18
Edit creds

taravancil:
Takes care of #1092

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -321,11 +321,12 @@ def edit_account():
                 "Password successfully changed!",
                 "notification")
         except Exception as e:
-            db_session.rollback()
             flash(
                 "An unknown error occurred, please inform your administrator",
                 "error")
-    
+            app.logger.error("Password change for '{}' failed: {}".format(
+                user, e))
+            db_session.rollback()
     return render_template('edit_account.html')
 
 
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -219,7 +219,6 @@ def admin_new_user_two_factor():
 
     return render_template("admin_new_user_two_factor.html", user=user)
 
-
 @app.route('/admin/reset-2fa-totp', methods=['POST'])
 @admin_required
 def admin_reset_two_factor_totp():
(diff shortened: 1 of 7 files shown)


securedrop/journalist.py
```diff
@@ -292,6 +296,80 @@ def admin_delete_user(user_id):
...
+        try:
+            db_session.add(user)
+            db_session.commit()
+            flash(
+                "Password successfully changed!",
+                "notification")
+        except Exception as e:
+            db_session.rollback()
+            flash(
+                "An unknown error occurred, please inform your administrator",
```
garrettr: In order for the administrator to be able to respond effectively, they need some info about what happened. For now, can you add a call to `app.logger.error` with the contents of `e` so they'll at least have something to go off of?

In general, this exception handling needs an overhaul but at least logging the error is a start that will aid future debugging and incident response.

securedrop/journalist_templates/base.html
```diff
@@ -17,6 +17,8 @@
     <div id="logout">
       {% if g.user and g.user.is_admin %}
       <a href="{{ url_for('admin_index') }}">Admin</a> | 
+      {% else %}
+      <a href="{{ url_for('edit_account') }}">Edit Account</a>
```
garrettr: Add a `|` after "Edit Account" so there's a visual separator between it and "Logout".

garrettr: Also, do we want to show "Edit account" on admin's pages too? They can always edit their own info through the Admin Interface, but they might appreciate a shortcut to their own settings via "Edit account" as well.

securedrop/journalist_templates/edit_account.html
```diff
@@ -0,0 +1,32 @@
+{% extends "base.html" %}
+{% block body %}
+<h2>Edit your account</h2>
+
+<form method="post">
+  <input name="csrf_token" type="hidden" value="{{ csrf_token() }}"/>
+  <p>
+    <label for="password">Change password</label>
+    <input name="password" id="password" type="password" />
+    <em class="light">Leave blank for no change</em>
```
garrettr: This is no longer necessary because the only thing you're allowing them to change is their password. This *was* necessary (albeit arguably not a great UX) on the admin's pages because they could also change the user's username and toggle `is_admin`, so there needed to be a way to signal "I'm just changing the username and/or the is_admin toggle, and I don't want to reset their password to an empty password at the same time".

garrettr:
@taravancil Beyond those line comments, this looks great! I've tested everything and encountered no issues. Thanks for writing comprehensive passing unit tests as well!

taravancil:
No problem! I'll work on those changes now.

PR #1099, opened 2015-08-24
doc update with better suggestion for build-deps

chadmiller:
Single .deb downloads don't get security updates, so if the OS has a recent
vagrant version, we should prefer it and suggest developers use it. Tell them
how to find out from APT what would be installed, so they can compare.

### DIFF
--- a/docs/develop.md
+++ b/docs/develop.md
@@ -38,14 +38,21 @@ We recommend using the latest stable version of Vagrant (`1.7.4` at the time of
 this writing), which might be newer than what is in the Ubuntu repositories.
 
 If `apt-cache policy vagrant` says your candidate version is not at least
-1.4, you should download the current version from
-https://www.vagrantup.com/downloads.html. We *do not* recommend using an older
-version of Vagrant from Ubuntu's package repositories. E.g, version 1.5.4
+1.7, you should download the current version from
+https://www.vagrantup.com/downloads.html and then install it.
+
+```sh
+sudo dpkg -i vagrant.deb    # if you downloaded vagrant.deb from vagrantup.com
+
+sudo apt-get install vagrant    # OR this, if your OS vagrant is recent enough
+```
+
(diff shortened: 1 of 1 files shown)


garrettr:
Great point @chadmiller, thanks for helping clarify the best way to get Vagrant on Linux (so many options, oh my!).

My only question is - have you tested running the SecureDrop VM's with Vagrant 1.7.4? I'm currently using 1.7.2 with no problems. The only reason I ask is that we've encountered issues in the past where upgrading Vagrant lead to breakage (see #802 for one example).

conorsch:
For what it's worth, I've been running 1.7.3 with no problems, and it looks like 1.7.4 was a minor bugfix release [just a week later](https://github.com/mitchellh/vagrant/blob/master/CHANGELOG.md#174-july-17-2015). Just now I upgraded to 1.7.4 and I'll ride this out for a bit and report back here.

garrettr:
I'm also upgrading to 1.7.4 to make sure there are no problems before merging this.

garrettr:
@chadmiller Sorry for the delay on this! I've been using Vagrant 1.7.4 with SecureDrop for over a month now without issues, so your advice in this PR is good and will not cause any problems AFAICT.

Unfortunately, while I was testing 1.7.4 to make sure it wouldn't cause any problems with our development environment, I was also working on a complete re-working on the SecureDrop documentation, which we merged in #1158. That work has totally rotted this PR, since we converted the documentation from Markdown to ReST and also moved `docs/develop.md` to `docs/development/getting_started.rst`. I'm sorry! :sob:

I'd still like to include the substance of this PR into our new developer documentation. Are you willing to rebase this pull request (or submit a new one, if that's easier for you) on top of the latest docs in the develop branch? If not, just let me know and I'll handle it.

garrettr:
@chadmiller Since you haven't replied, I'm assuming you're ok with me rebasing this on top of the new documentation.

PR #1101, opened 2015-09-02
Update version fixes

garrettr:
Fix some minor issues in the `update_version.sh` script that were discovered while testing the 0.3.5 release.

* The OSSEC version number for the `securedrop-ossec-*` packages was hardcoded, which broke because we've upgrades OSSEC since I last ran `update_version.sh`. I changed the sed regex to be more generic so it should update all version-y looking strings correctly (the format is `$OSSEC_VERSION+$SECUREDROP_VERSION`).
* A previous refactor broke the final "success" echo statements because it got rid of `$TAGFILE`. I added it back in and cleaned up the message a bit.


garrettr:
- [ ] Merge into release/0.3.5
- [ ] Merge into develop

conorsch:
:+1:

garrettr:
Merged into develop in ae783a7d4db68092a10f9b989c45f566e66867f2.

PR #1102, opened 2015-09-02
Extend mail queue lifetime to 14 days

ageis:
Default `maximal_queue_lifetime` is 5 days. The idea here is if SMTP breaks for any reason longer than 5 days and shorter than 2 weeks (ex. change of certificate or fingerprint or credentials, or mail server outage) OSSEC alerts that may contain important info can still get delivered to the admin instead of purged from the deferred queue. Such alerts will still be in OSSEC's alerts.log of course but since that hardly ever gets checked I think we should widen the window for admins to receive them via e-mail and allow for the possibility that SMTP functionality could break for a period longer than 5 days. Open for discussion.

### DIFF
--- a/install_files/ansible-base/roles/ossec_server/templates/main.cf
+++ b/install_files/ansible-base/roles/ossec_server/templates/main.cf
@@ -57,3 +57,4 @@ mydestination = $myhostname, localhost.localdomain , localhost
 mynetworks = 127.0.0.0/8 [::ffff:127.0.0.0]/104 [::1]/128
 mailbox_size_limit = 0
 recipient_delimiter = +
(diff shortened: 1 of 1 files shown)


garrettr:
Ideally we'd never been in a situation where SMTP would go down like that, and this will probably become a moot point eventually as we transition away from OSSEC in future releases. For now, this is a fine change to make for a little extra assurance.

garrettr:
Note: this went into develop, not release/0.3.5. I'm not merging it into 0.3.5 because we're fixing the SMTP fingerprint problem in that release, and this change (while minor, and likely not to cause problems) has not been adequately tested for a release.

Issue #1072, opened 2015-07-18
Update firewall docs


conorsch:
The firewall appliance recommended in `docs/hardware.md` is no longer available. From the [Netgate APU2 page](http://store.netgate.com/NetgateAPU2.aspx):

> The APU2 system is no longer available.
The APU1D board and kit are available by special order only. Please email for availability. 

It's ambiguous whether the [APU4 unit](http://store.netgate.com/APU4.aspx) will become available again:

>OUT OF STOCK.
The APU4 equivalent is the VK-T40E directly available from the pfSense microsite.

Then, on the [VK-T40E page](http://store.pfsense.org/VK-T40E/):

>The VK-T40E has been replaced. Please select one of these fanless replacement options:
SG-2220: Replacement for the VK-T40E with dual core Intel C2338 CPU, 2 GB RAM, 2 Intel GbE Ports.
SG-2440: Upgrade with dual core Intel C2358 CPU, 4 GB RAM, 4 Intel GbE Ports.

The [SG-2220](http://store.pfsense.org/SG2220/) only has two ports, WAN and LAN, so that's out.  The [SG-2440](http://store.pfsense.org/SG2440/), however, has four ports (WAN, LAN, OPT1, and OPT2) which could accommodate the Admin Workstation and remove the need for a switch. It's $500, rather than ~$300 for the APUs.

We should select a new recommended unit and update the docs.

Thanks to @Taipo for pointing this out (in the comments in #732).

garrettr:
We got an SG-2440 delivered today and will be testing it (and updating the docs) this week.

conorsch:
Waiting for release candidate for 0.3.5, will run through test install with new hardware then.

conorsch:
The pfSense Guide URL in the docs is broken: http://data.sfb.bg.ac.rs/sftp/bojan.radic/Knjige/Guide_pfsense.pdf Consider linking to pfSense project wiki instead: https://doc.pfsense.org/index.php/Main_Page

garrettr:
@conorsch That's unfortunate, the linked guide was really excellent. I'm pretty sure it was a copy of "pfSense: The Definitive Guide", which, according to the pfSense project wiki, is meant to be only available to Gold Subscribers. Which is unfortunate, as I found it to be much better than any of the online resources, especially for pfSense newbies.

Some options for resolving that particular issue are:

1. Encourage SecureDrop users to become pfSense Gold subscribers if they want to get access to that doc.
[...]

PR #1104, opened 2015-09-04
update firewall docs for SG-2440

harlo:
updating firewall doc for the new pfsense box, as per https://github.com/freedomofpress/securedrop/issues/1072.  NB: line 195 references a firewall rule template (at `install_files/network_firewall/rules`) which I suppose should be modified to fit this new configuration as well.


garrettr:
@harlo These docs looks good! I will test them with a fresh install later today.

garrettr:
@harlo The firewall rules for DNS need to support `TCP/UDP` for the protocol, since [DNS may use TCP][1] is either the request or response is larger than the size of a single packet.

[1]: http://www.networkworld.com/article/2231682/cisco-subnet/cisco-subnet-allow-both-tcp-and-udp-port-53-to-your-dns-servers.html

conorsch:
@harlo New screenshots look great! We're now correctly recommending `TCP/UDP` for DNS traffic. Looks ready for merge now, @garrettr.

garrettr:
Great, I'm going to run through an install today using these updated docs as a final check, will then merge.

harlo:
closing in favor of #1106

PR #1105, opened 2015-09-08
Confirmation prompts

### LINKED ISSUE #560: Require confirmation for sources deleting replies
As a source, I expect the “Delete” button to require a confirmation prompt, because SOP on online sites is that any destructive action should require a confirmation prompt.

This is difficult for us because of our "no javascript on the source interface" policy. We may be able to do it with CSS.

taravancil:
This adds a confirmation prompt to deleting replies in the source interface and changing a codename in the journalist interface.

Fixes #560

### DIFF
--- a/securedrop/journalist_templates/col.html
+++ b/securedrop/journalist_templates/col.html
@@ -10,7 +10,11 @@
       <a href="/">All Sources</a>
       <i class="fa fa-chevron-right"></i>
       <strong>{{ source.journalist_designation }}</strong>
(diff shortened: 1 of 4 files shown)


garrettr:
@taravancil First of all, sorry for the delay in reviewing this. I was feeling under the weather all week and let these PR's get backed.

Second of all, this is a *fantastic* PR! I just tested it on my development VM and could find nothing wrong with it. I love that it's CSS-only, because that's a requirement on the Source Interface. It's cool that you made it CSS-only on the Document Interface too! The styling fits in nicely with the current UI.

I did some preliminary QA stuff:

* testing deletion with multiple replies on the source interface
* testing refreshes at various times mid-way through the prompt process

...and encountered no issues. This is good to merge!

P.S. today's my birthday and this might be the best gift I've received so far! Thanks :birthday: :grinning:

taravancil:
I'm glad you like it! I felt it made sense to stick with the CSS-only prompt since it's only a few lines of CSS and it works like a charm.

I hope you're feeling better and that you have a great birthday! :tada:

Issue #1106, opened 2015-09-11
separate sections in firewall docs for OPT1+OPT2 and OPT1 only configs


harlo:
so admins can follow along according to the hardware they have...

conorsch:
Resolved by #1113.

PR #1108, opened 2015-09-14
3- and 4-NIC instructions

harlo:
including configuration options for 3- and 4-NIC configurations where applicable, as per https://github.com/freedomofpress/securedrop/issues/1106


conorsch:
Huge improvement! In the future, hopefully we can standardize on the 4-port model. For now, though, plenty of folks have the 3-port already, and only the 4-port is available for purchase, so providing instructions for both is the way to go.

Already targeting `release/0.3.5`, too. lgtm! :+1:

garrettr:
Only thing you're going to want to do here is re-build the table of contents to take into account all the new headers you added. Managing the TOC should be handled automatically once we move to a more sophisticated system for our docs, but for now I've been using [doctoc](https://github.com/thlorenz/doctoc) (you can use the [web service](https://doctoc.herokuapp.com/) if you don't want to install it via `npm`).

garrettr:
Since the firewall rules have changed, we also need to update the iptables definitions of those rules in `install_files/network_firewall/rules`. These rules are not used directly (since the firewall is pfSense, not iptables, and is configured manually, not automatically), but they are intended to be a well-defined version of the network firewall rules.

I have a feeling that most people probably don't bother to read this file, and instead are just using the screenshots of the pfSense WebGUI included in this document.

@harlo What do you think? Should we update the `rules` file, or delete it and just have the screenshots be the canonical definition of the network firewall rules?

harlo:
I actually flagged this here: https://github.com/freedomofpress/securedrop/pull/1104.  Final decision up to you, though!  (what do you think, @conorsch?)

garrettr:
Another note: my understanding was that NTP is unusual in that it runs on port 123 on *both* the client and the server. However, I'm tailing the network firewall logs during the initial install of Ubuntu on the application server and seeing a lot of denies for UDP packets from random ports on the application server sent to 123 on an Ubuntu NTP server.

I will update the screenshots to reflect the fact that NTP should be allowed to originate from any port.

garrettr:
Updated the screenshots to fix the NTP rules on my branch.

conorsch:
Great catch on the NTP protocols, @garrettr! Regarding the `rules` file, I say we update it—assuming we can trivially export it from the pfSense GUI. I'm also skeptical about the practical utility of that file, but it serves as strong documentation for technical admins, it's greppable unlike the images, and updating it also leaves the door open to virtualizing the firewall down the line for testing and development.

garrettr:
@conorsch The problem with the `rules` file is that it's *not* exportable from pfSense (pfsense != iptables). It was a solution that James came up with a while back to describe the firewall rules in a well-specified way back when people were trying to use a variety of firewalls with SecureDrop (e.g., we'd do an installation where they wanted to use the Cisco firewall they already had lying around rather than buy a pfSense appliance). It was completely created and is totally updated by hand, and cannot be used directly by any network firewall appliance that I know of (iptables is usually used for host firewalls, not for network firewalls).

conorsch:
Drat. Let's remove it, then. It's out of date, and inaccurate docs are worse than no docs. We can revisit down the road if have virtualization of the firewall, since then we can trivially dump a list of the rules.

garrettr:
Given that we've pretty much standardized on pfSense at this point, I'm in favor of getting rid of the rules file and using the screenshots for now. Alternatively, I might be able to export something usable from the pfSense WebGUI to *replace* the firewall rules. I think it exports some XML thing.

Eventually, of course, we'll figure out how to automate this with Ansible and all of this will be a thing of the past :smile:

conorsch:
Superseded by #1113.

PR #1111, opened 2015-09-15
Make delete confirmation unclickable when hidden

### LINKED ISSUE #1110: Delete confirmation can be triggered accidentally
The new confirmations in #1105 work by adding hidden links that have their `display:` property toggled via CSS pseudo-selectors. However, even when they are hidden they are still clickable. For example, on the source's `/lookup` page:

hidden delete button for replies

As you can see, when you mouse over the area where the delete button is "hidden" (not yet activated by clicking the visible "Delete" button with the trashcan icon), the cursor turns into a pointer, indicating a clickable element. Clicking in this apparently empty space causes the reply to be deleted.

All of the confirmation buttons in #1105 should not be clickable unless they are visible. This is probably achievable in all of the browsers we care about via CSS' [pointer-events](http://stackoverflow.com/a/10276157/1093000).

taravancil:
Fixes #1110

### DIFF
--- a/securedrop/static/css/journalist.css
+++ b/securedrop/static/css/journalist.css
@@ -220,11 +220,11 @@ button.small, a.btn.small, .btn.small {
 }
 
 .confirm-prompt {
-  display: none;
+  visibility: hidden;
 }
 
 .confirm-prompt:target {
-  display: block;
+  visibility: visible;
 }
 
 p.serious {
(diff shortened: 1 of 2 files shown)


garrettr:
How about instead of using opacity and z-index, we just use `visibility: hidden` and `visibility: visible`? I just tested it and it works (and does not allow clicks when the element is hidden). That seems simpler and less hacky.

taravancil:
Yeah, that is simpler. I'll make both of the confirmation prompts use the `visibility` property for the sake of consistency.

garrettr:
@taravancil :+1: for consistency

garrettr:
Squashed and merged in 0e05fd50fcc4c3aaa9f9bd9605c6366e8c7a2c85.

PR #1113, opened 2015-09-17
Update Network Firewall docs

garrettr:
Updated Network firewall docs, fully tested with a 0.3.5 release candidate on hardware.

This incorporates @harlo's commits from #1108, which documents *both* the old 3-NIC configuration (based on the Netgate APU 2) and the new 4-NIC configuration (based on the SG-2440). In addition, it:

* Clarifies a lot of wording based on doing a test install and following these docs word for word
* Removes references to the pfSense Guide because the link to the PDF broke
* Provides recommendations and documents the process for keeping pfSense up to date
* Recommends adding DNS servers to the Manual networking configuration on the Admin Workstation to avoid problems with launching the Unsafe Browser in subsequent sessions.
* Fixes the NTP firewall rules, which were incorrectly restricting the source port to 123


conorsch:
Excellent rewrite. Axing the outdated iptables example for a redacted XML dump of the pfSense config is a big improvement. I'm reminded that our Ansible playbooks don't currently support configuration of secondary DNS servers—I'll open an issue.

PR #1130, opened 2015-09-24
Update spectests for 0.3.5

conorsch:
Brings the serverspec tests up to par with the current state of `develop`, now that `0.3.5` has shipped.

The changes are mostly cosmetic, but c5c468f points out a problem with upgrading to the newest version of Vagrant in Snap CI—there's a dependency conflict with versions of `bundler`. The long-term solution is to strictly enforce the version of `bundler` in Snap CI, but for now, sticking with Vagrant v1.7.2 in Snap.

Note that #1099 recommends updating the version recommendation for devs to Vagrant v1.7.4. The issue described in c5c468f is **not** a blocker to merging #1099, since the Serverspec docs recommend installing bundler via `apt-get`, which installs a matching version that Vagrant is happy with.

In a separate PR I'll work on resolving the bundler version mismatch in Snap.

### DIFF
--- a/snap_ci/pinned.sh
+++ b/snap_ci/pinned.sh
@@ -23,7 +23,7 @@ vagrant_version="${vagrant_version:-1.7.2}"
 vagrant_rpm="vagrant_${vagrant_version}_x86_64.rpm"
 vagrant_url="https://dl.bintray.com/mitchellh/vagrant/${vagrant_rpm}"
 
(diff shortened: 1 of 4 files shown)


garrettr:
@conorsch lgtm, is this ready to merge?

conorsch:
These changes landed as part of #1196, due to a rebase. Closing.

PR #1148, opened 2015-10-16
Update install.md

mr-grooms:
I found some misspellings.


garrettr:
Thanks for these fixes!

PR #805, opened 2014-12-13
Reboot every 24 hours to periodically wipe memory

garrettr:
This is a kludge, but a reasonably effective one. This change reboots
the servers every day at 5am. Previously this reboot would only happen
if required by the daily package upgrade (e.g. for a kernel upgrade),
but now it happens every day regardless of whether it is needed.

This ensures that the plaintext of submissions can reside in memory for
at most 24 hours after submission. Resolves #99, although we can and
should do better in a future release.

PR #1154, opened 2015-10-20
Fix PR #805 link in thread_model.md

darkk:


garrettr:
Thanks for catching this!

Issue #520, opened 2014-08-04
Multiple security issues with the gpg --recv-key command in install.md


diracdeltas:
`docs/install.md` instructs the admin to fetch the securedrop code signing key with the command `gpg --recv-key BD67D096`. Problems that I can see:
1. Short key IDs are insecure since generating collisions is trivial. See http://www.asheesh.org/note/debian/short-key-ids-are-bad-news.html. Easy to fix by using the full key ID `9092 EDF6 244A 1603 DCFD  C462 9A2B E67F BD67 D096`
2. However, even using long key IDs, `gpg --recv-key` has an open bug where GPG won't verify that the key ID of the received key matches the key ID that was requested (!!!). Here's the open upstream ticket (warning: self-signed cert): https://bugs.g10code.com/gnupg/issue1579.
3. On top of all this, gpg isn't fetching keys over SSL here (unless you've installed gnupg-curl or set up gpg2, and specify a SSL-only keyserver or keyserver pool - I use hkps://pgp.mit.edu ever since I convinced the person running it to support SSL). 

Impact:
At least fetching the package and signature happens over SSL. However, given this bug, code signing is useless against an active MITM with no special capabilities.

Mitigation:
Use the long key ID. Make sure the key is received over SSL, though that doesn't fix the second problem if the keyserver is malicious. Make the admin manually check that the long key ID of the received key matches the requested key ID until the upstream bug is fixed.

fowlslegs:
Funny, agl's issue was fixed the day after this issue was closed. It's been 14 months, so I'm sure changes have propogated to the gnupg binary in Tails. It should no longer be necessary to check the fingerprint separately when GPG now implicitly does so when fetching a key using its fingerprint; I will go ahead and remove this from install.md.

[...]

PR #1157, opened 2015-10-23
Use hkps SKS pool, remove redundant fingerprint verification

fowlslegs:
Tails ships with the SKS hkps pool certificate and is set to use the hkps pool
by default. Passing `--keyserver pool.sks-keyservers.net` opens the admin up to
MitM attacks, and so I have removed it.

Since August 2014, GPG's `--recv-key` will check that the fingerprint of the
received key matches the fingerprint passed on the commandline. Thus, it is no
longer necessary to separately run `gpg --fingerprint`.

This fixes the problems described in issue #520.

Signed-off-by: Noah Vesely <fowlslegs@riseup.net>

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -241,12 +241,11 @@ git clone https://github.com/freedomofpress/securedrop.git
 
 Before proceeding, verify the signed git tag for this release.
 
-First, download the *Freedom of the Press Foundation Master Signing Key* and verify the fingerprint.
+First, download the *Freedom of the Press Foundation Master Signing Key*. When the full fingerprint is passed to GPG's `--recv-key` option, GPG will implicitly check that the fingerprint of the key received matches.
 
-    gpg --keyserver pool.sks-keyservers.net --recv-key B89A29DB2128160B8E4B1B4CBADDE0C7FC9F6818
-    gpg --fingerprint B89A29DB2128160B8E4B1B4CBADDE0C7FC9F6818
+    gpg --recv-key B89A29DB2128160B8E4B1B4CBADDE0C7FC9F6818
 
-The Freedom of the Press Foundation Master Signing Key should have a fingerprint of "B89A 29DB 2128 160B 8E4B  1B4C BADD E0C7 FC9F 6818". If the fingerprint does not match, fingerprint verification has failed and you *should not* proceed with the installation. If this happens, please contact us at securedrop@freedom.press.
+If GPG informs you the fingerprint of the key received does not match, fingerprint verification has failed and you *should not* proceed with the installation. If this happens, please contact us at securedrop@freedom.press.
 
 Verify that the current release tag was signed with the master signing key.
 


garrettr:
> Since August 2014, GPG's --recv-key will check that the fingerprint of the received key matches the fingerprint passed on the commandline.

It's important to verify claims like this, especially in the land of Debian/Tails. This bug was fixed in gpg 1.14.16, but the current version in Tails (based on Debian Wheezy) is [1.14.12](https://packages.debian.org/wheezy/gnupg). However, this specific fix was backported into the version of GPG in Debian Wheezy (1.4.12-7+deb7u5, see [changelog](http://metadata.ftp-master.debian.org/changelogs//main/g/gnupg/gnupg_1.4.12-7+deb7u7_changelog)) so this change is still appropriate.

garrettr:
Two things about merging this PR:

1. I'm about to complete and merge a total overhaul of the docs. The overhaul includes transitioning from Markdown to ReST, so this change won't merge once the overhaul is done. You can see the changes in my [read-the-docs](https://github.com/garrettr/securedrop/tree/read-the-docs) branch.

    I'd just wait until read-the-docs is merged, since that will happen in the next few days, and then re-do this PR for the new ReST docs. Should be easy. Sorry about that!

2. If you're going to simplify the "fetch a gpg key from the keyservers" flow in the install documentation, you might as well do it across all of the docs for consistency. A quick ` grep -rn "\-\-recv-key" docs` reveals several other instances that should be changed.

garrettr:
I just merged my read-the-docs branch in #1158, so this should be rebased on top of the new docs in the latest develop, and updated to use ReST instead of Markdown syntax.

Since the filenames changes to (`install.md` -> `install.rst`), it's probably easier just to close this PR and redo it based on the latest develop.

garrettr:
@fowlslegs Are you planning to rebase this on top of the new documentation in develop? Otherwise, I'm going to close without merging.

fowlslegs:
I have another pull request ready. Sorry I haven't had the time to get anything done on this front in the last week.

fowlslegs:
So I haven't made the pull request yet because:

(i) I'm waiting to hear from @ioerror regarding questions about the SKS keyserver hidden service he uses for TorBirdy (meant to CC securedrop@freedom.press, but will fwd when I get a response).

(ii) In ubuntu_install.rst we do not specify that any particular OS should be used to download and verify the Ubuntu image (and it in fact includes instructions for OS X), whereas in set_up_tails_admin.rst, it is specified that all GPG actions are to be performed within Tails. I verified the oldest extended-support distros people might be running--Redhat and CentOS versions 6--have a patched version of GPG (and Wheezy has backported the fix). And GPGTools for Mac is using GPG 2.0.29. With this in mind I think it's safe to remove the separate fingerprint verification step from ubuntu_install.rst as well. I wanted to check with you on this first, however.

garrettr:
> With this in mind I think it's safe to remove the separate fingerprint verification step from ubuntu_install.rst as well. I wanted to check with you on this first, however.

Sgtm, thanks for doing the background research to make sure that's safe.

fowlslegs:
So I'm going to propose making another change in this upcoming commit as well. Going w/ line 220 [here](https://github.com/ioerror/torbirdy/blob/19cc1c71b9ee44554f3ff5ac6006da1380bd3a95/components/torbirdy.js), we have some confirmation that this hidden keyserver is to be trusted/ is part of the normal SKS network. I would like to use it in both ubuntu_install and set_up_tails_admin, but that requires we instruct admins to download Ubuntu over Tails. I was going to do this in another commit for granularity, but since these changes complement each other and are minor, I think we could do it in one.

Besides the fact we can use this hidden keyserver (=> better authentication than TLS, no metadata exposure--this is due to Tor not the use of HS), there are two other benefits to downloading Ubuntu over Tor.

First, if someone would like to anonymously host a SD instance, downloading Tails and Ubuntu Server 14.04 over clearnet in the same time frame is a fairly "fingerprintable" action, while downloading Tails alone is not so much. Ideally, they'd download Tails over Tor as well, but adding that to the docs is another discussion.

Second, when the Tails image is first verified and dd'ed--we'll call this t=0--the integrity of the Tails image is dependent on the integrity of the AW. Assume that both are uncompromised at t=0. Then, as t :arrow_forward: oo, it is much more likely that just the AW is compromised rather than both the AW and Tails (persistently) or just Tails (persistently). Thus, when it comes time to upgrade to Ubuntu Server 16.04, there is a potential security (not just anonymity) benefit to downloading Ubuntu from Tails.

PR #1158, opened 2015-10-27
Read the docs

garrettr:
This is a big change! I migrated all of the documentation in `docs/` from Markdown to Restructured Text (ReST), and we are now using Sphinx to build it. The new layout is defined in `index.rst` and is intended to work well with the Read the Docs (RTD) document hosting platform.

This is a thorough migration and revision of a lot of the documentation, but it is by no means complete. At this point, I believe we should merge it into master so we can continue to improve the documentation with the new formatting from ReST and structure from Sphinx/RTD, because they provide numerous benefits that help make the documentation more readable, navigable, and well-organized than what was previously possible with a bunch of Markdown files on Github.

This diff looks huge, but that's largely due to the auto-conversion of Markdown to ReST and the deletion of the original Markdown files. I also made a lot of changes, and tried as much as possible to put them into focused commits with clear commit messages.

If you are interested in how to work with the new documentation (including building it locally), see the new "Documentation Guidelines" document under the "Contributing" toctree (`development/documentation.rst`).

See the documentation in this pull request rendered on Read the Docs: https://securedrop.readthedocs.org


conorsch:
Massively awesome. Easier to read, easier to search, easier to edit. Great work! :metal:

PR #1160, opened 2015-10-28
Documentation nits and bumping Ansible version number

ageis:
Just a few very tiny things I noticed last week and tonight.

### DIFF
--- a/docs/development/getting_started.rst
+++ b/docs/development/getting_started.rst
@@ -145,7 +145,7 @@ development, staging, and prod (production).
    hardening applied but virtualized
 
 If you plan to alter the configuration of any of these machines, make sure to
-review the `Development Guide for Serverspec Tests </docs/development/spec_tests.rst>`__.
+review the :doc:`Development Guide for Serverspec Tests <spec_tests>`.
 
 Development
 ~~~~~~~~~~~
--- a/docs/development/getting_started.rst
+++ b/docs/development/getting_started.rst
@@ -145,7 +145,7 @@ development, staging, and prod (production).
    hardening applied but virtualized
 
 If you plan to alter the configuration of any of these machines, make sure to
-review the `Development Guide for Serverspec Tests </docs/spec_tests.md>`__.
+review the `Development Guide for Serverspec Tests </docs/development/spec_tests.rst>`__.
 
 Development
 ~~~~~~~~~~~
--- a/docs/development/getting_started.rst
+++ b/docs/development/getting_started.rst
@@ -6,11 +6,11 @@ Prerequisites
 
 SecureDrop is a multi-machine design. To make development and testing easy, we
 provide a set of virtual environments, each tailored for a specific type of
-development task. We use Vagrant and Virtualbox to conveniently develop with a
+development task. We use Vagrant and VirtualBox to conveniently develop with a
 set of virtual environments, and our Ansible playbooks can provison these
 environments on either virtual machines or physical hardware.
 
-To get started, you will need to install Vagrant, Virtualbox, and Ansible on
+To get started, you will need to install Vagrant, VirtualBox, and Ansible on
 your development workstation.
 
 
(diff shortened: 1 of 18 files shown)


docs/development/getting_started.rst
```diff
@@ -51,7 +51,7 @@ If you're using Ubuntu, you can install a sufficiently recent version of
 Ansible from backports (if you prefer): ``sudo apt-get install
 ansible/trusty-backports``
 
-.. note:: Tested: Ansible 1.8.4
+.. note:: Tested: Ansible 1.9.4
```
garrettr: Have you actually tested these documentation updates with Ansible 1.9.4?

ageis: I've spun up the staging environment successfully using Ansible 1.9.4. That's what I was running when I found the expired key, and later got through the whole playbook. Is that sufficient?

garrettr: Probably, yeah.

docs/development/getting_started.rst
```diff
@@ -145,7 +145,7 @@ development, staging, and prod (production).
    hardening applied but virtualized
 
 If you plan to alter the configuration of any of these machines, make sure to
-review the `Development Guide for Serverspec Tests </docs/spec_tests.md>`__.
+review the `Development Guide for Serverspec Tests </docs/development/spec_tests.rst>`__.
```
garrettr: Nope, this doesn't work. Did you even try building this locally to make sure that it worked?

First of all, when Sphinx renders this documentation, it creates **HTML** (`.html`) from ReStructuredText (`.rst`). So this link ending in `.rst` is bogus, and cannot be reasonably expected to return anything but a 404.

Historical background: before the switch to RTD, we relied on GitHub's automatic rendering of Markdown files to have presentable documentation for end users. As a result, we started using relative and root-relative links to achieve inter-documentation hyperlinking. This was a gross hack that was totally dependent on GitHub's (fortunately stable and consistent) URL scheme.

Now that we're using Sphinx (the builder underpinning RTD), we can *explicitly* refer to other parts of the documentation, instead of resorting to gross (*implicit*) hacks. I have already started doing this in some of the documentation that I migrated. You should use the [`:doc:`](http://sphinx-doc.org/markup/inline.html#role-doc) role to refer to another document, or the [`:ref:`](http://sphinx-doc.org/markup/inline.html#role-ref) role to refer to a section of another document.

docs/development/threat_model.rst
```diff
@@ -156,18 +156,18 @@ What the Workstations can achieve
 
 -  The **Admin Workstation** requires Tails with a persistent volume,
    which stores information such as GPG and SSH keys, as well as a
-   `database with passphrases </docs/install.md#passphrase-database>`__
+   `database with passphrases </docs/passphrases.rst>`__
```
garrettr: I'm not sure whether this link should be pointing to `passphrases.rst` (which is just a high-level overview of which passphrases are needed for each user role in SecureDrop) or the section about KeePassX passphrase databases in the Tails Guide.

docs/development/getting_started.rst
ageis: Ack. My bad. I'll dive into building with Sphinx so I know what I'm doing. I was curious about that - I had only checked that links worked when browsing via GitHub.com. Thanks for the explanation.

docs/ossec_alerts.rst
```diff
@@ -152,7 +152,7 @@ server in ``/etc/ssl/certs_local`` and the system CAs will be ignored
 when validating the SMTP relay TLS certificate.
 
 Save ``prod-specific.yml``, exit the editor and `proceed with the
-installation <install.md#install-securedrop>`__ by running the
+installation <install.rst#run-the-ansible-playbook>`__ by running the
```
garrettr: Use `:ref:` instead of anchor links.

docs/development/getting_started.rst
garrettr: There should be no more expectation that this stuff works while browsing around Github.com. If it does, that's nice, but I'm not going to go out of my way to make that work. That was the old way. This is the new.

ageis:
Fixed things up in the latest commit. I'm up and running with Sphinx now! No warnings in the build. re browsing on github.com, yeah, it won't work anymore. 

re the passphrases link, I'm not sure either, but maybe shouldn't worry about it.

garrettr:
@ageis Much better, thanks for addressing those issues!

> re the passphrases link, I'm not sure either, but maybe shouldn't worry about it.

Yeah, let's leave it as-is for now. We'll come back to it naturally as we continue to refine and improve the new docs.

garrettr:
Can you rebase this (to get rid of the annoying and superfluous "Merge branch" commits)? Once you do, it'll be ready to merge.

garrettr:
@ageis Once we've merged this, I'll do a final pass over everything in the repo to make sure there are no more dead links to the old Markdown documentation.

ageis:
@garrettr Ok, been rebased. Those were all the .md links I could find using grep.

garrettr:
Nice work, @ageis :smile: :+1:

PR #1165, opened 2015-10-30
Redirect source to /lookup if visiting /generate and already logged in

pwplus:
Add ability to log out at /logout.

Add test to ensure that /generate handles users who are already logged in appropriately

Improve test for generate when already logged in

Add log out button to source_templates/base.html

This is the exact same thing as this PR, but squashed and rebased to the current source since mine was 2 months outdated: https://github.com/freedomofpress/securedrop/pull/1080
I ended up getting the job at the company I interviewed for and had to get managerial approval before I could continue working on this project. Luckily, I got managerial approval!

### DIFF
--- a/securedrop/source.py
+++ b/securedrop/source.py
@@ -154,12 +154,10 @@ def generate_unique_codename(num_words):
 
 @app.route('/generate', methods=('GET', 'POST'))
 def generate():
-    # Popping this key prevents errors when a logged in user returns to /generate.
-    # TODO: is this the best experience? A logged in user will be automatically
-    # logged out if they navigate to /generate by accident, which could be
-    # confusing. It might be better to instead redirect them to the lookup
-    # page, or inform them that they're logged in.
-    session.pop('logged_in', None)
+    if logged_in():
+        flash("You were redirected because you are already logged in. If you want"
+              "to create a new account, you should log out first.", "notification")
+        return redirect(url_for('lookup'))
 
     num_words = 7
     if request.method == 'POST':
@@ -381,6 +379,15 @@ def login():
                     "Login failed for invalid codename".format(codename))
             flash("Sorry, that is not a recognized codename.", "error")
     return render_template('login.html')
+
+@app.route('/logout')
+def logout():
+    if logged_in():
+        session.clear()
+        flash("Thank you for logging out.", "notification")
+
+    return redirect(url_for('index'))
+        
 
 
 @app.route('/howto-disable-js')
--- a/securedrop/source_templates/base.html
+++ b/securedrop/source_templates/base.html
@@ -27,6 +27,9 @@
       {% endblock %}
 
       <div class="panel selected">
+	{% if 'logged_in' in session %}
+	<a href="{{ url_for('logout') }}" class="btn pull-right" id="logout">Log Out</a>
+	{% endif %}
         <hr class="no-line" />
 
         {% block body %}{% endblock %}
--- a/securedrop/tests/test_unit_source.py
+++ b/securedrop/tests/test_unit_source.py
@@ -86,6 +86,17 @@ class TestSource(TestCase):
         already_have_codename_link = soup.select('a#already-have-codename')[0]
         self.assertEqual(already_have_codename_link['href'], '/login')
 
+    def test_generate_already_logged_in(self):
+        self._new_codename()
+        # Make sure it redirects to /lookup when logged in
+        rv = self.client.get('/generate')
+        self.assertEqual(rv.status_code, 302)
+        # Make sure it flashes the message on the lookup page
+        rv = self.client.get('/generate', follow_redirects=True)
+        # Should redirect to /lookup
+        self.assertEqual(rv.status_code, 200)
+        self.assertIn("because you are already logged in.", rv.data)
+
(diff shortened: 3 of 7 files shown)


securedrop/source.py
```diff
@@ -151,12 +151,9 @@ def generate_unique_codename(num_words):
...
 @app.route('/generate', methods=('GET', 'POST'))
 def generate():
-    # Popping this key prevents errors when a logged in user returns to /generate.
-    # TODO: is this the best experience? A logged in user will be automatically
-    # logged out if they navigate to /generate by accident, which could be
-    # confusing. It might be better to instead redirect them to the lookup
-    # page, or inform them that they're logged in.
-    session.pop('logged_in', None)
+    if logged_in():
+        flash("You were redirected from {} because you are already logged in.".format(url_for('generate')), "notification")
```
garrettr: I think a better message here would be "You were redirected because you are already logged in. If you want to create a new account, you should log out first".

securedrop/source_templates/base.html
```diff
@@ -26,6 +26,9 @@
       {% endblock %}
 
       <div class="panel selected">
+	{% if 'logged_in' in session %}
+	<a href="{{ url_for('logout') }}" class="btn pull-right">Log Out</a>
+	{% endif %}
```
garrettr: Minor nit: Top right seems like a good location for this (it's a familiar location for a log out button), but I'm not sure if I like it better inside or outside of the `.panel` div. However, that's a minor design think that we'll probably end up revisiting later anyway, so I think it's fine to merge the way it is.

securedrop/tests/test_unit_source.py
```diff
@@ -86,6 +86,17 @@ def test_generate_has_login_link(self):
...
+        self._new_codename()
+        # Make sure it redirects to /lookup when logged in
+        rv = self.client.get('/generate')
+        self.assertEqual(rv.status_code, 302)
+        # Make sure it flashes the message on the lookup page
+        rv = self.client.get('/generate', follow_redirects=True)
+        # Should redirect to /lookup
+        self.assertEqual(rv.status_code, 200)
+        self.assertIn("because you are already logged in.", rv.data)
+
```
garrettr: You should also have a unit test for logging out, and maybe a functional test for clicking the "Logout" button in the Source Interface.

securedrop/source_templates/base.html
pwplus: I originally had it outside the .panel. It shifted the entire page down a button size; I felt that it looked very ugly.

pwplus:
I intend to write these tests in my free time over the next week.

securedrop/source.py
```diff
@@ -151,12 +151,10 @@ def generate_unique_codename(num_words):
...
 @app.route('/generate', methods=('GET', 'POST'))
 def generate():
-    # Popping this key prevents errors when a logged in user returns to /generate.
-    # TODO: is this the best experience? A logged in user will be automatically
-    # logged out if they navigate to /generate by accident, which could be
-    # confusing. It might be better to instead redirect them to the lookup
-    # page, or inform them that they're logged in.
-    session.pop('logged_in', None)
+    if logged_in():
+        flash("You were redirected because you are already logged in. If you want\
```
garrettr: Personally, for long strings, I prefer to use Python's line continuation behavior, which merges multiple quoted strings together automatically. Like this:

```py
flash("You were redirected because you are already logged in. If you want"
      "to create a new account, you should log out first.", "notification")
```

securedrop/source.py
```diff
@@ -379,6 +377,15 @@ def login():
             flash("Sorry, that is not a recognized codename.", "error")
     return render_template('login.html')
 
+@app.route('/logout')
+def logout():
+    if logged_in():
+        session.pop('logged_in', None)
```
garrettr: The session also contains the `codename`, which is a sensitive piece of information. Might as well clear the whole thing, all of the data is invalidated once a user logs out. `session.clear()` should do the trick.

securedrop/source_templates/base.html
garrettr: Ok, then this is fine for now. We can come back to it later if we decide to rework the design for the Source Interface in a more holistic manner.

garrettr:
@pwplus I added some more line comments. If you need help writing the additional tests, let me know!

securedrop/tests/test_unit_source.py
pwplus: Haven't gotten a functional test yet, but that should handle the unit test portion of logging out. I'll try to work out the functional tests, but any tips would greatly appreciated.

garrettr: For now, I'd just add a logout step for the source in `submit_and_retrieve_{file,message}.py`. You'll want to add `_source_logs_out` to `source_navigation_steps.py`, and test that you see the expected flashed message when they log out. You could also test that the session is cleared for good measure.

pwplus: There was no obvious ways to check the session in the functional test since it is outside the context of flask. The unit test does check whether session is cleared though.

garrettr:
Ah, I missed that you'd implemented the functional tests because you rebased. Lgtm!

Issue #967, opened 2015-04-01
proxy apt-get and dns over tor


dolanjs:
Currently apt-get and dns requests are not proxied over tor.  We should migrate both of those over tor.

conorsch:
#1204 suggests using `apt-transport-tor` to handle the apt half of this issue.

conorsch:
Closing in favor of #1204. Discussion will continue there.

PR #1169, opened 2015-11-03
Change Tor package repository to HTTPS transport

ageis:
deb.torproject.org repos are reliably being served over HTTPS. We can update our stuff accordingly to fetch Tor with transport layer security. (the `snapci` role already uses https://deb.torproject.org, not sure why that is)

### DIFF
--- a/install_files/ansible-base/roles/common-app/tasks/install_tor.yml
+++ b/install_files/ansible-base/roles/common-app/tasks/install_tor.yml
@@ -9,7 +9,7 @@
 
 - name: setup tor apt repo
   apt_repository:
-    repo: deb http://deb.torproject.org/torproject.org {{ ansible_lsb.codename }} main
+    repo: deb https://deb.torproject.org/torproject.org {{ ansible_lsb.codename }} main
     state: present
     update_cache: yes
   tags:
--- a/install_files/ansible-base/roles/common-mon/tasks/install_tor.yml
+++ b/install_files/ansible-base/roles/common-mon/tasks/install_tor.yml
@@ -9,7 +9,7 @@
 
 - name: setup tor apt repo
   apt_repository:
-    repo: deb http://deb.torproject.org/torproject.org {{ ansible_lsb.codename }} main
+    repo: deb https://deb.torproject.org/torproject.org {{ ansible_lsb.codename }} main
     state: present
     update_cache: yes
   tags:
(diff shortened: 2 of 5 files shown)


garrettr:
I'm not sure if this is a good idea. As we've discussed before, HTTPS provides only marginal advantages in the context of apt because apt verifies that the packages are signed independently of the security/authenticity of the underlying transport.

You could argue that if the Tor package signing key is compromised, using HTTPS prevents a subset of attackers who might seek to deliver signed malicious updates to SecureDrop instances. I don't find this argument very convincing, because it's hard to imagine an attacker sophisticated enough to obtain this signing key couldn't also compromise Tor's infrastructure and/or a certificate authority, in which case the protection from HTTPS is rendered moot.

You could argue that using HTTPS provides a privacy benefit, and may make it harder to learn specific details about the servers (e.g. the version of Tor they are currently running) but in the context of SecureDrop's current update fetching behavior (every 24h), I highly doubt this would make a meaningful difference. An attacker could just infer what version of Tor the servers are running.

In fact, this change might make it even _easier_ for some types of network attackers to fingerprint SecureDrop servers based on their traffic patterns. I would bet that the majority of servers that use Tor's package repos use HTTP, so switching to HTTPS might make SecureDrop stand out more.

This conversation can quickly spiral down a rabbit hole, as we have already begun exploring in #967 and in offline discussions. If the goal is to minimize SecureDrop's fingerprintability, this PR doesn't help and may even hurt. If the goal is to provide defense in depth for the case where an adversary can compromise Tor's signing key, I think the benefit is negligible.

Are there any convincing reasons for me to merge this?

garrettr:
The potential concerns with using APT+HTTP for SecureDrop are described in **iSEC-15FTC-8** from the [last iSec audit](https://freedom.press/securedrop-files/iSEC_OTF_FPF_SecureDrop_Deliverable_v1.1_2.pdf), but we agreed during conversations with the auditors that a better overall solution is to:

1. Mirror all the package repositories used by SecureDrop on a server controlled by Freedom of the Press Foundation (e.g Ubuntu's upstreams, Tor's repos, our own package repository)
2. Make this available as a Tor Hidden Service
3. Require SecureDrop to connect to the mirror's THS for updates

This solves the following problems:

1. apt traffic is encrypted (end-to-end due to the Tor Hidden Service protocol), so it cannot be easily MITM'ed or selectively blocked like non-Tor HTTP traffic can.
2. No third-party server (e.g. the Ubuntu or Tor update servers) gets direct requests from the SecureDrop servers, which eliminates several fingerprinting and DoS vectors.
3. Limits the attack surface available to a network-only attacker due to package updates to just Tor, which is already an accepted attack surface due to the need to host the Source Interface publicly.

garrettr:
Since nobody provided a convincing reason for me to merge this, closing without merging. I think we should return to this issue later with a more holistic approach, as outlined in https://github.com/freedomofpress/securedrop/pull/1169#issuecomment-153867317.

Issue #1166, opened 2015-11-01
Broken links


chemagallego:
Next links are broken: Error 404

https://github.com/freedomofpress/securedrop/blob/develop/docs/install.md

https://github.com/freedomofpress/securedrop/blob/develop/docs/source_user_manual.md

https://github.com/freedomofpress/securedrop/blob/develop/docs/journalist_user_manual.md

https://github.com/freedomofpress/securedrop/blob/develop/docs/develop.md

ageis:
This is because we converted our docs from Markdown to reStructuredText for [ReadTheDocs](https://docs.securedrop.org) in #1158.

garrettr:
@ageis While the ReadTheDocs migration was the cause of the broken links, that does not mean the issue can just be closed.

@chemagallego Where are you getting the broken links from? Are they on securedrop.org or another website? From your browser bookmarks? Somewhere else?

chemagallego:
1) Yesterday was broken but now It seems to work properly once again at https://securedrop.org/faq#media_orgs

![2](https://cloud.githubusercontent.com/assets/6845281/10879417/10ed2124-8154-11e5-8034-f645645e5310.PNG)

2) At https://github.com/freedomofpress/securedrop it fails now (Error 404) in the four links I show you below.

![1](https://cloud.githubusercontent.com/assets/6845281/10879455/6292acf6-8154-11e5-93cc-ba807db304f5.PNG)

Example...

![3](https://cloud.githubusercontent.com/assets/6845281/10879573/fcf532d2-8154-11e5-90d0-23ca6f81076c.PNG)

3) I'm not using bookmarks, just searching using any search engine.

garrettr:
@chemagallego Thanks for the detailed response! From it, I can see multiple issues that we need to address:

- [x] Links on securedrop.org, which appear to have since been fixed, as you said in your comment
- [ ] Links in the README. I am fixing them now.
- [ ] Search engine results. This is harder to fix, and I am not sure what the best approach is.

ageis:
Google actually has a tool for removing outdated links, I can cover that:

![removal](https://cloud.githubusercontent.com/assets/1161532/10888642/41beec22-8142-11e5-8eda-3adc7fc6b532.png)

conorsch:
[...]

PR #1174, opened 2015-11-04
Add placeholders to redirect people from old docs

garrettr:
Before we migrated to ReadTheDocs in #1158, we regularly shared links to
the documentation as links to the Markdown files on GitHub (because
GitHub would render the Markdown for us automatically). In order to
assist people who may still be using those old links, I've added
placeholders for some of our more popular documentation (based on a
quick skim of my Sent email and the broken links cited in #1166) that
explain the migration and redirect users to the correct pages in the new
docs.

### DIFF
--- a/docs/develop.md
+++ b/docs/develop.md
@@ -1,0 +1,20 @@
+## Notice: Documentation Moved
+
+**New documentation:** https://docs.securedrop.org
+
+If you are viewing this file on GitHub in your web browser, you are using an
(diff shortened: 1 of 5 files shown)


conorsch:
This is great. The unique URLs for each guide are particularly useful. Should prevent headaches for folks still using the old URLs, particularly for `install.md` and `hardware.md`.

Fine to merge. :+1:

PR #1182, opened 2015-11-07
Persistent automatic SecureDrop Tails setup (redo)

ageis:
This is PR #955 redone. Aspects of these modifications to `install.sh` have been extensively reviewed and tested in the past, but not on the latest version of Tails. We're doing the following things:
 
    * Detects whether user is an admin or journalist
    * Configures HidServAuth values and eliminates need to manually edit torrc_additions
    * Adds desktop and main menu launchers for both interfaces
    * Sets up SSH host aliases for mon and app server (if admin)
    * Updates inventory file to run playbooks over Tor (if admin) 
    * Use 'live-additional-software' to always install Ansible (if admin)
    * Updates the documentation to reflect

The shortcut setup will have to be updated to deal with TLS/HTTPS when that feature comes in. Those links are currently HTTP only.

### DIFF
--- a/docs/configure_admin_workstation_post_install.rst
+++ b/docs/configure_admin_workstation_post_install.rst
@@ -42,41 +42,11 @@ directory with these scripts, and run the install script:
 
 Type the Administration Password that you selected when starting Tails
 and hit enter. The install script will download some additional
-software, which may take a few minutes. It will then pop open a text
-editor with a file named ``torrc_additions``.
+software, which may take a few minutes.
 
-Edit this file, adding the ``HidServAuth`` information for each of the
-three authenticated hidden services that were created during the
-install process:
-
-#. ``app-document-aths``
-#. ``app-ssh-aths``
-#. ``mon-ssh-aths``
-
-You can double-click or use the 'cat' command to read the values from
-the corresponding files in ``install_files/ansible-base``. When you're
-done, ``torrc_additions`` will look something like this:
-
-::
-
-    # add HidServAuth lines here
-    HidServAuth gu6yn2ml6ns5qupv.onion Us3xMTN85VIj5NOnkNWzW # client: user
-    HidServAuth fsrrszf5qw7z2kjh.onion xW538OvHlDUo5n4LGpQTNh # client: admin
-    HidServAuth yt4j52ajfvbmvtc7.onion vNN33wepGtGCFd5HHPiY1h # client: admin
-
-.. tip:: An easy way to do this is to run ``cat *-aths`` from the
-	 ``install_files/ansible-base`` folder in a terminal window,
-	 and copy/paste the output into the opened text editor.
-
-When you are done, click **Save** and close the text editor. Once the
-editor is closed, the install script will automatically continue.
-
-.. todo:: This process is confusing and error prone. Since the
-          necessary information (the ``*-aths`` files) is in a
-          predictable location, can't we just skip the "manually
-          copy/paste into a text editor" step?
-
-The install scripts installs a Network Manager hook that runs on boot. It automatically adds the HidServAuth values from ``torrc_additions`` to the ``torrc`` and restarts Tor.
+This script installs a Network Manager hook that runs every time you connect to Tor. It automatically adds the HidServAuth values from ``torrc_additions`` to the ``torrc`` and restarts Tor.
+In addition, it creates desktop and main menu shortcuts for both interfaces, updates your Ansible inventory file to install playbooks over Tor in the future, directs Tails
+to install Ansible at the beginning of every session, and sets up SSH host aliases for the servers.
 
 .. note:: The only thing you need to remember to do is enable
           persistence when you boot the Admin Workstation. If you are
@@ -91,25 +61,11 @@ The install scripts installs a Network Manager hook that runs on boot. It automa
 SSH Host Aliases
 ----------------
 
-.. note:: This step is optional, but it makes it much easier to
-	  connect to the servers in order to administrate them in the future.
+The installation script in ``tails_files`` also sets up SSH host aliases for the admin.
 
-Create the file ``/home/amnesia/.ssh/config`` and add a configuration
-following the scheme below, replacing ``Hostname`` and ``User`` with
-the values specific to your setup:
-
-::
-
-    Host app
-      Hostname <app hostname>.onion
-      User <ssh username>
-    Host mon
-      Hostname <mon hostname>.onion
-      User <ssh username>
-
-Now you can simply use ``ssh app`` and ``ssh mon`` to connect to each
+You can simply use ``ssh app`` and ``ssh mon`` to connect to each
 server. This configuration will be persisted across reboots thanks to
-Tails' SSH Client persistence.
+Tails' SSH client persistence.
 
 Set up two-factor authentication for the Admin
 ----------------------------------------------
--- a/docs/configure_admin_workstation_post_install.rst
+++ b/docs/configure_admin_workstation_post_install.rst
@@ -6,43 +6,41 @@ Configure the Admin Workstation Post-Install
 Auto-connect to the Authenticated Tor Hidden Services
 -----------------------------------------------------
 
-The SecureDrop installation process adds multiple layers of
-authentication to protect access to the most sensitive assets in the
-SecureDrop system:
+The SecureDrop installation process adds multiple layers of authentication to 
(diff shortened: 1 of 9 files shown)


docs/configure_admin_workstation_post_install.rst
```diff
@@ -42,41 +42,11 @@ directory with these scripts, and run the install script:
...
-
-.. todo:: This process is confusing and error prone. Since the
-          necessary information (the ``*-aths`` files) is in a
-          predictable location, can't we just skip the "manually
-          copy/paste into a text editor" step?
-
-The install scripts installs a Network Manager hook that runs on boot. It automatically adds the HidServAuth values from ``torrc_additions`` to the ``torrc`` and restarts Tor.
+This script installs a Network Manager hook that runs every time you connect to Tor. It automatically adds the HidServAuth values from ``torrc_additions`` to the ``torrc`` and restarts Tor.
+In addition, it creates desktop and main menu shortcuts for both interfaces, updates your Ansible inventory file to install playbooks over Tor in the future, directs Tails
+to install Ansible at the beginning of every session, and sets up SSH host aliases for the servers.
```
garrettr: In the absence of documented criteria for formatting the docs, please adhere to the format of the surrounding source. In this case, that means hard wrapping lines at ~80 characters.

docs/configure_admin_workstation_post_install.rst
```diff
@@ -91,25 +61,11 @@ The install scripts installs a Network Manager hook that runs on boot. It automa
...
-
-::
-
-    Host app
-      Hostname <app hostname>.onion
-      User <ssh username>
-    Host mon
-      Hostname <mon hostname>.onion
-      User <ssh username>
+The installation script in ``tails_files`` also sets up SSH host aliases for the admin.
```
garrettr: So this is less "magic"-seeming, maybe add a note that you can see the host aliases added by the install script in `~/.ssh/config`.

docs/tails_guide.rst
```diff
@@ -147,37 +147,29 @@ installation by typing these commands into the terminal:
...
+and hit enter. This installation script does the following: 
 
-Edit the file, inserting the *HidServAuth* information for your
-SecureDrop instance that you received during the installation process.
-The values can be found in
-``install_files/ansible-base/app-document-aths``. This information
-includes the address to the Document Interface and your personal
-authentication string, as seen in the example below:
+* Downloads additional software
+* Installs a program that automatically and persistently configures Tor to access the
```
garrettr: Specify that this automatic configuration happens during boot

ageis: @garrettr I think of the NM hooks as contingent on the network connection in the session, rather than part of the "boot" process.

docs/tails_guide.rst
```diff
@@ -147,37 +147,29 @@ installation by typing these commands into the terminal:
...
 able to access to the *Document Interface*, even if they manage to steal
-the journalist's username, password, and two-factor authentication
-token.
-
-Our ``install.sh`` sets up an initialization script that automatically
-updates Tor's configuration to work with SecureDrop every time you login
-to Tails. As long as Tails is booted with the persistent volume enabled
-then you can open the Tor Browser and connect to the Document Interface
-as normal.
+the journalist's username, password, and two-factor authentication token.
```
garrettr: Actually, the primary reason we use an authenticated Tor Hidden Service for the Document Interface is so an adversary cannot access the Document Interface, which provides a layer of defense in depth and protects the Document Interface even if there is a security vulnerability in the authentication implemented in the Document Interface web application.

See https://github.com/freedomofpress/securedrop/pull/974#issuecomment-88643590 for more info on why this is an important layer of defense-in-depth.

install_files/ansible-base/inventory
```diff
@@ -5,7 +5,7 @@
 #
 # NOTE: after the install is complete, you can re-provision with Ansible but you will need to:
 # 1. Set up HidServAuth in your torrc with the values from app-ssh-aths and mon-ssh-aths
-# 2. Replace the IP addresses here with the corresponding .onion addresses
+# 2. Replace the IP addresses here with the corresponding addresses
```
garrettr: Why remove `.onion`? Seems like a useful hint to me.

ageis: It's required for the grep on line 83 of install.sh to work properly.

garrettr: Well, you could always make that regex a little smarter right? Something like "onion but not after a #" should be pretty easy right?

docs/tails_guide.rst
garrettr: That's fair, you can leave this as-is.

install_files/ansible-base/inventory
ageis: Updated with an inverse grep so that we don't need to edit that comment. This still works even if the comment is not present.

garrettr:
Some of the line wrapping in the `.rst` files is still pretty wack.

docs/tails_guide.rst
```diff
@@ -41,14 +41,17 @@ create a Tails USB stick. Here are some links to help you out:
 -  `Install onto a USB stick or SD
    card <https://tails.boum.org/doc/first_steps/installation/index.en.html>`__
 -  `Create & configure the persistent
-   volume <https://tails.boum.org/doc/first_steps/persistence/configure/index.en.html>`__
+   volume 
+<https://tails.boum.org/doc/first_steps/persistence/configure/index.en.html>`__
```
garrettr: e.g. this line wrapping is all messed up. Note that you only need one `_` to indicate a link (aka [reference name](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#reference-names)) in ReST.

In general, I prefer [external hyperlink targets](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#hyperlink-targets) to internal hyperlink targets because:

1. They are reusable
2. They make the text easier to read (no URLs all over the place) and also more closely resemble the rendered output
3. They help avoid line length problems like the one you have here.

ageis:
@garrettr I attempted to address issues in the latest commit... Let me know if I succeeded.

docs/tails_guide.rst
```diff
@@ -1,72 +1,70 @@
...
--  `Download and verify the Tails
-   .iso <https://tails.boum.org/download/index.en.html>`__
--  `Install onto a USB stick or SD
-   card <https://tails.boum.org/doc/first_steps/installation/index.en.html>`__
--  `Create & configure the persistent
-   volume <https://tails.boum.org/doc/first_steps/persistence/configure/index.en.html>`__
+.. _`Download and verify the Tails .iso`: https://tails.boum.org/download/index.en.html
+.. _`Install onto a USB stick or SD card`: https://tails.boum.org/doc/first_steps/installation/index.en.html
+.. _`Create & configure the persistent volume`: https://tails.boum.org/doc/first_steps/persistence/configure/index.en.html
+.. _`manually installing Tails onto a USB device on Mac OS X`: https://tails.boum.org/doc/first_steps/installation/manual/mac/index.en.html
```
garrettr: The references should be defined _after_ they are used, not before. Look at any of the other docs to see how this is done.

docs/overview.rst
```diff
@@ -56,52 +58,50 @@ administrators, the sources, and the journalists.
...
-`Tails <https://tails.boum.org>`__ and connects to the App and Mon
-servers over authenticated `Tor Hidden
-Services <https://www.torproject.org/docs/hidden-services.html>`__ and
-manages them using `Ansible <http://www.ansible.com/>`__.
+The SecureDrop servers are managed by a systems administrator; for larger 
+newsrooms, there may be a team of systems administrators. The administrator 
+uses a dedicated *Admin Workstation* running `Tails <https://tails.boum.org>`__ 
+and connects to the App and Mon servers over authenticated `Tor Hidden Services 
+<https://www.torproject.org/docs/hidden-services.html>`__ and manages them 
+using `Ansible <http://www.ansible.com/>`__.
```
garrettr: Do you need two underscores? Like I said in an earlier comment, I'm pretty sure you only need one.

ageis: That's the way those links were left from the original .md to .rst conversion. If I start converting two underscores to one, I'd like to do it in every doc for consistency.

garrettr: Ah, I see. That's annoying. Just leave it alone for now, if we do that I agree it should be done all at once. If it ain't broke, don't fix it.

PR #1185, opened 2015-11-08
Set Cache-Control header time of 30 minutes.

patrickod:
Previously we used
 * cache-control
 * pragma
 * expires

headers to ensure that assets being served were not cached at all by the
client, the theory being that we don't want to leave any forensic
evidence on the source machine indicating that they've used securedrop.

TorBrowser patches Firefox such that it does not persist cached assets
to disk, and keeps them solely in RAM.

Neither TorBrowser, nor our use of cache-control headers can prevent the
OS from swapping to disk, in which case all cached assets are persisted
to disk leaking information.

The size of the assets being served in securedrop is in the order of
tens of kilobytes now that we're minimizing them. It is highly unlikely
that these assets will be the straw to break the camels back memory
wise causing a swap to disk.

Without the caching headers the experience of loading securedrop is
disjointed, as the stylesheets load well after the page has begun
rendering. The experience of loading securedrop is *greatly* improved by
allowing short-term caching of these assets.

### DIFF
--- a/install_files/ansible-base/roles/app/templates/sites-available/document.conf
+++ b/install_files/ansible-base/roles/app/templates/sites-available/document.conf
@@ -12,10 +12,8 @@ XSendFile        On
 XSendFilePath    /var/lib/securedrop/store/
 XSendFilePath    /var/lib/securedrop/tmp/
 
-Header set Cache-Control "max-age=0, no-cache, no-store, must-revalidate"
+Header set Cache-Control "max-age=1800"
 Header edit Set-Cookie ^(.*)$ $1;HttpOnly
-Header set Pragma "no-cache"
-Header set Expires "-1"
 Header always append X-Frame-Options: DENY
 Header set X-XSS-Protection: "1; mode=block"
 Header set X-Content-Type-Options: nosniff
--- a/install_files/ansible-base/roles/app/templates/sites-available/source.conf
+++ b/install_files/ansible-base/roles/app/templates/sites-available/source.conf
@@ -10,10 +10,8 @@ AddType text/html .py
 
 XSendFile        Off
 
-Header set Cache-Control "max-age=0, no-cache, no-store, must-revalidate"
+Header set Cache-Control "max-age=1800, must-revalidate"
 Header edit Set-Cookie ^(.*)$ $1;HttpOnly
-Header set Pragma "no-cache"
-Header set Expires "-1"
 Header always append X-Frame-Options: DENY
 Header set X-XSS-Protection: "1; mode=block"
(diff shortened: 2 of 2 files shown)


patrickod:
Mike Perry had the following to say about the Tor Browser's caching behavior

```
15:48 <patrickod> mikeperry: ping if you're around?
15:48 <mikeperry> patrickod: yeah, somewhat here
15:49 <patrickod> myself and garret from FPF have a question about caching of assets in tor browser 
15:52 <patrickod> tl;dr in securedrop we set cache-control and related headers to avoid caching assets in the browser. the working theory is that might prevent leaking forensic information about the source's use of securedrop to their persistent storage
15:52 <patrickod> we're wondering if tor browser has any specific mitigations for this type of concern?
15:53 <patrickod> it degrades the user experience considerably, as the stylesheets loads after the page, and we're also concerned that this type of asset might be swapped to disk regardless of the cache header settings, in which case it's not a useful setting to have
15:53 <mikeperry> the default config does not cache anything to disk
15:53 <patrickod> so in our case we can set these cache headers and tor browser will keep cached assets in RAM and avoid writing to disk
15:54 <mikeperry> there is an option to enable disk records, and that enables all manner of disk storage
15:54 <mikeperry> including url history
15:57 <mikeperry> disabling caching entirely may reduce the risk of swapping this memory-cached data out to disk, though (since we do have a memory-based cache by default)
16:00 <patrickod> do you mean disabling via a pref in tor browser? or via cache control headers?
16:06 <mikeperry> disabling caching entirely via cache control headers may reduce the risk of swapping site data out to disk (since it will only be in the DOM and not kept in the memory cache for long periods)
16:21 <patrickod> mikeperry: would you mind if I immortalized this conversation in the resulting github issue?
16:24 <mikeperry> sure, so long as the "may" is clear. you could also set a cache lifetime of something like 10 minutes to have a similar effect. at least, I think
16:25 <mikeperry> about:cache and about:memory are your friends, as are gdb core dumps
16:34 <patrickod> +1 yep intend on discussing this, not just dumping the chatlog as sole justification
```

Mike notes that keeping assets in RAM for any amount of time may increase the possibility that they'll be swapped out to disk. 

Myself and @garrettr discussed this in person at the Aaron Swartz hackathon, and having tested it using onion services, we believe the improvement in the user's experience is worth this tradeoff.

garrettr:
Further forensic analysis of this would be an interesting project! In the meantime, I agree that the usability benefits outweigh the potential forensic concerns, especially since the forensic *benefits* are completely theoretical and cannot be guaranteed in any case due to OS-specific swap behavior.

Issue #1188, opened 2015-11-08
Delete Submissions that have had their Sources deleted


pwplus:
Right now if you delete a source, it does not delete the corresponding submissions from the database.

PR #1190, opened 2015-11-08
Delete submissions and replies when source deleted

pwplus:
This PR will delete submissions and replies linked to a source when a
source is deleted. This will fix https://github.com/freedomofpress/securedrop/issues/1188.

### DIFF
--- a/securedrop/db.py
+++ b/securedrop/db.py
@@ -132,7 +132,8 @@ class Submission(Base):
         "Source",
         backref=backref(
             'submissions',
-            order_by=id))
+            order_by=id,
+        cascade="delete"))
     filename = Column(String(255), nullable=False)
     size = Column(Integer, nullable=False)
     downloaded = Column(Boolean, default=False)
@@ -158,7 +159,8 @@ class Reply(Base):
             order_by=id))
 
     source_id = Column(Integer, ForeignKey('sources.id'))
-    source = relationship("Source", backref=backref('replies', order_by=id))
+    source = relationship("Source", backref=backref('replies', order_by=id,
+                          cascade="delete"))
 
     filename = Column(String(255), nullable=False)
     size = Column(Integer, nullable=False)


securedrop/db.py
```diff
@@ -77,6 +77,8 @@ class Source(Base):
     flagged = Column(Boolean, default=False)
     last_updated = Column(DateTime, default=datetime.datetime.utcnow)
     star = relationship("SourceStar", uselist=False, backref="source")
+    submissions_delete = relationship("Submission", cascade="delete")
+    replies_delete = relationship("Reply", cascade="delete")
```
garrettr: How does this interact with the backrefs to `Source` that are defined on `Submission` and `Reply`? Seems like a cleaner approach might be to set up the cascade on the existing backrefs (see the [SQLAlchemy documentation on cascades](http://docs.sqlalchemy.org/en/rel_0_9/orm/cascades.html)).

redshiftzero:
Thanks for this contribution! Would you be interested in writing a unit test for this functionality?

pwplusnick:
I would need to check with my current employer. While I'm generally free to write for projects outside the problem domain I work in, this is kind of on the fence. I'm going to guess that I could, but let me double check first.

conorsch:
> I would need to check with my current employer. 

@pwplus No sweat, your contributions thus far have been really helpful. @redshiftzero Could you tackle adding some tests here? You'll need to rebranch from @pwplus's fork in order to add the test commits, but he'll still get credit for the changes he wrote when it comes time to merge.

redshiftzero:
Sure will do @conorsch

pwplusnick:
They are taking a while to get back to me. It is probably best if you guys just take it from here.

redshiftzero:
Hey @pwplus I pulled your commit into a #1440 that supersedes this one with unit testing, so I’m closing this PR in favor of #1440. Thanks again for your work!

PR #1192, opened 2015-11-09
Rebase i18n

shirkevich:


mark-in:
Maybe it's better to perform translation on client side instead of the server one? Doing so on server side may leak metadata.

garrettr:
@mark-in This is translation of the UI elements, not the submissions sent by sources or the replies sent by journalists. It is a static process that occurs during the release process, so it is unrelated to metadata.

mark-in:
@garrettr, example case:
Some country have two official languages, e.g. Belgium. If you have SecureDrop set up with French UI it becomes unavailable for a Dutch-speaking source and *vice versa*. In order to switch language the source must leak their language preference to the Application Server.

Solution could be to serve both (or more) translations to a client as a bunch and to provide an UI for lang change on a client side.

mark-in:
Also cases: a source have something to leak to foreign media or foreign source have something to leak to local media (that is not native to them).
It seems reasonable to serve *all* available translations.

garrettr:
> Solution could be to serve both (or more) translations to a client as a bunch and to provide an UI for lang change on a client side.

Dynamically changing the UI like that would require JavaScript on the Source Interface, but we intentionally encourage sources to _disable_ Javascript.

Additionally, I really don't think it's a serious metadata leak. The language preference is stored in the session, so it is only available to the server while it is processing a request, and is never written to disk. This is also true of, for example, the source's codename, which is a _much_ more sensitive piece of information, as it both identifies them and encrypts the replies sent to them by the journalist.

mark-in:
> Dynamically changing the UI like that would require JavaScript on the Source Interface, but we intentionally encourage sources to disable Javascript.

No, it wouldn't. You can make i18n super easy in plain CSS. See a [nice example](http://www.smashingmagazine.com/2014/06/css-driven-internationalization-in-javascript/#css-driven-internationalization-improved) of this.

> Additionally, I really don't think it's a serious metadata leak. The language preference is stored in the session, so it is only available to the server while it is processing a request, and is never written to disk. This is also true of, for example, the source's codename, which is a much more sensitive piece of information, as it both identifies them and encrypts the replies sent to them by the journalist.

First of all, you never know how serious metadata leak is. So we should avoid any of them.
The source's codename is not sensitive at all - it tells *nothing* about the source. This is just a random id assigned to the source and is not provided by them. Native language *is* sensitive metadata and it is one more bit of identifying information that can help to deanonymize the source.

garrettr:
@mark-in Thanks for the link to the CSS-driven i18n, I hadn't seen that before and it was really interesting!

Unfortunately, that *still* does not solve the metadata problem you brought up originally. In order for the CSS-driven technique to work, the server needs to know which language the user wants the page to be in in order to set the `lang` attributes appropriately. This is what makes the CSS `:lang()` pseudo-selectors work.

> of course one should not use lang(): and should allow source to choose their language

I think you're confused here. `:lang()` is the key CSS pseudo-selector that allows this CSS-driven i18n to work in the first place. The blog post further goes on to describe using Javascript to dynamically re-translate the page without having to re-fetch it from the server. This is a nice technique, but obviously it requires Javascript.

We don't have to use the dynamic reload JS in SecureDrop, meaning switching the language would require a full page reload - annoying, but an acceptable tradeoff to avoid using JavaScript. However, the server still needs to know the source's language preference in order to set the `lang` attribute correctly.

There are only two ways to do this: on the server, or client-side via JS. I am sorry, but given our "no-Javascript" requirement I am fairly certain that **we cannot avoid leaking a source's language preference to the server**, at least temporarily.

Again, this language preference does *not* have to be stored persistently (in the database), and we would probably want to avoid doing that because, as you point out, that is a piece of identifying information about the source. This actually goes against typical practice, and negatively impacts usability for the source, because not storing their language preference means they will have to re-select it every time they log in to the Source Interface.

garrettr:
> The source's codename is not sensitive at all - it tells nothing about the source. This is just a random id assigned to the source and is not provided by them.

@mark-in This is incorrect. It is actually the _only_ uniquely identifying piece of information that both the source and the server share. And it *is* provided by the source, because it is their login credential.

garrettr:
I agree that we should strive to minimize metadata leaks, but in the context of supporting i18n I think it is acceptable to (transiently, not persistently) leak a source's language preference to the server. Please remember that the server also receives the *plaintext* of the source's submissions, which IMO are ***far*** more identifying than their language preference.

Please take a step back and look at the big picture, and adjust your perspective accordingly.

redshiftzero:
I know that there is a lot of demand for internationalization of SecureDrop and we appreciate this is incredibly important. We really want to translate and localize SecureDrop, but given some of the other [improvements](https://github.com/freedomofpress/securedrop/milestones) we're working on and the fact that we're a small team, we don't currently have the bandwidth to add this feature and begin managing translations. We really appreciate this contribution, but given that we can't add this right now and that this PR has been pending merge for almost two years, I am going to close this PR for now.

PR #1196, opened 2015-11-12
Reorganize Ansible playbooks

conorsch:
Substantial changes to the flow of the Ansible playbooks and structure of the roles. Most of the diff is due to files that have been moved or renamed—functional changes were kept to a minimum in order to facilitate review. Once merged, future changes to the Ansible roles will be easier to make and to review.

### Major changes
* Consolidates host-specific shared roles (e.g. `common-app`, `common-mon` -> `common`) 
* Use `group_vars` for `staging` and `development` hosts
* Manage group membership in `manage-groups` role for production instances
* Manage group membership in `Vagrantfile` for testing VMs
* Support `ANSIBLE_ARGS` environment variable for managing tags and skipped tags on testing machines

### Minor changes
* More readable task names (also added task names for plays)
* Several improvements to idempotence (e.g. `command` calls only report `changed` when appropriate)
* Typo fixes (e.g. iptables comments)

Happy to review in person since the diff is large. Once this PR is merged, there's a list of **ops**-tagged issues outstanding that I plan to knock out posthaste.

PR #1197, opened 2015-11-12
Install securedrop-keyring package

ageis:
This is the change to the playbook required to install the [SecureDrop keyring package](https://github.com/freedomofpress/securedrop-keyring) as part of the FPF apt repo, which will allow us to update the public key which is used to sign the [apt.freedom.press](https://apt.freedom.press) Release file and its expiry in the future without requiring manual action by SecureDrop instance admins. For context, our master signing key expired without us being aware on October 26, 2015, necessitating the 0.3.6 hotfix.

Per #1196 , the `install_fpf_repo` role is mostly the same, it just has slashes instead of underscores.

The actual package and its build process is simple and described in the linked GitHub repository. Merging this PR should occur concurrent with adding the package to our repository, which is a whole process.

### DIFF
--- a/install_files/ansible-base/roles/install-fpf-repo/tasks/main.yml
+++ b/install_files/ansible-base/roles/install-fpf-repo/tasks/main.yml
@@ -15,3 +15,11 @@
   tags:
     - apt
     - fpf_repo
+
+- name: ensure the fpf apt repo keyring package is installed
+  apt:
+    pkg: securedrop-keyring
+    state: present
+  tags:
+    - apt
+    - fpf_repo
--- a/install_files/securedrop-app-code/DEBIAN/control
+++ b/install_files/securedrop-app-code/DEBIAN/control
@@ -6,5 +6,5 @@ Homepage: https://securedrop.org
 Package: securedrop-app-code
 Version: 0.3.6
 Architecture: amd64
-Depends: python-pip,apparmor-utils,gnupg2,haveged,python,python-pip,secure-delete,sqlite,apache2-mpm-worker,libapache2-mod-wsgi,libapache2-mod-xsendfile,redis-server,supervisor
+Depends: python-pip,apparmor-utils,gnupg2,haveged,python,python-pip,secure-delete,sqlite,apache2-mpm-worker,libapache2-mod-wsgi,libapache2-mod-xsendfile,redis-server,supervisor,securedrop-keyring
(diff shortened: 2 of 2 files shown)


garrettr:
This is blocked on us creating the securedrop-keyring package (code for that is here: https://github.com/freedomofpress/securedrop-keyring) and pushing it to the apt package server, so it needs to be part of a release :weary:.

conorsch:
The underlying task lists have moved around a bit, so these changes will likely need to be rebased. We'll also need to update the documentation to change mention of the `securedrop-keyring` package from a note to an item in the definition list. As @garrettr mentioned, we'll block on this until we have a new release.

ageis:
Rebased on latest develop. Looks like this is being targeted for 0.4. We have 5 months before the key expires again, so might be worth considering for a point release.

ageis:
Latest commit makes securedrop-keyring a dependency of securedrop-app-code. This is intended to avoid requiring manual action by admins in order to install the keyring package before the next key expiration date. Instead, it can be pulled in automatically via the app code dependencies when the app code package is updated.

See also, Recommends or Suggests: https://www.debian.org/doc/debian-policy/ch-relationships.html#s-binarydeps

garrettr:
The securedrop-keyring package is the focus of the next planned minor release for SecureDrop (0.3.10)! @ageis did a fantastic job creating the [securedrop-keyring](https://github.com/freedomofpress/securedrop-keyring) package, and this pull request looks to be sufficient to integrate it with SecureDrop. As far as I can tell, all that remains to do for this release is:

- [ ] Rebase on a new release/0.3.10 branch based on the 0.3.9 tag
- [ ] Testing. We need to test:
 - [ ] New installations
 - [ ] Updating current installations via a new SecureDrop release (using the mechanism described in https://github.com/freedomofpress/securedrop/pull/1197#issuecomment-223462885).
 - [ ] Updating key expiry via the keyring package (the primary intended use case).
 - [ ] Rotating keys via the keyring package.

conorsch:
This PR targets develop, so I'll need to close it and rebase the changes to target a release branch, as @garrettr described above. In addition I'm going to create the deb package structure inside this repository, rather than maintain it separately, so we can make use of the existing build tooling. See [comments on the securedrop-keyring repo](https://github.com/freedomofpress/securedrop-keyring/issues/1#issuecomment-250622715).

During rebase I'll port over the existing checklist items and expand on the testing methodology.

conorsch:
Superseded by #1416.

PR #1200, opened 2015-11-13
Add document describing SecureDrop apt repository

ageis:
I figured we should have something describing the contents of our repo _somewhere_. I put it under /development.

I also fixed a small Sphinx error in the upgrade docs that were merged yesterday.

### DIFF
--- a/docs/development/apt_repo.rst
+++ b/docs/development/apt_repo.rst
@@ -1,0 +1,31 @@
+SecureDrop apt repository
+=========================
+
+This document contains a brief description of the Debian packages which are 
+hosted and maintained by Freedom of the Press Foundation in our apt repository 
+at `apt.freedom.press`_.
+
+linux-image-3.14.*-grsec
+    This package contains the Linux kernel image, patched with grsecurity.
+
+linux-headers-3.14.*-grsec
+    Header files related to the Linux kernel.
+
+securedrop-app-code
+    Packages the SecureDrop application code, Python pip dependencies and 
+    AppArmor profiles.
+
+`securedrop-keyring <https://github.com/freedomofpress/securedrop-keyring>`_
+    Packages the public signing key used in conjunction with this apt 
+    repository.
+
+securedrop-ossec-agent
+    Installs the pre-configured OSSEC agent for the SecureDrop App Server.
+
+securedrop-ossec-server
+    Installs the pre-configured OSSEC manager for the SecureDrop Mon Server.
+
+`securedrop-grsec <https://github.com/freedomofpress/grsec>`_
+    SecureDrop grsec kernel (metapackage depending on the latest version).
+
+.. _apt.freedom.press: https://apt.freedom.press
--- a/docs/development/apt_repo.rst
+++ b/docs/development/apt_repo.rst
@@ -15,10 +15,6 @@ securedrop-app-code
     Packages the SecureDrop application code, Python pip dependencies and 
     AppArmor profiles.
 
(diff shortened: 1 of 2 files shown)


conorsch:
Nice use of a definition list in ReST! Sort the items alphabetically, though. It's close to alphabetical now, close enough to drive one mad.

Also take out mention of `securedrop-keyring`, since it's not currently accurate. It'll be a one-line change to add it when appropriate, but we shouldn't ever have docs pointing to as-yet-fictitious features. Maybe extract the keyring definition and add it in a **Note** at the bottom, as with the new upgrade docs?

ageis:
@conorsch hmm, I was figuring that we might be able to do a repo update including the keyring (and perhaps an updated grsec kernel too) pretty soon, since I already have PR #1197 out, and just wait until after that's done to merge this, but that's up to @garrettr. I imagine you guys will want to test the package on your own, so yeah, I'll take it out if that's gonna take some time.

conorsch:
Fingers crossed we get the keyring package straightened out, but with testing it will definitely take time. Convert the definition for the keyring package to a **Note**, then the docs will accurately reflect the current state of the project.

conorsch:
lgtm! :+1:

docs/development/apt_repo.rst
```diff
@@ -0,0 +1,33 @@
...
+
+securedrop-app-code
+    Packages the SecureDrop application code, Python pip dependencies and 
+    AppArmor profiles.
+
+securedrop-ossec-agent
+    Installs the pre-configured OSSEC agent for the SecureDrop App Server.
+
+securedrop-ossec-server
+    Installs the pre-configured OSSEC manager for the SecureDrop Mon Server.
```
garrettr: This is not quite correct. The actual OSSEC packages are `ossec-server` and `ossec-agent`, which are documented and built in a separate repo: https://github.com/freedomofpress/ossec. We do this because OSSEC wasn't providing Ubuntu packages for a long time, only Debian, but we were able to make some minor modifications to get the Debian packages to work on Ubuntu. OSSEC now provides Ubuntu-compatible packages (see http://ossec.github.io/downloads.html), but we were disappointed with how long it took them to update the packages on that repo for 2.8.2 (over a month, for a release that fixed a [serious security vulnerability](http://ossec.github.io/blog/posts/2015-06-11-cve-2015-3222-fix.markdown.html)), so we have opted to continue providing OSSEC packages from our repo for the time being.

The `securedrop-ossec-*` packages just provide SecureDrop-specific configuration for OSSEC, so this description is mostly correct, but should be amended to mention the separate packages that actually install the OSSEC software.

ageis:
@garrettr Fixed the OSSEC package descriptions per your comments.

conorsch:
Nit:
> Installs the SecureDrop-specific configuration for the App Server.

should be:

> Installs the SecureDrop-specific OSSEC configuration for the Application Server.

and

conorsch:
That'll do it.

garrettr:
@ageis If you want this to be available, you should add it to one of the toctrees in `index.rst`. It should probably go under `devdocs`. Otherwise I'm pretty sure it won't get included by Sphinx, and therefore won't be available through what gets built out on ReadTheDocs.

ageis:
The doc is indeed under development, which is included in the Developer Documentation toctree via file glob/wildcard in `index.rst`

garrettr:
@ageis Thanks, missed that it was already in `development`. r+

Issue #952, opened 2015-03-26
delete the sasl_password file after postmap


dolanjs:
For the monitor servers postfix setup. Ansible creates the sasl_password file then runs the postmap task to enter it into the db. The postfix main.cf config points to the sasl_password db not the plain text file. We should secure-delete the original sasl_password file after running the postmap command.

garrettr:
Ok. If we do this, we also need to change the documentation for troubleshooting failure to send emails in `docs/ossec_alerts.md`, since the troubleshooting recommendation for "Authentication failure" involves editing `/etc/postfix/sasl_passwd`.

Instead, we should tell people to update the `sasl_password` in `prod-specific.yml` and re-run the task to configure postfix.

dolanjs:
@garrettr I added it to the original issue @ageis created about that troubleshooting, but the admin shouldn't be editing that file directly period. If there was a mistake in that file or a future password change the admin should modify the prod-specific.yml and re-run the playbook. Not edit the file. @ageis ran into issues with changing the IP addresses and sasl password because he was editing the files directly on the server instead of the ansible config file and re-running the playbook.

ageis:
Yeah just to get specific, my issue was with the SMTP relay port (not IP or password) and the corresponding iptables rule for outbound SMTP which depends on the Ansible variable. For that reason editing the relay hostname or sasl_passwd is easier/safer than changing the port, but I agree they should not do it. That's why in the most recent revision of ossec_alerts.md I added a note "We recommend modifying and re-running the Ansible playbook instead of doing things like this."

conorsch:
Dropping the 0.4 milestone since this will require careful testing, and the even new config tests in #1616 don't include functional testing of the OSSEC email alerts.

markldevine:
When I `less /etc/postfix/sasl_passwd.db` and scroll to the bottom, there is my clear text password to my Office 365 account.  This should be obfuscated to prevent instant-can't-unsee-it exposure.  "Keep Honest People Honest" with anything other than clear text.  Base64 at a minimum...

ageis:
According to the documentation, using hashed or encrypted passwords with Postfix/SASL requires a usage of an SQL or LDAP auxprop plugin, or it can be integrated with PAM.

PR #1201, opened 2015-11-16
Cosmetic changes to OSSEC docs

conorsch:
While following up on the docs discussion in #952, fixed some formatting issues left over from the Markdown -> ReST conversion (#1158). Makes the OSSEC docs conform more closely to the new contribution guidelines, particularly with regard to linewrapping.

Also adds a specific header **Useful log files for OSSEC** and a **Tip** about encrypting logs when sending via email.


ageis:
👍

PR #1205, opened 2015-11-17
Fixes fetching Source Interface THS URL

conorsch:
Recent changes (#1196) to how the Tor hidden service files are collected from the servers and written to the Admin Workstation yielded a regression in formatting: the `HidServAuth` string was inappropriately prepended to _all_ hidden service files. Instead, it should be prepended only to the ATHS files:

* `app-document-aths`
* `app-ssh-aths`
* `mon-ssh-aths`

but _not_ to the THS file (the Source Interface) file:

* `app-source-ths`

Thanks for spotting, @ageis!


conorsch:
@ageis By the way, you can now pass in `--tags admin,tor` to target rerunning just these tasks for testing the Tails files scripts.

garrettr:
lgtm

Issue #1175, opened 2015-11-04
Multiple administrators


tomlowenthal:
What if SecureDrop needs to have multiple independent administrators? How should that be managed?

conorsch:
Right now the Ansible playbook supports multiple admin user accounts, simply by appending to the `ssh_users` variable in `prod-specific.yml`. The value of `ssh_users` can be a string:

```
ssh_users: alice
```

or it can be a YAML list:

```
ssh_users:
  - alice
  - bob
```

In both cases, the Ansible playbook will "do the right thing" and grant admin privileges to both usernames. Specifically, the playbook will:

1. Create the username
2. Add the username to the `sudo` and `ssh` groups
3. Grant passwordless `sudo`

That's all well and good, but it largely assumes that both admins will be working from the same Tails device—not an option if the admins are not geographically collocated. Tasks that the Ansible playbook currently does _not_ support include SSH key management, which should be included in any solution that manages multiple admin accounts.

Regardless of how we improve the coverage of the Ansible tasks, we're still going to have the problem of sharing config state between multiple Tails devices. Since the site-specific values are stored in `prod-specific.yml`, and are not version-controlled beyond the occasional `git stash`, synchronizing that state because quite difficult, and simply must happen out-of-band.

We're looking into migrating away from the painful `prod-specific.yml` workflow anyway, but synchronizing vars state across Tails devices will remain problematic. Git submodules for managing encrypted vars files, anyone?

freddymartinez9:
I can tackle the docs on this but a few questions that this issue doesn't seem to address. 

> That's all well and good, but it largely assumes that both admins will be working from the same Tails device

Why does this have to be the case? I assume because they need to have the same THS Auth cookie?  I am also not sure of a use case when you need multiple independent admins with separate SSH access. Anyways feel free to assign it to me.

fowlslegs:
>  I am also not sure of a use case when you need multiple independent admins with separate SSH access.

One admin cannot investigate some problem or OSSEC alert, or carry out some necessary upgrade process required to keep their instance up-to-date.

freddymartinez9:
Just to be clear @fowlslegs, we would have to have each ssh user to reuse the Onion Service Auth Cookie? So part of the setup is making sure each admin say `alice` and `bob` would have the same Cookie (I _assume_ so, I just want to be explicit) ...

fowlslegs:
[...]

PR #1206, opened 2015-11-19
Validate journalist's input to torrc additions

ageis:
* Adds a regular expression to match HidServAuth line onion address + auth-cookie 
* Loops until they get it in the correct format

regexp are direct from the Tor manual: https://www.torproject.org/docs/tor-manual.html.en#HidServAuth

### DIFF
--- a/tails_files/install.sh
+++ b/tails_files/install.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 # SecureDrop persistent setup script for Tails 
 
+set -e
+
 # check for root
 if [[ $EUID -ne 0 ]]; then
   echo "This script must be run as root" 1>&2
--- a/tails_files/install.sh
+++ b/tails_files/install.sh
@@ -109,7 +109,11 @@ chmod 755 $INSTALL_DIR/99-tor-reload.sh
 # journalist workstation does not have the *-aths files created by the Ansible playbook, so we must prompt
 # to get the interface .onion addresses to setup launchers, and for the HidServAuth info used by Tor
 if ! $ADMIN; then
-  HIDSERVAUTH=$(zenity --entry --title="Hidden service authentication setup" --width=600 --window-icon=$INSTALL_DIR/securedrop_icon.png --text="Enter the HidServAuth value to be added to /etc/tor/torrc:")
+  REGEX="^(HidServAuth [a-z2-7]{16}\.onion [A-Za-z0-9+/.]{22})"                   
+  while [[ ! "$HIDSERVAUTH" =~ $REGEX ]];                                         
+  do                                                                              
+    HIDSERVAUTH=$(zenity --entry --title="Hidden service authentication setup" --width=600 --window-icon=$INSTALL_DIR/securedrop_icon.png --text="Enter the HidServAuth value to be added to /etc/tor/torrc:")
+  done  
   echo $HIDSERVAUTH >> $ADDITIONS
   SRC=$(zenity --entry --title="Desktop shortcut setup" --window-icon=$INSTALL_DIR/securedrop_icon.png --text="Enter the Source Interface's .onion address:")
   SOURCE="${SRC#http://}"


ageis:
I added another feature to this branch - instead of prompting for the admin's SSH user, we'll fetch it from `prod-specific.yml`.

Also, we use `set -e` so the script will fail if there's a bad exit status on any subcommand, instead of proceeding.

tails_files/install.sh
```diff
@@ -1,5 +1,27 @@
...
+   awk -F$fs '{
+      indent = length($1)/2;
+      vname[indent] = $2;
+      for (i in vname) {if (i > indent) {delete vname[i]}}
+      if (length($3) > 0) {
+         vn=""; for (i=0; i<indent; i++) {vn=(vn)(vname[i])("_")}
+         printf("%s%s%s=\"%s\"\n", "'$prefix'",vn, $2, $3);
+      }
+   }'
+}
```
garrettr: Black fuckin magic...

conorsch: Having a brittle bash function to parse YAML to make idempotent changes seems silly, given that we're using Ansible to parse YAML vars and make idempotent changes on the servers. 

Food for thought: require Ansible on Journalist Workstation to make the Tails configuration more sane?

tails_files/install.sh
```diff
@@ -61,14 +83,19 @@ if $ADMIN; then
...
+    fi
     cat > $INSTALL_DIR/ssh_config <<EOL
 Host app
   Hostname $APPSSH
-  User $SSHUSER
+  User $ssh_users
 Host mon
   Hostname $MONSSH
-  User $SSHUSER
+  User $ssh_users
```
garrettr: @conorsch I think we've been over this before, so sorry if I'm forgetting the conclusion. What does Ansible do if `ssh_users` is a list? Is that allowed, or does it really expect only one username to be listed there?

My memory is that it only expects one username to be listed there, and the only reason we haven't renamed that field to `ssh_user` is because it's in the version-controlled `prod-specific.yml`, and so change it will be disruptive to the typical stash-pull-pop update workflow for admins.

tails_files/install.sh
ageis: Yeah, it's not ideal. But just want to note we used this function in our 0.3pre migration scripts with some success. https://github.com/freedomofpress/securedrop/blob/develop/migration_scripts/0.3pre/upgrade.sh

garrettr: I say it's good enough for now. Generally, I think the goal should be be to centralize the configuration management for the SecureDrop servers and the Tails workstations into one script/command/wizard.

tails_files/install.sh
garrettr: I ask because if `ssh_users` has more than one entry, the SSH config auto-setup will fail because SSH expects a single username here. This is sufficiently confusing that, if we cannot change `ssh_users` to `ssh_user` due to the "merge conflict on the Admin Workstation during update" issue, we should probably document this clearly.

conorsch: @garrettr Ansible will Do The Right Thing with `ssh_users`, since the associated tasks are already using `with_items` (see `roles/common/tasks/create_users.yml`). If the value of `ssh_users` is a string, rather than a list, then Ansible will interpret it as a list with length of 1 while processing `with_items`. If it's a YAML list, then it'll be a list of length *n*, where *n* is the number of usernames provided.

You can check this by updating the vars file and running `ANSIBLE_ARGS="-t users" vagrant provision /staging/`. The sudoers config is a static file, and not a template, so doesn't change regardless of how many admin users are created. The tasks that create the admin users add the usernames to the `ssh` and `sudo` groups, which makes sudo work correctly for all accounts.

Therefore I think it's appropriate to leave the variable named as `ssh_users`, since it does indeed accommodate multiple usernames.

garrettr: @ageis In that case, you should handle the case where there are more than one `SSH_USERS` in this script, otherwise people could end up with a broken SSH config.

ageis: Our docs currently tell people that both servers should have the same username.

This would be hard to do, because if ssh_users is a list in an arbitrary order, then you have to figure out which server it's for. Which means spawning an SSH process and checking whether it successfully connects using that username or not. Is that too complicated?

conorsch: @ageis Yes, the usernames should be the same on both servers _for a given Ansible run_, but technically any number of usernames can be granted admin privileges on the servers. Unfortunately we don't currently have a great plan for maintaining multiple support for multiple administrators—see #1175. 

@garrettr is right, though: `ssh_users` in the YAML vars file can be either a string or a list, so we must support both use cases.

garrettr:
This still needs a story to resolve the issue with more than one `ssh_users`.

ageis:
@garrettr what do you make of my last comment? "if ssh_users is a list in an arbitrary order, then you have to figure out which server it's for. Which means spawning an SSH process and checking whether it successfully connects using that username or not." That is the only way I can see implementing this the way you would like... I can do it, but it will increase set up time, complexity and likelihood of script timeout/failure. I think I should also put this work in a separate branch so we can just get the HidServAuth line regex validation in sooner.

Again, our docs say the username on both servers "must be the same" https://docs.securedrop.org/en/latest/servers.html which should preclude us from having to support this case.

ageis:
Validating the SSH config might not be such a bad idea. It would prevent us from installing a broken ~/.ssh/config.

garrettr:
@ageis What did you make of @conorsch's last comment? The fact that we tell people to set the same username on both servers is unrelated to this issue (it's an artifact of Ansible). The issue is that people might want to provision _multiple_ admins. One way they could do this is by making `ssh_users` a list instead of a string and re-running the playbooks (although I believe the existing admin would need to set up the new admin's public key manually via `ssh-copy-id`).

Note that this _works_ in Ansible but _does not_ work in your script, which will barf if it encounters a non-string for `ssh_users`. That is the core of the problem. The problem is exacerbated by the variable being named `ssh_users` instead of `ssh_user`, which suggests that it's perfectly reasonable to make it a list of usernames.

Again, as @conorsch already pointed out, multiple admin support has problems (#1175). That is why we do not discuss it in the documentation. We never actually planned to support it, but we've received enough requests from SecureDrop admins for it that we've been trying to figure out how to make it work on an ad-hoc basis. I just want to make sure we don't totally break SecureDrop for people like that until we have a chance to engineer a better solution from the ground up.

Again, this might not need to be a technical solution. It might be sufficient to add a note to the documentation explaining this limitation in the Tails setup scripts. That would be fine. I just don't want people who reasonably added more than one username to `ssh_users` to have this script trash their Admin Tails Workstation.

ageis:
@garrettr This PR has been modified to include only functionality specified in the title. We can handle the ssh_users issue separately.

Issue #1161, opened 2015-10-29
Diffie-Hellman group parameters for SSH


ageis:
`/etc/ssh/moduli` contains the default prime numbers that are bundled with OpenSSH on Ubuntu.

In light of recent research on [weak DH](https://weakdh.org/)[¹](https://weakdh.org/imperfect-forward-secrecy-ccs15.pdf)[ ²](https://freedom-to-tinker.com/blog/haldermanheninger/how-is-nsa-breaking-so-much-crypto/) we should probably think twice about using common primes, or including any 1024-bit primes at all—and generate our own moduli of 2048 bits or greater.

The risk here is mitigated by the extensive firewalling preventing direct SSH access to either server from the open internet, so we are relying additionally on the not-super-strong 1024-bit RSA cryptography in the Tor hidden service protocol which is [in the process of being upgraded](https://lists.torproject.org/pipermail/tor-dev/2015-October/009762.html). An unprecedented investment of massive computation resources capable of breaking both Tor and the SSH key exchange at the same time would be catastrophic.

Fortunately, the servers are preferring Curve25519 and `diffie-hellman-group-exchange-sha256` in `KexAlgorithms`[³](https://github.com/freedomofpress/securedrop/commit/eadce5a52ed7cdd6aaf1fcf0edc9fbb48f32ba76).

I know absolutely nothing about discrete logarithmics or field sieve analysis, but I will quote from the paper which speculated about how the NSA is able to decrypt a majority of connections through a break in Diffie-Hellman key exchange:

> Since a handful of primes are so widely reused, the payoff, in terms of connections they could decrypt, would be enormous. Breaking a single, common 1024-bit prime would allow NSA to passively decrypt connections to two-thirds of VPNs and a quarter of all SSH servers globally. Breaking a second 1024-bit prime would allow passive eavesdropping on connections to nearly 20% of the top million HTTPS websites. In other words, a one-time investment in massive computation would make it possible to eavesdrop on trillions of encrypted connections.

Additional links:
https://security.stackexchange.com/a/41947
http://entropux.net/article/openssh-moduli/

garrettr:
> Fortunately, the servers are preferring Curve25519 and SHA2 in KexAlgorithms³.

This only matters if the SSH clients that typically connect to them also prefer or at least support these KexAlgorithms. In the case of SecureDrop SSH, that's Tails on the Admin Workstation. Does the default SSH client configuration on Tails support these algos?

[...]

PR #1207, opened 2015-11-19
Remove weak DH groups of <= 2000 bits

ageis:
This is a first approach to the problem in #1161.

Generating new prime numbers and testing them with `ssh-keygen` takes a lot of time, so various guides across the internet have recommended keeping the default primes, but just deleting the weak ones from `/etc/ssh/moduli`.

So we're doing that using a common awk command based on whether we have a SHA1 checksum of the moduli file shipped with `openssh-client`.

My only misgiving about this method concerns registering the moduli into a variable (it's ~200 kilobytes). Maybe it should be using a temporary file instead?

### DIFF
--- a/install_files/ansible-base/roles/restrict-direct-access/tasks/dh_moduli.yml
+++ b/install_files/ansible-base/roles/restrict-direct-access/tasks/dh_moduli.yml
@@ -3,6 +3,8 @@
   stat:                                                                         
     path: /etc/ssh/moduli                                                       
   register: dh_moduli                                                           
+
+# SHA1 checksum represents the latest default primes/moduli shipped with OpenSSH
                                                                                 
 - name: Remove weak DH moduli                                                   
   command: awk '$5 > 2000' /etc/ssh/moduli                                      
--- a/install_files/ansible-base/roles/common/tasks/dh_moduli.yml
+++ b/install_files/ansible-base/roles/common/tasks/dh_moduli.yml
@@ -1,0 +1,19 @@
+---                                                                             
+- name: Check whether Diffie-Hellman groups have been updated                   
+  stat:                                                                         
+    path: /etc/ssh/moduli                                                       
+  register: dh_moduli                                                           
+                                                                                
+- name: Remove weak DH moduli                                                   
+  command: awk '$5 > 2000' /etc/ssh/moduli                                      
+  register: new_dh_moduli                                                       
+  when: dh_moduli.stat.checksum == '2dc7b3b18a2d3884ca4604466b75cf7903e96a97'   
+                                                                                
+- name: Install updated DH moduli                                               
+  copy:                                                                         
+    dest: /etc/ssh/moduli                                                       
+    content: "{{ new_dh_moduli.stdout }}"                                       
+    owner: root                                                                 
+    group: root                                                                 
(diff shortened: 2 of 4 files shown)


garrettr:
This looks good to me except for the way it's integrated with the rest of the Ansible roles. Logically, shouldn't this go with the rest of the common SSH configuration? Ideally, it would be part of some "common SSH hardening" role, which doesn't currently appear to exist but maybe one day it will. Instead, it's in `roles/common/tasks/dh_moduli.yml`, which I think is the wrong place because there is nothing else related to the SSH configuration in any of the other roles listed in `roles/common/tasks/main.yml`.

conorsch:
The `restrict-direct-access` role is probably the best place for now. Also, let's add comments that explain the checksums being used, for posterity.

ageis:
@garrettr @conorsch Made the requested change.

install_files/ansible-base/roles/restrict-direct-access/tasks/dh_moduli.yml
```diff
@@ -0,0 +1,21 @@
...
+  when: dh_moduli.stat.checksum == '2dc7b3b18a2d3884ca4604466b75cf7903e96a97'   
+                                                                                
+- name: Install updated DH moduli                                               
+  copy:                                                                         
+    dest: /etc/ssh/moduli                                                       
+    content: "{{ new_dh_moduli.stdout }}"                                       
+    owner: root                                                                 
+    group: root                                                                 
+    mode: "0644"                                                                
+  when: dh_moduli.stat.checksum == '2dc7b3b18a2d3884ca4604466b75cf7903e96a97'
```
garrettr: FYI, using SHA1 is fine for now but when it comes out, Ansible 2.0 will support a variety of hash functions including SHA-{256,512}: http://docs.ansible.com/ansible/stat_module.html#options. Sweet!

conorsch: In the meantime, there's nothing stopping us from registering the SHA256 sum and using that for the comparison. I don't think it's necessary here, though, particularly with the upstream changes inbound.

conorsch:
This PR is looking good. One nit: can we elaborate on the comments about the checksum? At first glance it's not clear what this task list is accomplishing. Go nuts with the comment text—anyone who takes the time to read the YAML in the Ansible tasks should be rewarded for the digging.

ageis:
@conorsch What kind of elaboration? Like the version of openssh-client that shipped this version of the file, whence it first appeared? A link to info about weak DH?

install_files/ansible-base/roles/restrict-direct-access/tasks/dh_moduli.yml
fowlslegs: There's no point really in using a cryptographic hash function here at all as this file has come from a signed GPG package--we just need a (non-cryptographic) hash function to see if it's already been modified or not for idempotency. I think this was the wrong way to do it as if there are updates to this file provided by a new version of openssh it will no longer function correctly. The first function could be a call to awk with the inequality reversed (i.e., `command: awk '$5 > 2000' /etc/ssh/moduli`), and the following two functions would be predicated on `when: has_small_primes.stdout`.

PR #1210, opened 2015-11-24
Even more gender-neutral language

mark-in:


garrettr:
Nice, appreciate this change. I agree that "they" should be preferred when a gender-neutral pronoun is required.

Issue #1123, opened 2015-09-21
OSSEC alternatives


flamsmark:
OSSEC is a hassle. It provides convoluted, opaque and hard-to-parse alerts by email (how 20th century). Is there something better that SecureDrop could use instead? It would be lovely if that something supported contemporary notions like on-call rotations.

It would also be particularly convenient one didn't have to understand SecureDrop's innards to parse a notification. Did the machine just reboot successfully? Was this a regular software update? I would like to be able to glance at a notification and know the answers to these sorts of questions. Currently, I have no idea.

conorsch:
OSSEC grates on me, as well. Making sense of the email alerts is a challenge, particularly for new admins. What I'd like to see is a total overhaul of the Monitor Server, with more modern and less painful tooling. A few options to consider:

* [osquery](https://osquery.io/)
* [artillery](https://github.com/trustedsec/artillery) (Functionality equivalent to OSSEC to my eye, plus optional honeypot)
* [MozDef](https://github.com/jeffbryner/mozdef) (Impressive, but likely overkill for our use case)
* custom ELK dashboard

The custom ELK option seems like a workable compromise between functionality and ease of use. We could even take the existing OSSEC config and pipe its logs into ELK as another data point. The primary obstacle right now is that the ELK workflow mostly expects admins to come to it—meaning we'd need to bolt on alerting functionality by way of another tool.

[Riemann](http://riemann.io/concepts.html) is a possibility for adding alerts, but I'm open to suggestions. One neat aspect of Riemann is that it leaves the door open for how alerts are sent—it doesn't assume they're going to be email. One reason we're using email now is the ability to encrypt the alert contents. Receiving an encrypted email alert is essentially a pop-up notification (depending on device settings, of course) that says "Something's up with SecureDrop!" and you need to decrypt the message—likely on a different device—in order to get the details. In theory, this is a nice setup—but OSSEC is noisy, and the false positives likely reduce confidence in its alerts.

conorsch:
Given our security requirements, specifically that we want alerts to be encrypted, we should consider using Signal for shipping the alerts. Even before we migrate away from OSSEC, refining the alerting logic in OSSEC and then providing a Signal transport, rather than GPG-encrypted email, would go a long way in making the monitoring setup more useful to Admins.

freddymartinez9:
There is a [signal-cli ](https://github.com/AsamK/signal-cli) tool that may be leveraged here.

conorsch:
[...]

Issue #976, opened 2015-04-02
Convert SecureDrop visualization diagram to an open format


garrettr:
Follow-up to #975. Ideally any and all diagrams would have the following properties:

1. Open format
2. Version-controllable (e.g. XML, not a binary format)
3. Easily editable by potential contributors, ideally with open-source software. [draw.io][1] is also a nice solution that we have been using for some other things.

[1]: http://draw.io/

mark-in:
This issue seems to be resolved by #1211.

conorsch:
@ghost Yup, that SVG you contributed certainly does the trick. Thanks!

PR #1211, opened 2015-11-24
Make the diagram vectorish, smoother and clearer

mark-in:
This is how it looks like:
securedrop


fowlslegs:
Nice work!

How did you create this? Ideally, this could solve #976.

A few things:

* OSSEC also sends alert emails from the monitor server to the Admin's email.
* The Admin also uses TBB on Tails to access the web interface of the SecureDrop application running on the App Server.
* This doesn't illustrate how journalists and sources can carry out conversations over SecureDrop. The arrows to and from the App Server should be bidirectional.

Also, we may be switching to recommending FreeOTP as a HOTP/ TOTP app because Google Authenticator closed its source a couple years back (at least I'm pushing for that). Do leave that as is for now though.

mark-in:
> Nice work! How did you create this? Ideally, this could solve #976.

Thanks. Just simple Inkscape and graphics from Tango project.

> A few things:
> * OSSEC also sends alert emails from the monitor server to the Admin's email.

Email notifications are super-duper extreme because it leaks tons of sensitive metadata. I think this must be optional and not be included in overview diagram.
I looks reasonable to store PGP encrypted (or maybe not) on special authenticated HS hosted on MS. Or to host simple IMAP server on MS via HS with push notifications. Anyway there must not be any third party involved.

> * The Admin also uses TBB on Tails to access the web interface of the SecureDrop application running on the App Server.

Fixed in 4d2d387.

> * This doesn't illustrate how journalists and sources can carry out conversations over SecureDrop. The arrows to and from the App Server should be bidirectional.

Fixed in d8b5a15.

> Also, we may be switching to recommending FreeOTP as a HOTP/ TOTP app because Google Authenticator closed its source a couple years back (at least I'm pushing for that). Do leave that as is for now though.

I don't like GA either for the same reason. Fixed in 2d96f9b.

mark-in:
Please notify me before merge. Maybe I need to save result diagram in Inkscape SVG (w/ Inkscape extentions) to make it more convenient to edit for Inkscape users.
Unfortunately this format leaks lots of metadata. That's why I pushed plain SVG for the first time.

garrettr:
Awesome work, @mark-in! This is a great improvement over our current diagram, and I'm really happy we'll be able to incorporate it in an open format (e.g. not Microsoft Visio, :rage:).

To me, this looks ready to merge, but can you explain a little more about the pros/cons of SVG vs. "Inkscape  SVG" first?

garrettr:
> Email notifications are super-duper extreme because it leaks tons of sensitive metadata. I think this must be optional and not be included in overview diagram.
I looks reasonable to store PGP encrypted (or maybe not) on special authenticated HS hosted on MS. Or to host simple IMAP server on MS via HS with push notifications. Anyway there must not be any third party involved.

A couple of notes on this:

* OSSEC email notifications are technically optional (you can install SecureDrop without setting them up), but this is not recommended (how else would you receive notifications of unusual or potentially suspicious behavior on your instance?)
* These are notifications from OSSEC, and are not directly based on the behavior of sources (which, I agree, *would* leak sensitive metadata). The directories that are used to store data received from sources are specifically ignored by OSSEC, so we avoid sending file integrity change alerts (or similar) that could leak info about source behavior via OSSEC alerts.
* While we encourage organizations to use their own mail server to relay the OSSEC alert emails, most organizations do not actually run their own mail server, and so either directly or indirectly use a third party (typically Gmail, as seen it is [specially documented](https://docs.securedrop.org/en/latest/ossec_alerts.html#using-gmail-for-ossec-alerts)) to relay these emails. We consider this trade-off to be acceptable because the emails are encrypted and we are careful to avoid them leaking metadata about source behavior, but also think this situation could be improved.

We are well aware of the current limitations of SecureDrop's monitoring and are working on improving it (see #1123). In the meantime, it is best for diagrams to reflect the *current* state of SecureDrop, rather than an ideal but not-yet-implemented state.

mark-in:
> can you explain a little more about the pros/cons of SVG vs. "Inkscape SVG" first?

In details it's explained at http://wiki.inkscape.org/wiki/index.php/Inkscape_SVG_vs._plain_SVG .
In short, "Inkscape svg is basically the same as plain svg, just with a few extra commands (in separate namespaces) added, which the Inkscape tools use to keep track of their work" and "should render identically with or without the inkscape extensions". I'm not sure if it's really useful but it can be.
The point is that there are a lot of metadata in these extentions (Inkscape version, name of the user, path to file on user's machine etc). It should not be included into project.
So the choice is between plain SVG and Inkscape SVG with extra data cut out (MAT doesn't support SVG). Now I tend to think that plain SVG is better for having less pain with cutting out metadata.

mark-in:
>    OSSEC email notifications are technically optional (you can install SecureDrop without setting them up), but this is not recommended (how else would you receive notifications of unusual or potentially suspicious behavior on your instance?)

Maybe Pond?
Alerts should be visible only to the admin who can manage to install and use Pond (on Tails).

> While we encourage organizations to use their own mail server to relay the OSSEC alert emails, most organizations do not actually run their own mail server, and so either directly or indirectly use a third party (typically Gmail, as seen it is specially documented) to relay these emails. We consider this trade-off to be acceptable because the emails are encrypted and we are careful to avoid them leaking metadata about source behavior, but also think this situation could be improved.

Alerts should also be resistant to traffic analysis against MS. If one is sending email notifications then it can easily leak the location of SecureDrop Area in clearnet to ISP. 
One can send emails via Tor but then we have to incorporate one more Area in our diagram (it will become sophisticated).

garrettr:
@mark-in Ok. In general, it seems like preserving *more* of people's work and work-in-progress is usually helpful, so I'd tend to prefer to use Inkscape SVG over normal SVG. For example, it appears only Inkscape SVG preserves layer information (I inferred this from reading various comments on this issue, and have not tested this myself, so I could be wrong), and layer information is generally very useful to encourage further work with a particular image.

> The point is that there are a lot of metadata in these extentions (Inkscape version, name of the user, path to file on user's machine etc). It should not be included into project.

Yeah, there's been an [issue](https://bugs.launchpad.net/inkscape/+bug/168958) for this on their issue tracker since 2007 and it still hasn't been fixed. :disappointed:

Given all of this, while I would prefer to use "Inkscape SVG", the metadata concerns are pretty annoying and I think it's probably best just to avoid them by using SVG instead. So this PR should be ready to merge, unless you have anything additional you want to add to this discussion!

garrettr:
Re https://github.com/freedomofpress/securedrop/pull/1211#issuecomment-160693100: if you feel strongly about this, consider filing a follow-up issue. The discussion in this pull request should be focused on the diagram you've submitted, and any additional ideas you have about SecureDrop's architecture are best discussed in separate, focused issues.

mark-in:
> Given all of this, while I would prefer to use "Inkscape SVG", the metadata concerns are pretty annoying and I think it's probably best just to avoid them by using SVG instead. So this PR should be ready to merge, unless you have anything additional you want to add to this discussion!

By the way there is some script (https://github.com/hartwork/svgstrip/) to strip private data from Inkscape SVG but it looks too messy and inconvenient for every-save usage (it also breaks formatting). I think that it's better to leave plain SVG for now.
Ready to merge.

mark-in:
Should I fix something else?

garrettr:
@mark-in Sorry, no! This is good to go, thanks for the reminder.

PR #1178, opened 2015-11-06
Various changes to verification processes

fowlslegs:
This commit changes parts of various files relating to the verification process
of keys and software use in and by SecureDrop. It is a follow up to pull request
1157 and issue #520. Changes include:

* Removal of redundant verification of key fingerprint (since GPG now does it
  implicitly).
* Admin is now explicitly instructed to download and verify Ubuntu Server over
  Tor, for greater privacy and security (see comments on #1157 for details).
* SKS hidden keyserver is now used instead of non-HKPS SKS pool for greater
  privacy and security (see comments on #1157 for details).
* Improved exposition on key fingerprints in admin-oriented documentation, w/o
  being over-technical or introducing concepts not elsewhere mentioned.
* Lastly, since a modification to the documentation on creating an Ubuntu
  installation medium was already in place, I went ahead and gave a recommended
  usage of the dd command.

Signed-off-by: Noah Vesely <fowlslegs@riseup.net>

PR #1212, opened 2015-11-24
Improve verification docs

fowlslegs:
This is a follow-up from #1178 addressing improvements to the verification documents.

### DIFF
--- a/docs/ossec_alerts.rst
+++ b/docs/ossec_alerts.rst
@@ -113,11 +113,14 @@ installation:
 
     cp ossec.pub install_files/ansible-base/
 
-The fingerprint is a unique identifier for an encryption key that is
-longer than, but contains both the shortid and longid. The value for
-``ossec_gpg_fpr`` must be the full 40-character GPG fingerprint for this
-same key, with all capital letters and no spaces. Here's a series of
-commands that will retrieve and format the fingerprint per our
+The fingerprint is a unique identifier for an encryption (public) key. It is the
+output of running a cryptographic hash function (in this case SHA1) over the
+key, represented in hexadecimal. The short and long key IDs correspond to the
+last 8 and 16 hexadecimal digits (or 32 and 64 bits) of the fingerprint,
+respectively, and are thus a subset of the fingerprint. The value for
+``ossec_gpg_fpr`` must be the full 40 hexadecimal digit (160 bit) GPG
+fingerprint for this same key, with all capital letters and no spaces. Here's a
+series of commands that will retrieve and format the fingerprint per our
 requirements:
 
 ::
--- a/docs/ossec_alerts.rst
+++ b/docs/ossec_alerts.rst
@@ -93,7 +93,7 @@ from a trusted and verified source, but you can also request it from keyservers
 using the known fingerprint. Looking it up by email address or a shorter key ID
 format could cause you to obtain a wrong, malicious, or expired key. Instead, we
 recommend you type out your fingerprint in groups of four (just like GPG prints
-it) enclosed by parentheses.  The reason we suggest this formatting for the
+it) enclosed by double quotes.  The reason we suggest this formatting for the
 fingerprint is simply because it's easiest to type and verify correctly. In the
 code below simply replace ``<fingerprint>`` with your full, space-separated
 fingerprint:
@@ -101,6 +101,10 @@ fingerprint:
 Download your key and import it into the local keyring: ::
 
     gpg --keyserver hkp://qdigse2yzvuglcix.onion --recv-key "<fingerprint>"
+
+.. note:: It is important you type this out correctly. If you are not
+          copy-pasting this command, we recommend you double-check you have
+          entered it correctly before pressing enter.
 
 Again, when passing the full public key fingerprint to the ``--recv-key`` command, GPG
 will implicitly verify that the fingerprint of the key received matches the
@@ -111,13 +115,9 @@ argument passed.
              the installation. If this happens, please email us at
              securedrop@freedom.press.
 
-Next we export the key to a local file. Here it is safe to use your email
-address associated with the key instead of your fingerprint since we are
-performing a local action and do not need to worry about network-based attacks
-as when fetching a key from a keyserver. Replace ``<email>`` with this email
-address below. ::
+Next we export the key to a local file. ::
 
-    gpg --export -a <email> > ossec.pub
+    gpg --export -a "<fingerprint>" > ossec.pub
 
 
 Copy the key to a directory where it's accessible by the SecureDrop
@@ -132,10 +132,7 @@ for ``ossec_gpg_fpr`` must be the full 40 hexadecimal digit GPG fingerprint for
 this same key, with all capital letters and no spaces. The following command
 will retrieve and format the fingerprint per our requirements: ::
 
-    gpg --with-colons --fingerprint <email> | grep "^fpr" | cut -d: -f10
-
-Again, we are using our email instead of our fingerprint as a key ID here
-because we are performing a local action.
+    gpg --with-colons --fingerprint "<fingerprint>" | grep "^fpr" | cut -d: -f10
 
 Next you specify the e-mail that you'll be sending alerts to, as
 ``ossec_alert_email``. This could be your work email, or an alias for a
--- a/docs/ossec_alerts.rst
+++ b/docs/ossec_alerts.rst
@@ -87,30 +87,37 @@ If you have your GPG public key handy, copy it to
 install\_files/ansible-base and then specify the filename in the
 ``ossec_alert_gpg_public_key`` line of prod-specific.yml.
 
-If you don't have your GPG key ready, you can run GnuPG on the command
-line in order to find, import and export your public key. It's best to
-copy the key from a trusted and verified source, but you can also
-request it from keyservers using the known fingerprint. Looking it up by
-email address or a shorter key ID format could cause you to obtain a
-wrong, malicious or expired key. Just replace the word ``<fingerprint>``
-in the following example with your own:
+If you don't have your GPG key ready, you can run GnuPG on the command line in
+order to find, import, and export your public key. It's best to copy the key
+from a trusted and verified source, but you can also request it from keyservers
+using the known fingerprint. Looking it up by email address or a shorter key ID
+format could cause you to obtain a wrong, malicious, or expired key. Instead, we
+recommend you type out your fingerprint in groups of four (just like GPG prints
+it) enclosed by parentheses.  The reason we suggest this formatting for the
+fingerprint is simply because it's easiest to type and verify correctly. In the
+code below simply replace ``<fingerprint>`` with your full, space-separated
+fingerprint:
 
 Download your key and import it into the local keyring: ::
 
-    gpg --keyserver <fingerprint> hkp://qdigse2yzvuglcix.onion --recv-key C5986B4F1257FFA86632CBA746181433FBB75451
+    gpg --keyserver hkp://qdigse2yzvuglcix.onion --recv-key "<fingerprint>"
 
-As in the setup for the *Admin Workstation*, we encourage you to first
-write out the fingerprint in groups of four, double-check you have
-entered it correctly, remove the spaces, and then press "enter" to
-actually fetch the key (and implicitly verify the fingerprint of the key
-received matches the one requested).
+Again, when passing the full public key fingerprint to the ``--recv-key`` command, GPG
+will implicitly verify that the fingerprint of the key received matches the
+argument passed.
 
-Next we export the key to a local file. It's okay to type the email
-address your key is associated with in the place of ``<fingerprint>`` in
-the command below. Since it is a local operation we don't have to worry
-about the network-based attacks we do during import from a keyserver. ::
+.. caution:: If GPG warns you that the fingerprint of the key received
+             does not match the one requested **do not** proceed with
+             the installation. If this happens, please email us at
+             securedrop@freedom.press.
 
-    gpg --export -a <fingerprint> > ossec.pub
+Next we export the key to a local file. Here it is safe to use your email
+address associated with the key instead of your fingerprint since we are
+performing a local action and do not need to worry about network-based attacks
+as when fetching a key from a keyserver. Replace ``<email>`` with this email
+address below. ::
+
+    gpg --export -a <email> > ossec.pub
 
 
 Copy the key to a directory where it's accessible by the SecureDrop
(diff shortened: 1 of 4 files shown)


fowlslegs:
Okay, I think this is finally ready to be merged. Thanks @freedomofpress/employees for helping me find what I believe presents a very good balance between making the documentation instructions here as clear and concise as possible, while also ensuring everything is done in a very safe way.

docs/ossec_alerts.rst
```diff
@@ -90,39 +90,44 @@ install\_files/ansible-base and then specify the filename in the
...
+request it from keyservers using the known fingerprint. Looking it up by
+email address or a shorter key ID format could cause you to obtain a
+wrong, malicious or expired key. Just replace the word ``<fingerprint>``
+in the following example with your own:
 
-Download your key and import it into the local keyring:
+Download your key and import it into the local keyring: ::
 
-::
+    gpg --keyserver <fingerprint> hkp://qdigse2yzvuglcix.onion --recv-key C5986B4F1257FFA86632CBA746181433FBB75451
```
garrettr: The previous paragraph says "replace the word `<fingerprint>` in the following example with your own", but then the example has a specific fingerprint (`C5986B4F1257FFA86632CBA746181433FBB75451`) instead of `<fingerprint>`. I would prefer `<fingerprint>` to avoid potential confusion.

docs/ossec_alerts.rst
```diff
@@ -90,39 +90,44 @@ install\_files/ansible-base and then specify the filename in the
...
 
-    gpg --recv-keys 0xFD720AD9EBA34B1C
+As in the setup for the *Admin Workstation*, we encourage you to first
+write out the fingerprint in groups of four, double-check you have
+entered it correctly, remove the spaces, and then press "enter" to
+actually fetch the key (and implicitly verify the fingerprint of the key
+received matches the one requested).
```
garrettr: Ouch, I understand why you're recommending this but this is a comically overwrought procedure is it not?

docs/ossec_alerts.rst
```diff
@@ -90,39 +90,44 @@ install\_files/ansible-base and then specify the filename in the
...
 
-Export the key to a file:
+Next we export the key to a local file. It's okay to type the email
+address your key is associated with in the place of ``<fingerprint>`` in
+the command below. Since it is a local operation we don't have to worry
+about the network-based attacks we do during import from a keyserver. ::
 
-::
+    gpg --export -a <fingerprint> > ossec.pub
```
garrettr: If you're going to allow the use of an email address here, I think you should replace `<fingerprint>` with `<keyid>`. Fingerprints are something specific, and it doesn't _really_ make sense to say that they are interchangeable with email addresses.

Key IDs, on the other hand, are defined as anything that can identify a specific key, which includes names, email addresses, truncated fingerprints ("short" and "long" key ids) and of course fingerprints.

docs/ossec_alerts.rst
```diff
@@ -90,39 +90,44 @@ install\_files/ansible-base and then specify the filename in the
...
-The fingerprint is a unique identifier for an encryption key that is
-longer than, but contains both the shortid and longid. The value for
-``ossec_gpg_fpr`` must be the full 40-character GPG fingerprint for this
-same key, with all capital letters and no spaces. Here's a series of
-commands that will retrieve and format the fingerprint per our
-requirements:
-
-::
+The fingerprint is a unique identifier for an encryption (public) key.
+the short and long key ids correspond to the last 8 and 16 hexadecimal
```
garrettr: Capitalize the first word of a sentence.

docs/set_up_admin_tails.rst
```diff
@@ -55,25 +55,27 @@ Start by running the following commands to download the git repository.
...
-    gpg --fingerprint B89A29DB2128160B8E4B1B4CBADDE0C7FC9F6818
+    gpg --keyserver hkp://qdigse2yzvuglcix.onion --recv-key B89A29DB2128160B8E4B1B4CBADDE0C7FC9F6818
 
-The Freedom of the Press Foundation Master Signing Key should have a
-fingerprint of ``B89A 29DB 2128 160B 8E4B 1B4C BADD E0C7 FC9F
-6818``.
+It is important you type this out correctly. If you are not copy-pasting
+it, we recommend typing the fingerprint out with spaces like so ``B89A
+29DB 2128 160B 8E4B 1B4C BADD E0C7 FC9F 6818``, double-checking you have
+it correct, and then removing the spaces before pressing "enter."
```
garrettr: I would move the earlier sentence:

> When passing the full public key fingerprint to the ``--recv-key`` command, GPG will implicitly verify that the fingerprint of the key received matches the argument passed.

to be an admonition after this paragraph.

docs/set_up_admin_tails.rst
```diff
@@ -55,25 +55,27 @@ Start by running the following commands to download the git repository.
...
 
-.. caution:: If the fingerprint does not match, fingerprint
-	     verification has failed and you **should not** proceed
-	     with the installation. If this happens, please contact us
-	     at securedrop@freedom.press.
+.. caution:: If GPG warns you that the fingerprint of the key received
+             does not match the one requested **do not** proceed with
+             the installation. If this happens, please email us at
+             securedrop@freedom.press.
```
garrettr: Does anybody know what this warning actually looks like?

fowlslegs: We'll need to look through the GPG source to find it I think.

garrettr: Yes, I was thinking it'd be easiest to look at the GPG source. Maybe find the issue that fixed fingerprint verification when receiving keyservers, and look at the patch that was landed to resolve the issue?

docs/ossec_alerts.rst
fowlslegs: I found it easier to type `C598 6B4F 1257 FFA8 6632 CBA7 4618 1433 FBB7 5451` and then remove the spaces than to type `C5986B4F1257FFA86632CBA746181433FBB75451`. It's better to copy-paste, but I found it very hard to accurately enter 40 hexadecimal digits manually--I got entered the Ubuntu Image key wrong the first time I entered it for example.

docs/ossec_alerts.rst
fowlslegs: Pedantic ;-)

docs/set_up_admin_tails.rst
fowlslegs: sgtm

docs/ossec_alerts.rst
fowlslegs: I feel like that requires an explanation of what we mean by a key ID. I've just replaced it with email in a way that I feel doesn't suggest email addresses and fingerprints are in interchangeable.

docs/ossec_alerts.rst
fowlslegs: I found out `gpg --recv-key "C598 6B4F 1257 FFA8 6632 CBA7 4618 1433 FBB7 5451"` works. This removes the lengthiness of that instruction while still ensuring the fp is entered correctly if manually typed.

garrettr: @fowlslegs What version of GPG are you using? This doesn't work for me on GPG 2.0.28.

```sh
▶ gpg --recv-key "C598 6B4F 1257 FFA8 6632 CBA7 4618 1433 FBB7 5451"
gpg: "C598 6B4F 1257 FFA8 6632 CBA7 4618 1433 FBB7 5451" not a key ID: skipping
```

fowlslegs: Can confirm it works on 1.4.1{8,9} and doesn't work on 2.0.29.

fowlslegs: Tested it on Tails and it works. It seems like this is going nowhere on the GPGv1 branch, which Tails is not moving off of any time soon. I think it would be fine to use this command for the time being if we file an issue upstream asking them to add it to GPGv2. Since even Jessie defaults to GPGv1, it's going to be at least 2 years before we might have to consider this again.

docs/set_up_admin_tails.rst
fowlslegs: [Here's](http://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg.git;a=blobdiff;f=g10/keyserver.c;h=dca5e18bf94c809938c5994d92179b67e4c4dbcf;hp=7bf9830645d32c9e3315ab4971a45b4c677cdb34;hb=5230304349490f31aa64ee2b69a8a2bc06bf7816;hpb=8eab483a1c4817a2946624c7305f464089d1875e) the original diff which added the `keyserver_retrieval_filter` function. [The patch](https://bugs.debian.org/cgi-bin/bugreport.cgi?att=1;bug=725411;msg=5;filename=0001-filter-and-verify-keyserver-responses.patch) Debian applied to the oldstable package seems to use this function in the original form (although it has since been rewritten). So when a bad key is given `keyserver_retrieval_filter` returns 1, which in turn causes the the `import_keys_stream` function on line 1502 of the 1.4.12 branch (unpatched) to fail. Not sure exactly what error message is propagated though.

The
```
/* TODO: Warn here if the fingerprint we got doesn't match the one
we asked for? */
```
warning is still present in the `keyserver_import_fprint` function (which starts on line 1675 on the unpatched 1.4.12 branch, is not changed by the patch). Seems like to you will get a generic error that does not specifically tell you the key you received has a fingerprint different than the one you request. What that generic error says exactly I'm unsure.

I feel like I've spent too much time on this as is.

docs/ossec_alerts.rst
fowlslegs: Fixed in upcoming revision.

fowlslegs:
So this last commit addresses most of your concern. Two notes:

1. I think using the spaced, parentheses-enclosed fingerprint strings is the way to go for readability, typability and is a safe choice based on the slow changing nature of Tails.
2. We still don't know what exactly the warning GPG throws at users when a malicious keyserver serves them the wrong key.

Also, we need to
- [ ] file a bug upstream with GPG to accept this fingerprint formatting.

fowlslegs:
So this last commit addresses most of your concern. Two notes:

1. I think using the spaced, parentheses-enclosed fingerprint strings is the way to go for readability, typability and is a safe choice based on the slow changing nature of Tails.
2. We still don't know what exactly the warning GPG throws at users when a malicious keyserver serves them the wrong key.

Also, we need to
- [ ] file a bug upstream with GPG to accept this fingerprint formatting.

docs/ossec_alerts.rst
```diff
@@ -87,42 +87,55 @@ If you have your GPG public key handy, copy it to
...
+If you don't have your GPG key ready, you can run GnuPG on the command line in
+order to find, import, and export your public key. It's best to copy the key
+from a trusted and verified source, but you can also request it from keyservers
+using the known fingerprint. Looking it up by email address or a shorter key ID
+format could cause you to obtain a wrong, malicious, or expired key. Instead, we
+recommend you type out your fingerprint in groups of four (just like GPG prints
+it) enclosed by parentheses.  The reason we suggest this formatting for the
+fingerprint is simply because it's easiest to type and verify correctly. In the
+code below simply replace ``<fingerprint>`` with your full, space-separated
+fingerprint:
```
garrettr: "enclosed by parentheses"? Did you mean double quotes?

docs/ossec_alerts.rst
```diff
@@ -87,42 +87,55 @@ If you have your GPG public key handy, copy it to
...
+             does not match the one requested **do not** proceed with
+             the installation. If this happens, please email us at
+             securedrop@freedom.press.
 
-::
+Next we export the key to a local file. Here it is safe to use your email
+address associated with the key instead of your fingerprint since we are
+performing a local action and do not need to worry about network-based attacks
+as when fetching a key from a keyserver. Replace ``<email>`` with this email
+address below. ::
```
garrettr: Ok, that's true, but do we need to add this complexity to the docs? Why not just stay consistent and recommend they use the fingerprint for everything?

docs/set_up_admin_tails.rst
```diff
@@ -55,25 +55,28 @@ Start by running the following commands to download the git repository.
...
-    gpg --keyserver pool.sks-keyservers.net --recv-key B89A29DB2128160B8E4B1B4CBADDE0C7FC9F6818
-    gpg --fingerprint B89A29DB2128160B8E4B1B4CBADDE0C7FC9F6818
+    gpg --keyserver hkp://qdigse2yzvuglcix.onion --recv-key "B89A 29DB 2128 160B 8E4B 1B4C BADD E0C7 FC9F 6818"
 
-The Freedom of the Press Foundation Master Signing Key should have a
-fingerprint of ``B89A 29DB 2128 160B 8E4B 1B4C BADD E0C7 FC9F
-6818``.
+It is important you type this out correctly. If you are not copy-pasting this
+command, we recommend you double check you have entered it correctly before
+pressing enter.
```
garrettr: This would be a good admonition.

docs/ubuntu_install.rst
```diff
@@ -1,74 +1,72 @@
...
-of Ubuntu on it, you may begin the Ubuntu installation on both
-SecureDrop servers.
+To create the Ubuntu installation media, you can either burn the ISO image to a
+CD-R or create a bootable USB stick.  As a reliable method we recommend using
+the ``dd`` command to copy the hybrid ISO directly to a USB drive rather than a
+utility like UNetbootin which can result in errors. Once you have a CD or USB
+with an ISO image of Ubuntu on it, you may begin the Ubuntu installation on both
+SecureDrop servers. If your USB is mapped to /dev/sdb, you would use dd like so: ::
+
+    sudo dd conv=fdatasync if=ubuntu-14.04.2-server-amd64.iso of=/dev/sdb
```
garrettr: If we're requiring this happens on Tails and we don't have to document how to do this cross-platform anymore, we should probably detail how to make sure you're writing to the right `of`, since this can be a destructive action.

docs/set_up_admin_tails.rst
fowlslegs: It already seems like one to me. Should I make it a `.. note::` along with the other places similar text is present?

docs/ossec_alerts.rst
fowlslegs: Yes.

docs/ossec_alerts.rst
fowlslegs: Sure. Fixed in next commit.

docs/ubuntu_install.rst
fowlslegs: Good point. I've added a couple lines on using `lsblk` to the upcoming commit.

docs/set_up_admin_tails.rst
fowlslegs: I've now made this a note following the --recv-key commands in all the files.

garrettr:
@fowlslegs This (finally!) lgtm. Thanks for being patient with my many reviews.

fowlslegs:
No worries. They were all valid concerns IMO :-)

Issue #525, opened 2014-08-08
source ui: codename is displayed in plain sight


Hainish:
When submitting documents, the codename is displayed in plain sight.  Also, when the source enters their codename, the text entered is fully visible.  This is especially troublesomesome since we're going to be advising sources to log in from a public internet connection.

These should be masked to conceal the text when it is entered and also when it is displayed on the submission page.

garrettr:
Closed by #1215.

PR #526, opened 2014-08-08
Mask codename

Hainish:
Fixes #525

Allow sources the option to log in with a new route, login_masked, which does not reveal their codename as they are typing it in.

Masks the codename on the submission page, revealing it with text select.

PR #1215, opened 2015-12-07
Mask codename on login and submission pages

fowlslegs:
This commit masks sources' codenames altogether as they enter them into the
login page. On the submission page, it hides the codename by default, but
allows the source to click a lock to reveal and hide the codename. The lock icon
also changes from closed to open (and back) as the source hides and reveals,
respectively, the codename.

Signed-off-by: Noah Vesely <fowlslegs@riseup.net>

### DIFF
--- a/securedrop/static/css/source.css
+++ b/securedrop/static/css/source.css
@@ -392,6 +392,31 @@ button.center {
   font-size: 18pt;
 }
 
+input#show {
+  display:none;
+}
+#content {
+    display: block;
+    transition: opacity 1s ease-out;
+    opacity: 0; 
+    height: 0;
+    font-size: 0;
+    overflow: hidden;
+}
+#show:before {
+    color:blue;
+    content: "Show"
+}
+#show:hover.show:before {
+    color:gray;
+    content: "Hide"
+}
+#show:hover ~ span#content {
+    opacity: 1;
+    font-size: 100%;
+    height: auto;
+}
+
 input.codename {
   width: 400px;
   padding: 10px;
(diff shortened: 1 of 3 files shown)


fowlslegs:
Not expecting this to be merged/ this to be the final form. Just wanted to show a possible solution to #526. My commit actually looks pretty good, but we might want to do some renaming/ reorganizing with the CSS and possibly rewording with the lookup.html text.

fowlslegs:
This addresses issue #525.

garrettr:
> This addresses issue #525.

@fowlslegs Github tip: https://github.com/blog/1506-closing-issues-via-pull-requests

securedrop/static/css/source.css
```diff
@@ -392,6 +392,32 @@ button.center {
   font-size: 18pt;
 }
 
+input#show {
+  display:none;
+}
+#content {
+    display: block;
+    -webkit-transition: opacity 1s ease-out;
```
garrettr: This is probably unnecessary, all major web browsers have supported unprefixed `transition:` since 2013: http://www.web-plus-plus.com/Articles/css-transition-moz-and-webkit-vs-css3. Additionally, we tend to focus on Tor Browser/Firefox, since that is the strongly recommended client to use when accessing SecureDrop sites, and `-webkit-transition` will have no effect there.

fowlslegs: Alright, I added this in then rebased.

garrettr:
Closes #525.

PR #1221, opened 2015-12-09
Converts Travis CI tests to use trusty64 with sudo

conorsch:
The Travis CI tests have been running on legacy infrastructure. Indeed, on each and every run Travis reminds us:

> This job ran on our legacy infrastructure. Please read our docs on how to upgrade. [0]

We've been dragging our feet on the upgrade, because the new Docker-based containers explicitly do not allow the use of sudo, which is a hard requirement for our tests. But sudo-less Docker containers are
not our only option (praise the old gods and the new)! We also have the option of using Ubuntu 14.04 LTS builds with sudo enabled [1], which is a much better option for our use case, given that we explicitly support 14.04 for production instances, and our development VMs also run 14.04.

[0] https://docs.travis-ci.com/user/migrating-from-legacy
[1] https://docs.travis-ci.com/user/trusty-ci-environment/

The new test suite ran fine over here: https://travis-ci.org/conorsch/securedrop/builds/95879540 So I say let's enable it project-wide on the `develop` branch. I accept full responsibility for debugging the new test suite—I've had my head buried in tests lately anyway, so now's a fine time to make this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,12 @@
+# Use Ubuntu 14.04 LTS images, and explicitly require sudo.
+# Relevant docs: https://docs.travis-ci.com/user/trusty-ci-environment/
+sudo: required
+dist: trusty
+
 language: python
 python:
 - '2.7'
+
 virtualenv:
   system_site_packages: true
 before_install:


garrettr:
@conorsch What are the benefits of this change? Besides:

1. Faster CI runs
2. No more annoying warning messages about legacy infrastructure

conorsch:
@garrettr The change may not even speed up our builds noticeably—will need to gather more data and compare. I see the benefits as:

  1. Migrating away from legacy test infrastructure, which includes:
    * "No more annoying warning messages"
    * *Potentially* more reliable builds in Travis
  2. Standardizing on 14.04 for production, development, and tests.

The reasons _not_ to make this change are:

  1. Maybe the tests will fail more, and I don't want to debug that mess.

That last point definitely carries weight, since we've already burned a non-trivial amount of time debugging Travis, but the the fact that we've had problems in the past suggests to me that we should pay attention to iterations the Travis folks are making that may improve reliability for the service—which leads me to conclude that the change is a Good Thing.

garrettr:
Fortunately, this is easy to revert. I'm going to merge it now so we can so we can try it out. If we get an increased number of build failures, we can always revert it.

Issue #1023, opened 2015-05-11
Improve idempotence of Ansible playbooks


conorsch:
The provisioning process reports several false positives regarding changed state. In short, the Ansible playbooks are not entirely idempotent. Here's a summary of `changed` tasks after a _second_ run of the playbooks on VirtualBox hosts:

```
PLAY RECAP ******************************************************************** 
development                : ok=29   changed=5    unreachable=0    failed=0
app-staging                : ok=114  changed=29   unreachable=0    failed=0   
mon-staging                : ok=80   changed=21   unreachable=0    failed=0
build                      : ok=20   changed=14   unreachable=0    failed=0
```

Ideally, running the playbooks a second time would yield `changed=0` on all hosts. More realistically, we should try to get that number down to an informative value. Reducing it to 5-10% (from >20%) would be swell. During a normal provisioning run, there are many expected changes to hosts, so we'll need to consider carefully what constitutes a "changed" state:
- firewall rules opening, then closing (for authd registration)
- performing backup tasks, and fetching backup files (see also #942)
- reboots and associated wait_for tasks
## Improving idempotence

There are a few areas to focus on for improving Ansible idempotence meaningfully.
### Overzealous use of `shell` and `command` modules

By default the `shell` and `command` modules **always** return `changed=true`, partly to encourage the use of more structured modules, and partly because it's really the only sane default behavior. Consider adding custom `fail/when` tasks, or appending `changed_when` parameters to existing tasks to check for exit codes. Wherever possible, avoid use of `shell` and `command` in favor of more structured modules (e.g. `apt`).
### Specific regexp for `lineinfile` calls

If the regular expression for `lineinfile` additions isn't properly structured, Ansible will blithely append statements to files all day long. Examples:
- `/etc/default/haveged`
- `/etc/hosts`

A related case is the `lineinfile` call always reporting changed, e.g.:
- `/etc/network/iptables/rules_v4` (there are duplicate lines in here too)
- fetched `*ths*` files (local to admin workstation)
### Implications for testing

It may be worthwhile to triage idempotence checks by role. At present the test suite doesn't play well with isolating specific roles, but #1004 exposed individual roles via tags. Consider breaking out roles post-reorg and testing idempotence separately.
### List of changes
- [x]  `/etc/default/haveged`
- [x] `/etc/hosts`
- [x]  fetched `*ths*` files (local to admin workstation)
- [x] importing gpg keys
- [x] all `command` or `shell` invocations
[...]

PR #1225, opened 2015-12-11
General idempotence changes for Ansible tasks

conorsch:
Makes a variety of Ansible changes idempotent, including:

* paxctl flags
* OSSEC registration
* grub-reboot target
* iptables management

Uses group_vars to manage staging-only attributes that determine, for example, whether direct access over SSH is enabled. Consolidates the iptables rules into a single template (with exception of needing to register an ossec-agent, which still inserts special rules). The consolidation allows for culling several now-superfluous roles:

* allow-direct-access (folded into restrict-direct-access, managed by `defaults` and `group_vars`)
* remove-authd-exemptions (folded into restrict-direct-access and ossec-agent, which now uses `delegate_to` to clean up after the `authd` process)

See #1023 for progress tracking of the playbook idempotence. 

Also corrects file permissions on two OSSEC-related files, and updates the corresponding Serverspec tests. See #1006 for tracking progress on all file permissions fixes.


install_files/ansible-base/roles/app-test/tasks/apparmor_complain.yml
```diff
@@ -1,10 +1,49 @@
...
-# the aa-status would be painful.
+# The `aa-status` command does not support machine-readable output,
+# so we must parse it with sed. This is a read-only task that will
+# display only the profiles listed in "complain" mode. The results
+# of this check are used to conditionally set the profiles to complain
+# mode only if necessary, to ensure idempotence.
+
+# Unfortunately the `aa-complain` command does not yield useful output
+# sufficient for determining changed state, but rather announces
+# "Setting <profile> to complain mode" every time, which is inane.
```
garrettr: :joy:

install_files/ansible-base/roles/app-test/tasks/apparmor_complain.yml
```diff
@@ -1,10 +1,49 @@
...
+  # So let's check both the exit code and STDOUT to fail appropriately.
+  failed_when: apparmor_complaining_profiles_result.rc != 0 or
+               'does not exist' in apparmor_complaining_profiles_result.stdout
+
 - name: Set AppArmor profiles to complain mode.
   command: aa-complain /etc/apparmor.d/{{ item }}
+  # Using multiline YAML syntax ('>') to avoid excessive escaping of quotes in the conditional.
+  # Using `regex_replace` to substitute newlines because the AppArmor profile names are
+  # dot-separated, but `aa-status` the executable path in place of the filename.
+  # Doing it this way keeps the config DRY and will work with future profiles as well.
```
garrettr: I :heart: these comments

install_files/ansible-base/roles/grsecurity/tasks/paxctl.yml
```diff
@@ -0,0 +1,36 @@
...
+  # The desired flags should include:
+  #   - p NOPAGEEXEC
+  #   - m NOMPROTECT
+  #   - x NORANDEXEC
+  #   - e NOEMUTRAMP
+  # After running `paxctl -Cpm` in VirtualBox, all four of those flags are set.
+  # However, the current paxctl command only enforces p and m.
+  # The x and e flags are present by default, but can be removed via
+  # `paxctl -z <binary>`. TODO: We may want to update the command
+  # to set -Cpmxe explicitly. Warrants further testing on hardware.
```
garrettr: File a separate issue?

garrettr: We should probably revisit these Pax flags generally. My usual stance is to try to avoid removing any of grsecurity's protections unless it is absolutely necessary. Sometimes it is necessary, for example if the program needs to be rewritten to understand the error code grsec returns from an mmap for WX, or if it uses trampolined functions, etc.

The Pax rules for GRUB were cargo-culted from an old blog post about getting grsec working on Ubuntu, IIRC. It is possible that grub has since been updated such that they are no longer necessary, in which case we should allow it to take advantage of the full set of protections afforded by grsecurity.

garrettr:
This lgtm so I'm going to merge it, but please take a look at my comments anyway :grin:

install_files/ansible-base/roles/grsecurity/tasks/paxctl.yml
ageis: For what it's worth, the default paxctld configuration file sets flags for the following binaries:
```
/usr/bin/grub-script-check      E
/usr/bin/grub-bios-setup        E
/usr/sbin/grub-mkdevicemap      E
/usr/sbin/grub-probe            E
```
However, this is enabling EMUTRAP instead of disabling MPROTECT. I would like to learn more about the distinctions (trampolines, etc.) and why this is recommended. I've found in testing that these GRUB utilities don't work on that that configuration, ex. installing a new kernel will fail on the post-install hooks, but they do run without any PaX flags. In other words, the flags recommended by paxctld.conf cause breakage. Same for Python - enabling EMUTRAP is not enough, although this is what's recommended instead of going all out on disabling MPROTECT. We do still need to get to the bottom of this.

Issue #1233, opened 2015-12-19
Rewrite static/js/source.js removing jquery requirement for source interface


evilaliv3:
It seems to me that currently on the source interface jQuery is included but required only by: https://github.com/freedomofpress/securedrop/blob/develop/securedrop/static/js/source.js

It would be valuable to rewrite such library so to remove jQuery requirement.

From a small analysis few functions would be needed:

1) a function to show/hide html elements; this is doable in standard javascript with:

``` javascript
          document.getElementById("element_id").style.display = "none";  // hide
          document.getElementById("element_id").style.display = "block"; // show
```

2) a function for adding a class to html elements; this is doable in standard javascript with:

``` javascript
          document.getElementById("element_id").className += "additional_css_class";
```

3) fading/fadout functions; not immediatly doable but can it's fancy-only stuff that can be either removed or be implemented in CSS

redshiftzero:
Once #2405 is addressed, this can be picked back up based on the work so far in PR #1298 branch `kaizensoze-remove-jquery`

PR #1234, opened 2015-12-19
Upgrade jquery from 2.1.1 to 2.1.4

evilaliv3:
This commit updates jQuery dependency to latest minor revision.

It would be valuable start tracking that version in the build scripts in order to be sure to not using outdated  buggy versions.

The same is recommended also for the fontawesome inclusion.


garrettr:
Hey @evilaliv3, thanks for the PR (and good to see you, virtually, anyway!).

Are there any security issues fixed between this version of JQuery and the one currently in the tree?

In general, I think I'd prefer to just implement your suggestion from #1233 and get rid of the JQuery dependency instead of upgrading it, but we could also upgrade it for now until #1233 is implemented.

evilaliv3:
no as far as i know no security issue is mentioned.

yep with my suggestion it is possible to remove jquery from the source interface but what about the journalist interface, there javascript is currently used for other things?

garrettr:
@evilaliv3 Of course, good point!

PR #1237, opened 2016-01-04
Document 2FA setup in install guide

### LINKED ISSUE #1219: Document google-authenticator setup for login via tty
I had he problem that after the playbooks ran through smoothly I wasn't able 
to access mon and app via the .onion adresses. I tried logging in via tty   
and failed. I figured out that google-authenticator needs to be configured to
be able to log in via tty (which is supposed to happen after logging in via .onion for the first time ). maybe mentioning this as a note in the admin guide would be good for future references. tnx

conorsch:
Makes the 2FA setup more visible for first-time admins. Specifically:

* adds a link to the **Google Authenticator** Topic Guide to the end of the **Install SecureDrop** page
* adds steps to test 2FA over TTY to the **Test the installation** page

Also adds brief mention of FreeOTP as an alternative software solution. Resolves #1219.


ageis:
LGTM. I had to think about why you would recommend 2FA for TTY instead of SSH, but it makes sense. If the admin leaves their password lying around on a piece of paper, and a physical intruder gets access to the hardware, they'd need their phone too. Whereas access to the admin workstation + Persistence should only occur in admin's presence, then be powered off.

However, assuming you can get a TTY, couldn't you just boot in single user mode?

garrettr:
These changes are good, but agree with @ageis there is a broader question about whether we want to require 2FA for TTY login.

> However, assuming you can get a TTY, couldn't you just boot in single user mode?

Yes, you can. There are ways to restrict that (FDE, other stuff), but they all have complex tradeoffs for availability and we discourage their use. Frankly, the ability to rescue a local system via single user mode has been a lifesaver on more than one occasion.

As a result, I think the 2-Factor for TTY is security theater that only makes the admin's life harder. We should allow them to login via local TTY with just a password. If we want to make it harder for attackers with physical access to attack the boxes, then we should approach that in a holistic way at a later date.

garrettr:
Reviewed during issue triage session. Decided that this is a substantial improvement to the current documentation, and so should be merged.

We also agreed that 2FA for the admin's console login is overkill, and is more harmful to availability than it is beneficial for security. @conorsch is opening a followup issue to reconsider the required use of 2FA for console login.

PR #1239, opened 2016-01-19
Update deployment practices guide with nginx examples

ageis:


garrettr:
In general, I think these instructions might be a little bit _too_ specific. We always struggle with defining our audience in our documentation. In general, I would hope that an admin who is interested in installing SecureDrop is also capable of deploying TLS without detailed guidance from us. If you can't deploy TLS, you probably shouldn't be deploying SecureDrop.

That being said, this is a clear improvement over the current documentation, and this type of explicit instruction is already the norm in the documentation. I say we merge this for now, but generally keep the audience issue I just brought up in mind for future work on the documentation.

Issue #1241, opened 2016-01-27
Bug in Vagrant 1.8.1 breaks provisioning VM groups


ageis:
The example command `vagrant up /staging$/` from our development docs is currently broken in Vagrant 1.8.1 on both Linux and OSX.

Issue filed upstream here: https://github.com/mitchellh/vagrant/issues/6828

Filing this as an acknowledgment of the issue in case any one else encounters it; when it gets fixed upstream we can close this out, but possibly include a note advising against using that Vagrant version.

conorsch:
Looks like this is an upstream issue that isn't reliably reproduceable. For instance, I can't reproduce it on my machine, and I'm running Vagrant 1.8.1, and the same is true for one of the maintainers. If you're affected, @ageis, provide some debug logs and platform information on the upstream issue and we'll compare notes.

garrettr:
I'm also having this problem with Vagrant 1.8.1 on Mac, so I did what @conorsch suggested and commented on the upstream bug with my debug logs: https://github.com/mitchellh/vagrant/issues/6828#issuecomment-175747531

garrettr:
Found a workaround in one of the various issues on the upstream Vagrant repo. If you specify `--no-install-provider`, multimachine provisioning with regex works again. e.g.

```
vagrant up --no-install-provider /staging$/
```

works for me, event hough `vagrant up /staging$/` does not.

`¯\_(ツ)_/¯`

conorsch:
Astounding though it is, this regression is still live in the most recent versions of vagrant. To keep things simple, we can just update the examples in the dev docs to use full machine names, rather than regex, e.g. `vagrant up --no-provision app-staging mon-staging`.

garrettr:
@conorsch Yes, I sadly agree that would be the best (for now). I asked if there's an ETA for Vagrant 1.8.2, which includes a fix for this, on the upstream repo: https://github.com/mitchellh/vagrant/issues/6828#issuecomment-218876744.

conorsch:
@fowlsegs was bitten by this recently. We should remove all regexp-formatted vagrant commands from the dev docs and use explicit host groups instead. I don't really care if Hashicorp patch this; let's just stick with the explicit version, since that seems more stable.

Another reason we want to use the explicit version: since Vagrant 1.8(ish), provisioning happens per-host even if not all target hosts have been created. That's foolish. @fowlslegs mentioned this work around in #1304:

```
[...]

PR #1242, opened 2016-01-27
Update Apache access control directives for version 2.4

ageis:
The old Allow, Deny and Order directives are deprecated: https://httpd.apache.org/docs/2.4/howto/access.html

I haven't had a chance to test this in the staging environment yet due to #1241.

### DIFF
--- a/install_files/ansible-base/roles/app/templates/sites-available/document.conf
+++ b/install_files/ansible-base/roles/app/templates/sites-available/document.conf
@@ -42,7 +42,11 @@ ErrorDocument 500 /error
   Options {{ apache_dir_options | default('None') }}
   AllowOverride None
   <Limit GET POST HEAD>
+{% if apache_allow_from == "all" %}
+    Require all granted
+{% else %}
     Require ip {{ apache_allow_from | default('127.0.0.1') }}
+{% endif %}
   </Limit>
   <LimitExcept GET POST HEAD>
     Require all denied
@@ -53,7 +57,11 @@ ErrorDocument 500 /error
   Options {{ apache_dir_options | default('None') }}
   AllowOverride None
   <Limit GET POST HEAD>
+{% if apache_allow_from == "all" %}
+    Require all granted
+{% else %}
     Require ip {{ apache_allow_from | default('127.0.0.1') }}
+{% endif %}
   </Limit>
   <LimitExcept GET POST HEAD>
     Require all denied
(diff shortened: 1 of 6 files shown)


conorsch:
> I haven't had a chance to test this in the staging environment yet due to #1241.

You can run `vagrant up app-staging mon-staging` rather than `vagrant up /staging$/` as a workaround.

garrettr:
@ageis This needs to be tested before it can be merged. @conorsch provided a workaround for multimachine provisioning in the last comment, there's another workaround [here](https://github.com/freedomofpress/securedrop/issues/1241#issuecomment-191983630).

conorsch:
Also required for merge will be updated serverspec tests for the apache config. Should be a similarly minor change to what's presented in the PR, to be performed after manual testing of the config update.

ageis:
@conorsch If you didn't notice (as your previous comment somewhat suggests), the updated spec tests were in the original commit. I just rebased on the latest develop. 

This PR has one very glaring problem. Per the `apache_allow_from` variable, "all" is not a valid IP address and will cause the Apache2 service to outright fail. More rewriting needed.

ageis:
In the prod environment, you want `Require ip 127.0.0.1`. In staging, you want `Require all granted`. Since these directives are very different and don't work as well with the `apache_allow_from` var, I am not sure of the best approach yet. Perhaps it's to include a conditional in the config template itself.

ageis:
Ok, I fixed that issue with a conditional in the template, and tested it with both values of "all" and "127.0.0.1" with positive results (in the latter/production case, the interfaces are Tor-only / not available via the app-staging VM's forwarded ports), and the changes themselves have been tested as valid and behaving as expected. Now it remains to fix the spec tests. I am still a beginner in RSpec language so I'm not sure how to express that. It would be very simple if someone could tell me whether the Serverspec tests are ever run in production, cause if they are only for staging, then we can just run with the expected line in staging.

ageis:
@conorsch, are the spec tests supposed to work in both environments or what?

conorsch:
@ageis Pending merge of #1387, most spectests should pass on the develop branch, applied to any of the VMs.

ageis:
I'm interpreting that as meaning spec tests should pass in the prod VM. I will let you know when I've fixed the tests and this PR is ready to be merged.

ageis:
Fixed up tests by adding new per-environment spec_tests var for the custom line. This should be good to go.

`bundle exec rake spec` passed with two unrelated failures. As an aside, curious why this defaults to development and you can't run `bundle exec rake spec:staging`.

fowlslegs:
Fixes #1607.

fowlslegs:
Closing for reasons listed in https://github.com/freedomofpress/securedrop/issues/1607#issuecomment-296847513.

PR #1243, opened 2016-01-28
Add note about using location directive

ageis:
Added note about using location directives to put landing-page specific configuration on the same domain/site.


garrettr:
lgtm

PR #1246, opened 2016-02-04
Update Hi-DPI displays note

KwadroNaut:
Tails 2.0 release is out since a week, the bug about high resolution displays (https://labs.riseup.net/code/issues/8659 ) is resolved. Only a minor issue remains.

### DIFF
--- a/docs/hardware.rst
+++ b/docs/hardware.rst
@@ -256,14 +256,6 @@ any of the new generation of NUCs, which have names starting with "NUC5"
 These NUCs have wireless networking built into the motherboard, and it
 is impossible to physically remove.
 
-A note about Hi-DPI displays
-''''''''''''''''''''''''''''
-Older versions of Tails (before 2.0, based on Debian 8), had problems to scale the
-resolution correctly. This has mostly been fixed in current versions. You might want
-to use the addon Zoom Menu Elements if you find the Tor browser toolbar too small.
-Another option is avoiding laptops that use this type of display.
-Examples are MacBook/MacBook Pros with the Retina display, or the Dell Precision M3800.
-
 Tails USBs
 ~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


fowlslegs:
Even if Tor Browser may block enumeration of plugins (anti-fingerprinting), there could be other detectable or otherwise unexpected behavior of a 3rd-party addon. Therefore, we cannot recommend the installation of the addon you mention. If you want to change your commit to instead suggest people go into `about:config` and instead set `layout.css.devPixelsPerPx` to a value between 1.1 to 1.5, I would be happy to merge it.

garrettr:
@KwadroNaut I haven't had a chance to test Tails 2.0 yet, is there a problem with the Tor Browser toolbar is small and hard to use on Hi-DPI displays?

KwadroNaut:
I've been getting conflicting feedback regarding such displays. @fowlslegs comment is very important and true (don't install plugins). Maybe, until someone else comes by and tests, it's better to change the wording to prefer non High-DPI displays and give feedback if you *do* use them. I'll adapt the wording accordingly tomorrow.

garrettr:
How about we just remove the "A note about Hi-DPI displays" section entirely?

conorsch:
> How about we just remove the "A note about Hi-DPI displays" section entirely?

A voice of reason! This sounds like the best choice, although I hesitate to implement the change until I've seen the hidpi problem resolved with my own eyes.

garrettr:
@conorsch It's marked as resolved on the upstream Tails issue tracker as of 2.0: https://labs.riseup.net/code/issues/8659. I have confirmed that Tails 2.x has a reasonable screen resolution on my MacBook Pro with a Retina display (there are a few little pieces of the UI that appear kinda small, but overall it's pretty good).

KwadroNaut:
Sounds good.

garrettr:
Great, thanks for updating the PR @KwadroNaut! :+1:

Issue #1249, opened 2016-03-05
Docs as described may lead to DNS lookup failures


heartsucker:
A fresh Ubuntu install is defaulting to have the network gateway as the only DNS server in `resolv.conf`. Adding the Google DNS servers `8.8.8.8` and `8.8.4.4` as `nameserver`s to `/etc/resolvconf/resolv.conf.d/tail` should be added to the docs. Currently the firewall rules listed in the docs only allow DNS lookups to the external DNS servers.

ehartsuyker:
Ah, it seems the docs do say this in the "Ubuntu Installation Guide." However, since this doesn't seem like a place some one would necessarily look when reading the "Set up the Servers" part of the guide (since I'm guessing most people who are doing this could install Ubuntu in their sleep), I'm going to merge those two since several things like FDE and the lack of correct DNS entries will **definitely** break the SD setup. Expect a PR later.

conorsch:
As mentioned in #1254, the Ansible playbooks do handle the DNS configuration on the hosts—see `roles/common/tasks/harden_dns.yml`. Provided the hardware firewall is configured correctly, the DNS config should be settled early enough in the Ansible run to ensure resolution during lookups for new packages and whatnot.

conorsch:
The playbooks should be updated to ensure that the ntp package installation occurs before adding the FPF repo. If the FPF repo was added, the `apt_repository` module will automatically trigger an `apt-get update`, which could fail if the system time is wrong.

conorsch:
The changes in #2714 added assert statements to validate that DNS is working, removing the need for Admins to run the verification commands manually as a debugging step. As written, those new checks would resolve this issue entirely. 

However, in #2741, we prioritized a fail-fast-on-grsec strategy, which prevents the checks from usefully validating DNS early in the install process, since the `common` role now runs _after_ the FPF repo install logic again. So further action still required here. We could:

1. Break out the sanity checks into a tiny role and place that at the top of the first play, to balance "fail fast with no DNS" against "fail fast on incompatible kernels.
2. Reorder the roles to run "common" first. Common's a big role, so doing so would essentially revert the changes presented in #2741. 

Slight preference for 1.

eloquence:
Sounds good! If I don't hear otherwise by tomorrow, will break #1 into a new task we can take on in the next sprint.

eloquence:
Per above, closing this issue and focusing on the tightly scoped #3203.

Issue #1244, opened 2016-02-02
Recommended hardware will likely brick itself on first boot


heartsucker:
The Intel NUC D54250WYK has a BIOS issue that will cause the unit to brick themselves if they attempt to suspend. Which I found out the hard way by bricking the two servers I just bought.

https://communities.intel.com/thread/76708?start=0&tstart=0
https://communities.intel.com/message/359708#359708

Can someone add this to the documentation? (I might do it later when I'm not busy dis/re-assembling these)

conorsch:
@ehartsuyker Ouch, thanks for the report. Were you intentionally suspending the servers, or did they reach sleep state accidentally, for example due to a power button bump? We'll certainly add mention of this to the docs, but more background on your situation may prove be useful for making recommendations for first-time admins.

ehartsuyker:
This happened because I pushed the power button to power down, and the (complete lack of a) manual that comes with it does not specify that the power button causes them to enter suspend as opposed to powering off. However, from what forums are saying this can happen a number of ways. One user said he averaged something like only 6 days of uptime between brickings.

garrettr:
Holy crap, this is a terrible issue! Thanks for bringing it to our attention, @ehartsuyker. I've done a lot of messing around/rebooting by pressing the power button on NUCs, and never encountered something like this - maybe it's only an issue on newer models? Were you able to reset the BIOS to recover the devices, as described in some of the links you provided?

conorsch:
I count myself very lucky never to have experienced this one myself. Based on the documentation @ehartsuyker provided, though, it's definitely something we should document (and #1254 includes docs updates).

ehartsuyker:
@garrettr Yes, I was able to open them and reset the BIOS by unplugging the battery, so this is more a bricking where they won't power on, and not a brick that completely ruins the motherboard. I also updated the BIOS and have adopted the strategy of "hold the power button until it is definitely off." I haven't have the issue again, but I'm also probably not going to attempt to reproduce it. :)

conorsch:
Resolved via #1254.

Issue #1247, opened 2016-03-03
Unable to install SecureDrop because of errors with https://apt.freedom.press


ehartsuyker:
Full `ansible` output. I'll be email this to y'all too since this might be a security problem.

```
amnesia@amnesia:~/Persistent/securedrop/install_files/ansible-base$ ansible-playbook -i inventory -u berlinleaks -K --sudo securedrop-prod.yml
sudo password:

PLAY [mon;app;mon-prod;app-prod] **********************************************

GATHERING FACTS ***************************************************************
ok: [mon]
ok: [app]

TASK: [install_fpf_repo | install fpf apt signing gpg key] ********************
changed: [mon]
changed: [app]

TASK: [install_fpf_repo | setup fpf apt repo] *********************************
failed: [mon] => {"failed": true, "parsed": false}
invalid output was: Traceback (most recent call last):
  File "<stdin>", line 2207, in <module>
  File "<stdin>", line 430, in main
  File "/usr/lib/python2.7/dist-packages/apt/cache.py", line 440, in update
    raise FetchFailedException(e)
apt.cache.FetchFailedException: W:Failed to fetch https://apt.freedom.press/dists/trusty/main/binary-amd64/Packages  server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
, E:Some index files failed to download. They have been ignored, or old ones used instead.

failed: [app] => {"failed": true, "parsed": false}
invalid output was: Traceback (most recent call last):
  File "<stdin>", line 2207, in <module>
  File "<stdin>", line 430, in main
  File "/usr/lib/python2.7/dist-packages/apt/cache.py", line 440, in update
    raise FetchFailedException(e)
apt.cache.FetchFailedException: W:Failed to fetch https://apt.freedom.press/dists/trusty/main/binary-amd64/Packages  server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
, E:Some index files failed to download. They have been ignored, or old ones used instead.


FATAL: all hosts have already failed -- aborting

PLAY RECAP ********************************************************************
           to retry, use: --limit @/home/amnesia/securedrop-prod.retry

app                        : ok=2    changed=1    unreachable=0    failed=1
mon                        : ok=2    changed=1    unreachable=0    failed=1
```

This was verified on the server with `curl`.

garrettr:
Looking into this now, thanks for the report.

ehartsuyker:
[...]

Issue #1114, opened 2015-09-17
Support secondary DNS server(s) in Ansible playbooks


conorsch:
The Ansible configuration currently supports only a single string for a DNS server. Instead, the `dns_server` variable should be a list, and all IP addresses in it should be allowed out when writing the iptables rules.

The Network Firewall docs already recommend using two DNS servers (giving `8.8.8.8` and `8.8.4.4` as examples), but given the firewall rules, `8.8.4.4` will not be reachable by the servers. We should also update the Server Config docs with examples that include typing both DNS servers. Currently the docs only recommend setting `8.8.8.8`.

Changing the format of any variables in `prod-specific.yml` is likely to cause a merge conflict for admins when they pull down a new version from the git repo. Look into intelligent handling of the value to check whether it's iterable or a string, to retain backwards compatibility for admins with only a single value set.

eloquence:
This was fully realized via #5288. :)

PR #1254, opened 2016-03-06
Updates server install documentation

heartsucker:
Related:
https://github.com/freedomofpress/securedrop/issues/1249
https://github.com/freedomofpress/securedrop/issues/1244


conorsch:
These are mostly positive changes, particularly the note pertaining to #1244. The recommendation to manually set the DNS servers seems redundant, however, given that the Ansible playbooks already handle that setup in `roles/common/tasks/harden_dns.yml`. Admittedly the playbook only handles setting one DNS server, rather than two, but we have an issue open to track that as well:  #1114.

The encouragement to read the whole install guide, regardless of familiarity with installing Ubuntu, is frank and potentially quite useful to new admins, so thank you for that contribution.

ehartsuyker:
@conorsch When running `securedrop-prod.yml`, `install_fpf_repo` is the first role/task run. It uses a DNS record (`apt.freedom.press`), so if the admin didn't correctly set the DNS records at install time, this will fail in a non-obvious way.

I can update the docs here to remove that bit about DNS and open a PR to shuffle the playbooks so that the DNS is set first. If y'all are so inclined.

And actually https://github.com/freedomofpress/securedrop/issues/1247 is related in some sense in that things using TLS depend on `ntp`. Maybe these together are worth an issues or a partial refactor of `ansible` so that fresh install don't hit these little hiccups like I did.

garrettr:
> open a PR to shuffle the playbooks so that the DNS is set first

This seems reasonable to me, and removes what is currently an implicit dependency in the Ubuntu installation process (that you follow our documentation and set up DNS servers when installing Ubuntu). @conorsch What do you think?

The other option is to configure the network firewall to act as a DNS server, which would mean Ubuntu's default behavior of using the gateway for DNS would lead to correct DNS resolution. James and I debated doing that a while back, unfortunately I cannot recall exactly why we decided this was the way to go.

conorsch:
@garrettr Optimizing the Ansible playbooks to be more resilient is always a good thing, so I'm in favor of making changes here. Rather than patch the playbooks in master, however, we should work on getting develop merged in for the new release and make changes on top of that.

docs/development/getting_started.rst
```diff
@@ -327,7 +327,7 @@ Using Tor Browser with the development environment
 We strongly encourage sources to use the Tor Browser when they access
 the Source Interface. Tor Browser is the easiest way for the average
 person to use Tor without making potentially catastrophic mistakes,
-makes disable Javascript easy via the handy NoScript icon in the
+makes disabling Javascript easy via the handy NoScript icon in the
```
conorsch: :+1:

docs/hardware.rst
```diff
@@ -224,6 +224,11 @@ insert the cards into the NUC before it can be used. We recommend:
 	  updating the BIOS according to `these instructions
 	  <http://arstechnica.com/gadgets/2014/02/new-intel-nuc-bios-update-fixes-steamos-other-linux-booting-problems/>`__.
 
+.. note:: Some BIOS versions on the NUC will cause the server to
+	  `brick itself <https://communities.intel.com/message/359708>`__ if
+	  the device attempts to suspend. Some suggestions include disabling
+	  suspend in the BIOS as well as OS options like "wake on LAN".
+
```
conorsch: Great advice, see #1244 for context.

conorsch:
There are multiple changes in this PR:

1. documentation potential "bricking" of NUCs (see #1244)
2. fix minor typo in Tor Browser Bundle description
3. refactor docs flow for OS installation

1 and 2 are easy to review and should be merged straightaway. 3 requires more careful review, but in general the tone is frank and accessible, and definitely improves the project documentation. Note that the **Minor Admin Tasks** section explains steps for rectifying system state that can be automated away via Ansible, but isn't yet—see #1249.

+1 for merge.

conorsch:
Also includes FDE warning discussed in #1266, and therefore closes #1265.

garrettr:
Gonna trust @conorsch's +1 review for this one and merge.

PR #1256, opened 2016-03-08
Stick to Tails defaults for GPG keyserver options

fowlslegs:
In our last docs release, we switched to a hidden service SKS mirror. The
reasons for doing so were:

* Traffic never leaves the Tor network
* End-to-end TLS w/ self-authentication via the .onion URL

The hidden service keyserver has since gone down and this has caused a problem
for users trying to install SecureDrop, especially because GPG doesn't actually
inform the user when they specify an invalid keyserver address. Instead, they
are told that the operation they wanted to perform (e.g., `--search-keys`,
`--recv-key`) failed.

--

I propose we remove the `--keyserver` option altogether for the reasons that
follow.

In the Tails `gpg.conf` they default to using the SKS keyserver pool over
SSL/TLS (HKPS). The configuration file also specifies to replace the default
certificate store with just the singular self-signed certificate that is shared
among the TLS-capable SKS keyservers. This solves the CA problem in a different
way than the use of hidden services does. Traffic does leave the Tor network,
but only in encrypted form.

Further, the configuration specifies the `no-try-dns-srv` keyserver-option,
which though only documented in an upstream commit, prevents the leaking of DNS
(see https://trac.torproject.org/projects/tor/ticket/2846).

Lastly, sticking with the Tails `gpg.conf` defaults will be reliable since I
don't expect the entire SKS keyserver infrastructure to go down anytime soon.
Any problems our users have will be affecting all Tails and/ or SKS users, so we
can expect a timely fix in the unlikely chance of future problems.

--

One final thing to consider is the case that it has been discovered that the SKS
keyserver certificate is compromised. In this case, the certificate file will
not be replaced until the next version of Tails. It will be big news if this
does happen, so I'm sure we'll find out and quickly push a warning and
workaround to the documentation. Further, the risks implied by this problem are
still present with the use of hidden services, HKPS w/ the system certificate
store (as opposed to a self-signed cert file), and with no HKPS at all.

Signed-off-by: Noah Vesely <noah@freedom.press>

### DIFF
--- a/docs/ossec_alerts.rst
+++ b/docs/ossec_alerts.rst
@@ -100,7 +100,7 @@ fingerprint:
 
 Download your key and import it into the local keyring: ::
 
-    gpg --keyserver hkp://qdigse2yzvuglcix.onion --recv-key "<fingerprint>"
+    gpg --recv-key "<fingerprint>"
 
 .. note:: It is important you type this out correctly. If you are not
           copy-pasting this command, we recommend you double-check you have
--- a/docs/servers.rst
+++ b/docs/servers.rst
@@ -34,7 +34,7 @@ integrity with cryptographic signatures and hashes.
 First, we will download *Ubuntu Image Signing Key* and verify its
 *fingerprint*. ::
 
-    gpg --keyserver hkp://qdigse2yzvuglcix.onion --recv-key "C598 6B4F 1257 FFA8 6632 CBA7 4618 1433 FBB7 5451"
+    gpg --recv-key "C598 6B4F 1257 FFA8 6632 CBA7 4618 1433 FBB7 5451"
 
 .. note:: It is important you type this out correctly. If you are not
           copy-pasting this command, we recommend you double-check you have
--- a/docs/set_up_admin_tails.rst
+++ b/docs/set_up_admin_tails.rst
@@ -60,7 +60,7 @@ Signing Key*.
 
 .. code:: sh
 
-    gpg --keyserver hkp://qdigse2yzvuglcix.onion --recv-key "B89A 29DB 2128 160B 8E4B 1B4C BADD E0C7 FC9F 6818"
+    gpg --recv-key "B89A 29DB 2128 160B 8E4B 1B4C BADD E0C7 FC9F 6818"
 
 .. note:: It is important you type this out correctly. If you are not
           copy-pasting this command, we recommend you double-check you have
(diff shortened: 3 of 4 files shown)


fowlslegs:
Sorry for the lengthy commit message for a simple change. I just feel like with the recent Linux Mint disaster we need to be extra vigilant about our verification process.

conorsch:
@fowlslegs Great PR. The patient explanation is a good thing for those following along at home. Given the problems we've seen recently around this for new admins, I'm in favor of merging sooner rather than later, since the change would land in the docs immediately.

It's worth pointing out that we're also recommending a specific keyserver in `upgrade/0.3.5-to-0.3.6.rst`, although it's not a hidden service (pool.sks-keyservers.net) so I think it's OK to let go for now.

fowlslegs:
Okay, I changed that one as well since as the Tails `gpg.conf` (relevant section reproduced below) stands  passing `--keyserver pool.sks-keyservers.net` would represent a security downgrade.

    #-----------------------------
    # keyserver
    #-----------------------------

    # This is the server that --recv-keys, --send-keys, and --search-keys will
    # communicate with to receive keys from, send keys to, and search for keys on
    keyserver hkps://hkps.pool.sks-keyservers.net

    # Provide a certificate store to override the system default
    # Get this from https://sks-keyservers.net/sks-keyservers.netCA.pem
    keyserver-options ca-cert-file=/usr/local/etc/ssl/certs/hkps.pool.sks-keyservers.net.pem

    # Set the proxy to use for HTTP and HKP keyservers - default to the standard
    # local Tor socks proxy
    # It is encouraged to use Tor for improved anonymity. Preferrably use either a
    # dedicated SOCKSPort for GnuPG and/or enable IsolateDestPort and
    # IsolateDestAddr
    keyserver-options http-proxy=socks5-hostname://127.0.0.1:9050

    # Don't leak DNS, see https://trac.torproject.org/projects/tor/ticket/2846
    keyserver-options no-try-dns-srv

    # When using --refresh-keys, if the key in question has a preferred keyserver
    # URL, then disable use of that preferred keyserver to refresh the key from
    keyserver-options no-honor-keyserver-url

    # When searching for a key with --search-keys, include keys that are marked on
    # the keyserver as revoked
    keyserver-options include-revoked

Rebased and squashed this last change into the original commit.

fowlslegs:
Finally, I'm making one more change I'll push right after posting this and we should be good to go. I just noticed that the upgrade instructions are out of sync with the new verification instructions added in commit https://github.com/freedomofpress/securedrop/commit/d7d25c602166dd1cc7b1937b5bce28889701d45f and discussed at length in the issues referenced in the commit message.

garrettr:
This looks good; I agree that it's best to stick with Tails' defaults. That way, if it breaks, it's their fault 😁

garrettr:
@fowlslegs Can you rebase this on the latest develop?

Issue #749, opened 2014-11-16
Verify contents of prod-specific.yml before installing SecureDrop


runasand:
When installing SecureDrop, the admin is asked to fill out _prod-specific.yml_ before running an _ansible-playbook_ command which installs the software and configures/hardens the servers. As part of the installation process, the _ansible-playbook_ command should verify the contents of _prod-specific.yml_ before continuing. 

That is, we need to make sure the information we require for the install is present in the file, and we also need to make sure the information that is present is accurate. The installation process does not seem to complain about the fact that the IP address I entered for the Monitor Server is "192.168.1.522".

fowlslegs:
@conorsch Think this one should stick in the 0.4 milestone?

conorsch:
@fowlslegs Yes, thanks for nudge. Will step through the related issues here and evaluate the same.

conorsch:
#1348 has good targets for validation, but as described in https://github.com/freedomofpress/securedrop/pull/1348#issuecomment-240519667 and https://github.com/freedomofpress/securedrop/issues/1340#issuecomment-237020725, we'd do well to use Ansible's `assert` module to perform these checks, rather than a custom script.

conorsch:
Preparing a PR to resolve, waiting for review and merge of #1762. All the logic is present in the `validate-all-prod-vars` feature branch right now, but I'd like to cherry-pick some of @heartsucker's work in #1259 to strengthen the GPG validation.

conorsch:
Given feedback from first-time Admins, a common problem during installation is not enabling required persistence options, e.g. SSH keys. We should try to validate that the persistence options are indeed enabled during setup time, otherwise hard-to-debug problems come up, e.g. loss of SSH keys.

PR #1259, opened 2016-03-10
gpg key verification

### LINKED ISSUE #1257: GPG keys should be validated against fingerprints before provisioning
The application will fail if the GPG key fingerprint is wrong.

Related: https://github.com/freedomofpress/securedrop/issues/1252

heartsucker:
Fixes #1257 by importing the keys into `gpg2` on the admin work station and then checking the fingerprints it outputs.

This is slightly better than having the user provide just the key and then using `gpg2` to extract the fingerprint and using `set_fact` as it means the user has to screw up twice to cause the application to do Bad Things.

### DIFF
--- a/install_files/ansible-base/files/validate-gpg-key.sh
+++ b/install_files/ansible-base/files/validate-gpg-key.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+set -e
+set -o pipefail
+
+# Verifies a key at at given path matches a fingerprint
+# Does so by importing the key into gpg2 and checking output
+
+# Exit codes
+# 0 - fingerprint matches key
+# 1 - otherwise
+
+declare -r key_location="$1"
+declare -r fingerprint="$2"
+
+: {key_location:?'A path to a key must be provided (arg 1)'}
+: {fingerprint:?'A fingerprint must be provided (arg 2)'}
+
+# check if fingerprint is formatted correctly
+echo "$fingerprint" | egrep -q '^[0-9A-F]{40}$'
+
+# validate key against fingerprint
+gpg2 --import "$key_location"
+gpg2 --list-keys --fingerprint "$fingerprint" | \
+  grep 'Key fingerprint =' | \
+  sed -e 's/ //g' | \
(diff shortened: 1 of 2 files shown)


conorsch:
Thanks, I like the approach here. The bash script would be better structured as a tedious series of idempotent ansible tasks, but I can handle that part. There's a slew of improvements to the "validate" role that will abort the playbook run if the config is deemed invalid. Sanity checks for the GPG fingerprint would be wise!

conorsch:
Related to validation improvements e.g. #749, which we're tracking for 0.4. In those validation issues we've decided that using `assert` wherever possible is preferable, but in some cases we may need to resort to scripts, and this is a possible consideration for that. Leaving open for further review.

Have a few issues with the current implementation not testing quite what it it's designed to test—will comment in-line about that.

install_files/ansible-base/files/validate-gpg-key.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# check if fingerprint is formatted correctly
+echo "$fingerprint" | egrep -q '^[0-9A-F]{40}$'
+
+# validate key against fingerprint
+gpg2 --import "$key_location"
+gpg2 --list-keys --fingerprint "$fingerprint" | \
+  grep 'Key fingerprint =' | \
+  sed -e 's/ //g' | \
+  grep -q "$fingerprint"
```
conorsch: Importing the key and then checking the expected fingerprint doesn't test that the fingerprint matches the file. Consider the following:

1. Generate valid keypair and list pubkey filename and correct fingerprint in `prod-specific.yml`.
2. Run playbooks to set config on servers.
3. Generate _new_ keypair and update the pubkey filename, but _not_ the fingerprint in `prod-specific.yml`.
4. Run playbooks to set config on servers.

In step 4, the playbook will complete successfully, because the old fingerprint was never removed from the keyring. So the test passes when it shouldn't. A reasonable solution here would be to set up a temporary GNUPGHOME dir, and import the key there, so it contains _only_ the key currently being imported.

PR #1260, opened 2016-03-11
use ansible to enforce required ubuntu settings

### LINKED ISSUE #1249: Docs as described may lead to DNS lookup failures

heartsucker:
most ansible tasks have an implicit requirement on
  - the correct settings of nameservers
  - the correct system time

fixes #1249

### DIFF
--- a/install_files/ansible-base/roles/prerequesites/files/resolvconf-reconfigure.sh
+++ b/install_files/ansible-base/roles/prerequesites/files/resolvconf-reconfigure.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+set -e
+
+/usr/bin/expect<<EOF
+spawn dpkg-reconfigure -f readline resolvconf
+expect "updates?" { send "Yes\r" }
+expect "dynamic files?" { send "Yes\r" }
+EOF
+
+exit 0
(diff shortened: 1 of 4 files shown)


conorsch:
These changes seem overwrought. In order to ensure correct behavior of the apt repo config, the system should have the correct time set so that the TLS handshake for the HTTPS apt repo works. Provided the Admin has followed the setup guide—and thanks for your edits to the the Server Install Guide in #1254—then the server will query for system time during initial setup, which the firewall rules allow.

It does make sense to ensure that the ntp package is installed prior to the FPF apt repository (or any other HTTPS apt repo, for that matter). The playbook currently doesn't enforce that, which allows for an edge case of misconfigured system time. Will track that enhancement in #1249.

I'll also point out that the `common` role does already set the `nameserver` line to the DNS server IP, and it does so via a template, which generally provides fewer side-effects than `lineinefile`. (Normally I'd say template also grants cheap idempotence, but you handled that in your changes via judicious use of registered vars.)

PR #1262, opened 2016-03-13
Removed TODOs from the docs that might scare sources away

heartsucker:
This is one of two solutions that fix the current issue in the docs: the TODOs look like an unprofessional doubt-casting discussion that reduces SecureDrop's credibility.

This solutions hides the aforementioned (open a new GH issue, don't leave it in the docs) and makes it more likely a source will leak the docs.

The other solution would be to say "If you're absolutely worried about your identity being revealed, and you haven't removed documents from their systems, you should stop now because by not using Tor while researching document leaking services, your identity may have already been compromised."

Or something better than the phrasing on either of those because I'm no copy writer.

### DIFF
--- a/docs/source.rst
+++ b/docs/source.rst
@@ -1,12 +1,6 @@
 Source Guide
 ============
 
-.. todo:: There's a lot more to it than this, but then we begin to
-          duplicate content from individual organization's landing
-          pages and Micah's Intercept article. For example: what
-          computer should you use? what network should you be on? etc.
-
-	  
 Choose who to submit to
 -----------------------
 
@@ -21,16 +15,6 @@ Most organizations make their landing page prominently accessible from their mai
 
 Get the Tor Browser
 -------------------
-
-.. todo:: There is a classic chicken and egg problem with encouraging
-   sources to use Tor. If their network traffic is already being
-   monitored (e.g. at their workplace by the IT department, at their
-   home by their ISP, anywhere by FVEY...), and they don't know to use
-   Tor from the start (like Snowden did), then their initial interest
-   in using SecureDrop might be enough to make them a target. One
-   partial mitigation might be to make this section the first section
-   in this document, and encourage them to "stop, drop, and Tor
-   Browser" before continuing or clicking any more links.
 
 Each SecureDrop instance has a publicly available *Source Interface*: a website where sources can create anonymous accounts, submit files and messages, and check back for replies.
 


fowlslegs:
I believe these were put in place before our last hackathon, to aid participants wanting to work on our documenation. There are also quite a number more:

```
servers.rst:82:.. todo:: Not sure if we should encourage people to put a passphrase
create_admin_account.rst:42:.. todo:: Clarify how to set up TOTP/HOTP through ``./manage.py
onboarding.rst:44:.. todo:: Describe best practices for provisioning multiple Secure
development/documentation.rst:80:  .. todo:: I don't love this convention for a couple of reasons:
development/apparmor_profiles.rst:39:.. todo:: Explain the different behavior of the virtual environments (staging
development/getting_started.rst:114:.. todo:: The branching model should be documented separately, in a
development/getting_started.rst:198:.. todo:: Explain why we allow direct access on the staging environment
development/getting_started.rst:239:.. todo:: Replace all of this with nc, which is simpler, works well with
source.rst:4:.. todo:: There's a lot more to it than this, but then we begin to
source.rst:25:.. todo:: There is a classic chicken and egg problem with encouraging
configure_admin_workstation_post_install.rst:77:.. todo:: Do we still want to recommend/require this? Should it be
overview.rst:55:.. todo:: A picture of an actual physical setup (e.g. the office
generate_securedrop_application_key.rst:4:.. todo:: Tom complained about the name "SecureDrop Application GPG
generate_securedrop_application_key.rst:69:.. todo:: I would prefer a filename of ``SecureDrop.pub.asc``, as that
generate_securedrop_application_key.rst:79:.. todo:: The screenshot shows them saving the public key with a
generate_securedrop_application_key.rst:96:.. todo:: This is the wrong place for this. The first phase of the
journalist.rst:12:.. todo:: This whole document needs to be reworked, it's all over the
journalist.rst:24:.. todo:: This document recommends transferring documents from the SVS
tails_printing_guide.rst:19:.. todo:: Consider adding images of enabling persistence and setting an admin
terminology.rst:9:.. todo:: Pictures would be good for many of the objects defined here
conf.py:23:# If extensions (or modules to document with autodoc) are in another directory,
conf.py:36:extensions = ['sphinx.ext.todo', ]
conf.py:107:# If true, `todo` and `todoList` produce output, else they produce nothing.
conf.py:108:todo_include_todos = True
```

A short term solution could be to simply comment them out. The long term solution is to read through them all and address the concerns in our documentation.

conorsch:
I agree with the changes proposed here—the TODOs have been in place too long as it is, and there are more than just those addressed in this PR. +1 for merge

fowlslegs:
Was going to say we could easily comment out all the todos with `grep` and `sed`, but just realized there's actually a config option to leave them out of the Sphinx output.

Closing this PR in favor of that solution: https://github.com/freedomofpress/securedrop/pull/1264

PR #1264, opened 2016-03-25
No longer include todos in Sphinx output

fowlslegs:
This addresses https://github.com/freedomofpress/securedrop/pull/1262.
Basically, our 'todos' are notes we put in place for potential improvements of
the documenation, with the Aaron Swartz hackathon in mind. For next year, we can
simply switch back this setting if we'd like to encourage people to help us with
our docs again.

Signed-off-by: Noah Vesely <noah@freedom.press>

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -105,7 +105,7 @@ pygments_style = 'sphinx'
 #keep_warnings = False
 
 # If true, `todo` and `todoList` produce output, else they produce nothing.
-todo_include_todos = True
+todo_include_todos = False
 
 
 # -- Options for HTML output ----------------------------------------------


fowlslegs:
@conorsch Suggested the problem is with the coverage logic, and that everything should build fine with this change, but that we should blacklist `conf.py` to prevent these warnings in the future, since that is more reasonable than creating tests for documentation.

conorsch:
+1 for merge. Built the docs locally and toggled this change—the documentation is much more readable with the proposed update. Note that disabling TODOs does not affect other boxes such as Warning, Caution, and Note.

Issue #1152, opened 2015-10-19
Improve SecureDrop loading experience over Tor Hidden Services


garrettr:
In production, all of SecureDrop's web interfaces (the Source Interface and the Document Interface) are only available as Tor Hidden Services, which have extremely restricted bandwidth. In addition, SecureDrop intentionally avoids caching _any_ resources in the client's browser to avoid potential forensic infoleaks.

These qualities combine to make the SecureDrop web interface feel slow and glitchy, particularly when used over an especially slow hidden service connection:

![securedrop source interface performance](https://cloud.githubusercontent.com/assets/407302/10588496/50a10bb8-765b-11e5-8a1c-12f1b1297e1f.gif)
## Specific Issues to Fix
- [ ] Seeing the unstyled page before the CSS loads. This looks glitchy and unprofessional.
- [ ] Seeing the logo resize during the page load
- [ ] Reduce or eliminate the delay in showing the icons (e.g. cloud icon next to "Submit Documents" on the home page)
- [ ] Reduce or eliminate delay on page loads after the first in a session. This requires research, as it conflicts with our current position on avoiding caching web resources, and we would need to make sure that we can use some form of caching without the browser writing the resources to long term storage.
## Potential Resolutions
- Reordering or inlining resources (stylesheets, images)
- Reduce the size of currently used resources to load them faster with the limited bandwidth we have, e.g.
  - Reduce the size of logos
  - Reduce the icons files to just the icons that are actually used in SecureDrop
  - Minify JS and CSS
- Caching

garrettr:
I started working on this, but ran into an issue with the development cycle and our existing set of automated development environments. Rather than work around it, I think this is an instructive use case that could help us improve our development environments and automation.

Essentially, the problem is:

1. The `development` environment synchronizes changes to the application code (`securedrop/securedrop`) to the VM, so it automatically synchronizes the type of changes I need to make (changes to HTML templates, CSS, images, JS, etc.). _However_, it does not make the site available via Tor Hidden Service.
2. The `staging` environment _does_ make the site available via Tor Hidden Service, but it installs `securedrop-app-code` from the production apt server. In order to use `staging` to test my changes, I would need to:
    1. Stand up a testing apt server, edit `staging-specific.yml` and associated files to use that instead of the production apt server.
    2. Make changes.
    3. Rebuild the `securedrop-app-code` deb locally, using the `build` vm
    4. Upgrade the testing apt server with the new, testing deb.
[...]

PR #1267, opened 2016-03-27
Added mod_pagespeed to apache2

heartsucker:
Helps with #1152.


ageis:
This PR looks great, but one thing to consider with it is the overall desirability of adding more code that increases the attack surface. For instance, there was a previous version of SecureDrop that used a Google authenticator Apache module, which we later found out had vulnerabilities in it.

Can you produce any comparative metrics that show the impact of mod_pagespeed on the loading and browsing experience?

heartsucker:
I tried to boot a prod VM, but this happened: #1268

In the staging environment, it merges 4 CSS resources into 1, and 2 JS into 1. It also cuts the size of the HTML by about 30%. I'll get some better numbers later today (hopefully).

heartsucker:
@ageis
Without `mod_pagespeed`:
screen shot 2016-03-28 at 15 10 21

With `mod_pagespeed`:
screen shot 2016-03-28 at 15 12 22

In the example they give the exact same response time (new tor circuits / identities for both requests). But, the request time varies from ~1.5 seconds to upwards of 15 on repeated tests. What makes pagespeed worth it IMO is on the long / slow requests, the CSS arrives faster and makes the page usable faster.

There's actually another filter called `inline_css` that adds all the CSS assets directly to the HTML, but that was acting up on vanilla Firefox and TorBrowser, so I left it out to prevent unstyled pages. When that works well, it makes the page usable the second the HTML arrives, which was around 0.5-5 seconds.

install_files/ansible-base/roles/app/tasks/mod_pagespeed.yml
```diff
@@ -0,0 +1,61 @@
+---
+# TODO remove when mod_pagespeed is an fpf deb package
```
garrettr: Why we would make this an FPF deb package? Our general goal is to try to maintain as few deb packages as possible 😄 .

My guess: that's one way to avoid installing the package over HTTPS, which is more vulnerable to MITM than installing through `apt` and authenticating via GPG.

garrettr:
@heartsucker Thanks for the detailed comparisons and screenshots in https://github.com/freedomofpress/securedrop/pull/1267#issuecomment-202400575!

install_files/ansible-base/roles/app/tasks/mod_pagespeed.yml
heartsucker: Yeah, the idea was security since you guys already maintain other packages. Not trying to force more work on y'all. :D

fowlslegs:
I think this has the potential to detrimentally effect a website fingerprinting defense such as [ALPaCA](https://3tmaadslguc72xc2.onion). Specifically, the combining of resources by Apache being out of control of the webapp is problematic. Further, I don't think the difference would really be noticeable if you were to actually test this over Tor. And as @ageis points out, this also increases SD attack service, adding a significant amount of C code. I'm in favor of closing this PR for all the reasons above. Since #1268 is fixed, maybe you want to try testing this on prod again @heartsucker, but I don't think we should merge this even if we do see noticeable improvements.

fowlslegs:
Going to close for the reasons listed above. Not totally opposed if you can find a way to side-step the above-mentioned concerns.

Issue #1263, opened 2016-03-22
Backup role - support for large zip files


dbgdngit:
SecureDrop v0.3.6

I'm starting to use the backup role of securedrop-prod.yml for regular backup of key files.
It currently fails as the zip file is nearly 3GB in size
I can get it working by adding the allow64ZIP=True flag in the script 0.3_collect.py script
My understanding is that is the default in python 3.4 and above?

conorsch:
We're testing a rewrite of the backup strategy in #1273. Thanks for reporting this!

garrettr:
This will be fixed in #1273 because it switches to using tar instead of zip for the backup archive. The [default format](https://docs.python.org/2/library/tarfile.html#tarfile.DEFAULT_FORMAT) used by `tarfile` in Python 2.7 is `GNU_FORMAT`, for which the maximum archive size is [effectively unlimited](http://answers.google.com/answers/threadview?id=25116).

Note, however, that the backup archive is transferred over Tor, and that will be a significant bottleneck if your backup archive is more than a couple of GB in size. Another bottleneck to keep in mind is the amount of free space in the persistent volume on the Admin Workstation, which is where the backup archive is stored.

fowlslegs:
>the backup archive is transferred over Tor, and that will be a significant bottleneck if your backup archive is more than a couple of GB in size

Another place pylzma would be useful. Just noting this to x-ref https://github.com/freedomofpress/securedrop/issues/1282

conorsch:
Resolved by #1273, pending release of 0.3.7. Tracking backporting these changes into develop in #1313.

Issue #1127, opened 2015-09-21
Backups


flamsmark:
Issue #606 talks about disaster recovery, but doesn't offer a full recovery option. In my SecureDrop setup, I elected to attach an external hard drive to the application server for nightly incremental backups. Other admins may prefer to rotate backup hard disks, or execute periodic backups to a remote networked machine.

None of this is covered by current documentation. Encouraging good and safe backup practices should be part of the setup instructions.

conorsch:
The backup/restore role isn't documented because it needs work. For 0.4, I'm leaning toward splitting the backup role up into two types, one for critical server info like ATHS values, the other a backup of all the submissions on the App Server. See #942 for discussion of around chaging the backup role.

One the backup role is rewritten, documenting the recovery process is mandatory.

conorsch:
#1273 provides an overhaul of the backup strategy, and creates a Topic Guide in the docs dedicated to explaining the steps necessary for backup and restore. Scheduled for the 0.3.7 release.

garrettr:
#1273 significantly improved the backup and restore processes, and included detailed documentation. The process is still overly complex and requires care and attention from administrators who are performing backups, but for the purposes of the current Docs Sprint I think the current state is sufficient.

## Suggested next steps

* Test backup/restore for 0.4, using the current backup roles and documentation.
* Consider creating one or more helper scripts to assist in dealing with some of the current complexity in backups/restores.
  * For example, one pain point is systems that may have a large number of stored submissions. Currently, administrators are advised to check the total size of the stored submissions, and manually verify that they have sufficient disk space on both the App Server (where the backup archive is built, before being transferred to the Admin Workstation) and the Admin Workstation (where the backup archive is stored). This could be scripted fairly easily, assuming the script is run from the Admin Workstation and has SSH access to the servers.
  * Another pain point is the extremely slow transfer time (and potential for interruptions) when transferring a potentially large (multi-GB) backup archive over a Tor Hidden Service connection. This is harder to deal with, since SecureDrop's hardening limits all connectivity to the servers to be over Tor Hidden Services.

If folks are interested in prioritizing one or more of these next steps, we should file separate issues for each of them.

redshiftzero:
Reviewing the changes in #1273 I agree that addresses this issue. I've added a note for the 0.4 hardware testing to run through the backup/restore process as time permits. Do you want to add your point re: creating a helper script to a new issue? Then I think this is good to close :metal:

[...]

Issue #942, opened 2015-03-24
Time out for secure-delete of backups


dolanjs:
The ansible tasks to secure-delete the encrypted backup off the app server is set with a timeout of 300 seconds. For existing installs who have either a lot of sources or large submission this may not be enough. We should consider extending the timeout to a larger value. As Garrett mentioned we could also move this to the last task after the final reboot and make it just a fire and forget task.

garrettr:
Another idea that might make more sense would be to have two different types of backups:

1. Backup of essential configuration.
2. Backup of server state (`/var/lib/securedrop` directory).

Type 1 is simple and can be done as part of the ansible playbook, as it is now. Type 2 does not work well when run as part of the Ansible playbook once the amount of state stored on the server grows past a certain size - copying the backup is extremely time-consuming (over Tor), liable to be interrupted due to circuit flakiness (due to Tor), and the secure-remove step can take too long given the Ansible timeout (which is the top comment for this issue).

This basically reiterates the conversation we've already had in #606. For 0.4, I propose we:

1. Only backup essential configuration as part of the Ansible playbook (this has a small, fixed size, and is not problematic in the context of timeouts.
2. Figure out some solution for periodic backups of server state (type 2 backups).

conorsch:
I'm in favor of splitting up the backup role into two types. Making the secure delete process fire-and-forget is also a good compromise—simply incrementing the timeout counter is essentially kicking the can down the road. Any objections?

conorsch:
#1273 removed the use of secure-delete (via `srm`), so the concerns listed here are no longer relevant. Closing.

PR #1273, opened 2016-04-13
Better backup and restore

garrettr:
This improves the backup and restore Ansible roles. The primary goals are to make backups easier, faster, and less error-prone.

## Key Changes

* _Greatly_ simplifies the backup and restore scripts.
  * **Important Note**: this happens at the expense of flexibility. The new scripts **assume** that both the source and destination installations use the default paths for SecureDrop and Tor. Based on my experience, this is a reasonable assumption, and it is documented in comments on the scripts.
* Stops encrypting the backup files to the air-gapped key.
  * This does not provide a security benefit. All of the sensitive files (submissions) in the backup are already encrypted to the air-gapped key. Also, the Admin Workstation has full access (via SSH) to the App Server, so it can be trusted with data from the App Server and there is no need for an additional layer of encryption. 
  * Not encrypting the backup file makes restoring from backup *much* less complicated and error-prone.
* Stops "secure deleting" the backup files.
  * No security benefit.
  * Creates problems for submissions that have a non-trivial amount of submissions (e.g. > 100MB). Secure deleting this amount of data takes a *very* long time, which creates all kinds of problems.
  * This is a problem for systems that use hardware RAID, since `shred` and hardware RAID do not play nice.
* Renames the variable that needs to be set to trigger a restore, from `backup_zip` to `restore_file`, which is much clearer. In the past, some admins were confused and (reasonably) thought that `backup_zip` was the name of the file to backup *to*, rather than the file to restore *from*.

## Testing

- [x] Since this PR modifies `prod-specific.yml`, test: one should be able to `git pull` and then cleanly re-apply an existing production configuration, using the stash-pull-pop workflow.
- [x] Test that this handles large amounts of submissions well (both backup and restore)
  - [x] Backup large SeucreDrop over Tor
  - [x] Restore large SecureDrop over Tor
- [x] Ensure that restoring SSH ATHS does not cause problems for the admin.
 - [x] Does it prevent the restore play from completing?
 - [x] What is the best way to sync the Admin Workstation with the restored server state?


install_files/ansible-base/roles/backup/files/0.3_restore.py
```diff
@@ -1,155 +1,50 @@
 #!/usr/bin/python2.7
 """
+This script and backup archive should be copied to the App server and run by
+the Anisble playbook. When run (as root), it restores the contents of the 0.3
```
ageis: spell check on Anisble

garrettr:
Currently uploading several GB of data to a test SecureDrop instance so I can do a realistic test of backup and restore (i.e. to compare to a production server that may have been running for years and accumulated a significant number of submissions).

garrettr:
Verified that the stash-pull-pop workflow works with the changes to `prod-specific.yml` in this PR. On a test production machine, Git was able to automerge the file and no manual merge conflict resolution was required.

garrettr:
Verified migration by backing up the SecureDrop Test Instance with 2GB of submissions and then restoring it onto some staging VM's. It would be good to test the restore over Tor to make sure it's unlikely to time out.

install_files/ansible-base/roles/backup/tasks/main.yml
```diff
@@ -1,50 +1,6 @@
...
-  when: backup_zip is defined
-
-- name: if a restore file is defined secure delete it
-  command: srm /tmp/{{ backup_zip }}
-  async: 300
-  poll: 10
-  when: backup_zip is defined
+  # Perform a restoration if a restore file is specified
+- include: restore.yml
+  when: restore_file is defined
```
conorsch: Nit: `when: restore_file is defined and restore_file != ''`. An empty string should be treated the same as undefined.

install_files/ansible-base/roles/backup/tasks/restore.yml
```diff
@@ -0,0 +1,15 @@
+---
+- name: copy restore script to app server
+  copy: src=0.3_restore.py dest=/tmp/ owner=root mode=770
```
conorsch: For consistency, use the multiline YAML syntax here, as in the `file:` task below:

```
- name: Copy restore file to App Server.
  copy:
    src: "{{ restore_file }}"
    dest: /tmp/{{ restore_file }}
```

install_files/ansible-base/staging-specific.yml
```diff
@@ -46,11 +46,3 @@ sasl_username: "test"
...
 sasl_password: "password123"
 
-### Use for backup restores ###
-# If the `backup_zip` variable is defined ansible will copy the defined file to
-# the app server and run the 0.3_collect.py script to unzip and restore those
-# files.
-# ex: backup_zip: "sd-backup-2015-01-15--21-03-32.zip"
-# To use uncomment the following line and enter the correct decrypted zip
-# filename between the quotes.
-#backup_zip: ""
```
conorsch: Skipping the backup action by default makes sense for staging. :+1:

conorsch:
> On a test production machine, Git was able to automerge the file and no manual merge conflict resolution was required.

That's a pleasant surprise!

Backporting these changes onto `develop` will require manual conflict resolution, however, since the `develop` branch already has multiline YAML syntax implemented on all tasks for this role, but without the new functionality.

install_files/ansible-base/staging-specific.yml
garrettr: @conorsch Yeah. Obviously all of this config management needs an overhaul anyway but this seemed like a reasonable short-term change.

install_files/ansible-base/roles/backup/tasks/main.yml
garrettr: > An empty string should be treated the same as undefined.

Really? Obviously an empty string is an invalid value here, but if the variable is defined at all then somebody has modified the configuration to define it (even if they did so incorrectly). I am inclined to say "if you defined a variable, you're responsible for defining it correctly" (at least for now).

garrettr:
Testing a large restore over Tor right now. To do so, I:

* Copied the `sd-backup-*.tar.gz` backup file from the test server to my development machine.
* `vagrant up /staging$/` up a staging environment
 * I could've used `/prod$/`, but had already set up staging and didn't want to wait for a rebuild.
* Modified Vagrant for the staging environment so vagrant+ansible would do SSH over Tor, by cherry-picking the relevant changes for `SECUREDROP_SSH_OVER_TOR` from the develop branch.
 * Would not have had to do this if I had used `prod`, since `prod` fully emulates a production installation and always does SSH over Tor.
* Added `ansible.tags = ['backup']` to Vagrantfile so provisioning only runs the backup tag. This is to save time, since every Ansible command is happening over Tor now.
* `SECUREDROP_SSH_OVER_TOR=1 vagrant provision app-staging`
* ...wait...

garrettr:
Confirmed that a large restore over Tor eventually succeeded, although it took a very long time (2GB restore took ~5 hours).

garrettr:
Ansible also appeared to continue running successfully even after the restore script replaced the HidServAuth values and reloaded Tor, which is convenient (if a little surprising): [Ansible logs](https://gist.github.com/garrettr/1f808770d692f51b1e1a0059c336df5f).

garrettr:
> Ensure that restoring SSH ATHS does not cause problems for the admin.

Let's expand on this a little bit.

The data that gets backed up and restored from a SecureDrop includes the hidden service configuration information (from `/var/lib/tor/services`). This allows a restored SecureDrop to maintain the same .onion addresses for the Source Interface, Document Interface, and SSH interfaces.

The sequence of events for a backup and restore is as follows:

1. A SecureDrop was installed and has been running for some time.
2. An admin runs the `backup` role on the installed SecureDrop, creating a snapshot of that SecureDrop's state which is stored in an archive file.
3. A new SecureDrop is installation is performed. New hidden services are automatically generated.
4. In order to connect to the new installation to restore the backup onto it, the admin must update the Ansible inventory and configure Tor to connect to the SSH ATHS, which is done via `tails_files/install.sh`.
5. The admin runs the restore task, connecting to the new SecureDrop installation using the *new* ATHS credentials.
6. The restore role copies the **old** values over the new values, and reloads Tor.

Concerns to investigate:

- [ ] Does restoring the old SSH ATHS values break the Ansible session running the restore play?
- [ ] Once the restore role is done running, the files on the Admin Workstation (`ansible_base/*ths`, Ansible inventory, tails_files torrc_additions, etc.) will out of sync in relation to the restored state on the server. What is the best way to sync the Admin Workstation with the restored server state?

install_files/ansible-base/roles/backup/tasks/main.yml
garrettr: To clarify my stance on saying "if you defined a variable, you're responsible for defining it correctly": we can't comprehensively verify whether the values an admin sets for variables like `restore_file` are correct, so IMHO we should not bother doing a half-assed job (e.g. checking for one obviously invalid but also unlikely value, like `''`).

At some point we will overhaul the configuration mechanism for SecureDrop, and we'll probably add a layer of abstraction between the configuration settings that the admin touches versus the `.yml` that gets fed to Ansible (or whatever). I think we should leave this type of concern for that more comprehensive overhaul of SecureDrop configuration management, instead of doing a cursory job now.

garrettr:
Adding support-request since this work was initially motivated by multiple requests from SecureDrop admins who want to move their systems to different hardware. The "backup and restore" Ansible roles are the easiest way to do that, but they needed a bunch of work, which is described in detail in the top comment.

install_files/ansible-base/roles/backup/tasks/main.yml
garrettr: From in-person review, @conorsch agrees this is fine the way it is.

install_files/ansible-base/staging-specific.yml
garrettr: Note that this doesn't "skip the backup action" as @conorsch says. The backup role always runs. This simply removes an outdated comment, because this PR replaces `backup_zip` for the more descriptive variable `restore_file`.

garrettr:
>  Does restoring the old SSH ATHS values break the Ansible session running the restore play?

@fowlslegs investigated this in https://github.com/freedomofpress/securedrop/pull/1293#issue-153916849 and convincingly argued that, because reloading Tor marks circuits as dirty but *does not kill them*, restoring the SSH ATHS values will not break the currently running Ansible session *as long as it continues to use the same TCP connection*.

We noted that this may only work because we use `ControlPersist` in our SSH configuration for Ansible (https://github.com/freedomofpress/securedrop/pull/1293#issuecomment-221116353). It's important to test this thoroughly, and keep in mind that changing the Ansible SSH configuration could have a negative impact on the restore role.

garrettr:
> Once the restore role is done running, the files on the Admin Workstation (ansible_base/*ths, Ansible inventory, tails_files torrc_additions, etc.) will be out of sync in relation to the restored state on the server. What is the best way to sync the Admin Workstation with the restored server state?

In https://github.com/freedomofpress/securedrop/pull/1273/commits/cecd7e54865fca2a4cf5244ae1104f450a5dfb0c, I swapped the order of the restore role and the restrict_direct_access_app role, so when restrict_direct_access_app fetchs the Tor ATHS credentials, it does so _after_ the restore has been completed, which should keep the Admin Workstation in sync with the Tor ATHS credentials that are on the server after a restore.

garrettr:
Currently re-testing backup and restore to ensure:

1. Re-ordering the backup/restore and restrict_direct_access_app roles doesn't lead to problems with losing the Tor connection. With the roles swapped, the number of Ansible tasks that have to happen over a now-"dirty" connection is increased, and this might lead to problems.
2. State on the "Admin Workstation" is appropriately synchronized after performing a restore to a fresh installation.

garrettr:
After discussing the decision with @conorsch, I added the `backup` tag to the `{app,mon}_display_onions` tasks, which are responsible for fetching the ATHS credentials from the servers to the Admin Workstation. This allows an admin to run a backup and/or restore operation in isolation using `ansible-playbook install_files/ansible-base/securedrop-prod.yml --tags=backup`, which is faster, more convenient, and will make debugging easier in case something goes wrong with the backup and/or restore process.

garrettr:
I have fully tested backing up and restoring a small instance (~50MB restore file) in a variety of contexts using the production Vagrant environment. Everything looks good! I am going to kick off a test of restoring a 2GB restore file later today, the results of that should be available by tomorrow.

[Detailed testing notes](https://www.evernote.com/l/AAiIUyoi8jVPArVwE7XyUIN_-AUhUhvvL2U).

garrettr:
Tried to test a 2GB restore overnight last night, but it failed in the middle of the night. My computer probably tried to go to sleep during the transfer of the restore file. Trying it again now, and taking notes in the same Evernote from yesterday.

garrettr:
2GB restore succeeded! Detailed notes in the Evernote. Unfortunately, since this is a restore of the hardware test instance in our other office, I need to wait until tomorrow so I can shut down the test hardware instance in order to properly test the restore.

garrettr:
Rebased on top of `release/0.3.7` so I could write documentation for the new-and-improved backup and restore procedures for Sphinx/ReadTheDocs, which needed the backported ReadTheDocs documentation added to `release/0.3.7` in b39bf3868e2fe8bad8d5813951893068e239f85e. Had to force push the feature branch as a result of the rebase.

@conorsch I'd appreciate a review of the backup and restore documentation added in 90908aa3c9bf974525e17e5f0583b4c3e3ff04f9. You can pull it locally and build the documentation with `make html` or `sphinx auto-build`. For your convenience, I also published a copy on ReadTheDocs here: https://docs.securedrop.org/en/improve-backup-and-restore/backup_and_restore.html.

conorsch:
@garrettr Reviewed docs, pushed up some revisions. Mostly the changes strive for concision. There are a few niggling details that I'd like to improve about the docs, but are out of scope for this PR:

* SSH trouble shooting steps belong in **Test the installation**, to make docs DRY
* we're still requiring Admins to provide `-i inventory` on the command line when we can store that in the `ansible.cfg` file and save them the trouble

Give my edits a sanity check, but with their inclusion I'm happy with merging. :+1: Thanks for testing so extensively!

garrettr:
> There are a few niggling details that I'd like to improve about the docs, but are out of scope for this PR

Agreed; let's file follow-up issues for these. Good ideas, but we should stay on target to get 0.3.7 released ASAP.

conorsch:
> Good ideas, but we should stay on target to get 0.3.7 released ASAP.

:+1: :tada: :shipit: :shipit: :shipit:

garrettr:
> Give my edits a sanity check, but with their inclusion I'm happy with merging.

Sanity checked, lgtm. I added a note about how you may have to wait between deleting submissions on the Document Interface and actually reclaiming space in `/var/lib/securedrop/store` (due to file deletion being asynchronous, queued, and slow due the use of `srm`).

This is ready to merge.

garrettr:
> This is ready to merge.

Hold up. There's one more piece that @conorsch and I discussed this week that still needs some thought. Currently, the backup role runs by default whenever the full playbook is run. This could be problematic for production installations that have a lot of submissions.

If they don't plan ahead for the backup (in terms of available disk space on the Admin Workstation, time needed to transfer the backup archive over Tor, etc.), this could be a bad experience for them if they need to run a playbook for a reason other than backup/restore - e.g. we ask them to run the playbook for an upgrade that requires Ansible, such as 0.3.6).

I think we should consider _not_ running the backup role by default, so it _only_ gets run when the admin explicitly chooses to do so (following the new documentation introduced in 90908aa). @conorsch What do you think?

conorsch:
> I think we should consider _not_ running the backup role by default, so it _only_ gets run when the admin explicitly chooses to do so

Agreed, that's logical behavior, and will be particularly useful for long-running instances where too-frequent backups can clutter up the Admin Workstation. Worse, the long-running backups may discourage Admins from running the playbooks frequently.

The ideal solution to achieve the only-when-requested backup behavior is an isolated playbook encapsulating the backup and restore logic. That addition is best deferred until 0.4, however, due to substantial changes from the layout of the playbooks and roles that has already occurred in develop. As an intermediate solution, we can disable the backup role by default at the play level within `securedrop-prod.yml`, and reenable it conditionally if either `perform_backup=true` or `restore_file` is defined. 

14cd246 implements the intermediate solution, and tweaks the docs for upgrade accordingly. This should be tested on hardware again.

garrettr:
Awesome. I think this is finally done!!

conorsch:
Tracking a list of issues that will be resolved once this PR is merged:

* closes  #942
* closes #1127 
* closes #1263 

There are a few "related" issues that have already been linked via discussion. Not marking those for automatic close upon merge.

Issue #844, opened 2015-01-23
Intermittent Travis test failures (Directory not empty: '/tmp/securedrop/keys')


garrettr:
Recently it seems to me that Travis has begun failing more often than it should. Most of the failures are not actual test failures, but appear to be related to the setup and teardown for each test. Re-running the Travis build usually results in all tests passing.

I will post relevant logs in comments on this issue as they occur. It does not appear to be possible to recover the logs from a failed build once you've restarted it :disappointed:

garrettr:
Here's an example log for one of these intermittent failures: https://gist.github.com/garrettr/65271f6b1c31b03487a3

garrettr:
Another log for a similar intermittent failure: https://gist.github.com/garrettr/f0a0a603ee025ba16223

garrettr:
Here's another example, which looks like it might be a different flavor than the first two: https://gist.github.com/garrettr/04d58b02489c185c3081

garrettr:
Sometimes Pip is flaky too: https://gist.github.com/garrettr/dbe8e77b91fa6db1c4f9. I'm not sure if there's anything we could do about that (besides switching to a better CI system that allows us to use custom base images...)

garrettr:
Another rmtree-related intermittent test failure: https://gist.github.com/garrettr/ad6c791748abc310d267

garrettr:
And another one: https://gist.github.com/garrettr/3b778b308213f19c8e3b

garrettr:
This is an interesting one, in test_regenerate_valid_lengths: https://gist.github.com/garrettr/28f00ae6a2f45d44692a

garrettr:
Adding p-high, since this makes the CI less valuable and repeatedly wastes developer time.

garrettr:
Started looking at this today. There are approximately 3 distinct classes of errors reported in this issue, but the most common by far is errors in `shutil.rmtree` when it's called from `common.clean_root`.

Here's a [recent traceback](https://gist.github.com/garrettr/90c2ffb142f3ee69da20). I have a theory about what's going on here. I think this is a race condition caused by python-gnupg's use of daemon threads, similar to my theory about the cause of [this other intermittent test issue](https://github.com/freedomofpress/securedrop/issues/618#issuecomment-60293098).

Here's what I think happens:

[...]

PR #1276, opened 2016-04-15
Human friendly diceware wordlist

### LINKED ISSUE #1235: Totally unmemorizable stuff in the wordlist
With high probability a source get a phrase-like codename consisting of words. But sometimes one can get something like
`crest sill 86 crt plushy 13th rebut`.
Numbers are easy to forget. Words leave some trace in human memory (one can remember a word by its length/shape/etc). Numbers are not.
It's all because of [`wordlist`](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/wordlist#L6515). A possible fix is to replace all numbers with words generated additionally.

heartsucker:
Trading decreased entropy for increased memorability.

Closes #1235


garrettr:
Hey, @heartsucker, sorry for the delay on reviewing this! This looks like a really solid new wordlist to me.

> Trading decreased entropy for increased memorability.

This wordlist actually increases the entropy of the generated passphrases, because it has 8192 words and the current list only has 6800 words.

However, it's important to note that the SecureDrop wordlist has been filtered over time for a number of reasons, primarily in order to:

1. Remove words that are strange, unusual, or hard to remember
2. Remove words that are offensive, on their own or potentially in combination with other words
 * For example, your new wordlist includes the word "jew", which is not offensive on its own but can be made into an offensive phrase with the addition of some of the other words in your list.
3. At one point we removed some words because they contained special characters (e.g. `;`, `'`, etc.), and we preferred to filter these out of codenames to avoid possible injection attacks. I am pretty sure this is not an issue anymore (if it ever really was), but it is something to keep in mind.

From a quick glance at your list, it looks like you've done a good job removing difficult words and you have totally avoided special characters. I think it would be good to do another pass and remove words that could be considered offensive in the context of other words (e.g. "jew").

It's hard for me to say how much you or we should worry about this. A lot of words can seem offensive depending on their context, it's probably quite hard to remove them all (and you do so at the expense of entropy). I mostly want to avoid regressions from the current wordlist. I fired up Python and took the set difference of you new list and the current list, and put it in a gist for you. These are the new words that are in your list that are not in the old list (and therefore could be regressions, deserving an extra bit of scrutiny): [gist](https://gist.github.com/garrettr/842ae53082b2d6920113e59ccf75fbcc).

You should also remove [words that start with numbers or are just numbers](https://github.com/heartsucker/securedrop/blob/new-wordlist/securedrop/dictionaries/diceware.txt#L1-L286). Consensus seems to be that these are harder to memorize than ordinary words, and I would agree. They are specifically called out in the issue that you are trying to resolve (#1235).

heartsucker:
@garrettr Ah, it looked like you were still pulling from the EOWL which has something like 120k words, so that's why I thought "decrease in entropy."

Anyway, I tried to filter the list to remove anything that had the potential to be offensive, but I guess I wasn't careful enough. I have this wordlist in a [repo](https://github.com/heartsucker/diceware) in case anyone ever wants to reuse it, so it might be worth being overly cautious about words that could be offensive. Did you see any others?

I'll try to get to the changes you suggested later this week. I did this by hand the first time around, so it might take a few days to get an update as this is... tedious. To say the least.

garrettr:
> I did this by hand the first time around, so it might take a few days to get an update as this is... tedious. To say the least.

@heartsucker No doubt! That's why we haven't gotten around to it yet - so your efforts here are _hugely_ appreciated.

> Anyway, I tried to filter the list to remove anything that had the potential to be offensive, but I guess I wasn't careful enough.

You did a really good job - I also (cursorily) checked your wordlist for words that we removed in the various efforts to get rid of offensive words, and most of them were not present.

If I were you, I'd just skim the gist, since it contains all the "new" words in your wordlist. That way, at least there won't be any regressions from the current wordlist. If there are still potentially offensive combinations of words after that, that's ok - we'll deal with them as we have in the past.

heartsucker:
@garrettr I went over this again. I decided to leave in words that could be taken offensively because once I started trying to removed them it because a ridiculous game of "is this slang to any demographic for anything bad?" And the answer happens to be a crap ton of single words and word pairs can be stretched into offensiveness. Anyway. Let me know what you think. (And it took forever because I had to recreate the list of words I rejected the first time around)

garrettr:
@heartsucker I think that's fine. Like I said, I already checked and there aren't any serious specific regressions compared to the current wordlist as far as offensive words go, so if there is any overall regression I think it will be minor.

The CI failed due to an intermittent error (#844), so I restarted them. Once they are confirmed to pass, I'll merge this because otherwise it lgtm.

garrettr:
Discussed merging this internally. We still have some minor reservations about the new wordlist because it contains words that have lead to unfortunately offensive random phrases in past (e.g. "jew", "jewish"), so it is a slight regression in that respect. However, we also sympathize with your experience where, as you said:

> I decided to leave in words that could be taken offensively because once I started trying to removed them it because a ridiculous game of "is this slang to any demographic for anything bad?"

Same thing happens to us whenever we try to take a crack at getting rid of "all of the words that could _potentially_ be offensive taken in context with _any other word_ in this list". 😄 

So, I thought we should merge it now and deal with any candidates for removal (due to offensiveness, or for any other reason) on a case-by-case basis in follow-up issues.

garrettr:
Overall, I think this wordlist tends to generate _much_ nicer Diceware passphrases than the previous list did. Here's a sampling of some random 7-word phrases (the default length in SecureDrop):

```
detects vendor affair menlo conrad credo trombone
covet unary bale typo toe version hay
demark sloth crushes rangy veer modem prime
amend mystic den ripe weather drone bases
cube suits cortex spider arrived lynx jiffy
mood spud irons california body omaha beirut
snap media nurse meets ditches beets irony
banks biscuit gauss pops martyr jasmine dusk
```

So thanks a lot, @heartsucker! 💯

PR #1277, opened 2016-04-18
Run NetworkManager hook before software upgrade

### LINKED ISSUE #1274: Run tails_files Network Manager hook earlier
Currently, `tails_files` installs a Network Manager hook on the Admin and Journalist Workstations. The goal of this hook is to update the Tails system-wide torrc with the HidServAuth values necessary to access various authenticated Tor Hidden Services, then reload Tor.

When it was originally merged, we made it the very last network manager hook to run, out of an abundance of caution. We were afraid it might interfere with some of the built-in Network Manager hooks if it ran ahead of them.

However, I have done some investigation and found that this choice interacts badly when `live-additional-software.conf` is used. It is often used on the **Admin Workstation** to automatically install `ansible`.

If you examine `/etc/NetworkManager/dispatcher.d`, you will see that `70-upgrade-additional-software.sh` comes before `99-tor-reload.sh` (which is the hook placed by tails_files). The problem is that upgrading additional software requires an `apt-get update` at the minimum, and potentially an `apt-get upgrade` as well. These operations can be very slow on Tails, depending on whether the speed of the network connection and whether an upgrade is required in addition to an update.

Since Network Manager hooks are run in order, the tails_files tor reload won't run until upgrade-additional-software is done. This seems to typically take at least 2-3 minutes, and potentially much longer. In the meantime, none of the authenticated Tor Hidden Services can be accessed.

We should move the `99-tor-reload.sh` hook so it comes _before_ the `70-upgrade-additional-software.sh` hook. Additionally, it should come _after_ the `60-*` hooks, which make sure Tor is ready (since it would be good to have Tor working before trying to customize its configuration and reload it).

ageis:
Fixes #1274


tails_files/install.sh
```diff
@@ -162,13 +162,20 @@ if [ -f $XSESSION_RC ]; then
     fi
 fi
 
+# remove previous NetworkManager hook if present. it runs later than we'd like
+OLD_NM_HOOK=$TAILSCFG/custom-nm-hooks/99-tor-reload.sh
+if [ -f $OLD_NM_HOOK ]; then
+    rm -f $OLD_NM_HOOK > /dev/null 2>&1
+    rm -f $INSTALL_DIR/99-tor-reload.sh > /dev/null 2>&1
+fi
```
conorsch: Also remove old script from `/etc/NetworkManager/dispatcher.d/`, since it may be there already as configured by a prior version of the init script.

Might be wise to drop the conditional block altogether: we want to ensure removal of the old script no matter what.

garrettr: > Might be wise to drop the conditional block altogether: we want to ensure removal of the old script no matter what.
 
You want to keep the conditional; otherwise `rm` will exit with non-zero and the script will be terminated due to its use of `set -e`.

tails_files/install.sh
```diff
@@ -53,7 +53,7 @@ cp securedrop_icon.png $INSTALL_DIR
 cp document.desktop $INSTALL_DIR
 cp source.desktop $INSTALL_DIR
 cp securedrop_init.py $SCRIPT_PY
-cp 99-tor-reload.sh $INSTALL_DIR
+cp 65-tor-reload.sh $INSTALL_DIR
```
garrettr: Minor request - can we name the hook so it clearly indicates it's part of the SecureDrop-specific customizations? e.g. "65-configure-tor-for-securedrop.sh`

garrettr:
@ageis What's the story for existing Admin Workstations to upgrade in order to benefit from this? Can they just pull the new release and re-run `tails_files/install.sh`, or will that break things?

From a quick skim on my end, it looks like it might be ok to just re-run `install.sh`. Can you verify?

tails_files/install.sh
conorsch: No, `rm -f` will ensure the command exits 0 even if the file does not exist.

garrettr: Good point. I still prefer the conditional. Makes the intent more clear - this is supposed to be a one-time thing, cleaning up from an old version, and only needs to happen if that file is present. "Better explicit than implicit".

tails_files/install.sh
conorsch: +1

tails_files/install.sh
conorsch: Under no circumstances should the file `/etc/NetworkManager/dispatcher.d/99-tor-reload.sh` exist. In lieu of Ansible's `state=absent` pragma for ensuring file absence, the best we have in a bash script is `rm -f <path>`.

Right now the script checks if `/home/amnesia/Persistent/.securedrop/99-tor-reload.sh` exists, and removes `/etc/NetworkManager/dispatcher.d/99-tor-reload.sh` **if-and-only-if** it does. I see the presumed association here, but the conditional logic is actually implicit, not explicit, since it assumes that the existence of a file is predicated on the existence of another.

It's an edge case, but if I ctrl+c this script at the right moment, I can create a system state where two redundant nm hooks are present. That's not good, and I definitely don't want to debug that state months from now.

I suggest refactoring the conditional block as:

```
# Remove previous NetworkManager hook if present. The "99-" prefix
# caused the hook to run too late for apt persistence to work reliably.
for d in $TAILSCFG $INSTALL_DIR /etc/NetworkManager/dispatcher.d ; do
    rm -f "$d/99-tor-reload.sh" > /dev/null 2>&1
done
```

It may be more explicit to check for file existence per file within the for-loop—personally I find that a bit obtuse, but it's mostly a stylistic question at that point, and the per-file checks would still ensure file absence as intended.

ageis:
@garrettr yes, the script logic is designed so that it can be re-run cleanly without problems.

garrettr:
This lgtm. @ageis Can you test on Tails 2.x? If it works, then this is ready to merge.

Note that the _goal_ of this PR is to avoid a significant delay between when:

1. The desktop appears
2. One is able to connect to the SecureDrop ATHSs

So make sure that is explicitly tested. An easy way to do it is to tail syslog once the desktop comes up - you should see an entry indicating that the SecureDrop network manager hook ran. As long as it runs without a noticeable delay (e.g. within a few seconds of the desktop appearing), then I would consider this to be a resolution for #1274.

ageis:
I've tested this on a fresh Tails 2.2.1 and it works. I counted 40 seconds before the desktop appeared, and the ATHSes were available as soon as Tor was ready.

I'm unable to provide detailed information on when exactly the hook ran, because there is no syslog in Tails 2 and the systemd journal doesn't contain logging of NetworkManager-dispatcher.service... I _am_ able to tell you that `live-additional-software` ran a mere ~3 seconds after a message indicating that the NM dispatcher itself had started, so the time taken between configuring/reloading Tor for SecureDrop to proceeding to installing packages from apt was quite slim.

PR #1285, opened 2016-05-04
No suid for securedrop init

garrettr:
Originally, the custom Tails configuration for SecureDrop was in a script run by the normal Tails user (amnesia). At first, it was run from a desktop launcher. Later, there was a (failed) experiment to auto-run the custom configuration from the amnesia user's `~/.xsessionrc`.

In both cases, the user running the script was the unprivileged `amnesia`, but it needed to modify the system-wide `/etc/tor/torrc` owned by root. The solution was to create a setuid wrapper in C for the configuration script (since you cannot setuid a script). Since the Tails configuration *install script* (`tails_files/install.sh`) was run as root, it could build and setuid on the binary wrapper that, when run, would setuid, then call the custom configuration script (as root).

After some discussion with the Tails developers, we learned that Tails' persistence had a secret option for custom NetworkManager hooks. This was obviously a natural place for our custom Tor configuration to go, so we migrated to that, which allowed us to auto-run the custom configuration for SecureDrop while avoiding the issues we were having with the `~/.xsessionrc`-based solution. We switched to using a NetworkManager script instead of `~/.xsessionrc~`, leaving everything else unchanged.

However, switching to a NetworkManager hook makes the setuid wrapper unnecessary, since these hooks are run as root by default. This pull request removes the setuid wrapper, which:

* Simplifies the SecureDrop custom configuration
* Greatly speeds up `tails_files/install.sh`, since there is no need to update packages and install gcc (just to build a 5-line C setuid wrapper binary... 😭)
* Avoids potential security risks from setuid programs

The changes to `securedrop_init.py` are just to remove `if __name__ == '__main__'`, which is unnecessary because this is a single use script and does not define anything that another script could conceivably import. There appear to be a lot of changes, but all I did was reindent most of the file after removing that unnecessary conditional.

This PR also removes some trailing whitespace that I encountered in this area of the code.

### DIFF
--- a/tails_files/securedrop_init.py
+++ b/tails_files/securedrop_init.py
@@ -1,50 +1,49 @@
-#!/usr/bin/env python
+#!/usr/bin/python
 
 import os
 import sys
 import subprocess
 
 
-if __name__ == '__main__':
-    # check for root
-    if os.geteuid() != 0:
-        sys.exit('You need to run this as root')
+# check for root
+if os.geteuid() != 0:
+    sys.exit('You need to run this as root')
 
-    # paths
-    path_torrc_additions = '/home/amnesia/Persistent/.securedrop/torrc_additions'
-    path_torrc_backup = '/etc/tor/torrc.bak'
-    path_torrc = '/etc/tor/torrc'
+# paths
+path_torrc_additions = '/home/amnesia/Persistent/.securedrop/torrc_additions'
+path_torrc_backup = '/etc/tor/torrc.bak'
+path_torrc = '/etc/tor/torrc'
 
-    # load torrc_additions
-    if os.path.isfile(path_torrc_additions):
-        torrc_additions = open(path_torrc_additions).read()
+# load torrc_additions
+if os.path.isfile(path_torrc_additions):
+    torrc_additions = open(path_torrc_additions).read()
+else:
+    sys.exit('Error opening {0} for reading'.format(path_torrc_additions))
+
+# load torrc
+if os.path.isfile(path_torrc_backup):
+    torrc = open(path_torrc_backup).read()
+else:
+    if os.path.isfile(path_torrc):
+        torrc = open(path_torrc).read()
     else:
-        sys.exit('Error opening {0} for reading'.format(path_torrc_additions))
+        sys.exit('Error opening {0} for reading'.format(path_torrc))
 
-    # load torrc
-    if os.path.isfile(path_torrc_backup):
-        torrc = open(path_torrc_backup).read()
-    else:
-        if os.path.isfile(path_torrc):
-            torrc = open(path_torrc).read()
-        else:
-            sys.exit('Error opening {0} for reading'.format(path_torrc))
+    # save a backup
+    open(path_torrc_backup, 'w').write(torrc)
 
-        # save a backup
-        open(path_torrc_backup, 'w').write(torrc)
+# append the additions
+open(path_torrc, 'w').write(torrc + torrc_additions)
 
-    # append the additions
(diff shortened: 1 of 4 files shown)


garrettr:
This needs to be tested on Tails 2.x. @ageis Would you mind?

garrettr:
@ageis In the context of re-running this script on an already-configured Admin Workstation: should I add some code to `install.sh` to cleanup the old setuid wrapper, if it exists (like we did with the old NetworkManager hooks after moving them to run earlier in #1277)?

ageis:
@garrettr I mean, sure, since we've been doing the same for the other components. Reminds me that it's a little annoying that we often have two or more copies of things since we use `.securedrop` as a staging directory.

Rationale for this pull request looks great. Will test-run tomorrow.

ageis:
Before running the `install.sh` script, Tor is ready and working. After `securedrop_init.py` is executed, the Tor service is in an inactive and non-running state:
```
Job for tor.service failed. See 'systemctl status tor.service' and 'journalctl -xn' for details.
```
```
$ systemctl status tor.service
● tor.service - Anonymizing overlay network for TCP (multi-instance-master)
   Loaded: loaded (/lib/systemd/system/tor.service; disabled)
   Active: inactive (dead)

May 04 21:01:12 amnesia systemd[1]: Unit tor.service cannot be reloaded because it is inactive.
```
If Tor is then started manually, `securedrop_init.py` can be run with no problems. The problem is that it's trying to do a service reload after the service has died/been stopped for some as-yet unknown reason.
The torrc_additions are successfully appended to torrc and the configuration appears valid.

garrettr:
> Before running the install.sh script, Tor is ready and working.

There are a couple of confusing things about this.

If you boot Tails and run `systemctl status tor.service` (before doing anything else, including running `install.sh`), it says:

```
$ systemctl status tor.service
● tor.service - Anonymizing overlay network for TCP (multi-instance-master)
   Loaded: loaded (/lib/systemd/system/tor.service; disabled)
   Active: inactive (dead)
```

It says this even after Tails pops a "Tor is ready" notification, and I can access websites in the Tor Browser.

Seems like the systemctl status for the tor service might be broken :/

garrettr:
Actually, looks like Tails' systemd configuration for Tor (targets and services) just differs from the the default configuration. My gut feeling is that while `subprocess.call(['/usr/sbin/service', 'tor', 'reload'])` worked in Tails 1.x, it might not work in Tails 2.x due to these changes.

I'm currently exploring https://labs.riseup.net/code/issues/9393 to try to better understand the Tails systemd setup for Tor.

garrettr:
Looks like the correct way to reload system Tor on Tails 2.x is `systemctl reload tor@default.service`.

tails_files/securedrop_init.py
```diff
@@ -36,14 +36,12 @@
...
-                 'You can now connect to your SecureDrop\ndocument interface.'])
+try:
+    subprocess.check_call(['systemctl', 'reload', 'tor@default.service'])
+except subprocess.CalledProcessError:
+    sys.exit('Error reloading Tor')
+
+# notify the user
+subprocess.call(['tails-notify-user',
+                 'SecureDrop successfully auto-configured!',
+                 'You can now access the Document Interface.\nIf you are an admin, you can now SSH to the servers.'])
```
ageis: Works like a charm.

garrettr:
Ok, according to @ageis this PR is working so I am going to merge it. This PR has the 0.3.7 milestone, but I am going to merge it into develop because we decided that since there have been many changes to `tails_files` between 0.3.6 and develop, we should just backport the current `tails_files` from develop into 0.3.7 instead of trying to cherry-pick each individual change.

PR #1287, opened 2016-05-05
Use --force option with copy commands in tails_files

### LINKED ISSUE #1286: tails_files upgrade/re-run: existing file conflict with cp and set -e
If `install.sh` is executed a second time, the script will fail after the following message:
```
cp: ‘/home/amnesia/Desktop/document.desktop’ and ‘/live/persistence/TailsData_unlocked/dotfiles/Desktop/document.desktop’ are the same file
```

Also, this is more of an edge case, but if I manually clean up the desktop shortcuts in persistence (`/live/persistence/TailsData_unlocked/`) but forget to remove the symlinks in `~/Desktop`:
```
cp: not writing through dangling symlink ‘/home/amnesia/Desktop/document.desktop’
```

cp has a number of options that may be helpful in the first case. I think --update is probably the right one.

```
       -u, --update
              copy only when the SOURCE file is newer than the destination file or when the destination file is missing

       --remove-destination
              remove each existing destination file before attempting to open it (contrast with --force)

       -f, --force
              if an existing destination file cannot be opened, remove it and try again (this option is ignored when the -n option is also used)

       -n, --no-clobber
              do not overwrite an existing file (overrides a previous -i option)

```

ageis:
When cp is used in combination with set -e, and the source and destination are the same (as in an upgrade or re-run use case), the command and rest of the script will fail.

Intended to fix #1286. 

It is needed on lines 138 to 141 for sure. But this commit applies it liberally to all copy commands.

### DIFF
--- a/tails_files/install.sh
+++ b/tails_files/install.sh
@@ -35,7 +35,7 @@ if [ ! -d "$ANSIBLE" ]; then
 fi
 
 # detect whether admin or journalist
-if [ -f $ANSIBLE/app-document-aths ]; then
+if [ -f $ANSIBLE/app-ssh-aths ]; then
   ADMIN=true
 else
   ADMIN=false
--- a/tails_files/install.sh
+++ b/tails_files/install.sh
@@ -100,8 +100,6 @@ chmod 400 $ADDITIONS
 
 chown amnesia:amnesia $INSTALL_DIR/securedrop_icon.png
 chmod 600 $INSTALL_DIR/securedrop_icon.png
-chown amnesia:amnesia $INSTALL_DIR/document.desktop $INSTALL_DIR/source.desktop
-chmod 700 $INSTALL_DIR/document.desktop $INSTALL_DIR/source.desktop
 
 # journalist workstation does not have the *-aths files created by the Ansible playbook, so we must prompt
 # to get the interface .onion addresses to setup launchers, and for the HidServAuth info used by Tor
@@ -122,18 +120,28 @@ echo "Exec=/usr/local/bin/tor-browser $DOCUMENT" >> $INSTALL_DIR/document.deskto
 echo "Exec=/usr/local/bin/tor-browser $SOURCE" >> $INSTALL_DIR/source.desktop
 
 # copy launchers to desktop and Applications menu
-cp -pf $INSTALL_DIR/document.desktop $DESKTOP
-cp -pf $INSTALL_DIR/source.desktop $DESKTOP
-cp -pf $INSTALL_DIR/document.desktop $HOMEDIR/.local/share/applications
-cp -pf $INSTALL_DIR/source.desktop $HOMEDIR/.local/share/applications
+cp -f $INSTALL_DIR/document.desktop $DESKTOP
+cp -f $INSTALL_DIR/source.desktop $DESKTOP
+cp -f $INSTALL_DIR/document.desktop $HOMEDIR/.local/share/applications
+cp -f $INSTALL_DIR/source.desktop $HOMEDIR/.local/share/applications
 
 # make it all persistent
 sudo -u amnesia mkdir -p $DOTFILES/Desktop
 sudo -u amnesia mkdir -p $DOTFILES/.local/share/applications
-cp -pf $DESKTOP/document.desktop $DOTFILES/Desktop
-cp -pf $DESKTOP/source.desktop $DOTFILES/Desktop
-cp -pf $DESKTOP/document.desktop $DOTFILES/.local/share/applications
-cp -pf $DESKTOP/source.desktop $DOTFILES/.local/share/applications
+cp -f $DESKTOP/document.desktop $DOTFILES/Desktop
+cp -f $DESKTOP/source.desktop $DOTFILES/Desktop
+cp -f $DESKTOP/document.desktop $DOTFILES/.local/share/applications
+cp -f $DESKTOP/source.desktop $DOTFILES/.local/share/applications
+
+# set ownership and permissions
+chown amnesia:amnesia $DESKTOP/document.desktop $DESKTOP/source.desktop \
+  $DOTFILES/Desktop/document.desktop $DOTFILES/Desktop/source.desktop \
+  $HOMEDIR/.local/share/applications/document.desktop $HOMEDIR/.local/share/applications/source.desktop \
+  $DOTFILES/.local/share/applications/document.desktop $DOTFILES/.local/share/applications/source.desktop
+chmod 700 $DESKTOP/document.desktop $DESKTOP/source.desktop \
+  $DOTFILES/Desktop/document.desktop $DOTFILES/Desktop/source.desktop \
+  $HOMEDIR/.local/share/applications/document.desktop $HOMEDIR/.local/share/applications/source.desktop \
+  $DOTFILES/.local/share/applications/document.desktop $DOTFILES/.local/share/applications/source.desktop
 
 # remove xsessionrc from 0.3.2 if present
 XSESSION_RC=$TAILSCFG/dotfiles/.xsessionrc
--- a/tails_files/install.sh
+++ b/tails_files/install.sh
@@ -43,11 +43,11 @@ fi
 mkdir -p $INSTALL_DIR
 
 # copy icon, launchers and scripts
-cp securedrop_icon.png $INSTALL_DIR
-cp document.desktop $INSTALL_DIR
-cp source.desktop $INSTALL_DIR
-cp securedrop_init.py $SCRIPT_PY
-cp 65-configure-tor-for-securedrop.sh $INSTALL_DIR
+cp -f securedrop_icon.png $INSTALL_DIR
+cp -f document.desktop $INSTALL_DIR
+cp -f source.desktop $INSTALL_DIR
+cp -f securedrop_init.py $SCRIPT_PY
+cp -f 65-configure-tor-for-securedrop.sh $INSTALL_DIR
 
 # Remove binary setuid wrapper from previous tails_files installation, if it exists
 WRAPPER_BIN=$INSTALL_DIR/securedrop_init
@@ -74,7 +74,7 @@ Host mon
 EOL
     chown amnesia:amnesia $INSTALL_DIR/ssh_config
     chmod 600 $INSTALL_DIR/ssh_config
-    cp -p $INSTALL_DIR/ssh_config $HOMEDIR/.ssh/config
+    cp -pf $INSTALL_DIR/ssh_config $HOMEDIR/.ssh/config
     SSH_ALIASES=true
   fi
   # set ansible to auto-install
@@ -88,7 +88,7 @@ EOL
   fi
 else
   # prepare torrc_additions (journalist)
-  cp torrc_additions $ADDITIONS
+  cp -f torrc_additions $ADDITIONS
 fi
 
 # set permissions
@@ -124,18 +124,18 @@ echo "Exec=/usr/local/bin/tor-browser $DOCUMENT" >> $INSTALL_DIR/document.deskto
 echo "Exec=/usr/local/bin/tor-browser $SOURCE" >> $INSTALL_DIR/source.desktop
 
 # copy launchers to desktop and Applications menu
-cp -p $INSTALL_DIR/document.desktop $DESKTOP
-cp -p $INSTALL_DIR/source.desktop $DESKTOP
-cp -p $INSTALL_DIR/document.desktop $HOMEDIR/.local/share/applications
-cp -p $INSTALL_DIR/source.desktop $HOMEDIR/.local/share/applications
+cp -pf $INSTALL_DIR/document.desktop $DESKTOP
+cp -pf $INSTALL_DIR/source.desktop $DESKTOP
+cp -pf $INSTALL_DIR/document.desktop $HOMEDIR/.local/share/applications
+cp -pf $INSTALL_DIR/source.desktop $HOMEDIR/.local/share/applications
 
 # make it all persistent
 sudo -u amnesia mkdir -p $DOTFILES/Desktop
 sudo -u amnesia mkdir -p $DOTFILES/.local/share/applications
-cp -p $DESKTOP/document.desktop $DOTFILES/Desktop
-cp -p $DESKTOP/source.desktop $DOTFILES/Desktop
-cp -p $DESKTOP/document.desktop $DOTFILES/.local/share/applications
-cp -p $DESKTOP/source.desktop $DOTFILES/.local/share/applications
+cp -pf $DESKTOP/document.desktop $DOTFILES/Desktop
+cp -pf $DESKTOP/source.desktop $DOTFILES/Desktop
+cp -pf $DESKTOP/document.desktop $DOTFILES/.local/share/applications
+cp -pf $DESKTOP/source.desktop $DOTFILES/.local/share/applications
 
 # remove xsessionrc from 0.3.2 if present
 XSESSION_RC=$TAILSCFG/dotfiles/.xsessionrc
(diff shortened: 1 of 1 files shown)


garrettr:
lgtm, @conorsch what do you think?

conorsch:
Looks good, I agree that `-f` is warranted throughout.

garrettr:
Just a comment: I'm not totally in love with the use of `cp -p` in this script. It's not immediately clear why it's only used in some places and not others.

The reason is because `install.sh` is run as root, so by default copied files end up owned by root. That is good, even necessary for some of these files. Other files should be owned by `amnesia`, and there are some files that _must_ be to ensure correct functionality (e.g. `~/.ssh/config`).

If you follow the documentation, you `git clone` the SecureDrop repo as the amnesia user, so all of the files within it are automatically owned by amnesia. Therefore, `cp -p` makes the copied files owned by amnesia as well.

This works just fine, and is such a minor issue that I don't really feel like it warrants any changes. I just wanted to explain why it's being done, since I couldn't find an explanation in comments or anywhere in git. In general, I would say "explicit is better than implicit" and prefer an approach that explicitly sets ownership and permissions, rather than this one, which makes assumptions about the ownership of files in the git repo.

Also, there are some files that are copied with `-p` that probably don't need to be/shouldn't be. For example, @ageis why are the network manager hooks copied with `-p`?

ageis:
@garrettr You make a good point about "explicit is better than implicit". We were using `-p` cause we're doing some chown operations for files that need specific permissions or ownership applied. In the case of the NetworkManager hook, that does need to be copied with `-p` since it's set to 755 and marked executable, like the rest of the contents of `dispatcher.d`. Otherwise it would be 644 and wouldn't work. There might be instances where we don't need it but from a quick review I don't immediately see any.

garrettr:
@ageis https://github.com/freedomofpress/securedrop/pull/1287#issuecomment-217993861 doesn't make a lot of sense to me. How is `cp -p` better than copying, than explicitly setting ownership/permissions where necessary, in any scenario?

ageis:
@garrettr I think in this case it needs to be beared in mind that we move stuff to a persistent/staging installation directory, set the permissions there, and then move them from there onto other places. I think that's the process that should be re-examined, because there are files we don't really need living in `.securedrop`.

garrettr:
@ageis If there are files that don't need to be copied to `.securedrop`, then... fix it?

garrettr:
This PR still needs:

- [ ] Research: Avoid redundant files in `.securedrop`
- [ ] Fix: Set ownership and permissions explicitly (with `chown`/`chmod`) instead of implicitly (with `cp -p`).

ageis:
See latest commits. The desktop shortcuts in particular need to be **modified in place** (to add the source and document URLs), so they remain in `.securedrop`. But the NetworkManager hook doesn't need to be there, so it's been removed. There are cases where it seems more efficient, and less lines, to set ownership+permission once, in `.securedrop`, and simply use `cp -p` when copying onward to multiple destinations. But anyway, I think this is what you were looking for.

tails_files/install.sh
```diff
@@ -74,7 +74,7 @@ Host mon
 EOL
     chown amnesia:amnesia $INSTALL_DIR/ssh_config
     chmod 600 $INSTALL_DIR/ssh_config
-    cp -p $INSTALL_DIR/ssh_config $HOMEDIR/.ssh/config
+    cp -pf $INSTALL_DIR/ssh_config $HOMEDIR/.ssh/config
```
garrettr: Does the ssh config need to be copied to the `$INSTALL_DIR`? AFAICT it's already persisted because the [SSH persistence](https://tails.boum.org/doc/first_steps/persistence/configure/index.en.html#index3h2) includes `~/.ssh/config/`.

garrettr:
This needs to be rebased on top of the latest develop.

tails_files/install.sh
ageis: I figure since it is also prepared and modified in place (adding the hostname and username) then it makes sense to have it in the install dir. If you don't feel that's apt then we could change it to edit the .ssh/config directly.

conorsch:
Ran through a test of this on hardware. (More truthfully, I tested all `tails_files`-related changes staged for 0.3.7.) 

```
amnesia@amnesia:~/Persistent/securedrop/tails_files$ sudo ./install.sh 
[sudo] password for amnesia: 
cp: ‘/home/amnesia/Desktop/document.desktop’ and ‘/live/persistence/TailsData_unlocked/dotfiles/Desktop/document.desktop’ are the same file
```

I edited the `tails_files/install.sh` script and added `set -x` at the top to make it more verbose:

```
amnesia@amnesia:~/Persistent/securedrop/tails_files$ sudo ./install.sh 
[sudo] password for amnesia: 
+ [[ 0 -ne 0 ]]
+ HOMEDIR=/home/amnesia
+ PERSISTENT=/home/amnesia/Persistent
+ INSTALL_DIR=/home/amnesia/Persistent/.securedrop
+ ADDITIONS=/home/amnesia/Persistent/.securedrop/torrc_additions
+ SCRIPT_PY=/home/amnesia/Persistent/.securedrop/securedrop_init.py
+ TAILSCFG=/live/persistence/TailsData_unlocked
+ DOTFILES=/live/persistence/TailsData_unlocked/dotfiles
+ DESKTOP=/home/amnesia/Desktop
+ ANSIBLE=/home/amnesia/Persistent/securedrop/install_files/ansible-base
+ NMDISPATCHER=/etc/NetworkManager/dispatcher.d
+ SSH_ALIASES=false
+ source /etc/os-release
++ TAILS_PRODUCT_NAME=Tails
++ TAILS_VERSION_ID=2.3
+ [[ 2.3 =~ ^1\..* ]]
+ '[' '!' -d /live/persistence/TailsData_unlocked ']'
+ '[' '!' -d /home/amnesia/Persistent/securedrop/install_files/ansible-base ']'
+ '[' -f /home/amnesia/Persistent/securedrop/install_files/ansible-base/app-document-aths ']'
+ ADMIN=true
+ mkdir -p /home/amnesia/Persistent/.securedrop
+ cp -f securedrop_icon.png /home/amnesia/Persistent/.securedrop
+ cp -f document.desktop /home/amnesia/Persistent/.securedrop
+ cp -f source.desktop /home/amnesia/Persistent/.securedrop
+ cp -f securedrop_init.py /home/amnesia/Persistent/.securedrop/securedrop_init.py
+ WRAPPER_BIN=/home/amnesia/Persistent/.securedrop/securedrop_init
+ '[' -f /home/amnesia/Persistent/.securedrop/securedrop_init ']'
+ true
++ cat /home/amnesia/Persistent/securedrop/install_files/ansible-base/app-document-aths
++ cut -d ' ' -f 2
+ DOCUMENT=tyzwhx6om3vzksvj.onion
++ cat /home/amnesia/Persistent/securedrop/install_files/ansible-base/app-source-ths
+ SOURCE=rbted2wcey6yc2yv.onion
++ cat /home/amnesia/Persistent/securedrop/install_files/ansible-base/app-ssh-aths
++ cut -d ' ' -f 2
+ APPSSH=rgk4z2f3yrj7ou4o.onion
++ cat /home/amnesia/Persistent/securedrop/install_files/ansible-base/mon-ssh-aths
++ cut -d ' ' -f 2
+ MONSSH=5w7k3e553tgkko5u.onion
+ echo '# HidServAuth lines for SecureDrop'\''s authenticated hidden services'
+ cat - /home/amnesia/Persistent/securedrop/install_files/ansible-base/app-ssh-aths /home/amnesia/Persistent/securedrop/install_files/ansible-base/mon-ssh-aths /home/amnesia/Persistent/securedrop/install_files/ansible-base/app-document-aths
+ [[ -d /home/amnesia/.ssh ]]
+ [[ ! -f /home/amnesia/.ssh/config ]]
+ grep -q ansible /live/persistence/TailsData_unlocked/live-additional-software.conf
+ grep -v '^#.*onion' /home/amnesia/Persistent/securedrop/install_files/ansible-base/inventory
+ grep -q onion
+ chmod 755 /home/amnesia/Persistent/.securedrop
+ chown root:root /home/amnesia/Persistent/.securedrop/securedrop_init.py
+ chmod 700 /home/amnesia/Persistent/.securedrop/securedrop_init.py
+ chown root:root /home/amnesia/Persistent/.securedrop/torrc_additions
+ chmod 400 /home/amnesia/Persistent/.securedrop/torrc_additions
+ chown amnesia:amnesia /home/amnesia/Persistent/.securedrop/securedrop_icon.png
+ chmod 600 /home/amnesia/Persistent/.securedrop/securedrop_icon.png
+ true
+ echo 'Exec=/usr/local/bin/tor-browser tyzwhx6om3vzksvj.onion'
+ echo 'Exec=/usr/local/bin/tor-browser rbted2wcey6yc2yv.onion'
+ cp -f /home/amnesia/Persistent/.securedrop/document.desktop /home/amnesia/Desktop
+ cp -f /home/amnesia/Persistent/.securedrop/source.desktop /home/amnesia/Desktop
+ cp -f /home/amnesia/Persistent/.securedrop/document.desktop /home/amnesia/.local/share/applications
+ cp -f /home/amnesia/Persistent/.securedrop/source.desktop /home/amnesia/.local/share/applications
+ sudo -u amnesia mkdir -p /live/persistence/TailsData_unlocked/dotfiles/Desktop
+ sudo -u amnesia mkdir -p /live/persistence/TailsData_unlocked/dotfiles/.local/share/applications
+ cp -f /home/amnesia/Desktop/document.desktop /live/persistence/TailsData_unlocked/dotfiles/Desktop
cp: ‘/home/amnesia/Desktop/document.desktop’ and ‘/live/persistence/TailsData_unlocked/dotfiles/Desktop/document.desktop’ are the same file
```

Looks like we're trying to copy a symlink to a real file, and `cp` is appropriately balking at that request. Needs further testing.

ageis:
Yup, I just got around to testing all the latest stuff, and can confirm `cp
-f` is not sufficient, probably due to the symlink. Researching other
options now.

ageis:
Okay, all issues are resolved in the latest commits and the script completes successfully for me.

ageis:
Just a note of not much consequence — I did some testing to examine other assumptions within the script. One of them: why copy the shortcuts directly to persistence, because maybe the contents `~/Desktop` and `~/.local/share/applications` are persisted automatically as part of dotfiles? It turned out they are not, so no changes are warranted. Another thought I had: the shortcut customization uses an append, so won't it clobber on subsequent re-runs? No, it's fine since the fresh shortcut template is first copied out of the Git repo upon every execution.

conorsch:
### Admin Workstation Story: :+1:

The recent commits to this PR resolve issues on the Admin Workstation. The NetworkManager hooks are properly updated; before:

```
1 -rwxr-xr-x  1 root root  180 Apr 25 15:33 00-firewall.sh
1 -rwxr-xr-x  1 root root  313 Apr 25 15:33 00-save-env
3 -rwxr-xr-x  1 root root 2146 Mar 31  2015 01ifupdown
1 -rwxr-xr-x  1 root root  406 Apr 25 15:33 01-wait-for-notification-recipient.sh
4 -rwxr-xr-x  1 root root 3096 Apr 25 15:33 10-tor.sh
7 -rwxr-xr-x  1 root root 6997 Apr 25 15:33 20-time.sh
1 -rwxr-xr-x  1 root root  389 Apr 25 15:33 30-i2p.sh
2 -rwxr-xr-x  1 root root 1171 Apr 25 15:33 60-tor-ready.sh
1 -rwxr-xr-x  1 root root  224 Apr 25 15:33 60-ttdnsd.sh
1 -rwxr-xr-x  1 root root  250 Apr 25 15:33 70-upgrade-additional-software.sh
0 lrwxrwxrwx  1 root root   79 May 16 16:19 99-tor-reload.sh -> /lib/live/mount/persistence/TailsData_unlocked/custom-nm-hooks/99-tor-reload.sh
0 drwxr-xr-x  2 root root    3 Mar 31  2015 pre-down.d
0 drwxr-xr-x  2 root root    3 Mar 31  2015 pre-up.d
```

After:

```
1 -rwxr-xr-x  1 root root  180 Apr 25 15:33 00-firewall.sh
1 -rwxr-xr-x  1 root root  313 Apr 25 15:33 00-save-env
3 -rwxr-xr-x  1 root root 2146 Mar 31  2015 01ifupdown
1 -rwxr-xr-x  1 root root  406 Apr 25 15:33 01-wait-for-notification-recipient.sh
4 -rwxr-xr-x  1 root root 3096 Apr 25 15:33 10-tor.sh
7 -rwxr-xr-x  1 root root 6997 Apr 25 15:33 20-time.sh
1 -rwxr-xr-x  1 root root  389 Apr 25 15:33 30-i2p.sh
2 -rwxr-xr-x  1 root root 1171 Apr 25 15:33 60-tor-ready.sh
1 -rwxr-xr-x  1 root root  224 Apr 25 15:33 60-ttdnsd.sh
4 -rwxr-xr-x  1 root root  372 May 16 17:35 65-configure-tor-for-securedrop.sh
1 -rwxr-xr-x  1 root root  250 Apr 25 15:33 70-upgrade-additional-software.sh
0 drwxr-xr-x  2 root root    3 Mar 31  2015 pre-down.d
0 drwxr-xr-x  2 root root    3 Mar 31  2015 pre-up.d
```

The desktop icons for the Source and Document Interfaces now display noticeably faster upon boot, since they're no longer blocked by the apt hook. So the Admin story is looking good.

### Journalist Workstation Story: :-1:
Testing on the Journalist Workstation wasn't as smooth. The script was fooled into thinking I was testing on an Admin Workstation, an assumption that caused the script to fail:

```
$ sudo ./install.sh 
[sudo] password for amnesia: 
cat: /home/amnesia/Persistent/securedrop/install_files/ansible-base/app-ssh-aths: No such file or directory
cat: /home/amnesia/Persistent/securedrop/install_files/ansible-base/mon-ssh-aths: No such file or directory
cat: /home/amnesia/Persistent/securedrop/install_files/ansible-base/app-ssh-aths: No such file or directory
cat: /home/amnesia/Persistent/securedrop/install_files/ansible-base/mon-ssh-aths: No such file or directory
amnesia@amnesia:~/Persistent/securedrop/tails_files$ echo $?
1
```

This is caused by the conditional check:

```
# detect whether admin or journalist
if [ -f $ANSIBLE/app-document-aths ]; then
  ADMIN=true
else
  ADMIN=false
fi
```

The Journalist Workstation is likely to have the `app-document-aths` file, since it needed to be copied over to the JW during initial provisioning, so we should not expect its absence. Updating the if statement to check for existence of `app-ssh-aths` makes more sense, and causes the script to correctly identify the JW. 

The script then proceeds to prompt for the ATHS info, which is completely unnecessary and confusing, since the HidServAuth values have already been populated in the torrc by the previous versions of securedrop dotfiles. It would be far better to read the value if already present and avoid interactively prompting journalists to paste the information in. The zenity prompt is sudden and offers no explanation of how to look up the information requested, which will surely lead to many support requests (and/or beleaguered Admins).

I'll keep testing, just wanted to document the results so far.

ageis:
> Updating the if statement to check for existence of app-ssh-aths makes more sense, and causes the script to correctly identify the JW.
Agree completely.

garrettr:
In favor, as I always am, of explicit over implicit: how about, instead of trying to guess the type of workstation (admin or journalist) based on the files in `install_files/ansible-base`, we just require them to pass that as an argument, e.g. `./tails_files/install.sh admin` or `./tails_files/install.sh journalist`?

conorsch:
@garrettr You read my mind! :joy: A case statement that inspects `$1` is a great possible option here: if no args are provided, it can calmly state that one of the arguments is required. I'd like to try to minimize the copy/paste required for journalists, though, so I'll focus on that. (Right now the script is actually a step backwards for journalists in that they have to copy/paste the ATHS values even if they're already in place.)

garrettr:
Superseded by #1309.

PR #1289, opened 2016-05-06
Change yml config file paths to allow for copy-paste from dev docs

fowlslegs:
Perhaps in the past the install_files dir was a subdir of the securedrop dir...

### DIFF
--- a/docs/development/getting_started.rst
+++ b/docs/development/getting_started.rst
@@ -199,7 +199,7 @@ across the full stack.
 
 If you want to receive OSSEC alerts or change any other settings, you will need
 to fill out your local copy of
-``securedrop/install_files/ansible_base/staging-specific.yml``.
+``./install_files/ansible_base/staging-specific.yml``.
 
 .. code:: sh
 
@@ -214,7 +214,7 @@ Prod
 ~~~~
 
 You will need to fill out the production configuration file:
-``securedrop/install_files/ansible_base/prod-specific.yml``.  Part of the
+``./install_files/ansible_base/prod-specific.yml``.  Part of the
 production playbook validates that staging values are not used in
 production. One of the values it verifies is that the user Ansible runs as is
 not ``vagrant`` To be able to run this playbook in a virtualized environment


garrettr:
@fowlslegs It was not. I think that doc is just writing the path _including_ the repo root, as opposed to starting from it. Not sure if there is a consistent convention throughout the docs, or if I care enough to try to enforce one. This change seems reasonable, however.

garrettr:
Not merging now because I'm not sure how to deal with whether this should go into release/0.3.7 or develop or what.

fowlslegs:
It's mixed (i.e., no convention), but to me it makes sense to make the files/ paths relative to being inside the top level repo dir versus inside the parent dir of the top level repo dir. Besides the fact this avoids the confusion of there being  `securedrop/securedrop` dir (which becomes `./securedrop`), it also makes sense to cd into the top level dir of a repo you're going to be working with, and from there you should be able to  copy-paste the filenames/ commands from the dev docs.

I think it should be merged into 0.3.7 as with all documentation fixes that are not related to features that have not been backported into a stable release or the current stable rc from develop.

conorsch:
> It's mixed (i.e., no convention)

We should solve the issue throughout the docs, then. At the very least, throughout the developer docs.

> it makes sense to make the files/ paths relative to being inside the top level repo dir versus inside the parent dir of the top level repo dir

Agreed, that's a good pragma. :+1:

> I think it should be merged into 0.3.7

The purpose of this PR explicitly targets developer docs, so targeting develop actually makes sense in this case. Won't block on that, since the changes are rather minor, though helpful.

fowlslegs:
>We should solve the issue throughout the docs, then. At the very least, throughout the developer docs.

Okay, I don't mind making it uniform across the dev docs and then rebasing.

>The purpose of this PR explicitly targets developer docs, so targeting develop actually makes sense in this case. Won't block on that, since the changes are rather minor, though helpful.

Hmm, as opposed to the source/admin/journalist docs, we probably want the dev docs to be the same no matter what version of the docs you are looking at. I don't know what the best way to resolve this is though. Thoughts?

fowlslegs:
So I just read through the rest of the developer docs and it looks like the rest of the commands, with the exception of one where you're instructed to clone the repo in the step right before, assume you're already in the top level dir.

It makes sense to merge this PR to me because we're going to start defaulting to using the latest stable release for our readthedocs. Not everyone (and in fact I assume few people) will know to select the develop branch docs in particular. Since develop and release/0.3.7 have the same version of the docs as of present moment I propose we do the following to handle documentation PRs until the branching strategy shift in release/4.0.0:

(i) Disable viewing/building all branches in readthedocs besides the current stable and release candidate.
(ii) Merge new documentation PRs into the rc if they are (a) general improvements to the documentation about the system as is (this could include dev doc improvements) or (b) related to fixes that will come out in that rc.

garrettr:
> It's mixed (i.e., no convention), but to me it makes sense to make the files/ paths relative to being inside the top level repo dir versus inside the parent dir of the top level repo dir.

@fowlslegs Agree this is a good convention to use going forward. You should add a note to `docs/development/documentation.rst`.

garrettr:
https://github.com/freedomofpress/securedrop/pull/1289#issuecomment-217592539 is giving me a headache. This change is tiny, there's no reason to backport it into the next point release. It should be merged into develop.

If we want, maybe we can re-do the "pull all docs from develop" commit to include some additional changes that shake out during this rc process. Let's evaluate that once the rc is closer to being done.

fowlslegs:
The fix is obviously tiny here. The point was more about getting some sort of consistent git workflow for documentation. I'm fine with pulling against develop and then maybe merging really useful things when we cut the rc.

garrettr:
> The point was more about getting some sort of consistent git workflow for documentation.

That is the primary motivation for backporting the docs from develop into 0.3.7. I'm pretty sure the docs (at least, the production installation-relevant docs) in develop still apply to 0.3.x, so this is a good time to bring some docs into a release.

Once the release is tagged, ReadTheDocs will automatically created a "stable" version of the docs that points to the latest tag. We can send a link to this version of the docs to people who want to install SecureDrop. There will still be a "latest" version of the docs pointing to the development branch. This is the version of the docs that should be used by developers. We'll update the README to clarify which version of the docs should be used, depending on your use case (installation or development).

Once we have this distinction between "stable" and "latest", we can freely modify the documentation in develop without worrying about breaking the instructions for installing the stable version of SecureDrop. The fact that this is impossible right now is a huge hindrance for development.

Hopefully this helps you understand why changes to the development documentation have no place in the point release - that's not where most developers should be looking for documentation, anyway.

In general, documentation should only change on the stable branch if it is forced to by an external dependency. Since the stable branch's code and docs are stored in the same repo, they should remain in sync for a long time. Changes may need to be made to the stable documentation in order to account for external dependency changes (e.g. Ansible updates, Tails updates, etc.). If that is the case, we would have to cut a new point release to include the documentation changes. These changes would be based off of the docs in the last release, and would then have to be merged into the develop branch.

PR #1291, opened 2016-05-07
Don't make tooltips seem like clickable buttons.

### LINKED ISSUE #1258: Journalist interface: cursor changes to "clickable" when hovering over message/reply icons
SD 0.3.6

In the journalist interface, while on the conversation page for a given source, if the cursor is hovered over the message (page) or reply (left arrow) icons, it changes to the hand that signals "this item is clickable." They do nothing when clicked.

kaizensoze:
The cursor: pointer makes the tooltips seem clickable, which is
misleading, especially for the Message, Uploaded Document icons on the
Source page.

resolves #1258

### DIFF
--- a/securedrop/static/css/journalist.css
+++ b/securedrop/static/css/journalist.css
@@ -598,11 +598,6 @@ button.button-star:focus {
     outline:0;
 }
 
-/* Tooltips */
-time:not([title='']):hover, i:not([title='']):hover {
-  cursor: pointer;
-}
-
 div#logout {
   float:right;
   margin: 1em 1em 0 1em;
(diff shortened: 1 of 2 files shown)


garrettr:
@kaizensoze This PR also removes some trailing whitespace. Was that intentional, or a byproduct of your editor's configuration?

kaizensoze:
A byproduct of atom. I can remove the whitespace changes from the PR if desired.

garrettr:
@kaizensoze Nope, they're fine, just checking :)

garrettr:
We're working on a proper CONTRIBUTING.md to make requirements for contributors clear. Until then, it'd be a bit unreasonable of us to be so picky with PR's 😁 

Testing this now.

garrettr:
@kaizensoze Since the only change in that CSS block is to change `cursor` away from the default value, I think a better solution would just to be to delete that block entirely.

kaizensoze:
Ok, removed the tooltip css blocks.

garrettr:
I love the smell of deleted code in the morning 🔥

PR #1292, opened 2016-05-09
Adds version check to tails_files

ageis:
The systemd-specific changes in d413ac6 will only work on Tails 2.x which is based on Debian jessie. This introduces a Tails version check to prevent the user from proceeding further if they are running Tails 1.x.

### DIFF
--- a/tails_files/install.sh
+++ b/tails_files/install.sh
@@ -21,6 +21,13 @@ DESKTOP=$HOMEDIR/Desktop
 ANSIBLE=$PERSISTENT/securedrop/install_files/ansible-base
 SSH_ALIASES=false
 
+# check for Tails 2.x
+source /etc/os-release
+if [[ $TAILS_VERSION_ID =~ ^1\..* ]]; then
+  echo "This script must be used on Tails version 2.x or greater." 1>&2
+  exit 1
+fi
+
 # check for persistence
 if [ ! -d "$TAILSCFG" ]; then
   echo "This script must be run on Tails with a persistent volume." 1>&2
--- a/docs/configure_admin_workstation_post_install.rst
+++ b/docs/configure_admin_workstation_post_install.rst
@@ -30,6 +30,8 @@ configure Tor to access a set of ATHS. In order to persist these changes across
 reboots, the Tails instance must have persistence enabled (specifically, the 
 "dotfiles persistence").
 
+We also require that you are using Tails 2.x or greater.
+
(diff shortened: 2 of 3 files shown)


garrettr:
Good call, @ageis. Would you mind skimming the docs to make sure they're clear about using Tails 2.x?

tails_files/install.sh
```diff
@@ -21,6 +21,13 @@ DESKTOP=$HOMEDIR/Desktop
 ANSIBLE=$PERSISTENT/securedrop/install_files/ansible-base
 SSH_ALIASES=false
 
+# check for Tails 2.x
+source /etc/os-release
```
conorsch: Why `source` and not `grep`? Altering the environment that `install.sh` runs in should be done carefully. Since this is a read-only operation, `grep` seems the safer option.

ageis: Well, @conorsch, os-release is a simple file that contains two environment vars: TAILS_PRODUCT_NAME and TAILS_VERSION_ID. It's almost begging to be sourced instead of grepped. We can grep through if ya'll consider that safer. It's a file owned by root in the base system however, so I wasn't considering it as a source of malicious code.
The structure is like:
```
TAILS_PRODUCT_NAME="Tails"
TAILS_VERSION_ID="2.3"
````

conorsch: Good to know it's a small file, and therefore the environment changes are minimal. (That's true today, though may not be true into the future!) Expected the answer to be that grepping with `set -e` can be complicated, since grep by default exits non-zero if there wasn't a match. Either way, an explanation of the decision should absolutely be added in a comment.

garrettr: > Expected the answer to be that grepping with set -e can be complicated, since grep by default exits non-zero if there wasn't a match.

@conorsch Remember that `set -e` does **not** always mean the script will exit if a command returns non-zero. For example, it will not exit if a command returns non-zero in a conditional. Run `info bash` and search for `errexit` for more info.

garrettr: In this case, sourcing `/etc/os-release` seems fine to me. It appears to have been designed with that purpose in mind: https://www.freedesktop.org/software/systemd/man/os-release.html says

> It is possible to source the configuration from shell scripts

ageis:
Added docs update to mention version requirement.

garrettr:
Good enough!

conorsch:
:shipit:

PR #1293, opened 2016-05-10
Create backup_mon role, handle restore errors by reverting files

fowlslegs:
This commit builds on the 660f78d3a4e3e9e9a9991edad60ace5595340db8 and addresses the issues raised in https://github.com/freedomofpress/securedrop/pull/1273.

## Maintaining SSH connections during Tor reloads

In `/lib/systemd/system/tor@.service` we see the line:

`ExecReload=/bin/kill -HUP ${MAINPID}`

The default action on receiving a SIGHUP is to terminate the process, but
processes may trap the signal and handle it however they like. Daemons commonly
will reload their configuration files. In Tor’s `src/or/main.c` we see the
function, `do_hup` that handles SIGHUPs. In it the only subroutine that deals
with open circuits is:

```c
  /* Rotate away from the old dirty circuits. This has to be done
   * after we've read the new options, but before we start using
   * circuits for directory fetches. */
  circuit_mark_all_dirty_circs_as_unusable();
```

Note this does not close currently open circuits, it simply marks them such that
no new connections may be made over them:

```c
/** Mark <b>circ</b> so that no more connections can be attached to it. */
void
mark_circuit_unusable_for_new_conns(origin_circuit_t *circ)
{
  const or_options_t *options = get_options();
  tor_assert(circ);

  /* XXXX025 This is a kludge; we're only keeping it around in case there's
   * something that doesn't check unusable_for_new_conns, and to avoid
   * deeper refactoring of our expiration logic. */
  if (! circ->base_.timestamp_dirty)
    circ->base_.timestamp_dirty = approx_time();
  if (options->MaxCircuitDirtiness >= circ->base_.timestamp_dirty)
    circ->base_.timestamp_dirty = 1; /* prevent underflow */
  else
    circ->base_.timestamp_dirty -= options->MaxCircuitDirtiness;

  circ->unusable_for_new_conns = 1;
}
```

I think this is a conclusive explanation of why the SSH connection was
maintained for you @garrettr after the reload, and sufficient to conclude that
this is the expected behavior we should be able to rely on.

## Keeping ATHSs for SSH accessible

If an error happens during the restore process, we would like our admins to
retain the ability to remotely access the SD instance. This is implemented as
follows. Now, if a restore file is defined, the backup which is always done
before a restore, is not deleted from the remote server until the restore
process is over. This way, in case of an error, we can recover from this backup
file, reverting the files overwritten by the faulty restore_file. An error
message is also printed informing the admin that the restore was unsuccessful
and that we've reverted the files back for them.

## Keeping the servers in sync with the admin workstation

If the restore process completed successfully, we re-use the code which
originally syncs the onion address and auth cookies to the
`./install_files/ansible-base/*ths` files during the install process (as well as
the tasks that prepend `HidServAuth` to the appropriate files).

## Sets group_var backup_script_version

I think that setting a group_var makes more sense than sourcing `version.py`
because this means not needing to bump the version of the backup script w/ every
version bump.

## Replaces restore script with ansible tasks

Since the functionality of the restore script could be reproduced with pure
Ansible, it made sense to get rid of it.

## Restore and backup for the Monitor Server as well

It makes sense, especially in order to better support requests like multiple
admin support (see https://github.com/freedomofpress/securedrop/issues/1175), to
support backup and restore of Tor services and configuration for the mon server
as well.

## Adds support for testing backup and restore from staging VMs as well

Because sometimes you don't want to test over Tor until you've tested locally
many times over. Soon you will be able to use the `SECUREDROP_SSH_OVER_TOR`
environmental variable as well to easily switch between accessing your staging
machines locally and via Tor, making this all the more handy.


fowlslegs:
I screwed up the commit message, thinking that I could use markdown headers, but really just commenting out those lines. Also, I have a half-finished title. I'll fix this later, just let me know what you think of the actual contents and please test it out as well (works here). I'll change Here is the commit message as intended:

**Create backup_mon role, implement error-handling for backup roles**

This commit builds on the 660f78d3a4e3e9e9a9991edad60ace5595340db8 and addresses
the issues raised in https://github.com/freedomofpress/securedrop/pull/1273.

## Maintaining SSH connections during Tor reloads

In `/lib/systemd/system/tor@.service` we see the line:

`ExecReload=/bin/kill -HUP ${MAINPID}`

The default action on receiving a SIGHUP is to terminate the process, but
processes may trap the signal and handle it however they like. Daemons commonly
will reload their configuration files. In Tor’s `src/or/main.c` we see the
function, `do_hup` that handles SIGHUPs. In it the only subroutine that deals
with open circuits is:

```c
  /* Rotate away from the old dirty circuits. This has to be done
   * after we've read the new options, but before we start using
   * circuits for directory fetches. */
  circuit_mark_all_dirty_circs_as_unusable();
```

Note this does not close currently open circuits, it simply marks them such that
no new connections may be made over them:

```c
/** Mark <b>circ</b> so that no more connections can be attached to it. */
void
mark_circuit_unusable_for_new_conns(origin_circuit_t *circ)
{
  const or_options_t *options = get_options();
  tor_assert(circ);

  /* XXXX025 This is a kludge; we're only keeping it around in case there's
   * something that doesn't check unusable_for_new_conns, and to avoid
   * deeper refactoring of our expiration logic. */
  if (! circ->base_.timestamp_dirty)
    circ->base_.timestamp_dirty = approx_time();
  if (options->MaxCircuitDirtiness >= circ->base_.timestamp_dirty)
    circ->base_.timestamp_dirty = 1; /* prevent underflow */
  else
    circ->base_.timestamp_dirty -= options->MaxCircuitDirtiness;

  circ->unusable_for_new_conns = 1;
}
```

I think this is a conclusive explanation of why the SSH connection was
maintained for you @garretr after the reload, and sufficient to conclude that
this is the expected behavior we should be able to rely on.

## Keeping ATHSs for SSH accessible

If an error happens during the restore process, we would like our admins to
retain the ability to remotely access the SD instance. This is implemented as
follows. Now, if a restore file is defined, the backup which is always done
before a restore, is not deleted from the remote server until the restore
process is over. This way, in case of an error, we can recover from this backup
file, reverting the files overwritten by the faulty restore_file. An error
message is also printed informing the admin that the restore was unsuccessful
and that we've reverted the files back for them.

# Keeping the servers in sync with the admin workstation

If the restore process completed successfully, we re-use the code which
originally syncs the onion address and auth cookies to the
`./install_files/ansible-base/*ths` files during the install process (as well as
the tasks that prepend `HidServAuth` to the appropriate files).

## Sets group_var backup_script_version

I think that setting a group_var makes more sense than sourcing `version.py`
because this means not needing to bump the version of the backup script w/ every
version bump.

## Replaces restore script with ansible tasks

Since the functionality of the restore script could be reproduced with pure
Ansible, it made sense to get rid of it.

## Restore and backup for the Monitor Server as well

It makes sense, especially in order to better support requests like multiple
admin support (see https://github.com/freedomofpress/securedrop/issues/1175), to
support backup and restore of Tor services and configuration for the mon server
as well.

## Adds support for testing backup and restore from staging VMs as well

Because sometimes you don't want to test over Tor until you've tested locally
many times over. Soon you will be able to use the `SECUREDROP_SSH_OVER_TOR`
environmental variable as well to easily switch between accessing your staging
machines locally and via Tor, making this all the more handy.

garrettr:
Looks you accidentally committed some `.swn` and `.swo` files. You should remove them from this PR (and considering adding them to your `.gitignore`).

install_files/ansible-base/roles/backup_app/tasks/backup.yml
```diff
@@ -1,18 +1,15 @@
 ---
-  # TODO the version is hardcoded into the collect.py script name and ansible
-  # task. This should be changed in the ansible task to use a group_var with the
-  # version or to read it from the securedrop_code version.py
-- name: copy the versions collect.py script to the target server /tmp directory
+- name: copy the version\'s collect.py script to the app server /tmp directory
   copy:
-    src: 0.3_collect.py
+    src: "{{ backup_script_version }}_collect_app.py"
```
garrettr: @conorsch Can you have subdirectories in an Ansible role's `files/` directory? I think it's a little neater to use subdirectories instead of interpolated filenames in a case like this, e.g.

```
roles/backup/files/0.3/collect_app.py
roles/backup/files/0.3/restore_app.py
...
```

conorsch: Absolutely. Other options include using `script` (not sure if that supports `chdir`...) or converting the backup script to a fully-fledged Ansible module. Since we expect it to always return "changed", i.e. always perform the action of collecting, zipping, and fetching, the module approach seems overwrought.

install_files/ansible-base/roles/backup_mon/files/0.3_collect_mon.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/python2.7
+"""
+This script is copied to the Mon server and run by the Ansible playbook. When
+run (as root), it collects all of the necessary information to backup the 0.3
+system and stores it in /tmp/sd-backup-mon-0.3-TIME_STAMP.tar.gz.
```
garrettr: Nit: this doesn't match the actual `backup_filename`.

install_files/ansible-base/roles/backup_app/tasks/backup.yml
fowlslegs: Is the idea of having versioning for this script to address potential issues of restoring from a older to newer version of SD?

garrettr: @fowlslegs You added the versioning, what do you think it's for?

fowlslegs: Adding it at the 'request' of the TODO comment you put in your original commit https://github.com/freedomofpress/securedrop/blob/660f78d3a4e3e9e9a9991edad60ace5595340db8/install_files/ansible-base/roles/backup/tasks/backup.yml#L2

install_files/ansible-base/roles/backup_mon/files/0.3_collect_mon.py
fowlslegs: Fixed.

fowlslegs:
Okay, nit fixed. Swapfiles removed. Original commit message fixed. This https://github.com/freedomofpress/securedrop/pull/1293#discussion_r62739181 conversation remains to be resolved. Also, this should be tested more/ on prod.

install_files/ansible-base/roles/backup_app/tasks/backup.yml
garrettr: > TODO comment you put in your original commit

I didn't put that there, James did, a long time ago (in c377f25974e986ac968842a8bf89224b9f865c29). I simply preserved it when refactored the backup task files in 660f78d3a4e3e9e9a9991edad60ace5595340db8.

Protip on finding which commit introduced a specific line: [Stack Overflow](http://stackoverflow.com/a/5816177).

I'm not certain of the specific motivation behind the original comment. In general, if we want to support backup and restore across different versions of SecureDrop, we should file an issue to explore that in more depth because it is a complex issue. For now, backup and restore is implicitly assumed to happen across the same version of SecureDrop. We call backup and restore across versions of SecureDrop "migration", and it's handled totally differently (see `migration_scripts`).

I understand that you saw the TODO and thought it would be smart to implement it while you were doing this other work. However, the way that you've refactored things to use `backup_script_version` isn't helpful in the short or long term, as far as I can tell. I think you should revert it.

garrettr:
> we re-use the code which originally syncs the onion address and auth cookies to the ./install_files/ansible-base/*ths files during the install process

You appear to have reused the code by copy-pasting it. Generally, if you ever find yourself copy-pasting large chunks of code, you should stop and ask yourself, "what am I doing wrong?". Might there be a more elegant solution?

fowlslegs:
If it was purely a copy-paste operation I would have just re-used the existing files. The addition of the `with` statements makes these roles more idempotent; there is no need to re-`fetch` the old values if the restore went wrong and we've reverted to the same files we were using before running the restore. Normally fetch is idempotent, but since we're using using `local_action: lineinfile` to format the `*ths` files the md5sums will not be the same and the block itself will effectively not be.

fowlslegs:
Should have just thought to do that ^ sooner. Still new at Ansible.

garrettr:
> Reuse display_onions task files in restore while preserving idempotence

@fowlslegs Ah, I didn't know you could do that either! Awesome 😄

garrettr:
Reviewing this in-person now.

> I think this is a conclusive explanation of why the SSH connection was maintained for you @garretr after the reload, and sufficient to conclude that this is the expected behavior we should be able to rely on.

We are not 100% sure, but think this might only work because we are using ControlPersist in our `ansible.cfg`. We may want to add a comment indicating that modifying Ansible's SSH configuration could have a negative impact on backup and restore and should be tested carefully if it is changed in the future.

install_files/ansible-base/roles/backup_app/tasks/backup.yml
```diff
@@ -1,18 +1,15 @@
...
     dest: /tmp/
     owner: root
     mode: 0770
```
garrettr: @conorsch says this needs to be quoted to avoid madness.

garrettr:
Just did an exhaustive review of this PR in-person with @fowlslegs and @conorsch. We decided that while there are some good changes in here, we should keep it simple and merge the minimal set of necessary changes to achieve our goals for 0.3.7. As a result, I am closing this PR and adding the remaining necessary changes to the original upstream PR, https://github.com/freedomofpress/securedrop/pull/1273.

Issue #1146, opened 2015-10-15
Run a current version of Ansible in the Admin Workstation


conorsch:
At time of writing, Tails 1.6 ships with Ansible 1.7.2 available in the Debian stable repository. Feature-wise, 1.7.2 is recent enough for our playbooks—the apache2_module was added in 1.6, for example—but it lacks several important bugfixes.

One example is [Get MemoryError when fetching large file](https://github.com/ansible/ansible/issues/11702), which was patched as of 1.7.3. I've seen this bug cause failures with the automatic backup role that runs at the end of the playbook. The failure aborts the backup process, and no encrypted tarball is written to the Admin Workstation. In order to force the backup, a manual rsync command is required, which sidesteps the MemoryError exception.

The problem is how to install a recent version of Ansible safely under Debian wheezy (which Tails 1.6 is based on). The [Ansible install docs](http://docs.ansible.com/ansible/intro_installation.html) include a long list of installation methods, but none includes GPG verification, which the current apt-get install process provides.

garrettr:
Potential solution worth investigating: as mentioned in https://github.com/freedomofpress/securedrop/issues/1035#issuecomment-140172267, Tails includes the Debian testing repos in their apt sources list by default. The current version of Ansible in Debian testing is [1.9.4](https://packages.debian.org/stretch/ansible).

conorsch:
Yes, `sudo apt-get install -y -s testing ansible` works! Haven't tested apt persistence in Tails with more recent version, will try that next.

garrettr:
@conorsch I bet the apt persistence would work just fine for packages installed from the testing repos. Have you had a chance to test it yet?

garrettr:
You should also test the workflow for migrating Admin Workstations that were using ansible from the stable repos, since that package will be cached locally due to the apt persistence. Can they just run `apt-get install -t testing ansible`, or do they need to explicitly remove the stable ansible package first?

conorsch:
@garrettr Haven't tested sufficiently yet. Persistence seems to work fine, but as you mention, it may be the stable version getting installed—haven't confirmed that it's the 1.9.x release that's getting reinstalled.

Also, installing ansible from the testing repo results in several interactive dpkg prompts. Sticking to the default suggestions for the confirmation prompts works fine, but will those prompts cause problems with apt persistence? Hopefully the Tails apt persistence correctly uses a Debian noninteractive environment, but will need to confirm to be sure.

[...]

PR #1294, opened 2016-05-10
Replace `sudo pip` with `virtualenv` or `mkvirtualenv` in dev docs

fowlslegs:
Using virtual environments makes it easier to manage multiple projects (or even
just branches of SecureDrop) that may require different versions of ansible.
Also, since pip does not check signatures on packages, it is better they are not
built as root.

### DIFF
--- a/docs/development/getting_started.rst
+++ b/docs/development/getting_started.rst
@@ -17,7 +17,7 @@ your development workstation.
 Ubuntu/Debian
 ~~~~~~~~~~~~~
 
-.. note:: Tested on: Ubuntu 14.04
+.. note:: Tested on: Ubuntu 14.04 and Debian Stretch
 
 .. code:: sh
 
@@ -46,6 +46,21 @@ which is signficantly out of date and known not to work with SecureDrop (`Github
 .. _`Vagrant Downloads page`: https://www.vagrantup.com/downloads.html
 .. _`GitHub #932`: https://github.com/freedomofpress/securedrop/pull/932
 
+.. warning:: We do not recommend installing vagrant-cachier. It destroys apt’s
+            state unless the VMs are always shut down/rebooted with Vagrant,
+            which conflicts with the tasks in the Ansible playbooks. The
+            instructions in Vagrantfile that would enable vagrant-cachier are
+            currently commented out.
+
+.. todo:: This warning is here because a common refrain during hackathons for
+          SecureDrop a while back was "setting up VMs is too slow, you should
+          use vagrant-cachier". We tried it and it had some nasty interactions
+          with Ansible, so we dropped it, and added this note to prevent other
+          people from making the same suggestion. Eventually, we should: (i)
+          Build our own base boxes to dramatically cut down on provisioning
+          times (ii) Remove this note as well as the commented vagrant-cachier
+          lines from the Vagrantfile
+          
 Either way, once you've installed Vagrant you should run:
 
 .. code:: sh
@@ -53,25 +68,37 @@ Either way, once you've installed Vagrant you should run:
     sudo dpkg-reconfigure virtualbox-dkms
 
 Finally, install Ansible so it can be used with Vagrant to automatically
-provision VMs. We recommend installing Ansible from PyPi with ``pip`` to ensure
-you have the latest stable version.
+provision VMs. We recommend installing Ansible from PyPi with ``pip``, which
+allows you to specifically install the version of Ansible available in Tails,
+currently 1.7.2 (5/6/16):
 
 .. code:: sh
 
     sudo apt-get install python-pip
-    sudo pip install -U ansible
 
-If you're using Ubuntu, you can install a sufficiently recent version of
-Ansible from backports (if you prefer): ``sudo apt-get install
-ansible/trusty-backports``
+Since the version of ansible required to provision SecureDrop VMs may flux out
+of sync with the version in your distro's repos, we recommend installing the
+version 1.8.4 of ansible to a virtual environment using `virtualenvwrapper` or
+`virtualenv`.
 
-.. note:: Tested: Ansible 1.9.4
+Using virtualenvwrapper:
 
-.. warning:: We do not recommend installing vagrant-cachier. It destroys apt's
-	     state unless the VMs are always shut down/rebooted with Vagrant,
-	     which conflicts with the tasks in the Ansible playbooks. The
-	     instructions in Vagrantfile that would enable vagrant-cachier are
-	     currently commented out.
+.. code:: sh
+
+    sudo apt-get install virtualenvwrapper
+    mkvirtualenv -p python2.7 securedrop
+    pip install ansible==1.8.4
+
+Using virtualenv (we recommend you `cd` into the  base directory of the repo
+before running these commands):
+
+.. code:: sh
+
+    sudo apt-get install virtualenv
+    virtualenv -p python2.7 .
+    . bin/activate
+    pip install ansible==1.8.4
+
 
 Mac OS X
 ~~~~~~~~
--- a/docs/development/getting_started.rst
+++ b/docs/development/getting_started.rst
@@ -68,18 +68,18 @@ Either way, once you've installed Vagrant you should run:
     sudo dpkg-reconfigure virtualbox-dkms
 
 Finally, install Ansible so it can be used with Vagrant to automatically
-provision VMs. We recommend installing Ansible from PyPi with ``pip``, which
-allows you to specifically install the version of Ansible available in Tails,
-currently 1.7.2 (5/6/16):
-
+provision VMs. We recommend installing Ansible from PyPi with ``pip``:
 .. code:: sh
 
     sudo apt-get install python-pip
 
-Since the version of ansible required to provision SecureDrop VMs may flux out
-of sync with the version in your distro's repos, we recommend installing the
-version 1.8.4 of ansible to a virtual environment using `virtualenvwrapper` or
-`virtualenv`.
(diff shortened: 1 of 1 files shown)


docs/development/getting_started.rst
```diff
@@ -17,7 +17,7 @@ your development workstation.
 Ubuntu/Debian
 ~~~~~~~~~~~~~
 
-.. note:: Tested on: Ubuntu 14.04
+.. note:: Tested on: Ubuntu 14.04 and Debian sid (May 6th)
```
garrettr: "Debian Sid (May 6th)" is that the best you can do? Seems like a pretty fuzzy/unhelpful reference point for "the version that these docs last worked on."

docs/development/getting_started.rst
```diff
@@ -46,32 +46,50 @@ which is signficantly out of date and known not to work with SecureDrop (`Github
 .. _`Vagrant Downloads page`: https://www.vagrantup.com/downloads.html
 .. _`GitHub #932`: https://github.com/freedomofpress/securedrop/pull/932
 
+.. warning:: We do not recommend installing vagrant-cachier. It destroys apt’s
+            state unless the VMs are always shut down/rebooted with Vagrant,
+            which conflicts with the tasks in the Ansible playbooks. The
+            instructions in Vagrantfile that would enable vagrant-cachier are
+            currently commented out.
+
```
garrettr: This warning is here because a common refrain during hackathons for SecureDrop a while back was "setting up VMs is too slow, you should use vagrant-cachier". We tried it and it had some nasty interactions with Ansible, so we dropped it, and added this note to prevent other people from making the same suggestion. Eventually, we should:

1. Build our own base boxes to dramatically cut down on provisioning times
2. Remove this note as well as the commented `vagrant-cachier` lines from the Vagrantfile

docs/development/getting_started.rst
```diff
@@ -46,32 +46,50 @@ which is signficantly out of date and known not to work with SecureDrop (`Github
...
 
     sudo apt-get install python-pip
-    sudo pip install -U ansible
 
-If you're using Ubuntu, you can install a sufficiently recent version of
-Ansible from backports (if you prefer): ``sudo apt-get install
-ansible/trusty-backports``
+Since the version of ansible required to provision SecureDrop VMs may flux out
+of sync with the version in your distro's repos, we recommend installing the
+latest stable ansible to a virtual environment using `virtualenvwrapper` or
```
garrettr: 1.7.2 is not the "latest stable Ansible" (didn't I already say this?)

docs/development/getting_started.rst
```diff
@@ -46,32 +46,50 @@ which is signficantly out of date and known not to work with SecureDrop (`Github
...
+Using virtualenvwrapper:
+
+.. code:: sh
+
+    sudo apt-get install virtualenvwrapper
+    mkvirtualenv -p python2.7 securedrop
+    pip install ansible==1.7.2
+
+Using virtualenv (must be in SecureDrop base directory before running these
+commands):
```
garrettr: Pretty sure you don't have to be; it's just the most common way and arguably the best way to do it (put the virtualenv in the project's root dir). But it could be anywhere on your system and I'm pretty sure it would work.

garrettr:
Overall, these changes are great. I'm actually surprised we left the `sudo pip install` instructions in for so long, that's definitely not best practice. I think for a lot of Python devs, they see that and automatically translate it to their environment management system of choice (virtualenv, virtualenvwrapper, anaconda, ...). Still, it is good to give explicit recommendations of best practices, which I think this PR does well.

conorsch:
>  I think for a lot of Python devs, they see that and automatically translate it to their environment management system of choice (virtualenv, virtualenvwrapper, anaconda, ...).

Guilty as charged. The change is a definite improvement, and obviously more "explicit".

docs/development/getting_started.rst
fowlslegs: Please be more polite. And I'll just remove the date or do you think Debian sid is too shifting a point of reference and just shouldn't be added?

garrettr: It's valuable to know that it worked on that system at a given point. Sorry if the message came off rude, it wasn't intended to be. I literally don't know if there is a more specific way to refer to versions of Debian Sid than the date. If there is, I think that would be preferable here.

docs/development/getting_started.rst
fowlslegs: Since we're hiding 'todos' now, I'll copy this note into a todo.

docs/development/getting_started.rst
conorsch: > do you think Debian sid is too shifting a point of reference and just shouldn't be added?

This resonates with me. Your inclination to add a timestamp to a rapidly changing distro is good. Maybe we can compromise with `Debian Sid (Q2 2016)` for a broader window. Even that strikes me as rather narrow, though—will we update it again in Q3 2016?

Another troubling point is that the docs claim our setup has been tested on Ubuntu 14.04 LTS. That _was_ true, but I believe I was the most active Ubuntu user, and I'm now primarily developing on Debian. I do have an Ubuntu 16.04 box that I can run through the environment setup on and confirm it works as expected.

docs/development/getting_started.rst
fowlslegs: Yeah, I rewrote that part too, but must have not committed it.

docs/development/getting_started.rst
fowlslegs: No, you don't. I just wrote that because like you said it's arguably best practice. I'll change it to "we recommend."

docs/development/getting_started.rst
garrettr: Right - the important thing here is the "Ubuntu 14.04" means the same thing today that it meant last year, which is not true *at all* for Debian Sid. We want to avoid our docs bit-rotting as much as we can.

Adding the date to "Debian Sid" obviously fixes it in time, but that doesn't seem very useful to me. e.g. if I'm running Debian Sid 6 months from now, try to use these docs, and it doesn't work, can I use "Debian Sid (May 6th)" to compare that system to my current system in order to try to figure out what changed that broke these instructions?

If adding the date is the best we can do, then that's fine. I don't know if there is a more precise/"useful for future reference" versioning scheme used by Debian. Using the date does not seem very helpful to me (according to that thought experiment I just described), but if I'm wrong then that's fine 😁

fowlslegs: In `/etc/os-release` I see `Debian GNU/Linux stretch/sid`. I think that this will change to `<v10 codename>/sid` after stretch is released. Maybe `stretch/sid` is too confusing though and it will be read as in both stretch and sid. Maybe we shouldn't include this addition at all, or perhaps we can just say `Debian stretch` as I believe that is what @conorsch is running and provides a more stable point of reference.

docs/development/getting_started.rst
fowlslegs: Also, 1.7.2 is failing for both @conorsch and I to setup staging VMs, even though he used it successfully in real-hw prod just days ago. I mean, I know they run through different playbooks, but the error is about `apt_key` not being present, when it's been in ansible far before 1.7.2. @conorsch was saying it might have to do with how they placed the libraries in the pip package (vs. the Debian package that was used for the real install). Anyway, I'm going to reinstate 1.8.4 for now.

docs/development/getting_started.rst
fowlslegs: Added.

docs/development/getting_started.rst
conorsch: > Anyway, I'm going to reinstate 1.8.4 for now.

This frustrates me deeply, but I agree with the determination: tell devs to use 1.8.4, because that's what we're using. We always test on hardware, in Tails, prior to a release, so we'll still catch any version-specific tangles during that process. It's good to keep in mind that we're intentionally conservative with module choice in order to retain support for older versions of Ansible—see #1146 for tracking more modern versions of Ansible for admins.

docs/development/getting_started.rst
garrettr: @fowlslegs I'll let you make a final call. Seems like there is no better option than "Debian Sid (May 10, 2016)" in which case, go for it.

fowlslegs: I say we go with `Debian Stretch` because it's a more stable point of reference and @conorsch is running it, so if it breaks we'll know. Just replaced it with that.

docs/development/getting_started.rst
```diff
@@ -46,32 +46,59 @@ which is signficantly out of date and known not to work with SecureDrop (`Github
...
     sudo apt-get install python-pip
-    sudo pip install -U ansible
 
-If you're using Ubuntu, you can install a sufficiently recent version of
-Ansible from backports (if you prefer): ``sudo apt-get install
-ansible/trusty-backports``
+Since the version of ansible required to provision SecureDrop VMs may flux out
+of sync with the version in your distro's repos, we recommend installing the
+version 1.8.4 of ansible to a virtual environment using `virtualenvwrapper` or
+`virtualenv`.
```
garrettr: You to seem to give conflicting advice about whether to use Ansible version 1.7.2 or 1.8.4...

fowlslegs: Missed the earlier one when I was correcting it.

fowlslegs: Fixed in https://github.com/freedomofpress/securedrop/pull/1294/commits/f7d10d394b9aa50f4e1ed2c6c5259b8c9f2c2c9a

fowlslegs:
So this commit is the result of using GH's new squash-merge @garrettr? Just checking.

garrettr:
@fowlslegs That's correct.

PR #1295, opened 2016-05-10
Change yml config file paths to allow for copy-paste from dev docs

fowlslegs:
>... it makes sense to make the files/ paths relative to being inside the top level repo dir versus inside the parent dir of the top level repo dir. Besides the fact this avoids the confusion of there being securedrop/securedrop dir (which becomes ./securedrop), it also makes sense to cd into the top level dir of a repo you're going to be working with, and from there you should be able to copy-paste the filenames/ commands from the dev docs.

-- https://github.com/freedomofpress/securedrop/pull/1289#issuecomment-217563245

### DIFF
--- a/docs/development/documentation.rst
+++ b/docs/development/documentation.rst
@@ -86,3 +86,12 @@ Style Guide
                italicized, which is confusing when used near
                references to the terminology.
 
+* Use absolute paths when referring to files outside the SecureDrop repository.
+  Exceptions made for when it's clear from the surrounding context what the
+  intended working directory is. For files inside the SecureDrop directory,
+  write them as `./some_dir/file`, where `.` is the top level directory of the
+  SecureDrop repo. Since by default the git repo will be cloned under the name
+  `securedrop` and it also contains a `securedrop` subdirectory this is intended
+  to avoid confusion.  Exceptions made for when it's clear from the context
+  we're outside of the SecureDrop repo, but would like to somehow interact with
+  it (e.g., we just cloned the repo and now we're going to `cd` into it).
--- a/docs/development/getting_started.rst
+++ b/docs/development/getting_started.rst
@@ -199,7 +199,7 @@ across the full stack.
 
 If you want to receive OSSEC alerts or change any other settings, you will need
 to fill out your local copy of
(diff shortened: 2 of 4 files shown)


garrettr:
A quick grep shows this is not the only location of `securedrop/` in the documentation. You might as well get them all at once if you're going to change the convention across the docs. Also, you should add a note (as I said on the last PR: https://github.com/freedomofpress/securedrop/pull/1289#issuecomment-217980487) that this is the preferred convention.

fowlslegs:
Okay, I only did the dev docs at first, but now all of them. There are some times `securedrop/` is appropriate as I explained in https://github.com/freedomofpress/securedrop/pull/1289#issuecomment-217563245 and also in the new documentation guideline I added. Rebased so that the commit has a more appopriate title.

docs/development/documentation.rst
```diff
@@ -86,3 +86,12 @@ Style Guide
...
 
+* Use absolute paths when referring to files outside the SecureDrop repository.
+  Exceptions made for when it's clear from the surrounding context what the
+  intended working directory is. For files inside the SecureDrop directory,
+  write them as `./some_dir/file`, where `.` is the top level directory of the
+  SecureDrop repo. Since by default the git repo will be cloned under the name
+  `securedrop` and it also contains a `securedrop` subdirectory this is intended
+  to avoid confusion.  Exceptions made for when it's clear from the context
+  we're outside of the SecureDrop repo, but would like to somehow interact with
+  it (e.g., we just cloned the repo and now we're going to `cd` into it).
```
garrettr: Well said :+1:

docs/development/getting_started.rst
```diff
@@ -199,7 +199,7 @@ across the full stack.
 
 If you want to receive OSSEC alerts or change any other settings, you will need
 to fill out your local copy of
-``securedrop/install_files/ansible_base/staging-specific.yml``.
+``./install_files/ansible_base/staging-specific.yml``.
```
ageis: It's arguable that including the single period for current working directory - `./install_files` vs. the absolute sufficiency of `install_files/` is complicating with two extra characters and shell conventions. Just pointing out there's another choice ;)

Issue #1284, opened 2016-05-04
Document upgrading Tails workstations to Tails 2.x


garrettr:
Tails 2.x has been out for some time now, and the Tails 1.x is no longer receiving updates. Therefore it is important for security that we encourage SecureDrop users to upgrade Tails for the components of the SecureDrop architecture that are based on Tails:

* Admin Workstation
* Journalist Workstation
* Secure Viewing Station (SVS)

There is [no auto-upgrade mechanism for Tails 1.x to Tails 2.x](https://tails.boum.org/news/version_2.0/index.en.html#index4h1), so we need to provide clear documentation for upgrading each of the above listed components.

We have made improvements to tails_files in #1206 and #1277 and according to @ageis it is working without issues on Tails 2.x.

@harlo has a private draft of upgrade documentation on [Hackpad](https://securedrop.hackpad.com/Tails-1.x-to-2.x-upgrade-s3jHsmIinS3).

We have started the release integration branch in release/0.3.7, and we have backported the ReadTheDocs documentation from develop, so I think that would be the best place for this documentation to live. Once we tag the release, it will automatically be promoted to the "stable" version of the docs on ReadTheDocs.

garrettr:
This was merged into release/0.3.7 in #1303. Some follow-up issues were identified in release QA testing for 0.3.7, described in #1315, and resolved in #1317. Therefore, this issue is resolved. It will be merged into the 0.4 development branch (develop) in #1313.

PR #1296, opened 2016-05-10
Tails 1.x to 2.x upgrade docs

harlo:
documentation for upgrading Tails, migrating and backing-up data as per https://github.com/freedomofpress/securedrop/issues/1284


docs/upgrade_to_tails_2x.rst
```diff
@@ -0,0 +1,138 @@
+Upgrade Tails from 1.x to 2.x
+=============================
+
+Newer versions of SecureDrop leverage improvements to the Tails OS since the introduction of Tails 2.0 (and subsequent versions). It is critical to upgrade all Tails USBs to the latest version before upgrading SecureDrop. For some admins, this might be complicated, since the upgrade from 1.x to 2.x must be done manually. Please follow this guide to updating the entire system, and special attention to only use the Secure Viewing Station computer so you benefit from its airgap while transfering sensitive data.
```
garrettr: I would also add a note saying that Tails 1.x is no longer receiving updates, so it is vulnerable to a growing list of security vulnerabilities. Everyone should upgrade to Tails 2.x, which will be getting automatic upgrades for the foreseeable future.

docs/upgrade_to_tails_2x.rst
```diff
@@ -0,0 +1,138 @@
...
+
+Insert your Transfer Device into a USB port and select it from the left column. Brand new devices sometimes have pre-configured partitions, which you will need to remove. Select any block of partitioned data, and click the minus (``-``) button to remove any unwanted partitions.
+
+|Unwanted Bloatware Partition|
+
+Click the plus (``+``) button to register a new partition to cover the entire available space. This should automatically be filled out for you in the wizard.
+
+Be sure to do a full wipe of the existing data during this step, especially if you choose to repurpose an older USB stick. Select "Overwrite existing data with zeroes (Slow)" from the "Erase" options. This does mean that the reformatting step will take some extra time, but it is the best way to insure that previous data from older transfers, or manufacturer bloatware, is removed from the Transfer Device.
+
+Give your new partition a complex, diceware-generated passphrase.
```
garrettr: Maybe we should just say "strong passphrase"? I'd rather use that throughout the docs, and then provide our own definition of "strong passphrase" in one canonical location (if you think that would be useful).

docs/upgrade_to_tails_2x.rst
```diff
@@ -0,0 +1,138 @@
...
+|Root Terminal|
+
+The Nautilus window should show both the Transfer Device and the TailsData partition as mounted.
+
+|Migrate Data 1|
+
+Copy the all data from the TailsData partition onto the Transfer Device **except**:
+
+-	``persistence.conf``: in older versions of Tails, this file might have slightly different directives in it that could temporarily brick a Tails 2.x USB.
+-	``claws-mail`` folder: Claws Mail is no longer included in Tails. The OS uses Icebird instead. Some users might not have this folder, so if you don't see it there, do not be alarmed.
```
garrettr: Are the tabs after the `-` necessary?

docs/upgrade_to_tails_2x.rst
```diff
@@ -0,0 +1,138 @@
...
+
+|Migrate Data 2|
+
+Insure that all critical data has been successfully copied.  Specifically, be sure the the ``gnupg``, ``bookmarks``, and ``Persistent`` folders are completely copied.  Any loss of data from these folders could prevent users from accessing submissions.
```
garrettr: s/Insure/Ensure

docs/upgrade_to_tails_2x.rst
```diff
@@ -0,0 +1,138 @@
...
+	You'll need these old values for the following step.
+
+Re-install Securedrop with ``cd ~/Persistent/securedrop/tails_files && sudo ./install.sh``. Once completed, test your access to the Document Interface, and, if you're a Secure Drop admin, test your ssh connection to the application and monitor servers.
+
+If you did this correctly, move on to Step 7.
+
+7. Reformat the Transfer Device
+-------------------------------
+
+Repeat step 1 to restore a USB stick to a pristine state. While it probably isn't necessary to physically destroy a Transfer Device (because LUKS-encrypted data is very hard to forensically recover), you could smash the device with a hammer until the chips containing its flash memory are broken up into pieces before disposal.
```
garrettr: I would definitely recommend a couple of passes of `/dev/urandom` over the drive though. Is there a GUI tool in Tails that can do this? If not, this guide is aimed for admins so I think it would be ok to tell them how to do it on the command line.

garrettr:
Overall, looking really good. I just rendered the docs locally, doing another skim through them now. @conorsch Would you mind taking a look as well?

garrettr:
Also, you should add these new docs to the TOC in `index.rst`. It's up to you where/how you want to add it. Two options that occur to me:

1. Put it under "Topic Guides"
2. Rename the "Upgrade SecureDrop" to "Upgrade Guides", and put it there.

Of those two, I think prefer Option 1.

docs/upgrade_to_tails_2x.rst
fowlslegs: I feel like there is a GUI tool that will do that, but anyway, the way they recommend doing it on the cryptsetup wiki is a good deal faster than using `/dev/urandom`:

```
cryptsetup open --type plain <target-device> <container-name>
dd if=/dev/zero of=/dev/mapper/<container-name>
cryptsetup close <container-name>
```

garrettr: > the way they recommend doing it on the cryptsetup wiki is a good deal faster than using /dev/urandom

Next time, please provide a link when you quote an external resource. For others, wiping dm-crypt volumes is discussed in section 2.19 of the cryptsetup FAQ, [here](https://gitlab.com/cryptsetup/cryptsetup/wikis/FrequentlyAskedQuestions).

garrettr: Also, I think your comment incorrectly copies the procedure from the wiki. The whole point of their procedure is to create an ephemeral volume with a random key (`-d /dev/urandom`) on top of the block device to be wiped. That way, `dd if=/dev/zero` *ultimately* writes a pseudorandom stream of bytes to the block device (since the zeros are being encrypted under a random, ephemeral key by dm-crypt).

Your steps don't provide a random key to `cryptsetup open`. Running that command prompts for the existing LUKS passphrase. As a result, the commands as you describe them *do not overwrite the LUKS header*. This is a problem because that is where the anti-forensic countermeasures are implemented (see section 5.2 of the [LUKS paper](http://tomb.dyne.org/New_methods_in_HD_encryption.pdf)).

In general, you should be ok as long as you can reliably overwrite the LUKS header. This saves you a lot of time, since the LUKS header is much smaller than the entire block device. This technique is documented in detail on the [Arch wiki](https://wiki.archlinux.org/index.php/Dm-crypt/Drive_preparation#Wipe_LUKS_header).

Unfortunately, ***reliably overwriting*** the LUKS header is not guaranteed to work on an SSD, because you can't reliably overwrite *anything* on an SSD. This is noted on the Arch Wiki, and they link to section 5.19 of the [cryptsetup FAQ](https://gitlab.com/cryptsetup/cryptsetup/wikis/FrequentlyAskedQuestions#5-security-aspects) for more info on securely wiping SSDs. Another good resource for background is [Wei '11](https://www.usenix.org/legacy/events/fast11/tech/full_papers/Wei.pdf).

This distinction is important because we're talking about wiping the transfer device, which is almost always a USB drive and therefore almost always using flash memory. Unfortunately, there are no easy answers here that I know of. I would probably stick with the technique from the cryptsetup wiki (also documented with some extra helpful detail on the [Arch wiki](https://wiki.archlinux.org/index.php/Dm-crypt/Drive_preparation#dm-crypt_wipe_on_an_empty_disk_or_partition)). According to Wei's paper,

>  In most cases, overwriting the entire disk twice was sufficient to sanitize the disk, regardless of the previous state of the drive.

That, combined with the anti-forensic properties of LUKS, seems "good enough" IMHO. Of course, the only way to be sure is to physically destroy the drive, which you've already mentioned as an option.

docs/upgrade_to_tails_2x.rst
garrettr: I would rewrite this first paragraph to something like:

If you've been using SecureDrop for a while, you're probably using a version of Tails from the 1.x release series on your Tails USBs for the Admin Workstation, Journalist Workstation, and Secure Viewing Station. Tails has switched over to the 2.x release series, and is no longer supporting the 1.x release series, which means Tails 1.x systems are no longer receiving security updates. As a result, your Tails system will have have a gradually increasing number of security vulnerabilities, which could be used to compromise the security of your SecureDrop instance. As a result, we ***strongly recommend*** upgrading to the latest version of Tails (2.3 at the time of this writing).

In addition to the security concerns, newer versions of SecureDrop leverage improvements to the Tails OS since the introduction of Tails 2.0 (and subsequent versions). It is critical to upgrade all Tails USBs to the latest version before upgrading SecureDrop.

For some admins, this might be complicated, since the upgrade from 1.x to 2.x must be done manually. Please follow this guide to updating the entire system, and special attention to only use the Secure Viewing Station computer so you benefit from its airgap while transfering sensitive data.

conorsch:
Walking through the upgrade guide on hardware, will report back with any edits or clarifications.

conorsch:
Great docs, @harlo! The screenshots are magnificent. To test this, I used a test-only hardware instance of SD with Tails 1 USB sticks for all roles. I walked through the steps as you describe and was able to confirm a fully functioning system by logging in over SSH, submitting a document to the Source Interface, downloading it via the Journalist Workstation, and finally decrypting it on the SVS.

I've opened a PR against your feature branch with a few copy edits. The commit messages are verbose (:tada:), and I tried to keep the changes focused to ease review. If you agree with the changes there, please merge on your fork and we'll see the commits here. Alternatively, I'm happy to commit directly to your fork. Additional changes I'll discuss in a separate comment.

conorsch:
A few suggestions to improve the clarity of this guide:

### Rename "Transfer Device" to "Backup Device"

The Tails stick used for storing the backups is a failsafe mechanism, in case something goes wrong during the upgrades. In that sense, it's not a "transfer" device at all. More to the point, reusing the "Transfer Device" terminology may be confusing to Admins, since we already use that term for a specific concept in SD architecture, and we don't don't mean the same thing here.

### Break out Troubleshooting steps to separate section
The current docs flow looks like this:

1. Prepare the master Tails USB
2. Prepare the Transfer Device
3. Backup a Tails USB
4. Upgrade a Tails SB
5. Restore data from a Transfer Device *[see above: "Backup Device" may be a better name!]*
6. Reinstall SecureDrop *[nit: "Reinstall SecureDrop dotfiles" may cause fewer heart attacks]*
7. Reformat the Transfer Device

5 and 6 are designed to be skipped, so how about:

1. Prepare the master Tails USB
2. Prepare the Transfer Device
3. Backup a Tails USB
4. Upgrade a Tails SB
5. Reformat the Transfer Device

**Troubleshooting**

1. Restore data from a Transfer Device
2. Reinstall SecureDrop 

### How many passes?
The guide instructs Admins to duplicate a lot of labor, advising a full backup-upgrade-reboot-verify loop for each device. Given that we've encrypted the backup device and are airgapped, is it reasonable to instruct Admins to store the backups within subdirectories? The backup device could then be used as a cold storage recovery device in the event of a disaster.

### What about remote admins?
The guide assumes the Admin has physical access to the Journalist and SVS workstations. What are our recommendations if the Admin is not physically copresent? Should the Admin contact the Journalists and tell them to follow the guide?

------------------------------------------
I'm happy to handle updates implementing the above, but wanted to open for the floor to discussion before doing so. Holler back and we'll hash it out.

conorsch:
@harlo Thanks for merge! @garrettr, thoughts on further changes explained above? Remember that we're going to have to instruct folks to run the `tails_files/install.sh` script on the Journalist Workstation as well, in order to clear up the network hooks so the desktop icons and ATHS values are configured before the apt hook.

garrettr:
(clarified with @conorsch that "changes explained above" refers to https://github.com/freedomofpress/securedrop/pull/1296#issuecomment-218911351, not the commit after it)

> Rename "Transfer Device" to "Backup Device"

:+1:

Especially since this guide recommends reformatting and wiping the "transfer device"/"backup device" - if they do this with their existing transfer device, they'll have to recreate it, which could be annoying or confusing. Better use a unique name.

For sourcing hardware, I think the note in section 2 about using new vs. repurposing USB's is sufficient.

> Break out Troubleshooting steps to separate section

:+1:

> How many passes?

I think that's reasonable, and might save time. Not sure how to best explain this in the guide.

> What about remote admins?

My inclination is to say "let them figure it out." The docs are clear that **all** Tails USBs must be upgraded, and describe clearly how that should be done. It's up to the individual instance operators to figure out the logistical details. We can't do everything for them.

garrettr:
Note: the current docs do not list an airgapped computer in the "You will need:" section, but reference one throughout. This should be fixed.

Please clearly (but briefly) explain _why_ we are recommending using an airgap for these processes. Our docs have an unfortunate tendency to skimp on the "why," and I think this often confuses people, or leads them to use workarounds that compromise security because they don't understand the reasoning for doing things a certain way.

conorsch:
Thanks for merge, @harlo! Looks like I screwed up the formatting for section headers:

rtd-navbar-sections-wrong

Those **Finishing up** and **Troubleshooting** sections should not be broken out. I'll patch that and update this thread to request a final review.

conorsch:
New PR submitted targeting this feature branch. Strictly cleanup, no new content. Once that's merged, @harlo, can you tag @garrettr for a final review on these docs? I'm happy with them as they stand post-merged.

garrettr:
Superseded by #1303.

PR #1180, opened 2015-11-07
Bundle and minimize CSS & JS assets

patrickod:
First step in addressing freedomofpress/securedrop#1152

Uses flask-assets to group the CSS assets, run them through
cssmin, and link to them in the index.html and base.html templates
for both journalists and sources

Adds both flask-assets and cssmin as securedrop depedencies

PR #1297, opened 2016-05-11
Fix javascript

kaizensoze:
Specifying type="javascript" for the script tags seems to break
javascript.

### DIFF
--- a/securedrop/journalist_templates/base.html
+++ b/securedrop/journalist_templates/base.html
@@ -13,7 +13,7 @@
 
     {% assets filters="jsmin", output="gen/journalist.js",
       "js/libs/jquery-2.1.4.min.js", "js/journalist.js" %}
-      <script type="javascript" src="{{ ASSET_URL }}"></script>
+      <script src="{{ ASSET_URL }}"></script>
     {% endassets %}
 
     {% block extrahead %}{% endblock %}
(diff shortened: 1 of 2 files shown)


garrettr:
I am confused. First of all, if

> Specifying type="javascript" for the script tags seems to break javascript.

Then doesn't your PR, which adds `type="javascript"` to script tags, well... break javascript?

Also, `javascript` is the default `type` on a `script` tag, so I do not understand how this change is meaningful.

garrettr:
And, apparently I can't read. Thanks to @conorsch for pointing that out.

conorsch:
Cheers! Still though, this PR is definitely lacking in context. Can you provide an explanation of the problem this PR is supposed to solve, @kaizensoze?

garrettr:
@conorsch I understand it. `javascript` is an invalid type for a script tag. Removing it and letting it to revert to the default type for script tag allows the script to work again. Looks like #1180 broke some stuff and we didn't realize it at the time.

I've manually tested this and confirmed that Javascript isn't working on develop, but this PR fixes it. Nice catch, @kaizensoze!

kaizensoze:
Sorry, I should've provided a much better explanation in the PR.

conorsch:
@kaizensoze No worries! It was rather self-explanatory. Thanks for the contribution!

PR #1298, opened 2016-05-11
Remove jQuery dependency

kaizensoze:
reference used: http://youmightnotneedjquery.com/

As fun as it was becoming more familiar with vanilla javascript, I probably should've asked about this one before starting it since, on the one hand there's no longer a reliance on jquery, but on the other hand, there's now technically more [securedrop-related] javascript code.


garrettr:
> As fun as it was becoming more familiar with vanilla javascript, I probably should've asked about this one before starting it since, on the one hand there's no longer a reliance on jquery, but on the other hand, there's now technically more [securedrop-related] javascript code.

In my opinion, it's definitely a positive choice to remove jQuery (per #1233) from the Source Interface. This is because the only reason there is any JS on the source interface is to encourage sources to disable it, and provide a convenient pop-up guide on how to do that in Tor Browser. Given this extremely focused use of JS, with no plans for expansion, jQuery is overkill here.

I think it could also be argued that it's _currently_ overkill on the Document Interface. The DI uses JS for much more than the SI does, but its usage is still very simple compared to a typical modern web app. The original reason why it was included is that it encouraged rapid prototyping on the DI. I still think it's somewhat useful for that, but:

* We plan to do a massive re-architecture of the whole journalist workflow in the near future, which will obsolete the current DI.
* SecureDrop is only intended to be used with the Tor Browser, so one of the key reasons for using jQuery (normalizing behavior across browsers) doesn't apply.

So I think this PR is a candidate for merge. It will need testing, however, to ensure that there are no regressions.

fowlslegs:
@freedomofpress/employees I think this is a valuable contribution, but don't feel comfortable merging it myself. Maybe somebody else who knows JS well can.

securedrop/static/js/journalist.js
```diff
@@ -2,93 +2,172 @@
...
+      });
+    });
+  }
 
-  var filter_codenames = function(value){
-    if(value == ""){
-      $('ul#cols li').show()
+  var filter_codenames = function(value) {
+    if (value == "") {
+      [].forEach.call(document.querySelectorAll('ul#cols li'), function(el) {
```
garrettr: Is there any reason you're using `[].forEach.call(...)` instead of:

```js
document.querySelectorAll('ul#cols li').forEach(function(el) { ... });
```
?

kaizensoze: I honestly don't remember but I'll take a look.

kaizensoze: I believe this is the reason: https://css-tricks.com/snippets/javascript/loop-queryselectorall-matches/.

kaizensoze:
Merge conflicts should be resolved.

securedrop/static/js/journalist.js
garrettr: @kaizensoze Ah, thanks for clarifying. I had tested this in the Chrome Javascript console, where it appears they have fixed this issue: querySelectorAll returns an Array instead of a NodeList, and therefore it has `forEach` available. I just tested it again in the Tor Browser Javascript console and can confirm that in Tor Browser querySelectorAll returns a NodeList, which does not have `forEach` - so we do need some kind of workaround.

That said, I'm not a fan of `[].forEach.call`. As the author of the CSS-Tricks article that you linked points out:

> Fair warning, Todd Motto explains why this method is a rather hacky, detailing over 10 problems with it.

and links to Todd Motto's [Ditch the [].forEach.call(NodeList) hack](https://toddmotto.com/ditch-the-array-foreach-call-nodelist-hack/).

I would prefer it if code that loops over the results of querySelectorAll either:

1. Used a for loop
2. Used our own `forEach` method (Motto's [recommendation](https://toddmotto.com/ditch-the-array-foreach-call-nodelist-hack/#recommendations)).

kaizensoze: Ok I'll look into replacing [].forEach.call with a custom forEach method later tonight.

kaizensoze:
I've made the changes to replace `[].forEach` with a custom `forEach`; however, I'm unable to test the changes since I'm getting an error when doing a fresh `vagrant up`:

```
TASK [development : Install apt package dependencies for Development environment.] ***
failed: [development] (item=['development_dependencies']) => {"failed": true, "item": ["development_dependencies"], "msg": "No package matching 'development_dependencies' is available"}
```

fowlslegs:
0.4 @garrettr?

fowlslegs:
I merged this in the `alpaca` test branch a few weeks back because it greatly reduces the page size. Worked great!

While this increases the performance of the page load a-la https://github.com/freedomofpress/securedrop/issues/1152, the real reason I merged it into that test branch is that the Alpaca website fingerprint defense is more effective the smaller a page is, and the fewer resources it includes.

I'm definitely in favor of rebasing/ cleaning up this branch and merging it into develop, maybe for `0.4.1`.

redshiftzero:
I agree, let's clean this up for 0.4.1.

redshiftzero:
Thanks for this contribution @kaizensoze. This PR looks good and is a candidate for merge and release in SecureDrop 0.4.2, but it will need to be rebased onto current `develop` prior to merge. Are you interested in doing that? Note we can now also remove the unminified js in this PR `jquery-2.1.1.js`. Let us know!

kaizensoze:
I went ahead and rebased upstream develop into this branch. The only changes I'm not too sure about are the ones in securedrop/static/js/journalist.js. Don't really have the bandwidth to go through and test myself.

dachary:
> Don't really have the bandwidth to go through and test myself.

Took a quick look at the diff and indeed ... this is going to be significant work to test properly :-)

kaizensoze:
I had already tested all the functionality on submitting the first version, but yeah, post-rebase, it'll take some time.

dachary:
yeah, unfortunately we don't have any JS tests at the moment... I see there still are conflicts in this PR, did you push your rebase ?

kaizensoze:
Nope, the uncertainty with journalist.js threw me off so I wasn't sure how to finish the rebase and figured I'd get some feedback on how to merge it before pushing the entire rebase.

heartsucker:
Hey @kaizensoze. We had a little chat about this at the most recent engineering meeting. Since none of us are JS experts, to ensure the functionality stays consistent, it was suggest that we implement Selenium tests first that match the current behavior and then apply your patch on top and ensure they still pass. Does this sound like something you could implement? Also, apologies for the inconvenience. Front end stuff has been a pain point for use for as long as I've been on the project. :/

kaizensoze:
Understandable. Unfortunately I don't have the bandwidth to write selenium tests.

heartsucker:
Tagging @dachary and @redshiftzero. I think we should do a manual test and then merge this as this will let help us solve the problem with Flask-Assets and Flask-Babel easier if we have only one JS file to deal with. Also, since we're about to do a bit of a refactor on the tests along with the refactor of the code for #2057 and #2294, adding this in might not even make sense now since @dachary has had trouble with the JS flag in Tor Browser in Selenium.

redshiftzero:
For now, I'm going to close until the necessary Selenium tests for merge are written. I've pushed the work here to `kaizensoze-remove-jquery` branch so that they are saved, and the branch name is posted in ticket #1233 for picking back up by a contributor so the changes here can be safely merged in. Thanks again @kaizensoze :smile:.

Issue #856, opened 2015-01-29
Upgrading securedrop-app-code should remove unused Python dependencies


garrettr:
The postinst script of `securedrop-app-code` upgrades the Python dependencies with the following command:

```
pip install --no-index --find-links=/var/securedrop/wheelhouse -r /var/www/securedrop/requirements/securedrop-requirements.txt
```

The only downside here is that it does not automatically remove unused dependencies. For example, when I upgraded the Python dependencies with `pip_update.sh` in 9868f3a7a3d39e63748ed8e4305bdf0a8d07918c, `argparse` and `wsgiref` were no longer being used and were automatically removed from the requirements file by `pip-dump`. However, upgrading the package on the server (which runs the above `pip install`) does not trigger their removal.

garrettr:
We should revisit this the next time we have to upgrade the Python dependencies the securedrop-app-code .deb package.

garrettr:
Just discussed this with @conorsch. We started working on this because of https://github.com/freedomofpress/securedrop/issues/856#issuecomment-86775726, and @fowlslegs has a good start towards a resolution in #1302. However, that solution makes me a little nervous and would require significant testing prior to release. Furthermore, this issue is not high priority. Since the outdated dependencies are no longer used by the application, I do not think they pose a significant security risk.

As a result, I am removing this from the 0.3.7 milestone. We should revisit it again in the near future.

garrettr:
@fowlslegs and I discussed this in person. We agreed that it is difficult to do this well because SecureDrop's Python dependencies are installed into the system-wide pool of Python packages, which makes it hard to disambiguate SecureDrop's Python packages from other Python packages.

The "right answer" is to properly isolate SecureDrop's deployment environment. There are numerous techniques for doing this, and doing so would be a significant change in the current deployment stack; therefore, they should be evaluated carefully and considered for a later release.

garrettr:
In general, [pip-sync](https://github.com/nvie/pip-tools) seems like a good tool to use here. The only thing preventing us from just dropping it in is:

> SecureDrop's Python dependencies are installed into the system-wide pool of Python packages, which makes it hard to disambiguate SecureDrop's Python packages from other Python packages.

from https://github.com/freedomofpress/securedrop/issues/856#issuecomment-221106486. If we isolate those environments, then it should be straightforward to use `pip-sync` to keep the installed dependencies in sync with the requirements files.

fowlslegs:
[...]

PR #1302, opened 2016-05-18
Autoremove unused pip dependencies in app code postinst script

fowlslegs:
This addition to the `postinst` script in the app code package removes all pip
packages that neither belong to the OS (i.e., were installed by system packages)
nor are present in the requirements file in the package that was just unpacked.
The script itself just tries to uninstall all pip packages not declared in the
requirements file, but pip is smart and will not let you uninstall OS owned
packages.

Use of arrays required bash, which has been added to the `Depends:` just to be
thorough. And a duplicate copy of `python-pip` was removed from `Depends:` as
well.

### DIFF
--- a/install_files/securedrop-app-code/DEBIAN/control
+++ b/install_files/securedrop-app-code/DEBIAN/control
@@ -6,5 +6,5 @@ Homepage: https://freedom.press/securedrop
 Package: securedrop-app-code
 Version: 0.3.6
 Architecture: amd64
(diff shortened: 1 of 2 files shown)


fowlslegs:
https://github.com/freedomofpress/securedrop/issues/856

fowlslegs:
Closing because the approach in https://github.com/freedomofpress/securedrop/issues/1472 is preferable.

Issue #1300, opened 2016-05-11
SecureDrop release notifications


garrettr:
There is currently no straightforward way for SecureDrop users to be informed when we release an update of SecureDrop.

In the past, we have blogged about releases that get audited on the Freedom of the Press Foundation blog, but those posts are focused on the results of the audit, and we do not write them for point/maintenance releases.

Our current policy is to privately inform all of the SecureDrop administrators that we are in contact with when a new release comes out. It tells them what to expect and provides instructions if any manual steps are required on the part of the administrator. There are two problems with this approach:

1. It does not help the administrators of instances that we are unaware of.
 * This is a shame because an important goal of making SecureDrop open source, and working as hard as we have been on the installation process and documentation recently, is to encourage people to install and maintain their own systems without our involvement.
2. It creates a repetitive support burden for Freedom of the Press Foundation (writing and sending emails to dozens of administrators, dealing with associated PGP issues, replying to their queries, etc.)

As a result, I propose that we create a dedicated channel for publishing notifications of SecureDrop releases, along with any additional helpful information (e.g. manual upgrade instructions, if they are necessary).

Proposed requirements:

* It should probably be a blog.
* It should be hosted on https://securedrop.org.
* It should have an RSS/Atom feed to people can easily integrate into their notification tool of choice.
* It should be easy for us to update and edit.

ageis:
I've been working on a blogging setup at https://securedrop.org/news, with RSS/Atom available at https://securedrop.org/news/feed. Still in progress: some styling/design aspects, adding the RSS link to the main page.

ageis:
The news portal and RSS feed is live and ready to go. I still need to run others through the publication process.

The "Read more link was originally a button. @garrettr suggested a different style of link that doesn't take up vertical space, so that instead the link is at the end of the preceding line of the summary. I agreed that would look cleaner, so I set about trying to make it happen. When I couldn't do it, I consulted some CSS wizards on IRC. They confirmed my suspicion, that the current HTML markup of these elements will not allow that to happen. Drupal generates the markup, and there's nothing we can do to change it. To be specific, there's a <p> element encapsulating the article summary, and a paragraph defines a block.... Without being able to get the "Read more" link to appear inside of the <p>, then it is not possible to get the layout to flow as if the link were inside of that block. We are stuck with what we have.

garrettr:
> We are stuck with what we have.

[...]

PR #1303, opened 2016-05-19
Upgrade Tails 1x to 2x Guide

garrettr:
I'm creating a new PR because Harlo's original branch is on her fork, which we don't all have permission to push to, and that was hindering Conor and I's ability to work on it. This new branch is on the freedomofpress remote, so we should all be able to push to it. Please treat this newfound power carefully (no rebasing, force pushing, etc.).

See my commit messages for the changes that I've made. I think this is almost ready to merge, but I would appreciate a final review from both @conorsch and @harlo.

Finally, I have not had time to test the **Troubleshooting** instructions, but from reading them I am a little concerned that they might not be complete or consistent. For example, in **2. Reinstall the SecureDrop Tails Configuration**, there's a warning that says "Copy ~/Persistent/.securedrop/torrc_additions to a place like your desktop. You’ll need these old values for the following step.", but then that file is never mentioned again. 😕


garrettr:
From the todo that I added in https://github.com/freedomofpress/securedrop/pull/1303/commits/559c3d809ab53bba2837a396ecfe77ce18ac3753:

> However, they also cannot re-run ``tails_files/install.sh`` at the moment, since they have 0.3.6 checked out and that ``tails_files`` doesn't work on Tails 2.x. At this point, we need to direct them to the upgrade guide for 0.3.7 (which is yet to be written).

I'm making a new issue to write the 0.3.7 upgrade guide, which will block this issue.

I think the best way to resolve this will be to write the 0.3.7 upgrade guide, then add a final section at the end saying something like "once you've upgrade to Tails 2.x, you should re-run the SecureDrop Tails setup script that is included in SecureDrop 0.3.7", then provide a link to the relevant part of the 0.3.7 upgrade guide.

garrettr:
On second thought, do we need a separate guide? As far as I can tell, the only manual step the admins will need to do is to re-run `tails_files/install.sh` from 0.3.7 after upgrading to Tails 2.x. In that case, might as well just add that step to this guide rather than creating a new one.

conorsch:
> On second thought, do we need a separate guide?

Discussed this in person with @garrettr. Thanks to #1300 we have prominent messaging for upgrade notifications to Admins, so we can emphasize the importance of upgrading the Tails devices in the release announcement. Rather than link out to a separate "Upgrade guide" with all of one step, we'll run through the usual GPG verification song-and-dance, and ask Admins to run the `install.sh` script again to update the NM hooks.

conorsch:
The changes here are solid. One small addition: the **Troubleshooting** section currently contains a single step, numbered **1.**, but there's no corresponding Step 2. Let's add a section title **Verify the store** on L296, after the `find` command and before the section link up to the top. That will help guide Admins to the second step back up to the top.

garrettr:
This still needs me to write the documentation for updating the Tails auto-configuration after upgrading to Tails 2.x.

conorsch:
> the documentation for updating the Tails auto-configuration after upgrading to Tails 2.x

That basically boils down to `cd ~/Persistent/securedrop/tails_files && sudo ./install.sh`. The plan is to state that in the release notification channel (described in #1300).

garrettr:
@conorsch I think it needs to go into this documentation as well. I'm ok with repeating ourselves a little bit. Think it's important that docs stand alone.

> That basically boils down to cd ~/Persistent/securedrop/tails_files && sudo ./install.sh.

You also need to pull from GitHub, check out the new release (>= 0.3.7), and verify the signed tag (_before_ re-running `./install.sh`).

PR #1305, opened 2016-05-20
update_version.sh autoinstalls deps, prepares changelog.md 

fowlslegs:
Now instead of telling us to install devscripts and git, update_version.sh does
it itself. Since we're only running it from our Ubuntu build VM, we don't need
to worry about system compatibility.

This also puts the subheader for the new version in changelog.md and sets vim to
drop us right into the place in the changelog we're going to be writing in.

Because we are lazy, lazy devs.

### DIFF
--- a/update_version.sh
+++ b/update_version.sh
@@ -7,9 +7,7 @@ dch_installed=$?
 command -v git > /dev/null
 git_installed=$?
 if [ $dch_installed -ne 0 ] || [ $git_installed -ne 0 ]; then
-  echo "You must run this on a system with dch and git available (in order to edit the Debian package changelog and commit message)"
-  echo "If you are on Debian/Ubuntu, apt-get install devscripts git"
-  exit 1
+  sudo apt-get install -y devscripts git
 fi
 
 set -e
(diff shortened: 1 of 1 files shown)


update_version.sh
```diff
@@ -7,9 +7,7 @@ dch_installed=$?
 command -v git > /dev/null
 git_installed=$?
 if [ $dch_installed -ne 0 ] || [ $git_installed -ne 0 ]; then
-  echo "You must run this on a system with dch and git available (in order to edit the Debian package changelog and commit message)"
-  echo "If you are on Debian/Ubuntu, apt-get install devscripts git"
-  exit 1
+  sudo apt-get install -y devscripts git
```
garrettr: I admit this is *incredibly* half-assed, but part of the reasoning for checking this stuff was to make sure that this script is only run in the Vagrant build machine (the comment kinda indicates that that was the goal).

Would you mind adding a test before this change to that effect? Something like "if $(whoami) is not "vagrant", exit with an explanatory message about only running this in the build VM".

garrettr: Minor: If you're just gonna install the dependencies, you can probably remove or at least simplify the conditionals that check if those dependencies are already installed. For example, `apt-get install -y` won't do anything wrong if the packages you want are already installed. [Some ideas from StackOverflow](http://stackoverflow.com/questions/1298066/check-if-a-package-is-installed-and-then-install-it-if-its-not).

garrettr:
I freely admit that the update_version script sucks. These changes are reasonable minor improvements, but I'm not sure if I think they should go into 0.3.7. My inclination is to merge them into develop instead.

conorsch:
> My inclination is to merge them into develop instead.

+1

> Would you mind adding a test before this change to that effect? Something like "if $(whoami) is not "vagrant", exit with an explanatory message about only running this in the build VM".

+1 Especially important now that the script automatically installs packages, so it shouldn't be run willy-nilly on developer workstations. I'd check for both user "vagrant" and hostname "build".

fowlslegs:
Closed in favor of #1433.

Issue #888, opened 2015-02-18
usage of undocumented / internal python stdlib class _TemporaryFileWrapper 


ThomasWaldmann:
... by secure_tempfile module.

There is no docs about why it is not simply using NamedTemporaryFile or TemporaryFile.

It looks like the computed random filename has better randomness as it is using 32 bytes from os.urandom().

OTOH, it could be checked whether a filename is needed at all or maybe a nameless file created by TemporaryFile is maybe not even better. Or whether the less great randomness of the stdlib generated filename is not also good enough for the purpose.

garrettr:
There are several things that could potentially be improved here.

## Use of _TemporaryFileWrapper

The motivation for using this class is to get the "auto delete on close" behavior. We are injecting this into Flask (see `request_that_secures_file_uploads.py`), so we don't have an opportunity to do explicit cleanup and instead must rely on close eventually being called. It would not be hard to implement this ourselves, but there is some subtlety and corner cases that need handling, so why not re-use the existing implementation in _TemporaryFileWrapper?

I guess the alternative would be:

1. Don't inherit from `_TemporaryFileWrapper`
2. In `__init__`, `self.file = tempfile.NamedTemporaryFile(...)`
3. When `SecureTemporaryFile` is GC'ed, `self.file` will be GC'ed, and `.close` will be called automatically (right?)

The last step would need to be verified. I am not sure whether Flask calls `.close` on the `_secure_file_stream` explicitly or if it is implicit from the GC.

## Filenames

I think the 32 bytes of randomness for the filename is actually overkill, and the filenames generated by `tempfile._mkstemp_inner` would work just as well. The only thing you want to avoid is accidentally overwriting other files, or running out of possible filenames under normal use. I didn't write code to make sure `tmp_file_id` is unique before using, and just assume that the probability of a random 32-byte collision is small enough that it would not be a concern. The approach taken in tempfile is to use a much shorter filename with less entropy, but it loops over the filenames and [makes sure it has a unique name](https://hg.python.org/cpython/file/2.7/Lib/tempfile.py#l243) before continuing.

In retrospect, I prefer the approach taken by tempfile and think it would be good to change `SecureTemporaryFile` to do that instead.

## Do we need a name at all?

[...]

PR #1306, opened 2016-05-20
Pip package upgrade for release 0.3.7

### LINKED ISSUE #946: Use upstream version of python-gnupg
See https://github.com/freedomofpress/securedrop/issues/250#issuecomment-86237456 for context. We have worked with the upstream maintainer to fix the issues that were preventing us from using the upstream version with SecureDrop. Now that they are fixed, we should start using the upstream and discontinue our use of the outdated `gnupg-securedrop` [fork](https://pypi.python.org/pypi/gnupg-securedrop/1.2.5-9-g6f9d63a-dirty).

### LINKED ISSUE #1301: Upgrade SecureDrop's Python dependencies
We've been blocked on issuing a new release for a long time, mostly due to the increasing support burden for the growing number of SecureDrop installations. As a result, SecureDrop's Python dependencies (which are packaged and installed via a Pip wheel in the `securedrop-app-code` .deb package) are out of date, and should be updated ASAP.

Related: #487, #856.

fowlslegs:
Upgrades SecureDrop's Python dependencies to the latest versions (resolves #1301). Rewrites the `pip_update.sh` helper script to use `pip-compile`, which helps disambiguate top-level dependencies from nested dependencies. The top-level dependencies are now listed in `securedrop/requirements/*.in`, and pip-compile compiles these into helpfully commented requirements files: `securedrop/requirements/*.txt`.

Updates the Python gnupg package to use the current upstream version instead of the outdated pinned version that we were using before (resolves #946). To make SecureDrop compatible with the upstream, needed to:

1. API change for encryption (encrypt_fn --> encrypt)
2. Add our custom anti-forensic stream wrapper type to gnupg's recognized stream-like types


fowlslegs:
https://github.com/freedomofpress/securedrop/issues/1301

fowlslegs:
https://github.com/freedomofpress/securedrop/issues/946

fowlslegs:
I think all the kinks in the update script have been smoothed out, which is the less important part, but will be useful as we intend to track upstream much more closely than in the relative gap in development that we've spent working on infrastructure projects.

I did a little bit of manual testing of the new dependencies in the staging VMs, but mostly I relied on the unit tests.

pip_update.sh
```diff
@@ -2,35 +2,52 @@
...
+pip install --upgrade pip==8.1.1
+pip install pip-tools
+
+# Compile new requirements (.txt) files from our top-level dependency (.in)
+# files. See http://nvie.com/posts/better-package-management/
+for r in "securedrop-requirements" "test-requirements"
+do
+  # Maybe pip-tools will get its act together and standardize their cert-pinning
+  # syntax and this line will break. One can only hope.
+  pip-compile -o $r".txt" $r".in"
```
garrettr: Are these double quotes necessary?

securedrop/secure_tempfile.py
```diff
@@ -1,12 +1,16 @@
 import base64
 import os
 from tempfile import _TemporaryFileWrapper
+# python-gnupg will not recognize our SecureTemporaryFile as a stream-like type
+# and will attempt to call encode on it, thinking it's a string-like type. To
+# avoid this we add it the list of stream-like types.
+from gnupg._util import _STREAMLIKE_TYPES
+_STREAMLIKE_TYPES.append(_TemporaryFileWrapper)
```
garrettr: Why not `SecureTemporaryFile` instead?

pip_update.sh
fowlslegs: Yes, otherwise the interpreter will think you're trying to reference the non-existent `r.txt` variable.

securedrop/secure_tempfile.py
fowlslegs: I just put it at the top because I liked the formatting better (i.e., having it right next to the import block and then leaving the rest of the file to the sft class), and because `_TemporaryFileWrapper` is also a stream-like type. I doubt it's a type will ever be used in SecureDrop, but at least it's correct appended.

So maybe it could matter less to you to see it moved to the bottom, and instead it's really important to you that we're as granular as possible. Or maybe you were just curious.

Personally, I don't see much of a difference either way, so if you feel passionately about one approach, we can change it.

garrettr: It's kind of a hack that we use an internal type (`_TemporaryFileWrapper`, prepended by a `_`), and we've already had at least one person complain about it: https://github.com/freedomofpress/securedrop/issues/888. I'd prefer to minimize our use if possible. Also, appending `SecureTemporaryFile` instead of `_TemporaryFileWrapper` is more explicit and that is almost always better.

pip_update.sh
conorsch: pip-compile -o $r".txt" $r".in"

There is a more readable solution for interpolating vars in Bash:
```
pip-compile -o "${r}.txt" "${r}.in"
```

See here: https://google.github.io/styleguide/shell.xml?showone=Variable_expansion#Variable_expansion

securedrop/secure_tempfile.py
```diff
@@ -77,3 +78,8 @@ def read(self, count=None):
 
     def close(self):
         return _TemporaryFileWrapper.close(self)
+
+# python-gnupg will not recognize our SecureTemporaryFile as a stream-like type
+# and will attempt to call encode on it, thinking it's a string-like type. To
+# avoid this we add it the list of stream-like types.
+_STREAMLIKE_TYPES.append(_TemporaryFileWrapper)
```
garrettr: ❤️

garrettr:
This is good to go, thanks @fowlslegs! Need to figure out the best order to merge all of these into release/0.3.7, but once I do, this will be merged.

PR #1198, opened 2015-11-13
Fix duplicate lines in install.sh related to removing ~/.xsessionrc

ageis:
Apparently some code was included twice - my mistake. I'll make a point to test the new tails_files script on the latest Tails in the forthcoming days.

PR #1309, opened 2016-05-24
Overhaul install.sh for Tails environment configuration

conorsch:
## Motivation
The `install.sh` script is used for bootstrapping a new Tails device for SecureDrop, either an Admin or a Journalist Workstation, and has assumed responsibility for a variety of important config tasks:

* prepare ATHS `HidServAuth` entries for `/etc/tor/torrc`
* install NetworkManager hook for updating `/etc/tor/torrc`
* create desktop icons for Document and Source Interfaces (clickable!)

The script also configures noncritical, but still useful, functionality that's relevant only to Admins:

* add `ansible` to apt persistence
* create SSH host aliases
* update static inventory file with .onion addresses

Last but certainly not least, the script cleans up deprecated solutions that may be in place, due to changes in our config and our strategy for provisioning Tails workstations. A prime example of such cleanup is in #1285. Such cleanup requires manual invocation, which we'll need to recommend in upgrade documentation for the new release. (The proverbial "carrot" for such manual intervention is that the new NetworkManager hook order configures access to the Document Interface much faster than it used to! The speed-up is noticeable, a different of several minutes in some casual testing.)

### Idempotence (kinda)
Since we'll be instructing Admins to rerun the `install.sh` script, we must make sure it's smart enough to handle first-run use, and upgrade logic. Therefore this refactor implements a somewhat superficial idempotence. The script tries hard to find the necessary information (mostly ATHS info) from a variety of locations, before prompting the user interactively to type it in. Mostly this tenacity is designed to save Journalists from the hassle of looking up the relevant `HidServAuth` line in their torrc and pasting it into a dialog box.

### Readability == maintainability
The refactor also implements a variety of best practices for shell scripting. See the [Google Shell Style Guide](https://google.github.io/styleguide/shell.xml#Formatting) for a rough reference. Changes include lowercase var names, using bracket notation when interpolating vars in strings, organizing all logic into functions callable from a succinct `main()` function, and adding mountains of comments.

### Other PRs
Several files have changed in this PR. That's because it bottles up several closely related PRs, all of which pertained to `tails_files/install.sh`. The complete list of related PRs accounted for here is:

* #1182 
* #1198 
* #1206 
* #1277 
* #1285 
* #1287 
* #1292  

As such, there are a few docs changes along for the ride, as well as updates to other scripts within `tails_files`. All are intended for shipping with the impending 0.3.7 release. 

## Has it been tested?

Certainly! But perhaps you can do better. I've made sure that running this script Just Works:tm: on Admin and Journalist Workstations running 0.3.6 of SecureDrop. No prompts are levied at the user. Removing relevant ATHS from the host config causes the prompts to display, which is good. Deprecated NetworkManager hooks are cleaned up properly, and stay gone even after reboots (took me until 063058a to nail that one down). 

See if you can find any inappropriate assumptions the script is making. By reorganizing the script into a list of functional definitions, with the only actual running code called in the final `main` block, I've tried to make it readable and intelligible. If reading through the script doesn't make its utility obvious to you, say so, and I'll improve the comments and labeling wherever possible.


tails_files/install.sh
```diff
@@ -68,23 +87,360 @@ if [ -f $XSESSION_RC ]; then
...
+}
+
+function lookup_document_aths_url()
+{
+  # First look for the flat file containing the exact value we want.
+  # The Admin Workstation will certainly have this file, and the Journalist
+  # Workstation probably, but not definitely.
+  if [ -f "${securedrop_ansible_base}/app-document-aths" ] ; then
+    app_document_aths="$(awk '{ print $2 }' ${securedrop_ansible_base}/app-document-aths)"
+  # Failing that, check for ATHS info in the existing Tor config. A previously
```
ageis: Thinking ahead, we should probably support provisioning the same JW to be used for new ATHSes, i.e. rotated hidden service private keys and URLs. This happened during [Heartbleed](https://securedrop.org/news/securedrop-and-openssl-vulnerability) and it could be necessary again with next-generation Hidden Services (elliptic curve), or server compromise. However, if rotating keys and URLs is not currently supported in the SecureDrop server playbooks, then it shouldn't be in tails_files.

tails_files/install.sh
```diff
@@ -68,23 +87,360 @@ if [ -f $XSESSION_RC ]; then
...
+SD_COMPLETE_MSG1
+
+  # We're assuming the SSH aliases have been set up as long as ~/.ssh/config exists.
+  # Intentionally padding with an empty line to improve readability.
+  if is_admin_workstation; then
+    cat <<SD_COMPLETE_MSG2
+
+The App Server's SSH hidden service address is: $(lookup_app_ssh_aths_url)
+The Monitor Server's SSH hidden service address is: $(lookup_mon_ssh_aths_url)
+SSH aliases are set up. You can use them with 'ssh app' and 'ssh mon'.
```
ageis: Since the conditional is upon `if_admin_workstation`, this message about the aliases could still be output (as misleading) even if per line https://github.com/freedomofpress/securedrop/pull/1309/files#diff-a44dd5740ffe7a0588a02b9c152ec8edR255 there's an existing SSH config. I think you probably already know that though and made a judgment call that it's okay. That would only happen if an old, customized Tails had been re-purposed for SecureDrop, so it's not likely.

garrettr:
Reviewing the code now. Just wanted to say this is an exemplary pull request. Thank you.

tails_files/install.sh
```diff
@@ -1,60 +1,79 @@
...
+  # Detect whether the script is being run on Admin or Journalist Workstation.
+  # Under no circumstances should the Journalist Workstation have SSH info,
+  # so we'll assume that if the ATHS file for SSH info on the Application Server
+  # is present, we're running on the Admin Workstation, otherwise Journalist.
+  if [ -f $securedrop_ansible_base/app-ssh-aths ]; then
+    # In Bash, 0 equates to Boolean true and 1 to Boolean false.
+    return 0
+  else
+    return 1
 fi
```
garrettr: Nit: indent on same level as opening statement.

garrettr:
Something that has always confused me is that `install.sh` assumes it's possible that `app-{source-ths,document-aths}` will be available in `~/Persistent/securedrop/install_files/ansible-base/` before the install script is run on the Journalist Workstation. For this to happen, an admin or the journalist would have to manually transfer the relevant files to the correct directory, *but this not clearly documented anywhere*, at least as far as I can tell. I would expect to find a mention of it [here](https://docs.securedrop.org/en/latest/onboarding.html#set-up-automatic-access-to-the-document-interface), but these docs only say:

> you should enter the lines from app-document-aths and app-source-ths when prompted to by the install.sh script.

Seems to weird to do something automagical when the base conditions for its success are not documented. I think it would be appropriate to update the relevant documentation as part of this pull request.

garrettr:
I appreciate the automagic in the `lookup_*ths_url` functions, but I am concerned it might lead to unwanted outcomes.

Here's an example:

1. A SecureDrop installation is performed, and an Admin Workstation and Journalist Workstation are set up.
2. Some time later, the admin decides to reinstall the SecureDrop from scratch, and does not bother with backing up and restoring the THS values for whatever reason. (this scenario has happened several times, FWIW)
3. The admin wants to update the Journalist Workstation, so they copy the relevant `*ths` files - but they *do not* place them in `$securedrop_ansible_base`, because, at least at the moment, nothing in the documentation indicates that they should do that (see https://github.com/freedomofpress/securedrop/pull/1309#issuecomment-221307645).
4. They re-run `install.sh` on the Journalist Workstation. It falls through the conditionals to find the old SecureDrop installation's ATHS values in the Desktop icons, and uses them (incorrectly).
5. The admin does not understand why the journalist workstation won't connect to the new SecureDrop.

I am not sure what the best solution is here, but here are some ideas:

1. Automagic is more trouble than it's worth. Prefer explicit over implicit, even it's inconvenient. Prompt every time (but continue with the memoizing, so you only have to prompt once for each value).
2. Clearly document best practices for Succeeding With AutoMagical Setup Scripts (tm). That is, explain that everything will work better if you move the right files to the right locations in `$securedrop_ansible_base` before running `install.sh` (although it's not clear to me if this is really more ergonomic than just prompting for the values).
3. If you continue with automagic, print out when and where you automagically found a HidServAuth value to help admins understand what is going on, and aid in debugging. (e.g. "Found HidServAuth for the Document Interface in a Desktop icon, using that...").
4. Tell Admins that they have to wipe and recreate the Persistent volumes on all of their Tails USBs every time they re-install SecureDrop. "It's the only way to be sure".

tails_files/install.sh
```diff
@@ -68,23 +87,360 @@ if [ -f $XSESSION_RC ]; then
...
+  cp -f 65-configure-tor-for-securedrop.sh "${tails_live_persistence}/custom-nm-hooks"
+  cp -f 65-configure-tor-for-securedrop.sh "$network_manager_dispatcher"
+  chown root:root \
+    "${tails_live_persistence}/custom-nm-hooks/65-configure-tor-for-securedrop.sh" \
+    "${network_manager_dispatcher}/65-configure-tor-for-securedrop.sh"
+  chmod 755 \
+    "${tails_live_persistence}/custom-nm-hooks/65-configure-tor-for-securedrop.sh" \
+    "${network_manager_dispatcher}/65-configure-tor-for-securedrop.sh"
+
+  # Run the SecureDrop init script. Among other things, implements the torrc
```
garrettr: > Among other things

Actually, the _only_ thing the init script does is implement the torrc additions.

tails_files/install.sh
```diff
@@ -68,23 +87,360 @@ if [ -f $XSESSION_RC ]; then
...
+    # During initial provisioning of the Journalist Workstation, the ATHS file
+    # for the Document Interface may not be present. If not, prompt for the info.
+    aths_regex="^(HidServAuth [a-z2-7]{16}\.onion [A-Za-z0-9+/.]{22})"
+    # Loop while reading the input from a dialog box.
+    while [[ ! "$document_aths_info" =~ $aths_regex ]]; do
+      document_aths_info="$(zenity --entry \
+        --title='Hidden service authentication setup' \
+        --width=600 \
+        --window-icon=${securedrop_dotfiles}/securedrop_icon.png \
+        --text='Enter the HidServAuth value to be added to /etc/tor/torrc:')"
```
garrettr: Please make sure you are *very clear* about whether or not "the HidServAuth value" does or does not include the "HidServAuth" prefix. This is a common point of confusion for admins at this stage in the process.

ageis: Might this be a matter for the docs? I suppose the Zenity prompt could be made more verbose. I'd like to point out that we have a doc for "Configure the Admin Workstation Post-install" but we do not have a doc for "Configure the Journalist Workstation".

garrettr: Certainly it should be in the documentation, but it doesn't hurt to make sure it's clear in the prompt as well. Put yourself in the admin's shoes. Imagine what would make their life easier.

> I'd like to point out that we have a doc for "Configure the Admin Workstation Post-install" but we do not have a doc for "Configure the Journalist Workstation".

That's incorrect, see https://docs.securedrop.org/en/latest/onboarding.html#set-up-automatic-access-to-the-document-interface. I'm not saying those docs couldn't be improved, but they do at least exist.

tails_files/install.sh
```diff
@@ -68,23 +87,360 @@ if [ -f $XSESSION_RC ]; then
...
+}
+
+function set_permissions_on_securedrop_dotfiles()
+{
+  # Follow up function to enforce correct permissions and ownership
+  # on SecureDrop dotfiles in the staging directly. Separated
+  # into function to aid readability.
+  chown root:root $securedrop_init_script
+  chmod 700 $securedrop_init_script
+  chown root:root $torrc_additions
```
ageis: There's a problem here. Because the previous function `copy_securedrop_dotfiles()` has not copied the `torrc_additions` template into `.securedrop`, we are chown'ing a file that does not exist yet.

`chown: cannot access '/home/amnesia/Persistent/.securedrop/torrc_additions': No such file or directory`

To fix this, you can either cp `torrc_additions` from tails_files in the preceding function, or re-order the program flow so that `configure_torrc_additions()` comes earlier.

tails_files/install.sh
conorsch: Thus the comment:

> We're assuming the SSH aliases have been set up as long as ~/.ssh/config exists.

We can't do much better than that in Bash. I'm on the fence about the utility of the SSH aliases in the first place, but our docs currently reference them, so adding support.

tails_files/install.sh
conorsch: Thank you kindly, fixed in 91c9a3a.

tails_files/install.sh
conorsch: Fixed in 89b4862. :smiley:

tails_files/install.sh
conorsch: Adjusted the call order in a97ca49. Can you test again on a fresh Tails environment and confirm the new order does the job?

tails_files/install.sh
conorsch: > Thinking ahead, we should probably support provisioning the same JW to be used for new ATHSes

Out of scope for this PR. Providing playbooks for rotation and upgrade is a higher priority than altering this script, but both should be handled in a separate discussion.

conorsch:
@garrettr Great comments, thanks for the thoughtful review. 

> Something that has always confused me is that `install.sh` assumes it's possible that `app-{source-ths,document-aths}` will be available in `~/Persistent/securedrop/install_files/ansible-base/` on the Journalist workstation before the install script is run. [...] I think it would be appropriate to update the relevant documentation as part of this pull request.

That's absolutely true. I myself routinely place the `*ths` files in `ansible-base` on the Journalist Workstation when testing a fresh install or setting up a live instance, but our docs don't recommend it. The single biggest reason to adhere to this practice is to avoid mistakes, so I'm not typing in Onion URLs or futzing with copy/pasting the wrong files. Given that the pragma is intended to save time and reduce mistakes, it should indeed be part of our docs. However, we still can't assume that the files will be there on the Journalist Workstation.

> If you continue with automagic, print out when and where you automagically found a HidServAuth value to help admins understand what is going on, and aid in debugging. 

That's solid advice in several respects: the script is quiet for several seconds, then gets chatty printing the "success" message. Having more informative output would be helpful to everyone, so I'll add in a few `echo` statements to make the flow obvious.

ageis:
I have reviewed this PR both visually, line by line, and tested it on fresh and upgraded Tails 2.x. Most of my comments and concerns have been addressed and all important features or fixes that I added in previous commits/PRs have been included. 

Great work on the comments and encapsulating each piece in functions. This will make maintaining, understanding and contributing to `tails_files` much easier in the future.

There are two things I would raise at this point, but should not be considered blockers:

* The `tails_files/torrc_additions` template, which contains only a comment, is no longer necessary, since the script creates the file in `.securedrop` via cat+append. It can be removed from the repo.
* We've discussed not needing to prompt for the SSH username. This could be done with `parse_yaml()`, but beyond the issue with that being a dirty hack, there is another issue which is that `ssh_users` can take a list.

You personally have my vote for merge.

conorsch:
Updated the Admin and Journalist Workstation setup docs with strong recommendations to reuse the flat files for A/THS setup:

* `app-document-aths`
* `app-source-ths`

Even included a scary warning about **not** copying over the corresponding SSH info files. I've also added `echo` statements throughout the script as @garrettr suggested, and will rerun the scripts. Please build the docs locally from this PR and read through **Configure the Admin Workstation Post-Install** and **Onboard Journalists**.

Still need to rework the zenity dialog prompts to request explicit formatting for the HidServAuth values.

tails_files/install.sh
conorsch: Improved the Journalist Workstation docs somewhat in 4e5319e. Still need to strengthen the guiding text in the interactive zenity prompt—will need to test multiline string handling.

garrettr:
> Please build the docs locally from this PR and read through Configure the Admin Workstation Post-Install and Onboard Journalists.

👍

PR #1314, opened 2016-06-02
Document re-installing the Tails auto-config

garrettr:
This is the final step in upgrading an existing Tails (1.x) USB to Tails 2.x.

r? @conorsch

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -65,6 +65,7 @@ anonymous sources.
    upgrade/upgrade-0.3.x.rst
    upgrade/0.3.x-to-0.3.5.rst
    upgrade/0.3.5-to-0.3.6.rst
+   upgrade/0.3.6-to-0.3.7.rst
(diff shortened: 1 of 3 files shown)


conorsch:
These look good. I'll follow this guide during a final round of QA to make sure everything works as expected. Some thoughts:

* I'd expect the stash-pull-verify-pop documentation to go in a separate Upgrade guide specifically for 0.3.7. We did the same for 0.3.6, recommending the key rotation. The topic guide for Tails can link out to the general SecureDrop upgrade guide.
* We don't explicitly mention that the tails_files script only needs to be rerun on the Admin and Journalist Workstations. It's obvious to you and me that that's the case, but let's be as clear as possible in the docs.

Will report back with feedback after testing on hardware—let's keep this open until then.

garrettr:
r+ from in-person review with @conorsch.

PR #1320, opened 2016-06-03
updated diceware attribution

heartsucker:
Updated readme to link to correct diceware list.


garrettr:
Dammit, I totally failed at due diligence on merging #1276! Oh well, thanks for catching this 😄

PR #1321, opened 2016-06-09
Remove previous little poor wordlist

nogoegst:
I guess we don't need it anymore (since #1276).


garrettr:
Good catch, @nogoegst! I should've made sure the old wordlist was removed in #1276, but this PR takes care of it.

PR #1322, opened 2016-06-09
Fixes to the 0.3.6 upgrade docs

ageis:
These two issues were reported to me in person by @flamsmark.

Ping @garrettr.

### DIFF
--- a/docs/upgrade/0.3.5-to-0.3.6.rst
+++ b/docs/upgrade/0.3.5-to-0.3.6.rst
@@ -96,6 +96,11 @@ Greeter*.
     sudo apt-get update
     sudo apt-get install ansible
 
+#. Change your current working directory to where the files Ansible needs are
+stored ::
+
+    cd install_files/ansible-base
+
 #. Run the Ansible playbooks (substitute the admin account on the servers for 
    ``<user>``) ::
 
--- a/docs/upgrade/0.3.5-to-0.3.6.rst
+++ b/docs/upgrade/0.3.5-to-0.3.6.rst
@@ -91,7 +91,7 @@ Greeter*.
      git stash pop
 
 #. Make sure you have Ansible installed. If running ``which ansible`` returns 
-   ``ansible not found``, you should ::
+   nothing, you should ::
 
     sudo apt-get update
     sudo apt-get install ansible


docs/upgrade/0.3.5-to-0.3.6.rst
```diff
@@ -91,11 +91,16 @@ Greeter*.
      git stash pop
 
 #. Make sure you have Ansible installed. If running ``which ansible`` returns 
-   ``ansible not found``, you should ::
+   nothing, you should ::
```
garrettr: Huh. I wonder if that changed between Tails 1.x and 2.x?

ageis: @garrettr I don't think so. I ran `which` on Tom's admin workstation which uses Tails 1.x, and I ran version 2.4 on a VirtualBox VM on my computer, and saw the same behavior.

garrettr:
These are good changes. The tricky part is figuring how where to merge them. Obviously they _can_ go into develop, but SecureDrop end-users (sources, journalists, and administrators) are supposed to read the "stable" docs based on the latest signed release tag. Since they seem like the intended audience, these changes should go into the "stable" docs - but currently, those only get updated when we do a new release.

I'm disinclined to do a point release just to update documentation. According to the current release protocols, we are supposed to keep the following in sync at all times:

1. Latest signed git tag
2. Version of packages of apt.freedom.press

Otherwise, the concern was that things would get confusing quickly (the latest release is 0.3.8, but my source interface's home page says 0.3.7?!? Oh, but that really means it's running 0.3.8? Oh, actually it's ambiguous? ...do not want).

Since doing a new point release is a lot of work, due to:

1. Building new packages.
 * Even if the package contents don't change, the release protocol dictates that we should release "no-op" packages to keep the version numbers in sync everywhere. We did this for 0.3.6, where the relevant changes were only to the Ansible playbooks and documentation.
2. Airgapped signing process is manual
3. Need to write a blog post, tweet about it, answer the inevitable questions that come in, etc.

Therefore, since these changes are relatively minor, my inclination is to merge them into develop, and prioritize shipping 0.4 (which will be based on develop) ASAP. What do you think, @conorsch and @flamsmark?

FWIW, I'm happy to provide more detail if any of my explanation of the issue is confusing to anybody. I would totally understand if it _is_ confusing. This is also a good opportunity for constructive criticism of the SecureDrop release protocol. There is copious internal documentation of the release protocol, but I am reluctant to share it directly for reasons of opsec. If people are *really* interested, I would take the time to excerpt the relevant bits into a public document.

docs/upgrade/0.3.5-to-0.3.6.rst
garrettr: Huh, it says "foobar not found" on my Mac, but it's reasonable to expect differences between BSD `which` (on Mac OS X) and Linux's `which`. Anyway. Thanks for double-checking on both version families of Tails, @ageis 💯

flamsmark:
It's tough. Currently, an admin needs to use sysadmin critical-thinking skills (and knowledge of Ansible!) in order to make the jump from 0.3.5 to 0.3.6. That seems like a showstopper to me, so I think a point release of 0.3.6.1 is justified.

garrettr:
Merged into `develop` for 0.4 and `release/0.3.7.post1` for 0.3.7.post1.

Issue #1325, opened 2016-06-10
update_version.sh should update new documentation


freddymartinez9:
Currently the "set_up_admin_tails" documentation looks like this, I suspect that between every release FPF just increments the number (from .3.6 to .3.7). Luckily we have computers to automate small tasks for us :-)  Is there a programmatic way to do this? 
```
cd securedrop/
git checkout 0.3.6
git tag -v 0.3.6
```

freddymartinez9:
Also I can't seem to tag the latest release, or at least I am getting thrown an error verifying @garrettr's key 
```
user@hostname: git checkout 0.3.7
HEAD is now at 3a7e1dd... SecureDrop 0.3.7
user@hostname: git tag -v 0.3.7
object 3a7e1ddc63e56282ad28982f8a4552ad689d0e63
type commit
tag 0.3.7
tagger Garrett Robinson <garrett@freedom.press> 1464915950 -0700
SecureDrop 0.3.7
gpg: Signature made Thu 02 Jun 2016 08:22:48 PM CDT using RSA key ID 4833B9A3
gpg: Can't check signature: public key not found
error: could not verify the tag '0.3.7'
```
my git repo is up to date from the latest release:

```
user@hostname:git log -p | head -n 6
commit 3a7e1ddc63e56282ad28982f8a4552ad689d0e63
Author: Garrett Robinson <garrett@freedom.press>
Date:   Thu Jun 2 17:52:40 2016 -0700

    SecureDrop 0.3.7
```

ageis:
Hey @freddymartinez9. Per our installation documentation, you should have fetched the key to the Admin Workstation during the install process. See: https://docs.securedrop.org/en/latest/set_up_admin_tails.html under "Verify the Release Tag". If you run that `gpg --recv-key` command in that doc, then you will be able to verify the git tag.

ageis:
@freddymartinez9 Also, I'm noting from your shell prompt that you don't seem to be on Tails. What is `user@hostname`?

freddymartinez9:
yes it is a tails drive. the user is not amnesia because I was working
around an issue with DNS in tails (failing to resolve) so I've loaded the
Persistent volume in a dedicated qubes VM. If that's an absolute no-no from
a security perspective, let me know. I suspect that also triggered the GPG
error (no keyring to check against). I'll test again...

ageis:
[...]

PR #1326, opened 2016-06-10
Bump up current release in Set up Admin Tails doc

ageis:
Reported in #1325. The 'Set up the Admin Workstation' guide is not tracking the current release.


garrettr:
Subsumed by #1336.

PR #1329, opened 2016-06-11
flask auto-reset on updated templates

heartsucker:
Change causes Flask to auto-reset whenever a template is added, removed, or updated. This should make it easier to develop the HTML side of things.

Also, removed a couple of unused imports.

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-import sys
 import os
 from datetime import datetime
 import functools
@@ -8,7 +7,6 @@ from flask import (Flask, request, render_template, send_file, redirect, flash,
                    url_for, g, abort, session)
 from flask_wtf.csrf import CsrfProtect
 from flask.ext.assets import Environment
-from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
 from sqlalchemy.exc import IntegrityError
 
 import config
@@ -18,11 +16,11 @@ import store
 import template_filters
 from db import (db_session, Source, Journalist, Submission, Reply,
                 SourceStar, get_one_or_else, NoResultFound,
-                WrongPasswordException, BadTokenException,
                 LoginThrottledException, InvalidPasswordLength)
 import worker
 
 app = Flask(__name__, template_folder=config.JOURNALIST_TEMPLATES_DIR)
+app.debug = False  # safe default
 app.config.from_object(config.JournalistInterfaceFlaskConfig)
 CsrfProtect(app)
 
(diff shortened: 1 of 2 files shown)


fowlslegs:
Thanks for the pull @heartsucker. I'll review this and hopefully merge it within the next few days.

garrettr:
@heartsucker Can you explain a little more about why this is useful? Maybe a specific development scenario where this change would have helped you?

I ask because I can't think of any time I've wanted the Flask application to auto-reload when the templates change. The templates are read anew every time a response is generated, you all you have to do to test a template change is refresh the page. I can't see any benefit to triggering application auto-reload on template change.

Note that Flask application auto-reloading != the page you are on auto-reloading when underlying static assets (templates, js, css) change. If that's the behavior you are seeking, I think you need to use something more sophisticated, like [flask-reloaded](https://github.com/bollwyvl/flask-reloaded).

heartsucker:
> The templates are read anew every time a response is generated

That's what I've read elsewhere, but on my Flask apps and on the version SD is running, that is not the observed behavior. Maybe I'm getting cached HTTP responses or Flask is caching the rendered output. Anyway.

The idea is that if someone is tinkering with the app and makes an HTML change, they can just ctrl-R and get the udpated content on the page. For me the only way to get that immediately was to restart the app. And yes, I know this isn't having the webpage auto reload in the browser.

Flask Reloaded would actually be a better fix for faster development. If this rapid developement fix is something you guys think would benefit other devs, I can test that and leave notes here. If the opinion is that having the app refresh on changes like this is unnecessary, we can just close this.

garrettr:
@heartsucker This works for me:

```
$ vagrant up development
$ vagrant ssh
# in development VM
$ cd /vagrant/securedrop
$ ./manage.py run
```

Open one of the development interfaces in your web browser, say `localhost:8080` for the development Source Interface. Now open one of the templates (e.g. `source_templates/index.html`) in your preferred editor on your local machine. Edit the file to produce a visible change, and save it. Go back to your browser and reload the page.

I just tested this and can confirm that I see the results of the changed template immediately upon reloading the page, which is the expected behavior.

Are you doing anything differently?

heartsucker:
@garrettr That is exactly what I'm doing (and effectively what I'm doing in my other projects). I'm going to attribute this to my ability to attract the strangest of Heisenbugs and say that this isn't worth merging.

If you're curious:
```
heartsucker@archimedes:~$ vagrant --version
Vagrant 1.8.1
heartsucker@archimedes:~$ VBoxManage --version
5.0.20r106931
heartsucker@archimedes:~$ python --version
Python 2.7.11
heartsucker@archimedes:~$ ansible --version
ansible 2.1.0.0
  config file =
  configured module search path = Default w/o overrides
```

garrettr:
> ansible 2.1.0.0

I'm surprised that's working for you; I think we're still waiting to refactor the playbooks to be Ansible 2.0-compatible. But that probably isn't the cause of your issue with reloading.

garrettr:
What OS are you on? Just a guess, but is this possibly an issue with Vagrant's synced folders?

heartsucker:
@garrettr I'm on OSX 10.11.5. I don't think it's an issue with the synced folders because I have this issue with Flask running it directly on my Mac for other projects and also on my Debian 8 machine. And the configs are basically the same since I copied a bit of the project the from here. :D

[here](https://github.com/berlin-leaks/berlinleaks.org/) and [here](https://github.com/heartsucker/heartsucker.com)

mdrose:
@heartsucker My experience is that I also get a newly generated page on new requests without having to restart Flask. I'm also running Debian 8. Not sure what the difference might be.

garrettr:
Closing for now, since this issue appears to be unique to @heartsucker's environment.

garrettr:
@heartsucker Just noticed this in the changelog for Flask 0.11, which was released recently (May 29th, 2016):

> Templates are no longer automatically reloaded outside of debug mode. This can be configured with the new TEMPLATES_AUTO_RELOAD config key.

Might this be related to the problems you were having? I don't have time to investigate further, but thought it might be helpful for me to share this with you.

We'll probably update the version of Flask used by SecureDrop fairly soon, so if that fixes the issue for you, it'll be coming down the pipe (at least in develop) shortly.

heartsucker:
I don't think this is related to the problem because (I think) Flask was
already in debug mode when I was encountering this. I'd have to double
check the diff.
Am 05.07.2016 22:52 schrieb "Garrett Robinson" <notifications@github.com>:

PR #1214, opened 2015-12-03
Integrate build and staging playbooks

conorsch:
Rewrites the `install-local-packages` role to be more friendly to first-time developers. Includes error-handling and provides informative messages that link out to the docs, explaining the need to build the OSSEC packages separately.
### Major changes
- Consolidates all package building into `build-debian-package`, customized via vars in wrapper roles:
  - `build-securedrop-app-code`
  - `build-securedrop-ossec-agent`
  - `build-securedrop-ossec-server`
- Chains the building of Debian packages to the staging playbook, so packages are automatically rebuilt on each run of the staging playbook
- Provides an easy method for rebuild-and-reinstall in staging: `ANSIBLE_ARGS="-t rebuild" vagrant provision /staging/`, and also the external repo for OSSEC base packages
### Minor changes
- Removes sundry bash scripts for building packages
- Updates Serverspec tests with new regression checks for building deb packages
- Creates `ossec_version` variable to provide a single point of change for future updates
- Updates developer documentation to include mention of the `rebuild` tag

Also included here are a few changes to the CI scripts, that were convenient to make while updating the spectests. The changes aren't sufficient yet to activate the automatic Snap-CI runs for PRs, but getting quite close. 

Resolves #1040. Nudges forward—but does not resolve—issues #1023, #1067, and #1208.

PR #1332, opened 2016-06-13
Add Apparmor rule to allow Apache to execute uname

### LINKED ISSUE #1331: Spurious OSSEC alerts due to Apache AppArmor profile
When apache2 starts the WSGI application, it tries to exec `uname`, which triggers an AppArmor denial, which in turn triggers an OSSEC alert:

```
OSSEC HIDS Notification.
[timestamp redacted]

Received From: (app) [ip address redacted]->/var/log/syslog
Portion of the log(s):

[timestamp redacted] app kernel: [[redacted]] audit: type=1400 audit([redacted]): apparmor="DENIED" operation="exec" profile="/usr/sbin/apache2" name="/bin/uname" pid=3989 comm="sh" requested_mask="x" denied_mask="x" fsuid=33 ouid=0

--END OF NOTIFICATION
```

It might be worthwhile to investigate the cause of this alert, since I do not believe this was an issue when 0.3 was released. We did a thorough audit to find (and write rules to avoid) spurious OSSEC alerts around that time. My hunch is that apache or some component related to it has been updated since then, and the update changed the behavior in such a way as to cause the AppArmor denial that we are seeing today.

This issue was noticed during QA testing for 0.3.7, but we decided not to fix it then because we wanted to avoid additional scope creep for 0.3.7. It has recently also been reported to us by an admin as a cause of OSSEC alert spam, which indicates to us that a fix should be prioritized for the next point release.

fowlslegs:
At some point Apache started shelling out to the uname tool to gather some info about the system as observed in https://github.com/freedomofpress/securedrop/issues/1331. We've decided it should be fine for Apache to run uname and to add a rule to the AppArmor profile to allow for this to happen. This will fix the OSSEC alerts triggered by the Apparmor denial events. Resolves #1331.

### DIFF
--- a/install_files/ansible-base/usr.sbin.apache2
+++ b/install_files/ansible-base/usr.sbin.apache2
@@ -11,6 +11,7 @@
 
   /bin/dash rix,
   /bin/touch rix,
+  /bin/uname rix,
   /dev/null w,
   /dev/urandom r,
   /etc/apache2/apache2.conf r,


garrettr:
This PR looks great, testing it now.

garrettr:
This PR makes a change to the apparmor profiles, which are installed by the Debian packages. As we know, it is somewhat difficult to test with custom Debian packages in the 0.3.x release series. @conorsch What's the best way to test this, in your opinion? Build temporary packages, store them in `build/`, and run Vagrant with the `install_local_packages` tag enabled?

conorsch:
@garrettr Try temporarily rebasing on top of #1214, which will automatically build deb packages and install them in the staging environment.

garrettr:
@conorsch Unfortunately there are too many merge conflicts with #1314 and #1332 for that approach to be convenient.

garrettr:
Testing procedure:

1. Checkout this PR (with `hub`, `alias git=hub` and `git checkout https://github.com/freedomofpress/securedrop/pull/1332`)
2. Build the new packages: `vagrant provision build`. `securedrop-app-code-$VERSION-$ARCH.deb` will contain the updated AppArmor profile.
3. Make sure you have all the necessary packages for a local package install in your `build/` directory. They are listed in the `local_deb_packages` variables for app and mon in the playbooks. If you need some packages that you don't feel like building (e.g. `ossec-{agent,server}-*.deb`), just download them from `https://apt.freedom.press` and place them in `build/`.
4. Edit `Vagrantfile` and comment the line `ansible.skip_tags =  [ "install_local_pkgs" ]` for the staging environment.
5. New staging environment with locally built packages: `vagrant destroy -f /staging$/ && vagrant up --no-install-provider /staging$/`
6. `vagrant ssh app-staging` and verify that there are no apparmor errors related to apache2 calling uname in `/var/log/syslog`. The errors would have been triggered when the server started up. You can also make a couple of requests to the staging Source Interface (`localhost:8082`) to make sure this doesn't start up some WSGI worker processes which in turn could trigger the error.

With that, I've verified that this PR fixes the reported issue.

PR #1334, opened 2016-06-14
Move wordlist back to original location

### LINKED ISSUE #1328: config.py will break in next release becuase of changed wordlist
Because `config.py` is in the `.gitignore`, during the release where the new wordlist is pushed out, old systems will break because the wordlist has moved. Currently people's dev environment will be "broken" out of the box.

Solutions:
1. Move the wordlist back to its old location
2. Force push the single line change and mention the possibility merge conflicts in the release notes
3. Include this manual step in the next set of release notes

I think Option 2 makes the most sense.

heartsucker:
Fixes #1328


garrettr:
Hm, this is a reasonable approach. I'm a little reluctant to commit a symlink because they can create problems on filesystems that do not support them (e.g. FAT or NTFS on Windows, according to this [Stack Overflow post](http://stackoverflow.com/questions/11662868/what-happens-when-i-clone-a-repository-with-symlinks-on-windows)).

This might not seem like a problem because SecureDrop only runs on Ubuntu 14.04, but I think it *could* be a problem for Windows users who wish to contribute to SecureDrop. For example, if you clone the repo on Windows and run `vagrant up development`, which syncs the repo root under `/vagrant` on the Linux virtual machine, what happens to the `wordlist` symlink?

heartsucker:
I haven't developed on Windows ever, so I didn't think about the symlink issue. I think the most simple thing would be for me to just move this back to the original location. Yes yes?

garrettr:
@heartsucker Yes, I think that would the simplest solution for now.

If you want, you could file a follow-up issue to move it to `dictionaries/`, which I do agree is the best location for it. In order to do that, however, we would need to come up with stories to address updating the configuration for both development and production environments.

heartsucker:
@garrettr Updated. I'll make issues for those.

heartsucker:
Actually, what's the point of having the wordlist configurable? Shouldn't we just hard code that and say "you have to use our wordlist." I think that actually makes more sense now that I'm really thinking about it.

heartsucker:
Just kidding, I answered my own question. Because using absolute paths in the config makes us refactor resistant and helps us avoid saying `open('../../../wordlist')` or whatever.

garrettr:
Thanks for doing this @heartsucker, and for filing the follow-up. 💯 

I agree that this definitely sucks, and is a good example of what we should keep in mind when we eventually refactor SecureDrop's configuration management.

PR #1336, opened 2016-06-14
Fix version in 0.3.7 docs

### LINKED ISSUE #1325: update_version.sh should update new documentation

garrettr:
Resolves #1325 for 0.3.7.post1. Subsumes #1326.


garrettr:
I did a quick `ag 0.3.6 docs/` and didn't see anything other instances of the version number that seemed like they needed to be bumped to 0.3.7.

fowlslegs:
Ran `rgrep '0\.3\.[0-6]' docs` and found `docs/conf.py` still has some 0.3.5 lines that should be auto-updated as well. As an aside, I stopped using `ag` because I found that it behaves differently in grep and sometimes doesn't find things as I expect--not sure exactly what the syntactically differences are, but I guess something that could be looked into (obviously not related here due to the different arguments).

garrettr:
Nice catch on the references to the version number in `conf.py`, @fowlslegs. Fixed in the last commit.

PR #1338, opened 2016-06-15
Update wording for deployment best practices

### LINKED ISSUE #1337: Update Deployment Best Practices Recommendation
Should make the Washington Post "Whole Site" recommendation more generalized.

https://github.com/freedomofpress/securedrop/blob/develop/docs/deployment_practices.rst#whole-site-changes

The example of using `ssl.washingtonpost.com` is no longer applicable, since that domain just redirects to www now, which is being served entirely over HTTPS.

jmhale:
Updates a line item under the Whole Site section for Deployment Best Practices.

https://github.com/freedomofpress/securedrop/blob/develop/docs/deployment_practices.rst#whole-site-changes

The Washington Post example no longer applies, since `ssl.washingtonpost.com` just redirects to `www` now, which is served entirely by HTTPS.

Fixes #1337

### DIFF
--- a/docs/deployment_practices.rst
+++ b/docs/deployment_practices.rst
@@ -306,7 +306,7 @@ analysis.
 
 #. Make your entire site available through HTTPS.
 
-   - That way, visits to your landing page won't stand out as the only encrypted traffic.
+   - That way, visits to your landing page won't stand out as the only encrypted traffic to your site.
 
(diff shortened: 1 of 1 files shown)


garrettr:
Nice change, thanks @jmhale! I can't recall why it has that weirdly specific reference to `ssl.washingtonpost.com`. Also, the recommendation is much clearer now.

PR #1342, opened 2016-06-19
Fixes some indentation issues in html/css

mdrose:
Noticed a couple tabs in the html templates and css. I went ahead and replaced all the tabs with spaces and also fixed some nearby indentation inconsistencies.


garrettr:
Thanks! I <3 consistency (and spaces).

Issue #1340, opened 2016-06-16
Update SecureDrop deployment and configuration process for admins


heartsucker:
As mentioned in a few places (at least once [here](https://github.com/freedomofpress/securedrop/pull/1334#issuecomment-226544917)), the current deployment of Ansible is rather delicate.

Some goals of this should be to:
- Reduce the likelihood that an admin overwrites their `config.py` (or values for that wherever they are loaded)
- Have all files be kept in the git repo with a config be read from somewhere else on disk
- Have some sort of configuration validation (to avoid things like #1257)

One solution would be a simple CLI tool that is both scriptable and interactive that reads a config from elsewhere and validates it, or generates one.
### For example

```
anmnesia@amnesia:~/Persistent/securedrop/$ ./config-tool load-config ../securedrop-configs/config.yml
Loading old config from /home/amnesia/Persistent/securedrop-configs/config.yml
Validating...
WARN - config has extraneous key 'foo' with value 'bar'
ERROR - config has an invalid GPG key for key 'admin-key-fpr' with value '0x123abc'
```

```
anmnesia@amnesia:~/Persistent/securedrop/$ ./config-tool new-config --output ../securedrop-configs/config.yml --foo bar --baz quux
Validating...
ERROR - Missing key 'wat'
```

This `.yml` would still be used the way the current `.yml`: to power ansible deploys where all the moving around of details is still done within ansible as is some validation. This would just provide a nice user interface to help admins get things right on the first try.

heartsucker:
Just leaving this as a note while looking for an unrelated issue:

This also affects issues #749 and #924.

conorsch:
There are a couple of viable approaches here. There's already a small amount of variable validation performed in the `validate` role, which runs automatically as part of the prod playbook (but not staging, since it checks for staging/dev vars and fails if found), so I'm inclined to continue work in that direction.

Currently, the validate role uses `fail` tasks with conditional logic via `when` statements to bail out on a bad config. It'd be more intuitive to switch to using Ansible's [assert module](http://docs.ansible.com/ansible/assert_module.html). Should we need more sophisticated analysis of variable values than what's provided by Ansible's Jinja2 filters, we can add custom filter_plugins to the validate role, which gives us the power of raw Python.

conorsch:
I also see this as related to #1129. Any work done to aid in bootstrapping the initial config should consider implications on general configurability for the deployment.

redshiftzero:
[...]

PR #1347, opened 2016-06-26
more unix-y cli for manage.py

heartsucker:
First step toward #1340

### DIFF
--- a/securedrop/manage.py
+++ b/securedrop/manage.py
@@ -1,22 +1,16 @@
 #!/usr/bin/env python
 
-import atexit
 import sys
 import os
-import select
 import shutil
 import subprocess
-import unittest
-import readline  # makes the add_admin prompt kick ass
-from getpass import getpass
 import signal
-from time import sleep
-
 import qrcode
 import psutil
 
+from getpass import getpass
+from argparse import ArgumentParser
 from db import db_session, Journalist
-
 from management import run
 
 # We need to import config in each function because we're running the tests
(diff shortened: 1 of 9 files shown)


conorsch:
Good changes, I agree that using hyphens is a more intuitive approach to subcommands. Great job updating the docs, as well. Just grepped around and found a lingering mention of the old `add_admin` style, in `docs/upgrade/0.2.1-to-0.3.x.rst`. That should be updated to the new style as well, then I'm happy with merging.

heartsucker:
@conorsch Updated.

securedrop/manage.py
```diff
@@ -213,27 +207,40 @@ def listdir_fullpath(d):
...
-        "reset",
-        "add_admin",
-        "clean_tmp"]
-    help_str = "./manage.py {{{0}}}".format(','.join(valid_cmds))
+def get_args():
+    parser = ArgumentParser(prog=__file__,
+                            description='A tool to help admins manage and devs hack')
+
+    subparsers = parser.add_subparsers()
+    subparsers.required = True
```
fowlslegs: In my tests a singular subcommand was required even if this line was ommitted.

securedrop/manage.py
```diff
@@ -213,27 +207,40 @@ def listdir_fullpath(d):
...
+    subparsers.dest = 'subcommand'
```
fowlslegs: The function name is already saved to the `parser.func` attribute which we call with `args.func()`. This writes the title argument of the relevant `add_parser` subcommand to `parser.subcommand`, which is unused/ unecessary. Since this line of code seems to do nothing, I suggest removing it.

fowlslegs:
Closed in favor of https://github.com/freedomofpress/securedrop/pull/1368.

fowlslegs:
Thanks @heartsucker :smile_cat:

PR #1348, opened 2016-06-29
Config tool for admins

heartsucker:
More work towards #1340.

Not ready for merge (incomplete). This PR is for a review of functionality. Namely:
1. Do we like the interface?
2. Do we like the idea of the master config that stores all the values?

Second, for continued work, I was planning on extending this to:
1. Strip the config to the minimal values and place them where they need to go. i.e., put all the `ansible` related stuff under `install_files/ansible-base/prod-specific.yml`
2. Skip 1, and rewrite ansible to look outside the `git` repo for configs

### DIFF
--- a/securedrop/config_tool.py
+++ b/securedrop/config_tool.py
@@ -1,0 +1,327 @@
+#!/usr/bin/env python
+
+import argparse
+import logging
+import re
+import sys
+import yaml
+
+from os import path
+from yaml.scanner import ScannerError
+
+
+"""
+A CLI tool to help admins configure their SecureDrop instances. Uses argparse, so run this file from
+your terminal with --help to get instructions on how to use this.
+"""
+
+
+def init_logger():
+    l = logging.getLogger(__file__)
+    l.setLevel(logging.DEBUG)
+    ch = logging.StreamHandler(sys.stdout)
+    ch.setLevel(logging.DEBUG)
+    ch.setFormatter(logging.Formatter('%(levelname)s - %(message)s'))
+    l.addHandler(ch)
+    return l
+
+log = init_logger()
+
+
+class ConfigToolException(Exception):
+
+    def __init__(self, msg=None):
+        super(ConfigToolException, self).__init__(msg)
+
+
+def check_ip_address_valid(ip):  # "good enough"
+    m = re.match(r"^(\d{1,3})\.(\d{1,3})\.(\d{1,3})\.(\d{1,3})$", ip)
+    if bool(m) and all(map(lambda n: 0 <= int(n) <= 255, m.groups())):
+        log.debug("'{ip}' was a valid ipv4 address".format(ip=ip))
+        return True
+    else:
+        log.error("'{ip}' was not a valid ipv4 address".format(ip=ip))
+        return False
+
+
+def check_ssh_users(users):
+    def check_user(user):
+        regex = r'^[a-z][-a-z0-9]*$'
+        if isinstance(user, str) and re.match(regex, user):
+            log.debug("'{user}' was a valid Unix user name".format(user=user))
+            return True
+        else:
+            log.debug("'{user}' was not valid Unix user name. Regex: {regex}".format(user=user, regex=regex))
+            return False
+
+    if isinstance(users, str):
+        return check_user(users)
+    elif isinstance(users, list):
+        return all(map(check_user, users))
+    else:
+        log.error('SSH users need to be a string or list of strings')
+        return False
+
+
+def check_dns_servers(servers):
+    if isinstance(servers, str):
+        return check_ip_address_valid(servers)
+    elif isinstance(servers, list):
+        return all(map(check_ip_address_valid, servers))
+    else:
+        log.error('DNS servers need to be a string or list of strings')
+        return False
+
+
+def check_gpg_public_key(key_str):
+    key_lines = key_str.splitlines()
+    return len(key_lines) >= 2 and \
+        key_lines[0] == '-----BEGIN PGP PUBLIC KEY BLOCK-----' in key_str and \
+        (key_lines[-1] == '-----END PGP PUBLIC KEY BLOCK-----' or
+         key_lines[-2] == '-----END PGP PUBLIC KEY BLOCK-----')
+
+
+def check_gpg_public_key_fingerprint(fpr):
+    return bool(re.match(r'^[0-9A-F]{40}$', fpr))
+
+
+def check_email_address(addr):
+    return '@' in addr
+
+
+"""
+This object represents what a valid config should look like and contains test functions to validate
+the data in configs we generate.
+"""
+MASTER_CONFIG = {
+    'app': {
+        'ip_address': check_ip_address_valid,
+        'hostname': lambda x: True,  # no hostname validation
+        'secure_drop': {
+            'header_image': lambda x: True if not x else path.exists(x),
+            'gpg_public_key': check_gpg_public_key,
+            'gpg_public_key_fingerprint': check_gpg_public_key_fingerprint,
+        }
+    },
+    'mon': {
+        'ip_address': check_ip_address_valid,
+        'hostname': lambda x: True,  # no hostname validation
+        'ossec': {
+            'gpg_public_key': check_gpg_public_key,
+            'gpg_public_key_fingerprint': check_gpg_public_key_fingerprint,
+            'email_address': check_email_address,
+            'smtp': {
+                'relay': lambda x: isinstance(x, str) and bool(x),
+                'port': lambda x: isinstance(x, int) and 0 < x < 65536,
+                'sasl': {
+                    'username': lambda x: isinstance(x, str) and bool(x),
+                    'password': lambda x: isinstance(x, str) and bool(x),
+                    'domain': lambda x: isinstance(x, str) and bool(x),
+                }
+
+            }
+        }
+    },
+    'provisioning': {
+        'ssh_users': check_ssh_users,
+        'dns_servers': check_dns_servers
+    },
+}
+
+
+def load_file(path):
+    """
+    Loads a config as YAML from a given path
+    :param path: Path to the file to load
+    :return: A dict if successful, raises an Exception otherwise
+    """
+
+    log.debug("Loading file from path '{path}'".format(path=path))
+    f = None
+
+    try:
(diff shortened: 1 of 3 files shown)


securedrop/config_tool.py
```diff
@@ -0,0 +1,327 @@
...
+log = init_logger()
+
+
+class ConfigToolException(Exception):
+
+    def __init__(self, msg=None):
+        super(ConfigToolException, self).__init__(msg)
+
+
+def check_ip_address_valid(ip):  # "good enough"
```
jesusaurus: If we `import ipaddress` (which requires a dependency on py2-ipaddress for python2) then we can use `ipaddress.ip_address(ip)` and check for a `ValueError`, and that would give us free ipv6 support for this check too.

securedrop/config_tool.py
```diff
@@ -0,0 +1,327 @@
...
+        (key_lines[-1] == '-----END PGP PUBLIC KEY BLOCK-----' or
+         key_lines[-2] == '-----END PGP PUBLIC KEY BLOCK-----')
+
+
+def check_gpg_public_key_fingerprint(fpr):
+    return bool(re.match(r'^[0-9A-F]{40}$', fpr))
+
+
+def check_email_address(addr):
+    return '@' in addr
```
jesusaurus: It might be slightly better to split the string on '@' and check that both sides are not `None` (and possibly also check that there is a '.' on the right side). But no matter what, this function is going to be sad.

securedrop/config_tool.py
heartsucker: SecureDrop only uses IPv4 for now, so this works fine. I did this because it doesn't introduce a dependency.

securedrop/config_tool.py
heartsucker: [This](https://stackoverflow.com/questions/201323/using-a-regular-expression-to-validate-an-email-address#201378) is why it's not worth dong actual email address validation. It's very complicated, and this tool is just meant to keep admins from making dumb mistakes.

conorsch:
Thank you for your input here, @jesusaurus. I agree with @heartsucker's comments about IPv4 being required, so we're safe leaving out IPv6. (As much as I'm loathe to pretend like IPv6 conversion isn't happening, it's currently well outside the scope of the problem SecureDrop is trying to address.)

Regarding the intent of these changes, as I mentioned briefly in #1340, I believe we'd gain a lot by using validators in Ansible to verify that the prod vars are structured in a sane way. As @heartsucker mentioned, the goal here is to prevent silly mistakes. There are several examples of common errors that first-time Admins make, and the docs don't quite make quite clear enough, e.g. should one manually remove spaces from fingerprints before adding to the vars file? That's something machine can, and should, check before running the playbooks. That's related to #1146, however, since we'd want to make sure we're keeping abreast of built-in filters with Ansible, so we don't have to reimplement and maintain the filters ourselves if we can avoid it.

We're currently wrapping up backporting the recent releases onto the develop branch, so I'll take a look at the work that's been done in develop and how it compares to the changes @heartsucker is proposing. I expect to start that work by next week.

securedrop/config_tool.py
```diff
@@ -0,0 +1,327 @@
...
+        log.debug("'{ip}' was a valid ipv4 address".format(ip=ip))
+        return True
+    else:
+        log.error("'{ip}' was not a valid ipv4 address".format(ip=ip))
+        return False
+
+
+def check_ssh_users(users):
+    def check_user(user):
+        regex = r'^[a-z][-a-z0-9]*$'
```
fowlslegs: I think this should be `r'^[a-zA-Z_][a-zA-Z0-9_-]*$`.

securedrop/config_tool.py
```diff
@@ -0,0 +1,327 @@
+#!/usr/bin/env python
+
```
fowlslegs: Should add `# -*- coding: utf-8 -*-` to be consistent with rest of our Python source.

securedrop/config_tool.py
```diff
@@ -0,0 +1,327 @@
...
+class ConfigToolException(Exception):
+
+    def __init__(self, msg=None):
+        super(ConfigToolException, self).__init__(msg)
+
+
+def check_ip_address_valid(ip):  # "good enough"
+    m = re.match(r"^(\d{1,3})\.(\d{1,3})\.(\d{1,3})\.(\d{1,3})$", ip)
+    if bool(m) and all(map(lambda n: 0 <= int(n) <= 255, m.groups())):
```
fowlslegs: You don't have to use `bool` here. `""`, `0`, `set()`, `None` and the trivial instance/ zero object of most of the builtin types is equivalent to `False` when `if` testing a condition.

securedrop/config_tool.py
```diff
@@ -0,0 +1,327 @@
...
+    if isinstance(servers, str):
+        return check_ip_address_valid(servers)
+    elif isinstance(servers, list):
+        return all(map(check_ip_address_valid, servers))
+    else:
+        log.error('DNS servers need to be a string or list of strings')
+        return False
+
+
+def check_gpg_public_key(key_str):
```
fowlslegs: Importing it to an ephemeral keyring with `gnupg` would be a better test.

securedrop/config_tool.py
```diff
@@ -0,0 +1,327 @@
...
+    key_lines = key_str.splitlines()
+    return len(key_lines) >= 2 and \
+        key_lines[0] == '-----BEGIN PGP PUBLIC KEY BLOCK-----' in key_str and \
+        (key_lines[-1] == '-----END PGP PUBLIC KEY BLOCK-----' or
+         key_lines[-2] == '-----END PGP PUBLIC KEY BLOCK-----')
+
+
+def check_gpg_public_key_fingerprint(fpr):
```
fowlslegs: It would be good to verify with `gnupg` that the fingerprint matches the public key. @conorsch and I have discussed before though that one should not have to enter the fingerprint, and it should just be derived from the public key provided automatically w/o user input when it is needed.

securedrop/config_tool.py
fowlslegs: While not suggesting you do this 100% correct (if such a thing is possible in this case :smirk_cat:), it's probably worth it to be a little more strict. I think @jesusaurus has the right idea here that would help catch, e.g., a `,` typed instead of a `.`.

securedrop/config_tool.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+"""
+This object represents what a valid config should look like and contains test functions to validate
+the data in configs we generate.
+"""
+MASTER_CONFIG = {
+    'app': {
+        'ip_address': check_ip_address_valid,
+        'hostname': lambda x: True,  # no hostname validation
```
fowlslegs: Is there a reason the `check_ssh_users()` function wouldn't work here?

securedrop/config_tool.py
```diff
@@ -0,0 +1,327 @@
...
+        'secure_drop': {
```
fowlslegs: Nit: write as `securedrop` since it's one word.

securedrop/config_tool.py
```diff
@@ -0,0 +1,327 @@
...
+            'header_image': lambda x: True if not x else path.exists(x),
```
fowlslegs: Try `lambda:x True if not x else bool(imghdr.what(x))`. `imghdr` is a built-in library.

securedrop/config_tool.py
```diff
@@ -0,0 +1,327 @@
...
+            'gpg_public_key': check_gpg_public_key,
+            'gpg_public_key_fingerprint': check_gpg_public_key_fingerprint,
+        }
+    },
+    'mon': {
+        'ip_address': check_ip_address_valid,
+        'hostname': lambda x: True,  # no hostname validation
```
fowlslegs: Again, see if you can reuse the SSH username validator for this.

securedrop/config_tool.py
```diff
@@ -0,0 +1,327 @@
...
+    """
+    Loads a config as YAML from a given path
+    :param path: Path to the file to load
+    :return: A dict if successful, raises an Exception otherwise
+    """
+
+    log.debug("Loading file from path '{path}'".format(path=path))
+    f = None
+
+    try:
```
fowlslegs: `with` will do context management for you, so you don't have to worry about the error-handling you do with this `try`,  `except`, `finally` block here.

fowlslegs:
I didn't see @conorsch's comment here before I began the review, but I'm going to stop now because I agree that we should be preferring Ansible's built-in validators. If those are not sufficient, we may want to write an additional Ansible module that could include some of this code, but we'll wait for @conorsch to make a determination on that.

fowlslegs:
Also, @heartsucker see http://nvie.com/posts/better-package-management/ and `pip_update.sh` to learn about how we now handle Python dependencies. One should no longer manually edit the `*requirements.txt` files anymore.

securedrop/config_tool.py
heartsucker: Fixed.

securedrop/config_tool.py
heartsucker: Fixed.

securedrop/config_tool.py
heartsucker: Fixed.

securedrop/config_tool.py
heartsucker: @fowlslegs Github's line comment is ambiguous here. Can you clarify?

securedrop/config_tool.py
heartsucker: Fixed.

securedrop/config_tool.py
heartsucker: @fowlslegs I wanted the error handling done manually so we could print out slightly more useful error messages for the admins. I don't think there's a way to do with `with`?

securedrop/config_tool.py
heartsucker: Fixed.

conorsch:
@heartsucker I'm going to close this, since it uses a custom script instead of `assert` tasks in the existing `validate` role. Leaving #1340 open, and tracking validation updates primarily in #749, targeting 0.4 to land most of the changes. You identified some fine testing targets in this PR, some of which may get added to the automated config testing described in #1580.

PR #1349, opened 2016-07-02
Converted CSS to SASS

heartsucker:
I wanted to a add new features and CSS and is terrible to work with compared to SASS. This PR includes:
- Convert all CSS to SASS
- Combine duplicate SASS rules
- Add SASS compilation to Ansible / builds


garrettr:
Thanks for this PR, @heartsucker! I've been wanting to do this for a long time, since I agree CSS is terrible to work with compared to alternatives such as SASS. I'll review this by the end of the week.

heartsucker:
Hey guys. This is in need of some rebasing, but before I get to it, can someone take a look and let me know if this seems to be a good candidate for merge?

securedrop/sass/source.sass
```diff
@@ -0,0 +1,206 @@
...
+    padding: 10px
+    font-weight: bold
+
+p#codename
+    font-weight: bold
+    /*
+     * HACK: Firefox likes to add non-existent leading and/or trailing whitespace
+     * to <span>s, but never does it to <p>s. This hack uses a <p> for the
+     * codename so we can get the right copy behavior, and uses inline-block plus
+     * tweaks the margin to make it look like a span
```
garrettr: Close this comment.

securedrop/sass/source.sass
```diff
@@ -0,0 +1,206 @@
...
+
+    .confirm-prompt
+        position: absolute
+        width: 300px
+        right: 5%
+        top: -9%
+        font-size: 14px
+
+        p
+            *
```
garrettr: Is this line necessary?

securedrop/source_templates/base.html
```diff
@@ -3,8 +3,9 @@
   <head>
     <meta charset="utf-8">
     <title>SecureDrop | Protecting Journalists and Sources</title>
-    {% assets filters="cssmin", output="gen/source.css",
-      "css/normalize.css", "css/source.css", "css/font-awesome.css" %}
+
+    {# TODO why doesn't a link to /static/source.css work? #}
```
garrettr: We should resolve this TODO before merging. It's best not to leave TODOs in merged code, if possible.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/build_securedrop_app_code_deb.yml
```diff
@@ -48,6 +57,15 @@
   tags:
     - build
 
+- name: Compile sass to css
+  shell: sass --update sass:static/css --style compressed
+  args:
+    chdir: /tmp/{{ securedrop_app_code_deb }}/var/www/securedrop/
+#  register: sass_compile_result
+#  changed_when: sass_compile_result.stdout != ''
```
garrettr: This code is commented out, can it be removed?

securedrop/management/run.py
```diff
@@ -151,12 +154,15 @@ def run():
...
-                         'cyan'),
+        lambda: DevServerProcess('Source Interface',
+                                 ['python', 'source.py'],
+                                 'blue'),
+        lambda: DevServerProcess('Document Interface',
+                                 ['python', 'journalist.py'],
+                                 'cyan'),
+        lambda: DevServerProcess('SASS Compiler',
+                                 ['sass', '--watch', 'sass:static/css'],
+                                 'magenta'),
```
heartsucker: If one of the subprocesses fails to start or throws and exception, the monitor will leave the other processes dangling. Each call to DevServerProccess needs a try/except around it.

securedrop/sass/source.sass
heartsucker: Probably not? I'm not a CSS expert, so I made as few changes as possible when porting it to SASS.

securedrop/source_templates/base.html
heartsucker: I'll take a look at this since there have been some updates to some of the dependencies since I wrote this and the behavior might have changed. Also, suggestions welcome since this one seems a bit non-intuitive.

heartsucker:
@garrettr I'm not really a web frontend guy, but more a guy who can do some frontend stuff when he needs. I happened to do it in SASS because that's what I know, but some poking around does seem to indicate that there are more advantages to SCSS. If y'all want that, I can make those change along with all of the rest of this.

securedrop/sass/source.sass
garrettr: @heartsucker Gotcha. Let's leave it in for now, it can be addressed in a follow-up if necessary.

securedrop/management/run.py
garrettr: @heartsucker Thanks for the explanation! I'll take a look at this myself - I think there might be a way to more gracefully handle that case within DevServerProcess.

fowlslegs:
> I'm not really a web frontend guy

Me either.

> poking around does seem to indicate that there are more advantages to SCSS.

From a little research, I saw a decent amount of opinion to that point as well. E.g., http://thesassway.com/editorial/sass-vs-scss-which-syntax-is-better.

>  If y'all want that, I can make those change along with all of the rest of this.

If that isn't too much work for you, it seems to me like that might be good, just because it seems like SCSS might be better for other contributors to work with in the future. Since I don't do much myself with the frontend codebase myself, I'm not going to make an absolute call one way or the other. Unless one of the other SD team members has strong opinions, we'll leave the decision up to you.

garrettr:
> I happened to do it in SASS because that's what I know, but some poking around does seem to indicate that there are more advantages to SCSS. If y'all want that, I can make those change along with all of the rest of this.

@heartsucker I think SASS is fine. I polled some front-end folks that I know and people seem to be pretty evenly split on which one is "better". If we end up deciding to change to SCSS later, that seems to be trivial to do via `sass-convert`.

heartsucker:
This was tested on the dev VM, but it's not passing Travis because of this mystery:

```
TASK: [development | Install sass Ruby gem] *********************************** 
<localhost> state=latest name=sass
changed: [localhost] => {"changed": true, "name": "sass", "state": "latest", "version": "3.4.22"}

TASK: [development | Compile sass to css] ************************************* 
<localhost> chdir=/home/travis/build/freedomofpress/securedrop/securedrop
failed: [localhost] => {"changed": true, "cmd": "sass --update sass:static/css", "delta": "0:00:00.004013", "end": "2016-12-03 15:14:05.668050", "rc": 127, "start": "2016-12-03 15:14:05.664037", "warnings": []}
stderr: /bin/sh: 1: sass: not found
```

It's saying the gem was installed (globally), but then it's not on the path. Anyone have any experience debugging the Travis environment? Also, I added verbosity to build in hopes that it might tell us something, but alas.

redshiftzero:
Hey @heartsucker, I took a peek at this PR yesterday and it does look to be a weird PATH issue: one way to troubleshoot is to add a couple of post-install tasks in that Ansible play before you try to execute `sass`, e.g.:

```
- name: Check sass is in gem list
  command: gem list

- name: Check where gems are being installed
  command: gem environment
```

Since you put those -vv verbosity flags (good call) you can inspect the output and see what's going on with the PATH. This might not be the Best Way to debug this but it’s what I do  ¯\_(ツ)_/¯. Give that a whirl, if things still look confusing let me know and I’ll take a deeper look.

heartsucker:
@redshiftzero I already did most of that (and them some), but even when putting calls `sass` with the full path `/usr/lib/gems/sass` (or whatever it was), it still fails. I've never dealt with Ruby, so whatever seems to be going on with RVM is new to me. Anywho. I can probably get to this this weekend.

heartsucker:
So I've spent a couple of hours on this, and all I have so show is a new error message:

```
/usr/bin/env: ruby_executable_hooks1.9.1: No such file or directory
```

Any rubyists have any idea what do to here?

heartsucker:
Ok, so in commit b9c78be I did just about the kludgiest thing I've done in a while. It works, but it makes the build take over a minute longer than it should.

This whole bit needs to happen because the SASS files need to be compiled to CSS since some of the functional tests require DOM elements to be visible.

heartsucker:
Pinging for review/merge.

heartsucker:
Doubling down on the request for merge because there are a bunch of UI tickets that got opened today, and it would be nice to do this first.

heartsucker:
@garrettr? @fowlslegs? @redshiftzero? Bueller?

conorsch:
Whoa, sorry about the lag on this one, @heartsucker! Thanks for your persistence, getting this in pronto. (Yes, the hack in b9c78be is indeed adventurous approach, but it works well enough for the dev env!)

PR #1358, opened 2016-07-06
Adds unread submission icon and select all unread button to source view

### LINKED ISSUE #1353: Indicate unread submissions in source view
In the "Sources" view of the Document Interface, we indicate the number of unread submissions for each source, and provide a handy button to download _just_ the unread submissions as an archive.

screen shot 2016-07-05 at 5 51 04 pm

However, in the individual "Source" view, we _do not_ indicate which submissions are unread.

screen shot 2016-07-05 at 5 51 16 pm

This makes it difficult to figure out which submissions are unread in the Source view, particularly for sources with many submissions. It should be easy to fix this by adding an icon to the row for each submission in the Source view.

We might also consider adding a "Select all unread" button to the Source view, or another "Download unread" button on the source view (to save journalists from having to hit the Back button and potentially scroll through a long list of sources when they realize they want to download all unread after already having gone to a Source page).

mdrose:
Resolves #1353. Only adds space for an icon if there's at least one unread submission. If all submissions have been read, then the page appears as it would before this PR. Not sure if it's better this way or if it's better to just have an empty space regardless. I was planning on using an open envelope icon for read submissions, making it more clear what the icon means, but, surprisingly, font awesome doesn't have one.

Planning to make the "select all unread" button as well. Also considering adding javascript that removes the icon when the journalist clicks the download button.

Screenshot:
unread

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -438,7 +438,8 @@ def index():
 def col(sid):
     source = get_source(sid)
     source.has_key = crypto_util.getkey(sid)
-    return render_template("col.html", sid=sid, source=source)
+    unread_subs = [submission for submission in source.submissions if not submission.downloaded]
+    return render_template("col.html", sid=sid, source=source, unread_subs=unread_subs)
 
 
 def delete_collection(source_id):
--- a/securedrop/journalist_templates/col.html
+++ b/securedrop/journalist_templates/col.html
@@ -28,7 +28,15 @@
       </div>
       <ul id="submissions" class="plain submissions">
         {% for doc in source.collection %}
-          <li class="submission"><input type="checkbox" name="doc_names_selected" value="{{ doc.filename }}" class="doc-check"/>
+          <li class="submission">
+            <input type="checkbox" name="doc_names_selected" value="{{ doc.filename }}" class="doc-check"/>
+            {% if unread_subs %}
+              {% if doc in unread_subs %}
+                <span title="Unread" class="unread"><i class="fa fa-envelope"></i></span>
+              {% else %}
+                <span class="unread"></span>
+              {% endif %}
+            {% endif %}
             {% if doc.filename.endswith('reply.gpg') %}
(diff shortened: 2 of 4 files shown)


mdrose:
Added the select unread button. It adds no new jquery, so it's compatible with the direction of #1298. It does add to the existing jquery function that adds in the buttons, however, but merging that should be pretty trivial.

select_unread

conorsch:
Excellent! Thanks very much for this, @mdrose. The new UI is a big improvement, and implements feature in high demand. We do have a conflict with #1298, however. I'm inclined to merge this usability enhancement, and rebase #1298 on top of it if we decide to go forward with excising jQuery entirely.

> I was planning on using an open envelope icon for read submissions, making it more clear what the icon means, but, surprisingly, font awesome doesn't have one.

A surprise indeed. While it's a bit counterintuitive that "read" shows no icon:

securedrop-unread-icons

your approach of omitting an icon seems the best strategy for now. Besides, adding the unread notifications is a big enough win that I don't think we need to stress over a corresponding "read" icon.

redshiftzero:
This is a really great contribution! I just checked this out and tested the following functionality:

* Verified that the documents displayed as unread are indeed unread
* Downloaded a document, verified that the unread icon was no longer displayed
* Used the “Select unread” button to download only the unread documents

Everything looked good to me! 

Going forward, we’ll definitely want to have some tests for this new feature. It would be good to break out the `unread_subs = [submission for submission in source.submissions if not submission.downloaded]` line into a separate function perhaps `get_unread_submissions(source)` such that unit tests can be written for that piece. Let me know if you’d be interested in working on that! If not, no worries - we can merge in your contribution and make that minor tweak.

redshiftzero:
Hey @mdrose I pulled your commits into PR #1443 that supersedes this one with unit testing, so I’m closing this PR in favor of #1443. Thanks for your great work on SecureDrop!

mdrose:
Sorry I didn't get back to you, @redshiftzero. I hadn't checked github in awhile. Thanks for writing unit tests for this.

Issue #1355, opened 2016-07-06
Assign journalists to specific sources


garrettr:
It would be useful to be able to assign a specific source to a given journalist, so that journalist (and the other journalists) know who is handling them.

In the ORM, it's trivial to create a foreign key between `Source` and `Journalist`. It's also easy to display the assigned Journalist in the Sources and Source views on the Document Interface.

Since journalists may be assigned, then re-assigned, it might be useful to track and display the history of assignments in the Source view (similar to how GitHub displays assignees to issues, for example).

In the simplest implementation of this feature, any Document Interface user would be able to assign (or re-assign) a journalist to a source. It would be more complicated, but do-able, to base this capability on some kind of permissions system (e.g. only "admins" or "editors" can assign/re-assign sources or journalists). I am not sure if permissions for assignment would be useful, so I am inclined to implement the simplest version of this first, and implement/track permissions in a follow-up issue if that feature is requested.

It's an open question whether it would be useful to let Sources know who is assigned to them. My inclination is to say that this would not be useful, and could potentially be used to target or harm specific journalists, so I would prefer not to implement this unless somebody can make a strong case for doing so.

fowlslegs:
> It's an open question whether it would be useful to let Sources know who is assigned to them. My inclination is to say that this would not be useful, and could potentially be used to target or harm specific journalists, so I would prefer not to implement this unless somebody can make a strong case for doing so.

This should certainly not be the default. Hesistant to even add it as an optional setting. A journalist can always choose to reveal their identity in a message to the source.

heartsucker:
A source who frequently checks also might see themselves repeatedly
reassigned which could be off putting or might give the impression of a
lack of organization among the news org. Not to mention if they know this
feature exists and see none, this will know they are unassigned which could
be even worse.

I agree that sources should not know what journalist they are assigned to.
Am 11.07.2016 8:59 nachm. schrieb "Noah Vesely" <notifications@github.com>:

fowlslegs:
[...]

Issue #464, opened 2014-06-22
In journalist interface, create comprehensive filters when JS is enabled


Hainish:
We currently allow javascript use in the journalist interface, while gracefully degrading when js is disabled.  For journalists that have JS enabled, we should create a comprehensive filter bar above the list of sources.  This expands out the filter box we have currently for source names.  We can add extra criteria: {starred, unstarred}, {has unread docs, doesn't}.

This is also useful for the list of submissions.  We can filter by {read, unread}, {starred, unstarred}, {msgs, docs}

redshiftzero:
This was implemented back in #465, so closing.

redshiftzero:
Whoops, just kidding. Looks like it didn't get merged! Reopening this and tagging with Reading Room

eloquence:
We discussed this issue in our old issue review yesterday, and I also brought it up with @huertanix from the newsroom team today. We don't see that this part of the UI currently needs work, but of course we'll continue to look for usability and workflow improvements. Closing this issue for now.

dachary:
@Hainish feel free to re-open at any time if you still think this important ;-)

Issue #92, opened 2013-10-22
Prototype encrypting data client-side with the system's public key


dtauerbach:
Right now as I understand it, the source uploads a sensitive document, that document is sent over Tor to the hidden service running on the source server, that source server encrypts the document, and it is only decrypted on the SVS. This means that if the source server is somehow compromised, an attacker could recover the plaintext of the document before it is encrypted.

Channeling some of the feedback from Patrick Ball at the techno-activism event tonight, it might make sense to instead encrypt on the client with the public key of the system. That way, if the source server is compromised, the data will still be protected, so long as the SVS is secure. Since the SVS has a higher security model than the source server.

The way that was suggested to accomplish this is via a browser extension, or baking keys into the browser. In addition to being a lot of work, this brings up the whole can of worms that comes with key distribution (e.g. does the browser extension/patch server as a CA?)

In the shorter term, one could just provide the public key with Javascript, and encrypt the document using it before sending it to the source server. There are two issues I see with this: first, adding Javascript may open up an attack vector if no Javascript is being used right now. Second, the attacker we've presumed to have control of the source server could modify the Javascript to include a different public key. The second problem I think is solvable with a super basic browser add-on or something that detects when a client sees unexpected Javascript. Not all clients have to run this. Given the attacker does not who has submitted documents, she must attack everyone to attack her target. That means even if a small percentage of people run the testing add-on, it will still make an effective attack (against everyone) detectable.

[There should be a separate bug for if and how to move the conversation with the journalist to use a somewhat similar client-side approach.]

redshiftzero:
Yesterday at FOSDEM I had a chat with @tasn, who has written a [nice browser extension](https://github.com/tasn/webext-signed-pages) that verifies the PGP signature of web pages. This is something that is worth experimenting with in the context of SecureDrop. In SecureDrop releases, we'd ship JavaScript that encrypts client-side to the instance's public key, and this js code would be signed with the SecureDrop release key. The browser extension would verify the sig and only execute the JavaScript if the signature verifies (we'd either need to TOFU the release key, or we'd need the release key baked into the extension). We'd fall back to server-side crypto for sources that have JavaScript turned off entirely. We'd also (eventually) need to get this browser extension bundled into Tor Browser. 

[...]

PR #1359, opened 2016-07-12
Added source-assignment functionality

mdrose:
Spent a little time on #1355. This implements a simple version of journalist assignment. 

This links a journalist to sources in the database, displays the current value in index.html and gives a reassign link that opens up a dropdown to pick a new journalist. It probably needs some UX improvement to be merged, though. 

Using a select dropdown might work if you have a small number of journalists, but if you have a lot, it's probably cumbersome. Maybe a combo box is the way to go? I also thought of making some sort of journalist directory that opens up, but would that be overkill? Not sure how many journalists the average user has. It might also be a good idea to give the option to add names for journalist users. Going only by username could make it less than clear who is being assigned to a source.

This is also implemented without javascript. I'm not really sure what the guidelines are for javascript on document interface. The javascript here seems very minimal. Is this by design? This PR would probably be better with some javascript.

Let me know if you guys have any suggestions.

assignment1

Journalist assignment:
assignment2

### DIFF
--- a/securedrop/db.py
+++ b/securedrop/db.py
@@ -77,6 +77,8 @@ class Source(Base):
     flagged = Column(Boolean, default=False)
     last_updated = Column(DateTime, default=datetime.datetime.utcnow)
     star = relationship("SourceStar", uselist=False, backref="source")
+    journalist_id = Column(Integer, ForeignKey('journalists.id'))
+    journalist = relationship("Journalist", backref="assigned")
 
     # sources are "pending" and don't get displayed to journalists until they
     # submit something
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -430,8 +430,27 @@ def index():
             Submission.query.filter_by(source_id=source.id,
                                        downloaded=False).all())
 
-    return render_template('index.html', unstarred=unstarred, starred=starred)
+    journalists = Journalist.query.order_by(Journalist.username).all()
 
+    return render_template('index.html', unstarred=unstarred, starred=starred, journalists=journalists)
+
+@app.route('/change-assignment/<sid>', methods=('POST',))
+@login_required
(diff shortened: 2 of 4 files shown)


fowlslegs:
Note we "currently allow javascript use in the journalist interface, while gracefully degrading when js is disabled" (see https://github.com/freedomofpress/securedrop/issues/464#issue-36235331 and https://github.com/freedomofpress/securedrop/issues/92#issuecomment-26852285). Journalist assignment should be still possible (through a less pretty or usable interface if need be) when JS is disabled.

Compromise of the server does not mean compromise of past documents, or future documents in the case the source uses PGP. It is a reasonable precaution for a journalist not to use JS on their JW. This security/usability trade-off will eventually be solved by the reading room client, a native application that will query a new (non-HTTP) journalist application server--which we can script to our hearts content because all code will be client-side, and served by signed distribution packages. In the meantime though, this is a nice feature to add.

Note: have yet to review the PR code itself, but will try to get to that this coming week.

mdrose:
Well, in that case, it seems like this sort of interface is necessary for basic functionality and as a fallback to a future, more usable, Javascript interface. I'll try to get a unit test out for this soon, @redshiftzero.

redshiftzero:
Awesome @mdrose! Let me know if you need help or feedback.

fowlslegs:
Added "Needs unit/functional tests" label. Going to wait for those to review.

micahflee:
I haven't actually tested it, but just reading this PR and looking at the screenshots, this looks great to me.

redshiftzero:
I cherry picked your commit @mdrose, made a tiny change and then added a unit test in a superseding PR #1491 so we can get this in SecureDrop - thanks again for your contributions!! 😸

Issue #1346, opened 2016-06-25
Merge changes from 0.3.8 into develop


garrettr:
This is blocking on #1313 and #1345.

Issue #1345, opened 2016-06-25
Merge changes from 0.3.7.post1 into develop


garrettr:
This is blocking on #1313.

PR #1363, opened 2016-07-30
Docs: add linebreak so the line gets code directive

jrabbit:
fixes 
escreenshot from 2016-07-29 20-52-02


conorsch:
Thanks @jrabbit, great catch. The problem only exists in the develop branch; the ReadTheDocs config defaults to showing stable when browsing docs.securedrop.org, so at least the stable docs are currently accurate.

Let me take a look at the develop/master docs differences prior to merging.

conorsch:
Ah, looks like develop docs are still outdated, pending #1345 and #1346. The stable version in RTD looks OK because 0.3.8 includes the change, but develop doesn't yet. Merging. Thanks again, @jrabbit!

Issue #1343, opened 2016-06-20
Journalist replies only support ASCII characters


garrettr:
This is an oversight/classic Python 2.x problem. When a journalist submits a reply to a source, and includes non-ASCII characters in the reply, it causes an error:

```
ERROR:journalist:Exception on /reply [POST]
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1817, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1477, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1381, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1475, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1461, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/var/www/securedrop/journalist.py", line 86, in wrapper
    return func(*args, **kwargs)
  File "/var/www/securedrop/journalist.py", line 448, in reply
    output=store.path(g.sid, filename))
  File "/var/www/securedrop/crypto_util.py", line 161, in encrypt
    armor=False)
  File "/usr/local/lib/python2.7/dist-packages/gnupg/gnupg.py", line 972, in encrypt
    stream = _make_binary_stream(data, self._encoding)
  File "/usr/local/lib/python2.7/dist-packages/gnupg/_util.py", line 547, in _make_binary_stream
    thing = thing.encode(encoding)
UnicodeEncodeError: 'ascii' codec can't encode character u'\\u2019' in position 8: ordinal not in range(128)
```

mdrose:
@garrettr I just tried this and was unable to reproduce it on either develop or 0.3.8. I put some Chinese and Russian text in a journalist reply and it sent with no errors. The message appeared normally on the source side, as well.

garrettr:
@mdrose Huh, that's interesting. We got the bug report from the private support tracker that we use with SecureDrop administrators. I haven't had a chance to try to reproduce yet but will do today or tomorrow.

fowlslegs:
Resolved by #1370.

PR #1370, opened 2016-08-13
Encode UTF-8 and/or convert to BytesIO stream all GPG input

### LINKED ISSUE #1360: Unicode in messages throws internal server error
Writing unicode as as a message throws an internal server error. SecureDrop should support unicode in messages for those of us whose instances are used by non-English speakers. Not only does this happen, we get the generic "internal server error" message that doesn't explain what went wrong or how. This is very bad for sources who upon reaching a SD instance will be unable to submit, not see why, and presumably assume SD isn't a service worth using.

Moving to Python 3 (#997) is probably the easiest way to do this since we get better unicode handling for minimal overhead.

fowlslegs:
This converts plaintext inputs for encryption to `_io.BytesIO` streams before passing to `python-gnupg`'s encrypt function. This should resolve #1360 and #1343--duplicate issues--, however, it would be wise to additionally change the encoding used by mod_wsgi so it's `UTF-8` instead of `C` (as mentioned in https://github.com/freedomofpress/securedrop/issues/1360#issuecomment-239593607) and also to submit a PR upstream to `python-gnupg` so that we can uniformly pass `_io.BytesIO` streams to all `python-gnupg` functions we use, as discussed in the commit message of 56f2308 here.

### DIFF
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -5,6 +5,7 @@ from base64 import b32encode
 
 from Crypto.Random import random
 import gnupg
+from gnupg._util import _is_stream, _make_binary_stream
 import scrypt
 
 import config
@@ -154,6 +155,9 @@ def encrypt(plaintext, fingerprints, output=None):
         fingerprints = [fingerprints, ]
     fingerprints = [fpr.replace(' ', '') for fpr in fingerprints]
 
+    if not _is_stream(plaintext):
+        plaintext = _make_binary_stream(plaintext, "utf_8")
+
     out = gpg.encrypt(plaintext,
                       *fingerprints,
                       output=output,
@@ -165,7 +169,7 @@ def encrypt(plaintext, fingerprints, output=None):
         raise CryptoException(out.stderr)
 
 
-def decrypt(secret, plain_text):
+def decrypt(secret, ciphertext):
     """
     >>> key = genkeypair('randomid', 'randomid')
     >>> decrypt('randomid', 'randomid',
@@ -174,8 +178,7 @@ def decrypt(secret, plain_text):
     'Goodbye, cruel world!'
     """
     hashed_codename = hash_codename(secret, salt=SCRYPT_GPG_PEPPER)
-    return gpg.decrypt(plain_text, passphrase=hashed_codename).data
-
+    return gpg.decrypt(ciphertext, passphrase=hashed_codename).data
 
 if __name__ == "__main__":
     import doctest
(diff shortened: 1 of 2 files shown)


fowlslegs:
Sorry for the churn on this one. I had a lot of back and forth with how to handle this best, but I think I've come to a good conclusion. PR is now official ready for review and should not be subject to further changes.

garrettr:
@fowlslegs This approach seems reasonable to me. Can we add a test for this to prevent future regressions?

fowlslegs:
Added a unit test @garrettr

securedrop/tests/test_unit_integration.py
```diff
@@ -260,6 +260,20 @@ def test_reply_normal(self):
...
+        # method looks at other locale env vars.
+        lc_all_orig = os.getenv("LC_ALL") 
+        os.environ["LC_ALL"] = "C"
+        try:
+            self.helper_test_reply('Ivan Krsti\u0107', True)
+        finally:
+            if lc_all_orig:
+                os.environ["LC_ALL"] = lc_all_orig
+            else:
+                os.environ["LC_ALL"] = ""
```
garrettr: A more compact way to write this is:

```
finally:
    os.environ["LC_ALL"] = lc_all_orig or ""
```

garrettr: Also, this test does not work. It passes even if I comment out the call to `_make_binary_stream` in `crypto_util.encrypt`.

securedrop/crypto_util.py
```diff
@@ -174,8 +179,27 @@ def decrypt(secret, plain_text):
...
+    :param thing: The thing to turn into a encoded stream.
+    :rtype: ``io.BytesIO``
+    :returns: The encoded **thing**, wrapped in an ``io.BytesIO``.
+    """
+    encoding = "utf_8"
+
+    if type(thing) is not str:
+        thing = thing.encode(encoding)
+
+    return BytesIO(thing)
```
garrettr: Why not use `python-gnupg`'s `_make_binary_stream` with an explicit `encoding` parameter? This function seems like an unnecessary duplication of logic from the python-gnupg function.

One could argue that it's bad practice to use "internal" functions from other libraries (prefixed with `_`). Meh. DRY is a good principle. The original code in `crypto_util.encrypt` was also good:

```
if isinstance(plaintext, unicode):
    plaintext = plaintext.encode('utf8')
```

fowlslegs:
Patch rewritten with `_make_binary_stream` from `python-gnupg` library as suggested. @garrettr

garrettr:
Double-checked that the test is working correctly (and fails when we don't pre-encode the submission in `crypto_util.encrypt`). :shipit:

garrettr:
FWIW, I merged that with a merge commit instead of as a "squash merge". Recent internal discussions have indicated a preference for merge commits instead of squash merges, so I'm going to start using them instead.

PR #1333, opened 2016-06-14
Prevent OSSEC alarms when the Tor HS is scanned

redshiftzero:
Adds a rule to fix #1330.

PR #1374, opened 2016-08-15
Adds OSSEC rule to suppress email alerts when the HS is scanned

### LINKED ISSUE #1330: Spurious OSSEC alerts due to THS port scanning
We received a report on our private support portal about frequent OSSEC alerts triggered by the following message in `/var/log/tor/log`:

```
[timestamp redacted] [warn] connection_edge_process_relay_cell (at origin) failed.
```

After enabling `info`-level logging for Tor, we were able to correlate this `[warn]`-level message with preceding `[info]`-level messages:

```
[timestamp redacted] [info] rend_service_set_connection_addr_port(): No virtual port mapping exists for port 443 on service qn4qfeeslglmwxgb
[timestamp redacted] [info] connection_exit_begin_conn(): Didn't find rendezvous service (port 443)
[timestamp redacted] [warn] connection_edge_process_relay_cell (at origin) failed.
```

These messages are triggered by an attempt to open a connection to a hidden service on a port that it does not listen on (e.g. the Source Interface only listens on port 80, but these messages were triggered by attempts to connect on 443). You can easily reproduce it by trying to open a connection to a given hidden service on a port that it does not listen on.

At this point, we decided to analyze some of the logs from the Tor Hidden Services we operate (e.g. for our websites) using our new ELK stack, and found that we were seeing the same patterns.

These log messages appear to be caused by various automated crawlers that try various common "ports of interest" on publicly advertised hidden services (e.g. [OnionScan](https://github.com/s-rah/onionscan/blob/master/onionscan.go)). This type of crawling happens constantly and is usually benign (e.g. for research, testing new tools, etc.).

Since this event is typically benign, it should *not* trigger dozens of OSSEC alerts a day. These alerts are spurious, waste admin's time and attention, and make it difficult to separate signal from noise in the alerts sent by OSSEC. I propose that we develop a mitigation to avoid sending spurious alerts and include it in the next point release (0.3.8).

fowlslegs:
Do not alert attempted connection to the Tor HS on a port the server is not listening on. Supercedes #1333. Fixes #1330.

### DIFF
--- a/install_files/securedrop-ossec-server/var/ossec/rules/local_rules.xml
+++ b/install_files/securedrop-ossec-server/var/ossec/rules/local_rules.xml
@@ -115,3 +115,16 @@
     <options>no_email_alert</options>
   </rule>
 </group>
+
+<!--
+  Do not alert on attempted connections to the Tor HS on a port
+  the server is not listening on. Events are produced by
+  automated crawling/scanning.
+-->
+<group name="tor hs scans">
+  <rule id="200001" level="0">
+    <if_sid>1002</if_sid>
+    <match>connection_edge_process_relay_cell</match>
+    <options>no_email_alert</options>
+  </rule>
+</group>


fowlslegs:
The one thought I had on this is it would be nice to be alerted if a connection was attempted on the private onion services. However, in order to get one of the addresses, an adversary would need to have access to a developer or journalist workstation. This is outside our current threat model, however, access to a journalist workstation is not equated with access to the app server or secure viewing station, so on the other hand this is not a total or un-recoverable compromise. On the other hand, an adversary would have to be pretty stupid to compromise the JW just to try to connect a bunch of non-listening ports, meanwhile sounding all those OSSEC alerts to blow their cover. So, on the whole, I've come to the conclusion that this is good as is. Just wanted to share my thoughts for future reference on the issue if it were to come up.

Issue #1381, opened 2016-08-18
Bento box does not boot and mount correctly in Debian 8


heartsucker:
This was not a problem before with the old Ubuntu box.

My system:
```
heartsucker@pythagoras:~/code/freedom-of-press/securedrop (config-tool)$ VBoxManage --version
4.3.36_Debianr105129
heartsucker@pythagoras:~/code/freedom-of-press/securedrop (config-tool)$ uname -a
Linux pythagoras 3.16.0-4-amd64 #1 SMP Debian 3.16.7-ckt25-2+deb8u3 (2016-07-02) x86_64 GNU/Linux
```

Command output:
```
heartsucker@pythagoras:~/code/freedom-of-press/securedrop (config-tool)$ vag up
Bringing machine 'development' up with 'virtualbox' provider...
==> development: Box 'bento/ubuntu-14.04' could not be found. Attempting to find and install...
    development: Box Provider: virtualbox
    development: Box Version: >= 0
==> development: Loading metadata for box 'bento/ubuntu-14.04'
    development: URL: https://atlas.hashicorp.com/bento/ubuntu-14.04
==> development: Adding box 'bento/ubuntu-14.04' (v2.2.9) for provider: virtualbox
    development: Downloading: https://atlas.hashicorp.com/bento/boxes/ubuntu-14.04/versions/2.2.9/providers/virtualbox.box
==> development: Successfully added box 'bento/ubuntu-14.04' (v2.2.9) for 'virtualbox'!
==> development: Importing base box 'bento/ubuntu-14.04'...
==> development: Matching MAC address for NAT networking...
==> development: Checking if box 'bento/ubuntu-14.04' is up to date...
==> development: Setting the name of the VM: development
==> development: Clearing any previously set network interfaces...
==> development: Preparing network interfaces based on configuration...
    development: Adapter 1: nat
==> development: Forwarding ports...
    development: 8080 (guest) => 8080 (host) (adapter 1)
    development: 8081 (guest) => 8081 (host) (adapter 1)
    development: 22 (guest) => 2222 (host) (adapter 1)
==> development: Running 'pre-boot' VM customizations...
==> development: Booting VM...
==> development: Waiting for machine to boot. This may take a few minutes...
    development: SSH address: 127.0.0.1:2222
    development: SSH username: vagrant
    development: SSH auth method: private key
==> development: Machine booted and ready!
[development] GuestAdditions versions on your host (4.3.36) and guest (5.0.26) do not match.
[...]

PR #1379, opened 2016-08-17
Update deps 0.3.9

fowlslegs:
This upgrades our Python dependencies for the 0.3.9 release as discussed in #1352. Notably, it resolves #1351 by upgrading Selenium to be compatible with the latest version of Firefox, so we can properly run our functional tests. Further discussion of dependency updates is in https://github.com/freedomofpress/securedrop/issues/1352#issue-163540675.

This update does not fix any security issues as far as we are aware and is thus non-criticial. It does relieve us of one dependency, `colorama`, which is small piece of good news as we seek to keep the `securedrop-app-code` package as small and auditable as possible.

It also addresses the (lack of) use of the `-U`/`--upgrade` option w/ `pip`. This means we know explicitly tell `pip` to overwriting existing installed packages (or entries in our `*requirements.txt` files) with the latest or specified versions of those packages we provide.

Lastly, this PR updates the changelog and increments the version number of SD.

It should be noted that while this PR passes all tests, it increases the chance that the `source.async_keygen()` operation will not finish before each unit test and one may more often see warnings that appear as follows:

```
[2016-08-16 20:01:40,796] ERROR in source: async_genkey for source 
(sid=KSNQ2YQGK545HHBJUNJMVIOJHY4EASBMRLNEZGUXROSNKM4FHLP3VRJ6IZ5VRN5X2
JOKGTFWXO2KLS7E7QCIYC6RPCMQHCGAYX546CQ=):
(sqlite3.OperationalError) unable to open database file
```

We have concluded that this the result of the `tearDown()` in our tests, which deletes the contents of the `SECUREDROP_DATA_ROOT` before the key is generated. After generating the key, the `async_keygen()` function goes on to try to access entries in the SQL database that no longer exist, throwing this exception. This is similar to problems we see with `python-gnupg`'s use of daemon threads (see #844, #618, #1350).

Issue #1208, opened 2015-11-20
Improve developer docs


conorsch:
The new docs format is a joy. The recent hackathon shed light on a few aspects of our setup that cause problems for first-time contributors. 

### Documentation expansion
- [x] explanation of the automated tests (both travis and snap)
- [x] useful commands for rebuilding packages
- [ ] ~document need to build ossec packages for staging~ (no longer necessary; see #1668)
- [x] useful commands for running tests locally
- [x] clearer descriptions of the staging environments
- [ ] ~explain use of `vagrant-digitalocean` plugin for remote testing~ (no longer necessary; see #1666)
- [x] document management scripts for `pip` requirements (see https://github.com/freedomofpress/securedrop/pull/1348#issuecomment-240582768)
- [x] explicit version recommendations for:
  - [x] vagrant
  - [x] ansible
  - [x] virtualbox
- [x] document CI pipeline (especially any manual actions required by FPF staff)

### Documentation reformatting
- [x] breakout the environment descriptions into separate page
- [x] clean up .travis.yml file (linting only)
- [x] consistently make `code` blocks copypastable (no prefixed `$`)
- [x] update "copyright" year in docs
- [x] add SD logo to navbar

Right now much of the developer docs are folded into a **Getting Started** section. Breaking out the environment descriptions to a dedicated page would make the docs easier to navigate.

conorsch:
#1381 shows a clear need for explicit versioning in VirtualBox. Do the same for Ansible (related: #1146).

conorsch:
#1306 and #1379 changed the workflow for updating Python dependencies via the `pip_update.sh` script. [From a comment in](https://github.com/freedomofpress/securedrop/pull/1385#issuecomment-242007476) #1385:

> That should not be necessary with the current pip-compile-based workflow. The idea is that the requirements/*.in files list direct dependencies of SecureDrop, and we use pip-compile to generate the requirements/*.txt files from the *.in files, which contain all of the dependencies (including dependencies of dependencies, and so on, for SecureDrop).

> Whenever the *.in files are updated, pip_update.sh should be run to update the *.txt files in tandem. This has been the practice for all of the pull requests related to updating the Python dependencies so far (see #1306 and #1379 for examples).

That's very important to document clearly for devs.

fowlslegs:
[...]

PR #1385, opened 2016-08-23
Add pip-{compile,sync} to dev vm provisioning

heartsucker:
Ansible errors out otherwise because it can't find the `requirements.txt` files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 # ignore the external config file that will contain digital ocean api tokens
 snap.rb
 
+# ignore ansible retry files
+*.retry
+
 # ignore the production config files that don't end in .example
 securedrop-app-conf.yml
 securedrop-mon-conf.yml
--- a/install_files/ansible-base/roles/development/tasks/main.yml
+++ b/install_files/ansible-base/roles/development/tasks/main.yml
@@ -15,9 +15,21 @@
     - apt
     - development
 
-- name: Install pip dependencies for SecureDrop.
+- name: Install pip-tools
   pip:
-    requirements: "{{ securedrop_pip_requirements }}"
+    name: '{{ item }}'
+    state: latest
+  with_items:
+    - pip # pip-tools requires pip 7 or higher
+    - pip-tools
+  tags:
+    - pip
+    - development
+
+- name: Install and sync dependencies for SecureDrop
+  shell: pip-sync securedrop-requirements.txt test-requirements.txt
+  args:
+    chdir: '{{ securedrop_repo }}/securedrop/requirements/'
   tags:
     - pip
     - development


heartsucker:
Now that I'm looking at the code base more, I can't figure out for the life of me why this is needed. The commit this branches from (6230a4a) has both the `.txt` and `.in` files committed to the repo, yet with out this current change, the provisioning breaks every time including on fresh VMs. Am I taking crazy pills?

garrettr:
> Now that I'm looking at the code base more, I can't figure out for the life of me why this is needed.

It shouldn't be. Can you give a more detailed STR? When I run `vagrant destroy -f development && vagrant up development` on the latest develop, provisioning is successful.

What version of Ansible are you running?

conorsch:
Also can create and destroy development VMs fine. Running Ansible v1.8.4 in a virtualenv specifically for SecureDrop.

heartsucker:
@conorsch / @garrettr 

STR
```
vagrant destroy -f development
vagrant up development --no-provision
vagrant ssh -c 'sudo ln -s /opt/VBoxGuestAdditions-4.3.10/lib/VBoxGuestAdditions /usr/lib/VBoxGuestAdditions'
vagrant reload
vagrant provision
```

I tried this multiple times yesterday with the same results, but today when I did it again, vagrant downloaded VBoxGuestAdditions version `4.3.36` over `4.3.10`. The problem has gone away. I have no idea what the underlying problem was, but as of commit f292ba6, everything works.

This can probably be close, though keeping the `pip-sync` task in might help keep people's dev VMs in sync with whatever the current list of requirements is.

conorsch:
Definitely can't reproduce the failure with those steps. Sounds like this is still related to the virtualbox guest additions divergence, as discussed in #1381. Have you tried the [vagrant-vbguest plugin](https://github.com/dotless-de/vagrant-vbguest)? To try it:

```
vagrant plugin install vagrant-vbguest
```

I don't use it myself, but it may help with your use case, since your vbguest version is far out of sync with what's inside the base boxes. If it doesn't resolve your issue with running the VMs under Debian Jessie, make sure to remove the plugin (i.e. `vagrant plugin uninstall vagrant-vbguest`), otherwise it will aggressively run on every up and add additional provisioning time.

heartsucker:
@conorsch I already have that plugin installed and was using it during the previous attempts (forgot to mention). I can no longer reproduce the issue, so I think this can be closed or the current modified change (only adding `pip-sync` instead of `pip install`) should be merged in.

garrettr:
> This can probably be close, though keeping the pip-sync task in might help keep people's dev VMs in sync with whatever the current list of requirements is.

That should not be necessary with the current `pip-compile`-based workflow. The idea is that the `requirements/*.in` files list direct dependencies of SecureDrop, and we use `pip-compile` to generate the `requirements/*.txt` files from the `*.in` files, which contain all of the dependencies (including dependencies of dependencies, and so on, for SecureDrop).

Whenever the `*.in` files are updated, `pip_update.sh` should be run to update the `*.txt` files in tandem. This has been the practice for all of the pull requests related to updating the Python dependencies so far (see #1306 and #1379 for examples).

I can *sort-of* see why it may seem appropriate to use pip-sync to keep a development environment's Python dependencies in sync over time, but IMHO if there is a pull request that updates the Python dependencies, you should just re-provision your VMs (which is fast because we use Ansible). I think that the synchronization of state between the configuration and the environment is best handled by the configuration management tool (Ansible) and not by a multitude of specialized tools such as `pip-sync`.

If it can be shown that Ansible is insufficient for this purpose, or that using `pip-sync` in this manner would be a useful addition for a specific use case, then I would consider including it.

heartsucker:
You're right. I suppose the fact that we don't yet use virtual environments makes `pip-sync` a little useless for now.

conorsch:
> Whenever the *.in files are updated, pip_update.sh should be run to update the *.txt files in tandem. This has been the practice for all of the pull requests related to updating the Python dependencies so far (see #1306 and #1379 for examples).

Thanks for clarifying, @garrettr. Our dev docs don't mention that anywhere. Admittedly it's a recent change, but we need to document that for sanity. I think the best solution here is to update thedev docs with detailed explanation on managing Python dependencies via `pip_update.sh`, then we can close this issue. Will update #1208 with an item for the pip dependencies, then we can close this issue.

garrettr:
I totally agree, @conorsch. We should've made sure it was documented in #1306 before it was merged. Thanks for adding a todo item to #1208 so we don't forget!

conorsch:
Thanks for your active participation here, @heartsucker. Closing this issue out, since we're deferring to resolving this problem with better developer documentation in #1208. If these problems bite you again, please let me know!

PR #1376, opened 2016-08-15
Merges 0.3.7 release into develop branch

conorsch:
Backports changes from the 0.3.7 release on top of the develop branch. Does *not* include changes from:

* 0.3.7-post1 (tracked in #1345)
* 0.3.8 (tracked in #1346)

Conflicts were plentiful, mostly to documentation and Ansible playbooks. All were resolved manually. One merge conflict stood out as deserving particular attention:

```
diff --cc securedrop/crypto_util.py
index c24e94d,487b22a..0000000
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@@ -157,17 -154,11 +157,25 @@@ def encrypt(plaintext, fingerprints, ou
          fingerprints = [fingerprints, ]
      fingerprints = [fpr.replace(' ', '') for fpr in fingerprints]
  
++<<<<<<< HEAD
 +    if isinstance(plaintext, unicode):
 +        plaintext = plaintext.encode('utf8')
 +
 +    encrypt_fn = gpg.encrypt if isinstance(
 +        plaintext,
 +        str) else gpg.encrypt_file
 +    out = encrypt_fn(plaintext,
 +                     *fingerprints,
 +                     output=output,
 +                     always_trust=True,
 +                     armor=False)
++=======
+     out = gpg.encrypt(plaintext,
+                       *fingerprints,
+                       output=output,
+                       always_trust=True,
+                       armor=False)
++>>>>>>> 0.3.7
      if out.ok:
          return out.data
      else:
```

Looks like the develop branch had a commit revising the encrypt function, traceable to          657dfa9f8034a5895f68a06e89d518347e3d6114. In resolving this conflict, **I accepted the 0.3.7 version of the file,** to ease the way forward with backporting release branches onto develop (see #1345, #1346).

Closes #1313.

PR #1387, opened 2016-08-23
Updates Serverspec tests in develop branch

conorsch:
Following up on the recent backporting of release branches onto develop, some Serverspec checks needed to be modified in order to stay in sync with the current expectations for machine state in the `develop` branch. All the changes are to tests, save one, which strengthens the logic for inferring IPv4 addresses only in the staging environment (the previous logic was not bullet-proof).

There are still a few failures in the tests, but are best handled in separate discussions:
- development machine may not have the proper version of selenium (see #1376, and @heartsucker's recent #1385). @fowlslegs can you review the changes here and advise on the correct verisons?
- the tor apt repo pubkey recently changed, and the tests check for an explicit match; updates to the tests should only be made once the relevant ansible task is updated, since the ansible config applies a pubkey. Fresh installations do not seem to be affected, so I don't consider this critical, but I'm glad the tests alerted us to it.
- swap tests are failing; this is because the ansible config disables swap, but doesn't ensure that a swap partition, if one exists, is removed from `/etc/fstab`. This means that after a reboot, swap is reenabled. Wasn't an issue in the old Ubuntu boxes, but since switching to Bento boxes, [swap is on by default](https://github.com/chef/bento/blob/11c5e3aad265e67cc7f5768a373b685c08e3a362/scripts/common/minimize.sh#L15-L22).

Related: #1376, #1377, #1378.


conorsch:
Develop has changed quite a bit recently, particularly due to merging 0.3.10 changes in #1436, and updates to the Travis test suite in #1439. Will rebase on top of latest develop and update the tests as necessary.

conorsch:
Rebased on top of current `develop`. All tests are passing on the development VM. Most tests are passing on staging hosts, except for the swap checks, which now fail since the move to Bento boxes. That's OK in my mind, but I can also temporarily disable those known-failing checks prior to the upcoming hackathon, so new devs aren't confused. 

Have not verified the prod spec tests, due to `nc` version mismatches on Debian Stretch breaking the SSH-over-Tor integration; will solve that separately.

See the [dev docs](https://docs.securedrop.org/en/latest/development/spec_tests.html) for how to run these tests.

redshiftzero:
This looks good to me - I've verified that all serverspec tests are passing on the development VM.

redshiftzero:
Hey @conorsch, I added a small note in the docs regarding an issue I ran into on OS X (that I suspect other contributors may run into)

conorsch:
Thanks, @redshiftzero! I'll fold that docs update into the changes I'm making for #1208.

Issue #1118, opened 2015-09-21
byobu


flamsmark:
`tmux` is okay, but `byobu` is substantially easier to use. It's installed by default in Ubuntu and packaged in Debian.

conorsch:
Definitely open to supporting `byobu`, but I'd prefer not to change the default functionality without extensive testing. Our current strategy of favoring `tmux` over `screen` is a good compromise between stability and future-forwardness. I'd be open to parameterizing the program used as a screen-equivalent, so the playbooks default to using `tmux`, but admins can specify an additional variable in `prod-specific.yml` and set to `byobu` instead.

For context, see #943. We're using a terminal multiplexer strictly for the persistent session functionality. Without one, restarting the `tor` service can result in lockouts on production hosts, requiring physical access for recovery. Given the use case of lockout prevention, the bells and whistles of `byobu` seem like something that should be opt-in, rather than enabled by default.

flamsmark:
This seems fine as an admin-configurable option, but see also #1129.

fowlslegs:
Before I became a pure `tmux` junky, `byobu` was my gateway drug. For those who don't know `tmux`, `byobu` is substantially easier to learn--think `nano` vs `vim`--basic `byobu` operation keybindings are listed on the bottom of the screen in the same manner as nano, actually. For those who do know `tmux`, `byobu` still does everything you'd practically want in the ideally limited time you'll spend SSHed into your SecureDrop instance (it actually uses `tmux` as a backend by default, so really should be able to tap into all of the functionality).

eloquence:
Still worth considering making the switch (or considering other alternatives), esp. as part of overall admin experience improvements.

PR #1390, opened 2016-08-24
Dev tools

### LINKED ISSUE #1369: Consider installing some debug/ development tools in development VMs
As long as we make sure that the tools we choose to install (and their dependencies) won't affect the code that SD executes at any time, it should be safe (and useful) to install, e.g., `ipython` in the development VMs by default. Of course, these tools should stay out of prod just in case. We'd probably want Ansible to install them w/ `pip` instead of using the out-of-date apt packages in trusty.

garrettr:
Closes #1369.


conorsch:
Consider `byobu` in addition to `tmux`, since we have a request for it in #1118 and I'd like to use it more before replacing tmux with it. Clocks in at 1.5MB download. This PR is fine to merge as-is, though. :shipit:.

garrettr:
> Consider byobu in addition to tmux, since we have a request for it in #1118 

Note that the request in #1118 is for byobu to be installed by default in all environments, not just the development environment. It also suggests using byobu instead of tmux for the default terminal multiplexer. Let's resolve that issue separately.

Issue #1366, opened 2016-08-10
Write test suite for our bash scripts


fowlslegs:
We rely on a number of bash scripts for development, deployment, package configuration, and admin maintenance. As of now, we don't have any sort of test suite for these scripts. Historically some of these scripts have not been as robust and trusty as we would like. Something I've seen recommend by @ageis and elsewhere is https://github.com/sstephenson/bats. We might want to look into alternatives as well. Creating a test suite for our bash scripts would help us ensure that changes to them do not break new or existing SD deployments/instances.

garrettr:
In general, I would prefer for us to move away from using bash scripts as much as possible. Bash is simply an awful scripting language. Rather than invest resources in writing tests for bash scripts, I would generally prefer to rewrite them in Python wherever possible (following [Tails' example](https://tails.boum.org/blueprint/Port_shell_scripts_to_Python/)).

fowlslegs:
100% agree @garrettr. On the note of investing resources wisely, we should probably wait to see how the deployment story changes with the rearchitecture before spending time re-writing our current bash scripts in Python.

conorsch:
The more relevant problem here is general lack of testing for the Tails environment overall. The virtualized environments for working on SecureDrop omit Tails representation altogether, requiring manual testing on hardware as part of the release process. Virtualizing a Tails environment and forcing Admin interaction in a GUI VM may be worthwhile to help catch issues with the Tails config.

>I would prefer for us to move away from using bash scripts as much as possible

Truth. In particular it's frustrating to have Ansible for enforcing state on the SecureDrop servers, but not for configuring the Journalist Workstation. (Although we can assume Ansible is installed on the Admin Workstation, Journalists don't have it, so none of the configuration scripts use Ansible for setting up Tails.) We're going against the grain with Tails by overloading the persistence functionality to rewrite system-wide configs, so it shouldn't come as a surprise that we're getting bitten by oversights in the run-once scripts such as `tails_files/install.sh` (e.g. #1394).

PR #1393, opened 2016-08-26
Lock to prevent parallel execution of tails_files/install.sh

ageis:
Ideally only one copy of this script should be running at once, otherwise there might be weird errors and edge cases. We can stop this scenario from happening with `flock`.

### DIFF
--- a/tails_files/install.sh
+++ b/tails_files/install.sh
@@ -27,12 +27,29 @@ torrc_additions="${securedrop_dotfiles}/torrc_additions"
 document_aths_info_global=''
 source_ths_url_global=''
 
+# Use flock to prevent parallel execution.
+function lock() {
+    local lockfile=/tmp/tails_files.lock
+    # create lock file
+    eval "exec 200>/$lockfile"
+    # acquire the lock
+    flock -n 200 \
+        && return 0 \
+        || return 1
+}
+
+function exit_locked_error()
+{
+    echo "Only one instance of install.sh can be run at a time." 1>&2
+    exit 1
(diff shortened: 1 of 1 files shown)


conorsch:
:-1: 

The `tails_files/install.sh` logic is already convoluted enough, and I don't want to introduce additional complexity, particularly since we don't have any automated tests for this script (see #1366). We're already seeing problems related to the automagic configuration approach in Bash (#1394).

> Ideally only one copy of this script should be running at once

That seems a fair assumption we can make as maintainers—adding more untested code to this script is asking for trouble.

Thoughts @garrettr, @fowlslegs?

ageis:
I think it's a little unfair to characterize this as "complexity-introducing untested code". It's a very basic implementation of a locking mechanism that is commonly used in many shell scripts. Here some external resources in case it's any help.

* http://www.kfirlavi.com/blog/2012/11/06/elegant-locking-of-bash-program/
* https://stackoverflow.com/questions/1715137/the-best-way-to-ensure-only-1-copy-of-bash-script-is-running
* http://wiki.bash-hackers.org/howto/mutex
* https://tobrunet.ch/2013/01/follow-up-bash-script-locking-with-flock/
* http://bencane.com/2015/09/22/preventing-duplicate-cron-job-executions/

fowlslegs:
Agree that because it is not important, crucial or necessary, and this already convoluted script should be rewritten in Python, or just replaced when we switch to the 1.0 infrastructure, that we shouldn't merge this one.

PR #1395, opened 2016-08-27
Renamed "document interface" to "journalist interface"

### LINKED ISSUE #1384: Proposal: Change all references of "Document Interface" to "Journalist Interface" to be consistend with code
Document interface is ambiguous because it could mean the source interface because sources submit documents, or it could mean the the journalist interface because the journalist reads documents.

I think this is entirely a documentation issue (plus one of two lines in some `bash` script), and doesn't impact sources in any way.

heartsucker:
Fixes #1384


fowlslegs:
I'm personally in favor of merging this. I've always found Document Interface to be a confusing term when both the source and the journalist interact with a document interface. Source Interface/ Journalist Interface also mirrors the associated `source.py`/ `journalist.py` source code files. If enough SD devs agree with me we should have this rebased to make sure this covers all uses, since there have been some recent changes in `develop`. The reason to keep it Document Interface would be for historical continuation, but I don't think that's a good enough reason--the new name is so clear that people should immediately grasp what it is referring to.

heartsucker:
@fowlslegs Rebased!

fowlslegs:
Thanks. I would merge this, but I can't unilaterally make that decision. Will wait for others input (hopefully soon).

fowlslegs:
@harlo What do you think of this change?

fowlslegs:
@freedomofpress/employees thoughts?

justintroutman:
@fowlslegs I think this is a sensible change. It's consistent with the function of the interface.

conorsch:
The idea here is spot-on. Not only do I already sometimes catch myself saying "Journalist Interface" to clarify, but I've noticed that some folks intuitively believe that "Document Interface" refers to the Source Interface—since that's where documents should be submitted. Makes sense, if a bit convoluted. Naming both interfaces consistently according to role is the clearest option available to us.

fowlslegs:
Alright, it's happening! We're going to try to rebase again for you and make a new superceding PR with your commit, since we've already made you rebase once.

heartsucker:
Small addendum if you guys are rebasing now. The old `.desktop` file for the Admin should be removed in the script that creates them. To prevent clutter/confusion. Give me a minute and I can do it.

redshiftzero:
Just rebased this on develop in branch `heartsucker-rename-document-interface`. @heartsucker  if you want to make that small change in another commit I can cherry-pick it

heartsucker:
@redshiftzero Pushed.

redshiftzero:
Thanks @heartsucker! I see you rebased anyway so I'll destroy my branch and merge this in when the Travis builds complete

heartsucker:
@redshiftzero I had to push my commit somewhere, so I stole your rebasing work and claimed it as my own. :)

redshiftzero:
Hah no worries 😸  Thanks for these contributions!

PR #1396, opened 2016-08-29
A few small updates to docs/deployment_practices.rst

ageis:
See commit messages for details.

### DIFF
--- a/docs/deployment_practices.rst
+++ b/docs/deployment_practices.rst
@@ -45,10 +45,10 @@ is being monitored would be trivially discovered.
 If the landing page is deployed on the same domain as another site, you
 might consider having some specific configuration (such as the security
 headers below) apply only to the /securedrop request URI. This can be done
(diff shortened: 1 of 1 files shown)


fowlslegs:
> Content-Security-Policy : Defined by W3C Specs as standard header, used by Chrome version 25 and later, Firefox version 23 and later, Opera version 19 and later.

It's sufficient to say nobody should be running these browsers anymore and they have bigger problems than what headers the SD landing page sets if they are.

The cipher suite recommendations are the current recommendations of https://cipherli.st/, and comport with recommendations I've read from a number of other sources (e.g., [this recent iSec paper on perfect forward security](https://github.com/iSECPartners/publications/blob/master/whitepapers/perfect_forward_security.pdf?raw=true)).

I've confirmed the nginx HTTPS redirect syntax from other sources.

So everything checks out, going to go ahead and merge this one. Can't believe we were recommending 3DES as a fallback until up to now!

Issue #774, opened 2014-12-04
Custom message on source interface homepage


micahflee:
At The Intercept we have a terrible spam problem with our securedrop instance. We're planning on editing our homepage template to write a simple message, something like: "This a secure submission system for sources, not for general inquiries" and maybe other instructions.

It would be cool if a custom message were officially supported, so we didn't have to edit the template.

b4rton:
1. Because your Source page is necessarily generic, it makes no reference to the journalist or organization hosting it. That could be disorienting or even alarming. Is the source in the right place? In the ideal world, I'd like to be able to put my organization's name and logo alongside the SecureDrop logo. With or without that, how about intro language like: 

"Welcome to the SecureDrop site for confidential communications with [journalist or organization]. If you are seeing this page, your connection is encrypted and anonymous."

2. I think "submit documents" could scare some people off. Some potential sources don't think of themselves as sources at all, or are just dipping a toe, or might get the message that a message is unwelcome without documents. To avoid the presumption:

~~Submit documents for the first time~~
**First time here?**

~~If this is your first time submitting documents to journalists, start here.~~
**If this is your first visit, click the button below to begin a conversation or send a document.**

~~Submit Documents~~
**Get started**

~~Already submitted something?~~
**Been here before?**

~~If you have already submitted documents in the past, log in here to check for responses.~~
**If you have already sent a message or document, use your 'secret phrase' to check for a reply.**

~~Check for a response~~
**Log in**

----------

I would edit the first-time visitor page as well, if possible:

~~Remember this code and keep it secret~~
**Please remember this secret phrase and keep it secure**

~~To protect your identity, we're assigning you a unique codename.~~
**The secret phrase is unique to you and shields your identity. You will need it when you return to the site.**


~~Store this code somewhere safe, or memorize it. It can be used to compromise your anonymity.~~
**Store your secret phrase somewhere safe, or memorize it. Anyone who steals it can read your messages and our replies.**

~~Important: After you submit documents and messages, you will need this code to log in and read replies. This is the only way we can communicate back to you. You can also use your code to log in to submit more files and messages.~~
**Important: After you send a message or document, please return to the site and check for a reply. (Depending on circumstances, that may take a few days.) You will need your secret phrase to log back in. There is no other way for us to respond to you. If you forget it, you will have to begin again as a first-time visitor.**

---

On the page for returning visitors ("Log in to check for responses"):
~~Enter your code name~~
**Enter your secret phrase**

[...]

Issue #1129, opened 2015-09-22
General configurability of individual deployments


tomlowenthal:
As I was pondering issue #1118, I felt a twinge of something that I was repeatedly in the back of my head during the install and configuration of @pressfreedom's system. It seemed a little long & off-topic for that issue, so I decided to separate it out here.

I feel a little queasy about the use of Ansible and adding more options to `prod-specific.yml`. `tmux` versus `byobu` is a primarily cosmetic/usability/convenience change. It's one that helps admins who run their systems a particular way make SecureDrop easy-to-use, convenient, and consistent with their other environments. It may be worth considering this category of configuration changes and how best to manage it in general. What about an admin who prefers a different default shell, or one with plenty of custom shell config files, or… &c. 

My point is that having general configurability depend on upstream SecureDrop development of specific support for options like this is not an ideal solution. This directly impinges on the automatic-update process and use of Ansible for continuous provisioning. This makes it difficult to (confidently) execute minor tweaks to the system — like `tmux` versus `byobu` — but also makes more important changes less reliable. If I edit my `fstab` or add new `cron` jobs or so on: will those changes persist, or will they be lost like tears in rain to some future update, or (worse) conflict with a future unattended change and result in things breaking?

I think that an ideal approach would carefully delineate what SecureDrop relies on and updates/changes automatically, giving admins plenty of room to color inside the lines. Then — when configuration details are determined in the Ansible roles — provide room for configuration. I would suggest that `prod-specific.yml` should remain a place primarily for SecureDrop-specific core configuration which relates to the fundamental functionality of the system. Additional configuration choices (like — say — which terminal multiplexer to use) should probably live in another file.which might be longer and more convoluted, but which there would be no general obligation to edit.

flamsmark:
This general tenor of approach might also present a way forward for #67.

conorsch:
Since we're currently using Ansible to handle deployment and configuration, the most straightforward approach to providing site-specific configurability is via variable overrides. Right now the prod playbook expects all site-specific vars to be found in `prod-specific.yml`, which is version controlled with default vars, and forces admins to use a stash-pull-pop workflow to upgrade versions of securedrop.

[...]

PR #1399, opened 2016-09-01
Add custom notification text to source interface

redshiftzero:
This PR implements issue #774, enabling an admin to put some custom text on the source interface using a new option `custom_notification` in `prod_specific.yml`. When this variable contains some text, it will look like this: 

custom_notification

An admin can instead leave this option empty and the source interface will look as before.


conorsch:
Clean implementation so far, but tests are failing. You'll need to create a mapping for the template var in via `app.jinja_env.globals` in `source.py`.

conorsch:
Sorry, I misread the error message. Here it is in full:

> TASK [app : Copy custom notification text to config.py.] ***********************

> fatal: [localhost]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: {{ custom_notification_text }}: 'custom_notification_text' is undefined\n\nThe error appears to have been in '/home/travis/build/freedomofpress/securedrop/install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml': line 144, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Copy custom notification text to config.py.\n  ^ here\n"}

That's an undefined value at the Ansible level—the Jinja template looks good!

redshiftzero:
Whoops - forgot to add the new variable to `travis.yml`. Tests are working now.

install_files/ansible-base/group_vars/securedrop_application_server.yml
```diff
@@ -45,6 +45,9 @@ local_deb_packages:
 
 ### Used by the app role ###
 
+# Custom text on source interface
+custom_notification: "{{ custom_notification_text }}"
```
conorsch: Why the wrapper var here? Looks like this should simply be `custom_notification_text: ""`.

install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml
```diff
@@ -141,6 +141,14 @@
     - gpg
     - securedrop_config
 
+- name: Copy custom notification text to config.py.
+  lineinfile:
+     dest: "{{ securedrop_code }}/config.py"
+     regexp: "^CUSTOM_NOTIFICATION"
+     line: "CUSTOM_NOTIFICATION = '{{ custom_notification }}'"
```
conorsch: 1. Var should be referred to as `custom_notification_text` for consistency.
2. Tack on a conditional to ensure Ansible responds sanely to presence or absence of this var:
```
when: custom_notification_text is defined and custom_notification_text != ''
```

conorsch:
The Ansible vars are inconsistently named; depending on the vars file, it's either referred to as `custom_notification` or `custom_notification_text`. I think it should be the latter in all Ansible files, and the former in all Jinja templates. 

Even if that is fixed, however, we still have a breaking change for production installs: any changes to the `prod-specific.yml` template are problematic, because of the `pop-stash-pull` git workflow we mandate for upgrading tagged versions of SecureDrop on Admin Workstations. This is technical debt that we'll need to resolve for 0.4 (see #1129 for a discussion about how to fix). To see the problem, try applying the changes presented in this PR in a manner similar to how an Admin would do so:

````
git checkout -b develop-test-merge develop
cd install_files/ansible-base
# edit prod-specific.yml by hand, filling in custom vars
git stash # save site-specific vars
git merge custom-notification
git pop # merge conflict!
```

You can try moving the new var to the bottom of `prod-specific.yml`, which _should_ avoid a merge conflict. Just tried that, and looks to behave well. Another approach is to have the Ansible task check for definedness of the var—I've added an inline comment with an example.

redshiftzero:
Thanks for the detailed feedback. I’ve fixed the inconsistent naming and have added your suggested check as well as moved the new var to the end of `prod_specific.yml`. The `pop-stash-pull` workflow now results in a merge with no conflicts.

install_files/ansible-base/group_vars/travis.yml
```diff
@@ -13,3 +13,6 @@ securedrop_header_image: ""
 # install_files/ansible-base/roles/app/files/ directory
 securedrop_app_gpg_public_key: "test_journalist_key.pub"
 securedrop_app_gpg_fingerprint: "65A1B5FF195B56353CC63DFFCC40EF1228271441"
+
+# Custom text on source interface
+custom_notification_text: ""
```
fowlslegs: The no newline warning strikes again!

redshiftzero:
OK, added the newline and added a screenshot and explanation to the docs (also rebased). Take a peek @fowlslegs and let me know how it looks!

install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml
```diff
@@ -141,6 +141,15 @@
     - gpg
     - securedrop_config
 
+- name: Copy custom notification text to config.py.
+  lineinfile:
+     dest: "{{ securedrop_code }}/config.py"
+     regexp: "^CUSTOM_NOTIFICATION"
+     line: "CUSTOM_NOTIFICATION = '{{ custom_notification_text }}'"
+  when: custom_notification_text is defined and custom_notification_text != ''
```
fowlslegs: This overwrites `config.py` if you change `prod-specific.yml` and reprovision. I don't consider this because I edit `config.py` myself to test things, but I suppose our intention is that admins never touch it themselves (is that correct @conorsch?). Maybe we should set the `group_var` for the staging and development VMs to automatically note that this is a testing instance, and specifically, whether it's a staging or development box as a developer convenience, especially in the context of https://github.com/freedomofpress/securedrop/pull/1458.

docs/install.rst
```diff
@@ -99,6 +99,12 @@ following:
 -  The fingerprint of your SMTP relay (optional):
    ``smtp_relay_fingerprint``
 
+Optionally, you can also have custom notification text be displayed on the source interface. The source interface with a custom notification message is shown here (the custom notification appears after the bolded "Note:"):
```
fowlslegs: Hard wrap these at 80.

install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml
redshiftzero: Good idea @fowlslegs: I have added `custom_notification` text to the development and staging VMs

docs/install.rst
redshiftzero: Done!

redshiftzero:
Hey @fowlslegs I've made your suggested changes - should be good to go!

PR #1404, opened 2016-09-07
Make backups work

KwadroNaut:
On close examination, backups were not being made. The test in the backuprole perform_backup == true was not evaluated as expected. Known ansible problem, fixed upstream, but we're using the slightly older packaged version in Tails. See https://github.com/ansible/ansible/issues/9369
Evaluates to false:
(perform_backup is defined and perform_backup == true)
Compares the string:
(perform_backup is defined and perform_backup == 'true')
Or uhm, make it look like php:
(perform_backup is defined and perform_backup|bool == true)

### DIFF
--- a/install_files/ansible-base/securedrop-prod.yml
+++ b/install_files/ansible-base/securedrop-prod.yml
@@ -49,7 +49,9 @@
       # By default backups will NOT run, since they can claim a substantial
       # amount of disk space. Enable backups explicitly via a backup var, or
       # implicitly if performing a restore.
-      when: (perform_backup is defined and perform_backup == true) or
+      # Due to ansible version in current Tails, we have to pipe to bool
+      # https://github.com/ansible/ansible/issues/9369
+      when: (perform_backup is defined and perform_backup|bool == true) or
             (restore_file is defined and restore_file != '')
       tags: backup
 


fowlslegs:
@conorsch Can you confirm and merge this if this is the case?

conorsch:
@KwadroNaut Thanks for reporting this. We're discussing inadequate testing procedures regarding matching Ansible versions across multiple environments in #1146. Looks like this problem is specific to older versions than even what we're running in the dev environment, but pre-release QA on hardware should have caught this.

@fowlslegs Can you reproduce this error in Tails running Ansible 1.7.2 installed via apt? We'll definitely need to update the playbooks in a new version so we can point people to the upgrade docs if you can confirm.

conorsch:
Thanks for writing this, @KwadroNaut. We've received similar feedback on the backup behavior from a few Admins, so we definitely need to get this in. Will confirm the behavior on test hardware, then merge if all goes well.

conorsch:
Confirmed resolution with the patch presented. Another admin also pointed out that the [backup docs](https://docs.securedrop.org/en/stable/backup_and_restore.html#run-the-backup-ansible-role) don't include the `-u <username>` option in the ansible-playbook invocation, though they should.

Merging. We should get this out in the next release.

PR #1405, opened 2016-09-07
Missing username deploying ansible-playbook in doc

KwadroNaut:
It wouldn't connect if not specifying your SSH username, when it's different as your current username.


fowlslegs:
Yep, thanks!

Issue #1409, opened 2016-09-20
Rewrite bash scripts in Python


redshiftzero:
Rewrite `pip_update.sh` and `upgrade_version.sh` in Python such that they are cleaner and easier to test

fowlslegs:
```
% find . -not -path '*/\.*' -type f -name '*.sh'
./install_files/ansible-base/roles/app-test/files/xvfb_display.sh
./install_files/ansible-base/roles/ossec-server/templates/send_encrypted_alarm.sh
./install_files/securedrop-ossec-server/var/ossec/setup_gmail_test.sh
./migration_scripts/0.3pre/upgrade.sh
./pip_update.sh
./securedrop/test.sh
./tails_files/install.sh
./tails_files/65-configure-tor-for-securedrop.sh
./update_version.sh
./build/build_generic_package.sh
./snap_ci/development_clean.sh
./snap_ci/prod_clean.sh
./snap_ci/staging_clean.sh
./snap_ci/pinned.sh
```

Some of these will have to remain as bash scripts, some of these are low-priority for conversion, and others are basically irrelevant altogether this point. Besides the two you've mentioned, I'd say `tails_files/install.sh` is the top candidate for conversion to Python.

cowlicks:
Here is an alternative to `pip_update.sh` in python

```python
#!/usr/bin/env python

import os
import sys
import subprocess

def which(program):
    def is_exe(fpath):
        return os.path.isfile(fpath) and os.access(fpath, os.X_OK)

    fpath, fname = os.path.split(program)
    if fpath:
        if is_exe(program):
            return program
    else:
        for path in os.environ["PATH"].split(os.pathsep):
            path = path.strip('"')
            exe_file = os.path.join(path, program)
            if is_exe(exe_file):
                return exe_file

    return None

dir_ = os.path.dirname(os.path.realpath(__file__))
requirements_dir = os.path.join(dir_, 'securedrop/requirements')
venv='review_env'

if 'VIRTUAL_ENV' in os.environ:
    print('Please deactivate your virtualenv before running this script.')
    sys.exit(1)

install='sudo apt-get install -y'.split()

install_pip = install + 'python-pip'.split()
install_venv = install + 'python-virtualenv'.split()
mkvirtualenv = 'virtualenv -p python2.7'.split() + [venv]

big_command = """
virtualenv -p python2.7 {venv}
source "{venv}/bin/activate"

pip install --upgrade pip
pip install pip-tools

[...]

PR #1416, opened 2016-10-03
Installs securedrop-keyring package for managing apt signing key

conorsch:
Provides build scripts to create a new Debian package `securedrop-keyring`, so the expiration date on the offline master signing key can be extended at regular intervals. The new `securedrop-keyring` package is set as a dependency of the existing SecureDrop packages:

* securedrop-app-code
* securedrop-ossec-agent
* securedrop-ossec-server

The dependency declaration enables automatic updates to the signing key via the nightly cron-apt process, so no Admin intervention is required in order to rotate the signing key (as was the case with [0.3.6](https://github.com/freedomofpress/securedrop/blob/develop/changelog.md#036) last year).

Regarding testing:

- [x] Rebase on a new release/0.3.10 branch based on the 0.3.9 tag
- [ ] Testing. We need to test:
 - [x] New installations
 - [x] Updating current installations via a new SecureDrop release (using the mechanism described in https://github.com/freedomofpress/securedrop/pull/1197#issuecomment-223462885).
 - [x] Updating key expiry via the keyring package (the primary intended use case).
 - [ ] Rotating keys via the keyring package.
- [ ] Update the expiration date on the production code signing key and publish via the keyring package.

Reimplementation of #1197; supersedes and therefore closes #1197.


conorsch:
During testing, using a staging apt server and a disposable test key. Patch for applying the test apt URL and pub key can be found here: https://gist.github.com/conorsch/63eb1d15070901d0abb890dbccc2cca9

Prior to merge a rebase is required, to swap out the test key with the production offline master signing key, after its expiration date has been pushed forward into 2017; thus the **In progress** label.

ageis:
Your solution does not provide for [key removal](https://github.com/freedomofpress/securedrop-keyring/blob/master/debian/prerm). If the key or repository is compromised, it should be possible to remove and untrust it, but as it stands, it will persist on the system even if the package is uninstalled.

install_files/ansible-base/roles/install_fpf_repo/tasks/main.yml
```diff
@@ -3,9 +3,15 @@
   apt_key: >
```
ageis: I don't really understand why you're relying on Ansible to actually add the key into the trusted keyring or why this approach is superior to the previous work that I did. All your package does is put it on the filesystem and help with rotations.

I recommend that you add an [install](https://www.debian.org/doc/manuals/maint-guide/dother.en.html#install) file so it's clear that you have files destined for `/etc/apt/trusted.gpg.d`.

conorsch:
> Your solution does not provide for key removal.

False, key removal is absolutely supported. See terminal session below: upon removal of the `securedrop-keyring` package, the associated apt key is automatically removed, because we are using a discrete keyring specifically for this package. Note that the `securedrop-app-code` and `securedrop-ossec-agent` packages are also removed as well, since we've set a dependency on `securedrop-keyring` in their control files. That's a side-effect of rolling this change out without requiring Admin intervention, and is defensible, but must be acknowledged all the same.

```
vagrant@app-staging:~$ aptitude search securedrop~n     
i   securedrop-app-code                                                     - Packages the SecureDrop application code pip dependencies and apparmor profiles. 
i   securedrop-grsec                                                        - SecureDrop grsec kernel (metapackage depending on the latest version)            
i A securedrop-keyring                                                      - Provides an apt keyring for SecureDrop-related packages, so the master signing ke
i   securedrop-ossec-agent                                                  - Installs the securedrop pre-configured OSSEC agent                               
p   securedrop-ossec-server                                                 - Installs the pre-packaged OSSEC server                                           
vagrant@app-staging:~$ apt-key list | tail -n 5                
/etc/apt/trusted.gpg.d/securedrop-keyring.gpg
---------------------------------------------
pub   2048R/AE2C8E02 2016-09-28 [expired: 2016-09-30]
uid                  SecureDrop TEST apt repo key for 0.3.10-rc (TEST KEY, DO NOT USE IN PRODUCTION) <securedrop@freedom.press>

vagrant@app-staging:~$ sudo apt-get remove securedrop-keyring
Reading package lists... Done
Building dependency tree       
The following packages were automatically installed and are no longer required:
  apache2 apache2-bin apache2-data apache2-mpm-worker expect haveged
  inotify-tools libapache2-mod-wsgi libapache2-mod-xsendfile libapr1
  libaprutil1 libaprutil1-dbd-sqlite3 libaprutil1-ldap libhavege1
  libinotifytools0 libjemalloc1 libsqlite0 libtcl8.6 ossec-agent
  python-chardet-whl python-colorama python-colorama-whl python-distlib
  python-distlib-whl python-html5lib python-html5lib-whl python-meld3
  python-pip python-pip-whl python-requests-whl python-setuptools-whl
  python-six-whl python-urllib3-whl python-wheel redis-server redis-tools
  secure-delete sqlite sqlite3 ssl-cert supervisor
Use 'apt-get autoremove' to remove them.
The following packages will be REMOVED:
  securedrop-app-code securedrop-keyring securedrop-ossec-agent
0 upgraded, 0 newly installed, 3 to remove and 0 not upgraded.
After this operation, 0 B of additional disk space will be used.
Do you want to continue? [Y/n] y
(Reading database ... 43436 files and directories currently installed.)
Removing securedrop-app-code (0.3.10-rc3) ...
Removing securedrop-ossec-agent (2.8.2+0.3.10-rc3) ...
Removing securedrop-keyring (0.1.0+0.3.10-rc3) ...
vagrant@app-staging:~$ apt-key list 
/etc/apt/trusted.gpg
--------------------
pub   1024D/437D05B5 2004-09-12
uid                  Ubuntu Archive Automatic Signing Key <ftpmaster@ubuntu.com>
sub   2048g/79164387 2004-09-12

pub   1024D/FBB75451 2004-12-30
uid                  Ubuntu CD Image Automatic Signing Key <cdimage@ubuntu.com>

pub   4096R/C0B21F32 2012-05-11
uid                  Ubuntu Archive Automatic Signing Key (2012) <ftpmaster@ubuntu.com>

pub   4096R/EFE21092 2012-05-11
uid                  Ubuntu CD Image Automatic Signing Key (2012) <cdimage@ubuntu.com>

pub   4096R/FC9F6818 2014-10-26 [expires: 2016-10-27]
uid                  Freedom of the Press Foundation Master Signing Key
sub   4096R/4833B9A3 2014-10-26

/etc/apt/trusted.gpg.d/deb.torproject.org-keyring.gpg
-----------------------------------------------------
pub   2048R/886DDD89 2009-09-04 [expires: 2020-08-29]
uid                  deb.torproject.org archive signing key
sub   2048R/219EC810 2009-09-04 [expires: 2018-08-30]

vagrant@app-staging:~$
```

As you can see in the full output of `apt-key list` at the end of the terminal session, no apt key related to SecureDrop exists on the system after package removal.

ageis:
@conorsch You are right, I spoke too soon. Since the keyring file is provided by the package, it will be removed with the package. Interesting. That would be clearer if there were an install file like I mentioned in my other comment.

conorsch:
Reviewed these changes in person with @garrettr, and we're happy with the implementation as proposed. Still need to update this PR with the production pubkey info, then a rebase for cleanup, and we're set to cut the new release before the end of the week.

install_files/ansible-base/roles/install_fpf_repo/tasks/main.yml
```diff
@@ -3,9 +3,15 @@
...
     state=present
     data="{{ lookup('file', 'fpf-signing-key.pub') }}"
+    keyring=/etc/apt/trusted.gpg.d/securedrop-keyring.gpg
```
ageis: Just came back to double check that this file is raw instead of ASCII-armored. That results in issues. Looks like you already figured that out. 👍

changelog.md
```diff
@@ -1,5 +1,20 @@
 # Changelog
 
+## 0.3.10
+
+Creates new Debian package `securedrop-keyring` for managing the SecureDrop
+Release Signing Key. Rotates the signing key currently in use by setting
+a dependency on the other Debian packages, so that currently running deployments
+will have their apt keyrings updated via automatic nightly updates.
```
garrettr: This wording is confusing. We set the dependency to ensure that the securedrop-keyring package will be installed automatically through cron-apt on existing installations. We are also rotating the signing key, but that is an orthogonal concern.

install_files/securedrop-keyring/DEBIAN/preinst
```diff
@@ -0,0 +1,31 @@
...
+   h="$(mktemp -d)"
+   trap "rm -rf '$h'" EXIT
+
+   if gpg --homedir="$h" \
+          --batch --no-tty --no-default-keyring --keyring /etc/apt/trusted.gpg \
+          --list-key 0xBFFDF68849BC8719F1BE0191321A7B10AE2C8E02 > /dev/null 2>&1 ; then
+      gpg --homedir="$h" \
+          --batch --no-tty --no-default-keyring --keyring /etc/apt/trusted.gpg \
+          --no-auto-check-trustdb \
+          --delete-key 0xBFFDF68849BC8719F1BE0191321A7B10AE2C8E02 || true
```
garrettr: What key is this fingerprinting referring to? I expected this to be the old (about to be rotated) "Freedom of the Press Foundation Master Signing Key", but it is not:

```
securedrop ~> gpg -k "Freedom of the Press Foundation Master Signing Key"
pub   4096R/0xBADDE0C7FC9F6818 2014-10-26 [expires: 2016-10-27]
      Key fingerprint = B89A 29DB 2128 160B 8E4B  1B4C BADD E0C7 FC9F 6818
uid                 [  full  ] Freedom of the Press Foundation Master Signing Key
sub   4096R/0xE02E0B594833B9A3 2014-10-26
```

changelog.md
conorsch: Since release 0.3.10 rotates the signing key, rotation of the signing key should be included in the changelog. We could omit mention of the new package dependencies, but in light of the expiration that happened in October 2015, where we required Admin intervention on all instances, it makes sense to include mention of the approach that enables us to update currently running instances automatically.

install_files/securedrop-keyring/DEBIAN/preinst
conorsch: You're right! That's _not_ the key we want to remove—that's a testing key used temporarily on the staging apt server. Updated with the currently active fingerprint `0xB89A29DB2128160B8E4B1B4CBADDE0C7FC9F6818`, and rebased for clean commit history. 

New commit implementing correct fingerprint is cd36f6909c4abbcf5ec950fe174f9cf70731914c (was 26559ee334e85e466f5bd90ae2baca94ccc83d12).

conorsch:
> Do you think we should also add securedrop-keyring as a dependency of the securedrop-grsec metapackage? 

Sure, that sounds reasonable, but those changes won't happen in this repository—the securedrop-grsec metapackage build logic is maintained over in https://github.com/freedomofpress/ansible-role-grsecurity. Opened up an issue over there to track that change, since we need to release a new kernel soon anyway.

changelog.md
garrettr: @conorsch Agreed! I simply think it would be clearer if the changelog was written to express those two things separately, e.g.

"Creates new Debian package `securedrop-keyring` for managing the SecureDrop Release Signing Key. Adds the new keyring package as a dependency to the other SecureDrop packages so that currently running deployments will have their apt keyrings updated via automatic nightly updates.

In 0.3.10, `securedrop-keyring` rotates the code signing key used for SecureDrop. Since the keyring package will be installed automatically as a dependency of the other SecureDrop packages, this rotation will happen automatically and does not require admin intervention."

But this is a minor point. If you prefer the original wording, leave it and we'll ship it.

garrettr:
@conorsch Ah, I had forgotten we moved the package building code for securedrop-grsec from this repo to the ansible-role-grsecurity repo. I agree that should be handled separately.

install_files/securedrop-keyring/DEBIAN/preinst
```diff
@@ -0,0 +1,31 @@
...
+   h="$(mktemp -d)"
+   trap "rm -rf '$h'" EXIT
+
+   if gpg --homedir="$h" \
+          --batch --no-tty --no-default-keyring --keyring /etc/apt/trusted.gpg \
+          --list-key 0xB89A29DB2128160B8E4B1B4CBADDE0C7FC9F6818 > /dev/null 2>&1 ; then
+      gpg --homedir="$h" \
+          --batch --no-tty --no-default-keyring --keyring /etc/apt/trusted.gpg \
+          --no-auto-check-trustdb \
+          --delete-key 0xB89A29DB2128160B8E4B1B4CBADDE0C7FC9F6818 || true
```
garrettr: Minor note: DRY for the key id? A typo here would be hard to spot.

conorsch: Struck me as strange as well, but I decided to reuse DKG's approach. A bit cargo-culty, but it definitely works as expected. Going forward, as we modify the `preinst` script to accommodate future rotations or other features, we can make the script more readable.

What I'd really like is the ability to reuse the keyid from an Ansible variable, so it's DRY across the entire project—that will take more consideration to do sanely, though.

changelog.md
conorsch: Not a bad edit at all, but let's stick with what's in place given time constraints. In other messaging surrounding the release, such as blog post announcement, we can be explicit about the specifics.

install_files/securedrop-keyring/DEBIAN/preinst
```diff
@@ -0,0 +1,31 @@
...
+# In SecureDrop versions before 0.3.10, the Ansible playbooks used
+# the apt_key module to add the signing key to /etc/apt/trusted.gpg.
+# It's cleaner to use the trusted.gpg.d subdirectory, since we can
+# update that trivially in future versions of the keyring package.
+#
+# Therefore let's clean up prior versions of the key installed
+# to the general apt keyring, to ensure we only have one signing key
+# installed for authenticating securedrop-related packages.
+
+if [ -e /etc/apt/trusted.gpg ] && which gpg >/dev/null; then
```
fowlslegs: `if [[ -e /etc/apt/trusted.gpg ]] && command -v gpg; then`

install_files/securedrop-keyring/DEBIAN/preinst
```diff
@@ -0,0 +1,31 @@
...
+   (
+   h="$(mktemp -d)"
+   trap "rm -rf '$h'" EXIT
+
+   if gpg --homedir="$h" \
```
fowlslegs: `"${h}"`

install_files/securedrop-keyring/DEBIAN/preinst
```diff
@@ -0,0 +1,31 @@
...
+# update that trivially in future versions of the keyring package.
+#
+# Therefore let's clean up prior versions of the key installed
+# to the general apt keyring, to ensure we only have one signing key
+# installed for authenticating securedrop-related packages.
+
+if [ -e /etc/apt/trusted.gpg ] && which gpg >/dev/null; then
+   (
+   h="$(mktemp -d)"
+   trap "rm -rf '$h'" EXIT
```
fowlslegs: `trap "rm -rf ${h}" EXIT`?

install_files/securedrop-keyring/DEBIAN/preinst
```diff
@@ -0,0 +1,31 @@
...
+
+   if gpg --homedir="$h" \
+          --batch --no-tty --no-default-keyring --keyring /etc/apt/trusted.gpg \
+          --list-key 0xB89A29DB2128160B8E4B1B4CBADDE0C7FC9F6818 > /dev/null 2>&1 ; then
+      gpg --homedir="$h" \
```
fowlslegs: `"${h}"`

install_files/securedrop-keyring/DEBIAN/preinst
fowlslegs: Continuing with my shell style comments: `readonly KEY_FPR='0xB89A29DB2128160B8E4B1B4CBADDE0C7FC9F6818'` should be initialized right after `set -e`. If you want to use the keyid from an Ansible variable this should definitely be a Python script, which would bring us one step closer to resolving #1409 as well.

install_files/securedrop-keyring/DEBIAN/preinst
```diff
@@ -0,0 +1,31 @@
...
+# the apt_key module to add the signing key to /etc/apt/trusted.gpg.
+# It's cleaner to use the trusted.gpg.d subdirectory, since we can
+# update that trivially in future versions of the keyring package.
+#
+# Therefore let's clean up prior versions of the key installed
+# to the general apt keyring, to ensure we only have one signing key
+# installed for authenticating securedrop-related packages.
+
+if [ -e /etc/apt/trusted.gpg ] && which gpg >/dev/null; then
+   (
```
fowlslegs: These parens are unnecessary and hurt readability in my opinion because at first you wonder if they're doing anything.

install_files/securedrop-keyring/DEBIAN/preinst
fowlslegs: Forgot `> /dev/null 2>&1` at the end of this line.

garrettr:
@fowlslegs It's too close to the release deadline to propose changes like that to the preinst script. As @conorsch has explained in this issue and elsewhere, most of that code was copied or adapted from the Tor Project's similar keyring package. If it ain't broke, don't fix it.

conorsch:
Thank you @garrettr, and thank you @fowlslegs for the detailed review. As @garrettr mentioned, we're too close to the current key expiry to modify this PR anymore—doing so would require circling back around with rebuilding the packages and stepping through the airgapped signing procedure on the staging apt server all over again.

fowlslegs:
I was late to reviewing this, and 100% agree we shouldn't make more, costly-re-review-inducing changes at this point. Had I posted my review comments a couple weeks ago, I think we should implemented them. I think the value of the "if it ain't broke, don't fix it" proverb only stretches so far.

PR #1418, opened 2016-10-10
Add diagram of securedrop database

redshiftzero:
This PR adds a diagram to the documentation showing the structure of the database - useful to get a quick high level view of what (and how) data is stored in securedrop's SQLite database.


garrettr:
This looks great, @redshiftzero! Out of curiosity, do you know of any tools that can automatically generate entity-relationship diagrams? This diagram looks great, but it'd be nice if we didn't have to manually update it as the database schema changes over time.

redshiftzero:
I totally agree - but I'm not aware of a tool like that.  It would be great to have a script that would automagically produce this diagram based on `securedrop/db.py`

conorsch:
Great diagram! The changes here don't display the diagram anywhere in the docs, though. How about adding it under the Developer Docs, so we can link to it in the future?

redshiftzero:
Good point! I just edited the tail of the Getting Started developer doc to show this diagram in "Tips and Tricks" as well as point to the other helpful diagrams in the `/diagrams` directory.

conorsch:
Now we're talking. Thanks, @redshiftzero!

Issue #1170, opened 2015-11-03
Submission categories


tomlowenthal:
At CPJ, there are multiple teams who check SecureDrop for submissions. Other newsrooms may have a similar situation — different teams or authors who are not particularly interested in submissions meant for each others. CPJ staff would very much like it if there were a way for a source to pick a topic or department for their submission: metadata that is visible to journalists checking the document interface.

In a more advanced version, it would be best to be able to assign users to these groups so that someone only sees relevant material when they log in, and don't have the opportunity to mark other teams' submissions as read or delete them (more from the perspective of practical logistics than security).

garrettr:
Great idea! This should be relatively straightforward to implement. You'll need to add some way to manage these categories (and access control based on categories) to the Admin portion of the Document Interface.

taravancil:
I'm working on this! I'll submit the finished PR tomorrow as my contribution to the hackathon.

A few questions as I finish up:
* Should non-admins be able to create new categories or add/remove them from submissions? I've reserved these capabilities for admins, but I can easily change that.
* I'm thinking about adding the ability to make a category private, so that sources can't see the category or add it to submissions. I imagine this will be useful if an organization uses categories like "high-priority", "Journalist Name", "team A", etc. Good/bad idea?

garrettr:
Let's focus on MVP for this initial implementation, and file follow ups for more complex functionality if desired. As described, the goal for this issue appears to be:

* A source chooses a category or categories for their submission from a drop-down menu when they submit.
    * You should choose early on whether a submission can have one or multiple categories, since this defines the underlying representation in the database. Depending on this choice, the Source UI should be a <select> or a list of checkboxes.
* The categories are free-form text labels. Categories can be created, edited, and deleted by admin users on the Document Interface.
    * You will need to consider and document the correct behavior for submissions if one, several, or all of their categories are deleted.
* There should be always be a default/uncategorized category available.
* Journalist users will only be able to see submissions from a subset of categories. Which journalists can see which categories should be configurable by admin users on the Document Interface. This forms the basis for access control within the Document Interface.
    * Note that all submissions will still be encrypted with the same offline submission key - access control occurs at the level of the Document Interface, not at the cryptographic level. This is actually necessary to achieve category agility. It may be desirable to add flexibility at the cryptographic layer for various reasons, including access control, but that is outside the scope of this issue.

taravancil:
[...]

PR #1422, opened 2016-10-17
Tagging of Sources and Submissions

redshiftzero:
This PR implements tagging, similar to what's suggested in #1170, but more using GitHub-style labels. I see there is another PR for journalist assignment, imho it would be cool to have both of these features. How this labelling feature works:
- An admin user defines a set of labels for both submissions and sources. There is a new panel in the admin view enabling them to add and delete tags. I thought it might be good to restrict this to admins for OPSEC reasons, but happy to discuss further and modify this. I also thought that it would be worth having separate categories of labels for submissions and sources. There is also an OPSEC warning that is always there in the add tag admin view:

admin-view
- Any user can then associate any source or submission with one or more labels. Any user can also remove labels from any source or submission. Here's the source view with some labels:

unfiltered-source-view

I put the adding or removing of all labels on the submission view so it didn't clutter the source view too much. In the submission view you can add or remove labels to that particular source, and add or remove labels to specific submissions. Here's the view of a source's submissions with some labels:

submission-view
- A journalist or an admin user can also filter the source view to show only sources with a specific label. This can be done by clicking the relevant label in "Filter by label". They can return to the unfiltered view by clicking again on the label:

filter-view

Some example situations where tags implemented in this way might be useful: 
- A point person that checks SecureDrop every day could look at the submissions, and then might tag those with a "team" - e.g. "environment". This enables the point person on the environment team to then log in and spend time downloading and decrypting only the submissions they need to attend to. 
- A point person might also use the tags for setting priorities: e.g. "high-P". 
- Users might use the tags to describe the status of working with a source or submission: e.g. "waiting on response", "journo needed", etc. If multiple people are logging into SecureDrop, this gives each user a way to know what the status of handling each source/submission is without having to discuss in person or on another less secure platform. This is especially important for users and admins distributed over a wide geographic area without the ability to confer in person.

Let me know your thoughts on this feature.. if/when we agree it looks OK then I'll write some unit tests for this. Some additional documentation should also be written for this but I'm happy to do so if this is OKed.


securedrop/static/css/journalist.css
```diff
@@ -499,6 +503,40 @@ p#no-replies {
...
 
+.tag-label {
+  display: inline-block;
+  border-bottom: 1px solid #ddd;
+  text-align: left;
+  border: 1px solid #dedede;
+  padding: 3px;
+  margin-right: 1px;
+  font-size: 8pt;
+  background-color: #7683a9;
```
ageis: I'd consider trying to find a new color to add to the whole palette, #7683a9 is a bit overused in SD.

fowlslegs:
I consider adding features like this like this (intentional redundancy) simply too dangerous. I'm ready to have a meeting about the reading room client and get started on it next week if you are. Let's do this the right way using a multi-party secure communication scheme, where we can provide a rich, featureful, native interface for with less worry.

redshiftzero:
Let's consider how SecureDrop is likely being used: if a point person is checking SecureDrop or multiple people are working with a specific source, these individuals have to exchange some basic ticketing-type information about these sources and submissions: e.g. “hey Alice, look at source N, it’s about the environment, can you get someone from your team to look at this”, “hey Bob, an initial response was sent to source N to let them know that we got their documents, can you follow up with them“, etc.

Perhaps this is done via email - which is actually quite bad given our stance that journalist devices are often easily compromised and thus any information about sources should _not_ be stored on them. In a situation where admins and journalists are distributed (or frequently traveling, working from home, etc.), they currently have no option but to discuss source matters like this using their regular devices and email, Signal, etc.

Perhaps at another news organization the admins decide that since all SD admins and journalists are in the same office they can set up a weekly face to face meeting without devices to discuss SecureDrop sources and submissions. In this case the administrators may decide that tag information is unnecessary: they simply never add any tags to the system, the journalists are unable to add tags, and in this case the journalist interface appears exactly as it was before. 

In this context, I think adding a feature like this enhances security by giving SecureDrop administrators the option of moving some of this high-level ticketing info onto the SecureDrop server. Documentation accompanying this feature should include some safe suggested tags as well as some examples of what not to do - we should explain _what_ we consider dangerous and _why_. 

Also, of course I agree that we should explore the reading room client (and the other SD 1.0 projects) further, but we should also be realistic about resources: rearchitecting a significant piece of SecureDrop — as well as making sure it is properly documented and tested — is a considerable undertaking, which even optimistically will take 6 months or more. Certainly it should be explored, but in the meantime, I think we should continue to merge PRs that improve SecureDrop and that have an appropriate balance between usability and security. 

Comments/thoughts from anyone else? @garrett @conorsch? Interested observer (that means you person reading this right now)?

fowlslegs:
> Let's consider how SecureDrop is likely being used: if a point person is checking SecureDrop or multiple people are working with a specific source, these individuals have to exchange some basic ticketing-type information about these sources and submissions: e.g. “hey Alice, look at source N, it’s about the environment, can you get someone from your team to look at this”, “hey Bob, an initial response was sent to source N to let them know that we got their documents, can you follow up with them“, etc.

> Perhaps this is done via email - which is actually quite bad given our stance that journalist devices are often easily compromised and thus any information about sources should not be stored on them. In a situation where admins and journalists are distributed (or frequently traveling, working from home, etc.), they currently have no option but to discuss source matters like this using their regular devices and email, Signal, etc.

This is a good argument, but I'm not sure that tagging alone is enough to supplant the communication that may be taking place across un-encrypted channels or on insecure endpoints.

On another note, journalists will probably think it's lame they can't create labels, and this will cause a headache for admins when they make requests for new labels. I think this should be a full access feature, if it is to be a web app feature.

I'm re-considering my initial opinion, and would like to look over this with you tomorrow in-person and discuss more.

securedrop/journalist.py
```diff
@@ -333,6 +349,162 @@ def edit_account():
...
+def admin_delete_source_label_type(tag_id):
+    delete_source_label(tag_id)
+    return redirect(url_for('admin_index'))
+
+
+@app.route('/admin/create_source_label_type/', methods=('POST',))
+@admin_required
+def admin_create_source_label_type():
+    text_to_display = request.form['text_to_display']
+    all_labels = [x.label_text for x in get_all_defined_source_tags()]
```
garrettr: [`.exists()`](http://docs.sqlalchemy.org/en/latest/orm/query.html#sqlalchemy.orm.query.Query.exists) would probably be helpful here.

securedrop/journalist.py
```diff
@@ -333,6 +349,162 @@ def edit_account():
...
+    if text_to_display not in all_labels:
+        db_session.add(SourceLabelType(label_text=text_to_display))
+        db_session.commit()
+    else:
+        flash('Tag already exists!', 'source')
```
garrettr: I'm pretty sure `source` is not a valid flashed message category. You probably want to use `error` here.

securedrop/journalist.py
```diff
@@ -333,6 +349,162 @@ def edit_account():
...
+
+
+@app.route('/admin/delete_submission_label_type/<int:tag_id>', methods=('POST',))
+@admin_required
+def admin_delete_submission_label_type(tag_id):
+    delete_submission_label(tag_id)
+    return redirect(url_for('admin_index'))
+
+
+def create_source_tag(source_id, label_id):
```
garrettr: During in-person code review, we discussed how this is not DRY - this basically re-implements the same CRUD functionality for SourceTags that was already implemented for SubmissionTags, but duplicates code instead of reusing it.

This is fine for now - DRY is a good principle but not an absolute one, and it's often easier to maintain code that hasn't tried to get too fancy and abstracted in the service of avoiding repetition.

That said, this is a common problem and there are a number of elegant solutions to it. In Django-land, the most common solution is generic views. For inspiration, check out these Django docs:

* [Class-based views](https://docs.djangoproject.com/en/1.10/topics/class-based-views/)
* [Generic views](https://docs.djangoproject.com/en/1.10/topics/class-based-views/generic-display/)

There are probably a ton of projects that do this kind of thing for Flask via extensions. This is probably worth researching further, but for now this code is ok and suitable for merge.

securedrop/journalist.py
```diff
@@ -333,6 +349,162 @@ def edit_account():
...
+    source = Source.query.filter(Source.id == source_id).first()
+    db_session.add(SourceTag(source_id=source.id,
+                             label_id=label_id))
+    db_session.commit()
+
+
+def delete_source_tag(source_id, label_id):
+    source = Source.query.filter(Source.id == source_id).first()
+    matching_tag = SourceTag.query.filter(and_(SourceTag.label_id == label_id,
+                                               SourceTag.source_id == source.id)).all()
```
garrettr: In general, these ORM queries read like queries written by an experienced SQL programmer (no offense!!) 😆 That being said, there is *probably* a more expressive and ORM-ish way to write this. In general, if we're going to use an ORM I think it's best to maximize its expressive potential.

This code is ok to merge, but I encourage you to research the ORM a bit more in depth to see if you can better leverage the ORM to make this query cleaner and more readable.

securedrop/journalist.py
```diff
@@ -333,6 +349,162 @@ def edit_account():
...
+    for tag in matching_tag:
+        db_session.delete(tag)
+    db_session.commit()
+
+
+@app.route("/col/add_source_label/<sid>/<label_id>", methods=('POST',))
+@login_required
+def add_source_label(sid, label_id):
+    source = get_source(sid)
+    create_source_tag(source.id, label_id)
```
garrettr: In general, since we are using an ORM, it is preferable to pass around objects returned from the ORM (representing rows in the database) instead of primary keys, like you are doing here.

securedrop/journalist.py
```diff
@@ -408,20 +580,16 @@ def remove_star(sid):
...
 
-    # Long SQLAlchemy statements look best when formatted according to
-    # the Pocoo style guide, IMHO:
-    # http://www.pocoo.org/internal/styleguide/
-    sources = Source.query.filter_by(pending=False) \
-                          .order_by(Source.last_updated.desc()) \
-                          .all()
     for source in sources:
         star = SourceStar.query.filter_by(source_id=source.id).first()
+        source_labels.update({source.id: get_source_tags(source.id)})
```
garrettr: I'm not sure how this could be better, but in general whenever I see SQL queries in a for loop I get a little concerned. There's often a better way to do this with a single query, which has better runtime characteristics (O(1) for the number of sources instead of O(n)).

securedrop/journalist.py
```diff
@@ -430,15 +598,85 @@ def index():
...
+                          .order_by(Source.last_updated.desc()) \
+                          .all()
+
+    starred, unstarred, labels, all_labels = get_stars_and_labels(sources)
+
+    return render_template('index.html', unstarred=unstarred, starred=starred,
+                           source_labels=labels, all_labels=all_labels)
+
+
+@app.route('/filter_by_label/<int:label_id>', methods=('POST',))
```
garrettr: In general, I like to think of views as having a 1-to-1-to-1 correspondence between the view's URL, view function, and template file. I think it makes the behavior of the web application more consistent and easier to modify.

Filtering a view is a classic case where GET parameters would be a good move. I would encourage you to:

1. Merge `filter_index` into `index` so the index view has a single view function
2. Supporting filtering via GET params, e.g. `?filter=foo,bar`. You'll have to take care to handle multiple filter arguments, and make sure you're url encoding/decoding properly.

garrettr: We discussed this functionality and you mentioned that it is currently only possible to filter by a single SourceTag at a time. It would be nice if you could filter by an arbitrary number of tags, which would be AND'ed together.

securedrop/journalist.py
```diff
@@ -430,15 +598,85 @@ def index():
...
 
 
 @app.route('/col/<sid>')
 @login_required
 def col(sid):
     source = get_source(sid)
+
+    # Tags on source
+    source_labels = get_source_tags(source.id)
+    selected_source_labels = [x.SourceLabelType.label_text for x in source_labels]
```
garrettr: Better to make this an extension of the query (`SELECT label_text FROM ...`) instead of using a list comprehension.

securedrop/journalist.py
```diff
@@ -430,15 +598,85 @@ def index():
...
+    all_source_labels = get_all_defined_source_tags()
+
+    unselected_source_labels = []
+    for possible_label in all_source_labels:
+        if possible_label.label_text not in selected_source_labels:
+            unselected_source_labels.append(possible_label)
```
garrettr: If you can do this in a DB query instead of a for loop, that would be better. I bet the ORM could make this clean and expressive, but I am not sure how.

securedrop/journalist_templates/admin.html
```diff
@@ -38,4 +38,54 @@
...
+<h2>Add/Delete Label Types</h2>
+<p><b>Operational Security Warning:</b> Note that these labels are stored in plaintext on the
+SecureDrop server.</p>
+
+<p>
+  <form id="source_tag_creation" method="post">
+    <input name="csrf_token" type="hidden" value="{{ csrf_token() }}" />
+    <input name="text_to_display" type="text" placeholder="Label Text" />
+    <button type="submit" class="btn" formaction="/admin/create_source_label_type/{{ text_to_display }}"><i class="fa fa-plus"></i> Add source label</button>
+    {% with errors = get_flashed_messages(category_filter=["source"]) %}{% if errors %}
```
garrettr: Nice! It's important to disambiguate handling this set of flashed messages from the set that is handled in `journalist_templates/flashed.html`, which is included in `journalist_templates/base.html`. Using specific categories is a good way to do this, but I would encourage you to use a more explicit name than `source`.

We reviewed `journalist_templates/flashed.html` and determined that it will only flash messages with category `notification` or `error`, so you are safe from the templates accidentally flashing the same message twice.

securedrop/db.py
```diff
@@ -411,6 +411,46 @@ def __init__(self, journalist):
...
+
+class SourceTag(Base):
+    __tablename__ = "source_tags"
+    id = Column(Integer, primary_key=True)
+    source_id = Column(Integer, ForeignKey('sources.id', ondelete='CASCADE'), nullable=False)
+    label_id = Column(Integer, ForeignKey('source_label_type.id', ondelete='CASCADE'), nullable=False)
+
+    def __repr__(self):
+        return "<Source tag ID={}: label_id={}>".format(self.source_id,
+                                                        self.label_id)
```
redshiftzero: Implemented in 8f3a1a2

securedrop/journalist.py
```diff
@@ -333,6 +349,162 @@ def edit_account():
...
+
+
+def get_all_defined_submission_tags():
+    """Get all defined submission tags"""
+    submission_tags = db_session.query(SubmissionLabelType).all()
+    return submission_tags
+
+
+def get_source_tags(source_id):
+    source = Source.query.filter(Source.id == source_id).first()
```
redshiftzero: Replaced with `filter_by` where possible in c12f4b0

securedrop/journalist.py
```diff
@@ -333,6 +349,162 @@ def edit_account():
...
+    source_tags = db_session.query(SourceTag, SourceLabelType) \
+                            .join(SourceLabelType) \
+                            .filter(SourceTag.source_id == source.id) \
+                            .group_by(SourceLabelType.label_text).all()
+    return source_tags
+
+
+def get_submission_tags(submission_id):
+    submission = Submission.query.filter(Submission.id == submission_id).first()
```
redshiftzero: Good point, I ended up removing these particular lines (by passing around the objects as suggested further on in the code review such that these queries were no longer necessary) but noted for the future and I checked I did not do this anywhere else in the diff.

securedrop/journalist.py
```diff
@@ -333,6 +349,162 @@ def edit_account():
...
+    submission_tags = db_session.query(SubmissionTag, SubmissionLabelType) \
+                            .join(SubmissionLabelType) \
+                            .filter(SubmissionTag.submission_id == submission.id) \
+                            .group_by(SubmissionLabelType.label_text).all()
```
redshiftzero: Yep this was a bad query: removed the entire thing by making better use of SQLAlchemy's `relationship` directive in 476830e

securedrop/journalist.py
```diff
@@ -333,6 +349,162 @@ def edit_account():
...
+@admin_required
+def admin_delete_source_label_type(tag_id):
+    delete_source_label(tag_id)
+    return redirect(url_for('admin_index'))
+
+
+@app.route('/admin/create_source_label_type/', methods=('POST',))
+@admin_required
+def admin_create_source_label_type():
+    text_to_display = request.form['text_to_display']
```
redshiftzero: Done in f875cf5

securedrop/journalist_templates/admin.html
redshiftzero: Yep renamed in 33ea94c

securedrop/journalist.py
redshiftzero: As we realized later on, this is a valid category but was not clear at all, renamed in 	33ea94c

securedrop/journalist.py
redshiftzero: Great points. I've merged these two views into one in a3b36fe, and filtering by an arbitrary number of labels via GET parameters is now implemented in eea97e6

securedrop/journalist.py
redshiftzero: Fixed in 10936d0

securedrop/journalist.py
redshiftzero: :+1: - unnecessary list comprehension removed

securedrop/journalist.py
redshiftzero: Haha no offense 😸 . Replaced with more ORM-ish query: 

`SourceTag.query.filter_by(label_id=label_id).filter_by(source_id=source.id).delete()`

securedrop/journalist.py
redshiftzero: Yep, totally. I've removed the query I added to this for loop and instead am using the `tags` attribute I added to Source and Submission models (using SQLAlchemy's `relationship` directive) to get the tags for the source.

securedrop/journalist.py
redshiftzero: Removed this query entirely just by using the `source.tags` attribute directly

securedrop/journalist.py
redshiftzero: Word - rewrote the db relationships such that the `source.tags` and `submission.tags` attributes are used to get the tags associated with each object. That moves most of the new logic from the views to the models 🤓 .

securedrop/journalist.py
redshiftzero: DRYed things up in commit fe6e1ec. 

Also nice point on the views: I researched this a bit and found the following for Flask:

- Flask's [Pluggable Views](http://flask.pocoo.org/docs/0.12/views/), the Flask version of Django's generic views
- [flask-classy](https://github.com/apiguy/flask-classy): a Flask extension to provide class-based views

I haven't used these in this PR but it's something I'm happy to look into for a subsequent PR

redshiftzero:
Thanks again for this feedback @garrettr!! It was really helpful. I've made improvements based on the code review and added comments inline describing how I addressed each one. Generally, I've DRYed up the code, made better use of the ORM, and have moved logic from the templates and views into the models and utility functions. Also, I've added functional and unit tests for the new functionality in this PR.

redshiftzero:
Also, here are some unnecessary gifs.

Here is the admin enabling tagging on the interface by adding source labels for her offices ("nyc", "sf", etc.) and a submission labels "response sent":

sd_tagging_admin

Journalist using the tagging as she does her work:

sd_tagging_journo

fowlslegs:
Do we want this in the 0.4 milestone?

fowlslegs:
Alright, I wanted to rebase this to play with some ideas I have on how we could keep tags encrypted-at-rest, and that ended up taking a couple hours, because there was a lot of tricky manual merging and reviewing of past PRs. Then I had to figure out how to convert your CSS changes to SASS (see https://github.com/freedomofpress/securedrop/pull/1422/commits/010e1988dc69d348dd9ad594d0ae594e77bce458 for the full story), which was a challenge because I barely know anything about either of them. Figured I may as well push the rebase here then to save you the future trouble.

We talked earlier today about how we needed to have more discussion of this PR and it's security implications. I do think it's great as a feature, and am of the overall opinion that it's one we should bring it to SD. That said, in addition to some ideas I have regarding how we can reduce some of the security implications of adding it, I think after test-driving it that the UX for this really needs to be re-evaluated. As we bring more user-facing features to SD I think it's important we be working more closely w/ UX folks, and I think @ninavizz would probably be willing to help us re-design the tagging UX here in a way that works much better and is compatible with our current design.

fowlslegs:
So just had a good talk with @ninavizz, and am also getting close to finishing my implementation of transparent encryption-at-rest for the label text in the `encrypted-at-rest-tagging` branch. Nina is going to get back to me next week with some wireframes for a labeling UX refactor, and I should be finished with my backend work by then, so I'll be able to start implementing whatever she comes up with.

We agreed in our convo that this needs to be a granular refactor of just the labeling implementation, as much as Nina would love to is very capable of improving myriad other JI elements. We also agreed the wireframes should fit in with our current styling--ensuring a granular pull request--while being "forward-looking," by which we mean that it can both bring us closer to her overall 1.0 redesign vision, and later become part of it with minimal modification.

ninavizz:
Addendum: "my design vision," thus far is only mine. As such, I have no attachment to it—and want to ensure everyone else gets that. Design is a team sport, with users it's intended beneficiaries—and thus far most of my work has been pretty contained to my own ideas, as that was the initial ask: "you're an outsider, let us know what you think this should be!"

Had a great chat with Noah, yesterday. Yep, will be making a few sets of reccos, with different degrees of technical impact to each. The existing UI has many 101-Heuristics usability issues going on, and I wouldn't be doing any justice to users or to you all, if those didn't additionally get flagged, w/ lowest-impact-possible solutions offered. Separate from a minimum-viable impact Tagging solution, offered.

redshiftzero:
Closing until the project is in a position to make changes like this

PR #1424, opened 2016-10-20
Extended explanation of why on-prem dedicated hardware is required

### LINKED ISSUE #1423: security implications of running SecureDrop on rented servers
I'm planning to setup SecureDrop for a small news organization that currently does not have a secure location for servers of its own.

Rented dedicated servers from a large, reliable provider seem like an acceptable option, since data from submissions is encrypted at rest. I see multiple warnings against this in the docs, but no rationale. **What are the actual security risks of having both SecureDrop servers hosted outside the organization?**

Might be good to elaborate on this topic in the docs, since many organizations are likely to consider the rented server option.

redshiftzero:
Resolves #1423

### DIFF
--- a/docs/hardware.rst
+++ b/docs/hardware.rst
@@ -22,28 +22,51 @@ These are the core components of a SecureDrop instance.
 -  **Network Firewall**: 1 physical computer that is used as a dedicated
    firewall for the SecureDrop servers.
 
-.. warning:: We are often asked if it is acceptable to run SecureDrop
-	     on cloud servers (e.g. Amazon EC2, DigitalOcean, etc.)
-	     instead of on dedicated hardware. This request is
-	     generally motivated by a desire for cost savings and
-	     convenience; however, cloud servers are trivially
-	     accessible and manipulable by the provider that operates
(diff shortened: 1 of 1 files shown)


fowlslegs:
We all went over this in Slack/ Google Docs before you submitted, and it's already been reviewed. It also builds fine. My only concern is that aesthetically it's a rather huge warning box. It will look better on docs.securedrop.org than the image I'm posting below because the stylesheet we use there (see the warning at the top of https://docs.securedrop.org/en/stable/development/documentation.html for example), but still maybe there's a way for us to still draw attention to this fact without putting multiple paragraphs in a yellow warning box. My only idea is to just not make it a warning at all--it's a large enough section that people can't miss it. If anyone else has other thoughts do share 

2016-10-19-224020_1181x1385_scrot

Note: I had to make FF fullscreen and zoom down to 67% just to make that image fit on my HiDPI screen.

redshiftzero:
Fair point - I've merged this into the main text in the section and removed the "Warning box". When people read the section on "Servers" they can't miss it.

PR #1425, opened 2016-10-20
Get sources to restart TB after logout

### LINKED ISSUE #1155: Reset identity via Torbutton both after using SecureDrop
It would be good to instruct everyone accessing the SecureDrop web interface to reset their identity after visiting the page. Even though Tor Browser does isolate most of it's state between sites and closes the rendezvous point circuit to once you've navigated away from a site, it's probably safest to do things this way because not everything has been wiped from memory. For example, if you press ctrl+shift+t (on Linux) after closing a tab, Tor Browser will re-open it. So it would be best practice to reset identity after using SD if you wish to continue using Tor Browsers.

redshiftzero:
Resolves #1155, adds a notification to encourage sources to restart Tor Browser when they log out, updates the docs to tell sources to do this, and also refreshes the screenshots in the source interface since the source interface has changed a bit (e.g. explicit Logout button was added, masking of the source codename). Here's the notification that appears:

logout-notification


fowlslegs:
Alright, I tested this manually locally, and we also have our unit test passing locally and on Travis. This was the implementation Garrett expressed preference for in #1155, and I agree. Merging.

PR #1426, opened 2016-10-20
Explain why a journo might want to rename a source

### LINKED ISSUE #558: Explain why a journalist would want to rename a source
We provide a "Generate a new codename for this source" button on each source's page, but there is no explanation of *why* a journalist would want to do this.

The idea here is to allow journalists to arrive at an easy to remember and communicate identifier, without the potential for leaking any information about the source or their submissions.

We should add language both to the journalist interface and to documentation explaining the purpose of this option.

Additional improvements to make here would be:

1. Move the button to generate a codename so it's next to the codename. I think it would work well if we just used the little "spinning arrows"/recycle icon, and it had a tooltip to explain what it does.

redshiftzero:
Closes #558


fowlslegs:
Lgtm. Going to go ahead and merge this one.

PR #1428, opened 2016-10-23
Use pytop's builtin option for valid window for TOTP codes

fowlslegs:
This commit replaces our code with functionally equivalent code already in the
pyotp library.


redshiftzero:
Just tested this, lgtm. Thanks @fowlslegs!

PR #1436, opened 2016-11-01
Merge 0.3.10 into develop

garrettr:
Resolves #1430.

The history for this pull request includes the commits from both 0.3.9 and 0.3.10. This is a bit confusing! After some investigation, we realized that the original pull request that merged the 0.3.9 changes into develop (#1411) squashed all of the the changes into a single commit, which failed to include the full 0.3.9 history in the develop branch. All of the _changes_ were included, just not the original commits. This was an exception to earlier practice. We can't remember why we did things differently for merging 0.3.9 into develop, but believe it was because it was a difficult merge and various git shenanigans ensued.

In a meeting, we agreed that it is preferable to include the full history of both 0.3.9 and 0.3.10 in develop in order to make the development history clear and hopefully make future merges from the 0.3.x release series into develop easier. Thus, this pull request includes the commits from both 0.3.9 and 0.3.10. However, the actual changes that are being applied are only from 0.3.10, and have been carefully reviewed by the team.

PR #1439, opened 2016-11-02
Version-locks Ansible in Travis tests

### LINKED ISSUE #1438: Version-lock Ansible in Travis tests
The automated tests in Travis currently run `pip install ansible` during the install phase, which will always install the latest stable version via pip. That's not good for our use case. Even though the Travis tests currently only run against the development environment, meaning it's not technically crucial to match the version of Ansible used on Admin Workstations (see #1146), we're still seeing [build failures due to syntax changes](https://travis-ci.org/freedomofpress/securedrop/builds/172420770) in upstream Ansible.

Let's update the Travis config to install a 1.x version of Ansible, as we recommend in the developer docs.

conorsch:
Rather than install the latest version of Ansible (currently 2.2.0.0), let's install an older version to match what we recommend for developers. Recent syntax deprecations in Ansible v2 cause the current playbooks to fail.

Confirmed a [working branch build](https://travis-ci.org/freedomofpress/securedrop/builds/172730427) based on these changes. Targeting the `merge-0.3.10-into-develop` branch so we can confirm passing tests on #1436 prior to merge.

Closes #1438.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_install:
 - echo HRNGDEVICE=/dev/urandom | sudo tee /etc/default/rng-tools
 - sudo /etc/init.d/rng-tools restart
 install:
-- pip install ansible
+- pip install ansible==1.8.4
 - pip install coveralls
 script:
 - echo localhost > inventory


redshiftzero:
Just tested this, lgtm. Merging..

fowlslegs:
Note Ansible 1.9.6 is the last version in the 1.X series. I've used it to provision every type of VM without error. It likely fixes bugs in 1.8.4 (that we've yet to trigger) without making any breaking changes like the 2.X series of releases would. What do you think about upgrading to this version @conorsch?

fowlslegs:
Ah, forgot about that one.

PR #1417, opened 2016-10-05
Improve unit test coverage of securedrop app code

redshiftzero:
This PR adds some more unit tests for the securedrop app code mostly for the admin interface which was an old "TODO" in the code (seen in the diff below). There are new tests for the functionality in `store.py`, `journalist.py` and `template_filters.py`.

Issue #47, opened 2013-10-09
Include a style guide for contributors


diracdeltas:
We should have consistent coding styles.

garrettr:
This is still an issue. We've had several pull requests come through recently that have fixed a number of PEP8 violations, which has helped. I think we should re-orient this issue to not only recommend a consistent coding style, but to enforce it. Therefore, the requirements for this issue should be:

- [ ] Make coding style adherence a requirement for contribution, and document the requirement in our contribution guidelines (`docs/develop.md`).
- [ ] Enforce this requirement via our CI pipeline. For example, Let's Encrypt checks code style as part of their [Travis automated tests](https://github.com/letsencrypt/lets-encrypt-preview/blob/master/.travis.yml#L20).

PR #1440, opened 2016-11-02
Delete submissions and replies when a source is deleted 

### LINKED ISSUE #1188: Delete Submissions that have had their Sources deleted

redshiftzero:
This PR is to merge in the work of @pwplus from #1190 which deletes submissions and replies when the source is deleted. I just added a unit test. Closes #1188.

### DIFF
--- a/securedrop/db.py
+++ b/securedrop/db.py
@@ -132,7 +132,8 @@ class Submission(Base):
         "Source",
         backref=backref(
             'submissions',
-            order_by=id))
+            order_by=id,
+        cascade="delete"))
     filename = Column(String(255), nullable=False)
     size = Column(Integer, nullable=False)
     downloaded = Column(Boolean, default=False)
@@ -158,7 +159,8 @@ class Reply(Base):
             order_by=id))
 
     source_id = Column(Integer, ForeignKey('sources.id'))
-    source = relationship("Source", backref=backref('replies', order_by=id))
+    source = relationship("Source", backref=backref('replies', order_by=id,
+                          cascade="delete"))
 
     filename = Column(String(255), nullable=False)
     size = Column(Integer, nullable=False)
--- a/securedrop/db.py
+++ b/securedrop/db.py
@@ -130,10 +130,9 @@ class Submission(Base):
     source_id = Column(Integer, ForeignKey('sources.id'))
     source = relationship(
         "Source",
-        backref=backref(
-            'submissions',
-            order_by=id,
-        cascade="delete"))
+        backref=backref("submissions", order_by=id, cascade="delete")
+        )
+
     filename = Column(String(255), nullable=False)
     size = Column(Integer, nullable=False)
     downloaded = Column(Boolean, default=False)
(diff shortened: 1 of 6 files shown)


securedrop/tests/test_unit_journalist.py
```diff
@@ -4,7 +4,7 @@
 from cStringIO import StringIO
 import unittest
 import zipfile
-
+import pdb
```
conorsch: Remove `pdb` import left over from debugging.

securedrop/db.py
```diff
@@ -132,7 +132,8 @@ class Submission(Base):
         "Source",
         backref=backref(
             'submissions',
-            order_by=id))
+            order_by=id,
+        cascade="delete"))
```
garrettr: Nit: formatting. Either put all of the parameter to `backref` on one line, or left-align them. If you can fit them all of one line, that is preferable.

securedrop/tests/test_unit_journalist.py
```diff
@@ -208,18 +217,29 @@ def test_edit_hotp(self):
...
+        journalist.delete_collection(source.filesystem_id)
+
+        # Source should be gone
+        results = db_session.query(Source).filter(Source.id == source.id).all()
+        self.assertEqual(results, [])
+
+        # Submissions should be gone
+        results = db_session.query(Submission.source_id == source.id).all()
+        self.assertEqual(results, [])
+
```
garrettr: I think you should also create a test reply as well, and verify that it is deleted as well (in order to test the change to `class Reply`).

securedrop/db.py
redshiftzero: Fixed!

securedrop/tests/test_unit_journalist.py
redshiftzero: Ah, good point - I have now written this in a separate unit test (and added a helper function to setup test replies in the database).

redshiftzero:
Thanks for your review @garrettr! Let me know if these changes don't address your concerns.

securedrop/db.py
```diff
@@ -158,7 +158,10 @@ class Reply(Base):
             order_by=id))
 
     source_id = Column(Integer, ForeignKey('sources.id'))
-    source = relationship("Source", backref=backref('replies', order_by=id))
+    source = relationship(
```
redshiftzero: I agree with your points here, and I definitely agree with you that the first is more readable than the second.  I’ll note that: 

```
    source = relationship("Source",
                          backref=backref("replies", order_by=id, cascade="delete"))
```

is not PEP8 compliant due to limits on characters per line. Instead I decided to opt for the ‘hanging intent’ style used in many of the tests in #1417 for readability. Happy to discuss further. 

Also: in order to address these kinds of style issues in the codebase, I suggest we tackle #47 and #886 soon and define and clean up the style in a few style-only PRs (I’m not going to change the lines you mentioned above in this PR since they’re not in the diff and I don’t want to confuse the history here, but the lines you point out are definitely something for a style PR once we’ve defined what constitutes good style for this project.)

redshiftzero:
So it looks like the PGP key is indeed being deleted when the source is deleted (done in `journalist.delete_collection`) but the deletion of the source documents can take a second due to the Redis worker. However I’ve added additional unit tests to explicitly check this behavior. 

Check it out and let me know if this doesn’t address your concerns @fowlslegs! Also rebased on current develop.

fowlslegs:
More good stuff @pwplus @redshiftzero!! Merging!

PR #1435, opened 2016-11-01
Ansible syntax update

fowlslegs:
Resolves #1427. Updates Ansible syntax, so it stops spitting out deprecation warnings at us. Will help in migrations to Ansible 2.

Note: I have not been able to test this comprehensively yet because doing so is blocking on https://github.com/freedomofpress/securedrop/issues/1430.

PR #1442, opened 2016-11-03
Wrap vars in with_items: statements

### LINKED ISSUE #1427: Turn off Ansible deprecation warnings or update to new syntax
Running the development playbook (all playbooks actually) spits out a whole bunch of deprecated syntax warnings. While Ansible 1.X still cooperates with this deprecated syntax, it would be better to update all the roles to Ansible 2. In the meantime, how do you feel about setting `deprecation_warnings=False` in `ansible.cfg`?

fowlslegs:
The use of raw vars in Ansible with_items statements has been deprecated because
of it's ambiguity: is a a string or a var?! To help make ourselves compatible
with future Ansible verions and get rid of some of those pesky deprecation
warnings (#1427), we're appropriately wrapping them in handlebars. Helps resolve
#1427.

### DIFF
--- a/install_files/ansible-base/roles/app-test/tasks/apparmor_complain.yml
+++ b/install_files/ansible-base/roles/app-test/tasks/apparmor_complain.yml
@@ -46,4 +46,4 @@
   # Doing it this way keeps the config DRY and will work with future profiles as well.
   when: >
     "/{{ item | regex_replace('\\.', '/') }}" not in apparmor_complaining_profiles_result.stdout_lines
-  with_items: apparmor_profiles
+  with_items: "{{ apparmor_profiles }}"
--- a/install_files/ansible-base/roles/app-test/tasks/dev_setup_xvfb_for_functional_tests.yml
+++ b/install_files/ansible-base/roles/app-test/tasks/dev_setup_xvfb_for_functional_tests.yml
@@ -9,7 +9,7 @@
   apt:
     name: "{{ item }}"
     state: latest
-  with_items: test_apt_dependencies
+  with_items: "{{ test_apt_dependencies }}"
   tags:
     - apt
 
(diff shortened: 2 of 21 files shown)


fowlslegs:
Supercedes #1435.

conorsch:
As I mentioned in https://github.com/freedomofpress/securedrop/pull/1435#issuecomment-258035641, these changes will need to be explicitly tested and confirmed working with:

* ansible==1.8.4 (via pip) in staging VMs
* ansible==1.7.2 (via apt) on Tails

If both environments honor the changes, we should be good for merge.

conorsch:
Ironically, Ansible warns about the allegedly superfluous presence of the handlebar syntax in with_items loops when run under old versions:

```
 [WARNING]: It is unnecessary to use '{{' in loops, leave variables in loop
expressions bare.
```

My, how times changed. While you're right that eventually we'll need to solve this problem to tackle #1146, it's not truly necessary at present, and the Ansible output will generate deprecation warnings no matter what we do (until #1146 is resolved). Still, I'm inclined to merge here—just need to repeat tests on hardware Tails.

fowlslegs:
Assigned @justintroutman @trevortimm because they're going to each be doing practice installs. I figured they may as well test this PR out since they'll be working with the hardware. Talk to me before you install SD to the app/mon server, and I'll help you test this.

conorsch:
In #1146 we've decided to push for upgrading Ansible to 2.x during the 0.4 release. Therefore we should merge this into develop and hammer out the config tests described in #1580 to make sure the playbooks respond intelligently to out-of-date versions of Ansible.

Waiting for #1563 to be closed out, to avoid complicating the backporting of 0.3.11 changes onto develop. I'll clean up any conflicts that arise in this branch as a result of the backport.

conorsch:
#1563 is closed out, so this should be good for merge. Related tasks for Ansible-v2 conversion tracked in #1146.

PR #1453, opened 2016-11-05
Hardcode the use of virtualbox in the Vagrantfile

ajvb:
Without this, a developer (such as myself) may unknowingly
use a different provider (e.g. vmware) where it is advised to
use virtualbox.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,6 +1,9 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
+# Make sure to use virtualbox
+ENV['VAGRANT_DEFAULT_PROVIDER'] = 'virtualbox'
+
 Vagrant.configure("2") do |config|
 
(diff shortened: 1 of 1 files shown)


conorsch:
@ajvb This approach overrides a support behavior of vagrant to reference the `VAGRANT_DEFAULT_PROVIDER` env var. If a developer has `VAGRANT_DEFAULT_PROVIDER=libvirt` specified, that provider should be used during `vagrant up`. Perhaps check if the env var is defined prior to setting, to avoid clobbering.

To be clear, we currently only support virtualbox as a local backend, but we should implement that support in a way consistent with vagrant's docs and existing tooling.

ajvb:
Per conversing about this in person, I'm going to close this PR.

VMWare is working just fine for me for development, so if the securedrop team decides they want something like this, further work could go into using Vagrant's built-in "choose this provider before this provider" in a cleaner way.

PR #1454, opened 2016-11-05
Adds temporary fix to time skew problems in development VMs

fowlslegs:
A temporary fix for the development VM playbook, that fixes a time skew problem that was preventing a contributor and myself from accessing the document interface while working at the Aaron Swartz day hackathon. This should be resolved in a nicer way sometime next week, but is needed to keep things running smooth for this weekend in the meantime.


ajvb:
This fixed the problem for me :)

PR #1455, opened 2016-11-05
Adding a github helper to add relevant labels to PRs

Hainish:


conorsch:
A description of how this tool is supposed to be used would be :ok_hand:.

fowlslegs:
Worked for me and seems idempotent. You can see the results, as submissions have been tagged appropriately. I used a personal access token with `public_repo` permissions.

fowlslegs:
`config.json` is added to the dir-level `.gitignore`, which should prevent an accidental push of the token to the repository.

fowlslegs:
Since this is only to be used by SD devs, we should not add documentation. @conorsch: first `cd` into the dir, then:
```
npm install
cp config.json.example config.json
# Generate a personal access token with public_repo access and add it to the empty relevant field of config.json
js index.js
```

github_helpers/index.js
```diff
@@ -68,7 +68,28 @@ function github_process_prs(most_recent_pr_checked, pr_cb){
...
+                if(file.filename.match(/^snap_ci\//)){
+                  labels.add("snap_ci");
+                }
+
+                if(file.filename.match(/^spec_tests\//)){
+                  labels.add("spec_tests");
+                }
+
+                if(file.filename.match(/^securedrop\/tests\//) ||
+                  file.filename.match(/^securedrop\/source.py$/)){
```
fowlslegs: Why are you adding the "tests" tag whenever `securedrop/source.py` is modified?

github_helpers/index.js
```diff
@@ -0,0 +1,134 @@
...
+                  labels.add("tests");
+                }
+
+                if(file.filename.match(/^securedrop\/tests\/functional\//)){
+                  labels.add("functional_tests");
```
fowlslegs: So functional tests will be labeled with both "tests" and "functional tests." I think just "tests" should be sufficient, and you can remove this label.

github_helpers/README.md
```diff
@@ -0,0 +1,32 @@
...
+    npm install
+    cp config.json.example config.json
+
+Enter your GitHub token info into `config.json`.
+
+## Running
+
+### With Docker
+
+Set your `$GITHUB_TOKEN`, and run
```
fowlslegs: I really don't like the idea of setting this as an environment variable. Doesn't seem safe.

github_helpers/index.js
Hainish: Good catch, I copied that line from above without changing the relevant file

github_helpers/README.md
Hainish: This is to illustrate that the below line should not be run as-is.  How is this any less safe than specifying it in a `config.json` file?

fowlslegs: My concern is some (honest) software reading in the value of `GITHUB_TOKEN`, and performing some action based on the fact it exists. I'm not a huge fan of the OS environment model, but I especially believe secrets should be kept out of it.

Hainish: I'm not completely sure I understand your concern, but I've changed the wording to remove the suggestion setting `$GITHUB_TOKEN` in the host env.

fowlslegs:
Looks good! Merging!

fowlslegs:
Thanks Bill!

PR #1462, opened 2016-11-06
Fix ansible-base path in virtual_environments.rst

ajvb:
Simple documentation fix.


redshiftzero:
Good catch, thanks @ajvb!

PR #1463, opened 2016-11-06
Adding .editorconfig (editorconfig.org)

Hainish:


conorsch:
You really don't like writing PR descriptions, do you? :stuck_out_tongue_winking_eye:

conorsch:
Could you add space/2 settings for `*.rb` files, as well? The serverspec tests are ruby, and even if they get converted to python toolkits in the future, we should still document the proper settings for formatting at present.

For bash, looks like we have mixed examples between 2 and 4, so let's leave that out for now.

Hainish:
Cool, I just added the convention for *.rb files

PR #1460, opened 2016-11-06
Create a more usable folder structure for downloads

ajvb:
This PR comes based on a conversation with @fowlslegs
and further information can be found here:
https://github.com/freedomofpress/securedrop/issues/383#issuecomment-258656017

Fixes #383

PR #1467, opened 2016-11-07
Download unread/ all selected sources

fowlslegs:
* Adds buttons/ functionality for downloading either all submissions or just unread submissions from one or more selected sources.
* Refactors download backend download functionality.
* Chooses better icons for buttons.
* Implements unit tests for this new functionality.

Since the git history is messy on this one, but I still want all 3 of us to get credit, I'm going to rebase this PR into 3 commits: one from each of us.


securedrop/tests/test_unit_journalist.py
```diff
@@ -7,15 +7,17 @@
...
 import crypto_util
 import journalist
 import common
-from db import db_session, Source, Journalist, InvalidPasswordLength
+from db import (db_session, Source, Journalist, InvalidPasswordLength,
+                Submission)
 
 # Set environment variable so config.py uses a test environment
 os.environ['SECUREDROP_ENV'] = 'test'
+import config
```
ajvb: Where is config being called? Or is this simply for making sure the flask settings are set?

fowlslegs: Currently, the `securedrop/config.py` modules reads the environment variable `SECUREDROP_ENV`, and sets the configuration differently based on its value. We'd like to deprecate the reading/ setting of environment variables in our code entirely, and instead be read from static INI files.

Although I'm still considering how to do this, I believe that we need to create a separate module for each INI file, for each environment setting. This will involve breaking up `securedrop/manage.py`. Each new module will read only a single INI file. All functionality it provides will involve calls to generators of other modules with initialization arguments, so that no other modules besides these are actually reading from a configuration file.

fowlslegs:
Want this one to get at least one more review @redshiftzero @garrettr

securedrop/tests/test_unit_journalist.py
ajvb: >Currently, the securedrop/config.py modules reads the environment variable SECUREDROP_ENV, and sets the configuration differently based on its value.

Ok awesome. Could I suggest leaving a comment explaining that its needed even though it is not directly called?

>We'd like to deprecate the reading/ setting of environment variables in our code entirely, and instead be read from static INI files.

👍 makes complete sense. Only thing to consider is that environment variables are much easier to work with when it comes to Docker / containers than INI files in my experience.

fowlslegs: Ah, I see now. We don't actually access any of the attributes in config directly in this module. I have to refresh my memory on what happens if we import some other modules which import config before setting this environment variable. I think that could be bad, but need to confirm.

fowlslegs:
Note: blocking on https://github.com/freedomofpress/securedrop/pull/1440.

securedrop/journalist_templates/col.html
```diff
@@ -85,7 +85,7 @@
     <input name="csrf_token" type="hidden" value="{{ csrf_token() }}"/>
     <input type="hidden" name="sid" value="{{ sid }}"/>
     <input type="hidden" name="col_name" value="{{ source.journalist_designation }}"/>
-    <button type="submit" class="danger"><i class="fa fa-times"></i> Delete collection</button>
+    <button type="submit" class="danger"><i class="fa fa-trash-o"></i> Delete Collection</button>
```
redshiftzero: Good call on changing these icons!

securedrop/tests/test_unit_journalist.py
```diff
@@ -415,27 +418,115 @@ def test_edit_hotp(self):
...
 
-    def test_bulk_download(self):
+    def test_download_selected_from_source(self):
         sid = 'EQZGCJBRGISGOTC2NZVWG6LILJBHEV3CINNEWSCLLFTUWZJPKJFECLS2NZ4G4U3QOZCFKTTPNZMVIWDCJBBHMUDBGFHXCQ3R'
         source = Source(sid, crypto_util.display_id())
         db_session.add(source)
         db_session.commit()
-        files = ['1-abc1-msg.gpg', '2-abc2-msg.gpg']
-        filenames = common.setup_test_docs(sid, files)
+        files = ['1-abc1-msg.gpg', '2-abc2-msg.gpg', '3-abc3-msg.gpg', '4-abc4-msg.gpg']
```
redshiftzero: You can adapt `add_source_and_submissions()` in #1440 (once it's merged in) here

fowlslegs:
Hmm, all tests were passing on my machine. I'll have to retry with a clean branch and VM.

fowlslegs:
Okay, still everything passes locally. I rebased and pushed again--we'll see if it passes on Travis this time.

fowlslegs:
@redshiftzero Can you pull this one down and test locally? I'm not sure why Travis is running into `sqlite3.OperationalError`s. Thoughts?

redshiftzero:
Hmm - all tests are passing locally for me as well...

redshiftzero:
Also: I see you added a commit with a major refactor of the test suite - can we maybe put this refactor in a separate PR? Else the merging of this PR will be delayed as those changes definitely warrants a careful re-review... up to you

fowlslegs:
@redshiftzero Problem is I'd have to rewrite the tests for download functionalities with the current test structures, which are now soon (hopefully) going to be outdated anyway. Alternatively, if you think this is an okay idea, we could also make an exception here and merge this without tests, since I've already written two passing* sets of tests for the functionality that you've been able to confirm--the ones I wrote pre-#1440 and the ones introduced in 0307265.

* The latter only seems to be passing locally.

I _really_ don't want to write a 3rd set of tests for this functionality, that is just going to be replaced days later by my 2nd set.

fowlslegs:
Maybe let's figure out Travis first?

redshiftzero:
Oh sorry - I wasn't suggesting writing of a third set of tests, instead just moving the last commit into a separate PR, rebase this one on develop as is since the first three commits are already reviewed and approved. Then we merge this one in (assuming we get a passing Travis build on the first three commits - which at least _was_ the case when I reviewed this last week).

fowlslegs:
I think rebasing will require me to rewrite tests because #1440 changed some things.

fowlslegs:
Ah, nvm.

fowlslegs:
Okay, that worked. My mistake. Split into two PRs.

fowlslegs:
@redshiftzero Ready for re-review.

fowlslegs:
Okay, I've gotten 3 approved reviews, and all 3 checks have passed. Going to merge this one myself in this case because I don't want to block #1460 any longer, which has already been open for 10 days.

PR #1471, opened 2016-11-12
Make securedrop PEP8-compliant (mostly) and add PEP8 pre-commit hook

ajvb:
This PR is purely style fixes.

It makes the entire `securedrop/` directory PEP8 compliant except for a few cases of `E402 module level import not at top of file` within the `test/` directory.

This is because of overriding config.py like so:
```
import sys
...
# Set environment variable so config.py uses a test environment
os.environ['SECUREDROP_ENV'] = 'test'
import config
```

As well, this PR creates the `git_helpers/` directory and adds a pep8 pre-commit hook.

One last thing, I would highly recommend adding CodeClimate (or similar) to do PEP8-on-PR checking per our conversation in #886. This will make sure that these style "problems" don't come back.

---------------------------------------------------------
Example pep8 pre-commit hook output:
```
$ git commit
PEP8 style violations have been detected.  Please fix them
or force the commit with "git commit --no-verify".

./securedrop/source.py:26:80: E501 line too long (85 > 79 characters)

```


securedrop/db.py
```diff
@@ -109,7 +108,8 @@ def documents_messages_count(self):
             for submission in self.submissions:
                 if submission.filename.endswith('msg.gpg'):
                     self.docs_msgs_count['messages'] += 1
-                elif submission.filename.endswith('doc.gz.gpg') or submission.filename.endswith('doc.zip.gpg'):
+                elif submission.filename.endswith('doc.gz.gpg') or \
+                        submission.filename.endswith('doc.zip.gpg'):
```
fowlslegs: I'm not sure if the spacing on this bottom line is right. Further, I think we should follow PEP8 advice here:

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

I also think think line breaks should come before binary operators (see https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator).

securedrop/db.py
```diff
@@ -184,7 +184,8 @@ class SourceStar(Base):
 
     def __eq__(self, other):
         if isinstance(other, SourceStar):
-            return self.source_id == other.source_id and self.id == other.id and self.starred == other.starred
+            return self.source_id == other.source_id and \
```
fowlslegs: Parens instead of backslash, break before binary operator.

securedrop/journalist.py
```diff
@@ -115,13 +113,17 @@ def login():
...
             else:
                 try:
                     user = Journalist.query.filter_by(
                         username=request.form['username']).one()
                     if user.is_totp:
-                        login_flashed_msg += " Please wait for a new two-factor token before logging in again."
+                        login_flashed_msg += (
+                            " Please wait for a new "
+                            "two-factor token before logging in again."
+                        )
```
fowlslegs: Could probably be 2 lines.

securedrop/journalist.py
```diff
@@ -186,8 +188,13 @@ def admin_add_user():
...
                 form_valid = False
-                flash("Your password is too long (maximum length {} characters)".format(
-                        Journalist.MAX_PASSWORD_LEN), "error")
+                flash(
+                    "Your password is too long "
+                    "(maximum length {} characters)".format(
+                        Journalist.MAX_PASSWORD_LEN
+                    ),
+                    "error"
+                )
```
fowlslegs: In this one it's not just for compactness--I think the spacing you have here actually hurts readability.

securedrop/db.py
```diff
@@ -131,7 +131,7 @@ class Submission(Base):
     source = relationship(
         "Source",
         backref=backref("submissions", order_by=id, cascade="delete")
-        )
+    )
```
fowlslegs: Not a fan of this C-style syntax to be honest. I prefer to fit more content in a page view, so long as it doesn't hurt readability. IMO the right-hand side is not so complicated that putting it into two lines would hurt readability. I do use it [here](https://github.com/freedomofpress/securedrop/pull/1467/files#diff-930c74950e483448f44f7d78910b23a8R155) for example, where the two vars in the brackets would otherwise be flush with the rest of the arguments, making it otherwise harder to tell they were nested in a list.

ajvb:
@fowlslegs Totally agree with the parenthesis over using backslashes, never saw that in PEP8.

In regards to the conversation that occurred in #886, what are your thoughts on:
1. Adding `pep8` checking to Travis build in this PR?
2. Change "required line length" to something more reasonable like 100-120?

fowlslegs:
1. I'd say yes. The general consensus at FPF seems to be we're in favor of making SD PEP8-compliant and enforcing that. One thing I just remembered is there is already a `pylintrc` file under `securedrop/`. You might want to look into modifying that and running `pylint` in Travis. I think it would be the way to go because it does a lot beyond PEP8. Not sure whether running that in Travis will make the results most accessible though, or what will be a failing condition as I've not used Pylint myself.
2. My personal vote is 99 based on a sample size of 2, and the fact I like odd numbers `乁໒( ͡◕ ᴥ ◕͡ )७ㄏ`

fowlslegs:
Okay, internal survey says FPF prefers 79/ official PEP8 decree.

conorsch:
Ouch, badly in need of a rebase. Thank you for writing this, @ajvb! We'll get it over the finish line once the test churn settles down. @redshiftzero @fowlslegs is now a good time to tackle rebasing this?

conorsch:
Ah, #1516 is still outstanding, let's get that in!

fowlslegs:
Putting this in the 0.4 milestone issue because the associated issue, https://github.com/freedomofpress/securedrop/issues/886, currently is. As noted there, big changes in `develop` will require a messy rebase, and we need to discuss if this is something we want in 0.4 in general.

fowlslegs:
I think we should either close this PR, or commit to reworking it for an upcoming release such as `0.4.1`.

ajvb:
@fowlslegs Agreed. Which would you advise?

redshiftzero:
After 0.4 is released, let's rework this and get it into 0.4.1 (and maybe add `flake8` to CI while we're at it)

fowlslegs:
Testing story here is relevant https://github.com/freedomofpress/securedrop/pull/1894#issue-239207662.

ajvb:
@redshiftzero Sounds good!

redshiftzero:
Following up now that 0.4 has been released. First, thanks again for your contributions @ajvb :gem:. Unfortunately, as we can see from the conflicts, the codebase has changed significantly since this PR was written. However since then, @dachary has gradually flake8'd almost all the code in `securedrop/` (just one last PR to go for that :tada:), and added `flake8` checks to Travis CI. 

However, I think the pre-commit hook in this PR would be really handy for developers. I've worked on a few projects that have `flake8` in CI and it's pretty annoying to forget to `flake8`, push up code, wait for tests to pass, only to have it fail due to linting. @ajvb would you be interested in reworking this PR or opening a superseding PR to add a `flake8 securedrop/` pre-commit hook?

redshiftzero:
Thanks again for your PR @ajvb. Unfortunately we are closing this PR in light of superseding PR #2237. 

If you are interested in trying to contribute to SecureDrop again, please feel free - you're welcome to drop by our [Gitter](https://gitter.im/freedomofpress/securedrop) or take a look at anything tagged "hackathon" (these tend to be smaller issues).

ajvb:
@redshiftzero Sounds good! My apologizes for being so unresponsive. Glad this was resolved nonetheless :)

PR #1474, opened 2016-11-14
Refactor tests/common.py

fowlslegs:
To be rebased after #1467 is merged!

Rebase + refactor tests for new download functionality
This rebases the new download unread and download all functionality in the index
view of the journalist app on the current develop, and in the process does some
minor refactoring of the test suite.

* Organizes imports: first come external package imports, then the
  SECUREDROP_ENV environment variable is set, and finally the SD module imports.
  They are now alphabetized. Also removes some unused imports, and additional
  addition or removal was done as necessary to support the refactor.

* Prefers use of APIs for object construction versus explicit assignment. For
  example, a call to :function:`crypto_util.genrandomid()` is preferred to the
  assignment `sid = 'somestring'`.

* Class-based rewrite of common.py: in order to better organize the
  functionality, and encourage better OOP and DRY practices, all the loose
  functions in `securedrop/tests/common.py` have become (mostly) static or class
  methods various organizational classes. This allowed us to write a lot of
  code.

* More accurate test method names: a lot of test method names were vaguely
  descriptive. A common motif was to just end test names in 'success.' It's
  better to name a method `test_admin_login_redirects_to_index` if it just tests
  that after successful login you're redirected to the index page, than
  `test_admin_login_success`. There are many things that should happen on
  success--:obj:`flask.g` should be set correctly, calls to the database should
  be made, session should be set in a client-side cookie, and that cookie should
  be set--the list goes on. Let's be specific about which thing that's supposed
  to happen happened successfully. (Side note: we do really need to do more
  comprehensive tests in a lot of places where we only tests that the app
  redirects correctly, instead of checking that all state and context is
  modified--or not--in the expected way).

* Better async support: The new :class:`tests.common.Async` provides a little
  bit better asynchronous functionality for a variety of situations, and
  deprecates the use of sleep statements.

* General cleanup: PEP8 stuff, and nicer formatting--some other minor touches.


securedrop/tests/common.py
```diff
@@ -1,120 +1,291 @@
...
 
-def clean_root():
-    shutil.rmtree(config.SECUREDROP_DATA_ROOT)
+        :param bool is_admin: Whether the user is an admin.
 
+        :returns: A 2-tuple. The first entry, the :class:`db.Journalist`
+                  initialized. The second, their password string.
+        """
+        username = crypto_util.genrandomid()
+        user_pw = crypto_util.genrandomid()
```
fowlslegs: Don't think it's a big deal in this case, but we can hard-code a seed with `random.seed()` in `common.SetUp.setup()`. I'd rather do that and make the API calls for everything rather than hardcoding in actual values in these cases, as it is a better way to test functions are working in conjunction with each other as expected. This was also, if any of the API calls we make change, we don't have to change every fixture such that it could be a state produced by calls to those functions.

redshiftzero: Yep, specifying a random seed sounds good 👍

redshiftzero:
Ping whenever this one is ready for re-review - happy to look at it

fowlslegs:
Unfortunately there's this bug that's only surfacing in Travis. I'm making some debugging commits I'm going to clean up later, but I don't know how better to do this, so it's happening.

fowlslegs:
Ready for review now.

securedrop/tests/common.py
```diff
@@ -1,120 +1,272 @@
...
+from functools import wraps
+import gnupg
+import mock
 import os
 import shutil
-import uuid
 import subprocess
-
-import gnupg
+import time
```
redshiftzero: [In PEP8](https://www.python.org/dev/peps/pep-0008/#imports), standard library imports come before third party imports (e.g. `time` before `gnupg`)

securedrop/tests/test_unit_journalist.py
```diff
@@ -2,356 +2,393 @@
...
 
         self._login_user()
         for admin_url in admin_urls:
-            res = self.client.get(admin_url)
-            self.assertStatus(res, 302)
+            resp = self.client.get(admin_url)
+            self.assertStatus(resp, 302)
 
-    def test_admin_authorization_for_posts(self):
+    def test_admin_page_restrction_http_posts(self):
```
redshiftzero: Spelling mistake: `test_admin_page_restriction_http_posts`

securedrop/tests/test_unit_journalist.py
```diff
@@ -2,356 +2,393 @@
...
+        self.assertIn('Add user', resp.data)
 
-        res = self.client.post(
-            url_for('admin_add_user'),
-            data=dict(username='dellsberg', password='not',
-                      password_again='thesame', is_admin=False))
-
-        self.assertIn('Passwords didn', res.data)
-
-    def test_admin_add_user_failure_password_too_long(self):
```
redshiftzero: Did this test go away?

securedrop/tests/test_journalist.py
```diff
@@ -67,44 +78,26 @@ def test_col_un_star_call_db(self, db_session):
...
-        journalist.request.form.__getitem__.return_value = action
 
     @classmethod
     def tearDownClass(cls):
         # Reset the module variables that were changed to mocks so we don't
         # break other tests
         reload(journalist)
 
-
+seshy = None
```
fowlslegs: No idea...

securedrop/tests/test_journalist.py
```diff
@@ -113,19 +106,16 @@ def tearDown(self):
 
     @patch('db.Journalist._scrypt_hash')
     @patch('db.Journalist.valid_password', return_value=True)
-    def test_login_with_valid_length_password_calls_scrypt(
-            self, mock_scrypt_hash, mock_valid_password):
-        self.login(self.username, self.password)
+    def test_login_calls_scrypt(self, mock_scrypt_hash, mock_valid_password):
```
fowlslegs: Just trying to make the convention of using `login` or, abstracted, just the action (here `login`) when the action will be successful, and only adding additional words to test names to when there is a problem with the action.

securedrop/tests/test_unit_source.py
```diff
@@ -80,90 +82,90 @@ def test_generate_has_login_link(self):
...
-            self.assertIn("Submit documents and messages", rv.data)
+            self.assertEqual(resp.status_code, 200)
+            self.assertIn("Submit documents and messages", resp.data)
             self.assertTrue(session['logged_in'])
-            common.logout(c)
+            resp = c.get('/logout', follow_redirects=True)
 
         with self.client as c:
-            rv = self.client.post('/login', data=dict(codename='invalid'),
+            resp = self.client.post('/login', data=dict(codename='invalid'),
```
fowlslegs: I think it works either way, but I can change it because it's not a sensible block.

securedrop/tests/common.py
fowlslegs: OooOOoo

securedrop/tests/common.py
```diff
@@ -1,120 +1,272 @@
...
-    for d in (config.SECUREDROP_DATA_ROOT, config.STORE_DIR,
-              config.GPG_KEY_DIR, config.TEMP_DIR):
-        if not os.path.isdir(d):
-            os.mkdir(d)
+    @staticmethod
+    def init_admin():
+        return TestJournalist.init_journalist(True)
+
+    @staticmethod
+    def mock_verify_token(self):
```
fowlslegs: Yeah.

securedrop/tests/test_journalist.py
redshiftzero: Maybe `test_valid_login_calls_scrypt`? There are multiple unit tests for `login` testing `scrypt` calls, good to be precise here

securedrop/tests/test_journalist.py
```diff
@@ -113,19 +106,16 @@ def tearDown(self):
...
+        Journalist.login(self.user.username, self.user_pw, 'mocked')
         self.assertTrue(mock_scrypt_hash.called,
                 "Failed to call _scrypt_hash for password w/ valid length")
 
     @patch('db.Journalist._scrypt_hash')
-    def test_login_with_invalid_length_password_doesnt_call_scrypt(
-            self, mock_scrypt_hash):
-        print "test_login_with_invalid_length_password_calls_scrypt"
+    def test_invalid_login_doesnt_call_scrypt(self, mock_scrypt_hash):
```
redshiftzero: :+1:

securedrop/tests/common.py
```diff
@@ -1,120 +1,272 @@
...
+        :raises: An :exc:`AssertionError`.
+        """
+        start_time = time.time()
+        while time.time() - start_time < timeout:
+            if job.result == cls.redis_success_return_value:
+                return
+            elif job.result not in (None, cls.redis_success_return_value):
+                assert False, 'Redis worker failed!'
+            time.sleep(0.1)
+        assert False, 'Redis worker timed out!'
```
fowlslegs: That makes sense when caller functions need to perform exception handling based on what goes wrong. In this case the test runner will just fail the unit test and the message to the developer will be sufficiently clear.

securedrop/tests/common.py
```diff
@@ -1,120 +1,272 @@
...
-# It has been intentionally omitted from `config.py.example`
-# in order to isolate the test vars from prod vars.
-# When refactoring the test suite, the TEST_WORKER_PIDFILE
-# TEST_WORKER_PIDFILE is also hard-coded in `manage.py`.
-TEST_WORKER_PIDFILE = "/tmp/securedrop_test_worker.pid"
 
+class TestJournalist:
+    """Test fixtures for working with :class:`db.Journalist`.
+    """
+    @staticmethod
```
redshiftzero: If [there's no need for `__init__()`](https://stackoverflow.com/questions/9100616/why-does-pylint-warn-about-no-init) and only static methods, why are these classes? Why is this better than [module-level functions e.g. in the relevant files](https://stackoverflow.com/questions/19620498/how-to-create-an-utility-class-correctly)? Apologies if I'm missing something here and this is a pattern I am not aware of

securedrop/tests/test_unit_journalist.py
fowlslegs: No I renamed it `test_admin_edits_user_password_too_long_warning` since it doesn't test that the operation actually fails (i.e., that the database record is unmodified)--it only checks a warning is shown. Think it's important to be specific about that sort of thing.

securedrop/tests/test_journalist.py
redshiftzero: But you added this, no?

securedrop/tests/common.py
fowlslegs: Nvm--answered too quick. It's supposed to be there. Look at how it's used in `test_journalist` and `test_unit_journalist`. Should probably rename the variable to `app`.

securedrop/tests/test_journalist.py
fowlslegs: Maybe it got accidentally pasted in there since it's out of context. Anyway, removed now.

securedrop/tests/common.py
fowlslegs: Because I find the namespacing helpful. The way I would change it is to break this into a few different modules: `tests.utils.async`, `tests.utils.journalist`, etc.. I find it organizationally useful and think it will encourage code reuse.

fowlslegs: One of the problems with the old `common` module is that it was really disorganized, and people weren't using it well (I presume partly for this reason), or were rewriting the same utilities elsewhere.

securedrop/tests/test_unit_store.py
```diff
@@ -47,19 +48,14 @@ def test_get_zip(self):
...
-        open(os.path.join(source_dir, old_filename), 'w').close()
-
-        new_filestem = 'abc2'
-        expected_filename = '1-abc2-msg.gpg'
-        actual_filename = store.rename_submission(sid, old_filename,
-                                                  new_filestem)
+        source, _ = TestSource.init_source()
+        old_journalist_filename = source.journalist_filename
+        old_filename = TestSubmission.submit(source, 1)[0].filename
+        new_journalist_filename = 'nestor_makhno'
```
redshiftzero: 🙀

securedrop/tests/test_unit_store.py
```diff
@@ -1,26 +1,29 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
+
 import os
 import unittest
 import zipfile
+
+from common import SetUp, TearDown, TestSource, TestSubmission
+import crypto_util
```
redshiftzero: `crypto_util` imported but not used

fowlslegs:
I'm going to split `common` up into a `utils` namespace, with several sub-packages instead of abusing `staticmethod`.

fowlslegs:
Ready for re-review.

fowlslegs:
Woops forgot to fix the functional tests as well. Impending commit!

fowlslegs:
Okay, this should pass Travis in a minute. Has been rebased on the latest develop at time of writing.

fowlslegs:
Could use another rebase just to clean up the git history, but I wanted to preserve it until this PR gets approved.

fowlslegs:
Rebased on the latest develop.

conorsch:
Can't get all the tests to pass locally on this branch; getting consistent failures on a redis worker timeout. Full test output here: https://gist.github.com/conorsch/bed571f440da3cfcb4d0edd8455c82ce 

Are you able to run the entire test suite locally and have all tests pass, @fowlslegs?

fowlslegs:
Yes. As a rule, I won't say a PR is ready for review unless it is passing all tests locally and on Travis.

redshiftzero:
Hmm... just provisioned a VM on this branch, ran the test suite a few times and everything seems to be passing for me (no Redis worker timeouts).

conorsch:
@redshiftzero Thanks for testing! That's good enough for me, I say good ahead and mark as approved. I'll step through the logs on my local VMs if it happens again. Given that you two and also Travis agree the tests are passing, we can be pretty confident these aren't breaking changes.

As for the actual subject of the PR, the separation looks clean, and in general the tests are a bit more readable. Hopefully these changes lower the bar for devs updating the tests in the future. 

LGTM! :checkered_flag:

redshiftzero:
Just gave the changes another look - this is much better, lots of good improvements. Let's get this merged in - @fowlslegs can you rebase and clean up the git history?

redshiftzero:
I see you've rebased, the git history cleaned up, and all checks are passing. Thanks @fowlslegs! I'mg going to merge this one in. If additional refactoring is needed in this part of the code, let's do it in another PR.

Issue #1477, opened 2016-11-15
Replace the Font Awesome icons with images


fowlslegs:
The context of this is that the Torbutton security slider provides a better user experience and greater exploitation protections than disabling JS alone, however, it breaks Font Awesome icons in SD as they are used currently. This is a legitimate decision to protect users on behalf of the TBB devs, as remote fonts/icons may contain malware.

See #1024 and specifically https://github.com/freedomofpress/securedrop/issues/1024#issuecomment-153211534, where @micahflee made the suggestion for this issue.

redshiftzero:
Note this was done for the source interface in #1480 but could also be done on the admin and journalist interface to remove this dependency. I used the SVG images from [this repo](https://github.com/encharm/Font-Awesome-SVG-PNG) to replace the icons.

ninavizz:
Many of the icons used in the interface (that I observed) are more distracting than helpful. In general, buttons should never have icons unless it's next/prev buttons as arrows, or the circle-arrow on a reload button. Buttons should always be either text or icons, but rarely both.

fowlslegs:
> Many of the icons used in the interface (that I observed) are more distracting than helpful. In general, buttons should never have icons unless it's next/prev buttons as arrows, or the circle-arrow on a reload button. Buttons should always be either text or icons, but rarely both.

Curious what you think of Github's use of icon + text buttons or Gmail's (see photos in https://play.google.com/store/apps/details?id=com.google.android.gm). Personally, I think in both cases what they've done both looks good, and is helpful to the user. When text is accompanied by icons, it can often help me find what I'm looking for faster than just scanning the text.

ninavizz:
@fowlslegs So: there's ALWAYS exceptions, to rules. My "never use icons on buttons" guidance, is a good rule to follow if you have no UX person on your team full-time. I'm looking at the "Open" button above, and that's done great. The icon itself is perfectly balanced with the text, and it appears to be a very faint green. 

Designers know how to spec these things out and make them work both to visually look rad, and to ease cognitive processes for users. I agree—when icons on buttons are done well, they're the best. They're too often done poorly, tho.

Like: I *can* do HTML and CSS, but there's no way I'd advise anything I fully coded to go into a live environment. Unless it's, like, HTML plopped into a Drupal site's page body. Even then, my HTML and CSS won't be able to reflow nicely.

ninavizz:
[...]

Issue #1476, opened 2016-11-15
Fix areas the Journalist Interface does not gracefully degrade when Security Slider is High


psivesely:
From some old issues made before I started here, we used to encourage the idea that some JS was acceptable as long as the interface gracefully degraded when it was disabled. There are some areas now where this doesn't quite seem to be the case.

JS is the `#1` tool for anyone wishing to exploit a browser. Security by compartmentalization is a fundamental component of the SD security model. We're kind of forcing users to compromise the strength of that compartmentalization when we make the interface inaccessible w/o JS, even though it's fair to say most would not disable JS, even if the JI did degrade gracefully, simply because they wouldn't consider it. We should still consider the users who are disabling JS on their JWs, but better we should fix this for all users long-term via the reading room.

I'm (re-)creating this issue to (i) create a list of some functionality that doesn't transition nicely between JS, and (ii) try to incorporate some practices for testing into the developer docs that will help contributors prevent the list in (i) from building up again.

The list (to be updated continually):
- [ ] The select all/ select none buttons on the JI index do nothing, but remain when JS is disabled.
- [ ] Replace font awesome icons with images instead of font characters in the JI as in https://github.com/freedomofpress/securedrop/pull/1480/commits/5e104459d890be8960cffe65d1fd5d0992811fb0.
- [ ] TOTP QR codes do not render when security slider is high

fowlslegs:
Okay, the situation is not as bad as I initially thought. After a brief reflection, I remembered testing my first PR involving making the codename hover-to-show in the source interface, and noticed that even that was broken in my testing environment. I used uBlock Origin to block scripts in Chromium, which I was using for test convenience (as now it seems TBB overrides custom network settings to allow direct localhost access every time you move the slider, and that gets annoying). It seems like uBlock Origin still goes above and beyond even the Torbutton slider on High in terms of breaking things--not going to look into exactly what all it's doing right now. That said, I think my point still applies, I'm just going to **edit the issue title** to reflect this mistake.

fowlslegs:
X-link https://github.com/freedomofpress/securedrop/issues/1544. It's up for debate whether we think that lack of "gracefully degrading" is a bug, but if we decide it's not we can't say we really support a Security Slider High compatible workflow on the JI, which is sad, and I believe we should.

redshiftzero:
While I think this is worth doing, I don't think it's high priority enough to justify putting into 0.4. can we move this off the 0.4 release?

[...]

PR #1480, opened 2016-11-29
Get sources to set Tor Browser security slider to High

### LINKED ISSUE #1024: Rethink javascript warning
Tor Browser 4.5 (and Tails 1.4) include much more fine-grained security settings than before. Rather than telling users to disable js using NoScript, I think it makes more sense to tell them to adjust to a more secure setting -- either High or Medium-High.

screenshot from 2015-05-13 19 23 50

redshiftzero:
SecureDrop’s source interface probably has one of the best use cases for the “High” security level of Tor Browser’s security slider. However, the icons we use looked bad/broken on the source interface using the “High” security level (described in #1024). This PR adds an explanation to the source guide telling them how to use the security slider, replaces all Font Awesome icons on the source interface with SVG graphics (as suggested by #1024 and #1477), and adds an explanation to the source interface if JavaScript is enabled to turn up the security slider to High. Note that turning up the security slider handles turning off JavaScript as well. I verified that nothing else breaks on the source interface other than the Font Awesome icons. Closes #1024.


fowlslegs:
Text on this page needs to be fixed. Otherwise, lgtm. 2016-12-02-113034_1517x890_scrot

Adding a task to https://github.com/freedomofpress/securedrop/issues/1476 to do the same for the JI.

redshiftzero:
Thanks for the review @fowlslegs! Updated and the page looks like this now: 

screen shot 2016-12-02 at 6 25 06 pm

fowlslegs:
We received an email:

>Does anyone know when this merged PR will make it into a release?
https://github.com/freedomofpress/securedrop/pull/1480

>Would it be possible for it to be kinda soon? :)

>At [redacted] we're doing a series a videos on digital security that
[redacted] are hosting. One of them will be part of [redacted]'s
[redacted], and it will be a video tutorial [redacted]: [redacted]
including how to use SecureDrop as a source.

>I'm writing the script for it and I've run into a usability issue. I
include visiting torproject.org in a normal browser, downloading Tor
Browser, opening it, and loading [redacted] SecureDrop. It has the big red bar
at the top that says, "We recommend disabling JavaScript to protect your
anonymity: Learn how to disable it, or ignore this warning to continue."
But the "Learn how to disable it" instructions are wrong (and only work
if you're security slider is already at low), and if I do change the
slider to high and reload all of the icons break. The PR fixes this.

>So the choice right now, for the script of this video, is to either
ignore the warning, or to heed the warning but not the way it suggests
and end up with broken icons. Both of those options kind of suck.

>I don't know exactly when we're filming, but probably in early February.
Would it be possible to get this PR in a release before then? If not, I
suppose worse case I could manually apply that PR to [redacted]'s instance.
But I'd rather avoid doing hacky things like that if I can help it.

If we are to make a 0.3.11 release, which seems probable and wise given our long list of TODOs for 0.4, I think we should try to include this PR. Thoughts @freedomofpress/securedrop-maintainers?

micahflee:
I just tested this PR, and unfortunately doesn't completely fix the issue.

- Unfortunately, it looks like (as of the latest Tor Browser) setting the security slider to High says, "Some types of images are disabled," and this appears to include `svg` images, which means that the images are still broken.
- It doesn't update `/etc/apparmor.d/usr.sbin/apache2` to allow apache2 to read all the new images in `static/i/font-awesome`

micahflee:
Also, it looks like some of the Tor Browser UI has changed, so the instructions text should be updated. It says to click the onion and choose "Privacy and Security Settings", but now it's just "Security Settings".

micahflee:
I also noticed a few other issues. Right now the buttons with icons in them are white on blue, and the css hover affect makes them blue on white. In order to retain this affect, each icon needs two versions of it, and the css needs to work in such a way that it switches to a different icon on hover.

Also, I noticed some other small style changes that seem to happen -- like, in `/generate` the button at the bottom has an icon and says "Continue", but when changing away from Font Awesome the icon is aligned differently, etc.

heartsucker:
@micahflee Could you screen shot this? I just made changes to these elements that made it into the mainline.

micahflee:
I took the svg files and converted them to 20x20 pngs. This is the index page (the button location, size, and icon vertical centering is all a bit off):

screenshot from 2017-02-06 21 10 08

And you can't see the mouse cursor in the image, but this is the index page when I'm hovering over "Submit Documents" -- the icon png is white on white, so you can't see it.

screenshot from 2017-02-06 21 10 43

Here's the "Continue" button on the generate page.

screenshot from 2017-02-06 21 11 06

This wasn't exhaustive at all, there could be some other issues too. But that was just what I quickly noticed. Everything should be checked thoroughly obviously.

heartsucker:
@micahflee Ah yes, one of @ninavizz's recommendation was actually to strip the icons from the buttons entirely, so those won't be there in the future (and aren't in current `develop`).

However, if SVGs don't work in TB with high security settings, then we might have some other problems as a few more have been added.

And as for the centering, it doesn't look like that in FF as of commit b83f2197 (and I can't get TB to talk to localhost, so I can't confirm there).

fowlslegs:
@heartsucker Once you figure out the legitimate issues and branches affected it would be awesome if you could file them in the tracker. I think you're more on top of the UI/UX changes project than me, so you're in a better position to do this. Thanks!

ninavizz:
Damn. I've been in touch with the TOR peeps on making that control more usable. :/ Will see what's going on with it. The slider should be a toolbar-button actionable control, the whole TBB setup as-is, stinks. I also had a minor meltdown when first new to Tor, trying to figure-out where that damn thing was.

ninavizz:
Sorry—this is my 2nd late-to-the-game request, but could we possibly *just* change the language of the text on this page, to the attached? The text as-is, is quite overwhelming. Especially the red text (which also, being red, looks like an inline error message—even tho the user hasn't done anything). If this request sends eyes rolling or knuckles whitening, ignore! Looking forward to getting UX stuff more synced w/ release schedules (it ain't supposed to be done as separate projects, cough). 

sd_generate

PR #1486, opened 2016-12-05
Improve usable folder structure for downloads

### LINKED ISSUE #383: Bulking downloading folder structure
I've noticed that in practice, it's really helpful if each message/document is in it's own folder. The reason is because in Tails, you double-click a .gpg file to decrypt it, and it creates the new file in the same folder. If it's a doc, then you right-click on the .zip to extract it to get the original file that was uploaded. Sometimes (if the source is savvy and follows the journalist key instructions) this is another .gpg file with a different name, and you double-click on that to decrypt it. And sometimes, it decrypts into a .zip or .tar.gz or something that the source sent, so you extract that. Etc, etc.

Basically, each message generates at least 2 files, sometimes 3 (if they post a gpg-encrypted message in the box). And each doc generates at least 3 files (.zip.gpg, .zip, and whatever is extracted from the .zip), sometimes way more. It gets really disorganized to have all of the files just sitting there in the same folder.

(Also, sometimes a source tries to upload the same document twice because they weren't sure if it worked the first time. If the two .zip.gpg files are in the same folder, you end up extracting the zip files over each other when the contents have the same filename.)

So anyway, I propose that the bulk download feature (and any future bulk download features, such as #372) shouldn't just put all the messages/docs into a zip file, but rather should build a directory structure like this:

```
| source_codename
|--| year-month-date
|--|--| doc_number
|--|--|--| document.gpg
```

So for example, if the source codename is "Ottoman mate" and they uploaded a message and then a document on May 22, the directory structure would be:

```
ottoman-mate
ottoman-mate/2014-05-22
ottoman-mate/2014-05-22/1
ottoman-mate/2014-05-22/1/1-ottoman_mate-msg.gpg
ottoman-mate/2014-05-22/2
ottoman-mate/2014-05-22/2/2-ottoman_mate-doc.zip.gpg
```

Make sense? I don't think the directory structure needs to be exactly like this, as long as it's organized sequentially and each message/doc is in it's own folder. (And the dates are super useful, especially if there's weeks of back-and-forth between journalist and source, to look back -- but the date could just be the date of last activity.)

redshiftzero:
This PR is to get in the work of @ajvb from the Aaron Swartz hackathon improving the usable folder structure for downloads. Supersedes #1460 and closes #383.

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -692,7 +692,7 @@ def download(zip_basename, submissions):
     attachment_filename = "{}--{}.zip".format(
         zip_basename, datetime.utcnow().strftime("%Y-%m-%d--%H-%M-%S"))
 
-    # Mark the submissions that are about to be downloaded as such
+    # Mark the submissions that have been downloaded as such
     for submission in submissions:
         submission.downloaded = True
     db_session.commit()
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -687,19 +687,16 @@ def download(zip_basename, submissions):
     :param list submissions: A list of :class:`db.Submission`s to
                              include in the zipfile.
     """
+    zf = store.get_bulk_archive(submissions,
+                                zip_directory=zip_basename)
+    attachment_filename = "{}--{}.zip".format(
+        zip_basename, datetime.utcnow().strftime("%Y-%m-%d--%H-%M-%S"))
+
     # Mark the submissions that are about to be downloaded as such
     for submission in submissions:
         submission.downloaded = True
     db_session.commit()
 
-    filenames = [store.path(submission.source.filesystem_id,
-                            submission.filename)
-                 for submission in submissions]
-
-    zf = store.get_bulk_archive(filenames,
-                                zip_directory=zip_basename)
-    attachment_filename = "{}--{}.zip".format(
-        zip_basename, datetime.utcnow().strftime("%Y-%m-%d--%H-%M-%S"))
     return send_file(zf.name, mimetype="application/zip",
                      attachment_filename=attachment_filename,
                      as_attachment=True)
--- a/securedrop/store.py
+++ b/securedrop/store.py
@@ -68,17 +68,27 @@ def path(*s):
     return absolute
 
 
-def get_bulk_archive(filenames, zip_directory=''):
+def get_bulk_archive(items_selected, zip_directory=''):
     zip_file = tempfile.NamedTemporaryFile(prefix='tmp_securedrop_bulk_dl_',
                                            dir=config.TEMP_DIR,
                                            delete=False)
+    sources = set([i.source.journalist_designation for i in items_selected])
+    # The below nested for-loops are there to create a more usable
+    # folder structure per #383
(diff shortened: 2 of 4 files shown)


redshiftzero:
Rebased on current develop. @fowlslegs take a peek at this one when you get a chance

redshiftzero:
Rebased 😇

securedrop/store.py
```diff
@@ -68,17 +68,26 @@ def path(*s):
...
+        for source in sources:
+            submissions = [s for s in items_selected if s.source.journalist_designation == source]
+            for index, submission in enumerate(submissions, start=1):
+                filename = path(submission.source.filesystem_id,
+                                submission.filename)
+                verify(filename)
+                zip.write(filename, arcname=os.path.join(
+                    zip_directory,
+                    source,
+                    "%s_%s" % (index, submission.source.last_updated.date()),
```
fowlslegs: This is not the index we want here. We want `submission.filename.split('_')[0]`.

securedrop/journalist.py
```diff
@@ -665,22 +665,19 @@ def download(zip_basename, submissions):
...
-    :param list submissions: A list of :class:`db.Submission`s to 
+    :param list submissions: A list of :class:`db.Submission`s to
                              include in the zipfile.
     """
+    zf = store.get_bulk_archive(submissions,
+                                zip_directory=zip_basename)
+    attachment_filename = "{}--{}.zip".format(
+        zip_basename, datetime.utcnow().strftime("%Y-%m-%d--%H-%M-%S"))
+
     # Mark the submissions that are about to be downloaded as such
```
fowlslegs: If you change the order of operations in this function, this comment uses the right tense.

redshiftzero:
(I'm waiting on merge of #1474 to make changes here)

fowlslegs:
You should just go ahead with this one.

redshiftzero:
ACK

securedrop/store.py
redshiftzero: It looks to me like `index` is the right variable in order to construct the filename directory as described in [this comment](https://github.com/freedomofpress/securedrop/issues/383#issuecomment-258656017).

securedrop/store.py
```diff
@@ -68,17 +68,26 @@ def path(*s):
     return absolute
 
 
-def get_bulk_archive(filenames, zip_directory=''):
+def get_bulk_archive(items_selected, zip_directory=''):
```
redshiftzero: Renamed to `selected_submissions` and docstring added

securedrop/journalist.py
redshiftzero: Comment changed!

redshiftzero:
OK fixed up based on your review @fowlslegs (thanks for the comments!), tests passing, and rebased on current develop. Should be good to go!

fowlslegs:
Needs rebase.

redshiftzero:
OK, rebased again.

redshiftzero:
OK removed `enumerate` everywhere and rebased on current develop, this one should be good to go!

PR #1492, opened 2016-12-09
Bump unit test coverage a few percent

redshiftzero:
Adds a new file of unit tests for `db.py` and adds no cover lines where appropriate

### DIFF
--- a/securedrop/tests/test_unit_db.py
+++ b/securedrop/tests/test_unit_db.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-import logging
 import os
 import unittest
 
@@ -13,9 +12,6 @@ import crypto_util
 from utils import db_helper, env
 from db import (db_session, Journalist, Submission, Source, Reply,
                 get_one_or_else)
-
-logging.basicConfig()
-logger = logging.getLogger(__name__)
 
 
 class TestDatabase(TestCase):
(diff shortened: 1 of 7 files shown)


heartsucker:
I think it might be worthwhile to write some super simple tests for the overridden `__repr__` functions instead of `no cover`-ing them because I've actually seen production bugs where some edge case calls a broken `repr` and everything blows up because it was never called for the like 8 months it was in the code base.

redshiftzero:
Ah, great point @heartsucker! I'll revise this.

redshiftzero:
OK simple tests added for `__repr__` lines (and `no cover` lines removed) for the reason @heartsucker mentions above.

(Also note that the coveralls check below is showing the difference since the last build on this branch, not from develop 😿 ... this PR increases coverage 4%)

securedrop/tests/test_unit_db.py
```diff
@@ -0,0 +1,97 @@
...
+from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
+
+import journalist
+import crypto_util
+import common
+os.environ['SECUREDROP_ENV'] = 'test'
+from db import (db_session, Journalist, Submission, Source, Reply,
+                get_one_or_else)
+
+import logging
```
fowlslegs: You should move this up to the top for PEP8.

securedrop/tests/test_unit_db.py
```diff
@@ -0,0 +1,97 @@
...
+import unittest
+
+from flask_testing import TestCase
+import mock
+from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
+
+import journalist
+import crypto_util
+import common
+os.environ['SECUREDROP_ENV'] = 'test'
```
fowlslegs: You don't need this if you're not importing `config`.

fowlslegs:
Needs rebase.

redshiftzero:
OK, rebased on current develop to use the utility functions, all changes made and mocked the logger. Should be good to go!

PR #1493, opened 2016-12-11
Added zxcvbn lib for password strength checking

### LINKED ISSUE #980: Add password validation during user creation

heartsucker:
Dropbox has the library `zxcvbn` that does password strength checking in a number of ways (dictionary attacks, 1337 sp34k, etc.) and gives an all around score (0-4). The app checks that this score is at least 3 before accepting a journalist password. Diceware (n>=4) and KeePass passwords score high enough to pass.

This PR **only** addresses new users/admins. It does not attempt to warn current users/admins that their passwords are weak. I could add something that forces password resets, but given the heightened focus on UX, I'm holding off on that until someone give a second opinion.

Mostly fixes #980

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -19,7 +19,8 @@ import template_filters
 from db import (db_session, Source, Journalist, Submission, Reply,
                 SourceStar, get_one_or_else, NoResultFound,
                 WrongPasswordException, BadTokenException,
-                LoginThrottledException, InvalidPasswordLength)
+                LoginThrottledException, InvalidPasswordLength,
+                WeakPasswordError)
 import worker
 
 app = Flask(__name__, template_folder=config.JOURNALIST_TEMPLATES_DIR)
@@ -188,6 +189,9 @@ def admin_add_user():
                 form_valid = False
                 flash("Your password is too long (maximum length {} characters)".format(
                         Journalist.MAX_PASSWORD_LEN), "error")
+            except WeakPasswordError:
+                form_valid=  False
+                flash("Your password is too weak. Try adding more words." 'error')
             except IntegrityError as e:
                 form_valid = False
                 if "username is not unique" in str(e):
@@ -274,6 +278,9 @@ def admin_edit_user(user_id):
                       "(maximum length {} characters)".format(
                       Journalist.MAX_PASSWORD_LEN), "error")
                 return redirect(url_for("admin_edit_user", user_id=user_id))
+            except WeakPasswordError:
+                flash("Your password is too weak. Try adding more words.", 'error')
+                return redirect(url_for("admin_edit_user", user_id=user_id))
 
         user.is_admin = bool(request.form.get('is_admin'))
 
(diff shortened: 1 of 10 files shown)


securedrop/journalist.py
```diff
@@ -188,6 +189,9 @@ def admin_add_user():
                 form_valid = False
                 flash("Your password is too long (maximum length {} characters)".format(
                         Journalist.MAX_PASSWORD_LEN), "error")
+            except WeakPasswordError:
+                form_valid=  False
+                flash("Your password is too weak. Try adding more words." 'error')
```
redshiftzero: Typo (missing comma): ` flash("Your password is too weak. Try adding more words.", 'error')`

redshiftzero:
This is a really nice PR @heartsucker: just pulled this down and took a look at this. It's a clean implementation, but there are a few broader issues to consider here: the first being that this adds the additional dependency of `zxcvbn`. Ideally we would avoid the additional of unnecessary dependencies. Thoughts  @fowlslegs @garrettr @conorsch?

While @fowlslegs and I were talking about this issue around the FPF water cooler, we thought having the server generate a passphrase in the same manner as we are already doing for the source codenames might be the best approach as described in [the comments](https://github.com/freedomofpress/securedrop/issues/980#issuecomment-92471105) on #980. This way no additional dependency is introduced and we can have a journalist user cycle their credentials after first login (since administrators will be generating their accounts and passing the initial creds to the journalist). It might also be useful to have a message "Store this passphrase in your KeePassX"  after the journalist's changes their password as well as a message "Securely transfer these credentials to your journalist user" on the administrator add user page.

fowlslegs:
@redshiftzero and I were talking about this one yesterday and decided it would be better to just generate Diceware passphrases for journalists. This:
1. Can be easily implemented in a way such that current users on systems that upgrade to a release that includes this feature won't need to change their passwords.
2. Guarantees sufficiently strong passwords as long as the server is not compromised (we have bigger problems to worry about in that case).
3. Would add way less code to our application and would not require adding yet another 3rd-party library.

fowlslegs:
I think we should close this one in favor of server-generated Diceware passphrases.

redshiftzero:
Well before we close, @heartsucker are you interested to modify this PR to implement the diceware passphrase generation described? Let us know your thoughts!

fowlslegs:
@redshiftzero I agree. Was going to wait in any case to hear from @heartsucker before proceeding, but it'd probably be easier to just open another PR.

heartsucker:
@fowlslegs @redshiftzero Yeah, close this. I can add the feature to auto-generate diceware passphrases in another PR.

conorsch:
> Ideally we would avoid the additional of unnecessary dependencies. 

:+1:

> Yeah, close this. I can add the feature to auto-generate diceware passphrases in another PR.

Roger that!

Issue #880, opened 2015-02-13
Require session timeout for Source Interface and Document Interface


runasand:
A journalist can be logged in to the Document Interface, remain inactive for 10 hours, and then continue working without the system requiring her to log back in again. The same is true for sources on the Source Interface. If possible, consider timing out sessions that have been inactive for a long time.

garrettr:
This was suggested and debated in an [earlier issue](https://github.com/freedomofpress/securedrop/issues/204). 

To recap: this is actually a little tricky. The current cookies are "session" cookies, which means they do not have an "Expires:" value set. Therefore, they only last as long as the user's browser session. The easiest way to time out sessions is to have the cookies expire. Unfortunately, this means they will no longer be session cookies, in which case Firefox will [write them to disk](https://dxr.mozilla.org/mozilla-central/source/netwerk/cookie/nsCookieService.cpp#4268). We intentionally use session cookies, which are stored in memory and never written to disk, as an anti-forensic measure.

It is possible that one of the Tor Browser patches might change this behavior in a way that could mitigate this forensic concern, but I haven't had time to look into it.

garrettr:
Note that the link to the Firefox implementation in the above comment is no longer exactly correct (although it is close) because the exact line number has changed due to modifications to the file since the last comment. If you are looking for the code, look around this comment:

```c
// if it's a non-session cookie and hasn't just been read from the db, write it out.
```

heartsucker:
@redshiftzero Hackathon candidate.

redshiftzero:
As described in [this comment](https://github.com/freedomofpress/securedrop/issues/880#issuecomment-76455408), this is not trivial, but #1494 would be a good starting point.

heartsucker:
Actually, my old change (#1494) doesn't alter the session/persistent nature of the cookies. It just adds and explicit timeout to the cookie. This covers the case where the browser is left open so that it will time out for the next request. I don't think @garrettr's comment applies to mine.

redshiftzero:
I took a look at how Tor Browser handles cookies. In Tor Browser, [`network.cookie.lifetimePolicy`](https://developer.mozilla.org/en-US/docs/Cookies_Preferences_in_Mozilla) is set to `2`:

<img width="784" alt="screen shot 2017-09-28 at 9 35 45 am" src="https://user-images.githubusercontent.com/7832803/30978782-70c28282-a430-11e7-857c-8bcdf07512de.png">

[...]

Issue #204, opened 2013-12-06
Flask cookies leak (server-side) session values


dolanjs:
The configured secret key, which is needed to use the flask session feature, is not used to encrypt the session values in the cookie, but only to sign them. This means that the cookie can be decoded to show the values in plain text.

PoC:

```
>>> cookie_str = "eJw9zMsOQ0AYQOFXaf61BVobSRcuGbVApC6Z2TSYMYohUYo23r3VRZcnOfneUPSUdZlgoL_hkIMORF0qYpkVVYmSW2aTpZrsOtqTpDMvHCTTS8OxSCb6e9BKrsYSRHiGTYLiMZS3sW9Y9-f8V6j5dVLhFInADlesxsfACU9-jYQvkMBRrBHbUILUVbwIL57NFRKezztXthnnjIJeZu2DSdD2e97uX30cJrZ9AOFDPgE"
>>> zlib.decompress(base64.urlsafe_b64decode(cookie_str+b"="*(-len(cookie_str) % 4)))
```

```
{"codename": {"b":"Z2xhZCBhd2Z1bCBkaW50IG5vZWwgcGF0dHkgYmVudCBhd2FyZSAxOTYw"},
"csrf_token":{" b":"NzQ5NjVhYWFmODQyY2U3OGQ4NjFmNmFmYTU5ZDA1OWI1MTYxMDg1ZQ=="},
"flagged":false,
"logged_in":true}
```

```
>>> base64.b64decode('Z2xhZCBhd2Z1bCBkaW50IG5vZWwgcGF0dHkgYmVudCBhd2FyZSAxOTYw')
'glad awful dint noel patty bent aware 1960'
```

There are two ways to mitigate this leak: One is to implement an own session interface that uses encryption for the (un)serialization and pass it to the app via the app.session_interface config. The second method is, to only store a session id in the cookie and create a server-side session store with files or a database.

Reported by crew53: SD-01-012 Flask cookies leak (server-side) session values

fowlslegs:
I'm working on EtM scheme for client-side session based on AES-CTR-128 and HMAC-SHA-224, since IMO that's the best option provided by our current crypto toolkit, PyCrypto. I'd love to at some point switch our crypto over to libsodium, but that's another story.

[...]

PR #1494, opened 2016-12-11
Added expiration to sessions

heartsucker:
A quick fix to #880. Also, this should be compatible with with the fix @fowlslegs is proposing in https://github.com/freedomofpress/securedrop/issues/204#issuecomment-260478199.

It works by manually setting a session expiration time in the cookie's data and then checks it before each request either extending it or logging the user out. Sessions are set to expire after 30 minutes.

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from datetime import datetime
+from datetime import datetime, timedelta
 import functools
 
 from flask import (Flask, request, render_template, send_file, redirect, flash,
@@ -65,6 +65,13 @@ def get_source(filesystem_id):
 @app.before_request
 def setup_g():
     """Store commonly used values in Flask's special g object"""
+    if 'expires' in session and datetime.utcnow() >= session['expires']:
+        session.clear()
+        flash(gettext('You have been logged out due to inactivity'), 'error')
+
+    session['expires'] = datetime.utcnow() + \
+        timedelta(minutes=getattr(config, 'SESSION_EXPIRATION_MINUTES', 30))
+
     uid = session.get('uid', None)
     if uid:
         g.user = Journalist.query.get(uid)
@@ -162,6 +169,7 @@ def login():
 @app.route('/logout')
 def logout():
     session.pop('uid', None)
+    session.pop('expires', None)
     return redirect(url_for('index'))
 
 
--- a/securedrop/source_app/__init__.py
+++ b/securedrop/source_app/__init__.py
@@ -1,3 +1,4 @@
+from datetime import datetime, timedelta
 from flask import (Flask, render_template, flash, Markup, request, g, session,
                    url_for, redirect)
 from flask_babel import gettext
@@ -75,6 +76,16 @@ def create_app(config):
         g.text_direction = i18n.get_text_direction(g.locale)
         g.html_lang = i18n.locale_to_rfc_5646(g.locale)
         g.locales = i18n.get_locale2name()
+
+        if 'expires' in session and datetime.utcnow() >= session['expires']:
+            session.clear()
+            flash(gettext('You have been logged out due to inactivity'),
+                  'error')
+
+        session['expires'] = datetime.utcnow() + \
+            timedelta(minutes=getattr(config,
+                                      'SESSION_EXPIRATION_MINUTES',
+                                      30))
 
         # ignore_static here because `crypto_util.hash_codename` is scrypt
         # (very time consuming), and we don't need to waste time running if
--- a/securedrop/tests/test_source.py
+++ b/securedrop/tests/test_source.py
@@ -202,13 +202,13 @@ class TestSourceApp(TestCase):
             login_test(' ' + codename + ' ')
             login_test(' ' + codename)
 
-    def _dummy_submission(self):
+    def _dummy_submission(self, client):
         """
         Helper to make a submission (content unimportant), mostly useful in
         testing notification behavior for a source's first vs. their
         subsequent submissions
         """
-        return self.client.post('/submit', data=dict(
+        return client.post('/submit', data=dict(
             msg="Pay no attention to the man behind the curtain.",
             fh=(StringIO(''), ''),
         ), follow_redirects=True)
@@ -221,7 +221,7 @@ class TestSourceApp(TestCase):
         """
         with self.client as client:
             new_codename(client, session)
-            resp = self._dummy_submission()
+            resp = self._dummy_submission(client)
             self.assertEqual(resp.status_code, 200)
             self.assertIn(
                 "Thank you for sending this information to us.",
@@ -230,7 +230,7 @@ class TestSourceApp(TestCase):
     def test_submit_message(self):
         with self.client as client:
             new_codename(client, session)
-            self._dummy_submission()
+            self._dummy_submission(client)
             resp = client.post('/submit', data=dict(
                 msg="This is a test.",
                 fh=(StringIO(''), ''),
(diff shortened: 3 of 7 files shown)


fowlslegs:
Sorry, but I already have implemented similar functionality in my branch, so I'm going to close this one.

fowlslegs:
Good PR though :smile_cat:

fowlslegs:
If you want to connect with us more on Gitter that would be awesome. We're trying to increase organization via Github comments, labels, & assignments in order to make collaboration with outside contributors easier without the need to rely much on either tools, but it's clearly an imperfect system, and still an organization-in-progress. Just noting this because we've closed a couple of your PRs that addressed "help wanted" issues that more organization and/or communication would've saved you from writing in the first place.

heartsucker:
Oh hai everyone. I'm back with this PR again. Rebased on develop.

Pros: sessions only good for 30 minutes

Cons: Exact login times leaked in the cookies. This only really matters for sources and could be mitigated with @fowlslegs suggestion about using some sort of AEAD to encrypt/authenticate the session.

dachary:
@heartsucker it would be good to have the timeout configurable for test purposes. The test could verify that the session expires after say 0.1 sec ?

trevortimm:
@heartsucker can you flesh out a little more exactly how the timing of log in leaked to the cookies could potentially affect (or not affect) sources?

dachary:
@heartsucker this implementation looks fine. I'm a little surprised this does not already exist in flask since it's a common pattern.

heartsucker:
@trevortimm There's a discussion on #204 that covers a lot of this. At the end of the day, if someone has the cookie, they have the codename which means they win. So having the exact time of a session isn't terribly important. Thus, the con is really a very minor one that I'm more pointing out as a way of making note that we have very minimally expanded the amount of information an attacker could receive rather than stating it as an argument against this PR.

On the source side, this mostly just protects against leaving a computer unattended. This is likely in that a long upload may mean they have to leave the computer unattended for a long time and popping their session could be very useful here.

heartsucker:
Also, this is strange. I don't get test failures locally. Travis did just update their trusty image today, so that could be related?

dachary:
@heartsucker looking at flask/sessions.py it seems possible to set permanent_session_lifetime to 30 minutes to get the same result. Am I missing something ? The **permanent** is the name of the variable is troubling ;-)

garrettr:
> looking at flask/sessions.py it seems possible to set permanent_session_lifetime to 30 minutes to get the same result. Am I missing something?

@dachary Yes, see https://github.com/freedomofpress/securedrop/issues/880#issuecomment-76455408.

Note that it's been a while since I've audited Firefox/Tor Browser's cookie handling code, and doing an up-to-date review is probably worthwhile. Also note that Tor Browser does things differently from Firefox, usually with an eye towards minimizing privacy risks/data leakage, so they may have recent (since 2015) patches that would help with this problem.

It would also be good for somebody to clearly articulate the threat model in this case (and yes, it's my bad for not having done so already). Here's a rough sketch:

1. When a user logs in to the source interface, they are given a signed session cookie that permits them to log in to their account.
2. This cookie does not expire. Cookie expiration is usually implemented with the `Expires:` key for HTTP cookies), but setting an expiration changes the cookie from a _session cookie_ to a _persistent cookie_. The last time I audited the Firefox cookie code (in 2015), session cookies are persisted in RAM and cleared when the browser exits, while persistent cookies are persisted on disk and deleted once they expire (originally discussed in #880 and others).
3. We generally try to minimize the chance of the Source's web browser persisting anything to disk, because we cannot reliably wipe the data in a way that prevents recovery through forensic disk analysis.
    * Note that there are a lot of problems with this approach in general. Since we do not control the copy of Tor Browser used by the Source, nor their operating system, we can only make a best effort.
4. On the other hand, @heartsucker and others are correct that cookies without an expiration are effectively "lifetime passes" for access to a Source's account, so if an adversary were somehow able to obtain one, it would give them unlimited access to the Source's account. **This is a good place to threat model, discuss how an adversary might realistically obtain a Source's session cookie, and then discuss the best way to mitigate that risk.**

heartsucker:
@dachary Missed your earlier comment. Yes, this could be configured or patched for testing. This was more meant as "reopen for review of functionality" thread and then if we decide this is something we want, I can polish it up.

dachary:
@heartsucker FWIW 💯 on implementing this session timeout :-)

heartsucker:
So the threat model with this feature would prevent capturing a cookie from being an unlimited access token, but the problem is that a source's codename is stored in the session. Once this is implemented, we can leave just a session token in the session use that for the current login behavior.

My proposal would be to merge this (already useful for the journalist) and then work on a scheme to alter the login behavior of the source.

heartsucker:
Proposed during the engineering meeting: look at Flask-Login as it's better supported than it was in the past and therefore is a candidate to offload code to something with more eyes than just SD. However, this depends on how Tor Browser handles cookies. It is only an option if TB treats all cookies as session cookies. Writing anything to disk is not acceptable.

redshiftzero:
Needs a rebase but if everyone agrees with [my logic here](https://github.com/freedomofpress/securedrop/issues/880#issuecomment-332894501), then I think this is a worthwhile incremental improvement.

heartsucker:
Pinging @redshiftzero for final review on this one.

securedrop/journalist.py
```diff
@@ -65,6 +65,13 @@ def get_source(filesystem_id):
 @app.before_request
 def setup_g():
     """Store commonly used values in Flask's special g object"""
+    if 'expires' in session and datetime.utcnow() >= session['expires']:
+        session.clear()
+        flash('You have been logged out due to inactivity', 'error')
+
+    session['expires'] = datetime.utcnow() + \
+        timedelta(minutes=getattr(config, 'SESSION_EXPIRATION_MINUTES', 30))
```
redshiftzero: It's good that this has a default of 30 minutes, but otherwise where is this config option defined? (x-linking with #1966). Or are you just assuming that it isn't defined?

heartsucker: Correct. I'm assuming it isn't defined, and I use `getattr` everywhere. Should someone really want to override it, it will pick it up. The unit/integration tests assume it's undefined and have checks to `del` the field in the `finally` block to not affect other tests.

securedrop/journalist.py
```diff
@@ -65,6 +65,13 @@ def get_source(filesystem_id):
 @app.before_request
 def setup_g():
     """Store commonly used values in Flask's special g object"""
+    if 'expires' in session and datetime.utcnow() >= session['expires']:
+        session.clear()
+        flash('You have been logged out due to inactivity', 'error')
```
redshiftzero: `gettext('You have been logged out due to inactivity')` 🇩🇪  🇸🇦  🇫🇷  🇳🇴

securedrop/source_app/__init__.py
```diff
@@ -76,6 +77,15 @@ def setup_g():
         g.html_lang = i18n.locale_to_rfc_5646(g.locale)
         g.locales = i18n.get_locale2name()
 
+        if 'expires' in session and datetime.utcnow() >= session['expires']:
+            session.clear()
+            flash('You have been logged out due to inactivity', 'error')
```
redshiftzero: `gettext` here too

(:eyes: on #2391)

securedrop/tests/test_source.py
```diff
@@ -368,47 +382,45 @@ def test_failed_normalize_timestamps_logs_warning(self, call, logger):
...
-        self.assertNotIn('logged_in', session.keys())
-        self.assertNotIn('codename', session.keys())
+            # Now the journalist deletes the source
+            filesystem_id = crypto_util.hash_codename(codename)
+            crypto_util.delete_reply_keypair(filesystem_id)
+            source = Source.query.filter_by(filesystem_id=filesystem_id).one()
+            db_session.delete(source)
+            db_session.commit()
 
-        logger.assert_called_once_with(
```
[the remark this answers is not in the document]
heartsucker: Lost in the rebase D:

heartsucker:
@redshiftzero Things are fixed, and I pushed enough times to kill transient CircleCI errors. :D

PR #1501, opened 2016-12-19
DRY up journalist interface index template using include

redshiftzero:
This PR is to use Jinja2's `include` statement to DRY up the journalist interface index. This prevents (almost) duplicating the code within the `for source in starred` and `for source in unstarred` loops. This is worth merging before e.g. #1422 adds more features to the journalist interface.

### DIFF
--- a/securedrop/journalist_templates/_source_row.html
+++ b/securedrop/journalist_templates/_source_row.html
@@ -1,0 +1,47 @@
+{% set docs = source.documents_messages_count()['documents'] %}
+{% set msgs = source.documents_messages_count()['messages'] %}
+<li class="source" data-source-designation="{{ source.journalist_designation|lower }}">
+  <time class="date" title="{{ source.last_updated|datetimeformat }}" datetime="{{ source.last_updated|datetimeformat(fmt="%Y-%m-%d %H:%M:%S%Z") }}">{{ source.last_updated|datetimeformat(relative=True) }}</time>
+  <div class="designation">
+    {% if source.star.starred %}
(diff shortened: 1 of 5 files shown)


conorsch:
Oh, this looks marvelous! Will test locally before giving :+1:.

redshiftzero:
Sorry, just realized I forgot to use the `starred` attribute in the template. I've fixed and rebased on latest develop. Also here's a gif of my testing it:

dry-journo

fowlslegs:
Okay, seems like 2 was fixed with that last push. And 3 was in fact never support (issue incoming). Can you confirm/ look into 1?

fowlslegs:
It would be good to have a functional test that tests the select all/none buttons work.

redshiftzero:
I’ve rebased on current develop and added a functional test for starring sources: `_journalist_stars_and_unstars_message()` (called in `test_submit_and_retrieve_happy_path`) to guard against future regressions.

It looks like 3 (now issue #1504) is actually implemented in `journalist.js` for the `delete_collection` and `delete_collections` classes. However it also looks like 1 and 3 were not introduced in this PR since they don't appear to be working on `develop` either. Investigating now, if someone can confirm they also see this that'd be 💯 . Either way more tests are needed here.

redshiftzero:
OK confirmed that these were existing issues on `develop`, but in this PR this is now fixed and I've also added functional tests for the "Select None" and "Select All" buttons to prevent regressions. Also the confirmation box upon clicking delete does appear now, as you can see in this gif that shows the JavaScript functionality on the journalist interface: 

fix_js

thus merge of this PR means we can also close #1504.

PR #1505, opened 2016-12-22
Update generate_securedrop_application_key.rst

mac-tech:
Prevents loss of the SecureDrop private key at first reboot.

### DIFF
--- a/docs/generate_securedrop_application_key.rst
+++ b/docs/generate_securedrop_application_key.rst
@@ -10,6 +10,15 @@ only be decrypted and read on the *Secure Viewing Station*.
 We will now generate the *SecureDrop Submission Key* on the
 *Secure Viewing Station*.
 
+Configure Persistent Volume
+---------------------------
+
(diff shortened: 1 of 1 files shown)


fowlslegs:
This information is already under https://docs.securedrop.org/en/latest/set_up_tails.html#enable-persistent-storage. If we were to add something to the SVS setup section of the guide to help ensure people didn't miss that, I believe it should be a warning at the beginning of the section to make sure to have set up persistence with a link back to that section.

mac-tech:
I think that reminder with link back would be a good idea. The recommendation to just turn all features in the tails setup guide is too general and does not convey which features are actually required for what purpose.

fowlslegs:
That sounds like a reasonable docs improvement. If you'd like to fix this PR to the issues we both just raised that would be :100:.

fowlslegs:
Closing this PR, but opened issue #1518. If you want to make one that more precisely resolves the issue and takes into account the existing documentation on creating persistent volumes you missed, do feel encouraged to do so! Thanks.

PR #1509, opened 2016-12-23
Force journalists to use diceware passwords

heartsucker:
This is the initial work for fixing #980 (superseding #1493). Forces journalists to use auto-generated diceware passwords and does not allow user input for passwords.

This PR is **not** ready for merge, but is here for initial functionality review before fixing all of the unit/functional tests.


securedrop/db.py
```diff
@@ -269,17 +274,20 @@ def _scrypt_hash(self, password, salt, params=None):
 
     MAX_PASSWORD_LEN = 128
 
-    def set_password(self, password):
+    @staticmethod
+    def check_password_acceptable(password):
         # Enforce a reasonable maximum length for passwords to avoid DoS
-        if len(password) > self.MAX_PASSWORD_LEN:
+        if len(password) > Journalist.MAX_PASSWORD_LEN:
```
heartsucker: I think this does need to be static so it can be preemptively used without creating a Journalist instance. As in, we check the password N times, then make one instance and set it.

Or is that not a good approach?

fowlslegs: > I think this does need to be static so it can be preemptively used without creating a Journalist instance.

You can do the same thing with a class method. The main advantage in this case is that, by writing `cls.MAX_PASSWORD_LEN`, even if the `Journalist` class where to change names, the method would still work w/o requiring you to replace `Journalist` in the method body with the new class name. In this case it's unlikely we'll change the name of the `Journalist` class, but I still think it's good practice. This SO answer http://stackoverflow.com/a/12179752 might give a little more info. on when it's better to use one or the other.

heartsucker: Ah yes, you're totally right.

heartsucker:
@fowlslegs In reply to 3, the edge case about a reboot / server failure between the password being changed and the flashed message being send to the client, I think the best thing to do would be to redirect a confirmation page before changing the password (but persist all the other changes). The admin would have to click "yes, confirm password change" on a page with text saying "this is the new password, here's where to save it." Or something to that effect. Then the password change is persisted, and a confirmation is flashed. Done.

I'm only asking because it seems like UI/UX is now becoming a priority, and I figured I should check in before making this change. @ninavizz might want to give input here too.

fowlslegs:
Maybe it is possible to use a JS `window.confirm` popup box if JS is enabled, or otherwise fall back to this confirm password change page? This is part of our larger make things nice easily with JS if it's enabled, but still provide full functionality--"graceful degradation" (see https://github.com/freedomofpress/securedrop/issues/1476)--if it's not.

heartsucker:
@fowlslegs: I made the changes. One thing doesn't work. I'm trying to use the JS to add a field to the form, but for the life of me I can't get it to work. Do you think you could take a peek? i think everything else works here.

fowlslegs:
If you think this is ready can you fix the existing tests and write new ones to thoroughly test the new functionality?

heartsucker:
Hey @fowlslegs, can i get some assistance with Selenium. Right now I have correct behavior in the browser, but I can't get it to work in the functional tests. Right now, I'm dying at `tests/functional/journalist_navigation_steps.py:291`. I have reset the password, but I can only log in with the old password. This does not happen in the browser. I've tried reloading the ORM object and a few other things, but it seems like the old password is... cached somehow?

heartsucker:
Just kidding. Nailed it. :D

fowlslegs:
Nice! I'll try to re-review Friday. Prob too busy tomorrow.

heartsucker:
@fowlslegs Whenever you get around to checking this, there was an error on the VM:

```
OSError: [Errno 39] Directory not empty: '/tmp/securedrop'
```

I think restarting the build should get this green.

fowlslegs:
Build restarted!

heartsucker:
@fowlslegs I'm rebasing this at the moment and think there might be a better solution than the JS `window.confirm` you proposed. I think it would be easier to implement something like the two step form like `#delete-all-confirm` on the source interface. It would be one less route, one less HTML page, and one less thing to test with Selenium. Good idea, bad idea?

fowlslegs:
@heartsucker Don't feel annoying pinging me if it's been a couple of days after I said I would review and I haven't. Everyone forgets.

Anyway, to answer your question, sounds like a good idea to me! Ping me when you've got that in and then I'll finally review.

heartsucker:
@fowlslegs No worries. I was busy with other tickets and would have been just endlessly rebasing this. :)

heartsucker:
@fowlslegs Yo yo. I think this new version is good to go. Can you take a peek?

fowlslegs:
Going to review tomorrow, so I can actually focus on getting some https://github.com/freedomofpress/fingerprint-securedrop work done today.

heartsucker:
@fowlslegs Rebased, but not manually checked since the 0.3.11 merge. I'll probably have time to get to that this weekend.

heartsucker:
This is tagged with the `0.4.2` label. Is this something I should rebase? Possibly after i18n lands?

dachary:
@heartsucker this pull request has a number of conflicting changes with i18n / l10n indeed. However I vote that we try to merge it regardless of the progress done with i18n. I volunteer to rebase myself in case there are conflicts so you don't have to carry the burden of doing that.

This is a large commit and it would be great if you could split it into smaller, self contained commits to facilitate review ?

redshiftzero:
I agree - let's try to merge this in, this is a nice feature that we should get into 0.4.2 (as [you point out](https://github.com/freedomofpress/securedrop/issues/2035#issuecomment-318908265) @heartsucker it also fixes #2035).

heartsucker:
Oh hai. I have this basically done except for one functional test that fails because using the admin interface to submit a password never seems to actually submit the password. I'm only passingly familiar with using the Selenium driver, so I think someone else might need to take a peek at this to tell me what I'm doing wrong. :/

`securedrop/tests/functional/journalist_navigation_steps.py:344` is where everything dies.

Annoyingly, this works when I run the app and in the other parts of the test suite, so I don't know what's happening here.

heartsucker:
Echoing what I said in the chat: this should be ready for review as it passes the test suite.

heartsucker:
Ping.

securedrop/db.py
```diff
@@ -274,20 +287,31 @@ def _scrypt_hash(self, password, salt, params=None):
         return scrypt.hash(str(password), salt, **params)
 
     MAX_PASSWORD_LEN = 128
-    MIN_PASSWORD_LEN = 12
+    MIN_PASSWORD_LEN = 14
```
redshiftzero: Why increase `MIN_PASSWORD_LEN`? (Not against it, just wondering)

securedrop/journalist.py
```diff
@@ -359,19 +346,53 @@ def admin_delete_user(user_id):
...
+    try:
+        user.set_password(password)
+    except PasswordError:
+        flash('You submitted a bad password! Password not changed.', 'error')
+        return
 
-    return render_template('edit_account.html')
+    try:
+        db_session.commit()
+    except Exception:  # pragma: no cover
```
redshiftzero: Hmm: is there a specific exception you are intending to catch here? Can we mock `db_session.commit()` to add a unit test instead of no cover?

securedrop/db.py
heartsucker: Because the minimum of 7 "diceware-like" is 14. `'a ' * 7`. Well, actually 13, but whatever. It's just for testing.

securedrop/manage.py
```diff
@@ -66,31 +66,50 @@ def add_journalist():  # pragma: no cover
...
-
-        if len(password) > Journalist.MAX_PASSWORD_LEN:
-            print('Your password is too long (maximum length {} characters). '
-                  'Please pick a shorter '
-                  'password.'.format(Journalist.MAX_PASSWORD_LEN))
-            continue
+        answer = raw_input("I understand that passwords are now "
+                           "autogenerated. (y/N): ").lower()
+        if answer in ('y', 'yes'):
+            return
```
heartsucker: Done.

securedrop/manage.py
```diff
@@ -66,31 +66,50 @@ def add_journalist():  # pragma: no cover
...
+    '''Function used to test suite mocking'''
+    while True:
+        answer = raw_input(
+                'Will this user be using a YubiKey [HOTP]? (y/N): ').lower()
+        if answer in ('y', 'yes'):
+            return True
+        elif answer in ('n', 'no'):
+            return False
+        else:
+            print 'Invalid answer. Please type "y" or "n"'
```
heartsucker: Done.

securedrop/journalist_templates/edit_account.html
```diff
@@ -2,47 +2,48 @@
...
-    <label for="password_again">Change password (confirm)</label>
-    <input name="password_again" id="password-again" type="password">
-  </p>
-
-  {# Only admin users may modify the admin status of other users. #}
-  {% if user %}
-  <p>
-    <input name="is_admin" id="is-admin" type="checkbox" {% if user.is_admin %}checked{% endif %}>
-    <label for="is_admin">Is Administrator</label>
+    <br><span id="password" class="password">{{ password }}</span>
```
heartsucker: Double `<br>` ftw.

securedrop/journalist_templates/edit_account.html
```diff
@@ -2,47 +2,48 @@
...
   </p>
-  {% endif %}
-
-  <button class="sd-button" type="submit" id="update">UPDATE</button>
+  <button class="sd-button btn" type="submit" id="reset-password">RESET PASSWORD</button>
```
heartsucker: Done.

securedrop/journalist_templates/edit_account.html
heartsucker: That's more ugly, actually. We can make it pretty in another PR.

securedrop/journalist.py
heartsucker: Done.

securedrop/tests/test_journalist.py
```diff
@@ -257,111 +260,92 @@ def test_admin_edits_user_password_success_response(self):
...
+                      password=overly_long_password),
             follow_redirects=True)
 
-        self.assertMessageFlashed('Your password must be between {} and {} '
-                                  'characters.'.format(
-                                      Journalist.MIN_PASSWORD_LEN,
-                                      Journalist.MAX_PASSWORD_LEN), 'error')
+        print resp.data.decode('utf-8')
+        self.assertMessageFlashed('You submitted a bad password! '
+                                  'Password not changed.', 'error')
```
heartsucker: Because there is nothing that should be able to happen on accident and all of the password submission/etc is outside of the user's control. More detailed messages won't provide actionable steps.

heartsucker:
Note: the function `commit_account_changes` is no longer called by the endpoints this PR affects. This is where the most of the coverage decrease comes from. The endpoint for `admin_edit_user` has tests. This should be addressed elsewhere.

securedrop/journalist.py
```diff
@@ -359,19 +346,53 @@ def admin_delete_user(user_id):
...
+    while True:
+        password = crypto_util.genrandomid(7)
         try:
-            edit_account_password(g.user, request.form['password'],
-                                  request.form['password_again'])
-        except (PasswordMismatchError, InvalidPasswordLength):
-            return redirect(url_for('edit_account'))
+            Journalist.check_password_acceptable(password)
+            return password
+        except PasswordError:  # pragma: no cover
```
heartsucker: Done.

securedrop/journalist.py
```diff
@@ -188,13 +184,10 @@ def admin_add_user():
                                       otp_secret=otp_secret)
                 db_session.add(new_user)
                 db_session.commit()
-            except InvalidPasswordLength:
+            except PasswordError:
+                flash('There was an error with the autogenerated passoword. '
```
redshiftzero: `passoword` -> `password`

redshiftzero:
Looks like a bug has crept in - I'm trying to make an admin from the command line and I get:

```
vagrant@development:/vagrant/securedrop$ ./manage.py add_admin
  File "./manage.py", line 409, in <module>
    _run_from_commandline()
  File "./manage.py", line 402, in _run_from_commandline
    rc = args.func(args)
TypeError: add_admin() takes no arguments (1 given)
```

(Relevant: #1982)

heartsucker:
Fixed the bug. Travis should be happy in a minute.

redshiftzero:
Cool - taking another look!

Also writing out here to summarize our discussion in person: since we've had a long back and forth in this PR, once the basic functionality and sufficient testing is down, I'm going to make two followup issues: one for design/UI cleanup and one for documentation. Note that we should tackle updating the documentation today when this is merged (even without updated screenshots, which we can do prior to release) so that any developers trying to work on SecureDrop are not confused by this new functionality.

redshiftzero:
Staging failure looks like a flake, restarting

redshiftzero:
Note: Coverage is decreasing here because @heartsucker has removed `no cover` from functions that were improperly excluded from the coverage computation in `manage.py` and added some test coverage for them. See coverage [output here](https://coveralls.io/builds/12943050). I am not going to block merge on this and further unit tests should be addressed in #1982.

PR #1510, opened 2016-12-23
Removes generated files in deb package

### LINKED ISSUE #1507: Make sure securedrop-app-code doesn't ship any .pyc files
Reported in https://github.com/freedomofpress/securedrop/issues/308#issuecomment-266144351. More historical info provided in https://github.com/freedomofpress/securedrop/issues/308#issuecomment-266169708. This probably just requires a small change in the app code deb build playbook somewhere to make sure it doesn't happen in the future.

heartsucker:
Fixes #1507

### DIFF
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/build_securedrop_app_code_deb.yml
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/build_securedrop_app_code_deb.yml
@@ -13,48 +13,60 @@
   tags:
     - pip
 
+- name: Clean build path
+  file:
+    path: /tmp/{{ securedrop_app_code_deb }}/
+    state: absent
+  tags:
+    - build
+    - cleanup
+
 - name: Copy install_files/securedrop-app-code dir to build path
-  command:  cp -R {{ securedrop_app_code_deb_config }} /tmp/{{ securedrop_app_code_deb }}/
+  command: cp -R {{ securedrop_app_code_deb_config }} /tmp/{{ securedrop_app_code_deb }}/
   args:
     creates: /tmp/{{ securedrop_app_code_deb }}/
   tags:
     - build
 
-- name: Create var directory in build path.
+- name: Create directories in build path.
   file:
+    dest: '/tmp/{{ securedrop_app_code_deb }}/{{ item }}'
     state: directory
-    dest: /tmp/{{ securedrop_app_code_deb }}/var/
-  tags:
-    - build
-
-- name: Create www directory in build path.
-  file:
-    state: directory
-    dest: /tmp/{{ securedrop_app_code_deb }}/var/www
-  tags:
-    - build
-
-- name: Create wheelhouse directory in build path.
-  file:
-    state: directory
-    dest: /tmp/{{ securedrop_app_code_deb }}/var/securedrop
+  with_items:
+    - /var
+    - /var/www
+    - /var/www/securedrop
+    - /var/www/securedrop/static
+    - /var/securedrop
   tags:
     - build
 
 - name: Copy securedrop to build path
-  command: cp -R {{ build_app_code_dir }} /tmp/{{ securedrop_app_code_deb }}/var/www/
-  args:
-    creates: /tmp/{{ securedrop_app_code_deb }}/var/www/{{ build_app_code_dir }}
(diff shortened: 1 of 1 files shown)


heartsucker:
Can someone reset the build? I got a network error.

```
TASK: [app-test | Download Firefox 46.0.1 for compatibility with Selenium 2.53.6.] *** 
failed: [localhost] => {"failed": true}
msg: failed to create temporary content file: The read operation timed out
```

redshiftzero:
Sure - build restarted!

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/build_securedrop_app_code_deb.yml
```diff
@@ -86,6 +86,17 @@
     - build
     - apparmor
 
+- name: Remove .pyc and other generated files
+  shell: >
+    (find /tmp/{{ securedrop_app_code_deb }} -name '*.pyc' -type f &&
+     find /tmp/{{ securedrop_app_code_deb }} -name __pycache__ -type d) &&
```
fowlslegs: Remove that closing paren on this line.

heartsucker:
@fowlslegs I'm also a fan of the constructive approach. Want me to scrap this version and do it that way?

fowlslegs:
That sounds good to me!

heartsucker:
So when looking at this, I noticed some annoying thing switching back and forth between the branches. There are some shell commands that use the `creates` argument to file copy or deb build steps. I think these should be removed, and there should be a clean step added to ensure everything is wiped before starting the build again. The idempotence of the current build is actually a bit problematic in that it makes it hard to work across multiple branches without manual intervention.

fowlslegs:
>So when looking at this, I noticed some annoying thing switching back and forth between the branches. There are some shell commands that use the `creates` argument to file copy or deb build steps. I think these should be removed, and there should be a clean step added to ensure everything is wiped before starting the build again. The idempotence of the current build is actually a bit problematic in that it makes it hard to work across multiple branches without manual intervention.

Or even the same branch after you've modified some files! Yes, you should add a step that deletes the `/tmp/{{ securedrop_app_code_deb }}/` directory if it exists before beginning to copy any files to the build path. Then you can get rid of those `creates` directives since they won't be doing anything.

In general, it is preferred to use Ansible modules when a suitable one exists over `shell` or `command`. You probably want to create a dict that maps files (or, regular expressions, such as `{{ build_app_code_dir }}/*.py`, whenever possible) to sub-directories of the `/tmp//{{ securedrop_app_code_deb }}`. I would look to the `copy` module and the `with_fileglob` and `with_dict` looping directives. Not positive you can use them together--as in putting `with_fileglob: {{ item.key }}` and `with_dict: {{ app_code_deb_mappings }}` in the same task--but that seems like a good way to do it. The mapping dict goes under `.../build-securedrop-app-code-deb-pkg/defaults/main.yml`.

heartsucker:
@fowlslegs I looked into this some more, and we can't use the `remote_src` in the `copy` module because it doesn't support recursive remote copies (also, it's `ansible` 2.0+). Using the `synchronize` module is a bit hideous because of doing a remote copy to self by delegating to self.

Last, using your suggestions leaves me a wee bit nervous as I have been [bitten a few times](https://heartsucker.com/blog/2016/06/15/how-ansible-took-down-our-servers) by trying to do even simple things with `ansible` loops. I'm pretty sure the behavior isn't consistent across versions.

Using shell / `cp` as above might with a single `with_items` not be a best practice for this but a) it definitely works, and b) it is unambiguous. I think for now, this is the most satisfactory solution.

conorsch:
It seems to me that a single task running something like `find -iname *.pyc* -delete` would go a long way here, and be easily maintainable.

fowlslegs:
It would be easier to just do `git clean` with w/e the option is that ignores local and global `.gitignore`s and that would get rid of the `__pycache__`, `.retry`, `.cache`, `.un~`, etc. files as well as the `.pyc`s. The reason a constructive approach is better is that we don't even want all the files in source control in the `securedrop/` directory in our app code.

I might try my hand at branching off this PR next week to try something if you don't mind @heartsucker.

heartsucker:
@fowlslegs Go nuts.

fowlslegs:
Closing in favor of #1612.

PR #1520, opened 2017-01-12
Add SourceClear integration to .travis.yml

### LINKED ISSUE #1519: Consider use of Github integration(s) to help us keep up on dependency management
Presently, we intermittently check on our Python dependencies to see if new versions are available. If so, we do our best to review the changes, and if we discover there are any vulnerabilities in package versions shipped in our latest app code deb, release a hot-fix. Ideally, we should be notified about updates to dependencies, and vulnerabilities in libraries we use ASAP. There exist several Github integrations, all with free plans for open-source projects that can do this for us. Foremost among them, in my opinion, from my limited research, is [SRC:CLR](https://github.com/integrations/src-clr), though [VersionEye](https://github.com/integrations/versioneye) and [WhiteSource](https://github.com/integrations/whitesource) are also considerations.

These integrations should not serve as a replacement for manual review of dependency updates, but, in my opinion, as a necessary supplement.

fowlslegs:
Resolves #1519 (if we like it--needed to create this PR to test!).

### DIFF
--- a/.srcclr.json
+++ b/.srcclr.json
@@ -1,0 +1,3 @@
+{
+   "pipRequirementsFile": "securedrop/requirements/securedrop-requirements.txt"
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,10 +32,10 @@ script:
 - sudo chown root:root securedrop/config.py
 - sudo sh -c "export DISPLAY=:1; cd securedrop && ./manage.py test"
 after_success:
-  cd securedrop && coveralls
-notifications:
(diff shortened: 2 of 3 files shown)


fowlslegs:
Results from running with CLI tool locally:

```
~/src/fpf/securedrop(sourceclear-integration|✔)
% srcclr scan
Unable to scan and generate report for directory "/home/fowlslegs/src/fpf/securedrop"
java.lang.String cannot be cast to java.io.File
```

PR #1521, opened 2017-01-13
Remove source choice in length of codename

### LINKED ISSUE #1508: Remove source choice in number of words in codename
I think it would look a bit nicer and streamline the user experience if we got rid of the option to choose the number of Diceware words in the generate view for sources. I don't see a reason to make the codename longer than 7 words considering the strong scrypt parameters we've chosen. I assume it can be a confusing option for sources who don't understand how Diceware serves as an entropy encoding mechanism or how their codename fits into the security model of SD (i.e., most all sources).

redshiftzero:
This removes the source choice in length of codename to make the source interface less confusing:

source-step4

Suggested by @fowlslegs and 👍ed  by UX MVP @ninavizz. Closes #1508.


securedrop/tests/test_unit_source.py
```diff
@@ -15,7 +15,7 @@
 import utils
 
 
-class TestSourceApppp(TestCase):
```
fowlslegs: Appppppppppppppppppppppppppppppppppppppppppppppppppppppppppppps!

Issue #98, opened 2013-10-24
Allow sources to contact specific journalists


micahflee:
I had a discussion with Dave Maass, media relation coordinator at EFF. He pointed out that some sources might contact a specific media organization because they want to get in touch with a specific journalist, and don't want anyone else to have access.

A way that this could work is:

Each journalist can have their own airgapped viewing station computer, and their own Tails USB stick with their own PGP key that only they know the passwords to.

Whenever a source sends a message or uploads a document, they get to choose which journalists have access to what they're about to send. The default is all journalists, and each time the source logs back in the checkboxes are in the same state as the last time the form was submitted.

When journalists login to their Tor hidden service URL, they only see sources that have chosen to encrypt information to them.

Media orgs should also be able to choose to just have one public key and share it (like how SecureDrop currently works), and the interface should be slightly cleaner in that case.

garrettr:
If this is ever implemented, we should hide the recipients as suggested in #13. See https://github.com/freedomofpress/securedrop/issues/13#issuecomment-86202424 for context.

ninavizz:
Yo Peeps... now that I see this functionality has been added for the 0.4 release, I figure this should get some feedback from those working on it—as the existing Journalist UI is... err, problematic, and it'd be nice to see this elegantly folded-in to an updated UI.

[SD_JournoOhDotFour-01g.pdf](https://github.com/freedomofpress/securedrop/files/746687/SD_JournoOhDotFour-01g.pdf)

redshiftzero:
Feedback added in [this comment](https://github.com/freedomofpress/securedrop/issues/1540#issuecomment-277035572).

PR #1523, opened 2017-01-17
Enable journalists to upload their GPG public key

redshiftzero:
This is a first step toward #98 (enabling sources to submit to specific journalists). In order to have sources select the journalists that they want to send their document to, we need in the SD database 1) more than just a username to show sources: probably we need for each journalist at least their name and a brief blurb about what they work on and 2) we need their public keys. 

This PR adds these to the database and adds a couple of new routes enabling a journalist to submit: their PGP key, full name, and a bit of profile text that they can use to describe themselves to potential sources (e.g. "Crime Reporter", "National Security Reporter", etc.). The intention here is that this information (name and blurb) would eventually be displayed to sources that want to send documents directly to a specific person and then on the backend documents can be encrypted to the journalist's public key. Journalists are able to get to their profiles by clicking the new link "Profile" in the upper right corner of the interface:

profile-linky

Their PGP key, name, blurb now appears on the journalist interface under a new profile route `/profile/<username>`:

profile

They can click "Edit my profile" to modify any of this information:

edit-profile

Logged-in users are able to look at one another's profiles if they are so inclined (but are unable to edit them as the "Edit my profile" link will not appear). 

The profile page also displays the sources that the journalist is assigned to. This enables journalists to remind themselves what they are assigned to as well as potentially for collaboration (they can quickly see what one of their colleagues is assigned to). 

Will add more tests before merge if this general approach is useful/reasonable, thoughts/comments/questions/concerns welcome

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -14,6 +14,7 @@ from sqlalchemy.exc import IntegrityError
 import config
 import version
 import crypto_util
+import forms
 import store
 import template_filters
 from db import (db_session, Source, Journalist, Submission, Reply,
@@ -27,6 +28,7 @@ app.config.from_object(config.JournalistInterfaceFlaskConfig)
 CsrfProtect(app)
 
 assets = Environment(app)
+
 
 app.jinja_env.globals['version'] = version.__version__
 if getattr(config, 'CUSTOM_HEADER_IMAGE', None):
@@ -407,6 +409,34 @@ def remove_star(sid):
     make_star_false(sid)
     db_session.commit()
     return redirect(url_for('index'))
+
+
+@app.route('/profile/<username>')
+@login_required
+def profile(username):
+    user = Journalist.query.filter_by(username=username).one()
+    sources = Source.query.filter_by(journalist_id=g.user.id).all()
+    return render_template("profile.html", user=user, sources=sources)
+
+
+@app.route('/edit-profile/', methods=['GET', 'POST'])
+@login_required
+def edit_profile():
+    form = forms.EditProfileForm(request.form)
+    if request.method == 'GET':
+        # Populate the inputs with the existing values in the database
+        form.full_name.data = g.user.full_name
+        form.about.data = g.user.about
+        form.pgp_key.data = g.user.pgp_key
+    elif request.method == 'POST' and form.validate():
+        user = Journalist.query.filter_by(username=g.user.username).one()
+        user.full_name = form.full_name.data
+        user.about = form.about.data
+        user.pgp_key = form.pgp_key.data
+        db_session.commit()
+        flash('Successfully edited your profile!', 'notification')
+        return redirect(url_for('profile', username=g.user.username))
+    return render_template("edit_profile.html", form=form)
 
 
 @app.route('/')
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -415,7 +415,7 @@ def remove_star(sid):
 @login_required
 def profile(username):
     user = Journalist.query.filter_by(username=username).one()
-    sources = Source.query.filter_by(journalist_id=g.user.id).all()
+    sources = Source.query.filter_by(journalist_id=user.id).all()
     return render_template("profile.html", user=user, sources=sources)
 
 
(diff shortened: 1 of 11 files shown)


redshiftzero:
It sounded like from AFK discussion that this is a direction that is useful so I've added unit tests for it and added a feature enabling administrators to edit this information in the database in the not-that-unlikely-situation that journalists are slow in putting their profile information and GPG keys on the SecureDrop server. An administrator can just go in and add this information for the users herself by going to the admin interface and clicking on the "Edit Profile" link of the user she wishes to edit:

new-edit-profile-link

In addition, a link now appears on the profile page of any user if the person viewing the page is an administrator. The link directs them to the edit profile link for that user:

admin-link-edit-profile

(Note: I followed the same approach taken during the creation of `admin_edit_user` and `edit_user` to make the profile edit routes for regular users and administrators separate.)

fowlslegs:
Do we want this in the 0.4 milestone?

redshiftzero:
I'm fine with pushing this until a future release in order to keep 0.4 constrained to shipping develop.

redshiftzero:
Hey @fowlslegs saw you were taking a stab at reviewing this. Note there is one remaining to do here (sorry, just took another look at this after having eyes off it for a couple of months): I need to add validation on `username` in `profile()`. Let me know if you notice anything else. Merge of this is a few releases out due to requiring database migrations support, but any comments either in general or detail are very welcome!

conorsch:
> Merge of this is a few releases out due to requiring database migrations support

We should set a milestone to queue up all pending changes blocked by migration support—tentatively 0.5, could land earlier though, especially if we manage to release more rapidly, and implement a migration strategy that can be applied via automatic updates.

securedrop/journalist_templates/profile.html
```diff
@@ -0,0 +1,52 @@
...
+    </div>
+  </p>
+  {% endfor %}
+{% else %}
+<p>User is not currently assigned to any sources.</p>
+{% endif %}
+
+<p><b>PGP public key:</b></p>
+{% if user.pgp_key: %}
+<pre>{{ user.pgp_key }}</pre>
```
fowlslegs: Maybe we don't want their entire PGP public key to be displayed here. I think a fingerprint would look nicer. We can have a "Download public key" link next to it, but journos shouldn't need each other's public keys on their JWs, so what's the point anyway?

securedrop/journalist.py
```diff
@@ -409,6 +411,55 @@ def remove_star(sid):
...
+        user.full_name = form.full_name.data
+        user.about = form.about.data
+        user.pgp_key = form.pgp_key.data
+        db_session.commit()
+        flash('Successfully edited your profile!', 'notification')
+        return redirect(url_for('profile', username=g.user.username))
+    return render_template("edit_profile.html", form=form, user=g.user)
+
+
+@app.route('/admin/edit-profile/<int:user_id>', methods=('GET', 'POST'))
```
fowlslegs: I think the `/admin/edit-profile/<username>/` URL routing scheme is nicer, but not super important.

securedrop/journalist.py
```diff
@@ -409,6 +411,55 @@ def remove_star(sid):
...
+        form.full_name.data = user.full_name
+        form.about.data = user.about
+        form.pgp_key.data = user.pgp_key
+    elif request.method == 'POST' and form.validate():
+        user.full_name = form.full_name.data
+        user.about = form.about.data
+        user.pgp_key = form.pgp_key.data
+        db_session.commit()
+        flash('Successfully edited the profile for {}!'.format(user.username),
+              'notification')
```
fowlslegs: Probably want `success` instead of `notification` here.

securedrop/journalist.py
```diff
@@ -409,6 +411,55 @@ def remove_star(sid):
...
+        form.full_name.data = g.user.full_name
+        form.about.data = g.user.about
+        form.pgp_key.data = g.user.pgp_key
+    elif request.method == 'POST' and form.validate():
+        user = Journalist.query.filter_by(username=g.user.username).one()
+        user.full_name = form.full_name.data
+        user.about = form.about.data
+        user.pgp_key = form.pgp_key.data
+        db_session.commit()
+        flash('Successfully edited your profile!', 'notification')
```
fowlslegs: Probably want `success` instead of `notification` here.

redshiftzero:
Closing until the project is in a position to make changes like this

Issue #972, opened 2015-04-01
Add endpoint to source interface to query package version metadata


garrettr:
Now that we are providing automatic package upgrades, it would be helpful for us to be able to verify that an upgrade successfully landed on every known SecureDrop instance. We can sort of do this for `securedrop-app-code` because the version is displayed in the footer of each page on the Source Interface, but we cannot query the version status of the other packages we provide.

Since these packages are maintained and upgraded publicly, our package repo is public, and the servers auto-upgrade within 24 hours, there is not much security risk in leaking the versions of these packages (it's not something an attacker wouldn't just be able to easily infer). However, being unaware that package upgrades are failing and not being able to respond quickly to rectify such a situation _is_ a serious security concern. Providing an endpoint that lists the versions of installed packages would help us monitor these systems.

I am not sure whether we should list the installed versions of all packages, all relevant packages, or only the packages that we provide. The 3rd option seems the safest to start with and also the most immediately useful. It is also generally useful to know what versions of Tor and Apache (at a minimum) the servers are running, especially when novel attacks on Tor Hidden Services are published (womp womp). However, publishing those might give an attacker more of an advantage and we should discuss the tradeoff.

This is related to the idea that FPF should receive some subset of sanitized OSSEC alerts. For example, we could also learn if an upgrade had failed by setting up cron-apt alerts to _also_ email FPF. These email alerts would be encrypted, and could also include the relevant logs, which would be helpful for debugging. However, I do not think these efforts are mutually exclusive, and both may be developed in tandem, evaluated, and the best solution refined over time.

fowlslegs:
@heartsucker Just a note that if something does not completely resolve an issue try to avoid use of ["keywords for closing issues"](https://help.github.com/articles/closing-issues-via-commit-messages/#closing-an-issue-in-the-same-repository).

In https://github.com/freedomofpress/securedrop/pull/1528, @heartsucker created a new view for the app `/metadata` that lists the SVS key fingerprint and SD app code version as JSON. It should now be trivial to add additional information to this view, but we should further discuss the trade-offs of listing any additional package versions or metadata before doing so.

heartsucker:
Whoops. Sorry, y'all.

eloquence:
[...]

PR #1528, opened 2017-01-19
added metadata route to source app

heartsucker:
In part fixes https://github.com/freedomofpress/securedrop/issues/972. This gives a bare minimum solution so that automation can quickly check the PGP fingerprint and app code version. This leaks zero information that can't be quickly scraped from the web interface with `grep` anyway.

### DIFF
--- a/securedrop/source.py
+++ b/securedrop/source.py
@@ -7,7 +7,7 @@ import subprocess
 from threading import Thread
 import operator
 from flask import (Flask, request, render_template, session, redirect, url_for,
-                   flash, abort, g, send_file, Markup)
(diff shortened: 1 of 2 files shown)


fowlslegs:
While this doesn't give us any information we can't glean from scraping version string or downloading the public GPG key from the server, it does structure that data and create a dedicated, centralized view of it. I think this is actually the perfect first step because it allows us to test out monitoring + logging of this metadata, without having to make the decision upfront regarding trade-offs to listing other package versions (e.g., other SD-served packages, Tor, Apache).

PR #1452, opened 2016-11-05
Replaced wordlist with well researched EFF wordlist

commandnotfound:
Replaced existing wordlist with the well researched EFF wordlist (more information found at https://www.eff.org/deeplinks/2016/07/new-wordlists-random-passphrases). This was discussed in #1361.

PR #1529, opened 2017-01-20
Update README.md Word List Links

benjic:
The existing link to the word list is broken after the list was split into nouns and adjectives. The README.md has been updated to include links to each of these files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,4 +34,4 @@ See the [Development Guide](https://docs.securedrop.org/en/latest/development/ge
 
 SecureDrop is open source and released under the [GNU Affero General Public License v3](/LICENSE).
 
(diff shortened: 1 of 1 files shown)


fowlslegs:
We still use a single Diceware list (EFF's new long list) to generate source codenames (passphrases). The nouns and adjectives lists are used to generate a different codename for each source that only the journalist sees (as they should not have access to the codename/passphrase generated for the source, but still need some mechanism by which to distinguish & remember sources). For reference:

> New codenames are generated with two new dictionaries, located in
deaddrop/dictionaries, with an adjective and noun pair. This is
memorable for the journalist. These dictionaries were generated with the
python natural language toolkit, using the wordnik database.

-- https://github.com/freedomofpress/securedrop/commit/c8f5d7aa47e9938d6d39225af8a1787ed6c74a2c

> Replaced wordlist with well researched EFF wordlist

-- https://github.com/freedomofpress/securedrop/pull/1452/commits/e8046e2f80bd615ccbe5335034afe4ac9847b05d

So if you want to fix this PR to reflect the above, I'd be happy to merge. Thanks!

PR #1535, opened 2017-01-27
Updated codename in Readme to match the one in screenshot

tldr:
Just fixed the Readme to display the correct codename in text as shown in the referred screenshot.


redshiftzero:
Good catch! Merging.

Issue #1536, opened 2017-01-28
Recommended 0.4 SOURCE UI Fixes


ninavizz:
Creating this as a new issue, and breaking-up different parts that can be more clearly referenced by @fowlslegs and @redshiftzero in their subsequent #helpwanted PR UX directives. :)

———————————
UX Implementation Recommendation: Should PRs be created with an interest in any of the below UX reccos implemented, cut-and-paste the listed items as sought, below, and include the PDF (or PNGs of the PDF), in the PR (sorry, I know devs hate PDFs!). I’m thinking that’d be the easiest way to make sure everything gets pushed thru smoothly? I’ll make art as it’s sought for implementation, just ‘cc me on those PRs. Will usually be able to do those SVGs w/in a few hours, s'long as it's not the middle of the night. :)

**_Please reference the wireframes here, for change implementations. _FILE UPDATED_ Sunday afternoon... Danke Shoen!_** 
[SD_SourceOhDotFour-02b.pdf](https://github.com/freedomofpress/securedrop/files/737711/SD_SourceOhDotFour-02b.pdf)

## All Screens
1. Left-Align each of the headers and change their color to `#7985AA`. Note: center-aligned content is spoken to in additional points. It should all get left-aligned, the center-align stuff is getting outta control—tho I presume made sense with fewer page items.

2. If a user is using a passphrase for the first time, number the H1s on the New Passphrase and Submit Stuff screens 1-2-3 so that users are more clearly lead through a path and know to return later, for getting replies (see page 3 of the above PDF).

3. Add a secondary-action button to all screens, to the left of the “Submit” button. Please style as shown—white fill w/ `#7985AA` stroke. The primary action button needs to be “Submit” with “Cancel” a clear secondary action (unless a different secondary action is specified) on all screens.

4. **DONE!** Thx @heartsucker <strike>Change all button text to be all-caps, regular weight (not bold), with generous character-spacing (as shown).</strike>

5. **DONE!** Thx @heartsucker <strike>Kill all button icons</strike>

6. Change “Log Out” to “Exit”

7. Pick one: either implement the weeble lockup for the exit button or the codename/lamp widget & exit button lockup.

## “Get Codename” screen particulars:

8a. Left-Align the header and change its color to `#7985AA`

8b. Change header text to “1. Welcome!” 

8c. Add instructional text below “1. Welcome!” header: This codename is what you will use in future visits, to receive messages from our journalists in response to what you submit on the next screen.

[...]

PR #1538, opened 2017-01-29
Ui fixes

heartsucker:
Makes some of the suggested changes in #1536 and fixes an incomplete rebase from #1349.


securedrop/sass/_base.sass
```diff
@@ -20,10 +20,14 @@
...
   h1, h2, h3
     font-weight: bold
+
+  h2, h3
     text-align: center
 
   h1
+    text-align: left
     font-size: 30px
+    color: #004080
```
redshiftzero: Seems like from the mockups that the text that we want this color and left alignment applied to is actually the h2 text?

screen shot 2017-01-31 at 2 11 52 pm

securedrop/source_templates/index.html
```diff
@@ -34,15 +34,16 @@
           </h2>
           <hr class="cut-out" />
           <p>If this is your first time submitting documents to journalists, start here.</p>
-          <a href="/generate" class="btn alt block" id="submit-documents-button"><img id="warning-close" src="{{ url_for('static', filename='i/font-awesome/white/cloud-upload.svg') }}" width="17px" height="17px"> Submit Documents</a>
+          {# adding a break between 'submit' and 'documents' to make both containers equal sized #}
+          <a href="/generate" class="sd-button btn alt block" id="submit-documents-button">SUBMIT<br/>DOCUMENTS</a>
```
redshiftzero: Looks like these containers are not equal sized in Tor Browser:

screen shot 2017-01-31 at 1 46 07 pm

securedrop/sass/journalist.sass
```diff
@@ -74,3 +74,54 @@ button.small-danger, a.btn.small-danger, .btn.small-danger
...
+    select
+      margin-left: 2%
+      width: 23%
+
+    .btn
+      background-color: #999
+    
+      &:hover
+        color: #999
+        background-color: white
```
redshiftzero: Verified this fixes the CSS on the journalist interface :+1:

securedrop/source_templates/index.html
heartsucker: I'm not seeing it display like this. What OS are you testing this on? I'm on Firefox 51.0.1 / Debian Stretch. Also, how are you getting TB to reach localhost?

redshiftzero: Check out [this link](https://docs.securedrop.org/en/latest/development/tips_and_tricks.html#using-tor-browser-with-the-development-environment) for how to allow TB to reach localhost. I was running Tor Browser 6.5 on Mac OS X when I made that screenshot. Check it out in Tor Browser (it's based on Firefox 45.7.0 so the CSS might render differently) and let me know if you can reproduce what I see above

securedrop/sass/_base.sass
heartsucker: Actually the bigger problem is that most places we use `h2`, we should be using `h1`, and the only places we use `h1`, should definitely be `h2`. Ha. I'll make some more changes.

securedrop/source_templates/index.html
heartsucker: The "no proxy" option isn't working for me. Also TB 6.5. Hmm. If you want to branch this and make a tweak, go ahead.

redshiftzero: Sure, I'll take a look

redshiftzero: Ok branched off this and in `fix-box-sizes-on-src-interface-index` there is a single commit 74f612b4ef42679d1f6f05f223f34700cbccd254 fixing these boxes to be the same size. Now in Tor Browser this looks like:

modified

I'll submit a PR against this repo to get that commit merged in after I merge this branch in

redshiftzero:
Hey @heartsucker - thanks for addressing these comments - everything looks good on my end now. If you fix the functional test failure I will merge this in 😄

heartsucker:
@redshiftzero Sounds good. It might take a few days for me to get to it. New job 'n all.

heartsucker:
Two things:

1) How the eff do you get a container to be the width of the stuff inside it without setting `min-width` to some pixel value?

2) Is cancel button outline / text supposed to be the same color as the continue button?

screenshot_2017-02-02_21-09-52

redshiftzero:
1. Try `width: auto` - also if you're talking about for the buttons, if you remove the `width: 35%` from the btn primary then the button will be the width of the content inside it: 

screen shot 2017-02-02 at 1 10 19 pm

2. For that button we are using the color suggested by @ninavizz - @ninavizz is the different color intentional for a UX reason? Unless there is a UX reason to make them different, I think it's fine to change to be the same color (note you'll also need to modify the rollover color).

ninavizz:
Hey @heartsucker—my bad on that hex call, sorry abt that! I got sloppy, it should visually match the primary button—yes, yes. :) 

I updated my spec to cite #727c9b as the recco'd stroke color (it's a taaaad darker than the button blue). 

Thx for catching & asking!

ninavizz:
P.S. I mis-spec'd that deep blue for the header, too—that should also be the same muted blue as the primary button. We probz need a styleguide... I'll start *something* tonite, that just covers these basics. :)

ninavizz:
Ok, I created #1552 as a go-to place to solidly document global basic things between all three UIs: Journalist, Source, and Admin. I had used `#004080` I had for hyperlink-blue, and confused that for the header's blue in my haste to document earlier notes. `#7985AA` is the global "muted blue" I want to see carry into headers. Really sorry for the runaround on that!

heartsucker:
> if you remove the width: 35% from the btn

D'oh. Note to self: don't debug on the S-bahn. :(

heartsucker:
@redshiftzero Should be good to merge now.

redshiftzero:
@heartsucker I took another look at this - looking much better - just a couple more small things 🙏 :

There was a rebase snafu on `/login`:

screen shot 2017-02-03 at 10 06 09 am

1. Remove extra codename field (leaving the one with `id="login-with-existing-codename"` since the functional tests are looking for that)
2. Cancel button should be at the same y-position as the continue button (once you fix the nesting of the `<p>`s that should happen naturally 😇 
3. Cancel button should redirect back to the index

Finally, the index looks like this in Tor Browser: 

screen shot 2017-02-03 at 10 05 52 am

4. But my SecureDrop-CSS-spider-sense tells me that if you change the classes on the "Check for a response" button to `sd-button btn primary index` then this button will be fixed :100:

Also thanks @ninavizz for that quick response 💎

ninavizz:
Are Sources now being asked to enter their passphrase 2x? :(

Cancel (or secondary) button should always be visually paired w/ the primary button—no more than ~4px space, between them. Once I'm fully competent with this Terminal/dev-build stuff (that @fowlslegs was extra-awesome spending THREE hours walking me through, on Tuesday!) that's still super brain-hurts-new to me, I'll do all I can to comment directly on the code itself. 

Longer-term, a UI library of common elements—such as button pairs!—with "how to modify w/o breaking code or visual standards" notes will be a nice thing to create, so all these repeat-elements can be way easier to wham-bam cut-and-paste/modify. More time spent building + designing, less time spent replicating the saaaame things over and over again. The Y!UI library was built when I was at Yahoo!, and maan, that thing was awesome!

heartsucker:
No, this  was just a messy rebase I didn't check carefully enough. I only
checked the things I fought changed and not everything. Some of the changes
broke the design and some are just functionally wrong (double password).
Changes soon (hopefully).

Am 03.02.2017 22:22 schrieb "Nina Eleanor Alter" <notifications@github.com>:

ninavizz:
Cool! I (and lots of others in the UX/security community) just find that whole "Enter your password in a second time!" pattern-standard to be absurdly asinine... and wanted to make sure it wasn't sought, here. I should trust my experience with you thus far, that you're not an advocate for piffy security-theatrics... :D

heartsucker:
@redshiftzero Apologies for the sloppy rebase and time wasted. This one should do it, though. However, two notes about things that look less than ideal. Not sure if these should be tackled here or in another issue.

1) Without JS enabled, it is not obvious what the `download` and `delete` buttons do (IMO).

screenshot_2017-02-04_09-46-13

2) The `reset` buttons here have too much text in them, so I think there should be text to the left explaining what the button does, and then `reset totp` and `reset yubikey` buttons.

screenshot_2017-02-04_09-46-50

@ninavizz Comments / suggestions?

redshiftzero:
@heartsucker no worries. Thanks for making these changes! I've checked this out and it now looks great - I'm going to merge this in now given that the changes are solid and we've already had a long back and forth in this PR. But! If you want to make another followup PR you could make the following changes to address these (IMHO valid but much smaller) concerns: 

1. Change button text to "Download Selected" and "Delete Selected" - which may involve making some CSS changes such that this text fits in one line when JS is enabled.  

2. I agree with your suggested solution here, bearing in mind that "Reset Google Authenticator" makes more sense to non-technical folk than "Reset TOTP" (people acting as administrators / point people for their SecureDrop instance might not be super techy).

heartsucker:
@redshiftzero You got it. :D

PR #1545, opened 2017-01-31
multi threading enabled

flipchan:
this will make the app run much faster

### DIFF
--- a/securedrop/source.py
+++ b/securedrop/source.py
@@ -460,4 +460,4 @@ def write_pidfile():
 if __name__ == "__main__":  # pragma: no cover
     write_pidfile()
     debug = getattr(config, 'env', 'prod') != 'prod'
-    app.run(debug=debug, host='0.0.0.0', port=8080)
+    app.run(debug=debug, host='0.0.0.0', port=8080, threaded=True)


heartsucker:
In production, `apache2` and `mod_wsgi` handle forking / multithreading.

From the [flask docs](http://flask.pocoo.org/docs/0.12/api/#api):

> Do not use run() in a production setting. It is not intended to meet security and performance requirements for a production server. Instead, see Deployment Options for WSGI server recommendations.

Usage of `app.run()` is limited to the dev environment, so I'm not sure this is needed.

(Also, not a maintainer, so take this with a grain of salt.)

fowlslegs:
@heartsucker is right that the Flask server is only used for development as recommended. My concern about enabling threading is that could make debugging more difficult. Also, I do not believe this will speed up anything in the development environment--manual interaction via a browser and runs of the test suite are synchronous processes, and so only need one thread anyway.

redshiftzero:
I'm gonna go ahead and close this one for the reasons @heartsucker and @fowlslegs describe - we don't use this in production and we only need one thread for development and testing (since we run the tests in series). 

However if you'd like to work further on SecureDrop, go ahead and check out the issues tagged "help wanted" 😃

PR #1546, opened 2017-01-31
removed 'scary' words from wordlist

### LINKED ISSUE #1541: "Treason" as a passphrase word option for Sources? #fail
https://docs.securedrop.org/en/latest/journalist.html

Super random: but, the third-down screenshot I came upon earlier today, and it shows a created passphrase with the word "treason" in it. Imagining what a n00b Government Accountability type leaker might think upon seeing that—and all I can think, is task abandonment, buh-bye.

Words such as "treason" and "traitor" or "prison" or "murder" or "stalking" or "jail" or "guilty" or "betrayal" or "deported" or "exile" etc. Would it be possible to assemble a list of likely trigger words to be blocked from inclusion in passphrases presented to Sources?

heartsucker:
Fixes #1541

### DIFF
--- a/securedrop/wordlist
+++ b/securedrop/wordlist
@@ -75,7 +75,6 @@ aflutter
 afoot
 afraid
 afterglow
-afterlife
 aftermath
 aftermost
 afternoon
@@ -92,7 +91,6 @@ aging
 agnostic
 agonize
 agonizing
-agony
 agreeable
 agreeably
 agreed
@@ -272,7 +270,6 @@ apply
 appointee
 appraisal
 appraiser
-apprehend
 approach
 approval
 approve
@@ -295,7 +292,6 @@ arise
 armadillo
 armband
 armchair
-armed
 armful
 armhole
 arming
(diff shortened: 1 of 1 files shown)


securedrop/wordlist
```diff
@@ -5079,7 +4960,6 @@ ranger
 ranging
 ranked
 ranking
-ransack
```
redshiftzero: Think this can go back in

redshiftzero:
(Note: removing 178 words from the list reduces entropy from 12.89 to 12.85 bits per word - a negligible change - so security-wise I think this is good to merge)

securedrop/wordlist
```diff
@@ -1012,7 +989,6 @@ chihuahua
 childcare
 childhood
 childish
-childless
```
heartsucker: My reasoning on this one would a that a whistleblower could lose custody of a child as a result of their actions, and this word might be "triggering" in that regard.

redshiftzero: Hmm fair point, feel free to leave in

securedrop/wordlist
heartsucker: I pulled this one because leak -> police raid -> ransack seemed like too easy a jump. Could put it back it still.

redshiftzero:
Looks good, merging!

PR #1555, opened 2017-02-04
Add note about Tails 3.0 needing a 64-bit machine

tkell:
@heartsucker, let me know if that's the right place in the docs for this.

### DIFF
--- a/docs/hardware.rst
+++ b/docs/hardware.rst
@@ -68,8 +68,19 @@ servers (for example, we do so to support our CI pipeline), doing so
 in order to run it on cloud servers is at your own risk and without
 our support or consent.
 
+.. _Workstations:
 Workstations
 ~~~~~~~~~~~~
+.. note:: SecureDrop depends on the Tails operating system for its bootable USB drives.
+	The upcoming version of Tails, Tails 3.0, will no longer support 32-bit computers.
+
+	The current targeted release date for Tails 3.0 is **June 13, 2017**.
+
+	 To see if you have a 64-bit machine, run ``uname -m`` from a terminal.
+
+	If you see ``x86_64``, then Tails should work on your current machine.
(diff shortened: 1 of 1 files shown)


heartsucker:
@tkell I actually think this would make sense to have more detail in the `Workstations` section with some details about how to check whether hardware is 64/32 bit *before* anyone buys it. The `Tails USBs` section could make note of the previous warning and direct users to read that in case they didn't see it.

fowlslegs:
What @heartsucker said. Except ending up w/ 32-bit hardware is only a possibility if you're buying some really old used machines. Even my 11 year old T60 has a 64-bit processor.

conorsch:
I'm in favor of merging this as-is, because a note ahead of the upcoming release is better than no mention at all. While it's theoretically a good idea to mention that machines should be 64-bit prior to purchase, most orgs aren't actually procuring new hardware for workstation purposes—they're repurposing old machines, for the most part, and that's a fine way to save costs.

tkell:
@heartsucker, @fowlslegs, moved it to `Workstation` as per your request, added an (attempt) at an internal link in the note in `Tails USB`.

fowlslegs:
You can build (and automatically rebuild every time a change to a rst file is saved to disk) the docs following the instructions here: https://docs.securedrop.org/en/stable/development/documentation.html#documentation-guidelines.

fowlslegs:
Closing in favor of https://github.com/freedomofpress/securedrop/pull/1570. Your commits are preserved in that PR @tkell. See http://docutils.sourceforge.net/docs/user/rst/quickref.html#implicit-hyperlink-targets and https://github.com/freedomofpress/securedrop/pull/1570/commits/b83a81ca9c4e9c31e219e0d096f1aa566a5448b0 to see how to make internal links to titled sections in reStructuredText.

PR #1561, opened 2017-02-06
Libvirt support for development machine

fowlslegs:
* Consolidate comments explaining memory requirements in Vagrantfile.
* Add support for libvirt provider for development machine.
  * ~~Uses new `available_cpus` Ruby function to ensure all CPUs are used.~~
  * Although `config.vm.synced_folder` defaults to the exact settings in this
  commit, if NFS support is not present on the host, Vagrant will fall back
  to unidirectional `rsync` syncing. It's better to fail and make the user
  setup NFS support than leave w/o bidirectional syncing support.*
* Use all available CPUs w/ the Virtualbox provider.

\* See [discussion below](https://github.com/freedomofpress/securedrop/pull/1561#issuecomment-278498306) as to whether this is completely accurate.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -40,6 +40,11 @@ Vagrant.configure("2") do |config|
     development.vm.provider "virtualbox" do |v|
       v.memory = 1024
     end
+    development.vm.provider "libvirt" do |lv|
+      lv.memory = 1024
+      lv.cpus = available_vcpus
+      config.vm.synced_folder './', '/vagrant', type: 'nfs', disabled: false
+    end
   end
 
   # The staging hosts are just like production but allow non-tor access
@@ -244,4 +249,25 @@ end
 def internal_network_name
   repo_root = File.expand_path(File.dirname(__FILE__))
   return File.basename(repo_root)
+end
+
+def available_vcpus
+  # Increase number of virtual CPUs in guest VM.
+  # Rather than blindly set it to "2" or similar,
+  # inspect the number of VCPUs on the host and use that,
+  # to minimize compile time.
+  available_vcpus = case RUBY_PLATFORM
+    when /linux/
+      `nproc`.to_i
+    when /darwin/
+      `sysctl -n hw.ncpu`.to_i
+    else
+      1
+    end
+  # If you want to restrict the resources available to the guest VM,
+  # uncomment the return line below, and Vagrant will use half the
+  # number available on the host, rounded down.
+  # (Ruby will correctly return the quotient as an integer.)
+  # return available_vcpus / 2
+  return available_vcpus
 end
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -39,11 +39,9 @@ Vagrant.configure("2") do |config|
     # more memory than the previous version did.
     development.vm.provider "virtualbox" do |v|
       v.memory = 1024
-      v.customize ["modifyvm", :id, "--cpus", available_vcpus]
     end
     development.vm.provider "libvirt" do |lv|
       lv.memory = 1024
-      lv.cpus = available_vcpus
       config.vm.synced_folder './', '/vagrant', type: 'nfs', disabled: false
     end
   end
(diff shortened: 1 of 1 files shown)


conorsch:
Looking pretty good. Was able to take the development and staging machines for a spin. Pleasantly surprised that the private networking settings assigned the static IP addresses correctly under libvirt, same as vbox. Under development, I'll need to rig up NFS support, will report back with results for that.

Remove the `available_cpus` logic, though. That was a hack for the grsec repo to speed up long-running builds, but in general it's aggressive to the point of impolite to claim all available VCPUs on a VM host for use in just one guest. Especially for folks running many VMs at one time, that's an undeserved performance hit.

fowlslegs:
Changes made!

`aptitude install nfs-kernel-server portmap` will get you a working NFS server. Be sure to test that syncing is bidirectional. And download + mutate the latest 2.3.1 bento box if you haven't.

msheiny:
> if NFS support is not present on the host, Vagrant will fall back to unidirectional rsync syncing.

This didn't happen for me FYI - it just failed. I'm going to try installing the NFS package now

msheiny:
So I'm fine with this PR now - I resolved all my NFS issues. I'm concerned that someone on a Mac might have an issue and the rsync fallback could fail but I dont have one to test. If that happens we could figure a conditional check that would map it back to rsync easy enough sooo I'm good for margin .

:+1:

conorsch:
With an alley-oop from @msheiny to get the NFS settings working locally (firewall rules were too restrictive), this also plays nicely with me. I'd like to get sign-off from someone using VirtualBox that this doesn't break their setup—@heartsucker, you're using VirtualBox under Debian Stretch, aren't you? Can you take this for a spin and confirm nothing breaks?

conorsch:
Deserves mention in #1208 once we testdrive it for a while.

fowlslegs:
>> if NFS support is not present on the host, Vagrant will fall back to unidirectional rsync syncing.

> This didn't happen for me FYI - it just failed. I'm going to try installing the NFS package now

Only true when you omit the sync method argument. At least according the Vagrant Libvirt Provider README it defaults to NFS. Maybe that's not true anymore, but the behavior I experienced was that if that line isn't there, rsync would be used.

> I'm concerned that someone on a Mac might have an issue

Vagrant Libvirt Provider only supports KVM & QEMU, even though Libvirt runs on macOS. Also, this is intended for grsec users who have problems with VirtualBox. VB in general works fine on macOS.

> I'd like to get sign-off from someone using VirtualBox that this doesn't break their setup—@heartsucker, you're using VirtualBox under Debian Stretch, aren't you? Can you take this for a spin and confirm nothing breaks?

I've tested this branch w/ VB--just a development machine though. If anyone wants to test the staging or prod machines please do.

conorsch:
@redshiftzero Can you spin up a staging environment based off this PR? I'd like to confirm the virtualbox settings are not affected by the addition of libvirt config lines.

redshiftzero:
@conorsch: I was able to set up a staging environment using this branch with no problems on VirtualBox 💯

conorsch:
Magnificent, thanks for testing, @redshiftzero. Going in!

redshiftzero:
Staging environment does work as reported, but I got the following error when attempting to provision a development VM using VirtualBox... failure on `vagrant up` on this branch:

```
==> development: Setting hostname...
NFS requires a host-only network to be created.
Please add a host-only network to the machine (with either DHCP or a
static IP) for NFS to work.
```

conorsch:
Drat, that's precisely the kind of regression I was worried about. Thanks for following up, @redshiftzero. We'll need to make sure the nfs settings only apply to libvirt hosts.

fowlslegs:
@conorsch Surely we can use a Ruby `if` branching statement in the Vagrantfile, right? The reason I guess it worked for me under VB was that I have NFS setup, but we definitely don't want to add another dependency on the road to getting SD up and running.

conorsch:
> Surely we can use a Ruby `if` branching statement in the Vagrantfile, right? 

Unfortunately not! Due to the way Vagrantfiles are parsed, all config lines are read and evaluated, regardless of provider. There _are_ still options for conditional logic, though—moving discussion to #1582.

Issue #1220, opened 2015-12-08
ossec emails rejected due to "@ossec.server" sender domain


neo-luddite:
ossec alerts didn't work for me out of the box because the sender
email is @ossec.server which gets rejected because the domain doesn't
exist. my workaround are postfix generic tables for manipulating the
sender domain.

ageis:
The OSSEC alerts require some configuration. The sender e-mail should be sasl_username@sasl_domain. You are routing through another mail server, correct? Is it one you operate, or a different third party? Is this the staging or production environment?

This might be a legit issue.
Line 41 in ./group_vars/securedrop_monitor_server.yml: `postfix_hostname: ossec.server`

garrettr:
@neo-luddite Did you configure the OSSEC alerts per our documentation? See https://docs.securedrop.org/en/latest/install.html and https://docs.securedrop.org/en/latest/ossec_alerts.html.

neo-luddite:
I'm using one of our production mailservers as a mail relay. 
In production-specific.yml  I entered the following values accordingly to my setup:

> smtp_relay: "mail.example.net"
> smtp_relay_port: "25"
> sasl_username: "abc"
> sasl_domain: "example.net"
> sasl_password: "password"

/var/ossec/etc/ossec.conf: 
> \<email_from\>ossec@freedom.press\</email_from\>

mail.log:
> postfix/smtpd[1761]: connect from localhost[127.0.0.1]
> postfix/smtpd[1761]: 71B0DA0048: client=localhost[127.0.0.1]
> postfix/cleanup[1763]: 71B0DA0048: message->id= 20151214101336.71B0DA0048@ossec.server>
> postfix/qmgr[1555]: 71B0DA0048: from=<ossec@freedom.press>, size=612, nrcpt=1 (queue active)
> postfix/smtpd[1761]: disconnect from localhost[127.0.0.1]
> postfix/local[1764]: 71B0DA0048: to=<ossec@ossec.server>, orig_to=<root@localhost>, relay=local, delay=0.12, delays=0.08/0/0/0.04, dsn=2.0.0, status=sent (delivered to command: /usr/bin/procmail)
>postfix/qmgr[1555]: 71B0DA0048: removed
>postfix/pickup[1554]: A0A33A09A1: uid=1002 from=<ossec>
>postfix/cleanup[1763]: A0A33A09A1: message-id=<20151214101336.A0A33A09A1@ossec.server>
[...]

Issue #894, opened 2015-03-03
OSSEC email from option


dolanjs:
Currently the OSSEC email from option is a static value `securedrop@freedom.press`. This value was chosen because it was not triggering any common spam filters. But some orgs may want a custom value. 

Changing this value should be done via debconf so that the file is not marked as locally modified which would cause issues for future auto updates.

conorsch:
Resolved via #1565.

PR #1565, opened 2017-02-07
Override from address on outbound ossec emails

msheiny:
This allows re-mapping of the default ossec@server_name address that
is seen on the FROM line coming out of the ossec box. Post fix is
handling this rather than ossec.

Added this for customization and because some SMTP servers will
not allow sending email from an address that is not the address
associated with the user account. Noticed this specifically as a default
rule on Office365 recently.


conorsch:
Good catch. Closes #1220, where this was first mentioned.

Unfortunately we cannot trivially update `prod-specific.yml` because it will cause merge conflicts with site-specific changes. Make sure to test how it merges:

1. edit `prod-specific.yml` with custom vars
2. `git stash save site-specific-changes`
3. checkout this feature branch
4. `git stash pop`

If you have to resolve conflicts manually, we'll need another way. Often appending the new var to the edit of the prod-specific.yml file helps avoid conflicts, but that's quite unintuitive, given that all the OSSEC vars are grouped together. We might need a smarter way to manage site vars in 0.4.

msheiny:
> If you have to resolve conflicts manually, we'll need another way. Often appending the new var to the edit of the prod-specific.yml file helps avoid conflicts, but that's quite unintuitive, given that all the OSSEC vars are grouped together. We might need a smarter way to manage site vars in 0.4.

What about? We make a new `example-prod-specific.yml` with example data and then we `gitignore` the `prod-specific.yml` file?

conorsch:
We will most likely have to migrate completely away from prod-specific.yml and require a new vars layout, yes. But for the nonce I'd prefer to retain backwards compatibility on the prod-specific.yml file, simply because we haven't broken it yet, and I want to keep that streak going, even in develop (for now).

msheiny:
@conorsch i removed the var from `prod-specific.yml` for now. Definitely want to revisit the vars issue though in a later release.

fowlslegs:
Closes https://github.com/freedomofpress/securedrop/issues/894.

conorsch:
We've recently had Admins contact us to debug the OSSEC alerts, and the changes proposed in this PR resolved the issue. Since mail handlers are increasingly restricting the allowed from values, in an attempt to fight spam, we need to fix this issue for 0.4. Without these changes, the OSSEC setup cannot work with many mail providers.

Since @msheiny removed the new variable from prod-specific.yml, sidestepping any stash/pop conflicts, I no longer have concerns about breaking the Admin workflow. Still, several changes should be made prior to merge:

* saner declaration of default vars (e.g. via role defaults, rather than in template)
* update OSSEC docs to include this option in the troubleshooting

I'll handle implementing those during review.

conorsch:
Rebased on top of latest develop, to take advantage of new config tests.

conorsch:
Out of an abundance of caution, we're preserving the existing default behavior of the FROM line for OSSEC emails being set to `ossec@ossec.server`. That's not a value we can change the default for lightly, given that it could break alerting for Admins with currently working setups. The addition to the OSSEC guide documentation should clarify the issue for anyone who runs into problems—I placed the new entry near the top of the troubleshooting guide because it's becoming an increasingly common issue.

conorsch:
Satisfied with the latest changes. Added a config test. @redshiftzero can you sanity-check the prose in the docs addition? If you're OK with that then we're good for merge here.

redshiftzero:
Docs look good to me!

PR #1570, opened 2017-02-10
Add note about Tails 3.0 needing a 64-bit machine

fowlslegs:
Supersedes #1555.


conorsch:
Includes @tkell's commits, :+1:. LGTM!

Issue #1563, opened 2017-02-07
Release SecureDrop 0.3.11


redshiftzero:
This is a tracking issue for SecureDrop 0.3.11 - tasks may get added or modified. 

# Finish release candidate (0.3.11-rc1)
- [x] Make new PR targeting `release/0.3.11` branch cherry picking commits in #1522 and #1480 and updating for latest version of Tor Browser - @redshiftzero
- [x] Cherry pick commits from #1520 (SourceClear integration), #1165 (adds the explicit `/logout` route), and #1186 (needed such that the styling on the interface is not horrible), the commits from #1445 that fix the broken test suite into `release/0.3.11`, #1521 (remove source choice in number of words from interface) - @redshiftzero
- [x] Review and merge above PR - @garrettr 
- [x] Build test debs - @conorsch
- [x] Stand up test apt server with 0.3.11-rc1 packages - @msheiny

# QA
- [x] Unit tests pass on 0.3.11-rc1 staging VM
- [x] Test upgrade from 0.3.10 works correctly on staging VM
- [x] Interactive testing produces produce nothing strange in logs in staging VM
- [x] Test upgrade from 0.3.10 works on prod w/ test repo debs
- [x] Test install (not upgrade) of 0.3.10 works on prod w/ test repo debs

# Release
- [x] Build final Debian packages for 0.3.11
- [x] Release 0.3.11 
- [x] Publish blog post about 0.3.11 Debian package release

# Post-release
- [x] Merge release changes into development branch - @redshiftzero

fowlslegs:
Coveralls is being run in `.travis.yml`, but there's no `securedrop/.coveragerc` in branch `release/0.3.11`, hence the ever-pending coveralls CI status (see, e.g., https://github.com/freedomofpress/securedrop/pull/1578). We probably want to either:

1. Copy the `securedrop/.coveragerc` from `develop`.
2. Stop installing & running coveralls in `.travis.yml`.

We could also just ignore it. I'm in favor of 1 provided the checklist here is up-to-date; if we haven't built the test debs, this is a change that we can make super quick and won't require testing as it's outside of the app code. It'd be nice to know what type of coverage we have on our actual production code, not just our develop branch.

conorsch:
Coveralls changed added by @redshiftzero in cf48c3d7f30164dbde336f5f54bf6ea1ad325dfc.

freddymartinez9:
Great job with the release @redshiftzero !!!

[...]

PR #1578, opened 2017-02-11
Updating 0.3.11 "JS Warning" text

ninavizz:
The prior text, "We recommend changing your Security Slider to high" does not identify the Tor Browser as the the requested task's location. It will be confusing to users using SecureDrop for the first time, to read this—as their brains are in SecureDrop land, and it's more intuitive at that point to look for said Security Slider in SecureDrop—not Tor. 

Also—ZOMG, Nina's first PR!! :)


securedrop/source_templates/index.html
```diff
@@ -62,7 +62,7 @@
     Included here so the images can preload while the user is first
     reading the page. Hidden by default. -->
     <div class="bubble">
-      <p>You appear to be using the Tor Browser. You can turn the Security Slider to High in 4 easy steps!</p>
+      <p>You appear to be using the Tor Browser. In Tor's "Security Settings" pane (accessible from the toolbar's onion-button), we reccommend turning Security Slider to High—it's easy-peasy!</p>
```
redshiftzero: `'onion-button), we reccommend '` -> `'onion button), we recommend'`
`'easy-peasy!'` -> `'easy!'`

redshiftzero: Also, it seems like we are now explaining how to get to the security slider twice in this bubble. Once in this line and then again in the ordered list below...

ninavizz: That's ok! Instructions are often redundant. The point of the edit, is that currently users have to mouse-over something to access that ordered-list... and they shouldn't have to. The single line of text should explain all they need to know, to directly act. But if they STILL don't get it, then they have the ordered list. Standard pattern. 

The less stuff users have to access to know what to do, the better. :D

ninavizz: @redshiftzero also, I'm not clear on what you changed or want me to change... there's sooo much going on in this PR, now (and I kinda intended it to target 0.3.11, was that bad?). The use of -> and -> and -> is easy for geeks to figure-out, but keeping the written directions as clear sentences, will go farther with both speaking clearly to users, and keeping a cadence in the tone of the writing. Smoothly cadenced writing = less anxiety. :)

Separately: I saw you added an error-bang into the "please quit Tor" end-message, and that it's red. While that message is unlikely to dissuade users from using the product, the user never did anything "wrong," so that's a poor use of both the icon and the color—hence, the purple bubble w/ the fingerprint recommendation. Dialogs also shouldn't have numbered-points in them. It's just not a convention, and is odd. Are the fellas just trying to bang-out that release w/o the recommended dialogs, or did y'all want more time to iterate on them?

conorsch: > and I kinda intended it to target 0.3.11, was that bad?

We're frozen on the features we we want to land in `0.3.11`, so I assume @redshiftzero was targeting develop to make sure your changes still get included in the near future, even though the release ship has sailed on 0.3.11. We just got through final QA on that (see #1563).

Looks like the automatic branch retargeting via GitHub made some ugly decisions, so we'll have to cherry-pick your contributions and resubmit a new PR that supersedes this one. Smoke and mirrors, largely—we'll take care of the dirty work there. 

Your UX feedback continues to enlighten, thanks very much for your time!

fowlslegs:
@ninavizz Just a tip you probably want to add the lines
```
[pull]
    rebase = true
```
to your `~/.gitconfig`. This might not all make sense yet, but here's an explanation of why: http://mislav.net/2013/02/merge-vs-rebase/.

securedrop/source_templates/index.html
ninavizz: Coolio, can dig it. 

Longer-term, text-edits and ux tweeks can hopefully be rolled-out more incrementally. I'm heartened by the enthusiasm for all my recommendations, and yes, admittedly, do squeal with glee whenever I see a full page of them implemented in code—but a good goal to target, is ongoing UX updates. With time, with time. :)

garrettr:
0.3.11 has been released, so it is too late to make any further changes to it.

Issue #1574, opened 2017-02-10
SVGs are disabled with Tor Browser high security; replace with PNGs


heartsucker:
@micahflee reported that SVGs are disabled by the high security setting in Tor Browser here https://github.com/freedomofpress/securedrop/pull/1480#issuecomment-277787564.

Possible solutions would be to either a) have a fall back to PNG or b) ditch the SVGs entirely and only use PNGs (or some other format).

freddymartinez9:
are there a lot of SVGs? This should be low-hanging fruit unless I messed up my basic linux commands...

```
:securedrop b_meson$ find . -name '*.svg' 
./docs/diagrams/SecureDrop.svg
```

redshiftzero:
FYI this is done in #1567 targeted for the next release

heartsucker:
As of commit b83f2197:
```heartsucker@pythagoras:~/code/freedom-of-press/securedrop (develop)$ find . -name '*.svg'
./docs/diagrams/SecureDrop.svg
./securedrop/static/i/svg/hand_with_fingerprint.svg
./securedrop/static/i/svg/success_checkmark.svg
./securedrop/static/i/font-awesome/white/comments.svg
./securedrop/static/i/font-awesome/white/times.svg
./securedrop/static/i/font-awesome/white/refresh.svg
./securedrop/static/i/font-awesome/white/cloud-upload.svg
./securedrop/static/i/font-awesome/white/arrow-circle-o-right.svg
./securedrop/static/i/font-awesome/black/info-circle.svg
./securedrop/static/i/font-awesome/black/times.svg
./securedrop/static/i/font-awesome/black/lock.svg
./securedrop/static/i/font-awesome/black/info.svg
./securedrop/static/i/font-awesome/black/trash.svg
./securedrop/static/i/font-awesome/black/cloud-upload.svg
./securedrop/static/i/font-awesome/black/exclamation-triangle.svg
./securedrop/static/i/relieved_face.svg
./securedrop/static/fonts/fontawesome-webfont.svg
./securedrop/source_templates/svg/server_upload.svg
```

heartsucker:
@redshiftzero Your PR doesn't include the new ones for the 0.4 release.

```
./securedrop/static/i/svg/hand_with_fingerprint.svg
./securedrop/static/i/svg/success_checkmark.svg
./securedrop/static/i/relieved_face.svg
./securedrop/source_templates/svg/server_upload.svg
```

[...]

PR #1581, opened 2017-02-15
Merge 0.3.11 changes into develop

redshiftzero:
This merges in the changes from the 0.3.11 release. I've converted the CSS changes to SASS. 

On current `develop` note that some of the icons from the mixed icon and text buttons have been removed. However the icons are still present on buttons without text, e.g. the refresh codename button, so the CSS styling we added for 0.3.11 is still necessary.


redshiftzero:
OK so after discussion with @garrettr and @conorsch, we're going to merge in these changes while keeping the icon styling used in 0.3.11. I've redone the merge and resolved merge conflicts to clobber changes in develop. This means some of the icons are back, and we're just going to leave it as is until after 0.4.

I made the following additional changes for consistency:

* In 2368d22 I added PNGs for the four new SVGs that were in `develop` on the source interface (and updated the templates/styling/AppArmor rules as necessary). This closes #1574.

* In 901084a I've replaced the screenshots in the docs on the source interface. The docs/screenshots changes in this PR show the current state of `develop` with the new changes from 0.3.11.

fowlslegs:
I think 4 and 9 can probably wait because they were already problems to some degree and it will require some discussion to figure out how we want to resolve them. I removed one of the points because it was based on my incorrect use of `find`. Everything else represents a small, novel issue introduced by this PR, and should be fixed before merging.

redshiftzero:
Thanks for review @fowlslegs! I've made changes in commit dd1e3c9. Note that I tried to keep things scoped as closely as possible on getting the 0.3.11 changes merged into develop. If there are problems that are existing issues from develop or non-ideal aspects of the 0.3.11 release, I think it's appropriate to address in a followup PR. Responses to your comments point-by-point:

1. Fixed and updated asserts in the unit tests. The text in `assertIn` now checks that the text for the right template is displayed, so if the wrong template is used in the future the tests will fail.
2. Specified height/width of the png so that it doesn't render as an oval. Any other formatting stuff is the same way on develop, so this is one of those things for a followup PR.
3. Yep so this `upload-large-gray.png` image was added for the 0.3.11 release, but the image is different in develop. This is a case where I modified the image to match develop. In commit dd1e3c9 I have removed the image and AppArmor rule from the repo.
4. As you suggest, I will leave this one for someone else to address in develop since this was not introduced in this PR.
5. I do think you're right that the button on the right looks better and I have removed the `<br>`. Unfortunately, the differences between these two buttons aren't due to CSS or `<br>` differences (add a few characters to the right button and you will see that the icon alignment is the same). I also agree that these buttons might look better with the icons twice the size. Let's do additional styling tweaks in a followup PR. Note that we need to review all UI changes in develop so we might not want to spend too much time making further small tweaks until we are sure we are keeping the styling as is, but once we do then let's make small CSS tweaks.
6. Shrank the icon down a couple of pixels so the buttons are the same size.
7. Specified height and width of the image so it displays at the correct resolution.
8. As this is how the icon is displayed in the release, I'm going to leave as is and we can address in a followup PR.
9. This was unfortunately already the case in develop, but a subsequent PR resolving this makes sense for sure (again, we will have to all agree on the desired UI changes first).

conorsch:
Great conversation here. Looks like @redshiftzero has already addressed the final blockers for merge. Spinning up a development VM now to review the interface changes locally.

heartsucker:
@fowlslegs In number 4, `success` and `notification` are different, because they lead to different styles in the SASS.
https://github.com/freedomofpress/securedrop/blob/b83f21976b4bd26519d2daeb5bbe2185544175cc/securedrop/sass/_base.sass#L314-L316

fowlslegs:
@heartsucker Oh, duh, I kind of saw that as I was reviewing, but it didn't register. I think it's still worth reviewing how we could consolidate some templates to use these too, but that's a future item I'll make another issue for.

@redshiftzero Re-reviewing now!

conorsch:
LGTM! Thanks for the hard work on this one, all. (@heartsucker it's grand to have you contributing so regularly. Onward to 0.4!)

PR #1583, opened 2017-02-15
Adjusts vagrant shared folder behavior under libvirt

### LINKED ISSUE #1582: Fix networking on development VM
I get the following failure when attempting to provision a development VM using VirtualBox:

```
==> development: Setting hostname...
NFS requires a host-only network to be created.
Please add a host-only network to the machine (with either DHCP or a
static IP) for NFS to work.
```

See #1561 for additional context

conorsch:
The NFS strategy for mounting /vagrant in the development machine should
only take effect if the libvirt provider is currently active. Let's
reference the `override` pragma, documented here [0], to achieve.

Supplements changes in #1561. Closes #1582.

[0] https://www.vagrantup.com/docs/providers/configuration.html

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -40,9 +40,9 @@ Vagrant.configure("2") do |config|
     development.vm.provider "virtualbox" do |v|
       v.memory = 1024
     end
-    development.vm.provider "libvirt" do |lv|
+    development.vm.provider "libvirt" do |lv, override|
       lv.memory = 1024
-      config.vm.synced_folder './', '/vagrant', type: 'nfs', disabled: false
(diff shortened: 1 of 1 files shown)


conorsch:
@redshiftzero I'll need you to test this. Happy to pair on this in realtime so we can iterate faster. Confirmed that the change still runs NFS shares under libvirt on Linux. The real question is, does it work for you under VirtualBox on OS X?

Test by running `vagrant up development` on this branch and report results!

redshiftzero:
Verified this works on OS X, thanks @conorsch!

conorsch:
Golly, glad to hear it!

PR #1590, opened 2017-02-21
Allow admin to set reboot time in securedrop-prod.yml

### LINKED ISSUE #1515: Allow the admin to configure automatic reboot time
The SecureDrop servers automatically reboot every 24 hours to periodically wipe memory (#805). The admin should be allowed to choose when the servers reboot; 5am is not an ideal time for everyone. I, personally, would like to be available when the servers reboot in case they don't come back up on their own.

fowlslegs:
Fixes #1515.


ageis:
Related, but more documentation about the nightly reboots would be good. The only doc that mentions the fact that this happens AFAIK is [ossec_alerts](https://docs.securedrop.org/en/stable/ossec_alerts.html). Under the [set up Ubuntu](https://docs.securedrop.org/en/stable/servers.html?highlight=ubuntu#continue-the-installation) doc there an instruction to "configure your time zone", but no mention of why that might be relevant or important in the context of scheduled tasks.

freddymartinez9:
What is to prevent someone from putting in a value above 23? We aren't really checking the value anywhere as far as I can tell...

conorsch:
> What is to prevent someone from putting in a value above 23?

@freddymartinez9 Cron will automatically error out with a "bad hour" message in that case, and Ansible will honor that error by halting execution of the playbook. So, we have basic sanity-checking on the values by default.

conorsch:
@fowlslegs Have you tested how this change would behave with site-specific changes to `prod-specific.yml`? From a glance at that update, I suspect a merge conflict will occur if an Admin tries to pop their stashed changes on top of this new vars file.

fowlslegs:
Popping worked fine for me.

fowlslegs:
Rebased!

conorsch:
This breaks the staging environment due to the new config var being undefined. It should have a default declared in the role, with top-level exposure in the `prod-specific.yml` file so Admins are aware they can edit it.

> Popping worked fine for me.

For me as well. We'll need to do extensive testing of the stash/pop workflow prior to 0.4, especially as we reevaluate the Ansible versioning story (#1146).

conorsch:
The wrong file was templatized here. The current implementation writes the cron logic to the cron-apt config destination, which breaks cron-apt. The new config tests (#1616) catch this, so I'll rebase this branch on the latest develop and fix the template conversion during review.

conorsch:
With revisions, all config tests are now passing. The reboot time is now configurable, and the new config var retains the previous default behavior of "4:00AM, according to localtime on the server," which means that the staging environment is working well with the new config item.

Once tests pass, I'll merge.

Issue #1412, opened 2016-09-20
Remove extra test wrapper


redshiftzero:
Get rid of `test.sh` that is used to run the functional tests

fowlslegs:
Working on this... have a question:

* There's no reason not to merge `test_{unit_journalist,journalist,_single_start}.py` into a single file, right? They are all `unittest.TestCase` objects testing the journalist module--I'd combine the imports, but they'd each stay their own class.

@garrettr

fowlslegs:
Pending PR is https://github.com/freedomofpress/securedrop/pull/1527.

Issue #1594, opened 2017-03-01
Sass 3.5 will require Ruby>=2.0.0


nvesely:
Currently Ubuntu Trusty has 1.9.3 in it's repos. We'll need to pin Sass<3.5 if Ubuntu doesn't upgrade Ruby by the time Sass 3.5 is released. Sass is currently on version 3.4.23.

garrettr:
Does SASS need to be installed on the servers? Wouldn't it be better to generate the final CSS from the SASS as part of the build process, and only ship the CSS in `securedrop-app-code.deb`?

heartsucker:
Sass is not needed for the servers. Only for the build. The 'build' role
does the compilation during the Debian packaging. However, we could nail
the version down in the 'gem install' step to give us longer until we have
to deal with this.

Am 01.03.2017 10:01 nachm. schrieb "Garrett Robinson" <
notifications@github.com>:

Does SASS need to be installed on the servers? Wouldn't it be better to
generate the final CSS from the SASS as part of the build process, and only
ship the CSS in securedrop-app-code.deb?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/freedomofpress/securedrop/issues/1594#issuecomment-283468647>,
or mute the thread

fowlslegs:
Not a direct concern for production systems, but needed by the development and build Vagrant machines (the latter of which generates the final CSS for the app code package w/ `sass`).

eloquence:
This gem has reached end-of-life; closing in favor of #4452.

PR #1527, opened 2017-01-18
Run tests directly with pytest

fowlslegs:
Changes/ new features include:

- Easy runs of 1+ selected functional/unit tests, callable by short names.
- Pass any options through to pytest when running any unit or functional
  test(s), or execute arbitrary pytest commands in the test environment.
- Run docstring tests.
- Create non-admin journalists as well.
- Better exception handling and error reporting.
- Ensures cleanup of test environment _always_ happens.
- Some work towards Python 3 compatibility for eventual transition.
- Adds sick SD ASCII art to `./manage.py run`.
- Fix `crypto_util` module doctests.
- Update pytest section syntax in `setup.cfg`.

Supersedes https://github.com/freedomofpress/securedrop/pull/1516 and resolves https://github.com/freedomofpress/securedrop/issues/1412.

PR #1464, opened 2016-11-06
Automatically builds deb packages in staging environment

conorsch:
Reimplementation of changes presented in #1214. The staging playbook now also includes the build-deb-pkgs playbook, so the latest app code in the local git repo is packaged and installed on staging machines. This is much more intuitive behavior for a "staging" environment than what exists at present: Ansible configs are pulled in from the local git repo, but application code is pulled in from the FPF apt repository (apt.freedom.press), installing whatever the latest production release is.

The prod environment still installs hosted deb packages from apt.freedom.press, as intended. To test:

```
$ vagrant up --no-provision build /staging/
$ vagrant provision /staging/
```

If you do not bring up the build machine, the playbook will fail and instruct you to do so. Hurray for informative error messages. Also includes a few test updates to track changes brought in via #1445.

Supersedes and therefore closes #1214. Closes #1040.

PR #1612, opened 2017-03-10
Refactor build-securedrop-app-code-deb-pkg role

### LINKED ISSUE #1507: Make sure securedrop-app-code doesn't ship any .pyc files

fowlslegs:
* Constructively builds app code package as discussed in #1510. Use of globs ensures that both new and old modules, JS files, etc. will be included in packages, while editor swap files and bytecode will not. This is done with a Python script because it was nicer to write than Ansible and idempotence is not important here (we delete and re-create the build directory on each playbook run).
* Pins Sass<3.5 as discussed in  #1594.
* We force updating the CSS our SASS generates and fail immediately on error.
* Cleanup is now done just before a build happens. This makes it easier to introspect the package contents after running the playbook; you don't have to decide before running it that you want to pass `--skip-tags=cleanup`.

Tbh, I think tags should be removed from this playbook as I don't see a point for them. Would be happy to make a follow-up commit to do that, but want to get @conorsch's input.

Fixes #1507. Partially #1594. Supersedes #1510.


fowlslegs:
Alright, so it looks like I'll have to re-include the tests and test config files in order for us to be able to run them from the `staging` and `prod` Vagrant machines. A cleaner alternative would be to sync `securedrop/tests` to `/var/www/securedrop/tests`, and that should actually work fine once we merge #1527, which fixes #1412. So I'm going to label this one blocked for now and come back to it later.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/copy_app_code_files.py
```diff
@@ -0,0 +1,42 @@
...
+         'source_templates': ['*.html'],
+         'static/js': ['*.js'],
+         'static/js/libs': ['*.js']}
+PATHS = ['static/fonts', 'static/i']
+
+os.makedirs(BUILD_DIR)
+
+for rel_path, glob_list in GLOBS.iteritems():
+    try:
+        os.makedirs(os.path.join(BUILD_DIR, rel_path))
```
heartsucker: I think you want `exist_ok=True` here instead of the `try/except`, though that is a `python3` option. Since this is a build script and not the main app, it might be worth it to update to `python3` to have one less thing to change in the future.

conorsch:
Rebased on top of latest develop, starting review.

conorsch:
Added some tests to track the changes here, particularly the "no .pyc files" allowed check. Ran a build against the 0.3.11 source (since 0.3.12 isn't backported yet; see #1636) and compared with previous debs for 0.3.10. The diff is here: https://gist.github.com/conorsch/e0aa2b2bf16bc814d4e8bb06ec92fe20 The major changes are:

1. No `.pyc` files—that's good!
2. No webassets dotfiles—that's good!
3. Document interface renamed to Journalist interface. See #1614 for complications there.
4. Tests are absent from the deb package. Related to changes in #1527. We'll need to work out a strategy to run the tests against app-staging. 

Eventually we'll need to modify the build logic to push/pull over the SSH transport, otherwise the build process can never run in CI. Willing to defer that work for now, since this PR importantly omits .pyc files in the deb packages, which is an important contribution (and closes #1507). This is solid progress for now, so I'm going to merge, and continue hashing out the staging test environment toward closing #1614 and possibly #1464.

conorsch:
Earlier @fowlslegs said:

> A cleaner alternative would be to sync `securedrop/tests` to `/var/www/securedrop/tests`, and that should actually work fine once we merge #1527, which fixes #1412. So I'm going to label this one blocked for now and come back to it later.

That seems like a decent solution, so I'm going to add it here and rerun the analysis. Have concerns about the import path hacks that will break, but will find that out soon enough.

conorsch:
Modified the additive logic (first described in @heartsucker's #1510) to use raw rsync, rather than a custom Python script. See the "FILTER RULES" section of the rsync manpage for full details, but in summary, we declare an explicit list of globbed includes, then exclude everything else. The whitelist approach is vastly superior to the prior build logic (e.g. #1507), and the rsync-based implementation is consistent with the logic @fowlslegs initially wrote for this PR.

Touched up some of the config tests to track regressions here. Was tempted to expand the config test suite related to the deb packages, but punted those additional asks to #1637, so we can get this in and keep going. In particular, it's difficult to validate the functionality of the new build logic without automated builds in the staging environment; see #1464 for an example of that. 

The new diff is here: https://gist.github.com/conorsch/e0aa2b2bf16bc814d4e8bb06ec92fe20 Only one item stands out to me in that list: the mode of `/var/www/securedrop` is 700 versus 755 across the two versions. That's a separate problem for the config tests to shake out when #1464 gets rolling.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/rsync-filter
```diff
@@ -0,0 +1,36 @@
...
+include source_templates/
+include source_templates/*.html
+include static/
+include static/fonts/
+include static/fonts/**
+include static/i/
+include static/i/**
+include static/i/font-awesome/
+include static/i/font-awesome/**
+include static/i/tipbox/
```
fowlslegs: Is this a really old legacy folder @conorsch? Not present in any of our active branches afaicr.

conorsch: Assuming you mean `static/i/tipbox/`, I see it on present development:

```
./securedrop/static/i/tipbox
./securedrop/static/i/tipbox/tipbox-hed-j-all.png
./securedrop/static/i/tipbox/tipbox-logo.png
./securedrop/static/i/tipbox/tipbox-hed-submit3.png
./securedrop/static/i/tipbox/tipbox-hed-user.png
./securedrop/static/i/tipbox/tipbox-hed-j-single.png
./securedrop/static/i/tipbox/tipbox-hed-submit1.png
```

Now as far as whether those includes are _used_, that's a separate matter! 

Will follow-up with some tests-related changes to deb packages build logic, since the additive approach I took wasn't quite additive enough. Resisting the urge to rsync the entire tests dir and stick with the selective approach.

Issue #1606, opened 2017-03-02
Journalist Interface Apache logging is broken


conorsch:
SecureDrop by design never logs interactions on the Source Interface, to reduce the threat of information disclosure on source behavior, whether by compromise, discovery, or forensic analysis. The Journalist Interface, however, should be logged so that Admins have access to information about Journalists using the system. 

The Journalist Interface logging is currently broken, and only outputs the log entry `combined` for every event. This problem is caused by an absent config line for `LogFormat` that defines the behavior of the combined setting. Proposed addition to `/etc/apache2/apache2.conf`: 

```
LogFormat "%h %l %u %t \"%r\" %>s %O \"%{Referer}i\" \"%{User-Agent}i\"" combined
```

That line is taken from the default Apache configs when installed under Ubuntu Trusty, and should work well for our use case. 

We received this report via our support portal. Many thanks to the Admin who reported it.

conorsch:
Simple fix, marking for 0.4 milestone.

PR #1615, opened 2017-03-11
Fix Apache logging formatting

fowlslegs:
Fixes https://github.com/freedomofpress/securedrop/issues/1606.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -75,6 +75,9 @@ Vagrant.configure("2") do |config|
     staging.vm.provider "virtualbox" do |v|
       v.memory = 1024
     end
+    staging.vm.provider "libvirt" do |lv, override|
+      lv.memory = 1024
+    end
     staging.vm.provision "ansible" do |ansible|
       ansible.playbook = "install_files/ansible-base/securedrop-staging.yml"
       ansible.verbose = 'v'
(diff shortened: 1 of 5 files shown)


conorsch:
Blocking here until we get #1616, so the tests can be updated to track this change. There's currently a test marked "expected failure" in that PR that targets these log formats.

conorsch:
#1616 is merged, so will add tests to track regression here, then we're good for merge.

fowlslegs:
@conorsch I'd actually like to write the tests here, but I wanted to ask you what you'd do here first. Would it be sufficient to just test that this line is present in this file, or would you actually try to parse the log to make sure it's sensible? I'm guessing probably just the former, but was just wondering if there's anything more robust (but perhaps less complicated than log parsing) I could do.

conorsch:
A few considerations:

1. Under no circumstances should a line in the journalist interface log match `^combined$`; (currently they all do). So that should be its own test, to guard against regressions here.
2. Match for a rough approximation of the format we're looking for; we don't want false failures because an overly specific regex didn't consider certain allowed characters or patterns in the log.
3. The log will be empty unless the Journalist Interface has been interacted with.

Mostly 3 will cause problems; we don't yet have a cohesive "run all the tests" flow, so it's hard to predict that the log lines will be present prior to running the config tests. But as long as we have 1 landing int his PR, I'm happy.

conorsch:
Rebased on top of latest develop.

conorsch:
Rebased on top of latest develop. Made some modifications to the new local packages logic (#1612) to support running the application tests inside app-staging, to help with testing that the config change works as expected. Update the associated config test (which was previously marked as an expected failure) so we can track this regression going forward.

Issue #1614, opened 2017-03-11
Investigate Document -> Journalist Interface renaming side-effects


conorsch:
In #1395 the "Document Interface" was renamed to "Journalist Interface". There are a few problems with the current implementation, chiefly that there's no migration strategy in place for currently running instances. 

If we ship that change to the Ansible playbooks, then production instances will generate a new Authenticated Tor Hidden Services (ATHS) on next run, and lock out Journalists from the interface by updating the Apache vhost to listen to the _new_ service, using a HidServAuth value that Journalists won't have on their workstations.

Currently the `app-prod` machine is broken in the develop branch, and fails during the provisioning run, due to the AppArmor profile changes dropping support for the old vhost file. That kind of makes sense: the prod machines install from the FPF apt repo, and changes in e.g. #1464 are intended to provide a more robust staging environment that would allow for testing the new changes. It makes running app-prod impossible right now, though, which has terrible implications for pre-release testing.

So the question becomes: do we want to commit to the rename and write migration tasks, or back out the changes to simplify the scope of the 0.4 release?

heartsucker:
On the server side, I believe this is easily fixed via a `preinst` script addition on new packages that can move the hidden service dirs. Ensuring that admins have up to date playbooks that don't overwrite these changes might be a bit trickier (for example in `install_files/ansible-base/roles/tor-hidden-services/tasks/configure_tor_hidden_services.yml`).

How much leeway is there expected to be with admins running old versions of playbooks?

conorsch:
@heartsucker Leveraging the Debian package scripts is really our only option here, short of backing out the changes (which I'm loath to do), so I agree. We'll need to straighten out the package build scripts in order to test this adequately, e.g. #1612 and #1464.

conorsch:
#1464 and #1612 are merged, so drilling down on this now.

conorsch:
In order to test the upgrade path here, we should:

1. Provision staging VMs based on the 0.3.12 release tag. Doing so will ensure we get the 0.3.12 packages from the apt repo, and the old Ansible config, referencing the "Document Interface," will be used to configure the tor and apache services.
2. Check out a feature branch based on current `develop` and patch the `preinst` script to handle the migration.
3. Build new deb packages with the patched `preinst` script and evaluate success.

[...]

PR #1616, opened 2017-03-11
Converts config tests from Serverspec to TestInfra

### LINKED ISSUE #1580: Convert ServerSpec tests to TestInfra
Substantial work has been done on config tests, currently implemented in [ServerSpec](http://serverspec.org/) (#1006). The config tests are not currently run automatically via a CI pipeline (#1067), but should be. Given recent efforts to improve the application tests (#1445, #1457, #1466, #1474, #1516), we should make a similar push to test the system hardening specifications. 

The core developers already have strong familiarity with Python and [pytest](http://docs.pytest.org/en/latest/contents.html), so we should convert the existing config tests to [TestInfra](https://testinfra.readthedocs.io/en/latest/#), which is pytest-based. Doing so would allow us to remove Ruby-based tooling from the repo and focus on a Python-only development environment for the foreseeable future.

The config tests must be updated before we make a concerted effort to improve the Ansible versioning situation on Admin Workstations (#1146, #1531). We'll also need a strong config testing baseline in order to evaluate the upgrade from Ubuntu Trusty to Xenial (#1530).

conorsch:
All SecureDrop tests are now Python-based, with the config tests using [testinfra](https://testinfra.readthedocs.io/en/latest/), which is itself a [pytest](http://docs.pytest.org/en/latest/plugins.html) module. The config test conversion allows us to remove the Ruby-based Serverspec tooling, which was _not_ running as part of CI (see #1067), and was underutilized by developers.

Items of note:

* Updated the developer documentation on the config tests (related to #1208, but doesn't resolve it).
* The prod tests aren't reliable yet. Need to route the SSH connections reliably over Tor, and need to resolve issues with the `app-prod` provisioning flow (#1614) in order to fix.
* The diff is massive, but only affects config tests and docs, plus the gitignore file.

To test, read the new docs included in this PR and if anything isn't clear, yell at me for writing bad docs. You should be able to get passing tests on development, build, app-staging, and mon-staging. 

Closes #1580.


docs/development/config_tests.rst
```diff
@@ -0,0 +1,103 @@
...
+------------
+
+.. code:: sh
+
+    pip install -r testinfra/requirements.txt
+
+Running the tests
+-----------------
+
+In order to run the tests, each VM will be created and provisioned, if
```
redshiftzero: Great docs. One note: the voice here makes it sound like the test suite will create the machine for you - it might be useful to make this more explicit, something like: "each VM will need to be created and provisioned - i.e. be sure you run `vagrant up <machine>` on the relevant machine before running the testinfra suite for that machine" (this exact language not necessary).

redshiftzero:
This looks really great. I gave this a spin using Ansible 1.8.4 and Vagrant 1.8.5 following the documentation in this PR. Note using `pytest==3.0.6`, I got the following warning:
```
==================================================== pytest-warning summary ====================================================
WP1 None Module already imported so can not be re-written: testinfra
```

It looks like other people have [had a similar issue](https://github.com/philpep/testinfra/issues/125),  downgrading and using `pytest==2.9.0` did not produce the warning. Not sure if y'all saw something like this before during testing.

Anyway, here's what I found:

## testinfra: development

### Testing procedure

```
$ vagrant up
$ ./testinfra/test.py development
```

### Result

I got one failure:

```
===================================================================== FAILURES ======================================================================
___________________________________________ test_development_app_directories_exist[ansible://development] ___________________________________________
[gw1] darwin -- Python 2.7.12 /usr/local/opt/python/bin/python2.7
File = <class 'testinfra.modules.base.GNUFile'>

    def test_development_app_directories_exist(File):
        """
        Ensure that application code directories are created
        under /vagrant for the development environment, rather than
        /var/www as in staging and prod.
    
        Using a separate check from the data directories because /vagrant
        will be mounted with different mode.
        """
        f = File("/vagrant/securedrop")
        assert f.is_directory
        assert f.user == "vagrant"
        assert f.group == "vagrant"
>       assert oct(f.mode) == "0700"
E       assert '0755' == '0700'
E         - 0755
E         + 0700
```

## testinfra: staging

### Testing procedure

```
$ vagrant up /staging/
$ ./testinfra/test.py app-staging
$ ./testinfra/test.py mon-staging
```

### Result

Result for app-staging: 2 failures, [details in this gist ](https://gist.github.com/redshiftzero/15d935976002f701b9acac3c67f7a46f#file-app-staging)

Result for mon-staging: 5 failures, [details in this gist](https://gist.github.com/redshiftzero/15d935976002f701b9acac3c67f7a46f#file-mon-staging)

conorsch:
@redshiftzero Thanks for the detailed review! You're right, the docs should provide better guidance on creating and provisioning the VMs beforehand—the old serverspec docs did so, and that language should not have been removed. Will re-add.

Regarding the dev error, that sounds like a provider-specific issue, e.g. libvirt vs. virtualbox. There's no need to track that explicit mode on the dev VM, so I'll simply remove that check, because the failure is not helpful to developers. 

Regarding the staging failures, that sounds like the VMs were not rebooted prior to running the tests, since nearly all the networking checks fail. We _removed_ the automatic reboot at the end of the playbooks from the staging VMs, simply to get them up and running faster, but I'm not convinced that was the right approach. (Prod VMs always reboot at the end of a playbook run.) For now, I'll document the need to reboot the hosts in the developer docs, but we should consider enforcing the reboot in staging, same as prod.

redshiftzero:
Ah, makes sense. Adding a note to the docs works :+1:. To confirm, I rebooted both VMs and reran the testinfra suites, which reduced the failures to three. Remaining failures on app-staging are [here](https://gist.githubusercontent.com/redshiftzero/b3b6c51eb4e051be1b8c66611b9e0f80/raw/0b63a838030bf9c48f4c9a38de0657a0c97dfb1e/app-staging) and for mon-staging are [here](https://gist.githubusercontent.com/redshiftzero/b3b6c51eb4e051be1b8c66611b9e0f80/raw/0b63a838030bf9c48f4c9a38de0657a0c97dfb1e/mon-staging). 

It looks like two of the failures are due to the hardcoded iptables rulesets in the tests - I seem to have different uids on `debian-tor` on both VMs and for `postfix` on `mon-staging`: 

```
vagrant@mon-staging:~$ id -u postfix
107
vagrant@mon-staging:~$ id -u debian-tor
106
vagrant@app-staging:~$ id -u redis
107
vagrant@app-staging:~$ id -u debian-tor
106
```

If you look at the line by line difference in the iptables rules in the links above you'll see that the differing uids are the only issue, so if this is resolved these two iptables tests should pass.

conorsch:
> It looks like two of the failures are due to the hardcoded iptables rulesets in the tests - I seem to have different uids on `debian-tor` on both VMs and for `postfix` on `mon-staging`: 

Drat, tried to KISS on those. Fortunately @msheiny wrote logic to infer those values dynamically, same as we were doing in the Serverspec tests, so I'll remove the skip markers on those and iron them out.

conorsch:
@redshiftzero Updated the networking tests to include dynamic vars for UIDs and GIDs. All tests should now pass for you on development, app-staging, and mon-staging. Let me know if you can get any to fail!

msheiny:
Okay running through now...
- [ ]  Development
- [ ]  app-staging
- [ ]  mon-staging

msheiny:
> assert '** No agent available.' == 'app-staging-10.0.1.2 is available.'

I think this was failing last week too when I wrote the test... im still waiting for my staging environment to come up to confirm its still happening.

msheiny:
okay i got everything passing on my end under libvirt. Had to reboot the staging boxes pre test as @conorsch recommended

conorsch:
Thanks @msheiny! Going in.

Issue #1601, opened 2017-03-02
Explicitly set height and width of PNGs to ensure they are rendered correctly


fowlslegs:
We replaced SVGs with PNGs in the SI in order to have Security Slider High compatible icons, but it turns out if the height and width are not specified in the HTML templates they can be rendered the wrong size and scaled non-proportionately. It would be good to go through each one and ensure the icons are the right size, so they line up properly with their accompanying text, and are rendered proportionately. See, e.g., https://github.com/freedomofpress/securedrop/pull/1581/commits/dd1e3c90830ccf19323a79a40f94c2ce6f478251.

PR #1619, opened 2017-03-14
Refactor account changes logic

### LINKED ISSUE #1597: Reset 2FA buttons in the `/admin/edit/<int:user_id>` view are way out of alignment
The reset TOTP button is not justified left and there is no div before the reset TOTP and HOTP buttons as there is in the `/account` view. This would be a good time to DRY out the templates here by making a single template for these two views. There is no security risk in doing so as we rely on the `@login_required` decorator, not choice of template, which is irrelevant if you know the right URL to send the right HTTP request to because you've read the SD source. Obviously, the template would not show the admin-only capabilities to the non-admin users, as a matter of design, not security.

fowlslegs:
* If a user tries to change a password for an account and enters the existing
  password, the operation is now idempotent. Similarly, if an admin "changes"
  a user's username to itself, this operation is also idempotent. The new
  `commit_account_changes` function only writes to the database if the `db.User`
  object of the user's account being changed has been locally modified, which it
  determines via the `Session.is_modified(db.User)` method.
* The `/admin/edit/<int:user_id>` and `/account` routes now use the same
  template. Since the former provides a superset of the functionality of the
  latter, and there is no security risk in doing so, DRYing out the templates
  here made sense. While this fixes #1597, combining the templates also has
  potential to expose elements meant for admins to users in some future
  regression. To defend against this, I introduced a new functional test module,
  `tests.functional.make_account_changes`, which provides comprehensive checks
  to this regard. I also did some DRYing of these route functions since they
  both change passwords, commit account changes to the database, and flash
  messages regarding the success or failure of these operations.
* I also made some improvements to the relevant flashed messages for account
  changes, making progress towards #1476, #1601, #1602. It seemed like
  "success" and "failure" were better categorical indicators than
  "notification" and "failure," so I went with that and used the green
  checkmark PNG we have already employed in the source interface. I changed all
  the flashed icons here to PNGs while I was at it to make a small step towards
  making the JI security-slider-friendly, and explicitly specified their sizes
  to make sure they rendered correctly.
* Changing some of the flashed messages in the process of my refactor, I had to
  change the corresponding unit tests as well. To test both the content and
  category of the flashed message, I decided to take advantage of Flask-Testing's
  `assertMessageFlashed` method. This required installing an additional testing
  dependency, `blinker`, which I did without changing any other package versions.


securedrop/journalist_templates/edit_account.html
```diff
@@ -1,31 +1,82 @@
...
 </form>
 <br />
-<form method="post" action="{{ url_for('account_reset_two_factor_hotp') }}" id="reset-two-factor-hotp">
+<form method="post" action="{{ hotp_reset_url}}" id="reset-two-factor-hotp">
+{% if user %}
+  <input name="uid" type="hidden" value="{{ user.id }}"/>
+{% endif %}
   <input name="csrf_token" type="hidden" value="{{ csrf_token() }}"/>
-  <button class="sd-button" type="submit" class="pull-right">RESET TWO FACTOR AUTHENTICATION (HOTP YUBIKEY)</button>
+  <button class="sd-button" type="submit" class="pull-right">RESET TWO FACTOR AUTHENTICATION (HARDWARE TOKEN)</button>
```
redshiftzero: Can we add some additional explanatory text not on the button indicating that Yubikey is the hardware token and Google Authenticator is the mobile app?

securedrop/journalist_templates/edit_account.html
```diff
@@ -1,31 +1,82 @@
...
+  {% set hotp_reset_url = url_for('admin_reset_two_factor_hotp') %}
+{% else %}
+  {% set totp_reset_url = url_for('account_reset_two_factor_totp') %}
+  {% set hotp_reset_url = url_for('account_reset_two_factor_hotp') %}
+{% endif %}
+
+<form method="post" action="{{ totp_reset_url }}" id="reset-two-factor-totp">
+{% if user %}
+  <input name="uid" type="hidden" value="{{ user.id }}"/>
+{% endif %}
```
redshiftzero: I see why you tried to DRY up these two. In my view, the original templates for these views were pretty simple albeit with some duplication. Now there is a single template, but it has ~20 lines of basically "if this is view x, then do thing x, else do thing y for view y". This makes the template more complicated which might be a bit tricky if the template is used both for admin and non-admin views (as it is here). If there is duplication, imho a cleaner approach is to use [include](http://jinja.pocoo.org/docs/2.9/templates/#include) and [macros](http://jinja.pocoo.org/docs/2.9/templates/#macros) where possible in two separate templates. I don't think this necessarily needs to be changed for merge. Just putting my .2 cents and seeing what you think 💯

securedrop/journalist.py
```diff
@@ -306,31 +333,14 @@ def edit_account():
...
-                flash("Your password is too long "
-                      "(maximum length {} characters)".format(
-                      Journalist.MAX_PASSWORD_LEN), "error")
-                return redirect(url_for("edit_account"))
-
         try:
-            db_session.add(user)
-            db_session.commit()
-            flash(
-                "Password successfully changed!",
```
redshiftzero: Just noting here that this specific flashed message will not be displayed in favor of the more generic "account updated" message. I think this is fine, however just wanted to draw attention to this in case a UX observer wants to say that this is a Bad Thing.

securedrop/journalist.py
```diff
@@ -247,40 +247,67 @@ def admin_reset_two_factor_hotp():
...
 
     if request.method == 'POST':
-        if request.form['username'] != "":
+        if request.form['username']:
             new_username = request.form['username']
-            if Journalist.query.filter_by(username=new_username).one_or_none():
-                flash("Username {} is already taken".format(new_username),
+            if new_username == user.username:
+                pass
+            elif Journalist.query.filter_by(
```
fowlslegs: That's funny. [PEP8](https://www.python.org/dev/peps/pep-0008/) says:

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

I'm not going to change it unless you really think it's necessary.

fowlslegs:
Okay, ready for re-review @redshiftzero

securedrop/journalist.py
redshiftzero: Well, as is, this is neither PEP8 nor consistent with the current style:

`securedrop/journalist.py:293:17: E125 continuation line with same indent as next logical line`

However I'm going to merge this as is and then in the next release after 0.4 we will make sure to finally add style / linting checks to CI so that we can automate this discussion 😉 .

PR #1623, opened 2017-03-15
Install tmux in common role

### LINKED ISSUE #1375: Bento Box 2.2.9 does not include tmux
`tmux` is launched upon SSH login for our `staging` and `prod` VMs and our production installs. In the latest Bento Box (version 2.2.9) this dependency is missing. Since Bento Boxes simply run the Ubuntu installer, they should include the packages in a default Ubuntu install--no more or less. The current script for building these boxes is using Ubuntu 14.04.1 ISO [1], which is actually out of date and we should make a PR to bump them up to Ubuntu 14.04.5 (but that's a side note). The solution is to, with Ansible, explicitly ensure that `tmux` is installed across all VMs (and thus also for our production installs).

[1] https://github.com/chef/bento/blob/master/ubuntu-14.04-amd64.json#L254

fowlslegs:
Fixes #1375.

### DIFF
--- a/install_files/ansible-base/roles/common/tasks/install_tmux.yml
+++ b/install_files/ansible-base/roles/common/tasks/install_tmux.yml
@@ -1,11 +1,10 @@
 ---
-- name: Install ntp for ntpd.
+- name: Install tmux.
   apt:
-    pkg: ntp
+    pkg: tmux
     state: latest
     force: yes
     update_cache: yes
     cache_valid_time: 3600
   tags:
     - apt
-    - ntp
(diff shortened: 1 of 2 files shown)


conorsch:
Looks good. I'm going to block pending merge of #1616, so we can add a config test to ensure this package is present. We didn't have one before, and still don't, and having one would have caught the regression described in #1375.

conorsch:
Added associated test to #1616, since the package is already present in the base image, so the test is currently passing.

fowlslegs:
@conorsch The arguments I passed to the `apt` module were just copied from the `install_ntp.yml` tasklist, fwiw.

PR #1626, opened 2017-03-16
Permanently disables swap in securedrop-app-code preinst

### LINKED ISSUE #1620: Swapfile not really disabled
The Ansible config tries to disable swapfile on the Application and Monitor Servers, via `swapoff -a`. This works, but only for the current boot cycle. If a machine is configured with a swapfile in `/etc/fstab`, that swapfile will be restored on a subsequent reboot. Since the machines reboot nightly, the `swapoff -a` approach is close to useless.

In order to disable swap effectively, the first-run Ansible config should ensure that no swap entries exist in fstab, removing them if found.

conorsch:
Checks for active swap on system, and if found:

  * disables it so the system won't write to it
  * shreds the contents so any sensitive data is destroyed

Regardless of whether an active swap config was disabled, any fstab
entry of type "swap" will be commented out, effectively disabling it
permanently by ensuring it won't be reenabled on subsequent reboots.

Closes #1620.

### DIFF
--- a/install_files/securedrop-app-code/DEBIAN/preinst
+++ b/install_files/securedrop-app-code/DEBIAN/preinst
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # preinst script for securedrop-app-code
 #
 # see: dh_installdeb(1)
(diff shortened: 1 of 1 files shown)


install_files/securedrop-app-code/DEBIAN/preinst
```diff
@@ -13,12 +13,46 @@ set -e
...
+    # Search for path format in output.
+    swap_partition=$(grep '/' <<< "${swap_result}" | perl -ane 'print $F[0]')
+
+    # If a filepath for swap was found, disable it.
+    if [[ -n "${swap_partition}" ]]; then
+        echo "Disabling swap..."
+        # Disable all active swap.
+        swapoff --all
+        # Securely erase swap partition.
+        shred "${swap_partition}"
```
redshiftzero: Do we have a sense of roughly how long this should take? We should probably include in our pre-release notification to administrators a note about this

conorsch: During initial testing in VMs it was on the order of 1-5 minutes. In most cases the swap partition in the VMs was either completely unused, or only minimally used. Either way, the shred operation is going to vary based on the size of the partition. Will test further on hardware so we can provide a reasonable estimate.

conorsch:
This solution prioritizes fixing the config currently running on production Application Servers; it ignores any active swap on Monitor Servers. That's fine for now, but we should update the Ansible config to enforce swap absence on both machines prior to 0.4.

install_files/securedrop-app-code/DEBIAN/preinst
```diff
@@ -13,12 +13,46 @@ set -e
...
+
+function permanently_disable_swap() {
+    # Swap usage is prohibited in the context of SecureDrop, due to risk of
+    # forensic discovery recovering plaintext submissions that were written
+    # to disk prior to encrypting. If a swapfile is active, disable it,
+    # and disable all swap entries from /etc/fstab, to prevent mounting
+    # after reboots.
+
+    # Check if swapfile is currently enabled.
+    swap_result=$(swapon --summary)
```
fowlslegs: Nit: "Declare function-specific variables with `local`. Declaration and assignment should be on different lines." Also: "Always quote... command substitutions...."
```
local swap_result
swap_result="$(swapon --summary)"
```

install_files/securedrop-app-code/DEBIAN/preinst
```diff
@@ -13,12 +13,46 @@ set -e
...
+
+    # Search for path format in output.
+    swap_partition=$(grep '/' <<< "${swap_result}" | perl -ane 'print $F[0]')
```
fowlslegs: See previous comment. Also, this doesn't account for the case there's more than one swap partition (unlikely, but I don't want to discredit it, and I don't think supporting that scenario should significantly complicate this script). In fact, if there are multiple swap partitions, this command will return a string of their concatenated mapped paths. Then the `shred` command below will error out.

install_files/securedrop-app-code/DEBIAN/preinst
fowlslegs: (`swap partition size` * `3`) / `sequential write speed`. 5 minutes max with low-end hardware (80MB/s HDD) and a big swap partition (8GB). Except that could be wrong if `shred` uses a blocking randomness source. I'm going to go with we probably don't need a warning and for most people using SSDs it will be about 30s. We can test on our NUC.

conorsch:
Nits fixed and squashed. Merging, will proceed with prerelease testing. If no showstoppers during testing, we'll announce the upcoming 0.3.12 release later today, slated for publishing in the apt repo on Monday 2017-03-20.

PR #1628, opened 2017-03-17
Run testinfra under travis

msheiny:
Right now we can only run the `development` tests which we'll obviously need to look into expanding, but just that baseline is a good step.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ before_install:
 - echo HRNGDEVICE=/dev/urandom | sudo tee /etc/default/rng-tools
 - sudo /etc/init.d/rng-tools restart
 install:
+- pip freeze -l
 - pip install -r testinfra/requirements.txt
 script:
 - echo localhost > inventory
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,8 @@ before_install:
 - echo HRNGDEVICE=/dev/urandom | sudo tee /etc/default/rng-tools
 - sudo /etc/init.d/rng-tools restart
 install:
-- pip install ansible==1.8.4
 - pip install coveralls
+- pip install -r testinfra/requirements.txt
 script:
 - echo localhost > inventory
 - ansible-playbook -i inventory -vv --syntax-check install_files/ansible-base/securedrop-travis.yml
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ python:
 - '2.7'
 
 virtualenv:
-  system_site_packages: false
+  system_site_packages: true
 before_install:
 - 'for e in /.packer-env/*; do echo -n "${e}: "; cat "${e}"; done'
 - sudo apt-get update -qq
@@ -18,8 +18,8 @@ before_install:
 - echo HRNGDEVICE=/dev/urandom | sudo tee /etc/default/rng-tools
 - sudo /etc/init.d/rng-tools restart
 install:
-- pip install -r testinfra/requirements.txt
 - pip install coveralls
+- pip install -r testinfra/requirements.txt
 script:
 - echo localhost > inventory
 - ansible-playbook -i inventory -vv --syntax-check install_files/ansible-base/securedrop-travis.yml
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ script:
   # securedrop_user default www-data
 - sudo chown root:root securedrop/config.py
 - sudo sh -c "export DISPLAY=:1; cd securedrop && ./manage.py test"
+- ./testinfra/test.py
 after_success:
   cd securedrop && coveralls && cd ..  # cd back to repo root for srcclr run
 addons:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ python:
 - '2.7'
 
 virtualenv:
-  system_site_packages: true
+  system_site_packages: false
 before_install:
 - 'for e in /.packer-env/*; do echo -n "${e}: "; cat "${e}"; done'
 - sudo apt-get update -qq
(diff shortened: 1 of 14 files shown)


msheiny:
heyyyyy @conorsch / @redshiftzero  - this is ready for review now. some of the commit history is messy - lots of back and forth to make things work. Im fine with squash merging this if it makes the history cleaner.

.travis.yml
```diff
@@ -18,8 +18,10 @@ before_install:
 - echo HRNGDEVICE=/dev/urandom | sudo tee /etc/default/rng-tools
 - sudo /etc/init.d/rng-tools restart
 install:
-- pip install ansible==1.8.4
-- pip install coveralls
+- pip freeze -l
+# Preinstalled travis pytest is not tracked by us
+- pip uninstall pytest -y
+- pip install -r testinfra/requirements.txt
```
fowlslegs: I think it would be better to merge this file with `securedrop/requirements/test-requirements.in` and just let the existing role that installs the test deps install these. Even if we don't always need all of them in the course of development, it shouldn't add a very significant amount of time to provisioning. This way we also don't have to worry about version clash. You could also remove the `install` section here entirely as long as you are installing with `pip install -U -r <requirements file>`. Since `pip-tools` fixes the lower-level requirements in the `.txt` file it produces, one doesn't have to worry about `-U` producing different results if non-top-level requirements are updated on PyPi.

testinfra/development/test_development_application_settings.py
```diff
@@ -120,12 +127,12 @@ def test_development_default_logo_exists(File):
...
-    assert oct(f.mode) == "0644"
-    # TODO: Ansible task declares mode 400 but not as string, needs to be fixed
-    # and tests updated. Also, not using "mode" in tests below because umask
-    # on snapci machines differs from the /vagrant folder in dev VM.
-    # Fixing Ansible task may fix differing perms.
+    assert f.user == sd_test_vars.securedrop_user
+    assert f.group == sd_test_vars.securedrop_user
+    # check if logo is the default securedrop png
+    if f.md5sum == "92443946d5c9e05020a090f97b62d027":
+        assert oct(f.mode) == "0664"
```
redshiftzero: Should this be `0644`? That's what I have locally in my development VM

.travis.yml
msheiny: The issue with your suggestion is that I moved the `ansible` requirement into the `testinfra/requirements.txt` and you need ansible before you can install the `test-requirements.txt`.

I guess we could go back to a single travis-requirements.txt file that just tracks the ansible version and then everything else could be shoved in `test-requirements.txt` ? Would that be a good compromise? The reason I took `ansible` out of the `travis.yml` because I thought it was better to track all the pip dependencies in dedicated requirements files - its more obvious where to look to when you need to find them

testinfra/development/test_development_application_settings.py
msheiny: Yeah what the hell... that's failing locally for me now too. We should be setting permissions in ansible on that whole folder, I guess I tweaked it to work in CI without going back and then confirming it was still passing locally :( Let me try pushing a fix.

.travis.yml
conorsch: I don't agree that we should shoehorn these pip dependencies into `test-requirements`. The test requirements are already problematic in that they get installed on target hosts (such as app-staging) simply to run the tests. We haven't had a "developer requirements" for some odd reason, and the testinfra requirements here are at least close to that, in that they install required packages such as ansible, rather than merely _recommending_ a specific version of Ansible in the docs.

I wouldn't be opposed to moving these requirements into the requirements directory with the others, and calling them `dev-requirements` or similar, but I figured we could shake that out later.

testinfra/development/test_development_application_settings.py
msheiny: So i took a look at my prod SD instance, it also has permission `644` across that folder, `664` seems to only apply to `travis`. Technically I couldn't find a place where we are setting that permission for the folder (correct me if im mistaken) so I'm tempted to just remove the mode check unless its a custom logo in the test. Then subsequently open up a future issue to enforce permissions recursively on that folder and then write an adjacent test to test for that.

conorsch: Different umask in both environments most likely. Agreed on punt the check to a separate issue; we'll readd the check once the first-run config actually enforces it.

msheiny: k - i tweaked it a bit to better reflect the ansible code, see the latest commit i just pushed

conorsch:
Looks good to me! Test suite is still passing locally after the most recent changes.

.travis.yml
fowlslegs: I think shoehorning is the wrong word here when these both these requirements files are for test dependencies. App code and spec tests are both important and should be run not only on our development machines, but staging and prod as well. The requirements need to be installed for that to happen.

The more requirements lists we create, the greater room there is for conflict. testinfra should catch these, but we still have to resolve them manually. That's one of the reasons we use `pip-tools`, but even that approach is complicated by too many requirements lists. (Speaking of which, either way you should be using `pip-tools` in order to lock down your second-level dependencies for reproducible tests).

I think that isolation of the testing and app code dependencies is a good thing, that may be well solved by your suggestion of running tests from a separate machine. I think this is something we need to tackle separately, however. Adding `pytest-xdist` and `testinfra` to our `test-requirements.in` (everything else in `testinfra/requirements.txt` is already there except Ansible, which could be installed on its own) does not seem like a problematic move when afaik this lack of isolation has only been a problem in theory.

On the other hand, it would (i) be a sensible logical grouping, (ii) enable us to use `pip-tools` and our existing `pip_update.sh` script to update these periodically and easily along with our other requirements, and (iii) make it easier for developers to just run the testinfra suite on the staging and prod machines--if the tools aren't just there, we'll be less likely to go to the annoying process of installing them separately, and the testinfra suite will get less love and less use on these machines.

PR #1629, opened 2017-03-20
Test with tor browser

tkell:
This has:
- Updated the Ansible files to get `tbeselenium`, TBB, and a legacy version of FireFox.  
- Updated the functional tests to use TBB,  and fix failing tests caused by that.

Potential problem areas / odd things:
- In order to read from localhost and set cookies, I had to pass the following config to TBB:
```
pref_dict = {
        'network.proxy.no_proxies_on': '127.0.0.1',
        'browser.privatebrowsing.autostart': False
}
```
- For some reason, when creating the `tbselenium browser driver, the current directory gets changed, so I change it back.  
- I was not sure where to put the legacy FireFox, so it is in `/usr/bin/firefox45/`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ script:
   # travis needs the config.py file ran owned by root in other environments it is the
   # securedrop_user default www-data
 - sudo chown root:root securedrop/config.py
+# replacing with my own stuff to try to get a log file written
 - sudo sh -c "export DISPLAY=:1; cd securedrop/ && pytest tests/"
 - ./testinfra/test.py
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,6 @@ script:
   # travis needs the config.py file ran owned by root in other environments it is the
   # securedrop_user default www-data
 - sudo chown root:root securedrop/config.py
-# replacing with my own stuff to try to get a log file written
 - sudo sh -c "export DISPLAY=:1; cd securedrop/ && pytest tests/"
 - ./testinfra/test.py
 after_success:
--- a/install_files/ansible-base/roles/app-test/tasks/download_legacy_ff.yml
+++ b/install_files/ansible-base/roles/app-test/tasks/download_legacy_ff.yml
@@ -1,0 +1,18 @@
+---
+- name: Download Firefox 45.7.0.esr as a fallback if Tor Browser does not work
+  sudo: no
+  get_url:
+    dest: /home/vagrant/
+    url: https://ftp.mozilla.org/pub/firefox/releases/45.7.0esr/linux-x86_64/en-GB/firefox-45.7.0esr.tar.bz2
+    sha256sum: fbaa185abf9b7abc9d66ae037a34f192ced42f6e799d251253827cd0fbf3ae89
+
+# Hoping that we don't need dependencies because we've got them for FF 46
+
+# untar and move - the executable is `firefox` in the below folder
+- name: Extract Firefox 45.7.0esr
+  sudo: yes
+  unarchive:
+    copy: no
+    src: "/home/vagrant/firefox-45.7.0.esr.tar.bz2"
+    dest: "/usr/bin/firefox45/"
+
--- a/install_files/ansible-base/roles/app-test/tasks/download_legacy_ff.yml
+++ b/install_files/ansible-base/roles/app-test/tasks/download_legacy_ff.yml
@@ -5,13 +5,11 @@
     dest: /home/vagrant/
     url: https://ftp.mozilla.org/pub/firefox/releases/45.7.0esr/linux-x86_64/en-GB/firefox-45.7.0esr.tar.bz2
     sha256sum: fbaa185abf9b7abc9d66ae037a34f192ced42f6e799d251253827cd0fbf3ae89
-# Hoping that we don't need dependencies because we've got them for FF 46
 
 - name: Make target directory.
   sudo: yes
   file: path=/usr/bin/firefox45 state=directory
 
-# untar and move - the executable is `firefox` in the below folder.
 - name: Extract Firefox 45.7.0esr
   sudo: yes
   unarchive:
--- a/install_files/ansible-base/roles/app-test/tasks/download_legacy_ff.yml
+++ b/install_files/ansible-base/roles/app-test/tasks/download_legacy_ff.yml
@@ -5,14 +5,17 @@
     dest: /home/vagrant/
     url: https://ftp.mozilla.org/pub/firefox/releases/45.7.0esr/linux-x86_64/en-GB/firefox-45.7.0esr.tar.bz2
     sha256sum: fbaa185abf9b7abc9d66ae037a34f192ced42f6e799d251253827cd0fbf3ae89
-
 # Hoping that we don't need dependencies because we've got them for FF 46
 
-# untar and move - the executable is `firefox` in the below folder
+- name: Make target directory.
+  sudo: yes
+  file: path=/usr/bin/firefox45 state=directory
+
+# untar and move - the executable is `firefox` in the below folder.
 - name: Extract Firefox 45.7.0esr
   sudo: yes
   unarchive:
     copy: no
-    src: "/home/vagrant/firefox-45.7.0.esr.tar.bz2"
-    dest: "/usr/bin/firefox45/"
+    src: /home/vagrant/firefox-45.7.0esr.tar.bz2
+    dest: /usr/bin/firefox45/
 
--- a/install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
+++ b/install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
@@ -63,3 +63,12 @@
   copy:
     content: "{{ tbb_directory }}"
     dest: "/opt/.tbb_path_file"
+
+# If we're on Travis, CI will fail unless the TBB dir is owned by root.
+- name: Chown TBB directory to root, if on Travis
+  file:
+    dest: "{{ tbb_directory }}"
+    owner: root
+    group: root
+    recurse: yes
+  when: non_default_securedrop_user == "travis"
--- a/install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
+++ b/install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
@@ -53,11 +53,6 @@
   args:
     creates: "{{ tbb_directory }}/tor-browser_{{ tbb_locale }}"
 
-- name:  Save TBB directory to be read later.
-  copy:
-    content: "{{ tbb_directory }}"
-    dest: "/opt/.tbb_path_file"
-
 # If we're on Travis, CI will fail unless the TBB dir is owned by root.
 - name: Chown TBB directory to root, if on Travis
   file:
--- a/install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
+++ b/install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
@@ -2,19 +2,19 @@
 - name: Create tbb directory.
   sudo: true
   file:
-    path: "{{ fpsd_tbb_directory }}"
+    path: "{{ tbb_directory }}"
     state: directory
-    owner: "{{ fpsd_system_account }}"
-    group: "{{ fpsd_system_account }}"
+    owner: "{{ system_account }}"
+    group: "{{ system_account }}"
     mode: "0770"
 
-- name: Download TBB {{ fpsd_tbb_release }} tarball and signature.
+- name: Download TBB {{ tbb_release }} tarball and signature.
   get_url:
     url: "{{ item }}"
-    dest: "{{ fpsd_tbb_directory }}/{{ item|basename }}"
+    dest: "{{ tbb_directory }}/{{ item|basename }}"
   with_items:
-    - https://dist.torproject.org/torbrowser/{{ fpsd_tbb_release }}/tor-browser-linux{{ fpsd_tbb_arch }}-{{ fpsd_tbb_release }}_{{ fpsd_tbb_locale }}.tar.xz
-    - https://dist.torproject.org/torbrowser/{{ fpsd_tbb_release }}/tor-browser-linux{{ fpsd_tbb_arch }}-{{ fpsd_tbb_release }}_{{ fpsd_tbb_locale }}.tar.xz.asc
+    - https://dist.torproject.org/torbrowser/{{ tbb_release }}/tor-browser-linux{{ tbb_arch }}-{{ tbb_release }}_{{ tbb_locale }}.tar.xz
+    - https://dist.torproject.org/torbrowser/{{ tbb_release }}/tor-browser-linux{{ tbb_arch }}-{{ tbb_release }}_{{ tbb_locale }}.tar.xz.asc
 
 - name: Install gnupg-curl
   sudo: true
@@ -22,24 +22,24 @@
     name: gnupg-curl
     state: present
 
-- name: Download TBB {{ fpsd_tbb_release }} signing key.
+- name: Download TBB {{ tbb_release }} signing key.
   sudo: true
-  command: gpg --keyserver {{ fpsd_keyserver }} --recv-key {{ fpsd_tbb_signing_key }}
+  command: gpg --keyserver {{ keyserver }} --recv-key {{ tbb_signing_key }}
   register: gpg_import_tor_browser_devs_key_result
   changed_when: "'imported: 1' in gpg_import_tor_browser_devs_key_result.stderr"
 
-- name: Verify TBB {{ fpsd_tbb_release }} signature.
+- name: Verify TBB {{ tbb_release }} signature.
   command: >
     gpg --verify
-    tor-browser-linux{{ fpsd_tbb_arch }}-{{ fpsd_tbb_release }}_{{ fpsd_tbb_locale }}.tar.xz.asc
-    tor-browser-linux{{ fpsd_tbb_arch }}-{{ fpsd_tbb_release }}_{{ fpsd_tbb_locale }}.tar.xz
+    tor-browser-linux{{ tbb_arch }}-{{ tbb_release }}_{{ tbb_locale }}.tar.xz.asc
+    tor-browser-linux{{ tbb_arch }}-{{ tbb_release }}_{{ tbb_locale }}.tar.xz
   register: gpg_verify_result
   changed_when: false
   args:
-    chdir: "{{ fpsd_tbb_directory }}"
+    chdir: "{{ tbb_directory }}"
 
-- name: Extract TBB {{ fpsd_tbb_release }} archive.
+- name: Extract TBB {{ tbb_release }} archive.
   unarchive:
     copy: no
-    src: "{{ fpsd_tbb_directory }}/tor-browser-linux{{ fpsd_tbb_arch }}-{{ fpsd_tbb_release }}_{{ fpsd_tbb_locale }}.tar.xz"
-    dest: "{{ fpsd_tbb_directory }}"
+    src: "{{ tbb_directory }}/tor-browser-linux{{ tbb_arch }}-{{ tbb_release }}_{{ tbb_locale }}.tar.xz"
+    dest: "{{ tbb_directory }}"
--- a/install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
+++ b/install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
@@ -4,8 +4,8 @@
   file:
     path: "{{ tbb_directory }}"
     state: directory
-    owner: "{{ non_default_securedrop_user }}"
-    group: "{{ non_default_securedrop_user }}"
+    owner: "{{ securedrop_user }}"
+    group: "{{ securedrop_user }}"
     mode: "0770"
 
 - name: Stat TBB tarball
@@ -71,4 +71,4 @@
     owner: root
     group: root
     recurse: yes
-  when: non_default_securedrop_user == "travis"
+  when: securedrop_user == "travis"
--- a/install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
+++ b/install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
@@ -8,11 +8,6 @@
     group: "{{ securedrop_user }}"
     mode: "0770"
 
-- name: Stat TBB tarball
-  stat:
-    path: "{{ tbb_directory }}/tor-browser-linux{{ tbb_arch }}-{{ tbb_release }}_{{ tbb_locale }}.tar.xz"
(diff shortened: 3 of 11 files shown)


install_files/ansible-base/roles/app-test/tasks/download_legacy_ff.yml
```diff
@@ -0,0 +1,21 @@
+---
+- name: Download Firefox 45.7.0.esr as a fallback if Tor Browser does not work
+  sudo: no
+  get_url:
+    dest: /home/vagrant/
+    url: https://ftp.mozilla.org/pub/firefox/releases/45.7.0esr/linux-x86_64/en-GB/firefox-45.7.0esr.tar.bz2
+    sha256sum: fbaa185abf9b7abc9d66ae037a34f192ced42f6e799d251253827cd0fbf3ae89
+# Hoping that we don't need dependencies because we've got them for FF 46
```
fowlslegs: We should test fallback works as expected, and then remove this line.

securedrop/tests/functional/functional_test.py
```diff
@@ -31,19 +30,31 @@ class FunctionalTest():
...
     def _create_webdriver(self):
-        log_file = open('tests/log/firefox.log', 'a')
-        log_file.write(
-            '\n\n[%s] Running Functional Tests\n' % str(
-                datetime.now()))
+        log_file = open('tests/log/tbb.log', 'a')
+        log_msg = '\n\n[%s] Running Functional Tests\n' % str(datetime.now())
+        log_file.write(log_msg)
         log_file.flush()
-        firefox = firefox_binary.FirefoxBinary(log_file=log_file)
```
fowlslegs: You should preserve use of a FF log file, to aid in debugging TBB/FF issues in the future.

install_files/ansible-base/roles/app-test/tasks/main.yml
```diff
@@ -14,3 +14,7 @@
 - include: dev_setup_xvfb_for_functional_tests.yml
 
 - include: disable_mprotect_ff_plugin_container.yml
+
+- include: download_verify_tbb.yml
+
+- include: download_legacy_ff.yml
```
fowlslegs: Nit: `download_ff_esr.yml` is probably a better name.

fowlslegs:
Also, you can see this failed in Travis https://travis-ci.org/freedomofpress/securedrop/builds/213201178. For testing your Ansible, the way to do it is `vagrant destroy -f development && vagrant up development && vagrant provision development`. So you provision twice. It should succeed both times, and you should see `changed: 0` the second time (if it isn't zero and the changed tasks are not tasks you added, it's fine) meaning your tasks are idempotent. Then you can SSH in, run `./manage.py test` from `/vagrant/securedrop`, and if that succeeds, push your changes, and ping me for re-review.

Properly testing in the `staging` and `prod` machines is a bit more complicated and requires building a new app-code-package via `vagrant up build`, copying it from `build/` on your host to the target VM (I recommend the `vagrant scp` plugin), and installing it in the target VM. It's a "todo" to simplify this workflow, but for now if you can get it passing in Travis, and have tested it locally in a `development` VM as described above, I'll take care of testing it in these other environments for you.

tkell:
@fowlslegs, regarding installing `firefox-esr` with `apt`, I am getting told that:

```
 firefox-esr : Depends: libhunspell-1.3-0 (>= 1.3.3) but it is not going to be installed
               Depends: libstdc++6 (>= 4.9) but 4.8.4-2ubuntu1~14.04.3 is to be installed
               Recommends: gstreamer1.0-libav but it is not going to be installed
               Recommends: gstreamer1.0-plugins-good but it is not going to be installed```

Looking at the Ubuntu sources, [here](http://packages.ubuntu.com/search?keywords=libhunspell-1.3-0) suggests that `libhunspell` does not have the correct version for any Trusty sources.   Likewise for `libstdc++6`, [here](http://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=libstdc%2B%2B6&searchon=names).

I could try to install those dependencies from source, but I feel like that is just as bad as pinning a version of `firefox-esr`.   (and the Debian Jessie packages do not have 45.7.0esr, so we can't just apt install that).  Your thoughts?

**Comment by fowlslegs** [comment]
Here are two thoughts:

1. Download https://download.mozilla.org/?product=firefox-45.7.0esr-SSL&os=linux64&lang=en-US, extract to `/opt`, & `ln -s /opt/.../firefox-esr /usr/bin/firefox-esr`. Pin the SHAsums in Ansible.

2. Just use FF from Ubuntu repos.

Personally, I really think you should just go with the latter for simplicity, and since hopefully we just won't have to be falling back. I think that setting a few profile options using regular `selenium` for FF stable, such as enabling private browsing mode, would be helpful in making it behave a bit more like TBB in potential edge cases.

Looks like you're getting close, so once things are working you should push to your branch, and then try to rebase on the latest develop. Looking at https://github.com/freedomofpress/securedrop/pull/1527 (read all the comments and/ or just look at the changes because the direction changed a bit through progression from the initial comment) before you rebase should be helpful.

Thanks for persevering thus far; resolving this seems to have involved a few curve-balls.

**Comment by tkell** [comment]
@fowlslegs, got this to a better place, I think.  LMK your thoughts, and then I will wrestle through the rebase.

- There are 5 items that say `changed`, but none of them are mine.
- Functional tests work with both TBB and Firefox.
- I don't know why the ansible `unarchive` command did not want to work with the TBB tar, but it did not.
- Also went with a regular apt call for Firefox, as per your comment above.

**Comment by fowlslegs** [comment]
Looks good to me if you want to go ahead and rebase. The fact `unarchive` is not working for you is weird because that tasklist is copied from https://github.com/freedomofpress/fingerprint-securedrop/blob/master/roles/crawler/tasks/download-verify-tbb.yml, which works fine. Anyway, don't worry about that for now. I'll take another look once this is rebased and see if I can figure out what's going wrong.

**Comment by tkell** [comment]
@fowlslegs, rebased.  Thanks for all your help with this!

(And for sure, the unarchive thing is super-duper odd.  Computers.)

**Comment by tkell** [comment]
Grumble, fails because there's no user named `vagrant` on Travis.  Working on it.

**Comment by tkell** [comment]
Grumble++.  Travis is not happy with something to do with Firefox profiles, and/or Firefox log files.  These both work on vagrant invoked locally, which is odd and frustrating.  Will work on them more this weekend.

**Comment by fowlslegs** [comment]
I think you just want `os.path.abspath(os.path.join(__file__, '../log/firefox.log'))`.

**Comment by tkell** [comment]
OK, so on Travis, we never manage to connect to the WebDriver.  This works fine on my Vagrant. 
 What is Travis doing that is different?

**Comment by tkell** [comment]
Phew.  @fowlslegs, all yours for a final review.

File: securedrop/tests/functional/functional_test.py [hunk 4]
```diff
@@ -73,7 +85,7 @@ def _create_firefox_webdriver(self, log_file):
 
     def _create_webdriver(self):
         log_file_path = 'tests/log/firefox.log'
-        abs_log_file_path = os.path.abspath(log_file_path)
+        abs_log_file_path = os.path.abspath(os.path.join(__file__, '../../log/firefox.log'))
```
fowlslegs: My bad earlier, but this should be `abspath(join(dirname(__file__), '../log/firefox.log'))`. Have two `..` works, but is a little funky logically because then we're calling `abspath` on `/.../securedrop/tests/functional/functional_test.py/../../log/firefox.log`.

.travis.yml
```diff
@@ -33,6 +33,7 @@ script:
   # travis needs the config.py file ran owned by root in other environments it is the
   # securedrop_user default www-data
 - sudo chown root:root securedrop/config.py
+# replacing with my own stuff to try to get a log file written
```
fowlslegs: Not sure what this comment is explaining--perhaps an artifact from earlier troubleshooting?

install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
```diff
@@ -0,0 +1,69 @@
+---
+- name: Create system account user
```
fowlslegs: This code was written with the use case of supporting Tor Browser and multiple human users and a system account. Here we'll only be running the tests as a single user, whether that's `travis` or `vagrant`, so to simplify things it's best if we don't create this system account and perform operations as it.

securedrop/tests/functional/functional_test.py
```diff
@@ -33,20 +34,69 @@ class FunctionalTest():
...
-    def _create_webdriver(self):
-        log_file = open(join(LOG_DIR, 'firefox.log'), 'a')
-        log_file.write(
-            '\n\n[%s] Running Functional Tests\n' % str(
-                datetime.now()))
-        log_file.flush()
+
+    def _create_tor_browser_webdriver(self, abs_log_file_path):
+        # Creating the TorBrowser driver changes the working directory,
+        # so we need to save and then restore the current one.
```
fowlslegs: Why is this actually a problem? What happens if we don't change back directories?

securedrop/tests/functional/functional_test.py
```diff
@@ -33,20 +34,69 @@ class FunctionalTest():
...
+        old_dir = os.getcwd()
+
+        # Read the TorBrowser location from Ansible
```
fowlslegs: Why not just hardcode the static directory TBB lives under and get rid of this suite as well as the task that writes this file above?

securedrop/tests/functional/functional_test.py
```diff
@@ -33,20 +34,69 @@ class FunctionalTest():
...
+        with open('/opt/.tbb_path_file') as f:
+           path_to_tbb = f.readline().strip()
+        path_to_tbb = path_to_tbb + os.path.sep + "tor-browser_en-US"
+        path_to_tbb = os.path.abspath(path_to_tbb)
+
+        # Due to Travis having opinions, we need to chown the tor dir
+        # to be owned by root when we're running CI.
```
fowlslegs: This is much better done by Ansible. First, make the change I recommend regarding removing the system account use above, then see if this is still needed. If it is, then `become: True` in Ansible when running under Travis _and_ in the relevant Ansible explain exactly why it's necessary (more info than "Travis has opinions" would be helpful :smile:).

securedrop/tests/functional/functional_test.py
```diff
@@ -33,20 +34,69 @@ class FunctionalTest():
...
+        if '/home/travis/' in path_to_tbb:
+            for root, dirs, files in os.walk(path_to_tbb):
+                for d in dirs:
+                    os.chown(os.path.join(root, d), 0, 0)
+                for f in files:
+                    os.chown(os.path.join(root, f), 0, 0)
+
+        # Don't use Tor when reading from localhost,
+        # and turn off private browsing.
+        # We need this to make functional tests work,
```
fowlslegs: Be more explicit here. We need to turn off private browsing because we won't be able to access the browser's cookies in private browsing mode. Since we use session cookies in SD anyway (in private browsing mode all cookies are set as session cookies), this should not affect session lifetime.

securedrop/tests/functional/functional_test.py
```diff
@@ -33,20 +34,69 @@ class FunctionalTest():
...
+        os.chdir(old_dir)
+
+        return driver
+
+    def _create_firefox_webdriver(self, log_file):
         firefox = firefox_binary.FirefoxBinary(log_file=log_file)
         return webdriver.Firefox(firefox_binary=firefox)
 
+    def _create_webdriver(self):
+        log_file_path = 'tests/log/firefox.log'
```
fowlslegs: This object reference is created and then nothing more happens.

securedrop/tests/functional/functional_test.py
```diff
@@ -33,20 +34,69 @@ class FunctionalTest():
...
+        abs_log_file_path = os.path.abspath(os.path.join(__file__, '../../log/firefox.log'))
+        log_file = open(abs_log_file_path, 'a')
```
fowlslegs: You should use the `with` context manager here so this file is closed properly after use.

securedrop/tests/functional/functional_test.py
```diff
@@ -33,20 +34,69 @@ class FunctionalTest():
...
+        log_msg = '\n\n[%s] Running Functional Tests\n' % str(datetime.now())
+        log_file.write(log_msg)
+        log_file.flush()
+
+        if 'SD_USE_FALLBACK_BROWSER' in os.environ:
+            driver = self._create_firefox_webdriver(log_file)
```
fowlslegs: Why are you passing a file object here instead of a string?

securedrop/tests/functional/functional_test.py
tkell: I will swear that this dir change broke the tests - maybe the switch to pytest fixed it?  Regardless, I've taken it out.

securedrop/tests/functional/functional_test.py
tkell: I recall that I had it hard-coded, and then changed it to

`tbb_directory: "/home/{{ ansible_user|default(ansible_ssh_user|default(lookup('env', 'USER'))) }}/.local/tbb"`

I wasn't sure how to recreate that logic with python, so I saved the result out to a file.

tkell:
@fowlslegs, fixes made except the path where we put TBB - see my query above.  Thanks!

securedrop/tests/functional/functional_test.py
fowlslegs: `abspath(join(expanduser('~'), '.local/tbb'))`

securedrop/tests/functional/functional_test.py
```diff
@@ -33,19 +34,50 @@ class FunctionalTest():
...
-            '\n\n[%s] Running Functional Tests\n' % str(
-                datetime.now()))
-        log_file.flush()
-        firefox = firefox_binary.FirefoxBinary(log_file=log_file)
-        return webdriver.Firefox(firefox_binary=firefox)
+        abs_log_file_path = os.path.abspath(join(dirname(__file__), '../log/firefox.log'))
+        with open(abs_log_file_path, 'a') as f:
+            log_msg = '\n\n[%s] Running Functional Tests\n' % str(datetime.now())
+            f.write(log_msg)
+            f.flush()
```
fowlslegs: You shouldn't need flush here. Python will automatically flush here before closing the file.

install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
```diff
@@ -0,0 +1,74 @@
...
+- name: Create tbb directory.
+  sudo: true
+  file:
+    path: "{{ tbb_directory }}"
+    state: directory
+    owner: "{{ non_default_securedrop_user }}"
+    group: "{{ non_default_securedrop_user }}"
+    mode: "0770"
+
+- name: Stat TBB tarball
```
fowlslegs: This line should not be necessary as `get_url` below is idempotent.

install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
```diff
@@ -0,0 +1,74 @@
...
+- name: Download TBB {{ tbb_release }} tarball and signature.
+  get_url:
+    url: "{{ item }}"
+    dest: "{{ tbb_directory }}/{{ item|basename }}"
+  with_items:
+    - https://dist.torproject.org/torbrowser/{{ tbb_release }}/tor-browser-linux{{ tbb_arch }}-{{ tbb_release }}_{{ tbb_locale }}.tar.xz
+    - https://dist.torproject.org/torbrowser/{{ tbb_release }}/tor-browser-linux{{ tbb_arch }}-{{ tbb_release }}_{{ tbb_locale }}.tar.xz.asc
+  when: not tbb_stat.stat.exists
+
+- name: Install gnupg-curl
```
fowlslegs: This task should probably be moved to wherever we're installing other dependencies.

install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
```diff
@@ -0,0 +1,74 @@
...
+  args:
+    creates: "{{ tbb_directory }}/tor-browser_{{ tbb_locale }}"
+
+- name:  Save TBB directory to be read later.
+  copy:
+    content: "{{ tbb_directory }}"
+    dest: "/opt/.tbb_path_file"
+
+# If we're on Travis, CI will fail unless the TBB dir is owned by root.
+- name: Chown TBB directory to root, if on Travis
```
fowlslegs: Can you provide a traceback of the error?

install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
```diff
@@ -0,0 +1,74 @@
...
+  command: >
+    gpg --verify
+    tor-browser-linux{{ tbb_arch }}-{{ tbb_release }}_{{ tbb_locale }}.tar.xz.asc
+    tor-browser-linux{{ tbb_arch }}-{{ tbb_release }}_{{ tbb_locale }}.tar.xz
+  register: gpg_verify_result
+  changed_when: false
+  args:
+    chdir: "{{ tbb_directory }}"
+
+- name: Install the package "xz utils"
```
fowlslegs: What happens exactly when you try
https://github.com/freedomofpress/fingerprint-securedrop/blob/4c8df3d55afad04075674dbd5ce1899d38155256/roles/crawler/tasks/download-verify-tbb.yml#L37?

install_files/ansible-base/roles/app-test/tasks/main.yml
```diff
@@ -14,3 +14,7 @@
 - include: dev_setup_xvfb_for_functional_tests.yml
 
 - include: disable_mprotect_ff_plugin_container.yml
+
+- include: download_verify_tbb.yml
+
+- include: download_ff_esr.yml
```
fowlslegs: I'm thinking this tasklist should get merged with `dev_setup_xvfb_for_functional_tests`, which could probably be renamed to something like `install_functional_test_dependencies`. Do also remove the FF 46.0.1 related tasks as that's what these changes are intended to replace.

install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
```diff
@@ -0,0 +1,74 @@
+---
+- name: Create tbb directory.
+  sudo: true
+  file:
+    path: "{{ tbb_directory }}"
+    state: directory
+    owner: "{{ non_default_securedrop_user }}"
```
fowlslegs: You want `securedrop_user`.

securedrop/tests/functional/functional_test.py
```diff
@@ -33,19 +34,50 @@ class FunctionalTest():
...
+                     'browser.privatebrowsing.autostart': False
+                    }
+        driver = TorBrowserDriver(path_to_tbb,
+                                  pref_dict=pref_dict,
+                                  tbb_logfile_path=abs_log_file_path)
+        return driver
+
+    def _create_firefox_webdriver(self, abs_log_file_path):
+        with open(abs_log_file_path, 'a') as f:
+            firefox = firefox_binary.FirefoxBinary(log_file=f)
```
tkell: When I pass `abs_log_file_path`, I get told that `AttributeError: 'str' object has no attribute 'fileno'`.

Looking at the [docs](https://seleniumhq.github.io/selenium/docs/api/py/webdriver_firefox/selenium.webdriver.firefox.firefox_binary.html), there appears to be no way to pass a path to a file, so I guess we're stuck with sending the file handle.

fowlslegs: The docs you linked say to send a file object, so what you're doing here is actually correct--my bad. I just got confused because `TorBrowserDriver` takes a string for the `log_file` keyword argument, whereas `FirefoxBinary` takes a file object--the former would ideally be made consistent with the latter since the latter is its upstream, but irrelevant to what we want to do here, and no big deal.

I think you'll need to back out of using `with` here, since our `FirefoxBinary` initializer is going to create an attribute in our `firefox` instance that is an obj ref to the same file object that `fh` is an obj ref to. When the `with` context manager exits, it will call `fh.close()`, which might prevent logging later on... although it seems like that would cause some fatal exceptions somewhere. I haven't looked at the code in a while, but can do that when I'm done with the current thing. Anyway, here's what I mean:

```
In [25]: with open('README.MD') as fh:
    ...:     x = fh
    ...:

In [26]: x.write('yo')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-26-f0d743a38770> in <module>()
----> 1 x.write('yo')

ValueError: I/O operation on closed file.
```

install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
tkell: Hm.  I don't see an obvious place for it?  

`gnupg-curl` is only needed for the TBB install, so it seems like it could stay in that file (maybe at the top?).  Happy to move it if you have a suggestion though!

securedrop/tests/functional/functional_test.py
```diff
@@ -33,19 +34,50 @@ class FunctionalTest():
...
         s = socket.socket()
-        s.bind(("localhost", 0))
+        s.bind(('127.0.0.1', 0))
         port = s.getsockname()[1]
         s.close()
         return port
 
+
+    def _create_tor_browser_webdriver(self, abs_log_file_path):
+        # Read the TorBrowser location from Ansible
```
tkell: `expanduser` fails for some reason related to the fact that TBB changes directories when it creates a driver.  I found a workaround using `getpass.getuser()` though.

install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
tkell: Got an error.  The message was along the lines of `"Failed to find handler for <filename>. Make sure the required command to extract the file is installed."

install_files/ansible-base/roles/app-test/tasks/download_verify_tbb.yml
tkell: Not now, as I've got work travel until the start of May - and it took me a few hours to get Travis to spit out the log file I needed to read.

I feel like the error message was pretty clear in that it expected the files to be owned by `0`?

install_files/ansible-base/roles/app-test/tasks/main.yml
tkell: ... I think I may have outsmarted myself with trying to just get latest FF from apt.  I feel like we need FF 45 or 46 in order to use Selenium 2.53, as latest FF needs Geckodriver.  

We could keep the current FF 46, as we're now only using it as a fallback, or try to work out what Geckodriver needs?  But it may well need a more recent version of Selenium as well - which seems hard to have as well as 2.53.

fowlslegs: @garrettr suggested dropping support for a fallback browser, and I agree especially because it would simplify the Ansible and Python here a moderate amount. Tor Browser or bust!

PR #1630, opened 2017-03-20
Add note on removing speakers from the SVS

fowlslegs:
Wanted to get more feedback on this before we consider merge. Should we include information on why this is a good idea and helps enforce the airgap? (It's because of cooperative side-channels for information exfiltration using ultrasonic frequencies.) Also, should I just merge this line into the paragraph above it if we're not going to expand on it?

My inclinations, responding to my own questions, are no and yes, respectively.

And one more--is this process too difficult on some machines to expect of admins?

### DIFF
--- a/docs/set_up_svs.rst
+++ b/docs/set_up_svs.rst
@@ -18,12 +18,12 @@ adapters. If the machine has network ports you can't physically remove, you
 should clearly cover these ports with labels noting not to use them. For an even
 safer approach, fill a port with epoxy to physically disable it. We also
 recommend you remove the speakers from the device (or just cut the audio cables
-if that's easier) . This is to prevent exfiltration of data from the airgap via
-ultrasonic (i.e., above the range of human hearing) waves (for more read
-https://arstechnica.com/security/2013/12/scientist-developed-malware-covertly-jumps-air-gaps-using-inaudible-sound/).
-If you have questions about repurposing hardware for the *Secure Viewing
-Station*, contact the `Freedom of the Press Foundation
-<https://securedrop.org/help>`__.
+if that's easier). This is to prevent `exfiltration of data from the airgap via
+ultrasonic audio
+<https://arstechnica.com/security/2013/12/scientist-developed-malware-covertly-jumps-air-gaps-using-inaudible-sound/>`__,
+which cannot be heard by humans. If you have questions about repurposing
+hardware for the *Secure Viewing Station*, contact the `Freedom of the Press
+Foundation <https://securedrop.org/help>`__.
 
 You should have a Tails drive clearly labeled “SecureDrop Secure Viewing
 Station”. If it's not labeled, label it right now, then boot it on the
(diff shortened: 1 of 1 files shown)


fowlslegs:
Note that this concern becomes mostly irrelevant in the context of the hypothetical, Qubes-based JW/SVS machine. If the compartmentalization can be broken in Qubes, then speaker access will be not be as much a concern as network access. My sense is that this is a ways off, however, and think it's better to include this recommendation in the meantime.

docs/set_up_svs.rst
```diff
@@ -21,6 +21,9 @@ epoxy to physically disable it. If you have questions about repurposing
 hardware for the *Secure Viewing Station*, contact the `Freedom of the
 Press Foundation <https://securedrop.org/help>`__.
 
+We also recommend you remove all speakers from the device, instead using
+earphones to listen to audio submissions.
```
redshiftzero: I think this is fine to include, ideally with a link to a blog/article/paper that explains to admins what the concern is. Hopefully as the admins are taking apart the SVS anyway to pull out other hardware then doing this as well is not too difficult (with that said, the list of modifications to the SVS is becoming _quite_ onerous...)

This flows best if it is moved to after the sentence ending in "physically disable it.".

conorsch: Agreed, let's get it in, but with an explanation about _why_ we recommend it.

fowlslegs:
https://arxiv.org/abs/1608.03431 Bring in the concrete!

fowlslegs:
Also, ready for re-review @redshiftzero @conorsch

docs/set_up_svs.rst
```diff
@@ -12,14 +12,18 @@ drive or a DVD) and physically transfer the *Data Transfer Device* to
...
-epoxy to physically disable it. If you have questions about repurposing
-hardware for the *Secure Viewing Station*, contact the `Freedom of the
-Press Foundation <https://securedrop.org/help>`__.
+internal hard drive, we recommend that you physically remove any any internal
+storage devices or networking hardware such as wireless cards or Bluetooth
+adapters. If the machine has network ports you can't physically remove, you
+should clearly cover these ports with labels noting not to use them. For an even
+safer approach, fill a port with epoxy to physically disable it. We also
+recommend you remove the speakers from the device (or just cut the audio cables
+if that's easier) . This is to prevent exfiltration of data from the airgap via
```
redshiftzero: Extra space here:

`easier) .` -> `easier).`

docs/set_up_svs.rst
```diff
@@ -12,14 +12,18 @@ drive or a DVD) and physically transfer the *Data Transfer Device* to
...
+ultrasonic (i.e., above the range of human hearing) waves (for more read
+https://arstechnica.com/security/2013/12/scientist-developed-malware-covertly-jumps-air-gaps-using-inaudible-sound/).
```
redshiftzero: Make this a link?

conorsch: Definitely make it a link, I don't like the in-line URL either.

conorsch: > This is to prevent **exfiltration of data from the airgap via ultrasonic audio**, which cannot be heard by humans. 

The bold part should be the link.

PR #1632, opened 2017-03-23
Add PR template

redshiftzero:
GitHub has this nice [pull request template](https://github.com/blog/2111-issue-and-pull-request-templates) feature. This is an easy way to let contributors know what kinds of information they should include to aid in review as well as what checks they should be doing before they submit their PR.


conorsch:
This looks good. Long-term I'd like to link out to the docs on the test suites, but we need to _write_ those docs first :innocent:, so I'm fine with merging as-is for a template.

PR #1633, opened 2017-03-23
Fix typo in install.sh

xanderhades:
Missing double quote ending at line 345

### DIFF
--- a/tails_files/install.sh
+++ b/tails_files/install.sh
@@ -342,7 +342,7 @@ function create_desktop_shortcuts()
 
   # Remove historic launchers and icons
   rm -f "${amnesia_desktop}/document.desktop" \
-        "${amnesia_home}/.local/share/applications/document.desktop
+        "${amnesia_home}/.local/share/applications/document.desktop"
 
   # make it all persistent
   sudo -u amnesia mkdir -p "${tails_live_dotfiles}/Desktop"


conorsch:
Good catch, @xanderhades! Looks like this slipped in as part of #1395, and fortunately only affects the `develop` branch (i.e. hasn't shipped with a release yet). See #1614 for a list of complications with the interface renaming attempt.

By the way, we could have caught this was a simple `bash -n tail_files/install.sh` in CI.

fowlslegs:
Thanks! This bug never should have slipped in. Time for us to start at least linting our shell scripts in CI, even if proper testing is less practical given the nature of their use. @freedomofpress/securedrop-maintainers what do you think of running https://github.com/caarlos0/shell-ci-build in Travis? Shellcheck caught this bug fwiw.

Good thing this is only in `develop` and never made it into a release!

xanderhades:
Are you sure it was only in develop branch? I cloned the repo and had the install.sh script error while trying to configure a journalist Tails USB .

conorsch:
@xanderhades That's why I made sure to mention it—in the 0.3.12 release, the bug doesn't exist. If you're installing a production instance, make sure to use the latest stable tag, described here: https://docs.securedrop.org/en/stable/set_up_admin_tails.html#verify-the-release-tag

xanderhades:
Ahhhh I missed that.
Thanks Conor!

PR #1641, opened 2017-04-06
Fix coveralls checks

redshiftzero:
## Status

Ready for review

## Description of Changes

Small followup PR to recent changes to the SecureDrop application tests: 
* Coveralls checks were not working due to an outdated relative path in `.travis.yml`, which means that the coverage results were not sent to Coveralls. 
* The SecureDrop application tests were being included in the coverage calculation. 

## Testing

1. Look at the checks. 
2. Notice that Coveralls is working. 
3. Merge this PR! :sparkles:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ script:
 - sudo sh -c "export DISPLAY=:1; cd securedrop/ && pytest tests/"
 - ./testinfra/test.py
 after_success:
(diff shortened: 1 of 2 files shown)


conorsch:
Magnificent! Thanks for fixing, @redshiftzero. The change to moving the tests directory first came in with #1527, but in #1612 I incompletely reverted the rename, breaking the coveralls tests in CI. Glad we're back in action now.

PR #1643, opened 2017-04-07
Remove unused static resources, cleanup configs

fowlslegs:
* The "tipbox" images are not used anywhere.
* CSS compiled by sass is placed under `/static/css/` and the JS we use is under
  `/static/js`. `/static/gen/` is a legacy-use folder.
* Tor Browser doesn't download the fonts we provide--probably an
  anti-exploitation mechanism--so no point keeping them around.


conorsch:
Thanks for tackling these, @fowlslegs. I'm rebuilding staging debs frequently as part of #1614, and will keep these changes in mind while I step through edits there. 

At a glance, the `tipbox/` changes look good, but I disagree that the `static/gen/` allowances in AppArmor can be removed—they seem to be required even under staging as of current develop. Will continue to test and report back.

fowlslegs:
> I disagree that the static/gen/ allowances in AppArmor can be removed

Somehow missed that Flask-Assets (which I want to get rid of--a different story) puts the JS it minifies and merges in that dir (I did `grep`). The naming scheme used for that operation is less than ideal IMO, but we'll leave it. Will back out that backing out.

conorsch:
> Somehow missed that Flask-Assets (which I want to get rid of--a different story) 

I sympathize, but let's keep it for 0.4 to simplify the changes necessary to get the branches reconciled, so we can get back to rapid development. 

The changes in this PR partly satisfy the concerns listed in #1646, so I'm going to tweak this branch as part of review so we can close that issue. 

>  Will back out that backing out.

No sweat, I'll handle it—I've got lots of recent testing feedback, and will keep the commit messages verbose to explain the changes as they land.

conorsch:
Why remove _all_ the fonts? At a glance, the `fonts/fontawesome-webfont.eot` is still used, perhaps others. I think this PR should remove the `tipbox/` entries and nothing else, since the removals are overly aggressive on several fronts.

conorsch:
> The changes in this PR partly satisfy the concerns listed in #1646, so I'm going to tweak this branch as part of review so we can close that issue.

I take that back: we need the staging environment fully operational in order to evaluate the consequences of these changes, so I'm going to submit a separate PR to resolve #1646, and block here. Once the up-and-coming PR is merged, we can use that work to test the changes presented here.

conorsch:
Rebased on top of latest develop, to include changes from #1649.

fowlslegs:
Ready for re-review.

conorsch:
Thanks @fowlslegs. There are yet more AppArmor allowances that must be included (and were missed in #1646), so stepping through all the changes now.

conorsch:
The blanket approach to removing fonts causes a lot of problems, including 404s on the Journalist Interface when fetching the console. In general, the changes in this PR are not well tested and do not address a known problem, so rather than continue to review, I'm going to close this out and focus on 0.4-related changes. To be specific:

> CSS compiled by sass is placed under `/static/css/` and the JS we use is under
  `/static/js`. `/static/gen/` is a legacy-use folder.

Not true, but addressed already by 0b2fcec.

> Tor Browser doesn't download the fonts we provide--probably an anti-exploitation mechanism--so no point keeping them around.

Not true. Compare the screenshots below. 

### With fonts preserved (as in current develop)

journalist-interface-fewer-404s

### With fonts removed (as proposed in this PR)

journalist-interface-font-404s

It's true that webfont loading is blocked if Journalists set the Security Slider to High, but we can't assume that's been done. The Journalist Interface already requires JS to be usable, so it's inappropriate to assume that the Journalist user account will be navigating with a browser as hardened as the security we recommend for Source user accounts. Besides, even if we chose to remove the fonts, that change should make sure to remove all references to the fonts, including in CSS files where appropriate.

> The "tipbox" images are not used anywhere.

This is probably correct, but given the inaccurate assumptions elsewhere in these changes, I'd want to be very certain this is so before merging anything. @fowlslegs if you feel strongly that the tipbox images should be removed, please open another issue to track that change. Right now, trying to stay on track for the 0.4 branch reconciliation project.

conorsch:
The changes in c88373c are necessary for a working and interactive Staging environment, so I'll resubmit that as a separate PR. That change _should_ have made it into #1646, but didn't.

fowlslegs:
I'm going to re-open this and add three commits that:

1. [ ] replace use of SVGs in the JI w/ PNGs.
2. [ ] add necessary AppArmor allowances for any new PNGs, and
3. [ ] remove references to the webfonts in the CSS.

The PR can be slated for merge into 0.4.1--I agree it's not a 0.4 essential and is best put off until then. Going to label as blocked and tag for that milestone as well.

redshiftzero:
If these changes should be in 0.4.1, let's 1) make sure there are issues for each proposed change here, 2) decide as a team during sprint prep that it should be addressed in 0.4.1 and 3) create well-scoped PRs addressing each issue. Until then, I'm closing this.

PR #1645, opened 2017-04-10
Update Grub PaX flags in grsecurity role

### LINKED ISSUE #1621: PaX flags inconsistently declared on GRUB binaries
The GRUB binaries require PaX flags when running under grsecurity, in order to avoid errors when installing new kernels and running `update-grub` via the postinst kernel hooks. We're now enforcing those flags via the `securedrop-grsec` metapackage, which is maintained in a separate repo (see discussion at https://github.com/freedomofpress/ansible-role-grsecurity/issues/90). 

The Ansible config contains logic to set the PaX flags at install time, and the current implementation uses less precise flags than are present in the grsec metapackage. The flags should be updated to match those set in the grsec repo, and ideally be set idempotently to provide meaningful "changed" output during run of the playbooks.

fowlslegs:
PaX flags here now match what we set in the securedrop-grsec metapackage (see
https://github.com/freedomofpress/ansible-role-grsecurity/pull/91/commits/406078267426faf01c16e21eb5a32f129795a045). Fixes #1621.

### DIFF
--- a/install_files/ansible-base/roles/grsecurity/tasks/paxctl.yml
+++ b/install_files/ansible-base/roles/grsecurity/tasks/paxctl.yml
@@ -23,15 +23,5 @@
 - name: Adjust paxctl headers on grub binaries.
   command: paxctl -zCE {{ item.item }}
   with_items: "{{ paxctl_grub_header_check.results }}"
-  # The desired flags should include:
-  #   - p NOPAGEEXEC
-  #   - m NOMPROTECT
-  #   - x NORANDEXEC
-  #   - e NOEMUTRAMP
-  # After running `paxctl -Cpm` in VirtualBox, all four of those flags are set.
-  # However, the current paxctl command only enforces p and m.
-  # The x and e flags are present by default, but can be removed via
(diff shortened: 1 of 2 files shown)


install_files/ansible-base/roles/grsecurity/tasks/paxctl.yml
```diff
@@ -32,5 +33,5 @@
   # The x and e flags are present by default, but can be removed via
   # `paxctl -z <binary>`. TODO: We may want to update the command
   # to set -Cpmxe explicitly. Warrants further testing on hardware.
-  when: "item.stdout != '- PaX flags: -p---m-x-e-- [{{ item.item }}]' or
+  when: "item.stdout != '- PaX flags: -p---m--E--- [{{ item.item }}]' or
```
conorsch: The PaX output doesn't match what we're trying to set, so this task will run every time, breaking idempotence.

conorsch:
The PaX flag checks in the config tests are currently marked as "expected failure," so I'll update those to run and expect a pass.

PR #1649, opened 2017-04-17
Fix locally built deb packages under staging environment

conorsch:
## Status

Ready for review

## Description of Changes

Updates the staging environment configuration, particularly the package build logic, so that the application code runs under Apache, as intended. Since merging #1464, we now have a tighter loop on testing app code changes under the Apache setup. Removed the strategy where AppArmor profiles run under "complain" (rather than "enforce"), since most of the problems fixed in this PR slipped in due to the complain strategy.

Resolves #1646. 

Changes proposed in this pull request:

* Stop setting "complain" on AppArmor profiles under staging.
* Remove new `util.py` app code file, which didn't have an AppArmor entry.
* Add AppArmor entries for Flask-Asset-generated CSS and JS files, and include those empty directories in the package build logic.
* Don't assume the `CUSTOM_NOTIFICATION` value will be defined in `config.py`; it will in newly installed instances, but not on those that upgraded from a prior version.

## Testing

How should the reviewer test this PR?

1. Create staging environment from scratch and provision it fully.
2. Ensure that the app tests and config tests pass on all staging hosts.

## Deployment

Any special considerations for deployment? Consider both:

These changes are intended to restore the usefulness of the staging environment, so that locally built deb packages can be used to evaluate how code changes affect a prod-like serving environment, including Apache vhosts, Tor Onion Services, and AppArmor profiles. These changes are necessary to review e.g. #1642 and #1643.

### If you made changes to the app code:

- [x] Unit and functional tests pass on the development VM
- [x] Unit and functional tests pass on the app-staging VM

### If you made changes to the system configuration:

- [x] Testinfra tests pass on the development VM
- [x] Testinfra tests pass on the app-staging VM
- [x] Testinfra tests pass on the mon-staging VM
- [x] Testinfra tests pass on the build VM

PR #1650, opened 2017-04-18
Migrates Document -> Journalist interface via preinst

### LINKED ISSUE #1614: Investigate Document -> Journalist Interface renaming side-effects

conorsch:
Implements a hook-based migration strategy for renaming the Document
Interface to Journalist Interface (for consistency with the role-based
"Source Interface" name). Services migrated include:

  * ATHS config
  * Apache vhost config

These config updates are performed as part of the `preinst` stage in the
`securedrop-app-code` deb package. Care has been taken to ensure that
the config update logic is idempotent, so that the same preinst hooks
can run on any future version of SecureDrop and ensure the same results.

## Status

Ready for review.

## Description of Changes

Fixes #1614.

## Testing

1. Provision staging VMs based on the 0.3.12 tag.
2. Check out this PR and install locally build deb packages.
3. Run all config and app test suites, confirm passing.

In manual testing locally, config test suites showed green across the board, save a single failure on App Staging and Mon Staging, caused by an iptables rule being updated between 0.3.12 and current develop. That's an acceptable failure; rerunning the Ansible playbooks in their entirety would resolve.

## Deployment

The strategy here is to migrate currently running production instances over to the new Interface naming scheme (i.e. Document -> Journalist interface). We cannot coordinate with Admins to install the 0.4 update manually, since the servers will automatically install any new deb packages via the nightly updates. Thus the migration logic must exist in the app-code package, or be backed out of develop. See #1614 for discussion.

Will update the testing results after #1649 is merged. Rebasing on top of that branch locally, test output was great. Omitting that successful feedback from the manual test confirmation below until the passing is true of the PR branch.

- [ ] Unit and functional tests pass on the app-staging VM

- [x] Testinfra tests pass on the build VM
- [ ] Testinfra tests pass on the app-staging VM
- [ ] Testinfra tests pass on the mon-staging VM

### DIFF
--- a/install_files/securedrop-app-code/DEBIAN/preinst
+++ b/install_files/securedrop-app-code/DEBIAN/preinst
@@ -46,6 +46,32 @@ function permanently_disable_swap() {
     perl -i -apne '$F[2] eq "swap" && /^[^\s#]+?/ && s/^/#/g' /etc/fstab
 }
 
+function convert_document_to_journalist_interface() {
+    # Helper function to migrate the old "Document Interface" config files
+    # to the new "Journalist Interface" naming scheme. Affects tor and apache.
+
+    if [[ -d /var/lib/tor/services/document ]]; then
+        # Move tor service hostname and keys.
+        mv /var/lib/tor/services/document /var/lib/tor/services/journalist
+        # Update torrc to use new tor service hostname directory.
+        perl -pi -e 's#^(HiddenServiceDir /var/lib/tor/services/)document#$1journalist#' /etc/tor/torrc
+        # Bounce tor service.
+        service tor restart
+    fi
+
+    if [[ -f /etc/apache2/sites-available/document.conf ]]; then
+        # Stop apache prior to migrating vhost.
+        service apache2 stop
+        # Convert apache vhost settings to use the new interface name.
+        perl -p -e 's/document/journalist/' /etc/apache2/sites-available/document.conf \
+            > /etc/apache2/sites-available/journalist.conf
+        rm -f /etc/apache2/sites-available/document.conf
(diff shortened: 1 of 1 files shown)


conorsch:
Rebased on top of latest develop, to include changes from #1649, which are required for testing this PR.

conorsch:
Rebased on top of latest develop. Now going to step through the entire testing check list. @redshiftzero would you mind doing the same? I'd like more eyes on this, since it's moving around prod secrets on running hosts—want to make sure the logic is bullet proof.

redshiftzero:
Will do :+1:

redshiftzero:
QA checklist is looking good from my end now. Just got the one failure on the test_custom_notification unit test on `app-staging`.

PR #1663, opened 2017-04-24
Validates username choice prior to installation

### LINKED ISSUE #1029: Playbook should fail if user creation fails
If the user specified in `ssh_user` directive fails, the playbook should fail out.

conorsch:
## Status

Ready for review

## Description of Changes

Fixes #1029.

* Strengthens the username validation prior to installing the SecureDrop application.

## Testing

The testing story here is tricky. For ages we've been using `--skip-tags validate` to omit all checks for the validation logic in the local development environment—even in prod. That choice implies the only way to test this change adequately is to check against a hardware instance—since virtualized instances will have username set to `vagrant`, which will fail these checks. I don't expect you to test against hardware while reviewing this PR. 

More substantial changes to the validation implementation, and the testing thereof, will be required in order to satisfy #749. 

## Deployment

Only affects the Admin Workstation. We'll need to test extensively prior to releasing 0.4, including on hardware, but additional changes are coming in response to #749 (and others). At present I've added a sketch of a config test to confirm that bad vars result in playbook failure, but the test is marked as "skipped" for now, since the default staging env requires vars that won't pass validation.

- [x] Testinfra tests pass on the mon-staging VM

### DIFF
--- a/install_files/ansible-base/roles/validate/defaults/main.yml
+++ b/install_files/ansible-base/roles/validate/defaults/main.yml
@@ -1,0 +1,8 @@
+---
+# List of prohibited usernames for use in production contexts.
+# The Admin should set a unique username for the instance, to be
+# used in all subsequent tasks.
+securedrop_validate_disallowed_users:
+  - amnesia
+  - root
+  - vagrant
--- a/install_files/ansible-base/roles/validate/tasks/main.yml
+++ b/install_files/ansible-base/roles/validate/tasks/main.yml
@@ -3,11 +3,7 @@
 #credentials are not used when running the app and mon playbooks for
 #production.
 
-- debug: msg="verifying ssh_users is not set to vagrant"
-  failed_when: ssh_users == "vagrant"
-  tags:
-    - validate
-    - debug
+- include: validate_users.yml
 
 - debug: msg="verifying securedrop_app_gpg_fingerprint is not the journalist test key"
   failed_when: securedrop_app_gpg_fingerprint == "65A1B5FF195B56353CC63DFFCC40EF1228271441"
--- a/install_files/ansible-base/roles/validate/tasks/validate_users.yml
+++ b/install_files/ansible-base/roles/validate/tasks/validate_users.yml
@@ -1,8 +1,22 @@
 ---
-- name: Validate Admin username.
+# Check the value of `ssh_users` in the prod-specific.yml file to ensure
+# we're using a unique username. If the playbook is running with a username
+# of e.g. `amnesia`, then the vars file has been configured incorrectly.
+- name: Validate Admin username (specified in vars).
   assert:
     that:
       - "ssh_users != item"
       - "ssh_users != ''"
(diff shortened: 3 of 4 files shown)


fowlslegs:
Are you going to add `testinfra/ansible` to `test.py` anywhere? The test you wrote should pass as-is. I don't see why we shouldn't include it now.

conorsch:
> Are you going to add `testinfra/ansible` to `test.py` anywhere?

Could, but it'll get skipped. Was going to punt till #749. 

> The test you wrote should pass as-is.

It'll get skipped if we add it, but it won't pass. For one, the `validate` role doesn't run on staging—maybe it should—and for another, `validate` when run against staging will fail loudly, given that it checks for the absence of the hard-coded staging vars.

fowlslegs:
> Could, but it'll get skipped.

I meant to imply removing that decorator.

>  For one, the validate role doesn't run on staging—maybe it should—and for another, validate when run against staging will fail loudly, given that it checks for the absence of the hard-coded staging vars.

We could also add another `target_role` to that `test.py::get_target_roles()` (whether it's `app-prod` or something else) that just runs `testinfra/ansible` too. Or we can wait and do this some other way--you should make the call here.

conorsch:
> I meant to imply removing that decorator.

Removing the `@pytest.mark.skip` decorator will cause the test to fail in _all_ environments currently declared in the Vagrantfile. The validate role fails liberally if any development-related secrets are active in the Ansible vars, so substantial reworking is necessary before we can remove the skip decorator.

> We could also add another `target_role` to that `test.py::get_target_roles()` (whether it's `app-prod` or something else) that just runs `testinfra/ansible` too.

That sounds like a viable long-term strategy; the Ansible version checks currently in `testinfra/common/test_platform.py` should be moved into that new dir, as well. Didn't move the version check as part of this PR, so the diff was manageable and the intended functionality change clear.

fowlslegs:
> Removing the @pytest.mark.skip decorator will cause the test to fail in all environments currently declared in the Vagrantfile. The validate role fails liberally if any development-related secrets are active in the Ansible vars, so substantial reworking is necessary before we can remove the skip decorator.

I get what you're saying I think. If we add the validate role to any playbook, that playbook will fail when run with Vagrant. While we kind of want that to be true, it would also make development "impossible." We could add `--skip-tags=validate` to the playbook command run in the Vagrantfile, but that would create conflicts with how this test passes options via the `ANSIBLE_ARGUMENTS` environment var, and when we want to do that via manual invocation. So merging for now then.

conorsch:
> I get what you're saying I think. If we add the validate role to any playbook, that playbook will fail when run with Vagrant. While we kind of want that to be true, it would also make development "impossible."

Bingo. Now that we have role defaults for the validate logic, we _could_ override the staging vars and permit the "vagrant" username in staging. Tested that locally, and it works as expected. Perhaps the prod env should have a non-vagrant username by default, but that'd require custom base images, and we're not ready to deliver that yet.

PR #1664, opened 2017-04-24
Use a dict instead of chain of elifs for test.py branching

fowlslegs:
## Status

Ready for review.

## Description of Changes

Just makes for cleaner, more concise code, where all the definitions are made upfront in the function.

## Testing

Just run `./test.py` with each of the keys in the dict (and keys not in the dict), and make sure it works.


conorsch:
That _is_ more readable—thanks @fowlegs!

Issue #1468, opened 2016-11-09
Consider merging the ossec repo into this one


fowlslegs:
I'm just going to quote a comment I made in a relevant PR that more or less explains the proposal I'm making and the reasoning behind it:

> @conorsch and I just talked for a minute, and I expressed concerns about the process of building the OSSEC packages via a separate repo being cumbersome. It also seems possible if not likely that a developer could forget to pull, rebuild, and re-copy-over new OSSEC packages as work happens in that repo. I proposed the possibility of merging that repo into this one. By merging the OSSEC repository into the SD repository, I see us gaining two things:

> 1. We could more fully realize the aspiration of this PR, which I see as being to automate the (re-)building and (re-)installation of local deb packages from the latest version of the source for use in the staging test VMs.

> 2. The OSSEC packages might get more love, since they'll be more visible.

---https://github.com/freedomofpress/securedrop/pull/1464#issuecomment-259299390

conorsch:
During recent testing sprints, we discovered that the modes on files and directories within `/var/ossec/` are not consistently declared. This is because the `postinst` scripts in the FPF-maintained Debian packages clobber permissions on every install. Worse, the permissions set via `postinst` are not well structured, e.g. `chmod -R 550 ${OSSEC_HOME}` needlessly setting executable bits on all files.

We don't have a tight testing loop on these settings now, since they're maintained in a [separate repository](https://github.com/freedomofpress/ossec), so I'm in favor of merging those files in here and writing config tests to validate the setup.

conorsch:
Turns out this is necessary for the new CI pipeline described in #1067, so adding to the 0.4 milestone.

Issue #951, opened 2015-03-26
avoid extra reboots


dolanjs:
Currently in the develop version, under certain situations, the playbook will cause 2 reboots in the common role. 1 during installing the grsec kernel and 1 during the cron-apt task to perform system upgrades. We should either reorganize the order of these tasks or change the reboot tasks to an ansible handler to avoid the extra reboot.

conorsch:
Responding to the original spirit of this issue, I don't much object to a dual reboot in the playbook, particularly when the order is:

1. Reboot if security updates were just installed
2. Reboot if a grsecurity-patched Linux kernel was just installed

When bootstrapping a system, it makes sense to update everything to a predictable baseline and then start specialized configuration tasks, such as setting up SecureDrop. Therefore I suggest we allow two reboots during the initial installation.

_However_, I recently noticed that in some cases there could be _three_ reboots during a fresh install, since the grsecurity role doesn't handle checking for the kernel version intelligently. That's due to a regression introduced in #1196, and is easily fixed with a second include of the `check_installation.yml` task list, to refresh the value of `running_grsec`.

conorsch:
This is not a priority to refactor, certainly not for the 0.4 release. We have sane rebooting logic as part of the external grsecurity install role (https://github.com/freedomofpress/ansible-role-grsecurity), and we'd do well to incorporate that logic into this repo for configuring SecureDrop, rather than duplicating the logic.

conorsch:
Tackling this issue. Pulled it onto the 0.4 milestone because we're seeing inconsistent reboot behavior between local Vagrant VMs and the EC2 CI setup (#1666). We need to make sure reboots work consistently and reliably in all environments, most especially hardware.

The reason why reboot behavior is working in some locations but not in others involves how Ansible handles the combination of `local_action` and `wait_for`, particularly how hostvars such as `ansible_host` are resolved (e.g. `ansible_host` becomes `localhost` in a wait_for action delegated to localhost). The safest way to reliably side-step this problem to take a two-step approach where a separate task registers the desired hostname target to be waited for in a uniquely named fact, then references that uniquely named fact in the subsequent wait_for task.

Additionally we should write a single implementation of reboot-and-wait-for, then include that where necessary to keep the config DRY and predictable.

PR #1666, opened 2017-04-25
Adds CI pipeline for testing staging environment on AWS

msheiny:
## Status

Ready for review :+1:

## Description of Changes

Fixes 
* #1067 -- AUTOMATIC testing of staging in CI utilizing CircleCI. Development will still be run under travis, I'd like to break off a new ticket to also run development in CircleCI in parallel.
* #1146 -- So I started to tackle this a bit. I'm using `ansible 2.2.1.0` during CI to deploy securedrop. This is getting used in staging and development testing in CI (in travis and staging) and passing. Had to make very minimal tweaks to get working... Was kind of surprised. There are some other legacy attributes we are using that I want to do a full cleanup on. I'll open a separate PR for that as part of the development under CircleCI effort.
* Lots of little string comparison fixes in the `testinfra` tests.

Changes
* Refactors the `build-generic-pkg` role to work against remote machines, this was necessary to get CI working properly. Previous role and playbook logic assumed that securedrop files were being mounted via virtual-machine shared storage mount.

## Testing

You should really test locally as well as test trigger the CI. If you want to test locally, you'll first off need AWS creds. You'll also need the following environment variables set:
* `BUILD_NUM` - arbitrary string, just dont conflict with anyone else on the team or MORE importantly the build number that is running in CI. 
* `FPF_CI` - set to `true`
* `FPF_GRSEC` - set to `false` to skip grsec tests
* `CI_AWS_VPC_ID`
* `CI_AWS_REGION` - currently `us-west-1`
* `CI_AWS_TYPE` -- been doing `m3.medium` so far

## Deployment

This is going to break current users playbook workflow without further fleshed out documentation. It makes some changes that are incompatible with the previously utilized version of `ansible 1.7.4` so they'll need to run a current version of ansible under a virtualenv under tails. 

2. New installs.

Need to clean up documentation and write additional code (recommend we do this in #1146 ) to bootstrap the provisioning using the more modern version of ansible than what is supplied in the tails repos.

- [x] Testinfra tests pass on the staging VM


msheiny:
Rebasing on `develop`

msheiny:
Fudge --- so just realize there are two caveats here I forgot to add:

- I realized I need to fix the problem of the test results not affecting build results. In other words, CircleCI always reports success if scripts dont bomb out with a `0` regardless of test results. CircleCI really didnt have (correct me if anyone can find it) a straight-forward way to ALWAYS run a task so I needed to shove all this under the test run which ignores exit-code results. In our case, we really need the teardown script to run everytime (even on cancel) otherwise we are going to have dangling EC2 instances / keys / sg_groups.
- this will not run properly on fork PRs - this is a security issue (re: our build infrastructure) that i will need to figure out soon. Basically my idea is to have a manual verification gate where we see a fork PR, give it a quick manual review, and then approve the build via comment key-word. I'm pretty sure we'll have to switch to Jenkins (or some other build system) to make this happen. Other idea is we have another integration branch reserved for fork PRs, basically, workflow would be you manually review, and then merge the PR into this integration branch -- once it's on an internal branch CircleCI can run on it automatically. We should hash best way forward out-of-band of this PR.

conorsch:
This is a tremendous PR! Great work, @msheiny. To aid in review, I'm inclined to slice and dice a bit. Looks like we can shave off the OSSEC repo merge (#1468) into its own PR, especially since you kindly encapsulated those changes into ca43dcd32046545918080f10de00a0ead484d2d3. Once we merge that smaller PR, then the diff here will be more manageable, and we can step through all the CI logic.

msheiny:
> To aid in review, I'm inclined to slice and dice a bit. Looks like we can shave off the OSSEC repo merge (#1468) into its own PR, especially since you kindly encapsulated those changes into ca43dcd. 

Okay Incoming --> https://github.com/freedomofpress/securedrop/pull/1668

msheiny:
Couple other things that i'll need to document better, there are a few helper commands in the `Makefile` that are involved in running CI environment. 

`make ci-go` -- this will do a spin-up, provision, and test (dumping results to junit xml files)
`make ci-teardown` --- this will teardown CI environment (this should always get run no matter what in ci)
`make ci-debug` --- if you need to debug in the CI environment (sshing to CircleCI) - you'll want to run this command early on in the build process. This will prevent the build from tearing down on failure. 

there's a few other helper commands to run each step one off -- like one for just stand-up, one for provision, one for just test.

conorsch:
>  I realized I need to fix the problem of the test results not affecting build results. In other words, CircleCI always reports success if scripts dont bomb out with a `0` regardless of test results. CircleCI really didnt have (correct me if anyone can find it) a straight-forward way to ALWAYS run a task

Sounds like a bash trap would work well here. See here for an example: https://github.com/freedomofpress/securedrop/blob/ccc2148ba087c263a99527d35c9e167ce8e178c5/snap_ci/staging_clean.sh#L6-L22 That way, all the cleanup/teardown functionality will run when the script exits, regardless of test pass or failure.

conorsch:
#1668 is undergoing final review; once that PR lands, this PR should be rebased on top of the latest develop, with the OSSEC addition commit (https://github.com/freedomofpress/securedrop/pull/1666/commits/ca43dcd32046545918080f10de00a0ead484d2d3) excised.

During rebase there may be other conflicts, but @redshiftzero and I did a lot of manual testing to try to preserve the functionality you're bringing together here. Make sure you're aware of [the `use_ssh_args` change](https://github.com/freedomofpress/securedrop/pull/1668/files#r113562990), in case that bites you.

I'll let you do the honor of rebasing, @msheiny, since you may have other commit-related cleanup to perform.

msheiny:
Refactored commits and rebased ontop of #1668 . Had to do some substantial refactoring of the generic-build logic to utilize sync/fetch. Got it to past locally, testing in CI now.. and i'll test locally with vagrant as well.

msheiny:
Sweet. All tests passing now. I'm going to leave this alone and wait for feedback.

msheiny:
@conorsch cleaned up the description for yas :+1:

msheiny:
> I realized I need to fix the problem of the test results not affecting build results. In other words, CircleCI always reports success if scripts dont bomb out with a 0 regardless of test results. 

FYI - This is fixed now thanks to the power of TRAPS!!

conorsch:
This is a fantastic PR. Eager to get in so we can utilize the automated tests as we near the finish line on the upcoming 0.4 release. A few thoughts from initial review:

* The `build` config tests are not run, but should be.
* There are some unnecessary `block`s encapsulating task lists in the CI playbooks, which should be removed.
* Some general clean up of task names to make it more obvious what the playbook is doing.
* Needs docs! Added another checkbox to #1208 to track that change, so we don't block merge here.

I'll see if I can satisfy those requested edits during review.

install_files/ansible-base/roles/common/tasks/cron_apt.yml
```diff
@@ -84,14 +84,13 @@
...
-      port={{ ansible_ssh_port }}
-      delay=20
-      state=started
+  local_action: wait_for
+  args:
+    host: "{{ ansible_host if not amazon_builder else ansible_ec2_public_hostname }}"
+    port: "{{ ansible_port | default(22) }}"
+    delay: 20
+    search_regex: OpenSSH
+    state: started
```
conorsch: Not a fan of the AWS-specific logic here; why can't we use the IPv4 address in all contexts?

install_files/ansible-base/roles/grsecurity/tasks/apply_grsec_lock.yml
```diff
@@ -12,14 +12,14 @@
...
+  # after rebooting. The Ansible variable {{ ansible_host }} refers
   # to the host just rebooted, and this syntax will work in production,
   # vagrant/virtualbox, and snap-ci environments.
 - name: Wait for server to come back.
   local_action:
     module: wait_for
-      host={{ ansible_ssh_host }}
-      port={{ ansible_ssh_port }}
+      host={{ ansible_host }}
+      port={{ ansible_port | default('22') }}
```
conorsch: Careful: while Ansible 2 adds support for `ansible_host`, it doesn't remove support for `ansible_ssh_host`; the SecureDrop production inventory file (`install_files/ansible-base/inventory`) still uses `ansible_ssh_host`, so we should continue to support it. 

See example here of how we can support both: https://github.com/freedomofpress/ansible-role-grsecurity/blob/7cdd39b3afae1d09797467e55f82ba214ae0dbe6/roles/install-grsec-kernel/tasks/validate_install.yml#L16-L23 Although I'd prefer that the default logic were reversed, so that the newer var takes precedence, i.e.

```
ansible_host|default(ansible_ssh_host|default(inventory_hostname))
```

instead of:

```
ansible_ssh_host|default(ansible_host|default(inventory_hostname))
```

conorsch:
The `build` VM config tests are failing locally, due to the reorg of the build vars. That must be fixed prior to merge. Will also activate the build config tests in CI (which are also currently failing), but haven't done so yet in this branch, since I want to confirm that the other changes I've made to cause the tests to fail.

Will need to rebase on top of develop given conflict caused by merge of #1680. During that change will tweak the reboot/wait_for logic a bit and test with multiple Ansible versions.

conorsch:
Build config working locally, including config tests. If the new logic doesn't break the CircleCI run, I'll push up another commit activating the build config tests in CI, as well. Then the last remaining change is cleaning up the reboot-and-wait conflicts with current develop.

testinfra/build/test_legacy_paths.py
```diff
@@ -0,0 +1,20 @@
...
+
+@pytest.mark.parametrize('build_path', [
+    '/tmp/build-',
+    '/tmp/rsync-filter',
+    '/tmp/src_install_files',
+    '/tmp/build-securedrop-keyring',
+    '/tmp/build-securedrop-ossec-agent',
+    '/tmp/build-securedrop-ossec-server',
+])
+def test_build_ossec_apt_dependencies(File, build_path):
```
msheiny: Nooiicee - thanks for adding this :heart:

conorsch:
Going to take this for one last spin to confirm local builds are working smoothly, then good for merge. The relevant docs have *not* been updated yet:

* Ansible version in dev docs
* Explanation of CI in dev docs

Punting those to #1208.

conorsch:
Having trouble sketching out a reliable reboot strategy that works equally well in both CI and local Vagrant boxes. The CI tests are currently passing—_all_ of them—but the `wait_for` tasks fail under Vagrant staging. 

I suggest repurposing the reboot-related #951, dragging it onto the 0.4 milestone, and focusing on reboot logic there. That way, we can get the new CI pipeline merged and use it for vetting all the upcoming tickets pre-0.4.

Thoughts, @redshiftzero?

redshiftzero:
Realistically we have been merging all kinds of stuff that has been breaking the staging environment due to lacking staging in tests so I agree that we need this in ASAP. I'm fine with merging this in and then pull in #951 or create another ticket to fix the reboot logic.

conorsch:
Thanks @redshiftzero, I heartily agree. Will provided a detailed description of the behavior based on notes.

PR #1668, opened 2017-04-25
Merge in OSSEC repo

### LINKED ISSUE #1468: Consider merging the ossec repo into this one

msheiny:
The flow of building the ossec package is tightly integrated with ossec
deployment and testing. We debated utilizing submoduling it in, but decided
it would make more sense to just bring it in and mange in this repo from here on.

## Status

Ready for review 

## Description of Changes

Fixes #1468 and sprinkles in changes from https://github.com/freedomofpress/ossec/pull/10 .

Well the big change is that the ossec repo files are brought in directly under the securedrop repo. Second, there have been some changes to assist in building and fetching back ossec package on a remote server. This was needed for CI work.

## Testing

Ehh - really needs to be tested as part of https://github.com/freedomofpress/securedrop/pull/1666 but I guess you could confirm local vagrant up is still working.Though honestly I havent messed with that part yet :|

## Deployment

> Should have no affect
2. New installs.
> Should have no affect


conorsch:
Cleaned up some of the OSSEC build logic in https://github.com/freedomofpress/ossec/pull/11, so I'll pull that in here. Will also shove the included roles into the existing `install_files/ansible-base/roles/` directory, and update the playbook so that the build VM will automatically builds these packages, as well.

conorsch:
Pulled in the [external repo](https://github.com/freedomofpress/ossec) entire, including commit history, via `git subtree`. Integrated the OSSEC build logic with the existing `build` VM, so now that VM builds _all_ SecureDrop packages. 

Simplified the logic for fetching back the built deb packages: now the role _always_ uses the `fetch` module regardless of the backend provider. That way, the build logic works the same whether you're building locally in VMs, or on remote hosts in CI for staging.

install_files/ansible-base/roles/build-ossec-deb-pkg/tasks/main.yml
```diff
@@ -0,0 +1,138 @@
...
+- name: Conditional copy over current directory for repo build
+  synchronize:
+    # All Debian package source trees are located in `install_files/<package>`
+    # in the root of the SecureDrop git repository.
+    src: ../../../../ossec-{{ purpose }}/
+    dest: "{{ repo_src_path }}"
+    # Ansible v1 does not support the `use_ssh_args`, which was added in 2.0.
+    # When resolving #1146, to upgrade Ansible, this option should be
+    # activated to permit builds in CI.
+    #   use_ssh_args: yes
```
conorsch: @msheiny Note this change. The role is now compatible with Ansible v1.8.4, but for #1666 you may need to edit this option. Made sure to preserve Ansible v1 support since #1146 is still open.

conorsch:
aba3ef635cb76db8ac44b1dc31779f1611a93d43 includes docs updates toward satisfying #1208.

redshiftzero:
Hmm, I got a failure while provisioning the build VM on `TASK: [build-ossec-deb-pkg | Conditional copy over current directory for repo build]`. Error messages:

```
 *** 
msg: Warning: Permanently added '[127.0.0.1]:2222' (ECDSA) to the list of known hosts.
rsync: failed to set times on "/tmp/build/ossec-server/.": Operation not permitted (1)
rsync: recv_generator: mkdir "/tmp/build/ossec-server/DEBIAN" failed: Permission denied (13)
*** Skipping any contents from this failed directory ***
rsync: recv_generator: mkdir "/tmp/build/ossec-server/etc" failed: Permission denied (13)
*** Skipping any contents from this failed directory ***
rsync: recv_generator: mkdir "/tmp/build/ossec-server/usr" failed: Permission denied (13)
*** Skipping any contents from this failed directory ***
rsync error: some files could not be transferred (code 23) at /BuildRoot/Library/Caches/com.apple.xbs/Sources/rsync/rsync-47/rsync/main.c(992) [sender=2.6.9]
```

conorsch:
Drat, I'm guessing you have an older version of rsync than I do—I'm using `3.1.2`. You're probably 2.x with substantially different options support on macOS. Not sure how to craft a BSD-rsync-compatible line that won't break on Linux hosts; it might be as simple as adding:

```
  rsync_opts:
    - "--rsync-path 'sudo rsync'"
```

to the `synchronize` task. Try upgrading Ansible to a more recent release (i.e. `pip install ansible==2.2.1`) and see if the synchronize module got smart enough to handle this for us.

redshiftzero:
Yep I am using rsync version 2.6.9. Your suggestion of using Ansible 2.2.1 worked - I was able to successfully provision a build VM. :metal:

redshiftzero:
Ah good, your rsync option also worked. So that's probably a good solution for now in order to avoid what the pros call "mission creep".

conorsch:
Good to know both options (upgrading Ansible and explicitly setting `--rsync-path`) work for you. Since #1146 is still open, and requires a lot of ancillary changes, I'm reticent to upgrade Ansible outside the context of #1146. So, what I'd prefer to do is use the `rsync_opts` addition, as long as it works well under both versions of Ansible.

Confirmed the `--rsync-path` approach worked for me, on Linux, with rsync v3.1.2 and Ansible v1.8.4. Trying again with Ansible v2.2.1....

redshiftzero:
Confirming that Ansible 2.2.1 and the `--rsync-path` option also works.

conorsch:
Confirmed the `--rsync-path` approach works for me on Linux with rsync v3.1.2 and Ansible v2.2.1 _and v2.3.0 (since it never hurts to be curious). @redshiftzero if you're good on your end, I'll add a commit implementing that approach, since it seems the safest bet not to breaking #1666.

Issue #1530, opened 2017-01-25
Support for 16.04 Ubuntu Release


b-meson:
Current SD Administrators are getting the following OSSEC alarm:

```
Subject:  Cron <root@mon> test -x /usr/sbin/anacron || ( cd / && run-parts --report /etc/cron.weekly )
Body:
/etc/cron.weekly/update-notifier-common:
New release '16.04.1 LTS' available.
Run 'do-release-upgrade' to upgrade to it.
```

If you're an administrator who is diligently checking your OSSEC alarms, you might be tempted to upgrade which will break the application server and require a SD reinstall. Additionally, there is another set of alarms that trigger every two minutes 

```
OSSEC HIDS Notification.
2017 Jan 24 

Received From: mon->/var/log/syslog
Rule: 1002 fired (level 2) -> "Unknown problem somewhere in the system."
Portion of the log(s):

Jan 24 23:47:29 mon ntpd[908]: error resolving pool 3.ubuntu.pool.ntp.org: Temporary failure in name resolution (-3)

 --END OF NOTIFICATION
```

freddymartinez9:
Someone with admin access in the SD Support Portal should send a message to all the other admins to ignore this `do-release-upgrade` alarm.

conorsch:
Ouch, sorry to hear it, @freddymartinez9. Thanks for reporting. We've had a number of Admins reach out to us via the support portal and inquire about the schedule of moving to 16.04. For now we're definitely recommending that Admins **not** upgrade to 16.04 Xenial, due to the problems you've described. We'll prioritize testing there and see if we can't update the config to support it in the near future.

freddymartinez9:
@conorsch ah okay that makes sense. I will try to send an OSSEC PR to suppress the ntpd alarms and double check that `ntpd` service is functioning properly on 16.40 as well.

r4v5:
The ntp alarm is because of unrelated DNS problems `pool.ntp.org` is having. That said, it probably doesn't need to fire every two minutes.

freddymartinez9:
had a discussion with @r4v5 and it appears that its the actual DNS resolvers having issues. The alarm is triggering every few minutes. We have thought about making it trigger once every six hours in case the pool is down. I'll try to implement this logic in the PR.

redshiftzero:
For the ugprade here we will need to verify that the Apache access control directives that we use are not deprecated in the version of Apache used in 16.04 (see prior issue #1607).

[...]

PR #1669, opened 2017-04-25
Timing and token reuse

fowlslegs:
## Status

Ready for review.

## Description of Changes

Eliminates some timing leaks and prevents the same valid token from being used to authenticate twice in a row.

### DIFF
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 from base64 import b32encode
+import hmac
 import os
 import subprocess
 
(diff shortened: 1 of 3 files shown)


redshiftzero:
Hmm, builds seem to be reliably failing on Travis on this - you didn't see any test failures locally?

conorsch:
Also seeing 22 failures locally, same as in Travis.

conorsch:
`hmac.compare_digest()` was [added in Python 2.7.7](https://docs.python.org/2/library/hmac.html), but Ubuntu Trusty includes only Python 2.7.6. So these tests will never pass with the current implementation, nor did they prior to submission.

conorsch:
We cannot trivially upgrade the version of Python running on SecureDrop instances, so we'll need to consider another path. Worth nothing that #1530 considers upgrading the Trusty hosts to Xenial, which would get us newer versions.

fowlslegs:
Re-posting from Slack:

> Too much multitasking, sorry.
>
>Tests were running and I preemptively checked them, then started on something else and submitted the PR w/o thinking.

Also, re-opening because I just replaced `hmac.compare_digest` with another constant-time comparison function from Django.

fowlslegs:

PR #1671, opened 2017-04-26
Remove journalist assignment

redshiftzero:
## Status

Ready for review

## Description of Changes

Unfortunately, since database migrations support (#1419) did not make it into the feature set for 0.4, we can't ship the journalist assignment feature introduced in #1491 to prod instances. This PR removes this feature. 

I've added to the checklist in #1658 so that we can bring it in later in the 0.4.x release series when database migrations is implemented.

## Testing

How should the reviewer test this PR?

1. Provision development VM
2. Run unit/functional tests
3. Submit a document to the source interface
4. Make admin account, sign into journalist interface
5. Verify journalist assignment box is gone

## Screenshot

Journalist assignment field is gone:

screen shot 2017-04-26 at 8 31 32 am

### If you made changes to the app code:

- [x] Unit and functional tests pass on the development VM

Issue #1658, opened 2017-04-21
Backed out changes for the release of SecureDrop 0.4


redshiftzero:
This is a tracking issue for changes backed out for the upcoming release of SecureDrop 0.4. We want to release 0.4 as soon as possible in order to resolve the large diff between `master` and `develop` (issue #1559). To ease this process both in terms of deployment and minimizing disruptive changes for users, we are going to back out some changes in `develop`. These changes will be reconsidered for release in the 0.5.x release series:

- [ ] Display of journalist username on source interface (UX improvement). Added in PR #1534, removed in PR #1657. See also issue #1592.
- [ ] Assigning journalists to sources. Added in #1491, removed in PR #1671. Requires implementation of #1419. 

Items will be added to this list.

dachary:
@redshiftzero is it still useful ?

dachary:
@redshiftzero or not ?

redshiftzero:
Ah so, I think this is still useful as we do want these changes in (and these items are not otherwise tracked anywhere). I just edited the issue text to say "0.5.x release series" though 😇

heartsucker:
Closing this in favor of discussions in #1355 and #1592 since those are more specific to the actual changes and this is a bit of combined issue for those two.

Issue #1189, opened 2015-11-08
Handle Submissions with no Sources in currently running instances.


pwplus:
We need to handle submissions database entries in currently running instances that have no Source. This is a follow up issue to https://github.com/freedomofpress/securedrop/issues/1188.

pwplus:
@garrettr Can you add this as a milestone? Thanks!

fowlslegs:
X-link: https://github.com/freedomofpress/securedrop/issues/1419.

redshiftzero:
note: the correct place to address this ticket now is in an alembic revision (with no `downgrade` implemented as this is a destructive action)

eloquence:
We have privately received a patch for this that we'll consider, therefore moving to near-term backlog for inclusion in a future sprint.

PR #1673, opened 2017-04-26
Back out deletion of submissions when a source is deleted

redshiftzero:
## Status

Ready for review

## Description of Changes

This was a addition in #1440, but similarly to #1671 until we have a migration path
for making changes like this to the database we cannot ship this yet
to instances (it adds an ON DELETE CASCADE constraint). 

Pushed to 0.4.x release series.

Added to checklist in #1658. 

## Testing

1. Provision dev VM.
2. Make sure dev tests pass locally.

### DIFF
--- a/securedrop/db.py
+++ b/securedrop/db.py
@@ -132,7 +132,7 @@ class Submission(Base):
     source_id = Column(Integer, ForeignKey('sources.id'))
     source = relationship(
         "Source",
-        backref=backref("submissions", order_by=id, cascade="delete")
+        backref=backref("submissions", order_by=id)
         )
 
     filename = Column(String(255), nullable=False)
@@ -162,7 +162,7 @@ class Reply(Base):
     source_id = Column(Integer, ForeignKey('sources.id'))
     source = relationship(
         "Source",
-        backref=backref("replies", order_by=id, cascade="delete")
+        backref=backref("replies", order_by=id)
         )
 
     filename = Column(String(255), nullable=False)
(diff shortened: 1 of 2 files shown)


redshiftzero:
Hmm, so I just read about how SQLAlchemy does this a little more, and it looks like this might actually be done in the ORM and so okay to leave in. From the [docs](http://docs.sqlalchemy.org/en/latest/orm/cascades.html): 

> The behavior of SQLAlchemy’s “delete” cascade has a lot of overlap with the ON DELETE CASCADE feature of a database foreign key, as well as with that of the ON DELETE SET NULL foreign key setting when “delete” cascade is not specified. Database level “ON DELETE” cascades are specific to the “FOREIGN KEY” construct of the relational database; SQLAlchemy allows configuration of these schema-level constructs at the DDL level using options on ForeignKeyConstraint which are described at ON UPDATE and ON DELETE.

But for safety, need to test the upgrade to check

redshiftzero:
Yeah, in further digging around on how this is done in SQLAlchemy, this is indeed ORM only. In order to set ON DELETE CASCADE one has to explicitly set the [ForeignKeyConstraint](http://docs.sqlalchemy.org/en/latest/core/constraints.html#sqlalchemy.schema.ForeignKeyConstraint). So! We can leave this in and I'm closing this PR.

conorsch:
Manual testing of the upgrade strategy here went well, so agreed we can close. Even when this feature lands, it'll only help for _future_ delete actions. We should consider cleaning up dangling database references in a future migration.

redshiftzero:
Yep, totally. We have the cleanup still to do in issue #1189.

Issue #889, opened 2015-02-18
potential timing attacks on Journalist.login and user.valid_password?


ThomasWaldmann:
it first checks for the user being valid (existing), raises exception if not.
then verifies token, raises exception if not.
then verifies password, raises exception if not.

an attacker can see the timing (maybe even the kind of exception / error msg) and make educated guesses about what was wrong.

user.valid_password does not use a constant time string comparison internally. see pyotp.utils.strings_equal for a better way.

ageis:
Interesting, would just note that the attacker would have to compromise the journalits's Tor HidServAuth value first to try this.

garrettr:
@ageis is right. In addition:

Theoretically, an attacker could use timing attacks based on the current structure of `Journalist.login` to learn the following (in order):

1. Username
2. Token
3. Password

I do not think learning the username is very useful for an adversary, and mitigating that is probably not worth the complexity. Note that we intentionally check the two-factor token *before* the password - this is to prevent password brute force attempts for attackers who don't have the two-factor token.

I think the throttling (implemented in `verify_token` and `throttle_login`) makes it very difficult to exploit anything like this. The TOTP token comparison is constant time, and even if it wasn't you're only allowed to make up to 5 login attempts / minute so you don't get a lot of guesses for the 6-digit value.

Finally, I don't think `valid_password` using a non-constant time string comparison matters because it is comparing the hashed passphrases. Unless you have a preimage attack on the hash function (scrypt), I can't see how this would help you guess the passphrase.

So, at the moment I don't think any of the valid issues raised lead to exploitable security bugs. Please let me know if you think I have overlooked anything!

fowlslegs:
All the points Garrett makes are good. I'd just like to reinforce one point. For a cryptographically secure hash function, the [avalanche effect](https://en.wikipedia.org/wiki/Avalanche_effect) will be very strong. Therefore, the scrypt hash of `passworc` is going to be totally different from that of `password`, and the way in which it differs will be unpredictable for _any_ PPT adversary. Therefore, even if we didn't have all these other mitigations to even trying out an admin/journalist password, it makes a negligible (in the cryptographic sense) difference whether we use a constant-time string comparison function here or not.

PR #1676, opened 2017-04-26
Removing timing information from login

fowlslegs:
## Status

Ready for review!

## Description of Changes

Removes non-exploitable timing information from `valid_password()`, and negligibly useful--but hypothetically exploitable in a highly specific context--timing information from `verify_token()`. See commit messages for more detailed information.

### DIFF
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 from base64 import b32encode
+import hmac
 import os
 import subprocess
 
@@ -191,6 +192,24 @@ def decrypt(secret, ciphertext):
     """
     hashed_codename = hash_codename(secret, salt=SCRYPT_GPG_PEPPER)
     return gpg.decrypt(ciphertext, passphrase=hashed_codename).data
+
+
+# Taken from Django Source Code (with some modification)
+def constant_time_compare(val1, val2):
+    """Returns True if the two strings are equal, False otherwise. Tries
+    to use :meth:`hmac.compare_digest()`, but falls back to an
+    implementation from Django if that's not available.
+    """
+    try:
+        hmac.compare_digest(val1, val2)
+    except AttributeError:
+        if len(val1) != len(val2):
+            return False
+        result = 0
+        for x, y in zip(val1, val2):
+            result |= ord(x) ^ ord(y)
+        return result == 0
+
 
 if __name__ == "__main__":
     import doctest
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -2,7 +2,6 @@
 # -*- coding: utf-8 -*-
 
 from base64 import b32encode
-import hmac
 import os
 import subprocess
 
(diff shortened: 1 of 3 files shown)


fowlslegs:
Rebased & edited first commit message to credit @EdOverflow.

fowlslegs:

fowlslegs:
Oh, I already found and fixed that, but apparently in another branch. The problem is that `compare_digest` iterates over its inputs and the initial value of `last_token` is `None`, which of course does not implement the iterator interface. The simple fix is not doing the comparison the first time a user logs in.

redshiftzero:
Looks like you have a typo in the name of the `pyotp` dependency:

```
--------------------------------------------------------------------------------
ERROR in journalist [journalist.py:114]:
Login for 'test' failed: global name 'pytop' is not defined
--------------------------------------------------------------------------------
```

fowlslegs:
Going to write another test then because that means that code path is not being tested.

fowlslegs:
Okay, looks ready to me @redshiftzero !

fowlslegs:
Rebased.

fowlslegs:
Rebased.

redshiftzero:
This is a writeup from the SecureDrop Engineering Team as followup for transparency since we got an inquiry why we awarded points to a Bugcrowd researcher. To clarify: this is not an exploitable security vulnerability. It cannot be used to compromise codenames, sources, submissions, communications, or private keys. 

## Issue

The `valid_password()` and `verify_token()` functions in `Journalist.login` in `db.py` were not using constant time comparisons, which was reported to us on Bugcrowd as a potential timing attack.

## Why this isn’t exploitable

The non-constant-time comparison in `valid_password()` did not compare passwords directly, it was comparing salted scrypt password hashes. Provided scrypt is a secure hash function, this cannot be exploited to leak any information about the underlying password due to scrypt's preimage resistance.

The non-constant-time comparison in the `verify_token()` function did leak timing information, but it could not be exploited due to the throttling of login attempts and token reuse protection. 

It’s worth noting how SecureDrop's architecture provides defense in depth for both of these attacks: for an attacker to get to this point, they would first need to comprise the authentication provided by the Authenticated Tor Hidden Service used for the Journalist Interface.

## Why we awarded Bugcrowd “points” to the researcher who reported these issues

While this isn’t exploitable, this is probably the best submission we have received thus far through Bugcrowd, and we decided to award a few points to thank the researcher for their close inspection of the SecureDrop code—and to hopefully motivate more code review and clear reports. We also agreed that, while not strictly necessary in these specific cases, using constant time comparisons when performing comparisons against secret data is generally best practice for writing secure code. 

## Why we made the code change in this PR

We also decided to go ahead and change all these comparisons to be constant time in 8cc163d31af77a86961754535ce98c060bbe04ed, for 1. defense in depth, 2. to make sure we are using best practices for writing secure code in SecureDrop, and 3. as we had received and had to handle reports of this issue several times (both on Bugcrowd and GitHub (e.g. see #889 for this issue being reported again)).

PR #1677, opened 2017-04-27
Strengthen totp reuse protections

fowlslegs:
## Status

This PR depends on #1676 and  #1676 is merged. That said, it is ready for review, but you might just want to wait for the rebase to happen so the diff is more clear.

## Description of Changes

Strengthens TOTP protections:

* Protection is only applied to TOTP users because I believe the process of updating the HOTP counter is sufficient to defend against such attacks.
* Once a TOTP token has been used as part of a valid login process, that token, as well as the previous and following tokens are rejected. This is to prevent MitM and shoulder-surfing type attacks. Since in practice it seems unlikely a journalist would login and logout just to login again within a ~60s window, this shouldn't be of inconvenience.
* Since we're using `last_access` in place of `last_token`, I (i) added a "`# deprecated`" comment to the relevant column to remind us to remove it when we do our next database migration and (ii) moved the updating of last access from the journalist app code to the database code, where it seems to fit better, and helps the reader understand the protection.
* Protection is applied in the `login()` function instead of the `verify_token()` function because `verify_token()` is called from other functions, and with the re-implementation this could create problems in the case a user logs in and then tries to re-generate a TOTP token.
* Since the class of and conditions under which exceptions are raised slightly changed in `db.Journalist.login()`, I updated `journalist.login()` to reflect that.
* Adds Google-style docstrings.

## Testing

Since `LOGIN_HARDENING` is set to `False` during the tests, this will require manual review. With both the `development` and `staging` machines, create a `Journalist`, and try logging in, out, and back in again. You should be flashed a warning on your second login attempt notifying you to wait 60 seconds. Wait that long and then try again. Login should succeed.

We could create tests which temporary set `LOGIN_HARDENING` to `True`, to make sure a token can't be re-used. I think by mocking `datetime` we could do thorough testing without making such a test take forever. Thoughts?

### DIFF
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 from base64 import b32encode
+import hmac
 import os
 import subprocess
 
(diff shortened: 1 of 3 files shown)


fowlslegs:
Setting to blocked for now because this is really blocking on another PR, but also, I have some thoughts on simplifying the protection, and may want to add some tests as well once I come to a conclusion on how best to do so.

fowlslegs:
I think that this implementation of TOTP hardening (as well as the original) are may be bypassed by a race condition, so expect a follow-up commit that uses some synchronization primitive to avoid this. Leaving open as I work for feedback.

fowlslegs:
Just had a talk with @garrettr about our threat model here, and reviewed the https://github.com/google/google-authenticator-libpam. This is going to go under one more revision that should provide the same protections as the current implementation, but will be less of an annoyance to the user in certain (mostly training-related) UX edge cases we discussed, and will work more similarly to the google-authenticator-libpam implementation of token re-use protection.

PR #1678, opened 2017-04-27
Adds manage.py aliases and updates tmp dir cleaning cron job

### LINKED ISSUE #1675: Updated `manage.py` subcommands should have aliases
In #1347 we converted the subcommands for the `manage.py` script to use hyphens instead of underscores, e.g. `add_admin` -> `add-admin`. That works well for the most part, assuming existing Admins read the docs on the update, but it will cause problems for currently running instances, because a cron job exists for the `clean_tmp` subcommand, and that subcommand will no longer exist in 0.4. We should do two things to resolve:

1. Implement subcommand aliases in `manage.py` so that the old subcommands still work. 
2. Update the Ansible config to ensure state=absent on the old cronjob, and state=present on the new one. Doing so will ensure that Admins running the playbooks do not create duplicate cron jobs.

If we don't do this, the nightly cron job for cleaning the temp directory will fail on all Application Servers going forward. Therefore this'll need to go on the 0.4 milestone, lest we break the temp cleanup logic currently used in production.

fowlslegs:
## Status

Ready for review.

## Description of Changes

Fixes #1675.

## Testing

Reviewers can test `manage.py` in any VM. To test the playbook changes, provision a staging VM using `master`, then switch to this branch and re-provision. You should see the old cronjob commented out and the new one present.

### DIFF
--- a/install_files/ansible-base/roles/app/tasks/setup_cron.yml
+++ b/install_files/ansible-base/roles/app/tasks/setup_cron.yml
@@ -1,7 +1,16 @@
 ---
-- name: Add cron job to clean SecureDrop tmp dir daily.
+- name: Remove cron job to clean SecureDrop tmp dir daily (old manage.py syntax).
   cron:
     name: cleanup SecureDrop temp dir
+    job: "{{ securedrop_code }}/manage.py clean_tmp"
+    special_time: daily
+    state: absent
+  tags:
+    - cron
+
+- name: Add cron job to clean SecureDrop tmp dir daily (new manage.py syntax).
+  cron:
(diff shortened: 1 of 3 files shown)


conorsch:
Reviewing now! Changes look great, just want to test the cron job updates to be absolutely sure. Will spin up 0.3.12 machines locally then flip to this PR branch and confirm the that the old cron job is removed and replaced with the new one.

conorsch:
Works like a charm, and changes are idempotent. Updated the config tests to track both state=present on the new form, and state=absent on the old form. :+1:

PR #1679, opened 2017-04-28
Adds error handling to OSSEC encrypted email script

### LINKED ISSUE #858: When ossec_gpg_fpr is invalid, OSSEC sends empty emails
**STR**
1. Fill out the OSSEC information in `staging-specific.yml`, following the instructions in `docs/ossec_alerts.md`.
2. Mess up the value of `ossec_gpg_fpr`. In my case, I accidentally deleted a character of the fingerprint while manually removing whitespace (which is present by default in the output of `gpg --with-fingerprint`).
3. `vagrant up /staging$/
4. Note that the OSSEC emails you receive contain correct-seeming subject lines, but the emails themselves are empty.

At a minimum, this should trigger errors on the installed system. From a cursory examination, I did not see any errors from encrypting the emails in `/var/ossec/logs/ossec.log` or `/var/log/mail.log`. Even better, we should prevent the install from happening if this kind of error is apparent in the configuration.

conorsch:
## Status

Ready for review

## Description of Changes

Fixes #858.

* Validates the GPG fingerprint vars in `prod-specific.yml` before attempting to use the values in templates. This is designed to prevent known-bad values from landing on the servers.
* Makes the encrypted email notification script more resilient to failure, sending a simple plaintext email if encryption failed. This is designed to inform the Admin about bad values that somehow slipped into the Monitor Server.
* Updates documentation with explicit reference to the new OSSEC email error message.

A few notes on the implementation here. The validation logic via `assert` is rather straightforward, and chips away at the long list of concerns documented in #749. There are substantial changes to the `send_encrypted_alarm.sh` script, however. I _really_ didn't want to make major edits to such a simple script, but it was necessary to avoid the situation of encryption errors meaning that empty emails are sent. 

The problem with the old implementation is that the pipeline (i.e. chain of commands all joined by `|`) approach _always_ sends mail regardless of what is passed through to the `/usr/bin/mail` program. Mail _does_ have a convenient option `-E` that silently discards empty messages, which would have prevented the empty emails. Simply avoiding sending empty messages isn't enough, though: we want to inform the Admin that something is awry, to encourage investigation and resolution.

So we're _not_ using the `-E` option, instead we'll build the message body, encrypt it, then ensure both that the body is not an empty string, and that the pipeline exited non-zero. If those conditions are satisfied, we send the encrypted message as normal. If either of those conditions are not satisfied, then we fail over to sending the general "you should check it out" message.

## Testing

1. Edit `install_files/ansible-base/group_vars/staging.yml` and add working GPG key info and OSSEC email info. You'll need working OSSEC email delivery in order to test this PR! (Don't forget about the new `ossec_from_address` var, added in #1565; you'll probably need to set that to the "from" address.)
2. Provision staging servers.
3. Confirm receipt of encrypted OSSEC alerts via email, and that you can decrypt them.
4. SSH into `mon-staging` and edit `/var/ossec/send_encrypted_alarm.sh`; remove a single character from the fingerprint, to make it invalid (GPG will fail).
5. Confirm that next OSSEC alerts are plaintext and contain the message `Failed to encrypt OSSEC alert. Investigate the mailing configuration on the Monitor Server.`

Tip: `service ossec restart` is a convenient way to trigger an alert to fire; works on both App and Mon machines.

## Deployment

Since these changes only apply to the Ansible config, existing instances will not get the benefit of the additional error handling automatically—they'll need to run the playbook again. That's OK by me, and I don't think we need to require Admins to run the playbook to get this specific update, since it's mostly useful for first-time Admins performing the initial installation—that's when the OSSEC configuration is most painful. Hopefully the docs update will alleviate some of that pain.

### DIFF
--- a/install_files/ansible-base/roles/ossec-server/templates/send_encrypted_alarm.sh
+++ b/install_files/ansible-base/roles/ossec-server/templates/send_encrypted_alarm.sh
@@ -1,5 +1,49 @@
 #!/bin/bash
+# Handler script to encrypt OSSEC alerts prior to mailing.
+# Called via the `.procmailrc` for user `ossec`.
 
-/usr/bin/formail -I "" | \
-    /usr/bin/gpg --homedir /var/ossec/.gnupg --trust-model always -ear "{{ ossec_gpg_fpr }}" | \
-    /usr/bin/mail -s "$(echo $SUBJECT | sed -r 's/([0-9]{1,3}\.){3}[0-9]{1,3}\s?//g' )" {{ ossec_alert_email }}
+# Intentionally NOT setting `set -e`, since we'll do our own error handling.
+# Rather than exit immediately on errors, we should gracefully inform
+# the Admin that there was a problem encrypting a message.
+#set -e
+set -o pipefail
+
+
+# Read alert message content from STDIN, as passed by OSSEC. Store the message
+# in a var so we can iteratively process it, halting execution before sending
+# if the GPG encryption command failed. If we simply chain a bunch of pipes
+# together, then the email will be sent with an empty body, should the GPG
+# encryption fail.
+ossec_alert_text="$(< /dev/stdin)"
+
+# Primary "send email to Admin" functionality.
+function send_encrypted_alert() {
+
+    local encrypted_alert_text
+    # Try to encrypt the alert message. We'll inspect the exit status of the
+    # pipeline to decide whether to send the alert text, or the default
+    # failure message.
+    encrypted_alert_text="$(printf "${ossec_alert_text}" | \
+        /usr/bin/formail -I '' | \
+        /usr/bin/gpg --homedir /var/ossec/.gnupg --trust-model always -ear '{{ ossec_gpg_fpr }}')"
+
+    # Error handling.
+    if [[ -z "${encrypted_alert_text}" || $? -ne 0 ]]; then
+        send_plaintext_fail_message
+    else
+        echo "${encrypted_alert_text}" | \
+            /usr/bin/mail -s "$(echo $SUBJECT | sed -r 's/([0-9]{1,3}\.){3}[0-9]{1,3}\s?//g' )" '{{ ossec_alert_email }}'
+    fi
+}
+
+# Failover alerting function, in case the primary function failed.
+# Usually a failure is related to GPG balking on the encryption step;
+# that may be due to a missing pubkey or something reason.
+function send_plaintext_fail_message() {
+    printf "Failed to encrypt OSSEC alert. Investigate the mailing configuration on the Monitor Server." | \
+        /usr/bin/formail -I "" | \
+        /usr/bin/mail -s "$(echo $SUBJECT | sed -r 's/([0-9]{1,3}\.){3}[0-9]{1,3}\s?//g' )" '{{ ossec_alert_email }}'
+}
+
+# Encrypt the OSSEC notification and pass to mailer for sending.
+send_encrypted_alert
--- a/docs/ossec_alerts.rst
+++ b/docs/ossec_alerts.rst
@@ -326,6 +326,18 @@ after the first playbook run, try setting ``ossec_from_address`` in
 ``prod-specific.yml`` to the full email address used for sending the alerts,
 then run the playbook again.
 
+Message failed to encrypt
+~~~~~~~~~~~~~~~~~~~~~~~~~
+If OSSEC cannot encrypt the alert to the GPG public key for the Admin
+email address (configured as ``ossec_alert_email`` in ``prod-specific.yml``),
+the system will send a static message instead of the scheduled alert:
+
+  Failed to encrypt OSSEC alert. Investigate the mailing configuration on the Monitor Server.
+
+Check the GPG configuration vars in ``prod-specific.yml``. In particular,
+make sure the GPG fingerprint matches that of the public key file you
+exported.
+
 Troubleshooting SMTP TLS
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/install_files/ansible-base/roles/validate/tasks/main.yml
+++ b/install_files/ansible-base/roles/validate/tasks/main.yml
@@ -5,29 +5,10 @@
 
 - include: validate_users.yml
 
-- debug: msg="verifying securedrop_app_gpg_fingerprint is not the journalist test key"
-  failed_when: securedrop_app_gpg_fingerprint == "65A1B5FF195B56353CC63DFFCC40EF1228271441"
-  tags:
-    - validate
-    - debug
-
-- debug: msg="verifying securedrop_app_gpg_fingerprint is not the admin test key"
-  failed_when: securedrop_app_gpg_fingerprint == "600BC6D5142C68F35DDBCEA87B597104EDDDC102"
-  tags:
-    - validate
-    - debug
-
-- debug: msg="verifying ossec_gpg_fpr is not the journalist test key"
-  failed_when: securedrop_app_gpg_fingerprint == "65A1B5FF195B56353CC63DFFCC40EF1228271441"
-  tags:
-    - validate
-    - debug
-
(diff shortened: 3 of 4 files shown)


conorsch:
Supersedes #1259. The one neat thing that #1259 did was attempt to verify that the declared fingerprint matched the associated pubkey file, via a `local_action` script. That functionality is not included in this PR, but I think the additional error handling should cover us.

install_files/ansible-base/roles/ossec-server/templates/send_encrypted_alarm.sh
```diff
@@ -1,5 +1,49 @@
...
+    fi
+}
+
+# Failover alerting function, in case the primary function failed.
+# Usually a failure is related to GPG balking on the encryption step;
+# that may be due to a missing pubkey or something reason.
+function send_plaintext_fail_message() {
+    printf "Failed to encrypt OSSEC alert. Investigate the mailing configuration on the Monitor Server." | \
+        /usr/bin/formail -I "" | \
+        /usr/bin/mail -s "$(echo $SUBJECT | sed -r 's/([0-9]{1,3}\.){3}[0-9]{1,3}\s?//g' )" '{{ ossec_alert_email }}'
```
redshiftzero: It would be useful to replace the subject with something like "OSSEC Notification - Mailing misconfiguration" or alternatively append " - Mailing Error" to the subject, such that in a storm of OSSEC alerts (not an unlikely scenario) the administrator can quickly see that encrypted alerts are not being sent properly.

redshiftzero:
I've verified following the testing instructions that the functionality in `send_encrypted_alarm.sh` is working like a charm :100:

install_files/ansible-base/roles/ossec-server/templates/send_encrypted_alarm.sh
conorsch: Great idea, I'll try to add that. Should be as simple as overriding the `SUBJECT` env var.

redshiftzero:
So it looks like `include + with_items` is not supported in Ansible 1.8.4, is the `validate` role intended to be for Ansible 2.2.1 only?

conorsch:
> So it looks like `include + with_items` is not supported in Ansible 1.8.4, is the `validate` role intended to be for Ansible 2.2.1 only? 

That was sloppy of me; I've been hopping back and forth between v1 and v2 in virtualenvs during review of #1666. I'm inclined to say "upgrade to 2.2.2 locally" and we'll hammer out the transition snags in #1146.

redshiftzero:
Gave the validate role a spin on staging to test some intentionally incorrect GPG configurations and the validate role catches the misconfigurations as expected:

```
TASK [validate : Validate GPG info.] *******************************************
fatal: [app-staging]: FAILED! => {
    "assertion": "' ' not in item", 
    "changed": false, 
    "evaluated_to": false, 
    "failed": true
}
```

```
TASK [validate : Validate GPG info.] *******************************************
fatal: [app-staging]: FAILED! => {
    "assertion": "item|length == 40", 
    "changed": false, 
    "evaluated_to": false, 
    "failed": true
}
```

```
TASK [validate : Validate GPG info.] *******************************************
fatal: [app-staging]: FAILED! => {
    "assertion": "item == item|regex_replace('[^A-F0-9]', '')", 
    "changed": false, 
    "evaluated_to": false, 
    "failed": true
}
```

install_files/ansible-base/roles/ossec-server/templates/send_encrypted_alarm.sh
```diff
@@ -1,5 +1,49 @@
...
+# Primary "send email to Admin" functionality.
+function send_encrypted_alert() {
+
+    local encrypted_alert_text
+    # Try to encrypt the alert message. We'll inspect the exit status of the
+    # pipeline to decide whether to send the alert text, or the default
+    # failure message.
+    encrypted_alert_text="$(printf "${ossec_alert_text}" | \
+        /usr/bin/formail -I '' | \
+        /usr/bin/gpg --homedir /var/ossec/.gnupg --trust-model always -ear '{{ ossec_gpg_fpr }}')"
```
fowlslegs: I think you should add `--no-batch` here as it is "highly recommended" in the GPG manual. `--with-colons`, also recommended for unattended mode is irrelevant here, but you could specify the last recommended option for unattended use by adding the flag `--status-fd 2` and then putting `2> /dev/null` after the GPG command. Also, to make it explicit the input is going to come from stdin you'd end the GPG args with `-`. Of course this works just the way it is, but my preference is to be explicit and follow all recommendations.

>        --no-batch
              Use batch mode.  Never ask, do not allow  interactive  commands.   --no-batch  disables
              this option.  Note that even with a filename given on the command line, gpg might still
              need to read from STDIN (in particular if gpg figures that the input is a detached sig‐
              nature  and no data file has been specified).  Thus if you do not want to feed data via
              STDIN, you should connect STDIN to g‘/dev/null’.

              It is highly recommended to use this option along  with  the  options  --status-fd  and
              --with-colons for any unattended of gpg.

>        --status-fd n
              Write  special  status  strings  to the file descriptor n.  See the file DETAILS in the
              documentation for a listing of them.


>       --with-colons
              Print  key  listings delimited by colons. Note that the output will be encoded in UTF-8
              regardless of any --display-charset setting. This format is useful when GnuPG is called
              from  scripts  and  other  programs as it is easily machine parsed. The details of this
              format are documented in the file ‘doc/DETAILS’, which is included in the GnuPG  source
              distribution.

install_files/ansible-base/roles/ossec-server/templates/send_encrypted_alarm.sh
```diff
@@ -1,5 +1,49 @@
...
+
+    # Error handling.
+    if [[ -z "${encrypted_alert_text}" || $? -ne 0 ]]; then
+        send_plaintext_fail_message
+    else
+        echo "${encrypted_alert_text}" | \
+            /usr/bin/mail -s "$(echo $SUBJECT | sed -r 's/([0-9]{1,3}\.){3}[0-9]{1,3}\s?//g' )" '{{ ossec_alert_email }}'
```
fowlslegs: Can you explain this regexp?

install_files/ansible-base/roles/ossec-server/templates/send_encrypted_alarm.sh
fowlslegs: It's also not explained anywhere in the script that SUBJECT is an env var, and I really think that env vars should be avoided at all costs (I won't reiterate my arguments to that extent--there are plenty of good articles on why env vars/ global state/ singletons suck). Even if you intend to keep it as such, you should do proper variable substitution when referencing it's value (`"${SUBJECT}"`).

install_files/ansible-base/roles/ossec-server/templates/send_encrypted_alarm.sh
redshiftzero: Hmm, looks like this is just from the original script

install_files/ansible-base/roles/ossec-server/templates/send_encrypted_alarm.sh
conorsch: You're right, it's not explicitly stated that SUBJECT is an env var, but nor was it in the [previous version of the script](https://github.com/freedomofpress/securedrop/blob/666f11fac726b5057657e432c3cce4d700cf65d3/install_files/ansible-base/roles/ossec-server/templates/send_encrypted_alarm.sh). We don't have much wiggle room on the env var approach: this script is run in a context where SUBJECT is defined via an environment variable.

install_files/ansible-base/roles/ossec-server/templates/send_encrypted_alarm.sh
conorsch: The regular expression was added as part of #915, to scrub IP addresses from the subject line (which is observable in plain text, since it's outside the encrypted message body).

PR #1680, opened 2017-04-28
Don't upgrade tor during playbook runs

### LINKED ISSUE #964: Apt upgrade via Ansible could break SSH over Tor sessions
There is a small edge case around the cron-apt tasks in the ansible playbook for existing installs that could stop the tor service and lock everyone out of the system until the servers are manually rebooted.

For existing installs subsequent ansible-playbook runs use ssh over a tor sessions. If there is an update to the tor package when the ansible playbook is ran again the cron-apt tasks will attempt to install them. The pre-install script for the tor package will stop tor which will also kill the ssh over tor session that ansible is using. This will leave tor stopped on the server until it is manually rebooted.

conorsch:
## Status

Ready for review

## Description of Changes

Fixes #964.

* If playbook is being run over Tor (via SSH ATHS), and the `tor` package is eligible for an upgrade, don't upgrade it, since an upgrade would break the SSH connection and cause the playbook to fail.
* If playbook is running over local IPv4 address, i.e. during the initial installation, go ahead and upgrade tor if possible, since we want to make sure we have the latest patches prior to generating ATHS info.

## Testing

You should perform two sets of tests to validate that both the first-run behavior and the existing-instance cases are handled well. To test the first-run behavior:

1. Provision staging machines.
2. Downgrade tor package: `sudo apt-get install tor=0.2.4.27-1build0.14.04.1 -y --force-yes`
3. Rerun common role: `ANSIBLE_ARGS="--tags common" vagrant provision /staging/`
4. Confirm tor package is upgraded (since SSH is _not_ over Tor).

To test the existing-instance behavior:

1. Add SSH ATHS values to your host torrc file, and restart tor _on host machine_.
2. Comment out the "Sandbox" line in `/etc/tor/torrc` (since it's not supported on the older version of tor) on guest machine.
3. Downgrade tor package: `sudo apt-get install tor=0.2.4.27-1build0.14.04.1 -y --force-yes` on guest machine.
4. Rerun common role over Tor: `SECUREDROP_SSH_OVER_TOR=1 ANSIBLE_ARGS="--tags common" vagrant provision /staging/`
5. Confirm tor package is **not** upgraded (since SSH _is_ over Tor). If the package is upgraded, the playbook will fail.

## Deployment

Made sure to handle first-run during install time while preventing playbook breakage when run against already provisioned instances. So, no special considerations necessary here.

- [ ] Unit and functional tests pass on the app-staging VM

For the app-staging VM, I got a single failure, for the `custom_notification` string check. Seems like the logic for writing that is perhaps not deterministic, given how Vagrant is generating the INI inventory file. Can be solved with host_vars for clearer variable precedence. 

- [x] Testinfra tests pass on the mon-staging VM


conorsch:
> So if the task that runs aptitude hold tor is not skipped, then it will always register it's result as changed?

That's correct: by default all invocations of the `command` module reported changed=true.

Issue #1687, opened 2017-05-04
Handle BadTokenException in /admin/2fa


redshiftzero:
# Steps to reproduce

1. Provision development VM
2. Make an administrator account
3. Sign in 
4. Click "Admin"
5. Click "Edit" on the admin account
6. Click "Reset two factor authentication (Google Authenticator)"
7. Put in "123456" as the verification code
8. Put in "123456" again as the verification code

# What should happen

Ideally a flashed message would appear saying "Two factor token failed to verify".

# What actually happens

The `BadTokenException` in `Journalist.verify_token()` is not handled:

![screen shot 2017-05-04 at 4 39 38 pm](https://cloud.githubusercontent.com/assets/7832803/25729191/a74188aa-30e8-11e7-868d-aad7e6852a48.png)

# Proposed Fix

Catch the exception and flash the friendly error message as suggested above. Whoever fixes this ideally should also write a unit test to guard against future regressions.

PR #1692, opened 2017-05-08
Handle BadTokenException in /admin/2fa

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes defect #1687 reported by a SecureDrop user.

- Adds regression unit test
- Handles the `BadTokenException` in `/admin/2fa` and flashes a message to the user

## Post Mortem

A bug like this can slip into production because:
* Overuse of mocks
* No 2FA unit or functional tests
* "if running the tests do A else do B" logic in app code, meaning that B will never be tested

## Testing

1. Provision development VM
2. Make an administrator account
3. Sign in
4. Click "Admin"
5. Click "Edit" on the admin account
6. Click "Reset two factor authentication (Google Authenticator)"
7. Put in "123456" as the verification code
8. Put in "123456" again as the verification code

You should see a flashed message "Two factor token failed to verify"

## Deployment

No special considerations for deployment

### DIFF
--- a/securedrop/tests/test_2fa.py
+++ b/securedrop/tests/test_2fa.py
@@ -1,0 +1,68 @@
+# -*- coding: utf-8 -*-
+import os
+import random
+
+from flask import url_for
+from flask_testing import TestCase
+
+os.environ['SECUREDROP_ENV'] = 'test'
+from db import (db_session, BadTokenException)
+import journalist
+import utils
+
+# Smugly seed the RNG for deterministic testing
+random.seed('¯\_(ツ)_/¯')
+
+
+class TestJournalist2FA(TestCase):
+    def create_app(self):
+        return journalist.app
+
+    def setUp(self):
+        utils.env.setup()
+
+        self.admin, self.admin_pw = utils.db_helper.init_journalist(
+            is_admin=True)
+
+    def tearDown(self):
+        utils.env.teardown()
+        # TODO: figure out why this is necessary here, but unnecessary in all
+        # of the tests in `tests/test_unit_*.py`. Without this, the session
+        # continues to return values even if the underlying database is deleted
+        # (as in `shared_teardown`).
+        db_session.remove()
+
+    def _login_admin(self):
+        valid_token = self.admin.totp.now()
+        resp = self.client.post(url_for('login'),
+                                data=dict(username=self.admin.username,
+                                          password=self.admin_pw,
+                                          token=valid_token))
+
+    def test_bad_token_fails_to_verify_on_admin_new_user_two_factor_page(self):
+        self._login_admin()
+
+        # Create and submit an invalid 2FA token
+        invalid_token = unicode(int(self.admin.totp.now()) + 1)
+        resp = self.client.post(url_for('admin_new_user_two_factor',
+                                        uid=self.admin.id),
+                                data=dict(token=invalid_token))
+
+        self.assertIn('Two factor token failed to verify', resp.data)
+
+        # last_token should be set to the invalid token we just tried to use
+        self.assertEqual(self.admin.last_token, invalid_token)
+
+        # Submit the same invalid token again
+        resp = self.client.post(url_for('admin_new_user_two_factor',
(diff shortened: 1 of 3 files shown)


securedrop/tests/test_2fa.py
```diff
@@ -0,0 +1,68 @@
...
+        self.admin, self.admin_pw = utils.db_helper.init_journalist(
+            is_admin=True)
+
+    def tearDown(self):
+        utils.env.teardown()
+        # TODO: figure out why this is necessary here, but unnecessary in all
+        # of the tests in `tests/test_unit_*.py`. Without this, the session
+        # continues to return values even if the underlying database is deleted
+        # (as in `shared_teardown`).
+        db_session.remove()
```
fowlslegs: That is weird since the first line of the last function is this one. It's not weird that you need to call this method on `db_session` despite the database being deleted (that part makes sense), but that calling it the first time doesn't work.

securedrop/tests/test_2fa.py
```diff
@@ -0,0 +1,68 @@
...
+        resp = self.client.post(url_for('login'),
+                                data=dict(username=self.admin.username,
+                                          password=self.admin_pw,
+                                          token=valid_token))
+
+    def test_bad_token_fails_to_verify_on_admin_new_user_two_factor_page(self):
+        self._login_admin()
+
+        # Create and submit an invalid 2FA token
+        invalid_token = unicode(int(self.admin.totp.now()) + 1)
```
fowlslegs: Why not just `u'000000'` or `u'invalid'`?

redshiftzero: Sure, fixed in 1f4d4b2

securedrop/tests/test_2fa.py
redshiftzero: As so I copied this thinking it was boilerplate from `test_journalist.py`, but after testing it's not necessary so I have deleted it in 1f4d4b2

redshiftzero:
Thanks for review @fowlslegs! Good call on `/account/2fa`, I've added the same exception handling and a similar unit test there too. 

Note the only code I am trying to test here is the exception handling in `/admin/2fa`, not the code in `/admin/reset-2fa-totp` or `/admin/reset-2fa-hotp`. It's just intended to be a regression test to catch the bug reported in #1687, which is indeed caught by the unit tests as written here. 

Also note that I'm not trying to do major refactoring here. There is plenty to refactor, but going crazy refactoring without excellent tests is a recipe for a Really Bad Time: we will probably break things and not realize. Here I'm just trying to get things from "broken and untested" to "working and tested".

In terms of catching bugs like this in the future, I totally agree that docstrings are great and additional documentation would be solid. But of the **utmost importance** is that functionality like this is tested with both unit (unless the functionality is from a dependency and is under test there) and end-to-end functional/integration testing. We have code paths that are currently untested, as evidenced by this bug found in production. The test suite _should be_ expressive and natural documentation pinning down the expected behavior of the code.

fowlslegs:
Let's talk in person about this one tomorrow.

fowlslegs:
I think the better solution to this is just to move the TOTP re-use protection code out of `verify_token()` and into `Journalist.login()`. Such code does not belong in a function which serves to confirm to a user that the update of their OTP secret succeeded.

redshiftzero:
Sure, a fair point! Added that change in 0c8e561. Let me know if you have further thoughts on this one

securedrop/tests/test_2fa.py
```diff
@@ -0,0 +1,93 @@
...
+
+from flask import url_for
+from flask_testing import TestCase
+
+os.environ['SECUREDROP_ENV'] = 'test'
+from db import db_session
+import journalist
+import utils
+
+# Smugly seed the RNG for deterministic testing
```
fowlslegs: I love how this has become boilerplate.

redshiftzero:
Ah, I left the tests as is here as they are intended as regression tests - as in, they reproduce this bug this PR is fixing such that if it were to come back it's caught

PR #1694, opened 2017-05-08
Handle IntegrityError on /admin/add 

### LINKED ISSUE #1693: Handle IntegrityError on /admin/add
# Steps to reproduce

1. Provision development VM
2. Make an administrator account
3. Sign in 
4. Click "Admin"
5. Click "Add user" button
6. Fill in the form, setting the username to username of an existing user
7. Click "Add user"

# What should happen

Ideally a flashed message would appear saying "That username is already in use".

# What actually happens

screen shot 2017-05-08 at 3 36 59 pm

# Proposed Fix

Catch the exception and flash the friendly error message as suggested above. Whoever fixes this ideally should also write a unit test to guard against future regressions.

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #1693, an unhandled exception produced when an admin tried to create a user with the username of an existing user.

* Adds regression unit test to reproduce defect #1693 
* On `/admin/add` we now handle the exception, roll back the transaction, and show a flashed message to the user

## Testing

1. Provision development VM
2. Make an administrator account
3. Sign in
4. Click "Admin"
5. Click "Add user" button
6. Fill in the form, setting the username to username of an existing user
7. Click "Add user"

You should see the following:

screen shot 2017-05-08 at 4 22 34 pm

## Deployment

### DIFF
--- a/securedrop/tests/test_manage.py
+++ b/securedrop/tests/test_manage.py
@@ -1,10 +1,43 @@
 # -*- coding: utf-8 -*-
 
 import manage
+import mock
+from StringIO import StringIO
+import sys
 import unittest
+import __builtin__
+
+import utils
 
 
 class TestManagePy(unittest.TestCase):
     def test_parse_args(self):
         # just test that the arg parser is stable
         manage.get_args()
+
+
+class TestManagementCommand(unittest.TestCase):
+    def setUp(self):
+        utils.env.setup()
+
+    def tearDown(self):
+        utils.env.teardown()
+
+    @mock.patch("__builtin__.raw_input", return_value='N')
+    @mock.patch("manage.getpass", return_value='testtesttest')
+    @mock.patch("sys.stdout", new_callable=StringIO)
+    def test_exception_handling_when_duplicate_username(self, mock_raw_input,
+                                                        mock_getpass,
+                                                        mock_stdout):
+        """Regression test for duplicate username logic in manage.py"""
+
+        # Inserting the user for the first time should succeed
+        return_value = manage._add_user()
+        self.assertEqual(return_value, 0)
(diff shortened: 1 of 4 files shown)


redshiftzero:
Thanks for review @fowlslegs! I've done the same to `manage.py` and added a similar regression unit test (albeit with a lot more mocked out functions because the functionality in `_add_user()` in `manage.py` is more complicated than that in `journalist.py`).

securedrop/tests/test_manage.py
```diff
@@ -1,10 +1,48 @@
...
+    def create_app(self):
+        return journalist.app
+
+    def setUp(self):
+        utils.env.setup()
+
+    def tearDown(self):
+        utils.env.teardown()
+
+    @mock.patch("__builtin__.raw_input", return_value='N\n')
```
fowlslegs: Does just `'raw_input'`  not work? I thought everything in `__builtin__` is implicitly, unqualified-ly imported into each new namespace.

It seems like the `return_value` should not have the trailing `\n` (same goes for the line below). Otherwise wouldn't it be the case that every time an account was created from `manage.py` that the username would contain a newline character? I know it doesn't matter as far as our primary concern, which is ensuring this test is testing what it's supposed to be testing, but it does seem like a legitimate secondary concern to try to mock the return values of functions realistically.

redshiftzero: `raw_input` does not work:  
```
TypeError: Need a valid target to patch. You supplied: 'raw_input'
```

redshiftzero: :+1: removed `/n` in ba782d9

redshiftzero:
OK using `unittest.TestCase` as the base class as of ba782d9

PR #1702, opened 2017-05-12
Minor content changes/corrections.

justintroutman:
## Status

Ready for review.

## Description of Changes

Small batch of minor content changes (e.g., version number, spelling correction).


docs/conf.py
```diff
@@ -59,9 +59,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '0.3.12'
+version = '0.4'
 # The full version, including alpha/beta/rc tags.
-release = '0.3.12'
+release = '0.4'
```
redshiftzero: We don't want to change any version stuff here since that is one of the tasks that is done by update_version.sh, a script we run when cutting a new release

PR #1703, opened 2017-05-12
Infra-378 CI teardown

msheiny:
## Status

Read for review

## Description of Changes

In CI, we have been seeing the issue that any of the `testinfra` tests fail - the clean-up task was not being called. This makes sense upon further investigation of the current bash trap layout.

To fix this, I broke out the CI testing bash script with it's own clean-up trap. I tried to put this trap at the highest level script in an earlier branch/commit and I remember having issues problems. I'm crossing fingers that these latest changes fix that issue - they did for me locally using the same CI scripts.

## Testing

Nothing to do manually here, just confirm:

- CI environment gets torn down at commit `142a8d5` (which introduced a guaranteed CI failure)
- CI environment gets torn down at commit `fcecb75` (which reverted that CI failure)

## Deployment

N/A - this only affects CI

N/A - only affects CI scripts

### DIFF
--- a/devops/scripts/ci-tester.sh
+++ b/devops/scripts/ci-tester.sh
@@ -1,0 +1,42 @@
+#!/bin/bash -e
+#
+#
+#
+trap "make ci-teardown" ERR
+
+# Quick hack to pass IPs off to testinfra
+for host in app mon; do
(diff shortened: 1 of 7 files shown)


msheiny:
rebased to latest `develop`

PR #1700, opened 2017-05-11
Fixes reboot-and-wait-for strategy across all platforms

conorsch:
## Status

Ready for review.

## Description of Changes

Fixes #951. Changes introduced in #1666 caused variability in the "reboot and wait for" strategy (RAWFS) used in several places throughout the site config. To merge #1666 and boost the automated tests, we opted for the RAWFS that worked well with AWS EC2 instances—but failed on Vagrant  local VMs. The new strategy works equally well in both contexts.

Changes proposed in this pull request:

* Notifies a single handler for reboot-and-wait-for (DRY)
* Reduces number of reboots from 3 to 1 on clean staging install

There are some non-obvious side-effects of the handler approach, namely:

  1. The handler is declared in the `common` role and therefore is only available to roles within the same play. 
  2. The only way to notify a handler is to have a task that reports changed=true; thus the dummy tasks that simply run `/bin/true` based on file existence.

Down the road I'd like to move toward leaning more heavily on upstream for the Ansible config, which would allow us to remove many of the custom roles for common hardening actions, e.g. sysctl and iptables management. In particular, we have a [well-maintained grsecurity role](https://github.com/freedomofpress/ansible-role-grsecurity) that we're _not_ using in this project yet. Once we include that work, we'll be in a much stronger position. 

## Testing

  1. Confirm CircleCI staging env passes.
  2. Confirm a fresh staging build on Vagrant VMs passes.

## Deployment

The changes presented here modify the implementation of reboots in the playbooks, but with an eye toward preserving existing functionality. That said, vigorous QA on hardware is warranted, although I suggest we include that on the general 0.4 QA testing list.

### If you made changes to the app code:

- [x] Unit and functional tests pass on the development VM
- [ ] Unit and functional tests pass on the app-staging VM

### If you made changes to the system configuration:

**Note:** Had to bump Ansible down to v2.2.1, since 2.2.2 showed errors with testinfra. If you receive the same results, say so, and we'll make sure to address in #1146.

- [x] Testinfra tests pass on the development VM
- [ ] Testinfra tests pass on the app-staging VM
- [ ] Testinfra tests pass on the mon-staging VM

PR #1707, opened 2017-05-13
Safely reload tor service when connecting over Tor

### LINKED ISSUE #940: ansible notify task to reload tor

conorsch:
## Status

Ready for review.

## Description of Changes

Fixes #940. (Requires #1700, so review and merge that first.)

* If connected over tor and the tor service needs to be restarted, don't do so blindly; send the restart command via `async`, then sit down and wait for 90 seconds before attempting to connect again, in order to give the Authenticated Tor Hidden Services for SSH time to come back up.

## Testing

1. Provision staging VMs.
2. Add the ATHS values to your host system's torrc.
3. `export SECUREDROP_SSH_OVER_TOR=1`.
4. Confirm via `vagrant ssh-config app-staging` that the Onion URLs are listed.
5. Edit `install_files/ansible-base/roles/tor-hidden-services/templates/torrc`, e.g. adding comments, then rerun the playbook over Tor `vagrant provision /stag/`. 

Protip: you can use `ANSIBLE_ARGS="--tags tor"` to run _only_ the tor role, which should save a lot of time. _Without_ the new handler, editing the torrc file while connected over tor will cause the playbook to fail. With the new handler, things should proceed well.

It's a shortcoming that we're forced to hardcode the wait time. 90s seems like a sane value; during testing, it never took longer than 60s for the services to come back up (polled them in a bash while loop), so 90s is an attempt to be conservative.

## Deployment

This is designed specifically to prevent Admin frustration if torrc changes land. Right now they haven't, so we can reasonable expect torrc files to be the same between 0.3.12 and 0.4—still, this sound strategy rules out a potential source of frustration and breakage.

## Checklist

Leaning on CI checks here, and I want confirmation from manual testing via vagrant that the reboot works OK over tor. We need manual verification because the prod tests don't run in CI (yet), and prod does SSH over tor.

### DIFF
--- a/install_files/ansible-base/roles/tor-hidden-services/handlers/main.yml
+++ b/install_files/ansible-base/roles/tor-hidden-services/handlers/main.yml
@@ -1,5 +1,3 @@
 ---
 - name: restart tor
-  service:
-    name: tor
-    state: restarted
+  include: restart-tor-carefully.yml
--- a/install_files/ansible-base/roles/tor-hidden-services/handlers/restart-tor-carefully.yml
+++ b/install_files/ansible-base/roles/tor-hidden-services/handlers/restart-tor-carefully.yml
@@ -1,0 +1,35 @@
+---
+# Meta handler to bounce tor service sanely. If the SSH connection
+# for the remote host is over Tor (i.e. host ends in `.onion`), then
+# bounce the tor service via fire-and-forget, wait, then reestablish
+# the connection after polling for the service to come back up.
+
+# Registering a concise variable for use in conditionals, essentially
+# deciding "Are we connected via SSH over Tor or not?"
+- name: Register host name to wait for.
+  set_fact:
+    _hostname_to_wait_for: "{{ ansible_ssh_host|default(ansible_host) }}"
+
+# If we're not connected over Tor, bounce the service as usual.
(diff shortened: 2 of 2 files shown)


conorsch:
Blocked by #1700; once that's in, this PR should be rebased on top of latest develop.

conorsch:
Rebased on top of latest develop, to include changes from #1700.

msheiny:
Coool :+1:  -- I got this passing on my end under vagrant - thanks for the detailed testing instructions as usual. 

> It's a shortcoming that we're forced to hardcode the wait time. 90s seems like a sane value; during testing, it never took longer than 60s for the services to come back up (polled them in a bash while loop), so 90s is an attempt to be conservative.

I would have preferred something dynamic here but yeah I agree that sounds like more trouble than its worth for this out-lier scenario. 

Anywhoo I didnt realize you could have handlers call entirely other playbooks like this. I'm learning something new today!!! :books:

Issue #1634, opened 2017-03-23
Linting/testing of our shell scripts


fowlslegs:
#1633 was a reminder that our shell scripts often only receive manual review before making it into develop, and that some basic linting--not even proper tests, which are a lot more difficult to implement and run regularly given the context in which most of these scripts are run--could go a long way.

Shellcheck seems like a good option and has [a section in the README](https://github.com/koalaman/shellcheck#travis-ci-setup) on use in Travis. Shellcheck caught #1633. I think it would be a good idea to add it to CI.

Would love to hear other ideas and thoughts on what we could do to improve (especially cheaply) the robustness of our shell script analysis.

conorsch:
Running `shellcheck tail_files/install.sh` locally shows some good output, so I'm inclined to go with that solution for CI. There's a barebones approach in `bash -n tails_files/install.sh` that would have caught the error described in #1633, but shellcheck finds more, and provides more informative output.

conorsch:
Versioning is difficult here:

* Travis: 0.4.6
* Ubuntu Trusty (e.g. dev VM): 0.3.3
* Debian Stretch (example workstation): 0.4.4-4

The differences are substantial enough that a single ruleset cannot be used in all environments. So we're going to have to use a container to bottle up the functionality and make the versioning standard across all environments.

PR #1709, opened 2017-05-15
INFRA-384 - These AWS Costs are too dang high!! 

### LINKED ISSUE #1708: Convert CI `build` VM to docker container
During the `Great CI Event` (GCE) of 2017 - where `staging` provisioning and tests were thrown into the pipeline, the price of such greatness was a stab to FPF wallets. Here is a dramatic re-enactment of what went down:

:hocho: :moneybag: 

To explain, we spent more than desired on cloud services during the course of introducing the automated CI system for a few days. There are a number of initiatives currently underway to reign in these cloud costs. The first low hanging fruit is to convert the `build` vm (the machine that handles packaging debian packages) into a docker instance. This requires a number of tweaks to playbooks, to the testinfra scripts, and to the circle CI configuration to make this work-right.

msheiny:
## Status

:checkered_flag:  Ready for review!

## Description of Changes

* Closes  #1708  - Specifically, moves our ec2 `build` instance to a docker instance to save us CI costs :moneybag: . Had to make tweaks throughout to adjust the playbooks. Added an upstream compatible version of the `synchronize.py` module that is friendly with docker.
* modifications to how testinfra was being called against the build instance under CI
* switched to circleci 2.0 which allows remote docker building/triggering. It looks like I currently lost the ability for junit.xml test parsing. address that issue in  #1710 (I'm getting fatigued from running these tests over and over so even though its a trivial 2 lines to add, I'm bouncing that to another ticket.)

## Testing

If this works in CI :pray: - than we are good to go. For local testing, you'll need AWS credentials as well as a working docker setup.* Key item to note is that in CI i started using `t2.small` instances so while you are laying out the environment variables make sure you set `export CI_AWS_TYPE=t2.small`. 

Note - if you are on a mac it looks like docker recently released an alternative to the ol' docker-machine + virtualbox combo that takes advantage of native OSX virtualization frameworks. See [docker for mac](https://docs.docker.com/docker-for-mac/install/). Another solution is to use `docker-machine` with a cloud provider like AWS / DigitalOcean. I probably owe you some documentation updates on this. Created ticket #1711 to address.

## Deployment

Had to bring in upstream `synchronize.py` module from `ansible 2.3.0` in order for our rsync logic to play nice with docker. When we decide to switch project to ansible 2.3+ in the future, we can come back and tear this version out of our repo.

Not applicable.

Not applicable.


.circleci/config.yml
```diff
@@ -0,0 +1,33 @@
+version: 2
+jobs:
+  build:
+    working_directory: ~/sd
+    docker:
+      - image: msheiny/pressfreedomci:latest
```
msheiny: in case you are wondering this image is built from -- https://github.com/freedomofpress/docker-debian-systemd . Probably want to rebrand and shove under the fpf banner (instead of under the docker hub `msheiny` name) in the future.

msheiny: Umm nope - nevermind see line 27 under `devops/playbooks/aws-ci-startup.yml` for reference to ^^ image. The image above is from the `Dockerfile` below..... Fudge on second thought, I should add that to this repo for tracking. It is merely used as a `jump box` image to lauch other docker images from.

```
FROM debian:stretch


ENV DOCKER_BUCKET get.docker.com
ENV DOCKER_VER 17.05.0-ce
ENV DOCKER_SHA256_x86_64 340e0b5a009ba70e1b644136b94d13824db0aeb52e09071410f35a95d94316d9

RUN apt-get update && \
    apt-get install -y flake8 make virtualenv ccontrol libpython2.7-dev \
            libffi-dev libssl-dev libyaml-dev python-pip curl git &&\
    apt-get clean

RUN curl -L -o /tmp/docker-${DOCKER_VER}.tgz https://get.docker.com/builds/Linux/x86_64/docker-${DOCKER_VER}.tgz; \
	echo "${DOCKER_SHA256_x86_64} /tmp/docker-${DOCKER_VER}.tgz" | sha256sum -c -; \
	cd /tmp && tar -xz -f /tmp/docker-${DOCKER_VER}.tgz; \
	mv /tmp/docker/* /usr/bin

CMD /bin/bash
```

redshiftzero: :+1: to adding this Dockerfile to the repo

devops/ansible_env
```diff
@@ -1,3 +1,4 @@
+export BUILD_NUM=${CIRCLE_BUILD_NUM}
```
redshiftzero: Ah, won't this not work if we try to run this locally? I get the following error when trying to spin this up:

```
TASK [Ensure CI environment vars are declared.] ********************************
    "assertion": "aws_ec2_ci_tags.build_num != \"\"", 
    "changed": false, 
    "evaluated_to": false, 
    "failed": true
}
```

(I don't hit this if I replace your line with `export BUILD_NUM=redshiftzero`)

.circleci/config.yml
```diff
@@ -0,0 +1,33 @@
...
+      # the working_directory creates that directory first
+      - run: rm -r ~/sd
+      - checkout:
+          path: ~/sd
+      - run:
+          name: Installation pre-reqs
+          command: |
+            easy_install -U pip
+            pip install pip-tools
+            pip-sync testinfra/requirements.txt
```
fowlslegs: Since this is a fresh filesystem why not skip installing `pip-tools` and just `pip install -U -r testinfra/requirements.txt`?

devops/playbooks/aws-ci-startup.yml
```diff
@@ -1,10 +1,39 @@
...
+        image: "{{ docker_image_sd_build }}"
+        command: /lib/systemd/systemd
+        devices:
+          - "/sys/fs/cgroup:/sys/fs/cgroup:ro"
+        privileged: true
+      tags: docker
+      with_items: "{{ docker_containers }}"
+  vars:
+    ci_environment: "{{ ansible_env.CI_SD_ENV | mandatory }}"
+    docker_image_sd_build: "msheiny/debian-jessie-systemd"
```
fowlslegs: I don't see an upstream repository listed on https://hub.docker.com/r/msheiny/debian-jessie-systemd/.

devops/playbooks/aws-ci-startup.yml
```diff
@@ -1,10 +1,39 @@
...
+
+    - name: Start build container
+      docker_container:
+        name: "{{ item }}"
+        hostname: "{{ item }}"
+        image: "{{ docker_image_sd_build }}"
+        command: /lib/systemd/systemd
+        devices:
+          - "/sys/fs/cgroup:/sys/fs/cgroup:ro"
+        privileged: true
```
fowlslegs: Assuming your image was built from https://github.com/freedomofpress/docker-debian-systemd, you shouldn't need this (unless there's some reason other than systemd that you're specifying it).

devops/scripts/ci-tester.sh
```diff
@@ -31,7 +31,7 @@ if [ "$?" == "0" ]; then
     esac
 fi
 
-if [ -z "$TEST_REPORTS" ]; then
+if [ -z "$TEST_REPORTS" ] || [ "$TEST_REPORTS" == '${CIRCLE_TEST_REPORTS}' ]; then
```
msheiny: > Also, you have ${CIRCLE_TEST_REPORTS} wrapped in single, not double, quotes, so it will be treated as a literal string and variable expansion will not happen.

Yeah... that is intentional. There is a bug in CircleCI 2.0 where all of a sudden the env vars i defined from the web UI arent getting expanded. I'm trying to catch that scenario and fix it without breaking currently passing functionality in CircleCI 1.0.

.circleci/config.yml
msheiny: Ive run into issues before in CI -- i think it was circleCI.... where they had a preinstalled version of a python dependency i was installing and had trouble with `install -U` getting me to the right version. Since we are now running under a docker image we control -- i dont think thats as much of an issue really. I mean, I'm fine with taking it out.

devops/playbooks/aws-ci-startup.yml
msheiny: `you shouldn't need this ` 

Yeah .. i "shouldnt" but i kept trying to make it work without it and it was failing. That honestly could have been in combination with another setting I was using at the time. I'm not dying to go back and debug though in this ticket though but I'll be happy to debug in another PR/ticket. Since the image is primarily running in CI in an environment outside of our server's I dont see this as urgent to get into this PR.

fowlslegs: :+1:

devops/scripts/ci-runner.sh
```diff
@@ -10,9 +10,11 @@ ansible staging -b -m hostname -a "name={{inventory_hostname}}"
 # Hotfix for missing hostname attached to localhost
 ansible staging -b -m lineinfile -a "dest=/etc/hosts line='127.0.0.1 localhost {{inventory_hostname}}' regexp='^127\.0\.0\.1'"
 
+SSH_USER=`whoami`
```
msheiny: :heavy_check_mark:

fowlslegs:
Oh yeah, https://google.github.io/styleguide/shell.xml is kind of our unofficial style guide. We've tried to stay pretty close to it without enforcing everything as some strict requirement. Also, running `shellcheck` on your scripts is always a good idea (see https://github.com/freedomofpress/securedrop/issues/1634).

.circleci/config.yml
```diff
@@ -0,0 +1,33 @@
...
+    docker:
+      - image: msheiny/pressfreedomci:latest
+        environment:
+          FPF_CI: true
+          CI_SD_ENV: staging
+          CI_AWS_TYPE: t2.small
+          FPF_GRSEC: false
+          TEST_REPORTS: /root/sd
+    steps:
+      # This is really dumb that Im doing this, basically
```
msheiny: ehh so what happened when i tested this...

Basically say you have a working directory defined, like i have `working_directory: ~/sd` .
Securedrop kept getting cloned to `~/sd/securedrop` and then all future commands were running out of `~/sd` . I tried to clone to `~/sd` but obviously git complains that the directory already exists.  Now that i look at the docs.. I could have actually just defined a different `working_directory` for each instance of `run` ..... actually yeah let me do that. its the same number of lines and then i lose all that comment crap.

devops/playbooks/aws-ci-startup.yml
msheiny: That was set-up rather hastily to be honest. I didnt have docker hub build off a repository, I built it locally and then pushed it up from that freedomofpress systemd repo you pointed to before.

.circleci/config.yml
fowlslegs: Say you have pypkg 1.0 installed. If you want to install a different version, regardless if it's an earlier or later version, you have to pass `-U`. It could be the case that you installed the right version with your `pip` command, but the preinstalled version was not installed with pip and is located under a different directory tree which comes earlier in the `PYTHONPATH`. Usually this isn't the case as where pip installs is before where package managers install in the `PYTHONPATH`.

As we've all experienced irregularities with `pip-sync` it seems best if you do take it out.

.circleci/config.yml
fowlslegs: > Securedrop kept getting cloned to ~/sd/securedrop

That doesn't seem right. According to the reference [here](https://circleci.com/docs/2.0/configuration-reference/#checkout): it should be the case that that step will fail. Of course, as I'm discovering, this documentation sucks and doesn't make sense. I mean if path defaults to working_directory, and working_directory is created if it doesn't exist, then checkout without a path argument should always fail.

Just follow your heart on this one I guess :heart:

devops/scripts/ci-tester.sh
fowlslegs: Uhh, okay?

devops/playbooks/aws-ci-startup.yml
fowlslegs: Would be good to have a repo for it. For the FOSS points and all that. Out of the scope of this PR clearly.

msheiny:
Doing lots of interactive rebasing on the last two commits. Make sure you do a fresh pull if you are messing with this branch.

devops/scripts/ci-tester.sh
```diff
@@ -16,27 +16,28 @@ done
...
     "development")
         ./testinfra/test.py development
         ;;
     esac
 fi
 
-if [ -z "$TEST_REPORTS" ]; then
-    export TEST_REPORTS=`pwd`
+if [[ -z "${TEST_REPORTS}" ]] || [[ "${TEST_REPORTS}" == '${CIRCLE_TEST_REPORTS}' ]]; then
+    TEST_REPORTS=$(pwd)
```
fowlslegs: `TEST_REPORTS="$(pwd)"`

testinfra/vars/app-staging.yml
```diff
@@ -22,8 +22,8 @@ apache_source_log: /var/log/apache2/source-error.log
 apache_allow_from: all
 
 dns_server: "8.8.8.8"
-mon_ip: 10.0.1.3
-app_ip: 10.0.1.2
+mon_ip: 172.16.1.203
+app_ip: 172.16.1.69
```
conorsch: I spy a regression here... these changes should not have been committed to version control.

msheiny: ahhhhhhhh :hankey: I think i should fix that functionality to use a file that is ignored from git. Its too easy to make a mistake like I just did.... :cry:

PR #1719, opened 2017-05-19
Test request limit size on JI

fowlslegs:
We set this parameter for the JI Apache config, but do not validate it, as we do for the SI. Since staging is in big churn right now as far as local spin up goes, I'm testing this change by making a PR.


fowlslegs:
Looks like that worked as expected, going to rebase and force push in a sec.

PR #1720, opened 2017-05-19
Reject empty replies

### LINKED ISSUE #1715: Empty replies can be sent to a source
# Steps to reproduce

1. Provision development VM
2. Submit some documents through the source interface.
3. Create a journalist account.
4. Click on a source.
5. Attempt to send an empty reply.

# What should happen

A flashed message should appear indicating that text must be entered in the reply field.

# What actually happens

The empty reply is uploaded.

# Comment

Whoever fixes this ideally should definitely write a functional or unit test to guard against future regressions.

fowlslegs:
## Status

So ready.

## Description of Changes

Flashes warning about and rejects zero-length replies. Fixes #1715.

Also, does proper logging and rollback in case of error (e.g., one journalist may delete a source while another is writing a reply). Logging is cautious because of the sensitive nature of replies. See in-line comment for details.

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -599,9 +599,7 @@ def reply():
         flash("An unexpected error occurred! Please check the application "
               "logs or inform your adminstrator.", "error")
         # We take a cautious approach to logging here because we're dealing
-        # with responses to sources. TODO: It may be possible to provide more
-        # detailed/ informative log errors here, but we'll need to ensure no
-        # confidential or possibly deanonymizing information ends up there.
+        # with responses to sources.
         app.logger.error("Reply from '{}' failed: {}!".format(g.user,
                                                               exc.__class__))
         db_session.rollback()
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -595,7 +595,7 @@ def reply():
         db_session.add(g.source)
         db_session.add(reply)
         db_session.commit()
-    except:
+    except Exception as exc:
         flash("An unexpected error occurred! Please check the application "
               "logs or inform your adminstrator.", "error")
         # We take a cautious approach to logging here because we're dealing
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -592,7 +592,6 @@ def reply():
     reply = Reply(g.user, g.source, filename)
 
     try:
-        db_session.add(g.source)
         db_session.add(reply)
         db_session.commit()
     except Exception as exc:
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -607,7 +607,8 @@ def reply():
         db_session.rollback()
     else:
         flash("Thanks! Your reply has been stored.", "notification")
-        return redirect(url_for('col', sid=g.sid))
+
+    return redirect(url_for('col', sid=g.sid))
 
 
 @app.route('/regenerate-code', methods=('POST',))
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -576,18 +576,38 @@ def download_single_submission(sid, fn):
 @app.route('/reply', methods=('POST',))
 @login_required
 def reply():
+    msg = request.form['msg']
+
+    # Reject empty replies
+    if not msg:
+        flash("You cannot send an empty reply!", "error")
+        return redirect(url_for('col', sid=g.sid))
+
     g.source.interaction_count += 1
     filename = "{0}-{1}-reply.gpg".format(g.source.interaction_count,
                                           g.source.journalist_filename)
-    crypto_util.encrypt(request.form['msg'],
+    crypto_util.encrypt(msg,
                         [crypto_util.getkey(g.sid), config.JOURNALIST_KEY],
                         output=store.path(g.sid, filename))
     reply = Reply(g.user, g.source, filename)
-    db_session.add(reply)
-    db_session.commit()
 
-    flash("Thanks! Your reply has been stored.", "notification")
-    return redirect(url_for('col', sid=g.sid))
+    try:
+        db_session.add(g.source)
+        db_session.add(reply)
+        db_session.commit()
+    except:
+        flash("An unexpected error occurred! Please check the application "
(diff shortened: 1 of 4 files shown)


securedrop/journalist.py
```diff
@@ -575,18 +575,38 @@ def download_single_submission(sid, fn):
...
+        db_session.commit()
+    except:
+        flash("An unexpected error occurred! Please check the application "
+              "logs or inform your adminstrator.", "error")
+        # We take a cautious approach to logging here because we're dealing
+        # with responses to sources. TODO: It may be possible to provide more
+        # detailed/ informative log errors here, but we'll need to ensure no
+        # confidential or possibly deanonymizing information ends up there.
+        app.logger.error("Reply from '{}' failed: {}!".format(g.user,
+                                                              exc.__class__))
```
redshiftzero: `exc` is undefined

securedrop/journalist.py
```diff
@@ -575,18 +575,38 @@ def download_single_submission(sid, fn):
...
                         [crypto_util.getkey(g.sid), config.JOURNALIST_KEY],
                         output=store.path(g.sid, filename))
     reply = Reply(g.user, g.source, filename)
-    db_session.add(reply)
-    db_session.commit()
 
-    flash("Thanks! Your reply has been stored.", "notification")
-    return redirect(url_for('col', sid=g.sid))
+    try:
+        db_session.add(g.source)
```
redshiftzero: I think as written this will throw an `InvalidRequestError` because the underlying source has been deleted. But stepping back a bit: also recall that adding back a source doesn't regenerate the source directory / source GPG key that will be deleted when the source is deleted, so this line can be removed.

fowlslegs: >I think as written this will throw an InvalidRequestError because the underlying source has been deleted. 

Only in scenario I described where one journalist deletes a source while another journalist is writing a reply to that source.

>But stepping back a bit: also recall that adding back a source doesn't regenerate the source directory / source GPG key that will be deleted when the source is deleted, so this line can be removed.

So while I wrote this just to be explicit, it's actually harmful in either (i) the case described above or (ii) the same case as above except where a source deletes their own account instead of another journalist. In either of these cases as you point out the source will get added back to the database, but the necessary and complimentary filesystem actions executed during `Source` initialization will not be performed. Not only does this screw up system state, it partially un-deletes an account and associated data that is supposed to be deleted.

My explicitness is not necessary as items returned by queries already present in the database are automatically tracked in the session without needed to be explicitly added:

> add() is used to place instances in the session. For transient (i.e. brand new) instances, this will have the effect of an INSERT taking place for those instances upon the next flush. For instances which are persistent (i.e. were loaded by this session), they are already present and do not need to be added.

-- http://docs.sqlalchemy.org/en/latest/orm/session_basics.html#adding-new-or-existing-items

redshiftzero: Agreed

fowlslegs: I realized right after posting this that what I was saying wasn't consistent, but needed to investigate some things in order to properly correct myself. If it's true that query results are implicitly tracked, then there's really no significance in explicitly calling `db_session.add(g.source)`. That is, it should neither solve nor create problems; everything should behave the same under the same conditions regardless of the presence of that line. Note, I have omitted it now because it's redundant, not for other reasons discussed above.

In the source is deleted scenario, no exception will be raised--regardless of when the deletion of the source happens between when the query for the source object in `journalist.get_source()` succeeds and when `db_session.commit()` is called in `journalist.reply()`--until `db_session.commit()` is called. In this interim period the state of the session is detached from the state of the database. If the source is deleted a `StaleDataError` will be raised upon that call to `db_session.commit()`.

I've changed the error handling to take this into account. Now, when this exception is encountered, the database is rolled back, and then a query is made against the database to confirm that this was the cause of the exception (namely, our source has been deleted). If so, an appropriate message is flashed and the journalists is returned to the index view.

fowlslegs:
Revisions complete! All yours @redshiftzero.

fowlslegs:
Rebased.

fowlslegs:
Rebased.

garrettr:
@fowlslegs This looks good to me, but can you write a test for the race condition you fixed in 125149c? If you identified a potential bug (even one that is unlikely to be encountered in normal use, like this one), you should write a test to prevent future regressions. Also, adding such a test would probably fix the decrease in test coverage that is caused by this PR in its current form.

fowlslegs:
Good call @garrettr. That error handling block wasn't in the right place because `get_source()` is called earlier in the request handling, and would return a 404 without even running the `reply()` function. This gave me the opportunity to handle this problem on application-wide (see https://github.com/freedomofpress/securedrop/pull/1720/commits/7a4fe21e17ec1dc4eebf6f1eca900807b73fee42).

I also discovered that an adversary that has access to the JI ATHS URL and HidServAuth value may be able to enumerate Source filesystem identifiers through lookup timing leaks and response generation in the `journalist.get_source()` function. I modified the `journalist.setup_g()` s.t. operations on a user-provided "sid" are only executed for logged in clients (see https://github.com/freedomofpress/securedrop/pull/1720/files#diff-2af57f469c54eb2e19fbb1a0e2a5b35aR91, https://github.com/freedomofpress/securedrop/pull/1720/files#diff-7556fa4cb7c3f25d1b29b08e1ef2e04cR109, and https://github.com/freedomofpress/securedrop/pull/1720/files#diff-7556fa4cb7c3f25d1b29b08e1ef2e04cR123).

Finally, I did write a test for the added logging functionality in `reply()`.

Ready for re-review.

fowlslegs:
Had some thoughts on this. Want to rework it into 3 PRs.

1. Just the `if` block that rejects/ `flash`es on empty replies + the relevant tests.
2. General exception handling for the `reply` function (+ relevant tests).
3. General exception handling for `journalist.get_source()` (+ relevant tests). I think a better idea than what I've done here is _not_ to modify `db.get_one_or_else()`. Instead, `get_source()` should just not use it and do exception handling itself. That way we get to keep using `get_one_or_else()` as it was written and intended to be used in the appropriate places.

Issue #1724, opened 2017-05-22
Add application tests to staging CI


msheiny:
Currently the application tests are running under the `development` CI environment (travis) but not under the `staging` environment (circleCI). Lets get those tests running and fail the build if they fail.

A stretch goal is to spit out results to junit.xml. This allows the CI environment to parse test results and display them with G-U-I power! :boom:

redshiftzero:
Another one that was addressed by #1731, so closing.

PR #1731, opened 2017-05-25
App-tests in CI

### LINKED ISSUE #1729: `app-test` role bombs out against iptables rules
Specifically in the following scenario:

* testing against a staging+ environment
* server has already been rebooted at least once after provision

So far I've caught this issue for the [Download Firefox 46.0.1 for compatibility with Selenium 2.53.6](https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/app-test/tasks/dev_setup_xvfb_for_functional_tests.yml#L24) task. Once the system has rebooted once, the iptables restrictions are in place (which restricts inbound/outbound traffic) and it is unable to pull the download for Firefox. Previously this wasnt an issue because the download location was not an ephermeral space, but for `CI` fixes this was changed to go under `/tmp` so the task will always try to re-download.

msheiny:
## Status

:+1: Ready for review 

## Description of Changes

* Fixes #1729 - `app-test` role failing on second run of playbook against a staging server that has been rebooted.
* Addresses #1724 - Adds application tests into staging CI. Im not :100: % certain how these failing tests will affect CI passing/failing status yet. 
* Added a `tor` and `test` docker instance for utilizing tor in CI to perform checks outside of the `app` and `mon` boxes. In the future, we should plan to throw all tor specific tests that need to be run outside of the SD  CI servers on this `test` instance. The `tor` docker container is merely a socks proxy for other docker containers.
* Fixes a BIG regression with CSS files not getting propagated to the app deb package. Added accompanying testinfra tests to test for this scenario to prevent future problems.

## Testing

Well this is all specifically targeted at CI right now - I would recommend testing locally to confirm that the docker changes work against your machine (specifically you wacky Mac users :man_dancing: ). To recreate locally:

- Source the CI environment variables (make sure you change the vpc-id accordingly - youll get a prompt... youll see what i mean)
```
. ./devops/scripts/local-setup.sh
```
- Ensure your docker environment is ready to go. You GRSEC peeps -- `docker-machine` is your friend.
- Run `make ci-debug` - to prevent CI from tearing down on error
- Run `make ci-go`
- Wait until you see the process chugging along on ansible on the `built` stage (probably like 2 minutes in). Open another window, and jump into the `test` docker container and make sure you can run curl commands via the tor proxy:
```bash
$ docker exec -it test bash
cab015fde045$  curl -s --socks5-hostname \"${TOR_PROXY}\":9050 https://check.torproject.org/ | grep "Congratulations"
```
- If the above command worked tor is working from within the containers, now just wait til the rest of the tests run. XML file should dump into `junit/junit.xml` and you should see about 100 tests with the class prefix `apptest`

:tada: 

## Deployment

N/A

N/A

- [x] Testinfra tests pass on the app-staging CI
- [x] Testinfra tests pass on the mon-staging CI


msheiny:
FYI a last minute change I made allows us to download the raw junit xml files at the end of a run. Its in the webUI under `Artifacts`
selection_024

conorsch:
Rebased to modify the #1729 resolution; old commit was 29d07dd4c8851cc86d806925bce094f3bdae3191, and was sufficient, but not idempotent. Cherry-picked in c63e8b2286804e6271df4cd9224127a56561de8b which was already written to address concerns in #1023.

testinfra/functional/test_tor_interfaces.py
```diff
@@ -0,0 +1,25 @@
...
+
+    # extract onion url from saved onion file
+    onion_url_raw = open(os.path.dirname(__file__)+"/../../install_files/ansible-base/{}".format(site['file']),'ro').read()
+    onion_url = re.search("\w+\.onion", onion_url_raw).group()
+
+    # point curl to our tor proxy box
+    curl_tor = "curl -s --socks5-hostname \"${TOR_PROXY}\":9050 "
+    site_scrape = Command.check_output(curl_tor+onion_url)
+    
+    assert Command.check_output(curl_tor+"-o /dev/null -w \"%{http_code}\" "+onion_url) == "200"
```
conorsch: `curl` command construction is difficult to read over several lines; would prefer a `.format()`-style interpolation.

devops/inventory/staging
```diff
@@ -19,3 +19,6 @@ staging
 
 [build]
 build ansible_connection=docker
+
+[test]
+test ansible_connection=docker
```
conorsch: The `test` host is too vaguely named. At first glance, it looks like a throwaway box that won't actually be used for anything. A more descriptive name such as `apptestclient` would be well met.

redshiftzero:
Looks like there might be some permissions issues:

```
TASK [app-test : Download Firefox 46.0.1 for compatibility with Selenium 2.53.6.] ***
fatal: [app-staging]: FAILED! => {"changed": false, "failed": true, "msg": "Destination /opt not writable"}
```

conorsch:
@redshiftzero The permissions issue you mention was caused by my sloppy cherry-picking; please pull and try again.

redshiftzero:
This is provisioning well for me, all tests pass, the CSS is now back on the app server :metal: and I see the junit XML test output (nice!). I wasn't able to run curl commands via the tor proxy however.. 

```
root@test:/# curl -s --socks5-hostname \"${TOR_PROXY}\":9050 https://check.torproject.org/ | grep "Congratulations"
root@test:/# exit
```

I verified that `$TOR_PROXY` is defined, which it is.

conorsch:
Kicked off a build triggered locally, via the makefile targets, and _everything passed_. Huzzah! Confirmed that the CSS is looking good. I'm going to make a few cosmetic touchups to address the in-line comments I made, then approve. If CI checks continue to pass, we should be good to go.

devops/playbooks/aws-ci-startup.yml
```diff
@@ -8,10 +8,18 @@
...
+    - name: Tor docker proxy
+      docker_container:
+        name: tor
+        image: "hsaito/torbox"
+        state: started
       when: "{{ docker_containers != [] }}"
+      tags: docker
 
-    - name: Start build container
+    - name: Start containers
```
conorsch: Will make these task names more explicit, e.g. `Tor docker proxy` -> `Start Tor docker proxy for app-test environment.`

conorsch:
Quite pleased with these tests, thanks for the rapid turnaround @msheiny! Addressed the inline nits and rebased on top of develop.

msheiny:
Needs a restamp with @redshiftzero 's latest commit .  Feel free to press merge after that, thanks for the thorough review peeps and for cleaning up my work as usual :)

conorsch:
Confirmed that the intentionally failing app test does indeed cause the CI run to fail. That's good. The output on the app test suite is hideous, but I can live with that. Removing the `ignore_errors` task means the playbook exits before the test results are collected, so no artifacts are posted on failing runs (which is when they're most needed):

missing-junit-artifacts

Since we've switch to Ansible v2, a try/except block should allow us to fetch back the results files regardless of error. Will try that now.

msheiny:
> Since we've switch to Ansible v2, a try/except block should allow us to fetch back the results files regardless of error. Will try that now.

So one thing I'm stil curious about is -- will CircleCI report a job as a failure if the junit.xml shows failure but the exit codes are 0 across the board. Thats what I was ultimately trying to reproduce here (but didnt think about the scenario where a test run just totally bombed out). 

I think we'll have to add ensure that test runs dont raise a non-zero exit code AND slap our own logic where  make sure the number of tests is above a certain threshold and if they aren't, raise an error to bubble that up to CI.

msheiny:
heyyy @conorsch -- `c89ac21a` wont solve that issue because the tests arent copied into ~/sd/junit until the last line of [`devops/scripts/ci-tester.sh`](https://github.com/freedomofpress/securedrop/blob/RepairBrokenDebz/devops/scripts/ci-tester.sh#L48). Essentially all the tests can munged together as one big test junit.xml file at the end of the all the test runs. To be honest, I'm not even sure if this behavior is needed anymore in CircleCI 2.0 -- it might be smart enough to parse all the test files as is (but I dunno without testing).

conorsch:
Thanks @msheiny. I misunderstood the failure mode for CI, and it appears we've never had result fetching working in failure, so it's not a regression. I'm fine merging with the ugly test output since it still honestly reports pass/fail for app tests (plus there are a lot of great fixes in this PR).

Will remove the REVERTME commit and repush, then merge on passing.

Issue #1544, opened 2017-01-30
Update all documentation for 0.4 release


heartsucker:
By the looks of things, there are enough major changes that there will need to be significant alterations to the docs. This should include.

# Docs issues for 0.4
- [x] Update source guide screenshots from Tails (#2010)
- [x] Update journalist guide screenshots with ones from Tails (#2011)
- [x] Update admin guide screenshots with ones from Tails (#2012)
- [x] Update "Set up the Network Firewall" Tails screenshots and put the Tails version next to them
- [x] Update "Set up the Network Firewall" section to instruct how to use the template
- [x] Update "Set up the Data Transfer Device" pages and put the Tails version next to them (in #1695)
- [x] Update "Generate the SecureDrop Submission Key" pages and put the Tails version next to them. As this is done, we can also fix #1142.
- [x] Update custom notification screenshot on the "Install SecureDrop" page (#1992)
- [x] Write docs describing EV cert procurement for HTTPS deployment (#1605)

fowlslegs:
Definitely worth waiting until we're at a "stable" point. The UI will never be static, but after we get in all of @ninavizz's suggestions there'll be a lot less churn.

heartsucker:
Agreed. Just putting it here so it could be added to the 0.4 project to ensure it gets done.

fowlslegs:
I think this should also include convergence of the "latest" and "stable" versions of the documentation on our website.

conorsch:
@fowlslegs I wouldn't converge them—"stable" still has merit as a label in that it reflects the most recently _shipped_ release, and "latest" represents whatever the head of current development includes.

edenemmanuel:
Someone who recently set up the latest stable version of SecureDrop emailed the following comments, which I'm attaching to this issue:

1. The data transfer device pages seem to be for another version of tails. I’m running the latest 2.11. At the very least, you should put the version number next to one of the screenshots.
2. The firewall section could be a lot better.  You ship a sample configuration with a bunch of redacted fields. How about simply walking the user through how to change the template and how to load it?  For the screenshots on the pf2440, the one you recommend, it would be nice if you had that info perhaps in a table or in text or something that I don’t have to squint at a low-res jpg to try and figure out what you have. I get that firewalls are complicated, but if somebody buys straight off your list of hardware and puts the computers on a dedicated subnet inside the firewall, should be able to give me a better cookbook approach.
-----
[...]

PR #1740, opened 2017-05-26
Update network firewall templates and docs

### LINKED ISSUE #1122: Firewall template config
The firewall setup instructions are monotonous and fiddly and an easy place to make mistakes. Although FPF can't offer template configs for every firewall out there, since you already recommend pfsense and have pfsense-specific documentation with pfsense screenshots, why not just include a pfsense config file that an admin can edit/upload, saving a whole load of fiddly web-interface-clicking.

Better yet — using a config file might mean that the firewall can be configured over SSH, removing the need for a password and allowing the option of turning off the firewall's web interface altogether. Since the firewall can be physically reset in-person without trouble, this shouldn't represent a huge usability/lockout hurdle.

### LINKED ISSUE #1648: Update external links for firewall documentation
In the current [documentation for the hardware firewall](https://docs.securedrop.org/en/stable/network_firewall.html), we link out to some rather dated pfSense material, such as the wiki, which [hasn't been updated since November 2015](https://doc.pfsense.org/index.php?title=Main_Page&action=history). We should instead consider linking out to the much more user-friendly [NetGate documentation](https://www.netgate.com/docs/sg-2440/index.html). 

In particular, the video on that page for resetting the firewall has come in handy. It also has excellent documentation on [connecting to the console](https://www.netgate.com/docs/sg-2440/connect-to-console.html), for multiple platforms.

redshiftzero:
## Status

Ready for review

## Description of Changes

This PR refreshes the network firewall setup (ticks two more check boxes for #1544). It:

   * Updates the docs and screenshots
   * Fixes old broken links and adds more useful links (how to console in, how to reset the firewall, etc.). This closes #1648.

It also closes #1122, to provide docs and describe how to use the firewall template:

   * Adds a description of setting up the firewall by importing the templates.
   * Replaces the earlier template with several templates corresponding to the aliases, interfaces, and firewall rules sections (these were exported directly from the pfSense firewall without redaction). 
   * The docs still include updated manual instructions on how to configure the firewall in case for some reason a user has trouble with the templates and prefers to just point and click and copy the screenshots. 

## Testing

I've tested this by setting the firewall up manually, exporting the templates, messing up the firewall config (deleting a bunch of rules, changing the interface configs), and then importing the firewall templates one at a time and verifying that the settings return to the correct values. 

Review this for grammar, spelling, inconsistencies etc. now and when we are testing the fresh hardware install of 0.4 test debs we can do a factory reset of the firewall and follow the template instructions to triple check that this is smooth.

## Deployment

Only relevant for new installs

## Checklist

- [x] Doc linting passes locally


docs/network_firewall.rst
```diff
@@ -322,38 +365,133 @@ to configure the OPT1 interface. Go to **Interfaces ▸ OPT1**, and check
...
+
+|Add Firewall Alias|
+
+Click **Save** to add the alias.
+
+Keep adding aliases until the screenshot matches what is shown here:
+
+|Firewall IP Aliases Pre Save|
+
+Finally, hit **Apply Changes**. This will save your changes. You should see a
```
justintroutman: Replace "hit" with "click" for consistency.

docs/network_firewall.rst
```diff
@@ -322,38 +365,133 @@ to configure the OPT1 interface. Go to **Interfaces ▸ OPT1**, and check
...
+
+Scroll down to "Restore Backup" and install each one of the template files for
+**Restore Areas** Interfaces, Aliases, and Firewall Rules, in that order:
+
+- Interfaces config: ``install_files/network_firewall/interfaces-config-pfSense.xml``
+- Aliases: ``install_files/network_firewall/aliases-config-pfSense.xml``
+- Firewall rules: ``install_files/network_firewall/filter-config-pfSense.xml``
+
+Note that none of the template filters are encrypted. Click "Restore Configuration"
+to restore each file. After this is done, verify that the rules have taken
```
justintroutman: I understand the expression "have taken", but maybe "have been restored properly" would be clearer here.

redshiftzero:
Thanks for your review and comments @justintroutman - I've made your suggested changes in c9419dd and added a note of which version of Tails was used to make these screenshots as suggested in #1544. 

Let me know if there are any other comments on this one!

justintroutman:
LGTM!

Issue #1319, opened 2016-06-03
Consider using include statements to DRY out docs


conorsch:
In the upgrade docs for each release, we instruct Admins to pull down the latest release via git, verify the GPG signature on the tag, then checkout out the release branch and proceed with any manual intervention required by the release.

Now that we're using ReadTheDocs (since #1158), we can create a snippet of text to be included dynamically in multiple documents. Think "macro", but without nightmares. Here are some examples of repetitious docs:
- https://docs.securedrop.org/en/latest/set_up_admin_tails.html#verify-the-release-tag
- https://docs.securedrop.org/en/latest/upgrade/0.3.x-to-0.3.5.html#upgrade-procedure
- https://docs.securedrop.org/en/latest/upgrade/0.3.5-to-0.3.6.html#upgrade-procedure
- (pending release 0.3.7)

Repetition is soggy—let's _DRY it out!_ :sunglasses:

We didn't implement a reusable snippet for the 0.3.7 release, but we _did_ reconcile develop docs with master, so we're in a good position to start doing reusing snippets starting in 0.3.8. The tag verification workflow is the most obvious example of redundancy in the docs, but there may be others. Let's track any additional candidates in this issue as well.

zmanian:
I started looking at this but I haven't been able to find any documentation on reusable text snippets in ReStructuredText/ReadTheDocs files. It would be helpful if someone could point at what feature is intended to be used.

redshiftzero:
Sure! Check out [this blog](http://tychoish.com/posts/large-sphinx-deployments/) under "Single Source Content" regarding reStructuredText's `include` directive

conorsch:
Originally recorded in #1208, I'm moving the following examples for places in the docs to use include statements over here:

### Consider using `include` statements to DRY out docs
- [ ] ssh connectivity test:
  - [ ] `servers.rst`
  - [ ] `test_the_installation.rst`
  - [ ] `upgrade_to_tails_2x.rst`
- [ ] ansible-playbook commands
  - [ ] `install.rst`
  - [ ] `backup_and_restore.rst`
  - [ ] `upgrade/0.3.5-to-0.3.6.rst`
  - [ ] `upgrade/0.3pre-to-0.3.x.rst`
  - [ ] `upgrade/0.3.x-to-0.3.5.rst`
- [ ] git clone tasks
  - [ ] `development/getting_started.rst`
  - [ ] `tails_guide.rst`
[...]

PR #1746, opened 2017-05-26
Add "Admin Responsibilities" section to Administrator Guide

### LINKED ISSUE #1727: Add Admin Responsibilities section in Admin Guide
It would be useful to have a clear checklist in the admin guide of what responsibilities SecureDrop admins have, e.g.:

1. Maintain awareness of SecureDrop updates and take any required manual action if requested in the release blog (e.g. run playbooks, update Tails drives).
2. Be informed of and apply updates to your network firewall. If you're using the suggested network firewall by FPF you can subscribe to the RSS feed [here](https://www.netgate.com/feed.xml).
3. Monitor OSSEC alerts for unusual activity and report strange events to FPF through the support portal.
4. Apply security updates to Tails drives as they are released.

...etc.

redshiftzero:
## Status

Ready for review

## Description of Changes

Closes #1727.

  * Adds an admin responsibilities guide to clearly delineate which maintenance tasks are required and expected for SecureDrop admins.

## Testing

Are there responsibilities we want to add here? Are there other useful tips we can be providing?

Otherwise review for spelling, grammar, typesetting issues.

## Checklist

- [x] Docs linting passing locally

### DIFF
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -2,8 +2,8 @@ Administrator Guide
 =====================
 
 You (the administrator) should have your own username and password, plus
-two-factor authentication through either the Google Authenticator app on
-your smartphone or a YubiKey.
+two-factor authentication through either the Google Authenticator app
+on your smartphone or a YubiKey.
 
 .. _Responsibilities:
 
@@ -27,9 +27,9 @@ subscribing to the `SecureDrop RSS Feed`_ to stay apprised of new updates.
 
 Most often, the SecureDrop server will automatically update via apt. However,
 occasionally you will need to run the Ansible playbooks. We will inform you in
-the release blog when this is the case. If you are onboarded to our support portal,
-we will let you know in advance of major releases if manual intervention will be
-required.
+the release blog when this is the case. If you are onboarded to our `SecureDrop
+Support Portal`_, we will let you know in advance of major releases if manual
+intervention will be required.
 
 .. _`SecureDrop Release Blog`: https://securedrop.org/news
 .. _`SecureDrop RSS Feed`: https://securedrop.org/news/feed
@@ -43,9 +43,12 @@ network, you will want to ensure that critical security patches are applied.
 Be informed of potential updates to your network firewall. If you're using the
 suggested network firewall by FPF you can subscribe to the `Netgate RSS Feed`_
 to be alerted when releases occur. If critical security updates need to be
-applied, you can do so through the firewall's pfSense WebGUI.
+applied, you can do so through the firewall's pfSense WebGUI. Refer to our
+:ref:`Keeping pfSense up to date` documentation or the official `pfSense
+Upgrade Docs`_ for further details on how to update the suggested firewall.
 
 .. _`Netgate RSS Feed`: https://www.netgate.com/feed.xml
+.. _`pfSense Upgrade Docs`: https://doc.pfsense.org/index.php/Upgrade_Guide
 
 Keep your Tails Drives Updated
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


docs/admin.rst
```diff
@@ -1,9 +1,71 @@
...
+applied, you can do so through the firewall's pfSense WebGUI.
+
+.. _`Netgate RSS Feed`: https://www.netgate.com/feed.xml
+
+Keep your Tails Drives Updated
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You should apply updates to your Tails drives as they are released, as they
+can contain critical security fixes. Subscribe to the `Tails RSS Feed`_ to be
+alerted of new releases. The online Tails drives once booted and connected to Tor
```
justintroutman: Comma separation might help readability here: "The online Tails drives, once booted and connected to Tor, will alert you if upgrades are available."

redshiftzero:
Thanks for taking a look at this @justintroutman! I made your suggested change in a subsequent commit (which I squashed into the one commit here since it was such a minor change).

docs/admin.rst
```diff
@@ -1,9 +1,71 @@
...
+Keep your SecureDrop Server Updated
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You should maintain awareness of SecureDrop updates and take any required
+manual action if requested in the `SecureDrop Release Blog`_. We recommend
+subscribing to the `SecureDrop RSS Feed`_ to stay apprised of new updates.
+
+Most often, the SecureDrop server will automatically update via apt. However,
+occasionally you will need to run the Ansible playbooks. We will inform you in
+the release blog when this is the case. If you are onboarded to our support portal,
```
fowlslegs: A link to the support portal somewhere in this section might be helpful.

docs/admin.rst
```diff
@@ -1,9 +1,71 @@
...
+we will let you know in advance of major releases if manual intervention will be
+required.
+
+.. _`SecureDrop Release Blog`: https://securedrop.org/news
+.. _`SecureDrop RSS Feed`: https://securedrop.org/news/feed
+
+Keep your Network Firewall Updated
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
fowlslegs: I feel like this section could provide a little more detailed instructions. For older instances, it seems like admins need to manually update then upgrade to the latest 2.3.X release. For newer instances (2.3.0 and beyond), it seems like auto-updates are default, but upgrades must still be done manually. They should probably just consult the instructions below for more explicit instructions on doing so, as they should be better than what we can provide.

https://doc.pfsense.org/index.php/Firmware_Updates
https://doc.pfsense.org/index.php/Upgrade_Guide

...I wonder if there's a way to do auto-upgrades like `cron-apt`. I think most admins are not regularly upgrading their pfSense box. And while I think adding this documentation may increase the amount who do, I don't think it will resolve the problem in the same way automatic upgrades would.

docs/admin.rst
```diff
@@ -1,9 +1,71 @@
 Administrator Guide
 =====================
 
-At this point, you (the administrator) should have your own username and
-password, plus two-factor authentication through either the Google
-Authenticator app on your smartphone or a YubiKey.
+You (the administrator) should have your own username and password, plus
+two-factor authentication through either the Google Authenticator app on
```
fowlslegs: Can you mention FreeOTP here too?

docs/admin.rst
conorsch: Agreed, linking to support portal is appropriate here, as we do it elsewhere in the docs already. Prime candidate for a reusable include as described in #1319, but to keep the wheels turning, let's just copy/paste it: `_SecureDrop Support Portal: https://securedrop-support.readthedocs.io/en/latest/`

redshiftzero:
OK thanks for the review @fowlslegs and @conorsch! I've added those fixes. Let me know if there are any further comments.

PR #1750, opened 2017-05-26
Remove old upgrade guides due to irrelevance and outdated information.

### LINKED ISSUE #1737: Remove old upgrade guides from docs.securedrop.org
We should remove the upgrade guides for older versions (namely, from `0.2.1` through `0.3.7`) of SecureDrop, given that they're largely irrelevant and contain outdated information.

justintroutman:
## Status

Ready for review

## Description of Changes

Removing old upgrade guides, because they're irrelevant and contain outdated information.

## Fixes

This fixes #1737


justintroutman:
Commit 1a73f6b should address these linting errors.

Issue #1689, opened 2017-05-06
Test upgrade path in CI


redshiftzero:
Related to #1681: it would be really great (at a future point) for CI to be catching issues that are introduced that _do not_ break new installs but _do_ break upgrades on existing instances.

msheiny:
So I'm really thinking we need this as part of `0.4` release - we can time-box it but I think its worth trying to bring in. In essence, testing upgrades in particular, is a huge time-drain and its going to hurt more when we try to particularly test nuances of the jump between ansible versions and tails versions.

(taking some of this convo from chat)
I see testing strategy for this issue as a two part problem:
1. ensure users are able to upgrade tails from 2 -> 3 without breaking their ability to reach their servers and perform upgrades. 
2. ensure we do not break the server when running a new playbook against a `0.3.12` server

Re: `1` -  Client-side testing for this scenario is going to be a pain in the ass to try and automate. At least as the docs currently describe it (creating a backup tails stick, upgrading from 2 -> 3, testing web/ssh/ansible access still exists). We can do particularly pieces of it but overall i think this is best left as a manual check in QA. 
Re: `2` -- This should be much easier to do but still has a few hairy moving pieces. High-level workflow I see:
* Create an image of a running `0.3.12` server (staging) - remove tor auth keys (so there isnt a conflict)
* Clone off the image during CI run 
* Resprinkle tor tor auth pieces
* Spin up ephemeral apt server and point securedrop CI server here
* Perform upgrade and record results

Im concerned about the length of time that CI takes if we just tack this on to our current staging run. We'd have to play it smart and only run on particular branches depending on how much additional time it adds on.

msheiny:
![index](https://user-images.githubusercontent.com/1727935/36178648-797bb3f0-1111-11e8-8e28-79c9978a146f.png)

msheiny:
Okay so it sounds like we are going to pivot this ticket slightly and aim for running under nested virtualizaiton. Going to break this into two tickets:

* The actual upgrade automation testing logic to be able to run under libvirt/kvm --> #3018
* The work to get that ^ running in CI with nested virtualization --> this ticket.

msheiny:
As a new direction for this ticket and to scope the work, I propose the following high-level goals:

- [ ] Ensure that we can in-fact do nested virtualization in Google Cloud as advertised w/ libvirt + virtualbox. Discovering this early will totally shape the work of the rest of the ticket. - estimate 2 hr
[...]

PR #1762, opened 2017-05-31
Deprecate `prod-specific.yml` vars file

### LINKED ISSUE #1758: Migrate away from `prod-specific.yml`
The current strategy for configuring SecureDrop at install time is to fill out the `prod-specific.yml` vars file prior to running the playbooks. This works well enough for the first-run, but quickly becomes problematic during upgrades, since that file is version-controlled, and yet expected to contain site-specific customizations. Thus we recommend a tedious [stash-pull-pop workflow](https://docs.securedrop.org/en/stable/upgrade/0.3.6-to-0.3.7.html#upgrade-procedure) to allow upstream changes from this repo to land on Admin Workstations without clobbering site-specific customizations.

In order to provide a more customizable experience in general (#1129), and to support improvements for 0.4 (#1605, #1028, #749) specifically, we should abandon the current `prod-specific.yml` vars file in favor of a more durable approach to customizing the installation. 

We can add a new vars file target that is a non-existent file by default, but will automatically be included when it exists. A helper script or playbook can prompt the Admin for required custom vars, then handle writing the vars to disk in YAML format. Existing instances should have the contents of the `prod-specific.yml` vars file migrated over to the new location. We'll of course need to update the documentation with the new strategy, and possibly write an upgrade guide for Admins for 0.4.

conorsch:
## Status

Ready for review.

## Description of Changes

Fixes #1758. Migrates away from the long-standing `prod-specific.yml` vars file, used for providing site-specific variables and optional customizations. The new file is called `group_vars/all/site-specific`. Includes a new dedicated playbook called `securedrop-configure.yml` that interactively prompts for site-specific secrets, validates them via the `validate` role, then writes them to disk, in a non-version-controlled file.

Admins who previously filled out the `prod-specific.yml` playbook need not take any special action. The first time the `securedrop-prod.yml` playbook is run after upgrading to 0.4, the old vars file will be automatically copied to the new location, and its contents read from there for the remainder of the playbook.

Also updated the docs throughout, to use the new filename where appropriate, and also to reference the new playbook in the install docs. Avoided expanding the validation logic, since we have separate issues for those (e.g. #749).

### Scenario 1: upgrades
1. Check out `0.3.12` and fill out the `prod-specific.yml` file with production-like secrets.
2. Provision prod VMs.
3. Check out this feature branch and reprovision the prod VMs.
4. Confirm that the second provisioning run completed successfully (no undefined vars anywhere).
5. Confirm that the `group_vars/all/site-specific` file was created and its values match those in the `prod-specific.yml` file you manually edited previously.

### Scenario 2: new installs
1. Check out this feature branch, and ensure you have a clean repo (no customizations to `prod-specific.yml`).
2. Hop into the `install_files/ansible-base` directory and run the new configuration playbook: `ansible-playbook securedrop-configure.yml`. 
3. Fill out all the prompts, with the same secrets you used in Scenario 1.
4. Confirm no errors with the validation prompts (playbook will fail if errors are found).
5. Confirm values in `group_vars/all/site-specific` are formatted correctly.
6. Confirm you can provision VMs with the new vars file.

## Deployment

Ensured that existing Admins will not need to take special action to migrate the vars—the new prod playbook will handle that for them. The big win is for first-time Admins on the initial install: the combination of interactive prompts and built-in variable validation should provide bumpers enough that misconfigurations do not slip in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,9 @@
 securedrop-app-conf.yml
 securedrop-mon-conf.yml
 
+# ignore site-specific vars file, since we can't predict contents
+install_files/ansible-base/group_vars/all/site-specific
+
 # ignore securedrop-app-code wheelhouse archive
 wheelhouse
 
--- a/docs/backup_and_restore.rst
+++ b/docs/backup_and_restore.rst
@@ -79,7 +79,7 @@ the SecureDrop servers.
    ansible -i inventory -u <SSH username> -m ping all
 
 .. tip:: If you forgot your SSH username, it is the value of the ``ssh_users``
-         variable in ``prod-specific.yml``.
+         variable in ``group_vars/all/site-specific``.
 
 If this command fails (usually with an error like "SSH Error: data could not be
 sent to the remote host. Make sure this host can be reached over ssh"), you need
--- a/docs/development/virtual_environments.rst
+++ b/docs/development/virtual_environments.rst
@@ -66,7 +66,7 @@ mon-prod
     This is like a production installation with all of the system
     hardening active, but virtualized, rather than running on hardware.
     You will need to configure prod-like secrets in
-    ``install_files/ansible-base/prod-specific.yml``, or export
+    ``install_files/ansible-base/group_vars/all/site-specific``, or export
     ``ANSIBLE_ARGS=="--skip-tags validate"`` to skip the tasks
     that prevent the prod playbook from running with Vagrant-specific info.
 
@@ -183,7 +183,7 @@ Prod
 ----
 
 You will need to fill out the production configuration file at
-``install_files/ansible-base/prod-specific.yml`` with custom secrets.
+``install_files/ansible-base/group_vars/all/site-specific`` with custom secrets.
 The production playbook validates that staging values are not used in
 production. One of the values it verifies is that the user Ansible runs as is
 not ``vagrant`` To be able to run this playbook in a virtualized environment
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -16,6 +16,8 @@ Tor. Once that's done, you can install Ansible:
     sudo apt-get update
     sudo apt-get install ansible
 
+.. _configure_securedrop:
+
 Configure the Installation
 --------------------------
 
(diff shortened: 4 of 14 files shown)


msheiny:
Hey @conorsch -- I have some high-level concerns regarding this approach and wanted to start a dialogue :)

As it currently is laid out, I think `ansible` is not going to be the best tool for validating and taking in variables. Here is an example of the workflow I just stepped through... 

* Executed script
* Made my way through each prompt looking up IP values, fingerprints, and filepaths, smtp values and carefully entering those in.
* At the end, as a new user you see the ansible output (which is kind of jarring for users not accustomed to that) and I get the following "mysterious" dialogue which dumps me back out to prompt.

So yeah, I'm pretty sure thats a bug since the current `ansible_user` is going to be `amnesia` when the script is run on tails ... but anyways the point was that ansible has jarring output here and then ontop of that the user has to go back and enter everything over again if any item was wrong. As a short-term work-around I would at the very least throw the `validate` role after the `Save site-specific information as local vars file` task so whatever the user enters gets thrown into the file but then throws an error at the end.

As a frequent ansible user, this type of output makes total sense to me but will be jarring and cryptic to a new user.
selection_025

msheiny:
I also have a question outside the scope of this ticket? What is the point of asking the user for the fingerprint along with public key file? I'm assuming its a validation thing? Can't we just read in the file, validate its a public key, and spit out the key details and ask the user to confirm they dropped the right key?

msheiny:
Okay I didnt want to start pushing commits on this until I got your feedback... anywhoo to fix some issues I  added the following changes. Note that I had to skip the user check when running under Tails in validate and I also moved the validate role to AFTER the var_prompt section.

```
diff --git a/install_files/ansible-base/roles/validate/tasks/validate_users.yml b/install_files/ansible-base/roles/validate/tasks/validate_users.yml
index 0cc46a97..2a32852e 100644
--- a/install_files/ansible-base/roles/validate/tasks/validate_users.yml
+++ b/install_files/ansible-base/roles/validate/tasks/validate_users.yml
@@ -9,6 +9,7 @@
       - "ssh_users != ''"
       - "item not in ssh_users"
   with_items: "{{ securedrop_validate_disallowed_users }}"
+  when: "'Tails' not in ansible_lsb.id"

 # Beyond checking what's declared in the vars file, we must validate the
 # username used to connect to the host. The SSH config forbids SSH connections,
@@ -20,3 +21,4 @@
       - "ansible_ssh_user|default('') != item"
       - "remote_user|default('') != item"
   with_items: "{{ securedrop_validate_disallowed_users }}"
+  when: "'Tails' not in ansible_lsb.id"
diff --git a/install_files/ansible-base/securedrop-configure.yml b/install_files/ansible-base/securedrop-configure.yml
index 2682e803..05eb7e27 100644
--- a/install_files/ansible-base/securedrop-configure.yml
+++ b/install_files/ansible-base/securedrop-configure.yml
@@ -105,10 +105,6 @@
         msg: >-
           Validating user-entered variables...

-  roles:
-    - role: validate
-      tags: validate
-
   tasks:
     - name: Save site-specific information as local vars file.
       lineinfile:
@@ -156,3 +152,11 @@
           var_value: "{{ sasl_username }}"
         - var_name: sasl_password
           var_value: "{{ sasl_password }}"
+
+- name: Validate site-specific information.
+  hosts: localhost
+  connection: local
+  gather_facts: yes
+  roles:
+    - role: validate
+      tags: validate
```

conorsch:
@msheiny Great review. I agree that we should save any variables to disk even if the validation fails, so that we can resume from that state if necessary. Doing so would certainly go a long way in reducing Admin frustration. As to your point about the specific validation failures, e.g. `ansible_user` not working in Tails, those refinements are best handled in #749, which can drill down hard on the validation logic.

In the interest of avoiding feature creep, I'm inclined to stick with the `ansible-playbook` invocation for the 0.4 release, although I remain open to more durable scripting thereafter. Once big usability improvement would be judicious use of the `msg` parameter in the `assert` tasks, to provide an informative message about the failure, which is now possible since we've moved to Ansible v2. Again, that's best handled in #749, though.

One thought about having a wrapper script is that passing extra vars to a playbook with `vars_prompt` will skip prompts for vars already defined. So to _resume_ the configure playbook:

```
ansible-playbook securedrop-configure.yml -e @group_vars/all/site-specific
```

The logic around when/whether to provide that auto-resume capability is subject to debate, and I worry would complicate the scaffolding provided for 0.4. Will tinker with that idea and circle back here, otherwise I'm comfortable considering for review the playbook-based strategy, albeit with @msheiny's suggestion of don't-forget-to-write-vars.

conorsch:
Created a very lightweight wrapper script called `configure` that calls the playbook. Using the `./configure` invocation allows auto-resume of inputs, so that vars already specified in the vars file are not asked for again. If the Admin wishes to completely reconfigure the instance, then `./configure --force` will prompt for everything.

The wrapper script is Bash because a Python implementation would be more complicated and require thorough review. Willing to revise the script in subsequent PRs, likely post-0.4, and having a `./configure` file leaves the door open to substantial changes that would be transparent to the Admin. 

Omitted the changes you posted about not validating SSH usernames under Tails, based on value of `ansible_lsb.id`—those types of changes are best handled in the resolution for #749, and I'll improve the associated messaging as well.

msheiny:
@conorsch this is becoming a huge time-suck to test for me. That's not a jab at you ( :heart: you buddy ) , thats a challenge with our current tails architecture. I'm increasing feeling we need to handle upgrade testing in an automated way as part of #1689 in the `0.4` QA process. Doing it manually is a huge pain in the ass.

You indicated up top that I need to test this with the `prod` playbook but we determined in chat that well, thats not realistic - because prod is pulling in production deb packages which dont play nicely with our latest apache config settings in ansible (an apparmor conflict). So now I'm running with a staging level playbook but making a bunch of modifications to the inventory and tags to `mock` a production. I also dont have any original tails 2 sticks left so im trying to pip install the previous ansible version to mock an original tails environment and hitting compilation errors. tl;dr - im burning a lot of time trying to mimic an upgrade test and think this is better spent punting to QA. Highly advocating that we automate that process but thats a discussion best left in #1689 .

I've successfully tested scenario #2 though - so I'm going to green light this PR for now because well, I think between the team we will tackle remaining issues in some upcoming PRs that are in the queue.

msheiny:
Oh btw -- kicked off CI manually again ... not sure why that failed :confused:

conorsch:
> You indicated up top that I need to test this with the `prod` playbook but we determined in chat that well, thats not realistic - because prod is pulling in production deb packages which dont play nicely with our latest apache config settings in ansible (an apparmor conflict).

You're right, that was misleading on my part. Staging is definitely the way to go here, due to the template conflicts you identify. It's possible to simply reuse the generate vars file for prod for the staging hosts, but that's clunky.

Thanks for the thorough review here—I agree there will continue to be refinements as we pull in resolutions to #749 and #1146, then proceed with vigorous QA prior to release.

Issue #1728, opened 2017-05-24
Possible race condition when two Journalists Reply to same Source simultaneously


psivesely:
There may be a race condition in `db.Journalist.reply()`. There are potentially many unexpected outcomes of two Journalists sending a reply to the same source at the same time, and I believe I'm only part of the way to narrowing down the most likely ones. Nonetheless, I'll share my initial determinations and concerns.

Here's the important bit of `db.Journalist.reply()`:

```py
    g.source.interaction_count += 1
    filename = "{0}-{1}-reply.gpg".format(g.source.interaction_count,
                                          g.source.journalist_filename)
    crypto_util.encrypt(msg,
                        [crypto_util.getkey(g.sid), config.JOURNALIST_KEY],
                        output=store.path(g.sid, filename))
    reply = Reply(g.user, g.source, filename)

    try:
        db_session.add(reply)
        db_session.commit()
```

Also important, the function used to assign `g.source`:

```py
def get_source(sid):
    """Return a Source object, representing the database row, for the source
    with id `sid`"""
    source = None
    query = Source.query.filter(Source.filesystem_id == sid)
    source = get_one_or_else(query, app.logger, abort)

    return source
```

So the worry is that the thread-local `g.source` object is the same for both request threads in the case where the second thread executes `get_source()` before the first executes `commit()`.

It might be easiest to break down possible outcomes by operation. I confirmed experimentally (`ipython` + tmux's `synchronize-panes` feature) that you can bring a source into your session, inc the `interaction_count` and `commit()` in two interpreters simultaneously and no errors occur. So in such a situation I believe we can assume `interaction_count` will be increased just by one despite two replies, and 

The database itself will not be updated until `commit()` is called, so both threads can update the `source.interaction_count`, but instead of increasing by two, an increase of only one happens _without exception_.

Next, let's look at `crypto_util.encrypt()`. Since this the only I/O operation in between execution of our functions/methods of interest, it's almost certainly the most time consuming and highly likely to be a point of overlap (i.e., both threads are statistically most likely to have some overlap in their execution of this procedure). Since `interaction_count` increases are not synced between the threads here, they will both be the same, so `filename` will be the same. This means both replies will be written to the same file. Looking at `gnupg._meta._encrypt`, we see how this is handled:

```py
        if output:
            if getattr(output, 'fileno', None) is not None:
                ## avoid overwrite confirmation message
                if getattr(output, 'name', None) is not None:
                    output_filename = output.name
[...]

PR #1768, opened 2017-05-31
Adds exception handling to journalist.reply()

fowlslegs:
## Status

Ready for review.

## Description of Changes

All changes target `journalist.reply()`:

* Adds error handling to database transactions to handle unlikely, but possible errors resulting from racing threads modifying the same database object. Error handling involves rollback of the session, logging that tries to straddle the line between usefulness and not writing sensitive data to disk, and flashing a generic error message on the client to let them know the operation was not successful.
  * Adds a unit test that ensures that error handling is working as described.
  * Adds to the docstring to describe the added functionality.

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -578,7 +578,16 @@ def download_single_submission(sid, fn):
 def reply():
     """Attempt to send a Reply from a Journalist to a Source. Empty
     messages are rejected, and an informative error message is flashed
-    on the client.
+    on the client. In the case of unexpected errors involving database
+    transactions (potentially caused by racing request threads that
+    modify the same the database object) logging is done in such a way
+    so as not to write potentially sensitive information to disk, and a
+    generic error message is flashed on the client.
+
+    Returns:
+       flask.Response: The user is redirected to the same Source
+           collection view, regardless if the Reply is created
+           successfully.
     """
     msg = request.form['msg']
     # Reject empty replies
@@ -593,11 +602,24 @@ def reply():
                         [crypto_util.getkey(g.sid), config.JOURNALIST_KEY],
                         output=store.path(g.sid, filename))
     reply = Reply(g.user, g.source, filename)
-    db_session.add(reply)
-    db_session.commit()
 
-    flash("Thanks! Your reply has been stored.", "notification")
-    return redirect(url_for('col', sid=g.sid))
+    try:
+        db_session.add(reply)
+        db_session.commit()
+    except Exception as exc:
+        flash("An unexpected error occurred! Please check the application "
+              "logs or inform your adminstrator.", "error")
+        # We take a cautious approach to logging here because we're dealing
+        # with responses to sources. It's possible the exception message could
+        # contain information we don't want to write to disk.
+        app.logger.error(
+            "Reply from '{}' (id {}) failed: {}!".format(g.user.username,
+                                                         g.user.id,
+                                                         exc.__class__))
+    else:
+        flash("Thanks! Your reply has been stored.", "notification")
+    finally:
+        return redirect(url_for('col', sid=g.sid))
 
 
 @app.route('/regenerate-code', methods=('POST',))
--- a/securedrop/tests/test_journalist.py
+++ b/securedrop/tests/test_journalist.py
@@ -9,6 +9,7 @@ import zipfile
 from flask import url_for, escape
 from flask_testing import TestCase
 from mock import patch, ANY, MagicMock
+from sqlalchemy.orm.exc import StaleDataError
 
 os.environ['SECUREDROP_ENV'] = 'test'
 import config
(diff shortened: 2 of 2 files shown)


securedrop/journalist.py
```diff
@@ -593,11 +602,25 @@ def reply():
...
+        flash("An unexpected error occurred! Please check the application "
+              "logs or inform your adminstrator.", "error")
+        # We take a cautious approach to logging here because we're dealing
+        # with responses to sources. It's possible the exception message could
+        # contain information we don't want to write to disk.
+        app.logger.error(
+            "Reply from '{}' (id {}) failed: {}!".format(g.user.username,
+                                                         g.user.id,
+                                                         exc.__class__))
+        db_session.rollback()
```
garrettr: @fowlslegs I don't think you need to call `.rollback()` explicitly. If an exception is raised before or during `.commit()`, then the transaction will have completed, and `.rollback()` will be called automatically when the scoped_session is removed in `shutdown_session`. For details, see: http://flask.pocoo.org/docs/0.12/patterns/sqlalchemy/ and http://docs.sqlalchemy.org/en/latest/orm/contextual.html.

securedrop/tests/test_journalist.py
```diff
@@ -41,6 +42,26 @@ def setUp(self):
...
+                   side_effect=exception_class(exception_msg)):
+            resp = self.client.post(url_for('reply'),
+                                    data={'sid': sid, 'msg': '_'})
+
+        # Notice the "potentially sensitive" exception_msg is not present in
+        # the log event.
+        mocked_error_logger.assert_called_once_with(
+            "Reply from '{}' (id {}) failed: {}!".format(self.user.username,
+                                                         self.user.id,
+                                                         exception_class))
```
garrettr: Should there be a test for the new variety of flashed message that a journalist could observe if a reply submission results in an error?

securedrop/journalist.py
fowlslegs: You're right, I just picked up this practice because I saw Jen started adding this line to exception handling blocks. Bad cargo culting on my part.

securedrop/tests/test_journalist.py
fowlslegs: Yeah, I think there was in the PR that was split up into multiple superceding PRs (including this one). Must have gotten lost in that process. Good catch.

securedrop/tests/test_journalist.py
```diff
@@ -41,6 +42,43 @@ def setUp(self):
...
+
+    @patch('journalist.app.logger.error')
+    def test_reply_error_flashed_message(self, mocked_error_logger):
```
garrettr: Don't mock `journalist.app.logger.error` if you aren't going to use the mock anywhere

fowlslegs: Fixed.

fowlslegs:
Changes made, commits squashed, rebased on develop.

garrettr:
@redshiftzero Do you want to review this? I think it's good to go, but since I know you've done a fair amount of work with SQLAlchemy and the database code I thought your perspective might be valuable here.

securedrop/journalist.py
```diff
@@ -593,11 +602,24 @@ def reply():
...
+        # with responses to sources. It's possible the exception message could
+        # contain information we don't want to write to disk.
+        app.logger.error(
+            "Reply from '{}' (id {}) failed: {}!".format(g.user.username,
+                                                         g.user.id,
+                                                         exc.__class__))
+    else:
+        flash("Thanks! Your reply has been stored.", "notification")
+    finally:
+        return redirect(url_for('col', sid=g.sid))
```
redshiftzero: Question: is this PR intended to handle the case where the journalist tries to reply to a recently deleted source? It looks like from the use of the `StaleDataError` exception in the unit tests that the answer is yes - but in this case, this redirect to `url_for('col', sid=g.sid)`, will 404 if the source was deleted, so the flashed message above will not be displayed, no?

fowlslegs: > Question: is this PR intended to handle the case where the journalist tries to reply to a recently deleted source?

Yes and no. It's intended to handle the race condition where the source is deleted after `journalist.setup_g()` has finished executing in the request thread, but before `db_session.commit()` is called in `journalist.reply()`. I use the base `Exception` class because I also believe it's possible exceptions other than `StaleDataError` made be raised under certain race-condition-related circumstances (for more see https://github.com/freedomofpress/securedrop/issues/1728).

> will 404 if the source was deleted, so the flashed message above will not be displayed, no?

As explained in https://github.com/freedomofpress/securedrop/pull/1720#issuecomment-305307167, the case in which the source is deleted before the request handling starts will be taken care of in a separate PR. There is a rough WIP branch `gracefully-handle-posts-involving-deleted-sources` (note: am now implementing error handling for _all_ types of requests, not just POSTS).

redshiftzero: As we just discussed f2f, this 404 is occurring in the `finally` block, and any case where the source is deleted won't be gracefully handled. To handle this situation one could delete this `finally` block and instead:

1. Redirect to the home page if the source is deleted so that a flashed message is displayed and no 404 occurs. Nice advantage: one could also then flash a message specific to this situation ("Sorry, the source was deleted by another user."). You can simply write a unit test for the first case just by taking one of the tests below and adding a line where you delete the source.
2. Else you redirect to the `url_for('col', sid=g.sid)` and flash the message as you are currently doing. And you've already got the unit tests for this written :+1:

Issue #1617, opened 2017-03-11
Harden Python package download process


psivesely:
`pip` does not encourage use of or check Python package signatures. That doesn't mean we can't do better than just running `pip wheel` in the `build_securedrop_app_code_deb` role. Here are a few things we could do to harden the package download process:

- [ ] Check signatures on the packages that provide them (last time I checked a while back 4/19 upstream projects signed their packages).
- [x] Pin SHA256sums of packages. This is something `pip` supports enforcing via the `requirements.txt` file.
- [ ] Pin PyPi's cert. It hasn't changed in years.
- [ ] Use Tor.
- [ ] Try to reduce the number of dependencies SD uses. There are numerous packages we use that have barely if at all changed in years, and that we only use tiny snippets from.
- [ ] File issues with upstream package developers encouraging them to sign their packages.

legoktm:
I'll bring this up as an agenda item in the next server development hangout (if we have time), my current feeling is that there's nothing actionable here besides the work being done for reproducible wheels (https://github.com/freedomofpress/securedrop/issues/5901).

Going through the list:
* Check signatures on the packages that provide them - instead of checking signatures, we're auditing the code itself through diff reviews
* Pin SHA256sums of packages - done.
* Pin PyPi's cert - dubious value IMO, SHA256sums prevent us from getting compromised packages. And we're moving in the direction of cutting pypi out of the build process by hosting our own wheels.
* Use Tor - doesn't help with hardening the package download process?
* Try to reduce the number of dependencies SD uses - a very good goal that we're slowly making progress on, but is better handled in a separate or more scoped ticket.
* File issues with upstream package developers encouraging them to sign their packages - see above, we're not using package signatures anymore.

legoktm:
No objections in today's hangout, so closing.

PR #1770, opened 2017-06-01
Update Python dependencies 0.4

### LINKED ISSUE #1769: Update Python dependencies for the 0.4 release

fowlslegs:
## Status

Ready for review.

## Description of Changes

Fixes #1769.


securedrop/requirements/update_python_dependencies
```diff
@@ -0,0 +1,46 @@
...
+
+pushd "${WORKING_DIR}"
+# Create a temporary virtualenv for the SecureDrop Python packages in our
+# requirements directory and make sure it's cleaned up on exit.
+trap 'rm -rf "${VENV}" && popd' EXIT
+
+virtualenv -p python2 "${VENV}"
+# shellcheck disable=SC1090
+source "${VENV}"/bin/activate
+pip install -U pip-tools
```
garrettr: Wondering if we really need to manage creating and destroying a virtualenv just to install pip-tools. We are already require the developer running this script to install pip and virtualenv before they can run it; is it too much more to ask them to install pip-tools as well? Is the separate virtualenv required for `pip-compile`?

garrettr: Just tested running `pip-compile` in a new virtualenv—it does not install dependencies into the virtualenv, which I think is further evidence that we don't need it.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/defaults/main.yml
```diff
@@ -14,5 +14,6 @@ securedrop_app_rsync_opts:
   - "--exclude=*.git"
   - "--exclude=*aths"
   - "--exclude=*.deb"
+  - "--exclude=*update_python_requirements"
```
redshiftzero: Should this be `update_python_dependencies`?

redshiftzero: Should this be `update_python_dependencies`?

fowlslegs: This should be omitted entirely based on the reworking of this role to use the `synchronize` module's `.rsync-filter` file support.

securedrop/requirements/update_python_dependencies
fowlslegs: 1. This script runs faster than you can type the commands yourself. The fact we're creating and destroying a virtual environment is a small performance penalty compared to the procedural reproducibility it provides.
2. This code is highly simple and stable. The maintenance burden is pretty much negligible.
3. Scripting the dependency update process codifies the process, which is a way of both documenting the process and providing reproducible results.
4. This script will play an integral process in resolving #1617. Especially once #1617 is implemented, being able to simply run this script will be much easier than trying to remember the `pip-compile` switch which pins hashes in the generated `.txt` file, or the one which would let you pin the PyPi cert for connecting to that  (or copying commands from documentation).

> Just tested running pip-compile in a new virtualenv—it does not install dependencies into the virtualenv, which I think is further evidence that we don't need it.

This is the expected and desired behavior. Developers should maintain a persistent virtual environment that contains the `develop-requirements.txt` dependencies as described in the documentation. This script is just used (/ will just be used) to provide a safe, reproducible dependency update mechanism, and should only be run in practice by a single developer once every few weeks to months.

garrettr: I feel like you misunderstood my review comment. I'm not saying we should do away with `update_python_dependencies`—I think it's great. I just think it might be kind of unnecessary (and a waste of time, over repeated runs) to create (then destroy) a new virtualenv *just* to install pip-tools every time this script is run.

`pip-tools` is already included in `develop-requirements.in`. Couldn't we just update the docs to say that `update_python_dependencies` has to be run from an environment where those requirements are installed (e.g. as we already do in https://docs.securedrop.org/en/latest/development/documentation_guidelines.html), and then simply the script a bit?

I just really don't see the point of creating a virtualenv, installing pip-tools, then destroying it again. Is there some use case I am missing?

fowlslegs: This provides a simple, reliable, and relatively fast way to ensure we have the latest `pip` and `pip-tools`. Further, it doesn't make any modifications to user-created virtual environments nor does it make changes that may move any of our virtual machines further away from our production configuration. Having the latest versions installed is important because (i) it will allow us to do the most with #1617 and (ii) it might avoid problems between older version of these packages and PyPi or the options we specify to pip in this script, among other potential incompatibilities.

Basically, the time expenditure is very little (an extra few seconds) and repeated runs of the script are not realistic in practice.

securedrop/requirements/update_python_dependencies
```diff
@@ -0,0 +1,48 @@
...
+# does not work reliably, so instead we quit with an error message.
+if [[ -n "${VIRTUAL_ENV}" ]]; then
+  printf '\nERROR: Please deactivate your virtualenv then re-run this script.\n'
+  exit 1
+fi
+
+# If pip and virtualenv are not available, try to install them with apt-get. If
+# apt-get is not available, quit with an error message.
+if ! command -v pip virtualenv > /dev/null; then
+  if command -v sudo apt-get > /dev/null; then
```
garrettr: You are using `command` incorrectly here. According to the [docs](https://www.gnu.org/software/bash/manual/html_node/Bash-Builtins.html#Bash-Builtins), the usage is:

```
command [-pVv] command [arguments …]
```

That is, it only takes one command at a time. As a consequence, you cannot pass multiple arguments and rely on the return code to check for their existence. Here's an example Bash session:

```
vagrant@app-staging:~$ command -v echo
echo
vagrant@app-staging:~$ echo $?
0
vagrant@app-staging:~$ command -v notinstalled
vagrant@app-staging:~$ echo $?
1
vagrant@app-staging:~$ command -v echo notinstalled
echo
vagrant@app-staging:~$ echo $?
0
```

securedrop/requirements/update_python_dependencies
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+
+# Usage: ./update-python-dependencies
```
redshiftzero: Since you're changing the name of `pip-update.sh`, these docs still need updating: https://docs.securedrop.org/en/latest/development/tips_and_tricks.html#upgrading-or-adding-python-dependencies

testinfra/build/test_securedrop_deb_package.py
```diff
@@ -119,6 +119,20 @@ def test_deb_package_control_fields_homepage(File, Command, deb):
...
 @pytest.mark.parametrize("deb", deb_packages)
+def test_deb_package_contains_no_update_dependencies_file(File, Command, deb):
+    """
+    Ensures the update_python_dependencies script is not shipped via the
+    Debian packages.
+    """
+    deb_package = File(deb.format(
+        securedrop_test_vars.securedrop_version))
+    # Using `dpkg-deb` but `lintian --tag package-installs-python-bytecode`
+    # would be cleaner. Will defer to adding lintian tests later.
```
redshiftzero: Is this comment re: `lintian --tag package-installs-python-bytecode` relevant here? It looks like this is leftover from copypasta-ing the `test_deb_package_contains_no_pyc_files` test

redshiftzero:
Also note that the staging CI failures here are due to the versions in `pip_deps` in `testinfra/vars` needing to be updated

securedrop/requirements/update_python_dependencies
fowlslegs: https://docs.securedrop.org/en/latest is down. The stable branch is still up.

securedrop/requirements/update_python_dependencies
fowlslegs: Fixed.

securedrop/requirements/update_python_dependencies
fowlslegs: @redshiftzero I might have a better workflow than the one described in that discussion. We should talk IRL about it.

testinfra/build/test_securedrop_deb_package.py
fowlslegs: It's on all tests that use `dpkg-deb --contents {}.format(deb_package.path)`. Kept it so that this test wouldn't be missed for upgrade if/when that suggested change happens. Maybe it's better to track these as an issue rather than `TODO` comments as suggested in similar cases, but that's out of scope of this PR to change.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/defaults/main.yml
fowlslegs: Fixed.

securedrop/requirements/update_python_dependencies
fowlslegs: Fixed.

fowlslegs:
> Also note that the staging CI failures here are due to the versions in pip_deps in testinfra/vars/app-staging.yml needing to be updated

Fixed. 

... Addressed all issues, rebased, and squashed extra commits. Ready for re-review.

securedrop/requirements/update_python_dependencies
garrettr: @fowlslegs I'm looking into https://docs.securedrop.org/en/latest being down. FWIW, I notice that https://securedrop.readthedocs.io/en/latest/ is also down, so I think the problem is with Read the Docs, not our DNS.

garrettr: I just manually triggered a rebuild of the "latest" docs from the Read the Docs administrative interface, and https://docs.securedrop.org/en/latest is back. Not sure what went wrong there. Skimmed the build logs but don't see anything unusual. We should keep an eye on it in case it happens again.

PR #1776, opened 2017-06-06
Add MAX_CONTENT_LENGTH to Flask config for development environment

### LINKED ISSUE #1714: File upload limit does not work in the development environment
# Steps to reproduce

1. Provision development VM
2. Go to source interface
3. Generate a codename
4. Submit a >0.5 GB file

# What should happen

A flashed message should appear indicating that the file is too large for submission.

# What actually happens

The file uploads.

# Comment

Whoever fixes this ideally should definitely write a functional or unit test to guard against future regressions.

redshiftzero:
## Status

Ready for review

## Description of Changes

* Adds `MAX_CONTENT_LENGTH` to dev Flask config to fix divergent behavior in the development environment, see analysis by @garrettr in #1714. Closes #1714.

## Testing

1. Provision development
2. Try to upload a file larger than 500 MB
3. You should see: 

screen shot 2017-06-05 at 7 53 46 pm

## Deployment

No concerns for deployment as this only effects the development environment. On new installs, this line will exist in their `config.py` but will not be executed as it's in the dev block. And on existing installs, this change is only in `config.py.example`, and so this code will not be present in the `config.py` at all.

### DIFF
--- a/securedrop/config.py.example
+++ b/securedrop/config.py.example
@@ -55,6 +55,9 @@ if env == 'prod':
 elif env == 'dev':
     # Enable Flask's debugger for development
     FlaskConfig.DEBUG = True
+    # Use MAX_CONTENT_LENGTH to mimic the behavior of Apache's LimitRequestBody
+    # in the development environment. See #1714.
+    FlaskConfig.MAX_CONTENT_LENGTH = 524288000
 elif env == 'test':
     FlaskConfig.TESTING = True
     # Disable CSRF checks to make writing tests easier


garrettr:
This definitely works, although I'd prefer to store the value of the maximum upload size in a single canonical location to prevent potentially confusing divergences between the development and staging/prod environments in the future. @conorsch It would be pretty easy to make the maximum upload size an Ansible var and template the relevant configuration files, correct?

conorsch:
> It would be pretty easy to make the maximum upload size an Ansible var and template the relevant configuration files, correct?

Agreed, will take a look!

conorsch:
I take it back: declaring the var once is difficult, due to the `lineinfile`-strategy used to update `config.py`. We'll need to rework that config strategy in a future release, because it's a pain to treat that file gingerly.

Long story short, we can easily add a `securedrop_max_upload_size` var and use that in the Apache templates. We can't add a simple `lineinfile` task to update the `config.py` file since the line being targeted won't exist in prod templates (and indeed shouldn't), and we don't want to append it. Technically we could use a two-pass approach where we grep for the presence of the `FlaskConfig.MAX_CONTENT_LENGTH` string and then modify it if-and-only-if it's found, but that approach screams brittle. We're already using a long list of tasks where a template would suffice, as long as we properly handle the one-time generated values like the journo/source secret keys.

So, since the PR as written only affects the dev env, I'm happy merging as-is. Down the road we'll need to migrate the `config.py` values to something more amenable to updates.

garrettr:
> So, since the PR as written only affects the dev env, I'm happy merging as-is. Down the road we'll need to migrate the config.py values to something more amenable to updates.

Sgtm; I was afraid there would be something tricky like that. Agree we should merge this for now as a stopgap.

PR #1777, opened 2017-06-06
Fix flashed messages

### LINKED ISSUE #1763: Submission on source interface does not produce flashed message
# Description

A first time submission to the source interface does not produce a flashed message in the staging environment using Tor Browser.

# Steps to reproduce

1. Provision staging VMs
2. Load the source interface in Tor Browser.
3. Generate a new codename.
4. Submit a test message "blah".

# What should happen

A flashed message like this one should appear:

screen shot 2017-05-30 at 7 17 17 pm

# What actually happens

No flashed message is shown. If you refresh the page, the flashed message will then appear. Subsequent submissions will produce the flashed message. Also note that this problem does not appear in Chrome.

garrettr:
## Status

Ready for review

## Description of Changes

Fixes #1763.

* Only use `Cache-Control` on static assets; do not cache HTML in the browser
* Updates the testinfra tests to reflect the changes to the Apache configuration
* Add a stopgap to reduce the memory consumption of the build VM in the staging environment. The overall memory pressure of the staging environment has been causing lockups and crashes on developer machines with <= 8GB of RAM.

## Testing

1. `vagrant provision /staging/ && vagrant reload /staging/`
2. For both the Source Interface and the Journalist Interface:
    1. Navigate the interface in the Tor Browser. Use the Network panel to confirm that HTML responses have `Cache-Control: "no-store"` while other response types have `Cache-Control: "max-age=3600".
    2. Confirm that flashed notifications are now working as expected (e.g. original STR from #1763).

## Deployment

Since this is a modification to the Apache configuration, administrators of existing instances will need to re-run the updated Ansible playbooks to deploy this change.

Note that since #1763 was never deployed to production, this is not currently a bug in production. Production still works just fine, although the user experience is suboptimal because of the lack of caching (see #1152).

2. New installs.

No special considerations.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -146,6 +146,17 @@ Vagrant.configure("2") do |config|
         'securedrop:children' => %w(development),
       }
     end
+
+    # TODO: For some reason, the build VM is defaulting to using 1GB of RAM.
+    # It does not need this much RAM, and the staging environment has been
+    # causing issues (hangs and crashes) on developer machines that have <= 8GB
+    # RAM, so hopefully setting this to a smaller value will help with that
+    # issue. We should look into why this is being automatically allocated 1GB
+    # of RAM instead of the expected 512MB default.
+    build.vm.provider "virtualbox" do |v|
+      v.memory = 512
+    end
+
     build.vm.provider "libvirt" do |lv, override|
       override.vm.synced_folder './', '/vagrant', type: 'nfs', disabled: false
     end
--- a/install_files/ansible-base/roles/app/templates/sites-available/journalist.conf
+++ b/install_files/ansible-base/roles/app/templates/sites-available/journalist.conf
@@ -1,18 +1,27 @@
 ServerName {{ apache_listening_address | default('127.0.0.1') }}
 <VirtualHost {{ apache_listening_address | default('127.0.0.1') }}:8080>
 
-DocumentRoot /var/www/securedrop/static
-Alias /static /var/www/securedrop/static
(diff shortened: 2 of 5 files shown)


Vagrantfile
```diff
@@ -146,6 +146,17 @@ Vagrant.configure("2") do |config|
         'securedrop:children' => %w(development),
       }
     end
+
+    # TODO: For some reason, the build VM is defaulting to using 1GB of RAM.
+    # It does not need this much RAM, and the staging environment has been
+    # causing issues (hangs and crashes) on developer machines that have <= 8GB
+    # RAM, so hopefully setting this to a smaller value will help with that
+    # issue. We should look into why this is being automatically allocated 1GB
+    # of RAM instead of the expected 512MB default.
```
conorsch: Best guess is that it's a bug in how the Vagrantfile declarations are resolved, since that's a notoriously difficult to predict process. The behavior only affects virtualbox—I cannot reproduce in libvirt, which always assigns 512MB to the build VM—so difficult to debug. Happy with the hardcoded solution, since it's explicit.

conorsch:
Changes LGTM. Nice work on light annotations in the Apache templates.

To be clear, the regression this PR is fixing never shipped, so production instances don't need to run playbooks in order to get this "fix". On the other hand, the optimization that caused the regression also never shipped, so Admins _will_ need to run the playbooks in order to take advantage of the smarter asset loading. (Full details in #1763.) Newly installed instances will get the latest-and-greatest functionality out of the box.

PR #1779, opened 2017-06-06
Creates dynamic inventory script for managing host info

### LINKED ISSUE #1774: Automate ansible inventory with dynamic script
We have a current expectation now for admins to manage the ansible `inventory` file ( located in `install_files/ansible-base` ) themselves which is kind of painful. Additionally we already collect and/or manage the relative bits of information that are needed for ansible to connect - so lets create a simple dynamic inventory script so admins dont have to touch that file any longer! Yayyyy

From a high level I see the dynamic script looking like this:

```
if ths files dont exist:
   pull data from site-specific vars
else:
   pull connection address from ths files and user info from site-specific vars
```

conorsch:
## Status

Ready for review.

## Description of Changes

Provides a dynamic inventory script that intelligently uses local IPv4 addresses (read from the required YAML vars file), or Onion URLs for SSH over ATHS, if available. Removes the need for Admins to pass the `-i` flag on every ansible/ansible-playbook command, and also to edit the static inventory file by hand. Removed the static-inventory-munging logic from the `tails_files/install.sh` script, as well as the as-yet-unshipped `manage-groups` role, which was assigning group membership information via tasks, as a stopgap measure until we had dynamic-inventory support. Docs have been updated.

The new script isn't very relevant for the dev env, since it should only be used in prod contexts. Still, this change constitutes a big time saver for Admins, and should also reduce the number of mistakes and misconfigurations during setup. 

Closes #1774. 

## Testing

1. Execute `./inventory-dynamic` within the ansible-base directory, and confirm you get valid JSON.
2. If you don't, make sure you have `group_vars/all/site-specific` (run the `./configure` script to create if not).
3. If `app-ssh-aths` and `mon-ssh-aths` exists, you should see Onion URLs—if they don't, you should see local IPv4 addresses, read from the YAML vars.
4. Review documentation and make sure the updates are clear.

## Deployment

Designed as a time-saver for Admins, and also to reduce mistakes from manual editing of files, so definitely has implications for deployment. The inventory script as written will work equally well during initial install, and for existing instances that  have already had ATHS files fetched back to the Admin Workstation. If experienced SD Admins do not read the changelog, and continue to pass `-i inventory` at the command-line, nothing will break—the `-i inventory` will override the default `inventory-dynamic` declared in `ansible.cfg`.

If an Admin for an existing instance uses the new dynamic inventory file, it will fail first, due to absence of the new `group_vars/all/site-specific` file, and display a message about running the `./configure` script to port the vars over. Due to the automatic documentation in failure mode, this should be an acceptable migration strategy.

## Checklist

We don't have Tails or prod testing in CI yet, so expecting manual verification of the changes here.


docs/backup_and_restore.rst
```diff
@@ -160,7 +160,7 @@ backup:
 .. code:: sh
 
    cd install_files/ansible-base
-   ansible-playbook -i inventory -u <SSH username> -K -t backup securedrop-prod.yml -e restore_file="<your backup archive filename>"
+   ansible-playbook -u <SSH username> -K -t backup securedrop-prod.yml -e restore_file="<your backup archive filename>"
```
msheiny: This isnt particular to this line, but I think you should add logic to also abstract away the ssh_username and just read it from the site-specific vars. I would be fine if that was moved to another PR also. Thoughts?

conorsch: Let's move to another issue. It's a great idea to standardize on that var, but we'd almost surely break backwards compatibility to do so. Will need to think about the smartest way to handle the YAML parsing.

msheiny:
Im not able to get ansible to take advantage of the inventory script on my run under tails 3 rc1 ... getting errors about the host list being empty. I can help debug in the AM but FYI im getting blocked from doing a full test using ansible right now. Were you able to get it running?

msheiny:
Im comparing output from other working inventory scripts ... i think its the `_meta` tag that is not properly scoped. Only `hostvars` should be under the `_meta` tag as far as I can tell. I cant further debug now because I have to run -- but im betting thats the issue. 

selection_024

docs/backup_and_restore.rst
conorsch: Opened #1789 to discuss. I agree that it's a good idea, but would prefer to keep it out of this PR to aid in review.

conorsch:
@msheiny Good catch on the `_meta` block, that's been patched in 1d2610bf07018eeb5b53c114c0fb89fea4de9dac. Ready for re-review!

docs/backup_and_restore.rst
msheiny: think you meant #1796 ;)

install_files/ansible-base/inventory-dynamic
```diff
@@ -0,0 +1,135 @@
...
+    if not host_is_tails():
+        for host, hostvars in inventory["_meta"]["hostvars"].iteritems():
+            (inventory["_meta"]
+                      ["hostvars"]
+                      [host]
+                      ["ansible_ssh_common_args"]) = SSH_COMMON_ARGS
+    return inventory
+
+
+if __name__ == "__main__":
```
msheiny: Technically, there are a few things missing for this to be `compatible` with the ansible inventory spec. BUT it's working right now, so I'm not going to hold up the PR for this but FYI from the [docs](http://docs.ansible.com/ansible/dev_guide/developing_inventory.html) this script is missing:

* a `--list` option that well, does pretty much what its doing now
* a `--host <hostname>` option to return variables from a hostname or return an empty dict

( i also noticed our working infra inventory script for digitalocean is missing these) soo yeah :man_shrugging:

conorsch: Looks like the use of `_meta.hostvars` supersedes the user of `--host <hostname>` invocation: 

>  In Ansible 1.3 or later, if the inventory script returns a top level element called “_meta”, it is possible to return all of the host variables in one inventory script call. When this meta element contains a value for “hostvars”, the inventory script will not be invoked with `--host` for each host.

As for `--list`, the one and only action of the script is to list out all known hosts, so no need to add special logic there.

conorsch:
Conflicts created by merging #1781—will resolve locally.

conorsch:
@msheiny Rebased to resolve conflicts. I'm happy with the diff as-is now. Can you re-review?

PR #1655, opened 2017-04-20
WIP: Tkell test with tor browser

nvesely:
Fixes: #1569

Run the functional tests with Tor Browser. Supersedes #1629.

PR #1780, opened 2017-06-07
Remove dead /howto-disable-js, add explicit warning to use Tor Browser

### LINKED ISSUE #1738: Source Interface instructions contain broken/bad URL
When a Source follows the instructions to bump up the Security Slider settings in Tor Browser, we provide rationale for why, and include two links:

* https://www.wired.com/threatlevel/2013/09/freedom-hosting-fbi
* https://www.theguardian.com/world/interactive/2013/oct/04/egotistical-giraffe-nsa-tor-document

The second link is broken. It returns an HTTP 200, but the document it's linking to is not displayed. At first glance it appears to be a mixed content/CSP error, in which case we could contact TG folks to correct it. Maybe it's simpler to find different coverage and link to that.

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #1738.

   * Remove dead `/howto-disable-js` view
   * Add `/use-tor` view for anyone that is not using Tor Browser
   * Modifies `source.js` to customize the warning bar along the top of the index page

## Testing

Provision a development VM.

### Case 1: Source uses Tor Browser, JS is enabled
1. Set Tor Browser Slider to low so that Javascript is enabled.
2. A purple warning bar should appear:

screen shot 2017-06-06 at 5 08 32 pm

3. Click on the "Learn how to set it to high" link on the source interface index.
4. A bubble should appear explaining how to set the security slider to high.

### Case 2: Source does not use Tor Browser, JS is enabled
1. A purple warning bar should appear:

screen shot 2017-06-06 at 5 12 16 pm

2. Click on the "Learn how to install it" link on the source interface index.
3. The link should go to a page that has a strong recommendation to install it.

### Case 3: Source uses Tor Browser, JS is disabled
1. No purple warning bar appears.

## Deployment

No special requirements for deployments for new or existing installs.


securedrop/source_templates/index.html
```diff
@@ -12,7 +12,8 @@
     {% endassets %}
   </head>
   <body>
-    <div class="warning"><strong>We recommend turning the Security Slider to High to protect your anonymity:</strong> <a id="disable-js" href="/howto-disable-js">Learn how to set it to high</a>, or ignore this warning to continue. <img id="warning-close" src="{{ url_for('static', filename='i/font-awesome/times-white.png') }}" width="12px" height="12px"></div>
+    <div class="js-warning warning"><strong>We recommend turning the Security Slider to High to protect your anonymity:</strong> <a id="disable-js" href="">Learn how to set it to high</a>, or ignore this warning to continue. <img id="js-warning-close" src="{{ url_for('static', filename='i/font-awesome/times-white.png') }}" width="12px" height="12px"></div>
+    <div class="tor-warning warning"><strong>We strongly recommend using Tor Browser to access SecureDrop:</strong> <a id="disable-js" href="{{ url_for('not_using_tor_browser_warning') }}">Learn how to install it</a>, or ignore this warning to continue. <img id="tor-warning-close" src="{{ url_for('static', filename='i/font-awesome/times-white.png') }}" width="12px" height="12px"></div>
```
fowlslegs: You should change the `id` attribute here to `use-tbb` or something.

securedrop/static/js/source.js
```diff
@@ -1,9 +1,17 @@
 // Warn about using Javascript
```
fowlslegs: Can you update this comment?

securedrop/source_templates/index.html
conorsch: Nit: we're duplicating `id`s for `disable-js`. We should use a dedicated id for the tor-browser warning if we need to refer to the element by id.

garrettr:
> I think you want to get rid of the function call on line 38 of securedrop/static/js/source.js

@fowlslegs is right. In addtion, @redshiftzero, as long as we're cleaning up dead code, you might as well also get rid of the Array of `TBB_UAS`, since it was superseded by `TBB_UA_REGEX` ages ago.

garrettr:
> I would also say TBB_UAS and tbb_version() could be cleaned up, except that I think it would be a good idea to actually use those (well, at least the tbb_version() function) to detect and warn about out-of-date TB versions. The current version is 45. If someone is running an older version with a different UA, and also has not disabled JS, we should get them to upgrade first. Then they'll get the disable JS popup the next time they try to use SD after upgrading. While this won't generally detect when a client is using a slightly out-of-date TB, it will detect when a client is using a very out-of-date TB.

I like this idea, @fowlslegs! But it is out of scope for this PR and for the 0.4 release generally. Please file a follow-up issue so we don't forget about it 😸 .

garrettr:
> I agree with Garrett that source.js needs test, and feel strongly that unless there's a really good reason they should be part of this PR, as opposed to a follow-up issue (although one may also be created to tackle the need for more coverage of our JS).

I think this is worth taking a look at as part of this PR, but **don't agree that it should be a requirement for this PR to land**, for two reasons:

1. Since we don't already have a client-side JS testing pipeline, @redshiftzero would have to create one. That may or may not be easy, but if it's not easy I don't think it should block this PR from being merged. We may be able to stopgap fully or at least partially by adding some checks related to this messaging to the Selenium-based functional tests.
2. Since this PR hinges on a difference in behavior between browsers (Tor Browser vs. "not Tor Browser"), modifying the test environments to have an additional browser available and wiring up the tests to optionally use it is a large enough task that I think it probably deserves a dedicated PR.

fowlslegs:
@garrettr I'm neither sure if functional tests provide the only way to properly test this code, nor am I sure if that is what you are suggesting (although both could be true). I do think they should be part of a larger testing story, surely.

Regarding point 2 above, and in the context of functional tests, I think that once #1655 is merged, we should be able to test this by just changing the UA and security slider settings rather than rigging up a second browser.

To be clear, I didn't state that tests should be an absolute requirement for this PR to land. Maybe I could have been more verbose, but I was more or less hoping @redshiftzero would look into what a proper testing story here would look like, then implement a part or all of that if possible. And if it wouldn't be practical to implement the whole story as part of this PR, an explanation of why would suffice to satisfy the "unless there's a really good reason" exception. In short, it seems sensible to me to have a strong understanding of what needs to be done for testing before making an exception for merging code without tests.

redshiftzero:
I think we want to keep the code on line 38 of `source.js` - it's what displays this nice bubble to explain to the source how to turn up the security slider:

screen shot 2017-06-08 at 2 21 49 pm

Let me know if you disagree. But I've added a comment in 3437629 explaining this function for the future.

Re: improving the testing story, I'll look into that now and add a commit if smallish and file an issue for a subsequent PR if not.

redshiftzero:
OK tests added

fowlslegs:
> OK tests added

Does this mean ready for re-review or do you have more changes yet?

redshiftzero:
Drat, the tests were the last thing to add but I broke CI so I will have to investigate. Will ping and assign you for re-review

redshiftzero:
OK I've added a new functional test for the warning added in this PR: it verifies that if someone is not using Tor Browser (in this case Firefox) then the new banner warning instructing sources to go and download Tor Browser is displayed. 

I experimented a bit adding an additional test for the existing functionality (the warning when a source is using Tor Browser and JavaScript) by setting the User Agent of Tor Browser in Selenium, but I had some trouble getting that to pass in CI (potentially issues caused with `window.navigator.mimeTypes` not being set to mimic Tor Browser), so I'm going to punt that until #1655.

Ready for re-review!

garrettr:
After the back-and-forth about line 38 of source.js, I decided that it would be quick and easy for me to do some significant cleanup of the security slider frontend code. I discussed it with @redshiftzero and she agreed to allow me to push my changes to her branch for inclusion as part of this pull request. I just pushed, so I'm flagging @redshiftzero for review on her own PR to make sure she's cool with the commits that I've added to it 😄 .

garrettr:
Actually, it appears I can't flag @redshiftzero for review on her own PR. I guess that makes sense. Take a quick look anyway, @redshiftzero (although we already reviewed most of this stuff in person)!

redshiftzero:
Changes look good to me! :+1: :shipit:

securedrop/source_templates/use-tor-browser.html
```diff
@@ -0,0 +1,7 @@
+{% extends "base.html" %}
+{% block body %}
+<h1>You Should Use Tor Browser</h1>
+<p>If you are not using Tor Browser, you <strong>may not be anonymous</strong>.</p>
+<p>If you want to submit information to SecureDrop, we <strong>strongly advise you</strong> to install Tor Browser and use it to access our site safely and anonymously.</p>
+<p>Copy and paste the following address into your browser and follow the instructions to download and install Tor Browser: <pre>https://www.torproject.org/download.html.en</pre></p>
```
garrettr: I think this might be a better link to direct folks to instead: https://www.torproject.org/projects/torbrowser.html.en. I think the design is cleaner, and it has more detailed installation instructions (with screenshots!).

It might be worth hopping into the tor-dev IRC channel and asking what the "official" recommendation is.

fowlslegs: Agree we should link this page for the reasons @garrettr listed.

Issue #1551, opened 2017-02-02
Test SecureDrop on Tails 3.0


garrettr:
We will need to test SecureDrop on Tails 3.0 (especially the functionality provided by `tails_files`) before it is released. The first beta just became available: https://tails.boum.org/news/test_3.0-beta1/index.en.html.

In addition, we will need to devise, test, and document an upgrade path for all SecureDrop users, because this is a major version change (from Tails 2.x -> 3.x), and I expect it will not be possible for them to automatically upgrade their Tails devices.

garrettr:
We should also contact the Tails project and find out how long they plan to support the 2.x release series, because the primary motivation for migrating SecureDrop users is to ensure they will continue to be able to receive security updates in a timely manner.

freddymartinez9:
would be happy to test this if there is a need to have multiple people test besides FPF developers.

fowlslegs:
@freddymartinez9 Yeah, especially if you would be willing to switch to using Tails 3 beta for your JWs/AW that would definitely help us catch bugs.

heartsucker:
Leaving a note about linking this to the issue that (presumably) triggered it. #1549

freddymartinez9:
I was thinking about how to do this testing and it seems to me we need to someone to test using a fresh install of SD using tails 3.0 with persistent storage, especially the `tails_files` scripts. Separately, I will `dd` my current Tails 2.0 USBs (JA/AW) to a new USB, upgrade to tails 3.0 on that (while preserving persistent storage) and test away. My feeling is that someone at FPF should test the former, and I can test the latter.

freddymartinez9:
The only issue I found is that Tails 3.0 seems to have dropped support for the `.kdb` extension in support of `.kdbx` (Keypass 2 DB). You can go to KeePass's Settings > Database > Import KeePass 1 database > Import but this is a one-way function. This might be fine but needs to be documented.

The update process I followed: Download Tails 3.0beta1 to a clean USB device. For all Tails devices, I created a clone of my USB devices on a reasonably secure OS (Qubes) `sudo dd if=/dev/xvdi of=/dev/xvdj bs=4M; sync;` for Admin and for the SVS. Update both to tails 3.0 beta1 by cloning. Login to Journalist Workstation as normal (FWIW: I tested with TBB security slider set to High). Move archives to the DTD and on to the SVS. Decrypt archives as always. 

[...]

PR #1782, opened 2017-06-07
Validates all prod vars prior to installation

### LINKED ISSUE #749: Verify contents of prod-specific.yml before installing SecureDrop

### LINKED ISSUE #1028: Second file path option for OSSEC public key is wrong
In the instructions for prod-specific.yml, we give users the choice on putting their public key for ossec notifications at 2 locations.  The longer of the two should be changed to `install_files/ansible-base/roles/ossec_server/files`, which is where ansible is actually looking for it.

### LINKED ISSUE #1257: GPG keys should be validated against fingerprints before provisioning

conorsch:
## Status

Ready for review.

(Confirmed with manual testing that the presence of the `openssh-client` directory is sufficient indication that SSH dotfiles persistence is enabled. This PR is no longer marked "Work in progress.")

## Description of Changes

Validates all site-specific production variables prior to running the prod playbook. Designed to catch common misconfiguration errors _before_ the playbook is run, to prevent late-stage failures that require excessive troubleshooting. Since we're switching to Ansible v2 (#1146), we can leverage the `msg` parameter on the `assert` task, which allows us to provide custom failure messages for variable validation, so Admins know where the problem lies and what to change.

The new validation logic requires Tails 3 (#1551) or greater as host OS, and additionally checks that persistence is enabled, and specifically that SSH dotfiles persistence is enabled. There are also functional checks that confirm the specified SMTP relay is accessible. It also pulls in @heartsucker's GPG validation logic presented in #1259, to confirm that the GPG public key files and full fingerprints match.

At some point in the future, we may wish to port the validation logic to pure Python, which would allow us to loop over input prompts when values don't pass validation, rather than collecting all variables and analyzing them as a set. 

Fixes #749. Fixes #1257. Obsoletes and therefore closes #1028. Includes and therefore closes #1259. 

## Testing

Requires testing on hardware, with an Admin Workstation on Tails 3 (#1551) configured with persistence and SecureDrop-specific config.

1. Boot into the Admin Workstation, with persistence enabled.
2. Run `./configure` in the ansible-base directory (see #1762 for more info) and configure working prod-compatible vars for GPG keys and OSSEC email authentication.
3. Confirm that the validation logic passes.

Then, to check failure mode:

1. Edit `group_vars/all/site-specific` with junk values, e.g. alter a single character in one of the GPG fingerprints.
2. Rerun `./configure`.
3. Confirm that the failure message is informative.

## Deployment

Mostly useful for first-time Admins during install time, but since the validation logic is run as part of the prod playbook, the changes should also catch future misconfiguration if required vars are not declared. Previously, such failures would only occur during the playbook run, which due to the Tor over ATHS strategy, could occur 10 minutes or more after executing the playbook command. 

## Checklist

Requiring manual testing on hardware to confirm functionality.


conorsch:
Good news, looks like the persistence check as written is sufficient, and runs with normal user privileges. Tested on Tails 3 rc1. Results:

### Persistence with no SSH dotfiles config
Persistence volume configured and unlocked, but the "SSH" option was _not_ checked in the Configure Persistence Volume settings.
```
amnesia@amnesia:~$ ls -l /live/persistence/TailsData_unlocked/
total 24
-rw------- 1 tails-persistence-setup tails-persistence-setup     0 Jun  7 18:19 live-additional-software.conf
drwx------ 2 root                    root                    16384 Jun  7 18:15 lost+found
-rw------- 1 tails-persistence-setup tails-persistence-setup    43 Jun  7 18:16 persistence.conf
drwx------ 3 amnesia                 amnesia                  4096 Jun  7 18:20 Persistent
```

### Persistence with SSH config
Persistence volume configured and unlocked, _with_ the "SSH" option checked in the Configure Persistence Volume settings. Machine was rebooted to ensure the SSH config option was active in the currently booted environment.

```
amnesia@amnesia:~$ ls -l /live/persistence/TailsData_unlocked/ 
total 40
drwx------ 2 amnesia                 amnesia                  4096 Jun  7 18:24 dotfiles
drwx------ 3 amnesia                 amnesia                  4096 Jun  7 18:24 gnupg
-rw------- 1 tails-persistence-setup tails-persistence-setup     0 Jun  7 18:19 live-additional-software.conf
drwx------ 2 root                    root                    16384 Jun  7 18:15 lost+found
drwxr-xr-x 2 root                    root                     4096 May 20 14:48 nm-system-connections
drwx------ 2 amnesia                 amnesia                  4096 Jun  7 18:24 openssh-client
-rw------- 1 tails-persistence-setup tails-persistence-setup   221 Jun  7 18:21 persistence.conf
drwx------ 3 amnesia                 amnesia                  4096 Jun  7 18:27 Persistent
```

Will clean up PR accordingly, and mark as ready for review.

conorsch:
@msheiny Ready for review.

install_files/ansible-base/roles/validate/tasks/validate_ossec_info.yml
```diff
@@ -0,0 +1,69 @@
...
+      - "'@' not in sasl_username"
+
+- name: Validate SASL password for OSSEC config.
+  assert:
+    that:
+      # Ensure not the dummy value specified in example vars file.
+      - sasl_password != "password123"
+      - sasl_password|length >= 8
+    msg: >-
+      Password for OSSEC email account must be strong.
```
msheiny: I dunno if this is a useful validation test. I mean, its a good security-practice check but what if the specific account an org is using has 8 characters and there is ? That kind of validation seems better left up to the organization and their email provider rather than during the setup process. The message is getting encrypted so is it our place to mandate the length of the SMTP account to send the email out?

conorsch: I hear that. Definitely felt gross as soon as I wrote a check for password strength, but was trying to determine that a real password had indeed been typed in. `|length > 1` didn't feel right.

msheiny:
Soo do you think it is out of scope to add an SMTP cred verification to this script? I think we hit that issue a lot in the field (at least ive seen it a few times). Thats fair if you think its better left to another PR - just throwing it out here.

conorsch:
Considered a functional test that would try to auth, but the most common auth problem is not a mistyped password, it's a 2FA step-up verification problem. Definitely _don't_ want to add any validation tasks that fail when the config is actually valid, so perhaps I'm being overly conservative for now. This is already such a large win 0.3.12 -> 0.4, I'm inclined to play it safe and help as much as we can, but without going overboard.

Another thing that could be useful is adding docs links in the custom error messages. Didn't do that for two reasons:

1. updating the URLs over time would be a pain
2. most of the URLs would point to either the install guide or the OSSEC guide

msheiny:
To recap our in-person, some might use the adjective `sexy`, in-person discussion. There is a workflow problem now if validation fails on a field and you want to go back and fix. I currently suggest that you add messaging post failure of the `configure` script telling the user to go back and edit the `site-specific` vars file directly to correct the mistake and re-run the validation script. This isnt the most ideal workflow but for the scope of this PR i think it makes the most sense. The alternative would be rewriting the validation logic / configure script in `python` which would provide more control over the prompt and validate process.

conorsch:
>  I currently suggest that you add messaging post failure of the `configure` script telling the user to go back and edit the `site-specific` vars file directly to correct the mistake and re-run the validation script.

This is the best compromise, and definitely worth adding. The solution of "validate only after all vars are entered" coupled with "write all vars to disk regardless of validation state" makes for a tricky workflow, but messaging can provide clarity. :+1:

install_files/ansible-base/roles/validate/tasks/validate_gpg_info.yml
```diff
@@ -1,20 +1,24 @@
 ---
 
-- name: Validate GPG info.
+- name: Validate GPG fingerprints.
   assert:
     that:
```
msheiny: Found a great edge case for you :) I typed the file name wrong and it passed this section but when it went to the script it failed and it was driving me crazy. Just says the fingerprints dont match (so i assumed it was reading my file in wrong). Can you detect if the file actually exists here?

msheiny: errr `i assumed it was reading my file right` is what i meant to say.

conorsch: Great catch! Was surprised there's no `|abspath` filter in Ansible that would let you expand a path. Since Ansible looks in many different locations to expand a path, it'd be nice to introspect that a bit. Just occurred to me that a `lookup('file', filename)` would do the job nicely.

conorsch:
Added the requested improvements, @msheiny. The GPG pubkey file checks are much improved when the files cannot be found:

### Before:
```
TASK [validate : Confirm public key file and fingerprint match.] ***************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "rc": 1,
"stderr": "", "stdout": "Creating temporary GPG config dir for testing key
import...\nImporting pubkey file from 'conor.asc2'...\nFailed! Specified
fingerprint does NOT match pubkey file.\nCleaning up temporary GPG config
dir...\n", "stdout_lines": ["Creating temporary GPG config dir for testing key
import...", "Importing pubkey file from 'conor.asc2'...", "Failed! Specified
fingerprint does NOT match pubkey file.", "Cleaning up temporary GPG config
dir..."]}
```

### After:
```
TASK [validate : Confirm GPG public key files exist locally.] ******************
 [WARNING]: Unable to find 'conor.asc2' in expected paths.

fatal: [localhost]: FAILED! => {"failed": true, "msg": "The conditional check
'lookup('file', item.keyfile)' failed. The error was: An unhandled exception
occurred while running the lookup plugin 'file'. Error was a <class
'ansible.errors.AnsibleError'>, original message: could not locate file in
lookup: conor.asc2"}

TASK [validate : GPG public key file validation failed.] ***********************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg":
"Failed to find the public key file 'conor.asc2' on the Admin Workstation. You
should manually edit the file `group_vars/all/site-specific` and confirm that
the public key filenames are set correctly in vars `ossec_alert_gpg_public_key`
and `securedrop_app_gpg_public_key`. The public key files should exist in the
`install_files/ansible-base/` directory."}
```

Used a role default var to provide the template string for pointing to the `group_vars/all/site-specific` file, so we have a single point of update to adjust the messaging in the future.

PR #1783, opened 2017-06-08
Add issue template and more details to the contributor guide

redshiftzero:
## Status

Ready for review

## Description of Changes

 * Update pull request template. I didn't add to run staging locally because I don't think we are expecting people to do this.
 * Add issue template
 * Add information to the contributor guide regarding our current explicit merge strategy and cleaning up one's git history via interactive rebase

## Testing

Check out the typesetting and most importantly see if you agree with these guidelines 😃 

## Deployment

N/A

## Checklist

- [x] docs linting passed locally

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,6 +11,7 @@ Changes proposed in this pull request:
 ## Testing
 
 How should the reviewer test this PR?
+Write out any special testing steps here.
 
 ## Deployment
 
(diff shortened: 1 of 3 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -29,3 +30,6 @@ Any special considerations for deployment? Consider both:
 
 - [ ] Testinfra tests pass on the development VM
```
garrettr: Discussed this in person with @redshiftzero and @conorsch. We agreed that this recommendation has the potential to be misleading/confusing, because a significant portion of the system configuration is *not* applied in the development VM, and in many cases you would need to ensure the testinfra tests on passing on one of the staging machines.

For example, in testing #1777, the testinfra tests passed on the development VM because #1777 primarily modified the Apache configuration, which is not applied in the development VM (but it applied in the app-staging VM).

We agreed that the most parsimonious way to resolve this potential confusion in the context of this PR is to change the wording of this checklist item to something like the following:

redshiftzero:
Thanks for looking this over @garrettr! I modified the system configuration checklist item as suggested, so this one should be good to go.

PR #1789, opened 2017-06-10
tests: store: test relative path error handling

dachary:

### DIFF
--- a/securedrop/tests/test_store.py
+++ b/securedrop/tests/test_store.py
@@ -22,11 +22,23 @@ class TestStore(unittest.TestCase):
         utils.env.teardown()
         db_session.remove()
 
-    def test_verify(self):
+    def test_verify_path_not_absolute(self):
         with self.assertRaises(store.PathException):
             store.verify(os.path.join(config.STORE_DIR, '..', 'etc', 'passwd'))
-        with self.assertRaises(store.PathException):
+
+    def test_verify_in_store_dir(self):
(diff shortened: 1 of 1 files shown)


securedrop/tests/test_store.py
```diff
@@ -27,6 +27,12 @@ def test_verify(self):
             store.verify(os.path.join(config.STORE_DIR, '..', 'etc', 'passwd'))
         with self.assertRaises(store.PathException):
             store.verify(config.STORE_DIR + "_backup")
+        STORE_DIR = config.STORE_DIR
+        with self.assertRaisesRegexp(store.PathException,
+                                     'is not absolute'):
```
redshiftzero: In `store.py` there are two exceptions that can be thrown that would match the regexp here - when the path passed as an arg is not absolute (line 44) _and_ when `config.STORE_DIR` is not absolute (line 36). Let's change the regexp to be more precise so that this assert can only pass if the behavior on line 36 is correct. For example:

```
with self.assertRaisesRegexp(store.PathException,
            'config.STORE_DIR\(\S*\) is not absolute'):
```

dachary:
* split in three functions
* made sure STORE_DIR is reset in case of error with a **finally**
* fixed the store_dir_not_absolute regexp to not be ambiguous

dachary:
thanks for your reviews :-) repushed with the suggested changes.

PR #1790, opened 2017-06-10
tests: db: reply returns prematurely

dachary:

### DIFF
--- a/securedrop/tests/utils/db_helper.py
+++ b/securedrop/tests/utils/db_helper.py
@@ -60,8 +60,8 @@ def reply(journalist, source, num_replies):
         replies.append(reply)
         db.db_session.add(reply)
 
-        db.db_session.commit()
-        return replies
+    db.db_session.commit()
+    return replies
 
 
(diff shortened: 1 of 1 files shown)


redshiftzero:
I suggest we just go and and merge this one in. It sounds like we are in agreement that the only line in the diff here is a solid change. Does that sound reasonable @fowlslegs?

dachary:
@fowlslegs I repushed with the suggested change (commit outside of the loop). Block shifting is indeed an easy way to get things wrong in python. Not just with vim ;-)

dachary:
@fowlslegs feel free to merge this one or #1795, I'm glad this is fixed either way.

fowlslegs:
I don't care at all. I'm unable to merge either because I either need a SD team member to approve my PR, or we need to manually run the CI on this PR as those two things, respectively, are blockers to me being able to merge one of them due to our Github-enforced merge policies.

dachary:
@fowlslegs I see ;-) I'll close this one in favor of your  #1795 so there is no duplicate.

PR #1792, opened 2017-06-12
tests: crypt_util: encrypt error condition

dachary:

### DIFF
--- a/securedrop/tests/test_crypto_util.py
+++ b/securedrop/tests/test_crypto_util.py
@@ -5,6 +5,7 @@ import unittest
 os.environ['SECUREDROP_ENV'] = 'test'
 import config
 import crypto_util
+import store
 import utils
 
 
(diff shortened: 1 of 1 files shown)


securedrop/tests/test_crypto_util.py
```diff
@@ -23,3 +24,23 @@ def test_clean(self):
             crypto_util.clean('foo bar`') # backtick is not currently allowed
         with self.assertRaises(crypto_util.CryptoException):
             crypto_util.clean('bar baz~') # tilde is not currently allowed
+
+    def test_encrypt(self):
+        journalist, _ = utils.db_helper.init_journalist()
+        source, _ = utils.db_helper.init_source()
+        replies = utils.db_helper.reply(journalist, source, 1)
```
fowlslegs: Why are you creating a `Reply` here? Think this should probably just be removed.

securedrop/tests/test_crypto_util.py
```diff
@@ -23,3 +24,23 @@ def test_clean(self):
             crypto_util.clean('foo bar`') # backtick is not currently allowed
         with self.assertRaises(crypto_util.CryptoException):
             crypto_util.clean('bar baz~') # tilde is not currently allowed
+
+    def test_encrypt(self):
+        journalist, _ = utils.db_helper.init_journalist()
```
redshiftzero: You can remove this `init_journalist()` line as you should only need 

```
source, _ = utils.db_helper.init_source()
```

for the setup for the function under test here

dachary:
* split in two functions
* removed the journalist & reply lines

dachary:
* used assertGreater instead of assert

dachary:
@redshiftzero @fowlslegs hopefully addressed your comments, thanks for the careful review.

Issue #1796, opened 2017-06-12
Standardize on single Admin username


conorsch:
# Feature request

## Description

The Ansible config used to set up the servers at install time contains a var called `ssh_users`, which should contain the username of the Admin user account on the remote servers, created during the OS install. The username will be configured with passwordless sudo, and whitelisted for SSH access in the pam config.

To minimize error-prone typing on the command-line by Admins, it would be great to reuse the remote username throughout the site config. We haven't been reusing it to date because _technically_ Admins could specify that var as a list, which makes it impossible to choose the correct username. Given the improvements to validation logic in #1782, we should:

* enforce that the `ssh_users` var contains a single variable name
* update all documentation to remove `-u <username>` from Ansible commands
* ensure that playbooks detect a YAML list and fail with an informative message, stating the requirement to update the var locally

The trade-off for standardizing on string-over-list is that _some_ Admins may have declared the `ssh_users` var to be a list. It's impossible to state the percentage of Admins doing so accurately, since we don't have access to site-specific vars for production instances, but it's a safe bet to say that most Admins are _not_ declaring the var as a list. In fact, none of our documentation includes examples of doing so. 

The advantage for standardizing on string-over-list is that we can simply the documented commands and streamline the config process. Lack of the `-u <username>` option is a common stumbling block that generates support requests, as well, so it seems like a ripe target for improvement.

## User Stories

As a new Admin, I want to declare the username for SSH once and only once during configuration of the Admin workstation.

As an existing Admin, I want subsequent playbook runs to inform me of changes I must make to my local vars file, if the vars cannot be understood without prompting.

msheiny:
Ohhhhh i like that `user stories` is part of the issue template now :)

This is a really solid written issue as-is - I dont have anything to add and agree with the implementation and testing high-level strategies. I would just ensure that #1781 gets in before this is in PR state so that the appropriate changes are also made to the `securedrop-admin` script (which provides a flag to provide a username on the cli). I'm not 100% we should kill that option... actually yeah i cant think of a good reason to leave it in.

conorsch:
Good, glad we agree on the strategy here. There are a few places username consolidation should take place:

* [ ] `securedrop-admin` script (#1781)
* [ ] dynamic inventory script (#1779)
* [ ] `validate` role logic (to provide an informative error message about transition)
* [ ] throughout Admin documentation (to remove the `-u <username>` flag from Ansible commands)

Any other targets should be added to this list.

[...]

PR #1803, opened 2017-06-14
Configures Tails workstations via Ansible

### LINKED ISSUE #1793: Port Tails setup to Ansible logic
# Feature request

## Description

The existing setup logic for provisioning Admin and Journalist Workstations is to run the `tails_files/install.sh` script. That script contains about 500 lines of Bash, and contains some tricky grep and sed logic that is difficult to test reliably (which has caused problems in the past, e.g. #1633). The sole reason we're using Bash to configure the Tails environment is that Ansible is only available on the Admin Workstation, but not the Journalist Workstation. Give the tooling presented in #1781 to wrap the Ansible bootstrap logic, this is a fine opportunity to port the Bash logic over to stable Ansible playbooks, so we can make use of DRY vars.

Additionally, much of the logic contained in `tails_files/install.sh` contains legacy cleanup tasks. Due to the Tails 2 -> 3 migration for 0.4 (#1551), Admins must configure new Tails volumes, so much of the legacy cleanup tasks can be removed.

## User Stories

As an Admin, I want to run a single script and configure the Tails environment post-install. The script should understand target workstation roles such as "Admin" and "Journalist" and configure them accordingly.

Relevant issues: #1533, #1176, #1177.

conorsch:
## Status

Ready for review.

## Description of Changes

Removes `tails_files/install.sh` bash script and ports all Tails configuration logic to Ansible. Updates the `securedrop-admin tailsconfig` command to trigger the dedicated Tails config playbook. Works for both Admin and Journalist Workstations. Does not require specification by Admin—if the `app-ssh-aths` and `mon-ssh-aths` files are present in the `install_files/ansible-base` directory, then the SSH access is configured. If the Journalist Interface ATHS info is missing, configuration halts with an informative error message.

Fixes #1793.

## Testing

1. Boot up Tails 3 with a fresh persistence volume.
2. Run through an install on prod VMs (or hardware, as you will) via `./securedrop-admin install`.
3. After installation, run `./securedrop-admin tailsconfig` to use the new Ansible-based Tails setup logic.

Note that 2) is optional, since you may already have `-aths` files locally. As long as those still refer to machines with active Tor interfaces (so you can test the interfaces), it's OK to reuse them, since the subject of this PR is the Tails config. Note that 1) is required!

After running 3), confirm the following:

1. SecureDrop desktop icons exist for both Source and Journalist Interfaces.
2. Double-clicking the desktop icons load the respective interfaces in Tor Browser.
3. Running `ansible all -m ping` within the `ansible-base` directory returns a green OK message for each host.

Then reboot Tails, and confirm that 1-3 above are still true. Doing so will confirm that the persistence options were correctly configured.

Remember that use of prod VMs will cause failure due discrepancies in the AppArmor profile for Apache, caused by the Document -> Journalist Interface renaming (#1614). Edit the `/etc/apparmor.d/usr.sbin.apache2` file on the Application Server and add an exemption for the journalist.conf file in sites-available, then bounce apparmor and apache.

## Deployment

Makes first-time setup of a new instance substantially easier for Admins. Should greatly reduce the opportunity for user error during the configuration and install process. Not as relevant for existing Admins, although the new `tailsconfig` logic will run if/when new Journalists are onboarded to an existing instance.

## Checklist

Relying on manual testing to validate Tails-based config.


conorsch:
Had some problems connecting to SSH over Tor post-provisioning during local testing. Might have screwed up my network settings though, so will need to try again.

Also noticed that the desktop icons do not appear immediately—only after a reboot of Tails. That's different behavior than we saw in Tails 2, although I'm not sure what the root cause is yet.

Note that there's a minor regression presented in the Ansible logic, in that the Admin SSH username is omitted from the SSH aliases. Opened #1796 to discuss separately, so not handling in this PR.

install_files/ansible-base/roles/tails-config/defaults/main.yml
```diff
@@ -0,0 +1,43 @@
...
+# to ensure absence from all locations succinctly.
+tails_config_deprecated_directories:
+  - "{{ tails_config_live_persistence }}"
+  - "{{ tails_config_securedrop_dotfiles }}"
+  - "{{ tails_config_network_manager_dispatcher }}"
+  - "{{ tails_config_network_manager_dispatcher }}/custom-nm-hooks"
+
+tails_config_deprecated_config_files:
+  - 70-tor-reload.sh
+  - 99-tor-reload.sh
```
conorsch: Logic like these legacy cleanup tasks isn't strictly necessary, since we're requiring Admins to install Tails 3 on a fresh stick during upgrade to 0.4. Depending on how they copy the dotfiles over, though, it's conceivable something could be dangling, so I figured better safe than sorry. Also, we know have a dedicated task list for handling legacy cleanup tasks on Tails in the future.

securedrop-admin
```diff
@@ -128,8 +128,11 @@ def install_securedrop(args):
 
 def run_tails_config(args):
     """Configure Tails environment post SD install"""
-    subprocess.check_call(['sudo',
-                            os.path.join(SD_DIR, 'tails_files/install.sh')])
+    activate_venv()
+    subprocess.check_call([os.path.join(ANSIBLE_PATH, 'securedrop-tails.yml'),
```
dachary: shouldn't there be an *['ansible'...* before os.path.join ? Or is there some magic making the .yml executable ?

fowlslegs: Yes, a shebang: https://github.com/freedomofpress/securedrop/pull/1803/files#diff-a4345ca9831def9cc2022200e0df2e2eR1.

dachary: Nice :-)

msheiny:
Burning through a local install on tails now

msheiny:
Ohhh so I hit an interesting issue -- my securedrop is cloned to `~/securedrop` instead of `~/Persistent/securedrop`. Interesting I've been able to get this far in the process without something yelling at me (i blame my previous work on the installation script).

msheiny:
To clarify -- the issue is the script is breaking in certain areas at the tailsconfig where it is looking for the tor files to mix them in the system torrc.

conorsch:
@msheiny The new validation checks should fail out if `~/Persistent/securedrop` does not exist: https://github.com/freedomofpress/securedrop/pull/1803/files#diff-dc14710315347f520dea7178bacdf5d6

msheiny:
:laughing:  okay i see --- i had a left over `securedrop` folder in `~/Persistent` from a previous run... 

So I guess we need a ticket to cover the case `when an FPF person has two copies of `securedrop` in his home folder and he's rsyncing files through a virtualized instance of `Tails` and accidentally kicks off the install from wrong folder` validation step?

msheiny:
`SSH aliases are set up. You can use them with 'ssh app' and 'ssh mon'. "`

This messaging is confusing because I forgot that we have to specify the username currently. Would recommend taking that from the config and dropping into the sshconfig file... yes i realize there is another ticket covering the user stuff ;)

securedrop-admin
conorsch: I'm actually not much of a fan of the executable playbook pragma, simply because it's not common in the Ansible community. Added it here to be consistent with the tooling changes as of #1781.

conorsch:
Right, SSH aliases are punted to #1796.

> Note that there's a minor regression presented in the Ansible logic, in that the Admin SSH username is omitted from the SSH aliases. Opened #1796 to discuss separately, so not handling in this PR.

install_files/ansible-base/roles/tails-config/tasks/cleanup_legacy_artifacts.yml
```diff
@@ -0,0 +1,25 @@
...
+    path: "{{ tails_config_live_persistence }}/dotfiles/.xsessionrc"
+    state: absent
+
+- name: Remove deprecated Document Interface desktop icons.
+  file:
+    state: absent
+    path: "{{ item }}"
+  with_items:
+    - "{{ tails_config_amnesia_home }}/Desktop/document.desktop"
+    - "{{ tails_config_amnesia_home }}/.local/share/applications/document.desktop"
```
msheiny: I would add that you should also remove `Report_an_error.desktop` and `tails-documentation.desktop`. These icons clutter the desktop and our users are likely hitting errors with SD --- if anything I would add a link to OUR website and OUR docs. That second part can be a new PR (up to you) but I think the removal of existing icons should go in this PR.

install_files/ansible-base/securedrop-tails.yml
```diff
@@ -0,0 +1,23 @@
...
+  connection: local
+  gather_facts: yes
+  roles:
+    - role: tails-config
+      tags: tails-config
+  tasks:
+    - name: Configuration complete.
+      debug:
+        msg: >-
+          Successfully configured Tor and set up desktop bookmarks for SecureDrop!
```
msheiny: so one issue i hit was that my desktop icons didnt show up right away -- i know thats not a new issue and cant find a specific fix online (gnome's refresh `alt`+`f2` --> `r`) didnt resolve it for me. rebooting did tho. so i dunno if we want to say `hey if your icons arent showing up on desktop, please reboot`

@conorsch responded in chat: `i noticed that too mike, you can quote my observation when you comment on github. my gut tells me there's some handler we can run to refresh the desktop, bu ti don't know what it is`

install_files/ansible-base/roles/tails-config/tasks/cleanup_legacy_artifacts.yml
conorsch: Great idea. Will need to confirm that one-time removal sticks after rebooting. If that's the case, will push up changes.

conorsch:
Unfortunately the Tails desktop icons are baked in. Removing them from `~/Persistent/Desktop` only works for the current session, since the icons are copied to that location from `/etc/skel`. Removing the icons in `/etc/skel/Desktop` doesn't work across reboots due to Tails not storing system state changes.

Technically we could shoehorn some customization logic into the `securedrop_init.py` script that's run via NetworkManager, but I don't want to overload that script more than absolutely necessary. If anyone feels strongly about removing the Tails-specific desktop icons, open a separate issue to discuss.

Issue #1797, opened 2017-06-13
gpg --recv-key hangs on tails-3.0


dachary:
# Bug

## Description

docs/set_up_admin_tails.rst suggests:

<pre>
    gpg --recv-key "2224 5C81 E3BA EB41 38B3 6061 310F 5612 00F4 AD77"
</pre>

but it hangs and timesout.

## Steps to Reproduce

* Install https://tails-dl.urown.net/tails/stable/tails-amd64-3.0/tails-amd64-3.0.iso with persistence and boot with an admin password.
* Connect to internet (via wifi, no firewall)
* Verify connectivity with sudo apt-get update
* Open a terminal
* $ cat /home/amnesia/.gnupg/dirmngr.conf 
use-tor
keyserver hkp://jirk5u4osbsr34t5.onion
* gpg --debug-all --recv-key "2224 5C81 E3BA EB41 38B3 6061 310F 5612 00F4 AD77"

## Expected Behavior

The key is imported.

## Actual Behavior

gpg: reading options from '/home/amnesia/.gnupg/gpg.conf'
gpg: enabled debug flags: packet mpi crypto filter iobuf memory cache memstat trust hashing ipc clock lookup extprog
gpg: DBG: [not enabled in the source] start
gpg: DBG: chan_3 <- # Home: /home/amnesia/.gnupg
gpg: DBG: chan_3 <- # Config: /home/amnesia/.gnupg/dirmngr.conf
gpg: DBG: chan_3 <- OK Dirmngr 2.1.18 at your service
gpg: DBG: connection to the dirmngr established
gpg: DBG: chan_3 -> GETINFO version
gpg: DBG: chan_3 <- D 2.1.18
gpg: DBG: chan_3 <- OK
gpg: DBG: chan_3 -> KS_GET -- 0x22245C81E3BAEB4138B36061310F561200F4AD77
gpg: keyserver receive failed: Connection timed out

## Comments

It is entirely possible I'm doing something stupid, this is my first time with tails ;-) I tried --keyserver but gpg would not accept that option.

dachary:
Filed an issue at tails but I have the distinct feeling this is not where it should be reported: https://labs.riseup.net/code/issues/12689

redshiftzero:
[...]

PR #1804, opened 2017-06-14
doc: add a note about the GPG keyserver being unreachable

dachary:

### DIFF
--- a/docs/set_up_admin_tails.rst
+++ b/docs/set_up_admin_tails.rst
@@ -66,6 +66,15 @@ First, download and verify the **SecureDrop Release Signing Key**.
           copy-pasting this command, we recommend you double-check you have
           entered it correctly before pressing enter.
 
+.. note:: If the command hangs and evenutally fails with a timeout, it
+          could indicate the default GPG key server is down or
+          unreachable. As a workaround, another keyserver can be
+          specified by adding the --keyserver option to the
+          gpg --recv-key command. In our experience, the SKS keyserver
+          pool is usually a reliable alternative, so try re-running
+          the command with ``--keyserver
+          hkp://pool.sks-keyservers.net``.
+
 When passing the full public key fingerprint to the ``--recv-key`` command, GPG
 will implicitly verify that the fingerprint of the key received matches the
 argument passed.


fowlslegs:
I'm more inclined to add a note including a workaround (viz., instructions on specifying a high-availability HKPS keyserver). I'd rather give admins an immediate fix than send them to go comment on a bug tracker issue that may be closed at any point in the future.

dachary:
@fowlslegs that makes sense. I'll look for a workaround.

fowlslegs:
@dachary The first comment @redshiftzero posted suffices `gpg --keyserver hkp://pool.sks-keyservers.net --recv-key "2224 5C81 E3BA EB41 38B3 6061 310F 5612 00F4 AD77"` really.

In the conception/ threat model of PGP, keyservers are not supposed to be inherently trusted. It actually was a problem in the past that GPG used to trust that the key you received was the one you requested. Only after Adam Langley reported this issue a few years back did GPG fix this: the client now checks the public key received matches the fingerprint specified with the `--recv-key` operation.

If we trust that this verification process works, then the real threat now from "untrusted" keyservers is that they could respond to a client with malformed data that exploits some GPG 0-day. This is not something I think we should be particularly concerned about (we need to at some level trust the tools we use). Further, as everything in Tails is routed over Tor, IP enumeration of SD key lookups is also not a worry.

This said, as I mentioned above I still think we should use a HKPS server, simply because it provides that extra measure of security, and there are numerous well-known, high-availability HKPS keyservers (unfortunately I'm not aware of any onion service keyservers that haven't had periodic uptime problems). Tails, IIRC, used to include the certificate for the SKS HKPS keyserver pool. I'll try to look into if that's still there.

fowlslegs:
https://github.com/freedomofpress/securedrop/issues/520#issuecomment-150511234 is relevant.

fowlslegs:
Sorry, adding more as I look further. Apparently, they are using an experimental OnionBalance-based SKS pool (https://labs.riseup.net/code/projects/tails/repository/revisions/3c68e5ff4cc0dd38dad2e69e4627f03292d834e1) and they have removed the SKS key they used to store under `/usr/local/etc/ssl/certs/hkps.pool.sks-keyservers.net.pem` (https://labs.riseup.net/code/projects/tails/repository/revisions/345176e5a265047c8341cafd16194c256cc70b8e).

The good news is that

> Since gnupg2 (2.1.15-9), "If no keyserver is explicitly configured, dirmngr will use the built-in default of hkps://hkps.pool.sks-keyservers.net" (dirmngr(8)). So perhaps we can drop our custom settings.

-- https://labs.riseup.net/code/issues/11905.

So, if we can confirm that `gpg --keyserver hkps://pool.sks-keyservers.net --recv-key "2224 5C81 E3BA EB41 38B3 6061 310F 5612 00F4 AD77"` works out of the box like I think it will, I think that is our best choice for a workaround for when the experimental OnionBalance-based SKS pool is down.

dachary:
@fowlslegs you're absolutely right ! Does that look better now ?

garrettr:
I think this is the consensus of the comments on this issue so far, but just to be clear: We should prefer the Tails defaults for gpg wherever possible, since they customize the default GPG configuration (as @fowlslegs pointed out) with the goal of making common tasks (like key lookups) more secure/anonymous.

I do think it is good to provide a suggestion for an alternative in case the Tails defaults do not work as expected, although I would also say that the defaults have changed significantly in Tails 3 and may now be more reliable. If the defaults are more reliable, then suggesting an alternative may be unnecessary.

garrettr:
> If the defaults are more reliable, then suggesting an alternative may be unnecessary.

Ah, but I see now that this was all prompted by #1797, so clearly the new default are *not* more reliable, and suggesting an alternative remains necessary 😅 .

docs/set_up_admin_tails.rst
```diff
@@ -66,6 +66,11 @@ First, download and verify the **SecureDrop Release Signing Key**.
           copy-pasting this command, we recommend you double-check you have
           entered it correctly before pressing enter.
 
+.. note:: If the command hangs and evenutally fails with a timeout, it
+          could indicate the GPG key server is down or unreachable. As
```
garrettr: `s/GPG key server/default GPG key server` is slightly clearer

docs/set_up_admin_tails.rst
```diff
@@ -66,6 +66,11 @@ First, download and verify the **SecureDrop Release Signing Key**.
...
+          a workaround, another keyserver can be specified: for
+          instance ``--keyserver hkp://pool.sks-keyservers.net``.
```
garrettr: I would add a little more detail to this sentence so to try to make it clearer to SecureDrop administrators who may be unfamiliar with GPG, e.g: "As a workaround, another keyserver can be specified by adding the `--keyserver` option to the `gpg --recv-key` command. In our experience, the SKS keyserver pool is usually a reliable alternative, so try re-running the command with `--keyserver hkp://pool.sks-keyservers.net`".

fowlslegs:
I'm inclined to label this as blocked for now, and to wait to see how Tails responds further in that ticket. If they decide to stick with the current setup, it looks like due to changes in GPG the workaround in case of the OnionBalance SKS pool being down will be a little more involved (might have to edit `~/.gnupg/dirmngr.conf`).

garrettr:
I can reproduce @fowlslegs issue from https://github.com/freedomofpress/securedrop/pull/1804#pullrequestreview-44147845, `gpg --keyserver hkp://pool.sks-keyservers.net --recv-key "2224 5C81 E3BA EB41 38B3 6061 310F 5612 00F4 AD77"` fails for me consistently on Tails 3 stable.

garrettr:
FWIW, using the secure `hkps` pool works for me consistently on Tails 3: `gpg --keyserver hkps://hkps.pool.sks-keyservers.net --recv-key "2224 5C81 E3BA EB41 38B3 6061 310F 5612 00F4 AD77"`.

fowlslegs:
I tried using the SKS HKPS keyserver as well earlier and ran into the same problem. It doesn't seem like changes to relevant config files have been made since January, so it doesn't seem like there should be a discrepancy between Tails 3 beta 4 and Tails 3 stable.

Note: Garrett is testing with virtualized Tails 3, while I've been testing on hardware. While this seems unlikely to be the explanation for this discrepancy, we should confirm that on Tails 3 stable, running on real hardware, that the command `gpg --keyserver hkps://hkps.pool.sks-keyservers.net --recv-key "2224 5C81 E3BA EB41 38B3 6061 310F 5612 00F4 AD77"` reliably works.

docs/set_up_admin_tails.rst
dachary: Updated as suggested, thanks !

dachary:
this is all quite confusing indeed (i.e. the fact that hkps://hkps.pool.sks-keyservers.net fails in some contexts and hkp://hkps.pool.sks-keyservers.net in other contexts...). I'm curious to read what tails maintainers have to say about that...

dachary:
Let's re-open if this surfaces again.

fowlslegs:
@garrettr Thoughts on re-opening? Not sure I agree we should wait until this problem potentially surfaces again.

fowlslegs:
Also, @freedomofpress/securedrop-maintainers

garrettr:
I agree with @fowlslegs that it would be good to resolve this for 0.4, because based on recent experiences (e.g. #1797) it seems like it might be somewhat likely for folks to encounter issues with `gpg --recv-key` and the new default keyserver setup in Tails 3.

dachary:
@garrettr thanks for working on this, closing in favor of you pull request :-) #1842

PR #1791, opened 2017-06-11
tests: source: unit test for delete_all

dachary:
Signed-off-by: Loic Dachary <loic@dachary.org>

PR #1808, opened 2017-06-14
db: add unit test for Journalist.throttle_login

dachary:
## Status

Ready for review

## Description of Changes

Adds unit tests for Journalist.throttle_login

A minor change to the db Journalist class is also done to expose the constants used by throttle_login

## Testing

pytest -v -s -k test_throttle_login tests/test_db.py

### DIFF
--- a/securedrop/db.py
+++ b/securedrop/db.py
@@ -368,11 +368,11 @@ class Journalist(Base):
                     return True
             return False
 
+    _LOGIN_ATTEMPT_PERIOD = 60  # seconds
+    _MAX_LOGIN_ATTEMPTS_PER_PERIOD = 5
+
     @classmethod
     def throttle_login(cls, user):
-        _LOGIN_ATTEMPT_PERIOD = 60  # seconds
-        _MAX_LOGIN_ATTEMPTS_PER_PERIOD = 5
-
         # Record the login attempt...
         login_attempt = JournalistLoginAttempt(user)
         db_session.add(login_attempt)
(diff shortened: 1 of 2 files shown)


fowlslegs:
> `pytest -v -s -k test_throttle_login tests/test_db.py`

Another way to write this is `pytest -v tests/test_db.py::TestDatabase::test_throttle_login`. Avoids regex matching as I prefer. Why do you pass the `-s`?

securedrop/db.py
```diff
@@ -368,26 +368,26 @@ def verify_token(self, token):
                     return True
             return False
 
+    _LOGIN_ATTEMPT_PERIOD = 60  # seconds
+    _MAX_LOGIN_ATTEMPTS_PER_PERIOD = 5
```
fowlslegs: I think it is slightly nicer to set these as class attributes as you've done. Good call!

**Edit:** I see exactly why now :smile:

securedrop/tests/test_db.py
```diff
@@ -78,3 +78,10 @@ def test_journalist_string_representation(self):
     def test_source_string_representation(self):
         test_source, _ = db_helper.init_source()
         test_source.__repr__()
+
+    def test_throttle_login(self):
+        journalist, _ = db_helper.init_journalist()
+        for _ in range(Journalist._MAX_LOGIN_ATTEMPTS_PER_PERIOD):
+            Journalist.throttle_login(journalist)
+        with self.assertRaises(LoginThrottledException):
```
fowlslegs: `assertRaisesRegexp` would be even better.

fowlslegs:
I would merge this, but I don't know how to run CircleCI. If anyone else wants to :+1:.

garrettr:
> I would merge this, but I don't know how to run CircleCI. If anyone else wants to 👍.

@fowlslegs The CircleCI build runs automatically, but since we're currently using their free plan, *all* of our CI jobs (from all of our GitHub repositories) must share a small set of underlying containers. That, plus the fact that the SecureDrop CircleCI tests take a long time anyway, means you just have to wait a while for the CircleCI results to come in.

You can see more detailed CircleCI progress on their dashboard: https://circleci.com/dashboard. If you don't already have an account, just sign up with your GitHub account, everything clicks neatly into place after that in my experience.

fowlslegs:
This PR is from a fork so CircleCI doesn't run automatically for security reasons. I am already set up with CircleCI, but am not an admin there, which is required to flip whatever setting prevents CircleCI from running on PRs from forked repos. I already explained the deal to @dachary [here](https://github.com/freedomofpress/securedrop/pull/1791#issuecomment-307889658), hence why I was quite terse above.

dachary:
@fowlslegs 

> Why do you pass the -s

in this instance it does not make a difference. When a run is long I find it helpful to have the output to contemplate. Not strictly necessary ;-)

securedrop/tests/test_db.py
dachary: Since this exception is unique to this error condition, I did not think of using assertRaisesRegexp. Could you please explain why it would be better ?

fowlslegs: You're not wrong. I would use `assertRaisesRegexp` as a defensive programming measure in case`LoginThrottledException` does find a second use in the future. The reason not to wait until then is just to be careful, so you don't forget if/ when `LoginThrottledException` finds a second use.

I realize in this specific example it's highly unlikely, but I just have a general mindset of defensive programming and making test assertions as comprehensive as practical. I approved this PR because it definitely is good as it was merged, and making this change was not worth burning another review cycle on.

dachary: thanks for explaining :-)

PR #1812, opened 2017-06-15
tests: add unit test for generate_unique_codename

dachary:


garrettr:
This PR reminds me that we should double-check the chosen value for `Source.MAX_CODENAME_LEN` in light of #1452, especially because the average word length in the EFF "long" wordlist is [longer](https://www.eff.org/deeplinks/2016/07/new-wordlists-random-passphrases) than the average word length in the original Diceware list:

> The result is our own list of 7,776 words [.txt] suitable for use in dice-generated passphrases. The words in our list are longer (7.0 characters) on average, than Reinhold's Diceware list (4.3 characters). This is a result of banning words under 3 characters as well as prioritizing familiar words over short but unusual words.

garrettr:
Ok, the longest word in the new wordlist is 9 characters long. In addition, #1521 removed the source's ability to choose the length of their codename: 7 words is now the only option. So the longest possible codename (including spaces) is `9*7+6 = 69`, so the current `Source.MAX_CODENAME_LEN` is still ok.

We should probably update the comment in `generate_unique_codename` (starts with "The maximum length of a word in the wordlist is 6 letters") to reflect the current state of affairs.

securedrop/tests/test_source.py
```diff
@@ -318,3 +318,24 @@ def test_login_with_overly_long_codename(self, mock_hash_codename):
             self.assertFalse(mock_hash_codename.called,
                              "Called hash_codename for codename w/ invalid "
                              "length")
+
+    @patch('source.app.logger.warning')
+    def test_generate_unique_codename(self, mock_warning):
+        codename = source.generate_unique_codename()
+        self.assertGreater(len(codename), 0)
+        mock_warning.assert_not_called()
```
garrettr: I don't see the benefit of including this test. It is guaranteed to succeed unless changes are made to the wordlist, but it is also not a good test for catching potential issues caused by changes to the wordlist.

For example, if we added a word that was longer than `Source.MAX_CODENAME_LEN` to the wordlist, we would want to avoid that, or reconsider the maximum codename length DoS protections (see 3102925d369c5110e0c50a80684814a5efa30011 for context). If we shipped a release with such a change to the wordlist, then we potentially be in danger of hitting the warning condition tested here; however, this test is very unlikely to catch such a potentially problematic change, because the probability of it generating a problematic phrase at random is very small.

securedrop/tests/test_source.py
```diff
@@ -318,3 +318,24 @@ def test_login_with_overly_long_codename(self, mock_hash_codename):
...
+    @patch('source.app.logger.warning')
+    def test_generate_unique_codename_too_long(self, mock_warning, mock_genrandomid):
+        self.too_long = True
+        def genrandomid(num_words):
+            if self.too_long:
+                self.too_long = False
+                return 'O' * Source.MAX_CODENAME_LEN * 2
+            else:
+                return 'not so random'
+        mock_genrandomid.side_effect = genrandomid
```
garrettr: You could simplify the mock by using an [iterator](https://docs.python.org/dev/library/unittest.mock.html#unittest.mock.Mock.side_effect) instead of a function for `side_effect`, e.g.

```py
    @patch('source.crypto_util.genrandomid')
    @patch('source.app.logger.warning')
    def test_generate_unique_codename_too_long(self, mock_warning, mock_genrandomid):
        mock_genrandomid.side_effect = [
            'O' * Source.MAX_CODENAME_LEN * 2,
            'not too long'
        ]
        codename = source.generate_unique_codename()
        mock_warning.assert_called()
        self.assertEqual(codename, 'not too long')
```

dachary:
> We should probably update the comment in generate_unique_codename

Updated in https://github.com/freedomofpress/securedrop/pull/1813/files

securedrop/tests/test_source.py
dachary: My intent was to run the code path used most of the time and assert it does not display any warning. This should never fail and if it does, we know there is a regression. If you still think this is not worth testing I'll remove it.

securedrop/tests/test_source.py
dachary: Nice ! Thanks for the tip :-)

fowlslegs:
Looking at the corresponding app code part of this PR, I _really_ didn't like how `source.generate_unique_codename()` was working, so I created this WIP branch: https://github.com/freedomofpress/securedrop/commit/41e01fcb185fad9a146c527a9ab9fb2e561ae9c1 . I'm thinking that instead of writing tests to verify bad application behavior, we should fix the application behavior, and then write tests for that.

So I'm board with merging #1813, but I think I'd rather see this PR closed in favor of one based on the branch above (or if anyone has other suggestions of how to better fix this situation).

PR #1813, opened 2017-06-15
source: set Source.NUM_WORDS to 7

dachary:
## Status

Ready for review

## Description of Changes

Instead of keeping the constant in generate_unique_codename
argument (which is never called with anything but the default), create a
NUM_WORDS constant (next to MAX_CODENAME_LEN). The two are related since
MAX_CODENAME_LEN is a safeguard that assumes a given number of words and
a given average word length.

Also modify the test that had the same constant accordingly.

### DIFF
--- a/securedrop/db.py
+++ b/securedrop/db.py
@@ -86,6 +86,7 @@ class Source(Base):
     interaction_count = Column(Integer, default=0, nullable=False)
 
     # Don't create or bother checking excessively long codenames to prevent DoS
+    NUM_WORDS = 7
     MAX_CODENAME_LEN = 128
 
     def __init__(self, filesystem_id=None, journalist_designation=None):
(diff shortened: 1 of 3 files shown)


dachary:
also updated the comment to reflect the reality described at https://github.com/freedomofpress/securedrop/pull/1812#issuecomment-308837950

garrettr:
> also updated the comment to reflect the reality described at #1812 (comment)

excellent, thanks. this is ready to merge once CI passes.

garrettr:
Note: the CircleCI tests were not automatically run on this PR. They are currently not being run on any PR's submitted by external contributors because there are security concerns related to automatically running arbitrary code on our CI infrastructure.

We are working on improving this situation so we can get better test coverage on external contributions (see #1826). In the meantime, I will be applying the steps from the "Short-term mitigation" section of the description of #1826.

Since this PR only modifies the web application code, it is unlikely to break the configuration tests, and the fact that the unit and functional test suites are passing on Travis CI should be sufficient due diligence. I am going to merge this without running the CircleCI tests.

PR #1815, opened 2017-06-15
tests: unit test for source.errorhandler(404)

dachary:


securedrop/tests/test_source.py
```diff
@@ -28,6 +28,11 @@ def setUp(self):
     def tearDown(self):
         utils.env.teardown()
 
+    def test_page_not_found(self):
+        response = self.client.get('/UNKNOWN')
+        self.assertEqual(response.status_code, 404)
+        self.assertIn("locate what you requested", response.data)
```
garrettr: I would prefer this include the full custom error message ("Sorry, we couldn't locate what you requested."), so it's clearer exactly what this test is trying to achieve: that is, that a 404 error results in a response that includes the content of our custom 404 error template (`notfound.html`). It also might be worth adding a comment or docstring to help clarify the intent of this test.

securedrop/tests/test_source.py
```diff
@@ -28,6 +28,11 @@ def setUp(self):
     def tearDown(self):
         utils.env.teardown()
 
+    def test_page_not_found(self):
+        response = self.client.get('/UNKNOWN')
+        self.assertEqual(response.status_code, 404)
```
fowlslegs: Flask-Testing provides some handy methods for the web application context. E.g., you could use `self.assert404(response)` here.

dachary:
@garrettr thanks for the review :-) The suggested comment was added and the error message looked for in full.

securedrop/tests/test_source.py
```diff
@@ -28,6 +28,11 @@ def setUp(self):
...
+        self.assertIn("locate what you requested", response.data)
+
```
fowlslegs: To add to what Garrett was saying, you could also add the line
```
        self.assertTemplateUsed('notfound.html')
```
to ensure the correct template is being called.

garrettr: This is a good suggestion! That would be an even better test.

dachary:
even better, very convenient helpers :-)

Issue #1821, opened 2017-06-15
Remove web application pidfiles


garrettr:
The original `./manage.py` commands to manage starting and stopping web application servers in the development environment used pidfiles to manage the lifetime of those process (see #458). When this function of `./manage.py` was rewritten in #1016, the pidfiles became redundant, but the code related to them was not cleaned up.

The following code should be removed:

1. `write_pidfile` in both `source.py` and `journalist.py`
2. `config.{SOURCE,JOURNALIST}_PIDFILE`

Note that the `WORKER_PIDFILE` is still necessary and should not be touched.

Note that the production deployment story for changes to the web application configuration (`config.py`) is currently fraught. There are several potential issues to look out for, although I cannot recall them off the top of my head. The production deployment aspect of this issue deserves careful consideration before merge/deployment.

dachary:
I did the easy part of the work at https://github.com/freedomofpress/securedrop/pull/1822 and I would appreciate pointers to verify / do the right thing after modifying config.py. Thanks !

dachary:
@garrettr existing installations may have pid files but they are not used. The config.py.example now contains no *_PIDFILE variables and the next time it is used by install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml and copied to config.py that will create a config.py file without *_PIDFILE. Existing config.py will still have *_PIDFILE and they will not be removed, they will just not be used anywhere. I did not find tests verifying all config.py.example variables are in the config.py or any other kind of consistency check of the kind. The existing pidfiles are in the /tmp directory and will be cleaned up after the next reboot.

You mentionned that "the production deployment story for changes to the web application configuration (config.py) is currently fraught. " and I probably missed something. Would you be so kind as to point me to a file or an example so I can better understand what is at stake ?

fowlslegs:
> You mentionned that "the production deployment story for changes to the web application configuration (config.py) is currently fraught. " and I probably missed something. Would you be so kind as to point me to a file or an example so I can better understand what is at stake ?

Hopefully you'll get an answer here or in response to https://github.com/freedomofpress/securedrop/pull/1822#issuecomment-309488035.

dachary:
I believe this has been addressed with https://github.com/freedomofpress/securedrop/pull/1822, please re-open if you disagree.

PR #1822, opened 2017-06-15
source,journalist: pidfiles are not used anymore

dachary:
## Status

Ready for Review

## Description of Changes

Remove the dead code and verify the pid files are not created.

## Testing

In addition to running config tests ( testinfra/app/test_appenv.py ),  the journalist and source app should be run from the development environment to verify they start and manually check there is no /tmp/*.pid file.

## Deployment

1. Upgrading existing production instances: the existing `{SOURCE,JOURNALIST}_PID` variables in config.py remain and are a noop.
2. New installs: the config.py copied over from config.py.example does not have `{SOURCE,JOURNALIST}_PID` but the installation procedure is otherwise unchanged since it does not mention these variables.

### DIFF
--- a/securedrop/config.py.example
+++ b/securedrop/config.py.example
@@ -28,8 +28,6 @@ WORD_LIST = os.path.join(SECUREDROP_ROOT, 'wordlist')
 NOUNS = os.path.join(SECUREDROP_ROOT, 'dictionaries/nouns.txt')
 ADJECTIVES = os.path.join(SECUREDROP_ROOT, './dictionaries/adjectives.txt')
 
-JOURNALIST_PIDFILE = "/tmp/journalist.pid"
-SOURCE_PIDFILE = "/tmp/source.pid"
 WORKER_PIDFILE = "/tmp/securedrop_worker.pid"
 
 # "head -c 32 /dev/urandom | base64" for constructing public ID from source codename
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -738,13 +738,6 @@ def flag():
                            codename=g.source.journalist_designation)
 
 
-def write_pidfile():
-    pid = str(os.getpid())
-    with open(config.JOURNALIST_PIDFILE, 'w') as fp:
-        fp.write(pid)
-
-
 if __name__ == "__main__":  # pragma: no cover
-    write_pidfile()
     debug = getattr(config, 'env', 'prod') != 'prod'
     app.run(debug=debug, host='0.0.0.0', port=8081)
--- a/securedrop/source.py
+++ b/securedrop/source.py
@@ -444,13 +444,6 @@ def internal_error(error):
     return render_template('error.html'), 500
 
 
-def write_pidfile():
-    pid = str(os.getpid())
-    with open(config.SOURCE_PIDFILE, 'w') as fp:
-        fp.write(pid)
-
-
 if __name__ == "__main__":  # pragma: no cover
-    write_pidfile()
     debug = getattr(config, 'env', 'prod') != 'prod'
     app.run(debug=debug, host='0.0.0.0', port=8080)
--- a/testinfra/app/test_appenv.py
+++ b/testinfra/app/test_appenv.py
@@ -21,6 +21,11 @@ def test_app_wsgi(File, Sudo):
         assert f.contains("^import logging$")
         assert f.contains("^logging\.basicConfig(stream=sys\.stderr)$")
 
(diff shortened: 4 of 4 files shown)


conorsch:
Good changes to `config.py.example`. We don't have a good story for updating that config file in-place, even with subsequent playbook runs. A regression test in `testinfra/app/` to ensure that PID files are absent would be great.

garrettr:
@dachary ^ That pull request template is meant to be filled out by you, in case that wasn't clear.

dachary:
@garrettr it was clear: very convenient template. I filed the first part "Work in progress" and saved the rest for later ;-)

dachary:
@conorsch will do. First opportunity to dive into that part of the tests. Yeah !

garrettr:
Added the "pending additional work" label since this should *not* be merged until it addresses the potential production deployment issues described in #1821.

dachary:
I don't know ansible and not able to complete this immediately. Closing it for now so someone else can take over.

dachary:
I'll give it a shot with my newly acquired ansible knowledge.

dachary:
@conorsch I added something in testinfra/app at https://github.com/freedomofpress/securedrop/pull/1822/commits/d42664de279e031bdcb73b0e6511016fbaedf7b2 . Is it what you expected ?

fowlslegs:
> @conorsch I added something in testinfra/app at d42664d . Is it what you expected ?

That should work, but only when the application is running. Not a terribly important regression test to have IMO, so I think that's fine as is.

@conorsch is there a good thread with more info on the problems with `config.py` we've had thus far and how we've endeavored to solve them. I feel like most of that conversation has happened IRL.

garrettr:
> @conorsch is there a good thread with more info on the problems with config.py we've had thus far and how we've endeavored to solve them. I feel like most of that conversation has happened IRL.

@fowlslegs Here is one relevant past issues https://github.com/freedomofpress/securedrop/issues/1328. This only affected development environments, but such potential breakage is worth keeping in mind and avoiding if possible.

To understand the issues with updating `config.py` in production, you need to grok `install_files/ansible-base/roles/tasks/initialize_securedrop_app.yml`. @conorsch's comments are very helpful.

So far, I'm pretty the only way we've "endeavored to solve them" is to not update config.py in production. It is probably worth filing a separate issue/PR to handle "Reliably updating the SecureDrop configuration."

In the long term, I think the informal consensus is that we'd like to move away from using Python files for configuration management (for a number of reasons). However, I think we're probably stuck with it for at least the next release or so.

dachary:
@garrettr thanks for the pointers. In the case of this change is it OK to just leave the `{JOURNALIST,SOURCE}_PID` variables in existing config.py files ? Or should they be stripped by ansible when found ? It is not as pretty as it could be :-)

dachary:
@fowlslegs thanks for the review :-)

dachary:
Updated the PR description, I think it is ready for review now.

dachary:
$ ./testinfra/test.py app-staging

<pre>
...
[gw5] PASSED testinfra/app/test_appenv.py::test_pidfile[ansible://app-staging] 
...
</pre>

fowlslegs:
I don't think we need to care about cleaning the vars from older installs.

garrettr:
Discussed this in a meeting today. As it stands right now, this PR makes no attempt to clean up the defunct `{SOURCE,JOURNALIST}_PIDFILE` variables from existing versions of `config.py` in currently deployed development or production environments; however, we agreed that we cannot see how leaving these defunct variables in the SecureDrop configuration could cause any problems.

We have various open issues intended to address the current shortcomings with maintaining and updating the SecureDrop application configuration, but we agreed that improving the configuration maintenance story is a significant enough undertaking that it should wait until after the planned 0.4 release.

In light of these two points, I agree that it is acceptable to merge this PR for the 0.4 release.

garrettr:
In the context of #1826, this PR modifies the testinfra tests, so I believe it is necessary to ensure that the CircleCI test suite passes before merging it. I will follow the procedure outlined in Option 2 of the "Short-term mitigation" section of #1826 to do so.

fowlslegs:
Closing in favor of #1864.

garrettr:
Re-opening to conform with the protocol described in Option 2 of #1826.

fowlslegs:
Intermittent CircleCI failure. Restarted the build.

garrettr:
All tests are passing on #1864, so this is good to merge.

PR #1823, opened 2017-06-15
Minor refactor journalist 2fa

fowlslegs:
## Status

Ready for review.

## Description of Changes

I found it pretty purposeless to introduce a second object reference to the user object, so I got rid of that. I also think it's generally best to avoid this sort of aliasing, which I believe makes it easier to make mistakes.

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -334,16 +334,14 @@ def admin_delete_user(user_id):
 @app.route('/account', methods=('GET', 'POST'))
 @login_required
 def edit_account():
-    user = g.user
-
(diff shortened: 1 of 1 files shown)


fowlslegs:
Spurious Travis failure, but I figured out the cause. PR coming to fix that. In the meantime I've just restarted the build. Chances are high it won't run into that intermittent bug again.

fowlslegs:
Oh yeah, had to rebase and force push anyway.

dachary:
These are only cosmetic changes and indeed easier to read. :+1: (is there another way for me to express a review ?)

Issue #1816, opened 2017-06-15
Testinfra tests are skipped in Travis


conorsch:
# Bug

## Description

All of the config tests in the Travis CI run are marked as "skipped." There are a few config tests that are not relevant on the Travis host, since that host is a combination of an Ansible controller and the development VM, given the localhost provisioning strategy, but we don't want to skip _everything_.

While working on #1736 and running the config tests locally, I discovered that #1770 broke the development VM config tests by upgrading Python dependencies without touching the config tests. Travis CI should have caught that, but didn't. Note that the pip checks may be one of the config tests we need to skip in Travis, due to the insufficient separation of Python library environments. Still, worth a shot.

## Steps to Reproduce

1. Commit a breaking change in a feature branch.
2. Confirm that `./testinfra/test.py development` fails locally as a result of the breaking change.
3. Push feature branch to remote and see the Travis checks fail.

## Expected Behavior

Config tests run in Travis. If not all of them, then most of them.

## Actual Behavior

No config tests run in Travis. All are marked as "SKIPPED."

## Comments

The `testinfra/test.py` wrapper script has gotten ugly. The solution is either to add a "travis" target to that script, or fix the handling of the "development" target to execute against localhost when running on Travis. It's also possible to execute testinfra directly, bypassing the wrapper, since the default behavior of testinfra is to execute against localhost.

msheiny:
Urggggghh good catch @conorsch

dachary:
I tried this on the vagrant development vm. But it's missing the "testinfra" script. Not sure what I'm missing here.
<pre>
vagrant@development:/vagrant$ ./testinfra/test.py development
Running Testinfra suite against 'development'...
Target roles:
    - testinfra/app-code
    - testinfra/development
testinfra -vv -n auto --connection ansible --ansible-inventory .vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory --hosts development testinfra/app-code testinfra/development
Traceback (most recent call last):
  File "./testinfra/test.py", line 135, in <module>
    run_testinfra(target_host)
  File "./testinfra/test.py", line 132, in run_testinfra
    subprocess.check_call(testinfra_command)
  File "/usr/lib/python2.7/subprocess.py", line 535, in check_call
    retcode = call(*popenargs, **kwargs)
  File "/usr/lib/python2.7/subprocess.py", line 522, in call
    return Popen(*popenargs, **kwargs).wait()
[...]

PR #1824, opened 2017-06-15
Reorganizes top-level Ansible vars for cleaner variable precedence

### LINKED ISSUE #1736: Fix intermittent custom_notification test failure
The `custom_notification` test is occasionally failing in the staging environment. When I tested #1731 it did not fail (using docker and AWS) but I have seen it fail with staging provisioned locally using vagrant. @conorsch mentioned that this is a test failure only occurring in the vagrant environment. We should resolve this for 0.4.

conorsch:
## Status

Work in progress.

## Description of Changes

Fixes #1736, which pointed out a problem caused by our use of nested groups in the Vagrant Ansible config. Takes most of the top-level vars and shoves them into role defaults, which simplifies [variable precedence](https://docs.ansible.com/ansible/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable) and makes it a clear-cut case where and how to override elements where appropriate.

Although the diff looks substantial, note that the vars have _moved_ to a more convenient location. In a few cases, a few vars were removed, only if they were never included anywhere (and were therefore left over from previous versions of the Ansible config). In other cases, vars were used by unnecessarily providing too much indirection, such as `non_default_securedrop_user` which just pointed to `securedrop_user`. Removed the former and preserved the latter. The functionality change that allows the test to pass is the removal of the nested group memberships within the `Vagrantfile`.

## Testing

Run the application test suite _locally_ in Vagrant and confirm that 100% passes. You'll need to do so on a fresh staging build, since bad variable precedence from a previous build likely poisoned your config.py.

## Deployment

Affects the dev env. Made sure not to alter the way variables land on different hosts, just cleaned up the layout so it's more intuitive and consistent with best practices. We can now properly use host var overrides where appropriate and be confident the override will take effect (#1736 showed that sometimes it didn't).


conorsch:
Downgraded to "work in progress" since the Travis CI is failing. There are related changes to improve Travis (#1816) but I wanted to submit them separately, rather than bolt on to this for review—even though that's how I wrote the changes.

conorsch:
I mistakenly submitted this from a fork, so I'm going to close and resubmit from a feature branch on the origin.

PR #1829, opened 2017-06-16
Explicitly select bs4 HTML parser

### LINKED ISSUE #1820: Fix bs4 pytest HTML parser warning
# Bug

## Description

I spotted this pytest warning on a recent Travis build. Seems like a quick fix to get this resolved.

```
test_integration.py::TestIntegration::test_delete_collection
  /usr/local/lib/python2.7/dist-packages/bs4/__init__.py:181: UserWarning: No parser was explicitly specified, so I'm using the best available HTML parser for this system ("html5lib"). This usually isn't a problem, but if you run this code on another system, or in a different virtual environment, it may use a different parser and behave differently.
  
  The code that caused this warning is on line 11 of the file /usr/local/bin/pytest. To get rid of this warning, change code that looks like this:
  
   BeautifulSoup(YOUR_MARKUP})
  
  to this:
  
   BeautifulSoup(YOUR_MARKUP, "html5lib")
  
    markup_type=markup_type))
```

fowlslegs:
## Status

## Description of Changes

Fixes #1820.

Adding this to get rid of the pytest warning the latest version of bs4 throws if you do not specifically select your HTML parser.

### DIFF
--- a/securedrop/source_templates/lookup.html
+++ b/securedrop/source_templates/lookup.html
@@ -70,7 +70,8 @@
           <input name="csrf_token" type="hidden" value="{{ csrf_token() }}"/>
           <input type="hidden" name="reply_filename" value="{{ reply.filename }}" autocomplete="off"/>
           <a href="#confirm-delete-{{ reply.filename }}" class="delete">
-          <img src="{{ url_for('static', filename='i/font-awesome/trash-black.png') }}" width="20px" height="20px">
+            <img src="{{ url_for('static', filename='i/font-awesome/trash-black.png') }}" width="20px" height="20px">
+          </a>
           <div id="confirm-delete-{{ reply.filename }}" class="confirm-prompt">
             <p>Delete this reply?
               <a href="#delete">Cancel</a>
--- a/securedrop/requirements/test-requirements.in
+++ b/securedrop/requirements/test-requirements.in
@@ -2,6 +2,7 @@ beautifulsoup4
 blinker
 coveralls
 Flask-Testing
+html5lib
 mock
 pip-tools
 py
--- a/securedrop/requirements/test-requirements.txt
+++ b/securedrop/requirements/test-requirements.txt
@@ -16,18 +16,20 @@ first==2.0.1              # via pip-tools
 flask-testing==0.6.2
 flask==0.12.2             # via flask-testing
 funcsigs==1.0.2           # via mock
+html5lib==0.999999999
 idna==2.5                 # via requests
 itsdangerous==0.24        # via flask
 jinja2==2.9.6             # via flask
 markupsafe==1.0           # via jinja2
 mock==2.0.0
-pbr==3.0.1                # via mock
+pbr==3.1.1                # via mock
 pip-tools==1.9.0
 py==1.4.34
 pytest-cov==2.5.1
-pytest==3.1.1
-requests==2.17.3          # via coveralls
+pytest==3.1.2
+requests==2.18.1          # via coveralls
 selenium==2.53.6
-six==1.10.0               # via mock, pip-tools
+six==1.10.0               # via html5lib, mock, pip-tools
 urllib3==1.21.1           # via requests
+webencodings==0.5.1       # via html5lib
 werkzeug==0.12.2          # via flask
--- a/securedrop/tests/test_integration.py
+++ b/securedrop/tests/test_integration.py
@@ -80,7 +80,7 @@ class TestIntegration(unittest.TestCase):
         rv = self.journalist_app.get('/')
         self.assertEqual(rv.status_code, 200)
         self.assertIn("Sources", rv.data)
-        soup = BeautifulSoup(rv.data)
+        soup = BeautifulSoup(rv.data, 'html5lib')
 
         # The source should have a "download unread" link that says "1 unread"
         col = soup.select('ul#cols > li')[0]
@@ -90,7 +90,7 @@ class TestIntegration(unittest.TestCase):
         col_url = soup.select('ul#cols > li a')[0]['href']
         resp = self.journalist_app.get(col_url)
         self.assertEqual(resp.status_code, 200)
-        soup = BeautifulSoup(resp.data)
+        soup = BeautifulSoup(resp.data, 'html5lib')
         submission_url = soup.select('ul#submissions li a')[0]['href']
         self.assertIn("-msg", submission_url)
         span = soup.select('ul#submissions li span.info span')[0]
@@ -105,7 +105,7 @@ class TestIntegration(unittest.TestCase):
         # delete submission
         resp = self.journalist_app.get(col_url)
         self.assertEqual(resp.status_code, 200)
-        soup = BeautifulSoup(resp.data)
+        soup = BeautifulSoup(resp.data, 'html5lib')
         doc_name = soup.select(
             'ul > li > input[name="doc_names_selected"]')[0]['value']
         resp = self.journalist_app.post('/bulk', data=dict(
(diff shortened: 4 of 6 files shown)


garrettr:
Useful [notes](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#installing-a-parser) from the BeautifulSoup4 documentation on selecting a parser.

garrettr:
@fowlslegs Ah, I see you added the explicit dependency for `html5lib` while I was writing my review. 👍

Any thoughts on whether `html.parser` might be a better choice than `html5lib`? Now that you've added the explicit dependency for `html5lib`, I think this PR satisfies my review's request for changes, but I am still slightly perturbed by the fact that some of our tests appear to depend on `html5lib`'s "extremely lenient" parsing.

fowlslegs:
I thought `html5lib` would be the best choice because it "parses pages the same way a web browser does" according to the docs you linked.

According to [the html5lib-python README](https://github.com/html5lib/html5lib-python):

> html5lib is a pure-python library for parsing HTML. It is designed to conform to the WHATWG HTML specification, as is implemented by all major web browsers.

fowlslegs:
Reading through https://www.crummy.com/software/BeautifulSoup/bs4/doc/#differences-between-parsers, I do agree however that "extremely lenient" may not be ideal in this case, and it might be best to reject the generally good practice of trying to make your tests as close as possible to production.

fowlslegs:
Okay, I think I'm going to redo this with `html.parser` and just fix the HTML to satisfy those failing tests you mentioned. Sound good?

garrettr:
@fowlslegs Sounds good to me! I'm ok with either approach. I would say it's worthwhile for you to attempt to switch to `html.parser`, but timebox it, and if it takes to too long we should just stick with `html5lib`.

fowlslegs:
So getting the tests to pass with `html.parser` required only a single character change. That said, it seems like in a lot of the HTML we've added added a `/` before the closing `>` in order to maintain XHTML compatibility (correct me if I'm wrong here, as I'm just guessing at why these `/`s exist), and I'm not sure we want this.

Question as far as wrapping this PR up goes: do you think I should I make this change everywhere now, or leave this as-is (I'll of course need to amend my commit message before merging) and open a follow-up issue? I'm leaning toward the latter, but want to hear from @garrettr first, who's got the most historical perspective on the repo and might be able to explain what I'm guessing at above.

securedrop/source_templates/lookup.html
```diff
@@ -67,7 +67,7 @@ <h2 class="headline">Get Replies</h2>
       <div class="reply">
         <time class="date" title="{{ reply.date|datetimeformat }}" datetime="{{ reply.date }}">{{ reply.date|datetimeformat(relative=True) }}</time>
         <form id="delete" class="message" method="post" action="/delete" autocomplete="off">
-          <input name="csrf_token" type="hidden" value="{{ csrf_token() }}"/>
+          <input name="csrf_token" type="hidden" value="{{ csrf_token() }}">
           <input type="hidden" name="reply_filename" value="{{ reply.filename }}" autocomplete="off"/>
           <a href="#confirm-delete-{{ reply.filename }}" class="delete">
```
garrettr: Side note, noticed while reviewing: this `<a>` start tag appears to be missing a corresponding closing tag.

garrettr:
> That said, it seems like in a lot of the HTML we've added added a / before the closing > in order to maintain XHTML compatibility (correct me if I'm wrong here, as I'm just guessing at why these /s exist), and I'm not sure we want this.

We are using HTML5 (`base.html` has `<!doctype html>`), so there is no reason to have or attempt to maintain XHTML compatibility. As far as I know, these trailing slashes are just mistakes and aren't being added or maintained intentionally. According to the [HTML5 spec](https://www.w3.org/TR/html5/syntax.html#start-tags):

> Then, if the element is one of the void elements, or if the element is a foreign element, then there may be a single "/" (U+002F) character. This character has no effect on void elements, but on foreign elements it marks the start tag as self-closing.

So this trailing slash on an `<input>` element (which is a void element) is unnecessary, but according to the spec it should be allowed (it's supposed to be ignored). I'm actually surprised that it appears to be causing the failure in `html.parser` - are you sure that's the true cause? Seems like it might be a red herring, and the true cause is some actually invalid HTML in the same neighborhood (e.g. the lack of a closing tag that I pointed out in https://github.com/freedomofpress/securedrop/pull/1829#pullrequestreview-45000839).

> Question as far as wrapping this PR up goes: do you think I should I make this change everywhere now, or leave this as-is (I'll of course need to amend my commit message before merging) and open a follow-up issue?

Definitely make the minimal changes required to get the tests passing with `html.parser`, and file a follow-up issue to address removing the redundant trailing slashes holistically. You might also consider filing a follow-up issue to propose auto-validating HTML as part of our test suite and/or CI pipeline, which could help prevent this kind of issue in the future.

fowlslegs:
Good catch on the missing `</a>` tag. That totally ruined the ability for sources to delete a single reply, and this PR will need to add a functional test to prevent regressions before it's ready for merge.

That said, adding the missing `</a>`, and then reverting the deletion of that slash resulted in the tests failing in the same manner they were before.

fowlslegs:
> are you sure that's the true cause?

I'm not, but I agree there needs to be full understanding before the correct solution can be arrived at. I'll keep looking.

fowlslegs:
I couldn't find anything via manual review, so I tried to use a linter to help, but still haven't been able to figure this one out.

fowlslegs:
Requesting help on this one @freedomofpress/securedrop-maintainers.

fowlslegs:
Relevant: https://github.com/freedomofpress/securedrop/issues/1820#issuecomment-311189486.

fowlslegs:
Checks have been passing. The problem is instead that the change that makes them pass is [this single character diff](https://github.com/freedomofpress/securedrop/pull/1829/files#diff-e87f970bac26dbcd91cd38d562e6f9e0R70). @garrettr and I felt uncomfortable with the fact we don't understand why that's the case, and that's been the hold up--that nobody has been able to determine why.

I'm going to fall back to the original state of this PR, where `html5lib` is being used and also tracked in our dependencies (instead of incidentally being provided by an apt package), and will keep in the commit that adds the missing `</a>` tag. Then I'll open an issue about switching to the more strict `html.parser`.

fowlslegs:
Ah, forgot to update the testinfra tests this time.

fowlslegs:
Back to square on basically lol, but I'm pretty sure checks should pass and consider this merge-able.

fowlslegs:
Alright, so we've now reached the inevitable stage where checks are failing because Python dependencies aren't being upgraded in Travis. Specifically, requests is not being upgraded when the development Ansible role is run because earlier in Travis another version is installed development-requirements. And then the failure to upgrade to the correct version in test-requirements causes testinfra to fail. Still thinking about what to do about it though...

garrettr:
> Good catch on the missing </a> tag. That totally ruined the ability for sources to delete a single reply, and this PR will need to add a functional test to prevent regressions before it's ready for merge.

@fowlslegs How did you test this? I just tested deleting a single reply from the Source Interface in Tor Browser on the develop branch (after confirming that the missing `</a>` tag in question is still missing), and deleting a single reply succeeded without issues. My manual test fits with the automated test results for #1898, which added functional testing for deleting a reply on the source interface, and passed without fixing the missing `</a>` tag. Note that modern web browser's HTML parsers are generally quite good at fixing minor HTML validation issues.

To be clear: I think it's _great_ that this investigation lead to additional functional tests being written in #1898 (thanks, @redshiftzero!). I also think it would be good to fix the missing `</a>` tag for the next release, and it would be good to eventually integrate some HTML linting into our automated test suite as suggested in #1897. We shouldn't rely on web browsers to fix issues with our HTML; however, it's also important to make sure issues are described accurately and are based on observations, not assumptions.

redshiftzero:
Closing in favor of #1937.

PR #1830, opened 2017-06-16
Eliminate need for source.generate_unique_codename

fowlslegs:
## Status

Ready for review.

## Description of Changes

Inspired by #1812.

* By verifying that it's impossible for `crypto_util.genrandomid()` to generate a codename longer than our maximum in our tests, we no longer have to rely on the not-so-great `source.generate_unique_codename()` function.
* While writing this test I discovered that a failure to correctly strip the contents of our wordlist files resulted in the last "word" of each list object being `''`, the empty string. In addition to fixing this I wrote a test
  to prevent regressions.
* While it would be nice to keep NUM_CODENAME_WORDS together with MAX_CODENAME_LEN, that produced a circular imports problem. I resolved it in the clearest way I could think of that did not require significant refactoring.

### DIFF
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-
 from base64 import b32encode
 import os
 import subprocess
@@ -11,6 +10,7 @@ from gnupg._util import _is_stream, _make_binary_stream
 import scrypt
 
 import config
+import db
 import store
 
 # to fix gpg error #78 on production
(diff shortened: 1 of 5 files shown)


dachary:
Am I right to assume this won't be merged before 0.4 is released ?

fowlslegs:
> Am I right to assume this won't be merged before 0.4 is released ?

Yeah, I think that sounds right. We'll talk about it as a team once our glorious and revered leader @redshiftzero is back, but I suspect that'll be the case. Going to label as blocked for now.

fowlslegs:
> Sure, but that wasn't the original or primary purpose of generate_unique_codename—the original purpose was to prevent multiple sources from accidentally being assigned the same codename.

When writing this PR I did recognize this code was intended to do this, but decided to remove it as doing so serves no practical purpose.

> Yes, I know the probability of this actually occurring is miniscule, but better safe than sorry, right?

Miniscule is an understatement if anything. The probability is negligible in the cryptographic sense of the word. And when the probability of something happening is cryptographically negligible it's safe to forget about it. We already implicitly rely on this assumption elsewhere by trusting the cryptography we use, we should rely on it here too.

fowlslegs:
Cannibalized part of this for #1901, but am going to hold off on unnecessary app code changes until after 0.4. Will resurrect in a new PR then.

Issue #1826, opened 2017-06-15
Full test coverage for external contributions


garrettr:
Our automatic CI currently only achieves partial coverage for external contributions (defined as pull requests created from a branch on a forked remote, not the origin). Specifically, all tests except for the CircleCI test suite are run automatically. This is because there are security concerns related to automatically running arbitrary code on some of our CircleCI-related infrastructure.

Since the only tests that do not automatically run are the CircleCI tests, we have full test coverage for everything except the [configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html), which only run on CircleCI.

Unfortunately, this makes it difficult for maintainers to apply due diligence to external contributions before merging them. Ideally, every pull request would satisfy the invariant "If it was merged, it passed **all** of the tests." Satisfying this invariant makes debugging and reasoning about regressions _much_ easier.

## Long-term resolution

We have some ideas for how to fix this, which are being tracked on a private issue tracker internal to the Freedom of the Press Foundation. This public issue is meant to serve as a placeholder to provide some transparency regarding this situation; it will be updated as progress is made on the issue internally.

## Short-term mitigation

In the meantime, maintainers have one of two choices for performing due diligence on external contributions before merge:

1. If the partial automatic CI passes, and the PR is small and/or unlikely to break the configuration tests, you may merge without running the CircleCI tests. To do so: temporarily disable the [CircleCI status check requirement](https://github.com/freedomofpress/securedrop/settings/branches/develop), merge the external contribution, then re-enable the CircleCI status check requirement.
2. Otherwise, you should run the CircleCI tests. To do so:
    1. Review the contribution and ensure that it does not attempt to do anything malicious once it is running on our CI infrastructure.
    2. Checkout the branch from the fork's remote locally, then push it to the origin remote.
    3. Create a new pull request for the branch you pushed to the origin. The action of creating a new pull request is what triggers the full CI test suite to run.
    4. Comment on the original PR with a link to the temporary PR, explaining why it was created.
    5. Once you have confirmed the full test suite is passing, merge the original PR, then close the temporary PR.

redshiftzero:
Given #2070, from now on we should only use Option 2 to merge PRs from forks. That is, we should pull the changes into another PR so that the full CI suite can run before merging.

conorsch:
I propose we pull the lever and start building all forked PRs by default:

[...]

PR #1838, opened 2017-06-16
Reorganizes top-level Ansible vars for cleaner variable precedence

### LINKED ISSUE #1736: Fix intermittent custom_notification test failure

conorsch:
## Status

Ready for review.

## Description of Changes

Fixes #1736, which pointed out a problem caused by our use of nested groups in the Vagrant Ansible config. Takes most of the top-level vars and shoves them into role defaults, which simplifies [variable precedence](https://docs.ansible.com/ansible/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable) and makes it a clear-cut case where and how to override elements where appropriate.

Although the diff looks substantial, note that the vars have _moved_ to a more convenient location. In a few cases, a few vars were removed, only if they were never included anywhere (and were therefore left over from previous versions of the Ansible config). In other cases, vars were used by unnecessarily providing too much indirection, such as `non_default_securedrop_user` which just pointed to `securedrop_user`. Removed the former and preserved the latter. The functionality change that allows the test to pass is the removal of the nested group memberships within the `Vagrantfile`.

This PR was originally submitted from a fork as #1824. Resubmitting from origin, so CI runs (see #1826). 

## Testing

Run the application test suite _locally_ in Vagrant and confirm that 100% passes. You'll need to do so on a fresh staging build, since bad variable precedence from a previous build likely poisoned your config.py.

## Deployment

Affects the dev env. Made sure not to alter the way variables land on different hosts, just cleaned up the layout so it's more intuitive and consistent with best practices. We can now properly use host var overrides where appropriate and be confident the override will take effect (#1736 showed that sometimes it didn't).

- :red_circle:  [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

The development VM currently has broken config tests due to merge of #1770. Opened #1816 to track and will resolve there.


msheiny:
Im still pressing through review -- but I think that #1816 should get resolved first and that this PR should be rebased on that to ensure proper dev CI tests run against these changes.

conorsch:
Well it's hard to do separately, unfortunately, since the resolution for #1816 edited a lot of the same files as this one did, so I've got that resolution queued up in a separate branch. Will open a PR immediately, then you can confirm the changes in Travis as well as Circle.

conorsch:
@msheiny Done—see #1859 for inclusion of the Travis changes.

conorsch:
Thanks @msheiny, just what I was looking for. Will rebase #1859 on top of the latest develop post-merge to expedite review there.

PR #1840, opened 2017-06-16
doc: /staging/ is app-staging

dachary:


dachary:
it feels like the **/staging/** was intentional and had a semantic once ?

fowlslegs:
`/staging/` is a regular expression which matches both `app-staging` and `mon-staging` (https://www.vagrantup.com/docs/multi-machine/#controlling-multiple-machines). Consult https://docs.securedrop.org/en/latest/overview.html#infrastructure , https://docs.securedrop.org/en/latest/development/virtual_environments.html#staging , and https://github.com/freedomofpress/securedrop/blob/develop/Vagrantfile#L49 for details.

dachary:
Ah, that makes sense now, thanks :-) It looks like the minimum Vagrant version requirements should be updated then because it does not work with Vagrant 1.8.1 which is the default for the recommended Ubuntu 16.04.
<pre>
loic@fold:~/software/securedrop/securedrop$ vagrant up /staging/
The machine with the name '/staging/' was not found configured for
this Vagrant environment.
loic@fold:~/software/securedrop/securedrop$ vagrant --version
Vagrant 1.8.1
loic@fold:~/software/securedrop/securedrop$ sudo apt-cache policy vagrant
vagrant:
  Installed: 1.8.1+dfsg-1
  Candidate: 1.8.1+dfsg-1
  Version table:
 *** 1.8.1+dfsg-1 500
        500 http://fr.archive.ubuntu.com/ubuntu xenial/universe amd64 Packages
        500 http://fr.archive.ubuntu.com/ubuntu xenial/universe i386 Packages
        100 /var/lib/dpkg/status
</pre>

dachary:
It looks like this is a bug that [was fixed in 1.8.3](https://github.com/mitchellh/vagrant/blob/master/CHANGELOG.md#183-june-10-2016)

fowlslegs:
> It looks like this is a bug that was fixed in 1.8.3

Yeah, I remember us dealing with this now, but I thought we had addressed it already in the documentation. Thanks for that PR.

PR #1844, opened 2017-06-17
doc: remove extra leading $

dachary:
## Status

Ready for review

## Description of Changes

The SecureDrop documentation style guide's recommendations for quoting shell commands and shell output are not followed consistently throughout the documentation. This pull request applies the recommended formatting consistently across all of the documentation.

### DIFF
--- a/docs/servers.rst
+++ b/docs/servers.rst
@@ -319,7 +319,7 @@ First, generate the new SSH keypair:
 
 ::
 
-    $ ssh-keygen -t rsa -b 4096
+    ssh-keygen -t rsa -b 4096
 
 You'll be asked to "enter file in which to save the key." Type
 **Enter** to use the default location.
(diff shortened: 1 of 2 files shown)


conorsch:
Thanks, @dachary, these are solid changes and definitely in line with our recommendations. As @garrettr mentions, having the style guide enforced throughout the docs is a good idea, however be aware [that the style guide declares](https://docs.securedrop.org/en/latest/development/documentation_guidelines.html#style-guide):

> In the context of a terminal session, with both typed commands and printed output text, then use `$`, but only on the typed command lines

So it's a bit tricky to enforce consistently, and will definitely take manual review of the grep output on a case-by-case basis.

dachary:
@garrettr the `find` was helpful, thank you ! It caught another instance and the commit was updated & repushed. Also added a meaningful description to the pull request, using the template.

@conorsch I verified manually that all other occurences of a leading $ are in blocks that contain shell output.

garrettr:
@dachary Thanks for following up! With those changes this PR LGTM.

In the context of #1826, this PR only changes documentation, so I think it can be merged without CircleCI coverage.

PR #1845, opened 2017-06-17
doc: hardware: tails-3.0 is out and only supports x86_64

dachary:


garrettr:
In the context of #1826, this PR only changes documentation, so it is safe to merge without the CircleCI test coverage.

PR #1846, opened 2017-06-17
doc: onboarding: unify italics on Workstations

dachary:


garrettr:
In the context of #1826, this PR only changes the documentation and so does not need the CircleCI test coverage.

PR #1847, opened 2017-06-17
tests: store: unit test for secure_unlink & delete_source_directory

dachary:


garrettr:
In the context of #1826, this PR adds additional web application unit tests, which are covered by TravisCI and Coveralls (which both passed). Therefore, CircleCI test coverage is unnecessary for this PR and it can be merged without running on CircleCI.

PR #1848, opened 2017-06-17
cleanup: journalist: remove useless NoResultFound import

dachary:
## Status

Ready for review

## Description of Changes

NoResultFound imported from journalist.py actually is the same
as the previously imported exception by the same name from sqlalchemy


dachary:
@garrettr repushed with the suggested change and added a meaningful comment for the PR. Thanks for your review :-)

PR #1852, opened 2017-06-17
ansible: remove redundant handler

dachary:


dachary:
@conorsch I'm not sure this is actually redundant (because I'm so new to ansible ... the chances of me being mistaken are high ;-). What do you think ?

conorsch:
Thanks for the keen eye, @dachary. In this case, I'd prefer to keep the duplicated handler as-is. In Ansible nomenclature, _roles_ are bundles of _tasks_ and _vars_, and can include additional helpers such as _modules_ and _handlers_. 

When a role is executed, it exports its _handlers_ into the namespace used for the given _play_. So in the case you point out, the `app-test` handler is not technically required, since that role can use the handler in the `app` role. This only works because the two roles are included within the same play—if the `app-test` role were moved to a different play, and your changes were merged, the playbook would fail.

To keep the roles as "modular" as possible, I'd like to keep the handlers they require explicitly declared within the roles, rather than implicitly inherited from other roles.

Down the road (i.e. post-0.4), we'd do well to start ripping out a lot of these custom roles and leaning more heavily on community-maintained solutions. Now that we have decent configuration testing (#1616), we can make that transition more confidently, and reduce the amount of code we're directly responsible for maintaining within this repository.

dachary:
@conorsch thanks for the detailed explanation :-) I am not able to find where app-test uses the **restart apache2** handler ? 

BTW: #1616 is a really impressive set of commits!

conorsch:
@dachary Well I'll be! You're right, it's _not_ included in the `app-test` role, so I'm also comfortable removing it, as you initially suggested.

dachary:
It looks like the travis failure is unrelated to this change ?

<pre>
=================================== FAILURES ===================================

____________________ TestIntegration.test_filenames_delete _____________________

self = <tests.test_integration.TestIntegration testMethod=test_filenames_delete>

    def test_filenames_delete(self):

        """Test pretty, sequential filenames when journalist deletes files"""

        # add a source and submit stuff

        self.source_app.get('/generate')

        self.source_app.post('/create')

        self.helper_filenames_submit()

    

        # navigate to the collection page

        resp = self.journalist_app.get('/')

        soup = BeautifulSoup(resp.data)

        first_col_url = soup.select('ul#cols > li a')[0]['href']

        resp = self.journalist_app.get(first_col_url)

        self.assertEqual(resp.status_code, 200)

        soup = BeautifulSoup(resp.data)

    

        # delete file #2

>       self.helper_filenames_delete(soup, 1)

tests/test_integration.py:514: 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

tests/test_integration.py:613: in helper_filenames_delete

    utils.async.wait_for_assertion(lambda: self.assertFalse(

tests/utils/async.py:47: in wait_for_assertion

    return assertion_expression()

tests/test_integration.py:614: in <lambda>

    any([os.path.exists(store.path(sid, doc_name)) for doc_name in checkbox_values])))

E   AssertionError: True is not false
</pre>

dachary:
@conorsch ok then :-)

conorsch:
Agreed @dachary, looks unreleated, and likely ephemeral. To do due diligence here, I'm going to run through the workflow described in #1826 so that we have our entire CI suite validating this change.

dachary:
Closing in favor of #1854. @conorsch this is a lot of work for a trivial cleanup. I'll refrain from semi-gratuitous cleanups like this one in the future.

PR #1853, opened 2017-06-18
docs: recommend vagrant 1.8.5 instead of >= 1.8

dachary:
The instructions to bring up the virtual machines makes use of regular
expressions (vagrant up /staging/). It does not work with vagrant 1.8.1
which is the default supported version for the recommended operating
system version (Ubuntu 16.04).

Although the recommended version is 1.8.5, the check suggested with
apt-cache policy suggests that any version above 1.8 is going to
work. Raise the minimum version to 1.8.5 to avoid this pitfall.


dachary:
thanks for the merge :-)

PR #1828, opened 2017-06-16
Fix usage of shutil.rmtree()

fowlslegs:
## Status

Ready for review.

## Description

* Use of the `ignore_errors` keyword argument simpler than manually catching :exc:`OSError`s and letting them `pass`.
* Fixes intermittent errors cause by racing teardown scripts (e.g., https://travis-ci.org/freedomofpress/securedrop/builds/243419555#L870).
* Ensures that `shutil.rmtree()` is always run in the `_can_decrypt_with_key` helper function in the `test_integration.py` module.

### If you made changes to the app code:

- [x] Unit and functional tests pass on the development VM

PR #1854, opened 2017-06-19
ansible: remove redundant handler

conorsch:
"restart apache2" is not used by the app-test role and
is already defined by the "app" role.

## Status

Ready for review.

## Description of Changes

Resubmitting @dachary's #1852, to get the full CI suite to run (as described in #1826). Therefore closes #1852.

## Testing

Validate all CI checks pass. Details of implementation were already hashed out in #1852 discussion thread.

### DIFF
--- a/install_files/ansible-base/roles/app-test/handlers/main.yml
+++ b/install_files/ansible-base/roles/app-test/handlers/main.yml
@@ -1,10 +1,5 @@
 ---
 ## SecureDrop Interfaces
-- name: restart apache2
-  service:
-    name: apache2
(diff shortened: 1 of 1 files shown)


fowlslegs:
Another intermittent Travis app code test failure with `shutil.rmtree()`. Restarted the build. Will be following up on https://github.com/freedomofpress/securedrop/pull/1828 today.

dachary:
@fowlslegs could it be that tests run in parallel and use directories that have the same name, racing against each other ?

PR #1859, opened 2017-06-19
Ensures config tests run in Travis CI

### LINKED ISSUE #1816: Testinfra tests are skipped in Travis

conorsch:
## Status

Ready for review.

## Description of Changes

Fixes #1816. The testinfra config tests were being skipped in the Travis run, which allowed some regressions to slip in regarding dependency version checks (that only affected the development VM).

Originalyl marked with "blocked" label because it includes the large diff submitted as #1838, and therefore #1838 must be reviewed and merged first. That's now done.

## Testing

1. Confirm that all CI checks pass.
2. For Travis, confirm the checks not only pass, but display the testinfra run as green PASSED rather than yellow SKIPPED. To be explicit:

### Should look like this:

travis-ci-testinfra-yes

### Not this:
travis-ci-testinfra-no

## Deployment

No concerns, these are dev-env changes.


conorsch:
Rebased on top of latest develop, after merging #1838. Ready for review.

conorsch:
Digging into the coveralls failure—seems to me these changes should _not_ result in a coverage decrease.

conorsch:
Looks like we had a recent spike in coverage, and this PR reduced it:

coveralls-spike-up-and-down

Tried another rebase on top of develop to ensure that the very latest changes are included in this PR. Will reassess once CI reports back.

conorsch:
Rebase satisfied Coveralls. :1st_place_medal:

.travis.yml
```diff
@@ -12,6 +12,7 @@ virtualenv:
   system_site_packages: true
 before_install:
   - 'for e in /.packer-env/*; do echo -n "${e}: "; cat "${e}"; done'
+  - printenv | sort # dump Travis environment for debugging
```
garrettr: @conorsch Do we want to leave this change in permanently? Seems like it was included for debugging while this work was in progress, and may no longer be needed.

testinfra/conftest.py
```diff
@@ -26,19 +25,17 @@ def securedrop_import_testinfra_vars(hostname, with_header=False):
     Vars must be stored in `testinfra/vars/<hostname>.yml`.
     """
     filepath = os.path.join(os.path.dirname(__file__), "vars", hostname+".yml")
-    file_load = open(filepath, 'r')
+    hostvars = yaml.load(open(filepath, 'r'))
```
garrettr: Prefer `yaml.safe_load`, unless you really need `load`: https://security.openstack.org/guidelines/dg_avoid-dangerous-input-parsing-libraries.html.

.travis.yml
conorsch: Yes, the env var dump should stay. It doesn't add to build time, the output folds automatically in the web interface, and it's invaluable in case of a failure. Rather than having to add it ad-hoc in the future, I just want it in there.

testinfra/conftest.py
conorsch: Was not on my radar, :tophat: to you, sir. Will add. Gonna toss that `open` into a context manager while I'm in there.

.travis.yml
```diff
@@ -21,22 +22,23 @@ before_install:
 install:
   - pip freeze -l
   # Preinstalled travis pytest is not tracked by us
-  - pip uninstall pytest -y
+  # Preinstalled travis pbr is not tracked by us
```
fowlslegs: Instead of uninstalling Python packages that overlap between what Travis includes and what we use, it would better to simply run `pip install -r testinfra/requirements.txt` with the `-U` flag. This is preferable because (i) if the Python packages in Travis are updated, we don't have to worry about additional overlaps in the future and (ii) if they happen to be on the same version of an overlapping package, we don't have to reinstall it.

Last time I checked `pip install -r reqs.txt` will not upgrade or downgrade an already installed package even if requirements file specifies a different version of that package, whereas `pip install -U -r reqs.txt` will upgrade/downgrade that package to the version specified in the requirements file.

You can get rid of the three `pip freeze -l` commands I see in this diff too then, which look like they are used for debugging. Ensuring the right dependencies are present in Travis should always be confirmed by testinfra.

conorsch: Thanks for pointing out the `pip` idiosyncrasies, will update the install commands with the `-U` option and remove the explicit removals. Still keeping the `pip freeze -l` commands since we have several pip installs from various requirements files and I want that info in there for debugging purposes.

testinfra/conftest.py
```diff
@@ -25,7 +25,8 @@ def securedrop_import_testinfra_vars(hostname, with_header=False):
     Vars must be stored in `testinfra/vars/<hostname>.yml`.
     """
     filepath = os.path.join(os.path.dirname(__file__), "vars", hostname+".yml")
-    hostvars = yaml.load(open(filepath, 'r'))
+    with open(filepath, 'r') as f:
```
fowlslegs: `r` is the default mode fwiw, but no need to change this of course :smile:

conorsch: Explicit is better than implicit! :stuck_out_tongue_winking_eye:

testinfra/conftest.py
```diff
@@ -26,19 +25,18 @@ def securedrop_import_testinfra_vars(hostname, with_header=False):
     Vars must be stored in `testinfra/vars/<hostname>.yml`.
     """
     filepath = os.path.join(os.path.dirname(__file__), "vars", hostname+".yml")
-    file_load = open(filepath, 'r')
+    with open(filepath, 'r') as f:
+        hostvars = yaml.safe_load(f)
```
fowlslegs: I noticed this in `securedrop-admin`:

```
106:            site_config_vars = yaml.safe_load(site_config_file.read())
```

so tested and confirmed that `pyyaml` will take either a `file` or `str` object.

conorsch:
@fowlslegs Implemented your `pip install -U` recommendation in 4fd5bea, but that broke the build. I'm inclined to revert that commit. Before I do, anything you see wrong with it? For comparison:

* [working build](https://travis-ci.org/freedomofpress/securedrop/builds/244740978)
* [failing build](https://travis-ci.org/freedomofpress/securedrop/builds/244745224)

Oh, pip.

fowlslegs:
Alright, so if you look at the last line in the JSON stdout list of [this line in the failing build](https://travis-ci.org/freedomofpress/securedrop/builds/244745224#L866):

> Successfully installed beautifulsoup4-4.6.0 blinker-1.4 certifi-2017.4.17 chardet-3.0.4 coverage-4.4.1 coveralls-1.1 docopt-0.6.2 first-2.0.1 flask-testing-0.6.2 funcsigs-1.0.2 idna-2.5 mock-2.0.0 pbr-3.0.1 pip-tools-1.9.0 py-1.4.34 pytest-3.1.1 pytest-cov-2.5.1 requests-2.17.3 selenium-2.53.6 urllib3-1.21.1

You don't, however, see a line indicating `pbr==1.10.0` was uninstalled. I think it may be that they are getting installed simultaneously, but to different places on the PYTHONPATH. Alternatively, `pip` could have changed behavior since the last time I tested. I did admit it had been a while. Since we have a way that works, I'm inclined to say we should just stick with. The intention of the suggestion was to save time in the future, not cost it in the now.

You could try

```
for dep in $(pip freeze -l | cut -d'=' -f1 | tr '[:upper:]' '[:lower:]' | tr '-' '_'); do python -c "import ${dep}; print(${dep}.__file__)"; done || true
```

to test my theory, but it isn't necessary. 

...I really do not like `pip`.

testinfra/vars/travis.yml
```diff
@@ -1,4 +1,7 @@
 ---
 securedrop_user: travis
-securedrop_code: {{ BUILD_DIR }}/securedrop
+# The directory Travis runs builds in varies by PR, so we cannot hardcode it
+# in testvars. See `conftest.py`, where the `securedrop_data` var is read from
+# a Travis-specific environment variable.
+# securedrop_code: /home/travis/build/freedomofpress/securedrop
```
msheiny: Thanks for the verbose comments :+1:

conorsch:
Thanks for review, @msheiny. Testinfra tests are skipping for me locally against the Vagrant development VM, but that's also true of the develop branch, not just this PR, so will handle in #1866.

PR #1860, opened 2017-06-19
Prefer use of yaml.safe_load

fowlslegs:
## Status

Ready for review, but needs more local testing. Wanted to figure out what needs to be done there, but wanted to briefly consult my good friend @conorsch first.

## Description of Changes

Always prefer `yaml.safe_load()` to `yaml.load()`. I can't really imagine a time where we'd need to use the latter, anyway.

## Testing

Maybe @conorsch could help with the story here. It seems like as long as you can provision and SSH into all the machines we can confirm this change doesn't break anything. Let me know if you agree that's sufficient, or there's more this could affect.

### DIFF
--- a/install_files/ansible-base/inventory-dynamic
+++ b/install_files/ansible-base/inventory-dynamic
@@ -42,7 +42,7 @@ def lookup_local_ipv4_address(hostname):
     """
     try:
         with open(SECUREDROP_SITE_VARS, 'r') as f:
-            site_vars = yaml.load(f)
+            site_vars = yaml.safe_load(f)
             app_ip = site_vars['app_ip']
             monitor_ip = site_vars['monitor_ip']
 


conorsch:
Seems quite reasonable. Will require testing in Tails to be absolutely sure, though.

fowlslegs:
If @garrettr recalls correctly (see https://gitter.im/freedomofpress/securedrop?at=594865db5bf0bef94cc18463) then this only affects production environments as only `install_files/ansible-base/ansible.cfg` uses that script to dynamically generate it's inventory. Can you confirm this @conorsch and whether it applies to the `prod` virtual environment?

fowlslegs:
Rebased.

redshiftzero:
Agree - the diff looks fine, but we need to test it prior to merge. Next time someone runs through a test install they should check out this branch.

conorsch:
Reviewing #2083 now, which requires testing in an Admin Workstation, so will make sure to run this through its paces when testing for 2083 is complete.

conorsch:
Satisfied there's no breakage here. Confirmed via manual testing that the relevant `securedrop-admin` subcommands behave as expected, and executing the dynamic inventory file directly (after activating virtualenv for yaml dependency) displays correct output, as well.

Going to mark as approved, and once CI passes we can merge.

PR #1810, opened 2017-06-15
Minor refactor + test coverage for secure_tempfile.py

fowlslegs:
## Status

Ready for review.

## Description of Changes

* Test coverage for `secure_tempfile.py` is now 100%, and the new test module `test_secure_tempfile.py` alone provides 100% coverage of `secure_tempfile.py`.
* In the process minor refactoring was done:
  * Orders imports according to PEP8.
  * Adds standard UTF-8 encoding header.
  * Improves documentation of :class:`SecureTemporaryFile` and it's methods.
  * Converts `SecureTemporaryFile` to a new-style class for use with `super()`.
  * Minor formatting changes: both PEP8 related and personal opinions.
  * Removes purposeless `try`/`except` block.
  * Ensures :meth:`read()` cannot be called before :meth:`write()`. I think this was the intended behavior (it's at least the behavior promised by the original docstring), so I consider this a bug fixed.

### If you made changes to the app code:

- [x] Unit and functional tests pass on the development VM

PR #1863, opened 2017-06-20
Secure tempfile refactor alternative test fix

garrettr:
Minimal alternative fix for broken tests in #1810.

I do not think this should necessarily supersede #1810; I am just posting it as an alternative way to fix the tests. @fowlslegs' work in #1810 seems to me to be a good start on unit tests for `store.save_file_submission`, so I think ideally my work from this branch would be incorporated into #1810, which would then be merged.

@fowlslegs What do you think?

### DIFF
--- a/securedrop/secure_tempfile.py
+++ b/securedrop/secure_tempfile.py
@@ -1,93 +1,119 @@
+# -*- coding: utf-8 -*-
 import base64
 import os
 from tempfile import _TemporaryFileWrapper
 
+from gnupg._util import _STREAMLIKE_TYPES
 from Crypto.Cipher import AES
 from Crypto.Random import random
 from Crypto.Util import Counter
 
-from gnupg._util import _STREAMLIKE_TYPES
 
-class SecureTemporaryFile(_TemporaryFileWrapper):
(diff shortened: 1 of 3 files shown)


fowlslegs:
This is what I wanted to do at first, but couldn't figure out how to do it. Probably just bad oversight when reading the documentation.

I agree on the final code state we want to achieve here. I think that merging this PR, closing #1810, and then making a small follow up PR including [the test here](https://github.com/freedomofpress/securedrop/pull/1863/commits/f489353293f80ad720dadf94a93d23950ff8576e#diff-e72734cd811fe95d511a138acdd0c91bR74) (modified to make use of a spy)  to increase `store` coverage will be a slightly easier way to get to that state though.

fowlslegs:
Locally I dropped the middle two commits and then rebased on develop. Okay if I push that and then we merge?

garrettr:
@fowlslegs Force push that branch to this PR, I'll rubber-stamp it, then it should be good to go.

garrettr:
@fowlslegs LGTM! I can't approve this branch via GitHub review because I'm not allowed to approve my own branches, but I'd say it's good to go once CI passes.

PR #1865, opened 2017-06-21
Adds ipdb apt pkg to development_dependencies

### LINKED ISSUE #1369: Consider installing some debug/ development tools in development VMs

fowlslegs:
## Status

Ready for review.

## Description of Changes

Closes #1369 even harder.

### DIFF
--- a/install_files/ansible-base/group_vars/all/securedrop
+++ b/install_files/ansible-base/group_vars/all/securedrop
@@ -20,6 +20,7 @@ development_dependencies:
   - libssl-dev
   - ntp
   - python-dev
+  - python-ipdb
   - python-pip
   - ruby # to ensure `gem` command is available
   - tmux


fowlslegs:
Can't confirm configuration tests pass until https://github.com/freedomofpress/securedrop/issues/1866 is resolved.

dachary:
It is not trivial to install via pip, installing it with a package indeed is a good option.

fowlslegs:
The reason to install it via apt as opposed to using `sudo pip install` as we do with the `requirements.txt` files is that the install path for apt uses comes later on the PYTHONPATH versus the install path `sudo pip install` uses. So in the case `ipdb` brought in some dependency that SD also uses, SD would default to using the `sudo pip install` version from our `requirements.txt` file instead of the version brought in by apt.

Now that I think about it, it's actually tempting to install `ipython` and `ipdb` with `sudo pip install` because the latest versions are a lot more featureful than what the outdated Trusty repos provide. The reason this seems safe to do (in terms of ensuring SD is run/ tested with all the right dependency versions) is because testinfra provides a check that this is the case. The testinfra tests are what allow us to safely install `securedrop-requirements.txt` and then `test-requirements.txt` on the `development` machine. Those two requirements files often share dependencies, and if the latter had a different version of one than the former, it would overwrite it, and the testinfra tests would fail.

I think it's more simple just to stick to apt here, and not modify this PR though. Just wanted to explain our Python packages in `development` strategy, and double check in the process that everything made sense.

fowlslegs:
Worked around #1866 using the instructions in https://github.com/freedomofpress/securedrop/issues/1866#issuecomment-310200724.

PR #1868, opened 2017-06-21
journalist: gracefully handle HOTP format errors

dachary:
## Status

Ready for review

## Description of Changes

Fixes: https://github.com/freedomofpress/securedrop/issues/1869

In admin/reset-2fa-hotp, when the token is not properly formatted (odd
number of characters, non hexa character), a python trace is displayed
to the user.

When a known formatting error is detected, a user friendly error message
is displayed instead of the stack trace. Other errors are not caught and
will show a stack trace. Chances are they are not understandable for the
user anyway, even if beautified.

## Testing

pytest -v tests/test_journalist.py

## Deployment

It only changes the user experience and has no impact on deployment.

after

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -242,9 +242,17 @@ def admin_reset_two_factor_hotp():
     otp_secret = request.form.get('otp_secret', None)
     if otp_secret:
         user = Journalist.query.get(uid)
-        user.set_hotp_secret(otp_secret)
-        db_session.commit()
-        return redirect(url_for('admin_new_user_two_factor', uid=uid))
+        try:
+            user.set_hotp_secret(otp_secret)
+            db_session.commit()
+            return redirect(url_for('admin_new_user_two_factor', uid=uid))
+        except TypeError as e:
+            if ("Non-hexadecimal digit found" not in str(e) and
+                "Odd-length string" not in str(e)):
+                raise
+            flash("Invalid secret format: {}".format(e), "error")
+            db_session.rollback()
+            return render_template('admin_edit_hotp_secret.html', uid=uid)
     else:
         return render_template('admin_edit_hotp_secret.html', uid=uid)
 
--- a/securedrop/tests/test_journalist.py
+++ b/securedrop/tests/test_journalist.py
@@ -371,6 +371,44 @@ class TestJournalistApp(TestCase):
         self.assertRedirects(resp,
             url_for('admin_new_user_two_factor', uid=self.user.id))
 
+    def test_admin_resets_user_hotp_format_non_hexa(self):
+        self._login_admin()
+        old_hotp = self.user.hotp.secret
+
+        resp = self.client.post(url_for('admin_reset_two_factor_hotp'),
+                                data=dict(uid=self.user.id, otp_secret='ZZ'))
+        new_hotp = self.user.hotp.secret
+
+        self.assertEqual(old_hotp, new_hotp)
+        self.assertMessageFlashed('Invalid secret format: '
+                                  'Non-hexadecimal digit found', 'error')
+
+    def test_admin_resets_user_hotp_format_odd(self):
+        self._login_admin()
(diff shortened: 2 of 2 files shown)


securedrop/journalist.py
```diff
@@ -242,9 +242,17 @@ def admin_reset_two_factor_hotp():
...
     if otp_secret:
         user = Journalist.query.get(uid)
-        user.set_hotp_secret(otp_secret)
-        db_session.commit()
-        return redirect(url_for('admin_new_user_two_factor', uid=uid))
+        try:
+            user.set_hotp_secret(otp_secret)
+            db_session.commit()
+            return redirect(url_for('admin_new_user_two_factor', uid=uid))
+        except TypeError as e:
```
fowlslegs: Since the `TypeError` we're trying to catch is going to be raised by `user.set_hotp_secret()`, and not by the following two lines it makes sense to put the two following lines in an `else` block.

securedrop/journalist.py
```diff
@@ -242,9 +242,17 @@ def admin_reset_two_factor_hotp():
...
+            if ("Non-hexadecimal digit found" not in str(e) and
+                "Odd-length string" not in str(e)):
+                raise
```
fowlslegs: With nothing to catch this exception, the request aborts and the client receives an error page. I think a better solution is to just flash a generic error here that an unexpected error occurred. In this `except` block you could determine which message to flash with a series of `if`/`elif`/`else` statements that introspect the exception.

securedrop/journalist.py
```diff
@@ -242,9 +242,17 @@ def admin_reset_two_factor_hotp():
...
+            flash("Invalid secret format: {}".format(e), "error")
```
fowlslegs: To ensure consistent and safe messaging to the user, it's better to use our own strings here.

dachary:
I'll rework that but it will take me a little time, closing for now

fowlslegs:
No need to close an issue in this case. Unless the next iteration is going to be very drastically different, will take weeks to complete, or is not suitable for 0.4, I'd say keep PRs you need to rework open.

dachary:
I better like to reopen later if you don't mind. I feel pressure from open pull requests. I have to talk to my analyst about that ;-)

fowlslegs:
Sure thing :smile:

dachary:
resuming work on this but ... it appears I I can't re-open this issue ?

dachary:
> I think some validation of the OTP secret is in order, but that that should come in a separate PR. Right now, any even length hex string (containing any number of spaces) will be accepted besides the null string. In the case of the null string, that is handled here, but not gracefully (by which I mean there is no helpful user messaging via flash or any other mechanism). You could implement this too if you want, but I think a good implementation of set_hotp_secret would replace the need to perform any checks on the OTP string in admin_reset_two_factor_hotp(), so it will/ should be replaced in the future anyway.

I'd be happy to do that. Since we're very close to the 0.4 release, it is maybe better to make a minimal change does not change any of the logic and focuses exclusively on the exceptions that create a bad user experience. What do you say ?

securedrop/journalist.py
dachary: good point, done

securedrop/journalist.py
dachary: the downside of showing a "unexpected error happened" is that it hides the exception completely. The error may not be useful to the user but it could be copy/pasted to a support ticket and be useful to the person doing the diagnostic. That being said, if that's really what you want I'll do it :-)

securedrop/journalist.py

securedrop/journalist.py
```diff
@@ -242,9 +242,17 @@ def admin_reset_two_factor_hotp():
...
+            db_session.rollback()
```

dachary:
repushed

dachary:
Looks like I need to open a new pull request after all: https://github.com/freedomofpress/securedrop/pull/1879

PR #1870, opened 2017-06-21
unify two-factor spelling

dachary:
## Status

Ready for review

## Description of Changes

Change "two factor" in "two-factor" so it is unified in the
documentation and the user visible messages.

https://en.wiktionary.org/wiki/2fa


dachary:
in favor of the other one

fowlslegs:
The original gets merged according to our temporary practices outline in short-term strategy #2 in https://github.com/freedomofpress/securedrop/issues/1826.

dachary:
ack

PR #1871, opened 2017-06-21
doc: git describe --exact-match may fail

dachary:
## Status

Ready for review

## Description of Changes

When not on a tag, it will fail with:

    fatal: no tag exactly matches 'dc301206cd38e5f1c30a1687f19d2d0b7b5054cf'

which is not very useful. Changing to `git describe` alone which will show the tag if it matches the current commit or something like

    0.3.11-2103-gdc301206

otherwise which gives some kind of indication regarding the latest release.

### DIFF
--- a/docs/backup_and_restore.rst
+++ b/docs/backup_and_restore.rst
@@ -55,8 +55,8 @@ Backing Up
 
 Open a **Terminal** on the *Admin Workstation* and ``cd`` to your clone of the
 SecureDrop git repository (usually ``~/Persistent/securedrop``). Ensure you have
-SecureDrop version 0.3.7 or later checked out (you can run ``git describe
---exact-match`` to see what Git tag you've checked out).
+SecureDrop version 0.3.7 or later checked out (you can run ``git describe``
+to see what Git tag you've checked out).
 
 .. note:: The backups are stored in the *Admin Workstation*'s persistent volume.
           **You should verify that you have enough space to store the backups


dachary:
Functional [tests failed](https://travis-ci.org/freedomofpress/securedrop/builds/245455802?utm_source=github_status&utm_medium=notification), transient error it seems.

<pre>
tests/functional/test_admin_interface.py::AdminInterface::test_admin_interface FAILED

tests/functional/test_source_warnings.py::SourceInterfaceBannerWarnings::test_warning_appears_if_tor_browser_not_in_use FAILED

tests/functional/test_submit_and_retrieve_file.py::SubmitAndRetrieveFile::test_source_cancels_at_login_page FAILED

tests/functional/test_submit_and_retrieve_file.py::SubmitAndRetrieveFile::test_source_cancels_at_submit_page FAILED

tests/functional/test_submit_and_retrieve_file.py::SubmitAndRetrieveFile::test_submit_and_retrieve_happy_path FAILED

tests/functional/test_submit_and_retrieve_message.py::SubmitAndRetrieveMessage::test_submit_and_retrieve_happy_path FAILED

</pre>

garrettr:
Would not merge this. The use of `--exact-match` is intentional. The "backup and restore" documentation is intended for SecureDrop administrators, and using `--exact-match` adds an extra check that they are running an actual SecureDrop release (indicated by a signed git tag) before running any commands.

It is important to ensure administrators have a release (and not another commit) checked out because:

- We only intended for them to work with release code. If they have other code checked out, there are not guarantees that they are working with stable, complete code, and anything could happen.
- We only GPG-sign release tags, so it is safer for administrators to only use cryptographically verified code (as described in detail in the documentation).

dachary:
Thanks for taking the time to explain in detail !

PR #1872, opened 2017-06-21
CircleCI * dachary * unify two-factor spelling

fowlslegs:
Made to run the CircleCI tests on #1870.


fowlslegs:
Restarting because of some intermittent Selenium errors. Those in particular are not ones we can do much to resolve, but in general we really gotta get more of these intermittent failures resolved.

fowlslegs:
Restarted again. I think with the move back to the old Trusty Travis images this should finally pass.

fowlslegs:
Duh, have to rebase to get Travis to pass.

PR #1873, opened 2017-06-22
Raises main stack gap 64KB -> 1MB via sysctl

### LINKED ISSUE #1861: Mitigate "The Stack Clash"
Today Qulays published a [writeup](https://blog.qualys.com/securitylabs/2017/06/19/the-stack-clash) of a set of security vulnerabilities collectively called "The Stack Clash": their [security advisory](https://www.qualys.com/2017/06/19/stack-clash/stack-clash.txt) has the full technical details. Of special interest is the fact that Qualys was able to develop a PoC that works against grsecurity/PaX; see section `IV.1.6. Grsecurity/PaX` of the Qualys security advisory.

We should research this issue and determine what, if any, steps should be taken to mitigate the potential threat to SecureDrop installations.

conorsch:
## Status

Ready for review.

## Description of Changes

We're already setting this value via the `securedrop-grsec` metapackage, as described in #1861 and implemented in the grsec repo [0]. Let's also ensure it at install time by setting it directly along with the other
sysctl options.

Included a config test so that test suite will fail if the sysctl doesn't stick during CI or local development.

[0] https://github.com/freedomofpress/ansible-role-grsecurity/pull/100

Fixes #1861.

## Testing
CI it not sufficient here, since the CI hosts don't use grsecurity-patched kernels. Run through the provisioning with staging VMs, and confirm manually that `sudo sysctl vm.heap_stack_gap` shows `1048576`.

## Deployment

We already posted the new kernel packages in the apt repo, and the sysctl setting is set in the `securedrop-grsec` metapackage, so we're good to go for deployed instances. Adding the setting within Ansible is simply to guard against regressions, and also allows us to clean up the metapackage in the future.

### DIFF
--- a/install_files/ansible-base/roles/grsecurity/defaults/main.yml
+++ b/install_files/ansible-base/roles/grsecurity/defaults/main.yml
@@ -6,4 +6,8 @@ grsec_sysctl_flags:
     # rest will not be applied
   - name: "kernel.grsecurity.grsec_lock"
     value: "1"
-
+    # Stack clash mitigation, increasing main stack gap to 1MB.
+    # Storing as part of grsecurity vars, because sysctl option won't
+    # exist otherwise.
+  - name: "vm.heap_stack_gap"
+    value: "1048576"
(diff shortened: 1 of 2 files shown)


conorsch:
Rebased on top of latest develop to incorporate #1875, which should resolve the Travis failures.

Issue #1869, opened 2017-06-21
journalist: admin/reset-2fa-hotp gracefully handle formatting errors


dachary:
# Bug

## Description

In admin/reset-2fa-hotp, when the token is not properly formatted (odd
number of characters, non hexa character), a python trace is displayed
to the user.

## Steps to Reproduce

* loggin as a journalist admin
* click on reset hardware token (goes to admin/reset-2fa-hotp)
* type zz for the token

## Expected Behavior

An error message because Z is not an hexadecimal character

## Actual Behavior

A python traceback that displays an error message about the fact that there is non hexadecimal number

![before](https://user-images.githubusercontent.com/433594/27391047-2a4dceae-56a3-11e7-8672-521578d634a9.png)

## Comments

An error message would be more user friendly, typo are common

redshiftzero:
Thanks for this excellent report @dachary! You're right that we should gracefully handle this case. The only thing to add is that the traceback will only be shown in the development environment - in staging and production, the error will appear like this:

![screen shot 2017-06-26 at 2 38 00 pm](https://user-images.githubusercontent.com/7832803/27561515-498aa520-5a7d-11e7-8840-db9a758227b3.png)

PR #1879, opened 2017-06-25
journalist: gracefully handle HOTP format errors

dachary:
## Status

Ready for review

## Description of Changes

Fixes: https://github.com/freedomofpress/securedrop/issues/1869

In admin/reset-2fa-hotp, when the token is not properly formatted (odd
number of characters, non hexa character), a python trace is displayed
to the user.

When a known formatting error is detected, a user friendly error message
is displayed instead of the stack trace. Other errors are not caught and
will show a stack trace. Chances are they are not understandable for the
user anyway, even if beautified.

## Testing

pytest -v tests/test_journalist.py

## Deployment

It only changes the user experience and has no impact on deployment.

after

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -242,9 +242,28 @@ def admin_reset_two_factor_hotp():
     otp_secret = request.form.get('otp_secret', None)
     if otp_secret:
         user = Journalist.query.get(uid)
-        user.set_hotp_secret(otp_secret)
-        db_session.commit()
-        return redirect(url_for('admin_new_user_two_factor', uid=uid))
+        try:
+            user.set_hotp_secret(otp_secret)
+        except TypeError as e:
+            if "Non-hexadecimal digit found" in str(e):
+                flash("Invalid secret format: "
+                      "please only submit letters A-F and numbers 0-9.",
+                      "error")
+            elif "Odd-length string" in str(e):
+                flash("Invalid secret format: "
+                      "odd-length secret. Did you mistype the secret?",
+                      "error")
+            else:
+                flash("An unexpected error occurred! "
+                      "Please check the application "
+                      "logs or inform your adminstrator.", "error")
+                app.logger.error(
+                    "set_hotp_secret '{}' (id {}) failed: {}".format(
+                        otp_secret, uid, e))
+            return render_template('admin_edit_hotp_secret.html', uid=uid)
+        else:
+            db_session.commit()
+            return redirect(url_for('admin_new_user_two_factor', uid=uid))
     else:
         return render_template('admin_edit_hotp_secret.html', uid=uid)
 
--- a/securedrop/tests/test_journalist.py
+++ b/securedrop/tests/test_journalist.py
@@ -371,6 +371,56 @@ class TestJournalistApp(TestCase):
         self.assertRedirects(resp,
             url_for('admin_new_user_two_factor', uid=self.user.id))
 
+    def test_admin_resets_user_hotp_format_non_hexa(self):
+        self._login_admin()
+        old_hotp = self.user.hotp.secret
+
+        resp = self.client.post(url_for('admin_reset_two_factor_hotp'),
+                                data=dict(uid=self.user.id, otp_secret='ZZ'))
+        new_hotp = self.user.hotp.secret
+
+        self.assertEqual(old_hotp, new_hotp)
+        self.assertMessageFlashed(
+            "Invalid secret format: "
+            "please only submit letters A-F and numbers 0-9.", "error")
+
+    def test_admin_resets_user_hotp_format_odd(self):
+        self._login_admin()
+        old_hotp = self.user.hotp.secret
+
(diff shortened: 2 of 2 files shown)


securedrop/journalist.py
```diff
@@ -242,9 +242,19 @@ def admin_reset_two_factor_hotp():
...
         user = Journalist.query.get(uid)
-        user.set_hotp_secret(otp_secret)
-        db_session.commit()
-        return redirect(url_for('admin_new_user_two_factor', uid=uid))
+        try:
+            user.set_hotp_secret(otp_secret)
+        except TypeError as e:
+            if "Non-hexadecimal digit found" in str(e):
+                flash("Invalid secret format: non-hexadecimal digit found")
+            elif "Odd-length string" not in str(e)):
```
redshiftzero: Looks like the Travis builds are failing on this line, which should be:

`elif "Odd-length string" not in str(e):`

securedrop/journalist.py
```diff
@@ -242,9 +242,19 @@ def admin_reset_two_factor_hotp():
...
     if otp_secret:
         user = Journalist.query.get(uid)
-        user.set_hotp_secret(otp_secret)
-        db_session.commit()
-        return redirect(url_for('admin_new_user_two_factor', uid=uid))
+        try:
+            user.set_hotp_secret(otp_secret)
+        except TypeError as e:
+            if "Non-hexadecimal digit found" in str(e):
+                flash("Invalid secret format: non-hexadecimal digit found")
```
redshiftzero: Note: many journalists will be unfamiliar with the term "hexadecimal digit". Perhaps something along the lines of: "Please only submit letters A-F and numbers 0-9"

securedrop/journalist.py
```diff
@@ -242,9 +242,19 @@ def admin_reset_two_factor_hotp():
...
+            elif "Odd-length string" not in str(e)):
+                flash("Invalid secret format: odd-length string")
```
redshiftzero: The word "string" may not be clear to many end users, so we could say "odd-length secret" instead. Adding a statement like "Did you mistype the secret?" to prompt the user may help them understand what the implications of an odd-length string are.

securedrop/journalist.py
```diff
@@ -242,9 +242,19 @@ def admin_reset_two_factor_hotp():
...
+            else:
+                raise
```
fowlslegs: I think that flashing an "unexpected error" message and also logging the error is preferable to just letting Apache serve a 500 error. Logging is in general way less of a concern on the JI vs the SI, except in the context of certain interactions such as sending a reply to a Source.

Right now, we won't get much out of the fact we're logging these errors because we haven't implemented the opt-in, "securely send your logs to FPF to help us improve SD" feature. However, if/when this changes, logging such unexpected errors will be a means to help us improve exception handling, to catch bugs, and to generally increase code quality.

"Unexpected error" messaging is certainly not great, but maybe slightly better than a 500.

securedrop/journalist.py
```diff
@@ -242,9 +242,19 @@ def admin_reset_two_factor_hotp():
     otp_secret = request.form.get('otp_secret', None)
     if otp_secret:
         user = Journalist.query.get(uid)
-        user.set_hotp_secret(otp_secret)
-        db_session.commit()
-        return redirect(url_for('admin_new_user_two_factor', uid=uid))
+        try:
+            user.set_hotp_secret(otp_secret)
+        except TypeError as e:
+            if "Non-hexadecimal digit found" in str(e):
```
fowlslegs: Your tests ensure that if the exception messaging changes here, we'll detect it and be able to fix it here :+1:

dachary:
modified according to your comments & repushed, thanks for the reviews :-)

securedrop/journalist.py
```diff
@@ -242,9 +242,23 @@ def admin_reset_two_factor_hotp():
...
+                flash("Invalid secret format: "
+                      "please only submit letters A-F and numbers 0-9",
```
fowlslegs: Can you put a period at the end of the sentence here?

securedrop/journalist.py
```diff
@@ -242,9 +242,23 @@ def admin_reset_two_factor_hotp():
...
+                      "error")
+            elif "Odd-length string" in str(e):
+                flash("Invalid secret format: "
+                      "odd-length secret. Did you mistype the secret?",
+                      "error")
+            else:
+                flash("Unexpected error", "error")
```
fowlslegs: It would be better to flash an error more like https://github.com/dachary/securedrop/blob/develop/securedrop/journalist.py#L277 here. Also, this still needs logging. In previous iterations where you re-raised the `TypeError` in the `else` block, Apache would log it for us. Now, the exception will disappear, which we'd like to avoid.

securedrop/journalist.py

securedrop/journalist.py
dachary: done. I better like it too, thanks for the tip :-)

dachary:
updated & repushed

dachary:
\o/

Issue #1806, opened 2017-06-14
Improve Admin-facing messaging in `securedrop-admin` script


conorsch:
# Feature request

## Description

The new `securedrop-admin` script brought in with #1781 could be made more informative during run for first-time Admins. Since the various stages of the script take a long time to run—e.g. roughly 10 minutes for the apt installs—it's vital to display progress to Admins, otherwise it looks a lot like the script hung.

We'll be running this script a lot during QA ramping up to the 0.4 release, so tracking changes targeting that tool in this issue, to be implemented in the near future. 

## Examples
* [x] The apt install task tasks nearly 10 minutes, and displays no messages during that run time:
```
$ ./securedrop-admin setup
INFO: Installing dependencies. You'll be prompted for the temporary Tails root credentials. This was set on boot-up screen
[sudo] password for amnesia: 
dpkg-preconfigure: unable to re-open stdin: No such file or directory
INFO: Setting up virtualenv
```

* [x] The message `dpkg-preconfigure: unable to re-open stdin: No such file or directory` always displays during the apt install task. No explanation for the warning message is provided, so it's unclear whether the operation completed successfully.

* [x] Tails tries to update packages automatically, and we can detect the apt lock error message and provide a better failure mode than displaying the traceback:

```
$ ./securedrop-admin setup
INFO: Installing dependencies. You'll be prompted for the temporary Tails root credentials. This was set on boot-up screen
[sudo] password for amnesia: 
E: Could not get lock /var/lib/apt/lists/lock - open (11: Resource temporarily unavailable)
E: Unable to lock directory /var/lib/apt/lists/
Traceback (most recent call last):
  File "./securedrop-admin", line 164, in <module>
    args.func(args)
  File "./securedrop-admin", line 68, in envsetup
    libpython2.7-dev"]))
  File "/usr/lib/python2.7/subprocess.py", line 219, in check_output
    raise CalledProcessError(retcode, cmd, output=output)
subprocess.CalledProcessError: Command '['sudo', 'su', '-c', 'apt-get update &&                                             apt-get install -y                                             python-virtualenv                                             python-pip                                             ccontrol                                             virtualenv                                             libffi-dev                                             libssl-dev                                             libpython2.7-dev']' returned non-zero exit status 100
```

* [ ] Since the apt tasks take the longest of any step, we can break up the apt update and apt install tasks and provide messaging in between, as a progress indicator.

## User Stories
[...]

PR #1881, opened 2017-06-26
Ports configure logic to securedrop-admin script

### LINKED ISSUE #1857: Port `configure` logic to `securedrop-admin`
# Feature request

Port the bash logic in `install_files/ansible-base/configure` to the new `securedrop-admin` Python script. 

## Description

Consistent with the push to Pythonification described in #1409, and actually necessary now that we have dual wrappers on the `securedrop-configure.yml` playbook, because the `-f` option for clobbering vars files isn't passed any longer. See [discussion thread on 1855 for details](https://github.com/freedomofpress/securedrop/issues/1855#issuecomment-309520508).

## User Stories

As an Admin, if I make a mistake typing in variables at install time, I want a way to re-enter everything via an opt-in CLI flag. (Was already implemented in #1782, but implementation doesn't jive well with #1781.)

As an Admin, I want informative error output, so that if an error is encountered during setup, it's clear what I need to do to solve the problem.

conorsch:
## Status

Ready for review.

## Description of Changes

Fixes #1857.

The bash script at `install_files/ansible-base/configure` was already called via the `securedrop-admin`, providing a single entry point for Admins to memorize, and for the project documentation to cover. Since the bash script is not executed directly in the current docs, it's now been ported to a Python function inside the `securedrop-admin` management script. Functionality remains the same.

Related: #1409.

Does _not_ implement more user-friendly messaging as described in #1806—will open a separate PR for that.

## Testing

1. Boot into Tails with fresh persistence.
2. Run through the `./securedrop-admin setup` flow.
3. Run `./securedrop-admin sdconfig` and confirm the site-specific vars are created.
4. Run  `./securedrop-admin sdconfig` again and confirm you are _not_ prompted for the same vars (if you did not set a custom logo, you should see only one prompt, for the logo, which you can skip by pressing enter).
5.  Run  `./securedrop-admin sdconfig --force` and confirm you _are_ prompted for all the vars. Type some in differently, and confirm that you see the items as changed in the output.

## Deployment

Functionality remains the same as the old `./configure` script, so the major implications for deployment are adequate QA to confirm the changes work as described. 

## Checklist

Relying on manual testing using Tails.


dachary:
I ran the suggested tests (not to the letter but enough to get confirmation, the full log is at https://paste2.org/EyFIIXCg) and it worked.

The implementation is straightforward and lgtm too :-) Just one minor nit (or ignorance on my part). What is the env=os.environ.copy() for ? The documentation reads:

> If env is not None, it must be a mapping that defines the environment variables for the new process; these are used instead of inheriting the current process’ environment, which is the default behavior.

so it looks like a noop.

securedrop-admin
```diff
@@ -46,8 +46,21 @@ def activate_venv():
...
     activate_venv()
-    subprocess.check_call([os.path.join(ANSIBLE_PATH,'configure')],
-                            cwd=ANSIBLE_PATH, env=os.environ.copy())
+
+    ansible_cmd = [
+        'ansible-playbook',
+        # Passing an empty inventory file to override the automatic dynamic
+        # inventory script, which fails if no site vars are configured.
+        '-i', '/dev/null',
+        os.path.join(ANSIBLE_PATH, 'securedrop-configure.yml'),
```
fowlslegs: The trailing comma is unnecessary, but harmless. It is ignored by the Python interpreter. I just have noticed you using it in your code often, and just wanted to comment. I'm not sure if it's been done intentionally or not.

securedrop-admin
```diff
@@ -148,6 +161,9 @@ if __name__ == "__main__":
 
     parse_sdconfig = subparsers.add_parser('sdconfig', help=sdconfig.__doc__)
     parse_sdconfig.set_defaults(func=sdconfig)
+    parse_sdconfig.add_argument('--force', action='store_const', default=False,
+                                const=True,
+                                help="Clobber site-specific vars for reconfiguration")
```
fowlslegs: You can write this more simply as 
```
parse_sdconfig.add_argument('--force', action='store_true', help="Clobber site-specific vars for reconfiguration")
```

fowlslegs:
Just to be clear because there has been confusion in the past. I am not asserting that any of my in-line comments be "fixed."

securedrop-admin
conorsch: [PEP8 recommends it](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas) in multiline lists:

>When trailing commas are redundant, they are often helpful when a version control system is used, when a list of values, arguments or imported items is expected to be extended over time. The pattern is to put each value (etc.) on a line by itself, always adding a trailing comma, and add the close parenthesis/bracket/brace on the next line.

securedrop-admin
conorsch: Good suggestion; implemented in 190ff92ee98be91d9cfd79f79446bb5ac3792223.

securedrop-admin
```diff
@@ -46,8 +46,21 @@ def activate_venv():
...
+        '-i', '/dev/null',
+        os.path.join(ANSIBLE_PATH, 'securedrop-configure.yml'),
+    ]
+    # Honor `--force` flag to clobber existing (misconfigured) vars,
+    # otherwise read in the existing vars to supply defaults.
+    if os.path.exists(SITE_CONFIG) and args.force is False:
+        ansible_cmd.append('--extra-vars')
+        ansible_cmd.append("@{}".format(SITE_CONFIG))
+
+    subprocess.check_call(ansible_cmd, cwd=ANSIBLE_PATH, env=os.environ.copy())
```
msheiny: @dachary brought up a good question here regarding `env=os.environ.copy()` -- i'm assuming this was used for setting `ANSIBLE_ARGS` at the terminal. I wasn't able to reproduce getting them passed in with this in or out. Really this is a non-issue as far as I'm concerned. Definitely not blocking my review.

conorsch: Had thought it was necessary for the virtualenv to propagate, but minimal testing without it didn't break anything. So I'm inclined to think it's a no-op as well, as @dachary stated. Further testing required until I rip it out, though—thanks for review!

PR #1889, opened 2017-06-27
journalist: log unknown IntegrityError

dachary:
## Status

Ready for review

## Description of Changes

Unexpected IntegrityError (other than a duplicate journalist name) should be logged for debugging purposes.

## Testing

pytest -v -k test_admin_add_user_integrity_error tests/test_journalist.py

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -200,6 +200,8 @@ def admin_add_user():
                 else:
                     flash("An error occurred saving this user to the database",
                           "error")
+                    app.logger.error("Adding user '{}' failed: {}".format(
+                        username, e))
 
         if form_valid:
             return redirect(url_for('admin_new_user_two_factor',
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -198,7 +198,8 @@ def admin_add_user():
                     flash("That username is already in use",
                           "error")
                 else:
-                    flash("An error occurred saving this user to the database",
+                    flash("An error occurred saving this user to the database."
+                          " Please check the application logs.",
                           "error")
                     app.logger.error("Adding user '{}' failed: {}".format(
                         username, e))
--- a/securedrop/tests/test_journalist.py
+++ b/securedrop/tests/test_journalist.py
@@ -549,6 +549,10 @@ class TestJournalistApp(TestCase):
         mocked_error_logger.assert_called_once_with(
             "Adding user 'username' failed: (__builtin__.NoneType) "
             "None [SQL: 'STATEMENT'] [parameters: 'PARAMETERS']")
+        self.assertMessageFlashed(
+            "An error occurred saving this user to the database."
+            " Please check the application logs.",
+            "error")
 
     def test_admin_page_restriction_http_gets(self):
         admin_urls = [url_for('admin_index'), url_for('admin_add_user'),
--- a/securedrop/tests/test_journalist.py
+++ b/securedrop/tests/test_journalist.py
@@ -10,6 +10,7 @@ from flask import url_for, escape
 from flask_testing import TestCase
 from mock import patch, ANY, MagicMock
 from sqlalchemy.orm.exc import StaleDataError
+from sqlalchemy.exc import IntegrityError
 
 os.environ['SECUREDROP_ENV'] = 'test'
 import config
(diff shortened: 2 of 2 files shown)


securedrop/journalist.py
```diff
@@ -197,6 +197,8 @@ def admin_add_user():
                 else:
                     flash("An error occurred saving this user to the database",
```
fowlslegs: Can you update the flashed message too to encourage admins to check their logs? Something similar to what we use elsewhere should work. Nothing like pushing admins to report errors and help us improve exception handling.

securedrop/tests/test_journalist.py
```diff
@@ -459,6 +460,24 @@ def test_admin_add_user_without_username(self):
                                           is_admin=False))
         self.assertIn('Missing username', resp.data)
 
+    @patch('journalist.app.logger.error')
+    @patch('journalist.Journalist')
```
fowlslegs: `@patch('journalist.Journalist', side_effect=IntegrityError('STATEMENT', 'PARAMETERS', None))` should work here and you can get rid of lines 469-471 below.

securedrop/tests/test_journalist.py
```diff
@@ -459,6 +460,24 @@ def test_admin_add_user_without_username(self):
...
+    def test_admin_add_user_integrity_error(self,
```
fowlslegs: While it's not uniformly implemented, a common pattern that we try to use in our unit tests is:

```
setup block
empty line
action(s) we're testing
empty line
assertion(s)
```

So inserting a line before and after your assignment of `resp` (that assignment being the "action(s) we're testing") would bring this code into line with that style, which would be preferable. The reason for doing this is pretty simple: many including myself (and maybe you'll agree) find sticking to this pattern to aide readability and clarity of test purpose. Sometimes you'll see

```
setup block
empty line
action(s) we're testing
empty line
assertion(s)
empty line
more action(s) we're testing
empty line
more assertion(s)
```

or other variations. For more complicated tests, just go with whatever seems most clear to you.

securedrop/tests/test_journalist.py
dachary: better indeed, thanks for reminding me about this :-)

securedrop/tests/test_journalist.py
dachary: :-) I added the suggested empty lines.

securedrop/journalist.py

dachary:
@fowlslegs thanks a lot for the review :-) I made the suggested changes and repushed.

securedrop/journalist.py
```diff
@@ -195,8 +195,13 @@ def admin_add_user():
                     flash("That username is already in use",
                           "error")
                 else:
-                    flash("An error occurred saving this user to the database",
+                    flash("An error occurred "
```
fowlslegs: Starting this string on the next line might help condense the number of lines it takes up. Same goes for with `assertMessageFlashed` below.

securedrop/journalist.py
```diff
@@ -195,8 +195,13 @@ def admin_add_user():
...
+                          "saving this user to the database. "
+                          "Please check the application "
+                          "logs or inform your adminstrator.",
```
fowlslegs: Any of the routes with the `@admin_required` decorator are only accessible to admins, so you can get rid of the "or inform your administrator" here.

securedrop/journalist.py
```diff
@@ -195,8 +195,13 @@ def admin_add_user():
...
                           "error")
+                    app.logger.error("Adding user '{}' failed: {}".format(
```
fowlslegs: I don't think you should make this change here, but generally speaking it might be good to consider logging tracebacks instead of just exception messages in the application logs. @freedomofpress/securedrop-maintainers

securedrop/journalist.py

securedrop/journalist.py

securedrop/journalist.py
dachary: > I don't think you should make this change here,

Do you mean it should be done elsewhere ?

I agree it would be better to log the full stack trace. Unless I missed something (very possible ;-) it is not currently done anywhere but I'd be happy to work on that if you think it is necessary in the context of this pull request. As an admin I would very much like to have additional information in the logs whenever an error happens. Instead of nothing.

fowlslegs:
Merging w/ method 1 of #1826 considering the this is a small app code only change.

PR #1891, opened 2017-06-28
journalist: gracefully handle TOTP regeneration errors

dachary:
## Status

Ready for review

## Description of Changes

The db.Journalist.regenerate_totp_shared_secret ultimately calls os.urandom which may throw a NotImplementedError exception if the randomness source is not found. Log the error and display a user message suggesting to contact the administrator

## Testing

pytest -v -k test_admin_resets_user_totp_error tests/test_journalist.py
pytest -v -k test_user_resets_totp_error tests/test_journalist.py

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -240,9 +240,16 @@ def admin_reset_two_factor_totp():
     uid = request.form['uid']
     user = Journalist.query.get(uid)
     user.is_totp = True
-    user.regenerate_totp_shared_secret()
-    db_session.commit()
-    return redirect(url_for('admin_new_user_two_factor', uid=uid))
+    try:
+        user.regenerate_totp_shared_secret()
+    except Exception as e:
+        flash("An unexpected error occurred! Please check the application "
+              "logs or inform your adminstrator.", "error")
+        app.logger.error("Reset TOTP for '{}' failed: {}".format(uid, e))
+        return redirect(url_for("admin_edit_user", user_id=uid))
+    else:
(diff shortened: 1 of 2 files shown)


fowlslegs:
> The db.Journalist.regenerate_totp_shared_secret ultimately calls os.urandom which may throw a NotImplementedError exception if the randomness source is not found. Log the error and display a user message suggesting to contact the administrator

Has this happened to you in one of the development environments? Seems like if this happens we have much bigger problems than the need to display vaguely helpful error messages to users and log the exception.

dachary:
> Has this happened to you in one of the development environments? 

No

> Seems like if this happens we have much bigger problems than the need to display vaguely helpful error messages to users and log the exception.

This is one possible reason for an exception to reach the user. Maybe it's because the source of entropy is temporarily down ? Maybe it's because of something we can't recover from ? Maybe it will be another non fatal error in the future because the code path for TOTP regeneration is changed and new exceptions can happen ?

I think there is value is making sure the exception never leads to a traceback or blank page 500 error while still being logged for the admin to analyze.

That's me defending my pull request ;-)

fowlslegs:
I'm going to just leave this one open and not review today. Want to further discuss exception handling strategy with @redshiftzero tomorrow.

dachary:
> the Flask thing to do here is to add an errorhandler(500) on the journalist interface

It makes sense, much better. I'm closing this pull request and will open another one: it is a different approach and I feel mixing the two is going to be confusing for the reader in the future.

PR #1893, opened 2017-06-28
tests: journalist: add unit test for login hardening

dachary:
## Status

Ready for review

## Description of Changes

add unit test for login hardening and get rid of a hard coded constant while we're at it

## Testing

pytest -v -k test_login_throttle tests/test_journalist.py

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -114,7 +114,10 @@ def login():
             login_flashed_msg = "Login failed."
 
             if isinstance(e, LoginThrottledException):
-                login_flashed_msg += " Please wait at least 60 seconds before logging in again."
+                login_flashed_msg += (
+                    " Please wait at least {} seconds "
+                    "before logging in again.".format(
+                        Journalist._LOGIN_ATTEMPT_PERIOD))
             else:
                 try:
                     user = Journalist.query.filter_by(
--- a/securedrop/tests/test_journalist.py
+++ b/securedrop/tests/test_journalist.py
@@ -104,7 +104,7 @@ class TestJournalistApp(TestCase):
         resp = self.client.get(url_for('index'))
         self.assertRedirects(resp, url_for('login'))
 
-    def test_invalid_credentials(self):
+    def test_login_invalid_credentials(self):
         resp = self.client.post(url_for('login'),
                                 data=dict(username=self.user.username,
                                           password='invalid',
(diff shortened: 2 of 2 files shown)


dachary:
I deliberately chose to use **assertIn** instead of matching the entire flash message because

* it is what the other test about the login page does
* the flash message is composed and trying to match all of it is going to be somewhat fragile

that being said I'll match the entire message if you think it is best, I don't have strong feelings about that ;-)

redshiftzero:
This PR is not touching any of the config tests and it just adds web app unit test coverage and makes a small change to the web app code. As such, In the context of #1826, this is fine to merge in without running on CircleCI (Option 1 in #1826).

fowlslegs:
I'd like to review before merging. Should be able to get it done sometime in the next couple of hours.

securedrop/journalist.py
```diff
@@ -114,7 +114,10 @@ def login():
             login_flashed_msg = "Login failed."
 
             if isinstance(e, LoginThrottledException):
-                login_flashed_msg += " Please wait at least 60 seconds before logging in again."
+                login_flashed_msg += (
+                    " Please wait at least {} seconds "
```
fowlslegs: You can get rid of this leading space. Or not. Anyway, good practice of not hard-coding here :+1:

securedrop/tests/test_journalist.py
```diff
@@ -104,15 +105,36 @@ def test_unauthorized_access_redirects_to_login(self):
         resp = self.client.get(url_for('index'))
         self.assertRedirects(resp, url_for('login'))
 
-    def test_invalid_credentials(self):
+    def test_login_throttle(self):
```
fowlslegs: This is analogous to `test_throttle_login` in `test_db.py`, but tests the behavior instead of the implementation (thanks @garrettr for recently reminding me of this important distinction).

securedrop/tests/test_journalist.py
```diff
@@ -104,15 +105,36 @@ def test_unauthorized_access_redirects_to_login(self):
...
+    def test_login_invalid_credentials(self):
         resp = self.client.post(url_for('login'),
                                 data=dict(username=self.user.username,
                                           password='invalid',
                                           token='mocked'))
         self.assert200(resp)
         self.assertIn("Login failed", resp.data)
 
-    def test_valid_credentials(self):
+    def test_login_valid_credentials(self):
```
fowlslegs: Please remove this. Just kidding, I'm the guiltiest person here in terms of sneaking in unrelated fixes. It's totally fine to stay.

securedrop/journalist.py
dachary: the leading space is because it is append to the previous string

securedrop/tests/test_journalist.py
dachary: well, I separated them in two commits and it really looks funny and confusing when merged into a single diff :-)

PR #1896, opened 2017-06-28
source: adjust reply confirmation CSS

### LINKED ISSUE #1895: source: delete reply confirmation does not look good
# Bug

## Description

The CSS displaying the confirmation when a source deletes a reply does not look good (Firefox 53 & Chromium 53)

## Steps to Reproduce

* Send a message as a source
* Send a reply as a journalist
* Try to remove the reply

## Expected Behavior

Nice looking confirmation.

## Actual Behavior

delete

dachary:
## Status

Ready for review

## Description of Changes

Fixes #1895

Adjust the reply confirmation box slightly to the left and to the top so it fits in the reply frame and is nicely readable.

better

## Testing

* Send a message as a source
* Send a reply as a journalist
* Try to remove the reply

## Deployment

This is a GUI fix that has no impact on deployment.

### DIFF
--- a/securedrop/sass/source.sass
+++ b/securedrop/sass/source.sass
@@ -180,9 +180,10 @@ p#codename
   .confirm-prompt
     position: absolute
     width: 300px
-    right: 5%
-    top: -9%
+    right: 10%
     font-size: 14px
+    background: white
+    border: 1px solid $color_grey_dark
 
     p
       *


securedrop/sass/source.sass
```diff
@@ -180,8 +180,8 @@ p#codename
   .confirm-prompt
     position: absolute
     width: 300px
-    right: 5%
-    top: -9%
+    right: 10%
+    top: -20%
```
redshiftzero: When I look at current `develop` I agree that the delete reply confirmation does not render well in Tor Browser:
 
develop

With your PR, I see: 
your-pr

which definitely looks better :+1:. 

However if we set `top: 0%` (can just remove the line since it's the default behavior) then it's also nicely centered vertically:

my modification

Let me know what you think!

dachary:
> given that we require sources to use Tor Browser

Of course... silly me :-) I'll adjust as you suggested, thanks for the quick review !

dachary:
Fixed and repushed

dachary:
@fowlslegs good catch. What about with a white background ? (repushed so you can check for yourself).

white

fowlslegs:
Looks fine to me, but I don't want to make the call here.

redshiftzero:
I like that approach @dachary. My only thought is that it might look clearer/better with a light border:

screen shot 2017-06-28 at 5 30 25 pm

(added with: `border: 1px solid $color_grey_dark`)

dachary:
nicer indeed :-) Added and repushed.

redshiftzero:
Excellent, good to merge. In the context of #1826, since this is touching SASS only, I am going to merge using Option 1 described in #1826 (without running CircleCI).

PR #1898, opened 2017-06-28
Functional testing: Journalist submits reply and source deletes

redshiftzero:
## Status

Ready for review

## Description of Changes

Functional testing to test the flow of:
 * Journalist submits a reply to a source
 * Source goes in and deletes the reply

I'm adding this because a missing `</a>` tag found in #1829 caused concern that there was a regression in this flow, which was previously not functionally tested. 

## Testing

Check functional tests pass and that the flow described above is being appropriately tested

## Deployment

Just tests, no concerns for deployment


dachary:
:+1: useful and straightforward

PR #1901, opened 2017-06-28
Fix parsing of wordlists

### LINKED ISSUE #1900: Wordlists are not being parsed correctly
# Bug

## Description

`crypo_util.{words,nouns,adjectives}` all contain an empty string as their last element.

fowlslegs:
## Status

Ready for review.

## Description of Changes

Fixes #1900.

Without stripping the final newline character, each of these lists ends up with an empty string as its last element.

### DIFF
--- a/securedrop/tests/test_crypto_util.py
+++ b/securedrop/tests/test_crypto_util.py
@@ -20,6 +20,11 @@ class TestCryptoUtil(unittest.TestCase):
     def tearDown(self):
         utils.env.teardown()
 
+    def test_word_list_does_not_contain_empty_strings(self):
+        self.assertNotIn('', (crypto_util.words
+                              + crypto_util.nouns
+                              + crypto_util.adjectives))
+
     def test_clean(self):
         ok = (' !#%$&)(+*-1032547698;:=?@acbedgfihkjmlonqpsrutwvyxzABCDEFGHIJ'
               'KLMNOPQRSTUVWXYZ')
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -46,9 +46,9 @@ do_runtime_tests()
 
 gpg = gnupg.GPG(binary='gpg2', homedir=config.GPG_KEY_DIR)
 
-words = open(config.WORD_LIST).read().split('\n')
-nouns = open(config.NOUNS).read().split('\n')
-adjectives = open(config.ADJECTIVES).read().split('\n')
+words = open(config.WORD_LIST).read().rstrip('\n').split('\n')
+nouns = open(config.NOUNS).read().rstrip('\n').split('\n')
+adjectives = open(config.ADJECTIVES).read().rstrip('\n').split('\n')
 
 
 class CryptoException(Exception):


dachary:
nice catch :-1:

fowlslegs:
Restarting CircleCI because it encountered on of those intermittent failure conditions.

securedrop/tests/test_crypto_util.py
```diff
@@ -19,6 +19,20 @@ def setUp(self):
...
+        against regressions regarding not stripping file contents
+        correctly.
+        """
+        words = crypto_util.words
+        nouns = crypto_util.nouns
+        adjectives = crypto_util.adjectives
+
+        self.assertEqual(len(words), 7603)
+        self.assertEqual(len(adjectives), 8222)
+        self.assertEqual(len(nouns), 17949)
```
redshiftzero: Let's remove the asserts in lines 31-33 and keep just the assert below on line 34 - your assert in line 34 is a nice clean way to catch any null strings making it into these word lists in the future. You might also rename the test `test_word_lists_contain_no_null_strings`. (I say this because we've changed our word lists several times and for maintainability's sake, we don't want to edit this unit test every time we do that especially when these asserts are unnecessary to catch the bug in question).

fowlslegs: Change made.

fowlslegs:
Rebased on develop. Ready for re-review.

securedrop/tests/test_crypto_util.py
```diff
@@ -19,6 +19,15 @@ def setUp(self):
     def tearDown(self):
         utils.env.teardown()
 
+    def test_word_lists_imports(self):
```
redshiftzero: Can we name this something more precise, e.g. `test_word_list_does_not_contain_empty_strings`

securedrop/tests/test_crypto_util.py
```diff
@@ -19,6 +19,15 @@ def setUp(self):
...
+        """Make sure word count for each list is correct and Guard
+        against regressions regarding not stripping file contents
+        correctly.
```
redshiftzero: This comment should be removed or changed to just "No empty strings should appear in the wordlists."

securedrop/tests/test_crypto_util.py
```diff
@@ -19,6 +19,15 @@ def setUp(self):
...
+        """
+        self.assertNotIn('',   crypto_util.words
```
redshiftzero:
```
securedrop/tests/test_crypto_util.py:28:30: E127 continuation line over-indented for visual indent
securedrop/tests/test_crypto_util.py:28:30: W503 line break before binary operator
securedrop/tests/test_crypto_util.py:29:30: W503 line break before binary operator
```

fowlslegs: What is this output from?

fowlslegs: Fwiw PEP8 recommends line breaks before binary operators: https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator.

redshiftzero: It's from `pep8 securedrop/tests/test_crypto_util.py `

redshiftzero: Actually you can just ignore the comment and fix the other things. When #1913 is merged in flake8 will be in CI and that will automate this convo ;)

fowlslegs: Hopefully our PEP8 tool will follow PEP8 lol.

fowlslegs: Yeah, `flake8` does not complain about breaks before binary operators.

fowlslegs:
R e a d y  f o r  r e v i e w ! `flake8` has no complaints with this.

PR #1905, opened 2017-06-28
source: flake8 compliance

dachary:
## Status

Ready for review

## Description of Changes

makes db.py flake8 compliant

## Testing

* pytest tests

* pip install git+https://github.com/berkerpeksag/astor
* checkout develop and run cd securedrop ; python -m astor.rtrip . ; mv tmp_rtrip tmp_rtrip.orig
* checkout this pull request and run cd securedrop ; python -m astor.rtrip .
* diff -ru tmp_rtrip tmp_rtrip.orig
<pre>
diff -ru tmp_rtrip.orig/source.py tmp_rtrip/source.py
--- tmp_rtrip.orig/source.py	2017-06-28 17:29:13.956310744 +0200
+++ tmp_rtrip/source.py	2017-06-29 00:57:45.518421711 +0200
@@ -60,7 +60,8 @@
 
 
 def ignore_static(f):
-    """Only executes the wrapped function if we're not loading a static resource."""
+    """Only executes the wrapped function if we're not loading
+    a static resource."""
 
     @wraps(f)
     def decorated_function(*args, **kwargs):
@@ -177,7 +178,7 @@
     for reply in g.source.replies:
         reply_path = store.path(g.sid, reply.filename)
         try:
-            reply.decrypted = crypto_util.decrypt(g.codename, file(
+            reply.decrypted = crypto_util.decrypt(g.codename, open(
                 reply_path).read()).decode('utf-8')
         except UnicodeDecodeError:
             app.logger.error('Could not decode reply %s' % reply.filename)
</pre>

Comparing the sources normalized with the AST ensures white space changes do not modify the source behavior

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,6 +58,6 @@ script:
   # to alter the pip requirements, which would cause config tests to fail.
   - pip install -r securedrop/requirements/develop-requirements.txt
   - make docs-lint
-  - pushd securedrop; flake8 db.py store.py; popd
+  - pushd securedrop; flake8 db.py store.py source.py; popd
 after_success:
   cd securedrop/ && coveralls
--- a/securedrop/source.py
+++ b/securedrop/source.py
@@ -215,7 +215,7 @@ def lookup():
         try:
             reply.decrypted = crypto_util.decrypt(
                 g.codename,
-                file(reply_path).read()).decode('utf-8')
+                open(reply_path).read()).decode('utf-8')
         except UnicodeDecodeError:
             app.logger.error("Could not decode reply %s" % reply.filename)
         else:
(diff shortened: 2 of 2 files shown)


dachary:
rebased & added source.py to flake8 in travis

dachary:
rebased & repushed

fowlslegs:
Alright, so we've now reached the inevitable stage where checks are failing because Python dependencies aren't being upgraded in Travis. Specifically, `requests` is not being upgraded when the `development` Ansible role is run because earlier in Travis another version is installed  `development-requirements`. And then the failure to upgrade to the correct version in `test-requirements` causes testinfra to fail. Still thinking about what to do about it though...

PR #1916, opened 2017-06-29
tests: journalist: add assert to test_delete_source_deletes_submissions

dachary:
## Status

Ready for review

## Description of Changes

Add a missing assert in test_delete_source_deletes_submissions. Without it nothing verifies the intended side effect.

## Testing

pytest -v -s -k test_delete_source_deletes_submissions tests/test_journalist.py


redshiftzero:
This touches webapp unit tests only, so merging with Option 1 in #1826

PR #1923, opened 2017-06-30
Ensure compatibility between Python requirements files

### LINKED ISSUE #1807: Replace `update_python_dependencies` with Makefile
A suggestion: we should replace `update_python_dependencies` with a Makefile based on the one from this blog post: [A successful pip-tools workflow for managing Python package requirements](http://jamescooke.info/a-successful-pip-tools-workflow-for-managing-python-package-requirements.html).

The developer environment and workflow described in the blog post is so similar to our own that I think we could just drop the Makefile in with minimal changes. The only requirement is that we would have to add `pip-tools` as a dependency of the developer environment, but since we already require developers to install a set of Python dependencies as part of setting up their development environment, I think this is a totally sensible and straightforward additional requirement. The only other prerequisite is `make`, and since we use Makefiles elsewhere in the SecureDrop development environment, it is already a prerequisite.

Benefits of switching to a Makefile include:

* Simpler and easier to maintain
* Trivially supports more granular and developer-friendly workflows (e.g. just update one requirements file at a time)
* Automatically does "intelligent updating", only runs `pip-compile` when changes have been made to the corresponding `.in` file (this is a core feature of `make`)
* Almost all of `update_python_dependencies` is concerned with creating a virtualenv to install `pip-tools` in, but all of this code is unnecessary if we simply make `pip-tools` a requirement of the development environment.

### LINKED ISSUE #1918: Python dependency versions differ on development between local VMs and Travis
# Bug

## Description

In both environments, `pip` is invoked sequentially on multiple requirements files. When a particular dependency is present in more than one requirements file, the version installed will be from whichever requirements file `pip` is is first invoked on. In the local VM context, `pip` is invoked via Ansible in the `development` Ansible role first on `securedrop-requirements.txt` and then on `test-requirements.txt`. In Travis `pip` is first invoked on `testinfra/requirements.txt`, and then the `development` Ansible role is run [1].

In #1829 I added the version of `requests` in `test-requirements.txt` to the `test_development_pip_dependencies()` testinfra test, which caused Travis to fail because an older version is present in `testinfra/requirements.txt` (which, again, `pip` is invoked on first in the Travis context). I'm not sure why this and a number of the other `test-requirements.txt` dependencies are missing from the `test_development_pip_dependencies()` testinfra test, but I believe there may be even more conflicts.

I have not taken the time to check whether or not there is overlap between `testinfra/requirements.txt` and `securedrop-requirements.txt`, but if that is the case, or were to become the case in the future (it's certainly possible with how things now work) that would arguably be an even worse situation.

[1] In Travis `develop-requirements.txt` is also installed, but after the app code and testinfra tests are run, so we can mostly ignore this for the purpose of this discussion.

## Comments

I'm still brainstorming ideas on how to solve this problem in a way that is maintainable, verifiable (i.e., our tests should conclusive prove things are as they should be), easy to follow, and would keep the diff as small as possible. Please chime in if you have ideas to this extent; I intend to follow up myself in the near-term.

fowlslegs:
## Status

Ready for review.

## Description of Changes

Fixes #1918. Fixes #1807. Progress towards #1409.

Adds a new utility to ensure compatibility between our requirements file. An order of importance has been set, `securedrop > test > testinfra > develop`, to ensure that our app code dependencies are never held back in order to remain compatible with less important dependencies (instead, it's vice-versa). Whenever a change is made to a particular `.in` file, one of the new `make` commands I've introduced should be run according to the new developer documentation. For granularity, this command will only compile the file you've updated and the ones "downstream" of it if necessary, in order to maintain compatibility. Further, it detects if changes are only "cosmetic" and won't update ".txt" files if all that's changed is letter case or comments; it's mostly idempotent.

The technical documentation in `update_python_dependencies.py` is pretty good I think, and the strategy is pretty simple to understand. I've added further documentation for developers/ contributors in our developer docs. 

New testinfra tests confirm that this utility is working as intended.

Our existing Python dependencies were brought into compatibility using `make update-python-dependencies-testinfra` since there were no conflicts between the securedrop and test requirements, and I intended only to make the minimum necessary changes for compatibility.

Other changes:
* All Python packages pre-installed in Travis are uninstalled at the start of each run to ensure a clean slate.
* Ansible is unpegged in `develop-requirements.txt` as the #1146 is resolved.

I agree with @conorsch that it's better simply not to install dependencies side-by-side that belong on the same machine, but this provides a short-term solution as it may be some months before SecureDrop is able to implement and transition to the model he describes in #1918.

## Testing

Test that the changes to the testinfra and develop requirements have not broken anything and in fact show that the compatibility tool is working as intended (the new development testinfra tests will confirm this, as will Travis passing):

1. Install the new development requirements: `workon securedrop && pip-sync securedrop/requirements/develop-requirements.txt`.
2. Provision twice from a clean VM: `vagrant destroy development -f && vagrant up && vagrant provision`.
3. Run the configuration tests: `./testinfra/test.py development`.
4. Repeat steps 2-3 for the staging environment.

Make various changes to `in` files, and follow the developer documentation instructions on updating the `txt` files. Ensure the results are sane, idempotent, and granular. Changes to `testinfra-requirements.in`, e.g., should never induce changes to `securedrop-requirements.txt` or `test-requirements.txt`, if you use the make command as instructed in the new developer docs.

## Deployment

2. New installs.

I don't believe so.


fowlslegs:
Just failed because of intermittent errors related to #844 (and also CircleCI intermittent failures, which we should probably start tracking better for when our boiii @msheiny can get back on the SD CI grind).

fowlslegs:
Rebased and forced pushed--checks should pass now.

docs/development/tips_and_tricks.rst
```diff
@@ -40,21 +40,21 @@ Browser by navigating to ``127.0.0.1:8080`` and ``127.0.0.1:8081``.
...
-
+SecureDrop uses a custom Python utility to manage it's numerous requirements
+file and avoid conflicts between them. If you would like to add a Python
+dependency, instead of editing the ``securedrop/requirements/*.txt`` file or
+``testinfra/requirements.txt`` directly, please:
+
+  #. Edit the relevant ``*.in`` file in ``securedrop/requirements/`` or
+     ``testinfra/requirements.in``.
+  #. Run the apprioprate ``make`` command for the file you updated from the root
+     of the git repository. Options include
```
dachary: s/Options/Targets/ cf [A rule appears in the makefile and says when and how to remake certain files, called the rule’s targets](https://www.gnu.org/software/make/manual/html_node/Rules.html#index-target)

update_python_dependencies.py
```diff
@@ -0,0 +1,176 @@
...
+
+reqs_pref_order = collections.OrderedDict()
+reqs_pref_order['securedrop'] = sd
+reqs_pref_order['test'] = test
+reqs_pref_order['testinfra'] = testinfra
+reqs_pref_order['develop'] = develop
+
+
+def create_reqs_dict(file):
+    """Parses a requirements.txt file generated by `pip-compile` with
```
dachary: s/with//

dachary:
Why not having one requirements.in for all purposes ? It would mean more pip modules are installed and not used in various contexts. But it looks like the additional space required and the increased build time is not very significant. The complexity of maintaining separated requirements files seems to be  a recurring burden.

garrettr:
> Adds a new utility to ensure compatibility between our requirements file.

@fowlslegs Is this new utility necessary? Have you tried running `pip-compile` with multiple `*-requirements.in` files? Seems like that would achieve the outcome you are seeking, if I am  correctly understanding the issue you are trying to fix.

```
~> pip-compile --help
Usage: pip-compile [OPTIONS] [SRC_FILES]...

  Compiles requirements.txt from requirements.in specs.
```

dachary:
@fowlslegs would you like me to move forward with this ?

redshiftzero:
Agree with @dachary in [this comment](https://github.com/freedomofpress/securedrop/pull/1923#issuecomment-312765636) and @msheiny in ticket #2175. Let's just reduce all these pip requirements files in different contexts, this is getting very confusing and difficult to manage. 

Note: I am closing this PR but I am _not_ deleting this branch yet in case there are commits we want to salvage for #1807 or anything else.

msheiny:
> @fowlslegs Is this new utility necessary? Have you tried running pip-compile with multiple *-requirements.in files? Seems like that would achieve the outcome you are seeking, if I am correctly understanding the issue you are trying to fix.

I didnt see this PR -- but yeah I started doing this and can 👍 that is does indeed working .. sourcing requirements.in files from within another requirements.in (or sourcing requirements.txt from a requirements.in)

PR #1926, opened 2017-07-03
js: install jquery-2.1.4 sources

dachary:
## Status

Ready for review

## Description of Changes

Distribute the jquery-2.1.4 sources alongside the compressed version so people redistributing SecureDrop comply with the AGPL.

## Testing

No testing is necessary since only the minimized version of jquery is being used.

## Deployment

2. New installs.

This is an additional file that is not used in the code path. It will show up in existing environments and make them compliant with AGPL (should someone ask for a copy of the installed software).

### DIFF
--- a/install_files/ansible-base/usr.sbin.apache2
+++ b/install_files/ansible-base/usr.sbin.apache2
@@ -206,6 +206,7 @@
   /var/www/securedrop/static/i/tipbox/tipbox-hed-submit3.png r,
   /var/www/securedrop/static/i/tipbox/tipbox-logo.png r,
   /var/www/securedrop/static/js/journalist.js r,
+  /var/www/securedrop/static/js/libs/jquery-2.1.4.js r,
   /var/www/securedrop/static/js/libs/jquery-2.1.4.min.js r,
   /var/www/securedrop/static/js/source.js r,
   /var/www/securedrop/static/fonts/fontawesome-webfont.eot r,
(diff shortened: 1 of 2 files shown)


dachary:
@redshiftzero for legal reasons I think this should be in 0.4

install_files/ansible-base/usr.sbin.apache2
```diff
@@ -206,6 +206,7 @@
   /var/www/securedrop/static/i/tipbox/tipbox-hed-submit3.png r,
   /var/www/securedrop/static/i/tipbox/tipbox-logo.png r,
   /var/www/securedrop/static/js/journalist.js r,
+  /var/www/securedrop/static/js/libs/jquery-2.1.4.js r,
```
redshiftzero: Do we need to add this to the AppArmor rules if this is distributed only for legal reasons?

dachary: it will allow the user to load the source by changing the URL to end with **jquery-2.1.4.js** instead of the minified version which is the easiest way to comply. If the source is not available to the user who downloaded the minified version it gets more complicated.

That being said I'm not an apparmor expert and maybe there is no need for this line and the file is readable regardless ?

redshiftzero: You're right that's how AppArmor works, was just curious if making this file readable by users of SecureDrop web apps is strictly necessary to be compliant with AGPL if the uncompressed js were to be distributed only in the repo (e.g. we provide `.py` code only in the repo). Either way, this is OK, will approve for merge 🙃

dachary:
thanks !

redshiftzero:
OK staging CI [build passing](https://circleci.com/gh/freedomofpress/securedrop/739) in #1933. Merging!

PR #1928, opened 2017-07-03
Supports optional HTTPS on Source Interface

### LINKED ISSUE #1605: Provide first-class support for HTTPS on Source interface
A few organizations have started serving the Source Interface over HTTPS. We've worked with them via our support portal and provided patched Apache vhost configs for the TLS setup. Since that custom config is now stable (several orgs are using it with no reported problems so far), we should fold those features into the Ansible config as first-class features. A few considerations:

* [x] HTTPS should be opt-in
* [ ] Provide strong documentation:
  * [ ] Procurement of EV cert for .onion URL from DigiCert
  * [ ] Proving ownership of .onion URL
  * [ ] How to update the Ansible config to add HTTPS cert info
* [x] Certs must live in `/var/lib/ssl/` (see additional considerations in #949)
* [x] Ansible vars updates should not cause merge conflicts on pop/stash

At present, orgs using HTTPS on the Source Interface need to manually maintain the patched vhost configs, which is problematic for keeping the other system config up to date. Since we already plan to require Admin intervention during release of 0.4, it's a reasonable time to make this change, as well.

conorsch:
## Status

Ready for review.

## Description of Changes

Fixes #1605. Several organizations have started deploying HTTPS on the Source Interface, and we've worked with them to create custom vhost configs. Given that the config has been stable in production, we can now add top-level configuration to enable HTTPS if orgs wish to opt-in.

Most of the configuration must happen out-of-band, via manual changes on the Application Server. Since DigiCert provides only EV certificates for Onion URLs, the Admin will need to communicate with DigiCert directly and confirm ownership of the URL. The documentation provided in this PR links out to the relevant DigiCert contact pages, and explicitly mentions the Ansible vars that should be updated, but is light on details concerning the nitty-gritty of the EV process. I'm fine with that for now.

The config tests have been updated, as well. Noticed that in the Apache test suite, we were still hardcoding a few vars, rather than importing dynamically from the YAML vars file. Fixed that, and cleaned up some of the Apache templates while in there (056e96574630147a043fee65030b9d290c316c72). 

## Testing

1. Provision staging hosts as usual, and confirm you can access the Source Interface over HTTP at the Onion URL.
2. Generate a self-signed keypair (e.g. with `easy-rsa`) for use with the Source Interface. You should also generate a CA crt, for a total of three files that will land on the Application Server.
3. Update the site-specific vars as described in the new HTTPS docs. (I'm being intentionally vague here; if the docs aren't clear, request changes in the review.)
4. Rerun the config: `ANSIBLE_ARGS="--tags app,tor" vagrant provision /staging/`.
5. Confirm you can access the Source Interface over HTTPS. (You'll need to accept the self-signed cert.)
6. Confirm the HTTP URL redirects automatically to HTTPS.

## Deployment

For the few orgs already using HTTPS, we should reach out and describe the new vars flow during the 0.4 roll-out. For new orgs, the opt-in flow will default to the normal HTTP configuration, so no changes unless Admins wish to obtain an EV cert.

## Checklist

**Relying on manual testing to confirm functionality.** We don't have config testing for self-signed certs right now. In general, we're not testing multiple configuration paths in the Ansible setup, although that's something we should definitely start doing in the future.


conorsch:
The `-d` option added in f2b985b required further updates to the `securedrop-admin` script. Implemented. The change dismissed @msheiny's previous approval, but @redshiftzero will review as well, so waiting on a thumbs up.

install_files/ansible-base/roles/app/templates/sites-available/source.conf
```diff
@@ -1,5 +1,23 @@
...
+<VirtualHost 127.0.0.1:443>
+SSLEngine on
+SSLCertificateFile /var/lib/ssl/{{ securedrop_app_https_certificate_cert_src|basename }}
+SSLCertificateKeyFile /var/lib/ssl/{{ securedrop_app_https_certificate_key_src|basename }}
+SSLCertificateChainFile /var/lib/ssl/{{ securedrop_app_https_certificate_chain_src|basename }}
+
+SSLProtocol all -SSLv2 -SSLv3
+SSLHonorCipherOrder on
+SSLCompression off
+SSLCipherSuite EECDH+AES128:RSA+AES128:EECDH+AES256:RSA+AES256:EECDH+3DES:RSA+3DES:!MD5
```
redshiftzero: Will make a followup issue for updating the ciphersuites, but in the interest of incremental progress I'm not going to block merge on this

PR #1945, opened 2017-07-07
Fix update_version.sh script for 0.4

redshiftzero:
## Status

Ready for review

## Description of Changes

Closes #1942.

## Testing

How should the reviewer test this PR?
* Provision the build VM, check no snags are hit
* Make sure all relevant files are updated

## Deployment

Not used in deployment

PR #1946, opened 2017-07-07
Fixes Ansible group membership in prod VMs

### LINKED ISSUE #1943: Prod Vagrant VMs don't receive correct group membership
# Bug

## Description

Provisioning prod VMs fails because the group membership is wrong, and the `hosts: securedrop` play target matches nothing.

## Steps to Reproduce

1. `vagrant destroy -f /prod/`
2. `ANSIBLE_ARGS="--skip-tags validate" vagrant up /prod/`

## Expected Behavior

Provisioning proceeds at least until the installation of the `securedrop-app-code` deb package (at which point it will fail unless you're using rc debs for 0.4).

## Actual Behavior

Provisioning fails immediately with no successful tasks, because the first play targeted was the Monitor Server configuration play, since that is the first play referencing a host target other than "localhost" or "securedrop" (which is a group of both servers).

## Comments

This patch applied to the Vagrantfile will resolve:

```
index 543b0888..36c7f978 100644
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -124,10 +124,11 @@ Vagrant.configure("2") do |config|
       ansible.raw_arguments = Shellwords.shellsplit(ENV['ANSIBLE_ARGS']) if ENV['ANSIBLE_ARGS']
       # Taken from the parallel execution tips and tricks
       # https://docs.vagrantup.com/v2/provisioning/ansible.html
-      ansible.limit = 'all'
+      ansible.limit = 'all,localhost'
       ansible.groups = {
         'securedrop_application_server' => %w(app-prod),
         'securedrop_monitor_server' => %w(mon-prod),
+        'securedrop' => %w(app-prod mon-prod)
       }
     end
   end
```

There are two changes there:

* limit (required to get the localhost target to run)
* securedrop group (which in Tails/prod is handled by the dynamic inventory script #1774), but for non-Tails testing (e.g. virtualbox on mac), must be duplicated in the vagrantfile config.

conorsch:
## Status

Ready for review.

## Description of Changes

Closes #1943.

The changes presented in #1774 updated the prod inventory to use a dynamic inventory script, but did not include corresponding changes to the Vagrantfile for the prod VMs. Implemented here, specifically the "securedrop" group addition, which is used to target both the Application and the Monitor Servers in production.

In addition to the "securedrop" addition, also amended the "limit" option in the Ansible config for the prod VMs to include "localhost", which unintuitively is not considered when handling a limit of "all".
The explicit override enables localhost to be targeted by plays, e.g. the vars validation logic in the prod playbook.

## Testing

1. `vagrant destroy -f /prod/`
2 `ANSIBLE_ARGS="--skip-tags validate" vagrant up /prod/`
3. Confirm that the playbook executes normally.

Be advised that if you are not using the apt-test repo (for release candidate packages), then the playbook will fail upon installing the securedrop-app-code deb package at version 0.3.12 in the app role. This is OK. If you made it that far, the problem is resolved.

## Deployment

Only affects the dev env.

## Checklist

We don't run prod in CI yet, so relying on manual testing to confirm desired behavior.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -124,10 +124,11 @@ Vagrant.configure("2") do |config|
       ansible.raw_arguments = Shellwords.shellsplit(ENV['ANSIBLE_ARGS']) if ENV['ANSIBLE_ARGS']
       # Taken from the parallel execution tips and tricks
       # https://docs.vagrantup.com/v2/provisioning/ansible.html
-      ansible.limit = 'all'
+      ansible.limit = 'all,localhost'
       ansible.groups = {
         'securedrop_application_server' => %w(app-prod),
         'securedrop_monitor_server' => %w(mon-prod),
+        'securedrop' => %w(app-prod mon-prod)
       }
     end
   end


conorsch:
CI is failing because we need changes from #1945. Rebasing...

conorsch:
Rebased. Once CI checks pass, good to merge.

Issue #1944, opened 2017-07-07
Docs improvements to the HTTPS config


conorsch:
# Feature request

## Description

Received feedback out-of-band regarding the HTTPS on the Source interface docs (#1928). Summarizing here for implementation during the next docs sprint:

* [x] Make sure to include a link directly to DigiCert's docs on the process: https://www.digicert.com/blog/ordering-a-onion-certificate-from-digicert/
* [x] The validation page DigiCert will ask ask Admins to host must be accessible at `<onion_url>.onion/.well-known/pki-validation.html`.
* [x] State the 12-month max on Onion certs. Some orgs have managed to wrangle longer certs (15 months), but it took a lot of back-and-forth, and involved surprise revocations. Sticking with the 12-month max seems wise.
* [x] Encourage Patience in the docs. Since Onion certs are not yet common, the DigiCert team may try to adhere to inapplicable procedures, e.g. trying to verify an email sent to the WHOIS contacts for the domain (which won't work for Onion URLs).

Then, separate from the list above, but important in light of #1941, we should:

* [x] Explain the OCSP concerns, reference the #1941 issue, and state that we're working on improving the HTTPS config going forward.

## User Stories

As an Admin, I want to read through a frank accounting of the process to roll out HTTPS for my SecureDrop instances, so that I am informed about the trade-offs and complications involved with the switch.

garrettr:
> Make sure to include a link directly to DigiCert's docs on the process

There is already a [link](https://github.com/freedomofpress/securedrop/blame/400240b608e8a5db5ab9e8e81c0af947a81be673/docs/https_source_interface.rst#L32-L37) to Digicert's docs in the current documentation, so I consider that checklist item to be already addressed.

redshiftzero:
Addressed by #1948 and #1949, closing

PR #1948, opened 2017-07-07
Improve HTTPS documentation

garrettr:
## Status

Ready for review

## Description of Changes

Partial resolution for #1944. Addresses all of the checklist items in that issue except for the one about OCSP concerns, which @redshiftzero said she was already working on.

See commit messages for detailed list of changes.

## Testing

Build the documentation and review the revised HTTPS configuration documentation.

## Deployment

There are no special considerations for deployment, since this is a brand new feature that has not been deployed yet.

### DIFF
--- a/docs/https_source_interface.rst
+++ b/docs/https_source_interface.rst
@@ -4,9 +4,29 @@ HTTPS on the Source Interface
 The *Source Interface* for SecureDrop is automatically served over
 as a Tor Hidden Service, requiring a ``*.onion`` URL to access it.
 While Tor Hidden Services provide end-to-end encryption by default, as well
-as strong anonymity, high-profile organizations may wish to provide
-further verification to Sources that the SecureDrop instance belongs
-to the organization listed on the *Landing Page*.
+as strong anonymity, there are several reasons why you might want to consider
+deploying an additional layer of encryption and authentication via HTTPS:
+
+* The cryptographic primitives used by Tor Hidden Services are considered to be
+  outdated, and while there are no known compromises of the security of Tor
+  Hidden Services due to this issue, it may be wise to provide an additional
+  layer of transport encryption using stronger cryptographic primitives, which
+  is most easily achieved by setting up HTTPS on the Source Interface.
+
+  .. note:: This issue is being addressed by the Tor Project with their Next
+     Generation Onion Services design, but that design is still a work in
+     progress and is not expected to be deployed until December 2017 at the
+     soonest.
+
+* Extended Validation (EV) certificates, which are currently the only type of
+  certificates that may be issued for ``*.onion`` addresses, are intended to
+  attest to the identity of the organization running a service. This provides
+  an additional measure of authenticity (in addition to the organization's
+  *Landing Page* and the `SecureDrop Directory`_) to help assure sources that
+  they are communicating with the intended organization when they access a
+  given Source Interface.
+
+.. _`SecureDrop Directory`: https://securedrop.org/directory/
 
 Obtaining an HTTPS certificate for Onion URLs
 ---------------------------------------------
--- a/docs/https_source_interface.rst
+++ b/docs/https_source_interface.rst
@@ -1,22 +1,11 @@
 HTTPS on the Source Interface
 ======================================
 
-The *Source Interface* for SecureDrop is automatically served over
-as a Tor Hidden Service, requiring a ``*.onion`` URL to access it.
-While Tor Hidden Services provide end-to-end encryption by default, as well
-as strong anonymity, there are several reasons why you might want to consider
-deploying an additional layer of encryption and authentication via HTTPS:
-
-* The cryptographic primitives used by Tor Hidden Services are considered to be
-  outdated, and while there are no known compromises of the security of Tor
-  Hidden Services due to this issue, it may be wise to provide an additional
-  layer of transport encryption using stronger cryptographic primitives, which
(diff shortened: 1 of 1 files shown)


docs/https_source_interface.rst
```diff
@@ -4,45 +4,93 @@ HTTPS on the Source Interface
...
+``<onion_url>.onion/.well-known/pki-validation.html``. You will need to
+understand Apache configuration well enough to know how to temporarily
+provision such a page at this specific URL. Once you have received your
+certificate, you should restore the Apache configuration to the stock
+SecureDrop configuration.
+
+While the `CAB forum`_ has specified that ``.onion`` certificates may have a
+maximum lifetime of 15 months, we have heard that some folks have run into
+issues with such certificates, and currently it seems safest to give the
+certificte a validity period of 12 months.
```
redshiftzero: `certificte` -> `certificate`

docs/https_source_interface.rst
```diff
@@ -4,45 +4,93 @@ HTTPS on the Source Interface
...
+     progress and is not expected to be deployed until December 2017 at the
+     soonest.
+
+* Extended Validation (EV) certificates, which are currently the only type of
+  certificates that may be issued for ``*.onion`` addresses, are intended to
+  attest to the identity of the organization running a service. This provides
+  an additional measure of authenticity (in addition to the organization's
+  *Landing Page* and the `SecureDrop Directory`_) to help assure sources that
+  they are communicating with the intended organization when they access a
+  given Source Interface.
```
redshiftzero: Even though the crypto in onion services is a bit aged, as you correctly state, in my view the reason why we want to allow HTTPS on the source interface is the second reason you describe here: to demonstrate to sources that they are communicating with the onion service that is indeed controlled by the news organization they want to talk to. 

While an additional layer of encryption is a nice bonus, if we really believed that there may be a break in onion service crypto in the wild and that this was a likely vector of attack, then we'd make HTTPS on all interfaces a requirement and not optional. Basically my concern is making sure that SecureDrop is not pushing difficult security decisions to users - in this case the security decision "is Tor onion service crypto sufficient for my purposes?" is not a question that most users are in a position to answer (but we are, and we answered that question yes). That said, I think we should remove the comments about onion service crypto but leave your clear description in the second point. 

Let me know your thoughts!

(Btw: I'm cool with putting an small OCSP note for full disclosure and because it will be gone soon in the next minor release, but generally I think we should just instruct users on what the right decision is and then make that the default)

garrettr: > While an additional layer of encryption is a nice bonus, if we really believed that there may be a break in onion service crypto in the wild and that this was a likely vector of attack, then we'd make HTTPS on all interfaces a requirement and not optional.

All I can say here is that the fact that Tor Hidden Service's crypto primitives would be considered insecure if the system were being designed today has been and continues to be a strong motivation for the administrators who have already deployed HTTPS on their SecureDrops. Even if there is not a break right now, this is still a concern in the context of adversaries who may be storing Tor traffic with the aim of trying to break the crypto at a later date.

I think it would great to make HTTPS on all SecureDrop's required, or at least strongly recommended. The only reasons why we haven't done so already are:

1. It wasn't even possible to get a cert for an .onion until 2015
2. After the initial .onion certs were issued, there was a lot of debate as to whether they would be approved by the relevant standards bodies, and this approval process took a long time. We decided to wait and see because we didn't want to tell administrators to spend money on certificates, only to find that they would stop being supported.
3. Maintaining an HTTPS deployment over time requires a significant amount of technical know-how and attention from the SecureDrop instance administrators. They need to pay attention to cert expiry and a whole host of other issues. Due to SecureDrop's decentralized architecture, we were leery of encouraging people to do this unless we felt that were capable of managing the lifecycle of their HTTPS deployment on their own.

Finally, the Tor Project is aware of this issue, have already improved the cryptographic primitives in some areas of the Tor protocol due to this exact concern, and hidden services will start using a modern set of primitives once the next generation design is deployed. HTTPS is pretty much just a stop-gap until the next-generation design is available, but I still think it is worthwhile.

So: I think this potential motivation for deploying HTTPS on a SecureDrop instance should be kept in because it is a meaningful defense-in-depth measure.

> In my view the reason why we want to allow HTTPS on the source interface is the second reason you describe here: to demonstrate to sources that they are communicating with the onion service that is indeed controlled by the news organization they want to talk to.

In theory, yes, but of course EV certificates have a whole host of problems: most people don't know what they mean, the EV validation process has had issues in the past, and there are generally trust issues in the context of Certificate Authorities. So I think this is a helpful additional way to authenticate a SecureDrop instance that complements the other methods we've already developed (landing page, directory, printing it in the print publication), but I do not think of it as the primary motivation to deploy HTTPS on the Source Interface.

redshiftzero: Yep, I totally agree that the Tor Onion Service crypto is weak. My concern was primarily pushing this difficult crypto decision onto users and whether it makes sense to *strongly* encourage admins to migrate over to HTTPS primarily _for crypto reasons_ at the present time given that 1) as you correctly note, the crypto is being fixed upstream and 2) imho there are other more practical attacks that can compromise sources than breaking THS crypto that we might want to focus admin energy on. 

With that said, how about we swap around these two bullets: authenticity first, crypto second. We can generally leave the discussion on Onion Service crypto in place (with one or two minor edits inline, let me know your thoughts on those). Otherwise, let's merge this in and we can revise further in a future PR if necessary.

PR #1952, opened 2017-07-08
USB Printing on Tails 3.0

### LINKED ISSUE #1149: Conflicting documentation on difficulty of finding Tails-compatible printers
I noticed that the [Tails printing guide](https://github.com/freedomofpress/securedrop/blob/develop/docs/tails_printing_guide.md) states

> Normally, any printer should work with Tails "out of the box",

while the [printer section](https://github.com/freedomofpress/securedrop/blob/develop/docs/hardware.md#printers) of the hardware guide states

> Unfortunately, it is difficult to find printers that work with Tails, and it is increasingly difficult to find non-wireless printers at all.

The latter section was written later, and, at least in my experience attempting to set up various printers with GNU/Linux, it's much closer to the truth. You might want to link to or port some of the info from the hardware doc to the Tails printing doc.

Another thing you may wish to add to either of these pages is that http://www.openprinting.org/printers provides a database of Linux-compatible printers and you may wish to link to it.Tails does not necessarily ship with all the drivers necessary to run the printers in the database, but for older printers it would be a good indicator.

justintroutman:
## Status

Ready for review

## Description of Changes

Fixes #1149 insofar as it further expounds on installation for a new test case (e.g., non-wireless USB printer in Tails 3.0), and reorganizes the headings to document both printing via the terminal, as well as the GUI for both seamlessly auto-detected and installed printers, and those for which you must manually select drivers.

* 3c16996: 
Added screenshots showing the UI for installing an auto-detected USB printer in Tails 3.0 (affected file: `tails_printing_guide.rst`)

* 8817545: 
Restructured headings, added steps for printing via `lp` in the terminal, and added labels for the aforementioned screenshots. (affected file: `tails_printing_guide.rst`)

* c620830: 
Added new entry for tested printer (HP Deskjet F4200), and cleaned up the table of tested printers to be more compact (i.e., get rid of horizontal scrolling bar) and not included quickly outdated data (i.e., Amazon pricing). (affected file: `hardware.rst`)

* bde18a4:
Added steps and screenshots to document the process of installing a non-wireless USB printer that Tails 3.0 auto-detects without any need for manually selecting drivers; this uses the screenshots and labels included in the first two commits listed above.

- [x] Doc linting passed locally via `make docs-lint`


docs/tails_printing_guide.rst
```diff
@@ -7,12 +7,43 @@ potentially untrusted network. This is why **we strongly recommend using
...
-Normally, any printer should work with Tails "out of the box." Most
-difficulties stem from not selecting the right driver (extra software
-needed for the printer and computer to communicate). Luckily, Tails has
-a large number of drivers for just about any popularly manufactured
-printer on hand, without even having to download new drivers from the
-web.
+Installing and Printing via the Tails GUI
+-----------------------------------------
+
+Normally, any printer should work with Tails "out of the box." Let's look at
```
garrettr: While you're making changes to the printer docs, you might as well resolve #1149 by updating this sentence to more accurately reflect reality. The core concern of #1149 is that the printing guide says "any printer should work with Tails out of the box," which conflicts with language elsewhere in the documentation and with our experience generally. I suggest revising this sentence to something like:

> Finding a printer that works with Tails can be challenging because Tails is based on the Linux operating system, which often has second-class hardware support in comparison to operating systems such as Windows or macOS. We maintain a list of printers that have personally tested and gotten to work with Tails in the Hardware guide [link to printers section of hardware guide]; if possible, we recommend using one of those printers. In any case, this document outlines the usual set of steps that we follow when attempting to use a new printer with Tails, and provides some troubleshooting tips that you may find useful if you are trying to use a different printer.

docs/tails_printing_guide.rst
```diff
@@ -7,12 +7,43 @@ potentially untrusted network. This is why **we strongly recommend using
...
+Tails 3.0's typical flow for installing a USB-connected printer.
+
+First, you'll want to open your "Settings" panel, and double-click "Printers".
```
garrettr: Just to be super clear, I think you need to add a step prior to this one which says "use a USB cable to connect your printer to the computer running Tails, and turn the printer on."

docs/hardware.rst
```diff
@@ -408,13 +408,15 @@ To assist you, we have compiled the following partial list of
...
-+-------------------------+----------------+------------------+--------------------+--------------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------+
++-------------------------+--------------+----------------+--------------------+
+| Printer Model           | Testing Date | Tails Versions | Printer Type       |
++=========================+==============+================+====================+
+| HP Deskjet F4200        | 06/2017      | 3.0            | Color Inkjet       |
++-------------------------+--------------+----------------+--------------------+
+| HP LaserJet 400 M401n   | 06/2015      | 1.4            | Monochrome Laser   |
++-------------------------+--------------+----------------+--------------------+
+| HP Deskjet 6940         | 04/2015      | 1.3.2          | Monochrome Injket  |
++-------------------------+--------------+----------------+--------------------+
```
garrettr: Excellent. I'm really happy with the addition of the HP Deskjet F4200, so we have an up-to-date recommendation to provide for a printer that is known to work with Tails. I mostly like how you revised the layout of this table. I think it's ok to remove the price recommendations, since those tend to fluctuate a lot in my experience. However, I'm not sure if you want to remove the **Notes** column—or, if you do, you should make sure that information is preserved elsewhere in the documentation.

justintroutman: I agree -- this was also documented in `tails_printing_guide`, in the first set of screenshots, which pertain to the HP LaserJet 400 M401n; however, I agree this should be retained on the `hardware` page, since this is where we provide recommendations. As such, I added a `.. note::` to reflect notes associated with the two models we document fully in `tails_printing_guide` -- the Deskjet F4200 and LaserJet 400 M401n. I put this under the table of recommendations.

docs/tails_printing_guide.rst
justintroutman: Agreed. Simple. Done.

docs/tails_printing_guide.rst
justintroutman: Agreed. Your suggestion was great, so I integrated that, along with a description of the Linux Foundation's OpenPrinting database, and how to best use that for identifying other potentially suitable printers. (H/T @fowlslegs)

docs/hardware.rst
```diff
@@ -408,13 +408,26 @@ To assist you, we have compiled the following partial list of
...
+
+.. note:: We've documented both the HP Deskjet F4200 and HP LaserJet 400 M401n
+         with screenshots of the installation process, in our section on
+         `setting up printers in Tails
+         <https://docs.securedrop.org/en/stable/tails_printing_guide.html>`_.
```
garrettr: I recommend separating the link and target definition as much as possible; it generally makes the `.rst` source easier to read. See [docs](http://www.sphinx-doc.org/en/stable/rest.html#hyperlinks).

docs/hardware.rst
```diff
@@ -408,13 +408,26 @@ To assist you, we have compiled the following partial list of
...
+         While the F4200 installed automatically, the 400 M401n required that we
+         set "Make and model" to "HP LaserJet 400 CUPS+Gutenprint v5.2.9" when
+         manually configuring the
+         drivers.
```
garrettr: Two style notes:

1. The line wrapping on the last line is funky; `drivers.` should not be on its own line.
2. I recommend indenting lines after the first line in a directive to the same level as the marker, which is 3 spaces in this case (the marker is `.. `). I know we don't follow this style convention throughout the docs, but I'd like to change that (after 0.4, of course). The reStucturedText [docs on indentation](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#indentation) say it is a requirement:

    > Several constructs begin with a marker, and the body of the construct must be indented relative to the marker.

docs/tails_printing_guide.rst
```diff
@@ -7,12 +7,75 @@ potentially untrusted network. This is why **we strongly recommend using
...
+          whether or not a printer is wireless, so this will involve manually
+          checking models of interest, if you wish to use this resource as a
+          guide for purchasing a non-wireless printer suitable for use with
+          SecureDrop.
+
+With that in mind, this database is arguably the best resource for researching
+the compatibility of printers with Linux. As a tip for narrowing down your
+search, look for printers that are compatible with Debian, or Debian-based
+distributions like Ubuntu, since Tails itself is also Debian-based. This might
+increase the chances for a successfully seamless installation experience in
```
garrettr: "successfully seamless" is weird. Pick one 😄

docs/tails_printing_guide.rst
```diff
@@ -7,12 +7,75 @@ potentially untrusted network. This is why **we strongly recommend using
...
+
+Installing and Printing via the Tails GUI
+-----------------------------------------
+
+Let's look at Tails 3.0's typical flow for installing a USB-connected printer.
+
+First, connect the printer to your Tails-booted computer via USB, then turn the
+printer on.
+
+Now you'll want to open your "Settings" panel, and double-click "Printers".
```
garrettr: You need to explain how to find the **Settings** panel (it's in the **Applications** menu under **System Tools**). Check out how Tails does it in their [Printing docs](https://tails.boum.org/doc/sensitive_documents/printing_and_scanning/index.en.html).

Also, our [documentation guidelines](https://docs.securedrop.org/en/latest/development/documentation_guidelines.html#style-guide) require you to **bold** user interface elements, not "quote" them.

garrettr: You only need to single-click the **Printers** button. I know this sounds overly pedantic, since in this case a double-click seems to work just as well, but I've definitely done trainings where the difference between a single-click and a double-click tripped people up (e.g. trainer says to double-click a button that only needs a single-click, user double-clicks, the second click accidentally hits an element on a new page that was displayed after the first click, confusion ensues).

docs/tails_printing_guide.rst
```diff
@@ -7,12 +7,75 @@ potentially untrusted network. This is why **we strongly recommend using
...
+attempting to use a new printer with Tails, and provides some troubleshooting
+tips that you may find useful if you are trying to use a different printer.
+
+Installing and Printing via the Tails GUI
+-----------------------------------------
+
+Let's look at Tails 3.0's typical flow for installing a USB-connected printer.
+
+First, connect the printer to your Tails-booted computer via USB, then turn the
+printer on.
```
garrettr: Do we want to say "boot tails with persistence enabled, and set an admin passphrase" here (like we do at the start of the troubleshooting section)? Do you need admin privileges to add a printer? Or do you only need them to modify printer settings?

Even if you only need them to modify printer settings, I might advocate setting an admin passphrase anyway since if you encounter printer problems (which one often does on Tails), you'll need an admin passphrase, and it's annoying to get to that point and realize you have to reboot Tails just so you can set an admin passphrase.

Note that we encourage enabling persistence because Tails has printer persistence, so if persistence is enabled when you set up a printer your printer settings will be preserved across reboots.

docs/tails_printing_guide.rst
```diff
@@ -7,12 +7,75 @@ potentially untrusted network. This is why **we strongly recommend using
...
-needed for the printer and computer to communicate). Luckily, Tails has
-a large number of drivers for just about any popularly manufactured
-printer on hand, without even having to download new drivers from the
-web.
+Finding a printer that works with Tails can be challenging because Tails is
+based on the Linux operating system, which often has second-class hardware
+support in comparison to operating systems such as Windows or macOS.
+
+We `maintain a list of printers
+<https://docs.securedrop.org/en/stable/hardware.html#printers>`_ that we have
```
garrettr: When you are linking to other parts of the SecureDrop documentation, use [`:ref:`](http://www.sphinx-doc.org/en/stable/markup/inline.html#ref-role) or [`:doc:`](http://www.sphinx-doc.org/en/stable/markup/inline.html#role-doc). Doing so is more semantically correct, and also preserves the ability to build all of the output targets supported by Sphinx without breakage.

docs/tails_printing_guide.rst
```diff
@@ -7,12 +7,75 @@ potentially untrusted network. This is why **we strongly recommend using
...
+receive no errors in this process, you will then see the following screen,
+which indicates that the printer is "ready" for printing.
+
+|printer ready|
+
+For instances where your printer does *not* work out of the box, most difficulties
+stem from not selecting the right driver (extra software needed for the printer
+and computer to communicate). Luckily, Tails has a large number of drivers for
+just about any popularly manufactured printer on hand, without even having to
+download new drivers from the web.
```
garrettr: This seems like a natural break point where it would be appropriate to add another section heading. The last section was a walk-through of adding a printer successfully, while the next section is about troubleshooting a printer that doesn't work out of the gate. I would add a "Troubleshooting" section heading here.

docs/tails_printing_guide.rst
garrettr: According to @harlo, this may have changed in Tails 3. It seems admin privileges were required to add a printer in Tails 1.x, but they are no longer required in Tails 3.x. That would seem to be confirmed by this [resolved Tails issue](https://labs.riseup.net/code/issues/8443)

@justintroutman Can you test and confirm that Tails 3 does not need an admin passphrase to add a printer? If that is the case, then you should also update the Troubleshooting section of this document to remove the now-outdated and incorrect recommendations to set an admin password upon boot in order to print.

docs/hardware.rst
justintroutman: The example in those docs are much cleaner, I agree. This is another instance where I quickly looked at other .rst files in our docs for structural guidance, but this suggests some worthwhile clean-up in the future, for readability. I'll make the change.

docs/hardware.rst
justintroutman: Odd; this one I was at least aware of (indenting). Going to say this is residue from the unwieldy link/target style referenced in your previous comment. I'll clean this up.

docs/tails_printing_guide.rst
justintroutman: Fair; will note this to fix elsewhere in docs. Good to know it preserves build-without-breakage.

docs/tails_printing_guide.rst
justintroutman: Seamless it is!

docs/hardware.rst
garrettr: @justintroutman Yup, I would like to do some cleanup of the reStructuredText style conventions used throughout our documentation, and I'm sorry that the current lack of consistency was confusing for you. I'll file an issue with some notes about what stylistic conventions I think we should adopt going forward, and we can address this holistically after the 0.4 release.

docs/tails_printing_guide.rst
garrettr: Additional note: this actually matters even if you only care about building the HTML output target (e.g. for hosting on ReadTheDocs). Imagine the following scenario:

1. You are a developer working on the documentation.
2. You are editing `.rst` files in your editor and observing the corresponding changes to the HTML output in your browser (thanks to the sweet auto-reload provided by `make docs`).
3. You click this link in your local rendered copy of the HTML output.

You would want this link to go to the relevant section of *your local copy* of the HTML documentation, but instead it would take you to the externally hosted version of the documentation, which might differ significantly from your local copy (and would be confusing/incorrect behavior in any case).

garrettr:
@justintroutman Your Travis CI tests are currently failing due to #1971. To get them to pass (a prerequisite for merge), you'll need to rebase or merge your feature branch with `release/0.4` to get the latest commits, which include a fix from #1973.

docs/tails_printing_guide.rst
justintroutman: I confirmed that Tails 3 does not need an admin passphrase to add a printer. I'll update the docs to reflect this.

docs/tails_printing_guide.rst
justintroutman: That's fair -- upon re-testing, I realized I was clicking too quickly to notice the effect of a single click. Good call.

docs/tails_printing_guide.rst
justintroutman: After some research, I've used this convention, per the official Sphinx docs on `:ref`:

Here are the steps:

* Add the reference point `.. _tails_printing_guide:` above the section titled `Setting up a Printer in Tails` in `tails_printing_guide.rst`.

* In `hardware.rst`, link to this reference point by including `` :ref:`tails_printing_guide` `` within the paragraph text, which produces a link that inherits the title of the section being referred to. (One can customize the link text, but leaving it as-is seemed appropriate here.)

Here's the result:

ref

This passes linting and, as you pointed out, ensures that developers running a local copy will *only* be pointed to local copies of files; if this method is good etiquette, I'll employ it going forward. That, and make note of it for consideration when we further discuss revisions to the style guide.

garrettr: @justintroutman That's exactly how internal links within the documentation should be handled, well done 👍 . You can also grep for `:ref:` and `:doc:` to see additional examples of how the existing documentation handles internal links.

justintroutman:
@garrettr Inadvertent flub on my part with the merge commit; @redshiftzero and I will resolve.

redshiftzero:
Ready for re-review

docs/tails_printing_guide.rst
conorsch: I'm still seeing "successfully seamless" in the latest version of the docs here.

docs/tails_printing_guide.rst
```diff
@@ -1,43 +1,115 @@
...
+Let's look at Tails 3.0's typical flow for installing a USB-connected printer.
+
+If you've enabled persistence, boot with your persistent volume, and set an
+admin passphrase.
+
+Connect the printer to your Tails-booted computer via USB, then turn the
+printer on.
+
+Now, you'll want to single-click your way through **Applications** ▸
+**System Tools** ▸ **Settings** ▸ **Printers**.
```
conorsch: Line break between every sentence strikes me as excessive, and hurts readability. This is a formatting nit—won't block merge for it.

conorsch: Resolved in 3861a55aba134da0d1500a72877cbeb455a28ad5.

docs/tails_printing_guide.rst
conorsch: Resolved in 3861a55aba134da0d1500a72877cbeb455a28ad5.

conorsch:
Thanks for your help with the branch cleanup, @fowlslegs!

PR #1973, opened 2017-07-13
[0.4] Pin version of SASS ruby gem to 3.4.23 in development and CI

redshiftzero:
## Status

Ready for review

## Description of Changes

Cherry picking fix for issue #1971 contained in PR #1972 into `release/0.4`. 

## Testing

If builds pass, this is good to merge.

## Deployment

Does not run in prod, so no issues for deployment

### If you made changes to the app code:

- [ ] Unit and functional tests pass on the development VM

### If you made changes to the system configuration:

- [ ] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

PR #1974, opened 2017-07-13
Fix SSH connections breaking when tor is restarted

### LINKED ISSUE #1969: Tor Sandbox mode - breaking tor service on app/mon
# Bug

## Description

During an install sync from an admin workstation  -- tor is not coming back up correctly and I'm unable to re initiate connection.

## Steps to Reproduce

* Upgrade a `0.3.12` SD instance to `0.4rc1` by switching the apt repos to point to `apt-test.freedom.press`
* Boot into a `3.0.1` with persistence
* Ensure the latest `release/0.4` branch is cloned into ~/Persistent/securedrop
* Ensure the settings in `install_files/ansible-base/groups_vars/all/site-specific` match the settings on your SD instance.
* Drop all the relevant *ths files in `install_files/ansible-base` accordingly.
* Jump into the virtualenv (`source .venv/bin/activate`)
* cd into `install_files/ansible-base`, and run `./securedrop-prod.yml -t tor --diff -vv`

## Expected Behavior

Playbook will finish with minimal changes, and any changes that come up should not affect SD service permanently. Tor should reboot and still be accessible.

## Actual Behavior

Ansible is bombing out on the `Waiting for SSH connection (slow)...`. Before that you can see that `Sandbox 1` is getting added to the `/etc/tor/torrc` file and triggering a tor restart.

## Comments

We need to do two things:

* swap the `pause` logic with something smarter that takes advantage of `retries` and `command` (see https://github.com/freedomofpress/securedrop/blob/develop/devops/playbooks/aws-ci-teardown.yml#L48 and http://docs.ansible.com/ansible/playbooks_loops.html#do-until-loops). 

* consider ripping out the `sandbox 1` option before the 0.4 release. I'm not convinced yet this has seen enough testing internally to warrant it'll work in prod without incident.

conorsch:
## Status

Ready for review.

## Description of Changes

Since SSH connections in SecureDrop are routed over Authenticated Tor Hidden Services (ATHS), restarts of the tor daemon will sever existing SSH connections, thereby breaking Ansible playbook runs if a handler to restart tor is notified. Although we take care to ensure that tor service restart action completes successfully (#1707), the subsequent reconnection by Ansible is affected by SSH ControlPersist files.

When tor is restarted, the SSH connection is never formally severed, so SSH reasonably believes that the ControlPersist files on the Admin Workstation represent still-active connections to the remote machines. Once tor is restarted, this is no longer true. The solution is to clear out any existing ControlPersist files whenever tor is restarted. Subsequent Ansible tasks will then be forced to open a new ControlMaster, and fresh ControlPersist files will be automatically created, and the play will continue as normal.

Fixes #1969.

## Testing

1. Provision prod VMs on 0.3.12.
2. Set up access to the apt-test repo for 0.4-rc1 debs.
3. Update to 0.4-rc1 via `cron-apt -i -s`.
4. Checkout the release/0.4 branch. 
5. On the Tails Admin Workstation, run `./securedrop-admin setup` (to get the latest tooling).
6. Run `./securedrop-admin install` and confirm no breakage of tor.

There's a shortcut to 6) that is sufficient for testing: manually activate the virtualenv and run `ansible-playbook -vv --diff securedrop-prod.yml --tags tor` to isolate the tor-related tasks. 

## Deployment

Ensures that Admins upgrading from 0.3.12 to 0.4 can run the Ansible playbooks without breakage. 

## Checklist
Relying on manual testing of the Tails Admin Workstation environment, as well as upgrade-specific testing that does not exist in CI.

### DIFF
--- a/install_files/ansible-base/roles/tor-hidden-services/handlers/restart-tor-carefully.yml
+++ b/install_files/ansible-base/roles/tor-hidden-services/handlers/restart-tor-carefully.yml
@@ -25,11 +25,26 @@
   poll: 0
   when: _hostname_to_wait_for.endswith('.onion')
 
-# Hardcode a wait value and wait for the entire interval.
-# Cannot use `local_action/wait_for` because it's not smart
-# enough to find the SSH connection info required to connect
-# to the Onion URL.
-- name: Waiting for SSH connection (slow)...
-  pause:
-    seconds: 90
+# It's critical that we eliminate existing SSH connections, otherwise Ansible
+# will try to reuse the stale ControlPersist files, which no longer have an
+# active socket, causing a connection timeout, and halting execution of the
+# playbook with a cryptic message.
+- name: Clear out SSH sessions, to prevent reuse of stale ControlPersist file.
+  local_action:
+    module: file
+    args:
+      path: /home/amnesia/.ansible/cp
+      state: absent
+  run_once: yes
+  sudo: no
   when: _hostname_to_wait_for.endswith('.onion')
+
+- name: Waiting for SSH connection (slow)...
+  local_action: wait_for
+  args:
+    host: "{{ _hostname_to_wait_for }}"
+    port: "{{ ansible_ssh_port|default(ansible_port|default(22)) }}"
+    delay: 30
+    search_regex: OpenSSH
+    state: started
+  sudo: no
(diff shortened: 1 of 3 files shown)


msheiny:
rebased thanks to #1973

conorsch:
Thanks for the cleanup, @msheiny. Tests are failing because the `Sandbox` declaration was removed (as discussed, yes), but the config test suite wasn't updated. I'll fix that presently and push back up so we have green lights across the board in CI.

PR #1981, opened 2017-07-14
[0.4] source: not found page content is empty

dachary:
## Status

Ready for review

## Description of Changes

The #content id is set to hide its content by default, which is good
when used to hide the codename in the lookup.html page. Unfortunately
the same id (#content) is also used in the error.html and notfound.html
templates and hide their content.

Remove the extra <div id="content"> in both error.html and
notfound.html. Also add a functional test to verify the message is
visible when notfound.html is displayed.

## Testing

* navigate to the url /unlikely in the source interface
* see that it is an empty page

notfound

* apply the patch and see the not found page contains the expected text

expected-notfound

## Deployment

This is a GUI only issue with no deployment side effect

### DIFF
--- a/securedrop/source_templates/error.html
+++ b/securedrop/source_templates/error.html
@@ -1,11 +1,8 @@
 {% extends "base.html" %}
 {% block body %}
-<div id="content">
-
 <h1>Server error</h1>
 
 <p>Sorry, the website encountered an error and was unable to complete your request.</p>
 
 <p><a href="/login">Look up a codename...</a></p>
-</div>
 {% endblock %}
--- a/securedrop/source_templates/notfound.html
+++ b/securedrop/source_templates/notfound.html
@@ -1,11 +1,8 @@
 {% extends "base.html" %}
 {% block body %}
-<div id="content">
-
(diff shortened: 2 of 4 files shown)


securedrop/sass/source.sass
```diff
@@ -66,9 +66,6 @@ input#show
 #content
```
redshiftzero: It looks like this `content` id is also used in `lookup.html` on the source interface - what do you think about instead removing the surrounding `<div id="content">` for both `error.html` and `notfound.html`? It looks like that change fixes this and renders well:

screen shot 2017-07-16 at 12 43 04 pm

redshiftzero:
Also, we should add a functional test for `/unlikely` to check the text is displayed in the future

dachary:
@redshiftzero thanks for the review, I should have caught the other instance of hidden content in error.html. The functional test was added. The **id** was added to locate the message. The other way would be to verify part of the message is found. But this is fragile when the text changes and when the function is re-used for screenshots of l10n pages.

dachary:
travis seems to be in bad shape

Issue #1569, opened 2017-02-09
Default to Tor Browser for functional tests, keep Firefox as fallback


nvesely:
Since we're expecting sources to be using Tor Browser, it makes the most sense to be using it in our functional tests as well. When this project was started, there wasn't much in the way of tools to do this, but we've been using https://github.com/webfp/tor-browser-selenium for at least 6 months in https://github.com/freedomofpress/fingerprint-securedrop w/o issues. In fact, due to the slower moving nature of TB, it has been much more stable to drive than using Selenium with Firefox.

My plan is to keep Firefox as a fallback browser (just in case), especially because while tor-browser-selenium has been pretty well maintained over the last year, it is less likely to have the longevity of the Selenium project. It just seems good to have a backup in case of kinks. I'll wait until #1527 gets merged because I want the fallback browser to be callable like so: `./manage.py functional --all --fallback-browser -- --pdb -x`.

One thing to note is that `tbselenium` depends on `selenium<3`. So this potential conflicts with https://github.com/freedomofpress/securedrop/issues/1502 unless we can find a non-hacky way install and use two versions of Selenium. I think it's more important that we run TB for tests than use Selenium 3, as long as things are stable.

fowlslegs:
Updated tentative plans:

* Let the [`add-selenium-3-support`](https://github.com/freedomofpress/securedrop/compare/add-selenium-3-support) branch sit for the time being.
* Install `tbselenium` (0.2.1) and let that pull in the correct version of `selenium`.
* Install TBB 6.5 (latest stable)--we can steal the Ansible from https://github.com/freedomofpress/fingerprint-securedrop.
* Install FF 45.7.0esr (on which TBB 6.5 is based) as a fallback (this should work with the `selenium` that `tbselenium` pulls in if `tbselenium` is not driving TBB correctly for w/e reason).
* Implement the switch into the manage module as above.
* If/when `tbselenium` supports `selenium>=3` (there's a branch there working on this with and I might just do it myself if it hasn't happened in another month or two), we fix up/ cannibalize [`add-selenium-3-support`](https://github.com/freedomofpress/securedrop/compare/add-selenium-3-support) (mostly we want the functional test fixes, assuming things will break during this transition).

Thoughts on this plan?

tkell:
I like this idea, and agree that tests should match the desired environment.  I can take a run at this?

fowlslegs:
[...]

PR #1985, opened 2017-07-17
[0.4.1] source: fix inconsistent output when exiting from use-tor

### LINKED ISSUE #1978: inconsistent display when exiting /use-tor
# Bug

## Description

The display shows both "you are not using tor" and "Please select "New Identity" from the green Onion button in the Tor browser".

## Steps to Reproduce

* navigate to the source interface with firefox
* click on the "Learn how to install it" link
* click on the exit button

## Expected Behavior

expected

## Actual Behavior

inconsistent

dachary:
## Status

Ready for review

## Description of Changes

Fixes #1978 

When exiting a session from /use-tor, the login page shows
a flash message recommending to do some tor specific actions. This
should not be displayed if not using tor.

## Testing

The test environment does not have a tor browser and pretending firefox is a tor browser is non trivial. The change can be tested manually:

* in the development vm ./manage.py run
* with a web browser that is not tor
  * go to http://127.0.0.1:8080
  * go to http://127.0.0.1:8080/generate
  * go to http://127.0.0.1:8080/use-tor
  * click exit
  * verify the display is as follows:

good

* with a tor web browser
  * go to http://127.0.0.1:8080
  * go to http://127.0.0.1:8080/generate
  * go to http://127.0.0.1:8080/use-tor
  * click exit
  * verify the display is as follows:

goodtor

## Deployment

This is a user facing change only with no impact on deployment.

### DIFF
--- a/securedrop/tests/functional/source_navigation_steps.py
+++ b/securedrop/tests/functional/source_navigation_steps.py
@@ -161,5 +161,4 @@ class SourceNavigationSteps():
 
     def _source_logs_out(self):
         self.driver.find_element_by_id('logout').click()
-        notification = self.driver.find_element_by_css_selector('.important')
-        assert 'Thank you for exiting your session!' in notification.text
+        assert self.driver.find_element_by_css_selector('.important')
--- a/securedrop/source.py
+++ b/securedrop/source.py
@@ -399,7 +399,7 @@ def logout():
     if logged_in():
         session.clear()
         msg = render_template('logout_flashed_message.html')
-        flash(Markup(msg), "important")
+        flash(Markup(msg), "important hide-if-not-tor-browser")
     return redirect(url_for('index'))
 
 
--- a/securedrop/static/js/source.js
+++ b/securedrop/static/js/source.js
@@ -50,6 +50,7 @@ $(function(){
   } else {
     // If the user is not using Tor Browser, we want to encourage them to do so.
     $('.use-tor-browser').show();
+    $('.hide-if-not-tor-browser').hide();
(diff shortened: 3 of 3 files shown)


garrettr:
> The test environment does not have a tor browser and pretending firefox is a tor browser is non trivial.

@dachary Note that we do intend to switch to using Tor Browser for the functional tests fairly soon—see #1569 and the significant WIP in #1629.

dachary:
rebased & repushed

securedrop/tests/functional/source_navigation_steps.py
```diff
@@ -161,5 +161,4 @@ def _source_deletes_a_journalist_reply(self):
 
     def _source_logs_out(self):
         self.driver.find_element_by_id('logout').click()
-        notification = self.driver.find_element_by_css_selector('.important')
-        assert 'Thank you for exiting your session!' in notification.text
+        self.driver.find_element_by_css_selector('.important')
```
garrettr: This is an effective test (`find_element_by_css_selector` will raise `NoSuchElementException` if it cannot find any matching elements, which causes a test failure), but FWIW I generally prefer test assertions to be more explicit. I think that's a little harder to do in this case because this class doesn't inherit from `unittest.TestCase` or similar, and therefore doesn't provide as many useful assertion test methods.

dachary: I added an explicit ``assert`` to make clear this is what we are verifying, good point.

garrettr:
In the context of #1826, this PR only changes the web application code and the corresponding functional tests, so the CircleCI test coverage is not necessary for merge.

PR #1992, opened 2017-07-18
[0.4] Back out custom notification text on source interface

### LINKED ISSUE #1965: Back out CUSTOM_NOTIFICATION for 0.4
While QA'ing the 0.4 release, we realized that the `CUSTOM_NOTIFICATION` configuration option introduced in #1399 does not have a good deployment story for currently deployed instances because we generally do not have a good story for automated migrations of the SecureDrop web application configuration.

Currently, any changes that are made to `config.py.example` cannot be propagated to currently deployed instances because the [Ansible playbooks](https://github.com/freedomofpress/securedrop/blob/892cf52299f17ed9c70df0bb0bec68137101bab2/install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml#L29-L41) will avoid touching `config.py` if it has already been generated. Thus, *new* instances of 0.4 would be able to use the new `CUSTOM_NOTIFICATION` feature, but *existing* instances that are upgrading to 0.4 would not be able to take advantage of the feature. If the administrators of existing instances tried to modify the configuration on the Admin Workstation, those changes would simply not be propagated to the production machines.

We discussed this internally and agreed that the best course of action is to back out #1399 and any related changes for the 0.4 release, to avoid confusing and frustrating SecureDrop administrators with a new feature that would work inconsistently across old and new SecureDrop instances. We also agreed that it would be good to target the next release after 0.4, 0.4.1, for re-landing #1399, along with the configuration management changes required to support it consistently across current and new SecureDrop instances. I will file a separate follow-up issue to summarize the automated configuration upgrade issue and track work towards a resolution.

redshiftzero:
## Status

Ready for review

## Description of Changes

Closes #1965. Sadly, the deployment story for the new custom text on the source interface needs to be more carefully thought through for deployed instances of SecureDrop. In the interest of not blocking the upcoming release of 0.4, we're backing this out for now. 

## Testing

Check that all tests are passing and that there are no lingering `custom_notification` references anywhere. 

## Deployment

None


dachary:
Unless I'm mistaken, this will introduce a regression for installations where a custom notification is in use. The CUSTOM_NOTIFICATION will stay in the config.py file but no longer be rendered in the source index.

What about keeping this as an undocumented / deprecated feature instead ? When a sane implementation is in place, a migration path (manual or automated) will be available for installations that still use the old CUSTOM_NOTIFICATION.

redshiftzero:
Unfortunately, this `CUSTOM_NOTIFICATION` feature was only merged into `develop` and never made it into a release, so `CUSTOM_NOTIFICATION` won't be in the `config.py` of any deployed SecureDrop instances.

dachary:
Under the circumstances this is fortunate :-)

redshiftzero:
Thanks for review @garrettr 🤘 I decided to just do this backing out manually because the code in the diff crept in through a bunch of commits that I couldn't cleanly `git revert`

PR #1995, opened 2017-07-18
journalist: tests menu with id instead of text

dachary:
## Status

Ready for review

## Description of Changes

Looking up the menu text during testing is fragile, use the id instead.

## Testing

pytest tests

## Deployment

Test only no deployment issue.

### DIFF
--- a/securedrop/journalist_templates/base.html
+++ b/securedrop/journalist_templates/base.html
@@ -19,11 +19,11 @@
 
     {% if g.user %}
     <div id="logout">
-      <a href="{{ url_for('edit_account') }}">Edit Account</a> | 
+      <a href="{{ url_for('edit_account') }}" id="link_edit_account">Edit Account</a> |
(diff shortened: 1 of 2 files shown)


dachary:
> you should avoid adding stuff "just for tests" unless you really need to

Yes, this is good practice indeed. There is another reason to add **id** (although still in the context of tests), still in the context of i18n.

        admin_interface_link = self.driver.find_element_by_link_text('Admin')
        admin_interface_link.click()

does not work when the page is localized, 

        admin_interface_link = self.driver.find_element_by_id('link_admin_index')
        admin_interface_link.click()

does. The functions used on tests/functional to navigate to a given state and take a screenshot in a given language would not work without this change.

I thought about using a CSS selector instead (.something href ...) to avoid adding an id. But it seemed even sensitive because even the slightest change in the DOM could break it.

PR #1996, opened 2017-07-18
tests: functional: wait for app to respond

dachary:
## Status

Ready for review

## Description of Changes

There is a race between the navigator and the apps. Although rare, it is
possible the app are not responding yet when the first .get() is issued
to the browser. If that happens the test will just stop and fail instead
of waiting for all elements to load.

Poll the app URLs for 30 seconds before returning so there is no race
between the navigator and the apps.

## Testing

pytest tests/functional

## Deployment

this is for tests only, no side effect on deployment

### DIFF
--- a/securedrop/tests/functional/functional_test.py
+++ b/securedrop/tests/functional/functional_test.py
@@ -9,6 +9,7 @@ import signal
 import socket
 import time
 import traceback
+import requests
 
 from Crypto import Random
 from selenium import webdriver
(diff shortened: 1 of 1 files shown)


garrettr:
Since this is an external contribution, the CircleCI tests did not run because of #1826. Since the functional tests passed on Travis CI, I think it is fine to merge this without coverage from CircleCI.

dachary:
> I've never encountered this race condition before, but I believe that it could exist. 

I ran into it just a few times while taking screenshots for l10n in an automated way. They exercise this function often and have a better chance of racing. Since they take a screenshot in the end, it was obvious from the image that the service was not yet ready to respond :)

Thanks for the quick review !

garrettr:
@dachary Thanks for clarifying the specific context in which you encountered this error, that's really helpful!

PR #1983, opened 2017-07-17
[WIP] i18n

dachary:
## Status

Work in progress

## Description of Changes

i18n of source and journalist interface. This pull request is for reference and is not meant to be merged as is. Commits are proposed as separated pulll requests. Note that the [order of the commits as presented by github is not what you would get with git log](https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/). When viewed with git log the commits are grouped for clarity.

## Active pull requests waiting to be merged

These are the pull requests created from various commits in this branch.There should not be more than a few otherwise it will become difficult to manage.

* https://github.com/freedomofpress/securedrop/pull/2053
* https://github.com/freedomofpress/securedrop/pull/2056
* https://github.com/freedomofpress/securedrop/pull/2074

PR #1850, opened 2017-06-17
manage: clean-tmp removes old files to avoid races

dachary:
## Status

Ready for review

## Description of Changes

Fixes #1849

A file is considered old if it has not been modified in seven days (or
--days) and is removed from config.TEMP_DIR (or --directory).

The files in config.TEMP_DIR are for preparing downloads and they must
not be deleted before the user gets a chance to start the download. The
preparation of the file may be long for large collections but it will
always take less than a week.

It is not impossible for a the download to take more than a week if the
network is slow. However, once the download has begun it does not matter
if the file is removed: the server still has an open file descriptor and
keeps serving its content.

## Testing

The test_manage.py has been modified to test the new implementation.

## Deployment

1. Upgrading existing production instances.

The new implementation will leave temporary files for a week longer than the previous one.

2. New installs.

Nothing special.

PR #2004, opened 2017-07-20
manage: verbosity control

dachary:
## Status

Ready for review

## Description of Changes

Add logging and verbosity control (via the --verbose flag) to the manage.py subcommand. Also add pytest capture log which defines a convenient fixture for pytest to inspect logged messages.

## Testing

pytest -v pytest

## Deployment

The --verbose option is not used during deployment and has no side effects.

### DIFF
--- a/securedrop/manage.py
+++ b/securedrop/manage.py
@@ -3,6 +3,7 @@
 
 import argparse
 from getpass import getpass
+import logging
 import os
 import shutil
 import signal
@@ -17,6 +18,9 @@ os.environ['SECUREDROP_ENV'] = 'dev'  # noqa
 import config
 from db import db_session, init_db, Journalist
 from management.run import run
+
+logging.basicConfig(format='%(asctime)s %(levelname)s %(message)s')
+log = logging.getLogger(__name__)
 
 
 def reset():  # pragma: no cover
@@ -209,6 +213,7 @@ def clean_tmp():  # pragma: no cover
 def get_args():
     parser = argparse.ArgumentParser(prog=__file__, description='Management '
                                      'and testing utility for SecureDrop.')
+    parser.add_argument('-v', '--verbose', action='store_true')
     subps = parser.add_subparsers()
     # Run WSGI app
     run_subp = subps.add_parser('run', help='Run the Werkzeug source & '
@@ -247,10 +252,18 @@ def get_args():
     return parser
 
 
+def setup_verbosity(args):
+    if args.verbose:
+        logging.getLogger(__name__).setLevel(logging.DEBUG)
+    else:
+        logging.getLogger(__name__).setLevel(logging.INFO)
+
+
 def _run_from_commandline():  # pragma: no cover
     try:
         args = get_args().parse_args()
         rc = args.func()
+        setup_verbosity(args)
         sys.exit(rc)
     except KeyboardInterrupt:
         sys.exit(signal.SIGINT)
--- a/securedrop/tests/test_manage.py
+++ b/securedrop/tests/test_manage.py
@@ -9,10 +9,22 @@ import unittest
 import utils
 
 
-class TestManagePy(unittest.TestCase):
+class TestManagePy(object):
     def test_parse_args(self):
         # just test that the arg parser is stable
         manage.get_args()
+
+    def test_not_verbose(self, caplog):
+        args = manage.get_args().parse_args(['run'])
+        manage.setup_verbosity(args)
+        manage.log.debug('INVISIBLE')
+        assert 'INVISIBLE' not in caplog.text()
+
+    def test_verbose(self, caplog):
+        args = manage.get_args().parse_args(['--verbose', 'run'])
+        manage.setup_verbosity(args)
+        manage.log.debug('VISIBLE')
+        assert 'VISIBLE' in caplog.text()
 
 
 class TestManagementCommand(unittest.TestCase):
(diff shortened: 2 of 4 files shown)


dachary:
This pull request is in preparation of both [i18n](https://github.com/freedomofpress/securedrop/pull/1983) and a [manage clean_tmp bug fix](https://github.com/freedomofpress/securedrop/pull/1850). I also suspect it will be helpful in the context of Zack's work with a new manage sub-command to delete dangling submissions.

I originally made these as part of the [manage clean_tmp bug fix](https://github.com/freedomofpress/securedrop/pull/1850). But now that two pull requests depend on it, there is value in separating it to a standalone pull request. With this pull request merged we can make progress on i18n commits even if the bug fix to clean_tmp has not been resolved yet.

securedrop/manage.py
```diff
@@ -18,8 +19,11 @@
 from db import db_session, init_db, Journalist
 from management.run import run
 
+logging.basicConfig(format='%(asctime)s %(levelname)s %(message)s')
+log = logging.getLogger(__name__)
 
-def reset():  # pragma: no cover
+
+def reset(args):  # pragma: no cover
```
garrettr: Why did you add `args` to so many functions when it does not appear to be used in any of them?

securedrop/manage.py
```diff
@@ -247,10 +256,15 @@ def get_args():
     return parser
 
 
+def setup_verbosity(args):
+    logging.getLogger(__name__).setLevel(args.verbose)
```
garrettr: I don't like how you're overloading `args.verbose`, which, as a flag, is implied to be a boolean, with constants indicating the log level. I think it is better to decouple "what the user asked for" from "how the user's request should be implemented internally."

securedrop/manage.py
dachary: Because it is going to be used in both [i18n](https://github.com/freedomofpress/securedrop/pull/1983) and a [manage clean_tmp bug fix](https://github.com/freedomofpress/securedrop/pull/1850).

dachary:
I changed the handling of --verbose to be explicit as you suggested + added a test to fix coverage. Thanks for your review :-)

securedrop/manage.py
garrettr: Ok, but then it should be added by either of those PR's, not this one. I say this for two reasons:

1. In general, the changes in a PR should be tightly scoped and limited to only the changes necessary to address the corresponding issue. This makes reviewing, `git blame`ing, etc. much easier.
2. Adding `args` in this PR will create merge conflicts in your other PR's. I don't see why you wouldn't just wait to add it in one of the other PR's instead (where it is actually needed/used by other changes).

securedrop/manage.py
```diff
@@ -247,10 +252,18 @@ def get_args():
     return parser
 
 
+def setup_verbosity(args):
+    if args.verbose:
+        logging.getLogger(__name__).setLevel(logging.DEBUG)
+    else:
+        logging.getLogger(__name__).setLevel(logging.INFO)
```
garrettr: Thanks, this is exactly what I had in mind 👍

securedrop/tests/test_manage.py
```diff
@@ -14,6 +14,10 @@ def test_parse_args(self):
         # just test that the arg parser is stable
         manage.get_args()
 
+    def test_setup_verbose(self):
+        args = manage.get_args().parse_args(['run'])
+        manage.setup_verbosity(args)
```
garrettr: How is this a test? I don't see any assertions anywhere.

dachary: this is not a test, this is restoring coverage but I'll add a test as well

securedrop/manage.py
dachary: I created this PR to reduce the scope of the other PRs and gradually merge them. I did not do that originally (this PR was part of https://github.com/freedomofpress/securedrop/pull/1850) but after a few weeks with no review activity it seemed like a good idea to split it in this way. I'm happy to re-integrate it back to https://github.com/freedomofpress/securedrop/pull/1850 if you think this is the best course of action.

dachary: Another reason for splitting is that the two PRs that need this change are unrelated and it would not make much sense to have one depending on the other.

dachary:
with the tests added it looks more like a standalone pull request :-)

securedrop/manage.py
garrettr: @dachary I'm not suggesting you re-integrate them; I think separate PR's is definitely the way to go. However, I think that you should remove irrelevant changes from this PR, such as these new `args` parameters. Does that make sense?

securedrop/tests/test_manage.py
garrettr: There is no such thing as restoring coverage without adding tests. It is acceptable to explicitly mark code as not covered by tests (e.g. [`# pragma: no cover`](https://coverage.readthedocs.io/en/latest/excluding.html)), if you do not think it is necessary or worthwhile to test certain sections of code; however, it is not acceptable to trick the coverage analyzer into thinking new code is covered without actually testing the new code, which is what you did here.

dachary: It's better with asserts indeed. But it has value without assert. If the covered code has a typo the test will fail. Without the test a typo would go unnoticed. In any case I added the required tests.

dachary:
I removed the commit adding arguments.

garrettr:
In the context of #1826, this PR only affects `./manage.py` and the functional tests, so I think it is safe to merge without the CircleCI test coverage.

PR #2007, opened 2017-07-20
source: allow testing of tor2web

dachary:
## Status

Ready for review

## Description of Changes

Although it is possible to manually install an extension to the tor
browser to add a custom header and set X-tor2web in this way, it makes
the development environment more complicated. Besides, no other
source/journalist feature needs that kind of fixture and the effort does
not seem worth the reward.

Add a global variable TEST_X_TOR2WEB that allows tests to simulate the
presence of tor2web.

## Testing

pytest tests

## Deployment

This is only for tests, there are not deployment side effects.

### DIFF
--- a/securedrop/source.py
+++ b/securedrop/source.py
@@ -110,12 +110,15 @@ def setup_g():
         g.loc = store.path(g.sid)
 
 
+TEST_X_TOR2WEB = False
+
+
 @app.before_request
 @ignore_static
 def check_tor2web():
     # ignore_static here so we only flash a single message warning
     # about Tor2Web, corresponding to the initial page load.
-    if 'X-tor2web' in request.headers:
+    if 'X-tor2web' in request.headers or TEST_X_TOR2WEB:
         flash('<strong>WARNING:</strong> You appear to be using Tor2Web. '
               'This <strong>does not</strong> provide anonymity. '
               '<a href="/tor2web-warning">Why is this dangerous?</a>',


dachary:
I do not like this test only modification. The alternative makes the development environment significantly more complicated. My preference would be to install a firefox plugin that allows to set custom headers and pretend tor2web is active. I'd be happy to spend the time doing that if people  think it's worth the effort.

heartsucker:
> Add a global variable

This would go against some of the goals in 0.4.1 which would be to eliminate global state and use the `init_app()` Flask pattern.

Also, the better pattern would be to to just add it to the headers in the `GET` in the tests, no?

heartsucker:
Ah, this would be for the Selenium tests. Maybe this one could wait until after we make the apps more Flask-y?

dachary:
> Ah, this would be for the Selenium tests. 

Right, used [to generate screenshots](https://github.com/freedomofpress/securedrop/pull/1983/commits/9d64df5be71bc6615fd5f863b7bffc80d2a2271f#diff-848022a05ebb3b6032b0f390ac5bfc07R25) and verify it renders as it should, including with l10n.

Do you see a way to do the same without this ugly global variable ?

heartsucker:
Yeesh. I don't know enough about Selenium or the python driver, but I imagine it might be possible to have it set a custom header or user agent.

What you could do would be to set up a tiny flask app that takes a request and proxies it (with all headers / params / body) but adds a `X-Tor2Web` and `Forwarded` headers the way Tor2Web actually does. Then point the driver at this app for the Tor2Web stuff, and have the proxy point at the real app.

heartsucker:
Though once we have the the `init_app()` pattern, that could be injected into it as `init_app(..., force_tor2web=False)` and then use that. It wouldn't be global, but just on a per app basis. This would feel wrong in the sense that I think it's not always a good engineering practice to add things just to allow for testing if it makes the API / code too complex. Granted, there are other philosophies on this like the "Java dependency inject all things" approach. Different strokes, y'know.

dachary:
I think you have me convinced. We can just wait for the refactor and not test this specific message in the meantime, it's too much work for a transient patch. I like the proxy idea, simpler than installing a firefox plugin in the test environment...

Unless someone else has a brilliant idea in the next few days I'll close this pull request and create an issue to remember this needs work post-refactor.

dachary:
> Can you wait for a few weeks until we have time to work on that proposed refactor, or do you need this functionality urgently?

this is just for the visual validation of a single message. I think we can complete i18n without that :) Closing this for now and I'll create an issue as a reminder later. Thanks @garrettr @heartsucker for a quick review. I did not like the commit and to be honest I'm happy to kill it ;-)

PR #2015, opened 2017-07-21
[0.4] Full install documentation updates

redshiftzero:
## Status

Ready for review

## Description of Changes

Some minor doc updates I noticed while doing a 0.4 hardware install:
 * 794c735, baa3189: I ran through the firewall instructions and updated the last ancient screenshot and removed the recommendation to refresh the page during an upgrade if it's slow as the pfSense webapp now explicitly says not to do this
 * db1a388: state up front no passphrase necessary for SSH keys
 * 1d10c03, e1e2f26: Remove old comments from the SecureDrop 0.3 days, e.g. that we have 20 SecureDrops (also not gonna count them in docs anymore as we don't want to maintain that)
 * ff1cc05: fixing typo

Happy to break this into multiple PRs if necessary to aid in review

## Testing

Check everything reads fine and renders well locally.

## Deployment

Just docs


redshiftzero:
Adding 97cad3f because the AppArmor module isn't loaded on the Monitor server

conorsch:
New screenshot in network firewall docs looks great. You didn't by chance _also_ get an updated screenshot for `|Invoke auto upgrade|`, did you? Would fit well here if so.

redshiftzero:
OK the firewall update instructions were also out of date so I have rewritten this section and added new screenshots in 3ffb777.

conorsch:
New screenshots for the firewall upgrade workflow, hurray! :tada:

PR #2026, opened 2017-07-22
Fixes Application Server backup logic

### LINKED ISSUE #2019: Failure using Backup Ansible role
## Description

Backups fail using "Backup and restore" on 0.4

## Steps to Reproduce

1. Install 0.4 from test debs and upgrade admin workstation
2. `source .venv/bin/activate`
3. `cd install_files/ansible-base`
4. `ansible-playbook -t backup securedrop-prod.yml -e perform_backup=true`

## Expected Behavior

The backup completes successfully and a backup file `sd-backup*` should appear in the current working directory 

## Actual Behavior

Backup fails:

```
TASK [backup : fetch the backup file] ******************************************
fatal: [app]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'dict object' has no attribute 'stdout'\n\nThe error appears to have been in '/home/amnesia/Persistent/securedrop/install_files/ansible-base/roles/backup/tasks/backup.yml': line 17, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: fetch the backup file\n  ^ here\n"}
	to retry, use: --limit @/home/amnesia/Persistent/securedrop/install_files/ansible-base/securedrop-prod.retry
```

conorsch:
## Status

Ready for review.

## Description of Changes

Fixes #2019.

The `backup` role was using `async` for the creation of the backup tarball, but failed to wait on the results of that task before proceeding. Since the subsequent tasks operate on the backup tarball,
the role failed due to an undefined variable, specifically the `backup_filename` var.

The solution is to preserve the async nature of the task, since it can indeed take a while to complete, but make sure to poll for results, and pause execution of the role until the backup has completed.
For reference, see: http://docs.ansible.com/ansible/latest/async_status_module.html

## Testing

Try to perform a backup, following the backup guide in the docs. 

## Deployment

Backups are optional, and we don't have solid data on how frequently Admins actually use the functionality, but given that we're mandating Admin intervention for 0.4 in the form of Tails upgrades, it'd be good to resolve this prior to release.

## Checklist

Requires manual invocation of the backup logic in order to confirm resolution.

### DIFF
--- a/install_files/ansible-base/roles/backup/tasks/backup.yml
+++ b/install_files/ansible-base/roles/backup/tasks/backup.yml
@@ -1,27 +1,34 @@
 ---
-- name: copy the backup script to the target server
+- name: Copy the backup script to the target server.
   copy:
     src: backup.py
     dest: /tmp/
     owner: root
     mode: "0770"
 
-- name: run the backup script
+- name: Run the backup script (can take a while).
   # Setting args inline (against project style conventions) to work around
   # Ansible bug: https://github.com/ansible/ansible/issues/9693
   command: /tmp/backup.py chdir=/tmp
   # If there are a lot of submissions, this task could take a while.
   async: 3600
-  register: backup_filename
+  poll: 10
+  register: backup_script_result
 
-- name: fetch the backup file
+# The backup script emits the filename of the created tarball on stdout,
+# so we'll store it for use in subsequent tasks.
+- name: Store backup filename as host fact.
+  set_fact:
+    backup_filename: "{{ backup_script_result.stdout }}"
+
+- name: Fetch the backup tarball back to the Admin Workstation.
   fetch:
-    src: /tmp/{{ backup_filename.stdout }}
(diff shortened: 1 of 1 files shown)


redshiftzero:
Verified on a test instance that backup and restore now work successfully on 0.4 🎆

install_files/ansible-base/roles/backup/tasks/backup.yml
```diff
@@ -1,27 +1,34 @@
...
   copy:
     src: backup.py
     dest: /tmp/
     owner: root
     mode: "0770"
 
-- name: run the backup script
+- name: Run the backup script (can take a while).
   # Setting args inline (against project style conventions) to work around
   # Ansible bug: https://github.com/ansible/ansible/issues/9693
```
redshiftzero: Aside: looks like this bug is fixed now, but changing this is for another time

Issue #2031, opened 2017-07-23
Post offers of source for kernels


conorsch:
# Feature request

## Description

We should start posting explicit offers of kernel source packages for the custom-compiled kernels hosted in the FPF apt repository. We don't host source packages there, just the binary image packages (and also the header packages, even though those aren't used in the installation logic). For Admins who request source, we can provide those on demand, but we should also explicitly state this option.

The Software Freedom Conservancy has been in touch with recommendations about appropriate places to place the offer:

1. In the docs, in the install guide where grsecurity is installed, via a "Note" just below the https://docs.securedrop.org/en/latest/install.html#install-securedrop-servers section. 
2. As a comment in the FPF apt repo vars: https://github.com/freedomofpress/securedrop/blob/9f94a34e4f0c654d8e0d624342aecbd556c447f7/install_files/ansible-base/roles/install-fpf-repo/defaults/main.yml
3. Although not explicitly recommended by SFC, in the spirit of 2, also a comment in the grsecurity install tasks: https://github.com/freedomofpress/securedrop/blob/9f94a34e4f0c654d8e0d624342aecbd556c447f7/install_files/ansible-base/roles/grsecurity/tasks/from_fpf_repo_install_grsec.yml#L23
4. Configure a source package so that `apt-get source securedrop-grsec` displays the offer.

1-3 are all immediately achievable, and I support including them for the upcoming 0.4 release as docs updates. 4 will require a bit more research, but should be achievable in the near future (though not for 0.4). 

## Comments

In the interest of keeping the repo dry, perhaps we should add a top-level file to the repo that includes the explicit offer, and references (1-3) can point to that specific file. I'd at least like the top-level file to track the canonical text of the offer, and we can copy/paste it around as necessary to update it—although I suspect that will be a painful practice into the future.

## User Stories

As an Admin, I want a clear statement about where I can find source code for compiled packages running on my SecureDrop instance.

As a SecureDrop maintainer, I want a clear statement explaining the source request process, so that all Admins can be directed to the same information, and use the same contact methods for source requests.

conorsch:
We should also mention it in https://docs.securedrop.org/en/stable/development/apt_repo.html

redshiftzero:
Items 1-3 were completed in #2033, leaving just item 4 to complete. Moving to 0.4.1.

conorsch:
[...]

PR #2033, opened 2017-07-24
Adds offer for source for custom kernels

conorsch:
## Status

Ready for review.

## Description of Changes
Accomplishes most of the plan laid out in #2031. (Does not implement the `apt-get source <package>` option discussed there.)

Created a top-level file in the repo called `SOURCE_OFFER`, which includes a concise statement of offering source for the kernel packages, and links to itself via a public URL (in case someone is reading it offline). We can reuse the copy here in various places, and more importantly, we can link directly to this document from anywhere, giving us a single  point of update, and a single URL to sprinkle around in the  documentation. 

Updated the project documentation (both the install guide and the apt repo docs) with explicit mention of the source offer, and a link  directly to the SOURCE_OFFER file. Also added some comments to the Ansible YAML files in locations relevant to the apt repo and kernel install tasks.

## Testing

1. Render the docs locally and read through the changes.
2. Double-check the contact information is correct.
3. Observe the final "stable" URL https://github.com/freedomofpress/securedrop/blob/develop/SOURCE_OFFER and make sure it's appropriate, because I'd like to maintain it for a long time.
4. Observe the final "stable" URL and confirm it will be accurate when this PR is merged.

Note that the URL won't be live until we merge this PR, so don't be surprised at a 404.

## Deployment
None at all, this is documentation-only.

### DIFF
--- a/SOURCE_OFFER
+++ b/SOURCE_OFFER
@@ -1,0 +1,19 @@
+Freedom of the Press Foundation (FPF) maintains custom-compiled Linux
+kernel images for use in SecureDrop. The kernel source has been patched
+with the grsecurity modifications, and only binaries are available in the
+FPF apt repository.
+
+Under the terms of the GPL, any user with access to the binaries has a right
+to request the source code used to compile those binaries.
+You may submit a written request to:
+
+    Freedom of the Press Foundation
+    601 Van Ness Ave., Suite E371
+    San Francisco, CA 94102
+
+Or via email to:
+
+    source-offer@freedom.press
+
+A full copy of this message can be found online at
+https://github.com/freedomofpress/securedrop/blob/develop/SOURCE_OFFER
--- a/SOURCE_OFFER
+++ b/SOURCE_OFFER
@@ -3,7 +3,7 @@ kernel images for use in SecureDrop. The kernel source has been patched
 with the grsecurity modifications, and only binaries are available in the
 FPF apt repository.
 
-Under the terms of the GPL, any user with access to the binaries has a right
+Under the terms of the GPL, any user in possession of this offer has a right
 to request the source code used to compile those binaries.
 You may submit a written request to:
 
--- a/docs/development/apt_repo.rst
+++ b/docs/development/apt_repo.rst
@@ -39,5 +39,11 @@ securedrop-keyring
     Allows for managed key rotation via automatic updates, as implemented in
     `SecureDrop 0.3.10`_.
 
+.. note::
+   The SecureDrop install process configures a custom Linux kernel hardened
+   with the grsecurity patch set. Only binary images are hosted in the apt
+   repo. For source packages, see the `Source Offer`_.
+
 .. _SecureDrop 0.3.10: https://github.com/freedomofpress/securedrop/blob/c5b4220e04e3c81ad6f92d5e8a92798f07f0aca2/changelog.md#0310
 .. _apt.freedom.press: https://apt.freedom.press
+.. _`Source Offer`: https://github.com/freedomofpress/securedrop/blob/develop/SOURCE_OFFER
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -94,6 +94,13 @@ running the install, please submit a detailed `GitHub
 issue <https://github.com/freedomofpress/securedrop/issues/new>`__ or
 send an email to securedrop@freedom.press.
 
+.. note::
+   The SecureDrop install process configures a custom Linux kernel hardened
+   with the grsecurity patch set. Only binary images are hosted in the apt
+   repo. For source packages, see the `Source Offer`_.
+
+.. _`Source Offer`: https://github.com/freedomofpress/securedrop/blob/develop/SOURCE_OFFER
+
 Once the installation is complete, the addresses for each Tor Hidden
 Service will be available in the following files under
 ``install_files/ansible-base``:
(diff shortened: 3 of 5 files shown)


conorsch:
I briefly considered creating a URL that we fully control, e.g. `https://securedrop.org/source-offer`, since we could 301/302 to that over time to make sure it stays accurate. Still willing to do that, wanted feedback first. The current implementation doesn't involve redirects, which is nice, but will be harder to maintain.

dachary:
I'm reviewing the language carefully. It will take me an hour or so.

redshiftzero:
Excellent thanks @dachary! We will wait to merge this until after your review

dachary:
The answer (via email ?) to the written offer will be:

* a copy of the grsecurity patch required to re-create the same binary package
* a link to https://github.com/freedomofpress/ansible-role-grsecurity which contains the kernel sources on which to apply the patch as well as detailed instructions on how to re-create a binary kernel package functionally identical to the one published at apt.freedom.press

These two will constitute the **Complete and Corresponding Source Code**. Is my understanding correct ?

conorsch:
@dachary No, I was actually planning on handing over the linux-source deb package that we create as part of the build process, but do not publish. By way of example, the current kernel image deb package in the repo is:

```
linux-image-3.14.79-grsec_3.14.79-grsec-11.00.Custom_amd64.deb
```

Upon honoring a request for source, then, we could send over the source package:

```
linux-source-3.14.79-grsec_3.14.79-grsec-11.00.Custom_all.deb
```

Is that acceptable and compliant? The https://github.com/freedomofpress/ansible-role-grsecurity role repo doesn't actually house any source code for the kernel (or the grsecurity patch set)—it merely provides tooling and documentation. We'd originally thought publishing that tooling would help other projects adopt grsecurity, but with the patches under restricted access now, that no longer seems a realistic goal.

dachary:
can the source package be re-built by following instructions in https://github.com/freedomofpress/ansible-role-grsecurity ?

conorsch:
@dachary No, not unless we _also_ provide the grsecurity patch file. Willing to do that if you believe it's required for compliance. Even then, I'd like to make some quick changes to the https://github.com/freedomofpress/ansible-role-grsecurity logic such that one can provide a patch file from on-disk—the scripts currently assume one has basic auth access to the grsecurity project itself. But the manual guide https://github.com/freedomofpress/ansible-role-grsecurity/blob/a62731bf3c30ed6aab491e892693ce96ed501d87/docs/build.md will definitely work as long as one has the patch.

dachary:
Any third party should be able to verify the source you provide can indeed be used to create the binary. There are two ways to do that, if I understand correctly:

A) you publish the source package + instructions on how to use this binary package to rebuild a binary package
B) you publish the source package + the grsecurity patch + instructions on how to use https://github.com/freedomofpress/ansible-role-grsecurity  to rebuild a binary package

I think B) is better because it actually is the process you are also using to build the binary package. The option A) is a different way to rebuild the binary package and there is room for error. It would require you to document and verify another method of building the binary package which seems an unnecessary burden / risk.

Is that sensible ?

conorsch:
Option B makes sense, so I'm willing to update our internal documentation to include an explicit reference to the grsecurity patch file (and associated detached signature file) as part of honoring a source request. However, a point of clarification:

The source contained within the `linux-source` deb package has _already been patched_ with the grsecurity patch set. So by providing the source package, we already give someone the necessary information to configure and kick off a build—they do _not_ need to manually apply the grsecurity patch.

Given that information, perhaps A is the most realistic, in terms of empowering users to reproduce the package locally. It would require some small tweaks to the manual build docs (https://github.com/freedomofpress/ansible-role-grsecurity/issues/102), or maybe even its own section in the docs.

dachary:
> The source contained within the linux-source deb package has already been patched with the grsecurity patch set. So by providing the source package, we already give someone the necessary information to configure and kick off a build—they do not need to manually apply the grsecurity patch.

Right ! However going from the source package to the binary package needs to be documented. And as you very well know, it's not just a matter of running ``make-kpg`` ... there are dependencies, precautions and what not. All this context is an integral part of the instructions necessary for a third party to rebuild the binary. You already have all that precisely documented in https://github.com/freedomofpress/ansible-role-grsecurity/ and duplicating / verifying this process seems to be a source of error that is probably not necessary.

> It would require some small tweaks to the manual build docs (freedomofpress/ansible-role-grsecurity#102), or maybe even its own section in the docs.

I voiced my opinion but I'll defer to you better judgement on this, of course ;-) And I'm more than happy to run / test the process to verify it does work as expected.

SOURCE_OFFER
```diff
@@ -0,0 +1,19 @@
+Freedom of the Press Foundation (FPF) maintains custom-compiled Linux
+kernel images for use in SecureDrop. The kernel source has been patched
+with the grsecurity modifications, and only binaries are available in the
+FPF apt repository.
+
+Under the terms of the GPL, any user with access to the binaries has a right
```
dachary: s/with access to the binaries/in possession of this offer/

I'd be happy to explain the rationale, if you'd like :-)

conorsch: Revised! Thanks for the direct edit.

dachary:
:+1: !

redshiftzero:
Merging when checks pass. Thanks for review @dachary!

Issue #1882, opened 2017-06-26
HTTP only onion services produce warning in current TB


redshiftzero:
Firefox added [a warning](https://blog.mozilla.org/security/2017/01/20/communicating-the-dangers-of-non-secure-http/) to indicate when a user is interacting with a username or password input field on a page served over HTTP. In the current release of Tor Browser, 7.0.1, this is enabled. 

Unfortunately, this means that currently on the source interface (for the SecureDrop instances that don't use HTTPS on the source interface), a warning appears:

![screen shot 2017-06-26 at 2 01 48 pm](https://user-images.githubusercontent.com/7832803/27560440-054d7a9e-5a79-11e7-95c3-70a10e3e6891.png)

Similarly on the journalist interface:

![screen shot 2017-06-26 at 2 00 38 pm](https://user-images.githubusercontent.com/7832803/27560436-00bf40ac-5a79-11e7-8c5f-eb48ac9ba0ec.png)

This warning is intrusive (by design) and may scare away sources from submitting. The best solution here is to start using HTTPS for *both* the source and journalist interfaces (relevant: #1605 for the source interface). If there's another viable solution here (i.e. one that does not train users to click through security warnings), feel free to comment.

fowlslegs:
I'm going to file a bug upstream to see if Tor will remove this feature for `.onion`  sites.

fowlslegs:
Already being tracked in https://trac.torproject.org/projects/tor/ticket/21321 .

fowlslegs:
Looks like the Tor Trac ticket was given priority for resolution and labeled for July. Maybe we'll see it in one of the upcoming 7.0.X point releases in the next couple of months?

garrettr:
Looks like a PR was [posted](https://trac.torproject.org/projects/tor/ticket/21321#comment:46) in the upstream issue earlier today; fingers crossed that this will be fixed in an upstream release of Tor Browser soon!

garrettr:
I just discussed this issue with @redshiftzero. We agreed that it is challenging to address for a number of reasons.

We do not have a mechanism for receiving feedback from the anonymous sources who use SecureDrop, so it is difficult to gauge the potential impact of these warnings on end users; however, from our own recent encounters with the warnings from doing QA for SecureDrop 0.4, as well as reports from journalists who use SecureDrop, we have reason to believe these warnings are creating confusion and uncertainty for end users.

[...]

PR #2038, opened 2017-07-25
Explain insecure login warnings in Tor Browser

garrettr:
## Status

Ready for review.

## Description of Changes

Partial mitigation for #1882. Adds documentation for both sources and journalists to explain the erroneous "insecure login" warnings that they may encounter while using a recent version of Tor Browser on a SecureDrop instance that does not provide HTTPS.

My intention is to strike a reasonable balance between the different mitigation strategies described in https://github.com/freedomofpress/securedrop/issues/1882#issuecomment-317573646. Adding some documentation on this issue should help reassure end users who may be confused or frightened by these erroneous warnings.

I think it's best to add this explanation to the documentation, rather than directly to the affected pages in the SecureDrop web application, because:

* We're about to release SecureDrop 0.4, so any changes to the web application code at this point do not have a chance to be properly QA'ed and therefore are relatively *high risk*.
* Documentation changes are easier to review and less likely to cause significant breakage; therefore, they are relatively *low risk*.
* Adding language to the web application that directs users to ignore browser warnings is an anti-pattern, since it resembles malicious techniques used by phishing sites. I am loathe to instruct users to ignore browser warnings *at all*, but these warnings are very frightening and the upstream resolution is taking a long time, so I think it is acceptable in this case.

Eventually, once these erroneous warnings are resolved upstream in Tor Browser, we should be able to revert this PR.

-  [x] Doc linting passed locally


garrettr:
@redshiftzero FWIW, I could DRY this up a bit with an [`include`](http://docutils.sourceforge.net/docs/ref/rst/directives.html#include) directive, if you think that would be worthwhile.

PR #2053, opened 2017-08-01
i18n: extract/update/compile gettext from sources/templates

dachary:
## Status

Ready for review

## Description of Changes

manage.py translate --extract-update updates the sources and the result
is meant to be commited to the repository. It is used in development
environments.

manage.py translate --compile creates the .mo files that
are not to be commited to the repository. It is used for packaging
and in development environments.

## Testing

pytest -v tests/test_manage.py

## Deployment

It only adds to the existing manage.py command and is not actually used by any production deployment nor does it add files being deployed.

### DIFF
--- a/securedrop/manage.py
+++ b/securedrop/manage.py
@@ -10,6 +10,7 @@ import signal
 import subprocess
 import sys
 import traceback
+import version
 
 import psutil
 import qrcode
@@ -247,6 +248,56 @@ def clean_tmp(args):  # pragma: no cover
     return 0
 
 
+def translate(args):
+    messages_file = os.path.join(args.translations_dir, 'messages.pot')
+
+    if args.extract_update:
+        sh("""
+        set -xe
+
+        mkdir -p {translations_dir}
+
+        pybabel extract \
+        --charset=utf-8 \
+        --mapping={mapping} \
+        --output={messages_file} \
+        --project=SecureDrop \
+        --version={version} \
+        --msgid-bugs-address='securedrop@freedom.press' \
+        --copyright-holder='Freedom of the Press Foundation' \
+        {sources}
+
+        # we do not handle fuzzy translations yet
+        sed -i '/^#, fuzzy$/d' {messages_file}
+        """.format(translations_dir=args.translations_dir,
+                   mapping=args.mapping,
+                   messages_file=messages_file,
+                   version=args.version,
+                   sources=" ".join(args.source)))
+
+        if len(os.listdir(args.translations_dir)) > 1:
+            sh("""
+            set -xe
+            pybabel update \
+            --input-file {messages_file} \
+            --output-dir {translations_dir} \
+            --no-fuzzy-matching --ignore-obsolete
+            """.format(translations_dir=args.translations_dir,
+                       messages_file=messages_file))
+        else:
+            log.warning("no translations found (ok for tests, not otherwise)")
+
+    if args.compile:
+        sh("""
+        set -x
+
+        # stop here if there are no translations (for testing)
+        test $(ls {translations_dir} | wc -l) = 1 && exit 0
+
+        pybabel compile --directory {translations_dir}
+        """.format(translations_dir=args.translations_dir))
+
+
 def get_args():
     parser = argparse.ArgumentParser(prog=__file__, description='Management '
                                      'and testing utility for SecureDrop.')
@@ -286,7 +337,47 @@ def get_args():
     clean_tmp_subp_a = subps.add_parser('clean_tmp', help='^')
     clean_tmp_subp_a.set_defaults(func=clean_tmp)
 
+    set_translate_parser(subps)
+
     return parser
+
+
+def set_translate_parser(subps):
+    parser = subps.add_parser('translate',
+                              help='Update and compile translations')
+    translations_dir = 'translations'
+    parser.add_argument(
+        '--extract-update',
+        action='store_true',
+        help='run pybabel extract and pybabel update')
+    parser.add_argument(
+        '--compile',
+        action='store_true',
+        help='run pybabel compile')
+    mapping = 'babel.cfg'
+    parser.add_argument(
+        '--mapping',
+        default=mapping,
+        help='Mapping of files to consider (default {})'.format(
+            mapping))
+    parser.add_argument(
+        '--translations-dir',
+        default=translations_dir,
+        help='Base directory for translation files (default {})'.format(
+            translations_dir))
+    parser.add_argument(
+        '--version',
+        default=version.__version__,
+        help='SecureDrop version to store in pot files (default {})'.format(
+            version.__version__))
+    sources = ['.', 'source_templates', 'journalist_templates']
+    parser.add_argument(
+        '--source',
+        default=sources,
+        action='append',
+        help='Source file or directory to extract (default {})'.format(
+            sources))
+    parser.set_defaults(func=translate)
 
 
 def setup_verbosity(args):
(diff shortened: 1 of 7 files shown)


securedrop/tests/test_manage.py
```diff
@@ -52,3 +59,128 @@ def test_exception_handling_when_duplicate_username(self, mock_raw_input,
...
+        mo_file = os.path.join(locale_dir, 'LC_MESSAGES/messages.mo')
+
+        #
+        # Extract+update but do not compile
+        #
+        assert not os.path.exists(mo_file)
+        manage.translate(args)
+        assert not os.path.exists(mo_file)
+
+        #
```
redshiftzero: Can we split this into two separate tests? `test_translate_extract_and_update_without_compile` and `test_translate_compile_without_extract_and_update`?

securedrop/manage.py
```diff
@@ -210,6 +242,48 @@ def listdir_fullpath(d):
...
+        test $(ls {translations_dir} | wc -l) = 1 && exit 0
+
+        pybabel update \
+        --input-file {messages_file} \
+        --output-dir {translations_dir} \
+        --no-fuzzy-matching --ignore-obsolete
+        """.format(translations_dir=args.translations_dir,
+                   mapping=args.mapping,
+                   messages_file=messages_file,
+                   version=version.__version__,
```
redshiftzero: Perhaps we should add an option to either manually pass a `version` string or an abbreviated commit hash? The reason being, if we are preparing translations to be incorporated into a particular release, the version string here will not be updated and we may get confused about which version of the software the translation text files came from / are for.

securedrop/manage.py
```diff
@@ -210,6 +242,48 @@ def listdir_fullpath(d):
...
+        --charset=utf-8 \
+        --mapping={mapping} \
+        --output={messages_file} \
+        --project=SecureDrop \
+        --version={version} \
+        --msgid-bugs-address='securedrop@freedom.press' \
+        --copyright-holder='Freedom of the Press Foundation' \
+        {sources}
+
+        sed -i '/^#, fuzzy$/d' {messages_file}
```
redshiftzero: Add clarifying comment to this one

securedrop/manage.py
```diff
@@ -210,6 +242,48 @@ def listdir_fullpath(d):
...
+
+        test $(ls {translations_dir} | wc -l) = 1 && exit 0
```
redshiftzero: Add a clarifying comment here too, e.g. `# Continue if there's one file in translations directory as expected`

dachary: done

securedrop/manage.py
dachary: done

securedrop/manage.py
```diff
@@ -23,7 +25,37 @@
...
+        bufsize=1)
+    if stdin is not None:
+        proc.stdin.write(input)
+        proc.stdin.close()
+    lines = []
+    with proc.stdout:
+        for line in iter(proc.stdout.readline, b''):
+            line = line.decode('utf-8')
+            lines.append(line)
+            log.debug(line.strip().encode('ascii', 'ignore'))
```
dachary: because logging needs ascii so we need to convert lines that may contain non-ascii into something logging can accept. It could be an inconvenience if the non-ascii characters being ignored give a clue about an error. But I can't imagine any scenario where this actually is the case so I think we're good. It is possible I'm missing something though, do you have an idea in mind ?

securedrop/manage.py
```diff
@@ -23,7 +25,37 @@
 log = logging.getLogger(__name__)
 
 
-def reset():  # pragma: no cover
+def sh(command, input=None):
```
dachary: done (docstring)

securedrop/manage.py
```diff
@@ -23,7 +25,37 @@
...
+        args=command,
+        stdin=stdin,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        shell=True,
+        bufsize=1)
+    if stdin is not None:
+        proc.stdin.write(input)
+        proc.stdin.close()
+    lines = []
```
dachary: done

securedrop/manage.py
dachary: Ah, right. Completely missed that, good catch ! Added --version with a default to version.__version__

securedrop/tests/test_manage.py
dachary: Hum. I re-read the test and added a few asserts while I was at it. Could you explain the rationale for splitting this into two tests ? Unless I'm mistaken ``test_translate_compile_without_extract_and_update`` needs to do all ``test_translate_extract_and_update_without_compile`` does to prepare what it needs. Maybe it's just me missing a factorization and I'd be grateful if you could point me in the right direction.

dachary:
modified and repushed, thanks for the quick review @redshiftzero :-) The draft documentation 74b1887 is a bit ahead of its time because it documents a usage that assumes code/templates/javascript already is i18n. What do  you think ?

securedrop/manage.py
```diff
@@ -210,6 +248,51 @@ def listdir_fullpath(d):
...
+        --version={version} \
+        --msgid-bugs-address='securedrop@freedom.press' \
+        --copyright-holder='Freedom of the Press Foundation' \
+        {sources}
+
+        # we do not handle fuzzy translations yet
+        sed -i '/^#, fuzzy$/d' {messages_file}
+
+        # stop here if there are no translations (for testing)
+        test $(ls {translations_dir} | wc -l) = 1 && exit 0
```
redshiftzero: For these test lines can we either: do this in Python e.g. with `len(os.listdir(translations_dir)` and print a message if we exit prematurely on this line i.e. "Exited - No translations found!" or similar to do the same in bash, e.g. add a `|| print "Exited - No translations found!"`.

dachary: I'll rework that in python, sure :-)

dachary:
> I think some of the bash may a bit nicer to do directly with Python

When facing a library that has two APIs (cli & python classes) I think best to choose the one that is likely to be supported best. I think that in the context of manage.py the cli is the better choice because it is likely to be used by more people than the API and therefore more stable, and better maintained and documented. Of course I don't have any hard evidence, it's a gut feeling :-)

If you think the better choice is to use the API I'll change to that, just let me know.

dachary:
> Also re the documentation, thanks for writing it, I enjoyed reading it 😎 - but I take your point. How about we put that commit into a separate PR and we can review it from there? 

Great, I'll do that. My grammar definitely needs fixing ;-)

dachary:
rebased (now that pybabel is back in develop, the test should pass). I think it is ready for review now. Unless I'm mistaken the only remaining question is wether the python API should be used for pybabel instead of the cli API, to make it pure python instead.

redshiftzero:
Two changes:
 * Remove commit 579adca9219e00194c6494648b38ec82ac6994c9 as those changes are in the separate PR #2074
 * Address comment https://github.com/freedomofpress/securedrop/pull/2053/files#r131223172. 

Other than the one part of the code I commented on above, since this is an infrequently run developer only tool, I think it's fine to leave as is without making it pure python :+1:.

dachary:
I was sure this was done... but on the laptop I had at sha2017. And the push I thought I made must have failed without me noticing. Sorry for the confusion, I'll repush.

dachary:
Recovered the lost changes & repushed.

dachary:
rebased & repushed

dachary:
continued at https://github.com/freedomofpress/securedrop/pull/2120

PR #1924, opened 2017-07-01
i18n: add Flask-Babel requirement

dachary:
## Status

Ready for review

## Description of Changes

add Flask-Babel requirement

## Testing

vagrant up development # will install from securedrop-requirements.txt

## Deployment

1. Upgrading existing production instances.
2. New installs.

It will install new python modules that will not be used and will not interfere with existing or new deployment.

### If you made changes to the app code:

- [x] Unit and functional tests pass on the development VM

PR #2071, opened 2017-08-04
Revert "i18n: add Flask-Babel requirement"

### LINKED ISSUE #2070: Source Interface throwing a 500 in staging on develop
# Bug

## Description

The Source Interface is failing due to an AppArmor violation for apache:

>Aug  3 23:20:20 app-staging kernel: [  656.377951] audit: type=1400 audit(1501802420.976:32): apparmor="DENIED" operation="open" profile="/usr/sbin/apache2" name="/etc/timezone" pid=10855 comm="apache2" requested_mask="r" denied_mask="r" fsuid=33 ouid=0

## Steps to Reproduce

1. Check out `develop` branch.
2. Bring up staging VMs.
3. Try to load the Source Interface (you'll see a 500).
4. Check syslog and you'll see an error similar to the above.

## Expected Behavior

Source Interface renders well because apache is happy and able to do its work.

## Actual Behavior

Source Interface fails because apache is confined by the restrictive AppArmor profile and falls down before it can serve anything.

## Comments

The regression likely slipped in due to lack of proper CI on PRs coming from forks (#1826). Thankfully the the functional test introduced in #1731 caught the error, when none of the app test suite did.

conorsch:
## Status

Ready for review.

## Description of Changes

Change came in via #1924 but tests were not run (see #1826), so a
breaking regression due to the AppArmor profiles has been wrecking
builds in CI. We need to reland the requirements updates as part of the
i18n effort, but as a short-term fix, we're backing out the breaking
change from develop, pending a more durable and fully-tested
implementation.

Fixes #2070.

This reverts commit 10b41800d5524aef6c483c5c77d37f279e4fcecd.

## Testing

Confirm CircleCI passes. Bonus points if you spin up a staging environment and confirm a 200 GET on both Source and Journalist Interfaces.

## Deployment

No concerns for deployment, this bug only landed in develop and has never shipped.

## Checklist

Trust nothing but CircleCI.

### DIFF
--- a/securedrop/requirements/securedrop-requirements.in
+++ b/securedrop/requirements/securedrop-requirements.in
@@ -1,6 +1,5 @@
 cssmin
 Flask-Assets
-Flask-Babel
 Flask-WTF
 Flask
 gnupg
--- a/securedrop/requirements/securedrop-requirements.txt
+++ b/securedrop/requirements/securedrop-requirements.txt
@@ -4,11 +4,9 @@
 #
 #    pip-compile --output-file securedrop-requirements.txt securedrop-requirements.in
 #
-babel==2.4.0              # via flask-babel
 click==6.7                # via flask, rq
 cssmin==0.2.0
 flask-assets==0.12
-flask-babel==0.11.2
 flask-wtf==0.14.2
 flask==0.12.2             # via flask-assets, flask-wtf
 gnupg==2.3.0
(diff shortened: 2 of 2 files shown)


dachary:
Wow. Thanks for fixing this, I'll work on it and run CircleCI myself (or something that does the same).

PR #2078, opened 2017-08-04
Fix AppArmor rule denying access to fontawesome-webfont.svg

### LINKED ISSUE #2075: Apparmor denials for fontawesome-webfont.svg trigger OSSEC alerts
Don't think these are of great concern and have only occurred once but wanted to flag. See https://github.com/freedomofpress/securedrop/issues/1240 different file, same issue ;-)

Securedrop 0.4.1

# Bug

## Description

Received From: APP->/var/log/syslog
Portion of the log(s):

DATE app kernel: [19517572.494728] audit: type=1400 audit(1501844691.909:75): apparmor="DENIED" operation="open" profile="/usr/sbin/apache2" name="/var/www/securedrop/static/fonts/fontawesome-webfont.svg" pid=3891 comm="apache2" requested_mask="r" denied_mask="r" fsuid=33 ouid=33

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #2075.

Apache should have access to fontawesome-webfont.svg. If AppArmor
denies access to this file, a spurious OSSEC alert will be produced.

## Testing

Verify the path to the FontAwesome file is correct

## Deployment

There should be no issues deploying this to new or existing installs 

## Checklist

Relying on manual review and CI here

### DIFF
--- a/install_files/ansible-base/usr.sbin.apache2
+++ b/install_files/ansible-base/usr.sbin.apache2
@@ -210,6 +210,7 @@
   /var/www/securedrop/static/js/libs/jquery-2.1.4.min.js r,
   /var/www/securedrop/static/js/source.js r,
   /var/www/securedrop/static/fonts/fontawesome-webfont.eot r,
+  /var/www/securedrop/static/fonts/fontawesome-webfont.svg r,
   /var/www/securedrop/static/fonts/fontawesome-webfont.ttf r,
   /var/www/securedrop/static/fonts/fontawesome-webfont.woff r,
   /var/www/securedrop/static/fonts/fontawesome-webfont.woff2 r,


dachary:
Under which circumstances is fontawesome-webfont.svg used ?

conorsch:
> Under which circumstances is fontawesome-webfont.svg used ?

I too am having a heck of a time trying to reproduce the problem described in #2075. Tried both the Source and the Journalist Interfaces, navigating to all views I could find. It's worth noting that in Tor Browser, with the Security Slider set to "High," SVGs will not be loaded. (I assume that's how we missed this during QA.) Still, even with it set to "Low," I'm not able to trigger this condition on any page in the UI.

conorsch:
Spoke with @redshiftzero briefly today about inability to reproduce this. While I don't have strong objections to merge given the the diff (we ship the file, so we should indeed whitelist it in the profile), I want clear documentation of where it's loaded and why, since without that future maintainers will have a heck of a time reconstructing some of these changes.

redshiftzero:
I too was unable to reproduce this AppArmor failure by typical user interaction with the source and journalist interface on Tor Browser. Further, SVG fonts are [not supported](https://bugzilla.mozilla.org/show_bug.cgi?id=119490) by Firefox in favor of woff2 (which we do whitelist):

screen shot 2017-08-16 at 8 27 14 am

The SVG font is referred to only in `_font_awesome.sass` (a dependency). This alert would be triggered if someone directly downloads `http://<source or journalist URL>.onion/static/fonts/fontawesome-webfont.svg`, which might happen if someone is inspecting the resources in `journalist.css` and tries to download them all for some reason. 

I could go either way on this - either removing the SVG file from `_font_awesome.sass` or keeping it and adding the AppArmor rule. I err on the side of whitelisting the file however because `_font_awesome.sass` is a dependency, so we'd need to maintain whatever changes we make to it if/when we bump Font Awesome versions.

conorsch:
Slick report, @redshiftzero, thanks for documenting clearly!

PR #2081, opened 2017-08-05
Add HTML linting to CI and fix linting errors on all HTML templates

### LINKED ISSUE #1897: Lint HTML
To catch issues like the [missing](https://github.com/freedomofpress/securedrop/pull/1829/commits/05e5ddb05d96f2a107410d1c6e5fa63dcf286d72#r122849978) `</a>` tag found in #1829, we should find a good linter for HTML (e.g. [html-linter](https://pypi.python.org/pypi/html-linter/)), and add it to CI.

redshiftzero:
## Status

Ready for review

## Description of Changes

This PR closes #1897

* Adds [html-linter](https://pypi.python.org/pypi/html-linter/) to Travis CI
* Cleans up all our HTML templates so now we pass all linting checks (and thus follow the [Google HTML style guide](https://google.github.io/styleguide/htmlcssguide.html))

Linting errors will result in failed CI builds (019724b) and no linting errors result in passing builds (e5a2762).

### Disabled checks

Note that I disabled three checks. Here I'm going to briefly describe why for posterity. 

####  `optional_tag` 

This is an informational only linting warning about removing unnecessary closing tags. See [here](https://google.github.io/styleguide/htmlcssguide.html#Optional_Tags) for more explanation. I turned this off as it is optional and I personally prefer the more explicit style (for example, this linting check recommends the removal of `</p>` tags). If consensus is that this should be turned on, we can do so, but I recommend in a followup PR, as it would add _significantly_ to the diff.

####  `extra_whitespace`

This check gets confused by the fact we are linting Jinja2 templates, e.g. it reports an error on this line in `securedrop/source_templates/base.html`: 

```
<a href="{% if 'logged_in' in session %}{{ url_for('lookup') }}{% else %}{{ url_for('index') }}{% endif %}">
```

#### `indentation`

This would make some of our Jinja2 templates _less_ readable, e.g. when we have a code block like this:

```
{% if messages %}
  {% for category, message in messages %}
    {% if category != "banner-warning" %}
      <div class="flash {{ category }}">
        {% if category == "notification" %}
        <img src="{{ url_for('static', filename='i/font-awesome/info-circle-black.png') }}" height="16" width="20">
        {% elif category == "success" %}
        <img src="{{ url_for('static', filename='i/success_checkmark.png') }}" height="17" width="20">
        {% elif category == "error" %}
        <img src="{{ url_for('static', filename='i/font-awesome/exclamation-triangle-black.png') }}" height="17" width="20">
        {% endif %}
        {{ message }}
      </div>
    {% endif %}
  {% endfor %}
{% endif %}
```

it will complain about the nesting here. 

Most of the changes in this PR were pretty straightforward, but I know this is a big diff, so let me know if it needs to get split up into multiple PRs. Happy to do so to aid review :smile:

## Testing

Verify that all tests pass and inspect the diff. 

## Deployment

Just changes some of the HTML and functional tests, should introduce no problems in deployment.

### DIFF
--- a/securedrop/requirements/develop-requirements.txt
+++ b/securedrop/requirements/develop-requirements.txt
@@ -11,18 +11,21 @@ argh==0.26.2              # via sphinx-autobuild, watchdog
 asn1crypto==0.22.0        # via cryptography
 babel==2.4.0              # via sphinx
 backports-abc==0.5        # via tornado
-boto==2.47.0
-certifi==2017.4.17        # via requests, tornado
-cffi==1.10.0              # via cryptography
+bcrypt==3.1.3             # via paramiko
+boto==2.48.0
+certifi==2017.7.27.1      # via requests, tornado
+cffi==1.10.0              # via bcrypt, cryptography, pynacl
 chardet==3.0.4            # via requests
 click==6.7                # via pip-tools
 configparser==3.5.0       # via flake8
-cryptography==1.9         # via paramiko
-docutils==0.13.1          # via sphinx
+cryptography==2.0.3       # via paramiko
+docopt==0.6.2             # via html-linter, template-remover
+docutils==0.14            # via sphinx
 enum34==1.1.6             # via cryptography, flake8
 execnet==1.4.1            # via pytest-xdist
 first==2.0.1              # via pip-tools
-flake8==3.3.0
+flake8==3.4.1
+html-linter==0.4.0
 idna==2.5                 # via cryptography, requests
 imagesize==0.7.1          # via sphinx
 ipaddress==1.0.18         # via cryptography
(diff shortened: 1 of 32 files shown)


securedrop/requirements/develop-requirements.txt
```diff
@@ -11,50 +11,55 @@ argh==0.26.2              # via sphinx-autobuild, watchdog
 asn1crypto==0.22.0        # via cryptography
 babel==2.4.0              # via sphinx
 backports-abc==0.5        # via tornado
-boto==2.47.0
-certifi==2017.4.17        # via requests, tornado
-cffi==1.10.0              # via cryptography
+bcrypt==3.1.3             # via paramiko
+boto==2.48.0
```
dachary: could you maybe add one line in the commit comment explaining why the dependencies are updated as well ?

dachary:
Unless I'm mistaken html-linter will not catch a missing closing tags. I tried feeding it a file with just
<pre>
&lt;a>&lt</b>
</pre>
and it did not complain. Maybe I misunderstand the scope of this PR and a tool like [tidy](http://www.html-tidy.org/) is already used somewhere else to ensure the HTML is valid ? Reading [html-linter](https://pypi.python.org/pypi/html-linter/0.1.9) documentation 
<pre>
What is missing?

...

The abbility to validate the HTML using the tool HTML5 tidy and to integrate some schema.org or microdata validator.
</pre>
So it looks like it indeed assumes a validator ran before it.

redshiftzero:
Thanks for review @conorsch. Yep @dachary's point is very valid and after a tiny bit of research [html5validator](https://github.com/svenkreiss/html5validator) seems like it might do the job. I vote we merge this but reopen #1897 and I (or someone else if they get to it first) can do more experimenting and potentially add that HTML validator

conorsch:
> I vote we merge this but reopen #1897

:+1: !

Issue #2035, opened 2017-07-24
./manage.py add-admin prompts admins to confirm an invalid password


garrettr:
# Bug

## Description

The minimum password length check added to `./manage.py add-admin` in #1660 is not performed until after the password is entered twice, i.e. after both the `Password:` and `Confirm Password:` prompts.

This is slightly annoying for administrators because it makes them enter an invalid password twice before letting them know that it was invalid.

## Steps to Reproduce

1. Provision a development VM and ssh into it.
2. `cd /vagrant/securedrop`
3. `./manage.py add-admin`
4. Enter any username.
5. At the `Password:` prompt, enter a password that is less than 12 characters.
6. At the `Confirm Password:` prompt, enter the same password again.

## Expected Behavior

After a short password is entered at the `Password:` prompt, `./manage.py add-admin` should perform the password length check and let the user know that their desired password is too short.

## Actual Behavior

The user is prompted to confirm their invalid password before being told it is invalid. This is confusing and annoying, because it forces the admin to take a step back after being lead to think they had an entered an acceptable password.

## Comments

We should move the password length check so it is performed after the first `Password:` prompt.

redshiftzero:
Thanks for filing this @garrettr. We should definitely fix this, tagging for the next minor release - 0.4.1.

heartsucker:
Should we implement https://github.com/freedomofpress/securedrop/issues/980 (diceware passwords), this would be covered.

redshiftzero:
Adding blocked because as @heartsucker notes, this will be naturally solved by #1509.

redshiftzero:
Fixed by #1509

Issue #1982, opened 2017-07-16
Write unit tests for manage commands


zsck:
# Feature request

## Description

The functions in [manage.py](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/manage.py) are almost all marked with the `# pragma: no cover` to avoid having test coverage expect tests for them. There doesn't seem to be any particular reason why these shouldn't be tested.

redshiftzero:
Great point @zsck. Tagging this one for 0.4.1. I also wonder if some of the functionality in `manage.py` could be DRYed up (e.g. both `journalist.py` and `manage.py` adds users).

dachary:
https://github.com/freedomofpress/securedrop/pull/1850/files fixes a race and adds tests for one function. One step toward that goal ;-)

PR #2085, opened 2017-08-06
added min-length check for journalist usernames

### LINKED ISSUE #1682: There is no minimum username requirement for the JI
We specify `username = Column(String(255), nullable=False, unique=True)` among the `db.Journalist` class attributes, but `nullable=False` doesn't mean that `''` is an unacceptable string.

I think a nice way to do it would be to rename the `username` column as `_username`, and then a new `username` property whose `getter` returns `_username`, and whose `setter` will raise a new `InvalidUsernameException` if the value passed is the empty string. Since this would require a database migration, but is also not a pressing problem, it makes sense to wait for that to fix this.

An alternative is simply to put such a check into place everywhere username may be set.

heartsucker:
## Status

Ready for review

## Description of Changes

FIxes #1682

## Testing

Try to use `manage.py` or the admin interface to make a username with less than 8 characters. Expect an error.

## Deployment

This does not affect existing installations.


dachary:
<pre>
tests/test_journalist.py:540:80: E501 line too long (91 > 79 characters)
</pre>

securedrop/db.py
```diff
@@ -251,10 +251,17 @@ class Journalist(Base):
         "JournalistLoginAttempt",
         backref="journalist")
 
+    MIN_USERNAME_LEN = 8
+
     def __init__(self, username, password, is_admin=False, otp_secret=None):
+        if len(username) < self.MIN_USERNAME_LEN:
+            raise InvalidUsernameException(
+                    'Username "{}" is too short'.format(username))
```
dachary: it would help l10n to have 'Username "{username}" must be at least {min} character long".format(username=username, min=self.MIN_USERNAME_LEN)

dachary:
Not sure why the coverage is lower now ... it would be good to always have 

<pre>
modified   securedrop/tests/pytest.ini
@@ -1,4 +1,4 @@
 [pytest]
 testpaths = . functional
 usefixtures = setUptearDown
-addopts = --cov=../securedrop/
+addopts = --cov-report term-missing --cov=../securedrop/
</pre>

so we know where to look, don't you think ?

heartsucker:
@dachary I had to pop one function out to mock it so it shows up as uncovered. I added a `pragma: no cover` to it to catch that single line decrease.

dachary:
> out to mock it so it shows up as uncovered

Ah right ! To protect against typos would you mind adding a test to cover this one function ? I feel uncomfortable with "no cover" when it's not too complicated to do otherwise.

heartsucker:
I feel like that might be overkill. Someone from FPF want to comment on that?

heartsucker:
Ping.

securedrop/db.py
```diff
@@ -251,10 +251,18 @@ class Journalist(Base):
         "JournalistLoginAttempt",
         backref="journalist")
 
+    MIN_USERNAME_LEN = 8
+
     def __init__(self, username, password, is_admin=False, otp_secret=None):
+        if len(username) < self.MIN_USERNAME_LEN:
+            raise InvalidUsernameException(
+                        'Username "{}" must be {} characters long.'
+                        .format(username, Journalist.MIN_PASSWORD_LEN))
```
redshiftzero: This line should be: `.format(username, Journalist.MIN_USERNAME_LEN))`

securedrop/manage.py
```diff
@@ -66,9 +66,13 @@ def add_journalist():  # pragma: no cover
...
 
 
+def get_username():  # pragma: no cover
+    return raw_input('Username: ')
+
+
 def _add_user(is_admin=False):  # pragma: no cover
     while True:
-        username = raw_input('Username: ')
+        username = get_username()
```
redshiftzero: The user gets an ugly traceback when they enter an invalid username:

```
vagrant@development:/vagrant/securedrop$ ./manage.py add-admin
Username: jen
Password: 
Confirm Password: 
Will this user be using a YubiKey [HOTP]? (y/N): 
['Traceback (most recent call last):\n', '  File "./manage.py", line 108, in _add_user\n    otp_secret=otp_secret)\n', '  File "<string>", line 4, in __init__\n', '  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/orm/state.py", line 414, in _initialize_instance\n    manager.dispatch.init_failure(self, args, kwargs)\n', '  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/util/langhelpers.py", line 66, in __exit__\n    compat.reraise(exc_type, exc_value, exc_tb)\n', '  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/orm/state.py", line 411, in _initialize_instance\n    return manager.original_init(*mixed[1:], **kwargs)\n', '  File "/vagrant/securedrop/db.py", line 260, in __init__\n    .format(username, Journalist.MIN_PASSWORD_LEN))\n', 'InvalidUsernameException: Username "jen" must be 12 characters long.\n']
```

Two suggestions:

1. If the username is shorter than `Journalist.MIN_USERNAME_LEN`, then let's print a message like, e.g. "Username "jen" must be 8 characters long." 

2. The if block checking the username length should appear _after_ the username is input and _before_ the passwords are input. It would be annoying for a user to have to type out their desired password twice only to find their username is invalid (UX issue similar to #2035).

securedrop/tests/test_journalist.py
```diff
@@ -530,7 +532,17 @@ def test_admin_add_user_without_username(self):
                                           password='pentagonpapers',
                                           password_again='pentagonpapers',
                                           is_admin=False))
-        self.assertIn('Missing username', resp.data)
+        self.assertIn('Invalid username', resp.data)
+
+    def test_admin_add_user_too_short_username(self):
```
redshiftzero: Nice test :100:

securedrop/manage.py
```diff
@@ -66,9 +66,13 @@ def add_journalist():  # pragma: no cover
     return _add_user()
 
 
+def get_username():  # pragma: no cover
```
redshiftzero: I agree with @dachary here - let's either add a test (better) or remove the no cover. We should be removing the no covers from most of the functions in this file (see #1982).

securedrop/manage.py
redshiftzero: Also, since we're further complicating this one function (`_add_user`), this would also be a perfect opportunity to add tests for this one function (partially addresses #1982).

securedrop/db.py
heartsucker: Done.

redshiftzero:
Hey @heartsucker - is this one ready for re-review?

heartsucker:
Yarp.

securedrop/db.py
```diff
@@ -264,10 +264,18 @@ class Journalist(Base):
         "JournalistLoginAttempt",
         backref="journalist")
 
+    MIN_USERNAME_LEN = 8
```
redshiftzero: This is a very long minimum username length. What about 3 characters? That seems more reasonable.

securedrop/journalist.py
```diff
@@ -188,6 +184,9 @@ def admin_add_user():
                 flash('There was an error with the autogenerated password. '
                       'User not created. Please try again.', 'error')
                 form_valid = False
+            except InvalidUsernameException as e:
+                form_valid = False
+                flash('Invalid username: ' + str(e), "error")
```
redshiftzero: We should flash the same informative message we do on the command line: "Username blah is too short. Minimum characters: `Journalist.MIN_USERNAME_LEN`"

redshiftzero: In `admin_edit_user` we will also need an additional `elif` block to check if `new_username` is less than `Journalist.MIN_USERNAME_LEN`  - currently admins can set user passwords back to usernames that are too short.

securedrop/db.py
heartsucker: For the public record: the offline conversation says that there is no attack that is mitigated by longer usernames. Lowering to 3.

securedrop/journalist.py
heartsucker: `str(e)` will include the message saying that the username is too short and how long it needs to be. The username could be invalid for other reasons in the future (white space, uncode, who knows), so this is future proof.

You are right about the `admin_edit_user` block.

heartsucker: Done.

heartsucker:
@redshiftzero Ready for another round of review.

redshiftzero:
Excellent, will re-review shortly!

securedrop/tests/test_manage.py
```diff
@@ -43,9 +44,14 @@ def setUp(self):
     def tearDown(self):
         utils.env.teardown()
 
-    @mock.patch("__builtin__.raw_input", return_value='test')
-    def test_get_username(self, mock_stdin):
-        assert manage._get_username() == 'test'
+    @mock.patch("__builtin__.raw_input", return_value='jen')
+    def test_get_username_success(self, mock_stdin):
+        assert manage._get_username() == 'jen'
```
redshiftzero: :eyes:

securedrop/journalist.py
```diff
@@ -298,6 +297,13 @@ def admin_edit_user(user_id):
     if request.method == 'POST':
         if request.form['username']:
             new_username = request.form['username']
+
+            try:
+                Journalist.check_username_acceptable(new_username)
+            except InvalidUsernameException as e:
+                flash('Invalid username: ' + str(e))
```
redshiftzero: Should be: 

`flash('Invalid username: ' + str(e), 'error')`

Otherwise the flashed message style is inconsistent: 

screen shot 2017-08-24 at 11 10 23 am

heartsucker:
@redshiftzero Rebased and passing tests.

heartsucker:
@redshiftzero Rebased and passing tests.

PR #2046, opened 2017-07-26
docs: translator guide

dachary:
## Status

Ready for review

## Description of Changes

Fixes #2039

A guide for translators

### If you made changes to documentation:

- [x] Doc linting passed locally

PR #2088, opened 2017-08-07
Run full test suite for #2046

redshiftzero:
CI for #2046 (see #1826 for explanation of why this requires a PR)


redshiftzero:
#2046 was merged

PR #2101, opened 2017-08-11
[CI] Full test coverage for #2053

redshiftzero:
Full test coverage for #2053 - Do not merge


redshiftzero:
This branch was pushed to the shared remote in #2120 and a PR opened from there. This means that all tests will automatically run. Closing!

PR #2123, opened 2017-08-16
tests: join threads before teardown

### LINKED ISSUE #2122: transient test error Directory not empty: '/tmp/securedrop'
# Bug

## Description

Because source.py **async_keygen** races with every test using it implicitly, it creates various transient errors, a common one being [Directory not empty: '/tmp/securedrop'](https://travis-ci.org/freedomofpress/securedrop/builds/264935607?utm_source=github_status&utm_medium=notification)

## Steps to Reproduce

while : ; do pytest -v tests/test_source.py || break ; done

## Expected Behavior

the loop never ends

## Actual Behavior

the loop will break after a while, most likely with a **Directory not empty: '/tmp/securedrop'** error

## Comments

join the thread before teardown

dachary:
## Status

Ready for review

## Description of Changes

Fixes #2122 

If the tests left live threads, join them before proceeding with
teardown. The most common live thread is from the
source.py::async_keygen function. If not joined, the thread will
terminate after the teardown and is likely to fail because files have
been removed.

## Testing

while : ; do pytest -v tests/test_source.py || break ; done

## Deployment

This is test only, no deployment impact

### DIFF
--- a/securedrop/tests/utils/env.py
+++ b/securedrop/tests/utils/env.py
@@ -5,6 +5,7 @@ import os
 from os.path import abspath, dirname, exists, isdir, join, realpath
 import shutil
 import subprocess
+import threading
 
 import gnupg
 
@@ -59,6 +60,12 @@ def setup():
 
 
 def teardown():
+    # make sure threads launched by tests complete before
+    # teardown, otherwise they may fail because resources
+    # they need disappear
+    for t in threading.enumerate():
+        if t.is_alive() and not isinstance(t, threading._MainThread):
+            t.join()
     db_session.remove()
     try:
         shutil.rmtree(config.SECUREDROP_DATA_ROOT)


dachary:
retest

securedrop/tests/utils/env.py
```diff
@@ -59,6 +60,9 @@ def setup():
 
 
 def teardown():
+    for t in threading.enumerate():
```
redshiftzero: I know that you have a nice descriptive commit message :smile: and that we can git blame, but let's make sure this is as clear as possible for future developers (and future us) by adding a comment above this line explaining why we added this to the `teardown` (basically just what you have in the commit message).

dachary: good idea. I added a comment, possibly with typos... I'd be grateful if you could verify ;-)

PR #2135, opened 2017-08-16
i18n: package securedrop-app translation files

dachary:
## Status

Ready for review

## Description of Changes

Add the i18n files to the app package so they are installed and useable.

## Testing

* make build-debs

Note that the build process will use ./manage.py --verbose translate to build the .mo file. This must be verified to run successfully. There are no translations yet so it will not produce any file. But it should not fail either.

## Deployment

Although the app package now contains more files, they are not used and are not concern for existing deployments.

### DIFF
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/translations.yml
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/translations.yml
@@ -1,0 +1,23 @@
+---
+- name: Install dependencies for building translations
+  apt:
+    pkg: "{{ item }}"
+    state: present
+  with_items: "{{ appserver_dependencies }}"
+  tags: apt
+
+- name: Install pip dependencies for SecureDrop.
+  pip:
+    requirements: "{{ securedrop_code_filtered }}/requirements/securedrop-requirements.txt"
+  tags:
+    - pip
+
+- name: Compile PO to MO.
+  shell: >-
+    cp config.py.example config.py ;
+    trap 'rm config.py' EXIT ;
+    ./manage.py --verbose translate --compile
+  args:
+    chdir: "{{ securedrop_code_filtered }}"
+  environment:
+    PYTHONDONTWRITEBYTECODE: "true"
--- a/securedrop/translations/messages.pot
+++ b/securedrop/translations/messages.pot
@@ -1,0 +1,18 @@
+# Translations template for SecureDrop.
+# Copyright (C) 2017 Freedom of the Press Foundation
+# This file is distributed under the same license as the SecureDrop project.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2017.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: SecureDrop 0.3.12\n"
+"Report-Msgid-Bugs-To: securedrop@freedom.press\n"
+"POT-Creation-Date: 2017-07-03 15:42+0000\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=utf-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Generated-By: Babel 2.4.0\n"
+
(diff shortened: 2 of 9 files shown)


dachary:
rebased & repushed

dachary:
@redshiftzero thanks for the carefully review. I'll prepare the next item :-)

dachary:
rebased & repushed to trim the **18n: manage: make translate --compile more pythonic** commit that was merged via another pull request.

redshiftzero:
Arg sorry @dachary - rebase?

dachary:
rebased & repushed :-)

securedrop/translations/messages.pot
```diff
@@ -0,0 +1,18 @@
+# Translations template for SecureDrop.
+# Copyright (C) 2017 Freedom of the Press Foundation
+# This file is distributed under the same license as the SecureDrop project.
+# FIRST AUTHOR <EMAIL@ADDRESS>, 2017.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: SecureDrop 0.3.12\n"
```
conorsch: I spy a hardcoded version string! If we cannot make this a Jinja var, then it must be added to the `./update_version.sh` script so that it gets incremented during the normal release process.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/translations.yml
```diff
@@ -0,0 +1,18 @@
...
+- name: Install pip dependencies for SecureDrop.
+  pip:
+    requirements: "{{ securedrop_code_filtered }}/requirements/securedrop-requirements.txt"
+  tags:
+    - pip
+
+- name: Compile PO to MO.
+  shell: "cp config.py.example config.py ; trap 'rm config.py' EXIT ; PYTHONDONTWRITEBYTECODE=true ./manage.py --verbose translate --compile"
+  args:
+    chdir: "{{ securedrop_code_filtered }}"
```
conorsch: Crafty! Hard to read, though. Seems like the desired control flow is:

1. Ensure that `config.py` exists.
2. Run the `manage.py translate --compile` command.
3. Remove the `config.py` file.

We could definitely break this up into separate tasks to help readability, but perhaps the functionality would be clear to future maintainers. Still, a reformatting of the current logic like so:

```
diff --git a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/translations.yml b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/translations.yml
index 0c9a574d..a4653733 100644
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/translations.yml
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/translations.yml
@@ -13,6 +13,11 @@
     - pip
 
 - name: Compile PO to MO.
-  shell: "cp config.py.example config.py ; trap 'rm config.py' EXIT ; PYTHONDONTWRITEBYTECODE=true ./manage.py --verbose translate --compile"
+  shell: >-
+    cp config.py.example config.py ;
+    trap 'rm config.py' EXIT ;
+    ./manage.py --verbose translate --compile
   args:
     chdir: "{{ securedrop_code_filtered }}"
+  env:
+    PYTHONDONTWRITEBYTECODE: true
```

Strikes me as an improvement. (Note I haven't actually tested the patch above.)

dachary:
> Will prioritize doing so next week.

Since code freeze is tuesday and monday is labor day, it is reasonable to say this commit missed the merge window. It would have been good to merge it, of course. But it's not a big deal if it is not.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/translations.yml
dachary: Much better to read indeed, thanks for the diff :-)

securedrop/translations/messages.pot
dachary: This is a generated file (from ./manage.py translate) so it is updated every time it is re-generated.

The only reason for committing it is to communicate with the weblate server. It goes like this:

* strings changes in code and templates
* they are [committed to messages.pot](https://docs.securedrop.org/en/latest/development/i18n.html#updating-strings-to-be-translated)
* weblate.securedrop.club pulls the repository and updates its indices with the content of messages.pot

Does this seem like an acceptable workflow to you ?

dachary:
repushed with the suggested change

dachary:
> I do see a failing test:

Hum... I don't see how this failure can be related to this diff. Rebased & repushed. And updated the build instructions, good point !

dachary:
this tests pass now

conorsch:
:ok_hand: Looking good! Thanks for all the back-and-forth on this one, @dachary—glad it's finally going in!

dachary:
@conorsch would you like me to do something else before this can be merged ?

Issue #2134, opened 2017-08-16
Comprehensive testing for OSSEC rules


redshiftzero:
## Description

We should have automated testing to ensure that particular log events do or do not produce OSSEC alerts. We can begin by adding regression tests when patching the OSSEC rules.

It appears that the right tool for testing if particular log events do or do not produce OSSEC alerts is [`ossec-logtest`](https://ossec-docs.readthedocs.io/en/latest/manual/rules-decoders/testing.html). This is currently installed on `mon`. For example, if I want to test whether the log event described in PR #871 produces an alert, I can simply pass the event to `ossec-logtest`, it will read the rules that are currently configured, and inform me whether or not an OSSEC rule would be produced, and if so, what the alert level would be:

```
root@mon-staging:/home/vagrant# sudo echo "Feb 10 23:34:40 app-prod kernel: [  124.188641] grsec: denied RWX mmap of <anonymous mapping> by /usr/sbin/apache2[apache2:1328] uid/euid:33/33 gid/egid:33/33, parent /usr/sbin/apache2[apache2:1309] uid/euid:0/0 gid/egid:0/0" | /var/ossec/bin/ossec-logtest
2017/08/16 22:28:25 ossec-testrule: INFO: Reading local decoder file.
2017/08/16 22:28:25 ossec-testrule: INFO: Started (pid: 18973).
ossec-testrule: Type one log per line.



**Phase 1: Completed pre-decoding.
       full event: 'Feb 10 23:34:40 app-prod kernel: [  124.188641] grsec: denied RWX mmap of <anonymous mapping> by /usr/sbin/apache2[apache2:1328] uid/euid:33/33 gid/egid:33/33, parent /usr/sbin/apache2[apache2:1309] uid/euid:0/0 gid/egid:0/0'
       hostname: 'app-prod'
       program_name: 'kernel'
       log: '[  124.188641] grsec: denied RWX mmap of <anonymous mapping> by /usr/sbin/apache2[apache2:1328] uid/euid:33/33 gid/egid:33/33, parent /usr/sbin/apache2[apache2:1309] uid/euid:0/0 gid/egid:0/0'

**Phase 2: Completed decoding.
       decoder: 'iptables'

**Phase 3: Completed filtering (rules).
       Rule id: '100101'
       Level: '7'
       Description: 'grsec error was detected'
**Alert to be generated.
```

We can use this `ossec-logtest` utility to create new `testinfra` tests for our OSSEC rules. 

[...]

PR #2137, opened 2017-08-16
Add initial automated tests for OSSEC and whitelist overloaded Tor guard log event

### LINKED ISSUE #1670: Filter out Tor Guard overloaded log event in OSSEC
I propose we filter out the following alert:  

```
Received From: (app)
Portion of the log(s):

[warn] Your Guard <name> ($fingerprint) is failing a very large amount of
circuits. Most likely this means the Tor network is overloaded, but it
could also mean an attack against you or potentially the guard itself.
```

It provides some information, but there's no action that we're expecting the SecureDrop admins to take here. It is however an interesting example of a situation where we might want to collect this information for analysis in order to look at trends across SecureDrop instances (e.g. if we implement #973). As with #1667, in order to alleviate alert fatigue, if we're going to be sending email alerts to administrators, we should make sure that it's either critical information or requires some action. 

We should also fix the docs [here](https://docs.securedrop.org/en/latest/ossec_alerts.html#common-ossec-alerts) regarding this rule.

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #1670 and begins work towards #2134

  * 523447e - Adds initial example of automated testing for OSSEC rules based on log event in PR #871
  * dd40158 - Adds a failing test to reproduce spurious OSSEC alert described in ticket #1670
  * 32b18ad - Adds an OSSEC rule to suppress the spurious email notification for the log event in ticket #1670 and thus makes the failing test added in dd40158 pass

(Happy to break this up into separate PRs to aid in review)

## Testing

At minimum, verify the new tests run and pass in the CI staging environment.

If you want to be super thorough:

1. Provision staging on dd40158 and see that `test_overloaded_tor_guard_does_not_produce_alert` fails ("Alert to be generated" appears in the output of `ossec-logtest`)
3. Provision staging on 32b18ad and see that `test_overloaded_tor_guard_does_not_produce_alert` now passes ("Alert to be generated" does not appear in the output of `ossec-logtest`)

## Deployment

This just updates the OSSEC rules, so there should be no issues in deployment.  

- [x] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass on app and mon

### DIFF
--- a/install_files/securedrop-ossec-server/var/ossec/rules/local_rules.xml
+++ b/install_files/securedrop-ossec-server/var/ossec/rules/local_rules.xml
@@ -128,3 +128,16 @@
     <options>no_email_alert</options>
   </rule>
 </group>
+
+<!--
+  Do not send an email alert on overloaded Tor guard events.
+  These are purely informational notifications, but would be
+  a candidate for sending up to FPF for analysis in aggregate.
+-->
+<group name="tor guard overloaded">
+  <rule id="200002" level="0">
+    <if_sid>1002</if_sid>
+    <match>this means the Tor network is overloaded</match>
+    <options>no_email_alert</options>
+  </rule>
+</group>
--- a/testinfra/mon/test_ossec_ruleset.py
+++ b/testinfra/mon/test_ossec_ruleset.py
@@ -1,0 +1,21 @@
+import re
+
+
+alert_level_regex = re.compile(r"Level: '(\d+)'")
+
+
+def test_grsec_denied_rwx_mapping_produces_alert(Command, Sudo):
+    """Check that a denied RWX mmaping produces an OSSEC alert"""
+    test_alert = ("Feb 10 23:34:40 app kernel: [  124.188641] grsec: denied "
+                  "RWX mmap of <anonymous mapping> by /usr/sbin/apache2"
+                  "[apache2:1328] uid/euid:33/33 gid/egid:33/33, parent "
+                  "/usr/sbin/apache2[apache2:1309] uid/euid:0/0 gid/egid:0/0")
+
+    with Sudo():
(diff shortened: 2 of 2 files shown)


conorsch:
Hmm, I would expect the `RWX mmap of <anonymous mapping> by /usr/sbin/apache2` alert never to fire, given that we set `kernel.grsecurity.rwxmap_logging=0` in the grsecurity role. @redshiftzero have you observed that particular event on staging machines?

redshiftzero:
Hmm so this isn't `apache2`, but I see this event: 

`Aug 16 21:58:28 app-staging kernel: [    7.847843] grsec: denied RWX mmap of <anonymous mapping> by /usr/bin/landscape-sysinfo[landscape-sysin:279] uid/euid:0/0 gid/egid:0/0, parent /usr/share/landscape/landscape-sysinfo.wrapper[50-landscape-sy:269] uid/euid:0/0 gid/egid:0/0` 

in `/var/log/syslog` on `app-staging` provisioned today.

conorsch:
Spinning up VMs to test out these changes.

conorsch:
We have alerting tests! 😍 Thanks @redshiftzero!

I'll note now that we're definitely going to want to DRY out the alerting config tests by writing a reusable function (should technically be a fixture in pytest/testinfra nomenclature, but a helper function will work too), and reading in the alert metadata via the YAML testinfra vars files. Won't block merge for that, since we have several pending alert test PRs (e.g. #2143, #2152). 

Will need to circle back on the dev docs anyway to explain the alert testing workflow, we can handle DRY cleanup then.

redshiftzero:
Nice, thanks for reviewing this @conorsch - I'm with you on DRYing up the alerting config tests and adding developer docs to explain that people should be writing regression tests for OSSEC rule changes (to address in a followup).

PR #2157, opened 2017-08-20
Update backup documentation to use gksu nautilus in lieu of "Root Terminal" application

### LINKED ISSUE #2069: Tails 3 root nautilus backups
# Bug

## Can't start nautilus from root terminal

Since Tails 3, you'll have to do something like `gksu nautilus` to launch a root nautilus. You'll get a `GTK-Warning cannot open display: :1` **if you have run a root file manager before in the same session**

## Steps to Reproduce

Described here:
https://docs.securedrop.org/en/latest/upgrade_to_tails_3x.html#upgrade-the-tails-drives
https://docs.securedrop.org/en/latest/backup_workstations.html

## Actual Behavior

nope-0

## Comments

gksu nautilus from a regular terminal would work. When reproducing this bug, I found that it only happens when having run a root nautilus in the same sesion before.

edenemmanuel:
## Status

Ready for review

## Description of Changes

Fixes #2069. 

Updates documentation to not use the "Root Terminal" application and have the user start a root Nautilus that way. Instead, we have the user start the standard "Terminal" application and run `gksu nautilus`, as that is recommended as of Tails 3.x.

This will also add appropriate images and remove one image no longer used in the docs.

## Testing

Make sure it builds correctly, and run `make docs-lint` to check for text formatting errors. Furthermore, any reviewer should take care to check if spelling and grammar is appropriate, and for compliance with the [SecureDrop Documentation Style Guide](https://docs.securedrop.org/en/stable/development/documentation_guidelines.html#style-guide).

## Deployment

Upon release, documentation will be sent out to stable docs.

## Checklist
- [x] Compliance with SecureDrop Documentation Style Guide


dachary:
In the commit messages at 05b419e : s/remove remove/remove/

dachary:
s/
In the commit title at 202c0d7 s/Updates images for using gksu nautlus/Updates images for using gksu nautilus/

dachary:
The **root_terminal.png** file should also be removed.

edenemmanuel:
Ah, thanks for pointing out the typos @dachary. Probably should've slept more. :sweat_smile:.

Fixing.

edenemmanuel:
Added new commit to remove *root_nautilus.png* and force pushed updated commit messages. Thanks @redshiftzero for the help! :)

dachary:
It may be easier for the user to open the terminal via "Favorites" instead of "Utilities", don't you think ?

dachary:
This is the error I get when running `gksu nautilus`  on a fresh tails-3.1 with a persistent volume available and unlocked. It is the same error I get without a persistent volume unlocked. The error does not show when `gksu nautilus` is run again. And the error does not prevent navigation via "Other Locations".

screenshot from 2017-08-20 13-01-46

edenemmanuel:
Indeed, I was able to replicate that error in hardware. 

I still am fine with this, because in the two pieces of documentation affected, we recommend the admin use a "Master" Tails 3.x drive. The "Master" drive *should not* have persistence enabled in any case, rendering that issue moot.

Admittedly, it's not a perfect fix. It appears `gksu`, `gksudo`, and "Root Terminal" (provided by `gksu`) seems to [be on the way out](https://labs.riseup.net/code/issues/12738) in Tails 3.2 or in a future release. There's some talk of using  `pkexec` (which would be an improvement generally), but running `pkexec nautilus` does nothing for me as tested on Tails 3.1 w/ persistence enabled on hardware.

If there are still some admins who don't update their Tails sticks, we may have to apply another hotfix to our docs, but at present this is probably the best workaround for the time.

dachary:
> the best workaround for the time.

That's sensible indeed. The documentation should mention this error and the fact that it is not a concern, don't you think ?

redshiftzero:
Hmm, so I saw the same error that @dachary reported also on a Tails 3.1 master drive without persistence. Given this, I agree that mentioning this error in a small note box (along the lines of "If you see error X, it is not a concern and you should do Y") is probably the best course of action.

edenemmanuel:
Seems reasonable @redshiftzero @dachary, especially given the fact that @redshiftzero has seen it without persistence.

A note is a great idea. Will add.

Issue #1825, opened 2017-06-15
Stablize AWS / CircleCI CI Builds


msheiny:
# Bug

## Description

Occasionally the CircleCI staging run will fail during the provisioning step of the build on the [Wait for EC2 instances to boot](https://github.com/freedomofpress/securedrop/blob/develop/devops/playbooks/aws-ci-startup.yml#L133) task. This step attempts to connect to port 22 on both ec2 boxes (`app` and `mon`) from the circleci environment to check whether SSH is up. 

## Steps to Reproduce

Unable to exactly produce but this is seen in a number of recent random builds:
* https://circleci.com/gh/freedomofpress/securedrop/526
* https://circleci.com/gh/freedomofpress/securedrop/523
* https://circleci.com/gh/freedomofpress/securedrop/506

## Expected Behavior

The CI servers should spin up and be available on port 22 from the POV of the circleci builder.

## Actual Behavior

In each of scenario it seems either one or both of the AWS servers are unavailable over port 22 and the task times out and the servers are torn down. The job will error without much fanfare (but a lot of distracting ansible output). An example looks something like this:

```
TASK [Create custom temporary security_group for CI hosts.] ********************
changed: [localhost]

TASK [Create remote CI hosts on AWS.] ******************************************
changed: [localhost] => (item={u'server_type': u'mon-staging'})
changed: [localhost] => (item={u'server_type': u'app-staging'})

TASK [Wait for EC2 instances to boot.] *****************************************
failed: [localhost] (item={'_ansible_parsed': True, u'changed': True, '_ansible_no_log': False, u'instances': 
[...]
"msg": "Timeout when waiting for ec2-54-219-170-84.us-west-1.compute.amazonaws.com:22"}
ok: [localhost] => (item={'_ansible_parsed': True, u'changed': True, '_ansible_no_log': False, u'instances': 
[..] )
	to retry, use: --limit @/root/sd/devops/playbooks/aws-ci-startup.retry

PLAY RECAP *********************************************************************
localhost                  : ok=12   changed=7    unreachable=0    failed=1   

make[2]: Entering directory '/root/sd'
./devops/scripts/ci-teardown.sh
+ . devops/ansible_env
++ '[' -z 506 ']'
++ [[ 506 == \$\{* ]]
++ export RETRY_FILES_ENABLED=False
++ RETRY_FILES_ENABLED=False
++ export ANSIBLE_INVENTORY=./devops/inventory/staging
[...]

Issue #1637, opened 2017-03-30
Use lintian to validate deb package build process


conorsch:
Beyond simply ensuring that the debs don't have unwanted files (#1507), we should be use `lintian` to inspect the built deb packages in the config test suite and fail out if checks aren't satisfied.

By default lintian performs comprehensive validation, and we likely won't agree with every recommendation. Still, we can selectively opt in to certain checks via the `--tags` option. Good candidates for inclusion in the config test suite are:

* wrong-file-owner-uid-or-gid
* package-contains-vcs-control-file
* package-installs-python-bytecode
* non-standard-file-perm

A parametrized config test that runs all of the above would be grand. In the context of config tests that inspect warnings, the `--fail-on-warnings` flag should be used. See `lintian-info --tags <tag>` for an explanation of what a tag check entails.

t2d:
I started to do a very basic implementation at https://github.com/t2d/securedrop/commit/1f543c2a8ca274af6f89fb96a7005c356da9ef0f
But this actually makes the verifier fail. In what direction do you want this to progress?

dachary:
@t2d I'm sure @conorsch will be very happy to see progress in this direction :-)

Issue #308, opened 2014-02-19
Independently repeatable fully deterministic compilation


tildelowengrimm:
It should be possible for independent folks to compile SecureDrop from source and achieve exactly the same binary. Once this is possible, SecureDrop's normal release process should rely on multiple independent builders.

For added safetly, the normal release process should also expect that others are secretly building SecureDrop the same way, and provide a mechanism for issuing alerts the official builds are incorrect. This mechanism should be tested regularly (but not often). In a test, all the official builders collude to make an innocuous change (perhaps the addition of whitespace), and publish their build. Then you see how long it takes for someone to sound the alarm.

dpkp:
gzip's default behavior is to include the timestamp of the uncompressed file, which for tarballs is the timestamp of the intermediate tar archive. That means that your tarball is including the timestamp of the build run, which should explain why your bytes are different on each build even if the source is unchanged. Try this instead if you still want to use tar+gz:
```
tar cf foobar.tar foobar/
gzip -n foobar.tar
```
also see http://superuser.com/questions/705877/compressing-compressed-tar-gz-files-deterministically

fowlslegs:
So this post was from a while ago and obviously things have changed since this was last being discussed. Whatever was the past, it makes sense now to work on making our `.deb` packages reproducible. Everything else is already deterministic in the sense that our repository contains no executable binaries. Here is Debian's guide to doing so: https://wiki.debian.org/ReproducibleBuilds/Howto. Since we have automated the process of building the packages we ship by using a 'build' VM provisioned by Ansible, it should be easy to standardize package versions for our toolchain and make the other changes necessary so that anyone can build SecureDrop packages reproducible from the comfort of their beds.

fowlslegs:
Also, note at some point down the road we would like to migrate to Debian. This may not be for years due to the massive migration and support costs, but perhaps sometime before April 2019, when the current Ubuntu LTS hits EOL. The amd64 testing repository of Debian (i.e., Debian 9/Stretch) is already at 89.5% reproducibility (https://tests.reproducible-builds.org/reproducible.html). Thus, a *down-to-the-kernel* reproducible SD app and mon server is possible in the coming years. I could also say we should go further and recommend the ASUS KGPE-D16 for which we could reproducibly build a blobless version of coreboot called libreboot for a *down-to-the-boot-firmware* reproducible system (but I won't go there, I'll just link it for fun and interest https://libreboot.org/docs/hcl/kgpe-d16.html).

[...]

PR #2160, opened 2017-08-20
Refactor Debian Package Build logic

msheiny:
## Status

🚥 Ready for Review 🚥 

## Description of Changes

So I first started this PR to address #1825 BUTT in order to do that I needed to start converting the CI logic to moleculev2 (believe me it makes things much easier ). This branch started getting too big so I figured lets start testing it as-is and pull a PR. *ahem* 

So what does this PR contain? It converts the current docker build logic for creating debian packages:

- Utilizes a non-privileged Ubuntu trusty docker image (vs. previously utilized Debian jessie image)
- Provides tooling for local dev and removes need for Vagrant build flow
- Takes advantage of molecule which will provide much easier to maintain in the long-run.

## Testing

In order to test the following things need to occur:

- Deb packages need to get thrown against CI ✔️ 
- a full vagrant run of the staging environment locally needs to occur

## Deployment

Doesnt really affect deployments in the wild BUT this does affect building debian packages for deployment (by SD maintainers)


dachary:
I was confused by this PR until I realized molecule is used both for testing and to produce debian packages. I think it would be saner to have a role whose sole purpose is to build deb packages using docker. And molecule testing this role actually produces the artifcats it is supposed to. Does that make sense ?

msheiny:
> I was confused by this PR until I realized molecule is used both for testing and to produce debian packages. I think it would be saner to have a role whose sole purpose is to build deb packages using docker. And molecule testing this role actually produces the artifcats it is supposed to. Does that make sense ?

Well I mean there is already a playbook `install_files/ansible-base/build-deb-pkgs.yml` and several ansible roles that do that:
* ./install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg
* ./install_files/ansible-base/roles/build-generic-pkg
* ./install_files/ansible-base/roles/build-ossec-deb-pkg

`molecule/builder/playbook.yml` is a symlink to `install_files/ansible-base/build-deb-pkgs.yml`

This molecule scenario just takes advantage of all of those playbooks/roles + provides a scaffolding for running the testinfra tests afterwards. You just have to run `make build-debs` now vs. `vagrant up build`.   If you really want to skip testing and just build the debs you can run `molecule converge -s builder`

dachary:
I can't help but think using a test framework as a tool to produce artifacts (deb in this case) is a clever hack that we'll eventually regret. Why not just make it so `make build-debs` provisions docker instead of vagrant + builds the packages without requiring molecule ? And use molecule solely for the purpose of testing that.

msheiny:
The nice thing about moleculev2 is that there is very little abstraction being done here -- and I will concede that one of the things I didnt like about moleculev1 was that there was a ton of weird abstraction going on. In fact, I remember @conorsch was arguing when I laid out the CI scripts that are currently being used that we should use molecule v1 -- but it was just a huge pain in the butt.

BUT Moleculev2 in itself is basically a clever wrapper around a bunch of other tools: vagrant, docker, ansible, testinfra, flake8, ansible-lint. I was essentially doing what molecule is doing before with bash scripts. The problem with controlling this logic with bash scripts is that we have to do this same thing across multiple repos. So overtime  I start to get a little better at organizing the scripts with every new repo, and when I look back at older repos it doesnt make sense. So then I'm maintaining and updating a slew of bash scripts across multiple repos -- which is super painful and leads to confusion when people go to update. Using something like molecule brings a standard to the chaos -- and is very easy to rip out later if we deem it a problem.

Basically I am doing what molecule is doing but using my own shell

msheiny:
Tested locally with staging VMs 👍

dachary:
> The problem with controlling this logic with bash scripts is that we have to do this same thing across multiple repos.

I was not suggesting to revert to bash scripts :-)

> is very easy to rip out later if we deem it a problem.

I respectfully disagree and encourage you to consider an alternative approach.

conorsch:
This PR doesn't close an issue, which makes it difficult to review. That seems to be part of the problem with understanding its intent, and how effectively it meets the implicit goals. 

> start converting the CI logic to moleculev2 (believe me it makes things much easier )

@msheiny We need a discrete issue for the Molecule v2 migration. I'll open one. I agree with you that migrating to Molecule v2 would allow us to delete a ton of custom bash from the repo, but there's no issue tracking that discussion, so it's difficult for other folks to follow along. 

Also let's consider again the fact that we have a dozen or so _custom_ roles in this repository, which means we're not making use of community-maintained or even FPF-maintained configuration logic, which makes refactors like this one cumbersome. Now that we have stronger configuration testing, I would be happy ripping out the custom roles in favor of reusable ones (likely via `git subtree`) and rely on the config tests to guard against regressions.

> I was not suggesting to revert to bash scripts :-)

@dachary Unfortunately that's what we've got right now, for the CI tooling located in the `devops/` directory. Moving to Molecule v2 is a great idea for CI, but again, that discussion belongs in a separate issue. 

> I can't help but think using a test framework as a tool to produce artifacts (deb in this case) is a clever hack that we'll eventually regret.

A fair point! However we're already using Vagrant, a tool intended for developer environments, to build debs, and have been doing so for years. By swapping in Molecule for controlling the Vagrant logic, we get much more powerful testing and linting for free. Still interested in adding stronger validation of the build process (e.g. #308, #1637), and adding those features _without_ including Molecule would involve a ton of custom scripting, and I'd much rather rely on a strong community-maintained solution.

msheiny:
rebased off `develop` to solve potential merge conflict

msheiny:
> This PR doesn't close an issue, which makes it difficult to review. That seems to be part of the problem with understanding its intent, and how effectively it meets the implicit goals.

👍 -- @conorsch see #2198

msheiny:
@conorsch ping

conorsch:
Reviewing!

conorsch:
@msheiny Could you rebase this on top of latest develop? I'm already seeing some logic missing from this branch, e.g. #2191.

msheiny:
rebased 👍

conorsch:
@heartsucker Would appreciate a functional test from you, meaning confirming that staging still works on your machine. You should: 

1. Destroy all local VMs.
2. Check out this feature branch.
3. Run `make build-debs`. 
4. Run `vagrant provision /staging/`.
5. Run config tests for app-staging and mon-staging.
6. Run app tests on app-staging (including `--page-layout` tests). 

If any errors show up, please post them here.

install_files/ansible-base/securedrop-staging.yml
```diff
@@ -1,10 +1,22 @@
...
+
+    - set_fact:
+        expected_debs: "{{hostvars['app-staging']['local_deb_packages'] + hostvars['mon-staging']['local_deb_packages']}}"
+        found_debs: "{{ search_deb_results.files | map(attribute='path') | list }}"
+
+    - name: Fail if package is missing from build dir
+      fail:
+        msg: "run 'make build-debs' - ../../build/{{item}} missing"
+      when: "'../../build/'+item not in found_debs"
+      with_items: "{{ expected_debs }}"
```
conorsch: Nit: It would be slightly more readable to use a `|basename` filter on the `found_debs` vars declaration, then the `when` conditional wouldn't need to prefix the build path.

molecule/builder/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:14.04
+
+RUN apt-get -y update && apt-get install -y python rsync sudo bash \
+            devscripts git aptitude ipython libssl-dev ntp \
+            python-dev python-pip python-ipdb ruby tmux vim paxctl && \
+            aptitude -y dist-upgrade && \
+            apt-get clean
```
conorsch: Bummer to have these hard-coded rather than vars-based, but that's Docker for ya. Worth adding some config tests to track these explicitly, since many of them were present implicitly in the Ubuntu Trusty build VM.

molecule/builder/create.yml
```diff
@@ -0,0 +1,24 @@
...
+    docker_hash: "{{ lookup('file', '../docker_hashes.yml') | from_yaml }}"
+  tasks:
+    - name: Create molecule instance(s)
+      docker_container:
+        name: "{{ item.name }}"
+        hostname: "{{ item.name }}"
+        image: "{{ item.image }}@{{ docker_hash[item.hash_name] }}"
+        state: started
+        recreate: True
+        command: "tail -f /dev/null"
```
conorsch: Crafty. I would have used `sleep infinity` instead, but either works fine for a forever process.

molecule/builder/tests/conftest.py
```diff
@@ -0,0 +1,15 @@
...
+import yaml
+
+
+def pytest_namespace():
+    """ Return dict of vars imported as 'securedrop_test_vars' into pytest
+        global namespace
+    """
+    filepath = os.path.join(os.path.dirname(__file__), "vars.yml")
+    with open(filepath, 'r') as f:
+        return dict(securedrop_test_vars=yaml.safe_load(f))
```
conorsch: Hmm, we'll need to duplicate these conftest.py files across multiple scenarios in Molecule (has bearing on #2218). I wonder if there's a better way?

msheiny: Well FYI , #2218 doesnt need this because I was unable to use molecule to kick off the testinfra tests there in a way that wouldn't cause a huge refactor. In that PR, we are calling the testinfra tests with a similar shell script that we currently have (took existing and stripped away unneeded bits)

docs/development/getting_started.rst
```diff
@@ -128,6 +137,7 @@ different version, the path to ``virtualenvwrapper.sh`` will differ. Running
     source /usr/local/bin/virtualenvwrapper.sh
     mkvirtualenv -p python2 securedrop
     pip install -r securedrop/requirements/develop-requirements.txt
+    pip install -r testinfra/requirements.txt
```
conorsch: While testing this PR branch, I intuitively reran the `develop-requirements` files and confirmed that Molecule was _not_ added, even though it's a requirement for development now. We should fix the requirements story quite soon. Willing to let it pass here and in #2218, and continue tracking in #1918 & #2175.

dachary:
> Run make build-debs.

It requires permissions to talk to the docker daemon which is not the case for the non-root users. It can be fixed easily by adding the user to the docker group and then running newgrp docker. It would be nice to have a more explicit message about that. Currently it looks like
<pre>
    TASK [Destroy molecule instance(s)] ********************************************
    failed: [localhost] (item=(censored due to no_log)) => {"censored": "the output has been hidden due to the fact that 'no_log: true' was specified for this result"}
</pre>
and after disabling the
<pre>
#  no_log: "{{ not lookup('env', 'MOLECULE_DEBUG') | bool }}"
</pre>
line in **molecule/builder/destroy.yml** it looks like
<pre>
    TASK [Destroy molecule instance(s)] ********************************************
failed: [localhost] (item={'image': u'quay.io/freedomofpress/ubuntu:trusty', 'hash_name': u'ubuntu', 'name': u'build', 'groups': [u'build']}) => {"failed": true, "item": {"groups": ["build"], "hash_name": "ubuntu", "image": "quay.io/freedomofpress/ubuntu:trusty", "name": "build"}, "msg": "Error connecting: Error while fetching server API version: ('Connection aborted.', error(13, 'Permission denied'))"}
</pre>
which is not really more explicit.

heartsucker:
@conorsch Am testing now.

heartsucker:
Other than this one error, everything works.

```
(venv) heartsucker@pythagoras:~/code/freedom-of-press/securedrop (StabilizeCI)$ ./testinfra/test.py app-staging
Running Testinfra suite against 'app-staging'...
Target roles:
    - testinfra/app
    - testinfra/app-code
    - testinfra/common
    - testinfra/development/test_xvfb.py
  File "/home/heartsucker/code/freedom-of-press/securedrop/venv/local/lib/python2.7/site-packages/_pytest/config.py", line 379, in _importconftest
    mod = conftestpath.pyimport()
  File "/home/heartsucker/code/freedom-of-press/securedrop/venv/local/lib/python2.7/site-packages/py/_path/local.py", line 680, in pyimport
    raise self.ImportMismatchError(modname, modfile, self)
ImportMismatchError: ('conftest', '/vagrant/testinfra/conftest.py', local('/home/heartsucker/code/freedom-of-press/securedrop/testinfra/conftest.py'))
ERROR: could not load /home/heartsucker/code/freedom-of-press/securedrop/testinfra/conftest.py

  File "./testinfra/test.py", line 136, in <module>
    run_testinfra(target_host)
  File "./testinfra/test.py", line 132, in run_testinfra
    subprocess.check_call(testinfra_command)
  File "/usr/lib/python2.7/subprocess.py", line 186, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['testinfra', '-vv', '-n', 'auto', '--connection', 'ansible', '--ansible-inventory', '.vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory', '--hosts', 'app-staging', 'testinfra/app', 'testinfra/app-code', 'testinfra/common', 'testinfra/development/test_xvfb.py']' returned non-zero exit status 4
```

conorsch:
@heartsucker Run the command `./testinfra/test.py app-staging` from your host machine, not the VM guest. Then it should work for you! 

The failed attempt to run the testinfra checks inside the guest may have generated conflicting cache files—try running `find -type f -iname '*.pyc' -delete'` from your host machine prior to rerunning testinfra.

conorsch:
@heartsucker My apologies, I misread your terminal session. Looks like you did indeed run the command from your host machine, not from the guest. Try the pyc purge and let me know if that helps.

heartsucker:
Yes, that `find` command did it. How annoying.

Anywho. Green across the board.

redshiftzero:
Note that a testinfra test for the development environment needs updating here

conorsch:
> Note that a testinfra test for the development environment needs updating here

Hmm, all tests were passing previously, must have been the rebase. Go ahead and fix, I'm happy to re-review.

redshiftzero:
Got passing tests on `11e8e0e` locally. Let's see if Travis agrees.

dachary:
I confirm the following instructions (just modified to make them a little more explicit) worked for me as well :-)

Make sure docker ps works (without sudo)

1. Destroy all local VMs.
2. Run `pip install -r testinfra/requirements.txt`
2. Check out this feature branch.
3. Run `make build-debs`. 
4. Run `vagrant provision /staging/`.
5. Run config tests for app-staging and mon-staging testinfra/test.py app-staging ; testinfra/test.py mon-staging
6. Run app tests on app-staging (including `--page-layout` tests). 
<pre>
vagrant@app-staging:~$ sudo bash
root@app-staging:~# cd /var/www/securedrop/
root@app-staging:/var/www/securedrop# pytest -v --page-layout tests
</pre>

Issue #2165, opened 2017-08-22
vagrant provision build fails with ansible 2.3.2


dachary:
# Bug

## Description

It looks like quoting rules changed between ansible 2.2.1 and ansible 2.3.2 which causes a "sudo rsync" command not found error when running vagrant provision build (full sample output at https://paste2.org/4gV8d96f)

## Steps to Reproduce

* virtualenv /tmp/v
* source /tmp/v
* pip install ansible # with a ton of other dependencies from https://gist.github.com/trinitronx/bdfd3cbb338767e58299 if needed
* vagrant up build

## Expected Behavior

The packages are produced

## Actual Behavior

Fails with:
<pre>
fatal: [build]: FAILED! => {"changed": false, "cmd": "/usr/bin/rsync --delay-updates -F --compress --delete-after --archive --rsh=/usr/bin/ssh -S none -i /home/loic/.vagrant.d/insecure_private_key -o Port=22 -o StrictHostKeyChecking=no --rsync-path='sudo rsync' --out-format=<<CHANGED>>%i %n%L /home/loic/software/securedrop/securedrop/install_files/securedrop-app-code/ vagrant@192.168.121.234:/tmp/securedrop-app-code-0.4.2-amd64", "failed": true, "msg": "Warning: Permanently added '192.168.121.234' (ECDSA) to the list of known hosts.\r\nbash: sudo rsync: command not found\nrsync: connection unexpectedly closed (0 bytes received so far) [sender]\nrsync error: remote command not found (code 127) at io.c(235) [sender=3.1.2]\n", "rc": 127}
</pre>

## Comments

The workaround is to install explicitly ansible 2.2.1 as mentionned in securedrop/requirements/develop-requirements.txt

* pip install ansible==2.2.1

Another workaround is to apply the following patch but there may be other ansible version issues because of non backward compatible changes.

<pre>
modified   install_files/ansible-base/action_plugins/synchronize.py
@@ -365,5 +365,5 @@ class ActionModule(ActionBase):
                 # remote then add privilege escalation here.
                 if self._play_context.become_method == 'sudo':
-                    rsync_path = "'sudo rsync'"
+                    rsync_path = "sudo rsync"
                 # TODO: have to add in the rest of the become methods here
</pre>

ageis:
This is due to https://github.com/ansible/ansible/issues/23575.

dachary:
[...]

PR #2166, opened 2017-08-22
fail unless ansible version 2.2.3 is used

dachary:
## Status

Ready for review

## Description of Changes

Fixes: #2165 

fail unless ansible version 2.2.* is used

## Testing

* pip install ansible==2.2.1
* vagrant provision development # succeeds
* pip install ansible==2.3.2
* vagrant provision development # fails with the following
<pre>
SecureDrop restriction: only Ansible 2.2.* is supported. 
</pre>

## Comments

It will not affect developers because using ansible 2.3.* fails anyway and there is zero chance any developer is using this version.

There is a chance someone deploying SecureDrop has ansible 2.3.* installed and did not run into any compatibility problem. The corrective action in this case will be for her/him to **pip install ansible==2.2.3** or re-run the relevant **pip install -r ...requirements.txt** which will do the same.

## Deployment

It will fail a deployment that has ansible 2.3.* installed instead of ansible 2.2.* . The will not affect existing instances because the unattended upgrades do not rely on ansible.

## Checklist

I think all cases are covered by the run of the CI (CircleCI + Travis). Some of the ansible playbooks (at least those using synchronize) will fail with version 2.3+.

### DIFF
--- a/install_files/ansible-base/callback_plugins/ansible_version_check.py
+++ b/install_files/ansible-base/callback_plugins/ansible_version_check.py
@@ -1,0 +1,28 @@
+# -*- encoding:utf-8 -*-
+from __future__ import absolute_import, division, print_function, unicode_literals
+
+import sys
+
+import ansible
+
+try:
+    # Version 2.0+
+    from ansible.plugins.callback import CallbackBase
+except ImportError:
+    CallbackBase = object
+
+
+def print_red_bold(text):
+    print('\x1b[31;1m' + text + '\x1b[0m')
+
+
+class CallbackModule(CallbackBase):
+    def __init__(self):
+        # Can't use `on_X` because this isn't forwards compatible with Ansible 2.0+
+        required_version = '2.2'  # Keep synchronized with group_vars/all/main.yml
+        if not ansible.__version__.startswith(required_version):
+            print_red_bold(
+                "SecureDrop restriction: only Ansible {version}.* is supported. "
+                .format(version=required_version)
+            )
+            sys.exit(1)


dachary:
@msheiny this fails as expected for 
<pre>
ANSIBLE_ARGS="--tags=app-deb" vagrant provision build
</pre>
with
<pre>
TASK [build-securedrop-app-code-deb-pkg : Check for ansible version] ***********
fatal: [build]: FAILED! => {
    "assertion": false, 
    "changed": false, 
    "evaluated_to": false, 
    "failed": true, 
    "msg": "Ansible version 2.2.1 is required"
}
	to retry, use: --limit @/home/loic/software/securedrop/securedrop/install_files/ansible-base/build-deb-pkgs.retry
</pre>

However I'm not sure this is the best place to have this check since it should really be done whenever ansible runs. Unless we have different requirements depending on the role but that does not seem to be the case.

It also is probably a good candidate for molecule testing, where the roles are verified to fail when the wrong ansible version is being used. Or is there a simpler way to do the same without requiring manual testing ?

dachary:
> A more durable approach would be to wire up a callback plugin to check the version, as described here: https://adamj.eu/tech/2016/08/19/restricting-the-ansible-version-in-use/

Excellent ! Thanks for the pointer :-) I updated the PR to have that.

dachary:
@conorsch updated the PR description

conorsch:
Testing new changes now!

conorsch:
Looks great! Confirmed that Ansible 2.3 fails on local VMs via vagrant:

ansible-failure-debian

Also tested in Tails, even though it's quite difficult to install the "wrong" version of Ansible in tails, ever since #1781. To test, I activated the virtualenv within tails manually:

```
source ~/Persistent/securedrop/.venv/bin/activate
torify pip install ansible==2.3
```

Then tried the playbooks again and it failed, as it should:

ansible-failure-tails

The callback plugin is a solid approach—nice work, @dachary!

dachary:
> The callback plugin is a solid approach

Thank you for suggesting it :-) I gained a few XP in ansible !

PR #2173, opened 2017-08-22
made docs hostable from vagrant

### LINKED ISSUE #2170: Expose docs on port 8000 of vagrant development machine
This makes it easier for users to view the docs without having to install all the tools locally.

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #2170.

Expose port 8000 on `development` Vagrant vm.

```
vagrant reload development
vagrant ssh development
cd /vagrant
make docs
```

Then go to http://localhost:8000 in the browser.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+PWD := $(shell pwd)
+
 .PHONY: ci-spinup
 ci-spinup:
 	./devops/scripts/ci-spinup.sh
@@ -36,7 +38,7 @@ docs-lint:
 .PHONY: docs
 docs:
 # Spins up livereload environment for editing; blocks.
-	make -C docs/ clean && sphinx-autobuild --host 0.0.0.0 docs/ docs/_build/html
+	make -C docs/ clean && sphinx-autobuild `[ $(PWD) = /vagrant ] && echo '--host 0.0.0.0'` docs/ docs/_build/html
 
 .PHONY: flake8
 flake8:
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ docs-lint:
 .PHONY: docs
 docs:
 # Spins up livereload environment for editing; blocks.
-	make -C docs/ clean && sphinx-autobuild docs/ docs/_build/html
+	make -C docs/ clean && sphinx-autobuild --host 0.0.0.0 docs/ docs/_build/html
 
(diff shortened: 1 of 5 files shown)


redshiftzero:
On a fresh provision of this branch:

```
vagrant@development:/vagrant$ make docs
make -C docs/ clean && sphinx-autobuild --host 0.0.0.0 docs/ docs/_build/html
make[1]: Entering directory `/vagrant/docs'
Makefile:12: *** The 'sphinx-build' command was not found. Make sure you have Sphinx installed, then set the SPHINXBUILD environment variable to point to the full path of the 'sphinx-build' executable. Alternatively you can add the directory with the executable to your PATH. If you don't have Sphinx installed, grab it from http://sphinx-doc.org/.  Stop.
make[1]: Leaving directory `/vagrant/docs'
make: *** [docs] Error 2
```

redshiftzero:
Yep on a fresh provision of the development VM, I do not have `sphinx` installed.

heartsucker:
@redshiftzero I added some dependencies to the dev vm that resolve this. Can you recheck?

conorsch:
Works for me, and does not break the config tests. I wonder, though, whether we're solving the right problem: seems like a local docker container would be a smoother workflow, albeit with a docker dependency. For another day, I suppose.

The in-VM docs-building logic is markedly slower than the in-virtualenv approach, running on Debian Stretch, so I plan to continue using `make docs` _outside_ the dev VM.

Not thrilled about the listen-on-all-interfaces logic, even though host-based firewalls should deny it. A conditional based on whether PWD is `/vagrant` would suffice to append the `--host 0.0.0.0` option if-and-only-if the command is invoked from the dev VM.

heartsucker:
@conorsch I think this helps new devs get everything setup so it's worth it. Agreed that docker is a better long term solution. Also, I added conditional logic to address your concern.

redshiftzero:
Hmm, I see this is coming from `docs-expose-vagrant-port` - let's make sure we are _not_ using the prefix `docs-` for branches unless it is a PR containing _only_ documentation (as CI will skip running the staging tests).

heartsucker:
@redshiftzero Got it. I thought it was "doc-y stuff that doesn't affect app code and ansible." I'll reopen.

PR #2191, opened 2017-08-24
Reboot only when necessary

### LINKED ISSUE #1704: Reboot staging servers on first provision
Currently the staging servers need to be manually rebooted after they are provisioned in order for the iptables config to take. Ideally this would be a task done the first time the servers are provisioned.

conorsch:
## Status

Ready for review.

## Description of Changes

Fixes #1704. Adjusts the reboot logic to avoid unnecessary reboots at the end of the prod playbook. The new logic reboots at the end of the playbook if-and-only-if the playbook run constitutes a first-time install, as determined by the firewall rules. Complements #1700.

## Testing
1. CI should pass! Changed the staging logic by _removing_ the custom reboot playbook from the Circle CI run. 
2. Provision staging VMs locally and confirm that you can run the config tests _without_ manually rebooting the machines. All config tests should pass.
3. Perform manual installation on prod VMs and confirm that the playbook does _not_ wait for the VMs to come back after rebooting at the end of the play (doing so would cause the playbook to fail, since the remote addresses change during that reboot, from local IPv4 to Onion URLs). 
4. Confirm that a _second_ run of `./securedrop-admin install` (after running `tailsconfig`, of course) does _not_ reboot the servers. 

## Deployment
Yes, there's the potential for breakage on the first-time install logic, which is why we should test via Tails. The changes as advertised will enable Admins to reapply the config without immediately rebooting the servers, which may be undesirable during normal work hours.


redshiftzero:
Verified all `app-staging` and `mon-staging` config tests pass in the staging environment _without_ a reboot :muscle:. Moving on to prod testing

PR #2196, opened 2017-08-25
i18n: manage: make translate --compile more pythonic

dachary:
## Status

Ready for review

## Description of Changes

Make it resemble the above test instead of having a shell equivalent. It is cosmetic only.

## Testing

* pytest -v test/test_manage.py

## Deployment

None.

### DIFF
--- a/securedrop/manage.py
+++ b/securedrop/manage.py
@@ -295,13 +295,9 @@ def translate(args):
         else:
             log.warning("no translations found (ok for tests, not otherwise)")
 
-    if args.compile:
+    if args.compile and len(os.listdir(args.translations_dir)) > 1:
         sh("""
         set -x
-
-        # stop here if there are no translations (for testing)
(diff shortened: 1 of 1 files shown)


securedrop/manage.py
```diff
@@ -295,13 +295,9 @@ def translate(args):
         else:
             log.warning("no translations found (ok for tests, not otherwise)")
 
-    if args.compile:
+    if args.compile and len(os.listdir(args.translations_dir)) > 1:
```
heartsucker: This latter check does nothing since there is a default value for this argument. What logic was this trying to capture?

dachary: it won't try to compile a translation directory that has only one file in it (i.e. the messages.pot file). This is useful now because there indeed are no translations. And it may also be useful later if a test fixture runs it and has no translation.

heartsucker: Ah yes. Makes sense. Approved!

Issue #2202, opened 2017-08-25
CI - produce junit test results for circleCI to digest


msheiny:
# Feature request

## Description

The CircleCI process takes a good ~20 minutes -- whenever it fails under the application tests it is very annoying to sift through the current output to determine what failed. Lets export these to junit and get them absorbed by CircleCI so we can get some clean output 👍 

## User Stories

As a developer, I want to quickly determine what tests are failing in CI so I can debug them.

msheiny:
Technically this ticket as written has been completed. There is additional work here that can be done to make it easier for developers to see why CI is failing. I'll open a ticket for that

PR #2206, opened 2017-08-26
[Do not merge] Investigating application tests in Circle CI

redshiftzero:
## Status

Never merge

## Description of Changes

Just making this PR to investigate how application test failures currently appear in Circle CI


redshiftzero:
(Was just seeing what an application test failures looked like on Circle CI in the context of #2202)

PR #2233, opened 2017-08-31
tests: use make -k instead of listing targets

dachary:
## Status

Ready for review

## Description of Changes

Using **make -k** ensures all targets run even if one fails and we don't need to explicitly duplicate the list of lint targets in the travis file

## Testing

* push a flake8 error
* look at the travis output and verify the html-lint checks run although the target is *after* flake8 in the Makefile

## Deployment

test only

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,10 +59,6 @@ script:
   # to alter the pip requirements, which would cause config tests to fail.
   # The develop requirements include sphinx tooling for docs linting.
   - pip install -r securedrop/requirements/develop-requirements.txt
-  # Intentionally *not* using `make lint` wrapper so as not to skip linting
-  # steps if one fails. Continue linting in CI, and report full status.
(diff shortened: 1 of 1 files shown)


dachary:
Verified an intentional flake8 error does not prevent the following target from beingrun:

https://travis-ci.org/freedomofpress/securedrop/builds/270420010?utm_source=github_status&utm_medium=notification

<pre>
flake8 --exclude='config.py' testinfra securedrop-admin \

		securedrop/*.py securedrop/management \

		securedrop/tests/functional securedrop/tests/*.py

securedrop/journalist.py:20:80: E501 line too long (161 > 79 characters)

make: *** [flake8] Error 1

html_lint.py --printfilename --disable=optional_tag,extra_whitespace,indentation \

make: Target `lint' not remade because of errors.

The command "make -k lint" exited with 2.
</pre>

PR #2237, opened 2017-08-31
Added pre-commit hook

### LINKED ISSUE #2234: Provide pre-commit hook for developers
# Feature request

## Description

As our testing and linting improves, we should provide an easy way for developers to catch formatting mistakes prior to opening a PR. A pre-commit that runs at least the flake8 tests for Python files would be grand.

Work was already done toward both PEP8 compliance (now resolved, see #886) and providing a pre-commit hook in #1471. We may be able to pluck out some commits from that branch done by @ajvb. If not, we should reimplement from scratch.

## User Stories

As a SecureDrop developer, I want to catch common formatting mistakes before opening a PR, to save time for both me and my other contributors (who would need to review the faulty PR).

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #2234. Supersedes #1471

Creates a simple pre-commit hook, adds docs about it.

## Testing

`cp git/pre-comit .git/hooks/` and then try to commit. The linters should run.

### DIFF
--- a/docs/development/contributor_guidelines.rst
+++ b/docs/development/contributor_guidelines.rst
@@ -33,17 +33,26 @@ Code Style
 ----------
 
 We use code linters to keep a consistent code quality and style. These linters
-also run in CI and will produce build failures. To avoid this, we suggest running
-the linters locally as you work.
+also run in CI and will produce build failures. To avoid this, we have included
+a git pre-commit hook. You can install it with the following command run at the
+root of the repository:
+
+.. code:: sh
+
+    ln -sf ../../git/pre-commit .git/hooks/pre-commit
+
 
 .. note::
-  The code linters are installed automatically on the Development VM.
-  To install the linting tools locally on your host machine, from the root
-  of the repo you can run the following:
+  The code linters are installed automatically on the Development VM, but for
+  the pre-commit hook to work, you will need to install the linting tools
+  locally on your host machine. From the root of the repo you can run the
+  following:
 
   .. code:: sh
 
      pip install -r securedrop/requirements/develop-requirements.txt
+
+
 
 Python
 ~~~~~~
--- a/git/pre-commit
+++ b/git/pre-commit
@@ -1,0 +1,30 @@
+#!/bin/bash
+set -e
+
+# This file will exist in $project_root/.git/hooks/, so we need to move back up
+# to $project_root. This apparent mess needs to happen because macOS doesn't
+# have `readlink -f`.
+declare src="$0"
+declare dir=''
+
+while [ -h "$src" ]; do
+  dir="$(cd -P "$( dirname "$src")" && pwd)"
+  src="$(readlink "$src")"
+  [[ "$src" != /* ]] && src="$dir/$src"
+done
+
+cd -P "$(dirname "$src")/../"
+
+##### This is the actual pre-commit hook. #####
+
+if [ -z "$VIRTUAL_ENV" ]; then
+    # If the developer has a virtual environment in the canonical location, then
+    # this will attempt to use that. Otherwise, assume the necessary things are
+    # installed globally.
+    declare -r venv='./venv/bin/activate'
+    if [ -f "$venv" ]; then
+      source "$venv"
+    fi
+fi
+
+make lint


git/pre-commit
```diff
@@ -0,0 +1,28 @@
...
+
+while [ -h "$src" ]; do
+  dir="$(cd -P "$( dirname "$src")" && pwd)"
+  src="$(readlink "$src")"
+  [[ "$src" != /* ]] && src="$dir/$src"
+done
+
+cd -P "$(dirname "$src")/../../"
+
+##### This is the actual pre-commit hook. #####
```
dachary: We should either explain the new requirements (pip install requirements/securedrop-development.txt) or we could just not run make lint if the requirements are not satisfied.

I'm in favor of making this pre-commit hook optional rather than mandatory, therefore leaning toward not doing anything if the requirements are not met.

heartsucker: If you don't want to run it, just don't copy it into `.git`. Or use `--no-verify`. :)

docs/development/contributor_guidelines.rst
```diff
@@ -33,18 +33,27 @@ Code Style
...
 We use code linters to keep a consistent code quality and style. These linters
-also run in CI and will produce build failures. To avoid this, we suggest running
-the linters locally as you work.
+also run in CI and will produce build failures. To avoid this, we have included
+a git pre-commit hook. You can install it with the following command run at the
+root of the repository:
+
+.. code:: sh
+
+    cp git/pre-commit .git/hooks/
```
conorsch: Is it folly to recommend a symlink here? The `cp` actual will not honor modifications to the pre-commit hook.

git/pre-commit
dachary: Oh, I wrongly assumed it would be installed by default! Sorry for the noise. I should have read the guidelines :-)

git/pre-commit
```diff
@@ -0,0 +1,28 @@
...
+
+# If the developer has a virtual environment in the canonical location, then
+# this will attempt to use that. Otherwise, assume the necessary things are
+# installed globally.
+declare -r venv='./venv/bin/activate'
+if [ -f "$venv" ]; then
+  source "$venv"
+fi
```
conorsch: I appreciate the automagic here, but for the record I do _not_ use that path for my development virtualenv. I _do_ use that path when testing inside Tails, for use with `securedrop-admin`. But on my host machine, I use `virtualenvwrapper`, so the actual virtualenv files are stored under `~/.virtualenvs`.

I recommend wrapping all of this virtualenv logic in an if conditional that skips it if the environment variable `VIRTUAL_ENV` is defined, which it will be if a virtualenv is already active.

dachary:
I did not test and wrongfully approved. Now I have no clue how to dismiss my review...

conorsch:
@dachary Got your back. :wink:

dachary:
@conorsch thanks! Did i make a mess with this review, sorry everyone...

git/pre-commit
```diff
@@ -0,0 +1,28 @@
...
+declare src="$0"
+declare dir=''
+
+while [ -h "$src" ]; do
+  dir="$(cd -P "$( dirname "$src")" && pwd)"
+  src="$(readlink "$src")"
+  [[ "$src" != /* ]] && src="$dir/$src"
+done
+
+cd -P "$(dirname "$src")/../../"
```
conorsch: Much of this path handling would be simplified by writing in Python, no? Not complaining, and not even requesting the change, just pointing it out.

docs/development/contributor_guidelines.rst
heartsucker: Good point.

git/pre-commit
```diff
@@ -0,0 +1,28 @@
...
+
+# If the developer has a virtual environment in the canonical location, then
+# this will attempt to use that. Otherwise, assume the necessary things are
+# installed globally.
+declare -r venv='./venv/bin/activate'
+if [ -f "$venv" ]; then
+  source "$venv"
+fi
+
+make lint
```
conorsch: Running the docs-lint task every time adds considerable delay here. Compare:

```
$ time make lint > /dev/null

real    0m11.934s
user    0m14.372s
sys     0m0.200s
$ time sh -c 'make flake8 ; make html-lint' > /dev/null

real    0m1.923s
user    0m4.516s
sys     0m0.096s
```

The extra ten seconds will add a ton of time when rebasing. Maybe skip the docs-lint step for interactive work and leave it in CI?

git/pre-commit
heartsucker: I added the logic about the env-var, but left the `venv` logic since it's still useful.

git/pre-commit
heartsucker: This doesn't run when rebasing, so I think that's not an issue.

heartsucker: In theory, we could actually do a parse of `git status --porcelain` and then only run the relevant tests, though that would quickly become a pain to maintain.

docs/development/contributor_guidelines.rst
```diff
@@ -33,18 +33,27 @@ Code Style
...
 We use code linters to keep a consistent code quality and style. These linters
-also run in CI and will produce build failures. To avoid this, we suggest running
-the linters locally as you work.
+also run in CI and will produce build failures. To avoid this, we have included
+a git pre-commit hook. You can install it with the following command run at the
+root of the repository:
+
+.. code:: sh
+
+    ln -sf git/pre-commit .git/hooks/pre-commit
```
conorsch: Symlink won't work with this syntax, since the target must be relative. Ideally this would work:

```
ln -sf ../../git/pre-commit .git/hooks/pre-commit
```

but then the Makefile target can't be found, likely due to the filepath madness in the while loop. So just change it back to a `cp` and I'll look into the specifics on a rainy day. Would much rather have a `cp`-able version available for everyone to use pronto.

heartsucker: I just pushed a new version. It work with a symlink and plays nicely with the paths.

git/pre-commit
```diff
@@ -0,0 +1,31 @@
...
+    # If the developer has a virtual environment in the canonical location, then
+    # this will attempt to use that. Otherwise, assume the necessary things are
+    # installed globally.
+    declare -r venv='./venv/bin/activate'
+    if [ -f "$venv" ]; then
+      source "$venv"
+    fi
+fi
+
+pwd
```
conorsch: Nit: remove the `pwd`? Looks like that's left over from testing. New version works for me as a symlink! (Also tested removing the `pwd` and committing again—still works.)

heartsucker: Done!

conorsch:
Works for me! Thanks very much, @heartsucker!

conorsch:
> A little concerned about the length of time to commit, but if it becomes annoying we can rip out docs linting at a later date.

Agreed. Maybe a pre-push hook to run the docs-linting. Let's start using it and find out!

PR #2240, opened 2017-09-01
tests: make test_manage.py run from any directory

### LINKED ISSUE #2239: Application test suite execution is not directory-agnostic
# Bug

## Description

Two of the application tests, related to translation subcommands for `manage.py`, are not directory-agnostic. They pass only when pytest is executed from the `securedrop/` directory, but not if the fullpath to the directory is provided. Only these two tests are fussy about the directory.

## Steps to Reproduce
To confirm the failure:

1. Provision development VM and SSH in.
2. Run `pwd` and confirm you're in `/home/vagrant`. 
2. Run `pytest -v /vagrant/securedrop/tests --page-layout`.
3. Observe that out of all 229 tests, 2 fail. 

To confirm the cause:
1. Provision development VM and SSH in.
2. Run `cd /vagrant/securedrop`.
3. Run `pytest -v tests --page-layout`.
3. Observe that all 229 tests pass.

## Expected Behavior

The application test suite passes regardless of where the "run the tests" command was executed.

## Actual Behavior

Two tests fail when called from any directory other than `/vagrant/securedrop`. The two tests in question are:

```
tests/test_manage.py::TestManage::test_translate_compile_code_and_template FAILED
tests/test_manage.py::TestManage::test_translate_compile_arg FAILED
```

These tests were added as part of #2120. 

## Comments

The requirement for a specific directory for test running is not a showstopper, but does indicate we've got a more brittle test suite than we should. Discovered it during Travis debugging for #2216. Until this issue is patched, I'll just preserve the current directory-based test invocation and include a comment warning about its requirement for now. 

Looks like some rather simple touch-ups to the two tests adding path canonicalization, e.g. via `os.path.abspath` and `__file__`, may resolve the problem.

dachary:
## Status

Ready for review

## Description of Changes

Fixes #2239

The application test suite passes regardless of where the "run the tests" command was executed.

## Testing

1. Provision development VM and SSH in.
2. Run `pwd` and confirm you're in `/home/vagrant`. 
2. Run `pytest -v /vagrant/securedrop/tests --page-layout`.
3. Observe that out of all 229 tests pass

## Deployment

test only

### DIFF
--- a/securedrop/tests/i18n/babel.cfg
+++ b/securedrop/tests/i18n/babel.cfg
@@ -1,4 +1,4 @@
-[python: tests/i18n/*.py]
+[python: **/tests/i18n/*.py]
 
-[jinja2: tests/i18n/*.html]
+[jinja2: **/tests/i18n/*.html]
 extensions=jinja2.ext.autoescape,jinja2.ext.with_
--- a/securedrop/tests/test_manage.py
+++ b/securedrop/tests/test_manage.py
@@ -2,6 +2,7 @@
 
 import argparse
 import os
+from os.path import abspath, dirname, exists, getmtime, join, realpath
 os.environ['SECUREDROP_ENV'] = 'test'  # noqa
 import config
 import logging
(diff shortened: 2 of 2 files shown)


conorsch:
Test run from $HOME with normal user privileges in development VM shows that coverage doesn't work:

```
Coverage.py warning: Module ../securedrop/ was never imported. (module-not-imported)
Coverage.py warning: No data was collected. (no-data-collected)


====================================================================== warnings summary =======================================================================
:171
  'pytest_runtest_makereport' hook uses deprecated __multicall__ argument

None
  pytest_funcarg__caplog: declaring fixtures using "pytest_funcarg__" prefix is deprecated and scheduled to be removed in pytest 4.0.  Please remove the prefix and use the @pytest.fixture decorator instead.
  pytest_funcarg__capturelog: declaring fixtures using "pytest_funcarg__" prefix is deprecated and scheduled to be removed in pytest 4.0.  Please remove the prefix and use the @pytest.fixture decorator instead.

========================================================== 229 passed, 3 warnings in 471.36 seconds ===========================================================
ERROR: Failed to generate report: No data to report.
```

That's a separate issue, though, so will file for later follow-up.

PR #2250, opened 2017-09-02
YubiKey [HOTP], secret

comradekingu:
In line with https://github.com/freedomofpress/securedrop/blob/develop/docs/yubikey_setup.rst


redshiftzero:
Hmm, checks aren't passing here - @comradekingu can you rebase this on current `develop`?

dachary:
noticed a few more instances where YubiKey is spelled Yubikey instead and added them to this commit. For test purposes this commit was moved to https://github.com/freedomofpress/securedrop/pull/2265

PR #2251, opened 2017-09-02
Log out

comradekingu:


dachary:
@redshiftzero if you agree with the change in wording I can help with the tests. What do you think ?

dachary:
See also https://github.com/freedomofpress/securedrop/pull/2357

redshiftzero:
Thanks @comradekingu and @dachary (:heart: for your offer - I was fixing up #2301 and figured I'd pull this commit in too), rebased these commits on top of i18n changes and added test changes in superseding PR #2358, let me know if you notice any issues! 😃

PR #2256, opened 2017-09-06
run srm with minimal dependencies

### LINKED ISSUE #844: Intermittent Travis test failures (Directory not empty: '/tmp/securedrop/keys')

dachary:
## Status

Work in progress

## Description of Changes

Fixes #844

When srm is delegated to a redis worker via rq, the modules it needs
will be imported in another python interpreter. To keep this to a
minimum, the srm function is isolated in a single file which only
depends on subprocess.

The former implementation was included in the store file which imports
gnupg. But gnupg runs processes and create directories when it is
imported. It will race with the journalist wsgi threads (each thread
has its own python interpreter) on a production server. And it will
also race with the journalist application run in a test environment,
inadvertendly re-creating the key directory after the teardown
function of the test removed the temporary directory in which it ran.

The redis server cannot be replaced by a thread run from the
journalist application because the production server has dozens of
them ran in each wsgi thread.

## Testing

* cherry-pick **reproducer for #844** to develop
* run  `while : ; do pytest -v tests/test_journalist.py || break ; done` in the development VM
* verify that it quickly (less than 5 runs) fails with
<pre>
>           assert not os.path.exists(config.SECUREDROP_DATA_ROOT)  # reproducer for #844
E           AssertionError
</pre>
* cherry-pick **isolate the srm function** on top of **reproducer for #844**
* run  `while : ; do pytest -v tests/test_journalist.py || break ; done` in the development VM
* verify that it loops with no failure during an hour

## Deployment

* The new rm.py file was added to apparmor for usr.sbin.apache2
* The code delegating the removal of files is not changed, it is just moved into a separate file which has 100% coverage.

### DIFF
--- a/securedrop/tests/utils/env.py
+++ b/securedrop/tests/utils/env.py
@@ -70,9 +70,5 @@ def teardown():
     try:
         shutil.rmtree(config.SECUREDROP_DATA_ROOT)
     except OSError as exc:
-        os.system("find " + config.SECUREDROP_DATA_ROOT)  # REMOVE ME, see #844
         if 'No such file or directory' not in exc:
             raise
-    except:
-        os.system("find " + config.SECUREDROP_DATA_ROOT)  # REMOVE ME, see #844
-        raise
--- a/securedrop/tests/utils/env.py
+++ b/securedrop/tests/utils/env.py
@@ -69,6 +69,9 @@ def teardown():
     db_session.remove()
     try:
         shutil.rmtree(config.SECUREDROP_DATA_ROOT)
+        import time
+        time.sleep(0.1)
+        assert not os.path.exists(config.SECUREDROP_DATA_ROOT)  # safeguard for #844
     except OSError as exc:
         if 'No such file or directory' not in exc:
             raise
--- a/securedrop/tests/utils/env.py
+++ b/securedrop/tests/utils/env.py
@@ -69,8 +69,6 @@ def teardown():
     db_session.remove()
     try:
         shutil.rmtree(config.SECUREDROP_DATA_ROOT)
-        import time
-        time.sleep(0.1)
         assert not os.path.exists(config.SECUREDROP_DATA_ROOT)  # safeguard for #844
     except OSError as exc:
         if 'No such file or directory' not in exc:
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
@@ -144,6 +144,8 @@
   /var/www/securedrop/journalist_templates/login.html r,
   /var/www/securedrop/request_that_secures_file_uploads.py r,
   /var/www/securedrop/request_that_secures_file_uploads.pyc rw,
+  /var/www/securedrop/rm.py r,
+  /var/www/securedrop/rm.pyc rw,
   /var/www/securedrop/secure_tempfile.py r,
   /var/www/securedrop/secure_tempfile.pyc rw,
   /var/www/securedrop/source.py r,
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -15,6 +15,7 @@ from sqlalchemy.sql.expression import false
 import config
 import version
 import crypto_util
+from rm import srm
 import store
 import template_filters
 from db import (db_session, Source, Journalist, Submission, Reply,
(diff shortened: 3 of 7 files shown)


redshiftzero:
It looks like you added a new file `path.py` (where the `path` and `verify` functions are) that was not added in d00e7159c3933377dc5afce5e20b2a940c628220?

dachary:
@redshiftzero I believe this is the right way to deal with the problem, as explained in https://github.com/freedomofpress/securedrop/issues/844#issuecomment-327433547

This draft implementation is split in two commits: one implementing & testing the thread implementation and the other trimming redis, workers and supervisor. I did this to make sure I was not missing a significant blocker.

I'll refrain from working on it until you get a chance to voice your opinion. Since the #844 does not happen very often, it should not be a blocker. However, for some reason it was a blocker yesterday on one of my PR for mysterious reasons. So.... it may be a good thing to address than in the not too distant future ;-)

redshiftzero:
Hey @dachary - so I haven't combed this over, but from taking a quick look at the diff here and reading your reasoning in [#844 (comment)](https://github.com/freedomofpress/securedrop/issues/844#issuecomment-327433547), I see what you mean about Redis being overkill. Given that this simplifies a clearly error prone area of the code, makes it easier to maintain, and it resolves the transient error (full disclosure: I haven't tested whether this does resolve the transient error yet) then I'm cool with your approach here so far. :+1:

dachary:
@redshiftzero thanks for the high level review. I'll work on it some more so it can be reviewed in detail. To be continued :-)

garrettr:
This looks great, but unfortunately I only have time to give it a quick skim. Just one note: whoever is reviewing/QA'ing this should examine the interaction between Python processes/threads as they are spawned by the web server (which, in turn, depends on both the web server and its specific configuration) and the threads that they in turn spawn as part of the operation of `Shredder`.

For example, if the web server spawns threads to serve requests, and each of these threads `.join`s on the thread running the `srm` process, and there is a maximum number of threads allowed (there usually is), a journalist who deletes a large number of files might end up accidentally triggering a DOS.

Also, make sure you test it on very large files—shred is significantly slower on larger files (e.g. close to the maximum submission size). 

Overall, I think this is a solid approach and agree that `redis` for this use case is overkill (I implemented that in a moment of desperation, IIRC). My main reason for doing so is that it maintains a persistent queue and handles submissions one at a time, so there is no risk of DOS—it might take a while, but as long as the server continues running, it will securely delete every deleted submission eventually.

dachary:
> if the web server spawns threads to serve requests, 

This is an excellent point ! Is there more than one flask journalist app instance at any given time ? Will wsgi spawn multiple journalist app ?

dachary:
@garrettr a system wide lock ensures there only is one srm loop at anytime https://github.com/freedomofpress/securedrop/pull/2256/commits/dc8d947905049c2999a7be262da5e682df11f8d5

That takes care of races between processes or concerns about involuntary DOS because of multiple srm.

If I'm not mistaken, there is one journalist app per wsgi thread. Since each of them create an additional shredder thread, we effectively double the number of threads. However, since only one loop is active at a time this does not significantly increase the CPU consumption: all other shredder threads will be blocked waiting on the exclusive file lock. And since there are at most 60 threads according to install_files/ansible-base/roles/app/templates/sites-available/journalist.conf

<pre>
WSGIDaemonProcess journalist processes=2 threads=30 ...
</pre>

there is no risk of overflowing system limits and the impact on performances should be negligible.

Am I missing something else ?

garrettr:
All I'm saying is: stress test it (e.g. delete > 60 large submissions at once) and profile it to make sure it doesn't do anything undesirable.

dachary:
@garrettr ack. I'm glad you caught that early, it would have been complicated to diagnose later on. 👍

dachary:
having the **while true ; do srm /var/lib/securedrop/shredder/* ; done** loop running as a daemon is a good idea. Using a thread launched from the journalist app is a **terrible** idea. Will rework to just replace redis with this simple daemon.

dachary:
Or ... we can just srm from the journalist app after returning the response to the user ? And use a global lock to make sure there is only one srm going on at any given time. This is a simple problem there should be a simple solution...

dachary:
The scope of this pull request far exceeded the need to fix a transient error that happens frequently enough to be a nuisance. It would have been good to trim redis in favor of a simpler solution, if there was one. Since it turns out this is more complicated than I thought, let's just fix the problem in a minimal way.

The test instructions were updated and the reproducer included in the PR for posterity. It is tricky enough that we don't want to forget about it.

securedrop/tests/utils/env.py
```diff
@@ -69,10 +69,7 @@ def teardown():
     db_session.remove()
     try:
         shutil.rmtree(config.SECUREDROP_DATA_ROOT)
+        assert not os.path.exists(config.SECUREDROP_DATA_ROOT)  # safeguard for #844
```
redshiftzero: :+1:

PR #2258, opened 2017-09-06
Run update_version.sh in development VM

### LINKED ISSUE #2257: Update update_version.sh for 0.4.3
## Description

We removed the build VM where the `update_version.sh` script expects to be run, so we'll need to make some edits to `update_version.sh`. 

We'll also need to make sure that `dch` is installed in the VM we use to run `update_version.sh`. 

Ideally one could run the `update_version.sh` locally since it only requires git and basic bash utilities, but `dch` (which is provided by the `devscripts` apt package) does not appear to be available packaged in homebrew for macOS, so it's simpler to run this in the development VM for now and add `devscripts` as a requirement.

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #2257.

 * Install `git` and `devscripts` in the development VM so that we can run `update_version.sh` there
 * Update warnings in `update_version.sh` to direct user to run the script in the development VM

## Testing

Provision development VM on this branch, attempt to run, should encounter no issues.

## Deployment

Release process only

### DIFF
--- a/install_files/ansible-base/group_vars/all/securedrop
+++ b/install_files/ansible-base/group_vars/all/securedrop
@@ -26,6 +26,8 @@ development_dependencies:
   - tmux
   - vim
   - libffi-dev # for sphinx
+  - git
+  - devscripts # for dch
 
 # These profiles are referenced by multiple machines, such as Application Server
 # for direct copying at install time, and the build machine for including them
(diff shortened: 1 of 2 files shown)


redshiftzero:
Excellent, thanks for that rapid review @conorsch!

redshiftzero:
Disabling coverage check here because 1) it's unnecessary since the code in this diff is not included in the coverage computation and 2) this slow check is currently blocking me. Filed #2261.

PR #2264, opened 2017-09-07
terminal, UID

comradekingu:


securedrop/source_templates/why-journalist-key.html
```diff
@@ -8,12 +8,12 @@
...
 <ul>
 <li>If you are using <a href='https://tails.boum.org/'>Tails</a>, you can double-click the <code>.asc</code> file you just downloaded and it will be automatically imported to your keyring.</li>
-<li>If you are using Mac/Linux, open the Terminal. You can import the key with <code>gpg --import /path/to/key.asc</code>.</li>
+<li>If you are using Mac/Linux, open the terminal. You can import the key with <code>gpg --import /path/to/key.asc</code>.</li>
 </ul>
 </li>
 <li>Encrypt your submission.
 <ol>
-<li>You will need to be able to identify the key (this is called the "user ID" or uid). Since the public key's filename is the key's fingerprint (with .asc at the end), you can just copy and paste that. (don't include the <code>.asc</code>!)</li>
+<li>You will need to be able to identify the key (this is called the "user ID" or UID). Since the public key's filename is the key's fingerprint (with .asc at the end), you can just copy and paste that. (don't include the <code>.asc</code>!)</li>
```
dachary: I'm not against using UID instead of uid on principle. Is there a reason for it ? When I list gpg keys, it is lower case but maybe that's bad practice ?
<pre>
pub   dsa1024 2001-02-03 [SCA]
      24F2720EC693FEA0A564F8DEF1D2CCC8497A176D
uid           [ultimate] Loic Dachary (OuoU) <loic@debian.org>
uid           [ultimate] Loic Dachary (OuoU) <loic@senga.org>
uid           [ultimate] Loic Dachary (OuoU) <loic@gnu.org>
uid           [ultimate] Loic Dachary <loic@dachary.org>
sub   elg1024 2001-02-03 [E]
</pre>

comradekingu: Then again "user ID" was never my understanding of it, so you have me confused.
`
username@localhost:~$ gpg --list-keys UniqueID smari@mailpile.is
pub   4096R/D13C70DA 2013-09-13
uid                  Smári McCarthy <smari@mailpile.is>
`
Following standard rules for abbreviation it should be UID either way.

dachary: Ok for uid.

dachary:
continued at https://github.com/freedomofpress/securedrop/pull/2279

Issue #2282, opened 2017-09-11
Pfsense null configuration XML tag format has changed


redshiftzero:
## Description 

It looks like there have been some changes to the XML null configuration tag format used by pfSense, now requiring `<tag></tag>` instead of the `<tag/>` format currently used in the templates that SecureDrop is shipping. 

Upstream issue: https://redmine.pfsense.org/issues/6893

## Steps to Reproduce

1. Download firewall templates.
2. Attempt to load.

## Expected Behavior

Template loads obviating the need for manual config.

## Actual Behavior

The template does not load. 

## Comments

This necessitates the templates to be regenerated using the (latest) updated format. 

Given the tight timeline until 0.4.3 is released, I suggest that we push the resolution of this into 0.4.4 to give sufficient time to generate and test the updated firewall templates. In the meantime, people installing SecureDrop will need to fall back to the manual configuration, and the documentation should indicate that manual configuration is necessary to avoid frustration in current users. 

@freddymartinez9 discovered this while testing #2281 on hardware.

freddymartinez9:
I realized yesterday that I am in the middle of a fresh install with the latest pfSense. When I have a SD install complete with a clean (factory reset device pfSense) I can do an export of their configurations, so we won't have to manually fix these tags.

freddymartinez9:
pushed this commit https://github.com/freedomofpress/securedrop/commit/ac4ce01382bd23a156140ba1d3700f7171292a77 but it requires a fresh pfSense firewall reset.  I will try to test this week

freddymartinez9:
I tested my commit https://github.com/freedomofpress/securedrop/commit/ac4ce01382bd23a156140ba1d3700f7171292a77 just now using a working SecureDrop install. Here is what I did:
 1) Manually install SecureDrop and the firewall
 2) Factory reset the pfSense firewall router
 3) Using the clearnet web browser (Unsafe browser) I went to my branch and downloaded the `.xml` files as "raw" from GitHub (**note**: this won't work using Tor due to the `clearnet` user chroot in Tails).
 4) Confirmed the factory reset was successful, no firewall rules were in the pfSense admin GUI. Also the admin user's password is reset. 
  5)  Follow the steps in our [docs](https://docs.securedrop.org/en/stable/network_firewall.html) up to "SecureDrop Configuration"
 6) Restore the `xml` files for each section.
 7) Reboot the firewall (important step)
 8) Test connectivity to the `.onion` using both Tor Browser and ssh.

[...]

PR #2283, opened 2017-09-11
Instruct users not to use XML templates for firewall config

### LINKED ISSUE #2277: The xml file aliases-config-pfSense.xml is missing starting tag.
# Bug

## Description

The xml file [aliases-config-pfSense.xml](https://github.com/freedomofpress/securedrop/blob/develop/install_files/network_firewall/aliases-config-pfSense.xml) is missing the starting tag.

The file starts with 
`aliases>` should start with `<aliases>`

redshiftzero:
## Status

Ready for review

## Description of Changes

 * Fixes #2277 (needed if we're going to use these XML files as a reference).
 * Adds documentation in light of #2282 to notify users that the templates exist purely as a reference. We can revert de527dd when #2282 is addressed.

## Testing

Review the small edits to the templates.

Review the docs.

## Deployment

None, docs only. 


redshiftzero:
Thanks for the rapid review @conorsch!

PR #2116, opened 2017-08-15
[WIP] i18n

dachary:
## Status

Work in progress

## Description of Changes

i18n of source and journalist interface. This pull request is for reference and is not meant to be merged as is. Commits are proposed as separated pulll requests. Note that the [order of the commits as presented by github is not what you would get with git log](https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/). When viewed with git log the commits are grouped for clarity.

## Active pull requests waiting to be merged

These are the pull requests created from various commits in this branch.There should not be more than a few otherwise it will become difficult to manage.

* https://github.com/freedomofpress/securedrop/pull/2343
* https://github.com/freedomofpress/securedrop/pull/2344
* https://github.com/freedomofpress/securedrop/pull/2345
* https://github.com/freedomofpress/securedrop/pull/2346
* https://github.com/freedomofpress/securedrop/pull/2347
* https://github.com/freedomofpress/securedrop/pull/2348
* https://github.com/freedomofpress/securedrop/pull/2349

PR #2301, opened 2017-09-14
Less !, ID, shorter

comradekingu:


dachary:
@comradekingu do you mind if we wait until the i18n of SecureDrop is merged in before working on these improvements ? The first batch of changes was extremely useful as it helped figure out and polish a good workflow for tiny changes. However this requires a fair amount of manual work on my part because it creates conflicts with the not-yet-merged part if of the [i18n commit series](https://github.com/freedomofpress/securedrop/pull/2116).

I realize this is just me being lazy here and if the relevant parts of i18n are not merged in two weeks from now, I'll do the needful regardless.

What do you think ?

comradekingu:
@dachary We do it your way. This is all bonus stuff anyway.
Could do a more in-depth set of changes, thus far it has been a bit superficial.
Let me know when you have the time to look into things.

redshiftzero:
Thanks @comradekingu and @dachary, rebased these commits on top of i18n changes and added test changes in superseding PR #2358, let me know if you notice any issues! 😃

PR #2306, opened 2017-09-14
i18n: load the locale depending on user preferences

dachary:
## Status

Ready for review

## Description of Changes

Set the locale of the source or journalist interface as follows, by
order of precedence:

  - l request argument or session['locale']
  - browser suggested locale via Accept-Languages
  - config.DEFAULT_LOCALE
  - 'en_US'

When the l request argument is specified, the locale is saved in the
session. If the l request argument is empty, the locale from the session
is cleared and the Accept-Languages header takes precedence again.

The locale requested with the l argument or via the Accept-Languages
header is matched against the available locales, as found in the
translations directory. The babel.core.negotiate_locale function is used
to find a match and it will, for instance, match:

   fr to fr_FR
   fr-FR to fr_FR
   zh-CN to zh_Hans_CN (script part)
   zh_Hans to zh_Hans

In order to improve the chances to find a match, the Accept-Languages
values are normalized via the core.Locale.parse function which knowns
about aliases, default script for Chinese etc.

## Testing

* pytest --page-layout tests/pages-layout and visually verify all pages look as they should
* pytest tests will run the additional tests that should could all modified lines
<pre>
modified   securedrop/tests/pytest.ini
@@ -1,4 +1,4 @@
 [pytest]
 testpaths = . functional
 usefixtures = setUptearDown
-addopts = --cov=../securedrop/
+addopts = --cov-report term-missing --cov=../securedrop/
</pre>
will help verify this is the case.

## Deployment

The code makes use of new config variables. It will do nothing and behave in a backward compatible way if they are not present. This should be verified by upgrading from a 0.4.3 fresh install.

### DIFF
--- a/securedrop/i18n.py
+++ b/securedrop/i18n.py
@@ -1,0 +1,74 @@
+#
+# SecureDrop whistleblower submission system
+# Copyright (C) 2017 Loic Dachary <loic@dachary.org>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+from flask import request, session
+from flask_babel import Babel
+from babel import core
+
+import config
+import os
+
+LOCALES = set(['en_US'])
+
+
+def setup_app(app):
+    babel = Babel(app)
+    assert len(list(babel.translation_directories)) == 1
+    for dirname in os.listdir(next(babel.translation_directories)):
+        if dirname != 'messages.pot':
+            LOCALES.add(dirname)
+
+    babel.localeselector(get_locale)
+
+
+def get_locale():
+    """
+    Get the locale as follows, by order of precedence:
+    - l request argument or session['locale']
+    - browser suggested locale, from the Accept-Languages header
+    - config.DEFAULT_LOCALE
+    - 'en_US'
+    """
+    locale = None
+    accept_languages = set()
+    for l in request.accept_languages.values():
+        if '-' in l:
+            sep = '-'
+        else:
+            sep = '_'
+        try:
+            accept_languages.add(str(core.Locale.parse(l, sep)))
+        except:
+            pass
+    if 'l' in request.args:
+        if len(request.args['l']) == 0:
+            if 'locale' in session:
+                del session['locale']
+            locale = core.negotiate_locale(accept_languages, LOCALES)
+        else:
+            locale = core.negotiate_locale([request.args['l']], LOCALES)
+            session['locale'] = locale
+    else:
+        if 'locale' in session:
+            locale = session['locale']
+        else:
+            locale = core.negotiate_locale(accept_languages, LOCALES)
+
+    if locale:
+        return locale
+    else:
+        return getattr(config, 'DEFAULT_LOCALE', 'en_US')
--- a/securedrop/tests/test_i18n.py
+++ b/securedrop/tests/test_i18n.py
@@ -1,0 +1,159 @@
+#
+# SecureDrop whistleblower submission system
+# Copyright (C) 2017 Loic Dachary <loic@dachary.org>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Affero General Public License for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+import argparse
+import logging
+import os
+
+from flask import request, session, render_template_string
+from flask_babel import gettext
+from werkzeug.datastructures import Headers
+
+os.environ['SECUREDROP_ENV'] = 'test'  # noqa
+import config
+import i18n
+import journalist
+import manage
+import source
+import version
+
+
+class TestI18N(object):
+
+    def verify_i18n(self, app):
+        not_translated = 'code hello i18n'
+        translated_fr = 'code bonjour'
+
+        for accepted in ('unknown', 'en_US'):
+            headers = Headers([('Accept-Language', accepted)])
+            with app.test_request_context(headers=headers):
+                assert not hasattr(request, 'babel_locale')
+                assert not_translated == gettext(not_translated)
+                assert hasattr(request, 'babel_locale')
+                assert render_template_string('''
+                {{ gettext('code hello i18n') }}
+                ''').strip() == not_translated
+
+        for lang in ('fr_FR', 'fr', 'fr-FR'):
+            headers = Headers([('Accept-Language', lang)])
+            with app.test_request_context(headers=headers):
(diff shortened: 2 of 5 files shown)


dachary:
@redshiftzero this is the last one before we can merge all gettext() wrapping of strings. It is more subtle than it looks because it was squashed and patched at least a dozen times over the past two months. The rest will be easier ;-)

dachary:
<pre>
i18n.py                                   38      0     18      0   100%
</pre>

The claimed decreased coverage is a mystery to me...

dachary:
https://circleci.com/gh/freedomofpress/securedrop/2046?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

<pre>
"Timeout when waiting for search string OpenSSH in ec2-52-53-233-88.us-west-1.compute.amazonaws.com:22"
</pre>
re-launching

securedrop/i18n.py
```diff
@@ -0,0 +1,76 @@
...
+import os
+
+LOCALES = set(['en_US'])
+babel = None
+
+
+def setup_app(app):
+    global babel
+    babel = Babel(app)
+    assert 1 == len(list(babel.translation_directories))
```
heartsucker: Pedantic: Yoda assertion (`expected == test_case` and not `test_case == expected`)

securedrop/i18n.py
```diff
@@ -0,0 +1,76 @@
...
+            sep = '-'
+        else:
+            sep = '_'
+        try:
+            accept_languages.add(str(core.Locale.parse(l, sep)))
+        except:
+            pass
+    if 'l' in request.args:
+        if len(request.args['l']) == 0:
+            if 'locale' in session:
```
heartsucker: I feel like we shouldn't pop this off the session until after we've gotten an acceptable alternative. Is there a reason you're popping it early? If so, can you put that in a comment.

securedrop/i18n.py
```diff
@@ -0,0 +1,76 @@
...
+
+import config
+import os
+
+LOCALES = set(['en_US'])
+babel = None
+
+
+def setup_app(app):
+    global babel
```
heartsucker: So looking at this, it looks like we never use the `babel` object outside this function, so to avoid us adding more global state, I think it's safe to not save the output of `Babel(app)` anywhere. If we have to save it later, we should attach it to the `app` object itself so that it's unique per app instance to help with future testing.

securedrop/tests/test_i18n.py
```diff
@@ -0,0 +1,148 @@
...
+os.environ['SECUREDROP_ENV'] = 'test'  # noqa
+import config
+import i18n
+import journalist
+import manage
+import source
+import version
+
+
+class TestI18N(object):
```
heartsucker: So, so approve of everything about these tests. :D

securedrop/i18n.py
dachary: Oh, nice ! It was needed in the beginning and then never questioned it again. Removed.

securedrop/i18n.py
dachary: The **1** is indeed what is expected. Am I missing something ?

securedrop/i18n.py
dachary: in which case do we not have an acceptable alternative ?

securedrop/i18n.py
heartsucker: I think traditionally it's written `some_operation_or_var == constant`.

heartsucker:
Checked the screenshots. Things look good.

securedrop/i18n.py
heartsucker: None, actually, but since this function doesn't also set the `session['locale']`, I feel like it shouldn't modify it. Also, neither the J nor S app sets the session locale. They only assign it to `g`.

dachary: >  since this function doesn't also set the session['locale']
but it does, in the else part of the if

securedrop/i18n.py
dachary: ah, ok, will change that

securedrop/tests/test_i18n.py
```diff
@@ -0,0 +1,148 @@
...
+
+        for app in (journalist.app, source.app):
+            app.config['BABEL_TRANSLATION_DIRECTORIES'] = config.TEMP_DIR
+            i18n.setup_app(app)
+            self.verify_i18n(app)
+
+    @classmethod
+    def teardown_class(cls):
+        reload(journalist)
+        reload(source)
```
redshiftzero: Thought: we could have a test to guard against forgetting that `DEFAULT_LOCALE` may not be set in `config.py`, something along the lines of:

```
def test_verify_default_locale_en_us_if_not_defined_in_config(self):
        DEFAULT_LOCALE = config.DEFAULT_LOCALE
        try:
            config.DEFAULT_LOCALE = None
            with app.test_client() as c:
                c.get('/')
                assert session.get('locale') == 'en_US'
                assert not_translated == gettext(not_translated)
        finally:
            config.DEFAULT_LOCALE = DEFAULT_LOCALE
```

Note: I have not tested this and it may need some modification, just jotting down the general idea

securedrop/tests/test_i18n.py
```diff
@@ -0,0 +1,148 @@
...
+            c.get('/?l=en_US', headers=Headers([('Accept-Language', 'fr_FR')]))
+            assert session.get('locale') == 'en_US'
+            assert not_translated == gettext(not_translated)
+
+            c.get('/', headers=Headers([('Accept-Language', 'fr_FR')]))
+            assert session.get('locale') == 'en_US'
+            assert not_translated == gettext(not_translated)
+
+            c.get('/?l=', headers=Headers([('Accept-Language', 'fr_FR')]))
+            assert session.get('locale') is None
```
redshiftzero: Why would we want `locale` to be `None` and not `en_US` here?

dachary: Because the session is not always set. It is set when **l=XXX** and unset when **l=**

securedrop/tests/test_i18n.py
dachary: Good idea ! To be perfectly honest I thought about adding a similar test and conveniently forgot about it ... Your scaffold was useful and adapted. Repushed !

dachary:
@heartsucker is there anything else you need modified ?

heartsucker:
@dachary 

The docs say:

```python
"""
Get the locale as follows, by order of precedence:
- l request argument
- browser suggested locale, from the Accept-Languages header
- config.DEFAULT_LOCALE
- app.config['BABEL_DEFAULT_LOCALE']
"""
```

But the value is also pulled from the session in some cases. The docs should be updated to reflect that (and possibly a test to ensure a request with a session containing a locale retains that value on the next request.

dachary:
> But the value is also pulled from the session in some cases. The docs should be updated to reflect that 

Good catch ! I updated as follows:

    - l request argument or session['locale']
    - browser suggested locale, from the Accept-Languages header
    - config.DEFAULT_LOCALE
    - 'en_US'

with both l and session on the same line since they have equal precedence because they really are the same thing. And changed

   - app.config['BABEL_DEFAULT_LOCALE']

to 

    - 'en_US'

The former was true in a previous version of the implementation but it was decided to move to something explicit and easier to control from the user / developer perspective.

> (and possibly a test to ensure a request with a session containing a locale retains that value on the next request.

I believe the current tests verify that, in particular:

<pre>
            c.get('/?l=fr_FR', headers=Headers([('Accept-Language', 'en_US')]))
            assert session.get('locale') == 'fr_FR'
            assert translated_fr == gettext(not_translated)

            c.get('/', headers=Headers([('Accept-Language', 'en_US')]))
            assert session.get('locale') == 'fr_FR'
            assert translated_fr == gettext(not_translated)
</pre>

heartsucker:
> with both l and session on the same line since they have equal precedence because they really are the same thing.

But they're not, right? Someone could start a session with `de-DE` and switch to `fr-FR` via the `l` query parameter. Session should be below `l`, no?

Anyway, that's me being pedantic. I'm marking this as accepted since this change is minimal.

dachary:
> But they're not, right? Someone could start a session with de-DE and switch to fr-FR via the l query parameter. Session should be below l, no?

Well yes, in a way your right because the session cannot override l while the other is true. But my thinking was that l will always be the same as session['locale']. There are no situation where l and session['locale'] are different. A contrario the header can be different from the session and the session will have precedence. 

But I agree with you, it is a fine point and thanks for agreeing to something that is debatable :-)

PR #2308, opened 2017-09-14
i18n gettext sources

dachary:
## Status

Ready for review

## Description of Changes

Add gettext() around translatable strings. And move all javascript strings to a template for translation purposes.

## Testing

* pytest --page-layout test/pages and verify visually that all pages look as they should.

## Deployment

It is a runtime change only and will not change how deployment / upgrades are done.


dachary:
This should be a straighforward suite of commits. With the notable exception of the **i18n: hack javascript with templates**. If you think it is too much of a hack I'll research alternatives.

securedrop/template_filters.py
```diff
@@ -9,7 +10,7 @@ def datetimeformat(dt, fmt=None, relative=False):
     if relative:
         time_difference = _relative_timestamp(dt)
         if time_difference:
-            return '{} ago'.format(time_difference)
+            return gettext('{time} ago').format(time=time_difference)
```
heartsucker: A problem I see with this is that some languages might not say "$time ago" and that the values may be ordered a different way. This should duplicate the "ago" in all the strings. Also, this will help the translators.

securedrop/i18n.py
```diff
@@ -0,0 +1,76 @@
...
+
+import config
+import os
+
+LOCALES = set(['en_US'])
+babel = None
+
+
+def setup_app(app):
+    global babel
```
heartsucker: As noted in the other PR, this does not need to be global.

dachary:
> Additional note: Are we not translating manage.py?

No, we're assuming it is only used by admins who would presumably be lost anyway if they could not read that much english.

securedrop/journalist.py
```diff
@@ -400,24 +418,27 @@ def _set_diceware_password(user, password):
...
+            'error')
         app.logger.error('Failed to update a valid password.')
         return
 
     # using Markup so the HTML isn't escaped
-    flash(Markup("<p>The password was successfully updated! Don't forget to "
-                 'save it in your KeePassX database. The new password is: '
-                 '<span><code>{}</code></span></p>'.format(password)),
-          'success')
+    flash(Markup("<p>" + gettext(
```
heartsucker: This should be: 
```python
    flash(Markup(("<p>" + gettext(
        "The password was successfully updated! Don't forget to "
        "save it in your KeePassX database. The new password is: " 
        ' <span><code>{password}</code></span>') + '</p>').format(passworkd=password)),
        'success')
```

Because the HTML got unevenly split which could be confusing. Also, some languages don't use the colon that way and it would be unnatural to force them to.

securedrop/template_filters.py
dachary: It is indeed the case in French (il y a 2 jours => 2 days ago) but it is consistent and this factorization did not cause problem. I propose we wait until someone complains that it is causing problem (arabic comes to mind as the rules are significantly different).

dachary: and Dutch + Norwegian are also ok with it it seems.

securedrop/i18n.py
dachary: yes, this commit will be discarded when the other PR is merged in

securedrop/journalist.py
dachary: Note the gettext closing parenthesis is right after **password is:** so the string does not contain any html tag.

securedrop/template_filters.py
heartsucker: In German it's "vor 2 Tagen" for "2 days ago," so that's already one.

securedrop/journalist.py
heartsucker: Yeah, you're right, but my point still stands about about the colon. I guess we can leave it for now since I'm pretty sure non-English speakers are used to some oddities when using translated sites, should that be a problem.

securedrop/i18n.py
heartsucker: 👍

heartsucker:
> This should be a straighforward suite of commits. With the notable exception of the i18n: hack javascript with templates. If you think it is too much of a hack I'll research alternatives.

Just as a note to other reviewers, the reason I did this originally was to avoid us having to have multiple JS assets that we'd have to transmit. We could do something like have the `journalist.js` have the locale suffix like `journalist-fr-FR.js` and then do some logic in the template to avoid having to retransmit the strings every time using that `include` jinja statement.

securedrop/journalist.py
dachary: hum, you're right. It looks like chinese does not have a problem with that. Do you know of a language that does ? Maybe arabic  (@redshiftzero do you know ?). If you don't mind I'll leave it this way unless we have evidence it's wrong. I'm being lazy (of course, always am ;-). But also it means translators in Dutch, Chinese and Norwegian will have to update their translation if we change that (these are the three languages with active translators at the moment, not counting French of course). I'd like to make sure this is actually useful before we do the change and ask them that kind of effort.

heartsucker: I can't think of one. Let's leave it for now.

heartsucker:
Replying to myself: It seems https://github.com/python-babel/flask-babel/issues/108 and https://github.com/python-babel/babel/issues/444 were related originally. Since they've been addressed (or so it seems), I think we actually *should* localize the JS properly to minimize the amount of data we're sending and also to get rid of the JS/HTML hack I used used and @dachary copied.

dachary:
> I think we actually should localize the JS properly

I must confess my ignorance about how that could be done. I understand the **assets** in jinja are a way to pre-process CSS/JS in the same way you could with a makfile minifying JS or collating CSS. Only you stay in the jinja universe and do not need to learn new tools.

I also understand that babel has to know about the **assets** extension (http://flask-assets.readthedocs.io/en/latest/#babel-configuration) otherwise it will fail to parse the jinja templates properly.

I do not however understand how we could combine jinja and babel in the context of a JS file. @heartsucker it would be great if you could point me in the right direction :-)

heartsucker:
@dachary I'll take a look at that one later.

Also, you have a split sentence "Click the" and "Tor icon in the toolbar above" need to be a single string.

securedrop/template_filters.py
```diff
@@ -22,19 +23,20 @@ def _relative_timestamp(dt):
...
     if diff < 45:
-        return '{} second{}'.format(int(diff), '' if int(diff) == 1 else 's')
+        return ngettext('{n} second', '{n} seconds', int(diff)).format(
+            n=int(diff))
     elif diff < 90:
-        return 'a minute'
+        return gettext('a minute')
     elif diff < 2700:
-        return '{} minutes'.format(int(max(diff / 60, 2)))
+        return gettext('{n} minutes').format(n=int(max(diff / 60, 2)))
```
redshiftzero: For languages that have a dual form (e.g. Arabic), we'll need to do something else here for the "2 minute" case. But let's do this in a followup.

securedrop/journalist.py
redshiftzero: Not sure how Arabic handles colons tbqh - agree with leaving as is until we find a language that motivates changes here

heartsucker:
Approved because the JS hack is totally good enough and it's better to have marginally slower page loads *but* also have i18n than to worry ourselves with this too much and delay the feature.

heartsucker:
Also, all the small things I mentioned can be fixed in other issues.

dachary:
\o/ here we are !

PR #2281, opened 2017-09-11
Fix pfSense firewall templates

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #2277. 

Changes proposed in this pull request:
 * Fixes typo in firewall template (missing `<` in starting `<aliases>` tag)
 * Fixes incorrect subnet mask on one interface (I verified the other subnet masks are correctly set as `/24`)

## Testing

In an ideal world of infinite QA time, you'd factory reset a pfSense firewall and test out these templates are working.

Barring that, at least spot check the diff and make sure these changes seem reasonable. 

## Deployment

Used on first time install only

## Checklist

Firewall only, no tests

PR #2317, opened 2017-09-15
Docs firewallxml

b-meson:
## Status

Ready for review

## Description of Changes

Fixes #2281 

## Testing

 1) Manually install SecureDrop and the firewall
 2) Factory reset the pfSense firewall router
 3) Using the clearnet web browser (Unsafe browser) I went to my branch and downloaded the `.xml` files as "raw" from GitHub (note: this won't work using Tor due to the `clearnet` user chroot in Tails).
Confirmed the factory reset was successful, no firewall rules were in the pfSense admin GUI. Also the admin user's password is reset.
 4) Follow the steps in our docs up to "SecureDrop Configuration"
 5) Restore the `xml` files for each section.
 6) Reboot the firewall (important step)
 7) Test connectivity to the .onion using both Tor Browser and ssh.

## Deployment

None


freddymartinez9:
image
It was squashed during a rebase but included in the commit messages @redshiftzero. It should correspond to [this line](https://github.com/freedomofpress/securedrop/pull/2317/files#diff-3b769446231031fdf3ff2bc2c0524a86R332) but if I'm missing something, please let me know.

docs/network_firewall.rst
```diff
@@ -465,20 +469,38 @@ to add a rule.
...
+First navigate to **Diagnostics ▸ Backup & Restore**:
 
 As an alternative to the provided screenshots, you can examine the provided
-``.xml`` templates as a reference:
+``.xml`` templates as a reference. However, because of the Tails chroot,
+you can not download the ``.xml`` files directly. Using the Unsecure Browser,
+download the files from `GitHub <https://github.com/freedomofpress/securedrop/tree/develop/install_files/network_firewall/>`__
+as "Raw" and save them as plaintext documents.
+
+|Backup & Restore|
```
redshiftzero: This "Backup and Restore" screenshot should be right after "First navigate to **Diagnostics ▸ Backup & Restore**:" right?

docs/network_firewall.rst
```diff
@@ -465,20 +469,38 @@ to add a rule.
...
+SecureDrop Configuration (Templates)
+------------------------------------
+
+Alternatively, you can load the provided ``.xml`` templates.
+
+First navigate to **Diagnostics ▸ Backup & Restore**:
 
 As an alternative to the provided screenshots, you can examine the provided
-``.xml`` templates as a reference:
+``.xml`` templates as a reference. However, because of the Tails chroot,
```
redshiftzero: This "templates as a reference" text should be removed, since these templates are no longer to be used as reference.

redshiftzero:
Friendly note that you should avoid squashing other people's commits into your own in the future

redshiftzero:
This closes #2282

freddymartinez9:
changes made and rebased on top of latest `develop`

docs/network_firewall.rst
```diff
@@ -465,20 +469,37 @@ to add a rule.
 Once you've set up the firewall, exit the Unsafe Browser, and continue
 with the "Keeping pfSense up to date" section below.
 
-Configuration Reference Templates
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+SecureDrop Configuration (Templates)
+------------------------------------
+
+As an alternative to the manually provided screenshots, you can use our ``.xml``
+templates. However, because of the Tails chroot, you can not download the
```
redshiftzero: Is it explained what the `Tails chroot` is? We should briefly explain (without assuming the reader is highly technical :wink:).

docs/network_firewall.rst
```diff
@@ -465,20 +469,37 @@ to add a rule.
...
+``.xml`` files directly. Using the Unsecure Browser, download the files from
```
redshiftzero: `Unsecure Browser` -> [`Unsafe Browser`](https://tails.boum.org/contribute/design/Unsafe_Browser/)

docs/network_firewall.rst
conorsch: Agreed, let's elide "chroot" altogether. We can simply provide instructions to use the Unsafe Browser. It may be reasonable to provide a `curl` command (coupled with `torify` if necessary).

docs/network_firewall.rst
```diff
@@ -465,20 +469,37 @@ to add a rule.
...
+`GitHub <https://github.com/freedomofpress/securedrop/tree/develop/install_files/network_firewall/>`__
+as "Raw" and save them as plaintext documents.
```
conorsch: Wait, why are we asking folks to download files from the develop branch? These files should be local on disk already—and have their integrity validated with the signed git tag from the latest release.

freddymartinez9:
We discussed this issue a bit further and it requires additional testing inside Tails admin drive. Getting the files from the verified checkout (which is GPG signed) is preferred but it is unclear how much latitude we have to transfer files from `amnesia` user and the `clearnet` user. Removing the milestone and we will revisit after the next release

conorsch:
As great as it is to have a template and upload it, we cannot instruct Admins to upload unverified sources. Given the Tails chroot strategy, our hands are tied on what we can do with the Unsafe Browser—and apparently "upload from Persistence" is not on that least.

I vote we close this issue, since these changes will never get merged in their current state. We can continue discussion in a separate issue.

msheiny:
> Given the Tails chroot strategy, our hands are tied on what we can do with the Unsafe Browser—and apparently "upload from Persistence" is not on that least.

This part is inaccurate - granted it is annoying for an admin to get files in there, we'd want to have a script do it. But I was able to drop files into `/var/lib/unsafe-browser/chroot/home/clearnet` and read them in the unsafe-browser

Here's a screenshot

image

conorsch:
@msheiny That is fantastic news! So we could add a task to `tailsconfig` that drops the firewall template into the correct directory. Then it's available _if_ an Admin wants to use it, and will be cleaned up on shutdown, same as the other ephemeral files. 

@b-meson Can you take a stab at implementing the above approach? Happy to re-review if you can get that working!

b-meson:
This looks like a good approach, I will give it a go. Thanks @msheiny for taking a look at this. 

@conorsch we should do this as a separate `./securedrop-admin` command. The `setup` command and `tailsconfig` are not ideal candiates: the former is done after the Ubuntu servers are installed and the latter is after the SecureDrop playbooks run. I'll look into adding a new command...

conorsch:
@b-meson Good call. A new subcommand for `firewall-template` or similar would indeed be ideal.

b-meson:
I tried to rename the `docs-firewallxml` branch and rename it in order to trigger tests, since I am now pulling in changes in via ansible but it was deleted. No worries, I will continue working the `firewallansible` branch and reopen this PR when it is ready.

PR #2320, opened 2017-09-15
Autocomplete local_servers aliases for pfSense

freddymartinez9:
## Status

Ready for review

## Description of Changes

Fixes autocomplete for an alias. Since we didn't have `mon` above the `local_servers`, autocomplete was failing in the pfSense GUI. This is a small fix. 

### DIFF
--- a/docs/network_firewall.rst
+++ b/docs/network_firewall.rst
@@ -416,8 +416,8 @@ Make aliases for the following:
 - ``admin_workstation``: ``10.20.1.2``
 - ``app_server``: ``10.20.2.2``
 - ``external_dns_servers``: ``8.8.8.8, 8.8.4.4``
-- ``local_servers``: ``app_server, monitor_server``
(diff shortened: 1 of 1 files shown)


redshiftzero:
Hey @freddymartinez9 - I'm not sure I understand this bug. Can you first make an issue (or link to one if you already made one) and then explain here how modifying this documentation fixes it?

freddymartinez9:
If you take a look at how we add aliases (screenshot here) we can add aliases by IP or by _other aliases_. The `local_servers` is an alias that is includes `mon` and `app`. 
 
add_firewall_alias

So as you are adding aliases, the pfSense GUI helpfully auto-completes aliases that you have already saved. So the order matters in that case. By moving the `mon` server above the `local_servers`, it will autocomple in the GUI. 

Another note: the pfSense GUI sorts the saved aliases alphabetically so there is no need to change our screenshots.

PR #2325, opened 2017-09-16
journalist: unify case of relative datetime

dachary:
## Status

Ready for review

## Description of Changes

The relative datetime displayed on the journalist administration page
is all in lowercase except the string Never.

## Testing

* login as a journalist with admin rights
* click on **Admin**
* create a new user
* verify it shows **never** instead of **Never**

## Deployment

No code change, no deployment troubles ahead.


dachary:
thanks for being flexible here :-)

PR #2326, opened 2017-09-17
Tests for manage.py

heartsucker:
## Status

Work in progress

## Description of Changes

Mostly addresses #1982. This doesn't fully close it due to some issues listed below. Fully closing this issue would require a major change to how we do tests.

Adds unit tests to some of the `manage.py` subcommands.

### `manage.py clean-temp` requires `sudo`

Otherwise it can't list all the processes' open files and errors out. Running it with `sudo` breaks because we change the `config.TEMP_DIR` value at the start of the tests. Because we have to shell out, we'd have to make `manage.py` accept a path value that allows us to dynamically load modules so we can write our own `config.py` from the original, change the values, write it to `/tmp` and then pass it back to `manage.py`.

### `db_session` has odd caching

There are errors when attempting to verify that `manage.py reset` purges the database errors with the following steps:

1. Insert `Journalist`
2. Run `manage.reset(args=None)`
3. Expect inserted Journalist to not be present under a new `Journalist.query`

However, the journalist is present. If instead I do:

```python
j = Journalist(**kwargs)
db_session.add(j)
db_session.commit()
manage.reset(args=None)
Jornalist.query.filter(username=j.username).one()
```

It raises an exception saying that row was deleted when I attempt to access the `j.username` field.

Thus, it both caches the result and acknowledges that the DB was deleted at the same time. This has been... frustrating.

## Testing

```bash
vagrant ssh
cd /vagrant/securedrop/
pytest -v tests/
```

### DIFF
--- a/securedrop/tests/test_manage.py
+++ b/securedrop/tests/test_manage.py
@@ -82,6 +82,40 @@ class TestManagementCommand(unittest.TestCase):
         self.assertIn('ERROR: That username is already taken!',
                       sys.stdout.getvalue())
 
+    @mock.patch("manage._get_username", return_value='test-user-56789')
+    @mock.patch("manage._get_yubikey_usage", return_value=False)
+    @mock.patch("manage._get_username_to_delete",
+                return_value='test-user-56789')
+    @mock.patch('manage._get_delete_confirmation', return_value=True)
+    def test_delete_user(self,
+                         mock_username,
+                         mock_yubikey,
+                         mock_user_to_delete,
+                         mock_user_del_confirm):
+        return_value = manage._add_user()
+        self.assertEqual(return_value, 0)
+
+        return_value = manage.delete_user(args=None)
+        self.assertEqual(return_value, 0)
+
+    @mock.patch("manage._get_username_to_delete",
+                return_value='does-not-exist')
+    @mock.patch('manage._get_delete_confirmation', return_value=True)
+    @mock.patch("sys.stdout", new_callable=StringIO)
+    def test_delete_non_existent_user(self,
+                                      mock_user_to_delete,
+                                      mock_user_del_confirm,
+                                      mock_stdout):
+        return_value = manage.delete_user(args=None)
+        self.assertEqual(return_value, 0)
+        self.assertIn('ERROR: That user was not found!',
+                      sys.stdout.getvalue())
+
+    @mock.patch("__builtin__.raw_input", return_value='test-user-12345')
+    def test_get_username_to_delete(self, mock_username):
+        return_value = manage._get_username_to_delete()
+        self.assertEqual(return_value, 'test-user-12345')
+
 
 class TestManage(object):
 
--- a/securedrop/tests/test_manage.py
+++ b/securedrop/tests/test_manage.py
@@ -116,6 +116,12 @@ class TestManagementCommand(unittest.TestCase):
         return_value = manage._get_username_to_delete()
         self.assertEqual(return_value, 'test-user-12345')
 
+    def test_reset(self):
+        return_value = manage.reset(args=None)
+        self.assertEqual(return_value, 0)
+        assert os.path.exists(config.DATABASE_FILE)
+        assert os.path.exists(config.STORE_DIR)
+
 
 class TestManage(object):
 
--- a/securedrop/manage.py
+++ b/securedrop/manage.py
@@ -62,7 +62,7 @@ def sh(command, input=None):
     return "".join(lines_of_command_output)
 
 
-def reset(args):  # pragma: no cover
+def reset(args):
     """Clears the SecureDrop development applications' state, restoring them to
     the way they were immediately after running `setup_dev.sh`. This command:
     1. Erases the development sqlite database file.
--- a/securedrop/manage.py
+++ b/securedrop/manage.py
@@ -192,10 +192,23 @@ def _add_user(is_admin=False):
         return 0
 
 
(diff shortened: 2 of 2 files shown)


dachary:
> manage.py clean-temp requires sudo

Note that https://github.com/freedomofpress/securedrop/pull/2121 resolves that and also tests the function. It is ready for review and the only reason it did not receive more attention is because the race it resolves is really rare. If you feel like reviewing it ... that's definitely an interesting one ;-)

securedrop/tests/test_manage.py
```diff
@@ -82,6 +82,46 @@ def test_exception_handling_when_duplicate_username(self,
...
+    @mock.patch("__builtin__.raw_input", return_value='test-user-12345')
+    def test_get_username_to_delete(self, mock_username):
+        return_value = manage._get_username_to_delete()
+        self.assertEqual(return_value, 'test-user-12345')
+
+    def test_reset(self):
+        return_value = manage.reset(args=None)
+        self.assertEqual(return_value, 0)
+        assert os.path.exists(config.DATABASE_FILE)
+        assert os.path.exists(config.STORE_DIR)
```
redshiftzero: These asserts would pass even if reset _didn't_ reset the database

redshiftzero: Ah ok just read the issues you described in the PR description. What about adding asserts that do the following:

```
>>> j = db.Journalist('myjournalist', 'correct horse battery staple generic passphrase hooray', False)
>>> db.db_session.add(j)
>>> db.db_session.commit()
>>> username_that_should_be_gone = j.username
>>> manage.reset(args=None)
0
>>> db.db_session.remove()  # Close that session and get a new one
>>> db.Journalist.query.filter_by(username=username_that_should_be_gone).one()
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/orm/query.py", line 2820, in one
    raise orm_exc.NoResultFound("No row was found for one()")
sqlalchemy.orm.exc.NoResultFound: No row was found for one()
``` 

You do have to save the username in `username_that_should_be_gone` because otherwise when you try to access the object `j`, SQLalchemy will first try to refresh this object in the database, will find see that there is no row for that primary key, and will throw `ObjectDeletedError`.

heartsucker: > These asserts would pass even if reset didn't reset the database

Yes, but part of the reset is to recreate those. I should have put a comment saying "ensure these exist after the reset."

> You do have to save the username in username_that_should_be_gone because...

Yes, I was doing that and then querying again but somehow the driver appeared to be caching the result because even with the DB gone, it was saying that the row still existed. I'll see if I can reproduce.

heartsucker: I'll see if I can debug this further, but it might be sparse test coverage here. Which is ok because an admin should never use this in production, so if it's a bit wonky, we can get by with occasional manual changes IMO.

heartsucker: Ok, really, I have no idea what's happening here.

cmd: `pytest tests/test_manage.py::TestManagementCommand::test_reset ; sqlite3 /var/lib/securedrop/db.sqlite 'select * from journalists;'`

```
self = <tests.test_manage.TestManagementCommand testMethod=test_reset>

    def test_reset(self):
        username = crypto_util.genrandomid()
        journo = Journalist(username=username,
                            password=crypto_util.genrandomid())
        db_session.add(journo)
        db_session.commit()
    
        # ensure the user was created
        Journalist.query.filter(Journalist.username == username).one()
    
        return_value = manage.reset(args=None)
        self.assertEqual(return_value, 0)
    
        # ensure the user was deleted
>       assert (Journalist.query.filter(Journalist.username == username)
                .one_or_none()) is None
E       AssertionError: assert <Journalist proofing fiscally cycle vanilla aflutter freezable unread tissue> is None
E        +  where <Journalist proofing fiscally cycle vanilla aflutter freezable unread tissue> = <bound method Query.one_or_none of <sqlalchemy.orm.query.Query object at 0x7f98593a6a90>>()
E        +    where <bound method Query.one_or_none of <sqlalchemy.orm.query.Query object at 0x7f98593a6a90>> = <sqlalchemy.orm.query.Query object at 0x7f98593a6a90>.one_or_none
E        +      where <sqlalchemy.orm.query.Query object at 0x7f98593a6a90> = <bound method Query.filter of <sqlalchemy.orm.query.Query object at 0x7f985938b1d0>>(<sqlalchemy.orm.attributes.InstrumentedAttribute object at 0x7f985a575b30> == 'proofing fiscally cycle vanilla aflutter freezable unread tissue')
E        +        where <bound method Query.filter of <sqlalchemy.orm.query.Query object at 0x7f985938b1d0>> = <sqlalchemy.orm.query.Query object at 0x7f985938b1d0>.filter
E        +          where <sqlalchemy.orm.query.Query object at 0x7f985938b1d0> = Journalist.query
E        +        and   <sqlalchemy.orm.attributes.InstrumentedAttribute object at 0x7f985a575b30> = Journalist.username

tests/test_manage.py:135: AssertionError
```

The `sqlite3` command outputs nothing after the test. wat.

heartsucker: Also, reloading the `db` module, sleeping, deleting the `journo` object all have no effect on this. 😭

redshiftzero: :sob: -----> :smile: 

check out #2359 - it's a PR into this branch doing what I described in the comment above. let me know what ya think

Issue #2098, opened 2017-08-10
Docs: Say that rerunning ./securedrop-admin install is safe


redshiftzero:
## Description

We should remind people in the docs (and maybe in the output of the script) that they can repeatedly `./securedrop-admin install` as much as they want without breaking anything. If they get a transient error, they should feel comfortable to fire it off again.

conorsch:
This situation improved vastly as of 0.4, so definitely appropriate to state explicitly that rerunning is not going to cause things to blow up. We should probably resolve #1704 first, though.

edenemmanuel:
@conorsch should we consider #1704 a blocker for #2098? It looks like that'll be resolved soon anyway, rendering my question moot.

conorsch:
> should we consider #1704 a blocker for #2098?

No, PRs welcome to close #2098.

> It looks like that'll be resolved soon anyway, rendering my question moot.

Aye! :wink:

Issue #2266, opened 2017-09-07
sdconfig fingerprint should allow white space separated form


dachary:
# Feature request

## Description

The `securedrop-admin sdconfig` input for the `securedrop_app_gpg_fingerprint` and `ossec_gpg_fpr` should ignore whitespace. 

## User Stories

When installing securedrop I want to copy/paste the output of `gpg --list-key` without manually removing the white space when answering the `securedrop-admin sdconfig` prompt for `securedrop_app_gpg_fingerprint` and `ossec_gpg_fpr`

conorsch:
Hmm, not a bad idea. We do already check for whitespace and fail if it's found: https://github.com/freedomofpress/securedrop/blob/afa97acb49f2d0f95208ff6f1c456718ae5e681d/install_files/ansible-base/roles/validate/tasks/validate_gpg_info.yml#L10-L11 but I see what you're saying, we could just strip it. Not opposed.

conorsch:
#2267 was solid progress here, but was inadequately QA'd in Tails, and doesn't work as advertised. For one, the inner quotes should be single quotes, not double quotes, otherwise Ansible throws an error:

![tails-quotes-in-fingerprint-whitespace-stripping](https://user-images.githubusercontent.com/657862/30355670-a4ce9726-97e9-11e7-9999-4f191434ef56.png)

Notably `yamllint` catches that error:

```
yamllint install_files/ansible-base/securedrop-configure.yml -d relaxed
install_files/ansible-base/securedrop-configure.yml
  60:81     warning  line too long (84 > 80 characters)  (line-length)
  65:81     warning  line too long (84 > 80 characters)  (line-length)
  159:67    error    syntax error: expected <block end>, but found '<scalar>'
```

so I'll get a PR up for #2255 pronto. Even with the quoting fixed, however, the associated validation tasks still fail if whitespace is provided in the fingerprint. Looks like we'll need to import the vars if writing to the local site-specific vars file reported changes, to make sure we have the var value _after_ the whitespace has been stripped. (Loosely related: #2095.)

kushaldas:
This one is fixed in https://github.com/freedomofpress/securedrop/commit/b7b6ca2e77d6ff8ae82ac87abaadc12896940ee8#diff-a3b5e93ae26d81aad5d29a02b86500deR201 Closing the ticket.

dachary:
@kushaldas thanks !

PR #2337, opened 2017-09-19
document how to rerun Install playbooks if needed

freddymartinez9:
## Status

Ready for review 

## Description of Changes

When an ansible playbook fails to run (for example if you have whitespace in the fingerprint, see #2266) there are no clear instructions for what an admin can do to rerun the playbook. If it is a minor issue, like a typo, the documentation should point to a way to fix these issues. This is one of the cases that I found while doing a new SD install last week and if I had to look it up as a person familiar with SecureDrop, then we should document it. 

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -70,7 +70,9 @@ match your environment: ::
 The script will automatically validate the answers you provided, and display
 error messages if any problems were detected. The answers you provided will be
 written to the file ``install_files/ansible-base/group_vars/all/site-specific``,
-which you can edit manually to provide further customization.
+which you can edit in case of errors such as typos before rerunning the script. 
+You can also run ``./securedrop-admin sdconfig --force`` to remove your entire 
+configuration file and start over.
 
 When you're done, save the file and quit the editor.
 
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -92,7 +92,10 @@ The install process will take some time, and it will return
 the terminal to you when it is complete. If an error occurs while
 running the install, please submit a detailed `GitHub
 issue <https://github.com/freedomofpress/securedrop/issues/new>`__ or
-send an email to securedrop@freedom.press.
+send an email to securedrop@freedom.press. You can also attempt to rerun
+the Ansible playbooks if you had a minor issue, such as a typo, after making
+edits to the file that are read from ``/install_files/ansible_base/prod_specific.yml``.
```
redshiftzero: I think you want to cover two cases here, the case of an admin making a mistake during `sdconfig`, and the case of Tor being flaky (or some other intermittent error), in which case the admin would want documentation to:

1. Describe that they can manually edit `install_files/ansible-base/site-specific/group_vars/all/site-specific` (used instead of `prod-specific.yml`) _or_ re-run `./securedrop-admin sdconfig --force` to retype all vars in `site-specific`.

2.  Describe that it's safe to rerun `./securedrop-admin install` if they see an error and let them know they aren't going to break anything (this closes #2098).

redshiftzero: Also let's put this troubleshooting above "make an issue or email us" :-)

freddymartinez9:
rebased on top of the latest developed. `make docs-lint` continues to pass locally

PR #2344, opened 2017-09-20
l10n: support localized ascii wordlist for passphrases

dachary:
## Status

Ready for review

## Description of Changes

Support ascii localized wordlist for passphrases and add a french diceware list. If there is no wordlist for the current locale, fallback to english. A directory is used to store the wordlists organized by localed but a symbolic preserves the old naming convention for backward compatibility.

## Testing

* pytest -v tests
* The source interface should be accessed from the staging virtual machine to verify the apparmor list is updated to allow access to the wordlist at their new location in the directory tree

## Deployment

The old english wordlist is replaced with a symbolic link to the new english wordlist.

2. New installs.

Nothing specific.


dachary:
@heartsucker I should have made it clear this pull request is exclusively for ascii based wordlists. Working with unicode wordlists is going to be another kind of challenge. Note that the French wordlist introduced in this pull request is only made of words with no accented letters.

heartsucker:
> this pull request is only made of words with no accented letters.

Does this mean it was chosen as such or that it just chopped the accents off to make it ascii? I think it might be better to solve this problem now before we start adding localized wordlists.

dachary:
> Does this mean it was chosen as such

It was: the words do not have accented letters converted to non-accented equivalent

dachary:
https://circleci.com/gh/freedomofpress/securedrop/2354?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

<pre>
Slave 'gw24' crashed while running 'molecule/builder/tests/test_build_dependencies.py::test_build_dependencies[ansible://build-python-dev]'
</pre>

dachary:
https://circleci.com/gh/freedomofpress/securedrop/2356?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

<pre>
CommandResult(command='aptitude --simulate -y safe-upgrade', exit_status=255, ...lecule/aws/.molecule/securedrop-ci-2356-config: No such file or directory\r\n").rc
</pre>

dachary:
@conorsch @msheiny does it looks like the CircleCI errors at https://github.com/freedomofpress/securedrop/pull/2344#issuecomment-331615346 and https://github.com/freedomofpress/securedrop/pull/2344#issuecomment-331700211 could have been introduced by the recent refactor of the test logic ?

securedrop/tests/test_crypto_util.py
```diff
@@ -133,14 +133,29 @@ def test_basic_encrypt_then_decrypt_multiple_recipients(self):
...
+        id = crypto_util.genrandomid(locale=locale)
         id_words = id.split()
 
         self.assertEqual(id, crypto_util.clean(id))
         self.assertEqual(len(id_words), crypto_util.DEFAULT_WORDS_IN_RANDOM_ID)
         for word in id_words:
-            self.assertIn(word, crypto_util.words)
+            self.assertIn(word, crypto_util._get_wordlist(locale))
+
+    def test_genrandomid(self):
```
redshiftzero: Solid tests :+1:. One thought: where possible let's name these tests to be more descriptive, e.g. `test_genrandomid_default_locale_is_en`. I like this approach because it makes for very clear natural documentation of the expected behavior of the code (even before you read the test).

README.md
```diff
@@ -38,4 +38,9 @@ We also have a public [Gitter channel](https://gitter.im/freedomofpress/securedr
 
 SecureDrop is open source and released under the [GNU Affero General Public License v3](/LICENSE).
 
-The [wordlist](/securedrop/wordlist) we use to generate source passphrases is based off a new [Diceware wordlist](https://www.eff.org/deeplinks/2016/07/new-wordlists-random-passphrases) from the EFF.
+## Wordlists
+
+The wordlist we use to generate source passphrases come from various sources:
+
+* [en.txt](/securedrop/wordlists/en.txt) is based off a new [Diceware wordlist](https://www.eff.org/deeplinks/2016/07/new-wordlists-random-passphrases) from the EFF.
+* [fr.txt](/securedrop/wordlists/en.txt) is based Matthieu Weber's [translated diceware list](http://weber.fi.eu.org/index.shtml.en).
```
redshiftzero: `based` -> `based off`

securedrop/crypto_util.py
```diff
@@ -47,7 +47,7 @@ def do_runtime_tests():
 
 gpg = gnupg.GPG(binary='gpg2', homedir=config.GPG_KEY_DIR)
 
-words = open(config.WORD_LIST).read().rstrip('\n').split('\n')
+language2words = {}
```
redshiftzero: Hmm, what is `language2words` intended for?

securedrop/crypto_util.py
```diff
@@ -77,8 +77,38 @@ def clean(s, also=''):
...
+
+        language2words[locale] = open(
+            wordlist_path).read().rstrip('\n').split('\n')
+
+    return language2words[locale]
+
+
+def genrandomid(words_in_random_id=DEFAULT_WORDS_IN_RANDOM_ID, locale='en'):
+    return ' '.join(random.choice(_get_wordlist(locale))
+                    for x in range(words_in_random_id))
```
redshiftzero: On the journalist side the source's `journalist_designation` will continue to be two English words - because the two wordlists used for this designation are in English and are separate from the wordlist here (this doesn't block merge, just noting).

redshiftzero:
(Note #2377 was merged)

README.md
dachary: done

securedrop/crypto_util.py
dachary: Added a comment above to clarify:

    # map code for a given language to a localized wordlist

securedrop/crypto_util.py
dachary: Yes, it could be improved in a separate pull request. The focus of this pull request is to have passphrases that are easier to remember when a localized wordlist is available. And since the journalist does not need to remember the source designation, it is less of a problem if it always is in english.

securedrop/tests/test_crypto_util.py
dachary: done

dachary:
> Remind me what was the takeaway there? If needed we can do a bit more digging with people we know that are interested in translated wordlists (and worst case we add another project to Weblate).

I thought it would just be about translating the english wordlist. But it looks like it is more complicated and the wordlist needs to be adapted for each language and culture. I did not dig deeper than this because it felt like a topic that requires time and patience. Also I have the feeling that people may be studying this problem and working on new wordlists.

dachary:
@heartsucker do you approve of an ascii-only diceware list support ?

heartsucker:
> and worst case we add another project to Weblate

Translating the wordlists is not a good idea. The whole point is unique words that are simple and common in the native language.

dachary:
testing upgrade from 0.4.3 as follows:

* vagrant destroy app-staging # mon-staging is already up
* git checkout -b release/0.4.3 origin/release/0.4.3
* make build-deb
* cp -a build build-0.4.3 # to avoid rebuilding later, handy
* vagrant up app-staging
*

Issue #1966, opened 2017-07-12
Automated web application configuration migration


garrettr:
The SecureDrop web application's configuration is read from `config.py`, which is generated from a template, `config.py.example`, by the `initialize_securedrop_app` [task](https://github.com/freedomofpress/securedrop/blob/892cf52299f17ed9c70df0bb0bec68137101bab2/install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml#L29-L151) during the initial run of the Ansible playbooks.

During the initial development of the SecureDrop auto-upgrade strategy, we faced a dilemma in managing `config.py`. The dilemma is mostly captured in comments ([1](https://github.com/freedomofpress/securedrop/blob/892cf52299f17ed9c70df0bb0bec68137101bab2/install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml#L29-L33), [2](https://github.com/freedomofpress/securedrop/blob/892cf52299f17ed9c70df0bb0bec68137101bab2/install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml#L59-L65)) in the relevant Ansible role, but I will summarize it here for clarity.

At the time of development, the most straightforward way to provide an auto-upgrade capability for SecureDrop was to use Debian packages. The SecureDrop application code is provided by the `securedrop-app-code` package, and production instances check https://apt.freedom.press periodically for updates to this package. Files that contain instance-specific configuration cannot be included in the Debian package because they would be overwritten on update.

Additionally, the instance-specific configuration depends on configuration values that must be provided by the SecureDrop administrator during installation, so the task of generating and maintaining the web application configuration fell to the Ansible playbooks, which are configured with vars by the administrator during installation and run from the Admin Workstation.

Finally, there are some configuration values which automatically generated by the Ansible playbooks. For security, these values must be unique, random, and generated from a secure CSPRNG (e.g. `/dev/urandom`). These values [are](https://github.com/freedomofpress/securedrop/blob/892cf52299f17ed9c70df0bb0bec68137101bab2/install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml#L71-L133):

* `source secret key`—the Flask `SECRET_KEY` for the Source Interface, used as an HMAC key to secure Flask's sessions from tampering.
* `journalist secret key`—same, but for the Journalist Interface.
* `scrypt id pepper`—random value used as an instance-wide salt (aka "pepper") for deriving Source's filesystem ID's from their codenames.
[...]

PR #2349, opened 2017-09-21
i18n: add language selection to source/journalist

dachary:
## Status

Ready for review

## Description of Changes

Add language selection to source/journalist interfaces. It will not show by default because there is less than one supported language at the moment. It needs to look better (web design welcome) but the code logic should remain the same.

## Testing

* visual verification can be tried on http://i18n.securedrop.club for both the source and the journalist interface

## Deployment

The new configuration variable SUPPORTED_LOCALES is optional and existing installations will not be affected.

2. New installs.

A comment about the SUPPORTED_LOCALES variable is included. An admin can manually define it on the app server for test purposes although it is not documented. Another pull request should implement a way to set this variable during the sdconfig stage, when localization is officially supported.


securedrop/source_templates/index_locales.html
```diff
@@ -0,0 +1,11 @@
+{% if g.locales|length > 1 %}
```
heartsucker: I don't like the idea of the `index_locales`. This should just reuse the other `locales` bit and we can use CSS to make it work in both locations. This feels like unnecessarily duplicating code.

securedrop/i18n.py
```diff
@@ -72,3 +75,47 @@ def get_locale():
...
+        return sorted(locales)
+    unsupported = set(supported) - set(locales)
+    if unsupported:
+        raise Exception("config.py SUPPORTED_LOCALES contains {} which is not "
+                        "among the locales found in the {} "
+                        "directory: {}".format(list(unsupported),
+                                               babel.translation_directories,
+                                               locales))
+    if default_locale and default_locale not in supported:
+        raise Exception("config.py SUPPORTED_LOCALES contains {} "
```
heartsucker: We should make this a `LocaleNotFound` exception or something similar so we can catch it with higher specificity in the unit test.

securedrop/tests/test_i18n.py
```diff
@@ -133,6 +166,9 @@ def test_i18n(self):
         pybabel init -i {d}/messages.pot -d {d} -l zh_Hans_CN
         sed -i -e '/code hello i18n/,+1s/msgstr ""/msgstr "code chinese"/' \
               {d}/zh_Hans_CN/LC_MESSAGES/messages.po
+        pybabel init -i {d}/messages.pot -d {d} -l nb_NO
```
heartsucker: Pedantic: Could you add a newline here since this is a slight wall of text?

securedrop/i18n.py
dachary: Good idea, done !

securedrop/tests/test_i18n.py
dachary: Done. Indeed less confusing.

dachary:
modified & repushed. About the index_locale.html I agree that we could DRY this with CSS. However, there is 100%  certainty this will be replaced by something prettier in the immediate future: that's my excuse for not doing it. That being said, if you think it should be done, I'll do it. I don't feel strongly about it.

heartsucker:
I think it's fine if it's ugly for now, and if you remove `index_locales`, the next diff will be cleaner too.

dachary:
@heartsucker removed local_index.html as suggested. It feels cleaner indeed, thanks for pushing in the right direction !

source-index

heartsucker:
@dachary If this gets merged without the prettification of the locales (which I presume it will), I'm leaving it up to you to open a follow up ticket for making it look nice.

dachary:
> I'm leaving it up to you to open a follow up ticket for making it look nice.

Will do :-)

redshiftzero:
Linking to #1966 for `SUPPORTED_LOCALES`

PR #2353, opened 2017-09-21
jinja: implement localized filesizeformat() and datetimeformat() filters

dachary:
## Status

Ready for review

## Description of Changes

The babel.units module provides support for localized digital units
display. It is not identical to the filesizeformat() filter natively
provided by jinja because:

* it only supports powers of 1024 (i.e. decimal prefixes like kB, MB
  etc. and not binary prefixes like KiB, MiB etc.)
* it only supports units up to the terabyte and not yoctobyte

The advantage of having translations of the prefixes in dozens of
languages outweight the incompatibility.

The babel.dates implements localized display of relative and regular date functions.

## Testing

* pytest -v tests

## Deployment

This is only a user facing change.

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -42,6 +42,7 @@ else:
     app.jinja_env.globals['use_custom_header_image'] = False
 
 app.jinja_env.filters['datetimeformat'] = template_filters.datetimeformat
+app.jinja_env.filters['filesizeformat'] = template_filters.filesizeformat
 
 
 @app.teardown_appcontext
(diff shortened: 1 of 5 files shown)


heartsucker:
The request headers had `Accept-Language: en-US,en;q=0.5` and yet I got what appears to be a Norwegian string for the relative time.

screenshot_2017-09-21_13-06-37

dachary:
@heartsucker I can't reproduce the problem, could you please give me detailed instructions to get the same result ?

heartsucker:
@dachary I think this has something to do with the flask application not clearing sessions correctly and some values are left in the cookie. As in, a fresh login in a browser window with an old session will keep the language. I think. I'm working on some other things right now, but I'm pretty sure my rejection of this PR is based on an unrelated issue. I'll get back to it when I'm done with my current PR.

heartsucker:
Clearing my sessions has resolved this.

Issue #2230, opened 2017-08-30
Merge in first translation (French)


redshiftzero:
Still a little way to go yet to complete the i18n process, but I'm filing this as a reminder when we merge in the first SecureDrop translation:

- [ ] Merge in `.po` file exported from Weblate
- [ ] Remove `@pytest.mark.xfail` from `test_deb_package_contains_mo_file` (from #2135)

@dachary let me know if you disagree that French is a good first target for including in a release, I figured you would be on board for that 😉 .

dachary:
@redshiftzero it's going to be a close race between French and Norwegian :-) Both are currently 100% translated. But French is a little ahead because a reviewer from localizationlab stepped in. Suspens !

heartsucker:
TODO:

- [ ] Add checks that `<html lang="fr">` when the locale is French for
  - [ ] `journalist_templates/base.html`
  - [ ] `source_templates/base.html`
  - [ ] `source_templates/index.html`

This should be sufficient for the future. If it works with two languages, it will work with the rest.

PR #2355, opened 2017-09-21
i18n: support for <html lang="$lang">

### LINKED ISSUE #2354: Add <html lang="$some_lang"> to HTML templates
# Feature request

## Description

All pages should have the `<html lang="en-US">` or other appropriate language tag.

Reference: https://www.w3schools.com/Tags/ref_language_codes.asp

## User Stories

As a source/journalist, I want my browser to appropriately handle things like CSS text transforms.

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #2354.

Adds the RFC 5646 language tag to the HTML.

## Testing

`pytest tests`.

More complete testing depends on merging in some translations.

### DIFF
--- a/securedrop/i18n.py
+++ b/securedrop/i18n.py
@@ -22,7 +22,9 @@ from babel import core
 import collections
 import config
 import os
+import re
 
+LOCALE_SPLIT = re.compile('(-|_)')
 LOCALES = set(['en_US'])
 babel = None
 
@@ -129,3 +131,13 @@ def get_locale2name():
             locale = core.Locale.parse(l)
             locale2name[l] = locale.languages[locale.language]
     return locale2name
+
+
+def locale_to_rfc_5646(locale):
+    lower = locale.lower()
+    if 'hant' in lower:
+        return 'zh-Hant'
+    elif 'hans' in lower:
+        return 'zh-Hans'
+    else:
+        return LOCALE_SPLIT.split(locale)[0]
--- a/securedrop/tests/test_i18n.py
+++ b/securedrop/tests/test_i18n.py
@@ -19,6 +19,7 @@
 import argparse
 import logging
 import os
+import re
 
 from flask import request, session, render_template_string, render_template
 from flask_babel import gettext
@@ -32,9 +33,14 @@ import manage
 import pytest
 import source
 import version
+import utils
 
 
 class TestI18N(object):
+
+    @classmethod
+    def setup_class(cls):
+        utils.env.setup()
 
     def test_get_supported_locales(self):
         locales = ['en_US', 'fr_FR']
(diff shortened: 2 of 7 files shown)


dachary:
It looks like [w3school list](https://www.w3schools.com/Tags/ref_language_codes.asp) contradicts the [wikipedia list](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) for ISO-638-1 codes. If we are to believe wikipedia, getting the first part of the locale returned by get_locale() (without the **script** part that only exists in chinese), it would work. What do you think ?

heartsucker:
Yeah, I thought about just truncating it and hoping for the best. Let's see what other people think.

dachary:
For completeness, it can be done with the following babel snippet:
<pre>
from babel import core
core.parse_locale('fr_FR')[0] => fr
core.parse_locale('zh_Hant')[0] => zh
etc.
</pre>
This is essentially what get_locale() does with two useful additions:
* it will accept **-** instead of **_** 
* it will fallback to a supported locale if the requested locale is not availble (i.e. DEFAULT_LOCALE)
So using the following function
<pre>
def get_language():
    return get_locale().split('_')[0]
</pre>
should get you what you need for the lang attribute.

heartsucker:
@dachary Ok, I'm going to use that and then add the exceptions for `zh_Han{s,t}` as they are the two obvious outliers.

heartsucker:
@redshiftzero Can someone restart travis? It looks like a transient OS error.

redshiftzero:
Restarted :-)

heartsucker:
I think we have a new, different transient CircleCI error :(

msheiny:
> I think we have a new, different transient CircleCI error :(

I'm seeing that on my AWS branch --- 👍 and I ended up doing an `XFAIL` which i have to fix. I wasnt able to reproduce in CI manually though. its really annoying. is there a fancy way to re-run that multiple times until it passes with pytest?

securedrop/tests/test_i18n.py
```diff
@@ -173,6 +174,33 @@ def test_verify_default_locale_en_us_if_not_defined_in_config(self):
...
+        assert i18n.locale_to_rfc_5646('en-us') == 'en'
+        assert i18n.locale_to_rfc_5646('zh-hant') == 'zh-Hant'
+
+    def test_html_lang_correct(self):
+        app = journalist.app.test_client()
+        resp = app.get('/', follow_redirects=True)
+        html = resp.data.decode('utf-8')
+        assert re.compile('<html .*lang="en".*>').search(html), html
+
+        # TODO add check for a non-English language on the journalist app
```
redshiftzero: Can you remove these TODOs and add a @checklist item in #2230 to remind us we should add testing for `lang="fr"` when we merge in the French translation?

securedrop/i18n.py
```diff
@@ -129,3 +129,13 @@ def get_locale2name():
...
+
+
+def locale_to_rfc_5646(locale):
+    lower = locale.lower()
+    if 'hant' in lower:
+        return 'zh-Hant'
+    elif 'hans' in lower:
+        return 'zh-Hans'
+    else:
+        return core.parse_locale(locale.split('-')[0])[0]
```
dachary: <pre>
locale.split('-')[0]
</pre>
should be enough here ?

heartsucker:
@dachary @redshiftzero This is ready for re-review.

securedrop/tests/test_i18n.py
```diff
@@ -211,6 +219,29 @@ def test_verify_default_locale_en_us_if_not_defined_in_config(self):
...
+        resp = app.get('/', follow_redirects=True)
+        html = resp.data.decode('utf-8')
+        assert re.compile('<html .*lang="en".*>').search(html), html
+
+        app = source.app.test_client()
+        resp = app.get('/', follow_redirects=True)
+        html = resp.data.decode('utf-8')
+        assert re.compile('<html .*lang="en".*>').search(html), html
+
+        # check '/generate' too because '/' uses a different template
```
dachary: I don't see how this is different from the above ? did you mean app.get('/generate') ?

redshiftzero:
Once @dachary's [comment here](https://github.com/freedomofpress/securedrop/pull/2355/files/78879c19572f560f9bd78707b9b2e98e7eb93b12#r140652041) is addressed (a good catch), this is ready to merge.

heartsucker:
@redshiftzero @dachary ping

heartsucker:
CircleCI seems to have hiccuped. Restart?

redshiftzero:
Restarted

PR #2358, opened 2017-09-22
Add test changes for @comradekingu string changes on journalist interfaces

redshiftzero:
## Status

Ready for review

## Description of Changes

Adds test changes for @comradekingu's changes introduced in #2251 and #2301 (those commits are cherry-picked here).

## Testing

Do tests pass?

## Deployment

They may update the strings to be translated, but otherwise should be fine


redshiftzero:
Thanks for the super fast approval @heartsucker 🙃

PR #2364, opened 2017-09-23
tests: ignore config.py SUPPORTED_LOCALES during tests

dachary:
## Status

Ready for review

## Description of Changes

The test_i18n.py tests must ignore the content of the
SUPPORTED_LOCALES config.py variable when they run, otherwise the
result depends on it and it is unstable.

## Testing

* set SUPPORTED_LOCALES = ['en_US'] in config.py
* pytest -v tests/test_i18n.py fails with
<pre>
>           assert 'fr_FR' in locales
E           AssertionError: assert 'fr_FR' in ''
</pre>
* use this pull request and see it passes regardless of the content of SUPPORTED_LOCALES 

## Deployment

Test only.

### DIFF
--- a/securedrop/tests/test_i18n.py
+++ b/securedrop/tests/test_i18n.py
@@ -195,10 +195,17 @@ class TestI18N(object):
 
         manage.translate(args)
 
-        for app in (journalist.app, source.app):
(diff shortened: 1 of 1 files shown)


heartsucker:
Note: the contents of this are contained in #2355 in order to get it to pass the test suite.

PR #2366, opened 2017-09-23
l10n translations from weblate

dachary:
## Status

Ready for review

## Description of Changes

All translations from https://weblate.securedrop.club/, exclusively modifying the translations directory. We're not supporting any languages just yet, just populating the translations with everything we have and getting ready for continuous updates. The workflow for doing so is being worked on and will be documented in a separate pull request.

## Testing

In addition to the CI the following should be done manually:

* PAGE_LAYOUT_LOCALES='fr_FR,de_DE,nl,nb_NO' pytest -v --page-layout tests/pages-layout

and this will validate these languages do not have rendering issues.

## Deployment

These files are not currently used and should not interfere.


heartsucker:
heavybreathing

dachary:
Squashed ~50 commits to reduce the noise

dachary:
We're going about it differently, copying instead of merging.

PR #2367, opened 2017-09-23
docs: l10n and development workflow 

dachary:
## Status

Ready for review

## Description of Changes

The localization lifecycle is different from the development lifecycle
and uses different tools. The proposed workflow allows developers and
translators to work in parallel without blocking each other provided
the developers do not modified the localization files by other means.

Weblate needs to be able to push commits somewhere. Granting write
permissions to the official SecureDrop repository does not sound like
a sane thing to do. Instead, a dedicated bot user was created and owns
a fork of the SecureDrop repository, at http://lab.securedrop.club.

Weblate pushes commits with new translations to the i18n branch of
this repository. These commits are collected from time to time and
proposed as a pull request for merge into the develop branch.

In parallel developers modify the develop branch of the official
repository and may modify, remove or insert new strings to be
translated. From time to time a developer needs to:

* merge the develop branch into the i18n branch
* run ./manage.py translate --extract-update

Merging develop into i18n will never generate conflicts because
developers never modify the files same files as weblate.

To keep the develop branch as clean as possible, the commits created
by weblate are cherry-picked. Although it would be possible to merge
the i18n branch back into develop, the merge commits unique to the
i18n branch would also be added to the develop branch and make it more
difficult to read.

## Testing

* make docs
* firefox http://127.0.0.1:8000/development/i18n.html

## Deployment

Documentation only, not deployment considerations.


docs/development/l10n.rst
```diff
@@ -91,6 +91,27 @@ by clicking on it:
...
+
+will be displayed to the user as::
+
+  Edit user Jean-Claude
+
+The French translated string should look like::
+
+  Modifier l'utiliser {user}
+
+And it would be **incorrect** to translate the placeholder like so::
```
heartsucker: 👍

heartsucker:
Also, we need to update the internal FPF doc that reviewers need to be attentive with the merges since we can get XSS injections via the strings since a lot of them are called via `gettext('foo')|safe` or use the `Markup` type.

docs/development/i18n.rst
```diff
@@ -72,19 +80,102 @@ the file can be found in the function that created it in
...
+already in the `origin/develop` branch and the `patch they contain is unique`_
+so it is not confused with a previous cherry-pick of the same patch.
+
+Merging develop into the weblate fork
+-------------------------------------
+
+`Weblate`_ works on a long standing fork of the `SecureDrop git
+repository`_ and is exclusively responsible for the content of the
+`*.pot` and `*.po` files. It needs to merge the content of the
+`develop` branch back into its `i18n` branch to be able to extract
```
redshiftzero: these strings need to be wrapped in double backticks to render as fixed width

docs/development/i18n.rst
```diff
@@ -72,19 +80,102 @@ the file can be found in the function that created it in
...
+
+      $ git clone https://github.com/freedomofpress/securedrop
+      $ cd securedrop
+      $ git remote add lab http://lab.securedrop.club/bot/securedrop/tree/i18n
+      $ git fetch lab
+      $ git checkout -b wip-i18n origin/develop
+      $ h=$(git log --reverse --pretty='%H' --cherry-mark --right-only \
+            origin/develop...lab/i18n)
+      $ git cherry-pick $h
+      $ git push wip-i18n
```
redshiftzero: Thanks for writing this - this workflow seems reasonable! I have a concern regarding merging in translations after taking a look at #2366 - there will likely quickly become hundreds of automatic commits from Weblate littering our git history if we merge these automated commits as is. For example our last release was ~370 commits, which would quickly be swamped by automatic commits from translators doing their work. 

I took a look at how a few other projects, e.g. Tor and HTTPS everywhere, handle this issue to merge in translations, and I think they have a nice workflow - they have a [translation-only](https://gitweb.torproject.org/translation.git/) repository with one branch per project that commits are automatically pushed to when strings are translated. This repository is either added as a submodule, or changes are periodically [brought in](https://gitweb.torproject.org/bridgedb.git/commit/?id=fec3696d6826c8e8227e5286ffda5502249d3fc3) from the translation repo, bundling all the changes for one language at a time. 

Having a translation-only repository seems unnecessary given that you've set up `lab.securedrop.club` (:clap:), but squashing translation commits together into one commit per release per language for all the changes in that release is important in order for us to maintain the utility of our git history. This would add an additional step to the workflow above. What do you think?

dachary: This sounds like an excellent idea ! I'll have to think a little more about the details but it is going to be more robust, won't clutter the git history and should not cause any merge problem because git should be able to cope with the commit asymmetry.

dachary: I'm not so convinced this is the way to go anymore. The benefit of taking the time before making a decision ;-) I should point out that weblate does not create a commit each time a string is translated. That would be extremely noisy. When a translator starts work the language is locked. After a period of inactivity all the changes done during are committed into a single commit and pushed to the i18n branch. In other words we have a single commit per translation session.

At the beginning I made a lot of small changes and this may have given a different impression. I squashed all those into a single larger commit (https://github.com/freedomofpress/securedrop/pull/2366 now has ~100 commits instead of ~150). I stopped squashing when the first contributor started translating, to not remove his commit and preserve the attribution of his work.

These ~100 commits are spread over a period of ~100 days which is roughly one commit per day. I **hope** we can keep this level of engagement in the future ;-) And I also think is is unlikely to grow much more so we're in no real danger to drown under the localization commits.

# The proposed workflow

## Cons

* Localization commits are mixed with code commits and clutter the history

## Pros

* Translators are credited for their contribution in the main codebase
* a fork that is periodically merged back: there is no room for divergence (i.e a script that is in the translation repository but not in the securedrop repository)

# Workflow relying on squashed commits being copied from the translation repository to the secureDrop repository
 
## Cons

* merging/using translations with the latest from develop is more work than a merge
* translators contribution have to be acknowledged separately (i.e. not as code contributors)
* a separate repository is a risk of durable divergence

## Pros

* A single localization commit per release or so, less clutter

dachary:
@redshiftzero a friendly reminder about https://github.com/freedomofpress/securedrop/pull/2367#discussion_r141625589

docs/development/i18n.rst
redshiftzero: Great writeup @dachary 🚀 . And thanks for the clarification on the commit per modification, that's good to know :+1:. My reservation still stands, currently we have a few languages with active translators/reviewers (~10 languages or so, thanks to your hard work), and as we get more activity, even one commit per translation session is going to be a a bit of noise (for comparison, Tor's Transifex has something like ~150 languages).

But I hear you, and I agree it's important that contributors are acknowledged, so I think we should add a note to the README linking to the contributions page (or equivalent) for the repository that get commits directly from Weblate (note: we could make a `securedrop-translations` repository in the `freedomofpress` group and have Weblate push commits to that if that is easier).

I _think_ divergence shouldn't be an issue if Weblate is watching `develop` in this repository and then either commits are cherry-picked from a fork and squashed or the relevant translation file is copied over, but let me know if I'm overlooking a way that divergence can be introduced. 

Summarizing, I propose we:

0. Squash together all commits for a particular language (or otherwise just copy the translations for a particular language since it's one file) and propose them as PRs for merge into the main repository to avoid clutter and follow the practice that e.g. [Signal](https://github.com/WhisperSystems/Signal-Android/commit/edc00d858b79e47ad8f9e658008455f621adafbc) and Tor are using to incorporate translations. 
1. Add a "How to Contribute Translations to SecureDrop" to the project `README.md` pointing to the translation portal and the contributors page for the fork that we use to get commits from Weblate so that people can see the translator and reviewer contribution history. 

Let me know your thoughts. And I'm happy to do this or help the doer with this! 😄

heartsucker:
One more note. For source strings like "Open the X menu and press Y button," it's not enough to translate the word. Someone needs to use Arabic Tor Browser and check what the menu is actually called.

docs/development/i18n.rst
dachary: Let's do it this way :-)I don't have strong feelings about this and what you're suggesting makes sense. Regarding contributions I propose we add an unsorted list of contributors who have at least one commit (code or translation) to the blog post announcing a release. This list is easy to compile for each repository. What do you think ?

dachary:
@heartsucker I'll add a note about that, good point.

dachary:
@heartsucker added **docs: l10n instructions about a user navigation** commit to address what you raised. Does it look ok ?

dachary:
@redshiftzero I updated the documentation to copy files instead of merging the commits. The rest does not need change... unless I missed something ? If that's ok with you, we can copy over the first translations shortly after **release/0.4.4** is branched and iterate from there.

heartsucker:
@dachary I thought we were going to do something like squashing the commits per language to keep the amount of noise down, yes?

dachary:
@heartsucker I figured it would be easier to just copy the files over (the checkout does not get the commits, just the content of the files) and create a new commit instead. It is easier and does the same... unless I missed something ?

heartsucker:
@dachary Ah yes. Ok, so one thing that might be nice is to include a list of usernames in the squashed commit somehow so that we can give appropriate credit. Maybe we could setup a shell script that does the above plus scrapes all the usernames into a unique sorted list from the git log?

docs/development/l10n.rst
```diff
@@ -91,6 +91,40 @@ by clicking on it:
 
 |Weblate translate context screenshot|
 
+Instructions about a user navigation
+------------------------------------
+
+The strings my contain instructions relative to the user interface of
```
redshiftzero: Nit: `my` -> `may`

redshiftzero:
@heartsucker that's an interesting idea! In the discussion [here](https://github.com/freedomofpress/securedrop/pull/2367#discussion_r142560622) we talked about adding a note in the SecureDrop readme pointing to the contributions page of the repo that gets the Weblate commits pushed to it. Let's discuss that and your idea in this issue: #2432

docs/development/l10n.rst
dachary: Ooops, thanks for catching the typo :-)

dachary: Fixed & repushed

redshiftzero:
[Staging build](https://circleci.com/gh/freedomofpress/securedrop/2758) failure:

```
    TASK [Wait for server to come back.] *******************************************
    Tuesday 10 October 2017  09:22:41 +0000 (0:00:01.073)       0:00:03.799 *******
Makefile:19: recipe for target 'ci-go' failed
make: *** [ci-go] Terminated
Too long with no output (exceeded 10m0s)
```

redshiftzero:
[One more timeout](https://circleci.com/gh/freedomofpress/securedrop/2766):

```

    RUNNING HANDLER [Create remote CI hosts on AWS.] *******************************
    Wednesday 11 October 2017  14:02:30 +0000 (0:00:00.827)       0:01:15.069 *****
    ok: [localhost] => (item={'name': u'mon-staging', 'groups': [u'securedrop_monitor_server', u'mon-staging', u'staging', u'securedrop']})
    changed: [localhost] => (item={'name': u'app-staging', 'groups': [u'securedrop_application_server', u'app-staging', u'staging', u'securedrop']})
    ...snip...!
 "msg": "Timeout when waiting for ec2-13-56-184-34.us-west-1.compute.amazonaws.com:22"}
    
    PLAY RECAP *********************************************************************
    localhost                  : ok=13   changed=8    unreachable=0    failed=2
```

PR #2371, opened 2017-09-26
Smoothing wording replies source interface

KwadroNaut:
replies-0
It's not a button, doesn't require any action.

## Status

### Made a change to wording, need to regenerate .pot!

### DIFF
--- a/securedrop/source_templates/lookup.html
+++ b/securedrop/source_templates/lookup.html
@@ -50,7 +50,7 @@
 
 <hr class="no-line">
 
-<h2 class="headline">{{ gettext('Get Replies') }}</h2>
+<h2 class="headline">{{ gettext('Read Replies') }}</h2>
 
 <div id="replies">
   {% if replies %}


KwadroNaut:
I see your point. In other languages it would sound just as silly as in English. In for example a MUA a 'get replies/messages' would initiate a connection to a server, check for new messages etc. This isn't happening here, there would be a (list of) replies for the submitter or nothing. Your point, both headlines being a call to action, stands. So what about rephrasing it to "Read Replies"?

redshiftzero:
OK, that sounds good @KwadroNaut 😄  Go ahead and make that change.

dachary:
@KwadroNaut ping ?

KwadroNaut:
@dachary thanks for reminding me, changed as requested!

dachary:
@KwadroNaut  a short note to let you know that @redshiftzero  is currently busy with other things. Your PR is not overlooked, it's just a matter of waiting a few more days. Thanks for your patience :-)

redshiftzero:
In light of our updated [release schedule and relaxed string freeze](https://forum.securedrop.club/t/securedrop-vulnerability-and-return-to-regular-release-cycle/186), I'm going to merge this one. Thanks @KwadroNaut :)

Issue #2043, opened 2017-07-25
db.Source.journalist_designation is not a unique field


fowlslegs:
# Bug

## Description

`db.Source.journalist_designation` provides a 2-word designation which is used in the JI by journalists as the primary mechanism to identify and distinguish between sources (whose real identity and codename remain secret). Since the uniqueness of the field is not being enforced, it's possible that two Sources might end up with the same journalist designation, causing confusion.

## Comments

The fix is pretty straightforward. We should set the column/ attribute with `unique=True` and changes to this attribute/ field should be done with something like this pseudo-Python:
```py
while True:
    try:
        journo.journalist_designation = gen_display_id()
        db_session.add(journo)
        db_session.commit()
    except UniquenessConstraintException:
        pass
    else:
       break
````

redshiftzero:
Note: adding a [unique constraint](http://docs.sqlalchemy.org/en/latest/core/metadata.html#sqlalchemy.schema.Column.params.unique) to the `journalist_designation` is blocked on #1419

redshiftzero:
Well, we could add a `while` loop in `crypto_util.display_id()` similar to [the loop in `source_app.utils.generate_unique_codename()`](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/source_app/utils.py#L38) to make sure that we cycle the `journalist_designation` if a duplicate is generated, that _would_ work going forward. 

But - the right place to solve this is the database. Specifically we should add the unique constraint, and if a source with a duplicate `journalist_designation` is added to the database, we handle the `IntegrityError` thrown by rolling back the transaction and cycle the `journalist_designation`. Let's tackle after #1419.

prateekj117:
@redshiftzero I would like to work on this.

emkll:
Hi @prateekj117 !

Thanks for reaching out, your help would be appreciated here. After some discussion with @rmol, we would recommend you adopt the strategy described in the strategy described in https://github.com/freedomofpress/securedrop/issues/2043#issuecomment-332990259 (mirroring `generate_unique_codename` function as opposed to adding a database constraint and creating a database migration)

 The proposed fix for setting `unique` attribute for `journalist_designation` column may be somewhat risky due to the auto-updating of the web application: Database migrations are automatically performed when the application is updated. To make sure we cover all cases, the Alembic migration, we would need to handle the (unlikely) event a duplicate journalist designation already exists in the database by:
- renaming journalist designation in the database (to ensure database integrity)
[...]

PR #2377, opened 2017-09-28
Increase unit test coverage on the source interface

redshiftzero:
## Status

Ready for review

## Description of Changes

I have inspected the code on the source interface that is untested and the changes in this PR are scoped to these areas. 

* This PR adds unit tests for some untested code paths on the source interface.
* In one case there was uncovered code that was dead code, so I have removed it in 
 0b02e88
* In 330a6d9 there was a missing rollback so I added that

If breaking this up would be useful, I am happy to do so 😇 . Else I have tried to be descriptive in the commit messages for each change. 

## Testing

1. Does CI pass? 
2. Step commit by commit and examine my reasoning behind each change described in the commit message. 

## Deployment

Mostly unit tests, else the expected behavior is covered by the tests added here. 

### DIFF
--- a/securedrop/source_app/main.py
+++ b/securedrop/source_app/main.py
@@ -48,6 +48,7 @@ def make_blueprint(config):
         try:
             db_session.commit()
         except IntegrityError as e:
+            db_session.rollback()
             current_app.logger.error(
                 "Attempt to create a source with duplicate codename: %s" %
                 (e,))
--- a/securedrop/tests/test_source.py
+++ b/securedrop/tests/test_source.py
@@ -8,7 +8,8 @@ from bs4 import BeautifulSoup
 from flask import session, escape
 from flask_testing import TestCase
 
-from db import Source
+import crypto_util
+from db import db_session, Source
 import source
 import version
 import utils
@@ -357,3 +358,54 @@ class TestSourceApp(TestCase):
             self.assertFalse(mock_hash_codename.called,
                              "Called hash_codename for codename w/ invalid "
                              "length")
+
+    @patch('source.app.logger.warning')
+    @patch('subprocess.call', return_value=1)
+    def test_failed_normalize_timestamps_logs_warning(self, call, logger):
+        """If a normalize timestamps event fails, the subprocess that calls
+        touch will fail and exit 1. When this happens, the submission should
+        still occur, but a warning should be logged (this will trigger an
+        OSSEC alert)."""
+
+        self._new_codename()
+        self._dummy_submission()
+        resp = self.client.post('/submit', data=dict(
+            msg="This is a test.",
+            fh=(StringIO(''), ''),
+        ), follow_redirects=True)
+        self.assertEqual(resp.status_code, 200)
+        self.assertIn("Thanks! We received your message", resp.data)
+
+        logger.assert_called_once_with(
+            "Couldn't normalize submission "
+            "timestamps (touch exited with 1)"
+        )
+
+    @patch('source.app.logger.error')
+    def test_source_is_deleted_while_logged_in(self, logger):
+        """If a source is deleted by a journalist when they are logged in,
+        a NoResultFound will occur. The source should be redirected to the
+        index when this happens, and a warning logged."""
+
+        codename = self._new_codename()
+        resp = self.client.post('login', data=dict(codename=codename),
+                                follow_redirects=True)
+
+        # Now the journalist deletes the source
+        filesystem_id = crypto_util.hash_codename(codename)
+        crypto_util.delete_reply_keypair(filesystem_id)
+        source = Source.query.filter_by(filesystem_id=filesystem_id).one()
+        db_session.delete(source)
+        db_session.commit()
+
+        # Source attempts to continue to navigate
+        resp = self.client.post('/lookup', follow_redirects=True)
+        self.assertEqual(resp.status_code, 200)
+        self.assertIn('Submit documents for the first time', resp.data)
+        self.assertNotIn('logged_in', session.keys())
+        self.assertNotIn('codename', session.keys())
+
+        logger.assert_called_once_with(
+            "Found no Sources when one was expected: "
+            "No row was found for one()"
+        )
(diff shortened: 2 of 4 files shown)


securedrop/source_app/main.py
```diff
@@ -48,6 +48,7 @@ def create():
         try:
             db_session.commit()
         except IntegrityError as e:
+            db_session.rollback()
```
heartsucker: Good catch.

securedrop/tests/test_source.py
```diff
@@ -83,14 +86,44 @@ def test_generate_already_logged_in(self):
...
             resp = c.get('/generate')
             resp = c.post('/create', follow_redirects=True)
             self.assertTrue(session['logged_in'])
             # should be redirected to /lookup
             self.assertIn("Submit Materials", resp.data)
 
+    @patch('source.app.logger.warning')
+    @patch('crypto_util.genrandomid',
+           side_effect=[overly_long_codename, 'short codename'])
+    def test_generate_too_long_codename(self, genrandomid, logger):
```
heartsucker: 👍

securedrop/tests/test_source.py
```diff
@@ -136,6 +169,12 @@ def test_login_and_logout(self):
             self.assertTrue(not session)
             self.assertIn('Thank you for exiting your session!', resp.data)
 
+    def test_user_must_log_in_for_protected_views(self):
```
heartsucker: 👍

securedrop/tests/test_source.py
```diff
@@ -83,14 +86,44 @@ def test_generate_already_logged_in(self):
...
+        with self.client as c:
+            c.get('/generate')
+
+            # Create a source the first time
+            c.post('/create', follow_redirects=True)
+
+            # Attempt to add the same source
+            c.post('/create', follow_redirects=True)
+            logger.assert_called_once()
+            self.assertIn("Attempt to create a source with duplicate codename",
```
heartsucker: The logic on the endpoint this tests makes me a bit uneasy. First, this is the schema:

```sql
CREATE TABLE sources (
	id INTEGER NOT NULL, 
	filesystem_id VARCHAR(96), 
	journalist_designation VARCHAR(255) NOT NULL, 
	flagged BOOLEAN, 
	last_updated DATETIME, 
	pending BOOLEAN, 
	interaction_count INTEGER NOT NULL, 
	PRIMARY KEY (id), 
	UNIQUE (filesystem_id), 
	CHECK (flagged IN (0, 1)), 
	CHECK (pending IN (0, 1))
);
```

Note that the `journalist_designation` is not unique. I think we need to open an issue for this.

Second, once we *do* add that fix, the exception there will catch both errors and treat them as the same which is a little problematic. I also am not convinced it should blindly continue with logging the user in there. I actually think it should be a 500 error (maybe a 400). In the UI, that endpoint should never produce a codename that already exists. In the (astronomically tiny) chance it does, we don't want to log them in as another source. I think it should bail and then make the source retry.

securedrop/tests/test_source.py
```diff
@@ -304,3 +358,54 @@ def test_login_with_overly_long_codename(self, mock_hash_codename):
             self.assertFalse(mock_hash_codename.called,
                              "Called hash_codename for codename w/ invalid "
                              "length")
+
+    @patch('source.app.logger.warning')
+    @patch('subprocess.call', return_value=1)
+    def test_failed_normalize_timestamps_logs_warning(self, call, logger):
```
heartsucker: 🎉

dachary:
removing myself from the reviewer list. Not because I don't want to ;-) But because @heartsucker is already at it. If you think it deserves double review, I'd be happy to oblige.

securedrop/tests/test_source.py
redshiftzero: Good call 😎  - check out issue #2043 and my comments there re: the uniqueness of `journalist_designation`. 

On your second point, duplicate codenames won't be generated due to the unique constraint set on the `filesystem_id` column (recall that `filesystem_id` is [a hash of `codename`](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/source_app/utils.py#L54)).

redshiftzero:
Thanks for reviewing @heartsucker, answered your comment inline, lemme know your thoughts 😃

securedrop/tests/test_source.py
heartsucker: Yes, that is true, but the code here in `source_app/main.py`:

```python
    @view.route('/create', methods=['POST'])
    def create():
        filesystem_id = crypto_util.hash_codename(session['codename'])

        source = Source(filesystem_id, crypto_util.display_id())
        db_session.add(source)
        try:
            db_session.commit()
        except IntegrityError as e:
            db_session.rollback()
            current_app.logger.error(
                "Attempt to create a source with duplicate codename: %s" %
                (e,))
        else:
            os.mkdir(store.path(filesystem_id))

        session['logged_in'] = True
        return redirect(url_for('.lookup'))
```

Will continue on an `IntegrityError` which means in the small chance a source generates the same codename an existing source, we just log them in. That was what I had minor concern about.

But at the end of the day, this PR improves the existing state of things, and perhaps we should revisit this particular problem in another issue.

heartsucker:
Marking as approved because this PR is strictly positive, and my earlier comments were issues with existing code. Up to @redshiftzero to merge this now or with the changed I mentioned. If no changes are made, the issue I noted should be turned into a ticket so  we can address it later.

securedrop/tests/test_source.py
heartsucker: More notes: this endpoint could check if they're logged in or not, and if they are, it could continue, but if they're not logged in *and* they try to make a duplicate user name, they could bail.

redshiftzero: Hmm yeah, I hear ya. OK - for defense in depth I'm going to make a followup to `abort(500)` after this event is logged and fix the corresponding tests since it's quick and easy

redshiftzero:
Subsequent ticket in #2386 - merging!

PR #2380, opened 2017-09-28
i18n: the order of the accepted languages matters

### LINKED ISSUE #2379: i18n Doesn't respect user locale preferences
# Bug

I accept several languages, and send a header with a different priority. I prefer for example Dutch to German and a typical Accept-Language header here could look like: 
`Accept-Language:en-US,en;q=0.8,nl;q=0.6,fr;q=0.4,de;q=0.2
`
This will *not* display the Dutch interface but the French one. If I send one like:
`Accept-Language: en-US,en;q=0.8,nl;q=0.6,de;q=0.4
`
I would get the German one.

## Steps to Reproduce

Browse to http://source.i18n.securedrop.club/ and send such headers:

GET / HTTP/1.1
Host: source.i18n.securedrop.club
Connection: keep-alive
Pragma: no-cache
Cache-Control: no-cache
Upgrade-Insecure-Requests: 1
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chromium/63.0.XXXXX Safari/537.36
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8
Referer: http://i18n.securedrop.club/
Accept-Encoding: gzip, deflate
Accept-Language: en-US,en;q=0.8,nl;q=0.6,de;q=0.4

dachary:
## Status

Ready for review

## Description of Changes

Fixes #2379.

The language list from the Accept-Languages header is sorted by order
of preference. Store it in a list instead of a set to preserve the
order.

## Testing

It is difficult to verify manually because the order in which the set is stored depends on how the values are hashed. The best next thing is to read and run the test added to tests/test_i18n.py

I implemented the test with values that fail on my machine if the patch is not applied. With a little luck you may have the same on yours.

## Deployment

i18n is not deployed yet.

### DIFF
--- a/securedrop/tests/test_i18n.py
+++ b/securedrop/tests/test_i18n.py
@@ -80,6 +80,13 @@ class TestI18N(object):
                 {{ gettext('code hello i18n') }}
                 ''').strip() == translated_fr
 
+        # https://github.com/freedomofpress/securedrop/issues/2379
+        headers = Headers([('Accept-Language',
+                            'en-US;q=0.6,fr_FR;q=0.4,nb_NO;q=0.2')])
+        with app.test_request_context(headers=headers):
+            assert not hasattr(request, 'babel_locale')
+            assert not_translated == gettext(not_translated)
+
         translated_cn = 'code chinese'
 
         for lang in ('zh-CN', 'zh-Hans-CN'):
--- a/securedrop/i18n.py
+++ b/securedrop/i18n.py
@@ -66,14 +66,14 @@ def get_locale():
     - 'en_US'
     """
     locale = None
-    accept_languages = set()
+    accept_languages = []
     for l in request.accept_languages.values():
         if '-' in l:
             sep = '-'
         else:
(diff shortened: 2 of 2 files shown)


securedrop/tests/test_i18n.py
```diff
@@ -80,6 +80,13 @@ def verify_i18n(self, app):
                 {{ gettext('code hello i18n') }}
                 ''').strip() == translated_fr
 
+        # https://github.com/freedomofpress/securedrop/issues/2379
+        headers = Headers([('Accept-Language',
+                            'en-US;q=0.6,fr_FR;q=0.4,nb_NO;q=0.2')])
+        with app.test_request_context(headers=headers):
+            assert not hasattr(request, 'babel_locale')
+            assert not_translated == gettext(not_translated)
```
redshiftzero: :tada:

Issue #2312, opened 2017-09-14
Make better use of WTForm's validators


redshiftzero:
## Description

Some of the code in our webapp views are reimplementing functionality that exists in [WTForms' validators](https://wtforms.readthedocs.io/en/latest/validators.html). For example see the following code in `journalist.reply()`:

```
msg = request.form['msg']
    # Reject empty replies
    if not msg:
        flash("You cannot send an empty reply!", "error")
        return redirect(url_for('col', filesystem_id=g.filesystem_id))
```

We should instead just use WTForms' `DataRequired()` validator. Similarly, for issues like #2310 we should be using WTForms' existing form validators. See [here](https://github.com/lucyparsons/OpenOversight/blob/develop/OpenOversight/app/auth/forms.py) and [here](https://github.com/lucyparsons/OpenOversight/blob/develop/OpenOversight/app/main/forms.py) for examples of using WTForms validators. 

Note: In the context of #2294, I suggest we have a `forms.py` for each web application. 

## User Stories

As a SecureDrop developer, I don't want to roll my own validation code when it exists in a dependency I am already "using".

As a paranoid SecureDrop developer, I want to easily see what validation is performed on my user-supplied inputs in `forms.py`.

Dalek-Sec:
I will try to work on this.

PR #2388, opened 2017-09-30
Improve login form validation on source interface

### LINKED ISSUE #2376: Unexpected characters in source codename should not cause 500 error
## Description

On the source interface, unexpected characters in the source codename login a cause a `CryptoException` and 500 error.

## Steps to Reproduce

1. Go to `/login` on source interface.
2. Enter '[]' in the source codename.
3. Hit "Continue"

## Expected Behavior

A reasonable flow would be to flash a message informing the user of the expected format of the codename or which characters in the form input were invalid. 

## Actual Behavior

The generic 500 error page is shown due to a `CryptoException`:

screen shot 2017-09-27 at 12 16 36 pm

## Comments

The current logic here is overly complicated: `login` calls `source_app.utils.valid_codename` which calls `crypto_util.hash_codename` which calls `crypto_util.clean`. In `crypto_util.clean`, if there are unexpected characters in the codename _then_ a `CryptoException` will be thrown. We could just handle the `CryptoException` and flash a message. However, we should be validating these forms separately from `crypto_util.py` (see #2312) and _not_ throw `CryptoException`s for form validation errors.

redshiftzero:
## Status

Ready for review

## Description of Changes 

* This PR creates a `forms.py` file on the source interface with the login form defined using WTForm's `FlaskForm`. This form uses WTForm's validators (a dependency we already have installed on the SecureDrop app server, but were only using it for CSRF protection) to validate form inputs. A nice feature of this is that you can just look at that one file to see what validation is performed on form inputs. This is issue #2312. 
* The bug (a 500 error) caused by an invalid character working its way into `crypto_util.py` and throwing a `CryptoException` is fixed by this. Closes #2376.
* For displaying of form validation errors on the source interface login form, I followed the advice in this [guide to UI patterns best practices for input validation](http://uipatterns.io/input-validation). It looks like this:

loginform

Note there's one other form on the source interface (the submission form) but I wanted to handle refactoring that into `forms.py` in a separate PR to keep the changes small here. 

## Testing

Follow the STR in defect #2376 to test this

## Deployment

* `source_app/forms.py` already is added to `.rsync_filter` (since it copies all *.py files under `source_app`) so it will be shipped in the app code
* `source_app/forms.py` is added to the AppArmor rules in this PR

### DIFF
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
@@ -161,6 +161,8 @@
   /var/www/securedrop/source_app/api.pyc rw,
   /var/www/securedrop/source_app/decorators.py r,
   /var/www/securedrop/source_app/decorators.pyc rw,
+  /var/www/securedrop/source_app/forms.py r,
+  /var/www/securedrop/source_app/forms.pyc rw,
   /var/www/securedrop/source_app/info.py r,
   /var/www/securedrop/source_app/info.pyc rw,
   /var/www/securedrop/source_app/main.py r,
--- a/securedrop/sass/source.sass
+++ b/securedrop/sass/source.sass
@@ -261,3 +261,7 @@ div.bubble
     border-color: transparent transparent #fff #fff
     left: 33px
     top: -43px
+
+span.form-validation-error
+  color: red
+  font-size: small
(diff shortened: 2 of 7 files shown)


redshiftzero:
Good catch! I decided to go ahead and fix that color styling issue and squash it into the existing commits. This is one of the new form validation errors tested on Firefox:

screen shot 2017-10-02 at 2 20 35 pm

redshiftzero:
Since my force push dismissed your review @heartsucker, go ahead and :+1: this when you get a chance if all looks good.

PR #2389, opened 2017-10-02
added code owners for app / deployment code

heartsucker:
## Status

Ready for review

## Description of Changes

Set up code owners to allow us to split responsibilities with merging code.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,10 @@
+# application code
+/securedrop/ @redshiftzero @heartsucker @dachary
+
+# ansible / deployment
+/build/         @conorsch @msheiny
+/devops/        @conorsch @msheiny
+/install_files/ @conorsch @msheiny
+/molecule/      @conorsch @msheiny
+/testinfra/     @conorsch @msheiny
+/tails_files/   @conorsch @msheiny


.github/CODEOWNERS
```diff
@@ -0,0 +1,18 @@
...
+/molecule/      @conorsch
+/testinfra/     @conorsch
+/tails_files/   @conorsch
+
+/build/         @msheiny
+/devops/         @msheiny
+/install_files/ @msheiny
+/molecule/      @msheiny
+/tails_files/   @msheiny
+/testinfra/     @msheiny
```
redshiftzero: Hmm it sounds like from this:

> Order is important. The last matching pattern has the most precedence.

(from [this page](https://github.com/blog/2392-introducing-code-owners)) means that this would result in e.g. only @msheiny being requested on a PR that touches `testinfra` since that is the last matching pattern for `/testinfra/`. Maybe we should put @msheiny and @conorsch on the same line such that they are both requested? 

(Similar comment applies to the entire PR, e.g. the `securedrop/*` section should be `@redshiftzero @dachary @heartsucker` on one line)

.github/CODEOWNERS
```diff
@@ -0,0 +1,18 @@
+# application code
+/securedrop/ @redshiftzero
```
redshiftzero: not `securedrop/*`?

.github/CODEOWNERS
```diff
@@ -0,0 +1,18 @@
...
+/securedrop/ @heartsucker
```
redshiftzero: Let's also add @dachary as a code owner for `securedrop/*`

.github/CODEOWNERS
heartsucker: That wouldn't cover things in `source_app` since the don't match the glob. So it has to be `/securedrop/`.

.github/CODEOWNERS
heartsucker: 👍

.github/CODEOWNERS
heartsucker: Ah, you're totally right. Fixing.

PR #2393, opened 2017-10-03
text-align: right for rtl text directions

### LINKED ISSUE #2381: Text is aligned to the wrong side in rtl languages
## Description

Much of the text on the Arabic version of SecureDrop is aligned to the wrong side (left aligned, when it should be right aligned). Needs fixing pre-release:

screen shot 2017-09-28 at 5 14 14 pm

screen shot 2017-09-28 at 5 15 30 pm

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #2381

screenshot_2017-10-03_12-00-31

Uses the pseudo selector `dir()` to correctly align content for languages that are set RTL.

> Note: this only works in TB and FF according to [this](https://css-tricks.com/almanac/selectors/d/dir/).

I think this is a better solution than the trick I mentioned in https://github.com/freedomofpress/securedrop/issues/2381#issuecomment-333780521 on the grounds that it only leads to a tiny increase in the size of the CSS and for now it is a much simpler solution.

## Testing

Setup:

```bash
cd /vagrant/securedrop
mkdir -p translations/ar/LC_MESSAGES
curl http://lab.securedrop.club/bot/securedrop/raw/i18n/securedrop/translations/ar/LC_MESSAGES/messages.po > translations/ar/LC_MESSAGES/messages.po
./manage.py translate --compile
PAGE_LAYOUT_LOCALES='ar' pytest -v --page-layout tests/pages-layout
```

Using Tor Browser: Create a source (in Arabic locale), submit a doc. Create a journalist (in Arabic locale), respond. Check the response as a source. This should display most everything.

### DIFF
--- a/securedrop/sass/_base.sass
+++ b/securedrop/sass/_base.sass
@@ -28,6 +28,9 @@
     text-align: left
     font-size: 30px
 
+    &:dir(rtl)
+      text-align: right
+
   h2
     font-size: 25px
 
@@ -61,6 +64,9 @@
     margin-bottom: 20px
     text-align: left
 
+    &:dir(rtl)
+      text-align: right
+
   .doc-check
     margin-right: 10px
 
@@ -89,11 +95,20 @@
   .pull-left
     float: left
 
+    &:dir(rtl)
+      float: right
+
   .pull-right
     float: right
 
+    &:dir(rtl)
+      text-align: left
+
   #header
     float: left
+
+    &:dir(rtl)
+      float: right
 
     .powered
       margin-top: 10px
@@ -111,6 +126,9 @@
       border-bottom: 1px solid #ddd
       padding: 6px
       text-align: left
+
+      &:dir(rtl)
+        text-align: right
 
       &:last-child
         border-bottom: none
@@ -271,6 +289,9 @@
     text-align: left
     font-size: medium
 
+    &:dir(rtl)
+      text-align: right
+
   .flash
     display: flex
     flex-flow: row nowrap
@@ -348,7 +369,7 @@
 
   .logo
     width: 250px
-    margin-right: 20px
+    margin-right: 9px
     margin-top: 36px
     margin-left: 9px
 
@@ -359,11 +380,17 @@
   .submissions
     padding-left: 0px
 
+    &:dir(rtl)
+      padding-right: 0px
+
   .submission
     background: $color_grey_xlight
     border-bottom: 1px solid #ddd
     padding: 10px
     text-align: left
+
+    &:dir(rtl)
+      text-align: right
 
     &:last-child
       border-bottom: none
--- a/securedrop/sass/_source_index.sass
+++ b/securedrop/sass/_source_index.sass
@@ -22,6 +22,9 @@
       text-align: left
       margin: 0
 
+      &:dir(rtl)
+        text-align: right
+
     .index-row
       display: flex
       justify-content: space-between
--- a/securedrop/sass/journalist.sass
+++ b/securedrop/sass/journalist.sass
@@ -12,6 +12,9 @@
     margin-bottom: 1em
     border-radius: 5px
 
+    &:dir(rtl)
+      text-align: right
+
     .date
       float: right
       background-color: $color_grey_medium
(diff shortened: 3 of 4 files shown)


securedrop/sass/_base.sass
```diff
@@ -348,9 +366,7 @@
 
   .logo
     width: 250px
-    margin-right: 20px
-    margin-top: 36px
-    margin-left: 9px
+    margin-right: 20px 36px 20px auto
```
dachary: You meant **margin** here or is it really **margin-right** ?

dachary:
@heartsucker updated the tests instructions to add

    PAGE_LAYOUT_LOCALES='ar' pytest -v --page-layout tests/pages-layout

which will help verify without navigating. The technical side of the pull request looks good to me, the **dir** selector is handy.

I'll leave the review to Jen though because my Arabic is close to non-existent and I'm likely to miss something.

securedrop/sass/_base.sass
heartsucker: Yep. I did mean that.

securedrop/sass/_base.sass
```diff
@@ -28,6 +28,9 @@
     text-align: left
     font-size: 30px
 
+    &:dir(rtl)
+      text-align: right
+
```
redshiftzero: A fine approach. We can get fancy with SASS variables on a rainy day :+1:

securedrop/sass/_base.sass
```diff
@@ -89,12 +95,18 @@
   .pull-left
     float: left
 
+    &:dir(rtl)
+      float: right
+
   .pull-right
     float: right
```
redshiftzero: `pull-right` also needs:

```sass
    &:dir(rtl)
      float: left
```

securedrop/sass/_base.sass
```diff
@@ -348,9 +366,7 @@
 
   .logo
     width: 250px
-    margin-right: 20px
-    margin-top: 36px
-    margin-left: 9px
+    margin: 20px 36px 20px auto
```
redshiftzero: Hmm, this produces a weird offset in both Arabic and English:

screen shot 2017-10-03 at 1 46 50 pm

Suggestion: 

```
    margin-right: 9px
    margin-top: 36px
    margin-left: 9px
```

This renders well in both English:

screen shot 2017-10-03 at 1 50 45 pm

And Arabic:

screen shot 2017-10-03 at 1 50 36 pm

Other solutions welcome

securedrop/sass/_base.sass
heartsucker: As I mentioned [here](https://github.com/freedomofpress/securedrop/issues/2381#issuecomment-333780521), I think that won't work quite as slick as we need because the variables are static after they're compiled, so we'd still have to use jinja at a minimum to pull down the right CSS. I feel like there has to be a better way to do this, but I could be wrong.

dachary:
@heartsucker  about the error you got

<pre>
            language2words[locale] = open(
>               wordlist_path).read().rstrip('\n').split('\n')
E IOError: [Errno 40] Too many levels of symbolic links: '/vagrant/securedrop/wordlist'
</pre>

could you please `ls -lR /vagrant/securedrop` ? There is a symlink loop but I can't figure out how that could have happened. I'd like to figure that out in case someone else runs into the same problem.

redshiftzero:
I'm seeing this test failure on `develop` only locally using VirtualBox - `ls -lR /vagrant/securedrop` results [here](https://gist.github.com/redshiftzero/dc946b8d5c5a2e3641a94db83ae9d018)

dachary:
I'm seeing a fail with:

<pre>
PAGE_LAYOUT_LOCALES='ar' pytest -v --page-layout tests/pages-layout/test_journalist.py::TestJournalistLayout::test_admin_edit_hotp_secret[ar]
...
  File "/vagrant/securedrop/journalist.py", line 236, in admin_new_user_two_factor
    user=user.username),
KeyError: u'\u0627\u0644\u0645\u0633\u062a\u062e\u062f\u0645'
</pre>

meaning the arabic translation broke a **{user}** placeholder.

heartsucker:
> There is a symlink loop

I fixed that one, so I must have copy/pasted the wrong error.

PR #2426, opened 2017-10-08
Source: only show codename on first session

### LINKED ISSUE #2327: Only show source codename on first session
# Feature request

## Description

A source is shown their codename on the `/lookup` page to help them remember it on their first session. This hint should not be shown again after the source logs out. The reason being is in the future, we maybe remove the actual codename from the cookie to prevent it from being used to login indefinitely (either by encrypting it or using some uuid that maps to the codename per session). If we show this on subsequent logins, an attack could steal the cookie, not learn the codename, but login see the codename, and then login again indefinitely.

Plus, is a source has logged in successfully, they definitely know their codename, and there is little (nothing?) to be gained by showing it again.

squeed:
## Status

Ready for review

## Description of Changes
When a source logs back in, they can can retrieve their codename, which isn't good. It means that if a session is compromised, all future sessions are also compromised.

Instead, only show the codename on first login. Given that the user had to supply their codename for all subsequent sessions, re-displaying it has no value.

Fixes #2327.

## Testing
1. Create a source and submit a document
2. See that your code name is visible in the source "home page"
3. Log out and log back in
4. See that your code name is no longer supplied

## Deployment
No special considerations

### DIFF
--- a/securedrop/source_app/main.py
+++ b/securedrop/source_app/main.py
@@ -38,6 +38,7 @@ def make_blueprint(config):
 
         codename = generate_unique_codename()
         session['codename'] = codename
+        session['new_user'] = True
         return render_template('generate.html', codename=codename)
 
     @view.route('/create', methods=['POST'])
@@ -91,6 +92,7 @@ def make_blueprint(config):
             codename=g.codename,
             replies=replies,
             flagged=g.source.flagged,
+            new_user=session.get('new_user', None),
             haskey=crypto_util.getkey(
                 g.filesystem_id))
 
--- a/securedrop/source_templates/lookup.html
+++ b/securedrop/source_templates/lookup.html
@@ -94,14 +94,17 @@
 
 <hr class="no-line">
 
+{% if new_user %}
 <div class="code-reminder" id="codename-hint">
   <div id="codename-hint-visible">
     <img class="pull-left" src="{{ url_for('static', filename='i/font-awesome/lock-black.png') }}" width="17px" height="20px"> {{ gettext('Remember your codename is:') }}
     <a id="codename-hint-show" class="show pull-right visible-codename" href="#codename-hint-visible">{{ gettext('Show') }}</a>
     <div id="codename-hint-content" class="hidden-codename codename">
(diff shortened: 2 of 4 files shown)


securedrop/source_app/main.py
```diff
@@ -201,7 +203,9 @@ def login():
         if form.validate_on_submit():
             codename = request.form['codename'].strip()
             if valid_codename(codename):
-                session.update(codename=codename, logged_in=True)
+                session.update(codename=codename,
+                               logged_in=True,
+                               new_user=False)
```
heartsucker: I think we even could be lazier here and not set this to `False` ever and just let it be absent in the session. This relates to my other comment.

securedrop/source_app/main.py
```diff
@@ -91,6 +92,7 @@ def lookup():
             codename=g.codename,
             replies=replies,
             flagged=g.source.flagged,
+            new_user=session['new_user'],
```
heartsucker: this should get `session('new_user', None)` so that if it's not present it won't throw a `KeyError`.

securedrop/source_templates/lookup.html
```diff
@@ -94,14 +94,17 @@ <h2 class="headline">{{ gettext('Get Replies') }}</h2>
...
 
+{% if new_user %}
 <div class="code-reminder" id="codename-hint">
   <div id="codename-hint-visible">
     <img class="pull-left" src="{{ url_for('static', filename='i/font-awesome/lock-black.png') }}" width="17px" height="20px"> {{ gettext('Remember your codename is:') }}
     <a id="codename-hint-show" class="show pull-right visible-codename" href="#codename-hint-visible">{{ gettext('Show') }}</a>
     <div id="codename-hint-content" class="hidden-codename codename">
-      <p>{{ codename }} <a id="codename-hint-hide" class="pull-right" href="#codename-hint">{{ gettext('Hide') }}</a></p>
+      <a id="codename-hint-hide" class="pull-right" href="#codename-hint">{{ gettext('Hide') }}</a>
+      <p>{{ codename }}</p>
```
heartsucker: Note: This change makes it possible to double click the text to select all of it without it selecting the "Hide" button too.

dachary:
Since this is a feature change, I think @redshiftzero should be the one to merge.

Issue #2334, opened 2017-09-18
An admin editting their own account from the admin interface should redirect to the nomal self-acocunt-edit page


heartsucker:
# Feature request

## Description

If an admin attempts to navigate to admin edit pages for their own account, they should be redirected to the normal account edit page.

If we implement #2304, then an admin will be able to update their own account without entering their own password. For this reason, among others, we should not try to be DRY with the normal account edit pages / endpoints and the admin account edit pages / endpoints. The logic is fairly simple, and they should be different things entirely. This will also become necessary when we split things into Blueprints with #2294 since these two endpoints will live in different places and keeping them DRY will just lead to spaghetti code.

## User Stories

As an admin, I don't want to accidentally circumvent any account update logic while updating my own account.

As dev, I don't want unnecessary OOP-style indirection and inheritance in my jinja templates.

zenmonkeykstop:
Current functionality is acceptable. JI Admin interface won't be massively overhauled until a potential Qubes version is created.

PR #2417, opened 2017-10-08
Improve the login form usability by making input fields larger

nicolasiensen:
## Status

Ready for review

## Description of Changes

In this commit, we are changing the login form style so the input fields are larger, especially on big screens.

The inputs are set to a maximum width of 445px, the ideal size in order to display all the characters of the password if the user is using diceware passphrases.

The login form is now also mobile-friendly since it has a responsive width and the input fields will adapt to the screen size as well.

Fixes #2288.

## Testing

These are some screenshots of the new login form:

On a laptop screen
securedrop

On Nexus 5X
securedrop nexus 5x

On iPad
securedrop ipad

## Deployment

No special considerations for deployment.

### If you made changes to the app code:

- [x] Unit and functional tests pass on the development VM

### If you made changes to the system configuration:

- ~[Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass~

### If you made changes to documentation:

- ~Doc linting passed locally~

Issue #2288, opened 2017-09-12
Textboxes for admin and journalist logins should be longer


huertanix:
# Feature request
Textboxes for admin and journalist logins should be longer.

## Description
In login pages where long diceware passphrases are expected to be typed in, the textbox for the passphrase should be long enough to fit a diceware passphrases of the word size recommended.

## User Stories
As a journalist or admin, I want to be able to enter my very long diceware passphrase and make sure I didn't miss or misspell any words before logging in and realizing I done goofed, which would make me have to type a very long passphrase in another time.

redshiftzero:
Noting that in the SecureDrop web applications the codename will not be visible since the form input type is `password` - however I think your point still stands, for example if a user is typing a long password it is useful to see the feedback of the number of obfuscated characters increasing. [This blog](https://baymard.com/blog/form-field-usability-matching-user-expectations) echoes @huertanix's advice here and has some other good background on good UX on forms.

Issue #2253, opened 2017-09-05
Stabilize AWS staging builders connectivity during play


msheiny:
# Bug

## Description

We've been seeing frequent time-outs to AWS builders during the middle to the end of a playbook run. Note that we already have ticket #1825 / #2218 to handle logic at the `beginning` of the run but not `mid-run`. 

## Steps to Reproduce

There is no simple way to reproduce this error - it occurs under CircleCI sporadically without warning. Usually a re-run of the job will resolve.

## Expected Behavior

The CI playbook should finish all the way through without connectivity errors.

## Actual Behavior

The CI playbook dies mid-way through (not at any particular task) with the following ansible error:

> `fatal: [app-staging]: FAILED! => {"changed": true, "failed": true, "rc": 1, "stderr": "Connection to ec2-xx-xx-xx-xx.us-west-1.compute.amazonaws.com closed.\r\n", "`

## Comments

There are a few things we can try (sorted from easiest to more time intensive):
* Switching to a different ec2 region
* Attempt to perform a rescue task against the entire playbook if it fails with the connectivity issue
* Switch to a different cloud provider -- we cant really jump up on ec2 instance size due to budget constraints but I've noticed connectivity improvements before between t2.small and the larger sizes (at least on us-west-1). On another provider we benefit as a side-effect because we can get per minute charging since all of our builders are under an hour (where as amazon rounds up to closest hour).

### Timeout reboot

In the output of CircleCI, if it ends with
<pre>
Wait for server to come back. ----------------------------------------- 602.03s
</pre>
it means the server did not come back after 10 minutes.

### Timeout launch

In the output of CircleCI, if 
<pre>
Wait for EC2 instances to boot
</pre>
fails **twice** (there is a  **Destroy EC2 instances (for relaunch)** in between), it means the timeout in securedrop/molecule/aws/aws-launch.yml is not enough.

redshiftzero:
Just saw another AWS connectivity issue in [this build](https://circleci.com/gh/freedomofpress/securedrop/2666):

```
    TASK [Wait for server to come back.] *******************************************
    Friday 06 October 2017  23:38:19 +0000 (0:00:01.187)       0:00:03.889 ********
    ok: [mon-staging -> localhost]
[...]

PR #2430, opened 2017-10-08
Require credentials when changing password

### LINKED ISSUE #2304: User does not need to enter their old password to change their password
## Description

A user can change their password without entering their old password first. 

## Steps to Reproduce

1. Login
2. Go to change password form

## Expected Behavior

User must enter old password and then generate the new password.

## Actual Behavior

User can generate their new password without entering their old one. 

## Comments 

Related: #2300, #2303. This is a low priority to fix, but filing nevertheless.

jerr0328:
Users will have to enter their password and TOTP token when changing their passwords. Admins will also need to do the same for their own accounts, although this isn't specifically validated in the backend (I consider it a sanity check more than security from the admin side).

## Status

Ready for review

## Description of Changes

Fixes #2304 .

- Users must enter their existing password and two-factor code when changing their passwords.
- Admins changing their own password via the Admin console will also be prompted for their current password and token.

## Testing

Besides running unit tests, this can be validated as a user by doing the following:

1. Login as a user
2. Go to account page (click on username in top right).
3. See login failure by failing to provide valid existing password.
4. See additional failure by failing to provide valid two-factor code.
5. Enter valid credentials and notice the password is appropriately changed.

## Deployment

No special deployment considerations

## Checklist

- [X] Unit and functional tests pass on the development VM

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -26,7 +26,8 @@ import store
 import template_filters
 from db import (db_session, Source, Journalist, Submission, Reply,
                 SourceStar, get_one_or_else, LoginThrottledException,
-                PasswordError, InvalidUsernameException)
+                PasswordError, InvalidUsernameException,
+                BadTokenException, WrongPasswordException)
 import worker
 
 app = Flask(__name__, template_folder=config.JOURNALIST_TEMPLATES_DIR)
@@ -142,7 +143,10 @@ def validate_user(username, password, token, error_message=None):
     """
     try:
         return Journalist.login(username, password, token)
-    except Exception as e:
+    except (InvalidUsernameException,
+            BadTokenException,
+            WrongPasswordException,
+            LoginThrottledException) as e:
         app.logger.error("Login for '{}' failed: {}".format(
             username, e))
         if not error_message:
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -131,42 +131,57 @@ def admin_required(func):
     return wrapper
 
 
+def validate_user(username, password, token, error_message=None):
+    """
+    Validates the user by calling the login and handling exceptions
+    :param username: Username
+    :param password: Password
+    :param token: Two-factor authentication token
+    :param error_message: Localized error message string to use on failure
+    :return: Journalist user object if successful, None otherwise.
+    """
+    try:
+        return Journalist.login(username, password, token)
+    except Exception as e:
+        app.logger.error("Login for '{}' failed: {}".format(
+            username, e))
+        if not error_message:
+            error_message = gettext('Login failed.')
+        login_flashed_msg = error_message
+
+        if isinstance(e, LoginThrottledException):
+            login_flashed_msg += " "
+            period = Journalist._LOGIN_ATTEMPT_PERIOD
+            # ngettext is needed although we always have period > 1
+            # see https://github.com/freedomofpress/securedrop/issues/2422
+            login_flashed_msg += ngettext(
+                "Please wait at least {seconds} second "
+                "before logging in again.",
+                "Please wait at least {seconds} seconds "
+                "before logging in again.", period).format(seconds=period)
+        else:
+            try:
+                user = Journalist.query.filter_by(
+                    username=username).one()
+                if user.is_totp:
+                    login_flashed_msg += " "
+                    login_flashed_msg += gettext(
+                        "Please wait for a new two-factor token"
+                        " before trying again.")
+            except:
+                pass
+
+        flash(login_flashed_msg, "error")
+        return None
+
+
 @app.route('/login', methods=('GET', 'POST'))
 def login():
     if request.method == 'POST':
-        try:
-            user = Journalist.login(request.form['username'],
-                                    request.form['password'],
-                                    request.form['token'])
-        except Exception as e:
-            app.logger.error("Login for '{}' failed: {}".format(
-                request.form['username'], e))
-            login_flashed_msg = gettext('Login failed.')
-
-            if isinstance(e, LoginThrottledException):
-                login_flashed_msg += " "
-                period = Journalist._LOGIN_ATTEMPT_PERIOD
-                # ngettext is needed although we always have period > 1
-                # see https://github.com/freedomofpress/securedrop/issues/2422
-                login_flashed_msg += ngettext(
-                    "Please wait at least {seconds} second "
-                    "before logging in again.",
-                    "Please wait at least {seconds} seconds "
-                    "before logging in again.", period).format(seconds=period)
-            else:
-                try:
-                    user = Journalist.query.filter_by(
-                        username=request.form['username']).one()
-                    if user.is_totp:
-                        login_flashed_msg += " "
-                        login_flashed_msg += gettext(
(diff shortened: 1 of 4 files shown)


dachary:
https://circleci.com/gh/freedomofpress/securedrop/2711?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

<pre>
'Timeout when waiting for ec2-54-219-169-171.us-west-1.compute.amazonaws.com:22'})
</pre>

securedrop/journalist_templates/edit_account.html
```diff
@@ -27,14 +27,19 @@
 <p>{{ gettext('SecureDrop now uses automatically generated diceware passwords.') }}</p>
 <p>{{ gettext('Your password will be changed immediately, so you will need to save it before pressing the "Reset Password" button.') }}</p>
 
-{% if user %}
+{% if user and g.user != user %}
```
dachary: Why is **g.user != user** necessary here ?  It's an extra security check ?

securedrop/journalist.py
```diff
@@ -120,38 +120,50 @@ def wrapper(*args, **kwargs):
...
+    :param password: Password
+    :param token: Two-factor authentication token
+    :return: Journalist user object if successful, None otherwise.
+    """
+    try:
+        return Journalist.login(username, password, token)
+    except Exception as e:
+        app.logger.error("Login for '{}' failed: {}".format(
+            username, e))
+        login_flashed_msg = gettext('Login failed.')
```
dachary: In the context of a password change, this error message should be different. Something like **Incorrect password or two-factor code** maybe ?

securedrop/journalist_templates/edit_account.html
jerr0328: It's basically to check if the admin user is updating their own password. I don't consider it so much a security check, but instead a protection from an admin changing their own password instead of another user and getting locked out.

securedrop/journalist.py
jerr0328: I did notice the error wasn't exactly the best. I'm wondering if "Authentication failed" would be better to support both? The other parts about waiting or using a different 2FA code refer to logging in, so I'd like to try to keep it generic (and hopefully help reduce the amount of text that needs to be translated).

I'll change this depending on context (basically passing in the string to the function), and change the rest of the notices to be generic with something like "before trying again" (instead of "before logging in again").

securedrop/journalist_templates/edit_account.html
dachary: Good call ! 👍

dachary:
https://circleci.com/gh/freedomofpress/securedrop/2720?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
<pre>
"Timeout when waiting for ec2-13-57-59-142.us-west-1.compute.amazonaws.com:22"
</pre>

dachary:
Manual tests

* Admin changes its own password
* Admin changes the password of another user
* Regular user changes its own password

dachary:
@jerr0328 this is ready to merge :-) But we need to kick CircleCI again, because of the transient error. You can do that by squashing the two commits into a single one and re-pushing.

securedrop/journalist.py
```diff
@@ -120,38 +120,53 @@ def wrapper(*args, **kwargs):
...
+    Validates the user by calling the login and handling exceptions
+    :param username: Username
+    :param password: Password
+    :param token: Two-factor authentication token
+    :param error_message: Localized error message string to use on failure
+    :return: Journalist user object if successful, None otherwise.
+    """
+    try:
+        return Journalist.login(username, password, token)
+    except Exception as e:
```
heartsucker: I think we can this a bit more specific and only catch the crypto / no result cases. The normal 500 error handler will catch everything else.

This could be `except (InvalidUsernameException, BadTokenException, WrongPasswordException)`.

Sorry I didn't catch this earlier before we did 10,000 CircleCI builds. :/

securedrop/journalist.py
```diff
@@ -394,12 +409,17 @@ def edit_account():
...
 
-@app.route('/account/new-password', methods=['POST'])
+@app.route('/account/new-password', methods=('POST',))
 @login_required
 def new_password():
     user = g.user
-    password = request.form.get('password')
-    _set_diceware_password(user, password)
+    current_password = request.form.get('current_password')
+    token = request.form.get('token')
```
heartsucker: I don't object to this, but I think we might not need the token for them to reset their account. This isn't the usual pattern on other webapps, but also SD is more paranoid. I'll like @redshiftzero to comment on this to see if they have an opinion.

securedrop/journalist_templates/edit_account.html
heartsucker: Nice. I'm pretty this actually resolves #2334.

securedrop/journalist.py
kushaldas: The suggestion from @heartsucker sounds good to me.

securedrop/journalist.py
kushaldas: I would prefer the token should be present to change the user password.

redshiftzero: Yeah I hear you @heartsucker but I agree with @kushaldas - let's leave it in the PR as is with the token required. It's not much of an additional burden since they will need their 2FA token on hand anyway.

securedrop/journalist_templates/edit_account.html
```diff
@@ -27,14 +27,19 @@
...
   {% set password_reset_url = url_for('admin_new_password', user_id=user.id) %}
 {% else %}
   {% set password_reset_url = url_for('new_password') %}
+  <p>{{ gettext('Please enter your current password and two-factor code.') }}</p>
 {% endif %}
 
 <form action="{{ password_reset_url }}" method="post" id="new-password">
   <input name="csrf_token" type="hidden" value="{{ csrf_token() }}">
+  {% if not user or g.user == user %}
+  <input type="password" name="current_password" placeholder="{{ gettext('Current Password') }}">
```
redshiftzero: In UX feedback (see #2288 and associated PR #2417), we've heard that the form fields should be as long as the expected content within the form. You could use the same `login-form` class added in PR #2417 to achieve this in this form.

securedrop/journalist.py
redshiftzero: I agree with you both @heartsucker @kushaldas that we should only catch what we handle here and let 500s occur for any other exceptions. That said, since in this PR the author moved the existing validation code into its own function for reuse, I'm cool with making a followup issue to address improving the exception handling in this function.

heartsucker: Ok, marking this as approved. Whoever merges this should make the follow up issue.

dachary:
https://circleci.com/gh/freedomofpress/securedrop/2838?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

timeout provisioning

#2253

dachary:
@redshiftzero it's ready now :-)

heartsucker:
@redshiftzero Have your issues been addressed? This is passing CI I think it's good to merge.

redshiftzero:
Sorry for the delay - this looks excellent. Merging!

PR #2113, opened 2017-08-15
Tor Apt Repo mirror

msheiny:
## Status

Ready for review 👍 

## Description of Changes

Addresses #2106 , in response to the big ol' #2105 issue of 2017.

Changes proposed in this pull request:

Switches tor apt repo to point from `http://deb.torproject.org/torproject.org` to `http://tor-apt.ops.freedom.press`. Which is a snapshot in time of the upstream tor repo. 

## Testing

This switches the network location of the tor packages pulled in by the securedrop servers. So the big part in testing is to ensure that:
* tor doesnt break 
* securedrop-app deb packages correctly switch the repo to point to our mirror

## Deployment

Any special considerations for deployment? Consider both:

1. `Upgrading existing production instances.` --- the securedrop-app deb package contains basic bash logic run on updates that attempts to modify the tor apt repo location. This needs to be tested more before this PR is pulled in.
2. `New installs.` ---   Do the tor packages get installed and does securedrop onion services come up 👍 

### If you made changes to the app code:

- [ ] Unit and functional tests pass on the development VM

### If you made changes to the system configuration:

- [x] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

Issue #2105, opened 2017-08-14
After unattended update, can't connect to SecureDrop app


sighmon:
# Bug

## Description

Overnight one of the unattended upgrades left the SecureDrop app server unable to be reached from Tor Browser.

## Steps to Reproduce

SecureDrop 0.4.1

1. SecureDrop had been running upgrades flawlessly.
2. An unattended upgrade last night seems to have caused the break.

## Expected Behavior

Can access SecureDrop from Tor Browser.

## Actual Behavior

Unable to connect.

## Comments

* I've tried restarting the servers twice.
* I can still ssh in to both mon & app from the Admin workstation.
* Searching issues I came across #1078

From the OSSEC logs around the time it became unreachable:

```bash
Rule: 100012 fired (level 7) -> "Apparmor denied event"
Portion of the log(s):

Aug 14 XX:00:XX app kernel: [   36.599669] audit: type=1400 audit(XXX): apparmor="DENIED" operation="capable" pid=XXXX comm="apache2" capability=1  capname="dac_override"


Rule: 1002 fired (level 2) -> "Unknown problem somewhere in the system."
Portion of the log(s):

Aug 14 XX:01:XX app Tor[XXXX]: Onion service connection to [scrubbed] failed (connection refused)
```

byeskille:
Can confirm this also affects our setup.
Same situation, we are able to ssh in from the Admin Workstation, but not able to connect to either Source or Journalist Interface.

Checked a couple of the instances listed in the [SecureDrop directory](https://securedrop.org/directory), and many of them are also down.

**Can add that upgrades to tor around the same time might also be involved.**

From the OSSEC log alerts we see changes to tor related files/packages:
```
Rule: 550 fired (level 7) -> Integrity checksum changed.
Portion of the log(s):
Integrity checksum changed for: /etc/apparmor.d/abstractions/tor
--END OF NOTIFICATION

Rule: 2902 fired (level 7) -> New dpkg (Debian Package) installed.
Portion of the log(s):
2017-08-14 XX:36:XX status installed tor:amd64 0.3.0.10-1~trusty+1
--END OF NOTIFICATION

Rule: 2902 fired (level 7) -> New dpkg (Debian Package) installed.
Portion of the log(s):
2017-08-14 XX:36:XX status installed tor-geoipdb:all 0.3.0.10-1~trusty+1
--END OF NOTIFICATION

Rule: 552 fired (level 7) -> Integrity checksum changed again (3rd time).
[...]

PR #2441, opened 2017-10-13
Uses apt mirror for Tor via `securedrop-config` package

### LINKED ISSUE #2106: Proxy and serve up apt updates for all securedrop instances
# Feature request

## Description

Having SD instances go out on their own for apt updates is problematic because it has the side-effect of sometimes introducing critical breaking package updates to the ecosystem*. Instead we should be hosting and proxying updates to SecureDrop instances and performing QA when we see new packages hit upstream.

`*` An example of this happening this week is #2105 

## User Stories

As a securedrop maintainer, I want to ensure that upstream package updates to not break the securedrop system.

conorsch:
## Status

Ready for review.

## Description of Changes

Fixes #2106. Reimplements and supersedes #2113.

* Installs tor packages from `tor-apt.freedom.press` mirror.
* Migrates existing apt configs on running hosts to use _only_ the FPF tor mirror.
* Leverages new `securedrop-config` package to handle changes

## Testing

We need to test clean installs, which CI should do rather well. More important, however, we must ensure the upgrade 0.4.3 prod -> this PR is smooth. Given the imminent codefreeze for 0.4.4, we may have to utilize the QA period to make sure we've addressed the upgrade path sufficiently.

## Deployment

Yes, major implications for deployment. Affects running instances and the method they use to install tor, which is critical for SecureDrop (see #2105 for a reminder). 

### DIFF
--- a/install_files/ansible-base/group_vars/all/securedrop
+++ b/install_files/ansible-base/group_vars/all/securedrop
@@ -51,3 +51,5 @@ appserver_dependencies:
   - apparmor-utils
   - redis-server
   - supervisor
+
+tor_apt_repo_url: http://tor-apt.freedom.press
--- a/install_files/ansible-base/group_vars/all/securedrop
+++ b/install_files/ansible-base/group_vars/all/securedrop
@@ -52,4 +52,4 @@ appserver_dependencies:
   - redis-server
   - supervisor
 
-tor_apt_repo_url: http://tor-apt.freedom.press
+tor_apt_repo_url: https://tor-apt.freedom.press
--- a/install_files/ansible-base/group_vars/securedrop_application_server.yml
+++ b/install_files/ansible-base/group_vars/securedrop_application_server.yml
@@ -6,10 +6,11 @@ ip_info:
 
 ### Used by the install_local_deb_pkgs role ###
 local_deb_packages:
+  - "securedrop-keyring-0.1.1+{{ securedrop_app_code_version }}-amd64.deb"
+  - "securedrop-config-0.1.0+{{ securedrop_app_code_version }}-amd64.deb"
   - "securedrop-ossec-agent-2.8.2+{{ securedrop_app_code_version }}-amd64.deb"
   - "{{ securedrop_app_code_deb }}.deb"
   - "ossec-agent-2.8.2-amd64.deb"
-  - "securedrop-keyring-0.1.1+{{ securedrop_app_code_version }}-amd64.deb"
 
 # Configuring the tor hidden services
 tor_instances:
--- a/install_files/ansible-base/roles/tor-hidden-services/tasks/install_tor.yml
+++ b/install_files/ansible-base/roles/tor-hidden-services/tasks/install_tor.yml
@@ -9,7 +9,8 @@
 
 - name: Setup Tor apt repo.
   apt_repository:
-    repo: deb http://deb.torproject.org/torproject.org {{ ansible_lsb.codename }} main
+    filename: deb_torproject_org_torproject_org
+    repo: deb {{ tor_apt_repo_url }} {{ ansible_lsb.codename }} main
     state: present
   register: add_tor_apt_repo
   tags:
--- a/install_files/ansible-base/roles/tor-hidden-services/tasks/install_tor.yml
+++ b/install_files/ansible-base/roles/tor-hidden-services/tasks/install_tor.yml
@@ -9,7 +9,7 @@
 
 - name: Setup Tor apt repo.
   apt_repository:
-    filename: deb_torproject_org_torproject_org
+    filename: tor_apt_freedom_press
     repo: deb {{ tor_apt_repo_url }} {{ ansible_lsb.codename }} main
     state: present
   register: add_tor_apt_repo
--- a/install_files/ansible-base/roles/tor-hidden-services/tasks/install_tor.yml
+++ b/install_files/ansible-base/roles/tor-hidden-services/tasks/install_tor.yml
@@ -1,8 +1,8 @@
 ---
-- name: Install Tor project GPG signing key.
+- name: Remove Tor project GPG signing key.
   apt_key:
-    state: present
-    data: "{{ lookup('file', 'tor-signing-key.pub') }}"
+    state: absent
+    id: A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89
   tags:
     - apt
     - tor
@@ -17,20 +17,19 @@
     - apt
     - tor
 
-- name: Update apt cache.
-  apt:
-    update_cache: yes
-  when: add_tor_apt_repo.changed
-
+# Remove official Tor apt repo key, so only the SecureDrop Release
+# Signing Key is used, for the FPF tor apt mirror
 - name: Install Tor and Tor keyring packages.
   apt:
-    pkg: "{{ item }}"
-    state: latest
+    name: "{{ item.name }}"
+    state: "{{ item.state }}"
     update_cache: yes
     cache_valid_time: 3600
   with_items:
-    - deb.torproject.org-keyring
-    - tor
+    - name: tor
+      state: present
+    - name: deb.torproject.org-keyring
+      state: absent
   tags:
     - apt
     - tor
--- a/install_files/securedrop-config/DEBIAN/postinst
+++ b/install_files/securedrop-config/DEBIAN/postinst
@@ -13,7 +13,7 @@ case "$1" in
     # Repoint tor repositories to FPF mirror
     apt_security_list="/etc/apt/security.list"
     if [ -f "$apt_security_list" ]; then
-        sed -i 's/deb\.torproject\.org\/torproject\.org/tor-apt.ops.freedom.press/g' "$apt_security_list"
+        sed -i 's/deb\.torproject\.org\/torproject\.org/tor-apt.freedom.press/g' "$apt_security_list"
     fi
     ;;
 
--- a/install_files/securedrop-config/DEBIAN/postinst
+++ b/install_files/securedrop-config/DEBIAN/postinst
@@ -1,0 +1,29 @@
+#!/bin/sh
+# postinst script for securedrop-config
+
+set -e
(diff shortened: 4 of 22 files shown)


dachary:
Testing process I'm following:

* git checkout -b release/0.4.3 origin/release/0.4.3
* vagrant up --no-provision /staging/
* make build-debs

conorsch:
@dachary That sounds great! I'll be doing a similar operation on prod VMs. Noticed that our Ansible logic still installs the tor GPG pubkey, which it shouldn't. But the postinst hooks in `securedrop-config` remove the key, so it's not a large problem. I'll add a commit here to address.

testinfra/common/test_tor_mirror.py
```diff
@@ -0,0 +1,58 @@
...
+    c = Command('apt-key finger')
+    tor_gpg_pub_key_info = """/etc/apt/trusted.gpg.d/deb.torproject.org-keyring.gpg
+-----------------------------------------------------
+pub   2048R/886DDD89 2009-09-04 [expires: 2020-08-29]
+      Key fingerprint = A3C4 F0F9 79CA A22C DBA8  F512 EE8C BC9E 886D DD89
+uid                  deb.torproject.org archive signing key
+sub   2048R/219EC810 2009-09-04 [expires: 2018-08-30]"""
+
+    assert c.rc == 0
+    assert tor_gpg_pub_key_info in c.stdout
```
redshiftzero: @conorsch and I just had a brief discussion about this: this test is clearly passing _but_ we had discussed a workflow where we resign Tor packages on our mirror with the SecureDrop release key, so this key should be gone. If we keep this as is, we'll need to act prior to the `Valid-Until` expires (as in, do QA and update the packages on the mirror before the package expires).

install_files/ansible-base/group_vars/all/securedrop
```diff
@@ -51,3 +51,4 @@ appserver_dependencies:
   - redis-server
   - supervisor
 
+tor_apt_repo_url: http://tor-apt.freedom.press
```
emkll: `apt.freedom.press` supports HTTPS. Why isn't it being used here?

msheiny: so initially there was confusion on my behalf on how to do this with our subdomain and S3 buckets --- but after some much needed time away from that task. I know an excellent way to work around my issue to make it work. So yeah -- i agree, i dont think we should merge until this is https which requires a backend change

conorsch:
Bumped to 0.5.1 pending further QA. The 0.5 code freeze is hours away, so we'll focus on testing what we have—0.5 is already a substantial release—then make a special release for the Tor mirror strategy. New milestone for 0.5.1: https://github.com/freedomofpress/securedrop/milestone/37

conorsch:
Rebased and added a few tweaks to address requested changes. Ready for re-review: @msheiny, @redshiftzero, @emkll, @dachary.

msheiny:
working on reviewing this today

msheiny:
rebased on develop

install_files/ansible-base/roles/tor-hidden-services/tasks/install_tor.yml
```diff
@@ -1,35 +1,35 @@
...
+    id: A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89
   tags:
     - apt
     - tor
 
 - name: Setup Tor apt repo.
   apt_repository:
-    repo: deb http://deb.torproject.org/torproject.org {{ ansible_lsb.codename }} main
+    filename: tor_apt_freedom_press
+    repo: deb {{ tor_apt_repo_url }} {{ ansible_lsb.codename }} main
```
msheiny: This logic should be removed eventually since it is duplicating the logic of the new `securedrop-config` package. I'm not going to block on it for this review and instead I'll spin  up a new ticket.

msheiny:
hoping this is a one-off transient networking error .. it passed successfully on the parallel test 😕 

> fatal: [mon-staging]: FAILED! => {"changed": false, "failed": true, "module_stderr": "Shared connection to 54.219.150.128 closed.\r\n", "module_stdout": "Traceback (most recent call last):\r\n  File \"/tmp/ansible_tZsde9/ansible_module_apt_repository.py\", line 565, in <module>\r\n    main()\r\n  File \"/tmp/ansible_tZsde9/ansible_module_apt_repository.py\", line 553, in main\r\n    cache.update()\r\n  File \"/usr/lib/python2.7/dist-packages/apt/cache.py\", line 440, in update\r\n    raise FetchFailedException(e)\r\napt.cache.FetchFailedException: W:Failed to fetch https://tor-apt.freedom.press/dists/trusty/main/binary-amd64/Packages  HttpError403\r\n, E:Some index files failed to download. They have been ignored, or old ones used instead

install_files/ansible-base/group_vars/securedrop_application_server.yml
```diff
@@ -6,10 +6,11 @@ ip_info:
 
 ### Used by the install_local_deb_pkgs role ###
 local_deb_packages:
+  - "securedrop-keyring-0.1.1+{{ securedrop_app_code_version }}-amd64.deb"
+  - "securedrop-config-0.1.0+{{ securedrop_app_code_version }}-amd64.deb"
   - "securedrop-ossec-agent-2.8.2+{{ securedrop_app_code_version }}-amd64.deb"
   - "{{ securedrop_app_code_deb }}.deb"
   - "ossec-agent-2.8.2-amd64.deb"
-  - "securedrop-keyring-0.1.1+{{ securedrop_app_code_version }}-amd64.deb"
```
msheiny: hah! the tables have turned! not in alphabetical order 😛 !

install_files/securedrop-ossec-server/DEBIAN/control
```diff
@@ -6,7 +6,7 @@ Homepage: https://securedrop.org
 Package: securedrop-ossec-server
 Version: 2.8.2+0.5
 Architecture: amd64
-Depends: ossec-server,securedrop-keyring
+Depends: ossec-server,securedrop-keyring,securedrop-config
```
msheiny: soooo as a totally side discussion in my PR for #2592 - i made it a point not to install tor on ossec if one enables tor over local network. I don't know if that was the right call (im on the fence) - I'm just making a note that here you are making this package a pre-requisite for ossec server where in my PR I assume tor might not be installed on the mon server... Once again, totally outside the scope of this ticket but I would like to convo about what should be the right call there.

install_files/securedrop-config/DEBIAN/postinst
```diff
@@ -0,0 +1,29 @@
...
+case "$1" in
+    configure)
+
+    # Ensure official Tor repo entry is removed, so that only FPF mirror is used.
+    rm -f /etc/apt/sources.list.d/deb_torproject_org_torproject_org.list
+
+    # Repoint tor repositories to FPF mirror
+    apt_security_list="/etc/apt/security.list"
+    if [ -f "$apt_security_list" ]; then
+        sed -i 's/deb\.torproject\.org\/torproject\.org/tor-apt.freedom.press/g' "$apt_security_list"
```
msheiny: won't this add a duplicate apt source once the `securedrop-config` installs ?
* a copy in `/etc/apt/security.list` and
* a revision in `/etc/apt/sources.list.d/tor.apt.freedom.press.list` 

I thought apt would throw a warning for this type of thing ... I have to confirm this scenario actually exists first though

install_files/ansible-base/roles/tor-hidden-services/tasks/install_tor.yml
```diff
@@ -1,35 +1,35 @@
 ---
-- name: Install Tor project GPG signing key.
+- name: Remove Tor project GPG signing key.
   apt_key:
-    state: present
-    data: "{{ lookup('file', 'tor-signing-key.pub') }}"
+    state: absent
+    id: A3C4F0F979CAA22CDBA8F512EE8CBC9E886DDD89
```
msheiny: I think this removal logic should also be part of the effort of #2764 but I'm not blocking this marge on that . Good due diligence on removing it here thru ansible for now 👍

install_files/ansible-base/group_vars/securedrop_application_server.yml
conorsch: Actually the order here is important, since we're installing via `dpkg -i` there's no smart dependency resolution, so we must install the dependencies first, _then_ the packages that depend on them!

install_files/securedrop-config/DEBIAN/postinst
conorsch: `/etc/apt/security.list` is the list of repos referenced by `cron-apt` for nightly upgrades. The `/etc/apt/sources.list.d/tor.apt.freedom.press.list` location is where system apt finds the repository information for normal apt install tasks. You'll also see the FPF repo (`apt.freedom.press`) duplicated in `security.list`.

install_files/securedrop-ossec-server/DEBIAN/control
conorsch: Hm, interesting. To be clear, the `securedrop-config` package does not actually install tor, it just configures the FPF apt mirror _for_ tor. So there shouldn't be any conflict there.

But the discrepancy you point out is the direct result of trying to support customizations of deployments, so we haven't seen the last of that (related: #1129).

msheiny: > Hm, interesting. To be clear, the securedrop-config package does not actually install tor, it just configures the FPF apt mirror for tor. So there shouldn't be any conflict there.

Yeah thats a very good point but just something I figured I'd point out since it caught my attention. 

Regarding #1129 though-- i really dont want administrators to make those kinds of unique changes. If they are doing that, we need to try to officially support the features they are trying to add.

install_files/ansible-base/group_vars/securedrop_application_server.yml
msheiny: ...... but but..... dang it

Issue #2217, opened 2017-08-29
Move linting logic from Travis to Circle CI


conorsch:
# Feature request

## Description

The Travis CI config runs the linting checks as a final step, after validating the dev VM setup. We should consider moving the linting checks to Circle CI, since we have workflows as of #2149. Running all the linting checks first will keep CI runs briefer by exiting on linting errors and skipping the long-running staging test install.

## User Stories

As a SecureDrop developer, I want CI runs to be reliable, informative, and snappy. Failing fast is appropriate for linting errors, since I should be linting locally rather than trusting CI to do it for me.

redshiftzero:
Done in #2588

PR #2442, opened 2017-10-13
Stabilize aws round105

### LINKED ISSUE #2253: Stabilize AWS staging builders connectivity during play

msheiny:
## Status

Ready for review

## Description of Changes

Fingers crossed this will fix #2253 

* Utilize an FPF maintained AMI that pre-bakes some of the system level dependencies. Should only shave about 2 minutes off build time. 
* ~Utilize an EC2 reboot versus a system level reboot. I'm hoping this will solve stability issues.~ Completely disable reboot and mark XFAIL on a few iptables and ossec testinfra tests. Will have to solve this completely in another ticket. In order to move forward and unblock the community this was the right call.
* Rips out dynamic security group generation in-place of a static one

## Testing

Does this pass CI ? Try running again to confirm it's "stable"

## Deployment

Nope - only affects CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,7 @@ jobs:
       - run:
           name: Ensure environment torn down
           command: molecule destroy -s aws
-          when: always
+          when: on_fail
 
       - store_test_results:
           path: /root/sd/junit
--- a/molecule/aws/create.yml
+++ b/molecule/aws/create.yml
@@ -37,27 +37,6 @@
           vpc-id: "{{ aws_ec2_vpc_id }}"
       register: vpc_subnet_result
 
-    - name: Create custom temporary security_group for CI hosts.
-      ec2_group:
-        name: "{{ job_id }}"
-        description: Temporary rules for CI
-        region: "{{ aws_ec2_ci_region }}"
-        vpc_id: "{{ aws_ec2_vpc_id }}"
-        rules:
-          - proto: tcp
-            from_port: 22
-            to_port: 22
-            cidr_ip: "{{ ci_box_ip  }}/32"
-          - proto: tcp
-            from_port: 1514
-            to_port: 1515
-            cidr_ip: "{{ vpc_subnet_result.subnets[0].cidr_block }}"
-          - proto: udp
-            from_port: 1514
-            to_port: 1515
-            cidr_ip: "{{ vpc_subnet_result.subnets[0].cidr_block }}"
-      register: group_create_result
-
     - include: aws-launch.yml
 
     - name: Populate instance config dict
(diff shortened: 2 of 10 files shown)


molecule/aws/create.yml
```diff
@@ -105,6 +73,7 @@
     aws_ec2_ci_region: "{{ ansible_env.CI_AWS_REGION | default('us-west-1') }}"
     aws_ec2_vpc_id: "{{ ansible_env.CI_AWS_VPC_ID | mandatory }}"
     aws_ec2_boot_timeout: "500"
+    aws_ec2_sg_id: "sg-3b2e5f5d"
```
kushaldas: Having a proper security group for testing is a very good thing :) 💯

msheiny:
Okay now we are getting greens --> https://circleci.com/gh/freedomofpress/securedrop/2927
Re-running to see if that was a fluke

msheiny:
Still was catching failures after a re-run, decided to bump the kernel in the base image to see if that would resolve our intermittent issues. Got one pass in 2931 ... re-running to confirm back to back success.

msheiny:
AWwwwwwwwwwwww yay. I think stability is way improved with the newer kernels. I'm making new tickets to track adding back in reboot and stripping XFAIL right now.

.circleci/config.yml
```diff
@@ -69,6 +69,4 @@ workflows:
   securedrop_ci:
     jobs:
       - docs-lint
-      - staging-test:
-          requires:
-            - docs-lint
+      - staging-test
```
redshiftzero: Why the removal of `docs-lint`? Do we want `staging-test` to continue if doc linting fails?

msheiny: I was trying to shave 2 minutes to be honest ... i mean, you are right... we dont necessarily want staging-test to continue if doc linting fails.... dang it @redshiftzero and your sound logic

conorsch: Agreed with @redshiftzero, let's keep it. We're considering moving all the linting logic to non-AWS-ified Circle CI anyway (#2217).

msheiny: Okay I re-based and removed the commit - to restore previous test dependency flow. 👍

PR #2463, opened 2017-10-22
removed asserts from main code and added explicit exceptions

### LINKED ISSUE #2429: Don't use asserts in main codebase
# Feature request

## Description

We use asserts for critical codepaths and it could accidentally be optimized away in the future. We should use `if condition: raise Exception(msg)` instead.

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #2429

Removes asserts and uses exceptions to prevent the asserts from ever being optimized away.

## Testing

`pytest -vx tests` like usual.

### DIFF
--- a/securedrop/i18n.py
+++ b/securedrop/i18n.py
@@ -49,7 +49,11 @@ def setup_app(app, translation_dirs=None):
     app.config['BABEL_TRANSLATION_DIRECTORIES'] = translation_dirs
 
     babel = Babel(app)
-    assert len(list(babel.translation_directories)) == 1
+    if len(list(babel.translation_directories)) != 1:
+        raise AssertionError(
+            'Expected exactly one translation directory but got {}.'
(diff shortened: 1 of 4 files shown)


securedrop/i18n.py
```diff
@@ -49,7 +49,9 @@ def setup_app(app, translation_dirs=None):
     app.config['BABEL_TRANSLATION_DIRECTORIES'] = translation_dirs
 
     babel = Babel(app)
-    assert len(list(babel.translation_directories)) == 1
+    if len(list(babel.translation_directories)) != 1:
+        raise AssertionError('Multiple babel translation directories found.')
```
dachary: It could also be an empty list :-) What about:
<pre>
raise AssertionError('Expected exactly one translation directory but got {}.'.format(babel.translation_directories))
</pre>

heartsucker: Fixed.

PR #2469, opened 2017-10-22
implement jinja whitespace control in html templates

### LINKED ISSUE #2413: Strip unnecessary whitespace from templates
# Feature request

## Description

In relation to #1152, we can marginally improve the load speed on pages by using [whitespace control](http://jinja.pocoo.org/docs/dev/templates/#whitespace-control) in our jinja templates. This will only provide a small decrease in pages sizes, but every byte counts and it's relatively easy to implement.

## User Stories

As a source, I want my page to load 3 milliseconds faster.

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #2413.

Strips whitespace when templates are rendered.

## Testing

`pytest -vx tests`. Also, open any page and verify the rendered template has no leading whitespace where a `{% block foo %}` would be.

### DIFF
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -37,6 +37,8 @@ i18n.setup_app(app)
 
 assets = Environment(app)
 
+app.jinja_env.trim_blocks = True
+app.jinja_env.lstrip_blocks = True
 app.jinja_env.globals['version'] = version.__version__
 if getattr(config, 'CUSTOM_HEADER_IMAGE', None):
     app.jinja_env.globals['header_image'] = config.CUSTOM_HEADER_IMAGE
--- a/securedrop/source_app/__init__.py
+++ b/securedrop/source_app/__init__.py
@@ -38,6 +38,8 @@ def create_app(config):
 
     i18n.setup_app(app)
 
+    app.jinja_env.trim_blocks = True
+    app.jinja_env.lstrip_blocks = True
     app.jinja_env.globals['version'] = version.__version__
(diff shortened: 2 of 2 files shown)


dachary:
Wouldn't it be easier to set the [options globally](http://jinja.pocoo.org/docs/dev/api/#jinja2.Environment) ? Or are there specific cases where we do not want spaces to be trimmed [as described in the documentation](http://jinja.pocoo.org/docs/dev/templates/#whitespace-control) ?
<pre>
app.jinja_env.trim_blocks = True
app.jinja_env.lstrip_blocks = True
</pre>

heartsucker:
```
The lstrip_blocks option can also be set to strip tabs and spaces from the beginning of a line to the start of a block. (Nothing will be stripped if there are other characters before the start of the block.)
```

I hate Python docs because it's not clear this is a global option and not "configured by adding `-` at the start of each block."

dachary:
>  it's not clear

I also found the docs to be slightly confusing regarding the algorithm implemented. I was worried it would consume too much space on occasion. However it looks like the two options **trim_blocks** and **lstrip_blocks** do what is intended in this PR with no risk for the HTML produced because:

* it only trims the newline immediately following a **{{ }}**
* it only trims the leading white space preceeding a **{{ }}** with the equivalent of the regexp **^\s***. Meaning it will only trim white space if there is only white space from the beginning of the line to the beginning of the **{{ }}**

Do you have the same understanding of the slightly confusing documentation ?

PR #2481, opened 2017-10-24
Let's target 18.04 and skip 16.04 for the upgrade

KwadroNaut:
## Status

Ready for review

## Description of Changes

Fixes expectations.

Changes proposed in this pull request are tiny.

## Testing

Look up which get released when, and realize that putting effort into 18.04 is more fruitful in the long term.

## Deployment

1. Upgrading existing production instances: skipping is going to be though, if intermediate upgrade is necessary needs to be tested.


redshiftzero:
Let's discuss in #1530 @KwadroNaut

PR #2488, opened 2017-10-30
Underline links in sources list

### LINKED ISSUE #2130: UX: Journalists often click the checkbox to try to get to the source page
## Description

Noticed this UX issue watching first-time SecureDrop users at an on-site: journalists often will try to click on the checkbox to the _left_ of the source codename to try to navigate to the individual source page:

ux-checkbox

We need to make it more clear that the source codename is the element they should be clicking on to navigate to the individual source page (relevant: designs in #1540).

daonb:
## Status

Ready for review

## Description of Changes

Fixes #2130 

## Testing

This is a UX enhancement so testing is left to the user.

### Procedure

- Browser to http://127.0.0.1:8081/ & login
- Look at the sources list (if you have no sources, create some)
- Chose a source from the list in random
- Click it

### Pass/Fail

The test fails if you clicked the check-box the the left of the name.

### DIFF
--- a/securedrop/sass/_base.sass
+++ b/securedrop/sass/_base.sass
@@ -157,7 +157,6 @@
 
         a
           color: #333
-          text-decoration: none
           font-weight: bold
           margin: 5px
           display: inline-block


securedrop/sass/_base.sass
```diff
@@ -157,7 +157,6 @@
 
         a
           color: #333
-          text-decoration: none
```
dachary: Verified the **designation** class is only used in `securedrop/journalist_templates/_source_row.html` and the only anchor in the inner HTML is the link that is modified. There is no risk this change can have a side effect on the look of other links in the source or journalist interface. 
Note to self: this chunk of sass belongs to the journalist interface and should not be in _base.sass because it shares nothing with the source interface.

dachary:
@daonb I'll wait a few days before merging, just in case someone has comments on this.

heartsucker:
This is mergeable, but the issue @ninavizz was addressing in #2130 was that we should use CSS tricks to use underlines on links that look like they do in the New Yorker.

screenshot_2017-10-31_15-50-59

If you look very carefully, the `p` (and `y`, `j`, `q`) all hover above the underline.

```css
.ArticleBody__link___1FS03 {
    margin-left: 1px;
    text-shadow: 0 1px 0 #fff,0 2px 0 #fff,0 -1px 0 #fff,0 -2px 0 #fff,-1px 1px 0 #fff,-1px 2px 0 #fff,1px 1px 0 #fff,1px 2px 0 #fff,-1px 0 0 #fff,0 -3px 0 #fff;
    background: linear-gradient(#fff,#fff),linear-gradient(#fff,#fff),linear-gradient(#000,#000);
        background-repeat: repeat, repeat, repeat;
        background-position-x: 0%, 0%, 0%;
        background-position-y: 0%, 0%, 0%;
        background-size: auto auto, auto auto, auto auto;
    background-position: 0 87%,100% 87%,0 92%;
    background-repeat: no-repeat,no-repeat,repeat-x;
    background-size: .05em 1px,.05em 1px,1px 1px;
    position: relative;
    text-decoration: none;
    z-index: 0;
}
```

dachary:
Oh, indeed :-) I did not notice that ! Should I make an issue about this or does it exist already ?

heartsucker:
Ah whoops, this *does* fix #2130. I got some tabs mixed up. I thought she opened a ticked for this, but maybe it's buried in something else. Or we rejected it. Either way I can't find it right now. Let's pretend I didn't say anything until someone complains about underlines and then we'll revisit it.

PR #2494, opened 2017-10-31
Fixes 2311, validates 20 bytes Yubikey secret

### LINKED ISSUE #2311: Poor input validation on Yubikey HOTP setup
# Bug

## Description

When configuring a Yubikey for HOTP 2FA, the `./manage.py` script does not valid input properly. 
## Steps to Reproduce

For instance, a hardware Yubikey with secret key:

> cb a0 5f ad 41 a2 ff 4e eb 53 56 3a 1b f7 23 2e ce fc dc d7

that entire secret should be pasted as-is into the prompt **Please configure your YubiKey and enter the secret:**. However, if you make a mistake copy/pasting, and truncate the string, e.g.:

> cb a0 5f ad 41 a2 ff 4e eb 53 56 3a 1b f7

Then the script will still accept the bad value, and logins will be impossible with the newly created account. 
## Expected Behavior

`./manage.py` script fails if HOTP secret input is bad, forcing re-entry.

## Actual Behavior

`./manage.py` shows a lackadaisical attitude toward HOTP secret structure, then staunchly refuses logins using the Yubikey.

## Comments

Logs show, predictably, a bad token error:

>  ERROR in journalist: Login for 'conor4' failed: invalid token

kushaldas:
## Status

Ready for review

## Description of Changes

Fixes #2311 

Checking the length of the secret (20 bytes) before accepting a secret key in the manage.py

## Testing

Try to add a new user with an invalid key (length wise).

## Deployment

2. New installs.

### DIFF
--- a/securedrop/manage.py
+++ b/securedrop/manage.py
@@ -159,6 +159,13 @@ def _add_user(is_admin=False):
             otp_secret = raw_input(
                 "Please configure this user's YubiKey and enter the secret: ")
             if otp_secret:
+                tmp_str = otp_secret.replace(" ", "")
+                if len(tmp_str) != 40:
+                    print("The length of the secret is not correct. "
+                          "Expected 40 characters, but received {0}. "
+                          "Try again.".format(len(tmp_str)))
+                    continue
+            if otp_secret:
                 break
 
     try:
--- a/securedrop/tests/test_manage.py
+++ b/securedrop/tests/test_manage.py
@@ -21,6 +21,10 @@ import utils
 from db import Journalist, db_session
 
 
+YUBIKEY_HOTP = ['cb a0 5f ad 41 a2 ff 4e eb 53 56 3a 1b f7 23 2e ce fc dc',
+                'cb a0 5f ad 41 a2 ff 4e eb 53 56 3a 1b f7 23 2e ce fc dc d7']
+
+
 class TestManagePy(object):
     def test_parse_args(self):
         # just test that the arg parser is stable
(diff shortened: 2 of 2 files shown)


kushaldas:
Should not push code without enough coffee.

securedrop/manage.py
```diff
@@ -159,6 +159,12 @@ def _add_user(is_admin=False):
             otp_secret = raw_input(
                 "Please configure this user's YubiKey and enter the secret: ")
             if otp_secret:
+                tmp_str = otp_secret.replace(" ", "")
+                if len(tmp_str) != 40:
+                    print("The length of the secret is not correct. "
+                          "Try again.")
```
dachary: It would be very useful to mention the length of the secret provided and the expected length.

kushaldas: Will add that.

dachary:
would you mind squashing the commits ? it would make the git history easier to read.

securedrop/manage.py
```diff
@@ -159,6 +159,13 @@ def _add_user(is_admin=False):
             otp_secret = raw_input(
                 "Please configure this user's YubiKey and enter the secret: ")
             if otp_secret:
+                tmp_str = otp_secret.replace(" ", "")
+                if len(tmp_str) != 40:
+                    print("The length of the secret is not correct. "
+                          "Expected 40 characters, but, received {0}. "
```
redshiftzero: Nit: `, but,` -> `but`

Issue #2496, opened 2017-10-31
OSSEC syscheck does not monitor integrity of /boot


emkll:
# Bug

## Description

OSSEC file monitoring utility `syscheck` does not monitor files in `/boot/`

## Steps to Reproduce

Observe the configuration in `var/ossec/etc/ossec.conf` or syscheck logs of a given host in `/var/ossec/queue/syscheck/syscheck/`

## Expected Behavior

`var/ossec/etc/ossec.conf` should contain `/boot` in list of directories.

## Actual Behavior

`var/ossec/etc/ossec.conf` does not contain `/boot` in list of directories.

PR #2498, opened 2017-11-01
Modify ossec syscheck configuration to run syscheck every 7h and scan /boot

### LINKED ISSUE #2495: OSSEC syscheck does not reliably updated hashes of files on disk
# Bug
OSSEC syscheck does not reliably calculate hashes of files on the system. This makes comparing hashes of files against known good files more difficult.

## Description
When `syscheck` detects a change in the file, it attempts to hash it. However, at times it fails and returns an error like this one, from `/var/ossec/logs/ossec.log`:

```70/01/01 00:00:00 ossec-syscheckd: ERROR: Unable to run diff for /usr/bin/something```

Hash of file `/usr/bin/something` is never updated in the `syscheck` database.

## Steps to Reproduce

Run a SecureDrop instance for a sufficient amount of time to catch these types of errors in `/var/ossec/logs/ossec.log`.

## Expected Behavior
All files should be hashed as they are changed.

## Actual Behavior
Not all files are hashed as they are changed.

## Comments
`syscheck` allows periodic scanning of the filesystem, updating the `syscheck` database with hashes of all files that are tracked.

### PR DESCRIPTION
**Pull request by emkll**
## Status

Ready for review

## Description of Changes

Fixes #2495 , #2496 .

1. Modified ossec syscheck config to run every 7h to ensure files are properly scanned.
2. Modify ossec syscheck config to monitor integrity of files in /boot.
3. Modified infra tests for mon similar to https://github.com/freedomofpress/securedrop/pull/2490

## Testing

- Ensure syscheck scans every 7h by observing logs in `/var/ossec/logs/ossec.log`
- Ensure syscheck scans files in `/boot` by observing logs in:
    -  `/var/ossec/queue/syscheck/syscheck` for `mon`
    - `/var/ossec/queue/syscheck/($APP_HOST)$APP_IP->syscheck` for `app`
- Ensure the above works when updating existing SecureDrop installs (TODO)

## Deployment

New SecureDrop installs will use this updated configuration upon install.
Existing SecureDrop installs (this has not yet been tested) will need the configuration to be updated via apt.

- :white_check_mark:  [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

### DIFF
--- a/install_files/securedrop-ossec-agent/var/ossec/etc/ossec.conf
+++ b/install_files/securedrop-ossec-agent/var/ossec/etc/ossec.conf
@@ -5,7 +5,7 @@
 
   <syscheck>
     <alert_new_files>yes</alert_new_files>
-
+    <frequency>25200</frequency>
     <!-- Directories to check  (perform all possible verifications) -->
     <directories realtime="yes" check_all="yes" report_changes="yes">/etc,/usr/bin,/usr/sbin</directories>
     <directories realtime="yes" check_all="yes" report_changes="yes">/bin,/sbin</directories>
@@ -15,6 +15,8 @@
     <directories realtime="yes" check_all="yes" report_changes="yes">/var/lib/securedrop</directories>
     <directories realtime="yes" check_all="yes" report_changes="yes">/var/lib/tor/services/hostname</directories>
     <directories realtime="yes" check_all="yes" report_changes="yes">/var/lib/tor/lock</directories>
+    <directories realtime="yes" check_all="yes" report_changes="yes">/boot</directories>
+
 
     <!-- Files/directories to ignore -->
     <ignore>/var/lib/tor/services/source/private_keys</ignore>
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: install_files/securedrop-ossec-agent/var/ossec/etc/ossec.conf [hunk 1]
```diff
@@ -5,7 +5,7 @@
 
   <syscheck>
     <alert_new_files>yes</alert_new_files>
-
+    <frequency>25200</frequency>
```
redshiftzero: These changes look good - but why every 7 hours?

emkll: [Syscheck frequency default is 6h](https://ossec-docs.readthedocs.io/en/latest/manual/syscheck/) and incorrectly assumed that daily update/reboot would have interrupted the scans. However, it appears that production deployments of ossec do syschecks every ~30 mins. Syscheck frequency is therefore likely not the cause of #2495 . Will close this PR and scope one just for #2496.

PR #2502, opened 2017-11-01
Improve form validation on new user creation in journalist interface

### LINKED ISSUE #2500: Poor input validation on OTP secret field in journalist interface
## Description

500 error thrown due to lack of validation in new user form

Same issue as #2311, except that issue was for `manage.py`

## Steps to Reproduce

0. Sign in as administrator on journalist interface
1. Try to create new user using a Yubikey and provide a secret with an odd length

## Expected Behavior

Happy error message is shown indicating that the expected length of the secret is 40 characters

## Actual Behavior

screen shot 2017-10-31 at 3 01 26 pm

## Comments

Simple fix, check length of string and flash an appropriate error message

redshiftzero:
## Status

Ready for review

## Description of Changes

Closes #2500. Progress towards #2312. 

 * Use WTForms to validate new user form. This fixes #2500, a 500 error due to poor form validation on the journalist interface.
 * Adds regression test and some other unit tests

## Testing

Do tests pass? If you want to verify manually as well, you can try the Yubikey string '123' to make sure that a 500 error is _not_ thrown. You should see error messages under the relevant form fields if you provide e.g. an OTP secret that is too short:

screen shot 2017-11-01 at 11 09 18 am

## Deployment

The code changed here will be deployed in the `securedrop-app-code` deb package without issue.


redshiftzero:
Also if the journalist refactor is ready to review in the next couple days we can merge that first as I'm happy to rebase this PR :angel:

dachary:
@redshiftzero gentle reminder that it needs rebase

redshiftzero:
Thanks @dachary - rebased on latest `develop` :)

securedrop/journalist_app/forms.py
```diff
@@ -2,8 +2,34 @@
...
+        raise ValidationError(gettext('Field must be 40 characters long.'))
+
+
+class NewUserForm(FlaskForm):
+    username = TextField('username', validators=[
+        InputRequired(message=gettext('This field is required.')),
+        Length(min=Journalist.MIN_USERNAME_LEN,
+               message=gettext('Field must be at least {min_chars} '
+                               'characters long.'.format(
+                                   min_chars=Journalist.MIN_USERNAME_LEN))),
```
dachary: Same as above (i.e. hint about the number of characters according to the program).

securedrop/journalist_templates/admin_add_user.html
```diff
@@ -6,17 +6,39 @@
 
 <form method="post">
   <input name="csrf_token" type="hidden" value="{{ csrf_token() }}">
-  <input name="password" type="hidden" value="{{ password }}">
-  <p><input name="username" type="text" placeholder="{{ gettext('Username') }}"></p>
+  <p>
+    {{ form.password(value=password, id=False) }}
```
dachary: I'm curious about when an error can happen with the password field ? It has been generated and unless the user deliberately modifies the value (debugger etc.), it looks like it is going to always be good.

securedrop/journalist_templates/admin_add_user.html
```diff
@@ -6,17 +6,39 @@
...
+    <br>
+    {% for error in form.password.errors %}
+      <span class="form-validation-error">{{ error }}</span>
+    {% endfor %}
+  </p>
+
+  <p>
+    {{ form.username(placeholder=gettext('Username')) }}
```
dachary: **id=False** ? I don't see why it would hurt to have an id but ... since the password above has **id=False** this would be consistent don't you think ?

securedrop/journalist_templates/admin_add_user.html
```diff
@@ -6,17 +6,39 @@
...
   </p>
   <p>
-    <input name="is_hotp" id="is-hotp" type="checkbox">
+    {{ form.is_hotp(id="is-hotp") }}
     <label for="is_hotp">{{ gettext("Is using a YubiKey [HOTP]") }} </label>
-    <input name="otp_secret" type="text" placeholder="{{ gettext('HOTP Secret') }}" size="60"><br>
+    {{ form.otp_secret(placeholder=gettext('HOTP Secret'), size="60") }}
+    <br>
+    {% for error in form.is_hotp.errors + form.otp_secret.errors %}
+      <span class="form-validation-error">{{ error }}</span>
```
dachary: It looks like the display of multiple errors could be confusing because they will show on the same line. Maybe the **<br>** should be in the **for** loop ?

securedrop/journalist_templates/admin_add_user.html
redshiftzero: yep, I generally think it's good to have `id` defined for functional testing, I just set the password `id` to false above in the one field as by default WTForms would use `id="password"` which would create a duplicate `id` (`id="password"` is used elsewhere in the file).

securedrop/journalist_templates/admin_add_user.html
redshiftzero: Ah, true - removed

securedrop/journalist_templates/admin_add_user.html
redshiftzero: Multiple errors will render like this now: 
screen shot 2017-11-10 at 10 32 43 pm

redshiftzero:
Thanks for review! Fixed, squashed, repushed.

dachary:
@redshiftzero ready to merge, I assume the circleci failure is because of the **testinfra/requirements.txt** file that was in the way.

securedrop/journalist_app/forms.py
```diff
@@ -2,8 +2,44 @@
...
-from wtforms import TextAreaField
-from wtforms.validators import InputRequired
+from wtforms import (TextAreaField, TextField, BooleanField, HiddenField,
+                     ValidationError)
+from wtforms.validators import InputRequired, Optional
+
+from db import Journalist
+
+
+def otp_secret_validation(form, field):
```
heartsucker: We could regex this to ensure it's hex encoded to catch errors before they hit the app code. This isn't a blocker, but it would be nice because it would put all the errors in the same place.

redshiftzero:
OK rebased on current `develop` and tests are passing now 😄  Needs restamp for merge

securedrop/journalist_app/forms.py
redshiftzero: Good idea - want to file a followup ticket for that?

PR #2506, opened 2017-11-04
CSS based deletion confirmation, #295

### LINKED ISSUE #295: If journalists enable NoScript, they don't get warning when deleting source codename
Journalists who have scripts turned off do not get an 'are you sure?' warning when they go to delete a source code name. 

The means that journalists engaged in best security practices are ironically more at risk at slipping up and deleting a communications line with a legitimate source.

daonb:
## Status

Work in progress. Just index.html was updated, I will change `col.html` if this way is accepted.
I'm also not sure about the colors, should `Delete` be blue and `Confirm` red or both of them red?

## Description of Changes

Fixes #295 

Changes proposed in this pull request: Delete confirmation is based on CSS and does not require JS

## Testing

Send a message as a whistleblower, log as a journalist mark the message and press `Delete`. A confirm button will appear next to the delete button.

### DIFF
--- a/securedrop/journalist_templates/index.html
+++ b/securedrop/journalist_templates/index.html
@@ -12,7 +12,11 @@
         <button type="submit" name="action" value="download-all" class="small"><i class="fa fa-download"></i> {{ gettext('Download') }}</button>
         <button type="submit" name="action" value="star" class="small"><i class="fa fa-star"></i> {{ gettext('Star') }}</button>
         <button type="submit" name="action" value="un-star" class="small"><i class="fa fa-star-half-full"></i> {{ gettext('Un-star') }}</button>
-        <button type="submit" id="delete-collections" name="action" value="delete" class="small-danger"><i class="fa fa-trash-o"></i> {{ gettext('Delete') }}</button>
+        <input type="checkbox" id="delete-checkbox" name="input" class="small-danger" />
+        <label for="delete-checkbox" class="btn" id="delete-label"><i class="fa fa-trash-o"></i>{{ gettext('Delete') }}</label>
+        <button type="submit" id="delete-collections" name="action"
+        value="delete" class="small-danger"> {{
+            gettext('Confirm') }}</button>
       </p>
 
       {% if starred %}
--- a/securedrop/journalist_templates/index.html
+++ b/securedrop/journalist_templates/index.html
@@ -12,11 +12,11 @@
         <button type="submit" name="action" value="download-all" class="small"><i class="fa fa-download"></i> {{ gettext('Download') }}</button>
         <button type="submit" name="action" value="star" class="small"><i class="fa fa-star"></i> {{ gettext('Star') }}</button>
         <button type="submit" name="action" value="un-star" class="small"><i class="fa fa-star-half-full"></i> {{ gettext('Un-star') }}</button>
-        <input type="checkbox" id="delete-checkbox" name="input" class="small-danger" />
+        <input type="checkbox" id="delete-checkbox" name="input" class="small-danger">
         <label for="delete-checkbox" class="btn" id="delete-label"><i class="fa fa-trash-o"></i>{{ gettext('Delete') }}</label>
-        <button type="submit" id="delete-collections" name="action"
-        value="delete" class="small-danger"> {{
-            gettext('Confirm') }}</button>
+        <button type="submit" id="delete-collections" name="action" value="delete" class="small-danger">
+            {{ gettext('Confirm') }}
+        </button>
       </p>
 
       {% if starred %}
--- a/securedrop/sass/journalist.sass
+++ b/securedrop/sass/journalist.sass
@@ -98,3 +98,16 @@ button.small-danger, a.btn.small-danger, .btn.small-danger
 
 .journalist-reply__input
   max-width: 700px
+
+#delete-checkbox
+  display: none
+
+#delete-label
+  font-size: 13px
+  padding: 3px 7px
+
+#delete-collections
+  visibility: hidden
+
+#delete-checkbox:checked ~ #delete-collections
+  visibility: visible
--- a/securedrop/static/js/journalist.js
+++ b/securedrop/static/js/journalist.js
@@ -61,21 +61,6 @@ $(function () {
       }
     });
 
-  $("#delete-collection").submit(function () {
-    return confirm(get_string("collection-delete-confirm-string"));
-  });
-
-  $("#delete-collections").click(function () {
-    var checked = $(".panel ul#cols li :checkbox").filter(":visible").filter(function(index) {
-        return $(this).prop('checked');
-    });
-    if (checked.length > 0) {
-      return confirm(get_string("collection-multi-delete-confirm-string").supplant({ size: checked.length }));
-    }
-    // Don't submit the form if no collections are selected
-    return false;
-  });
-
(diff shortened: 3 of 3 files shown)


dachary:
@daonb please ping me explicitly with **@dachary** once you have resolved the errors (GitHub won't let me know that you did ;-).

heartsucker:
Ooooh snap. This is clean.

screenshot_2017-11-04_17-14-07

screenshot_2017-11-04_17-14-20

I would change the button to say "confirm delete" so it's very obvious what's happening. This will also future proof it if we choose to add other confirmations to other actions in the future.

heartsucker:
You also have a linting error:

```pytb
html_lint.py --printfilename --disable=optional_tag,extra_whitespace,indentation \
  File "/usr/local/bin/html_lint.py", line 66, in <module>
    sys.exit(html_linter.main(options))
  File "/usr/local/lib/python2.7/dist-packages/html_linter.py", line 1094, in main
    clean_html = template_remover.clean(io.open(filename).read())
  File "/usr/local/lib/python2.7/dist-packages/template_remover.py", line 258, in clean
    return _TemplateRemover(html_content, pattern).get_clean_content()
  File "/usr/local/lib/python2.7/dist-packages/template_remover.py", line 199, in get_clean_content
    assert False, 'Echo tags should be in just one line.'
AssertionError: Echo tags should be in just one line.
```

You can run the linters in the dev machine from the SD root with `make lint` or just `make html-lint` since you only changed the HTML.

heartsucker:
I also think it's very likely this will break the functional tests with Selenium since it will expect the click on "Delete" to send the request.

daonb:
@heartsucker seems like making it a *Confirm Delete* breaks the line and we're not sure it's worth it.
confirm delete

heartsucker:
@daonb what you could do is wrap the pair in its own div so that the latter appears below it and apply some styling. I'm not great with HTML / CSS, but something like:

```html
<div class="confirm-container">
  <button>Foo Action</button>
  <button>Foo Confirm</button>
</div>
...
```

and then force it to wrap below.

huertanix:
A quick design note: Having a confirmation button alongside top-level options seems somewhat dissimilar to the sort of confirmation dialogs that many users may be used to. I would recommend having the confirm button live inside a second row that can appear under the top-level options row with a light-red background with some copy asking the user to confirm or cancel:

"Are you sure you want to delete the selected [singular or plural whatever word we use to describe what is being deleted]?" [Yes] [Cancel]

dachary:
@daonb I volunteer to address the comments in case you don't have time right now. Just let me know ;-)

daonb:
@dachary Thanks for the offer. We have another meeting the coming Monday and I hope to finish it than. If I don't, it'll be all yours :-)

daonb:
@dachary it didn't happen :-( . what's more, that crooked apple have talked me into high-sierra which means I now have to download the box which takes over 2 hours and I only got an hour in this cafe before I sign off.
If you got time to complete this, I'd be happy if you do.

dachary:
@daonb :-) I'll do it, sure.

dachary:
@toast this is a draft to implement deletion confirmation without javascript. As @huertanix mentions, the proposed UX is somewhat unusual. Do you have an idea on how to make it better ? I can work on the implemenation based on a mockup, following the best practices you described when working on the i18n menu.

dachary:
rebased (trivial conflict because some content was added to the sass files in the meantime)

dachary:
transient error, dealt with in another pull request, this is a known error https://travis-ci.org/freedomofpress/securedrop/builds/312837803?utm_source=github_status&utm_medium=notification

<pre>
2017-12-07 08:35:56,101 DEBUG DELETE http://127.0.0.1:55140/hub/session/b39ab43b-3f7a-42f0-ad25-9693f5d2df2a {"sessionId": "b39ab43b-3f7a-42f0-ad25-9693f5d2df2a"}
Exception in thread Thread-2950:
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/travis/build/freedomofpress/securedrop/securedrop/source_app/utils.py", line 81, in async_genkey
    current_app.logger.error(
  File "/usr/local/lib/python2.7/dist-packages/werkzeug/local.py", line 347, in __getattr__
    return getattr(self._get_current_object(), name)
  File "/usr/local/lib/python2.7/dist-packages/werkzeug/local.py", line 306, in _get_current_object
    return self.__local()
  File "/usr/local/lib/python2.7/dist-packages/flask/globals.py", line 51, in _find_app
    raise RuntimeError(_app_ctx_err_msg)
RuntimeError: Working outside of application context.
</pre>

toast:
@dachary Yes we could try to convert this into a no JS modal dialog.

**Step 1:**
delete dialog modal - step 1 2x

**Step 2:**
delete dialog modal - step 2 2x

**Step 3:**
delete dialog modal - step 3 2x

In theory it's possible to do this client side with no JS by following a similar pattern to the menu.sass module; using a checkbox to toggle a modal dialog that interrupts the journalist. 

**But it'll be a bit more complicated because:**
When the modal opens, the checkbox's label would need to become an invisible overlay similar to menu.sass that closes the modal when the user clicks outside of it. However, to allow for a CANCEL action and the 'X' with no JS in the window, the close overlay would actually need to cover those two areas of the modal as well (see the blue area in image below). But the DELETE action would need to be placed above the content area to still allow for interaction - this could be done with z-index. Ideally the title and content text would also be placed above the cancel overlay to allow interaction:

z-indices-explanation 2x

Note that the blue area (close overlay) needs to extend to the edges of the screen, I just left it smaller here to see the difference between this and the semi-transparent gray overlay that is behind the entire modal.

All this will also all need to be kept inside the `<form>` so the post still works when the DELETE inside the modal is clicked (the other DELETE becomes the checkbox label).

I'm not sure if all this will work together neatly in the HTML structure, I haven't made a modal without JS before. Once created though, it could be useful for other dangerous actions elsewhere.

dachary:
> I'm not sure if all this will work together neatly in the HTML structure, I haven't made a modal without JS before.

It would be really nice if it did though. Using the z-index so the action is on top is a clever trick. But, as you point out, I can't imaging how the HTML structure should look like for all this to work together...

dachary:
@toast I propose we move this discussing [in the forum](https://forum.securedrop.club/t/css-based-modals/342/1) because its scope is larger than this PR. What do you think ?

toast:
Yep, makes sense @dachary

dachary:
@toast please let me know if you would like me to implement your proposal. I'd be happy to draft something for you to review.

P.S. Could you please let me know what your account name is on the [forum](https://forum.securedrop.club/) ?

SaptakS:
@toast I agree with your approach. Another approach of creating a modal I may propose is using anchor tags instead of checkbox. We can use the pseudo attribute `:target` for a div. I have made a small draft example using only css in codepen so that you all can have a look. Maybe this will work.
https://codepen.io/SaptakS/pen/xpByWY
@dachary do have a look.

dachary:
For the record @SaptakS volunteered to create a new pull request with a draft implementation. Closing this one for now @toast but feel free to re-open when you have time to work on it.

PR #2511, opened 2017-11-04
Containerize app tests

tmc:
## Status

Ready for review

## Description of Changes

Containerize application tests.

This runs the app tests at with parallelism of 3 in circle.

## Testing

make images test

## Deployment

N/A

## Checklist

-


tmc:
This is held up by some needed ci improvements. Opened #2515

securedrop/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM ubuntu@sha256:30204139c6ab96ebd75d72f34db390f28c4decd5e563488b4e485bf979397b67
+
+RUN apt-get update # 2017-11-04
```
dachary: Why pinning and then updating ?

dachary:
See also https://github.com/freedomofpress/securedrop/pull/2518

securedrop/Dockerfile
tmc: The update is necessary as the published images don't have any package data by default.

tmc:
Updated the FROM sha to match the current published ubuntu:xenial image.

securedrop/Dockerfile
tmc: specifically, https://github.com/tianon/docker-brew-ubuntu-core/blob/5fce3945d95630c2fc03c21ef8665d92bd824642/trusty/Dockerfile#L35

dachary: oh, of course. Not **upgrade** but **update**, sorry for the noise.

.circleci/config.yml
```diff
@@ -112,14 +112,47 @@ jobs:
...
+      DOCKER_API_VERSION=1.23
+    parallelism: 3
+    steps:
+      - run: sudo apt-get install -y make
+      - checkout
+
+      - run:
+          name: Build Docker images
+          command: |
+            set +o pipefail
```
heartsucker: Why are you adding a `+o pipefail`? That's the default, and I'm pretty sure we *want* the error code to propagate here to abort if an earlier command fails.

.circleci/config.yml
```diff
@@ -112,14 +112,47 @@ jobs:
...
+            docker images
+            fromtag=$(docker images |grep securedrop-test |head -n1 |awk '{print $2}')
+            cd securedrop && EXTRA_BUILD_ARGS="--cache-from securedrop-test:${fromtag}" make images TAG=${CIRCLE_SHA1}
```
heartsucker: We might want  to do `${fromtag:?}` to have to puke if the above `grep|head|awk` returns nothing.

.circleci/config.yml
tmc: the default is '-eo pipefail', the intent here is to prevent cache population failing from interrupting the build.

.circleci/config.yml
tmc: I think it's better to build without cache than to get into a stuck state if circle clears caches or cache population/loading fails in some manner.

securedrop/manage.py
```diff
@@ -194,6 +195,7 @@ def _add_user(is_admin=False):
                                              issuer_name='SecureDrop')
             qr = qrcode.QRCode()
             qr.add_data(uri)
+            sys.stdout = codecs.getwriter("utf-8")(sys.stdout)
```
dachary: Could you please update the commit message to explain why this is needed ?

dachary:
I tried the recommended test **make images** and then **make test**. The **make images** works fine but **make test** fails. The full log is at https://paste2.org/H89ED2XY . Maybe it's something missing in my environment ?

dachary:
Hum, no, it's leftovers from vagrant it seems.

dachary:
After the failure
<pre>
$ sudo docker ps --all
[sudo] password for loic: 
CONTAINER ID        IMAGE                                                      COMMAND             CREATED             STATUS                     PORTS               NAMES
f596a797b57d        securedrop-test:81f75b28f56fab2ebf8cba85052ef4435935b7ce   "./test -v"         5 minutes ago       Exited (4) 5 minutes ago                       securedrop-test
</pre>
the container is still there. And for some reason the **haveged** of the docker host uses 100% CPU.
<pre>
  558 root      20   0   12020   4584   1460 R  99.7  0.0   5:39.74 /usr/sbin/haveged --Foreground --verbose=1 -w 1024
</pre>

dachary:
I was able to run **make test** but it fails on deletion tests
<pre>
tests/test_crypto_util.py::TestCryptoUtil::test_genrandomid_default_locale_is_en PASSED
tests/test_crypto_util.py::TestCryptoUtil::test_get_wordlist PASSED
tests/test_crypto_util.py::TestCryptoUtil::test_getkey PASSED
tests/test_crypto_util.py::TestCryptoUtil::test_hash_codename PASSED
tests/test_crypto_util.py::TestCryptoUtil::test_word_list_does_not_contain_empty_strings PASSED
tests/test_db.py::TestDatabase::test_get_one_or_else_multiple_results PASSED
tests/test_db.py::TestDatabase::test_get_one_or_else_no_result_found PASSED
tests/test_db.py::TestDatabase::test_get_one_or_else_returns_one PASSED
tests/test_db.py::TestDatabase::test_journalist_string_representation PASSED
tests/test_db.py::TestDatabase::test_reply_string_representation PASSED
tests/test_db.py::TestDatabase::test_source_string_representation PASSED
tests/test_db.py::TestDatabase::test_submission_string_representation PASSED
tests/test_db.py::TestDatabase::test_throttle_login PASSED
tests/test_i18n.py::TestI18N::test_get_supported_locales PASSED
tests/test_i18n.py::TestI18N::test_i18n PASSED
tests/test_i18n.py::TestI18N::test_verify_default_locale_en_us_if_not_defined_in_config PASSED
tests/test_i18n.py::TestI18N::test_locale_to_rfc_5646 PASSED
tests/test_i18n.py::TestI18N::test_html_en_lang_correct PASSED
tests/test_i18n.py::TestI18N::test_html_fr_lang_correct PASSED
tests/test_integration.py::TestIntegration::test_delete_collection FAILED
tests/test_integration.py::TestIntegration::test_delete_collections FAILED
tests/test_integration.py::TestIntegration::test_filenames PASSED
</pre>
This is most likely an environmental problem (just like the *pyc*). Although we could try to fix that before merging, this is not the primary goal of this PR which is to allow containerized tests in the CI. We can later fix this to work in a developer environment as well.

securedrop/Makefile
```diff
@@ -0,0 +1,37 @@
...
+
+.PHONY: images
+images:
+	docker build $(EXTRA_BUILD_ARGS) -t $(IMAGE):$(TAG) .
+	# combine dockerfiles but strip out first COPY for better cache utilization
+	cat Dockerfile* |sed '/#skip-if-test$$/N;/#skip-if-test/d' | docker build $(EXTRA_BUILD_ARGS) -f - -t $(IMAGE)-test:$(TAG) .
+
+.PHONY: test
+test:
+	docker run --privileged \
```
dachary: why not **--rm** here ? Are there situations where we want to keep the container after a run ?

tmc: it should use --rm.

securedrop/manage.py
tmc: sure thing, https://circleci.com/gh/freedomofpress/securedrop/3618 is an example failure otherwise.

securedrop/Makefile
tmc: actually the caching of layers in circle relies on this.

msheiny: I major concerns with the `privileged` container use here. I dont think it should be a blocker for CI, but I think it should be a blocker from local development documentation and recommendation for users. I won't hold up the PR for it now

securedrop/Makefile
```diff
@@ -0,0 +1,37 @@
...
+
+.PHONY: assets
+assets:
+	sass
+
+.PHONY: images
+images:
+	docker build $(EXTRA_BUILD_ARGS) -t $(IMAGE):$(TAG) .
+	# combine dockerfiles but strip out first COPY for better cache utilization
+	cat Dockerfile* |sed '/#skip-if-test$$/N;/#skip-if-test/d' | docker build $(EXTRA_BUILD_ARGS) -f - -t $(IMAGE)-test:$(TAG) .
```
msheiny: Is this really necessary? I'd prefer breaking these dockerfiles over two separate files without this concatenation action. In other words if it takes a little longer for builds i'd rather have that versus this logic.

Makefile
```diff
@@ -2,6 +2,18 @@ DEFAULT_GOAL: help
 SHELL := /bin/bash
 PWD := $(shell pwd)
 
+.PHONY: images
+images:
+	$(MAKE) -C securedrop images
+
+.PHONY: test
```
msheiny: lets prepend this with `app-` . we have a number of different `test` suites and we need to make it more obvious which one we are referring to. also needs documentation for commands.

msheiny: or you can cherry pick `8e43edb1` in :)

securedrop/Makefile
conorsch: Agreed with @msheiny—although there's sophisticated cache optimization with the concatenation approach, it may pose problems for maintainability, particularly from new contributors. I'm fine trying out this approach and revising later, though, since the changes in this PR get us pretty far ahead of where we are now.

securedrop/Makefile
conorsch: Also agreed: `--privileged` is _only_ OK in CI, we should _not_ recommend developers run privileged containers on their workstations.

securedrop/Makefile
tmc: I concur that this is probably a little too fancy. I can simplify this and settle on one. I was thinking ahead to what publishing official images might look like and how cache sharing could assist in that situation. I'll back this out and it can be revisited in the future.

securedrop/Makefile
tmc: I agree, as discussed we need a way to increase entropy in CI and will have to instruct folks in other environment (such as Qubes) to increase the rate of entropy collection.

Makefile
tmc: done. Thanks!

securedrop/Makefile
tmc: pushed this change.

heartsucker:
Consider my review dismissed. I can't totally comment on this at the moment, and the previous rejection was based on minor things that were addressed.

conorsch:
Sweet, with the PaX flags added, I'll pull down the latest for local review!

tmc:
@conorsch get a chance to evaluate?

Issue #2522, opened 2017-11-06
refactor securedrop-admin sdconfig in python


dachary:
# Feature request

## Description

securedrop-admin sdconfig is implemented in python via install_files/ansible-base/securedrop-configure.yml

ansible is contorted and not user friendly in this context. What we want is to display a prompt to the user with a default or the previous value read from site_specific. And have immediate validation instead of post-submission validation.

The [click](http://click.pocoo.org/5/) library is already a SecureDrop dependency and could be used for that purpose.

See also https://github.com/freedomofpress/securedrop/issues/2266 and https://github.com/freedomofpress/securedrop/issues/2095

## User Stories

As an admin I want to get immediate validation when `securedrop-admin sdconfig` prompts me for a valid when configuring securedrop for the first time.

As an admin I want `securedrop-admin sdconfig` to display the values from `install_files/ansible-base/group_vars/all/site-specific` when running it the second time so that I can modify them instead of editing the file manually.

dachary:
Quoting @redshiftzero 
> i think replacing everything before "Validate site-specific information." in python makes sense and would lead to a much nicer user experience (one could still do some additional validation in python for some immediate user feedback but replacing the entire validate role seems like a big task)

dachary:
Quoting @msheiny 
> i'm totally on-board with you swapping that out -- i also had reservations about doing that code logic in ansible. ansible is a terrible replacement for doing that in python with proper argument parsing validation

dachary:
I will not use click because it is not actively maintained. Replacing argparse with click would not be a good idea because of that. There are [>300 issues opened](https://github.com/pallets/click/issues) with [very few commits](https://github.com/pallets/click/graphs/contributors) over the past few months and [62 pull requests opened](https://github.com/pallets/click/pulls) which is a lot considering the project has less than 1000 commits.

dachary:
* http://cooked-input.readthedocs.io/en/latest/ good match but very recent (will it be maintained ?)

kushaldas:
if not Click then we should stay with argparse. Even though most of the projects I came across use Click now :)

dachary:
> if not Click then we should stay with argparse.

💯

dachary:
I ended up using http://python-prompt-toolkit.readthedocs.io/ which does a great job

PR #2524, opened 2017-11-06
l10n: add SUPPORTED_LOCALES to sdconfig

### LINKED ISSUE #2516: add SUPPORTED_LOCALES to sdconfig
# Feature request

## Description

The value of **SUPPORTED_LOCALES** in config.py must be set via **./securedrop-admin sdconfig**

## User Stories

As a SecureDrop admin configuring the application server, I want to chose the languages supported by the instance among all supported languages.

dachary:
## Status

Ready for review

## Description of Changes

Fixes #2516

Allow the sysadmin installing (or updating) SecureDrop to specify the
content of the SUPPORTED_LOCALES variable when running

   ./securedrop-admin sdconfig

The updated value from the site-specific file will override the value
in config.py in the same way DEFAULT_LOCALE is overriden.

The list of locales entered via sdconfig is validated to match
existing directories from the securedrop/translations directory.

### Development VM

For each test case below repeat the following:

* ./securedrop-admin -d sdconfig
* enter the value when prompted for **Space separated list of supported locales**
* the validation will fails because it is not Tails but validation of **securedrop_supported_locales** is done
* ANSIBLE_ARGS="--tags securedrop_config" vagrant provision development
* verify that the task **Update SUPPORTED_LOCALES in config.py** runs successfully
* verify that securedrop/config.py has a SUPPORTED_LOCALES setting matching install_files/ansible-base/group_vars/all/site-specific
* remove the **securedrop_supported_locales** line in install_files/ansible-base/group_vars/all/site-specific

Test cases:

* empty string => securedrop_supported_locales: [] => passes validation
* ar     fr_FR (separated by multiple white space) => [ar, fr_FR] => passes validation
* foobar ar => [foobar, ar] => fails validation on foobar

### Verifying it applies to config.py post installation from packages

* make build-debs
* vagrant up /staging/
* vagrant ssh app-staging
* get the config.py file
* verify the value of SUPPORTED_LOCALES in config.py is consistent with the tests done in **Development VM** above

## Deployment

The variable **SUPPORTED_LOCALES** is added to config.py (when it exists or when it is new). 

### DIFF
--- a/install_files/ansible-base/roles/validate/files/validate-supported-locales.sh
+++ b/install_files/ansible-base/roles/validate/files/validate-supported-locales.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+# Verifies a all arguments match a translation directory
+#
+# Exit codes
+# 0 - all arguments match a translation directory
+# 1 - otherwise
+
+for translation in $1 ; do
+    if test "$translation" = en || test "$translation" = en_US ; then
+        continue
+    fi
+    directory="securedrop/translations/$translation"
+    if ! test -d "../../$directory" ; then
+        echo "$directory does not exist"
+        exit 1
+    fi
+done
--- a/install_files/ansible-base/securedrop-configure.yml
+++ b/install_files/ansible-base/securedrop-configure.yml
@@ -110,6 +110,12 @@
       default: ""
       private: yes
 
+    - name: securedrop_supported_locales
+      # the list is from the securedrop/translations repository
+      prompt: Space separated list of additional locales to support (ar de_DE en_US es_ES fr_FR nb_NO nl)
+      default: ""
+      private: no
+
   pre_tasks:
     - debug:
         msg: >-
@@ -175,6 +181,8 @@
           var_value: "{{ sasl_password }}"
         - var_name: securedrop_app_https_on_source_interface
           var_value: "{{ securedrop_app_https_on_source_interface | bool}}"
+        - var_name: securedrop_supported_locales
+          var_value: "{% if securedrop_supported_locales is string %}{{ securedrop_supported_locales.split() | to_yaml }}{% else %}{{ securedrop_supported_locales | to_yaml }}{% endif %}"
 
 - name: Validate site-specific information.
   hosts: localhost
--- a/install_files/ansible-base/roles/app/defaults/main.yml
+++ b/install_files/ansible-base/roles/app/defaults/main.yml
@@ -99,3 +99,5 @@ worker_logs_dir: /var/log/securedrop_worker
 
 # The locale used when the browser or the user do not have a preference
 securedrop_default_locale: en_US
+# The subset of the available locales that will be proposed to the user
+securedrop_supported_locales: []
--- a/install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml
+++ b/install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml
@@ -189,3 +189,12 @@
   when: db.stat.exists
   tags:
     - securedrop_config
+
+- name: Update SUPPORTED_LOCALES in config.py
+  blockinfile:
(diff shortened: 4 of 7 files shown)


dachary:
> Also the test should include building the deb packages if there is any previous commit related to this change.

The tests instructions in the PR message were updated with instructions to verify the config.py they create.

install_files/ansible-base/securedrop-configure.yml
```diff
@@ -110,6 +110,12 @@
       default: ""
       private: yes
 
+    - name: securedrop_supported_locales
+      # the list is from the securedrop/translations repository
+      prompt: Space separated list of supported locales (ar, de_DE, en_US, fr_FR, nb_NO, nl)
```
redshiftzero: Hmm, this is slightly confusing since this says space-separated list but then there's a comma-separated list. Maybe this could be `(ar de_DE en_US fr_FR nb_NO nl)` for clarity (or we can use `split(',')` below on line 185). 

Also, what this does currently is specify additional locales, so we could say this explicitly for clarity, e.g "Space separated list of additional locales to support (`ar de_DE fr_FR nb_NO nl`)"

install_files/ansible-base/securedrop-configure.yml
```diff
@@ -175,6 +181,8 @@
           var_value: "{{ sasl_password }}"
         - var_name: securedrop_app_https_on_source_interface
           var_value: "{{ securedrop_app_https_on_source_interface | bool}}"
+        - var_name: securedrop_supported_locales
+          var_value: "{% if securedrop_supported_locales is string %}{{ securedrop_supported_locales.split() | to_yaml }}{% else %}{{ securedrop_supported_locales | to_yaml }}{% endif %}"
```
redshiftzero: Oof - I see why you filed #2522 :+1: :smile:

install_files/ansible-base/roles/validate/files/validate-supported-locales.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+# Verifies a all arguments match a translation directory
+#
+# Exit codes
+# 0 - all arguments match a translation directory
+# 1 - otherwise
+
+for translation in "$@" ; do
```
redshiftzero: I think we want this to be `for translation in $@ ; do` - as it is right now the additional quotes mean that bash doesn't split on the whitespace in the string, e.g.: 

```
amnesia@amnesia:~/Persistent/securedrop$ ./install_files/ansible-base/roles/validate/files/validate-supported-locales.sh "foobar ar"
securedrop/translations/foobar ar does not exist
amnesia@amnesia:~/Persistent/securedrop$ ./install_files/ansible-base/roles/validate/files/validate-supported-locales.sh foobar ar
securedrop/translations/foobar does not exist
```

(this is the cause of the failure I saw for the test case `ar fr_FR`)

dachary:
thanks for the detailed review ! I'll work on that first thing tomorrow :-)

install_files/ansible-base/securedrop-configure.yml
dachary: Removed the **,** good idea ! Also changed the wording but kept **en_US** in the list. I think there is value in listing all available locales even though **en_US** will always be there.

install_files/ansible-base/roles/validate/files/validate-supported-locales.sh
dachary: Fixed. I wonder how it worked in the past. I'm tempted to blame ansible changing quoting rules. But not motivated to actually look for a proof because it's such a simple fix ;-)

dachary:
@redshiftzero fixed and repushed. Also updated the test instructions: they were so bad I could not repeat them ... sorry about that :-(

dachary:
Also added Spanish because it is now 100% \o/

redshiftzero:
Confirmed via manual testing in Tails that the previously failing test case is now working. Merging!

PR #2525, opened 2017-11-06
Improved page responsiveness and buttons clarity for source interface

nsalontay:
## Status

Ready for review

## Description of Changes

- Percent widths added to non-responsive elements in the source interface.
securedrop___protecting_journalists_and_sources

- In codename generation page, buttons changed from “Continue” and "Already have a codename?" to "Use new codename" and "Use existing codename" for clarity.
securedrop___protecting_journalists_and_sources

- Primary buttons (“Submit”, “Continue”, “Use New Codename”) in the source interface are now before the secondary buttons. This lends proximity between the content of pages and the primary action.
image
_Above: Text field is now responsive and buttons have been reversed so "Continue" is next to field rather than "Cancel"_

## Testing

I'm mainly concerned about the changes to classes ".message" and ".attachment" in _base.css. I'm not sure if they are used anywhere else in the application other than for the document submission form.

## Deployment
None

### DIFF
--- a/securedrop/source_templates/generate.html
+++ b/securedrop/source_templates/generate.html
@@ -39,12 +39,12 @@
 <div class="pull-right">
 <form id="create-form" method="post" action="/create" autocomplete="off">
   <input name="csrf_token" type="hidden" value="{{ csrf_token() }}">
-  <a id="already-have-codename" href="{{ url_for('main.login') }}" class="sd-button btn secondary">{{ gettext('ALREADY HAVE A CODENAME?') }}</a>
   <button type="submit" class="sd-button btn primary" id="continue-button">
     <img class="icon off-hover" src="{{ url_for('static', filename='i/font-awesome/fa-arrow-circle-o-right-white.png') }}" width="17px" height="17px">
     <img class="icon on-hover" src="{{ url_for('static', filename='i/font-awesome/fa-arrow-circle-o-right-blue.png') }}" width="17px" height="17px">
-     {{ gettext('CONTINUE') }}
+     {{ gettext('USE NEW CODENAME') }}
   </button>
+  <a id="already-have-codename" href="{{ url_for('main.login') }}" class="sd-button btn secondary">{{ gettext('USE EXISTING CODENAME') }}</a>
 </form>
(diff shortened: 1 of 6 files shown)


dachary:
> I'm mainly concerned about the changes to classes ".message" and ".attachment" in _base.css. I

These must be classes of found in an element which is a child of an element with the **snippet** class. There is only one element with such a class, in `securedrop/source_templates/lookup.html`.

securedrop/source_templates/generate.html
```diff
@@ -39,12 +39,12 @@
 <div class="pull-right">
 <form id="create-form" method="post" action="/create" autocomplete="off">
   <input name="csrf_token" type="hidden" value="{{ csrf_token() }}">
-  <a id="already-have-codename" href="{{ url_for('main.login') }}" class="sd-button btn secondary">{{ gettext('ALREADY HAVE A CODENAME?') }}</a>
   <button type="submit" class="sd-button btn primary" id="continue-button">
     <img class="icon off-hover" src="{{ url_for('static', filename='i/font-awesome/fa-arrow-circle-o-right-white.png') }}" width="17px" height="17px">
     <img class="icon on-hover" src="{{ url_for('static', filename='i/font-awesome/fa-arrow-circle-o-right-blue.png') }}" width="17px" height="17px">
-     {{ gettext('CONTINUE') }}
+     {{ gettext('USE NEW CODENAME') }}
```
dachary: What about **GET NEW CODENAME** ? I don't feel strongly about that but **USE** suggest you have a codename already. What really happens is that you'll be given a codename. What do you think ?

nsalontay: I was operating under the impression the codename earlier in the page becomes their assigned code name if they click this button.

Perhaps "Use this codename " would be even better, but "get" makes me think it'll be in the next page. Want to make it clear to the user that they should make note of the codename before proceeding.

dachary: > I was operating under the impression the codename earlier in the page becomes ...

You are correct and I was confused. Sorry for the noise.

PR #2549, opened 2017-11-08
Convert tests in SubmissionNotInMemoryTest to pytest and xfail them

### LINKED ISSUE #2548: Convert tests in SubmissionNotInMemoryTest to pytest and xfail them
## Description

In #476, some tests that dump memory and determine if memory contains secrets were merged in. These tests cannot pass. We do have secrets in memory (though not on disk, see #862). 

Since that time, we have made other changes to the test suite, including using `pytest` to run the tests, but we left this file behind (oops). Recently a contributor was containerizing our tests (#2511) and [ran into trouble](https://circleci.com/gh/freedomofpress/securedrop/3466) due to executing everything in `tests/functional`, including this file, and seeing them fail. Let's be clear about these tests by updating them so that they do run properly, and then mark them `xfail` so it is clear that they are not expected to pass.  

## User Stories

As a SecureDrop developer, I don't want to have files containing tests that don't work lying around not clearly marked `xfail` because it is extremely confusing :)

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #2548. 

* Update `SubmissionNotInMemoryTest` and mark them `xfail`

## Testing

Do you see:

```
tests/functional/test_submission_not_in_memory.py::TestSubmissionNotInMemory::test_message_is_not_retained_in_memory xfail
tests/functional/test_submission_not_in_memory.py::TestSubmissionNotInMemory::test_file_upload_is_not_retained_in_memory xfail
```

in passing CI? You can also remove `xfail` from these two tests and see them fail due to the presence of secrets in memory. 

## Deployment

None, tests only


install_files/ansible-base/group_vars/all/securedrop
```diff
@@ -50,4 +51,3 @@ appserver_dependencies:
   - apparmor-utils
   - redis-server
   - supervisor
-
```
kushaldas: Aren't we supposed to keep the newlines at the end of the file for POSIX standard?

heartsucker:
> I just wish to add a docs page, where we can mark which change was introduced in which release and point to the right issue/PR.

I think this is a more general problem and could be addressed in another issue / PR.

redshiftzero:
> I just wish to add a docs page, where we can mark which change was introduced in which release and point to the right issue/PR.

I think the [changelog](https://github.com/freedomofpress/securedrop/blob/develop/changelog.md) should satisfy this, which we update prior to each release. I might try [this changelog generator](https://github.com/skywinder/github-changelog-generator) next time as preparing the changelog manually is a bit of a bore :)

PR #2567, opened 2017-11-12
source: add random payload to each page

### LINKED ISSUE #2566: add random payload to
# Bug

## Description

HTML pages that have a fixed size makes it easier for a third party to [fingerprint SecureDrop](https://forum.securedrop.club/t/fingerprinting-securedrop-onion-services). 

## Steps to Reproduce

Do the following twice:

* Capture tor packets after the circuits are established
* Visit the  index page of the source interface

## Expected Behavior

The total size of the captured packets varies randomly.

## Actual Behavior

The total size of the captured packets does not change.

## Comments

Add a random payload at the end of each page.

dachary:
## Status

Ready for review

## Description of Changes

Contributes to #2566

add random payload to each page

## Testing

The CI is the only test. There is no fingerprinting  software that is widely accepted and that we could use to demonstrate this helps.

## Deployment

No impact, just the page size are of variable size.

### DIFF
--- a/securedrop/source_app/utils.py
+++ b/securedrop/source_app/utils.py
@@ -2,6 +2,7 @@ import subprocess
 
 from datetime import datetime
 from flask import session, current_app, abort, g
+import random
 from threading import Thread
 
 import crypto_util
@@ -96,3 +97,7 @@ def normalize_timestamps(filesystem_id):
                 "Couldn't normalize submission "
                 "timestamps (touch exited with %d)" %
                 rc)
+
+
+def randompayload(min=1, max=2048):
+    return "." * random.randint(min, max)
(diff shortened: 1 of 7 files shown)


securedrop/source_app/utils.py
```diff
@@ -96,3 +99,7 @@ def normalize_timestamps(filesystem_id):
                 "Couldn't normalize submission "
                 "timestamps (touch exited with %d)" %
                 rc)
+
+
```
heartsucker: We don't need to add random bytes. We only need to add size since after THS end-to-end encryption the bytes will look random anyway. Using `os.urandom` may exhaust our entropy, and we actually have checks that prevent key generation when there's insufficient entropy.

dachary: Ok, I changed that to just using dots.

dachary:
> I wouldn't want to implement a countermeasure of any sort without verify that it has the intended outcome (decreasing fingerprintability). 

It looks like the only way to find out is to implement this countermeasure and wait long enough for the researchers to either confirm or invalidate that it helps. We do not have access to the software or methodology they use to evaluate fingerprinting. Ideally they would share it. But it looks like this is not going to happen any time soon.

This countermeasure is the simplest possible and addresses something that is quite intuitive: fixed payloads are good for fingerprinting.

That being said I don't feel strongly about this.

redshiftzero:
This is a really clean implementation :1st_place_medal: and we should save these commits for reference. I do agree with @heartsucker's feedback and think we should close for now based on the reasoning I described here: https://forum.securedrop.club/t/fingerprinting-securedrop-onion-services/199/6.

PR #2569, opened 2017-11-12
Add i18n menu with flags

### LINKED ISSUE #2557: i8n language menu
# i8n language menu

## Description

A dropdown menu defaulted to the language set by the journalist org on installation. This keeps the main content area free from the switching language info; it remains focused on a single task - submitting documents.

This should simplify the interface for the majority of sources; assuming that most sources use a journalist org in their own language. For the situation where I'm an English speaking source submitting documents to Süddeutsche Zeitung then it will be a little harder to find the language switcher since it will say 'Deutsch'. The flag helps with this - makes it more findable.

**Mobile mockup EN**
i8n example navigation - mobile

**Mobile mockup DE**
i8n example navigation - mobile - de

**Desktop mockup**
i8n example navigation - desktop

toast:
## Status

The i8n application test is failing but I'm unsure how to fix it. Otherwise, ready for review.

## Description of Changes

Fixes #2557 - i8n language menu

**Note:** No Javascript is used. Instead the menu opens by using checkbox:checked and CSS sibling selector '~'

- Modules folder introduced to improve CSS encapsulation
- Using BEM syntax for better encapsulation of modules
- Visually-hidden module keeps a hidden element accessible to screenreaders - hides the checkbox.
- Menu module is generic allowing for reuse for other menus
- Flag module is generic allowing flag icons to be used in other contexts if needed

## Testing

- Check the menu opens. Check it closes when you click outside of it (on the rest of the page)
- It should work with Javascript turned off
- Layout should look correct on the home page and in the sidebar on signed in pages
- Several flag images added but more may be needed. I'm not sure if I had all the supported languages turned on.
- _Known issue:_ when the menu is open; horizontal and vertical browser scrollbars appear


dachary:
> Known issue: when the menu is open; horizontal and vertical browser scrollbars appear

The following patch fixes this, I think. But I'm not sure why it was given such a large width/height in the first place so.... maybe I'm missing something ?

<pre>
modified   securedrop/sass/modules/_menu.sass
@@ -51,8 +51,6 @@
   position: absolute
   top: -100vh
   left: -100vw
-  width: 200vw
-  height: 200vh
   background-color: transparent
   & .menu__trigger-text
     display: none
</pre>

dachary:
@toast I like those flags :-) Where did you find them ?

dachary:
I noticed that if you select arabic, it shows the page in arabic, this is good. But when you click the arabic flag again to chose another language, the page becomes all white. I suppose it has to do with top/left being inverted when left-to-right is active or something like that ? I can try to investigate that more if you'd like.

dachary:
@toast I added a commit to your pull request to fix the tests and remove a loop that was no longer useful in locales.html.

Very excited to see that making progress !

toast:
@dachary *on the flags*: I used and slightly adapted https://www.sketchappsources.com/free-source/823-flag-set-sketch-freebie-resource.html which is in turn inspired by https://dribbble.com/shots/1089488-Stripe-Flag-Set (but it's a more 'flat design' version). So credit to https://twitter.com/bdc for the original set and to https://twitter.com/lucchaissac for adapting them. They both publicly advertised them on the web as 'free'.

dachary:
@toast three commits were pushed to describe the problems I found and propose a tentative fix:

* we rely on https://pypi.python.org/pypi/html-linter/ to verify html and it is picky about using simple quotes or _ in CSS class names
* the **absolute** rule made the menu-trigger element jump to the left so I removed it but did not try to understand why it was necessary in the first place
* a rtl/ltr conditional on **visually-hidden** was added so ?l=ar displays instead of being hidden

These are all suggestions that are primarily meant to describe the problems I found. Feel free to discard them for something you think more appropriate.

I feel we're getting very close to merging this change :-)

dachary:
Also added the new module directory to the rsync file so it can be packaged properly.

toast:
@dachary thanks for these! Sorry I missed the linter; is there a way to run that separate to the application tests?

Unfortunately the linter disallowing underscore means that the form of [BEM syntax I was suggesting](https://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/) doesn't work. However, I'm very keen on this syntax or a similar one and would like to try to explain why:

BEM is a great way to encapsulate the CSS into smaller modules. The project currently has very large CSS files and I'm keen to have a way to split these (similar to 'extract method' refactoring). When you're editing a BEM module, you know that your changes won't affect code outside of that module, which makes it easier for developers with less CSS knowledge to still be able to make changes to it.

This is the meaning of the syntax:
* `.menu` - This is the block name and corresponds to the filename menu.sass.
* `.menu__items` - The double underscore denotes that 'items' is an _element_ that belongs to the block 'menu', so you can also expect to find it in the file named menu.sass.
* `.menu--open` - The double hyphen indicates a state or modifier that is applied to the original default state of the menu. Since no javascript was used for this menu there aren't many modifiers but there's an example for `.flag` which uses `.flag--ar` to modify the flag shown based on the locale.
* Single hyphens denote spaces in words. E.g. for `menu__trigger-text` 'trigger text' is a single element with spaces since it is two words.

**I'm wondering if there is a way to customize the html-linter to allow this syntax? Many linters allow you to switch off a particular rule, I couldn't see that in this case but perhaps I'm missing something?** An alternative could be to use a letter to denote the element with `e-` so that `.menu__items` would become `.menu-e-items` but this is a non-standard syntax and is less scannable for developers.

I realise this is a bit of a change from the current project conventions but I've seen it work very well for organising CSS on a lot of projects (both large and small). And many developers I've worked with, who found it very verbose to begin with, became keen on it as they realised that it makes it easier to reason about and edit the CSS.

I of course understand if it doesn't work for this team, it's just my suggestion and I'll be happy to adapt to work with the convention that you all prefer.

dachary:
@toast thanks for explaining this in detail. To be honest I'm not a big fan of the linter. It was introduced to detect wrongly formatted HTML. I'll propose to replace that with something else tomorrow.

redshiftzero:
Thanks for this explanation @toast! I agree that this convention produces cleaner and more maintainable CSS. 

> I'm wondering if there is a way to customize the html-linter to allow this syntax? Many linters allow you to switch off a particular rule, I couldn't see that in this case but perhaps I'm missing something?

A minimal change that would allow you to use this syntax would be to add `names` as one of the exceptions in the linter:

```
diff --git a/Makefile b/Makefile
index 050abf5..fc10c63 100644
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ flake8: ## Validates PEP8 compliance for Python source files.
 
 .PHONY: html-lint
 html-lint: ## Validates HTML in web application template files.
-       html_lint.py --printfilename --disable=optional_tag,extra_whitespace,indentation \
+       html_lint.py --printfilename --disable=optional_tag,extra_whitespace,indentation,names \
 
 .PHONY: yamllint
```

> is there a way to run that separate to the application tests?

Yep! A shortcut to running the linter is provided in the Makefile, so you can run locally:

```
make html-lint
```

dachary:
@toast I discarded the commit added to your pull request and changing the CSS names. And added an exception as suggested by @redshiftzero (thanks for that, much easier than switching to another linter :-).

dachary:
CircleCI fails https://circleci.com/gh/freedomofpress/securedrop/3898?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
<pre>
"msg": "Instance creation failed => InstanceLimitExceeded: Your quota allows for 0 more running instance(s). You requested at least 1"}
</pre>

dachary:
@redshiftzero thanks for the review, the 9 commits were squashed them into three so they are more readable.

PR #2523, opened 2017-11-06
Containerize development environment

tmc:
## Status

Work in progress

## Description of Changes
This supplies the ability to build development environments via docker. This simplifies getting started for new developers as well as allows development on Qubes. Based on #2511 

## Testing

make app-images app-dev

This brings up the source and journalist interfaces on :8080, :8081

## Deployment
N/A

## Checklist
N/A

PR #2572, opened 2017-11-13
Enable vnc in containerized test and dev

tmc:
## Status

Ready for review

## Description of Changes

Depends on #2523 (should be rebased once that lands).

Adds x11vnc on port 5901 for test and dev environments to aid in debugging selenium tests.

## Testing

Connect a vnc client to :5901 when running dev container via 'make images dev'

## Deployment
N/A

## Checklist
N/A

### DIFF
--- a/securedrop/Dockerfile
+++ b/securedrop/Dockerfile
@@ -10,6 +10,7 @@ RUN gem install sass -v 3.4.23
 # test requirements:
 RUN apt-get install -y firefox
 RUN apt-get install -y git xvfb haveged curl gettext paxctl
+RUN apt-get install -y x11vnc
 
 # pinned firefox version for selenium compatibility. See bd795e2f5865b4f6e6e1b88bcbbacef704675c74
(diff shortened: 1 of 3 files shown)


dachary:
@redshiftzero this was rebased. @tmc a few lines in the [developer guide](https://docs.securedrop.org/en/latest/development/setup_development.html) to explain how it can be used would be useful. Otherwise people won't know it exists. Not a blocker for the merge, just a thought.

msheiny:
> @redshiftzero this was rebased. @tmc a few lines in the developer guide to explain how it can be used would be useful. Otherwise people won't know it exists. Not a blocker for the merge, just a thought.

Hey @dachary -- we had a discussion at the last hack-a-thon with @tmc that we would agree to merge #2523 now even though it was failing app tests to keep things moving forward. The caveat was though its not considered "production" ready for us to recommend so we'll keep grinding on it in the background and add documentation when its ready.

PR #2573, opened 2017-11-13
Fixes #2161 allows multiple file upload

### LINKED ISSUE #2161: Multiple file upload
# Feature request

Multi-file upload on the Submit Materials (`/lookup`) page

Couldn't find any GitHub issue about this. Does one exist? 

It would be nice if sources had the ability upload more than one file at a time. Say, a max of 5.

kushaldas:
## Status

Ready for review

## Description of Changes

Fixes #2161 

Now a source can upload more than one file using the file selector
in the source view. This also includes the updates to the tests as
required.

## Testing

In the source view, try to select more than one document and submit them.

## Deployment

Nothing changed in the database. Nothing extra required for deployment.


kushaldas:
After this is merged I will do another tests update to have the same style in submitting in the source view (in every test where we submit something).

kushaldas:
As I understood multiple file upload is [not yet supported in Selenium](https://github.com/seleniumhq/selenium-google-code-issue-archive/issues/2239).

securedrop/source_app/main.py
```diff
@@ -126,14 +133,15 @@ def submit():
...
-                    fh.stream))
+            for file_submitted in fh:
+                g.source.interaction_count += 1
+                fnames.append(
+                    store.save_file_submission(
+                        g.filesystem_id,
+                        g.source.interaction_count,
+                        journalist_filename,
+                        file_submitted.filename,
+                        file_submitted.stream))
```
heartsucker: We should commit after each add so that if the connection breaks, we will have saved some of the files. We should also flash a message "file `foo.txt`" saved per DB commit so the user only has to reupload a portion of the files after a connection breaks.

dachary:
Well, it's apparently impossible to revert to not having an opinion once you either approved or request change... so: I'm approving but what I'm really saying is that I'm leaving it to someone else to review this pull request.

securedrop/source_app/main.py
heartsucker: Normally I wouldn't recommend this but since this is over a THS, I think this is advisable.

securedrop/source_app/main.py
```diff
@@ -113,27 +120,39 @@ def submit():
...
+                submission = Submission(g.source, fname)
+                db_session.add(submission)
+                db_session.commit()
+
+                html_msg = render_template(
+                    'next_submission_flashed_message.html',
+                    html_contents="{0}".format(file_submitted.filename))
+                flash(Markup(html_msg), "info")
```
redshiftzero: Hmm, this merits further inspection - there is untrusted user-supplied input (the filename of the submission) passed to `Markup` - which is equivalent to Jinja's `|safe`). From the Flask documentation:

> If you can trust a variable and you know that it will be safe HTML (for example because it came from a module that converts wiki markup to HTML) you can mark it as safe by using the Markup class or by using the |safe filter in the template.

kushaldas: I was thinking about the same, also to use `secure_filename` and escape it before putting it into a Markup string.

heartsucker: `secure_filename` does not strip out markup, only things like `../`.

```python
    if isinstance(filename, text_type):
        from unicodedata import normalize
        filename = normalize('NFKD', filename).encode('ascii', 'ignore')
        if not PY2:
            filename = filename.decode('ascii')
    for sep in os.path.sep, os.path.altsep:
        if sep:
            filename = filename.replace(sep, ' ')
    filename = str(_filename_ascii_strip_re.sub('', '_'.join(
                   filename.split()))).strip('._')

    # on nt a couple of special files are present in each folder.  We
    # have to ensure that the target file is not such a filename.  In
    # this case we prepend an underline
    if os.name == 'nt' and filename and \
       filename.split('.')[0].upper() in _windows_device_files:
        filename = '_' + filename

    return filename
```

Also, that flashed message is pretty giant, so if someone submits 10 files, then it will fill up the page. Maybe we could flash tiny messages for each file and that big one at the end like "All your files were uploaded." We could also clear the flashed messages before flashing the big one so they have one that says "it's all here" and not all the incremental ones too.

kushaldas: That is why I said both secure_filename and escape.

`Markup.escape(secure_filename(filename))`

emkll: It seems like the `flash` function escapes the problematic characters: 
```
<div class="message">
  <p>&lt;script&gt;alert();&lt;:script&gt; 2.txt</p>
</div>
```
 While potentially redundant, I agree that using `Markup.escape()` will ensure the snippet won't be copy-pasted out of context.

kushaldas: Thanks, I will update the PR accordingly.

securedrop/tests/test_integration.py
```diff
@@ -584,18 +584,18 @@ def test_login_after_regenerate_hotp(self):
...
+            'fh[]': [],
+        }, follow_redirects=True)
+        self.source_app.post('/submit', data={
+            'msg': "This is a test.",
+            'fh[]': [(StringIO('This is a test'), 'test.txt'), ],
+        }, follow_redirects=True)
+        self.source_app.post('/submit', data={
+            'msg': "",
+            'fh[]': [(StringIO('This is a test'), 'test.txt'), ],
+        }, follow_redirects=True)
```
redshiftzero: Woohoo thanks for these tests :tada:

securedrop/source_app/main.py
```diff
@@ -113,27 +120,40 @@ def submit():
...
+                        file_submitted.stream))
+                submission = Submission(g.source, fname)
+                db_session.add(submission)
+                db_session.commit()
+
+                filename = Markup.escape(file_submitted.filename)
+                html_msg = render_template(
+                    'next_submission_flashed_message.html',
+                    html_contents="{0}".format(filename))
+                flash(Markup(html_msg), "info")
```
redshiftzero: So the flashed messages look a bit wonky:

screen shot 2017-11-14 at 4 14 42 pm

Primarily because the styling on the background of the new flashed messages is not green, and because the new flashed messages consume a lot of space on the screen. An alternative approach (that is simpler design wise) could be creating a list of files successfully uploaded as they complete and printing "Files x, y, and z successfully uploaded" in the existing success box.

securedrop/source_templates/lookup.html
```diff
@@ -29,7 +29,7 @@ <h1 class="headline">{{ gettext('Submit Materials') }}</h1>
   <div class="snippet">
     <div class="attachment grid-item center">
       <img class="center" src="{{ url_for('static', filename='i/server_upload.png') }}" width="73px" height="62px">
-      <input type="file" name="fh" autocomplete="off">
+      <input type="file" name="fh[]" autocomplete="off" multiple>
```
redshiftzero: Currently a source can upload single files up to 500 MB in size. After your PR, a source can upload _unlimited_ files, each up to 500 MB in size. My concern here is [file space exhaustion](https://www.owasp.org/index.php/Unrestricted_File_Upload) due to the unrestricted uploads. Maybe we should limit this to 10 files or so - what do you think? Then we can add descriptive text saying "Attach up to 10 files at a time, each up to 500 MB in size", as done in this UI screenshot: http://ui-patterns.com/users/32814/collections/upload-file/screenshots/17805.

securedrop/source_app/main.py
kushaldas: I actually removed the green background from the filenames in the flash messages. We can have a css class to make them look decent. 
I will have to check how to keep flashing messages in the same success box.

securedrop/source_templates/lookup.html
kushaldas: Limiting the actual number of uploads sounds good. I was thinking to do a simple list slicing in the server side too, and have only 10 files as submissions at any given submit operation.

kushaldas:
After searching/talking more on the subject, my current understanding is that without JS we can not have any way to keep showing all the filenames in the same place. 
Also, in the Flask implementation, all the flash messages are actually in a simple list in the `session` object. This also means that all of them goes to next view and then disappears together. 

I would prefer to go for a better design (HTML+CSS) for the filename info boxes. I have asked one of my friend to submit a patch for the same (on top of my work).

securedrop/source_templates/lookup.html
redshiftzero: Hey @kushaldas - it looks like this comment still needs to be addressed.

docs/source.rst
```diff
@@ -100,14 +100,11 @@ Once you have generated a codename and put it somewhere safe, click
...
 
-You will next be brought to the submission interface, where you may
-upload a document, enter a message to send to journalists, or both. You
-can only submit one document at a time, so you may want to combine
-several files into a zip archive if necessary. The maximum submission
-size is currently 500MB. If the files you wish to upload are over that
-limit, we recommend that you send a message to the journalist explaining
-this, so that they can set up another method for transferring the
-documents.
+You will next be brought to the submission interface, where you may upload one
```
redshiftzero: Once the comment below is addressed, you can convert "you may upload one or more documents" -> can become "you may upload up to 10 documents"

securedrop/source_templates/lookup.html
kushaldas: More reading +testing the interface shows that [500MB size limit](https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/app/templates/sites-available/source.conf#L58) is per request, not on per file. Means even if the source uploads 100 files, the total size can not exceeds 500MB per request.

Also because we don't have any JS on the client side, I could not find any way to restrict the maximum number of files to be attached.

kushaldas: The official Apache documentation is [available here](https://httpd.apache.org/docs/2.4/mod/core.html#limitrequestbody).

redshiftzero:
During the last engineering meeting I brought up a security question about this PR: what happens if an attacker submits many small files that exceed the maximum number of open files? 

The block size on my staging instance is 4KB (relevant because this is the smallest file that can be created):

```
root@app-staging:/home/vagrant# dumpe2fs /dev/mapper/vagrant--vg-root | grep -i "block size"
dumpe2fs 1.42.9 (4-Feb-2014)
Block size:               4096
```

And the max number of open files is 1024:

```
root@app-staging:/home/vagrant# ulimit -n
1024
```

with the current 500 MB limit, an attacker can submit sufficient small 4 KB+ files to hit the max limit. I'm curious if 1. this is a valid concern and 2. what happens if they do this?

emkll:
@redshiftzero I actually tested this out:
- set my ulimit to 2048 locally and launched tor browser
- uploaded  ~1900 files (the most my browser instance would let me)

The open file count for `www-data` went over 2000.

The application was still responding on the source interface on another Tor Browser instance, but I suspect a DDoS is still possible.

emkll:
With the help of some trolls, we seem to have reached a limit. This limit is hit when ~8000 files are being uploaded concurrently.

An exception is thrown when writing to the database:
```
[Fri Dec 22 19:26:59.971167 2017] [:error] [pid 1452:tid 3240042301184] [remote 127.0.0.1:8546]     context)
[Fri Dec 22 19:26:59.971169 2017] [:error] [pid 1452:tid 3240042301184] [remote 127.0.0.1:8546]   File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/engine/default.py", line 470, in do_execute
[Fri Dec 22 19:26:59.971172 2017] [:error] [pid 1452:tid 3240042301184] [remote 127.0.0.1:8546]     cursor.execute(statement, parameters)
[Fri Dec 22 19:26:59.971175 2017] [:error] [pid 1452:tid 3240042301184] [remote 127.0.0.1:8546] OperationalError: (sqlite3.OperationalError) database is locked [SQL: u'UPDATE sources SET interaction_count=? WHERE sources.id = ?'] [parameters: (819, 6)]
```
some files become inaccessible to the app server

```
[Fri Dec 22 19:28:30.587491 2017] [core:error] [pid 1453:tid 3240026969856] (13)Permission denied: [client 127.0.0.1:33434] AH00132: file permissions deny server access: /var/www/securedrop/static/css/source.css.map
```

As a result, some connections timeout:
```
[Fri Dec 22 19:30:10.558245 2017] [reqtimeout:info] [pid 1453:tid 3240105457408] [client 127.0.0.1:33441] AH01382: Request header read timeout
```
And the server returns an error, source interface becomes unusable until the files are uploaded/processed:
screenshot from 2017-12-22 15-01-14

However, I am happy to report that such a large volume of submissions does not pose a problem on the journalist interface. 

If there isn't a way to do limit the number of files uploaded without the use of JavaScript, would an acceptable compromise be of making multi-file upload a feature that could easily be toggled on and off ?

securedrop/source_app/main.py
```diff
@@ -104,7 +104,14 @@ def lookup():
...
     def submit():
         msg = request.form['msg']
-        fh = request.files['fh']
+        files = request.files
+        fh = files.getlist('fh[]')
+
+        #  Based on http://flask.pocoo.org/docs/0.12/patterns/fileuploads/
+        #  if user does not select file, browser also
+        #  submit a empty part without filename
+        if fh and not fh[0].filename:
```
redshiftzero: So since `fh` is a list of `FileStorage` objects we can just enforce the max number of files here in Python using the length of `fh` (e.g. `if len(fh) < NUMBER_OF_MAX_FILES` then we proceed with the upload, else we flash a message "You can only upload up to 10 files" or similar). Once that's in place it should address the concern @emkll and I were discussing in this PR and we should be good to go to merge this in! What do you think?

conorsch: That sounds _great_—a fine compromise and requires no JS. Please add! Happy to re-review by hammering away with submissions to confirm we have a limit. 

UX consideration: we may want to inform the uploader about the file max _before_ uploading—but happy to test before commenting more about design implications.

kushaldas: I idea is that because fh is a list in the server side, means we already have all the data in the memory (or open file descriptors). So, forcing on the list, means we are doing it after everything is on the server.

puiterwijk: Do note that with the default Werkzeug `stream_factory`, the files will have be attempted to be generated/opened at the moment you call the first `request.form` or `request.files` (https://github.com/pallets/werkzeug/blob/master/werkzeug/formparser.py#L45).

So what you would need to do is override `werkzeug.wrapper.BaseRequest._get_file_stream` (which you have access to as `flask.request._get_file_stream` I think), to return a different stream_factory (maybe always use BytesIO, or do something smarter like "count and abort" in the stream factory), and this must be setup before `_load_form_data` gets called (i.e. before you access `request.form` or `request.files`).

All this is to work around the fact that I don't see any limiting code in https://github.com/pallets/werkzeug/blob/master/werkzeug/formparser.py#L488.
I think that ideally, Werkzeug/Flask would check the max number of files opened.

(Note that this is just from reading the Werkzeug sources. I have not actually tried this)

kushaldas: I can actually achieve this easily (say take only first 10 files and discard rest) if I directly patch the `formparser.py` (tested on my development vm). I will try to figure out a good way to money patch at the runtime.

kushaldas: I have a patch ready for monkey patching which takes the first 10 files in the request and discards the rest. Will update the PR after I manage to test locally in stage vms.

kushaldas:
@dachary @redshiftzero Can I keep the monkeypatch file and the actual 3 lines of monkey patch code outside of lint? I don't want to make any formatting changes to the money patch as it will easier to maintain along with any future upstream change.

dachary:
> Can I keep the monkeypatch file and the actual 3 lines of monkey patch code outside of lint? 

I'm good with any hack as long is the proper patch is actively worked on and has a good chance to land in the not-too-distant future. For future reference it seems a good idea to create a separate issue to track the progress of the proper patch upstream. I'll set myself a reminder to verify it is making progress every month or so. This is something we should have in a bug tracking software (i.e. tooling to track upstream contributions) but manually doing it also works ;-)

redshiftzero:
Thanks for going down the rabbit hole here :)  I agree with @dachary in that I’m fine with making a modification like this as long as they are minimal, maintainable, and have unit test coverage. If removing linting checks on a file makes the code overall more maintainable, then I’m certainly fine with that. A few things: 

1. Note that [we are currently overriding the default Flask Request](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/request_that_secures_file_uploads.py) with `RequestThatSecuresFileUploads` (in `source_app.__init__` we set `app.request_class = RequestThatSecuresFileUploads` in order to encrypt submissions as they stream in). 
2. It’s not clear to me from the commit history here (in the large diff in 559b4fb) where the modification to `_get_file_stream` is done? It might be cleaner to take a similar approach to the overriding of `RequestThatSecuresFileUploads` if possible and override only the function or method of interest such that we don’t need to commit this entire file. What do you think? Let me know if I’m misunderstanding something on this point.

dachary:
@kushaldas I looked into what could be missing after the rebase I did but did not come up with anything.

<pre>
$ git --no-pager  log --oneline --cherry-mark --right-only old-2161-multifile-upload...2161-multifile-upload
= 61b599a7c Pin to older Travis Trusty image
= c5ec2c4f9 Skips flake8 checks for monkeypatch
= d2b6c0e73 Adds the monkeypatch for file count in Flask request
= fb6519c51 Fixes test for the new CSS and use Unicode in filenames
+ fe8ed3b57 styling: Show individual file submissions inline with ellipsis effect
= 4a0c71414 Escape the filename before showing in the UI
= d77fd6a3b Shows flash message for each file on submission
= 8215702fd Updates docs for #2161 fix
+ 083b95dda Fixes #2161 allows multiple file upload
</pre>

Shows only two commits are different between the old and the new branch. **styling: Show individual file submissions inline with ellipsis effect** is where the conflict was resolved but it apparently did not loose anything, just moved the SASS stanza from one file to another.

<pre>
$ diff -u --ignore-matching-lines '^[^+-]'   <(git log --patch --grep 'styling: Show individual file submissions inline with ellipsis effect' old-2161-multifile-upload) <(git log --patch --grep 'styling: Show individual file submissions inline with ellipsis effect' 2161-multifile-upload)
--- /dev/fd/63	2018-01-08 17:17:44.792162532 +0100
+++ /dev/fd/62	2018-01-08 17:17:44.800162637 +0100
@@ -7,11 +7,11 @@
     Signed-off-by: Saptak Sengupta <saptak013@gmail.com>
     Signed-off-by: Kushal Das <kushal@freedom.press>
 
-diff --git a/securedrop/sass/_base.sass b/securedrop/sass/_base.sass
-index f8a803e6d..3b111e3d4 100644
---- a/securedrop/sass/_base.sass
-+++ b/securedrop/sass/_base.sass
-@@ -333,6 +333,17 @@
+diff --git a/securedrop/sass/modules/_flash.sass b/securedrop/sass/modules/_flash.sass
+index a061b6e62..758394d3d 100644
+--- a/securedrop/sass/modules/_flash.sass
++++ b/securedrop/sass/modules/_flash.sass
+@@ -38,6 +38,17 @@
        i
          color: #D62727
</pre>

The **Fixes #2161 allows multiple file upload** also has a different patch-id but it appears to be the result of changes in the context of the diff. The content of the diff compare equal

<pre>
diff -u --ignore-matching-lines '^[^+-]'   <(git log --patch --grep 'Fixes #2161 allows multiple file upload' old-2161-multifile-upload) <(git log --patch --grep 'Fixes #2161 allows multiple file upload' 2161-multifile-upload)
</pre>

kushaldas:
@redshiftzero Made the changes as suggested by you in last week. Now we have a small patch to carry.

securedrop/source_app/main.py
```diff
@@ -143,19 +163,16 @@ def submit():
             if msg and not fh:
                 html_contents = gettext('Thanks! We received your message.')
             elif not msg and fh:
-                html_contents = gettext('Thanks! We received your document.')
+                html_contents = gettext('Thanks! We received your'
```
pierwill: A source who has only submitted a single file might be concerned when told that "document(s)" have been received.

Might `len(fh) == 1` and `len(fh) >= 1` be used in more `elif` statements to send the appropriate response?

heartsucker:
I attempted to upload 4 files, and this was the result:
screenshot_2018-01-10_23-24-48

If all the files had long names, I wouldn't be able to read all of them which would be annoying if I'm trying to figure out what files did or did not get uploaded. These should always display the full name.

securedrop/source_app/main.py
```diff
@@ -113,27 +120,40 @@ def submit():
...
+                fname = (
+                    store.save_file_submission(
+                        g.filesystem_id,
+                        g.source.interaction_count,
+                        journalist_filename,
+                        file_submitted.filename,
+                        file_submitted.stream))
+                submission = Submission(g.source, fname)
+                db_session.add(submission)
+                db_session.commit()
```
heartsucker: Another note about this. We don't catch any errors, so if somethings goes bad (db failure, FS failure, etc), the source will get the generic 500 error page. I think there are two solutions.

1. Keep going, collect all errors, and display all errors at the end.
2. Halt immediately, display which file erred and which files were not processed.

redshiftzero:
First, thanks for your diligent work exploring how we could safely implement this. Taking a step back here to recap: this is a feature that is a nice-to-have, but is not necessary. As I'm reviewing the diff again, my primary concern is maintainability - we're adding a fair bit of monkeypatched code to Flask to implement this safely. As you point out, we also can't functionally test this due to the lack of Selenium multi-file upload support so this would add to the pre-release manual testing (or we rely only on unit tests). Before investing further work here, we should evaluate whether this feature is worth the burden - after consideration, I'm on the side of probably not, but I can be convinced if people strongly disagree. What do you think @kushaldas? We can also discuss this further in the meeting tomorrow. 

Otherwise I agree with @heartsucker's points. Additionally, we'd also at least need unit test coverage of two additional base cases:
* check a submission with `MAX_FILE_ALLOWED` files is successful (since it's a boundary value), 
* checks a submission with greater than `MAX_FILE_ALLOWED` files fails

redshiftzero:
@kushaldas and I discussed this yesterday - @kushaldas suggested that we first try to get the functionality in Flask and then take another run at this at a later date. Until then closing, and marking the original issue blocked.

PR #2583, opened 2017-11-17
test: use urandom in test

tmc:
## Status

Work in progress

## Description of Changes

## Testing

make app-images app-test

## Deployment
N/A

## Checklist

N/A

### DIFF
--- a/securedrop/test
+++ b/securedrop/test
@@ -8,6 +8,9 @@ redis-server &
 
 echo -n 4096 > /proc/sys/kernel/random/write_wakeup_threshold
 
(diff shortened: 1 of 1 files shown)


dachary:
My interpretation of "Work in progress" is because @tmc wanted to demonstrate the benefit. However there is enough evidence/documentation explaining **urandom** will be less of a blocker than **random**. In the context of tests we do not care about the quality of entropy therefore there is no negative impact if we force every tool to use urandom instead of random, even for GPG or ssh keys.

PR #2585, opened 2017-11-17
Test low entropy

tmc:
## Status

Ready for review

## Description of Changes

Adds test coverage for low-entropy branch.

## Testing

unit tests

## Deployment
N/A

## Checklist
N/A

### DIFF
--- a/securedrop/tests/test_source.py
+++ b/securedrop/tests/test_source.py
@@ -312,6 +312,38 @@ class TestSourceApp(TestCase):
             self.assertIn("Thanks! We received your message and document",
                           resp.data)
 
+    @patch('source_app.main.async_genkey')
+    @patch('source_app.main.get_entropy_estimate')
+    def test_submit_message_with_low_entropy(self, get_entropy_estimate,
+                                             async_genkey):
+        get_entropy_estimate.return_value = 300
+
+        with self.client as client:
+            new_codename(client, session)
+            self._dummy_submission(client)
+            resp = client.post('/submit', data=dict(
+                msg="This is a test.",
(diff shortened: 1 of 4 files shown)


securedrop/tests/test_source.py
```diff
@@ -312,6 +312,38 @@ def test_submit_both(self):
             self.assertIn("Thanks! We received your message and document",
                           resp.data)
 
+    @patch('source_app.main.async_genkey')
+    @patch('source_app.main.get_entropy_estimate')
+    def test_submit_message_with_low_entropy(self, get_entropy_estimate,
+                                             async_genkey):
+        get_entropy_estimate.return_value = 300
```
dachary: this is the only difference between the two tests. It is problematic in the sense that a regression in the code that would make both tests run the same code path would not raise an error. The logs should be examined to assert a different message is issued, or any other proof that the expected code path has run.

securedrop/tests/test_source.py
```diff
@@ -312,6 +312,38 @@ def test_submit_both(self):
...
+
+        with self.client as client:
+            new_codename(client, session)
+            self._dummy_submission(client)
+            resp = client.post('/submit', data=dict(
+                msg="This is a test.",
+                fh=(StringIO(''), ''),
+            ), follow_redirects=True)
+            self.assertEqual(resp.status_code, 200)
+            self.assertFalse(async_genkey.called)
```
tmc: This last assert is the interesting difference.

dachary: Oh.... right.

Issue #2591, opened 2017-11-20
SecureDrop validate role fails on Tails 3.3


msheiny:
# Bug

## Description

Under Tails `3.3` - the validate script will bomb out because it is trying to shell out to `nslookup` but the deb package `dnsutils` is not installed.

A short summary of the issue.

## Steps to Reproduce

Under a fresh tails 3.3:
* clone down securedrop, cd into that directory
* run `./securedrop-admin setup`
* run `./securedrop-admin sdconfig` or `./securedrop-admin install`

## Expected Behavior

You'll get prompted for variables and everything should pass validation.

## Actual Behavior

The validation role bombs out with the following output:

```
TASK [validate : Perform SMTP lookup check.] *******************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "cmd": ["torify", "nslookup", "-vc", "-retry=3", "-timeout=10", "-fail", "mail.electricembers.net", "8.8.8.8"], "delta": "0:00:00.029899", "end": "2017-11-20 16:01:33.483484", "failed": true, "rc": 1, "start": "2017-11-20 16:01:33.453585", "stderr": "ERROR: nslookup cannot be found.", "stderr_lines": ["ERROR: nslookup cannot be found."], "stdout": "", "stdout_lines": []}
...ignoring

TASK [validate : Validate SMTP relay connection.] **************************************************************
fatal: [localhost]: FAILED! => {
    "assertion": false, 
    "changed": false, 
    "evaluated_to": false, 
    "failed": true, 
    "msg": "The SMTP relay domain failed during lookup. This domain is the server contacted for authentication in order to send OSSEC email notifications. You should manually edit the file `group_vars/all/site-specific` and confirm that the `smtp_relay` var is set correctly."
}
	to retry, use: --limit @/home/amnesia/Persistent/securedrop/install_files/ansible-base/securedrop-configure.retry

```

## Comments

We really shouldn't be shellin' out and instead should be taking advantage of either the `dig` lookup or stop using ansible and use python scripts. As the fastest short-term work-around we could install `dnsutils` during the setup step.

conorsch:
[...]

PR #2592, opened 2017-11-20
Allow SSH over local network instead of Tor

### LINKED ISSUE #2091: Allow optional local management of SD servers without tor
# Feature request

## Description

Currently a production securedrop install configures the servers so that they are only manageable remotely over tor. There are particularly use-cases where this could be overkill -- such as when an admin only needs to "remotely" manage the servers on the local network.

We should allow the setting of a boolean flag (disabled by default to leave the behavior as-is) that will disable ssh over tor and restrict ssh by a local admin workstation IP address (we'll throw warnings if users try to set non-RFC1918 addresses).

I realize that there is value to ssh over tor -- but I also see the perspective of: 
> If i don't need to remotely manage SD outside of the local building why am I opening up a remote network service for anyone on tor to access it? Yes I realize it's an authenticated service but its still a possible attack vector that isn't present if I just restrict SSH to the local network

## User Stories

As an admin, I'd like to manage my securedrop servers from my local network and do not require access remotely.

msheiny:
## Status

Ready for review

## Description of Changes

* Fixes #2091  - Allows an admin to enable SSH over local network instead of Tor. 
* Kind of, sorta helps with #2591 on fresh install only -- need a fix in a different PR to permanently address

If an admin toggles this setting (optional, its off by default) though, this PR will:

* not install `tor` on the monitor server
* add in iptables rules restricting SSH to the same interface and IP space that the admin is utilizing to install on. For instance, if the admin has multiple interfaces and comes in over a particular IPv4 addr, the iptables rules will restrict SSH to that interface and to the local /cidr network of that interface's ip address.
* allow admin to set ssh over local network option under `securedrop-admin sdconfig`

## Testing

I've tested this locally under a full provisioning (boot to tails, running `./securedrop-admin`commands). So I could use eyes on vagrant work-flow . CI now takes advantage of this mode (since essentially it was running staging before anyways).

## Deployment

1. Upgrading existing production instances. --- If an admin were to run this against a previously setup pair of instances - this might not close the old Tor ssh exception specifically on the `mon` server. I just thought of that. I don't have a good story here right now. I'd like to push off to another ticket so this PR doesnt get too crazy (if theres no objections).
2. New installs. -- No crazy considerations, admin will just need to answer the prompt at `sdconfig`

## Checklist

I made changes to both system config and docs. 


install_files/ansible-base/roles/restrict-direct-access/templates/rules_v4
```diff
@@ -90,26 +94,20 @@
...
-{% endif %}
+# Permit direct SSH access.
+# Allowed for staging and optionally for production (disables ssh over tor)
+{% for interface in ansible_interfaces -%}
+  {%- set int_details = hostvars[inventory_hostname]['ansible_'+interface].ipv4 -%}
+  {%- set net_string = '/'.join([int_details.network,int_details.netmask]) -%}
+  {%- if ssh_ip|ipaddr(net_string) -%}
+    {%- set ssh_local_net = net_string|ipaddr('cidr') -%}
+    {%- set ssh_int = interface -%}
+      {%- if enable_ssh_over_tor == true or 'staging' in group_names -%}
```
conorsch: The `if enable_ssh_over_tor` conditional should be the governor of whether the entire "Permit direct SSH access" block is entered. Let's move this to the top of the block and iterate over the interfaces only if required.

The `netmask` attributes on ifaces will also require additional pip requirements, but I see you already added those to the admin requirements file.

msheiny: unfortunately -- if you do that logic (only loop over this `if enable_ssh_over_tor`) you are unable to use these values outside that loop. That bit me hard while debugging and I couldnt find a way to set jinja variables global inside a loop

docs/ssh_over_local_net.rst
```diff
@@ -0,0 +1,77 @@
...
+    TASK [debug]
+    ***********************************************************************
+    ok: [localhost] => {
+        "msg": "You will need to fill out the following prompts in order to
+        configure your SecureDrop instance. After entering all prompts, the
+        variables will be validated and any failures displayed. See the docs
+        for more information https://docs.securedrop.org/en/stable"
+        }
+
+    Enable SSH over Tor - alternatively ssh over local net [true]: no
```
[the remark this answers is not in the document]
msheiny: 👍 i really like this wording change

msheiny:
rebased on `develop`

msheiny:
rebased again to fix an issue with a previous rebase ... :|

install_files/ansible-base/roles/restrict-direct-access/templates/rules_v4
msheiny: for reference of similar issue I ran into when trying to add this under a giant conditional at the top previously -- https://stackoverflow.com/questions/17925674/jinja2-local-global-variable

i did just find a post of someone assigning to a dictionary instead though to get around this --> https://stackoverflow.com/questions/9486393/jinja2-change-the-value-of-a-variable-inside-a-loop .

I honestly dont think its a huge deal as-is. can you chip in with more context on why you think the current logic is going to be problematic?

conorsch:
I realize this PR has undergone significant rebasing, but CI is failing at the moment, so I am _not_ re-reviewing. Unlikely to make it into 0.6 before merge window closes @msheiny. :disappointed:

msheiny:
rebased on `develop` . pushed additional fixes. might fail testinfra.... gotta test locally as well

msheiny:
@conorsch ping for re-review. Passing CI now

docs/ssh_over_local_net.rst
```diff
@@ -0,0 +1,78 @@
...
+
+* Allows remote administration outside of the local network
+* Provides anonymity to an administrator while logging into the SecureDrop
+  back-end.
+* Can mitigate against an attacker on your local network attempting to exploit
+  vulnerabilities against the ssh daemon.
+
+Most administrators will need ssh access during the course of running a
+SecureDrop instance and a few times a year for maintanence. So the
+potential short-falls of having ssh over tor aren't usually a big deal.
```
redshiftzero: Docs look great :) One addition: it looks like the comments [here](https://github.com/freedomofpress/securedrop/pull/2592#discussion_r153304245) need addressing

msheiny:
ping @redshiftzero  -- try again ;) I'm surprised yours failed on first pass.. if anything it would always fail on second pass (after iptables has been laid down and server rebooted) -- that would explain why its passing in CI with the current misnamed `s/rules/iprules/` issue you noticed.

install_files/ansible-base/group_vars/securedrop_monitor_server.yml
```diff
@@ -34,7 +34,7 @@ authd_iprules:
 
 # Declare Monitor Server as OSSEC server role.
 ossec_is_server: yes
-authd_rules:
+authd_iprules:
```
redshiftzero: Ahh my bad, I should have been more specific in my comment: `authd_iprules` is defined above already in this file (redefining it like so in this commit will cause the playbook to bail out when it gets [here](https://github.com/freedomofpress/securedrop/blob/a8915e0740caefc4664a509b3f39997c97a090b4/install_files/ansible-base/roles/ossec/tasks/register.yml#L78)). Since these authd rules do not change, we can safely remove this second definition.

msheiny: ahahah im slow. rebased and removed.

securedrop/requirements/develop-requirements.txt
```diff
@@ -62,6 +62,7 @@ marshmallow==2.13.5       # via molecule
 mccabe==0.6.1             # via flake8, pylint
 molecule==2.4.0
 monotonic==1.4            # via fasteners
+netaddr==0.7.19
```
emkll: Given the securedrop-admin refactor, this maybe should be added to [`admin/requirements-ansible.in`](https://github.com/freedomofpress/securedrop/blob/develop/admin/requirements-ansible.in) to avoid getting this error at install time:
```
TASK [restrict-direct-access : Copy IPv4 iptables rules.] ******************************************************
fatal: [app]: FAILED! => {"changed": false, "msg": "AnsibleFilterError: The ipaddr filter requires python-netaddr be installed on the ansible controller"}
fatal: [mon]: FAILED! => {"changed": false, "msg": "AnsibleFilterError: The ipaddr filter requires python-netaddr be installed on the ansible controller"}
```

docs/ssh_over_local_net.rst
```diff
@@ -0,0 +1,78 @@
...
+for most organizations. If you happen to require SSH restricted to the local
+network instead please continue to read.
+
+
+.. _ssh_over_local:
+
+Configuring SSH for local access
+--------------------------------
+
+.. warning:: It is important that your firewall is configured adequately if you
```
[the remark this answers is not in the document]
msheiny: do you think an expansion of docs are really needed? i ask because we only enable ssh via keys - i don't see the increased threat vector here. If someone has physical access to the server room, the entire system can be compromised regardless of ssh status. If they own the firewall, i guess they could hammer on ssh ? They could technically do that if there is a vulnerability in tor and/or an adversary gets ahold of someone's tails stick + persistence creds.

install_files/ansible-base/roles/restrict-direct-access/templates/rules_v4
```diff
@@ -90,26 +94,20 @@
...
--A OUTPUT -o {{ ansible_default_ipv4.interface }} -p tcp --sport 22 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT
--A OUTPUT -p udp --dport 53 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT
--A INPUT -p udp --sport 53 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT
-{% endif %}
-
-{% endif %}
+# Permit direct SSH access.
+# Allowed for staging and optionally for production (disables ssh over tor)
+{% for interface in ansible_interfaces -%}
+  {%- set int_details = hostvars[inventory_hostname]['ansible_'+interface].ipv4 -%}
```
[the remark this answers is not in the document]
msheiny: incoming fix for this one ... :crossed_fingers:

install_files/ansible-base/roles/restrict-direct-access/templates/rules_v4
```diff
@@ -90,26 +94,20 @@
...
+  {%- set net_string = '/'.join([int_details.network,int_details.netmask]) -%}
+  {%- if ssh_ip|ipaddr(net_string) -%}
+    {%- set ssh_local_net = net_string|ipaddr('cidr') -%}
+    {%- set ssh_int = interface -%}
+      {%- if not enable_ssh_over_tor -%}
+-A INPUT -i {{ ssh_int }} -s {{ ssh_net_in_override|default(ssh_local_net) }} -p tcp --dport 22 -m state --state NEW,ESTABLISHED,RELATED -j ACCEPT
```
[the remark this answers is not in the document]
msheiny: awesome feedback! i agreee. incoming fix

securedrop/requirements/develop-requirements.txt
msheiny: dohhhhhhhhhh thanks for pointing this out buddy!!

docs/ssh_over_local_net.rst
```diff
@@ -0,0 +1,78 @@
...
+.. note:: Whichever network you install from will be the one that SSH is
+          restricted to post-install. This will come into play particularly if
+          you have multiple network interfaces.
+
+The workflow for configuration will look like this (answer 'no' or 'false') to
+the following :
+
+.. code:: sh
+
+    $ ./securedrop-admin sdconfig
```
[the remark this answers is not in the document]
msheiny: this should work now

msheiny:
ping @conorsch / @redshiftzero  / @emkll  -- did a few tweaks .. waiting for CI to pass :crossed_fingers: 

The biggest change I made in the last few commits is smooth over the transition story. Meaning, if an installation is already configured over tor, but the setting is changed to switch to ssh over local network, this process actually works. It was kinda broken before.. its not the smoothest experience (requires two runs of the playbook when going from tor-> localnet) but I tried to add messaging to ease this and ive tested multiple times swapping back and forth and it works good.

redshiftzero:
Hmm. So I've tried several times and this PR keeps bailing out for me on first provisioning:  

```
TASK [ossec : Register OSSEC agent.] *******************************************
fatal: [app-staging]: FAILED! => {"changed": true, "cmd": ["/var/ossec/bin/agent-auth", "-m", "10.0.2.15", "-p", "1515", "-A", "app-staging"], "delta": "0:00:00.004416", "end": "2018-03-30 18:50:09.589774", "msg": "non-zero return code", "rc": 1, "start": "2018-03-30 18:50:09.585358", "stderr": "2018/03/30 18:50:09 ossec-authd: INFO: Started (pid: 6916).\n2018/03/30 18:50:09 ossec-authd: Unable to connect to 10.0.2.15:1515", "stderr_lines": ["2018/03/30 18:50:09 ossec-authd: INFO: Started (pid: 6916).", "2018/03/30 18:50:09 ossec-authd: Unable to connect to 10.0.2.15:1515"], "stdout": "", "stdout_lines": []}
```

I'm testing in staging VMs - in `staging.yml`... `enable_ssh_over_tor` is indeed set to `false`. Sorry for this not-very useful feedback, posting in case anyone else has seen this while testing this PR

msheiny:
> Hmm. So I've tried several times and this PR keeps bailing out for me on first provisioning:

This has gotta be a weird virtualbox / kvm testing difference . Hrmmmmmmmm. Okay I'll do some poking on Monday. No need to do more digging! Thanks for testing!

conorsch:
> This has gotta be a weird virtualbox / kvm testing difference 

In VirtualBox under vagrant, eth0 is always reserved for NAT, via 10.0.2.15. See here: https://github.com/hashicorp/vagrant/issues/2093 In SD staging, we create eth1 with a private network so that the OSSEC registration can happen between VMs.

In KVM/libvirt under vagrant, eth0 is an IP assigned to a generated IP (assigned within a block), e.g.

```
vagrant@mon-staging:~$ ip route get 8.8.8.8
8.8.8.8 via 192.168.121.1 dev eth0  src 192.168.121.228 
    cache
```

In SD staging, we also create eth1 with a private network for the OSSEC registration, to be consistent with the other environments. In prod, only one net iface is likely, and the same local IP and iface will be used for both remote calls and OSSEC registration and communication.

I have not taken a close look at the testing problems reported during review of this PR, but based on @redshiftzero and @msheiny's comments, perhaps this'll shed some light on what needs to change.

emkll:
Vars/sdconfig issues were fixed, thanks @msheiny ! However, when running an new install with this branch in prod VM's, the install keeps timing out very early in the install process: 
```
TASK [Gathering Facts] *********************************************************
ok: [app]
ok: [mon]

TASK [Include restrict role early when using ssh over localnet] ****************

TASK [restrict-direct-access : Wait for all Tor hidden services hostname files.] ***************************************************************
failed: [app] (item={u'service': u'source', u'filename': u'app-source-ths'}) => {"changed": false, "elapsed": 300, "item": {"filename": "app-source-ths", "service": "source"}, "msg": "Timeout when waiting for file /var/lib/tor/services/source/hostname"}
```

I suspect this is due to the `restrict-direct-access` [being run as pre-task](https://github.com/freedomofpress/securedrop/blob/22129260ff5fad12d5fe091bd8f5a5a28807b020/install_files/ansible-base/securedrop-prod.yml#L18) Because the first task of the `restrict-direct-access role` is to [wait for a tor config](https://github.com/freedomofpress/securedrop/blob/22129260ff5fad12d5fe091bd8f5a5a28807b020/install_files/ansible-base/roles/restrict-direct-access/tasks/fetch_tor_config.yml), this might explain why it doesn't work on new installs (but probably does work on existing installs).

msheiny:
> However, when running an new install with this branch in prod VM's, the install keeps timing out very early in the install process:

Shit one of the many problems of not testing the prod playbook in CI :( That whole logic needs to be consolidated into one playbook. I thought i had a ticket for this searching.. got an incoming patch for that issue.

Also thanks to @conorsch for illuminating @redshiftzero 's issue with this branch.. got an incoming fix for that :crossed_fingers:  as well

emkll:
Thanks @msheiny , I can confirm the install now completes without error. However, I am unable to ssh into the monitor server. This might be related to the ton of OSSEC notifications I've been receiving for this instance:
```


OSSEC HIDS Notification.
2018 Apr 04 20:07:37

Received From: mon-prod->/var/log/syslog
Portion of the log(s):

Apr  4 20:07:37 mon-prod Tor[2083]: 838 connections have failed:



 --END OF NOTIFICATION
```
Perhaps this has something to do with the Tor configuration?

msheiny:
one thing i just noticed in testing, when going from a ssh over local instance to ssh over tor --> you have to manually restart the servers in order to get tor over ssh working. I'm kind of reluctant to keep going to try and add that logic into ansible cause im worried about how it will affect a fresh ssh over tor install . Any objections to beefing up the documentation to add that note or should i keep going to try and automate it? Uggggggh this feature is sooo annoying.

emkll:
Thanks for the fix @msheiny, that fixed my clean install issue on staging VM, and managed to successfully upgrade as well. Great touch on the immediately enabling local ssh :1st_place_medal: .

I'm still receiving the following ossec alerts from the mon server:
```


OSSEC HIDS Notification.
2018 Apr 05 18:10:09

Received From: mon-prod->/var/log/syslog
Portion of the log(s):

Apr  5 18:10:08 mon-prod Tor[3537]: 61 connections have failed:



 --END OF NOTIFICATION
```
This appears to be due to the new iptables rules on the mon server which no longer have the IP tables rule that allows tcp outbound for tor (`-A OUTPUT -p tcp -m owner -uid-owner 106 -m state --state NEW, RELATED, ESTABLISHED -m comment --comment "tor instance that provides ssh access" -j ACCEPT`). The Tor network cannot bootstrap, and a flood of emails are sent to the admin. I guess you could either uninstall, kill the process or re-introduce those iptables rules. On my upgraded instance, only the following rules allow TCP outbound on 80, 8080, 443 and 587:
```
-A OUTPUT -p tcp -m multiport --dports 80,8080,443 -m owner --uid-owner 0 -m state --state NEW,RELATED,ESTABLISHED -m comment --comment "apt updates" -j ACCEPT
-A OUTPUT -p tcp -m tcp --dport 587 -m owner --uid-owner 108 -m state --state NEW,RELATED,ESTABLISHED -m comment --comment "Allow ossec email alerts out" -j ACCEPT
```

msheiny:
okay @emkll i made some changes to reflect what i learned trying to push it to my production instance on hardware. the only potential potential issue I see is that if someone has an admin network that is behind a NAT and tries to use this feature.. that will totally fall down. Maybe I should update docs to reflect that.

docs/ssh_over_local_net.rst
```diff
@@ -0,0 +1,88 @@
...
+          restricted to post-install. This will come into play particularly if
+          you have multiple network interfaces.
+
+The setting that controls SSH over LAN access is set during the `sdconfig` step
+of the install. Below is an example of what the prompt will look like. You can
+answer either 'no' or 'false' when you are prompted for `Enable SSH over Tor`:
+
+.. code:: sh
+
+    $ ./securedrop-admin sdconfig
```
emkll: I would suggest recommending `./securedrop-admin setup` to pull in the new requirements (netaddr)

emkll:
I've tested disabling ssh over tor and re-enabling it on both prod vms and a hardware instance. I can confirm that the functionality now works as expected. Thanks for the quick fixes @msheiny , LGTM.

msheiny:
Rebased on develop to pull in #3272 fix

conorsch:
Having an exceedingly hard time breaking this, which is fantastic. Going to run through yet another install on clean prod VMs, this time validating that opting out of ssh-over-tor from the get-go works smoothly. Very optimistic this'll get in before the current sprint is over.

install_files/ansible-base/roles/restrict-direct-access/tasks/iptables.yml
```diff
@@ -36,12 +36,39 @@
     - iptables
     - permissions
 
+- name: Determine admin network - first find admin net dev
+  set_fact:
+    admin_dev: "{{ lookup('pipe','/bin/ip r get '+ssh_ip)|regex_search('(?<=dev )\\w+') }}"
```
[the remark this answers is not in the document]
conorsch: I'm actually OK with this, @redshiftzero. Use of the prod VMs has become _highly_ dependent on a virtualized Tails setup, and we'd do well to go more in that direction, e.g. removing the Ansible integration from the Vagrantfile for the "prod" VMs altogether. Holler if you disagree; we can discuss separately.

conorsch:
Repeated runs with both clean-install-without-tor and converted-install worked well. Used prod VMs from virtualized Tails environment exclusively. We'll surely get in more quality time with this puppy during pre-release QA for 0.7, but as it stands, it addresses all the major edges we should accommodate. Fantastic work here, @msheiny.

PR #2615, opened 2017-11-26
tests: one rqworker per session is enough

### LINKED ISSUE #2613: env.py and conftest.py appear to race to start an rqworker process which can cause test failures
# Bug
env.py and conftest.py appear to race to start an rqworker process which can cause test failures

## Description

Working on a minimal and reliable repro but I see in my containerized qubes environment sometimes an rqworker comes up via env.py (not processing the test queue) which prevents the test worker from coming up and can trigger failures.

## Steps to Reproduce

One way to simulate this problem is to run a worker in the fashion env.py does by hand (not listening to the test queue) and then run pytest.

## Expected Behavior
Deterministic test behavior.

## Actual Behavior
Non-deterministic test behavior.

dachary:
## Status

Ready for review

## Description of Changes

The attempt to run another rqworker in securedrop/tests/utils/env.py
does not serve any purpose and races with the run from
securedrop/tests/conftest.py

Fixes #2613

## Testing

If the travis tests pass, it means the change is good and the double run was unecessary.

## Deployment
Nope.

### DIFF
--- a/securedrop/tests/utils/env.py
+++ b/securedrop/tests/utils/env.py
@@ -52,11 +52,6 @@ def setup():
     init_db()
     # Do tests that should always run on app startup
     crypto_util.do_runtime_tests()
-    # Start the Python-RQ worker if it's not already running
-    if not exists(TEST_WORKER_PIDFILE):
-        subprocess.Popen(["rqworker",
-                          "-P", config.SECUREDROP_ROOT,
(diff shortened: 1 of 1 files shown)


dachary:
@tmc ready for your review :-)

tmc:
lgtm!

PR #2603, opened 2017-11-22
Support a `.well-known` www directory for external CA validation

msheiny:
## Status

Ready for review

## Description of Changes

Fixes #2602

Changes proposed in this pull request:

Addresses a scenario when an administrator requires to prove onion ownership to an external CA. Adds a `.well-known` folder to the application server and updates apache and apparmor rules to allow serving arbitrary html files from here. 

## Testing

* Ensure CI is passing (added a test for the debian app code pkg)
* Spin up a staging instance, navigate to the source onion address, drop a file in `/var/www/securedrop/.well-known/` (extension `.html`), and attempt to navigate to that in a tor browser session.

## Deployment

Any special considerations for deployment? Consider both:

1. Upgrading existing production instances.      --- Does not help out with existing production installs fully. An admin will have to re-run the install playbooks to make the appropriate apache changes.
2. New installs. -- Nothing special needs to be done during new deployments.

### If you made changes to the system configuration:

- [ ] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

### If you made changes to documentation:

- [x] Doc linting passed locally

PR #2621, opened 2017-11-27
Explicit docs to reactivate the virtualenv

b-meson:
## Status

Ready for review

## Description of Changes

Small fix to the documentation for admins who want to enable HTTPS on their onion site. Related to #2603 but doesn't need to go in at the same time. 

### DIFF
--- a/docs/https_source_interface.rst
+++ b/docs/https_source_interface.rst
@@ -110,9 +110,10 @@ following: ::
     securedrop_app_https_certificate_chain_src: ca.crt
 
 The filenames should match the names of the files provided to you by DigiCert,
-and should be saved inside the ``install_files/ansible-base/`` directory. Then rerun
-the configuration: ::
+and should be saved inside the ``install_files/ansible-base/`` directory. You'll
+rerun the configuration scripts: ::
 
+    ./securedrop-admin setup
     ./securedrop-admin install
 
 The webserver configuration will be updated to apply the HTTPS settings.


docs/https_source_interface.rst
```diff
@@ -104,9 +104,10 @@ following: ::
     securedrop_app_https_certificate_chain_src: ca.crt
 
 The filenames should match the names of the files provided to you by DigiCert,
-and should be saved inside the ``install_files/ansible-base/`` directory. Then rerun
-the configuration: ::
+and should be saved inside the ``install_files/ansible-base/`` directory. You'll
+reactivate the virtual environment before running the Ansible playbook: ::
```
redshiftzero: The audience we should be catering for in our documentation is a SecureDrop administrator that is not a developer. They shouldn't be expected to know what a virtual environment or an Ansible playbook is, as these things should be abstracted away by `securedrop-admin`. So, let's remove the diff in these two lines. Btw note that `./securedrop-admin install` does reactivate the virtual environment without needing to run `./securedrop-admin setup` first.

However, I think it's a fine practice for admins to run `./securedrop-admin setup` after updating SecureDrop locally in case there are any changes to packages (e.g. #2592 installs `dnsutils` locally as a workaround for #2591).

dachary:
@b-meson a gentle reminder to address @redshiftzero comments :-)

b-meson:
@redshiftzero @dachary I think this PR is ready for review.

conorsch:
Why are these changes necessary? The `install` subcommand already activates the virtualenv under the hood.

b-meson:
@conorsch that was to address the point @redshiftzero made a comment (click show outdated above) that the `setup` command could change in the future (as it did in #2591).

Issue #2099, opened 2017-08-10
Retry SMTP relay and SASL domain verification tasks


redshiftzero:
## Description

The "Validate SASL domain" and "Validate SMTP relay" tasks (in `validate_ossec_info.yml` in the `validate` role) timeout pretty often, so let's either increase the timeout or retry them.

conorsch:
Working on this now. Two quick thoughts on implementation:

* We'll need to stop using `assert` on the lookups because then we can't use `retries` effectively. If we use `command`, then we're OK.
* Consider using `nslookup` rather than `host`, which itself has a `retry=<n>` flag.

@redshiftzero Can you run `nslookup -version` for me and share the results?

redshiftzero:
FYI I don't have a version argument on my `nslookup`:

```
$ nslookup -version
*** Invalid option: version
```

but the `retry=number` option is present.

PR #2622, opened 2017-11-27
Use host for SASL and SMTP domain validation

### LINKED ISSUE #2591: SecureDrop validate role fails on Tails 3.3

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #2591.

  * Replace nslookup with host in `validate` role so dnsutils is not required (fix for #2591).

## Testing

Run `./securedrop-admin sdconfig` on Tails 3.3

## Deployment

Tails only

## Checklist

Tails env, needs manual QA

### DIFF
--- a/install_files/ansible-base/roles/validate/tasks/validate_ossec_info.yml
+++ b/install_files/ansible-base/roles/validate/tasks/validate_ossec_info.yml
@@ -49,23 +49,23 @@
   set_fact:
     _smtp_hostname_query_command: >-
       {% if ansible_lsb.id == "Tails" -%}torify {% endif %}
-      nslookup -vc -retry=3 -timeout=10 -fail {{ smtp_relay }} 8.8.8.8
+      host -W=10 -R=3 -T -4 {{ smtp_relay }} {{ dns_server }}
     _sasl_hostname_query_command: >-
       {% if ansible_lsb.id == "Tails" %}torify {% endif %}
-      nslookup -vc -retry=3 -timeout=10 -fail {{ sasl_domain }} 8.8.8.8
+      host -W=10 -R=3 -T -4 {{ sasl_domain }} {{ dns_server }}
 
 - name: Perform SMTP lookup check.
   command: "{{ _smtp_hostname_query_command }}"
   changed_when: false
   register: smtp_validate_result
-  # We'll inspect return code in subsequent task to provide an informative
-  # failure message.
+  # We'll inspect the output of this command in the next task to see if
+  # an address was found.
   ignore_errors: yes
 
 - name: Validate SMTP relay connection.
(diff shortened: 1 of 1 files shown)


conorsch:
Nice fallback. These changes actually revert to the initial implementation, using `host`, in b7977d13ee2a6f41fe58f060ab22118ab72c1e67. We switched to `nslookup` to satisfy #2099, which we're losing support for here. 

We'll likely want to tweak the `host` options, however, including at least `-T` and `-4`, as described in 785383695c3784755a68ac6f51a8813a3629b9fd. We should also append the `{{ dns_server }}` var in case Admins specify a custom server other than the default of 8.8.8.8.

redshiftzero:
Good calls - added those and `-W=10` (10 second timeout) and `-R=3` (3 retries) so that we get the same behavior as the previous nslookup approach

PR #2638, opened 2017-11-30
Revert "l18n: install gettext on app server"

### LINKED ISSUE #2632: Make sure gettext lands on existing servers
## Description

We just discovered this in a code review  - in #2348 for fuzzy translations we added a more modern version of `gettext` which is installed via Ansible (i.e. this won't get installed on existing servers). We definitely need the modern version of `gettext` in `development` or `build` environments when we do the message extraction and compiling of translations, but we need to ensure:

1. That `gettext` will be installed on existing servers 
2. If we do it via aptly, that the version that is installed on existing servers is sufficient to run the application

dachary:
## Status

## Description of Changes

Fixes #2632

This reverts commit a15f5947e6f0437f759ee321071815279bc29105. It may have been necessary at some point, maybe for building packages.

## Testing

If it passes CI, it is good.

## Deployment

No.


conorsch:
Fantastic! Agreed that the presented changes resolve the concerns mentioned in #2632. 

While we're here, can you comment on why the modern `gettext` is installed from the Zesty repos in `app-test` rather than in the `development` role? Looks like it's because the application test suite includes testing of the `manage.py` i18n subcommands. Specifically, when the `--compile` flag is passed to the `translate` subcommand, the test suite will fail without the Zesty version. Can you confirm, @dachary?

dachary:
> test suite includes testing of the manage.py i18n subcommands. Specifically, when the --compile flag is passed to the translate subcommand, the test suite will fail without the Zesty version. 

I confirm, yes.

PR #2648, opened 2017-11-30
[0.5] Fix #2647: Make sure source interface logout preserves user locale

### LINKED ISSUE #2647: Logout message in localized language and landing page in default language
# Bug
## Description

A source logging out of session in non-default language may yield a home page that has different languages. 

## Steps to Reproduce

0. SecureDrop instance must support more than 1 language
1. Navigate to source interface
2. Change to another language
2. Click on submit document (or equivalent)Change language to non-default language 
4. Logout
5. Observe multilingual SecureDrop login page (Session Ending box in chosen language, Landing page in default language

## Expected Behavior

Logout should preserve user's language settings ?

## Actual Behavior

The session settings are cleared and the user's language selection is not preserved

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #2647.

   * Preserve user locale after logout on the source interface. 

## Testing

0. Open source interface
1. Change language
2. Submit doc
3. Logout
4. Page should not be multilingual:

screen shot 2017-11-30 at 2 15 31 pm

## Deployment

Not yet in prod

### DIFF
--- a/securedrop/source_app/main.py
+++ b/securedrop/source_app/main.py
@@ -221,8 +221,11 @@ def make_blueprint(config):
         if logged_in():
             msg = render_template('logout_flashed_message.html')
 
-            # clear the session after we render the message so it's localized
+            # Clear the session after we render the message so it's localized
+            # If a user specified a locale, save it and restore it
+            user_locale = g.locale
             session.clear()
+            session['locale'] = user_locale
 
             flash(Markup(msg), "important hide-if-not-tor-browser")
         return redirect(url_for('.index'))
--- a/securedrop/tests/test_source.py
+++ b/securedrop/tests/test_source.py
@@ -194,7 +194,9 @@ class TestSourceApp(TestCase):
 
             # sessions always have 'expires', so pop it for the next check
             session.pop('expires', None)
-            self.assertTrue(not session)
+
+            self.assertNotIn('logged_in', session)
+            self.assertNotIn('codename', session)
 
             self.assertIn('Thank you for exiting your session!', resp.data)
 


dachary:
clever :-)

securedrop/source_app/main.py
```diff
@@ -221,8 +221,12 @@ def logout():
         if logged_in():
             msg = render_template('logout_flashed_message.html')
 
+            # save locale so we can restore it
+            user_locale = session['locale']
```
dachary: The locale key does not always exist.

redshiftzero: oof, good call... `en_US` default it is

dachary: it would be better to just not set it to avoid hardcoding the en_US default here

dachary: if not set it should fallback to the DEFAULT_LOCALE set in config.py

redshiftzero:
Working on this, will ping when ready for review

redshiftzero:
OK ready - I set it to use first `config.DEFAULT_LOCALE` and then as a fallback `en_US`. The reason I set the fallback to `en_US` is since `DEFAULT_LOCALE` was set via Ansible's `lineinfile`, it might not be in place for some existing installs.

securedrop/source_app/main.py
```diff
@@ -221,8 +221,12 @@ def logout():
         if logged_in():
             msg = render_template('logout_flashed_message.html')
 
-            # clear the session after we render the message so it's localized
+            # Clear the session after we render the message so it's localized
+            # If a user specified a locale, save it and restore it
+            default_locale = getattr(config, 'DEFAULT_LOCALE', 'en_US')
```
dachary: I think it would be enough to not set locale at all, even if DEFAULT_LOCALE does not exist, because [en_US will be used as a last resort](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/i18n.py#L100). Unless you found a case where it fails ?

redshiftzero: True, it seems like using `g.locale` here would satisfy all concerns, it will always be defined and will have the default set via `g.locale = i18n.get_locale()`. Modifying...

PR #2649, opened 2017-11-30
[0.5] app: use lazy_gettext instead of gettext in wtforms

### LINKED ISSUE #2644: Missing translation for 'This field is required' across all languages
# Bug

## Description

When navigating to `/login` and pressing submit on an empty, the default (English) error message string is displayed.

## Steps to Reproduce

1. Click on the localized equivalent of 'Check for a response`
2. Click submit (while leaving an empty string)
3. Observe the error message.

## Expected Behavior
The error message should be displayed in the localized language.

## Actual Behavior

The message is displayed in English.

dachary:
## Status

Ready for review

## Description of Changes

Fixes #2644

The wtforms classes are evaluated outside of the context of a
request. When gettext is used to set a parameter, it has no way to
know which language is going to be used and default to english.

We need to use lazy_gettext which delays the evaluation until the
string is used, i.e. when the user session is active and the language
preferences can influence the translation.

## Testing

* vagrant ssh development
* cd /vagrant/securedrop
* ./manage.py run
* firefox http://localhost:8080/login
* switch to non default language
* type enter for an empty passphrase
* verify the error message is not displayed in english

## Deployment

User interface only.

### DIFF
--- a/securedrop/journalist_app/forms.py
+++ b/securedrop/journalist_app/forms.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from flask_babel import gettext
+from flask_babel import lazy_gettext
 from flask_wtf import FlaskForm
 from wtforms import (TextAreaField, TextField, BooleanField, HiddenField,
                      ValidationError)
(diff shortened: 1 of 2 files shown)


redshiftzero:
Test failure looked like a flake, restartin'

redshiftzero:
oh this is into `develop` not `release/0.5`, will make a PR to move your commit over

dachary:
Shoooot. My bad :-(

PR #2671, opened 2017-12-04
docs: development: require commits to be signed

dachary:
## Status

Ready for review

## Description of Changes

Require commits to be signed by contributors.

## Testing

N/A

## Deployment

N/A

### DIFF
--- a/docs/development/contributor_guidelines.rst
+++ b/docs/development/contributor_guidelines.rst
@@ -1,6 +1,12 @@
 Contributing Guidelines
 =======================
 
+Signing commits
+---------------
+
+Commits should be signed, as `explained in the GitHub documentation <https://help.github.com/articles/signing-commits-using-gpg/>`_.
+This helps verify commits proposed in a pull request are from the expected author.
+
 Branching Strategy
 ------------------
 


docs/development/contributor_guidelines.rst
```diff
@@ -1,6 +1,12 @@
 Contributing Guidelines
 =======================
 
+Signing commits
+---------------
+
+Commits must be signed, as `explained in the GitHub documentation <https://help.github.com/articles/signing-commits-using-gpg/>`_.
+This helps verify commits proposed in a pull request are from a the expected author.
```
redshiftzero: My concern with this is that it further increases the technical bar for contributors and may deter contributions - how about we have this as a recommendation but not a requirement?

conorsch: Agreed on requirement—`s/must/should/` would suffice. Also, typo in "from a the": remove "a ".

Even if we enforce signing across the board for all commits, we're still losing full integrity by clicking the "merge" button on the GitHub interface, since the merge commit isn't signed. Still, for PR review, having commits marked as verified due to valid signatures is a Good Thing.™

dachary:
Repushed with a should instead of must, you're right, this was a bit much ;-)

docs/development/contributor_guidelines.rst
dachary: done

conorsch:
@dachary Remember that docs-only PRs can skip staging CI by prefixing the branch name with `docs-`! https://docs.securedrop.org/en/stable/development/documentation_guidelines.html#testing-documentation-changes

dachary:
> Remember that docs-only PRs can skip staging CI by prefixing the branch name with

Thanks for reminding me for the Nth time... I'll remember it this time 😊

PR #2676, opened 2017-12-05
[0.5] l18n: arabic is postponed to the next version

dachary:
## Status

Ready for review

## Description of Changes

Arabic needs a little more work, let's not include it in 0.5

### Development VM

For each test case below repeat the following:

* ./securedrop-admin -d sdconfig
* enter the value when prompted for **Space separated list of supported locales**
* the validation will fails because it is not Tails but validation of **securedrop_supported_locales** is done
* ANSIBLE_ARGS="--tags securedrop_config" vagrant provision development
* verify that the task **Update SUPPORTED_LOCALES in config.py** runs successfully
* verify that securedrop/config.py has a SUPPORTED_LOCALES setting matching install_files/ansible-base/group_vars/all/site-specific
* remove the **securedrop_supported_locales** line in install_files/ansible-base/group_vars/all/site-specific

Test cases:

* empty string => securedrop_supported_locales: [] => passes validation
* nl     fr_FR (separated by multiple white space) => [nl, fr_FR] => passes validation
* foobar nl => [foobar, nl] => fails validation on foobar

## Deployment

Same as https://github.com/freedomofpress/securedrop/pull/2524 The variable **SUPPORTED_LOCALES** is added to config.py (when it exists or when it is new). 


redshiftzero:
- [x] Verified that Arabic now fails validation (as does `foobar nl`):

```
TASK [validate : Verify supported locales exist] ***************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "rc": 1, "stderr": "", "stdout": "securedrop/translations/ar does not exist\n", "stdout_lines": ["securedrop/translations/ar does not exist"]}
```

- [x] Verified running with supported locales as an empty string passes validation and produces `en_US` as the only supported locale in `config.py`

- [x] Verified running with supported locales set to `nl` and `fr_FR` passes validation and produces `nl`, `fr_FR` and `en_US` as the supported locales.

PR #2684, opened 2017-12-06
i18n: do not expose unsupported locales

### LINKED ISSUE #2681: Disable languages that are not supported
# Bug

## Description

The source and journalist interface can be switched to a non-supported language by:

* l=fr_FR : this is ok as the person doing that knows details that are unlikely to confuse anyone
* Accept-Languages is set to fr_FR : this is not good because it will lead the person into thinking French is supported

## Steps to Reproduce

* Set the SUPPORTED_LOCALES to ['en_US'] only
* Change the preferred language of your browser to French and visit the source interface

## Expected Behavior

* The page displays in English

## Actual Behavior

* The page displays in French

dachary:
## Status

## Description of Changes

Fixes #2681

The SUPPORTED_LOCALES list is enforced so the user cannot select a
language that it does not contain. The restriction based on the
SUPPORTED_LOCALES variable is done at setup and used to narrow the
content of the LOCALES variable.

The SUPPORTED_LOCALES content was previously only used to select the
languages displayed via the language menu. This is incorrect because
it would allow a user to add ?l=unsupported to see the interface in
one of the available languages even if it is not in the
SUPPORTED_LOCALES list. This was also true if the browser it configured
to prefer a given locale.

## Testing

* Set the SUPPORTED_LOCALES to ['en_US'] only
* Change the preferred language of your browser to French and visit the source interface
* Visit the source interface and manually add **l=fr_FR** to the URL
* In both cases the interface must show in English

## Deployment

Only changes the locale selection logic, no impact on deployment.

## Comment

The **i18n: do not expose unsupported locales** commit is standalone. In the spirit of minimizing the changes for 0.5.1 it can be cherry-picked.

### DIFF
--- a/securedrop/journalist_app/admin.py
+++ b/securedrop/journalist_app/admin.py
@@ -70,7 +70,8 @@ def make_blueprint(config):
                 return redirect(url_for('admin.new_user_two_factor',
                                         uid=new_user.id))
 
-        return render_template("admin_add_user.html", password=make_password(),
+        return render_template("admin_add_user.html",
+                               password=make_password(config),
                                form=form)
 
     @view.route('/2fa', methods=('GET', 'POST'))
@@ -171,7 +172,7 @@ def make_blueprint(config):
 
             commit_account_changes(user)
 
-        password = make_password()
+        password = make_password(config)
         return render_template("edit_account.html", user=user,
                                password=password)
 
--- a/securedrop/i18n.py
+++ b/securedrop/i18n.py
@@ -20,7 +20,6 @@ from flask_babel import Babel
 from babel import core
 
 import collections
-import config
 import os
 import re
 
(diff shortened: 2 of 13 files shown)


securedrop/journalist_app/admin.py
```diff
@@ -171,7 +172,7 @@ def edit_user(user_id):
 
             commit_account_changes(user)
 
-        password = make_password()
+        password = make_password(config)
```
redshiftzero: Thought: if during `__init__.create_app()` we replaced `app.config.from_object(config.SourceInterfaceFlaskConfig)` with `app.config.from_object(config)` then we could use the application instance `app.config['MY_VAR']` (or in a blueprint `current_app.config['MY_VAR']`) to access config instead of passing `config` around.

dachary: thanks for the tip. I need to educate myself a bit more.

PR #2690, opened 2017-12-07
Change "password" to "passphrase" throughout docs

### LINKED ISSUE #2176: Consistent usage of password/passphrase in docs/code
# Feature request

## Description

The docs and the code should consistently use password or passphrase as much as possible. Currently it seems like a hodgepodge of both. There may be a reason for this, but it seems like this could be consolidated down to just "passphrase" everywhere.

pierwill:
## Status

Ready for review.

## Description of Changes

Contributes to #2176 (in docs, at least).

Changed "password" to "passphrase" throughout the main docs.

I tried to make sure nothing creeped in that didn't sound right. For instance: I didn't change "password manager" to "passphrase manager". I also left the tails upgrade and tails admin docs untouched to avoid changing anything where **password** appears onscreen.

### DIFF
--- a/docs/servers.rst
+++ b/docs/servers.rst
@@ -189,10 +189,10 @@ firewall. If you did not, adjust these settings accordingly.
 Continue the installation
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You can choose whatever username and password you would like. To make
-things easier later you should use the same username and same password
-on both servers (but not the same password as username). Make sure to
-save this password in your admin KeePassX database afterwards.
+You can choose whatever username and passphrase you would like. To make
+things easier later you should use the same username and same passphrase
+on both servers (but not the same passphrase as username). Make sure to
+save this passphrase in your admin KeePassX database afterwards.
 
 Click 'no' when asked to encrypt the home directory. Then configure your
 time zone.
@@ -215,7 +215,7 @@ information on them stays private in case they are seized or stolen.
 While FDE can be useful in some cases, we currently do not recommend
 that you enable it because there are not many scenarios where it will be
 a net security benefit for SecureDrop operators. Doing so will introduce
-the need for more passwords and add even more responsibility on the
+the need for more passphrases and add even more responsibility on the
 admin of the system (see `this GitHub
 issue <https://github.com/freedomofpress/securedrop/issues/511#issuecomment-50823554>`__
 for more information).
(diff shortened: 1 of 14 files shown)


docs/configure_admin_workstation_post_install.rst
```diff
@@ -39,7 +39,7 @@ with these scripts (``~/Persistent/securedrop/``), and run the install script:
 
     ./securedrop-admin tailsconfig
 
-Type the Administration Password that you selected when starting Tails and hit
+Type the Administration Passphrase that you selected when starting Tails and hit
```
dachary: This one we must keep because it is a reference to tails as shown in the greeter screen below.

a

pierwill: Ah good catch :)

pierwill:
Pushed that change.

conorsch:
Changes as presented apply _only_ to docs, as @pierwill pointed out. As such, does not satisfy #2176. Although the docs are now somewhat inconsistent with the terminology in the code base, it's more important that we update the phrasing in the docs, since that's what _most_ people will be reading.

Making comparable changes to the code base will involve heavy lifting:

```
$ grep -ri password securedrop -c | cut -d: -f2 | paste -sd+ | bc
462
```

Let's leave that aside for now, and merge as-is. #2176 will be closed automatically on merge of this PR. I'm OK with that.

docs/servers.rst
```diff
@@ -189,10 +189,10 @@ firewall. If you did not, adjust these settings accordingly.
 Continue the installation
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You can choose whatever username and password you would like. To make
-things easier later you should use the same username and same password
-on both servers (but not the same password as username). Make sure to
-save this password in your admin KeePassX database afterwards.
+You can choose whatever username and passphrase you would like. To make
+things easier later you should use the same username and same passphrase
+on both servers (but not the same passphrase as username). Make sure to
```
b-meson: this is awkward considering now. I think most *nix users set usernames but then set passwords. I dont know of any cases where a *nix user has set a passphrase as a username for the servers. Can we reword this?

b-meson:
Made one small comment to the PR.

docs/servers.rst
dachary: I'll merge this as is because it is consistent. I agree with you that someone thinking "passphrase" is very unlikely to use the username as a "passphrase". But this is a useful warning, IMHO.

PR #2699, opened 2017-12-08
these docs add support for a clean OS X python install

### LINKED ISSUE #2698: OS X developer guide does not work
# Bug

## Description

Running through the OS X developer guide throws a stacktrace and doesn't actually work on a fresh, new computer.

## Steps to Reproduce

In a new OS X machine run the `pip install -U virtualenvwrapper` described [here](https://raw.githubusercontent.com/freedomofpress/securedrop/develop/docs/development/setup_development.rst).

## Expected Behavior

Following the developer guide should let a new developer get their dev environment working.

## Actual Behavior

See below stacktrace
```
Installing collected packages: virtualenv, pbr, six, stevedore, virtualenv-clone, virtualenvwrapper
Exception:
  File "/Library/Python/2.7/site-packages/pip-9.0.1-py2.7.egg/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/Library/Python/2.7/site-packages/pip-9.0.1-py2.7.egg/pip/commands/install.py", line 342, in run
    prefix=options.prefix_path,
  File "/Library/Python/2.7/site-packages/pip-9.0.1-py2.7.egg/pip/req/req_set.py", line 784, in install
    **kwargs
  File "/Library/Python/2.7/site-packages/pip-9.0.1-py2.7.egg/pip/req/req_install.py", line 851, in install
    self.move_wheel_files(self.source_dir, root=root, prefix=prefix)
  File "/Library/Python/2.7/site-packages/pip-9.0.1-py2.7.egg/pip/req/req_install.py", line 1064, in move_wheel_files
    isolated=self.isolated,
  File "/Library/Python/2.7/site-packages/pip-9.0.1-py2.7.egg/pip/wheel.py", line 345, in move_wheel_files
    clobber(source, lib_dir, True)
  File "/Library/Python/2.7/site-packages/pip-9.0.1-py2.7.egg/pip/wheel.py", line 323, in clobber
    shutil.copyfile(srcfile, destfile)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/shutil.py", line 83, in copyfile
    with open(dst, 'wb') as fdst:
IOError: [Errno 13] Permission denied: '/Library/Python/2.7/site-packages/virtualenv.pyc'
```

b-meson:
## Status

Ready for review

## Description of Changes

Fixes #2698.

## Testing

Ran through a clean SecureDrop install (from scratch including VirtualBox, Vagrant, python etc) on a brand new OS X machine. 

### DIFF
--- a/docs/development/setup_development.rst
+++ b/docs/development/setup_development.rst
@@ -122,7 +122,7 @@ different version, the path to ``virtualenvwrapper.sh`` will differ. Running
 .. code:: sh
 
     sudo easy_install pip # if you don't already have pip
-    pip install -U virtualenvwrapper
+    sudo -H pip install -U virtualenvwrapper --ignore-installed six
     source /usr/local/bin/virtualenvwrapper.sh
     mkvirtualenv -p python2 securedrop
 


docs/development/setup_development.rst
```diff
@@ -122,7 +122,7 @@ different version, the path to ``virtualenvwrapper.sh`` will differ. Running
 .. code:: sh
 
     sudo easy_install pip # if you don't already have pip
-    pip install -U virtualenvwrapper
+    sudo -H pip install -U virtualenvwrapper --ignore-installed six
```
redshiftzero: Why do you need `sudo`?

b-meson: This will not work without `sudo`. You get a permission denied error.  That's why you pass the `-H` flag, to install in the users' home directory. Correct me if I'm wrong @redshiftzero but I think pip recommends _not_ using sudo with pip but in this case I think the correct behavior is to use `sudo` and `-H`.

I think this is the upstream [issue](https://github.com/pypa/pip/issues/3165)

conorsch: The traceback in #2698 showed attempts to write the pip libs to `/Library/Python/2.7/site-packages/`, thus the root perms. Personally, I use `pip install --user` to avoid having to elevate perms for pip libs. Not sure that'll work out of the box on macOS, though.

Might be worthwhile to recommend a homebrew workflow, if possible.

The `--ignore-installed six` strikes me as a more glaring problem, particularly because there's no explanation for it.

b-meson: @conorsch I looked for a homebrew solution for pip install virtualenvwrapper and couldn't easily find one.

redshiftzero: Did you try `brew install pyenv-virtualenvwrapper`? I think that should work and obviate the need for `sudo pip`

conorsch: Must defer to folks using macOS full-time, but I vastly prefer a homebrew workflow for bootstrapping the dev env. Reasons homebrew is rad:

* common for developers to have it already
* setup docs can be linked to, so we don't have to maintain
* no sudo required to install pip libs (warrants further testing)

Let's shoot for a homebrew solution and fall back from there if possible. @redshiftzero I think you're honestly the best candidate for nailing these steps down, mind creating a fresh env—possibly in a new user—and running through the doc?

redshiftzero: Fair, I'll put this in my queue for today

redshiftzero:
brew install worked for me locally, so added a commit making that change

b-meson:
Hmm, I did a pip uninstall of `virtualenvwrapper` and tried to `brew install pyenv-virtualenvwrapper` and the `virtualenvwrapper.sh` script doesn't exist. This is on a brand new user I created (named python). 

```
zerohistory:~ python$ source /usr/local/bin/virtualenv
virtualenv        virtualenv-clone  
zerohistory:~ python$ source /usr/local/bin/virtualenv
```

redshiftzero:
OK if you're still having trouble let's just leave it as is since at least one other contributor reported success. Restamping

PR #2706, opened 2017-12-08
Admin guide: Add basic commands for admins SSHed into the servers

redshiftzero:
## Status

Ready for review

## Description of Changes

 * Our admin guide is very sparse and we've had support requests to provide a basic reference for admins when they are SSHed into their SecureDrop server
 
Note that this is not intended to be exhaustive

## Testing

Render docs locally, give it a read through

## Deployment

Docs only


b-meson:
Looks great to me @redshiftzero

PR #2708, opened 2017-12-09
Reword source template text

pierwill:
## Status

Work in progress

## Description of Changes

I've reworded some text that appears on the codename generator page. I think temporally it makes more sense for the reader this way.

## Testing

I've called this a work in progress since I've edited code and I'm not familiar with the tests, etc.

### DIFF
--- a/securedrop/source_templates/generate.html
+++ b/securedrop/source_templates/generate.html
@@ -32,7 +32,7 @@
 <hr class="no-line">
 
 <p class="explanation">{{ gettext('Because we use none of the traditional means to track users of our <strong>SecureDrop</strong>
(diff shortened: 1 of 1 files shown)


dachary:
@pierwill looks better to me ! But I'll let a native english speaker review :-)

b-meson:
This also look good to be but I'm not sure what needs to change from a functional point of view since this is changing strings used by the app code.

dachary:
> what needs to change from a functional point of view since this is changing strings used by the app code.

Nothing more than [updating the translations](https://docs.securedrop.org/en/latest/development/i18n.html) :-)

PR #2714, opened 2017-12-10
Automate post Ubuntu install checks

### LINKED ISSUE #2129: Automate post-Ubuntu install checks verifying DNS, NTP is working
## Description

Porting the idea suggested by @conorsch [in a comment here](https://github.com/freedomofpress/securedrop/issues/2094#issuecomment-321684537) to its own ticket. 

We should automate these three steps:

```
host freedom.press
sudo apt-get update
sudo ntpdate ntp.ubuntu.com
```

which we currently have admins do manually on both app and mon

## User Stories

As a user installing SecureDrop, I want to minimize the manual steps I must do in my install so that the install is fast and error free

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #2129.

 * Replace a few manual steps that need to be done on each server with tasks in the `common` ansible role to reduce chance of admin error

## Testing

`vagrant provision /staging/`

## Deployment

No special concerns

### DIFF
--- a/docs/servers.rst
+++ b/docs/servers.rst
@@ -352,31 +352,3 @@ the below commands. You should not be prompted for a passphrase
     app
     $ ssh <username>@<Monitor IP address> hostname
     mon
-
-Minor Admin Tasks
------------------
-
-DNS
-~~~
-
-The network firewall rules are set up to disable DNS traffic to the gateway,
-so if your system has not set nameservers, DNS queries will fail. You can
-test this by running ``host freedom.press``. If the host isn't found,
-or there is some other sort of failure, check the pfSense logs. You may see
-UDP traffic to the gateway on port 53 being blocked.
-
-If this is the case, you need add the following lines to ``/etc/resolvconf/resolv.conf.d/tail``
-
-.. code::
-
-    nameserver 8.8.8.8
-    nameserver 8.8.4.4
-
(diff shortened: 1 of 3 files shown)


dachary:
This looks like a legitimate race:

<pre>
    fatal: [mon-staging]: FAILED! => {"changed": false, "cmd": ["ntpdate", "ntp.ubuntu.com"], "delta": "0:00:00.003812", "end": "2017-12-10 04:10:12.525629", "failed": true, "rc": 1, "start": "2017-12-10 04:10:12.521817", "stderr": "10 Dec 04:10:12 ntpdate[1824]: the NTP socket is in use, exiting", "stderr_lines": ["10 Dec 04:10:12 ntpdate[1824]: the NTP socket is in use, exiting"], "stdout": "", "stdout_lines": []}
</pre>

redshiftzero:
If `ntpd` is started then the socket `ntpdate` wants to use will already be in use - but that's okay because the time will be accurate. So, I used `ignore_errors` on the first NTP task and added a subsequent task in f520ec4 that will inspect the exit code and `stderr` of the `ntpdate` such that:

* On first run, then `ntpdate` will have an exit code of 0, both tasks pass. 
* On subsequent runs, `ntpdate` will have a non-zero exit code since the socket is in use, both tasks pass.
* If `ntpdate` fails for _some other reason_, then the playbook will fail.

PR #2717, opened 2017-12-12
docs: abstract firewall rules definition

dachary:
## Status

Ready for review

## Description of Changes

It is not uncommon for network engineers working in a media to have
the necessary expertise to translate the firewall recommendations and
implement them on the hardware they already have.

It is error prone to browse the long description using the pfSense web
interface and figure out what needs to be implemented in an abstract
way.

A paragraph is added, precisely describing the intended rules, for the
benefit of skilled network engineers and in the spirit of minimizing
errors.

## Testing

Carefully read the pfSense instructions and verify nothing is missing from the abstract description.

## Deployment

N/A

### DIFF
--- a/docs/network_firewall.rst
+++ b/docs/network_firewall.rst
@@ -7,7 +7,11 @@ USB* to access the Network Firewall's web interface for configuration.
 
 Unfortunately, due to the wide variety of firewalls that may be used, we
 do not provide specific instructions to cover every type or variation in
-software or hardware. This guide is based on pfSense, and assumes your
+software or hardware.
+However, if you have the necessary
+expertise, we provide `abstract firewall rules`_ that can be
+implemented with iptables, Cisco IOS etc.
+This guide is based on pfSense, and assumes your
 firewall hardware has at least three interfaces: WAN, LAN, and OPT1. For
 hardware, you can build your own network firewall (not covered in this
 guide) and `install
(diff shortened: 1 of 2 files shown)


docs/network_firewall.rst
```diff
@@ -578,3 +582,25 @@ Once it is complete, you will see a notification of successful upgrade:
...
+The pfSense instructions using the web interface can also be precisely
+described as follows:
+
+* Disable DHCP (in case the firewall is providing a DHCP server by default)
+* Disallow all traffic by default (inbound or outbound)
+* Allow UDP OSSEC (port 1514) from *Application Server* to *Monitor Server*
+* Allow TCP ossec agent auth (port 1515) from *Application Server* to *Monitor Server*
+* Allow TCP/UDP DNS from *Application Server* and *Monitor Server* to the IP of know name servers
+* Allow UDP NTP from *Application Server* and *Monitor Server* to all
+* Allow TCP any port from *Monitor Server* to all
```
redshiftzero: TCP outbound on all ports is also needed for the application server - also I would add a specific note in parentheses for both these TCP onbound on all port that it's for Tor, e.g. "this is needed for making connections to the Tor network" since administrators often go off-script on this point.

docs/network_firewall.rst
```diff
@@ -578,3 +582,25 @@ Once it is complete, you will see a notification of successful upgrade:
...
+-----------------------
+
+The pfSense instructions using the web interface can also be precisely
+described as follows:
+
+* Disable DHCP (in case the firewall is providing a DHCP server by default)
+* Disallow all traffic by default (inbound or outbound)
+* Allow UDP OSSEC (port 1514) from *Application Server* to *Monitor Server*
+* Allow TCP ossec agent auth (port 1515) from *Application Server* to *Monitor Server*
+* Allow TCP/UDP DNS from *Application Server* and *Monitor Server* to the IP of know name servers
```
redshiftzero: `know` -> `known`

docs/network_firewall.rst
```diff
@@ -578,3 +582,25 @@ Once it is complete, you will see a notification of successful upgrade:
...
+* Allow UDP NTP from *Application Server* and *Monitor Server* to all
+* Allow TCP any port from *Monitor Server* to all
+* Allow TCP 80/443 from *Application Server* and *Monitor Server* to all
+* Allow TCP ssh from *Admin Workstation* to all
```
redshiftzero: Instead of all, this should be to the Application Server and Monitor Server only

docs/network_firewall.rst
```diff
@@ -578,3 +582,25 @@ Once it is complete, you will see a notification of successful upgrade:
...
+described as follows:
+
+* Disable DHCP (in case the firewall is providing a DHCP server by default)
+* Disallow all traffic by default (inbound or outbound)
+* Allow UDP OSSEC (port 1514) from *Application Server* to *Monitor Server*
+* Allow TCP ossec agent auth (port 1515) from *Application Server* to *Monitor Server*
+* Allow TCP/UDP DNS from *Application Server* and *Monitor Server* to the IP of know name servers
+* Allow UDP NTP from *Application Server* and *Monitor Server* to all
+* Allow TCP any port from *Monitor Server* to all
+* Allow TCP 80/443 from *Application Server* and *Monitor Server* to all
```
redshiftzero: TCP 80/443 is _only_ needed for the admin workstation (to access the pfSense web interface), otherwise this should be disabled (a common misconfiguration is TCP 80/443 inbound to the application and monitor servers because people forget that all the webserver traffic is over Tor onion services)

docs/network_firewall.rst
dachary: done

docs/network_firewall.rst
dachary: I changed to ```Allow TCP 80/443 from *Admin Workstation* to all (in case there is a need to access the web interface of the firewall)``` is that better ?

docs/network_firewall.rst
dachary: done

docs/network_firewall.rst
dachary: I'm assuming you meant **outbound** instead of **onbound** :-) Added the suggested wording: thanks !

dachary:
fixed and repushed 🍨

PR #2720, opened 2017-12-12
docs: better installation/update instructions for the custom logo

dachary:
## Status

Ready for review

## Description of Changes

It is confusing that the custom logo is mentionned twice in the
list of elements to prepare before running sdconfig, under two
slightly different names (custom header vs custom logo).

The details about the custom logo (size) are better included in a
dedicated section to keep the list short.

## Testing

N/A

## Deployment

N/A

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -44,6 +44,22 @@ parentheses. When the source interface is displayed in French (for
 instance), people submitting documents will expect a journalist fluent
 in French is available to read them and followup.
 
+Custom logo for the source and journalist interface
+---------------------------------------------------
+
+An image to replace the SecureDrop logo on the *Source Interface* and
+*Journalist Interface* can be provided during the installation. It
+should be prepared in advance and available from the *Admin
+Workstation*.
+
+- Recommended size: ``500px x 450px``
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -44,6 +44,22 @@ parentheses. When the source interface is displayed in French (for
 instance), people submitting documents will expect a journalist fluent
 in French is available to read them and followup.
 
+Custom logo for the source and journalist interface
+---------------------------------------------------
+
+An image to replace the SecureDrop logo on the *Source Interface* and
+*Journalist Interface* can be provided during the installation. It should
+be prepared in advance and available from the *Admin Workstation*, either
+via an URL or from the persistent tails directory.
```
redshiftzero: can we remove "either via an URL or from the persistent tails directory." - I feel like people _might_ get confused and think they can provide just a URL instead of downloading the image

otherwise this all looks excellent! (and this is an extremely common source of confusion, so these docs will be really helpful :tada: 💃 )

dachary:
@redshiftzero modified and repushed: thanks for the suggestion, it was indeed confusing 🥇

Issue #1681, opened 2017-04-28
Get Prod environment testing running in CI


msheiny:
Depends on first getting CI specific builds of grsecurity in the pipeline, as currently under discussion outside of Github issues. We are also working on the best way to handle forked PRs. 

I have a high-level design of where I think the production testing should fit into our pipeline. This is based on how long the tests take though and whether we are able to get them running in parallel to the staging tests with ease.
![index](https://cloud.githubusercontent.com/assets/1727935/25544689/997ecb7e-2c29-11e7-9636-957a4e8d8922.png)

zenmonkeykstop:
Discussed in backlog grooming session today. Agreed to close as the only real testing benefit over current staging config would be a more accurate OSSEC configuration.

PR #2729, opened 2017-12-12
move meta dep. from grsecurity role -> playbook

### LINKED ISSUE #2728: Install-fpf-repo role should not be a dependency of grsecurity role
# Feature request

## Description

Ansible meta dependencies (as defined in a role's `meta/main.yml`) should not be used in securedrop. They are annoying to track down for developers coming back in later, trying to determine what is being run.

Let's make sure all role calls are done at the playbook level.

## User Stories

As a securedrop maintainer, I want to immediately be able to look at a role playbook and visually see all roles that are being called.

msheiny:
## Status

Ready for review

## Description of Changes

During a security audit of trying to track down dependencies, this was
overlooked at first pass and makes the logic trickier to debug. Lets
stop using meta dependencies and instead always use top-level
dependencies.

Fixes #2728

## Testing

Does it pass CI and not contain any other nits? 👍 

## Deployment

1. Upgrading existing production instances. -- nope
2. New installs. -- nope

### DIFF
--- a/install_files/ansible-base/roles/grsecurity/meta/main.yml
+++ b/install_files/ansible-base/roles/grsecurity/meta/main.yml
@@ -1,5 +1,0 @@
----
-# Add dependency for FPF repo installation role,
-# since the securedrop-grsec package is hosted there.
-dependencies:
-  - install-fpf-repo
--- a/install_files/ansible-base/securedrop-prod.yml
+++ b/install_files/ansible-base/securedrop-prod.yml
@@ -19,6 +19,7 @@
   roles:
     - { role: common, tags: common }
     - { role: tor-hidden-services, tags: tor }
+    - { role: install-fpf-repo, tags: fpf }
     - { role: grsecurity, when: grsecurity, tags: [grsec, grsecurity] }
   become: yes
 


dachary:
> Ready for review / Work in progress

@msheiny not merging because I'm not sure if you want to check something more. Feel free to merge though. Happy to get rid of the only `meta` left in the ansible tree ;-)

msheiny:
Actually...I take back my original comments ... we should test this in prod... because:
* the staging environment doesnt install grsecurity  ( we have an agenda to fix that, let me make an official ticket though)
* and staging uses a separate playbook which happens to include the fpf-role separately (we should also make a ticket to utilize the same exact playbook in prod and in staging)

dachary:
I'm confused: when I do `vagrant up /staging/` I get
<pre>
vagrant@app-staging:~$ uname -a
Linux app-staging 3.14.79-grsec #1 SMP Wed Jun 21 19:57:53 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
</pre>
CircleCI does a different staging ?

msheiny:
Update: 

We already have a ticket for production install testing in CI --> #1681 
Grsec support in CI staging is now tracking under --> https://github.com/freedomofpress/ansible-role-grsecurity-build/issues/11

>CircleCI does a different staging ?

Yeah CircleCI does a number of tweaks. The relevant piece is that it skips tags for `grsec` completely. This is because the currently compiled grsec kernel doesnt have the modules for Xen that are needed to run in AWS (the system will just fall over).

dachary:
If I do `vagrant provision app-staging` on my laptop using your branch, it will however install grsec and therefore verify there is no regression, right ?

conorsch:
Thanks right, @dachary. It's only the _CI_ staging that skips grsec installs—local VMs will still freceive grsec-patched kernels.

msheiny:
> I stand by my approval then :-)

The @dachary stamp of approval!!! 🎖  I DID IT!!!! Woooooooooooo!

PR #2741, opened 2017-12-13
Push grsecurity tasks to top of the install

### LINKED ISSUE #2737: Ansible install -- Place grsecurity logic before other tasks
# Feature request

## Description

The grsecurity kernel we compile can be problematic on certain hardware and typically an admin won't discover this until about half-way through the install process. Let's move the grsecurity higher up the install chain to discover issues earlier in the install.

## User Stories

As a securedrop admin, I want to know ASAP if the FPF kernel is going to 💥 on my hardware.

msheiny:
## Status

Ready for review

## Description of Changes

Move grsec logic to the top of the play. Reasoning behind move is:
* we want grsec on the box as fast as possible
* we want a fail fast experience so if grsec kernel bombs out on a particular piece of hardware we arent also locking out the box and installing un-necessary networking

Fixes #2737 

## Testing

Test a local prod run ... make sure CI is passing.

## Deployment

1. Upgrading existing production instances.  -- Nope
2. New installs. -- Nope

### DIFF
--- a/install_files/ansible-base/securedrop-prod.yml
+++ b/install_files/ansible-base/securedrop-prod.yml
@@ -17,10 +17,10 @@
 - name: Add FPF apt repository and install base packages.
   hosts: securedrop
   roles:
-    - { role: common, tags: common }
-    - { role: tor-hidden-services, tags: tor }
     - { role: install-fpf-repo, tags: fpf }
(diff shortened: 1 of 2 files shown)


conorsch:
Super straightforward. Will run through the paces in VMs to be absolutely sure.

Issue #2742, opened 2017-12-13
Swap ansible  command escalations to use become instead


msheiny:
# Feature request

## Description

We are using some really bad pragma inside ansible tasks such as this:

```
  command: >
    su -s /bin/bash -c 'gpg
    --homedir {{ securedrop_data }}/keys
    --import {{ securedrop_data }}/{{ securedrop_app_gpg_public_key }}' {{ securedrop_user }}
```

where we should be utilizing the ansible [become](http://docs.ansible.com/ansible/latest/become.html) feature whenever possible. 

## User Stories

As a securedrop developer ... something something ansible best practices.

conorsch:
@msheiny If we port the staging environment to Molecule (in Vagrant, just like it is in AWS for CI), we'd get `ansible-lint` checks for free to report on these.

eloquence:
We should still address this, but also consider moving this kind of logic into the packages (e.g., postinst) wherever possible; see for example #2641.

PR #2748, opened 2017-12-14
Dynamically opens firewall during OSSEC agent registration

### LINKED ISSUE #2478: Re-run of playbook against existing server fails on OSSEC registration
# Bug

## Description

Upon re-run of the install playbook against an existing production instance, I found a really weird condition where a few things line up to cause the ossec agent registration tasks to break. ossec fails to detect a registered client AND then bombs out on attempt to register. This is against an existing install that's already been rebooted (and iptables clamped down).

## Steps to Reproduce

Attempt to run `./securedrop-admin install` against an existing installation where the IP is listed twice under the ossec-agent list tool. I'm not certain of when this scenario would have occurred but I've found it on mine so I'm guessing it may be present somewhere else.

## Expected Behavior

Playbook runs successfully.

## Actual Behavior

Playbook bombs out at task `Register OSSEC agent.` under `install_files/ansible-base/roles/ossec-agent/tasks/agent_config.yml` . I confirmed on the app server that there is no iptables hole punched for port 1515 outbound.

## Comments

* https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/ossec-server/tasks/authd.yml#L23 -- detection of existing agent registered can be incorrect. For instance, in my prod, my agent is listed registered as - `{{ app_ip }}-{{ app_ip }} is available.` (vs the expected `{{ app_hostname }}-{{ app_ip }} is available.`)
* okay, that scenario should be fine and ansible attempts to re-register BUT the firewall rule needed (`1515`) will not open right away b/c the logic in https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/ossec-agent/tasks/agent_config.yml#L17 is flawed. The opening up of the port will not take affect until the handlers are flushed. Which is not done until the role runs ... but it'll fail before that. The entire iptables logic is problematic here and I'm assuming it works on first install because there is no iptables restrictions on in a default ubuntu installation.

conorsch:
## Status

Ready for review.

## Description of Changes

Fixes #2478.

* Combines `ossec-agent` and `ossec-server` roles into `ossec`.
* Moves Postfix config logic into separate role `postfix`.
* Dynamically opens and closes firewall ports during agent registration
* Minor: implements a single conversion as described in #2742

While the diff looks intimidating, it's mostly just mv operations: all the duplicated logic across both OSSEC rules has now been simplified. If you're curious about the motivation behind the implementation, the commit messages make great evening reading.

The two commits with titles matching `^Creates` should probably be squashed into one, to tell a better story to future maintainers. I kept them as two to aid in review. Chime in if you have have a preference either way.

## Testing

1. Install on prod VMs and confirm receipt of OSSEC alerts.
2. Rerun prod playbook with no config changes and confirm it works.
3. Deauthorize app as agent [as described here](https://github.com/freedomofpress/securedrop/issues/2478#issuecomment-351213581), then rerun playbook and confirm no errors.
4. Rerun prod playbook _again_ after 3 and _again_ confirm no errors.

## Deployment

Should smooth out the re-run story for Admins. The goal is to make use of the `./securedrop-admin install` flow more reliable. Hopefully this does that! We'll need a lot of QA to gather enough data to confirm the issue is knocked down flat. 

## Checklist

I have not performed QA in prod VMs yet. Since this is a substantial change, several maintainers (other than me) should confirm good functionality. 

I will note I'm getting a single test failure on `test_ossec_connectivity`, due to the `@pytest.mark.xfail(strict=True)` marker, which throws an error **because the test is actually passing.** We'll see what CI has to say about that.


emkll:
Tried this with staging VMs and the problem persists :(. 
This is how I tested:
1. `git checkout 0.5`
2. `make build-debs`
3. `vagrant up /staging/`
4. de-register app-staging agent in ossec manage_agents on mon-staging
5. `git checkout ossec-combined-role` 
6. `make build-debs`
7. `vagrant provision /staging/`

I do see the changes to the firewall by the Ansible run:
```
changed: [mon-staging] => (item={u'chain': u'INPUT', u'proto': u'tcp', u'cstate': u'NEW,ESTABLISHED,RELATED', u'jump': u'ACCEPT', u'dest': u'app-staging', u'match': u'state', u'dest_port': 1515}) => {"chain": "INPUT", "changed": true, "failed": false, "flush": false, "ip_version": "ipv4", "item": {"chain": "INPUT", "cstate": "NEW,ESTABLISHED,RELATED", "dest": "app-staging", "dest_port": 1515, "jump": "ACCEPT", "match": "state", "proto": "tcp"}, "rule": "-p tcp -d app-staging -m state -j ACCEPT --destination-port 1515 --state NEW,ESTABLISHED,RELATED", "state": "present", "table": "filter"}
[...]
changed: [mon-staging] => (item={u'chain': u'OUTPUT', u'proto': u'tcp', u'cstate': u'ESTABLISHED,RELATED', u'jump': u'ACCEPT', u'source': u'app-staging', u'source_port': 1515, u'match': u'state'}) => {"chain": "OUTPUT", "changed": true, "failed": false, "flush": false, "ip_version": "ipv4", "item": {"chain": "OUTPUT", "cstate": "ESTABLISHED,RELATED", "jump": "ACCEPT", "match": "state", "proto": "tcp", "source": "app-staging", "source_port": 1515}, "rule": "-p tcp -s app-staging -m state -j ACCEPT --source-port 1515 --state ESTABLISHED,RELATED", "state": "present", "table": "filter"}
```

install_files/ansible-base/roles/ossec/tasks/register.yml
```diff
@@ -0,0 +1,99 @@
...
+  changed_when: false
+
+# Gnarly vars retrieval logic in this task. The "register" action above applies
+# only to the Monitor Server, so the Application Server won't be able to access
+# the value (and the play will fail as a result). So on both hosts, let's look up
+# the registered value by referencing the hostvars for the Monitor Server, then
+# copy the result to a more conveniently named fact on both hosts.
+- name: Set host fact for OSSEC registration state.
+  set_fact:
+    ossec_agent_already_registered: "{{ hostvars[groups.securedrop_monitor_server.0].ossec_list_agents_result.stdout == app_hostname +'-'+app_ip+' is available.' }}"
```
msheiny: so i ran into a similar issue that i had before ... `hostvars[groups.securedrop_monitor_server.0]` doesnt populate at all for me. Running under tails `3.3`, ansible `2.3.2.0`, i added a debug right before the `set_fact` here ...

`debug: var="{{ hostvars[groups.securedrop_monitor_server.0] }}"

response is `VARIABLE IS NOT DEFINED` on both servers and then  `start authd` is always firing for me. I had to fix a commit for the iptables rules --- and now its working... but its always firing. Note that I am running this with explicit virtualenv, and the under the ansible dir:

`./securedrop-prod.xml -t ossec --diff --ask-become-pass`

emkll:
Tested @msheiny's commit and it works! I can confirm that it correctly re-registers the the app ossec agent:
```
root@mon-staging:/home/vagrant# /var/ossec/bin/list_agents -a
app-staging-10.0.1.2 is available.
```

conorsch:
Digging into the authd-always-firing issue that @msheiny pointed out...

conorsch:
Uh-oh, looks like there's an install-time bug now, introduced by the reordering changes:

```
TASK [postfix : Create procmail log file.] *************************************
fatal: [mon]: FAILED! => {"changed": false, "checksum": "da39a3ee5e6b4b0d3255bfef95601890afd80709", "failed": true, "gid": 0, "group": "root", "mode": "0644", "msg": "chown failed: failed to look up user ossec", "owner": "root", "path": "/var/log/procmail.log", "size": 0, "state": "file", "uid": 0}
```

Will try to reproduce and patch as part of addressing @msheiny's concerns.

conorsch:
> looks like there's an install-time bug now [...] Will try to reproduce and patch 

Yes, was able to reproduce in prod VMs. The issue does _not_ occur in staging VMs, due to the local deb package install logic landing prior to postfix configuration. The `ossec` user is created via postinst scripts in the ossec deb package, so this makes sense. To resolve in the prod config, I simply moved the postfix configuration logic after the OSSEc configuration logic. Updated the staging playbook to match.

> Can you verify the behaviour I'm seeing with the start authd always firing?

As for the problem @msheiny describes, I am unable to reproduce. I can confirm that authd does _not_ fire every time—see this example of a rerun against already configured prod VMs:

```
TASK [ossec : Check whether Application Server is registered as OSSEC agent.] **
********************************
ok: [mon] => {"changed": false, "cmd": ["/var/ossec/bin/list_agents", "-a"], "de
lta": "0:00:00.007961", "end": "2018-01-09 01:23:56.939846", "rc": 0, "start": "
2018-01-09 01:23:56.931885", "stderr": "", "stderr_lines": [], "stdout": "app-pr
od-10.0.1.4 is available.", "stdout_lines": ["app-prod-10.0.1.4 is available."]}

TASK [ossec : Set host fact for OSSEC registration state.] *********************
********************************
ok: [app] => {"ansible_facts": {"ossec_agent_already_registered": true}, "change
d": false}
ok: [mon] => {"ansible_facts": {"ossec_agent_already_registered": true}, "change
d": false}

TASK [ossec : Start authd.] ****************************************************
********************************

TASK [ossec : Add firewall exemption for OSSEC agent registration (both servers)
] ******************************

TASK [ossec : Register OSSEC agent.] *******************************************
********************************
```

Full terminal output here: https://gist.github.com/conorsch/fdb1cd4a5936186aca063496c90341fa. The command used to generate the above output was `ansible-playbook securedrop-prod.yml --tags postfix,ossec -v --diff`, with virtualenv manually activated in a Tails VM.

So I'm happy with these changes as they stand. Can I get a re-review, @msheiny and @emkll? (@kushaldas if you've succeeded in setting up your system to handle QA on SD, feedback welcome!)

kushaldas:
Tried to do a fresh install.

```
TASK [ossec : Set host fact for OSSEC registration state.] *****************************************************
fatal: [app]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'dict object' has no attribute 'ossec_list_agents_result'\n\nThe error appears to have been in '/home/amnesia/Persistent/securedrop/install_files/ansible-base/roles/ossec/tasks/register.yml': line 14, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n# copy the result to a more conveniently named fact on both hosts.\n- name: Set host fact for OSSEC registration state.\n  ^ here\n"}
```

conorsch:
Thank you for the feedback, @kushaldas. Given your results paired with @msheiny, looks like this needs more work. I'm going to step away from this PR for now to focus on 0.5.1 QA. Let's leave it open, and we'll try to get it in for 0.5.2.

emkll:
I can confirm that the upgrade path still works, I've testing against staging and prod vms following these steps:
1. `git checkout 0.5`
2. `make build-debs`
3. `vagrant up /staging/`
4. de-register app-staging agent in `ossec manage_agents` on `mon-staging`
5. `git checkout ossec-combined-role`
6. `make build-debs`
7. `vagrant provision /staging/`

redshiftzero:
I ran through this in prod VMs and like @msheiny also saw `authd` always firing. @kushaldas I suspect the error you experienced was due to `mon` falling over earlier in the install (cc #2885)

kushaldas:
I will do another round of tests on fresh vms, and report back tomorrow.

msheiny:
Testing now against my prod

msheiny:
yeppp @emkll i'm seeing that same error message now

conorsch:
Able to reproduce the "authd always firing" behavior mentioned above. Looks like manually removing an agent somehow causes it never to be readded. On subsequent repeated runs, the config logic _always_ determines that the Application Server is not registered—which is true:

```
vagrant@mon-prod:~$ sudo /var/ossec/bin/list_agents -a
** No agent available
```

So we're screwing up somewhere with the auth flow. Switching to manual commands for the OSSEC registration to debug.

emkll:
Testing the latest commit 97e1b1a and it appears to have worked on my physical instance:
```
./list_agents -a
app-10.20.2.2 is available.
```
I also now correctly receive email alerts for the app server

conorsch:
@msheiny's latest changes have resolved the problem for me, as well. Tested many times in staging and even after manually removing `app-staging` as an agent (again, as described in https://github.com/freedomofpress/securedrop/issues/2478#issuecomment-351213581), the new logic correctly reauths the agent. Additionally, the new logic is idempotent, and shows no changes on the OSSEC config if the agent is already present.

_However_, we now have a failure on OSSEC connectivity in CI. Looks like it's caused by a pessimistic strict xfail on the OSSEC connectivity check, so will investigate the behavior via local testinfra runs and patch onto this PR if the test now passes (as it should).

conorsch:
Confirmed that removing the strict xfail on the OSSEC config check caused test suite to pass locally. The same test failed twice consecutively in CI, so pushed up. Waiting on CI to pass. My tests-only change automatically dismissed approvals from both @msheiny and @emkll, so as long as CI passes here, we should be good to go—please re-review!

conorsch:
CI is passing! :tada: :metal:

PR #2753, opened 2017-12-16
Set Referrer-Policy in vhost config + landing page guidelines

ageis:
## Status

Ready for review.

## Description of Changes

Fixes: #2762

[Referrer-Policy](https://www.w3.org/TR/referrer-policy/) is a new HTTP security header which was recommended and specified by W3C at the beginning of 2017. This setting allows us to instruct browsers to never send a 'Referer' header for all requests originating from the site. It would be extremely rare or unheard of for any off-site links to be present on a SecureDrop page, but it's worth including this in the headers anyhow.

Adds another security header to Apache2 virtual host configuration files, landing page deployment recommendations, and tests.

## Testing

n/a

## Deployment

No special considerations. If merged in then the change will propagate with new installs.

## Checklist

n/a, haven't familiarized myself with or tried running your tests locally but I will next time

### DIFF
--- a/docs/deployment/landing_page.rst
+++ b/docs/deployment/landing_page.rst
@@ -130,6 +130,7 @@ If you use Apache, you can use these:
     Header set X-Download-Options: noopen
     Header set X-Permitted-Cross-Domain-Policies: master-only
     Header set Content-Security-Policy: "default-src 'self'"
+    Header set Referrer-Policy "no-referrer"
 
 If you intend to run nginx as your webserver instead, this will work:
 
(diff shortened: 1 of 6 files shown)


dachary:
@ageis I'll wait a few days before merging, in case someone has a different opinion about this.

conorsch:
Looks fine, won't affect running installs. 

> The tor browser design document suggests web sites should be using the Referrer-Policy header.

Great resource @dachary, and precisely the kind of information that should be included with changes of this nature, preferably in an issue @ageis, so that it's clear not just what is being changed but _why_ it's being changed.

dachary:
> that should be included with changes of this nature, preferably in an issue

Created #2762 to match this PR, good call 👍

Issue #344, opened 2014-04-01
Admin interface


micahflee:
It would be useful if there were an admin web interface, in addition to source and document interfaces, that let organizations customize their SecureDrop instance.

It could let them:
- Set an organization name (so the HTML title isn't always SecureDrop, but rather something like "The Intercept - SecureDrop")
- Upload a logo/header image
- Add custom submission instructions/other blocks of text
- Add a custom footer

This would also be a good place to let admins manage journalist users as well:
- Add new users (creates new authenticated hidden services and Google Auth tokens), and also set their public key for #98 
- Regenerate credentials for a user
- Revoke user privileges
- Maybe import/export all user data, to facilitate a fresh install

This could simplify the installation process too. Rather than specifying journalist users ahead of time during installation, it could just give the admin access to this interface.

redshiftzero:
Of the features here, most of them have now been implemented, and we have filed more specific followups for the issues that are still to be completed:

> Set an organization name (so the HTML title isn't always SecureDrop, but rather something like "The Intercept - SecureDrop")

Tracked in #1575

> Upload a logo/header image

Added in #2769

> Add custom submission instructions/other blocks of text

Tracked in #3044

> Add a custom footer

Tracked in #1967

> Maybe import/export all user data, to facilitate a fresh install

Tracked in #606 

All of the admin interface features regarding adding/removing new users and regenerating credentials was added in #522.

In terms of the OSSEC data, since the application server cannot modify contents of the monitor server, we cannot easily add these features to the Admin Interface. If there is a specific suggestion for doing so, it should be suggested in a followup issue.

Closing in favor of these scoped issues.

PR #2769, opened 2017-12-29
Enable admin to update logo via web app

redshiftzero:
## Status

Ready for review

## Description of Changes

Implements item from #344, enabling administrators to upload their custom logo/image via the web application (the current command line only approach to update the logo is a common source of confusion for admins).

If this generally looks fine, I can add some docs and screenshot functional tests onto this PR

## Testing

Log in to journalist interface and update the image

## Deployment

We should consider deprecating the command line path for updating the logo and have admins use the web app instead. The reason for this is that they otherwise will need to keep the current logo they want to use current in their admin workstation for subsequent runs of `securedrop-admin install`

admin-logo-update


dachary:
great idea ! 👍

dachary:
> We should consider deprecating the command line path for updating the logo

This sounds right. Do you plan on doin that in this pull request ?

dachary:
Known transient error https://travis-ci.org/freedomofpress/securedrop/builds/322986451?utm_source=github_status&utm_medium=notification

<pre>
...", "Selecting previously unselected package python3-six.", "Preparing to unpack .../python3-six_1.5.2-1ubuntu1_all.deb ...", "Unpacking python3-six (1.5.2-1ubuntu1) ...", "Selecting previously unselected package python3-debian.", "Prepclose failed in file object destructor:
sys.excepthook is missing
lost sys.stderr
The command "ansible-playbook -i inventory -vv --connection=local install_files/ansible-base/securedrop-development.yml
  " failed and exited with 1 during .
Your build has been stopped.
</pre>

securedrop/journalist_app/__init__.py
```diff
@@ -52,6 +52,16 @@ def handle_csrf_error(e):
         template_filters.rel_datetime_format
     app.jinja_env.filters['filesizeformat'] = template_filters.filesizeformat
 
+    @app.template_filter('autoversion')
+    def autoversion_filter(filename):
+        """Use this template filter for cache busting"""
+        try:
+            timestamp = str(path.getmtime(filename[1:]))
+        except OSError:
```
dachary: what about **if os.path.exists** instead of try / except ?

securedrop/tests/test_journalist.py
```diff
@@ -678,6 +678,52 @@ def test_admin_add_user_integrity_error(self,
...
 
+    def test_logo_upload_with_valid_image_succeeds(self):
+        # Save original logo to restore after test run
+        logo_image_location = os.path.join(config.SECUREDROP_ROOT,
+                                           "static/i/logo.png")
+        with open(logo_image_location) as logo_file:
+            original_image = logo_file.read()
+
+        self._login_admin()
+
```
dachary: this should be in a **try finally** so the image is restored even if the test fail

redshiftzero:
Thanks for fast review @dachary! Changes made and repushed 🎄

dachary:
I tried it manually and it works beautifully 💯 I'll merge this week-end if nobody objects.

Two thoughts:

* It would be helpful to add more information about the image cache on the config page. Although the image is updated immediately in the form, the logo displayed elsewhere is still cached and shows the old one. Something like: "Hit Control-F5 to clear the cache and see the new logo everywhere" maybe ? An alternative would be to **|autoversion** all logo.png but that would effectively always invalidate the cache, always, for everyone. Another idea would be to have a more sophisticated **autoversion**, used on every logo.png that adds the ?XXXX everywhere but only during X hours after the logo.png is modified.

* I tried to think about ways this new feature could be misused by a malicious journalist who has admin privileges, but I could not imagine anything. Maybe @emkll will have a different opinion ?

PR #2772, opened 2017-12-30
Fix async genkey logging

### LINKED ISSUE #2609: async_genkey: thread safety issue in async_genkey by calling current_app in separate thread.
# Bug
async_genkey calls current_app in some conditions and can raise `RuntimeError: Working outside of application context.`

## Description

async_genkey calls current_app in some conditions and can raise `RuntimeError: Working outside of application context.` Example: https://gist.github.com/tmc/095d490608c7cc5e2aa5f4825cd7a4a3#file-gistfile1-txt-L1122

## Steps to Reproduce

This is a flake I have running tests inside of a container in a qubes 3.2 machine.

## Expected Behavior
To not break in case of an underlying error.

## Actual Behavior
Breaks when trying to call current_app.

## Comments
There's clearly another underlying issue here but it shouldn't blow up in this manner.

tmc:
## Status

Ready for review.

## Description of Changes

This skips calling current_app which is invalid to do from a separate thread (which async_genkey runs in).

Fixes #2609.

## Testing
CI

## Deployment
N/A

## Checklist
N/A

### DIFF
--- a/securedrop/source_app/utils.py
+++ b/securedrop/source_app/utils.py
@@ -1,3 +1,4 @@
+import logging
 import subprocess
 
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


tmc:
This failure appears to be a flake, can someone re-run?

PR #2778, opened 2017-12-31
Split base.sass into multiple file imports

### LINKED ISSUE #2777: Split base.sass into multiple files
# Split base.sass into multiple files

## Description

Suggested refactoring. Currently base.sass is one very large file and it is hard to see what parts of the CSS are responsible for which design components of the interface. Giving each of these parts a file with a clear name should make maintenance easier because the responsibility / encapsulation boundaries of each part will be clearer. This will help move closer to proper design modules that could be listed in a style guide.

toast:
## Status

Ready for review

## Description of Changes

Refactoring to split base.sass into multiple file imports.

Fixes #2777 

    - Folders created for Global (order dependent) and Modules (aim to make these BEM syntax)
    - Each component in the interface now has a named module file
    - Added new sass folders to rsync-filter

## Testing

Have a quick look through all the pages in the app for style discrepancies. Theoretically a refactoring only but has potential to affect all styles.

### DIFF
--- a/securedrop/sass/_base.sass
+++ b/securedrop/sass/_base.sass
@@ -65,6 +65,20 @@
 @import "modules/select-container"
 // Document actions button - Seems to be unused. Delete?
 @import "modules/document-actions-button"
+// Grid item - part of the grid. Doesnt seem to be used very much though.
+@import "modules/grid-item"
+// Option - might be unused. Delete?
+@import "modules/option"
+// Logo - Small and large versions of the branding; can also be the journalist org brand (not SD).
+@import "modules/logo"
+// Submissions - container for all of the submissions
+@import "modules/submissions"
+// Submission - an individual submission
+@import "modules/submission"
+// List plain and starred - two types of lists
+@import "modules/list-plain-and-starred"
+// Logout - the logout button (top right)
+@import "modules/logout"
 
 =base_rules
   // LIBRARIES
@@ -104,78 +118,13 @@
   +banner-warning
   +select-container
   +document-actions-button
-
-  .grid-item
-    display: inline-block
-    letter-spacing: normal
-
-  .option
-    display: inline-block
-    width: 30%
-    min-width: 300px
-    min-height: 310px
-    margin: 1%
-    padding: 25px
-    border: 1px solid #c3c3c3
-    vertical-align: top
-    position: relative
-
-    &.alt
-      background-color: $color_purple_medium
-      color: white
-
-    &.selected
-      width: 500px
-      height: auto
-
-  .logo
-    width: 250px
-    margin-right: 9px
-    margin-top: 36px
-    margin-left: 9px
-
-    &.small
-      width: 150px
-      margin-top: 20px
-
-  .submissions
-    padding-left: 0px
-
-    &:dir(rtl)
-      padding-right: 0px
-
-  .submission
-    background: $color_grey_xlight
-    border-bottom: 1px solid #ddd
-    padding: 10px
-    text-align: left
-
-    &:dir(rtl)
-      text-align: right
-
-    &:last-child
-      border-bottom: none
-
-    .file
-      min-width: 300px
-      display: inline-block
-
-      &.reply
-        font-size: small
-
-    .info
-      font-size: 12px
-      color: #666666
-
-  ul.plain
-    list-style: none
-
-  ul.starred
-    margin-bottom: 1.33em
-
-  div#logout
-    float: right
-    margin: 1em 1em 0 1em
+  +grid-item
+  +option
+  +logo
+  +submissions
+  +submission
+  +list-plain-and-starred
+  +logout
 
   label
     cursor: pointer
--- a/securedrop/sass/_base.sass
+++ b/securedrop/sass/_base.sass
@@ -15,6 +15,16 @@
 @import modules/link
 // Headline - The typography hierarchy
 @import modules/headline
+// Main content area - Outer container area for the whole page (see also: Panel)
+@import modules/main-content-area
+// Codename and password - common style for secrets
+@import modules/codename-and-password
+// Document actions - seems to be unused. Delete?
+@import modules/document-actions
+// Doc check - Checkbox spacing. In journalist interface only?
+@import modules/doc-check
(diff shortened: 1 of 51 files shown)


dachary:
@toast there is a legitimate error here https://circleci.com/gh/freedomofpress/securedrop/5548?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

<pre>
Step 21/30 : RUN sass --force --stop-on-error --update sass:static/css --style compressed
 ---> Running in 7d81f642c9f4
DEPRECATION WARNING:
Sass 3.5 will no longer support Ruby 1.9.3.
Please upgrade to Ruby 2.0.0 or greater as soon as possible.

  directory static/css
Error: Invalid US-ASCII character "\xE2"
        on line 72 of sass/_base.sass
        from line 1 of sass/journalist.sass
  Use --trace for backtrace.
</pre>

dachary:
@toast I love this cleanup :-) For the record, in order to make sure nothing was removed or changed by accident, we need to run `sass` and compare the output in `static/css` to verify they are identical.

dachary:
@toast thanks for fixing the error :-) I tried to compare the sass output with and without this change to verify there is no unintended modification. Unfortunately the output is very different (~2000 lines of diff). Do you maybe have a suggestion about how to do that ? That is making sure the changes do not introduce a regression in the display.

dachary:
@toast I created screenshots with and without your change and they look good which is encouraging.

https://github.com/freedomofpress/securedrop/commit/b059ceb9b7567745dfa8fe30c4d32e3df93328d4

* git checkout -b wip-dachary-compare origin/develop
* pytest -v --page-layout tests/pages-layout
* git add tests/pages-layout/screenshots/en_US
* git commit -m 'orig'
* git reset --hard toast:develop
* pytest -v --page-layout tests/pages-layout
* git add tests/pages-layout/screenshots/en_US
* git commit -m 'new'
* git push wip-dachary-compare
* firefox visit-the-new-commit # to see the screenshots side by side

toast:
Sorry @dachary I didn't realise splitting base in one go like this would make the diff so awkward. 

In this gist I've inlined all of the new mixin calls (but not the old ones) back into _base.sass:
https://gist.github.com/toast/b5479cfeaa4b9d82851076688176e102

A diff between that and the _base.sass from 196c6562cc9c5b110ae4e197aa7b5c2fa8f912aa (before the changes) shows that most of the mixin changes are moving things up or down in the file or renaming:

* Renaming +font_rules to +font-awesome-iconography
* Renaming +button_rules to +button and moving it further down to sit with the other modules
* Moving `a` rules and `.select` rules further up
* Splitting `.aside, .flash, .banner-warning` into three separate parts (instead of combining them in one rule), which means that I've needed to make copies of the encapsulated rules under that combined selector to put them into each of those 3 files for `.aside`, `.flash`, and `.banner-warning` separated modules.

Happy to rework this or do other suggestions you might have to make this less awkward. Good to know as I was about to go do the same with source_index.sass / journalist.sass, I'll do those in smaller pieces.

dachary:
Thanks for proposing to rework, it would be really. For review of course but also in case we find a bug later one for forensics. I used the gist your provided and compared with develop but it still is difficult to separate what is just splitting existing content into separate files, moving content for clarity and improvements.

For instance I ran into the following hunk:

before

and because you told me about moving blocks so they are grouped when it is meaningfull, I was able to track it down to:

after

Now it is still unclear to me why this blocked needed to be moved and if it was in a separate commit with a comment explaining why it is beneficial, it would be most helpful.

Another approach that would make things simpler would be to have a commit that just splits content so that running SASS produces exactly (or trivial to compare) the same output as before for sake of comparing. You would then have less to explain and I fully trust you to create a hierarchy that makes sense.

What do you think ?

toast:
No problem. I'll have a go at doing separate commits for the different parts.

dachary:
> I'll have a go at doing separate commits for the different parts.

This is much appreciated, thank you :-)

securedrop/sass/global/_body-tag.sass
```diff
@@ -0,0 +1,6 @@
+=body-tag
+  body
+    font-family: "HelveticaNeue-Light", "Helvetica Neue Light", "Helvetica Neue", Helvetica, Arial, "Lucida Grande", sans-serif
+    font-weight: 300
+    padding: 0
```
dachary: I noticed this is not strictly copy/paste but 0px changed to 0 but ... whatever the unit zero still is zero ;-)

securedrop/sass/_base.sass
```diff
@@ -89,69 +95,9 @@
...
-
-  .banner-warning
-    border: 1px solid $color_error_border
-    background-color: $color_error_background
-
-    i
-      color: #D62727
+  +aside
+  +flash
+  +banner-warning
```
dachary: :+1: the refactored aside, flash and banner-warning were carefully verified to exactly contain the concatenation of the corresponding combined selectors. While combined selectors are a terse way to express the same (i.e. more DRY), the motivation behind this refactor is most probably maintainability. When someone fixes or improves **flash** it is easier for them to not be forced to think about how it will influence **banner-warning** or **aside**. There is no reason why those would always share part of their CSS.

securedrop/sass/_base.sass
```diff
@@ -98,14 +102,8 @@
...
-
-  div
-    &#select-container
-      display: inline
-
-    &.document-actions button
-      font-size: 14px
-      padding: 6px 15px
+  +select-container
+  +document-actions-button
```
dachary: These are not an exact copy of the original. They were both under the same **div** but this factorization did not make sense from a functional point of view and there is no harm in breaking it.

dachary:
Note: I made a number of comments while reviewing, just as I was verifying everything. GitHub trashed everything for some reason. Oh well.

securedrop/sass/global/_body-tag.sass
toast: Sorry, unconscious habit of mine to strip the unit off in CSS whenever it's a 0 (because as you say, the unit is redundant).

toast:
No worries, thanks for your advice getting the PR ready.

PR #2584, opened 2017-11-17
Add codecov

tmc:
## Status

Ready for review

## Description of Changes

Changes proposed in this pull request:

Adds codecov support.

Sample report: https://codecov.io/gh/freedomofpress/securedrop/branch/add-codecov

## Testing

N/A

## Deployment

N/A

PR #2792, opened 2018-01-03
Remove mentions of coveralls, update requirements, update badge

### LINKED ISSUE #2261: Investigate and fix slow coveralls checks or move to codecov
The coverage checks in some cases are taking a very long time or not returning. We need to investigate what is going on and fix it or migrate to some other coverage tool.

redshiftzero:
## Status

Ready for review

## Description of Changes

Closes #2261. 

 * Codecov reports are working thanks to #2584: https://codecov.io/gh/freedomofpress/securedrop
 * This PR just updates requirements, docs, and the badge in the README to remove coveralls references

## Testing

Reviewing diff should be sufficient

## Deployment

None, dev/test environment only

### DIFF
--- a/securedrop/requirements/test-requirements.txt
+++ b/securedrop/requirements/test-requirements.txt
@@ -7,17 +7,12 @@
 attrs==17.4.0             # via pytest
 beautifulsoup4==4.6.0
 blinker==1.4
-certifi==2017.4.17        # via requests
-chardet==3.0.4            # via requests
 click==6.7                # via flask, pip-tools
-coverage==4.4.1           # via coveralls, pytest-cov
-coveralls==1.1
-docopt==0.6.2             # via coveralls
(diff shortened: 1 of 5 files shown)


securedrop/requirements/test-requirements.txt
```diff
@@ -27,9 +22,7 @@ pip-tools==1.9.0
 pluggy==0.6.0             # via pytest
 py==1.5.2
 pytest-cov==2.5.1
-pytest==3.3.1
```
dachary: Regression from 3.3.1 to 3.1.1 is probably not intended ? It will cause trouble with the caplog API for tests

redshiftzero: Ahh.. good call. Fixing

redshiftzero: Updated :)

PR #2797, opened 2018-01-05
circleci: do not cache docker layers

dachary:
## Status

Ready for review

## Description of Changes

When a RUN apt-get update layer is cached it may become obsolete. This
will manifest with apt-get install failing on 404 because the cache
references packages that no longer exist.

The developer who sees this error is unlikely to link that to the
fact that docker layers are cached accross runs. And it will not occur
to them to change the line of the Dockerfile to force an update.

The benefit of maintaining a cache is to speed up tests (skipping
apt-get install and pip install on every run). But this benefit is
less appealing when developer are confused by the errors it may
induce.

## Testing

N/A

## Deployment

N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,38 +25,10 @@ jobs:
       - checkout
       - setup_remote_docker
 
-      - restore_cache:
-          keys:
-            - v1-sd-layers-{{ .Branch }}-
-            - v1-sd-layers-
-          paths:
-            - /caches/layers.tar.gz
-
-      - run:
-          name: Load image layer cache
-          command: |
-            set +o pipefail
-            docker load -i /caches/layers.tar |true
-
       - run:
           name: Build Docker images
           command: |
-            set +o pipefail
-            docker images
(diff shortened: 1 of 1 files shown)


tmc:
I'm a bit on the fence regarding the tradeoff here. Ideally I'd prefer to leave the caching in and have a bot automatically open an issue/pr when the cache is inappropriately stale.

dachary:
>  Ideally I'd prefer to leave the caching in and have a bot automatically open an issue/pr when the cache is inappropriately stale.

I'm all for that but ... how do you figure out the cache is stale ?

heartsucker:
The problem with caching is in part that it's best practice to make your `apt` commands all one `RUN` joined by `&&`. If we do this, there should be no problems with caching. I'm opening another PR with these changes for us to check. Even if we decide to purge the cache in this PR, the changes I make to the `Dockerfile` should be kept since the are more in line with best practices.

dachary:
Closing in favor of #2822

PR #2802, opened 2018-01-07
journalist: the admin cannot delete itself

### LINKED ISSUE #1562: Prevent admin journalists from deleting themselves

singuliere:
## Status

Ready for review

## Description of Changes

Fixes #1562

## Testing

* cd securedrop
* make dev
* verify the trashbin cannot be clicked for self

## Deployment

N/A

### DIFF
--- a/securedrop/journalist_app/admin.py
+++ b/securedrop/journalist_app/admin.py
@@ -208,7 +208,13 @@ def make_blueprint(config):
     @admin_required
     def delete_user(user_id):
         user = Journalist.query.get(user_id)
-        if user:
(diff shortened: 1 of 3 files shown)


dachary:
This looks great :-) I would merge right away but let's give it a few days so people can voice their concerns.

Note: while testing as instructed, I ran into a problem for which [there is a proposed fix](https://github.com/freedomofpress/securedrop/pull/2804)

PR #2803, opened 2018-01-07
daily email alert to journalists about submissions received in the past 24h

### LINKED ISSUE #1195: Submission email notifications
Checking SecureDrop every day is a huge hassle. Instead, it should tell me about new submissions, but without leaking any information.

Every day at 6am, SecureDrop sends me an encrypted email noting the number of unread submissions and the number of submissions received in the last 24 hours, even if these numbers are zero. If submission categories are used, each category's “owner” is sent the same info about their category.

dachary:
## Status

Ready for Review

## Description of Changes

Fixes #1195 

A cron job runs daily on the app server and updates the

    /var/lib/securedrop/submissions_today.txt

file which contains the number of submissions sent in the past 24h, as
created by the manage.py how_many_submissions_today command.

The OSSEC agent on the app server runs a command daily, displaying
the content of /var/lib/securedrop/submissions_today.txt. The output
of the command is sent to the OSSEC server.

A new rule is defined on the OSSEC server to send a mail to when the
output is received from the OSSEC agent running on the app server.

A new procmail rule is definied on the OSSEC server to catch mails
encrypt mails containing the /var/lib/securedrop/submissions_today.txt
string and send them to the email defined by the
journalist_alert_email ansible variable.

A new set of (optional) ansible variables, similar to
ossec_alert_gpg_public_key, ossec_gpg_fpr, ossec_alert_email are
defined: journalist_alert_gpg_public_key, journalist_gpg_fpr,
journalist_alert_email. They are used to upload a journalist public
key to the OSSEC server and inserted into the send_encrypted_alarm.sh
script which handles mails received by procmail.

The modified send_encrypted_alarm.sh script takes one
argument (journalist or ossec) and dispatches the mail read from
stdin to the corresponding recipient.

Integration tests are implemented to verify the following:

  * manage.py how_many_submissions_today
  * the app OSSEC agent sends a mail to the journalist address
  * cover all branches of send_encrypted_alarm.sh

## Testing

* make build-debs
* vagrant up /staging/
* ./testinfra/test.py staging
<pre>
Running Testinfra suite against 'staging'...
Target roles:
    - testinfra/ossec
==================================================================== test session starts =====================================================================
platform linux2 -- Python 2.7.13, pytest-3.3.1, py-1.5.2, pluggy-0.6.0 -- /home/loic/software/securedrop/virtualenv/bin/python2
cachedir: .cache
rootdir: /home/loic/software/securedrop/securedrop, inifile: setup.cfg
plugins: testinfra-1.7.1, xdist-1.21.0, forked-0.2
collected 10 items                                                                                                                                           

testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_procmail[ansible://app-staging] SKIPPED                                              [ 10%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_send_encrypted_alert[ansible://app-staging] SKIPPED                                  [ 20%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_missing_journalist_alert[ansible://app-staging] SKIPPED                              [ 30%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_ossec_rule_journalist[ansible://app-staging] SKIPPED                                 [ 40%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_journalist_mail_notification[ansible://app-staging] SKIPPED                          [ 50%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_procmail[ansible://mon-staging] PASSED                                               [ 60%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_send_encrypted_alert[ansible://mon-staging] PASSED                                   [ 70%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_missing_journalist_alert[ansible://mon-staging] PASSED                               [ 80%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_ossec_rule_journalist[ansible://mon-staging] PASSED                                  [ 90%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_journalist_mail_notification[ansible://mon-staging] PASSED                           [100%]

====================================================================== warnings summary ======================================================================
None
  Module already imported so cannot be rewritten: testinfra

===================================================== 5 passed, 5 skipped, 1 warnings in 106.35 seconds ======================================================
</pre>

## Comments

See the [forum discussion](https://forum.securedrop.club/t/simplifying-testinfra-tests/382) for the rationale behind the **testinfra: remove some XXX tests duplicating Ansible** commits.

## Deployment

* The journalist emails is not set for existing installations, nothing will happen
* During sdconfig the admin will be prompted for the email of the journalist and the encryption key. If set, the daily notification mails will be sent.


dachary:
@emkll this is the reason why `sdconfig` was refactored :-) This is the implementatoin we [discussed a few weeks ago](https://forum.securedrop.club/t/journalist-mail-notifications/90/10). It is mostly OSSEC related, reason why I invite you for a first review. It is mostly complete but still missing a few bits (documentation, sdconfig etc.).

dachary:
Not sure this error comes from this PR, restarting https://circleci.com/gh/freedomofpress/securedrop/7409?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

<pre>
 RUNNING HANDLER [tor-hidden-services : restart tor (simple)] *******************
    Sunday 28 January 2018  00:00:10 +0000 (0:00:00.071)       0:05:17.882 ********
    changed: [mon-staging]
    fatal: [app-staging]: FAILED! => {"changed": false, "msg": " * Stopping tor daemon...\n   ...done.\n * Starting tor daemon...\n   ...fail!\n"}
</pre>

dachary:
@emkll this is ready for review, no more work in progress :-)

emkll:
@dachary thank you for your patience! I have performed preliminary testing of this feature (on commit ab9dc10c98e646134f514ed15fe9151664bfd21dd, so apologies if the findings below have been addressed, I will test on the new commits late today). This is how I tested it:
* change frequency of  the submissions today`/install_files/securedrop-ossec-agent/var/ossec/etc/ossec.conf`
* make build-debs
* add journalist account info in `site-specific`
* spin up prod VMs
* install will error out because the apt repos contain a version of `securedrop-app-code` incompatible with the current install playbooks
* scp `securedrop-osssec-server` (to mon), `securedrop-ossec-client` and `securedrop-app-code` (to app)
* `sudo apt remove` and `sudo dpkg -i` the packages above
* reboot both servers (though probably not necessary
* submit documents
* ssh to server and run `./manage.py how-many-submissions-today` in `/var/www/securedrop`

Here are some findings:

* https://github.com/freedomofpress/securedrop/issues/2955
* Admins also receive an email informing them of submission status. I am not sure if this is intended, but I think this should not happen. Should an admin want submission information, they should set their email as journalist email or add themselves to a mail group with other journalists. 
```
OSSEC HIDS Notification.
2018 Feb 01 12:54:27

Received From: (app-prod) 10.0.1.4->head -1 /var/lib/securedrop/submissions_today.txt | grep '^[0-9]*$'
Rule: 400600 fired (level 7) -> "Number of submissions received in the past 24h."
Portion of the log(s):

ossec: output: 'head -1 /var/lib/securedrop/submissions_today.txt | grep '^[0-9]*$'':
2


 --END OF NOTIFICATION

```

* I seem to be receiving duplicate emails, for both admin and journalist:
screenshot from 2018-02-01 08-11-54

* I would consider removing 'SecureDrop' from the title line of the emails (but happy to discuss)

I will be doing some further testing today,  with more attention into the size of the emails sent to the journalists.

dachary:
Thanks a lot for the careful testing. The `sdconfig` implementation now in this PR will save you a number of manual steps. Hopefully the generated packages are now compatible and if not this definitely is a bug.

- [ ] Good catch on the duplicate emails. The integration test verify the mails reach the intended recepient but they do not verify no extra mail is sent. The admin is indeed not meant to receive a copy. I've not verified that but I have a hunch about why it happens.
- [ ] I will remove SecureDrop from the title

emkll:
@dachary Just noticed this as well: I have also received a significant number of AppArmor events. Could you check to see if you have those as well? Based on the timestamps, my guess is that this might have something to do with the custom install process *or* something that happens when the `submissions_today.txt` file does not exist:

```
OSSEC HIDS Notification.
2018 Feb 01 01:15:53

Received From: (app-prod) 10.0.1.4->/var/log/syslog
Portion of the log(s):

Feb  1 01:15:53 app-prod kernel: [ 8159.402185] audit: type=1400 audit(1517447753.355:20): apparmor="DENIED" operation="exec" profile="/usr/sbin/apache2" name="/usr/bin/file" pid=3466 comm="apache2" requested_mask="x" denied_mask="x" fsuid=33 ouid=0



 --END OF NOTIFICATION
```

dachary:
- [ ] About the apparmor alerts, I'll double check. 

However `submissions_today.txt` is never accessed via apache2 (nor created by it) therefore I suspect the source of the alerts is something different.

emkll:
@dachary I have tested the latest changes in a staging environment. I can confirm that the `sdconfig` works well. However, I've observed the following test failure when running `./testfinfra/test.py staging` (note that the last test takes a while to fail):
```
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_procmail[ansible://app-staging] SKIPPED                                                                                                                              [  8%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_process_submissions_today[ansible://app-staging] SKIPPED                                                                                                             [ 16%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_send_encrypted_alert[ansible://app-staging] SKIPPED                                                                                                                  [ 25%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_missing_journalist_alert[ansible://app-staging] SKIPPED                                                                                                              [ 33%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_ossec_rule_journalist[ansible://app-staging] SKIPPED                                                                                                                 [ 41%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_journalist_mail_notification[ansible://app-staging] SKIPPED                                                                                                          [ 50%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_procmail[ansible://mon-staging] PASSED                                                                                                                               [ 58%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_process_submissions_today[ansible://mon-staging] PASSED                                                                                                              [ 66%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_send_encrypted_alert[ansible://mon-staging] PASSED                                                                                                                   [ 75%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_missing_journalist_alert[ansible://mon-staging] PASSED                                                                                                               [ 83%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_ossec_rule_journalist[ansible://mon-staging] PASSED                                                                                                                  [ 91%]
testinfra/ossec/test_journalist_mail.py::TestJournalistMail::test_journalist_mail_notification[ansible://mon-staging] FAILED                                                                                                           [100%]

[...]
        #
        # wait until at least one notification is sent
        #
>       assert self.wait_for_command(
            mon,
            "mailq | grep -q journalist@ossec.test")
E       AssertionError: assert False
E        +  where False = <bound method TestJournalistMail.wait_for_command of <test_journalist_mail.TestJournalistMail object at 0x35d7d528cd0>>(<testinfra.host.Host object at 0x35d7d334a90>, 'mailq | grep -q journalist@ossec.test')
E        +    where <bound method TestJournalistMail.wait_for_command of <test_journalist_mail.TestJournalistMail object at 0x35d7d528cd0>> = <test_journalist_mail.TestJournalistMail object at 0x35d7d528cd0>.wait_for_command

testinfra/ossec/test_journalist_mail.py:205: AssertionError
```

The good news is that I don't observe any AppArmor rules denied, so you can safely ignore my comment above :)

dachary:
> test_journalist_mail_notification[ansible://mon-staging] FAILED 

I suspect this is because mon-staging was up for some time and maybe used for testing. It does not happen in CI. But ... that also means the test is too fragile and I need to make sure it is more robust. 

- [ ] Note to self: test_journalist_mail_notification sometime fails and restarting ossec on mon make it run again. Make it more robust.

dachary:
@emkll thanks for the careful testing, I'll work to fix those :-)

dachary:
Creating a copy directory:
<pre>
mkdir /var/spool/mail/backup
chmod 777 /var/spool/mail/backup
</pre>

And adding

<pre>
VERBOSE=yes
MAILDIR=/var/mail/
DEFAULT=$MAILDIR
LOGFILE=/var/log/procmail.log
SUBJECT=`formail -xSubject:`

:0 c
backup
</pre>

at the beginning of `/var/ossec/.procmailrc` shows the original email from OSSEC looks like

<pre>
From ossec@freedom.press  Sun Feb 11 21:18:20 2018
Return-Path: <ossec@freedom.press>
X-Original-To: root@localhost
Delivered-To: root@localhost
Received: from notify.ossec.net (mon-staging [127.0.0.1])
        by ossec.server (Postfix) with SMTP id 9A897140419;
        Sun, 11 Feb 2018 21:18:20 +0000 (UTC)
To: <root@localhost>
From: OSSEC HIDS <ossec@freedom.press>
To: <journalist@localhost>
Date: Sun, 11 Feb 2018 21:18:20 +0000
Subject: OSSEC Notification - (app-staging) 10.0.1.2 - Alert level 7
Message-Id: <20180211211820.9A897140419@ossec.server>


OSSEC HIDS Notification.
2018 Feb 11 21:18:16

Received From: (app-staging) 10.0.1.2->head -1 /var/lib/securedrop/submissions_today.txt | grep '^[0-9]*$'
Rule: 400600 fired (level 7) -> "Number of submissions received in the past 24h."
Portion of the log(s):

ossec: output: 'head -1 /var/lib/securedrop/submissions_today.txt | grep '^[0-9]*$'':
0



 --END OF NOTIFICATION
</pre>

With **two** **To:** lines, hence the duplicate

dachary:
@emkll I believe the problems you found were resolved. Also discovered ossec will run all commands when started: there is no need for the perl onliner that raised an eyebrow ;-)
The duplicate is resolved by swapping the two lines in the procmailrc. And also got rid of the **c** flag which is only meaningfull when delivering mail, not when calling a command. The title of the mail was also changed.

I did not run a manual test with a real mail address but I'll do that now.

dachary:
The mail no longer goes to ossec and journalist, just to journalist. This much is good. But it is sent twice by ossec for a reason I don't understand. Looking into that now.

dachary:
@emkll I resolved the double email by removing code: less code, less bugs ;-) Instead of adding a new mail to OSSEC, I changed the procmailrc file to looking into the content of the notification and intercept it when it is a daily notification.

The double email was probably because the rule was added in a new group and that created a duplication. I was not able to find a good documentation of the semantics of groups in OSSEC so I can't confirm. But it does not matter since the added OSSEC logic was removed. The only change is a new rule to bump the priority of alerts reporting on the content of `/var/lib/securedrop/submissions_today.txt` to 7 so a mail notification is sent.

install_files/ansible-base/roles/ossec/tasks/configure_server.yml
```diff
@@ -21,11 +25,13 @@
   command: >
     gpg
     --homedir /var/ossec/.gnupg
-    --import /var/ossec/{{ ossec_alert_gpg_public_key }}
+    --import /var/ossec/{{ item }}
   become: yes
   become_user: "{{ ossec_group }}"
-  register: add_ossec_gpg_key_result
-  changed_when: "'imported: 1' in add_ossec_gpg_key_result.stderr"
+  when: "'{{ item }}' != ''"
```
emkll: If I understand, this is opt-in. If `site-specific` does not contain an entry for `journalist_alert_gpg_public_key`, it should throw an error such as `journalist_alert_gpg_public_key' is undefined`. Perhaps this statement should be replaced with something like ` {{ item }} is defined and item != '' `

dachary: Good catch ! I'll fix this.

dachary:
Fixed with the following. Setting defaults for **journalist_gpg_fpr** and **journalist_alert_email** although they are not currently used by the Ansible playbooks. I also removed them from **prod-specific.yml** because this file is obsolete and kept around for backward compatibility (or maybe it's just dead code that can be removed but that would need verification) and modifying it does not make sense.

<pre>
diff --git a/install_files/ansible-base/prod-specific.yml b/install_files/ansible-base/prod-specific.yml
index e01b36c2..c0ec5f94 100644
--- a/install_files/ansible-base/prod-specific.yml
+++ b/install_files/ansible-base/prod-specific.yml
@@ -34,9 +34,6 @@ securedrop_app_gpg_fingerprint: ""
 ossec_alert_gpg_public_key: ""
 ossec_gpg_fpr: ""
 ossec_alert_email: ""
-journalist_alert_gpg_public_key: ""
-journalist_gpg_fpr: ""
-journalist_alert_email: ""
 smtp_relay: ""
 smtp_relay_port: ""
 sasl_username: ""
diff --git a/install_files/ansible-base/roles/ossec/defaults/main.yml b/install_files/ansible-base/roles/ossec/defaults/main.yml
index 4395f9a5..c0704bfa 100644
--- a/install_files/ansible-base/roles/ossec/defaults/main.yml
+++ b/install_files/ansible-base/roles/ossec/defaults/main.yml
@@ -10,3 +10,14 @@ ossec_group: ossec
 # will execute tasks conditionally based on the var values.
 ossec_is_server: False
 ossec_is_client: False
+
+#
+# The journalist_* variables are optional and may not be present at
+# all in the site-specific file. Set an empty string as the default
+# for all of them to make it easier for tasks to test if they are set
+# or not by comparing to the empty string instead of testing if they
+# are defined and if they are set to the empty string.
+#
+journalist_alert_gpg_public_key: ""
+journalist_gpg_fpr: ""
+journalist_alert_email: ""
</pre>

The **ossec: send journalist about the number of submissions in the past 24h** commit was updated with this diff. And the series rebased. And force pushed :-) Thanks again for your excellent review !

emkll:
@dachary :+1: I've completed the testing of the PR, it looks great! Thank you for your diligence and for addressing my comments. From an implementation perspective and based on my testing, this PR seems to perform all functions as advertised and I believe the overall approach is sound.

Consider this approved from my perspective. Given the scope of changes, I would like to give time to others to review/comment/merge this PR. I would also like to discuss some fine implementation details at today's engineering meeting, specifically the email title and whether the notification should be a binary Y/N rather than the number of submissions.

dachary:
Thanks for your careful review: you caught significant problems and it significantly improved the quality of the PR 👍 I agree this should wait at least a few days and would benefit from another pair of eyes.

> I would also like to discuss some fine implementation details at today's engineering meeting, specifically the email title and whether the notification should be a binary Y/N rather than the number of submissions.

I stand ready to make the necessary change.

dachary:
During engineering meeting today @emkll suggested the mail sent is binary (are there submissions or not) instead of containing the number of submissions. The suggestion was approved and noone objected. It is implemented with the following fix:

<pre>
@@ -14,7 +14,11 @@ function main() {
     count=$(echo "$stdin" | perl -ne 'print scalar(<>) and exit if(/ossec: output/);')
     if [[ "$count" =~ ^[0-9]+$ ]] ; then
         export SUBJECT="Submissions in the past 24h"
-        printf "Submissions received in the past 24h: %-*d" 20 "$count" | $sender journalist
+        if [[ "$count" -gt "0" ]] ; then
+            printf "Submissions are waiting"
+        else
+            printf "NO submissions received"
+        fi | $sender journalist
     else
         export SUBJECT="SecureDrop Submissions Error"
         (
</pre>

Feel free to suggest alternate wording :-)

redshiftzero:
@conorsch or @msheiny y'all should review this too

securedrop/manage.py
```diff
@@ -398,10 +398,25 @@ def init_db(args):
...
     parser = argparse.ArgumentParser(prog=__file__, description='Management '
                                      'and testing utility for SecureDrop.')
     parser.add_argument('-v', '--verbose', action='store_true')
+    parser.add_argument('--data-root',
+                        default=config.SECUREDROP_DATA_ROOT,
+                        help=('directory in which the securedrop '
+                              'data is stored'))
+    parser.add_argument('--store-dir',
+                        default=config.STORE_DIR,
+                        help=('directory in which the documents are stored'))
```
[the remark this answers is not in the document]
dachary: because it was needed for tests with the shell based implementation. But what you proposed with the ORM is cleaner so it is no longer needed. I think it is good that all manage.py options can override the defaults from config because it helps with tests in general. If you're not against it I'd like to keep this commit adding --store-dir but I don't feel strongly about it.

install_files/ansible-base/roles/ossec/files/process_submissions_today.sh
```diff
@@ -0,0 +1,56 @@
...
+    local stdin
+    sender=${1:-send_encrypted_alarm}
+    stdin="$(< /dev/stdin)"
+
+    local count
+    count=$(echo "$stdin" | perl -ne 'print scalar(<>) and exit if(/ossec: output/);')
+    if [[ "$count" =~ ^[0-9]+$ ]] ; then
+        export SUBJECT="Submissions in the past 24h"
+        if [[ "$count" -gt "0" ]] ; then
+            printf "Submissions are waiting"
```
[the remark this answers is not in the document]
dachary: done

install_files/ansible-base/roles/ossec/files/process_submissions_today.sh
```diff
@@ -0,0 +1,56 @@
...
+        else
+            printf "NO submissions received"
```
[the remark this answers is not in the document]
dachary: done

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/build_securedrop_app_code_deb.yml
```diff
@@ -54,6 +54,17 @@
...
+  file:
+    state: directory
+    dest: "{{ securedrop_app_code_deb_dir }}/etc/cron.daily"
+
+- name: Copy crontab.
+  copy:
+    src: cron.daily
+    dest: "{{ securedrop_app_code_deb_dir }}/etc/cron.daily/securedrop"
+    mode: 0755
+
```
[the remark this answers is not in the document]
dachary: good catch, it is indeed a much more logical place

securedrop/manage.py
```diff
@@ -398,10 +398,25 @@ def init_db(args):
     os.chown('/var/lib/securedrop/db.sqlite', user.pw_uid, user.pw_gid)
 
 
+def how_many_submissions_today(args):
+    count_file = os.path.join(args.data_root, 'submissions_today.txt')
+    sh("""
+    find {dir} -type f -a -mmin -1440 | wc -l > {count_file}
+    """.format(dir=args.store_dir,
+               count_file=count_file))
```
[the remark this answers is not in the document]
dachary: much better indeed, done !

install_files/ansible-base/roles/ossec/templates/send_encrypted_alarm.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+# shellcheck disable=SC2086
```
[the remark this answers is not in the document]
dachary: that was because of $gpg_fpr which is already in quotes but ... it is also in $() so we can quote it. Fixed and exception removed.

docs/install.rst
```diff
@@ -69,6 +69,15 @@ continuing:
...
 -  The username of the system admin
 
+You can also, optionally, require that :doc:`daily notifications about
+the number of submissions <journalist>` received in the past 24h are
+sent via email so journalists know if it is worth checking the
+*Journalist Interface*. For this you will need:
+
+-  The journalist alerts GPG key
+-  The journalist alerts GPG key fingerprint
+-  The email address that will receive the journalist alerts
```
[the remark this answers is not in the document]
dachary: done

docs/install.rst
```diff
@@ -69,6 +69,15 @@ continuing:
    can add more later)
 -  The username of the system admin
 
+You can also, optionally, require that :doc:`daily notifications about
+the number of submissions <journalist>` received in the past 24h are
```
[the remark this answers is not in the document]
dachary: done

dachary:
@redshiftzero to facilitate review I address comments in commits marqued **SQUASH** so they can be squashed into the corresponding commit once they are to your satisfaction.

install_files/ansible-base/roles/ossec/tasks/configure_server.yml
```diff
@@ -21,11 +25,13 @@
   command: >
     gpg
     --homedir /var/ossec/.gnupg
-    --import /var/ossec/{{ ossec_alert_gpg_public_key }}
+    --import /var/ossec/{{ item }}
   become: yes
   become_user: "{{ ossec_group }}"
-  register: add_ossec_gpg_key_result
-  changed_when: "'imported: 1' in add_ossec_gpg_key_result.stderr"
```
redshiftzero: I think you can preserve these two lines such that this task is idempotent

admin/securedrop_admin/__init__.py
```diff
@@ -256,6 +293,19 @@ def __init__(self, args):
...
+             u'Local filepath to journalist alerts GPG public key (optional)',
+             SiteConfig.ValidateOptionalPath(self.args.ansible_path),
+             None],
+            ['journalist_gpg_fpr', '', str,
+             u'Full fingerprint for the journalist alerts '
+             u'GPG public key (optional)',
+             SiteConfig.ValidateOptionalFingerprint(),
+             self.sanitize_fingerprint],
+            ['journalist_alert_email', '', str,
+             u'Email address for receiving journalist alerts',
```
redshiftzero: `(optional)` should also appear for the email address for receiving journalist alerts

install_files/ansible-base/roles/ossec/tasks/configure_server.yml
conorsch: Two quick notes: 

1. `when` can accept a list, such that _all_ listed statements must evaluate to true for the task to execute
2. `when` statements should _not_ include Jinja curly braces—there's an Ansible deprecation warning to this effect now.

dachary:
@conorsch thanks for the when: deprecation of {{ }}, this is fixed in https://github.com/freedomofpress/securedrop/pull/2803/commits/aa8e10e5570b7f2021ea6345c4eaf07b59346b9b

install_files/ansible-base/roles/ossec/tasks/configure_server.yml
dachary: right. I removed them because I was not sure about the precedence of when/with_items compared to changed_when. Done

admin/securedrop_admin/__init__.py
dachary: done

install_files/ansible-base/roles/ossec/files/process_submissions_today.sh
```diff
@@ -0,0 +1,61 @@
...
+        fi | $sender journalist
+    else
+        export SUBJECT="SecureDrop Submissions Error"
+        (
+            echo "$0 failed to find the number of submissions in the following OSSEC alert"
+            echo
+            echo "$stdin"
+        ) | $sender ossec
+    fi
+}
```
redshiftzero: Since OSSEC by default [groups alerts](https://github.com/ossec/ossec-hids/blob/72641d6f22c63b97f290ae22d47a79032b56d0fd/etc/internal_options.conf#L59) together, it's possible to get a "SecureDrop Submissions Error" sent to the administrator if there are multiple alerts that occur in a short time span. This alert contains both the contents of the submissions text file on disk along with the other alert(s) that occur. One way to reproducibly trigger this is by rebooting the VMs:

```
/var/ossec/process_submissions_today.sh failed to find the number of submissions in the following OSSEC alert

>From ossec@freedom.press  Sat Feb 24 18:44:02 2018
Return-Path: <ossec@freedom.press>
X-Original-To: root@localhost
Delivered-To: root@localhost
Received: from notify.ossec.net (mon-staging [127.0.0.1])
	by ossec.server (Postfix) with SMTP id 7170C1202F6
	for <root@localhost>; Sat, 24 Feb 2018 18:44:02 +0000 (UTC)
To: <root@localhost>
From: OSSEC HIDS <ossec@freedom.press>
Date: Sat, 24 Feb 2018 18:44:02 +0000
Subject: OSSEC Notification - (app-staging) 10.0.1.2 - Alert level 7
Message-Id: <20180224184402.7170C1202F6@ossec.server>


OSSEC HIDS Notification.
2018 Feb 24 18:43:50

Received From: (app-staging) 10.0.1.2->netstat -tan |grep LISTEN |grep -v 127.0.0.1 | sort
Rule: 533 fired (level 7) -> "Listened ports status (netstat) changed (new port opened or closed)."
Portion of the log(s):

ossec: output: 'netstat -tan |grep LISTEN |grep -v 127.0.0.1 | sort':
tcp        0      0 0.0.0.0:111             0.0.0.0:*               LISTEN     
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN     
tcp        0      0 0.0.0.0:55436           0.0.0.0:*               LISTEN     
tcp        0      0 0.0.0.0:6001            0.0.0.0:*               LISTEN     
tcp6       0      0 :::111                  :::*                    LISTEN     
tcp6       0      0 :::40520                :::*                    LISTEN     
tcp6       0      0 :::6001                 :::*                    LISTEN     
Previous output:
ossec: output: 'netstat -tan |grep LISTEN |grep -v 127.0.0.1 | sort':
tcp        0      0 0.0.0.0:111             0.0.0.0:*               LISTEN     
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN     
tcp        0      0 0.0.0.0:53212           0.0.0.0:*               LISTEN     
tcp        0      0 0.0.0.0:6001            0.0.0.0:*               LISTEN     
tcp6       0      0 :::111                  :::*                    LISTEN     
tcp6       0      0 :::54287                :::*                    LISTEN     
tcp6       0      0 :::6001                 :::*                    LISTEN     



 --END OF NOTIFICATION



OSSEC HIDS Notification.
2018 Feb 24 18:43:50

Received From: (app-staging) 10.0.1.2->head -1 /var/lib/securedrop/submissions_today.txt | grep '^[0-9]*$'
Rule: 400600 fired (level 7) -> "Number of submissions received in the past 24h."
Portion of the log(s):

ossec: output: 'head -1 /var/lib/securedrop/submissions_today.txt | grep '^[0-9]*$'':
1



 --END OF NOTIFICATION
```

dachary: **very** good catch ! Will add [do_not_group](https://ossec-docs.readthedocs.io/en/latest/syntax/head_ossec_config.email_alerts.html#element-do_not_group) to prevent that

install_files/securedrop-ossec-server/var/ossec/rules/local_rules.xml
```diff
@@ -190,3 +190,13 @@
     <options>no_email_alert</options>
   </rule>
 </group>
+
+<group name="Ossec daily notifications">
+  <rule id="400600" level="7" >
+    <if_sid>530</if_sid>
+    <match>ossec: output: 'head -1 /var/lib/securedrop/submissions_today.txt</match>
+    <description>Number of submissions received in the past 24h.</description>
```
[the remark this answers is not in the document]
dachary: done

securedrop/manage.py
```diff
@@ -399,10 +401,38 @@ def init_db(args):
     os.chown('/var/lib/securedrop/db.sqlite', user.pw_uid, user.pw_gid)
 
 
+def how_many_submissions_today(args):
```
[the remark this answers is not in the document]
dachary: done

install_files/ansible-base/roles/ossec/files/process_submissions_today.sh
dachary: do_not_group does not actually help because the alerts are grouped and do_not_group applies to mails not being grouped which is a different problem. I'm now trying a different approach which is to lower the level of the alert (level 1) so it is not grouped with other level 7 alerts. And I'm creating an ossec.conf email_alerts dedicated to this rule. It may not work as I think it does though...

dachary: I made it so two alerts are sent at the same time and got two separate emails. Although this is not documented, I think it makes sense that the level 1 alert regarding submissions is not grouped with level 7 alerts. @redshiftzero it would be nice if you could repeat what you did to see the problem and verify if it really is gone. If you can document the steps I'd be happy to do it myself. I tried to reboot the staging vm but I doubt that's what you did because it requires manually running the ossec server and we're no longer in the same situation as the one you experienced ;-)

<pre>
# mailq
-Queue ID- --Size-- ----Arrival Time---- -Sender/Recipient-------
17FC41403F4     1105 Sun Feb 25 14:18:29  ossec@ossec.server
(Host or domain name not found. Name service error for name=smtp.faketld type=A: Host not found)
                                         journalist@ossec.test

15D251403F6     1512 Sun Feb 25 14:18:29  ossec@ossec.server
(Host or domain name not found. Name service error for name=smtp.faketld type=A: Host not found)
                                         ossec@ossec.test

-- 2 Kbytes in 2 Requests.
# postcat -q 15D251403F6 | gpg --homedir /var/ossec/.gnupg/ --decrypt                                           
OSSEC HIDS Notification.
2018 Feb 25 14:18:27

Received From: (app-staging) 10.0.1.2->syscheck
Rule: 550 fired (level 7) -> "Integrity checksum changed."
Portion of the log(s):

Integrity checksum changed for: '/var/ossec/etc/ossec.conf'
Size changed from '4608' to '4657'
Old md5sum was: 'b998b945f355a8d37883e124960b0214'
New md5sum is : '124e883d6e69498ae3a14a697332afee'
Old sha1sum was: '52f8d74b7087eb69efaae51921d8b47714855656'
New sha1sum is : '22e0dfca73d67879ca89e815ef9530b4784ae638'



 --END OF NOTIFICATION

# postcat -q 17FC41403F4 | gpg --homedir /var/ossec/.gnupg/ --decrypt                                           

There has been no submission activity in the past 24 hours.
You do not need to login to SecureDrop.
</pre>

dachary:
@redshiftzero your [comments](https://github.com/freedomofpress/securedrop/pull/2803#discussion_r170430840) were addressed in [1](

install_files/ansible-base/roles/ossec/files/process_submissions_today.sh
redshiftzero: Nice fix - with this change the submissions error message does not occur :clap:.

For posterity, I verified this by:

0. checking out the latest changes here
1. `make build-debs`
2. `vagrant provision /staging/`
3. `vagrant reload /staging/` - this is what produced the error originally
4. Checking my OSSEC alerts

dachary: That makes me curious because ... I don't have OSSEC or postfix running when I provision staging? Did you tweak the vagrant files ? Or am I misisng something ?

redshiftzero: Ah so when I'm testing a PR where it would be convenient to have both speedy SSH _and_ OSSEC emails, I edit the staging group variables like so:

```
diff --git a/install_files/ansible-base/group_vars/staging.yml b/install_files/ansible-base/group_vars/staging.yml
index c27e5a96..66612ab5 100644
--- a/install_files/ansible-base/group_vars/staging.yml
+++ b/install_files/ansible-base/group_vars/staging.yml
@@ -31,21 +31,10 @@ apache_logging_level: "info"
 ### Used by the mon role ###
 # The OSSEC alert GPG public key has to be in the install_files/ansible-base/ or
 # install_files/ansible-base/roles/app/files/ directory
-ossec_alert_gpg_public_key: "test_admin_key.pub"
-ossec_gpg_fpr: "53E1113AC1F25027BA5D475B1141E2BBB5E53711"
-ossec_alert_email: "ossec@ossec.test"
-journalist_alert_gpg_public_key: "test_admin_key.pub"
-journalist_gpg_fpr: "53E1113AC1F25027BA5D475B1141E2BBB5E53711"
-journalist_alert_email: "journalist@ossec.test"
-smtp_relay: "smtp.faketld"
-smtp_relay_port: "587"
-sasl_username: "test"
-sasl_domain: "ossec.test"
-sasl_password: "password123"
 
 # Disable Postfix in staging, so we don't hammer Google mail relays
 # with known-bad credentials.
-postfix_enable_service: no
+postfix_enable_service: yes
```

dachary:
@redshiftzero added a [test to verify unencrypted messages have the same size](https://github.com/freedomofpress/securedrop/pull/2803/commits/aef0cde3e4aedc670a00fb5fc054dbd4a200171b) and checked that it fails if we remove characters in one of the messages.

install_files/ansible-base/roles/ossec/files/process_submissions_today.sh
dachary: nice trick !

redshiftzero:
The last thing we want to verify is that the ciphertexts produced in the two cases (submissions vs. no submissions) cannot be distinguished - i.e. we want the ciphertext lengths in these two cases to be identically distributed. 

Note that there is some variation in the length of ciphertexts due to the slight differences in the lengths of the OpenPGP packet used to set up the (random) symmetric session key. You can see this in the following output, which shows the (differing) length and OpenPGP packet output of two ciphertexts that were produced by encrypting identical plaintexts (OpenPGP packet output using [`pgpdump`](http://www.mew.org/~kazu/proj/pgpdump/en/)): 

```
ciphertext has length:  878
Old: Public-Key Encrypted Session Key Packet(tag 1)(524 bytes)
	New version(3)
	Key ID - 0x55B23724E7A8DF03
	Pub alg - RSA Encrypt or Sign(pub 1)
	RSA m^e mod n(4091 bits) - ...
		-> m = sym alg(1 byte) + checksum(2 bytes) + PKCS-1 block type 02
New: Symmetrically Encrypted and MDC Packet(tag 18)(72 bytes)
	Ver 1
	Encrypted data [sym alg is specified in pub-key encrypted session key]
		(plain text + MDC SHA1(20 bytes))

ciphertext has length:  874
Old: Public-Key Encrypted Session Key Packet(tag 1)(523 bytes)
	New version(3)
	Key ID - 0x55B23724E7A8DF03
	Pub alg - RSA Encrypt or Sign(pub 1)
	RSA m^e mod n(4087 bits) - ...
		-> m = sym alg(1 byte) + checksum(2 bytes) + PKCS-1 block type 02
New: Symmetrically Encrypted and MDC Packet(tag 18)(72 bytes)
	Ver 1
	Encrypted data [sym alg is specified in pub-key encrypted session key]
		(plain text + MDC SHA1(20 bytes))
```

This variation should not depend on the message content, so this variation is fine. However, one issue that _can_ impact this PR is that by default, OpenPGP will [compress plaintexts](https://tools.ietf.org/html/rfc4880#section-13.3.1) prior to encrypting, which can introduce a message length difference. However, @emkll noted it's possible to just turn off compression - he's currently testing this and will report back here.

emkll:
In my initial tests (when this functionality sent the number of submissions), I had measured the armored payload of the encrypted email. As @redshiftzero suggests, one can go further by using `pgpdump` to get more details in the various packet in the GPG packets. With N=29, all base 64 GPG-armor blocks were the same size and all Packets displayed by gpgdump (tags 1 and 18) were all the same size.

I have (briefly) tested the scenario tested above and I can confirm the behavior observed, specifically variance in tag 18 (Symmetrically Encrypted and MDC Packet). Though my sample size is limited (N=10),  emails where there's a submission are longer on average. Perhaps @redshiftzero might have some data to share later today.

When disabling GPG compression ( by invoking gpg with `compression-algo none` `send_encrypted_alarm.sh`), all GPG-armored portions of the email are of identical length, and `pgpdump` reports identical lengths for all emails (whether there is or isn't a submission): for all emails, I observe:
```
Old: Public-Key Encrypted Session Key Packet(tag 1)(524 bytes)
[...]
New: Symmetrically Encrypted and MDC Packet(tag 18)(151 bytes)
```
It sounds like the hypothesis of GPG compression as stated above is likely in this scenario. Given the padding is done with spaces, it may allow an attacker distinguish between submissions and no submission email (and likely why this behavior was not observed in the previous iteration, as the padding was always the same, due to 1 digit)

Given that:

0. An attacker does not have the GPG private key
1. GPG encryption uses a random symmetric key to prevent symantic disclosure
2. Encrypted message length is always the same (with compression enabled)
3. Messages are sent at the same exact time every day

I cannot think of a situation where an attacker with visibility on the email traffic could distinguish between either versions of the email.

dachary:
@emkll @redshiftzero compression is disabled when sending journalist notifications at [1](https://github.com/freedomofpress/securedrop/pull/2803/commits/ffd230d091b863dc2759d920b28695fd3ebf92d4)

dachary:
@emkll rebased & squashed & repushed !

dachary:
rebased and repushed, resolved a trivial conflict due to SDConfig being merged

dachary:
@conorsch can I interest you in a review of this pull request? I realize I'm a little needy but I have good reasons :-) Or is it an excuse? Anyways: here it is. @emkll and @redshiftzero spent time reviewing the functional aspect and this is all fresh in their mind. And mine as well. Things will go smoothly if a review from you happen in the next few days.

Of course, if you do not have time to review it will not be catastrophic! I'm just pushing a little because the time is right. The worst that can happen if you review in a week or two is that people will have to refresh their memory when that happens. And possibly a few rebase but that should not be a major problem.

install_files/ansible-base/roles/ossec/templates/send_encrypted_alarm.sh
```diff
@@ -16,34 +16,56 @@ set -o pipefail
...
+    local alert_email
+    local compression
+    if [[ "$recipient" = "journalist" ]] ; then
+        gpg_fpr='{{ journalist_gpg_fpr }}'
+        alert_email="$JOURNALIST_EMAIL"
+        if [[ "$alert_email" = "" ]] ; then
+            echo "journalist alert email unset, no notification sent"
+            return
+        fi
+        compression="--compress-algo none"
```
redshiftzero: I had to modify this to `compression="--compress-algo=none"` or I get a failure to encrypt message

redshiftzero: Hmm, also the failure message went to the journalist and not the admin - I think this is ok (otherwise the journalist might wonder "where's my daily email") but we could add a "Please contact your administrator" string

dachary: > I had to modify this to ...

Ouch. It means the tests do not run this code path for some reason... I'll look into that.

dachary:
> I had to modify this to compression="--compress-algo=none" or I get a failure to encrypt message

So... we need **--compress-algo=none** instead of **--compress-algo none** because "$compression" makes it a single argument instead of two because of the quotes. The reason it did not trigger an error in the tests is because of the other bug: errors should **not** be sent to the journalist, they should be sent to the admin. The test does not decrypt the message, it just verifies the mail is sent to the expected recipient: the journalist. And since the error mail was sent to the journalist... the test passed. By fixing the two problems in https://github.com/freedomofpress/securedrop/pull/2803/commits/42ddbc3604f7fb290d740a3ef992fe4971970d7a we know the tests now cover the validity of the compression argument. If it did not, the mail would got to the admin instead of the journalist.

@redshiftzero please let me know if that is good enough or if I should add more tests.

redshiftzero:
Sounds good, thanks for the quick fix @dachary. 

In terms of turning off compression, I did a quick sanity check by collecting between ~50-200 ciphertexts for each of the following cases:

* Submissions, compression on
* No submissions, compression on
* Submissions, compression off
* No submissions, compression off

In the following plot showing normalized ciphertext length distributions based on this data, one can see in the left panel with compression on, the distributions of ciphertexts with and without submissions can be clearly distinguished. The right panel shows the result with the compression turned off - all observed emails were exactly the same length. 

journo_notifications

(Note that the submissions bars are shown to be thinner for readability, but the error bar in the x-direction represents the ciphertext bin width, which is equal for both the submission and no submission cases. Error bars in counts are computed assuming a Poisson distribution)

dachary:
> I did a quick sanity check by collecting between

I really love the graph and scientific approach. It provides an aura of respectability to this pull request that is impressive. I'm curious about the snippets of code / tools you used to do that :-)

redshiftzero:
:) sure! Just pushed the [Python notebook I used here](https://github.com/redshiftzero/securedrop-2803/blob/master/journalist-notifications-pr-2803.ipynb)

dachary:
@conorsch a gentle reminder about reviewing the Ansible parts of this pull request. Please let me know if there is anything I can do to make it easier for you to review.

dachary:
@conorsch ping?

dachary:
rebased to resolve conflicts because the test_manage.py became pytest friends :tada: and io.open got cleaned up :tada:. Nothing changed otherwise @conorsch, I promise ;-)

dachary:
@conorsch rebased & repushed just to get a clean run but no conflicts.

conorsch:
Spent some time with this one today. I've got prod VMs running with the new config active, and although OSSEC alerts are streaming in fine, I've yet to receive any journalist notifications. (I made sure to run `./securedrop-admin sdconfig` and fill out the relevant fields). Going to let these VMs run for a few days _without_ triggering any special behavior and confirm that the notifications come in without further nudging.

Made sure to make at list one submission, which I'll expect to find in my inbox shortly. 

A visual review of the diff looks remarkably pleasant. Considerable testing as been added, on the underlying logic to enable the new functionality is surprisingly concise. Even the longwinded bash scripts are pleasant to read and understand. Thanks for your careful attention to detail here, @dachary—looking forward to getting this in!

Will report back tomorrow with results from the long-running config change.

docs/journalist.rst
```diff
@@ -57,6 +57,25 @@ Journalist Interface <yubikey_setup>`.)
...
 |Journalist Interface Login|
 
+Daily journalist alerts about submissions
+-----------------------------------------
+
+When a SecureDrop has little activity and receives only a few
+submissions every other week, checking the *Journalist Interface*
+daily only to find there is nothing is a burden. It is more
+convenient for journalists to be notified daily via email about the
+number of submissions received in the past 24h.
```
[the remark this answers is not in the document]
dachary: done. Somehow this slipped... thanks for the reminder.

dachary:
> Update: haven't received any messages yet.

Could you please try to submit one document today? And tomorrow morning, if you did not receive any e-mail, could you 
<pre>
ls -l /var/lib/securedrop/submissions_today.txt
cat /var/lib/securedrop/submissions_today.txt
</pre>
If that file exists and contains a number different from zero, it will help me figure out what's not working as expected.

If you receive an email tomorrow, it probably means there is a corner case  when bootstraping the system.

Thanks a lot for your patience on this one: it's really good we get a chance to diagnose the problem now :-)

conorsch:
@dachary The problem was indeed that `/var/lib/securedrop/submissions_today.txt` didn't exist yet. I've submitted another doc to the test instance and am currently waiting again. If the @daily marker doesn't trigger the cron job within the next few hours, I'll kick it off manually to expedite review. 

It occurs to me that perhaps a slightly more frequent cron schedule would help to keep the "submissions _today_" logic accurate. That's a nit, though. This is definitely a huge step up in terms of actionable information for journalists using the system.

dachary:
> If the @daily marker doesn't trigger the cron job within the next few hours, I'll kick it off manually to expedite review.

We can wait another 24h, it really is valuable to be in a realistic scenario, thanks again for your patience doing this.

conorsch:
My previous attempt at testing was misguided: I used prod VMs, mostly to validate the Tails workflow on setting up the new vars, but failed to install the locally built debs. Retried in staging with overridden vars (similar to what @redshiftzero described in https://github.com/freedomofpress/securedrop/pull/2803#discussion_r170720578) and can confirm working messages:

```
There has been submission activity in the past 24 hours.   
You should login and check SecureDrop. 
```

Overall the Admin-facing workflow is impressively smooth. Because the new functionality is disabled by default, I'm comfortable with this going out as-is. Noticed some extremely small typos that could be cleaned up, will comment in-line.

install_files/ansible-base/roles/ossec/templates/send_encrypted_alarm.sh
conorsch: I didn't encounter this problem; @redshiftzero do you still recommend the change here?

securedrop/manage.py
```diff
@@ -295,6 +327,14 @@ def get_args():
     return parser
 
 
+def set_were_there_submissions_today(subps):
+    parser = subps.add_parser(
+        'were-there-submissions-today',
+        help=('Update the file indicating '
+              'iff submissions were received in the past 24h'))
```
conorsch: Nit: `s/iff/whether/`

docs/install.rst
```diff
@@ -60,6 +60,21 @@ continuing:
    can add more later)
 -  The username of the system admin
 
+You can also, optionally, configure a :doc:`daily notifications
+<journalist>` about whether or not submission activity occurred in the
```
conorsch: Nit: remove "a " in "a daily notifications".

install_files/ansible-base/roles/ossec/templates/send_encrypted_alarm.sh
redshiftzero: I have _not_ run on the latest changes here, but when I last tested I did have to make this change (to generate emails for my compression vs no compression testing last month). I would need to check again in VMs and verify it's no longer the case... I can look into this on Monday

redshiftzero: Ok, just re-ran on the latest, and things are indeed working as is, so this change is not needed (not 100% sure what was happening before, but since I made some modifications to test the ciphertext lengths it's possible I introduced an unrelated issue).

redshiftzero:
The lint CI job is consistently failing due to taking too long ([build here](https://circleci.com/gh/freedomofpress/securedrop/11427)), this is a new one...

redshiftzero:
Ok waited a while, restarted and now the lint job is passing 🙃. Going to chalk this one up to fluctuations in the Circle CI force

redshiftzero:
TIL about CircleCI workflows: to failed jobs that are dependencies for other jobs: one must [do this](https://circleci.com/docs/2.0/workflows/#rerunning-a-workflows-failed-jobs) or the parent jobs will not run

redshiftzero:
Merging, thanks @dachary for your patience on this one and thanks to @emkll and @conorsch for the reviews and testing 😃

PR #2804, opened 2018-01-07
tests: only upload coverage information if running on CircleCI

dachary:
## Status

Ready for review

## Description of Changes

Trying to upload to codecov while running **make dev** on the local machine will likely result in failure. Unless the developer has codecov credentials properly setup which is rare.

## Testing

* cd securedrop
* make test

## Deployment

N/A

### DIFF
--- a/securedrop/bin/test
+++ b/securedrop/bin/test
@@ -9,12 +9,14 @@ ln -s /dev/urandom /dev/random
 
 x11vnc -display :1 -autoport 5901 -shared &
 
-touch tests/log/firefox.log
-function finish {
-  cp tests/log/firefox.log /tmp/test-results/logs/
-  bash <(curl -s https://codecov.io/bash)
-}
-trap finish EXIT
+if [ -n "${CIRCLE_BRANCH}" ] ; then
+    touch tests/log/firefox.log
+    function finish {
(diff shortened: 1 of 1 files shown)


securedrop/bin/test
```diff
@@ -9,12 +9,14 @@ ln -s /dev/urandom /dev/random
 
 x11vnc -display :1 -autoport 5901 -shared &
 
-touch tests/log/firefox.log
-function finish {
-  cp tests/log/firefox.log /tmp/test-results/logs/
-  bash <(curl -s https://codecov.io/bash)
-}
-trap finish EXIT
+if test "${CIRCLE_BRANCH}" ; then
```
heartsucker: Not a blocker, but I think the `[ -n "${CIRCLE_BRANCH}" ]` notation is more clear than `test` and would slightly prefer that as it makes skimming code easier.

dachary: right ! done :-)

dachary:
rebased & repushed, thanks for your review @heartsucker :-)

heartsucker:
I'm going to let someone else merge this since there's not a related issue / discussion, and i don't want to be the only person voting yes on this.

PR #2805, opened 2018-01-08
tests: no need to remove *.pyc before docker build

dachary:
## Status

Ready for review

## Description of Changes

because they are already excluded by .dockerignore. In addition the
--no-run-if-empty option is not implemented on some operating systems
and will fail

## Testing

On MacOS verify that 

* make test

works.

## Deployment

N/A

### DIFF
--- a/securedrop/.dockerignore
+++ b/securedrop/.dockerignore
@@ -1,4 +1,5 @@
-*.pyc
+**/__pycache__
+**/*.pyc
 .sass-cache
 static/gen
 static/.webassets-cache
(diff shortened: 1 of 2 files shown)


dachary:
@tmc turns out *.pyc was already in .dockerignore which makes me wonder how the error documented in c971ac9eb3c3c6e2fda42985f1ec263f03bc4fdd could have happened. Let see if it shows up again to clarify this mystery

dachary:
> We should add __pycache__ to the .dockerignore as part of this PR

Good idea ! Rebased & modified & repushed.

tmc:
I originally implemented dockerignore but had to reference docs. We need ‘**/*.pyc’ and the same for pycache.

dachary:
Good catch, @tmc is this better now ?

tmc:
LGTM

Issue #2727, opened 2017-12-12
Ansible install - utilize state present instead of latest


msheiny:
# Feature request

## Description

Using `state: latest` with the `apt` module forces apt_cache to refresh and is unnecessary on a re-run of the ansible install playbooks. The nightly cron job is responsible for this, lets try to speed up the ansible install runner so it's not so burdensome to run again over tor to make changes. 

Particuarly, we should swap all those cases of `latest` to utilize `present` instead.

This issue should probably be done in the same PR as #2724 . 

## User Stories

As a securedrop admin, when I want to make app changes and run `securedrop-admin install` again, I want that process to be as quick as possible.

dachary:
so the way forward would be to check all apt and replace `latest` with `present` when we don't care about `latest`, right ? In some cases we actually want `latest` and we'll keep it.

kushaldas:
The latest packages will be installed by the cron job every night. That will save the time for `apt` cache check for every `apt` install using the `latest` value.

dachary:
Yes, but if package-1.0 is present and package-1.1 is available (and maybe we want the `latest`), if we're using `present` we will be stuck with package-1.0. Or am I missing something ?

conorsch:
This would still be a solid improvement. We're needlessly re-updating the apt lists several times throughout the install flow. More recent versions of ansible-lint discourage use of state=latest. Would certainly review a PR if someone wanted to tackle this conversion.

Issue #1742, opened 2017-05-26
Create dedicated playbooks for common Admin tasks


conorsch:
The current workflow for updating the instance configuration is to update relevant Ansible vars, then run the prod playbook in its entirety. That's rather effective, but it's quite slow, given the SSH-over-Tor setup. For common tasks that Admins have, we should create small, tightly scoped playbooks that handle only the required task, and do not also run all the setup scripts. A few examples:

* perform backup and fetch tarball back to Admin Workstation
* collect logs, encrypt to FPF support key, fetch back encrypted tarball to Admin Workstation
* rotate SecureDrop Application GPG Key (including cleaning up references to old pubkeys in the app keyring)

Heavy use of tagging (which will land in 0.4) helps to target certain functionality within the prod playbook, but that's a poor substitute for dedicated workflows handling common operations.

eloquence:
Relevant related tickets:
- #5297 could significantly simplify the Journalist Workstation provisioning story (at least until Qubes, for which we have separate provisioning code)
- #3325 suggests making languages configurable via the web UI; moving other configuration settings into the web UI would reduce the reliance on playbooks altogether.

Issue #1532, opened 2017-01-25
Customized ISO distribution of SD Application and Monitor servers


msheiny:
Since we are already tightly controlling the OS experience through ansible policies on first playbook install I think we should invest more time to move some of the logic upfront during the install process. We should maintain our own spin of Ubuntu with baked in:

- SecureDrop GPG key and repos already in place
- Basic security hardening tasks in place
- Strip out all non essential software for SD
- Automate as much of the install process as possible
- bad-ass `ASCII` art and customized ISO bootup logos

You could also make the argument (though I'm not really) that the gap between vanilla Ubuntu install and SD install is a vulnerable time for exploitation. We have no idea where administrators will stage these servers. It could have been set up in a relatively insecure network and then moved to an isolated location for the SD installation piece. In other words, this would close the gap of exposure between Vanilla Ubuntu and hardened SD configuration.  

Now when I say `maintain` our own distro ... I mean take the stock upstream Ubuntu/Debian and sprinkle in our bits. I don't advocate that FPF gets in the OS shipping business and I feel there is a distinction here with what I'm proposing.

eloquence:
Given the effort and complexity, we're unlikely to implement this; in any event, we'll want to consider server OS alternatives first (#5517). Closing.

PR #2809, opened 2018-01-08
Support tags in securedrop-admin script

### LINKED ISSUE #2624: Permit passing `--tags` to Ansible
# Feature request

## Description

Related to #1742, which would ease the workflow for Admins when updating configuration of an already running SecureDrop instance. Rather than create dedicated playbooks for each task, we leverage Ansible "tags" to run only subsets of the larger playbook. Since prod installs route SSH over Tor, an entire playbook run can take nearly half an hour. For a one-line configuration change, that's a long time.

Let's update the `./securedrop-admin install` command to support an optional `--tags` argument for running only specific portions of the playbook. We could reuse this logic on future securedrop-admin subcommands to run only portions of the configuration logic.

## User Stories

As a SecureDrop Administrator, I don't want to wait 30 minutes every time I need to update one setting. If you expect me to wait 30 minutes for a single file change, I'm likely to log in interactively and edit the files manually, sidestepping the error-checking and validation within the Ansible logic.

As a SecureDrop maintainer, I want to test configuration changes in a prod-like environment (i.e. SSH over Tor), and waiting 30 minutes for each playbook run slows down development.

conorsch:
## Status

Ready for review.

## Description of Changes

Fixes #2624.

Each time a SecureDrop Admin wishes to alter the server configuration,
they must run the *entire* production playbook. For an already running
instance, with the connection routed over Tor, the run typically takes
20-30 minutes. Sometimes the underlying change could be made manually
(albeit less reliably) via manual edits in 1-2 minutes. Let's permit
Admins to filter by `--tags` to run only subsets of changes.

We may wish to fine-tune the use of tags throughout the configuration
logic to ensure that tagged subsets of tasks are meaningful and useful.
At present, one useful subset already well tagged is "ossec", e.g.
`./securedrop-admin install --tags ossec`.

## Testing

1. Use Prod VMs and a Tails-based Admin Workstation (virtualized is OK).
2. Install as normal until setup is fully provisioned.
3. Run _only_ the OSSEC tasks with `./securedrop-admin install --tags ossec`. 
4. Run _only_ the "common" tasks with `./securedrop-admin install --tags common`.
5. Observe faster runs without errors.

## Deployment
Should not break any existing configuration logic. The goal is to make configuration more convenient and therefore more predictable: we want to encourage Admins to use the tooling provided for determining server state. If that's going to happen, the tooling must be convenient enough that it's more attractive than ad-hoc manual changes (which are not tracked and easily overwritten).

### DIFF
--- a/securedrop/securedrop-admin
+++ b/securedrop/securedrop-admin
@@ -215,9 +215,18 @@ def install_securedrop(args):
                    "servers.")
         sdlog.info("The sudo password is only necessary during initial "
                    "installation.")
-        subprocess.check_call([os.path.join(ANSIBLE_PATH,
-                                            'securedrop-prod.yml'),
-                              '--ask-become-pass'], cwd=ANSIBLE_PATH)
+        ansible_cmd = [
+            'ansible-playbook',
+            os.path.join(ANSIBLE_PATH, 'securedrop-prod.yml'),
+            '--ask-become-pass',
+        ]
+        if args.tags:
+            ansible_cmd.append('--tags')
+            ansible_cmd.append(','.join(args.tags))
+
+        sdlog.info("Preparing to run Ansible command:")
+        sdlog.info(ansible_cmd)
+        subprocess.check_call(ansible_cmd, cwd=ANSIBLE_PATH)
 
 
 def backup_securedrop(args):
@@ -310,7 +319,12 @@ if __name__ == "__main__":
 
     parse_install = subparsers.add_parser('install',
                                           help=install_securedrop.__doc__)
+
     parse_install.set_defaults(func=install_securedrop)
+    parse_install.add_argument(
+        '-t', '--tags', action='append',
+        default=[],
+        help="Run only configuration matching specified Ansible tags")
 
     parse_tailsconfig = subparsers.add_parser('tailsconfig',
                                               help=run_tails_config.__doc__)


dachary:
you'll be happy to learn there now are (a few ;-) tests for securedrop-admin ``securedrop/tests/test_securedrop-admin.py`` :-)

securedrop/securedrop-admin
```diff
@@ -310,7 +319,12 @@ if __name__ == "__main__":
 
     parse_install = subparsers.add_parser('install',
                                           help=install_securedrop.__doc__)
+
     parse_install.set_defaults(func=install_securedrop)
+    parse_install.add_argument(
+        '-t', '--tags', action='append',
+        default=[],
+        help="Run only configuration matching specified Ansible tags")
```
redshiftzero: Question: do we want admins to know what Ansible is? We should try to abstract away that we're running an Ansible playbook under the hood. For the purpose of this feature, they just need to roughly know what the possible tags are and what they do. In lieu of a SecureDrop admin option like `securedrop-admin list-tags` (running `--list-tags`) we can just list this in the documentation, along with a sentence or so describing what running with that tag will do.

heartsucker: Agree that we shouldn't have to expose ansible to the admins.

conorsch:
Great points, @redshiftzero and @heartsucker. 

> We should try to abstract away that we're running an Ansible playbook under the hood. 

I agree that we should be abstracting more of the technical details about software choices from SecureDrop Administrators. What's really important is that we clearly identify "this task will configure the servers." As for whether it's using Ansible, or Salt, or fairy dust, that's unimportant, and so much so we'd do well to mask it with a prettier interface. Such was the motivation for the `securedrop-admin` script in the first place (came in via #1781). 

> I think if we want admins to be able to run tasks, we need to make them into playbooks.

Agreed! This PR would close #2624, but #1742 tracks playbook addition or conversion in order to make discrete tasks more reliable (and faster running). 

> Point being, we don't know how these might be run and we might forget to tag all the transitive dependencies of a single task which would break things.

This is quite close to the mark, in my mind. To be frank, during development of config changes (e.g. #2478), I frequently _manually activate the virtualenv_ to get access to `ansible-playbook` on the command line, so that I can pass `--tags` to target a given play or role. I do this because SSH over Tor is slow, and testing in prod VMs is the best way to evaluate candidate functionality before merge. In effect, I'm not using the `securedrop-admin` 100% of the time, when I definitely should be—but doing so would vastly increase the QA time before opening a new PR, or when reviewing an already open PR.

So perhaps my initial description was a bit disingenuous: I'm mostly concerned about lack of proper QA. While I admit it's improper to "cheat" by manually activating the virtualenv, it's hardly feasible to block review for several days simply because I'm waiting for known-good aspects of the entire prod playbook to run. By testing the tags option ourselves, we can shake out any inconsistencies (as @heartsucker rightly points out there may well be!) _before_ documenting the feature for Admins.

Feel free to close this issue @redshiftzero, but please be aware of the also-undocumented divergence from scripting testing procedures as a result of dev env friction (specifically in the prod-VM/Tails setup).

redshiftzero:
Thanks for this detailed explanation @conorsch. While I do think this would be useful for us during QA or for admins who, e.g. want to deploy the changes in #2478, but don't want to re-run the entire playbook due to the long run time, I agree with @heartsucker - really we would have to test that each set of tasks corresponding to a particular tag work as anticipated when executed individually. If we're not testing that (and we aren't) then we could only include this as an undocumented option and test it before requesting an admin run it. As including undocumented options due to lack of testing is an unwise decision, I think the right call here is to close this PR, and encourage instead any changes that can be made to decrease the run time of the Ansible installation playbook (e.g. #2727, #1532).

Issue #2760, opened 2017-12-20
./securedrop-admin sdconfig failure on SMTP host checking


kushaldas:
# Bug

## Description

Some random times, the `./securedrop-admin sdconfig` command fails on validation on SMTP .

## Steps to Reproduce

.`$ ./securedrop-admin sdconfig`

## Expected Behavior

It should pass.

## Actual Behavior

```
TASK [validate : Perform SMTP lookup check.] *******************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "cmd": ["torify", "host", "-W=10", "-R=3", "-T", "-4", "smtp.gmail.com", "8.8.8.8"], "delta": "0:00:03.061367", "end": "2017-12-20 17:55:05.760958", "failed": true, "rc": 1, "start": "2017-12-20 17:55:02.699591", "stderr": "", "stderr_lines": [], "stdout": ";; connection timed out; no servers could be reached", "stdout_lines": [";; connection timed out; no servers could be reached"]}
...ignoring

TASK [validate : Validate SMTP relay connection.] **************************************************************
fatal: [localhost]: FAILED! => {
    "assertion": "'has address' in smtp_validate_result.stdout", 
    "changed": false, 
    "evaluated_to": false, 
    "failed": true, 
    "msg": "The SMTP relay domain failed during lookup. This domain is the server contacted for authentication in order to send OSSEC email notifications. You should manually edit the file `group_vars/all/site-specific` and confirm that the `smtp_relay` var is set correctly."

```

## Comments

We can remove the whole validate or something similar.

conorsch:
Marking as `ignore_errors: yes` seems the sane approach here, so we still inform the Admin, but don't abort execution of the install. However, the rewrite in #2758 means we'll need another strategy, so let's wait until that's merged to tackle this error.

redshiftzero:
If we hit this during 0.5.1 QA, I suggest we add `ignore_errors: yes` to the `Validate SMTP relay connection` task (as @conorsch suggests) for 0.5.1, as this has been reported as a source of frustration by current admins

conorsch:
I've already hit this error during QA, and encourage the stopgap solution of setting `ignore_errors: yes`. After release, we can circle back and attempt a more elegant solution that provides "warning" messages (via `debug`) to Admins.

eloquence:
Since we've disabled this check and don't see high importance of re-enabling, @conorsch @redshiftzero and I agreed to close this issue for the time being.

PR #2815, opened 2018-01-09
Stopgap for #2760: ignore errors on host checks in validate role

redshiftzero:
## Status

Ready for review

## Description of Changes

Stopgap for #2760 (intentionally not closing that issue until we understand _why_ this is an intermittent failure). 

## Testing

Run `./securedrop-admin sdconfig` and ensure the host checks do not cause a failure

## Deployment

Admins will need to check out 0.5.1 in their admin workstation to get this (when shipped)

## Checklist

Requires manual testing in Tails


conorsch:
Looks great. Will test locally in prod VMs to be absolutely sure.

msheiny:
@conorsch i made a change and tested in prod VMs so I'm not going to hit merge here unless ya'll are on the same page.

conorsch:
OK, @msheiny, tested your changes and those work, as well. Normally I'd prefer the smallest diff possible, but I can see the scary red error text being confusing to Admins, so I agree with the change. We can always cleanly revert your commit as part of properly addressing #2760.

dachary:
For the record I object to removing this check without understanding why it does not work.

redshiftzero:
Fair @dachary. Unfortunately for many users this validation erroneously reporting a failure means that they are unable to proceed with an install or update (unless they are sufficiently technical to modify the playbook themselves, which is not something most admins can do). Since this validation feature is not necessary, I think removing this from production SecureDrops until we have a validation check that works (at minimum for all SecureDrop testers) - as will be provided by your work in #2758 - is the most conservative course of action. What do you think?

conorsch:
@dachary There are now two changes in this PR:

* @redshiftzero's setting of ignore_errors=yes
* @msheiny's full removal of the known-to-be-bad failing checks, pending final resolution

Do you object to either, or both? My concern at present is that the failure rate is so high, it's preventing Admins from running the playbooks with an otherwise working config. We are intentionally leaving #2760 open to dig into the root cause, since you're right: we don't understand the _why_ yet.

But given that code freeze is a few hours away, I do not think it's realistic to understand the problem adequately at present, and instead suggest we'll need to do so as part of the 0.5.2 work. Thoughts?

dachary:
Ho, sorry, I should have clarified that I object on principle. I don't really have strong feelings otherwise I would have formally object with a negative review ;-)

Issue #2827, opened 2018-01-11
500 internal server error on instance config page due to missing AppArmor rule


redshiftzero:
## Description

I went to the new "Instance Config" page and got a 500 internal server error due to a missing AppArmor rule.

## Steps to Reproduce

0. Install `0.5.1-rc1` on prod VMs
1. Create an administrator account.
2. Click on "Instance Config" in the admin interface:

![screen shot 2018-01-10 at 6 41 26 pm](https://user-images.githubusercontent.com/7832803/34806050-e51e1f0e-f635-11e7-8261-28e8d99cc3c1.png)

## Expected Behavior

You go to the instance config page and all is merry and well. 

## Actual Behavior

![screen shot 2018-01-10 at 6 42 11 pm](https://user-images.githubusercontent.com/7832803/34806067-fd0b123e-f635-11e7-96aa-7dfb13760d6c.png)

## Comments

A peek at the logs reveals the culprit:

```
[Thu Jan 11 02:38:16.105986 2018] [:error] [pid 3754:tid 2943491983104] [remote 127.0.0.1:51806] IOError: [Errno 13] Permission denied: '/var/www/securedrop/journalist_templates/config.html'
```

Updating the AppArmor profile for Apache will fix this.

redshiftzero:
I'm surprised that CI didn't catch this, since there were unit and functional tests for this feature that will pass in the development environment due to the lack of AppArmor, but should fail in staging..

dachary:
> I'm surprised that CI didn't catch this

Ouch, it means something is broken in the CI :-(

kushaldas:
I can also verify the bug.

redshiftzero:
Closed by #2827

PR #2829, opened 2018-01-11
[Do not merge] Circle CI with intentionally failing application test

redshiftzero:
## Status

Testing only 

## Description of Changes

In #2827, it was questioned how an important missing AppArmor rule was not caught in CI given the presence of several unit and functional tests for features requiring the page AppArmor denied Apache access to. This branch has an intentionally failing test. 

## Testing

Does Circle CI pass? If so, we need to fix that.

## Deployment

-

## Checklist

-


redshiftzero:
Mystery resolved during today's engineering meeting. We need to run the functional tests via a test client so that bugs in the Apache AppArmor rules are discovered.

PR #2837, opened 2018-01-11
[0.5.1] tests: wait for source key generation in functional tests

dachary:
## Status

Ready for review

## Description of Changes

All functional tests assume the source GPG key is created when a
document or a message is submitted. But it really is generated
asynchronously and therefore races with the tests. It is fine most of
the time but leads to subtle bugs.

For instance when a journalist functional test tries to reply to a
source. If the GPG is being created the selector will not be
displayed: instead the "flagged for reply" display will show and the
test will fail to reply.

This race was incorrectly fixed by waiting for the key to be generated
before deleting it, because it happened relatively frequently.

A support function is added to wait until the GPG key for a given
source becomes available. It is called whenever a source submits a
document and effectively waits until the async_genkey completes.

## Testing

N/A (covered by CI)

## Deployment

N/A

### DIFF
--- a/securedrop/tests/functional/functional_test.py
+++ b/securedrop/tests/functional/functional_test.py
@@ -25,6 +25,7 @@ import config
 import db
 import journalist
 from source_app import create_app
+import crypto_util
 import tests.utils.env as env
 
 LOG_DIR = abspath(join(dirname(realpath(__file__)), '..', 'log'))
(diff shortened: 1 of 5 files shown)


dachary:
@redshiftzero targetting 0.5.1 because it showed up frequently when running update-user-guides

dachary:
> I wonder why we don't just use 

Yeah, it's tempting to cleanup :-)

PR #2843, opened 2018-01-12
tests: wait up to 60 seconds to generate a GPG key

dachary:
## Status

Ready for review

## Description of Changes

If the machine is slow, 5 seconds may not be enough. Since there is no entropy shortage (urandom in use), 60 seconds should be plenty.

## Testing

Use a slow machine and verify functional tests pass.

## Deployment

N/A


dachary:
merci !

PR #2863, opened 2018-01-14
tests: create $USER in Dockerfile and allow sudo

### LINKED ISSUE #2862: tests: create $USER in Dockerfile and allow sudo
# Feature request

## Description

create $USER in Dockerfile and allow sudo

## User Stories

As a developer I want to be able to **docker run -v $(pwd):$(pwd)** as myself and not root so my files do no belong to root in the end. All my files are belong to me !

dachary:
## Status

Ready for review

## Description of Changes

Fixes #2862

When run from the developer environment, it allows for docker run
--user $USER to work. Instead of running scripts as root by default,
it allows to run them using this unprivileged user and use sudo when
root privileges are necessary. It also allows to modify host files
exposed via --volume without changing their ownership.

## Testing

* make -C securedrop images
<pre>
$ docker run --user=$USER -ti securedrop-test:$(docker images -q securedrop-test |head -n1) bash
loic@b2eb0286b61d:/app$ sudo id
uid=0(root) gid=0(root) groups=0(root)
</pre>

## Deployment

N/A

### DIFF
--- a/securedrop/Dockerfile
+++ b/securedrop/Dockerfile
@@ -1,5 +1,9 @@
 # ubuntu:14.04 as of 2017-12-15
 FROM ubuntu@sha256:084989eb923bd86dbf7e706d464cf3587274a826b484f75b69468c19f8ae354c
+ARG USER_NAME
+ENV USER_NAME ${USER_NAME:-root}
+ARG USER_ID
+ENV USER_ID ${USER_ID:-0}
 
 RUN apt-get update && \
     apt-get install -y python-pip libpython2.7-dev libssl-dev secure-delete \
@@ -27,6 +31,8 @@ RUN echo deb http://archive.ubuntu.com/ubuntu/ xenial main > /etc/apt/sources.li
 COPY requirements requirements
 RUN pip install -r requirements/securedrop-app-code-requirements.txt && \
     pip install -r requirements/test-requirements.txt
+
+RUN if test $USER_NAME != root ; then useradd --no-create-home --home-dir /tmp --uid $USER_ID $USER_NAME && echo "$USER_NAME ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers ; fi
 
 WORKDIR /app
 
--- a/securedrop/Makefile
+++ b/securedrop/Makefile
@@ -12,7 +12,10 @@ lint-full:  ## Run the python linter with nothing disabled
 
 .PHONY: images
 images: ## Build the docker images
-	docker build $(EXTRA_BUILD_ARGS) -f Dockerfile -t $(IMAGE)-test:$(TAG) .
+	docker build $(EXTRA_BUILD_ARGS) -f Dockerfile \
+	   --build-arg=USER_ID=$(shell id -u) \
+	   --build-arg=USER_NAME=$(USER) \
+	   -t $(IMAGE)-test:$(TAG) .
 
 TESTFILES ?= tests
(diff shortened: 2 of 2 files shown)


dachary:
@tmc I'm not sure about $USER being MacOS compatible, do you have a better idea ?

dachary:
@msheiny first time I see this one https://circleci.com/gh/freedomofpress/securedrop/6541?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

<pre>
    RUNNING HANDLER [tor-hidden-services : restart tor (simple)] *******************
    Sunday 14 January 2018  11:34:24 +0000 (0:00:00.047)       0:04:55.279 ********
    changed: [mon-staging]
    fatal: [app-staging]: FAILED! => {"changed": false, "failed": true, "msg": " * Stopping tor daemon...\n   ...done.\n * Starting tor daemon...\n   ...fail!\n"}
    
</pre>

dachary:
See https://github.com/freedomofpress/securedrop/pull/2865#issuecomment-357544435 for a reply to https://github.com/freedomofpress/securedrop/pull/2863#pullrequestreview-88688071

tmc:
I wonder how crazy it would be to have some sort of integration test for dev environments.. circle can run osx jobs.

tmc:
We'd need to test this at least on MacOS but I'm not sure the complexity is worthwhile. Also note that dev-shell moved away from a make target.

dachary:
> We'd need to test this at least on MacOS but I'm not sure the complexity is worthwhile.

It simplifies the implementation of https://github.com/freedomofpress/securedrop/pull/2865 and all similar commands (for i18n for instance, make translate-message, make translate-desktop, make translate-compile etc.) that we will implement and that have a side effect on the current working directory.

dachary:
@redshiftzero would you mind testing on MacOS ?

dachary:
> Also note that dev-shell moved away from a make target.

Right ! Updated the tests accordingly.

dachary:
@heartsucker @squeed @tmc here is a quick summary of the discussions that happened over the past week and there conclusions:

* it was suggested to not add a user but sudo needs one, setting the numeric uid is not enough and sudo will fail
* it was suggested to move the logic to create a user to the ENTRYPOINT but it does not work because it  relies on sudo which requires a user
* it was suggested to run all root commands in a dedicated image so the script can run without ever using sudo but that does not work to run daemons daemons

It would be great if you could revisit this PR with these conclusions in mind and let me know if you have any objection to this PR

Thanks a lot for your diligent reviews on this tiny PR !

heartsucker:
Can you not start the container as root, create the user, and have it sudo to that user?

dachary:
> Can you not start the container as root, create the user, and have it sudo to that user?

It is possible to go this way and I tried it (it also is a nightmare because of shell evaluation) but I fail to see the advantage of doing so. It's more complicated to use because you then need to:

    docker run --env=USER_ID=$(id -u) --env=USER_NAME=$(USER) ...

instead of 

    docker run --user=$(USER) ..

redshiftzero:
Tested this PR on macOS, works as advertised for me: i.e. `id -u` and `$USER` both work in macOS ✅

redshiftzero:
Hey @heartsucker are you happy with the changes here?

heartsucker:
Merging because this is the best solution, but grumbling because it still feels wrong.

PR #2865, opened 2018-01-14
docs: refactor make update-user-guides to run as non-root

### LINKED ISSUE #2864: docs: make update-user-guides must not run as root
# Feature request

## Description

make update-user-guides must run as non-root to avoid creating files owned by root in the git repository as a side effect

## User Stories

As a developer I want to be able to make update-user-guides and not worry about files being owned by root as a side effect

dachary:
## Status

Ready for review

## Description of Changes

Depends #2862
Fixes #2864 

When running update-user-guides it creates files that are on the host,
to update screenshots. The script runs as root in the docker
container and these files must be chown to developer user to avoid
permission problems. But a number of other files can also be created
as root as a side effect (pyc, etc.) and ensuring they all belong to
the user is error prone.

Instead of running the script as root, run it as the develper user and
rely on sudo for actions that require root permissions in the
container.

A VNC proxy is added to enable the developer to run

    xtightvncviewer localhost:1

and see the X11 display used by the tests, for debugging purposes.

## Testing

* make -C securedrop images update-user-guides
* (while it is running) xtightvncviewer localhost:1

## Deployment

N/A

### DIFF
--- a/securedrop/Makefile
+++ b/securedrop/Makefile
@@ -54,9 +54,10 @@ update-user-guides: ## Runs the page layout tests to regenerate screenshots
 	docker run \
 		--name $(IMAGE)-update-user-guides \
 		--rm \
-		-e USER_ID=$(shell id -u) \
-		-v $(shell git rev-parse --show-toplevel):$(shell git rev-parse --show-toplevel) \
-		-w $(shell pwd) \
+		--user $(USER) \
+		--volume $(shell git rev-parse --show-toplevel):$(shell git rev-parse --show-toplevel) \
+		--workdir $(shell pwd) \
+		--publish 5901:5901 \
 		-ti $(IMAGE)-test:$(TAG) \
 		./bin/update-user-guides
 
--- a/securedrop/bin/update-user-guides
+++ b/securedrop/bin/update-user-guides
@@ -2,11 +2,19 @@
 
 set -e
 export DISPLAY=:1
-Xvfb :1 -screen 0 1024x768x24 -ac +extension GLX +render -noreset &
-redis-server &
 
 rm /dev/random
 ln -s /dev/urandom /dev/random
+function run_xvfb() {
+    Xvfb :1 -screen 0 1024x768x24 -ac +extension GLX +render -noreset >& /tmp/xvfb.out || cat /tmp/xvfb.out
+}
+
+function run_redis() {
+    redis-server >& /tmp/redis.out || cat /tmp/redis.out
+}
+
+run_xvfb &
+run_redis &
 
 if ! test -f config.py ; then
     trap "rm config.py" EXIT
--- a/securedrop/bin/update-user-guides
+++ b/securedrop/bin/update-user-guides
@@ -13,8 +13,13 @@ function run_redis() {
     redis-server >& /tmp/redis.out || cat /tmp/redis.out
 }
 
+function run_x11vnc() {
+    x11vnc -display :1 -autoport 5901 -shared >& /tmp/x11vnc.out || cat /tmp/x11vnc.out
+}
+
 run_xvfb &
 run_redis &
+run_x11vnc &
 
 if ! test -f config.py ; then
     trap "rm config.py" EXIT
--- a/securedrop/bin/update-user-guides
+++ b/securedrop/bin/update-user-guides
@@ -3,8 +3,6 @@
 set -e
 export DISPLAY=:1
 
-rm /dev/random
-ln -s /dev/urandom /dev/random
 function run_xvfb() {
     Xvfb :1 -screen 0 1024x768x24 -ac +extension GLX +render -noreset >& /tmp/xvfb.out || cat /tmp/xvfb.out
 }
@@ -21,14 +19,15 @@ run_xvfb &
 run_redis &
 run_x11vnc &
 
+sudo rm /dev/random
+sudo ln -s /dev/urandom /dev/random
+
 if ! test -f config.py ; then
     trap "rm config.py" EXIT
     make test-config
 fi
 
 sass --force --stop-on-error --update sass:static/css --style compressed
-chown -R "$USER_ID" static .sass-cache
 
 pytest -v tests/pages-layout --page-layout
 cp tests/pages-layout/screenshots/en_US/*.png ../docs/images/manual/screenshots
-chown -R "$USER_ID" tests/pages-layout/screenshots ../docs/images/manual/screenshots ./*/__pycache__ ./*/*/__pycache__ || true


dachary:
> $USER should probably be $(id -u) for best compatibility. 

This is for the user name which is required when creating the user. The user id is taken from $(id -u) indeed.

> But also this feels very un-docker-y in that it's breaking the isolation between host and container. 

The idea here is that the container provides all the dependencies to run a single action (create screenshots). It is indeed not meant to provide any isolation, the idea is to produce a side effect on the host without the need to install all dependencies on the host itself.

> I've also run into strange problems with containers running as users that are mapped to the host OS. Also, some containers don't support sudo very well.

I'm not aware of such problem, do you have any pointers ? I've been using docker in this way for years but it's entirely possible I missed the associated problems.

> Basically, I feel like we shouldn't be doing this at all since it's just saving some minor developer annoyance.

I think it's worth saving a minor developer annoyance. But there is more: a number of commands (sass for instance or ./manage.py translate-desktop etc.) have the same pattern.

securedrop/Dockerfile
```diff
@@ -29,6 +31,8 @@ COPY requirements requirements
 RUN pip install -r requirements/securedrop-app-code-requirements.txt && \
     pip install -r requirements/test-requirements.txt
 
+RUN if test $USER_NAME != root ; then useradd --no-create-home --home-dir /tmp --uid $USER_ID $USER_NAME && echo "$USER_NAME ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers ; fi
```
heartsucker: You need to export the `ARG` value with `ENV`.

https://docs.docker.com/engine/reference/builder/#using-arg-variables

heartsucker:
@dachary  Replying to what you're saying about there being multiple places that use this pattern. I think it's entirely possible to extend the docker workflow to get us to a point where we don't need to worry about this. I suspect we'll eventually get rid of the dev VM and only need `vagrant` for staging/prod testing. I feel like this is shoehorning multiple workflows together. And really my biggest objection is that having to add a user that's an argument to `sudoers` just *feels wrong*.

dachary:
> I think it's entirely possible to extend the docker workflow to get us to a point where we don't need to worry about this. 

I'm interested :-) How would that work ?

dachary:
@heartsucker ping ?

squeed:
I'm not sure why the `useradd` is there - did you try without it? Docker should create the various `/etc/passwd` entries.

squeed:
@heartsucker is generally right - doing and `useadd` is sort of un-docker-y. Ideal would be the ability to run almost nothing as root, especially I don't, as a rule, trust Docker isolation.

This will require a bit of a more careful redesign; and generally involves setting up a `securedrop-builder` container that is run "once" as root. All subsequent commands are then run as your own user.

dachary:
> I'm not sure why the useradd is there - did you try without it? Docker should create the various /etc/passwd entries.

Yes

```
loic@raise:~$ docker run ubuntu:14.04 grep $USER /etc/passwd
loic@raise:~$ docker run ubuntu:14.04 grep root /etc/passwd
root:x:0:0:root:/root:/bin/bash
loic@raise:~$ 
```

dachary:
> This will require a bit of a more careful redesign; and generally involves setting up a securedrop-builder container that is run "once" as root. All subsequent commands are then run as your own user.

Could you explain more ? That sounds interesting ! Also I'd be curious to get an explanation about the why user is "undockery".

squeed:
The name of the user doesn't matter, only their numeric ID, especially if you're concerned about having random 'root' files pooped all over the bind-mounted directory. So the right thing is generally to do:

```
$ docker run --user $(id -u):$(id -g) alpine id
uid=1000 gid=1000
```

dachary:
> docker run --user $(id -u):$(id -g) alpine id

The goal is to run scripts under ``id -u`` by default, otherwise they create root files and that is not good. But we also want these scripts to be able to sudo when appropriate and this is where adding an entry in /etc/passwd is relevant.

securedrop/Dockerfile
dachary: done

dachary:
@heartsucker I added ENV in addition to ARG as suggested.

dachary:
@heartsucker regarding your suggestion to add an entrypoint that would run as root to add the users instead of doing that at build time, it turns out to complicate things. If the container runs with --user it will not have the ability to create the user and fail. If it runs as root, the script will need to explicitly change to running under the desired id (relying on variables maybe ?) and this defeats the purpose which is to have scripts run as the current user by default and sudo when needed. It is also possible that I'm missing a simplification that you're seeing, in which case I'd be grateful if you could explain in more details.

heartsucker:
I was imagining something like this:

```bash
docker run -e USER_ID=$(id -u)  \
    <other env flags> \
    -u $(id -u):$(id -g) \
    -w /src \
    -v $(pwd):/src/ \
    securedrop/docs-maker
```

in the `Dockerfile`:

```
RUN echo 'ALL ALL = (ALL) NOPASSWD: ALL' > /tmp/sudoer && visudo -f /tmp/sudoers
```
And the entrypoint would be

```bash
#!/bin/bash
set -eu

# maybe add some if/else checking to make `docker exec` work
sudo groupadd <args>
sudo useradd <args>

if [ -z "$1" ]; then
    # do build
else
  $@
end
```

heartsucker:
I see @squeed commented too. Ignore what I just said. I have either had too much or not enough coffee.

dachary:
That won't work because sudo requires a name. Unless there is a way to dispense of that ?

<pre>
loic@raise:/tmp/l$ docker build --tag t .
Sending build context to Docker daemon  2.048kB
Step 1/2 : FROM ubuntu:14.04
 ---> 67759a80360c
Step 2/2 : RUN echo 'ALL ALL = (ALL) NOPASSWD: ALL' >> /etc/sudoers
 ---> Running in 35b0c769028f
 ---> 3c2328b8d3f1
Removing intermediate container 35b0c769028f
Successfully built 3c2328b8d3f1
Successfully tagged t:latest
loic@raise:/tmp/l$ docker run -ti t bash
root@d47dde36ca11:/# exit
loic@raise:/tmp/l$ docker run -ti --user $(id -u) t bash
I have no name!@745457854bac:/$ sudo bash
sudo: unknown uid 1000: who are you?
</pre>

dachary:
@heartsucker were your concerns addressed ?

dachary:
@heartsucker do you mean setting a password or suppressing the warning because it is frightening ?

heartsucker:
That container has no network access or exposed ports, so it's not really a problem. It just dumps a ton of log messages to stdout that aren't helpful or useful to see. If we just write to `&> /dev/null`, then we'd lose the ability to debug CI build errors. So it might be useful to set the password.

heartsucker:
But also if you just silenced all the crap it outputs, I'd be happy with that.

dachary:
@heartsucker excellent idea ! I silenced the output of all daemons and display them only if they fail. How does that look now ?

Issue #2866, opened 2018-01-14
Move custom SQLAlchemy session management to Flask-SQLAlchemy


heartsucker:
# Feature request

## Description

We maintain our own code to handle the database operations, and it would make more sense to use someone else's code. This will also make usage of Alembic cleaner for #1419.

## User Stories

As a developer, I don't want to maintain custom code just to keep SecureDrop running.

heartsucker:
Work is tracked in branch [`flask-sqlalchemy`](https://github.com/freedomofpress/securedrop/tree/flask-sqlalchemy).

PR #2869, opened 2018-01-14
db: enable vacuum and secure delete

### LINKED ISSUE #2868: Enable vacuum and secure delete on sqlite
It would be valuable to enable secure delete and auto vacuum on the database when available.

Even if the database of securedrop holds very minimal metadata information this would reduce leakage of some metadata about the names of files and the number of submissios or past users enabled on the platform.

For sqlite this could be achieved by issuing the following queries:
```sql
PRAGMA secure_delete = ON
PRAGMA auto_vacuum = FULL
```

the two can be issued at the time of database creation and would garantee that entries are securely deleted from the database and not just marked as free.

In SQLAlchemy this can be achieved with:
```python
    engine.execute('PRAGMA secure_delete = ON')
    engine.execute('PRAGMA auto_vacuum = FULL')
```

a similar result could be probably achieved on other databases

references:
- https://www.sqlite.org/pragma.html#pragma_secure_delete
- https://www.sqlite.org/pragma.html#pragma_auto_vacuum

dachary:
## Status

Ready for review

## Description of Changes

Fixes #2868

Note that auto_vacuum will only be effective on newly created
databases because it needs to be done before any table are created.

* https://www.sqlite.org/pragma.html#pragma_secure_delete
* https://www.sqlite.org/pragma.html#pragma_auto_vacuum

## Testing

* vagrant ssh development
* rm -f /var/lib/securedrop/db.sqlite
* ./manage.py run &
* sqlite3 /var/lib/securedrop/db.sqlite 
<pre>
SQLite version 3.8.2 2013-12-06 14:53:30
Enter ".help" for instructions
Enter SQL statements terminated with a ";"
sqlite> pragma main.auto_vacuum;
1
sqlite> pragma main.secure_delete;
1
</pre>

## Deployment

Runtime options only.

### DIFF
--- a/securedrop/db.py
+++ b/securedrop/db.py
@@ -41,6 +41,9 @@ if config.DATABASE_ENGINE == "sqlite":
         config.DATABASE_ENGINE + ":///" +
         config.DATABASE_FILE
     )
+
+    engine.execute('PRAGMA secure_delete = ON')
+    engine.execute('PRAGMA auto_vacuum = FULL')
 else:  # pragma: no cover
     engine = create_engine(
         config.DATABASE_ENGINE + '://' +


securedrop/db.py
```diff
@@ -41,6 +41,8 @@
         config.DATABASE_ENGINE + ":///" +
         config.DATABASE_FILE
     )
+    engine.execute('PRAGMA secure_delete = ON')
```
heartsucker: This should be moved to `{journalist,source}_app/__init__.py`. This will behave badly when we complete #2866.

heartsucker: And it will probably need to be:

```python
with app.app_context():
    # do stuff
```

dachary: I don't see how it can be moved to ``{journalist,source}_app/__init__.py`` and it would be great if you could provide a snippet to show me the way. Thanks :-)

heartsucker: This *should* work:

```python
from sqlalchemy import text

def create_app(config):
    app = Flask(__name__,

app.config.from_object(config.JournalistInterfaceFlaskConfig)

with app.app_context():
    db_session.execute(text('PRAGMA secure_delete = ON''))
    db_session.exexute(text('PRAGMA auto_vacuum = FULL'))
    db_session.commit()
```

In fact, the `with app.app_context()` might not even be necessary since we have some no wrapper libs. Anyway, calling it directly from the app's init function should work.

dachary:
@heartsucker I updated the PR description with instructions to verify. And a note about the fact that auto_vacuum is only effective for newly created databases.

dachary:
tests fail with an error that looks like a race of some kind https://circleci.com/gh/freedomofpress/securedrop/6614?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link
<pre>
self = <sqlalchemy.dialects.sqlite.pysqlite.SQLiteDialect_pysqlite object at 0x7f5d10e06d10>
cargs = ('/tmp/securedrop/db.sqlite',), cparams = {}

    def connect(self, *cargs, **cparams):
>       return self.dbapi.connect(*cargs, **cparams)
E       OperationalError: (sqlite3.OperationalError) unable to open database file (Background on this error at: http://sqlalche.me/e/e3q8)

/usr/local/lib/python2.7/dist-packages/sqlalchemy/engine/default.py:410: OperationalError
</pre>

When the execute statements run from db.py this does not happen. I suppose it has to do with the fact that test_journalist.py and test_source.py do not use create_app yet. I pushed a commit with execute at the top level of db.py.

Any suggestion to properly fix this would be most welcome !

dachary:
@heartsucker ping ?

heartsucker:
> test_journalist.py and test_source.py do not use create_app yet

This is covered in #2895. Also, as noted in a previous version, this should be done in `{source,journalist}_app/__init__.py`. If we wait for #2895 to land, we can do it that way. However, as this stands there seems to be nothing wrong with it.

PR #2870, opened 2018-01-15
Edits to landing page docs

pierwill:
## Status

Ready for review

## Description of Changes

Various typo, style, and other fixes to landing page docs

Removed broken/outdated link from landing page docs

### DIFF
--- a/docs/deployment/landing_page.rst
+++ b/docs/deployment/landing_page.rst
@@ -15,7 +15,7 @@ your organization.
           implement minimum security requirements is sure to be noticed, and
           could undermine trust, discouraging possible sources.
 
-URL and location
+URL and Location
 ----------------
 
 Ideally you would not use a separate subdomain, but would use a path at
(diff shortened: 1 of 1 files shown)


b-meson:
I think this is ready for merge but I am not sure why [this line](https://github.com/freedomofpress/securedrop/pull/2870/files#diff-dfa4d01609c432bc4d7bccf86132f3f5R284) is present. Can you elaborate @pierwill

pierwill:
I think something should demarcate the introduction to this block of suggested text. Otherwise, there is no visual indication for the reader of the different between doc text and example text. I didn't think `literal` was the way to go. A [horizontal rule](https://gist.github.com/dupuy/1855764#horizontal-rules-transitions) seemed right to me.

pierwill:
horiz

heartsucker:
Agreeing with @bemeson. Since most likely these landing pages are set up on a corporate networks, something like `fail2ban` isn't going to solve the issue. They page is probably going to be accessible from the CMS that editors use to maintain the site. We shouldn't really start advocating for corporate security practices since that's and endless rabbit hole.

heartsucker:
I'm with @b-meson. We should drop that line. Most likely the landing pages are hosted on the CMS in a corporate network environment where they wouldn't even use `fail2ban`, but probably something like Kerberos. We shouldn't even try to dictate corporate IT policy other than the few basics about good headers and whatnot since that's an endless rabbit hole.

redshiftzero:
Ah, I think @b-meson was asking about the "Landing Page Content Suggestions" line - @pierwill's explanation makes sense there. In terms of the security requirements, while it might be a big topic, it is an important aspect of a SecureDrop deployment (since if an attacker can easily compromise the webserver hosting the landing page then they can trivially replace the onion address of the legitimate SecureDrop with the onion address of an attacker-controlled SecureDrop - if sources do not verify the onion address against the SecureDrop directory, or if the SecureDrop instance is not in the SecureDrop directory, this would not be detected by the source). I agree with @heartsucker that we shouldn't _dictate_, only suggest. That said, if we can point to more resources in this area for smaller organizations, they would find it valuable.

PR #2895, opened 2018-01-19
Added flask-sqlalchemy to replace homegrown session management

### LINKED ISSUE #2866: Move custom SQLAlchemy session management to Flask-SQLAlchemy

heartsucker:
## Status

Work in progress (need to discuss how we update existing instances)

## Description of Changes

Fixes #2866

## Testing

No special steps.

## Deployment

We will need to add the requirements file to the `.deb` and do a `pip install` to cover existing instances. This is already done in the `postinst` step, so no issues there.


redshiftzero:
This looks great so far @heartsucker - also I don't see any particular issues with the upgrade path here for current instances - this PR just modifies how the app code connects to the database, which should be fine for existing databases, no? (Don't forget to modify the `Initialize sqlite database.` task in `install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml` to use `db.create_all()`)

heartsucker:
> Don't forget to modify the Initialize sqlite database. task in install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml to use db.create_all()

Ah good catch on that. Will do.

Currently stuck debugging errors because the page layout tests behave differently on vagrant and CI. D:

heartsucker:
Note: this was rebased on top of #2905, so that one will need to be merged first.

heartsucker:
There is an error with key generation it seems since [this error](https://circleci.com/gh/freedomofpress/securedrop/7023?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) that is not present locally. Most likely the patch on entropy checking fails.

dachary:
@heartsucker it is unlikely the entropy patch not working (unless something in this commit series has that side effect) because it reliably works on develop. It can also be caused when the key generation fails for other reasons, not necessarily entropy. A good way to get a better error message is to comment out the async_gen wrapper so it does not silently happen in the background.

securedrop/tests/utils/db_helper.py
```diff
@@ -81,31 +83,31 @@ def mock_verify_token(testcase):
...
     """
     # Create source identity and database record
     codename = crypto_util.genrandomid()
     filesystem_id = crypto_util.hash_codename(codename)
     journalist_filename = crypto_util.display_id()
-    source = db.Source(filesystem_id, journalist_filename)
-    db.db_session.add(source)
-    db.db_session.commit()
+    source = models.Source(filesystem_id, journalist_filename)
+    models.db.session.add(source)
```
heartsucker: Annoyingly, this should just be `db.session` and not `models.db.sessions`. End result of a `sed` gone bad. Technically, this isn't wrong, but it's not really worth changing unless there's something else too.

securedrop/journalist_app/__init__.py
```diff
@@ -28,6 +29,21 @@ def create_app(config):
...
+        db_uri = (config.DATABASE_ENGINE + ":///" +
+                  config.DATABASE_FILE)
+    else:
+        db_uri = (
+            config.DATABASE_ENGINE + '://' +
+            config.DATABASE_USERNAME + ':' +
+            config.DATABASE_PASSWORD + '@' +
+            config.DATABASE_HOST + '/' +
+            config.DATABASE_NAME
+        )
```
redshiftzero: We can definitely get rid of the `else` part since everyone is using SQLite - also we've recently merged in raw SQL statements that are SQLite specific. Not required for merge since your diff here is just moving the `db_uri` logic unchanged into `create_app`, just noting

securedrop/manage.py
```diff
@@ -87,7 +87,7 @@ def reset(args):
         pass
 
     # Regenerate the database
-    init_db()
+    db.creat_all()
```
redshiftzero: Typo: should be `db.create_all()`

securedrop/models.py
```diff
@@ -73,7 +47,7 @@ def get_one_or_else(query, logger, failure_method):
         failure_method(404)
 
 
-class Source(Base):
```
redshiftzero: Hm I'm not sure we want to move away from `declarative_base` - we'll want to preserve that in order to [enable autogenerated migrations for alembic](http://alembic.zzzcomputing.com/en/latest/autogenerate.html#auto-generating-migrations), which we definitely want for developer agility

securedrop/manage.py
redshiftzero: This was resolved in the latest push, disregard

securedrop/models.py
heartsucker: I have used this pattern to generate at least one auto migration, so I'm fairly sure this works. If it doesn't, I can swap it out when I add alembic.

heartsucker: From `help(models)`:

```
    class Journalist(sqlalchemy.ext.declarative.api.Model)
     |  Method resolution order:
     |      Journalist
     |      sqlalchemy.ext.declarative.api.Model
     |      flask_sqlalchemy.model.Model
     |      __builtin__.object
```

So it looks like it still is `declarative_base` under the hood too.

redshiftzero: Ahh right, so we can use `db.metadata` instead of using `Base.metadata` as the target metadata for migrations. Sounds good!

install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml
```diff
@@ -152,12 +152,9 @@
...
 - name: Initialize sqlite database.
-  shell: >
-    su -s /bin/bash -c
-    "PYTHONPATH={{ securedrop_code }}
-    python -c 'import db; db.init_db()'"
-    {{ securedrop_user }}
-  when: not db.stat.exists
+  shell: './manage.py init-db --user {{ securedrop_user }}'
+  args:
+    chdir: '{{ securedrop_code }}'
```
redshiftzero: Just noticed on this task we lost `when: not db.stat.exists` which we want to preserve for idempotency, so let's add that back in

securedrop/tests/test_integration.py
```diff
@@ -17,223 +17,236 @@
...
         self.gpg = gnupg.GPG(homedir=config.GPG_KEY_DIR)
 
         # Patch the two-factor verification to avoid intermittent errors
-        patcher = mock.patch('db.Journalist.verify_token')
+        patcher = mock.patch('models.Journalist.verify_token')
         self.addCleanup(patcher.stop)
         self.mock_journalist_verify_token = patcher.start()
         self.mock_journalist_verify_token.return_value = True
 
         # Add a test user to the journalist interface and log them in
```
redshiftzero: Nit: comment is now incorrect: `# Add a test user to the journalist interface and log them in` -> `# Add a test user to the journalist interface`

securedrop/tests/test_integration.py
```diff
@@ -17,223 +17,236 @@
...
         # print Journalist.query.all()
         self.user_pw = "corret horse battery staple haha cultural reference"
-        self.user = Journalist(username="some-username",
-                               password=self.user_pw)
-        db_session.add(self.user)
-        db_session.commit()
-        self._login_user()
+        self.username = crypto_util.genrandomid()
```
redshiftzero: I'm fine with this, but if we do then we should specify the random seed for deterministic unit tests runs (see [here](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/tests/test_journalist.py#L25))

heartsucker:
@redshiftzero *pew pew*

Issue #2510, opened 2017-11-04
Tor warning not working on mobile 


redshiftzero:
# Description

Reported during Mozilla Festival 2017 session: user using SecureDrop + Orfox on their Android phone saw the warning on the source interface about not using Tor. They inquired thinking something might be wrong. We should fix this. 

We should also consider adding a warning for users on mobile indicating that for high security submissions they should use a regular computer due to the better mitigations for browser fingerprinting (if you know more about the browser fingerprinting mitigations on mobile Tor browsers, feel free to comment). 

# User Stories

As a SecureDrop source, I don't want to receive spurious warnings.

rr0ll:
Verified this on Android 4.4.4 with latest version of Orbot/Orfox and http://source.demo.securedrop.club

The message is coming from file https://github.com/freedomofpress/securedrop/blob/6e4ae446f86af0c40da332572d9c3fece6eb5c8c/securedrop/tests/functional/test_source_warnings.py

trying to locate where the test actually happens

huertanix:
Was just about to write up an issue about mobile-specific warning but it looks like that's already here.

Having talked with some folks, I get the general impression that people see us as asking whistleblowers to use TAILS (there's some verbiage about that being ideal here and there, even though we only explicitly require desktop Tor browser). Hence, people have been considering using mobile Tor (Orfox + Orbot) as what they see as the easier option, without fully understanding the related risks, and if those matter in many cases; e.g. screen sizes would be an important factor against an adversarial website but maybe not a SecureDrop instance?

redshiftzero:
Good points @huertanix. After discussing this at the Tor meeting today, since Orfox has not been audited and does not have all the privacy protections as the official Tor Browser, it should not be used for high security applications. So the way to go for this ticket is a customized warning for the Orfox mobile UA that lets sources know that official Tor Browser is the safest option, and encourage people to use it, i.e. by adopting text similar to what we use for the security slider  i.e. "We recommend using the Desktop-based Tor Browser to protect your anonymity".

eloquence:
Split into new issue per above and added to near-term backlog, we should be able to handle this soon, and at that point we can close both these issues.

legoktm:
Fixed in #3215 and again more recently after Orfox was replaced by Tor Browser for Android.

Issue #2406, opened 2017-10-06
Add QUnit tests for JavaScript


redshiftzero:
## Description

While we don't have much JavaScript, we should still have tests for it, e.g. using QUnit: https://qunitjs.com/

## User Stories

As a SecureDrop contributor, I want to know I'm not breaking the little JavaScript that exists when I modify it.

zenmonkeykstop:
Future versions of SecureDrop may use JS more extensively (for example for client-side encryption), so this issue is still relevant.

PR #2921, opened 2018-01-24
Add coverage of js functionality with Selenium and resolve js alert regression

### LINKED ISSUE #2405: Complete coverage of JS functionality with Selenium
# Feature request

## Description

As the title says, all JS functionality needs to be tested with Selenium. This is a blocker for #1298 and #1233.

### LINKED ISSUE #2920: User confirmation before 2FA reset no longer appears
## Description

There used to be a js confirmation before resetting 2FA tokens for users but this is no longer working

## Steps to Reproduce

1. Sign into journalist interface as admin
2. Go to admin panel and edit a user
3. On the edit user page, click the button to reset the user's 2FA authentication

## Expected Behavior

An alert appears, requesting the user confirm they wish to reset the 2FA token

## Actual Behavior

No alert appears / no confirmation is required

## Comments

Discovered during writing test cases for #2405 - there's a regression in `journalist.js`, this functionality no longer works:

```
  // Confirm before resetting two-factor authentication on edit user page
  $('form#reset-two-factor').submit(function(event) {
      return confirm(get_string("reset-user-mfa-confirm-string").supplant({ username: username }));
  });
```

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #2405 and fixes #2920 (happy to split this up into multiple PRs if desired)

  * Adds test coverage for js functionality using Selenium -- this should test everything that we can test with _Firefox_ based Selenium tests

We would either need to use Selenium with multiple browsers (i.e. Tor Browser and not Tor Browser) to test the remaining source interface functionality and/or have some js unit tests that [check that the right warnings are rendered based on various user agent strings](https://github.com/freedomofpress/securedrop/compare/test-coverage-source-interface-js#diff-2a4e34a719a23a3734bf355b523558beR13) (#2406). Imho adding js unit tests is easier than rigging up functional tests with many browsers (consider how we'd add test coverage for [resolutions](https://github.com/freedomofpress/securedrop/issues/2510#issuecomment-342007810) to defects like #2510). 

## Testing

Do tests pass? Is there functionality in the js that I have _not_ covered and could via Selenium? 

## Deployment

Just tests and js

### DIFF
--- a/securedrop/tests/functional/journalist_navigation_steps.py
+++ b/securedrop/tests/functional/journalist_navigation_steps.py
@@ -6,6 +6,7 @@ import gzip
 import os
 
 from selenium.common.exceptions import NoSuchElementException
+from selenium.webdriver.common.keys import Keys
 
 import tests.utils.db_helper as db_helper
 import crypto_util
@@ -78,17 +79,60 @@ class JournalistNavigationStepsMixin():
         self.driver.find_element_by_id('delete-selected').click()
         self._alert_wait()
 
-    def _journalist_verifies_deletion_of_one_submission_javascript(self):
+    def _journalist_clicks_delete_collections_javascript(self):
+        self.driver.find_element_by_id('delete-collections').click()
+        self._alert_wait()
+
+    def _journalist_clicks_delete_collection_javascript(self):
+        self.driver.find_element_by_id('delete-collection-button').click()
+        self._alert_wait()
+
+    def _journalist_uses_delete_selected_button_javascript(self):
         self._journalist_selects_first_doc()
         self._journalist_clicks_delete_selected_javascript()
         self._alert_dismiss()
+
         selected_count = len(self.driver.find_elements_by_name(
             'doc_names_selected'))
         assert selected_count > 0
+
         self._journalist_clicks_delete_selected_javascript()
         self._alert_accept()
         assert selected_count > len(self.driver.find_elements_by_name(
             'doc_names_selected'))
+
+    def _journalist_uses_delete_collection_button_javascript(self):
+        self._journalist_clicks_delete_collection_javascript()
+        self._alert_dismiss()
+
+        # After deletion the button will redirect us. Let's ensure we still
+        # see the delete collection button.
+        assert self.driver.find_element_by_id(
+            'delete-collection-button').is_displayed()
+
+        self._journalist_clicks_delete_collection_javascript()
+        self._alert_accept()
+
+        # Now we should be redirected to the index.
+        if not hasattr(self, 'accept-languages'):
+            assert "Sources" in self.driver.find_element_by_tag_name('h1').text
+
+    def _journalist_uses_delete_collections_button_javascript(self):
+        self.driver.find_element_by_id('select_all').click()
+
+        self._journalist_clicks_delete_collections_javascript()
+        self._alert_dismiss()
+
+        self.driver.find_element_by_id('select_all').click()
+        sources = self.driver.find_elements_by_class_name("code-name")
+        assert len(sources) > 0
+
+        self._journalist_clicks_delete_collections_javascript()
+        self._alert_accept()
+
+        # We should be redirected to the index without those boxes selected.
+        sources = self.driver.find_elements_by_class_name("code-name")
+        assert len(sources) == 0
 
     @screenshots
     def _admin_logs_in(self):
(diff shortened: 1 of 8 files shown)


securedrop/tests/functional/test_admin_interface.py
```diff
@@ -15,6 +15,12 @@ def test_admin_interface(self):
         self._new_user_can_log_in()
         self._admin_can_edit_new_user()
 
+    def test_admin_deletes_user(self):
+        self._admin_logs_in()
+        self._admin_visits_admin_interface()
+        self._admin_adds_a_user()
+        self._admin_can_delete_user()
```
dachary: Nit: s/_admin_can_delete_user/_admin_deletes_user/

securedrop/tests/functional/test_admin_interface.py
```diff
@@ -14,6 +14,10 @@ def test_admin_interface(self):
         self._admin_adds_a_user()
         self._new_user_can_log_in()
         self._admin_can_edit_new_user()
+        self._admin_visits_admin_interface()
```
dachary: What about having a separate function for this ? Similar to what there is in page-layout

```
    def test_admin_edit_hotp_secret(self):
        self._admin_logs_in()
        self._admin_visits_admin_interface()
        self._admin_adds_a_user()
        self._admin_visits_edit_user()
        self._admin_visits_reset_2fa_hotp()
```

securedrop/tests/functional/journalist_navigation_steps.py
```diff
@@ -78,18 +79,60 @@ def _journalist_clicks_delete_selected_javascript(self):
...
+
+        # After deletion the button will redirect us. Let's ensure we still
+        # see the delete collection button.
+        assert self.driver.find_element_by_id(
+            'delete-collection-button').is_displayed()
+
+        self._journalist_clicks_delete_collection_javascript()
+        self._alert_accept()
+
+        # Now we should be redirected to the index.
```
dachary: ``        if not hasattr(self, 'accept_languages'):
``

securedrop/tests/functional/journalist_navigation_steps.py
```diff
@@ -648,3 +691,36 @@ def _admin_enters_journalist_account_details_hotp(self, username,
...
             'input[name="is_hotp"]')
         hotp_checkbox.click()
+
+    def _journalist_uses_js_filter_by_sources(self):
+        self.wait_for(lambda: self.driver.find_element_by_id("filter"))
+
+        filter_box = self.driver.find_element_by_id("filter")
+        filter_box.send_keys("thiswordisnotinthewordlist")
+
+        sources = self.driver.find_elements_by_class_name("code-name")
```
dachary: Nit: ``assert len(sources) > 0``

securedrop/tests/functional/journalist_navigation_steps.py
```diff
@@ -648,3 +691,36 @@ def _admin_enters_journalist_account_details_hotp(self, username,
...
+
+        filter_box.clear()
+        filter_box.send_keys(Keys.RETURN)
+
+        for source in sources:
+            assert source.is_displayed() is True
+
+    def _journalist_uses_js_buttons_to_download_unread(self):
+        self.driver.find_element_by_id('select_all').click()
+        checkboxes = self.driver.find_elements_by_name('doc_names_selected')
```
dachary: Nit: ``assert len(checkboxes) > 0``

redshiftzero:
Thanks for your review @dachary :) made these changes, squashed, repushed

securedrop/tests/functional/test_source_warnings.py
```diff
@@ -14,3 +14,11 @@ def test_warning_appears_if_tor_browser_not_in_use(self):
...
         assert ("We recommend using Tor Browser to access SecureDrop" in
                 warning_banner.text)
+
+        # User should be able to dismiss the warning
+        warning_dismiss_button = self.driver.find_element_by_id(
+            'use-tor-browser-close')
+        warning_dismiss_button.click()
+
+        # The banner should now be hidden
+        self.wait_for(lambda: warning_banner.is_displayed() is False)
```
dachary: Note that:

        self.wait_for(lambda: True is False)

will pass the test. The **wait_for** function expects the function to raise an exception if it should wait some more. If it does not raise an exception, it will not wait even if the function returns False.

redshiftzero: doh, added an assert

PR #2922, opened 2018-01-24
Uses `max_fail_percentage` to force immediate Ansible exits

### LINKED ISSUE #2885: Failures on a single host should halt Ansible execution
# Bug

## Description

Right now, when running `./securedrop-admin install`, if an error is encountered on a single host, e.g. the Monitor Server, that host will drop off the play, but the playbook will continue to execute against the Application Server. That's very confusing, and can make reporting problems difficult: lots of scrollback is necessary even to determine that a host did indeed fail.

By way of example, the OSSEC registration bug documented in #2478 will lead to the only-one-host-failed scenario described above.

## Steps to Reproduce

Quite difficult to reproduce reliably. As mentioned above, #2748 is a fine candidate for observing the poor failure mode. 

## Expected Behavior
When encountering an error, Ansible stops executing, leaving the error text prominently displayed. The Admin can then use that error message to debug the underlying issue, or report the problem to the SecureDrop developers for additional support.

## Actual Behavior
If only one host fails, Ansible continues to execute against the remaining host, obscuring the fact that the error occurred at all, and complicating the procedure for sharing critical feedback required for support, such as verbatim error messages.

## Comments

We have a recourse here. There's an Ansible option called [max_fail_percentage](http://docs.ansible.com/ansible/latest/playbooks_delegation.html#maximum-failure-percentage): 

> By default, Ansible will continue executing actions as long as there are hosts in the group that have not yet failed.

Setting `max_fail_percentage` to `0` should enforce fail-fast behavior from Ansible with SecureDrop. While it would be ideal to set the option in `ansible.cfg`, it appears it's only supported at the play level—which means all playbooks should have it set _on each play_.

conorsch:
## Status

Ready for review.

## Description of Changes

Fixes #2885.

* sets `max_fail_percentage=0` on each and every play in all Ansible playbooks
* includes regressions tests to make sure any new plays include `max_fail_percentage=0`
* removes unused, nearly-three-year-old "upgrade" logic designed for SecureDrop 0.3
* uses new "ansible-config" Molecule scenario appropriate for local env testing (can be expanded in the future)

## Testing

There are actually two things we want to test:

1. Errors on one host during a prod run should cause immediate exit.
2. Any changes to Ansible playbooks that omit `max_fail_percentage: 0` should be caught by CI and cause the test suite to fail.

The first is the important one. It's not possible to test with staging, since staging already assumes that both hosts are present (via the ci-tweaks role) and will indeed fail fast if not. So, use prod VMs! Then:

1. Run `./securedrop-admin install`
2. During execution of the "common" role, perform hard shutdown on `mon-prod`.
3. Confirm that the entire playbook exits immediately, _without_ continuing to execute against `app-prod`.

Additional output for comparing the behavior under the develop branch can be found in https://github.com/freedomofpress/securedrop/issues/2885#issuecomment-360275071.

## Deployment

Designed to implement a "fail fast" strategy for production installs. Any error should cause immediate exit of the configuration run.


kushaldas:
Eh, I wanted mark it approved ;(

trishnaguha:
`any_errors_fatal: yes` can be used with `max_fail_percentage`. `any_errors_fatal` causes immediate abort of playbook execution when a single task fails against any of the hosts.

conorsch:
Thank you, @trishnaguha! Testing that locally now, will push up changes and edit the testing instructions when I confirm working!

conorsch:
Confirmed @trishnaguha's suggestions are exactly what we want. Updated testing instructions in the PR description. @kushaldas, please take for another spin—and thanks for calling in additional support on the Ansible logic!

conorsch:
@emkll 

> Tested with prod VMs and attempted hard shutdown of mon-prod during common tasks. It works well (if mon is shutdown, the action times out and the playbook is halted and app configuration does not continue)

Good! That's the "fail-fast" strategy we want at work, so thanks for confirming.

> However, for certain tasks such as TASK [common : Install cron-apt for unattended security upgrades.] and TASK [common : Ensure the system clock is set accurately.] the playbook does not exit immediately (nor after 60s) and hangs for quite long (undetermined amount of) time.

Yes, I've seen that too. You're right that the failure isn't _immediate_ in that after powering off, the task hangs for a while. However, as long as one host does not continue on being configured after the failed host is marked as failed, I'm comfortable with the implementation in terms of how it informs Admins about failed state, and increases the odds of competent debugging, but ensuring failed tasks remain near the bottom of the command output (and are therefore easy to document and share, e.g. in bug reports).

molecule/ansible-config/tests/test_max_fail_percentage.py
```diff
@@ -0,0 +1,80 @@
...
+    with open(playbook, 'r') as f:
+        playbook_yaml = yaml.safe_load(f)
+        # Descend into playbook list structure to validate play attributes.
+        for play in playbook_yaml:
+            assert 'max_fail_percentage' in play
+            assert play['max_fail_percentage'] == 0
+
+
+@pytest.mark.parametrize('playbook', find_ansible_playbooks())
+def test_max_fail_percentage(host, playbook):
```
conorsch: This test is named identically to the test above it, meaning it'll clobber, and only one test will be run. That's bad. It doesn't even pass flake8 linting, but CI missed that. See #2933 for tracking improvements to CI, and #2935 for fixing the test clobbering problem.

PR #2941, opened 2018-01-26
Test tor packages from `tor-apt-test` in release branches

### LINKED ISSUE #2919: QA 0.3.2.9-1~trusty+1 for release to tor-apt mirror
# Feature request

## Description

With the move of tor packaging to our own repository. We now have to QA each release as part of the release process. `0.3.2.9-1~trusty+1` is now included in the official upstream trusty repo (a bump from the current `0.3.1.9-1~trusty+1` installed on current servers).

Let's ensure that during QA of next release -- that latest tor package is also tested.

## User Stories

As a securedrop developer, I want to ensure latest stable tor is tested prior to release.

msheiny:
## Status

Work in progress

## Description of Changes

Partially fixes  #2919  -- performs testing against CI staging AFTER installing latest tor package in `tor-apt-test.freedom.press`

Ummm what I described right above this part... GEEEZ!!

## Testing

You want to check to make sure CI passed and confirm you are able to see a `.tor_version` artifact in circle CI and it's content matches version `0.3.2.9-1~trusty+1`

## Deployment

Well this PR affects CI specfically, but locally for QA you also want to make sure you are installing tor from tor-test-apt.freedom.press

- [] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

Runing tests locally now, I had to make a tweak to the tor apt test so it doesnt run in the release* branches (same behavior as when this tor apt test is supposed to run).


molecule/aws/side_effect.yml
```diff
@@ -9,6 +9,14 @@
     # WHEN REINSTATING REBOOT
     - include: reboot_and_wait.yml
       when: "false"
+    - include: tor_apt_test.yml
+      when: (lookup('env','CIRCLE_BRANCH')|default('na')).startswith('release')
```
conorsch: @msheiny Assigning defaults to env lookups often leads to surprises. I believe it's currently not possible to override an undefined env var with `|default`, but the lookup will return an empty string, which will satisfies the conditional logic here.

msheiny: ohhh shit you are totally right @conorsch - thanks for pointing that out

conorsch: Older versions of Ansible threw an uncatchable exception in the event of an undefined env var lookup, so the behavior has improved somewhat, but the silent ignore of `|default` warrants careful attention. Still, good to go here. 👌

PR #2948, opened 2018-01-28
Rewrite tests with pytest fixtures

### LINKED ISSUE #2877: Meta issue: No new tests until we have cleaned up the existing test infrastructure
# Feature request

## Description

We omit many best practices for our tests. One thing we should do is properly use `pytest`'s fixtures to generate our apps. This can be done with:

```python
@pytest.fixture(scope='session')
def source_app(config):
    app = source_app.create_app(config)
    app.testing = True
    app.config['WTF_CSRF_ENABLED'] = False
    app.config['SERVER_NAME'] = 'localhost'

    with app.app_context():
        db.create_all()
        with app.test_client() as app:
            yield app


@pytest.fixture(scope='session')
def journalist_app(config):
    app = journalist_app.create_app(config)
    app.testing = True
    app.config['WTF_CSRF_ENABLED'] = False
    app.config['SERVER_NAME'] = 'localhost'

    with app.app_context():
        db.create_all()
        with app.test_client() as app:
            yield app
```

Right now we have issues (#1444) that point out some of the errors we hit. Are are also manually mucking around with the `g` and object for some of our tests which is extremely brittle. The best thing to do to get a user/admin to a logged in state is to `POST` to `/login`.

In short, we really should work with fixtures and be mindful of the applications and context we're using since introducing #2459 (Flask-SQLAlchemy) will make it so that we have to pay attention to these contexts much more.

## User Stories

As a developer, I find that the hardest part of adding new features and refactoring is always, always in the tests. This is problematic since editing a test in parallel with the app code is how bugs get introduced.

heartsucker:
## Status

Ready for review

## Description of Changes

Work towards #2877 

Added `pytest` fixtures to centralize app creation logic for the source app

## Testing

`make test`

### DIFF
--- a/securedrop/tests/test_source.py
+++ b/securedrop/tests/test_source.py
@@ -31,6 +31,15 @@ class TestPytestSourceApp:
                 assert resp.status_code == 404
                 ins.assert_template_used('notfound.html')
 
+    def test_index(self, source_app):
+        """Test that the landing page loads and looks how we expect"""
+        with source_app.test_client() as app:
+            resp = app.get('/')
+            assert resp.status_code == 200
+            text = resp.data.decode('utf-8')
+            assert 'Submit documents for the first time' in text
+            assert 'Already submitted something?' in text
+
 
 class TestSourceApp(TestCase):
 
@@ -42,13 +51,6 @@ class TestSourceApp(TestCase):
 
     def tearDown(self):
         utils.env.teardown()
-
-    def test_index(self):
-        """Test that the landing page loads and looks how we expect"""
-        response = self.client.get('/')
-        self.assertEqual(response.status_code, 200)
-        self.assertIn("Submit documents for the first time", response.data)
-        self.assertIn("Already submitted something?", response.data)
 
     def test_all_words_in_wordlist_validate(self):
         """Verify that all words in the wordlist are allowed by the form
--- a/securedrop/tests/test_source.py
+++ b/securedrop/tests/test_source.py
@@ -16,8 +16,20 @@ import version
 from db import db
 from models import Source
 from utils.db_helper import new_codename
+from utils.instrument import InstrumentedApp
 
 overly_long_codename = 'a' * (Source.MAX_CODENAME_LEN + 1)
+
+
+class TestPytestSourceApp:
+
+    def test_page_not_found(self, source_app):
+        """Verify the page not found condition returns the intended template"""
+        with InstrumentedApp(source_app) as ins:
+            with source_app.test_client() as app:
+                resp = app.get('UNKNOWN')
+                assert resp.status_code == 404
+                ins.assert_template_used('notfound.html')
 
 
 class TestSourceApp(TestCase):
@@ -30,12 +42,6 @@ class TestSourceApp(TestCase):
 
     def tearDown(self):
         utils.env.teardown()
-
-    def test_page_not_found(self):
-        """Verify the page not found condition returns the intended template"""
-        response = self.client.get('/UNKNOWN')
-        self.assert404(response)
-        self.assertTemplateUsed('notfound.html')
 
     def test_index(self):
         """Test that the landing page loads and looks how we expect"""
--- a/securedrop/tests/test_source.py
+++ b/securedrop/tests/test_source.py
@@ -64,6 +64,31 @@ class TestPytestSourceApp:
                 assert 'Sorry, that is not a recognized codename.' in text
                 assert 'logged_in' not in session
 
+    def _find_codename(self, html):
+        """Find a source codename (diceware passphrase) in HTML"""
+        # Codenames may contain HTML escape characters, and the wordlist
+        # contains various symbols.
+        codename_re = (r'<p [^>]*id="codename"[^>]*>'
+                       r'(?P<codename>[a-z0-9 &#;?:=@_.*+()\'"$%!-]+)</p>')
+        codename_match = re.search(codename_re, html)
+        assert codename_match is not None
+        return codename_match.group('codename')
+
+    def test_generate(self, source_app):
+        with source_app.test_client() as app:
+            resp = app.get('/generate')
+            assert resp.status_code == 200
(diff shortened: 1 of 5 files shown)


kushaldas:
I think somehow the cache is causing trouble in CircleCI. `make test` passes properly in local system.

heartsucker:
@kushaldas That's what I thought too. I guess I'll open a ticket for that.

heartsucker:
For now, it seems like this PR is blocked by #3033. Once that is merged, the strange test failure logic related to this line:

```python
c = imp.load_module('c', *imp.find_module('config'))
```

should go away. But to be fair, I'm not positive that's the line that's causing problems, but it's definitely the most likely one.

securedrop/tests/test_source.py
```diff
@@ -74,462 +72,489 @@ def _find_codename(self, html):
...
+            ' ' + codename + ' ',
+            ' ' + codename,
+        ]
+
+        for codename_ in codenames:
+            with source_app.test_client() as app:
+                login_test(app, codename_)
+
+    @staticmethod
+    def _dummy_submission(app):
```
dachary: Nit: for clarity removal of the old _dummy_submission is in commit **pytest fixtures: test_csrf_error_page** but should be in commit **pytest fixtures: test_initial_submission_notification** where the static version is introduced.

securedrop/tests/test_source.py
```diff
@@ -74,462 +72,489 @@ def _find_codename(self, html):
...
+            assert resp.status_code == 200
+            resp = app.get('/lookup', follow_redirects=True)
+
+            # check that the session was cleared (apart from 'expires'
+            # which is always present and 'csrf_token' which leaks no info)
+            session.pop('expires', None)
+            session.pop('csrf_token', None)
+            assert not session, session
+            text = resp.data.decode('utf-8')
+            assert 'Your session timed out due to inactivity' in text
```
dachary: Why is it different from `You have been logged out due to inactivity` ? It indicates the test does not behave in the same way and this is either cause for concern or a detailed explanation in the commit message.

securedrop/tests/test_source.py
```diff
@@ -2,70 +2,68 @@
...
 from models import Source
+from source_app import main as source_app_main
 from utils.db_helper import new_codename
+from utils.instrument import InstrumentedApp
 
 overly_long_codename = 'a' * (Source.MAX_CODENAME_LEN + 1)
 
 
-class TestSourceApp(TestCase):
+class TestSourceApp:
```
redshiftzero: We can get rid of the `TestSourceApp` class now, which I think would be nice to flatten this file a bit (if you make this change you could do it in an additional commit if it's easier)

securedrop/tests/test_source.py
```diff
@@ -204,6 +204,38 @@ def test_user_must_log_in_for_protected_views(self, source_app):
...
+            resp = app.get('/login')
+            assert resp.status_code == 200
+            text = resp.data.decode('utf-8')
+            assert "Enter Codename" in text
+
+            resp = app.post('/login', data=dict(codename=codename),
+                            follow_redirects=True)
+            assert resp.status_code == 200
+            text = resp.data.decode('utf-8')
+            assert "Submit Materials" in text, text
```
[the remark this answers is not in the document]
dachary: If I'm not mistaken the idea here is to display `text` in case the assert fails but ... `pytest` instruments the `assert` and displays the actual values being compared when it fails. So +1 on the Nit ;-)

securedrop/tests/test_source.py
dachary: I think we need this class for `setup` and `teardown`?

securedrop/tests/test_source.py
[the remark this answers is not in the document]
heartsucker: Honestly, we should probably do that with all cases since seeing `...` in the error output is not very helpful, but that one probably got left in by mistake.

securedrop/tests/test_source.py
heartsucker: We don't use those anymore. Lifecycles are managed by `pytest` alone.

securedrop/tests/test_source.py
```diff
@@ -74,462 +72,489 @@ def _find_codename(self, html):
...
-            self.assertNotIn('logged_in', session)
-            self.assertNotIn('codename', session)
-
-            self.assertIn('Thank you for exiting your session!', resp.data)
+            resp = app.get('/journalist-key')
+            text = resp.data.decode('utf-8')
+            assert "BEGIN PGP PUBLIC KEY BLOCK" in text
+
+    def test_login_and_logout(self, source_app):
+        with source_app.test_client() as app:
```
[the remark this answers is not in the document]
heartsucker: Why is the grounds for that nit? I think it might do some silly things to sessions and whatnot (but could be wrong).

securedrop/tests/test_source.py
heartsucker: I don't move all the tests at once, so I leave it in the old class because there are tests that required it,

securedrop/tests/test_source.py
heartsucker: That string never appears in the source app, and also that test was prefixed with `_test` so it was never actually run before. We have higher coverage now that it is actually run.

heartsucker:
Pinging @dachary and @redshiftzero for rereview now that things are green again.

securedrop/tests/test_source.py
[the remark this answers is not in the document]
redshiftzero: Since the code in each block is a separate test case, having them use a different test client makes sense (really they should be in different tests entirely - but in the spirit of not expanding the scope of this PR beyond strictly rewriting the tests to use pytest fixtures, @dachary's suggestion seems reasonable)

securedrop/tests/test_source.py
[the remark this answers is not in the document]
heartsucker: Fixed

Issue #1261, opened 2016-03-11
Consider using unix sockets instead of localhost for Tor Hidden Services


garrettr:
# Background

SecureDrop currently uses the "standard" configuration of "expose service on localhost, then connect that to a hidden service via torrc" (see `install_files/ansible-base/roles/tor-hidden-services/templates/torrc`).

This configuration is problematic, because applications sometimes (fairly) assume that localhost has a different threat model than an externally exposed service. Most recently, this common configuration lead to some hidden services exposing sensitive information through the Apache mod_status module ([link](http://arstechnica.com/security/2016/02/default-settings-in-apache-may-decloak-tor-hidden-services/)). SecureDrop was never vulnerable to this particular issue because we have been aware of it for years and accordingly disable mod_status as well as the rest of the default Apache site configuration.

Nonetheless, I think this is a generally problematic configuration due to the mismatch in the intended use of localhost vs. its common (but frankly, inappropriate) use for externally exposed Tor Hidden Services. I am concerned that other, similar issues may be lurking.
# Alternatives to localhost

Riseup has a nice page on deploying hidden services (down right now, here's a [cached](https://webcache.googleusercontent.com/search?q=cache:yD4Jo8RNPg8J:https://help.riseup.net/en/security/network-security/tor/onionservices-best-practices+&cd=2&hl=en&ct=clnk&gl=us) copy) that discusses this issue in depth. Here are their proposed alternatives to localhost:

---

Be careful of localhost bypasses!

You should take very careful care to not accidentally expose things on your server that are restricted to the local machine. For example, if you provide /server-status in apache (from mod_status) to monitor the health of your apache webserver, that will typically be restricted to only allow access from 127.0.0.1, or you may have .htaccess rules that only allow localhost, etc.

There are a few ways you can solve this problem:

different machine: consider running the onion service on a different machine (real or virtual) than the actual service. This has the advantage that you can isolate the service from the onion service (a compromise of one doesn’t compromise the other) and helps with isolating potential information leaks

isolation: similarly to the above, you can also isolate tor and the service so it will run on a different network namespace than the service.

public ip: configure the onion service to connect to the public IP address of the service instead of localhost/127.0.0.1, this should make tor not pick 127.0.0.1 as the source address and avoid most misconfigurations. For example like this:
bc. HiddenServiceDir /var/lib/tor/hidden/ftp/
HiddenServicePort 80 192.168.1.1:81

[...]

PR #2963, opened 2018-02-02
Syscheck monitoring of /var/lib/tor/services/ 

### LINKED ISSUE #2960: Syscheck warning on missing directory specified in ossec agent configuration
# Bug

## Description

As of release 0.5.2, ossec logs contain a large number of the following warning message:
```
2018/02/02 14:12:46 ossec-syscheckd: WARN: Error opening directory: '/var/lib/tor/services/hostname': No such file or directory 
```

## Steps to Reproduce

* ssh into mon
* cat /var/ossec/logs/ossec.log
* observe a large amount of the following warning message:

```
2018/02/02 14:12:46 ossec-syscheckd: WARN: Error opening directory: '/var/lib/tor/services/hostname': No such file or directory 
```
## Expected Behavior

Warning should not appear in ossec logs, ossec configuration should specify files/folder that exist.

## Actual Behavior

Warning appears in ossec logs, ossec configuration does not contain files/folders that exist.

## Comments

https://github.com/freedomofpress/securedrop/blob/develop/install_files/securedrop-ossec-agent/var/ossec/etc/ossec.conf#L16 specifies the path, which has likely been removed as of tor 0.3.2.*

emkll:
## Status

Ready for review

## Description of Changes

Fixes #2960:
Modify ossec-agent configuration to monitor `/var/lib/tor/services/` directory.

## Testing

* `make build-debs`
* `vagrant up /staging/`
* `vagrant ssh mon-staging`
* `/var/ossec/queue/syscheck/\(app-staging\)\ 10.0.1.2-\>syscheck` should contain hashes all files and folders in `/var/lib/tor/services` and monitor/alert on changes.

## Deployment

Will require a new ossec-agent package to published.

### DIFF
--- a/install_files/securedrop-ossec-agent/var/ossec/etc/ossec.conf
+++ b/install_files/securedrop-ossec-agent/var/ossec/etc/ossec.conf
@@ -13,7 +13,7 @@
     <directories realtime="yes" check_all="yes" report_changes="yes">/var/securedrop</directories>
     <directories realtime="yes" check_all="yes" report_changes="yes">/var/www</directories>
     <directories realtime="yes" check_all="yes" report_changes="yes">/var/lib/securedrop</directories>
-    <directories realtime="yes" check_all="yes" report_changes="yes">/var/lib/tor/services/hostname</directories>
+    <directories realtime="yes" check_all="yes" report_changes="yes">/var/lib/tor/services</directories>
     <directories realtime="yes" check_all="yes" report_changes="yes">/var/lib/tor/lock</directories>
     <directories realtime="yes" check_all="yes" report_changes="yes">/boot</directories>
 


ageis:
@emkll @redshiftzero Hey both, just had a comment or thought here when I was looking at the list of directories that are being monitored. If we're speaking in terms of defense-in-depth, then you might want to monitor `/var/run/tor`. That's where the control socket and authentication cookie get written on Debian-based systems, so any changes would indicate someone has created a backdoor to the Tor process, which is explicitly disallowed, since you do not set a ControlSocket or CookieAuthentication in your torrc. However, there is a problem with doing this, which is that the PID file is in the same directory and subject to change.

conorsch:
@ageis #1261 describes switching to Unix sockets (i.e. files on disk) for the Tor service. If that gets implemented, then the socket path would ideally be monitored by OSSEC. As for watching anything under `/var/run/`, that'd generate too much noise for Admins without providing any meaningful information: files there are _expected_ to change.

msheiny:
Hey @emkll should this also be monitored on upgrades? I can't see any logic that would add it as-is.

PR #2695, opened 2017-12-08
Implement #2679: Compile translations in development environment

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #2679.

Changes proposed in this pull request:
* Adds compilation of translations to the dev env

## Testing

If you have already compiled translations, run `vagrant provision` to verify that the task added in this PR does not report changes made (due to an `mo` file already being present). 

If you have not already compiled translations, run `vagrant provision` and verify that the task in this PR runs successfully and produces a file e.g. in `/vagrant/securedrop/translations/fr_FR/LC_MESSAGES/messages.mo`.

## Deployment

Dev env only

## Checklist

Dev env only

PR #2971, opened 2018-02-05
docs: add instructions to developers to show the l10n menu

dachary:
## Status

## Description of Changes

When setting up the development virtual machine, the language menu
will not be displayed by default because the .po files are not
compiled into .mo files. We add instructions to setup the development
virtual machines so the menu is always displayed.

The translations need to be updated manually by running this
compilation step every time a source string changes in the source or
journalist interface.

## Testing

* follow the instructions and copy paste the line to check there is no typo

## Deployment

N/A

### DIFF
--- a/docs/development/virtual_environments.rst
+++ b/docs/development/virtual_environments.rst
@@ -45,6 +45,7 @@ To get started working with the development environment:
    ./manage.py run         # run development servers
    ./manage.py reset       # resets the state of the development instance
    ./manage.py add-admin   # create a user to use when logging in to the Journalist Interface
+   ./manage.py translate-messages --compile   # add languages and display the i18n menu
    pytest -v tests/        # run the unit and functional tests
 
 .. sidebar:: Note to Qubes users


dachary:
@SaptakS thanks :-)

docs/development/virtual_environments.rst
```diff
@@ -45,6 +45,7 @@ To get started working with the development environment:
    ./manage.py run         # run development servers
    ./manage.py reset       # resets the state of the development instance
    ./manage.py add-admin   # create a user to use when logging in to the Journalist Interface
+   ./manage.py translate-messages --compile   # add languages and display the i18n menu
```
redshiftzero: I think we'll just need to add this in the Dockerfile, since in #2695 I added this task in the Ansible development VM.

But for the language picker to be displayed, we'll need to add an Ansible task / RUN command that adds locales to `SUPPORTED_LOCALES` in `config.py` (probably only worth doing for Docker since the Ansible based development environment is slated for removal in 0.6).

dachary: Oh, right. I'll revise this PR accordingly, I was too quick. Thanks !

dachary:
closing as this will be a non-docs PR after all

Issue #1357, opened 2016-07-06
Make HTML/CSS responsive


heartsucker:
Currently many of the elements have fixed sizes (in pixels). This led to some ugly and poorly sized containers for the German translations in #1103 because of long words. It would probably look equally bad with very short words as in Chinese.

Before piling on more features, it might make sense to refactor the HTML/CSS (or hopefully SASS #1349) to support containers that resize depending on content size (as well as possibly monitor size). As I'm not usually a frontend dev, I'd roll it all by hand and skip a framework.

Currently there are a lot of features that require front changes, so it would be good to make this change before starting any of the others:  #1356, #1355, #1354, #1353,  #1171, #1170, (plus a nasty rebase of #1103)

heartsucker:
I can help with this, but it would be helpful if someone could merge #1349 first. We already have ugliness with the new wordlist because of the longer words.

![screenshot - 11102016 - 11 10 43 am](https://cloud.githubusercontent.com/assets/3998464/20172717/8a70ef36-a736-11e6-983b-3b608846781e.png)

conorsch:
Sorry about the holdup on #1349, @heartsucker—those changes are in `develop` now, so have at it!

heartsucker:
@conorsch No worries. I'm probably going to hold off on this since it looks like @ninavizz has some more major changes to the UI in the pipe, and any changes I make might conflict with that.

Also related: https://github.com/freedomofpress/securedrop/issues/1450

heartsucker:
Work tracked in branch [`responsive-css`](https://github.com/freedomofpress/securedrop/tree/responsive-css).

kushaldas:
@SaptakS please have a look :)

SaptakS:
@kushaldas sure. I will try to contribute and make the pages more responsive in case they aren't. Thanks. 😃

SaptakS:
There is one more thing that I wanted to ask. Apart from changing the fixed sizes (in pixels) to maybe percentage or `vw/vh`, one more thing that we often see is though things resize like they should with screen size, somethings appear better when done differently in mobile. 

Say, for example, in the /generate page, in wide screen, the differently sized buttons look good, but in narrower screens it looks weird:

### Wide Screen:
![screen shot 2018-01-04 at 6 12 49 pm](https://user-images.githubusercontent.com/9530293/34563985-fc704b74-f17a-11e7-81bb-00950ecf1bb9.png)

### Narrow Screen:
[...]

PR #2974, opened 2018-02-05
Fix some responsive issues in journalist interface

SaptakS:
## Status

Ready for review

## Description of Changes

* Use Flexbox in journalist interface similar to source interface
* Make minimum width of designation columns in journalist collections rows in percentage rather than pixels.
* Fix button wrapping and margins for smaller screens.

Related to  #1357 #2891 .

### DIFF
--- a/securedrop/journalist_templates/col.html
+++ b/securedrop/journalist_templates/col.html
@@ -24,8 +24,8 @@
       <p>
         <div id="select-container"></div>
         <button type="submit" name="action" value="download" class="small"><i class="fa fa-download"></i> {{ gettext('Download Selected') }}</button>
-        <a href="#delete-selected-confirmation-modal" id="delete-selected-link" class="btn small danger">
-          <i class="fa fa-trash-o"></i> {{ gettext('Delete Selected') }}
+        <a href="#delete-selected-confirmation-modal" id="delete-selected-link">
+          <button type="button" class="small danger"><i class="fa fa-trash-o"></i> {{ gettext('Delete Selected') }}</button>
         </a>
       </p>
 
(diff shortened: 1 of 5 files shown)


SaptakS:
Before minimum width in percentage:
screen shot 2018-02-05 at 3 06 22 pm

After minimum width in percentage:
screen shot 2018-02-05 at 3 07 04 pm

SaptakS:
Before button fix:
screen shot 2018-02-06 at 12 50 36 am

After button fix:
screen shot 2018-02-06 at 12 58 13 am

securedrop/journalist_templates/col.html
```diff
@@ -24,8 +24,8 @@
       <p>
         <div id="select-container"></div>
         <button type="submit" name="action" value="download" class="small"><i class="fa fa-download"></i> {{ gettext('Download Selected') }}</button>
-        <a href="#delete-selected-confirmation-modal" id="delete-selected-link" class="btn small danger">
```
dachary: could you please add to the commit message a short description of the problem for posterity ? It is unclear to me what removing the btn class solves.

SaptakS: So I didn't exactly remove the btn class. I created an entire button inside the <a> tag instead of making the <a> tag look like a button with the "btn" class. I did this because without it, in smaller screen, there was a partial wrapping rather than the entire button going to the next line as seen [here](https://github.com/freedomofpress/securedrop/pull/2974#issuecomment-363206840)

I will add a commit message for the same.

dachary: Oh, I overlooked that part of the diff, thanks for explaining. It makes total sense now. Good to merge once the commit message is amended so it is cristal clear to future readers.

SaptakS: @dachary updated the commit messages. Please check if it looks fine.

PR #2979, opened 2018-02-07
Remove global state of `config`

### LINKED ISSUE #2969: Replace `import config` with dependency injection
# Feature request

## Description

We still do `import config` everywhere in the app which breaks dependency injection and leaves us with a bit of spaghetti code. This negatively impacts tests in that it makes it quite hard to properly use fixtures in testing.

## User Stories

As a dev, I want to be able to dependency injection during tests.

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #2969.

Attempts to remove global state so we can properly do dependency injection during testing which will be critically important when we parallelize tests.

## Testing

`make images test`

### DIFF
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -1,18 +1,18 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-from base64 import b32encode
+import gnupg
 import os
+import scrypt
 import subprocess
 
+from base64 import b32encode
 from Cryptodome.Random import random
-import gnupg
+from flask import current_app
 from gnupg._util import _is_stream, _make_binary_stream
-import scrypt
 from typing import Dict, List, Text  # noqa: F401
 
 import config
-import store
 
 # to fix gpg error #78 on production
 os.environ['USERNAME'] = 'www-data'
@@ -182,7 +182,7 @@ def getkey(name):
 def encrypt(plaintext, fingerprints, output=None):
     # Verify the output path
     if output:
-        store.verify(output)
+        current_app.storage.verify(output)
 
     if not isinstance(fingerprints, (list, tuple)):
         fingerprints = [fingerprints, ]
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -12,51 +12,206 @@ from flask import current_app
 from gnupg._util import _is_stream, _make_binary_stream
 from typing import Dict, List, Text  # noqa: F401
 
-import config
-
 # to fix gpg error #78 on production
 os.environ['USERNAME'] = 'www-data'
-
-GPG_KEY_TYPE = "RSA"
-if os.environ.get('SECUREDROP_ENV') == 'test':
-    # Optimize crypto to speed up tests (at the expense of security - DO NOT
-    # use these settings in production)
-    GPG_KEY_LENGTH = 1024
-    SCRYPT_PARAMS = dict(N=2**1, r=1, p=1)
-else:  # pragma: no cover
-    GPG_KEY_LENGTH = 4096
-    SCRYPT_PARAMS = config.SCRYPT_PARAMS
-
-SCRYPT_ID_PEPPER = config.SCRYPT_ID_PEPPER
-SCRYPT_GPG_PEPPER = config.SCRYPT_GPG_PEPPER
-
-DEFAULT_WORDS_IN_RANDOM_ID = 8
-
-
-# Make sure these pass before the app can run
-# TODO: Add more tests
-def do_runtime_tests():
-    if config.SCRYPT_ID_PEPPER == config.SCRYPT_GPG_PEPPER:
-        raise AssertionError('SCRYPT_ID_PEPPER == SCRYPT_GPG_PEPPER')
-    # crash if we don't have srm:
-    try:
-        subprocess.check_call(['srm'], stdout=subprocess.PIPE)
-    except subprocess.CalledProcessError:
-        pass
-
-
-do_runtime_tests()
-
-gpg = gnupg.GPG(binary='gpg2', homedir=config.GPG_KEY_DIR)
-
-# map code for a given language to a localized wordlist
-language2words = {}  # type: Dict[Text,List[str]]
-nouns = open(config.NOUNS).read().rstrip('\n').split('\n')
-adjectives = open(config.ADJECTIVES).read().rstrip('\n').split('\n')
 
 
 class CryptoException(Exception):
     pass
+
+
+class CryptoUtil:
+
+    GPG_KEY_TYPE = "RSA"
+    DEFAULT_WORDS_IN_RANDOM_ID = 8
+
+    def __init__(self,
+                 scrypt_params,
+                 scrypt_id_pepper,
+                 scrypt_gpg_pepper,
+                 securedrop_root,
+                 word_list,
+                 nouns_file,
+                 adjectives_file,
+                 gpg_key_dir):
+        self.__securedrop_root = securedrop_root
+        self.__word_list = word_list
+
+        if os.environ.get('SECUREDROP_ENV') == 'test':
+            # Optimize crypto to speed up tests (at the expense of security
+            # DO NOT use these settings in production)
+            self.__gpg_key_length = 1024
+            self.scrypt_params = dict(N=2**1, r=1, p=1)
+        else:  # pragma: no cover
+            self.__gpg_key_length = 4096
+            self.scrypt_params = scrypt_params
+
+        self.scrypt_id_pepper = scrypt_id_pepper
+        self.scrypt_gpg_pepper = scrypt_gpg_pepper
+
+        self.do_runtime_tests()
+
+        self.gpg = gnupg.GPG(binary='gpg2', homedir=gpg_key_dir)
+
+        # map code for a given language to a localized wordlist
+        self.__language2words = {}  # type: Dict[Text, List[str]]
+
+        with open(nouns_file) as f:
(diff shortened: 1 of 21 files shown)


dachary:
I really like what you've done here :-) It's a baby step in the right direction and we can do something similar for other global states :100: 

Just one thing to help review the PR in detail: it would be nice to separate the whitespace-only changes of **dependency injection for crypto_util** in a separate commit so the diff actually highlights what has changed instead of showing as big walls of red/blue. The whitespace only does not even need to be proper python, just the indent needed to move functions in the new class.

securedrop/crypto_util.py
```diff
@@ -1,54 +1,207 @@
...
+
+        self.do_runtime_tests()
+
+        self.__gpg_key_dir = gpg_key_dir
+        self.gpg = gnupg.GPG(binary='gpg2', homedir=gpg_key_dir)
+
+        # map code for a given language to a localized wordlist
+        self.__language2words = {}  # type: Dict[Text, List[str]]
+
+        with open(nouns_file) as f:
```
kushaldas: Thank you for these context managers.  💯

kushaldas: We should try to move all the `open` calls into `io.open` calls. This will help us to another step towards Python3.

securedrop/crypto_util.py
```diff
@@ -1,54 +1,207 @@
...
+            self.nouns = f.read().rstrip('\n').split('\n')
```
kushaldas: We should use `f.read().splitlines()` here. Related [docs](https://docs.python.org/2/library/stdtypes.html#str.splitlines) for more reading.

securedrop/crypto_util.py
```diff
@@ -1,54 +1,207 @@
...
+                                    'wordlists',
+                                    locale + '.txt')
+                if os.path.exists(path):
+                    wordlist_path = path
+                else:
+                    wordlist_path = self.__word_list
+            else:
+                wordlist_path = self.__word_list
+
+            with open(wordlist_path) as f:
```
kushaldas: @dachary @heartsucker Are we sure that those `localename.txt` files are always in ASCII? Moving to `io.open` will help here too.

securedrop/crypto_util.py
```diff
@@ -1,54 +1,207 @@
...
+                content = f.read().rstrip('\n').split('\n')
```
kushaldas: We should use splitlines here too.

securedrop/crypto_util.py
```diff
@@ -1,54 +1,207 @@
...
+                               *fingerprints,
+                               output=output,
+                               always_trust=True,
+                               armor=False)
+        if out.ok:
+            return out.data
+        else:
+            raise CryptoException(out.stderr)
 
+<<<<<<< HEAD
```
kushaldas: `git` artifacts.

securedrop/crypto_util.py
```diff
@@ -57,6 +210,19 @@ def do_runtime_tests():
 
 class CryptoException(Exception):
     pass
+=======
```
kushaldas: `git` artifacts.

securedrop/crypto_util.py
```diff
@@ -57,6 +210,19 @@ def do_runtime_tests():
...
+        >>> key = self.genkeypair('randomid', 'randomid')
+        >>> self.decrypt('randomid',
+        ...   self.encrypt('Goodbye, cruel world!', str(key))
+        ... )
+        'Goodbye, cruel world!'
+        """
+        hashed_codename = self.hash_codename(secret,
+                                             salt=self.scrypt_gpg_pepper)
+        return self.gpg.decrypt(ciphertext, passphrase=hashed_codename).data
+>>>>>>> dependency injection for crypto_util.py
```
kushaldas: `git` artifacts.

securedrop/crypto_util.py
```diff
@@ -78,6 +244,7 @@ def clean(s, also=''):
     # scrypt.hash requires input of type str. Since the wordlist is all ASCII
     # characters, this conversion is not problematic
     return str(s)
+<<<<<<< HEAD
```
kushaldas: `git` artifacts.

securedrop/crypto_util.py
```diff
@@ -220,3 +387,5 @@ def decrypt(secret, ciphertext):
 if __name__ == "__main__":  # pragma: no cover
     import doctest
     doctest.testmod()
+=======
+>>>>>>> dependency injection for crypto_util.py
```
kushaldas: `git` artifacts in the above 2 lines.

securedrop/crypto_util.py
heartsucker: But why? Python3 has `open` as a function?

securedrop/crypto_util.py
dachary: yes they are (the two of them ;-)

securedrop/crypto_util.py
kushaldas: Because then we will be using the same function call in Python2 and in future in Python3. The `open` call in Python3 actually points to `io.open` 

If we keep using `open` function call now, means we will change the implementation when we switch to Python3. The suggestion we provide is to use `io.open` when a project tries to move towards Python3.

heartsucker: Ok, well we can do that in another PR. This is a restructuring of the current code to allow us to do dependency injection.

kushaldas: Okay.

securedrop/crypto_util.py
```diff
@@ -12,51 +12,206 @@
...
+        else:  # pragma: no cover
+            self.__gpg_key_length = 4096
+            self.scrypt_params = scrypt_params
+
+        self.scrypt_id_pepper = scrypt_id_pepper
+        self.scrypt_gpg_pepper = scrypt_gpg_pepper
+
+        self.do_runtime_tests()
+
+        self.__gpg_key_dir = gpg_key_dir
```
redshiftzero: Nit: Attribute `__gpg_key_dir` is not used

securedrop/crypto_util.py
```diff
@@ -12,51 +12,206 @@
...
+                content = f.read().rstrip('\n').split('\n')
+                self.__language2words[locale] = content
+
+        return self.__language2words[locale]
+
+    def genrandomid(self,
+                    words_in_random_id=None,
+                    locale='en'):
+        if words_in_random_id is None:
+            words_in_random_id = self.DEFAULT_WORDS_IN_RANDOM_ID
```
redshiftzero: Seems like setting `self.DEFAULT_WORDS_IN_RANDOM_ID` as the default kwarg for `words_in_random_id` would obviate the need for these two lines of code (beginning `if words_in_random_id is None:`)

securedrop/crypto_util.py
```diff
@@ -12,51 +12,206 @@
...
+
+    def hash_codename(self, codename, salt=None):
+        """Salts and hashes a codename using scrypt.
+
+        :param str codename: A source's codename.
+        :param str salt: The salt to mix with the codename when hashing.
+        :returns: A base32 encoded string; the salted codename hash.
+        """
+        if salt is None:
+            salt = self.scrypt_id_pepper
```
redshiftzero: Same comment here - why this pattern?

securedrop/crypto_util.py
```diff
@@ -12,51 +12,206 @@
...
-adjectives = open(config.ADJECTIVES).read().rstrip('\n').split('\n')
-
-
-class CryptoException(Exception):
-    pass
+    def decrypt(self, secret, ciphertext):
+        """
+        >>> key = self.genkeypair('randomid', 'randomid')
+        >>> self.decrypt('randomid',
+        ...   self.encrypt('Goodbye, cruel world!', str(key))
```
redshiftzero: This string was changed in #2982, let's keep those changes

securedrop/source_app/utils.py
```diff
@@ -67,8 +70,8 @@ def wrapper(*args, **kwargs):
 
 
 @async
-def async_genkey(filesystem_id, codename):
-    crypto_util.genkeypair(filesystem_id, codename)
+def async_genkey(crypto_util_, filesystem_id, codename):
```
redshiftzero: Passing this `crypto_util_` around is a bit unclear at first blush: let's include a comment in the code for the benefit of future maintainers?

securedrop/crypto_util.py
heartsucker: It breaks because `self` isn't defined at class definition time. I did this so we could leave the `DEFAULT_WORDS_IN_RANDOM_ID` attached to the class, but if you don't like this pattern I can make this constant part of the module and then set it as the default in the function.

heartsucker:
```
>>> class Foo():
...     X = 1
...     def bar(self, baz=self.X):
...         return baz
... 
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 3, in Foo
NameError: name 'self' is not defined
```

securedrop/crypto_util.py
heartsucker: Fixed

securedrop/source_app/utils.py
heartsucker: Fixed

securedrop/crypto_util.py
heartsucker: Fixed

securedrop/crypto_util.py
redshiftzero: ahhhh right, that's cool then

PR #2982, opened 2018-02-07
Change sad crypto_util doctest sample text

nvesely:
## Status

Ready for review

## Description of Changes

Given the loss of two great contributors to this project to suicide, I think it would be nice to remove this rather depressing doctest function sample text. The new text reads, in Spanish, "Good day, beautiful world!"

## Testing

`python -m doctest -v crypto_util.py` should suffice as only doctest function documentation is changing.


heartsucker:
Merging because this makes the world a little brighter.

PR #2988, opened 2018-02-09
source: genkey runs in a thread and must have its own session

### LINKED ISSUE #2978: async_genkey runs in a thread that has no flask application
# Bug

## Description

async_genkey tries to access the database in a separate thread and fails with the error like

<pre>
[2018-02-06 00:05:33,620] ERROR in utils: async_genkey for source (filesystem_id=AXT3PQIZMYAMHOZ4MIVX6ZGHCWP3GKWNI5UW2UMMANIGQJ6CUTYH45VTEP63X2DAWPRTS3PUPQBAFCLI4JMOZHVZPOZXGIKWMCFJQ3I=): No application found. Either work inside a view function or push an application context. See http://flask-sqlalchemy.pocoo.org/contexts/.
</pre>

See an example of failure probably due to this race https://gist.github.com/conorsch/b76b0ae5f81b1a78428d59e1b4666e7a#file-app-test-failures-log-L358

## Steps to Reproduce

Not sure.

## Expected Behavior

* pytest tests/test_integration.py runs with no failure.

## Actual Behavior

* pytest tests/test_integration.py runs sometime fails with
<pre>
No application found. Either work inside a view function or push an application context. See http://flask-sqlalchemy.pocoo.org/contexts/.
</pre>

dachary:
## Status

Ready for review

## Description of Changes

Fixes #2978 

The session available from Source.query belongs to the main thread
and is bound to a Flask app context. However, since async_genkey runs
in a separate thread, it is entirely possible that there is no active
app context when it tries to access the session and it will fail with
an error like:

    No application found. Either work inside a view function or
    push an application context.
    See http://flask-sqlalchemy.pocoo.org/contexts/

In addition the query from the thread races against the queries from
the main thread and it may lead to failure or corruption.

Since there is no need for async_genkey to work in an app_context, we
create a new engine and a new session for the duration of the thread.
Writing to SQLite from multiple processes or threads is protected by
locks (http://www.sqlite.org/wal.html)

## Testing

The modified code is extensively used in the integration tests.

## Deployment

N/A

### DIFF
--- a/securedrop/source_app/utils.py
+++ b/securedrop/source_app/utils.py
@@ -3,13 +3,14 @@ import subprocess
 
 from datetime import datetime
 from flask import session, current_app, abort, g
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
 from threading import Thread
 
 import crypto_util
 import i18n
 import store
 
-from db import db
 from models import Source
 
 
@@ -68,21 +69,23 @@ def async(f):
 
 
 @async
-def async_genkey(filesystem_id, codename):
+def async_genkey(db_uri, filesystem_id, codename):
     crypto_util.genkeypair(filesystem_id, codename)
 
     # Register key generation as update to the source, so sources will
     # filter to the top of the list in the journalist interface if a
     # flagged source logs in and has a key generated for them. #789
+    session = sessionmaker(bind=create_engine(db_uri))()
     try:
-        source = Source.query.filter(Source.filesystem_id == filesystem_id) \
-                       .one()
+        source = session.query(Source).filter(
+            Source.filesystem_id == filesystem_id).one()
         source.last_updated = datetime.utcnow()
-        db.session.commit()
+        session.commit()
     except Exception as e:
         logging.getLogger(__name__).error(
                 "async_genkey for source (filesystem_id={}): {}"
                 .format(filesystem_id, e))
+    session.close()
 
 
 def normalize_timestamps(filesystem_id):
--- a/securedrop/source_app/main.py
+++ b/securedrop/source_app/main.py
@@ -90,7 +90,8 @@ def make_blueprint(config):
         # Only do this if the journalist has flagged the source as one
         # that they would like to reply to. (Issue #140.)
         if not crypto_util.getkey(g.filesystem_id) and g.source.flagged:
-            async_genkey(g.filesystem_id, g.codename)
+            db_uri = current_app.config['SQLALCHEMY_DATABASE_URI']
+            async_genkey(db_uri, g.filesystem_id, g.codename)
 
         return render_template(
             'lookup.html',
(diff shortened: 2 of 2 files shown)


dachary:
@tmc @heartsucker it looks like this is a reasonable way forward. But I'm not very happy that we don't have a way to repeat the original problem, i.e. have the query run after the app_context is gone. Do you have a suggestion for that ?

dachary:
it does not work, I've seen another error of the same kind. My guess is that the scoped_session is similarly bound to the app. We need to find a way to obtain a scoped_session that is not bound to the app. Threat safety is good but one more step to go.

dachary:
Hum, actually the source of the problem was different: I incorrectly included the app multiple times, overriding the globals. That does not count but let's keep this in mind anyway.

securedrop/source_app/utils.py
```diff
@@ -75,8 +75,8 @@ def async_genkey(filesystem_id, codename):
     # filter to the top of the list in the journalist interface if a
     # flagged source logs in and has a key generated for them. #789
     try:
-        source = Source.query.filter(Source.filesystem_id == filesystem_id) \
-                       .one()
+        source = Source(db.scoped_session).query.filter(
```
heartsucker: I think this won't work because the `db` object is a proxy that maps back to the main flask `app`. The app might not have a context when this code executes in a new thread. I think the safest thing to do is pass the connection string in and let the thread spawn its own connection. Has this been tested in a scenario that reproduces the race condition?

heartsucker: Just saw your other comments on this. 😐 😐 😐 😐

dachary:
Thanks for the feedback, I suspected that would not be that simple ;-)

> the connection string

You mean the db_uri

<pre>
db_uri = (config.DATABASE_ENGINE + ":///" +
                  config.DATABASE_FILE)
</pre>

heartsucker:
@dachary Yes, passing in the db_uri via`current_app.config['SQLALCHEMY_DATABASE_URI']` would be sufficient, and then we can have it set up to do something like create it's own engine and session from the URI alone.

dachary:
@heartsucker thanks for the good tip :-) How does it look now ?

dachary:
@heartsucker thanks for your review :-) @redshiftzero since this is such a central piece of code, it would be nice if you could also look into this change ?

dachary:
Although we don't have a reproducer, we could add a test that calls **async_genkey** out of context. That will demonstrate it does not depend on flask. @redshiftzero if you think that's a good idea, I'll add that.

securedrop/source_app/utils.py
```diff
@@ -68,17 +69,18 @@ def wrapper(*args, **kwargs):
...
     # flagged source logs in and has a key generated for them. #789
+    session = sessionmaker(bind=create_engine(db_uri))()
     try:
-        source = Source.query.filter(Source.filesystem_id == filesystem_id) \
-                       .one()
+        source = session.query(Source).filter(
+            Source.filesystem_id == filesystem_id).one()
         source.last_updated = datetime.utcnow()
-        db.session.commit()
+        session.commit()
```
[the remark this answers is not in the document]
dachary: Good point ! Done.

dachary:
rebase & fixed & repushed ;-)

PR #2989, opened 2018-02-09
Allow blank sasl_domain in sdconfig

### LINKED ISSUE #2482: Installation does not like having no sasl_domain
# Bug

## Description

Our mail server auths with just a username, not a username and a domain.

## Steps to Reproduce

set sasl_domain to "" in the site-specific vars
run `./securedrop-admin install`

## Expected Behavior

 - pass validation
 - installs a config at /etc/postfix/sasl_password that looks like

```
[$smtp_relay]:$smtp_port      $sasl_username:$sasl_password
```

## Actual Behavior

 - Fails validation by trying to nslookup empty string
 - writes out a config with an `@` after the sasl_username

## Comments

I think the `@` is ok, it just looks super weird. It seems to be working though.

emkll:
## Status

Ready for review

## Description of Changes

SecureDrop configuration should now allow `sasl_domain` to be empty string. Fixes #2482 .

## Testing

Test both with sasl_domain set as fqdn or empty string and observe the configuration in `/etc/postfix/sasl_passwd`

## Deployment

Should only affect new installs, as existing installs should have their email setup correctly.

### DIFF
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -200,14 +200,6 @@ class SiteConfig(object):
                 message=("Must contain a @ and be set to "
                          "something other than ossec@ossec.test"))
 
-    class ValidateSASLDomain(Validator):
-        def validate(self, document):
-            if document.text == '' or SiteConfig.ValidateFQDN():
-                return True
-            raise ValidationError(
-                message=("Must contain a valid domain or be empty")
-            )
-
     def __init__(self, args):
         self.args = args
         translations = SiteConfig.Locales(
@@ -274,7 +266,7 @@ class SiteConfig(object):
              int],
             ['sasl_domain', "gmail.com", str,
              u'SASL domain for sending OSSEC alerts',
-             SiteConfig.ValidateSASLDomain(),
+             None,
              None],
             ['sasl_username', '', str,
              u'SASL username for sending OSSEC alerts',
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -200,6 +200,14 @@ class SiteConfig(object):
                 message=("Must contain a @ and be set to "
                          "something other than ossec@ossec.test"))
 
+    class ValidateSASLDomain(Validator):
+        def validate(self, document):
+            if document.text == '' or SiteConfig.ValidateFQDN():
+                return True
+            raise ValidationError(
+                message=("Must contain a valid domain or be empty")
+            )
+
     def __init__(self, args):
         self.args = args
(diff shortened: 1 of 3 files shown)


admin/securedrop_admin/__init__.py
```diff
@@ -200,6 +200,14 @@ def validate(self, document):
                 message=("Must contain a @ and be set to "
                          "something other than ossec@ossec.test"))
 
+    class ValidateSASLDomain(Validator):
+        def validate(self, document):
+            if document.text == '' or SiteConfig.ValidateFQDN():
```
dachary: It has been decided to not validate the domain (i.e. not try to resolve them) because we're not able to make that work reliably in Tails at this moment. Reason why there is no validation for the SASLDomain.

admin/tests/test_securedrop-admin.py
```diff
@@ -287,6 +287,26 @@ def test_validate_gpg_key(self, caplog):
...
+            '65A1B5FF195B56353CC63DFFCC40EF1228271441',
+
+            'sasl_username':
+            'alice',
+
+            'sasl_domain':
+            '',
+        }
+        assert site_config.ValidateSASLDomain()
+
```
dachary: Since there is no domain name validation at the moment, there is no need for a dedicated validation test function. The consistency check of the `sasl_domain` declaration above are taken care of by:

<pre>
verify_prompt_sasl_domain = verify_desc_consistency
</pre>

below

install_files/ansible-base/roles/postfix/templates/sasl_passwd
```diff
@@ -1 +1 @@
-[{{ smtp_relay }}]:{{ smtp_relay_port }}   {{ sasl_username }}@{{ sasl_domain }}:{{ sasl_password }}
+[{{ smtp_relay }}]:{{ smtp_relay_port }}   {{ sasl_username }}{% if sasl_domain %}@{{ sasl_domain }}{% endif %}:{{ sasl_password }}
```
dachary: Ideally we would have molecule tests checking both cases (empty/not empty) work but we don't have that yet. However it looks fine visually :-)

admin/securedrop_admin/__init__.py
emkll: Should I keep this class as-is nonetheless? I would have to create a new class anyways (e.g. `ValidateNothing`) since every other field is validated and no existing class satisfies the validation requirements. I would also remove the validation alltogether but it feels like a step backwards.

dachary: You can set it to **None** and it will not be validated at all.

PR #2990, opened 2018-02-09
tests: docker based development environment

### LINKED ISSUE #2902: Deprecate Ansible dev environment in favor of Docker dev environment
## Description

Currently we have two development environments: one provisioned with Ansible and one using Docker. Since the purpose of the development environment is to allow developers to iterate quickly on the application code, I think we should deprecate the Ansible-based environment in favor of the Docker environment once #2861 is merged. 

## User Stories

As a SecureDrop maintainer, I want to maintain a single development environment. 

As a SecureDrop developer, I want a development environment that spins up quickly with minimal dependencies so I can get to writing code as quickly as possible.

dachary:
## Status

Ready for review

## Description of Changes

Fixes #2902 

A docker image that contains all dependencies installed (deb and pip) replaces the former image. The script part (running Xvfb, ./manage.py etc.) is delegated to scripts in the bin directory instead. This minimal image is the basis to run tests, shell interaction, screenshot generation etc.

The creation of the image is no longer a separate makefile target: it is always created or updated before running a command in the container. Checking the image does not need to update does not take long and the delay imposed when running a command in the container is minimal.

The bin/dev-shell does something different (for docker build and docker run) if it runs in CircleCI or not. It would be unecessarily complicated to unify the two because:

* bind mounting the directory or running under the same user in the host and in the container is not required for CI
* the docker container does not run on the same host as the CI
* the CI needs explicit instructions to use the Docker cache

The bind/dev-deps daemons are run from with functions that capture their output to reduce the output cluter: it is only displayed in case a daemon fails.

A function to create a demo user with the same credentials as http://demo.securedrop.club is added to facilitate testing.

## Testing

* make docs
* firefox localhost:8000 and verify the documentation changes
* run the documentation samples to verify they work as expected
* verify `pytest --pages-layout tests` is run successfully by CI

## Deployment

N/A


dachary:
I'm puzzled by the lack of coverage. There is no error but https://codecov.io/gh/freedomofpress/securedrop/pull/2990/tree shows 0% coverage, confusing statistics and when clicking on https://codecov.io/gh/freedomofpress/securedrop/pull/2990/src/crypto_util.py it loops forever. Something is not right but ... what ?

securedrop/manage.py
```diff
@@ -446,6 +455,15 @@ def get_args():
     return parser
 
 
+def set_add_admin_parser(subps, name):
+    parser = subps.add_parser(name, help='Add an admin to the application.')
+    parser.add_argument('--password', help='force password')
+    parser.add_argument('--username', help='force username')
+    parser.add_argument('--totp', action='store_true', help='force TOTP')
+    parser.add_argument('--otp-secret', help='force OTP secret')
```
redshiftzero: I understand that you're doing this so we can create a test user, but since these options will be exposed to current SecureDrop administrators, I don't think we want to give them the ability to override any of these (for their own good) - what about a one off script that adds the test user (directly using `sqlalchemy`)?

securedrop/bin/run-test
```diff
@@ -1,13 +1,20 @@
...
+urandom
+run_sass
+maybe_create_config_py
 
 if [ -n "${CIRCLE_BRANCH:-}" ] ; then
     touch tests/log/firefox.log
     function finish {
         cp tests/log/firefox.log /tmp/test-results/logs/
-        cp /tmp/securedrop*log /tmp/test-results/logs/
         bash <(curl -s https://codecov.io/bash)
```
redshiftzero: I wonder if this is broken because codecov isn't searching in `/tmp/test-results` where the coverage reports are -- we could add a `cd /tmp/test-results` prior to running `bash <(curl -s https://codecov.io/bash)`

securedrop/manage.py
```diff
@@ -180,7 +187,13 @@ def _add_user(is_admin=False):
         user = Journalist(username=username,
                           password=password,
                           is_admin=is_admin,
-                          otp_secret=otp_secret)
+                          otp_secret=otp_secret,
+                          is_totp=not is_hotp)
+        if args.password:
+            user.pw_salt = user._gen_salt()
+            user.pw_hash = user._scrypt_hash(args.password, user.pw_salt)
```
redshiftzero: (note the salting and hashing is done by `Journalist.set_password` in `models.py` automatically)

securedrop/manage.py
redshiftzero: See script in this commit 4774568324965cc02f4dae2035d1d1fef71ba655 - you could run this instead, let me know what you think

kushaldas: I would also prefer this script as we will not expose anything to the admins. If we provide any command line option available, someone will try those and complain if something breaks.

dachary: Good point @redshiftzero @kushaldas, will use this script :-)

securedrop/bin/run-test
dachary: It occurs to me that it could be because codecov keeps the **absolute** path of the sources instead of the relative path. And since the working directory of the Docker container changed, that could create the confusion.

securedrop/manage.py
dachary: The Journalist.set_password function is not used because it enforces
seven words diceware passwords and the demo password does not comply.
Although it would make for a more educative example to use a seven
words password, it is more difficult to copy/paste, all pages
displaying the existing password would have to be modified and the
users would need to update.

docs/development/virtual_environments.rst
```diff
@@ -1,11 +1,14 @@
 Virtual Environments: Servers
```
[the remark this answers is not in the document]
dachary: Good point. I changed the **docs: deprecated Vagrant for Docker in the documentation** commit to insert a "Quick start" section which is most probably the only thing new developers will need to read.

dachary:
@kushaldas @redshiftzero thanks for your careful review :-) I think all your comments were addressed, please let me know if I forgot something.

dachary:
@redshiftzero the codecov problem is because the path in which tests run changed. Another PR was created to demonstrate the problem, setting the workdir to /app and [codecov shows coverage is preserved](https://github.com/freedomofpress/securedrop/pull/3019#issuecomment-365960591), modulo the newly added script which is untested because it is part of the development tools. It is fixed by moving everything to /app again.

redshiftzero:
hey @tmc @squeed @heartsucker @msheiny - this one also might also be of interest to you 😇

securedrop/bin/dev-deps
```diff
@@ -1,8 +1,46 @@
...
+
+function maybe_create_config_py() {
+    if ! test -f config.py ; then
+        append_to_exit "rm config.py"
+        make test-config
+    fi
+}
+
+function run_sass() {
+    sass --force --stop-on-error --update sass:static/css --style compressed
```
heartsucker: It would be better to use `--watch` instead of `--update` and then do `run_sass &` for live updates.

securedrop/create-demo-user.py
```diff
@@ -0,0 +1,41 @@
...
+    context.push()
+    valid_password = "correct horse battery staple profanity oil chewy"
+
+    try:
+        user = Journalist(username=username,
+                          password=valid_password,
+                          is_admin=is_admin)
+        user.otp_secret = otp_secret
+        user.pw_salt = user._gen_salt()
+        user.pw_hash = user._scrypt_hash(password, user.pw_salt)
```
heartsucker: Why are you hashing `password` but not `valid_password`?

testinfra/test.py
```diff
@@ -9,11 +9,7 @@
 import sys
 import tempfile
 
-# By default let's assume we're testing against the development VM.
-try:
-    target_host = sys.argv[1]
-except IndexError:
-    target_host = "development"
+target_host = sys.argv[1]
```
heartsucker: I would prefer to see us using `argparse` instead of this to make the scripts a little more documented to that we can always safely do `./script -h` and get something useful out.

securedrop/bin/dev-shell
```diff
@@ -1,29 +1,82 @@
...
+}
+
+function docker_run_circle() {
+    docker run -ti ${DOCKER_RUN_ARGUMENTS:-} --name securedrop-test "securedrop-test:${CIRCLE_SHA1}" "$@"
+}
+
+function docker_image() {
+    local out
+    out="$(mktemp)"
+    if ! docker build \
```
heartsucker: Minor nit: I feel like `cmd || exit 1` is a little easier to read than this. But this is minor and feel free to disagree.

securedrop/create-demo-user.py
dachary: The Journalist.set_password function is not used because it enforces
seven words diceware passwords and the demo password does not comply.
Although it would make for a more educative example to use a seven
words password, it is more difficult to copy/paste, all pages
displaying the existing password would have to be modified and the
users would need to update.

testinfra/test.py
dachary: This would be a valuable addition to test.py but I would prefer it is implemented in a separate pull request.

securedrop/bin/dev-shell
dachary: I prefer shell functions that return instead of abort. The primary benefit is that you can test such functions. When a shell script is cattered with exit statements, it is very difficult to write tests without disrupting the logic. It is like refactoring code that relies on throwing exception into code that relies on the return value of functions to handle errors.

securedrop/bin/dev-deps
dachary: Nice catch: that would have been a regression. Done.

dachary:
@heartsucker thanks for your review, updated & rebased & repushed !

securedrop/bin/dev-shell
heartsucker: Fair enough.

testinfra/test.py
heartsucker: Also fair.

redshiftzero:
@kushaldas were your comments addressed here?

PR #2992, opened 2018-02-10
tests: guard modal clicks against race

dachary:
## Status

Ready for review

## Description of Changes

Whenever a modal is clicked, verify the corresponding link is
displayed or not. This implicitly waits for a few seconds and gives a
better chance for the DOM state to change as expected. It will also
fail closer to the root cause when an error happends (wrong id etc.).

## Testing

* cd securedrop ; make test

## Deployment

N/A

### DIFF
--- a/securedrop/tests/functional/journalist_navigation_steps.py
+++ b/securedrop/tests/functional/journalist_navigation_steps.py
@@ -75,47 +75,55 @@ class JournalistNavigationStepsMixin():
     def _journalist_selects_first_doc(self):
         self.driver.find_elements_by_name('doc_names_selected')[0].click()
 
+    def _journalist_clicks_on_modal(self, click_id):
+        self.driver.find_element_by_id(click_id).click()
+
+        def link_not_displayed():
+            assert (
+                (not self.driver.find_elements_by_id(click_id)) or
+                (not self.driver.find_element_by_id(click_id).is_displayed()))
+        self.wait_for(link_not_displayed)
+
     def _journalist_clicks_delete_collections_cancel_on_modal(self):
-        self.driver.find_element_by_id('cancel-collections-deletions').click()
+        self._journalist_clicks_on_modal('cancel-collections-deletions')
 
     def _journalist_clicks_delete_selected_cancel_on_modal(self):
-        self.driver.find_element_by_id('cancel-selected-deletions').click()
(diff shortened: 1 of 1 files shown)


dachary:
@SaptakS I have seen an occasional failure on my laptop (3 / 4 times today) because the DOM did not have time to display. There is an implicit 10 seconds wait on click() and this should be enough in most cases. The click modal functions are factorized into a helper with additional waiting time so it has less chance of happening. There is no 100% guarantee that it will never happen but the extra wait time should be enough.

securedrop/tests/functional/journalist_navigation_steps.py
```diff
@@ -139,8 +147,10 @@ def _journalist_uses_delete_collection_button_confirmation(self):
 
         # After deletion the button will redirect us. Let's ensure we still
         # see the delete collection button.
-        assert self.driver.find_element_by_id(
-            'delete-collection-link').is_displayed()
+        def delete_collection_link_displayed():
+            assert self.driver.find_element_by_id(
+                'delete-collection-link').is_displayed()
+        self.wait_for(delete_collection_link_displayed, timeout=60)
```
SaptakS: @dachary I see timeout only in this wait_for and not the other wait_for function calls. Shouldn't it be there in others as well?

dachary: The default time of 10sec should be fine for the other wait_for. I set it to a generous 60 sec because this is where the race happened a few times today on my laptop.

dachary:
hopefully an AWS transient error https://circleci.com/gh/freedomofpress/securedrop/7933?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

<pre>
"Unable to change state for instance i-0463e0c5b623ce081, error: EC2ResponseError: 403 Forbidden\n<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Response><Errors><Error><Code>UnauthorizedOperation</Code><Message>You are not authorized to perform this operation.
</pre>

securedrop/tests/functional/journalist_navigation_steps.py
SaptakS: Okay. Makes sense then I guess.

Issue #2181, opened 2017-08-23
Improve handling of screenshots in functional tests


redshiftzero:
In light of [these tracebacks](https://github.com/freedomofpress/securedrop/pull/2141#issuecomment-324418928), I propose we:
 - [ ] Disable/remove the `@screenshots` decorator in the functional tests
 - [x] Use the pytest `--page-layout` option in CI
- [ ] ~~Add the screenshots as test artifacts~~

PR #2994, opened 2018-02-10
tests: remove @screenshot decorator because it obfuscates errors

dachary:
## Status

Ready for review

## Description of Changes

Fixes: #2181

@screenshot was a good idea in theory but has not been used for
debugging in practice. However, in some circumstances it obfuscates
stack traces in ways that are non trivial. It would be possible to fix
this. However since it is not used we can simplify the problem by
removing the decorator.

## Testing

CI covers the modified test code

## Deployment

N/A


dachary:
@tmc it bothers me to remove this code because it is potentially very useful. I'm not sure why we did not get to use it. Was it because we forget it exists when we need it most ?

PR #3002, opened 2018-02-13
Fixes #3001  initial support for type annotation

### LINKED ISSUE #3001: [RFE] Add type annotation
# Feature request

## Description

We need to add the initial type annotation in the Python code for #2795

kushaldas:
Co-authored-by: Brett Cannon <brett@python.org>

## Status

Ready for review

## Description of Changes

Fixes #3001 

We have the initial type annotation. Means `typing` module is added as an app dependency.

## Testing

<pre>
sudo apt install python3-dev
python3 -m venv .py3
source .py3/bin/activate
pip install wheel ; pip install mypy
mypy securedrop
</pre>

### DIFF
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -50,7 +50,7 @@ do_runtime_tests()
 gpg = gnupg.GPG(binary='gpg2', homedir=config.GPG_KEY_DIR)
 
 # map code for a given language to a localized wordlist
-language2words = {}  # type: Dict[Text,List[Text]]
+language2words = {}  # type: Dict[Text,List[str]]
 nouns = open(config.NOUNS).read().rstrip('\n').split('\n')
 adjectives = open(config.ADJECTIVES).read().rstrip('\n').split('\n')
 
@@ -81,6 +81,7 @@ def clean(s, also=''):
 
 
 def _get_wordlist(locale):
+    # type: (Text) -> List[str]
     """" Ensure the wordlist for the desired locale is read and available
     in the words global variable. If there is no wordlist for the
     desired local, fallback to the default english wordlist.
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -9,6 +9,7 @@ from Cryptodome.Random import random
 import gnupg
 from gnupg._util import _is_stream, _make_binary_stream
 import scrypt
+from typing import Dict, List, Text  # noqa: F401
 
 import config
 import store
@@ -49,7 +50,7 @@ do_runtime_tests()
 gpg = gnupg.GPG(binary='gpg2', homedir=config.GPG_KEY_DIR)
 
 # map code for a given language to a localized wordlist
-language2words = {}
+language2words = {}  # type: Dict[Text,List[Text]]
 nouns = open(config.NOUNS).read().rstrip('\n').split('\n')
 adjectives = open(config.ADJECTIVES).read().rstrip('\n').split('\n')
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,6 @@
+[mypy]
+ignore_missing_imports = True
+python_version = 2.7
+
+[mypy-config]
+ignore_errors = True
--- a/securedrop/journalist_app/utils.py
+++ b/securedrop/journalist_app/utils.py
@@ -20,6 +20,7 @@ from rm import srm
 
 
 def logged_in():
+    # type: () -> bool
     # When a user is logged in, we push their user ID (database primary key)
     # into the session. setup_g checks for this value, and if it finds it,
     # stores a reference to the user's Journalist object in g.
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -7,7 +7,7 @@ import binascii
 try:
     from cStringIO import StringIO
 except ImportError:
-    from StringIO import StringIO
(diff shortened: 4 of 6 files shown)


dachary:
There are two errors from the config.py file:

<pre>
$ mypy --py2 --ignore-missing-imports securedrop/journalist_app
securedrop/config.py:52: error: "Type[FlaskConfig]" has no attribute "USE_X_SENDFILE"
securedrop/config.py:58: error: "Type[FlaskConfig]" has no attribute "MAX_CONTENT_LENGTH"
</pre>

dachary:
I suggest adding the testing steps to `devops/docker/Dockerfile.linting` which would make the followup PRs easier to review.

kushaldas:
I don't to enable mypy on the CI as part of this PR. I think that should be done as a second PR. I will update the PR for the changes related to `config.py.example`.

securedrop/config.py.example
```diff
@@ -5,6 +5,8 @@ class FlaskConfig(object):
     DEBUG = False
     TESTING = False
     WTF_CSRF_ENABLED = True
+    USE_X_SENDFILE = False
```
dachary: As commented above these are the default values for production and should therefore be `USE_X_SENDFILE = True` to be consistent. It does not matter much because it is overriden later in the file but ... let's not confuse people ;-)

securedrop/config.py.example
```diff
@@ -5,6 +5,8 @@ class FlaskConfig(object):
...
+    MAX_CONTENT_LENGTH = 524288000
```
dachary: This one is consistent with production but it would need to be kept in sync with the value from `install_files/ansible-base/roles/app/templates/sites-available/journalist.conf`  and `install_files/ansible-base/roles/app/templates/sites-available/source.conf`. This is problematic because there is no update path for this value in the config.py file.

For instance imagine we raise that value in the apache files to 1GB. It will be updated via `./securedrop-admin ansible` but ... the value in config.py stays 500MB and the change has no effect.

I'm afraid we'll have to find another way to make mypy happy.

kushaldas: Btw, those values are from the current config.py.example itself.

dachary: disabling the check for this file seems like a good idea

securedrop/crypto_util.py
```diff
@@ -49,7 +50,7 @@ def do_runtime_tests():
 gpg = gnupg.GPG(binary='gpg2', homedir=config.GPG_KEY_DIR)
 
 # map code for a given language to a localized wordlist
-language2words = {}
+language2words = {}  # type: Dict[Text,List[Text]]
```
dachary: I think it would be better to limit the type declaration to something the static anlysis can actually verify, for instance:
<pre>
language2words = {}  # type: Dict[Any,Any]
</pre>
instead of what we know to be true at this moment in time. This will help maintainability. For instance, if the data structure changes in the future, there is a good chance this declaration will be overlooked and become an incorrect and confusing description.

kushaldas: 1. Usage of `Any` is the last option. We should try to avoid it as much as possible.
2. In future if we change the type, `mypy` will catch that and fail (error). That is the reason why use type annotation checking in CI.

dachary: Hum, I may have missed something then. If I change `type: Dict[Text,List[Text]]` to `type: Dict[int,int]` it does not throw an error. Why is that ?

kushaldas: @dachary Because we are yet to add type annotations in https://github.com/freedomofpress/securedrop/blob/8d6f9db6a6fbb1a33cf1324828de7c99e2ccb12c/securedrop/crypto_util.py#L83 mypy by design will not check/verify the usage unless the calling function is also typed. 

This enables us to have incremental type annotations. 

If you want to experiment, just add this in the line below that type annotation.

```
language2words[42] = 9001
```

dachary: Oh, I see. Would you mind adding annotations to **_get_wordlist** so it is properly enforced ? That would be a good example of a minimal, consistent use of mypy. 👍

dachary:
@kushaldas looks good once squashed: it will make a fine example. I tried to change:
<pre>
language2words = {}  # type: Dict[Text,List[str]]
</pre>
to
<pre>
language2words = {}  # type: Dict[int,List[str]]
</pre>
and got an error as expected:
<pre>
$ mypy --py2 securedrop/journalist_app
securedrop/crypto_util.py:107: error: Invalid index type "unicode" for "Dict[int, List[str]]"; expected type "int"
securedrop/crypto_util.py:110: error: Invalid index type "unicode" for "Dict[int, List[str]]"; expected type "int"
</pre>

Nice !

dachary:
@kushaldas ready to merge once squashed 💯

securedrop/crypto_util.py
```diff
@@ -50,7 +50,7 @@ def do_runtime_tests():
 gpg = gnupg.GPG(binary='gpg2', homedir=config.GPG_KEY_DIR)
 
 # map code for a given language to a localized wordlist
-language2words = {}  # type: Dict[Text,List[Text]]
```
dachary: this fixes the previous commit. If you would like to keep it as two commits, could you please move this to the previous commit ? Otherwise squashing into one is good because the two commits are closely related.

PR #3007, opened 2018-02-13
Fixes #3006 Adds Makefile target for type linting

### LINKED ISSUE #3006: [RFE] Add a devops container and related Makefile changes for mypy
# Feature request

## Description

We need a Makefile target to build and run mypy inside of a container for checking type annotation.

## User Stories

- I am developer, I want to run mypy on the codebase.
- Next, we will enable the same in the CI.

This is towards the goal of fixing #2795.

kushaldas:
Co-authored-by: Brett Cannon brett@python.org

## Status

Ready for review

## Description of Changes

Fixes #3006 

We now have a new Makefile target and related Dockerfile to check type annotations using mypy.
I am building the container on top of Fedora 27 as it has the latest Python3 packaged from the distribution.

## Testing

```
make ci-typelint-image
make ci-typelint
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,14 @@ ci-lint-image: ## Builds linting container.
 ci-lint: ## Runs linting in linting container.
 	docker run --rm -ti -v /var/run/docker.sock:/var/run/docker.sock securedrop-lint:${TAG}
 
+.PHONY: ci-typelint-image
+ci-typelint-image: ## Builds type glinting container.
+	docker build $(EXTRA_BUILD_ARGS) -t securedrop-typelint:${TAG} -f devops/docker/Dockerfile.typeannotation .
+
+.PHONY: ci-typelint
+ci-typelint: ## Runs type linting in container.
+	docker run --rm -ti -v /var/run/docker.sock:/var/run/docker.sock securedrop-typelint:${TAG}
+
 .PHONY: ansible-config-lint
 ansible-config-lint: ## Runs custom Ansible env linting tasks.
 	molecule verify -s ansible-config
--- a/devops/docker/Dockerfile.typeannotation
+++ b/devops/docker/Dockerfile.typeannotation
@@ -1,0 +1,10 @@
+FROM circleci/python:3.6.4-stretch
+
+
+RUN  sudo pip3 install mypy
+
+WORKDIR /src
+COPY . /src
+
+
+CMD ["sudo", "mypy", "./securedrop", "./admin"]


devops/docker/Dockerfile.typeannotation
```diff
@@ -0,0 +1,12 @@
...
+# To install mypy from pip
+RUN dnf install -y python3-devel python3-pip gcc
+
+RUN  pip3 install mypy
+
+WORKDIR /src
+VOLUME /src
+
+
+CMD ["mypy", "./securedrop", "./admin"]
```
dachary: It would be better to have that in `devops/docker/Dockerfile.linting` so all linting operations are in one place.

kushaldas: This requires a latest Python3 system. That is why I made a new container out of Fedora. Also, this mounts the directory as a VOLUME to the container, which is better as we don't have to build the actual container every time.

dachary: I don't understand what you mean by " build the actual container every time." ?

Makefile
```diff
@@ -27,6 +27,14 @@ ci-lint-image: ## Builds linting container.
 ci-lint: ## Runs linting in linting container.
 	docker run --rm -ti -v /var/run/docker.sock:/var/run/docker.sock securedrop-lint:${TAG}
 
+.PHONY: ci-typelint-image
+ci-typelint-image: ## Builds type linting container.
+	docker build $(EXTRA_BUILD_ARGS) -t securedrop-typelint:${TAG} -f devops/docker/Dockerfile.typeannotation .
+
+.PHONY: ci-typelint
+ci-typelint: ## Runs type linting in container.
+	docker run --rm -ti -v /var/run/docker.sock:/var/run/docker.sock -v ${PWD}:/src securedrop-typelint:${TAG}
```
dachary: that will not work because the docker container does not run on the same machine as the CI

devops/docker/Dockerfile.typeannotation
kushaldas: The linting container copies the source into the container, and then runs within that. Means, if I have made some changes, and then did not rebuild the container, it will not get those changes into the linting container.

Where as here we are just mounting the source code directory as a `VOLUME`.  But, your next comment makes it invalid. I will update the PR now.

kushaldas:
@msheiny Correct point about the root user, which I just copy pasted from the existing example. I will drop to an user now.

devops/docker/Dockerfile.typeannotation
msheiny: yeah @dachary - @kushaldas 's motivation makes sense here. Ideally we really want to be doing volume mounting.. I think honestly, I just made a change request about this. but it might make more sense to utilize circleci `machine` mode and use volume mounting everywhere instead of constantly having to rebuild the containers for linting

kushaldas:
Updated the PR as suggested by @dachary and @msheiny

kushaldas:
I hope we can call the Makefile target from one of the existing CI step, say the linting one.

kushaldas:
I am going to update this PR to use the circleci image. Thank you @dachary for pointing that out.

PR #3027, opened 2018-02-17
securedrop-admin: an empty config is {}, not None

dachary:
## Status

## Description of Changes

An empty config is {}, not None, otherwise it cannot be updated and will fail when starting
from scratch when non site-specific file exists

## Testing

In addition to the test added to ensure this regression stays fixed, one can manually test with

* git clone https://github.com/freedomofpress/securedrop
* cd securedrop/admin
* ./securedrop-admin setup
* ./securedrop-admin sdconfig

## Deployment

N/A


dachary:
@msheiny thanks for fixing the fixes :-)

kushaldas:
We should be able to find issues like these using `mypy` in future. Though we will have to keep adding type annotations 👍

PR #3032, opened 2018-02-19
tests: merge Dockerfile.typeannotation with Dockerfile.linting

dachary:
## Status

Ready for review

## Description of Changes

This saves us the trouble of adding typeannotation to the CI.

## Testing

* make ci-lint-image ci-lint

## Deployment

N/A


dachary:
@kushaldas this also enables typeannotation in the CI :-)

dachary:
Verified that mypy [actually runs from CI](https://circleci.com/gh/freedomofpress/securedrop/8631?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

<pre>
if [[ ! -d .python3/.venv ]] ; then \
  virtualenv --python=python3 .python3/.venv && \
  .python3/.venv/bin/pip3 install mypy ; \
        fi
.python3/.venv/bin/mypy ./securedrop ./admin
</pre>

.dockerignore
```diff
@@ -1,6 +1,11 @@
 .git
 .vagrant
+**/_build
```
heartsucker: Can you add these to the `.gitignore` too?

devops/docker/Dockerfile.linting
```diff
@@ -1,7 +1,8 @@
-FROM circleci/python:2
+FROM circleci/python:2-stretch
```
heartsucker: For linting, could we use an alpine based image to cut down on image sizes?

dachary: I'd rather not because it would impose that people know about Alpine in terms of packaging, dependencies etc. We have to know about stretch already because it's what Tails is made from. If you feel strongly about saving on image size versus reducing the amount of knowledge necessary to maintain CI, I'd be happy to discuss that in a thread in the forum.

.dockerignore
dachary: good idea, will do !

dachary:
@heartsucker fixed & repushed ! Thanks for your quick review :-)

devops/docker/Dockerfile.linting
heartsucker: Ok, so this isn't a blocker for this but I'll post to the forum some time for that.

devops/docker/Dockerfile.linting
```diff
@@ -15,4 +16,6 @@ RUN sudo pip install pyenchant==2.0.0
 WORKDIR /src
 COPY . /src
 
-CMD ["sudo", "make", "--keep-going", "lint"]
+RUN sudo make install-mypy
```
heartsucker: For caching reasons, this should be above the `COPY` command. Also, shouldn't we mount `$PWD:/src` instead of copying everything in to the container? It makes for a marginally faster edit/test cycle.

dachary: this docker image is designed to work for CI that does not allow mounting volumes and contrary to the admin and securedrop images, it does not have the required special case that would allow mounting . We could have a separate PR to optimize this for development and allow mounting. If you feel strongly this must be done for this PR to be merged I'd be happy to do the needful.

dachary:
> Can we somehow say to only run one part of the lint?

You're right, it is not convenient and I'll update accordingly.

dachary:
@heartsucker 

* modified the .gitignore
* made it so mypy is not rebuilt every time in the **tests: rebuild mypy in the lint docker image if the Makefile changes** commit

@kushaldas made it convenient to copy/paste the Makefile target to run part of the lint in the commit **tests: move the ticker shell helper to a separate file** (please kindly take a look at the commit message for more information)

PR #3033, opened 2018-02-19
Fixes #3031 type annotation for application config

### LINKED ISSUE #3031: [RFE] Have type annotation for config object
# Feature request

## Description

The config object we pass around to create the Flask applications should be properly annotated.

## User Stories

Related to #2795

kushaldas:
## Status

Ready for review.

## Description of Changes

Fixes #3031 

We now pass around `SDConfig` which can be checked by mypy.

## Testing

`make ci-typinglint-image ci-typelint`

### DIFF
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -18,15 +18,21 @@ from journalist_app.utils import get_source, logged_in
 from models import Journalist
 from store import Storage
 
+MYPY = False
+if MYPY:
+    from sdconfig import SDConfig  # noqa: F401
+
 _insecure_views = ['main.login', 'static']
 
 
 def create_app(config):
+    # type: (SDConfig) -> Flask
     app = Flask(__name__,
                 template_folder=config.JOURNALIST_TEMPLATES_DIR,
                 static_folder=path.join(config.SECUREDROP_ROOT, 'static'))
 
     app.config.from_object(config.JournalistInterfaceFlaskConfig)
+    app.sdconfig = config
 
     CSRFProtect(app)
     Environment(app)
@@ -74,7 +80,7 @@ def create_app(config):
     app.jinja_env.trim_blocks = True
     app.jinja_env.lstrip_blocks = True
     app.jinja_env.globals['version'] = version.__version__
-    if hasattr(config, 'CUSTOM_HEADER_IMAGE'):
+    if config.CUSTOM_HEADER_IMAGE:
         app.jinja_env.globals['header_image'] = config.CUSTOM_HEADER_IMAGE
         app.jinja_env.globals['use_custom_header_image'] = True
     else:
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -18,8 +18,13 @@ from journalist_app.utils import get_source, logged_in
 from models import Journalist
 from store import Storage
 
-MYPY = False
-if MYPY:
+import typing
+# https://www.python.org/dev/peps/pep-0484/#runtime-or-type-checking
+if typing.TYPE_CHECKING:
+    # flake8 can not understand type annotation yet.
+    # That is why all type annotation realted import
+    # statements has to be marked as noqa.
+    # http://flake8.pycqa.org/en/latest/user/error-codes.html?highlight=f401
     from sdconfig import SDConfig  # noqa: F401
 
 _insecure_views = ['main.login', 'static']
@@ -80,7 +85,7 @@ def create_app(config):
     app.jinja_env.trim_blocks = True
     app.jinja_env.lstrip_blocks = True
     app.jinja_env.globals['version'] = version.__version__
-    if config.CUSTOM_HEADER_IMAGE:
+    if hasattr(config, 'CUSTOM_HEADER_IMAGE'):
         app.jinja_env.globals['header_image'] = config.CUSTOM_HEADER_IMAGE
         app.jinja_env.globals['use_custom_header_image'] = True
     else:
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -22,7 +22,7 @@ import typing
 # https://www.python.org/dev/peps/pep-0484/#runtime-or-type-checking
 if typing.TYPE_CHECKING:
     # flake8 can not understand type annotation yet.
-    # That is why all type annotation realted import
+    # That is why all type annotation relative import
     # statements has to be marked as noqa.
     # http://flake8.pycqa.org/en/latest/user/error-codes.html?highlight=f401
     from sdconfig import SDConfig  # noqa: F401
@@ -86,7 +86,8 @@ def create_app(config):
     app.jinja_env.lstrip_blocks = True
     app.jinja_env.globals['version'] = version.__version__
     if hasattr(config, 'CUSTOM_HEADER_IMAGE'):
-        app.jinja_env.globals['header_image'] = config.CUSTOM_HEADER_IMAGE
+        app.jinja_env.globals['header_image'] = \
+            config.CUSTOM_HEADER_IMAGE  # type: ignore
         app.jinja_env.globals['use_custom_header_image'] = True
     else:
         app.jinja_env.globals['header_image'] = 'logo.png'
--- a/securedrop/sdconfig.py
+++ b/securedrop/sdconfig.py
@@ -1,0 +1,65 @@
+# -*- coding: utf-8 -*-
+
+import config as _config
+
+MYPY = False
+if MYPY:
+    from typing import List  # noqa: F401
+
+
+class SDConfig(object):
+    def __init__(self):
+        # type: () -> None
+        self.ADJECTIVES = _config.ADJECTIVES
+        self.DATABASE_ENGINE = _config.DATABASE_ENGINE
+        self.DEFAULT_LOCALE = _config.DEFAULT_LOCALE
+        self.FlaskConfig = _config.FlaskConfig
+        self.GPG_KEY_DIR = _config.GPG_KEY_DIR
+        self.JOURNALIST_KEY = _config.JOURNALIST_KEY
+        self.JOURNALIST_TEMPLATES_DIR = _config.JOURNALIST_TEMPLATES_DIR
+        self.JournalistInterfaceFlaskConfig = \
+            _config.JournalistInterfaceFlaskConfig
+        self.NOUNS = _config.NOUNS
+        self.SCRYPT_GPG_PEPPER = _config.SCRYPT_GPG_PEPPER
+        self.SCRYPT_ID_PEPPER = _config.SCRYPT_ID_PEPPER
+        self.SCRYPT_PARAMS = _config.SCRYPT_PARAMS
+        self.SECUREDROP_DATA_ROOT = _config.SECUREDROP_DATA_ROOT
+        self.SECUREDROP_ROOT = _config.SECUREDROP_ROOT
+        self.SESSION_EXPIRATION_MINUTES = _config.SESSION_EXPIRATION_MINUTES
+        self.SOURCE_TEMPLATES_DIR = _config.SOURCE_TEMPLATES_DIR
+        self.STORE_DIR = _config.STORE_DIR
+        self.SUPPORTED_LOCALES = []  # type: List[str]
+        self.SourceInterfaceFlaskConfig = _config.SourceInterfaceFlaskConfig
+        self.TEMP_DIR = _config.TEMP_DIR
+        self.WORD_LIST = _config.WORD_LIST
+        self.WORKER_PIDFILE = _config.WORKER_PIDFILE
+        self.env = _config.env
+        self.os = _config.os
+        # Below 4 attributes are only for non-sqlite env
+        self.DATABASE_USERNAME = ""
+        self.DATABASE_PASSWORD = ""
+        self.DATABASE_HOST = ""
+        self.DATABASE_NAME = ""
+        self.DATABASE_FILE = ""
+        # This by default is an empty string
+        self.CUSTOM_HEADER_IMAGE = ""
+
+        if _config.DATABASE_ENGINE != "sqlite":
+            self.DATABASE_USERNAME = _config.DATABASE_USERNAME  # type: ignore
+            self.DATABASE_PASSWORSD = \
+                _config.DATABASE_PASSWORD   # type: ignore
+            self.DATABASE_HOST = _config.DATABASE_HOST  # type: ignore
+            self.DATABASE_NAME = _config.DATABASE_NAME  # type: ignore
+        else:
+            self.DATABASE_FILE = _config.DATABASE_FILE
+
+        if hasattr(_config, 'CUSTOM_HEADER_IMAGE'):
+            self.CUSTOM_HEADER_IMAGE = \
+                _config.CUSTOM_HEADER_IMAGE  # type: ignore
+
+        if hasattr(_config, 'SUPPORTED_LOCALES'):
+            self.SUPPORTED_LOCALES = \
+                _config.SUPPORTED_LOCALES  # type: ignore
+
+
+config = SDConfig()  # type: SDConfig
--- a/securedrop/sdconfig.py
+++ b/securedrop/sdconfig.py
@@ -2,56 +2,47 @@
 
 import config as _config
 
-MYPY = False
-if MYPY:
-    from typing import List  # noqa: F401
+import typing
+# https://www.python.org/dev/peps/pep-0484/#runtime-or-type-checking
+if typing.TYPE_CHECKING:
+    # flake8 can not understand type annotation yet.
+    # That is why all type annotation realted import
+    # statements has to be marked as noqa.
+    # http://flake8.pycqa.org/en/latest/user/error-codes.html?highlight=f401
+    from typing import List, Dict  # noqa: F401
 
 
 class SDConfig(object):
     def __init__(self):
         # type: () -> None
-        self.ADJECTIVES = _config.ADJECTIVES
-        self.DATABASE_ENGINE = _config.DATABASE_ENGINE
-        self.DEFAULT_LOCALE = _config.DEFAULT_LOCALE
-        self.FlaskConfig = _config.FlaskConfig
-        self.GPG_KEY_DIR = _config.GPG_KEY_DIR
-        self.JOURNALIST_KEY = _config.JOURNALIST_KEY
-        self.JOURNALIST_TEMPLATES_DIR = _config.JOURNALIST_TEMPLATES_DIR
-        self.JournalistInterfaceFlaskConfig = \
-            _config.JournalistInterfaceFlaskConfig
-        self.NOUNS = _config.NOUNS
-        self.SCRYPT_GPG_PEPPER = _config.SCRYPT_GPG_PEPPER
-        self.SCRYPT_ID_PEPPER = _config.SCRYPT_ID_PEPPER
-        self.SCRYPT_PARAMS = _config.SCRYPT_PARAMS
-        self.SECUREDROP_DATA_ROOT = _config.SECUREDROP_DATA_ROOT
-        self.SECUREDROP_ROOT = _config.SECUREDROP_ROOT
-        self.SESSION_EXPIRATION_MINUTES = _config.SESSION_EXPIRATION_MINUTES
-        self.SOURCE_TEMPLATES_DIR = _config.SOURCE_TEMPLATES_DIR
-        self.STORE_DIR = _config.STORE_DIR
-        self.SUPPORTED_LOCALES = []  # type: List[str]
-        self.SourceInterfaceFlaskConfig = _config.SourceInterfaceFlaskConfig
-        self.TEMP_DIR = _config.TEMP_DIR
-        self.WORD_LIST = _config.WORD_LIST
-        self.WORKER_PIDFILE = _config.WORKER_PIDFILE
-        self.env = _config.env
-        self.os = _config.os
-        # Below 4 attributes are only for non-sqlite env
-        self.DATABASE_USERNAME = ""
-        self.DATABASE_PASSWORD = ""
-        self.DATABASE_HOST = ""
-        self.DATABASE_NAME = ""
-        self.DATABASE_FILE = ""
-        # This by default is an empty string
-        self.CUSTOM_HEADER_IMAGE = ""
+        if hasattr(_config, 'FlaskConfig'):
+            self.FlaskConfig = _config.FlaskConfig  # type: ignore
 
-        if _config.DATABASE_ENGINE != "sqlite":
+        if hasattr(_config, 'JournalistInterfaceFlaskConfig'):
+            self.JournalistInterfaceFlaskConfig = \
+                _config.JournalistInterfaceFlaskConfig  # type: ignore
+
+        if hasattr(_config, 'SourceInterfaceFlaskConfig'):
+            self.SourceInterfaceFlaskConfig = \
+                _config.SourceInterfaceFlaskConfig  # type: ignore
+
+        if hasattr(_config, "DATABASE_ENGINE"):
+            self.DATABASE_ENGINE = _config.DATABASE_ENGINE  # type: ignore
+
+        if hasattr(_config, "DATABASE_FILE"):
+            self.DATABASE_FILE = _config.DATABASE_FILE  # type: ignore
+
+        if hasattr(_config, "DATABASE_USERNAME"):
             self.DATABASE_USERNAME = _config.DATABASE_USERNAME  # type: ignore
-            self.DATABASE_PASSWORSD = \
-                _config.DATABASE_PASSWORD   # type: ignore
+
+        if hasattr(_config, "DATABASE_PASSWORD"):
+            self.DATABASE_PASSWORD = _config.DATABASE_PASSWORD  # type: ignore
+
+        if hasattr(_config, "DATABASE_HOST"):
             self.DATABASE_HOST = _config.DATABASE_HOST  # type: ignore
+
+        if hasattr(_config, "DATABASE_NAME"):
             self.DATABASE_NAME = _config.DATABASE_NAME  # type: ignore
-        else:
-            self.DATABASE_FILE = _config.DATABASE_FILE
 
         if hasattr(_config, 'CUSTOM_HEADER_IMAGE'):
             self.CUSTOM_HEADER_IMAGE = \
@@ -60,6 +51,77 @@ class SDConfig(object):
         if hasattr(_config, 'SUPPORTED_LOCALES'):
             self.SUPPORTED_LOCALES = \
                 _config.SUPPORTED_LOCALES  # type: ignore
+
+        # Now we will fill up existing values from config.py
+        if hasattr(_config, "ADJECTIVES"):
+            self.ADJECTIVES = _config.ADJECTIVES  # type: str
+
+        if hasattr(_config, "DATABASE_ENGINE"):
+            self.DATABASE_ENGINE = _config.DATABASE_ENGINE  # type: str
+
+        if hasattr(_config, "DATABASE_FILE"):
+            self.DATABASE_FILE = _config.DATABASE_FILE  # type: str
+
+        if hasattr(_config, "DEFAULT_LOCALE"):
+            self.DEFAULT_LOCALE = _config.DEFAULT_LOCALE  # type: ignore
+
+        if hasattr(_config, "GPG_KEY_DIR"):
+            self.GPG_KEY_DIR = _config.GPG_KEY_DIR  # type: str
+
+        if hasattr(_config, "JOURNALIST_KEY"):
+            self.JOURNALIST_KEY = _config.JOURNALIST_KEY  # type: str
+
+        if hasattr(_config, "JOURNALIST_TEMPLATES_DIR"):
+            self.JOURNALIST_TEMPLATES_DIR = \
+                _config.JOURNALIST_TEMPLATES_DIR  # type: str
+
+        if hasattr(_config, "NOUNS"):
+            self.NOUNS = _config.NOUNS  # type: str
+
+        if hasattr(_config, "SCRYPT_GPG_PEPPER"):
+            self.SCRYPT_GPG_PEPPER = _config.SCRYPT_GPG_PEPPER  # type: str
+
+        if hasattr(_config, "SCRYPT_ID_PEPPER"):
+            self.SCRYPT_ID_PEPPER = _config.SCRYPT_ID_PEPPER  # type: str
+
+        if hasattr(_config, "SCRYPT_PARAMS"):
+            self.SCRYPT_PARAMS = _config.SCRYPT_PARAMS  # type: Dict[str,int]
+
+        if hasattr(_config, "SECUREDROP_DATA_ROOT"):
+            self.SECUREDROP_DATA_ROOT = \
+                _config.SECUREDROP_DATA_ROOT  # type: str
+
+        if hasattr(_config, "SECUREDROP_ROOT"):
+            self.SECUREDROP_ROOT = _config.SECUREDROP_ROOT  # type: str
+
+        if hasattr(_config, "SESSION_EXPIRATION_MINUTES"):
+            self.SESSION_EXPIRATION_MINUTES = \
+                _config.SESSION_EXPIRATION_MINUTES  # type: ignore
+
+        if hasattr(_config, "SOURCE_TEMPLATES_DIR"):
+            self.SOURCE_TEMPLATES_DIR = \
+                _config.SOURCE_TEMPLATES_DIR   # type: str
+
+        if hasattr(_config, "STORE_DIR"):
+            self.STORE_DIR = _config.STORE_DIR  # type: str
+
+        if hasattr(_config, "SUPPORTED_LOCALES"):
+            self.SUPPORTED_LOCALES = _config.SUPPORTED_LOCALES  # type: ignore
+
+        if hasattr(_config, "TEMP_DIR"):
+            self.TEMP_DIR = _config.TEMP_DIR  # type: str
+
+        if hasattr(_config, "WORD_LIST"):
+            self.WORD_LIST = _config.WORD_LIST  # type: str
+
+        if hasattr(_config, "WORKER_PIDFILE"):
+            self.WORKER_PIDFILE = _config.WORKER_PIDFILE  # type: str
+
+        if hasattr(_config, "TRANSLATION_DIRS"):
+            self.TRANSLATION_DIRS = _config.TRANSLATION_DIRS  # type: ignore
+
+        if hasattr(_config, "env"):
+            self.env = _config.env  # type: str
 
 
 config = SDConfig()  # type: SDConfig
--- a/securedrop/sdconfig.py
+++ b/securedrop/sdconfig.py
@@ -6,7 +6,7 @@ import typing
 # https://www.python.org/dev/peps/pep-0484/#runtime-or-type-checking
 if typing.TYPE_CHECKING:
     # flake8 can not understand type annotation yet.
-    # That is why all type annotation realted import
+    # That is why all type annotation relative import
     # statements has to be marked as noqa.
     # http://flake8.pycqa.org/en/latest/user/error-codes.html?highlight=f401
     from typing import List, Dict  # noqa: F401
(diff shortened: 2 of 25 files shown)


securedrop/models.py
```diff
@@ -327,10 +327,12 @@ def set_hotp_secret(self, otp_secret):
 
     @property
     def totp(self):
+        # type: () -> str
```
heartsucker: These type annotations are incorrect. `TOTP` and `HOTP` return objects that have to be called to create or verify tokens.

securedrop/source_app/__init__.py
```diff
@@ -92,7 +96,7 @@ def handle_csrf_error(e):
     app.jinja_env.filters['filesizeformat'] = template_filters.filesizeformat
 
     for module in [main, info, api]:
-        app.register_blueprint(module.make_blueprint(config))
+        app.register_blueprint(module.make_blueprint(config))  # type: ignore
```
heartsucker: Does this not return `None`?

kushaldas:
This needs more work for fixing the lint errors. Will try that in some time.

securedrop/journalist_app/__init__.py
```diff
@@ -18,15 +18,21 @@
...
 
 
 def create_app(config):
+    # type: (SDConfig) -> Flask
     app = Flask(__name__,
                 template_folder=config.JOURNALIST_TEMPLATES_DIR,
                 static_folder=path.join(config.SECUREDROP_ROOT, 'static'))
 
     app.config.from_object(config.JournalistInterfaceFlaskConfig)
+    app._config = config
```
heartsucker: We should assign this to `sdconfig` because I feel like that is more pythonic since accessing "protected" members implies monkey-patchiness.

securedrop/sdconfig.py
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8 -*-
+
+import config as _config
+
+class SDConfig(object):
+    def __init__(self):
+        # type: () -> None
+        self.ADJECTIVES = _config.ADJECTIVES
+        self.DATABASE_ENGINE = _config.DATABASE_ENGINE
+        self.DATABASE_FILE = _config.DATABASE_FILE
```
heartsucker: This attribute is not always present and should be under the `else` of `if _config.ENGINE != 'sqlite'`.

securedrop/sdconfig.py
```diff
@@ -0,0 +1,50 @@
...
+        if _config.DATABASE_ENGINE != "sqlite":
+            self.DATABASE_USERNAME = _config.DATABASE_USERNAME  # type: ignore
+            self.DATABASE_PASSWORSD = _config.DATABASE_PASSWORD   # type: ignore
+            self.DATABASE_HOST = _config.DATABASE_HOST  # type: ignore
+            self.DATABASE_NAME = _config.DATABASE_NAME  # type: ignore
+            
+        if hasattr(_config, 'CUSTOM_HEADER_IMAGE'):
+            self.CUSTOM_HEADER_IMAGE = _config.CUSTOM_HEADER_IMAGE  # type: ignore
+
+config = SDConfig()  # type: SDConfig
```
heartsucker: Nit: missing newline at EOF.

securedrop/models.py
kushaldas: Ah, these went in by mistake :(

securedrop/source_app/__init__.py
kushaldas: mypy/typing still can not handle module level typing that well as it sees the list as `List[ModuleType]` and `ModuleType` does not have that function `make_blueprint`.

securedrop/sdconfig.py
kushaldas: Ah, means it will be there as  empty string as default value.

securedrop/sdconfig.py
kushaldas: Yup, will make these changes.

securedrop/sdconfig.py
heartsucker: Actually i was wrong. This field will be present and have a valid value because `config.py` loads it correctly as `path.join(SECUREDROP_DATA_ROOT, 'db.sqlite')`. But we should still only load it if the engine is `sqlite`.

securedrop/source_app/__init__.py
heartsucker: :+1:

securedrop/journalist_app/__init__.py
kushaldas: Can we fix this in a future PR?

heartsucker: We already renamed everything in this PR so I feel like we might as well rename it right the first time if we agree that it makes more sense to name it `sdconfig`.

heartsucker:
Test error:

```
    TASK [build-securedrop-app-code-deb-pkg : Compile PO to MO.] *******************
    fatal: [build]: FAILED! => {"changed": true, "cmd": "cp config.py.example config.py ; trap 'rm config.py' EXIT ; ./manage.py --verbose translate-messages --compile", "delta": "0:00:00.321362", "end": "2018-02-19 16:29:58.608011", "msg": "non-zero return code", "rc": 1, "start": "2018-02-19 16:29:58.286649", "stderr": "Traceback (most recent call last):\n  File \"./manage.py\", line 24, in <module>\n    from sdconfig import config\n  File \"/tmp/securedrop-app-code-0.5.2-amd64/var/www/securedrop/sdconfig.py\", line 57, in <module>\n    config = SDConfig()  # type: SDConfig\n  File \"/tmp/securedrop-app-code-0.5.2-amd64/var/www/securedrop/sdconfig.py\", line 27, in __init__\n    self.SUPPORTED_LOCALES = _config.SUPPORTED_LOCALES\nAttributeError: 'module' object has no attribute 'SUPPORTED_LOCALES'", "stderr_lines": ["Traceback (most recent call last):", "  File \"./manage.py\", line 24, in <module>", "    from sdconfig import config", "  File \"/tmp/securedrop-app-code-0.5.2-amd64/var/www/securedrop/sdconfig.py\", line 57, in <module>", "    config = SDConfig()  # type: SDConfig", "  File \"/tmp/securedrop-app-code-0.5.2-amd64/var/www/securedrop/sdconfig.py\", line 27, in __init__", "    self.SUPPORTED_LOCALES = _config.SUPPORTED_LOCALES", "AttributeError: 'module' object has no attribute 'SUPPORTED_LOCALES'"], "stdout": "", "stdout_lines": []}
```

securedrop/tests/test_journalist.py
```diff
@@ -1207,9 +1206,7 @@ def test_journalist_session_expiration(self):
                         resp.data.decode('utf-8'))
         finally:
             if has_session_expiration:
-                config.SESSION_EXPIRATION_MINUTES = old_expiration
```
dachary: The deletion of the the two lines below need to be explained in detail. In particular it should be clearly articulated in the commit message why these two lines become redundant and exactly which part of the code using it exist and how they are not impacted by this change. I'm sure you did your due diligence but it needs to be spelled out for the reader.

heartsucker:
Also, you're missing the AppArmor rules for `sdconfig.py`.

kushaldas:
>* a detailed analysis of each existing config.py variable and how it is taken into account in sdconfig.py. It is worth the extra effort because config.py is such a central piece in SecureDrop and even the slightest mistake could have epic proportions.

I made the `SDConfig` class based on all the use cases I found in the source at the `securedrop` directory. I hope that is what you meant by analysis of each variable. If you know about any special corner case, please comment and I will update the PR accordingly.

dachary:
On example of things I'm worried about is: if anywhere in the code or the tests a config variable is modified (that should not be the case but ...) it will no longer be modified in config.py, it will be modified in the copy kept in the SDConfig class instance. In order to verify this never is the case, there would be a need to check everywhere a config.py variable is used to visually verify this is indeed not a concern.

Since the tests are passing we know nothing obviously bad happened. But the tests do not cover every aspect of the code. When reviewing a PR we rely on the tests but also on our ability to check the code impacted by the change. In the case of the config.py refactor, we would need to check a very large amount of code to be sure nothing was overlooked.

Please don't get me wrong: I like the fact that we're trying to move to a class instead of global variables. I'm just saying this should be done very carefully. And every decision must be clearly explained. For instance, why is it necessary to have a copy of the os module in the SDConfig class ?

dachary:
Here is an idea: why not make the SDConfig class a simple wrapper around whatever exists in config.py ? Instead of copying what it contains, define a __getattr__ function that returns the matching config.py variable ? So we know it is transparent. And define __setattr__ to raise an exception as we do **not** want any part of the code modifying what config.py contains.

securedrop/journalist_app/__init__.py
```diff
@@ -18,15 +18,21 @@
 from models import Journalist
 from store import Storage
 
+MYPY = False
+if MYPY:
```
dachary: A comment explaining the F401 exception is because of mypy would be better. if MYPY is too much of a hack.

securedrop/sdconfig.py
```diff
@@ -0,0 +1,65 @@
...
+    def __init__(self):
+        # type: () -> None
+        self.ADJECTIVES = _config.ADJECTIVES
+        self.DATABASE_ENGINE = _config.DATABASE_ENGINE
+        self.DEFAULT_LOCALE = _config.DEFAULT_LOCALE
+        self.FlaskConfig = _config.FlaskConfig
+        self.GPG_KEY_DIR = _config.GPG_KEY_DIR
+        self.JOURNALIST_KEY = _config.JOURNALIST_KEY
+        self.JOURNALIST_TEMPLATES_DIR = _config.JOURNALIST_TEMPLATES_DIR
+        self.JournalistInterfaceFlaskConfig = \
```
dachary: this is surprising, why is it necessary ? and the same question for self.os and all other modules.

securedrop/sdconfig.py
```diff
@@ -0,0 +1,65 @@
...
+        self.TEMP_DIR = _config.TEMP_DIR
+        self.WORD_LIST = _config.WORD_LIST
+        self.WORKER_PIDFILE = _config.WORKER_PIDFILE
+        self.env = _config.env
+        self.os = _config.os
+        # Below 4 attributes are only for non-sqlite env
+        self.DATABASE_USERNAME = ""
+        self.DATABASE_PASSWORD = ""
+        self.DATABASE_HOST = ""
+        self.DATABASE_NAME = ""
```
dachary: it would be better to have a \_\_setattr__ function that getattr(_config, attr) so the class is transparent. And it should **only** allow getting attributes that are upppercase only and underscore. To eliminate any places in the code that try to access something different from config.py

securedrop/journalist_app/__init__.py
kushaldas: This is the formal way suggested by the mypy team for Python2.

securedrop/sdconfig.py
kushaldas: This is required because we can not dynamically add any random attribute to a class when we are also doing type checks on the class. This is the reason to create the `SDConfig` class.

kushaldas:
> Here is an idea: why not make the SDConfig class a simple wrapper around whatever exists in config.py ? Instead of copying what it contains, define a getattr function that returns the matching config.py variable ? So we know it is transparent. And define setattr to raise an exception as we do not want any part of the code modifying what config.py contains.

Explicit is better than implicit, and as I said in other comments, the current way will help us to add more detailed type checks in future.

kushaldas:
> And every decision must be clearly explained. For instance, why is it necessary to have a copy of the os module in the SDConfig class ?

I kept it to make sure that I don't break any code accidentally. This way it is same like the `config` module and existing code will not break.

securedrop/journalist_app/__init__.py
kushaldas: I will add the details on the same in the development docs, on how to add type annotation in our Python2 code.

heartsucker:
> Here is an idea: why not make the SDConfig class a simple wrapper around whatever exists in config.py ? Instead of copying what it contains, define a __getattr__ function that returns the matching config.py variable ? So we know it is transparent. And define __setattr__ to raise an exception as we do not want any part of the code modifying what config.py contains.

If we do that, we would need to allow `__setattr__` to work if and only if we're in test mode (`os.environ['SECUREDROP_ENV'] == 'test'). Some tests require modification of the variables to work correctly.

heartsucker:
Also @kushaldas I think one of the worries is that some of the variables will break like how in the test suite we broke because `config.SUPPORTED_LOCALES` isn't always present. I think we might need to check more cases to make sure we're not raising any `AttributeError`s.

dachary:
@kushaldas I don't think adding type annotation is incentive enough to refactor config.py in a more fragile way. Surely there must be a way to add type annotation for class attributes that are dynamically created by \_\_getattr__: it is a common use case. I suspect it also is possible to declare the type of a global variable so it can be used to validate functions that make use of them and that would allow you to add type annotations to config.py directly instead of refactoring it. If none of these options are possible with mypy, that makes me wonder if it is mature enough to be used.

kushaldas:
>Also @kushaldas I think one of the worries is that some of the variables will break like how in the test suite we broke because config.SUPPORTED_LOCALES isn't always present. I think we might need to check more cases to make sure we're not raising any AttributeErrors.

That is why we have all the config attributes available by default, the tests are modifying them as required.

```
$ rg config.SUPPORTED_LOCALES
securedrop/tests/test_i18n.py
241:        fake_config.SUPPORTED_LOCALES = [
246:            assert i18n.LOCALES == fake_config.SUPPORTED_LOCALES
287:        fake_config.SUPPORTED_LOCALES = ['fr_FR', 'en_US']

securedrop/tests/test_template_filters.py
114:        fake_config.SUPPORTED_LOCALES = ['en_US', 'fr_FR']
118:            assert i18n.LOCALES == fake_config.SUPPORTED_LOCALES

securedrop/tests/test_journalist.py
1326:        fake_config.SUPPORTED_LOCALES = ['en_US', 'fr_FR']
```

In future we can modify the code to check the existence of any class attribute as standard `True` value check , that is:

```
if config.SUPPORTED_LOCALES:
```

kushaldas:
As an example:

```
self.SUPPORTED_LOCALES = []  # type: List[str]
```

The above declaration will make sure that in our code we will not be able to assign a string to the supported locales by mistake (mypy will catch that).

kushaldas:
@heartsucker @dachary To have  proper type annotation, we are giving up on many fancy dynamic way which Python supports, but at the same time, we will have much better type safety before seeing run time errors.

kushaldas:
>@kushaldas I don't think adding type annotation is incentive enough to refactor config.py in a more fragile way. Surely there must be a way to add type annotation for class attributes that are dynamically created by __getattr__: it is a common use case. I suspect it also is possible to declare the type of a global variable so it can be used to validate functions that make use of them and that would allow you to add type annotations to config.py directly instead of refactoring it. If none of these options are possible with mypy, that makes me wonder if it is mature enough to be used.

I am going to argue that having a proper class and defined attributes is a better, explicit and stable way than having a module whose attributes can change at run time. Just show the SDConfig definition to any new Python programmer, they will be able to tell you exactly what all attributes our configuration object will have. 
In the case of current `config.py`, that is not the case. 

As far stability of `mypy`, Dropbox has more than 100K lines of code annotated, I forgot the line count of Instagram though.

securedrop/sdconfig.py
```diff
@@ -0,0 +1,65 @@
...
+        self.JOURNALIST_TEMPLATES_DIR = _config.JOURNALIST_TEMPLATES_DIR
+        self.JournalistInterfaceFlaskConfig = \
+            _config.JournalistInterfaceFlaskConfig
+        self.NOUNS = _config.NOUNS
+        self.SCRYPT_GPG_PEPPER = _config.SCRYPT_GPG_PEPPER
+        self.SCRYPT_ID_PEPPER = _config.SCRYPT_ID_PEPPER
+        self.SCRYPT_PARAMS = _config.SCRYPT_PARAMS
+        self.SECUREDROP_DATA_ROOT = _config.SECUREDROP_DATA_ROOT
+        self.SECUREDROP_ROOT = _config.SECUREDROP_ROOT
+        self.SESSION_EXPIRATION_MINUTES = _config.SESSION_EXPIRATION_MINUTES
```
redshiftzero: Note: `SESSION_EXPIRATION_MINUTES` was recently added and may not be defined in many existing `config.py`

kushaldas: Okay, I will give another pass on this file and make all of them as default and then if available in `config.py`, then only add here.

securedrop/sdconfig.py
```diff
@@ -0,0 +1,65 @@
...
+if MYPY:
+    from typing import List  # noqa: F401
+
+
+class SDConfig(object):
+    def __init__(self):
+        # type: () -> None
+        self.ADJECTIVES = _config.ADJECTIVES
+        self.DATABASE_ENGINE = _config.DATABASE_ENGINE
+        self.DEFAULT_LOCALE = _config.DEFAULT_LOCALE
```
redshiftzero: `DEFAULT_LOCALE` may not be defined in instances installed prior to 0.4.3

kushaldas: Yes, this is taken care of in the current update.

securedrop/sdconfig.py
```diff
@@ -0,0 +1,83 @@
...
+        # static configuration file, we should remove any dynamic
+        # attribution set/get code like below.
+        attributes = [
+            'ADJECTIVES', 'DATABASE_ENGINE', 'DATABASE_FILE', 'DEFAULT_LOCALE',
+            'GPG_KEY_DIR', 'JOURNALIST_KEY', 'JOURNALIST_TEMPLATES_DIR',
+            'NOUNS', 'SCRYPT_GPG_PEPPER', 'SCRYPT_ID_PEPPER', 'SCRYPT_PARAMS',
+            'SECUREDROP_DATA_ROOT', 'SECUREDROP_ROOT',
+            'SESSION_EXPIRATION_MINUTES', 'SOURCE_TEMPLATES_DIR', 'STORE_DIR',
+            'SUPPORTED_LOCALES', 'TEMP_DIR', 'WORD_LIST', 'WORKER_PIDFILE',
+            'TRANSLATION_DIRS', 'env', 'os'
```
redshiftzero: Why is `os` included in this list?

kushaldas: Because that will keep things exactly same in current `config.py`. I did not want to make any random breaking changes in this PR. We can check for if there is random usage of these modules, and then remove them later in another PR.

That list of items were made by another small script I wrote auto-generate.

securedrop/sdconfig.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(self):
+        # type: () -> None
+        self.FlaskConfig = _config.FlaskConfig
+        self.JournalistInterfaceFlaskConfig = \
+            _config.JournalistInterfaceFlaskConfig
+        self.SourceInterfaceFlaskConfig = _config.SourceInterfaceFlaskConfig
+
+        self.ADJECTIVES = ""
+        self.DATABASE_ENGINE = ""
+        self.DEFAULT_LOCALE = ""
```
redshiftzero: You'll want to set this to `en_US` as a default since the current logic in the application is:

```
getattr(config, 'DEFAULT_LOCALE', 'en_US')
```

to handle the case where `DEFAULT_LOCALE` was not set in existing instances

securedrop/sdconfig.py
redshiftzero: But `os` is just a module used by `config.py` - is there a possible path where removing `os` from this list could conceivably produce a breaking change?

securedrop/journalist_app/__init__.py
redshiftzero: I don't understand why we have this `MYPY = False` here - can we add a comment explaining why this is the case? To the code or in the commit message

securedrop/sdconfig.py
dachary: I think all attributes that are modules (os, etc.) must be removed.

securedrop/sdconfig.py
dachary: I feel it is wrong to write code that is not used (because getattr will actually lookup something else) for the sole purpose of indirectly allowing type checking. It is a hack. If there must be a hack to allow mypy to do its job, it needs to be outside of the code base, not inside the codebase.

securedrop/source_app/__init__.py
```diff
@@ -79,7 +84,7 @@ def handle_csrf_error(e):
     app.jinja_env.trim_blocks = True
     app.jinja_env.lstrip_blocks = True
     app.jinja_env.globals['version'] = version.__version__
-    if getattr(config, 'CUSTOM_HEADER_IMAGE', None):
+    if config.CUSTOM_HEADER_IMAGE:
```
dachary: to keep things simple, SDConfig should not automagically change the semantic of attributes (in this case it means CUSTOM_HEADER_IMAGE exists even if not defined in config.py).

securedrop/sdconfig.py
kushaldas: Okay.

securedrop/sdconfig.py
kushaldas: @dachary Yes, doing it in the next commit.

kushaldas:
>But os is just a module used by config.py - is there a possible path where removing os from this list could conceivably produce a breaking change?

Just now checked removing it, no breakage, so will update that.

>I don't understand why we have this MYPY = False here - can we add a comment explaining why this is the case? To the code or in the commit message

I will update the code with newer *TYPE_CHECKING* value and point to the PEP in the code as a comment.
The `if MYPY:` was the official hack to handle things, now so many people used it that it became part of the PEP :)

dachary:
> I will update the code with newer TYPE_CHECKING 

As per https://www.python.org/dev/peps/pep-0484/#runtime-or-type-checking. Having:

<pre>
import typing

if typing.TYPE_CHECKING:
    import expensive_mod
</pre>

To guard all imports that are only meaningful in the context of type checking is a clean and explicit way to do it, :100: :-)

securedrop/source_app/__init__.py
kushaldas: This is related to [Truth value testing](https://docs.python.org/3/library/stdtypes.html#truth-value-testing). 

We know that there is an attribute of our config named `CUSTOM_HEADER_IMAGE`, but  use it only if it has a `Truth` value, means someone set it up  using config file.

In this PR I also clearing up the related code which are doing getattr or setattr (except the test cases).  Pushing another update in some time.

dachary: such a change must be in a separate commit, with the explained rationale and benefits so it can be properly reviewed. Changing the semantics in the same commit a refactor is done is harder to review and may lead to subtle errors.

kushaldas: Correct, but this has to be inside of that commit. As if we kept the old `hasattr` call, it would find the `CUSTOM_HEADER_IMAGE` all the time, and the would execute the next two lines. 

Now, it will only do if there is a proper value. The rest of any such calls are replaced in a newer commit in the PR.

securedrop/sdconfig.py
kushaldas: There is no more `getattr` in any piece of code, one of the latest commit actually removed all such dynamic attribution handling. See https://github.com/freedomofpress/securedrop/pull/3033/commits/0f55e8fa50373fe36e599dd199ecef8fbcbf52d1

dachary: ah, so I'm back to my original suggestion https://github.com/freedomofpress/securedrop/pull/3033#discussion_r169270000

dachary:
@kushaldas as discussed on gitter I second the idea of have a SDConfig contructor that does the following for each knonw variable in config.py

```
if hasattr(_config, SOME_THING):
    self.SOME_THING = _config.SOME_THING # type: str
```
and the SDConfig class would have no other method. It would effectively make it transparent, robust and mypy compatible. Kudos to @heartsucker for suggesting this in the first place.

securedrop/sdconfig.py
```diff
@@ -0,0 +1,112 @@
...
+
+        if _config.DATABASE_ENGINE != "sqlite":
+            self.DATABASE_USERNAME = _config.DATABASE_USERNAME  # type: ignore
+            self.DATABASE_PASSWORD = \
+                _config.DATABASE_PASSWORD   # type: ignore
+            self.DATABASE_HOST = _config.DATABASE_HOST  # type: ignore
+            self.DATABASE_NAME = _config.DATABASE_NAME  # type: ignore
+        else:
+            self.DATABASE_FILE = _config.DATABASE_FILE
+
```
dachary: please remove the lines above, they are not necessary

securedrop/sdconfig.py
```diff
@@ -0,0 +1,112 @@
...
+    # That is why all type annotation realted import
+    # statements has to be marked as noqa.
+    # http://flake8.pycqa.org/en/latest/user/error-codes.html?highlight=f401
+    from typing import List, Dict  # noqa: F401
+
+
+class SDConfig(object):
+    def __init__(self):
+        # type: () -> None
+        self.FlaskConfig = _config.FlaskConfig
```
heartsucker: I think we should use the `hasattr` guard for every single member of the `config.py` module. This covers us to be dead sure that we're not going to raise exceptions at runtime on a prod instance.

securedrop/sdconfig.py
```diff
@@ -0,0 +1,112 @@
...
+            self.ADJECTIVES = _config.ADJECTIVES  # type: str
+
+        if hasattr(_config, "DATABASE_ENGINE"):
+            self.DATABASE_ENGINE = _config.DATABASE_ENGINE  # type: str
+
+        if hasattr(_config, "DATABASE_FILE"):
+            self.DATABASE_FILE = _config.DATABASE_FILE  # type: str
+
+        if hasattr(_config, "DEFAULT_LOCALE"):
+            self.DEFAULT_LOCALE = _config.DEFAULT_LOCALE  # type: ignore
```
redshiftzero: I'm curious why we have a few variables that have `# type: ignore`? If this has been answered already and I missed it, we should probably explain in the commit message

securedrop/sdconfig.py
```diff
@@ -0,0 +1,112 @@
...
+            self.DATABASE_HOST = _config.DATABASE_HOST  # type: ignore
+            self.DATABASE_NAME = _config.DATABASE_NAME  # type: ignore
+        else:
+            self.DATABASE_FILE = _config.DATABASE_FILE
+
+        if hasattr(_config, 'CUSTOM_HEADER_IMAGE'):
+            self.CUSTOM_HEADER_IMAGE = \
+                _config.CUSTOM_HEADER_IMAGE  # type: ignore
+
+        if hasattr(_config, 'SUPPORTED_LOCALES'):
```
redshiftzero: In Python 2 `hasattr` will return false on any exception - not just an `AttributeError` as one might expect, so an alternative here that does not swallow exceptions (since this can create debugging issues in the future) would be to adopt an [EAFP](https://blogs.msdn.microsoft.com/pythonengineering/2016/06/29/idiomatic-python-eafp-versus-lbyl/) approach to handle the `AttributeError` only (as shown [in this blog](https://hynek.me/articles/hasattr/) - kudos to @kushaldas for pointing me to this). What do you think @dachary @heartsucker?

dachary: > Which is almost never what you want because it shadows errors in properties:

Although this does not apply here given the simplistic nature of config.py as we have it now, I must confess I did not know at all about this. And I'm happy to have been educated, thanks @kushaldas 

I'm 100% ok to replace all hasattr in this function with try / except even though it is longer

securedrop/sdconfig.py
```diff
@@ -0,0 +1,112 @@
+# -*- coding: utf-8 -*-
+
+import config as _config
+
+import typing
+# https://www.python.org/dev/peps/pep-0484/#runtime-or-type-checking
+if typing.TYPE_CHECKING:
+    # flake8 can not understand type annotation yet.
+    # That is why all type annotation realted import
```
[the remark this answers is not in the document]
kushaldas: Will be fixed.

securedrop/sdconfig.py
kushaldas: I will add that in the commit message. They are ignored because just in case the attribute is missing (which we expect) `mypy` will complain. That is why now all the attributes have that. We will improve these in future (after 0.6).

securedrop/sdconfig.py
```diff
@@ -0,0 +1,174 @@
...
+    # statements has to be marked as noqa.
+    # http://flake8.pycqa.org/en/latest/user/error-codes.html?highlight=f401
+    from typing import List, Dict  # noqa: F401
+
+
+class SDConfig(object):
+    def __init__(self):
+        # type: () -> None
+        try:
+            self.FlaskConfig = _config.FlaskConfig  # type: ignore
```
dachary: please do not include modules, only variables

securedrop/sdconfig.py
```diff
@@ -0,0 +1,174 @@
...
+        except AttributeError:
+            pass
+
+        try:
+            self.JournalistInterfaceFlaskConfig = \
```
dachary: please do not include modules, only variables

securedrop/sdconfig.py
```diff
@@ -0,0 +1,174 @@
...
+                _config.JournalistInterfaceFlaskConfig  # type: ignore
+        except AttributeError:
+            pass
+
+        try:
+            self.SourceInterfaceFlaskConfig = \
```
dachary: please do not include modules, only variables

securedrop/sdconfig.py
```diff
@@ -0,0 +1,174 @@
...
+        except AttributeError:
+            pass
+
+        try:
+            self.TRANSLATION_DIRS = _config.TRANSLATION_DIRS  # type: ignore
+        except AttributeError:
+            pass
+
+        try:
+            self.env = _config.env  # type: ignore
```
dachary: please do not include modules, only variables

securedrop/tests/test_journalist.py
```diff
@@ -53,8 +53,7 @@ def tearDown(self):
     @patch('crypto_util.CryptoUtil.genrandomid',
            side_effect=['bad', VALID_PASSWORD])
     def test_make_password(self, mocked_pw_gen):
-        class fake_config:
-            pass
+        fake_config = SDConfig()
```
dachary: please remove this line. If it was necessary it should be **fake_config = SDConfig** (i.e. without the parenthesis). It does not fail because the call does not actually try to use it. But it is a type violation of some kind and the purpose of this PR is precisely to help with type violations so... we don't want that ;-)

securedrop/tests/test_journalist.py
```diff
@@ -53,8 +53,7 @@ def tearDown(self):
...
         assert (journalist_app.utils.make_password(fake_config) ==
```
dachary: replace fake_config with SDConfig

securedrop/tests/test_journalist.py
```diff
@@ -1315,10 +1314,7 @@ def tearDown(self):
         utils.env.teardown()
 
     def get_fake_config(self):
-        class Config:
-            def __getattr__(self, name):
-                return getattr(config, name)
-        return Config()
+        return SDConfig()
```
dachary: thanks to the new **SDConfig** there is no need for **get_fake_config(self)** and all calls to **get_fake_config()** can be replaced with **SDConfig()** instead :+1:

securedrop/tests/test_template_filters.py
```diff
@@ -19,10 +19,7 @@
 class TestTemplateFilters(object):
 
     def get_fake_config(self):
-        class Config:
-            def __getattr__(self, name):
-                return getattr(config, name)
-        return Config()
+        return SDConfig()
```
dachary: thanks to the new **SDConfig** there is no need for **get_fake_config(self)** and all calls to **get_fake_config()** can be replaced with **SDConfig()** instead :+1:

securedrop/tests/test_i18n.py
```diff
@@ -56,10 +56,7 @@ def tearDown(self):
         self.__context.pop()
 
     def get_fake_config(self):
-        class Config:
-            def __getattr__(self, name):
-                return getattr(config, name)
-        return Config()
+        return SDConfig()
```
dachary: thanks to the new **SDConfig** there is no need for **get_fake_config(self)** and all calls to **get_fake_config()** can be replaced with **SDConfig()** instead :+1:

securedrop/sdconfig.py
heartsucker: Why not? These modules impact how the app behaves and we need to be able to inject them in tests.

dachary: Could you please show me where in the code it would make a difference? I honestly thought it was entirely unecessary.

kushaldas: @dachary That is not a module, it is class which is part of current `config.py`.

securedrop/sdconfig.py
kushaldas: @dachary That is not a module, it is class which is part of current `config.py`. https://github.com/freedomofpress/securedrop/blob/develop/securedrop/source_app/__init__.py#L30

securedrop/sdconfig.py
kushaldas: https://github.com/freedomofpress/securedrop/blob/develop/securedrop/journalist_app/__init__.py#L29

@dachary That is not a module, it is class which is part of current `config.py`.

securedrop/sdconfig.py
kushaldas: @dachary `env` is not a module, it is a variable of type `str`.

securedrop/tests/test_journalist.py
kushaldas: I will not remove these as the current PR brings in only the limited changes required to have an API level similar thing by replacing `config` module with `SDConfig` class. We can update the tests in a future PR where we update how config works.

securedrop/tests/test_journalist.py
kushaldas: I will update the tests in a new PR where will change how we deal with configuration.

securedrop/tests/test_template_filters.py
kushaldas: I will update the tests in a new PR where will change how we deal with configuration. This PR is keeping things in API level same, and we should not update any tests, that will confirm that our old will work in the same way with the new `SDConfig`.

securedrop/tests/test_i18n.py
kushaldas: I will update the tests in a new PR where will change how we deal with configuration. This PR is keeping things in API level same, and we should not update any tests, that will confirm that our old will work in the same way with the new `SDConfig`.

securedrop/sdconfig.py
dachary: correct, we need to keep that

securedrop/sdconfig.py
dachary: correct, but it is only used internally and not anywhere else in the code so let's not needlessly expose it

securedrop/sdconfig.py
dachary: correct, we need to keep that

securedrop/sdconfig.py
dachary: correct, but it is only used internally and not anywhere else in the code so let's not needlessly expose it

securedrop/tests/test_journalist.py
dachary: ok. Then you need to change this to **fake_config = SDConfig** to be exact (i.e. without the parenthesis).

securedrop/tests/test_journalist.py
dachary: ok

securedrop/tests/test_template_filters.py
dachary: ok

securedrop/tests/test_i18n.py
dachary: ok

securedrop/tests/test_journalist.py
kushaldas: I should not, this test was broken in some way. Let me explain it below.

```
>>> class fake_config:
...     pass
... 
>>> type(fake_config)
<type 'classobj'>
```

As you can see above that `fake_cofnig` is actually a variable of type `classobj` (or in other words, it is a class). But, if you read the implementation of `[make_password](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/journalist_app/utils.py#L201)` you will see that actually wants an object in the argument named `config`. Because of the dynamic dance with has/get attr calls, this never failed. But, we should not pass around a `class` variable where our code expects an object with some certain attributes.

This is another good example why we should defined attributes in our code base and should not try to handle things in random dynamic way. Maybe in future we will start adding type annotation to our tests too.

dachary: Oh, so this actually is a bug fix. It is worth separating bug fixes from refactor (in separate commits). I commented assuming the purpose of the change was to replace fake_config which is the focus of all other changes in this commit, hence my confusion. All good then.

securedrop/sdconfig.py
heartsucker: I think we should keep everything exposed exactly as it was in this PR and then a second one to restructure it.

heartsucker: Wait, sorry. You're right. This one is just a subclass of the things we use. I misunderstood you. Yes, this one should not be part of `SdConfig`.

securedrop/sdconfig.py
heartsucker: It's used for starting the app in the development environment and as a safeguard against listening on `0.0.0.0` in prod.

dachary: ah, right it's
<pre>
    debug = getattr(config, 'env', 'prod') != 'prod'
</pre>
in journalist.py and source.py. It is worth a comment to point to these. I looked for `config.env` for usage and did not find anything, reason for my remark.

securedrop/tests/test_journalist.py
kushaldas: >Oh, so this actually is a bug fix. It is worth separating bug fixes from refactor (in separate commits). I commented assuming the purpose of the change was to replace fake_config which is the focus of all other changes in this commit, hence my confusion. All good then.

Correct, this bug fix was not intentional, but just using the new `SDConfig` solved that bug, that is why it is here. I understand your idea about keeping the PRs separate.

securedrop/sdconfig.py
kushaldas: @heartsucker @dachary Okay, will remove that one.

securedrop/sdconfig.py
kushaldas: ^^ this is the exact reason I don't want any dynamic handling of attributes in the code base :+1:

dachary:
@kushaldas did you investigate why this PR looses 2% in code coverage ?

heartsucker:
> @kushaldas did you investigate why this PR looses 2% in code coverage ?

Likely because all the `except AttributeError`'s are never triggered.

securedrop/sdconfig.py
dachary: Pleaes add the following above `self.env` to avoid similar confusion when reading the code (because this discussion is unlikely to be linked):
<pre>
# although config.env is intended to be private, it needs to be exposed for journalist.py / source.py to know if it is running in production or not.  
</pre>

dachary:
`securedrop/journalist_app/__init__.py` lost 2% according to codecov although there only is one line not covered. It's probably nothing but it's worth checking, otherwise what's the point of having coverage reports ?

redshiftzero:
Looks like we're getting a coverage decrease (for the interested observer, I determined this by examining the coverage results from `make test` on `develop` and this branch) in the existing files as the following code block inside the if statement will not be executed in `journalist_app/__init__.py`, `crypto_util.py`, `source_app/__init__.py`, `journalist_app/utils.py`:

```
if typing.TYPE_CHECKING:
  from sdconfig import SDConfig  # noqa: F401
```

securedrop/sdconfig.py
```diff
@@ -0,0 +1,169 @@
...
+
+        try:
+            self.DATABASE_ENGINE = _config.DATABASE_ENGINE  # type: ignore
+        except AttributeError:
+            pass
+
+        try:
+            self.DATABASE_FILE = _config.DATABASE_FILE  # type: ignore
+        except AttributeError:
+            pass
```
redshiftzero: It looks like this is a duplicate - `DATABASE_FILE` is defined above

dachary:
Unless there is a super urgent need to merge this 3 days old pull request who beats all records of attention in history, I'd like to get a chance to review it one last time after all is addressed, during the week-end maybe.

redshiftzero:
(Btw for the interested observer - `CUSTOM_NOTIFICATION` and `ORG_NAME` are config options that never shipped, and `JOURNALIST_PIDFILE` and `SOURCE_PIDFILE` were removed as they were unused - see #1822 and #1821 for the full history on that)

dachary:
I propose we wait another 48h before merging this (i.e. after the 0.6 branch is cut) so we have plenty of time to address whatever we may have missed during the 0.7 cycle. That will also give an opportunity to add the [comment I suggested](https://github.com/freedomofpress/securedrop/pull/3033/files#r170091763).

redshiftzero:
Since we've branched off for SecureDrop 0.6, merging this in...

PR #3035, opened 2018-02-20
Fixes #2114 Adds clear caution messages

### LINKED ISSUE #2114: ./securedrop-admin tailsconfig fails if parent not named "securedrop"
# Bug

## Description

Running `tailsconfig` on a workstation Tails fails unless the directory has a particular name. This doesn't seem to be documented.

## Steps to Reproduce

Place copy of securedrop repo in `~/Persistent` of a properly-configured (dotfiles etc) Tails 3.1 instance running with an admin password enabled. Ensure repo is in dir named `securedrop-develop`, as it is named by default if you downloaded a zip file of the repo from github and unzipped it (not the recommended approach, more below). Copy your working `app-journalist-aths` and `app-source-ths` into `securedrop-develop/install_files/ansible-base`. `cd` into the repo dir and run `./securedrop-admin setup` and then `./securedrop-admin tailsconfig`. Enter admin password as prompted.

## Expected Behavior

Install works, just as it would if the dir were named "securedrop".

## Actual Behavior

 Install fails, see 
[securedrop-install-error.txt](https://github.com/freedomofpress/securedrop/files/1226552/securedrop-install-error.txt). I tried this twice with each directory name and the exact same config files and steps to ensure I wasn't borking things with anything other than the dir name. Always worked in `~/Persistent/securedrop` and failed in `~/Persistent/securedrop-develop`.

The use of a hardwired dir name seems to emerge in the session like so
```
TASK [tails-config : Migrate site-specific configuration to new location.] *****
fatal: [localhost]: FAILED! => {"changed": true, "cmd": ["cp", "/home/amnesia/Persistent/securedrop/install_files/ansible-base/prod-specific.yml", "/home/amnesia/Persistent/securedrop/install_files/ansible-base/group_vars/all/site-specific"], "delta": "0:00:00.002211", "end": "2017-08-15 21:20:57.049969", "failed": true, "rc": 1, "start": "2017-08-15 21:20:57.047758", "stderr": "cp: cannot stat '/home/amnesia/Persistent/securedrop/install_files/ansible-base/prod-specific.yml': No such file or directory", "stdout": "", "stdout_lines": [], "warnings": []}
```

## Comments

People shouldn't be downloading the zip file, they should be using the git install, so they can use the inbuilt gpg verification and check out a particular tag.  If they do that, and don't rename anything, this won't be an issue. I ultimately did this.

That said, it's never explicitly documented that the dir must be named `securedrop`. That should probably be stated explicitly here:

https://docs.securedrop.org/en/stable/set_up_admin_tails.html#download-the-securedrop-repository

And/or here:

https://docs.securedrop.org/en/stable/onboarding.html

OR the code should be fixed to not depend on a particular dir name. IMO.

kushaldas:
We should warn the administrators to only use the given git
commands to download the SecureDrop Git repository. This also
adds another caution message about not skipping the GPG
verification step.

## Status

Ready for review

## Description of Changes

Fixes #2114 

This adds two more caution blocks in the docs.

## Testing

`make docs` and read the updated `set_up_admin_tails.html#download-the-securedrop-repository` section.

### DIFF
--- a/docs/set_up_admin_tails.rst
+++ b/docs/set_up_admin_tails.rst
@@ -53,9 +53,16 @@ Start by running the following commands to download the git repository.
 .. note:: Since the repository is fairly large and Tor can be slow,
 	  this may take a few minutes.
 
+.. caution:: Do not download SecureDrop Git repository as a Zip file,
+             or any other means. Only download by using the given git
+             command.
+
 
 Verify the Release Tag
 ~~~~~~~~~~~~~~~~~~~~~~
+
+.. caution:: Do not skip this step as this steps validates the files
+             in your Git repository.
 
 First, download and verify the **SecureDrop Release Signing Key**.
 


kushaldas:
Once again missed to give the right branch name for the docs patches :(

PR #3039, opened 2018-02-20
tests: use TAG when possible and default to git rev-parse HEAD

dachary:
## Status

Ready for review

## Description of Changes

It is important to obey TAG in CircleCI because it will *not* always
match git rev-parse HEAD and will lead to inconsistencies.

## Testing

* make ci-lint locally to verify the defaults work
* the CircleCI should always succeed (difficult to verify except by running multiple times)

## Deployment

N/A

### DIFF
--- a/devops/scripts/dev-shell-ci
+++ b/devops/scripts/dev-shell-ci
@@ -7,6 +7,7 @@ set -eu
 source "${BASH_SOURCE%/*}/ticker"
 
 TOPLEVEL=$(git rev-parse --show-toplevel)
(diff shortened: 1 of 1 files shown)


dachary:
looks like a transient error that would require better timing guards in functional tests https://circleci.com/gh/freedomofpress/securedrop/8735?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

<pre>
E       NoSuchElementException: Message: Unable to locate element: {"method":"css selector","selector":"input[name=\"username\"]"}
</pre>

PR #3046, opened 2018-02-21
Release script: do not create tag for release candidates

### LINKED ISSUE #2950: Release candidates deb packages are considered higher version than the final release
# Bug

## Description

During the QA process - one might want to test a RC version of SD on a physical instance. If you do that though, it becomes annoying to install the final release candidate because debian/ubuntu considers: 

`0.5.1` <  `0.5.1-rc#`

## Steps to Reproduce

* Install a release candidate debian package, 
* attempt to upgrade it to the latest,
* cry in agony at the uncruel world. 

## Expected Behavior

I expect the prod release will be recognized by apt as a higher version than a release candidate.

## Actual Behavior

Apt requires a force install of the lower version and there is a possibility the old release candidate could auto-upgrade if one doesnt properly clean out their apt repo list.

## Comments

Luckily! ol' @msheiny has a fix for you -- utilize the `~rc#` versioning scheme instead of `-rc#` . For example `0.5.2~rc1`. Notice the squiggly line instead of the flat line for all you people that can't discern squiggles. Is that a thing? I feel like thats a medical condition.

redshiftzero:
## Status

Ready for review

## Description of Changes

Closes #2950

Previously, we used the version string 0.x.y-rcN for the Nth release
candidate for 0.x.y. Unfortunately, this rc version string was
considered a higher version than the actual release.

In 0.5.2 and beyond, we use the version string 0.x.y\~rcN for the Nth
release candidate for 0.x.y. This versioning scheme is correct, in
that Debian and Ubuntu recognize 0.x.y as a higher version than
0.x.y~rcN.

This commit updates our release script to create valid tags for
release candidates. Since the character "~" is not
a valid character in a git tag, without this change the script
will fail.

## Testing

1. A version string with a `~` produces a tag file with the name `0.x.y-rcZ`:

```
$ ./update_version.sh 0.5.3~rc1
...snip...
[ok] Version update complete and committed.
Since this is a release candidate, no tag has been generated.
```

2. A version string without `~` also produces a tag file:

```
$ ./update_version.sh 0.5.3
...snip...
[ok] Version update complete and committed.
     Please continue the airgapped signing process with the tag file: 0.5.3.tag
```

3. A version string with `-rc` in it exits: 

```
$ ./update_version.sh 0.5.7-rc1
Release candidates should use the versioning 0.x.y~rcZ!
```

## Deployment

Release process only

## Checklist

N/A

### DIFF
--- a/securedrop/Dockerfile
+++ b/securedrop/Dockerfile
@@ -6,7 +6,8 @@ ARG USER_ID
 ENV USER_ID ${USER_ID:-0}
 
 RUN apt-get update && \
-    apt-get install -y python-pip libpython2.7-dev libssl-dev secure-delete \
+    apt-get install -y devscripts \
+                       python-pip libpython2.7-dev libssl-dev secure-delete \
                        gnupg2 ruby redis-server firefox git xvfb haveged curl \
                        gettext paxctl x11vnc enchant
 
--- a/update_version.sh
+++ b/update_version.sh
@@ -10,10 +10,12 @@ if ! which dch > /dev/null ; then
   exit 1
 fi
 
-# Since we're running in a VM, we won't have access to ~/.gitconfig. So the
+cd "$(git rev-parse --show-toplevel)"
+
+# Since we may be running in a container, we may not  have access to ~/.gitconfig. So the
 # repo-level git user config file must be set.
-if ! grep -q '^\[user\]' /vagrant/.git/config; then
-    echo 'Please set your git user config in /vagrant/.git/config and retry!'
+if ! grep -q '^\[user\]' .git/config; then
+    echo 'Please set your git user config in .git/config and retry!'
     exit 1
 fi
 
--- a/update_version.sh
+++ b/update_version.sh
@@ -58,7 +58,7 @@ sed -i "s@$(echo "${OLD_VERSION}" | sed 's/\./\\./g')@$NEW_VERSION@g" docs/conf.
 
 # Update the changelog
 sed -i "s/\(## ${OLD_VERSION}\)/## ${NEW_VERSION}\n\n\n\n\1/g" changelog.md
-vim +5 changelog.md
+vi +5 changelog.md
 
 export DEBEMAIL="${DEBEMAIL:-securedrop@freedom.press}"
 export DEBFULLNAME="${DEBFULLNAME:-SecureDrop Team}"
--- a/update_version.sh
+++ b/update_version.sh
@@ -63,12 +63,22 @@ dch -b -v "${NEW_VERSION}" -D trusty -c install_files/securedrop-app-code/usr/sh
 # Due to `set -e`, providing an empty commit message here will cause the script to abort early.
 git commit -a
 
-# Initiate the process of creating a signed tag, using the workflow for signing with the airgapped signing key.
-git tag -a "${NEW_VERSION}"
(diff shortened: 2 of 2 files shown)


dachary:
I would like to have tags for RC, even if they do not reflect exactly the release number because tags do not allow for ~ in their name. It makes it so much easier for scripts.

What about this: from now on we tag releases with 0.X.Y.Z where X is the major release number, Y is the minor release number and Z is the release candidate with rc%02d for the release candidates and release for the final release, because **release** is lexicographically higher than **rc...** :

* 0.X.Y.rc%02d is the release 0.X.Y~rc%02d
* 0.X.Y.release is the release 0.X.Y

In other words we have natural lexicographically ordered set of tags which has a direct mapping to the bizarre exception introduced by ~ in Debian packages.

What do you think ?

redshiftzero:
In the spirit of transparency, I agree that we should ideally be pushing up tags for release candidates, so I've updated this PR to generate tag files for them as well.  I think we should keep the current versioning where stable releases as just `0.x.y` - as that is a common pragma (e.g. in [SemVer](https://semver.org/#spec-item-11)) and our admins are used to it (which matters (for now) since we have them check out git tags).

conorsch:
Looks great, @redshiftzero. Ran through the tag generation locally and the handling of the new `x.y.z~rc1` format works great. _However_, the script also proceeds blissfully even with `x.y.z-rc1`. That might be something we want to detect and bail out on.

There are a few quibbles I have with the existing `update_version.sh` script:

* doesn't _really_ need to be run in the dev VM, which is deprecated in favor of containers anyway
* much of the manual entry is rote: e.g. `See changelog.md`, filling out the version for both the commit message and the tag name; we can pass those via commands to reduce repetitive tasks and minimize mistakes.

Happy to open a separate issue if you agree with the above objections, @redshiftzero. I won't block merge on the permitted hyphens for this PR.

redshiftzero:
Commit c4c020d added for version strings with `-rc`, updated PR description

dachary:
@redshiftzero I pushed a few changes to enable running in the Docker container. Feel free to discard if that's not the right time to do that and I'll rework them later.

dachary:
<pre>
$ securedrop/bin/dev-shell ../update_version.sh 0.5.3-rc1
Docker image build in progress  done !
Release candidates should use the versioning 0.x.y~rcZ!
$ securedrop/bin/dev-shell ../update_version.sh 0.5.3
Docker image build in progress  done !
[fix-release-shell-script 2ed9724] kjdf
 12 files changed, 22 insertions(+), 12 deletions(-)
[ok] Version update complete and committed.
A tag has been generated: 0.5.3.tag
If you wish to release this version, please continue the airgapped signing process with the tag file.
$ securedrop/bin/dev-shell ../update_version.sh 0.5.3~rc1
Docker image build in progress  done !
dch warning: new version (0.5.3~rc1) is less than
the current version number (0.5.3).
[fix-release-shell-script 0af9e3c] kjdf
 12 files changed, 22 insertions(+), 12 deletions(-)
[ok] Version update complete and committed.
A tag has been generated: 0.5.3-rc1.tag
If you wish to release this version, please continue the airgapped signing process with the tag file.
</pre>

Works as advertized :100:

redshiftzero:
No worries, just ran through your additions - :+1: from me!

PR #3054, opened 2018-02-21
tests: makedirs is better than system("mkdir -p")

dachary:
## Status

Ready for review

## Description of Changes

makedirs is better than system("mkdir -p")

## Testing

* CI should pass

## Deployment

N/A

### DIFF
--- a/securedrop/tests/pages-layout/functional_test.py
+++ b/securedrop/tests/pages-layout/functional_test.py
@@ -43,7 +43,8 @@ class FunctionalTest(functional_test.FunctionalTest):
         self.log_dir = abspath(
             os.path.join(dirname(realpath(__file__)),
                          'screenshots', self.accept_languages))
-        os.system("mkdir -p " + self.log_dir)
+        if not os.path.exists(self.log_dir):
+            os.makedirs(self.log_dir)
         firefox = self._prepare_webdriver()
         profile = webdriver.FirefoxProfile()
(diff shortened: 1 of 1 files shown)


dachary:
@emkll that was a test to see if you were actually reading the code. And no, I'm not making that up: 7 months ago I already knew you'd be with us.

emkll:
Would there be a situation where the folder already exists? In which case I would suggest you check `if os.path.exists(self.log_dir).

`mkdir -p` returns 0 even if the specified folder exists, wheras normal `mkdir` returns 1 (and i suspect makedirs will throw an exception.

conorsch:
Yes, @emkll is correct, `os.makedirs` raises an OSError if the directory already exists. Existence checking does seem the sane way to sidestep—we don't want to `pass` on OSError, since an OSError could also mean that the directory could not be created, due to permissions or otherwise.

Also, the default mode is `0777`, which I find a bit permissive, but as the codepath is only relevant to tests, should be fine.

Docs: https://docs.python.org/2/library/os.html#os.makedirs

dachary:
> mkdir -p returns 0 even if the specified folder exists,

Ah! here is my excuse for using **mkdir -p** :-D

dachary:
> I would suggest you check if os.path.exists(self.log_dir).

Done. And now we have a beautifully pythonic code :tada:

PR #3055, opened 2018-02-21
WIP - Bandit static analysis

emkll:
## Status

Ready for Review
## Description of Changes

- Adds Bandit static code security scanner to CI for entire codebase
- CI will fail if the scanner detects a high or medium severity vulnerability in the codebase
- Excludes test-related folders
- Fixed (or annotated) all medium or high security findings to pass ci

## Testing

- CI must pass
- Fixes must not introduce regressions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,6 +116,12 @@ safety: ## Runs `safety check` to check python dependencies for vulnerabilities
 		|| exit 1; \
 	done
 
+# Bandit is a static code analysis tool to detect security vulnerabilities in Python applications
+# https://wiki.openstack.org/wiki/Security/Projects/Bandit
+.PHONY: bandit
+bandit: ## Run bandit excluding manage.py
+	@bandit -r securedrop/ -x securedrop/manage.py,securedrop/tests/ -ll
+
 .PHONY: update-pip-requirements
 update-pip-requirements: ## Updates all Python requirements files via pip-compile.
 	make -C admin update-pip-requirements
--- a/Makefile
+++ b/Makefile
@@ -119,8 +119,8 @@ safety: ## Runs `safety check` to check python dependencies for vulnerabilities
 # Bandit is a static code analysis tool to detect security vulnerabilities in Python applications
 # https://wiki.openstack.org/wiki/Security/Projects/Bandit
 .PHONY: bandit
-bandit: ## Run bandit excluding manage.py
-	@bandit -r securedrop/ -x securedrop/manage.py,securedrop/tests/ -ll
+bandit: ## Run bandit with medium level excluding test-related folders
+	@bandit --recursive . --exclude admin/.tox,admin/.venv,admin/.eggs,molecule,testinfra,securedrop/tests,.tox,.venv -ll
 
 .PHONY: update-pip-requirements
 update-pip-requirements: ## Updates all Python requirements files via pip-compile.
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -85,7 +85,7 @@ class SiteConfig(object):
                 cmd = 'torify ' + cmd
             cmd += ' ' + (dns and dns or '8.8.8.8')
             try:
-                result = subprocess.check_output(cmd, shell=True,
+                result = subprocess.check_output(cmd.split(' '),
                                                  stderr=subprocess.STDOUT)
             except subprocess.CalledProcessError as e:
                 result = e.output
--- a/securedrop/journalist.py
+++ b/securedrop/journalist.py
@@ -9,4 +9,4 @@ app = create_app(config)
 
 if __name__ == "__main__":  # pragma: no cover
     debug = getattr(config, 'env', 'prod') != 'prod'
-    app.run(debug=debug, host='0.0.0.0', port=8081)
+    app.run(debug=debug, host='0.0.0.0', port=8081)  # nosec
--- a/securedrop/manage.py
+++ b/securedrop/manage.py
@@ -313,7 +313,7 @@ def translate_messages(args):
                    sources=" ".join(args.source)))
 
         changed = subprocess.call("git diff --quiet {}".format(messages_file),
-                                  shell=True)
+                                  shell=True)  # nosec
 
         if changed and len(os.listdir(args.translations_dir)) > 1:
             sh("""
@@ -360,7 +360,7 @@ def translate_desktop(args):
                    sources=" ".join(args.source)))
 
         changed = subprocess.call("git diff --quiet {}".format(messages_file),
-                                  shell=True)
+                                  shell=True)  # nosec
 
         if changed:
             for f in os.listdir(args.translations_dir):
--- a/securedrop/request_that_secures_file_uploads.py
+++ b/securedrop/request_that_secures_file_uploads.py
@@ -24,7 +24,7 @@ class RequestThatSecuresFileUploads(wrappers.Request):
             # note in `config.py` for more info. Instead, we just use
             # `/tmp`, which has the additional benefit of being
             # automatically cleared on reboot.
-            return SecureTemporaryFile('/tmp')
+            return SecureTemporaryFile('/tmp')  # nosec
         return BytesIO()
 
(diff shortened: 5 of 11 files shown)


Makefile
```diff
@@ -116,6 +116,12 @@ safety: ## Runs `safety check` to check python dependencies for vulnerabilities
 		|| exit 1; \
 	done
 
+# Bandit is a static code analysis tool to detect security vulnerabilities in Python applications
+# https://wiki.openstack.org/wiki/Security/Projects/Bandit
+.PHONY: bandit
+bandit: ## Run bandit with medium level excluding test-related folders
+	@bandit --recursive . --exclude molecule,testinfra,securedrop/tests,admin/.tox -ll
```
conorsch: This is passing CI, _but_ additional dirs should be excluded so that devs can run the target locally:

* `admin/.venv/`
* `admin/.eggs/`
* `.venv/`

Those directories will be created during testing in Tails, e.g. using prod VMs.

dachary: and `.tox`

securedrop/request_that_secures_file_uploads.py
```diff
@@ -24,7 +24,7 @@ def _secure_file_stream(self, total_content_length, content_type,
             # note in `config.py` for more info. Instead, we just use
             # `/tmp`, which has the additional benefit of being
             # automatically cleared on reboot.
-            return SecureTemporaryFile('/tmp')
+            return SecureTemporaryFile('/tmp')  # nosec
```
heartsucker: Why does this need to be marked `# nosec`?

securedrop/manage.py
```diff
@@ -313,7 +313,7 @@ def translate_messages(args):
                    sources=" ".join(args.source)))
 
         changed = subprocess.call("git diff --quiet {}".format(messages_file),
```
heartsucker: Since we're calling this with `--quiet`, can we not disable shell here too?

securedrop/manage.py
```diff
@@ -360,7 +360,7 @@ def translate_desktop(args):
...
```
heartsucker: Since we're calling this with `--quiet`, can we not disable shell here too?

securedrop/request_that_secures_file_uploads.py
emkll: There is a rule in bandit that is a string match for /tmp `/tmp`:
```
>> Issue: [B108:hardcoded_tmp_directory] Probable insecure usage of temp file/directory.
   Severity: Medium   Confidence: Medium
   Location: ./securedrop/request_that_secures_file_uploads.py:27
```
Which warns against the usage of /tmp to store application data (https://www.owasp.org/index.php/Insecure_Temporary_File). Given the way `filesystem_id` is generated and files are encrypted as they are streamed to disk, it shouldn't affect the confidentiality of the submission. As for integrity, perhaps using an authenticated encryption mode (GCM) *might* be an improvement, but would obviously require very careful consideration.

heartsucker: Could you open a ticket for this? Not necessarily for an AEAD, but for maybe moving it to `/var/run/securedrop/tmp`.

securedrop/journalist.py
```diff
@@ -9,4 +9,4 @@
 
 if __name__ == "__main__":  # pragma: no cover
     debug = getattr(config, 'env', 'prod') != 'prod'
-    app.run(debug=debug, host='0.0.0.0', port=8081)
+    app.run(debug=debug, host='127.0.0.1', port=8081)
```
heartsucker: This will break dev (I think) since the container is part of the `10.0.0.0/8` network. Prod also doesn't start SD with main. It uses WSGI to start the app, so the debug check here is actually unnecessary. Also fun fact, adding that line was my first ever commit to SD.

securedrop/source.py
```diff
@@ -9,4 +9,4 @@
 
 if __name__ == "__main__":  # pragma: no cover
     debug = getattr(config, 'env', 'prod') != 'prod'
-    app.run(debug=debug, host='0.0.0.0', port=8080)
+    app.run(debug=debug, host='127.0.0.1', port=8080)
```
heartsucker: This will break dev (I think). See other comment.

securedrop/manage.py
emkll: Good call, unfortunately the change does not appear trivial (to me). Changing that line to:
```
changed = subprocess.call(['git', 'diff', '--quiet', messages_file])
```
makes `tests/test_i18n.py::TestI18N::test_i18n` fail with `OSError: [Errno 2] No such file or directory`. Any ideas on what's going on?

dachary: not really
<pre>
>>> import subprocess
>>> subprocess.call(['git', 'rev-parse', 'HEAD'])
f0c32641464892897f172bd601e3522cbba518f3
</pre>
so `message_file` must be not found but ... how ? I suspect this is unrelated. Does it work if you revert back to using shell=True ?

heartsucker: It could be an absolute path thing. Maybe `path.abspath(messages_file)`?

securedrop/journalist.py
emkll: Just tried this out, and you are correct. Given Flask's inability to bind to multiple interfaces, and WSGI is used to start the app, I will revert the change and add the nosec annotation. Let me know if you disagree

securedrop/manage.py
emkll: @dachary it does indeed pass with:
```
changed = subprocess.call(['git', 'diff', '--quiet', messages_file], shell=True)
```

securedrop/journalist.py
heartsucker: Nope, that's exactly what I would do.

securedrop/manage.py
dachary: this is really wierd, what kind of side effect can shell=True have to make a difference ?

securedrop/request_that_secures_file_uploads.py
emkll: https://github.com/freedomofpress/securedrop/issues/3067

securedrop/manage.py
emkll: Using `os.path.abspath(messages_file)` still throws the error. Given the `sh` function in the file, I suggest we keep the nosec annotation.

heartsucker: Totally fair. Can you open a ticket to investigate this? Could be something a newbie could look in to too since it's is only a couple line change.

emkll: https://github.com/freedomofpress/securedrop/issues/3071

admin/securedrop_admin/__init__.py
```diff
@@ -85,7 +85,7 @@ def lookup_fqdn(self, fqdn, dns=None):
                 cmd = 'torify ' + cmd
             cmd += ' ' + (dns and dns or '8.8.8.8')
             try:
-                result = subprocess.check_output(cmd, shell=True,
+                result = subprocess.check_output(cmd,
```
redshiftzero: Since `cmd` is a string, don't we need to convert to a list using e.g. `cmd.split(' ')` or similar?

emkll: Yes, we do. Good catch!

admin/securedrop_admin/__init__.py
```diff
@@ -85,7 +85,7 @@ def lookup_fqdn(self, fqdn, dns=None):
                 cmd = 'torify ' + cmd
             cmd += ' ' + (dns and dns or '8.8.8.8')
             try:
-                result = subprocess.check_output(cmd, shell=True,
+                result = subprocess.check_output(cmd.split(' '),
```
heartsucker: Not going to block on this because what you've done is a strict improvement, but maybe we should programmatically construct a list and then validate input instead of splitting on spaces like this. The former feels more robust.

redshiftzero:
One of the admin tests - `TestSiteConfig.test_lookup_dns` - needs updating - check out the [documentation here](https://docs.securedrop.org/en/latest/development/admin_development.html) on doing that.

PR #3057, opened 2018-02-21
Improvements in image upload functionality

### LINKED ISSUE #2801: Recommend dimensions for logo(s) uploads
# Recommend dimensions for logo(s) uploads

## Description

Currently logos vary in size dramatically between different [deployed instances](https://securedrop.org/directory). This makes it much more difficult to have a clean design for the top bar layout on mobile (in particular) and also restricts the design options on desktop.

This issue is an extension to #2769 to encourage journalists to upload a logo that will display well on desktop and mobile screens within the constraints of the design and allow for consistent responsive layout as per #2551. Journalists should be warned that logos which have different dimensions to these cannot be guaranteed to display optimally and shown an example guide of preparing the logo including desired clearance (whitespace).

For example:

* Upload desktop square logo (recommended size 150x150) (png)
* Upload square mobile logo (recommended size 50x50) (png) OR Upload landscape mobile logo (recommended size 100x50) (png)

The above sizes are a best guess at recommended dimensions based on the current design and assume clearance is included in the image and it is centered correctly within that space. My assumption is that it's best to avoid retina for now to prioritise download speed on the Tor network. Tor browser doesn't allow SVG in high security mode so PNG at 1x for now.

Would be useful to add a mockup to this issue to show layout of instructions to journalists and upload controls.

### LINKED ISSUE #2807: Admin logo image upload: Resize images server-side if they are too large
## Description

We don't optimize uploaded images server-side so this makes for a slow experience for sources and journalists if a news organization elects to use a too large logo image. The large image may be transferred to the client which will then be resized to a smaller size anyway. It would make a lot more sense to do resizing server-side. What do we think (cc: @toast) about the following flow:

1. Admin uploads image(s), which may or may not be the correct size.
2. We resize the image(s) server-side and display them to the user. (we could use `PyImageOptimizer` as suggested in #2801 but since it looks like that dependency is not widely used / has not seen recent activity, and the functionality is simple I would advocate we implement this functionality in the application code ourselves). 
3. The uploaded image is shown, and admins can resubmit if they don't like how it has been resized.  

We should avoid adding additional JavaScript such that the journalist interface degrades gracefully without js, so this means that we can't have the kind of image cropping interface we are used to from other websites. 

Related issue: #2801 

## User Stories

As a SecureDrop source or journalist, I don't want to wait for a large logo image to unnecessarily download so that my Tor Onion Service browsing experience is faster.

### LINKED ISSUE #3043: Image uploaded by journalist is always saved as a PNG file
# Bug

## Description

Regardless of the type of image the journalist uploads (JPG, JPEG or PNG), the image `securedrop/static/i/logo.png`  is always saved with a .png extension. This leads to image viewing applications not being able to read and display the file correctly.

## Steps to Reproduce

As an Admin, navigate to 'Instance Configuration' in the journalist app, and upload a JPG/JPEG image as logo.

Open 'logo.png' in the directory `securedrop/static/i/`.

## Expected Behavior

The image should be displayed without any errors.

## Actual Behavior

screenshot from 2018-02-21 05-32-52

## Comments

Tested using Gnome `eog` on Debian Stretch.

### LINKED ISSUE #3058: Check for a valid image file when uploading logo
# Bug

## Description

In the journalist application, when the Admin uploads an image file for logo, no check is performed to determine if it is a valid image file, or a corrupted/unreadable one. The application currently only checks for a valid extension.

Thus if someone uploads a corrupt/unreadable or non-image file with an extension like JPG, JPEG or PNG, it gets updated and shows up like:

screenshot from 2018-02-22 19-43-45

## Steps to Reproduce

Upload a non-image file with `.png` extension.

## Expected Behavior

The image should be ignored and user should be prompted to upload again.

## Actual Behavior

The incorrect image file gets uploaded.

## Comments

As suggested by @heartsucker, I'll try to write a fix for this and integrate into #3057

aydwi:
## Status

Ready for review

## Description of Changes

This PR-

1. Fixes #2807 by resizing the uploaded logo to a maximum width of 500px, or a maximum height of 450px (whichever is higher) while maintaining the original aspect ratio. Maximum limit is chosen in accordance with the recommended size by SecureDrop docs. Images with dimensions below this constraint are not resized.

2. Fixes #3043 by converting JPG/JPEG image files to PNG before saving them. Thus no matter which format the Admin chooses to upload, the final image is always a PNG file.

3. Possibly fixes #2801 by adding a message at the logo upload screen informing the user about the recommended image size as per the docs.

fofvv3l

4. Fixes #3058 by only accepting valid image files.

screenshot from 2018-02-22 22-27-11

## Dependencies Introduced

1. Python `Pillow`, installed via `pip`

2. `libjpeg-dev`, installed via `apt`

## Testing

1. After successfully uploading, image size can be verified by checking the dimensions of `securedrop/securedrop/static/i/custom_logo.png`.

2. After successfully uploading a JPG/JPEG image, it can be verified that `securedrop/securedrop/static/i/` always contains `custom_logo.png`

3. Journalist app should display the recommended size on the upload page.

4. Try to upload a non-image file with an extension like JPG/JPEG/PNG. A warning message should appear that image cannot be processed.

### DIFF
--- a/securedrop/journalist_app/admin.py
+++ b/securedrop/journalist_app/admin.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+from PIL import Image
+
 import os
 
 from flask import (Blueprint, render_template, request, url_for, redirect, g,
@@ -31,11 +33,19 @@ def make_blueprint(config):
         form = LogoForm()
         if form.validate_on_submit():
             f = form.logo.data
-            static_filepath = os.path.join(config.SECUREDROP_ROOT,
-                                           "static/i/logo.png")
-            f.save(static_filepath)
-            flash(gettext("Image updated."), "logo-success")
-            return redirect(url_for("admin.manage_config"))
+            custom_logo_filepath = os.path.join(config.SECUREDROP_ROOT,
+                                                "static/i/custom_logo.png")
+            try:
+                with Image.open(f) as im:
+                    im.thumbnail((500, 450), resample=3)
+                    im.save(custom_logo_filepath, "PNG")
+                flash(gettext("Image updated."), "logo-success")
+            except Exception:
+                flash("Unable to process the image file."
+                      " Try another one.", "logo-error")
+            finally:
+                return redirect(url_for("admin.manage_config"))
+
         else:
             for field, errors in form.errors.items():
                 for error in errors:
--- a/securedrop/journalist_templates/config.html
+++ b/securedrop/journalist_templates/config.html
@@ -23,7 +23,7 @@
 <p>{{ gettext('Here you can update the image displayed on the SecureDrop web interfaces:') }}</p>
 
 <p>
-  <img src="{{ url_for('static', filename='i/logo.png')|autoversion }}" class="logo small" alt="SecureDrop" width="250px">
+  <img src="{{ url_for('main.select_logo') }}" class="logo small" alt="SecureDrop" width="250px">
 </p>
 
 <form method="post" enctype="multipart/form-data">
@@ -32,6 +32,9 @@
     {{ form.logo(id="logo-upload") }}
     <br>
   </p>
+  <h5>
+    {{ gettext('Recommended size: 500px * 450px') }}
+  </h5>
   <button type="submit" class="sd-button" id="submit-logo-update">
     <i class="fas fa-pencil-alt"></i>{{ gettext('UPDATE LOGO') }}
   </button>
--- a/securedrop/tests/test_journalist.py
+++ b/securedrop/tests/test_journalist.py
@@ -5,8 +5,10 @@ import io
 import random
 import unittest
 import zipfile
+import base64
 
 from cStringIO import StringIO
+from io import BytesIO
 from flask import url_for, escape, session, current_app, g
 from flask_testing import TestCase
 from mock import patch
@@ -954,9 +956,11 @@ class TestJournalistApp(TestCase):
 
         try:
             self._login_admin()
-
+            # Create 1px * 1px 'white' PNG file from its base64 string
             form = journalist_app_module.forms.LogoForm(
-                logo=(StringIO('imagedata'), 'test.png')
+                logo=(BytesIO(base64.decodestring
+                      ("iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQ"
+                       "VR42mP8/x8AAwMCAO+ip1sAAAAASUVORK5CYII=")), 'test.png')
             )
             self.client.post(url_for('admin.manage_config'),
                              data=form.data,
@@ -977,8 +981,10 @@ class TestJournalistApp(TestCase):
         resp = self.client.post(url_for('admin.manage_config'),
                                 data=form.data,
                                 follow_redirects=True)
-        self.assertMessageFlashed("Upload images only.", "logo-error")
-        self.assertIn('Upload images only.', resp.data)
+        self.assertMessageFlashed("You can only upload JPG/JPEG"
+                                  " or PNG image files.", "logo-error")
+        self.assertIn("You can only upload JPG/JPEG"
+                      " or PNG image files.", resp.data)
 
     def test_logo_upload_with_empty_input_field_fails(self):
         self._login_admin()
--- a/install_files/ansible-base/group_vars/all/securedrop
+++ b/install_files/ansible-base/group_vars/all/securedrop
@@ -51,5 +51,6 @@ appserver_dependencies:
   - apparmor-utils
   - redis-server
   - supervisor
+  - libjpeg-dev
 
 tor_apt_repo_url: https://tor-apt.freedom.press
--- a/install_files/securedrop-app-code/DEBIAN/control
+++ b/install_files/securedrop-app-code/DEBIAN/control
@@ -6,5 +6,5 @@ Homepage: https://securedrop.org
 Package: securedrop-app-code
 Version: 0.6~rc2
 Architecture: amd64
-Depends: python-pip,apparmor-utils,gnupg2,haveged,python,python-pip,secure-delete,sqlite,apache2-mpm-worker,libapache2-mod-wsgi,libapache2-mod-xsendfile,redis-server,supervisor,securedrop-keyring,securedrop-config
+Depends: python-pip,apparmor-utils,gnupg2,haveged,python,python-pip,secure-delete,sqlite,apache2-mpm-worker,libapache2-mod-wsgi,libapache2-mod-xsendfile,redis-server,supervisor,securedrop-keyring,securedrop-config,libjpeg-dev
 Description: Packages the SecureDrop application code pip dependencies and apparmor profiles. This package will put the apparmor profiles in enforce mode. This package does use pip to install the pip wheelhouse
--- a/molecule/builder/Dockerfile
+++ b/molecule/builder/Dockerfile
@@ -8,6 +8,6 @@ LABEL image_name="$NAME"
 
 RUN apt-get -y update && apt-get install -y python rsync sudo bash \
             devscripts git aptitude ipython libssl-dev ntp \
-            python-dev python-pip python-ipdb ruby tmux vim paxctl && \
+            python-dev python-pip python-ipdb ruby tmux vim paxctl libjpeg-dev && \
             aptitude -y dist-upgrade && \
             apt-get clean
--- a/securedrop/Dockerfile
+++ b/securedrop/Dockerfile
@@ -9,7 +9,7 @@ RUN apt-get update && \
     apt-get install -y devscripts \
                        python-pip libpython2.7-dev libssl-dev secure-delete \
                        gnupg2 ruby redis-server firefox git xvfb haveged curl \
-                       gettext paxctl x11vnc enchant
+                       gettext paxctl x11vnc enchant libjpeg-dev
 
 RUN gem install sass -v 3.4.23
 
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -27,7 +27,7 @@ if typing.TYPE_CHECKING:
     # http://flake8.pycqa.org/en/latest/user/error-codes.html?highlight=f401
     from sdconfig import SDConfig  # noqa: F401
 
-_insecure_views = ['main.login', 'static']
+_insecure_views = ['main.login', 'main.select_logo', 'static']
 
 
 def create_app(config):
--- a/securedrop/journalist_app/forms.py
+++ b/securedrop/journalist_app/forms.py
@@ -59,5 +59,6 @@ class LogoForm(FlaskForm):
     logo = FileField(validators=[
         FileRequired(message=gettext('File required.')),
         FileAllowed(['jpg', 'png', 'jpeg'],
-                    message=gettext('Upload images only.'))
+                    message=gettext("You can only upload JPG/JPEG"
+                                    " or PNG image files."))
     ])
--- a/securedrop/journalist_app/main.py
+++ b/securedrop/journalist_app/main.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+import os
+
 from datetime import datetime
 from flask import (Blueprint, request, current_app, session, url_for, redirect,
                    render_template, g, flash, abort)
@@ -42,6 +44,14 @@ def make_blueprint(config):
         session.pop('uid', None)
         session.pop('expires', None)
         return redirect(url_for('main.index'))
+
+    @view.route('/org-logo')
+    def select_logo():
+        if os.path.exists(os.path.join(current_app.static_folder, 'i',
+                          'custom_logo.png')):
+            return redirect(url_for('static', filename='i/custom_logo.png'))
+        else:
+            return redirect(url_for('static', filename='i/logo.png'))
 
     @view.route('/')
     def index():
--- a/securedrop/journalist_templates/base.html
+++ b/securedrop/journalist_templates/base.html
@@ -34,7 +34,7 @@
       <div class="container">
         {% block header %}
         <div id="header">
-          <a href="{{ url_for('main.index') }}"><img src="/static/i/{{ header_image }}" class="logo small" alt="SecureDrop" width="250px"></a>
+          <a href="{{ url_for('main.index') }}"><img src="{{ url_for('main.select_logo') }}" class="logo small" alt="SecureDrop" width="250px"></a>
           {% include 'locales.html' %}
           {% if use_custom_header_image %}
           <div class="powered">
--- a/securedrop/requirements/securedrop-app-code-requirements.in
+++ b/securedrop/requirements/securedrop-app-code-requirements.in
@@ -17,3 +17,4 @@ scrypt
 SQLAlchemy
 typing
 Werkzeug==0.12.2
+Pillow
--- a/securedrop/requirements/securedrop-app-code-requirements.txt
+++ b/securedrop/requirements/securedrop-app-code-requirements.txt
@@ -17,6 +17,7 @@ itsdangerous==0.24        # via flask
 jinja2==2.10
 jsmin==2.2.2
 markupsafe==1.0           # via jinja2
+pillow==5.0.0
 psutil==5.4.3
 pycryptodomex==3.4.7
 pyotp==2.2.6
--- a/securedrop/source_app/main.py
+++ b/securedrop/source_app/main.py
@@ -40,6 +40,14 @@ def make_blueprint(config):
         session['new_user'] = True
         return render_template('generate.html', codename=codename)
 
+    @view.route('/org-logo')
+    def select_logo():
+        if os.path.exists(os.path.join(current_app.static_folder, 'i',
+                          'custom_logo.png')):
+            return redirect(url_for('static', filename='i/custom_logo.png'))
+        else:
+            return redirect(url_for('static', filename='i/logo.png'))
+
     @view.route('/create', methods=['POST'])
     def create():
         filesystem_id = current_app.crypto_util.hash_codename(
--- a/securedrop/source_templates/base.html
+++ b/securedrop/source_templates/base.html
@@ -17,7 +17,7 @@
         {% block header %}
         <div id="header">
           <a href="{% if 'logged_in' in session %}{{ url_for('main.lookup') }}{% else %}{{ url_for('main.index') }}{% endif %}">
-            <img src="/static/i/{{ header_image }}" class="logo small" alt="SecureDrop" width="250px">
+            <img src="{{ url_for('main.select_logo') }}" class="logo small" alt="SecureDrop" width="250px">
           </a>
           {% include 'locales.html' %}
           {% if use_custom_header_image %}
--- a/securedrop/source_templates/index.html
+++ b/securedrop/source_templates/index.html
@@ -26,7 +26,7 @@
            See _source_index.sass for a more full understanding. #}
         <div class="index-wrap">
           <div class="header">
-            <img src="/static/i/{{ header_image }}" alt="SecureDrop">
+            <img src="{{ url_for('main.select_logo') }}" alt="SecureDrop">
             {% if use_custom_header_image %}
             <div class="powered">
               {{ gettext('Powered by') }}<br>


redshiftzero:
This looks good on first inspection, note you have a few linting errors (once this is fixed then the full staging tests will run):

```
./securedrop/tests/test_journalist.py:753:80: E501 line too long (159 > 79 characters)
./securedrop/journalist_app/admin.py:41:50: E251 unexpected spaces around keyword / parameter equals
./securedrop/journalist_app/admin.py:41:52: E251 unexpected spaces around keyword / parameter equals
```

Note you can add `# noqa` after a line if you think there is a compelling reason it should not be linted

aydwi:
I'll correct the linting issues. May I ask what does **staging-test-with-rebase** do?

conorsch:
> May I ask what does staging-test-with-rebase do?

The `staging-test-with-rebase` phase will:

1. Rebase the PR branch on top of the latest "develop" branch (thus the name!)
2. Provision "staging" Application and Monitor Servers
3. Run configuration tests against to validate server state
4. Run application tests to validate webapp functionality

We also had a `staging-test` phase, but rather than run both, we only run one. (See #2932 for recent changes.)

securedrop/journalist_templates/config.html
```diff
@@ -32,6 +32,9 @@ <h2>{{ gettext('Logo Image') }}</h2>
     {{ form.logo(id="logo-upload") }}
     <br>
   </p>
+  <h5>
+    {{gettext('Recommended size: 500px * 450px')}}
```
heartsucker: Two things here:

1. We use `{{ gettext(...) }}` (with spaces) stylistically to improve readability.
2. Is there a reason to suggest images that are 500x450 pixels but we thumbnail to 500x500 pixels max?

securedrop/tests/test_journalist.py
```diff
@@ -744,9 +748,11 @@ def test_logo_upload_with_valid_image_succeeds(self):
 
         try:
             self._login_admin()
-
+            # Create 1px * 1px 'white' PNG file from its base64 string
             form = journalist_app.forms.LogoForm(
-                logo=(StringIO('imagedata'), 'test.png')
+                logo=(BytesIO(base64.decodestring
+                ("iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQ"
```
heartsucker: For the other reviewers, this was something I gave @aydwi out of band, and that is definitely a 1x1 white pixel.

securedrop/journalist_app/admin.py
```diff
@@ -34,6 +36,11 @@ def manage_config():
             static_filepath = os.path.join(config.SECUREDROP_ROOT,
```
heartsucker: We should save this to a temp file, then once we've successfully resized it move it to the old location. Right now if we throw an error thumbnailing it, we'll have a broken image.

It should go:

1. Save to tmp
2. Thumbnail that overwrites tmp
3. Move tmp to `static/i/logo.png` using `os.rename`

securedrop/journalist_app/admin.py
```diff
@@ -34,6 +36,11 @@ def manage_config():
...
                                            "static/i/logo.png")
             f.save(static_filepath)
+
+            with Image.open(static_filepath) as im:
```
heartsucker: If someone gives us a PNG or JPG that Pillow doesn't understand, the use will get the generic 500 error page. We should try/catch this to and `flash` something like "We couldn't process that image. Try another one."

securedrop/journalist_templates/config.html
aydwi: 1. Thanks for letting me know. I'll adjust that.
2. `Pillow` suggests giving size as a 2-tuple. The `thumbnail` method automatically adjusts the lower dimension according to aspect ratio. 500 is the upper limit.

If a user uploads square image with edge length greater than 500px, the end result will be 500px by 500px image, since aspect ratio is preserved. It will be resized to 500px by 450px only when the initial aspect ratio was 10:9.

I think the size 500px by 450px comes from the size of default SecureDrop logo. Maybe we can recommend a square image like 500px by 500px.

securedrop/tests/test_journalist.py
aydwi: Yes Thanks @heartsucker. I added a comment to clarify that to anyone reading the code as well. Also, I think including an image whose size is greater than 500px can also help test the resize function, although the base64 string will be very large.

securedrop/journalist_app/admin.py
aydwi: I did not write any exception handling here because I think Flask handles the type of images, and only JPG/JPEG/PNG make through. But that is a good point, `Pillow` may encounter something unexpected. I'll update the code.

securedrop/journalist_templates/config.html
heartsucker: I'm fine with either, but what we suggest and what we thumbnail to should be the same number just to avoid unexpected behavior.

securedrop/journalist_app/admin.py
heartsucker: I'm rather sure that the `FileAllowed` validator only checks the extension, and it's possible someone has a valid PNG that Pillow doesn't understand. We should try to catch these cases.

heartsucker:
It looks like we're having `mypy` errors, but this is all we get for the output.

```
make: Target 'lint' not remade because of errors.
if [[ ! -d .python3/.venv ]] ; then \
  virtualenv --python=python3 .python3/.venv && \
  .python3/.venv/bin/pip3 install mypy ; \
        fi
.python3/.venv/bin/mypy ./securedrop ./admin
Makefile:28: recipe for target 'ci-lint' failed
make: *** [ci-lint] Error 2
Exited with code 2
```

@kushaldas Some help here?

securedrop/journalist_templates/config.html
aydwi: I changed the maximum possible height of thumbnail to 450px. If defualt logo can be created in 500*500, we can start recommending 500*500

aydwi:
@heartsucker If I change the message to 'You can only upload files ending in .png, .jpg, or .jpeg', should I also make changes to `securedrop/translations` directory?

dachary:
The lint error is not the last line of the output because we intentionally run all lint checks even if one of them fail. The error at https://circleci.com/gh/freedomofpress/securedrop/8936?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link is:

<pre>
flake8 --exclude='config.py,.venv/'
./securedrop/tests/test_journalist.py:754:17: E128 continuation line under-indented for visual indent
./securedrop/tests/test_journalist.py:755:17: E128 continuation line under-indented for visual indent
</pre>

heartsucker:
@aydwi No, you don't need to do that. That will be taken care of by a different process.

securedrop/journalist_app/admin.py
aydwi: Made this change

securedrop/journalist_app/admin.py
```diff
@@ -33,9 +35,30 @@ def manage_config():
...
             f.save(static_filepath)
-            flash(gettext("Image updated."), "logo-success")
-            return redirect(url_for("admin.manage_config"))
+
+            try:
+                with Image.open(static_filepath) as im:
+                    imcopy = im.copy()
+                    imcopy.thumbnail((500, 450), resample=3)
+                    imcopy.save(temp_static_filepath, "PNG")
+                if os.path.exists(static_filepath):
```
heartsucker: I think this guard is unnecessary since we're already catching the `IOError`. Also if that path doesn't exist either because `static/i/` is missing then we have big problems, or if `static/i/logo.png` is missing, then we should just copy it anyway.

securedrop/journalist_app/admin.py
```diff
@@ -33,9 +35,30 @@ def manage_config():
...
+                    os.rename(temp_static_filepath, static_filepath)
+                flash(gettext("Image updated."), "logo-success")
+
+            except IOError:
```
heartsucker: I would change this to `except Exception` because I'm fairly sure that `Image` raises Pillow specific errors on all sorts of things, and we want to alert the user in all cases.

securedrop/journalist_app/admin.py
```diff
@@ -33,9 +35,30 @@ def manage_config():
...
+                with open(static_filepath, 'w') as img_file:
```
heartsucker: This `with` block is unnecessary becuse `os.rename` is effectively atomic, so I think it's acceptable to assume that we never end up with a broken `logo.png`.

securedrop/journalist_app/admin.py
```diff
@@ -33,9 +35,30 @@ def manage_config():
             f = form.logo.data
             static_filepath = os.path.join(config.SECUREDROP_ROOT,
                                            "static/i/logo.png")
+            temp_static_filepath = os.path.join(config.SECUREDROP_ROOT,
+                                                "static/i/temp_logo.png")
+            with open(static_filepath) as img_file:
```
heartsucker: As noted in other comments, I don't think we need to read this in. `os.rename` should just do the right thing on its own.

securedrop/journalist_app/admin.py
```diff
@@ -33,9 +35,30 @@ def manage_config():
             f = form.logo.data
             static_filepath = os.path.join(config.SECUREDROP_ROOT,
                                            "static/i/logo.png")
+            temp_static_filepath = os.path.join(config.SECUREDROP_ROOT,
+                                                "static/i/temp_logo.png")
```
heartsucker: I'm not sure we should save temp files in `./static`. I think we can use the `SECUREDROP_DATA_ROOT` and make sure there's a `tmp` in there. This would provide us with safe place for temp files.

aydwi:
@heartsucker I'll try to incorporate these changes. Can I know why "staging-test-with-rebase" is failing?

securedrop/journalist_app/admin.py
aydwi: >  or if static/i/logo.png is missing, then we should just copy it anyway.

I think I read once that the path to which we are moving should exist before `os.rename()` is called.

msheiny:
@aydwi  its failing during app installation 

`No distributions at all found for pillow==5.0.0 (from -r /var/www/securedrop/requirements/securedrop-app-code-requirements.txt (line 20)`

i don't think uppercase should matter but maybe pip interprets the requirements file differently than the cli tool? i dunno thats my first guess is to try setting it `Pillow` instead of `pillow` ?

aydwi:
@msheiny I built the requirements file with `pip-compile --output-file securedrop/requirements/securedrop-app-code-requirements.txt securedrop/requirements/securedrop-app-code-requirements.in`.

I added `Pillow` to the `.in` file, but in `.txt` file, it came as `pillow`. Is it a good approach to manually edit it?

msheiny:
huh @aydwi .... that cant be it then.. if pip-tools is handling it, should be fine.... hrmmm

securedrop/journalist_app/admin.py
aydwi: I'll update it

heartsucker:
@msheiny @aydwi 

```
heartsucker@pythagoras:/tmp$ virtualenv -p `which python2` venv
Running virtualenv with interpreter /usr/bin/python2.7
Using base prefix '/usr'
New python executable in /tmp/venv/bin/python2.7
Also creating executable in /tmp/venv/bin/python
Installing setuptools, pip, wheel...source venv/bin/actdone.
heartsucker@pythagoras:/tmp$ source venv/bin/activate
(venv) heartsucker@pythagoras:/tmp$ pip install pillow==5.0.0
Collecting pillow==5.0.0
  Using cached Pillow-5.0.0-cp35-cp35m-manylinux1_x86_64.whl
Installing collected packages: pillow
Successfully installed pillow-5.0.0
```

securedrop/journalist_app/admin.py
aydwi: Since `f.save(static_filepath)` always writes logo.png, this `with` block replaces it with current image id it is broken. Without this restoration, only a message will be flashed, but the broken image will remain saved as logo.png

heartsucker: Yes, which is why I'm suggesting that we have an endpoint that selects and uses something called `custom_logo.png` so we don't end up with a broken `logo.png`. If the app terminates while the write in that `with` block happens, we'll have a broken logo. You're more IO than necessary, and that could be a problem. Also, the advantage of having and endpoint that selects `custom_logo.png` is that if someone runs the Ansible playbooks, they won't overwrite the uploaded image.

dachary:
https://circleci.com/gh/freedomofpress/securedrop/8969?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link shows in the **Run Debian builds** part which is green / success:

<pre>
    TASK [build-securedrop-app-code-deb-pkg : Create pip wheel archive for Debian package requirements.] 
    changed: [build]
    
    TASK [build-securedrop-app-code-deb-pkg : Install dependencies for building translations] ***
    changed: [build] => (item=[u'gnupg2', u'haveged', u'python', u'python-pip', u'secure-delete', u'sqlite', u'apparmor-utils', u'redis-server', u'supervisor', u'libjpeg-dev'])
</pre>

But ** Create pip wheel archive for Debian package** does

<pre>
- name: Create pip wheel archive for Debian package requirements.
  command: >
    pip wheel
    -r {{ securedrop_code_filtered }}/requirements/securedrop-app-code-requirements.txt
    -w {{ securedrop_app_code_deb_dir }}/var/securedrop/wheelhouse
  tags: pip
</pre>
which should fail to populate the wheelhouse with pillow ... because the dependency it needs (libjpeg-dev) is installed **after**. So it looks like pip wheel does not fail when it should and we have  wheelhouse that does not contain pillow, reason why it fails to install from the package (because the package does not download from the net, it installs from locally available wheelhous exclusively).

dachary:
I tried manually and failed to reproduce with
<pre>
$ cd securedrop
$ ./bin/dev-shell bash
$ sudo bash
# apt-get remove --purge libjpeg-dev
# pip uninstall pillow
# pip wheel -r requirements/securedrop-app-code-requirements.txt -w /tmp/w
# ls /tmp/w/Pillow*
Pillow-5.0.0-cp27-none-linux_x86_64.whl
</pre>
Trying `make build-debs` now to verify that Pillow is misisng from the wheelhouse. If it is not missing I'll rebase & repush to confirm.

securedrop/journalist_app/admin.py
aydwi: Before moving on to logo endpoint approach, I want to discuss my current approach and what purpose the two `with` blocks are serving.

Case 1-

User tries to upload logo for the first time
The image is a valid png file - A.png
The default SecureDrop logo is backed up
`f.save` writes A.png
In the `try` block, no exception occurs while working with image, and `os.rename()` executes
Thus the resized image is written at `static_filepath`

Case 2-

User takes a non-image file B.png (by renaming any other type of file to png) and tries to upload
The current logo is backed up by first `with` block
`f.save` writes B.png (which is a valid file, but not a valid image file)
Pillow raises `IOError` in the `try` block while trying to read the file, and `os.rename()` does not execute
In the `except` block, second `with` overwrites `static_filepath` with previously backed up image, which is not broken
Thus, the broken image is discarded, and a warning is shown

If the `with` blocks are removed, the non-image file will be saved. So I they are not unnecessary in the current approach.

dachary:
you need to add `libjpeg-dev` in molecule/builder/Dockerfile which is where the package is built and the wheelhouse populated (the development environment is different, reason why it succeeds). Without it the build fails with:
<pre>
The headers or library files could not be found for jpeg,

a required dependency when compiling Pillow from source.



Please see the install instructions at:

   https://pillow.readthedocs.io/en/latest/installation.html



  File "<string>", line 1, in <module>

  File "/tmp/pip_build_root/pillow/setup.py", line 804, in <module>

    raise RequiredDependencyException(msg)

__main__.RequiredDependencyException:



The headers or library files could not be found for jpeg,

a required dependency when compiling Pillow from source.



Please see the install instructions at:

   https://pillow.readthedocs.io/en/latest/installation.html





----------------------------------------
  Failed building wheel for pillow
  Running setup.py bdist_wheel for psutil
  Destination directory: /tmp/z
  Running setup.py bdist_wheel for pycryptodomex
  Destination directory: /tmp/z
  Running setup.py bdist_wheel for scrypt
  Destination directory: /tmp/z
  Running setup.py bdist_wheel for sqlalchemy
  Destination directory: /tmp/z
  Running setup.py bdist_wheel for webassets
  Destination directory: /tmp/z
  Running setup.py bdist_wheel for wtforms
  Destination directory: /tmp/z
Successfully built cssmin flask-assets gnupg itsdangerous jsmin markupsafe psutil pycryptodomex scrypt sqlalchemy webassets wtforms
Failed to build pillow
</pre>
**but** it does not trigger an error because **pip wheel** never returns on error when it fails, it just outputs and error message.

aydwi:
@dachary Okay

dachary:
For the record https://github.com/freedomofpress/securedrop/pull/3065 will fail the build in a more debuggable way in the future.

dachary:
@aydwi actually you need https://github.com/freedomofpress/securedrop/pull/3065 (you can cherry-pick the two commits it contains) otherwise adding `libjpeg-dev` in `molecule/builder/Dockerfile` will not have any effect (you can read the diff and the commit messages at #3065 if you would like to know more).

securedrop/journalist_app/admin.py
heartsucker: That wouldn't make sense for the API to do that because you'd always have to create a new file at the destination before doing a move. Also:

```
heartsucker@pythagoras:~$ mkdir /tmp/wat
heartsucker@pythagoras:~$ cd !$
cd /tmp/wat
heartsucker@pythagoras:/tmp/wat$ touch foo
heartsucker@pythagoras:/tmp/wat$ python2 -c 'import os; os.rename("foo", "bar")'
heartsucker@pythagoras:/tmp/wat$ ls
bar
```

securedrop/journalist_app/admin.py
```diff
@@ -33,9 +35,30 @@ def manage_config():
...
+                with Image.open(static_filepath) as im:
+                    imcopy = im.copy()
+                    imcopy.thumbnail((500, 450), resample=3)
+                    imcopy.save(temp_static_filepath, "PNG")
+                if os.path.exists(static_filepath):
+                    os.rename(temp_static_filepath, static_filepath)
+                flash(gettext("Image updated."), "logo-success")
+
+            except Exception:
+                with open(static_filepath, 'w') as img_file:
```
heartsucker: I can't find the initial comment on this (or maybe it was just in chat), but the logic for catching errors shouldn't include rewriting the file from memory contents. That's more error prone than other solutions. This is also fixed by adding and `sd_logo` endpoint (maybe named more descriptively) that contains a switch for selecting the correct logo.

heartsucker:
CI is failing with this:

> FAILED testinfra/app/test_appenv.py::test_ensure_logo

I think this might be because we are messing with the original logo file. If we create a new file then move it with `os.rename` it will not have the correct permissions. The logo needs to be owned by `root`.

aydwi:
@heartsucker We wouldn't wanna revert to in-place resizing of logo. Is there a way permissions can be granted? Anyways I'm working on a way to handle this, will share soon.

securedrop/journalist_app/admin.py
aydwi: I'll remove the unnecessary guard then

heartsucker:
@aydwi In chat, @dachary corrected me that I have it backwards. The logo needs to be owned by `www-data`, and during tests it ends up owned by root. I also said that I think we shouldn't be overwriting the original `logo.png`. We should add an endpoint that:

- checks for the existence of a file under `SECUREDROP_DATA_ROOT/admin/logo.png` and serves that
- OR if it's missing serves the original `static/i/logo.png`

When we do this, we don't have to ever touch the original `logo.png`. This will unbreak the `testinfra` test and also have our code be more robust against failures.

aydwi:
I'm working on it

dachary:
In case you don't know about it @aydwi github does not notify when you push new commits. You should explicitly ping @heartsucker and let him know you have something new addressing his comments. If you already knew just ignore me :-P

securedrop/journalist_app/admin.py
```diff
@@ -35,24 +35,14 @@ def manage_config():
...
                                            "static/i/logo.png")
-            temp_static_filepath = os.path.join(config.SECUREDROP_ROOT,
-                                                "static/i/temp_logo.png")
-            with open(static_filepath) as img_file:
-                current_img = img_file.read()
-            f.save(static_filepath)
-
             try:
-                with Image.open(static_filepath) as im:
+                with Image.open(f) as im:
```
heartsucker: I'm not sure why you're copying the image and than saving the copy. The original is still being overwritten so we can still end up in a place where we have a broken image.

aydwi: I was trying to work with a copy of image instead of modifying it in place. That is unnecessary though, I'll remove it.

dachary:
@aydwi a gentle reminder about the need to rebase this pull request

aydwi:
@dachary Sure. I'll rebase it with develop 🙂

aydwi:
@heartsucker Pushed the changes to implement the endpoint approach and squashed the commits

dachary:
@aydwi you need another rebase because changes to tests were merged in the past 24h. Good practice :-)

aydwi:
@dachary Will do

kushaldas:
@aydwi you should also clean up the commit message in this PR.

aydwi:
@kushaldas Sure

heartsucker:
I'm going to merge this because it is a strict improvement over the current behavior. However, I am going to open up some followup tickets to tighten this up a little bit (for example, if we want to save in `/var/lib/securedrop` instead of `static` because then we get slightly nicer automatic cache/busting behavior).

PR #3064, opened 2018-02-23
Fixes #3063 uses io.open call everywhere

### LINKED ISSUE #3063: [RFE] Use io.open instead of open function call in the codebase
# Feature request

## Description

We should start using [`io.open`](https://docs.python.org/2.7/library/io.html#io.open) instead of the builtin function `open`  in Python2 so that we can move another step towards Python3. 

In Python3 `open` function call is actually the same of `io.open`. But, that is not the same in Python2. That is why moving to `io.open` in Python2 is [step forward](https://docs.python.org/3/howto/pyporting.html#update-your-code) to Python3 land.

Always remember, [the clock is ticking](https://docs.python.org/3/howto/pyporting.html#update-your-code).

kushaldas:
We are now using `io.open` call instead of `open` builtin.
This will help us to move forward one step to the Python3 world.
https://docs.python.org/3/howto/pyporting.html#update-your-code

## Status

Ready for review

## Description of Changes

Fixes #3063 

## Testing

`make test`

### DIFF
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -26,6 +26,7 @@ instances.
 import argparse
 import logging
 import os
+import io
 import re
 import string
 import subprocess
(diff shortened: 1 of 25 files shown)


kushaldas:
>securedrop/static/i/logo.png should be removed.

:( Will remove that.

kushaldas:
@heartsucker I was waiting for this, should have mentioned in the message above. Can I also make all of them into `with` statement in the same PR?

kushaldas:
Nah, it broke properly. Will fix these first.

redshiftzero:
Marking this as "pending additional work" for now - ping when you'd like reviewers to take another look @kushaldas

kushaldas:
>You seem to still be missing a few:

Oh they used to be in a different file, updating the patch again.

heartsucker:
Thanks @kushaldas

PR #3065, opened 2018-02-23
build: pip wheel must fail when it does not build a python module

dachary:
## Status

Ready for review

## Description of Changes

It's not a catastrophy to not fail when pip wheel fails because it will eventually fail when installed. But it's non-intuitive to see a failure during installation and much easier to debug if it fails early.

## Testing

* Add **pillow==5.0.0** to `securedrop/requirements/securedrop-app-code-requirements.txt`
* make build-debs # it must fail
* Add libjpeg-dev to **molecule/builder/Dockerfile**
* make build-debs # it must succeed

## Deployment

N/A

### DIFF
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/build_securedrop_app_code_deb.yml
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/build_securedrop_app_code_deb.yml
@@ -33,10 +33,11 @@
 - include: sass.yml
 
 - name: Create pip wheel archive for Debian package requirements.
-  command: >
-    pip wheel
-    -r {{ securedrop_code_filtered }}/requirements/securedrop-app-code-requirements.txt
-    -w {{ securedrop_app_code_deb_dir }}/var/securedrop/wheelhouse
+  shell: |
+    pip wheel \
+      -r {{ securedrop_code_filtered }}/requirements/securedrop-app-code-requirements.txt \
+      -w {{ securedrop_app_code_deb_dir }}/var/securedrop/wheelhouse 2>&1 | tee /tmp/w.out
+      ! grep -i --quiet 'Failed to build' /tmp/w.out
   tags: pip
 
 - include: translations.yml
--- a/molecule/builder/Dockerfile
+++ b/molecule/builder/Dockerfile
@@ -1,5 +1,5 @@
-# ubuntu:14.04 as of 2017-12-15
-FROM ubuntu@sha256:084989eb923bd86dbf7e706d464cf3587274a826b484f75b69468c19f8ae354c
+ARG IMAGE
+FROM $IMAGE
 
 RUN apt-get -y update && apt-get install -y python rsync sudo bash \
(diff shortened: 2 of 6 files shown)


dachary:
@msheiny this is modifying something you did, could you please disagree with me ? :-D

heartsucker:
```
$ make build-debs 
/bin/bash: molecule: command not found
Makefile:108: recipe for target 'build-debs' failed
```

We should include setting up the `venv` as a dependency for `make build-debs`. *grumbles*

dachary:
@heartsucker thanks for the quick review :-)

redshiftzero:
@msheiny you cool with this one?

msheiny:
checking @redshiftzero ...

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/build_securedrop_app_code_deb.yml
```diff
@@ -33,10 +33,11 @@
...
 - name: Create pip wheel archive for Debian package requirements.
-  command: >
-    pip wheel
-    -r {{ securedrop_code_filtered }}/requirements/securedrop-app-code-requirements.txt
-    -w {{ securedrop_app_code_deb_dir }}/var/securedrop/wheelhouse
+  shell: |
+    pip wheel \
+      -r {{ securedrop_code_filtered }}/requirements/securedrop-app-code-requirements.txt \
+      -w {{ securedrop_app_code_deb_dir }}/var/securedrop/wheelhouse 2>&1 | tee /tmp/w.out
+      ! grep -i --quiet 'Failed to build' /tmp/w.out
```
msheiny: Some super fancy ansible pros would gasp at the usage of error handling like this.... luckily, I think this is fine to do this here ;) Ive seen others break this logic up over two tasks (one task runs, the other task checks the output of the first and fails conditionally with text present). I honestly think the way you went makes more sense.

molecule/builder/Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM ubuntu:14.04
+# ubuntu:14.04 as of 2017-12-15
+FROM ubuntu@sha256:084989eb923bd86dbf7e706d464cf3587274a826b484f75b69468c19f8ae354c
```
msheiny: I don't like embedding the hash right into the Dockerfile. It's more annoying to over-ride programtically in the future if need be. I got a couple commits coming in with some slightly tweaked logic.

dachary: cool. Note that it's also used in other Docker files so maybe you'd like to fix in a uniform way.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/build_securedrop_app_code_deb.yml
dachary: :-) I'm glad we agree on this.

dachary:
@msheiny :+1:

PR #3066, opened 2018-02-23
docker: move the docker image capture logic in ticker

dachary:
## Status

Ready for review

## Description of Changes

If DOCKER_BUILD_VERBOSE=true is set, do not capture the docker image
output which is useful for debuging.

## Testing

* make -C securedrop dev
* DOCKER_BUILD_VERBOSE=true make -C securedrop dev
* make -C admin test
* DOCKER_BUILD_VERBOSE=true make -C admin test
* make ci-lint
* DOCKER_BUILD_VERBOSE=true make ci-lint

## Deployment

N/A

### DIFF
--- a/devops/scripts/ticker
+++ b/devops/scripts/ticker
@@ -12,13 +12,24 @@ function display_dots() {
 
 function ticker() {
     local pid
+    local out
+    local status
+
+    out="$(mktemp)"
     display_dots &
     pid="$!"
-    if "$@" ; then
-        status=0
+    status=0
+    if test -n "${DOCKER_BUILD_VERBOSE:-}" ; then
+        "$@" || status=1
+    else
+        echo "Run with DOCKER_BUILD_VERBOSE=true for more information"
+        "$@" >& "$out" || status=1
+    fi
+
+    if test $status = 0 ; then
         kill -USR1 $pid
     else
-        status=1
+        cat "$out"
         kill -USR2 $pid
     fi
     wait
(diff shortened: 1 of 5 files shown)


devops/scripts/ticker
```diff
@@ -12,13 +12,24 @@ function display_dots() {
...
+    out="$(mktemp)"
     display_dots &
     pid="$!"
-    if "$@" ; then
-        status=0
+    status=0
+    if test -n "${DOCKER_BUILD_VERBOSE:-}" ; then
+        "$@" || status=1
+    else
+        echo "Run with DOCKER_BUILD_VERBOSE=true for more information"
```
redshiftzero: We should also add this to the documentation, as I expect new contributors might miss this in the output but might see it in a [tip](http://docutils.sourceforge.net/docs/ref/rst/directives.html#specific-admonitions) box in the documentation, e.g. "Tip: If you're having trouble with your build, try..."

dachary:
@redshiftzero good idea for the documentation, thanks for suggesting it. Added to the [SQUASH: docker: move the docker image capture logic in ticker](https://github.com/freedomofpress/securedrop/pull/3066/commits/02855649ae782494d16514f85d42311742f405f4) commit.

dachary:
> I'm not opposed to always producing the full build output

My personal preference is to have full debug output always because I really dislike cryptic error messages. But I also understand why some people find that confusing and prefer terse output when all goes well so they can actually pay attention to the output when something bad happens.

heartsucker:
> My personal preference is to have full debug output always because I really dislike cryptic error messages.

Same. I would rather have full debug output always.

redshiftzero:
This looks great - just needs commits squashed prior to merge

PR #3079, opened 2018-02-28
i18n: use bin/dev-shell to run translate-{message,desktop}

dachary:
## Status

Ready for review

## Description of Changes

use bin/dev-shell to run translate-{message,desktop}

## Testing

* make -C securedrop translate
* see that .pot and .po files are modified

## Deployment

N/A

### DIFF
--- a/docs/development/i18n.rst
+++ b/docs/development/i18n.rst
@@ -250,18 +250,19 @@ The ``develop`` branch can now be merged into ``i18n`` as follows:
       $ git fetch lab
       $ git checkout -b i18n lab/i18n
       $ git merge origin/develop
-      $ make translate
+      $ make -C securedrop translate
 
-The ``manage.py`` command examines all the source files, looking for
-strings that need to be translated (i.e. gettext('translate me') etc.)
-and update the ``*.pot`` and ``*.po`` files, removing, updating and inserting
-strings to keep them in sync withe the sources. Carefully review the
-output of ``git diff``. Check ``messages.pot`` first for updated strings,
-looking for formatting problems. Then review the ``messages.po`` of one
-existing translation, with a focus on ``fuzzy`` translations. There is
-no need to review other translations because they are processed in the
-same way. When you are satisfied with the result, it can be merged
-with:
+The ``translate`` Makefile target relies on the ``manage.py`` command
+to examine all the source files, looking for strings that need to be
+translated (i.e. ``gettext('translate me')`` etc.)  and update the
+``*.pot`` and ``*.po`` files, removing, updating and inserting strings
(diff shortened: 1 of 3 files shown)


docs/development/i18n.rst
```diff
@@ -250,18 +250,19 @@ The ``develop`` branch can now be merged into ``i18n`` as follows:
...
-looking for formatting problems. Then review the ``messages.po`` of one
-existing translation, with a focus on ``fuzzy`` translations. There is
-no need to review other translations because they are processed in the
-same way. When you are satisfied with the result, it can be merged
-with:
+      $ make -C securedrop translate
+
+The ``translate`` Makefile target relies on the ``manage.py`` command
+to examine all the source files, looking for strings that need to be
+translated (i.e. gettext('translate me') etc.)  and update the
```
redshiftzero: Nit: Wrap `gettext('translate me')` in double backticks so it renders as fixed width

docs/development/i18n.rst
```diff
@@ -250,18 +250,19 @@ The ``develop`` branch can now be merged into ``i18n`` as follows:
...
+``*.pot`` and ``*.po`` files, removing, updating and inserting strings
+to keep them in sync withe the sources. Carefully review the output of
```
redshiftzero: withe -> with

docs/development/i18n.rst
dachary: done

docs/development/i18n.rst
dachary: done

dachary:
@redshiftzero thanks for catching the typos :-) Fixed and rebased and repushed !

dachary:
there is no reason for `make translate` to need config.py, let's wait for #3087 to be merged and this will be resolved.

redshiftzero:
hmm, it would be convenient for bin/dev-shell to have `config.py` generated anyway - that way a developer can hop into the container using `bin/dev-shell` and e.g. use the ORM interactively, otherwise they'll get the same `ImportError`. that said you're right that making that change is not necessary for _this_ PR given the changes in #3087 so will review that when tests are passing and then come back to this one :-)

dachary:
> hmm, it would be convenient for bin/dev-shell to have config.py generated anyway

I [documented how to do it manually](https://github.com/freedomofpress/securedrop/pull/3093/files). The `bin/dev-shell` is deliberately minimalist. But if this tip is not enough, maybe we can create a `bin/interactive` script that do things you always need when running interactively in the container? I must confess I very seldom run interactively so ... I'm not sure what we really need in this situation.

dachary:
Move the commit over to https://github.com/freedomofpress/securedrop/pull/3087

PR #3092, opened 2018-03-01
securedrop-admin: symlink ancient .venv to admin/.venv

### LINKED ISSUE #3088: ./securedrop-admin setup misses to update Ansible to 2.4.2
# Bug

This is in the update story from 0.5.2.

## Steps to Reproduce

To do QA work for `0.6-rc2` I moved into the tag, and ran the following command which worked as it was supposed to do:  `./securedrop-admin setup`.

Next, as part of the QA process I did the following steps.

- `source .venv/bin/activate`
- `cd install_files/ansible-base`
- `ansible-playbook -vv --diff sequredrop-qa.yml`

Please specify your environment if that is necessary to reproduce the bug (if in
doubt, include it).

## Expected Behavior

The ansible-playbook command should work properly.

## Actual Behavior

It failed saying  Ansible 2.4.2.x is required.

# Feature request

## Description

A short summary of the idea.

## User Stories

If appropriate

dachary:
## Status

Ready for review

## Description of Changes

Fixes #3088 

Prior to securedrop-admin moving to the admin directory, its
virtualenv directory was in the .venv directory at the toplevel of the
git tree. It is now in the admin/.venv directory instead.

For the regular admin following the instructions from the
documentation the existence of the ancient .venv will not be noticed
and can be ignored.

However the more knowledgeable SecureDrop admin who manually source
the .venv will be confused. We remove the .venv and symlink the new
admin/.venv so the change is transparent.

## Testing

* ./securedrop-admin setup # with 0.5.2
* see .venv exits and is a directory
* /securedrop-admin setup # with 0.6 rc2
* see .venv exists and is a symlink to admin/.venv
* source .venv/bin/activate
* ansible --version
ansible 2.4.2.0

## Deployment

1. SecureDrop admin following the documentation will not be affected. It will however fix a bug for the SecureDrop admin who source .venv even though it is not documented to exist 
2. N/A this is only for upgrades

### DIFF
--- a/securedrop-admin
+++ b/securedrop-admin
@@ -5,6 +5,10 @@
 # the top level of the SecureDrop repository
 d=$(dirname "$0")
 if test "$1" = "setup" ; then
+   # the .venv symlink is to not confuse devs used to
+   # finding the directory at the top-level, prior to 0.6
+   rm -fr "$d/.venv"
+   ln -s "$d/admin/.venv" "$d/.venv"
    shift
    exec python "$d/admin/bootstrap.py" "$@"
 else


dachary:
@kushaldas @msheiny does that sound like a sensible way to address the issue?

msheiny:
ehhhh i dunno.... maybe destroy and symlink instead? nah, i think im over-thinking it .. this will only affect us devs since we dont active advertise the location  of the virtualenv. i do have to ask though... would it just make more sense to move the location of the virtualenv back to the root folder from the securedrop-admin script instead?

dachary:
>  symlink instead

Good idea. Modified with that!

> move the location of the virtualenv back to the root folder from the securedrop-admin script instead

The admin directory is self contained right now and that would break this property.

redshiftzero:
I'll hold off on merge for 8+ hours to give the next shift of SecureDroppers a chance to test and comment if they like :) I've added @kushaldas and @msheiny to requested reviewers

Issue #3096, opened 2018-03-02
docs: Document `securedrop-admin update` and `securedrop-admin check_for_updates`


eloquence:
# Bug

## Description

SecureDrop 0.6 introduces a `securedrop-admin update` and a `securedrop-admin check_for_updates` command (https://github.com/freedomofpress/securedrop/pull/3026). This is great as it significantly simplifies the administration of Tails workstations. However, these commands are not referenced in the documentation yet.

It is worth noting that the `update` command also checks for updates, so running `check_for_updates` by itself is mostly useful as a pre-flight check. But it should also be documented.

## User Stories

As a SecureDrop administrator, I would like to get clear instructions for how to keep the workstations I am responsible for up-to-date.

redshiftzero:
Thanks for filing @eloquence - adding a section about updating the SecureDrop Tails workstation in the [Administrator Guide](https://docs.securedrop.org/en/latest/admin.html) should do the trick.

redshiftzero:
Closed by #3117 and #3120

PR #3117, opened 2018-03-06
Update Admin Guide to explain how to update SecureDrop workstations

redshiftzero:
## Status

Ready for review

## Description of Changes

Towards #3096 (not closing until changes backported into 0.6 release branch)

* Update admin guide on how to update workstations based on changes in 0.6

## Testing

Does this render and read well? 

## Deployment

Docs only


docs/admin.rst
```diff
@@ -52,15 +52,25 @@ Upgrade Docs`_ for further details on how to update the suggested firewall.
...
+
+* **Tails updates**: for each *Admin Workstation*, *Journalist Workstation*, and
+  *Secure Viewing Station*
+* **SecureDrop workstation updates**: for each *Admin Workstation* and
+  *Journalist Workstation*
 
-You should apply updates to your Tails drives as they are released, as they
-can contain critical security fixes. Subscribe to the `Tails RSS Feed`_ to be
+You should apply Tails updates to your Tails drives as they are released, as they
+often contain critical security fixes. Subscribe to the `Tails RSS Feed`_ to be
```
emkll: Out of scope of this PR, but what are you thoughts about this:
 `check_for_update` could run `/usr/local/bin/tails-version` and parse the running version, and we could store the latest version in a constant somewhere.

docs/admin.rst
```diff
@@ -187,6 +197,24 @@ possession when they attempt to log in to SecureDrop.
...
+In addition to Tails updates, once you have upgraded to SecureDrop 0.6 or later,
+you can keep the SecureDrop software up to date on each *Journalist Workstation*
+and *Admin Workstation* using the ``securedrop-admin`` CLI. In each workstation:
+
+.. code:: sh
+
+  cd ~/Persistent/securedrop
+  ./securedrop-admin setup
+  ./securedrop-admin update
+
```
emkll: Are there cases where we would have to run `./securedrop admin setup` after pulling the latest tag? Or do you think this be in release notes on an as-needed basis?

redshiftzero: This... is a good point. I'll add `securedrop-admin setup`, since if there are _not_ changes to the requirements then it doesn't hurt.

docs/admin.rst
redshiftzero: Clever! Running `/usr/local/bin/tails-version` might come in handy, e.g. if we know that a certain Tails release series no longer works well with our current workstation logic, we could display a message gently encouraging the user to upgrade Tails

redshiftzero:
Updated, thanks for review @kushaldas and @emkll :)

PR #1567, opened 2017-02-08
Update security slider instructions and source interface icons

redshiftzero:
In #1480, I removed Font Awesome icons from the source interface in order to direct sources to use Tor Browser's security slider without seeing breakage or weirdness. However, @micahflee noticed that since that PR was merged 1) the UI has changed - requiring an update to the instructions on the source interface and documentation and 2) the SVG icons used to replace Font Awesome also no longer work. Also the hover over effects did not work for the icons.

I've cherry-picked the commits from #1480 and follow-up PR #1522 and added commits to address the above issues in this PR, updating the docs and source interface and replacing the SVGs with PNGs. The rollover effects are back through some hacky CSS to change the image displayed on hover:

style_3_11

Important note: This PR _does not_ update `/etc/apparmor.d/usr.sbin/apache2`, also noted by @micahflee as a problem.

PR #3121, opened 2018-03-08
Add globe icon in the locale dropdown for better understanding

SaptakS:
As pointed out [here](https://forum.securedrop.club/t/language-menu-is-confusing-when-the-user-does-not-read-the-current-language/407), language menu is confusing when the user does not read the current language. Adding a globe icon helps understand that the dropdown is about locales for people who don't read the current language.

## Status
Ready for review

## Description of Changes
Added globe icon before(left of) the locale name and moved the trigger arrow after(right of) the locale name.
screen shot 2018-03-08 at 1 40 08 pm

## Testing

I did the following steps for testing:
1. Add locales in `SUPPORTED_LOCALES` list in the `securedrop/config.py` file
2. run `./i18n_tool.py translate-messages`
3. run `./manage.py run`
4. Open source interface and journalist interface to verify a globe icon appears in the language dropdown.

- [X] Unit and functional tests pass on the development VM


securedrop/source_templates/locales.html
```diff
@@ -2,8 +2,9 @@
 <div class="menu">
   <input id="menu-1-checkbox" class="menu__checkbox visually-hidden" type="checkbox" >
   <label for="menu-1-checkbox" class="menu__trigger">
-    <span class="menu__trigger-arrow">▼</span>
+    <i class="fa fa-globe menu__trigger-icon"></i>
```
redshiftzero: So the source interface is intended to render well even with the Tor Browser security slider set to its safest setting. With the Tor Browser set to safest, unfortunately this icon does not render:

screen shot 2018-03-08 at 9 23 48 am

However converting the icon to a PNG should do the trick - that's what we did with the other icons on the source interface - see #1567 for more.

SaptakS: Got it. Will do that. That answers my question of why some icons weren't visible in tor previously. Guess I fiddled with the safe settings.

SaptakS: Btw, just curious, why isn't this needed in journalist interface?

SaptakS: @redshiftzero thanks for clearing that out. If we are reducing the amount of JS code in journalist interface then I think replacing icons with PNGs is also an important thing to do.

redshiftzero: Totally, since you're making that PNG for the source interface anyway, if you elect to also use it on the journalist interface, it'd be a nice bonus to this PR :tada: (but not required for merge)

SaptakS: Sure. I will use the globe PNG for both the interfaces.

SaptakS: @redshiftzero done. I have added the svg file as well just in case sometime in future we think of resizing the image or something, it will be easier from svg file. Let me know if I should remove the svg file from the PR though. 😃

PR #2757, opened 2017-12-18
replace Google Authenticator with FreeOTP

dachary:
## Status

Ready for review

## Description of Changes

Fixes #2705

Google Authenticator is not Free Software, prefer FreeOTP which is available for Android and iOS and tested to work with SecureDrop. It also simplifies the instructions because the interface is the same on both iOS and Android.

## Testing

* Verify the FreeOTP setup instructions at securedrop/journalist_templates/account_new_two_factor.html and docs/servers_2fa.rst work
* Check the URLs at docs/terminology.rst are working
* Verify the images/manual/screenshots/journalist-admin_new_user_two_factor_totp.png screenshot looks good with the updated strings

## Deployment

N/A 

Although there are changes to the deployment instructions, they only relate to the App generating the two-factor authentication token. They can only change deployment if FreeOTP fails where Google Authenticator succeeds.

### If you made changes to the app code:

- [x] Unit and functional tests pass on the development VM

### If you made changes to documentation:

- [x] Doc linting passed locally

Issue #2705, opened 2017-12-08
Recommend a FOSS OTP tool


agharbeia:
# Feature request

## Description
Google's authenticator is proprietary, and FreeOTP,  which is mentioned once in the UI, strings is no longer maintained, even though it could probably still be functional on current Androids.

Perhaps the **main recommendation** and in-line examples should be of FOSS OTP tools, such as [andOTP](https://github.com/andOTP/andOTP) or [FreeOTP](https://github.com/freeotp/freeotp-android), by using their names in the user's directions, with Google's as a secondary recommendation, if at all.

## User Stories
When advising users in similar situation, I recommend FOSS whenever possible. This is even more relevant in contexts of security and privacy.

KwadroNaut:
FreeOTP still work as of Nougat, I see activity this month here: https://github.com/freeotp/freeotp-android/commits/master What exactly isn't maintained anymore? andOTP mentioned that they're doing this now as a side project. I like your suggestion.

dachary:
> Google's authenticator is proprietary

I did not realize that. There is a Free Software [version of Google Authenticator](https://github.com/google/google-authenticator) but it explicitly says the Android app is **not** Free Software. Does anyone know more about that ?

conorsch:
> FreeOTP [...] is no longer maintained

@agharbeia Citation? I agree with @KwadroNaut: there's still activity on the GitHub repo. Have you seen an announcement by the maintainers that the project is deprecated?

Big fan of recommending FreeOTP. We're already doing that (see 476223f26 via #1746), but happy to give it priority. We apparently only recommend FreeOTP in comparatively few locations:

```
$ grep -riPo '(freeotp|google(\s+auth\w+))' docs | cut -d: -f2 | sort | uniq -c
      1 freeotp
      3 FreeOTP
     18 Google Authenticator
```

Would gladly review a PR that mentions FOSS alternatives first and Google Authenticator second. Unclear to me what FOSS alternatives should be for iOS users—both FreeOTP and andOTP mentioned above appear to be Android-only.

agharbeia:
You are right, @KwadroNaut and @conorsch. it seems FreeOTP is still alive after all.
I should have rechecked before claiming that. The last time I had, a couple of months ago, it seemed to me to have been dormant for some time.

agharbeia:
You are right, @KwadroNaut and @conorsch. FreeOTP seems to still be alive!
[...]

PR #3125, opened 2018-03-08
journalist: better error message with the HOTP secret length is not 40

dachary:
## Status

Ready for review

## Description of Changes

better error message with the HOTP secret length is not 40

## Testing

N/A

## Deployment

N/A

### DIFF
--- a/securedrop/journalist_app/forms.py
+++ b/securedrop/journalist_app/forms.py
@@ -14,8 +14,8 @@ def otp_secret_validation(form, field):
     strip_whitespace = field.data.replace(' ', '')
     if len(strip_whitespace) != 40:
         raise ValidationError(gettext(
-            'Field must be 40 characters long but '
-            'got {num_chars}.'.format(
+            'HOTP secrets are 40 characters long - '
+            'you have entered {num_chars}.'.format(
                 num_chars=len(strip_whitespace)
             )))
 
--- a/securedrop/tests/test_journalist.py
+++ b/securedrop/tests/test_journalist.py
@@ -688,7 +688,7 @@ class TestJournalistApp(TestCase):
                                           is_admin=None,
                                           is_hotp=True,
                                           otp_secret='123'))
-        self.assertIn('Field must be 40 characters long', resp.data)
+        self.assertIn('HOTP secrets are 40 characters', resp.data)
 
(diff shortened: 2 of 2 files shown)


dachary:
@ei8fdb here you go

ei8fdb:
Merci 💥

redshiftzero:
Good call on this string improvement @ei8fdb! 

Thanks for making the PR @dachary - looks like there's just one test update (`TestJournalistApp.test_admin_add_user_yubikey_odd_length`) and then we're good to merge

ageis:
Just wanted to make a minor point for the future; HOTP/TOTP secrets are only 40 characters when they're based on SHA-1. There are some sites where they let you configure the HMAC for 2FA up to SHA256 and SHA512, so with an arbitrary change or algorithm deprecation you could have up to 64 or 128 characters in your HOTP secret.

@dachary I see with #2705 / #2757 we're now recommending FreeOTP everywhere; where this is explicitly configurable on the interface, if we are validating this string as 40 chars then I think you should also add a hint to the docs wherever journalist users are setting up 2FA that it _must_ be SHA1 with a 6-digit code. I'm not suggesting HMAC/SHA-1 is broken, but it may be prudent or necessary to support stronger hashes in the future, or there may come a time when it's not FreeOTP's default algorithm.

dachary:
@ageis I'm unclear where and how that should be phrased. Would you mind drafting a pull request for review?

kushaldas:
>@dachary I see with #2705 / #2757 we're now recommending FreeOTP everywhere; where this is explicitly configurable on the interface, if we are validating this string as 40 chars then I think you should also add a hint to the docs wherever journalist users are setting up 2FA that it must be SHA1 with a 6-digit code. I'm not suggesting HMAC/SHA-1 is broken, but it may be prudent or necessary to support stronger hashes in the future, or there may come a time when it's not FreeOTP's default algorithm.

This will be really helpful as the configuration window is otherwise very confusing.

dachary:
@redshiftzero the test is fixed.

@kushaldas @ageis do you feel the (good points) you raised should be addressed in this PR? Or would it be ok if I make a separate issue to followup?

ageis:
@dachary let's address that separately.

Issue #3124, opened 2018-03-08
/account/reset-2fa-hotp does not catch TypeError


dachary:
# Bug

## Description

When modifying the HOTP as a journalist (not in the admin interface), TypeError are not caught and a stack trace shows when debug is activated. Otherwise it would bubble to whatever catches server errors.

## Steps to Reproduce

* login as a journalist
* go to /account/account
* click /account/reset-2fa-hotp
* enter three letters
* click save

## Expected Behavior

An error message similar to what happens with /admin/reset-2fa-hotp

## Comments

The /admin/reset-2fa-hotp error handling code should be shared with /account/reset-2fa-hotp

redshiftzero:
Nice catch, I can reproduce this: 

![screen shot 2018-03-08 at 12 12 03 pm](https://user-images.githubusercontent.com/7832803/37173956-fb929a9c-22c9-11e8-9aea-7a40770332c3.png)

redshiftzero:
Fixed by #3128 and #3127

PR #3127, opened 2018-03-08
Handle TypeErrors on /account/reset-2fa-hotp, add regression tests

redshiftzero:
## Status

Ready for review

## Description of Changes

Towards #3124 (not closing until these changes are backported into the 0.6 release branch). 

 * Adds regression tests for the unhandled `TypeError`s for the user 2FA HOTP reset functionality. A `TypeError` is possible via the submission of an odd-length OTP secret, and another is possible via non-hex characters in the OTP secret.
 * Fixes by using a common function for validation shared with the admin route (which was not subject to the bug in #3124)

## Testing

I would test both user and admin HOTP reset for the following cases:
1. Odd length input: '123'
2. Non hex characters in input: 'zz'
3. Expected even length hex input e.g. '1234'

The following gif shows the expected behavior:

hotp-bugfix

## Deployment

No special considerations

- [x] Unit and functional tests pass in the development container

### DIFF
--- a/securedrop/journalist_app/account.py
+++ b/securedrop/journalist_app/account.py
@@ -6,7 +6,7 @@ from flask_babel import gettext
 
 from db import db
 from journalist_app.utils import (make_password, set_diceware_password,
-                                  validate_user)
+                                  validate_user, validate_hotp_secret)
 
 
 def make_blueprint(config):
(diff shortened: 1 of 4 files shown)


redshiftzero:
Thanks for the rapid review 😄

PR #3132, opened 2018-03-09
source: reword the tor2web warning for clarity

dachary:
## Status

Ready for review

## Description of Changes

clarify what publisher means in the context of tor2web

## Testing

N/A

## Deployment

N/A

### DIFF
--- a/securedrop/source_templates/tor2web-warning.html
+++ b/securedrop/source_templates/tor2web-warning.html
@@ -1,7 +1,7 @@
 {% extends "base.html" %}
 {% block body %}
 <h1>{{ gettext('Why is there a warning about Tor2Web?') }}</h1>
-<p>{{ gettext('<a href="tor2web.org">Tor2Web</a> is a proxy service that lets you browse Tor Hidden Services (.onion sites) without installing Tor. It was designed to facilitate anonymous publishing.') }}</p>
-<p>{{ gettext('Tor2Web only protects publishers, not readers. If you upload documents to us using Tor2Web, you are <strong>not anonymous</strong> and could be identified by your ISP or the Tor2Web proxy operators. Additionally, since Tor2Web sites typically do not use HTTPS, it is possible that your connection could be MITM\'ed by a capable adversary.') }}</p>
-<p>{{ gettext('If you want to submit information, you are <strong>strongly advised</strong> to install <a href="torproject.org/download.html.en">Tor</a> and use it to access our site safely and anonymously.') }}</p>
+<p>{{ gettext('Using <a href="tor2web.org">Tor2Web</a> to connect to SecureDrop will not protect your anonymity.') }}</p>
+<p>{{ gettext('It could be possible for anyone listening to Internet usage (your government, your Internet provider), to identify you.') }}</p>
+<p>{{ gettext('We <strong>strongly advise</strong> you to use the <a href="www.torproject.org/projects/torbrowser.html">Tor browser</a> instead.') }}</p>
 {% endblock %}


dachary:
suggested by @Jobava https://weblate.securedrop.club/translate/securedrop/securedrop/ro/?checksum=f65986a0192f8403

securedrop/source_templates/tor2web-warning.html
```diff
@@ -2,6 +2,6 @@
 {% block body %}
 <h1>{{ gettext('Why is there a warning about Tor2Web?') }}</h1>
 <p>{{ gettext('<a href="tor2web.org">Tor2Web</a> is a proxy service that lets you browse Tor Hidden Services (.onion sites) without installing Tor. It was designed to facilitate anonymous publishing.') }}</p>
-<p>{{ gettext('Tor2Web only protects publishers, not readers. If you upload documents to us using Tor2Web, you are <strong>not anonymous</strong> and could be identified by your ISP or the Tor2Web proxy operators. Additionally, since Tor2Web sites typically do not use HTTPS, it is possible that your connection could be MITM\'ed by a capable adversary.') }}</p>
+<p>{{ gettext('Tor2Web only protects publishers (owners of hidden web services), not readers. If you upload documents to us using Tor2Web, you are <strong>not anonymous</strong> and could be identified by your ISP or the Tor2Web proxy operators. Additionally, since Tor2Web sites typically do not use HTTPS, it is possible that your connection could be MITM\'ed by a capable adversary.') }}</p>
```
heartsucker: We introduce a term ("publishers"), explain it, then never use it again. I feel like this whole bit should be rewritten to leave out the details about who's protected by hidden services since the enduser might not now what they are. Also, I think the canonical way to name them is now "onion services."

dachary: @heartsucker do you have a better wording in mind? Not being a native English speaker, I'm a bit hesitant to try that myself ;-)

heartsucker: Actually, that seems be covered by the previous paragraph. I think we could actually cut out this whole sentence:

> Tor2Web only protects publishers (owners of hidden web services), not readers.

it's a bit redundant and adds terms that users don't need to actually know in order to understand what's happening.

dachary: True! I changed that to use exactly the same terms as the sentence above so it is clearer.

heartsucker: That kinda addresses it, but I still think we might be over explaining here. Can @ei8fdb give some input here maybe?

dachary:
@heartsucker I remember being confused myself when I first discovered the notion of **Tor Hidden Service** and it was not immediately clear that it is a web server in this case, hence the publisher is the owner of the **Tor Hidden Service**. I maybe thought there were third parties involved, maybe for anonymization purposes?

ei8fdb:
At the moment, I can't give a complete answer. This is difficult, if we were to give a "simple" message we would be making assumptions, and this is an important moment for the source - "Do I upload documents via Tor2Web, or do I install Tor Browser?" 

My thoughts are something like the following:

> Using SecureDrop with Tor2Web will not protect your anonymity. 
> 
> It could be possible for anyone listening to Internet usage (your government, your Internet provider),  to identify you.
> 
> If you want to submit information to {ORGNAME} <strong>we strongly advise</strong> installing the Tor browser {LINK-TO-TOR-BROWSER-DOWNLOAD} and upload documents using that.

dachary:
Since nobody objected to @ei8fdb rewording, I went a head and used it. IMHO it is better but a native English speaker should also review ;-)

kushaldas:
This looks good to me. :+1: from me.

securedrop/source_templates/tor2web-warning.html
```diff
@@ -1,7 +1,7 @@
...
-<p>{{ gettext('Tor2Web only protects publishers, not readers. If you upload documents to us using Tor2Web, you are <strong>not anonymous</strong> and could be identified by your ISP or the Tor2Web proxy operators. Additionally, since Tor2Web sites typically do not use HTTPS, it is possible that your connection could be MITM\'ed by a capable adversary.') }}</p>
-<p>{{ gettext('If you want to submit information, you are <strong>strongly advised</strong> to install <a href="torproject.org/download.html.en">Tor</a> and use it to access our site safely and anonymously.') }}</p>
+<p>{{ gettext('Using SecureDrop with <a href="tor2web.org">Tor2Web</a> will not protect your anonymity.') }}</p>
+<p>{{ gettext('It could be possible for anyone listening to Internet usage (your government, your Internet provider), to identify you.') }}</p>
+<p>{{ gettext('If you want to submit information to us we <strong>strongly advised</strong> installing the <a href="www.torproject.org/projects/torbrowser.html">Tor browser</a> and upload documents using that.') }}</p>
```
heartsucker: Current: 

> 'If you want to submit information to us we <strong>strongly advised</strong> installing the <a href="www.torproject.org/projects/torbrowser.html">Tor browser</a> and upload documents using that.'

Should be

> 'If you want to submit information to us we <strong>strongly advised</strong> installing the <a href="www.torproject.org/projects/torbrowser.html">Tor browser</a> using it to visit SecureDrop.'

We should always say files instead of documents, and we also want to keep the messaging consistent in that we accept files and messages alike. We could replace "using it to visit SecureDrop" with "using it to send us files and messages," but that would imply that visiting SD without TB is safe when it is not.

heartsucker: @ei8fdb I would like your input on this since I might be a bit off with my analysis of the exact phrasing here.

dachary: I suggest we just remove **and upload documents using that** entirely. It is implied since the sentence starts with **If you want to submit information to us**. Leaving it sounds like: **If you want to send a us message, use the phone to send us a message** ;-)

heartsucker: Event better.

dachary: What about **If you want to submit information to us we <strong>strongly advise</strong> you to use the...** ?

dachary: Or even **We strong advise you to use the tor browser instead.**

dachary:
I reworded once more @ei8fdb @heartsucker what do you think?

ei8fdb:
Halo! I saw this thread today but only got time now to focus on it.

I think @dachary's take on my wording is good. 👍 

I think it conveys:

-  the important message ("Using Tor2Web (essentially) removes your anonymity"),

- puts the decision in the hands of the user (i.e. "If you need to be anonymous, don't use Tor2Web"), and,

- provides them with a recommended action (i.e. "Download the Tor browser, and visit the Tor network address").

Right now, it is our best judgement on how to write content for users. The ideal situation is we get an understanding fro users - this is why user research is so important.

👍

PR #3140, opened 2018-03-11
Journalist pytest fixtures

heartsucker:
## Status

Ready for review

## Description of Changes

Towards #2877.

This starts moving some of the journalist interfaces tests to the new pytest fixtures. This isn't all of the test in `test_journalist.py`, because some of them are going to be a little complicated. Also, these PRs are giant nightmares to review, so I'm trying to get this in so that between this and #2948, there shouldn't be any blockers on other contributors refactoring the other tests (except functional which will be another huge block).

## Testing

```
cd securedrop
make test TESTFILES=test/test_journalist.py
```

### DIFF
--- a/securedrop/tests/test_journalist.py
+++ b/securedrop/tests/test_journalist.py
@@ -271,6 +271,24 @@ class TestPytestJournalistApp:
         text = resp.data.decode('utf-8')
         assert edit_account_link in text
 
+    def test_user_has_link_to_edit_account_page_in_index_page(self,
+                                                              journalist_app):
+        with journalist_app.app_context():
+            user, password = utils.db_helper.init_journalist()
+            username = user.username
+            otp_secret = user.otp_secret
+
+        with journalist_app.test_client() as app:
+            resp = app.post('/login',
+                            data=dict(username=username,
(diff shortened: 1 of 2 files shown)


securedrop/tests/test_journalist.py
```diff
@@ -760,8 +906,8 @@ def test_admin_renames_user(self):
         # therefore asserting it has been created
         Journalist.query.filter(Journalist.username == new_user).one()
 
-    @patch('journalist_app.admin.current_app.logger.error')
-    @patch('journalist_app.admin.Journalist',
+    @patch('journalist_app_module.admin.current_app.logger.error')
+    @patch('journalist_app_module.admin.Journalist',
```
redshiftzero: As is, this patching should be applied to `journalist_app` e.g. `@patch('journalist_app.admin.Journalist', ...`

redshiftzero:
Thanks for the PR :)The test failure in CI is legitimate (I do also see the same result locally - can you double check your test report in the PR description?)

heartsucker:
@redshiftzero Repushed. Sorry about that. I think I lost something in all the rebases D:

PR #3174, opened 2018-03-18
tests: associate screenshots with source strings and upload them to weblate

dachary:
## Status

:red_circle: Work in progress :red_circle: 

## Description of Changes

TBD

## Testing

* DOCKER_RUN_ARGUMENTS='-e PAGE_LAYOUT_CAPTURE_L10N=true -e PAGE_LAYOUT_LOCALES=en_US' ./bin/dev-shell ./bin/run-test -v tests/pages-layout

## Deployment

N/A


dachary:
@heartsucker I'm moderately happy with [the l10n capture](https://github.com/freedomofpress/securedrop/pull/3174/commits/96930d37d9f9029a75af2c37f72a73b972b90896) and wonder if you have advices on how to make it more elegant?

securedrop/tests/functional/functional_test.py
```diff
@@ -43,6 +47,44 @@ def __call__(self, driver):
             return True
 
 
+@contextmanager
+def capturel10n(app, what):
```
heartsucker: I have not nearly enough idea what's happening here. Why are you `yield/return`-ing? Can you use longer variable names and function names?

dachary: It is necessary to yield so the mocks do not go out of scope. They need to exist for the duration of the context. Is there a better way to do this?

heartsucker: `yield`-ing is fine, but I've never seen the `yield/return` pattern like how you did. Can you explain that?

dachary: Oh, that... well, it's a way of doing nothing if the environment variable is not set. Very hackish I don't like it.

PR #3198, opened 2018-03-27
Update versions strings on develop to 0.6

redshiftzero:
## Status

Ready for review

## Description of Changes

Currently version strings point to `0.6~rc2`. Let's update to the latest release and then going forward we can add these version bumping commits to `develop` in preparation for the next release. 

 - cherry picked from commit 4cbab07fda4fb18264a4901042e4ac6c9860be47

## Testing

Does CI pass? Are there `0.6~rc2` version strings anywhere left on `develop`?

## Deployment

Already in prod


dachary:
I propose we bump develop to **0.7~rc1**  instead. It has the advantage that develop has a higher version than 0.6 and we can trivially run upgrade tests. The commits that modify the 0.6* versions can be unique to the `release/0.6` branch, unless I'm missing something there is no need to get them in develop. When `release/0.7` is branched, the **0.7-rc1** tag can be set and the develop branch can be bumped to **0.8~rc1**  to stay ahead. And all 0.7* versions after that are unique to `release/0.7`. And so on.

What do you think?

redshiftzero:
Actually sure why not, I can also think of no compelling reason to wait (so if we're both wrong, this will be an interesting learning experience at minimum 🙃). But let me close this PR as the branch name is misleading, will start a new one and link back to this

dachary:
Agreed to share the blame if things go sideways :-P

PR #3210, opened 2018-03-31
Add user research evidence section to issue feature request template

redshiftzero:
## Status

Ready for review

## Description of Changes

* This is a followup from [the UX meeting on Monday](https://lab.securedrop.club/main/ux/wikis/ux-meetings/20180326) where we discussed adding a section in the issue template to point to relevant SecureDrop user research to better motivate feature requests. 

## Testing

Let me know if the wording could be improved here @ei8fdb @belenbarrospena and I can modify it 

## Deployment

N/a


dachary:
As a contributor, I would be intrigued by this **user research** question. What am I expected to provide? Why does it matter? Can I get help doing that? Is this going to require days of my time? Are there examples I could look into (https://lab.securedrop.club/main/ux/wikis/ux-design-decisions/tails-ws-gui-updater for instance)?

I took a look at http://opensourcedesign.net/resources/ and maybe there is something suitable there but I did not find it.

Ideas?

belenbarrospena:
I agree some background information would be good, but I wonder if the issue template is the right place for it. 

What about creating a page in the UX wiki about what the user research template section actually means and how to complete it? Then we could link to the page from the template.

dachary:
> What about creating a page in the UX wiki about what the user research template section actually means and how to complete it? 

Sounds like the best thing to do, indeed!

belenbarrospena:
Here is a first stab at it 

https://lab.securedrop.club/main/ux/wikis/User-research-evidence-in-feature-requests

redshiftzero:
This document looks excellent @belenbarrospena! I'll link out to this page in the template

PR #3075, opened 2018-02-27
Local automated upgrade testing

msheiny:
## Status

Ready for review

## Description of Changes

Changes proposed in this pull request:

Fixes first check-box in #3018 

Add two scenarios:
* `vagrant_packager` -- builds two vagrant boxes ready for redistribution
* `upgrade` --- fires up boxes from the first scenario (pulled from s3), and chucks all locally built deps to a local apt server.

## Testing

How should the reviewer test this PR?

* `Caveat` - this PR testing will only work under a system with libvirt/kvm. There is another ticket to also add support for virtualbox. 

Packager testing:

* Run `make vagrant_package` - be prepared to enter a sudo password about 20 min in . 
* you'll find the vagrant box files in molecule/vagrant_packager/build/*.box - you can manually import those if you want to mess around with em
* Peek at `molecule/vagrant_packager/push.yml` dont run that now... but glance it over :)

Upgrade testing (local deb packages):
* `make build-debs`  -  make you sure have a version of debian packages that are higher than `0.6` if the `build/` directory. If you don't, running this command should give you `0.7.0~rc1` 
* Run `molecule converge -s upgrade` to get `.0.6` SD servers up. You'll get passed an onion address at the end. Navigate to that in a web-browser and note the version at the bottom.
* Run `molecule side-effect -s upgrade`  - Navigate back to the onion address and confirm version has bumped.

Upgrade testing (apt-test proxy):
1. Run `molecule converge -s upgrade` to get `.0.6` SD servers up.
2. Run `molecule login -s upgrade --host app-staging`  - To get a shell onto the app server
3. In the app server, run `sudo apt-get update` (confirm no ERRORS , you'll see a warning about duplicate repos but thats not an error and is a diff issue)
4. Run `apt-cache policy securedrop-config` (same terminal ^^) and you should see all the packages pulled in from above
5. Now back at the securedrop root dir run `QA_APTTEST=yes molecule converge -s upgrade -- --diff  -t apt`
6. Follow step `2 -> 4` again. This time you should see versions of securedrop-config pulled in from apt-test (this should have a few versions back and look like below):

```
vagrant@app-staging:~$ apt-cache policy securedrop-config                                                
securedrop-config:
  Installed: 0.1.0+0.6
  Candidate: 0.1.1+0.7.0~rc3
  Version table:
     0.1.1+0.7.0~rc3 0
        500 https://apt.freedom.press/ trusty/main amd64 Packages
     0.1.1+0.7.0~rc2 0
[...]

Issue #1967, opened 2017-07-12
Re-land CUSTOM_NOTIFICATION feature


garrettr:
Follow-up to #1965. Reminder to re-land #1399 once we've resolved #1966.

conorsch:
When we reimplement this feature, let's try to avoid use of `config.py` entirely. Updating the custom notification shouldn't require running the Ansible scripts. An Admin should be able to input custom notification text via the Admin web interface. Not only would this be better UX, it would also completely sidestep the config update difficulties described in #1966).

redshiftzero:
Note that reimplementing this will require either #1419 (the better implementation stores this information in the database - we should be enabling administrators to modify much of the text appearing in the templates on the source and journalist interfaces) or #1966 to deploy

redshiftzero:
We should ensure that this feature is implemented in such as way as to handle unicode (needed for many other languages).

eloquence:
As part of re-landing this, I would suggest replacing all instances of the term "notification" with "notice". In UI terms, "notifications" are generally user-specific alerts/updates (such as the notification email we intend to send about new submissions). Overloading this term for instance-level notices that are permanently displayed is IMO a recipe for confusion.

redshiftzero:
for 0.9.0, re-implementation of this (it has the database migration, needs tests writing) is in `source-interface-notice`:


### Admin interface - Instance config page 

<img width="551" alt="screen shot 2018-06-05 at 3 31 01 pm" src="https://user-images.githubusercontent.com/7832803/41007913-fdc5ec34-68dc-11e8-9372-9df5beb33ddc.png">

### Source interface footer

<img width="606" alt="screen shot 2018-06-05 at 3 31 09 pm" src="https://user-images.githubusercontent.com/7832803/41007914-fde4a534-68dc-11e8-852b-0fbf9d31db33.png">

Thought: We might also want to allow the admin to create multiple source notices for the different languages their instance supports

conorsch:
No longer blocked, now that we've landed db migrations. (cc @redshiftzero )

PR #3211, opened 2018-03-31
Use alembic to version control sqlite database

### LINKED ISSUE #1419: Schema migrations
Some new features (e.g. #1170) are going to require schema migrations. We should integrate e.g. [`sqlalchemy-migrate`](https://sqlalchemy-migrate.readthedocs.io/en/latest/) to make it easier for us to make changes to the databases on the production instances.

### LINKED ISSUE #3244: Add test harness to ensure alembic up/downgrades work as intended
# Feature request

## Description

Subtask of #1419

We need a test framework that ensures that every `alembic` database upgrade and downgrade applies cleanly.

As a note, whatever we do to load data into the database **cannot depend on** `models.py` because the database schema will always match `alembic`'s `head` revision, so using it to test an upgrade makes no sense since we need to use the out-of-date models. This means we have to hand craft code to do the inserts and selects that we want.

## Initial Proposal

Branch: [`migration-test-harness`](https://github.com/freedomofpress/securedrop/tree/migration-test-harness)

This proposal uses dynamic module loading and pytest parametrization to ensure that every new migration applies the tests in the same way. If a migration is not present, running `make test` will error out because it won't be able to find the correct module.

- Test harness: https://github.com/freedomofpress/securedrop/blob/cebd70fbccf866783b67406efba6023661f1b363/securedrop/tests/test_alembic.py#L109-L159
- Migration tester: https://github.com/freedomofpress/securedrop/blob/cebd70fbccf866783b67406efba6023661f1b363/securedrop/tests/migrations/migration_faac8092c123.py

## User Stories

As a dev, I want to know that my DB migration scripts work in a prod-like environment.

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #1419, fixes #3244

This adds [`alembic`](http://alembic.zzzcomputing.com/) to SD for revision controlling the sqlite database.

This includes an initial migration that exactly maps the current schema created by `db.create_all()` to the schema created by migration. All newly provisioned SD instances will have identical DB's to the ones currently out in the wild. A second revision is added to add the security pragmas to current instances.

There is basic testing for the upgrade/downgrade scripts that scripts the `alembic/verisons/` directory and runs all of them. This means all new revisions get **by default** some kind of testing.

There are automatic backups of the DB if and only a migration would be applied.

## Testing

Checkout `develop` (with fresh debs), provision a staging VM, checkout this branch, re-provision with fresh debs. Test out the interface.

## Deployment

New and old instances have their DBs added to version control via `postinst`.


securedrop/.rsync-filter
```diff
@@ -1,4 +1,10 @@
 exclude config.py
+exclude *.pyc
```
dachary: This should not be necessary because everything is explicitly included. Why do you think it is?

install_files/securedrop-app-code/DEBIAN/postrm
```diff
@@ -19,14 +19,20 @@ set -e
 # the debian-policy package
 . /usr/share/debconf/confmodule
 
+clean_pyc() {
+    find /var/www/securedrop/alembic -name '*.pyc' -delete
```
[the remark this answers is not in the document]
heartsucker: `lintian` disagrees with you. [It says](https://lintian.debian.org/tags/command-with-path-in-maintainer-script.html) we *shouldn't* prepend the full path because `$PATH` is "set to a reasonable value."

install_files/securedrop-app-code/DEBIAN/postinst
```diff
@@ -16,6 +16,38 @@ set -x
 # for details, see http://www.debian.org/doc/debian-policy/ or
 # the debian-policy package
 
+database_migration() {
+    database_dir='/var/lib/securedrop'
+    database_file="$database_dir/db.sqlite"
+
+    # Ensure the directory exists (needed for fresh installs)
+    if [ ! -e "$database_dir" ]; then
+        mkdir -m 0770 "$database_dir"
```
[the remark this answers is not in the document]
heartsucker: Just as a note, `-m` and `-p` are incompatible, so when you see that I use a `if/then` guard in the commit, that's why.

securedrop/alembic/versions/faac8092c123_enable_security_pragmas.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def upgrade():
+    conn = op.get_bind()
+    conn.execute(sa.text('PRAGMA secure_delete = ON'))
+    conn.execute(sa.text('PRAGMA auto_vacuum = FULL'))
+
+
+def downgrade():
+    pass
```
redshiftzero: Why no downgrade?

Btw, as a consistency check for migrations, I think a good practice is to do the upgrade, then a downgrade, and then verify that the operations went smoothly and that one is back where one started. If we don't do this, we should be aware we are taking on technical debt as we won't be able to move between database revisions smoothly if needed

securedrop/alembic/versions/faac8092c123_enable_security_pragmas.py
```diff
@@ -0,0 +1,26 @@
...
+revision = 'faac8092c123'
+down_revision = '15ac9509fc68'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    conn = op.get_bind()
+    conn.execute(sa.text('PRAGMA secure_delete = ON'))
+    conn.execute(sa.text('PRAGMA auto_vacuum = FULL'))
```
redshiftzero: Good call on adding this revision - this is a good opportunity for us to get instances that installed after we established these pragmas consistent with recent installs

securedrop/alembic/versions/faac8092c123_enable_security_pragmas.py
heartsucker: This was intentional. From [the docs](https://sqlite.org/pragma.html):

> The default setting for auto-vacuum is 0 or "none", unless the SQLITE_DEFAULT_AUTOVACUUM compile-time option is used. The "none" setting means that auto-vacuum is disabled.

and

> The default setting for secure_delete is determined by the SQLITE_SECURE_DELETE compile-time option and is normally off.

Because I'm not totally sure what the actual state of the DB is before this migration, I'm leaving it alone.

heartsucker:
> Related: we should definitely prioritize getting #3075 merged for 0.7 - that PR is a good place to add the script that loads in a bunch of prod like data into sqlite prior to upgrade for realistic testing of schema migrations in CI

I think these can be done in pytest as integration tests which would actually make it easier for developers to test that their migrations do the right thing.

redshiftzero:
We discussed in a call today that we should maintain this as a feature branch and hold on merging into `develop` until the required automated testing (on a prod-like database) is in place. When that's done and we all agree that we can safely run database migrations in `postinst` on production SecureDrop instances, then this is ready to merge in.

heartsucker:
Also we need #3273 reviewed / merged.

install_files/securedrop-app-code/DEBIAN/control
```diff
@@ -6,5 +6,5 @@ Homepage: https://securedrop.org
 Package: securedrop-app-code
 Version: 0.7.0~rc1
 Architecture: amd64
-Depends: python-pip,apparmor-utils,gnupg2,haveged,python,python-pip,secure-delete,sqlite,apache2-mpm-worker,libapache2-mod-wsgi,libapache2-mod-xsendfile,redis-server,supervisor,securedrop-keyring,securedrop-config,libjpeg-dev
+Depends: python-pip,apparmor-utils,gnupg2,haveged,python,python-pip,secure-delete,sqlite3,apache2-mpm-worker,libapache2-mod-wsgi,libapache2-mod-xsendfile,redis-server,supervisor,securedrop-keyring,securedrop-config,libjpeg-dev
```
redshiftzero: This change will be a problem for existing installs due to #3376 - is it strictly necessary?

heartsucker: Sort of. The docs say that using the `sqlite3 $path_to_db .database` command is the most future-proof way to create a database file. That said, right now using `touch` does seem to work, but that goes against what the developers say we should do.

heartsucker: Aslo use of `sqlite` to do this creates a database file that is incompatible with version 3 and causes errors about being unable to open the file.

heartsucker:
@redshiftzero I saw this comment in the email about your review but it's missing here:

> I don't have a problem with this being here for the sake of being explicit, but isn't the else condition below never going to execute due to `os.environ['SECUREDROP_ENV'] = 'dev'` on line 23?

This is on the `reset` function in `manage.py`.

I didn't realize we were setting the env var at the head of the file, and tbh I think we should stop doing that everywhere. These should all be explicit options set on the CLI (etiher as an env var or as a CLI option) with the one exception being setting `SECUREDROP_ENV = 'testing'` at the head of `conftest.py`.

It's too hard to track down where options are being set, and importing a module can permanently alter the behavior of the rest of the application (e.g., some tests run after the manage tests have the wrong env var).

heartsucker:
Note: I had to rebase to get the CI w/ rebase tests to pass because it couldn't merge the very first commit.

install_files/securedrop-app-code/DEBIAN/postinst
```diff
@@ -16,8 +17,55 @@ set -x
...
+        sqlite_tables=$(sqlite3 "$database_file" .tables)
+
+        if ! echo "$sqlite_tables" | grep -q journalists; then
+            # This is a fresh install, run the full migrations.
+            alembic upgrade head
+        else
+            # This is an old install, so mark that the scheme is up to date
+            # as of the version of migration 15ac9509fc68_init.py when we first
+            # started version controlling the database
+            alembic stamp 15ac9509fc68
```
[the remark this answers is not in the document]
heartsucker: What do you mean "template this out"?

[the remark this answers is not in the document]
msheiny: Thinking something like this so its not hard-coded. This value will theoretically change frequently right? If my understanding is incorrect then I retract my comment :)

https://github.com/freedomofpress/securedrop/blob/4bd7d8f218317083427dfc6ee994c6a673795f05/install_files/securedrop-grsec/DEBIAN/control.j2#L3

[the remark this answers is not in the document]
heartsucker: Ah. No, we want to hard coded because the migration needs to always start at the `init` revision. This line only happens if the database is empty *and* not under version control yet.

eloquence:
We agreed to try to get this merged in the current sprint (5/30-6/13), which will need to happen if we want to ship it with 0.8.0. Issues flagged during sprint planning:

- needs further investigation of backup/recovery behavior
- needs full QA by another developer, including making and testing an actual schema change

install_files/securedrop-app-code/DEBIAN/control
```diff
@@ -6,5 +6,5 @@ Homepage: https://securedrop.org
 Package: securedrop-app-code
 Version: 0.8.0~rc1
 Architecture: amd64
-Depends: python-pip,apparmor-utils,gnupg2,haveged,python,python-pip,secure-delete,sqlite,apache2-mpm-worker,libapache2-mod-wsgi,libapache2-mod-xsendfile,redis-server,supervisor,securedrop-keyring,securedrop-config
+Depends: python-pip,apparmor-utils,gnupg2,haveged,python,python-pip,secure-delete,sqlite3,apache2-mpm-worker,libapache2-mod-wsgi,libapache2-mod-xsendfile,redis-server,supervisor,securedrop-keyring,securedrop-config
```
redshiftzero: Good news on this front: today I did an upgrade test of this PR in staging (with the Depends field as is here) and the upgrade went fine. This dependency change is ok because the [`sqlite3` package](https://packages.ubuntu.com/trusty/sqlite3) is in `security.ubuntu.com/ubuntu/ trusty-security/main`, so we don't hit #3376:

```
vagrant@app-staging:~$ apt-cache policy sqlite3
sqlite3:
  Installed: 3.8.2-1ubuntu2.1
  Candidate: 3.8.2-1ubuntu2.1
  Version table:
 *** 3.8.2-1ubuntu2.1 0
        500 http://archive.ubuntu.com/ubuntu/ trusty-updates/main amd64 Packages
        500 http://security.ubuntu.com/ubuntu/ trusty-security/main amd64 Packages
        100 /var/lib/dpkg/status
     3.8.2-1ubuntu2 0
        500 http://archive.ubuntu.com/ubuntu/ trusty/main amd64 Packages
```

heartsucker:
Note: It would be nice to remove `qa_loader.py` from the debian package for actual releases to prevent some super unlikely case of an admin running it and putting "hard coded" (from a fixed random seed) credentials into their DB.

install_files/securedrop-app-code/DEBIAN/postinst
```diff
@@ -16,8 +17,55 @@ set -x
...
+            # This is an old install, so mark that the schema is up to date
+            # as of the version of migration 15ac9509fc68_init.py when we first
+            # started version controlling the database.
+            alembic stamp 15ac9509fc68
+        fi
+    fi
+
+    if ! alembic current | grep -q '(head)'; then
+        date=$(date +%Y-%m-%d-%H-%M-%S)
+        cp /var/lib/securedrop/db.sqlite "/var/lib/securedrop/backups/$date-db.sqlite"
```
redshiftzero: Some of these databases might not be _that_ small, so especially for long-running instances, we'll need some way to clean up the database backups. How about: once the database migration occurs successfully, we delete the backup? (in `postinst`)

docs/development/database_migrations.rst
```diff
@@ -0,0 +1,171 @@
...
+~~~~~~~~~~~~~~~~~~~
+
+Once you are satisfied with your new model, ``alembic`` can auto-generate migrations using
+SQLAlchemy metadata and comparing it to the schema of an up-to-date SQLite database. To generate a
+new migration use the following steps.
+
+.. code-block:: none
+
+    cd securedrop/
+    ./bin/dev-shell
```
redshiftzero: Then after hopping into the container we need to:

```
source bin/dev-deps
maybe_create_config_py
```

or we get an `ImportError` (due to no `config.py`) from `alembic/env.py` when we generate the migration

redshiftzero:
CI failure here is legit due to postinst logic, also the development environment is no longer working (found this out while working on a test db migration for #1967) on this branch, both those things need investigating and resolving

Roughly my test plan here so far is:

# Test drive stamping: 0.7.0 -> 0.8.0

1. `git checkout 0.7.0`
2. `make build-debs`
3. `vagrant up /staging/`
4. `git checkout alembic`
5. `make build-debs`
6. `vagrant provision /staging/`

This I did last week and things looked good (changes have been made since then to the logic so it does need to be run through again prior to merge)

# Test drive first migration: 0.8.0 -> 0.9.0

1. `git checkout alembic`
2. `make build-debs`
3. `vagrant up /staging/`
4. `git checkout source-interface-notice` (or your favored branch with a database migration here)
5. `make build-debs`
6. `vagrant provision /staging/`

wip

# Other possible test cases

1. Case where there is insufficient disk space to do the backup -> current behavior is that the upgrade will fail and manual action will be required by admins to resolve

install_files/securedrop-app-code/DEBIAN/control
redshiftzero: Recapping gitter comments/discussions for posterity:

Directly using `sqlite` to create a database will create a version 2.x database. This version is *totally incompatible* with version 3 (see the sqlite project's history of database file format changes [here](https://www.sqlite.org/formatchng.html)).

Since at least SecureDrop 0.3, we have used `db.create_all()` to create databases in production, and this method does create a version 3 database. Thus no 2->3 migration issues should be encountered. This logic is tested via the 0.7->0.8 upgrade path testing.

redshiftzero:
I'm going to merge this in now, but @kushaldas if you get chance this week, I would still be interested in any thoughts you have on this one (especially any improvements that can be addressed prior to 0.8.0 being released).

heartsucker:
Eeeeeeeeeeeee this is so exciting :D :D :D :D

Issue #3216, opened 2018-04-02
Journalist codenames with spaces at the end are still valid


dachary:
# Bug

## Description

A journalist name with a space at the end cannot login

## Steps to Reproduce

* create a journalist user with a space at the end of the journalist name

## Expected Behavior

* I am able to login as a journalist even when my name ends with a space

## Actual Behavior

* I cannot login if my journalist user name ends with a space

redshiftzero:
I was not able to reproduce this behavior as described - let me know if I misunderstood: 

1. Using the Admin Interface, I create a user with a space at the end
2. I _am_ able to login as that user without issue

I wrote a unit test in #3218 to verify this and include the case in our test suite

dachary:
Double checking. Most likely I'm doing something stupid.

dachary:
I'm not able to repeat the problem. It appears I shot myself in the foot a few times and thought it was real... sorry for the noise!

PR #3218, opened 2018-04-02
Unit tests: Verify journalist username containing whitespace can login

redshiftzero:
## Status

Ready for review

## Description of Changes

Relates to #3216.

* There may be journalist accounts in production 
that have whitespace in the username, so we should ensure they are
able to login going forward by including this case in our test suite.

## Testing

Does this test pass? Does the logic properly test that a user with a whitespace-containing username can login? 

## Deployment

None, this is merely a guard for future Stewards of SecureDrop


redshiftzero:
thanks for reviewing!

PR #3256, opened 2018-04-11
Additional pytest fixtures for test_journalist.py

heartsucker:
## Status

## Description of Changes

Towards #2877 

- Adds additional pytest fixtures to `test_journalist.py`.
- Fixes a potential vulnerability where we may attempt to use TOTP on an account with HOTP or vice versa

## Testing

`make test`


securedrop/models.py
```diff
@@ -327,11 +327,17 @@ def set_hotp_secret(self, otp_secret):
...
+        else:
+            raise ValueError('{} is not using TOTP'.format(self))
 
     @property
     def hotp(self):
-        return pyotp.HOTP(self.otp_secret)
+        if not self.is_totp:
+            return pyotp.HOTP(self.otp_secret)
+        else:
+            raise ValueError('{} is not using HOTP'.format(self))
```
redshiftzero: Just clarifying the "potential vulnerability" language in the PR description - we're not using these properties insecurely anywhere in our production code, that said, this is a solid addition for defense in depth.

securedrop/models.py
```diff
@@ -317,12 +317,12 @@ def regenerate_totp_shared_secret(self):
         self.otp_secret = pyotp.random_base32()
 
     def set_hotp_secret(self, otp_secret):
-        self.is_totp = False
         self.otp_secret = base64.b32encode(
             binascii.unhexlify(
                 otp_secret.replace(
                     " ",
                     "")))
+        self.is_totp = False
```
redshiftzero: Can you document why you are doing this in the commit message? 

Note that the validation of the secret value itself should be done prior to `set_hotp_secret` being called - i.e. in `manage._add_user` and in the 2FA-related views and/or forms (the exception being only when we do some test setup where we might not run the validation).

PR #3260, opened 2018-04-12
Alembic migration test harness

heartsucker:
## Status

Ready for review

## Description of Changes

Toward #3244

This adds a basic test harness that will be required for all migrations. It uses parameterized pytest functions that loop over all the migrations and run as series of tests against them. Developers will not have to write the boiler plate code for this and will only have to define the checks they want run against the data to ensure consistency. The steps necessary to write a migration test are described in the documentation.

## Testing

`make test TESTFILES=tests/test_alembic.py`


docs/development/database_migrations.rst
```diff
@@ -74,3 +80,80 @@ command to see an output of the SQL that will be generated.
 .. code-block:: none
 
     alembic upgrade head --sql
+
+Testing Migrations
+~~~~~~~~~~~~~~~~~~
+
+The test suite already comes with a test runner (``test_alembic.py``) hat run a series of checks to
```
redshiftzero: Nit: `hat` -> `that`

...🎩

docs/development/database_migrations.rst
```diff
@@ -74,3 +80,80 @@ command to see an output of the SQL that will be generated.
...
+
+.. code-block:: none
+
+    tests/migrations/
+    ├── __init__.py
+    ├── migration_15ac9509fc68.py
+    └── migration_faac8092c123.py
+
+
+Your module **MUST** contain the classes with the following classes with the following attributes.
```
redshiftzero: Nit: `the classes with the following classes with the following attributes` -> `the following classes with the following attributes`

securedrop/tests/migrations/migration_15ac9509fc68.py
```diff
@@ -0,0 +1,211 @@
...
+import random
+import string
+
+from datetime import datetime
+from sqlalchemy import text
+
+from db import db
+from journalist_app import create_app
+
+random.seed('ᕕ( ᐛ )ᕗ')
```
redshiftzero: yay determinism

securedrop/tests/migrations/migration_15ac9509fc68.py
```diff
@@ -0,0 +1,211 @@
...
+def random_username():
+    len = random.randint(3, 64)
+    return random_chars(len)
+
+
+def random_chars(len, chars=string.printable):
+    return ''.join([random.choice(chars) for _ in range(len)])
+
+
+def random_datetime(nullable):
```
redshiftzero: These functions are going to be used a _lot_ in migration testing - at some point we'll want to move these functions to a utility file

securedrop/tests/migrations/migration_15ac9509fc68.py
```diff
@@ -0,0 +1,211 @@
...
+            for jid in range(1, 200, 10):
+                for sid in range(1, 200, 10):
+                    self.add_reply(jid, sid)
+
+            for sid in range(1, 200, 10):
+                self.add_source_star(sid)
+
+            for sid in range(1, 200, 8):
+                for _ in range(random.randint(1, 3)):
+                    self.add_submission(sid)
```
redshiftzero: So in order to test the broken FK submission case in #1189 we could have an `add_orphaned_submissions` static method that adds some submissions, and then deletes their sources. Nice thing: since the cascade delete is enforced at the ORM level and not the database, then the submissions will not be deleted. Also, if we want to defer including this particular case to the Mega Pre-Release Test script I'm cool with that

securedrop/tests/test_alembic.py
```diff
@@ -24,32 +28,123 @@ def list_migrations(cfg_path, head):
...
+
+    reverted_schema = get_schema(app)
+
+    # The initial migration is a degenerate case because it creates the table
+    # 'alembic_version', but rolling back the migration doesn't clear it.
+    if len(migrations) == 1:
+        reverted_schema = list(filter(lambda x: x[2] != 'alembic_version',
+                                      reverted_schema))
+
+    assert reverted_schema == original_schema
```
redshiftzero: Sweet. Since this also ensures the same FK relationships exist before and after the revert, this covers the cases where I wanted us to do upgrade, downgrade, upgrade.

securedrop/tests/migrations/migration_15ac9509fc68.py
heartsucker: Fixed.

securedrop/tests/test_alembic.py
heartsucker: This actually doesn't *fully* cover that case because of non-SQL things like moving files around on the FS or purging old keys from the GPG keyring. We can add that test later.

securedrop/tests/migrations/migration_15ac9509fc68.py
```diff
@@ -0,0 +1,179 @@
...
+            for sid in range(1, self.SOURCE_NUM, 10):
+                self.add_source_star(sid)
+
+            for sid in range(1, self.SOURCE_NUM, 8):
+                for _ in range(random.randint(1, 3)):
+                    self.add_submission(sid)
+
+            # create "abandoned" submissions (issue #1189)
+            for sid in range(self.SOURCE_NUM, self.SOURCE_NUM + 50):
+                self.add_submission(sid)
```
redshiftzero: Great, thanks :+1:

creviera:
Hey @heartsucker - Your PR looks good and the docs update is great :tada: :tada: - since @redshiftzero already gave this a thorough review, I'm mostly interested in investigating the `test_alembic_head_matches_db_models` test failure.

I was able to reproduce by removing the `-k-test_alembic.py` workaround and running `make test`:
```
E               AssertionError: Schema for ('table', 'replies', 'replies') did not match:                                                                             
E               Left:                                                                                                                                                 
E               CREATE TABLE replies (                                                                                                                                
E                       id INTEGER NOT NULL,                                                                                                                          
E                       uuid VARCHAR(36) NOT NULL,                                                                                                                    
E                       source_id INTEGER,                                                                                                                            
E                       journalist_id INTEGER,                                                                                                                        
E                       filename VARCHAR(255) NOT NULL,                                                                                                               
E                       size INTEGER,                                                                                                                                 
E                       is_downloaded BOOLEAN,                                                                                                                        
E                       CONSTRAINT pk_replies PRIMARY KEY (id),                                                                                                       
E                       CONSTRAINT fk_replies_journalist_id_users FOREIGN KEY(journalist_id) REFERENCES users (id),                                                   
E                       CONSTRAINT fk_replies_source_id_sources FOREIGN KEY(source_id) REFERENCES sources (id),                                                       
E                       CONSTRAINT uq_replies_uuid UNIQUE (uuid),                                                                                                     
E                       CONSTRAINT ck_replies_is_downloaded CHECK (is_downloaded IN (0, 1))                                                                           
E               )                                                                                                                                                     
E               Right:                                                                                                                                                
E               CREATE TABLE replies (                                                                                                                                
E                       id INTEGER NOT NULL,                                                                                                                          
E                       uuid VARCHAR(36) NOT NULL,                                                                                                                    
E                       source_id INTEGER,                                                                                                                            
E                       journalist_id INTEGER,                                                                                                                        
E                       filename VARCHAR(255) NOT NULL,                                                                                                               
E                       size INTEGER,                                                                                                                                 
E                       is_downloaded BOOLEAN,                                                                                                                        
E                       CONSTRAINT pk_replies PRIMARY KEY (id),                                                                                                       
E                       CONSTRAINT uq_replies_uuid UNIQUE (uuid), 
E                       CONSTRAINT fk_replies_source_id_sources FOREIGN KEY(source_id) REFERENCES sources (id), 
E                       CONSTRAINT fk_replies_journalist_id_users FOREIGN KEY(journalist_id) REFERENCES users (id), 
E                       CHECK (is_downloaded IN (0, 1))
E               )

tests/test_alembic.py:60: AssertionError                                                                                                                              
```

## Alembic CHECK contraints

I briefly looked into alembic CHECK constraints wondering if it wasn't fully implemented feature. I read that SQLAlchemy doesn’t "reflect CHECK constraints on any backend." And that SQLite CHECK constraints require manual migration. Before going down that rabbit hole, I looked to see if anything else looked :fish: y

## Running tests in random order

`make test` (without `-k-test_alembic.py`) passes when you remove `pytest-random-order=global`. `pytest-random-order=global` is there to randomize the order in which tests are run in order to prevent tests from passing just because they run after other tests. The test that is having an issue is `test_alembic_head_matches_db_models` and the problem goes away when you run this test first, which explains why this PR's workaround works.

To double check the test that is now passing, I manually compared the table schemas and saw that they matched:

```
> sqlite3 /tmp/pytest-of-creviera/pytest-67/test_alembic_head_matches_db_m0/models/svs.sqlite

sqlite> .schema replies
CREATE TABLE replies (
        id INTEGER NOT NULL,
        uuid VARCHAR(36) NOT NULL,
        source_id INTEGER,
        journalist_id INTEGER,
        filename VARCHAR(255) NOT NULL,
        size INTEGER,
        is_downloaded BOOLEAN,
        CONSTRAINT pk_replies PRIMARY KEY (id),
        CONSTRAINT uq_replies_uuid UNIQUE (uuid),
        CONSTRAINT fk_replies_source_id_sources FOREIGN KEY(source_id) REFERENCES sources (id),
        CONSTRAINT fk_replies_journalist_id_users FOREIGN KEY(journalist_id) REFERENCES users (id),
        CONSTRAINT ck_replies_is_downloaded CHECK (is_downloaded IN (0, 1))
);
```
```
> sqlite3 /tmp/pytest-of-creviera/pytest-67/test_alembic_head_matches_db_m0/alembic/svs.sqlite          
sqlite> .schema replies                                                                                                                                               
CREATE TABLE replies (                                                                                                                                                
        id INTEGER NOT NULL,                                                                                                                                          
        uuid VARCHAR(36) NOT NULL,
        source_id INTEGER,
        journalist_id INTEGER,
        filename VARCHAR(255) NOT NULL,
        size INTEGER,
        is_downloaded BOOLEAN,
        CONSTRAINT pk_replies PRIMARY KEY (id),
        CONSTRAINT fk_replies_journalist_id_users FOREIGN KEY(journalist_id) REFERENCES users (id),
        CONSTRAINT fk_replies_source_id_sources FOREIGN KEY(source_id) REFERENCES sources (id),
        CONSTRAINT uq_replies_uuid UNIQUE (uuid),
        CONSTRAINT ck_replies_is_downloaded CHECK (is_downloaded IN (0, 1))
);

```
## Thoughts

Interesting how the CHECK constraint has been moved over after the change above. It makes it seem that this is indeed something alembic is capable of doing and that tests are just messing with the state somehow. Still strange behavior though. Why would only one constraint not be copied over?

It'll take some time to track down which tests are stepping on each other's toes or if there's something wrong with the way pytest is setup. Probably best way to start is by studying the configuration changes made in this PR and look for any shared code like where db files are saved, etc. It would also be helpful to look into how tests are run when they're in separate files.

PR #3273, opened 2018-04-16
QA data loader for alembic testing

heartsucker:
## Status

Ready for review

## Description of Changes

Toward #3244 

Adds a master dataloader that dumps data in to the database / storage dir to allow us to test migrations.

## Testing

Spin up a staging machine and `./qa_loader.py`

### DIFF
--- a/docs/development/database_migrations.rst
+++ b/docs/development/database_migrations.rst
@@ -81,8 +81,8 @@ command to see an output of the SQL that will be generated.
 
     alembic upgrade head --sql
 
-Testing Migrations
-~~~~~~~~~~~~~~~~~~
+Unit Testing Migrations
+~~~~~~~~~~~~~~~~~~~~~~~
 
 The test suite already comes with a test runner (``test_alembic.py``) that runs a series of checks
 to ensure migration's upgrade and downgrade commands are idempotent and don't break the database.
@@ -157,3 +157,24 @@ accurately test your migration, and it will likely break during future code chan
 should use as few dependencies as possible in your test including other ``securedrop`` code as well
 as external packages. This may be a rather annoying requirement, but it will make the tests more
 robust aginst future code changes.
+
+Release Testing Migrations
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In order to ensure that migrations between from the previous to current version of SecureDrop apply
+cleanly in production-like instances, we have a helper script that is designed to load
+semi-randomized data into the database. You will need to modify the script ``qa_loader.py`` to
+include sample data. This sample data should intentionally include edge cases that might behave
+strangely such as data whose nullability is only enforced by the application or missing files.
+
+During QA, the release manager should follow these steps to test the migrations.
+
+1. Checkout the previous SecureDrop release
+2. Build Debian packages locally
+3. Provision staging VMs
+4. ``vagrant ssh app-staging``
+5. ``sudo su``
+6. ``cd /var/www/securedrop && ./qa_loader.py``
+7. Checkout the release candidate
+8. Re-provision the staging VMs
+9. Check that nothing went horribly wrong
--- a/securedrop/qa_loader.py
+++ b/securedrop/qa_loader.py
@@ -162,6 +162,8 @@ def new_abandoned_submission(config, source_id):
     submission = Submission(source, filename)
     submission.source_id = source_id
     db.session.add(submission)
+    db.session.commit()
+    delete_source(source_id)
 
 
 def delete_source(source_id):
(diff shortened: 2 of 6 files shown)


heartsucker:
Note that this doesn't use `factory_boy` as suggested earlier because when i did it that way, it was honestly way more obtuse and hard to understand than just using a few hand-rolled functions.

securedrop/qa_loader.py
```diff
@@ -0,0 +1,223 @@
...
+    journalist.hotp_counter = (random.randint(-1000, 1000)
+                               if random_bool() else None)
+    journalist.created_on = random_datetime(nullable=True)
+    journalist.last_access = random_datetime(nullable=True)
+
+    db.session.add(journalist)
+
+
+def new_source():
+    fid_len = random.randint(0, 32)
```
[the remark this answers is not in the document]
redshiftzero: Note that we also want to avoid very small non-zero `fid_len` as there is a high probability of hitting `IntegrityError` for them also (I can attest for this random seed `fid_len = random.randint(4, 32)` worked no problem)

securedrop/qa_loader.py
```diff
@@ -0,0 +1,223 @@
...
+        # TODO add submissions without sources for #1189
+
+
+def arg_parser():
+    parser = ArgumentParser(
+        path.basename(__file__),
+        description='Loads data into the database for testing upgrades')
+    parser.add_argument('-m', '--multiplier', type=positive_int, default=100,
+                        help=('Factor to multiply the loaded data by '
+                              '(default 100)'))
```
redshiftzero: sweet!

docs/development/database_migrations.rst
```diff
@@ -157,3 +157,12 @@ accurately test your migration, and it will likely break during future code chan
...
 robust aginst future code changes.
+
+Release Testing Migrations
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In order to ensure that migrations between from the previous to current version of SecureDrop apply
+cleanly in production-like instances, we have a helper script that is designed to load
+semi-randomized data into the database. You will need to modify the script ``qa_loader.py`` to
+include sample data. This sample data should intentionally include edge cases that might behave
+strangely such as data whose nullability is only enforced by the application or missing files.
```
redshiftzero: For release managers and release testers, let's add a bit of docs saying:

1. Provision staging or prod VMs.
2. `sudo su`
3. `cd /var/www/securedrop`
4. `./qa_loader.py`

securedrop/qa_loader.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+def fake_file(config, source_fid):
+    source_dir = path.join(config.STORE_DIR, source_fid)
+    if not path.exists(source_dir):
+        os.mkdir(source_dir)
+
+    filename = random_chars(20, nullable=False, chars=string.ascii_lowercase)
+    filename = '1-' + filename + '-msg.gpg'
+    f_len = random.randint(1, 1024*1024*500)
```
redshiftzero: I see why you're doing this, but sadly this produces a bunch of very large files and my test VMs ran out of space very rapidly :innocent:. 

What about reducing the number of files that have very large sizes? e.g. by drawing from a steep exponential distribution `int(math.floor(random.expovariate(10.0) * 1024 * 1024 * 500))`? We'll get mostly smaller files but (with low probability) a few bigger ones thrown in - this should significantly reduce disk space needed and still be a realistic test (note: I haven't tested this yet and it likely needs tweaking)

securedrop/qa_loader.py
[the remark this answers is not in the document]
heartsucker: I included empty FID's because that's allowed by the DB and could (but very, very likely won't) cause problems during future migrations.

heartsucker:
@redshiftzero Can you review this locally? I'm getting merge conflicts with `develop` during CI, and I'd rather just deal with that rebase once on the `alembic` branch.

securedrop/qa_loader.py
```diff
@@ -0,0 +1,226 @@
...
+                    new_reply(config, jid, sid)
+
+        for jid in range(1, JOURNALIST_COUNT * multiplier, 10):
+            new_journalist_login_attempt(jid)
+
+        for sid in range(SOURCE_COUNT * multiplier,
+                         SOURCE_COUNT * multiplier + multiplier):
+            new_abandoned_submission(config, sid)
+
+        # TODO add submissions without sources for #1189
```
redshiftzero: I think we do want this - `new_abandoned_submission` almost handles this situation, it looks like we just need to delete the source(s) that the abandoned submissions are associated with. Using raw SQL to do the deletion should do the trick since the submission cascade deletion is enforced at the ORM level.

heartsucker:
I just noticed in the UI that you can't add emoji/reactions to a review. Only commends. So.

👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍

PR #3315, opened 2018-04-26
remove incorrect references  to gksu nautilus

### LINKED ISSUE #3314: Docs Update Admin Workstation backup docs to remove gksu reference
## Description

Reported in: https://forum.securedrop.club/t/gksu-missing-in-tails-3-6-2/661

## Steps to Reproduce

Follow admin workstation docs in https://forum.securedrop.club/t/gksu-missing-in-tails-3-6-2/661

## Expected Behavior

All is well and your workstation is backed up!

## Actual Behavior

`sudo: gksu: command not found`

## Comments

The suggestion in https://forum.securedrop.club/t/gksu-missing-in-tails-3-6-2/661 seems reasonable: instruct Admins to use nautilus instead

b-meson:
## Status

Ready for review. 

## Description of Changes

Fixes #3314. The documentation was incorrect as there was no reference to setting an administrator passphrase when logging in. It also references a very confusing "tails upgrade" which is not what we are doing. There is references to a command `gksu` which doesn't exist in Tails.  Additionally, I update screenshots and linked out to the official Tails documentation for backing up a persistent volume. Finally, remove all references to "master drives" as this language should be avoided. 


docs/backup_workstations.rst
```diff
@@ -2,46 +2,45 @@ Backup the Workstations
...
+
+.. note:: The steps in this section should be performed for each *Secure Viewing
+ Station*, *Journalist Workstation*, and *Admin Workstation* USB drive in
+ your organization.
 
 Preparing the Backup Device
 ---------------------------
 
-Navigate to **Applications** ▸ **Utilities** ▸ **Disks**.
+First you must boot the *primary Tails USB* drive. Ensure you have an administrator
```
redshiftzero: `have` -> `set`

docs/backup_workstations.rst
```diff
@@ -2,46 +2,45 @@ Backup the Workstations
...
+password set during the login screen. Then navigate to **Applications** ▸ **Utilities** ▸ **Disks**.
```
redshiftzero: `during` -> `at`

docs/backup_workstations.rst
```diff
@@ -91,31 +89,29 @@ mounted and ready to access.
...
-.. warning::
-            Make sure you use keep the `Terminal` window open while you perform
-            the backups. Otherwise, the `Nautilus` window will close.
+.. warning:: Make sure you use keep the `Root Terminal` window open while
+ you perform the backups. Otherwise, the `Nautilus` window will close.
 
-Make sure you create a directory on the backup drive to store the data from the
-drive you are backing up:
+ Make sure you create a directory on the backup drive to store the data from the
+ drive you are backing up:
```
redshiftzero: Remove spaces at the beginning of the lines in this paragraph, it causes it to be interpreted as part of the warning box above

docs/backup_workstations.rst
```diff
@@ -134,14 +130,23 @@ for each device you backup.
 Finally, once you have completed the steps described in this section for each
 USB drive, unmount the Backup partition and store the drive somewhere safely.
 
+Creating Backup Tails Drives
+----------------------------
+
+There are situations when you may need to create a new Tails Drive with your
+Persistent volumes saved. For example, you may find a USB is failing and would
+like to save your data. This is a separate workflow and the official
```
redshiftzero: Hmm, this sounds confusing as the purpose of the above workflow is to backup and save the data on the Tails drives. Perhaps instead of this paragraph, we can say in a note at the top of this document "This workflow will create a single USB drive with the backup data from all Tails drives. If instead you'd like to create a single duplicate backup Tails drive, you should follow the official documentation at... "

b-meson: @redshiftzero I am okay with a `.. note: ` at the top of this. I will explore tomorrow.

KwadroNaut:
gksu is indeed missing from current Debian testing, see as well https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=892768. The introduction of gksu was because of another bug: https://github.com/freedomofpress/securedrop/issues/2069 Can you test if this bug isn't happening anymore? [Opening a root file manager before in the same session]

I'm sitting somewhere Tails-less, here's another idea: l`aunch nautilus like a regular user (amnesia), then type in the location bar (ctrl-l): admin:///etc/` It should ask the administrator password and you should be able to do whatever you need to do with all the files.

redshiftzero:
Ah, thank you for pointing this out @KwadroNaut! I just checked in Tails 3.6.2 and I was _unable_ to reproduce the bug described in #2069, so instructing admins to launch Nautilus directly should work now.

b-meson:
I think keeping it simpler is best. It is best not to require admins to  type `admin://`; this will increase support burden since people are more comfortable with GUIs. Also the [Tails docs](https://tails.boum.org/doc/first_steps/persistence/copy/index.en.html) for backing up the Persistent volume recommend use a Root Terminal. While that is not what we are doing here, I think keeping our document similar to the Tails docs is the correct approach. Thoughts @redshiftzero @KwadroNaut?

redshiftzero:
Since the bug regarding launching Nautilus from the Root Terminal appears to be resolved, I think no additional changes are necessary here @b-meson - thanks for the docs!

KwadroNaut:
@b-meson thanks for testing, I'm more than happy to keep it simple with using the root terminal, thanks for merging as well. The admin:///path/ is for gui-users, to be typed into nautilus' location bar btw, if you'd ever need it.

PR #3327, opened 2018-04-30
Fixes: #3303 Have proper exit code in securedrop-admin

### LINKED ISSUE #3303: Improve return codes in securedrop-admin
## Description

I noticed while making a change in #3300 that the `return 1` in `securedrop-admin update` did not produce an eventual `sys.exit(1)` as I expected - it always returned `sys.exit(0)`. Looking at the logic in `securedrop-admin/__init__.py`:

```
        try:
            args.func(args)
        except KeyboardInterrupt:
            sys.exit(0)
        except Exception as e:
            raise SystemExit(
                'ERROR (run with -v for more): {msg}'.format(msg=e))
        else:
            sys.exit(0)
```

it looks like we `sys.exit(0)` unless there is an exception, i.e. we're not passing the return codes through to `sys.exit()`. Also, we `sys.exit(0)` for a `KeyboardInterrupt`. 

A suggested improvement is to exit non-zero for a `KeyboardInterrupt` and to pass the return codes through to `sys.exit()` (@kushaldas let me know if you have additional thoughts on this). I think this is a small but worthwhile change (I expect most of the diff will be in the unit tests), and I advocate that we resolve it during the extended (2 week) QA period.

kushaldas:
## Status

Ready for review.

## Description of Changes

Fixes #3303 

## Testing

Run `./securedrop-admin` and check for the exit codes in bash.

### DIFF
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -497,9 +497,9 @@ def install_securedrop(args):
                "servers.")
     sdlog.info("The sudo password is only necessary during initial "
                "installation.")
-    subprocess.check_call([os.path.join(args.ansible_path,
-                                        'securedrop-prod.yml'),
-                          '--ask-become-pass'], cwd=args.ansible_path)
+    return subprocess.check_call([os.path.join(args.ansible_path,
+                                 'securedrop-prod.yml'), '--ask-become-pass'],
+                                 cwd=args.ansible_path)
 
 
 def backup_securedrop(args):
@@ -512,7 +512,7 @@ def backup_securedrop(args):
         'ansible-playbook',
         os.path.join(args.ansible_path, 'securedrop-backup.yml'),
     ]
-    subprocess.check_call(ansible_cmd, cwd=args.ansible_path)
+    return subprocess.check_call(ansible_cmd, cwd=args.ansible_path)
 
 
 def restore_securedrop(args):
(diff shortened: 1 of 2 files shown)


admin/securedrop_admin/__init__.py
```diff
@@ -512,7 +512,7 @@ def backup_securedrop(args):
         'ansible-playbook',
         os.path.join(args.ansible_path, 'securedrop-backup.yml'),
     ]
-    subprocess.check_call(ansible_cmd, cwd=args.ansible_path)
+    return subprocess.check_call(ansible_cmd, cwd=args.ansible_path)
```
redshiftzero: - The logic here makes sense if this returns 0, as a `sys.exit(0)` will occur in `main` ✅ 
- If the command returns non-zero, then this will raise a `subprocess.CalledProcessError`, which gets caught in `main` and always produces a `SystemExit` with return code 1 regardless of the return code on `subprocess.CalledProcessError`, so I think we want to refine this further. I understand that might be a bit much for pre-release changes, but let's at least file a followup to address?

admin/securedrop_admin/__init__.py
```diff
@@ -713,17 +715,18 @@ def main(argv):
...
+            return_code = args.func(args)
         except KeyboardInterrupt:
-            sys.exit(0)
+            sys.exit(-1)
         except Exception as e:
             raise SystemExit(
                 'ERROR (run with -v for more): {msg}'.format(msg=e))
         else:
-            sys.exit(0)
+            sys.exit(return_code)
```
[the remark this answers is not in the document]
redshiftzero: Ah then we would get into the except block. Never mind!

admin/securedrop_admin/__init__.py
redshiftzero: Actually I'll retarget the existing issue since we cleaned up enough for 0.7.0

Issue #3237, opened 2018-04-05
Convert .deb build from Ansible to Docker


heartsucker:
# Feature request

## Description

We use `molecule` + `ansible` to build our debian packages. This is a bit annoying because there is little caching that is saved between builds. This means the dev + build + deploy + test cycle is *very* slow (10 minutes), which means changing a single line in a `postinst` script is tedious to test.

We should move our build logic into `Dockerfile`s so we can take advantage of caching. We would still use `molecule` as the wrapper that handles the builds, but the bulk of the logic would be in the `Dockerfile` itself. If the last steps are copying in the app code and `DEBIAN` dir, subsequent builds will be only seconds long.

Tagging @msheiny since he probably has opinions on this. :D

## User Stories

As a developer testing deployment / builds for the Debian packages, I want better caching so I can spend less time waiting for builds to finish.

msheiny:
:+1:  I'm not a fan of the current ansible scripts however I do think ansible can continue to call docker and we should continue to use molecule to wrap it so we can continue to utilize our testinfra tests. Basically I want to keep the testinfra tests in the flow because i dont want to lose test coverage in this migration.

heartsucker:
Ok, yeah so `molecule` -> `ansible` -> `docker` would be how we do the build. Not yanking ansible entirely, but things like "install build dependencies" and so on would happen in the container so we cache them between builds.

conorsch:
:+1: on switching to Dockerfiles for the package build logic. The lack of caching there is maddening. Optimistically we'd get parallelization (#2939) for free, as well. 

>  want to keep the testinfra tests in the flow because i dont want to lose test coverage in this migration.

Agreed, but wiring up the Docker/Molecule integration could be tricky. As a fallback, we could we run the testinfra tests against the deb packages on the host machine, presumably placed there by the build containers via volume mount.

msheiny:
> Agreed, but wiring up the Docker/Molecule integration could be tricky

Yeah -- i mean we do it all the time but I agree the caching logic sucks with ansible + docker. You have to do all these stupid work-arounds to get it to intelligently rebuild.

conorsch:
What if we just expand the Dockerfile template to include the required packages? We'd probably need more than one template, particularly for the app-code deb, but for the others they should be pretty minimal. We already have the `install_files/ansible-base/roles/build-generic-pkg/files/build_generic_package.sh` script that's perfect for containers.

PR #3331, opened 2018-05-01
Speed up debian build process

### LINKED ISSUE #2939: Parallelize deb package builds
# Feature request

## Description

Running `make build-debs` now takes around 20m to complete, given how many Debian packages we're building. Since we switched to containers for the build process, and we're heavily reusing build logic in serial, we're in a good position to leverage the `builder` Molecule scenario to spin up containers concurrently and build Debian packages in parallel.

Specifically, we should be able to consolidate:

* Five (5) invocations of the `build-generic-pkg` role
* Two (2) invocations of the `build-ossec-deb-pkg` role

No changes to the `build-securedrop-app-code-deb-pkg`, which must run separately, due to the unique logic. The minor customizations at the play level in the `build-deb-pkgs.yml` playbook can be ported to Molecule-specific host vars to accommodate parallel execution.

Whether the change is actually a timesaver bears further testing, but I'm confident we can build more quickly than 20m. (If anyone has substantially different build times, please comment; my workstation often runs several virtualized environments, so that's certainly a factor.)

## User Stories
As a developer, I want to test changes quickly and effectively, and not spend a significant portion of my day waiting for build artifacts to validate changes.

msheiny:
## Status

Ready for Review

## Description of Changes

Fixes #2939 . 
Partially addresses #3237  - moves all apt dependencies into docker for caching

This PR parallelizes the debian build logic and moves all the apt packaging logic into docker (and out of ansible). 

## Testing

* Run `time make build-debs` - confirm you receive no errors. The timing here will be a first run when docker has to build the image completely.
* Run `time make build-debs` again. This time the docker image doesnt have to get rebuilt and you should see a dramatic time change.
* To compare with existing switch back to `develop` and run `time make build-debs`

If you want to clean out all these docker images first you can run the following selective nuke command `docker image prune -a -f --filter "label=image_name=trusty-sd-builder-app"`

## Deployment

This code re-mixes the build process a bit and removes some tests. Give careful review and feedback to that when reviewing this PR.

We want to make sure that there are no regression in the build process so I would say that during QA we should be looking out for any weird behavior during install or involving missing files.


msheiny:
okay did some time comparisons specifically looking at CI jobs:

```
Previous build logic - ~10 min
New logic CI (from scratch) - ~8 min
New logic CI (from cache) - 5:32 
```

99% of the time we will be building from cache unless the dockerfile is updated in a particular PR. For some reason CI is way way faster than when I attempt to build locally from scratch (upwards of 20 min). I wonder if that is related to grsec or a docker version ... hmm...

msheiny:
> Is there a particular reason why it needs to be done with this new build strategy?

If you don't pin the versions, we need another way to detect when there are package updates and tell docker to rebuild the underlying image. Otherwise someone could be building with super old packages regardless.

msheiny:
> As a compromise, perhaps CI could check for security updates for installed packages (for example apt-get -s dist-upgrade |grep "^Inst" |grep -i trusty-security)? While imperfect, it would be an improvement.

hah . sorry, I totally ignored your second comment. yeah im fine with appending some kind of CI checker here. to recap convo out of band .. lets discuss... out of band? heh

molecule/builder/Dockerfile
```diff
@@ -6,8 +6,25 @@ ARG NAME
...
 LABEL image_name="$NAME"
 
-# Build dependencies must not be added here, they are listed
-# in install_files/ansible-base/group_vars/all/securedrop
-RUN apt-get -y update && apt-get install -y python rsync sudo bash aptitude && \
-            aptitude -y dist-upgrade && \
-            apt-get clean
+# Non-version specific tooling required
+# Followed by pegged dependencies
+RUN apt-get -y update && apt-get install -y make unzip python python-pip rsync sudo bash aptitude paxctl \
```
heartsucker: Is there a reason you're not pinning everything?

.circleci/config.yml
```diff
@@ -154,10 +159,28 @@ jobs:
 
       - setup_remote_docker
 
+      - run:
+          name: Load image layer cache
+          command: |
+            set +o pipefail
+            docker load -i /caches/deb-builder.tar |true
```
heartsucker: Why `pipefail` to `true`. Won't that just exit the same as if you ran the command directly?

.circleci/config.yml
```diff
@@ -163,7 +163,7 @@ jobs:
           name: Load image layer cache
           command: |
             set +o pipefail
-            docker load -i /caches/deb-builder.tar |true
+            docker load -i /caches/deb-builder.tar.gz |true
```
heartsucker: Same here with `pipefail` to `true`.

.circleci/config.yml
```diff
@@ -154,10 +159,28 @@ jobs:
 
       - setup_remote_docker
 
+      - run:
+          name: Load image layer cache
+          command: |
+            set +o pipefail
```
heartsucker: Aaaaaand here too.

.circleci/config.yml
msheiny: hah honestly i just copied and pasted the logic that was there before that wasnt written by me. Im not opposed to tearing it out as I dont have any justification for it being there.

molecule/builder/Dockerfile
msheiny: i didnt think it made sense to peg some of these auxiliary tools .... buttt as with `pipefail` i dont have a good argument on why we shouldn't pin here as well.

conorsch:
Clear improvement performance-wise: 

```
First build on speedy-debs: 16m46s
Second build on speedy-debs: 8m56s
Third build on speedy-debs: 8m44s

First build on develop: 12m33s
Second build on develop: 14m54s
```

The pinning strategy definitely needs to be refined. The suggestions by @emkll for CI checks are great. Would be willing to merge this work without pinning, but as we already have solid discussion on that end, let's try to revise accordingly. 

Stepping away, since we need a rebase, and there are some outstanding proposals to implement here.

msheiny:
@conorsch / @emkll  / @heartsucker  - i did some changes on the container handling to hopefully address security concerns brought up during this review. Particularly by building and pushing to our quay.io account for snapshot-ing instead of relying on version control pegs (which are annoying to maintain). I also added a test to see if there are security updates to that base container... im not 100% that we should have that check run on every PR... but I think its the quickest compromise I could get in for now.

molecule/builder/Dockerfile
```diff
@@ -1,13 +1,33 @@
-ARG IMAGE
-FROM $IMAGE
+# April 28th, 2018
```
heartsucker: Nit: Can you include the version you pulled to get this sha? E.g., `trusty-2018-04-28`. I find this makes it easier to track things down when I have to revisit dockerfiles.

molecule/builder/Dockerfile
```diff
@@ -1,13 +1,33 @@
...
+        python-pip \
+        rsync \
+        ruby \
+        secure-delete \
+        sqlite \
+        unzip \
+        && apt-get clean \
+        && rm -rf /var/lib/apt/lists/*
+
+RUN paxctl -cm /usr/bin/python2.7
```
heartsucker: Nit: You could make these all one `RUN` command to reduce the number of layers that are saved to the dev's disk.

molecule/builder/create.yml
```diff
@@ -8,38 +8,14 @@
     molecule_ephemeral_directory: "{{ lookup('env', 'MOLECULE_EPHEMERAL_DIRECTORY') }}"
     molecule_scenario_directory: "{{ lookup('env', 'MOLECULE_SCENARIO_DIRECTORY') }}"
     molecule_yml: "{{ lookup('file', molecule_file) | from_yaml }}"
+    image_hash: e9f1683b1fe756751e3ec6481860414988eceaa25663e2e98dd5223e6c3b2856
```
heartsucker: Where does this hash come from? Can you make a note so it looks less like a magic number.

heartsucker:
Tested with

```
virtualenv -p `which python2` venv
source venv/bin/activate
pip install -r securedrop/requirements/develop-requirements.txt
make build-deb
vagrant up /provision/
```

and then doing one full submission with the source via Tor Browser.

Issue #3367, opened 2018-05-04
[0.7.0] Journalist notification sends email every time the app server is manually restarted


emkll:
# Bug

## Description

Every time a SecureDrop instance running 0.7.0~rc2 is rebooted, a journalist notification indicating that there's a new submission is sent, despite there not being any new submissions. Note that this particular instance was upgraded from 0.6.

## Steps to Reproduce

* Install 0.7.0~rc2 and enable journalist notifications
* Press the power button on the instance to initiate shutdown of the app server
* Press the power button again to start the app server
* Observe an email in the journalist inbox stating that `There has been submission activity in the past 24 hours.`

## Expected Behavior

The email should only be sent at a regular interval to minimize information leakage and the information sent to journalists should be accurate.

## Actual Behavior

The email is sent at every instance reboot and the information sent to the journalist is not accurate.

redshiftzero:
Really good catch @emkll. Defect #3368 is my reproducing this issue along with another issue: alerts continuing to be grouped. Investigating...

redshiftzero:
Ah, indeed, I've been rebooting a bunch of times, and sometimes I get the alert separately, reproducing this bug. Sometimes I get the alerts grouped, which produces #3368. The intermittent nature of this is why the resolution of alert grouping described [here](https://github.com/freedomofpress/securedrop/pull/2803#discussion_r170454765) appeared to be sufficient during review.

Issue #3368, opened 2018-05-04
[0.7.0] Journalist and admin notifications can group and cause "SecureDrop Submissions Error"


redshiftzero:
## Description

While trying to reproduce #3367, I discovered a related issue, where a "SecureDrop Submissions Error" is sent to the administrator.

## Steps to Reproduce

1. Enable journalist notifications.
2. Reboot the app server (I just powered off the VM)

## Expected Behavior

No journalist notification is sent

## Actual Behavior

A notification entitled "SecureDrop Submissions Error" is sent to the administrator, clipped:

```
OSSEC HIDS Notification.
2018 May 04 06:35:13

Received From: (app) 10.0.1.4->netstat -tan |grep LISTEN |grep -v 127.0.0.1 | sort
Rule: 533 fired (level 7) -> "Listened ports status (netstat) changed (new port opened or closed)."
Portion of the log(s):

ossec: output: 'netstat -tan |grep LISTEN |grep -v 127.0.0.1 | sort':
tcp        0      0 0.0.0.0:111             0.0.0.0:*               LISTEN     
tcp        0      0 0.0.0.0:54579           0.0.0.0:*               LISTEN     
tcp6       0      0 :::111                  :::*                    LISTEN     
tcp6       0      0 :::34144                :::*                    LISTEN     
Previous output:
ossec: output: 'netstat -tan |grep LISTEN |grep -v 127.0.0.1 | sort':
tcp        0      0 0.0.0.0:111             0.0.0.0:*               LISTEN     
tcp        0      0 0.0.0.0:35431           0.0.0.0:*               LISTEN     
tcp6       0      0 :::111                  :::*                    LISTEN     
tcp6       0      0 :::34203                :::*                    LISTEN     



 --END OF NOTIFICATION



OSSEC HIDS Notification.
2018 May 04 06:35:13

Received From: (app) 10.0.1.4->head -1 /var/lib/securedrop/submissions_today.txt | grep '^[0-9]*$'
Rule: 400600 fired (level 1) -> "Boolean value indicating if there were submissions in the past 24h."
Portion of the log(s):

ossec: output: 'head -1 /var/lib/securedrop/submissions_today.txt | grep '^[0-9]*$'':
1
```

## Comments

I found a similar issue during [review](https://github.com/freedomofpress/securedrop/pull/2803#discussion_r170430840), that should be the first thing to investigate

redshiftzero:
[...]

Issue #3316, opened 2018-04-26
Debian package securedrop-app-code is not getting upgraded


kushaldas:
# Bug

## Description

In my prod vm `securedrop-app-code` package is not getting upgraded.

## Steps to Reproduce

Follow the QA steps to fetch the latest packages from apt-test repo.

## Expected Behavior

We should get the 0.7~RC1 packages.

## Actual Behavior

```
$ dpkg -al | grep secure
ii  openssh-client                       1:6.6p1-2ubuntu2.10                        amd64        secure shell (SSH) client, for secure access to remote machines
ii  openssh-server                       1:6.6p1-2ubuntu2.10                        amd64        secure shell (SSH) server, for secure access from remote machines
ii  openssh-sftp-server                  1:6.6p1-2ubuntu2.10                        amd64        secure shell (SSH) sftp server module, for SFTP access from remote machines
ii  secure-delete                        3.1-6                                      amd64        tools to wipe files, free disk space, swap and memory
ii  securedrop-app-code                  0.6                                        amd64        Packages the SecureDrop application code pip dependencies and apparmor profiles. This package will put the apparmor profiles in enforce mode. This package does use pip to install the pip wheelhouse
ii  securedrop-config                    0.1.1+0.7.0~rc1                            all          Establishes baseline system state for running SecureDrop.
ii  securedrop-grsec                     4.4.115+r1                                 amd64        Metapackage providing a grsecurity-patched Linux kernel for use
ii  securedrop-keyring                   0.1.1+0.7.0~rc1                            amd64        Provides an apt keyring for SecureDrop-related packages, so the master signing key used for SecureDrop packages can be updated via apt.
ii  securedrop-ossec-agent               2.8.2+0.7.0~rc1                            amd64        Installs the securedrop pre-configured OSSEC agent
ii  ssh-import-id                        3.21-0ubuntu1                              all          securely retrieve an SSH public key and install it locally
```

```
$ sudo apt-get upgrade
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Calculating upgrade... Done
The following packages have been kept back:
  securedrop-app-code
0 upgraded, 0 newly installed, 0 to remove and 1 not upgraded.

```

kushaldas:
It seems if I manually do a `sudo apt-get update` and then `sudo apt upgrade`, it pulls in the right dependencies and the upgraded package.

kushaldas:
In my latest testing I could not reproduce this error. I am closing this issue.

[...]

PR #3374, opened 2018-05-05
ossec: resolve journalist notification racing with reboots

dachary:
## Status

Ready for review

## Description of Changes

Mitigates: #3368 

The app server is rebooted every 24h and will send a notification at
boot time. The ossec server is also rebooted and will immediately send
the email to the journalist, regardless of when the previous mail was
sent (mail frequency is not a feature of ossec-maild). Always running
the localfile command at boot time is an undocumented OSSEC behavior
https://github.com/ossec/ossec-hids/issues/1415 in 2.8.2 as well as
2.9.3.

This guarantees exactly one mail will be sent daily.

Setting the 25 hours frequency element is a safeguard:

* against the following race a) command runs because the 24h period
  expires, b) the server reboots shortly after because it reboots
  every 24h, c) command runs again after the server is rebooted,
  causing two notifications to be sent in a row

* in case the server does not reboot for some reason, the notification
  will still be sent every 25h

Fixes: https://github.com/freedomofpress/securedrop/issues/3367

## Testing

* Reboot the app server and verify the notification is sent

## Deployment

N/A

### DIFF
--- a/install_files/ansible-base/roles/ossec/files/process_submissions_today.sh
+++ b/install_files/ansible-base/roles/ossec/files/process_submissions_today.sh
@@ -1,10 +1,28 @@
 #!/bin/bash
 
+SENT_STAMP=/var/ossec/logs/journalist_notification_sent.stamp
+
 function send_encrypted_alarm() {
     /var/ossec/send_encrypted_alarm.sh "$1"
 }
 
 function main() {
+    if modified_in_the_past_24h "${SENT_STAMP}" ; then
+        logger "$0 journalist notification suppressed"
+    else
+        handle_notification "$@"
+    fi
+}
+
+function modified_in_the_past_24h() {
+    local stamp
+    stamp="$1"
+    test -f "${stamp}" && \
+        find "${stamp}" -mtime -1 | \
+            grep --quiet "${stamp}"
+}
+
+function handle_notification() {
     local sender
     local stdin
     sender=${1:-send_encrypted_alarm}
@@ -24,6 +42,7 @@ function main() {
             echo "There has been no submission activity in the past 24 hours."
             echo "You do not need to login to SecureDrop."
         fi | $sender journalist
+        touch "${SENT_STAMP}"
     else
         export SUBJECT="SecureDrop Submissions Error"
         (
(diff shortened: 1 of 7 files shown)


dachary:
@emkll the second commit of this PR **ossec: never send more than one journalist notification per 24h** it good to have but I'm not sure if we want it for 0.7.0. My concern is about adding code that does not fix a bug, that close to the release. In practice journalists who could be inconvenienced would be the one associated with a sysadmin that not only configures journalists notifications but also manually reboots for some reason.

install_files/securedrop-ossec-agent/var/ossec/etc/ossec.conf
```diff
@@ -108,7 +108,7 @@
   <localfile>
     <log_format>full_command</log_format>
     <command>head -1 /var/lib/securedrop/submissions_today.txt | grep '^[0-9]*$'</command>
-    <frequency>86400</frequency>
+    <frequency>90000</frequency> <!-- 25 hours -->
```
redshiftzero: Question re: #3368 - does change mean that this `localfile` won't be considered a daily notification and thus `do_not_group` will not apply?

redshiftzero: Ah okay nevermind, I see that the corresponding rule in `local_rules.xml` is using `rule_id=400600` ✅

dachary:
@redshiftzero removed the extra commit (and test does not need to run because it already did on the first commit ;-)

redshiftzero:
Okay after looking at this more and taking a step back, the choices are:

First commit only: Minimal change that indeed resolves #3367. This is a clever fix. The thing is it also produces behavior worthy of another bug report "Additional reboots send submission notifications to journalists" (I know that a note in the documentation is added in this PR, but some admins will likely miss the note and it is odd and unexpected behavior unless one is familiar with the implementation). The other issue is that these additional notifications might actually be incorrect - indicating that submissions have or have not occurred in the last 24 hours when the opposite is true (since the `submissions_today.txt` file has not been updated on the application server). 

Both commits: Much larger diff in functionality and thus high risk given the time we have to test before release. It does resolve both issues however. 

Am I understanding this correctly? If I am, I think we actually have to merge both commits to resolve all issues prior to release....

dachary:
@redshiftzero I'd rather stick to the simpler fix. 

* The risk of introducing a regression with the larger fix are too high this close to the release, despite testing. The functionality being optional, it won't impact all SecurDrop instance. And in the subset of instances that activate it, it will only affect the even smaller subset.
* If we go for the larger patch, we take the risk of announcing a functionality that is broken for everyone trying it because we missed something.

That being said, I'm deferring to your better judgement :-)

redshiftzero:
> The risk of introducing a regression with the larger fix are too high this close to the release, despite testing.

Totally agree. 

Well, the silver lining to pushing back the release by a week due to other significant issues (#3316), is that we now have time to carefully QA both commits, and this will fix all outstanding issues (if something doesn't work as we expect, we should have enough time to resolve it). Thanks for your work on this @dachary.

dachary:
The fix enforcing a 24h delay was re-pushed

emkll:
Thanks @dachary, seems like the first commit it working as intended, I receive emails upon every reboot, as well as at 4AM local time when my instance is scheduled to reboot.

I've been testing the 2nd commit of this PR (856b180e3014f0e739ca8ea4f469533c1bde4751), but I still receive more than 1 notification per 24h
screenshot from 2018-05-08 10-18-20

I suspect this is due to permissions on the `/var/ossec/tmp/` folder, owned by root, for which the `ossec` group has no write permissions, and as such can't write the stamp file:
```
root@mon:/var/ossec/tmp# ls -lah
total 8.0K
dr-xr-x---   2 root ossec 4.0K May  8  10:16 .
dr-xr-x--- 14 root ossec 4.0K May  8  09:49 ..
```

dachary:
Added the **make journalist notifications resilient to double ossec alerts** commit to mitigate https://github.com/freedomofpress/securedrop/issues/3368 since we're unable to reproduce it.

dachary:
@emkll good catch! I moved it to the logs directory because it's already writable by the ossec user.

emkll:
Thanks @dachary for the quick fixes. I tested the latest changes and can confirm this addresses the issues we've been seeing in the 0.7.0 RCs.

One behavior that I've noticed is as follows: 
Admin adds journalist notifications, instance reboots before daily reboot time, the notification triggers. Following notification occurs only at the 2nd daily reboot time. It seems that addressing this edge case would introduce a lot of complexity and I perhaps we should leave it as-is.

I'd appreciate if someone else could do a quick sanity-check/review of this PR, but consider it approved from my perspective.

install_files/ansible-base/roles/ossec/files/process_submissions_today.sh
```diff
@@ -34,33 +53,50 @@ function main() {
...
     cat
 }
 
-function test_main() {
+function test_handle_notification() {
     shopt -s -o xtrace
     PS4='${BASH_SOURCE[0]}:$LINENO: ${FUNCNAME[0]}:  '
 
-    echo BUGOUS | main test_send_encrypted_alarm | \
+    echo BUGOUS | handle_notification test_send_encrypted_alarm | \
```
conorsch: Nit: should this be "BOGUS", as in a fake submission? (I realize this came in via #2803, just noting for clarity's sake.)

testinfra/ossec/test_journalist_mail.py
```diff
@@ -203,6 +214,21 @@ def test_journalist_mail_notification(self, host):
...
+        #
+        # The second notification within less than 24h is suppressed
+        #
+        with app.sudo():
+            self.service_restarted(app, "ossec")
+        assert self.wait_for_command(mon, """
+        grep --count 'notification suppressed' /var/log/syslog > /tmp/after
+        test $(cat /tmp/before) -lt $(cat /tmp/after)
+        """)
+
```
conorsch: Great tests! Thanks for the careful attention here!

emkll:
Thanks for the quick review @conorsch , merging!

dachary:
@conorsch @emkll thanks for the careful review :-)

> Admin adds journalist notifications, instance reboots before daily reboot time, the notification triggers. Following notification occurs only at the 2nd daily reboot time. It seems that addressing this edge case would introduce a lot of complexity and I perhaps we should leave it as-is.

I second your opinion @emkll

PR #3378, opened 2018-05-08
Make unread status on submissions and sources clearer

### LINKED ISSUE #3159: UX: Distinction between read and unread content is too subtle
# Feature Request

## Description

The distinction between read and unread content in the journalist interface is too subtle right now. We've had at least one report of an organization missing this feature entirely. See the screenshot below:

unreadux

We should discuss a few different UX approaches here, but will likely want to take a hint from other messaging systems such as common email UIs, which tend to use bold text as an additional visual indication of read/unread status. 

If we cannot target a fix for the next release, we should add some documentation as a stopgap measure.

## User Stories

As a journalist, I want to be able to quickly understand which collections or messages have been read/unread.

harrislapiroff:
## Status

Ready for review

## Description of Changes

Fixes #3159.

* Alter the appearance of the submission list to make a clearer distinction between read and unread documents

## Testing

In the journalist interface, from the source list view, observe that sources with unread documents look different, like so:

screen shot 2018-05-08 at 1 58 08 pm

Visit a source and observe that read/unread files look different, like so:

screen shot 2018-05-08 at 10 56 20 am

To change a file from read to unread state, download it. You will need to refresh the page.

## Deployment

No deployment considerations

### DIFF
--- a/securedrop/journalist_templates/_source_row.html
+++ b/securedrop/journalist_templates/_source_row.html
@@ -1,18 +1,22 @@
 {% set docs = source.documents_messages_count()['documents'] %}
 {% set msgs = source.documents_messages_count()['messages'] %}
-<li class="source" data-source-designation="{{ source.journalist_designation|lower }}">
+<li class="source {% if source.num_unread != 0 %}unread{% else %}read{% endif %}" data-source-designation="{{ source.journalist_designation|lower }}">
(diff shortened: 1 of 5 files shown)


harrislapiroff:
Okay, this is ready to be reviewed. 🎉

PR #3384, opened 2018-05-09
[0.7.0] ossec: resolve journalist notification racing with reboots

dachary:
## Status

Ready for review

## Description of Changes

Backport #3374

Mitigates: #3368 

The app server is rebooted every 24h and will send a notification at
boot time. The ossec server is also rebooted and will immediately send
the email to the journalist, regardless of when the previous mail was
sent (mail frequency is not a feature of ossec-maild). Always running
the localfile command at boot time is an undocumented OSSEC behavior
https://github.com/ossec/ossec-hids/issues/1415 in 2.8.2 as well as
2.9.3.

This guarantees exactly one mail will be sent daily.

Setting the 25 hours frequency element is a safeguard:

* against the following race a) command runs because the 24h period
  expires, b) the server reboots shortly after because it reboots
  every 24h, c) command runs again after the server is rebooted,
  causing two notifications to be sent in a row

* in case the server does not reboot for some reason, the notification
  will still be sent every 25h

Fixes: https://github.com/freedomofpress/securedrop/issues/3367

## Testing

* Reboot the app server and verify the notification is sent

## Deployment

N/A

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -75,6 +75,9 @@ worth checking the *Journalist Interface*. For this you will need:
           the GPG private key, it is not possible to specify multiple
           GPG keys.
 
+.. note:: The journalist notification is sent after the daily reboot
+          of the *Application Server*.
+
 You will have to copy the following required files to
 ``install_files/ansible-base``:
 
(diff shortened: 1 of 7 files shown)


dachary:
I would merge it if I had permissions but it appears I don't have permission to merge in release/0.7

PR #3386, opened 2018-05-09
Wip kingu source string suggestions

dachary:
## Status

Ready for review

## Description of Changes

Various source string suggestions from @comradekingu 

## Testing

N/A

## Deployment

N/A

### DIFF
--- a/securedrop/journalist_templates/col.html
+++ b/securedrop/journalist_templates/col.html
@@ -13,7 +13,7 @@
       <a href="#change-codename-{{ source.journalist_designation }}" title="{{ gettext('Generate a new random codename for this source. We recommend doing this if the first random codename is difficult to say or remember. You can generate new random codenames as many times as you like.') }}" class="plain" id="regenerate-codename-btn"><span id="regenerate-codename-btn-label">{{ gettext('Change codename') }}</span></a>
       <div id="change-codename-{{ source.journalist_designation }}" class="confirm-prompt">{{ gettext('Are you sure you want to generate a new codename?') }}
         <a href="#regenerate-code">{{ gettext('Cancel') }}</a>
-        <button class="sd-button" type="submit">{{ gettext('CONFIRM') }}</button>
+        <button class="sd-button" type="submit">{{ gettext('Confirm') }}</button>
       </div>
     </p>
   </form>
@@ -68,7 +68,7 @@
       {% with %}
         {% set modal_data = {
                               "modal_id": "delete-selected-confirmation-modal",
-                              "modal_header": gettext('Delete Confirmation'),
+                              "modal_header": gettext('Confirmation of deletion'),
                               "modal_body": gettext('Are you sure you want to delete the selected documents?'),
                               "cancel_id": "cancel-selected-deletions",
                               "submit_id": "delete-selected",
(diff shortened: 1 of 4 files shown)


ei8fdb:
[That OSSEC alerts work" "Update logo](https://github.com/freedomofpress/securedrop/pull/3386/commits/ed5037e4889cb9e3dbb6b1e129bd0955aff00413)

-- looks good. The word logo is an abbreviation of "logotype" so no need to uppercase it.

----

[Confirmation of deletion, Confirm, Afterwards, next time](https://github.com/freedomofpress/securedrop/pull/3386/commits/9abc924083395747d31ba37aa1c6403c4650fc15)

Line 16 -- [It's good practice to label buttons with what it does](https://www.uxmatters.com/mt/archives/2012/05/7-basic-best-practices-for-buttons.php) - what it does for *the user*, not necessairly in terms of it's internal system action. 

Confirm doesn't really mean anything, when read in isolation. I'd suggest relabeling the button to "Yes, generate new codename". (This approach will be used again below)

Since the action of "generating a new codename" is probably an uncommon task for a user to do, once they have clicked on "Change codename", good practise is help them complete their goal quickly - have the next main action button (i.e. the new codename generation) closer to the text/question.

LIne 71 -- Providing clearer button labelling on button id="delete-selected" removes the need for the <h2> (and so localisation) in
<div id="delete-selected-confirmation-modal" class="modal-dialog">.

I've swapped the ordering of the actions, putting the [Delete documents] (again in this case, the main action) on the left. I've no evidence showing the need for this, so if you prefer, you can choose to ignore  this suggestion if you like.

Also, since there is a secondary action button (Cancel) in the modal, there is no real need for the "<a href="#close" title="Close" class="close">X</a>" as the cancel button achieves the same outcome as clicking the cancel button.

securedrop

Line 104 - looks good.

securedrop

Line 115 -- The original (and therefore the suggested) line of content read strangley for me. 

I understand it as meaning (essentially):

You can delete this collection of files. If you do delete this collection of files, a) they will disappear from SecureDrop, and b) the source that submitted them will not be able to login to SecureDrop again.

My assumption is that most users would expect behaviour **A** above ("files will disappear from SecureDrop"). Therefore there shouldn't be a need for this explanation. My suggestion would be this line can be removed. 

However behaviour **B** ("the source that submitted the files will not be able to login") probably is not expected behaviour. 

Therefore it is a very good idea to have this explanation. The question is: where should this "warning" be put. 

In this case, I'd recommend one of the following:

1. 

securedrop

**OR**

2. 

securedrop

followed by this:

securedrop

-----

[SecureDrop will...](https://github.com/freedomofpress/securedrop/pull/3386/commits/513a36adfe5839c9a0b1d1fb09dd566873d2c8a4)

Line 8 -- makes sense 👍 
Line 10 -- makes sense 👍 

[Confirmation of deletion, Delete collection, Flag this…](https://github.com/freedomofpress/securedrop/pull/3386/commits/ca7b4e341447a4e5a18ad30b8a305b4e16f1eb91)

Line 110, 122 -- lowercase makes sense. 👍 

Line 129 -- Again, as above, labelling the buttons with their actions will make this clearer and also remove the need for extra UI elements and localisation.

securedrop

----

[Confirmation of deletion ](https://github.com/freedomofpress/securedrop/pull/3386/commits/e6f19f7d433d9f108893d82572448fbc67ca4298)

Line 46 -- again, labelling buttons with their actions will make this clearer and also remove the need for extra UI elements and localisation.

securedrop

ei8fdb:
@heartsucker @comradekingu if the above look OK, and you'd prefer I can make the content changes if you'd prefer. 

I might need some help in deleting strings, reodering button locations..

comradekingu:
@ei8fdb 
""Cancel" sounds like it gets additionally gets rid of the files i though. Just "Yes" and "X" or "No" would suffice"
"Deleting these files will prevent the submitter from logging in" (and put it below the button?"

dachary:
@ei8fdb  thank you for taking the lead on this: I'll close this pull request so you can open another one with the changes you suggest.

PR #3406, opened 2018-05-14
improve logic in securedrop-admin update

### LINKED ISSUE #3405: Securedrop-admin update should verify tag signature prior to checking out the tag
# Bug
## Description
SecureDrop admin update logic introduced in 0.6 checks out a tag before verifying that it was properly signed with the SecureDrop release signing key. Should an incorrect (non-production-ready) or malicious tag be selected, the error will appear to an administrator once. Should the admin ignore this error, the SecureDrop repository may no longer update to production-ready code.

As a SecureDrop administrator, I would like to check out and run only code that has been certified by the SecureDrop team as production ready.

## Expected Behavior
updater should only checkout code if the signature is valid. The order should be as follows:
1- fetch
2- tag -v
3- checkout

## Actual Behavior

The order is:
1- fetch
2- checkout
3- tag -v

## Comments

All upgrades to 0.7 should use the manual method of validating tags in the order described as above in the blog post, and can safely use the GUI updater in subsequent SecureDrop releases.

redshiftzero:
## Status

ready for review

## Description of Changes

Fixes #3405 

* Verify tag sig prior to checking out tag

## Testing

Carefully inspect the logic and test `securedrop-admin update` manually in Tails

## Deployment

We are asking users to manually do the update in 0.7.0 to make sure they verify the singing key fingerprint manually prior to setting up the automated workstation updates going forward

### DIFF
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -654,8 +654,10 @@ def update(args):
         good_sig_text = 'Good signature from "SecureDrop Release Signing Key"'
         bad_sig_text = 'BAD signature'
         # To ensure that an adversary cannot name a malicious key good_sig_text
-        # we check that bad_sig_text does not appear.
-        if sig_result.count(RELEASE_KEY) == 1 and \
+        # we check that bad_sig_text does not appear and that the release key
+        # appears on the second line of the output.
+        gpg_lines = sig_result.split('\n')
+        if RELEASE_KEY in gpg_lines[1] and \
                 sig_result.count(good_sig_text) == 1 and \
                 bad_sig_text not in sig_result:
             sdlog.info("Signature verification successful.")
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -37,6 +37,7 @@ from prompt_toolkit.validation import Validator, ValidationError
 import yaml
 
 sdlog = logging.getLogger(__name__)
+RELEASE_KEY = '22245C81E3BAEB4138B36061310F561200F4AD77'
 
 
 class FingerprintException(Exception):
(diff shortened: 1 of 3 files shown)


admin/securedrop_admin/__init__.py
```diff
@@ -648,14 +646,33 @@ def update(args):
...
+        good_sig_text = 'Good signature from "SecureDrop Release Signing Key"'
+        bad_sig_text = 'BAD signature'
+        # To ensure that an adversary cannot name a malicious key good_sig_text
+        # we check that bad_sig_text does not appear.
+        if sig_result.count(RELEASE_KEY) == 1 and \
+                sig_result.count(good_sig_text) == 1 and \
+                bad_sig_text not in sig_result:
+            sdlog.info("Signature verification successful.")
+        else:  # If anything else happens, fail and exit 1
+            sdlog.info("Signature verification failed.")
```
dachary: This is very well done :100:

admin/securedrop_admin/__init__.py
```diff
@@ -648,14 +646,33 @@ def update(args):
...
+                                             stderr=subprocess.STDOUT,
+                                             cwd=args.root)
+
+        good_sig_text = 'Good signature from "SecureDrop Release Signing Key"'
+        bad_sig_text = 'BAD signature'
+        # To ensure that an adversary cannot name a malicious key good_sig_text
+        # we check that bad_sig_text does not appear.
+        if sig_result.count(RELEASE_KEY) == 1 and \
+                sig_result.count(good_sig_text) == 1 and \
+                bad_sig_text not in sig_result:
```
dachary: Nit: since this is a very delicate logic, it would be nice to have test cases for all branches of the expression (i.e. one case where the name of the key is **Good signature from "SecureDrop Release Signing Key"**, another where the name of the key contains **22245C81E3BAEB4138B36061310F561200F4AD77**). Absolutely not a blocker for the merge though.

redshiftzero:
I modified the logic slightly to also verify the release key fingerprint appears where expected to mitigate the scenario in cb53f676ea28ab24be603f08810ef10380cf4388, I think that this is the best thing we can do with the current behavior of `git tag -v`

redshiftzero:
thanks for the reviews @emkll and @dachary!

PR #3408, opened 2018-05-14
docs: make style guide fixes throughout

pierwill:
## Status

Ready for review

## Description of Changes

Mostly adding italics to the term "Journalist Interface"

## Checklist

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -25,10 +25,10 @@ on network speed and computing power.
 
 .. _configure_securedrop:
 
-Localization of the source and journalist interfaces
-----------------------------------------------------
+Localization of the *Source Interface* and *Journalist Interface*
+-----------------------------------------------------------------
 
-The source and journalist interface are translated in the following
+The *Source Interface* and *Journalist Interface* are translated in the following
 languages:
 
 .. include:: includes/l10n.txt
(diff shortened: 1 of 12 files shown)


docs/install.rst
```diff
@@ -25,10 +25,10 @@ on network speed and computing power.
 
 .. _configure_securedrop:
 
-Localization of the source and journalist interfaces
-----------------------------------------------------
+Localization of the User Interfaces
```
dachary: We do not have **User Interfaces** anywhere else. Would it be a problem to use 

    *Source Interface* and *Journalist Interface* 

instead?

docs/threat_model/threat_model.rst
```diff
@@ -337,14 +337,14 @@ What a compromise of the journalist's property can achieve
...
-   -  Access the Hidden Service values used by the Journalist Interface.
+   -  Access the Hidden Service values used by the *Journalist Interface*.
    -  Access SSH keys and passphrases for the *Application Server* and the
       *Monitor Server*.
    -  Access the journalist's personal GPG key.
 
 -  An attacker with journalist access to the *Journalist Interface* can:
 
-   -  Change the codenames associated with sources within the Interface.
+   -  Change the codenames associated with sources within the interface.
```
dachary: I don't think we should use a capitalize form here.

docs/install.rst
pierwill: :heavy_check_mark:

docs/threat_model/threat_model.rst
pierwill: I agree. The diff looks like I did make it lowercase, but I'll check again.

dachary: Oh, you are right! I read it the other way around, my bad.

Issue #3395, opened 2018-05-11
Post-install SSH instructions should appear before asked to SSH into app, not after


huertanix:
# Feature request

## Description

At this point in the docs: https://docs.securedrop.org/en/release-0.6/create_admin_account.html, admins are asked to SSH into `app` however are not told that the way to ssh into `app` is now totally different. Instructions for logging into SD servers are in the documentation, but appear a few steps later: https://docs.securedrop.org/en/release-0.6/test_the_installation.html. These instructions should be moved to before admins are asked to ssh into `app`.

## User Research Evidence

When installing SecureDrop, I attempted to ssh in via IP address and I wondered why that wasn't working.

## User Stories

As a SecureDrop admin, I'd like to know when the way a thing is done has changed to something different.

kstohr:
Spoke with @kushaldas As in many cases SecureDrop is handling setting up the first admin acct, we'll add a cross reference here for those admin's installing on their own.

huertanix:
@kstohr Thanks; I'm not sure what you mean by SecureDrop handling setting up the first admin account?

eloquence:
This was resolved via #3414 a while ago.

PR #3414, opened 2018-05-15
Add ssh instructions link to create_admin_account.rst

kstohr:
## Status

Ready for review

## Description of Changes
https://github.com/freedomofpress/securedrop/issues/3395

## Testing
make docs-lint - passed

### DIFF
--- a/docs/create_admin_account.rst
+++ b/docs/create_admin_account.rst
@@ -29,7 +29,7 @@ running a command on the *Application Server*. After that, the Journalist
 Interface admin can create additional accounts through the web
 interface.
 
-To create the first admin account, SSH to the *Application Server*, then:
+If you are installing SecureDrop yourself, to create the first admin account, :doc:`SSH to the Application Server <test_the_installation>`, then:
 
 .. code:: sh
 
@@ -65,3 +65,4 @@ Interface Guide <admin>`.
            following the :ref:`Updating Logo Image` documentation.
 
 .. |TorBrowser| image:: images/torbrowser.png
+


dachary:
Could you please remove the **updated setup_development doc** commit from this PR? It sneaked in from another PR ;-)

dachary:
> If you are installing SecureDrop yourself

Could you please clarify why this part is necessary? In which case would the reader be someone who does not install SecureDrop themselves?

kstohr:
@dachary I am afraid I cannot clarify. I discussed this ticket with @kushaldas prior to updating the text and that wording came from @kushaldas. According to him, this part of the setup process is often supervised by SeccureDrop trainers. In general they would prefer to setup the first admin user for their partners rather than have them SSH and do it themselves. If, however, "they are installing SecureDrop themselves" then they will need to SSH in to create that first user. If this does not answer your question, I think we'd need @kushaldas to weigh in as that wording was discussed and agreed upon with him.

dachary:
> According to him, this part of the setup process is often supervised by SecureDrop trainers

Ah, that makes sense now. I don't think we distinguish between admins being in training or on their own elsewhere in the documentation. Unless I'm mistaken we always assume the admin is on their own.

dachary:
> If you have preferred wording, let me know.

Not changing **To** into **If you are installing SecureDrop yourself, to** would be good enough to clear the confusion.

dachary:
@kstohr this needs rebasing because there were changes merged that conflict with yours.

dachary:
rebased to resolve the conflict

PR #2491, opened 2017-10-31
minor typos in user facing strings

dachary:
## Status

Ready for review / Work in progress

## Description of Changes

Fixes #2401

Typo fixes courtesy Allan Nordhøy

## Testing

User facing changes only. If the test pass it means nothing was disrupted.

## Deployment

Nothing to be done.

### If you made changes to the app code:

- [x] Unit and functional tests pass on the development VM

PR #3416, opened 2018-05-15
Clarify Delete button label

### LINKED ISSUE #2419: Phrase "Delete Collection" is ambiguous
# Feature request

## Description

The button "DELETE COLLECTION" should really say "DELETE SOURCE" because when a collection is deleted, so is the source and that's not obvious.

Reported @xella

antsar:
## Status

Ready for review

## Description of Changes

Fixes #2419.

This pull request changes the label of this button:

screen shot 2018-05-15 at 1 15 41 pm

The original text was `DELETE COLLECTION`. The new text is `DELETE SOURCE AND SUBMISSIONS`. This text was proposed by @kushaldas in #2491 ([comment permalink](https://github.com/freedomofpress/securedrop/issues/2419#issuecomment-335828970)).

## Testing

1. Run app.
2. Look at button.

## Deployment

This will impact existing installs, but does not require any special procedures to deploy.

## Checklist

N/A

### DIFF
--- a/securedrop/journalist_templates/col.html
+++ b/securedrop/journalist_templates/col.html
@@ -119,7 +119,7 @@
     <input type="hidden" name="filesystem_id" value="{{ filesystem_id }}">
     <input type="hidden" name="col_name" value="{{ source.journalist_designation }}">
     <a href="#delete-collection-confirmation-modal" id="delete-collection-link">
-      <button type="button" class="sd-button danger"><i class="far fa-trash-alt"></i> {{ gettext('DELETE COLLECTION') }}</button>
+      <button type="button" class="sd-button danger"><i class="far fa-trash-alt"></i> {{ gettext('DELETE SOURCE AND SUBMISSIONS') }}</button>
     </a>
 
     <!-- Confirmation modal for entire collection deletion -->


dachary:
I believe this is the kind of modification that requires usability testing before being merged. I like your change @antsar and it seems like there is a consensus it is a good wording if I read the issue correctly. However the opinion of developers should be confirmed by usability testing with actual SecureDrop users.

Only recently did we consider usability testing: this is new ground ;-) I defer to @redshiftzero and @ei8fbd better judgment.

antsar:
Thanks for the feedback, @dachary. When you say "The translations (messages.po) must not be modified", do you mean that they should not be _manually_ modified, or they should not be modified _at all_?

The [i18n doc](https://docs.securedrop.org/en/latest/development/i18n.html#updating-strings-to-be-translated) instructs developers to run `make translate` and review/commit the `messages.pot` files. In addition to updating the `messages.pot` files, `make translate` also updated the `messages.po` files (translations).

Should the `*.po` files be excluded from the commit even though `make translate` changed them? If so, the documentation could probably use a sentence clarifying this. I'd be happy to open a doc PR if this is the case.

dachary:
> "The translations (messages.po) must not be modified", do you mean that they should not be manually modified, or they should not be modified at all?

I mean they must not be modified at all in this PR. It will be done at a later time and you don't need to convern yourself with that.

dachary:
> The i18n doc instructs developers to run make translate and review/commit the messages.pot files.

Now that you mention it... I understand the confusion. The implicit assumption is that synchronizing the code and the translation platform is done in batch and not continuously by each developer. Mainly because it requires permissions to push to a repository where [weblate](https://weblate.securedrop.club) will find new strings. The details are [in the i18n documentation](https://docs.securedrop.org/en/latest/development/i18n.html#merging-develop-into-the-weblate-fork).

The i18n documentation should start with a short introduction explaining the workflow. It developped from notes taken while working on the workflow and that is clearly missing.

ei8fdb:
👍 👍

PR #3439, opened 2018-05-16
Add explanatory text to screensaver.

### LINKED ISSUE #1975: Mitigate confusion on source interface "screen-saver"
# Feature request

## Description

Part of a commit that went in to address feature request #447 introduced a fade-out to the source interface after a minute of elapsed idle time. It's definitely a cool feature that I think we should keep, but it's a little confusing to users without any type of messaging. There should be some minimal text on-top of the white screen explaining succinctly to the user that everything is okay.

💡 Bonus "semi-serious but joking if I'm pressed about why we need it" feature --- it would be cool to have a [boss-key](https://en.wikipedia.org/wiki/Boss_key) to bring up a fake spreadsheet page if someone is in a public location and wants to suddenly hide what they are doing.

## User Stories

As a user, when I step away from the source interface after a minute of idle time - I'd like to not freak out and think I lost any work in progress when I return to a blank page :)

Example:

selection_027

antsar:
## Status

Ready for review

## Description of Changes

Fixes #1975.

This PR adds an explanatory message to the ["screen saver"](https://github.com/freedomofpress/securedrop/issues/447) in the source interface. The message reads `Interface hidden. Move your mouse here to show.`, and looks like this:

### Implementation detail

The screen saver originally used a CSS animation to set `opacity: 0` on the `<html>` element. Since all page contents should be inside this element and we want to display text on the screensaver, I had to change the approach.

This PR uses a CSS animation to show a full-page `<div>` with a white background over the page contents.

### Possible improvements / discussion

I've gone with the minimal implementation to make this easier to review/merge. However, it seems reasonable to consider some further enhancements. If these sound valuable, I'd be happy to open follow-up issue(s) and/or PR's. Feedback appreciated.

- **Activate even if mouse is over the window**
    The screen saver only activates if the mouse is left outside of the window. If the source leaves their cursor over the window, it will never activate. Not sure if this can be achieved without JS, but it could be a progressive enhancement.

- **"Boss Key"**
    #1975 also suggests adding a [Boss Key](https://en.wikipedia.org/wiki/Boss_key). This would also require JS, but might be a useful way to allow a source to obfuscate the interface when an adversary is present.

- **Replace page title and favicon**
    When the screensaver and/or boss mode is active, we could remove the favicon, and replace the title with something generic (eg. `Interface Hidden`). This also requires JS, but it might be a useful enhancement.

## Testing

1. Open the Source interface in a browser.
2. Move your cursor outside of the window.
3. Wait two minutes for screen saver to appear.

## Deployment

Nothing special AFAIK.

### DIFF
--- a/securedrop/sass/source.sass
+++ b/securedrop/sass/source.sass
@@ -5,53 +5,67 @@
 
 +base_rules
 
-@keyframes fadein
+@keyframes screensaver-stop
   from
-    opacity: 0
+    height: 100%
 
   to
-    opacity: 1
+    height: 0
 
 
-@-webkit-keyframes fadein
+@-webkit-keyframes screensaver-stop
   from
-    opacity: 0
+    height: 100%
 
   to
-    opacity: 1
+    height: 0
 
 
-@keyframes fadeout
+@keyframes screensaver-start
   0%
-    opacity: 1
+    height: 0
 
   98%
-    opacity: 1
+    height: 0
 
   100%
-    opacity: 0
+    height: 100%
 
 
-@-webkit-keyframes fadeout
+@-webkit-keyframes screensaver-start
   0%
-    opacity: 1
+    height: 0
 
   98%
-    opacity: 1
+    height: 0
 
   100%
-    opacity: 0
+    height: 100%
 
 
-html
-  animation: fadeout 120s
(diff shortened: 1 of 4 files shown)


dachary:
@toast @SaptakS your review would be much appreciated here :-)

PR #3448, opened 2018-05-16
Return correct exit code from subprocess in securedrop-admin

### LINKED ISSUE #3303: Improve return codes in securedrop-admin

antsar:
## Status

Ready for review

## Description of Changes

Fixes #3303.

This PR updates `securedrop_admin.main()` to pass through any non-zero exit codes returned by a `subprocess` call.

PR #3327 *partially* solved this problem by making `securedrop-admin` `exit 1` whenever a subprocess has a non-zero exit code. With this PR, it will return the specific code instead of `1`.

## Testing

A unit test is included. For manual testing, cause any subprocess to produce a non-zero exit code.

## Deployment

This should not impact new or existing deployments, unless there are shell scripts out there which rely on the old behavior.


admin/securedrop_admin/__init__.py
```diff
@@ -794,6 +795,10 @@ def main(argv):
             return_code = args.func(args)
         except KeyboardInterrupt:
             sys.exit(-1)
+        except subprocess.CalledProcessError as e:
+            print('ERROR (run with -v for more): {msg}'.format(msg=e),
+                  file=sys.stderr)
+            sys.exit(e.returncode)
```
heartsucker: i think we should always just return 1 on error since the other numbers aren't going to be useful for an admin. I only like other numbers to be used if there are very specific, document cases and they mean something actionable so that someone can catch them in a shell script.

E.g,. `curl`

```
EXIT CODES
       There  are  a  bunch of different error codes and their corresponding error messages
       that may appear during bad conditions. At the time of this writing, the  exit  codes
       are:

       1      Unsupported protocol. This build of curl has no support for this protocol.

       2      Failed to initialize.

       3      URL malformed. The syntax was not correct.
```

That said, printing the failed command and the exit code could be useful so they could pass it back to the team in a bug report.

antsar: Makes sense. That behavior is already implemented by PR #3327, so I guess #3303 (and this PR) can just be closed then.

redshiftzero: > i think we should always just return 1 on error since the other numbers aren't going to be useful for an admin.

Note that these CLI commands are also called by e.g. the updater GUI, to prevent code duplication, and in the future likely other GUIs used in the Admin/Journalist workstation. The goal being to provide a more user-friendly experience for less-technical admins/journalists, while also preserving the CLI functionality for the command-line inclined among us. 

@kushaldas since you had strong opinions on the return codes, can you comment on what you think would be reasonable behavior here?

heartsucker: Also as a note, these error codes wouldn't necessarily be reliable because we could run multiple subprocesses that return 5 on some error, and we wouldn't know which one popped that error code up. If we need to make a particular error actionable, we should try / catch / raise a custom exception type.

kushaldas: I would still love to have different error codes. If required, let us assign different error codes for that whole file (all `sys.exit` combinations).

heartsucker: Ok, so this would need to be defined as constants at the module level like

```
ERR_EXIT_SOMETHING_NAUGHTY = 1
ERR_EXIT_ANOTHER_SADNESS = 2
```

And we shouldn't just blindly return the subprocess's error code.

antsar: Thanks for the input! I've added explicit exit codes:

```
EXIT_SUCCESS = 0
EXIT_SUBPROCESS_ERROR = 1
EXIT_INTERRUPT = 2
```

redshiftzero:
Hey @kushaldas, also take a peek at this one when you get a chance :heart:

PR #3450, opened 2018-05-17
Remove .close in modals.

ei8fdb:
## Status

Work in progress

## Description of Changes
Removal of the X .close in modal dialogs. 

All modal dialogs have a **cancel**/**don't do that thing** button. The **cancel**/**don't do that thing** button serves the purpose of cancelling the intended action much better. The button provides the user a clear action to cancel this action.

That means the X (a.close) is not needed.

Fixes #.

## Testing

1. Choose an individual file/msg or collection to be deleted
2. Click "Delete" button to delete the individual/collection of files/msgs in the SecureDrop journalist interface
3. When the modal is displayed, the X that shows [in this image](https://imgur.com/YFo4m75) should not be seen.

## Deployment

2. New installs.

None.

### DIFF
--- a/securedrop/journalist_templates/_confirmation_modal.html
+++ b/securedrop/journalist_templates/_confirmation_modal.html
@@ -1,7 +1,6 @@
 <div id="{{ modal_data.modal_id }}" class="modal-dialog">
   <a href="#close" class="external"></a>
   <div>
-    <a href="#close" title="{{ gettext('Close') }}" class="close">X</a>
     <h2>{{ modal_data.modal_header }}</h2>
     <p>{{ modal_data.modal_body }}</p>
     <a href="#close" id="{{ modal_data.cancel_id }}" title="{{ gettext('Cancel') }}" class="btn sd-button">{{ gettext('Cancel') }}</a>
--- a/securedrop/sass/modules/_confirm-modal.sass
+++ b/securedrop/sass/modules/_confirm-modal.sass
@@ -30,21 +30,5 @@
     border-radius: 10px
     background: white
 
-  .close
-    background: $color_grey_medium
-    color: white
-    line-height: 25px
-    position: absolute
-    right: -12px
-    text-align: center
-    top: -10px
-    width: 24px
-    text-decoration: none
-    font-weight: bold
-    -moz-border-radius: 12px
-    border-radius: 12px
-    -moz-box-shadow: 1px 1px 3px #000
-    box-shadow: 1px 1px 3px #000
-
     &:hover
       background: $color_purple_medium


redshiftzero:
Hey @SaptakS, I think @ei8fdb's rationale in the PR description is reasonable, but since you added this feature, what do you think?

ei8fdb:
@SaptakS ah, I didn't know you added the modal. I'd have asked you if I did. 

The rational is that the action to achieve the close is duplicated - by the cancel/no/don't do this button and the X/close window.

SaptakS:
@redshiftzero @ei8fdb it's true that the action is redundant, but I thought from a UX point-of-view many users have a tendency to close using the X button. And hence I kept it. But given that securedrop has a very button oriented UX, I think removing the X doesn't really harm.

redshiftzero:
Hmm, I want to keep this because I find myself using it... Any others have thoughts on this one? Maybe @huertanix?

dachary:
I also like the X and use it.

heartsucker:
Do we have an example of other sites with modals to compare? Just because we use it doesn't mean it's a good pattern. Do all users understand that `X == Cancel`? Or does it look like there might be 3 options. I'm in favor of removing it even if people use it because then there is no ambiguity in what's happening.

huertanix:
The x is a common UX pattern used for dialogs in a few different places. I was just on a flight booking site and it had it on the top-right corner, it's also on every Google service dialog that I can think of and most modal ads. As some have said, big name brands using doesn't mean we need it, and having a clearly visible, localized "cancel" is another common pattern users will likely find familiar enough to use without a problem.

tl;dr: this works too imho.

redshiftzero:
Thanks for the input everyone, based on the majority view (i.e. several people commented that they use this), I'm going to close this PR for now. I think it's fine to have multiple paths to take the same action if it provides the smoothest experience for end users and doesn't introduce much maintenance burden. Let's pay attention to this in user tests and on-sites (cc @huertanix @olivemartini @harlo :sunglasses:) and if it changes, we can re-evaluate.

ei8fdb:
I've neglected this thread. Apologies, I've been busy getting the Qubes work finished.

The purpose of this proposal wasn't for any usability issue, the X or the close/delete/cancel call-to-action serve the same purpose.

It was more for a point of redundancy - there are 2 UI elements which lead to the same outcome.

I would not expect a user would find the X being a usability issue.

The question is bigger than the X/cancel button - from the interviews so far, no SD user has mentioned they delete submissions regularly.

(This was more a [simple paper-cut](https://jancborchardt.net/usability-in-free-software#Identify-paper-cuts)

So instead of introducing the interface friction (the modal) **after** they've decided to delete, it could be introduced **before** they delete it (as with [deleting a Github repository](https://help.github.com/articles/deleting-a-repository/) - step 6).

I'll add it to my usability testing list and do some testing.

huertanix:
@ei8fdb I like the Github UX flow for deleting a repo. I'm away from my SD test instance so I can't run through the current delete flow rn, but could you outline the differences in the current vs proposed change (the addition of the cancel button has been mentioned, but what else is there proposed)?

ei8fdb:
@huertanix Here is the user flow for deleting a repo on Github: 

<blockquote class="imgur-embed-pub" lang="en" data-id="a/F9qwqNO"><a href="//imgur.com/F9qwqNO"></a></blockquote><script async src="//s.imgur.com/min/embed.js" charset="utf-8"></script>

and this is the user flow for deleting a submission collection, OR a single file. Essentially they all follow the same flow.

<blockquote class="imgur-embed-pub" lang="en" data-id="a/rUb39Zq"><a href="//imgur.com/rUb39Zq"></a></blockquote><script async src="//s.imgur.com/min/embed.js" charset="utf-8"></script>

The main difference is:

In the Github version - 

1. The user confirms (by navigting to settings-delete-then manually typing the repo name) they want to delete  a submission collection/file **before** they've hit the delete call-to-action.

2. In the current design - 

the user confirms they want to delete a submission collection/file **after** they've hit the delete call-to-action (by again clicking the delete in the modal).

PR #3456, opened 2018-05-18
Yet more journalist pytest

heartsucker:
## Status

Ready for review

## Description of Changes

Towards #2877

More pytest fixtures for `test_journalist.py`. Also GH is rendering this diff terribly, so like in previous PRs, you should review individual commits.

## Testing

`make test TESTFILES=tests/test_journalist.py`


securedrop/tests/test_journalist.py
```diff
@@ -767,6 +767,27 @@ def test_user_resets_hotp(journalist_app, test_journo):
...
 
+def test_user_resets_user_hotp_format_odd(journalist_app, test_journo):
+    old_secret = test_journo['otp_secret']
+
+    with journalist_app.test_client() as app:
+        _login_user(app, test_journo['username'], test_journo['password'],
+                    test_journo['otp_secret'])
+
+        with InstrumentedApp(journalist_app) as ins:
+            app.post('/account/reset-2fa-hotp',
```
redshiftzero: Just a note for posterity that `uid=self.user.id` was removed in the dict here, but it was not actually needed as the view function in question uses the `g.user` object (this comment applies to the other uses of this `/account/reset-2fa-hotp` in this diff)

securedrop/tests/test_journalist.py
```diff
@@ -809,6 +809,40 @@ def test_user_resets_user_hotp_format_non_hexa(journalist_app, test_journo):
...
+    with journalist_app.test_client() as app:
+        _login_user(app, test_journo['username'], test_journo['password'],
+                    test_journo['otp_secret'])
+
+        mocker.patch('models.Journalist.set_hotp_secret',
+                     side_effect=TypeError(error_message))
+
+        with InstrumentedApp(journalist_app) as ins:
+            app.post('/account/reset-2fa-hotp',
+                     data=dict(uid=test_journo['id'],
```
redshiftzero: Nit: `uid=test_journo['id']` not necessary here (since it's `/account/reset-2fa-hotp` not `/admin/reset-2fa-hotp`)

securedrop/tests/test_journalist.py
```diff
@@ -896,6 +896,17 @@ def test_admin_resets_hotp_with_missing_otp_secret_key(journalist_app,
     assert 'Change Secret' in resp.data.decode('utf-8')
 
 
+def test_admin_new_user_2fa_redirect(journalist_app, test_admin):
+    with journalist_app.test_client() as app:
+        _login_user(app, test_admin['username'], test_admin['password'],
+                    test_admin['otp_secret'])
+        with InstrumentedApp(journalist_app) as ins:
+            resp = app.post(
+                '/admin/2fa?uid={}'.format(test_admin['id']),
```
redshiftzero: What about: `url_for('admin.new_user_two_factor', uid=test_admin['id'])`? (I just think `url_for` is cleaner and the right tool for the job here). The same comment applies to the other places in the diff where this kind of string manipulation for generating query parameters is done

securedrop/tests/test_journalist.py
```diff
@@ -1034,6 +1034,31 @@ def test_admin_add_user_yubikey_correct_length_with_whitespace(
     assert 'Enable YubiKey (OATH-HOTP)' in resp.data.decode('utf-8')
 
 
+def test_admin_sets_user_to_admin(journalist_app, test_admin):
+    new_user = 'rwinner'
```
redshiftzero: Ah.. hmm.. let's stick with user `'admin-set-user-to-admin-test'` ...

securedrop/tests/test_journalist.py
heartsucker: When we moved to `pytest`, we removed the other calls to `url_for` because we weren't pre-pushing an app context before executing the test functions. This is something that `Flask-Testing` was doing, but it wasn't clear what would happen when there were multiple app/request contexts pushed. I can re-add this feature to the fixtures, and I think it will be fine.

PR #3463, opened 2018-05-19
Edit and revise installation guide

pierwill:
## Status

Ready for review

## Description

Extensive edits and revisions to `install.rst`.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -4,25 +4,28 @@ Install SecureDrop
 Install Prerequisites
 ---------------------
 
-SecureDrop has some dependencies that need to be loaded onto the *Admin
-Workstation* prior to the installation of the server.
-
-To load these dependencies, from the base of the SecureDrop repository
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -117,45 +124,47 @@ the remote servers. ::
...
 
-The install process will take some time, and it will return
-the terminal to you when it is complete.
+The installation process will take some time. It will return you to the
+terminal prompt when complete.
 
-If an error occurs while
-running the install, please check all of the details of the error output.
+If an error occurs while running the install, please save the error output
+in its entirety for reference and troubleshooting.
```
dachary: what about "save and inspect"? The original wording did not recommend to save the output and this is useful advice. But it would be good to keep the encouragement to look at the output and maybe detect the simpler errors.

pierwill: I agree. I've tried another wording. Let me know if it works!

pierwill:
(I force pushed to remove some whitespace.)

PR #3472, opened 2018-05-23
Integration tests for securedrop-admin sdconfig

### LINKED ISSUE #3461: [admin CLI integration testing] securedrop-admin sdconfig
## Description

We should have integration tests for the `securedrop-admin sdconfig` command. In my opinion, decent integration tests here would be at least the cases for:

* HTTPS on source interface enabled
* HTTPS on source interface not enabled 
* Journalist notifications enabled
* Journalist notifications not enabled

Trying to make sure all possible variations are covered is a mission better accomplished via the existing unit tests (for example, testing many inputs for validation of a single item should be done in the unit tests), but the integration tests of at least the above common cases are useful as an additional safety net.

kushaldas:
## Status

Ready for review

## Description of Changes

Fixes #3461.

We have integration tests for `./securdrop-admin sdconfig`.

## Testing

```
make -C admin test
```


admin/tests/test_integration.py
```diff
@@ -0,0 +1,416 @@
...
+    child.expect('Local IPv4 address for the Application Server\:', timeout=2)
+
+
+def verify_ipv4_monserver_prompt(child):
+    child.expect('Local IPv4 address for the Monitor Server\:', timeout=2)
+
+
+def verify_hostname_app_prompt(child):
+    child.expect('Hostname for Application Server', timeout=2)
+    child.expect_exact('\x1b[0ma\x1b[0mp\x1b[0mp\x1b[36D\x1b[36C\x1b[?7h\x1b[0m\x1b[?12l\x1b[?25h', timeout=2)  # noqa: E501
```
redshiftzero: Can you add a comment explaining why we have these `expect_exact` lines?

admin/tests/test_integration.py
```diff
@@ -0,0 +1,416 @@
...
+    with open(os.path.join(SD_DIR, 'install_files/ansible-base/group_vars/all/site-specific')) as fobj:    # noqa: E501
+        data = fobj.read()
+    assert data == OUTPUT1
+
+
+def test_journalist_alert():
+    output_file = os.path.join(SD_DIR, 'install_files/ansible-base/group_vars/all/site-specific')  # noqa: E501
+    # We want to start in a clean state
+    if os.path.exists(output_file):
+        os.remove(output_file)
```
redshiftzero: We should do this state cleaning in the previous test teardown, no?

admin/tests/test_integration.py
```diff
@@ -0,0 +1,413 @@
...
+    child.expect('Local IPv4 address for the Application Server\:', timeout=2)
+
+
+def verify_ipv4_monserver_prompt(child):
+    child.expect('Local IPv4 address for the Monitor Server\:', timeout=2)
+
+
+def verify_hostname_app_prompt(child):
+    child.expect('Hostname for Application Server', timeout=2)
+    # For any default input, the text comes in the following format
```
dachary: It would help maintenance to have a helper that builds the mangled string from the expected input, **"app".split().join("\x1b[0m")** or something similar. Just an idea ;-)

kushaldas: The current form just works with copy paste from the test failures (as the real output in the test), that way we can modify them faster in future. That is why I kept it like this. The upstream also mentioned that those strings may change in future. That is why I tried to keep them as simple as possible.

redshiftzero: For the record: most of these characters are ANSI escape sequences. Here's me putting in a breakpoint and using a regex I stole from [StackOverflow](https://stackoverflow.com/questions/14693701/how-can-i-remove-the-ansi-escape-sequences-from-a-string-in-python):

```
(Pdb) ansi_escape = re.compile(r'\x1B\[[0-?]*[ -/]*[@-~]')
(Pdb) child.before
' \x1b[0m1\x1b[0m0\x1b[0m.\x1b[0m2\x1b[0m0\x1b[0m.\x1b[0m3\x1b[0m.\x1b[0m2\x1b[52D\x1b[52C\x1b[?7h\x1b[0m\x1b[?12l\x1b[?25h\x1b[?25l\x1b[?7l\x1b[52D\x1b[0m\x1b[J\x1b[0mLocal IPv4 address for the Monitor Server: \x1b[0m1\x1b[0m0\x1b[0m.\x1b[0m2\x1b[0m0\x1b[0m.\x1b[0m3\x1b[0m.\x1b[0m2\x1b[52D\x1b[0m\r\r\n\x1b[J\x1b[?7h\x1b[0m\x1b[?12l\x1b[?25h\x1b[?2004l\x1b[?1l\x1b[6n\x1b[?2004h\x1b[?25l\x1b[0m\x1b[?7l\x1b[0m\x1b[J\x1b[0m'
(Pdb) cleaned_text = child.before
(Pdb) ansi_escape.sub('', cleaned_text).
' 10.20.3.2Local IPv4 address for the Monitor Server: 10.20.3.2\r\r\n'
```

redshiftzero: so in fb9dd9ac01e7693b34b202b805bfdf942ea72167 I'm just stripping off these characters and then asserting the defaults are present - makes for more readable integration test failures (what future maintainers want), e.g.:

```
    def verify_smtp_port_prompt(child):
        child.expect('SMTP port for sending OSSEC alerts\:', timeout=2)
>       assert ANSI_ESCAPE.sub('', child.buffer) == ' 387'  # Expected default
E       AssertionError: assert ' 587' == ' 387'
E         -  587
E         ?  ^
E         +  387
E         ?  ^
tests/test_integration.py:213: AssertionError
```

redshiftzero:
I'll leave this open a bit so others have an opportunity to comment on these changes, but I'm happy with this going in. Thanks @kushaldas!

Issue #3478, opened 2018-05-29
journalist notifications: races against reboot time


dachary:
# Bug

## Description

The race condition goes as follows:

  - machine reboots
  - notification sent
  - machine reboots 24h later but reboots 1 minute faster than the previous day
  - notification is sent 23h59 minutes after the last one and suppressed

The notification will be sent the next day but the information it
contains will not reflect what happened in the past 24h because the cron job

   manage.py were-there-submissions-today

updates it daily, one hour before the machine reboots.

## Steps to Reproduce

* manually reboot the app server
* verify the journalist notification is sent
* manually reboot the app server before 24h00 passed, effectively simulating what happens when the machine reboots faster the next day
* verify the journalist notification is supprssed by looking for **suppressed** in /var/log/syslog in the mon server

## Expected Behavior

* The journalist notification is sent every day, give or take 30 minutes

## Actual Behavior

* The journalist notification is suppressed if 23h59 passed instead of 24h00

PR #3479, opened 2018-05-29
journalist notifications: allow 30 minutes variance in reboot times

dachary:
## Status

Ready for review

## Description of Changes

Fixes: #3478

The race condition goes as follows:

  - machine reboots
  - notification sent
  - machine reboots 24h later but reboots 1 minute faster than the previous day
  - notification is sent 23h59 minutes after the last one and suppressed

The notification will be sent the next day but the information it
contains will not reflect what happened in the past 24h because the cron job

   manage.py were-there-submissions-today

updates it daily, one hour before the machine reboots.

To resolve this race condition (or make it extremely unlikely) we
allow for notifications to be sent at most every 23h30.

Case 1: machine is slower to reboot the next day:

  - machine reboots
  - notification sent
  - machine reboots 24h later but reboots 5 minutes slower than the previous day
  - notification is processed 24h05 minutes after the last one, it is
    more tha 23h30 and therefore the notification is sent

Case 2: machine is faster to reboot the next day:

  - machine reboots
  - notification sent
  - machine reboots 24h later but reboots 8 minutes faster than the previous day
  - notification is processed 23h52 minutes after the last one, it is
    more tha 23h30 and therefore the notification is sent

## Testing

Testinfra tests changing the time of the app server are implemented. I do not think manual testing is possible for this race condition but I'm open to ideas.

## Deployment

* `./securedrop-admin install` must be run to fix the bug
* The cron job `ansible-base/roles/app/tasks/setup_cron.yml` updating the number of submissions in the past 24h will change to happen one hour before the reboot instead of midnight
* As a result it is possible that the journalist notification does not reflect the 

### DIFF
--- a/install_files/ansible-base/roles/app/tasks/setup_cron.yml
+++ b/install_files/ansible-base/roles/app/tasks/setup_cron.yml
@@ -20,6 +20,7 @@
   cron:
     name: Update the number of submissions in the past 24h
     job: "{{ securedrop_code }}/manage.py were-there-submissions-today"
-    special_time: daily
+    # 0 -> 23, 1 -> 0, 2 -> 1, ... 23 -> 22
+    hour: "{{ (daily_reboot_time + 23) % 24 }}"
   tags:
     - cron
--- a/install_files/ansible-base/group_vars/staging.yml
+++ b/install_files/ansible-base/group_vars/staging.yml
@@ -64,3 +64,4 @@ install_local_packages: true
 # the Apache service is configured correctly.
 securedrop_app_install_from_repo: False
 
+daily_reboot_time: 4 # An integer between 0 and 23
--- a/install_files/ansible-base/roles/ossec/files/process_submissions_today.sh
+++ b/install_files/ansible-base/roles/ossec/files/process_submissions_today.sh
@@ -17,8 +17,18 @@ function main() {
 function modified_in_the_past_24h() {
     local stamp
     stamp="$1"
+    #
+    # 24h is 1440 minutes but we subtract 30min to avoid the following race condition
+    #
+    # - machine reboots
+    # - notification sent
+    # - machine reboots 24h later but reboots 1 minute faster than the previous day
+    # - notification is sent 23h59 minutes after the last one and suppressed
+    #
+    local one_day
+    one_day=1410
     test -f "${stamp}" && \
-        find "${stamp}" -mtime -1 | \
+        find "${stamp}" -mmin "-${one_day}" | \
             grep --quiet "${stamp}"
 }
 


install_files/ansible-base/roles/app/tasks/setup_cron.yml
```diff
@@ -20,6 +20,6 @@
   cron:
     name: Update the number of submissions in the past 24h
     job: "{{ securedrop_code }}/manage.py were-there-submissions-today"
-    special_time: daily
+    hour: "{{ daily_reboot_time - 1 }}"
```
heartsucker: This needs to have a ` | default(4)` because prod playbooks won't have that value in them, so current admins will have things break.

heartsucker: Also. Why -1?

heartsucker:
This will need to be included in the release notes telling admins to run the script because we can't update prod instances, so the bug won't be "resolved" for them.

install_files/ansible-base/roles/app/tasks/setup_cron.yml
```diff
@@ -20,6 +20,7 @@
   cron:
     name: Update the number of submissions in the past 24h
     job: "{{ securedrop_code }}/manage.py were-there-submissions-today"
-    special_time: daily
+    # 0 -> 23, 1 -> 0, 2 -> 1, ... 23 -> 22
+    hour: "{{ (daily_reboot_time + 23) % 24 }}"
```
heartsucker:
```
>>> for x in range(0,24):
...     print((x + 23) % 24)
... 
23
0
1
2
3
4
5
6
7
8
9
10
11
12
13
14
15
16
17
18
19
20
21
22
```

heartsucker:
> Why would admins need to run the script?

The crontab entry needs to be changed.

dachary:
I need to test the jinja2 expression is correct

install_files/ansible-base/roles/app/tasks/setup_cron.yml
heartsucker: > This needs to have a `| default(4)` because prod playbooks won't have that value in them, so current admins will have things break.

For other reviewers, this statement was false. There is a default in a vars file in ansible, so adding a `default` filter is unnecessary.

heartsucker:
@dachary I think this is good to merge since there's not an easy way for us to manipulate the crontab other than through the admin script. Anything else is too brittle, and asking admins to fun `securedrop-admin install` periodically to fix non-critical bugs seems reasonable. I think spending additional development time on tweaking `postinst` to make this work is not worth the benefits.

dachary:
@conorsch thanks for the review :-)

dachary:
@heartsucker thanks for the merge :-)

Issue #1145, opened 2015-10-14
Remove old apt config


conorsch:
The current Ansible config adds the Freedom of Press apt repository to `/etc/apt/sources.list.d/apt_freedom_press.list`, with a repo URL of `https://apt.freedom.press trusty main`. That works great. 

However, long-running instances of SecureDrop may still have the old repo URL specified in a different file: `/etc/apt/sources.list.d/apt_pressfreedomfoundation_org.list`, with repo URL `https://apt.pressfreedomfoundation.org`. We changed the apt repo URL in November 2014 (see 97fefe45f4240d36d4d2162d7844cfa05f70ca8a), but the config does not ensure that the old repo is removed.

The discrepancy may cause problems during upgrades for long-running instances. After we update the Ansible playbooks, admin intervention will be necessary to fix the issue.

conorsch:
As with #2725, handling via `securedrop-keyring` sounds reasonable, so we're not dependent on Admin intervention. This is a cleanup task, only observed on rare occasions, so not critical to resolve.

eloquence:
We made it over the Ubuntu 14.04>16.04 without issues since this was discovered, and the number of instances that could be impacted by this is tiny. Curious if a cleanup task is still required here? If so, we may want to get it out before #4768 (Bionic/Focal).

eloquence:
We'll resolve this as part of #4768.

eloquence:
As far as I can tell, this file is [not included in backups](https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/backup/files/backup.py), so with the current reinstall -> restore path for moving to Ubuntu 20.04, the issue would be resolved implicitly as part of the upgrade. Please let me know if I'm wrong about that. 

I'll leave it open in case we do revisit the question of in-place upgrades.

eloquence:
The only officially supported migration path to Ubuntu 20.04 was a reinstall from a backup (which does not include this old file), so this should be fully resolved. Closing.

PR #3494, opened 2018-06-04
Update kernels to 4.4.135

emkll:
## Status

This needs to be merged after the debs are uploaded to the apt server.
Ready for review

## Description of Changes

* Bump kernels to 4.4.135
* Add linux-firmware-image for hardware compatibility
* Preserve 3.14.79 and 4.4.115 kernels in case 4.4.135 kernels do not boot
* Preserve existing kernel boot order

## Testing
- [ ] Ensure `apt-test.freedom.press` is used for staging VMs
- [ ] Ensure 'apt.freedom.press` is used for prod VMs
- [ ] Make sure the apt-test key does not get added to prod VMs

Please update this PR with the testing matrix below:

Test | VMs  | Clean install- NUC | PowerEdge R620 | Clean install - Mac Mini | Upgrade - Mac mini | Upgrade - NUC w/ 3.14| HP Proliant DL385 G7
| ------------ | :---: | ------------- | ------------- | ------------- | ------------- | ------------- | ------------- |
| Boots | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:| | :heavy_check_mark: | :heavy_check_mark: |
| Paxtest | :heavy_check_mark: | :heavy_check_mark: | | :heavy_check_mark: | | :heavy_check_mark: |
| [Spectre/Meltdown test](https://meltdown.ovh) | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | | :heavy_check_mark: | :heavy_check_mark: |
| End-to-end  | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: | | :heavy_check_mark: |
| Testinfra tests | :heavy_check_mark: | N/A | N/A | N/A | N/A | N/A | N/A|
| No errors/grsec denies in logs after 24+h  | :heavy_check_mark: |:heavy_check_mark: | | | | :heavy_check_mark: |
| No (unusual) ossec alerts after 24+h  | :heavy_check_mark: | :heavy_check_mark: | | | | :heavy_check_mark: |

## Deployment

Kernel debs (both the `securedrop-grsec` metapackageand `linux-image-4.4.135-grsec_4.4.135-grsec-1_amd64` need to be uploaded to the apt server).

If the instances fail to boot, instructions in https://docs.securedrop.org/en/stable/upgrade/0.5.x_to_0.6.html are still valid.


b-meson:
I installed on a Dell PowerEdge R620. Preliminary tests suggest it is going well. 
# Basic Server Testing
 - [x] I can access both the source and journalist interfaces after upgrade and it is upgraded to `kernel-update-4.4.135`.
 - [x] I can SSH into both machines over Tor
 - [x] AppArmor is loaded on app and mon
 - [x] Both servers are running grsec kernels
 - [x] iptables rules loaded
 - [x] OSSEC emails begin to flow after install
 - [x] OSSEC emails are decrypted to correct key and I am able to decrypt them

# Command Line User Generation
 - [x] Can successfully add admin user

b-meson:
I did a fresh installed on a HP Proliant DL385 Gen 7th. Preliminary tests suggest it is well supported. 

# Basic Server Testing
 - [x] I can access both the source and journalist interfaces after upgrade and it is upgraded to `kernel-update-4.4.135`.
 - [x] I can SSH into both machines over Tor
 - [x] AppArmor is loaded on app and mon
 - [x] Both servers are running grsec kernels
 - [x] iptables rules loaded
 - [x] OSSEC emails begin to flow after install
 - [x] OSSEC emails are decrypted to correct key and I am able to decrypt them

# Command Line User Generation
 - [x] Can successfully add admin user

# Command Line User Generation
 - [x] Can successfully add admin user

# Application Acceptance Testing

 - [x] On submit page, file less than 500 MB submitted successfully
 - [x] On submit page, short message submitted successfully

## Journalist Interface
###Login base cases
 - [x] Can log in with 2FA tokens
## Individual source page
 - [x] You can submit a reply and a flashed message and new row appears
 - [x] Clicking "Delete collection" and the source and docs are deleted

emkll:
To recap a conversation in standup today, we discussed the process of merging the kernel updates in the develop branch. Because 4.4.135 kernels will not be on `apt.freedom.press`, staging VMs (outside of CI) will be broken. We will need to make staging VMs point to `apt-test.freedom.press` at least temporarily, for staging VMs not to break. I have updated this pr to use apt-test for staging VMs only.

@conorsch I invite you to take a close look at the apt-test key addition logic, and would be interested in any suggestion to make the logic more robust, as we obviously want to ensure the apt-test key never makes it's way into a production install.

Given the scope of changes required and the kernel string version bump doesn't break CI, I would suggest we perhaps reconsider this approach, forgo staging VMs for testing and simply resort to prod VMs for the duration of the QA period.

conorsch:
>  take a close look at the apt-test key addition logic, and would be interested in any suggestion to make the logic more robust, as we obviously want to ensure the apt-test key never makes it's way into a production install.

So far so good! I'll tweak the logic somewhat to avoid use of `group_names` explicitly, and instead set the _default_ pubkey filepath to the prod key, and override the staging vars to the test key, same as you've down for the URL. 

> Given the scope of changes required and the kernel string version bump doesn't break CI, I would suggest we perhaps reconsider this approach, forgo staging VMs for testing and simply resort to prod VMs for the duration of the QA period.

Reasonable, however the prod VMs will still need to be bootstrapped to use the apt-test repos in order to QA the new kernel. With the changes to the test key you provide, staging will use the latest kernels, giving us much more visibility into interactions with the rest of the application. 

The ultimate problem remains the package dependency in `securedrop-grsec`: even though the apt-test key changes suffice to resolve the problem for the staging environment, I suspect we'll still have trouble with the prod environment post-merge (unless overridden each and every time the prod VMs are spun up). Testing now, will report back.

install_files/securedrop-grsec/DEBIAN/postinst
```diff
@@ -0,0 +1,48 @@
...
+case "$1" in
+    configure)
+
+    # If the SecureDrop instance is running a rolled-back kernel, this will
+    # preserve the kernel boot priority specfied in its ordinal form by its
+    # canonical form. (In this specific case, 1>2 as GRUB_DEFAULT in
+    # /etc/default/grub indicates the instance is running 3.14.79-grsec)
+    # Otherwise, we want to run the latest 4.4 series kernel.
+    if grep -qE '^GRUB_DEFAULT=\"1>2\"' /etc/default/grub; then
+        sed -i 's/^\(GRUB_DEFAULT=\)"1>2"/\1\"Advanced options for Ubuntu>Ubuntu, with Linux 3.14.79-grsec\"/' /etc/default/grub
```
conorsch: There's a case where this regex won't catch a valid configuration: if single quotes instead of double quotes were used to encase the value. (Haven't tested, but perhaps no quotes at all should be supported, as well.) We should expand the regex to handle a set of possible quote characters, to make sure we're preserving rolled back instances.

testinfra/common/test_fpf_apt_repo.py
```diff
@@ -15,9 +18,17 @@ def test_fpf_apt_repo_present(File):
...
+    # If the var fpf_apt_repo_url test var is apt-test, validate that the
+    # apt repository is configured on the host
+    if test_vars.fpf_apt_repo_url == "https://apt-test.freedom.press":
+        f = File('/etc/apt/sources.list.d/apt_test_freedom_press.list')
+        assert f.contains('^deb \[arch=amd64\] '
+                          'https:\/\/apt-test\.freedom\.press trusty main$')
+    else:
+        f = File('/etc/apt/sources.list.d/apt_freedom_press.list')
+        assert f.contains('^deb \[arch=amd64\] https:\/\/apt\.freedom\.press '
+                          'trusty main$')
```
conorsch: Clunky logic, but you're right to implement it! We could start forcing a specific filepath for the apt repo, which would help with problems like #1145. Here, the only conditional is the filepath—we consolidate the assert statement to reference `re.escape(test_vars.fpf_apt_repo_url`.

install_files/ansible-base/roles/grsecurity/tasks/from_fpf_repo_install_grsec.yml
```diff
@@ -29,7 +29,7 @@
   apt:
     pkg: securedrop-grsec
     state: latest
-  async: 300
+  async: 500
```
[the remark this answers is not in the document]
conorsch: Reporting multiple failures now. Perhaps 300 * n, where n is the number of images to install.

install_files/securedrop-grsec/DEBIAN/postinst
```diff
@@ -0,0 +1,48 @@
...
+    fi
+    # update grub to set the new default
+    update-grub2
```
conorsch: Whether we need to run this `update-grub` task here is dependent on the results of the grep above, so suggest moving it inside the if block.

conorsch:
Excellent attention to detail here, @emkll. The new postinst regex in particular bears careful scrutiny in QA, but the test plan you've laid out will ensure we're keeping a close eye on behavior.

Issue #3247, opened 2018-04-09
Remove 2FA for TTY logins


eloquence:
# Epic

## Description

As discussed in #1283, 2FA for TTY logins (keyboard attached to App or Mon server) does not buy us additional security; the user at this point can skip 2FA via single user mode.

We should phase out 2FA for TTY logins for all installs to avoid configuration drift. Since this is a change to existing security configuration of the server, clear communication, up to and including a manual confirmation step for existing installs, is essential. 

## Tasks

- [x] Announce that we are phasing out 2FA for TTY logins - #3248
- [ ] Remove 2FA for TTY logins for all installs (must show clear message, potentially confirmation step, for existing installs).
- [ ] If appropriate, create docs for 0.8.0 update (e.g., instruct admins to run `securedrop-admin install`)

KwadroNaut:
I'd like to hear a reasoning for confirmation? If Ida want to keep it, she wants a setup that's different from default. When she wants things to be up-to-date, she should *not* be asked anything while upgrading, otherwise her instance will be outdated?

msheiny:
Yeah im with @KwadroNaut here... i dont understand why we need a confirmation. We want all securedrops to be uniform in their settings.

emkll:
I agree, no confirmation will avoid any potential edge cases. Since this upcoming behavior was well documented and communicated in the release notes, and there were no objections, let's move forward without a confirmation step.

PR #3504, opened 2018-06-08
Add generic kernel troubleshooting/rollback docs

### LINKED ISSUE #3498: [0.8.0] Add kernel troubleshooting docs
As part of the 0.8.0 release, we are anticipating that we will ship an upgrade from grsecurity-kernel patched kernel version 4.4.115 to 4.4.135 (this will not impact instances with downgraded kernels). This is being done via  #3494.

While  we don't anticipate new issues with this kernel and will do hardware testing both on officially recommended hardware and on some additional server models, we must provide downgrade instructions to users in case of unanticipated issues that aren't immediately resolvable.

# Implementation

For the 0.5 to 0.6 upgrade, we provided extensive kernel troubleshooting docs here:
 https://docs.securedrop.org/en/release-0.7/upgrade/0.5.x_to_0.6.html

I suggest adding a generalized version of these instructions to the admin guide, and referencing them in the upgrade instructions for 0.7 to 0.8, instead of duplicating the whole page.

eloquence:
These instructions are largely identical to the troubleshooting section
in the 0.5->0.6 upgrade docs, but they do not reference specific
kernel versions.

Resolves #3498

## Status

Ready for review

## Recommended testing procedure

I recommend testing whether this procedure works for the planned 4.4.115->4.4.135 kernel update as part of the review.

## Checklist


eloquence:
One part I'm not entirely clear on (CC @msheiny @emkll) are the implications of #3247 for these docs. But here's my understanding - please tell me if I'm off base:

- Our goal is to remove 2FA for 0.8.0, but that change would only take place after a playbook run, which will never happen if the kernel panics. So these instructions still apply for 0.8.0.

- After 0.8.0, users will be able to boot into either kernel without using single user mode. We can revise these instructions to say so once 0.8.0 has shipped.

docs/kernel_troubleshooting.rst
```diff
@@ -0,0 +1,183 @@
+Troubleshooting Kernel Updates
+==============================
+For security reasons, new versions of the Linux kernel running on your your
```
dachary: s/your your/your/

docs/kernel_troubleshooting.rst
```diff
@@ -0,0 +1,183 @@
...
+against :ref:`recommended hardware <Specific Hardware Recommendations>`. If
+things do go wrong (e.g., the server does not boot after a kernel update), 
+the following instructions will help you to roll back to the previous, 
+working kernel. You can then `report compatibility issues`_ to us so we can work
+together to resolve them as quickly as possible.
+
+First, you need to physically access each server. Power down the server
+(safely if possible), attach required peripherals (keyboard, monitor),
+and power the server back up. If you did not configure Google
+Authenticator for console access, you will need to use single user mode
```
dachary: We refer to FreeOTP by default elsewhere in the documentation. A link to servers_2fa may help in case the person does not remember what it is about exactly because they chose Google Authenticator instead.

docs/kernel_troubleshooting.rst
```diff
@@ -0,0 +1,183 @@
...
+analysis.
+
+You can share ``server-facts.log``, ``syslog`` and ``dmesg`` with us as follows:
+
+-  If you are a member of our Support Portal, please create a new issue
+   and attach the files to it.
+-  Alternatively, email us at securedrop@freedom.press 
+   (`GPG encrypted <https://securedrop.org/sites/default/files/fpf-email.asc>`__) 
+   with the subject “SecureDrop kernel facts” and the files attached.
+
```
emkll: I suggest adding some language to discourage users from sharing raw logs and sensitive information to public forums such as GitHub or securedrop.club, but rather to share only high-level information.

docs/kernel_troubleshooting.rst
emkll: Suggest a bit more clarity here: Kernel updates address known bugs and security vulenrabilities in the Linux kernel. It is important for your SecureDrop instance to run the latest version.

docs/kernel_troubleshooting.rst
```diff
@@ -0,0 +1,183 @@
...
+login prompt and log in.
+
+Once you are logged in, check to see if you have network access. If you do, then  
+your instance is having an issue with the newer kernel. In that case, we need to 
+temporarily set an older kernel as the default.
+
+Roll Back to the Old Kernel
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. important:: It is of critical importance for the near term security of your instance
```
emkll: Perhaps remove near-term?

docs/kernel_troubleshooting.rst
eloquence: thanks! fixed.

docs/kernel_troubleshooting.rst
eloquence: per discussion during standup, this seems obsolete anyway. reworded accordingly.

docs/kernel_troubleshooting.rst
eloquence: done!

docs/kernel_troubleshooting.rst
eloquence: reworded

docs/kernel_troubleshooting.rst
eloquence: done!

PR #3506, opened 2018-06-08
Added passlib for more flexible password hashing

### LINKED ISSUE #2918: Use passlib for password hashing
# Feature request

## Description

Using [`passlib`](https://pythonhosted.org/passlib/) for password hashing will give us future proof and easily migrateable password management should we change password hashing algorithms/parameters.

Possibly conflicts with #1609 depending on what the underlying cryptolib is.

## User Stories

As a dev, I don't want to have to write boilerplate migration code if we bump the number of scrypt rounds.

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #2918

- Adds `passlib` library for handing password hashing and verification
- Replace `scrypt` with `argon2` 
- Adds database column & migration for new hashes

> Note: This shouldn't be merged until after the 0.8.0 release is branched off `develop`

## Testing

Check the manual tests.

```
make test
```

Do a manual test

```
git checkout develop
make build-debs
vagrant up /staging/

# add a user, login

git checkout passlib
make build-debs
vagrant provision /staging/

# check can still login
# add new user
```

## Deployment

This includes legacy support for old password hashing schemes, so old journalists accounts will still be able to log in. This adds a new apt dependency, but it's in the security lists so we won't error out.

### DIFF
--- a/docs/development/setup_development.rst
+++ b/docs/development/setup_development.rst
@@ -120,7 +120,7 @@ created by default when running ``make dev``. In addition, sources and
 submissions are present. The test users have the following credentials:
 
 * **Username:** ``journalist`` or ``dellsberg``
-* **Password:** ``WEjwn8ZyczDhQSK24YKM8C9a``
+* **Password:** ``correct horse battery staple profanity oil chewy``
 * **TOTP secret:** ``JHCO GO7V CER3 EJ4L``
 
 .. note:: The password and TOTP secret are the same for both accounts for
--- a/securedrop/alembic/versions/2d0ce3ee5bdc_added_passphrase_hash_column_to_.py
+++ b/securedrop/alembic/versions/2d0ce3ee5bdc_added_passphrase_hash_column_to_.py
@@ -1,0 +1,53 @@
+"""added passphrase_hash column to journalists table
+
+Revision ID: 2d0ce3ee5bdc
+Revises: faac8092c123
+Create Date: 2018-06-08 15:08:37.718268
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '2d0ce3ee5bdc'
+down_revision = 'faac8092c123'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('journalists', sa.Column('passphrase_hash', sa.String(length=256), nullable=True))
+
+
+def downgrade():
+    # sqlite has no `drop column` command, so we recreate the original table
+    # then load it from a temp table
+
+    op.rename_table('journalists', 'journalists_tmp')
+
+    op.create_table('journalists',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('username', sa.String(length=255), nullable=False),
+        sa.Column('pw_salt', sa.Binary(), nullable=True),
+        sa.Column('pw_hash', sa.Binary(), nullable=True),
+        sa.Column('is_admin', sa.Boolean(), nullable=True),
+        sa.Column('otp_secret', sa.String(length=16), nullable=True),
+        sa.Column('is_totp', sa.Boolean(), nullable=True),
+        sa.Column('hotp_counter', sa.Integer(), nullable=True),
+        sa.Column('last_token', sa.String(length=6), nullable=True),
+        sa.Column('created_on', sa.DateTime(), nullable=True),
+        sa.Column('last_access', sa.DateTime(), nullable=True),
+        sa.PrimaryKeyConstraint('id'),
+        sa.UniqueConstraint('username')
+    )
+
+    conn = op.get_bind()
+    conn.execute('''
+        INSERT INTO journalists
+        SELECT id, username, pw_salt, pw_hash, is_admin, otp_secret, is_totp,
+               hotp_counter, last_token, created_on, last_access
+        FROM journalists_tmp
+    ''')
+    
+    op.drop_table('journalists_tmp')
--- a/securedrop/alembic/versions/2d0ce3ee5bdc_added_passphrase_hash_column_to_.py
+++ b/securedrop/alembic/versions/2d0ce3ee5bdc_added_passphrase_hash_column_to_.py
@@ -1,7 +1,7 @@
 """added passphrase_hash column to journalists table
 
 Revision ID: 2d0ce3ee5bdc
-Revises: faac8092c123
+Revises: fccf57ceef02 
 Create Date: 2018-06-08 15:08:37.718268
 
 """
@@ -11,7 +11,7 @@ import sqlalchemy as sa
 
 # revision identifiers, used by Alembic.
 revision = '2d0ce3ee5bdc'
-down_revision = 'faac8092c123'
+down_revision = 'fccf57ceef02'
 branch_labels = None
 depends_on = None
 
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -19,6 +19,7 @@ except ImportError:
 from flask import current_app, url_for
 from itsdangerous import TimedJSONWebSignatureSerializer, BadData
 from jinja2 import Markup
+from passlib.hash import argon2
 from sqlalchemy import ForeignKey
 from sqlalchemy.orm import relationship, backref
 from sqlalchemy import Column, Integer, String, Boolean, DateTime, Binary
@@ -30,6 +31,9 @@ from db import db
 LOGIN_HARDENING = True
 if os.environ.get('SECUREDROP_ENV') == 'test':
     LOGIN_HARDENING = False
+    ARGON2_PARAMS = dict(memory_cost=2**3, rounds=1, parallelism=1)
+else:
+    ARGON2_PARAMS = dict(memory_cost=2**16, rounds=4, parallelism=2)
 
 
 def get_one_or_else(query, logger, failure_method):
@@ -269,13 +273,13 @@ class InvalidPasswordLength(PasswordError):
        password length.
     """
 
-    def __init__(self, password):
-        self.pw_len = len(password)
+    def __init__(self, passphrase):
+        self.passphrase_len = len(passphrase)
 
     def __str__(self):
-        if self.pw_len > Journalist.MAX_PASSWORD_LEN:
-            return "Password too long (len={})".format(self.pw_len)
-        if self.pw_len < Journalist.MIN_PASSWORD_LEN:
+        if self.passphrase_len > Journalist.MAX_PASSWORD_LEN:
+            return "Password too long (len={})".format(self.passphrase_len)
+        if self.passphrase_len < Journalist.MIN_PASSWORD_LEN:
             return "Password needs to be at least {} characters".format(
                 Journalist.MIN_PASSWORD_LEN
             )
@@ -302,6 +306,7 @@ class Journalist(db.Model):
 
     created_on = Column(DateTime, default=datetime.datetime.utcnow)
     last_access = Column(DateTime)
+    passphrase_hash = Column(String(256))
     login_attempts = relationship(
         "JournalistLoginAttempt",
         backref="journalist")
@@ -322,9 +327,6 @@ class Journalist(db.Model):
             self.username,
             " [admin]" if self.is_admin else "")
 
-    def _gen_salt(self, salt_bytes=32):
-        return os.urandom(salt_bytes)
-
     _LEGACY_SCRYPT_PARAMS = dict(N=2**14, r=8, p=1)
 
     def _scrypt_hash(self, password, salt):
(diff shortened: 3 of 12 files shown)


heartsucker:
Ok, not gonna lie I am actually blown away that I managed to get a PR in that passed CI on the first attempt. This isn't a useful thing to say, but it's late, and I expected to have build error emails. Anyway.

docs/development/setup_development.rst
```diff
@@ -103,7 +103,7 @@ created by default when running ``make dev``. In addition, sources and
 submissions are present. The test users have the following credentials:
 
 * **Username:** ``journalist`` or ``dellsberg``
-* **Password:** ``WEjwn8ZyczDhQSK24YKM8C9a``
```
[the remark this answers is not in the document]
heartsucker: If we really want to keep it, I can do that, but it would mean adding (a small amount) of complexity to get code. It's already very terse, and we'd basically end up factoring out a single function call to get this behavior.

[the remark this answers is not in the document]
dachary: I am in favor of your change.

securedrop/models.py
```diff
@@ -31,6 +32,9 @@
 # precisely control which code paths are exercised.
 if os.environ.get('SECUREDROP_ENV') == 'test':
     LOGIN_HARDENING = False
+    ARGON2_PARAMS = dict(memory_cost=2**3, rounds=1, parallelism=1)
```
redshiftzero: hey is this strictly necessary (i.e. did this provide a significant speedup)? I'm asking because having test-only code paths in prod code is in my opinion an anti-pattern and this if/else in particular is one that we can hopefully snip out some point soon

securedrop/models.py
```diff
@@ -259,28 +264,29 @@ def __repr__(self):
...
-        self.check_password_acceptable(password)
+    def set_password(self, passphrase):
+        self.check_password_acceptable(passphrase)
+
+        # "migrate" from the legacy case
+        if not self.passphrase_hash:
+            self.passphrase_hash = \
+                argon2.using(**ARGON2_PARAMS).hash(passphrase)
+            self.pw_hash = None
+            self.pw_salt = None
```
redshiftzero: setting `pw_salt` to None is such an important implementation detail that it should have a comment to indicate to everyone that in `passlib.hash.argon2` salts are auto-generated and stored in the `passphrase_hash` column

securedrop/qa_loader.py
```diff
@@ -64,6 +64,12 @@ def new_journalist():
                                          nullable=False),
                             pw,
                             random_bool())
+    if random_bool():
+        # to add legacy passwords back in
```
redshiftzero: nice 😎

securedrop/models.py
heartsucker:
```
heartsucker@7bee4bf4317f:/home/heartsucker/code/freedomofpress/securedrop/securedrop$ python -m timeit 'from passlib.hash import argon2; argon2.using(memory_cost=2**3, rounds=1, parallelism=1).hash("test")'
10000 loops, best of 3: 189 usec per loop
heartsucker@7bee4bf4317f:/home/heartsucker/code/freedomofpress/securedrop/securedrop$ python -m timeit 'from passlib.hash import argon2; argon2.using(memory_cost=2**16, rounds=4, parallelism=1).hash("test")'
10 loops, best of 3: 284 msec per loop
```

So I'm gonna say no, not necessary.

securedrop/models.py
heartsucker: Fixed in both place.

securedrop/models.py
heartsucker: This was removed.

redshiftzero:
Hey @heartsucker, can you confirm you were able to run through your test plan successfully? I'm having trouble upgrading in staging VMs:

```
# dpkg -i /root/securedrop-app-code-0.8.0~rc1-amd64.deb
(Reading database ... 51586 files and directories currently installed.)
Preparing to unpack .../securedrop-app-code-0.8.0~rc1-amd64.deb ...
Unpacking securedrop-app-code (0.8.0~rc1) over (0.8.0~rc1) ...
dpkg: dependency problems prevent configuration of securedrop-app-code:
 securedrop-app-code depends on libpython2.7-dev; however:
  Package libpython2.7-dev is not installed.

dpkg: error processing package securedrop-app-code (--install):
 dependency problems - leaving unconfigured
 securedrop-app-code
# apt-cache policy libpython2.7-dev
libpython2.7-dev:
  Installed: (none)
  Candidate: 2.7.6-8ubuntu0.4
  Version table:
     2.7.6-8ubuntu0.4 0
        500 http://security.ubuntu.com/ubuntu/ trusty-security/main amd64 Packages
```

heartsucker:
@redshiftzero You are right. I was using a dirty vagrant box when I did that test. I pushed a change that should fix it.

heartsucker:
Aww crap. I overwrote my fixes for your comments with a force push. This is gonna need a fix before you re-review.

conorsch:
@heartsucker Recoverable via `git reflog` ?

heartsucker:
@conorsch Possibly, but given that I don't access to one of the machines the code is on, fixing again is probably easier. :)

heartsucker:
Can someone kick the CI job since I think this is a transient error with PyPI. It's been acting up with me today.

heartsucker:
> Perhaps reinstalling the securedrop-app-code package as part of the backup script might be an easy way to fix this and ensure forward compatibility for backups.

This being the back job that ansible does?

> It seems like a new column / migration is required if we decide to update parameters or algorithm. 

If we update the params, we can always do:

```python
if self.passphrase_hash.startswith('$argon2i$known_old_args$'):
    new_hash = argon2.using(**ARGS).hash(submitted_pw)
else:
    new_hash = None

valid = argon2.verify(submitted_pw)
if valid and new_hash:
    self.passphrase_hash = new_hash

return valid
```

This works because the params are stored as part of the hash string:

```
$argon2i$v=19$m=65536,t=4,p=2$N+ac856zNuZ8r7U25nwvJQ$oBmgTQZXZvrM+nzpMyXyWw
```

> Finally, something we should at least be aware of is that passlib appears to be the work of a single developer, and there have been no commits in the past year.

Should we reach out and discuss maintainability? Also for a lib like this, I imagine there's not a whole lot to do once the basic features are in place.

heartsucker:
Booping this PR for final sign off.

zenmonkeykstop:
Test plan fails for me at the vagrant provision /staging/ step - looks like libpython2-dev is not getting installed:
`failed: [app-staging] (item=linux-image-generic-lts-xenial) => {"changed": false, "cmd": ["apt-get", "remove", "-y", "linux-image-generic-lts-xenial"], "delta": "0:00:00.551679", "end": "2018-07-12 21:07:06.030297", "item": "linux-image-generic-lts-xenial", "msg": "non-zero return code", "rc": 100, "start": "2018-07-12 21:07:05.478618", "stderr": "E: Unmet dependencies. Try 'apt-get -f install' with no packages (or specify a solution).", "stderr_lines": ["E: Unmet dependencies. Try 'apt-get -f install' with no packages (or specify a solution)."], "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\nPackage 'linux-image-generic-lts-xenial' is not installed, so not removed\nYou might want to run 'apt-get -f install' to correct these:\nThe following packages have unmet dependencies:\n securedrop-app-code : Depends: libpython2.7-dev but it is not going to be installed", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information...", "Package 'linux-image-generic-lts-xenial' is not installed, so not removed", "You might want to run 'apt-get -f install' to correct these:", "The following packages have unmet dependencies:", " securedrop-app-code : Depends: libpython2.7-dev but it is not going to be installed"]}
`

heartsucker:
@zenmonkeykstop Fixed in the latest commit. The explanation is in the commit message.

conorsch:
@heartsucker Regarding the new dependency logic addressing @zenmonkeykstop's comment, the implementation proposed installs all the appserver dependencies on _both_ servers, so mon-staging will have redis, supervisor, etc. The logic only affects staging hosts, not prod, but would still prefer to clean it up a bit.

What if we switched from using `dpkg -i <deb>` to `apt install ./<deb>` in the install_local_packages logic? Would have to test, but that may be intelligent enough to fetch the dependencies listed in the control file. It'd also avoid the need to reinstall the app dependencies on mon-staging.

conorsch:
Needs a rebase for conflicts with alembic, also for passing CI given merge of #3654. Following up on my previous comment about the liberal installation of dependencies, will test the proposed resolution locally and push up a patch if that resolves.

heartsucker:
@conorsch @zenmonkeykstop I think we actually don't need the last commit I added that included the dependencies. We only need it for exactly the test case above when updating a staging box from a non-`libpython2.7-dev`-having VM to one that requires it. Adding a fix will just be cruft. I think it's sufficient to leave it out.

securedrop/alembic/versions/2d0ce3ee5bdc_added_passphrase_hash_column_to_.py
```diff
@@ -0,0 +1,53 @@
...
+Create Date: 2018-06-08 15:08:37.718268
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '2d0ce3ee5bdc'
+down_revision = 'faac8092c123'
```
redshiftzero: test CI job is failing here as `down_revision` needs to be updated to `fccf57ceef02` after merge of #3619

securedrop/tests/migrations/migration_2d0ce3ee5bdc.py
```diff
@@ -0,0 +1,212 @@
...
+    @staticmethod
+    def add_source():
+        filesystem_id = random_chars(96) if random_bool() else None
+        params = {
+            'filesystem_id': filesystem_id,
+            'journalist_designation': random_chars(50),
+            'flagged': bool_or_none(),
+            'last_updated': random_datetime(nullable=True),
+            'pending': bool_or_none(),
+            'interaction_count': random.randint(0, 1000),
```
redshiftzero: and we'll need to add the UUID column here otherwise the unique constraint on UUID will cause an integrity error if UUID is null everywhere in the sources table

heartsucker:
@redshiftzero @conorsch Ok thiiiiiiissss time I think I got it :D

redshiftzero:
These changes look good, thanks @heartsucker! 

When you get a chance @zenmonkeykstop can you give this an upgrade test? Once you confirm a smooth upgrade, I'll approve this for merge.

zenmonkeykstop:
Tested as follows:

- Cloned fresh copy of freedomofpress:securedrop and ran:
```
 make build-debs # needed or else apt server setup fails in upgrade scenario
 molecule converge -s upgrade
```
  - logged into app-staging VM and used manage.py to add 'admin' user
  - verified app version in journalist interface is 0.8.0
  - logged into journalist interface with admin user and created journo user (did not log in as journo)
  - switched to passlib branch and ran:
```
 make build-debs
 molecule side-effect -s upgrade
```
  - verified app version is 0.9.0~rc1
  - logged into journalist interface with journo user successfully
  - logged in with admin user successfully
  - created a new journo2 user, logged in as journo2 user successfully.

Looks good to me based on the above testing.

redshiftzero:
Beautiful, thanks for the crystal clear report @zenmonkeykstop!

zenmonkeykstop:
Egg on face time: I took a look at the app db after finishing testing and realized the passphrase_hash column hadn't been populated. Looks like there was a flaw in my upgrade testing - the debs built on the develop branch before switching to the passlib were being used for the upgrade! Revised procedure is as follows:

  - cloned fresh copy of freedomofpress:securedrop

   - switched to passlib branch and ran:

```
$ make build-debs 
$ molecule converge -s upgrade
```
  - logged into app-staging VM and used manage.py to add 'admin' user
  - logged into journalist interface with admin user, verified app version is 0.8.0, and created journo user (did NOT log in with journo)
- Ran:
```
$ molecule side-effect -s upgrade
```
  - verified app version is 0.9.0~rc1
  - logged into journalist interface with journo user successfully
  - logged in with admin user successfully
  - created a new journo2 user, logged in as journo2 user successfully.
  - logged into app-staging VM, checked sqlite db, verified passphrase_hash field populated and pw_hash field empty for all users above

Luckily the results are the same, no functional change, but this time the db change is verified as well!

conorsch:
Fantastic, thanks for following up, @zenmonkeykstop! Good to have the initial results confirmed now. N.B. I opened #3669 to track docs creating for the upgrade testing scenario. Had we had strong docs for that flow, we could have minimized the confusion around what specifically was being tested and how.

PR #3539, opened 2018-06-18
Docs: Edit admin guide

pierwill:
## Status

Work in progress

## Description of Changes

So far: Edits the "Responsibilities" section.

## Questions

Do we want the admin guide to read entirely in terms of responsibilities? Currently, after the responsibilities section, it drops into a how-to guide ("Now you can add new logins for the journalists…"). Seems like maybe this is left over from a previous installation doc, maybe?

I'll probably go ahead try something out, then update this PR, but feedback from FPF folks would be welcome. (@redshiftzero @dachary @conorsch)

## Testing

- activate virtualenv
- `make docs`
- http://localhost:8000/admin.html

### DIFF
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -3,32 +3,36 @@ Admin Guide
 
 .. include:: ./includes/provide-feedback.txt
 
-You (the admin) should have your own username and passphrase, plus
-two-factor authentication through either the FreeOTP app
-on your smartphone or a YubiKey.
+The SecureDrop architecture contains multiple machines and hardened servers.
+While we have automated many of the installation and maintenance tasks, a
+skilled Linux admin is required to responsibly run the system.
+
+This section outlines the tasks the admin is responsible for in order to
+ensure that the SecureDrop server continues to be a safe place for sources to
+talk to journalists.
 
 .. _Responsibilities:
 
 Responsibilities
 ----------------
 
-The SecureDrop architecture contains multiple hardened servers, and while we have
-automated many of the installation and maintenance tasks, a skilled Linux
-admin and some manual intervention is required to responsibly run the system.
+Maintaining Credentials
+~~~~~~~~~~~~~~~~~~~~~~~
 
-This section outlines the tasks the admin is responsible for in order to
-ensure that the SecureDrop server continues to be a safe place for sources to
-talk to journalists.
+The admin should have her own username, passphrase, and two-factor
+authentication method (via smartphone app or YubiKey). Admins are also
+responsible for tracking user credentials and encouraging best practices. (See
+:doc:`passphrases` and :doc:`passphrase_best_practices`.)
 
-Keep Your SecureDrop Server Updated
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Updating the SecureDrop Servers
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You should maintain awareness of SecureDrop updates and take any required
+The admin should be aware of all SecureDrop updates and take any required
 manual action if requested in the `SecureDrop Release Blog`_. We recommend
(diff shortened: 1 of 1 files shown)


redshiftzero:
> Do we want the admin guide to read entirely in terms of responsibilities? Currently, after the responsibilities section, it drops into a how-to guide ("Now you can add new logins for the journalists…"). Seems like maybe this is left over from a previous installation doc, maybe?

Ha, yep. It was a solely how-to guide and then the admin responsibilities section got added. We didn't intend for the entire admin guide to read in terms of responsibilities, as I think it makes sense that part of the guide is how tos and FAQs. Reworking this to flow a bit more nicely would be :100:. Honestly at this point for both the journalist and admin guides, they have outgrown the single page of documentation they currently are. It might be less overwhelming for users to have a separate top level breakout section of the docs for each of these guides, with individual pages for the different topics (e.g. what we do for "Install SecureDrop"). What do you think @pierwill?

pierwill:
@redshiftzero Okay that all makes sense. I'll do a more structural revision. Is there anything we know for sure should be broken off to another doc or moved? What would you like to do with this branch? Shall I add commits or work on a new one?

pierwill:
This is ready for review. @redshiftzero @dachary 

I've edited the "responsibilities" section and restructured the doc after that. The structure is a bit tangled, so this doc will need more substantial revision later, but I think this is a good fix for now.

docs/admin.rst
```diff
@@ -3,47 +3,52 @@ Admin Guide
...
 .. include:: ./includes/provide-feedback.txt
 
-You (the admin) should have your own username and passphrase, plus
-two-factor authentication through either the FreeOTP app
-on your smartphone or a YubiKey.
+The SecureDrop architecture contains multiple machines and hardened servers.
+While we have automated many of the installation and maintenance tasks, a
+skilled Linux admin is required to responsibly run the system.
+
+This section outlines the tasks the admin is responsible for in order to
```
redshiftzero: Wait, shouldn't this paragraph be under the "Responsibilities" heading?

docs/admin.rst
```diff
@@ -3,47 +3,52 @@ Admin Guide
...
-admin and some manual intervention is required to responsibly run the system.
+Maintaining Credentials
+~~~~~~~~~~~~~~~~~~~~~~~
 
-This section outlines the tasks the admin is responsible for in order to
-ensure that the SecureDrop server continues to be a safe place for sources to
-talk to journalists.
+The admin should have her own username, passphrase, and two-factor
+authentication method (via smartphone app or YubiKey). Admins are also
+responsible for tracking user credentials and encouraging best practices. (See
```
redshiftzero: What does it mean by "tracking user credentials"? (I thought for a second it might be interpreted as "keep a record of all user's passwords", which we don't want)

docs/admin.rst
```diff
@@ -3,47 +3,52 @@ Admin Guide
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-You should maintain awareness of SecureDrop updates and take any required
+The admin should be aware of all SecureDrop updates and take any required
 manual action if requested in the `SecureDrop Release Blog`_. We recommend
 subscribing to the `SecureDrop RSS Feed`_ to stay apprised of new updates.
 
 Most often, the SecureDrop server will automatically update via apt. However,
-occasionally you will need to run the Ansible playbooks. We will inform you in
+occasionally you will need to run Ansible playbooks. We will inform you in
```
redshiftzero: ah er oops, this is old. Bonus points for replacing "run Ansible playbooks" with "run ``securedrop-admin install``" (double backticks for fixed-width)

docs/admin.rst
pierwill: managing?

pierwill:
Pushed requested changes.

PR #3546, opened 2018-06-20
Disable do-release-upgrade OSSEC notification

### LINKED ISSUE #3205: Suppress OSSEC alerts asking SecureDrop administrators to upgrade to Xenial
Per the original report in #1530, administrators may under some circumstances receive OSSEC alerts asking them to upgrade their SecureDrop servers to Ubuntu Xenial. Doing so would be *a very bad idea* as Ubuntu Xenial is not in fact supported yet (see #3204). We should therefore investigate whether administrators are still receiving such alerts under any circumstances.

Tasks:
- [x] Verify whether these alerts are still sent
- [ ] If appropriate, suppress them

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #3205

 * Currently admins are getting weekly alerts encouraging them to upgrade to Xenial, but upgrading to Xenial actually breaks SecureDrop. This addition to the `postinst` scripts of the securedrop OSSEC packages will suppress that notification.

## Testing

These instructions assume that you are testing this in staging VMs.

#### Make sure you're getting OSSEC emails

First, since the alert does not appear in vagrant VMs, and it actually only appears once per week, we should address both of those things as setup.

```
git checkout develop
make build-debs
```

Now make sure you have prod like creds and modify your `staging.yml` group vars to enable postfix (i.e. apply [this diff](https://github.com/freedomofpress/securedrop/pull/2803#discussion_r170720578)):

```
vagrant up /staging/
```

Now ensure OSSEC alerts are flowing.

#### Make the `do-release-upgrade` alert happen, and frequently

On each machine, SSH in and in `/etc/update-manager/release-upgrades` replace `Prompt=never` with `Prompt=lts` (this is what is set on hardware). 

Also delete the flag file at `/var/lib/ubuntu-release-upgrader/release-upgrade-available` if it exists.

Finally for easy testing, in `/etc/crontab` run the weekly job every 5 mins: 

```
*/5 * * * * root test -x /usr/sbin/anacron || ( cd / && run-parts --report /etc/cron.weekly)
```

At this point you should be getting the alert about `do-release-upgrade` to your OSSEC email address every 5 minutes. Wait and verify this is the case.

Note: I only ever get the alert from mon on both hardware and prod VMs. However I think it is best to apply this change to both hosts to ensure this alert doesn't pop up in the future for anyone. 

### Test the fix

```
git checkout  resolve-do-release-notification
make build-debs
vagrant provision /staging/
```

- [x] Verify on each server that `/etc/update-manager/release-upgrades` now has `Prompt=never`

At this point, the alert will continue to fire, because the weekly job is running `/usr/lib/ubuntu-release-upgrader/release-upgrade-motd`, and the logic there is to only update the "do we need to do an upgrade" status every 24 hours. This is tracked via the flag file that we deleted earlier, so you need to do one of the following:

1. Wait 24+ hours to wait for the alert to go away 
2. Edit `/usr/lib/ubuntu-release-upgrader/release-upgrade-motd` to expire the flag file more rapidly OR
3. Delete the stamp file at `/var/lib/ubuntu-release-upgrader/release-upgrade-available` so the check (the actual checking for a new release is done in `/usr/lib/ubuntu-release-upgrader/check-new-release`) is performed again (this is what I did)

Once one of these three things is done:

- [x] Verify the alert no longer appears

## Deployment

Will be automatically deployed when SD OSSEC packages update

### DIFF
--- a/install_files/securedrop-ossec-agent/DEBIAN/postinst
+++ b/install_files/securedrop-ossec-agent/DEBIAN/postinst
@@ -18,6 +18,8 @@ set -e
 
 case "$1" in
     configure)
+        # Disable do-release-upgrade notification
+        sed -i 's/Prompt=.*/Prompt=never/' /etc/update-manager/release-upgrades
         service ossec restart
     ;;
 
--- a/install_files/securedrop-ossec-agent/DEBIAN/postinst
+++ b/install_files/securedrop-ossec-agent/DEBIAN/postinst
@@ -18,8 +18,6 @@ set -e
 
 case "$1" in
     configure)
-        # Disable do-release-upgrade notification
-        sed -i 's/Prompt=.*/Prompt=never/' /etc/update-manager/release-upgrades
         service ossec restart
     ;;
 
--- a/install_files/securedrop-config/DEBIAN/postinst
+++ b/install_files/securedrop-config/DEBIAN/postinst
@@ -4,6 +4,10 @@
 set -e
 set -x
 
+disable_upgrade_prompt() {
+    # Disable do-release-upgrade notification
+    sed -i 's/Prompt=.*/Prompt=never/' /etc/update-manager/release-upgrades || true
+}
 remove_2fa_tty_req() {
     # The goal here is to remove legacy 2FA req on TTY logins
     # Lets prevent this from bombing out the install though if it fails
@@ -23,6 +27,7 @@ case "$1" in
         sed -i 's/deb\.torproject\.org\/torproject\.org/tor-apt.freedom.press/g' "$apt_security_list"
     fi
     remove_2fa_tty_req
+    disable_upgrade_prompt
     ;;
 
     abort-upgrade|abort-remove|abort-deconfigure)
--- a/install_files/securedrop-ossec-server/DEBIAN/postinst
+++ b/install_files/securedrop-ossec-server/DEBIAN/postinst
@@ -29,6 +29,9 @@ case "$1" in
         chown root:${GROUP} ${OSSEC_HOME}/rules/local_rules.xml
         chmod 440 ${OSSEC_HOME}/rules/local_rules.xml
 
+        # Disable do-release-upgrade notification
+        sed -i 's/Prompt=.*/Prompt=never/' /etc/update-manager/release-upgrades
+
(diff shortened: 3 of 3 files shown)


redshiftzero:
Apologies for the novella in the testing plan here, but it's a little annoying to test this so I erred on the side of more detailed. Let me know if anything is confusing

msheiny:
:heart: the detailed test plan .. stepping thru now

install_files/securedrop-ossec-agent/DEBIAN/postinst
```diff
@@ -18,6 +18,8 @@ set -e
 
 case "$1" in
     configure)
+        # Disable do-release-upgrade notification
+        sed -i 's/Prompt=.*/Prompt=never/' /etc/update-manager/release-upgrades
```
msheiny: Just a quick observation.. i think it would make more sense to drop this in `securedrop-config` since that package already goes on both servers. thoughts @redshiftzero ?

redshiftzero: Ahhh, yep - that is probably better

msheiny: :+1:  k i can make that change locally and test against it and commit when all is well

redshiftzero: :+1:

msheiny:
on a side note.. you dont have to wait for cron to fire... if you execute the same cron command as root with the same arguments.. just verify no output is shown.. like this:

```bash
$ cd /
$ SHELL=/bin/sh PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin  run-parts --report /etc/cron.weekly
```

redshiftzero:
Thanks for the review! and definitely, looking forward to getting that upgrade testing on VirtualBox 😍

msheiny:
> Apologies for the novella in the testing plan here, but it's a little annoying to test this so I erred on the side of more detailed. Let me know if anything is confusing

Just to be clear.. i super appreciated the full testing plan :)

redshiftzero:
mergin' and backportin'...

PR #3556, opened 2018-06-22
Speed up admin tests in CI

### LINKED ISSUE #3550: The file context in the admin docker build is too dang high!!!
## Description

The `admin` utility has its own set of tests which are conducted under docker. When a developer runs the environment, volume mounting is run and a tox directory will likely be created (`admin/.tox`). This is fine, shouldn't hurt anything BUT that folder is close to 400MB which is problematic because the docker build process tries to add all those files in as context. This can be a very slow operation on both the initialization and the chown process. So lets remove those from the context!!!

## Steps to Reproduce

* cd into admin
* make sure you have a huge `.tox` folder!!
* time DOCKER_BUILD_VERBOSE=true ./bin/dev-shell echo hey

## Expected Behavior

Build elements zoom by the context adding portion and lots of time is spent on apt and pip.

## Actual Behavior

I mean.. it should take a while to build.. especially the apt portion but you'll also see a significant time spent just adding context and chowning files which is annoying for developers.

## Comments

A dockerignore file is needed here.. easy peasy.. PR incoming

msheiny:
## Status

Ready for review

## Description of Changes

Fixes #3550 

* Prevents adding 350MB+ of un-necessary file context to the admin docker build process
* ~Removes sudo permissions~
* Runs the container as regular user
* ~Drops (seemingly) un-necessary ansible + friends dependencies~\
* Utilize docker caching in CircleCI (when there has been no admin folder changes)

Regarding that last point, can you chime in here @dachary .. i couldnt see any reason why the full bootrstrap args were needed here (at least for the current tests implemented) but maybe I am missing something.

I was able to get the full docker container down from around 900MB to around 300MB and speed up build and run time here.

## Testing

```bash
# clean out your current cache
$ docker rmi -f securedrop-admin && docker container prune -f && docker image prune -f
$ git checkout develop
$ `time make -C admin test`
# then jump back to this branch and compare times
$ docker rmi -f securedrop-admin && docker container prune -f && docker image prune -f
$ `time make -C admin test`
```

## Deployment

Only affects CI


dachary:
I like that it runs non-root, excellent idea :-)

The sudo is useful when this same container is used for interactive debuging / development with `bin/dev-shell bash`.

Running `bootstrap.py` instead of cherry-picking what is strictly necessary is to avoid duplication and facilitate maintenance. If someone adds a package in `install_apt_dependencies` in `bootstrap.py`, they will likely be confused when the test fails because it is not added in the Dockerfile as well. It is a tradeoff and maybe it should be reconsidered.

admin/Dockerfile
```diff
@@ -5,16 +5,21 @@ ENV USER_NAME ${USER_NAME:-root}
...
 
 WORKDIR /opt
-COPY bootstrap.py .
-COPY requirements.txt .
-RUN python bootstrap.py -v
+RUN chown -R $USER_NAME /opt
+USER ${USER_NAME}
 ENV VIRTUAL_ENV /opt/.venv
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+COPY requirements.txt .
```
dachary: Why is it necessary?

msheiny: hah i have no answer here... i guess its not

msheiny:
> If someone adds a package in install_apt_dependencies in bootstrap.py, they will likely be confused when the test fails because it is not added in the Dockerfile as well. It is a tradeoff and maybe it should be reconsidered.

Yeah that makes sense..... maybe i can add an argument and separate function to bootstrap for dependencies needed in both?

> The sudo is useful when this same container is used for interactive debuging / development with bin/dev-shell bash.

I could add a new target like `./bin/dev-shell-root` ? My concern about the sudo up rules is that it allows a process running in the container to elevate too easily without explicit consideration by the dev. I dunno, I'm probably being way too overly paranoid.

msheiny:
also note that i should add some kind of docker caching here for circle.. that should knock another couple minutes off builds when there are no changes in CI

heartsucker:
Agree with @dachary. I usually do my development in a container before adding it to the dockerfile. I would like to be able to easily have root without adding too much complexity to the dev env. And the SD dev env is by far the most complex I have ever dealt with, so I'm in favor of minimizing code paths / branches / etc.

msheiny:
so technically this is passing now... but something is up with the circle caching logic... the checksum keeps changing for the admin folder... im a little confused here

msheiny:
ohhh i get it the tar command isnt an accurate check to see what changed.. its scooping up timing information.. okay fixing

msheiny:
ugggggggggggggggggggggghhhhhhhh i saved 2 whole seconds in CI... i suspect theres something weird with tox trying to reinstall the pip files though... that seems problematic but its not a high priority so i'm dropping my timebox shield down.

So yeah don't merge as-is... I'm gonna to work on other items though anddd i'm just gonna close this PR for now :|

PR #3561, opened 2018-06-23
Fix one major omission, three minor issues

### LINKED ISSUE #3558: [QA] Kernel debugging docs clarifications
## Description

Recently performed an [upgrade test of a rolled back hardware instance](https://github.com/freedomofpress/securedrop/issues/3512#issuecomment-399575916), following the new [kernel debugging topic guide](
https://docs.securedrop.org/en/latest/kernel_troubleshooting.html). I'll list them here in order of importance:

* [ ] At the end, after updating the kernel to the latest (i.e. reverting the rollback), **there's no instruction to run update-grub**. If the instructions are followed exactly, the server will continue to boot into the old kernel, until _another_ image is installed, which would trigger update-grub automatically.
* [ ] While editing the boot options manually, neither server showed the "quiet" option. The "linux" line ended in "ro". We should amend the language to "append `single`" rather than removing a different option.
* [ ] Docs recommend `sudo update-grub2`, rather than the more common `sudo update-grub`. Given that "update-grub2" is just a symlink to "update-grub", I suggest we use the more simple `sudo update-grub`. 

Those were the major issues I noticed. If those are resolved, I'm happy with the kernel docs for 0.8.0.

eloquence:
- Add reminder to run `update-grub` (!)
- Change `update-grub2` to `update-grub`
- Do not instruct users to replace `quiet` with `single`, as `quiet`
  may not be in the boot options
- Change `shutdown -r now` to `reboot` and `shutdown -P now` to
  `poweroff` for consistency and readability

Resolves #3558

## Status

Ready for review 

## Checklist

### DIFF
--- a/docs/kernel_troubleshooting.rst
+++ b/docs/kernel_troubleshooting.rst
@@ -37,10 +37,9 @@ shut down the server.
 
 Once you hit a key, you will be able to interact with the menu with the
 up (⬆) and down (⬇) keys. Select “Ubuntu” as shown above, and press “e”
-to edit the boot options. In the line that begins “linux”, replace the
-word “quiet” with “single”. Note that the word "quiet" may be wrapped. Once you
-have performed the replacement, the output on your console should look similar to
-the screenshot below.
+to edit the boot options. In the line that begins with “linux”, add the word
+“single” at the end. When you are done, the output on your console should look
+similar to the screenshot below.
 
 |GRUB in edit mode|
 
@@ -146,7 +145,7 @@ This change still has to be applied to take effect on the next boot:
 
 .. code:: sh
 
-  sudo update-grub2
+  sudo update-grub
 
 Now you can reboot into the old, working kernel.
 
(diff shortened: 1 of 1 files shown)


docs/kernel_troubleshooting.rst
```diff
@@ -243,14 +242,22 @@ by issuing the following command:
...
 it. Verify  that it is working correctly by logging in using your *Admin
 Workstation*. If everything is working as expected, you can make the same change
-to ``/etc/default/grub`` on your *Application Server* as well. You can do so
-from your *Admin Workstation* and reboot the server using the command
-``sudo shutdown -r now``.
+to ``/etc/default/grub`` on your *Application Server* as well. Remember to again
+run the command ``sudo update-grub`` when you are done.
+
+You can make the change on the *Application Server* from your *Admin Workstation*
+and reboot the server using the command ``sudo shutdown -r now``.
```
conorsch: Nit: elsewhere in the guide we use `sudo reboot`, can we use that here, as well, for consistency's sake?

PR #3564, opened 2018-06-25
Login throttling should be per-user

### LINKED ISSUE #3563: Login throttling should be per-user
# Description

This is a denial of service bug that is fortunately behind the journalist interface ATHS. Login throttling is currently not per-user, it's global, i.e. if a single user is blocked from logging in due to login throttling, then all users are. We should fix this in 0.8.0.

# Steps to reproduce

1. Provision dev env
2. Try to login 6 times with incorrect credentials as journalist user. You will start seeing the following in the log output:

```
--------------------------------------------------------------------------------
ERROR in utils [/Users/redshiftzero/Documents/Github/securedrop/securedrop/journalist_app/utils.py:83]:
Login for 'journalist' failed: throttled (6 attempts in last 60 seconds)
--------------------------------------------------------------------------------
```

3. Try to login for the first time as dellsberg

# Expected behavior

Dellsberg can login and is not impacted by the tricksy behavior of the journalist user

# Actual behavior

Dellsberg is denied access to the journalist interface:

```
172.17.0.1 - - [23/May/2018 23:00:13] "POST /login HTTP/1.1" 200 -
172.17.0.1 - - [23/May/2018 23:00:13] "GET /org-logo HTTP/1.1" 302 -
--------------------------------------------------------------------------------
ERROR in utils [/Users/redshiftzero/Documents/Github/securedrop/securedrop/journalist_app/utils.py:83]:
Login for 'dellsberg' failed: throttled (7 attempts in last 60 seconds)
————————————————————————————————————————
```

# Resolution

The fix here is the addition of an additional filter in the ORM query we use to filter by user when computing `attempts_within_period` in `Journalist.throttle_login`. A fix _must_ include a regression test.

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #3563.

 * Adds a regression test for bug in #3563 
 * Resolve #3563 bug by making the login throttling per-user

## Testing

- [ ] Reproduce original bug following STR in #3563
- [ ] Checkout the regression test without the fix (i.e. check out e128b04) and verify it fails
- [ ] Apply the fix in 369b472, the regression test should now pass and logging in as the second user in the STR should now result in success

## Deployment

To be deployed in `securedrop-app-code` deb package tomorrow

### DIFF
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -407,6 +407,7 @@ class Journalist(db.Model):
         login_attempt_period = datetime.datetime.utcnow() - \
             datetime.timedelta(seconds=cls._LOGIN_ATTEMPT_PERIOD)
         attempts_within_period = JournalistLoginAttempt.query.filter(
+            JournalistLoginAttempt.journalist_id == user.id).filter(
             JournalistLoginAttempt.timestamp > login_attempt_period).all()
         if len(attempts_within_period) > cls._MAX_LOGIN_ATTEMPTS_PER_PERIOD:
             raise LoginThrottledException(
--- a/securedrop/tests/test_journalist.py
+++ b/securedrop/tests/test_journalist.py
@@ -188,6 +188,49 @@ def test_login_throttle(journalist_app, test_journo):
         models.LOGIN_HARDENING = False
 
 
+def test_login_throttle_is_not_global(journalist_app, test_journo, test_admin):
+    """The login throttling should be per-user, not global. Global login
+    throttling can prevent all users logging into the application."""
+
+    # Overwrite the default value used during testing
+    # Note that this may break other tests if doing parallel testing
+    models.LOGIN_HARDENING = True
+    try:
+        with journalist_app.test_client() as app:
+            for _ in range(Journalist._MAX_LOGIN_ATTEMPTS_PER_PERIOD):
+                resp = app.post(
+                    url_for('main.login'),
+                    data=dict(username=test_journo['username'],
+                              password='invalid',
+                              token='invalid'))
+                assert resp.status_code == 200
+                text = resp.data.decode('utf-8')
+                assert "Login failed" in text
+
+            resp = app.post(
+                url_for('main.login'),
+                data=dict(username=test_journo['username'],
+                          password='invalid',
+                          token='invalid'))
+            assert resp.status_code == 200
+            text = resp.data.decode('utf-8')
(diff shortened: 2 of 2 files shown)


redshiftzero:
(once merged this should get backported into 0.8.0)

redshiftzero:
Very good point, marking as wip for a bit

redshiftzero:
Actually, upon investigation, this is actually the case on `develop` also - this is due to the fact we check if the username is valid prior to checking the number of prior login attempts.

redshiftzero:
And since in the `journalist_login_attempt` table we have `journalist_id` as a foreign key to the journalist table, we can't actually store or track login attempts for non-journalists users... so this is something that we should file as a followup and address in 0.9.0

redshiftzero:
(here's a regression test for that other bug for later)

```
diff --git a/securedrop/tests/test_journalist.py b/securedrop/tests/test_journalist.py
index 13426872..098ddb8b 100644
--- a/securedrop/tests/test_journalist.py
+++ b/securedrop/tests/test_journalist.py
@@ -231,6 +231,38 @@ def test_login_throttle_is_not_global(journalist_app, test_journo, test_admin):
         models.LOGIN_HARDENING = False
 
 
+def test_login_throttle_works_for_invalid_usernames(journalist_app,
+                                                    test_journo, test_admin):
+
+    # Overwrite the default value used during testing
+    # Note that this may break other tests if doing parallel testing
+    models.LOGIN_HARDENING = True
+    try:
+        with journalist_app.test_client() as app:
+            for _ in range(Journalist._MAX_LOGIN_ATTEMPTS_PER_PERIOD):
+                resp = app.post(
+                    url_for('main.login'),
+                    data=dict(username='invalid',
+                              password='invalid',
+                              token='invalid'))
+                assert resp.status_code == 200
+                text = resp.data.decode('utf-8')
+                assert "Login failed" in text
+
+            resp = app.post(
+                url_for('main.login'),
+                data=dict(username='invalid',
+                          password='invalid',
+                          token='invalid'))
+            text = resp.data.decode('utf-8')
+            assert ("Please wait at least {} seconds".format(
+                Journalist._LOGIN_ATTEMPT_PERIOD) in text)
+
+    finally:
+        models.LOGIN_HARDENING = False
+
+
 def test_login_invalid_credentials(journalist_app, test_journo):
     with journalist_app.test_client() as app:
         resp = app.post(url_for('main.login'),
```

PR #3568, opened 2018-06-25
Implement logic to prevent checkout of branch when named the same as our signed tag

### LINKED ISSUE #3567: Implement logic to guard against checkout of branch in the case of a duplicate branch and tag
## Description

We distribute code to workstations via checkout of a signed git tag. If there is a branch with the same as the tag, then the tag will verify, but when we actually `git checkout <tag name>` then the branch, not the tag, will be checked out. 

### Short term

As a guard, implement logic to verify there is no duplicate tag and branch prior to checkout. 

### Long term

We should stop distributing code in this way and implement https://github.com/freedomofpress/securedrop/issues/3502

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #3567.

 * Ensures that securedrop-admin update will not check out branch if it is named the same as our signed tag

## Testing

- [x] The regression test alone should fail: 55fd98e (`make -C admin test`)
- [x] After the fix in f86f380 and the test updates in d74ffb8 are applied, all tests should pass

### Manual testing

 - checkout the latest prod tag locally (`git checkout 0.7.0`)
 - make a branch of the same name (`git checkout -b 0.7.0`)
 - go to this branch (`git checkout logic-checkout-branch-tag`) and run `securedrop-admin update` manually - you should not successfully update to latest due to the duplicate tag/branch. Instead, you should see `Signature verification failed` (this should trigger the user to contact their admin/support)

## Deployment

Unfortunately, we should recommend users do the update on the command line, and they should apply the same logic in this PR manually

### DIFF
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -686,7 +686,18 @@ def update(args):
         if RELEASE_KEY in gpg_lines[1] and \
                 sig_result.count(good_sig_text) == 1 and \
                 bad_sig_text not in sig_result:
-            sdlog.info("Signature verification successful.")
+            # Finally, we check that there is no branch of the same name
+            # prior to reporting success.
+            cmd = ['git', 'show-ref', '--heads', '--verify', '--quiet',
+                   'refs/heads/{}'.format(latest_tag)]
+            try:
+                # We expect this to produce a non-zero exit code, which
+                # will produce a subprocess.CalledProcessError
+                subprocess.check_output(cmd)
+                sdlog.info("Signature verification failed.")
+                return 1
+            except subprocess.CalledProcessError:
+                sdlog.info("Signature verification successful.")
         else:  # If anything else happens, fail and exit 1
             sdlog.info("Signature verification failed.")
             return 1
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -688,16 +688,21 @@ def update(args):
                 bad_sig_text not in sig_result:
             # Finally, we check that there is no branch of the same name
             # prior to reporting success.
-            cmd = ['git', 'show-ref', '--heads', '--verify', '--quiet',
+            cmd = ['git', 'show-ref', '--heads', '--verify',
                    'refs/heads/{}'.format(latest_tag)]
             try:
                 # We expect this to produce a non-zero exit code, which
                 # will produce a subprocess.CalledProcessError
-                subprocess.check_output(cmd)
+                subprocess.check_output(cmd, stderr=subprocess.STDOUT)
                 sdlog.info("Signature verification failed.")
                 return 1
-            except subprocess.CalledProcessError:
-                sdlog.info("Signature verification successful.")
+            except subprocess.CalledProcessError, e:
+                if 'not a valid ref' in e.output:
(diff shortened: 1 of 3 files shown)


admin/securedrop_admin/__init__.py
```diff
@@ -686,7 +686,18 @@ def update(args):
...
+            # prior to reporting success.
+            cmd = ['git', 'show-ref', '--heads', '--verify', '--quiet',
+                   'refs/heads/{}'.format(latest_tag)]
+            try:
+                # We expect this to produce a non-zero exit code, which
+                # will produce a subprocess.CalledProcessError
+                subprocess.check_output(cmd)
+                sdlog.info("Signature verification failed.")
+                return 1
+            except subprocess.CalledProcessError:
```
emkll: Are there any other circumstances/potential edge cases where this `CalledProcessError` exceptions can be raised? Would it be superfluous to verify the output text like in the admin integration tests (e.g. checking for the substring `not a valid ref`)?

redshiftzero: Hmm, I can't think of a reason off the top of my head, but for the sake of defense in depth, let me take a look at adding this (should be quick)

redshiftzero: this logic is added in e9998c5, good suggestion! also added 8fd25b0 for tests to cover the `else` case in the diff

conorsch:
## Testing

The test plan as described validated the behavior presented in the PR. Details below, lifted from the template in the submission text.

- [x] The regression test alone should fail: 55fd98e (`make -C admin test`)
- [x] After the fix in f86f380 and the test updates in d74ffb8 are applied, all tests should pass

### Manual testing

 - :heavy_check_mark:  checkout the latest prod tag locally (`git checkout 0.7.0`)
 - :heavy_check_mark: make a branch of the same name (`git checkout -b 0.7.0`)
 - [x] go to this branch (`git checkout logic-checkout-branch-tag`) and run `securedrop-admin update` manually - you should not successfully update to latest due to the duplicate tag/branch. Instead, you should see `Signature verification failed` (this should trigger the user to contact their admin/support)

## Addenda

Not request changes because these are minor comments that should not block release. Listing them here to kickstart discussion.

1. Consider adding a call to `git branch -D <tag_name>` _prior_ to the checkout command. That'd still ensure that only the tag is used.
2. Consider removing `--all` from the `git fetch` command. Do we really want to retrieve info from multiple remote endpoints? We may be better off avoiding mistakes if we simply fetch from the origin. Either way, the hardcoded sig verification will guard against both mistakes and nefarious actions. 

Looking good, let's get it in.

redshiftzero:
Regarding your addenda:

> Consider adding a call to git branch -D <tag_name> prior to the checkout command. That'd still ensure that only the tag is used.

this is a good point, and we could implement this logic in a later release (preferred: cut our losses and start using apt packages as soon as possible). The one downside here is that we wouldn't know that a MitM attempt had been performed.

> Consider removing --all from the git fetch command. Do we really want to retrieve info from multiple remote endpoints? We may be better off avoiding mistakes if we simply fetch from the origin. 

Yep... that's a good point.

That said, not implementing these changes right now as this logic has been tested by multiple people, and release is <24 hours away, it would be high-risk to make further modifications.

Issue #3176, opened 2018-03-19
Switch to using git-lfs for image files


msheiny:
# Feature request

## Description

Our git repository is rather large because of a lot of early binaries that were shoved into source control. We've stopped doing that but we continue to shove image file formats (`*.png` for example) into our docs which frequently get refreshed. 

To prevent size creep, lets start using [git-lfs](https://git-lfs.github.com/) for all image files. This change will require all users who are contributing doc edits and/or building docs to have `git-lfs` installed first. If one is just making code changes they do not need `git-lfs` installed ... 

## User Stories

As a securedrop developer, I want to try to decrease the growing size of the securedrop repo and follow best practices by not putting non-text files in source control directly.

redshiftzero:
I'm in favor of this, using LFS is not that much of an additional burden given that we already ask documentation writers to use both git and sphinx

eloquence:
This was done (for screenshots etc.) as part of the docs repo migration (#5435).

PR #3575, opened 2018-06-26
docs: update screenshots for 0.8.0

dachary:
## Status

Ready for review

## Description of Changes

https://docs.securedrop.org/en/latest/development/i18n.html#the-day-of-the-release

* git checkout release/0.8
* make -C securedrop update-user-guides
* commit
* cherry-pick to release/develop

## Testing

Verify visually that each image is changed in a sensible way.

## Deployment

N/A


heartsucker:
Do all of these need to get merged? This is a lot of binary data to stuff into `git`. if two screen shots are "pretty damn close" I'd day that's good enough to leave out as long as the intent is still clear.

msheiny:
@heartsucker what about ... enabling git-lfs here a la #3176 . it adds an extra burden on the developers but otherwise requires no change to all other tooling (like CI, Tails, etc.)

redshiftzero:
@heartsucker makes a good point, how about we: inspect these screenshots and see which ones have more changes beyond just the version string change, and just update those? Medium term (i.e. after 0.8.0 is released) we can look into #3176 as @msheiny suggests

redshiftzero:
ok so there are five that have UI changes:

```
journalist-clicks_on_source_and_selects_documents.png
journalist-col_flagged.png
journalist-col_has_no_key.png
journalist-composes_reply.png
journalist-index_javascript.png
```

proposal: we remove the other images and squash into one commit

redshiftzero:
actually, in the interests of time (i.e. avoiding another CI run), i'm going to merge this in as-is so that we can get these changes in for 0.8.0, will file followup issue after release for the best practice here going forward

PR #3417, opened 2018-05-15
Docs ssh instructions output

kstohr:
## Status

Ready for review 

## Description of Changes
Changes proposed in this pull request:
https://github.com/freedomofpress/securedrop/issues/3348

## Testing
make docs-lint - passed

PR #3591, opened 2018-06-28
Clarify docs re output of ssh hostname commands

### LINKED ISSUE #3348: SSH instructions to verify connection need style tweaks
# Feature request
Re-style verification commands here: https://docs.securedrop.org/en/release-0.6/servers.html#set-up-ssh-keys to show that `app` and `mon` are server output and not part of the command to run.

## Description
Change this block:

> $ ssh \<username\>@\<App IP address\> hostname
> app
> $ ssh \<username\>@\<Monitor IP address\> hostname
> mon

...to make it look less like the app and mon bits are part of the command. Existing styleguide might not be sufficient to address this since it already seems to comply with it: https://docs.securedrop.org/en/release-0.6/development/documentation_guidelines.html#style-guide. Also maybe add in a bit as to what the goal of the command is. e.g. avoid admins assuming this is doing anything other checking that you can get a hostname from a remote server.

## User Research Evidence

While installing SecureDrop, I very literally made this mistake and was confused as to why `ssh <username>@<App IP address> hostname app` was failing. I had assumed the goal of the command was to create a kind of alias for the IP address with a three-letter hostname rather than simply returning the hostname from the remote machine. 

## User Stories

As a SecureDrop Admin, I would like to decrease my risk of making mistakes when setting up a system with high stakes involved.

edenemmanuel:
## Status

Ready for review

## Description of Changes

Fixes #3348. Supersedes #3417.

The codeblock at the end of the page "Set Up the Servers" is ambiguous to some people. We can make this less ambiguous by adding a brief line about what sort of output the admin should expect.

Long term solution may involve updating the style guide around how we do code blocks, or providing more clarification earlier in the document.

Special thanks to @kstohr for doing the vast majority of the work to fix #3348. We really appreciate it! 😁 

## Checklist

### DIFF
--- a/docs/servers.rst
+++ b/docs/servers.rst
@@ -353,4 +353,6 @@ the below commands. You should not be prompted for a passphrase
     $ ssh <username>@<Monitor IP address> hostname
     mon
 
-If you have successfully connected to the app via SSH, the terminal output will be name of the app to which you have connected ('app' or 'mon') as shown above.
+If you have successfully connected to the server via SSH, the terminal
+output will be name of the server to which you have connected ('app'
+or 'mon') as shown above.
--- a/docs/servers.rst
+++ b/docs/servers.rst
@@ -352,3 +352,5 @@ the below commands. You should not be prompted for a passphrase
     app
     $ ssh <username>@<Monitor IP address> hostname
     mon
+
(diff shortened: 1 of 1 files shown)


redshiftzero:
thanks @edenemmanuel !

PR #3617, opened 2018-06-29
Smarter rebase on-top of github target logic

### LINKED ISSUE #2942: CI jobs should rebase on target branch, not always develop
## Description

We rebase on `develop` for all CI jobs, which is groovy for PRs into `develop`, but we don't want to do this for PRs into `master` or any `release*` branch. Currently we have logic in `securedrop/devops/scripts/rebase-develop.sh` that looks like this:

```
# If current branch is not master or doesnt start with release...
if [[ "$CIRCLE_BRANCH" != "master"  && "$CIRCLE_BRANCH" != release*  ]]; then
```

which will prevent the rebase step from running if we're on the `master` or `release*` branches. But since the `$CIRCLE_BRANCH` env var stores the current branch, not the branch we're trying to merge into, the rebasing logic will still run for PRs into `master` or `release*`. 

See also: [feature](https://discuss.circleci.com/t/only-build-pull-requests-targetting-specific-branch/6082) [requests](https://discuss.circleci.com/t/create-a-circle-target-branch-envar/10022) over at Circle CI

## User Stories

As a developer, I don't want to get conditioned into looking at a red build and saying "oh _that_ red build is ok"

msheiny:
Ready for review

Fixes #2942

* Modifies rebase script to pull from github API instead of hard-coding `develop`

Start by populating the following env vars:
* `CIRCLE_PULL_REQUEST`
* `CIRCLE_BRANCH`  - Keep in mind that this looks like a full URL ala `https://github.com/freedomofpress/securedrop/pull/3556`. You might ask, why didnt @msheiny use `CIRCLE_PR_NUMBER` to just grab the number like @redshiftzero suggested!!? Well unfortunately that env var only applies to [forked PRs](https://circleci.com/docs/2.0/env-vars/).

Okay now run .. `devops/scripts/rebase-ci.sh`. Make sure stuff doesnt explode. If you have dangling git changes you should get errors about a rebase.

Only affects CI. I did take note to look at the parameters from a forked PR on the github API and I'm convinced it'll work fine *famous last words*

Caveat -- One particular bit I am concerned about is Github API limits... its 60 per hour per host. Now I think we are safe here because we are executing on multiple builders which should change IP addresses.. but something to watch out for.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,8 +8,8 @@ jobs:
       - run: apt-get install -y make
       - checkout
       - run:
-          name: Rebase on-top of latest develop
-          command: ./devops/scripts/rebase-develop.sh
+          name: Rebase on-top of github target
+          command: ./devops/scripts/rebase-ci.sh
       - setup_remote_docker
 
       - run: TAG=${CIRCLE_SHA1} make ci-lint-image
@@ -27,8 +27,8 @@ jobs:
       - run: apt-get install -y make
       - checkout
       - run:
-          name: Rebase on-top of latest develop
-          command: ./devops/scripts/rebase-develop.sh
+          name: Rebase on-top of github target
+          command: ./devops/scripts/rebase-ci.sh
       - setup_remote_docker
 
       - restore_cache:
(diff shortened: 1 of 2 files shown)


msheiny:
@conorsch i rebased and removed that dangling branch :+1: .. waiting for tests to pass :crossed_fingers:

PR #3620, opened 2018-07-03
Docs sg3100 firewall

### LINKED ISSUE #3520: Recommended SG-4860 firewall has reached end-of-sale
The [officially recommended](https://docs.securedrop.org/en/stable/hardware.html#network-firewall) pfSense 4860 firewall has reached end-of-sale. Netgate recommends SG-3100 and XG-7100 as alternatives.

We should evaluate alternatives and update our recommendation. See #2605 for additional background on alternatives like the SG-3100.

b-meson:
## Status

Ready for review

## Description of Changes

Fixes #3520. 

## Testing

A reviewer should factory reset a SG-3100 and follow the 3NIC configuration steps. After setting the firewall rules, they should be able to run through an end-to-end SecureDrop installation and be able to reach the Source interface, the Journalist Interface, and receive OSSEC alerts. 

## Notes

I noticed at the very bottom of the `Disable DHCP server on Firewall` screenshot, there is an _Admin Gateway IP_ which we could probably crop out since it may be confusing. Additionally, this PR does **not** touch the `.xml` files which I think we should add as its own issue / commit.


docs/network_firewall.rst
```diff
@@ -1,29 +1,26 @@
...
+IOS etc. This guide is based on pfSense, and assumes your firewall hardware has
+at least three interfaces: WAN, LAN, and OPT1. For hardware, you can build
+your own network firewall (not covered in this guide) and `install pfSense 
+<https://doc.pfsense.org/index.php/Installing_pfSense>`__ on it. For most
+installations, we recommend buying a dedicated firewall appliance with
+pfSense pre-installed, such as the one recommended in the
+:ref:`Hardware Guide <hardware_guide>`.
+
+We currently recommend the `pfSense SG-3100
+<https://store.netgate.com/SG-3100.aspx>`__, which has 6 interfaces: WAN,
```
emkll: nit: It actually has 3 interfaces/NICs (WAN, OPT, and LAN) but it does indeed have 6 ports (WAN, OPT and LAN1 through LAN4)

docs/network_firewall.rst
```diff
@@ -99,6 +99,38 @@ chosen:
 -  Monitor Gateway: ``10.20.3.1``
 -  Monitor Server (OPT2) : ``10.20.3.2``
 
+3 NIC Example (SG-3100)
```
emkll: It it possible to make the numbering 5. b. for `3 NIC Example` and 5. a. for the `4 NIC Example`? Admins might think they must complete both.

docs/network_firewall.rst
```diff
@@ -99,6 +99,38 @@ chosen:
...
+'''''''''''''''''''''''
+
+As described earlier, the SG-3100 has an internal switch on the LAN interface
+which means we can place the *Application Server* and *Admin Workstation* on
+the same subnet and gateway. This example assumes you place the
+*Admin Workstation* on LAN1 and the *Application Server* on LAN2.
```
emkll: I don't think it matters which LAN ports the admin workstation and application server are connected to (as long as it's one of the four LAN ports) since it's a simple switch.

docs/network_firewall.rst
```diff
@@ -99,6 +99,38 @@ chosen:
...
+-  Admin Gateway: ``10.20.2.1``
+-  Admin Workstation: ``10.20.2.3``
+
+.. raw:: html
+
+   <!-- -->
+
+-  Application Subnet: ``10.20.2.0/24``
+-  Application Gateway: ``10.20.2.1``
+-  Application Server (OPT1): ``10.20.2.2``
```
emkll: Application server is now on LAN, not OPT1

docs/network_firewall.rst
```diff
@@ -99,6 +99,38 @@ chosen:
...
+
+.. raw:: html
+
+   <!-- -->
+
+-  Monitor Subnet: ``10.20.3.0/24``
+-  Monitor Gateway: ``10.20.3.1``
+-  Monitor Server (OPT2) : ``10.20.3.2``
```
emkll: Monitor server in this case in on OPT1, not OPT2

docs/network_firewall.rst
```diff
@@ -216,12 +261,20 @@ Setup Wizard
...
-   (``10.20.1.1``) and the subnet mask (``/24``) of the *Admin Subnet*.
-   Click **Next**.
+#. **4 NIC Example:**
+   For "Configure LAN Interface", use the IP address of the *Admin Gateway*
+   (``10.20.1.1``) and the subnet mask (``/24``) of the *Admin Subnet*. Click
+   **Next**.
 
    |Configure LAN Interface|
 
+#. **3 NIC Example (SG-3100):**
```
emkll: Same as above, would it be possible to make them 5.a. and 5.b.?

docs/network_firewall.rst
```diff
@@ -371,8 +439,28 @@ In order to use the firewall to isolate the *Application Server* and the *Monito
...
+The easiest way to set up your firewall rules is to look at the screenshots of
+a correctly configured firewall and edit the interfaces, aliases, and firewall
+rules on your firewall to match them.
+
+
+
+4 NIC Example
+~~~~~~~~~~~~~
+
+If you are using a firewall that has a dedicated port for each component of
```
emkll: nit: dedicated NIC or dedicated interface

docs/network_firewall.rst
```diff
@@ -486,6 +562,114 @@ message "The changes have been applied successfully". Once you've set up the
...
+
+.. warning:: Be sure not to delete the Anti-Lockout Rule on the LAN interface.
+    Deleting this rule will lock you out of the pfSense WebGUI.
+
+Add or remove rules until they match the following screenshots by clicking **Add**
+to add a rule.
+
+**LAN interface:**
+
+|3 NIC Firewall LAN Rules|
```
emkll: nit: you can move the admin workstation TCP */* rule under the deny from LAN to OPT1 which should prevent the workstation from reaching the server over any port that 22.

docs/network_firewall.rst
```diff
@@ -371,8 +439,28 @@ In order to use the firewall to isolate the *Application Server* and the *Monito
 Server* from each other, we need to connect them to separate interfaces, and then set
 up firewall rules that allow them to communicate.
 
+Setup the Firewall Rules
```
eloquence: Nit: setup->set up (only one word when used as a noun)

docs/network_firewall.rst
```diff
@@ -486,6 +562,114 @@ message "The changes have been applied successfully". Once you've set up the
...
+~~~~~~~~~~~~~~~~~~~~~~~
+
+The below guide assumes you are using a 3 NIC firewall such as the SG-3100.
+While the SG-3100 has an integrated switch, you may need to add a switch to
+the LAN interface if you use a different firewall.
+
+Set Up LAN
+''''''''''
+
+Although we setup the LAN interface during the Setup Wizard we need to make
```
eloquence: setup->set up

docs/network_firewall.rst
b-meson: It does not. I was slightly confused about which port to put the App and Admin servers so I figured I would be more verbose just to help future admins.

b-meson:
@emkll instead of this: 

`Disable the anti-lockout rule (system -> advanced -> admin access -> disable webConfigurator anti-lockout rule), ensuring that a rule to allow communication from workstationto the router's web interface is in place.`

Can we instead modify the anti-lockout rule to only allow routing from the `admin workstation`?

docs/network_firewall.rst
b-meson: I attempted to do here but it actually looks much worse and is less clear. I did make that change in the other suggested section.

docs/network_firewall.rst
b-meson: I just removed the line `This example assumes you place the *Admin Workstation* on LAN1 and the *Application Server* on LAN2.` As you said, it shouldn't matter but i think having too much language could be confusing.

b-meson:
I pushed an update and rebased on top of `develop`. We are still missing the revisions to the "anti lockout rule" issue that @emkll raised. I am also waiting to hear back on my question:

> I noticed at the very bottom of the "Disable DHCP Server on Firewall" screenshot, there is an Admin Gateway IP which we could probably crop out since it may be confusing.

emkll:
> I noticed at the very bottom of the "Disable DHCP Server on Firewall" screenshot, there is an Admin Gateway IP which we could probably crop out since it may be confusing.

Good catch ! Since it's fairly straightforward to crop out, I suggest we do so.

> Can we instead modify the anti-lockout rule to only allow routing from the admin workstation?

The anti-lockout rule automatically created by pfsense and cannot be modified. I suggest we caution the admin that the rules will need to be applied with care and in order as to not get locked out from the admin UI (as it would require a console reset to regain access to the interface).

b-meson:
I added another commit 2f988cf8712eb78c10294ddc8d06560344ee37db and rebased it on top of develop. @emkll ready for another round of reviews.

docs/network_firewall.rst
```diff
@@ -486,6 +563,128 @@ message "The changes have been applied successfully". Once you've set up the
...
+
+-  IPv4 Configuration Type: Static IPv4
+-  IPv4 Address: ``10.20.2.1`` (Application Gateway IP)
+
+Make sure that the CIDR routing prefix is correct (``/24``). Leave everything else
+as the default. **Save** and **Apply Changes**.
+
+|3 NIC LAN Interface|
+
+Disable Anti-Lockout Rule
```
emkll: nit: At line 668, we refer to the (old) anti-lockout rule. Can we either move this anti-lockout related section down there or move the warning here to avoid confusion?

Issue #789, opened 2014-12-06
When sources that are "flagged" generate a gpg key they don't move to the top


micahflee:
The list of sources is sorted chronologically by latest activity.

If there's a source that's flagged for response and then they login and therefore generate a gpg key, SecureDrop 0.3 doesn't consider that new activity. This means that the journalist is in no way notified that they can now write a response to that source if they want to.

SecureDrop 0.2.1 did consider it new activity, and that source rose to the top of the list.

This is an extremely minor issue though, since in most cases now the "flag for response" feature will never be used anyway, unless there's an entropy DoS.

PR #3637, opened 2018-07-13
source: do not set last_updated on key generation

### LINKED ISSUE #3635: journalist notification incorrectly claims submissions were received
## Description

A journalist notification mail is sent claiming a submission was received in the past 24h. But after checking the journalist interface, there is nothing. 

## Steps to Reproduce

* drain the source of entropy
* as a new source submit a document and check that the source is flagged because no new key was created
* wait 24h and see a journalist notification is mailed because there is a new submission
* login back as the source using the same passphrase and check that the source is no longer flagged
* wait 24h and see a journalist notification is mailed although there is no new submission

## Expected Behavior

* no journalist notification is sent when a source key generation is delayed

## Actual Behavior

* a journalist notification is sent when a source key generation is delayed

dachary:
## Status

Ready for review

## Description of Changes

Fixes #3635 

The last_updated field is updated every time a submission is
sent. Also updating it when the source key is generated has a
different meaning and creates a problem.

A: there is enough entropy and the key is created when a message or
   document is submitted. The last_updated field is set twice, with
   the same value, give or take a few seconds.

B: there is not enough entropy and the key is created long after the
   initial document or message was submitted, when the source
   reconnects using the same passphrase. The last_updated is set and
   that confuses the manage.py::were_there_submissions_today which
   assumes the field is only set when a submission is received.

Since no code logic relies on the last_updated field being set when
the key is generated, the async_genkey is modified to no longer set
the last_updated field.

## Testing

TBD

## Deployment

Existing deployments will no longer set the `last_updated` field when a key is created. This has no consequence as this field is not exposed to admins and users.

### DIFF
--- a/securedrop/source_app/utils.py
+++ b/securedrop/source_app/utils.py
@@ -86,7 +86,6 @@ def async_genkey(crypto_util_, db_uri, filesystem_id, codename):
     try:
         source = session.query(Source).filter(
             Source.filesystem_id == filesystem_id).one()
-        source.last_updated = datetime.utcnow()
         session.commit()
     except Exception as e:
         logging.getLogger(__name__).error(


dachary:
This is not going to be that simple: it would [introduce a regression](https://github.com/freedomofpress/securedrop/issues/789)

PR #3688, opened 2018-08-02
Modify template to allow journalist to download own replies

### LINKED ISSUE #3674: [reply refactor] Allow journalists to download replies from journalist interface
After #3673 is implemented, we should allow journalists to download replies from the journalist interface UI. Note that for long-running SecureDrop instances, there will be old replies encrypted only to the source key that should be unavailable for download.

Epic: #3097

emkll:
## Status

Ready for review

## Description of Changes

Fixes #3674 

Journalists can now download their replies encrypted to the submission
key. They are automatically marked as read in the interface and have a
reply icon to avoid confusion with sources messages and unread count.

## Testing

* Submit document/message as source
* Log into journalist account
* Reply to source
* Journalist must be able to download reply
* The reply in the conversation thread must not be confusing
## Deployment

This will be deployed in the SecureDrop app code deb package.

Since this is a journalist-facing change, we should inform users of this in the release notes.

### DIFF
--- a/securedrop/journalist_templates/col.html
+++ b/securedrop/journalist_templates/col.html
@@ -31,28 +31,21 @@
 
       <ul id="submissions" class="plain submissions">
         {% for doc in source.collection %}
-          <li class="submission {% if not doc.downloaded %}unread{% endif %}">
-            {% if not doc.filename.endswith('reply.gpg') %}
-              {% if not doc.downloaded %}
+          <li class="submission {% if not doc.downloaded and not doc.filename.endswith('reply.gpg') %}unread{% endif %}">
+              {% if doc.filename.endswith('reply.gpg') %}
+                <input type="checkbox" name="doc_names_selected" value="{{ doc.filename }}" class="doc-check">
+                <span title="Reply" class="icon"><i class="fa fa-reply pull right"></i></span>
+              {% elif not doc.downloaded %}
                 <input type="checkbox" name="doc_names_selected" value="{{ doc.filename }}" class="doc-check unread-cb">
                 <span title="Unread" class="icon"><i class="fa fa-envelope"></i></span>
               {% else %}
-                <input type="checkbox" name="doc_names_selected" value="{{ doc.filename }}" class="doc-check">
-                <span class="icon"><i class="fa fa-envelope-open"></i></span>
+                <input type="checkbox" name="doc_names_selected" value="{{ doc.filename }}" class="doc-check unread-cb">
+                <span title="Read" class="icon"><i class="fa fa-envelope-open"></i></span>
               {% endif %}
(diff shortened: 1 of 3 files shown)


securedrop/journalist_templates/col.html
```diff
@@ -31,28 +31,21 @@
...
-            {% if doc.filename.endswith('reply.gpg') %}
-              <span class="file reply"><span class="filename">{{ doc.filename }}</span></span>
-              <span class="info"><span title="{{ doc.size }} bytes">{{ doc.size|filesizeformat() }}</span></span>
-            {% else %}
-              <a class="file {% if not doc.downloaded %}unread{% else %}read{% endif %}" href="{{ url_for('col.download_single_submission', filesystem_id=filesystem_id, fn=doc.filename) }}">
-                <i class="fa fa-download"></i> <span class="filename">{{ doc.filename }}</span>
-              </a>
-              <span class="info"><span title="{{ doc.size }} bytes">{{ doc.size|filesizeformat() }}</span></span>
-            {% endif %}
+            <a class="file {% if not doc.downloaded and not doc.filename.endswith('reply.gpg') %}unread{% else %}read{% endif %}" href="{{ url_for('col.download_single_submission', filesystem_id=filesystem_id, fn=doc.filename) }}">
```
redshiftzero: Since we're using `col.download_single_submission` here, we'll need to modify the logic in that view function a bit due to the following `try`, `except` block that's in there now: 

```
        try:
            Submission.query.filter(
                Submission.filename == fn).one().downloaded = True
            db.session.commit()
        except NoResultFound as e:
            current_app.logger.error(
                "Could not mark " + fn + " as downloaded: %s" % (e,))
```

The issue here is that it's expecting a `Submission` object so that it can mark the file as downloaded. We only want to enter this `try`, `except` block if we are indeed downloading a `Submission` (not a `Reply`), otherwise we won't find the object, and we'll log an error (which will produce an OSSEC alert).

Also, since we're now using this `download_single_submission` logic to download both submissions and replies, I think for clarity we should rename it `download_single_file` or similar. What do you think?

securedrop/tests/test_journalist.py
```diff
@@ -1344,6 +1344,26 @@ def test_passphrase_migration_on_reset(journalist_app):
...
+                       fn=submissions[0].filename)
+    reply_url = url_for('col.download_single_file',
+                        filesystem_id=replies[0].source.filesystem_id,
+                        fn=replies[0].filename)
+
+    with journalist_app.test_client() as app:
+        resp = app.get(subm_url)
+        assert resp.status_code == 302
+        resp = app.get(reply_url)
+        assert resp.status_code == 302
```
redshiftzero: :100:!

PR #3690, opened 2018-08-03
Preserve conversation history only for journalists

### LINKED ISSUE #3675: [reply refactor] Preserve conversation history between sources and journalists
In order for journalists to view the conversation history between sources and journalists, when a source deletes the journalist reply from the _source inbox_, it should not delete the reply from the _journalist's_ inbox. This is a major UX win for journalists while preserving the current behavior of minimizing data available if a source codename is compromised. 

We could do this by adding a column to the replies table that `allow_source_read` (`default=True`) or something similar that we can filter by when showing replies to sources. When the source deletes the reply from their inbox, we set this to `False`. 

Epic: #3097

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #3675

 - Adds new column to track which replies should appear in the source inbox
 - Preserves (encrypted) conversation history for journalists (from the source's perspective, no changes will occur)

## Testing

1. Submit document as source
2. Submit reply as journalist
3. Verify that the reply is displayed without a checkmark, e.g. see last row here:

screen shot 2018-08-03 at 11 23 24 am

4. Delete reply as source
5. Verify that the reply still exists on the journalist side, can be downloaded and decrypted, and a checkmark now appears next to it, e.g. see last row here:

screen shot 2018-08-03 at 11 24 22 am

## Deployment

Deployed in securedrop-app-code package, significant change on journalist side so needs to be described in pre-release announcement and release notes

### DIFF
--- a/docs/journalist.rst
+++ b/docs/journalist.rst
@@ -75,7 +75,7 @@ Journalist Interface <yubikey_setup>`.)
 
 Reset Passphrase or Two-factor Authentication Token
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-If necessary journalists may reset their user passphrase or two-factor authentication token in their user profile. To navigate to your user profile, log in to the *Journalist Interface* and click on the link in the upper right of the screen where it says **Logged on as <your user name>.** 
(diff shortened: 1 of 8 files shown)


redshiftzero:
This is rebased on the changes in #3688 and ready for review. Btw, the UX changes are in cfa090d3c4607f79e288b697a89923a8f3c183e7 and I've updated the screenshots in the test plan so you can see what the proposed flow will look like going forward, any feedback / suggestions welcome on that.

redshiftzero:
Thanks for the review @emkll! and great point on the docs, added a commit for that

PR #3644, opened 2018-07-17
Converts Vagrantfile to Molecule for staging

conorsch:
## Status

Ready for review.

## Description of Changes

Progress towards #3208. So far, only the libvirt scenario has been ported. We also need support for VirtualBox, which should be straightforward. I propose sprinkling a wee wrapper into the mix, so that we can have a `make staging` that will work on Virtualbox, Libvirt, or Qubes (#3624), as appropriate.

There's also a small change in here to satisfy #3629, which should work equally well in prod and staging environments, with the same logic applicable to staging regardless of developer OS. 

The major win will come when we integrate the config tests into Molecule, as well as begin testing against Xenial as an additional platform. Those changes are not presented here, but will open up shortly.

### Libvirt
**Requires a libvirt-enabled workstation.** 

1. Run `make staging`. 
2. Confirm it defaults to the libvirt scenario, based on stdout.
3. Confirm zero errors during provisioning.
4. Log into app: `molecule login -s libvirt-staging -h app-staging`
5. View source interface in Tor Browser, confirm it's working well.

### Qubes
**Requires a Qubes workstation.**

1. Run `make staging`.
2. Confirm it defaults to the Qubes scenario, based on stdout.
3. When machines reboot, run `qvm-start sd-app sd-mon` in `dom0` or an Admin-API-enabled dev VM, so bring the standalone machines back up.
4. Confirm final reboot task does not error out, as it did in #3629.

### VirtualBox
**Recommended testing machine: a Mac laptop.** You can also try setting `VAGRANT_DEFAULT_PROVIDER=virtualbox` beforehand, then following the steps below.
1. Run `make staging`.
2. Confirm it defaults to the VirtualBox scenario, based on stdout. 
3. Confirm zero errors during provisioning.
4. Log into app: `molecule login -s virtualbox-staging -h app-staging`
5. View source interface in Tor Browser, confirm it's working well.

## Deployment
Yes, the reboot tweak should be tested against prod VMs in Tails. All other changes are dev-env-only, so far.

### If you made changes to the server application code:

- [ ] Linting (`make ci-lint`) and tests (`make -C securedrop test`) pass in the development container

### If you made changes to `securedrop-admin`:

- [ ] Linting and tests (`make -C admin test`) pass in the admin development container

### If you made changes to the system configuration:

- [ ] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

### If you made non-trivial code changes:

- [ ] I have written a test plan and validated it for this PR

[...]

PR #3703, opened 2018-08-08
Integrates testinfra tests with Molecule "verify" action

### LINKED ISSUE #3205: Suppress OSSEC alerts asking SecureDrop administrators to upgrade to Xenial

conorsch:
## Status

Ready for review.

## Description of Changes

Completes and therefore closes #3205. Follows up on #3644.

Integrates the testinfra config test suite with Molecule, so that it runs as a `verify` action. Deleted the custom `testinfra/test.py` wrapper script that managed the connection info for remote hosts.

Mostly this is a move operation, shoving the testinfra tests into the `molecule/` dir. Some minor tweaks were necessary inside the tests given the path change, but the tests still validate the same functionality they previously did. 

### CI
1. Make sure CI passes!
2. Inspect a successful CI run and check whether the test info is still downloadable as artifacts.

### Local env

1. Run `molecule test -s libvirt-staging` (requires Linux) and confirm all tests passing.

## Deployment
Dev-env only. Reorganizes tests, but preserves the content of what they're testing. 


molecule/aws/molecule.yml
```diff
@@ -29,20 +30,25 @@ provisioner:
   config_options:
     defaults:
       callback_whitelist: "profile_tasks, timer"
-  inventory:
-    links:
-      group_vars: ../../install_files/ansible-base/group_vars
-      host_vars: ../../install_files/ansible-base/host_vars
   env:
-    ANSIBLE_ROLES_PATH: ../../install_files/ansible-base/roles
+    ANSIBLE_CONFIG: ../../install_files/ansible-base/ansible.cfg
```
msheiny: Noiceeeeee :taco:

devops/scripts/combine-junit-test-results.sh
```diff
@@ -0,0 +1,12 @@
...
+set -e
+
+
+# Directory should already exist, from the CI Molecule scenario,
+# but let's be certain.
+mkdir -p junit
+
+# Combine tests for export as artifacts. There are at least two XML files,
+# one from the application tests, and another from the config tests.
+./molecule/testinfra/combine-junit.py ./*results.xml > ./junit/junit.xml
```
msheiny: I'm not sure about moving all the tests into a shared molecule/testinfra folder.. its not really a scenario (like the other folders) so this could be confusing later on. :man_shrugging: i guess its okay. just talking out loud here

molecule/testinfra/staging/app/test_tor_hidden_services.py
```diff
@@ -2,6 +2,7 @@
 import re
 
 
+testinfra_hosts = ["app-staging"]
```
msheiny: I would ideally love this to be an environment variable. I foresee the host names changing in the future but this isnt a deal-breaker for me. Just noting it for future implementations.

molecule/testinfra/staging/conftest.py
```diff
@@ -0,0 +1,91 @@
...
+    app_ip = lookup_aws_private_address(molecule_info, 'app-staging')
+    mon_ip = lookup_aws_private_address(molecule_info, 'mon-staging')
+
+    os.environ['APP_IP'] = app_ip
+    os.environ['MON_IP'] = mon_ip
+
+    # Make SSH calls more resilient, as we're operating against remote hosts,
+    # and running from CI. We've observed flakey connections in CI at times.
+    os.environ['ANSIBLE_SSH_RETRIES'] = '5'
+    ssh_args = [
```
msheiny: Love this approach to making the connection more stable :heart:

devops/scripts/combine-junit-test-results.sh
conorsch: @msheiny Totally agree, perhaps `devops/scripts/` would be the best location?

molecule/testinfra/staging/conftest.py
conorsch: It took me far too long to come up with that simple solution—been working reliably since. :ok_hand:

molecule/testinfra/staging/app/test_tor_hidden_services.py
conorsch: That's true, for example if/when we go back to testing e.g. `app-prod` regularly, we'll want to override here. Appending to the list seems sane. Willing to defer, since we don't have a clear picture of the future testing scenarios just yet.

devops/scripts/combine-junit-test-results.sh
msheiny: > `devops/scripts/`

yeah i could get down with that :)

conorsch:
Made the requested changes, @msheiny; ready for re-review!

PR #3714, opened 2018-08-16
Set 30 day retention period for apache logs

### LINKED ISSUE #304: Change default apache logrotate config
Change default apache logrotate to rotate every day and only retain 1 day. This should give enough time for any personnel to get on site for organizations that may opt not to enable ssh.
Since sensitive information could be in error messages the logs should be securely deleted.

starchy:
## Status

## Description of Changes

Fixes #304.

## Testing

## Deployment

2. New installs.

### DIFF
--- a/install_files/ansible-base/roles/app/files/apache2-logrotate.conf
+++ b/install_files/ansible-base/roles/app/files/apache2-logrotate.conf
@@ -1,0 +1,10 @@
+/var/log/apache2/*.log {
+  daily
+  missingok
+  rotate 30
+  create 640 root adm
+  sharedscripts
+  postrotate
+    /usr/bin/killall -HUP apache2
+  endscript
+}
--- a/install_files/ansible-base/roles/app/tasks/install_and_harden_apache.yml
+++ b/install_files/ansible-base/roles/app/tasks/install_and_harden_apache.yml
@@ -116,3 +116,12 @@
     - restart apache2
   tags:
     - apache
+
+- name: Retain no more than one month of apache logs
+  copy:
+    src: apache2-logrotate.conf
+    dest: /etc/logrotate.d/apache2
+    owner: root
+    mode: "0644"
+  tags:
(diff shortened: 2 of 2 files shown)


zenmonkeykstop:
Hi @starchy - this PR will probably need a rebase before review and merge. Do you have cycles available for it, or would you be OK with somebody else picking it up?

starchy:
I'm going on leave with limited internet access soon so it's probably best
if someone else grabs it. Thanks for asking!

eloquence:
Hi @starchy, sorry for the long silence on this PR. We're still interested in landing this change. There are no concerns in principle about reducing the retention from the perspective of tech support for newsrooms, though we may want to give organizations a heads up that this is coming.

Because this requires a run of the Ansible playbook to be applied to all configurations, one option we're considering is to fold this change into the release which will add support for the next LTS release of Ubuntu, where a playbook run will be required. This would ensure that we avoid too much configuration drift between new installs and existing ones. So just a heads up that this PR may still open quite a bit longer (the current LTS reaches end-of-life on April 30, 2021).

cfm:
It looks like we missed this pull request as part of the Xenial-to-Focal migration.  @zenmonkeykstop (CC: @eloquence), is this something you'd be interested in pulling into either v2.6.1 or v2.7.0?  Re: configuration drift, we could tack on a `postinst` fixup.

cfm:
Believe it or not...we'll pick this (back!) up for SecureDrop 2.12 towards compatibility with Ubuntu Noble.  We'll take on the task of adding an upgrade (whether `postinst`-based or otherwise) path.

legoktm:
As explained in https://github.com/freedomofpress/securedrop/issues/304#issuecomment-2557702119, in the meantime the default apache log retention has changed to 2 weeks, so we no longer need to reduce it. Thanks @starchy for the PR and sorry we weren't able to get to it in time.

PR #3719, opened 2018-08-17
Attempts to fix #2067 Pagination

### LINKED ISSUE #2067: Pagination
## Description

Currently, we have no pagination on results on either the SecureDrop source or journalist interface. For example, this means that if a news organization isn't great about deleting sources then they will have a list of potentially hundreds of sources appear whenever they load the journalist interface. This is bad for usability as it is overwhelming and it increases page load times, which are already slow due to Tor. 

## User Stories

As a journalist user, I want to see only the N most recent sources on the index so that I'm not overwhelmed with hundreds of sources.

zyphlar:
## Status

Ready for review

## Description of Changes

Fixes #2067 

- Paginates the list of sources in journalist interface
- Reduces SD logo redrawing during pageload (better UX)
- Handles paginated sorting of starred items
- Doesn't lose your spot when you star or unstar a source

## Testing

- Increase num_sources in create-dev-data.py to at least 15 (if you want to hand-edit `journalist_app/main.py` to set `per_page=50` to something like 1) or several hundred (if you don't want to edit that line, and have the time/cpu)
- Login as a journalist
- Note the pagination buttons along the bottom
- Click a button or visit the ?page=2 page
- Note items are paginated as expected
- Visit the ?page=0 and ?page=999 pages
- Note that a 404 error is returned since there are not that many pages
- Visit Page 2 and Star a source. Note that the starred item disappears (moves to the top of the list on page 1) but your view remains on Page 2.
- Stay on Page 2 and use some of the mass-edit buttons at the top. Note that they do not keep your view on Page 2, which *may or may not be expected behavior.*
- Use some of the filtering, mass-select, download, star, or delete functions at the top. Note that many of them only apply to sources visible on the current page, which *may or may not be expected behavior.*

## Deployment

I intentionally stayed away from adding dependencies or configuration requirements.


zyphlar:
Screenshot:

screenshot from 2018-08-16 01-56-04

Questions/thoughts since I'm not a Python/Flask pro:

- Is the `all()` function needed at the end of queries even if they're paginated?
- Is request.args.get() secure when type=int is set, even if it's passed straight into paginate()?
- I set per_page=50 rather than create a config.py constant. Perhaps others want to add this to let users customize pagination.
- The filtering and mass-selecting functions seem to operate only on sources visible in the table.
  - If users expect to be able to filter (search) the whole table, or select/download/star ALL sources, then @heartsucker's comment about JS/AJAX pagination definitely applies. However at that point this should basically be an API endpoint so that excess data is not sent over the wire.
  - Alternatively, these batch functions could be rewritten to operate on the entire dataset, not just what's visible.
- Should other mass functions redirect to the referring page instead of the index? I fixed the problem when clicking individual stars, but maybe other actions should keep the pagination as well (redirect to referrer, not index.)
- I didn't write tests, but if pointed in the right direction I probably could.

redshiftzero:
Great work here @zyphlar! 

> Is the all() function needed at the end of queries even if they're paginated?

Nope it's not necessary. For queries that are not paginated (`flask_sqlalchemy.BaseQuery`), as you correctly point out we have to use the `all()` method to get a list of (for us here) the `models.Source` objects. The reason why is because the `items` attribute on the paginated object (`flask_sqlalchemy.Pagination`) is already a list of the `models.Source` objects. 

> Is request.args.get() secure when type=int is set, even if it's passed straight into paginate()?

Since you're setting this variable to an integer, this is indeed safe to pass straight to `paginate()` (cc @emkll). 

> I set per_page=50 rather than create a config.py constant. Perhaps others want to add this to let users customize pagination.

Good call, I agree that we should defer customization of `per_page` in a subsequent pull request (aside: I think a default number per page of 15 is a good balance). 

> The filtering and mass-selecting functions seem to operate only on sources visible in the table.
> If users expect to be able to filter (search) the whole table, or select/download/star ALL sources, 
> then @heartsucker's comment about JS/AJAX pagination definitely applies. However at that point 
> this should basically be an API endpoint so that excess data is not sent over the wire.
> Alternatively, these batch functions could be rewritten to operate on the entire dataset, not just 
> what's visible.

Great questions. Currently, it is unclear from the user's perspective whether or not the "Select All" applies to only the current page or not. To investigate this as a user, I might click the next page, to see if they are also selected, only to discover that they are not. When I click back, my original checkmarks are gone, leaving me confused. So the right behavior from the user's perspective should be:

1. On the first page, when the user clicks "Select All", all checkboxes on the first (current) page are highlighted.
2. When a user browses to the second page of the index and navigates back to the first page, the checkboxes persist. 
2. User clicks the "Select All" on the second page, all checkboxes on the second (current) page are highlighted.
3. All action buttons along the top apply to all selected items, which in this example would be all items on the first and second pages.

What do you think? Feel free to disagree if you think this is not intuitive from the user's perspective. I was unsure myself, but then I looked at how other similar UIs handle this, and Gmail does what is described above. If you agree, we should modify how these batch functions are operating. 

When it comes to filtering/searching, the common UI pattern is that this is a search of the entire list of sources (see again e.g. Gmail). As such, we should do this logic in Flask instead of JavaScript (as we are currently doing). 

> Should other mass functions redirect to the referring page instead of the index? I fixed the problem when clicking individual stars, but maybe other actions should keep the pagination as well (redirect to referrer, not index.)

Totally, what you have done with the individual stars is intuitive :+1:. I think this should be the behavior for other actions 

> I didn't write tests, but if pointed in the right direction I probably could.

Here's where we have unit tests of the piece of SecureDrop you are modifying, additional tests for pagination would go there: https://github.com/freedomofpress/securedrop/blob/develop/securedrop/tests/test_journalist.py

zyphlar:
@redshiftzero thanks! Sorry for the lag.

So looking at how this is written, it seems designed to degrade gracefully without JS and use very non-JS web client-server architecture -- when you star something, `col.py#process` goes to the `utils.py#col_*` functions which either return a downloadable file or a 301 redirect to the index page.

The typical way to write what you're asking for would be to make the app 100% JS (React, Angular, Ember, lots of JQuery), handle pagination automagically behind the scenes, and also manage bulk operations automagically. Lots of AJAX/REST requests and buttons/forms that don't go anywhere except for more JS and AJAX/REST. I get the feeling you don't really want to go that route, for network, cpu, and browser-compatibility reasons.

The no-JS way to do this is sadly to click each selectbox individually and lose all your selectboxes each time you browse to a new page, even other paginated pages.

So I get the idea you want a half-JS graceful way of doing this... which means I somehow need to remember which selectboxes are checked across pageloads and then submit the full list via form submission when a bulk action button is clicked.

But I kinda want to check in with you before I go down that route because my first thought of using LocalStorage/Cookies and using JS to shove/mangle form submissions on the fly seems pretty janky and likely to cause some kind of headache (or shaking-of-heads) at some point. Any thoughts on that? I've done JS form mangling before, but sometimes it's more straightforward and less heartache to just design around it.

eloquence:
Hey @zyphlar, sorry for letting this sit. Touching base if you're still interested in working on this, with additional guidance. This PR is pretty old so a rebase may require a fair amount of attention to detail at this point (see conflicts).

zyphlar:
Hi! I'm happy to, the main issue was that the backend needs to support the
pages without bringing in too many third party libraries, so it's a design
decision which direction we should go. Can you take a peek and let me know
which direction you'd like to go, either faking it in the frontend with no
backend changes, or rearchitecting most controllers to support real
pagination, or bringing in outside dependencies to do it the way we'd
normally go it online?

eloquence:
Thanks @zyphlar; yes, we're generally very conservative when it comes to adding dependencies, and are trying to keep the front-end as lean as possible, as well. I've made a note to bring this up for further technical discussion with the team next week, if no maintainer is able to respond to you before then.

zyphlar:
The big question with pagination is, are we doing it to save on pageload
time, server cpu, client resources, or user clicks? Paginating purely
server-side lets you avoid querying and transmitting 10000 records each
pageload, whereas paginating purely client-side lets you rapidly develop a
feature that looks nice even though it can hog server/client resources. So
we need to know what the impetus behind the feature request is.

There's newer hybrid pagination facebook/google style where you lazy-load
pages as desired which can be the best of both worlds, but it also requires
the most amount of code on client and server side to support. So far most
of the development reminds me of lowest common denominator 2002-style
programming, which would indicate server-side pagination, which is fine and
just expands the scope of this task somewhat beyond my immediate comfort
zone; it'll take tinkering and I'd want approval before I hack up all your
controllers.

eloquence:
hey @zyphlar, thanks for your patience and sorry again for letting this sit for so long. After discussing it a bit more with the team, I'm going to close this PR and the underlying issue. We're not going to implement pagination in the JI for now:

- We generally want to discourage accumulating large numbers of sources on the server for security reasons.
- In terms of usability improvements for journalists, our primary focus is the new [SecureDrop Workstation](https://github.com/freedomofpress/securedrop-workstation), which ships with a [graphical client](https://github.com/freedomofpress/securedrop-client) that keeps a local copy of the source database in sync with the server, making access to existing sources very fast.
- If we did add this to the JI, we'd want to make improvements that would potentially balloon the scope of this PR.

Thanks for picking this up from our issues backlog in the first place, and hope you still consider contributing to SecureDrop in future. Would be great to see you [on Gitter](https://gitter.im/freedomofpress/securedrop)!

Issue #3734, opened 2018-08-24
Tooltip missing on read replies icon in single source page on journalist interface


zenmonkeykstop:
## Description

Read replies are downloadable in 0.9.0 from the single source page as another document type with a checkmark icon. The other document type icons have a tooltip on mouseover that describes their type, but the checkmark icon doesn't.

## Steps to Reproduce
In a 0.9.0 install:

- As a source, submit a message and remember your code phrase
- As a journalist, reply to the message
- As a source, delete the reply
- As a journalist, view the single source page and mouse over the checkmark icon for the reply document

## Expected Behavior
A tooltip is displayed indicating what the icon means

## Actual Behavior
Unlike the other icons in the document listing, nothing is displayed.

redshiftzero:
Just noticed while looking at this that some tooltips are not localized. Let's fix this tooltip in 0.9.0 and then add the localization in a followup PR, which we may or may not include in 0.9.0 (cc @kushaldas since that will mean string changes)

redshiftzero:
closed by #3735 #3738 

PR #3736 is up for localization followup

PR #3736, opened 2018-08-24
Journalist Interface: Localize tooltips

redshiftzero:
## Status

Ready for review

## Description of Changes

Noticed during resolution of #3734 

 - Localize tooltips

## Testing

Make sure all tooltips modified in this PR are displayed in the Journalist Interface

## Deployment

Requires string changes

Since we have _no_ string changes as part of 0.9.0 so far we _could_ just merge this into `develop` and fix in 0.10.0. That said, we have a few days until string freeze, so now would be a good time to resolve if we want to get it into 0.9.0, which would be nice if possible. What do you think @kushaldas? 


heartsucker:
I say we merge it in just to get it done because `gettext` falls back to English on string lookup failure, which means the behavior in the UI will remain unchanged.

PR #3737, opened 2018-08-24
Apply database migration on backup restore

### LINKED ISSUE #3732: SecureDrop backups from previous versions don't work if database migration has occurred
## Description

Backup restore (https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/restore/files/restore.py) script does not apply database migrations, and as such breaks the application upon backup restore. Manual workaround that appears to have no side-effects is to run `sudo dpkg-reconfigure securedrop-app-code` on the app server.

## Steps to Reproduce

1. Install SecureDrop 0.8.0
2. `securedrop-admin backup`
3. Upgrade to 0.9.0 (or higher)
4. `securedrop-admin restore`
5. Observe source and journalist interface return error 500

## Expected Behavior

The application should be operational.

## Actual Behavior

The source and journalist interfaces return error 500s.

## Comments

Running `sudo dpkg-reconfigure securedrop-app-code` calls the postinst script which will apply migration. Based on my testing, this seems to work reliably.

emkll:
## Status

Ready for review

## Description of Changes

Fixes #3732 .

Apply migrations as part of backup restore script.

## Testing

1. Install SecureDrop version < 0.9.0
2. `./securedrop-admin backup`
3. Upgrade SecureDrop to 0.9.0-rc1
4. Grab filename from backup tarball that was written to `install_files/ansible-base/`
5. `./securedrop-admin restore <filename>`
6. Go to the source interface and submit a document
7. Validate that the document can be retrieved on the journalist interface
8. Inspect the database to ensure the migrations have been successful, and that the following tables have UUIDs (`sqlite3 /var/lib/securedrop/db.sqlite`):
    a. replies
    b. submissions
    c. sources
    d. journalists

## Deployment

This will be used when restoring from backup. Code will be updated via SecureDrop Updater GUI.

### DIFF
--- a/install_files/ansible-base/roles/restore/files/restore.py
+++ b/install_files/ansible-base/roles/restore/files/restore.py
@@ -45,6 +45,8 @@ def main():
     # If the process exits with a non-zero return code, raises an exception.
     subprocess.check_call(['service', 'apache2', 'restart'])
     subprocess.check_call(['service', 'tor', 'reload'])
+    # Apply database migrations (if backed-up version < version to restore)
+    subprocess.check_call(['dpkg-reconfigure', 'securedrop-app-code'])
 
 
 if __name__ == "__main__":


zenmonkeykstop:
Do the test steps need a `securedrop-admin restore <file>` as well?

emkll:
Woops, updated the test plan with restore instructions. Thanks @zenmonkeykstop !

install_files/ansible-base/roles/restore/files/restore.py
```diff
@@ -45,6 +45,8 @@ def main():
     # If the process exits with a non-zero return code, raises an exception.
     subprocess.check_call(['service', 'apache2', 'restart'])
     subprocess.check_call(['service', 'tor', 'reload'])
+    # Apply database migrations (if backed-up version < version to restore)
```
heartsucker: Technically this comment is wrong. It will always attempt to migrate to `head` which will throw an exception if we restore a v3 version DB to a Debian package with v2 migration code (meaning one migration is missing) `alembic` will explode and yell at us that it can't do the thing.

PR #3742, opened 2018-08-25
remove task that removes tmux from bashrc

### LINKED ISSUE #3609: Remove deprecated ansible task that cleans up bashrc tmux additions
## Description

From `install_files/ansible-base/roles/common/tasks/create_users.yml`:

```yaml
# TODO: Backwards-compatibility. Previously, the SecureDrop bashrc additions
# for forcing a terminal multiplexer during interactive login sessions were
# added to ~/.bashrc for each admin user account. It's cleaner to add the
# config lines to /etc/profile.d, so all accounts get them by default.
# Post 0.4, we can remove this task, which is here to clean up the local
# bashrc files so that the additions are only sourced once.
- name: Clean up local bashrc config for admin accounts.
  lineinfile:
    dest: /home/{{ item }}/.bashrc
    line: '. /etc/bashrc.securedrop_additions'
    state: absent
  with_items: "{{ ssh_users }}"
  tags:
    - users
    - environment
```

heartsucker:
## Status

## Description of Changes

Fixes #3609

(see title)

## Testing

`make staging`

## Deployment

This might need a double check from the ops team, but this change has been here long enough that there shouldn't be any instances left that retain this line.

### DIFF
--- a/install_files/ansible-base/roles/common/tasks/create_users.yml
+++ b/install_files/ansible-base/roles/common/tasks/create_users.yml
@@ -33,19 +33,3 @@
   tags:
     - users
     - environment
-
-# Backwards-compatibility. Previously, the SecureDrop bashrc additions
-# for forcing a terminal multiplexer during interactive login sessions were
-# added to ~/.bashrc for each admin user account. It's cleaner to add the
-# config lines to /etc/profile.d, so all accounts get them by default.
-# Post 0.4, we can remove this task, which is here to clean up the local
-# bashrc files so that the additions are only sourced once.
-- name: Clean up local bashrc config for admin accounts.
-  lineinfile:
-    dest: /home/{{ item }}/.bashrc
-    line: '. /etc/bashrc.securedrop_additions'
-    state: absent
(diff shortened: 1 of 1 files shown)


install_files/ansible-base/roles/common/tasks/create_users.yml
```diff
@@ -33,19 +33,3 @@
   tags:
     - users
     - environment
-
-# Backwards-compatibility. Previously, the SecureDrop bashrc additions
```
redshiftzero: Questioning the comment here: we know from the 0.11.0 upgrades that we have instances that were installed prior to 0.4 that have not run the playbooks since. In which case, we do want to keep this task in for the next time they run `securedrop-admin install`. This PR should remove the part of the comment saying that we can remove this task post 0.4, as it's assuming that admins run the playbooks each release. 

tl;dr modifications to system config implemented only in Ansible will need to stay in the playbooks basically forever due to the long tail of instances not running playbooks in possibly years

heartsucker:
Closing in favor of #4018

PR #3760, opened 2018-08-31
Add 0.8.0 to 0.9.0 upgrade docs

### LINKED ISSUE #3751: [0.9.0] Add 0.8.0 to 0.9.0 upgrade guide to docs
## Description

As with [0.7.0 to 0.8.0](https://docs.securedrop.org/en/release-0.8/upgrade/0.7.x_to_0.8.html) and previous releases, we should add an upgrade guide so that folks looking for instructions, e.g., regarding the kernel update, can find them in a predictable location, not just in the blog.

eloquence:
Resolves #3751. Alongside some fairly boilerplate text, this also
includes a note regarding the new apt persistence behavior in
Tails 3.9.

## Status

Ready for review

## Checklist

### DIFF
--- a/docs/upgrade/0.8.0_to_0.9.0.rst
+++ b/docs/upgrade/0.8.0_to_0.9.0.rst
@@ -1,0 +1,133 @@
+Upgrade from 0.8.0 to 0.9.0
+===========================
+
+Updating the Tails Workstations
+-------------------------------
+
+We recommend that you update all Tails drives to version 3.9, which was released
+concurrently with SecureDrop 0.9.0 on September 5, 2018. Follow the
+graphical prompts on your workstations to perform this upgrade.
+
+For the *Journalist Workstations* and the *Admin Workstation*, the graphical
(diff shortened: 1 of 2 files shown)


docs/upgrade/0.8.0_to_0.9.0.rst
```diff
@@ -0,0 +1,123 @@
...
+Finally, run the following command:
+
+.. code:: sh
+
+  ./securedrop-admin setup
+
+Please note that this only updates the SecureDrop code on the workstation.
+Tails upgrades still have to be performed separately.
+
+New Dialogs in Tails 3.9
```
redshiftzero: thought: we should add a note about the database migrations: "Part of the release includes a significant database migration that can take some time, especially if you are not using an SSD or if you have a large number of sources and submissions. You should not be concerned as this is normal behavior during the upgrade, and the source/journalist interfaces will become available again once the upgrade has completed."

PR #3776, opened 2018-09-04
[0.9.0] Add 0.8.0 to 0.9.0 upgrade docs

emkll:
## Status

Ready for review

## Description of Changes

Backports #3760 into the 0.9 release branch

## Testing

Changes should be the same as #3760 

## Deployment

Docs only

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -86,6 +86,7 @@ anonymous sources.
    :name: upgradetoc
    :maxdepth: 2
 
+   upgrade/0.8.0_to_0.9.0.rst
    upgrade/0.7.x_to_0.8.rst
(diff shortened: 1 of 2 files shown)


emkll:
closed in favor of #3778

redshiftzero:
For future reference / for the interested observer, the reason why CI was failing here, is because another branch called `backport-3760-to-0.9` was pushed with identical changes to those here in `docs-backport-3760-to-0.9`. In a situation where there are two branches with the same changes (i.e. the same HEAD commit hash), then CircleCI will only run the build once, and my understanding is that we have to either open a new PR or add a new commit to get CI to rebuild

PR #3815, opened 2018-09-11
Add detail about membership of the docker group in Ubuntu/Debian quickstart.

ntoll:
## Status

Ready for review.

## Description of Changes

A minor documentation edit so docker related installation / setup has complete steps. Will save someone, somewhere 15 minutes of Google-ing.

## Testing

Build the docs and visit `/development/setup_development.html` to check.

## Deployment

None that I know of.

### DIFF
--- a/docs/development/setup_development.rst
+++ b/docs/development/setup_development.rst
@@ -36,14 +36,11 @@ be installed via the official documentation links:
 * `Docker CE for Ubuntu`_
 * `Docker CE for Debian`_
 
+Make sure to follow the `Post-installation steps for Linux`_, as well.
+
 .. _`Docker CE for Ubuntu`: https://docs.docker.com/install/linux/docker-ce/ubuntu/
 .. _`Docker CE for Debian`: https://docs.docker.com/install/linux/docker-ce/debian/
-
-.. warning::
-
-    The official documentation currently misses the detail that the user needs
-    to be in the ``docker`` group. This is easily achieved with
-    ``sudo adduser $USER docker`` and restarting your session.
+.. _`Post-installation steps for Linux`: https://docs.docker.com/install/linux/linux-postinstall/
 
 macOS
 ~~~~~
--- a/docs/development/setup_development.rst
+++ b/docs/development/setup_development.rst
@@ -39,6 +39,11 @@ be installed via the official documentation links:
 .. _`Docker CE for Ubuntu`: https://docs.docker.com/install/linux/docker-ce/ubuntu/
 .. _`Docker CE for Debian`: https://docs.docker.com/install/linux/docker-ce/debian/
 
+.. warning::
+
+    The official documentation currently misses the detail that the user needs
+    to be in the ``docker`` group. This is easily achieved with
+    ``sudo adduser $USER docker`` and restarting your session.
 
 macOS
 ~~~~~


docs/development/setup_development.rst
```diff
@@ -39,6 +39,11 @@ be installed via the official documentation links:
 .. _`Docker CE for Ubuntu`: https://docs.docker.com/install/linux/docker-ce/ubuntu/
 .. _`Docker CE for Debian`: https://docs.docker.com/install/linux/docker-ce/debian/
 
+.. warning::
+
+    The official documentation currently misses the detail that the user needs
```
heartsucker: By "official documentation" do you mean SecureDrop's or Docker's? Additionally, all commands can be `sudo`'d which is what Docker recommends since having a user in the `docker` group is actually a security hole as it effectively means commands can be run as root.

ntoll: Aha... I meant dockers docs. 

I didn't see the instructions for using `sudo` in their docs and the example commands I read in both docker and SecureDrop suggest (to me) that `sudo` wasn't expected, hence this change. However, please disregard this change given the security concerns of being in the `docker` group. In any case, would be good to know what advice to give (I was just logging this as I worked my way through the instructions). Thanks!

docs/development/setup_development.rst
```diff
@@ -39,6 +39,11 @@ be installed via the official documentation links:
...
+    to be in the ``docker`` group. This is easily achieved with
+    ``sudo adduser $USER docker`` and restarting your session.
```
kushaldas: Isn't this be `sudo adduser $USER -G docker` ?

ntoll: I don't think so... (from `adduser --help`):

```
adduser USER GROUP
   Add an existing user to an existing group
```

kushaldas: That is not the case for Fedora/CentOS/RHEL world. This command will not work :)

kushaldas: But, this is for Ubuntu/Debian, so skip this :)

conorsch:
Rather than include steps ourselves, we should link out to upstream docs wherever possible, to reduce maintenance burden on ourselves. I suggest the following patch in lieu of the "warning" box:

```
diff --git a/docs/development/setup_development.rst b/docs/development/setup_development.rst
index 27a4d5fb1..66214cd0a 100644
--- a/docs/development/setup_development.rst
+++ b/docs/development/setup_development.rst
@@ -36,8 +36,11 @@ be installed via the official documentation links:
 * `Docker CE for Ubuntu`_
 * `Docker CE for Debian`_
 
+Make sure to follow the `Post-installation steps for Linux`_, as well.
+
 .. _`Docker CE for Ubuntu`: https://docs.docker.com/install/linux/docker-ce/ubuntu/
 .. _`Docker CE for Debian`: https://docs.docker.com/install/linux/docker-ce/debian/
+.. _`Post-installation steps for Linux`: https://docs.docker.com/install/linux/linux-postinstall/
 
 
 macOS
```

Can you give that a shot, @ntoll, and see if it clarifies the setup for you?

ntoll:
+1 Fixing now.

redshiftzero:
good to go @conorsch?

PR #3831, opened 2018-09-30
actually fix totp window

### LINKED ISSUE #3830: TOTP window tests are still buggy
## Description

the TOTP window test is still buggy :eyes: :eyes: 

## Steps to Reproduce

It's flaky. Just run the 2fa tests a bunch of times.

## Expected Behavior

They pass and window correctly.

## Actual Behavior

They don't. 

```
>       assert now < window_end
E       assert datetime.datetime(2018, 9, 29, 13, 50, 32, 750672) < datetime.datetime(2018, 9, 29, 13, 50, 32)
```

## Comments

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #3830

The root cause of the bug is that if the `seconds` component of the time was `0`, then `seconds % 30` would evaluate to `0`, and we'd have a zero-length TOTP window, thus the tests would always fail.

## Testing

```
make test TESTFILES=test/test_2fa.py
```

### DIFF
--- a/securedrop/tests/test_2fa.py
+++ b/securedrop/tests/test_2fa.py
@@ -18,19 +18,25 @@ from utils.instrument import InstrumentedApp
 
 @contextmanager
 def totp_window():
-    # To ensure we have enough time during a single TOTP window to do the
-    # whole test, optionally sleep.
-    now = datetime.now()
-    num_secs_elapsed_in_time_window = now.second % 30
-    num_secs_left_in_time_window = 30 - num_secs_elapsed_in_time_window
+    '''To ensure we have enough time during a single TOTP window to do the
+       whole test, optionally sleep.
+    '''
 
-    if num_secs_left_in_time_window < 5:
-        time.sleep(30 - num_secs_elapsed_in_time_window)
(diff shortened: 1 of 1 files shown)


securedrop/tests/test_2fa.py
```diff
@@ -18,19 +18,24 @@
...
+    seconds_left_in_window = (now.second % 30) or 30
+
+    window_end = now.replace(microsecond=0) + \
+        timedelta(seconds=seconds_left_in_window)
+    window_end_delta = window_end - now
+
+    # if we have less than 5 seconds left in this window, sleep to wait for
+    # the next window
+    if window_end_delta < timedelta(seconds=5):
+        time.sleep(window_end_delta.second +
```
redshiftzero: looks like `window_end_delta` is a timedelta, and timedelta objects don't have `second` or `microsecond` attributes, so we'll get an `AttributeError`

redshiftzero:
btw the CI failure here is due to the 2FA tests: https://circleci.com/gh/freedomofpress/securedrop/18405

(I can help debug later, just flagging for now)

PR #3837, opened 2018-10-02
Fix xfails

heartsucker:
## Status

Ready for review

## Description of Changes

The tests we have marked as xfails were completely broken and have been since we moved to containerized test. Additionally, the `xfail` marker is overly permissive and allows *any* failure to trigger a test-pass-via-failure.

This PR
- Add `gdb` / `gcore` to the container
- Grants the container the `SYS_PTRACE` capability
- Fixes the borked parts of the tests

## Testing

```
make test TESTFILES=tests/functions/test_submission_not_in_memory.py
```


msheiny:
Soo i had to bone up on `ptrace` while looking over these changes .. fyi this was a good blog post about it http://www.alfonsobeato.net/c/modifying-system-call-arguments-with-ptrace/

msheiny:
so the docker image is indeed getting cached... but its cached on the start of the branch and it does indeed include the `gcore` binary (i inspected the cached image without doing any scripts).

Moving onto inspect capabilities

msheiny:
it was something really silly @heartsucker ... the capability fixes it... i pushed a fix . its basically a divergence issue between CircleCI and dev env. I got a potential PR to fix in #3839 that will hopefully squash this issue in the future. Thanks for putting up with this failing craft we call CI :rose:

heartsucker:
Copying message from chat to somewhere more findable: 

> I'm so proud I managed to crash two opsers' laptops today :sniff:
>
> Btw I'm inclined to say we just remove those xfails because they have never passed, have been completely and utterly broken for like a year, and now they are causing us a giant pain in the ass
>
> Also since we're using python which doesn't have an easy way to force scrub memory, those tests may never pass anyway
>
> Like I'm pretty sure they've been there as xfails since I joined the project

conorsch:
Given the proliferating workload generated by this extremely well intentioned cleanup effort (cf. #3840), I suggest we don't bother with _fixing_ the xfails right now, and simply let them remain as expected failures. It's possible to skip the tests in CI by referring to the `FPF_CI` env var _in the staging run_. 

Porting conversation from the Gitter channel: the changes presented here—presumably the SYS_PTRACE capability specifically—cause breakage for developers running grsecurity-patched kernels on their workstations. Until we suss out a resolution for that behavior—running these tests only in VMs, and not in containers, might suffice—let's defer.

In sum, the xfails can remain expected to fail. :slightly_smiling_face:

heartsucker:
> running these tests only in VMs

Does this mean the burden falls on grsec patched people to remember this, or does that mean I need to set ptrace only if we're not grsec? I can do either, but one is easier and falls more into the development happy path.

redshiftzero:
For posterity: these were introduced in #476, did not pass at that time, and were xfailed in #2548. 

> Also since we're using python which doesn't have an easy way to force scrub memory, those tests may never pass anyway

Indeed. @heartsucker: want to rescope this PR to just remove these failing tests?

PR #3840, opened 2018-10-02
Modify CI docker SD behavior to match dev env

### LINKED ISSUE #3839: Remove CircleCI docker divergence
## Description

The docker remote machine executor is great in terms of saving us potential $$$ but is disastrous for debugging. Let's use the circleci `machine` executor so we can use volume mapping with `docker`. If they end up charging an arm and a leg for it in the future we can move to another platform but at least then our test logic will match development env.

## User Stories

As a developer, it is super frustrating to have two parallel testing/building strategies for development and production.

msheiny:
## Status

Ready for review -- uncovered what look like legit linting failures

## Description of Changes

Fixes #3839 

* Utilize higher :dollar:  cost `machine` executor (at least they plan to charge more for it in the "future")
* Revamp docker caching to only rebuild when `Dockerfile` changes.

## Testing

* Ensure you can still spin up docker locally (tests and dev env)
* Ensure CI doesn't break :crossed_fingers: 

## Deployment

NOPE - only affects dev + CI

Umm  I didn't questionnaire... why you all up in my bizness!?

HEY! Whats with all the :fire: ?

I swear to god... 

Docs!? we have docs!?


msheiny:
holy $#!# - i want on a huge tangent trying to improve the story here and opened up a can of worms. This PR is def. WIP... not ready for review. feel free to unpack and improve if you are feeling gutsy

msheiny:
I want to point out here that `lint` is failing but:
* a majority of the shellcheck ones seem to be a side-effect of bumping the shellcheck version
* i'm actually confused on the mypy failure here though. I also get failures on `develop` when i run it locally. I thought maybe it was just a grsec thing but I'm getting the same error in CI here.

@kushaldas can you eye the mypy failure to make sure I'm not crazy?

redshiftzero:
> i'm actually confused on the mypy failure here though. I also get failures on develop when i run it locally. I thought maybe it was just a grsec thing but I'm getting the same error in CI here.

I get the same mypy error, so this is a legitimate issue, switching `pages-layout` to `pageslayout` _should_ resolve (dashes in Python packages are not allowed)

heartsucker:
Nit that we should use `pages_layout` just for readability.

redshiftzero:
PEP 8 discourages underscores but it is not explicitly disallowed, so either is fine

PR #3844, opened 2018-10-04
Functional pytest

### LINKED ISSUE #3836: Convert functional tests to use pytest/fixtures
## Description

Using pytest fixtures creates a single source for how we initialize the objects that tests rely on. This is generally more maintainable than having each test function/class/module initialize its resources how it sees fit.

Plus, once we move to having the functional tests run against external servers (using TB or otherwise), this would be a single place we could have a switch for altering the tests.

However, this is somewhat non-trivial if we want to ever add `pytest-xdist`. Pytest fixtures can be scoped in different ways, and xdist will parallelize load in a way that guarantees all free functions in a module go to the same worker or all functions in a class go to the same worker. We want to avoid having multiple webdrivers in existence at the same time (memory constraints), and we want to avoid having to recreate too many webdrivers as well as their boot time (on my low/mid-spec laptop) is roughly 2 seconds.

If we convert to the functional tests and ignore xdist support and slow tests, this is very easy. If we want all the bells and whistles, we have to think a little harder.

### Minimum Criteria

We need to continue using the pattern of one `app` per function because this guarantees a clean start for every test.

## User Stories

As a dev, I want to make changes to tests in one place when I update app code, and not be digging around in different places.

heartsucker:
## Status

Work in progress

## Description of Changes

Fixes #3836 

Use pytest for functional tests

## Testing

```
make test TESTFILES=tests/functional
```


redshiftzero:
We decided during sprint planning to merge `tbb.0.9.0` into `develop` first, then rebase this onto latest.

zenmonkeykstop:
Attempt at a rebase started here: https://github.com/zenmonkeykstop/securedrop/tree/functional-pytest-rebase

eloquence:
Hi @heartsucker, sorry for the long silence on this PR. We still are interested in moving towards more parallel execution of tests, and using this approach as a stepping stone. However, many changes to functional tests, especially Tor support, have landed since you worked on this, and major PR surgery would be required.

Our inclination would be to close this PR, and revisit when we prioritize test performance improvements. If you still have time for SecureDrop and would like to work on it again yourself, let us know, and we can set aside time in a future sprint to pair with you on it.

eloquence:
Closing per above comment.

PR #3850, opened 2018-10-07
Migrate to static config

### LINKED ISSUE #1966: Automated web application configuration migration

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #1966

- Define `config.json` configuration file and fields
- Create a migration tool to convert old `config.py` into new `config.json`
- Set defaults at runtime
- Populate missing config values via `ansible`

### New Installs
- Checkout this branch
- Build debs, provision staging VM
- Check that app works
- Back up the configs
- `dpkg -i` the app code deb (or just run `postinst`)
- Check that the it was idempotent and the configs weren't changed

### Old Installs (with ansible)

- Checkout the commit on `develop` that this branch
- Build debs, provision staging VM
- Check out the head of this branch
- Build debs, provision new VM
- Check that app works
- Check that config was migrated and contains the same values as `config.py`
- Back up the new configs to `/tmp` or `~`
- `dpkg -i` the app code deb (or just run `postinst`)
- Check that the it was idempotent and the configs weren't changed

### Old Installs (without ansible)

- Checkout the commit on `develop` that this branch
- Build debs, provision staging VM
- Check out the head of this branch
- Build debs, `scp` to VM
- Install debs manually (to simulate `cron-apt`)
- Check that app works
- Check that config was migrated and contains the same values as `config.py`
- Back up the new configs to `/tmp` or `~`
- `dpkg -i` the app code deb (or just run `postinst`)
- Check that the it was idempotent and the configs weren't changed

### Old Installs Backup/Restore

- Checkout the commit on `develop` that this branch
- Build debs, provision staging VM
- Run backup script
- Check out the head of this branch
- Build debs, provision new VM
- `rm /etc/securedrop/*.json`
- Run restore script
- Check that config was migrated and contains the same values as `config.py`

## Deployment

We need to be able to support the following cases:

1. `config.py` is not present
2. `config.py` present, `config.py` has partial configuration
3. `config.json` present, `config.json` has partial configuration


eloquence:
For the 11/28-12/12, we will aim to do a first time-boxed review; not planning to merge yet.

kushaldas:
I was planning to make the `SDConfig` class into a proper strictly defined class (this is just a comment).

molecule/testinfra/staging/app-code/test_securedrop_app_code.py
```diff
@@ -34,18 +35,6 @@ def test_securedrop_application_apt_dependencies(Package, package):
...
-def test_securedrop_application_test_locale(File, Sudo):
-    """
-    Ensure SecureDrop DEFAULT_LOCALE is present.
-    """
-    securedrop_config = File("{}/config.py".format(
-        securedrop_test_vars.securedrop_code))
-    with Sudo():
-        assert securedrop_config.is_file
-        assert securedrop_config.contains("^DEFAULT_LOCALE")
-        assert securedrop_config.content.count("DEFAULT_LOCALE") == 1
```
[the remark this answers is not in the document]
heartsucker: I don't think we need these tests because the new `Config` class will gracefully handle the missing value and set it to `en_US`.

heartsucker:
Before this is merged, an open question is: should `config.json` be split into `source-config.json` and `journalist-config.json`? I swear there was a reason I didn't do this at first, but I can't for the life of me remember what that might have been. Probably that this was just simpler? The advantage of this would be in the future we could have stricter permissions / AppArmor rules to prevent the source app from reading secret data from the journalist config.

install_files/ansible-base/roles/backup/files/backup.py
```diff
@@ -18,13 +18,15 @@ def main():
...
-    sd_config = os.path.join(sd_code, "config.py")
+    sd_config_py = os.path.join(sd_code, "config.py")
+    sd_config = os.path.join("/etc/securedrop/config.json")
     sd_custom_logo = os.path.join(sd_code, "static/i/logo.png")
 
     tor_hidden_services = "/var/lib/tor/services"
     torrc = "/etc/tor/torrc"
 
     with tarfile.open(backup_filename, 'w:gz') as backup:
+        backup.add(sd_config_py)
```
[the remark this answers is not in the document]
heartsucker: see the logic in `restore.py` to see if this does the trick.

kushaldas:
>Before this is merged, an open question is: should config.json be split into source-config.json and journalist-config.json? I swear there was a reason I didn't do this at first, but I can't for the life of me remember what that might have been. Probably that this was just simpler? The advantage of this would be in the future we could have stricter permissions / AppArmor rules to prevent the source app from reading secret data from the journalist config.

As we discussed last night, please go ahead and split them across into two files. This PR also needs `Test steps`.

heartsucker:
@kushaldas @redshiftzero I've made a few more changes. Basically all all the config generation logic has been moved into `populate_config.py` and doesn't exist in ansible (other than as a shell-out to it). There were some tests added for a couple of common cases for the state of the config files, but there are tons of edge cases and I'm not sure how thorough we want to be about the entire space of possibilities for this.

heartsucker:
Ah, also the configs were split into a source and journalist version, both as python classes and different JSON files. I also stripped out all global references to the config and everything is injected (for tests and the application).

install_files/ansible-base/roles/backup/files/backup.py
```diff
@@ -18,13 +18,15 @@ def main():
     sd_data = '/var/lib/securedrop'
 
     sd_code = '/var/www/securedrop'
-    sd_config = os.path.join(sd_code, "config.py")
+    sd_config_py = os.path.join(sd_code, "config.py")
+    sd_config = os.path.join("/etc/securedrop/config.json")
```
kushaldas: These config json file paths should be updated with the new 2 files.

molecule/builder/tests/test_securedrop_deb_package.py
```diff
@@ -151,15 +151,16 @@ def test_deb_package_control_fields_homepage(File, Command, deb):
 def test_deb_package_contains_no_config_file(File, Command, deb):
     """
     Ensures the `securedrop-app-code` package does not ship a `config.py`
-    file. Doing so would clobber the site-specific changes made via Ansible.
+    file or a `config.json`. Doing so would clobber the site-specific changes
```
kushaldas: Update comment to have two different file names.

molecule/testinfra/staging/app-code/test_securedrop_app_code.py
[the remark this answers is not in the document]
kushaldas: >I don't think we need these tests because the new Config class will gracefully handle the missing value and set it to en_US.

Correct, but, please make sure that is there is a test which checks for this default language value.

molecule/testinfra/staging/app-code/test_securedrop_app_code.py
```diff
@@ -64,17 +53,22 @@ def test_securedrop_application_test_journalist_key(File, Sudo):
 
     # Let's make sure the corresponding fingerprint is specified
     # in the SecureDrop app configuration.
-    securedrop_config = File("{}/config.py".format(
-        securedrop_test_vars.securedrop_code))
+    securedrop_config = File("/etc/securedrop/config.json")
```
kushaldas: This test will also need update.

securedrop/i18n.py
```diff
@@ -39,7 +39,7 @@ def setup_app(config, app):
     global LOCALES
     global babel
 
-    translation_dirs = getattr(config, 'TRANSLATION_DIRS', None)
+    translation_dirs = config.TRANSLATION_DIRS
```
kushaldas: @heartsucker I can not explain how happy I am to see these `*attr` calls being removed :heart: .

securedrop/sdconfig.py
```diff
@@ -11,154 +14,216 @@
...
+    WTF_CSRF_ENABLED = True
+
+    def __init__(self, secret_key, cookie_name, env):
+        # type: (str, str, str) -> None
+        self.SECRET_KEY = secret_key
+        self.SESSION_COOKIE_NAME = cookie_name
+
+        if env == 'prod':
+            # This is recommended for performance, and also resolves #369
+            self.USE_X_SENDFILE = True  # type: ignore
```
kushaldas: Can we please have all the possible attributes of the class with some default values? Let us say, locked down production values, and only enable/set in the other cases. Will really help to read/type annotations.

securedrop/sdconfig.py
```diff
@@ -11,154 +14,216 @@
...
-        except AttributeError:
-            pass
+        with open(config_file) as f:
+            self._json_config = json.loads(f.read())
 
-        try:
-            self.DATABASE_USERNAME = _config.DATABASE_USERNAME  # type: ignore
-        except AttributeError:
-            pass
+        self.SECUREDROP_DATA_ROOT = '/var/lib/securedrop/'
```
kushaldas: Please add the type annotations.

securedrop/sdconfig.py
heartsucker: Flask handles the default on its own, and I don't think we should set them ourselves.

securedrop/sdconfig.py
heartsucker: Can you check again that sufficient type annotations have been added?

securedrop/sdconfig.py
kushaldas: @heartsucker I am asking defaults to enable/use `mypy` in a better way.

install_files/securedrop-app-code/DEBIAN/postinst
```diff
@@ -85,6 +85,9 @@ case "$1" in
     done
 
     chown -R www-data:www-data /var/lib/securedrop /var/www/securedrop
+    
+    mkdir -p /etc/securedrop
+    chdmod 0750 /etc/securedrop
```
emkll: Is this meant to be `chmod`?

eloquence:
Recapping our discussion from sprint planning today:

We want to minimize the risks associated with this big (and important!) change by ensuring we express through automated tests the potential states that the configuration file can be in. The number of states should be manageable given that the file is not manually edited: when have the playbooks been run, and what features have been explicitly enabled? We must assume that there are long-running instances that have not run playbooks since pre-0.4. 

@heartsucker -- on the FPF side, we've committed to up to 8 person hours to help get this closer to the finish line, but we should make sure we're on the same page vis-a-vis the testing strategy. Let's discuss w/ @redshiftzero on Gitter, or during standup, next time you're around.

heartsucker:
As a note to reviewers, the test `test_populate_config.py` is the primary place one would want to look to check for what states of config files we are currently operating under. Additional tests would probably be added here.

eloquence:
Hi @heartsucker, this has been sitting for a long time and I think it's time to acknowledge that we're probably not going to land a version of this change, which would require significant PR surgery at this point, and which is quite high-risk in terms of potential for breaking existing configs.

Instead our bias is to prioritize moving as many settings as reasonably possible to the database (where we can use alembic migrations for change management) and the web-based admin UI.

I'm going to leave this open a few more days in case other folks want to make the case for revisiting this PR, and then close.

eloquence:
Closing PR per the above comment; let's keep the `static-config` branch around for now in case we want to revisit at a later date.

PR #3855, opened 2018-10-09
Added changes in UI of admin_add_login page

### LINKED ISSUE #3746: Admin interface registration form should list username limitations
## Description

Username field should have some smol help text indicating whether a username can have spaces in it, is case-sensitive, etc.

## User Research Evidence

During SecureDrop trainings, one of the questions asked by admins and journalists creating accounts on the SecureDrop admin interface is whether they can have spaces in their username or if the username is case-sensitive. The form should indicate these things.

nightwarrior-xxx:
## Status

Ready for review 

## Description of Changes

Fixes #3746 

* Changed the UI of the admin_add_user page

## Testing

Have a look at the admin_add_user page

### DIFF
--- a/admin/requirements.txt
+++ b/admin/requirements.txt
@@ -113,9 +113,9 @@ markupsafe==1.0 \
 netaddr==0.7.19 \
     --hash=sha256:38aeec7cdd035081d3a4c306394b19d677623bf76fa0913f6695127c7753aefd \
     --hash=sha256:56b3558bd71f3f6999e4c52e349f38660e54a7a8a9943335f73dfc96883e08ca
-paramiko==2.4.1 \
-    --hash=sha256:24fb31c947de85fbdeca09e222d41206781581fb0bdf118d2ef18f6e414cd388 \
-    --hash=sha256:33e36775a6c71790ba7692a73f948b329cf9295a72b0102144b031114bd2a4f3 \
+paramiko==2.4.2 \
+    --hash=sha256:3c16b2bfb4c0d810b24c40155dbfd113c0521e7e6ee593d704e84b4c658a1f3b \
+    --hash=sha256:a8975a7df3560c9f1e2b43dc54ebd40fd00a7017392ca5445ce7df409f900fcb
     # via ansible
 prompt-toolkit==1.0.15 \
     --hash=sha256:1df952620eccb399c53ebb359cc7d9a8d3a9538cb34c5a1344bdbeb29fbcc381 \
--- a/install_files/ansible-base/group_vars/all/securedrop
+++ b/install_files/ansible-base/group_vars/all/securedrop
@@ -53,5 +53,5 @@ securedrop_cond_reboot_file: /tmp/sd-reboot-now
 
 # If you bump this, also remember to bump in molecule/builder/tests/vars.yml
 securedrop_pkg_grsec:
-  ver: "4.4.144"
+  ver: "4.4.144-1"
   depends: "linux-image-3.14.79-grsec,linux-image-4.4.135-grsec,linux-firmware-image-4.4.135-grsec,linux-image-4.4.144-grsec,linux-firmware-image-4.4.144-grsec"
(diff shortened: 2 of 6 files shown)


nightwarrior-xxx:
Have a look a screen shot
screen_shot_of_changes made

nightwarrior-xxx:
@heartsucker Please review!
Regards,
@nightwarrior-xxx

heartsucker:
Hey @nightwarrior-xxx. There's still two diffs that are just whitespace. Also, could you squash your commits so this PR is just a single commit? It makes the history a little cleaner.

securedrop/journalist_templates/admin_add_user.html
```diff
@@ -33,9 +45,9 @@
...
     <br>
     {% for error in form.is_hotp.errors + form.otp_secret.errors %}
-      <span class="form-validation-error">{{ error }}</span><br>
+    <span class="form-validation-error">{{ error }}</span><br>
     {% endfor %}
   </p>
   <button type="submit" class="sd-button"><i class="fa fa-plus"></i>{{ gettext('ADD USER') }}</button>
 </form>
-{% endblock %}
+{% endblock %}
```
SaptakS: I think you should keep the existing newline in the file.

nightwarrior-xxx:
Hey @heartsucker,I just did indentation,which resulted in the removing of 2 white space you are talking about .
Regards,
@nightwarrior-xxx

heartsucker:
Heya. Some more notes.

screenshot_2018-10-10_20-03-23

I still think the help text is a little too small. My laptop is kinda shitty, so maybe it's just that the text doesn't render great, but i think this could be a problem for other users.

It would be nice to make the grey background a little bigger (using padding/margin) so that it's not hugging the text so tight.

nightwarrior-xxx:
Hey @heartsucker ,Cool I will do that.But what about those 2 spacing problems ,I just did indentation ,still do I change it as you said?
Regards,
@nightwarrior-xxx

heartsucker:
I think we need to call in the UX experts to make an assessment here. I'm mostly backend, so we'll need more input from people who can make a call on this.

ninavizz:
Ok! @SaptakS will you be in the UX call, in the AM PST? I think I get what's being asked-for here, but it's always nice to chat needs through. Regardless, can follow-up tomorrow late afternoon PST with whatever may be needed. Cheerio!

nightwarrior-xxx:
@heartsucker Should I close this PR and made a new PR in order to not to make history.Moreover I am facing issue in pulling also.

securedrop/sass/journalist.sass
```diff
@@ -101,3 +101,15 @@ button.small-danger, a.btn.small-danger, .btn.small-danger
 
 .journalist-reply__input
   max-width: 700px
+
+.journalist-username__notes
+  font: 12px Sans-Serif
+  font-weight: normal
+  letter-spacing: 0.05px
+
+.code_modified
```
SaptakS: Firstly, you need to put this code [here](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/sass/modules/_code.sass). Secondly, what is the benefit of removing the `padding` and the `border-bottom`? @ninaviz what do you think about this design vs the one we have for `.code` ?

nightwarrior-xxx: @SaptakS I removed padding for highlighted password part as it was taking more space on the upper side but later after @heartsucker said that font-size is too small so I increased the font size and added padding with suitable pixels.I will push PR after modifying it again.

heartsucker:
The current UI looks like this which I think is nearly there.

screenshot_2018-10-12_14-51-16

For the UX team, my one open question is: what should we do about multi-line passphrase? Because right now someone might forget to copy the last word of the passphrase and then not be able to log in.

nightwarrior-xxx:
@heartsucker I made a new PR because I was unable to fetch changes,which I tried hard to fetch but failed.New issue no. is #3873 .I will not repeat this mistake again.

Moreover,we can increase the border size of the form ,or reduce the length of password by one word if we want the password to be in one line
Regards,
@nightwarrior-xxx

eloquence:
Closing as duplicate of #3873.

PR #3867, opened 2018-10-11
Updates libvirt setup instructions to cover Debian 9 correctly, tiny fix to install docs

### LINKED ISSUE #3860: Docs: libvirt instructions inaccurate for Debian 9
## Description

The libvirt setup documentation is inaccurate for Debian 9:

 - The section on checking the status of the libvirt/KVM install is Ubunut-only
 - virt-manager is not installed by default
 - vagrant-mutate plugin installation not mentioned.

## Steps to Reproduce

Try running thru docs with a Debian box.

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #3860 and also removes an outdated reference to the custom header image in the installation instructions.

## Testing

Docs-only PR. To test, verify the libvirt setup process in Debian 9.

### DIFF
--- a/docs/development/virtual_environments.rst
+++ b/docs/development/virtual_environments.rst
@@ -149,7 +149,7 @@ Install libvirt and QEMU:
 .. code:: sh
 
    sudo apt-get update
-   sudo apt-get install libvirt-bin libvirt-dev qemu-utils qemu
+   sudo apt-get install libvirt-bin libvirt-dev qemu-utils qemu virt-manager
    sudo /etc/init.d/libvirt-bin restart
 
 Add your user to the libvirtd group:
@@ -159,8 +159,7 @@ Add your user to the libvirtd group:
    sudo addgroup libvirtd
    sudo usermod -a -g libvirtd $USER
 
-Install Vagrant along with the required plugins for converting and using
-libvirt boxes:
+Install  the  required Vagrant plugins for converting and using libvirt boxes:
 
 .. code:: sh
 
@@ -171,6 +170,14 @@ libvirt boxes:
    valid provider. In this case, remove Vagrant with ``sudo apt-get remove
    vagrant`` and reinstall it.
 
+Log out, then log in again. Verify that libvirt is installed and KVM is available:
+
+.. code:: sh
+  
+   libvirtd --version
+   kvm-ok
+
+
 Debian 9 setup
 ^^^^^^^^^^^^^^
 
@@ -179,9 +186,10 @@ Install Vagrant, libvirt, QEMU, and their dependencies:
 .. code:: sh
 
    sudo apt-get update
-   sudo apt-get install -y vagrant vagrant-libvirt libvirt-daemon-system qemu-kvm
+   sudo apt-get install -y vagrant vagrant-libvirt libvirt-daemon-system qemu-kvm virt-manager
    sudo apt-get install -y ansible rsync
    vagrant plugin install vagrant-libvirt
+   vagrant plugin install vagrant-mutate
    sudo usermod -a -G libvirt $USER
    sudo systemctl restart libvirtd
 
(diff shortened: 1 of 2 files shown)


zenmonkeykstop:
@edenemmanuel wanna take a look if you get a chance?

edenemmanuel:
Oh hello hi hola, I'm installing Debian in a VM working on it. I have some ideas for good one liners that check for virtualization support without `cpu-checker` as well.

Let's see what I can whip up by tomorrow evening.

docs/development/virtual_environments.rst
```diff
@@ -195,18 +203,15 @@ Add your user to the kvm group to give it permission to run KVM:
...
-
-Log out, then log in again. Verify that libvirt is installed and KVM is
-available:
+Log out, then log in again. Verify that libvirt is installed:
 
 .. code:: sh
 
-   libvirtd --version
-   kvm-ok
+   sudo libvirtd --version
```
edenemmanuel: Using `sudo` sounds good to me. Not sure why that's needed. We really shouldn't be if the user is in the `libvirt` group, but this should be fine for now.

docs/development/virtual_environments.rst
```diff
@@ -195,18 +203,15 @@ Add your user to the kvm group to give it permission to run KVM:
...
+
```
edenemmanuel: A good one liner to check if virtualization support is ready could be:

`egrep -c '(vmx|svm)' /proc/cpuinfo`

This will show the number of lines that either have `vmx` (for Intel virtualization support) or `svm` (for AMD virtualization support). If the number is greater than 0, then support is enabled.

What do y'all think?

docs/development/virtual_environments.rst
```diff
@@ -159,8 +159,7 @@ Add your user to the libvirtd group:
    sudo addgroup libvirtd
    sudo usermod -a -g libvirtd $USER
 
-Install Vagrant along with the required plugins for converting and using
-libvirt boxes:
+Install  the  required Vagrant plugins for converting and using libvirt boxes:
```
heartsucker: Nit: Double spaces around `the`.

docs/development/virtual_environments.rst
```diff
@@ -171,6 +170,14 @@ libvirt boxes:
    valid provider. In this case, remove Vagrant with ``sudo apt-get remove
    vagrant`` and reinstall it.
 
+Log out, then log in again. Verify that libvirt is installed and KVM is available:
+
+.. code:: sh
+  
+   libvirtd --version
+   kvm-ok
```
heartsucker: Is this (`kvm-ok`) a command? Because when I followed the docs originally, this wasn't installed.

docs/development/virtual_environments.rst
heartsucker: Agree that we should remove `sudo`.

docs/development/virtual_environments.rst
heartsucker: Might want to make it `flags\s*:.* (vmx|svm) ` to tighten it down a bit.

docs/development/virtual_environments.rst
heartsucker: Ah wait that's Ubuntu. You removed it on the Debian ones. Nvm.

edenemmanuel: Yep!

docs/development/virtual_environments.rst
edenemmanuel: Ah, so my issue is that there doesn't seem to be a reason why we need `sudo` but currently that is the case. Even if your local user is part of the `libvirt` and `kvm` groups, this is needed.

heartsucker: There sholdn't be. You have to log out and login *completely*. Not just the terminal session. If you still need `sudo`, something went wrong.

heartsucker: Oh man. I think you were right. I just had a live `sudo` session so I didn't hit that same error, but I hit it now.

zenmonkeykstop: Looks like it's needed for checking libvirtd on Debian, but `sudo` isn't a requirement for regular use. If virt-manager and the cli tools are good, then we good?

docs/development/virtual_environments.rst
zenmonkeykstop: Gonna add that last one!

docs/development/virtual_environments.rst
```diff
@@ -203,12 +203,14 @@ Add your user to the kvm group to give it permission to run KVM:
...
-Log out, then log in again. Verify that libvirt is installed:
+Log out, then log in again. Verify that libvirt is installed and your system
+supports KVM:
 
 .. code:: sh
 
    sudo libvirtd --version
-
+   [ `egrep -c 'flags\s*:.*(vmx|svm)' /proc/cpuinfo` -gt 0 ] &&  \
+   echo "KVM supported!" || echo "KVM not supported..."
```
edenemmanuel: Nice.

Issue #3862, opened 2018-10-10
Sources with a null value for last_updated will break the web UI


heartsucker:
## Description

If a source is created without a `last_updated` filed, it will cause the following error on the journalist index page.

```
172.17.0.1 - - [10/Oct/2018 18:49:40] "GET / HTTP/1.1" 500 -
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 2309, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 2295, in wsgi_app
    response = self.handle_exception(e)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1741, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python2.7/dist-packages/flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/home/heartsucker/code/freedomofpress/securedrop/securedrop/journalist_app/main.py", line 79, in index
    starred=starred)
  File "/usr/local/lib/python2.7/dist-packages/flask/templating.py", line 135, in render_template
    context, ctx.app)
  File "/usr/local/lib/python2.7/dist-packages/flask/templating.py", line 117, in _render
    rv = template.render(context)
  File "/usr/local/lib/python2.7/dist-packages/jinja2/environment.py", line 1008, in render
    return self.environment.handle_exception(exc_info, True)
  File "/usr/local/lib/python2.7/dist-packages/jinja2/environment.py", line 780, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/home/heartsucker/code/freedomofpress/securedrop/securedrop/journalist_templates/index.html", line 1, in top-level template code
    {% extends "base.html" %}
[...]

PR #3871, opened 2018-10-12
Made last_updated required in the database schema

mdrose:
Addresses #3862

Prevents issues arising if there ever happens to be a null value in last_update on the Sources table by making it a required field. I can't think of a situation where a null value in last_updated would be valuable. If that's the case, it might as well be enforced in the schema.

### DIFF
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -55,7 +55,7 @@ class Source(db.Model):
     filesystem_id = Column(String(96), unique=True)
     journalist_designation = Column(String(255), nullable=False)
     flagged = Column(Boolean, default=False)
-    last_updated = Column(DateTime)
+    last_updated = Column(DateTime, nullable=False)
     star = relationship("SourceStar", uselist=False, backref="source")
 
     # sources are "pending" and don't get displayed to journalists until they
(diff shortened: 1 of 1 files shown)


redshiftzero:
@heartsucker recall that we had defect #3635 in the journalist notifications due to the default value in the `last_updated` field, which was then removed in #3648, so we either need to keep the current behavior, or we need to modify the journalist notifications logic. 

Clarying question: is bug #3862 an issue that can occur with the existing application logic? My understanding is no, since `last_updated` will be set upon submission, and sources without submissions should not appear in the journalist interface UI.

mdrose:
@redshiftzero Yeah, it looks like this schema change would be in conflict with the notification code. That code searches the Source table. Should we want make this change, the logic for that would have to be changed to search the Submissions table, checking the related Source.last_updated, or something similar, avoiding notifications where a Source was created, but submitted nothing that day.

mdrose:
@redshiftzero Yeah, it looks like this schema change would be in conflict with the notification code. That code queries the Source table. Should we want make this change, the logic for that would have to be changed to query the Submissions table, checking the related Source.last_updated, or something similar, avoiding notifications where a Source was created, but submitted nothing that day.

heartsucker:
I have a fix that supersedes this one in #3875

redshiftzero:
Hey so, I hear you, but let's avoid code changes that will require both a significant database migration and a modification to the journalist notifications logic (since the journalist notifications are difficult to QA) for an issue that should only be encountered in the development environment. I merged #3875 since it's a minimal and uncontroversial resolution. Thanks y'all :heart:

PR #3872, opened 2018-10-12
docker ignore tox/venv

### LINKED ISSUE #3550: The file context in the admin docker build is too dang high!!!

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #3550

Docker ignore generated resources.

## Testing

Run the tests, get a big tox directory, rebuild the container with `DOCKER_BUILD_VERBOSE=true`, see the following line:

```
Docker image build in progress Sending build context to Docker daemon  202.2kB
```

### DIFF
--- a/admin/.dockerignore
+++ b/admin/.dockerignore
@@ -1,0 +1,4 @@
+.tox
+.venv
+**/__pycache__
+**/*.pyc


redshiftzero:
hey @msheiny since you also had a PR tackling #3550 earlier, what do you think of this change? (i'm cool with merge if you want to approve)

PR #3873, opened 2018-10-12
Added changes to UI of the admin_add_user page

nightwarrior-xxx:
## Status

Ready for review 

## Description of Changes

Fixes #3855 

 Added changes to the admin_add_user page


nightwarrior-xxx:
@SaptakS  Please review.I made a new PR as I was unable to pull changes from upstream.I will not repeat this mistake next time.

eloquence:
I would recommend re-scoping this PR to only fix the actual problem identified in #3746 and handle other design changes separately.

Regarding the actual text, we should ensure we have consensus on what the recommendations are. 

> "Username must not contain your original name"

Do you mean to say "real name"? If so, what is the underlying security reason?

> "Username must be unique"

I don't know how useful this one is; as far as @huertanix stated in #3746, the main questions that come up are case-sensitivity and spaces. We should not add text just for the sake of adding text; most systems require usernames to be unique, so if this is not a common point of confusion, I don't think we need to say so here.

We should, however, state if the username may contain spaces (it can).

nightwarrior-xxx:
@eloquence Gotcha.That means we should just add only these points 
- Username can contain spaces.
- Username is case sensitive.

eloquence:
Yes, and per my earlier comment, I suggest we remove the other style changes from this PR. Keeping PRs scoped clearly to the issues they resolve simplifies review.

Once this is done, could you also upload a before/after screenshot? @ninavizz may have some final styling suggestions.

Thanks a lot! :)

nightwarrior-xxx:
@eloquence @ninavizz Please review .
```Page before modification```
currentsecuredropaddadminuserpage

```Page after modification```
page after modification

 
Regards,
@nightwarrior-xxx

**Comment by eloquence** [comment]
Thanks, @nightwarrior-xxx ! I would suggest two changes:

- Don't add a "Username" label to the username input. To be sure, there are arguments in favor of labels; the problem right now is one of inconsistency. The other input on this form lacks a label, as does the sign-in form. If we want to have a discussion about adding labels to inputs, it should be done consistently across the board, and separately from this PR.

- Remove the new styling of the password. This is a separate issue from the one you're looking to resolve with this PR, and additional style changes should be applied consistently and discussed separately.

**Comment by nightwarrior-xxx** [comment]
Hey,
@eloquence and @ninavizz Please review!!!

```current page```
currentsecuredropaddadminuserpage
```page after modification```
page_after_modifications2

Regards,
@nightwarrior-xxx

**Comment by ninavizz** [comment]
@nightwarrior-xxx Looks good to me for asap deploy; getting fussy on things, I'd love an `&nbsp;` before the "A" in "Add" on the button, and maybe 5-10px of padding between the HOTP field and the checkbox text above it. The latter points are not blockers, tho. Thx!! :)

**Comment by eloquence** [comment]
Let's keep this out of the current PR, @ninavizz, I'd prefer to keep this scoped exactly to fixing the username issue identified in the bug. Thanks :)

File: securedrop/sass/journalist.sass [hunk 1]
```diff
@@ -101,3 +101,9 @@ button.small-danger, a.btn.small-danger, .btn.small-danger
 
 .journalist-reply__input
   max-width: 700px
+
+.journalist-username__notes
+  font: 12px Sans-Serif
+  font-weight: normal
+  letter-spacing: 0.05px
```
eloquence: Why is this here? I cannot visually tell much of a difference from this fraction of a pixel increase.

eloquence:
Aside from the minor question about the CSS, I have no concerns about this PR. There's a slight increase in whitespace above the username field, but I don't see much of a consistency issue with the rest of the form.

securedrop/sass/journalist.sass
ninavizz: .05 looks like an erroneous attempt at entering .5, perhaps? I <3 .5px letterspacing, personally.

nightwarrior-xxx:
@ninavizz  @eloquence 
Please review!!
```Before changes```
currentsecuredropaddadminuserpage
```After Modifications```
page_after_modification3

**Comment by nightwarrior-xxx** [comment]
@heartsucker Please review

File: securedrop/sass/modules/_main-content-area.sass [hunk 2]
```diff
@@ -7,8 +7,11 @@
...
 
+      &.flex-end
+        align-items: flex-end
+
     // This query is to reorder the flex box elements so that they correctly form the
     // two boxes (submit/returning) when the screen is sized down
     @media only screen and (max-width: 768px)
       .container
-        display: block
+        display: block
```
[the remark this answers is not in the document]
heartsucker: Hmm. My comment in the review doesn't seem to be appearing. Anyway. This is still missing a newline at the EOF.

nightwarrior-xxx:
@heartsucker and @redshiftzero Please review

PR #3909, opened 2018-10-30
Nested Virtualization of Staging Env in CI

### LINKED ISSUE #3702: Switch CircleCI staging environment to GCE + libvirt VMs
## Description

Our staging environment is currently directly provisioned to Xen-based VMs running on AWS, due to the lack of support for nested virtualization. That means we can't use the `grsecurity` kernel, and the environment differs from our locally run staging VMs in other ways. Moreover, we can't add support for new tests such as upgrade testing (#1689). 

We've previously considered workarounds (see #2892), but we now have established that we can use GCE's nested-virtualization capabilities (see #3452). The next logical step will be to switch the staging environment spun up by CircleCI to GCE running our libvirt-based staging VMs.

## User Stories

As a developer, I want CI results to reflect the preferred staging environment as much as possible, so that bugs and errors are caught early in development.

msheiny:
## Status

Ready for review

## Description of Changes

Fixes #3702 

* Drops AWS staging logic
* Introduces GCE staging logic
* Now with :100: more GRSEC tests (previously there were none)

## Testing

To replicate CI env locally:

* Get access to Google service account credentials (talk to @msheiny :heart_eyes: )
* export that service account as an environment variable called `GOOGLE_CREDENTIALS`
* source ci env file (`. devops/gce-nested/ci-env.sh`)
* Start the environment (`devops/gce-nested/gce-start.sh`)
* Start the test (`devops/gce-nested/gce-runner.sh`)
* Kill the test (`devops/gce-nested/gce-stop.sh`)

1 GRSEC tests is failing and has been `xfail`ed. Making a new ticket to track fix for that. The problem is reproducible locally and existed before introduction of this PR.

## Deployment

Only affects CI


msheiny:
> Update those docs! https://docs.securedrop.org/en/release-0.10.0/development/testing_continuous_integration.html still talks about AWS and lack of grsec testing—no longer true!

I'll make those docs changes in a separate PR with a `docs-*` prefix. Iterating on the various nits that I'm sure will come up when updating documentation is really annoying to do when the feedback loop from the staging tests are so long. Or I'd be willing to start collab in something like etherpad first.

conorsch:
> I'll make those docs changes in a separate PR

A PR into this PR would work; that way we can skip the long-running staging logic, _and_ still make sure that the dev docs are accurate, since the CI changes and accompanying docs will land in develop together.

conorsch:
Rebased on top of 36cbb217dcd1941f2b509c58f354c3eb77073c7f (latest develop), to re-run CI, including changes in particular from #3913.

molecule/testinfra/staging/common/test_grsecurity.py
```diff
@@ -160,8 +143,7 @@ def test_apt_autoremove(Command):
     assert "The following packages will be REMOVED" not in c.stdout
 
 
-@pytest.mark.skipif(os.environ.get('FPF_GRSEC', 'true') == "false",
-                    reason="Need to skip in environment w/o grsec")
+@pytest.mark.xfail(strict=True)
```
heartsucker: Can you comment the reason for `xfail`-ing this instead of skipping? And if possible, can you narrow the scope of the `xfail` to the exception we expect to see?

conorsch:
Just noticing that we lost the "Don't run staging CI if branch begins with `docs-`" logic... if I find more issues like that, may punt to follow-up tickets in order to get this in.

conorsch:
Resolved the outstanding concerns, _except_ for: 

> `devops/jenkins/TorNightlyPipeline` will start to fail post merge if not updated; it still references the old AWS logic, as well

It's true the AWS references should not be there, but in order to resolve, we need override support on the GCE logic to provide custom vars. Probably a wrapper script is best; I'd prefer to observe the Jenkins failures (which are only shown internally, since they run on a nightly basis, and will _not_ break CI on PRs to this repo). @msheiny feel free to chime in here, but I suggest we merge as-is and follow-up on refining the Jenkins logic on the backend.

Waiting to approve until CI passes, since I made changes to the logic as part of final review.

msheiny:
>  If it's cleaner, happy to excise in this PR, if it's going to remain a dead codepath for a while.

I have another pending PR in infra to squash the job from jenkins (its still there but the trigger to run it is disabled so it never runs). I can pull an accompanying PR in SD to kill it but i'm fine if you want to purge it here instead.

.circleci/config.yml
```diff
@@ -113,75 +113,33 @@ jobs:
...
-          name: Installation pre-reqs
-          command: pip install -U -r securedrop/requirements/develop-requirements.txt
-
-      - run:
-          name: Check Python dependencies for CVEs
-          command: make safety
-
-      - run:
-          name: Run static security testing on source code
-          command: make bandit
```
redshiftzero: was removing these steps (static analysis and checking for CVEs) intentional? 

ok if not (I will add them again), just was attempting to make some unrelated CI changes over in `xenial-pgp-journalist` and noticed this

conorsch: No, that was _not_ intentional, @redshiftzero — my oversight during review. I'll open an issue to track the re-add, fine if it comes in via your Xenial-related CI changes.

PR #3922, opened 2018-11-06
Add client documentation. First draft.

ntoll:
## Status

Ready for review.

## Description of Changes

Adds a basic first draft of documentation for the SecureDrop client.

More docs.

## Testing

`make docs` then visit: http://127.0.0.1:8000/client.html

## Deployment

Any special considerations for deployment? None.


docs/client.rst
```diff
@@ -0,0 +1,137 @@
+The SecureDrop Client Application
+=================================
+
+As part of the ongoing work to make a user (journalist?) friendly workstation
```
heartsucker: Just use "journalist" here.

docs/client.rst
```diff
@@ -0,0 +1,137 @@
...
+The source code, and related issues are `hosted on GitHub <https://github.com/freedomofpress/securedrop-client>`_.
+
+Developer Setup
+---------------
+
+Clone the repository::
+
+    git clone git@github.com:freedomofpress/securedrop-client.git
+
+Create and enable a new Python3 virtualenv::
```
heartsucker: Why is this step here? Are we suggesting they install `pipenv` within a venv? I feel like to avoid confusion we might just suggest they `sudo pip install pipenv` or `pip install --user pipenv`

docs/client.rst
```diff
@@ -0,0 +1,137 @@
...
+
+    python3 -m venv my_virtualenv
+    source my_virtualenv/bin/activate
+
+Set up the Python dependencies::
```
heartsucker: This duplicates the steps that are in the client `README`. I think it would be better to say "follow the steps in the readme to init the dev env".

docs/client.rst
```diff
@@ -0,0 +1,137 @@
...
+
+    pip install pipenv
+    pipenv install --dev
+    pipenv shell
+
+Install the packaged dependencies::
+
+    sudo apt install python3-pyqt5 python3-pyqt5.qtsvg xvfb
```
heartsucker: I feel like running non-dev related commands in the `pipenv` should be discouraged because it's not necessary and also you lose things like history or some people might have non-standard setups this could conflict with.

I also feel like it's generally better to suggest doing `pipenv run make check` because if someone forgets to enter the dev shell, they are going to get confusing errors and then have to back track further to see where they went wrong.

docs/client.rst
```diff
@@ -0,0 +1,137 @@
...
+
+How to Find Help
+----------------
+
+If you would like to report a problem `submit a new issue <https://github.com/freedomofpress/securedrop-client/issues/new>`_.
+
+If you'd like to chat with other developers working on the client drop
+into our `Gitter chat channel for the project <https://gitter.im/freedomofpress/securedrop>`_.
+
+Every week day at 10am (Pacific Time) we take part in a public daily stand-up,
```
heartsucker: Might want to make it more clear that we don't do it *every* weekday because of holidays and also it seems to be Fridays are spotty.

docs/client.rst
```diff
@@ -0,0 +1,137 @@
...
+usually via a `meeting on Jitsi <https://meet.jit.si/QuickWizardsDanceHigh>`_
+(although the details of each daily meeting are published on the Gitter channel
+five minutes before the start of the meeting). All are welcome to contribute.
+
+Otherwise, read on.
+
+Client Architecture
```
heartsucker: This feels whole section feels like too much detail outside the client repo. if we want this level of detail, we should make a docs build inside the client that contains this to make it easier to keep things in sync.

docs/client.rst
```diff
@@ -0,0 +1,137 @@
...
+* ``storage.py`` - contains the functions needed for interacting with a remote SecureDrop API and the local database.
+* ``utils.py`` - generic utility functions needed throughout the application.
+* ``gui`` - this namespace contains two modules: ``main.py`` (containing the ``Window`` class through which all interactions with the user interface should happen) and ``widgets.py`` (containing all the custom widgets used by the ``Window`` class to draw the user interface).
+
+We try very hard to leep the application logic and UI code cleanly separated.
+Furthermore, we try equally hard to ensure the main GUI code always remains
+unblocked. For instance look at how the ``APICallRunner`` is used in
+``logic.py`` to make unblocked network calls to the remote API.
+
+We expect all classes, methods and functions to have docstrings describing the
```
heartsucker: We don't enforce this among ourselves and should clean this up before requiring new contributors do do the same. Additionally, we should include `:param xxx:` and `:return:` markup.

docs/client.rst
```diff
@@ -0,0 +1,137 @@
...
+intention behind the code. We expect such docstrings to **remain up to date**
+as the code evolves. ;-)
+
+If possible, please use `Python type hints <https://docs.python.org/3.5/library/typing.html>`_
```
heartsucker: We don't enforce this among ourselves and should clean this up before requiring new contributors do do the same.

docs/client.rst
```diff
@@ -0,0 +1,137 @@
...
+the ``tests/test_logic.py`` file.
+
+To run the complete test suite simply type::
+
+    make check
+
+Our code style checkers, full test suite and coverage checker will run and
+report any errors.
+
+We use the `PyTest testing framework <https://docs.pytest.org/en/latest/>`_ for
```
heartsucker: We don't enforce this (docs/intent) among ourselves (well enough) and should clean this up before requiring new contributors do do the same. Further, sometimes a test is very obvious or commends scattered through are sufficient. However, I *would* support making this mandatory where each test has some sort of cucumber like specification in the docstring.

docs/client.rst
```diff
@@ -0,0 +1,137 @@
...
+code base. If you're unsure how to achieve this, please don't hesitate to get
+in touch via Gitter or mention this in your description of any pull requests
+you submit.
+
+Contributing
+------------
+
+Out open issues are `on GitHub <https://github.com/freedomofpress/securedrop-client/issues>`_.
+
+Please remember that we have a code of conduct and expect all contributors to
```
heartsucker: +1

docs/client.rst
redshiftzero: yep on friday AM pacific time FPF has 10% time and as such don't hold standup :-D

docs/client.rst
redshiftzero: I agree with the DRY sentiments here. That said I think we should keep all the "getting started" type developer docs in one place (i.e. docs.securedrop.org) - so we should migrate `README.md` content here and just provide a pointer to this page of documentation in the docs (providing a pointer to very detailed docs is what we do for the securedrop server readme, which I think is pretty nice)

ntoll:
Hi folks, I've made corrections and clarifications in light of the feedback. Please feel free to change/add/remove/update anything. :-)

PR #3937, opened 2018-11-21
Fixing broken reference of Tails installation page.

ultimatecoder:
## Status

Ready for review

## Testing

Verify the existing reference of Tails OS installation page is broken. Assert proposed link is working.

### DIFF
--- a/docs/set_up_tails.rst
+++ b/docs/set_up_tails.rst
@@ -36,9 +36,9 @@ create a bootable Tails USB drive. Follow the instructions at these
 links and then return to this page:
 
 -  `Download and verify the Tails
-   .iso <https://tails.boum.org/download/index.en.html>`__
+   .iso <https://tails.boum.org/install/download/index.en.html>`__
 -  `Install onto a USB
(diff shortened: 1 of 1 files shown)


docs/set_up_tails.rst
```diff
@@ -38,7 +38,7 @@ links and then return to this page:
 -  `Download and verify the Tails
    .iso <https://tails.boum.org/download/index.en.html>`__
 -  `Install onto a USB
-   drive <https://tails.boum.org/doc/first_steps/installation/index.en.html>`__
+   drive <https://tails.boum.org/install/index.en.html>`__
```
redshiftzero: Looks good, but replacing this link would make both the "Download and verify" (in the line above) and the "Install onto a USB drive" links go to the same place (the first link redirects to the change you have in this diff). Can we replace the first link ("Download and verify the Tails .iso") with https://tails.boum.org/install/download/index.en.html ?

ultimatecoder:
@redshiftzero  I have updated the docs according to your comment. Please review the changes again. Thanks :relaxed:

ultimatecoder:
@redshiftzero  Thanks for reviewing :blush:

Issue #3891, opened 2018-10-22
Update Ansible to 2.4.5 or later


emkll:
## Description

Ansible 2.4.2 is used in:
* https://github.com/freedomofpress/securedrop/blob/develop/securedrop/requirements/develop-requirements.txt
* https://github.com/freedomofpress/securedrop/blob/develop/admin/requirements.txt

There is currently a CVE associated to this version of Ansible: CVE-2018-10855:

>Ansible 2.5 prior to 2.5.5, and 2.4 prior to 2.4.5, do not honor the no_log task flag for failed tasks. When the no_log flag has been used to protect sensitive data passed to a task from being logged, and that task does not run successfully, Ansible will expose sensitive data in log files and on the terminal of the user running Ansible.

Securedrop does not use `no_log`, and as such this vulnerability does not directly affect Securedrop. Updating Ansible will be required for CI to pass safety checks.

## User Stories

As an admin, packages that are updated and do not have CVEs associated to them is good.

ultimatecoder:
Ansible 2.4 is EOL [source](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#release-status). Shouldn't it be upgraded to the lowest supported version which is [2.5.9](https://github.com/ansible/ansible/blob/stable-2.5/changelogs/CHANGELOG-v2.5.rst)?
Ansible v2.5.9 is greater than v2.5.5 as mentioned at [CVE-2018-10855](https://www.cvedetails.com/cve/CVE-2018-10855/).

redshiftzero:
Good flag @ultimatecoder, we should indeed update to at least the lowest supported version

ultimatecoder:
@redshiftzero Thanks!

PR #3943, opened 2018-11-27
ossec: set permissions on gpg homedir and contents

### LINKED ISSUE #3928: Rotating ossec key fails due to directory permissions on /var/ossec on mon server
## Description

If the ossec alert public key is changed on the admin workstation after an initial install, and `./securedrop-admin install` is run again, the install/update will fail with an Ansible error related to  directory permissions on `/var/ossec/.gnupg`

## Steps to Reproduce
with SD release 0.10.0:
- install 0.10.0 on app and mon servers via usual flow, including ossec setup
- change the file `~/Persistent/securedrop/install-files/ansible_base/ossec.pub` to a new pubkey
- run `./securedrop-admin install` again

## Expected Behavior
Installation completes without error

## Actual Behavior
Installation fails, as ossec user on the mon server doesn't have permission to write to `/var/ossec/.gnupg`

redshiftzero:
## Status

Ready for review

## Description of Changes

In bug #3928, problems with the permissions on the OSSEC gpg
keyring resulted in an error when attempting to import the OSSEC
public key.

This commit adds Ansible tasks to set the expected permissions
on the gpg homedir and its contents prior to attempting to import
the key

Fixes #3928.

note: some of this syntax i.e. `with_items` will need to get changed as part of #3891 

## Testing

0. Follow STR in #3928 and ensure the error does not occur

1. Test that fresh installs continue to work with this change 

2. Verify that the permissions I claim to be "the expected permissions" are indeed the right permissions

## Deployment

People are hitting this bug in prod when running `securedrop-admin install`, this will resolve as part of the workstation update in 0.11

## Checklist

[heads up did not get a chance to run tests locally, relying on CI and manual testing]

### DIFF
--- a/install_files/ansible-base/roles/ossec/defaults/main.yml
+++ b/install_files/ansible-base/roles/ossec/defaults/main.yml
@@ -21,3 +21,9 @@ ossec_is_client: False
 journalist_alert_gpg_public_key: ""
 journalist_gpg_fpr: ""
 journalist_alert_email: ""
+
+# These files should be created once an OSSEC key is imported.
+gpg_keyring_files:
+  - pubring.gpg
+  - secring.gpg
+  - trustdb.gpg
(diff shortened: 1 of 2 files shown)


redshiftzero:
thanks for review!

PR #3958, opened 2018-12-04
Set reply UUID

### LINKED ISSUE #3957: allow clients to set UUID for a reply
## Description

Currently we return the UUID of a reply after the reply has been stored. In order for clients with spotty connectivity to be able to store a reply locally and then confirm it has been stored by the server, they need to have some sort of ID that is sent to the server and then returned. We might as well make this the UUID.

## User Stories

As a developer working on the client, I want to be able to send a bunch of replies over a very slow / spotty connection and know which ones were stored and which ones failed.

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #3957

Allow API consumer to set the `Reply.uuid` in the request to allow clients to easily track in-flight message in their own cache/db.

## Testing

```bash
make test TESTFILES=tests/test_journalist_api.py::test_set_reply_uuid
```

Also play with the api some?

## Deployment

This is a non-breaking change to the API.

### DIFF
--- a/docs/development/journalist_api.rst
+++ b/docs/development/journalist_api.rst
@@ -357,6 +357,9 @@ Requires authentication. Clients are expected to encrypt replies prior to
 submission to the server. Replies should be encrypted to the public key of the
 source.
 
+Including the ``uuid`` field in the request is optional. Clients may want to
+pre-set the ``uuid`` so they can track in-flight messages.
+
 .. code:: sh
 
   POST /api/v1/sources/<source_uuid>/replies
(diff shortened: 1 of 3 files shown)


docs/development/journalist_api.rst
```diff
@@ -357,6 +357,9 @@ Requires authentication. Clients are expected to encrypt replies prior to
 submission to the server. Replies should be encrypted to the public key of the
 source.
 
+Including the ``uuid`` field in the request is optional. Clients may want to
+pre-set the ``uuid`` so the can track in-flight messages.
```
redshiftzero: Spelling mistake: `the` -> `they`

PR #3980, opened 2018-12-10
Remove unused sshd config options

### LINKED ISSUE #3979: Cleanup sshd config
## Description
As part of the removal of 2FA for interactive logins #3247, further cleanup is required for the sshd configuration on SecureDrop instances:

1. ChallengeResponseAuthentication is no longer required, as it was necessary for the 2FA PAM module.
2. UsePAM is no longer required, as no other PAM modules are used
3. PasswordAuthentication should be set to no (ChallengeResponseAuthetnication set to yes would override PasswordAuthentication set to no)

emkll:
## Status

Ready for review

## Description of Changes

Fixes #3979 

Disables `ChallengeResponseAuthentication`, `UsePam`, and `PasswordAuthentication` in sshd config. 
Updates the builder hash (security updates were required for passing tests).

#### Clean install scenario
- [ ] Ensure sshd configuration is as described (see tests)
- [ ] Ensure you can log in via ssh, through either tor or local network
- [ ] Ensure ssh access via Ansible works after reboot/bouncing sshd service (e.g. `./securedrop-admin backup`)

#### Upgrade install scenario
Use upgrade testing guide https://docs.securedrop.org/en/release-0.10.0/development/upgrade_testing.html or manually perform a clean install on 0.10.0 and install debs build from this branch. Confirm the following:
- [ ] Debs install successfully
- [ ] Ensure sshd configuration is as described (see tests, perform additional visual review of `/etc/ssh/sshd_config` to ensure is is identical to the one provisioned in clean installs of this branch.
- [ ] Ensure you can log in via ssh, through either tor or local network
- [ ] Ensure ssh access via Ansible works after reboot/bouncing sshd service (e.g. `./securedrop-admin backup`)

## Deployment

Existing production will see their sshd configuration change by way of the `securedrop-config` package postinst script as part of the next SecureDrop release.

2. New installs.

Sshd config for new installs will be provisioned at install-time via Ansible.

### DIFF
--- a/install_files/securedrop-config/DEBIAN/postinst
+++ b/install_files/securedrop-config/DEBIAN/postinst
@@ -13,6 +13,12 @@ remove_2fa_tty_req() {
     # Lets prevent this from bombing out the install though if it fails
     auth_file=/etc/pam.d/common-auth
     sed -i "/^auth\ required\ pam_google.*/d" ${auth_file} || true
+    # Since we are removing the pam_google module, we must ensure password
+    # authentication for sshd is disabled to ensure only key authentication is used.
+    grep -qF "PasswordAuthentication no" /etc/ssh/sshd_config || echo "PasswordAuthentication no" >> /etc/ssh/sshd_config
+    sed -i "/^UsePAM\ /s/\ .*/\ no/" /etc/ssh/sshd_config
+    sed -i "/^ChallengeResponseAuthentication\ /s/\ .*/\ no/" /etc/ssh/sshd_config
+    service ssh restart
 }
 
 case "$1" in
--- a/install_files/ansible-base/roles/restrict-direct-access/templates/sshd_config
+++ b/install_files/ansible-base/roles/restrict-direct-access/templates/sshd_config
@@ -18,8 +18,7 @@ IgnoreRhosts yes
 RhostsRSAAuthentication no
 HostbasedAuthentication no
 PermitEmptyPasswords no
-# This is needed for libpam-google-authenticator package for 2fa SSH access.
-ChallengeResponseAuthentication yes
+ChallengeResponseAuthentication no
 KerberosAuthentication no
 KerberosGetAFSToken no
 GSSAPIAuthentication no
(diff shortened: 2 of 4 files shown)


install_files/securedrop-config/DEBIAN/postinst
```diff
@@ -13,6 +13,12 @@ remove_2fa_tty_req() {
     # Lets prevent this from bombing out the install though if it fails
     auth_file=/etc/pam.d/common-auth
     sed -i "/^auth\ required\ pam_google.*/d" ${auth_file} || true
+    # Since we are removing the pam_google module, we must ensure password
+    # authentication for sshd is disabled to ensure only key authentication is used.
+    grep -qF "PasswordAuthentication no" /etc/ssh/sshd_config || echo "PasswordAuthentication no" >> /etc/ssh/sshd_config
+    sed -i "/^UsePAM\ /s/\ .*/\ no/" /etc/ssh/sshd_config
+    sed -i "/^ChallengeResponseAuthentication\ /s/\ .*/\ no/" /etc/ssh/sshd_config
+    service ssh restart
```
redshiftzero: sorry, just noticed this - is using service ssh restart in postinst the right practice here? https://www.debian.org/doc/debian-policy/ch-opersys.html#running-initscripts says we should be using invoke-rc.d

(note this did work for me as expected during testing, just wondering if we should change it to the recommendation)

emkll: Good point @redshiftzero , thanks for the link. From my understanding of that document, the intent of the guideline is for packagers to be mindful of other services and dependencies of these services on the host, to ensure all services change state graciously. Given that the state of services on SecureDrop instances is known (all SD instances should have the same services running), this seems less important in this context, but happy to make the change. Note that we do already use this method for apache: https://github.com/freedomofpress/securedrop/blob/develop/install_files/securedrop-app-code/DEBIAN/postinst#L119

redshiftzero: ok thanks for the explanation, since we (~)know the state of running SecureDrops, let's leave as is

emkll:
Confirmed, tests were failing for me locally as well (they were passing only a few hours ago). I've pushed an updated to the container image (and changed the hash in this PR). Deb build tests should be passing now.

PR #3994, opened 2018-12-17
Avoid leaking information in GPG key creation and expiration dates

### LINKED ISSUE #3912: Expiration date on PGP key leaks date of source's first contact
## Description

When we autogen PGP keys, the expiration is set to 1 year. This can be used to infer the date a source first used SD. We may want to remove this expiration field altogether to get around this issue.

garrettr:
## Status

Ready for review

## Description of Changes

Fixes #3912.

1. Use a fixed date for the creation time of each source's GPG reply key, to avoid leaking the approximate date at which they first created a SecureDrop account, per discussion in https://github.com/freedomofpress/securedrop/issues/3912#issuecomment-447719726.
2. Reply keypairs should not expire, per discussion in https://github.com/freedomofpress/securedrop/issues/3912#issuecomment-447716901

## Testing

I added a unit test, `tests/test_crypto_util.py::test_genkeypair_should_use_obfuscated_creation_and_expiration_dates`. 

I confirmed the test fails on develop and passes with the changes made in this PR.

## Deployment

Filed #3995 to address this issue.

2. New installs.
No special considerations that I can think of. I don't know of any places where we use the creation/expiration dates of the reply keys in SecureDrop, so I think this should be a fairly low-risk change.

I was unable to run `make ci-lint` on macOS, got an error:
```
devops/scripts/dev-shell-ci run make --keep-going lint typelint
devops/scripts/dev-shell-ci: line 23: sha256sum: command not found
make: *** [ci-lint] Error 127
```
File #3996 to address this issue.

### DIFF
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -49,6 +49,10 @@ class CryptoUtil:
     # https://www.newyorker.com/news/news-desk/strongbox-and-aaron-swartz
     DEFAULT_KEY_CREATION_DATE = date(2013, 5, 14)
 
+    # '0' is the magic value that tells GPG's batch key generation not
+    # to set an expiration date.
+    DEFAULT_KEY_EXPIRATION_DATE = '0'
+
     def __init__(self,
                  scrypt_params,
                  scrypt_id_pepper,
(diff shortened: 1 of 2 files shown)


garrettr:
> I'd like someone who remembers how to pick apart the PGP packets to take a look to make sure there's not a date that's embedded in there somewhere that we're not considering.

@heartsucker Good idea. Have you tried https://github.com/kazu-yamamoto/pgpdump?

heartsucker:
Tests need to be kicked, but this is good to merge. For whoever merges, we need to open the followup ticket to migrate keys on old instances to erase the expiration date.

redshiftzero:
This looks good to me, @emkll take a peek at this and let me know if you are cool with this change (doesn't need to be a full review, just so you are aware)

PR #4011, opened 2019-01-02
cron-apt remove action should be after security

### LINKED ISSUE #4003: Broken packages block updates due to cron-apt run order
## Description

The nightly `cron-apt` operations to install updates consist of three discrete actions, executed in the following order:

1. update apt lists for whitelisted repos
2. remove unwanted distro-maintained kernels (leaving only SD-specific hardened kernels)
3. install any new package versions available in whitelisted repos

If any packages fail to install, _all_ future updates will fail to install. 

## Steps to Reproduce

At a high level, steps to reproduce are:

1. Install a "broken" package (one that errors out at e.g. the `postinst` stage)
2. Attempt to upgrade to a newer, fixed version of same package via cron-apt
3. Observe failure to upgrade

More specifically, a practical scenario can be created with:

1. Set up `upgrade` scenario VMs based on 0.9.1. 
2. Edit `/etc/hosts` and duplicate the `mon` IP line a few times.
3. Run `cron-apt -i -s` to install debs at version 0.10.0 or newer.
4. Observe failure.
5. Build patched/fixed debs for e.g. 0.12.0~rc1 and copy to local apt server (in `upgrade` scenario)
6. Run `cron-apt -i -s` on the host with a broken package
7. Observe that the newer 0.12.0~rc1 is _not_ installed, and the system is still in a broken state.

## Expected Behavior
Updates to broken packages are applied.

## Actual Behavior
Updates are never installed if a package is in a broken state.b

## Comments

If we flip steps 2 & 3 in the cron-apt flow, i.e. install updates immediately after updating the lists, we should sidestep this problem.

emkll:
## Status

Ready for review
## Description of Changes

Fixes #4003 .

In the event the host has broken packages installed (e.g. failed postinst), cron-apt will fail at the remove action before upgrading the package. Upgrades to the broken package will not be installed and the system will remain in a broken state: 

Change `1-remove` to `9-remove` in cron-apt actions for Ansible install as well as in securedrop-config postinst script (and delete `1-remove` via the same)

### Clean install scenario:
- [ ] Successfully provision staging or prod environment on this branch
- [ ] `/etc/cron-apt/action.d/` contains files as expected (0-update 5-security and 9-remove)
- [ ] `sudo cron-apt -i -s` completes without error

### Upgrade scenario 
Using (https://docs.securedrop.org/en/release-0.10.0/development/upgrade_testing.html):
* Checkout this branch
* Install SecureDrop <=0.10.0:
  * `make build-debs`
  * `molecule converge -s upgrade`
  * copy and install `securedrop-config` deb package produced by this branch
* Insert duplicate entry for mon in `/etc/hosts` (either app or mon)
* Attempt an upgrade to 0.11.0:
  * `/etc/apt/security.list` should contain apt.freedom.press
  * run `cron-apt -i -s` and observe ossec failing to install
* Attempt an upgrade to 0.12~rc1:
  * `/etc/apt/security.list` should contain apt-test.freedom.press
  * run `cron-apt -i -s` and observe ossec failing to install

- [ ] Install of ossec fails, but other packages are still upgraded (e.g. `apt list --installed | grep securedrop-app-code` should return that version of securedrop-app-code installed is 0.12.0~rc1

## Deployment

- Existing installs will be updated via `securedrop-config` apt package
- New installs will be provisioned by the correct config via Ansible


conorsch:
Proceeding with review of the upgrade scenario. Using a slightly modified test plan from what's in the OP, so documenting here for clarity:

## Upgrade to 0.9.1 -> 0.11.0 -> 0.12.0~rc1

* [ ] Check out this branch
* [ ] `make build-debs` (so the local debs used by the upgrade scenario include the changes presented here)
* [ ] `echo "0.9.1" > molecule/shared/stable.ver` (to force use of 0.9.1 boxes as upgrade target)
* [ ] `make upgrade_start` (to prep boxes on version 0.9.1)
* [ ] Confirm "0.9.1" version string on Source Interface
* [ ] Insert duplicate entry for mon in /etc/hosts on app-staging host
* [ ] Run `sudo apt-get install --only-upgrade securedrop-config` (to get latest deb from this branch, via local apt)
* [ ] Comment out override for apt.freedom.press in /etc/hosts on app-staging host (to use prod, not local, apt server)
* [ ] Run `sudo rm -r /var/lib/apt/lists` to purge local apt versions from apt cache
* [ ] Manually run `sudo cron-apt -i -s` inside app-staging host (to pull 0.11.0 packages via prod apt)
* [ ] Observe failure on securedrop-ossec-agent package upgrade in cron-apt output
* [ ] Confirm "0.11.0" version string on Source Interface (:tada:!)
* [ ] Re-run `sudo rm -r /var/lib/apt/lists` to purge prod apt versions from apt cache
* [ ] Re-enable override for apt.freedom.press in /etc/hosts on app-staging host (to use local, not prod, apt server)
* [ ] Re-run `sudo cron-apt -i -s` inside app-staging host (to pull 0.12.0~rc1 packages via local apt)
* [ ] Observe failure on securedrop-ossec-agent package upgrade in cron-apt output again
* [ ] Confirm "0.12.0~rc1" version string on Source Interface (:tada:!)

Issue #4013, opened 2019-01-03
[xenial] Journalist replies are not displayed in source interface


emkll:
## Steps to Reproduce

* Spin up a Xenial staging environment (`make build-debs-xenial` and `molecule converge -s libvirt-staging-xenial`)
* Submit a document or message from the source interface
* Reply to said document or message
* Observe empty reply on the source interface

## Expected Behavior

Source should be able to view journalist reply
## Actual Behavior

Journalist reply is blank in source interface

## Comments

Several changes were made in gpg 2.1 which is being used by Xenial [0], specifically the removal of the secret keyring

[0] : https://www.gnupg.org/faq/whats-new-in-2.1.html

redshiftzero:
I took a look at this today, here's where I'm at so far: I can reproduce this bug in a Xenial-based SecureDrop dev container. 

The secret keys for decrypting the replies are available, e.g.:

```
>>> self.gpg.list_keys(secret=True)
[...]

Issue #3970, opened 2018-12-05
[xenial] Implement `securedrop-admin os-release-upgrade` feature


eloquence:
## Description

As part of #3204, we may want to implement an `os-release-upgrade` feature for `securedrop-admin`. This would be run on the *Admin Workstation*, and guide the administrator through the upgrade process from Trusty to Xenial.

This tool may offer to perform a backup/restore, and perform the `do-release-upgrade` step as well.

Depending on resolution to #3968, we may be able to eliminate prompts caused by the upgrade itself -- or at least provide some in-tool guidance.

## User Stories

As an administrator, I would like to have some guardrails for the system upgrade, so that the room for error on my part is reduced.

eloquence:
For now we've decided against this, since the upgrade is unlikely to be non-interactive, and will lean heavily on docs.

Issue #3661, opened 2018-07-26
Support tor-browser-selenium tests against app-staging in CI


conorsch:
## Description

The tor-browser-selenium epic (#3659) tracks work enabling the functional test suite to target "remote" servers, rather than executing Selenium within the `app-staging` VM. Once the test suite stabilizes in the `tbb-0.9.0` branch, we should modify the CI logic for the staging environment to use the new external test client. A brief outline of the work entails:

* [ ] Modify or remove `molecule/aws/side_effect.yml` to run the new functional test suite logic.
* [ ] Remove Ansible logic installing test dependencies in app-staging from `install_files/ansible-base/roles/app-test`.
* [ ] Ensure external client test logic works well in local staging, as well as CI

As we get closer to targeting this work, the outline above may change. Discuss often during team meetings to keep the outline above an accurate depiction of the planned work.

## User Stories

As a developer, I want CI to take over as much of the human-driven interaction testing as possible. 

As a maintainer, I recognize that substantial changes to testing logic will require corresponding updates to our CI framework.

PR #4015, opened 2019-01-03
Update AppArmor profile for Xenial

### LINKED ISSUE #3962: [xenial] Update AppArmor rules to work on Trusty, Xenial
Explicit rules are required for Apache `mpm` worker/event changes. `gpg2` policy should permit links via `/var/lib/securedrop/keys/* l` or similar.

For simplicity, we should update AppArmor rules to work for both Trusty and Xenial, using a single template file for both distros.

Part of #3204.

emkll:
## Status

Ready for review
## Description of Changes

Fixes #3962 . Opened #4013 as follow up, uncovered in https://github.com/freedomofpress/securedrop/issues/3962.

## Testing
Spin up a staging environment on Xenial:
- `make build-debs-xenial`
- `molecule converge -s libvirt-staging-xenial`

#### Verify there are no further AppArmor violations:
- Perform actions on source and journalist interface (for example, submit documents and reply to those documents, or run the molecule verify scenario)
- Log into the app server and confirm:
- [ ] no `apparmor="DENIED"` in `/var/log/syslog` or in `journalctl -a`

#### Verify that keys are not being monitored by syscheck:
- Log into the mon server, `sudo su `and run `./agent-control -r -a` in `/var/ossec/bin` to force a syscheck run
- Wait until `INFO: Ending syscheck scan` appears in /var/ossec/logs/ossec.log`:
- [ ] syscheck does not monitor the private keys (hashes are in `/var/ossec/queue/syscheck/(app-staging 10.0.1.2->syscheck)`

#### Ensure there is no breakage in Trusty
- `make build-debs`
- `molecule converge -s libvirt-staging`
- [ ] Ensure the application behaves as expected.

## Deployment

These settings will be provided to new and existing installs via deb packages.


emkll:
Thanks for the review @kushaldas ! 

I'm having some trouble reproducing the error you describe locally, could you please provide steps I can use to reproduce the error you've encountered?  You state that `This is broken`: I assume you mean a submission triggers error 500? Are there any AppArmor violations?

I've also updated the test plan with the steps I've used to test the Trusty scenario.

kushaldas:
> You state that `This is broken:` I assume you mean a submission triggers error 500? 

Yes, the step I took and the error message is given below in that comment.

kushaldas:
@emkll Please go ahead as my vagrant+molecule seems to be in broken state.

emkll:
Thanks for the review @conorsch, I've addressed your comments:

`. Good catch on the syscheck alert. The file you mention is a flag to indicate that the migration to GnuPG 2.1 was completed (since it's a 1-time migration) [0]. That file would produce at most a 1-time alert (which would have gotten lost in the myriad of other syscheck that might occur during an upgrade). I've nonetheless added the ignore rule in syscheck, so that will be one alert less upon upgrade.

2. I've also added a test to ensure no AppArmor violations are in syslog. I think it's a good idea, but I am not sure how effective it is in it's current form, given the lack of interaction with the application. Happy to amend that test if you have any ideas. 

[0] : https://www.gnupg.org/documentation/manuals/gnupg/GPG-Configuration.html

conorsch:
> I am not sure how effective it is in it's current form, given the lack of interaction with the application

Certainly it will be more useful once #3661 lands, but even in the near-term, devs can still do manual browser interaction, then run the verify logic to assert no denials via AppArmor. One foot in front of the other, so to speak.

> so that will be one alert less upon upgrade.

Given the cascade of OSSEC alerts the upgrade will generate, perhaps an `upgrade` target should silence OSSEC (as well as disable Apache and other pre-upgrade steps) ; the `install` target would then hook everything back up without any changes. Relevant to #3970.

Issue #3916, opened 2018-11-02
PaX flags are unset during install time


conorsch:
## Description

The grsecurity pax flag checks on the grub binaries are currently failing. Those don't run in CI—yet; see #3909—so we haven't been carefully observing them. 

## Steps to Reproduce

1. Provision local staging: `make staging`.
2. Run testinfra suite: `molecule verify -s libvirt-staging` (substitute `virtualbox-staging` if appropriate)
3. Observe PaX flag errors
4. Re-run staging logic: `molecule converge -s libvirt-staging`
5. Re-run testinfra suite:  `molecule verify -s libvirt-staging` (substitute `virtualbox-staging` if appropriate)
6. Observe all tests pass [0]

## Expected Behavior

The fresh install story configures PaX flags in a persistent manner.

## Actual Behavior

The fresh install story sets PaX flags, but the are overwritten if updates to grub are available.

## Comments

We can resolve this problem by moving the "common" role _before_ the "grsecurity" role; it should be the first in the play. We'll upgrade all packages, then after updates to grub are installed, the PaX flags will be set. 

In #2741 we moved the grsecurity role up to fail faster. That minor convenience isn't worth the risk of misconfiguration. Since #2703, we've been managing PaX flags on grub for _existing_ installs—but we still have an install time window where they can be overwritten. 

[0] Regarding "all tests pass", I actually saw _one_ still fail, but it's unrelated:

<details>
```
    =================================== FAILURES ===================================
    ____ test_securedrop_application_test_journalist_key[ansible://app-staging] ____
    [gw3] linux2 -- Python 2.7.13 /home/conor/.virtualenvs/sd/bin/python2
    
    File = <class 'testinfra.modules.base.GNUFile'>, Sudo = <sudo>
    
        def test_securedrop_application_test_journalist_key(File, Sudo):
            """
            Ensure the SecureDrop Application GPG public key file is present.
            This is a test-only pubkey provided in the repository strictly for testing.
            """
            pubkey_file = File("{}/test_journalist_key.pub".format(
                securedrop_test_vars.securedrop_data))
            # Sudo is only necessary when testing against app hosts, since the
            # permissions are tighter. Let's elevate privileges so we're sure
            # we can read the correct file attributes and test them.
            with Sudo():
                assert pubkey_file.is_file
    >           assert pubkey_file.user == "root"
    E           AssertionError: assert 'www-data' == 'root'
    E             - www-data
    E             + root
    
[...]

PR #4021, opened 2019-01-07
Remove common-auth pam customizations

### LINKED ISSUE #3963: [xenial] Investigate if PAM common-auth customizations are still needed
The PAM `common-auth` customizations include declarations for `pam_ecryptfs.so` which prove problematic on Xenial; commenting out appears to resolve. More research required. We may be able to remove these additions entirely, now that we no longer support 2FA via challenge-response. 

Part of #3204.

emkll:
## Status

Ready for review
## Description of Changes
PAM customizations were necessary to allow 2FA for console logins. Since, these configurations are no longer necessary due to the phasing out of 2FA for console logins and `/var/log/auth.log` in trusty and `syslog` in xenial, (encryptfs.so), let's replace the FPF-configuredpam.d/common-auth file with the upstream-maintained common-auth.

Closes #3963 for new installs or installs on which `./securedrop-admin install` is run. 

## Testing
- [ ]  Ensure `/etc/pam.d/common-auth` provided by this PR is consistent with the upstream version(s) in Ubuntu 14.04 and 16.04.

#### Clean install
- [ ] Provision a Trusty staging environment (e.g. `molecule converge -s libvirt-staging`) and ensure SSH access is preserved on app and mon servers.
- [ ] Provision a Xenial staging environment (e.g. `molecule converge -s libvirt-staging-xenial`) and ensure SSH access is preserved on app and mon servers.
- [ ] Error message `/lib/security/pam_encryptfs.so: cannot open shared object file: No such file or directory` does not appear in `/var/log/auth.log` (Trusty) or `journalctl -e` (Xenial) does not contain after running the playbooks.

#### Upgrade testing
- `git checkout develop`
- Provision a Prod environment on `develop` (using`./securedrop-admin install`)
- Check out this branch and re-run `./securedrop-admin install`
- [ ] Ensure ssh access is preserved on app and mon servers.
- [ ] Error message `/lib/security/pam_encryptfs.so: cannot open shared object file: No such file or directory` does not appear in `/var/log/auth.log` (Trusty) or `journalctl -e` (Xenial) does not contain after running the playbooks.

## Deployment

Since there is no negative impact on running instances (Trusty or Xenial) other than errors in the logs, let's minimize the risk to automatically patch the file on running instances via postinst, and instead only ship this change via Ansible:

1. Existing production installs will be updated via Ansible
2. New installs via Ansible.

### DIFF
--- a/install_files/ansible-base/roles/restrict-direct-access/files/common-auth
+++ b/install_files/ansible-base/roles/restrict-direct-access/files/common-auth
@@ -1,4 +1,26 @@
-auth    [success=1 default=ignore]      pam_unix.so nullok_secure
-auth    requisite                       pam_deny.so
-auth    required                        pam_permit.so
-auth    optional                        pam_ecryptfs.so unwrap
+#
+# /etc/pam.d/common-auth - authentication settings common to all services
+#
+# This file is included from other service-specific PAM config files,
+# and should contain a list of the authentication modules that define
+# the central authentication scheme for use on the system
+# (e.g., /etc/shadow, LDAP, Kerberos, etc.).  The default is to use the
+# traditional Unix authentication mechanisms.
+#
+# As of pam 1.0.1-6, this file is managed by pam-auth-update by default.
+# To take advantage of this, it is recommended that you configure any
+# local modules either before or after the default block, and use
+# pam-auth-update to manage selection of other modules.  See
+# pam-auth-update(8) for details.
+
+# here are the per-package modules (the "Primary" block)
+auth	[success=1 default=ignore]	pam_unix.so nullok_secure
+# here's the fallback if no module succeeds
+auth	requisite			pam_deny.so
+# prime the stack with a positive return value if there isn't one already;
+# this avoids us returning an error just because nothing sets a success code
+# since the modules above will each just jump around
+auth	required			pam_permit.so
+# and here are more per-package modules (the "Additional" block)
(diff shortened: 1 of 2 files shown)


conorsch:
Started reviewing this, haven't run through all the steps yet. On the surface this looks like exactly what what we want. On clean Trusty & Xenial VMs (i.e. before any SD-specific customizations are applied), the `/etc/pam.d/common-auth` files match what's presented here, with the small exception of `pam_cap` missing from the Trusty config, which was already discussed in #3963.

After verifying the rest of the functionality as described, I foresee two small additions:

* config test to check that `ecryptfs` does not appear in `/etc/pam.d/common-auth`
* config test to check that the log message does not appear in `/var/log/auth.log`

Before adding the latter test in particular, I'll a) confirm that the log message readily appears on develop and b) paste example log lines into #3963 to use in structure the test.

conorsch:
Thanks for the tests and the rebase, @emkll. Verified that there are no unwanted side-effects on the Trusty story, and all config tests are passing locally.

For Xenial, SSH functionality is working as expected, and all the config tests are passing (except #3916, since I ran several provisioning commands). However, I'm I notice that auth.log contains:

```
/lib/security/pam_cap.so: cannot open shared object file: No such file or directory
```

Given that the pam_cap include was taken from the upstream config, and no problems were observed with SSH access, I suggest we ignore this for now and move forward with the other migration tasks.

Issue #3877, opened 2018-10-15
Apache is removing ETag headers


redshiftzero:
## Description

We send ETags in responses from a couple of journalist API endpoints ([reference](https://docs.securedrop.org/en/release-0.9/development/journalist_api.html#download-a-submission-get)), but Apache is stripping them off via `Header unset Etag`

## Steps to Reproduce

0. Provision staging servers
1. Add a test source, a test submission, and a test journalist user
2. Authenticate to the API via submitting your credentials to `/api/v1/token` as described [here](https://docs.securedrop.org/en/release-0.9/development/journalist_api.html#authentication)
3. Try to download a response via a GET request to `/api/v1/sources/<source_uuid>/submissions/<submission_uuid>/download` where `source_uuid` and `submission_uuid` are valid

## Expected Behavior

response has `ETag` header 

## Actual Behavior

response does not have `ETag` header 

## Comments

This is similar to #3772 so a similar fix will do here (modifying Apache configs in `postinst` of `securedrop-app-code` package). In terms of priority, this is not a showstopping bug (users can still download files via the API without issue), so 0.11.0 or a bugfix release after 0.10.0 is fine.

redshiftzero:
When this bug is addressed, we can revert https://github.com/freedomofpress/securedrop-sdk/commit/6170a337914cb34da0286da0f0ed79c0657f46bd in the SDK

heartsucker:
From that ticket:

> ETags are being stripped from staging/production servers due to an Apache misconfiguration.

I am rather sure this was intentional and was done as part of the no-cache headers to prevent anything from being left on a source's computer.

redshiftzero:
No you're right @heartsucker, we should only modify the journalist Apache config

emkll:
Recapping discussion in gitter with @heartsucker @redshiftzero 
To close this item ticket, we should simply remove `Unset Etag` directive from the Journalist interface config, which should allow for the values that are set in the Application/API to be sent to the client and not stripped by Apache.
This configuration change should be sufficient to report download resumption as well (due to Range RequestHeader not being unset).

redshiftzero:
This was closed by #4023

PR #4023, opened 2019-01-08
Allow DELETE HTTP method for journalist interface

### LINKED ISSUE #3977: DELETE methods not allowed by apache journalist interface config
## Description

This is the issue causing https://github.com/freedomofpress/securedrop-client/issues/173

## Steps to Reproduce

1. Provision staging VMs
2. Get a valid API token [[ref]](https://gist.github.com/redshiftzero/934093c4f8e0f47c53e1add475345e34) 
3. Add a source via the source interface
4. Attempt to remove a star from a source (this is an idempotent operation so one should not need to first add the star to hit the endpoint):

```
curl -X DELETE -H "Content-Type: application/json" -H "Authorization: Token mytokenhere" --proxy socks5h://127.0.0.1:9150 mystagingserver.onion/api/v1/sources/my-source-uuid-here/remove_star
```

## Expected Behavior

`{"message":"Star removed"}`

## Actual Behavior

A 403 occurs, and we get redirected to the login page. 

This redirect-for-403 behavior only happens for the API in staging because Apache is [sending 403s to `/notfound`](https://github.com/freedomofpress/securedrop/blame/eaa2d45d371b64d5bb82317209d686c63063fec8/install_files/ansible-base/roles/app/templates/sites-available/journalist.conf#L42), causing the [`request.path`](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/journalist_app/__init__.py#L136) to be `/notfound` (instead of something that begins with `thisismeonion.onion/api/blahhh`), meaning that code execution will continue until [the redirect here](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/journalist_app/__init__.py#L140).

## Comments

The cause is the 403, and this is happening because of the [LimitExcept](https://httpd.apache.org/docs/current/mod/core.html#limitexcept) directive we are using in the Apache configs. These allow only GET POST and HEAD requests.

To prove this to oneself, you can edit `/etc/apache2/sites-enabled/journalist.conf` and add DELETE as an allowed request method.

This is similar to bugs #3772 and #3877. In light of this we should:

1. Revisit #1775 and,
2. Incorporate some API endpoint testing which would have detected all three bugs into at least the pre-release QA process (these could eventually be in CI as part of the external test client described in #3661)

emkll:
## Status

Ready for review 
## Description of Changes

Fixes #3977 , #3877 

Certain Journalist API operations require DELETE (e.g. delete or unstar a source). This will ensure that Apache allows the DELETE operation for new installs (or existing installs after an Ansible run).

Furthermore, this will disable Apache stripping Etag headers on the journalist interface in order to receive Etag information from the API (containing sha256 sum of file for integrity). Currently all Etag headers should be `sha256sum:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855` due to a bug tracked in #4032 .

From the test plan below, it is very clear that we need some integration tests in order to validate these configurations (tracked #4030)
#### Clean install scenario
- Check out this branch
- ./securedrop-admin install
- [ ] install completed without error
- [ ] `/etc/apache2/sites-enabled/journalist.conf` contains expected changes (or run testinfra suite)
- Create admin user for the app
- Navigate to the source interface and submit a document/message
- Attempt to delete source via API (see below for example commands)
- [ ] The deletion was successful, DELETE method was allowed by Apache.
#### Upgrade scenario

- Install SecureDrop 0.12.1
- Check out this branch
- ./securedrop-admin install
- [ ] install completed without error
- [ ] `/etc/apache2/sites-enabled/journalist.conf` contains expected changes (or run testinfra suite)
- Create admin user for the app
- Navigate to the source interface and submit a document/message
- Attempt to delete source via API

#### Example test commands for api (curl)
0. Ensure aths token is in torrc
1.  Get auth token:
```
torify curl -X POST \
  http://$JOURNALIST_INTERFACE.onion/api/v1/token \
  -H 'Content-Type: application/json' \
  -d '{
    "username": "$USERNAME", 
    "passphrase": "$PASSPHRASE", 
    "one_time_code": "$TOTP"
}'
```
2. Use the above token to get list of sources (to obtain `source_uuid`):
```
torify curl -X GET \
  http://$JOURNALIST_INTERFACE.onion/api/v1/sources \
  -H 'Authorization: Bearer $YOUR_TOKEN_GOES_HERE' \
  -H 'Content-Type: application/json' 
```
3. Get the source based on `source_uuid` obtained above:
```
torify curl -X GET \
  http://$JOURNALIST_INTERFACE.onion/api/v1/sources/$SOURCE_UUID \
  -H 'Authorization: Bearer $YOUR_TOKEN_GOES_HERE' \
  -H 'Content-Type: application/json' 
```
4. Get the download URL of a file from the`source_uuid` :
```
torify curl -O \
  http://$JOURNALIST_INTERFACE.onion/api/v1/sources/$SOURCE_UUID/submissions/$SUBMISSION_UUID/download \
  -H 'Authorization: Bearer $YOUR_TOKEN_GOES_HERE' \
  -H 'Content-Type: application/json' 
```
5. Get the headers/Etag value :
```
torify curl -I \
  http://$JOURNALIST_INTERFACE.onion/api/v1/sources/$SOURCE_UUID/submissions/$SUBMISSION_UUID/download \
  -H 'Authorization: Bearer $YOUR_TOKEN_GOES_HERE' \
  -H 'Content-Type: application/json' 
```
Should return the expected Etag value in the headers:
```
HTTP/1.1 200 OK
Date: Mon, 14 Jan 2019 19:22:02 GMT
Server: Apache
Content-Disposition: attachment; filename=1-better_wrap-msg.gpg
Cache-Control: public, max-age=43200
Expires: Tue, 15 Jan 2019 07:22:02 GMT
X-Frame-Options: DENY
Etag: "sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
Cache-Control: no-store
Referrer-Policy: no-referrer
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
X-Download-Options: noopen
X-Content-Security-Policy: default-src 'self'
Content-Security-Policy: default-src 'self'
Last-Modified: Mon, 14 Jan 2019 18:54:00 GMT
Content-Length: 592
Content-Type: application/pgp-encrypted
```
6. Delete the source:
```
curl -X DELETE \
  http://$JOURNALIST_INTERFACE.onion/api/v1/sources/$SOURCE_UUID \
  -H 'Authorization: Bearer $YOUR_TOKEN_GOES_HERE' \
  -H 'Content-Type: application/json' 
```
7. Ensure the source is deleted:
```
curl -X GET \
  http://$JOURNALIST_INTERFACE.onion/api/v1/sources \
  -H 'Authorization: Bearer $YOUR_TOKEN_GOES_HERE' \
  -H 'Content-Type: application/json' 
```
## Deployment
Due to the risks associated with modifying the Apache configuration and due to this change being opt-in for users of the SecureDrop Workstation, this change will initially require an Ansible run to be applied.

1. New installs: Via Ansible / securedrop-admin install
2. Existing installs/upgrades: Via Ansible / securedrop-admin install

### DIFF
--- a/install_files/ansible-base/roles/app/templates/sites-available/journalist.conf
+++ b/install_files/ansible-base/roles/app/templates/sites-available/journalist.conf
@@ -36,13 +36,6 @@ Header unset Etag
 # Limit the max submitted size of requests.
 LimitRequestBody 524288000
 
-#Redirect error pages to ensure headers are sent
-ErrorDocument 400 /notfound
-ErrorDocument 401 /notfound
-ErrorDocument 403 /notfound
-ErrorDocument 404 /notfound
-ErrorDocument 500 /error
-
 <Directory />
   Options None
   AllowOverride None
--- a/install_files/ansible-base/roles/app/templates/sites-available/journalist.conf
+++ b/install_files/ansible-base/roles/app/templates/sites-available/journalist.conf
@@ -53,11 +53,11 @@ ErrorDocument 500 /error
 <Directory /var/www/>
   Options {{ apache_dir_options | default('None') }}
   AllowOverride None
-  <Limit GET POST HEAD>
+  <Limit GET POST HEAD DELETE>
     Order allow,deny
     allow from {{ securedrop_app_apache_allow_from }}
   </Limit>
-  <LimitExcept GET POST HEAD>
+  <LimitExcept GET POST HEAD DELETE>
     Order deny,allow
     Deny from all
   </LimitExcept>
@@ -66,11 +66,11 @@ ErrorDocument 500 /error
 <Directory /var/www/securedrop>
   Options {{ apache_dir_options | default('None') }}
   AllowOverride None
-  <Limit GET POST HEAD>
+  <Limit GET POST HEAD DELETE>
     Order allow,deny
     allow from {{ securedrop_app_apache_allow_from }}
   </Limit>
-  <LimitExcept GET POST HEAD>
+  <LimitExcept GET POST HEAD DELETE>
     Order deny,allow
     Deny from all
   </LimitExcept>
--- a/install_files/ansible-base/roles/app/templates/sites-available/journalist.conf
+++ b/install_files/ansible-base/roles/app/templates/sites-available/journalist.conf
@@ -31,7 +31,6 @@ Header set X-Content-Type-Options: nosniff
 Header set X-Download-Options: noopen
 Header set X-Content-Security-Policy: "default-src 'self'"
 Header set Content-Security-Policy: "default-src 'self'"
-Header unset Etag
 
 # Limit the max submitted size of requests.
 LimitRequestBody 524288000
--- a/molecule/testinfra/staging/app/apache/test_apache_journalist_interface.py
+++ b/molecule/testinfra/staging/app/apache/test_apache_journalist_interface.py
@@ -16,7 +16,6 @@ wanted_apache_headers = [
   "Header set X-Content-Security-Policy: \"default-src 'self'\"",
   "Header set Content-Security-Policy: \"default-src 'self'\"",
   'Header set Referrer-Policy "no-referrer"',
-  'Header unset Etag',
 ]
 
 
--- a/molecule/testinfra/staging/app/apache/test_apache_journalist_interface.py
+++ b/molecule/testinfra/staging/app/apache/test_apache_journalist_interface.py
@@ -51,11 +51,11 @@ common_apache2_directory_declarations = """
 <Directory /var/www/>
   Options None
   AllowOverride None
-  <Limit GET POST HEAD>
+  <Limit GET POST HEAD DELETE>
     Order allow,deny
     allow from {apache_allow_from}
   </Limit>
-  <LimitExcept GET POST HEAD>
+  <LimitExcept GET POST HEAD DELETE>
     Order deny,allow
     Deny from all
   </LimitExcept>
@@ -64,11 +64,11 @@ common_apache2_directory_declarations = """
 <Directory {securedrop_code}>
   Options None
   AllowOverride None
-  <Limit GET POST HEAD>
+  <Limit GET POST HEAD DELETE>
     Order allow,deny
     allow from {apache_allow_from}
   </Limit>
-  <LimitExcept GET POST HEAD>
+  <LimitExcept GET POST HEAD DELETE>
     Order deny,allow
     Deny from all
   </LimitExcept>
--- a/securedrop/journalist_app/api.py
+++ b/securedrop/journalist_app/api.py
@@ -309,7 +309,7 @@ def make_blueprint(config):
         user = get_user_object(request)
         return jsonify(user.to_json()), 200
 
-    def _handle_http_exception(error):
+    def _handle_api_http_exception(error):
         # Workaround for no blueprint-level 404/5 error handlers, see:
         # https://github.com/pallets/flask/issues/503#issuecomment-71383286
         response = jsonify({'error': error.name,
@@ -318,6 +318,6 @@ def make_blueprint(config):
         return response, error.code
 
     for code in default_exceptions:
-        api.errorhandler(code)(_handle_http_exception)
+        api.errorhandler(code)(_handle_api_http_exception)
 
     return api
--- a/securedrop/tests/test_journalist.py
+++ b/securedrop/tests/test_journalist.py
@@ -2024,3 +2024,9 @@ def test_does_set_cookie_headers(journalist_app, test_journo):
         observed_headers = response.headers
         assert 'Set-Cookie' in observed_headers.keys()
         assert 'Cookie' in observed_headers['Vary']
+
+
+def test_app_error_handlers_defined(journalist_app):
+    for status_code in [400, 401, 403, 404, 500]:
+        # This will raise KeyError if an app-wide error handler is not defined
+        assert journalist_app.error_handler_spec[None][status_code]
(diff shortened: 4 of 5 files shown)


install_files/ansible-base/roles/app/templates/sites-available/journalist.conf
```diff
@@ -53,11 +53,11 @@ ErrorDocument 500 /error
 <Directory /var/www/>
   Options {{ apache_dir_options | default('None') }}
   AllowOverride None
-  <Limit GET POST HEAD>
+  <Limit GET POST HEAD DELETE>
```
redshiftzero: I think we should remove the `ErrorDocument` directives above as part of this ticket. I'm advocating this because they override the API error handlers we have implemented in the application code. Recall these Apache directives in the journalist config are why we saw redirects in #3977 instead of nice JSON responses. Let me know if you see a reason to keep them in the Apache config?

emkll: After taking a look at the removal of the ErrorDocument directives, the default Apache pages for errors do not return the security headers that are configured for the journalist interface (i.e.: x-xss-protection referrer-policy ...) While these error pages are static, I think we should still provide the headers for those pages. However, this meas we will need to provide ErrorDocuments for each error (400, 401, 403, 404 and 500)

conorsch:
Summarizing discussion today among @emkll, @redshiftzero, and myself: outstanding work here prior to merge is:

* [ ] Adding functional tests to verify the behavior works as intended
* [ ] Appending torify/curl calls to the manual QA test plan for upcoming releases
* [ ] Opening a follow-up ticket to track stronger integration testing for these endpoints, since #3977 will be closed by this PR. Progress on stronger integration testing is blocked by #3661

redshiftzero:
The automated (unit tests since functional tests need to get done in a followup) tests we need for the error handlers are in 9625c4c3c467c8130beea89964d4fc52fde17884, one can verify that these tests work by commenting out the registration of the error handlers [here](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/journalist_app/__init__.py#L96) and [here](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/journalist_app/api.py#L313).

redshiftzero:
While I don't want to increase the scope of this PR further, we _should_ also resolve https://github.com/freedomofpress/securedrop/issues/3877 with this PR. The reason is that this is another API bug that we have worked around _for now_ on the SDK/client side, but we should resolve while we are requesting a `securedrop-admin install` run to enable the API.

emkll:
Thanks for the tests @redshiftzero these look good, and thanks for opening the much-needed #4030 .I agree, let's expose Etags for Journalist interface. I have pushed a commit, tested, and updated the test plan accordingly.  I've opened #4032 to track an issue with Etags.

This PR should now be ready for re-review.

emkll:
Rebased on latest `develop`, this is ready for another review.

rmol:
:+1: I ran through both scenarios -- fresh install and upgrade (though I started at 0.12.0, not 0.11.0) -- and was able to verify the Etag hashes and delete a source through the API.

One note for anyone else testing this: since this PR was filed, we've started requiring `Authorization: Token` instead of `Authorization: Bearer`.

emkll:
rebased on latest `develop`

redshiftzero:
The testinfra failure [here](https://circleci.com/gh/freedomofpress/securedrop/24673) looks legit (`test_apache_journalist_interface_vhost`)

rmol:
> The testinfra failure [here](https://circleci.com/gh/freedomofpress/securedrop/24673) looks legit (`test_apache_journalist_interface_vhost`)

Yep. Just needs a space removed in the expected snippet:
```
diff --git a/molecule/testinfra/staging/app/apache/test_apache_journalist_interface.py b/molecule/testinfra/staging/app/apache/test_apache_journalist_interface.py
index 1249b2c08..cc1a5341a 100644
--- a/molecule/testinfra/staging/app/apache/test_apache_journalist_interface.py
+++ b/molecule/testinfra/staging/app/apache/test_apache_journalist_interface.py
@@ -63,7 +63,7 @@ common_apache2_directory_declarations = """
 <Directory {securedrop_code}>
   Options None
   AllowOverride None
-   <Limit GET POST HEAD DELETE>
+  <Limit GET POST HEAD DELETE>
     Order allow,deny
     allow from {apache_allow_from}
   </Limit>
```

emkll:
Thanks for the reviews/comments, I've addressed the test failure described above.
@conorsch would you mind doing a visual review of this (specifically removal of the ErrorDocument directives).

conorsch:
Changes look good. Based on the diff of the `ErrorDocument` removal, I'm inclined to agree with @emkll:

> While these error pages are static, I think we should still provide the headers for those pages. However, this meas we will need to provide ErrorDocuments for each error (400, 401, 403, 404 and 500)

Let's revisit the error routes when we add new tests validating that the headers are set, even on error. Some of the functional testing work you've been doing in other contexts is relevant here, @emkll, although we could certainly get away with some barebones `curl` checks to cover our bases. More important, the `ErrorDocument` directive will likely _never_ work for our needs, given the Client's expectation of full JSON results.

Since the changes are written affect only the Journalist Interface, no concerns with these changes going in as presented. Doing so will unblock pending work on the Client, and we can shore up the functional testing of the server side components post merge of the long-standing TBB work, when all of our functional tests will be running outside the VMs.

PR #4035, opened 2019-01-15
ci: Add app test job using xenial

redshiftzero:
## Status

Ready for review

## Description of Changes

Progress towards #4013, if there are no issues with the below then I will file followup issues for each class of app test failure

As mentioned in #4013, that bug was one of several issues that are surfaced by running the application test suite on Ubuntu Xenial. 

In the next couple of weeks we are planning to make a lot of changes here in SecureDrop core to ensure that SecureDrop works as expected for Xenial. To support faster iterations and multiple developers to help out with this effort, I'm proposing in this PR that we add a CI job that runs the Xenial app tests on develop (in addition to the CI job that runs the Trusty app tests). This job _is expected to fail until all app issues are resolved_. 

Why? Because this will be far, far faster than:

1. running tests manually on Xenial to review
2. everyone piling onto a single branch to try to resolve different issues while not stepping on each others toes
3. maintaining a feature branch, which I think we should generally avoid such that we if there are other changes being made that will cause additional app test changes to be made, we don't discover them at time of merge of the feature branch into develop

Thoughts?

## Testing

1. Are there now two CI jobs running here: `xenial-app-tests` and `trusty-app-tests`, where trusty passes but xenial does not? 

2. Can you now `make test-xenial` locally and run tests, that will fail? 

## Deployment

This is CI and dev env only


redshiftzero:
OK after discussion in standup, we're going to indeed merge this CI job and then fix the test failures iteratively (cc @kushaldas @emkll), thanks for review @heartsucker! 

For the future, the `trusty-app-tests` job is required for merge, but not the `xenial-app-tests` job. PRs made before this PR will have the `trusty-app-tests` job instead called `tests`, if that job is passing and you want to merge the PR, you can `@` me in the thread to merge the PR if you don't have permissions (avoids asking contributors to rebase, which is annoying).

PR #4052, opened 2019-01-21
added "filename" to reply creation api response

### LINKED ISSUE #4047: include `filename` in response when posting a reply
## Description

This request is motivated by implementing the replies feature in the SD client. When we make a reply, we get an object back from the API like

```json
{
    "message": "Message created or whaterver",
    "uuid": "some-uuud-abc123"
}
```

This is sufficient for us to track in-flight message, but it is not enough information for us to be able to correctly order replies. We need the `filename` that the server generates as well to be able to display the messages in the correct order. This is especially true if the Tor connections that are made cause messages to reach the server out of order.

We might as well return the entire object and drop the `message` field humans will not be seeing this and such messages are best conveyed the HTTP status code.

Without returning this field, if we want to have correct ordering of messages, for every reply we send send, we must make an additional call to fetch the reply object from the API to retrieve the `filename` for ordering.

## User Stories

As a consumer of the API, I want to make as few calls as possible in order to display information in a UI.

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #4047

Adds the `filename` to the API response when creating a reply.

## Testing

```bash
make test TESTFILES=tests/test_journalist_api.py::test_authorized_user_can_add_reply
```

### DIFF
--- a/securedrop/journalist_app/api.py
+++ b/securedrop/journalist_app/api.py
@@ -267,7 +267,8 @@ def make_blueprint(config):
                     raise e
 
             return jsonify({'message': 'Your reply has been stored',
-                            'uuid': reply.uuid}), 201
+                            'uuid': reply.uuid,
+                            'filename': reply.filename}), 201
 
     @api.route('/sources/<source_uuid>/replies/<reply_uuid>',
                methods=['GET', 'DELETE'])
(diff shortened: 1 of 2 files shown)


heartsucker:
@kushaldas I accidentally committed a hack I use during development so I don't have to use TOTP when working on the client. That was resolved by a force push.

PR #4074, opened 2019-01-25
Don't bump SecureDrop version strings in docs for release candidates

### LINKED ISSUE #4072: Latest docs specify an RC instead of a Release version
## Description

`update_version.sh` bumps version string in docs even when an RC is created. There might be a small period of time (e.g. during release QA) where that Tag exists (albeit not signed)

## Steps to Reproduce

https://docs.securedrop.org/en/latest/set_up_admin_tails.html?highlight=git%20checkout and observe instructions to check out 0.12.0~rc1 tag

## Expected Behavior

The tag should be the latest release (as of today, 0.11.1)
## Actual Behavior

The tag is 0.12.0~rc1

emkll:
## Status

Ready for Review

## Description of Changes

Fixes #4072

For release canadidates, `update_version.sh` will update docs strings to the previous SecureDrop release.

## Testing

- [x] Ensure the docs in develop/latest always point to the last (signed) release of SecureDrop

#### Ensure update_version.sh was properly updated:
- [x] `securedrop/bin/dev-shell ../update_version.sh 0.13.0~rc2` should not update doc strings (remains `0.12.1`
- [x] `securedrop/bin/dev-shell ../update_version.sh 0.13.0` should update doc strings to `0.13.0`
- [x]  `securedrop/bin/dev-shell ../update_version.sh 0.14.0~rc1` should not update doc strings
- [x]  `securedrop/bin/dev-shell ../update_version.sh 0.15.0` should update doc strings to `0.15.0`

## Deployment

Dev env only


update_version.sh
```diff
@@ -52,9 +52,12 @@ sed -i "s/^\(securedrop_app_code_version: \"\).*/\1$NEW_VERSION\"/" install_file
...
 sed -i "s@$(echo "${OLD_VERSION}" | sed 's/\./\\./g')@$NEW_VERSION@g" molecule/builder/tests/vars.yml
 
-# Update the version that we tell people to check out in the install doc
-sed -i "s@$(echo "${OLD_VERSION}" | sed 's/\./\\./g')@$NEW_VERSION@g" docs/set_up_admin_tails.rst
-sed -i "s@$(echo "${OLD_VERSION}" | sed 's/\./\\./g')@$NEW_VERSION@g" docs/conf.py
+# Update the version that we tell people to check out in the install doc if an actual release.
+# We should not update version string in the docs with a release candidate is created
+if [[ ! $NEW_VERSION == *~rc* ]]; then
+  sed -i "s@$(echo "${OLD_VERSION}" | sed 's/\./\\./g')@$NEW_VERSION@g" docs/set_up_admin_tails.rst
+  sed -i "s@$(echo "${OLD_VERSION}" | sed 's/\./\\./g')@$NEW_VERSION@g" docs/conf.py
```
redshiftzero: So there's the following scenario:

1. Bump 0.11.0 to 0.11.0-rc1: docs don't get updated as expected, so docs say 0.11.0
2. Bump 0.11.0-rc1 to 0.12.0: docs don't get updated since this script is using the 0.11.0-rc1 version string from `__version__.py`, i.e. not 0.11.0

We could do something like just taking the beginning part of the version string before `~rc` from the OLD_VERSION and _then_ apply these sed commands - what do you think?

emkll: Ah right, good point. What about using `molecule/shared/stable.ver`? I've updated the PR and the test plan based on your comments and it should address your comments

PR #4077, opened 2019-01-28
Added base OS info to metadata endpoint #4059

### LINKED ISSUE #4059: [xenial] Add base OS info to metadata endpoint
## Description

Part of #3204. In order to have better visibility into which SecureDrop instances are actively updating to from Ubuntu Trusty to Ubuntu Xenial (and actively engage with admins as needed), it makes sense to expose this info via the existing metadata endpoint.

To avoid breaking existing uses, this should be done in a new `server_os` variable.

From a security perspective, this does not give an attacker key information -- for now, both releases receive security updates; once 14.04 has reached EOL, it will also no longer receive SecureDrop package updates, which will be reflected in its already exposed SecureDrop version.

# User Stories

As a SecureDrop administrator, I'd like to be actively alerted by the support team when I have to perform critical security updates, so that I can act on them with appropriate priority.

As a SecureDrop support team member, I'd like to know whether an instance is following recommended upgrade procedures, so I can increase the urgency of outreach efforts if required.

nightwarrior-xxx:
## Status

Ready for review

## Description of Changes

Fixes #4059

### DIFF
--- a/docs/development/setup_development.rst
+++ b/docs/development/setup_development.rst
@@ -130,9 +130,6 @@ development.
 * **Password:** ``correct horse battery staple profanity oil chewy``
 * **TOTP secret:** ``JHCO GO7V CER3 EJ4L``
 
-.. tip:: We're hosting a demonstration instance of SecureDrop so that anyone can see how   
-    it works. The demo site generates Time-based One-time Password (TOTP) that you can use with the password above. To get a generated token, go to: http://demo.securedrop.club/
-
 If you need to generate the six digit token, use the TOTP secret in
 combination with an authenticator application that implements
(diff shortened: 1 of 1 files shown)


kushaldas:
@nightwarrior-xxx Please squash the commits into one. Also, add corresponding tests for this change.

zenmonkeykstop:
Hi @nightwarrior-xxx - in addition to the changes requested by @kushaldas above, can you update the PR to include a basic test plan for this feature change? It doesn't have to go into massive detail, just steps to verify the change and the expected values of the new variable would be fine. 

Thanks again for this!

nightwarrior-xxx:
@redshiftzero Yup, sorry for being too late !!

PR #4088, opened 2019-01-29
return user id when creating auth token

### LINKED ISSUE #4081: Return current journalist ID when creating a token
## Description

The endpoint that generates auth tokens needs to return the current user's UUID so the client knows which user it is without having to hit the user data endpoint.

heartsucker:
## Status

## Description of Changes

Fixes #4081

Return the journalist's UUID when they create an auth token.

## Testing
- Verify the name `journalist_uuid` is a reasonable name for this field
- `make test` like usual

## Deployment

This is a backwards compatible API change.


kushaldas:
Kicking the failed GUI tests.

PR #4080, opened 2019-01-28
Stores Trusty & Xenial deb packages side by side

conorsch:
## Status

Work in progress.

## Description of Changes

Fixes #3961. Fixes #4051.

Changes proposed in this pull request:

* **Removes tor-apt repo** ; tor debs are now stored in a single, consolidated apt repo
* Adds customization of `securedrop-app-code` package for discrete dependencies under Trusty & Xenial
* Builds Xenial debs inside Xenial container
* Deb packages are now stored in subdirs: `build/trusty/` and `build/xenial/`. 

## Testing

Given the magnitude of changes here, we must confirm that both staging and prod environments are fully functioning, on both Trusty and Xenial. 

### Staging environment
* [ ] `make build-debs && make staging` passes without error
* [ ] All infra tests pass `molecule verify -s {libvirt,virtualbox}-staging` 
* [ ] `make build-debs-xenial && make staging-xenial` passes without error
* [ ] All infra tests pass `molecule verify -s {libvirt,virtualbox}-staging-xenial` 

** Note that due to https://github.com/freedomofpress/securedrop/issues/3916 some PaX-related tests might fail, and there should be a note referring to that issue in the output.
### Production environment
```
vagrant box add bento/ubuntu-16.04
# Ensure you have vagrant mutate installed
vagrant plugin install vagrant-mutate
vagrant mutate bento/ubuntu-16.04 libvirt
```

If testing xenial, you must change the base box in the `Vagrantfile`: replace `bento/ubuntu-14.04` with `bento/ubuntu-16.04` for `app-prod` and `mon-prod`

```
vagrant up --no-provision /prod/
```

In Tails, update the vars to use the `apt-test` server. Specifically, that entails:
1. In `install_files/ansible-base/roles/install-fpf-repo/defaults/main.yml`, set  `apt_repo_url: https://apt-test.freedom.press` 
2. `cp install_files/ansible-base/roles/install-fpf-repo/files/{apt-test-signing-key,fpf-signing-key}.pub`

Then, proceed with testing. Confirm the following in your report:

* [ ] Clean installation (prod vms) on Xenial using `apt-test.freedom.press` completes without error (Basic server testing is successful)
* [ ] Clean installation (prod vms) on Trusty using `apt-test.freedom.press` completes without error (Basic server testing is successful)

## Deployment

At present, the "xenial" apt channel is only served up from non-prod URLs. We'll update the prod config to do the same as part of the 0.12.0 release.

[...]

Issue #4031, opened 2019-01-14
Upgrade Tor to 0.3.5.7 (or  0.3.4.10)


emkll:
## Description

SecureDrop servers are currently tracking the 0.3.4.x series of Tor stable. A new Tor Stable series was released with support for authenticated v3 onion URLS (0.3.5.x). This should now unblock the use of v3 authenticated onion services on Journalist and SSH interfaces.

## User Stories

As a SecureDrop user and admin, I would like to use better cryptography for encrypting data in transit and authenticating onion services.
[0] https://blog.torproject.org/new-releases-tor-0357-03410-and-03311

emkll:
It appears our current hidden service configuration (specifically the Journalist interface) does not work under the 0.3.5.X Tor series in Xenial (0.3.5.7-1~xenial+1):
```
Jan 29 16:25:10 app-prod tor[1332]: Jan 29 16:25:10.123 [notice] Tor 0.3.5.7 running on Linux with Libevent 2.0.
21-stable, OpenSSL 1.0.2g, Zlib 1.2.8, Liblzma 5.1.0alpha, and Libzstd N/A.
Jan 29 16:25:10 app-prod tor[1332]: Jan 29 16:25:10.123 [notice] Tor can't help you if you use it wrong! Learn h
ow to be safe at https://www.torproject.org/download/download#warning
Jan 29 16:25:10 app-prod tor[1332]: Jan 29 16:25:10.123 [notice] Read configuration file "/usr/share/tor/tor-ser
vice-defaults-torrc".
Jan 29 16:25:10 app-prod tor[1332]: Jan 29 16:25:10.124 [notice] Read configuration file "/etc/tor/torrc".
Jan 29 16:25:10 app-prod tor[1332]: Jan 29 16:25:10.134 [warn] Hidden service option HiddenServiceAuthorizeClien
t is incompatible with version 3 of service in /var/lib/tor/services/journalist
Jan 29 16:25:10 app-prod tor[1332]: Jan 29 16:25:10.134 [warn] Failed to parse/validate config: Failed to config
ure rendezvous options. See logs for details.
Jan 29 16:25:10 app-prod tor[1332]: Jan 29 16:25:10.135 [err] Reading config failed--see warnings above.
Jan 29 16:25:10 app-prod systemd[1]: tor@default.service: Control process exited, code=exited status=1
Jan 29 16:25:10 app-prod systemd[1]: Failed to start Anonymizing overlay network for TCP.
[...]

Issue #2951, opened 2018-01-29
Migrate SecureDrop onion interfaces to v3 onion services


redshiftzero:
## Description

Next generation onion services are now in the latest stable Tor release, 0.3.2.9. We should have a migration path to enable v3 onion services alongside v2 onion services for source interfaces. Since our journalist and app/mon SSH onion services are stealth onion services (i.e. they use `HiddenServiceAuthorizeClient stealth`), ~~we can't migrate those over until v3 stealth onion service support lands in stable, but we _can_ migrate the source interfaces~~ (support for v3 stealth onion services was added in 0.3.5.x, so we can migrate all interfaces).  

This can be done by adding an additional block in `torrc` on the app server containing `HiddenServiceVersion 3` alongside v2, e.g.:

```
HiddenServiceDir /var/lib/tor/services/source
HiddenServicePort 80 127.0.0.1:80

HiddenServiceDir /var/lib/tor/services/v3-source
HiddenServicePort 81 127.0.0.1:81
HiddenServiceVersion 3
```

We can continue to support v2 onion addresses for the foreseeable future until we have all instances moved over to v3. Given the action required from news organizations to add the v3 onion address to their landing pages / physical newspapers / etc., and the messaging they will need to need to do with any long-term sources to encourage them to move over to the v3 address, we'll need to coordinate with news organizations to make this change. 

## Subtickets

- [x] #4627 `securedrop-admin sdconfig` should set `v2_onion_services` and `v3_onion_services` variables
- [x] #4628 add v3 onion support to `tor-hidden-services` ansible role 
- [x] #4629 Modify `securedrop-admin tailsconfig` to use v3 if available
- [ ] #4630 set `Alt-Srv` header in Apache configs
- [x] #4631 investigate v2 only messaging options
- [x] #4632 documentation for admins

## User Stories

As a SecureDrop source, I want to use the most secure version of onion services when I use a SecureDrop instance so that I minimize risk to myself.

cyphar:
v3 onion services provide numerous safety advantages for both admins and sources, with massive improvements to the cryptography involved (invalidating most of the concerns in [the docs](https://docs.securedrop.org/en/release-0.13.0/https_source_interface.html)). It also makes it harder for HSDirs to know whether they are hosting the descriptors without knowing the .onion address, as well as many other improvements.

[...]

PR #4092, opened 2019-01-31
Explicitly declare onion services as v2 for existing installs

emkll:
## Status

Ready for review

## Description of Changes

Towards #4031 , this is required to ensure torrc config is compatible with tor 0.3.5.x series. 
Modify the torrc file in place to explicitly declare current onion services as v2 onion services. Since securedrop-config is installed after tor hidden services are configured, it will also perform the modification (and be a no-op once #4080  is merged at install-time)

## Testing
- [ ] Do we actually need this? Ansible changes in #4080 will make these changes in the torrc configuration. If we require news orgs to run the playbook prior to Xenial upgrade, this addition might be superfluous.
- [ ] In both cases, ensure that the resulting configuration makes sense per the documentation (https://www.torproject.org/docs/tor-onion-service.html.en) and will not make it more difficult to make configuration changes when migrating to v3 onion services (#2951)

### Clean install

- Make debs and provision staging environment on this branch
- [ ] Hidden services configurations in `/etc/tor/torrc` on app and mon server specify version 2 for all hidden services
- [ ] Source, journalist and ssh interfaces are accessible

### Upgrade testing

- Build debs and provision existing SecureDrop 0.11.1 install with these debs
- [ ] Hidden services configurations in `/etc/tor/torrc` on app and mon server specify version 2 for all hidden services
- [ ] Source, journalist and ssh interfaces are accessible

## Deployment

These changes will be deployed to new and existing SecureDrop instances via `securedrop-config` deb package.

### DIFF
--- a/install_files/securedrop-config/DEBIAN/postinst
+++ b/install_files/securedrop-config/DEBIAN/postinst
@@ -39,6 +39,16 @@ allow_apt_user_in_iptables() {
            "$rules_v4"
     fi
 }
+# Tor 0.3.5.x series now defaults to v3 onion URLs, but SecureDrop currently
+# uses v2 onion URLs. We must explictly set this definition in torrc to avoid
+# breakage when upgrading from Tor 0.3.4.x to 0.3.5.x.
+set_v2_hidserv_in_torrc() {
+    if [ -f /etc/tor/torrc ]; then
+        if ! grep -q HiddenServiceVersion /etc/tor/torrc ; then
+            perl -pi -e 's/^(HiddenServiceDir.*)$/$1\nHiddenServiceVersion 2/' /etc/tor/torrc
+        fi
+    fi
+}
 
 case "$1" in
     configure)
@@ -61,6 +71,7 @@ case "$1" in
     fi
 
     allow_apt_user_in_iptables
+    set_v2_hidserv_in_torrc
     ;;
 
     abort-upgrade|abort-remove|abort-deconfigure)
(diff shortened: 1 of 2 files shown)


redshiftzero:
In my opinion, it's worth adding this for the following scenario that @heartsucker is mentioning above:

1. Admin makes a backup during the pre-0.12 release time period where we are instructing admins to backup (and the backup contains tor configs). 
2. Admin upgrades to Xenial.
3. Admin runs `securedrop-admin install`. 
4. Admin runs restore using an earlier backup file, putting old `torrc` back on disk breaking the onion services. 

But in this case, since we don't `dpkg-reconfigure securedrop-config` during the restore process, I think this change wouldn't resolve the above scenario. What do y'all think about adding `subprocess.check_call(['dpkg-reconfigure', 'securedrop-config'])` [here](https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/restore/files/restore.py#L49)?

conorsch:
100% agree we want the changes in postinst. At present, the plan for upgrade is:

1. Admin confirms instance is running at least 0.12.0 
2. Performs manual `do-release-upgrade` tasks
3. Runs `./securedrop-admin install` to enforce state

We don't _also_ want to have to run `./securedrop-admin install` _before_ the `do-release-upgrade` task, if we can avoid it. These changes satisfy in that regard.

@redshiftzero brings up a great point on the restore script, happy to accommodate there. Still testing the functional changes presented here. I feel strongly there should be corresponding config tests added, will append those as part of review.

conorsch:
So far I've validated the staging clean install story configures the services correctly, on both Trusty & Xenial. Logged in interactively to inspect the torrc and confirm the v2 strings were there. Upgrade testing is next; see box metadata in #4093, will use those to base the upgrades on.

conorsch:
Upgrade scenario from 0.11.1 works smoothly, confirmed resolution of both Source & Journalist Interfaces. Did _not_ test upgrade for Xenial, since that's not terribly relevant (we plan to have folks upgrade to Xenial _after_ release of 0.12.0).

PR #4094, opened 2019-02-01
[xenial] gpg 2.1+ compliance + fix decryption of journalist replies

### LINKED ISSUE #3622: Change `gnupg` to `pretty_bad_protocol` in requirements files
## Description

In [this commit](https://github.com/isislovecruft/python-gnupg/commit/5d756ee44c1fe6904c81304573c1b958092c9f51) the package `gnupg` was renamed to `pretty_bad_protocol`. If we want to keep getting updates, we need to change the name.

## User Research Evidence

Honestly, GPG is pretty bad. The maintainer seems to be on to something with the name change.

## User Stories

As a dev, I want to get the latest updates on my package.

As a dev, I want to be reminded about the badness of GPG every time I use it.

### LINKED ISSUE #4013: [xenial] Journalist replies are not displayed in source interface

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #3622 
Fixes #4013

 - This PR makes SecureDrop work with gpg 2.1, while preserving functionality for earlier versions of gpg used on trusty (the gpg config options passed to the gpg binary will now vary based on the version of gpg that is used on the server). gpg 2.1 introduced some other significant changes, the most important of which is now requiring the use of gpg-agent. 

thoughts welcome on this as I had to sidestep some API weirdness / bugs on the gpg side which I note in the commit messages

## Testing

- [ ] All xenial app test should now pass (and after this PR is merged we will make that a required CI job)
- [ ] All trusty app tests should also still pass
- [ ] In the xenial staging environment, a source should now be able to decrypt journalist replies
- [ ] In the xenial dev environment (`BASE_OS=xenial make dev`), a source should now be able to decrypt journalist replies
- [ ] In the trusty dev environment (`make dev`), a source should still be able to decrypt journalist replies
- [ ] In the trusty staging environment, a source should still be able to decrypt journalist replies

## Deployment

New installs and upgrades both will have the now required gpg-agent conf (for gpg 2.1+) written to disk via `postinst` of `securedrop-app-code`. A reboot will then occur.

### DIFF
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
@@ -105,6 +105,7 @@
   /var/lib/securedrop/db.sqlite-journal w,
   /var/lib/securedrop/keys/* rwl,
   /var/lib/securedrop/keys/*.app-staging.* w,
+  /var/lib/securedrop/keys/gpg-agent.conf r,
   /var/lib/securedrop/keys/private-keys-v1.d/* rw,
   /var/lib/securedrop/keys/openpgp-revocs.d/* rw,
   /var/lib/securedrop/keys/pubring.gpg r,
--- a/install_files/securedrop-app-code/debian/postinst
+++ b/install_files/securedrop-app-code/debian/postinst
@@ -84,6 +84,19 @@ case "$1" in
       chmod 0700 "$dir"
     done
 
+    # Ensure required gpg-agent.conf is in place, see #4013.
+    if [ -e "/var/lib/securedrop/keys/gpg-agent.conf" ]; then
+
+        # gpg-agent.conf does exist, update it if needed.
+        if ! grep -qE '^allow-loopback-pinentry$' /var/lib/securedrop/keys/gpg-agent.conf; then
+            echo allow-loopback-pinentry >> /var/lib/securedrop/keys/gpg-agent.conf
+        fi
+
+    else
+        # gpg-agent.conf does not yet exist, create it.
+        echo allow-loopback-pinentry > /var/lib/securedrop/keys/gpg-agent.conf
+    fi
+
     chown -R www-data:www-data /var/lib/securedrop /var/www/securedrop
 
     pip install --no-index --find-links=/var/securedrop/wheelhouse --upgrade \
--- a/securedrop/bin/dev-deps
+++ b/securedrop/bin/dev-deps
@@ -46,8 +46,8 @@ function reset_demo() {
     # Create gpg-agent.conf
     echo allow-loopback-pinentry > /var/lib/securedrop/keys/gpg-agent.conf
 
-    # Kill gpg-agent(s) so it picks up the new config on restart.
-    pkill -f gpg-agent
+    # Kill gpg-agent(s) if they exist so it picks up the new config on restart.
+    pkill -f gpg-agent || true
     # Note that we should avoid `gpgconf --kill gpg-agent` since the pkill command will
     # handle killing multiple gpg-agent processes if they exist (this is what we want).
 
--- a/securedrop/bin/dev-deps
+++ b/securedrop/bin/dev-deps
@@ -37,10 +37,30 @@ function run_sass() {
 }
 
 function reset_demo() {
+    # Set up gpg keys directory structure.
     sudo mkdir -p /var/lib/securedrop/{store,keys,tmp}
     sudo chown -R "$(id -u)" /var/lib/securedrop
     cp ./tests/files/test_journalist_key.pub /var/lib/securedrop/keys
-    gpg --homedir /var/lib/securedrop/keys --import /var/lib/securedrop/keys/test_journalist_key.pub >& /tmp/gpg.out || cat /tmp/gpg.out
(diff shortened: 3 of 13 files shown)


install_files/securedrop-app-code/DEBIAN/postinst
```diff
@@ -84,6 +84,9 @@ case "$1" in
       chmod 0700 "$dir"
     done
 
+    # Ensure required gpg-agent.conf is in place, see #4013
+    echo allow-loopback-pinentry > /var/lib/securedrop/keys/gpg-agent.conf
+
```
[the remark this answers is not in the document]
heartsucker: Personally I think this should be something like

```bash
if ! grep -qE '^allow-loopback-pinentry$' /var/lib/securedrop/keys/gpg-agent.conf; then
    echo allow-loopback-pinentry >> /var/lib/securedrop/keys/gpg-agent.conf
fi
```

just to make this a bit more future-proof. But this is fine to merge as is imo because CI will catch this.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
```diff
@@ -105,6 +105,7 @@
   /var/lib/securedrop/db.sqlite-journal w,
   /var/lib/securedrop/keys/* rwl,
   /var/lib/securedrop/keys/*.app-staging.* w,
+  /var/lib/securedrop/keys/gpg-agent.conf rw,
```
heartsucker: Does `apache` need to be able to write to this file? Because `postinst` is run as `root`, so with `r` might not be needed.

emkll:
Yes, it only happens once: the first time a source decrypts a reply once the server was upgraded. I cannot reproduce by creating a new source (and replying to this new source), nor by rebooting.

heartsucker:
Does the upgrade do something like symlink `gpg2` to `gpg` or anything weird like that?

redshiftzero:
Ah gotcha, I expect this is happening due to the private key migration done (once) when running gpg2.1 for the first time

emkll:
On xenial, gpg still points to gpg (v1), no symlink. I think you are right @redshiftzero . Also I did have a source set up on that server, so a migration was necessary. It doesn't seem to generate an ossec alert, so I think we should keep a close eye on the apache logs during the QA period.

redshiftzero:
ok so one idea that _should_ handle this situation (as well as validate the theory above re: this `ValueError` being due to the private key migration) is to run the migration manually, e.g. `gpg2 --import < /var/lib/securedrop/keys/secring.gpg` during `postinst` of `securedrop-app-code`, let me give that a try

install_files/securedrop-app-code/debian/postinst
```diff
@@ -97,6 +97,12 @@ case "$1" in
         echo allow-loopback-pinentry > /var/lib/securedrop/keys/gpg-agent.conf
     fi
 
+    # Migrate private keyring to gpg2.1 if needed
+    if [ ! -d "/var/lib/securedrop/keys/private-keys-v1.d" ]; then
```
kushaldas: This is a good approach for the upgrade story, in future we can remove these postinst steps (after a few releases)

heartsucker: We probably never can. We had instances that were pre 0.4 recently, and that release was a looooooong time ago.

redshiftzero: So while we have instances that had not ran their Ansible playbooks since before 0.4, the server side was up to latest. Since we can monitor what version the securedrop-app-code package is on via the source interface metadata endpoint (and soon we'll be able to monitor also the base OS via the source interface metadata endpoint), we should be able to remove this from postinst once we have everyone on xenial and post-0.12.0.

redshiftzero:
OK so I tested 97965191994488327c296e813e2b5a8c3afa993d via the following procedure:

1. Provision staging VMs
2. `sudo do-release-upgrade` to upgrade to xenial
3. Upgrade securedrop-app-code package

But 97965191994488327c296e813e2b5a8c3afa993d did not work: I still saw the exception, because the migration failed to work due to the fact that importing secret keys appeared to require the secret key passphrase, which we don't have (a behavior change in gpg2.1, similar to why 826aa963f88d192cc37834bd1b6be506e20d901f was necessary). Furthermore, thinking about this a bit more, even if this did work, `postinst` would not be the right place for trying to run the manual migration because organizations will be updating to 0.12.0 _before_ they do the upgrade to xenial, and the a manual migration will only work if organizations are already using gpg 2.1. So we need to go back to the drawing board on this one. 

Since the note @emkll found won't produce an OSSEC alert (and recall source-error.log does not exist in prod), and this high-priority bug is resolved by this PR, I suggest we should merge this in as is. That said, it is still an unhandled exception, so I'm going to file a followup, and we should try another method to resolve if we have time before 0.12.0 (since the fix will be smaller than the diff in this PR, it's something we could do during QA depending on how things go). 

Note to reviewers: I've made no other changes since the testing was done above (I force pushed to drop 97965191994488327c296e813e2b5a8c3afa993d), so an additional round of testing is not necessary.

PR #4102, opened 2019-02-06
update name of models.py to db.py since it changed

deeplow:
## Status

Ready for review

## Description of Changes
The current version of the securedrop-client no longer has a models.py. It has been renamed to db.py and the docs should reflect that change (it was introduced in commit 7b0c560004 over in the securedrop-client repo)


deeplow:
You're welcome! Thanks for merging my first pr @redshiftzero (even though a very small one). 
Hopefully I'll be able do contribute more next time :)

PR #4061, opened 2019-01-23
Upgrade apt without following redirects on first install, fix CVE-2019-3462

emkll:
## Status

Ready for review 

## Description of Changes

Fixes #4058 .

Changes proposed in this pull request:

### Install logic
- [ ] Initial apt update uses the correct flags that mitigate the vulnerability [0]
- [ ] The install completes without error
- [ ] This also works in Xenial: `molecule converge -s {libvirt,virtualbox}-staging-xenial
### Container updates
- [ ] The trusty-based builder container contains an updated apt package and the upstream hash is valid.
- [ ] The xenial-based dev container contains an updated apt package and the upstream hash is valid.
- [ ] The trusty-based dev container contains an updated apt package and the upstream hash is valid.
- [ ] The stretch-based admin container contains an updated apt package and the upstream hash is valid.
- [ ] The trusy-based dev container was uploaded to container registry.

### Existing installs
Existing installs have already been updated using cron-apt
### New installs
This PR should fix the issue for new installs. Since Ubuntu ISOs will not contain the apt fixes, we change the install logic to update apt before performing any type of other apt operations

### If you made changes to `securedrop-admin`:

- [X] Linting and tests (`make -C admin test`) pass in the admin development container

### If you made changes to the system configuration:

- [X] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

### If you made non-trivial code changes:

- [X] I have written a test plan and validated it for this PR

[0] https://justi.cz/security/2019/01/22/apt-rce.html

PR #4115, opened 2019-02-09
[xenial] Ensures server packages required by ./securedrop-admin install are present on Xenial

### LINKED ISSUE #4109: [xenial] Xenial OS installer does not include some required packages if only "OpenSSH Server" selected
## Description

During the Xenial OS install, the user is prompted to select software for installation. Under Trusty, SecureDrop required that only `OpenSSH server` be selected. Xenial differs in having a preselected `standard system utilities` option. 

If the user deselects `standard system utilities`, selects `OpenSSH server`, and proceeds with the installation, `./securedrop-admin install` fails at the Gathering Facts step, claiming that it can't connect via SSH.

However, the issue is not with SSH, but with the fact that python is not installed on Xenial with the above software selection choices. (To prove this, turn on -vvvv logging in ansible, grab the SSH command run by gather_facts, and run it at the console.)

It turns out that 3 packages are required to complete the installation, `python`, `apt-transport-https`, and `dnsutils`.

## Steps to Reproduce
(Tested on 5-series NUCs)
- Proceed with a Xenial SecureDrop installation, deselecting standard system utilities and selecting OpenSSH server at the Software Selection dialog

## Expected Behavior
Installation finishes without error

## Actual Behavior
Installation fails during `./securedrop-admin install` as described above

## Comments
Selecting `standard system utilities` may solve this issue, but it also installs unnecessary packages, which is bad.
Installing the packages on the servers manually adds more complexity to the installation procedure, also bad.
One alternative is to install the required packages via a preliminary playbook that disables gather_facts and runs a raw ssh command:
```
#!/usr/bin/env ansible-playbook
---
- name Install required packages on SecureDrop servers
  hosts: securedrop
  any_errors_fatal: yes
  gather_facts: no
  become: yes
  pre_tasks:
    - name: 'install packages'
      raw: 'sudo apt-get -y install python apt-transport-https dnsutils'

```
(Thanks to @kushaldas for this suggestion)

This can be run at the command line, but a proper fix would probably be to add it to the securedrop-admin install function,  just before the main `securedrop-prod.yml` is run. (One downside is that the user would be prompted for the server sudo password twice.)

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Fixes #4109 .

Adds an Ansible role that ensures `securedrop-admin install`'s  missing package prerequisites (python, apt-transport-https, dnsutils) are installed on Xenial systems set up with only the `OpenSSH server` option checked in the OS installer's 'Software selection` dialog.

### Prod VMs
- Modify Vagrantfile if necessary to use 16.04 images for `app-prod` and `mon-prod`
- Create prod VMs: `vagrant up --no-provision /prod/`
- Use `vagrant ssh <vm-name-here` to connect to VMs, and remove the packages listed above. Verify that `/usr/bin/python` in particular has been removed. 
- On your Tails VM Admin Workstation, modify `/Persistent/securedrop/install_files/ansible-base/roles/install-fpf-repo/defaults/main.yml` to point to the `apt-test.freedom.press` test repo and to use `apt-test-signing-key.pub` as the repo signing key
- proceed with the SecureDrop installation

### Hardware
- Install Xenial on app and mon servers with the usual options, ensuring that `standard system utilities` is deselected and `OpenSSH server` is selected on the Software Selection dialog.
- On your Admin Workstation, modify `~/Persistent/securedrop/install_files/ansible-base/roles/install-fpf-repo/defaults/main.yml` to point to the `apt-test.freedom.press` test repo and to use `apt-test-signing-key.pub` as the repo signing key
- Proceed with installation as normal.

## Deployment

This only affects new Xenial installs, and nothing special is required for deployment.

(note: tested on prod VMs)


kushaldas:
@zenmonkeykstop This also needs a rebase against the develop to fix the `mypy` issue.

conorsch:
Appended a commit moving the logic already present via #4061 into the new role.

The outstanding question I have now is: will the `raw` apt-get task work against fresh hardware hosts? Recall that the VM images we use for testing _already_ have passwordless sudo enabled. Given the `sudo` call in the raw task, coupled with the fact that we haven't yet enabled passwordless sudo as part of the install flow, the task may block indefinitely on a password prompt. That would be an inscrutable error for Admins to face as part of the first install experience.

Hardware testing required in order to confirm the above. If it does prove to be a problem, then we're faced with a few options:

1. Conditionally run the task if it's reasonable to assume this is a first time install (e.g. `ansible_host` does not end in `.onion`)
2. Attempt to safely pass in sudo password since it was collected via `securedrop-admin install` ; prefer to avoid, given that we'd be reinventing Ansible wheels doing so
3. Update the install docs to instruct admins to install python2 manually, as part of the pubkey bootstrapping
4. _Optional_ Contingent on 3); we could fail if python2 is not installed, and point Admins to the docs to reduce confusion if the manual install step is overlooked

Slight preference for 3 and optionally 4, but defer to the team for input.

zenmonkeykstop:
Apologies @conorsch I think I may have nuked your commit (albeit with one doing something similar.)

I won't be in a position to test against hardware until Saturday, if someone else can test against hardware I'd appreciate it. I'm optimistic that it will work if raw commands use a similar mechanism to gather_facts.

conorsch:
@zenmonkeykstop Thanks for reply, no surprise we're stepping on toes a bit with this one. :innocent: Just added some more fixes on top of yours, pulling in some work that was done by @emkll in tandem. I believe this is now ready for review. cc @redshiftzero 

I remain skeptical about the sudo password behavior, but testing on hardware required to get clarity there.

emkll:
Hey @kushaldas it seems like you are using the prod apt server (apt.freedom.press) to test under xenial. There is currently no Xenial channel hosted on that server: You must use apt-test.freedom.press. The Xenial channel will be hosted on prod apt servers starting with the 0.12.0 release

kushaldas:
>Hey @kushaldas it seems like you are using the prod apt server (apt.freedom.press) to test under xenial. There is currently no Xenial channel hosted on that server: You must use apt-test.freedom.press. The Xenial channel will be hosted on prod apt servers starting with the 0.12.0 release

@emkll  can you please add the steps for the same in the test plan? This will help in future reference.

zenmonkeykstop:
Apologies for the omission @kushaldas, I've updated the test plan.

redshiftzero:
@kushaldas please take another look at this one when you get a chance today, let's try to get this merged in for 0.12.0~rc2 which we should get done before the weekend

PR #4116, opened 2019-02-11
[xenial] Updated release-upgrades Prompt for Trusty, Xenial as appropriate

### LINKED ISSUE #4104: [xenial] Reinstate LTS upgrade notification
In #3546, we set `Prompt` to `never` in  `/etc/update-manager/release-upgrades` to suppress premature OSSEC notifications. In order to perform the update to Xenial via `do-release-upgrade`, it needs to be reset to `lts`. We can do so in postinst so the admin does not have to do it manually.

To avoid spurious 18.04 notifications, this should be done on Trusty, while on Xenial `Prompt` should be set to `never`. We can inspect `lsb_release -sc` to avoid forking the build logic.

As a stretch goal, we should consider customizing the OSSEC notification that will be sent to 14.04 users. The default notification will say:

```
/etc/cron.weekly/update-notifier-common:
New release '16.04.1 LTS' available.
Run 'do-release-upgrade' to upgrade to it.
```

Ideally, the notification should point to: 

  https://securedrop.org/xenial-upgrade

This is the redirect also pointed to via #4055. Having the OSSEC alert point to this link will help prevent admins from attempting the upgrade without following the documentation.

Part of #3204.

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #4104

Update the `Prompt` in the release manage file depending on system codename. Additionally, attempt to rewrite the OSSEC alerts to prevent admins from running `do-release-upgrade` and breaking everything.

## Testing

1. Trusty
    - Build trusty debs
    - Install on trusty box
    - Verify that `/etc/update-manager/release-upgrades` contains `Prompt=lts`
    - Verify that `/usr/lib/ubuntu-release-upgrader/check-new-release` contains "Visit https://securedrop.org/xenial-upgrade for more information" (and so does  `/var/lib/ubuntu-release-upgrader/release-upgrade-available` if it exits)
2. Xenial
    - Build xenial debs
    - Install on xenial box
    - Verify that `/etc/update-manager/release-upgrades` contains `Prompt=never`
    - Verify that `/usr/lib/ubuntu-release-upgrader/check-new-release` does not contain "Visit https://securedrop.org/xenial-upgrade for more information" (and neither does  `/var/lib/ubuntu-release-upgrader/release-upgrade-available` if it exits)

## Deployment

All installs on Trusty will have files modified to the state SD requires. All installs on Xenial will have their files reverted to the default state.

### DIFF
--- a/install_files/securedrop-config/DEBIAN/postinst
+++ b/install_files/securedrop-config/DEBIAN/postinst
@@ -4,10 +4,28 @@
 set -e
 set -x
 
-disable_upgrade_prompt() {
-    # Disable do-release-upgrade notification
-    sed -i 's/Prompt=.*/Prompt=never/' /etc/update-manager/release-upgrades || true
+
+# Issue #4104
+# Set Prompt=never on Xenial
+# Set Prompt=lts on Trusty
+update_release_prompt() {
+    set -e
+
+    declare -r upgrade_config='/etc/update-manager/release-upgrades'
+
+    declare -r release="$(lsb_release -sc)"
+    if [ "$?" -ne 0 ]; then
+        echo 'Unable to detect LSB codename' >&2
+        return 1
+    fi
+
+    if [[ "$release"  == trusty ]]; then
+        sed -i 's/Prompt=.*/Prompt=lts/' "$upgrade_config"
+    else
+        sed -i 's/Prompt=.*/Prompt=never/' "$upgrade_config"
+    fi
 }
+
 remove_2fa_tty_req() {
     # The goal here is to remove legacy 2FA req on TTY logins
     # Lets prevent this from bombing out the install though if it fails
@@ -70,7 +88,7 @@ case "$1" in
 
     manage_tor_repo_config
     remove_2fa_tty_req
-    disable_upgrade_prompt
+    update_release_prompt
 
     # Remove cron-apt action should occur after security upgrades to avoid breaking
     # automatic upgrades (see issue #4003)
--- a/install_files/securedrop-config/DEBIAN/postinst
+++ b/install_files/securedrop-config/DEBIAN/postinst
@@ -4,23 +4,32 @@
 set -e
 set -x
 
+update_release_available_script() {
+    # The script /etc/cron.weekly/update-notifier-common runs the command
+    # /usr/lib/ubuntu-release-upgrader/release-upgrade-motd which runs the command
+    # /usr/lib/ubuntu-release-upgrader/check-new-release whose output is written to the "stamp" file
+    # /var/lib/ubuntu-release-upgrader/release-upgrade-available which is picked up by OSSEC.
+    #
+    # To prevent the OSSEC alerts from from telling the user to run 'do-release-upgrade' which
+    # may break their system, we update both the script and the existing "stamp" file.
+
+    for file in /usr/lib/ubuntu-release-upgrader/check-new-release /var/lib/ubuntu-release-upgrader/release-upgrade-available; do
+        if [ -f $file ]; then
+            sed -i "s|Run 'do-release-upgrade' to upgrade to it\\.|Visit https://securedrop.org/xenial-upgrade for more information|" "$file"
+        fi
+    done
+}
 
 # Issue #4104
 # Set Prompt=never on Xenial
 # Set Prompt=lts on Trusty
 update_release_prompt() {
     set -e
+    upgrade_config='/etc/update-manager/release-upgrades'
 
-    declare -r upgrade_config='/etc/update-manager/release-upgrades'
-
-    declare -r release="$(lsb_release -sc)"
-    if [ "$?" -ne 0 ]; then
-        echo 'Unable to detect LSB codename' >&2
-        return 1
-    fi
-
-    if [[ "$release"  == trusty ]]; then
+    if [ "$(lsb_release -sc)" = trusty ]; then
         sed -i 's/Prompt=.*/Prompt=lts/' "$upgrade_config"
+        update_release_available_script
     else
         sed -i 's/Prompt=.*/Prompt=never/' "$upgrade_config"
     fi
--- a/install_files/securedrop-config/DEBIAN/postinst
+++ b/install_files/securedrop-config/DEBIAN/postinst
@@ -18,6 +18,11 @@ update_release_available_script() {
             sed -i "s|Run 'do-release-upgrade' to upgrade to it\\.|Visit https://securedrop.org/xenial-upgrade for more information|" "$file"
         fi
     done
+
+    # remove the file in case it's empty
+    rm -f /var/lib/ubuntu-release-upgrader/release-upgrade-available
+    # force re-run the update script to trigger an OSSEC alert
+    /usr/lib/ubuntu-release-upgrader/check-new-release -q > /var/lib/ubuntu-release-upgrader/release-upgrade-available &
 }
 
 revert_update_release_available_script() {
--- a/install_files/securedrop-config/DEBIAN/postinst
+++ b/install_files/securedrop-config/DEBIAN/postinst
@@ -20,6 +20,14 @@ update_release_available_script() {
     done
 }
 
+revert_update_release_available_script() {
+    for file in /usr/lib/ubuntu-release-upgrader/check-new-release /var/lib/ubuntu-release-upgrader/release-upgrade-available; do
+        if [ -f $file ]; then
+            sed -i "s|Visit https://securedrop\\.org/xenial-upgrade for more information|Run 'do-release-upgrade' to upgrade to it.|" "$file"
+        fi
+    done
+}
+
 # Issue #4104
 # Set Prompt=never on Xenial
 # Set Prompt=lts on Trusty
@@ -32,6 +40,7 @@ update_release_prompt() {
         update_release_available_script
     else
         sed -i 's/Prompt=.*/Prompt=never/' "$upgrade_config"
+        revert_update_release_available_script
     fi
 }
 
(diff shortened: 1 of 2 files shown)


conorsch:
These changes appear well structured, based on visual review. Will step through the test plan to validate functionality. Flagging that we should add config tests to verify the LTS upgrade strings are set correctly on both Trusty & Xenial.

redshiftzero:
We should also verify as part of review that the OSSEC notification that is triggered includes the expected text: `"* Visit https://securedrop.org/xenial-upgrade for more information"`

conorsch:
Stepped through Trusty test plan today:


* [x] Build trusty debs
* [x] Install on trusty box
* [x] Verify that /etc/update-manager/release-upgrades contains Prompt=lts
* [x] Verify that /usr/lib/ubuntu-release-upgrader/check-new-release contains "Visit https://securedrop.org/xenial-upgrade for more information" 
* [ ] :red_circle: and so does /var/lib/ubuntu-release-upgrader/release-upgrade-available if it exits)
* [ ] OSSEC notification that is triggered includes the expected text: "* Visit https://securedrop.org/xenial-upgrade for more information"

Will leave these VMs running overnight to see if OSSEC emails contain the correct text (although I used staging machines, I made sure to configure OSSEC via email). Given that the `/var/lib/ubuntu-release-upgrader/release-upgrade-available` file didn't satisfy the test plan (file was zero-bytes on both app-staging and mon-staging), perhaps not.

emkll:
Went through the Xenial test plan, LGTM: 
- [x] Build xenial 
- [x] Install on xenial box
- [x] Verify that `/etc/update-manager/release-upgrades` contains `Prompt=never`
- [x] Verify that `/usr/lib/ubuntu-release-upgrader/check-new-release` does not contain "Visit https://securedrop.org/xenial-upgrade for more information" (and neither does  `/var/lib/ubuntu-release-upgrader/release-upgrade-available` if it exits)

Also did an upgrade test under trusty, since these will be shipped to Trusty installs via the securedrop-config package, LGTM:
- [x] local upgrade successful
- [x] Verify that `/etc/update-manager/release-upgrades` contains `Prompt=never`
- [x] Verify that `/usr/lib/ubuntu-release-upgrader/check-new-release` does not contain "Visit https://securedrop.org/xenial-upgrade for more information" (and neither does  `/var/lib/ubuntu-release-upgrader/release-upgrade-available` if it exits)

I am (seemingly unrelated) error while upgrading, likely introduced by #4080, for which I will open an issue/PR:
```
+ pip install --no-index --find-links=/var/securedrop/wheelhouse --upgrade -r /var/www/securedrop/requirements/securedrop-app-code-requirements.txt
Ignoring indexes: https://pypi.python.org/simple/
Requirement already up-to-date: alembic==0.9.9 in /usr/local/lib/python2.7/dist-packages (from -r /var/www/securedrop/requirements/securedrop-app-code-requirements.txt (line 7))
Could not find any downloads that satisfy the requirement argon2-cffi==18.1.0 in /usr/local/lib/python2.7/dist-packages (from -r /var/www/securedrop/requirements/securedrop-app-code-requirements.txt (line 8))
Downloading/unpacking argon2-cffi==18.1.0 (from -r /var/www/securedrop/requirements/securedrop-app-code-requirements.txt (line 8))
Cleaning up...
No distributions at all found for argon2-cffi==18.1.0 in /usr/local/lib/python2.7/dist-packages (from -r /var/www/securedrop/requirements/securedrop-app-code-requirements.txt (line 8))
Storing debug log for failure in /home/vagrant/.pip/pip.log
 subprocess installed post-installation script returned error exit status 1
 securedrop-app-code
```

heartsucker:
Someone other than myself or @conorsch needs to sign on off on this since now we've both pushed here and @conorsch would be the primary reviewer. I've look at his additions, and it looks good. @emkll wanna give us the thumbs up here?

emkll:
Happy to do so, @heartsucker . 

I am also seeing the same behavior @conorsch is experiencing when upgrading in trusty:

After running the cron.weekly updater-notifier-common, `/var/lib/ubuntu-release-upgrader/release-upgrade-available` is 0 bytes. No OSSEC emails are sent as a result.

heartsucker:
I'm in the middle of another PR review right now so I can't start up a VM, but I think what's happening is that the permissions are wrong on that file and it's breaking things because cron cant' access it.

heartsucker:
I think what's happening is in some cases the `/usr/lib/ubuntu-release-upgrader/release-upgrade-motd` part of all of the above runs with a `&` (in the background), so the file `/var/lib/ubuntu-release-upgrader/release-upgrade-available` doesn't exist yet when we `sed` it. On a staging machine, if I remove that file then run `dpkg-reconfigure securedrop-config`, everything seems to work. Should we add a force `rm` to that in the `postinst` just to ensure we blow it up to trigger it's recreation? I'm still not totally sure how this ends up in the OSSEC alerts (plugin/rule?) but I think this should do the trick.

heartsucker:
I just added a line that should trigger all the above logic. If that's not actually the fix, feel free to remove it.

conorsch:
Confirming that the `/var/lib/ubuntu-release-upgrader/release-upgrade-available` file is now correctly populated for me. Built twice on Trusty and it was true both times.

Why the `&` in the postinst, though? If it's to avoid aborting the script due to a non-zero exit code, then we should use `|| true` instead. If it's to avoid long run times of the postinst, it appears to save only 1-1.5s. At first I was concerned we had a race condition here, and we'd need a corresponding `wait` for that process to exit, but as long as we don't care about success, it doesn't seem cause problems.

Have not yet verified the Xenial story functionality, nor confirmed the OSSEC emails, so more testing still required.

emkll:
Thanks for the fix @heartsucker ! I can also confim that `/var/lib/ubuntu-release-upgrader/release-upgrade-available` is now properly populated in the trusty scenario.
Also tested again in Xenial and the test plan is successful. 

The final piece is the ossec alert and unfortunately I still have not received an ossec alert requesting me to upgrade. I do get a syscheck alert informing me that the file was changed.

I can't seem to find a relevant ossec alert/rule for this in the upstream repo [0]
[0] : https://github.com/ossec/ossec-hids/tree/master/etc/rules


The easiest path would be to add a custom ossec notification (that would require xenial-specific logic to remove):
```
diff --git a/install_files/securedrop-ossec-agent/var/ossec/etc/ossec.conf b/install_files/securedrop-ossec-agent/var/ossec/etc/ossec.conf
index 9b14f4a81..651fff66d 100644
--- a/install_files/securedrop-ossec-agent/var/ossec/etc/ossec.conf
+++ b/install_files/securedrop-ossec-agent/var/ossec/etc/ossec.conf
@@ -115,6 +115,12 @@
   </localfile>
 
   <localfile>
+    <log_format>full_command</log_format>
+    <command>/usr/lib/ubuntu-release-upgrader/check-new-release -q</command>
+    <frequency>864900</frequency> <!-- 24 hours -->
+  </localfile>
+
+  <localfile>
     <log_format>syslog</log_format>
     <location>/var/log/kern.log</location>
   </localfile>
diff --git a/install_files/securedrop-ossec-server/var/ossec/rules/local_rules.xml b/install_files/securedrop-ossec-server/var/ossec/rules/local_rules.xml
index 97f0cfe62..9994f6c37 100644
--- a/install_files/securedrop-ossec-server/var/ossec/rules/local_rules.xml
+++ b/install_files/securedrop-ossec-server/var/ossec/rules/local_rules.xml
@@ -200,4 +200,11 @@
   </rule>
 </group>
 
-
+<group name="Daily dist upgrade notifications">
+  <rule id="400700" level="7" >
+    <if_sid>530</if_sid>
+    <options>alert_by_email</options> <!-- force email to be sent -->
+    <match>ossec: output: '/usr/lib/ubuntu-release-upgrader/check-new-release -q</match>
+    <description>This alert is to inform admins to upgrade to Xenial</description>
+  </rule>
```

heartsucker:
> Why the `&` in the postinst, though?

That Python script will download the `do-release-upgrade` and other required packages in the background it seems. I think it can be a very heavy process, and it's ok if it doesn't exit in `postinst`. There were some posts complaining about it being very slow so it seems like to avoid that we don't want to run in the foreground.

heartsucker:
About that OSSEC rule, I think we could ship either two different deb packges or two different conf files and then move the right one into place on `postinst`. I guess we could also apply/unapply the diff with `diff` too. Thoughts?

heartsucker:
So as this stands, we don't have the OSSEC rule for this. Since @conorsch seems to do the most Debian packaging stuff here, it would be good to get your feedback on whether or not we should update those rules via some `postinst` logic.

emkll:
Given the short timeline for 0.12.0 and the complexities around the management of separate ossec rules, we could also write a very simple cron job on the mon server that simply calls `send_encrypted_alarm.sh` to send an email to the admin, only if the base os is Trusty.

heartsucker:
@emkll, take a peeksy at the change I pushed. This should do the thing.

emkll:
@heartsucker thanks for the change. We've discussed this in the engineering meeting and concluded that perhaps more research was required on the ossec front. Any objections to removing ce49036 and opening a follow up to monitor/update the ossec rule? We can keep the scope of this PR to only the release-upgrades for the purposes of admins running  `do-release-upgrade`
The release-upgrades is a very important/significant change, and might be best to merge just that part for QA, and following up on the ossec portion later.
What do you think?

heartsucker:
I think that's reasonable. I'll for push with this change yanked.

heartsucker:
@kushaldas Can you stamp this one again? I added a commit since your review, then removed it, so it's unchanged since your last approval.

PR #4133, opened 2019-02-14
fix restore logic to recreation of onion services

### LINKED ISSUE #3960: [xenial] Verify Trusty backup -> Xenial recovery story
We should ensure that a SecureDrop backup completed on 14.04 can successfully be recovered on 16.04. Whether we ask admins to perform this step manually, or whether we automate it, it may be a required part of the Xenial migration and will certainly be highly recommended.

If clean upgrades to Xenial are not yet implemented one should complete this ticket by following these steps instead:

1. Create a backup on 14.04 server
2. Create fresh install on 16.04 and then attempt to run the restore

Part of #3204, may result in follow-up issues.

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #3960

Alter the restore logic to ensure that the onion services from the backup are what end up running on prod, and not the new values that may have been generated by Ansible.

## Testing

- Provision 0.11.1 machine (trusty)
- Run backup script
- Provision 0.12.0 machine (xenial)
- Run restore script
- Ensure restore script exist successfully
- Ensure tor ssh access to machine using values from 0.11.1 backup
- Ensure web app "just works"

## Deployment

This does not change deployment, but it does change restore from a previous deployment. The admin will run the same commands on xenial as on trusty.


install_files/ansible-base/roles/restore/tasks/main.yml
```diff
@@ -8,15 +8,15 @@
 
 - name: Copy backup tarball of config info to Application Server.
   copy:
-    src: "{{ restore_file }}"
-    dest: /tmp/{{ restore_file }}
+    src: /home/heartsucker/Downloads/sd-backup-2019-02-14--11-34-21.tar.gz #"{{ restore_file }}"
```
emkll: Are these the values that were used for testing?

heartsucker: lol crap yes :neutral_face:

heartsucker:
While I was eating lunch I realized on deficiency here is that the ATHS values aren't copied back to the admin workstation, and the torrc isn't updated. This would leave an admin locked out. Part of the restore role should include re-provisioning the workstation.

heartsucker:
@emkll Does that mean that the issue I raised above is not present? I haven't done a Tails install in a while and this seemed like it might cause issues.

emkll:
@heartsucker : based on my local testing on this branch if you run `securedrop-admin install` right after performing the restore, it will fetch the onion URLs from the server and overwrite the local ones for the Journalist and Source interfaces. I pushed d1fe88f to clarify the documentation, let me know if you disagree.

heartsucker:
Ok I think that is sufficient for this. I'm giving the thumbs up to that docs change, so you can approve this or we can wait for another homie to come by and do the deed.

install_files/ansible-base/roles/restore/files/restore.py
```diff
@@ -35,18 +36,29 @@ def verify_args():
 def main():
     verify_args()
 
+    # Remove the /var/lib/tor/services directories to purge values that may have been
+    # generated by running the ansible playbooks
+    for d in ['journalist', 'source']:
```
redshiftzero: @heartsucker @emkll: for posterity can you explain why this change was necessary? Does the stanza below not overwrite the `/var/lib/tor/services` config files on disk in the correct locations?

emkll: In retrospect, it does not seem strictly necessary, unless I am missing something: the tar extraction should squash existing values on the server

heartsucker: I'm pretty sure V3 onion services add additional files, and I can't remember how `tar -x` works and if those extra files would cause bad behavior in the future.

redshiftzero: Based on the commit history and comments, it's not clear why this stanza is needed... it seems like a reasonable approach here is to determine if it's necessary. If it is, let's document why (at least in the PR). If we actually don't need this stanza, then we can just remove it in the spirit of reducing LOC to maintain.

PR #4135, opened 2019-02-14
Fix file rename bug

creviera:
## Status

Ready for review

## Description of Changes

Fixes https://github.com/freedomofpress/securedrop-client/issues/238

No longer update the names of old submissions in the `submissions` table when `journalist designation` changes. If we didn't do this, we would either have to keep track of the previous journalist designations or change names of files in the `data` directory for the client to fix this bug.

## Testing

1. Run the client
2. Send a message or document as a source
3. Change that source's name from the journalist interface
4. The old messages should no longer show "Content deleted"


creviera:
Need to think about this more.

Issue #4155, opened 2019-02-19
Make Xenial development/CI environments the default


redshiftzero:
## Description

New installs beginning with 0.12.0 will be done on Xenial. Xenial will become the primary base OS for SecureDrop, and we will deprecate support for Trusty when it EOLs in April. We should thus:

- [x] Make the default CI staging job that is run per-PR xenial (#4158)
- [x] Make the trusty CI staging job run nightly (#4158)
- [x] Rename the makefile targets that are `staging-xenial` and `test-xenial` just `staging` and `test` (renaming the existing `staging` to `staging-trusty` so that we keep them around to ensure we don't break anything for installs that have yet to migrate) (#4224)
- [x] Make `BASE_OS` default be `xenial` instead of `trusty` in the securedrop application Makefile (#4213)
- [x] Update and test Qubes staging setup docs on Xenial base OS (dev only)
- [x] Update Vagrantfile to use Xenial instead of Trusty (probably only used for Prod VMs at this point) (#4213)
- [x] Port "upgrade" scenario to Xenial, so VM upgrade testing can be performed on Xenial by default (#4237)
- [x] Update default staging target to Xenial (#4310)

redshiftzero:
everything here was done, anything we forgot should get a new dedicated issue

PR #4157, opened 2019-02-19
[xenial] Update fresh install guide

### LINKED ISSUE #4103: [xenial] Update the install guide to account for the use of Xenial instead of Trusty as the server base OS
## Description

The current install docs for SecureDrop apply to Trusty-based installations only. As part of the Xenial update, they should be modified to remove references to Trusty and ensure that  installations post 0.12.0 are completed on top of Xenial instead.

## User Stories

As a SecureDrop administrator, I would like to install a system with an OS that has LTS.

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #4103

- Updates for new installs on Xenial (ran through this on fresh install on hardware on Mac Minis to verify there was no other deviation)

## Testing

- [ ] Verify there are no lingering references in the install guide to trusty 

## Deployment

Applies to new installs only


docs/development/qubes_staging.rst
```diff
@@ -58,7 +58,7 @@ In ``dom0``:
 You may need to edit the filepath above if you downloaded the ISO to a
 different location within the ``sd-dev`` VM. Choose **Install Ubuntu**.
 For the most part, the install process matches the
-:ref:`hardware install flow <install_trusty>`, with a few exceptions:
+:ref:`hardware install flow <install_ubuntu>`, with a few exceptions:
 
   -  Server IP address: use value returned by ``qvm-prefs sd-trusty-base ip``, with ``/24`` netmask suffix
```
kushaldas: Should we still keep all the vm names as *trusty? I think we should update those too.

redshiftzero: so this PR doesn't instruct people to use Xenial for the Qubes staging VMs, that's tracked as a task in #4155

PR #4158, opened 2019-02-19
[ci] move trusty app tests to nightly job and use circle ci branch filtering logic

### LINKED ISSUE #3699: Docs-only PRs from forks still run staging env in CI
## Description

CI shouldn't run the staging tests when a PR is prefixed with `docs-`. At least in the case of PRs from forks, it seems to be doing so.

## Steps to Reproduce

1. Take a gander at #3670, and confirm the PR branch begins with `docs-`.
2. Inspect the [associated CI build for that PR](https://circleci.com/gh/freedomofpress/securedrop/16688) and confirm it did in fact run the entire staging scenario.

## Expected Behavior

CI does not run staging env when branch name for PR begins with `docs-`. 

## Actual Behavior

On at least one fork, CI did run the staging env, despite the fact that the branch name for the PR began with `docs-`.

## Comments

It would be helpful to have a verbose wrapper script to make inspecting CI state easier. The script could print informative messages such as "Detected branch name 'docs-libvirt-setup', skipping staging run...".

redshiftzero:
## Status

Ready for review

## Description of Changes

This PR makes two improvements to reduce total CI time:
 - Beginning with 0.12.0, news organizations will be migrating to Xenial, and all new installs will be on Xenial. Thus, we should move the Trusty app test job to a nightly job, and monitor it for failures until Trusty is deprecated in the coming months. This reduces total CI time per PR, as the trusty job took 20+ minutes. Related to #4155.
 - It turns out that [CircleCI enables filtering by branches now](https://circleci.com/docs/2.0/configuration-reference/#branches), which we can use to replace our custom logic for avoiding unnecessary jobs on `docs-` branches, currently used for the staging job only. The app-tests were also being run for `docs-` branches, so we can use the same CircleCI branch filtering logic to easily exclude the staging/application test jobs run for `docs-` branches. Indeed, previously jobs would be kicked off and then would rapidly pass, but with the CircleCI branch filtering, jobs won't even get kicked off, so this should reduce build pileups. This _should_ also close #3699.

Note: We'll need to disable the trusty app test job as a required build to merge, holding off on that until we agree

# Testing

Verify that the test build [here](https://github.com/freedomofpress/securedrop/commits/docs-ci-reduce-jobs) (branch `docs-ci-reduce-jobs`) only ran two jobs:

screen shot 2019-02-19 at 11 1
0 54 pm

## Deployment

CI only

## Checklist

CI only

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -257,7 +257,6 @@ workflows:
   securedrop_ci:
     jobs:
       - lint
-      - trusty-app-tests
       - xenial-app-tests
       - admin-tests
       - updater-gui-tests
@@ -276,3 +275,4 @@ workflows:
     jobs:
       - static-analysis-and-no-known-cves
       - staging-test-with-rebase-xenial
+      - trusty-app-tests
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -257,11 +257,27 @@ workflows:
   securedrop_ci:
     jobs:
       - lint
-      - xenial-app-tests
-      - admin-tests
-      - updater-gui-tests
+      - xenial-app-tests:
+          filters:
+            branches:
+              ignore:
+                - /docs-.*/
+      - admin-tests:
+          filters:
+            branches:
+              ignore:
+                - /docs-.*/
+      - updater-gui-tests:
+          filters:
+            branches:
+              ignore:
+                - /docs-.*/
       - static-analysis-and-no-known-cves
       - staging-test-with-rebase:
+          filters:
+            branches:
+              ignore:
+                - /docs-.*/
           requires:
             - lint
   nightly:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -273,7 +273,7 @@ workflows:
               ignore:
                 - /docs-.*/
       - static-analysis-and-no-known-cves
-      - staging-test-with-rebase:
+      - staging-test-with-rebase-xenial:
           filters:
             branches:
               ignore:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -257,12 +257,27 @@ workflows:
...
+            branches:
+              ignore:
+                - /docs-.*/
+      - updater-gui-tests:
+          filters:
+            branches:
+              ignore:
+                - /docs-.*/
       - static-analysis-and-no-known-cves
       - staging-test-with-rebase:
```
emkll: Should we also run Xenial infra tests on each PR instead of the default (Trusty) tests?

redshiftzero:
Once this is approved, I'll remove the requirement for `ci/circleci: staging-test-with-rebase` and `ci/circleci: trusty-app-tests` in favor of the corresponding Xenial jobs

devops/gce-nested/ci-go.sh
```diff
@@ -16,11 +16,6 @@ set -o pipefail
 # Assume we're running against Trusty; Xenial also supported.
 target_platform="${1:-trusty}"
```
emkll: Should we open a follow-up task to completely eliminate trusty references at some point in the future, and/or replace the defaults with xenial?

redshiftzero: so I have task #4155 filed for this which this PR does not close (wanted to get this PR in first so that we can not trigger quite so many builds since we're getting a lot of queueing in circleCI)

PR #4169, opened 2019-02-21
Ensure version of tor installed is from FPF repo

### LINKED ISSUE #4162: [0.12.0] Validate Tor version at install time
## Description
If the apt server used for a SecureDrop install (e.g., apt.freedom.press) does not serve a `tor` package, the install logic will install the Tor package from  Ubuntu upstream, which is quite old (0.2.4.27). It is the only package that mirrored on apt.freedom.press that also exists in upstream Ubuntu apt repos.

We should implement a verification in the install logic that ensures the version of Tor at install time is the one provided by FPF (and not the upstream-provided version).

I have added it to the pre-flight checklist for 0.12.0: https://github.com/freedomofpress/securedrop/issues/4124

emkll:
## Status

Ready for review 

## Description of Changes

Fixes #4162 . 

Because we don't want to couple Tor versions with SecureDrop releases, and that we probably won't go back in Tor versions, specifying  `>=` the current Tor version seems like a sensible approach, given the versions being served in Ubuntu Trusty and Xenial apt repos:
- 0.2.4.27 (Trusty)
- 0.2.9.14 (Xenial)

## Testing
- [ ] The test/assertion introduced make sense, we don't want to pin specific tor versions to SecureDrop releases.
- [ ] Using prod VMs or hardware, check out this branch and attempt installing against prod apt servers. This should fail because it will be installing Tor from Ubuntu's repos, as apt-tor.freedom.press was removed.
- [ ] Using prod VMs or hardware, check out this branch and attempt installing against test apt server (apt-test.freedom.press). This should succeed because Tor packages are served by the test apt server.

CI/staging scenarios are not an accurate test as they also include apt-test in their source list (for kernel tests).

## Deployment

This change only affects new installs, and will be provided via the Ansible install logic.

### DIFF
--- a/install_files/ansible-base/roles/tor-hidden-services/tasks/install_tor.yml
+++ b/install_files/ansible-base/roles/tor-hidden-services/tasks/install_tor.yml
@@ -37,7 +37,15 @@
     apt-cache policy tor | sed -e 's/^\s*Installed:\ \(\S*\)/\1/g;tx;d;:x'
   changed_when: false
   register: extract_tor_version
-  when: "'amazon' in ansible_product_version"
+
+# Ubuntu upstream repositories serve a version of tor that is very old. Since
+# FPF apt servers host this same package, let's ensure that the FPF-provided
+# Tor package is installed by checking we are using a recent version.
+
+- name: Ensure correct Tor version installed.
+  assert:
+    that: extract_tor_version.stdout is version('0.3.4.9', '>=')
+    fail_msg: "Tor package was not found on FPF apt server."
 
 - name: Dump Tor version to file (for reporting)
   copy:


install_files/ansible-base/roles/tor-hidden-services/tasks/install_tor.yml
```diff
@@ -37,7 +37,15 @@
     apt-cache policy tor | sed -e 's/^\s*Installed:\ \(\S*\)/\1/g;tx;d;:x'
   changed_when: false
   register: extract_tor_version
-  when: "'amazon' in ansible_product_version"
+
+# Ubuntu upstream repositories serve a version of tor that is . Since
```
redshiftzero: This comment trails off - maybe "a version of tor that is very old"?

Otherwise I agree with the logic here... now stepping through the test plan

redshiftzero:
✅Using prod VMs, check out this branch and attempt installing against prod apt servers. Fails as expected.
✅Using prod VMs, check out this branch and attempt installing against test apt server (apt-test.freedom.press). Succeeds as expected.

Gonna push a wee change to resolve that nit inline...

PR #4172, opened 2019-02-21
Clean up restore script and clarify restore docs, follow up to 4133

### LINKED ISSUE #4171: [0.12.0] Followup tasks for backup and restore
## Description

This is a followup from #4133 to:

- [ ] Update the docs to indicate that `securedrop-admin install` [here](https://github.com/freedomofpress/securedrop/pull/4133/files#diff-ed8769ff4c99e7e369511dcbe0ee78f9R159) will only fetch back the *THS files for SSH over LAN, not SSH over Tor. We should indicate in the docs for admins that have the default SSH over Tor configured, that they need to manually extract the *THS files from the restore tarball.
- [ ] Remove or document this stanza: https://github.com/freedomofpress/securedrop/pull/4133/files#diff-ff45d239e10e0aa98937629a72d83336R41

emkll:
## Status

Ready for review

## Description of Changes

Fixes #4171 

## Testing
- [ ] Restore logic is functional
- [ ] Docs properly reflect the workstation re-provisioning scenario once ssh access to the servers has been established.

## Deployment

This PR is limited to docs and Ansible logic shipped to Admin Workstations via git.


zenmonkeykstop:
I don't think this is necessarily the way to go about this - and it adds an extra `./securedrop-admin install` step after the restore.

If folks are restoring from a backup, their naive expectation would be that the config including JI and SI URLs would be those from the backup that they're restoring. This looks like it would remove them and then require you to add them again from the admin workstation via `install`.

It would be better IMO, if the `restore` command copied the appropriate files into /var/lib/tor/services on the app server and then created the appropriate app server \*ths files in ~/Persistent/securedrop/install_files/ansible-base based on the service definitions from the backup tarball.

redshiftzero:
So the current behavior with or without this PR is that the restore will copy the *ths files into `/var/lib/tor/services`. The stanza of code being removed in this diff is being removed due to it being unnecessary (see https://github.com/freedomofpress/securedrop/pull/4133/files#r258292925). 

I agree with you @zenmonkeykstop regarding the desired behavior of `securedrop-admin restore`: it should first save the old `ths` values (just in case) somewhere in the Admin Workstation and then write the new ones to disk based on the contents of the restore tarball. Since this is an improvement and we are in feature freeze, I recommend we do this in a subsequent release and target just keeping the restore working as is both trusty->trusty and trusty->xenial, with sufficient docs, for 0.12.0. With respect to the `securedrop-admin install`, perhaps just removing the paragraph about `securedrop-admin install` from the documentation, and instead instructing admins that they can get the onion service information from the tarball is the way to go. Thoughts?

heartsucker:
A V3 Onion service creates the following files:

```
authorized_clients
hostname
hs_ed25519_public_key
hs_ed25519_secret_key
```

A V2 Onion service only creates these

```
hostname
private_key
```

I think it is useful to explicitly remove those files since we will never need them.

emkll:
Makes sense @heartsucker . The current torrc template does however guard against the creation of V3 services https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/tor-hidden-services/templates/torrc by specifying the hidden service version, and unlikely that v3 onions would be created on the app server otherwise. This more destructive approach could introduce potentially confusing and destructive edge cases for admins on restore (if we don't change the script when we do the v2->v3 migration). 

Ultimately, I don't feel very strongly about this change, provided the restore functionality works as expected.

heartsucker:
> if we don't change the script when we do the v2->v3 migration

So whenever we do this, we should run parallel support of v2 and v3 since a lot of newspapers have published print onion addresses, so the deprecation plan for this should be very long. We should create new V3 services in different directories with and run SD with both v2 and v3 for a while.

Because I had this thought in mind, this seemed safe.

> by specifying the hidden service version, and unlikely that v3 onions would be created on the app server otherwise

However, this is also correct so I think this is fine to merge.

PR #4184, opened 2019-02-25
[ci] run page layout tests in all supported languages for i18n-* branches

### LINKED ISSUE #4183: add CI job for pre-release translation merges
## Description

Before SecureDrop regular releases, we merge translations from the fork that Weblate commits to back into this repository. The person who merges the translations needs to run the page layout test suite in all supported languages to ensure that no HTML/functionality was broken by the translated strings. This test suite currently takes 1+ hours and is run locally on developer machines. Instead of manually running these tests, we should run them in CI in a new job that is executed only for translation merges, e.g. those branches that are prefixed with `i18n-`.

## User Stories

As a SecureDrop release manager and/or localization manager, I want to reduce the manual work I am doing to prevent errors and to allocate my time to more interesting tasks.

redshiftzero:
## Status

Ready for review

## Description of Changes

Closes #4183

## Testing

- [ ] Ensure the workflow kicked off by this branch has only two jobs: `translation-tests` and `lint`. This is because this branch is prefixed with `i18n-`.
- [ ] The above two jobs should pass.
- [ ] Ensure the workflow kicked off by `normal-ci-jobs-should-run` contains the normal CI jobs - this branch is identical to this one except for an additional dummy commit (added otherwise CircleCI will display the CI job results in the checks for this PR since they are tracked by commit hash and not branch name).

## Deployment

This is CI only for developer/release agility

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,46 @@
 ---
+common-steps:
+  - &createcachedir
+    run:
+      name: Ensure cache dir exists and permissions are good
+      command: |
+        sudo mkdir -p /caches && sudo chown circleci: -R /caches
+
+  - &restorecache
+    restore_cache:
+      key: v1-sd-layers-{{ checksum "securedrop/dockerfiles/xenial/Dockerfile" }}
+      paths:
+        - /caches/layers.tar.gz
+
+  - &loadimagelayers
+    run:
+      name: Load image layer cache
+      command: |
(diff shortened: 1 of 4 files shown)


redshiftzero:
hey @kushaldas check this one out - it would be cool if for the final 0.12.0 release we stop running the page layout tests locally and instead rely on CI to speed up the release process a bit. what do you think?

PR #4186, opened 2019-02-25
Adds new languages: ro is

kushaldas:
## Status

Ready for review

## Description of Changes

Adding new languages Romanian and Icelandic.

## Testing

While merging back the translations from weblate, we should have commits for these two languages.

## Deployment

None, dev only.


securedrop/i18n_tool.py
```diff
@@ -40,10 +40,12 @@ class I18NTool(object):
         'fr_FR': {'name': 'French', 'desktop': 'fr', },
         'el': {'name': 'Greek', 'desktop': 'el', },
         'hi': {'name': 'Hindi', 'desktop': 'hi', },
+        'is': {'name': 'Icelandic', 'desktop': 'is', },
         'it_IT': {'name': 'Italian', 'desktop': 'it', },
         'nb_NO': {'name': 'Norwegian', 'desktop': 'nb_NO', },
         'nl': {'name': 'Dutch', 'desktop': 'nl', },
         'pt_BR': {'name': 'Portuguese, Brasil', 'desktop': 'pt_BR', },
+        'ro': {'name': 'Romanian', 'desktop': 'ro', },
```
redshiftzero: two other places we should update: 

docs here: https://github.com/freedomofpress/securedrop/blob/develop/docs/includes/l10n.txt
locales to test against in CI here: https://github.com/freedomofpress/securedrop/blob/develop/securedrop/bin/translation-test#L20

kushaldas: @redshiftzero I am guessing both of those should be updated after we merge the translations. Otherwise, the commands and tests will fail.

securedrop/i18n_tool.py
```diff
@@ -126,7 +126,7 @@ def translate_desktop(self, args):
         if args.compile:
             pos = filter(lambda f: f.endswith('.po'),
                          os.listdir(args.translations_dir))
-            linguas = map(lambda l: l.rstrip('.po'), pos)
+            linguas = map(lambda l: l[:-3], pos)
```
redshiftzero: given this change we should make this PR from a branch not prefixed in `i18n-`

redshiftzero:
Closing this in favor of #4193 - as described above I ran this from a branch not prefixed with `i18n-` such that we get full test runs in CI

PR #4218, opened 2019-03-01
extend staging timeout for long ansible tasks

### LINKED ISSUE #4220: CI timeout during package upgrade in xenial staging scenario
## Description

When upgrading distribution packages in staging, the CI run times out due to lack of output. This is likely due to a large amount of packages on the both app and mon due to nested virt

## Steps to Reproduce

Observe CI output of these runs

- https://circleci.com/gh/freedomofpress/securedrop/23846
- https://circleci.com/gh/freedomofpress/securedrop/23827
- https://circleci.com/gh/freedomofpress/securedrop/23886

```
    TASK [common : Perform safe upgrade to ensure all the packages are updated.] ***
make: *** [ci-go-xenial] Terminated
Warning: Permanently added '35.230.112.0' (ECDSA) to the list of known hosts.

scp: /home/sdci/securedrop-source/junit/*xml: No such file or directory
Too long with no output (exceeded 10m0s)
```

## Expected Behavior

Packages should be updated, no timeouts out

## Actual Behavior

Very often there is a timeout when upgrading packages, resulting in a failed `xenial-staging-with-rebase` step.

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes timeout errors in CI, closes #4220 

## Testing

Does CI pass? In particular this Ansible step:

```
Perform safe upgrade to ensure all the packages are updated
```

## Deployment

CI only.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -249,6 +249,7 @@ jobs:
       - run:
           name: Run Staging tests on GCE
           command: make ci-go
+          no_output_timeout: 20m
 
       - run:
           name: Ensure environment torn down
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -276,6 +276,7 @@ jobs:
       - run:
           name: Run Staging tests on GCE (Xenial)
           command: make ci-go-xenial
+          no_output_timeout: 20m
 
       - run:
           name: Ensure environment torn down


heartsucker:
Even with extending this to 20 minutes, we're timing out. I think we might be hitting some kind of interactive prompt, or we may need to update the base box to the latest version to minimize the number of packages that are being updated during the step that fails.

heartsucker:
Hmm actually this build passed that step, but without the extended timeout. So really I have no idea what's going on here. https://circleci.com/gh/freedomofpress/securedrop/23880?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

emkll:
Thanks @heartsucker for the fix. The problem indeed appears to be intermittent and might be dependent on network conditions. I've appended a commit 8243d9b to your branch, which should also cover the per-PR step, based on your changes:

The timeout was added to `staging-test-with-rebase`, whereas `staging-test-with-rebase-xenial` is step that is  being run per-PR (this was introduced very recently, see #4158). The naming is somewhat counter-intuitive, though Makefiles are to be renamed as part of #4155

PR #4231, opened 2019-03-06
Try harder to attach to an existing tmux session

### LINKED ISSUE #4221: During 14.04->16.04 upgrade, tmux is unable to re-attach
An organization performing a Xenial upgrade reported that they were unable to re-attach to the Application Server after the `sshd` restart. They attributed this to the fact that tmux itself was upgraded before then. They were able to manually re-attach by finding the old process ID via `pgrep` and re-attaching via `/proc/$pid/exe`. 

They did not experience the same issue upgrading the Monitor Server.

We should attempt to understand the root cause and, if warranted, add an update to the documentation to help users work around this issue.

rmol:
## Status

Ready for review

## Description of Changes

If in the course of an ssh/tmux session the tmux package is upgraded and the ssh connection broken, the next ssh attempt will encounter an error because of the tmux protocol version mismatch.

The old tmux can be used to reattach by searching for the tmux process under /proc and using its reference to the old tmux executable. This change attempts to do that automatically if a first "tmux attach" attempt fails, but we can see that a previous session still exists.

Fixes #4221.

## Testing

This is not deterministically reproducible, but I've had it happen on at least one of the servers during each upgrade from Trusty to Xenial, provided they are set up to use SSH over Tor. It seems more likely to occur on the app server, so try running `securedrop-admin install` with this branch, then execute `sudo do-release-upgrade` on the app server first. When you reconnect after libssl et al. are upgraded and Tor restarted, you should automatically resume the tmux session. If instead you get an error about the protocol mismatch, the fix hasn't worked.

### DIFF
--- a/install_files/securedrop-config/DEBIAN/control
+++ b/install_files/securedrop-config/DEBIAN/control
@@ -4,7 +4,7 @@ Priority: optional
 Maintainer: SecureDrop Team <securedrop@freedom.press>
 Homepage: https://securedrop.org
 Package: securedrop-config
-Version: 0.1.2+0.13.0~rc1
+Version: 0.1.3+0.13.0~rc1
 Architecture: all
 Description: Establishes baseline system state for running SecureDrop.
  Configures apt repositories.
--- a/install_files/securedrop-config/etc/profile.d/securedrop_additions.sh
+++ b/install_files/securedrop-config/etc/profile.d/securedrop_additions.sh
@@ -2,7 +2,12 @@
 
 which tmux >/dev/null 2>&1 || return
 
-function tmux_attach_via_proc() {
+tmux_attach_via_proc() {
+    # If the tmux package is upgraded during the lifetime of a
+    # session, attaching with the new binary can fail due to different
+    # protocol versions. This function attaches using the reference to
+    # the old executable found in the /proc tree of an existing
+    # session.
     pid=$(pgrep --newest tmux)
     if test -n "$pid"
     then
--- a/install_files/securedrop-config/etc/profile.d/securedrop_additions.sh
+++ b/install_files/securedrop-config/etc/profile.d/securedrop_additions.sh
@@ -1,0 +1,17 @@
+[[ $- != *i* ]] && return
+
+which tmux >/dev/null 2>&1 || return
+
+function tmux_attach_via_proc() {
+    pid=$(pgrep --newest tmux)
+    if test -n "$pid"
+    then
+        /proc/$pid/exe attach
+    fi
+    return 1
+}
+
+if test -z "$TMUX"
+then
+    (tmux attach || tmux_attach_via_proc || tmux new-session)
+fi
--- a/molecule/testinfra/staging/common/test_user_config.py
+++ b/molecule/testinfra/staging/common/test_user_config.py
@@ -1,6 +1,5 @@
-import hashlib
-import os
 import re
+import textwrap
 
 
 def test_sudoers_config(host):
@@ -41,18 +40,32 @@ def test_sudoers_tmux_env(host):
     """
 
     host_file = host.file('/etc/profile.d/securedrop_additions.sh')
-    source_file = os.path.abspath(
-        os.path.join(
-            os.path.dirname(__file__),
-            '../../../../install_files/securedrop-config/etc/profile.d',
-            'securedrop_additions.sh'
-        )
-    )
-    expected_content = open(source_file).read()
-    h = hashlib.sha256()
-    h.update(expected_content)
+    expected_content = textwrap.dedent(
+        """\
+        [[ $- != *i* ]] && return
 
-    assert host_file.sha256sum == h.hexdigest()
+        which tmux >/dev/null 2>&1 || return
+
+        tmux_attach_via_proc() {
+            # If the tmux package is upgraded during the lifetime of a
+            # session, attaching with the new binary can fail due to different
+            # protocol versions. This function attaches using the reference to
+            # the old executable found in the /proc tree of an existing
+            # session.
+            pid=$(pgrep --newest tmux)
+            if test -n "$pid"
+            then
+                /proc/$pid/exe attach
+            fi
+            return 1
+        }
+
+        if test -z "$TMUX"
+        then
+            (tmux attach || tmux_attach_via_proc || tmux new-session)
+        fi"""
+    )
+    assert host_file.content_string == expected_content
 
 
 def test_tmux_installed(host):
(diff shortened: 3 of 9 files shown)


emkll:
I could reliably reproduce the tmux  bug using prod VMs: setting the host to sleep or disconnecting the network would very reliably trigger the bug. I can confirm that for instances provisioned on this branch, this appears to fix the problem! :tada: 

However, the fix only gets applied for new installs, or installs on which the installer was run, as the bashrc additions are added to the servers as part of the `create_user` task of the `common` role.

Our documentation currently does not instruct users to run `./securedrop-admin install`  before beginning the upgrade process [0]. This file could be updated on running instances via the `securedrop-config` package, since it's installed on both machines. Is this problem severe/frequent enough to warrant us deploying the fix in an unattended fashion? Can someone think of another way to deploy this to existing instances given our March 19 release for 0.12.1?


[0] : https://docs.securedrop.org/en/latest/upgrade/xenial_upgrade_in_place.html

conorsch:
Agreed with @emkll's concerns: in order for this change to be useful to Admins performing the Trusty -> Xenial upgrade, we want the helper functions in place _before_ the upgrade is started—ideally without requesting yet another run of `./securedrop-admin install` prior to the upgrade. 

@rmol Take a stab at porting this new logic to the `securedrop-config` package (package files at `install_files/securedrop-config/`. Note that we already have a file for tmux settings written to `/etc/`; see here: https://github.com/freedomofpress/securedrop/blob/543357a5e691be62e4a19d87f46dc9e00a866e46/install_files/ansible-base/roles/common/files/bashrc_securedrop_additions

Those customizations, with the improvements you're presenting here, should be taken out of the Ansible logic and placed in the `securedrop-config` packaging logic. Haven't run the config tests on this branch yet, but I have a few comments there—will tack on in-line.

rmol:
Moved `/etc/profile.d/securedrop_additions.sh` creation to the `securedrop-config` package. Started with Trusty, installed the updated .deb manually, and confirmed that during `do-release-upgrade` I could reconnect and reattach. 

Also confirmed that having the existing file in `/etc/profile.d` doesn't cause problems when installing the updated package, either on Trusty or Xenial.

conorsch:
Beginning local test. Given that the change is implemented via debian package modifications, we must use staging to test. Intend to use the following test plan:

1. Run `rm -rf build/*` to ensure no old debs are present
2. Run `make build-debs-trusty` to create new debs used for testing
3. Set `enable_ssh_over_true: true` in `install_files/ansible-base/group_vars/staging.yml`
4. Run `make staging-trusty` to create Trusty VMs
5. Add `app-ssh-aths` info to host `torrc`
6. Open an interactive shell via SSH-over-Tor
7. Run `sudo do-release-upgrade` while connected over Tor 
8. Confirm able to reconnect to tmux after service bounces disconnect the session

Will report back when testing is done.

conorsch:
Finished test plan. Can confirm I'm able to reconnect to tmux after a service bounce on the App Server. Encountered a few test failures that are quick fixes, so I'll comment in-line with some pointers.

install_files/securedrop-config/DEBIAN/control
```diff
@@ -4,7 +4,7 @@ Priority: optional
 Maintainer: SecureDrop Team <securedrop@freedom.press>
 Homepage: https://securedrop.org
 Package: securedrop-config
-Version: 0.1.2+0.13.0~rc1
+Version: 0.1.3+0.13.0~rc1
```
conorsch: The version bump here is absolutely correct. It requires matching changes in other places, though:

* group_vars for staging (already handled)
* config tests vars (specifically, see `molecule/builder-trusty/tests/vars.yml`)

The rationale here is that we want to ensure we're installing _exactly_ a certain version, and also running the package checks on _exactly_ that same version. By default, multiple package versions can pile up in `build/`. We could automatically run `rm -rf build/*` to avoid the pile-up, but that seems a bit heavy-handed for the dev env.

molecule/testinfra/staging/common/test_user_config.py
```diff
@@ -38,16 +40,19 @@ def test_sudoers_tmux_env(host):
...
+    expected_content = open(source_file).read()
+    h = hashlib.sha256()
+    h.update(expected_content)
 
-    assert f.contains("^{}$".format(non_interactive_str))
-    assert f.contains("^if which tmux >\/dev\/null 2>&1; then$")
-
-    assert 'test -z "$TMUX" && (tmux attach || tmux new-session)' in f.content
-    assert f.contains(tmux_check)
+    assert host_file.sha256sum == h.hexdigest()
```
conorsch: Your motivation to DRY out the config tests here is understandable, but most of the config tests are _intentionally_ duplicative, to enable us to refactor the provision logic confidently. For example, if you modify `install_files/securedrop-config/etc/profile.d/securedrop_additions.sh` locally, then build packages and install them, you should reasonably expect the corresponding config tests to fail. These changes would cause them to pass. 

A broader discussion is warranted here about the pattern used for hardcoding values in the config tests. For now, simply for the sake of consistency, I'd prefer to keep them. Even a multiline string asserting an exact match on `host_file.contents` or `host_file.content_string` would be sufficient.

install_files/securedrop-config/etc/profile.d/securedrop_additions.sh
```diff
@@ -0,0 +1,17 @@
...
+    then
+        /proc/$pid/exe attach
+    fi
+    return 1
+}
+
+if test -z "$TMUX"
+then
+    (tmux attach || tmux_attach_via_proc || tmux new-session)
+fi
```
conorsch: Not requesting additional comments in this file because the logic is cleanly presented and the corresponding commit message is top-notch.

conorsch:
@kushaldas The wait timeout you describe should only occur if you've overridden `enable_ssh_over_tor: true`, as part of the modified test plan. The staging playbook isn't smart enough to read that var, which is a shortcoming, but needn't block additional testing. After the reboot, you can continue to follow the subsequent test plan steps.

install_files/securedrop-config/DEBIAN/control
rmol: Fixed `config_version` in vars.yml.

molecule/testinfra/staging/common/test_user_config.py
rmol: Reverted to comparing the exact current content.

install_files/securedrop-config/etc/profile.d/securedrop_additions.sh
rmol: That was a good not-request, though. No one should have to go find my top-notch commit message when trying to understand this. :smile:

conorsch:
New test report from @emkll looks great, and the additional config tests are a mighty fine addition. Running the build logic locally to confirm all tests passing, then will mark approved and merge.

PR #4243, opened 2019-03-06
Link directly to i18n repo contributors list in README

eloquence:
The intention is to make the names of all i18n contributors more
discoverable. We separately prepare a list of contributors to the
i18n for each release as part of the release notes.

## Status

Ready for review


eloquence:
Note that this page only shows users with GitHub accounts associated with their commits. It's IMO better than just linking to the repo (which doesn't tell you anything unless you know what you're looking for). There's also https://weblate.securedrop.org/stats/ which is only a top 10.

PR #4249, opened 2019-03-08
[docs] Make it clearer that key generation must be done on the SVS

### LINKED ISSUE #4076: [docs] Make it clearer that key generation must be done on the SVS
https://docs.securedrop.org/en/release-0.11.1/generate_securedrop_application_key.html currently does not explicitly state that the submission keypair must be created on the air-gapped SVS. This may be inferred from the sequence of steps, and from the conclusion ("At this point, you are done with the Secure Viewing Station for now"), but for the sake of avoiding ambiguity, we should make it explicit at the beginning of the document.

eloquence:
Resolves #4076

## Status

Ready for review

## Checklist

### DIFF
--- a/docs/generate_securedrop_application_key.rst
+++ b/docs/generate_securedrop_application_key.rst
@@ -7,7 +7,16 @@ of this key is only stored on the *Secure Viewing Station* which is never
 connected to the Internet. SecureDrop submissions can only be decrypted and
 read on the *Secure Viewing Station*.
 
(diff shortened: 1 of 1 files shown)


eloquence:
FWIW, I debated with myself on whether an "important" banner is warranted, then decided in favor of it; I felt that this document was a good opportunity to being reinforcing the idea that the private key is especially sensitive and that its generation warrants additional attention.

PR #4250, opened 2019-03-08
Adds instructions on verifying server locales before running install playbook

### LINKED ISSUE #4219: Ansible scripts depend on English locale
## Description

After the upgrade to xenial and 0.12.0, ./securedrop-admin install fails on the ‘tor-hidden-services: Ensure correct Tor version installed’ step.

It turns out both app and mon had /etc/default/profile written with some local locale settings (LANG=nb_NO.UTF-8). The unprivileged user does not use these settings, but when the ansible script escalates privileges, env is set with this locale. This, in turn, makes the script in `ansible_base/installed_files/ansible_base/tor-hidden-services/tasks/install-tor.yml` fail, because it explicitly tests for `installed`.

I am not sure why the default profile was written this way, or why it changed between versions of Ubuntu. I do however believe I had to set the keyboard langauge during initial setup (due to an automated password with characters I had no idea how to type on an English keyboard with a physical Norwegian layout). I know I could possibly have changed the keyboard layout without changing the locale, but the docs did not instruct me to not change it. Anyway, I don't think it's good to require users to keep the locale at English, and that locale-dependent scripts should be fixed.

A possible replacement for manually parsing apt-cache is apt-list, which has a slightly less human output.

### LINKED ISSUE #4242: [docs] Document how to run install playbooks when server locale is non-English
Complements #4219. `securedrop-admin install` fails if the servers are set up with a non-English locale. This is due to direct English language tests against strings like "installed". As upgrades to Xenial are now underway, let's make sure we add workaround documentation to the upgrade-in-place docs ASAP. We can remove it once the fix is for #4219 in & 0.12.1 is out the door.

Note that the [server setup docs](https://docs.securedrop.org/en/release-0.12.0/servers.html) we ask admins to follow explicitly say: "Follow the steps to select your language, country and keyboard settings." We must therefore assume that some other SecureDrop instances may use a non-English locale on the servers (beyond the one that originally reported this problem).

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #4242

Instructions added to upgrade-in-place docs, describing how to check and set server locales before running `./securedrop-admin install`

## Testing
- Docs-only PR - check for clarity and accuracy
- (optional) follow upgrade-in-place instructions on a system originally set up with a non-English language/locale and verify that the post-upgrade install run completes.

### DIFF
--- a/docs/upgrade/xenial_upgrade_in_place.rst
+++ b/docs/upgrade/xenial_upgrade_in_place.rst
@@ -246,6 +246,24 @@ stepping through the configuration using:
 
   ./securedrop-admin sdconfig
 
+Next, you may need to update your servers' default locale settings. This step is
+required if your servers were installed with language and locale settings that
+are incompatible with the SecureDrop installer's Ansible playbook. You can check
+your settings as follows:
+
+.. code:: sh
+ 
+  ssh app sudo locale | grep LANG=
+  ssh mon sudo locale | grep LANG=
+
+If the ``LANG`` variable is set to anything other than a blank value or an 
+English variant, you must change it using the following commands:
+
+.. code:: sh
+ 
+  ssh app sudo update-locale --reset LANG=  
+  ssh mon sudo update-locale --reset LANG=  
+
 Finally, install the Ubuntu 16.04 version of the server application code and
 configuration:
 
--- a/docs/upgrade/xenial_upgrade_in_place.rst
+++ b/docs/upgrade/xenial_upgrade_in_place.rst
@@ -257,12 +257,13 @@ your settings as follows:
   ssh mon sudo locale | grep LANG=
 
 If the ``LANG`` variable is set to anything other than a blank value or an 
-English variant, you must change it using the following commands:
+English variant, you must change your locale settings using the following 
+commands:
 
 .. code:: sh
  
-  ssh app sudo update-locale --reset LANG=  
-  ssh mon sudo update-locale --reset LANG=  
+  ssh app sudo update-locale --reset LC_ALL=C 
+  ssh mon sudo update-locale --reset LC_ALL=C
 
 Finally, install the Ubuntu 16.04 version of the server application code and
 configuration:


eloquence:
I'm wondering if we should recommend changing the locale back after they've run the playbook. This is likely to get them into trouble until #4219 is fixed, but it's possible that some admins legitimately rely on the locale change to effectively administer their servers.

zenmonkeykstop:
@eloquence I added language around that and then thought better of it - they'd have to repeat those steps each time they ran `./securedrop-admin install`, so consistence would demand updating docs everywhere in that case. There aren't many admin tasks on the servers that aren't Ansible-mediated, and for those that are, we provide instructions that assume an English locale, so I'd be more inclined to leave as-is and fix the playbooks.

eloquence:
In that case, should we update the install docs and recommend choosing English there, explaining that the documentation assumes an English-language server configuration? (Can be done in a separate issue/PR.)

zenmonkeykstop:
It won't be necessary when the Ansible playbook is made language-independent,but yeah, if that fix isn't imminent then we should update the install instructions IMO.

eloquence:
What do you think about this approach:

- If a fix for #4219 gets in for 0.12.1, we revert this docs change on 0.12.1 and don't make any additional docs changes. 

- If a fix for #4219 doesn't get in, we keep this docs change and also update the docs to recommend English during the install.

I mainly want to make sure that we don't send mixed messages about whether or not it's okay to change the locale. :)

rmol:
I'm hoping that the fix for #4219 will make it into 0.12.1, but if it doesn't, I have a couple of changes to request after learning more about [the locale environment variables](https://www.gnu.org/software/gettext/manual/html_node/Locale-Environment-Variables.html). The command run on the servers should be `sudo update-locale --reset LC_ALL=C`, as `LC_ALL` has a higher priority than `LANG`, and specifying the `C` locale is safer than leaving the variable empty, which would cause it to be ignored. If a system is setting locale variables somewhere other than `/etc/default/locale`, these subtleties might bite us.

zenmonkeykstop:
@rmol did you have a chance to test this as part of an upgrade-in-place procedure

rmol:
I have, and `LC_ALL=C` worked. 

The process suggested another change, which should probably be added to the "Install SecureDrop" page: on a fresh install, `ssh app sudo update-locale` fails because sudo can't prompt for a password. It needs to be `ssh -t ...`.

zenmonkeykstop:
OK, will update PR accordingly, thanks!

zenmonkeykstop:
Closing this, as the docs change won't get in in time to make a significant difference before the code fix goes live.

PR #4252, opened 2019-03-08
Control locale during Ansible runs

### LINKED ISSUE #4219: Ansible scripts depend on English locale

rmol:
## Status

Ready for review

## Description of Changes

To avoid non-English system locales from breaking tasks that parse command output expecting English, set the environment variable LC_ALL=C in each play of securedrop-prod.yml.

Fixes #4219.

## Testing

Install app and mon with a non-English locale. Verify that running `securedrop-admin install` completes without errors. If it breaks on tasks in `install_tor.yml`, or anywhere else, this fix has not worked.

- [X] Linting and tests (`make -C admin test`) pass in the admin development container

### DIFF
--- a/install_files/ansible-base/securedrop-prod.yml
+++ b/install_files/ansible-base/securedrop-prod.yml
@@ -2,6 +2,8 @@
 ---
 - name: Ensure validation is run before prod install
   hosts: localhost
+  environment:
+    LC_ALL: C
   max_fail_percentage: 0
   any_errors_fatal: yes
   connection: local
@@ -10,6 +12,8 @@
 
 - name: Prepare servers for installation
   hosts: securedrop
+  environment:
+    LC_ALL: C
   gather_facts: no
   max_fail_percentage: 0
   any_errors_fatal: yes
@@ -19,6 +23,8 @@
 
 - name: Add FPF apt repository and install base packages.
   hosts: securedrop
+  environment:
+    LC_ALL: C
   max_fail_percentage: 0
   any_errors_fatal: yes
   pre_tasks:
@@ -49,6 +55,8 @@
 
 - name: Configure OSSEC.
   hosts: securedrop
+  environment:
+    LC_ALL: C
   max_fail_percentage: 0
   any_errors_fatal: yes
   roles:
@@ -58,6 +66,8 @@
 
 - name: Configure mailing utilities.
   hosts: securedrop_monitor_server
+  environment:
+    LC_ALL: C
   max_fail_percentage: 0
   any_errors_fatal: yes
   roles:
@@ -67,6 +77,8 @@
 
 - name: Configure SecureDrop Application Server.
   hosts: securedrop_application_server
+  environment:
+    LC_ALL: C
   max_fail_percentage: 0
   any_errors_fatal: yes
   roles:
@@ -81,6 +93,8 @@
   # connection. After that point the admin will to proxy traffic over tor.
 - name: Lock down firewall configuration for Application and Monitor Servers.
   hosts: securedrop
+  environment:
+    LC_ALL: C
   max_fail_percentage: 0
   any_errors_fatal: yes
   roles:
@@ -89,6 +103,8 @@
 
 - name: Reboot Application and Monitor Servers.
   hosts: securedrop
+  environment:
+    LC_ALL: C
   max_fail_percentage: 0
   any_errors_fatal: yes
   vars:
--- a/install_files/ansible-base/securedrop-backup.yml
+++ b/install_files/ansible-base/securedrop-backup.yml
@@ -3,6 +3,8 @@
   hosts: securedrop_application_server
   max_fail_percentage: 0
   any_errors_fatal: yes
+  environment:
+    LC_ALL: C
   roles:
     - role: backup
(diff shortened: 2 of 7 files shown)


emkll:
I will be testing this will the following plan
- [ ] Set up Prod VMs with french locale (setting the contents of /etc/default/locale [0]) and running `sudo locale-gen fr_FR && sudo locale-gen fr_FR.UTF-8 && sudo update-locale`
sudo locale-gen ru_RU.UTF-8
- [ ] Reboot and ensure `locale` returns [0], that various tooling is in French
- [ ] Install on this branch from Tails, and ensure the testing completes successfully




[0] Contents of /etc/default/locale: 
```
LANG=fr_FR.UTF-8
LANGUAGE="fr_FR.UTF-8"
LC_CTYPE="fr_FR.UTF-8"
LC_NUMERIC="fr_FR.UTF-8"
LC_TIME="fr_FR.UTF-8"
LC_COLLATE="fr_FR.UTF-8"
LC_MONETARY="fr_FR.UTF-8"
LC_MESSAGES="fr_FR.UTF-8"
LC_PAPER="fr_FR.UTF-8"
LC_NAME="fr_FR.UTF-8"
LC_ADDRESS="fr_FR.UTF-8"
LC_TELEPHONE="fr_FR.UTF-8"
LC_MEASUREMENT="fr_FR.UTF-8"
LC_IDENTIFICATION="fr_FR.UTF-8"
LC_ALL="fr_FR.UTF-8"
```

install_files/ansible-base/securedrop-prod.yml
```diff
@@ -2,6 +2,8 @@
 ---
```
emkll: Would it make sense to also apply these changes to `securedrop-staging.yml`?

install_files/ansible-base/securedrop-prod.yml
```diff
@@ -2,6 +2,8 @@
...
 - name: Ensure validation is run before prod install
   hosts: localhost
+  environment:
+    LC_ALL: C
```
emkll: What do you think about validating that this variable is set (via `assert` or otherwise) at the start of each play?

install_files/ansible-base/securedrop-prod.yml
rmol: Sure.

install_files/ansible-base/securedrop-prod.yml
conorsch: Rather than an assert statement, we can add a config test (to be run in CI) that confirms the environment is set as expected _on every play_. See here for an example: https://github.com/freedomofpress/securedrop/blob/10a2eeedc3c22fd39e6a7eda4ca8f4bca1ca6024/molecule/ansible-config/tests/test_max_fail_percentage.py

conorsch: @rmol has already added the corresponding config test here, which will help guard against regressions in the future: if new plays are added _without_ the explicit locale declaration, CI will report an error.

install_files/ansible-base/securedrop-prod.yml
conorsch: Summarizing out of band discussion, we should _also_ add the locale blocks to the other "prod" SD playbooks, such as `-logs` and `-backup`. Technically only `-logs` uses `register`, which is how we first encountered the error we're aiming to resolve here, but adding coverage to the other playbooks will also let us broaden the config test and reuse the existing "check all prod SD playbooks" logic.

I'll tack on a commit to implement this before final review.

conorsch:
Tacked on a commit porting the locale overrides to _all_ production playbooks, with corresponding config test update.

kushaldas:
@emkll I tried `sudo locale-gen fr_FR && sudo locale-gen fr_FR.UTF-8 && sudo update-locale` many times, but, could not set the locale properly. What am I missing?

rmol:
@kushaldas I think you need to pass to `update-locale` the environment variables you want active in `/etc/default/locale`, e.g. `sudo update-locale LANG=fr_FR.UTF8`. I'd add the `--reset` flag as well, which will comment out any variables present in the file that you don't pass.

PR #4257, opened 2019-03-12
replaced drag-and-drop backup step with rsync

### LINKED ISSUE #4256: Workstation backup procedure throws errors if gpg-agent socket files are present
## Description

The documentation for backing up the SecureDrop Workstation and SVS USBs recommends using Nautilus with escalated privileges to copy the source USBs' persistent volume contents to an encrypted backup device. However, Nautilus can't copy socket files. If the source USB was not shut down correctly, there may be socket files left behind by gpg-agent in the PV's `gnupg` directory, that cause Nautilus to throw errors, interrupting the backup process.

This issue will likely only affect older USBs, as the socket file location was changed in gpg 2.1.13 to be /run/user/$UID instead of ~/.gnupg, if available. (Current Tails gpg version is 2.1.18.)

## Steps to Reproduce

- Create a Tails USB with a persistent volume, and start it with the PV enabled and an admin password set
- Set up a GPG key on the Tails USB
- copy the socket files from `/run/user/1000/gnupg` to `/~/.gnupg/`
- shut down the Tails USB, and follow the documentation in the 0.12.0 SecureDrop release to back it up. 

## Expected Behavior
Backup completes without errors.

## Actual Behavior
Nautilus displays an error when trying to copy socket files from `/media/amnesia/TailsData/gnupg`

## Comments

It probably makes sense to use a CLI tool like rsync instead of Nautilus as part of the backup process, and rsync has a  `--no-specials`  flag specifically to skip sockets. The socket files shouldn't be backed up in the first place, as they're created by `gpg-agent` on startup.

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #4256 

Docs-only update: replaces root Nautilus step in workstation backup instructions with rsync command that skips special files that shouldn't be backed up.

## Testing

- Docs-only PR - review for clarity and accuracy.
- Follow workstation backup procedure and verify that it completes successfully, and that it's possible to recreate a workstation USB from the backup by copying the contents of the backup into the persistent volume of a new Tails USB.

### DIFF
--- a/docs/backup_workstations.rst
+++ b/docs/backup_workstations.rst
@@ -94,43 +94,35 @@ mounted and ready to access.
 
 |Backup and TailsData Mounted|
 
-Open a Nautilus window with administrative privileges by first going to
+Open a terminal with administrative privileges by going to
 **Applications** ▸ **System Tools** ▸ **Root Terminal**.
 
 |Root Terminal|
 
-Type ``nautilus`` at the terminal prompt and hit enter. You'll need to type in
-your administrative passphrase you set at the Tails login screen.
-
-|Start Nautilus|
-
-.. note:: When you run ``nautilus``, you may run into an error where Nautilus
- complains that it can't create a required folder. If that happens, just
- click OK and continue normally.
-
+.. note:: 
  If you can't find the "Root Terminal" window, it might be because an
  administrator passphrase wasn't set when you logged in to Tails. If
  that's the case, you'll need to restart Tails and set one at the login screen.
 
-.. warning:: Make sure you use keep the `Root Terminal` window open while
- you perform the backups. Otherwise, the `Nautilus` window will close.
+Next, create a directory on the Backup USB for the device to be backed up - the
+command below creates a directory named ``admin-backup``:
 
-Make sure you create a directory on the backup drive to store the data from the
-drive you are backing up:
+.. code:: sh
+ 
+  mkdir /media/amnesia/Backup/admin-backup
 
-|Make Folders for All Drives|
+Then, copy the contents of the device's persistent volume to the directory using 
+``rsync``:
 
-Copy over everything in the TailsData partition to the relevant folder on the
-Backup drive. You can simply drag to select all the files and then copy and
-paste them to the relevant folder on the Backup drive.
+.. code:: sh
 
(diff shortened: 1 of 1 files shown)


docs/backup_workstations.rst
```diff
@@ -94,43 +94,35 @@ mounted and ready to access.
...
-
+.. note:: 
  If you can't find the "Root Terminal" window, it might be because an
  administrator passphrase wasn't set when you logged in to Tails. If
  that's the case, you'll need to restart Tails and set one at the login screen.
 
-.. warning:: Make sure you use keep the `Root Terminal` window open while
- you perform the backups. Otherwise, the `Nautilus` window will close.
+Next, create a directory on the Backup USB for the device to be backed up - the
+command below creates a directory named ``admin-backup``:
```
heartsucker: Possible nit: does Tails name all devices `Media` by default, or should we maybe call this `MyUsb` for extra clarity?

conorsch: Tails doesn't name all devices `Media` by default, but _will_ mount all USB drives inside `/media/` by default. As the docs stand, the path `/media/amnesia/Backup` matches the earlier instruction to name the drive "Backup", including screenshots indicating the same name.

conorsch:
Followed the docs from top to bottom to perform a backup of an Admin Workstation. The machine I used only had two (2) USB ports, so I plugged in a USB hub, as recommended. It's out of scope for this PR, but let's consider revising these docs to create the backup drive _as a Tails drive_—then we only need two USB ports, and we can copy via rsync or otherwise into `~/Persistent/` on the backup drive. 

As far as validating the intended fix, the new command works as intended. Two suggestions on the rsync command flags:

1. No need to use `-z` / `--compress`; since this is a local transfer, not over a network, the compression is more overhead and actually slowed down transfers slightly in my unscientific testing. 
2. The `-vh` flags make output _very_ verbose: one line per file. Instead, we could use `--info=progress2`, which provides a nice progress indicator, and the entire output from the command fits on a single terminal screen.

So, condensing those recommendations into a new rsync command:

```
rsync -a --info=progress2 --no-specials --no-devices /media/amnesia/TailsData/ /media/amnesia/Backup/admin-backup
```

Then this is fine to merge!

zenmonkeykstop:
Command looks good, updating.

PR #4263, opened 2019-03-14
use vim in update_version.sh

heartsucker:
## Status

Ready for review

## Description of Changes

Use `vim` instead of `vi` in `update_version.sh` because `vi` is so annoying to use in comparison.

### DIFF
--- a/securedrop/dockerfiles/trusty/Dockerfile
+++ b/securedrop/dockerfiles/trusty/Dockerfile
@@ -6,7 +6,7 @@ ARG USER_ID
 ENV USER_ID ${USER_ID:-0}
 
 RUN apt-get update && \
-    apt-get install -y devscripts \
+    apt-get install -y devscripts vim \
                        python-pip libpython2.7-dev libssl-dev secure-delete \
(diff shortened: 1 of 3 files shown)


heartsucker:
If any devs are running that, it's probably worth maintaining. I'll add it and some comments about why then push again.

heartsucker:
I pushed a fix that takes your diff and collapses it into a single layer. Take a peeksy again?

heartsucker:
This test is consistently failing, but it looks like a flake we've seen before. https://circleci.com/gh/freedomofpress/securedrop/24581?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

heartsucker:
I can't reproduce that bug on my laptop. Gonna just keep hitting "rebuild" until something happens. :neutral_face:

heartsucker:
JK that last failure was def an HKP issue and not the QPainter one we've seen a bunch of times. :sob:

heartsucker:
CI crashed because of 20 minutes with no output.

PR #4288, opened 2019-03-21
Extract VNC changes from Tor Browser functional test branch

### LINKED ISSUE #4287: Port support for VNC observability from `tbb-0.9.0` to existing functional tests in `develop`
In the `tbb-0.9.0` branch for functional testing via the Tor browser we added support for observing running tests via VNC, which is extremely helpful for debugging. See #3817. 

As part of incrementally merging the key changes in this branch, let's port this specific functionality to work with the existing functional tests in `develop`.

rmol:
## Status

Ready for review

## Description of Changes

These changes were cherry-picked from the tbb-0.12.0 branch. The VNC server used in the functional test container is switched to tightvncserver, and scripts are changed to ensure that it can be reached on both Mac and Linux developer machines.

Fixes #4287.

## Testing

Run the functional tests with `securedrop/bin/dev-shell bin/run-test -v tests/functional`. In another local shell, run `make -C securedrop func-vnc` to verify that you can connect to the test container with VNC and watch the functional tests run.

### DIFF
--- a/devops/scripts/vnc-docker-connect.sh
+++ b/devops/scripts/vnc-docker-connect.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+# shellcheck disable=SC2086
+#
+#
+# Connect to a docker test instance's VNC session
+
+set -e
+
+# Bomb out if container not running
+docker inspect securedrop-dev >/dev/null 2>&1 || (echo "ERROR: SD container not running."; exit 1)
+
+
+# Find our securedrop docker ip
+SD_DOCKER_IP="$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' securedrop-dev)"
+
+# Quit if the VNC port not found
+nc -w5 -z "$SD_DOCKER_IP" 5901 || (echo "ERROR: VNC server not found"; exit 1)
+
+# Are we running on gnome desktop?
+if [ "$(echo \"$XDG_DATA_DIRS\" | sed 's/.*\(gnome\).*/\1/')" == "gnome" ]; then
+    remote-viewer "vnc://${SD_DOCKER_IP}:5901" 2>/dev/null &
+
+# Maybe we are running macOS
+elif [ "$(uname -s)" == "Darwin" ]; then
+    open "vnc://${SD_DOCKER_IP}:5901" &
+
+else
+    echo "Not sure what the VNC terminal cli arguments are for your OS."
+    echo "Please add to $0 and make a pull-request"
+fi
--- a/devops/scripts/vnc-docker-connect.sh
+++ b/devops/scripts/vnc-docker-connect.sh
@@ -9,6 +9,11 @@ set -e
 # Bomb out if container not running
 docker inspect securedrop-dev >/dev/null 2>&1 || (echo "ERROR: SD container not running."; exit 1)
 
+# Maybe we are running macOS
+if [ "$(uname -s)" == "Darwin" ]; then
+    open "vnc://${USER}:freedom@127.0.0.1:5901" &
+    exit 0
+fi
 
 # Find our securedrop docker ip
 SD_DOCKER_IP="$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' securedrop-dev)"
@@ -18,11 +23,7 @@ nc -w5 -z "$SD_DOCKER_IP" 5901 || (echo "ERROR: VNC server not found"; exit 1)
 
 # Are we running on gnome desktop?
 if [ "$(echo \"$XDG_DATA_DIRS\" | sed 's/.*\(gnome\).*/\1/')" == "gnome" ]; then
-    remote-viewer "vnc://${SD_DOCKER_IP}:5901" 2>/dev/null &
-
-# Maybe we are running macOS
-elif [ "$(uname -s)" == "Darwin" ]; then
-    open "vnc://${SD_DOCKER_IP}:5901" &
+    remote-viewer "vnc://${USER}:freedom@${SD_DOCKER_IP}:5901" 2>/dev/null &
 
 else
     echo "Not sure what the VNC terminal cli arguments are for your OS."
--- a/devops/scripts/vnc-docker-connect.sh
+++ b/devops/scripts/vnc-docker-connect.sh
@@ -21,14 +21,17 @@ SD_DOCKER_IP="$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddre
 # Quit if the VNC port not found
 nc -w5 -z "$SD_DOCKER_IP" 5901 || (echo "ERROR: VNC server not found"; exit 1)
 
-# Are we running on gnome desktop?
-if [ "$(echo \"$XDG_DATA_DIRS\" | sed 's/.*\(gnome\).*/\1/')" = "gnome" ]; then
-    echo -e "[virt-viewer]\ntype=vnc\nhost=${SD_DOCKER_IP}\nport=5901\npassword=freedom" > /tmp/func-vnc.ini
-    remote-viewer /tmp/func-vnc.ini 2>/dev/null &
-elif [ "$(echo \"$XDG_DATA_DIRS\" | sed 's/.*\(ubuntu\).*/\1/')" = "ubuntu" ]; then
-    echo -e "[virt-viewer]\ntype=vnc\nhost=${SD_DOCKER_IP}\nport=5901\npassword=freedom" > /tmp/func-vnc.ini
-    remote-viewer /tmp/func-vnc.ini 2>/dev/null &
-else
-    echo "Not sure what the VNC terminal cli arguments are for your OS."
-    echo "Please add to $0 and make a pull-request"
+if [ ! $(which remote-viewer) ]
+then
+    printf "\nError: We use the remote-viewer utility to reach Docker via VNC,\n"
+    printf "and it is not installed. On Debian or Ubuntu, install it with\n"
+    printf "'sudo apt install virt-viewer', or if you use another VNC client,\n"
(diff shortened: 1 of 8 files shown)


heartsucker:
Works on Debian 9. Letting others confirm on other OSes.

rmol:
@kushaldas Yes, `vncpasswd` can seg fault. I added a retry loop around it in [securedrop/bin/dev-deps](https://github.com/freedomofpress/securedrop/pull/4288/files#diff-54a1021fdb59e21d71ed44d5433a8a8a), but maybe we can take another look at the need for the passwd -- @zenmonkeykstop, what prompted [that](https://github.com/freedomofpress/securedrop/commit/b53b657caf3a60ef8df03adfaf0417475a8b151c)? Or, maybe there's another VNC server option -- again, I don't know the history for the [switch from x11vnc](https://github.com/freedomofpress/securedrop/commit/af16906b69224df356c4bbc1465169b286c4fb89).

kushaldas:
Only segfault is different, and a `buffer overflow` is scary :)

rmol:
In this case, I don't think it's that scary, as `vncpasswd` is running as an unprivileged user, invoked only by the `run-test` script. But if we want to eliminate the problem, we could:

- Revert to x11vnc (Maybe @msheiny could weigh in on the motivation for the switch?).
- Omit the authentication (though apparently it was necessary for Mac users, @zenmonkeykstop?).
- Create the password file another way. There's at least one Python implementation of the obfuscation used, which we could incorporate.

Open to suggestions. I'd probably pick the first option.

rmol:
I've reverted to x11vnc from tightvncserver. The reasons for the switch have been lost to time, the tests work and are visible with x11vnc, and it does support the passwords required on Macs, so I'd rather switch back than have to maintain workarounds for tightvncserver bugs.

Works for me, @kushaldas and @heartsucker, please check again when you have time, and I'll find a Mac user to test there as well.

zenmonkeykstop:
Failing for me on Qubes in a debian-9-based standaloneVM - docker output includes the following:

```
(securedrop) user@sd-dev:~/securedrop/securedrop$ ./bin/dev-shell  bin/run-test -v tests/functional
Run with DOCKER_BUILD_VERBOSE=true for more information
Docker image build in progress  done !
stored passwd in file: /tmp/vncpasswd
03/04/2019 14:27:00 passing arg to libvncserver: -rfbauth
03/04/2019 14:27:00 passing arg to libvncserver: /tmp/vncpasswd
03/04/2019 14:27:00 passing arg to libvncserver: -rfbport
03/04/2019 14:27:00 passing arg to libvncserver: 5901
03/04/2019 14:27:00 x11vnc version: 0.9.13 lastmod: 2011-08-10  pid: 14
03/04/2019 14:27:00 XOpenDisplay(":1") failed.
03/04/2019 14:27:00 Trying again with XAUTHLOCALHOSTNAME=localhost ...

03/04/2019 14:27:00 ***************************************
03/04/2019 14:27:00 *** XOpenDisplay failed (:1)

*** x11vnc was unable to open the X DISPLAY: ":1", it cannot continue.
*** There may be "Xlib:" error messages above with details about the failure.

Some tips and guidelines:

** An X server (the one you wish to view) must be running before x11vnc is
   started: x11vnc does not start the X server.  (however, see the -create
   option if that is what you really want).

** You must use -display <disp>, -OR- set and export your $DISPLAY
   environment variable to refer to the display of the desired X server.
 - Usually the display is simply ":0" (in fact x11vnc uses this if you forget
   to specify it), but in some multi-user situations it could be ":1", ":2",
   or even ":137".  Ask your administrator or a guru if you are having
   difficulty determining what your X DISPLAY is.

** Next, you need to have sufficient permissions (Xauthority) 
   to connect to the X DISPLAY.   Here are some Tips:

 - Often, you just need to run x11vnc as the user logged into the X session.
   So make sure to be that user when you type x11vnc.
 - Being root is usually not enough because the incorrect MIT-MAGIC-COOKIE
   file may be accessed.  The cookie file contains the secret key that
   allows x11vnc to connect to the desired X DISPLAY.
 - You can explicitly indicate which MIT-MAGIC-COOKIE file should be used
   by the -auth option, e.g.:
       x11vnc -auth /home/someuser/.Xauthority -display :0
       x11vnc -auth /tmp/.gdmzndVlR -display :0
   you must have read permission for the auth file.
   See also '-auth guess' and '-findauth' discussed below.

** If NO ONE is logged into an X session yet, but there is a greeter login
   program like "gdm", "kdm", "xdm", or "dtlogin" running, you will need
   to find and use the raw display manager MIT-MAGIC-COOKIE file.
   Some examples for various display managers:

     gdm:     -auth /var/gdm/:0.Xauth
              -auth /var/lib/gdm/:0.Xauth
     kdm:     -auth /var/lib/kdm/A:0-crWk72
              -auth /var/run/xauth/A:0-crWk72
     xdm:     -auth /var/lib/xdm/authdir/authfiles/A:0-XQvaJk
     dtlogin: -auth /var/dt/A:0-UgaaXa

   Sometimes the command "ps wwwwaux | grep auth" can reveal the file location.

   Starting with x11vnc 0.9.9 you can have it try to guess by using:

              -auth guess

   (see also the x11vnc -findauth option.)

   Only root will have read permission for the file, and so x11vnc must be run
   as root (or copy it).  The random characters in the filenames will of course
   change and the directory the cookie file resides in is system dependent.

See also: http://www.karlrunge.com/x11vnc/faq.html
...
```

rmol:
@zenmonkeykstop Yeah, I started running into this in the big all-tbb merge too. I think the problem is that Xvfb sometimes hasn't fully started by the time we start x11vnc. Reordering things in run-test seems to have fixed it, but I'm planning on doing something more reliable.

zenmonkeykstop:
@rmol fair enough, like maybe a test for xvfb's availability with a couple of sleep/retries?

FWIW, same issue under OS X

zenmonkeykstop:
Added a `sleep 30` between the xvfb and x11vnc startups, which cleared the above error. So the x11vnc bit itself is fine, it just needs the aforementioned reliability fix.

rmol:
I've added a wait for Xvfb to be fully available.

zenmonkeykstop:
Confirmed that VNC working with no issues on both Qubes(debian-9) and OS X with this fix. LGTM!

eloquence:
Real lint failure for a change:

```
In ./securedrop/bin/dev-deps line 9:
    for i in {1..10}
    ^-- SC2034: i appears unused. Verify it or export it.
```

zenmonkeykstop:
VNC still working fine on Qubes(debian-9) and OS X with the shellcheck fix above. :+1: from me.

devops/scripts/vnc-docker-connect.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Find our securedrop docker ip
+SD_DOCKER_IP="$(docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}' securedrop-dev)"
+
+# Quit if the VNC port not found
+nc -w5 -z "$SD_DOCKER_IP" 5901 || (echo "ERROR: VNC server not found"; exit 1)
+
+if [ ! "$(which remote-viewer)" ]
+then
+    printf "\nError: We use the remote-viewer utility to reach Docker via VNC,\n"
```
emkll: this was a very helpful message :)

securedrop/bin/dev-shell
```diff
@@ -40,12 +40,14 @@ function docker_run() {
 
     find . \( -name '*.pyc' -o -name __pycache__ \) -delete
     docker run \
-	   --rm \
+       -p 127.0.0.1:5901:5901 \
```
emkll: 5901 is only exposed only to localhost :+1:

securedrop/dockerfiles/trusty/Dockerfile
```diff
@@ -1,5 +1,5 @@
-# ubuntu:14.04 as of 2019-01-22
-FROM ubuntu@sha256:cac55e5d97fad634d954d00a5c2a56d80576a08dcc01036011f26b88263f1578
+# ubuntu 14.04 image from 2019-03-12
+FROM ubuntu@sha256:6612de24437f6f01d6a2988ed9a36b3603df06e8d2c0493678f3ee696bc4bb2d
```
emkll: :+1:

securedrop/dockerfiles/xenial/Dockerfile
```diff
@@ -1,5 +1,5 @@
-# ubuntu 16.04 image - 2019-01-22
-FROM ubuntu@sha256:e4a134999bea4abb4a27bc437e6118fdddfb172e1b9d683129b74d254af51675
+# ubuntu 16.04 image from 2019-03-12
+FROM ubuntu@sha256:58d0da8bc2f434983c6ca4713b08be00ff5586eb5cdff47bcde4b2e88fd40f88
```
emkll: :+1:

devops/scripts/vnc-docker-connect.sh
```diff
@@ -0,0 +1,37 @@
...
+    printf "and it is not installed. On Debian or Ubuntu, install it with\n"
+    printf "'sudo apt install virt-viewer', or if you use another VNC client,\n"
+    printf "consider adding it to this script:\n"
+    printf "\n%s\n\n" "$(realpath $0)"
+    printf "and submitting a pull request.\n\n"
+    exit 1
+fi
+
+rv_config="${TMPDIR:-/tmp}/func-vnc.ini"
+echo -e "[virt-viewer]\ntype=vnc\nhost=${SD_DOCKER_IP}\nport=5901\npassword=freedom" > "${rv_config}"
```
emkll: Given VNC is exposed only to localhost and it's a dev container/test data, I think this is fine. We should follow up with a better way (maybe automatically generate a password?) in the future.

Issue #4024, opened 2019-01-09
Add latest version of e1000e module, built for grsec kernel


zenmonkeykstop:
## Description

Intel 7-series NUCs use the e1000e kernel driver for built-in NIC support under Linux. The version that ships with Ubuntu 14.04.5, and the version that's currently installed with SD's grsec kernels, is out-of-date and does not support the Ethernet chipset on new NUCs. Version 3.4.2.1, available as a source tarball on the Intel support site, does. The module shipped with the grsec kernel should be updated to this version if possible. This will allow for a wider range of NUC hardware to be added to the SecureDrop HCL.

More info on the build process for this driver is available at https://downloadcenter.intel.com/download/15817 - The build process has been confirmed to work for 14.04.5's most recent 4.4.0 kernel with recent updates, but has not been tested for the SD grsec kernel as it requires a linux-headers package.

## User Stories
As an organization investigating SecureDrop, I want to have a choice of NUC hardware that's up-to-date and commercially available.

eloquence:
As part of the 1/9-/1/23 sprint, we've committed to a 4 person hour time-boxed investigation (likely led by @emkll and @zenmonkeykstop) to determine (and, if doable within the timebox, implement) the best way to ship/enable the driver with the next SecureDrop kernel release.

redshiftzero:
This won't get in for 0.12.0, bumping this off the milestone into a potential 0.12.1 point release milestone

ageis:
I used to have e1000e cards for a few years before switching to igb and have built the driver many times (often complains about fPIC). FYI, the module that ships in the Linux kernel is typically a newer version than the source one can obtain from Intel.

zenmonkeykstop:
Did some more tests. *TL;DR:* the e1000e module provided with the current 4.4.167-grsec kernel doesn't have an alias defined for the NIC in the NUC7i5BNH, despite being the same version as the e1000e module provided with the 4.4.0131-generic kernel in Ubuntu 16.04.6, which does. This is why it doesn't detect the NIC.

## Device info
All testing done against an Intel NUC7i5BNH, with the latest available BIOS version (0072). NIC device info from `lspci` is as follows:
```
00:1f.6 Ethernet controller [0200]: Intel Corporation Ethernet Connection (4) I219-V [8086:15d8] (rev 21)
	DeviceName: Onboard - Ethernet
	Subsystem: Intel Corporation Ethernet Connection (4) I219-V [8086:2068]
[...]

PR #4308, opened 2019-03-29
Bump SecureDrop kernel to 4.4.177

emkll:
## Status

Ready for review

## Description of Changes

Towards #4024 
Simply bumps strings for grsec kernel, the kernel work was done by @zenmonkeykstop in https://github.com/freedomofpress/ansible-role-grsecurity-build/pull/46

## Test Plan

First:
- [x] Metapackage and kernel images uploaded to apt-test (Thanks @rmol !)

To validate the functionality of these kernels:

- [ ] Clean install using apt-test.freedom.press is successful
- [ ] intel e1000e NIC is functional using the 4.4.177-grsec kernel provided
- [ ] CI passes

## Deployment

New and existing installs will be upgraded via apt

### DIFF
--- a/install_files/ansible-base/group_vars/all/securedrop
+++ b/install_files/ansible-base/group_vars/all/securedrop
@@ -52,5 +52,5 @@ securedrop_cond_reboot_file: /tmp/sd-reboot-now
 
 # If you bump this, also remember to bump in molecule/builder/tests/vars.yml
 securedrop_pkg_grsec:
-  ver: "4.4.167"
-  depends: "linux-image-4.4.162-grsec,linux-firmware-image-4.4.162-grsec,linux-image-4.4.167-grsec,linux-firmware-image-4.4.167-grsec"
+  ver: "4.4.177"
+  depends: "linux-image-4.4.167-grsec,linux-firmware-image-4.4.167-grsec,linux-image-4.4.177-grsec,linux-firmware-image-4.4.177-grsec"
--- a/molecule/builder-trusty/tests/vars.yml
+++ b/molecule/builder-trusty/tests/vars.yml
@@ -3,7 +3,7 @@ securedrop_version: "0.13.0~rc1"
 ossec_version: "3.0.0"
 keyring_version: "0.1.2"
 config_version: "0.1.3"
-grsec_version: "4.4.167"
+grsec_version: "4.4.177"
 
 # These values will be interpolated with values populated above
 # via helper functions in the tests.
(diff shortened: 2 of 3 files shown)


emkll:
This should now be ready for review. CI will staging-test-with-rebase will fail on all PRs due to 4.4.177 currently being served by apt-test, and the tests are checking for 4.4.167

rmol:
@zenmonkeykstop @emkll Hate to end the week with bad news, but it's probably just something I've done wrong. I tried a fresh install of SD on Xenial, by changing the apt repo to apt-test in my develop checkout and running `securedrop-admin install`, and the installation timed out while waiting for my NUCs to reboot. 

I logged in on their consoles to find they're both now running the 4.4.177-grsec kernel, but with no Ethernet interfaces configured. Running `modprobe e1000e` does nothing. The interface doesn't show up in `ip link` (there's just the loopback), and the driver isn't attached to the NIC in the output of `lspci`. The PCI ID is the same as above (8086:15d8), rev 21. 

These are NUC 715BNHs. One has the same BIOS revision as above, 72, while the other is at 76. The NICs worked fine for the first part of the installation.

zenmonkeykstop:
Hi @rmol - if the kernel version bumped, I don't think that you did anything wrong. Could you do me a favour and run `modinfo e1000e | grep -i 15d8` to see if there's an alias for the chipset defined in the module that it's running?

rmol:
@zenmonkeykstop There is not.

conorsch:
Summarizing discussion from today's standup: @rmol's surprising result of broken hardware support was caused by a mix-up on the backend, with the `apt-test` repo. We're sorting that out now, having rebuilt with the patches presented in https://github.com/freedomofpress/ansible-role-grsecurity-build/pull/46. CI is passing here, validating that the new kernel versions are serving as expected, via the `apt-test` repo.

We'll proceed with manual QA on hardware prior to the 0.12.2 point release. Approving and merging, given that the changes are presented are behaving well (as evidenced by CI), and also because the new test vars are required for _other_ PRs to pass CI now, given the automatic rebase in CI.

At a later date, let's consider a branch naming pragma with `kernel-`, similar to what we do with `docs-` (to skip the full CI run, which would skip _just_ the kernel version tests.

emkll:
thanks for the review @rmol I have updated apt-test with kernels that should now be e1000e compatible. The previous kernels were built on master instead of the e1000e feature branch, as I was using a disposable VM and forgot to explicitly check out the correct WIP remote branch, my apologies.

PR #4309, opened 2019-04-01
added fn to prevent multiple GUI updater instances

### LINKED ISSUE #3999: On the Admin Workstation,  N SecureDrop updater processes are started if N network interfaces are enabled in Tails
## Description

If a Tails Admin Workstation has more than one network interface, the SecureDrop network manager hook that checks for updates will run for each active NIC. (For example, if a workstation has both Ethernet and Wifi enabled.)

This is confusing to end users and may result in multiple update processes clobbering each other's changes.

## Steps to Reproduce

On an Admin Workstation:
- revert to an earlier SecureDrop version:
```
cd ~/Persistent/securedrop
git checkout 0.10.0
```
- enable multiple network connections (eg Ethernet and Wifi) and wait for their Tor connections to come up

## Expected Behavior
A single instance of the Securedrop Updater is started.

## Actual Behavior
Multiple instances of the SecureDrop Updater are started.

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #3999.

Adds a function based on the one in securedrop-client as per @redshiftzero's [recommendation](https://github.com/freedomofpress/securedrop/issues/3999#issuecomment-448674976), to prevent multiple running instances of the GUI updater.

## Testing

In a Tails Admin or Journalist workstation:

1.  Check out this branch in the `~/Persistent/securedrop` application directory:
```
cd ~/Persistent/securedrop
git checkout 3999-singleton-gui-updater
```
2. Disable existing Tails network connections and re-enable them 1 by 1. (If the Tails workstation only has one network connection, disable and enable it repeatedly, allowing a minute or two for the Tor connection and networkmanager hook to run between each toggle.) The first network connection should spawn a SecureDrop GUI updater window. Second and subsequent runs should pop up an alert dialog saying "SecureDrop Updater is already running." No other updater windows should spawn.
3. In the SecureDrop GUI updater window, complete the update and verify that it finishes successfully.

### DIFF
--- a/journalist_gui/journalist_gui/SecureDropUpdater.py
+++ b/journalist_gui/journalist_gui/SecureDropUpdater.py
@@ -5,12 +5,33 @@ import subprocess
 import os
 import re
 import pexpect
+import socket
+import sys
 
 from journalist_gui import updaterUI, strings, resources_rc  # noqa
 
 
 FLAG_LOCATION = "/home/amnesia/Persistent/.securedrop/securedrop_update.flag"  # noqa
 ESCAPE_POD = re.compile(r'\x1B\[[0-?]*[ -/]*[@-~]')
+
+
+def prevent_second_instance(app: QtWidgets.QApplication, name: str) -> None:  # noqa
+
+    # Null byte triggers abstract namespace
+    IDENTIFIER = '\0' + name
+    ALREADY_BOUND_ERRNO = 98
+
+    app.instance_binding = socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM)
+    try:
+        app.instance_binding.bind(IDENTIFIER)
+    except OSError as e:
+        if e.errno == ALREADY_BOUND_ERRNO:
+            err_dialog = QtWidgets.QMessageBox()
+            err_dialog.setText(name + ' is already running.')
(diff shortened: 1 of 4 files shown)


heartsucker:
If we wanted to add some testing, this is an example. https://github.com/mdrose/securedrop-client/blob/c228dae8487a431e21c6320665cfa51193738b7d/tests/test_app.py Though this is such a similar implementation that I assume @zenmonkeykstop knew about it.

zenmonkeykstop:
@zenmonkeykstop ... may have forgotten to see if there were tests corresponding to @redshiftzero's example code. Added now.

journalist_gui/journalist_gui/SecureDropUpdater.py
```diff
@@ -13,6 +15,25 @@
 ESCAPE_POD = re.compile(r'\x1B\[[0-?]*[ -/]*[@-~]')
 
 
+def prevent_second_instance(app: QtWidgets.QApplication, name: str) -> None:  # noqa
```
emkll: Unlike the client, we don't use type hints in the updater GUI, even though it is just supported by the version of Python here. But I see no harm in having these here

PR #4312, opened 2019-04-01
Reduce setup time per test by cloning GPG homedir

heartsucker:
## Status

Ready for review

## Description of Changes

Create one master GPG dir and clone it. Each GPG home creation takes ~2 seconds on my laptop, so even running a single module's tests can be minutes longer than necessary.

## Testing

Does CI pass?

### DIFF
--- a/securedrop/tests/conftest.py
+++ b/securedrop/tests/conftest.py
@@ -107,8 +107,10 @@ def config(tmpdir):
     cnf.TEMP_DIR = str(tmp)
     cnf.DATABASE_FILE = str(sqlite)
 
+    DEVNULL = io.open(os.devnull, 'w')
     # create the db file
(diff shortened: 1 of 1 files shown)


eloquence:
As discussed today, this requires clearer specification and technical input from the rest of the team, once that's done we can consider it for the next sprint. :)

eloquence:
Thanks @heartsucker and sorry for the late response. We agreed that this is still potentially a worthwhile improvement if it does not cause test flakiness; adding to near-term backlog for consideration in a future sprint.

nabla-c0d3:
A logic similar to this PR (ie. re-using the GPG directory to speed up the tests) was implemented in https://github.com/freedomofpress/securedrop/pull/5683 .

eloquence:
This was resolved via a new, independent PR, #5683.

PR #4313, opened 2019-04-02
[xenial docs] Bump from 0.12.0 to 0.12.1 where version is specified

eloquence:
## Description

During the point release we bumped the version in the prep doc, but not in the detailed upgrade docs.

## Status

Ready for review

## Checklist

### DIFF
--- a/docs/upgrade/xenial_backup_install_restore.rst
+++ b/docs/upgrade/xenial_backup_install_restore.rst
@@ -157,7 +157,7 @@ code is up-to-date and validated. From a terminal, run the following commands:
 
  cd ~/Persistent/securedrop
  git fetch --tags
- git tag -v 0.12.0
+ git tag -v 0.12.1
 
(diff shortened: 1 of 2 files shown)


eloquence:
> It might make sense to keep them in sync with other stable versions used across the docs so that these version strings don't get forgotten again:

I agree; there are a couple places where the Xenial docs need to continue to refer to the 0.12 series, which need to be either rewritten or written so they can be easily exempted from the regex, but that should be doable. Let me make a separate PR for that.

> However it seems like these docs should be removed once the Xenial transition is complete? What do you think?

There will still be some Trusty instances in the wild after April 30, possibly well after. I think we'll want to do a refactor in May where we remove the "upgrade in place" instructions and instruct people to explicitly follow the "reinstall from backup" procedure.

eloquence:
Abandoning in favor of new PR with automation.

PR #4325, opened 2019-04-08
Disable Source Interface for Trusty when Trusty is end-of-life

### LINKED ISSUE #4305: Disable Source Interface on instances running Trusty after April 30
> **NOTE: If you are an administrator and would like to receive help with the upgrade from Trusty to Xenial, please contact us via securedrop@freedom.press ([GPG encrypted](https://securedrop.org/sites/default/files/fpf-email.asc)), or open an issue in our support portal if you are a member.**

As [previously communicated](https://securedrop.org/news/advisory-why-you-must-manually-upgrade-your-securedrop-servers-april-30/), news organizations must upgrade to Ubuntu 16.04 by April 30, as Ubuntu 14.04 reaches EOL on that date. The Journalist Interface of impacted instances has [displayed a warning](https://github.com/freedomofpress/securedrop/issues/4027) since March 4, and extensive communications continue through all available channels. 

In our previous advisory, we have indicated that we will remove Trusty instances from the [SecureDrop directory](https://securedrop.org/directory/) after April 30, and that “we cannot guarantee that your SecureDrop instance will continue to function after that date."

Running an unsupported operating system on the servers presents an unacceptable risk to sources and newsrooms alike. For this reason, after April 30, news organizations running SecureDrop on Trusty should no longer be able to communicate with sources until they upgrade. The Source Interface should show a neutral message similar to the following (final language TBD):

> This SecureDrop instance is currently unavailable, pending a server upgrade. Please check back again later.

All source interface functionality should be disabled. The journalist interface should continue to work as before. We should also plan additional comms through all available channels in early April (tracked internally).

## User Stories

As a source, I want to have confidence that all my interactions via SecureDrop meet the highest standard of information security, so I face minimal risks of deanonymization by an adversary.

As a news organization, I want my SecureDrop instance to be protected against external compromise at all times, so that existing submissions and correspondence are not exposed to adversaries.

## Acceptance Criteria

**Given** that my SecureDrop servers are running Trusty after April 30
**When** I connect to the server’s Source Interface
**Then** I should see a maintenance notice
**And** all other Source Interface functionality should be disabled

**Given** that my SecureDrop servers were running Trusty after April 30, but have since been upgraded to Xenial following [instructions](https://securedrop.org/xenial-upgrade)
**When** I connect to the server’s Source Interface
**Then** the Source Interface should operate normally
**And** no warning message should be displayed

emkll:
## Status

Ready for Review
## Description of Changes

Fixes #4305: Disables the source interface for Trusty instances after Ubuntu 14.04 Trusty's EOL date (2019-04-30).

#### Target branch
- [ ] Does it make sense
#### Copy
- [ ] The message for disabled source interfaces makes sense

#### Test on 0.12.1 (Trusty)
1. `git checkout release/0.12.1`
2. `git cherry-pick -x <commit hash(es) in this PR>`
3. `make build-debs`
4. `molecule converge -s libvirt-staging`
5. Modify values in `/var/www/securedrop/source_app/__init__.py:L111` (change `TRUSTY_DISABLE_DATE` to any past date)
6. Navigate to the the source interface and confirm:
- [ ] Source interface is disabled
- [ ] Any other URL path (except static assets) returns this new page
- [ ] No AppArmor errors in /var/log/syslog

#### Next steps
- [ ] Should we open a follow-up ticket to track further disabling of the Source Interface (Apache or THS level)?

## Deployment

New and existing installs will be upgraded automatically though the securedrop-app-code deb package


securedrop/source_app/decorators.py
```diff
@@ -18,7 +18,7 @@ def ignore_static(f):
     a static resource."""
     @wraps(f)
     def decorated_function(*args, **kwargs):
-        if request.path.startswith('/static'):
```
emkll: Would moving the org-logo route behind /static make more sense here?

heartsucker:
I think we should not merge this into develop and instead merge it directly into the 0.12 release branch. Adding it to develop just means code we will have to maintain until we completely yank all Trusty every from the codebase. I think it is appropriate to deprecate Trusty code in develop and make Trusty specific changes just in the 0.12 branch. Isn't that the point of having multiple release branches like this?

securedrop/source_app/decorators.py
heartsucker: We could make code changes to create a `Blueprint` whose `/static` route is set to be behind `/static/org` and whose content is exactly the single file that is uploaded. Doing this cleanly is actually just a little tricky, and our current code does not easily support this at the moment.

securedrop/source_app/__init__.py
```diff
@@ -106,6 +106,11 @@ def handle_csrf_error(e):
     for module in [main, info, api]:
         app.register_blueprint(module.make_blueprint(config))  # type: ignore
 
+    # Disables the app if the server is running Trusty past its EOL date.
+    app.config['XENIAL_VER'] = "16.04"
```
heartsucker: Instead of adding these as config options, we could still mock them by making them constants in the `disable` module.

securedrop/source_app/disable.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def disable_app(app):
+
+    @app.before_request
+    @ignore_static
+    def disable_ui():
+        if(platform.linux_distribution()[1] != app.config['XENIAL_VER'] and
+                date.today() > app.config['TRUSTY_DISABLE_DATE']):
+            g.locale = i18n.get_locale(app.sdconfig)
```
heartsucker: Are these added here because we haven't yet called the other `before_request` handlers? If this was done to follow my suggestion in the original ticket of running this handler before all other handlers, that was to *prevent* any other code from being hit to minimize code paths on a possibly insecure instance. What we could do is split `setup_g` into multiple smaller handlers that each do one thing like `init_i18n1` and `check_expires` and such. This would make some of the code more reusuable and put less logic into this handler.

securedrop/tests/test_source.py
```diff
@@ -21,6 +24,9 @@
 from utils.instrument import InstrumentedApp
 
 overly_long_codename = 'a' * (Source.MAX_CODENAME_LEN + 1)
+SOURCE_ENDPOINTS = ['main.index', 'main.lookup', 'main.generate', 'main.login',
```
heartsucker: We may want to add other endpoints called `allowed endpoints` for CSS and other static content and check that those endpoints aren't blocked by this feature.

securedrop/source_app/__init__.py
emkll: done

securedrop/source_app/disable.py
emkll: good point, done

securedrop/tests/test_source.py
emkll: good point, done

PR #4327, opened 2019-04-09
Fixes more Python2 and Python3 common codebase

kushaldas:
## Status

Work in progress

## Description of Changes

Fixes a few more failed tests.

## Testing

`make test`

`make test-python3`

## Deployment

2. New installs.


securedrop/crypto_util.py
```diff
@@ -259,6 +260,8 @@ def decrypt(self, secret, ciphertext):
         """
         hashed_codename = self.hash_codename(secret,
                                              salt=self.scrypt_gpg_pepper)
+        if six.PY2:  # Python2
```
heartsucker: This is a stylistic comment only. This duplicates the logical line of code. I think it would make more sense to do:

```python
data = self.gpg.decrypt(ciphertext, passphrase=hashed_codename).data

if not six.PY2: # in python 3 the output is bytes, but we need str
    data = data.decode('utf-8')

return data
```

This makes it easier to spot the python2/3 differences at a quick glance.

securedrop/secure_tempfile.py
```diff
@@ -47,7 +48,12 @@ def __init__(self, store_dir):
         """
         self.last_action = 'init'
         self.create_key()
-        self.tmp_file_id = base64.urlsafe_b64encode(os.urandom(32)).decode('utf-8').strip('=')
+        self.temp_file_id = ""
+        if six.PY2:
```
heartsucker: As in the other commend, we duplicate logic, and may make sense to use a more minimal branching on the python2/3 code.

securedrop/secure_tempfile.py
```diff
@@ -83,7 +89,12 @@ def write(self, data):
             raise AssertionError('You cannot write after reading!')
         self.last_action = 'write'
 
-        if isinstance(data, str):  # noqa
+        # This is the old Python related code
+        if six.PY2:  # noqa
```
heartsucker: This is overly complicated for python2/3 compatibility. The code should be:

```python
if isinstance(data, six.binary_type):
    data = data.decode('utf-8')
```

securedrop/tests/test_i18n_tool.py
```diff
@@ -314,9 +315,13 @@ def r():
         ])
         assert 'l10n: updated nl' not in r()
         assert 'l10n: updated de_DE' not in r()
-        message = str(git('--no-pager', '-C', 'securedrop', 'show',
-                          _cwd=d, _encoding='utf-8'))
-        assert "Loïc" in message
+        if six.PY2:
```
heartsucker: This could just use `six.string_type(git(...))`.

securedrop/tests/test_i18n_tool.py
```diff
@@ -347,7 +352,11 @@ def r():
         ])
         assert 'l10n: updated nl' in r()
         assert 'l10n: updated de_DE' not in r()
-        message = str(git('--no-pager', '-C', 'securedrop', 'show',
-                      _cwd=d))
+        if six.PY2:
```
heartsucker: Likewise here for `six.string_type(git(...))`

securedrop/tests/test_integration.py
```diff
@@ -176,7 +178,10 @@ def test_submit_file(source_app, journalist_app, test_journo):
         decrypted_data = journalist_app.crypto_util.gpg.decrypt(resp.data)
         assert decrypted_data.ok
 
-        sio = BytesIO(decrypted_data.data)
+        if six.PY2:  # Python2
```
heartsucker: This feels logically wrong. In the python2 path, we're using "write bytes" against a `StringIO` which feels like it could break under non-Ascii strings. I think we should use `six.moves.StringIO` and then an if/else to ensure that we're using the `six.string_type`.

This is test-only code, so if this passes CI, we might be ok. I just worry that as we move more of the code over to using `six` that this may break.

securedrop/tests/test_journalist.py
```diff
@@ -1145,8 +1146,12 @@ def test_admin_add_user_integrity_error(journalist_app, test_admin, mocker):
                 "error")
 
     log_event = mocked_error_logger.call_args[0][0]
-    assert ("Adding user 'username' failed: (builtins.NoneType) "
-            "None [SQL: 'STATEMENT'] [parameters: 'PARAMETERS']") in log_event
+    if six.PY2:
```
heartsucker: Nit: We could extract the `builtin` to a var and format these strings. On first glace I missed the dunder around `builtin` and didn't even see why we were branching.

securedrop/crypto_util.py
kushaldas: Doing this.

securedrop/secure_tempfile.py
kushaldas: Doing this.

securedrop/secure_tempfile.py
kushaldas: This did not work for me, I just kept the old code for Python2 branch. We are planning to drop python2 support as soon as possible, and we will remove all `six` for the same.

securedrop/tests/test_i18n_tool.py
kushaldas: `str.u` does not work for our test case and creates exception on `six` itself.

securedrop/tests/test_i18n_tool.py
kushaldas: `str.u` does not work for our test case and creates exception on `six` itself.

securedrop/tests/test_integration.py
kushaldas: I am not modifying any test code here, I am keeping what ever is there, we can modify the test in future, but in a different PR.

securedrop/tests/test_journalist.py
kushaldas: I find this is easier for now as we will throw out all Python2 branch/code in next release.

securedrop/tests/test_i18n_tool.py
redshiftzero: `six.text_type` should work here

securedrop/tests/test_integration.py
redshiftzero: I think what we want here is `six.BytesIO` - `StringIO.StringIO` on Python 2, `io.BytesIO` on Python 3

redshiftzero: (though this did pass CI)

PR #4330, opened 2019-04-10
Add workaround docs for MemoryError during backup

eloquence:
Time-sensitive due to Xenial updates in progress.

This will be resolved with #4326 in 0.12.1, though we may want to
keep some version of this documentation around in case people
still encounter issues during the playbook run.

Credit to @rmol for authoring the rsync command used here, which
has been used during real-world upgrades.

## Status

Ready for review

### Checklist

### DIFF
--- a/docs/includes/backup-warning.txt
+++ b/docs/includes/backup-warning.txt
@@ -1,0 +1,19 @@
+.. note::
+  When dealing with larger backups, the ``securedrop-admin backup`` command may
+  fail with a ``MemoryError`` at this stage of the operation: "Fetch the backup
+  tarball back to the Admin Workstation".
+  
+  If this happens, a backup was successfully generated, but it is still
+  on the server. Run this command from your ``~/Persistent/securedrop``
+  directory to copy the backup your *Admin Workstation*:
+  
+  .. code:: bash
+
+    rsync -av --progress --partial app:$(ssh app ls -1rt /tmp/sd-backup* | tail -1) install_files/ansible-base/
+  
+  If the transfer fails or is interrupted, you can simply run this command again
+  to resume it.
+
+  Note that this method will only work if you have first run the 
+  ``securedrop-admin backup`` command, and the backup has successfully progressed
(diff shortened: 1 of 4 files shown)


docs/includes/backup-warning.txt
```diff
@@ -0,0 +1,19 @@
+.. note::
+  When dealing with larger backups, the ``securedrop-admin backup`` command may
+  fail with a MemoryError at this stage of the operation: "Fetch the backup
```
heartsucker: Nit: Backticks around "MemoryError" would be more consistent with existing style.

eloquence: Should be fixed. :)

docs/includes/backup-warning.txt
```diff
@@ -0,0 +1,19 @@
...
+  fail with a ``MemoryError`` at this stage of the operation: "Fetch the backup
+  tarball back to the Admin Workstation".
+  
+  If this happens, a backup was successfully generated, but it is still
+  on the server. Run this command from your ``~/Persistent/securedrop``
+  directory to copy the backup your *Admin Workstation*:
+  
+  .. code:: bash
+
+    rsync -av --progress --partial app:$(ssh app ls -1rt /tmp/sd-backup* | tail -1) install_files/ansible_base/
```
emkll: is there a reason why it's `ansible_base`? Backups are normally `install_files/ansible-base`

rmol: That is a typo that I didn't catch. It should be ansible-base.

eloquence: ah good catch, I'm guessing that's a hyphen/underscore typo in the original command

rmol: It is. You transcribed it correctly.

rmol: We didn't catch it during the upgrade because we ended up using rsync to transfer the backup back to the app server.

eloquence: (Fixed.)

PR #4349, opened 2019-04-16
Logout to revoke tokens

### LINKED ISSUE #3933: Support API token blacklisting and add API /logout endpoint
## Description

We should:

* store a table of blacklisted API tokens server-side
* add a /logout endpoint to add API tokens to the blacklist
* ensure that tokens used to access token_required endpoints are not in the blacklist, else we 403

heartsucker:
## Status

Ready for review

## Description of Changes

Fixes #3933

Added a `/logout` endpoint to the API. Added a `revoked_tokens` DB table.

## Test Plan

(edited by @redshiftzero)

### Preconditions

1. set `TOKEN_EXPIRATION_MINS` in `securedrop/journalist_app/api.py` to a smaller value, like  5 minutes or so.
2. replace `./manage.py run` with `bash` in `securedrop/bin/run` (this is so we can re-run the dev server manually to check the token cleanup works as expected with all the dev env setup in `securedrop/bin/run`)

### Testing

1. Run dev env with `make -C securedrop dev`
2. Start dev server with `./manage.py run`
3. Get a token with:

```
curl -X POST -H "Content-Type: application/json" --data '{"username":"journalist","passphrase":"correct horse battery staple profanity oil chewy","one_time_code":"875674"}' 127.0.0.1:8081/api/v1/token
```

4. Confirm you can revoke the token

```
$ curl -X POST -H "Content-Type: application/json" -H "Authorization: Token yourtokengoeshere" 127.0.0.1:8081/api/v1/logout

{
  "message": "Your token has been revoked."
}
```

5. Attempt to revoke the token once more to confirm that your token no longer works on the logout endpoint:

```
curl -X POST -H "Content-Type: application/json" -H "Authorization: Token yourtokengoeshere" 127.0.0.1:8081/api/v1/logout

{
  "error": "Forbidden",
  "message": "API token is invalid or expired."
}
```

6. Stop the server, and inspect the database:

```
$ sqlite3 /var/lib/securedrop/db.sqlite
select * from revoked_tokens;
```

Confirm that you see a single row in this database and that it contains your token expired.

7. _Provided it has been less than 5 minutes since you created that token_ restart the server and run once more with `./manage.py run`.

8. Confirm that you can still not use this token (this confirms it was not removed from the database incorrectly):

```
curl -X POST -H "Content-Type: application/json" -H "Authorization: Token yourtokengoeshere" 127.0.0.1:8081/api/v1/logout

{
  "error": "Forbidden",
  "message": "API token is invalid or expired."
}
```

9. Wait _until 5 minutes have passed_ from when you first created this token (since we set 5 minutes as the expiration time) and restart the server one last time. 

10. Confirm that the token is now gone from the revoked_tokens table in the database following the same steps as above:

```
$ sqlite3 /var/lib/securedrop/db.sqlite
select * from revoked_tokens;
```

(and for good measure if you like you can also verify that you still cannot use the token because it’s expired). 

## Deployment

Nothing special. Backwards compatible API change.

### DIFF
--- a/docs/development/journalist_api.rst
+++ b/docs/development/journalist_api.rst
@@ -56,11 +56,22 @@ HTTP Authorization header:
   Authorization: Token eyJhbGciOiJIUzI1NiIsImV4cCI6MTUzMDU4NjU4MiwifWF0IjoxNTMwNTc5MzgyfQ.eyJpZCI6MX0.P_PfcLMk1Dq5VCIANo-lJbu0ZyCL2VcT8qf9fIZsTCM
 
 This header will be checked with each API request to see if it is valid and
-not yet expired. Tokens currently expire after 8 hours, but note that clients
-should use the expiration time provided in the response to determine when
-the token will expire. After the token expires point, users must
-login again. Clients implementing logout functionality should delete tokens
-locally upon logout.
+not yet expired. Tokens currently expire after 8 hours. 
+
+Logout
+------
+
+Clients should use the logout endpoint to invalidate their token:
+
+``POST /api/v1/logout`` with the token in the HTTP Authorization header
+and you will get the following response upon successful invalidation of the
+API token:
+
+.. code:: sh
+
+  {
+  "message": "Your token has been revoked."
+  }
 
 Errors
 ~~~~~~
--- a/securedrop/alembic/versions/b58139cfdc8c_add_checksum_columns_revoke_table.py
+++ b/securedrop/alembic/versions/b58139cfdc8c_add_checksum_columns_revoke_table.py
@@ -1,4 +1,4 @@
-"""add checksum columns
+"""add checksum columns and revoke token table
 
 Revision ID: b58139cfdc8c
 Revises: f2833ac34bb6
@@ -35,6 +35,16 @@ def upgrade():
 
     with op.batch_alter_table('submissions', schema=None) as batch_op:
         batch_op.add_column(sa.Column('checksum', sa.String(length=255), nullable=True))
+
+    op.create_table(
+        'revoked_tokens',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('journalist_id', sa.Integer(), nullable=True),
+        sa.Column('token', sa.Text(), nullable=False),
+        sa.ForeignKeyConstraint(['journalist_id'], ['journalists.id'], ),
+        sa.PrimaryKeyConstraint('id'),
+        sa.UniqueConstraint('token')
+    )
 
     try:
         app = create_app(config)
@@ -77,6 +87,8 @@ def upgrade():
 
 
 def downgrade():
+    op.drop_table('revoked_tokens')
+
     with op.batch_alter_table('submissions', schema=None) as batch_op:
         batch_op.drop_column('checksum')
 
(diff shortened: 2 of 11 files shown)


securedrop/alembic/versions/e9fa666b089a_added_revoked_tokens_table.py
```diff
@@ -0,0 +1,32 @@
...
+Create Date: 2019-04-16 12:06:56.610815
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'e9fa666b089a'
+down_revision = 'f2833ac34bb6'
```
redshiftzero: flagging for other maintainers: once #4314 is merged, `down_revision = 'b58139cfdc8c'`

heartsucker: Yes, this needs to be merged with the other migration before this PR can be accepted.

redshiftzero:
Regarding the ever-increasing size of the blacklisted token table (cc @emkll @heartsucker), what about the following:

1. We store the both the token and the expiration timestamp (recall that there is an 8 hour expiration on the tokens, beyond which we don't need to store them in the blacklist) in the database. 
2. We write a small function that will check for expired tokens and delete them from the blacklist table. We could use [`@app.before_first_request`](http://flask.pocoo.org/docs/1.0/api/#flask.Flask.before_first_request) to execute this on application start such that every day we clean out old tokens. 

Thoughts?

redshiftzero:
Actually we can skip step 1 and just try to validate the tokens as part of the function described in step 2, since expired tokens won't validate.

redshiftzero:
I've implemented my suggestions above (since we are so close to the 0.13.0 cutoff, I didn't want to merge this without the cleanup of expired and revoked API tokens implemented). To facilitate testing from other people who are less familiar with the app code, I added a more detailed steps to test in the PR description above, which I have followed to verify this functionality works as expected. 

If someone could inspect/approve my last four commits and comment here if they look good, then we can merge - I :+1: @heartsucker's changes (not explicitly approving this PR to prevent accidental merge of unreviewed code in my commits).

rmol:
I ran through the test plan, and it worked as expected. Took me a minute to realize I needed to make a request after restarting the server (steps 9 and 10) to get the expired revoked token purged, because _before_first_request_, but that got me wondering if Flask has a better place for housekeeping like this, like maybe [record_once](http://flask.pocoo.org/docs/0.12/api/#flask.Blueprint.record_once)? Seems like it would be better to get this done without delaying any user request, though it's unlikely to take too long.

Also, it's not clear to me from the Flask docs what happens when a function registered either way throws an exception -- do we need to be sure that doesn't happen?

redshiftzero:
> Seems like it would be better to get this done without delaying any user request, though it's unlikely to take too long.

Agreed - and good idea re: `record_once`. So I tried this out in [this branch](revoke-token-blueprint) as an experiment (there's a generic exception raised to test the unexpected exception behavior), and here's the issue: since that method is called at blueprint registration time, if there is an unexpected exception, then the application initialization will fail - whereas if we use `before_first_request`, then if there _is_ an unexpected exception, the application will otherwise continue on merrily. Given that behavior, I think `before_first_request` seems like the most robust approach, even though the first request processing time will be a bit delayed. Let me know what you think!

rmol:
I was afraid of something like that with `record_once`. Looks good to me as it is then. :+1:

PR #4402, opened 2019-05-03
[docs] Add instructions to backup and restore workstation USBs using rsync

### LINKED ISSUE #4397: [docs] Workstation backup procedure should include a section on deploying the backup back on a thumbdrive
## Description

[Backup the Workstations](https://docs.securedrop.org/en/release-0.12.2/backup_workstations.html)  procedure lacks information on how to deploy the backup back on a thumbdrive.

## User Research Evidence

The procedure is assumed to be obvious and straightforward (just copy the files back), but there are a number of questions that need to be answered, like:
1. when setting up persistent storage on a new Tails thumbrive, to be populated from the backup, what options of the Tails persistent storage interface the user needs to choose?
1. does copy-pasting files suffice? Or does the user need to set ownership/permissions in some specific way too?
1. Should the persistent storage partition be cleared out first or not? Should files be overwritten on it? 

## User Stories

Alice wants to redeploy a backed-up Admin Workstation. She creates a Tails thumbdrive, but is unsure what options to select when setting up persistence for the backup, and afterwards what to do with files that might get overwritten when copying the backup onto the persistent storage partition.

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Fixes #4397 .

Changes proposed in this pull request: Docs changes to workstation backup instructions, to switch from manual drag-and-drop to `rsync` and include a section on restoring from a backup.
 
## Testing
Docs-only PR:
 
- work through the documented process and verify that it produces a working Workstation USB
- review for clarity and correctness

## Deployment
Will be deployed by standard docs build once merged.

### DIFF
--- a/docs/backup_workstations.rst
+++ b/docs/backup_workstations.rst
@@ -1,7 +1,10 @@
+Backing Up and Restoring Workstations
+=====================================
+
 .. _backup_workstations:
 
 Backup the Workstations
-=======================
+-----------------------
 
 .. note::  This workflow will create a single USB drive with the data backed up
  from all Tails drives. If instead you'd like to create a single duplicate
@@ -16,7 +19,7 @@ unexpected may happen.
 In all these cases, it is useful to have a backup of your data for each device.
 
 What You Need
--------------
+~~~~~~~~~~~~~
 
   #. You will need your *existing SecureDrop Tails USB sticks* (*Admin
      Workstation*, *Journalist Workstation*, and *Secure Viewing Station*).
@@ -44,7 +47,7 @@ hub which may reduce transfer speeds.
  your organization.
 
 Preparing the Backup Device
----------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 First you must boot the *primary Tails USB* drive. Ensure you set an administrator
 password set at the login screen. Then navigate to **Applications** ▸ **Utilities** ▸ **Disks**.
@@ -94,17 +97,14 @@ when you :ref:`created your Tails devices <set_up_tails>`.
 You should now have both the Backup and TailsData partition to be backed up
 mounted and ready to access.
 
+Create the backup using Rsync
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
 |Backup and TailsData Mounted|
 
-Open a terminal with administrative privileges by going to
-**Applications** ▸ **System Tools** ▸ **Root Terminal**.
+Open a terminal by going to
+**Applications** ▸ **Favorites** ▸ **Terminal**.
 
-|Root Terminal|
-
-.. note::
- If you can't find the "Root Terminal" window, it might be because an
- administrator passphrase wasn't set when you logged in to Tails. If
- that's the case, you'll need to restart Tails and set one at the login screen.
 
 Next, create a directory on the Backup USB for the device to be backed up - the
 command below creates a directory named ``admin-backup``:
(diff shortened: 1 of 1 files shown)


docs/backup_workstations.rst
```diff
@@ -118,16 +115,83 @@ Then, copy the contents of the device's persistent volume to the directory using
...
+          the drive can be safely removed. Wait until its entry  is removed from
+          the lefthand column of the file manager.
+
+Restoring a Workstation from a Backup
+=====================================
+
+To recreate a backed-up *Admin Workstation*, *Journalist Workstation*, or 
+*Secure Viewing Station* Tails USB,  you will need your Backup USB, a master
+Tails USB, and a fresh USB stick to be set up as the new device. You will also 
+need an airgapped machine, as the process is similar to that originally used to
```
eloquence: Overall these docs are super-clear, thank you!

I'm a bit confused by this part; you are referring the need to have a master _and_ the new Tails device, each with a persistent volume. What's unclear to me from reading the instructions from here on is which persistent volume we're restoring to, and whether a master device is always needed.

In short, I think that part may still require some untangling and explanation upfront, to help the user understand at this point in the process what the role of each Tails drive is. The use of a term like "master" may be unfamiliar, as well.

zenmonkeykstop: That's a fair point, it could do with some upfront definitions so I'm not just winging it in how I refer to each device. Will update.

zenmonkeykstop: have now done so!

docs/backup_workstations.rst
```diff
@@ -118,16 +115,90 @@ Then, copy the contents of the device's persistent volume to the directory using
...
+=====================================
+
+To recreate a backed-up *Admin Workstation*, *Journalist Workstation*, or 
+*Secure Viewing Station* Tails USB,  you will need 
+
+- your Backup USB containing the persistent volume to be restored,
+- a fresh USB stick to be set up as the new Tails USB,  
+- an airgapped machine and a USB with Tails already installed, referred to as 
+  the host Tails USB in this document. The host Tails USB is only used to 
+  transfer files between the Backup USB and the new Tails USB.
```
eloquence: Should we say something here about the persistent volume of the host USB, or would that make things more confusing? For example:

> It does not require a persistent volume; if one is configured, you do not need to unlock it during this process.

I'm thinking it might help to give folks peace of mind that they're not accidentally copying to the wrong persistent volume.

zenmonkeykstop: In order to do so, you'd need to use a completely different path, as the PV is mounted under `/live/persistence/` rather than under the default location for attached storage in `/media/amnesia,' so my feeling is that this is probably not worth the added complexity. There's no mention of the PV in the step where the USB is started and it wouldn't be a problem if they did unlock it anyway.

eloquence: :+1:

docs/backup_workstations.rst
```diff
@@ -118,16 +115,90 @@ Then, copy the contents of the device's persistent volume to the directory using
...
-USB drive, unmount the Backup partition and store the drive somewhere safely.
+USB drive, unmount the Backup partition by clicking its Eject button. Wait until
+the Backup USB can be safely removed, and store it somewhere safely.
+
+.. note:: After the Eject button is clicked, it may be take some time before 
+          the drive can be safely removed. Wait until its entry  is removed from
+          the lefthand column of the file manager.
+
+Restoring a Workstation from a Backup
+=====================================
```
eloquence: This translates to a level 1 header, which means that it will be rendered similarly to a separate chapter both in the sidebar navigation and the table of contents:

backup

I think this is probably not what we want (as fas I can tell it's the only doc with more than one H1); instead I suggest we keep "Backup" as the H1 and create "Creating" and "Restoring" H2s.

zenmonkeykstop: Done and done!

eloquence:
Looks great to me on visual review. I'll take a pass tomorrow at following these instructions (both backup and restore). I have never backed up the Tails drives for my own test install, so this will be a useful exercise for me as well. :)

eloquence:
Just making a note from my first pass review that I encountered issues with USB drives that use identical serial numbers being mapped to the same disk in Tails, which can confuse the GNOME Disks Tool, prevent ejecting drives in Nautilus, and probably has other unintended effects. When this happens, it's impossible to do the drive-jockeying work required for a backup. This appears to be the (likely WONTFIX) issue described in https://github.com/storaged-project/udisks/issues/490 .

I don't know how common it is that USB drives from the same manufacturer share the same serial number string; it happened to me with [these TOPESEL drives](https://www.amazon.com/gp/product/B0756VYD2P/). I don't think we need to address it in this PR, but given that many news organizations are likely to order a batch of USB drives from a single manufacturer, I don't think we can rule it out as an obscure scenario, so I'll submit a follow-up docs PR to add a very small note as a precautionary measure.

eloquence:
Another quick note from following the instructions: "Click on the disk on the left side column" should specify which disk (it's the drive they want to back up). Unrelated to the goal of this PR, so can be updated separately.

docs/backup_workstations.rst
```diff
@@ -118,16 +118,92 @@ Then, copy the contents of the device's persistent volume to the directory using
...
+.. _restore_workstations:
+
+Restoring a Workstation from a Backup
+-------------------------------------
+
+To recreate a backed-up *Admin Workstation*, *Journalist Workstation*, or 
+*Secure Viewing Station* Tails USB,  you will need 
+
+- your Backup USB containing the persistent volume to be restored,
+- a fresh USB stick to be set up as the new Tails USB,
```
eloquence: Right now you're using "fresh USB" and "new USB" interchangeably to refer to the recovery target. Terms like "fresh" and "new" can be confusing when juggling multiple USB drives (at least 3, possibly more if they're doing multiple restores), Tails cloning processes, etc.

I would reword this as follows:
> a fresh USB stick to serve as the restored *Admin Workstation*, *Journalist Workstation* or *Secure Viewing Station* USB, referred to as the "workstation USB" in this document

and then update subsequent sections, "Prepare the new workstation USB", etc.

zenmonkeykstop: I refer to it as the "fresh USB stick" until Tailsified, at which stage it changes to the "new Tails USB" - "workstation USB" would be ambiguous IMO as there are other workstation USBs referred to in the doc.

How about the "blank USB stick" before formatting, the "new workstation USB" afterwards?

eloquence: > How about the "blank USB stick" before formatting, the "new workstation USB" afterwards?

:+1:, that sounds very clear to me :)

zenmonkeykstop: Aaand done.

eloquence: LGTM!

eloquence:
I have verified these instructions by creating and restoring an Admin Workstation USB backup, and verifying that it does indeed work as expected. The instructions are clear and complete, and LGTM to merge by a maintainer. :) 

The Tails community does appear to want to make this process easier, so we may want to comment on https://redmine.tails.boum.org/code/issues/9889 to give input on what an ideal process would look like from our perspective. My sense is that a "Clone everything" feature as described [here](https://tails.boum.org/blueprint/backups/#index12h2) would be a huge win in terms of usability.

docs/backup_workstations.rst
```diff
@@ -118,16 +118,92 @@ Then, copy the contents of the device's persistent volume to the directory using
...
+folder to the new workstation USB's persistent volume. For example, if the backup
+folder to be copied is named ``admin-backup``, run the following command:
+
+.. code:: sh
+ 
+  sudo bash -c "rsync -a --info=progress2 --no-specials --no-devices \
+      /media/amnesia/Backup/admin-backup/ /media/amnesia/TailsData && sync"
+
+.. note:: Note that the first path argument ends with a ``/``, but the second 
+          path argument should not!
```
conorsch: Nit: The trailing slash on the _second_ argument makes no difference: with or without, the _contents_ of `/media/amnesia/Backup/admin-backup/` (but not the `admin-backup` dir itself) will be copied into `/media/amnesia/TailsData/`. 

The trailing slash on the _first_ argument is indeed crucial. Perhaps we can place a trailing slash on both args, for consistency, and update the "note" to something like:

> The directory names in the command above end with a ``/`` to ensure that only the contents of the directories are transferred, without recreating the parent directories at the destination. If you omit the trailing ``/``, the files will not be copied to the correct location for Tails to operate.

docs/backup_workstations.rst
```diff
@@ -118,16 +118,92 @@ Then, copy the contents of the device's persistent volume to the directory using
...
 .. code:: sh
 
-  rsync -a --info=progress2 --no-specials --no-devices \
-      /media/amnesia/TailsData/ /media/amnesia/Backup/admin-backup
+  sudo bash -c "rsync -a --info=progress2 --no-specials --no-devices \
+      /media/amnesia/TailsData/ /media/amnesia/Backup/admin-backup && sync"
 
-Once complete, unmount the TailsData partition.
+.. note:: Note that the first path argument ends with a ``/``, but the second 
+          path argument should not!
```
conorsch: See comment below regarding trailing slashes and rsync; same nit applies here.

zenmonkeykstop: Added similar abbreviated version of the above.

PR #4413, opened 2019-05-07
[docs] Standardize Submission Key language and usage

### LINKED ISSUE #2579: Terminology guide should explain "Submission Key"
# Feature request

## Description

The docs in the Terminology guide explain common jargon used throughout the project, e.g.

* Source
* Journalist
* Application Server
* Admin Workstation

No mention, however is made of the *SecureDrop Submission Key* in the Terminology guide. Admins have asked for clarification between various forms of reference to the submission key, e.g.

* Application Server Key
* SecureDrop Key
* SVS key
* Airgapped private key

Having a single reference to point to as the "correct" terminology would greatly clarify matters for autodidactic Admins.

## User Stories
As a SecureDrop Administrator, I want to ensure I'm communicating clearly when making support requests. After reading the docs, I'm still confused about what to call the "SecureDrop Submission Key".

As a SecureDrop maintainer, I want to point to a single answer in the docs to explain crucial components of the project, both to existing Admins and first-time contributors.

### LINKED ISSUE #4122: [docs] Journalist guide assumes that submission key requires a passphrase
The [journalist guide](https://docs.securedrop.org/en/latest/journalist.html) currently states:

> A prompt will ask you for the application PGP key passphrase to decrypt the document.

Aside from the inconsistent phrasing (see #2579),  this assumption is not warranted. The SVS docs [state](https://docs.securedrop.org/en/latest/generate_securedrop_application_key.html) that it is safe to generate the submission key without a passphrase, and they explain the process for doing so.

This statement therefore needs to be qualified to avoid confusion.

eloquence:
This also revises the SVS description slightly (nowhere else do we explicitly instruct people to use the SVS for _generating_ journalist keys, and in this context it feels more confusing than helpful). It renames a file that used the old "application key" term.

I chose to use the shorter "Submission Key" instead of "SecureDrop Submission Key" since other terms like *Journalist Workstation* are also SecureDrop-specific, and the convention to italicize such terms makes it reasonably clear that they carry a context-specific meaning.

I've not disabled whitespace trimming for this PR as we should IMO generally avoid trailing whitespace, but note that this introduces some whitespace removal along the way.

Resolves #2579
Resolves #4122

## Status

Ready for review

## Checklist

### DIFF
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -383,7 +383,7 @@ Currently, when one admin pushes changes in ``site-specific`` to the server, the
 changes will not sync to the local ``site-specific`` file on the remaining admin workstations.
 Without being aware of changes made to ``site-specific``, admins run the risk of pushing old
 information to the servers. This can affect the receipt of OSSEC alerts, viability of the
-Submission Key, among other critical components of the SecureDrop environment.
+*Submission Key*, among other critical components of the SecureDrop environment.
 
 There are multiple ways to avoid pushing out-of-date information to the servers.
 We recommend admins establish a secure communication pipeline to alert fellow admins
--- a/docs/generate_submission_key.rst
+++ b/docs/generate_submission_key.rst
@@ -1,15 +1,15 @@
-Generate the SecureDrop Submission Key
-======================================
+Generate the *Submission Key*
+=============================
 
 When a document or message is submitted to SecureDrop by a source, it is
-automatically encrypted with the *SecureDrop Submission Key*. The private part
+automatically encrypted with the *Submission Key*. The private part
 of this key is only stored on the *Secure Viewing Station* which is never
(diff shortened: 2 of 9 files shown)


eloquence:
If this looks good, I recommend we open a follow-up issue to similarly standardize references to the OSSEC keypair.

One remaining inconsistency is that we refer to the Submission Key as the "journalist key" (in the `journalist-key` and `why-journalist-key` URLs, and in the page title "Why download the journalist's public key?") in the *Source Interface*. I would also recommend addressing this as a follow-up code change.

conorsch:
> I've not disabled whitespace trimming for this PR as we should IMO generally avoid trailing whitespace

Big fan of cleaning up here. :ok_hand:

> I recommend we open a follow-up issue to similarly standardize references to the OSSEC keypair.

Totally agree.

Issue #4411, opened 2019-05-07
Feature idea: per-instance exporting of Tor metrics for monitoring by Prometheus


ageis:
## Description

Every SecureDrop runs at least two Tor daemons. From the perspective of the IT administrator at a news organization, it might be useful for them to have the ability to ingest statistics about Tor activity.  I have some prior experience with setting up introspection and monitoring for a fleet of exit nodes, and I recommend https://github.com/atx/prometheus-tor_exporter. It is based on stem plus the prometheus_client PyPi library. This means it could be integrated directly into the SecureDrop web application, or run separately as a systemd service.

By an exporter what is meant is that metrics are published to an HTTP endpoint. This can then be scraped by Prometheus, where then Prometheus may be the backend for a powerful visualization platform like Grafana, or an alerting/notification solution like Alertmanager. One probably would not want this traffic information to be published publicly due to the risk of timing attacks and deanonymization, etc.. Well, Prometheus is able to scrape targets via Tor when combined with an HTTP proxy such as Polipo or Privoxy, so the metrics could be served over an authenticated Tor hidden service. Even just polling the page manually every once in a while, say, only from the internal network, can provide insight. 

This would help FPF+admins keep an eye on issues in the Tor network, instances that may not be reachable or able to establish circuits, extremely low or high amounts of traffic, plus whether an instance has updated. You could also just put the service there and make it optional whether orgs want to turn it on. Or this could be a feature specific to "dev".

With the open source exporter I linked, you get the following information updated on a regular basis: 

```  Name              |  Description
--------------------|-----------------------
tor_written_bytes   | Running total of written bytes.
tor_read_bytes      | Running total of read bytes.
tor_version{version="..."} | Tor daemon version as a tag
tor_version_status={version_status="..."} | Tor daemon version status as a tag
tor_network_liveness | Network liveness (1.0 or 0.0)
tor_reachable{port="OR\|DIR"} | Reachability of the OR/DIR ports (1.0 or 0.0)
tor_circuit_established | Indicates whether the daemon is capable of establishing circuits (1.0 or 0.0)
tor_dormant | Indicates whether tor is currently active (1.0 or 0.0) (note that 1.0 means "dormant", see the specs for details)
tor_effective_rate | Shows the effective rate of the relay
tor_effective_burst_rate | Shows the effective burst rate of the relay
tor_fingerprint{fingerprint="..."} | Node fingerprint as a tag
tor_nickname{nickname="..."} | Node nickname as a tag
[...]

PR #4414, opened 2019-05-08
Adds admin-viewable-only internal Flask and SecureDrop metrics

ageis:
Internal metrics! Somewhat related to #4411, but instruments the SecureDrop web application with its own exporter that can help admins and developers get a sense of general activity as well as the performance via response times, frequency of requests for specific static assets, and the appearance of any HTTP errors on Flask views/routes/endpoints.

Affects the journalist interface only, it would not be good thinking to implement this in the source app; there's no reason.

Two new Python libraries are required: prometheus_client and prometheus_flask_exporter. For more info, please see:
- https://github.com/prometheus/client_python
- https://github.com/rycus86/prometheus_flask_exporter

At `/metrics` we have HTTP request metrics provided by the flask_exporter, which is by default tracking all routes and paths within the journalist app.

To disinclude an endpoint/view in the future just prepend the decorator:
```
    @internal_metrics.do_not_track()
```
Secondly, there're five internal SecureDrop metrics being exported, one as info,
the rest as gauges. (there are other types one can instrument with, such as Counter, Enum, Summary, Histogram...)

Those are:

-  securedrop_version=version.__version
-  securedrop_users[username, id, created, last] = len(Journalist.query.all)
-  securedrop_submissions[size_bytes, downloaded, author] = len(Submission.query.all)
-  securedrop_replies = len(Reply.query.all)
-  securedrop_sources[designation, flagged, last, interactions] = len(Source.query.all)
-  securedrop_logins[when, user_id, username] = len(JournalistLoginAttempt.query.all)

Analytical possibilities are plenty. You could graph submission sizes, login trends, source and journalist profilicness in comparison with others, etc. were-there-submissions-today? Well, has the securedrop_submissions metric incremented? :P I anticipate some debate about whether everyone thinks this is a great idea, whether it adds unnecessary code or could be a risk in some way. I personally don't see that. And in any event, we wouldn't be alone in this approach; there's plenty of inspiration out there:

- https://medium.com/@andrei.chernyshev/prometheus-python-flask-8487c3bc5b36
- https://blog.codeship.com/monitoring-your-synchronous-python-web-applications-using-prometheus/
- https://burhan.io/flask-application-monitoring-with-prometheus/
- https://blog.insightdatascience.com/monitoring-your-flask-application-on-kubernetes-with-prometheus-1d06db99b244
- https://github.com/kubernetes-for-developers/kfd-flask

## Status

Ready for review. Not sure what would be needed in terms of tests, feel free to supply guidance on that.

## Description of Changes

Adds new route `/metrics` to the journalist interface, with a button link to it on the admin page. Adds two Python libraries to requirements. 

## Testing

Check out my branch and stand up a dev server using the Docker method e.g. `make dev`. Then login to the journalist interface, go to /metrics and check it out.

## Deployment

No special considerations for deployment; I'm not sure if app code is still installed from a Debian package but I assume you have sound means of updating the Pip requirements.

## Checklist

I'll complete the below part later. Tests are failing due to mypy, which I doubt is my fault.

### DIFF
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -16,14 +16,15 @@ import platform
 
 from crypto_util import CryptoUtil
 from db import db
-from journalist_app import account, admin, api, main, col
+from journalist_app import account, admin, api, main, metrics, col
 from journalist_app.utils import (get_source, logged_in,
                                   JournalistInterfaceSessionInterface,
                                   cleanup_expired_revoked_tokens)
+
 from models import Journalist
 from store import Storage
 from worker import rq_worker_queue
-
+from prometheus_flask_exporter import PrometheusMetrics
 import typing
 # https://www.python.org/dev/peps/pep-0484/#runtime-or-type-checking
 if typing.TYPE_CHECKING:
@@ -84,7 +85,8 @@ def create_app(config):
         adjectives_file=config.ADJECTIVES,
         gpg_key_dir=config.GPG_KEY_DIR,
     )
-
+    internal_metrics = PrometheusMetrics(app)
+    internal_metrics.info('securedrop_version', 'SecureDrop version', version=version.__version__)
     app.config['RQ_WORKER_NAME'] = config.RQ_WORKER_NAME
     rq_worker_queue.init_app(app)
 
@@ -174,6 +176,7 @@ def create_app(config):
     app.register_blueprint(col.make_blueprint(config), url_prefix='/col')
     api_blueprint = api.make_blueprint(config)
     app.register_blueprint(api_blueprint, url_prefix='/api/v1')
+    app.register_blueprint(metrics.make_blueprint(config), url_prefix='/metrics')
     csrf.exempt(api_blueprint)
 
     return app
--- a/securedrop/journalist_app/metrics.py
+++ b/securedrop/journalist_app/metrics.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*-
+
+from flask import Blueprint, Response
+from models import Journalist, JournalistLoginAttempt, Reply, Source, Submission
+from journalist_app.decorators import admin_required
+from prometheus_client import Gauge, generate_latest, CONTENT_TYPE_LATEST
+
+SECUREDROP_USERS = Gauge(
+    "securedrop_users",
+    "Number of journalists who can access SecureDrop and their details",
+    labelnames=["username", "user_id", "created", "last"],
+)
+
+SECUREDROP_SUBMISSIONS = Gauge(
+    "securedrop_submissions",
+    "Number of submissions, from which source, their size in bytes, whether they've been read",
+    labelnames=["size_bytes", "downloaded", "author"],
+)
+
+SECUREDROP_REPLIES = Gauge(
+    "securedrop_replies", "Number of replies from journalists to sources"
+)
+
+SECUREDROP_SOURCES = Gauge(
+    "securedrop_sources",
+    "Number of sources who have made submissions, whether they're flagged, number of interactions and when they last logged in",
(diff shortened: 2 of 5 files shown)


redshiftzero:
This is an interesting approach, though I do want to flag the following before you spend more time working on this: in terms of FPF monitoring, we don't have the HidServAuth information to access individual news organization's Journalist Interfaces (nor do we want to start collecting this information). I'm also not aware of any news organization _using_ Prometheus so I think we need to take a few steps back (relevant: check out [these issues](https://github.com/freedomofpress/securedrop/labels/goals%3A%20reduce%20IDS%20noise) related to monitoring).

Regarding the additional risk introduced by these code paths and maintenance considerations: we'd be adding two dependencies (which we'd need to review and start tracking security issues therein) - one of the added dependencies has 58 GitHub stars and is not used by any other major project afaict. We should avoid adding new/rarely used dependencies unless there is a very strong motivation as we'll need to take over maintenance of dependencies if they fall into disrepair _or_ revert the shipped functionality that uses them, something that we should also avoid.

ageis:
> This is an interesting approach, though I do want to flag the following before you spend more time working on this: in terms of FPF monitoring, we don't have the HidServAuth information to access individual news organization's Journalist Interfaces (nor do we want to start collecting this information). I'm also not aware of any news organization using Prometheus so I think we need to take a few steps back (relevant: check out these issues related to monitoring).

OK. If any news organizations are using Kubernetes, then they are using Prometheus without being aware of it. I'm not specifically motivated for FPF to do collection either, this is there for developers and admins. (but, see users at bottom of https://prometheus.io/)

> Regarding the additional risk introduced by these code paths and maintenance considerations: we'd be adding two dependencies (which we'd need to review and start tracking security issues therein) - one of the added dependencies has 58 GitHub stars and is not used by any other major project afaict. We should avoid adding new/rarely used dependencies unless there is a very strong motivation as we'll need to take over maintenance of dependencies if they fall into disrepair or revert the shipped functionality that uses them, something that we should also avoid.

This is a fair point that is valid for that and the repo I cited in #4411. I can assure you it won't ever be a concern for prometheus_client, the main library however.

ageis:
One idea (provided enough RAM) could be to expand the functions of the Monitor Server by putting stuff like ELK (Wazuh version of OSSEC) and Prometheus+Grafana on there. So there could be some dashboards hosted on another authenticated hidden service to make everything more DevOpsy and SRE-ish.

KwadroNaut:
> One idea (provided enough RAM)

It might be currently underutilized, but I'm a bit affraid to beef up requirements too much, right now older hardware can be easily repurposed.

redshiftzero:
Hey @ageis, thanks for working on this. It's a cool proof of concept, but I'm going to close this PR unmerged. Unfortunately it's not something we can add to SecureDrop given the constraints I mention above plus I don't really think that admin users would get a ton of value from Prometheus-exportable metrics about e.g. submissions/source/journalist activity. If you're interested in contributing, check out the [help wanted](https://github.com/freedomofpress/securedrop/labels/help%20wanted) label or the [OSSEC label](https://github.com/freedomofpress/securedrop/labels/goals%3A%20reduce%20IDS%20noise) which we would accept patches for since many of those are issues that we know admins want fixes for.

ninavizz:
FWIW, It is an aspiration to post more in Gitter and in the UX Repo, information learned from newsrooms about what value users experience in which types of things... and where probable user needs are that are currently unmet. As a UX contributor, I'd love to see more community input on feature ideas/planning; but they need clear user motivations and needs mapping validated against data, first, on an aside from the technical dependencies stuff redshiftzero cited. Lean resources availability good times.  FYI, though—I was personally very intrigued by the idea of learning more about SD users, from such an integration!

PR #4416, opened 2019-05-08
Remove Trusty-specific logic

emkll:
## Status

Ready for review

## Description of Changes

Fixes #4341 

Removes Trusty specific logic from:

* Application logic on Source and Journalist for Trusty-specific warnings
* CI targets
* Molecule Upgrade scenarios
* Molecule libvirt Scenarios
* Molecule Virtualbox scenarios
* Molecule Build scenarios

## Testing

- [ ] `make build-debs`  completes successfully (should be covered by CI)
- [ ] `molecule converge -s libvirt-staging-xenial`  completes successfully (should be covered by CI)
- [ ] `make staging` completes successfully in VirtualBox on a Mac
- [ ] `make staging`  completes successfully in Qubes
- [ ] Upgrade scenario completes successfully (`make upgrade-start` and  `make upgrade-test-local`)
- [ ] There are no more references to Trusty that need to be removed

## Deployment

1. Changes to Source and Journalist interfaces will be shipped via the securedrop-app-code package
2. All other changes are dev/ci only

### If you made non-trivial code changes:

- [x] I have written a test plan and validated it for this PR

PR #4422, opened 2019-05-09
Remove Trusty-specific logic in Ansible scripts

### LINKED ISSUE #4341: Remove Trusty-specific logic from application code and dev env
## Description

Some platform-specific logic was added to handle the upgrade from Trusty to Xenial and should be removed:
- Source Interface Submission disable logic: https://github.com/freedomofpress/securedrop/pull/4325
- Admin Interface alert for end-of-life OS: https://github.com/freedomofpress/securedrop/pull/4055

There is also some Trusty-specific logic in the development environment: 
- [Docker images](https://github.com/freedomofpress/securedrop/tree/develop/securedrop/dockerfiles/trusty)
- [CI targets](https://github.com/freedomofpress/securedrop/blob/develop/.circleci/config.yml#L59
- The use of `BASE_OS` in Makefile targets
- Upgrade testing scenarios and boxes

Since Trusty will no longer be supported after 0.12.2, we should remove this logic to improve the maintainability of the codebase and reduce the risk of errors.

emkll:
## Status

Ready for review
## Description of Changes

Fixes #4341 : Some Trusty-specific logic was not completely removed in #4416 

## Testing

- [ ] CI passes
- [ ] There are no further references to trusty

## Deployment

These changes are limited to the Ansible scripts and development environment. The install process will no longer work under Trusty


redshiftzero:
some other places we should update as part of this change:

1. we have some xenial/trusty branching in `securedrop-config`'s `postinst` for issue #4104, we could remove that now

2. `test_platform` in `molecule/testinfra/staging/common` needs update

3. `molecule/vagrant-packager/package.py` has trusty references (will be confusing if we don't at least update the comments to indicate we removed trusty logic)

PR #4425, opened 2019-05-09
Add first and last name to journalist table and web app

### LINKED ISSUE #4251: Add Identity text fields in Admin interface
## Description

In the Qubes Workstation Client, journalists will have first-initial/last-initial square "ID Badges" to represent themselves and each other, in message threads and task logs. Example, just below. 

In order to support this, the existing Admin interface needs to have capabilities to support those ID badges with data.

Screenshots of the most obvious solution to this, are below. Per the convo thread below this comment, alternative options could ideally be considered and tested prior to an implementation.
- FN/LN fields should not be required, as they'll only deliver value on the Workstation Client
- Both name fields should be longer than the existing username field, to accommodate for lengthy English hyphenated names and lengthy first and last names in non-Western alphabets/cultures.
- When onboarding a newsroom to the Workstation Client product, it should be made clear that adding FN/LN entries for all journalists (even if just as single initials) is recommended.
- In attached mockups, a subtle few tweaks were also made to the list of users in the Admin list. Sorry, I had to. Only done to facilitate legibility w/ existing assets and type styles.
   - Left-align all fields
   - Increased linespacing (or top/bottom row/table padding)
   - Add a pale grey horizontal line between entries. 
      - This will help make the list more readable with multiple empty FN/LN entries.
   - Reduce type size and make grey, the timestamp entries. Another legibility tweak. 
   - Move the "Edit" column to be the first on the left, and the "Delete" column to be the last on the right.

## User Research Evidence

The UX kid just says to (and the Workstation Client will surface collaboration info much less easily, if this capability is not added). "LL" for Lois Lane, and "SY" for Steve Yzerman is an existing messaging-app and email-app paradigm that has been validated in testing as having low cognitive friction, while also being easy to implement (s'long as the columns in data tables already exist, hence this Issue). 

## User Stories

As the Admin for the Daily Planet, I want to give my SecureDrop journalist users first names and last names. This will help surface cross-team activity clearly and in a fashion that's easy to comprehend, across the super sweet new Qubes Workstation Client. Because the Qubes Workstation is SO awesome and all our newsroom's journalists want to use it, I'd also like the list of all users tidied-up a touch so that it's more legible beyond 3-4 line items (a little ux bird hinted that'd help).

As a Journalist using the Qubes Workstation Client, I want to be able to see my own activity reflected in an easy to recognize fashion. Likewise, I want to be able to see the contributions and activities of my colleagues in a similarly easy to recognize fashion. 

image

image

creviera:
## Description of Changes

Resolves #4251

* Add first and last name columns to the Journalists table
* Add ability to specify first and last name when creating a new account as an admin
* Add ability to change first and last name of a user as an admin
* Add ability to change first and last name as a user

## Test plan for reviewers

- [x] Successfully add a new account as an admin from the journalist interface 
- [x] Successfully change the name of an account that is not yours as an admin from the admin page
- [x] Successfully change your own name as a non-admin from the accounts page
- [x] Successfully change your own name to empty strings a non-admin from the accounts page


securedrop/journalist_app/admin.py
```diff
@@ -172,6 +176,24 @@ def edit_user(user_id):
                 else:
                     user.username = new_username
 
+            try:
+                first_name = request.form['first_name']
+                Journalist.check_name_acceptable(first_name)
+                if first_name != user.first_name:
```
heartsucker: These `if` checks are unnecessary since SQLAlchemy uses listeners generated by the metaclass to check for changes to the object.

securedrop/journalist_app/admin.py
```diff
@@ -172,6 +176,24 @@ def edit_user(user_id):
...
+                    user.first_name = first_name
+            except FirstOrLastNameError as e:
+                flash('Invalid name: ' + str(e), 'error')
```
heartsucker: These flashes need to be wrapped with a `gettext` and use a format string like `Invalid name: {name}`. This is done because some languages might not use a colon and want to reorder things.

securedrop/journalist_templates/edit_account.html
```diff
@@ -4,14 +4,30 @@
...
   <form method="post">
     <input name="csrf_token" type="hidden" value="{{ csrf_token() }}">
     <p>
-      <label for="username">{{ gettext('Change username') }}</label>
+      <label for="username">{{ gettext('Username') }}</label>
       <input name="username" id="username" type="text" placeholder="{{ user.username }}">
     </p>
+    <p>
+      <label for="first_name">{{ gettext('First name') }}</label>
+      {% if not user.first_name %}
```
heartsucker: Nit: I think it's more readable to just do `user.first_name or ''` in the one place this is used. If not, I think doing this is a little more readable.

```jinja
{% set first_name = user.first_name or '' %}
```

securedrop/models.py
```diff
@@ -291,6 +291,23 @@ class InvalidUsernameException(Exception):
...
 
 
+class FirstOrLastNameError(Exception):
+    """Generic error for names that are invalid."""
+
+
+class InvalidNameLength(FirstOrLastNameError):
+    """Raised when attempting to create a Journalist with an invalid name length."""
+
+    def __init__(self, name):
```
heartsucker: Possibly also here with a `super` with a more clear message instead of just the `__str__` below.

securedrop/models.py
```diff
@@ -400,13 +428,31 @@ def set_password(self, passphrase):
 
         self.passphrase_hash = argon2.using(**ARGON2_PARAMS).hash(passphrase)
 
+    def set_name(self, first_name, last_name):
```
heartsucker: We check the min length is greater than zero which is always true. If zero is an acceptable length, we sjould just remove this code path (and the same in the exceptions and validations and such).

securedrop/tests/migrations/migration_b58139cfdc8c.py
```diff
@@ -206,6 +225,18 @@ def check_downgrade(self):
                 except NoSuchColumnError:
                     pass
 
+            journalists_sql = "SELECT * FROM journalists"
+            journalists = db.engine.execute(text(journalists_sql)).fetchall()
+            for journalist in journalists:
+                try:
```
heartsucker: Looking at these makes me realize I wrote my test for the `checksum` column wrong. All three of these `NoSuchColumnError` checks should be:

```python
with pytest.raises(NoSuchColumnError):
    row['column_name']
```

This will catch the case where the column exists. The two you added have the edge case of passing when those columns are `None`, and the one I wrote is even more wrong of catching none of the cases where that column exists.

securedrop/tests/migrations/migration_b58139cfdc8c.py
```diff
@@ -182,7 +198,10 @@ def load_data(self):
 
     def check_downgrade(self):
         '''
-        Verify that the checksum column is now gone.
+        - Verify that the checksum column is now gone.
+        - Verify that the migration worked by checking that the journalist added in load_data has
```
heartsucker: This line of the docstring is wrong. On the downgrade the `first_name` and `last_name` columns won't exist. I think this was meant for `check_upgrade`.

securedrop/journalist_templates/edit_account.html
heartsucker: Also these checks exist in two places and we may want to DRY it some. Though this may not be possible.

heartsucker:
Also I should have noted in the above review that this does pass the test plan described.

creviera:
In case @redshiftzero finishes this PR while I work on finishing another outstanding PR, here's what's left:

* Test (as you can see :))
* Deciding on how to handle null first and last names, see notes below and refer to @ninavizz for UX feedback, this may or may not require an update to the code.

---

Currently we have a username field that cannot be `None`. We use a `placeholder` to show the current username in gray when editing the form. When you edit the field, the placeholder text will not entirely disappear, which works for username since we don't allow null values anyway. 

The problem with implementing first and last name this way is that we do want to allow null values for first and/or last name (see https://github.com/freedomofpress/securedrop/issues/4251#issuecomment-492763053 if you want to continue the conversation).

You can see how username and first and last name are implemented in the same way here:
placeholder-always-in-background-no-none-values-allowed

There is no way to save null values here.

Solution A:
Clicking "Update" with only the placeholder text erases the name. See the gif above to imagine how that might look. The two downsides to this solution are (1) it's inconsistent with username because saving a username field with None does not erase the username (username cannot be None anyway), and (2) it may look confusing to users trying to erase the first and last name fields when the placeholder text is still in the background.

Solution B (the one I went with in this PR using simple CSS:
editable-and-blank-for-none-value

The two downsides to this solution are (1) it's inconsistent with username, and (2) i can't figure out a CSS (non-javascript) way of making it so that the field remains black text after editing in between saves. To address (1), we could implement username this way too.

securedrop/alembic/versions/b58139cfdc8c_add_checksum_columns_revoke_table.py
```diff
@@ -36,6 +36,10 @@ def upgrade():
     with op.batch_alter_table('submissions', schema=None) as batch_op:
         batch_op.add_column(sa.Column('checksum', sa.String(length=255), nullable=True))
 
+    with op.batch_alter_table('journalists', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('first_name', sa.String(length=255), nullable=True))
+        batch_op.add_column(sa.Column('last_name', sa.String(length=255), nullable=True))
```
redshiftzero: since this isn't making it into 0.13.0 (feature freeze has passed), we'll need to move the migrations for this feature into a new migration targeted for 0.14.0

ninavizz:
I'm not sure exactly what the question is, here—but I would prefer no placeholder text exist for `first name` or `last name` fields (or even the `username` field), and to make the text black inside all 3. 

Allie, I know we discussed the color-change weirdness between the click-into and static states of that placeholder text, after you approached me about how to resolve that issue last week... though zooming-out from the broader interaction, no placeholder text at all makes more sense to me, in all 3 fields. 

Placeholder texts can be really nice to have, but too often cause more confusion than they add value. They really need to be done well, or not done at all—and I know they're done inconsistently throughout the Admin/Journalist UIs. For the name things, I'd rather just be safe and not have any placeholders—and then have the text be black, when clicked-into and typed with.

Also, a lot more padding to the left of the label `last name`, so there's a clear margin between it and the `first name` field. 

Feeling like an inconsistent jerk, recommending the above. Let me know if it makes things harder or easier, or if you'd like to chat this through.

creviera:
My description of the problem isn't super clear. I tried! 

The problem is that we've been trying to figure out a way to add the first and last name fields using the current design, or something similar, so that we can keep the scope of this change small.

If we stick with the current design of showing saved database values in text boxes, then I think it's desirable to allow the user to clear and see an empty text box when they want to erase their name.

If we remove placeholders all together, then there should be a place to show what's currently saved in the db. This is a larger user-facing change that needs to be discussed further - next UX meeting perhaps!

I'm curious what @redshiftzero thinks. Any other ideas on how to get this change out and keep it small? Do you think we should discuss a redesign of how we show current database values and placeholders before moving forward with this change?

ninavizz:
Oyy... `facepalm`. Thank you for the additional explanation, Allie! Your initial one was plenty clear to anyone familiar with the code. I should be more familiar with the code. Also: ouch wrt the problem at hand. Also wanting to avoid scope-creep. Will fiddle around with Erik's SD instance today, to see what makes sense for now. Also curious what others think; and, wanting to keep any "redesign" of things out of this 'lil ticket. <3

redshiftzero:
I think what makes sense is that the "placeholder" text here just provides the current value of that field if it's defined. If there is no value defined, then there is no placeholder text. If we were to get rid of this placeholder text then we need to display elsewhere what the current values of these fields are. In terms of the discrepancy between the username and first/last name, I agree that we should be consistent so I would update the username field to behave as the first/last name fields do.

ninavizz:
Is it possible to make text that is a "Current value in data table" vs a "placeholder," black? That would be my only real ask with this—but I don't know if that's possible. Placeholder = gray, existing value previously saved to account = black, is the standard pattern for this that I'd like to see followed as much as can be possible w/o scope-creep.

heartsucker:
> I think what makes sense is that the "placeholder" text here just provides the current value of that field if it's defined. If there is no value defined, then there is no placeholder text.

I think this makes the most sense. @ninavizz, does this work? I can't tell if you're in agreement or not.

ninavizz:
So: I don't want to hold-up merging this PR! I agree with the essense of what Jen's proposing @heartsucker, but it is sooooo confusing when saved data table values are reflected back to the user in gray text. I see that a lot, and it's a bad, highly confusing anti-pattern. 

I'm only just now learning of all the nuanced non-standard patterns employed by the _Admin_ and _Journalist_ experiences, and while I want to ensure they get "corrected," I don't want to encourage wasted cycles on experiences that may not exist in 6-12mos. And tbh, I don't know what it would take to make all saved data table values black, and ONLY placeholder text grey. Which is how it's supposed to be.

securedrop/alembic/versions/b58139cfdc8c_add_checksum_columns_revoke_table.py
redshiftzero: implemented in 541d402f5bc19decaadc0d923a21a2ddeeb33884

creviera:
I think part of the reason this conversation has been so confusing is because of the term "placeholder." Just to make sure everyone understands, a [placeholder](https://www.w3schools.com/TAGS/att_input_placeholder.asp) describes the expected value of an input field. By default, this appears gray and when you click into the empty box is remains visible.

In production, we use it in two different ways: 

**(1)** to describe the expected value, e.g. [showing "Username" in the username textbox of the new user form](https://github.com/creviera/securedrop/blob/develop/securedrop/journalist_templates/admin_add_user.html#L14)

**(2)** to display what is currently stored in the database, e.g. [showing the actual username stored for the current user for the existing user form](https://github.com/creviera/securedrop/blob/develop/securedrop/journalist_templates/edit_account.html#L13)

There are a couple different ways to show default values in the text boxes. You can use the [value](https://www.w3schools.com/tags/att_input_value.asp) attribute or you can use the placeholder attribute as we currently do. They have different behaviors. The value attribute is simple. The text appears black, and you can delete it completely. The placeholder attribute I described above: it appears gray and remains visible when the textbox is empty.

I originally wanted to make it so that when you clicked a textbox the placeholder would disappear, but since we don't use javascript, I came up with [Solution B](https://github.com/freedomofpress/securedrop/pull/4425#issuecomment-494171835) using CSS. This solution also makes it so that the text boxes are empty when there is no database value saved -- this was never an issue, but people have brought it up in comments I noticed.

Hopefully with all the feedback here, @redshiftzero can come to a reasonable solution, whether that be change the username field to act like Solution B or just use value attributes and ditch placeholders. Or something fancier!? :poodle:

ninavizz:
@creviera I've appreciated all the depth and clarity you've provided in this PR, around patterns that are both non-standard and inconsistent across the experience; an experience largely built by volunteer contributors, or a likely overworked dev (or two) wearing too many hats. I mean, the current experience **_is_** a vast improvement upon command-line only stuff, right? 

Your attention to detail in making sure it all gets done right, and asking questions where quick assumptions could otherwise be made, I feel are so important in the product's current growth phase. Your communication has been very clear, to me—I'm just the only kid on the thread w/o any familiarity with how all this is coded. Which was missing context, early in the convo, that you promptly and clearly shed light upon. Thank you for _not_ giving this PR the `shove out the door, go go bye` treatment. I've appreciated it. 😸

redshiftzero:
OK this should be ready for re-review: previous review comments are addressed, the migration and migration tests are moved into a new migration for 0.14.0, and tests are added

To @ninavizz' point any placeholder text (e.g. "First name") is gray, and any db-populated values (e.g. "Alice") are black, thanks @creviera and @ninavizz both for the explanations 💖

redshiftzero:
I'm signing off on Allie's changes here, just want my commits to get a review and then this should be good to merge

rmol:
I ran through the test plan, and everything worked. Input coloring was as @redshiftzero [last described](https://github.com/freedomofpress/securedrop/pull/4425#issuecomment-494602942). I don't see any problems with the code.

Two comments/questions: 
- Should we add labels to the new user form inputs, for consistency and to eliminate some of the problems with the placeholder values? (Did I miss this being considered above?)
- Just gonna lob this little firecracker in: first and last names are a [pretty shaky choice any more](https://shinesolutions.com/2018/01/08/falsehoods-programmers-believe-about-names-with-examples/). It's not too late to go with a single `name` column. It'd be easier to change it now than in a migration after first and last have been deployed. (So glad we're distributed and I'm out of reach. :slightly_smiling_face:)

creviera:
@rmol, ha, that firecracker has gone off a few times already. I'm glad you're bringing it up in written form, and I agree with you. 

I think we should come to a consensus here, so I'll try to capture the arguments I've heard for two names: (1) we want to be able to greet people by a preferred nickname rather than a full name; two fields make it users can provide a short version of how they want to be addressed, and (2) journalists would have an easier time recognizing people's badges by first and last name initials. 

Arguments for one name: (1) see rmol's link: https://shinesolutions.com/2018/01/08/falsehoods-programmers-believe-about-names-with-examples/ (2) it simplifies rules like these: https://github.com/freedomofpress/securedrop/issues/4251#issuecomment-494189922

ninavizz:
@rmol I totally agree with you in spirit; however, the only reason for the two names, is to facilitate the first-initial/last-initial ID badge in Workstation Client. That was what got this ticket going in the first place. That's literally it. Likewise, it is nice to be able to address users by their first name only, in messaging (like "Welcome, Kushal!"). For this reason, most ux and content folks hate the single text field for a users name... despite developers preferring it.

For internationalization, where surnames are used first, we can also just swap the ordering in how the badges are done (if research would validate that—I've never heard of nor done this). Especially for internationalization, though, when folks have multiple names (most countries except for the US); "How you're called socially" and "Your surname" remain the two only identifiers anyone cares about, in the B2B or Enterprise context.

From the user's POV: imagine looking at a conversation thread in a shared client; one source, many colleagues. Would you rather see a single-letter on the ID badge, or two?

image

In enterprise and B2B contexts, two name fields is very standard; namely, because the stiffies in suits in corporate America would freak the f* out if people were tracked by a single db value. Also, legal reasons (like, employer lawsuits are easier when users are more clearly ID'd). This really is a business app, and needs to follow many of those paradigms for user comfort and org compliance—on a whole aside from how we speak to users in UI text.

You're missed in the weekly UX meetings, btw! 😄

redshiftzero:
@rmol thanks for the review! I added a small HTML change to address your comment re: the new user form :rocket:

Thanks for the background/explanation @ninavizz - let's leave as first name / last name :-)

rmol:
> For this reason, most ux and content folks hate the single text field for a users name... despite developers preferring it. [...] "How you're called socially" and "Your surname" remain the two only identifiers anyone cares about

I think many sites that have a single field for the full name also add a "what should we call you" field. That's how I've done it where possible/permitted. It respects people two ways: getting their name exactly right, and asking them how they want to be addressed.

ninavizz:
Yeah, I really love the "What should we call you?" field. A lot. For our purposes with newsroom teams, it's as much about us addressing the authenticated user in system messaging, as it is about surfacing each user's activity to be easily recognized by colleagues as being theirs. 

I have seen the "What should we call you?" field go wrong in workplace situations like this, and it can run the gamut of simply confusing, to really funny. 

Especially once, with a colleague of mine who put "Playful Guy" into that field... which then showed-up in Team Activity streams for others in the company to see. He was really into SF's queer swingers scene, so its inference was not work friendly—but fortunately it was a VERY small consultancy with a bunch of us that had all previously worked together, elsewhere, and we just thought it was hilarious... (he was soooo embarrassed, which admittedly we also all got a big kick out of!) but it has remained with me, since, as the best-ever example of "What should we call you?" going bad. Then again, maybe most professionals wouldn't think to put their S&M monikers, in such a field for work stuff.

PR #4431, opened 2019-05-11
Adds NUC7 hardware recommendation and relegates NUC5s and Mac minis to legacy

### LINKED ISSUE #3826: Review and update hardware recommendations
We provide [hardware recommendations](https://docs.securedrop.org/en/stable/hardware.html#specific-hardware-recommendations) in our docs. There are a couple of problems with the current docs, plus likely more nor yet identified:

- The recommended Gigabyte BRIX (BXi5-5575) is out of stock with most retailers.
- The Intel NUC5i5MYHE is also starting to get long in the tooth; the most current models are 8th gen.
- We don't yet provide a 1U recommendation that would make sense for some larger news organizations with standard server management and procurement practices.

We should carefully review and update these recommendations. This breaks down into a few sub-tasks, e.g.:

- [x] Test 7th generation NUC and attempt to remediate any issues, e.g., with our grsec kernel
- [x] Add warning regarding latest generation Mac Minis (which don't work with Ubuntu without major effort)
- [x] Remove Gigabyte BRIX from recommendations. There's no compelling reason to continue to recommend this model as we already have a small form factor recommendation.
- [ ] Update docs to 7th generation NUCs as our standard recommendation
- [ ] Update docs to include instructions for disabling wireless in the BIOS
- [ ] Procure any desired 1U models to test against, in addition to the ones we already use during QA
- [ ] Based on common hardware choices, internal testing and product lifecycle, add a 1U recommendation

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #3826.
Adds recommendation for NUC7i5BNH servers including information on securing hardware, removes Gigabyte Brix, moves NUC5s and Mac Mini to a **previously recommended** section for legacy support reasons.

## Testing

Docs-only, review for clarity and correctness.

## Deployment

Docs will be deployed by standard process when merged.


docs/hardware.rst
```diff
@@ -279,35 +279,70 @@ We currently recommend the Intel NUC for SecureDrop servers.
...
+to disable wireless functionality:
+
+- before installation of the RAM and storage, disconnect the wireless antennae 
+  leads.
+
+|NUC7 leads|
+
+- before the initial OS installation, boot into the BIOS by pressing **F2** at
+  startup, navigating to **Advanced > Devices > Onboard Devices**, and disabling
+  unwanted hardware - everything except **LAN**.
```
eloquence: Nit: boot into the BIOS by pressing F2, navigating, and disabling -> boot into the BIOS by pressing F2, navigate, and disable.

docs/hardware.rst
```diff
@@ -279,35 +279,70 @@ We currently recommend the Intel NUC for SecureDrop servers.
...
+
+We have tested and can recommend the `NUC7i5BNH <https://www.intel.com/content/www/us/en/products/boards-kits/nuc/kits/nuc7i5bnh.html>`__ - these tend to be readily available in 
+retail stores.
+
+The NUC7i5BNH has soldered-on wireless components, which cannot easily be
+removed. For security reasons, we recommend that you take the following steps
+to disable wireless functionality:
+
+- before installation of the RAM and storage, disconnect the wireless antennae 
+  leads.
```
eloquence: Just unplug? The photo is nice, can we maybe just highlight them as in https://d3nevzfk7ii3be.cloudfront.net/igi/4qIxSBDCk6GkDLfs.medium ?

docs/hardware.rst
```diff
@@ -279,35 +279,70 @@ We currently recommend the Intel NUC for SecureDrop servers.
...
+However, it may be harder to find a retail source for this model.
+
+Previous Server Recommendations
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Intel 5th-gen NUC
+~~~~~~~~~~~~~~~~~
+
+We previously recommended the
+`NUC5i5MYHE <https://www.intel.com/content/www/us/en/products/boards-kits/nuc/kits/nuc5i5myhe.html>`__, however, it is now EOL, and difficult to find. We will continue to support and
```
eloquence: Nit: is now EOL -> has reached end-of-life (just for clarity for folks less familiar with terms like "EOL")

docs/hardware.rst
```diff
@@ -279,35 +279,70 @@ We currently recommend the Intel NUC for SecureDrop servers.
...
+The NUCs typically come as kits, and some assembly is required. You will need to
+purchase the RAM and hard drive separately for each NUC and insert both into the
+NUC before it can be used. We recommend:
+
+-  2x 240GB HDDs (2.5")
+-  1x memory kit of compatible 2x4GB sticks
+   -  You can put one 4GB memory stick in each of the servers.
+
+.. note:: We recommend the use of HDDs over SSDs, as it's more difficult to 
+          securely delete data from SSDs.
```
eloquence: The previous recommendation is to use SSDs. I'm assuming that was performance-motivated? Does the concern re: secure deletion outweigh performance considerations, even in a context like ours, where the most sensitive data is stored only in encrypted form?

docs/hardware.rst
```diff
@@ -317,24 +352,19 @@ before it can be used. We recommend:
...
-.. caution:: We have previously recommended the 2014 Apple Mac Minis (part
-  number MGEM2) for installing SecureDrop. The 2018 Apple Mac Mini (part number
-  MRTR2 or MRTT2) is not a viable candidate for installing SecureDrop, due to
-  hardware support issues. The instructions below apply if you want to
-  (re-)install SecureDrop on the 2014 version.
+2014 Mac Minis
+~~~~~~~~~~~~~~
 
+We previously recommended the 2014 Apple Mac Minis (part number MGEM2)
+for installing SecureDrop. These will soon be officially obsolete. Unfortunately
```
eloquence: Perhaps link "officially obsolete" to https://support.apple.com/en-us/HT201624 to make it clearer what this means in the context of Apple hardware?

eloquence:
Overall looks great, a few small comments inline. Note that I only have the Mac Mini 2014 at hand, so can't verify any of the newly added technical details.

docs/hardware.rst
zenmonkeykstop: The impression I got from recent research was that a) wear leveling makes verifiable deletion from SSDs problematic and b) we'll probably be moving to a reduced number of passes for secure deletion anyway, reducing the time for the system's most significant I/O-bound task.

docs/hardware.rst
zenmonkeykstop: done.

docs/hardware.rst
zenmonkeykstop: done.

docs/hardware.rst
zenmonkeykstop: Done.

docs/hardware.rst
emkll: In both cases, we can't really provide meaningful mitigations against seizure or forensics without at least full disk encryption in place. Furthermore, modern filesystems (Xenial default partion format is ext4) have journaling functionality, would also make some data available post-deletion unless disabled.

As stated above, HDDs provide better guarantees than SSDs for deleting single files (though SSDs might offer better Secure Deletion of the entire drive). We should also keep in mind that we use the `secure_delete` pragma for the (unencrypted) database as well: https://github.com/freedomofpress/securedrop/blob/develop/securedrop/alembic/versions/faac8092c123_enable_security_pragmas.py#L21

zenmonkeykstop: ...but this is something that needs more discussion so I'll revert the change for now.

docs/hardware.rst
rmol: Had the same question, and haven't opened mine up yet -- can the leads just be popped off or do they need to be cut?

zenmonkeykstop: They pop off. I ran them outside the case on mine, but best fix would probably be to stick some electrical tape or heatshrink tubing on the ends to stop them shorting something out. Either way I don't feel it's a level of detail that's worth getting into.

Issue #3959, opened 2018-12-04
Update screenshots on Weblate


redshiftzero:
## Description

One thing we forgot about during the Weblate transition was the updating of the screenshots shown in the "Source Information" panel on the right hand side of each [individual string translation page](https://weblate.securedrop.org/translate/securedrop/securedrop/ro/?type=all). The screenshots are no longer present: 

<img width="465" alt="screen shot 2018-12-04 at 9 53 07 am" src="https://user-images.githubusercontent.com/7832803/49462576-41c99b80-f7ab-11e8-9338-98b427cdd747.png">

We need to investigate and figure out how to update these and do so (one can do it through the admin interface but ideally we'd automate this, note there is an API endpoint for the screenshot). Relevant part of Weblate docs:

https://docs.weblate.org/en/latest/admin/translating.html#screenshots

## User Stories

As a SecureDrop translator, I want a screenshot of the application when translating a particular string so I understand the context.

redshiftzero:
here's the API endpoint used to update the screenshots: https://docs.weblate.org/en/latest/api.html#post--api-screenshots-(int-pk)-file-

eloquence:
I'll take a stab at investigating this further during the 7/23-8/5 sprint, will loop in @rmol before or after his PTO.

eloquence:
Here's how I would approach this:
1. Ensure that screenshots at https://weblate.securedrop.org/screenshots/securedrop/securedrop/ are consistently named in a manner that allows us to infer the layout test that generates them;
2. Create a script that queries all screenshots using the Weblate API and, where a name match can be found against a layout test result in English, uploads the local PNG via API;
3. Test this in the Weblate sandbox;
4. If all looks good, run said script in prod to get us up-to-date;
5. Go through all Weblate strings to make sure correct screenshot associations are in place; if warranted, add additional screenshots.

Does that make sense at a high level? If so, @rmol, I would be happy to collaborate with you on getting this done.

rmol:
Yeah, I think this is a pragmatic approach. I had hoped to find a reliable way to avoid the manual work of associating source strings and screenshots, but I haven't, and given the rate of change, I'm not sure it'd be worth the effort. I think it will be easy enough to keep on top of this as part of the localization manager duties, once we finish this initial push. 

I'd suggest we just clean out the existing screenshots and upload the new set named after the tests. It should at least be possible to write a quick script to list source strings grouped by location in the source code, which would make the initial association work more efficient.

[...]

PR #4451, opened 2019-05-20
Fix Marionette, low-entropy and mock-induced test flakes.

### LINKED ISSUE #4433: Test flakes on application tests
## Description

We're seeing occasional failures on the new TBB test logic.

## Steps to Reproduce

1. Inspect CI run output: https://circleci.com/gh/freedomofpress/securedrop/27610
2. Observe failure

## Expected Behavior
Application tests pass reliably. 

## Actual Behavior
Application tests fail occasionally. 

## Comments

The failing tests (observed so far) are:

* `tests/test_2fa.py::test_totp_reuse_protections`
* `tests/test_2fa.py::test_totp_reuse_protections2`
* `tests/test_2fa.py::test_bad_token_fails_to_verify_on_admin_new_user_two_factor_page`
* `tests/test_2fa.py::test_bad_token_fails_to_verify_on_new_user_two_factor_page`
* `tests/test_journalist_api.py::test_user_cannot_get_an_api_token_with_wrong_2fa_token`

Note that none of those are within the `functional/` subdir, but several do emit webdriver-related errors:

```
        elif exception_class == UnexpectedAlertPresentException:
            alert_text = None
            if 'data' in value:
                alert_text = value['data'].get('text')
            elif 'alert' in value:
                alert_text = value['alert'].get('text')
            raise exception_class(message, screen, stacktrace, alert_text)
>       raise exception_class(message, screen, stacktrace)
E       selenium.common.exceptions.WebDriverException: Message: Socket timeout reading Marionette handshake data: Resource temporarily unavailable (os error 11)
```

Also worth noting that the tests are emitting a warning:

```
test_integration.py::test_filenames
  /usr/local/lib/python3.5/dist-packages/flask/sessions.py:208: UserWarning: "localhost" is not a valid cookie domain, it must contain a ".". Add an entry to your hosts file, for example "localhost.localdomain", and use that instead.
    ' "{rv}.localdomain", and use that instead.'.format(rv=rv)
  /usr/local/lib/python3.5/dist-packages/flask/sessions.py:208: UserWarning: "localhost" is not a valid cookie domain, it must contain a ".". Add an entry to your hosts file, for example "localhost.localdomain", and use that instead.
    ' "{rv}.localdomain", and use that instead.'.format(rv=rv)
```

rmol:
## Status

Ready for review

## Description of Changes

In tests/functional/functional_test.py, FunctionalTest.setup mocked things up, and if an exception were thrown after that in setup, the mocks weren't getting cleaned up. Subsequent two-factor tests were then failing because models.Journalist.verify_token would always return true. This replaces setup/teardown with pytest fixtures and ensures that the mocks are stopped. (Thanks @redshiftzero for finding this one.)

Also retry web driver creation to work around intermittent Marionette problems.

We had redundant session timeout tests, so I removed one.

In a couple of places we were changing models.LOGIN_HARDENING and not resetting it to the original value, so those have been fixed.

Also, because we just set the servers' names to "localhost" in the tests, Flask was warning that it wasn't a valid cookie domain. So I've appended ".localdomain" to shut it up. We have enough noise in our
test output.

In test_alembic, the whitespace pattern '\s*' could be empty, potentially causing trouble for split(). Make it r'\s+'.

Patch get_entropy_estimate in tests/test_integration.py and tests/test_source.py to prevent failures when CI machines run low on entropy. In test_source.py, the call to logger.assert_called_once_with
in test_failed_normalize_timestamps_logs_warning fails if entropy is low in CI -- it's failing not because the message is repeated, but because the mock is called more than once, when the submit route logs
that it can't generate the source's key.

Fixes #4433.

## Testing

Running `make -C securedrop test` should pass, but is not sufficient to verify these changes. Most of the flakes only reliably (heh) happened in CI, due to low entropy or test ordering after exceptions like failures to set up web drivers because of Firefox/geckodriver bugs. (Tests can run in a different order in CI because of how they're divided up and run concurrently.)

A truly diligent test would be to set up CI under your account, create a branch from `rmol/fix-4433-tbb-flakes`, add a trivial change and push to get CI run. 

At this point, having been through that process a lot over the last week, I would say it's optional if the CI for this PR passes.

## Deployment

The changes are all in tests; this should not affect deployment.

### DIFF
--- a/securedrop/tests/pageslayout/test_source.py
+++ b/securedrop/tests/pageslayout/test_source.py
@@ -131,24 +131,3 @@ class TestSourceLayout(
     def test_why_journalist_key(self):
         self._source_why_journalist_key()
         self._screenshot('source-why_journalist_key.png')
-
-
-@pytest.mark.pagelayout
-class TestSourceSessionLayout(
-        functional_test.FunctionalTest,
-        source_navigation_steps.SourceNavigationStepsMixin,
-        journalist_navigation_steps.JournalistNavigationStepsMixin):
-
-    def setup(self):
-        self.session_length_minutes = 0.03
-        super(TestSourceSessionLayout, self).setup(
-            session_expiration=self.session_length_minutes)
-
-    def test_source_session_timeout(self):
-        self._source_visits_source_homepage()
-        self._source_clicks_submit_documents_on_homepage()
-        self._source_continues_to_submit_page()
-        self._source_waits_for_session_to_timeout(self.session_length_minutes)
-        self._source_enters_text_in_message_field()
-        self._source_visits_source_homepage()
-        self._screenshot('source-session_timeout.png')
--- a/securedrop/tests/pageslayout/test_source.py
+++ b/securedrop/tests/pageslayout/test_source.py
@@ -131,3 +131,21 @@ class TestSourceLayout(
     def test_why_journalist_key(self):
         self._source_why_journalist_key()
         self._screenshot('source-why_journalist_key.png')
+
+
+@pytest.mark.pagelayout
+class TestSourceSessionLayout(
+        functional_test.FunctionalTest,
+        source_navigation_steps.SourceNavigationStepsMixin,
+        journalist_navigation_steps.JournalistNavigationStepsMixin):
+
+    session_expiration = 5
+
+    def test_source_session_timeout(self):
+        self._source_visits_source_homepage()
+        self._source_clicks_submit_documents_on_homepage()
+        self._source_continues_to_submit_page()
+        self._source_waits_for_session_to_timeout()
+        self._source_enters_text_in_message_field()
+        self._source_visits_source_homepage()
+        self._screenshot('source-session_timeout.png')
--- a/securedrop/dockerfiles/xenial/python2/Dockerfile
+++ b/securedrop/dockerfiles/xenial/python2/Dockerfile
@@ -30,7 +30,7 @@ RUN curl -LO https://ftp.mozilla.org/pub/firefox/releases/${FF_ESR_VER}/linux-x8
 
 COPY ./tor_project_public.pub /opt/
 
-ENV TBB_VERSION 8.0.8
+ENV TBB_VERSION 8.0.9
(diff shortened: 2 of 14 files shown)


securedrop/tests/pageslayout/test_source.py
```diff
@@ -131,24 +131,3 @@ def test_tor2web_warning(self):
...
-            session_expiration=self.session_length_minutes)
-
-    def test_source_session_timeout(self):
-        self._source_visits_source_homepage()
-        self._source_clicks_submit_documents_on_homepage()
-        self._source_continues_to_submit_page()
-        self._source_waits_for_session_to_timeout(self.session_length_minutes)
-        self._source_enters_text_in_message_field()
-        self._source_visits_source_homepage()
-        self._screenshot('source-session_timeout.png')
```
redshiftzero: So while this is very similar to our functional test, it's worth keeping because the page layout test suite:

1. Is ran on translation merges in all supported locales - if we remove test coverage, if there was breakage in the exported strings/HTML placeholders from Weblate (in this case, the string indicating that the session timed out), we may not catch it for that locale.
2. Also generates screenshots for [updating the SecureDrop user guides](https://docs.securedrop.org/en/release-0.12.2/development/documentation_guidelines.html#updating-screenshots) and for context for translators in Weblate (pending task to upload those screenshots here: #3959)

(I think it is a valid point for us to combine the functional tests and the page layout tests into a single test suite with e.g. a flag for `--screenshot` instead of `--page-layout` and pass in the test locales via env var (and default to English for most developer test runs). We _can_ do that but given all the churn in the tests suites I think we should hold off on this battle for now)

redshiftzero:
otherwise this lgtm (exceptions during functional test setup now shouldn't cause 2FA test failures 🌈 ), just needs that one comment addressing and then this is good to merge

securedrop/tests/pageslayout/test_source.py
rmol: Right. Will save the pruning for a future overhaul. Restored.

PR #4459, opened 2019-05-23
api: expose journalist first and last name if set

redshiftzero:
## Status

Ready for review

## Description of Changes

This is a followup from #4425. In this PR we expose the journalist first name and last name added in  #4425 on:
 - the user endpoint
 - the reply endpoints (since this is the place where we want to use them in the client, for each conversation)

## Testing

Reading the diff and tests should be sufficient, but here are steps to test an endpoint that has the first name / last name added with curl for convenience: 

0. Get a token `curl -X POST -H "Content-Type: application/json" --data '{"username":"journalist","passphrase":"correct horse battery staple profanity oil chewy","one_time_code":"ADDYOURSHERE"}' 127.0.0.1:8081/api/v1/token`

1. Get `curl -X GET -H "Content-Type: application/json" -H "Authorization: Token ADDYOURSHERE" 127.0.0.1:8081/api/v1/user`

```
{
  "first_name": "",
  "is_admin": true,
  "last_login": "2019-05-23T19:07:13.967833Z",
  "last_name": "",
  "username": "journalist",
  "uuid": "e1f5636c-d141-4cac-bc6d-7ed1e6d06dae"
}
```

## Deployment

No special considerations

- [ ] Linting (`make lint`) and tests (`make -C securedrop test`) pass in the development container (I'm relying on CI, only ran the API tests)


rmol:
:+1: Tested these endpoints as well as the reply endpoints, to check the doc changes. All looks good.

securedrop/models.py
```diff
@@ -288,6 +288,8 @@ def to_json(self):
             'filename': self.filename,
             'size': self.size,
             'journalist_username': self.journalist.username,
+            'journalist_first_name': self.journalist.first_name or '',
```
heartsucker: I think these should be `null`/`None` the name isn't set. That seems to more clearly communicate the absence of a value.

securedrop/models.py
```diff
@@ -704,7 +706,9 @@ def to_json(self):
             'username': self.username,
             'last_login': self.last_access.isoformat() + 'Z',
             'is_admin': self.is_admin,
-            'uuid': self.uuid
+            'uuid': self.uuid,
+            'first_name': self.first_name or '',
```
heartsucker: Likewise here.

securedrop/models.py
kushaldas: In Python empty strings are generally preferred to mark unset string values.

heartsucker: Yes but this is a JSON API, not just an internal Python-only data representation. When I get data from an API, I would expect a lack of information to have either a missing key or `null` as the value, not `""`. These are very different things.

redshiftzero: hmm yeah, fair point (I looked at some other JSON APIs, e.g. Github and this is what they are also doing), fixed and repushed

redshiftzero:
hey @kushaldas this needs a restamp (I did make a change above so your review was dismissed)

PR #4461, opened 2019-05-24
Clarified language about localhost setup for local development

### LINKED ISSUE #4460: Documentation of how to use Tor Browser while developing belongs in the development setup page
## Description

In `docs/development/setup_development.rst`, where we describe the addresses of the local source and journalist interfaces, we should also describe how to reach them with the Tor Browser.

We currently explain this in `docs/development/tips_and_tricks.rst`, which is probably only found after contributors realize they can't reach the dev servers with Tor Browser, if at all. 

Let's move those instructions.

## User Research Evidence

Contributors have stumbled across this while trying to change UI elements that only appear when using the Tor Browser. Let's eliminate this obstacle.

drewmassey:
## Status

Ready for review  

## Description of Changes

Fixes #4460 

- Slight text change to make setting up a local deployment with Tor obvious.
## Testing

Non, text only. 

## Deployment

2. New installs.
Negatory.

### DIFF
--- a/docs/development/setup_development.rst
+++ b/docs/development/setup_development.rst
@@ -130,9 +130,7 @@ following ports:
 * Source Interface: `localhost:8080 <http://localhost:8080>`__
 * *Journalist Interface*: `localhost:8081 <http://localhost:8081>`__
 
-If you are using Tor to connect to your local Docker instance, make sure to set 
-"No Proxy" for 0.0.0.0, otherwise Tor will not be able to connect (make sure you 
-understand the security implications for your setup before doing so).
+You should use Tor Browser to test web application changes, `see here for instructions <https://docs.securedrop.org/en/release-0.12.2/development/tips_and_tricks.html#using-tor-browser-with-the-development-environment>`.
 
 A test administrator (``journalist``) and non-admin user (``dellsberg``) are
 created by default when running ``make dev``. In addition, sources and
--- a/docs/development/setup_development.rst
+++ b/docs/development/setup_development.rst
@@ -130,7 +130,8 @@ following ports:
 * Source Interface: `localhost:8080 <http://localhost:8080>`__
 * *Journalist Interface*: `localhost:8081 <http://localhost:8081>`__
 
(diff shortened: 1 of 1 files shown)


docs/development/setup_development.rst
```diff
@@ -130,6 +130,10 @@ following ports:
 * Source Interface: `localhost:8080 <http://localhost:8080>`__
 * *Journalist Interface*: `localhost:8081 <http://localhost:8081>`__
 
+If you are using Tor to connect to your local Docker instance, make sure to set
```
redshiftzero: let's either:

* move the [relevant text](https://docs.securedrop.org/en/release-0.12.2/development/tips_and_tricks.html#using-tor-browser-with-the-development-environment) from the tip page with the heading "Using Tor Browser with the Development Environment" to this page, i.e. in a new section before the `Setting Up a Multi-Machine Environment` section.
* Or just link out to the section https://docs.securedrop.org/en/release-0.12.2/development/tips_and_tricks.html#using-tor-browser-with-the-development-environment with some text like e.g. "You should use Tor Browser to test web application changes, see here for instructions.".

drewmassey: Agreed, having DRY docs is as nice as having DRY code. @redshiftzero I took the second option and applied it here for your review.

docs/development/setup_development.rst
```diff
@@ -130,6 +130,8 @@ following ports:
 * Source Interface: `localhost:8080 <http://localhost:8080>`__
 * *Journalist Interface*: `localhost:8081 <http://localhost:8081>`__
 
+You should use Tor Browser to test web application changes, `see here for instructions <https://docs.securedrop.org/en/release-0.12.2/development/tips_and_tricks.html#using-tor-browser-with-the-development-environment>`.
```
redshiftzero: ah this should be an internal hyperlink target: see e.g. http://docutils.sourceforge.net/docs/user/rst/quickref.html#internal-hyperlink-targets 

otherwise this LGTM! thanks @drewmassey

drewmassey: Pushed a fix; I'm a bit of an rst n00b so want to confirm that this approach doesn't run afoul of the problem here:
https://stackoverflow.com/questions/15394347/adding-a-cross-reference-to-a-subheading-or-anchor-in-another-page

PR #4462, opened 2019-05-24
Resolve Issue 4454

### LINKED ISSUE #4454: Update source interface instructions for Tor Browser 8.5
## Description

Tor Browser 8.5 moved the onion icon to the right of the URL bar:

Screen Shot 2019-05-22 at 3 51 01 PM

## Comments

The point of the bubble used to direct the user to the Tor onion icon. We should move it to point to its new location in TB 8.5.

drewmassey:
## Status

Ready for review 

## Description of Changes

Fixes #4454 .

## Testing

- Spin up local environment
- Confirm that "learn how to set it safest" warning that comes up as described in the issue is on the right, not left side of the screen.

## Deployment

2. New installs.
As mentioned in the Issue thread, this does not try to tell the difference between Tor 8.5 and before.


securedrop/source_templates/index.html
```diff
@@ -92,9 +92,9 @@ <h2 class="welcome-text">
         <div class="row-reverse"> <img id="security-slider-info-close" class="close" src="{{ url_for('static', filename='i/font-awesome/black/times.svg') }}" width="20px" height="20px"></div>
       <p>{{ gettext('You appear to be using the Tor Browser. You can move the Security Slider to Safest in four steps.') }}</p>
       <ol>
-        <li>{{ gettext('Click the <img src="{icon}" alt="Tor icon"> Tor icon in the toolbar above').format(icon=url_for("static", filename="i/toronion.png"))  }}</li>
```
redshiftzero: let's git rm toronion.png if it's not used anywhere else

securedrop/source_templates/index.html
```diff
@@ -92,9 +92,9 @@ <h2 class="welcome-text">
...
-        <li>{{ gettext('Click <strong>Security Settings...</strong>') }}</li>
-        <li>{{ gettext('Move the Slider to <strong>Safest</strong>, then click <strong>OK</strong>') }}</li>
+        <li>{{ gettext('Click the <img src="{icon}" alt="shield icon"> shield icon in the toolbar above').format(icon=url_for("static", filename="i/shield.png"))  }}</li>
```
redshiftzero: we use [AppArmor](https://en.wikipedia.org/wiki/AppArmor) to explicitly whitelist files that the apache process should have (in this case read) access to, this can be done via the following diff which should be part of this change:

```diff
diff --git a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2 b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
index 63282e143..e7d3c7071 100644
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
@@ -264,7 +264,7 @@
   /var/www/securedrop/static/i/font-awesome/refresh-white.png r,
   /var/www/securedrop/static/i/font-awesome/times-white.png r,
   /var/www/securedrop/static/i/font-awesome/trash-black.png r,
-  /var/www/securedrop/static/i/toronion.png r,
+  /var/www/securedrop/static/i/shield.png r,
   /var/www/securedrop/static/i/hand_with_fingerprint.png r,
   /var/www/securedrop/static/i/success_checkmark.png r,
   /var/www/securedrop/static/i/server_upload.png r,
```

securedrop/source_templates/index.html
redshiftzero: (just checked, looks like it isn't)

securedrop/source_templates/index.html
drewmassey: I don't have a lot of experience with AppArmor, but I'm assuming that it isn't enabled on the local development environment? Otherwise I would have expected this gotcha to have tripped me up pre-PR.

redshiftzero: yep AppArmor is only enabled in the [staging and production environments](https://docs.securedrop.org/en/release-0.12.2/development/virtual_environments.html#staging) which we definitely don't expect new contributors to spin up (i.e. totally reasonable not to have known pre-review that apparmor rule needed updating)

PR #4464, opened 2019-05-25
Remove Font Awesome References

### LINKED ISSUE #4450: Strip unnecessary CSS from source interface
The CSS served via the *Source Interface* currently includes the full [Font Awesome Free 5.0.6](https://fontawesome.com/) styling. As the name suggests, Font Awesome is an icon font; the purpose of the CSS styles is to enable you to render its glyphs using syntax like `<i class="fas fa-camera"></i>`. However, the font is not actually used on the *Source Interface* (because it conflicts with Tor browser's behavior in the "Safest" security setting). Instead, we serve up static images rendered from the few glyphs that we use as icons.

The Font Awesome CSS is therefore likely wholly unnecessary. The issue here is to:
- Verify whether any part of the Font Awesome CSS is required for current *Source Interface* styling
- Strip the stylesheet as appropriate.

drewmassey:
## Status

Ready for review 
## Description of Changes

Fixes #4450.

- removes font awesome library
- removes loading references to it in `_base.sass`

## Testing

## Deployment

2. New installs.
_Nope_


securedrop/sass/_base.sass
```diff
@@ -2,7 +2,6 @@
 
 // LIBRARIES
 @import libraries/normalize
-@import libraries/font-awesome-iconography
```
redshiftzero: ah so what we want to do here is move this `font-awesome` import into `journalist.sass`. the reason being is that you are correct we do not use anything in the `font-awesome` library in the source interface but: we _do_ on the journalist interface (127.0.0.1:8081 in the dev env), e.g. these icons are all font awesome icons:

Screen Shot 2019-05-25 at 11 42 47 AM

drewmassey:
Argh, sorry missing that in the code @redshiftzero. 43efa01 should resolve that. Thanks for bearing with me while I get situated.

redshiftzero:
The test failures do seem to be legitimate here - functional tests also fail locally which one can do via the following:

```
$ cd securedrop
$ bin/dev-shell bin/run-test tests/functional/
```

And if you run `make dev` and log into the journalist interface, the font awesome icons don't appear (when you do that note that the globe icon in the language dropdown is a PNG, it's not rendered using font awesome). I think the culprit might be the order of SASS imports?

drewmassey:
@redshiftzero I think this fixes it? The font awesome stuff shows up in journalist and all the tests pass locally.

PR #4475, opened 2019-05-28
Update translations for 0.13.0

rmol:
## Status

Ready for review

## Description of Changes

New translations for 0.13.0 release.

## Testing

For each supported language, check that each of the three strings changed in 0.13.0 has a plausible-looking translation. 

Check out this branch, then run `securedrop/bin/dev-shell ./i18n_tool.py --verbose translate-messages --compile` to create the `.mo` files.

Then save this script as `check_translations.py`:

```python
#!/usr/bin/env python3

import gettext
import os
import sys


SUPPORTED_LANGUAGES = [
    "ar",
    "de_DE",
    "es_ES",
    "en_US",
    "el",
    "fr_FR",
    "it_IT",
    "nb_NO",
    "nl",
    "pt_BR",
    "tr",
    "zh_Hant",
]

TESTS = [
    "Sorry we haven't responded yet!",
    (
        "Our SecureDrop recently experienced a surge of activity. For security "
        "reasons, the creation of a two-way communication channel was delayed until "
        "you checked in again."
    ),
    (
        "Please rest assured that we were able to download your submission, and check "
        "back again later for a reply."
    ),
]


if __name__ == "__main__":
    translation_dir = sys.argv[1]

    for lang in SUPPORTED_LANGUAGES:
        print("Checking {}...".format(lang))
        os.environ["LANGUAGE"] = lang
        gettext.bindtextdomain("messages", translation_dir)
        missing = False
        for t in TESTS:
            translation = gettext.gettext(t)
            print("  {}".format(translation))
            if lang != "en_US" and translation == t:
                missing = True
                print('MISSING TRANSLATION FOR "{}"'.format(t), file=sys.stderr)

        if not missing:
            print("Looks good.")
```

When run with the translations directory as its only argument (e.g. `check_translations.py securedrop/translations`, from the root of your working copy), you should see the three strings translated for each supported language, with no reports of missing translations, and no markup or other suspicious content in the translated strings.

Additionally (edited by @redshiftzero):

- [ ] Ensure there is a passing CI translation job (ran on all pages-layout tests) - you need to do this by triggering a CI run from the shared remote
- [ ] Review the diff manually to verify there are no unusual or unexpected changes in the diff

### If you made changes to the server application code:

- [x] Linting (`make lint`) and tests (`make -C securedrop test`) pass in the development container

PR #4478, opened 2019-05-29
Set up Tor in bin/translation-test

rmol:
Since we currently start both browsers for all functional tests, Tor needs to be running in the test container.

## Status

Ready for review

## Description of Changes

This sets up Tor in the container used for translation tests.

## Testing

Running `make translation-test` on latest develop should produce nothing but errors like `Error creating Tor Browser web driver: SOCKS port 9050 is not listening` because Tor's not running.

Running `make translation-test` on this branch should produce no errors.

## Deployment

This only affects the translation tests.

### DIFF
--- a/securedrop/bin/translation-test
+++ b/securedrop/bin/translation-test
@@ -5,8 +5,11 @@ set -euo pipefail
 
 source "${BASH_SOURCE%/*}/dev-deps"
 
-run_xvfb &
+run_xvfb
+run_tor &
 run_redis &
+setup_vncauth
+run_x11vnc &
 urandom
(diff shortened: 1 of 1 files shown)


redshiftzero:
can you push this commit directly to https://github.com/freedomofpress/securedrop/pull/4475 ? (saves a CI run so we can merge the two together)

redshiftzero:
I've pushed this commit to #4475 so we can merge them together

redshiftzero:
[note the test jobs did not fail here, I cancelled them to free up containers to process the CI queue]

PR #4487, opened 2019-05-29
Add make target to test docs for broken links

### LINKED ISSUE #4410: Check documentation links as part of docs linting
## Description

Sphinx [linkcheck](https://www.sphinx-doc.org/en/master/usage/builders/index.html#sphinx.builders.linkcheck.CheckExternalLinksBuilder) allows the verification of links with the `requests` library to ensure that the links are still valid and active. It might be useful to run this regularly or as part of CI to catch dead or broken links.
## User Stories
As a user, clicking on a link and getting a 404 can be a a frustrating experience.

pierwill:
`make docs-linkcheck` can now be used to test the docs for broken links.

closes #4410

## Status

Ready for review

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,16 +33,20 @@ typelint: install-mypy ## Runs type linting
 ansible-config-lint: ## Runs custom Ansible env linting tasks.
 	molecule verify -s ansible-config
 
+.PHONY: docs
+docs: ## Build project documentation in live reload for editing
+# Spins up livereload environment for editing; blocks.
+	make -C docs/ clean && sphinx-autobuild docs/ docs/_build/html
+
 .PHONY: docs-lint
 docs-lint: ## Check documentation for common syntax errors.
 # The `-W` option converts warnings to errors.
 # The `-n` option enables "nit-picky" mode.
 	make -C docs/ clean && sphinx-build -Wn docs/ docs/_build/html
 
-.PHONY: docs
-docs: ## Build project documentation in live reload for editing
-# Spins up livereload environment for editing; blocks.
-	make -C docs/ clean && sphinx-autobuild docs/ docs/_build/html
+.PHONY: docs-linkcheck
+docs-linkcheck: ## Check documentation for broken links
+	make -C docs/ clean && sphinx-build -b linkcheck -Wn docs/ docs/_build/html
 
 .PHONY: flake8
 flake8: ## Validates PEP8 compliance for Python source files.
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,7 +71,7 @@ jobs:
 
       - run:
           name: Run documentation linting
-          command: make docs-lint
+          command: make docs-lint && make docs-linkcheck
 
   app-tests:
     machine:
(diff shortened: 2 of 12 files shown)


Makefile
```diff
@@ -33,16 +33,20 @@ typelint: install-mypy ## Runs type linting
...
 docs-lint: ## Check documentation for common syntax errors.
 # The `-W` option converts warnings to errors.
 # The `-n` option enables "nit-picky" mode.
 	make -C docs/ clean && sphinx-build -Wn docs/ docs/_build/html
 
-.PHONY: docs
-docs: ## Build project documentation in live reload for editing
-# Spins up livereload environment for editing; blocks.
-	make -C docs/ clean && sphinx-autobuild docs/ docs/_build/html
+.PHONY: docs-linkcheck
```
redshiftzero: can we add a note about this makefile target to our docs on... working on the docs? (meta) 

and let's also run this in CI: you can do this by adding another run step that runs `make docs-linkcheck` in our circle ci config file [here](https://github.com/freedomofpress/securedrop/blob/develop/.circleci/config.yml#L72)

pierwill: Thanks! Force pushed those changes.

pierwill: Is this the right approach?

https://github.com/freedomofpress/securedrop/pull/4487/files#diff-1d37e48f9ceff6d8030570cd36286a61R74

pierwill:
CI failed. Should we treat broken links as errors (the current behavior), or just report them in the CI logs?

redshiftzero:
nope this good - broken links as errors definitely is what we want. however: want to fix the broken link flagged as part of this PR? (looks like there's just one that is actually broken if I'm reading the output correctly, the rest are redirects)

conorsch:
> however: want to fix the broken link flagged as part of this PR? (looks like there's just one that is actually broken if I'm reading the output correctly, the rest are redirects)

Pushed a small change for a broken link to the build logic in the repo. Upon rerunning, seeing an error on the localhost URL to the dev env. We should be able to update that to a codeblock, which will force devs to copy/paste rather than click. Not aware of a `# noqa` pragma for Sphinx linkcheck, which would be ideal.

zenmonkeykstop:
It's possible to skip URLs by adding a directive in the conf.py - setting it up to skip r'http://localhost(:\d+?)/?' gets past that particular URL, but there's a bunch of other failures. Annoyingly the build fails when it hits the first broken link. Will append to the PR once I get a successful build.

conorsch:
@zenmonkeykstop Thanks for jumping in. I, too, am disappointed there's no (apparent) method to omit checking of certain URLs.

> Annoyingly the build fails when it hits the first broken link

Coupled with that problem, I'd almost rather find the URLs ourselves and check them via requests. Once we get to 100% passing, the fast-fail behavior will be less onerous, but without exemptions, I'm not sure we can get this branch passing.

zenmonkeykstop:
Links are fixed and tests are passing - @pierwill if you wanna address @redshiftzero 's review, probably by adding a stanza like 
```
      - run:
          name: Run documentation link check
          command:  make docs-linkcheck
```
as a step in the lint job (as opposed to &&ing commands together) then this should be good to go.

redshiftzero:
this was added in the lint job here: https://github.com/freedomofpress/securedrop/pull/4487/files#diff-1d37e48f9ceff6d8030570cd36286a61R74 which is a fine approach, so approving

PR #4493, opened 2019-05-31
Update i18n.rst, add list-translators to i18n_tool.py

### LINKED ISSUE #4441: Update localization management guide for recent changes to string freeze
## Description

We combined string and feature freeze, so we need to update the documentation here: https://docs.securedrop.org/en/release-0.12.2/development/i18n.html#release-management

We should combine the tasks in the two sections, as well as remove the task "Backport every commit changing a source string to the release branch"

rmol:
## Status

Ready for review

## Description of Changes

Restructured the Internationalization docs (`docs/development/i18n.rst`) to make the order of operations clearer for developers and localization managers. Updated the process now that string and feature freeze are synchronized.

Added a `list-translators` function to `i18n_tool.py` to simplify translator attribution at release time.

Fixes #4441.

## Testing

Run `make docs`, navigate to http://localhost:8000/development/i18n.html and read critically.

Test the `i18n_tools.py` changes with `make -C securedrop list-translators` and `make -C securedrop list-all-translators`.

## Deployment

This is pretty much dev-only and shouldn't affect deployment, as the changes to `i18n_tool.py` didn't involve compilation.


securedrop/i18n_tool.py
```diff
@@ -201,15 +202,15 @@ def require_git_email_name(git_dir):
                'git -C {d} config --get user.email > /dev/null'.format(
                    d=git_dir))
         if subprocess.call(cmd, shell=True):  # nosec
-            if six.u('docker') in io.open('/proc/1/cgroup').read():
```
kushaldas: Any particular reason to remove the usage of `six`?

rmol: Yes. This PR includes changes from https://github.com/freedomofpress/securedrop/pull/4482 where that change is explained. Sorry for not documenting it here too:

> Use of six.u was corrupting translator names. According to
https://pythonhosted.org/six/index.html#six.u -- "On Python 2, u()
doesn't know what the encoding of the literal is. Each byte is
converted directly to the unicode codepoint of the same value. Because
of this, it's only safe to use u() with strings of ASCII data."

> Of course some translators' names were not just ASCII, so I've
replaced usage of six.u with u"", as it should be safe on both Python
2 and Python 3 > 3.3, per PEP-0414.

docs/development/i18n.rst
```diff
@@ -1,437 +1,461 @@
...
-the original file which includes all the translations collected from
-the ``.po`` files.
+* Go to the `Weblate repository page for SecureDrop`_.
 
-This can be done by running the following command::
+|Weblate commit Lock|
 
-    securedrop/bin/dev-shell ./i18n_tool.py --verbose translate-desktop --compile
+* Click ``Commit``.
+* Click ``Push``.
```
kushaldas: @rmol I would love to keep the text from last PR on this (as in develop).

```
* Click ``Push`` (we do these two steps, so that uncommitted changes
  in the repo do not create merge conflicts.)
```

rmol: I moved the explanation before the two steps, just above. Do you want this in addition to that, or do you mean you'd prefer that it just be explained here?

PR #4505, opened 2019-06-04
Add workaround for CircleCI's problems filtering from forks

### LINKED ISSUE #4479: CI branch filtering logic doesn't work from forks
## Description

We heavily use Circle's branch filtering logic: https://github.com/freedomofpress/securedrop/blob/develop/.circleci/config.yml#L355

However the logic doesn't seem to be working for PRs from forks: https://github.com/freedomofpress/securedrop/pull/4475

We should revise our branch filtering rules such that the correct CI jobs run from forks (e.g. `i18n` anywhere in the branch name triggers the translation test job)

rmol:
## Status

Ready for review

## Description of Changes

CircleCI's branch filtering does not work properly with pull requests
from forks. The `CIRCLE_BRANCH` variable contains something like `/pull/3`
in this case. This means that docs and i18n PRs from forks are not
being tested as we wish; the translation tests are not run for i18n
PRs, and the app tests are being run for docs PRs.

A CircleCI [feature request](https://discuss.circleci.com/t/only-build-pull-requests-targetting-specific-branch/6082/6) to improve this was closed without
explanation, so I'm incorporating a workaround suggested in the
CircleCI forums: using the GitHub API to obtain the real branch name
for PRs from forks, and skipping tests if it doesn't match. Not all
steps of the relevant jobs are skipped, but the most expensive ones
are.

Also, stop skipping static-analysis-and-no-known-cves, as it doesn't
take that long, and might prevent problems from sneaking in on
branches with inaccurate names.

Fixes #4479.

## Testing

We're going to test this with two PRs, one for docs and one for i18n, to verify that the right tests run and pass.

## Deployment

This will only affect CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,6 +98,9 @@ jobs:
           name: Run tests
           no_output_timeout: 20m
           command: |
+            BRANCH_MATCH=$(devops/scripts/match-ci-branch.sh "^(i18n|docs|update-builder)")
+            echo "match-ci-branch.sh said: ${BRANCH_MATCH}"
+            if [[ $BRANCH_MATCH =~ ^found ]]; then echo "Skipping: ${BRANCH_MATCH}"; exit 0; fi
             export TESTFILES=$(cd securedrop; circleci tests glob 'tests/test*py' 'tests/**/test*py' |circleci tests split --split-by=timings |xargs echo)
             docker rm -f securedrop-test-xenial-py2 || true
             fromtag=$(docker images |grep securedrop-test-xenial-py2 |head -n1 |awk '{print $2}')
@@ -155,6 +158,9 @@ jobs:
       - run:
           name: Run tests
           command: |
+            BRANCH_MATCH=$(devops/scripts/match-ci-branch.sh "^(i18n|docs|update-builder)")
+            echo "match-ci-branch.sh said: ${BRANCH_MATCH}"
+            if [[ $BRANCH_MATCH =~ ^found ]]; then echo "Skipping: ${BRANCH_MATCH}"; exit 0; fi
             export TESTFILES=$(cd securedrop; circleci tests glob 'tests/test*py' 'tests/**/test*py' |circleci tests split --split-by=timings |xargs echo)
             docker rm -f securedrop-test-xenial-py3 || true
             fromtag=$(docker images |grep securedrop-test-xenial-py3 |head -n1 |awk '{print $2}')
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -98,6 +98,11 @@ jobs:
           name: Run tests
           no_output_timeout: 20m
           command: |
+            CIRCLE_PR_BRANCH=$([[ -n ${CIRCLE_PR_NUMBER} ]] && (curl -s https://api.github.com/repos/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}/pulls/${CIRCLE_PR_NUMBER} | jq -r '.head.ref') || "")
+            echo "CIRCLE_PR_BRANCH: ${CIRCLE_PR_BRANCH}"
+            if [[ $CIRCLE_BRANCH =~ ^i18n || $CIRCLE_PR_BRANCH =~ ^i18n ]]; then echo "Skipping: i18n branch"; exit 0; fi
+            if [[ $CIRCLE_BRANCH =~ ^docs || $CIRCLE_PR_BRANCH =~ ^docs ]]; then echo "Skipping: docs branch"; exit 0; fi
+            if [[ $CIRCLE_BRANCH =~ ^update-builder || $CIRCLE_PR_BRANCH =~ ^update-builder ]]; then echo "Skipping: update-builder branch"; exit 0; fi
```
redshiftzero: hmm I'm wondering if we should add a helper script in `devops/` for this logic for DRYness reasons/readability? what do you think? (it would take a list of prefixes of branches to skip) then we can call out to it in these run steps

rmol: Yeah, good call. Will do.

.circleci/config.yml
```diff
@@ -292,7 +320,13 @@ jobs:
       - run: apt-get install -y make virtualenv python-pip enchant
```
redshiftzero: we'll need to install jq here too no?

.circleci/config.yml
```diff
@@ -292,7 +320,13 @@ jobs:
...
       - checkout
       - setup_remote_docker
-      - run: make ci-deb-tests
+      - run:
+          name: Test Debian package build
+          command: |
+            BRANCH_MATCH=$(devops/scripts/match-ci-branch.sh "^update-builder")
+            echo "match-ci-branch.sh said: ${BRANCH_MATCH}"
+            if [[ $BRANCH_MATCH =~ ^found ]]; then echo "Skipping: ${BRANCH_MATCH}"; exit 0; fi
```
redshiftzero: `if [[ $BRANCH_MATCH =~ ^found ]]` -> `if ! [[ $BRANCH_MATCH =~ ^found ]]` here since we only want to skip the test when we _don't_ find `update-builder` in the branch name

(btw: the `deb-tests` job isn't getting kicked off when testing from a fork since we're using circle ci's branch filtering on `/update-builder-.*/`. I think this makes sense since in practice anyone who is doing builder updates has write access to the repo.)

Issue #4355, opened 2019-04-18
Create securedrop-app-code Debian package with Python 3 dependencies


eloquence:
Now that we have basic Python 3 compatibility (#997, resolved through #4239), we need to ensure that all the `securedrop-app-code` dependencies transition to Python 3 as well. The plan of record is to use `dh-virtualenv` to ship these dependencies in the package itself, instead of depending on the system Python libraries. 

This will allow us to use consistent dependency management processes for SecureDrop Workstation and SecureDrop core, and will get us closer to resolving #3407.

eloquence:
For the 4/17-5/1 sprint, Kushal will prepare at least a draft PR for broader team input, which will not use FPF's PyPI mirror yet.

redshiftzero:
hey @kushaldas what is the status of this?

redshiftzero:
can you push up whatever wip you have when you get a chance @kushaldas ?

kushaldas:
I did not manage to work on this enough during that sprint, but, I have other WIP work on my backup system (a `make clean` removed from my laptop during the last trip). I will push it here with details of things needs to be done.

redshiftzero:
please push up whatever wip you have locally so @rmol can check it out since y'all will be working together next sprint to get the combined `mod_wsgi` changes and this issue merged together.

kushaldas:
I pushed https://github.com/freedomofpress/securedrop/pull/4555 to show the changes I made for a debian python3 package using `dh-virtualenv`

Issue #3376, opened 2018-05-07
cron-apt update logic only allows addition of apt dependencies from /etc/apt/security.list 


redshiftzero:
## Description

During QA of 0.7.0, in #3316 we discovered that the `securedrop-app-code` package was not upgrading. It was held back due to the addition of the `libjpeg-dev` apt dependency in #3057. [Further spelunking by @msheiny indicated](https://github.com/freedomofpress/securedrop/issues/3316#issuecomment-387181852) that we can only add dependencies in repos listed in `/etc/apt/security.list`, or existing instances will not update. This means that currently we can't add any apt dependencies in repos not on those lists. 

## User Stories

As a SecureDrop developer, I want to be able to add new apt dependencies to the `securedrop-app-code` package so I can ship new features.

redshiftzero:
My understanding is that the reason to restrict upgrades only to repositories in `/etc/apt/security.list` was such that critical security upgrades are applied promptly even though this means if a security upgrade causes some other issue, it may break something (a good decision given the nature of the project and the relative risk). Presumably this is also why e.g. `unattended-upgrades` behaves in the same way in its default configuration - i.e. only doing security updates automatically. 

If we permit upgrades for more packages, without doing something like #2106 (proxying all apt packages and pulling the lever to send them to instances e.g. when CI passes) it increases the probability of a production failure, as more untested (untested with SecureDrop) code will be running on instances. Without further investigation, it's unclear to me how much the chance of a failure increases, and thus whether or not this is an acceptable risk. My pause here to ask this question is because both outages in the history of the SecureDrop project have been due to untested updates (#2105, #1364). If we estimate the risk of such failures occurring to be low, then I'd be in favor of permitting upgrades for all installed packages. 

Perhaps a lower risk option that still enables us to add apt dependencies in the future, while not opening the door to all updates, might be an idea @msheiny had - to host those packages on FPF's repository. This does introduces some maintenance burden and as such we should only introduce dependencies when we actually need them, as we need to monitor them for updates and update them after CI passes, as we now do with Tor. It is certainly a lesser burden than both proxying all apt updates, and also better than having to recover from an outage.

msheiny:
> If we permit upgrades for more packages, without doing something like #2106 (proxying all apt packages and pulling the lever to send them to instances e.g. when CI passes) it increases the probability of a production failure, as more untested (untested with SecureDrop) code will be running on instances. 

[...]

PR #4518, opened 2019-06-10
Bundle mod_wsgi in securedrop-app-code package

### LINKED ISSUE #4292: Install libapache2-mod-wsgi-py3, which can only be installed if libapache2-mod-wsgi is not there
## Description

For Python3 we will have to install `libapache2-mod-wsgi-py3` which can only be installed if `libapache2-mod-wsgi` is not there. Means, we can not just autoupgrade to Python3 easily,

## User Stories

I am SecureDrop admin, and I want to move to the latest Python3 based SecureDrop smoothly without any userinput.

rmol:
# Status

Ready for review

## Description of Changes

To move to Python 3, we need to remove the dependency on Ubuntu's
libapache2-mod-wsgi package. Since the libapache2-mod-wsgi-py3 package
is in the universe section, this change bundles mod_wsgi in our
securedrop-app-code package.

For now, we have to simply drop libapache2-mod-wsgi as a dependency of
securedrop-app-code. Actually uninstalling it disables the Apache
module, breaking the Apache configuration. So we're leaving it
installed and modifying /etc/apache2/mods-available/wsgi.load to point
to our bundled mod_wsgi. This should be safe even if
libapache2-mod-wsgi is updated, as we're specifying --force-confdef
and --force-confold for cron-apt upgrades.

At some point when we establish more control over the state of /etc/
on SecureDrop servers and can ensure our bundled module stays
configured, we'll be able to remove the libapache2-mod-wsgi package.

While testing these changes in prod VMC, I ran into difficulties with
direct SSH, so I tweaked the Vagrantfile to support that better.

Fixes #4292.

## Testing

[ ] Check out this branch.
[ ] Run `make build-debs` and confirm only the apt up-to-date test fails.

### Clean install

[ ] Run `make staging` and confirm no errors during installation.
[ ] Browse to the Source Interface and confirm that the footer says `Powered by SecureDrop 0.14.0~rc1`.
[ ] Proceed with functional interactive testing of the Source/Journalist interfaces.

### Upgrade behavior

The "upgrade" scenario requires use of libvirt for VM driver.

[ ] Run `make upgrade-start` to provision local apt VMs.
[ ] Visit the source interface (the onion address is at the end of `make upgrade-start` output) and confirm that the footer says `Powered by SecureDrop 0.13.0` (we haven't upgraded yet).
[ ] Run `make upgrade-test-local`.
[ ] Browse to the Source Interface and confirm that the footer says `Powered by SecureDrop 0.14.0~rc1`.
[ ] Proceed with functional interactive testing of the Source/Journalist Interfaces.

## Deployment

Switching to Python 3 is kind of a big deal.


build/.gitignore
```diff
@@ -1,4 +0,0 @@
-*.deb
-trusty/*.deb
-xenial/*.deb
-*.tar.gz
```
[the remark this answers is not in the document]
rmol: That was ... an accident after cleaning build manually. But good point, I'll create `.empty`.

install_files/securedrop-app-code/debian/postinst
```diff
@@ -142,6 +142,9 @@ case "$1" in
     # the ability to send signals to unconfined peers.
     service apache2 stop
 
+    # Point Apache to our bundled mod_wsgi
+    mod_wsgi-express module-config > /etc/apache2/mods-available/wsgi.load
```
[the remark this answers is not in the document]
rmol: Will do.

build/.gitignore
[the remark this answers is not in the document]
rmol: The .gitignore wasn't completely redundant, so I just restored it.

rmol:
@conorsch @kushaldas Minimally invasive version ready for (re)consideration. It just uses Python 3 when building `securedrop-app-code`; the rest of the changes can be done later and elsewhere.

Makefile
```diff
@@ -103,16 +103,16 @@ safety: ## Runs `safety check` to check python dependencies for vulnerabilities
...
 	make -C admin update-pip-requirements
-	pip-compile --output-file securedrop/requirements/develop-requirements.txt \
+	pip-compile --generate-hashes --output-file securedrop/requirements/develop-requirements.txt \
 		admin/requirements-ansible.in \
 		securedrop/requirements/develop-requirements.in
-	pip-compile --output-file securedrop/requirements/test-requirements.txt \
+	pip-compile --generate-hashes --output-file securedrop/requirements/test-requirements.txt \
 		securedrop/requirements/test-requirements.in
 	pip-compile --generate-hashes --output-file securedrop/requirements/securedrop-app-code-requirements.txt \
 		securedrop/requirements/securedrop-app-code-requirements.in
```
conorsch: When generating the reqs files for the Application Server, both test and prod reqs, we must take care to use a Python3 virtualenv, otherwise we'll wind up with Py2 deps from the dev env's `pip-compile`. Consider wrapping just the app code calls to pip-compile in a container; see example here https://github.com/freedomofpress/securedrop.org/blob/d9388f62220e45f5fdd6d50ec2652680cb48805c/Makefile#L43-L51 . (Likely these changes were previously present in the diff, and dropped after discussion of punting on the dev env migration.)

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/build_securedrop_app_code_deb.yml
```diff
@@ -16,9 +16,21 @@
...
 - include: sass.yml
 
+- name: Install Debian packages required to build wheels
+  apt:
+    name:
+      - apache2-dev
+      - python3-dev
+      - python3-pip
+      - python3-wheel
+    state: present
```
conorsch: These py3 dependencies are _in addition_ to the py2 dependencies declared in `molecule/builder-xenial/Dockerfile`. We should instead modify the dependencies in the Dockerfile for the build, push a new image, and store the hash as part of this PR. Happy to append to this PR if you agree. Storing the dependencies in the container image definition will help keep local builds snappy, leveraging the cached layers.

install_files/securedrop-app-code/debian/control
```diff
@@ -6,5 +6,5 @@ Homepage: https://securedrop.org
 
 Package: securedrop-app-code
 Architecture: amd64
-Depends: python-pip,apparmor-utils,gnupg2,haveged,python,secure-delete,sqlite3,${dist:Depends},libapache2-mod-wsgi,libapache2-mod-xsendfile,redis-server,supervisor,securedrop-keyring,securedrop-config,libpython2.7-dev
+Depends: python3-pip,apparmor-utils,gnupg2,haveged,python3,secure-delete,sqlite3,${dist:Depends},libapache2-mod-xsendfile,redis-server,supervisor,securedrop-keyring,securedrop-config,devscripts
```
conorsch: In order to avoid upstream updates to `libapache2-mod-wsgi` breaking our customizations to the `/etc/apache2/mods-available/wsgi.load`, we should pin the current-latest version of `libapache2-mod-wsgi` in the dependencies. This is a temporary measure—we still intend to remove that package later.

Makefile
rmol: Done. Slightly differently than your example, as I just used our dev shell with Python 3.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/build_securedrop_app_code_deb.yml
rmol: I've left this in for now, but as soon as we can get the image updated I'll remove this step.

install_files/securedrop-app-code/debian/control
rmol: Pinned. I think I got the version specification correct, as securedrop-app-code is happily installing, but please double check.

rmol:
@conorsch Test plan updated; please review, and thanks for the help.

rmol:
@kushaldas This latest round of fixes included a change to how we collect the unhashed requirements when building the `securedrop-app-code` Debian package, to avoid pulling in dependencies that are not declared in its control file. It should have fixed the problem you saw, so if you've got the time to try it again, I'd appreciate it.

conorsch:
@rmol Test plan still appears out of date to me. Suggestions:

## Testing
### Clean install
* [ ] Run `make build-debs` and confirm _only_ the apt up-to-date test fails
* [ ] Run `make staging` and confirm no errors during installation
* [ ] Browse to the Source Interface and confirm no obvious errors
* [ ] Proceed with functional interactive testing of the Source/Journalist interfaces

### Upgrade behavior
The "upgrade" scenario requires use of libvirt for VM driver. 

* [ ] Run `make build-debs` and confirm _only_ the apt up-to-date test fails
* [ ] Run `make upgrade-start` to provision local apt VMs
* [ ] Log into app machine and run `sudo cron-apt -i -s`, confirm no errors
* [ ] Proceed with functional interactive testing of the Source/Journalist Interfaces

Ideally we should modify `molecule/upgrade/side_effect.yml` to use `cron-apt -i -s` rather than an apt upgrade, to simulate more accurately the upgrade behavior prod instances will see. That's fine to do in this PR. 

The test plan should _not_ include use of a py3 virtualenv, based on visual review of the diff. Right now, I'm seeing CI errors that may be related to CI running on your fork, rather than in the main repo. Building the debs locally, I see test errors that aren't the apt up-to-date test, relating to `Flask-Babel` from the requirements files.

Although it's not been discussed as in the scope of #4292 to date, even once we get this PR merged, we'll still have the supervisord process running via `python2`, which isn't good. That's best handled in a separate issue, though—perhaps we can drop supervisor altogether (installing from apt means we're lagging behind the py3 support upstream) and use systemd to keep the service up.

Will take another pass through this with functional testing tomorrow.

rmol:
@conorsch I don't know what happened to my edit. I think I was still previewing, went to do something else, came back, it looked good, and I never hit the update button. :roll_eyes: Anyway, it looked much like yours except I didn't realize `make upgrade-test-local` wasn't already using cron-apt.

Turns out trying cron-apt was a _great_ suggestion, because we can't upgrade `securedrop-app-code` with our current cron-apt configuration. Dependencies like `python3-pip` aren't available in the repos in `/etc/apt/security.list`, so `securedrop-app-code` just gets skipped. Historically, pip's been installed as part of `securedrop-admin install`, so it's always been available.

@emkll had the (reluctant) idea of adding a cron-apt action to explicitly install `securedrop-app-code`. This works, but it does take two rounds of cron-apt: one to insert the action, and another to actually run it, installing `securedrop-app-code` and its dependencies with the regular apt cache. And we're left with the non-security action. We could remove this hack in a point release, but the whole thing feels shaky.

We're hoping you're horrified enough to think of something better. I've pushed a couple of commits for review; with any luck we can rebase-erase them. :slightly_smiling_face:

install_files/securedrop-app-code/debian/control
```diff
@@ -6,5 +6,5 @@ Homepage: https://securedrop.org
 
 Package: securedrop-app-code
 Architecture: amd64
-Depends: python-pip,apparmor-utils,gnupg2,haveged,python,secure-delete,sqlite3,${dist:Depends},libapache2-mod-wsgi,libapache2-mod-xsendfile,redis-server,supervisor,securedrop-keyring,securedrop-config,libpython2.7-dev
+Depends: ${dist:Depends}, apparmor-utils, gnupg2, haveged, libapache2-mod-xsendfile, libapache2-mod-wsgi (= 4.3.0-1.1build1), python3, python3-pip, redis-server, secure-delete, securedrop-config, securedrop-keyring, sqlite3, supervisor
```
redshiftzero: so if we deployed this change using #4355 then we wouldn't need to install `python3-pip` (problematic as it's from the universe channel) right? we wouldn't need to pip install the wheels because we'd ship a virtualenv cc @kushaldas @rmol @conorsch

rmol: I've only read the [dh-virtualenv docs](https://dh-virtualenv.readthedocs.io/en/latest/index.html) but yes, it looks like that would be a better way to get this done.

eloquence:
Per discussions on Friday and today, this change won't land in 0.14.0, in significant part due to the problems updating dependencies not in `/etc/apt/security.list` (#3376) and our desire to avoid universe channel dependencies. Instead, the revised goal is to ship this change using `dh-virtualenv` (bypassing the need to install `pip` from the universe channel), and to land it with 0.15.0. 

During the current sprint, we'll want to continue to explore _how_ this change could be implemented using `dh-virtualenv`, pending a team review of the current WIP implementation of `dh-virtualenv` usage by @kushaldas.

install_files/securedrop-app-code/debian/control
eloquence: Once we switch to `dh-virtualenv`, will this change still require any other package dependency updates that conflict with the current cron-apt update logic (#3376)?

rmol:
Closing in favor of #4622.

Issue #4358, opened 2019-04-19
Ansible commands sometimes fail with "waiting for privilege escalation prompt"


eloquence:
When you run a command like `securedrop-admin logs` -- a fairly simple playbook -- the playbook may sometimes fail with the error "Timeout (62s) waiting for privilege escalating prompt:" on both `app` and `mon`, even though both servers are reachable via `ssh app` and `ssh mon` from the same workstation.

Removing the `~/.ansible` directory on the *Admin Workstation* resolves and ensures successful playbook run, so this appears to be related to Ansible control-path files as previously described in #1957.

Unfortunately no clear steps to reproduce yet; I've encountered it once today after running the logs command twice, with no significant difference in the two runs, so it may be related to Tor network issues as well.

PR #4522, opened 2019-06-13
Added workstation rebuild docs

### LINKED ISSUE #4266: [docs] Document from-scratch reprovisioning of Admin Workstation
If an admin does not have a working copy of their Admin Workstation, they may need to manually re-provision it from scratch -- using either a partially working copy, or only the data on their server. We have documented the process for updating *old* Admin Workstations in various update guides, but we have not documented the process for creating a new Admin Workstation for an existing installation.

# User Story

As an admin, I don't want to have to call FPF if my admin workstation is borked, so I can solve problems quickly on my own.

zenmonkeykstop:
## Status
Ready for review

## Description of Changes

Fixes #4266 
Adds documentation describing how to rebuild an Admin Workstation USB given physical access to the SecureDrop servers. Adds a new `securedrop-admin` action, `restrict_admin_access`, that locks down SSH access to the active Admin Workstation.

## Testing

- [ ] Follow the documented procedure and verify that it produces a working Admin Workstation and does not change the onion URLs, database or GPG keys associated with the instance.

- [ ] Set up a production instance and test the `./securedrop-admin reset_admin_access` target:
  - [ ] After completing installation and `tailsconfig`, run `./securedrop-admin reset_admin_access` and confirm that it leaves the file `~/.ssh/authorized_keys` on the the *Application* and *Monitor Servers* unchanged, with only the *Admin Workstation's* SSH pubkey present
  - [ ] Manually append multiple SSH pubkeys to the `~/.ssh/authorized_keys` file on the *Application* and *Monitor Servers*, run `./securedrop-admin test_admin_access`, and confirm that it updates the file `~/.ssh/authorized_keys` on the the *Application* and *Monitor Servers* to have only the *Admin Workstation's* SSH pubkey present
  - [ ] On the *Admin Workstation*, rename `~/.ssh/id_rsa.pub` to `~/.ssh/id_rsa.hideme`, run `./securedrop-admin test_admin_access`, and confirm that it exits with an error and without updating the file `~/.ssh/authorized_keys` on the the *Application* and *Monitor Servers*.

## Deployment
No special requirements

### DIFF
--- a/docs/rebuild_admin.rst
+++ b/docs/rebuild_admin.rst
@@ -21,9 +21,15 @@ rebuilding an *Admin Workstation* is as follows:
  #. Set up SSH access for the new *Admin Workstation*.
  #. Retrieve SecureDrop configuration settings from the  *Application* and *Monitor Server*.
  #. Back up and configure the SecureDrop application.
- #. Run the ``./securedrop-admin install`` and ``./securedrop-admintailsconfig``
+ #. Run the ``./securedrop-admin tailsconfig`` and ``./securedrop-admin install``
     commands from the new *Admin Workstation*.
  #. Complete post-rebuild tasks.
+
+
+.. important:: The rebuild process involves temporarily removing ``iptables`` 
+               rules on the *Application* and *Monitor Servers*, weakening their
+               security. Because of this, it's important to complete the rebuild
+               process promptly, to avoid leaving the servers in an insecure state.
 
 
 Step 1: Prepare the USB sticks
@@ -376,9 +382,14 @@ Once complete, configure the *Admin Workstation* using the command:
 This will set up desktop shortcuts for the *Source* and *Journalist Interfaces*,
 and configure adminstrative access to the servers.
 
-Once the ``./securedrop-admin tailsconfig`` command is complete, verify that you can connect to 
-the servers using ``ssh app`` and ``ssh mon``, accepting the host verification
-prompt if necessary.
+Once the ``./securedrop-admin tailsconfig`` command is complete: 
+ - verify that the ``Hostname`` references in ``~/.ssh/config`` have been updated 
+   to refer to Onion URLs instead of direct IP addresses,
+ - verify that you can connect to 
+   the servers using ``ssh app`` and ``ssh mon``, accepting the host verification
+   prompt if necessary,
+ - and verify that the desktop shortcuts for the *Source* and *Journalist Interfaces*
+   work correctly, opening their respective homepages in the Tor Browser.
 
 Next, back up the servers by running the following command in the terminal:
 
@@ -390,8 +401,8 @@ Next, back up the servers by running the following command in the terminal:
 Step 6: Run the ``./securedrop-admin install`` command
 ======================================================
 
-After the ``./securedrop-admin backup`` command completes successfully, you should  undo the
-changes made to enable temporary local SSH access, by running the following 
+After the ``./securedrop-admin backup`` command completes successfully, you should
+undo the changes made to enable temporary local SSH access, by running the following 
 command:
 
 .. code:: sh
@@ -419,5 +430,22 @@ We recommend completing the following tasks after the rebuild:
  - Back up your *Admin Workstation* using the process 
    :ref:`documented here <backup_workstations>`.
  - Delete invalid admin accounts in the *Journalist Interface*.
- - Delete invalid SSH public keys from the administration shell account on the
-   *Application* and *Monitor Servers*.
+ - Restrict SSH access to the *Application* and *Monitor Servers* to valid 
+   *Admin Workstions*. If your new *Admin Workstation* USB stick
+   is the only one that should have SSH access to the servers, you can remove 
+   access for any previous *Admin Workstations* from the terminal,  using the 
+   commands:
+
+   .. code:: sh
+
+     cd ~/Persistent/securedrop
+     ./securedrop-admin reset_admin_access
+
+   You can also selectively remove invalid keys by logging on to the *Application*
+   and *Monitor Servers* and editing the file ``~/.ssh/authorized_keys``, making 
+   sure not to remove the public key belonging to your new *Admin Workstation*.
+ - Optionally, set up :ref:`daily journalist alerts <daily_journalist_alerts>`, 
+   by running ``./securedrop-admin sdconfig`` and providing a valid
+   GPG key and fingerprint, along with the corresponding destination email address, then
+   running ``./securedrop-admin install`` again to update the server configuration.
+     
--- a/docs/rebuild_admin.rst
+++ b/docs/rebuild_admin.rst
@@ -1,0 +1,423 @@
+Rebuilding an *Admin Workstation* USB
+-------------------------------------
+
+In cases where an *Admin Workstation* USB stick has been lost or destroyed, and no
+backup exists, it is possible to rebuild one. In order to do so, you'll need
+ 
+ - physical access to the SecureDrop servers
+ - 2 USB sticks:
+
+   - Tails Master USB
+   - 1 replacement *Admin Workstation* USB (USB3 and 16GB or better recommended)
+
+The process requires experience with the Linux command line and Tails, and 
+can take up to 3 hours. If a backup of the SecureDrop application server is available,
+reinstalling the instance may be simpler. An outline of the steps involved in
+rebuilding an *Admin Workstation* is as follows:
+
+ #. Prepare the USB sticks.
+ #. (Optional) Boot the *Application* and *Monitor Server* in single user mode and  reset
+    the shell admin account password.
+ #. Set up SSH access for the new *Admin Workstation*.
+ #. Retrieve SecureDrop configuration settings from the  *Application* and *Monitor Server*.
+ #. Back up and configure the SecureDrop application.
+ #. Run the ``./securedrop-admin install`` and ``./securedrop-admintailsconfig``
+    commands from the new *Admin Workstation*.
+ #. Complete post-rebuild tasks.
+
+
+Step 1: Prepare the USB sticks
+==============================
+
+First, :ref:`create a new Admin Workstation USB <setup_install_tails>` 
+and set up a persistent volume with a strong passphrase.
+
+Once persistence has been set up, start up the *Admin Workstation* with 
+persistence enabled, :ref:`install the SecureDrop application code, and set up
+the KeePassX database <set_up_admin_tails>`.
+
+The *Admin Workstation* uses SSH with key authentication to connect to the servers,
+so you'll need to create a new SSH keypair for your SecureDrop instance. To do so,
+open a terminal by navigating to **Applications>Favorites>Terminal**,  and run 
+the following command:
+
+.. code:: sh
+
+ ssh-keygen -t rsa -b 4096
+
+When prompted to ``Enter file in which to save the key``, Press **Enter** to use
+the default location. When prompted for a passphrase, it's safe to leave it blank.
+
+
+Step 2: (Optional) Boot the servers in single-user mode
+=======================================================
+If you do not have the original password for the shell admin account on the
+*Application* and *Monitor Servers*, you'll need to reset the password on each
+server by booting in single user mode. In order to do so, you'll need physical
+access to the server, a keyboard, and a monitor. 
+
+First, connect a monitor and keyboard to the *Monitor Server*. Then reboot the server.
+When the GRUB menu appears, make sure the **Ubuntu** entry is highlighted, and 
+press **e** to edit boot options.
+In the boot options for Ubuntu, find the line that starts with ``linux`` and ends
+with ``ro``. Add ``single`` after ``ro``, separated by a space, and press
+**F10** to boot in single user mode.
+
+Reset the SecureDrop admin user's password
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Once the root prompt appears, you'll need to reset the password for the
+SecureDrop admin user. By default this user is named `sdadmin` and has UID 1000.
+However it may have been named differently during the installation of your 
+instance. You can use the command ``getent passwd 1000`` to check the username
+corresponding to UID 1000. Once you have the correct username, reset its password
+using the `passwd` command, for example:
+
+.. code:: sh 
+
+ passwd sdadmin
+
+.. important::
+ Make sure to select a strong password, and record it in the *Admin Workstation's*
+ KeePassX database.
+
+Finally, reboot the *Monitor Server* and verify that you can log in at the console
+using the new password.
+
+Repeat the process for the *Application Server*. Use the same username and 
+password as for the *Monitor Server* - this is required in order for the 
+``./securedrop-admin install`` command to work correctly.
+
+Step 3: Set up *Admin Workstation* access
+=========================================
+Next, you'll configure the servers to allow temporary SSH access from the new *Admin
+Workstation*.
+
+First, start the new *Admin Workstation* with persistence enabled and an administration
+password set.
+
+Next, connect the new *Admin Workstation* to the *Hardware Firewall* via the
+appropriate Ethernet port, and set up its static IP address. For more information
+on how to do so, see :ref:`this section in the firewall setup documentation
+<assign_static_ip_to_workstation>`. If you do not know the correct static IP
+address for the *Admin Workstation*, and you are using a recommended pfSense-based
+*Hardware Firewall*, you can retrieve the address by loggging into its admin
+interface and checking the settings under **Firewall ▸ Aliases**.
+
+.. note:: If you do not have login credentials for your pfSense firewall, check 
+ its user manual for instructions on resetting the administration password.
+
+Next, determine whether your instance was set up to allow adminstrative access
+via SSH over Tor, or via SSH over LAN. If you don't know which option was originally
+chosen, you can check as follows:
+
+ #. Log in to the *Application Server* via the console using the adminstration username
+    and password.
+ #. Check to see if an SSH hidden proxy service exists, using the command 
+    ``sudo cat /var/lib/tor/services/ssh/hostname``. If this file exists and
+    includes an Onion URL and authorization token, your instance is set up 
+    to use SSH over Tor and you should configure temporary SSH access
+    using :ref:`these instructions <rebuild_ssh_over_tor>`.
+    If not, your instance is set up to use SSH over LAN, and you should follow
+    :ref:`these instructions instead <rebuild_ssh_over_lan>`.
+
+.. _rebuild_ssh_over_tor:
+
+Configuring access for an SSH-over-Tor instance
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Direct SSH access is disabled when the SSH-over-Tor option is selected during
+installation. To temporarily re-enable it, you'll need to update ``iptables`` rules
+and change the sshd daemon's configuration.
+
+First, log on to the *Application Server* via the console, and run the following
+commands, substituting the *Admin Workstation's* static IP for ``<admin_static_ip>``:
+
+.. code:: sh
+
+  sudo iptables -I INPUT -p tcp --dport 22 -s <admin_static_ip> \
+    -m state --state NEW,ESTABLISHED -j ACCEPT
+  sudo iptables -I OUTPUT -p tcp --sport 22 -m state --state ESTABLISHED -j ACCEPT 
+
+Next, edit the file ``/etc/ssh/sshd_config``, changing the line:
+
+.. code-block:: none
+
+  ListenAddress 127.0.0.1:22                                    
+
+to:
+ 
+.. code-block:: none
+
+  ListenAddress 0.0.0.0:22                                   
+
+and deleting the line: 
+
+.. code-block:: none
+
+  PasswordAuthentication no                                         
+
+Then, restart ``sshd`` using the command ``sudo service sshd restart``.
+
+Finally, check the file ``/etc/network/interfaces``, and note down the network
+settings for the default Ethernet interface. You'll need them in the next step.
+
+Repeat the process above for the *Monitor Server*, making sure to note down its
+network settings as well.
+
+.. _rebuild_ssh_over_lan:
+
+Once the *Monitor Server* has been configured, proceed to :ref:`enable access from
+the new Admin Workstation <enabling_access_from_admin>`.
+
+Configuring access for an SSH-over-LAN instance
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+First, log on to the *Application Server* via the console and edit the file 
+``/etc/ssh/sshd_config``, deleting the line:
+
+.. code-block:: none
+
+  PasswordAuthentication no                                         
+
+Then, restart ``sshd`` using the command ``sudo service sshd restart``.
+
+Finally, check the file ``/etc/network/interfaces``, and note down the network
+settings for the default Ethernet interface. You'll need them in the next step.
+
+Repeat the process above for the *Monitor Server*, making sure to note down its
+network settings as well.
+
+.. _enabling_access_from_admin:
+
+Enabling access from the new *Admin Workstation*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+From the *Admin Workstation*, open a terminal and copy the *Admin Workstation's* 
+SSH public key to the servers, substituting the values for the server administration
+username and server IP addresses in the commands below and entering the admin account's 
+password when prompted:
+
+.. code:: sh
+
+  ssh-copy-id <admin-username>@<application-server-ip>
+  ssh-copy-id <admin-username>@<monitor-server-ip>
+  
+Next, create a file ``~/.ssh/config`` with contents as below, again substituting
+the appropriate values for your servers:
+
+.. code-block:: none
+ 
+  Host app
+    User <admin-username>
+    Hostname <application-server-ip>
+    ProxyCommand none
+
+  Host mon
+    User <admin-username>
+    Hostname <monitor-server-ip>
+    ProxyCommand none
+
+
+Finally, test direct SSH access from the terminal, using the commands ``ssh app`` and 
+``ssh mon``. It should be possible to connect without entering a password.
+
+Step 4: Retrieve SecureDrop configuration info from the servers
+===============================================================
+
+In addition to the account and networking information retrieved from the servers
+so far, you'll need to retrieve the following files and info:
+
+ - Tor hidden service URLs and tokens
+ - GPG *Submission Public Key* and OSSEC public key
+ - OSSEC alert configuration details
+ - (Optional) HTTPS configuration details
+
+Retrieve Onion hidden service info
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In order to connect to the Tor hidden services used by your instance, you will
+need to copy their details to the *Admin Workstation*. To do so, run the following
+commands from a Terminal window:
+
+.. code:: sh
+
+ cd ~/Persistent/securedrop/install_files/ansible-base
+ echo "HidServAuth $(ssh app sudo cat /var/lib/tor/services/journalist/hostname)" \
+    > app-journalist-aths
+ echo "$(ssh app sudo cat /var/lib/tor/services/source/hostname)" > app-source-ths
+
+If your instance was set up to use SSH over TOR, you'll also need to copy over the details 
+of the SSH proxy hidden services, by running the following commands:
+
+.. code:: sh
+
+ echo "HidServAuth $(ssh app sudo cat /var/lib/tor/services/ssh/hostname)" > app-ssh-aths
+ echo "HidServAuth $(ssh mon sudo cat /var/lib/tor/services/ssh/hostname)" > mon-ssh-aths
+
+Retrieve GPG Public Keys 
+~~~~~~~~~~~~~~~~~~~~~~~~
+There are 2 GPG public keys required by the instance configuration, that you will need to 
+copy to the new *Admin Workstation*.
+
+To copy the *Submission Public Key*, download it from the *Source Interface* and import
+it locally using the following commands:
+
+.. code:: sh
+
+ cd ~/Persistent/securedrop/install_files/ansible-base
+ curl http://$(cat app-source-ths)/journalist-key > SecureDrop.asc
+ gpg --import SecureDrop.asc
+
+Validate that the imported key's fingerprint matches the one on your 
+SecureDrop install. You can do this by first running the command:
+
+.. code:: sh
+
+  gpg --with-fingerprint --import-options import-show --dry-run --import SecureDrop.asc
+
+Then, compare the returned fingerprint value with that advertised by your *Source Interface*,
+using the command:
+
+.. code:: sh
+
+ curl http://$(cat app-source-ths)/metadata
+
+To copy the OSSEC alert public key, first list available keys on the monitor server:
+
+.. code:: sh
+ 
+ ssh mon sudo gpg --homedir=/var/ossec/.gnupg  -k
+
+Look for the key corresponding to the destination email address for OSSEC alerts. 
+Then, import it locally using the following commands (substituting the 
+appropriate email address for ``alerts@example.com``):
+
+.. code:: sh
+
+ ssh mon sudo gpg --homedir=/var/ossec/.gnupg --export --armor alerts@example.com > ossec.pub
+ gpg --import ossec.pub
+ 
+You will be prompted for the fingerprints for both keys during the next step. To view their fingerprints, use the command:
+
+.. code:: sh
+ 
+ gpg -k --fingerprint
+
+Retrieve OSSEC alert configuration details
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+In addition to the OSSEC alert GPG key you retrieved above, you'll also need 
+the following configuration information:
+
+ - SMTP server
+ - SMTP port
+ - SASL username
+ - SASL domain
+ - SASL password
+
+To retrieve these values, use the following command in the terminal:
+
+.. code:: sh
+
+ ssh mon sudo cat /etc/postfix/sasl_passwd
+
+This will return a line like:
+
+.. code:: sh
+ 
+ [smtp.gmail.com]:587 testossec@gmail.com:AwfulPassword
+
+In this example, ``smtp.gmail.com`` is the SMTP server, ``587`` is the SMTP port,
+``testossec`` is the SASL username, ``gmail.com`` is the SASL domain, and 
+``AwfulPassword`` is the SASL password.
+
+(Optional) Retrieve HTTPS certificate files
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If your *Source Interface* was configured to use HTTPS, you will need to copy
+three related files from the *Application Server* to the *Admin Workstation*.
+
+To retrieve these files, use the commands:
+
+.. code:: sh
+
+   cd ~/Persistent/securedrop/install_files/ansible-base
+   ssh app sudo tar -c -C /var/lib ssl/  | tar xvf - 
+
+These commands will create a directory named 
+``~/Persistent/securedrop/install_files/ansible-base/ssl``
+on the *Admin Workstation*, containing your instance's SSL certificate, 
+certificate key, and chain file. When prompted for the names of these files
+during the next step, you should specify them relative to the 
+``install_files/ansible-base`` directory, i.e. as ``ssl/mydomain.crt``.
+
+Step 5: Configure and back up the application
+=============================================
+
+Next, configure the application using the files and info retrieved in the 
+previous steps. To do so, connect to the Tor network on the
+*Admin Workstation*, open a Terminal and run the following commands:
+
+.. code:: sh
+
+ cd ~/Persistent/securedrop
+ ./securedrop-admin setup
+ ./securedrop-admin sdconfig
+
+.. note:: The ``./securedrop-admin setup`` command may take several minutes to complete, and may
+ fail due to network issues. If it fails, it's safe to run again.
+
+When prompted, fill in the values retrieved in the previous steps.
+
+Once complete, configure the *Admin Workstation* using the command:
+
+.. code:: sh
+
+  ./securedrop-admin tailsconfig
+
+This will set up desktop shortcuts for the *Source* and *Journalist Interfaces*,
+and configure adminstrative access to the servers.
+
+Once the ``./securedrop-admin tailsconfig`` command is complete, verify that you can connect to 
+the servers using ``ssh app`` and ``ssh mon``, accepting the host verification
+prompt if necessary.
+
+Next, back up the servers by running the following command in the terminal:
+
+.. code:: sh 
+ 
+ ./securedrop-admin backup
+
+
+Step 6: Run the ``./securedrop-admin install`` command
+======================================================
+
+After the ``./securedrop-admin backup`` command completes successfully, you should  undo the
+changes made to enable temporary local SSH access, by running the following 
+command:
+
+.. code:: sh
+
+ ./securedrop-admin install
+
+Once the command completes successfully, your new *Admin Workstation* is ready
+for use. To revert the changes made to enable temporary local SSH access, you
+should reboot the servers, by issuing the following commands in a terminal:
+
+.. code:: sh
+ 
+ ssh app sudo reboot
+ ssh mon sudo reboot
+
+Step 7: Post-rebuild tasks
+==========================
+
+We recommend completing the following tasks after the rebuild:
+
+ - Set up a new administration account on the *Journalist Interface*, by following
+   :doc:`these instructions <create_admin_account>`
+ - Verify that submissions can be decrypted, by going through the decryption 
+   workflow with a new submission.
+ - Back up your *Admin Workstation* using the process 
+   :ref:`documented here <backup_workstations>`.
+ - Delete invalid admin accounts in the *Journalist Interface*.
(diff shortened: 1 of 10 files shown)


docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+https://tails.boum.org/install/index.en.html
+
+Next, use the Tails Master USB to :ref:`create a new Admin Workstation USB <set_up_tails>` and set up a persistent volume with a strong passphrase.
+
+Once persistence has been set up, start up the Admin Workstation with 
+persistence enabled, :ref:`install the SecureDrop application code, and set up
+the KeePassX database <set_up_admin_tails>`.
+
+The Admin Workstation uses SSH with key authentication to connect to the servers,
+so you'l need to create a new SSH keypair for your SecureDrop instance. To do so,
```
eloquence: typo: `you'l`

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+ ssh-keygen -t rsa -b 4096
+
+When prompted to ``Enter file in which to save the key``, Press **Enter** to use
+the default location.
+
+
+Step 2: (Optional) Boot the servers in single-user mode
+=======================================================
+If you do not have the original password for the shell admin account on the
+*Apllication* and *Monitor Servers*, you'll need to reset the password on each
```
eloquence: typo: `Apllication`

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+In addition to the OSSEC alert GPG key you retrieved above, you'll also need 
+the following configuration information:
+
+ - SMTP server
+ - SMTP port
+ - SASL username
+ - SASL domain
+ - SASL password
+
+To  these, use the following command in the terminal:
```
eloquence: typo: double space

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+
+   cd ~/Persistent/securedrop/install_files/ansible-base
+   ssh app sudo tar -c -C /var/lib ssl/  | tar xvf - 
+
+These commands will create a directory named 
+``~/Persistent/securedrop/install_files/ansible-base/ssl``
+on the *Admin Workstation*, containing your instance's SSL certificate, 
+certificate key, and chain file. When prompted for the names of these files
+during the next step, you should specify them relative to the 
+``install_files/ansible-base directory``, ie as ``ssl/mydomain.crt``.
```
eloquence: - `directory` does not belong in pre-formatted string
- typo: `ie` (should be `i.e.`)

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+
+Once the ``tailsconfig`` command is complete, verify that you can connect to 
+the servers using ``ssh app`` and ``ssh mon``, accepting the host verification
+prompt if necessary.
+
+Next, back up the servers by running the following command in the Terminal:
+
+.. code:: sh 
+ 
+ /securedrop-admin backup
```
eloquence: Leading `.` missing

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
+Rebuilding an Admin Workstation USB
```
eloquence: italicize Admin Workstation (we do it even in chapter headings, albeit inconsistently)

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+-----------------------------------
+
+In cases where an Admin Workstation stick has been lost or destroyed, and no
```
eloquence: - italicize Admin Workstation
- maybe "USB stick" instead of just "stick"

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+ - physical access to the SecureDrop servers
+ - 2 USB sticks:
+
+   - Tails Master USB
+   - 1 replacement Admin Workstation USB (USB3 and 16GB or better recommended)
+
+The process requires experience with the Linux command line and Tails, and 
+can take up to 3 hours. If a backup of the SecureDrop application server is available,
+reinstalling the instance may be simpler. An outline of the steps involved is as 
+follows:
```
eloquence: I would say "an outline of the steps involved in rebuilding an *Admin Workstation*", just to resolve ambiguity with the previous sentence, or move the "reinstalling may be simpler" sentence somewhere else.

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+ #. Complete post-rebuild tasks.
+
+
+Step 1: prepare USB sticks
+==========================
+
+First, create a Tails Master USB using the installation procedure described here: 
+https://tails.boum.org/install/index.en.html
+
+Next, use the Tails Master USB to :ref:`create a new Admin Workstation USB <set_up_tails>` and set up a persistent volume with a strong passphrase.
```
eloquence: italicize Admin Workstation

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+
+Once persistence has been set up, start up the Admin Workstation with
```
eloquence: italicize Admin Workstation

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+ Make sure to select a strong password!
+
+Finally, reboot the *Monitor Server* and verify that you can log in at the console
+using the new password.
+
+Repeat the process for the *Application Server*. Use the same username and 
+password as for the *Monitor Server* - this is required in order for the 
+`./securedrop-admin install` command to work correctly.
+
+Step 3: Set up Admin Workstation access
```
eloquence: italicize Admin Workstation

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+-----------------------------------
+
+In cases where an Admin Workstation stick has been lost or destroyed, and no
+backup exists, it is possible to rebuild one. In order to do so, you'll need
+ 
+ - physical access to the SecureDrop servers
+ - 2 USB sticks:
+
+   - Tails Master USB
+   - 1 replacement Admin Workstation USB (USB3 and 16GB or better recommended)
```
eloquence: italicize Admin Workstation

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+First, create a Tails Master USB using the installation procedure described here: 
+https://tails.boum.org/install/index.en.html
+
+Next, use the Tails Master USB to :ref:`create a new Admin Workstation USB <set_up_tails>` and set up a persistent volume with a strong passphrase.
+
+Once persistence has been set up, start up the Admin Workstation with 
+persistence enabled, :ref:`install the SecureDrop application code, and set up
+the KeePassX database <set_up_admin_tails>`.
+
+The Admin Workstation uses SSH with key authentication to connect to the servers,
```
eloquence: italicize Admin Workstation

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+
+.. code:: sh
+
+ echo "HidServAuth $(ssh app sudo cat /var/lib/tor/services/ssh/hostname)" > app-ssh-aths
+ echo "HidServAuth $(ssh mon sudo cat /var/lib/tor/services/mon/hostname)" > mon-ssh-aths
+
+Retrieve GPG Public Keys 
+~~~~~~~~~~~~~~~~~~~~~~~~
+There are 2 GPG public keys required by the instance configuration, that you will need to 
+copy to the new Admin Workstation.
```
eloquence: italicize Admin Workstation

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+ #. (Optional) Boot the *Application* and *Monitor Server* in single user mode and  reset
+    the shell admin account password.
+ #. Set up SSH access for the new *Admin Workstation*.
+ #. Retrieve SecureDrop configuration settings from the  *Application* and *Monitor Server*.
+ #. Back up and configure the SecureDrop application.
+ #. Run the ``install`` and ``tailsconfig`` commands from the new *Admin Workstation*.
+ #. Complete post-rebuild tasks.
+
+
+Step 1: prepare USB sticks
```
eloquence: prepare -> Prepare the

(capitalization for consistency w/ other chapter headings)

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+
+.. important::
+ Make sure to select a strong password!
+
+Finally, reboot the *Monitor Server* and verify that you can log in at the console
+using the new password.
+
+Repeat the process for the *Application Server*. Use the same username and 
+password as for the *Monitor Server* - this is required in order for the 
+`./securedrop-admin install` command to work correctly.
```
eloquence: Command should be pre-formatted, not italicized.

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+public key to the servers, substituting the appropriate values for the username
+and server IP addresses in the commands below and entering the admin account's 
+password when prompted:
+
+.. code:: sh
+
+  ssh-copy-id <admin-username>@<application-server-ip>
+  ssh-copy-id <admin-username>@<monitor-server-ip>
+  
+Next, create a file ``~amnesia/.ssh/config`` with contents as below, again substituting
```
eloquence: Should be `~/.ssh/config`

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+``ssh mon``. It should be possible to connect without entering a password.
+
+Step 4: Retrieve SecureDrop configuration info from the servers
+===============================================================
+
+In addition to the account and networking information retrieved from the servers
+so far, you'll need to retrieve the following files and info:
+
+ - Tor hidden service URLs and tokens
+ - GPG Submission key and OSSEC public key
```
eloquence: Let's use *Submission Public Key* (italicized) per glossary

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+
+ echo "HidServAuth $(ssh app sudo cat /var/lib/tor/services/ssh/hostname)" > app-ssh-aths
+ echo "HidServAuth $(ssh mon sudo cat /var/lib/tor/services/mon/hostname)" > mon-ssh-aths
+
+Retrieve GPG Public Keys 
+~~~~~~~~~~~~~~~~~~~~~~~~
+There are 2 GPG public keys required by the instance configuration, that you will need to 
+copy to the new Admin Workstation.
+
+To copy the *Submission Key*, download it from the *Source Interface* and import
```
eloquence: *Submission Public Key*

(For the avoidance of confusion, I recommend always using the more specific phrase where possible)

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+it locally using the following commands
+
+.. code:: sh
+
+ cd ~/Persistent/securedrop/install_files/ansible-base
+ curl http://$(cat app-source-ths)/journalist-key > SecureDrop.asc
+ gpg --import SecureDrop.asc
+
+Validate that the inported key's fingerprint matches the one on your
```
eloquence: typo: `inported`

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+  ./securedrop-admin tailsconfig
+
+This will set up desktop shortcuts for the *Source* and *Journalist Interfaces*,
+and configure adminstrative access to the servers.
+
+Once the ``tailsconfig`` command is complete, verify that you can connect to 
+the servers using ``ssh app`` and ``ssh mon``, accepting the host verification
+prompt if necessary.
+
+Next, back up the servers by running the following command in the Terminal:
```
eloquence: Terminal->terminal

(not a proper noun)

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+reinstalling the instance may be simpler. An outline of the steps involved is as 
+follows:
+
+ #. Prepare the USB sticks.
+ #. (Optional) Boot the *Application* and *Monitor Server* in single user mode and  reset
+    the shell admin account password.
+ #. Set up SSH access for the new *Admin Workstation*.
+ #. Retrieve SecureDrop configuration settings from the  *Application* and *Monitor Server*.
+ #. Back up and configure the SecureDrop application.
+ #. Run the ``install`` and ``tailsconfig`` commands from the new *Admin Workstation*.
```
eloquence: Throughout, you refer to these commands using the short form ("the `install` command"). In other parts of the documentation, we generally use the long form ("the `securedrop-admin install` command", etc.). I would argue the long form is preferable, if only for familiarity; was there a reason you went with the short form here?

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+ #. Complete post-rebuild tasks.
+
+
+Step 1: prepare USB sticks
+==========================
+
+First, create a Tails Master USB using the installation procedure described here: 
+https://tails.boum.org/install/index.en.html
```
eloquence: For legibility, I would suggest changing this to "using [the official Tails installation guide](https://tails.boum.org/install/index.en.html)" or similar.

eloquence: Right now, you're pointing the user to the Tails download instructions, then to the SecureDrop documentation, which also includes those same download instructions -- including the use of a master to create additional Tails drives. 

How about only asking the user to perform the steps described in
https://docs.securedrop.org/en/release-0.13.0/set_up_tails.html#install-tails
to create a new *Admin Workstation* drive with a persistent volume?

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+instance. You can use the command ``getent passwd 1000`` to check the username
+corresponding to UID 1000. Once you have the correct username, reset its password
+using the `passwd` command, for example:
+
+.. code:: sh 
+
+ passwd sdadmin
+
+.. important::
+ Make sure to select a strong password!
```
eloquence: I'd suggest that we recommend generating and storing the password (or *passphrase*; see #2176) using KeePassX on the Admin Workstation, especially since we just instructed them to configure it.

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+If you do not have the original password for the shell admin account on the
+*Apllication* and *Monitor Servers*, you'll need to reset the password on each
+server by booting in single user mode. In order to do so, you'll need physical
+access to the server, a keyboard, and a monitor. 
+
+First, connect a monitor and keyboard to the *Monitor Server*. Then reboot the server.
+When the GRUB menu appears, make sure the **Ubuntu** entry is highlighted, and 
+press **e** to edit boot options.
+In the boot options for Ubuntu, find the line that starts with ``linux`` and ends
+with ``ro``. Add `` single`` after ``ro`` (make sure to include the space!) and press
```
eloquence: Sadly that `single` formatting won't work, leading or trailing whitespace is not permitted in restructured text inline markup. See [ye olde stackoverflowe](https://stackoverflow.com/questions/39151571/how-to-add-leading-trailing-spaces-to-monospace-text-in-sphinx-restructuredtext/39152201) for some possible workarounds, none of which look great :/

eloquence: Note also that we do have instructions with screenshots for booting into single user mode, here:
https://docs.securedrop.org/en/release-0.13.0/kernel_troubleshooting.html#boot-into-single-user-mode

Linking into that section could be confusing, but might be worth turning those into a re-usable content block?

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: either would be correct, but fixed :)

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: brevity, basically. Happy to change it.

docs/rebuild_admin.rst
zenmonkeykstop: done.

docs/rebuild_admin.rst
zenmonkeykstop: It'd need to be edited to remove reference to the new kernel. Edited to fix the code and added a clause to address the space.

docs/rebuild_admin.rst
zenmonkeykstop: done.

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,404 @@
...
+
+Next, use the Tails Master USB to :ref:`create a new Admin Workstation USB <set_up_tails>` and set up a persistent volume with a strong passphrase.
+
+Once persistence has been set up, start up the Admin Workstation with 
+persistence enabled, :ref:`install the SecureDrop application code, and set up
+the KeePassX database <set_up_admin_tails>`.
+
+The Admin Workstation uses SSH with key authentication to connect to the servers,
+so you'l need to create a new SSH keypair for your SecureDrop instance. To do so,
+open a Terminal and run the following command:
```
eloquence: Terminal -> terminal (one more)

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+persistence enabled, :ref:`install the SecureDrop application code, and set up
+the KeePassX database <set_up_admin_tails>`.
+
+The *Admin Workstation* uses SSH with key authentication to connect to the servers,
+so you'll need to create a new SSH keypair for your SecureDrop instance. To do so,
+open a Terminal and run the following command:
+
+.. code:: sh
+
+ ssh-keygen -t rsa -b 4096
```
eloquence: It also prompts for a passphrase, so we should repeat or include our recommendation from https://docs.securedrop.org/en/release-0.13.0/servers.html#set-up-ssh-keys

docs/rebuild_admin.rst
zenmonkeykstop: fixed.

docs/rebuild_admin.rst
zenmonkeykstop: fixed.

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+
+First, start the new *Admin Workstation* with persistence enabled and an administration
+password set.
+
+Next, connect the new *Admin Workstation* to the *Hardware Firewall* via the
+appropriate Ethernet port, and set up its static IP address. For more information
+on how to do so, see :ref:`this section in the firewall setup documentation
+<assign_static_ip_to_workstation>`. If you do not know the correct static IP
+address for the *Admin Workstation*, log in to the *Hardware Firewall's* admin
+interface and check the settings under ``Aliases``.
```
eloquence: - As "text directly taken from a user interface", the pointer to the "Aliases" section should be formatted in bold, like so:  **Firewall ▸ Aliases**
- If they have lost their Admin Workstation USB, that may include loss of credentials for the firewall, so I think we should include a pointer to how to reset those (perhaps https://docs.netgate.com/pfsense/en/latest/usermanager/locked-out-of-the-webgui.html for pfSense? haven't tested those instructions).

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+
+Next, determine whether your instance was set up to allow adminstrative access
+via SSH over Tor, or via SSH over LAN. If you don't know which option was originally
+chosen, you can check as follows:
+
+ #. Log in to the *Application Server* via the console using the adminstration username
+    and password.
+ #. Check to see if a file ``/var/lib/services/tor/ssh/hostname`` exists. If it
```
eloquence: - Should be `/var/lib/tor/services/ssh/hostname`, not  `/var/lib/services/tor/ssh/hostname`
- Since it requires `sudo`, maybe provide command to check? `sudo cat /var/lib/tor/services/ssh/hostname` or similar

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+
+Next, connect the new *Admin Workstation* to the *Hardware Firewall* via the
+appropriate Ethernet port, and set up its static IP address. For more information
+on how to do so, see :ref:`this section in the firewall setup documentation
+<assign_static_ip_to_workstation>`. If you do not know the correct static IP
+address for the *Admin Workstation*, log in to the *Hardware Firewall's* admin
+interface and check the settings under ``Aliases``.
+
+Next, determine whether your instance was set up to allow adminstrative access
+via SSH over Tor, or via SSH over LAN. If you don't know which option was originally
```
eloquence: Maybe note that SSH over Tor is the default?

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+and change the sshd daemon's configuration.
+
+First, log on to the *Application Server* via the console, and run the following
+commands, substituting the *Admin Workstation's* static IP for ``<admin_static_ip>``:
+
+.. code:: sh
+
+  sudo iptables -I INPUT -p tcp -dport 22 -s <admin_static_ip> \
+    -m state --state NEW,ESTABLISHED -j ACCEPT
+  sudo iptables -I OUTPUT -p tcp -sport 22 -m state --state ESTABLISHED -j ACCEPT
```
eloquence: Both `iptables` commands fail with "Bad argument '22'". The culprit appears to be `-sport`, `-dport` (should be `--sport`, `--dport`).

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+
+and deleting the line: 
+
+.. code-block:: none
+
+  PasswordAuthentication no                                         
+
+Then, restart ``sshd`` using the command ``sudo service sshd restart``.
+
+Finally, check the file ``/etc/networking/interfaces``, and note down the network
```
eloquence: Should be `/etc/network/interfaces`, not `/etc/networking/interfaces`

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+settings for the default Ethernet interface. You'll need them in the next step.
+
+Repeat the process above for the *Monitor Server*, making sure to note down its
+network settings as well.
+
+.. _rebuild_ssh_over_lan:
```
eloquence: I would recommend adding a note here to skip ahead to the section "Enabling access from the new *Admin Workstation*", since the next section does not apply to them, especially since we refer to "the next step" above.

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+First, log on to the *Application Server* via the console and edit the file 
+``/etc/ssh/sshd_config``, deleting the line:
+
+.. code-block:: none
+
+  PasswordAuthentication no                                         
+
+Then, restart ``sshd`` using the command ``sudo service sshd restart``.
+
+Finally, check the file ``/etc/networking/interfaces``, and note down the network
```
eloquence: As above, should be `/etc/network/interfaces`

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+settings for the default Ethernet interface. You'll need them in the next step.
+
+Repeat the process above for the *Monitor Server*, making sure to note down its
+network settings as well.
+
+Enabling access from the new *Admin Workstation*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+From the *Admin Workstation*, open a terminal and copy the *Admin Workstation's* 
+SSH public key to the servers, substituting the appropriate values for the username
```
eloquence: "the appropriate values" feels a bit handwavy -- I suggest we explicitly direct the user to the "address" definition in `/etc/network/interfaces`.

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+In order to connect to the Tor hidden services used by your instance, you will
+need to copy their details to the *Admin Workstation*. To do so, run the following
+commands from a Terminal window:
+
+.. code:: sh
+
+ cd ~/Persistent/securedrop/install_files/ansible-base
+ echo "HidServAuth $(ssh app sudo cat /var/lib/tor/services/journalist/hostname)" \
+    > app-journalist-aths
+ echo "$(ssh app sudo cat /var/lib/tor/services/journalist/hostname)" > app-source-ths
```
eloquence: Should be `/var/lib/tor/services/source/hostname`, not `/var/lib/tor/services/journalist/hostname`

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+
+If your instance was set up to use SSH over TOR, you'll also need to copy over the details 
+of the SSH proxy hidden services, by running the following commands:
+
+.. code:: sh
+
+ echo "HidServAuth $(ssh app sudo cat /var/lib/tor/services/ssh/hostname)" > app-ssh-aths
+ echo "HidServAuth $(ssh mon sudo cat /var/lib/tor/services/mon/hostname)" > mon-ssh-aths
```
eloquence: Should be `/var/lib/tor/services/ssh/hostname`, not `/var/lib/tor/services/mon/hostname`

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+ cd ~/Persistent/securedrop/install_files/ansible-base
+ curl http://$(cat app-source-ths)/journalist-key > SecureDrop.asc
+ gpg --import SecureDrop.asc
+
+Validate that the imported key's fingerprint matches the one on your 
+SecureDrop install. You can do this by first running the command:
+
+.. code:: sh
+
+ gpg -k --fingerprint
```
eloquence: This dumps a few different fingerprints, since Tails comes with a few keys preinstalled. Perhaps it would be worth verifying the fingerprint from the file before importing it. The short version to do that is `gpg --with-fingerprint SecureDrop.asc`, but that's ambiguous; the non-ambiguous version that worked for me on current Tails is `gpg --with-fingerprint --import-options import-show --dry-run --import SecureDrop.asc`.

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+
+ curl http://$(cat app-source-ths)/metadata
+
+To copy the OSSEC alert public key, first list available keys on the monitor server:
+
+.. code:: sh
+ 
+ ssh mon sudo gpg --homedir=/var/ossec/.gnupg  -k
+
+Look for the key corresponding to the destination email address for OSSEC alert. Then, import it locally using the following commands (substituting the appropriate email address for ``alerts@example.com``:
```
eloquence: - for OSSEC alert -> for OSSEC alerts
- missing closing parenthesis after `alerts@example.com`

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+
+ - Set up a new administration account on the *Journalist Interface*, by following
+   :doc:`these instructions <create_admin_account>`
+ - Verify that submissions can be decrypted, by going through the decryption 
+   workflow with a new submission.
+ - Back up your *Admin Workstation* using the process 
+   :ref:`documented here <backup_workstations>`.
+ - Delete invalid admin accounts in the *Journalist Interface*.
+ - Delete invalid SSH public keys from the administration shell account on the
+   *Application* and *Monitor Servers*.
```
eloquence: I think we should explain how to do this part (removal of invalid SSH public keys).

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+
+We recommend completing the following tasks after the rebuild:
+
+ - Set up a new administration account on the *Journalist Interface*, by following
+   :doc:`these instructions <create_admin_account>`
+ - Verify that submissions can be decrypted, by going through the decryption 
+   workflow with a new submission.
+ - Back up your *Admin Workstation* using the process 
+   :ref:`documented here <backup_workstations>`.
+ - Delete invalid admin accounts in the *Journalist Interface*.
```
eloquence: Can we link somewhere for this part?

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+.. code:: sh
+ 
+ ssh mon sudo gpg --homedir=/var/ossec/.gnupg  -k
+
+Look for the key corresponding to the destination email address for OSSEC alert. Then, import it locally using the following commands (substituting the appropriate email address for ``alerts@example.com``:
+
+.. code:: sh
+
+ ssh mon sudo gpg --homedir=/var/ossec/.gnupg --export --armor alerts@example.com > ossec.asc
+ gpg --import ossec.asc
```
eloquence: Note that [the default filename is ossec.pub](https://github.com/freedomofpress/securedrop/blob/develop/admin/securedrop_admin/__init__.py#L331), so it may make sense to use that here as well to avoid confusion later.

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,407 @@
...
+After the ``./securedrop-admin backup`` command completes successfully, you should  undo the
+changes made to enable temporary local SSH access, by running the following 
+command:
+
+.. code:: sh
+
+ ./securedrop-admin install
+
+Once the command completes successfully, your new *Admin Workstation* is ready
+for use.
```
eloquence: At this point, I was still able to access app and mon server over the local network, even though SSH over Tor was configured. Restarting SSH resolved this, but that suggests to me that the `iptables` update was not in effect either. Perhaps the simplest step to recommend here is to reboot both servers?

eloquence:
Thanks for your extensive work on this, it's quite an epic docs update! :)

I have fully tested this procedure with the following exceptions:
- I did not test the retrieval of HTTPS certificate files.
- I did not test most of the post-rebuild tasks (will at least do an end-to-end test, but I anticipate no issues there).
- I have not verified that the old iptables rules were correctly reinstated (could use your help with that).

I had to tweak some steps along the way (see comments), but with those edits, I was able to rebuild an Admin Workstation. At the end, `securedrop-admin install` failed twice with "Waiting for privilege escalation", a reminder that we really should pull #4358 into an upcoming sprint.

It'll be great to land this, but the procedure is of course quite burdensome; I definitely think this is an area where even light scripting/automation could make a significant difference.

docs/rebuild_admin.rst
zenmonkeykstop: Added a note - but complicated by the fact that folk may not be using a pfSense FW.

docs/rebuild_admin.rst
zenmonkeykstop: Fixed.

docs/rebuild_admin.rst
zenmonkeykstop: Not useful information at this point.

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: sounds good. added.

docs/rebuild_admin.rst
zenmonkeykstop: removed handwaviness.

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: done.

docs/rebuild_admin.rst
zenmonkeykstop: Fixed.

docs/rebuild_admin.rst
zenmonkeykstop: Same, but I'd put it in a separate doc describing off-boarding admins, linked to from here once it exists. I thought about adding a command to remove the authorized_keys files from the servers early on, but that might break access for legit admins.

docs/rebuild_admin.rst
zenmonkeykstop: For deleting admins? weirdly enough, no. The docs describe how to add them but not how to remove (probably because it seemed like a pretty trivial operation given the expected experience of instance admins.)

docs/rebuild_admin.rst
zenmonkeykstop: fixed

docs/rebuild_admin.rst
zenmonkeykstop: Good suggestion - the first nightly reboot will reset iptables and restart sshd, but adding in an explicit reboot isn't a big deal.

zenmonkeykstop:
I tested the HTTPS cert fetching process by creating the ``/var/lib/ssh`` directory manually, which, admittedly, isn't ideal. Added a reboot step to ensure the network changes get reverted before the next scheduled reboot.

zenmonkeykstop:
(Force-pushed to include CI fix)

docs/rebuild_admin.rst
eloquence: I agree a separate doc would be great, but perhaps we can add a bit more detail here until we have that? Revoking invalid SSH accounts feels like a crucial part of the process; IMO we should give less experienced admins a bit more guidance here (including caveats about the risks).

zenmonkeykstop: This is actually a pretty straightforward candidate for Ansiblization, excuse the temporary lint mess while I see about adding a playbook to clean up authorized_keys.

docs/rebuild_admin.rst
eloquence: LGTM.

docs/rebuild_admin.rst
eloquence: Leaving this open in case @emkll  or other reviewers have different suggestions for how to handle this part.

docs/rebuild_admin.rst
zenmonkeykstop: Added a securedrop-admin command and some more docs to address this - see updated test plan.

eloquence:
I've tested the `securedrop-admin reset_admin_access` per the test plan (I'm assuming `test_admin_access` is a typo/leftover) and can confirm that it works as intended:

- Keys other than the one on the currently used Admin Workstation are removed from `~/.ssh/authorized_keys` on both servers

- Repeatedly running the command does not alter state after that initial modification

- If `id_rsa.pub` does not exist, the command fails without making any modifications

It's not in the test plan, and I have _not_ verified the behavior with more than one SSH user as that's not an officially supported configuration.

Nice that Ansible gives us that `authorized_key` module so we don't have to write that logic ourselves :)

install_files/ansible-base/roles/reset-ssh-key/tasks/reset-ssh-key.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: Set authorized key, removing all the authorized keys already set
```
eloquence: The "removing all the authorized keys" bit may sound a bit scary in command output, perhaps shorten to something like "Set authorized key exclusively to the public key configured on this workstation"

install_files/ansible-base/roles/reset-ssh-key/tasks/reset-ssh-key.yml
```diff
@@ -0,0 +1,10 @@
...
+  authorized_key:
+    user: "{{ item }}"
+    key: "{{ lookup('file', '/home/amnesia/.ssh/id_rsa.pub') }}"
+    state: present
+    exclusive: True
+  with_items: "{{ ssh_users }}"
```
eloquence: The dynamic inventory check in https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/inventory-dynamic#L87 will throw an error when running this task with more than one user configured in `site-specific`. Given this, is the `with_items` loop really giving us anything?

zenmonkeykstop: Fair point. Updated to expect a single value.

install_files/ansible-base/roles/reset-ssh-key/tasks/reset-ssh-key.yml
zenmonkeykstop: Changed

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,436 @@
...
+ - Restrict SSH access to the *Application* and *Monitor Servers* to valid 
+   *Admin Workstions*. If your new *Admin Workstation* USB stick
+   is the only one that should have SSH access to the servers, you can remove 
+   access for any previous *Admin Workstations* from the terminal,  using the 
+   commands:
+
+   .. code:: sh
+
+     cd ~/Persistent/securedrop
+     ./securedrop-admin restrict_admin_access
```
eloquence: Should be `reset_admin_access`

eloquence:
Aside from one last nit, all my comments have been addressed and it LGTM! For final review by a maintainer, I would recommend _not_ duplicating the entire test plan (I've already stepped through it and confirmed that it works as advertised), but sanity-checking the overall flow, and paying particular attention to these components:

- the new `securedrop-admin reset_admin_access` command
- the `gpg` invocations in case there's some room for simplification there (see comment I left open).

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,436 @@
...
+appropriate email address for ``alerts@example.com``):
+
+.. code:: sh
+
+ ssh mon sudo gpg --homedir=/var/ossec/.gnupg --export --armor alerts@example.com > ossec.pub
+ gpg --import ossec.pub
+ 
+You will be prompted for the fingerprints for both keys during the next step. To view their fingerprints, use the command:
+
+.. code:: sh
```
emkll: There's also the journalist key, if journalist notifications are enabled

docs/rebuild_admin.rst
```diff
@@ -0,0 +1,436 @@
...
+Once complete, configure the *Admin Workstation* using the command:
+
+.. code:: sh
+
+  ./securedrop-admin tailsconfig
+
+This will set up desktop shortcuts for the *Source* and *Journalist Interfaces*,
+and configure adminstrative access to the servers.
+
+Once the ``./securedrop-admin tailsconfig`` command is complete, verify that you can connect to
```
emkll: Is there a reason why tailsconfig is done prior to install? the overview states install before tailsconfig https://github.com/freedomofpress/securedrop/pull/4522/files#diff-5aa6154ded0c00ed8eb19ab7a2b2031eR24

Since the ssh config is valid, maybe you want the user to confirm at this step that the desktop shortcuts for source and journalist interface are working?

docs/rebuild_admin.rst
zenmonkeykstop: fixed, d'oh

docs/rebuild_admin.rst
zenmonkeykstop: In that case, there's also the destination address for journalist alerts. I'll update the doc for both.

zenmonkeykstop: Added as a post-configuration step, as it's not necessary in order to get the Admin WS stick working.

docs/rebuild_admin.rst
zenmonkeykstop: Yup, that'd be prudent. Updated.

emkll:
Thanks for the fixes @zenmonkeykstop, changes look great. Perhaps one final comment: it might be prudent to state at the start of the guide the importance of completing these tasks in one pass as much as possible. Because the configuration will be changed by hand, this might pose some issues if config changes are done  in postinst or otherwise, and might be hard to test for (notably iptables and ssh config).

zenmonkeykstop:
Added an `important!` to address @emkll's last point.

Issue #4524, opened 2019-06-14
Release SecureDrop 0.13.1


redshiftzero:
This is a tracking issue for the upcoming release of SecureDrop 0.13.1 - tasks may get added or modified. 

**Feature freeze:** EOD Pacific time, Friday, June 14
**Release date:** Tuesday, June 18
 
_SecureDrop maintainers and testers:_ As you QA 0.13.1, please report back your testing results as comments on this ticket. File GitHub issues for any problems found, tag them "QA: Release", and associate them with the 0.13.1 milestone for tracking (or ask a maintainer to do so). 

Test debian packages will be posted on https://apt-test.freedom.press signed with [the test key](https://gist.githubusercontent.com/conorsch/ec4008b111bc3142fca522693f3cce7e/raw/2968621e8ad92db4505a31fcc5776422d7d26729/apt-test%2520apt%2520pubkey). An Ansible playbook testing the upgrade path is [here](https://gist.github.com/conorsch/e7556624df59b2a0f8b81f7c0c4f9b7d).

# Prepare release candidate (0.13.1~rc1)
- [x] Prepare 0.13.1-rc1 release changelog - @emkll 
- [x] Branch off 0.13.1~rc1 into release/0.13.1 - @emkll
- [x] Build debs and put up `0.13.1~rc1` on test apt server @emkll 

### Test plan

https://github.com/freedomofpress/securedrop/wiki/0.13.1-Test-plan
### QA matrix

https://docs.google.com/spreadsheets/d/19L1zj4otg2IoTxSKw5ZLaNjgUm7uZETvc37LkX6IQ1E/edit#gid=0

# Final release
- [x] Ensure builder in release branch is updated and/or update builder image @zenmonkeykstop @emkll 
- [x] Push signed tag @redshiftzero 
- [x] Build final Debian packages for 0.13.1 - @conorsch 
- [x] Upload Debian packages to apt test - @conorsch 
- [x] Pre-Flight: Test install and upgrade (both cron-apt on Trusty, and Ansible on Xenial) of 0.13.1 works w/ prod repo debs, test updater logic in Tails
- [x] Prepare and distribute release messaging

# Post release

- [x] Merge changelog back to `develop` #4537 
- [x] Update upgrade testing boxes
- [x] Update roadmap wiki page

rmol:
### Environment

- Install target: NUC7i5BNH
- Tails version: 3.14
[...]

PR #4528, opened 2019-06-17
[docs] Add 0.13.0 to 0.13.1 upgrade docs

eloquence:
## Status

Ready for review

Part of #4524

## Checklist

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -86,6 +86,7 @@ anonymous sources.
    :name: upgradetoc
    :maxdepth: 2
 
+   upgrade/0.13.0_to_0.13.1.rst
    upgrade/0.12.2_to_0.13.0.rst
    upgrade/xenial_after_april_30.rst
    upgrade/xenial_prep.rst
--- a/docs/upgrade/0.13.0_to_0.13.1.rst
+++ b/docs/upgrade/0.13.0_to_0.13.1.rst
@@ -1,0 +1,66 @@
+Upgrade from 0.13.0 to 0.13.1
+=============================
+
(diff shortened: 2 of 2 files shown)


eloquence:
Notes:

- I omitted the "Backup your workstations" reminder; as a rule, I would suggest including this reminder for minor or major level releases, but not for patch level releases, unless the patch has major implications for Admin Workstation operations.

- I omitted references to Ubuntu 16.04 as it feels like we've hit diminishing returns on any additional messaging regarding this upgrade in the docs.

eloquence:
Added `bandit` version pinning commit to get CI to green, per discussion w/ @redshiftzero and similar workaround in
freedomofpress/securedrop-client#428

Makefile
```diff
@@ -107,7 +107,7 @@ safety: ## Runs `safety check` to check python dependencies for vulnerabilities
 .PHONY: bandit
 bandit: ## Run bandit with medium level excluding test-related folders
 	pip install --upgrade pip && \
-        pip install --upgrade bandit!=1.6.0 && \
+        pip install --upgrade bandit==1.5.1 && \
```
zenmonkeykstop: It looks like the behaviour of the --exclude argument changed in 1.6.1 - you now have to include the root directory in exclude paths, i.e. ``bandit --recursive parent --exclude parent/child`` as opposed to the previous ``bandit --recursive parent --exclude child``. This might be better than pinning to an older version.

eloquence: Omitted from this PR and rebased to `develop` as this was fixed in #4530.

Issue #4296, opened 2019-03-25
`make update-pip-requirements` should be virtualized


heartsucker:
## Description

If `pip-tools` uses the system python version, requirements files will be generated that have compatible with the version of python used to execute `pip-compile`. To reproduce, look at the different files created by this command with `$python_version` set to `python2` and `python3`.

```bash
rm -f securedrop/requirements/*.txt
virtualenv -p $(which $python_version) venv
source venv
pip install pip-tools
make update-pip-requirements
```

heartsucker:
Also as a note, this needs to be the *exact* python version that is running the application. I have had issues with using pip-tools under 3.7 for apps that run under 3.5.

eloquence:
This has been resolved for a while; pip-compile runs are always executed via the dev-shell in a container.

PR #4544, opened 2019-06-20
Make Python 3 the default for development

### LINKED ISSUE #4415: Keep two different versions of requirements.txt file for Python2 and Python3
## Description

`pip-compile` creates different output for `requirements.txt` for Python2 and in Python3. Which in turn creates problems in installing those dependencies in the different versions. 

## Probable solution

Have two different requirements files for Python2 and Python3, for both dev and tests.

### LINKED ISSUE #4519: Use Python 3 in development environment
## Description

We should transition our development environment to Python 3. We're preparing to switch the source and journalist production servers to Python 3 with 0.15.0, and aligning development and production will reduce the likelihood of discovering problems only during deployment.

See also #4292.

rmol:
## Status

Ready for review

## Description of Changes

Make the default .venv Python 3, moving existing Python 2 versions out
of the way if necessary.  Compile and use separate requirements files
for each version, using .venv2 to update the Python 2 requirements.

Add Makefile targets to make it easier to create both virtualenvs;
devops/scripts/boot-strap-venv.sh can now be sourced as before or just
run to create and update the virtualenv.

Merge Makefile and securedrop/Makefile. I've left securedrop/Makefile
in place with targets echoing deprecation warnings, then invoking
their equivalents in the top-level Makefile.

Type linting no longer requires a dedicated virtualenv once we're on
Python 3, so mypy is now part of the normal development requirements.

Switch mypy checking from 2.7 to Python 3.5.

Our Python dependencies also required updates, starting with Molecule,
which didn't support Python 3.5 until 2.20.1, so would have required
us to run a more recent Python 3 for development than we deploy under.

The pip-tools package needed to be updated to work with pip 19.1.1,
which is installed in the new Python 3 virtualenv.

Fix many, many lint errors from lint packages updated for Python 3.

Fixes #4519. 
Fixes #4415.

## Testing

Try running various Makefile targets: 
- `make lint`
- `make test`
- `PYTHON_VERSION=2 make test`
- `make update-pip-requirements`
- `make update-python2-requirements`
- `make build-debs` (you should only get the usual apt update test failure)
- `make staging`
- `molecule verify -s libvirt-staging`

## Deployment

This should not include any changes to production deployment. 

Debian packages are still built with Python 2, and securedrop-app-code is still using the Python 2 version of libapache2-mod-wsgi. The only changes to building debs were to copy the new requirement file paths and specify that the Python 2-specific ones should be used when building wheels.

The only changes to the admin were to satisfy linters, and tests are passing.


.gitignore
```diff
@@ -140,9 +140,6 @@ junit/
 raw-test-output/
 *results.xml
 
-# Virtualenv on tails
-.venv
```
redshiftzero: we should keep this because this is where we keep the virtualenv on tails (admins/journalists use git checkouts on tails)

rmol: It's redundant; `.venv*` is present above.

devops/scripts/boot-strap-venv.sh
```diff
@@ -1,19 +1,39 @@
...
-        source .venv/bin/activate
-        set -u
-        pip install -r securedrop/requirements/develop-requirements.txt
+    if [ ! -d "$VENV" ]
+    then
+        p=$(which "python${PYTHON_VERSION}")
+        virtualenv -p "${p}" "${VENV}"
     fi
+
+    "${VENV}/bin/pip" install -q pip==19.1.1
```
redshiftzero: hmm remind me why we pin this version? (can we add a comment explaining for future us?)

journalist_gui/journalist_gui/resources_rc.py
```diff
@@ -873,18 +873,25 @@
 \x00\x00\x01\x62\xd9\x46\xbd\x12\
 "
 
-qt_version = QtCore.qVersion().split('.')
-if qt_version < ['5', '8', '0']:
+qt_version = QtCore.qVersion().split(".")
+if qt_version < ["5", "8", "0"]:
```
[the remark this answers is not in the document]
redshiftzero: hey can you expand on this comment? I see you're saying "on Tails 3.14.x we use `qt_version < 5.8.0`" but I don't think we want to make a change to the logic in `resources_rc.py` - do you agree? (just want to check I'm not missing something)

securedrop/Makefile
```diff
@@ -1,79 +1,71 @@
...
-lint: ## Run the python linter with the project's default error settings
-	find . -name '*.py' | xargs pylint --reports=no --errors-only \
-	   --disable=no-name-in-module \
-	   --disable=unexpected-keyword-arg \
-	   --disable=too-many-function-args \
-	   --disable=import-error \
-	   --disable=no-member \
-	   --max-line-length=100
+lint: ## Run all linters
+	@echo "\n    ►►► securedrop/Makefile is deprecated. Run 'make lint' in the repo root. ◀◀◀\n"
```
redshiftzero: let's file a followup after this PR is merged to update dev docs (since things Still Work here not going to append here since this diff is already large), but still worth doing next so that the Happy Contributor Path doesn't involve hitting deprecation warnings

securedrop-admin
```diff
@@ -5,10 +5,6 @@
 # the top level of the SecureDrop repository
 d=$(dirname "$0")
 if test "$1" = "setup" || test "$2" = "setup"; then
-   # the .venv symlink is to not confuse devs used to
-   # finding the directory at the top-level, prior to 0.6
-   rm -fr "$d/.venv"
-   ln -s "$d/admin/.venv" "$d/.venv"
```
redshiftzero: hmm... why was this removed?

Makefile
```diff
@@ -1,175 +1,389 @@
...
+	@$(VENV2)/bin/pip-compile \
+		--output-file securedrop/requirements/python2/develop-requirements.txt \
+		admin/requirements-ansible.in \
+		securedrop/requirements/python2/develop-requirements.in
+	@$(VENV2)/bin/pip-compile \
+		--output-file securedrop/requirements/python2/test-requirements.txt \
+		securedrop/requirements/python2/test-requirements.in
+	@$(VENV2)/bin/pip-compile --generate-hashes \
+		--output-file securedrop/requirements/python2/securedrop-app-code-requirements.txt \
+		securedrop/requirements/python2/securedrop-app-code-requirements.in
```
conorsch: For the `pip-compile` calls, we'd do better to wrap those in a container with a specific Python version, as described in #4296. Reasons:

* Developers may not use a predictable location for their venvs (`echo $VIRTUAL_ENV` on my machine shows `/home/user/.virtualenvs/sd`, for instance)
* Directly invoking a script _outside_ whatever the "active" virtualenv is will cause problems. By bottling up in a container, we'll get more predictable behavior on the pip updates.

The second point is critical for app-code dependency updates: we want to use specifically the version of Python used by apt servers, _not_ whatever version of Python is used on a developer laptop.

conorsch: For an example of where we use containerized pip-compile elsewhere, see https://github.com/freedomofpress/securedrop.org/blob/b074ade61d58220adfdd0fb183d3da535388c970/Makefile#L43-L51 For Xenial support, we want Python 3.5.2:

```
vagrant@app-staging:~$ python3 --version
Python 3.5.2
```

securedrop/models.py
```diff
@@ -30,7 +30,7 @@
     # statements has to be marked as noqa.
     # http://flake8.pycqa.org/en/latest/user/error-codes.html?highlight=f401
     from typing import Callable, Optional, Union, Dict, List, Any  # noqa: F401
-    from tempfile import _TemporaryFileWrapper  # noqa: F401
+    from tempfile import _TemporaryFileWrapper  # type: ignore # noqa: F401
```
redshiftzero: seems like we don't actually need this line for type checking? (I don't see `_TemporaryFileWrapper` in the type hints anywhere)

securedrop-admin
rmol: Because we want to keep the admin on Python 2 for now. I guess we could use .venv3 instead of .venv in the top level, for everything else, and rename everything once Python 2 is purged. I was just going down the path of using Python 3 by default and special-casing whatever we didn't want to migrate yet.

securedrop/models.py
rmol:
```

Comment (redshiftzero) [reply securedrop/models.py]: sorry by this line, I meant the import itself

Comment (rmol) [reply securedrop/models.py]: Yep, you're right.

File: securedrop/models.py [hunk 8]
```diff
@@ -670,10 +670,10 @@ def login(cls, username, password, token):
         return user
 
     def generate_api_token(self, expiration):
-        # type: (int) -> unicode
+        # type: (int) -> str
         s = TimedJSONWebSignatureSerializer(
             current_app.config['SECRET_KEY'], expires_in=expiration)
-        return s.dumps({'id': self.id}).decode('ascii')
+        return s.dumps({'id': self.id}).decode('ascii')  # type:ignore
```
redshiftzero: I was concerned this was a bug on our side but it looks like the type annotations for `itsdangerous` just need to be updated upstream: https://github.com/python/typeshed/issues/3081

securedrop/secure_tempfile.py
```diff
@@ -3,7 +3,7 @@
 import os
 import io
 import six
-from tempfile import _TemporaryFileWrapper
+from tempfile import _TemporaryFileWrapper  # type: ignore
```
redshiftzero: noting for myself/others: this change is necessary is because `_TemporaryFileWrapper` is not in the type annotations for [Python 3's tempfile](https://github.com/python/typeshed/blob/master/stdlib/3/tempfile.pyi) upstream (but it is in [Python 2's tempfile](https://github.com/python/typeshed/blob/master/stdlib/2/tempfile.pyi#L20)) - will file issue upstream in a bit

devops/scripts/boot-strap-venv.sh
rmol: It wasn't necessary.

Makefile
rmol: Added. I chose to do this in our existing Ubuntu 16.04 containers, as they should match production.

journalist_gui/journalist_gui/resources_rc.py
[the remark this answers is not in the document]
kushaldas: Ah, this was just a comment, this change in the PR is due to the `black` running on the Python script.

securedrop/models.py
redshiftzero: (now fixed upstream, next mypy release we should be able to remove `# type: ignore`, but until then needs to be left in)

.circleci/config.yml
```diff
@@ -99,12 +108,11 @@ jobs:
           no_output_timeout: 20m
           command: |
             BRANCH_MATCH=$(devops/scripts/match-ci-branch.sh "^(i18n|docs|update-builder)")
-            echo "match-ci-branch.sh said: ${BRANCH_MATCH}"
             if [[ $BRANCH_MATCH =~ ^found ]]; then echo "Skipping: ${BRANCH_MATCH}"; exit 0; fi
             export TESTFILES=$(cd securedrop; circleci tests glob 'tests/test*py' 'tests/**/test*py' |circleci tests split --split-by=timings |xargs echo)
             docker rm -f securedrop-test-xenial-py2 || true
             fromtag=$(docker images |grep securedrop-test-xenial-py2 |head -n1 |awk '{print $2}')
-            cd securedrop && DOCKER_RUN_ARGUMENTS=$(bash <(curl -s https://codecov.io/env)) DOCKER_BUILD_ARGUMENTS="--cache-from securedrop-test-xenial-py2:${fromtag:-latest}" make test
+            PYTHON_VERSION=2 DOCKER_RUN_ARGUMENTS=$(bash <(curl -s https://codecov.io/env)) DOCKER_BUILD_ARGUMENTS="--cache-from securedrop-test-xenial-py2:${fromtag:-latest}" make test
```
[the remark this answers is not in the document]
rmol: Fixed.

.circleci/config.yml
```diff
@@ -251,7 +234,7 @@ jobs:
           name: Install pip==18.0 to work pipenv/pip bug
           command: |
             cd journalist_gui
-            pipenv run pip install pip==18.0
+            pipenv run pip3 install pip==18.0
```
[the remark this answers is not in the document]
rmol: Reverted.

securedrop/Makefile
rmol: OK.

Makefile
```diff
@@ -1,183 +1,397 @@
 DEFAULT_GOAL: help
 SHELL := /bin/bash
 GCLOUD_VERSION := 222.0.0-1
-PWD := $(shell pwd)
+SDROOT := $(shell pwd)
 TAG ?= $(shell git rev-parse HEAD)
 STABLE_VER := $(shell cat molecule/shared/stable.ver)
+VENV := $(shell readlink -m .venv)
```
redshiftzero: [running through test plan] this is not portable on macOS (dev env needs to work on both mac and linux): `readlink`'s  `-m` option does not exist as mac is using a very old version of `readlink`

Makefile
```diff
@@ -1,183 +1,397 @@
...
 # The `-n` option enables "nit-picky" mode.
-	make -C docs/ clean && sphinx-build -Wn docs/ docs/_build/html
+	@. $(VENV)/bin/activate && make -C docs/ clean && sphinx-build -Wn docs/ docs/_build/html
+	@echo
 
 .PHONY: docs-linkcheck
-docs-linkcheck: ## Check documentation for broken links
-	make -C docs/ clean && sphinx-build -b linkcheck -Wn docs/ docs/_build/html
+docs-linkcheck:  ## Check documentation for broken links.
+	@. $(VENV)/bin/activate && make -C docs/ clean && sphinx-build -b linkcheck -Wn docs/ docs/_build/html
```
redshiftzero: summary of synchronous discussion regarding letting devs manage their virtualenvs:

1. add a note in the dev docs that the default virtualenv should be python 3 
2. update the paths in the dev docs to the python3/requirements files 
3. add a small note in the docs that "note: this project uses python 2.7 in production and tests in CI are also ran in python 2.7" - we could optionally add much more to this section, but the idea here is that soon after SecureDrop 0.14.0 is released, we won't need to have `master` contain any python 2.7 compliant code anymore for the webapps (to be clear: SecureDrop 0.14.0 is not shipping Python 3 in prod, but around the time of its release we will have at least a wip PR up with the migration path)
2. we'll remove the `$(VENV)/bin/activate` from the makefile and let the dev activate the Python 3 virtualenv

Makefile
rmol: Removed.

Makefile
rmol: Doc notes added, activation removed.

securedrop-admin
redshiftzero: while this change shouldn't be necessary to keep admins on Python 2 (since admins don't set up a virtualenv using the dev env logic, instead the logic in `securedrop-admin` only is used, and we don't have logic yet in that script to generate a python 3 virtualenv), it's fine to remove at this point. It's been a long time since 0.6 - worst case scenario that I'm aware of with this if there's an admin who hasn't run `securedrop-admin` since pre-0.6, is that they'll need to wait for `securedrop-admin setup` to re-set up the virtualenv.

securedrop/requirements/python3/securedrop-app-code-requirements.txt
```diff
@@ -0,0 +1,200 @@
+#
```
redshiftzero: Noting here for other maintainers the requirements diff compared to `securedrop-app-code-requirements.txt` on `develop` - this PR only removes dependencies from the python 3 version of the app code requirements: 

```diff 
 #
 alembic==0.9.9 \
     --hash=sha256:85bd3ea7633024e4930900bc64fb58f9742dedbc6ebb6ecf25be2ea9a3c1b32e
@@ -96,12 +96,6 @@ cryptography==2.6.1 \
     --hash=sha256:d4afbb0840f489b60f5a580a41a1b9c3622e08ecb5eec8614d4fb4cd914c4460 \
     --hash=sha256:d9ed28030797c00f4bc43c86bf819266c76a5ea61d006cd4078a93ebf7da6bfd \
     --hash=sha256:e603aa7bb52e4e8ed4119a58a03b60323918467ef209e6ff9db3ac382e5cf2c6
-enum34==1.1.6 \
-    --hash=sha256:2d81cbbe0e73112bdfe6ef8576f2238f2ba27dd0d55752a776c41d38b7da2850 \
-    --hash=sha256:644837f692e5f550741432dd3f223bbb9852018674981b1664e5dc339387588a \
-    --hash=sha256:6bd0f6ad48ec2aa117d3d141940d484deccda84d4fcd884f5c3d93c23ecd8c79 \
-    --hash=sha256:8ad8c4783bf61ded74527bffb48ed9b54166685e4230386a9ed9b1279e2df5b1 \
-    # via argon2-cffi, cryptography
 flask-assets==0.12 \
     --hash=sha256:6031527b89fb3509d1581d932affa5a79dd348cfffb58d0aef99a43461d47847
 flask-babel==0.11.2 \
@@ -116,10 +110,6 @@ flask-wtf==0.14.2 \
 flask==1.0.2 \
     --hash=sha256:2271c0070dbcb5275fad4a82e29f23ab92682dc45f9dfbc22c02ba9b9322ce48 \
     --hash=sha256:a080b744b7e345ccfcbc77954861cb05b3c63786e93f2b3875e0913d44b43f05
-ipaddress==1.0.22 \
-    --hash=sha256:64b28eec5e78e7510698f6d4da08800a5c575caa4a286c93d651c5d3ff7b6794 \
-    --hash=sha256:b146c751ea45cad6188dd6cf2d9b757f6f4f8d6ffb96a023e6f2e26eea02a72c \
-    # via cryptography
 itsdangerous==0.24 \
     --hash=sha256:cbb3fcf8d3e33df861709ecaf89d9e6629cff0a217bc2848f1b41cd30d360519 \
     # via flask
```

PR #4548, opened 2019-06-25
Replace "Cloud" upload icon with more neutral one

### LINKED ISSUE #4535: Update "Upload" icon in Source UI
## Description

Per #2508 and #1536, the upload icons were changed; first, to a datastore icon, and later to a cloud icon. The design rationale to add the datastore icon was to avoid use of cloud semiotics in the icon, which can scare users. The datastore icon confused users, however, so another cloud icon replaced it.

This issue is to replace the latest cloud icon with a more common upload icon that both speaks to the needs flagged in #1536, while also considering the user feedback that the (admittedly obscure) datastore icon was unclear.

## User Needs Evidence

See above

## Implementation

The attached files were created from the PNGs currently used in the Source UI. Same size, same placement of the art, same whitespace—so ideally, only the filenames would need to be changed in the code. 
- [Big arrow PNG](https://drive.google.com/open?id=1dXGEFu6QTx0yUowGKBVNAnHE_VzDBEyu)
- [Blue arrow PNG](https://drive.google.com/open?id=1ppo8iibYk6lu_KnzCLUD8Qo-0Vmz6Qmd)
- [White arrow PNG](https://drive.google.com/open?id=1exp_FQDC_XBgP1lz_ikEpQIq-tQV-6NY)

eloquence:
Adds rendering rule for the icon to accomplish two things:
  - give the icon room to breathe
  - force crisp rendering, to render large source image cleanly at smaller size (the source PNG is a bit larger than the default size to accommodate zooming)
    
The crisp rendering rule may make sense as a generic rule at a later point, but so far this seems to be the best candidate for using it.
    
The new icon was designed by Nina Alter (@ninavizz) and is not derived from FontAwesome, hence not in the `font-awesome` directory.

Resolves #4535

## Status

Ready for review

## Test plan

1) Verify the icons render as expected (in staging and development) **in Tor Browser in safest mode**.
2) Zoom into each page to double-check that zooming doesn't produce unexpected pixelation

The screens should look as follows:

Index page (default state):
index-page-normal

Index page (hover state):
index-page-hover

Upload screen:
upload-screen


install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
```diff
@@ -267,7 +267,7 @@
   /var/www/securedrop/static/i/font-awesome/trash-black.png r,
   /var/www/securedrop/static/i/font-awesome/black/guard.svg r,
   /var/www/securedrop/static/i/font-awesome/white/guard.svg r,
-  /var/www/securedrop/static/i/font-awesome/white/exclamation-circle.svg r,  
+  /var/www/securedrop/static/i/font-awesome/white/exclamation-circle.svg r,
```
eloquence: (My editor cleans up trailing wihtespace by default.)

PR #4551, opened 2019-06-25
Provide full MDS mitigations and move linkcheck to weekly cron

### LINKED ISSUE #4520: Update SecureDrop kernels to 4.4.181+
SecureDrop instances are currently running 4.4.177 grsecurity-patched kernels.

To benefit from the most recent round of security improvements, we should update to the latest grsecurity kernel version in the 4.4 series (as of this writing: 4.4.181).

emkll:
## Status

Ready for review

## Description of Changes

Fixes #4520 

Provides Kernel options to enable full MDS mitigations and disable SMT (see https://www.kernel.org/doc/html/latest/x86/mds.html#kernel-internal-mitigation-modes)
## Testing
- [ ] We want to move docs-linkcheck to a weekly cron job
This requires hardware testing - VM testing might return failures, if libvirt configuration does not pass proper parameters (see https://docs.openstack.org/nova/latest/admin/mitigation-for-Intel-MDS-security-flaws.html)
- build `securedrop-grsec-4.4.182-amd64.deb` (securedrop kernel metapackage) on this branch and install it on a hardware instance that has apt-test.freedom.press as sources.list (needs 4.4.182 kernel images)
- [ ] Download and run the script provided at https://meltdown.ovh . All tests should pass
- [ ] Kernel can boot in the previous kernel (4.4.177-grsec)
## Deployment

New and existing installs will be updated via deb packages (securedrop-grsec package)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,7 +71,7 @@ jobs:
 
       - run:
           name: Run documentation linting
-          command: make docs-lint && make docs-linkcheck
+          command: make docs-lint
 
   app-tests:
     machine:
@@ -328,6 +328,22 @@ jobs:
             if ! [[ $BRANCH_MATCH =~ ^found ]]; then echo "Skipping: ${BRANCH_MATCH}"; exit 0; fi
             make ci-deb-tests
 
+  docs-linkcheck:
+    machine:
+      enabled: true
+    steps:
+      - checkout
+      - *rebaseontarget
+      - *installenchant
+
+      - run:
+          name: Install development dependencies
+          command: pip install -U -r securedrop/requirements/develop-requirements.txt
+
+      - run:
+          name: Run documentation linting
+          command: make docs-lint && make docs-linkcheck
+
 workflows:
   version: 2
   securedrop_ci:
@@ -422,3 +438,4 @@ workflows:
     jobs:
       - deb-tests
       - translation-tests
+      - docs-linkcheck
--- a/install_files/securedrop-grsec/etc/default/grub.d/50-mds-smt.cfg
+++ b/install_files/securedrop-grsec/etc/default/grub.d/50-mds-smt.cfg
@@ -1,0 +1,3 @@
+# disables smt and provide full mds mitigations
+# see https://www.kernel.org/doc/html/latest/admin-guide/hw-vuln/mds.html
+GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX mds=full,nosmt"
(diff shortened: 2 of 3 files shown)


.circleci/config.yml
```diff
@@ -422,3 +438,4 @@ workflows:
     jobs:
       - deb-tests
       - translation-tests
+      - docs-linkcheck
```
conorsch: Note to other reviewers: `docs-linkcheck` cron settings are unrelated to the kernel updates, but sure are nice to have. :slightly_smiling_face:

install_files/securedrop-grsec/etc/default/grub.d/50-mds-smt.cfg
```diff
@@ -0,0 +1,3 @@
+# disables smt and provide full mds mitigations
+# see https://www.kernel.org/doc/html/latest/admin-guide/hw-vuln/mds.html
+GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX mds=full,nosmt"
```
conorsch: The `nosmt` settings may have a modest performance impact, but it appears acceptable: restricting the application processes to the number of cores (i.e. 2 on most recommended setups) for simultaneous execution shouldn't be a significant performance bottleneck, especially considering the other limitations such as network and disk IO. Even though we don't have quantitative metrics on application behavior prior to merge, it's worth flagging for consideration during QA. The typical submit, check, reply, check workflow should work just fine, with no appreciable slowdown. 

The rationale for disabling SMT is that it will help guard against future vulnerabilities in the same class. If the past 18 months have been any indication, we're likely not past the reactive cycle based on disclosures, so taking a more defensive posture, even at the cost of a slight performance hit, is well warranted.

molecule/testinfra/staging/common/test_grsecurity.py
```diff
@@ -192,3 +192,15 @@ def test_wireless_disabled_in_kernel_config(host, kernel_opts):
...
+def test_mds_mitigations_and_smt_disabled(host):
+    """
+    Ensure that full mitigations are in place for MDS
+    see https://www.kernel.org/doc/html/latest/admin-guide/hw-vuln/mds.html
+    """
+
+    grub_config_path = "/boot/grub/grub.cfg"
+    grub_config = host.file(grub_config_path)
+
+    assert grub_config.contains("mds=full,nosmt")
```
conorsch: Ideally we'd check `/sys/devices/system/cpu/vulnerabilities/mds`, but checking the mitigations in VMs won't give us an accurate picture about the mitigations on hardware, so fine as-is.

conorsch: Similarly, we'll continue to run the spectre/meltdown checker script manually during QA. If we added it to the test run here, we'd still see failures in VMs. Worth revisiting after the release if manual script runs are showing green on hardware.

PR #4555, opened 2019-06-26
Create debian package based on dh-virtualenv

kushaldas:
## Status

Work in progress

This has one commit for the `debian` directory for packaging and one for changes required to create a source tarball.


conorsch:
The diff presented here copy/pastes several version-controlled files, rather than using `git mv` to preserve history. As such, it's not a candidate for merge, but should still be a useful point of reference for @rmol in addressing #4355. In particular, take a look at `MANIFEST.in`—although it references new filepaths specific to this diff, the files referred to are an approximate list of what the package must include. Make sure to revise `setup.py` to remove `securedrop-proxy` references, including CLI scripts.

As far the commands you'll need to run in order to kick off the packaging workflow started here, see the build logic in the SDW packaging repo for examples:

* https://github.com/freedomofpress/securedrop-debian-packaging/blob/f32fec970a64ca0a6abfde52ca12fad042b70ea9/.circleci/config.yml
* https://github.com/freedomofpress/securedrop-debian-packaging/blob/f32fec970a64ca0a6abfde52ca12fad042b70ea9/scripts/build-debianpackage#L81

Just as with the SDW packaging logic, consider creating an ad-hoc command that you can use to rebuild just this package repeatedly during development and debugging. Don't plan to reuse the Molecule-based container build envs while iterating unless you see a pressing need to.

redshiftzero:
Superseded by #4622

PR #4557, opened 2019-06-27
Update SecureDrop screenshots for source, journalist and admin guides

### LINKED ISSUE #4529: Screenshots in source guide are out of date
The screenshots in the [source guide](https://docs.securedrop.org/en/release-0.13.0/source.html) are out of date (0.8.0rc1), which is leading to some inaccuracies:

- The "Tor browser" instructions will be updated as part of 0.14.0 (#4494), but the screenshots still reflect the old instructions, including an old screenshot of the security slider

- The "Flagged for reply" language was updated due to inaccuracies (#4420), but the language here is still the old language

There are smaller issues, such as the icon mismatch for submission (icon was updated in #2508).

Let's update these screenshots consistent with the current state of `develop`, ideally in time for 0.14.0 so we can push out updated Tor instructions both in docs and code at the same time.

eloquence:
This also includes a correction in the docs that describe a screenshot, which refers to the "Exit" button (it was once briefly called that, but is now back to "Log out").

Resolves #4529

## Status

Ready for review

## How to review

I recommend using GitHub's rich diff feature to ensure that the content of each screenshot matches the one it replaces. Click the "rich diff" icon for each file on the "Files changed" tab:

Screenshot from 2019-06-27 21-48-41

## Checklint

- [x] `make docs-lint` passed locally


eloquence:
I used @rmol's [sceenshots fixes](https://github.com/rmol/securedrop/commit/6f493c745a8e0532c7c112f2a5949627875ec0fd) to work around a bug in `develop` where all screenshots by the functional tests produce tofu instead of legible characters. It saved a fair bit of time, but I still had to manually crop several screenshots, and retook some to capture the right target area.

Note I forced the version string to 0.14.0 for a cleaner reading experience.

Even with John's fixes, my first run for the admin UI screenshots produced a lot of screens without the logo so I'm not including those in this PR; will do another run to see if that's flakiness or consistent.

eloquence:
With the changes in #4559 and more manual editing/re-taking of some screenshots, I was able to update the admin guide screenshots as well.

PR #4569, opened 2019-07-01
Removed units from the <img> tag

### LINKED ISSUE #4549: Incorrect (but compatible) width/height attributes in image markup
## Description

Many of the `` tags in SecureDrop's templates look like this:

```html
<img class="icon" src="some-image.png" width="20px" height="14px">
```

To the CSS-trained eye, this may look perfectly correct, and indeed it is rendered without problems as most rendering engines will happily ignore anything after the integer. However, this is technically not valid HTML, and https://validator.w3.org/ will report it as an error ("Bad value 14px for attribute height on element img: Expected a digit but saw p instead."). 

For `width` and `height` attributes, the value should be a non-negative integer without a unit (as pixels are always the unit). To avoid surprises, let's use valid markup throughout.

nightwarrior-xxx:
## Status

Ready for review

## Description of Changes

Fixes #4549 

## Testing

As suggested by @eloquence I have remove units from the image tag.

### DIFF
--- a/securedrop/journalist_templates/config.html
+++ b/securedrop/journalist_templates/config.html
@@ -23,7 +23,7 @@
 <p>{{ gettext('Here you can update the image displayed on the SecureDrop web interfaces:') }}</p>
 
 <p>
-  <img src="{{ url_for('main.select_logo') }}" class="logo small" alt="SecureDrop" width="250px">
+  <img src="{{ url_for('main.select_logo') }}" class="logo small" alt="SecureDrop" width="250">
 </p>
 
 <form method="post" enctype="multipart/form-data">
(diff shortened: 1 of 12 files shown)


nightwarrior-xxx:
@kushaldas Please take a look when you are free.

securedrop/journalist_templates/config.html
```diff
@@ -33,7 +33,7 @@ <h2>{{ gettext('Logo Image') }}</h2>
     <br>
   </p>
   <h5>
-    {{ gettext('Recommended size: 500px * 450px') }}
+    {{ gettext('Recommended size: 500 * 450') }}
```
eloquence: no reason to remove units here

nightwarrior-xxx:
@eloquence Lint is failing again :(

eloquence:
> Lint is failing again :(

Can you rebase your PR against the current state of `develop` in the main repo to see if that fixes things?

nightwarrior-xxx:
@eloquence Merging conflicts when I tried that. looks like I have to clone the whole repo again .

Issue #4128, opened 2019-02-13
Switch to keys.openpgp.org as default keyserver


eloquence:
The keyserver ecosystem is extremely flaky and poorly maintained. It routinely causes problems during setup and updates as `--recv-key` operations time out unpredictably (most recently #4100).  The keyserver ecosystem is also vulnerable to cert flooding attacks.

Per the discussion below, we want to:
- Switch to `keys.openpgp.org` as a well-maintained alternative to other keyserver implementations
- Support WKD as a fallback key distribution method in future (can be tracked in separate issues)

*2019-07-01: Updated per discussions below - @eloquence*

emkll:
I think this makes sense, and is already widely used by various project to host the keys used to sign apt packages.

There are two docs-level recommendations we should make:

1 - Since we are specifying a URL to an arbitrary file and not a fingerprint to import, we must guard against unwanted key changes, either due to error or the file being changed on server or in transit. We must validate the fingerprint prior to importing it to the gpg keyring (either programmatically for updates, or manually for first install):

```
gpg {--with-fingerprint, --with-colons} key.asc
```

2 - Any user that wants to validate signatures/web of trust would need to `gpg --refresh-key`

eloquence:
@emkll Thanks, that makes sense.

> gpg {--with-fingerprint, --with-colons} key.asc

What's the purpose of `--with-colons` here? It seems that the output without that option is easier to read when the verification step is performed by a human.

> 2 - Any user that wants to validate signatures/web of trust would need to `gpg --refresh-key`

Do you see this as necessary to document? If so, who's the target audience and how would they act on this information?

KwadroNaut:
1. I agree with @emkll here. The --with-colons is great for copy-pasting and scripts. Relying on human verification is a pretty bad idea since it's not too hard to create keys where the first and last couple of characters are the same. We, humans, tend to ignore those in the middle.

2. I'd suggest that admins do this, every time they do a manual upgrade or during troubelshooting.

eloquence:
WKD seems worth considering as a non-keyserver method for publishing the release key which is supported within recent versions of gpg as one of the `--auto-key-locate` methods.

https://wiki.gnupg.org/WKD

See draft spec: https://www.ietf.org/id/draft-koch-openpgp-webkey-service-08.txt

Being considered in gpgsync as well: https://github.com/firstlookmedia/gpgsync/issues/157

redshiftzero:
[...]

PR #4585, opened 2019-07-03
Modified CLI update function to use keys.openpgp.org for release key

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Towards #4128.

Updates the securedrop-admin tool to use only hkps://keys.openpgp.org when retrieving the release key for tag signature verification.

## Testing

On an Admin Workstation (virtual or USB stick):
- [ ] Verify that the update process works when a release key is already in the GPG keychain, using the following commands in a terminal:
```
cd ~/Persistent/securedrop
gpg --recv-key "22245C81E3BAEB4138B36061310F561200F4AD77"   # use Tails' default keyserver
git remote add zenmonkeykstop https://github.com/zenmonkeykstop/securedrop.git
git checkout openpgp-admin-cli-update
./securedrop-admin update
git status    # should now be on 0.13.1 tag
```
- [ ] Next, verify the process works with no pre-existing key:
```
gpg --delete-keys "22245C81E3BAEB4138B36061310F561200F4AD77"
git checkout openpgp-admin-cli-update
./securedrop-admin update
git status    # should now be on 0.13.1 tag
```

## Deployment
This will be deployed via the GUI updater, CLI update as above, or directly by checking out the release tag containing the change. Nothing special required.

### DIFF
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -683,13 +683,18 @@ def update(args):
                          'Good signature from "SecureDrop Release Signing ' +
                          'Key <securedrop-release-key@freedom.press>"']
         bad_sig_text = 'BAD signature'
-        # To ensure that an adversary cannot name a malicious key good_sig_text
-        # we check that bad_sig_text does not appear and that the release key
-        # appears on the second line of the output.
         gpg_lines = sig_result.split('\n')
+
+        # Check if any strings in good_sig_text match against gpg_lines[]
+        good_sig_matches = [s for s in gpg_lines if
+                            any(xs in s for xs in good_sig_text)]
+
+        # To ensure that an adversary cannot name a malicious key good_sig_text
+        # we check that bad_sig_text does not appear, that the release key
+        # appears on the second line of the output, and that there is a single
+        # match from good_sig_text[]
         if RELEASE_KEY in gpg_lines[1] and \
-                len([s for s in gpg_lines if
-                    any(xs in s for xs in good_sig_text)]) == 1 and \
+                len(good_sig_matches) == 1 and \
                 bad_sig_text not in sig_result:
             # Finally, we check that there is no branch of the same name
             # prior to reporting success.
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -40,6 +40,7 @@ from pkg_resources import parse_version
 
 sdlog = logging.getLogger(__name__)
 RELEASE_KEY = '22245C81E3BAEB4138B36061310F561200F4AD77'
+DEFAULT_KEYSERVER = 'hkps://keys.openpgp.org'
 EXIT_SUCCESS = 0
 EXIT_SUBPROCESS_ERROR = 1
 EXIT_INTERRUPT = 2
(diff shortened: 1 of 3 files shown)


kushaldas:
> It is still failing for me. Here is the output from the manual command.
> 
> ```
> $ gpg --batch --no-tty --recv-key --keyserver hkps://keys.opengpg.org 22245C81E3BAEB4138B36061310F561200F4AD77
> gpg: keyserver receive failed: Server indicated a failure
> ```

It seeems there was typo in my manual command, but, the `./securedrop-admin update` also failed a few times. I will update with the error message next time.

emkll:
It might make sense to add a test case for the updated key (with UID) the introduced in #4578, a test similar to https://github.com/freedomofpress/securedrop/blob/develop/admin/tests/test_securedrop-admin.py#L193 with different output, containing the new style key :

```
gpg: Signature made Thu 20 Jul 2017 08:12:25 PM EDT
gpg:                using RSA key 22245C81E3BAEB4138B36061310F561200F4AD77
gpg: Good signature from "SecureDrop Release Signing Key" [unknown]
gpg:                 aka "SecureDrop Release Signing Key <securedrop-release-key@freedom.press>" [unknown]
```

In cases where the key is directly imported or the key is refreshed, the (good) signature output should be as above. The existing case should cover non refreshed key.

kushaldas:
```
INFO: Update needed
INFO: Verifying signature on latest update...
gpg: keyserver receive failed: Connection closed in DNS
ERROR (run with -v for more): Command '['timeout', '45', 'gpg', '--batch', '--no-tty', '--recv-key', '--keyserver', 'hkps://keys.openpgp.org', '22245C81E3BAEB4138B36061310F561200F4AD77']' returned non-zero exit status 2

```

The error message.

redshiftzero:
@emkll makes a good point, [parametrizing](https://docs.pytest.org/en/latest/parametrize.html) `test_update_signature_verifies` to cover:

* the git output already in that test
* the git output @emkll flags above
* the following git output:

```
gpg: Signature made Thu 20 Jul 2017 08:12:25 PM EDT
gpg:                using RSA key 22245C81E3BAEB4138B36061310F561200F4AD77
gpg: Good signature from "SecureDrop Release Signing Key <securedrop-release-key@freedom.press>" 
``` 

would be good to add prior to merge

admin/securedrop_admin/__init__.py
```diff
@@ -667,29 +668,28 @@ def update(args):
...
+                         'Good signature from "SecureDrop Release Signing ' +
+                         'Key <securedrop-release-key@freedom.press>"']
         bad_sig_text = 'BAD signature'
         # To ensure that an adversary cannot name a malicious key good_sig_text
         # we check that bad_sig_text does not appear and that the release key
         # appears on the second line of the output.
         gpg_lines = sig_result.split('\n')
         if RELEASE_KEY in gpg_lines[1] and \
-                sig_result.count(good_sig_text) == 1 and \
+                len([s for s in gpg_lines if
```
emkll: To me, these two new lines introduced appear complex, especially for such a critical part of the verification code. Do you think it makes sense to clarify the logic here (either through comments or making the logic more explicit)?

zenmonkeykstop: I reorganized things to make it a little clearer and added comments.

admin/securedrop_admin/__init__.py
```diff
@@ -667,29 +668,33 @@ def update(args):
...
+                         'Good signature from "SecureDrop Release Signing ' +
+                         'Key <securedrop-release-key@freedom.press>"']
         bad_sig_text = 'BAD signature'
-        # To ensure that an adversary cannot name a malicious key good_sig_text
-        # we check that bad_sig_text does not appear and that the release key
-        # appears on the second line of the output.
         gpg_lines = sig_result.split('\n')
+
+        # Check if any strings in good_sig_text match against gpg_lines[]
+        good_sig_matches = [s for s in gpg_lines if
```
redshiftzero: This is looking good - what do you think about doing `good_sig_matches = set(gpg_lines).intersection(good_sig_text)` here for readability (and leaning on the standard library to do the intersection)?  `good_sig_text` can remain a list as `set.intersection` can take any iterable

zenmonkeykstop: I did this originally, but it means good_sig_text[] has to contain the entire line for each element, which kindof changes how it behaves. Maybe for the better, but my concern was that it might be brittle if the gpg output varies a little.

redshiftzero: hmm yeah this is a good point, ok let's leave as is

PR #4586, opened 2019-07-03
Update Admin Workstation setup instructions to use keys.openpgp.org

eloquence:
Includes instructions for importing the key from securedrop.org as a fallback (may be superseded by WKD fallback in future)

## Status

Ready for review 

## Test plan

Test all modified or added commands in Tails 3.14.2 with an empty keyring, and ensure that they behave as described in the documentation.

- [x] I have done this myself using a Tails 3.14 VM

## Description of Changes

Soldiering on towards #4128

## Checklint
- [x] `make docs-lint` is not sad

### DIFF
--- a/docs/set_up_admin_tails.rst
+++ b/docs/set_up_admin_tails.rst
@@ -61,55 +61,70 @@ Start by running the following commands to download the git repository.
 Verify the Release Tag
 ~~~~~~~~~~~~~~~~~~~~~~
 
-.. caution:: Do not skip this step as this steps validates the files
-             in your Git repository.
+.. important::
 
-First, download and verify the **SecureDrop Release Signing Key**.
+   It is crucial for the integrity of your installation that you carefully
+   follow the instructions below. By following these steps, you will verify
+   if your copy of the codebase has been approved by the SecureDrop
+   development team.
+
+Download and verify the **SecureDrop Release Signing Key** using the following
+command:
 
 .. code:: sh
 
-    gpg --recv-key "2224 5C81 E3BA EB41 38B3 6061 310F 5612 00F4 AD77"
+   gpg --keyserver hkps://keys.openpgp.org --recv-key \
+    "2224 5C81 E3BA EB41 38B3 6061 310F 5612 00F4 AD77"
 
-.. note:: It is important you type this out correctly. If you are not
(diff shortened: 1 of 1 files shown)


docs/set_up_admin_tails.rst
```diff
@@ -61,55 +61,70 @@ Start by running the following commands to download the git repository.
...
-   option to the ``gpg --recv-key`` command. In our experience, the SKS HKPS
-   keyserver pool is usually a reliable alternative, so try:
+   If the command still fails, the *keys.openpgp.org* keyserver may be down.
+   In that case, we recommend downloading the key from the SecureDrop website:
 
    .. code:: sh
 
-      gpg --keyserver hkps://hkps.pool.sks-keyservers.net --recv-key "2224 5C81 E3BA EB41 38B3 6061 310F 5612 00F4 AD77"
+      cd ~/Persistent
+      torify curl -LO https://securedrop.org/securedrop-release-key.asc
```
eloquence: Note: This redirect has never been used before; I created it today (using Wagtail's redirect feature), as the official URL (https://securedrop.org/documents/10/securedrop-release-key.asc) is more difficult to type reliably, and subject to change if a new version is uploaded (in which case we'd have to redirect anyway, as we're doing from https://securedrop.org/documents/1/securedrop-release-key.asc). Happy to use an alternative approach if preferable.

docs/set_up_admin_tails.rst
```diff
@@ -61,55 +61,70 @@ Start by running the following commands to download the git repository.
...
+   yet:
 
-.. _Tails is connected to Tor: https://tails.boum.org/doc/anonymous_internet/tor_status/index.en.html
+   .. code:: sh
 
-When passing the full public key fingerprint to the ``--recv-key`` command, GPG
-will implicitly verify that the fingerprint of the key received matches the
-argument passed.
+      gpg --with-fingerprint --import-options import-show --dry-run \
+        --import securedrop-release-key.asc
```
eloquence: The short version is `gpg securedrop-release-key.asc`, but that results in a warning ("gpg: WARNING: no command supplied.  Trying to guess what you mean ..."), which suggests to me that we shouldn't rely on this guessing behavior. This same command is also used in #4522.

PR #4592, opened 2019-07-04
ci: export junit XML test metadata

### LINKED ISSUE #3799: export application test metadata for circleCI
## Description

Upon test failures the [output](https://circleci.com/gh/freedomofpress/securedrop/17716) is a bit annoying to look through. At some point we had application [test metadata](https://circleci.com/docs/2.0/collect-test-data/) consumed by CircleCI. Let's bring that back for easy parsing of test results

redshiftzero:
## Status

Ready for Review

## Description of Changes

Fixes #3799

 - Circle CI decides how to split tests across the containers via various methods: we've been using [timings](https://circleci.com/docs/2.0/parallelism-faster-jobs/#splitting-by-timings-data). Circle gets these timings from test result metadata and _should_ iteratively improve the test distribution across the containers based on the run time of the tests. However, our test metadata has been broken for some time (this is bug #3799). This PR fixes the test metadata export.
 - Another reason it's nice to have this metadata exported as the failed test results populate a dashboard at https://circleci.com/build-insights (need to be a project contributor) which can be used to see at a glance which tests are flaky and which tests are the slowest. 

## Testing

Verify by clicking on an app-test Circle CI build on this PR that the "Test Summary" area is now populated and you now see information about the run time of our tests: 

Screen Shot 2019-07-03 at 7 30 03 PM

## Deployment

CI only

## Checklist

CI changes only

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -143,10 +143,6 @@ jobs:
       - *p2savecache
 
       - run:
-          name: Make test results directory
-          command: mkdir -p ~/test-results
-
-      - run:
           name: Run tests
           no_output_timeout: 20m
           command: |
@@ -157,10 +153,10 @@ jobs:
             PYTHON_VERSION=2 DOCKER_RUN_ARGUMENTS=$(bash <(curl -s https://codecov.io/env)) DOCKER_BUILD_ARGUMENTS="--cache-from securedrop-test-xenial-py2:${fromtag:-latest}" make test
 
       - store_test_results:
-          path: ~/test-results
+          path: ~/project/test-results
 
       - store_artifacts:
-          path: ~/test-results
+          path: ~/project/test-results
 
   python3-app-tests:
     machine:
@@ -181,10 +177,6 @@ jobs:
       - *savecache
 
       - run:
-          name: Make test results directory
-          command: mkdir -p ~/test-results
-
-      - run:
           name: Run tests
           command: |
             BRANCH_MATCH=$(devops/scripts/match-ci-branch.sh "^(i18n|docs|update-builder)")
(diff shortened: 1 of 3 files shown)


redshiftzero:
Hmm so I reran this a couple of times experimentally: the timings data is definitely available in Circle CI but `circleci tests split` is not balancing the tests across the containers as well as I would expect, e.g. in the long running container all these tests are being ran, including the very slow `tests/pageslayout/test_journalist.py`:

```
circleci@default-e742c04c-3bda-4717-ae1d-93f7c9980fc4:~/project/securedrop$ circleci tests glob 'tests/test*py' 'tests/**/test*py' | circleci --verbose tests split --split-by=timings
tests/functional/test_admin_interface.py
tests/functional/test_source_notfound.py
tests/pageslayout/test_journalist.py
tests/test_alembic.py
tests/test_i18n.py
tests/test_integration.py
tests/test_journalist_utils.py
tests/test_qa_loader.py
tests/test_store.py
INFO[2019-07-05T17:04:42Z] Clean exit without errors
```

which means we have one container that takes 28 minutes, and the other four finish _way_ faster (~< 15 minutes). It's unclear to me why this is happening.

redshiftzero:
Reworked this to _just_ contain the fixing of the junit XML metadata (no point in increasing number of parallel test runners if the tests are not well balanced across them)

rmol:
Checked the Circle builds, confirmed timing is present. We're definitely producing the junit.xml in the specified location.

Perhaps we could smooth out the uneven distribution you saw with more `parallelism`. I think I remember you saying that under our current plan, we can run as many containers as we want, just paying for the time? If that's the case, maybe we should go the other way and really crank dat up. :slightly_smiling_face: 

Another problem with splitting by timing, though, is the erratic runtime of tests in the CI environent. In the [python3-app-tests](https://circleci.com/gh/freedomofpress/securedrop/30699#tests/containers/0) we see this:
```
245.30s call     functional/test_admin_interface.py::TestAdminInterface::test_admin_deletes_user
```
That test consistently takes around ten seconds on my machine, most of which is setup; the actuall call is around two seconds. We can't expect much from the timing splitter with input like that.

I wonder if we'd be better off splitting by filename, and breaking down our bigger test files for more parallelism. We could explicitly isolate tests we know to be slow.

PR #4594, opened 2019-07-05
Remove outdated upgrade documentation

### LINKED ISSUE #4547: Consider removing or archiving outdated upgrade documentation
## Description

We should consider removing or archiving outdated SecureDrop and Tails upgrade documentation. While it is useful to have these docs available for posterity sake, it clutters up the TOC (and thus the sidebar) of the documentation, which can be overwhelming for new admins/potential admins. As an example, see the screenshot appended to this ticket.

We have precedent for doing this in #1750.

## Motive

Why the removal or archival? If an admin is using an older version of SecureDrop (esp. really old versions), they should not be encouraged to upgrade over several releases to the latest version. Rather, they should backup what they need (if any) and do a full reinstall.

Similarly with Tails drives, it is unrealistic to expect an admin or a journalist to upgrade to the latest version of Tails from the 1.x series. They would similarly be advised to backup and reinstall the latest version as appropriate.

## Potential fix

I would aim for complete removal, since one could view older versions of the documentation by selecting what version they'd like to view from the documentation itself. (We have precedent for this, as mentioned earlier.)

If that's undesirable, then I would advocate creating a separate TOC section entitled "Outdated Documentation" or something similar and moving these pages down there. It would keep the relevant bits of documentation clean.

If we decide to remove or archive the documentation, we should set a policy for how many upgrade documents would be appropriate to keep in the "current documentation". My initial stab at a policy would be to keep the last three releases.

This policy could be enshrined in the documentation guidelines.

## User Stories

As an admin or a journalist, I might be overwhelmed by the sheer number of articles available, particularly in the sidebar.

As an admin of a severely outdated instance, I might consider the availability of upgrade documentation to mean there is implicit support for upgrading from very old versions, rather than the much more appropriate "upgrade" method: to perform a backup and reinstall.

## Screenshot

Screenshot 2019-06-22 at 16 22 29

edenemmanuel:
## Status

Ready for review

## Description of Changes

Resolves #4547.

There is a lot of documentation for upgrading SecureDrop instances, SecureDrop workstations, or the server OS. Most of it is outdated and unsupported. It's wise to remove them and clean up any references.

### Checklist


edenemmanuel:
Thanks for the review @zenmonkeykstop!

I could certainly do a closer dive through the docs to check any references in prose referencing previous versions of SecureDrop. Not sure where we'd put information on accessing older versions of the documentation, but it could be something like a separate page or a note in the admin guide or similar.

What are your thoughts? Happy to add a commit.

zenmonkeykstop:
I may be  looking for you to gild the lily here - references to older versions outside the upgrade section are no big deal.

PR #4596, opened 2019-07-07
Updated logs playbook to clean up unencrypted tarballs

### LINKED ISSUE #4593: securedrop-admin logs command should clean up logs from admin workstation and server after task completion
## Description

When running `./securedrop-admin logs`, the unencrypted log tarball is stored in the user's home directory, root owned with a+r permissions. Both encrypted and non-encrypted log tarballs are also stored in install_files/ansible-base.

The `./securedrop-admin logs` command should clean up the unencrypted tarballs from both the admin workstation as well as both app and mon servers' home directories.

## User Stories

As an administrator, I do not want old logs to use up space nor to be available in the clear to a non-privileged user

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #4593 

Updates `securedrop-logs.yml` to delete unencrypted tarballs from the servers and Admin Workstation.

## Testing
On a VM or HW production instance, via the Admin Workstation:
- [ ] verify that `./securedrop-admin logs` completes without error
- [ ] verify that the unencrypted tarballs do not exist in `~/Persistent/securedrop/install_files/ansible-base`
- [ ] verify that the unencrypted tarballs do not exist in the server admin home directory for either the Application or Monitor Server.
- [ ] if you have access to the necessary key, verify that the encrypted tarballs can be decrypted.

## Deployment
Deployed when workstation USBs are updated.

### DIFF
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -41,6 +41,8 @@ from pkg_resources import parse_version
 sdlog = logging.getLogger(__name__)
 RELEASE_KEY = '22245C81E3BAEB4138B36061310F561200F4AD77'
 DEFAULT_KEYSERVER = 'hkps://keys.openpgp.org'
+SUPPORT_ONION_URL = 'http://support6kv2242qx.onion'
+SUPPORT_URL = 'https://support.freedom.press'
 EXIT_SUCCESS = 0
 EXIT_SUBPROCESS_ERROR = 1
 EXIT_INTERRUPT = 2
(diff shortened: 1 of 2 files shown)


zenmonkeykstop:
@kushaldas I didn't change the encryption key - it should be the one with fingerprint `734F6E707434ECA6C007E1AE82BD6C9616DABB79`

kushaldas:
> @kushaldas I didn't change the encryption key - it should be the one with fingerprint `734F6E707434ECA6C007E1AE82BD6C9616DABB79`

I think we found a bug, @conorsch according to my system that is not the `support` key, but, the other key we use. Can you please confirm?

PR #4597, opened 2019-07-08
Remove units from the image tag

### LINKED ISSUE #4549: Incorrect (but compatible) width/height attributes in image markup

nightwarrior-xxx:
## Status

Ready for review

## Description of Changes

Fixes #4549 

Removed units from the image tag

### DIFF
--- a/securedrop/journalist_templates/base.html
+++ b/securedrop/journalist_templates/base.html
@@ -33,7 +33,7 @@
       <div class="container">
         {% block header %}
         <div id="header">
-          <a href="{{ url_for('main.index') }}"><img src="{{ url_for('main.select_logo') }}" class="logo small" alt="SecureDrop" width="250px"></a>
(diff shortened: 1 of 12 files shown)


nightwarrior-xxx:
@eloquence Please take a look :)

eloquence:
@nightwarrior-xxx Looks good to me, leaving for a maintainer to merge. :)

PR #4607, opened 2019-07-09
Fix i18n_tool's list-translators function

rmol:
## Status

Ready for review

## Description of Changes

I needed to update it for my change to the commit format at 0.13.0, and the commit range examined for recent contributions needed to be improved to look until the last sync from Weblate.

## Testing

- Check out this branch.
- Run `make list-translators`. The output should only contain the contributors since the last Weblate sync (0aa921908). You can compare the list for a given language with a command like:
```
git log --format="%aN" 0aa921908..i18n/i18n securedrop/translations/ca/LC_MESSAGES/messages.po | sort | uniq
```
(Substitute the language you want to check for `ca` in the path.)

- Run `make list-all-translators`. Each language's contributor list should match the output of this:
```
git log --format="%aN:%s" i18n/i18n securedrop/translations/nl/LC_MESSAGES/messages.po | grep "Translated using Weblate" | sort | uniq | awk -F: '{print $1}'
```
Again, replace `ca` with the code you're trying to check.

## Deployment

The `list-translators` function should not affect deployment, but it's possible, so all tests should pass.

### DIFF
--- a/securedrop/i18n_tool.py
+++ b/securedrop/i18n_tool.py
@@ -284,20 +284,20 @@ class I18NTool(object):
             if updated:
                 self.upstream_commit(args, code)
 
-    def translators(self, args, path, range):
+    def translators(self, args, path, commit_range):
         """
         Return the set of people who've modified a file in Weblate.
 
         Extracts all the authors of translation changes to the given
-        path in the given range. Translation changes are identified by
-        the presence of "Translated using Weblate" in the commit
-        message.
+        path in the given commit range. Translation changes are
+        identified by the presence of "Translated using Weblate" in
+        the commit message.
         """
         translation_re = re.compile('Translated using Weblate')
 
         path_changes = git(
             '--no-pager', '-C', args.root,
-            'log', '--format=%aN\x1e%s', range, '--', path,
+            'log', '--format=%aN\x1e%s', commit_range, '--', path,
             _encoding='utf-8'
         )
         path_changes = u"{}".format(path_changes)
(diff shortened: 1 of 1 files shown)


kushaldas:
Tests timedout, rerunning.

kushaldas:
I can see the `replies` id as the test is expecting, digging more. 

i_can_see_replies_id

kushaldas:
If I put a `time.sleep(5)` before that assertion, it goes to next and fails again.

```

/home/kdas/code/securedrop/securedrop/tests/functional/test_source.py:20: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <tests.functional.test_source.TestSourceInterface object at 0x7fc8f6c48550>

    def _source_sees_no_codename(self):
        codename = self.driver.find_elements_by_css_selector(".code-reminder")
        time.sleep(60)
>       assert len(codename) == 0
E       AssertionError

```

kushaldas:
If I check manually then I don't see any codename, as shown in the screenshot below.

no_codename

But, the VNC for the test shows that there is a codename in the same page.

why_codename

zenmonkeykstop:
Output of provided git command doesn't match output of make:

- for `make list-translators`, Thalia Rahme is listed under `ar` but using git commands is listed under `ca`
- for `make list-all-translators` there are several other listings that are swapped or just appear in either make or git output.

rmol:
Thanks @zenmonkeykstop.  I think the `list-all-translators` discrepancies might be due to i18n_tool.py looking at the `.po` files for both the app and the desktop icons -- I only provided an example git command for checking the app `.po` file. So for example, one person contributed to the Spanish desktop icon translation (`install_files/ansible-base/roles/tails-config/templates/es_ES.po`), so shows up in `list-all-translators`, but wouldn't appear in the history of `securedrop/translations/es_ES/LC_MESSAGES/messages.po`.

I cannot explain how you saw Thalia Rahme under Catalan, though. I can't see a contribution from her to anything but Arabic in the git history. Could you leave the git command you used here?

zenmonkeykstop:
That checks out as an explanation - I was using the given git commands, concatenating results for all languages. Not able to reproduce the Arabic/Catalan discrepancy now, going to put it down to an issue with the concatenation. Good to go.

PR #4608, opened 2019-07-09
Add pytest flaky plugin

### LINKED ISSUE #4604: Add flaky decorator to all page layout tests
## Description

As discovered trying to rely on our `translation-test` CI job for merge of translations, we ran into frequent flakes in the page layout tests. We should add the [@flaky decorator](https://github.com/box/flaky) such that failures are automatically re-run.

We previously did this to reduce failures in the `securedrop-admin` tests in https://github.com/freedomofpress/securedrop/pull/4004 and https://github.com/freedomofpress/securedrop/pull/4466 so a similar approach should work nicely.

rmol:
## Status

Ready for review

## Description of Changes

Adds the pytest [flaky](https://github.com/box/flaky) plugin to automatically rerun failed tests. Adding it to the functional test classes with three retries seems to solve all of our famous flakes. We'll see how it does in CI.

I went with `flaky` over [pytest-rerunfailures](https://github.com/pytest-dev/pytest-rerunfailures) because the latter warns that it's not compatible with class-, module-, or package-level fixtures, which we're trying to move toward with #3836.

Fixes #4604.

## Testing

Check out the branch, run the tests a few hundred times. It's the only way to be sure.

## Deployment

This should only affect tests and test requirements.

### DIFF
--- a/securedrop/tests/test_alembic.py
+++ b/securedrop/tests/test_alembic.py
@@ -10,8 +10,6 @@ from alembic.script import ScriptDirectory
 from os import path
 from sqlalchemy import text
 
-from . import conftest
-
 from db import db
 from journalist_app import create_app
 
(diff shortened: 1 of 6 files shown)


rmol:
Unbelievable.

securedrop/tests/test_alembic.py
```diff
@@ -104,9 +102,7 @@ def test_alembic_head_matches_db_models(journalist_app,
     '''
     models_schema = get_schema(journalist_app)
 
-    config.DATABASE_FILE = config.DATABASE_FILE + '.new'
```
zenmonkeykstop: Had to ask about this change - apparently the updated version of pytest required only allows fixtures as decorators, hence the simplification.

zenmonkeykstop:
LGTM, all tests passing locally and in CI!

PR #4633, opened 2019-07-19
Support simultaneous dev containers

rmol:
## Status

Ready for review

## Description of Changes

Teach dev-shell to use a prefix for ports exposed in the dev containers, when necessary. This permits running multiple container-based tasks (anything run with dev-shell in the Makefile) simultaneously.

For example, if you run `make dev` with no other containers running, VNC will be on 5909, the source interface on 8080, and the journalist interface on 8081. If you then run `make test` while that container is running, the test container's services should be available at 15909, 18080, and 18081. Obviously the number of containers we can run with this simple scheme is limited, but probably sufficient.

## Testing

- Check out this branch.
- Run `make dev`.
- In another shell, run `make test`. You should see a notice that the container used for running tests will be offering services on different ports, e.g.:
```
************************************************************
Exposed Docker ports will be available on localhost prefixed
with "1", e.g. 18080 for the source interface, and 
18081 for the journalist interface.
************************************************************
```

The tests should run, and you should be able to interact with the dev servers normally.

## Deployment

This should only affect development activities using `dev-shell`.

### DIFF
--- a/securedrop/bin/dev-shell
+++ b/securedrop/bin/dev-shell
@@ -11,17 +11,15 @@ export PATH="/opt/venvs/securedrop-app-code/bin:$PATH"
 TOPLEVEL=$(git rev-parse --show-toplevel)
 BASE_OS=xenial
 
-function docker_image() {
-
 ## Get a short identifier for the working directory
 hashpath() {
     pwd | sum | awk '{print $1}'
 }
 
-## Get an integer prefix for exposed ports, to support running dev and
-## test, or containers from different directories simultaneously
+## Get an integer prefix for exposed ports, to support multiple containers
+
 get_port_prefix() {
-    port_prefix="${PORT_PREFIX:-}"
+    port_prefix="${PORT_PREFIX:-1}"
     tries=0
     while true
     do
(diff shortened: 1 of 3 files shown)


securedrop/bin/dev-shell
```diff
@@ -6,53 +6,91 @@
...
            -e PAGE_LAYOUT_LOCALES \
+           -e REPOROOT \
            --user "${USER:-root}" \
-           --volume "${TOPLEVEL}:${TOPLEVEL}" \
-           --workdir "${TOPLEVEL}/securedrop" \
-           --name securedrop-dev \
-           -ti ${DOCKER_RUN_ARGUMENTS:-} "securedrop-test-${1}-py${2}" "${@:3}"
+           --volume "${REPOROOT}:${REPOROOT}" \
+           --workdir "${REPOROOT}/securedrop" \
+           --name "sd-$(uuid)" \
```
zenmonkeykstop: The executable `uuid` wasn't present on my debian-9 dev VM, it had to be installed for the second container to be triggered. Is there any reason not to use, say, $port_prefix here instead?

rmol: Nope. It works find with the port prefix.

rmol:
I can't repro the inability to interrupt the container. I was seeing that a few weeks ago, outside the context of this PR, but Docker's been updated a couple of times since (on my Debian 9 system). Maybe check that it's current on your system.

rmol:
I've added a change to set the Flask ports from environment variables, but this means that what I'd intended as a small dev-only convenience turns into something that affects production. Not a huge deal, so if you're inclined, take a look and let me know if you think the extra usability is worth it. If so, we should probably table this until it can be scheduled properly.

zenmonkeykstop:
Maybe one way to avoid the production code change would be to explicitly set the ports per-Makefile target (`dev` always runs on 8080/8081, `test` on 18080/18081, random invocations of dev-shell on 2* or higher). That we we get multiple dev docker instances, but with the manage.py output still being accurate.

zenmonkeykstop:
Rebased this and rolled back the manage.py changes, opting instead to set it up to always have the `make dev` container use (5909,8080-8081) and other dev-shell spawned containers use higher ports (18080, 28080 etc). Test plan is unchanged.

rmol:
Thanks for resurrecting this one. With the latest revision, I was able to concurrently start several things: the dev server, a couple of test runs, translation tests, translator listings, and an update of pip requirements. 

When you get to ports above 64k, Docker throws this error:
```
docker: invalid publish opts format (should be name=value but got '127.0.0.1:65909:5909').
```
It's just bad error messaging, which was [fixed a few days ago](https://github.com/docker/cli/pull/2251). It's probably not useful to be running that many dev-shell things concurrently anyway; my machine was slowing down. 

I'd say :shipit: but we could also rework the port numbering scheme. What do you think?

zenmonkeykstop:
Updated port numbering scheme so successive containers use a port offset of 100 - first make test is 8180, next is 8280 etc. Now you can go completely nuts on containers :laughing: .

rmol:
The new port numbering scheme looks good. Numerous containers, VNC, everything worked in my testing. Now we just need a third party to approve.

Issue #2244, opened 2017-09-01
[docs] Make Tails upgrade documentation more generic


garrettr:
We currently have some excellent documentation for specific scenarios where we expect SecureDrop users will need to manually upgrade their Tails USBs:

* [Upgrade Tails from 1.x to 2.x](https://docs.securedrop.org/en/stable/upgrade_to_tails_2x.html)
* [Upgrade Tails from 2.x to 3.x](https://docs.securedrop.org/en/stable/upgrade_to_tails_3x.html)

However, we lack a _generic_ guide for manually upgrading Tails. Manual upgrades are also necessary in the following scenarios:

* Admins who upgraded their internet-connected Tails USBs (Admin/Journalist Workstation) from Tails 3.0 to Tails 3.0.1, which had a [bug](https://labs.riseup.net/code/issues/13426) that broke subsequent automatic upgrades. The official [fix](https://tails.boum.org/news/rescue_3.0.1/index.en.html) required a manual upgrade.
* In every case, the Secure Viewing Station (SVS) Tails must be manually upgraded, because the SVS is airgapped and unable to receive automatic upgrades over the network.
    * Clearly documenting this upgrade flow is especially important in the context of #2238. The initial RCE exploit was in the Nautilus file manager, and while it has been fixed upstream, it still [needs to be backported to Tails](https://labs.riseup.net/code/issues/14584). Once the backport is done, we expect a new version of Tails will be released, at which point SecureDrop administrators/journalists will need to manually upgrade their SVS' Tails to patch it against this specific vulnerability.

redshiftzero:
A contributor that wants to pick this documentation ticket up can start from branch `generic-tails-manual-upgrade-docs` as there is lots of work to build from in there.

eloquence:
Since this issue was written, we've formalized the habit of providing Tails upgrade instructions with each release of SecureDrop ([example](https://docs.securedrop.org/en/release-0.14.0/upgrade/0.13.1_to_0.14.0.html#updating-the-tails-workstations)). Where an upgrade has to be performed manually, we state so explicitly.

This enables admins to perform many SecureDrop-related maintenance tasks in one batch, in regular intervals. However, it still makes sense to add a general "Keeping all Tails workstations up-to-date" guide, especially for the air-gapped SVS. If we had a generic manual upgrade guide, we could regularly remind people to update their SVS in our release-to-release upgrade docs.

[...]

PR #4657, opened 2019-08-02
[docs] Overhaul Transfer Device and export recommendations

### LINKED ISSUE #3598: Explain risks of transferring information using USBs (breaking the airgap)
## Description

From `docs/journalist.rst`:

```rst
.. todo:: This document recommends transferring documents from the 
          *SVS* to the *Journalist Workstation*, without any
          discussion of the potential risks or mitigations that should
          be taken when doing so. A section needs to be added on why
          doing this could be risky, and what can be done to make this
          situation better.
```

### LINKED ISSUE #4434: Guidance in docs re: remembering Transfer Device passphrase is outdated
The [setup docs for the Transfer Device](https://docs.securedrop.org/en/release-0.12.2/set_up_transfer_device.html) instruct the administrator to use the "Remember Password" feature on each workstation, so they don't have to enter it by hand or store it in their password manager.

Unfortunately, persisting GNOME keyring data was [removed](https://redmine.tails.boum.org/code/issues/15275) in Tails 3.9. Keys persisted before then continue to work, but on any workstation set up since then, these instructions won't work.

The "Remember password" checkbox still appears; it just won't actually remember the passphrase after the Tails session is complete.

There's an upstream issue, https://redmine.tails.boum.org/code/issues/16718 , to consider reinstating that functionality; for now, we should update the docs in line with current behavior.

### LINKED ISSUE #4620: [docs] Switch Transfer Device recommendation to VeraCrypt
Our current expected SecureDrop workflow is:

**Journalist Workstation [Tails] → Transfer Device [USB or CD/DVD] → Secure Viewing Station [Tails] → Transfer Device [USB or CD/DVD] → Everyday Workstation [Windows/Mac/Linux]**

We currently recommend encrypting the Transfer Device with LUKS. LUKS is not supported on Windows ([LibreCrypt](https://github.com/t-d-k/LibreCrypt) is unmaintained) or Mac.

VeraCrypt is a better cross-platform option and one we'll likely adopt for the Qubes-based SecureDrop Workstation. We should also adopt it here so that this workflow actually makes sense.

### LINKED ISSUE #4646: [docs] Strengthen printer usage recommendations throughout
The [latest version of the Journalist Guide](https://docs.securedrop.org/en/latest/journalist.html#encrypting-and-moving-documents-to-your-everyday-workstation) addresses many issues with previous versions (e.g., it makes malware risks more explicit), but it still is fundamentally structured on the assumption that we want journalists to move files off the SVS in electronic form.

We should much more clearly highlight printing as preferred method to get documents off the SVS. Compared with all other methods, it has the strongest overall security properties (especially because it virtually eliminates malware risks). It is obviously not a fit for all use cases, but for documents that can be managed in print form, we want to encourage news organizations to consider it wherever possible.

We should also strongly encourage purchase of a non-WiFi-enabled, Tails-compatible printer in the [pre-install hardware checklist](https://docs.securedrop.org/en/latest/checklists/pre_install_hardware.html), and strengthen other references (e.g., right now we only say that "it is often useful" to have a printer).

### LINKED ISSUE #4670: Contradictory SVS data flow descriptions in diagrams
The data flow for moving information off the SVS is described inconsistently in these two diagrams:

- [overview diagram](https://docs.securedrop.org/en/release-0.14.0/_images/SecureDrop.png )
- [data flow diagram](https://docs.securedrop.org/en/release-0.14.0/_images/SecureDrop_DataFlow.png)

Per the work in #4657, we'll likely want to standardize on the version in the overview diagram, making it clear that both drives are encrypted, and referring to them using standardized nomenclature (*Transfer Device* and *Export Device*).

eloquence:
## Status

Ready for review

## Description of Changes

- More clearly prioritizes printing as the recommended method to get documents off the SVS whenever feasible;

- Introduces a recommended workflow using a VeraCrypt-encrypted *Export Device* intended to be clear and consistent, while still closely resembling current practices;

- Removes instructions to set up per-journalist GPG keys;

- Updates diagrams consistent with the above.

Resolves #4620
Resolves #4646
Resolves #4434
Resolves #4670
Resolves #3598

## Review plan

Review the documentation against the issues below. Does it now provide instructions to administrators that are clear, sensible, and consistent? Please note the significant updates to the diagrams in the documentation.

The biggest change for end users is the recommendation to provision a USB drive encrypted with VeraCrypt as the *Export Device*. I have done this and used it both in Tails and on my regular Linux workstation. I would recommend that the reviewer similarly provision a VeraCrypt drive and attempt to access it in a recent version of Tails and on their everyday workstation (bonus points for testing on Mac or Windows).

I have tested that VeraCrypt successfully mounts read-only drives, both using the application, and using the graphical integration in Tails. On the command line, cryptsetup must be invoked with the `--readonly` option for this to work.


eloquence:
Still a lot to do here, but I wanted to get the current WIP up for any early feedback. Most of this should be consistent with our recent discussions. Couple of areas where I think it could be useful for FPF to do a bit more hardware testing, if we haven't already done so:

- USB write blockers like [this one](https://www.cru-inc.com/products/wiebetech/usb_writeblocker/) seem like a relatively cost effective way to prevent infecting drives that may be taken back to the SVS. 

- VeraCrypt is still pretty cumbersome and slow, and a hardware encrypted drive honestly seems more user-friendly and practical -- maybe even for the *Transfer Device* and the SecureDrop Workstation.  If we can recommend such an option, it would seem to potentially bypass a lot of usability and cross-platform issues. There are hardware-encrypted NitroKeys, which are advertised as fully open source & open hardware w/ third party audits. If these hold up under scrutiny, it would be nice to be able to officially recommend them.

eloquence:
As a side note, the recommendation in this PR (to use separate USB drives for Export and Transfer Device) is in fact exactly identical to the architecture in [this diagram](https://docs.securedrop.org/en/release-0.14.0/_images/SecureDrop.png) where these are called "USB Key A" and "USB Key B". The only problem is that "USB Key A" and "USB Key B" are only referred to in this diagram and nowhere else in the docs.

docs/hardware.rst
```diff
@@ -13,9 +13,43 @@ Hardware Overview
...
+.. include:: includes/pre-install-hardware-optional.txt
+
+In the sections that follow, we provide additional details on most of these
+items.
+
+.. tip::
+
+    While a printer is not required, we highly recommend it. Printing documents
+    is generally far safer than copying them in digital form. See our
+    :doc:`journalist guide <journalist>` for more information.
```
conorsch: The journalist guide is large. Let's add a section link to the "Working with Documents" within the Journalist Guide to aid the reader a bit here.

docs/hardware.rst
```diff
@@ -192,7 +226,12 @@ Offline Printer
 ^^^^^^^^^^^^^^^
 
 It is often useful to print submissions from the *Secure Viewing Station* for
-review and annotation.
+review and annotation. Printing eliminates most of the risks associated with
+malware, aside from malicious QR codes that a journalist may be tempted to scan.
+It also removes invisible metadata that may identify the source of a given
+document (with the exception of `printer tracking information <https://www.eff.org/issues/printers>`__
+included in a scanned document, or similar forms of metadata that may not be
+easily visible or recognizable).
```
conorsch: Great URL; maybe worth adding to the Journalist Guide -> Working with documents, as well, particularly this passage:

> Printing a file is often safer than exchanging it electronically, and it has the additional benefits of removing embedded metadata (except for printer codes, watermarks, or similar identifiers that may not be visible to the naked eye).

docs/passphrases.rst
```diff
@@ -85,3 +84,13 @@ journalist will require the following passphrases:
...
+
+*Transfer Device* and *Export Device*
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+As noted in the :doc:`setup guide <set_up_transfer_and_export_device>`,
+we recommend using encrypted USB drives for transferring files to and from the
+*Secure Viewing Station*. Journalists performing these operations will need
+access to the encryption passphrase. Because the same passphrase has to be used
+on the source and the destination, we recommend storing this passphrase in the
+journalist's own existing password manager (e.g., on their mobile phone).
```
conorsch: Nit: "source and destination" is somewhat confusing here, given the use of the term "source" to denote user roles in the SecureDrop workflow. Perhaps rephrase as:

> Because the same passphrase has to be used frequently on multiple workstations (i.e., *Journalist Workstation* and *Secure Viewing Station*)

As far as the recommendation for using an individual password manager, rather than Tails's own KeePassX setup, that mostly makes sense for the _Export_ device, not the _Transfer_ device—but fine with the change to keep things consistent. The "Passphrase Best Practices" guide makes no mention of using an individual password manager, however—might be worth updating that guidance in line with these changes.

docs/set_up_transfer_and_export_device.rst
```diff
@@ -0,0 +1,210 @@
...
+Because files are copied between multiple computers, KeePassX in Tails is not
+necessarily the most convenient option for managing the encryption passphrases for
+your *Transfer Device* or your *Export Device*. While Tails itself gives you the
+option to "remember" passphrases, this option does not work across reboots.
+
+A simple alternative is to make sure that every journalist stores the
+*Transfer Device* and *Export Device* passphrases in their own password manager,
+which ideally will synchronize to their smartphone. See the Freedom of the Press
+Foundation guide for `choosing a password manager <https://freedom.press/training/blog/choosing-password-manager/>`__
+if you are not currently using one.
```
conorsch: As above, the guidance on choosing a password manager also belongs in the "Passphrase Best Practices" guide, in an adjacent doc.

docs/set_up_transfer_and_export_device.rst
```diff
@@ -0,0 +1,210 @@
...
+
+VeraCrypt is a good alternative to hardware-based encryption. It is actively
+maintained cross-platform software that has been independently audited and is
+free to use.
+
+VeraCrypt-encrypted media can be opened in the Tails operating system without
+installing additional software. However, to *create* your encrypted VeraCrypt
+drive in the first place, you will need to install the VeraCrypt software. The
+`guide by Freedom of the Press Foundation <https://freedom.press/training/encryption-toolkit-media-makers-veracrypt-guide/>`__
+provides instructions for encrypting storage media using VeraCrypt.
```
conorsch: Even following the URL, it's not immediately obvious that the user should create the VeraCrypt volume _on a machine other than the Tails-based Workstation_ (whether _Admin_ or _Journalist_). A bit more expectation management is warranted here, e.g that it's acceptable to use the daily-use machine (non-Tails-based) to provision the VeraCrypt volumes.

docs/set_up_transfer_and_export_device.rst
```diff
@@ -0,0 +1,210 @@
...
+protection switch, but it may be more difficult to consistently remember or
+enforce its usage.
+
+Full-size SD cards also come with physical write protection switches. However,
+this write protection is fully host-based (the host operating system can choose
+to ignore it), and should therefore be considered less secure against
+sophisticated malware.
+
+Please see our :doc:`journalist guide <journalist>` for additional
+recommendations regarding malware mitigation.
```
conorsch: As above, a link to the "Working with Documents" section within the Journalist Guide would be great.

docs/hardware.rst
eloquence: Done.

docs/hardware.rst
eloquence: Added link to malware section.

docs/passphrases.rst
eloquence: Good point re: confusing language, tried to clarify that. 

Re: the Transfer Device, one other concern about stashing that in KeePassX is making sure that it's always updated on two devices, without any automatic mechanism to keep them in sync. If those two password managers are out of sync, the journalist may complete a copy operation halfway, only to find that they don't have access to the cred when they need it. 

If they pull it from their smartphone's password manager, it'll either work on the *Journalist Workstation* or it won't work at all, reducing the likelihood of wasted time. So I think more than just a matter of consistency, having only a single copy of the passphrase is generally the better recommendation.

For orgs with many journalists, having it in KeePassX twice may be a better option (even if that won't work for the *Export Device*), but our docs also suggest that in those cases, it may be good to have multiple Transfer Devices. This idea that the Transfer Device/Export Device is _personal_ to the journalist is I think worth exploring further (especially as an alternative to the impractical and half-hearted suggestion to use GPG), but perhaps out of scope for this PR.

docs/set_up_transfer_and_export_device.rst
eloquence: Added!

docs/set_up_transfer_and_export_device.rst
eloquence: Agreed; I've added some language suggesting that they may want to use an everyday workstation, or a more controlled environment, to set up VeraCrypt drives.

docs/set_up_transfer_and_export_device.rst
eloquence: Done!

eloquence:
9a5349a52cf6b954cd0effdbdf676251d3fac682 is potentially the single most contentious part of this PR. It removes all references to setting up individual GPG keys for journalists, and instead provides instructions for using VeraCrypt.

As a reminder, VeraCrypt replaces LUKS as the previous recommendation --  a recommendation which was impossible to follow in practice. The documentation effectively instructed journalists to double-encrypt files, with LUKS *and* with their personal keys, and to do so in a manner that is impossible to decrypt on their everyday workstation (due to LUKS not being cross-platform). That said, whether to copy to Tails or to their everyday workstation -- even that part was not clearly spelled out.

Our docs provided minimal instructions for working with GPG keys (instead linking out to the EFF guides), and suggested that the private key be stored on the *Journalist Workstation* -- which is defined in the glossary as the Tails-based workstation used to access the journalist interface. That makes little sense in the real world, where journalists will need to leave Tails behind pretty quickly.

An easier option than introducing VeraCrypt would be to make the previous GPG instructions a bit more sensible, and to ask journalists to use an ordinary, unencrypted USB drive to copy files off the SVS, encrypting them first to their public key. This may well be current practice in at least some organizations.

The downsides of that approach are significant, IMO:

- Relying on GPG as the standard way to encrypt data signs organizations not already committed to GPG up for a lot of complexity, compared with using an encrypted drive.

- Encrypting individual submissions while keeping the drive unencrypted keeps metadata (time/data of export, etc.) in unencrypted form. There's a high risk of accidentally decrypting submissions directly on the *Export Device*, or accidentally copying data in unencrypted form.

- An adversary gaining access to an unencrypted *Export Device* may be able to successfully recover "deleted" files.

For these reasons, I don't think it's even a good idea to suggest it as an alternative to the VeraCrypt approach in the docs.

If all of this leaves you with a bad feeling, I share the same concern. I'm not a fan of any of these solutions. VeraCrypt is non-free software (due to legacy TrueCrypt bits), difficult to use (a lot of complexity in the UI that should be hidden behind progressive disclosure patterns and sensible defaults), and a bit flaky: I've had it randomly freeze on me twice during testing.

In terms of cross-platform usability, I've not tested them personally, but it seems hard to beat an integrated hardware-encrypted solution like the [Kingston DT-2000](https://www.kingston.com/us/usb-flash-drives/datatraveler-2000-encrypted-usb-flash-drive) with on-device PIN entry. And of course there more open solutions like the NitroKey encrypted drives that may be worth considering.

But if we have to choose between the devil and the deep blue sea, I think just recommending VeraCrypt and purging GPG from the docs is our best long term bet, and consistent with our current plan of record to use VeraCrypt on the SecureDrop Workstation. Really, much of this complexity is inherent in working with an air-gapped system, and the Workstation will greatly reduce the end user complexity here.

Finally, I'll mention Jen's suggestion to get files off the SVS by way of Tails+OnionShare. Tails does have pretty nice built-in OnionShare integration, and this would allow us to use LUKS for both the *Transfer Device* and the *Export Device*. It does introduce a new layer of complexity -- downloading files from Tor on the journalist's everyday workstation, and keeping the OnionShare on Tails running while that's happening, plus the usual concerns about Tor speed and reliability.

docs/threat_model/threat_model.rst
```diff
@@ -392,7 +392,7 @@ What Compromise of the Admin's Property Can Surrender
       alerts.
    -  Access the credentials for the account the encrypt alerts are sent
       to.
-   -  Access the admin's personal GPG key.
+   -  Access the admin's personal GPG public key, if stored there.
```
eloquence: At least I'm assuming that we're talking about the public key. If we're not, we've not made this clear anywhere else in the docs (as far as I can tell, we make only one other passing reference to the admin's personal GPG key, in `passphrases.rst` above).

docs/onboarding.rst
```diff
@@ -142,26 +131,32 @@ verify the journalist is set up for SecureDrop.
 
 The journalist should verify that they:
 
-1. Have their own *Journalist Tails USB* that they have verified they are able
```
eloquence: This section repeatedly italicizes terms that are not actually glossary terms. It may make sense to standardize the usage of *Admin Workstation USB* and *Journalist Workstation USB* and add it to the glossary; for now, I've just standardized on existing terminology.

eloquence:
OK, I'm going to call this ready for review! I've tried to be as thorough as possible in updating all existing references in the docs to the workflow that's being described here, but it's possible that I have missed some. I will take a couple more passes myself.

Because this PR impacts our training practices, it should not be considered ready for merge unless and until @harlo gives her +1.

docs/includes/pre-install-hardware-required.txt
```diff
@@ -1,12 +1,13 @@
 * 2 computers with memory and hard drives to use as the SecureDrop servers.
 * Mouse, keyboard, monitor (and necessary dongle or adapter) for
   installing the servers.
-* Dedicated physical computers for the Admin, Journalist, and Secure Viewing
-  Station that can boot to Tails. At *minimum* this is 2 computers.
+* Dedicated physical computers for the *Admin Workstation*, *Journalist
```
emkll: Do you mean that the SVS should use a dedicated physical computer? It's unclear to me which computers would be used for which

docs/journalist.rst
```diff
@@ -469,45 +464,59 @@ choose **Wipe** to delete them.
...
-Alternatively, you can use the tools provided within Tails to examine
-documents, or convert files from one format to another (e.g., export a Word
-document as PDF). Fully mitigating the risks of malware is beyond the scope
-of this documentation.
+As long as you are using the latest version of Tails on your *Secure Viewing Station*,
+you can open submitted documents with a low risk of malicious files compromising
+it. Tails only preserves changes to your persistent volume across reboots,
+making it difficult for malware to take hold. The air-gap prevents malware
+from "phoning home". Keeping your *Secure Viewing Station* up-to-date is the
+single best way to protect its security.
```
emkll: Do we provide docs on keeping up to date? If so, it might be useful to provide a link here.

docs/journalist.rst
```diff
@@ -469,45 +464,59 @@ choose **Wipe** to delete them.
...
+
+If you copy files you have received via SecureDrop to your everyday workstation
+and open them there in their original form, the risks from malware are
+significantly greater. SecureDrop does not scan for or remove malware, and your
+everyday workstation likely has a much greater `attack surface <https://en.wikipedia.org/wiki/Attack_surface>`__
+than the *Secure Viewing Station*.
```
emkll: While this is also considered attack surface, the "phoning home" aspect might be important to reiterate here.

docs/diagrams/SecureDrop_DataFlow.draw
```diff
@@ -1 +1 @@
...WOM2S9KbjW2P5ZIcNLeCcHpKwP1ekXqARXcKaW89MG6Q13wC8ji2gA90xSXYFhj4/q+1BOsjHakAY/0KMcC4BM7yYcP5iekmGVNkFs6FLD2S2S01Fmc2+cwDTK+2n+gy45EDd7+Svpm2ol2dCnyK/X9uWyk8c4BvdqVM5tJtAtocaz+FzdHIYGryvoqfuquU4W8+01dbKSFOtkIRtiNGNfWKFIb0UogacAW1szGcrg2aF7Yg5hmNqV2uKcoNGcDbPHVg7aTwao5NdK9TOsuZU4VjA2MOBaM/8ZChXMkGUkfRiz9viFXFK1nKj0ShlORVUx2rjxOsNfBYZhwiqMw97ob0Ch0x5WfTkaTX3abmrNLazUJZk0DXpTrGVeS2wDuEZCLIyVDAYRHdhoj1hBZ2S663NMo4TBTXpHjNkW8vExgBiBPNBUKi66vKOzFka4qT6so24Q8/Z2X5Bx3IbmVKBYLM80KQ1kPVRz+Kj6KIdch1GptDyLXGMy3nZJGTkRSPsDolPA25F5ZLCjgVocIGiDarJX7V5RTeSqh4XJ8TGFWlP6H47O4WIrTdGBSkiYE0egdtitK4nXA9BRVd4uTmrZYEMlzQ7QPNlWsFNo7raeasc0+MudoemZeA98EISG+0dLcQTpuZdThUafadwIi4S2uF8I53n97Li+7PgaPwqB332vCAE2UrYwaMArW4/Nsf9IYNKbtWnN73TCVtoQiu9pEuZBA/hp5VbzfGUUpHQRB0T0PVOQ2uHtC9+EwE2Sc8m5yy77ugCs91AFk8evaqKSlYuJecXJ2QLk9S4p3HmWvPe+dKO3nWDzGsNpHl5f8HGPoQE/4juO+5mB5urte4UVt5wb/m7/FNm8EZpP26vdoTH3+QyfO9mAeqp9/XLAfAXyPd+U/kZoT5A+fm1qZKXY0zk1UEm/NT79UGCTAp6MWvw5hjzS4PVN+7zR0NM6gMUf1/ECQGRbwXbj4/iv4iyz1vqIyfwKBT9zhH6+3O8rwbi3eHr+/v8USAib+zy/wdyZq86mQUYlZ7jeHF9fg2exCM8P1r6jaTvCvr3tu71qvH+QNHvyS/fq+AnKb8fsckT58e7HFH5drvLHz28coe2F7D7CMB3Pv6H/TRN3/2N7vu+3q1d6+45n7sTX/1eX1//ViQuP3Y1qKb/fXnnNcmuDkJwxf8D</diagram></mxfile>
```
emkll: I think the Export device should be outside of the airgapped area to make it clear to a user that the devices transfers across environments.

docs/journalist.rst
```diff
@@ -519,44 +528,88 @@ by default, which may reveal sensitive information about your SecureDrop
...
+   running on your everyday workstation, copying it in its original form carries
+   the risk of spreading malware to that computer. Make sure you understand the
+   risks, and consider other methods to export the document (e.g., print).
+
+If you must copy a file from your *Secure Viewing Station* to your everyday
+workstation in digital form, our :doc:`recommendation <set_up_transfer_and_export_device>`
+is that journalists are provided with an *Export Device*, typically a USB drive,
+which is encrypted using `VeraCrypt <https://www.veracrypt.fr/en/Home.html>`__.
+These instructions assume that you are following the recommended workflow.
+If you are unsure, ask your administrator.
```
emkll: I think moving this a bit closer to the top, we want to ensure that admins in an org are aware and comfortable with journalists  opening those files on their everyday workstations.

docs/passphrases.rst
```diff
@@ -85,3 +85,16 @@ journalist will require the following passphrases:
 The backup that is created during the installation of SecureDrop is also
 encrypted with the application's GPG key. The backup is stored on the
 persistent volume of the Admin Live USB.
+
+*Transfer Device* and *Export Device*
```
emkll: We should probably reiterate/link to the printer suggestion here as a reminder

docs/hardware.rst
```diff
@@ -147,41 +178,44 @@ two-factor authentication:
...
-From a security perspective, it is preferable to use write-once media such as
-DVD/CD-R discs because it eliminates the risk of exfiltration by malware that
-persists on the Transfer Device (e.g. `BadUSB <https://srlabs.de/badusb/>`__).
-
-On the other hand, using write-once media to transfer data is typically
-inconvenient and time-consuming. You should consider your threat model and
-choose your transfer device accordingly.
+Transfer Device(s) and Export Device(s)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+You need away to transfer encrypted submissions from the *Journalist
```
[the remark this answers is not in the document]
eloquence: Fixed.

docs/hardware.rst
```diff
@@ -147,41 +178,44 @@ two-factor authentication:
...
-most common transfer devices are DVD/CD-R discs and USB drives.
-
-From a security perspective, it is preferable to use write-once media such as
-DVD/CD-R discs because it eliminates the risk of exfiltration by malware that
-persists on the Transfer Device (e.g. `BadUSB <https://srlabs.de/badusb/>`__).
-
-On the other hand, using write-once media to transfer data is typically
-inconvenient and time-consuming. You should consider your threat model and
-choose your transfer device accordingly.
+Transfer Device(s) and Export Device(s)
```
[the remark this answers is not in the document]
eloquence: Reorganized in 22d5af4deea4d3ab7ea5f0668abeb3ff7cbbb021

docs/hardware.rst
```diff
@@ -147,41 +178,44 @@ two-factor authentication:
...
+Workstation* to the *Secure Viewing Station* to decrypt and view them, and a
+way to copy them from the *Secure Viewing Station* to the journalist's
+everyday workstation in decrypted form.
+
+We recommend using separate USB drives, in combination with encryption and
+careful data hygiene. We also urge you to consider using a printer or similar
+analog conversions to export documents from the *Secure Viewing Station*,
+whenever possible.
+
+You may want to employ write protection to make it impossible for malware to
```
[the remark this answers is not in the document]
eloquence: 22d5af4deea4d3ab7ea5f0668abeb3ff7cbbb021 uses language which I think is defensible, let me know if you disagree:

> When it is consistently applied and correctly implemented in hardware, write
protection can prevent the spread of malware to your *Secure Viewing Station*
from the computers used to read files stored on a *Transfer Device* or an
*Export Device*.

docs/hardware.rst
```diff
@@ -192,7 +226,12 @@ Offline Printer
 ^^^^^^^^^^^^^^^
 
 It is often useful to print submissions from the *Secure Viewing Station* for
-review and annotation.
+review and annotation. Printing eliminates :ref:`most of the risks associated with
+malware <malware_risks>`, aside from malicious QR codes that a journalist may be tempted to scan.
+It also removes invisible metadata that may identify the source of a given
```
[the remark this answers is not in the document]
eloquence: Reworked this section a bit in 786f6e21cba63e36182903064cc91ec570779ccd to further strengthen the recommendation to get a printer and to clarify the benefits/remaining risks without duplicating content. (My main goal here is to make sure the advantages are stated sufficiently clearly, before we go into "but! QR codes" type exceptions to the rule.)

docs/hardware.rst
```diff
@@ -478,33 +514,23 @@ persistent partition. We recommend getting something in the 16-64GB
...
-If you are using DVD/CD-R's for the transfer device, you will need *two*
-DVD/CD writers: one for burning DVDs from the *Journalist
-Workstation*, and one for reading the burned DVDs on the *SVS*. We
-recommend using two separate drives instead of sharing the same drive to
-avoid the potential risk of malware exfiltrating data by compromising
-the drive's firmware. We've found the DVD/CD writers from Samsung and LG
-to work reasonably well, you can find some examples
+If you want to use a setup based on CD-Rs or DVD-Rs, we've found the CDR/DVD
+writers from Samsung and LG to work reasonably well; you can find some examples
 `here <http://www.newegg.com/External-CD-DVD-Blu-Ray-Drives/SubCategory/ID-420>`__.
```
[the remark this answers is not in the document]
eloquence: not mine but good catch, fixed in 1490900204d86758e2ec53781c92cc8f18102180

docs/diagrams/SecureDrop_DataFlow.draw
eloquence: Done in 82c9436ccab0d070d09d1b86a183c0102232891a (also moved the "printed documents" icon for the same reason).

docs/journalist.rst
eloquence: Unfortunately we do not; that's #2244 and something I'd love @rocodes to take a stab at once she's completed her first SecureDrop install.

docs/includes/pre-install-hardware-optional.txt
```diff
@@ -0,0 +1,16 @@
+* a printer without wireless network support, to use in combination with the
+  *Secure Viewing Station*.
+* an external hard drive to expand the storage capacity of the
```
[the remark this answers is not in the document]
eloquence: This "can be encrypted" wording felt very handwavy to me, tbh, and I'd rather kill it than repeat it. I think we should keep the procurement lists (like the one you're commenting on) lean, but in 33bb90b5a3bee38928b59a647ce29940b9d404ec I've added an include to the two sections in the hardware docs that refer to encrypted external drives, to make it clearer that encryption is a requirement, and to call out specific areas of hardware compatibility that may warrant upfront research prior to procurement. I'm naming hardware RAID controllers and hardware-based encryption schemas as compatibility risks; happy to expand on that as warranted. Does that make sense?

docs/includes/pre-install-hardware-optional.txt
```diff
@@ -0,0 +1,16 @@
...
+  *Secure Viewing Station*.
+* an external hard drive for server backups.
+* a USB drive to store backups of some or all of your Tails workstation drives,
```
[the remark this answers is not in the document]
eloquence: The process described in https://docs.securedrop.org/en/release-0.14.0/backup_workstations.html suggests using the air-gapped SVS computer to perform the backups, and operates strictly on the command line (hence low risk of accidentally opening compromised files). In 257cdff91ba667ec33e054ace9e329c02f2c791c I've added a link directly to this documentation to make it clearer how this USB drive is intended to be used.

docs/includes/pre-install-hardware-optional.txt
```diff
@@ -0,0 +1,16 @@
+* a printer without wireless network support, to use in combination with the
+  *Secure Viewing Station*.
+* an external hard drive to expand the storage capacity of the
+  *Secure Viewing Station*.
+* an external hard drive for server backups.
```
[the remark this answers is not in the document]
eloquence: As above, would prefer to focus on requirements that have procurement implications in these checklists.

docs/includes/pre-install-hardware-required.txt
eloquence: I don't think the old language was clear on this point either, but I've tried to clarify it in d7f913bb67c9eb4ee82d3f056b88f69542a345dc.

docs/journalist.rst
eloquence: Ro has a WIP PR for that now here: #4694. If it gets merged in time, will add a link.

docs/journalist.rst
eloquence: Done in 46ec379e91862a74a08b814ab1f51265170f6477.

docs/journalist.rst
eloquence: This document is the Journalist Guide, so I don't think we can assume that admins will ever read it. A fundamental challenge with having a single Journalist Guide is that it may include instructions that differ from the ones the organization is giving to its users.

In terms of alerting admins to the risk management aspect, I would suggest that we add some language to the overview section: https://docs.securedrop.org/en/release-0.14.0/overview.html

docs/passphrases.rst
eloquence: Done in 2f62f1321.

docs/hardware.rst
```diff
@@ -147,41 +178,44 @@ two-factor authentication:
...
-
-On the other hand, using write-once media to transfer data is typically
-inconvenient and time-consuming. You should consider your threat model and
-choose your transfer device accordingly.
+Transfer Device(s) and Export Device(s)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+You need away to transfer encrypted submissions from the *Journalist
+Workstation* to the *Secure Viewing Station* to decrypt and view them, and a
+way to copy them from the *Secure Viewing Station* to the journalist's
+everyday workstation in decrypted form.
```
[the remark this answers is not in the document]
eloquence: Reworded in 74b7e23e7

eloquence:
I think this is ready for another round of review; @rocodes would be great if you could take a pass before we kick it over to Harlo and the training team for comments. :)

docs/glossary.rst
```diff
@@ -149,21 +149,20 @@ authentication for devices. We recommend using one of:
...
-*SVS*. While doing so is cumbersome, it reduces the risk of malware (that
-could be run simply by opening a malicious submission) exfiltrating
-sensitive data, such as the private key used to decrypt submissions or
-the content of decrypted submissions.
-
-When we use the phrase "sneakernet" we mean physically moving documents
-with the *Transfer Device* from one computer to another.
+The *Transfer Device* is the physical media (e.g., designated USB drive) used
+to  transfer encrypted documents from the *Journalist Workstation* to the
+*Secure Viewing Station*. Both computers run the Tails operating system
```
rocodes: [Nit/typo: there are 2 spaces in "to transfer"]

I would take out the sentence about "both computers run.." because this is just the definition of the transfer device.

"The *Transfer Device* is the physical media (e.g., designated USB drive) used to transfer encrypted documents from the *Journalist Workstation* to the *Secure Viewing Station*, where they can be decrypted."

docs/includes/pre-install-hardware-required.txt
rocodes: Additionally, the admin workstation needs an ethernet port or a dongle that has one, and the admin-ws and SVS need at least 2 USB ports, probably 3 if you count for times you need to insert yubikeys/transfer devices/backup devices.  This probably means purchasing a usb hub (with an ethernet port), if not 2 hubs (one with ethernet and one without), in order to have a dedicated device for the SVS and a device you can use with the the admin-ws.   

This will become more and more relevant as "old" computers repurposed for workstations move farther into Dongleland and have fewer onboard ports.

rocodes: re emkll's point: what about

"At least 2 computers that can boot to Tails: a dedicated physical computer for the *Secure Viewing Station*, and one or more computers for the *Admin workstation*/*Journalist workstation(s).*"

docs/journalist.rst
```diff
@@ -469,45 +464,61 @@ choose **Wipe** to delete them.
...
+your digital security staff before the export. Fully mitigating the risks of
+malware is beyond the scope of this documentation.
+
+.. tip::
+
+  Tails provides both graphical and command-line utilities that can be used to
+  convert files across formats (e.g., PDF to PNG). This can be a good way to
+  dramatically reduce the risk of malware, especially when converting to the
+  simplest possible format that can hold the information you need.
+
```
rocodes: I found the beginning of this section a little bit confusing.  What about something like:

SecureDrop itself does not protect you from receiving malicious submissions, but there are steps you can take to minimize the risk of malware. 

Using an up-to-date version of Tails on your Secure Viewing Station is the safest way to open and assess the documents you receive. Tails OS has features that offer more protection against compromise than your everyday computer, and viewing documents on an air-gapped system stops potential malware from "phoning home." 

Whenever possible, print hard copies of documents from the Secure Viewing Station instead of exporting them in digital form. This stops you from transferring potentially malicious files, as well as limits metadata from those files. (Note that printing a document may still preserve watermarks, printer codes, steganographically encoded data, or other information not visible to the naked eye.)

If you must copy files in digital form (e.g., because of the format or quantity of information), we recommend consulting with your administrator or your digital security staff beforehand. Fully mitigating the risks of malware is beyond the scope of this documentation.

..tip:: Converting files to simpler formats (e.g., PDF to PNG) can help reduce the risk of malware. Tails provides both graphical and command-line utilities for this.

eloquence:
Thank you, @rocodes!

> I am a little concerned overall about the length of all of our documents/it being an overwhelming amount of material for users

It absolutely is overwhelming, and suffering from attempting to serve an audience ranging from the individual journalist to the corporation with thousands of employees. Given how much variability there is in how SecureDrop is realized in practice, it may be worth thinking about a doc system that's a bit more templatized than this "one size fits all" approach.

For now, my concern has mainly been to ensure that we don't say things that don't make sense or are inconsistent, so we have a baseline from which we can think more carefully about the information architecture and usability of this documentation.

docs/journalist.rst
rocodes: The changes look great! In my opinion the list is much clearer.

harlo:
I'm comfortable with this, @eloquence; thanks for strongly enumerating the caveats up-front. :+1:

huertanix:
I read through the VeraCrypt updates and I think they look great; Only two minor things that might be :cool: to add in the future.
* Adding something along the lines of "read out guide to learn about how password managers work and how to pick one" linking out to [our guide](https://freedom.press/training/blog/choosing-password-manager/) when mentioning using a personal/phone password manager to keep the Export key passphrase in.
* Mentioning, either in the instruction bullet point asking the user to open VeraCrypt on their everyday workstation or wherever it makes more sense, that they need to first install VeraCrypt on their everyday workstation. This also might be a place to link to the VeraCrypt downloads page and our [guide on verifying open source software](https://freedom.press/training/verifying-open-source-software/).

eloquence:
Thanks @huertanix, @harlo and @olivemartini (who sent comments privately), much appreciated :)

> Adding something along the lines of "read out guide to learn about how password managers work and how to pick one" linking out to our guide when mentioning using a personal/phone password manager to keep the Export key passphrase in.

This is currently linked to, here:

https://github.com/freedomofpress/securedrop/blame/docs-transfer-to-export/docs/set_up_transfer_and_export_device.rst#L89-L93

I tried to address this in a single place ("Decide how to manage passphrases") since it comes up a few times.

I agree re: the VeraCrypt requirement, will clarify.

eloquence:
@huertanix VeraCrypt docs clarified in 920ac10eb81865e15a65554b64fd1296aa1f511e; feedback appreciated if you have time :)

docs/set_up_transfer_and_export_device.rst
```diff
@@ -0,0 +1,230 @@
...
+audits.
+
+VeraCrypt is a good alternative to hardware-based encryption. It is actively
+maintained cross-platform software that has been independently audited and is
+free to use.
+
+VeraCrypt-encrypted media can be opened in the Tails operating system and on
+common Linux distributions without installing additional software. To open
+VeraCrypt media on Windows or Mac workstations, or to create VeraCrypt drives,
+you need to install the VeraCrypt software. The `guide by Freedom of the Press
```
zenmonkeykstop: It might be worth being explicit here and saying that you need to create the Veracrypt volumes off Tails. It's possible that someone would read this as saying you should install additional software on Tails to create Veracrypt volumes. We don't want to encourage this IMO. Users shouldn't be customizing their Tails distros without good reason.

docs/set_up_transfer_and_export_device.rst
```diff
@@ -0,0 +1,230 @@
...
+password manager of each user who will use a given *Export Device*.
+
+Limiting write access
+~~~~~~~~~~~~~~~~~~~~~
+If you re-use the same *Export Device* for multiple copy operations, there is
+the risk of introducing malware to the *Secure Viewing Station* from your
+network. Depending on your threat model, there are steps you may want to take to
+mitigate that risk.
+
+One option is to restrict write access to the *Export Device* before it is
```
zenmonkeykstop: As a recommendation, this seems a bit unbalanced. To derive benefit from it, an organization using digital forensics tools would need to also have a dedicated workstation for working with files, strong protocols around everything they're doing, and would need to be applying those to *all* submissions that they copy from the SVS. Outside of this context, it seems unrealistic and creates a barrier to entry for organizations that are not as up-to-speed and want to do SecureDrop the "right" way.

docs/threat_model/threat_model.rst
```diff
@@ -473,7 +473,6 @@ What a Compromise of the Journalist's Property Can Achieve
    -  Access the Hidden Service values used by the *Journalist Interface*.
    -  Access SSH keys and passphrases for the *Application Server* and the
       *Monitor Server*.
-   -  Access the journalist's personal GPG key.
```
zenmonkeykstop: Threat model doc refers to SD 0.3 - unclear if it should be changed piecemeal, or edited in its entirety to include references to other assets. This change adds an Export USB and recommends the use of a password manager on the journalist's mobile device, neither of which are covered here.

eloquence: Per discussion earlier today, I've made minimal updates to the threat model consistent with the new terminology and updated data flow in 60cc802531e97019bdc76c1723a0dde6eff66494.

docs/set_up_transfer_and_export_device.rst
eloquence: Per discussion today, prioritized the low-cost option of a USB with a physical switch in 60cc802531e97019bdc76c1723a0dde6eff66494.

docs/set_up_transfer_and_export_device.rst
eloquence: Added a note to the effect that it should not be installed on the SecureDrop drives in 60cc802531e97019bdc76c1723a0dde6eff66494. (IMO another Tails drive may be a fine choice.)

docs/set_up_transfer_and_export_device.rst
```diff
@@ -0,0 +1,230 @@
...
+*Secure Viewing Station*. This is even more important than for the
+*Transfer Device*, as the risk of accidentally leaving decrypted files on the
+*Export Device* is significant.
+
+Because the *Export Device* will need to be mounted on both Tails and the
+journalist's everyday workstation, you will need to use an encryption scheme
+that works on both operating systems.
+
+This is the case for many hardware-encrypted USB drives. When considering
+a hardware solution, we recommend selecting a vendor that has fully opened
```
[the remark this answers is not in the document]
eloquence: Per discussion today, more clearly prioritized use of VeraCrypt in 60cc802531e97019bdc76c1723a0dde6eff66494 until such a time that we can provide an official recommendation for a hardware-encrypted drive.

docs/set_up_transfer_and_export_device.rst
```diff
@@ -0,0 +1,230 @@
...
+We recommend using a fully encrypted USB drive for copying files off the
+*Secure Viewing Station*. This is even more important than for the
+*Transfer Device*, as the risk of accidentally leaving decrypted files on the
+*Export Device* is significant.
+
+Because the *Export Device* will need to be mounted on both Tails and the
+journalist's everyday workstation, you will need to use an encryption scheme
+that works on both operating systems.
+
+This is the case for many hardware-encrypted USB drives. When considering
```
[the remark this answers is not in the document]
eloquence: This should be clearer in 60cc802531e97019bdc76c1723a0dde6eff66494.

docs/set_up_transfer_and_export_device.rst
emkll: Another area where write protection would be useful is also for the transfer device (if it is not wiped after each transfer): Enable write protection when connecting to the  SVS to avoid data exfiltration data via the transfer device to the journalist workstation. 

Perhaps prefacing the section with "something to consider" or adding (optional) to the title of the section, making it clear that it's an optional recommendation, which we can strengthen once we've used/practiced these a bit more.

eloquence: It strikes me as a lower risk scenario (because both machines are on Tails, and because the Transfer Device is only ever meant to hold `.gpg` files), but will poke at the language a bit.

eloquence:
@emkll I've tried to clarify the potential use of write protection for the *Transfer Device* as a defense-in-depth strategy in 86db46154d9939185c52c90debe9d570cacc5645 (amended existing commit, which seems more readable in this case). 

Regarding the threat model, note that it speaks already to the point you're mentioning, as it notes that an attacker will have access to SecureDrop submissions if the *Export Device* is in use:

> View, modify, and delete decrypted submissions, if they are stored in decrypted form on the *Secure Viewing Station*, or if the *Export Device* is in use.

If the journalist follows the recommended procedure, this will not be the case even if the *Transfer Device* is in use. 

I think the Export Device / Transfer Device risks could be more clearly separated from the device-related risks, but I also think the way this information is organized right now could be improved, so I would recommend leaving that to a follow-up PR.

docs/set_up_transfer_and_export_device.rst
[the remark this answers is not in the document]
zenmonkeykstop: :heavy_check_mark:

docs/set_up_transfer_and_export_device.rst
zenmonkeykstop: looks good.

PR #4673, opened 2019-08-15
Remove obsolete and redundant Tails Guide

### LINKED ISSUE #4566: Tails Guide instructs users to use `develop` version of the codebase, without verification
The Tails guide here:

https://docs.securedrop.org/en/release-0.13.1/tails_guide.html

includes instructions for setting up the `securedrop` repo on a Tails workstation. However, it does not include the download of the release key, the checkout of a specific tag, or the signature verification.

I've not tested if the graphical updater will succeed under these conditions to switch the user to a signed tag on the next update, but if they do not update, they'll be stuck on the ` develop` version of this codebase, which could cause issues attempting to use the `tailsconfig` command.

This may fail for other reasons: the guide does not include instructions for copying the required hidden service definitions to ensure that the `tailsconfig` command can use them.

This guide is also very unclear as to who its target audience is (admins provisioning Tails sticks? journalists?). We should carefully examine whether it should be consolidated with other existing guides. If not, to avoid such inconsistencies, let's consider using re-usable text blocks for instructions that must be repeated.

eloquence:
This guide was written with Tails 2 in mind (e.g., it refers to the old greeter screen), and does not instruct the user to perform proper verification steps when checking out the SecureDrop codebase. It is redundant with instructions elsewhere in the docs.

Resolves #4566

## Status

Ready for review

## Checklint

- [x] docs happily lintin'


docs/passphrases.rst
```diff
@@ -11,7 +11,9 @@ each role in a SecureDrop installation.
           password manager included in Tails, to generate and retain
           strong and unique passphrases. We have created a template
           passphrase database that you can use to get started. For more
-          information, see the :doc:`tails_guide`.
+          information, see the :ref:`KeePassX setup instructions <keepassx_setup>`,
+          which are identical for the *Admin Workstation* and the
+          *Journalist Workstation*.
```
redshiftzero: We don't want the admin to set up the keepass database once and then copy it to the journalist drives - this will leave the journalists with many admin credentials - can we add a note/warning to this effect?

eloquence: I've tried to clarify the process in the journalist onboarding doc, where I think admins are most likely to see it (it is the primary section of the docs that describes the process of creating new *Journalist Workstations*). See 5cb607c9a6e6ef07c67b0937f22ddc40497e3614 - how does that look to you?

PR #4678, opened 2019-08-16
Tighten Content Security Policy

evilaliv3:
With this pull request I propose to restrict the existing content security policy to the strictly necessary.

Existing policy:
```Content-Security-Policy: "default-src 'self'"```

Proposed policy:
```default-src 'none'; script-src 'self'; style-src 'self'; img-src 'self'; font-src 'self';```

Comment:
The current policy restricts any resource to be loaded from 'self' (current domain / same proto); this reduces the attack surface but still leaves open many kind of resources that are not used by the application;  for example ```object-src``` and ```connect-src``` could be set to 'none'.
This result could be achieved with the proposed policy that first block any kind of resource and then enables only a limited set of resources.

### DIFF
--- a/docs/deployment/landing_page.rst
+++ b/docs/deployment/landing_page.rst
@@ -168,7 +168,7 @@ If you use Apache, you can use these:
     Header set X-Content-Type-Options: nosniff
     Header set X-Download-Options: noopen
     Header set X-Permitted-Cross-Domain-Policies: master-only
-    Header set Content-Security-Policy: "default-src 'self'"
+    Header set Content-Security-Policy: "default-src 'none'; script-src 'self'; style-src 'self'; img-src 'self'; font-src 'self';"
     Header set Referrer-Policy "no-referrer"
 
 If you intend to run nginx as your webserver instead, this will work:
@@ -183,7 +183,7 @@ If you intend to run nginx as your webserver instead, this will work:
     add_header X-Content-Type-Options nosniff;
     add_header X-Download-Options: noopen;
     add_header X-Permitted-Cross-Domain-Policies master-only;
-    add_header Content-Security-Policy "default-src 'self'";
+    add_header Content-Security-Policy "default-src 'none'; script-src 'self'; style-src 'self'; img-src 'self'; font-src 'self';"
     add_header Referrer-Policy "no-referrer";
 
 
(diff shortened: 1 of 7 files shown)


evilaliv3:
I've now fixed all the unit tests in relation to the change.

The updater-gui-tests  seems to have failed for reasons independent from the pull request.

zenmonkeykstop:
`updater-gui-tests` seems flaky at the moment. I've kicked CI for your PR, thanks for fixing the unit tests!

eloquence:
Thanks again for this PR @evilaliv3. We're currently very focused on landing v3 Onion Service support and preparing for the 1.0.0 release, so it may still take us a couple of weeks to review and test this change, but we'll get to it as soon as we can.

zenmonkeykstop:
Looks good on first pass, with one exception - there's an inline style applied to the qrcode used for setting up TOTP 2fa that is blocked by the updated policy. I'm not sure if this is an existing bug though - it doesn't affect the scannability of the qrcode so it may jsut have been missed previously. Will confirm.

evilaliv3:
Yep actually if you leave inline styles you are prone to minor vulnerabilities that may tweak the interface or try to the confuse the user, but it would be minor.

The important part of the patch is to limit in general the default-src and especially blocking the connect-src to deny any malware to call back home

zenmonkeykstop:
Yup, I don't think adding unsafe-inline would be a good idea! Confirmed this is an existing bug, will file a separate issue for it.

emkll:
Noting for posterity that the changes will only be applied when `./securedrop-admin install` is run against servers, updates to the CSP will not be applied in an unattended fashion for existing installs. We should keep in mind that the Apache configs between unattended upgraded and installed from workstation on this branch will differ, until `./securedrop-admin install`is run. 

Given the transition to v3 onion services (which require an install run), these changes will likely applied sooner rather than later.

zenmonkeykstop:
@emkll do you think a followup PR with .deb script changes is warranted?

emkll:
I think it's a pretty risky change to make in an unattended fashion. Given the v3 onion migration, as long as an org migrates to v3 after 1.1.0, they will have these applied by the installer. Since that should be the case for many installs, I would argue that applying these to unattended instances might not be worth the risk (as long as we are aware of this, in case we are to make further unattended Apache config changes).


I'll open a ticket to track this for awareness

 think it might be worth opening a ticket to track this, but due to the v3 migration happening soon, we should have pretty good coverage since this will be shipped with 1.1.0

PR #4688, opened 2019-08-21
Fix path to requirements.txt for docs-linkcheck CI job

eloquence:
Resolves #4671; fixes CI failure caused by broken link along the way.

## Status

Ready for review

## Testing

Since this runs weekly, tested in separate commit in test branch (see below)

PR #4689, opened 2019-08-21
sshguard main site down; point to git repo

eloquence:
## Status

Ready for review

## Checklint

- [x] magic lint ball says "Yes."

## Also

Whitespace fixes, because why not.


eloquence:
cherry-picking into #4688 instead

PR #4715, opened 2019-08-28
[wip] Add HTTP Alternative Service header to source, journalist interfaces when both v2 and v3 onion services are enabled

### LINKED ISSUE #4630: [v3 onion migration] If v2 and v3 are both used, set Alt-Svc header in Apache configs
## Description

We can transparently move user traffic over to v3 using the HTTP Alternative Service (`Alt-Svc`) header. If v2 and v3 are both used, we should set the `Alt-Svc` header to point to the v3 onion. 

References for `Alt-Svc`:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Alt-Svc
https://tools.ietf.org/html/rfc7838

Subticket of #2951

redshiftzero:
## Status

work in progress

## Description of Changes

Fixes #4630

 - When both v2 and v3 are enabled, this PR adds an `Alt-Svc` header to send user traffic from v2 to v3 

## Testing

It's not possible as far as I can tell to determine that the Alt-Svc header is being used in Tor Browser (see my thoughts/explanation in #4630 for the full details) but I'll see if I can cook up something better on this for the 1.0.0 QA period. 

For now, you can verify that the header is being set by provisioning staging and visiting the v2 onions. Enabling developer tools in Tor Browser you should see: 

Screen Shot 2019-08-28 at 4 48 31 PM

To convince yourself that this is correct, you can read the [spec](https://tools.ietf.org/html/rfc7838) or you can check out [this site](https://perfectoid.space/test.php), which nicely shows via the color of the background in the page that is visited when the Alt-Svc is being used. It uses an Alt-Svc from Cloudflare:

Screen Shot 2019-08-28 at 4 49 05 PM

### Bonus

I'm marking these as bonus because I think since the reviewer will be inspecting the diff doing this testing as part of 1.0.0 QA is OK:

1. v2_onion_services=False, v3_onion_services=True -> No Alt-Svc header
2. v2_onion_services=True, v3_onion_services=False -> No Alt-Svc header

## Deployment

This will be ran the next time an administrator runs the Ansible playbook. The

- [x] I have written a test plan and validated it for this PR - well, I validated the non-bonus steps

### DIFF
--- a/install_files/ansible-base/roles/app/defaults/main.yml
+++ b/install_files/ansible-base/roles/app/defaults/main.yml
@@ -87,3 +87,9 @@ apache_disabled_modules:
 securedrop_default_locale: en_US
 # The subset of the available locales that will be proposed to the user
 securedrop_supported_locales: []
+
+# v2 Tor onion services are on / v3 Tor onion services are off by default for backwards
+# compatibility. Note that new installs after 1.0 will have v3 enabled by sdconfig which
+# will override these variables.
+v2_onion_services: true
+v3_onion_services: false
--- a/install_files/ansible-base/roles/app/tasks/install_and_harden_apache.yml
+++ b/install_files/ansible-base/roles/app/tasks/install_and_harden_apache.yml
@@ -31,6 +31,16 @@
   tags:
     - apache
 
+- name: Grab source interface v3 onion URL.
+  command: cat /var/lib/tor/services/sourcev3/hostname
+  # Read-only
+  changed_when: false
+  register: v3_source_hostname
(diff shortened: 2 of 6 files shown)


redshiftzero:
given how much other stuff needs to go in ahead of this yet for 1.0.0, i'm marking this wip until we can find a way to properly test

redshiftzero:
yeah I don't see the internal redirect error on develop in staging so is related to the changes here

redshiftzero:
closing this for now, see discussion in #4630

Issue #4691, opened 2019-08-22
investigate why app-test jobs are sometimes hanging in CI


redshiftzero:
## Description

I'm seeing a lot of our [python app-test CI jobs hanging](https://circleci.com/gh/freedomofpress/securedrop/32430
), e.g.:

```
tests/functional/test_admin_interface.py::TestAdminInterface::test_admin_interface PASSED [  0%]
tests/functional/test_admin_interface.py::TestAdminInterface::test_admin_edits_hotp_secret PASSED [  0%]
tests/functional/test_admin_interface.py::TestAdminInterface::test_admin_edits_totp_secret make: *** [test] Terminated
Too long with no output (exceeded 10m0s)
```

We should investigate this - a re-run resolves but it's still annoying for devs/contributors

emkll:
Another one: https://circleci.com/gh/freedomofpress/securedrop/33450

eloquence:
Another one:
https://circleci.com/gh/freedomofpress/securedrop/35018

eloquence:
Bump to 15 minutes does not appear to have materially impacted this flake; we still see failures at the HOTP/TOTP stage regularly, e.g.:

https://circleci.com/gh/freedomofpress/securedrop/36412

eloquence:
I'm not able to spot this app test failure in recent runs, perhaps resolved via #4995?

eloquence:
Closing as we don't seem to be seeing these failures anymore in recent builds.

PR #4718, opened 2019-08-29
Expose v2/v3 Onion URLs to webapp code

### LINKED ISSUE #4631: [v3 onion migration] investigate v2-only user messaging options
## Description

This is a stretch ticket for #2951. 

For organizations enabling v2 and v3 alongside (expected for most existing SecureDrop instances) we’d ideally expose to source and journalist users something like:

`“Warning: you are currently using 0123456789abcedf.onion, in the future please use our new improved onion (language here intentionally vague because we’re not going to take on the task of explaining the detailed differences of v2 versus v3 onions to non-technical users - could be something like “more secure”!) at <long address>.onion.” `

This ticket is to investigate methods to do this. While we could defer this until a future release, it would be wise for us to use some cycles now to think about this now in case we need Apache config changes to pass through v2/v3 info - it’s potentially brittle to make this kind of change in an auto update.

### LINKED ISSUE #4674: testinfra: facilitate running a single test or single file of tests
## Description

As a developer it's nice to be able to run a single test or single file of tests. It would be great to enable this for our testinfra test suite which is used for testing our staging environment: https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html#running-the-config-tests

conorsch:
## Status

Ready for review. 

## Description of Changes

Fixes #4631. Fixes #4674.

* Writes plaintext files to disk, readable by the `www-data` user, containing Onion URL info. Separate files are used for each URL.
* Updates documentation with a quick tip about running a subset of testinfra tests for developer velocity.

Files will exist only if the relevant generation of Onion URLs are configured in site-specific vars. That means that future implementations consuming these files must not assume file existence, and shold also validate the strings inside the files.

## Testing

CI will validate that the staging setup works well. Bust out your Tails/prod VM setup.

* [ ] Configure site-specific vars so that _only_ v2 URLs are enabled, install
* [ ] Log into app server (`ssh app`), confirm that `/var/lib/securedrop/source_v2_url` exists and matches the contents of `/var/lib/tor/services/source/hostname`.
* [ ] Confirm `/var/lib/securedrop/source_v3_url` does not exist
* [ ] Re-configure site-specific vars to enable v3 Onions, re-install
* [ ] Log into app server (`ssh app`), confirm that `/var/lib/securedrop/source_v2_url` exists and matches the contents of `/var/lib/tor/services/source/hostname`.
* [ ] Log into app server (`ssh app`), confirm that `/var/lib/securedrop/source_v3_url` exists and matches the contents of `/var/lib/tor/services/sourcev3/hostname`.

## Deployment
These changes are intended to enable future user-facing messaging regarding URL transitions. We aim to ship them in 1.0 to take advantage of the Admin process of enabling v3 URLs, which requires running the playbooks, to create these files. 

### DIFF
--- a/docs/development/testing_configuration_tests.rst
+++ b/docs/development/testing_configuration_tests.rst
@@ -32,7 +32,7 @@ For the staging VMs:
 
 The VMs will be set up using either the libvirt or virtualbox Vagrant VM provider,
 depending on your system settings. You'll need to use the appropriate commands below
-based on your choice of provider. 
+based on your choice of provider.
 
 Then, to run the tests:
 
@@ -40,7 +40,7 @@ libvirt:
 ~~~~~~~~
 
 .. code:: sh
-   
+
    molecule verify -s libvirt-staging
 
 virtualbox:
@@ -50,12 +50,15 @@ virtualbox:
 
    molecule verify -s virtualbox-staging
 
+.. tip:: To run only a single test, set ``PYTEST_ADDOPTS="-k name_of_test"``
+         in your environment.
+
 Test failure against any host will generate a report with informative output
 about the specific test that triggered the error. Molecule
 will also exit with a non-zero status code.
 
 .. note:: To build and test the VMs with one command, use the Molecule ``test``
-  action: ``molecule test -s libvirt-staging --destroy=never``, or ``molecule test -s virtualbox-staging --destroy=never``. 
+  action: ``molecule test -s libvirt-staging --destroy=never``, or ``molecule test -s virtualbox-staging --destroy=never``.
 
 Updating the Config Tests
 -------------------------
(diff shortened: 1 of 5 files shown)


conorsch:
Rebased to remove the dependency on #4715. Ready for review.

redshiftzero:
can you elaborate why you think prod VMs/Tails testing should be done here? (slower to test). it seems like the main thing we might to verify would be that for v2 only things work as expected (since the logic is straightforward I'm inclined to just rely on the testinfra tests passing for v2/v3)

conorsch:
> can you elaborate why you think prod VMs/Tails testing should be done here? 

Good point, I'm fine with staging VMs being used here. Bear in mind that this step:

> Configure site-specific vars so that only v2 URLs are enabled, install

Requires editing (and generating a diff that should not be committed) `install_files/ansible-base/group_vars/staging.yml` for the various setups.

conorsch:
`python3-app-tests` failing in CI with timeouts: https://circleci.com/gh/freedomofpress/securedrop/32940 Appears unrelated to changes in this PR, let's bounce a few times, then raise timeout if still unresolved.

redshiftzero:
ah that is #4691

conorsch:
Rebased on top of latest develop (d3d3ab79ac519e1545d78109c60cd605b3a29fa8) in an attempt to satisfy CI; as @redshiftzero points out, we've seen this before, so we may need to raise the timeouts here.

PR #4720, opened 2019-08-29
Fixes #4677 warns admin if v3 and https both are enabled

### LINKED ISSUE #4677: [v3 onion migration] warn user if they enable v3 and have https enabled
In `securedrop-admin sdconfig` when a user runs with `v3_onion_services=True`, we should:

1. (minimum) If HTTPS is also enabled, it should print a warning, but not prevent the writing of the config. They should be warned that this change will apply their configured cert to both their v2 and v3 interfaces and until they have a v3 cert, they should not share or advertise their v3 source URL with sources.
2. (better) Inspect the names in the configured cert and warn the user only: when the cert hasn't been updated to include their v3 URLs (predicated on the user having enabled v3 once already) OR the v3 URLs don't exist locally (since this means they are enabling v3 onion services for the first time, they have to run securedrop-admin install at least once in order to get the new URLs).

kushaldas:
## Status

Ready for review

## Description of Changes

Fixes #4677 This should be pulled in after #4710 is merged

We should decide the actual warning text and update in this PR. @redshiftzero 

## Testing

`./securedrop-admin sdconfig`

- If you enable v3 and https both, this should print an warning message
- If either of them disabled, the user should not see any warning.

## Deployment

2. New installs.

### DIFF
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -446,6 +446,19 @@ class SiteConfig(object):
         self.save()
         self.validate_gpg_keys()
         self.validate_journalist_alert_email()
+        self.validate_https_and_v3()
+        return True
+
+    def validate_https_and_v3(self):
+        """
+        Checks if https is enabled with v3 onion service.
+
+        :returns: False if both v3 and https enabled, True otherwise.
+        """
+        if self.config.get("v3_onion_services", False) and \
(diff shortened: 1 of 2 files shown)


conorsch:
#4710 is merged. I see conflicts, @kushaldas—can you rebase and mark ready for review when you're satisfied?

Issue #3204, opened 2018-03-29
Migrate SecureDrop servers to Ubuntu 16.04 (Xenial)


eloquence:
Ubuntu Trusty is reaching EOL in April 2019. We should upgrade SecureDrop servers to Ubuntu Xenial (16.04) before then. This will also unblock [some blocked issues](https://github.com/freedomofpress/securedrop/issues?q=is%3Aopen+is%3Aissue+label%3A%22blocked+on+ubuntu+upgrade%22).

(Please keep discussion about moving to 18.04 out of scope of this issue. We will consider the best path to 18.04, but will not immediately go from 14.04 to 18.04.)

Initially this epic captures preliminary only work; we will update it as we discover more work. The preliminary work must *only* impact the development environment and must not have production consequences.

Tasks:
- [x] Timeboxed install attempt (#3207)
- [x] Timeboxed upgrade attempt (#3491)
- [x] Add Xenial staging VMs  (https://github.com/freedomofpress/securedrop/issues/3206)
- [x] Enable installation on Xenial staging VMs (#3825)
- [x] Add host firewall support for _apt user introduced in 16.04 (#3952)
- [x] Update AppArmor rules to work on Trusty, Xenial (#3962)
- [x] Investigate if PAM common-auth customizations are still needed (#3963)
- [x] Support branched Trusty/Xenial packaging build logic (#3961)
- [x] Support running CI staging tests against both 14.04 and 16.04 (#3966)
- [x] Investigate/fix apt sources under xenial when upgrading  (#4017)
- [x] Display advisory in Journalist Interface for instances running 14.04 (#4027)
- [x] Resolve i18n related test failures #4036
- [x] Fix 2FA functionality in app code #4037
- [x] Fix `Unknown status message: u'IMPORT_OK'` related test failures -  #4038
- [x] Fix errors in functional tests - #4039
- [x] Add package/version information to `securedrop-admin logs` command (#3967)

## In current sprint
- [x] Verify Trusty backup -> Xenial recovery story (#3960)
- [x] Test upgrade path: Ubuntu 14.04.5 → 14.04.5 with all updates → 16.04 (#3965)
- [x] Journalist replies are not displayed in source interface (#4013)
- [x] Handle import_ok exception during upgrade  (#4108)
- [x] Ensure TestInfra test completeness on Trusty, Xenial (#3964)
- [x] Document the upgrade process (#4057)
- [x] Reinstate LTS upgrade notification (#4104)
[...]

PR #4725, opened 2019-09-02
Enable TLSv1.3 ciphers

evilaliv3:
This pull request adds support for a set of TLSv1.3 ciphers and is actually necessary in order to enable support for TLSv1.3.

### DIFF
--- a/install_files/ansible-base/roles/app/defaults/main.yml
+++ b/install_files/ansible-base/roles/app/defaults/main.yml
@@ -34,6 +34,9 @@ securedrop_app_https_certificate_chain_src: DigiCertCA.crt
 # The `SSLHonorCipherOrder` option is set to true, so ciphers below are
 # listed in order of preference.
 securedrop_app_https_ssl_ciphers:
+  - TLS13-AES-256-GCM-SHA384
+  - TLS13-CHACHA20-POLY1305-SHA256
+  - TLS13-AES-128-GCM-SHA256
   - ECDHE-ECDSA-AES256-GCM-SHA384
   - ECDHE-RSA-AES256-GCM-SHA384
   - ECDHE-ECDSA-CHACHA20-POLY1305
--- a/install_files/ansible-base/roles/app/defaults/main.yml
+++ b/install_files/ansible-base/roles/app/defaults/main.yml
@@ -34,9 +34,9 @@ securedrop_app_https_certificate_chain_src: DigiCertCA.crt
 # The `SSLHonorCipherOrder` option is set to true, so ciphers below are
 # listed in order of preference.
 securedrop_app_https_ssl_ciphers:
+  - TLS13-AES-128-GCM-SHA256
   - TLS13-AES-256-GCM-SHA384
   - TLS13-CHACHA20-POLY1305-SHA256
-  - TLS13-AES-128-GCM-SHA256
   - ECDHE-ECDSA-AES256-GCM-SHA384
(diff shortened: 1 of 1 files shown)


evilaliv3:
Thank you for your feedback @kushaldas. Please let me know which is your choice and i will make the proper adjustments.

Actually i just think that for a strange reason the order in that instructions are wrong.
Typically the order of ciphers is set in relation to most common implementations and their performance, but in this case it seems to me that the suggestion is just in the reverse order (from the less secure to the most secure).

@garrettr what do you think? is there anybody in @mozilla that we could involve in this evaluation and eventually notify this error?

redshiftzero:
so from the wiki page it looks like Mozilla is intentionally prioritizing `TLS_AES_128_GCM_SHA256` over `TLS_AES_256_GCM_SHA384` for performance reasons since both ciphersuites are considered secure ("The cipher suites are all strong and so we allow the client to choose, as they will know best if they have support for hardware-accelerated AES"). 

@evilaliv3 you're right that we would need to include the TLS ciphersuites in our configuration in order to support TLSv1.3 since we list our ciphersuites explicitly. However: I just checked and the version of openssl included in Ubuntu Xenial does not support TLSv1.3 (it's [1.0.2](https://packages.ubuntu.com/xenial/openssl)) - version 1.1.1 or later is required for TLSv1.3, so I think we might need to wait a bit for upstream before making this change. What we should probably do here is file an issue for awareness to make sure we update these ciphersuites when upstream (ubuntu) openssl supports TLSv1.3. What do y'all think?

evilaliv3:
Thank you for your feedback @redshiftzero, its now clear.

Actually 128 is prioritized over 256 for the reasons you mention.
Then CHACHA is left as last option but typical implementations use SSL_OP_PRIORITIZE_CHACHA to prioritize CHACHA over AES if the client declares CHACHA as first option.

As for OpenSSL the version 1.1.1 is included in the current Ubuntu LTS (bionic); is there any particular reason your codebase is still based on Ubuntu Xenial?

Here you can see our current implementation and the preferences we have set:
https://www.ssllabs.com/ssltest/analyze.html?d=try.globaleaks.org

Where you read 256P it means:  (P) This server prefers ChaCha20 suites with clients that don't have AES-NI (e.g., Android devices)

redshiftzero:
nope, no particular reason for lack of bionic support, the reason is basically that we haven't got around to it yet and we only recently transitioned to Xenial (shortly before the Trusty EOL, check out https://github.com/freedomofpress/securedrop/issues/3204 for the full details of what the Xenial support and migration entailed). 

At some point we will migrate instances onto the next LTS (unless we make a larger architectural change between now and the migration) though realistically it'll be a while before we make the move to Bionic - getting v3 tor onion service support out to production and removing all Python 2 code paths by the end of the year will be done first (mostly will be resolved in our next release but we do still have a Python 2.7 only CLI tool used by admins).

evilaliv3:
Clear, thank you @redshiftzero

redshiftzero:
hey @evilaliv3, i'm gonna close this one for now since we can't merge it for Xenial, but keep the branch around - when we implement #4768 (though it'll be a while 😇) you can refile again and we can bring it in :-) I filed #4769 for tracking

evilaliv3:
It make sense. Thank you @redshiftzero

If completing the migration to bionic will take time probaby:
- we could evaluate to backport openssl to xenial
- maybe you could just plan to skip bionic and prepare for 20.04 that will arrive next April?

PR #4729, opened 2019-09-03
Adds check for valid v3 config on Admin Workstation

### LINKED ISSUE #4681: [v3 onion migration] fail if v3 partial config is encountered
## Description

In `securedrop-admin install` when a user runs with `v3_onion_services=True`, we should:

* check for existing client auth files
* check for existing tor keypairs JSON file
* fail if one-but-not-both are found 

The goal is to provide a fail-fast mechanism that prevents breaking v3 services for orgs with multiple Administrators. See discussion https://github.com/freedomofpress/securedrop/pull/4652#issuecomment-522703768 for context. The "validate" role seems like the best place to run this logic, since that'll catch misconfigurations even if `securedrop-admin sdconfig` was _not_ run.

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Fixes #4681.

### Scenario 1: fresh install, no v3 config files present
- [x] `./securedrop-admin install` completes successfully

### Scenario 2: install with `tor_v3_keys.json` present and no `.auth_private` files present
- [x] `./securedrop-admin install` completes successfully

### Scenario 3a (SSH-over-Tor): install with `tor_v3_keys.json` present and all `.auth_private` files present
- [x] `./securedrop-admin install` completes successfully

### Scenario 3b (SSH-over-LAN): install with `tor_v3_keys.json` present and  `app-journalist.auth_private` file present
- [x] `./securedrop-admin install` completes successfully

### Scenario 4 (SSH-over-TOR or SSH-over-LAN): install with  `app-journalist.auth_private` file present and no `tor_v3_keys.json` file present
- [x] `./securedrop-admin install` fails with message about missing `tor_v3_keys.json`

### Scenario 5 (SSH-over-TOR): install with  SSH `.auth_private` files present and no `app-journalist.auth_private` file present.
- [x] `./securedrop-admin install` fails with message about missing `app-journalist.auth_private` file.

### DIFF
--- a/install_files/ansible-base/roles/validate/tasks/validate_tails_environment.yml
+++ b/install_files/ansible-base/roles/validate/tasks/validate_tails_environment.yml
@@ -26,3 +26,43 @@
       The SecureDrop git repository should be cloned
       to `~/Persistent/securedrop`.
   with_items: "{{ tails_persistence_check_result.results }}"
+
+- name: Check for v3 client auth files
+  stat:
+    path: "/home/amnesia/Persistent/securedrop/install_files/ansible-base/{{ item }}"
+  register: v3_client_auth_files
+  with_items:
+    - app-ssh.auth_private
+    - app-journalist.auth_private
+    - mon-ssh.auth_private
+
+- name: Count the number of v3 client auth files
+  set_fact:
+    v3_client_auth_file_count: "{{ v3_client_auth_files.results | selectattr('stat.exists') | list | count }}"
+
+- name: Check for Tor v3 key file
+  stat:
+    path: "/home/amnesia/Persistent/securedrop/install_files/ansible-base/tor_v3_keys.json"
+  register: v3_tor_key
+
+- name: Confirm that all necessary v3 client auth files are present
+  assert:
+    that:
+      - v3_client_auth_file_count == "0" or v3_client_auth_file_count == "3"
+    msg: >-
+      Some v3 onion service client auth files are missing. Please add
+      all 3 `.client_auth` files under
+      `~/Persistent/securedrop/install_files/ansible-base`.
+  when: v3_onion_services == True
+
+- name: Confirm that a valid set of v3 configuration files is present
+  assert:
+    that:
+      - v3_tor_key.stat.exists
+    msg: >-
+      Authentication files for v3 onion services were found, but the
+      corresponding `tor_v3_keys.json` file is missing. To enable updates
+      to an existing SecureDrop instance, please add this file under
+      `~/Persistent/securedrop/install_files/ansible-base`.
+  when:
+    - v3_onion_services == True and v3_client_auth_file_count == "3"
--- a/install_files/ansible-base/roles/validate/tasks/validate_tails_environment.yml
+++ b/install_files/ansible-base/roles/validate/tasks/validate_tails_environment.yml
@@ -1,4 +1,3 @@
----
 - name: Confirm host OS is Tails.
   assert:
     that:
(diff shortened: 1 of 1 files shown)


install_files/ansible-base/roles/validate/tasks/validate_tails_environment.yml
```diff
@@ -26,3 +26,43 @@
...
+    path: "/home/amnesia/Persistent/securedrop/install_files/ansible-base/tor_v3_keys.json"
+  register: v3_tor_key
+
+- name: Confirm that all necessary v3 client auth files are present
+  assert:
+    that:
+      - v3_client_auth_file_count == "0" or v3_client_auth_file_count == "3"
+    msg: >-
+      Some v3 onion service client auth files are missing. Please add
+      all 3 `.client_auth` files under
```
redshiftzero: to avoid confusion we should say `.client_auth` -> `.auth_private`

install_files/ansible-base/roles/validate/tasks/validate_tails_environment.yml
```diff
@@ -26,3 +26,43 @@
...
+
+- name: Check for Tor v3 key file
+  stat:
+    path: "/home/amnesia/Persistent/securedrop/install_files/ansible-base/tor_v3_keys.json"
+  register: v3_tor_key
+
+- name: Confirm that all necessary v3 client auth files are present
+  assert:
+    that:
+      - v3_client_auth_file_count == "0" or v3_client_auth_file_count == "3"
```
redshiftzero: what if SSH over LAN is used? we should allow `v3_client_auth_file_count` to be 1 in that case (we should allow it to be 1 _or_ 3 in case the admin is just turning off SSH over LAN on this run of the playbook)

install_files/ansible-base/roles/validate/tasks/validate_tails_environment.yml
zenmonkeykstop: check!

install_files/ansible-base/roles/validate/tasks/validate_tails_environment.yml
zenmonkeykstop: Added a separate assert and made them conditional on `enable_ssh_over_tor` instead.

redshiftzero:
hey can ye rebase out the TB version update in 298900a since it was merged into `develop`? (then will re-review)

install_files/ansible-base/roles/validate/tasks/validate_tails_environment.yml
```diff
@@ -26,3 +26,54 @@
...
+- name: Confirm that v3 auth files are present for SSH-over-LAN configuration
+  assert:
+    that:
+      - v3_client_auth_file_count == "0" or v3_client_auth_file_count == "1"
+    msg: >-
+      There are multiple `.auth_private` files under
+      `~/Persistent/securedrop/install_files/ansible-base`, most likely from
+      a previous installation. Please remove any `.auth_private` files except
+      for `app-journalist.auth_private`, and retry the install command.
+  when: v3_onion_services == True and enable_ssh_over_tor == False
```
redshiftzero: consider this scenario:

1. User has a SecureDrop on 0.14.0 using SSH over Tor (has 0 .auth_private files)
2. User updates to 1.0.0, adds v3 onion service support (now has 3 .auth_private files)
3. User then decides to turn off SSH over Tor and enable SSH over LAN at some later date. The validation here will fail since they have 3 `.private_auth` files in their admin workstation when they run the playbook. 

suggestion: we relax the validation here to allow 0, 1 or 3 .auth_private files

install_files/ansible-base/roles/validate/tasks/validate_tails_environment.yml
```diff
@@ -26,3 +26,54 @@
...
+
+- name: Confirm that v3 auth files are present for SSH-over-Tor configuration
+  assert:
+    that:
+      - v3_client_auth_file_count == "0" or v3_client_auth_file_count == "3"
+    msg: >-
+      Some v3 onion service client auth files are missing. Please add
+      all 3 `.auth_private` files under
+      `~/Persistent/securedrop/install_files/ansible-base`.
+  when: v3_onion_services == True and enable_ssh_over_tor == True
```
redshiftzero: consider a similar scenario for SSH over LAN:

1. User has a SecureDrop on 0.14.0 using SSH over LAN (has 0 .auth_private files)
2. User updates to 1.0.0, adds v3 onion service support (now has 1 .auth_private files)
3. User then decides to turn on SSH over Tor and disable SSH over LAN at some later date. The validation here will fail since they have only 1 `.private_auth` files in their admin workstation when they run the playbook. 

suggestion: we relax the validation here to allow 0, 1 or 3 .auth_private files

install_files/ansible-base/roles/validate/tasks/validate_tails_environment.yml
zenmonkeykstop: I think the appropriate thing to do might be to test for the ssh auth files separately, and if they're running `install` with ssh-over-lan selected, those files should not be present (so prompt them to move them away). `tailsconfig` doesn't reference `site-specific` (because JWs won't have it) so we can't check the  `enable-ssh-over-tor` toggle, and they'll get added to Tor's config if they're present, which we probably don't want.

install_files/ansible-base/roles/validate/tasks/validate_tails_environment.yml
zenmonkeykstop: Probably same deal, testing for SSH separately. So there should be be 0 or 2 ssh auth files.

install_files/ansible-base/roles/validate/tasks/validate_tails_environment.yml
redshiftzero: So we'd have the following scenario:

1. User installs their SecureDrop on 0.14.0 using SSH over Tor
2. They add v3 onion service support
3. They run `securedrop-admin tailsconfig`
4. They run `securedrop-admin sdconfig` to turn off SSH over LAN
5. They run `securedrop-admin install` to apply the change (connecting over Tor for the playbook run)
6. The install fails, and instructs them to move their SSH `.auth_private` files (move to somewhere that will persist but not delete them). 
7. They re-run `securedrop-admin install`

I'm concerned if the install fails in step 7 for some other reason (as we know there are all sorts of reasons why this might fail) and the admin runs tailsconfig again at some later date, they might end up in a confusing state unable to SSH in.

zenmonkeykstop: OK, simplest set of checks that cover all above scenarios:
- confirm that there are 0 or 2 SSH auth files in all cases
- confirm that there is a Journalist auth file if there are 2 SSH auth files and v3 enabled and ssh-over-tor enabled.
- confirm that there is a keys file if there is a journalist auth file and v3 enabled.

zenmonkeykstop:
OK, ready for another look @redshiftzero (apologies for the glut of force pushes, testing on a Tails VM as I go)

PR #4735, opened 2019-09-03
Correct cleanup of disconnected submission files

### LINKED ISSUE #4734: New submission cleanup could delete valid files of replies
## Description

In `securedrop/management/submissions.py`, the `find_disconnected_fs_submissions` returns any file found in the store that is not associated with a `Submission` record in the database. It should only return files that have neither a `Submission` nor `Reply` record, to avoid warning admins about and possibly deleting replies.

## Steps to Reproduce

- Run `make dev`
- In another shell:
  - Run `docker container ls` to get the dev container's ID
  - Run `docker exec -it "container-id" bash`
  - Run `./manage.py list-disconnected-fs-submissions`

## Expected Behavior

With a consistent database and filesystem, no disconnects should be reported. 

## Actual Behavior

The files of all replies in the database are reported as candidates for cleanup.

## Comments

The function simply needs to build a list of reply files as it does for submissions, and check the files under the store against the complete set of valid files.

rmol:
## Status

Ready for review

## Description of Changes

In `securedrop.management.submissions.find_disconnected_fs_submissions`, check replies as well as submissions for a valid association to any file found in the store, before reporting it as disconnected.

Fixes #4734.

## Testing

- Run `make dev`
- In another shell:
  - Run `docker container ls` to get the dev container's ID
  - Run `docker exec -it "container-id" bash`
  - Run `./manage.py list-disconnected-fs-submissions`

No disconnects should be reported.

- In the second shell:
  - `cd /var/lib/securedrop`
  - `sqlite3 db.sqlite`
    - `select * from submissions;`  -- there should be four records
    - `delete from submissions where id = 4;`  -- adjust if there is no record with ID 4
  - `cd -`
  - `./manage.py list-disconnected-fs-submissions`

The file whose submission you just deleted should be the only disconnect reported.

## Deployment

No concerns; this fixes a serious bug.

### DIFF
--- a/securedrop/tests/test_submission_cleanup.py
+++ b/securedrop/tests/test_submission_cleanup.py
@@ -13,7 +13,7 @@ def test_delete_disconnected_db_submissions(journalist_app, config):
     Test that Submission records without corresponding files are deleted.
     """
     with journalist_app.app_context():
-        source, codename = utils.db_helper.init_source()
+        source, _ = utils.db_helper.init_source()
         source_id = source.id
 
         # make two submissions
@@ -43,13 +43,18 @@ def test_delete_disconnected_fs_submissions(journalist_app, config):
     """
     Test that files in the store without corresponding Submission records are deleted.
     """
-    source, codename = utils.db_helper.init_source()
+    source, _ = utils.db_helper.init_source()
 
     # make two submissions
     utils.db_helper.submit(source, 2)
     source_filesystem_id = source.filesystem_id
     submission_filename = source.submissions[0].filename
     disconnect_path = os.path.join(config.STORE_DIR, source_filesystem_id, submission_filename)
+
+    # make two replies, to make sure that their files are not seen
+    # as disconnects
+    journalist, _ = utils.db_helper.init_journalist("Mary", "Lane")
+    utils.db_helper.reply(journalist, source, 2)
 
     # delete the first Submission record
     db.session.delete(source.submissions[0])
--- a/securedrop/management/submissions.py
+++ b/securedrop/management/submissions.py
@@ -9,11 +9,14 @@ from six.moves import input
 
 from db import db
 from rm import secure_delete
-from models import Source, Submission
+from models import Reply, Source, Submission
 from management import app_context
 
 
 def find_disconnected_db_submissions(path):
+    """
+    Finds Submission records whose file does not exist.
+    """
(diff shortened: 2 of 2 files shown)


securedrop/tests/test_submission_cleanup.py
```diff
@@ -43,14 +43,19 @@ def test_delete_disconnected_fs_submissions(journalist_app, config):
...
 
     # make two submissions
     utils.db_helper.submit(source, 2)
     source_filesystem_id = source.filesystem_id
     submission_filename = source.submissions[0].filename
     disconnect_path = os.path.join(config.STORE_DIR, source_filesystem_id, submission_filename)
 
+    # make two replies, to make sure that their files are not seen
+    # as disconnects
+    journalist, _ = utils.db_helper.init_journalist("Mary", "Lane")
```
zenmonkeykstop: Would it make sense to also delete one of the reply records here, to verify that reply files are being deleted correctly?

rmol: Reply deletion isn't part of submission cleanup (but see below), and it is already tested in  `test_integration.py` and `test_journalist.py`. It is worth nothing that [`test_integration._helper_test_reply`](https://github.com/freedomofpress/securedrop/blob/ae0ff0c36b80062097826852b02f667243c45a16/securedrop/tests/test_integration.py#L350) does not check that the file is actually gone, as is done in [`test_submit_file`](https://github.com/freedomofpress/securedrop/blob/ae0ff0c36b80062097826852b02f667243c45a16/securedrop/tests/test_integration.py#L242), but that _is_ tested, albeit somewhat indirectly, in [`test_journalist.test_delete_source_deletes_docs_on_disk`](https://github.com/freedomofpress/securedrop/blob/ae0ff0c36b80062097826852b02f667243c45a16/securedrop/tests/test_journalist.py#L1683), where a source collection with submissions and replies is deleted.

So ... it's probably worth reviewing all the tests involving reply deletion, in a separate issue. And I think if we have submission cleanup tools, we should have them for replies as well. Replies are smaller and less likely to have their deletion process interrupted as has happened for large submissions, but it is still technically possible for their files and database records to be disconnected. But I think that's out of scope for this fix, and possibly for 1.0.0.

redshiftzero: agreed, let's keep this scoped to submissions for 1.0.0

PR #4777, opened 2019-09-10
[1.0.0] Fix PR # for orphaned submission entry in changelog

eloquence:
## Status

Ready for review


redshiftzero:
seems like a good catch - @kushaldas what do you think?

kushaldas:
@redshiftzero I think this should be merged.

PR #4801, opened 2019-09-11
Rework Ansible restore role to detect Onion service version mismatch

### LINKED ISSUE #4786: Restoring from a v2 backup on an instance with v3 enabled breaks v3 onion services
## Description

When a v2 backup is used to restore an instance that has v3 services enabled (either v3 only or v2+v3), the v3 onion services are no longer enabled on the app server. 

## Steps to Reproduce

on prod vms or hardware:

- install a v2-only instance
- take a backup
- use `sdconfig` and `install` to enable v3
- restore from the v2backup 

## Expected Behavior
- data is restored, v3 services are still accessible

## Actual Behavior
- restore hangs when tor is restarted on the server, v3 services are no longer configured on the app server

## Comments
The backup tarball includes /etc/tor/torrc, which contains the onion service definitions. When it overwrites the v3 torrc, it breaks the services.

One approach to address this might be to break the onion service definitions out into files that are included in `torrc` via  %include. For example,a structure like:
```
/etc/tor/torrc.d/
                    10-v2-definitions.conf
                    20-v3-definitions.conf
```
and a line like:
```
%include /etc/tor/torrc.d
```
in the `torrc` proper.

Restoring from a v2 tarball would overwrite `/etc/tor/torrc.d/10-v2-definitions.conf` but shouldn't affect other files in the same directory.

rmol:
## Status

Ready for review

## Description of Changes

Make the restore playbook extract the Tor configuration from the backup, retrieve the app server's Tor configuration, and compare them. If the service versions differ, abort the restore with an explanation.

Fixes #4786.

## Testing

- Install the latest 1.0.0 release candidate on hardware or VMs, specifying only `v2_onion_services` with `sdconfig`/`site-specific`.
- Check out this branch on your admin workstation.
- Create a marker file under `/var/lib/securedrop/store` to test the restore. Send a submission through the source interface, or simply create a file manually. If you've got the time to try a largish file, great.
- Run `securedrop-admin backup` to backup the v2-only system.
- Delete the marker file.
- Run `securedrop-admin restore` to restore that backup. Confirm that the restore worked by verifying the presence of the marker file.
- Add v3 with `sdconfig`, then `securedrop-admin install`, then `securedrop-admin tailsconfig`.
- Run `securedrop-admin restore` with the backup file created earlier. The restore should abort with this message:
```
The Tor configuration on the app server offers version 2 and 3 services.
The Tor configuration in this backup offers version 2 services.

Restoring a backup with a different Tor configuration than the server 
is currently unsupported. If you require technical assistance, please 
contact the SecureDrop team via the support portal or at 
securedrop@freedom.press.
```

## Deployment

This reworks the restore playbook, replacing the `restore.py` script previously copied to the server with equivalent Ansible tasks, preventing hangs when Tor is restarted on the server. Those changes are straightforward but deserve scrutiny.


install_files/ansible-base/roles/restore/tasks/main.yml
```diff
@@ -1,22 +1,81 @@
...
+  connection: local
+  become: no
   file:
-    path: /tmp/{{ restore_file }}
+    path: "{{ torrc_check_dir.path }}"
     state: absent
+  when: torrc_check_dir.path is defined
+
+- name: Copy backup to application server
+  synchronize:
```
conorsch: Nit: `copy` module is sufficient here, since we're just copying a single file.

install_files/ansible-base/roles/restore/tasks/main.yml
```diff
@@ -1,22 +1,81 @@
...
+  become: no
+  tempfile:
+    state: directory
+  register: torrc_check_dir
+
+- name: Fetch current Tor configuration from app server
+  become: no
+  fetch:
+    src: /etc/tor/torrc
+    dest: "{{ torrc_check_dir.path }}"
```
conorsch: First thought was that we needn't copy the `torrc` back to the Admin Workstation, but doing so allows us to compare the files _before_ we copy the potentially-quite-large tarball to the remote. Great choice. :+1:

admin/securedrop_admin/__init__.py
```diff
@@ -736,6 +736,10 @@ def restore_securedrop(args):
     # but the securedrop-admin
     # script will be invoked from the repo root, so preceding dirs are likely.
     restore_file_basename = os.path.basename(args.restore_file)
+
+    # Would like readable output if there's a problem
+    os.environ["ANSIBLE_STDOUT_CALLBACK"] = "debug"
```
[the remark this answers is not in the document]
rmol: No, that was intentional, to present the output of `compare_torrc.py` more legibly. This should only affect the environment in the restore function.

install_files/ansible-base/roles/restore/tasks/main.yml
rmol: The idea was that with `partial: yes` an interrupted upload of the backup could be resumed.

rmol:
Re: simply diffing -- I was originally just using Ansible's `stat.checksum` to detect differences, but it's useful to determine how they differ, to be able to say which side has v2, v3, or both. It was going to be more useful when I was hoping to provide guidance for reconfiguring to enable the restore, but I ran out of time.

PR #4809, opened 2019-09-12
Add flaky to translation-test

### LINKED ISSUE #4808: Translation tests should use the flaky plugin too
## Description

The translation tests can fail due to Selenium flakiness, just like the regular tests, but `securedrop/bin/translation-test` isn't using the `flaky` plugin. We should add that.

rmol:
## Status

Ready for review

## Description of Changes

Adds the flaky pytest plugin to `securedrop/bin/translation-test`.

Fixes #4808.

## Testing

- Run `make translation-test`
- Verify that `flaky` appears in the list of pytest plugins
- Interrupt the tests after the first one passes, then verify that you see the `Flaky Test Report` in the output

## Deployment

This is dev/test only.

### DIFF
--- a/securedrop/bin/translation-test
+++ b/securedrop/bin/translation-test
@@ -24,6 +24,7 @@ for locale in ar de_DE es_ES fr_FR hi is it_IT nb_NO nl pt_BR ro ru sv tr zh_Han
 do
   PAGE_LAYOUT_LOCALES=$locale pytest \
     -v \
+    --force-flaky --max-runs=3 \
     --page-layout \
     --durations 10 \
     "$@"


zenmonkeykstop:
Failing for me on the same test - but it does not seem related to the `flaky` change.

zenmonkeykstop:
It fails because the button row wraps due to German having a higher character count than will fit, making the last button inaccessible to Selenium...

rmol:
Which was supposed to have been fixed by #4819 -- rebasing this onto develop this morning was supposed to fix that, and since CI passed I thought it had. Are you sure you're testing with the current branch?

zenmonkeykstop:
I rebased it locally and the test passed. Will merge.

rmol:
Whew! Thanks.

PR #4845, opened 2019-09-19
Fixes #4791 shows png files instead of SVG in source index

### LINKED ISSUE #4791: SVG in Tor warning banner not rendered if NoScript set to permit JS
## Steps to reproduce:
1. Set your Tor security level to "safest"
2. Set your NoScript setting to "Disable restrictions for this tab"
3. Load a SecureDrop source interface 

(Works with any live SecureDrop instance >= 0.14.0)

## Expected behavior

The security warning banner renders normally.

## Actual behavior

The security warning banner looks like this:

Screenshot from 2019-09-10 21-03-54

## Explanation

Tor enforces no consistency between the Tor security setting and the NoScript setting: You can have JavaScript enabled while your security level is nominally set to "safest" (which disables SVG rendering). The NoScript button is very visible, so it's not too much of an edge case to imagine that a user may have put their browser in this state intentionally or by accident.

These icons should probably be served up as PNG. That said, the banner is also incorrect in this situation (following the steps will not have any effect), so we may want to take this case into account in the next iteration on this warning.

kushaldas:
## Status

Ready for review

## Description of Changes

Fixes #4791

Adds 3 png files and also updates the source index tempate to use
the png files.

## Testing

- Set your Tor security level to "safest"
-  Set your NoScript setting to "Disable restrictions for this tab"
- Load a SecureDrop source interface

## Deployment

2. New installs.


kushaldas:
>  rules.

We should at least keep the svgs as source file for the png files. And, I think we should remove the AppArmor rules as you suggested.

kushaldas:
@emkll ping for bringing it back to view :)

redshiftzero:
> I think we should remove the AppArmor rules as you suggested

It looks like the svg files that are no longer used are still whitelisted by AppArmor in this PR - want to update that prior to merge @kushaldas?

redshiftzero:
hmm lint is failing but the failure looks unrelated to this diff - want to rebase on latest?

eloquence:
Closing out this year+-old PR, @kushaldas feel free to reopen if you want to pick this up again.

PR #4859, opened 2019-09-23
remove python 2 support in securedrop server code

### LINKED ISSUE #4699: remove Python 2 code paths in server code
## Description

If no major issues are found with the Python 2 to 3 transition during SecureDrop 1.0.0 QA, let's remove:

- [ ] the `PYTHON_VERSION` env var
- [ ] the Python 2 dockerfile
- [ ] the Python 2 application test CI job (and the Python 2-only DLC steps!)
- [ ] any Python 2 - only logic (e.g. all the `six` callouts) from the server side code _only_
- [ ] any developer docs referring to the above or the code needing to be both Python 2 and Python 3 compliant

Keep in mind that we still need to complete the migration of securedrop-admin to Python 3, which we'll do in the next regular release after 1.0.0 (i.e. 1.1.0). However the only function that is used during the staging provisioning process is both Python 2 and 3 compliant, so removing the Python 2 env var in the server code should not cause issues.

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #4699

 - Removes python 2 support from SecureDrop server code
 - Removes python 2 CI job
 - Removes python 2 dev env support

## Testing

For this one you should be able to rely on diff review and CI (application unit tests):
- `app-tests` job (this is the python 3 application test job) should pass
- all `six` calls from app code and tests should be gone
- developer documentation should not reference python 2 compliance

## Deployment

as of 1.0.0 the paths I'm removing are no longer ran in production, the other changes are mostly dev/CI facing


rmol:
Fall cleaning! This looks great. :slightly_smiling_face: 

Found a few lingering references in the docs (which I'd link to if GitHub didn't insist on rendering them :slightly_frowning_face: ): 
- In `contributor_guidelines.rst`, line 128: `SecureDrop uses Python2, that is why all type annotation/hints are mentioned as code comments.`
- In `setup_development.rst`, line 243, we're still recommending developers on Macs set up a Python 2 virtualenv.

And in `install_files/securedrop-app-code/debian/securedrop-app-code.triggers` the comment says we're registering interest in Python 2 interpreter changes, but we're not.

Makefile
```diff
@@ -146,9 +146,9 @@ safety:  ## Run `safety check` to check python dependencies for vulnerabilities.
 bandit: test-config ## Run bandit with medium level excluding test-related folders.
 	@command -v bandit || (echo "Please run 'pip install -U bandit'."; exit 1)
 	@echo "███ Running bandit..."
-	@bandit -ll --exclude ./admin/.tox,./admin/.venv,./admin/.eggs,./molecule,./testinfra,./securedrop/tests,./.tox,./.venv*,securedrop/config.py --recursive .
+	@bandit -ll --skip B322 --exclude ./admin/.tox,./admin/.venv,./admin/.eggs,./molecule,./testinfra,./securedrop/tests,./.tox,./.venv*,securedrop/config.py --recursive .
```
emkll: :+1: an interested observer can see https://docs.openstack.org/bandit/latest/api/bandit.blacklists.html#b322-input for more details

emkll: Note that we still have some Python2 code (securedrop-admin), but that will be moved to python3 before the next release, so I think it's fine to merge

rmol:
* https://github.com/freedomofpress/securedrop/blob/f4e3fcfbab156fd8ed8ba788dd3961cdac7289c3/install_files/ansible-base/group_vars/all/securedrop#L44
       
>       Is this package strictly required for SecureDrop now? Since we are using a virtualenv, this package should no longer be necessary ?

That's a good catch. I think it should be removed.

PR #4870, opened 2019-09-25
Apply mix fixes in install_files/securedrop-app-code/debian/copyright

evilaliv3:
Looking at the package i just noticed two minor declaration mistakes

1) The referenced repository is wrong
2) The copyright date is outdated

### DIFF
--- a/install_files/securedrop-app-code/debian/copyright
+++ b/install_files/securedrop-app-code/debian/copyright
@@ -1,5 +1,5 @@
 Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
-Upstream-Name: securedrop-proxy
+Upstream-Name: securedrop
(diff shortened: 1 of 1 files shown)


install_files/securedrop-app-code/debian/copyright
```diff
@@ -1,9 +1,9 @@
 Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 Upstream-Name: securedrop-proxy
```
rmol: Great catch! We should probably change `Upstream-Name: securedrop-proxy` to `Upstream-Name: securedrop` as well.

conorsch:
Aye, thanks, @rmol! Append a commit to save @evilaliv3 the trouble. Re-approving...

PR #4871, opened 2019-09-25
Fix current package version in setup.py

evilaliv3:
Looking at the packaging i just noticed that the version indicated in the setup.py is still the old rc1

It actually seems to not affect anything in the build of the package.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ long_description = "The SecureDrop whistleblower platform."
 
 setuptools.setup(
     name="securedrop-app-code",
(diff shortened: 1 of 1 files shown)


rmol:
We actually don't want to revert it to 1.0.0 from 1.1.0~rc1. This version number is managed with a script, `update_version.sh`. 

As you correctly point out, the version in setup.py doesn't affect anything, as we aren't really creating an distributable/installable Python package. It's just used in the dh-virtualenv build process.

evilaliv3:
I see, ok!

PR #4880, opened 2019-10-07
Copyedit and update Source Guide

eloquence:
- The "Get Tor" instructions as written didn't work; the email autoresponder requires the OS to be specified, and the GitHub downloads are currently not available.

- The instructions suggested visiting the .onion address _before_ "choosing who to submit to".

- Added a recommendation to verify the .onion address against the SecureDrop Directory

- Used more common terms in a couple of places (e.g., "USB Flash Drive" instead of "USB Key"), simplified Flag for Reply language

- Wrap to 80 characters

## Status

Ready for review 

## Checklint

- [x] All good on the linting front

### DIFF
--- a/docs/source.rst
+++ b/docs/source.rst
@@ -4,40 +4,84 @@ Source Guide
 Choosing the Right Location
 ---------------------------
 
-When national security is involved, we suggest you buy a new computer,
-a USB key and pay cash. In any case you must then find a busy
-coworking place or cyber cafe you don't regularly go to and sit at a
-place with your back to a wall to avoid cameras capturing information
-on your screen or keystrokes.
+When national security is involved, we suggest you buy a new computer and a
+USB flash drive, using cash. In any case you must then find a busy coworking
+place or cyber cafe you don't regularly go to and sit at a place with your back
+to a wall to avoid cameras capturing information on your screen or keystrokes.
 
 Get the Tor Browser
 -------------------
 
-Each SecureDrop instance has a publicly available *Source Interface*: a website where sources can create anonymous accounts, submit files and messages, and check back for replies.
+Each SecureDrop instance has a publicly available *Source Interface:* a website
+where sources can create anonymous accounts, submit files and messages, and
+check back for replies.
 
-Each Source Interface is only available as a *Tor Hidden Service*, which is a special type of website with an address ending in ".onion" that is only accessible through Tor. Tor is an anonymizing network that makes it difficult for anybody observing the network to associate a user's identity (e.g. their computer's IP address) with their activity (e.g. uploading information to SecureDrop).
+Each *Source Interface* is only available as an onion service, which is a
+special type of website with an address ending in ".onion" that is only
+accessible through Tor. Tor is an anonymizing network that makes it difficult
+for anybody observing the network to associate a user's identity (e.g., their
+computer's IP address) with their activity (e.g., uploading information to
+SecureDrop).
 
-The easiest and most secure way to use Tor is to download the Tor Browser from the `Tor Project website`_. The Tor Browser is a modified version of the Firefox web browser designed to protect your security and anonymity while using Tor. If there is a chance that downloading the Tor Browser raises suspicion, you have a few alternatives.
+The easiest and most secure way to use Tor is to download the Tor Browser from
+the `Tor Project website`_. The Tor Browser is a modified version of the Firefox
+web browser. It was designed to protect your security and anonymity while
+using Tor. If there is a chance that downloading the Tor Browser raises
+suspicion, you have a few alternatives, for example:
 
-* If your mail provider is less likely to be monitored, you can send a mail to gettor@torproject.org and a bot will answer with instructions
-* If you routinely use GitHub, you can use it to `download the Tor Browser <https://github.com/TheTorProject/gettorbrowser>`__
+* If your mail provider is less likely to be monitored, you can send a mail to
+  gettor@torproject.org with the text "linux", "windows" or "osx" in the body
+  (for your preferred operating system) and a bot will answer with instructions.
+* If you routinely use Git, you can use GitLab to
+  `download the Tor Browser <https://gitlab.com/thetorproject/gettorbrowser/tree/torbrowser-releases>`__.
 
-Once you have the Tor Browser, launch it and enter the ".onion" address for the Source Interface of the organization that you wish to submit to. You can find this address on the organization's *Landing Page*, or listed on the SecureDrop Directory.
-
-While using the Tor Browser on your personal computer helps hide your activity on the network, it leaves traces (of its own installation) on your local machine. For even more deniability, we recommend booting into a live system such as `Tails`_ for a higher level of security. Tails is specifically designed to run on your computer without leaving traces of your activity, and automatically routes all of your Internet browsing through Tor so you can easily access SecureDrop safely.
+While using the Tor Browser on your personal computer helps hide your activity
+on the network, it leaves traces of its own installation on your local
+machine. For even greater deniability and security, we recommend booting into the
+`Tails operating system`_ (typically from a USB stick). Tails is specifically
+designed to run on your computer without leaving traces of your activity, and
+automatically routes all of your Internet browsing through Tor so you can easily
+access SecureDrop safely.
 
 .. _`Tor Project website`: https://www.torproject.org/
-.. _`Tails`: https://tails.boum.org/
-
+.. _`Tails operating system`: https://tails.boum.org/
 
 Choose Who to Submit To
 -----------------------
+We recommend conducting all research related to your submission in Tor Browser.
+If you are unsure whether you are using Tor, you can visit the address
+https://check.torproject.org.
 
-Each SecureDrop instance is totally independent, and submissions to that instance are only available to journalists associated with that organization.
+All organizations operating SecureDrop have a *Landing Page* that provides their
+own organization-specific recommendations for using SecureDrop. We encourage you
+to consider an organization's *Landing Page* before submitting to them.
 
-All organizations have a *Landing Page* that provides their own organization-specific recommendations for using SecureDrop. We encourage you to consider an organization's *Landing Page* before submitting to them.
+.. note::
 
-Most organizations make their *Landing Page* prominently accessible from their main website's homepage. You can also find an incomplete list of organizations accepting submissions through SecureDrop on the `SecureDrop Directory`_ maintained by Freedom of the Press Foundation.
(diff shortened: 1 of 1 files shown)


eloquence:
If this looks good, I'd recommend approval but not merge until the issue with the new "Report an Error" form is resolved (see https://github.com/freedomofpress/securedrop.org/issues/648 ).

docs/source.rst
```diff
@@ -4,73 +4,127 @@ Source Guide
...
+  (for your preferred operating system) and a bot will answer with instructions.
+* If you routinely use Git, you can use GitLab to
+  `download the Tor Browser <https://gitlab.com/thetorproject/gettorbrowser/tree/torbrowser-releases>`__.
+
+While using the Tor Browser on your personal computer helps hide your activity
+on the network, it leaves traces of its own installation on your local
+machine. For even greater deniability and security, we recommend booting into the
+`Tails operating system`_ (typically from a USB stick). Tails is specifically
+designed to run on your computer without leaving traces of your activity, and
+automatically routes all of your Internet browsing through Tor so you can easily
```
emkll: Perhaps adding something that will evoke segmentation for the prospective source, something like: "By using tails, you reduce the risk of operating system level logging, indexing, and certain monitoring/management tools on your everyday workstation, as well as not interacting with your files/applications"

docs/source.rst
```diff
@@ -4,73 +4,127 @@ Source Guide
...
-* If you routinely use GitHub, you can use it to `download the Tor Browser <https://github.com/TheTorProject/gettorbrowser>`__
+All organizations operating SecureDrop have a *Landing Page* that provides their
+own organization-specific recommendations for using SecureDrop. We encourage you
+to consider an organization's *Landing Page* before submitting to them.
 
-Once you have the Tor Browser, launch it and enter the ".onion" address for the Source Interface of the organization that you wish to submit to. You can find this address on the organization's *Landing Page*, or listed on the SecureDrop Directory.
+.. note::
 
-While using the Tor Browser on your personal computer helps hide your activity on the network, it leaves traces (of its own installation) on your local machine. For even more deniability, we recommend booting into a live system such as `Tails`_ for a higher level of security. Tails is specifically designed to run on your computer without leaving traces of your activity, and automatically routes all of your Internet browsing through Tor so you can easily access SecureDrop safely.
+   Each SecureDrop instance is totally independent, and submissions to that
```
emkll: I propose "each SecureDrop instance is operated and administered independently by the news organization"

docs/source.rst
```diff
@@ -156,17 +210,16 @@ below message.
 |Delete received messages|
 
 If the server experiences a large number of new sources signing up at
-once and is overloaded with submissions, the journalist will flag your
-message on their end and you will see the message below. They can't
-write a reply to you until you've seen this message for security
-reasons. This will only happen the first time a journalist replies and
-with subsequent replies you will skip this step. Click **Refresh** or
-log in again to see if a journalist has responded.
+once and is overloaded with submissions, you may see the message below:
```
emkll: I propose changing overloaded to "experiences a surge in traffic" or "is experiencing high traffic volume"

docs/source.rst
```diff
@@ -4,73 +4,127 @@ Source Guide
...
+The easiest and most secure way to use Tor is to download the Tor Browser from
+the `Tor Project website`_. The Tor Browser is a modified version of the Firefox
+web browser. It was designed to protect your security and anonymity while
+using Tor. If there is a chance that downloading the Tor Browser raises
+suspicion, you have a few alternatives, for example:
+
+* If your mail provider is less likely to be monitored, you can send a mail to
+  gettor@torproject.org with the text "linux", "windows" or "osx" in the body
+  (for your preferred operating system) and a bot will answer with instructions.
+* If you routinely use Git, you can use GitLab to
```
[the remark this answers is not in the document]
eloquence: Reworded in 53e978e3a3ee5bfbdb81c2f144c00b4f60edb73b. I'm not very fond of having this here at all given the lack of verification instructions. Perhaps we should just describe the email service and leave it at that?

docs/source.rst
```diff
@@ -4,73 +4,127 @@ Source Guide
...
+  `download the Tor Browser <https://gitlab.com/thetorproject/gettorbrowser/tree/torbrowser-releases>`__.
+
+While using the Tor Browser on your personal computer helps hide your activity
```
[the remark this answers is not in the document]
eloquence: OK, added a note to that effect in  53e978e3a3ee5bfbdb81c2f144c00b4f60edb73b. I agree this is an important -- potentially life-altering -- warning to have, but am also worried about people completely misunderstanding it. What do you think about this wording, does it strike the right balance?

docs/source.rst
eloquence: I find that language a bit too technical to be honest, but I tried to make it clearer in  53e978e3a3ee5bfbdb81c2f144c00b4f60edb73b that there's additional OS-level activity logging Tails mitigates against. Does that work for you?

docs/source.rst
eloquence: Worded very similarly in  53e978e3a3ee5bfbdb81c2f144c00b4f60edb73b (I try not to overuse "news organization" since not all orgs using SecureDrop are strictly speaking news orgs).

docs/source.rst
```diff
@@ -4,73 +4,127 @@ Source Guide
...
-Choose Who to Submit To
------------------------
+.. tip::
 
-Each SecureDrop instance is totally independent, and submissions to that instance are only available to journalists associated with that organization.
+   If the organization does have an entry in the SecureDrop Directory, we
+   recommend comparing the address of the entry with the one on the
+   organization's own *Landing Page*. If the two addresses don't match, please
+   do not submit to this organization yet. Instead, please
+   `contact us <https://securedrop.org/report-an-error>`__  through th
```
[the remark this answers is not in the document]
eloquence: Excellent point, added in  53e978e3a3ee5bfbdb81c2f144c00b4f60edb73b.

docs/source.rst
eloquence: Yeah I can see why :), reworded in 53e978e3a3ee5bfbdb81c2f144c00b4f60edb73b.

docs/source.rst
ninavizz: "host organization" feels like a fit, perhaps? The language "instance" and "host" both used throughout, is what felt best to me when I took a stab at this a while ago.

docs/source.rst
ninavizz: Imposter syndrome is real; let's be mindful to not trigger that in users... :)

docs/source.rst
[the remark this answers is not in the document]
ninavizz: OK—is this, then, validation @emkll that we should take more aggressive measures at the beginning of the Source UI to get users to change their Tor browser's security settings? Relevant to this one edit—let's be sure to give them clear guidance that WILL assuredly protect them, and let them know that, if indeed it is decided to state that in its default config Tor will not conceal their use of Tor. It's ok to make extreme statements, but only if we offer direct remedies for users to take to protect themselves.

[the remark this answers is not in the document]
emkll: In this (very specific case), I don't think the Tor browser settings would change much to the overall risk. This is a subset of the first contact problem, where we need to bring more awareness about Tor and its properties more generally. This might, in the future, be useful language to include in landing pages, but I think that for now a disclaimer is warranted and also the best we can do. The wording seems quite good to me in 53e978e

docs/source.rst
emkll: Fair points, in restrospect the wording is definitely way too technical. I would however like to evoke the concept of segmenting work/personal and whistleblowing activities. Mixing these activities together greatly increases the odds of opsec failures. Either in this section or elsewhere in the document, it might be a useful concept to explain

docs/source.rst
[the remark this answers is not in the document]
emkll: Circling back to this comment, if https://github.com/freedomofpress/securedrop/pull/4880#discussion_r333578903 is overly technical, using git(lab) could, in some cases, also be considered overly technical.

The GitLab hosting resolves most of the issues the email approach solves (how to get tor if torproject.org is blocked). I'd be curious, from a source perspective, if there would be reticence to use email since they would either use their personal account or create a new one.

docs/source.rst
eloquence: I've tried to speak further to the opsec aspect in 8375ee541a8a9d57aa42aa7b841aa479c30b4a38, trying to use relatable examples of features and behaviors that may impact operational security.

ninavizz: Absolutely!! I would consider the audience this is written for, however; which is journalists and other people at host orgs. It's not a guide for sources, it's just a guide for using the Source UI. We still need to craft a "Source Guide," and I'd love to see that prioritized more highly. Because y'know, everything else we're doing is less of a priority (or something)? :/

docs/source.rst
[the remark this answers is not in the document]
ninavizz: Guys, what are the chances Torproject will be blocked—or, that anyone will be able to make sense of Gitlab, or know what GIt is? We're making a lot of assumptions here and designing for extreme worst-case scenarios. There _should_ be a content fork in there, for folks in at-risk locations where Tor may be blocked... but by default, I feel strongly that users need to be directed to the Tor website. Mostly because there is also helpful content framing what Tor is/does, etc. 

"Why shd I trust an installation from a place like this 'gitlab' place you're directing me to, vs the company that owns and makes Tor?" is the other reason. Non-technical folks just don't have a mental model of how the FOSS ecosystem works, and base their opsec decisions around anecdotal impressions we're trying to gently re-shape. 

Also, while folks in poorer regions may be using Linux, a majority of users will not. Nor will they be able to make sense of an installation that is not packaged for consumer-grade interaction.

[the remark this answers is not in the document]
eloquence: Note that the guide currently only speaks to the "likely to be monitored" scenario. That said, this is mostly old text, and I'm personally not fully convinced it's helpful. We say in the intro that the user should locate themselves in a place with Internet access they don't typically go to. Isn't that then also the best place to download Tor from? And in that case, are these instructions helping to mitigate that risk?

This may be lost in the diff, but my main goal with this edit was to fix obviously broken things, like the GitHub link that no longer works. While I understand how all these issues with the original text stand out during a review, I would recommend landing this change, then filing another issue for actually thinking through this Tor browser download scenario more carefully, and tailoring our advice to different audiences.

eloquence:
Added link to FPF guide (and some additional context) in 85e6e963eba3d41b56061ac4e8de35125bfacc05

PR #4890, opened 2019-10-09
Pin all dev and testing dependencies

### LINKED ISSUE #4686: We should run pip-compile with --allow-unsafe
## Description

Recent versions of `pip-tools` [add warnings when compiling requirements with `--generate-hashes` could result in uninstallable output](https://github.com/jazzband/pip-tools/pull/814). In our requirements files, we're getting these about `setuptools`, which `pip-tools` considers "unsafe" and refuses to hash. This means that we might not be able to update requirements in a virtualenv containing a version of `setuptools` too old to satisfy a requirement's new dependencies. Development virtualenvs can just be rebuilt, but this could break admin virtualenvs in the field.

We can persuade `pip-tools` to hash `setuptools` by adding the `--allow-unsafe` option wherever we invoke `pip-compile --generate-hashes`. This should allow trouble-free pip updates, and seems to be [safe according to at least one pypa developer](https://github.com/pypa/pip/issues/6459), though `pip-tools` [decided to be conservative and stop at the warnings](https://github.com/jazzband/pip-tools/issues/806).

emkll:
## Status

Ready for review

## Description of Changes

Towards #4686, pins hashes for test dependencies and pins versions/hashes of supervisor, pip and setuptools in develop-requirements.

As these dependencies are dev-facing, I did not perform a diff review (it was also unclear which version(s) were installed in the container). Happy to review them if the team judges it's necessary.

## Testing

- [x] The dev and admin-dev shells should be functional on a local developer's laptop.

## Deployment
Dev env only

As these dependencies are dev-facing, I did not review them. Happy to review them if the team judges it's necessary.


redshiftzero:
yep fine to not do diff review since they're dev/test only

I modified the PR description to indicate that this is towards #4686 but doesn't close since we're not pinning _all_ admin requirements (via using `--allow-unsafe`). I think we should do this because otherwise we're not pinning `setuptools` (pulled in by e.g. `ansible` in the admin workstation) and that could potentially cause breakage (not suggesting we fix for 1.1.0 but worth keeping open and addressing in a future PR). Let me know if you disagree.

PR #4911, opened 2019-10-15
[1.1.0] Fix changelog entries: docs updates were already included in 1.0.0

eloquence:
The docs updates re: HTTPS EV certs & the transfer/export docs PR were backported into 1.0.0 and are not novel to this release.

## Status

Ready for review

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -37,8 +37,7 @@
 
 ### Documentation
 
-* Transfer device and export recommendations (#4838).
-* Miscellaneous documentation fixes (#4844, #4853, #4874).
+* Miscellaneous documentation fixes (#4853, #4874).
 
 ## 1.0.0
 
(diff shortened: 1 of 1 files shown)


eloquence:
(NB: I've configured my editor to clean up trailing whitespace on save; it that's too annoying during review, I can turn it off, but I figure it's generally fine to clean up along the way.)

changelog.md
```diff
@@ -57,7 +56,7 @@
 ### Operations
 
 * Use dh-virtualenv and mod_wsgi to create securedrop-app-code package, run Python 3 version of the web applications on instances (#4622).
-* Adds support for v3 onion services for SecureDrop source, journalist, and SSH interfaces (#4652, #4710, #4690, #4675).
+* Adds support for v3 onion services for SecureDrop source, journalist, and SSH interfaces (#4652, #4710, #4690, #4675, #4844).
```
redshiftzero: hmm, the thing with these historical changelog edits is that we've already cut the 1.0.0 tag - let's just remove from the upcoming 1.1.0 release notes

eloquence: How about adding a 1.0.0 (Addendum) section above? Mostly concerned that #4838 in particular was a pretty hefty change (including to the threat model) and I want to make sure folks can find it if they look for it here.

redshiftzero: yeah i'm ok with this

redshiftzero:
OK after discussion with the RMs for 1.1.0 we decided not to add this addendum section as the extra overhead doesn't really seem worthwhile - we often merge docs changes after the tag is cut and we'd have an addendum section for most releases. versions of the docs from all recent versions are available on docs.securedrop.org if people want to see the docs from a particular release branch (if they don't want to just look at the release branch itself to see what got backported after release)

PR #4912, opened 2019-10-16
Fix securedrop-admin crash when no operation given under Python 3

### LINKED ISSUE #4910: securedrop-admin (setup) fails in Tails 4.0-rc1
## Description

Running `./securedrop-admin setup` in tails 4.0-rc1 (upgraded from 3.16) returns the following error:

```
INFO: Virtualenv already exists, not creating
INFO: Checking Python dependencies for securedrop-admin
ERROR: Failed to install pip dependencies. Check network connection and try again.
```
This was done in VMs, will try to reproduce on hardware.

## Steps to Reproduce

1. Upgrade Tails device from 3.16 to 4.0 (Boot into 4.0-rc1 drive and clone to upgrade 3.16 drive)
2. Boot into newly-upgraded tails drive
3. Verify that the internet is working (tor is bootstrapped, you can reach an external website over tor)
4. check out `1.1.0~rc2` tag
5. Run `./securedrop-admin setup` in ~/Persistent/securedrop
6. Observe error

## Expected Behavior

Securedrop-admin should run and the dependencies should be installed.

## Actual Behavior

Securedrop-admin fails and returns an error, the dependencies are not installed

## Comments

Per https://github.com/freedomofpress/securedrop/pull/4852/files#diff-b5e536cc161fcc0d62e661b4d6eae381R70-R73

When running the commands locally, I get
* `lsb_release --id --short` returns `Debian`
* `uname -a` returns `Linux amnesia 5.3.0-trunk-amd64 #1 SMF Debian 5.3.2-1~exp1 (2019-10-02) x86_64 GNU/Linux`

When i run ./securedrop-admin with no parameter, I get:
```
amnesia@amnesia:~/Persistent/securedrop$ ./securedrop-admin help
Could not find platform independent libraries <prefix>
Could not find platform dependent libraries <exec_prefix>
Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]
Fatal Python error: Py_Initialize: Unable to get the locale encoding
ImportError: No module named 'encodings'

Current thread 0x00007cf687450740 (most recent call first):
Aborted
```

rmol:
## Status

Ready for review

## Description of Changes

Fixes #4910.

Under Python 3, running securedrop-admin with no positional argument results in an ugly error, due to https://bugs.python.org/issue16308.

Under Python 3.7, we could simply add `dest` and `required` arguments to the `add_subparsers` call, but `required` isn't available in Python 3.5, so would break under Tails 3.

## Testing

Run `securedrop/bin/dev-shell bash` to start a shell in the dev container, then run:
- `cd ..`
- `./securedrop-admin setup`
- `./securedrop-admin -v`

You should see `Please specify an operation.` followed by the `securedrop-admin` help.

## Deployment

This only affects the admin workstation.

### DIFF
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -990,7 +990,12 @@ def parse_argv(argv):
                                             help=reset_admin_access.__doc__)
     parse_reset_ssh.set_defaults(func=reset_admin_access)
 
-    return set_default_paths(parser.parse_args(argv))
+    args = parser.parse_args(argv)
+    if getattr(args, 'func', None) is None:
+        print('Please specify an operation.\n')
+        parser.print_help()
+        sys.exit(1)
+    return set_default_paths(args)
 
 
 def main(argv):


emkll:
Thanks @rmol went through the fix and does what's advertised, `./securedrop-admin` now works and prints the help page, as expected.

I've also edited the PR description to say that it doesn't completely resolve #4910 : while it allows a user to run ./securedrop-admin with no argument, it does not address the issue that the venv in `admin/.ven3` needs to be destroyed/rebuilt for Tails 4. We should, as part of #4910 either
1. Provide docs for the 1.0.0 -> 1.1.0 release to indicate that the venv needs to be destroyed and rebuilt
2. Automate the process (this might be difficult as we cannot do this as part of the updater code, we would need to either create a new `./securedrop-admin` option or modify `./securedrop-admin setup` (check if tails3 and python version in venv is 3.5)

PR #4928, opened 2019-10-18
Add 1.0.0->1.1.0 upgrade guide

eloquence:
Includes instructions to rebuild the venv after the Tails 4 upgrade,
so securedrop-admin and the graphical updater continue to work with
Python 3.

Removed older upgrade guides along the way per standard procedure.

## Status

Ready for review 

## Test plan

The reviewer would follow the described procedure, or it should at least closely match the procedure used during QA.

## Checklint

- [x] Linter is as pleased as Punch


docs/upgrade/1.0.0_to_1.1.0.rst
```diff
@@ -0,0 +1,167 @@
...
+on your workstation by following these steps:
+
+- Change into the directory ``~/Persistent/securedrop/admin``.
+
+- Run the following commands from within this directory (make sure to type them
+  exactly as shown, as this is a destructive operation):
+
+.. code:: bash
+
+  rm -rf .venv
```
emkll: the `f` parameter was not necessary in my local testing

eloquence: dangerous force of habit -- fixed :)

PR #4932, opened 2019-10-18
Delete virtualenv if creation failed on setup

### LINKED ISSUE #4929: Python3 admin virtualenv is broken if improperly set up (doesn't contain pip3)
## Description

Initially discovered while reviewing https://github.com/freedomofpress/securedrop/pull/4927#issuecomment-543763957 , if the admin virtualenv is improperly setup, one must manually remove `admin/.venv3` to fix the virtualenv.

## Steps to Reproduce
This is very hard to reproduce, but you can do it by pressing ctrl+c when the virtualenv is being setup as part of an initial `./securedrop-admin setup` run. Subsequent `./securedrop-admin setup` invocations will no longer work

## Expected Behavior
`./securedrop-admin setup` should not fail

## Actual Behavior
`./securedrop-admin setup` fails with the following output:
```
amnesia@amnesia:~/Persistent/securedrop$ ./securedrop-admin setup
INFO: Virtualenv already exists, not creating
INFO: Checking Python dependencies for securedrop-admin
ERROR: Failed to install pip dependencies. Check network connection and try again.
```

emkll:
## Status

Ready for review

## Description of Changes

Fixes #4929 

## Testing

This test requires disconnecting the network to ensure the exception is properly thrown.
In both Tails 3 and Tails 4:
- remove the existing admin venv: `rm -r admin/.venv3`
- run `./securedrop-admin setup`
- While or immediately after apt dependencies are installed (preferably before `[INFO] Setting up virtualenv` appears) disconnect the network.
- The setup will fail after 60s timeout
- [ ] admin/.venv3 folder was correctly deleted

## Deployment

This will be deployed to new and existing installs via git.

### DIFF
--- a/admin/tests/test_securedrop-admin-setup.py
+++ b/admin/tests/test_securedrop-admin-setup.py
@@ -119,3 +119,19 @@ class TestSecureDropAdmin(object):
                                                                       ':o')):
                 bootstrap.clean_up_tails3_venv(venv_path)
                 assert os.path.exists(venv_path)
+
+    def test_envsetup_cleanup(self, tmpdir, caplog):
+        venv = os.path.join(str(tmpdir), "empty_dir")
+        args = ""
+        with pytest.raises(subprocess.CalledProcessError):
+            with mock.patch('subprocess.check_output',
+                            side_effect=self.side_effect_venv_bootstrap(venv)):
+                bootstrap.envsetup(args, venv)
+                assert not os.path.exists(venv)
+                assert 'Cleaning up virtualenv' in caplog.text
+
+    def side_effect_venv_bootstrap(self, venv_path):
+        # emulate the venv being created, and raise exception to simulate
+        # failure in virtualenv creation
+        os.makedirs(venv_path)
+        raise subprocess.CalledProcessError(1, ':o')
--- a/admin/bootstrap.py
+++ b/admin/bootstrap.py
@@ -187,6 +187,9 @@ def envsetup(args):
             sdlog.debug(e.output)
             sdlog.error(("Unable to create virtualenv. Check network settings"
                          " and try again."))
+            sdlog.debug("Cleaning up virtualenv")
+            if os.path.exists(VENV_DIR):
+                shutil.rmtree(VENV_DIR)
             raise
     else:
         sdlog.info("Virtualenv already exists, not creating")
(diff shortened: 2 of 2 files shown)


admin/tests/test_securedrop-admin-setup.py
```diff
@@ -119,3 +119,14 @@ def test_venv_cleanup_subprocess_exception(self, tmpdir, caplog):
                                                                       ':o')):
                 bootstrap.clean_up_tails3_venv(venv_path)
                 assert os.path.exists(venv_path)
+
+    def test_envsetup_cleanup(self, tmpdir, caplog):
+        venv_path = os.path.join(str(tmpdir), "empty_dir")
+        args = ""
+        with pytest.raises(subprocess.CalledProcessError):
+
+            with mock.patch('subprocess.check_output',
```
redshiftzero: OK so this diff looks good. one issue is that the test plan is not quite going to do the trick since at least for me `securedrop-admin setup` never times out (on Tails 4 beta2) so I'd need to ctrl-c which means the two cleanup lines won't execute. I have reproduced the original scenario however while testing other functionality on Tails, and I did need to manually remove the `admin/.venv3` directory in order to get `securedrop-admin setup` to complete on a subsequent call, so I think we should merge this.

Given that it's hard to test this scenario in Tails, we should ensure that the cleanup lines are being executed in this test by modifying this unit test to - as a side effect of the mocked `subprocess.check_output` call - create `venv_path` on disk. Then this test will verify that the cleanup works as expected.

admin/tests/test_securedrop-admin-setup.py
```diff
@@ -119,3 +119,19 @@ def test_venv_cleanup_subprocess_exception(self, tmpdir, caplog):
...
+                            side_effect=self.side_effect_venv_bootstrap(venv)):
+                bootstrap.envsetup(args, venv)
+                assert not os.path.exists(venv)
+                assert 'Cleaning up virtualenv' in caplog.text
+
+    def side_effect_venv_bootstrap(self, venv_path):
+        # emulate the venv being created, and raise exception to simulate
+        # failure in virtualenv creation
+        os.makedirs(venv_path)
+        raise subprocess.CalledProcessError(1, ':o')
```
redshiftzero: beautiful, thanks!

PR #4970, opened 2019-11-05
updated keepassx setup instructions to use Tails default location

### LINKED ISSUE #4335: Rename KeePassX template filename to match default KeePassX filename and adjust docs
## Description

Currently, the included template KeePassX DB file in `/tails_files` is named `securedrop-keepassx.kdbx` and the documentation asks admins to copy and save this file to whatever they want to name it to in the documentation and save it wherevs inside `~/Persistent`. The file should be renamed in the git repository to `keepassx.kdbx` and the docs should be updated to specifically ask that it be copied to the root of `~/Persistent` so that the file is automatically loaded when KeePassX is started rather than having to ask users to manually hunt for a database file in a desktop environment they're unfamiliar with. Tails went through an effort to make this work so that users who are already new to password managers could have one less thing they have to manually deal with (See: https://redmine.tails.boum.org/code/issues/12375#note-18).

## User Research Evidence

In observation of SecureDrop admins and journalists at SecureDrop trainings, many get frustrated with the amount of steps it takes to get to log in to the journalist interface. To remove this step and reduce the the confusion/surprise/"What do I do here", it should already be named the default/preferred KeePassX filename and located where Tails can load it automatically. Hunting for a DB file as a prerequisite to seeing your passwords is an alien process even for people using modern password managers.

## User Stories

As an admin or journalist, I want to be able to log in to the SecureDrop journalist interface without having to try to remember what a KeePassX database file is or where I put it/what it was named.

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #4335.

Updates the instructions for setting up the keepassx database, to use the Tails default location and save the user having to explicitly navigate to and open the correct database every time.

## Testing

Docs-only PR, review for clarity and correctness.

## Deployment

Deployed by docs build.

### DIFF
--- a/docs/set_up_admin_tails.rst
+++ b/docs/set_up_admin_tails.rst
@@ -154,23 +154,32 @@ store them securely. Once you have set up Tails with persistence and
 have cloned the repo, you can set up your personal password database
 using this template.
 
-You can find the template in ``tails_files/securedrop-keepassx.kdbx``
-in the SecureDrop repository that you just cloned.
-
 .. note::
 
    Earlier versions of Tails used KeePassX instead of KeePassXC.
    The provided template is compatible with both.
 
-To use the template:
+You can find the template in ``tails_files/securedrop-keepassx.kdbx``
+in the SecureDrop repository that you just cloned. To use the template:
+
+-  Copy the template to the Persistent folder - from a terminal, run the
+   command:
+
+   .. code:: sh
+
+     cp ~/Persistent/securedrop/tails_files/securedrop-keepassx.kdbx \ 
+        ~/Persistent/keepassx.kdbx
 
 -  Open the KeePassXC program |KeePassXC| which is already installed on
    Tails
 -  Select **Database**, **Open database**, and navigate to the location of
-   **securedrop-keepassx.kdbx**, select it, and click **Open**
+   **~/Persistent/keepassx.kdbx**, select it, and click **Open**
 -  Check the **password** box and hit **OK**
--  Click **Database** and **Save Database As**
--  Save the database in the Persistent folder
+-  Edit entries as required.
+-  Select **Database** and **Save Database** to save your changes.
+
+The next time you use KeepassXC, the database at ``~/Persistent/keepassx.kdbx``
+will be opened by default.
 
 .. tip:: If you would like to add a master password, navigate to **Database** and
    **Change master key**. Note that since each KeePassXC database is stored


huertanix:
Not exactly related, but this issue may/may not lead to some changes in Tails creating a default DB in this location when Persistence is activated in the future: https://redmine.tails.boum.org/code/issues/17212.

docs/set_up_admin_tails.rst
```diff
@@ -164,13 +164,24 @@ in the SecureDrop repository that you just cloned.
 
 To use the template:
 
+-  Copy the template to the Persistent folder - from a terminal, run the
+   commands:
+
+   .. code:: sh
+
+     cd ~/Persistent
+     cp securedrop/tails_files/keepassx.kdbx keepassx.kdbx
```
huertanix: Should we use absolute paths here instead of cd'ing into the destination?

docs/set_up_admin_tails.rst
```diff
@@ -164,13 +164,24 @@ in the SecureDrop repository that you just cloned.
...
-   **securedrop-keepassx.kdbx**, select it, and click **Open**
+   **~/Persistent/keepassx.kdbx**, select it, and click **Open**
 -  Check the **password** box and hit **OK**
--  Click **Database** and **Save Database As**
--  Save the database in the Persistent folder
+-  Edit entries as required.
+-  Select **Database** and **Save Database** to save your changes.
+
+The next time you use KeepassXC, the database at ``~/Persistent/keepass.kdbx``
+will be opened by default.
```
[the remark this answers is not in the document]
zenmonkeykstop: Yup, I thought about adding that as it happens. Will do.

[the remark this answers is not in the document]
zenmonkeykstop: Actually ended up renaming the template back to securedrop-keepassx.kdbx, as I was worried about the case where someone edited it in place by mistake. Better for the updater to throw an error than for it to silently fail and for a user to possibly lose their password db later.

docs/set_up_admin_tails.rst
zenmonkeykstop: Done. It makes the command a bit long, which is why I was initially against it, so it's now split across two lines.

docs/set_up_admin_tails.rst
```diff
@@ -154,23 +154,32 @@ store them securely. Once you have set up Tails with persistence and
...
 -  Select **Database**, **Open database**, and navigate to the location of
-   **securedrop-keepassx.kdbx**, select it, and click **Open**
+   **~/Persistent/keepassx.kdbx**, select it, and click **Open**
 -  Check the **password** box and hit **OK**
--  Click **Database** and **Save Database As**
--  Save the database in the Persistent folder
+-  Edit entries as required.
+-  Select **Database** and **Save Database** to save your changes.
+
+The next time you use KeepassXC, the database at ``~/Persistent/keepass.kdbx``
```
redshiftzero: typo? `~/Persistent/keepass.kdbx` -> `~/Persistent/keepassx.kdbx`

PR #4986, opened 2019-11-10
Update instructions for current version of Tor Browser

### LINKED ISSUE #4985: Update "Using Tor Browser with the Development Environment" docs for Tor Browser 9.x
## Description

The documentation for [Using Tor Browser with the Development Environment](https://docs.securedrop.org/en/latest/development/tips_and_tricks.html#using-tor-browser-with-the-development-environment) is no longer correct for Tor Browser 9.x. At some point since the docs were written, the Tor Browser preferences UI changed and the settings the documentation refers to are no longer available.

Fortunately, I poked around for a bit and was able to find an alternative method for allowing Tor Browser to bypass the proxy and access localhost. PR incoming.

garrettr:
## Status

Ready for review.

## Description of Changes

Fixes #4985.

Documents an alternative method of allowing Tor Browser to access localhost addresses that works in the current version of Tor Browser (9.0.1).

## Testing

Follow the updated instructions and confirm that this allows you to access localhost address in the current version of Tor Browser.

## Deployment

This change only affects developer documentation, so there are no deployment considerations.

### DIFF
--- a/docs/development/tips_and_tricks.rst
+++ b/docs/development/tips_and_tricks.rst
@@ -20,20 +20,20 @@ development. Unfortunately, it is not possible to access the local
 development servers by default, due to Tor Browser's proxy
 configuration.
 
-To test the development environment in Tor Browser, you need to add an
-exception to allow Tor Browser to access localhost:
+To test the development environment in Tor Browser, you need to modify Tor
+Browser's default settings to prevent localhost from being resolved by the
+proxy:
 
-#. Open the "Tor Browser" menu and click "Preferences..."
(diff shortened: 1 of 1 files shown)


garrettr:
I just noticed that `functional_test.py` uses the `network.proxy.no_proxies_on` to achieve the same result. I think `network.proxy.allow_hijacking_localhost` is equivalent and seems easier to get right, which is important for documentation of steps that will be carried out by a human rather than a script.

If anybody knows why we should prefer one pref over the other, I'd be interested to hear about it!

PR #4987, opened 2019-11-11
Removes journalist UI modal test xfail.

### LINKED ISSUE #4978: remove xfail on test_journalist_interface_ui_with_modal
## Description

In 2713cff7206296675140aa0b418f8c36b9197458 we xfailed `test_journalist_interface_ui_with_modal` due to that functionality not working well with the latest version of Tor Browser (9.0.1). This ticket is to investigate that functionality and ensure that it works again with latest Tor Browser such that we can unxfail that test.

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Fixes #4978

Adds a function to set TBB security level explicitly, uses it to switch TBB to correct state for journalist functional tests.

Change required because TBB 9.0.1 is being started with Javascript disabled. Rather than getting into managing individual preferences related to browser functionality, it makes sense to simulate user behaviour by toggling the security slider pref to get TBB 9.0.1 into the correct state. (This would also be useful for testing security warnings on the source interface, or any other featurest that rely on specific TBB security levels.)

## Testing

- [ ] Does CI pass, including `test_journalist_interface_ui_with_modal`?

## Deployment

Testing-only.

## Checklist

### DIFF
--- a/securedrop/tests/functional/functional_test.py
+++ b/securedrop/tests/functional/functional_test.py
@@ -28,6 +28,8 @@ from selenium.webdriver.common.by import By
 from selenium.webdriver.remote.remote_connection import LOGGER
 from selenium.webdriver.support import expected_conditions
 from selenium.webdriver.support.ui import WebDriverWait
+from selenium.webdriver.common.action_chains import ActionChains
+from selenium.webdriver.common.keys import Keys
 from sqlalchemy.exc import IntegrityError
 from tbselenium.tbdriver import TorBrowserDriver
 from tbselenium.utils import disable_js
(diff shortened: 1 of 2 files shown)


securedrop/tests/functional/functional_test.py
```diff
@@ -77,6 +79,28 @@ def _unused_port(self):
...
+            self.create_torbrowser_driver()
+        driver = self.torbrowser_driver
+
+        driver.get("about:config")
+        accept_risk_button = driver.find_element_by_id("warningButton")
+        if accept_risk_button:
+            accept_risk_button.click()
+        ActionChains(driver).send_keys(Keys.RETURN).\
+            send_keys("extensions.torbutton.security_slider").perform()
+        time.sleep(2)
```
redshiftzero: this looks great - one thought is that sleeps can be flaky (i.e. on slow machines), can we use `wait_for` here?

securedrop/tests/functional/test_journalist.py
```diff
@@ -59,14 +58,17 @@ def test_journalist_uses_index_delete_collections_button_modal(self):
...
-    @pytest.mark.xfail(reason="Filter widget not displayed in TBB 9.0.1")
     def test_journalist_interface_ui_with_modal(self):
         self._source_visits_source_homepage()
         self._source_chooses_to_submit_documents()
         self._source_continues_to_submit_page()
         self._source_submits_a_file()
         self._source_logs_out()
 
+        # Toggle security slider to force prefs change
+        self.set_tbb_securitylevel(1)
```
redshiftzero: it's unclear where these numbers came from, can we define these numbers as constants somewhere and reference them here?

securedrop/tests/functional/functional_test.py
zenmonkeykstop: Added in a couple, but didn't see an easy way to do so on the XUL panel changes from Selenium. Slowness *shouldn't* be an issue as these are local UI updates, not network calls.

securedrop/tests/functional/test_journalist.py
zenmonkeykstop: Done.

Issue #2331, opened 2017-09-18
Improve consistency of multi-factor auth terminology


heartsucker:
# Feature request

## Description

There are several place in the docs and apps where "two factor {code,token,credentials}" are used seemingly interchangeably.

We should use

- two factor secret: for the shared secret
- two factor code: the actual OTP

Google uses the phrase "authentication code" and we should at least stick to the phrase "code" in our docs / app.

This should also be included with variable names in the app. Using `totp_token` to parse what the docs call a "two factor code" is unnecessarily confusing.

## User Stories

As a translator, I want to have a smaller number of glossary terms to maintain.

As a (non-technical) user, I want to be able to simply follow directions without having to know if a two factor token is the same as a two factor code.

heartsucker:
This include the 2FA verification screen on the journalist app that says "Verification code."

rmol:
Our terminology for everything related to multi-factor authentication is internally inconsistent, doesn't match some popular web services that users might have experience with, and is wasting a lot of translator effort. 

I'd like to add to this ticket normalization of the terms for devices used in two-factor authentication. A  quick survey suggests that "security key" is the most common term for them:
- [Facebook](https://www.facebook.com/help/401566786855239)
- [GitHub](https://help.github.com/en/github/authenticating-to-github/configuring-two-factor-authentication#configuring-two-factor-authentication-using-a-security-key)
- [Google](https://support.google.com/accounts/answer/7539956)
- [Microsoft](https://support.microsoft.com/en-us/help/4463210/windows-10-sign-in-microsoft-account-windows-hello-security-key)

We should settle on that in code and docs.

ninavizz:
WRT "Security Key," I feel Duo does it very well on this page. Which, I realize, is only one part of the problem... https://guide.duo.com/u2f

eloquence:
Nina and I discussed this today, to ensure that we're agreeing on consistent terminology we can use for the SecureDrop Workstation as well. The proposal here makes sense to us:

- Two-Factor Code - (with "two" always spelled out) for the numeric codes the user enters
- Two-Factor Secret - for the seed used to manually configure a two-factor authentication device, in lieu of scanning a QR code
- Security Key - for dedicated physical security devices like YubiKeys that can be used to generate two-factor code;

A couple of notes on usage:

[...]

PR #5014, opened 2019-11-21
More consistent usage of two-factor related terminology

eloquence:
Towards #2331 

- Use "two-factor code" when referring to 6 digit codes
- Use "two-factor secret" for 2FA seeds
- Use "two-factor credentials" for reset operations
- Use "two-factor device" as umbrella term for YubiKey OR mobile
  app, but minimize usage and always explain it, to avoid
  confusion
- Avoid use of "authenticator"
- Avoid use of "token" in 2FA contexts
- Use "security key" to describe physical devices like YubiKeys
- Clearer error/success messages using this language
- Clearer instructions to the user on 2FA screens

## Status

Work in progress

### DIFF
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -163,7 +163,7 @@ enter.
 
 If FreeOTP was set up correctly, you will be redirected
 back to the Admin Interface and will see a confirmation that the
-two-factor token was verified.
+two-factor code was verified.
 
 .. include:: includes/otp-app.txt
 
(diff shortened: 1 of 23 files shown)


eloquence:
Opening for discussion, will need to update screenshots and potentially a couple more places in docs if this language looks good to everyone.

redshiftzero:
is this ready for final review? CI is passing

eloquence:
Screenshot updates incoming later today.

PR #5046, opened 2019-11-29
add confirmation flash when submission preferences are saved

### LINKED ISSUE #5009: Add confirmation message after updating submission preferences
#4879 made it possible to prevent sources from uploading documents by changing a setting in the *Admin Interface*:

Screen Shot 2019-11-19 at 16 41 15-fullpage

However, there is no confirmation message after changing this setting. At minimum, we should add a generic flashed message, which we could use when future preferences added to this page are changed, e.g., "Settings have been updated successfully."

# User Story

As an admin, I want to know at a glance that my actions on the preferences page have had the desired effect, so that I don't have to re-examine the state of specific checkboxes to be sure.

petevdp:
## Status

Ready for review

## Description of Changes

Fixes #5009.

Add a flashed message when a user succesfully submits a change with the submission preferences form. The current implementation doesn't differentiate between submissions which actually change anything, but that should be easy enough if it would be desirable.

### manual
* navigate to /admin/config
* choose a submission preference by checking/unchecking box
* hit 'update submission preferences'

A flashed message should show up, indicating that the setting has been successfully saved. 

## automated

I'm unsure if this feature needs automated testing. I'll be happy to add some if required!

- [x ] Linting (`make lint`) and tests (`make test`) pass in the development container

Unsure if I need to do this, as mentioned above.

### DIFF
--- a/securedrop/journalist_app/admin.py
+++ b/securedrop/journalist_app/admin.py
@@ -59,7 +59,7 @@ def make_blueprint(config):
         form = SubmissionPreferencesForm()
         if form.validate_on_submit():
             # The UI prompt ("prevent") is the opposite of the setting ("allow"):
-            flash(gettext("Preferences Saved."), "submission-preferences-success")
+            flash(gettext("Preferences saved."), "submission-preferences-success")
             value = not bool(request.form.get('prevent_document_uploads'))
             InstanceConfig.set('allow_document_uploads', value)
             return redirect(url_for('admin.manage_config'))
--- a/securedrop/journalist_app/admin.py
+++ b/securedrop/journalist_app/admin.py
@@ -59,6 +59,7 @@ def make_blueprint(config):
         form = SubmissionPreferencesForm()
         if form.validate_on_submit():
             # The UI prompt ("prevent") is the opposite of the setting ("allow"):
+            flash(gettext("Preferences Saved."), "submission-preferences-success")
             value = not bool(request.form.get('prevent_document_uploads'))
             InstanceConfig.set('allow_document_uploads', value)
             return redirect(url_for('admin.manage_config'))
--- a/securedrop/journalist_templates/submission_preferences_saved_flash.html
+++ b/securedrop/journalist_templates/submission_preferences_saved_flash.html
@@ -1,0 +1,10 @@
+{# these are flashed messages for the logo upload file verifiaction #}
+{% with messages = get_flashed_messages(with_categories=True, category_filter=["submission-preferences-success"]) %}
+  {% for category, message in messages %}
+    {% set category_status = category[23:] %}
+      <div class="flash {{ category_status }}">
+        <img src="{{ url_for('static', filename='i/success_checkmark.png') }}" height="17" width="20">
+        {{ message }}
+      </div>
+  {% endfor %}
+{% endwith %}
--- a/securedrop/journalist_templates/submission_preferences_saved_flash.html
+++ b/securedrop/journalist_templates/submission_preferences_saved_flash.html
@@ -1,6 +1,7 @@
-{# these are flashed messages for the logo upload file verifiaction #}
 {% with messages = get_flashed_messages(with_categories=True, category_filter=["submission-preferences-success"]) %}
   {% for category, message in messages %}
+{#  Get the end of the of the category message which
+    contains the category status.(success/error)#}
     {% set category_status = category[23:] %}
       <div class="flash {{ category_status }}">
         <img src="{{ url_for('static', filename='i/success_checkmark.png') }}" height="17" width="20">
--- a/securedrop/journalist_templates/config.html
+++ b/securedrop/journalist_templates/config.html
@@ -54,6 +54,7 @@
   <button type="submit" id="submit-submission-preferences">
     <i class="fas fa-pencil-alt"></i> {{ gettext('UPDATE SUBMISSION PREFERENCES') }}
   </button>
+  {% include 'submission_preferences_saved_flash.html' %}
 </form>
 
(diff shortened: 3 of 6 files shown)


securedrop/journalist_templates/submission_preferences_saved_flash.html
```diff
@@ -0,0 +1,10 @@
+{# these are flashed messages for the logo upload file verifiaction #}
```
eloquence: tiny typo: verifiaction->verification

securedrop/journalist_app/admin.py
```diff
@@ -59,6 +59,7 @@ def update_submission_preferences():
         form = SubmissionPreferencesForm()
         if form.validate_on_submit():
             # The UI prompt ("prevent") is the opposite of the setting ("allow"):
+            flash(gettext("Preferences Saved."), "submission-preferences-success")
```
eloquence: Saved->saved (for consistency with other messages)

eloquence:
Hi @petevdp, thanks for working on this! From my end, except for the two nits noted above, this looks good. Please give us a few days and a maintainer should be able to have a closer look.

From a UX perspective, it's a bit problematic that the confirmation only is noticeable when you scroll all the way to the bottom of the page:
Screenshot from 2019-12-03 17-41-31

That said, this in-line placement is consistent with how we handle logo updates on the same page. We may want to revisit that separately but I'd personally be OK with doing this in follow-up. CC @ninavizz for UX visibility.

petevdp:
@eloquence Yeah it seems like we might want to add a more generalized system for the sort of confirmation/notification system that would be good for situations like this. Maybe something JS based, or a redirect to another page with the desired message.

That sounds like it might justify creating a separate issue, but I'd still like to work on that.

securedrop/journalist_templates/submission_preferences_saved_flash.html
petevdp: f65646524ddf03548dd33ff4674bace15c894442

securedrop/journalist_app/admin.py
petevdp: 2d95ebf77522b070d28e22b6db3d7f6f14165b26

securedrop/journalist_templates/submission_preferences_saved_flash.html
redshiftzero: let's update or remove this comment (these are the flashed messages for the submission file preferences)

securedrop/journalist_templates/submission_preferences_saved_flash.html
```diff
@@ -0,0 +1,10 @@
...
+{% with messages = get_flashed_messages(with_categories=True, category_filter=["submission-preferences-success"]) %}
+  {% for category, message in messages %}
+    {% set category_status = category[23:] %}
```
redshiftzero: I realize that it's a magic number also in `securedrop/journalist_templates/logo_upload_flashed.html` but can you add a comment explaining where the 23 comes from?

securedrop/journalist_app/admin.py
```diff
@@ -59,6 +59,7 @@ def update_submission_preferences():
         form = SubmissionPreferencesForm()
         if form.validate_on_submit():
             # The UI prompt ("prevent") is the opposite of the setting ("allow"):
+            flash(gettext("Preferences saved."), "submission-preferences-success")
```
redshiftzero: the places to update such that this line is under test are: 
* in `securedrop/tests/functional/journalist_navigation_steps.py` there is a class `JournalistNavigationStepsMixin` which contains the steps used for our selenium tests. The methods you want to look at are ` _admin_disallows_document_uploads` and `_admin_allows_document_uploads`. [Here](https://github.com/freedomofpress/securedrop/blob/de48b1b9f9a7a7426499ec7177e23ac5d7f16925/securedrop/tests/functional/journalist_navigation_steps.py#L277) is an example where we are asserting that a flashed message appears. 
* in `securedrop/tests/test_journalist.py` there are two unit tests for updating the submission preferences in `test_prevent_document_uploads` and `test_no_prevent_document_uploads`. You can add an assert here too ([example](https://github.com/freedomofpress/securedrop/blob/de48b1b9f9a7a7426499ec7177e23ac5d7f16925/securedrop/tests/test_journalist.py#L1385)).

eloquence:
Hi @petevdp, just touching base - do you have time to poke further at this? If not we can finish it up. Just let us know :)

petevdp:
@eloquence Yeah I do, sorry for the lack of communication. I'm planning on working on it tomorrow!

eloquence:
Awesome, thanks for the update! :)

securedrop/journalist_templates/submission_preferences_saved_flash.html
petevdp: a8b2dc6e42c8056487ca16d95001295e794c5529 I added the comment here. I think there's probably a better way to word this though, so if you have a suggestion let me know.

zenmonkeykstop:
Thanks for the update @petevdp - pushed little a commit to fix linter errors, will review once CI is passing.

petevdp:
Awesome! Thanks for everyone's help, and sorry for the delay.

DrGFreeman:
Hi @petevdp, kudos for your contribution (got an email notification as I had commented on the issue). 

As a new contributor to the project, I found this [pre-commit hook](https://docs.securedrop.org/en/latest/development/contributor_guidelines.html#code-style) quite useful to avoid forgetting the linting checks.

petevdp:
@DrGFreeman Okay I'll install that, thanks!

PR #4496, opened 2019-05-31
Redirect to index after session expiration on /generate

redshiftzero:
## Status

Ready for review

## Description of Changes

Fixes #4490. 

Previous logic when sessions expire on `/generate`:
1. On `/generate` page, codename is added to the session.
2. When expiration occurs, session is cleared.
3. A flashed message indicating that the session has expired
would be added to the current session (this is done [via passing
`_flashes` on Flask's `session` object](https://github.com/pallets/flask/blob/cd4023d9d2ab630ce4f95856f065072ef8badb2b/flask/helpers.py#L449)).
4. Execution enters the view function associated with `/create`.
But `/create` expects codename to be in the session (which was
cleared in step 2), thus a `KeyError` will occur.

If you're wondering "but why does step 4 even happen" - it's because we don't immediately redirect to the index when a session expires. When the session expires when the `@login_required` decorator is on the view function corresponding to the request path, execution doesn't end up entering the view function itself, as the logic in `@login_required` will redirect the user to the `login` page if `logged_in` is not in the `session` object (and `logged_in` won't be in the `session` object, because it gets cleared in step 2). 

Logic with this diff when sessions expire on `/generate`:
1. On `/generate` page, codename is added to the session
2. When expiration occurs, session is cleared and user is immediately redirected to the index.

## Testing

1. Follow STR in #4490, confirm bug is fixed
2. Revert fix, make sure regression test fails

## Deployment

No special instructions (deployed via `securedrop-app-code` deb)

### If you made changes to the server application code:

- [x] Linting (`make lint`) and tests (`make -C securedrop test`) pass in the development container 

### If you made non-trivial code changes:

- [x] I have written a test plan and validated it for this PR

PR #5048, opened 2019-12-01
Handle multiple instances of source /generate

### LINKED ISSUE #4458: Duplicate Source Interface session triggers server error

DrGFreeman:
## Status

Work in progress

## Description of Changes

Fixes #4458.

When the source opens more than one browser tabs on /generate before clicking on SUBMIT DOCUMENTS:
- Flashes a message promting the source to verify their codename in /lookup after they click on SUBMIT DOCUMENTS in a fist instance of /generate.
- Redirects the source to /lookup when they click on SUBMIT DOCUMENTS in a second instance of /generate and displays a flash message prompting the source to verify their codename.
- Makes the codename in /lookup visible by default.

This eliminates the 500 server error reported in #4458 and reduces the potential of codename confusion following the generation of multiple codenames in different browser tabs or windows.

The behavior when only one instance of /generate is opened remains unchanged.

## Testing

## Deployment

2. New installs.

Choose one of the following:

### DIFF
--- a/securedrop/source_app/main.py
+++ b/securedrop/source_app/main.py
@@ -36,6 +36,12 @@ def make_blueprint(config):
                   "notification")
             return redirect(url_for('.lookup'))
 
+        # Note if a codename was already generated
+        if 'codename' in session:
+            session['multiple_codenames'] = True
+        else:
+            session['multiple_codenames'] = False
+
         codename = generate_unique_codename(config)
         session['codename'] = codename
         session['new_user'] = True
@@ -51,34 +57,40 @@ def make_blueprint(config):
 
     @view.route('/create', methods=['POST'])
     def create():
-        filesystem_id = current_app.crypto_util.hash_codename(
-            session['codename'])
-
-        source = Source(filesystem_id, current_app.crypto_util.display_id())
-        db.session.add(source)
-        try:
-            db.session.commit()
-        except IntegrityError as e:
-            db.session.rollback()
-            current_app.logger.error(
-                "Attempt to create a source with duplicate codename: %s" %
-                (e,))
-
-            # Issue 2386: don't log in on duplicates
-            del session['codename']
-
-            # Issue 4361: Delete 'logged_in' if it's in the session
-            try:
-                del session['logged_in']
-            except KeyError:
-                pass
-
-            abort(500)
+        if logged_in():
+            flash(gettext("You have already logged-in from a different browser tab.\n"
+                          "Please verify your codename below as it may differ from "
+                          "the one displayed on the previous page"),
+                  'notification')
         else:
-            os.mkdir(current_app.storage.path(filesystem_id))
+            if session['multiple_codenames']:
+                flash(gettext("Multiple codenames have been generated in different browser "
+                              "tabs. Please verify your codename below as it may differ from "
+                              "the one displayed on the previous page."),
+                      'notification')
 
-        session['logged_in'] = True
-        return redirect(url_for('.lookup'))
+            filesystem_id = current_app.crypto_util.hash_codename(
+                session['codename'])
+
+            source = Source(filesystem_id, current_app.crypto_util.display_id())
+            db.session.add(source)
+            try:
+                db.session.commit()
+            except IntegrityError as e:
+                db.session.rollback()
+                current_app.logger.error(
+                    "Attempt to create a source with duplicate codename: %s" %
+                    (e,))
+
+            else:
+                os.mkdir(current_app.storage.path(filesystem_id))
+
+            session['logged_in'] = True
+
+        if session['multiple_codenames']:
+            return redirect(url_for('.lookup', _anchor='codename-hint-visible'))
+        else:
+            return redirect(url_for('.lookup'))
 
     @view.route('/lookup', methods=('GET',))
     @login_required
--- a/securedrop/source_app/main.py
+++ b/securedrop/source_app/main.py
@@ -60,7 +60,7 @@ def make_blueprint(config):
         if logged_in():
             flash(gettext("You have already logged-in from a different browser tab.\n"
                           "Please verify your codename below as it may differ from "
-                          "the one displayed on the previous page"),
+                          "the one displayed on the previous page."),
                   'notification')
         else:
             if session['multiple_codenames']:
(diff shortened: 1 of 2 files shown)


DrGFreeman:
This PR follows the discussions in #4496.

It is a work in progress with placeholder messages for which I would like to get feedback from the team before going forward in updating the tests and implementing new ones.

Here is a demonstration of the proposed behavior:

1. Source opens `/generate` in tab A. Codename X is in session.
image
1. Source opens `/generate` in tab B. Codename Y is in session.
image
1. Source returns to tab A and clicks "SUBMIT DOCUMENTS". Codename Y is used to register the source. **Changes relative to the current behavior**: A flash message is added and the codename is made visible:
image
1. Source returns to tab B and clicks "SUBMIT DOCUMENTS". **Changes relative to current behavior**: No server error (500). The source is redirected to `/lookup`, a flash message is displayed and the codename is made visible:
image

If, alternatively, the source clicked "SUBMIT DOCUMENTS" in tab B first and then in tab A, the result would be similar except the flashed messages would be interverted.

There is no change in the interface behavior when only one `/generate` page was opened (i.e. no flash message, codename hidden).

Also, an alternative to making the codename visible would be to add a link in the flash message similar to when the source makes their first submission:
image

zenmonkeykstop:
From a UX perspective, this LGTM. Tested as follows:

## 2x get started

- ra `make dev` 
- opened 2 tabs pointing at `localhost:8080` and clicking through **Get Started** on both. Two different codenames generated
- clicked thru **Submit Documents** on first tab. First tab's codename displayed on reminder. Successful submit.
- clicked thru **Submit Documents** on second tab. Flash message displayed and first tab's codename displayed on reminder. Successful submit.
- logged out on first tab, got returned to Source Interface homepage
- tried to submit on second tab, got returned to SI homepage with expired session flash message (that might need to be updated to have more generic language as it says the session timed out due to inactivity).

## get started, login in separate tabs

- opened 2 tabs pointing at `localhost:8080` and clicked **Get Started** on one, **Log in** on the other. 
- logged in on 1st tab using existing codename - login successful
- clicked **Submit Documents** on 2nd tab. Flash message displayed and first tab's codename displayed on reminder.

## JI check
- checked Journalist interface, both submissions are present under a single codename corresponding with the SI codename (tested using replies).

DrGFreeman:
Thanks for the comments @zenmonkeykstop.

I marked the PR as WIP to make sure it was going in the right direction following the comments in #4996.

I still want to add functional tests before it is merged and would also appreciate feedback on the wording of the flash messages as I am not native English speaker and there may be better ways to phrase these.

securedrop/source_app/main.py
```diff
@@ -51,34 +57,40 @@ def select_logo():
...
-        except IntegrityError as e:
-            db.session.rollback()
-            current_app.logger.error(
-                "Attempt to create a source with duplicate codename: %s" %
-                (e,))
-
-            # Issue 2386: don't log in on duplicates
-            del session['codename']
-
-            # Issue 4361: Delete 'logged_in' if it's in the session
```
redshiftzero: confirming that #4391 was a fix required when we were deleting `session['codename']` in this view function and so is good to remove now

securedrop/source_app/main.py
```diff
@@ -51,34 +57,40 @@ def select_logo():
...
+        if logged_in():
+            flash(gettext("You have already logged-in from a different browser tab.\n"
+                          "Please verify your codename below as it may differ from "
+                          "the one displayed on the previous page"),
```
zenmonkeykstop: nit: End sentence with '.'

zenmonkeykstop:
The flash messages look good to me with the exception of one period :). With functional test updates, this should be good to go.

eloquence:
I think we're definitely closer to a resolution here, thanks for all your hard work on this @DrGFreeman. If some version of this lands, I'd like to give @ninavizz a chance to help with the final UI messaging at minimum. She suggested in the other PR that we might want to display the codename inline in the message; that could potentially be handled as a separate issue, though.

I do think we should talk a bit more whether this is the best approach, though. I'm getting a "multiple codenames" message on the subsequent page if I just reload the `/generate` page in the original tab (which the user may wish to do to generate a new codename). That's IMO a UX regression -- I suspect reloading the original tab is going to be more common than having more than one tab open. And it's confusing, because it warns the user about multiple tabs when they only have one open.

DrGFreeman:
Hi, thanks to all for the comments.

> I'm getting a "multiple codenames" message on the subsequent page if I just reload the `/generate` page in the original tab (which the user may wish to do to generate a new codename). That's IMO a UX regression.

@eloquence, I agree with you, this is not desirable. I should have thought of checking this in the first place.

I've been thinking about how to solve this issue elegantly for a while now and I think the fundamental issue preventing this is that we currently cannot uniquely identify each browser tab in which a codename is generated. As per [my previous comment](https://github.com/freedomofpress/securedrop/pull/4996#issuecomment-556854988), if we could do so, it would be possible to ensure the codename displayed in the tab from which the user clicks "SUBMIT DOCUMENTS" is the one used for the login and thus avoid having to use the flash message altogether. The only message required would be the "already logged in, please check your codename" message when the user clicks "SUBMIT DOCUMENTS" in the second tab. If someone experienced could comment on the feasibility of my proposal and potentially give some high level direction of how to make it work, I'd be glad to take the time and try to implement it.

> I'd like to give @ninavizz a chance to help with the final UI messaging at minimum. She suggested in the other PR that we might want to display the codename inline in the message;

Absolutely! That's why I set this PR to WIP.

I see three options with respect to displaying the codename to the user:
1. By making it visible at the bottom (what I have implemented here),
1. As a link in the flash message (as suggested at the bottom of my first comment on this PR),
1. Directly in the flash message.

I think that 1 and 2 have the advantage of allowing the source to hide back the codename while proceeding with their submission which cannot be done if it is included in the flash message.

securedrop/source_app/main.py
DrGFreeman: Good catch! I've added the period via 29f52d153194dceedd084dd95210c02ca276f5d2.

zenmonkeykstop:
Another option might be to move the regular "remember, your codename is.. " box up to the top, so that it's always front and center for new logins. Then if folks are getting the multiple tabs flash, the thing they need to check is just below it.

DrGFreeman:
Closing in favor of #5075 which resolves the [refresh issue]( https://github.com/freedomofpress/securedrop/pull/5048#issuecomment-562218281) identified by @eloquence above and in my opinion offers a much smoother user experience (source registered with the codename displayed in the tab from which they proceed to submit documents and only one flash message when the source is already logged-in).

PR #5083, opened 2019-12-19
Set Ansible connection timeout to 120s and updated playbooks to use Ansible's reboot module

### LINKED ISSUE #4358: Ansible commands sometimes fail with "waiting for privilege escalation prompt"

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #4358. (Well, mitigates it.)

This PR:
- updates the installer playbook, vagrant-packager, and upgrade Molecule scenarios to use the Ansible reboot module in most cases, or a simple `shutdown -r now` when there won't be a response from the rebooting servers.
- increases the Ansible connection timeout from 60s to 120s

### Installer:

On prod VMs or hardware:
- [ ] Complete an installation with SSH-over-Tor set and verify that it finishes without errors
- [ ] Update the config to use SSH-over-LAN, rerun the install (including the tailsconfig step when prompted) and verify that it finishes without errors
- [ ] Update random configuration fields via `./securedrop-admin sdconfig`, rerun `./securedrop-admin install` and verify that it completes without errors *or* if an SSH timeout is observed, that the timeout mentioned in the error message is 120sec approximately.

### Upgrade scenario
- [ ] On this branch, run through the "upgrade with local packages" scenario documented below, and verify that it completes successfully: 
https://docs.securedrop.org/en/release-1.2.0/development/upgrade_testing.html

### Vagrant packager
- [ ] on this branch, run `make vagrant-package` and verify that it completes successfully.

### DIFF
--- a/install_files/ansible-base/ansible.cfg
+++ b/install_files/ansible-base/ansible.cfg
@@ -2,12 +2,12 @@
 #ask_pass=False
 #ask_sudo_pass=False
 display_skipped_hosts=False
-timeout=60
+timeout=120
 # Use dynamic inventory script to determine local IPv4 addresses
 # from site-specific vars, or Onion URLs for SSH over ATHS.
 inventory=inventory-dynamic
 
 [ssh_connection]
 scp_if_ssh=True
-ssh_args = -o ControlMaster=auto -o ControlPersist=600s -o ConnectTimeout=60
+ssh_args = -o ControlMaster=auto -o ControlPersist=1200
 pipelining=True
--- a/devops/gce-nested/gce-runner.sh
+++ b/devops/gce-nested/gce-runner.sh
@@ -57,3 +57,5 @@ ssh_gce "make build-debs-notest"
 # so register a trap to ensure the fetch always runs.
 trap fetch_junit_test_results EXIT
 ssh_gce "make staging"
+# pause to allow for post-staging reboot
+sleep 30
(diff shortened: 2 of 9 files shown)


install_files/ansible-base/ansible.cfg
```diff
@@ -2,12 +2,12 @@
...
-timeout=60
+timeout=120
 # Use dynamic inventory script to determine local IPv4 addresses
 # from site-specific vars, or Onion URLs for SSH over ATHS.
 inventory=inventory-dynamic
 
 [ssh_connection]
 scp_if_ssh=True
-ssh_args = -o ControlMaster=auto -o ControlPersist=600s -o ConnectTimeout=60
+ssh_args = -o ControlMaster=auto -o ControlPersist=1200
```
rmol: This still seems backward to me -- I thought we'd seen problems due to lingering connections after reboot that required manual cleanup, which I'd think a shorter ControlPersist value might help -- but I can't argue with the results.

Issue #5084, opened 2019-12-28
Evaluate removal of apparmor rule for the shell execution of /bin/dash


evilaliv3:
Following the usage of the recent python-gnupg version (both in its original version and the patched version) that is using  python popen with ```shell=False```, it should be probably possible to remove ```/bin/dash``` from the requirements included in the apparmor profile.

In fact the recent versions of python-gnupg do not require anymore to launch a shell but use pipes.

This could probably significantly lower the possible attack surface previously required by the usage of python-gnupg.

evilaliv3:
@redshiftzero @conorsch: Considering the simplicity of the patch i considered opening a pull request so that we could see directly the impact of the change on the CI change.

We just applied this correction on GlobaLeaks and considering the value of the patch in an effective explot situation maybe could be valuable to apply this in Securedrop as well.

I've not currently found any other documentation about the need for this rule, do you remember any?

Issue #1389, opened 2016-08-23
Consider using dedicated WSGI app to run Python apps instead of Apache2


heartsucker:
All the web (Tor) facing services that comprise SecureDrop should be run by a separate process than the webserver. This would allow us to use AppArmor to completely separate out the responsibilities and permissions of Tor, the webserver, and the application itself. Breaching Tor won't get you to the app files (as is already true), but further, breaching Apache2 won't get you to access to anything except the config files.

Effectively, the webserver would just be a proxy that sets headers, and at this point we could also evaluate which server (Apach2, Nginx, Lighthttpd) is the most secure and has the best performance.

Gunincorn would be the WSGI app of choice, and it could be wrapped in an Upstart / Systemd script easily. Gunicorn can also bind to Unix sockets which would allow to use AppArmor to specify that only the webserver and SecureDrop processes could talk to the socket. If I remember right, Apache doesn't play nice with sockets which would necessitate a change of webserver.

Related: #1261

heartsucker:
Just kidding, more reading says that support for Unix sockets was added in Apache 2.4.7. [Link](https://httpd.apache.org/docs/trunk/mod/mod_proxy.html#proxypass).

However, the rest of this still stands.

fowlslegs:
We've talked briefly about this internally. It should be noted that Apache still buffers requests and replies (including plaintext submissions), and that OSSEC monitors that app files are not modified. Even if you just compromise Apache, you should still be able to spy on and modify all requests and responses.

I believe it is standard that TLS (which only The Intercept uses at the moment, but hopefully in the future will be better supported for `.onion`s) is handled by the web server, however, from some brief research it seems like you may be able to handle that from Gunicorn and use nginx (which Gunicorn recommends) as a proxy server. Don't quote me on this, but it _may_ be the case that if you do this nginx would only be able to observe TLS traffic.

I think there is still a lot more consideration that needs to go into this, but to me this seems like this idea represents a minor security improvement (if you can't ensure the web server only sees TLS traffic), and a moderate improvement (if you can ensure the web server only sees TLS traffic).

I also think there are also a number of unrelated reasons to prefer nginx over Apache. Another server I find interesting is caddy, but I don't know if it's been tested and audited enough to be a viable candidate for the SecureDrop platform.

zenmonkeykstop:
Relevant to #5084 - moving to Gunicorn or similar should allow for the apparmor rules there to be applied.

PR #5085, opened 2019-12-28
Remove apparmor rule for the shell execution of /bin/dash (#5084)

evilaliv3:
Implement correction as for ticket #5084

### DIFF
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
@@ -10,7 +10,6 @@
   capability net_bind_service,
   capability sys_ptrace,
 
-  /bin/dash rix,
   /bin/touch rix,
   /bin/uname rix,
   /dev/null w,


zenmonkeykstop:
Happy 2020 @evilaliv3 and thanks for this! It looks like `shell=True` is only going to be set for Windows systems in our current python-gnupg version, so this should be a change with no functional effect. Do you consider it ready for review? And if so, do you think there are any specific test cases we should be looking at?

evilaliv3:
Thank you @zenmonkeykstop; best wishes to you as well!

In my opinion yes the pull request is ready for merge and the restriction it applies is important as it fixes basically the main important failure in the design of python gnupg but unfortunately i don't think we have any option for a unit test here. I'm confident you should be enabling dash just in relation to previous versions of python gnupg that were requiring that in relation to popen but others of your team may confirm as well if this is the case.

zenmonkeykstop:
Thanks @evilaliv3 - I'll run through the usual functional tests on a production system (for that full apparmor experience :) ) and see how it goes. I'd agree with your assessment and it's unlikely we'll be downgrading the library to one that requires shell=True anytime soon.

zenmonkeykstop:
FTR - rough test plan is:

- [ ] the installation completes successfully on a prod system (HW or VM), with SI and JI available
- [ ] apparmor is enabled on the servers
- [ ]  source reply keys are generated successfully when a source account is created
- [ ]  submissions are encrypted successfully using the submission key
- [ ]  replies are encrypted successfully using the appropriate source reply key
- [ ]  replies are decrypted successfully in the SI
- [ ]  there are no apparmor errors in system logs

extra credit:
- [ ] basic server tests and application acceptance tests from 1.2.0 test plan pass.

zenmonkeykstop:
Just as a heads-up, I'm getting 2 `/bin/dash` apparmor denies on the initial request to the source or journalist interfaces, so something else relies on it. My guess is mod_wsgi but I'm still looking.

zenmonkeykstop:
dash is run when starting the WSGI daemon process, so unfortunately I don't think this apparmor change is applicable to SecureDrop, unless mod_wsgi can be reconfigured.

evilaliv3:
Ah! bummer! thank you for the update @zenmonkeykstop 

Actually what we could do probably is to sandbox the app in a different way so that each component stays in its sandbox.

E.g. giving /bin/sh to wsagi we are giving the entire codebase of securedrop this capability but maybe we could grant this to wsagi only and sandbox all the rest in a more precise manner.

evilaliv3:
Ah! bummer! thank you for the update @zenmonkeykstop 

Actually what we could do probably is to sandbox the app in a different way so that each component stays in its sandbox.

E.g. giving /bin/sh to wsagi we are giving the entire codebase of securedrop this capability but maybe we could grant this to wsagi only and sandbox all the rest in a more precise manner.

zenmonkeykstop:
No problem @evilaliv3 - #1389 is probably relevant here. I'm going to close this PR but it would be great to have your thoughts on the gunicorn+nginx architecture change mentioned in that ticket...

Issue #1584, opened 2017-02-15
Get rid of entropy checks


garrettr:
In `source.py`, we [check](https://github.com/freedomofpress/securedrop/blob/46a948ebadfb9b14bb7e902eff9f2be3f0e91fac/securedrop/source.py#L322-L326) for "available entropy" before generating a key. This is done because GPG blocks on "sufficient entropy" from `/dev/random` to generate a GPG key, and we don't want to block returning the HTTP response.

After discussing this with some more-knowledgeable security folks, a couple of things have come to my attention:

* libgcrypt has apparently improved their DRBG code since I last investigated this issue. It now appears to allow the use of `urandom` (although this may require some configuration, and it's not clear to me how/when this is done), will use `getrandom` if it's available, etc.
* libgcrypt has relaxed the formerly stringent checks to use `/dev/random` _and only_ `/dev/random`. So if we can't configure libgcrypt to use a non-blocking CSPRNG, we may now be able to trick it by symlinking `urandom`, or using `mknod`, etc.

Making this change could have a variety of potential benefits:

* Reduce code complexity and potential for bugs from `async_genkey` and the test helpers that wait for key generation.
* Eliminate "intermittent test errors" due to `async_genkey` (#844)
* Improve performance of tests that generate crypto keys

However, we may need to use a more recent version of GPG/libgcrypt to take advantage of these recent improvements. This might require upgrading our entire stack from Ubuntu 14.04LTS - a change that we want to make anyway, but one that is non-trivial.

dachary:
For the record https://www.2uo.de/myths-about-urandom/

dachary:
https://github.com/freedomofpress/securedrop/pull/2583

redshiftzero:
According to [this upstream ticket](https://dev.gnupg.org/T3894) libgcrypt 1.8 and above will use `getrandom()` if it's available, but unfortunately @emkll found that [xenial is using libgcrypt 1.6](https://packages.ubuntu.com/search?keywords=libgcrypt&searchon=names&suite=xenial&section=all).

However another option would be to use the `only-urandom` gcrypt config option. From [the docs](https://gnupg.org/documentation/manuals/gcrypt/Configuration.html) this option will:

> Always use the non-blocking /dev/urandom or the respective system call instead of the blocking /dev/random. If Libgcrypt is used early in the boot process of the system, this option should only be used if the system also supports the getrandom system call.

[...]

PR #5100, opened 2020-01-22
Cache CryptoUtil.getkey (redshiftzero's idea)

rmol:
## Status

Work in process

## Description of Changes

Adds caching to CryptoUtil.getkey, to reduce the number of expensive GPG key lookup operations. It uses CryptoUtil.keycache, an OrderedDict, so we can push out old items once we reach the cache size limit. Using functools.lru_cache would have taken care of that, but meant we couldn't avoid caching sources without keys, so delays in key generation would mean the source key would be unusable until the server were restarted.

The cache is primed in securedrop/journalist.py to avoid cold starts.

Credit to @redshiftzero for the solution.

## Testing

- Check out this branch. 

Save this script as `securedrop/securedrop/getkeytest.py`:
```
#!/usr/bin/env python3

import time

import pyotp
import requests


def api_url(path):
    return "http://localhost:8081/api/v1{}".format(path)


def get_all_sources(headers):
    start = time.perf_counter()
    get_all_sources_response = requests.get(api_url("/sources"), headers=headers)
    elapsed = time.perf_counter() - start
    sources = get_all_sources_response.json()["sources"]
    print("get_all_sources with {:d} sources took {:.2f} seconds".format(len(sources), elapsed))


if __name__ == "__main__":
    token_data = {
        "username": "journalist",
        "passphrase": "correct horse battery staple profanity oil chewy",
        "one_time_code": pyotp.TOTP("JHCOGO7VCER3EJ4L").now(),
    }
    token_response = requests.post(api_url("/token"), json=token_data).json()
    headers = {
        "Authorization": "Token {}".format(token_response["token"])
    }

    get_all_sources(headers)
    get_all_sources(headers)
```

- `export NUM_SOURCES=50`
- `make dev`

The dev server will take considerably longer to start than usual, as it creates the extra sources.

In another shell, once the server has finished populating the database and is fully ready:

- `docker exec -it securedrop-dev bash`
- `/opt/venvs/securedrop-app-code/bin/python3 getkeytest.py`  # this is run in the container

You should see output like this:
```
get_all_sources with 50 sources took 0.47 seconds
get_all_sources with 50 sources took 0.47 seconds
```

Now edit `securedrop/journalist.py` to comment out the call to `prime_keycache()` on line 22, and back in the container shell, run `getkeytest.py` again. If you get a key error about `token`, you were too quick at editing and your login attempt has been throttled. Wait 10-15 seconds and try again.

You should see output like this:
```
get_all_sources with 50 sources took 1.59 seconds
get_all_sources with 50 sources took 0.50 seconds
```
Note that this time the first call to `get_all_sources` is slower, as the cache hasn't been primed. 

## Deployment

The cache will increase memory consumption of both the source and journalist interfaces, but it's limited to 1000 keys, so should not be a problem.

### DIFF
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -55,6 +55,32 @@ def monkey_patch_delete_handle_status(self, key, value):
 gnupg._parsers.DeleteResult._handle_status = monkey_patch_delete_handle_status
 
 
+class FIFOCache():
+    """
+    We implemented this simple cache instead of using functools.lru_cache
+    (this uses a different cache replacement policy (FIFO), but either
+    FIFO or LRU works for our key fingerprint cache)
+    due to the inability to remove an item from its cache.
+
+    See: https://bugs.python.org/issue28178
+    """
+    def __init__(self, maxsize: int):
+        self.maxsize = maxsize
+        self.cache = collections.OrderedDict()  # type: collections.OrderedDict
+
+    def get(self, item):
+        if item in self.cache:
+            return self.cache[item]
+
+    def put(self, item, value):
+        self.cache[item] = value
+        if len(self.cache) > self.maxsize:
+            self.cache.popitem(last=False)
+
+    def delete(self, item):
+        del self.cache[item]
+
+
 class CryptoException(Exception):
     pass
 
@@ -73,8 +99,8 @@ class CryptoUtil:
     # to set an expiration date.
     DEFAULT_KEY_EXPIRATION_DATE = '0'
 
-    keycache = collections.OrderedDict()  # type: collections.OrderedDict
     keycache_limit = 1000
+    keycache = FIFOCache(keycache_limit)
 
     def __init__(self,
                  scrypt_params,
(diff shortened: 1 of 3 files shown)


securedrop/crypto_util.py
```diff
@@ -225,6 +226,7 @@ def delete_reply_keypair(self, source_filesystem_id):
         # The subkeys keyword argument deletes both secret and public keys.
         temp_gpg.delete_keys(key, secret=True, subkeys=True)
 
+    @lru_cache(maxsize=500)
```
redshiftzero: so i know you're working on the other bits of this (cache warmup) so will hold off on review

i'm wondering: what do you think is a reasonable number of sources here to cache? this `lru_cache` is just storing a dict so maybe we could make this much larger, like 2048 (the [docs say](https://docs.python.org/3/library/functools.html#functools.lru_cache) performance is best when we use powers of 2)

rmol: I went with 1000. I think that would be under 200kb, based on my crude estimates putting `filesystem_id` and `fingerprint` sized strings into `collections.OrderedDict`. 

I don't think the size needs to be a power of 2 with OrderedDict.

redshiftzero: afaik the LRU cache internally is using both a regular dict and a circular doubly linked list, I believe the latter is the reason for the powers of 2 suggestion in the docs

rmol: Right, the latest iteration is no longer using `functools.lru_cache`.

redshiftzero: hmm why do we not want to use the upstream decorator?

rmol: Doh, sorry, didn't update the PR description from the new commit message, which tried to explain that. Communication!

redshiftzero:
this all looks great! way faster. one thought: check out d04cdd4ea5ea0d2b1e5ee0cec0b6a688e20e94a0 which just encapsulates the cache so we can add a test (mostly for documentation purposes so it's super clear to future maintainers what's going on). if you like that you can cherry pick onto this branch or i can push directly

rmol:
Yeah, much nicer.

emkll:
LGTM based on visual review: `CryptoUtil.get_key` is only exposed once to the Source Interface, in the "flag for reply" flow [1]:  . This logic should rarely (if ever) be hit, and will be removed soon, per [2].Therefore, the caching introduced here should only effect the Journalist Interface, and not have any effect on the distinguishability of sources (re)visiting the source interface through timing.

[1] : https://github.com/freedomofpress/securedrop/blob/speedy-getkey/securedrop/source_app/main.py#L133
[2] :  https://github.com/freedomofpress/securedrop/issues/1584#issuecomment-571380760

PR #5175, opened 2020-03-30
Updates references to "journalists" in Source Interface to more neutral "team"

### LINKED ISSUE #5146: support for overriding interface strings
## Description

Our use of SecureDrop makes us an edge case:  We are a nonprofit law
firm, not a newsroom.  We would like our SecureDrop instance's source
interface to reflect that our "sources" are prospective clients and
our "journalists" are lawyers.

We have put together a rough [prototype][local-words] of how to
accomplish this kind of string-substitution while still playing
nicely with SecureDrop's translation via `gettext`.  Before we
develop this prototype further---in particular, to hook it into the
`instance_config` table via the admin interface---for submission as a
pull request, we would like to invite the SecureDrop team's feedback
on our goal and approach.

1. Is this kind of translation-respecting string-substitution a
   feature you would consider including in SecureDrop?  (We are aware
   of #1575 as a rough precedent without an easy resolution.)

2. Is this approach worth developing into a full-fledged pull
   request? or would you suggest another approach?  (We have considered
   proposing administrator-configurable message catalogs, but this
   seems like a very large hammer for the very small nail we are
   trying to hit.)

## User Research Evidence

We are one example of an organization whose security model benefits
from use of SecureDrop but to whom the labels "source" and "journalist"
do not apply.

## User Stories

As the administrator of a non-newsroom SecureDrop instance, I would
like to be able to change the words "source" and "journalist", as
they appear on the source interface, to reflect my organization's
use-case and roles.

[local-words]: https://github.com/freedomofpress/securedrop/compare/develop...wbaid:local-words

wbaid:
## Status

Ready for review

## Description of Changes

Closes #5146.

- Replaces instances of the phrase "our journalist[s]" with the phrase "our team", to reduce possible confusion when a source encounters a non-newsroom SecureDrop instance.

## Testing

Observe the changed strings in the source interface's templates.

## Deployment

No special considerations.

### DIFF
--- a/securedrop/source_templates/generate.html
+++ b/securedrop/source_templates/generate.html
@@ -7,7 +7,7 @@
   <p>{{ gettext('Please either write this codename down and keep it in a safe place, or memorize it.') }}</p>
 </aside>
 
-<p class="explanation">{{ gettext('This codename is what you will use in future visits to receive messages from our journalists in response to what you submit on the next screen.') }}</p>
+<p class="explanation">{{ gettext('This codename is what you will use in future visits to receive messages from our team in response to what you submit on the next screen.') }}</p>
 
 <hr class="no-line">
 
--- a/securedrop/source_templates/why-journalist-key.html
+++ b/securedrop/source_templates/why-journalist-key.html
@@ -1,6 +1,6 @@
 {% extends "base.html" %}
 {% block body %}
-<h1>{{ gettext("Why download the journalist's public key?") }}</h1>
+<h1>{{ gettext("Why download the team's public key?") }}</h1>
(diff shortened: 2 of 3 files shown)


zenmonkeykstop:
From a technical perspective there are no issues with this change. One concern might be that the move to more neutral language might make the purpose of the codename and gpg key less clear to sources. @ninavizz, any thoughts on UX impact of this change?

ninavizz:
I generally dislike use of the word "staff," however it could carry stronger implications of safety via implied NDAs and commitment to ethics, than the more broad "team." However, "team" is likely more accurate—as some instances may well be run by volunteers, with all-volunteer orgs being high on my list for probable candidates.

TL;DR, yes, this is a great change. Thx for the ping @zenmonkeykstop ! :)

PR #5178, opened 2020-04-01
Fixes #5176 Adds code+test to return replies without a journalist

### LINKED ISSUE #5176: deleted journalist causes 500 on GET /replies API endpoint
## Description

Deleted journalist causes GET /replies endpoint to 500 if there are any replies from that deleted journalist. 

## Steps to Reproduce

0. Submit a document or message as a source.
1. Reply to the source from journalist account A.
2. Delete the account of journalist account A (from another admin account). 
3. `GET /api/v1/replies`

## Expected Behavior

200 OK 

## Actual Behavior

```
172.17.0.1 - - [31/Mar/2020 20:57:10] "GET /api/v1/replies HTTP/1.1" 500 -
  File "/opt/venvs/securedrop-app-code/lib/python3.5/site-packages/flask/app.py", line 2309, in __call__
    return self.wsgi_app(environ, start_response)
  File "/opt/venvs/securedrop-app-code/lib/python3.5/site-packages/flask/app.py", line 2295, in wsgi_app
    response = self.handle_exception(e)
  File "/opt/venvs/securedrop-app-code/lib/python3.5/site-packages/flask/app.py", line 1741, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/opt/venvs/securedrop-app-code/lib/python3.5/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/opt/venvs/securedrop-app-code/lib/python3.5/site-packages/flask/app.py", line 2292, in wsgi_app
  File "/opt/venvs/securedrop-app-code/lib/python3.5/site-packages/flask/app.py", line 1815, in full_dispatch_request
  File "/opt/venvs/securedrop-app-code/lib/python3.5/site-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/opt/venvs/securedrop-app-code/lib/python3.5/site-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/opt/venvs/securedrop-app-code/lib/python3.5/site-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/opt/venvs/securedrop-app-code/lib/python3.5/site-packages/flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/redshiftzero/Documents/Github/securedrop-1.2.0-rc2/securedrop/securedrop/journalist_app/api.py", line 48, in decorated_function
    return f(*args, **kwargs)
  File "/Users/redshiftzero/Documents/Github/securedrop-1.2.0-rc2/securedrop/securedrop/journalist_app/api.py", line 310, in get_all_replies
    {'replies': [reply.to_json() for reply in replies]}), 200
  File "/Users/redshiftzero/Documents/Github/securedrop-1.2.0-rc2/securedrop/securedrop/journalist_app/api.py", line 310, in <listcomp>
    {'replies': [reply.to_json() for reply in replies]}), 200
  File "/Users/redshiftzero/Documents/Github/securedrop-1.2.0-rc2/securedrop/securedrop/models.py", line 289, in to_json
    'journalist_username': self.journalist.username,
AttributeError: 'NoneType' object has no attribute 'username'
```

## Comments

We should handle where the `journalist` is `None` [here](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/models.py#L289-L292)

It would be wise to also add test data to `create-dev-data.py` for the deleted journalist scenario (since that is used for development of securedrop-client).

### LINKED ISSUE #5177: Populate developer environment with an orphaned journalist ID
To better catch issues like #5176, it would be useful for the standard `make dev` developer environment to be populated with at least one reply associated with a journalist account that no longer exists.

kushaldas:
## Status

Ready for review.

## Description of Changes

Fixes #5176
Fixes #5177

We can safely return a reply by a journalist who is deleted from
the system. This patch also includes a test case, and dev-data
generation script also now adds a reply and then deletes the
journalist.

## Testing

- [ ] `create-dev-data.py` will add the first source with a journalist and then deletes the journalist
- [ ] https://github.com/freedomofpress/securedrop-sdk/blob/master/sdclientapi/__init__.py#L675 the list of replies for the first source can be fetched without any error

## Deployment

2. New installs.

Choose one of the following:

### DIFF
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -278,10 +278,10 @@ class Reply(db.Model):
 
     def to_json(self):
         # type: () -> Dict[str, Union[str, int, bool]]
-        username = ""
+        username = "deleted"
         first_name = ""
         last_name = ""
-        uuid = ""
+        uuid = "deleted"
         if self.journalist:
             username = self.journalist.username
             first_name = self.journalist.first_name
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -278,6 +278,15 @@ class Reply(db.Model):
 
     def to_json(self):
         # type: () -> Dict[str, Union[str, int, bool]]
+        username = ""
+        first_name = ""
+        last_name = ""
+        uuid = ""
+        if self.journalist:
+            username = self.journalist.username
+            first_name = self.journalist.first_name
+            last_name = self.journalist.last_name
+            uuid = self.journalist.uuid
         json_submission = {
             'source_url': url_for('api.single_source',
                                   source_uuid=self.source.uuid),
@@ -286,10 +295,10 @@ class Reply(db.Model):
                                  reply_uuid=self.uuid),
             'filename': self.filename,
             'size': self.size,
-            'journalist_username': self.journalist.username,
-            'journalist_first_name': self.journalist.first_name,
-            'journalist_last_name': self.journalist.last_name,
-            'journalist_uuid': self.journalist.uuid,
+            'journalist_username': username,
+            'journalist_first_name': first_name,
+            'journalist_last_name': last_name,
+            'journalist_uuid': uuid,
             'uuid': self.uuid,
             'is_deleted_by_source': self.deleted_by_source,
         }
--- a/securedrop/create-dev-data.py
+++ b/securedrop/create-dev-data.py
@@ -49,29 +49,48 @@ def main(staging=False):
                       test_otp_secret,
                       is_admin=False)
 
+        journalist_tobe_deleted = add_test_user("clarkkent",
+                                                test_password,
+                                                test_otp_secret,
+                                                is_admin=False,
+                                                first_name="Clark",
+                                                last_name="Kent")
+
         # Add test sources and submissions
         num_sources = int(os.getenv('NUM_SOURCES', 2))
-        for _ in range(num_sources):
+        for i in range(num_sources):
+            if i == 0:
+                # For the first source, the journalist who replied will be deleted
+                create_source_and_submissions(journalist_who_replied=journalist_tobe_deleted)
+                continue
             create_source_and_submissions()
+        # Now let us delete one journalist
+        db.session.delete(journalist_tobe_deleted)
+        db.session.commit()
 
 
-def add_test_user(username, password, otp_secret, is_admin=False):
+def add_test_user(username, password, otp_secret, is_admin=False,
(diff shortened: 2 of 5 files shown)


redshiftzero:
Looks good! CI is failing due to #5180 so we can resolve that first and then rebase this change

redshiftzero:
I think one change that does make sense is setting both the default username and uuid for journalist replies when the journalist does not exist to `deleted` from the API (and leaving first/last names to their empty values, which is what we do elsewhere in the API for absent first/last name fields). Thus:

* when we sync the client, we'll have a single row for deleted journalists with UUID `deleted` (what we use as a unique identifier for journalists). Then when we log (generally done with the UUID except for login), we'll see clearly in the logs that this is a deleted journalist. 
* if there is an existing journalist that just happens to have username `deleted`, they will be distinct in the `users` table in the client-side database as they will have different UUIDs. No unique constraint will be hit server-side as only one entry will exist with username `deleted` in the server database. 

I've tested this locally, so I'll push a commit with this change. If others disagree with this approach please comment.

eloquence:
Seems reasonable to me, I have a soft preference for using strings like `_deleted_` that are more unambiguously searchable and traceable to our code, but of course defer to y'all.

redshiftzero:
I'm approving @kushaldas' changes but would like someone to review the commit I added here. I've tested by applying this change to the dev env and syncing with the client, all works as expected (I went with `deleted` just so we can use as is in the UI without needing to strip off the `_`). There is an improvement we can make to preserve more conversation history as sources are deleted in the client (instead of associating replies with the `deleted` row when users are deleted), I'll file that over there, but the latest release of the client should work with this server-side change for the deleted journalist case.

securedrop/models.py
```diff
@@ -278,6 +278,15 @@ def __repr__(self):
 
     def to_json(self):
         # type: () -> Dict[str, Union[str, int, bool]]
+        username = "deleted"
+        first_name = ""
+        last_name = ""
+        uuid = "deleted"
```
emkll, line 284: :+1:  setting the uuid is a great idea

redshiftzero:
Thanks for review! I'll merge this now and create a patch for applying to our shared test instance.

PR #5199, opened 2020-04-17
Update ansible to 2.9.7

### LINKED ISSUE #5144: update Ansible due to CVE-2019-14864
## Description

[CVE-2019-14864](https://nvd.nist.gov/vuln/detail/CVE-2019-14864) is a vulnerability in Ansible's `no_log` flag for the splunk and sumologic plugins (sensitive data is incorrectly logged) but neither of which we're using. Regardless, we should update Ansible to a version that does not have this vulnerability in the next release.

@emkll also pointed out to me that this is a [good time](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html) to get onto the 2.8 series since the 2.7 series will become unmaintained when 2.10 is released (it's in development).

rmol:
## Status

Ready for review

## Description of Changes

Updates Ansible.

Fixes #5144.

## Testing

Ideally, `make clean` followed by `make venv && make build-debs && make staging`, then `molecule verify -s libvirt-staging-xenial` (or whatever your staging scenario is).

Even more ideally, run through a production install on VMs or hardware, using this branch.

## Deployment

It will affect admin workstations and development environments, but should not affect the content of deployed instances.

Choose one of the following:

I've done a timeboxed and therefore limited review of the differences between Ansible 2.7.13 and 2.9.7 ([hashes of the tarballs I examined are in the packaging wiki](https://github.com/freedomofpress/securedrop-debian-packaging/wiki/Ansible-2.7.13-to-2.9.7)) and did not spot any problems in changes to the core functionality. I also ran Bandit checks for high-severity problems on both versions, and found no new problems reported in 2.9.7.


rmol:
Agreed on all points. I'm addressing 1 and 2 right now, and think we might as well solve 3 by setting `ansible_python_interpreter` in `ansible.cfg`. @conorsch what do you think?

conorsch:
> we might as well solve 3 by setting ansible_python_interpreter in ansible.cfg

Sounds great, the commit you've added doing that looks like just what we need. Given that @emkll has already verified py3 support in prod env, will take a closer look at the upgrade & qubes-staging scenarios, make sure we're covered there.

emkll:
>Given that @emkll has already verified py3 support in prod env, will take a closer look at the upgrade & qubes-staging scenarios, make sure we're covered there.

To be clear, verified that python3 was installed on a fresh Ubuntu 16.04.6 target as follows
- In libvirt VMs, installed Ubuntu 16.04.06 from the ISO, *not* from vagrant boxes). Selected standard system utilities and openssh server
- /usr/bin/python3, /usr/bin/python3.5 is installed
- Ran `./securedrop-admin install` on this branch, against said VMs, and observed a successful install on those one-off VMs.

rmol:
OK, I'm still having trouble with Molecule. When I try `make upgrade-start` I get `ERROR! The requested handler 'restart tor' was not found in either the main handlers list nor in the listening handlers list`. 

I've tried adding `listen: restart tor` [here](https://github.com/freedomofpress/securedrop/blob/ab8d4a09f6b6a633398910ae4e008a5a316885c2/install_files/ansible-base/roles/tor-hidden-services/handlers/main.yml#L2) because ["Handlers using import* will not be triggered when notified by their name, as importing overwrites the handler’s named task with the imported task list."](https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse.html#tradeoffs-and-pitfalls-between-includes-and-imports) (and the classic way of specifying roles in a playbook, [as used here, if I'm understanding correctly](https://github.com/freedomofpress/securedrop/blob/ab8d4a09f6b6a633398910ae4e008a5a316885c2/molecule/upgrade/playbook.yml#L40) is equivalent to import). Same error.

I converted the [`roles:` block](https://github.com/freedomofpress/securedrop/blob/ab8d4a09f6b6a633398910ae4e008a5a316885c2/molecule/upgrade/playbook.yml#L40) to a `tasks:` block containing `import_role` or `include_role`. Same error.

I cannot get this playbook to actually fire the handler in Ansible 2.8.11 or 2.9.7.

Any ideas would be incredibly welcome.

conorsch:
Appended two commits. The first fixes the OSSEC package build by installing `python3-requests`, rather than `python-requests`, which was required due to interpreter changes. 

The second fixes the "upgrade" scenario. 

>  When I try `make upgrade-start` I get `ERROR! The requested handler 'restart tor' was not found in either the main handlers list nor in the listening handlers list.`

The problem was that the "upgrade" scenario uses the _old_ logic from the previous stable release of SecureDrop, cloning it into a separate subdir, which is then used to provision the servers. Since we jumped two minor versions of Ansible, we lost mutual compatibility between the two, so even though the handler task was happy as of 2.9, the combination of 1) a separate subdir with relative lookups for handlers and 2) the old-style "include" running under Ansible 2.9 was leading to the repeated failure to find the handler.

rmol:
@conorsch Works great; the upgrade scenario is now fine. I'm just working through the rest as a final check, but I think you've done it. Thanks!

rmol:
I've worked through the rest of the tests @emkll listed, and everything looks good. Merging.

PR #5223, opened 2020-05-04
Removed submission preference alert from top of config page

### LINKED ISSUE #5220: Instance configuration flash message is unstyled
## Description

In #5046 we added flash messages confirming configuration changes, with the `submission-preferences-success` category. They appear properly styled when presented via the `submission_preferences_saved_flash.html` template, but they're also presented at the top of the page without the logic that styles them as successes or errors, so they're shown as plain text.

## Steps to Reproduce

Start the dev server and navigate to the instance config, then toggle document uploads under **Submission Preferences**. You should see a page like this, with `Preferences saved.` at the top (plain) and bottom (styled):

image

## Expected Behavior

That the flash message would appear in one place, or at least be styled the same wherever it appears.

## Actual Behavior

The message at the top of the page is forlornly unadorned, while at the bottom it's miles of style.

## Comments

In [#5046](https://github.com/freedomofpress/securedrop/pull/5046#issuecomment-561438216) @eloquence pointed out that the message could easily be missed if it only appears at the bottom of the page, so we should probably just make sure that the same styling is applied to both occurrences.

zenmonkeykstop:
## Status

## Description of Changes

Fixes #5220

Adds document submission flash message to the list of messages *not* displayed at the top of pages, similarly to the handling for the logo form on the instance config page. (It was being displayed unstyled due to the message category not having a corresponding css style)

## Testing

- checkout this branch and run `make dev`
- log into the journalist interface using the test credentials, and navigate to the config section: `/admin/config`
- change the submissions preference at the bottom of the page
- [ ] the config page is re-rendered *without* a flash message at the top of the page, and *with* the "Preferences saved" message rendered correctly in the Submissions Preferences section.

### DIFF
--- a/securedrop/journalist_templates/flashed.html
+++ b/securedrop/journalist_templates/flashed.html
@@ -1,7 +1,7 @@
 {% with messages = get_flashed_messages(with_categories=true) %}
 {% if messages %}
   {% for category, message in messages %}
-    {% if category != "banner-warning" and category != "logo-success" and category != "logo-error"%}
+    {% if category != "banner-warning" and category != "logo-success" and category != "logo-error" and category != "submission-preferences-success" %}
       <div class="flash {{ category }}">
         {% if category == "notification" %}
         <img src="{{ url_for('static', filename='i/font-awesome/info-circle-black.png') }}" height="16" width="20">


rmol:
I thought we were going to make it _only_ appear at the top of the page?

zenmonkeykstop:
@rmol IMO there's a good case for some UX attention on the admin interface in general, including alert placement. But for now this is just a bug fix to restore the intent of the original PR.

PR #5224, opened 2020-05-04
Include various client related strings in test data

kushaldas:
## Status

Ready for review

## Description of Changes

Fixes #1080

Adds many different test strings to test `securedrop-client` text formatting.

## Testing

- [ ] `make dev` should have 2 sources
- [ ] `NUM_SOURCES=10 make dev` should create 10 sources
- [ ] `NUM_SOURCES=ALL make dev` shoudl create 43 sources
- [ ] Use the securedrop-client against the local server and look many strange looking texts :)

## Deployment

2. New installs.

Choose one of the following:


creviera:
> Do you want to add all the different test strings from the wiki?

Up to you. I think since this is not default, it would be great to add all the test strings from https://github.com/freedomofpress/securedrop-client/wiki/Message-Test-Data and also maybe a couple from https://github.com/minimaxir/big-list-of-naughty-strings/blob/master/blns.txt as you see fit.

kushaldas:
> > Do you want to add all the different test strings from the wiki?
> 
> Up to you. I think since this is not default, it would be great to add all the test strings from https://github.com/freedomofpress/securedrop-client/wiki/Message-Test-Data and also maybe a couple from https://github.com/minimaxir/big-list-of-naughty-strings/blob/master/blns.txt as you see fit.

@creviera  I updated with more text form` blns.txt`, creating at least 20 sources will be nicer to consume all strings properly.

creviera:
this lgtm - linter test fails though:

```
create-dev-data.py:91:85: E0001: invalid syntax (<unknown>, line 91) (syntax-error)
```

might want to move the special strings into a txt file?

kushaldas:
Our future test strings are so good that they are breaking Python linting tools  :snake::man_dancing: :snake: 

```
********** Module specialstrings
specialstrings.py:63:85: E0001: invalid syntax (<unknown>, line 63) (syntax-error)
************* Module create-dev-data
create-dev-data.py:18:0: E0001: Cannot import 'specialstrings' due to syntax error 'invalid syntax (<unknown>, line 63)' (syntax-error)
```

securedrop/create-dev-data.py
```diff
@@ -66,6 +67,10 @@ def main(staging=False):
                 )
                 continue
             create_source_and_submissions(i, num_sources)
+
+        client_test = int(os.getenv('CLIENT_TEST_DATA', 0))
+        for i in range(1, client_test):
+            create_source_and_submissions(i, client_test, client_test_data=True)
```
sssoleileraaa: so you're passing `client_test` in for `source_count`? could you instead pass `num_sources` and make `CLIENT_TEST_DATA` a boolean that says whether or not special strings should be used?

**UPDATE**
Let's talk more tomorrow about the design and how we imagine QA will use this feature as well as how we might use this for our screenshot tests. It'll also make it easier for me to review.

redshiftzero: Here's a version with Allie's suggested change above, which works nicely for me (feel free not to use this):

```diff
diff --git a/securedrop/create-dev-data.py b/securedrop/create-dev-data.py
index e689c6080..de5b8ea85 100755
--- a/securedrop/create-dev-data.py
+++ b/securedrop/create-dev-data.py
@@ -59,18 +59,17 @@ def main(staging=False):

         # Add test sources and submissions
         num_sources = int(os.getenv('NUM_SOURCES', 2))
+        client_test = bool(os.getenv('CLIENT_TEST_DATA', 0))
         for i in range(1, num_sources + 1):
             if i == 1:
                 # For the first source, the journalist who replied will be deleted
                 create_source_and_submissions(
-                    i, num_sources, journalist_who_replied=journalist_tobe_deleted
+                    i, num_sources, journalist_who_replied=journalist_tobe_deleted,
+                    client_test_data=client_test
                 )
                 continue
-            create_source_and_submissions(i, num_sources)
+            create_source_and_submissions(i, num_sources, client_test_data=client_test)

-        client_test = int(os.getenv('CLIENT_TEST_DATA', 0))
-        for i in range(1, client_test):
-            create_source_and_submissions(i, client_test, client_test_data=True)
         # Now let us delete one journalist
         db.session.delete(journalist_tobe_deleted)
         db.session.commit()
```

sssoleileraaa: further thoughts in here: https://github.com/freedomofpress/securedrop/pull/5224#pullrequestreview-428436671

rocodes:
@kushaldas Tried the changes in 2fca9dd and imo they look good and work well. I'm not going to weigh in on whether `make dev` should make automatically with this number of sources (adds maybe 10 seconds to make time?), so I'll leave others to weigh in on the right amount of test data to be the default, but I think this looks good. 

A few other points that came out of my conversation with @creviera: 
- There could be a useful integration between this and `qa_loader` down the line 
- The idea of reproducible data being populated during `make dev` paves the way nicely for more automated testing/screenshots in future too.

eloquence:
Personally I like the behavior as it is now; for new developers using the development environment, it doesn't add a huge performance penalty (creation of sources is pretty zippy; it takes 15 seconds to create 43 sources on my older X1), and for more experienced devs, it's easy enough to add `NUM_SOURCES`. I think in general there's some value in doing more testing with a more highly populated environment by default.

eloquence:
The one caveat is that all replies are identical right now. We've sometimes seen issues that arise only in messages, or only in replies, such as the recent styling differences reported in https://github.com/freedomofpress/securedrop-client/pull/1082#issuecomment-642255136, or the context menu bug in https://github.com/freedomofpress/securedrop-client/issues/1064 . I don't think this absolutely needs to be resolved in this PR, but it would be IMO preferable for messages and replies to both use the special strings file.

creviera:
Yeah, I agree that we can build on top of this change and that it doesn't need to do all the things. To be clear about what my review comment is requesting for changes:

> * Like messages, we should be able to populate replies with all the different test data

This is to catch bugs that only occur in the reply bubbles. Reply bubbles descend from the same class as message bubbles but they introduce new GUI behavior, like adding failure messages and changing the color bar, and so can we might miss a styling bug if we only show test data strings in message bubbles.

Also, this would allow us to see special strings in preview snippets, which would allow us to catch more bugs.

This can be made into a smaller issue that can be done outside of this PR.

> * The default configuration should be the most broadly useful during test. For me, this means it should be fast so that I can test as I develop (this is less important for QA testers but it doesn't hurt), as well as easily reproducible to faciliate creating STRs when bugs are found.

> So far your change meets this criteria but I think we can take it a step further and make it the default, i.e. running make dev should use all the test data strings. 

As Ro pointed out this would add a lot of time when starting `make dev`. The only problem I see with _not_ doing this is that we will have to know the magic number to pass to NUM_SOURCES in order to see all the test strings.

This doesn't need to hold up your PR, but it is confusing as is, because when you run `NUM_SOURCES=1 CLIENT_TEST_DATA=10 make dev` you get 10 sources with 2 different special strings each in 2 different messages.

eloquence:
> The only problem I see with not doing this is that we will have to know the magic number to pass to NUM_SOURCES in order to see all the test strings.

I really don't think that's desirable - testers should not have to count up the number of test cases or look for a magic number somewhere. If we don't want to default to importing all, IMO there should be a symbolic way to specify this, e.g., `NUM_SOURCES=ALL`.

creviera:
Personally, I still prefer:

`make dev` - sets up a test server with all the strings in the test data file
`NUM_SOURCES=1 make dev` - sets up a test server with 1 source that has the special strings from the top of the test data file list
`NUM_SOURCES=10 make dev` - sets up a test server with 10 sources that have the special strings from the top of the test data file list (might not cycle through all the strings)
`NUM_SOURCES=1000 make dev` - sets up a test server with 1000 sources that have the special strings from the test data file list (cycles through all the strings multiple times)

And in the _future..._ maybe something like:

`NUM_MESSAGES=200 NUM_SOURCES=10 make dev` - sets up a test server with 10 sources that each have 200 messages+replies that have special strings from the the test data file list (cycles through all the strings multiple times)

creviera:
Agreed, if we want to make `make dev` default to a smaller number of sources to speed things up, then `NUM_SOURCES=ALL make dev` is a nice solution.

securedrop/specialstrings.py
```diff
@@ -0,0 +1,90 @@
+from itertools import cycle
+
+
+submissions = cycle([
+    """This is a test submission without markup!""",
+    """This is a test submission with markup and characters such as \, \\, \', \" and ". """ +  # noqa: W605, E501
+    """<s$trong>This text should not be bold</strong>!""",
```
sssoleileraaa: `$` snuck in here

securedrop/create-dev-data.py
```diff
@@ -110,11 +111,12 @@ def create_source_and_submissions(
     # Generate some test submissions
     for _ in range(num_submissions):
         source.interaction_count += 1
+        submission_text = next(submissions)
```
sssoleileraaa, line 113: looks good! do you want to add special strings in the replies as part of this PR?

securedrop/specialstrings.py
```diff
@@ -0,0 +1,93 @@
...
...ouse and then regurgitate parts of it on the family room floor for why must they do that eats owners hair then claws head while happily ignoring when being called trip owner up in kitchen i want food. There's a forty year old lady there let us feast lick the curtain just to be annoying yet pet my belly, you know you want to; seize the hand and shred it! so leave hair on owner's clothes immediately regret falling into bathtub walk on car leaving trail of paw prints on hood and windshield. Cat dog hate mouse eat string barf pillow no baths hate everything. Check cat door for ambush 10 times before coming in eat all the power cords for fish i must find my red catnip fishy fish sleep everywhere, but not in my bed throw down all the stuff in the kitchen where is my slave? I'm getting hungry lick the curtain just to be annoying. Kitty power meow to be let out lie on your belly and purr when you are asleep shake treat bag. Cat fur is the new black pet me pet me don't pet me i vomit in the bed in the middle of the night making bread on the bathrobe stare at ceiling light but wake up wander around the house making large amounts of noise jump on top of your human's bed and fall asleep again.
```
rmol: Please find different copy for this.

kushaldas: @creviera ^^ can we just choose another random text from anywhere else with same length?

sssoleileraaa: hah! yes please find more sensible copy, @ninavizz might have a good recommendation here

ninavizz: This is dummy text I'd prepared a couple years ago, for client testing. What exactly are your concerns, @rmol?

https://github.com/freedomofpress/securedrop-ux/blob/master/dummydata.txt

rmol: I'd simply like sample text that isn't embarrassing.

sssoleileraaa: agreed, cat ipsum and probably all lorem-ipsum alternatives text generators could produce an embarrassing  sequence of words. we can use the regular lorem ipsum generator (all latin) for the test strings where we really need ridiculously long messages.

#### Update

I went ahead and updated the cat-ipsum generated text to use lorem-ipsum in both the wiki and in this pr since I'm to blame for such a poor choice in test data.

eloquence:
(Lint failure in CI looks legit)

eloquence:
@rmol want to take another spin?

rmol:
I am much happier with the OG lorem ipsum. _But in certain circumstances and owing to the claims of duty or the obligations of business it will frequently occur that pleasures have to be repudiated and annoyances accepted._ Such as a rebase. :smile:

securedrop/create-dev-data.py
```diff
@@ -56,8 +52,12 @@ def main(staging=False):
                                                 first_name="Clark",
                                                 last_name="Kent")
 
+        NUM_SOURCES = os.getenv('NUM_SOURCES', 2)
+        if NUM_SOURCES == "ALL":
+            # We want all strings
```
eloquence: Small request, please be a bit more explicit in the source code comment where the magic number 43 comes from. :) It's much larger than the actual number of strings, so we'll see every string repeatedly -- is that intentional?

eloquence:
Note that we're currently creating significantly more sources than there are special strings when `NUM_SOURCES=ALL`, so we'll see each special string repeatedly (~3 times, because each source pulls in 2 strings). It's not clear to me where the 43 comes from, and by what rule it's intended to be incremented if the special strings file is updated.
https://github.com/freedomofpress/securedrop/pull/5224#discussion_r447334860

sssoleileraaa:
I think this needs to be kicked twice due to the test timeout issue.

sssoleileraaa:
(referring to `Too long with no output (exceeded 35m0s): context deadline exceeded`)

eloquence:
Taking a quick look at 46a5d8783eb1c7495654c4a3f0a2c5519c3bf10f ...

securedrop/create-dev-data.py
```diff
@@ -56,8 +49,13 @@ def main(staging=False):
                                                 first_name="Clark",
                                                 last_name="Kent")
 
+        NUM_SOURCES = os.getenv('NUM_SOURCES', 2)
+        if NUM_SOURCES == "ALL":
+            # We want all strings, 14 sources will give all the strings based
+            # on current string count.
+            NUM_SOURCES = 14
```
eloquence: Technically this is 1 source too many since there are 25 strings and 2 strings per source, so we consume 28 strings, meaning one fully repeated couple.

securedrop/create-dev-data.py
eloquence: (Allie did so in 46a5d8783eb1c7495654c4a3f0a2c5519c3bf10f)

securedrop/create-dev-data.py
eloquence: elegantly resolved by @creviera in c83d8c4ab3c41c8cdd1362f220b99bdc7489a770

eloquence:
LGTM.

Test plan:
- [x] `make dev` should have 2 sources
- [x] `NUM_SOURCES=10` make dev should create 10 sources
- :x: `NUM_SOURCES=ALL make dev` should create 43 sources  -- test plan is inaccurate
- [x] `NUM_SOURCES=ALL make dev` creates the required number of sources to ingest all strings

Recent changes by Allie LGTM. May benefit from a squash before merge.

eloquence:
(Squashed)

eloquence:
Erm, that LGTM alert actually looks legit, looks like we're concatenating strings that should be separate per https://github.com/freedomofpress/securedrop-client/wiki/Message-Test-Data

conorsch:
@eloquence Nice catch, I agree, we should fix that prior to merge. Thanks, LGTM: https://lgtm.com/projects/g/freedomofpress/securedrop/rev/pr-7c7409194c43f3b116ae35ee0fa18f076d1125dfh

PR #5236, opened 2020-05-06
Stop calling random.seed in qa_loader.py

### LINKED ISSUE #5235: qa_loader.py uses a fixed random seed every run
## Description

Always using the same seed makes it impossible to run `qa_loader.py` multiple times with the same database, as supposedly random values aren't, causing unique constraint violations.

## Steps to Reproduce

- Run the dev server with `make dev`
- Start a shell in the container with `docker exec -it securedrop-dev-0 bash`
- In that shell, run `./qa_loader --journalist-count 1 --source-count 1`
- Run the same command a second time.

## Expected Behavior

That you could keep adding random journalists and sources to the database.

## Actual Behavior

You get `sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) UNIQUE constraint failed: journalists.username` because [`random.seed` is always called with the same value](https://github.com/freedomofpress/securedrop/blob/ec2220c3c2b9120d029b616d3a07647b175bc6ab/securedrop/qa_loader.py#L22).

rmol:
## Status

Ready for review

## Description of Changes

Fixes #5235.

## Testing

Check out this branch. Follow the #5235 steps to reproduce. The second `qa_loader.py` run should succeed.

## Deployment

Dev-only.

### DIFF
--- a/securedrop/qa_loader.py
+++ b/securedrop/qa_loader.py
@@ -19,9 +19,6 @@ from models import Journalist, JournalistLoginAttempt, Reply, Source, SourceStar
 from sdconfig import config as sdconfig
 
 
-random.seed("~(=^–^)")  # mrow?
-
-
 def random_bool():
     return bool(random.getrandbits(1))
 


eloquence:
Hit this issue today, thanks much for preparing a fix for it. :)

PR #5241, opened 2020-05-08
updated release management guide with new instructions for docs build

### LINKED ISSUE #5081: Show warning header on previous versions of documentation
## Description

It might be useful to show a warning at the top of the documentation when viewing versions older than the current release. It's possible to accidentally end up on older versions though searching or old links. It looks like this is a setting in readthedocs already: https://docs.readthedocs.io/en/stable/versions.html#version-warning

## User Research Evidence

I somehow ended up on the 1.0.0 documentation, I think either through a link from somewhere else or by my having attempted to visit it directly and my browser auto-completing the url of the documentation from when I was reading it a while ago. I got concerning far into the process before noticing.

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Resolves #5081.

Updates release management docs with instructions on updating public documentation.

## Testing

- Docs-only PR, review for clarity and correctness.

## Deployment

2. New installs.

### DIFF
--- a/docs/development/release_management.rst
+++ b/docs/development/release_management.rst
@@ -188,26 +188,10 @@ Release Process
 #. Ask Infrastructure to perform the DNS cutover to switch
    ``apt-qa.freedom.press`` to ``apt.freedom.press``. Once complete,
    the release is live.
-#. Make sure that the default branch of documentation is being built
-   off the tip of the release branch. Building from the branch instead
-   of a given tag enables us to more easily add documentation changes
-   after release. You should:
-
-   a. Log into readthedocs.
-   #. Navigate to **Projects** → **securedrop** → **Versions** →
-      **Inactive Versions** → **release/branch** → **Edit**.
-   #. Mark the branch as Active by checking the box and save your
-      changes. This will kick off a new docs build.
-   #. Once the documentation has built, it will appear in the version
-      selector at the bottom of the column of the.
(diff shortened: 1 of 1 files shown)


docs/development/release_management.rst
```diff
@@ -188,26 +188,10 @@ Release Process
...
-      Version**.
-   #. Select ``release/branch`` from the dropdown menu and save the
-      changes.
-   #. Verify that docs.securedrop.org redirects users to the
-      documentation built from the release branch.
-
+#. Issue a PR to merge the release branch changes into ``master``. Once the PR is
+   merged, verify that the `public documentation <https://docs.securedrop.org/>`_
+   refers to the new release version. If not, log in to ReadTheDocs and start a
+   build of the ``master`` version.
```
eloquence, line 194: The previous version said:
> Building from the branch instead of a given tag enables us to more easily add documentation changes after release.

With this new process, I think we should make the expectations for post-release docs updates even more explicit, to avoid confusion about what's permissible, and to avoid missing required steps.
 
What's the expected process going forward?
1) Will we still permit post-tag backports of docs changes into the release branch, without creating a new tag?
2) Will such docs changes then need to be individually ported into `master`?

eloquence: Per the discussion at our retro, we also want to clarify what merge strategy we want to use here (ideally one that is low-effort, since `master` is _only_ used for docs builds).

PR #5255, opened 2020-05-13
added Ansible parameter to use legacy password prompt text

### LINKED ISSUE #5254: Graphical updater fails repeatedly with "Administrator password incorrect"
During release QA for SecureDrop 1.3.0, we discovered a regression in the graphical SecureDrop updater for Tails workstations. Regardless of the user input, the updater fails with "Administrator password incorrect".

This appears to be the result of a change in Ansible output introduced as part of #5199. It does not appear to be related to #5169.

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #5254.

Sets `agnostic_become_prompt` to `False` in `ansible.cfg`. This sets the password prompt when ansible is called with `--ask-become-pass` to `SUDO password:`. When set to `True`, the prompt is set as `BECOME password:`. 

The default value was switched in Ansible 2.8, but the GUI updater expects the `False` and `SUDO password` version, so it needs to be set explicitly.

## Testing
In a prod instance (VMs or hardware):
- check out this branch in an Admin Workstation
- [ ] run `./securedrop-admin setup` - it completes without error
- follow the installation guide and verify that:
  - [ ] the password prompt for `./securedrop-admin install` is `SUDO password:`
  - [ ] the password prompt for `./securedrop-admin tailsconfig` is `SUDO password:`
  - [ ] both commands complete successfully.

## Deployment
This will be deployed with the next production release after merge - as it preserves existing behaviour it should not have any impact on the GUI updater.

### DIFF
--- a/install_files/ansible-base/ansible.cfg
+++ b/install_files/ansible-base/ansible.cfg
@@ -8,6 +8,9 @@ timeout=120
 # from site-specific vars, or Onion URLs for SSH over ATHS.
 inventory=inventory-dynamic
 
+[privilege_escalation]
+agnostic_become_prompt=False
+
 [ssh_connection]
 scp_if_ssh=True
 ssh_args = -o ControlMaster=auto -o ControlPersist=1200


zenmonkeykstop:
Agreed - I tested this via the GUI updater on Tails as follows:

- checked out this branch, tagged it locally as 1.3.0
- checked out 1.2.2, deleted the 1.2.2 tag (locally!)
- disabled signature verification in `admin/securedrop_admin/__init.py`, committed, and tagged as 1.2.2
- bumped network connection to trigger updater and followed regular flow

This worked, the tailsconfig phase completed successfully. This isn't a perfect test as it involves the surgery mentioned above, but it does increase my confidence in the fix .If someone else wants to run through a similar process, feel free.

rmol:
I repeated that process and can confirm that the updater has no problem with the escalation prompt after (this pending version of) 1.3.0 is checked out.

conorsch:
Merging based on @rmol's two-pass test report.

PR #5266, opened 2020-05-14
fix: Only run docs linting/linkcheck tests in master

### LINKED ISSUE #5264: modify circle ci configuration on master to only run docs related jobs
## Description

Since we currently are using master branch only for the purpose of building documentation, we should modify the jobs that are run on that branch to only those that are relevant for documentation purposes, which is the lint job only, as it runs the docs linting. We may even want to restrict further and only run the docs linting on master.

prateekj117:
## Status

Ready for review

## Description of Changes

As `master` branch is only used for the purpose of building documentation, we should modify the jobs that are run on that branch to only those that are relevant for documentation purposes, which is the lint job only, as it runs the docs linting. We may even want to restrict further and only run the docs linting on master.

Fixes #5264 

Ignored jobs other than `docs-linkcheck` for master branch. Also, configured `docs-linkcheck` to only run if changes are done in master branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -302,7 +302,7 @@ workflows:
             branches:
               only:
                 - master
-      - lint
+      - lint:
           filters:
             branches:
               ignore:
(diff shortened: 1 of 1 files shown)


prateekj117:
@redshiftzero One suggestion: I think we should rename the job `docs-linkcheck` to `docs-lint-linkcheck`. That gives a better description of the job now.

rmol:
Thanks. These CI failures are unrelated to your changes, and I think this is good to merge once we sort them out on our end.

PR #5167, opened 2020-03-18
Backport 1.2.2 changelog to develop

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Backports 1.2.2 changelog to develop

PR #5268, opened 2020-05-14
1.3.0 post release

creviera:
## Status

Ready for review 

## Description of Changes

Following post-release procedure: https://github.com/freedomofpress/securedrop/blob/develop/docs/development/release_management.rst#post-release

Cherry-picked:
* https://github.com/freedomofpress/securedrop/commit/c5a313d96779d6dd14a1abd16f9eec880d8efc9f
* https://github.com/freedomofpress/securedrop/commit/56281d6d5692648deba6ed9f57521b58b0a93244
* https://github.com/freedomofpress/securedrop/commit/873d58e110d67b68e697f88718ca2a30301fc316
* https://github.com/freedomofpress/securedrop/commit/0fdab9c70cf40d1f5fa0ec5a44a1aff0a084170b

## Testing

Make sure the correct commits were cherry-picked,  see: 
* https://github.com/freedomofpress/securedrop/compare/release/1.3.0
* https://github.com/freedomofpress/securedrop/pull/5167/commits


conorsch:
From https://github.com/freedomofpress/securedrop/blob/514b9e9726c557706f63a2fe3fdaead3e0eaefb1/docs/development/release_management.rst#post-release, two passages stand out for clarification:

> cherry-pick the release commits

and 

> bump the version numbers in preparation for the next release

We can handle those docs touch-ups in a follow-up PR!

creviera:
Added a commit after running `./update_version.sh 1.4.0rc1` but did not include the generated `1.4.0rc1.tag` file

docs/set_up_admin_tails.rst
```diff
@@ -131,8 +131,8 @@ signed with the release signing key:
 .. code:: sh
 
     cd ~/Persistent/securedrop/
-    git checkout 1.2.2
-    git tag -v 1.2.2
+    git checkout 1.4.0rc1
+    git tag -v 1.4.0rc1
```
eloquence, line 135: If you compare https://github.com/freedomofpress/securedrop/pull/5061/files you'll note how the docs were bumped from 1.1.0 to 1.2.0 during the 1.2.0 release, while other versions were bumped to 1.3.0rc1. 

The  problem with bumping them to 1.4.0rc1 here is that this particular page is end user documentation that is published to https://docs.securedrop.org/en/latest/set_up_admin_tails.html (which the user will see if they switch from the default `master` branch to `latest`, which is built from `develop`). We want that documentation to refer to a version that actually exists, in case a user tries to follow the docs.

docs/conf.py
```diff
@@ -57,9 +57,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '1.2.2'
+version = '1.4.0rc1'
```
eloquence, line 60: Similar to the comment below, this `conf.py` file is configuration for the documentation built to https://docs.securedrop.org/en/master/ . I believe it's expected to describe the current release, not the next one. See the changes to `conf.py` in a previous backport: https://github.com/freedomofpress/securedrop/pull/5061 for comparison

molecule/shared/stable.ver
```diff
@@ -1 +1 @@
-1.2.2
+1.4.0rc1
```
eloquence, line 1: Looking at https://github.com/freedomofpress/securedrop/pull/5061/files it looks like this version is also expected to describe the current production release, not the next one.

redshiftzero:
Cherry picked commits look good, regarding the edits to docs/conf.py/stable.ver: those were edited by `update_version.sh`, because as background: it basically looks to see if `~rc` is in the version number and if found, it _doesn't_ modify the docs/conf.py/stable.ver files. So, if you drop the last commit and re-run `update_version.sh` with `1.4.0~rc1` as the version number instead of `1.4.0rc1`, I think that will do the trick!

creviera:
bumped version to 1.4.0~rc1

PR #5269, opened 2020-05-18
Update Qubes docs references to debian-10; add tip re: aufs-dkms

### LINKED ISSUE #4965: docs encourage usage of Standalone VM
## Description

**TL;DR**: Development environment has root partition persistence at the same time it has passwordless sudo 

Qubes' architecture (particularly `non-persistent root partition`) allows for distrusting the kernel's privege escalation mechanisms (giving us the convenient "passwordless sudo"). But when creating a StandaloneVM those mechanism's are no longer effective.

The [docs](https://docs.securedrop.org/en/release-1.1.0/development/setup_development.html#qubes) encourage the creation of a `StandaloneVM` for development. For the understandable reasons of faster development. So far, from what I've seen in the project, after the initial there aren't that many changes one has to do with the packages because it is all with virtual environments.

If anyone is in the position where that is a problem, then I think they would have the knowldege to convert it to a StandaloneVM (which is quite simple. The same cannot be said for the reverse, I think)

## Comments

 Any solution here will increase the barrier of entry of a what is already quite a high one (In my case it took me a week just to get everything right with the staging and dev environments - with an [bug](https://github.com/freedomofpress/securedrop-workstation/issues/241) in the path).

So I don't know if this is something that you may want to address, but I just felt the need to bring it up.

If this is to be changed, I'd suggest bellow 3 possible approaches.

#### Approach 1 | little note in docs
Add note to the docs to encourage future developers opting for a AppVM instead for development while pointing them to the relevant qubes documentation on the implications (this would require minimal changes to the docs)

#### Approach 2 | AppVMs in docs
Change the docs to configure things as AppVMs.

#### Approach 3 | Replacing passwordless root access with Dom0 user prompt
Add a step that directs the user to: https://www.qubes-os.org/doc/vm-sudo/#replacing-passwordless-root-access-with-dom0-user-prompt which is what I currently do. But it's tricky and if the user messes up a step they could brick the VM (in the sense that they could get locked out)

eloquence:
## Description

The Qubes dev/staging docs still reference `debian-9`; we have since transitioned to `debian-10`. Also added a tip regarding the `aufs-dkms` dependency which can bite users when setting up Docker due to Qubes' custom kernel.

I did not attempt to resolve #4965 because we've not discussed that change as a team yet.

## Status

Ready for review

## Checklint
- [x] Linter says "ship it, fine with me!"

### DIFF
--- a/docs/development/qubes_staging.rst
+++ b/docs/development/qubes_staging.rst
@@ -15,7 +15,7 @@ Standalone (HVM) Qubes VMs for use with staging:
 - ``sd-staging-app-base``, a base VM for the *SecureDrop Application Server*
 - ``sd-staging-mon-base``, a base VM for the *SecureDrop Monitor Server*
 
-While the development VM, ``sd-dev``, is based on Debian 9, the other VMs
+While the development VM, ``sd-dev``, is based on Debian 10, the other VMs
 will be based on Ubuntu Xenial.
 
 .. note:: The staging server VM names were recently changed from ``sd-app`` and
--- a/docs/development/qubes_staging.rst
+++ b/docs/development/qubes_staging.rst
@@ -299,9 +299,9 @@ environment. In from the root of the SecureDrop project in ``sd-dev``, run:
 
    make staging
 
-The ``make staging`` command invokes the ``qubes-staging`` Molecule scenario. 
+The ``make staging`` command invokes the ``qubes-staging`` Molecule scenario.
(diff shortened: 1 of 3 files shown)


conorsch:
Tacked on a commit updating the core dev docs as well, to recommend continued use of "stable" Debian series. Did _not_ update the "tested with" notes, which still point to Stretch, since I haven't actually run through those exact steps on Buster recently. We can follow up with a small change there shortly.

PR #5273, opened 2020-05-19
Add decommissioning documentation

rocodes:
## Status

Ready for Review

## Description of Changes

Docs only, responds to support request for decommissioning docs 

## Testing

Manual review 

## Deployment
n/a

### DIFF
--- a/docs/decommission.rst
+++ b/docs/decommission.rst
@@ -25,6 +25,7 @@ SecureDrop instance.
       have a non-standard setup, you may have more than one *SVS USB*. It is
       important that you locate all of these USBs, since they hold the most
       sensitive data.
+
 #. **Optional: Save a backup.**
    If you want to save a backup of the *Application Server* (for example, to reinstall SecureDrop in the future using the same `.onion` address), follow
    our :doc:`backup guidelines <backup_and_restore>`. Once the backup has been
@@ -38,12 +39,31 @@ SecureDrop instance.
    them on the *SVS*, be sure you maintain access to the *Submission Private
    Key* found on the *SVS* so that you can decrypt them at a later time.
 #. **Optional: Delete submissions on the server.**
+   Log into the *Journalist Interface* and delete all sources to take advantage
+   of SecureDrop's secure deletion properties. Note that depending on the
+   number of sources on your server, it may take anywhere from several minutes
+   to an hour or more for the submissions to be completely deleted from the
+   server.
 
-   Log into the *Journalist Interface* and delete all sources to take advantage of SecureDrop's secure deletion properties. Depending on the number of sources on the server, you may need to perform this deletion in batches; please contact Support for assistance.
+   You can either leave the server ample time to complete this operation, or
+   monitor the progress by SSHing to the Application server and running
+
+   .. code:: sh
+
+      sudo journalctl -f
+
+   You will see repeated log lines that contain the following:
+
+   .. code:: sh
+
+      [Timestamp] app python [...] INFO Clearing shredder
+      [Timestamp] app python [...] INFO Files to delete: <number>
+
+   When the number of files to delete reaches 0, the process is complete.
 #. **Disconnect the firewall and the servers from the internet.**
    Be sure to inform your network administrator of any changes to devices on
    your network.
-#. **Wipe or destroy the USB drives.**
+#. **Wipe and destroy the USB drives.**
    Because the USB drives used for SecureDrop are all LUKS-encrypted,
    reformatting the USB drives (in particular, overwriting a portion of internal
    storage called the **LUKS header**) should be sufficient to make any existing
@@ -61,7 +81,7 @@ SecureDrop instance.
 
    You may also choose to destroy the drives by physical means, such as using a
    hammer or purpose-built shredder to pulverize or destroy the drive.
-#. **Wipe or destroy the storage drives on the servers.**
+#. **Wipe and destroy the storage drives on the servers.**
    SecureDrop submissions are stored GPG-encrypted on the *Application Server*.
    Unless your SecureDrop *Submission Key* is compromised (or a significant
    vulnerability in GPG is discovered), access to the servers does not guarantee
@@ -70,10 +90,10 @@ SecureDrop instance.
    That said, there may still be some sensitive information on the servers,
    including system logs and the SecureDrop database, which would yield
    information on the number of submissions and replies stored on the server.
-   This risk is partially mitigated by securely deleting submissions from the server, as described in a previous step.
-
-   Physically destroying or encrypting the storage drives on the servers are the
-   best ways to ensure that data on the drives cannot be recovered.
+   This risk is partially mitigated by securely deleting submissions from the
+   server, as described in a previous step; however, physically destroying or
+   encrypting the storage drives on the servers are the best ways to ensure
+   that data on the drives cannot be recovered.
 
    Physically destroying SSD drives is not as straightforward as destroying
    older hard drives, but drives can be pulverized, shredded, or incinerated,
@@ -91,10 +111,10 @@ SecureDrop instance.
    select "yes."
 #. **Destroy other Transfer or Export media, if applicable.**
 #. **Optional: Factory-reset the firewall.**
+#. **Update your Landing Page (tips page) to reflect the fact that your organization no longer has SecureDrop.**
 #. **Notify the SecureDrop Support team that your instance is no longer active.**
    If you have any questions about the decommissioning process, or about other
    secure communications options, please feel free to contact us at
    securedrop@freedom.press
    (`GPG encrypted <https://securedrop.org/sites/default/files/fpf-email.asc>`__)
    or via the `support portal <https://support.freedom.press/>`__.
-#. Update your Landing Page (tips page) to reflect the fact that your organization no longer has SecureDrop.
--- a/docs/decommission.rst
+++ b/docs/decommission.rst
@@ -1,0 +1,100 @@
+Decommission SecureDrop
+=======================
+
+The following steps will guide you through the decommissioning of your
+SecureDrop instance.
+
+#. **Put a notice in advance on your landing page to inform sources that your
+   instance will soon be retired.**
+   You may want to direct them to other secure methods of contacting you.
+#. **Locate and create an inventory of all your hardware.**
+     - *Journalist Workstation* USBs
+     - *Admin Workstation* USBs
+     - *Secure Viewing Station* USB
+     - *Secure Viewing Station* computer
+     - *Transfer* and *Export Devices* (USBs, optical drives, or external drives)
+     - Backup USBs/other storage media
+     - Servers
+     - Firewall
+
+   You may also want to inventory credentials, such as the email address or
+   alias and PGP key used for receiving OSSEC alerts, in order to retire them.
+
+   .. note:: The recommended SecureDrop setup includes only one *Secure Viewing
+      Station* USB. However, if you have been working remotely or
+      have a non-standard setup, you may have more than one *SVS USB*. It is
+      important that you locate all of these USBs, since they hold the most
+      sensitive data.
+#. **Optional: Save a backup.**
+   If you want to save a backup of the *Application Server* (for example, to reinstall SecureDrop in the future using the same `.onion` address), follow
+   our :doc:`backup guidelines <backup_and_restore>`. Once the backup has been
+   created, you can move it off of the *Admin Workstation* USB and onto an
+   encrypted device, such as a LUKS-encrypted drive. You will also require a
+   backup of the *Submission Key* found on the *SVS*.
+
+   If you do not require a server backup, you may choose to download specific
+   submissions, and store them in a secure manner (such as on an encrypted
+   drive). If you export and store these submissions without first decrypting
+   them on the *SVS*, be sure you maintain access to the *Submission Private
+   Key* found on the *SVS* so that you can decrypt them at a later time.
+#. **Optional: Delete submissions on the server.**
+
+   Log into the *Journalist Interface* and delete all sources to take advantage of SecureDrop's secure deletion properties. Depending on the number of sources on the server, you may need to perform this deletion in batches; please contact Support for assistance.
+#. **Disconnect the firewall and the servers from the internet.**
+   Be sure to inform your network administrator of any changes to devices on
+   your network.
+#. **Wipe or destroy the USB drives.**
+   Because the USB drives used for SecureDrop are all LUKS-encrypted,
+   reformatting the USB drives (in particular, overwriting a portion of internal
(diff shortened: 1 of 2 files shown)


rocodes:
@eloquence @zenmonkeykstop Tagging you for review as well, which I'm unfortunately not able to do in the Reviewers column.
  
This does not need to be merged in any hurry, but happy for someone to sanity-check these steps so I can share them with the org who requested them.

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
...
+   .. note:: The recommended SecureDrop setup includes only one *Secure Viewing
+      Station* USB. However, if you have been working remotely or
+      have a non-standard setup, you may have more than one *SVS USB*. It is
+      important that you locate all of these USBs, since they hold the most
+      sensitive data.
+#. **Decide whether to save a backup.**
+   If you want a backup, you can, follow our `backup guidelines`_. Once the
+   backup has been created, you can move it off of the *Admin Workstation* USB
+   and onto an encrypted device, such as a LUKS-encrypted drive. You will also
+   require the *SVS* USB and/or a backup of the SecureDrop Application Key.
```
eloquence, line 29: SecureDrop Application Key->*Submission Key* (standardized term)

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
...
+
+   .. caution:: Be **very** sure you are reformatting the right drive.
+      You may want to use the Secure Viewing Station laptop for this procedure
+      to reduce the risk of accidentally erasing a drive on your regular-use
+      machine.
+
+   You may also choose to destroy the drives by physical means, such as using a
+   hammer or purpose-built shredder to pulverize or destroy the drive.
+#. **Wipe or destroy the storage drives on the servers.**
+   SecureDrop submissions are stored GPG-encrypted on the Application server.
```
eloquence, line 52: Application server->*Application Server* (standardized term)

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
...
+   Unless your SecureDrop Application Key is compromised (or a significant
```
eloquence, line 53: *Submission Key*, again

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
...
+   - Servers
+   - Firewall
+
+   .. note:: The recommended SecureDrop setup includes only one *Secure Viewing
+      Station* USB. However, if you have been working remotely or
+      have a non-standard setup, you may have more than one *SVS USB*. It is
+      important that you locate all of these USBs, since they hold the most
+      sensitive data.
+#. **Decide whether to save a backup.**
+   If you want a backup, you can, follow our `backup guidelines`_. Once the
```
eloquence, line 26: "If you want a backup, you can, follow"->"If you want to save a backup, follow"

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
...
+   Physically destroying SSD drives is not as straightforward as destroying
+   older hard drives, and requires some research. Depending on your threat
+   model, you may choose instead to write over the information on the existing
+   drives. One way to do this is to re-install a clean version of Ubuntu server
+   with full-disk encryption enabled. During the disk partitioning portion of
+   the installation wizard, select *Guided - use entire disk and set up
+   encrypted LVM*.
+#. **Destroy other Transfer or Export media, if applicable.**
+#. **Factory-reset the firewall (optional).**
+#. **Notify the SecureDrop Support team that your instance is no longer active.**
```
eloquence, line 73: I'd make another reference to the landing page here, since the initial update you recommend is only to give sources a heads-up.

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
...
+   If you have any questions about the decommissioning process, or about other
+   secure communications options, please feel free to contact us at
+   securedrop@freedom.press
+   (`GPG encrypted <https://securedrop.org/sites/default/files/fpf-email.asc>`__)
+   or via the `support portal <https://support.freedom.press/>`__.
+
+.. _backup guidelines: https://docs.securedrop.org/en/master/backup_and_restore.html#backing-up
```
eloquence, line 80: To ensure stability of links, it's best to avoid using absolute URLs for references within the documentation itself; you can use a `:doc:` type link to refer to the whole doc, or insert a named reference in the target document near a specific anchor, and then use a `:ref:` link.

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
...
+
+The following steps will guide you through the decommissioning of your
+SecureDrop instance.
+
+#. **Put a notice in advance on your landing page to inform sources that your
+   instance will soon be retiring.**
+   You may want to direct them to other secure methods of contacting you.
+#. **Locate and create an inventory of all your hardware.**
+   - *Journalist Workstation* USBs
+   - *Admin Workstation* USB(s)
```
eloquence, line 12: nit: you use both `USB(s)` and `USBs`

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
...
+   - *Secure Viewing Station* USB
+   - *Secure Viewing Station* computer
+   - Transfer and Export devices (USBs, optical drives, or external drives)
```
eloquence, line 15: *Transfer* and *Export Devices* (standardized terms)

docs/decommission.rst
eloquence: This does not clearly speak to me right now to specific motivations or goals (why would I want to backup, and what steps do I need to follow to do so). One suggestion would be to organize it into two separate steps:

- download any submissions they want to hold on to, and store them in such a manner they'll be able to decrypt/access them securely (e.g., decrypt and store on an encrypted drive, or store alongside Submission Key), then _delete_ all submissions on server (which will trigger SD's own secure deletion);
- if they want to preserve the server configuration for any reason (e.g., may want to keep option open of spining up SD under the same .onion address later), follow the server backup procedures. Otherwise they can skip that step, as long as they've followed the previous one.

For a decommissioning document we can be clear that certain data will be _lost_ if they follow this process, such as any credentials stored on the Admin Workstation USB. I don't think the whole decommissioning process needs to be optionally reversible.

eloquence:
This is looking great already; added some nits, and thoughts re: backup. Not commenting on the data destruction aspect, very curious for @emkll's feedback on the recommendations here.

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
+Decommission SecureDrop
+=======================
+
+The following steps will guide you through the decommissioning of your
+SecureDrop instance.
+
+#. **Put a notice in advance on your landing page to inform sources that your
```
zenmonkeykstop, line 7: s/be retiring/be retired/
An instance can be retired, but it can't retire - if you see what I mean.

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
...
+   instance will soon be retiring.**
+   You may want to direct them to other secure methods of contacting you.
+#. **Locate and create an inventory of all your hardware.**
+   - *Journalist Workstation* USBs
```
zenmonkeykstop, line 11: Indentation of this list is broken.

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
...
+   That said, there is still some sensitive information on the servers,
+   including system logs and the SecureDrop database, which would yield
+   information on the number of submissions and replies stored on the server.
+
+   Physically destroying or encrypting the storage drives on the servers are the
+   best ways to ensure that data on the drives cannot be recovered.
+
+   Physically destroying SSD drives is not as straightforward as destroying
+   older hard drives, and requires some research. Depending on your threat
+   model, you may choose instead to write over the information on the existing
```
zenmonkeykstop, line 66: Might be worth avoiding domain-specific jargon here - non-security folk may have a woolly or no idea of their threat model, and other security folk might prefer a different term.

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
...
+   - Backup USBs/other storage media
+   - Servers
+   - Firewall
+
+   .. note:: The recommended SecureDrop setup includes only one *Secure Viewing
+      Station* USB. However, if you have been working remotely or
+      have a non-standard setup, you may have more than one *SVS USB*. It is
+      important that you locate all of these USBs, since they hold the most
+      sensitive data.
+#. **Decide whether to save a backup.**
```
zenmonkeykstop, line 25: s/Decide whether to/ Optional: /

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
...
+   access to the submissions and messages you have received.
+
+   That said, there is still some sensitive information on the servers,
+   including system logs and the SecureDrop database, which would yield
+   information on the number of submissions and replies stored on the server.
+
+   Physically destroying or encrypting the storage drives on the servers are the
+   best ways to ensure that data on the drives cannot be recovered.
+
+   Physically destroying SSD drives is not as straightforward as destroying
```
emkll, line 64: It could be interesting to have external references to these guidelines. Similar to USB flash drives, I was under the impression that physically destroying the NAND chips and DRAM chips here would be sufficient.

For example NIST-SP800-88 [1] recommends, for both SSDs and spinning platters, which is in-line with your recommendations for the USB drives above:
```
Shred, Disintegrate, Pulverize, or Incinerate by burning the device in a licensed incinerator.
```

[1] : https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-88r1.pdf

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
...
+   older hard drives, and requires some research. Depending on your threat
+   model, you may choose instead to write over the information on the existing
+   drives. One way to do this is to re-install a clean version of Ubuntu server
```
emkll, line 67: There's an option to "Overwrite free space" in the install process, this may need to be selected here, and would take extra time (a lot of time on spinning platter). I would suggest instead either referencing or suggesting ATA Secure Erase in the nist document mentioned above

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
...
+#. **Locate and create an inventory of all your hardware.**
+   - *Journalist Workstation* USBs
+   - *Admin Workstation* USB(s)
+   - *Secure Viewing Station* USB
+   - *Secure Viewing Station* computer
+   - Transfer and Export devices (USBs, optical drives, or external drives)
+   - Backup USBs/other storage media
+   - Servers
+   - Firewall
+
```
emkll, line 19: nit: There's also the email account for ossec notifications, perhaps that one can be deactivated as well

docs/decommission.rst
```diff
@@ -0,0 +1,80 @@
+Decommission SecureDrop
+=======================
+
+The following steps will guide you through the decommissioning of your
+SecureDrop instance.
+
+#. **Put a notice in advance on your landing page to inform sources that your
+   instance will soon be retiring.**
+   You may want to direct them to other secure methods of contacting you.
+#. **Locate and create an inventory of all your hardware.**
```
emkll, line 10: This is a great idea, and I wonder if we should also suggest this as part of the install process as well (this could be something tracked on the admin persistent volume)

docs/decommission.rst
rocodes: Yes, you're right--but the reason that I said it was 'not as straightforward' is that people might not have enough understanding to identify and destroy those specific chips, and might assume that the drive is inoperable when it is actually not (vs for example drilling holes through random parts of an HDD), so if they're opting for physical destruction they should do some research. This may seem overly paranoid, but it might not be considering we advise against FDE on the servers.  

I can be a bit more specific here though.

docs/decommission.rst
rocodes: This part is the main part that I wanted your review on so I'm glad you're here :) 

* Good point re overwriting space in install wizard, I'll put that in 
* I hesitated on mentioning the ATA Secure Erase because the implementation depends on the manufacturer and sometimes that might not cut it[1]:

> Of the 12 drives we tested[,...] [e]ight of the drives reported that they supported the ATA SECURITY feature set. One of these encrypts data, so we could not verify if the sanitization was successful. Of the remaining seven, only four executed the “ERASE UNIT” command reliably. Drive B’s behavior is the most disturbing: it reported that sanitization was successful, but all the data remained intact. In fact, the filesystem was still mountable. [...] The wide variance among the drives leads us to conclude that each implementation of the security commands must be individually tested before it can be trusted to properly sanitize the drive.

[1] https://www.usenix.org/legacy/events/fast11/tech/full_papers/Wei.pdf

docs/decommission.rst
rocodes: fixed in ff3b336

docs/decommission.rst
rocodes: fixed in ff3b336

docs/decommission.rst
rocodes: Fixed (and all similar standardization comments) in f38127f

docs/decommission.rst
rocodes: Updated in f38127f to reflect, as briefly as possible, the two different paths (server backup/ saving select files), and marked this step as 'optional.' 
As an aside/general point, I'm partly assuming that there will be some support contact before an org decides to follow these steps, so they may not be as fully verbose as some of our other documentation.

docs/decommission.rst
rocodes: Added in f38127f

docs/decommission.rst
rocodes: fixed in f38127f

docs/decommission.rst
rocodes: fixed in f38127f

docs/decommission.rst
rocodes: fixed in f38127f

docs/decommission.rst
rocodes: fixed in f38127 and all optional steps formatted this way for consistency

docs/decommission.rst
rocodes: See f38127f

docs/decommission.rst
rocodes: Added in f38127f

docs/decommission.rst
rocodes: Filed separate docs issue for myself in #5275

docs/decommission.rst
rocodes: See d89827e: I have included cursory information on ATA Secure Erase as one of the options for erasing the drives.

rocodes:
> Took another pass through this, thanks for the changes @rocodes, they look good to me. My final comment is regarding the drives: changes here have the heading `wipe or destroy`, I would suggest we recommend `wipe and destroy` as much as possible, given some drives being unencrypted, and destruction also prevents re-use.

amended in e88238c

PR #5284, opened 2020-05-30
Handle case of deleted journalists

### LINKED ISSUE #5232: We should not allow journalist username `deleted`
Because in https://github.com/freedomofpress/securedrop/pull/5178 we are now marking `deleted` as the name/uuid of any journalist account which is deleted from the system, we should not allow the term `deleted` as the journalist name (via the web interface). I think this will help in reducing confusion in future.

prateekj117:
## Status

Ready for review

## Description of Changes

Fixes #5232 

Because of #5178, deleted journalist `username` and `uuid` is set to `deleted`. We should not allow the deleted journalist to login. Also, we can differentiate it from other not deleted journalists whose username is deleted by comparing both the `username` and `uuid` rather than just username. Also, we should not allow adding new journalist with username `deleted` for less confusion. 

## Testing

Make a test journalist with username `deleted`, it would fail (due to changes in `form.py` file). Undo those changes and make a new journalist with username `deleted` and try logging in using credentials of that journalist, it should succeed. Change `uuid` of that journalist to `deleted`, it should fail to login again for that journalist.

### DIFF
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -132,6 +132,10 @@ password manager. Then, you will select whether you would like them to also be
 an admin (this allows them to add or delete other journalist accounts), and
 whether they will be using FreeOTP or a YubiKey for two-factor authentication.
 
+.. note::
+  We don't allow the username **deleted** as we use it to mark the
+  journalist which are deleted from the system.
+
 FreeOTP
 ^^^^^^^
 
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -134,7 +134,7 @@ whether they will be using FreeOTP or a YubiKey for two-factor authentication.
 
 .. note::
   We don't allow the username **deleted** as we use it to mark the
-  journalist which are deleted from the system.
+  journalists which are deleted from the system.
 
 FreeOTP
 ^^^^^^^
--- a/securedrop/journalist_app/forms.py
+++ b/securedrop/journalist_app/forms.py
@@ -40,7 +40,7 @@ def name_length_validation(form, field):
 def check_invalid_usernames(form, field):
     if field.data in Journalist.INVALID_USERNAMES:
         raise ValidationError(gettext(
-            "Invalid username"))
+            "This username is invalid because it is reserved for internal use by the software."))
 
 
 class NewUserForm(FlaskForm):
--- a/securedrop/journalist_app/forms.py
+++ b/securedrop/journalist_app/forms.py
@@ -37,10 +37,16 @@ def name_length_validation(form, field):
             .format(max_chars=Journalist.MAX_NAME_LEN)))
 
 
+def check_username_deleted(form, field):
+    if field.data == 'deleted':
+        raise ValidationError(gettext(
+            "Invalid username '{}'".format(field.data)))
+
+
 class NewUserForm(FlaskForm):
     username = StringField('username', validators=[
         InputRequired(message=gettext('This field is required.')),
-        minimum_length_validation
+        minimum_length_validation, check_username_deleted
     ])
     first_name = StringField('first_name', validators=[name_length_validation, Optional()])
     last_name = StringField('last_name', validators=[name_length_validation, Optional()])
--- a/securedrop/journalist_app/forms.py
+++ b/securedrop/journalist_app/forms.py
@@ -38,10 +38,9 @@ def name_length_validation(form, field):
 
 
 def check_invalid_usernames(form, field):
-    invalid_usernames = ['deleted']
-    if field.data in invalid_usernames:
+    if field.data in Journalist.INVALID_USERNAMES:
         raise ValidationError(gettext(
-            "Invalid username '{}'".format(field.data)))
+            "Invalid username"))
 
 
 class NewUserForm(FlaskForm):
--- a/securedrop/journalist_app/forms.py
+++ b/securedrop/journalist_app/forms.py
@@ -37,8 +37,9 @@ def name_length_validation(form, field):
             .format(max_chars=Journalist.MAX_NAME_LEN)))
 
 
-def check_username_deleted(form, field):
-    if field.data == 'deleted':
+def check_invalid_usernames(form, field):
+    invalid_usernames = ['deleted']
+    if field.data in invalid_usernames:
         raise ValidationError(gettext(
             "Invalid username '{}'".format(field.data)))
 
@@ -46,7 +47,7 @@ def check_username_deleted(form, field):
 class NewUserForm(FlaskForm):
     username = StringField('username', validators=[
         InputRequired(message=gettext('This field is required.')),
-        minimum_length_validation, check_username_deleted
+        minimum_length_validation, check_invalid_usernames
     ])
     first_name = StringField('first_name', validators=[name_length_validation, Optional()])
     last_name = StringField('last_name', validators=[name_length_validation, Optional()])
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -414,6 +414,7 @@ class Journalist(db.Model):
     MIN_USERNAME_LEN = 3
     MIN_NAME_LEN = 0
     MAX_NAME_LEN = 100
+    INVALID_USERNAMES = ['deleted']
 
     def __init__(self,
                  username: str,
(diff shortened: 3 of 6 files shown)


securedrop/journalist_app/forms.py
```diff
@@ -37,10 +37,16 @@ def name_length_validation(form, field):
             .format(max_chars=Journalist.MAX_NAME_LEN)))
 
 
+def check_username_deleted(form, field):
```
kushaldas: Convert it to `check_invalid_usernames` and verify against a list.

prateekj117:
@kushaldas Done. Ready for another review. Let me know if I missed something.

kushaldas:
@prateekj117 this PR needs a rebase with `develop`.

kushaldas:
@prateekj117 Also, can you please add a functional test for the same? As right now the error should propagate to the admin nicely.

prateekj117:
@kushaldas Done. Ready for another review.

prateekj117:
@kushaldas One thing that can be missing from here is I think we can also check the exception that is being raised in unit tests. Let me know if that is required.

kushaldas:
>Hey @kushaldas. The thing is, it's an endpoint, I am not sure how I will write a test for checking exception raising. app.post won't be able to check the exception as it happens on the server-side and app.post would just return a response. Can you share an example?

This whole code base is server side. I don't understand what do you mean by endpoint in this case. You have to test that the right error message is showing in the proper way in the web application, Please have a look at https://github.com/freedomofpress/securedrop/blob/develop/securedrop/tests/functional/test_source_notfound.py or any other file in that directory for examples.

prateekj117:
@kushaldas Now that I see the examples, I get it. I will correct it. Thanks !!!

prateekj117:
@kushaldas Done.

docs/admin.rst
```diff
@@ -132,6 +132,10 @@ password manager. Then, you will select whether you would like them to also be
 an admin (this allows them to add or delete other journalist accounts), and
 whether they will be using FreeOTP or a YubiKey for two-factor authentication.
 
+.. note::
+  We don't allow the username **deleted** as we use it to mark the
+  journalist which are deleted from the system.
```
rmol: Should be *journalists* (plural) which are deleted.

securedrop/journalist_app/forms.py
```diff
@@ -37,10 +37,17 @@ def name_length_validation(form, field):
             .format(max_chars=Journalist.MAX_NAME_LEN)))
 
 
+def check_invalid_usernames(form, field):
+    invalid_usernames = ['deleted']
```
rmol: We should keep the canonical list of invalid usernames on `models.Journalist` with other [constraints on usernames](https://github.com/freedomofpress/securedrop/blob/195ead9e16fceb44dbf1ea74f8414231d1ad52fa/securedrop/models.py#L414).

securedrop/journalist_app/forms.py
```diff
@@ -37,10 +37,17 @@ def name_length_validation(form, field):
...
+    if field.data in invalid_usernames:
+        raise ValidationError(gettext(
+            "Invalid username '{}'".format(field.data)))
```
rmol: Let's just leave the message at "Invalid username". The value is visible in the form field immediately above, and not having placeholders simplifies translation.

securedrop/models.py
```diff
@@ -641,12 +641,19 @@ def login(cls,
               username: str,
               password: str,
               token: str) -> 'Journalist':
+
+        invalid_usernames = ['deleted']
```
rmol: Again, let's use an attribute on `Journalist` instead of a local variable.

securedrop/models.py
```diff
@@ -641,12 +641,19 @@ def login(cls,
...
+
         try:
             user = Journalist.query.filter_by(username=username).one()
         except NoResultFound:
             raise InvalidUsernameException(
                 "invalid username '{}'".format(username))
 
+        if user.username in invalid_usernames and user.uuid in invalid_usernames:
```
rmol: We should raise if either username or UUID are invalid.

securedrop/models.py
```diff
@@ -641,12 +641,19 @@ def login(cls,
...
+            raise InvalidUsernameException(
+                "Invalid username '{}'".format(username))
```
rmol: Again, no reason to include the username in the message.

securedrop/tests/functional/journalist_navigation_steps.py
```diff
@@ -322,6 +322,27 @@ def _add_user(self, username, first_name="", last_name="", is_admin=False, hotp=
...
+        invalid_username = 'deleted'
+
+        self.safe_send_keys_by_css_selector('input[name="username"]', invalid_username)
+
+        self.safe_click_by_css_selector("button[type=submit]")
+
+        self.wait_for(lambda: self.driver.find_element_by_css_selector(".form-validation-error"))
+
+        error_msg = self.driver.find_element_by_css_selector(".form-validation-error")
+        assert "Invalid username '{}'".format(invalid_username) in error_msg.text
```
rmol: The expected string will have to change if the message doesn't include the username.

securedrop/tests/test_journalist.py
```diff
@@ -1076,6 +1076,65 @@ def test_admin_add_user(journalist_app, test_admin):
...
+        _login_user(app, test_admin['username'], test_admin['password'], test_admin['otp_secret'])
+
+        resp = app.post(url_for('admin.add_user'),
+                        data=dict(username=username,
+                                  first_name='',
+                                  last_name='',
+                                  password=VALID_PASSWORD,
+                                  is_admin=None))
+
+    assert "Invalid username '{}'".format(username) in resp.data.decode('utf-8')
```
rmol: The expected string will have to change if the message doesn't include the username.

securedrop/models.py
prateekj117: @rmol I don't think this should have the `or`. What if a user already has an invalid username, `deleted` in this case. They should be able to login. What I understood from [this](https://github.com/freedomofpress/securedrop/issues/5232#issuecomment-624740370) was that we can differentiate legitimate users from `deleted` user using uuid, so we need to check both.

rmol: Ah. Well, that's unfortunate, but yes, let's keep the `and`.

prateekj117:
@rmol Did the changes. Ready for another review.

securedrop/journalist_app/forms.py
```diff
@@ -37,10 +37,16 @@ def name_length_validation(form, field):
             .format(max_chars=Journalist.MAX_NAME_LEN)))
 
 
+def check_invalid_usernames(form, field):
+    if field.data in Journalist.INVALID_USERNAMES:
+        raise ValidationError(gettext(
+            "Invalid username"))
```
kushaldas: Or maybe a text saying "This username is not allowed in SecureDrop", @emkll @eloquence some intput here will be useful.

securedrop/models.py
```diff
@@ -641,12 +642,18 @@ def login(cls,
...
               password: str,
               token: str) -> 'Journalist':
+
         try:
             user = Journalist.query.filter_by(username=username).one()
         except NoResultFound:
             raise InvalidUsernameException(
                 "invalid username '{}'".format(username))
 
+        if user.username in Journalist.INVALID_USERNAMES and \
```
kushaldas, line 652: The idea was to stop the creation of invalid user names, but we should not block log in if someone already created such username. @redshiftzero mentioned this in https://github.com/freedomofpress/securedrop/issues/5232#issuecomment-624740370

prateekj117: @kushaldas Yes. That is why I also checked the value of `user.uuid`. Log in would not be interrupted if a user already has an invalid username. It will only be interrupted if both `user.username` and `user.uuid` is invalid.

securedrop/journalist_app/forms.py
eloquence: Given the 0.001% likelihood that any legit user is every going to see this message I'm not too fussed, but I'd suggest going with:
> This username is invalid because it is reserved for internal use by the software.

prateekj117: Should I make this change?

eloquence: Yes please, thank you :)

PR #5287, opened 2020-06-02
Add system configuration check script to securedrop-ossec packages

rmol:
## Status

Ready for review

## Description of Changes

Add checksdconfig.py to both securedrop-ossec-agent and securedrop-ossec-server, so that the configuration of both app and mon servers can be checked.

## Testing

- `make build-debs && make staging`
- as root on each server:
  - edit `/var/ossec/etc/ossec.conf` to change the frequency at which `checksdconfig.py` is run to 30 seconds
  - restart OSSEC with `systemctl restart ossec`
  -  run `iptables --rename-chain LOGNDROP FOO`
- wait 30 seconds, then as root on the mon server, verify that `/var/ossec/logs/alerts/alerts.log` contains level 12 alerts saying that there are system configuration errors on both servers, saying  `The iptables default drop rules are incorrect.` .

## Deployment

This adds a new script for checking the system configuration of both app and mon servers, with corresponding alerts that could be sent to the SecureDrop administrators.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,7 +72,6 @@ lib/
 lib64/
 parts/
 sdist/
-var/
 wheels/
 *.egg-info/
 .installed.cfg
--- a/install_files/securedrop-ossec-agent/var/ossec/checksdconfig.py
+++ b/install_files/securedrop-ossec-agent/var/ossec/checksdconfig.py
@@ -1,0 +1,90 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import argparse
+import subprocess
+import sys
+
+
+IPTABLES_RULES_UNCONFIGURED = {
+    "all": ["-P INPUT ACCEPT", "-P FORWARD ACCEPT", "-P OUTPUT ACCEPT"]
+}
+
+
+IPTABLES_RULES_DEFAULT_DROP = {
+    "policies": [
+        "-P INPUT DROP",
+        "-P FORWARD DROP",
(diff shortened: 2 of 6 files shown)


zenmonkeykstop:
- `make build-debs && make staging`
- as root on each server:
  - edit `/var/ossec/etc/ossec.conf` to change the frequency at which `checksdconfig.py` is run to 30 seconds
  - restart OSSEC with `systemctl restart ossec`
  -  run `iptables --rename-chain LOGNDROP FOO`
- wait 30 seconds, 
- [x] as root on the mon server, verify that `/var/ossec/logs/alerts/alerts.log` contains level 12 alerts saying that there are system configuration errors on both servers, saying  `The iptables default drop rules are incorrect.` .
- rebooted servers to restore iptables rules, then edited /etc/network/iptables/rules_v4 to break config, then rebooted again
- [x] verified that an alert was issued: 
  ```
  Rule: 400900 (level 12) -> 'Indicates a problem with the configuration of the SecureDrop servers.'
  ossec: output: '/var/ossec/checksdconfig.py': System configuration error: The iptables default drop rules are incorrect.
  ```

conorsch:
> Approved for merge once CI is unstuck

Reference to #5293. CI passing here: https://app.circleci.com/pipelines/github/freedomofpress/securedrop/249/workflows/54fb7ff2-8697-4fdd-90f9-0d7cfbd11b1a/jobs/40923/steps, so we are good to go

PR #5298, opened 2020-06-05
Update syntax for existing type annotations

### LINKED ISSUE #5282: Use PEP484 syntax for existing type annotations
As part of the wider type annotation work in https://github.com/freedomofpress/securedrop/issues/4399, we should update existing annotations to use the newer syntax now that we've migrated to Python 3.

pierwill:
## Status

Ready for review

## Description of Changes

Update syntax for existing type annotations.

Should fix #5282.

### DIFF
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -135,8 +135,7 @@ class CryptoUtil:
         if not rm.check_secure_delete_capability():
             raise AssertionError("Secure file deletion is not possible.")
 
-    def get_wordlist(self, locale):
-        # type: (Text) -> List[str]
+    def get_wordlist(self, locale: Text) -> List[str]:
         """" Ensure the wordlist for the desired locale is read and available
         in the words global variable. If there is no wordlist for the
         desired local, fallback to the default english wordlist.
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -135,7 +135,7 @@ class CryptoUtil:
         if not rm.check_secure_delete_capability():
             raise AssertionError("Secure file deletion is not possible.")
 
-    def get_wordlist(self, locale: Text) -> List[str]:
+    def get_wordlist(self, locale: 'Text') -> 'List[str]':
         """" Ensure the wordlist for the desired locale is read and available
         in the words global variable. If there is no wordlist for the
         desired local, fallback to the default english wordlist.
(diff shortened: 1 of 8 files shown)


pierwill:
Not sure what's holding up CI here...

kushaldas:
After digging more into this I think `pylint` is not ready yet for type annotations :( 

Even after I bring in all the imports into the main code from the `if typing.TYPE_CHECKING` condition, `pylint` is still failing. 

Related similar issues: 

- https://github.com/PyCQA/pylint/issues/3285
- https://github.com/PyCQA/pylint/issues/3382

We should discuss about this during the standup tonight. 

@pierwill Please hold of  any future PR like this until we figure out a way. Otherwise `pylint` will keep failing them.

redshiftzero:
It makes sense to migrate away from type comments now that we're on Python 3. To address the lint failures  we have the following options:

1. Not use pylint. Bad because it's a pretty useful tool.
2. Use pylint but litter the code with `noqa` exceptions. This is ugly.
3. Use pylint but add a global exception for E0601 / `used-before-assignment` which is causing the issue. Bad because it might cause bugs to be missed in the future for that specific lint rule. 
4. Add imports needed for type checking to the imports at runtime and combine them to sidestep the _other_ lint error @kushaldas found. Not a bad option but it means that it's unclear which imports are needed at runtime and which are needed solely for type checking, as well as obviously that there would be unnecessary imports at runtime. 
5. For the imports that are only performed when `TYPE_CHECKING`, we can express the type hint as a string literal (which mypy will still use). 

I think option 5 makes the most sense. We can revert this when PyCQA/pylint#3285 is addressed upstream. See commit d3707e6997fe0b96b379053149613202a3136a9b which does this on top of this PR and has a passing lint target. Let me know what you think @kushaldas @pierwill !

kushaldas:
>I think option 5 makes the most sense. We can revert this when PyCQA/pylint#3285 is addressed upstream. See commit d3707e6 which does this on top of this PR and has a passing lint target. Let me know what you think @kushaldas @pierwill !

Let us go with 5. We will revert back when `pylint` is ready for type hinting.

rmol:
Counterpoint: we're not even using pylint on our other projects, just flake8, which evidently doesn't have these problems with type annotations. I think it'd be worth reevaluating whether we really need both, instead of letting one linter's shortcomings dictate a different style of annotation to what we're using everywhere else.

redshiftzero:
Hmm fair, I'll file an issue for considering pylint removal, when testing this branch I did find more linting issues in pylint compared to flake8 (worth investigating what the difference is). However, we'll still get [F821](https://www.flake8rules.com/rules/F821.html) lint failures even with flake8 only due to our use of `if TYPE_CHECKING` which we don't hit in other repos because we do the type-checking imports also at runtime, i.e. we'd still need to use the string literal type annotations. Let me know if you think I'm misunderstanding something.

Issue #4399, opened 2019-05-01
Add type annotations to securedrop app code


redshiftzero:
In the securedrop application code we use [typing](https://pypi.org/project/typing/) and [mypy](https://mypy.readthedocs.io/en/latest/) and have annotations only in a few files. We should have individual PRs that:

- Add type hints to the codebase one file at a time (since small diffs easier to review)
- Enforce 100% annotation in those files in CI (done via this makefile target: https://github.com/freedomofpress/securedrop/blob/develop/Makefile#L37)

Please comment here what file you are taking to avoid duplication of work

See also: #2795

redshiftzero:
@rmol discovered this problem when reviewing #4408:

> We have a bigger problem with derivations of sqlalchemy.Model, as they effectively resolve to Any and thus aren't really checked. We talked to the mypy developers yesterday and have an idea for fixing this, but we'll tackle that in a later PR. It shouldn't hold this one up.

The idea mentioned there was:

```
# SqlAlchemy's Model is typed Any, so mypy won't check model classes
# unless we make them more specific.
if typing.TYPE_CHECKING:
    class BaseModel:
        def query(self, *args, **kwargs):
            pass
else:
    BaseModel = db.Model
```

pierwill:
👋 Volunteering to help with this!

pierwill:
Currently working on [source_app/utils.py](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/source_app/utils.py).

nabla-c0d3:
FYI after all my current PRs are merged, I will make a final PR to enforce/require type annotations across the whole repo.

nabla-c0d3:
This issue can now be closed: typing annotations are enforced across all files outside of the test suite.

PR #5300, opened 2020-06-05
Add type annotations for source_app/utils.py

pierwill:
## Status

Ready for review.

I've posed a few questions on specific types in the comments below.

## Description of Changes

This is work towards #4399.

### DIFF
--- a/securedrop/source_app/utils.py
+++ b/securedrop/source_app/utils.py
@@ -8,18 +8,24 @@ from sqlalchemy import create_engine
 from sqlalchemy.orm import sessionmaker
 from threading import Thread
 
+import typing
+
 import i18n
 import re
 
-from crypto_util import CryptoException
+from crypto_util import CryptoUtil, CryptoException
 from models import Source
+from sdconfig import SDConfig
+
+if typing.TYPE_CHECKING:
+    from typing import Optional  # noqa: F401
 
 
-def logged_in():
+def logged_in() -> bool:
     return 'logged_in' in session
 
 
-def valid_codename(codename):
+def valid_codename(codename: str) -> bool:
     try:
         filesystem_id = current_app.crypto_util.hash_codename(codename)
     except CryptoException as e:
@@ -32,7 +38,7 @@ def valid_codename(codename):
     return source is not None
 
 
-def generate_unique_codename(config):
+def generate_unique_codename(config: SDConfig) -> str:
     """Generate random codenames until we get an unused one"""
     while True:
         codename = current_app.crypto_util.genrandomid(
@@ -61,20 +67,23 @@ def generate_unique_codename(config):
             return codename
 
 
-def get_entropy_estimate():
+def get_entropy_estimate() -> int:
     with io.open('/proc/sys/kernel/random/entropy_avail') as f:
         return int(f.read())
 
 
-def asynchronous(f):
-    def wrapper(*args, **kwargs):
+def asynchronous(f):               # type: ignore
+    def wrapper(*args, **kwargs):  # type: ignore
         thread = Thread(target=f, args=args, kwargs=kwargs)
         thread.start()
     return wrapper
 
 
 @asynchronous
-def async_genkey(crypto_util_, db_uri, filesystem_id, codename):
+def async_genkey(crypto_util_: CryptoUtil,
+                 db_uri: str,
+                 filesystem_id: str,
+                 codename: str) -> None:
     # We pass in the `crypto_util_` so we don't have to reference `current_app`
     # here. The app might not have a pushed context during testing which would
     # cause this asynchronous function to break.
@@ -96,7 +105,7 @@ def async_genkey(crypto_util_, db_uri, filesystem_id, codename):
     session.close()
 
 
-def normalize_timestamps(filesystem_id):
+def normalize_timestamps(filesystem_id: str) -> None:
     """
     Update the timestamps on all of the source's submissions to match that of
     the latest submission. This minimizes metadata that could be useful to
@@ -115,7 +124,7 @@ def normalize_timestamps(filesystem_id):
                 rc)
 
 
-def check_url_file(path, regexp):
+def check_url_file(path: str, regexp: str) -> 'Optional[str]':
     """
     Check that a file exists at the path given and contains a single line
     matching the regexp. Used for checking the source interface address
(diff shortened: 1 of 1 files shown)


securedrop/source_app/utils.py
```diff
@@ -61,11 +63,10 @@ def generate_unique_codename(config):
             return codename
 
 
-def get_entropy_estimate():
+def get_entropy_estimate() -> int:
     with io.open('/proc/sys/kernel/random/entropy_avail') as f:
         return int(f.read())
 
-
 def asynchronous(f):
```
pierwill, line 70: Unsure how to type these asynchronous functions...

pierwill:
Unsure how to annotation the function at [line 70](https://github.com/freedomofpress/securedrop/pull/5300/files#diff-5cbe30f9891d7d06e461eb657c8be0e0R70).

securedrop/source_app/utils.py
```diff
@@ -8,18 +8,20 @@
...
+from sdconfig import SDConfig
 
-
-def logged_in():
+def logged_in() -> bool:
     return 'logged_in' in session
 
 
-def valid_codename(codename):
+def valid_codename(codename: str) -> bool:
```
pierwill, line 28: Should this be `-> Optional[bool]`?

securedrop/source_app/utils.py
```diff
@@ -74,7 +75,10 @@ def wrapper(*args, **kwargs):
 
 
 @asynchronous
-def async_genkey(crypto_util_, db_uri, filesystem_id, codename):
+def async_genkey(crypto_util_,
```
pierwill: Do we have a type for this `crypto_util_`?

securedrop/source_app/utils.py
```diff
@@ -61,11 +63,10 @@ def generate_unique_codename(config):
...
 
-def get_entropy_estimate():
+def get_entropy_estimate() -> int:
     with io.open('/proc/sys/kernel/random/entropy_avail') as f:
         return int(f.read())
 
-
 def asynchronous(f):
     def wrapper(*args, **kwargs):
         thread = Thread(target=f, args=args, kwargs=kwargs)
```
pierwill: Not sure how to annotate these functions...

securedrop/source_app/utils.py
```diff
@@ -32,7 +34,7 @@ def valid_codename(codename):
     return source is not None
 
 
-def generate_unique_codename(config):
+def generate_unique_codename(config: SDConfig) -> Optional[str]:
```
pierwill: On second look, thinking this should be `(config: SDConfig) -> str:`.

redshiftzero: yep `str` should be the return type here (the function will keep running until it finds a valid codename)

securedrop/source_app/utils.py
redshiftzero: `bool` is right since you'll either return a boolean _or_ it'll raise an exception (what Flask/Werkzeug `abort` is doing, which looks to be annotated properly as `typing.NoReturn` [here](https://github.com/python/typeshed/blob/master/third_party/2and3/werkzeug/exceptions.pyi#L184))

securedrop/source_app/utils.py
redshiftzero: that is the main object in `crypto_util.py` (`CryptoUtil`)

securedrop/source_app/utils.py
redshiftzero: there seems less value to adding static type annotations to this specific function so feel free to leave out, but if you want to go ahead:  `typing.Any` for args and kwargs seems reasonable here (since it's _such_ a generic function, if anyone disagrees here feel free to), and since this is a decorator (i.e. a function that wraps a function), type of `f` is `typing.Callable`

kushaldas:
> looks good!
> 
> one other thought: in other parts of the server codebase we have the imports needed only for type checking in a `if TYPE_CHECKING` block so they're not imported at runtime - @kushaldas (or others interested) want to keep doing that or no strong thoughts? I'll approve but not merge so folks have a chance to weigh in.

I prefer things under `if TYPE_CHECKING` if possible, but this is okay for me.

pierwill:
I'll take a look at how the conditional import is done elsewhere and add this in!

securedrop/source_app/utils.py
```diff
@@ -8,18 +8,25 @@
 from sqlalchemy.orm import sessionmaker
 from threading import Thread
 
+import typing
+
+if typing.TYPE_CHECKING:
```
kushaldas: Remove the import of `Optional` from under the if statement, that should fix the lint error.

pierwill: So, remove lines 13-14?

pierwill: Don't we need to import `Optional` somewhere here?

kushaldas: Sorry for the confusion, we have to keep the import statements, and maybe the best way is to mark `Optional` as strings in all of the usage lines, example is given below. 

```diff
diff --git a/securedrop/source_app/utils.py b/securedrop/source_app/utils.py
index e18766d00..fe01481b4 100644
--- a/securedrop/source_app/utils.py
+++ b/securedrop/source_app/utils.py
@@ -125,7 +125,7 @@ def normalize_timestamps(filesystem_id: str) -> None:
                 rc)
 
 
-def check_url_file(path: str, regexp: str) -> Optional[str]:
+def check_url_file(path: str, regexp: str) -> "Optional[str]":
     """
     Check that a file exists at the path given and contains a single line
     matching the regexp. Used for checking the source interface address
@@ -143,11 +143,11 @@ def check_url_file(path: str, regexp: str) -> Optional[str]:
         return None
 
 
-def get_sourcev2_url() -> Optional[str]:
+def get_sourcev2_url() -> "Optional[str]":
     return check_url_file("/var/lib/securedrop/source_v2_url",
                           r"^[a-z0-9]{16}\.onion$")
 
 
-def get_sourcev3_url() -> Optional[str]:
+def get_sourcev3_url() -> "Optional[str]":
     return check_url_file("/var/lib/securedrop/source_v3_url",
                           r"^[a-z0-9]{56}\.onion$")

```

eloquence:
@pierwill Looks like linter is still unhappy, would you like help resolving these linter issues?

pierwill:
That would be great! I haven't had time to troubleshoot this lately...

eloquence:
No worries, @kushaldas will poke at it a bit more tomorrow.

kushaldas:
This fixes the issue for me.

```diff
diff --git a/securedrop/source_app/utils.py b/securedrop/source_app/utils.py
index 3e430a81f..6c72a6de9 100644
--- a/securedrop/source_app/utils.py
+++ b/securedrop/source_app/utils.py
@@ -10,10 +10,6 @@ from threading import Thread
 
 import typing
 
-if typing.TYPE_CHECKING:
-    from typing import Optional  # noqa: F401
-
-
 import i18n
 import re
 
@@ -21,6 +17,9 @@ from crypto_util import CryptoUtil, CryptoException
 from models import Source
 from sdconfig import SDConfig
 
+if typing.TYPE_CHECKING:
+    from typing import Optional  # noqa: F401
+
 
 def logged_in() -> bool:
     return 'logged_in' in session


```

pierwill:
Squashed and linted!

PR #5316, opened 2020-06-16
Add SecureDrop 1.3.0->1.4.0 docs, OSSEC alert guidance

### LINKED ISSUE #5310: Chapter hierarchy issue in admin guide
Just noticing that #5301 introduced a small chapter hierarchy issue to https://docs.securedrop.org/en/latest/admin.html#passphrases-and-two-factor-resets (note how every section up to the FAQ is now a subsection of the Password reset/2FA guide).

eloquence:
## Status

Ready for review

## Description
- Adds our standard upgrade guidance, as well as specific docs pertaining to #5287. In the process, generalizes our docs for updating the system configuration a bit further, and adds standard guidance for ensuring you're running the correct version of the playbook
- Notes that Tails 4 upgrade docs may go away soon
- Resolves #5310

## Checklint

- [x] Linter says "Nothing to see here. Move along, folks"

### DIFF
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -422,10 +422,9 @@ There are two primary reasons why you may want to update the system configuratio
 
 - to change SecureDrop server configuration options. **Example:** You want to change
   the time of day at which the servers are automatically rebooted (default: 4:00 AM).
-- to fix a problem with the configuration state of your *Application* and/or
-  *Monitor Server*. **Example:** An administrator unfamiliar with the system has
-  directly modified the iptables rules, and you want to reinstate the correct
-  rules.
+- to restore a valid configuration state on your servers. **Example:** Another admin
+  has directly modified the iptables rules during troubleshooting, and you want
+  to reinstate the correct rules.
 
 In both cases, follow these steps:
 
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -194,9 +194,9 @@ Passphrases and Two-Factor Resets
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. warning:: Both of these operations will lock a user out of their
-   SecureDrop accounts. We recommend having users be physically present when
+   SecureDrop account. We recommend having users be physically present when
    resetting their passphrase or two-factor authentication. If this is not
-   possible, store passphrases and/or two-factor authentication secret in
+   possible, store the passphrase and/or two-factor authentication secret in
    your own password manager before securely transmitting them to the user
    in question, and delete them once the user has confirmed they can
    successfully log in.
@@ -423,8 +423,9 @@ There are two primary reasons why you may want to update the system configuratio
 - to change SecureDrop server configuration options. **Example:** You want to change
   the time of day at which the servers are automatically rebooted (default: 4:00 AM).
 - to fix a problem with the configuration state of your *Application* and/or
-  *Monitor Server*. **Example:** An administrator has locally modified your iptables
-  rules, and you want to reinstate the SecureDrop defaults.
+  *Monitor Server*. **Example:** An administrator unfamiliar with the system has
+  directly modified the iptables rules, and you want to reinstate the correct
+  rules.
 
 In both cases, follow these steps:
 
--- a/docs/admin.rst
+++ b/docs/admin.rst
@@ -191,14 +191,14 @@ possession when they attempt to log in to SecureDrop.
 
 
 Passphrases and Two-Factor Resets
----------------------------------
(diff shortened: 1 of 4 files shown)


docs/admin.rst
```diff
@@ -191,14 +191,14 @@ possession when they attempt to log in to SecureDrop.
...
----------------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. warning:: Both of these operations will lock a user out of their
    SecureDrop accounts. We recommend having users be physically present when
    resetting their passphrase or two-factor authentication. If this is not
-   possible, store passphrases and/or two-factor authentication secret in 
-   your own password manager before securely transmitting them to the user 
-   in question, and delete them once the user has confirmed they can 
+   possible, store passphrases and/or two-factor authentication secret in
```
rmol, line 199: I know this is just a whitespace change but it made me notice it should be "and/or two-factor authentication secret**s**" -- plural.

docs/admin.rst
```diff
@@ -418,33 +418,42 @@ Updating System Configuration
...
-If you want to update the system configuration, you should use the
-``securedrop-admin`` tool on the *Admin Workstation*. From
-``~/Persistent/securedrop``, run:
+There are two primary reasons why you may want to update the system configuration:
 
-.. code:: sh
+- to change SecureDrop server configuration options. **Example:** You want to change
+  the time of day at which the servers are automatically rebooted (default: 4:00 AM).
+- to fix a problem with the configuration state of your *Application* and/or
+  *Monitor Server*. **Example:** An administrator has locally modified your iptables
```
zenmonkeykstop, line 426: This example is a bit contrived, and might give admins the impression that modifications to iptables are something that they are expected to do. Something like `OSSEC SMTP settings are incorrect and need to be updated.` might make more sense

docs/upgrade/1.3.0_to_1.4.0.rst
```diff
@@ -0,0 +1,97 @@
...
+
+  rm ~/Persistent/.securedrop/securedrop_update.flag
+
+This will prevent the graphical updater from attempting to re-apply the failed
+update and has no bearing on future updates. You can now perform a manual
+update by  running the following commands: ::
+
+  cd ~/Persistent/securedrop
+  git fetch --tags
+  gpg --keyserver hkps://keys.openpgp.org --recv-key \
```
zenmonkeykstop, line 37: Would it be worth adding a note to say that the signing key is due to expire, so this is a mandatory step? I can see an  admin skipping it otherwise.

docs/admin.rst
eloquence: I'd like to distinguish between
- a configuration change that you or someone else made explicitly via `securedrop-admin sdconfig` vs.
- the system state that is enforced by the playbook.

It's IMO important to understand that running `securedrop-admin install` does more than just applying your site-specific configuration settings. And the reason I picked `iptables` as an example here is to make it easier for folks to recognize this section as relevant if they jump to it from the other chapter.

We could say something like "an administrator unfamiliar with the system" or other language to make it clear that such edits aren't desirable.

docs/upgrade/1.3.0_to_1.4.0.rst
eloquence: Sounds good, will do :)

docs/admin.rst
eloquence: good flag, tbh I think they should all be singular (common case is a single account), done in aa2d96968dc8da9d4556f9836273f2269ad95180

docs/admin.rst
eloquence: done in aa2d96968dc8da9d4556f9836273f2269ad95180

docs/upgrade/1.3.0_to_1.4.0.rst
eloquence: done in aa2d96968dc8da9d4556f9836273f2269ad95180

docs/admin.rst
zenmonkeykstop: Maybe instead of "to fix a problem..." the bulletpoint could just be "to restore the servers' configuration to a valid state. **Example:** to revert networking changes made for troubleshooting purposes."

eloquence: How about:
> to restore a valid configuration state on your servers. Example: Another admin has directly modified the iptables rules during troubleshooting, and you want to reinstate the correct rules.

zenmonkeykstop: sgtm

eloquence: done in 359a6194e0cbbaca6b3a03b0c223adb8f4a4c5ad

PR #5332, opened 2020-06-22
update pre-release docs - part 2

creviera:
## Status

Ready for review

## Description of Changes

Update RM docs to include details around the beginning steps of the pre-release process.

## Testing

Someone who's familiar with release management should make sure the changes are accurate and helpful.

Choose one of the following:

### DIFF
--- a/docs/development/release_management.rst
+++ b/docs/development/release_management.rst
@@ -39,10 +39,11 @@ Pre-Release
    <https://deb.tails.boum.org/dists/>`_. If so, request
    people participating in QA to use the latest release candidate.
 
-#. Ensure that a pre-release announcement is prepared and shared with the community
-   for feedback. Once the announcement is ready, coordinate with other team members to
-   send them to current administrators, post on the SecureDrop blog, and tweet
-   out a link.
+#. Work with the Communications Manager assigned for the release to prepare a pre-release 
+   announcement that will be shared on the support.freedom.press support portal, securedrop.org 
+   website, and Twitter. Wait until the day of the release before including an announcmement for a 
+   SecureDrop security update. For a point release, you may be able to skip the pre-release 
+   announcement depending on how small the point release is.
 
 #. Create a release branch.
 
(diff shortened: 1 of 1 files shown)


docs/development/release_management.rst
```diff
@@ -22,59 +22,48 @@ Pre-Release
...
+      numbers for each change. You will add these changes to the changelog in the next step.
 
-         git tag -d 1.x.y-rcN
-         git mktag < 1.x.y-rcN.tag > .git/refs/tags/1.x.y-rcN
+   b. Run ``update_version.sh``, passing it the new version in the format 
+      ``<major>.<minor>.<patch>~rcN`` (note how we use a tilde here instead of a dash). The script 
+      will open both the main repository changelog (``changelog.md``) and the one used for Debian 
+      packaging in an editor, giving you a chance to add the changes you collected. In the Debian 
+      changelog, we typically just refer the reader to the ``changelog.md`` file:
+      ``securedrop/bin/dev-shell ../update_version.sh 1.3.0~rc1``
```
conorsch, line 56: It's important that the update_version script is run with the `./securedrop/bin/dev-shell` prefix. The modified language here doesn't make that sufficiently clear. Consider embedding the full command inside the block, as done previously, rather than placing the example command at the very end.

docs/development/release_management.rst
```diff
@@ -22,59 +22,48 @@ Pre-Release
...
 
-#. Push the branch and tags:
+   c. Sign the commit added by the ``update_version.sh`` script:
+      ``git commit --all --amend --gpg-sign --no-edit``
```
conorsch, line 59: :rotating_light: Careful, that `--all` flag will take any local modifications in git repo and silently fold them into the previously created release commit. Remove the `--all` flag, then the command should do what we want: take the previously created commit, then sign it, without modifying the contents of the diff.

docs/development/release_management.rst
```diff
@@ -22,59 +22,48 @@ Pre-Release
...
+   d. Push the branch:
+      ``git push origin release/<major>.<minor>.<patch>``
 
-   * For subsequent release candidates and the final release version,
-     issue a PR with changelog and version changes into the
-     ``release/1.x.y`` branch, and push the signed tag once the PR is
-     merged.
+   e. Push the unsigned tag (you'll sign the <major>.<minor>.<patch> tag before the release goes 
+      out):
+      ``git push origin <major>.<minor>.<patch>-rcN``
```
conorsch, line 66: We seem to be a bit hand-wavy about the _why_ of tildes-versus-hyphens in the release candidate version strings. Consider a pop-out "Note" elsewhere in this doc to explain that Debian packages denote rcs with tildes, linking out to e.g. https://www.debian.org/doc/manuals/maint-guide/first.en.html#namever, but git tags don't support tildes, linking out to e.g. https://git-scm.com/docs/git-check-ref-format#_description

docs/development/release_management.rst
creviera: i see using `::` instead of backticks will show the command in a yellow box to make it stand out more... agreed this looks better!

docs/development/release_management.rst
creviera: removed `--noedit` as well just in case

creviera:
Comments addressed with a little extra reformatting and rewording.

creviera:
Running ci workflow since there was an error during the test sequence action: 'converge'

eloquence:
Kicking CI. Reminder @creviera, you can skip the long-running tests by prefixing the branch name with `docs-`

creviera:
oh yeah i keep forgetting. looks like the  'converge' error went away when you kicked it. this is ready for another review

PR #5335, opened 2020-06-23
Fix onion service version validation in securedrop-admin install

### LINKED ISSUE #5334: Install command fails on admin workstation if v2 services enabled, v3 services disabled.
## Description

If the `site-specific` file on an Admin Workstation has `v2_onion_services: true` and `v3_onion_services: false

## Steps to Reproduce

using an admin workstation USB:
- ensure that the workstation code has been upgraded to 1.4.0
- run `./securedrop-admin sdconfig` and set v2 services on, v3 services off
- run `./securedrop-admin install`

## Expected Behavior

- Install playbook is run, prompting user for server-side admin password

## Actual Behavior

- command fails with error `ERROR: Since you disabled v2 onion services, you must enable v3 onion services`

## Comments

This issue is 1.4.0-specific, 1.3.0-tagged version works correctly.

rmol:
## Status

Ready for review

## Description of Changes

Fixes #5334.

This was fallout from introducing non-interactive configuration validation during the install operation.

The `SiteConfig._config_in_progress` attribute was only being populated in `user_prompt_config`, so the v3 validator would never see v2 enabled during the install validation step, causing a spurious validation complaint if only v2 were enabled. This change ~copies the loaded config to `_config_in_progress` before calling `clean_config`~ changes `clean_config` to add each validated value to `_config_in_progress` (thanks @zenmonkeykstop), so validators can check the values of other configuration values.

## Testing

- Check out `develop`.
- Run `securedrop-admin sdconfig` to enable only v2 onion services.
- Run `securedrop-admin install`. It should abort with this error:
```
ERROR: Since you disabled v2 onion services, you must enable v3 onion services.
ERROR: Error loading configuration. Please run "securedrop-admin sdconfig" again.
ERROR (run with -v for more): Since you disabled v2 onion services, you must enable v3 onion services.
```
- Checkout the `fix-5334` branch.
- Rerun `securedrop-admin install`. You should not get the error, but be prompted for the sudo password.

### DIFF
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -424,6 +424,9 @@ class SiteConfig(object):
     def load_and_update_config(self, validate: bool = True, prompt: bool = True):
         if self.exists():
             self.config = self.load(validate)
+        elif not prompt:
+            sdlog.error('Please run "securedrop-admin sdconfig" first.')
+            sys.exit(1)
 
         return self.update_config(prompt)
 
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -637,6 +637,7 @@ class SiteConfig(object):
         try:
             with io.open(self.args.site_config) as site_config_file:
                 c = yaml.safe_load(site_config_file)
+                self._config_in_progress = c
                 return self.clean_config(c) if validate else c
         except IOError:
             sdlog.error("Config file missing, re-run with sdconfig")
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -624,6 +624,8 @@ class SiteConfig(object):
                         )
                         raise
                 clean_config[var] = transform(text) if transform else text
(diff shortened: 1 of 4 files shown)


admin/securedrop_admin/__init__.py
```diff
@@ -637,6 +637,7 @@ def load(self, validate=True):
         try:
             with io.open(self.args.site_config) as site_config_file:
                 c = yaml.safe_load(site_config_file)
+                self._config_in_progress = c
                 return self.clean_config(c) if validate else c
         except IOError:
             sdlog.error("Config file missing, re-run with sdconfig")
```
zenmonkeykstop, line 643: With the validation changes and `validate=true`, this exception is never thrown. A `KeyError: `securedrop_app_gpg_public_key'` is thrown instead, which is a little misleading.

zenmonkeykstop: This is unrelated  to the fix at hand tho and also already present in 1.4.0.

rmol: I've added a fix for this as well.

PR #5343, opened 2020-06-25
Clarify release roles, new package upload/signing workflow

eloquence:
## Status

Ready for review 

## Description of Changes

- Documents our rotating release management responsibilities.
- Documents the revised workflow for uploading and signing packages using the LFS repo.
- Includes trailing whitespace cleanup.

## Checklint
- [x] `make docs-lint` has examined this PR and reached a state of blissful transcendence

### DIFF
--- a/docs/development/release_management.rst
+++ b/docs/development/release_management.rst
@@ -260,6 +260,11 @@ Release Process
    ``apt.freedom.press``.
 #. The reviewer must delete the ``release`` branch so that it can be re-created
    during the next release.
+#. Issue a PR in the ``securedrop`` repository to merge the release branch
+   changes into ``master``. Once the PR is merged, verify that the
+   `public documentation <https://docs.securedrop.org/>`_
+   refers to the new release version. If not, log in to ReadTheDocs and start a
+   build of the ``master`` version.
 #. Create a `release
    <https://github.com/freedomofpress/securedrop/releases>`_ on GitHub
    with a brief summary of the changes in this release.
--- a/docs/development/release_management.rst
+++ b/docs/development/release_management.rst
@@ -17,6 +17,8 @@ roles for a SecureDrop release:
 - **Localization Manager:** to manage outreach to the translator community, and
   to coordinate translation updates of existing strings.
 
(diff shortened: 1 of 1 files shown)


docs/development/release_management.rst
```diff
@@ -8,23 +8,44 @@ must be done by people that have special privileges to do specific tasks
...
 
+In addition to the Release Manager, we typically recognize the following
+roles for a SecureDrop release:
+
+- **Deputy RM:** for additional time zone coverage, to delegate specific tasks,
+  and to act as backup in case of the RM becomes unavailable for any reason.
+
+- **Localization Manager:** to manage outreach to the translator community, and
+  to coordinate translation updates of existing strings.
+
```
rmol, line 19: We've also been assigning a deputy LM, for similar reasons as the deputy RM.

eloquence: (Added)

conorsch:
Holding off on merge so at least @creviera can chime in.

docs/development/release_management.rst
```diff
@@ -8,23 +8,46 @@ must be done by people that have special privileges to do specific tasks
...
+- **Deputy LM:** like the RM, this role is backed up by another team member.
+
+- **Communications Manager:** to prepare and distribute pre-release and
+  release messaging (including standard upgrade instructions, release notes,
+  social media posts, and support portal announcements)
+
+During the full release cycle, we also recognize the following role:
```
sssoleileraaa, line 26: Wondering if there should be a designated QA Manager who can give daily updates to the RM or deputy RM about bugs or issues the QA team is having, especially since there could be timezone differences where the RM might not be able to follow along as it happens?

eloquence: I've suggested that previously as well, but I think for now we should just capture current practices, and discuss this a bit more & potentially try it with 1.5.0. Does that make sense?

sssoleileraaa: that sounds good to me

sssoleileraaa:
@conorsch this is ready to go for launch

conorsch:
:rocket:

Issue #4225, opened 2019-03-05
friendly codenames


creviera:
## Description

Use curated word list for codenames/ journalist designations.

## User Stories

We have a feature to regenerate a codename for a source from the journalist interface. The user story is that a journalist wants to replace a codename that they find offensive. I'm not sure how often this is an issue, but using a curated list of words can help, such as: https://github.com/FogCreek/friendly-words

ninavizz:
So much love for allthings FogCreek!

eloquence:
For journalist designations, the existing dictionaries in `securedrop/dictionaries` have already been redacted, see:
https://github.com/freedomofpress/securedrop/commit/366c657b5f79b3dae510404ea5d6fed6adb0f6ca

It's unclear from the commit what methodology was used to do so.

PR #5357, opened 2020-07-01
Further redaction of wordlist

eloquence:
The main objective is to avoid combinations of words that could be offensive or distracting. Nouns that are inoffensive by themselves may become offensive in certain combinations.

The goal is not to entirely avoid humorous or mildly offensive combinations (that seems impossible given the size of the word list), so the mental test I applied is "could this be distracting if said out loud (in combination with other words)" or "is this inherently offensive of potentially offensive in combination". My sense is that this is an iterative process, and we should do at least one more pass before closing out #4225.

## Status

Ready for review

## Description of Changes

Towards #4225

## Testing

Inspect diff for overzealous removal or other words that should be removed

## Deployment

Should only impact newly generated sources, so no expected deployment impact

### DIFF
--- a/securedrop/dictionaries/adjectives.txt
+++ b/securedrop/dictionaries/adjectives.txt
@@ -785,7 +785,6 @@ big-bellied
 big-boned
 big-shouldered
 big-ticket
-bigoted
 bilateral
 biliary
 bilingual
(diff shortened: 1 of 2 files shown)


rmol (approved):
Yep. All right.

PR #5358, opened 2020-07-03
Fixes #5206 Delete Source Confirmation Modal

### LINKED ISSUE #5206: Clarify that deleting a source will prevent them from logging in
Currently the deletion modal in the Journalist Interface looks like this:

Screenshot from 2020-04-23 17-02-49

It is not clear from the language that deleting a source will prevent that source from logging in. That is made clear only if you delete a single source from its detail view that includes the list of files/messages/reply, where we show the following warning in the footer:

> Warning: If you do this, the files seen here will be unrecoverable and the source will no longer be able to login using their previous codename.

We should make this logic more apparent on the index page as well, as it is uncommon in most web apps for content deletion to also imply account deletion.

# User Story

As a journalist, I want to know what the effects of a deletion action are, so that any actions I take don't have unintended side effects.

danielpyon:
## Status

Ready for review

## Description of Changes

Fixes #5206.

Added `modal_warning to journalist_templates/_confirmation_modal.html` to display emphasized text of any warnings.

Changed `journalist_templates/index.html` to add warning text to `modal_warning`. (This text is identical to the one in `col.html`, line 104: "If you do this...")

## Deployment

We need to add the translation for the modal_warning, but this isn't very difficult because the text is already translated, just not specifically for this page.

### DIFF
--- a/securedrop/journalist_templates/_confirmation_modal.html
+++ b/securedrop/journalist_templates/_confirmation_modal.html
@@ -4,6 +4,9 @@
     <a href="#close" title="{{ gettext('Close') }}" class="close">X</a>
     <h2>{{ modal_data.modal_header }}</h2>
     <p>{{ modal_data.modal_body }}</p>
+    {% if modal_data.modal_warning %}
+      <p><em>{{ modal_data.modal_warning }}</em></p>
+    {% endif %}
     <a href="#close" id="{{ modal_data.cancel_id }}" title="{{ gettext('Cancel') }}" class="btn upper">{{ gettext('Cancel') }}</a>
     <button type="submit" id="{{ modal_data.submit_id }}" name="action" value="delete" class="{{ modal_data.submit_btn_type }} upper">{{ modal_data.submit_btn_text }}</button>
   </div>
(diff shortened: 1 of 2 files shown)


kushaldas (dismissed):
Approved. But I will wait for one look from @eloquence before merging. 

This is the what the UI looks like:

delete_ui_change

eloquence:
Thanks @danielpyon for your work on this.

It's a good idea to re-use the string to minimize translation impact. However, in this case I would recommend against it, because the context is different. The string originally occurs at the bottom of a source collection (where you look at the list of files belonging to a source). The text "the files seen here" makes sense in that context. It doesn't work as well on the list of sources, because 1) you're not looking at a list of files, but a list of sources, 2) not all the sources will be deleted, but only the selected ones.

I'd suggest this wording:
> Warning: If you do this, all files for the selected sources will be unrecoverable, and the sources will no longer be able to log in using their previous codename.

Feel free to tweak or suggest alternatives. :)

danielpyon:
I see, I didn't realize that the message was specific to the source collection. I'll fix it in a little bit.

kushaldas (approved):
Approved.

PR #5366, opened 2020-07-10
add v2 onion service deprecation warning

sssoleileraaa:
## Status

Ready for review

## Description of Changes

Fixes https://github.com/freedomofpress/securedrop/issues/5302

* show the following eol warning for v2 onion services if the `source_v3_url` file is missing:

```
Update Required: Your SecureDrop servers are still running v2 Onion Services. In February 2021, v2 Onion Services will be disabled, and your SecureDrop servers may become unreachable.
```

* show the following eol warning for v2 onion services if the `source_v2_url` file exists along with the `source_v3_url`:

```
Update Required: Your SecureDrop servers are still running v2+v3 Onion Services. In February 2021, v2 Onion Services will be disabled, and your SecureDrop servers may become unreachable.
```

## Feedback request

Would like review on language here. This is a WIP until we set up the `More info` redirects. Or perhaps we should create a separate information page in the JI? 

## Test Plan

1. Make sure no `source_v2_url` or `source_v3_url` files exist in `/var/lib/securedrop` 
2. Log into the JI in a Tor Browser
- [ ] Verify that the correct v2 warning appears
3. Add  a `source_v3_url` file in `/var/lib/securedrop` 
4. Log into the JI in a Tor Browser
- [ ] Verify that no warning appears
5. Add `source_v2_url` files in `/var/lib/securedrop` 
6. Log into the JI in a Tor Browser
- [ ] Verify that the correct v2+v3 warning appears

PR #5375, opened 2020-07-15
Fix sentence about onion services v2

comradekingu:
## Status

Ready for review

## Description of Changes

Fixes #.

https://weblate.securedrop.org/translate/securedrop/securedrop/en/?checksum=8bf29e52aaf6db39#comments

## Testing

See https://blog.torproject.org/v2-deprecation-timeline check whether SecureDrop deprecation differs from Tor plan.

## Deployment

2. New installs.

Choose one of the following:

### DIFF
--- a/securedrop/journalist_templates/base.html
+++ b/securedrop/journalist_templates/base.html
@@ -21,7 +21,7 @@
     {% if g.user %}
       {% if g.show_v2_onion_eol_warning %}
       <div id="v2-onion-eol" class="warning-banner">
-        {{ gettext('<strong>Update Required:</strong> Your SecureDrop servers are still running v2 Onion Services, which are being phased out for security reasons. In February 2021, v2 Onion Services will be disabled, and your SecureDrop servers may become unreachable. <a href="//securedrop.org/v2-onion-eol" rel="noreferrer">Learn More</a>') }}
+        {{ gettext('<strong>Update Required:</strong>Your SecureDrop servers are running onion services v2, which will be deprecated February 2021. Your SecureDrop servers may become unreachable. <a href="//securedrop.org/v2-onion-eol" rel="noreferrer">Learn More</a>') }}
       </div>
       {% endif %}
 
--- a/securedrop/journalist_templates/base.html
+++ b/securedrop/journalist_templates/base.html
@@ -21,7 +21,7 @@
     {% if g.user %}
       {% if g.show_v2_onion_eol_warning %}
       <div id="v2-onion-eol" class="warning-banner">
-        {{ gettext('<strong>Update Required:</strong>Your SecureDrop servers are running onion services v2, which will be deprecated February 2021. Your SecureDrop servers may become unreachable. <a href="//securedrop.org/v2-onion-eol" rel="noreferrer">Learn More</a>') }}
+        {{ gettext('<strong>Update Required:</strong>Your SecureDrop servers are running v2 onion services, which will be deprecated for security reasons come February 2021. Your SecureDrop servers may become unreachable. <a href="//securedrop.org/v2-onion-eol" rel="noreferrer">Learn More</a>') }}
       </div>
       {% endif %}
 


zenmonkeykstop:
Hi @comradekingu ! The February 2021 date refers to the plan for retirement of v2 onion service support from SecureDrop rather than from Tor in general. The point about capitalization is well-taken though. (SD documentation is very inconsistent in this regard so a larger fix is probably called for.)

eloquence:
Hi @comradekingu, thanks for this PR. As Kev notes, we're happy to standardize on the lower case "onion services", but would prefer to keep the wording otherwise as it is. 

As Kev noted above, SecureDrop's deprecation period of v2 onion services begins with the SecureDrop 1.5.0 release (July 28, 2020) and ends with the SecureDrop 2.0.0 release (February 2021), when support for v2 will be removed entirely. Tor's [deprecation timeline](https://blog.torproject.org/v2-deprecation-timeline) begins in September 2020, with support for v2 scheduled for removal in July 2021. While the term is often used imprecisely, _deprecation_ means that the use of a feature is discouraged, typically followed by the complete _removal_ of said feature.

comradekingu:
From my understanding the v2 comes after the "onion service". Is "for security reasons" important?
Feel free to change things around.

eloquence:
> From my understanding the v2 comes after the "onion service".

Both are valid in English, Tor (and the Internet at large) has generally preferred "v[2-3] onion services" except for the most recent blog post, see e.g.:
https://twitter.com/search?q=from%3Atorproject%20v3&src=typed_query&f=live

> Is "for security reasons" important?

Yes, during development we iterated on this language a couple of times, and agreed we wanted to bring attention to this aspect -- this is just not something that's "nice to have", but important for source/journalist security.

securedrop/journalist_templates/base.html
```diff
@@ -21,7 +21,7 @@
     {% if g.user %}
       {% if g.show_v2_onion_eol_warning %}
       <div id="v2-onion-eol" class="warning-banner">
-        {{ gettext('<strong>Update Required:</strong> Your SecureDrop servers are still running v2 Onion Services, which are being phased out for security reasons. In February 2021, v2 Onion Services will be disabled, and your SecureDrop servers may become unreachable. <a href="//securedrop.org/v2-onion-eol" rel="noreferrer">Learn More</a>') }}
+        {{ gettext('<strong>Update Required:</strong>Your SecureDrop servers are running onion services v2, which will be deprecated February 2021. Your SecureDrop servers may become unreachable. <a href="//securedrop.org/v2-onion-eol" rel="noreferrer">Learn More</a>') }}
```
comradekingu:
```suggestion
        {{ gettext('<strong>Update Required:</strong>Your SecureDrop servers are running v2 onion services, which will be deprecated for security reasons come February 2021. Your SecureDrop servers may become unreachable. <a href="//securedrop.org/v2-onion-eol" rel="noreferrer">Learn More</a>') }}
```

eloquence: That is still not an accurate use of the term "deprecated". Unless you feel really strongly about the phrase "phased out" or some other aspect of the original string, I would suggest we close this PR @comradekingu, I can open a separate one for changing the spelling of `onion services` to the lower case variant (which includes uses in the documentation), except for headlines.

comradekingu: How so? Thought it was better. "Taken out of service" isn't true, because they will still work upstream.
https://en.wikipedia.org/wiki/Deprecation
https://www.collinsdictionary.com/dictionary/english/deprecate

I only feel strongly about the quality of the original string, and how I don't want to see it again.
If the other possible meaning of "phased out" is left without the qualifying date, one would not have to reiterate how it will stop working in a new sentence. (?)

"Your SecureDrop servers may become unreachable when v2 onion services are deprecated for security reasons come February 2021." is a better sentence in my head.

Reading a lengthy passage to try to best translate is more taxing than trying to understand it, and a recipe for translations made with no oversight, by different people, or with the use of direct translations. SD is exactly that.

Edit: "Upgrade your SecureDrop servers to more secure v3 onion services so you can reach them after February 2021." is what I would go with.

eloquence: See my explanation here: https://github.com/freedomofpress/securedrop/pull/5375#issuecomment-658921716

The term "deprecation" is used for the period during which we recommend disabling, not for the removal itself. It's the latter (removal) which will occur in February 2021. This is also well-explained in the Wikipedia article you link to.

eloquence: > Edit: "Upgrade your SecureDrop servers to more secure v3 onion services so you can reach them after February 2021." is what I would go with.

This language won't work with the current implementation, because the same warning is displayed regardless of whether v3 services are enabled or not, _if_ v2 services are enabled, and your shorter language does not cover both cases. See the logic here: https://github.com/freedomofpress/securedrop/pull/5366/files#diff-4c2318c651fa1f217d6b6ed48e77e201R176

I do think it's worth considering a shorter warning if folks don't feel it's too late for that, so thanks for the suggestions -- will discuss at our team standup tomorrow. For now I've opened #5379 to standardize `onion service` usage throughout the codebase.

comradekingu: The idea that it could do both wasn't all that clear.

>Your SecureDrop server still supports deprecated v2 onion services, Make sure it has v3 ones so that it will be reachable come February 2021.

or

>Make sure your SecureDrop server will be reachable February 2021 by making sure it supports the safer v3 onion services, and not just v2 which will stop working.

eloquence:
We discussed this today at SecureDrop standup and decided to go with the slightly more verbose variant that's currently in `develop`. This string will only be seen by journalists, not sources, and we want to err on the side of being explicit about the impact. However, we're on board with standardizing to the lower case `onion services` variant as you suggest, and will do so separately. 

Thanks for your input here @comradekingu, and for advocating for brevity in UI messaging - a concern I generally share.

PR #5381, opened 2020-07-16
Ensure that generated journalist designation doesn't conflict with existing ones

### LINKED ISSUE #2043: db.Source.journalist_designation is not a unique field

prateekj117:
## Status

Ready for review

## Description of Changes

Fixes #2043 

Added a while loop in crypto_util.display_id() similar to the loop in [source_app.utils.generate_unique_codename()](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/source_app/utils.py#L38) to make sure we keep on generating a new `journalist_designation` until we find one not used already. 

## Deployment

2. New installs.

No.

### DIFF
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -176,23 +176,19 @@ class CryptoUtil:
     def display_id(self):
         """Generate random journalist_designation until we get an unused one"""
 
-        used_journalist_designation_hits = 0
-        matching_journalist_designation = '123'
-        journalist_designation = ''
+        tries = 0
 
-        while len(matching_journalist_designation) != 0:
-            if used_journalist_designation_hits >= 50:
-                raise ValueError("Too many duplicate journalist_designation hits")
+        while tries < 50:
+            new_designation = ' '.join([random.choice(self.adjectives),
+                                        random.choice(self.nouns)])
 
-            journalist_designation = ' '.join([random.choice(self.adjectives),
-                                               random.choice(self.nouns)])
+            collisions = Source.query.filter(Source.journalist_designation == new_designation)
+            if collisions.count() == 0:
+                return new_designation
 
-            matching_journalist_designation = Source.query.filter(
-                Source.journalist_designation == journalist_designation).all()
+            tries += 1
 
-            used_journalist_designation_hits = used_journalist_designation_hits + 1
-
-        return journalist_designation
+        raise ValueError("Could not generate unique journalist designation for new source")
 
     def hash_codename(self, codename, salt=None):
         """Salts and hashes a codename using scrypt.
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -16,6 +16,8 @@ from redis import Redis
 
 import rm
 
+from models import Source
+
 import typing
 # https://www.python.org/dev/peps/pep-0484/#runtime-or-type-checking
 if typing.TYPE_CHECKING:
@@ -172,8 +174,16 @@ class CryptoUtil:
                         for x in range(words_in_random_id))
 
     def display_id(self):
-        return ' '.join([random.choice(self.adjectives),
-                         random.choice(self.nouns)])
+        """Generate random journalist_designation until we get an unused one"""
+        while True:
+            journalist_designation = ' '.join([random.choice(self.adjectives),
+                                               random.choice(self.nouns)])
+
+            matching_journalist_designation = Source.query.filter(
+                Source.journalist_designation == journalist_designation).all()
+
+            if len(matching_journalist_designation) == 0:
+                return journalist_designation
 
     def hash_codename(self, codename, salt=None):
         """Salts and hashes a codename using scrypt.
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -175,15 +175,24 @@ class CryptoUtil:
 
     def display_id(self):
         """Generate random journalist_designation until we get an unused one"""
-        while True:
+
+        used_journalist_designation_hits = 0
+        matching_journalist_designation = '123'
+        journalist_designation = ''
+
+        while len(matching_journalist_designation) != 0:
+            if used_journalist_designation_hits >= 50:
+                raise ValueError("Too many duplicate journalist_designation hits")
+
             journalist_designation = ' '.join([random.choice(self.adjectives),
                                                random.choice(self.nouns)])
 
             matching_journalist_designation = Source.query.filter(
                 Source.journalist_designation == journalist_designation).all()
 
-            if len(matching_journalist_designation) == 0:
-                return journalist_designation
+            used_journalist_designation_hits = used_journalist_designation_hits + 1
+
+        return journalist_designation
 
     def hash_codename(self, codename, salt=None):
         """Salts and hashes a codename using scrypt.
--- a/securedrop/source_app/main.py
+++ b/securedrop/source_app/main.py
@@ -77,7 +77,7 @@ def make_blueprint(config):
             except ValueError as e:
                 current_app.logger.error(e)
                 flash(
-                    gettext("There was a temporary problem creating you account. "
+                    gettext("There was a temporary problem creating your account. "
                             "Please try again."
                             ),
                     'error'
--- a/securedrop/source_app/main.py
+++ b/securedrop/source_app/main.py
@@ -72,8 +72,18 @@ def make_blueprint(config):
             del session['codenames']
 
             filesystem_id = current_app.crypto_util.hash_codename(codename)
+            try:
+                source = Source(filesystem_id, current_app.crypto_util.display_id())
+            except ValueError as e:
+                current_app.logger.error(e)
+                flash(
+                    gettext("There was a temporary problem creating you account. "
+                            "Please try again."
+                            ),
+                    'error'
+                )
+                return redirect(url_for('.index'))
 
-            source = Source(filesystem_id, current_app.crypto_util.display_id())
             db.session.add(source)
             try:
                 db.session.commit()
--- a/securedrop/tests/functional/functional_test.py
+++ b/securedrop/tests/functional/functional_test.py
@@ -76,6 +76,8 @@ class FunctionalTest(object):
     driver_retry_count = 3
     driver_retry_interval = 5
 
+    journalist_designation_collision_test = False
+
     def _unused_port(self):
         s = socket.socket()
         s.bind(("127.0.0.1", 0))
@@ -260,6 +262,10 @@ class FunctionalTest(object):
                     }
 
                     self.admin_user["totp"] = pyotp.TOTP(self.admin_user["secret"])
+
+                    if self.journalist_designation_collision_test:
+                        self.source_app.crypto_util.adjectives = self.source_app.crypto_util.adjectives[:1]
+                        self.source_app.crypto_util.nouns = self.source_app.crypto_util.nouns[:1]
 
                     def start_source_server(app):
                         config.SESSION_EXPIRATION_MINUTES = self.session_expiration / 60.0
--- a/securedrop/tests/functional/functional_test.py
+++ b/securedrop/tests/functional/functional_test.py
@@ -76,8 +76,6 @@ class FunctionalTest(object):
     driver_retry_count = 3
     driver_retry_interval = 5
 
-    journalist_designation_collision_test = False
-
     def _unused_port(self):
         s = socket.socket()
         s.bind(("127.0.0.1", 0))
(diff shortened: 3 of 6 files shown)


prateekj117:
@emkll @rmol Ready for review.

securedrop/crypto_util.py
```diff
@@ -172,8 +174,16 @@ def genrandomid(self,
                         for x in range(words_in_random_id))
 
     def display_id(self):
-        return ' '.join([random.choice(self.adjectives),
-                         random.choice(self.nouns)])
+        """Generate random journalist_designation until we get an unused one"""
```
zenmonkeykstop, line 177: Is there a security issue with changing this from a constant-time operation to a loop that might leak information about the number of sources in the database? (given the likely size of SD databases, his is more theoretical, but it might be better to have the source generation just fail altogether and prompt the user to restart the process.)

prateekj117: @zenmonkeykstop Not sure how this would reveal the number of sources in the database.

rmol: It shouldn't take too many iterations to avoid a collision, so probably wouldn't give a good idea of the number of sources. As @emkll pointed out, the unique index on `journalist_designation` would help decrease the time this check takes, too, but as [discussed in the issue](https://github.com/freedomofpress/securedrop/issues/2043#issuecomment-658991658), that's going to be a more involved undertaking which we'll tackle later.

It would be good in any case to have an upper bound on the number of attempts. I think 50 would be quick enough while allowing for some really bad luck. If that limit is hit, we can flash an error and ask the visitor to retry.

prateekj117: @rmol Sounds good. I will add an upper limit to it. Should it be done in the same PR?

eloquence: Yes, it's best to do this as part of the same PR Prateek. Thanks again for working on this. :)

eloquence: @prateekj117 Just touching base, do you still have time to poke at this / would you like help to move it over the finish line?

prateekj117: Hey @eloquence, so sorry, had quite some things to wrap up this week. I will complete it within this week for sure.

prateekj117: @eloquence What exception should I raise when `display_id` hits duplicate `journalist_designation` 50 times.

prateekj117: Will just raise a `ValueError` for now.

rmol: @prateekj117 Simply raising `ValueError` will work. If you have time, this is a case where I think friendlier error handling than we're consistently providing in the source interface would help not scare away a new source, e.g.: 

```diff
diff --git i/securedrop/source_app/main.py w/securedrop/source_app/main.py
index 172849fbc..fa5f67e73 100644
--- i/securedrop/source_app/main.py
+++ w/securedrop/source_app/main.py
@@ -73,7 +73,19 @@ def make_blueprint(config):
 
             filesystem_id = current_app.crypto_util.hash_codename(codename)
 
-            source = Source(filesystem_id, current_app.crypto_util.display_id())
+            try:
+                source = Source(filesystem_id, current_app.crypto_util.display_id())
+            except ValueError as e:
+                current_app.logger.error(e)
+                flash(
+                    gettext(
+                        "There was a temporary problem creating your account. "
+                        "Please try again."
+                    ),
+                    "error"
+                )
+                return redirect(url_for('.index'))
+
             db.session.add(source)
             try:
                 db.session.commit()
```

But again, I think it's unlikely that the limit will be hit, and we could stand to make some general improvements here, so this could also be deferred.

prateekj117: @rmol Yes, I agree. Something like this would be much better. Will use this only.

prateekj117:
@eloquence @rmol Ready for another review.

securedrop/source_app/main.py
```diff
@@ -72,8 +72,18 @@ def create():
             del session['codenames']
 
             filesystem_id = current_app.crypto_util.hash_codename(codename)
+            try:
+                source = Source(filesystem_id, current_app.crypto_util.display_id())
+            except ValueError as e:
+                current_app.logger.error(e)
+                flash(
+                    gettext("There was a temporary problem creating you account. "
```
eloquence: typo: you->your

prateekj117: Oops. Thanks for pointing that out. Corrected it.

securedrop/crypto_util.py
```diff
@@ -172,8 +174,25 @@ def genrandomid(self,
...
+
+        used_journalist_designation_hits = 0
```
rmol: Definitely the right idea. I would just suggest that the code could be shortened a bit, and more importantly, the content of the `ValueError` could supply more context. SecureDrop admins might be seeing this error in an OSSEC alert, or while checking the Apache logs, and it's helpful for them and our support team to have a clue where the error came from.

```python
    def display_id(self):
        """Generate random journalist_designation until we get an unused one"""

        tries = 0
        while tries < 50:
            new_designation = ' '.join([random.choice(self.adjectives), random.choice(self.nouns)])

            collisions = Source.query.filter(Source.journalist_designation == new_designation)
            if collisions.count() == 0:
                return new_designation

            tries += 1

        raise ValueError("Could not generate unique journalist designation for new source")
```

rmol (changes requested):
This is just about there. Unfortunately you've been contributing so much that it's time for us to start asking you for _less_ code. :wink: I've got suggestions for shortening this, and also wanted to point out how error messages make their way through logs to SecureDrop admins, and why it's important that they carry some context, even if they're not what we show to sources or journalists.

Thanks again for your contributions. We really do appreciate them, though it might not seem like it from the feedback loop. :slightly_smiling_face:

securedrop/crypto_util.py
prateekj117: @rmol Ohh yes. This looks much more clear.

prateekj117:
@rmol It's good to know that my contributions are valued. And with all the good work SecureDrop does, it is a pleasure to contribute here. Though I would appreciate it if there would be fewer code suggestions as that would make me learn some more things along the way for sure.

Also, it's good that you peeps help me understand the underlying structure and where things might go wrong. I would be more than happy to take over more responsibilities in Freedom of Press or specifically SecureDrop. I am also open to Internships or maybe other roles if there are any in Freedom of Press or SecureDrop.

prateekj117:
@rmol @eloquence Ready for another review.

rmol:
OK, one last thing, which I should have asked about earlier: tests. We just need to ensure `display_id` actually catches a collision. I'd suggest a test function in `test_crypto_util.py` that alters `source_app.crypto_util.adjectives` and `source_app.crypto_util.nouns` to each be a list of one word, so after you create a source, any invocation of `display_id` will raise the `ValueError`.

eloquence:
LGTM.

- [x] Repeatedly creating new sources works as before, with a unique journalist designation per source. Logging in as any of those sources works as before.
- [x] After patching `crypto_util.py` to always return the journalist designation of an existing source, it errors out as expected, and `tries` is incremented to 50 as expected. The error message is displayed to the user as expected, and they are redirected to the index as expected.
Screen Shot 2020-08-18 at 11 08 53-fullpage

Note: This case is never expected to be seen by a user, but it's good to have it handled.

prateekj117:
@rmol Cool. Will add the tests.

prateekj117:
@rmol @eloquence Should I also add selenium frontend test to check that flash text is rendered properly?

kushaldas:
> @rmol @eloquence Should I also add selenium frontend test to check that flash text is rendered properly?

If there is any good way to do that, yes.

securedrop/tests/functional/source_navigation_steps.py
```diff
@@ -123,6 +123,25 @@ def submit_page_loaded():
 
         self.wait_for(submit_page_loaded)
 
+    def _source_chooses_to_submit_documents_with_colliding_journalist_designation(self):
+        self.source_app.crypto_util.adjectives = self.source_app.crypto_util.adjectives[0]
```
prateekj117: @kushaldas @rmol Using this does not affect the original list.

Directly importing and using source_app module shows:
```
    def _source_chooses_to_submit_documents_with_colliding_journalist_designation(self):
>       source_app.crypto_util.adjectives = source_app.crypto_util.adjectives[0]
E       AttributeError: module 'source_app' has no attribute 'crypto_util'
```

rmol: The functional tests can be tricky. To see the error, the source interface server started by the functional test has to have the wordlists truncated. That's going to require changing `tests.functional.functional_test.FunctionalTest` to support customizing how the source app gets created in that server process, and then this test is going to need to use a subclass of `FunctionalTest` where that new method is used to abbreviate the wordlists in the source app's `CryptoUtil`. _Then_ the test should encounter a collision and see the error message.

If that much test work seems like the tail wagging the dog for this relatively straightforward change, I empathize, and would be happy to implement it so we can get your work merged.

prateekj117: @rmol Oh ok. I get the point. The change does seem straightforward. Plus this a condition that might not ever be faced by anyone. Also, if the `flash` does fail, it will fail at many other places, which I guess would already have functional tests for this. But my views can be biased given that I have to implement this. Haha. So you peeps decide and let me know if a functional test for this is required are not.

rmol: I think adding the test is a good idea, to make sure the error message is correctly shown upon collision. But you absolutely don't have to be the one adding it. We can take care of that.

prateekj117: @rmol If it has to be added, I will give it a try once, and ping you peeps if I face any issue.

prateekj117:
@kushaldas @rmol Done. Ready for another review.

securedrop/tests/functional/functional_test.py
```diff
@@ -261,6 +263,11 @@ def sd_servers(self):
 
                     self.admin_user["totp"] = pyotp.TOTP(self.admin_user["secret"])
 
+                    if self.journalist_designation_collision_test:
```
rmol: We shouldn't introduce test-specific customization in this class. Instead, you could create the source app in a new method, and override that method in a derived class to limit the nouns and adjectives. Then `TestSourceInterfaceDesignationCollision` would use that subclass instead of the base `FunctionalTest`.

securedrop/tests/functional/source_navigation_steps.py
```diff
@@ -123,6 +123,22 @@ def submit_page_loaded():
 
         self.wait_for(submit_page_loaded)
 
+    def _source_chooses_to_submit_documents_with_colliding_journalist_designation(self):
```
rmol: If you use the existing convenience methods like `self._source_chooses_to_submit_documents()` in `TestSourceInterfaceDesignationCollision`, instead of adding this, we won't have another place that needs updating when page content changes.

rmol (changes requested):

prateekj117:
Cool. Will make these changes.

prateekj117:
@rmol Did the suggested changes. Ready for another review.

securedrop/tests/functional/functional_test.py
```diff
@@ -185,6 +185,11 @@ def disable_js_torbrowser_driver(self):
         if hasattr(self, 'torbrowser_driver'):
             disable_js(self.torbrowser_driver)
 
+    def start_source_server(self, app, source_port):
```
rmol: Where you're using this, `self.source_app` has already been set up, so you could just use that and skip passing it as an argument. That would also make the overridden method in `TestSourceInterfaceDesignationCollision` more consistent.

securedrop/tests/functional/test_source.py
```diff
@@ -1,5 +1,28 @@
 from . import source_navigation_steps, journalist_navigation_steps
 from . import functional_test
+from sdconfig import config
+
+
+class TestSourceInterfaceDesignationCollision(
+        functional_test.FunctionalTest,
+        source_navigation_steps.SourceNavigationStepsMixin):
+
+    def start_source_server(self, app, source_port):
```
rmol: See the note above about just using `self.source_app` instead of passing `app`.

rmol (changes requested):
This is also going to need a rebase to pick up recent fixes for container/CI breakage.

prateekj117:
@rmol Did the changes and rebased.

rmol (approved):
done. thanks for all the hard work @prateekj117 !

PR #5391, opened 2020-07-20
String fixes based on translator feedback

eloquence:
## Status

Ready for initial comments.

Will squash prior to promoting from draft.

Note: This PR is targeting 1.6.0, not 1.5.0.


securedrop/source_templates/why-journalist-key.html
```diff
@@ -14,10 +14,10 @@ <h1>{{ gettext("Why download the team's public key?") }}</h1>
 <li>{{ gettext('Encrypt your submission.') }}
 <ol>
 <li>{{ gettext('You will need to be able to identify the key (this is called the "user ID" or UID). Since the public key\'s filename is the key\'s fingerprint (with .asc at the end), you can just copy and paste that. (don\'t include the <code>.asc</code>!)')|safe }}</li>
-<li>{{ gettext('On all systems, open the Terminal and use this gpg command: <code>gpg --recipient &lt;user ID&gt; --encrypt roswell_photos.pdf</code>')|safe }}</li>
+<li>{{ gettext('On all systems, open the terminal and use this gpg command, replacing <code>&lt;user ID&gt;</code> and <code>&lt;filename&gt;</code> with their values:<br><code>gpg --recipient &lt;user ID&gt; --encrypt &lt;filename&gt;</code>')|safe }}</li>
```
rmol: Eliminating the Roswell references is good, but I wonder if we want to avoid greater-than and less-than symbols in example commands, to avoid shell redirection problems if they're blindly copied and pasted. It was like this before, but we could take the opportunity to quote the example user ID and filename instead.

rmol (dismissed):
These changes all look good to me. I have one optional suggestion.

securedrop/source_templates/why-journalist-key.html
eloquence: This is scope creeping a little bit, but what do you think about the approach in f836393439451bd144d46068c1645cd3aa564d94 of just pulling in the fingerprint so we don't have to use roundabout ways to describe it?

rmol: That's better. I added the actual filename to the import command too, and tweaked the formatting of the commands.

eloquence: A bit of whitespace does wonders sometimes. Thanks @rmol, that looks good to me. I'll close this draft, squash, and open a new PR from a new branch name, since that change goes a bit beyond minor fixes at this point.

PR #5404, opened 2020-07-23
Remove unused/disabled source onion service info files

### LINKED ISSUE #5400: [1.5.0] [QA] v2 onion services files are not removed during v3 migration, so deprecation warning still shows
## Description

During QA, I experienced the following: When migrating to v3 onion services, v2 onion services (`ths`) files were not removed from `install_files/ansible_base` on the Admin workstation, and the `/var/lib/securedrop/source_v2_url` file was still present on the app server (though the v2 url was no longer working).  This means that the v2 deprecation warning, which checks for the presence of the latter file, will still be shown despite v2 onion services being disabled.

See https://github.com/freedomofpress/securedrop/pull/5366/files#diff-4c2318c651fa1f217d6b6ed48e77e201R176 

## Steps to Reproduce

1. Clean install 1.5.0-rc2 with v2+v3 onion services enabled. Confirm presence of deprecation warning. 
2. `sdconfig` to disable v2.
3. `install`, `tailsconfig`
4. Observe deprecation warning and aforementioned files.

## Expected Behavior

No warning is shown

## Actual Behavior

Deprecation warning still present

rmol:
## Status

Ready for review

## Description of Changes

Fixes #5400.

If a site disables v2 onion services, remove the file used to decide to show the deprecation warning.

The v3 equivalent is also removed if v3 services are disabled, though downgrading to v2-only is not really possible -- more work would be needed in the playbooks, and that's probably not something we should support at this point.

## Testing

- Start with a deployment that has v2 onion services enabled. VM or hardware is most straightforward at this point, but I'll see if it can be tested in a staging environment.
- Run `securedrop-admin sdconfig` to enable v3 and disable v2 onion services.
- Run `securedrop-admin install`. This should complete successfully, and in the Ansible output you should see `Remove source v2 onion service info if not enabled` followed by `Expose source v3 onion service info to app`.
- Log in to the app server and list the contents of `/var/lib/securedrop`. There should  be no `source_v2_url`, just `source_v3_url`. 
- Run `securedrop-admin tailsconfig`. 
- Log in to the journalist interface. There should be no v2 deprecation warning.

## Deployment

This enables admins to disable v2 onion services and remove the deprecation warning from the journalist interface.

### DIFF
--- a/install_files/ansible-base/roles/app/tasks/copy_tor_url_info_to_app_dir.yml
+++ b/install_files/ansible-base/roles/app/tasks/copy_tor_url_info_to_app_dir.yml
@@ -15,7 +15,7 @@
   failed_when: false
   when: v3_onion_services
 
-- name: Expose Tor v2 Onion URL info to app
+- name: Expose source v2 onion service info to app
   copy:
     dest: /var/lib/securedrop/source_v2_url
     owner: www-data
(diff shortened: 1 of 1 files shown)


zenmonkeykstop (approved):
Test plan passes, LGTM!

PR #5414, opened 2020-07-27
Add upgrade documentation for 1.4.1->1.5.0

rocodes:
- Split Tails 4 upgrade docs into separate file to reference without referring to old upgrade documentation.
- Update references to Tails 4 docs in past version upgrade guides (text does not change).

## Status

Ready for Review

## Description of Changes

- Add upgrade documentation for 1.4.1->1.5.0
- Split out Tails 3-> Tails 4 migration docs so that we can refer to the migration process without linking to text embedded in an old upgrade guide. (Note that the actual text/content of old upgrade guides will not be altered as a result of this PR.)

## Testing

Manual review  

## Deployment

n/a


rocodes:
@eloquence: Have split out the Tails 3-> Tails 4 migration documentation as we discussed. The downside however is that it is also included in the Topic Guide, which may be redundant. 
Some possibilities for getting around this without annoying Sphinx are: add it to the "excludes" directive, or maybe put it in a subfolder of a greater depth than 2 (so that the `toctree` with `maxDepth 2` for the upgrade documentation doesn't complain about files that are not in the index).  Both of these seemed kind of hacky so I left it as is but welcome suggestions.

docs/upgrade_to_tails_4.rst
```diff
@@ -0,0 +1,59 @@
+Upgrading workstations to Tails 4
+---------------------------------
+
+.. important::
+
+   Before upgrading your *Admin Workstation* and your *Journalist Workstation*
+   to Tails 4, you must first ensure that the version of the SecureDrop code on
+   the workstation (which is used for administrative tasks and for configuring
+   the Tails desktop) is at 1.1.0 or a later version.
```
eloquence: This would be a good opportunity to switch to `|version|` for all references to 1.1.0 in this chapter, and to instruct admins to simply ensure that they're using the latest version. (1.1.0 is most likely fine, but there's no particular reason to keep carrying forward a reference to such an old release for this upgrade procedure.)

docs/upgrade/1.0.0_to_1.1.0.rst
```diff
@@ -60,109 +60,8 @@ Finally, run the following commands: ::
   ./securedrop-admin setup
   ./securedrop-admin tailsconfig
 
-.. _upgrade_to_tails_4:
+.. include:: ../upgrade_to_tails_4.rst
```
eloquence: The only reason we've kept this 1.0.0->1.1.0 upgrade guide around is that it includes the Tails 4 docs. Now that you've nicely split it off, we can do some pruning. I think you can safely prune all upgrade docs up to (not including) 1.3.0->1.4.0 as part of this PR.

eloquence:
This LGTM, suggested two small maintenance tweaks as part of this PR. I think it's fine that the Tails 4 guide appears in the messy "Topic Guides" section of the ToC; it's reasonable to have it included, and we're not going to keep it around for much longer, anyway.

docs/upgrade_to_tails_4.rst
rocodes: removed older upgrade guides in 60130e3d0

rocodes: Of course! fixed in 60130e3d0

docs/upgrade/1.0.0_to_1.1.0.rst
rocodes: Removed in 60130e3d0

rocodes:
Changes made as described above, and a couple slight wording tweaks for clarity.

eloquence:
(Kicking CI, re-reviewing)

eloquence:
Changes LGTM, one minor nit that can be cleaned up later, as well. Only two very distinct commits so no need to squash IMO.

zenmonkeykstop (approved):
LGTM based on Erik's feedback and visual review

PR #2298, opened 2017-09-13
Rebase to latest develop conditionally in CI

msheiny:
## Status

Ready for review

## Description of Changes

Fixes #2080 

Changes proposed in this pull request:

Provides scripting to perform a git rebase on top of develop during the `CircleCI` build process (this is already present in `Travis`).

There is a potential issue with a race conditions here that I'm not sure how best to resolve due to outstanding issues with passing context between workflow jobs in CircleCI (it breaks the ability to SSH debug). The gist of the problem is this -- there is potential for two jobs we have in CircleCI to be tested against different rebasing of develop if someone were to merge into develop at a particular moment in time during the test run. I dont see this as a hugee issue -- but it is something that I thought should be brought up and discussed.

## Testing

Really only applicable to CircleCI. So get out that CI magnifying glass!

## Deployment

Any special considerations for deployment?  Nope - only affects CI

PR #5424, opened 2020-07-29
Removes branch-based conditional execution of CI rebase script

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Updates references to defunct `master` branch in `rebase-ci.sh`.

## Testing

- [ ] CI passes

### DIFF
--- a/devops/scripts/rebase-ci.sh
+++ b/devops/scripts/rebase-ci.sh
@@ -14,8 +14,8 @@ if [[ -n "${CIRCLE_PULL_REQUEST:-}" ]]
 then
     GITHUB_PR_URL="https://api.github.com/repos/freedomofpress/securedrop/pulls/${CIRCLE_PULL_REQUEST##*/}"
 
-    # If current branch is not master or doesnt start with release...
-    if [[ "$CIRCLE_BRANCH" != "master"  && "$CIRCLE_BRANCH" != release*  ]]; then
+    # If current branch is not main or doesnt start with release...
+    if [[ "$CIRCLE_BRANCH" != "main"  && "$CIRCLE_BRANCH" != release*  ]]; then
 
       # Git will yell at you when trying to rebase without author/email configured
       git config --global user.email "ci@freedom.press"
--- a/devops/scripts/rebase-ci.sh
+++ b/devops/scripts/rebase-ci.sh
@@ -14,24 +14,20 @@ if [[ -n "${CIRCLE_PULL_REQUEST:-}" ]]
 then
     GITHUB_PR_URL="https://api.github.com/repos/freedomofpress/securedrop/pulls/${CIRCLE_PULL_REQUEST##*/}"
 
-    # If current branch is not main or doesnt start with release...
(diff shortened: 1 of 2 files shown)


devops/scripts/rebase-ci.sh
```diff
@@ -14,8 +14,8 @@ if [[ -n "${CIRCLE_PULL_REQUEST:-}" ]]
 then
     GITHUB_PR_URL="https://api.github.com/repos/freedomofpress/securedrop/pulls/${CIRCLE_PULL_REQUEST##*/}"
 
-    # If current branch is not master or doesnt start with release...
-    if [[ "$CIRCLE_BRANCH" != "master"  && "$CIRCLE_BRANCH" != release*  ]]; then
+    # If current branch is not main or doesnt start with release...
+    if [[ "$CIRCLE_BRANCH" != "main"  && "$CIRCLE_BRANCH" != release*  ]]; then
```
eloquence: 1. If we keep this `if` clause, should it say `stable`, or drop the `main` condition entirely, since no `main` branch exists in this repo?
2. Does this clause still make sense? It looks like we've carried it over from when Mike first implemented this back in #2298, which hardcoded a rebase against `develop`. That was subsequently fixed in #3617, and we're now always rebasing against the target branch. What does this `if` clause, now, accomplish?

zenmonkeykstop: That's a fair point, and there's no need to skip it for release branches either AFAICT. removed.

conorsch (approved):
LGTM. I see you rolled back your change to the i18n tool logic, which is now targeting upstream master, rather than main. @emkll is there any reason to hold off on a change like https://github.com/freedomofpress/securedrop/pull/5424/commits/10fe54b5ace4c040cf411d2289b3335e9bb44f46#diff-5c5e6b7fb4a6fb6bf1ea79d1c651e0fcL269-R269 ? If so, let's update the i18n repo and press on. 

@zenmonkeykstop depending on how the i18n repo changes shake out, might be prudent to squash here. No concerns with the core logic changes you're making, they're quite clear.

conorsch:
>  is there any reason to hold off on a change like 10fe54b#diff-5c5e6b7fb4a6fb6bf1ea79d1c651e0fcL269-R269 ? If so, let's update the i18n repo and press on.

Just had a mindmeld with @ketudb, and they concluded that the "main" branch needs to be pushed to the "securedrop-i18n" fork, then the change that @zenmonkeykstop reverted will work agan. We don't need to slip that into _this_ PR, so I'm not blocking merge, but would appreciate further input before we declare the branch migration process over and done with.

zenmonkeykstop:
There isn't actually a `main` branch on this repo, default is `develop` and `master` has effectively been defunct for a while (until the recent docs kerfuffle). Hence the revert.

conorsch:
@zenmonkeykstop Thanks for clarifying. Seems to me that the i18n script should be branching from `develop`, then, not from `master` (or `main`). Tangential for this PR, so fine to follow up later!

PR #5447, opened 2020-08-12
Fix Tor Browser capitalisation for readability (source interface)

gonzalo-bulnes:
## Status

Ready for review

## Description of Changes

When reviewing the usage of user-facing strings borrowed from the Tor Browser ([context][ctx]), I noticed that sometimes the English source would read `Tor browser` (lowercase). The three occurrences in this PR belong to the _source_ interface.

I propose using `Tor Browser` instead, primarily to make the instructions more readable to _sources_.

  [ctx]: https://forum.securedrop.org/t/spanish-es-locales-for-securedrop-in-relation-to-tor-browser/1352/4?u=gonzalo-bulnes

**Considerations**

- "Tor browser" has legitimate uses. While "Tor Browser" the name of one such product, you could browse the Tor network with a different browser.
- However, the three sets of instructions make reference specifically to _the_ Tor Browser.
- Using consistently "Tor Browser" as a product name takes advantage of our sensitivity to brands in order to reduce cognitive load of the _sources_. ("Which browser is this about? Ah! _The_ Tor Browser, I've seen that name!")
- There are other instances where "Tor Browser" is written as a product name, so this would also increase consistency.

## Testing

- [ ] Strings don't change length so this change should be very straightforward as far as the UI is concerned.
- [ ] Check for misspellings. 

## Deployment

 - [ ] These strings are internationalized. _I didn't modify the `.pot` file because I assumed strings would be extracted following a dedicated process._

### DIFF
--- a/securedrop/source_templates/logout.html
+++ b/securedrop/source_templates/logout.html
@@ -4,5 +4,5 @@
   </a>
   <br class="clearfix">
   <h1>{{ gettext('One more thing...') }}</h1>
-  <p id="click-new-identity-tor"> {{ gettext('Click the <img src={icon} alt="broom icon" width="16" height="16">&nbsp;<strong>New Identity</strong> button in your Tor browser\'s toolbar. This will clear your Tor browser activity data on this device.').format(icon=url_for('static', filename='i/torbroom-black.png'))  }}</p>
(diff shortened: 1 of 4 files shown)


gonzalo-bulnes:
Note: a staging test in Circle CI [timed out](https://app.circleci.com/pipelines/github/freedomofpress/securedrop/738/workflows/c9d9b399-449a-4835-8405-c9c4b37ca59c/jobs/43343). I couldn't re-run the workflow (all options appear disabled to me).

emkll:
Thanks for the changes @gonzalo-bulnes . The CI timeout is a flake (tracked in https://github.com/freedomofpress/securedrop/issues/5401). I just restarted the staging job

zenmonkeykstop (approved):
LGTM based on visual review.

PR #5448, opened 2020-08-12
Minor fix product names capitalisation for readability

gonzalo-bulnes:
## Status

Ready for review

## Description of Changes

This is a follow up on https://github.com/freedomofpress/securedrop/pull/5447 which fixes the case of the following product names: Tor, Tor Browser, Tails, SSH, Apache.

As far as I can tell, none of these strings are internationalized.
Some of them are public-facing (e.g. docs), others are code or configuration comments, other would likely be visible to _admins_ (comments of iptables rules).

**Considerations**

- Sometimes, specially in configuration comments, names could arguably refer to the executable name **or** the program name (e.g. `# start ssh service`). I opted for the comment to say `# start SSH service` when I saw that some comments would say `apache` (`Apache`) even when the executable name was actually `apache2`.

- In general, I assumed that comments were for human consumption and biased towards using product names when both the product and the executable name were an option.

- I kept commits granular in order to allow for cherry-picking. ~~(In particular, I'm not sure that touching the _iptables_ rules to fix the comment case is an obviously good idea auditability-wise, so I kept those changes in three separate commits.)~~

## Testing

- [ ] Make sure nothing was modified that results in commands being executed differently. ~~(iptables comments are arguably part of the command!)~~
- [ ] Read the comments and documentation text to confirm that they make sense.

## Deployment

- ~~:warning: If there are processes (or documentation) that depend on the exact wording/case of iptables comments, please call it out.~~ 
- [ ] Two strings are internationalized. I didn't modify the .pot file because I assumed strings would be extracted following a dedicated process.

- [ ] Linting (`make lint`)  pass in the development container (I didn't run it locally)
- [x] Tests (`make test`) pass in the development container

- [ ] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass (I didn't run it locally)

_See callout in the **deployment** section._


gonzalo-bulnes:
**Note**: `make test` locally had one failure (**test_login_and_logout**). I haven't investigated further at this point, but it looks like it may be a legit failure. 

The document body was truncated in the test output, but I think the test update may belong with [this change](https://github.com/freedomofpress/securedrop/pull/5447/files#diff-9038818d9dc776fb7530c4d6d3454d0dR7) in https://github.com/freedomofpress/securedrop/pull/5447. (And maybe my idea of separating the two pull requests wasn't the best approach.)

```
tests/test_source.py::test_login_and_logout FAILED

assert 'This will clear your Tor Browser activity data' in '<!DOCTYPE html>\n<html lang="en" dir="ltr">\n  <head>\n    <meta charset="utf-8">\n    <meta name="viewport" content=...">\n    SecureDrop is a project of Freedom of the Press Foundation.\n  </div>\n</footer>    </div>\n  </body>\n</html>'
```

**Edit**: this is fixed, I moved the test update to the PR where the internationalized strings were updated.

gonzalo-bulnes:
Note: there is [one failing step in the Circle CI build](https://app.circleci.com/pipelines/github/freedomofpress/securedrop/739/workflows/3d4012e2-795d-4ca4-beda-91be66567002/jobs/43350) that I believe to be the one that runs `make test`.

For what is worth, I confirm `make test` passes locally.

I couldn't re-run the workflow (all options appear disabled to me), but I'll re-trigger a build by fixing one more thing since I'm looking at Tor Browser strings.

rmol (dismissed):
Approving based on visual review. Someone whose setup currently supports the configuration tests should run those before merging.

install_files/ansible-base/roles/restrict-direct-access/templates/rules_v4
```diff
@@ -5,36 +5,36 @@
 :LOGNDROP - [0:0]
```
zenmonkeykstop: As this file is updated only when the install playbook is run, its changes will not be reflected on many production systems. This means that there will be configuration drift (albeit comments-only) between existing and new instances that may complicate support and automated tests. IMO it would be worth deferring these changes until the next mandatory install playbook run (currently the trusty->focal migration, but could happen before then).

molecule/testinfra/staging/app/iptables-app-staging.j2
```diff
@@ -4,8 +4,8 @@
 :OUTPUT DROP
```
zenmonkeykstop: Same point as above for the prod iptables template.

molecule/testinfra/staging/mon/iptables-mon-prod.j2
```diff
@@ -19,12 +19,12 @@
 -A INPUT -i lo -m comment --comment "Allow lo to lo traffic all protocols" -j ACCEPT
```
zenmonkeykstop: Same point as above for the prod iptables template.

molecule/testinfra/staging/mon/iptables-mon-staging.j2
```diff
@@ -17,14 +17,14 @@
 {% endfor -%}
```
zenmonkeykstop: Same point as above for the prod iptables template - if it's not updated then these shouldn't be either.

zenmonkeykstop (changes requested):
Thanks for this extensive readability fix! I have some concerns re: updates to iptables rules templates and associated tests as per comments, but otherwise it's looking good.

docs/deployment/landing_page.rst
```diff
@@ -120,7 +120,7 @@ services intercept requests between a potential source and the SecureDrop
 
 Do Not Hyperlink .onion Addresses
 ---------------------------------
-Because a visitor to your *Landing Page* may not be using the Tor browser yet,
+Because a visitor to your *Landing Page* may not be using the Tor Browser yet,
```
eloquence: Nit: if we're capitalizing the product name, we should IMO not also use the definite article (`the`), i.e. I would suggest changing such instances to "may not be using Tor Browser yet", in the same way you would say "may not be using Firefox" instead of "may not be using the Firefox".

zenmonkeykstop:
As per @rmol's [comment](https://github.com/freedomofpress/securedrop/pull/5448#pullrequestreview-466817440), testinfra tests are passing with no changes.

install_files/ansible-base/roles/restrict-direct-access/templates/rules_v4
gonzalo-bulnes: This makes a lot of sense to me, I suspected the iptables rules had potential to be a special case _somehow_.

docs/deployment/landing_page.rst
gonzalo-bulnes: I'll remove the articles in front of "Tor Browser" :+1:

gonzalo-bulnes (commented):
- I removed all changes in the `install_files/` directory, as per @zenmonkeykstop's [comment](https://github.com/freedomofpress/securedrop/pull/5448#discussion_r470002207). (:bulb: TIL)

- And I reviewed the use of the article 'the' in front of "Tor Browser" as per @eloquence' [comment](https://github.com/freedomofpress/securedrop/pull/5448#discussion_r470052606). (I hadn't thought of that either!)

:warning: Heads up:

- There are fewer files changed overall, because of removing the `install_files/` changes, but the article review touched some files previously untouched.
- I removed one `the` in internationalized strings in the _source interface_. (But didn't modify the `.pot` file, same as in #5447)

install_files/ansible-base/roles/restrict-direct-access/templates/rules_v4
gonzalo-bulnes: I removed all changes to the `install_files/` directory. https://github.com/freedomofpress/securedrop/pull/5448#pullrequestreview-467929827

molecule/testinfra/staging/mon/iptables-mon-prod.j2
gonzalo-bulnes: :+1:

molecule/testinfra/staging/mon/iptables-mon-staging.j2
gonzalo-bulnes: :+1:

molecule/testinfra/staging/app/iptables-app-staging.j2
gonzalo-bulnes: :+1:

zenmonkeykstop (approved):
Revised changes look good based on visual review, `make testinfra` passes, and string change is present in SI. This is good to go IMO once CI passes (there was an unrelated `staging-test-with-rebase` flake).

eloquence:
Just a note that we should squash those 10 commits, this can be done by the maintainer who merges the change.

gonzalo-bulnes:
I'm happy to squash the commits as well @eloquence  (not right now, so anyone feel free to do it if you prefer!)

zenmonkeykstop:
D'oh, forgot to squash, gonna leave it as is so as to avoid force-pushes on develop.

gonzalo-bulnes:
Oops. Thanks @zenmonkeykstop and everyone!

PR #5450, opened 2020-08-13
Add Weblate screenshot uploader tool

eloquence:
Uploads all page layout test results via Weblate's API, updating existing screenshots as appropriate.

## Status

Ready for review. Towards #3959 (resolves it once we complete a single run in production).

## Description of Changes

This is done as a standalone Python script, with the upload portion written as a `WeblateUploader` class that is as generic as possible, in case it is useful for others solving this problem in other contexts.

Includes documentation for SecureDrop maintainers, and built-in help.

## Test plan

- Ensure unrestricted access to the Weblate sandbox (infra team).
- Manually delete all screenshots in the sandbox.
- Obtain a sandbox API token and export it to an environment variable `WEBLATE_API_TOKEN`
- Export the sandbox URL to `WEBLATE_BASE_URL`
- Generate the screenshots in English via `LOCALES=en_US make translation-test`
- Run the tool via `securedrop/upload-screenshots.py`
- [ ] Observe that a total of 55 screenshots are newly created by checking the count via `/api/screenshots/` (the API has a convenient web UI you can use for this)
- Delete some screenshots, and modify another screenshot in a way that you'll be able to later recognize.
- Run the tool again.
- [ ] Observe that the screenshots you deleted are re-uploaded, and that the screenshot you modified was modified.

### DIFF
--- a/securedrop/upload-screenshots.py
+++ b/securedrop/upload-screenshots.py
@@ -24,7 +24,7 @@ SCREENSHOTS_GLOB = "*.png"
 # filename into the canonical title we give that screenshot in Weblate.
 #
 # Example conversion: "source-session_timeout.png" -> "source: session timeout"
-CANONICALIZATION_RULES = [(r"\.png$", ""), (r"-", ": "), (r"_", " ")]
+CANONICALIZATION_RULES = ((r"\.png$", ""), (r"-", ": "), (r"_", " "))
 
 # Weblate organizes internationalization work into projects and components,
 # which are part of many URLs, and need to be referenced in some API requests.
@@ -77,7 +77,7 @@ def main():
         request_limit=REQUEST_LIMIT,
         canonicalization_rules=CANONICALIZATION_RULES,
     )
-    uploader.safe_upload()
+    uploader.upload()
 
 
 class WeblateUploader(object):
@@ -94,7 +94,7 @@ class WeblateUploader(object):
         component,
         files,
         request_limit,
-        canonicalization_rules=[],
+        canonicalization_rules=(),
     ):
 
         if len(token) != 40:
@@ -123,16 +123,6 @@ class WeblateUploader(object):
             "Authorization": "Token {}".format(token),
         }
         self.session.headers.update(headers)
-
-    def safe_upload(self):
-        """
-        Uploads all files using the screenshots endpoint. Prior to uploading,
-        obtains the list of all existing screenshots. If a file with a
-        canonicalized title that corresponds to the filename already exists,
-        it will be updated, instead of a new upload being created.
-        """
-        existing_screenshots = self.get_existing_screenshots()
-        self.upload(existing_screenshots)
 
     def get_existing_screenshots(self):
         """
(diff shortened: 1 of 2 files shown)


securedrop/bin/upload-screenshots
```diff
@@ -0,0 +1,58 @@
...
+fi
+
+if [ -z "${WEBLATE_API_KEY:-}" ]; then
+   echo "WEBLATE_API_KEY environment variable is not set. This script requires"
+   echo "an API key to be configured to perform file uploads. You can obtain"
+   echo "one via this link: ${WEBLATE_BASE_URL}accounts/profile/#api"
+   exit 1
+fi
+
+BASEDIR=$(dirname "$0")
```
rmol: This works fine, but [elsewhere](https://github.com/freedomofpress/securedrop/blob/96debdc98be53bafac89b5ddaea761a0f71894fa/securedrop/bin/dev-shell#L11) we resolve the repo root with `TOPLEVEL=$(git rev-parse --show-toplevel)`. It's a little simpler and more robust than relative paths, should the script ever be moved. (I know, probably won't. :roll_eyes:)

rmol (commented):
This looks fine to me. I think we can probably even handle updates without rewriting in Python.

eloquence:
The script is now rewritten in Python, and also handles the updating of existing screenshots. Looking forward to review & perhaps running in prod soon. :)

securedrop/upload-screenshots.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def __init__(
+        self,
+        token,
+        base_url,
+        project,
+        component,
+        files,
+        request_limit,
+        canonicalization_rules=[],
```
rmol: Pylint reminds me that lists are [dangerous default arguments](https://docs.python.org/3/tutorial/controlflow.html#default-argument-values). I don't think it matters here.

securedrop/upload-screenshots.py
```diff
@@ -0,0 +1,218 @@
...
+        """
+        Derives a human-readable title from a filename using the defined
+        canonicalization rules, if any. This is used to later update the
+        screenshot.
+        """
+        for pattern, repl in self.canonicalization_rules:
+            filename = re.sub(pattern, repl, filename)
+        return filename
+
+    def upload(self, existing_screenshots=[]):
```
rmol: Same "dangerous default argument" warning here, and while again I don't think it's a problem, it made me think `safe_upload` could be eliminated if you just `get_existing_screenshots` at the beginning of `upload`.

securedrop/upload-screenshots.py
```diff
@@ -0,0 +1,218 @@
...
+
+def main():
+    """
+    Uses the generic WeblateUploader class below to run a SecureDrop screenshot
+    upload.
+    """
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
+    parser.add_argument(
```
rmol: I'm hard-pressed to think of a context in which this would be used where this would matter, but getting credentials from environment variables or files (with proper permissions) rather than command line arguments can be safer, since they can't be read by other users on the same system. The argument can be seen in `ps` output. Again, probably not an issue for this script, even if we eventually automate it.

rmol (dismissed):
This works perfectly. I made a couple of suggestions but am approving without requesting the changes. If you want to make them, I'm happy to rereview.

- Ensure unrestricted access to the Weblate sandbox (infra team).
- Manually delete all screenshots in the sandbox.
- Obtain a sandbox API token and export it to an environment variable `WEBLATE_TOKEN`
- Export the sandbox URL to `SANDBOX_URL`
- Generate the screenshots in English via `LOCALES=en_US make translation-test`
- Run the tool via `securedrop/upload-screenshots.py --token $WEBLATE_TOKEN --baseURL $SANDBOX_URL`
- [x] Observe that a total of 55 screenshots are newly created by checking the count via `/api/screenshots/` (the API has a convenient web UI you can use for this)
- Delete some screenshots, and modify another screenshot in a way that you'll be able to later recognize.
- Run the tool again.
- [x] Observe that the screenshots you deleted are re-uploaded, and that the screenshot you modified was modified.

securedrop/upload-screenshots.py
eloquence: I don't think so either, but it's reasonable to avoid. These rules are intended to not be mutable, so switched to a tuple in eccfba2cc2770b616f127fefebea1919ce30a92f

securedrop/upload-screenshots.py
eloquence: Refactored into one method with a Boolean argument to control its behavior, in eccfba2cc2770b616f127fefebea1919ce30a92f

securedrop/upload-screenshots.py
eloquence: That makes sense. It's a bit less user-friendly but I think still reasonably intuitive, modified in fd004de0070487c0586c6c9f6fa0119020a9f8fd

rmol (approved):

PR #5457, opened 2020-08-18
Enable no_implicit_optional in mypy config

nabla-c0d3:
## Status

Ready for review

## Description of Changes

To help with #4399 , this PR adds `no_implicit_optional = True` to the mypy config and then fixes the few type errors triggered by this change. 
This will eventually be the default setting in mypy ( https://github.com/python/mypy/issues/9091 ) and seemed easy enough to add to Securedrop now, before a lot of non-compatible type annotations get added.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,5 +1,6 @@
 [mypy]
 ignore_missing_imports = True
+no_implicit_optional = True
 python_version = 3.5
 
 [mypy-config]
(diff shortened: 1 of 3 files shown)


zenmonkeykstop (approved):
Looks good based on visual review, no other affected code found, and `make typelint` is passing, 

Approved and thanks, @nabla-c0d3 !

PR #5459, opened 2020-08-18
Add install docs for Fedora and document note on virtualenv 

### LINKED ISSUE #5438: Review and update developer docs
We should routinely review the development environment docs, here:

https://docs.securedrop.org/en/stable/development/setup_development.html

with an eye to: 
- are all instructions still accurate and up-to-date when setting up an environment from scratch;
- are the instructions clear and well-organized;
- do they work across different environments (Linux [different distros], Mac, Windows).

@joaedwar has offered to take a stab at this in the next SecureDrop sprint; exact scope TBD.

nebelsprachen:
## Status

Ready for review 

## Description of Changes

Fixes #5438 

Changes proposed in this pull request: Adds docs on installing in Fedora and other Linux distributions. Also, adds note on creating a virtulalenv in Python3.  

## Testing

1. Open docs site
2. Navigate to *Setting Up the Development Environment* section in docs
3. Navigate to *Documentation Guidelines* section

## Deployment

2. New installs.

Choose one of the following:

### DIFF
--- a/docs/development/documentation_guidelines.rst
+++ b/docs/development/documentation_guidelines.rst
@@ -13,19 +13,21 @@ To get started editing the docs:
 
 .. _clone_the_rep:
 
-#. Clone the SecureDrop repository:
+1. Clone the SecureDrop repository:
 
    .. code:: sh
 
       git clone https://github.com/freedomofpress/securedrop.git
 
-#. Install the dependencies:
+2. Install the dependencies:
 
    .. code:: sh
 
       pip install --no-deps --require-hashes -r securedrop/requirements/python3/develop-requirements.txt
 
-#. Build the docs for viewing in your web browser:
+.. include:: ../includes/virtualenv.txt   
+
+3. Build the docs for viewing in your web browser:
 
    .. code:: sh
 
--- a/docs/development/documentation_guidelines.rst
+++ b/docs/development/documentation_guidelines.rst
@@ -25,7 +25,7 @@ To get started editing the docs:
 
       pip install --no-deps --require-hashes -r securedrop/requirements/python3/develop-requirements.txt
 
-.. note:: Ensure you are running in a virtual environment, so you can install, upgrade, and remove packages using the program `pip`. Run the command `python3 -m venv .venv` to create and manage your virtual environments.
+.. include:: ../includes/virtualenv.txt
 
 #. Build the docs for viewing in your web browser:
 
--- a/docs/development/documentation_guidelines.rst
+++ b/docs/development/documentation_guidelines.rst
@@ -25,7 +25,7 @@ To get started editing the docs:
 
       pip install --no-deps --require-hashes -r securedrop/requirements/python3/develop-requirements.txt
 
-.. include:: ../includes/virtualenv.txt
+.. include:: ../includes/virtualenv.txt   
 
 #. Build the docs for viewing in your web browser:
 
--- a/docs/development/documentation_guidelines.rst
+++ b/docs/development/documentation_guidelines.rst
@@ -13,13 +13,13 @@ To get started editing the docs:
 
 .. _clone_the_rep:
 
-#. Clone the SecureDrop repository:
+1. Clone the SecureDrop repository:
 
    .. code:: sh
 
       git clone https://github.com/freedomofpress/securedrop.git
 
-#. Install the dependencies:
+2. Install the dependencies:
 
    .. code:: sh
 
@@ -27,7 +27,7 @@ To get started editing the docs:
 
 .. include:: ../includes/virtualenv.txt   
 
-#. Build the docs for viewing in your web browser:
+3. Build the docs for viewing in your web browser:
 
    .. code:: sh
 
--- a/docs/development/documentation_guidelines.rst
+++ b/docs/development/documentation_guidelines.rst
@@ -25,6 +25,8 @@ To get started editing the docs:
 
       pip install --no-deps --require-hashes -r securedrop/requirements/python3/develop-requirements.txt
 
+.. note:: Ensure you are running in a virtual environment, so you can install, upgrade, and remove packages using the program `pip`. Run the command `python3 -m venv .venv` to create and manage your virtual environments.
+
 #. Build the docs for viewing in your web browser:
 
    .. code:: sh
--- a/docs/development/setup_development.rst
+++ b/docs/development/setup_development.rst
@@ -24,24 +24,12 @@ Ensure Git is installed on your system. Use package management tools to update y
 
 Ubuntu or Debian GNU/Linux
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
-Run the following commands to update the packages and to install from the default Git repositories:
+Run the following commands to update the package index and to install Git and ``make``:
 
 .. code:: sh
 
    sudo apt-get update
    sudo apt-get install -y make git
-
-
-Fedora and Other Linux Distributions
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-.. note:: To install Docker Engine, you need the 64-bit version of Fedora 30 or higher.
-
-Run the following commands to update the packages and to install from the default Git repositories:
-
-.. code:: sh
-
-   sudo dnf update
-   sudo dnf install -y make git
 
 We recommend using the stable version of Docker CE (Community Edition) which can
 be installed via the official documentation links:
@@ -49,12 +37,32 @@ be installed via the official documentation links:
 * `Docker CE for Ubuntu`_
 * `Docker CE for Debian`_
 
+Make sure to follow the `post-installation steps for Linux`_.
 
-Make sure to follow the `Post-installation steps for Linux`_, as well.
 
-.. _`Docker CE for Ubuntu`: https://docs.docker.com/install/linux/docker-ce/ubuntu/
-.. _`Docker CE for Debian`: https://docs.docker.com/install/linux/docker-ce/debian/
-.. _`Post-installation steps for Linux`: https://docs.docker.com/install/linux/linux-postinstall/
+Fedora and Other Linux Distributions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. note:: To install Docker Engine, you need the 64-bit version of Fedora 30 or higher.
+
+Run the following commands to update the package index and to install Git and ``make``:
+
+.. code:: sh
+
+   sudo dnf install -y make git
+
+
+We recommend using the stable version of Docker CE (Community Edition) which can
+be installed via the official documentation link:
+
+* `Docker CE for Fedora`_
+
+Make sure to follow the `post-installation steps for Linux`_.
+
+.. _`Docker CE for Ubuntu`: https://docs.docker.com/engine/install/ubuntu/
+.. _`Docker CE for Debian`: https://docs.docker.com/engine/install/debian/
+.. _`Docker CE for Fedora`: https://docs.docker.com/engine/install/fedora/
+.. _`post-installation steps for Linux`: https://docs.docker.com/engine/install/linux-postinstall/
+
 
 macOS
 ~~~~~
--- a/docs/development/setup_development.rst
+++ b/docs/development/setup_development.rst
@@ -20,9 +20,11 @@ Quick Start
 The Docker based environment is suitable for developing the web application
 and updating the documentation.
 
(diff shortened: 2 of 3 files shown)


docs/development/documentation_guidelines.rst
```diff
@@ -25,6 +25,8 @@ To get started editing the docs:
 
       pip install --no-deps --require-hashes -r securedrop/requirements/python3/develop-requirements.txt
 
+.. note:: Ensure you are running in a virtual environment, so you can install, upgrade, and remove packages using the program `pip`. Run the command `python3 -m venv .venv` to create and manage your virtual environments.
```
zenmonkeykstop: This is definitely a practice to encourage, so I think you could explicitly make it part of the  docs setup process rather than relegating it to a note. Elsewhere in the [docs](https://docs.securedrop.org/en/stable/development/setup_development.html#setting-up-a-multi-machine-environment) we recommend using `virtualenvwrapper`, so it might make sense to have this be consistent with those instructions. Leveraging includes could make sense here.

docs/development/setup_development.rst
```diff
@@ -29,12 +29,23 @@ Ubuntu or Debian GNU/Linux
    sudo apt-get update
    sudo apt-get install -y make git
 
+
+Fedora and Other Linux Distributions
```
zenmonkeykstop, line 43: For consistency's sake, if we're adding a Fedora section for the dev environment setup we should consider adding it in the staging environment as well. That does get more complicated as there's all the virtualbox and/or libvirt setup to consider, though.

docs/development/setup_development.rst
```diff
@@ -29,12 +29,23 @@ Ubuntu or Debian GNU/Linux
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. note:: To install Docker Engine, you need the 64-bit version of Fedora 30 or higher.
+
+.. code:: sh
+
+   sudo dnf update
```
zenmonkeykstop: Both this and the original Ubuntu section are missing explanatory text as to why these commands should be run - I mean, it's pretty obvious but a line like "first make sure that git is available on your system" wouldn't hurt imo.

zenmonkeykstop (changes requested):

docs/development/setup_development.rst
```diff
@@ -3,7 +3,7 @@ Setting Up the Development Environment
 
 .. include:: ../includes/docs-branches.txt
 
-Prerequisites
+Overview
 -------------
```
eloquence: Nit: It makes no difference in the output, but for readability, I would recommend shortening the length of the underline that creates the headline formatting (the `-------------`) to match the length of the headline.

docs/development/setup_development.rst
eloquence: In the spirit of incremental improvements, adding a note about Fedora to the Docker instructions is IMO a good first step.

docs/development/documentation_guidelines.rst
nebelsprachen: I think the doc flow should be reorganized, with a section that calls out setting up the VM. I included the note as a place holder until that re-architecture is finalized. Do you think this note is okay for the time being, or should I hold off until we restructure the flow?

eloquence: You can follow the examples in `docs/includes` (`grep` through the docs to see how these are used) to create a re-usable text block `virtualenv.txt` which we can initially insert at this position and then place in other relevant sections of the text. 

Per discussion at standup, I suggest we go with `python3 -m venv .venv` as the canonical method (since it will always work without installing additional dependencies) and mention the existence of other convenience tools like `virtualenvwrapper`, which provides the `mkvirtualenv` shell function.

eloquence: Also, as just discussed, let's add instructions for how to activate a virtualenv.

docs/development/setup_development.rst
eloquence: I concur with @zenmonkeykstop that adding a bit more context here would be helpful, @joaedwar.

docs/includes/virtualenv.txt
```diff
@@ -0,0 +1,4 @@
+.. note:: Ensure you are running in a virtual environment, so you can install, upgrade, and       
+          remove packages using the program `pip`. Run the command `python3 -m venv .venv` to create and manage your virtual environments. To begin using the virtual environmen, use the command `source venv/bin/activate` to activate.
```
eloquence: In restructured text, double backticks (``) are used for inline code formatting, while single backticks italicize. All command references here should be enclosed in double backticks.

eloquence: Typo: "virtual environmen" -> "virtual environment"

eloquence: I would clarify the sentence ending with "using the program `pip`". a little bit. You can use `pip` to install packages with or without a virtualenv, what a virtualenv provides is a way to isolate the installation of specific dependencies of a specific application, so that the dependencies of another application (or updates to the OS at a whole) do not get in the way of that application's requirements

eloquence: Typo in the command: `source venv/bin/activate` -> `source .venv/bin/activate` (note leading `.`)

eloquence: Nit: "create and manage your virtual environments" -> "create and manage your virtual environment" (in this case you're just creating one, not multiple)

docs/development/documentation_guidelines.rst
```diff
@@ -25,6 +25,8 @@ To get started editing the docs:
 
       pip install --no-deps --require-hashes -r securedrop/requirements/python3/develop-requirements.txt
 
+.. include:: ../includes/virtualenv.txt
```
eloquence, line 28: The include directive breaks up the automatic numbering via the `#.` syntax (the 3rd item is again restarted at 1). You can fix that by just manually numbering the steps here, i.e. replacing `#.` with `1.`, `2.`, `3.`. You can verify the result of such changes by using the `make docs` autobuild and visiting http://localhost:8000/

eloquence:
Thanks @joaedwar, this is looking great. Left a few comments, don't hesitate to ping if I can help further. Once we merge this PR, I suggest we consolidate the virtualenv instructions you're adding here with the virtualenv-related instructions in the setup chapter: https://docs.securedrop.org/en/stable/development/setup_development.html

docs/development/setup_development.rst
eloquence: Noting a couple of things here on closer inspection:
1) `dnf update` is deprecated in favor of `dnf upgrade` which has the same meaning, see:
https://dnf.readthedocs.io/en/latest/cli_vs_yum.html#update-and-upgrade-commands-are-the-same

2) The `dnf update` / `dnf upgrade` commands behave _differently_ than `apt-get update`. These `dnf` commands attempt to update all packages installed on the system, while `apt-get update` only updates the package index.

To make the instructions equivalent to the Debian instructions, remove the `dnf update` command. Use only the `dnf install` command, which implicitly performs a metadata check to ensure it installs an up-to-date version.

docs/development/setup_development.rst
```diff
@@ -20,21 +20,36 @@ Quick Start
...
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
+Run the following commands to update the packages and to install from the default Git repositories:
 
 .. code:: sh
 
    sudo apt-get update
    sudo apt-get install -y make git
 
+
+Fedora and Other Linux Distributions
```
eloquence, line 43: Further down we are providing instructions for installing Docker under Ubuntu/Debian (the previous section), but they are now placed in the Fedora section.

docs/development/setup_development.rst
```diff
@@ -20,21 +20,36 @@ Quick Start
...
+.. code:: sh
+
+   sudo dnf update
+   sudo dnf install -y make git
+
 We recommend using the stable version of Docker CE (Community Edition) which can
 be installed via the official documentation links:
 
 * `Docker CE for Ubuntu`_
 * `Docker CE for Debian`_
```
eloquence: Note that Docker also provides instructions for installing Docker CE on Fedora, which we should link to in the appropriate section (see earlier comment about the current section structure).

https://docs.docker.com/engine/install/fedora/

docs/development/setup_development.rst
```diff
@@ -20,21 +20,36 @@ Quick Start
 The Docker based environment is suitable for developing the web application
 and updating the documentation.
 
+Ensure Git is installed on your system. Use package management tools to update your local package index. Once the update is complete, you can download and install Git:
 
 Ubuntu or Debian GNU/Linux
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
+Run the following commands to update the packages and to install from the default Git repositories:
```
eloquence: I would shorten to:
"Run the following commands to update the package index and to install Git and `make`"

Technically they are not Git repositories but distribution repositories, but we don't have to spell that out either way.

docs/development/setup_development.rst
```diff
@@ -20,21 +20,36 @@ Quick Start
...
 
    sudo apt-get update
    sudo apt-get install -y make git
 
+
+Fedora and Other Linux Distributions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. note:: To install Docker Engine, you need the 64-bit version of Fedora 30 or higher.
+
+Run the following commands to update the packages and to install from the default Git repositories:
```
eloquence: Same edit as above in the Debian section. As noted in another comment, just running `sudo dnf install git` will implicitly update the package index, while `sudo dnf update` updates all installed packages and is a deprecated alias for `sudo dnf upgrade`.

docs/includes/virtualenv.txt
```diff
@@ -0,0 +1,8 @@
+.. note:: You can install, upgrade, and remove packages using the program ``pip`` in an isolated environment. You can use ``venv`` or ``virtualenv`` to avoid installing Python             packages globally and to isolate the installation of dependencies of a specific application, so that the dependencies of another application do not interfere.
```
eloquence: Nit: Stray whitespace between "Python" and "packages". Regarding this text:

> You can install, upgrade, and remove packages using the program ``pip`` in an isolated environment. 

I think the goal of this paragraph is to provide a reusable explanation of how package installation works for people who are new to Python development. With this goal in mind, I would recommend wording like this:

> You can install, upgrade, and remove Python packages using `pip`, the Python package installer. By default, `pip` will attempt to install packages system-wide. Use `venv` or `virtualenv` to avoid ...

So that we:
1) introduce `pip` (what is its job in the Python world?)
2) explain its default behavior
3) show the value of virtual environments when using `pip` to manage dependencies.

eloquence:
Getting really close. Requesting a few final tweaks :). Thanks for all your work on this so far @joaedwar.

docs/development/setup_development.rst
```diff
@@ -35,11 +37,32 @@ be installed via the official documentation links:
...
 
-Make sure to follow the `Post-installation steps for Linux`_, as well.
+Make sure to follow the `post-installation steps for Linux`_.
+
+
+Fedora and Other Linux Distributions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. note:: To install Docker Engine, you need the 64-bit version of Fedora 30 or higher.
+
+Run the following commands to update the package index and to install Git and ``make``:
```
eloquence: Nit: commands->command

(It's just one command on Fedora)

docs/development/setup_development.rst
```diff
@@ -20,9 +20,11 @@ Quick Start
 The Docker based environment is suitable for developing the web application
 and updating the documentation.
 
+Ensure Git is installed on your system. Use package management tools to update your local package index. Once the update is complete, you can download and install Git:
```
eloquence: Right now this feels repetitive of the instructions that follow, and only partially (why mention Git and not `make` or Docker itself?). I would suggest we speak to the user's needs at a higher level here, e.g., language like the following (please feel free to reword as you see fit):

> Follow the instructions below to install the requirements for the Docker-based environment for your operating system.

eloquence:
Only two comments left from my end and then we can kick it over to a maintainer for a final pass, thanks for your patience. :) You can ignore the CI failure, that's a result of #5470 which is currently being worked on.

Now would be a good time to squash the 15 commits into 1 if you're comfortable doing so; I'm also happy to go over that process in our 1:1.

eloquence:
Update: #5470 has just been fixed, so if you rebase this PR on latest `develop` as part of squashing it, CI should go to green again.

eloquence:
Closing in favor of #5478.

PR #5464, opened 2020-08-22
Add type annotations to journalist_app.api

nabla-c0d3:
## Status

Ready for review.

## Description of Changes

To help with #4399 , this PR adds type annotations to journalist_app/api.py

### DIFF
--- a/securedrop/journalist_app/api.py
+++ b/securedrop/journalist_app/api.py
@@ -1,69 +1,69 @@
 import json
 
 from datetime import datetime, timedelta
+from typing import Tuple, Callable, Any
+
+import flask
+import werkzeug
 from flask import abort, Blueprint, current_app, jsonify, request
 from functools import wraps
+
 from sqlalchemy.exc import IntegrityError
 from os import path
 from uuid import UUID
-from werkzeug.exceptions import default_exceptions  # type: ignore
+from werkzeug.exceptions import default_exceptions
 
 from db import db
 from journalist_app import utils
 from models import (Journalist, Reply, Source, Submission,
                     LoginThrottledException, InvalidUsernameException,
                     BadTokenException, WrongPasswordException)
+from sdconfig import SDConfig
 from store import NotEncrypted
 
 
 TOKEN_EXPIRATION_MINS = 60 * 8
 
 
-def get_user_object(request):
-    """Helper function to use in token_required views that need a user
-    object
-    """
-    auth_token = request.headers.get('Authorization').split(" ")[1]
-    user = Journalist.validate_api_token_and_get_user(auth_token)
-    return user
+def _authenticate_user_from_auth_header(request: flask.Request) -> Journalist:
+    try:
+        auth_header = request.headers['Authorization']
+    except KeyError:
+        return abort(403, 'API token not found in Authorization header.')
+
+    if auth_header:
+        split = auth_header.split(" ")
+        if len(split) != 2 or split[0] != 'Token':
+            abort(403, 'Malformed authorization header.')
+        auth_token = split[1]
+    else:
+        auth_token = ''
+    authenticated_user = Journalist.validate_api_token_and_get_user(auth_token)
+    if not authenticated_user:
+        return abort(403, 'API token is invalid or expired.')
+    return authenticated_user
 
 
-def token_required(f):
+def token_required(f: Callable) -> Callable:
     @wraps(f)
-    def decorated_function(*args, **kwargs):
-        try:
-            auth_header = request.headers['Authorization']
-        except KeyError:
-            return abort(403, 'API token not found in Authorization header.')
-
-        if auth_header:
-            split = auth_header.split(" ")
-            if len(split) != 2 or split[0] != 'Token':
-                abort(403, 'Malformed authorization header.')
-            auth_token = split[1]
-        else:
-            auth_token = ''
-        if not Journalist.validate_api_token_and_get_user(auth_token):
-            return abort(403, 'API token is invalid or expired.')
+    def decorated_function(*args: Any, **kwargs: Any) -> Any:
+        _authenticate_user_from_auth_header(request)
         return f(*args, **kwargs)
     return decorated_function
 
 
-def get_or_404(model, object_id, column=''):
-    if column:
-        result = model.query.filter(column == object_id).one_or_none()
-    else:
-        result = model.query.get(object_id)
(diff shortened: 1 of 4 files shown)


securedrop/journalist_app/api.py
```diff
@@ -1,69 +1,69 @@
...
 
 
-def get_user_object(request):
-    """Helper function to use in token_required views that need a user
-    object
-    """
-    auth_token = request.headers.get('Authorization').split(" ")[1]
-    user = Journalist.validate_api_token_and_get_user(auth_token)
-    return user
+def _authenticate_user_from_auth_header(request: flask.Request) -> Journalist:
```
nabla-c0d3, line 28: There was code to parse the Authorization header in both get_user_object() and token_required(), so i combined them into just one function (because get_user_object() was causing a lot typing errors).

I would suggest passing the resulting/authenticated user as an argument to functions decorated with `@token_require` so that the views that need to know who the current user is don't need to parse the Auth header again (as done in all_source_replies() for example). 
I can make the change if that's useful.

kushaldas (commented):
At least for 3 people, including my local laptop: the branch is throwing this following error:

```
✦ ❯ make  typelint
███ Running mypy type checking...
securedrop/journalist_app/api.py:200: error: Incompatible return value type (got "werkzeug.wrappers.Response", expected "flask.wrappers.Response")
securedrop/journalist_app/api.py:209: error: Incompatible return value type (got "werkzeug.wrappers.Response", expected "flask.wrappers.Response")
Found 2 errors in 1 file (checked 104 source files)
make: *** [Makefile:124: typelint] Error 1
```

kushaldas:
@emkll please have a look at this, I am okay to approve, but want to know why a few of us saw that error.

nabla-c0d3:
Didn't have this error locally but I can fix it anyway, later today or tomorrow.

kushaldas:
> Didn't have this error locally but I can fix it anyway, later today or tomorrow.

Yes, I can see this in my laptop, but not in the other places. I can not even see this in the CI. You don't have to change anything right now. But, we should figure out why are we seeing this in some machines.

emkll:
Strange, I too am seeing the same error @kushaldas is describing. Those errors occur when running `make typelint` locally, either in a virtual environment or in the dev container. Looks like `typelint` is passing in CI. I double checked the version, they are the same in both CI and in my local virtualenv:

```
mypy==0.761
mypy-extensions==0.4.3
```
We should definitely resolve this discrepancy prior to merging.

@nabla-c0d3 if you want to take a stab at resolving the issue, that would be greatly appreciated. Otherwise we would be happy to investigate on your behalf, just let us know. Thank you for your continued contributions to the project!

nabla-c0d3:
I haven't been able to reproduce this error. I do get the same mypy error message if I replace:

```
def serve_file_with_etag(db_obj: Union[Reply, Submission]) -> flask.Response:
```

with 

```
def serve_file_with_etag(db_obj: Union[Reply, Submission]) -> werkzeug.Response:
```

which then returns the error:

```
$ make typelint
Running mypy type checking...
securedrop/journalist_app/api.py:200: error: Incompatible return value type (got "werkzeug.wrappers.Response", expected "flask.wrappers.Response")
securedrop/journalist_app/api.py:209: error: Incompatible return value type (got "werkzeug.wrappers.Response", expected "flask.wrappers.Response")
Found 2 errors in 1 file (checked 104 source files)
make: *** [typelint] Error 1
```

However this branch does have the correct `flask.Response` as the return type, which doesn't trigger the error in CI and on my computer.

nabla-c0d3:
I would suggest merging this PR (so I can continue adding type annotations elsewhere), and then investigating as part of #5516.

emkll:
Sorry for the delay @nabla-c0d3 . The source of the confusion on my end is likely related to the merge commit that included this change as part of conflict resolution in https://github.com/freedomofpress/securedrop/pull/5464/commits/a04143425ff2ad0575a2a23116326bc5e3eaa384

I've taken the liberty of squashing/rebasing this branch on latest `develop` (removing merge commits) for a final round of CI. `make typelint` is now passing locally and in https://github.com/freedomofpress/securedrop/compare/rebase-5464

securedrop/journalist_app/api.py
```diff
@@ -1,69 +1,69 @@
...
         return f(*args, **kwargs)
     return decorated_function
 
 
-def get_or_404(model, object_id, column=''):
-    if column:
-        result = model.query.filter(column == object_id).one_or_none()
-    else:
-        result = model.query.get(object_id)
+def get_or_404(model: db.Model, object_id: str, column: str) -> db.Model:
```
emkll, line 55: Looks like we always specify column as parameter here, so lgtm

emkll (approved):
Thanks for the changes and for your patience @nabla-c0d3 the changes looks good to me, with linting passing locally and CI still passing.

To aid the review of future PR's, it would be helpful if you could:

- Squash/rebase on latest `develop` instead of merging into your branch (though we would be happy to rebase on your behalf, let us know if you'd prefer to do it yourself)
- Either limit or separate code refactoring from type annotations as much as possible 

Thanks again for your contributions! We'll take a look at #5491 next.

securedrop/journalist_app/api.py
rmol: It would generally be better to make functional changes in separate PRs, particularly when they're optimizations and involve authentication, instead of mixing the refactoring with the type hinting.

I preferred the old name, `get_user_object`: it was accurate, if less precise, and a lot shorter.

securedrop/journalist_app/api.py
rmol: Agreed.

securedrop/journalist_app/api.py
```diff
@@ -134,41 +134,43 @@ def get_token():
...
+    def single_source(source_uuid: str) -> Tuple[flask.Response, int]:
         if request.method == 'GET':
             source = get_or_404(Source, source_uuid, column=Source.uuid)
             return jsonify(source.to_json()), 200
         elif request.method == 'DELETE':
             source = get_or_404(Source, source_uuid, column=Source.uuid)
             utils.delete_collection(source.filesystem_id)
             return jsonify({'message': 'Source and submissions deleted'}), 200
+        else:
+            abort(404)
```
rmol, line 153: This is a good addition, as Flask's implicit support for `HEAD` means we're setting up a 500 error when we fall out of the `if` without handling it, but I think a 405 (method not allowed) status would be more accurate here and the other places you've caught this.

PR #5458, opened 2020-08-18
Fixes #5453 updates cffi & argon2_cffi

kushaldas:
## Status

Ready for review 

## Description of Changes

Fixes #5453

Updates both the releases to the latest, these versions builds
properly on Python3.8 on Ubuntu Focal 20.04

## Testing

- [x] CI is green
- [x] Verify hashes introduced in the PR (for the source tar.gz files with the following)
- [x] https://github.com/freedomofpress/securedrop-debian-packaging/wiki/cffi-1.11.5-to-1.14.2
- [x] https://github.com/freedomofpress/securedrop-debian-packaging/wiki/argon2_cffi-18.1.0-to-20.1.0

## Deployment

Any special considerations for deployment? Consider both:

1. Upgrading existing production instances.
2. New installs.

### If you made changes to the server application code:

- [ ] Linting (`make lint`) and tests (`make test`) pass in the development container

### If you made changes to `securedrop-admin`:

- [ ] Linting and tests (`make -C admin test`) pass in the admin development container

### If you made changes to the system configuration:

- [ ] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

### If you made non-trivial code changes:

- [ ] I have written a test plan and validated it for this PR

### If you made changes to documentation:

- [ ] Doc linting (`make docs-lint`) passed locally

### If you added or updated a code dependency:

Choose one of the following:

- [x] I have performed a diff review and pasted the contents to [the packaging wiki](https://github.com/freedomofpress/securedrop-debian-packaging/wiki)
- [ ] I would like someone else to do the diff review

PR #5444, opened 2020-08-12
Adds initial builder-focal molecule scenario

kushaldas:
## Status

Ready for review

## Description of Changes

Fixes #5446

This will enable creating the builder image sd-builder-focal
from the Dockerfile in this commit. We can then update the next
image_hash in the next commit.

## Testing

- [ ] `make build-debs-focal` command should start building packages using the Focal container image. But, the actual debian package builds will fail. The fixes for the actual packages will come in future PRs.

## Deployment

Any special considerations for deployment? Consider both:

1. Upgrading existing production instances.
2. New installs.

### If you made changes to the server application code:

- [ ] Linting (`make lint`) and tests (`make test`) pass in the development container

### If you made changes to `securedrop-admin`:

- [ ] Linting and tests (`make -C admin test`) pass in the admin development container

### If you made changes to the system configuration:

- [ ] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

### If you made non-trivial code changes:

- [ ] I have written a test plan and validated it for this PR

### If you made changes to documentation:

- [ ] Doc linting (`make docs-lint`) passed locally

### If you added or updated a code dependency:

Choose one of the following:

- [ ] I have performed a diff review and pasted the contents to [the packaging wiki](https://github.com/freedomofpress/securedrop-debian-packaging/wiki)
- [ ] I would like someone else to do the diff review

PR #5465, opened 2020-08-24
Adds package builds for Focal 

### LINKED ISSUE #5449: Updating control & debian packaging files for Focal builds of securedrop-app-code and ossec packages
For building packages on Focal and Xenial both, we need different `control` files, and also different `rules`, and `apparmor` rules for the securedrop-app-code and ossec packages. 

These changes are mostly due to Python3.8 on Focal and also newer shared library dependency to build the ossec packages.

## On securedrop-app-code

- control
- rules
- securedrop-app-code.triggers
- usr.sbin.apache2
- `changelog-focal` must be installed as `changelog`

## ossec packages

- `control` in 4 different directories
- verify changelog file on these packages

Then, based on build-platform, we put the right files in the correct location using Ansible tasks.

kushaldas:
## Status

Ready for review

## Description of Changes

Fixes #5449 

This PR contains the changes required for `securedrop-app-code` and `ossec` packages
to be built on Ubuntu Focal. This requires #5458 to be merged first, so that I can rebase.

## Testing

- [ ] `make build-debs-focal` should create all the related packages for Focal
- [ ] `make build-debs` builds packages for Xenial.

## Deployment

2. New installs.

Choose one of the following:


install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/rules-focal
```diff
@@ -0,0 +1,51 @@
...
+override_dh_strip_nondeterminism:
+	find ./debian/ -type f -name '*.pyc' -delete
+	find ./debian/ -type f -name 'pip-selfcheck.json' -delete
+	find -type f -name RECORD -exec sed -i -e '/.*\.pyc.*/d' {} +
+	dh_strip_nondeterminism $@
+
+override_dh_virtualenv:
+	dh_virtualenv \
+		--python=/usr/bin/python3.8 \
+		--extra-pip-arg "--verbose" \
```
conorsch: The only changes I can find in the rules file between xenial & focal are:

1. python3.5 vs. python3.8
2. xenial contains `--setuptools`, whereas focal does not

Ideally we'd have this file templated so the rest is the same, but for now, can you add a comment explaining why setuptools is required for xenial but not for focal?

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/control-focal
```diff
@@ -0,0 +1,15 @@
+Source: securedrop-app-code
+Section: web
+Priority: optional
+Maintainer: SecureDrop Team <securedrop@freedom.press>
+Homepage: https://securedrop.org
+Build-Depends: debhelper (>= 9), dh-python, python3-all, python3-setuptools, dh-systemd, dh-virtualenv
+Standards-Version: 3.9.8
+X-Python3-Version: >= 3.8
```
conorsch: Is the `X-Python3-Version` field actually required? A quick glance at the dh-virtualenv source indicates it's not really used anywhere.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/control-focal
```diff
@@ -0,0 +1,15 @@
...
+
+Package: securedrop-app-code
+Architecture: amd64
+Conflicts: libapache2-mod-wsgi,supervisor
+Replaces: libapache2-mod-wsgi,supervisor
+Depends: ${dist:Depends}, ${misc:Depends}, ${python3:Depends}, apache2, apparmor-utils, coreutils, gnupg2, haveged, libapache2-mod-xsendfile, libpython3.8, paxctld, python3, redis-server, securedrop-config, securedrop-keyring, sqlite3
```
conorsch: The "Depends" line differs only in 3.5 -> 3.8 for the python packages. Can we not use the same name "python3" for both? I realize you'd have to touch xenial files to make this work. What do you think would be easiest to maintain for the remainder of the LTS window?

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2-focal
```diff
@@ -0,0 +1,321 @@
+# Last Modified: Wed Oct 29 08:16:32 2014
+#include <tunables/global>
+
+/usr/sbin/apache2 {
```
conorsch: As with other files, the changes between xenial & focal apparmor profiles are negligible, mostly python3.5 vs python3.8. I'm not sure I see the value in maintaining two difference profiles when we can just add the 3.8 exemptions to the pre-existing apparmor profile. cc @emkll for input

install_files/securedrop-ossec-agent/DEBIAN/control.j2
```diff
@@ -0,0 +1,18 @@
...
+{% if securedrop_build_focal_support %}
+Depends: libevent-2.1.7,libpcre2-8-0,ossec-agent,securedrop-keyring,securedrop-config
+{% else %}
+Depends: libevent-1.4-2,libpcre2-8-0,ossec-agent,securedrop-keyring,securedrop-config
+{% endif %}
+Replaces: ossec-agent
+Conflicts: securedrop-ossec-server
+Description: Installs the securedrop pre-configured OSSEC agent
+  This package installs an OSSEC agent pre-configured for the
+  SecureDrop app server.
```
conorsch, line 18: Solid approach to multi-distro support. :+1: However, elsewhere in the diff I see that ` install_files/securedrop-ossec-agent/DEBIAN/control → ...s/build-ossec-deb-pkg/files/control-focal. Given the `control.j2` introduction here, it apepars that `build-ossec-deb-pkg/files/control-focal` is no longer necessary, and should be removed.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/changelog-focal
```diff
@@ -0,0 +1,83 @@
...
+
+ -- SecureDrop Team <securedrop@freedom.press>  Wed, 17 Jun 2020 21:35:57 +0000
+
+securedrop-app-code (1.3.0+xenial) xenial; urgency=medium
+
+  * See changelog.md
+
+ -- SecureDrop Team <securedrop@freedom.press>  Tue, 12 May 2020 18:37:42 +0000
+
+securedrop-app-code (1.2.2+xenial) xenial; urgency=medium
```
conorsch: Focal changelog should not include historical xenial versions. Remove those, and start with the current/latest version for the focal history. (We did the same when moving from trusty to xenial.)

conorsch (changes requested):
First pass on visual review, requesting some cleanup between the various build directories. Much of the work here is _intentionally_ copy/pasted from the xenial dirs, and I appreciate the your caution in avoiding breaking xenial-specific logic, @kushaldas. See in-line for opportunities to consolidate, or if not, at least document explicitly what's different and why it's different under focal. 

Although I should have suggested it in #5444, the "make build-debs-focal" target is skipping all tests, even though the tests are present (but failing). We should either fix the packaging tests to get them passing, or else update the makefile target to include the "notest" string for clarity.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/control-focal
kushaldas: LTS releases are not supposed to upgrade Python version. That is why I wanted to go this (easy) way.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/rules-focal
kushaldas: I will add the `setuptools`, that was a mistake. Also, I will try to make it a template.

install_files/securedrop-ossec-agent/DEBIAN/control.j2
kushaldas: @conorsch These are two different packages :) `ossec-agent` and `securedrop-ossec-agent`.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/control-focal
kushaldas: Removed.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/rules-focal
kushaldas: I did and then remembered why i don't need it. By default it will install latest `setuptools` and with `--setuptools` flag we can now specify a specific version.

rmol:
It would run a higher risk of breaking the existing Xenial packaging, of course, but would it be possible (or worth the effort) to isolate all the release-specific differences (package dependency lists, paths, whatever) into variable files, and have those included at a single point, instead of sprinkling the "if focal" throughout?

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2-focal
emkll: I agree, with the comment above if possible it would be best to preserve a single AppArmor profile for both platforms. This make the build logic easier to maintain and will allow us to better track the AppArmor profile history over time. Allowing 2 or 1 specific versions of a Python interpreter does not provide any meaningful mitigations here.

Other then Python 3.5 -> 3.8, there are two additions:

```
/opt/venvs/securedrop-app-code/pyvenv.cfg rw
/run/systemd/userdb/io.systemd.DynamicUser rw
```
Are write permissions necessary here?

There are also 2 _removals_ ,  are they intentional?
```
   /var/www/securedrop/static/i/securedrop.png r,                                                                     
   /var/www/securedrop/static/i/securedrop_small.png r, 
```

emkll (commented):
Thanks @kushaldas , I agree with what others have said here, and I am pleased to report the packages are now built successfully :tada: 

As mentioned by a previous reviewer in https://github.com/freedomofpress/securedrop/pull/5465#pullrequestreview-474015580, this will _not_ run the testinfra tests for packages, which can be useful to guard against regressions when doing packaging logic changes. When running the focal builder with tests (`build-debs.sh test focal`), I am observing 22 failing tests.

We should address as part of this PR, or file a follow-up ticket (and track it in the epic). I would lean towards addressing the tests here to both ease review and prevent future regressions.

kushaldas:
>It would run a higher risk of breaking the existing Xenial packaging, of course, but would it be possible (or worth the effort) to isolate all the release-specific differences (package dependency lists, paths, whatever) into variable files, and have those included at a single point, instead of sprinkling the "if focal" throughout?

I am not sure, also the plan was to keep as minimal change as possible for this PR.

install_files/ansible-base/roles/build-ossec-deb-pkg/defaults/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
-build_ossec_deb_pkg_dependencies: ['libevent1-dev','libpcre2-dev']
+build_ossec_deb_pkg_dependencies: ['libevent-dev','libpcre2-dev']
```
eloquence: Why is this dependency changed?

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
```diff
@@ -280,6 +285,8 @@
   /var/www/securedrop/static/i/logo-footer.png r,
   /var/www/securedrop/static/i/no16-global.png r,
   /var/www/securedrop/static/i/no16.png r,
+  /var/www/securedrop/static/i/securedrop.png r,
+  /var/www/securedrop/static/i/securedrop_small.png r,
```
eloquence: As discussed, should not need to be modified.

install_files/ansible-base/roles/build-ossec-deb-pkg/defaults/main.yml
conorsch: +1, if we don't have to change this dependency, let's not. @kushaldas can you check whether the `libevent1-dev` is sufficient for both platforms? If not, can you provide documentation that `libevent-dev` is correct for both?

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/rules-focal
conorsch: > By default it will install

Does "it" refer to the `dh-virtualenv` package? Even if the `--setuptools` flag is truly not required for the build process—and I agree, it doesn't appear to be—we still have an entire file duplicated except for this one flag, so a comment should be added to explain why it's not necessary.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/templates/control.j2
```diff
@@ -5,11 +5,14 @@ Maintainer: SecureDrop Team <securedrop@freedom.press>
...
 
 Package: securedrop-app-code
 Architecture: amd64
 Conflicts: libapache2-mod-wsgi,supervisor
 Replaces: libapache2-mod-wsgi,supervisor
+{% if securedrop_build_focal_support %}
+Depends: ${dist:Depends}, ${misc:Depends}, ${python3:Depends}, apache2, apparmor-utils, coreutils, gnupg2, haveged, libapache2-mod-xsendfile, libpython3.8, paxctld, python3, redis-server, securedrop-config, securedrop-keyring, sqlite3
+{% else %}
 Depends: ${dist:Depends}, ${misc:Depends}, ${python3:Depends}, apache2, apparmor-utils, coreutils, gnupg2, haveged, libapache2-mod-xsendfile, libpython3.5, paxctld, python3 (>= 3.5), python3 (<< 3.6), redis-server, securedrop-config, securedrop-keyring, sqlite3
+{% endif %}
```
conorsch, line 17: Much improved! As discussed today in tech meeting, can we further consolidate the dependencies for `libpython3.8` -> `libpython3` for both platforms? Maybe that's not possible, but if it is, it'd greatly simplify the logic in play.

install_files/securedrop-ossec-server/DEBIAN/control.j2
```diff
@@ -6,7 +6,11 @@ Homepage: https://securedrop.org
 Package: securedrop-ossec-server
 Version: 3.6.0+1.6.0~rc1
 Architecture: amd64
+{% if securedrop_build_focal_support %}
+Depends: libevent-2.1.7,libpcre2-8-0,ossec-server,securedrop-keyring,securedrop-config
+{% else %}
 Depends: libevent-1.4-2,libpcre2-8-0,ossec-server,securedrop-keyring,securedrop-config
```
conorsch, line 12: As above, perhaps we can standardize on a single `libevent` metapackage that would work well. Can you check whether that's possible and add if so?

conorsch:
Only very small changes remain, mostly cleaning up package names and preferring cross-distribution metapackage names where possible. It may not be possible everywhere, but let's reduce the number of times we're forced to say if/else in the build logic.

> As mentioned by a previous reviewer in #5465 (review), this will not run the testinfra tests for packages, which can be useful to guard against regressions when doing packaging logic changes. When running the focal builder with tests (build-debs.sh test focal), I am observing 22 failing tests.

I took the liberty of re-enabling the tests, since recent changes by @kushaldas resulted in all of them passing for me locally. I noticed an apparently-unrelated test flake on the staging (xenial) job in CI, so bounced. 

After a few more package renames if possible, I think this is good to merge—then the next big item will be creating the staging env to _use_ these new packages.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/templates/control.j2
kushaldas: It seems this can not be done.

```
# apt install libpython3  
Reading package lists... Done
Building dependency tree       
E: Unable to locate package libpython3
```

install_files/securedrop-ossec-server/DEBIAN/control.j2
kushaldas:
```
# apt install libevent    
Reading package lists... Done
Building dependency tree       
E: Unable to locate package libevent
```

This is also not possible due to package naming.

install_files/ansible-base/roles/build-ossec-deb-pkg/defaults/main.yml
kushaldas: This has to updated based on `Focal` or `Xenial`.

conorsch:
I plan to re-review this tomorrow, although it'd be grand to get #5477 in first, so we can be sure there are no regressions with the Xenial build logic.

conorsch (approved):
Looks great! The separation between distros is well done here. I suspect we may want to use platform-specific vars in the Ansible logic as continue to drive toward full Focal support, but that's best addressed as part of #5468 anyway.

I made one more change, de-duplicating the packaging tests between Xenial & Focal, to include the new tests we added in #5477. Both packaging targets and all the corresponding tests are passing for me locally, will merge once CI passes.

PR #5472, opened 2020-08-31
Update dev container browser software and validation

rmol:
## Status

Ready for review

## Description of Changes

[Mozilla's keyserver went down, and they might just leave it that way.](https://github.com/mozilla/gpg.mozilla.org/issues/7) Their release signing key is available on another one, but that doesn't seem to be the result of any official policy I can find. The closest thing to guidance comes from a [blog post](https://blog.mozilla.org/security/2019/06/13/updated-firefox-gpg-key/) instructing us to obtain the key from the archive server along with the software. This changes the Dockerfile to do that.

Recent versions of geckodriver are being signed with the release key, so we'll now use it to verify that package too.

We were also using an expired version of the Tor release signing key, kept in our repo, so I'm following their [instructions](https://support.torproject.org/tbb/how-to-verify-signature/) for obtaining the key instead.

Finally, this updates all three packages.

Finally finally, this includes a workaround for the recent `setuptools` breakage.

Fixes #5470.
Fixes #5471.

## Testing

- Check out the `5470-keystone-kops` branch. 
- Start the dev server with `make dev`. Confirm that the development container is built and started properly.

## Deployment

Dev/test only.

### If you made changes to the server application code:

- [ ] Linting (`make lint`) and tests (`make test`) pass in the development container

### If you made non-trivial code changes:

- [x] I have written a test plan and validated it for this PR

PR #5443, opened 2020-08-10
Creates a symlink to the .so file in postint for mod_wsgi

kushaldas:
## Status

Ready for review

## Description of Changes

Fixes  in part of #4768 

On Focal, mod_wsgi expects a shared library for the python module at:

```
$VENV/lib/python3.8/site-packages/mod_wsgi/server/mod_wsgi-py38.so
```
But, the real shared library path is

```
$VENV/lib/python3.8/site-packages/mod_wsgi/server/mod_wsgi-py38.cpython-38-x86_64-linux-gnu.so"
```

So we are creating a symlink with the proper path if we are using Python 3.8 (that is on Focal)

## Testing

- [ ] `make build-deps` on this branch
- [ ] `make staging` to create a fresh staging environment from the debian packages. There should not be any error in the postinstall section
- [ ] Login to app `molecule -s libvirt-staging-xenial -h app-staging` (on mac point to virtualbox) and see that the following file does not exist.

```
/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/mod_wsgi/server/mod_wsgi-py38.so
```

- [ ] `sudo touch /usr/bin/python3.8` to emulate a python3.8 env in focal
- [ ] reinstall the `securedrop-app-code` package or reconfigure it, it should create a symlink as mentioned above.

## Deployment

Any special considerations for deployment? Consider both:

1. Upgrading existing production instances.
2. New installs.

### If you made changes to the server application code:

- [ ] Linting (`make lint`) and tests (`make test`) pass in the development container

### If you made changes to `securedrop-admin`:

- [ ] Linting and tests (`make -C admin test`) pass in the admin development container

### If you made changes to the system configuration:

- [ ] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

### If you made non-trivial code changes:

- [ ] I have written a test plan and validated it for this PR

### If you made changes to documentation:

- [ ] Doc linting (`make docs-lint`) passed locally

### If you added or updated a code dependency:

Choose one of the following:

- [ ] I have performed a diff review and pasted the contents to [the packaging wiki](https://github.com/freedomofpress/securedrop-debian-packaging/wiki)
- [ ] I would like someone else to do the diff review

PR #5484, opened 2020-09-04
Revert to old dh-virtualenv, use built-in venv module

### LINKED ISSUE #5480: Build defenses against setuptools
## Description

We have been regularly derailed by `setuptools` breakage (#5212, #5150, #5471, etc.), despite several attempts to control its installation and usage in our packaging process. And it's just suboptimal to have no control over the versions of tools used to build our production packages (#5109). We should figure out once and for all how to pin it to a working version absolutely everywhere.

We made some progress with the alternative build system of `dh-virtualenv` in #5472, but it has serious bugs, and there are ongoing questions about how to include recent versions of `dh-virtualenv` in the Xenial container. 

There are also some [promising suggestions in #5109](https://github.com/freedomofpress/securedrop/issues/5109#issuecomment-579648769).

rmol:
## Status

Ready for review

## Description of Changes

This takes another run at controlling `setuptools`: instead of installing a newer version of `dh-virtualenv` from Debian unstable for its ability to specify the `setuptools` version via the alternative build system, use it as before but with the Python 3 `venv` module, which will use the local version of `setuptools` that's already present, then replace that in the final package with the version pinned in the securedrop-app-code requirements.

This was the second plan of attack listed in #5480. The old version of `dh-virtualenv` available on Xenial did not provide a way to pass `--no-download` to `virtualenv`, but does offer `--builtin-venv`, which seems to be the way to go anyway, from our work on #5110. This approach lets us use the distribution-provided version of `dh-virtualenv`, which should be safer and more consistent, at least once the `dh-virtualenv` team is providing Focal packages.

Fixes #5480.

## Testing

- `cd molecule/builder-xenial && make`
- `export BUILDER_IMAGE="quay.io/freedomofpress/sd-docker-builder-xenial:2020_09_03"` # (adjust if your image name is different)
- `cd ../.. && make build-debs && make staging`
- Manually verify that the staging site is functioning, or if you're not on Qubes, run the configuration tests.

## Deployment

This changes the production packaging. Ignoring the recent vacillation, the difference to 1.5.0 is primarily the switch to using the Python 3 `venv` module to create the production virtual environment. One change to the AppArmor configuration was needed, for `pyvenv.cfg`.

### DIFF
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/main.yml
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/main.yml
@@ -29,13 +29,13 @@
 - name: Copy requirements.txt to prep directory
   command: cp "{{ securedrop_pip_requirements }}" "{{ securedrop_app_code_prep_dir }}/requirements.txt"
 
-- include: sass.yml
-
-- name: Install SecureDrop Python requirements in container
+- name: Control the version of setuptools used in the default construction of virtual environments
   shell: |
-    pip3 install --no-deps --no-binary :all: --require-hashes -r {{ securedrop_app_code_prep_dir }}/requirements.txt
-  tags:
-    - pip
+    pip3 download --require-hashes -r "{{ securedrop_app_code_prep_dir }}/requirements.txt" --dest /tmp/securedrop-app-code-requirements-download
+    rm -f /usr/share/python-wheels/setuptools-*.whl
+    mv /tmp/securedrop-app-code-requirements-download/setuptools-*.whl /usr/share/python-wheels/
+
+- include: sass.yml
 
 - include: translations.yml
 
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/main.yml
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/main.yml
@@ -31,7 +31,7 @@
 
 - name: Control the version of setuptools used in the default construction of virtual environments
   shell: |
-    pip3 download --require-hashes -r "{{ securedrop_app_code_prep_dir }}/requirements.txt" --dest /tmp/securedrop-app-code-requirements-download
+    pip3 download --no-deps --require-hashes -r "{{ securedrop_app_code_prep_dir }}/requirements.txt" --dest /tmp/securedrop-app-code-requirements-download
     rm -f /usr/share/python-wheels/setuptools-*.whl
     mv /tmp/securedrop-app-code-requirements-download/setuptools-*.whl /usr/share/python-wheels/
 
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/main.yml
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/main.yml
@@ -117,9 +117,6 @@
   args:
     chdir: "{{ securedrop_app_code_deb_dir }}"
   environment:
-    DH_PIP_EXTRA_ARGS: "--verbose --ignore-installed --no-deps --no-binary=:all: --no-cache-dir"
-    DH_UPGRADE_SETUPTOOLS: "46.0.0"
-    DH_VIRTUALENV_ARGUMENTS: "{{  '--python=/usr/bin/python3.8 --setuptools  46.0.0' if securedrop_build_focal_support else '--python=/usr/bin/python3.5 --setuptools' }}"
     DH_VIRTUALENV_INSTALL_ROOT: "/opt/venvs"
 
 - name: Find newly built Debian package
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/templates/control.j2
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/templates/control.j2
@@ -13,6 +13,6 @@ Replaces: libapache2-mod-wsgi,supervisor
 {% if securedrop_build_focal_support %}
 Depends: ${dist:Depends}, ${misc:Depends}, ${python3:Depends}, apache2, apparmor-utils, coreutils, gnupg2, haveged, libapache2-mod-xsendfile, libpython3.8, paxctld, python3, python3-distutils, redis-server, securedrop-config, securedrop-keyring, sqlite3
 {% else %}
-Depends: ${dist:Depends}, ${misc:Depends}, ${python3:Depends}, apache2, apparmor-utils, coreutils, gnupg2, haveged, libapache2-mod-xsendfile, libpython3.5, paxctld, python3 (>= 3.5), python3 (<< 3.6), python3-distutils, redis-server, securedrop-config, securedrop-keyring, sqlite3
+Depends: ${dist:Depends}, ${misc:Depends}, ${python3:Depends}, apache2, apparmor-utils, coreutils, gnupg2, haveged, libapache2-mod-xsendfile, libpython3.5, paxctld, python3 (>= 3.5), python3 (<< 3.6), redis-server, securedrop-config, securedrop-keyring, sqlite3
 {% endif %}
 Description: SecureDrop application code, dependencies, Apache configuration, systemd services, and AppArmor profiles. This package will put the AppArmor profiles in enforce mode.
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/templates/control.j2
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/templates/control.j2
@@ -11,8 +11,8 @@ Architecture: amd64
 Conflicts: libapache2-mod-wsgi,supervisor
 Replaces: libapache2-mod-wsgi,supervisor
 {% if securedrop_build_focal_support %}
-Depends: ${dist:Depends}, ${misc:Depends}, ${python3:Depends}, apache2, apparmor-utils, coreutils, gnupg2, haveged, libapache2-mod-xsendfile, libpython3.8, paxctld, python3, redis-server, securedrop-config, securedrop-keyring, sqlite3
+Depends: ${dist:Depends}, ${misc:Depends}, ${python3:Depends}, apache2, apparmor-utils, coreutils, gnupg2, haveged, libapache2-mod-xsendfile, libpython3.8, paxctld, python3, python3-distutils, redis-server, securedrop-config, securedrop-keyring, sqlite3
 {% else %}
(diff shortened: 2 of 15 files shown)


conorsch:
@rmol Did you happen to check whether these changes straighten out the `VIRTUAL_ENV` path in `post/activate`? (As discussed in https://github.com/freedomofpress/securedrop/pull/5472#issuecomment-685210436.) Let's aim to add a test for that, xfail for now if need be. 

Also prior to merge we'll need to rebuild and repush the xenial builder image.

rmol:
It does fix the activate scripts, but +1 to the test.

kushaldas (commented):
Initial testing on the staging worked with the general QA steps.  I am going to setup a `prod` instance, and then upgrade to this version. And then try again.

kushaldas (dismissed):
In a production vm the upgrade path worked well. Did the standard QA steps after that, and things seems to working as it should be (including `backup` & `restore`). Approved from my side.

conorsch:
> Also prior to merge we'll need to rebuild and repush the xenial builder image.

For future reference, we can build & push to aid in review next time. We needn't wait for approval, since the build logic is careful to pin explicit hashes of containers, so even if the image exists in the remote image repo, it won't be used in CI on other PRs until this PR is merged.

conorsch:
After some additional discussion, @rmol pointed out that we'll want to update the Focal build logic to match these changes. We're not yet running the `make build-debs-focal` target in CI, but soon will. 

So, after updating the logic for Focal to ensure working packages via the venv build system, we'll also need to build & push the Focal container, same as for Xenial. Therefore I'm dismissing @kushaldas's approving review, just to force another approval after the container changes land.

kushaldas:
NOTE:

The `setuptools` version on Focal is `44.0.0`. While using the same style of using `venv` in the Focal build is causing `pip` missing all the dependency packages for `securedrop-app-code` package and then failing there. The error output is more than anything I can copy-paste even on a separate file.

If I try manually in the container to install `setuptools` on a virtual environment, I seeing error similar to https://github.com/pypa/setuptools/issues/2029. 

So, for the experiment purpose, I decided to remove the old version of setuptools wheel from the container, and then wget `setuptools-46.0.0` there. This solves the initial packaging problem, all packages get built properly.  This is the diff of the Dockerfile i used

```diff
diff --git a/molecule/builder-focal/Dockerfile b/molecule/builder-focal/Dockerfile
index a75cde302..97b82e743 100644
--- a/molecule/builder-focal/Dockerfile
+++ b/molecule/builder-focal/Dockerfile
@@ -27,13 +27,15 @@ RUN apt-get -y update && apt-get upgrade -y && apt-get install -y \
         python3-all \
         python3-pip \
         python3-setuptools \
+        python3-venv \
         rsync \
         ruby \
         sqlite \
         sudo \
         tzdata \
         libevent-dev \
-        unzip
+        unzip \
+       wget
 
 
 # TEMPORARY: install dh-virtualenv from debian unstable, pending focal package:
@@ -44,6 +46,9 @@ RUN apt-get install -y debian-archive-keyring
 RUN ln -s /usr/share/keyrings/debian-archive-keyring.gpg /etc/apt/trusted.gpg.d/
 
 RUN apt-get update && apt-get install -y dh-virtualenv
+RUN wget https://files.pythonhosted.org/packages/70/b8/b23170ddda9f07c3444d49accde49f2b92f97bb2f2ebc312618ef12e4bd6/setuptools-46.0.0-py3-none-any.whl
+RUN rm -f /usr/share/python-wheels/setuptools-*.whl
+RUN mv setuptools-46.0.0*.whl /usr/share/python-wheels/
 
 RUN paxctl -cm /usr/bin/python3.8 && mkdir -p /tmp/build
 RUN apt-get clean \

```

But, then while trying to install the package on the Focal staging using https://github.com/freedomofpress/securedrop/pull/5486, the `apache2` service is failing to start. Looking at the error I can see the error is same in https://github.com/freedomofpress/securedrop/pull/5443. The .so file name is not what `mod_wsgi` looking.

We should not block this PR for the focal changes, I suggest that after we merge this, we can do a separate PR to fix the Focal package build.

If anyone wants to verify my findings so far, download https://kushaldas.in/focal.tar.gz into `build/` and untar, it will create the `focal` directory with all the packages from my system. Then do `molecule converge -s libvirt-staging-focal -- --skip-tags grsecurity` using the branch mentioned at #5486 , you should be able to reproduce the same apache2 error.

```sh
✦ ❯ sha256sum focal.tar.gz 
1a0082fbd689899d627a1a9faf9213cb642119f0bb197be9a9806986b3763564  focal.tar.gz
```

kushaldas (dismissed):
- [x] I can build the debian packages on both `Xenial` and on `Focal`.
- [x] `make staging` works
- [x] #5486 draft PR also works

Approved, we need to push new builder image for `Xenial` and `Focal`, and then restart the CI.

kushaldas (changes requested):
The service file tests need update:

```
**Environment=PYTHONPATH="/var/www/securedrop:/opt/venvs/securedrop-app-code/lib/python3.5/site-packages"**
```

did not happen (no ansible var substitution for `securedrop_venv` in the tests:

```
Environment=PYTHONPATH="/var/www/securedrop:{{ securedrop_venv }}/lib/python3.5/site-packages"
```

Seems like https://github.com/freedomofpress/securedrop/blob/5480-old-dhve-builtin-venv/molecule/testinfra/staging/vars/app-staging.yml#L15-L16 these two lines are causing trouble. 

The file paths are (and they are good in my Xenial staging instance):

- /lib/systemd/system/securedrop_source_deleter.service
- /lib/systemd/system/securedrop_rqworker.service
- /lib/systemd/system/securedrop_shredder.service
- /lib/systemd/system/securedrop_rqrequeue.service

molecule/testinfra/staging/vars/staging.yml
```diff
@@ -12,8 +12,8 @@ wanted_apache_headers:
   - 'Header unset Etag'
 
 securedrop_venv: /opt/venvs/securedrop-app-code
-securedrop_venv_bin: "/opt/venvs/securedrop-app-code/bin"
-securedrop_venv_site_packages: "/opt/venvs/securedrop-app-code/lib/python3.5/site-packages"
+securedrop_venv_bin: "{{ securedrop_venv }}/bin"
+securedrop_venv_site_packages: "{{ securedrop_venv }}/lib/python3.5/site-packages"
```
kushaldas: These two lines are breaking the CI (`testinfra` tests)

kushaldas (changes requested):

molecule/testinfra/staging/vars/staging.yml
kushaldas: Because this YAML file is loaded into Python as a dict, so no `jinja` style var substitution there.

rmol: reverted to plain strings.

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/templates/control.j2
```diff
@@ -11,8 +11,8 @@ Architecture: amd64
 Conflicts: libapache2-mod-wsgi,supervisor
 Replaces: libapache2-mod-wsgi,supervisor
 {% if securedrop_build_focal_support %}
-Depends: ${dist:Depends}, ${misc:Depends}, ${python3:Depends}, apache2, apparmor-utils, coreutils, gnupg2, haveged, libapache2-mod-xsendfile, libpython3.8, paxctld, python3, redis-server, securedrop-config, securedrop-keyring, sqlite3
+Depends: ${dist:Depends}, ${misc:Depends}, ${python3:Depends}, apache2, apparmor-utils, coreutils, gnupg2, haveged, libapache2-mod-xsendfile, libpython3.8, paxctld, python3, python3-distutils, redis-server, securedrop-config, securedrop-keyring, sqlite3
 {% else %}
-Depends: ${dist:Depends}, ${misc:Depends}, ${python3:Depends}, apache2, apparmor-utils, coreutils, gnupg2, haveged, libapache2-mod-xsendfile, libpython3.5, paxctld, python3 (>= 3.5), python3 (<< 3.6), redis-server, securedrop-config, securedrop-keyring, sqlite3
+Depends: ${dist:Depends}, ${misc:Depends}, ${python3:Depends}, apache2, apparmor-utils, coreutils, gnupg2, haveged, libapache2-mod-xsendfile, libpython3.5, paxctld, python3 (>= 3.5), python3 (<< 3.6), python3-distutils, redis-server, securedrop-config, securedrop-keyring, sqlite3
```
emkll: Looks like in both cases, `python-distutils` is not in the security channel, we would need to mirror the package and possibly its dependencies. I can't find that exact package in Xenial (maybe it's a metapackage?)

* https://packages.ubuntu.com/xenial/python3-distutils-extra
* https://packages.ubuntu.com/focal/python3-distutils

kushaldas: In `Xenial` the package `libpython3.5-stdlib` provides `distutils` module.

kushaldas (dismissed):
Tested with the following steps:

- [x] package built on `xenial`
- [x] tested on `xenial staging` by hand
- [x] package built on Focal with one patch fixing `ossec` dependency
- [x] finished staging install using @conorsch's focal branch

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/main.yml
```diff
@@ -29,13 +29,13 @@
...
 
-- include: sass.yml
-
-- name: Install SecureDrop Python requirements in container
+- name: Control the version of setuptools used in the default construction of virtual environments
   shell: |
-    pip3 install --no-deps --no-binary :all: --require-hashes -r {{ securedrop_app_code_prep_dir }}/requirements.txt
-  tags:
-    - pip
+    pip3 download --require-hashes -r "{{ securedrop_app_code_prep_dir }}/requirements.txt" --dest /tmp/securedrop-app-code-requirements-download
```
emkll: Is there a reason for omitting the `--no-deps`argument here? Looks like download supports it, and it may be prudent to add it here to avoid downloading transitive dependencies (other than the ones that are hash-pinned in requirements.txt). I guess we are checking them at the install phase later on, so it may not be strictly necessary

rmol: No, I suppose it could save some time. I'll add it.

kushaldas:
@rmol please cherry-pick https://github.com/freedomofpress/securedrop/pull/5486/commits/7682467ab573f23780a46665ddb27317f2289e8e this one in this PR.

conorsch (approved):
Re-tested on staging env. Confirmed functional app on Xenial. For Focal, confirmed that debs build, but did not install them.

Issue #4768, opened 2019-09-09
Support for Ubuntu 18.04


redshiftzero:
## Description

Xenial is the Ubuntu LTS release that is currently supported for SecureDrop and it will be supported until April 2021. 

Bionic is supported until April 2023. I think it would be wise for us in the next year or so to add support for Bionic to reduce the number of installs that will need to be migrated (even if we don't encourage people to migrate until towards the end of the Xenial support window). This means doing many tasks similar to #3204 (the epic tracking Xenial migration tasks), e.g. building packages for Bionic.

If you're thinking that we'll totally rearchitect SecureDrop by then so we don't need to support Bionic: even if a major rearchitecture is done of the server side in the next 1.5 years, we'll still want to ensure that legacy SecureDrops remain supported while the transition occurs. Keep in mind that at the time this ticket was filed the Xenial EOL is 1.5 years away, and I don't think we want to rely on us completing _both_ the Qubes workstation production release AND a production-ready rearchitecture of the server-side within that time window, even ignoring the fact that organizations will need at least a few months to migrate.

### Subtickets

TK

zenmonkeykstop:
Moving to Bionic for internal tooling also unlocks goodness like an up-to-date version of dh_virtualenv, with support for explicitly-set build package versions.

eloquence:
Perhaps we should consider enforcing v3 as part of the Bionic switch. We'll have to do a lot of outreach related to this upgrade, and having admins change the .onion address as part of an upgrade or reinstall seems like a good way to get everyone over the hump. This would also give us a clear date we can communicate to the Tor team by which we know that v2 is no longer supported for SecureDrop, which they could use for end-of-life planning (i.e. April 30, 2021, the EOL date for Ubuntu Xenial).

eloquence:
Recap of [discussions today](https://github.com/freedomofpress/securedrop/wiki/Tech-Meeting-Notes-2020-06-04):
- Potential upgrade strategies, roughly in decreasing order of desirability:
  - unattended upgrade to 18.04 or 20.04
  - attended upgrade to 18.04, then 20.04
  - reinstall on 20.04 (high effort for admins, though could still be least effort and least error-prone if other upgrade paths are very dicey)
  - attended upgrade to 18.04 (only buys us a couple of years)
- To assess feasibility of these strategies, we'll do exploratory spikes in future sprints, both for fresh installs and upgrades
[...]

PR #5486, opened 2020-09-04
Adds libvirt staging environment for Focal

conorsch:
## Status

Ready for review.

Towards #5468.

## Description of Changes

Early stages of Focal support for staging VMs. It's partial progress—areas not currently working are:

* drivers other than libvirt (no virtualbox, no qubes support for staging env)
* kernel tasks must be skipped
* testinfra tests are not passing (mostly kernel tasks & version strings, e.g. python paths)

## Testing

First, prepare the boxes for libvirt, same as required for Xenial:

```
vagrant box add bento/ubuntu-20.04
vagrant box mutate bento/ubuntu-20.04 libvirt
molecule converge -s libvirt-staging-focal -- --skip-tags grsecurity
```

Then browse to the Source Interface and interact with the app. Report any problems.

## Deployment
Yes, provisioning logic has changed a bit. An attempt was made to alter only Focal-related logic where possible, but in two places in particular Xenial-related logic was changed:

* the `aptitude` tasks for deciding whether to mark `tor` as held now uses failed_when=false, for Focal compatibility
* the base apt dependencies in the "common" role now use dist-specific vars, to handle Xenial & Focal differently. 

### If you made changes to the server application code:

- [ ] Linting (`make lint`) and tests (`make test`) pass in the development container

### If you made changes to `securedrop-admin`:

- [ ] Linting and tests (`make -C admin test`) pass in the admin development container

### If you made changes to the system configuration:

- [ ] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

### If you made non-trivial code changes:

- [ ] I have written a test plan and validated it for this PR

### If you made changes to documentation:

- [ ] Doc linting (`make docs-lint`) passed locally

### If you added or updated a code dependency:

Choose one of the following:

- [ ] I have performed a diff review and pasted the contents to [the packaging wiki](https://github.com/freedomofpress/securedrop-debian-packaging/wiki)
- [ ] I would like someone else to do the diff review

PR #5489, opened 2020-09-10
Creates a symlink to the .so file in postint for mod_wsgi

kushaldas:
## Status

Ready for review

## Description of Changes

Fixes  in part of #4768 

On Focal, mod_wsgi expects a shared library for the python module at:

```
$VENV/lib/python3.8/site-packages/mod_wsgi/server/mod_wsgi-py38.so
```
But, the real shared library path is

```
$VENV/lib/python3.8/site-packages/mod_wsgi/server/mod_wsgi-py38.cpython-38-x86_64-linux-gnu.so"
```

So we are creating a symlink with the proper path if we are using Python 3.8 (that is on Focal)

## Testing

- [ ] `make build-deps` on this branch
- [ ] `make staging` to create a fresh staging environment from the debian packages. There should not be any error in the postinstall section
- [ ] Login to app `molecule -s libvirt-staging-xenial -h app-staging` (on mac point to virtualbox) and see that the following file does not exist.

```
/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/mod_wsgi/server/mod_wsgi-py38.so
```

- [ ] `sudo touch /usr/bin/python3.8` to emulate a python3.8 env in focal
- [ ] reinstall the `securedrop-app-code` package or reconfigure it, it should create a symlink as mentioned above.

## Deployment

2. New installs.

Choose one of the following:

### DIFF
--- a/install_files/securedrop-app-code/debian/postinst
+++ b/install_files/securedrop-app-code/debian/postinst
@@ -93,6 +93,14 @@ function adjust_wsgi_configuration {
             perl -pi -e 's/^WSGIProcessGroup journalist.*\n//' "$journalist_conf"
         fi
     fi
+    # Create the WSGI so file link if only on Python3.8
+    wsgi_so_link="/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/mod_wsgi/server/mod_wsgi-py38.so"
+    wsgi_so="/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/mod_wsgi/server/mod_wsgi-py38.cpython-38-x86_64-linux-gnu.so"
+    if test -f "/usr/bin/python3.8"; then
+        if test ! -L  $wsgi_so_link; then
+            ln -s $wsgi_so $wsgi_so_link
+        fi
+    fi
 }
 
 # Manage PaX flags for web app, only required under Xenial.


kushaldas:
This is continuation of #5443

rmol:
I finally figured out why the `mod_wsgi` configuration was broken. The Ubuntu box used in the libvirt staging environment somehow lacks the `python3-distutils` package. When running `mod_wsgi-express module-config`, that causes [this import](https://github.com/GrahamDumpleton/mod_wsgi/blob/f0ff4d43a1148da6da6e17236fbb662d23b92cdb/src/server/__init__.py#L34) to fail, bypassing the determination of the shared library naming scheme, specifically [this](https://github.com/GrahamDumpleton/mod_wsgi/blob/f0ff4d43a1148da6da6e17236fbb662d23b92cdb/src/server/__init__.py#L38).

The generic Ubuntu installation used in the Qubes staging environment has `python3-distutils` by default, so I never saw this failure.

I think the best fix is adding `python3-distutils` as a dependency of `securedrop-app-code`, which I've done in #5484. So now if you build the Focal debs on that branch, then check out the branch of #5486, you get a working `mod_wsgi` configuration.

conorsch:
Thanks for the clarification, @rmol. #5484 is now merged, and #5486 has been rebased to include it, so we should be unblocked to proceed with review over there.

@kushaldas Should we then close this PR? Or would you prefer to test #5486 first?

kushaldas:
I thought I closed this one, it seems I only closed the issue  and forgot the pr.

PR #5491, opened 2020-09-12
Finish adding type annotations to journalist_app

nabla-c0d3:
## Status

Ready.

## Description of Changes

To help with #4399 , this PR adds type annotations to remaining files in journalist_app and enforces typing in this app.

### DIFF
--- a/securedrop/journalist_app/admin.py
+++ b/securedrop/journalist_app/admin.py
@@ -1,7 +1,10 @@
 # -*- coding: utf-8 -*-
 
 import os
+from typing import Optional
+from typing import Union
 
+import werkzeug
 from flask import (Blueprint, render_template, request, url_for, redirect, g,
                    current_app, flash, abort)
 from flask_babel import gettext
@@ -15,26 +18,30 @@ from journalist_app.decorators import admin_required
 from journalist_app.utils import (make_password, commit_account_changes, set_diceware_password,
                                   validate_hotp_secret, revoke_token)
 from journalist_app.forms import LogoForm, NewUserForm, SubmissionPreferencesForm
+from sdconfig import SDConfig
 
 
-def make_blueprint(config):
+def make_blueprint(config: SDConfig) -> Blueprint:
     view = Blueprint('admin', __name__)
 
     @view.route('/', methods=('GET', 'POST'))
     @admin_required
-    def index():
+    def index() -> str:
         users = Journalist.query.all()
         return render_template("admin.html", users=users)
 
     @view.route('/config', methods=('GET', 'POST'))
     @admin_required
-    def manage_config():
+    def manage_config() -> Union[str, werkzeug.Response]:
         # The UI prompt ("prevent") is the opposite of the setting ("allow"):
         submission_preferences_form = SubmissionPreferencesForm(
             prevent_document_uploads=not current_app.instance_config.allow_document_uploads)
         logo_form = LogoForm()
         if logo_form.validate_on_submit():
             f = logo_form.logo.data
+
+            if current_app.static_folder is None:
+                abort(500)
             custom_logo_filepath = os.path.join(current_app.static_folder, 'i',
                                                 'custom_logo.png')
             try:
@@ -55,7 +62,7 @@ def make_blueprint(config):
 
     @view.route('/update-submission-preferences', methods=['POST'])
     @admin_required
-    def update_submission_preferences():
+    def update_submission_preferences() -> Optional[werkzeug.Response]:
         form = SubmissionPreferencesForm()
         if form.validate_on_submit():
             # The UI prompt ("prevent") is the opposite of the setting ("allow"):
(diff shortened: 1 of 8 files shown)


eloquence:
Hi @nabla-c0d3, thanks again for your work on adding type annotations. Now that #5464 is merged, do you have time to finish up the follow-up PRs, or would you like our help with that?

nabla-c0d3:
@eloquence No problem, I will take care of the follow up PRs

nabla-c0d3:
Ready for review.

securedrop/journalist_app/admin.py
```diff
@@ -15,26 +18,30 @@
...
-    def manage_config():
+    def manage_config() -> Union[str, werkzeug.Response]:
         # The UI prompt ("prevent") is the opposite of the setting ("allow"):
         submission_preferences_form = SubmissionPreferencesForm(
             prevent_document_uploads=not current_app.instance_config.allow_document_uploads)
         logo_form = LogoForm()
         if logo_form.validate_on_submit():
             f = logo_form.logo.data
+
+            if current_app.static_folder is None:
```
rmol, line 43: Is it possible for `static_folder` to be `None`? It looks like we're always passing it in the `Flask` constructor in the source and journalist apps, and as a last resort, [`Flask` defaults it to "static"](https://flask.palletsprojects.com/en/1.1.x/api/#application-object).

securedrop/journalist_app/admin.py
```diff
@@ -55,18 +62,20 @@ def manage_config():
...
-    def update_submission_preferences():
+    def update_submission_preferences() -> Optional[werkzeug.Response]:
         form = SubmissionPreferencesForm()
         if form.validate_on_submit():
             # The UI prompt ("prevent") is the opposite of the setting ("allow"):
             flash(gettext("Preferences saved."), "submission-preferences-success")
             value = not bool(request.form.get('prevent_document_uploads'))
             InstanceConfig.set('allow_document_uploads', value)
             return redirect(url_for('admin.manage_config'))
+        else:
```
rmol, line 73: It's a nit, but the `else:` is unnecessary. You don't have to change it; we're doing it this way throughout and it should probably be addressed in a separate PR. I just wanted to note it for the person that takes that on.

securedrop/journalist_app/api.py
```diff
@@ -153,7 +153,7 @@ def single_source(source_uuid: str) -> Tuple[flask.Response, int]:
             utils.delete_collection(source.filesystem_id)
             return jsonify({'message': 'Source and submissions deleted'}), 200
         else:
-            abort(404)
+            abort(405)
```
rmol, line 156: Nice, though as long as we've correctly defined the allowed methods in `@api.route`, Flask will do this. You've made me realize, though, that it does not set the `Allow` header as required by the spec. Other weaknesses in our error handling have come up lately, so I think I'm going to file another issue and suggest we address this there.

securedrop/journalist_app/main.py
```diff
@@ -41,22 +45,24 @@ def login():
...
+    def logout() -> werkzeug.Response:
         session.pop('uid', None)
         session.pop('expires', None)
         session.pop('nonce', None)
         return redirect(url_for('main.index'))
 
     @view.route('/org-logo')
-    def select_logo():
+    def select_logo() -> werkzeug.Response:
+        if current_app.static_folder is None:
```
rmol, line 56: Same question as above.

rmol (changes requested):
This looks good. I have one question about the `static_folder` checks before merging.

securedrop/journalist_app/api.py
nabla-c0d3: This change was based on your comment at https://github.com/freedomofpress/securedrop/pull/5464#discussion_r492294392

securedrop/journalist_app/admin.py
nabla-c0d3: I'm not sure why but `current_app.static_folder` is typed as `Optional[str]`, so mypy will complain if None isn't checked here as the next line `join(current_app.static_folder etc.)` can't work if static_folder is None.

securedrop/journalist_app/admin.py
nabla-c0d3: I like the readability of having an explicit `else` but this method is small anyway.

securedrop/journalist_app/main.py
nabla-c0d3: Same thing, it is typed as `Optional[str]`

securedrop/journalist_app/admin.py
nabla-c0d3: This is the mypy error if you remove this:
```
securedrop\journalist_app\admin.py:43: error: Value of type variable "AnyStr" of "join" cannot be "Optional[str]"
```

securedrop/journalist_app/api.py
rmol: :facepalm: Right. Forgot about `HEAD` being implicitly allowed. I'll try to remember to gather all these into that forthcoming issue.

securedrop/journalist_app/admin.py
rmol: Ah, OK.

rmol (approved):

Issue #5503, opened 2020-09-16
Enable foreign key support


sssoleileraaa:
## Description

Right now we do not have foreign key support enabled for db.sqlite so we're not getting foreign key protections to ensure referential integrity, see https://docs.sqlalchemy.org/en/13/dialects/sqlite.html#foreign-key-support.

## Steps to Reproduce

```
sqlite> select * from sources;
1|db9d0ad6-b45c-4293-819f-6f181b87be91|5LL6MCTO4HJRVPRTHOEUHEV7PLLCKYET2OIL3PEB2NI27EADJZENCZCECXGDKDCZRS3IZLYR44EI5KHVCOWI2WYAV65OGPG4776O4XY=|neither carelessness|0|2020-09-16 07:11:19.777508|0|4|
2|bcc07e32-a366-4335-8db2-eda19601cc3d|HAMMETVY2ZO23ACL3YBNZS47ZFYVHNSI7EFQ63LGL52WKSMXMOGYLEYIHQZ7XE6B6XU2EVRC6JDILXKPJPCPOJMONDRMDMSMEMNIZPI=|featured down|0|2020-09-16 07:11:20.087783|0|4|
sqlite> insert into source_stars (source_id, starred) values ('a', 1);
sqlite> insert into source_stars (source_id, starred) values (99999999, 1);
sqlite> select * from source_stars;
1|a|1
2|99999999|1
```
To enable foreign keys and see them working:
```
sqlite> PRAGMA foreign_keys = ON;
sqlite> insert into source_stars (source_id, starred) values ('b', 1);
Error: FOREIGN KEY constraint failed
sqlite> insert into source_stars (source_id, starred) values (66666666, 1);
Error: FOREIGN KEY constraint failed
```

## Expected Behavior

For foreign keys to be enforced.

## Actual Behavior

Foreign keys are not enforced.

eloquence:
As part of the 9/17-10/1 sprint, we agreed during sprint planning to have a preliminary discussion and investigation focused on enabling foreign key support, e.g.:
- What do we gain by enforcing foreign key constraints?
- What changes do we need to make to existing code to do so?
- What migrations would have to be applied?
- What referential integrity violations could exist in long-running instances?
- How would we guard against unexpected side effects of making this change?

We'll set up an initial tech chat about this (probably a small group), and go from there, aiming to limit our commitment for this sprint to about about 4-8 hours total.

eloquence:
[...]

Issue #5467, opened 2020-08-26
Preserve deleted journalist UUIDs in database to attribute replies


emkll:
## Description

In https://github.com/freedomofpress/securedrop/pull/5178/files, we modified the model to default to a username and uuid set to `deleted` when a journalist is deleted. This is because the API will return the journalist's information via the `/replies` endpoint.

It is therefore impossible (both at the client and the server-side) to distinguish replies sent by distinct deleted journalists.

When deleting a journalist, we should consider:
- preserving their UUID
- setting their username to the `deleted` keyword
- clearing out all other information such as password, firstname and lastname

In the Admin interface user management page, we can either:
- display (but grey out) deleted journalists
- hide journalists that have their username set to `deleted`
Deleted users wouldn't need to appear in the Journalist Interface, as  grayed out/deactivated

## User Stories

As a journalist, I would like to differenciate replies that were written by different former colleagues

prateekj117:
Ok. This would also require changes in https://github.com/freedomofpress/securedrop/pull/5284. I will take a look at this.

rmol:
@prateekj117 We're going to handle this one, Prateek. It involves changes on the client as well that will have to be coordinated.

prateekj117:
Cool.

rmol:
@emkll :+1: I think keeping and scrubbing the `Journalist` record makes sense. The only problem I can see with it is that `username` has a unique constraint, so we can't set it to `deleted`. And to distinguish replies in the client, it would be useful to not just clear out first and last names. I'd suggest populating `username`, `first_name`, and `last_name` randomly using our word lists.

The advantages of keeping the `Journalist` record are:

- It requires less work on the client to keep journalist records in sync. Our current approach of spoofing the info of deleted journalists on replies returned in API responses has caused [problems](https://github.com/freedomofpress/securedrop-client/issues/1143), and I think it'll be easier to resolve those if we're just modifying the `Journalist` records.
- It doesn't require logic scattered around the code base to update the related entities that we keep, like `Reply`.
- It will make it easier to distinguish replies from multiple deleted journalists, instead of attributing them all to the `deleted` account that's currently being created in the client.

[...]

PR #5505, opened 2020-09-16
Add seen tables and update seen status based on downloads in JI

### LINKED ISSUE #5474: Migrate messages, files, and replies to `seen_by` architecture
As part of adding SecureDrop Client support for highlighting sources with unseen submissions (https://github.com/freedomofpress/securedrop-client/issues/187), we need to make server-side changes to record which journalist has seen a message, file, or reply. This breaks down into two main tasks:

- this issue: migrating the existing `downloaded` status and business logic to `seen_by` (for messages, files, and replies);

- issue #5475: adding API endpoints for updating and retrieving `seen_by` status

For now, this issue will have no user-facing impact on Journalist Interface users.

## Definitions

*"Seen" message, file, or reply:*
- in the *Journalist Interface*: message/reply/file download was initiated via browser by a specific user, or reply was sent by a specific user
- in the *SecureDrop Client*: message/reply/file widget (!) was displayed in conversation view by a specific user

*“Read”:* terminology currently used in the Journalist Interface to indicate that a submission has been downloaded by any user

## Technical specification

A working draft specification can be found here:
https://docs.google.com/document/d/1szPEKI1gTpRiM_YI87plKgiW39T7zT4Y4HG0uFnsJD4/edit#heading=h.5llqq3xyh04w

## Functional requirements

The following requirements are in scope of this issue:

- **"Seen by" attribution:** When I download a submission or reply via the Journalist Interface, it is recorded as seen by me in the database.

- **Replies are always seen:** When I send a reply via the Journalist Interface, it is recorded as seen by me in the database.

- **Feature equivalence:** The behavior of the Journalist Interface remains the same as it is today:
  - "Unread" submissions or replies are those that have not been seen by _any_ journalist
  - I can continue to use the "Download Unread" features of the Journalist Interface as before.
  - "Unread" submissions continue to be highlighted as such in the Journalist Interface.
  -  When I migrate an existing SecureDrop installation, all data about which submissions have been downloaded is still displayed in the Journalist Interface in the same manner as before.

sssoleileraaa:
# Description of Changes

Fixes https://github.com/freedomofpress/securedrop/issues/5474
Also fixes a bug found while working on this PR, see "Change 3"

## Change 1

Adds association tables so we know which journalists have seen a file, message, or reply:
  * seen_files
  * seen_replies
  * seen_messages

There has been significant consideration around journalist account deletion. While research is ongoing, we are going to continue to support keeping submissions as downloaded/seen even if the journalist who downloaded/saw the submission was deleted. And this will apply to replies as well. 

<details><summary>Note about unique constraints in sqlite</summary>
<p>
In sqlite, null is the absence of a value, so it is never equal to other null values and not considered a duplicate value. This means that it's possible to insert rows that appear to be duplicates if one of the values is null. So if we have:

```
id  message_id  journalist_id
---------------------------------
1   12                  7
2   12                  8
3   13                  8
4   13                  5
```

and then `journalist_id=8` is deleted, we would then have:

```
id  message_id  journalist_id
---------------------------------
1   12                  7
2   12                  
3   13                  
4   13                  5
```

and then `journalist_id=5` is deleted, we would then have:

```
id  message_id  journalist_id
---------------------------------
1   12                  7
2   12                  
3   13                  
4   13                  
```

Once/if we create a global "deleted" account in the db, we will have to do a data migration to aggregate these duplicate rows and update journalist_id to the journalist_id of the global deleted account.

</p>
</details>

<details><summary>More info about deletion</summary>
<p>

Currently, SecureDrop provides a global inbox, so we preserve journalist replies even when journalist accounts are deleted. The replies are no longer attributed to the journalist, but other journalists will still be able to read the source conversation (which may have replies from multiple journalists) until the conversation is deleted with the source. Currently, we set 'journalist_id' to null in the replies table if the journalist is deleted, which should violate the foreign key constraint on the 'replies' table for 'journalist_id', but due to this issue the constraint is not currently enforced: https://github.com/freedomofpress/securedrop/issues/5503. Also, we lose information about how unique individuals sent those replies that we are persisting without their accounts.

We're also considering full-deletion of journalists, meaning deleting the account and the replies made from that journalist. The issue I see with this is that there are other people sharing the same global inbox and viewing the same source conversations those replies are part of. Removing a journalist's replies when their account is deleted means altering the conversations that other people see. Deleting all the historical data connected to a journalist would also make any global features more challenging. Take global read/unread (aka seen/unseen, the current feature we're working on implementing) for instance. If a journalist was the primary person checking securedrop for a long time and then their account was deleted, potentially many old outdated sources would get bumped up to the top of the list as "unread" if they were the only journalists who read these messages.

As mentioned already, https://github.com/freedomofpress/securedrop/issues/5503 for referential integrity.

Another issue for debate is switching to using uuids as our primary keys. This would remove redundancy and hide sequential ordering of ids tied to account creation. If using ids has made significant performance improvements, I'd be interested in seeing the results. Otherwise, I'd be interested in benchmarking so that we know the tradeoffs.

So here were a couple ideas that were considered when designing the seen tables.

Idea 1:

```py
class SeenFile(db.Model):
    __tablename__ = "seen_files"
    file_id = Column(Integer, ForeignKey("submissions.id"), primary_key=True)
    journalist_id = Column(Integer, ForeignKey("journalists.id"), primary_key=True)
    file = relationship("Submission", backref="seen_files", cascade="delete")
    journalist = relationship("Journalist", backref="seen_files")
```

The association tables declare the foreign keys as primary keys so that the join operations are fast and prevent duplicate records of (journalist_id, message_id), (journalist_id, reply_id), and (journalist_id, file_id). This would require us to implement a global "deleted" journalist account, but there would be work involved in deciding when the "deleted" account should be created (upon first account deletion? by default?) and to be complete and organized about it, we would also want to do a data migration for replies with null journalist_ids. Also, there is some work around catching exceptions when journalists are deleted and making sure we re-attribute replies and seen records to the global "deleted" user. This third point would not be necessary if we implement https://github.com/freedomofpress/securedrop/issues/5467 (which I am in favor of doing in the future).

Idea 2 (the winner):

```py
class SeenFile(db.Model):
    __tablename__ = "seen_files"
    __table_args__ = (
        db.UniqueConstraint('file_id', 'journalist_id'),
    )
    id = Column(Integer, primary_key=True)
    file_id = Column(Integer, ForeignKey("submissions.id"), nullable=False)
    journalist_id = Column(Integer, ForeignKey("journalists.id"), nullable=True)
    file = relationship("Submission", backref="seen_files", cascade="delete")
    journalist = relationship("Journalist", backref="seen_files")
```

We still have the same foreign keys but there is a new sequential 'id' column that is the primary key and a unique constraint on (file_id, journalist_id) to prevent duplicate records. Also, journalist_id can now be null. This is a workaround solution for allowing null 'journalist_id' (once we turn on foreign key support we will have to update this table along with replies) to support deleted accounts.

</p>
</details>

## Change 2

The journalist interface no longer marks submissions as 'downloaded' and instead makes entries into the seen tables when:
  * files, messages, or replies are downloaded via the "Download Selected" button, by clicking on the filename link, and by clicking on the "<count> unread" link from the "Sources" page
  * files are downloaded from the client

The submissions table now has a `seen` property that is set if 'downloaded' is set or there is a corresponding 'seen' entry in the database. This is for backwards compatibility.

<details><summary>Note about design choice</summary>
<p>
With a preference for avoiding fake/ special numbers and accounts, it made the most sense to keep the 'downloaded' column in the 'submissions' table and to leave the data there. 'downloaded' can now be used for historical data where we don’t know which journalist saw a submission. Other options considered were:
  * data migration using a fake 'journalist_id' of a very large number in the 'seen_files' table and removal of 'downloaded'
  * data migration using a special journalist account to represent an "unknown" user, as distinct from "deleted", and removal of 'downloaded'
</p>
</details>

## Change 3

Fixed a bug where we were marking all files and messages as downloaded when the client opens from the JI endpoint `/sources/<source_uuid>/submissions/<submission_uuid>/download`, which is used in the background by the client just to get the latest messages and used to download files per user request. We now no longer mark files and messages as 'downloaded'.

Once we start using the new seen endpoints in the client we will be able to mark source messages as seen when a user clicks on the source and files as seen when a user clicks to open a file.

# Testing

## New tables and account deletion [Change 1]
0. Make sure alembic upgrade and downgrade work as expected (you can just check that tests pass as well
1. Check out this branch
2. `make dev`
3. `docker exec -it securedrop-dev-0 bash`
4. `sqlite3 /var/lib/securedrop/db.sqlite`
5. Check existence of tables
     - [ ] verify 'seen_replies', 'seen_files', and 'seen_messages' exist
6. Check deletion of a journalist
     -  Log in from JI as dellsberg and download a message
     - [ ] verify record exists in seen_messages table
     - Log in from JI as test-user and download the same message
     - [ ] verify new record exists in seen_messages table
     - Log in as journalist and delete dellsberg and test-user
     - [ ] verify journalist_id is now null for both records
7. Check unique constraints for non-nulls
     - `insert into seen_files (file_id, journalist_id) values (1, 1);` twice
     - [ ] verify constraint does not allow this
     - repeat for 'seen_messages' and 'seen_replies'
8. Check deletion of a file/message/reply
     - Send a file attachment from the source interface
     - Download it from the JI
     - [ ] verify seen record exists for this file
     - Now delete the file from the JI
     - [ ] verify record is deleted in the seen_files table
     - repeat for replies and messages
9. Check deletion of a source
     - Delete a source with seen records for replies, files, and messages
     - [ ] verify seen records no longer exist

## Behavior has not changed in the JI [Change 2]
1. Check out this branch
2. `make dev`
3. `docker exec -it securedrop-dev-0 bash`
4. `sqlite3 /var/lib/securedrop/db.sqlite`
5. Check x-unread link
     - Click on the "<x> unread" link for the source at the top of the list
     - [ ] verify JI works the same as before
     - [ ] verify the correct entries were added to the 'seen_messages' table for submissions
     - [ ] verify 'downloaded' was not set in the db
     - Repeat for files
6. Check "Download Unread" button
     - Send a file and message
     - Select the source with the unread file and message and click "Download Unread"
     - [ ] verify JI works the same as before
     - Select the source again with no unread files and messages and click "Download Unread"
     - [ ] verify that you get error message "No unread submissions in selected collections."
     - Send a couple files
     - Select all sources in the source list and click "Download Unread"
     - [ ] verify JI works the same as before
7. Check "Download Selected" button
     - Send a file, message, and reply
     - Select all conversation items for this source and click "Download Selected"
     - [ ] verify JI works the same as before
     - [ ] verify the correct entries were added to the 'seen_*' table
     - [ ] verify 'downloaded' was not set in the db
8. Check filename-link
     - Send a file
     - Click on the filename of that file from the JI
     - [ ] verify JI works the same as before
     - [ ] verify the correct entry was added to the 'seen_files' table
     - [ ] verify 'downloaded' was not set in the db
     - Repeat for message
9. Check unique constraint exception handling
     - [ ] verify that downloading the same file, message, or reply again and again works and that there is only one seen entry for each
10. Check that sending a reply as a journalist creates seen record
     - [ ] Send a reply to a source
     - [ ] Verify seen record exists in the seen_replies table

## New behavior in the JI [Change 3]
1. Restart `make dev`
2. Send 2 file attachments from the source interface as a new source
3. Visit JI to confirm that there are unread files
4. Log into the client
     - [ ] verify no files or messages are marked as read in the JI due to logging into the client
5. Download one file from the client
     - [ ] verify the file is not marked as read in the JI
     - [ ] verify 'downloaded' was not set in the db
6. Check that you can download a file that was seen via the JI
     - [ ] Download the unread file from the JI
     - [ ] Download the same file from the client
     - [ ] verify download was successful

## Migration testing

**Note:** the migration is in the postinst of the package

(test on staging or dev)
1. On your server running the latest securedrop, download some messages and files from the JI
2. Checkout this branch, rebuild, and restart the server (for staging: build the debs and make staging, for dev: run make dev)
3. Run `manage.py init-db` to apply the migration (Either ssh to your staging server or `docker exec -it securedrop-dev-0 bash` if you're testing on a dev server)
4. Revisit the JI and see that the same files and messages are still downloaded
5. Download some new messages, files, and replies (these will be marked as seen in the new seen tables)
6. From the latest release branch, either a) build the debs and make staging, or b) make dev
7. Checkout the latest release branch, rebuild, and restart the server (for staging: build the debs and make staging, for dev: run make dev)
8. Revisit the JI and see that only the messages and files you downloaded from step 2 are still downloaded
9. Download some more stuff to make sure nothing went horribly wrong during the downgrade

Or, instead of all the steps above, you can just do: https://docs.securedrop.org/en/stable/development/upgrade_testing.html?highlight=upgrade%20testing#upgrade-testing-using-molecule

## QA Testing

Remember to follow https://docs.securedrop.org/en/stable/development/database_migrations.html?highlight=alembic#release-testing-migrations during QA testing

# Upgrading existing production instances

N/A since database migrations are applied postinst

### DIFF
--- a/securedrop/alembic/versions/48a75abc0121_add_seen_tables.py
+++ b/securedrop/alembic/versions/48a75abc0121_add_seen_tables.py
@@ -1,0 +1,57 @@
+"""add seen tables
+
+Revision ID: 48a75abc0121
+Revises: 35513370ba0d
+Create Date: 2020-09-15 22:34:50.116403
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "48a75abc0121"
+down_revision = "35513370ba0d"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "seen_files",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("file_id", sa.Integer(), nullable=False),
+        sa.Column("journalist_id", sa.Integer(), nullable=True),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("file_id", "journalist_id"),
+        sa.ForeignKeyConstraint(["file_id"], ["submissions.id"]),
+        sa.ForeignKeyConstraint(["journalist_id"], ["journalists.id"]),
+    )
+
+    op.create_table(
+        "seen_messages",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("message_id", sa.Integer(), nullable=False),
+        sa.Column("journalist_id", sa.Integer(), nullable=True),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("message_id", "journalist_id"),
+        sa.ForeignKeyConstraint(["message_id"], ["submissions.id"]),
+        sa.ForeignKeyConstraint(["journalist_id"], ["journalists.id"]),
+    )
+
+    op.create_table(
+        "seen_replies",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("reply_id", sa.Integer(), nullable=False),
+        sa.Column("journalist_id", sa.Integer(), nullable=True),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("reply_id", "journalist_id"),
+        sa.ForeignKeyConstraint(["reply_id"], ["replies.id"]),
+        sa.ForeignKeyConstraint(["journalist_id"], ["journalists.id"]),
+    )
+
+
+def downgrade():
+    op.drop_table("seen_files")
+    op.drop_table("seen_messages")
+    op.drop_table("seen_replies")
--- a/securedrop/journalist_app/main.py
+++ b/securedrop/journalist_app/main.py
@@ -5,12 +5,11 @@ from datetime import datetime
 from flask import (Blueprint, request, current_app, session, url_for, redirect,
                    render_template, g, flash, abort)
 from flask_babel import gettext
-from sqlalchemy.sql.expression import false
 
 import store
 
 from db import db
-from models import Source, SourceStar, Submission, Reply
+from models import SeenReply, Source, SourceStar, Submission, Reply
 from journalist_app.forms import ReplyForm
 from journalist_app.utils import (validate_user, bulk_delete, download,
                                   confirm_bulk_delete, get_source)
@@ -112,10 +111,13 @@ def make_blueprint(config):
              config.JOURNALIST_KEY],
             output=current_app.storage.path(g.filesystem_id, filename),
         )
-        reply = Reply(g.user, g.source, filename)
 
         try:
+            reply = Reply(g.user, g.source, filename)
             db.session.add(reply)
+            db.session.flush()
+            seen_reply = SeenReply(reply_id=reply.id, journalist_id=g.user.id)
+            db.session.add(seen_reply)
             db.session.commit()
             store.async_add_checksum_for_file(reply)
         except Exception as exc:
--- a/securedrop/journalist_app/main.py
+++ b/securedrop/journalist_app/main.py
@@ -74,9 +74,9 @@ def make_blueprint(config):
                 starred.append(source)
             else:
                 unstarred.append(source)
-            source.num_unread = len(
-                Submission.query.filter_by(source_id=source.id,
-                                           downloaded=False).all())
+            submissions = Submission.query.filter_by(source_id=source.id).all()
+            unseen_submissions = [s for s in submissions if not s.seen]
+            source.num_unread = len(unseen_submissions)
 
         return render_template('index.html',
                                unstarred=unstarred,
@@ -172,13 +172,12 @@ def make_blueprint(config):
     def download_unread_filesystem_id(filesystem_id):
         id = Source.query.filter(Source.filesystem_id == filesystem_id) \
                          .filter_by(deleted_at=None).one().id
-        submissions = Submission.query.filter(
-            Submission.source_id == id,
-            Submission.downloaded == false()).all()
-        if submissions == []:
+        submissions = Submission.query.filter(Submission.source_id == id).all()
+        unseen_submissions = [s for s in submissions if not s.seen]
+        if unseen_submissions == []:
             flash(gettext("No unread submissions for this source."))
             return redirect(url_for('col.col', filesystem_id=filesystem_id))
         source = get_source(filesystem_id)
-        return download(source.journalist_filename, submissions)
+        return download(source.journalist_filename, unseen_submissions)
 
     return view
--- a/securedrop/journalist_app/utils.py
+++ b/securedrop/journalist_app/utils.py
@@ -9,16 +9,29 @@ import werkzeug
 from flask import (g, flash, current_app, abort, send_file, redirect, url_for,
                    render_template, Markup, sessions, request)
 from flask_babel import gettext, ngettext
-from sqlalchemy.exc import IntegrityError
-from sqlalchemy.sql.expression import false
 
 import i18n
 
 from db import db
-from models import (get_one_or_else, Source, Journalist, InvalidUsernameException,
-                    WrongPasswordException, FirstOrLastNameError, LoginThrottledException,
-                    BadTokenException, SourceStar, PasswordError, SeenFile, SeenMessage, SeenReply,
-                    Submission, RevokedToken, InvalidPasswordLength, Reply)
+from models import (
+    BadTokenException,
+    FirstOrLastNameError,
+    InvalidPasswordLength,
+    InvalidUsernameException,
+    Journalist,
+    LoginThrottledException,
+    PasswordError,
+    Reply,
+    RevokedToken,
+    SeenFile,
+    SeenMessage,
+    SeenReply,
+    Source,
+    SourceStar,
+    Submission,
+    WrongPasswordException,
+    get_one_or_else,
+)
 from store import add_checksum_for_file
 
 from sdconfig import SDConfig
@@ -161,34 +174,43 @@ def download(zip_basename: str, submissions: List[Union[Submission, Reply]]) -> 
     :param list submissions: A list of :class:`models.Submission`s to
                              include in the ZIP-file.
     """
-    zf = current_app.storage.get_bulk_archive(submissions,
-                                              zip_directory=zip_basename)
+    zf = current_app.storage.get_bulk_archive(submissions, zip_directory=zip_basename)
     attachment_filename = "{}--{}.zip".format(
-        zip_basename, datetime.datetime.utcnow().strftime("%Y-%m-%d--%H-%M-%S"))
+        zip_basename, datetime.datetime.utcnow().strftime("%Y-%m-%d--%H-%M-%S")
+    )
 
-    # mark as seen by the current user and update downloaded for submissions
-    journalist_id = g.get('user').id
+    # mark as seen by the current user
+    journalist_id = g.get("user").id
     for item in submissions:
-        try:
-            if item.filename.endswith('reply.gpg'):
+        if item.filename.endswith("reply.gpg"):
+            already_seen = SeenReply.query.filter_by(
+                reply_id=item.id, journalist_id=journalist_id
+            ).one_or_none()
+            if not already_seen:
                 seen_reply = SeenReply(reply_id=item.id, journalist_id=journalist_id)
                 db.session.add(seen_reply)
-            elif item.filename.endswith('-doc.gz.gpg'):
+        elif item.filename.endswith("-doc.gz.gpg"):
+            already_seen = SeenFile.query.filter_by(
+                file_id=item.id, journalist_id=journalist_id
+            ).one_or_none()
+            if not already_seen:
                 seen_file = SeenFile(file_id=item.id, journalist_id=journalist_id)
                 db.session.add(seen_file)
-                item.downloaded = True
-            else:
+        else:
+            already_seen = SeenMessage.query.filter_by(
+                message_id=item.id, journalist_id=journalist_id
+            ).one_or_none()
+            if not already_seen:
                 seen_message = SeenMessage(message_id=item.id, journalist_id=journalist_id)
                 db.session.add(seen_message)
-                item.downloaded = True
+        db.session.commit()
 
-            db.session.commit()
-        except IntegrityError:
-            pass  # expected not to store that a file was seen by the same user multiple times
-
-    return send_file(zf.name, mimetype="application/zip",
-                     attachment_filename=attachment_filename,
-                     as_attachment=True)
+    return send_file(
+        zf.name,
+        mimetype="application/zip",
+        attachment_filename=attachment_filename,
+        as_attachment=True,
+    )
 
 
 def delete_file_object(file_object: Union[Submission, Reply]) -> None:
@@ -354,19 +376,25 @@ def set_diceware_password(user: Journalist, password: Optional[str]) -> bool:
 
 
 def col_download_unread(cols_selected: List[str]) -> werkzeug.Response:
-    """Download all unread submissions from all selected sources."""
-    submissions = []  # type: List[Union[Source, Submission]]
+    """
+    Download all unseen submissions from all selected sources.
+    """
+    unseen_submissions = []  # type: List[Union[Source, Submission]]
+
     for filesystem_id in cols_selected:
-        id = Source.query.filter(Source.filesystem_id == filesystem_id) \
-                         .filter_by(deleted_at=None).one().id
-        submissions += Submission.query.filter(
-            Submission.downloaded == false(),
-            Submission.source_id == id).all()
-    if submissions == []:
-        flash(gettext("No unread submissions in selected collections."),
-              "error")
-        return redirect(url_for('main.index'))
-    return download("unread", submissions)
+        source = (
+            Source.query.filter(Source.filesystem_id == filesystem_id)
+            .filter_by(deleted_at=None)
+            .one()
+        )
+        submissions = Submission.query.filter_by(source_id=source.id).all()
+        unseen_submissions += [s for s in submissions if not s.seen]
+
+    if not unseen_submissions:
+        flash(gettext("No unread submissions in selected collections."), "error")
+        return redirect(url_for("main.index"))
+
+    return download("unread", unseen_submissions)
 
 
 def col_download_all(cols_selected: List[str]) -> werkzeug.Response:
--- a/securedrop/journalist_app/utils.py
+++ b/securedrop/journalist_app/utils.py
@@ -9,6 +9,7 @@ import werkzeug
 from flask import (g, flash, current_app, abort, send_file, redirect, url_for,
                    render_template, Markup, sessions, request)
 from flask_babel import gettext, ngettext
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.sql.expression import false
 
 import i18n
@@ -16,8 +17,8 @@ import i18n
 from db import db
 from models import (get_one_or_else, Source, Journalist, InvalidUsernameException,
                     WrongPasswordException, FirstOrLastNameError, LoginThrottledException,
-                    BadTokenException, SourceStar, PasswordError, Submission, RevokedToken,
-                    InvalidPasswordLength, Reply)
+                    BadTokenException, SourceStar, PasswordError, SeenFile, SeenMessage, SeenReply,
+                    Submission, RevokedToken, InvalidPasswordLength, Reply)
 from store import add_checksum_for_file
 
 from sdconfig import SDConfig
@@ -165,10 +166,25 @@ def download(zip_basename: str, submissions: List[Union[Submission, Reply]]) -> 
     attachment_filename = "{}--{}.zip".format(
         zip_basename, datetime.datetime.utcnow().strftime("%Y-%m-%d--%H-%M-%S"))
 
-    # Mark the submissions that have been downloaded as such
-    for submission in submissions:
-        submission.downloaded = True
-    db.session.commit()
+    # mark as seen by the current user and update downloaded for submissions
+    journalist_id = g.get('user').id
+    for item in submissions:
+        try:
+            if item.filename.endswith('reply.gpg'):
+                seen_reply = SeenReply(reply_id=item.id, journalist_id=journalist_id)
+                db.session.add(seen_reply)
+            elif item.filename.endswith('-doc.gz.gpg'):
+                seen_file = SeenFile(file_id=item.id, journalist_id=journalist_id)
+                db.session.add(seen_file)
+                item.downloaded = True
+            else:
+                seen_message = SeenMessage(message_id=item.id, journalist_id=journalist_id)
+                db.session.add(seen_message)
+                item.downloaded = True
+
+            db.session.commit()
+        except IntegrityError:
+            pass  # expected not to store that a file was seen by the same user multiple times
 
     return send_file(zf.name, mimetype="application/zip",
                      attachment_filename=attachment_filename,
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -722,6 +722,40 @@ class Journalist(db.Model):
         return json_user
 
 
+class SeenFile(db.Model):
+    __tablename__ = "seen_files"
+    __table_args__ = (db.UniqueConstraint("file_id", "journalist_id"),)
+    id = Column(Integer, primary_key=True)
+    file_id = Column(Integer, ForeignKey("submissions.id"), nullable=False)
+    journalist_id = Column(Integer, ForeignKey("journalists.id"), nullable=True)
+    file = relationship(
+        "Submission", backref=backref("seen_files", cascade="all,delete")
+    )
+    journalist = relationship("Journalist", backref=backref("seen_files"))
+
+
+class SeenMessage(db.Model):
+    __tablename__ = "seen_messages"
+    __table_args__ = (db.UniqueConstraint("message_id", "journalist_id"),)
+    id = Column(Integer, primary_key=True)
+    message_id = Column(Integer, ForeignKey("submissions.id"), nullable=False)
+    journalist_id = Column(Integer, ForeignKey("journalists.id"), nullable=True)
+    message = relationship(
+        "Submission", backref=backref("seen_messages", cascade="all,delete")
+    )
+    journalist = relationship("Journalist", backref=backref("seen_messages"))
+
+
+class SeenReply(db.Model):
+    __tablename__ = "seen_replies"
+    __table_args__ = (db.UniqueConstraint("reply_id", "journalist_id"),)
+    id = Column(Integer, primary_key=True)
+    reply_id = Column(Integer, ForeignKey("replies.id"), nullable=False)
+    journalist_id = Column(Integer, ForeignKey("journalists.id"), nullable=True)
+    reply = relationship("Reply", backref=backref("seen_replies", cascade="all,delete"))
+    journalist = relationship("Journalist", backref=backref("seen_replies"))
+
+
 class JournalistLoginAttempt(db.Model):
 
     """This model keeps track of journalist's login attempts so we can
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -229,6 +229,17 @@ class Submission(db.Model):
         }
         return json_submission
 
+    @property
+    def seen(self) -> bool:
+        """
+        If the submission has been downloaded or seen by any journalist, then the submssion is
+        considered seen.
+        """
+        if self.downloaded or self.seen_files.count() or self.seen_messages.count():
+            return True
+
+        return False
+
 
 class Reply(db.Model):
     __tablename__ = "replies"
@@ -729,7 +740,7 @@ class SeenFile(db.Model):
     file_id = Column(Integer, ForeignKey("submissions.id"), nullable=False)
     journalist_id = Column(Integer, ForeignKey("journalists.id"), nullable=True)
     file = relationship(
-        "Submission", backref=backref("seen_files", cascade="all,delete")
+        "Submission", backref=backref("seen_files", lazy="dynamic", cascade="all,delete")
     )
     journalist = relationship("Journalist", backref=backref("seen_files"))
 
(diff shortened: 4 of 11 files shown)


sssoleileraaa:
while I start working on tests and scripts for development/ qa, someone could start looking at my test plan and taking it for a spin (it's long, so I might have missed something, but it should be ready for a first pass)

securedrop/journalist_app/utils.py
```diff
@@ -165,10 +166,25 @@ def download(zip_basename: str, submissions: List[Union[Submission, Reply]]) ->
...
+                db.session.add(seen_file)
+                item.downloaded = True
+            else:
+                seen_message = SeenMessage(message_id=item.id, journalist_id=journalist_id)
+                db.session.add(seen_message)
+                item.downloaded = True
+
+            db.session.commit()
+        except IntegrityError:
+            pass  # expected not to store that a file was seen by the same user multiple times
```
creviera, line 187: i'm super confused as to why this IntegrityError handling works for single files/messages/replies when i download the same file many times, but somehow this doesn't catch the exception for bulk downloads :/

securedrop/journalist_app/utils.py
```diff
@@ -9,15 +9,16 @@
 from flask import (g, flash, current_app, abort, send_file, redirect, url_for,
                    render_template, Markup, sessions, request)
 from flask_babel import gettext, ngettext
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.sql.expression import false
```
eloquence: (no longer used in this file, hence lint error)

eloquence:
- [x] verify 'seen_replies', 'seen_files', and 'seen_messages' exist

```
sqlite> .schema seen_files
CREATE TABLE seen_files (
	id INTEGER NOT NULL, 
	file_id INTEGER NOT NULL, 
	journalist_id INTEGER, 
	PRIMARY KEY (id), 
	UNIQUE (file_id, journalist_id), 
	FOREIGN KEY(file_id) REFERENCES submissions (id), 
	FOREIGN KEY(journalist_id) REFERENCES journalists (id)
);
sqlite> .schema seen_replies
CREATE TABLE seen_replies (
	id INTEGER NOT NULL, 
	reply_id INTEGER NOT NULL, 
	journalist_id INTEGER, 
	PRIMARY KEY (id), 
	UNIQUE (reply_id, journalist_id), 
	FOREIGN KEY(reply_id) REFERENCES replies (id), 
	FOREIGN KEY(journalist_id) REFERENCES journalists (id)
);
sqlite> .schema seen_messages
CREATE TABLE seen_messages (
	id INTEGER NOT NULL, 
	message_id INTEGER NOT NULL, 
	journalist_id INTEGER, 
	PRIMARY KEY (id), 
	UNIQUE (message_id, journalist_id), 
	FOREIGN KEY(message_id) REFERENCES submissions (id), 
	FOREIGN KEY(journalist_id) REFERENCES journalists (id)
```

-  Log in from JI as dellsberg and download a message
- [x] verify record exists in seen_messages table
- Log in from JI as test-user and download the same message
- [x] verify new record exists in seen_messages table
- Log in as journalist and delete dellsberg and test-user
- [x] verify journalist_id is now null for both records
7. Check unique constraints for non-nulls
- `insert into seen_files (file_id, journalist_id) values (1, 1);` twice
- [x] verify constraint does not allow this
- repeat for 'seen_messages' and 'seen_replies'
- Send a file attachment from the source interface
- Download it from the JI
- [x] verify seen record exists for this file
- Now delete the file from the JI
- [x] verify record is deleted in the seen_files table

securedrop/models.py
```diff
@@ -141,139 +195,154 @@ def public_key(self, value: str) -> None:
...
         return json_submission
 
+    @property
+    def seen(self) -> bool:
+        """
+        If a submission has been downloaded or seen by any journalist, then the submssion is
+        considered seen.
+        """
+        seen_file = SeenFile.query.filter_by(file_id=self.id).all()
+        seen_message = SeenMessage.query.filter_by(message_id=self.id).all()
```
eloquence: <s>Do we technically need `.all()` here (for both queries) or would `.one()` suffice?</s> Never mind, given the exception behavior of .one() that probably doesn't make sense.

securedrop/tests/test_journalist.py
```diff
@@ -2095,11 +2095,6 @@ def test_download_all_selected_sources(journalist_app, test_journo):
             data=dict(action='download-all',
                       cols_selected=[bulk['source1'].filesystem_id]))
 
-        resp = app.post(
-            url_for('col.process'),
-            data=dict(action='download-all',
-                      cols_selected=[bulk['source1'].filesystem_id]))
-
```
eloquence, line 2102: Nice catch of the duplication there :)

eloquence:
- Click on the "<x> unread" link for the source at the top of the list
- [x] verify JI works the same as before
- [x] verify the correct entries were added to the 'seen_messages' table for submissions
- [x] verify 'downloaded' was not set in the db
- Repeat for files
- Send a file, message, and reply
- Select all conversation items for this source and click "Download Selected"
- [x] verify JI works the same as before
- [x] verify the correct entries were added to the 'seen_*' table
- [x] verify seen-* table entries added for each file, reply, and message.
- [x] verify 'downloaded' was not set in the db
7. Check filename-link
- Send a file, message, and reply
- Select all conversation items for this source and click "Download Selected"
- [x] verify JI works the same as before
- [x] verify the correct entries were added to the 'seen_*' table
- [x] verify seen-* table entries added for each file, reply, and message.
- [x] verify 'downloaded' was not set in the db
8. Check unique constraint exception handling
- [x] verify that downloading the same file, message, or reply again and again works and that there is only one seen entry for each
9. Check that sending a reply as a journalist creates seen record
- [x] Send a reply to a source
- [x] Verify seen record exists in the seen_replies table

## New behavior in the JI [Change 3]
1. Restart `make dev`
2. Send a file attachment from the source interface as a new source
- [x] 3. Visit JI to confirm that there are unread files and messages
4. Log into the client
- [x] verify JI no longer marks everything as read by refreshing the JI index page
5. Download the file from the client
- [x] verify the file is marked as read in the JI
- [x] verify entry exists in the seen_files table
- [x] verify 'downloaded' was not set in the db
- :x: Downloading a file from the client that's already seen causes a server-side exception due to unique constraint violation, and an error in the client

sssoleileraaa:
Thanks for the review! I add more tests to the test plan based on your feedback:

 Check that you can download a file that was seen via the JI
      - [ ] Download the unread file from the JI
      - [ ] Download the same file from the client
      - [ ] verify download was successful

Check deletion of a source
     - Delete a source with seen records for replies, files, and messages
     - [ ] verify seen records no longer exist

 Check "Download Unread" button
      - Send a file and message
      - Select the source with the unread file and message and click "Download Unread"
      - [ ] verify JI works the same as before
      - Select the source again with no unread files and messages and click "Download Unread"
      - [ ] verify that you get error message "No unread submissions in selected collections."
      - Send a couple files
      - Select all sources in the source list and click "Download Unread"
      - [ ] verify download contains expected content and JI works the same as before

I'll mark this as "Ready" after making the fix and adding a test section around doing a migration as well. While it's ready I might continue to add more unit tests.

sssoleileraaa:
one more rebase since the base `develop` branch changed...

securedrop/alembic/versions/48a75abc0121_add_seen_tables.py
```diff
@@ -0,0 +1,57 @@
+"""add seen tables
```
rmol, line 1: Still would have preferred last week's version of the tables from the spec, with the addition of the global downloaded/seen flag for dealing with undead unread submissions after journalist deletion, but I'm putting that on the record and putting it down. Moving on with my life. It'll be fine. :slightly_smiling_face:

securedrop/journalist_app/main.py
```diff
@@ -74,13 +87,11 @@ def index():
                 starred.append(source)
             else:
                 unstarred.append(source)
-            source.num_unread = len(
-                Submission.query.filter_by(source_id=source.id,
-                                           downloaded=False).all())
+            submissions = Submission.query.filter_by(source_id=source.id).all()
+            unseen_submissions = [s for s in submissions if not s.seen]
+            source.num_unread = len(unseen_submissions)
```
rmol, line 78: I'm sure this calculation could be done in the query, but only if you're curious. I don't think it would save much time.

securedrop/journalist_app/main.py
```diff
@@ -112,11 +123,13 @@ def reply():
...
         )
-        reply = Reply(g.user, g.source, filename)
-
         try:
+            reply = Reply(g.user, g.source, filename)
             db.session.add(reply)
             db.session.commit()
+            seen_reply = SeenReply(reply_id=reply.id, journalist_id=g.user.id)
+            db.session.add(seen_reply)
+            db.session.commit()
```
rmol, line 121: It might be better to make the `commit` immediately before the `SeenReply` is constructed a `flush` instead, so if there's  a problem creating the `SeenReply`, the reply won't have been committed.

securedrop/models.py
```diff
@@ -141,139 +194,154 @@ def public_key(self, value: str) -> None:
...
         }
         return json_submission
 
+    @property
+    def seen(self) -> bool:
+        """
+        If a submission has been downloaded or seen by any journalist, then the submssion is
+        considered seen.
+        """
+        seen_file = SeenFile.query.filter_by(file_id=self.id).all()
```
rmol: Could you use the backreferences (`.seen_files`, `.seen_messages`) here?

rmol (changes requested):
First pass looks good. Had a few suggestions, no blockers though.

securedrop/journalist_app/main.py
sssoleileraaa: you're right, i can be done as part of the query. i switched to doing it this way after getting an `sqlalchemy.exc.StatementError` when trying to do it this way:

```py
            unseen_submissions = Submission.query.filter_by(source_id=source.id).filter(Submission.seen == false()).all()
```

but now realizing that's because "seen" is not an sql boolean (false() is imported from sqlalchemy.sql.expression) because it's not a column, just a property. so i can just to this instead:

```py
            unseen_submissions = Submission.query.filter_by(source_id=source.id).filter(not Submission.seen).all()
```

securedrop/alembic/versions/48a75abc0121_add_seen_tables.py
sssoleileraaa: hah! it's not a bad idea. this pr adds a `seen` property to the `submissions` table for dealing with historical data where 'downloaded' was once set by users who may still have accounts or they may not. your idea could work with very little change in addition to what's in this PR if we:

  * started setting 'downloaded' (which would need to be renamed to something that describes that it means 'seen by at least one person') the first time a file/message/reply is seen (i don't think it's necessary to wait until a journalist is deleted to start setting it)
  * added a `downloaded` column (which would have the same brilliant name that indicates that it means 'seen by at least one person') and a `seen` property (similar behavior to the `seen` property on the `submissions` table) to the `replies` table

maybe we'll end up doing this if we decide to cascade delete records in the `seen` tables when a journalist is deleted (i'm glad you added your comment here so we can refer back to this when we start to talk journalist-deletion management more). also remember how we're going to need to prepare for enabling sqlite foreign key support, which means we can no longer keep updating replies from deleted journalists to have a `journalist_id` of null. this has implications on how we decide to manage deleted journalists. we may implement https://github.com/freedomofpress/securedrop/issues/5467 and then problem solved. or we may need to re-associate journalist data to an anonymized deleted account. this is all tbd.

securedrop/models.py
sssoleileraaa: forgot to respond to this. we would need to add backrefs here to both tables so that they're two way i think. not sure i see the benefit but do you have a design suggestion in mind? if so, i can take a look and update this tomorrow. i'm trying to keep the commits clean so i'll need to do some rebasing and commit shuffling tomorrow if we decide to make another change. speaking of, let's not squash the commits before merge because they are distinct units which adds clarity to this feature's history.

securedrop/alembic/versions/48a75abc0121_add_seen_tables.py
rmol: We can enable real foreign key handling with the current schema, as you've declared `journalist_id` nullable in the seen tables:
```
sqlite> PRAGMA foreign_keys = ON;
sqlite> insert into seen_replies (reply_id) values (1);
sqlite> select * from seen_replies;
1|1|
```

securedrop/journalist_app/main.py
rmol: You're getting tripped up by the leaky abstraction: `not Submission.seen` just evaluates to `false` before getting passed to SQLAlchemy, so this is always returning nothing. The generated query looks like this:

```
SELECT submissions.id, submissions.uuid, submissions.source_id, submissions.filename, submissions.size, submissions.downloaded, submissions.checksum 
FROM submissions 
WHERE false
```

Which returns no unseen submissions, so the sources are all always read.

It would need something like:
```
source.num_unread = (
    Submission.query.filter_by(source_id=source.id)
    .filter(
        and_(
            ~exists().where(SeenFile.file_id == Submission.id),
            ~exists().where(SeenMessage.message_id == Submission.id)
        )
    )
    .count()
)
```

where the query evaluates to:

```
SELECT submissions.id, submissions.uuid, submissions.source_id, submissions.filename, submissions.size, submissions.downloaded, submissions.checksum 
FROM submissions 
WHERE submissions.source_id = :source_id_1 AND NOT (EXISTS (SELECT * 
FROM seen_files 
WHERE seen_files.file_id = submissions.id)) AND NOT (EXISTS (SELECT * 
FROM seen_messages 
WHERE seen_messages.message_id = submissions.id))
```

If you want to push through this, you can use `current_app.logger.error("QUERY: %s", unseen_submissions.statement)` to see what SQLAlchemy is generating.

securedrop/models.py
rmol: You already have the backrefs in the seen tables, e.g. `SeenFile.file`. Right now you could use `if self.downloaded or self.seen_files or self.seen_messages:`, and if you add `lazy="dynamic"` to the `relationship` calls in the seen tables, you could use `if self.downloaded or self.seen_files.count() or self.seen_messages.count():`

securedrop/journalist_app/main.py
rmol: If we want to get too clever by half, we could add methods that return filter expressions: 
```
    @classmethod
    def unseen(cls):
        """
        If a submission has not been downloaded or seen by any journalist, then the submission is considered unseen.
        """
        return and_(
            ~exists().where(SeenFile.file_id == Submission.id),
            ~exists().where(SeenMessage.message_id == Submission.id)
        )
```

used like this:
```
        unseen_submissions += (
            Submission.query.filter_by(source_id=source.id)
            .filter(Submission.unseen())
            .all()
        )
```

Tricksy, but could reduce repetition.

sssoleileraaa:
one of the challenges here is that unit testing migrations seems to always pass locally so i've been pushing changes to this pr in order to see where the tests fail (not the most efficient method, so i'll look into updating https://docs.securedrop.org/en/stable/development/database_migrations.html?highlight=alembic#unit-testing-migrations once i figure out what steps are required to get migration tests to work locally)

securedrop/journalist_app/main.py
sssoleileraaa: since we're code freezing soon, let's go back to the way i had it originally

creviera:
> First pass looks good. Had a few suggestions, no blockers though.

sssoleileraaa:
> First pass looks good. Had a few suggestions, no blockers though.

Your comments have been addressed so this is ready for your :eye: :eye: again. Switching over to do an early review of https://github.com/freedomofpress/securedrop/pull/5513

securedrop/tests/migrations/migration_48a75abc0121.py
```diff
@@ -0,0 +1,383 @@
...
+            sql = "SELECT * FROM submissions"
+            submissions = db.engine.execute(text(sql)).fetchall()
+
+            sql = "SELECT * FROM replies"
+            replies = db.engine.execute(text(sql)).fetchall()
+
+            # Now seen tables exist, so you should be able to mark some files, messages, and replies
+            # as seen
+            for submission in submissions:
+                if submission.filename.endswith('-doc.gz.gpg') and random.choice([0, 1]):
```
rmol, line 261: These `random.choice` gates could mean you don't end up marking anything as seen, if you get really unlucky. 

It looks like the mark-as-seen helper functions don't do anything that you're not testing immediately after, when checking for unique constraint violations, so do we need these `for` loops and the helper methods at all?

securedrop/tests/migrations/migration_48a75abc0121.py
```diff
@@ -0,0 +1,383 @@
...
+                    self.mark_file_as_seen(submission.id, selected_journo_id)
+                elif random.choice([0, 1]):
+                    selected_journo_id = random.randint(0, self.JOURNO_NUM)
+                    self.mark_message_as_seen(submission.id, selected_journo_id)
+            for reply in replies:
+                if random.choice([0, 1]):
+                    selected_journo_id = random.randint(0, self.JOURNO_NUM)
+                    self.mark_reply_as_seen(reply.id, selected_journo_id)
+
+    def check_downgrade(self):
```
rmol, line 369: You're not using anything added in `load_data`, are you? If not, you could save the time spent populating the tables you're about to delete.

sssoleileraaa: the idea is to load a bunch of data with seen records and then downgrade to make sure that it goes smoothly even when there are seen records

securedrop/tests/migrations/migration_48a75abc0121.py
sssoleileraaa: yup, in the comment is says: `Now seen tables exist, so you should be able to mark some files, messages, and replies as seen`. i just wanted to mark some files, messages, and replies as seen (not all in th db)

> It looks like the mark-as-seen helper functions don't do anything that you're not testing immediately after, when checking for unique constraint violations, so do we need these for loops and the helper methods at all?

the helper functions are to make it easier to create journalists, submissions, seen records, etc. and to read what's going on. the for loop is to loop through all submissions so that we can make sure we are able to add seen records once there are seen tables

securedrop/tests/migrations/migration_48a75abc0121.py
rmol: I don't think it would matter, until we enable proper foreign key behavior (more [here](https://sqlite.org/lang_droptable.html)), at which time we'd need to do more than just drop the tables in the downgrade. But it's fine, just leave it.

securedrop/tests/migrations/migration_48a75abc0121.py
rmol: I understand what you're trying to do, but it might not be happening _at all_ in the for loops, because of the `random.choice` calls. You're not verifying the results of the for loops, and you're testing the same thing, the basic ability to insert to the seen tables, immediately below during the unique constraint checks. 

I would either remove the for loops altogether, or be more deliberate about the subset of records you mark seen, and then test that the inserted seen table records are as expected.

The helper functions are fine; you could also use them in the unique constraint testing.

rmol: You know, I just realized that because you are testing the insertion during the unique constraint check, it doesn't matter at all if the `random.choice` calls skip all that work. It's actually better, because it could save time whenever this migration is run, which is a lot in dev environments.

rmol (approved):
## New tables and account deletion [Change 1]
0. Make sure alembic upgrade and downgrade work as expected (you can just check that tests pass as well
1. Check out this branch
2. `make dev`
3. `docker exec -it securedrop-dev-0 bash`
4. `sqlite3 /var/lib/securedrop/db.sqlite`
5. Check existence of tables
     - [x] verify 'seen_replies', 'seen_files', and 'seen_messages' exist
6. Check deletion of a journalist
     -  Log in from JI as dellsberg and download a message
     - [x] verify record exists in seen_messages table
     - Log in from JI as test-user and download the same message
     - [x] verify new record exists in seen_messages table
     - Log in as journalist and delete dellsberg and test-user
     - [x] verify journalist_id is now null for both records
7. Check unique constraints for non-nulls
     - `insert into seen_files (file_id, journalist_id) values (1, 1);` twice
     - [x] verify constraint does not allow this
     - repeat for 'seen_messages' and 'seen_replies'
8. Check deletion of a file/message/reply
     - Send a file attachment from the source interface
     - Download it from the JI
     - [x] verify seen record exists for this file
     - Now delete the file from the JI
     - [x] verify record is deleted in the seen_files table
     - repeat for replies and messages
9. Check deletion of a source
     - Delete a source with seen records for replies, files, and messages
     - [x] verify seen records no longer exist

## Behavior has not changed in the JI [Change 2]
1. Check out this branch
2. `make dev`
3. `docker exec -it securedrop-dev-0 bash`
4. `sqlite3 /var/lib/securedrop/db.sqlite`
5. Check x-unread link
     - Click on the "<x> unread" link for the source at the top of the list
     - [x] verify JI works the same as before
     - [x] verify the correct entries were added to the 'seen_messages' table for submissions
     - [x] verify 'downloaded' was not set in the db
     - Repeat for files
6. Check "Download Unread" button
     - Send a file and message
     - Select the source with the unread file and message and click "Download Unread"
     - [x] verify JI works the same as before
     - Select the source again with no unread files and messages and click "Download Unread"
     - [x] verify that you get error message "No unread submissions in selected collections."
     - Send a couple files
     - Select all sources in the source list and click "Download Unread"
     - [x] verify JI works the same as before
7. Check "Download Selected" button
     - Send a file, message, and reply
     - Select all conversation items for this source and click "Download Selected"
     - [x] verify JI works the same as before
     - [x] verify the correct entries were added to the 'seen_*' table
     - [x] verify 'downloaded' was not set in the db
8. Check filename-link
     - Send a file
     - Click on the filename of that file from the JI
     - [x] verify JI works the same as before
     - [x] verify the correct entry was added to the 'seen_files' table
     - [x] verify 'downloaded' was not set in the db
     - Repeat for message
9. Check unique constraint exception handling
     - [x] verify that downloading the same file, message, or reply again and again works and that there is only one seen entry for each
10. Check that sending a reply as a journalist creates seen record
     - [x] Send a reply to a source
     - [x] Verify seen record exists in the seen_replies table

## New behavior in the JI [Change 3]
1. Restart `make dev`
2. Send 2 file attachments from the source interface as a new source
3. Visit JI to confirm that there are unread files
4. Log into the client
     - [x] verify no files or messages are marked as read in the JI due to logging into the client
5. Download one file from the client
     - [x] verify the file is not marked as read in the JI
     - [x] verify 'downloaded' was not set in the db
6. Check that you can download a file that was seen via the JI
     - [x] Download the unread file from the JI
     - [x] Download the same file from the client
     - [x] verify download was successful

I did not complete the migration testing as written; @zenmonkeykstop has volunteered to run through our upgrade testing scenario to validate the migrations instead.

zenmonkeykstop:
upgrade testing scenario passed successfully:
- ran `make build-debs && make upgrade-start`
- logged into SI and created a few sources and submissions
- logged into JI and downloaded a subset of submissions
- updated `molecule/upgrade/side-effect.yml` to stop the upgrade resetting the db /shrug
- ran `make upgrade-test-local`
- verified that read/unread states were preserved in JI
- verified that `seen_files` and `seen_messages` tables were created on app server db and are empty
- verified that submissions can still be downloaded
- verified that `seen_*` tables are updated when files downloaded from JI

PR #5506, opened 2020-09-17
Add /users endpoint

### LINKED ISSUE #5490: Add `/users` endpoint to API
In order to reliably update local user databases (see, for example, https://github.com/freedomofpress/securedrop-client/issues/1143), it would be useful for the API to expose a `/users` endpoint.

Currently, non-admin users do not have the ability to see a list of all users via the *Journalist Interface* , so we may only want to return the UUID for each user (and potentially a property like `is_deleted` or `is_locked` if/when such functionality is implemented, cf. #5467).

eloquence:
Resolves #5490
## Description

Adds the `/users` endpoint as described in the issue and subsequent comments, enumerating non-sensitive information for all users.

Documentation PR dependent on this one:
https://github.com/freedomofpress/securedrop-docs/pull/3

## Status

Ready for review.

## Test plan

Spin up a development environment from this PR.

- [ ] Observe that you see this endpoint listed on http://localhost:8081/api/v1
- [ ] Observe that you cannot access this endpoint at http://localhost:8081/api/v1/users without an authentication token
- [ ] Observe that you can access the list of users when providing a valid authentication token as part of the request (see example below)
- [ ] Observe that the list of users only includes the expected fields (UUID, username, first name, last name)

## Token usage example (curl)

```bash
$ curl -X POST -H "Content-Type: application/json" \ 
  -d '{"username":"journalist", "passphrase":"correct horse battery staple profanity oil chewy", "one_time_code": "123456"}' \ 
  'http://localhost:8081/api/v1/token'

$ curl -H "Authorization: Token TOKEN_HERE" http://localhost:8081/api/v1/users
```

## Checklist
- [x] Linting (`make lint`) and tests (`make test`) pass in the development container (tested for API tests)

### DIFF
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -710,15 +710,25 @@ class Journalist(db.Model):
 
         return Journalist.query.get(data['id'])
 
-    def to_json(self) -> 'Dict[str, Union[str, bool, str]]':
+    def to_json(self, all_info: bool = True) -> 'Dict[str, Union[str, bool, str]]':
+        """Returns a JSON representation of the journalist user. If all_info is
+           False, potentially sensitive or extraneous fields are excluded. Note
+           that both representations do NOT include credentials."""
+
         json_user = {
             'username': self.username,
-            'last_login': self.last_access.isoformat() + 'Z',
-            'is_admin': self.is_admin,
             'uuid': self.uuid,
             'first_name': self.first_name,
             'last_name': self.last_name
         }
+
+        if all_info is True:
+            json_user['is_admin'] = self.is_admin
+            try:
+                json_user['last_login'] = self.last_access.isoformat() + 'Z'
+            except AttributeError:
+                json_user['last_login'] = None
+
         return json_user
 
 
--- a/securedrop/journalist_app/api.py
+++ b/securedrop/journalist_app/api.py
@@ -66,6 +66,7 @@ def make_blueprint(config: SDConfig) -> Blueprint:
     def get_endpoints() -> Tuple[flask.Response, int]:
         endpoints = {'sources_url': '/api/v1/sources',
                      'current_user_url': '/api/v1/user',
+                     'all_users_url': '/api/v1/users',
                      'submissions_url': '/api/v1/submissions',
                      'replies_url': '/api/v1/replies',
                      'auth_token_url': '/api/v1/token'}
@@ -322,6 +323,13 @@ def make_blueprint(config: SDConfig) -> Blueprint:
     def get_current_user() -> Tuple[flask.Response, int]:
         user = _authenticate_user_from_auth_header(request)
         return jsonify(user.to_json()), 200
+
+    @api.route('/users', methods=['GET'])
+    @token_required
+    def get_all_users() -> Tuple[flask.Response, int]:
+        users = Journalist.query.all()
+        return jsonify(
+            {'users': [user.to_json(all_info=False) for user in users]}), 200
 
     @api.route('/logout', methods=['POST'])
     @token_required
--- a/securedrop/tests/test_journalist_api.py
+++ b/securedrop/tests/test_journalist_api.py
@@ -22,9 +22,9 @@ def test_unauthenticated_user_gets_all_endpoints(journalist_app):
     with journalist_app.test_client() as app:
         response = app.get(url_for('api.get_endpoints'))
 
-        expected_endpoints = ['current_user_url', 'submissions_url',
-                              'sources_url', 'auth_token_url',
-                              'replies_url']
+        expected_endpoints = ['current_user_url', 'all_users_url',
+                              'submissions_url', 'sources_url',
+                              'auth_token_url', 'replies_url']
         expected_endpoints.sort()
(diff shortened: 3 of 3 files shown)


securedrop/models.py
```diff
@@ -710,15 +710,27 @@ def validate_api_token_and_get_user(token: str) -> 'Union[Journalist, None]':
...
         return Journalist.query.get(data['id'])
 
-    def to_json(self) -> 'Dict[str, Union[str, bool, str]]':
+    def to_json(self, bare: bool = False) -> 'Dict[str, Union[str, bool, str]]':
+        """Returns a JSON representation of the journalist user. If a bare
+           representation is requested, potentially sensitive or extraneous
+           fields are excluded."""
+
+        try:
+            last_login = self.last_access.isoformat() + 'Z'
```
eloquence: When I first attempted to request a `to_json` representation of all users in the dev env, this raised an exception due to the `last_access` attribute never having been set. While not required for the bare representation, it seemed sensible to add a check here.

docs/development/journalist_api.rst
```diff
@@ -675,3 +675,37 @@ Response 200:
...
+Response 200:
+
+.. code:: sh
+
+  {
+    "users": [
+      {
+        "first_name": "Nellie",
+        "last_name": "Bly",
+        "username": "nbly",
```
sssoleileraaa: <3

docs/development/journalist_api.rst
```diff
@@ -675,3 +675,37 @@ Response 200:
...
+        "uuid": "2b3f05ef-3695-4522-88bd-f124d2e89d01"
+      },
+      {
+        "first_name": "Daniel",
+        "last_name": "Ellsberg",
+        "username": "dellsberg",
```
sssoleileraaa: a-ha! now i get the whole "dellsberg" reference in the test data

securedrop/models.py
sssoleileraaa: [nit/feel free to ignore] you could set last_login within the bare clause where bare is false since that's the only place it's used

[suggestion] instead of "bare" you could call it "all_details" or "all_info"

sssoleileraaa: i think it makes sense to also add an "all_details" check that defaults to false for the "/user" endpoint. otherwise, we'll have some admins with `is_admin` set and some admins without `is_admin` set depending on who has logged into the workstation. looking at the table directly as an admin could be misleading. also, we don't do anything with this information in the client and should probably wait to include it once we finish our client admin stories.

eloquence: >  you could set last_login within the bare clause where `bare` is false since that's the only place it's used

That makes sense; we'll still have to do the `try` check to avoid errors on newly created users, but I'll move it.

> instead of "bare" you could call it "all_details" or "all_info"

I like `all_info`, will use that. I was trying to avoid suggesting that it's a _full_ representation, since both representations exclude credentials.

eloquence: > also add an "all_details" check that defaults to false for the "/user" endpoint

Could you clarify? This PR doesn't modify the behavior of the `/user` endpoint (it currently already includes the `is_admin` Boolean), do you think that it should?

eloquence: Made the other two changes in aa0a5914e5dc55e6c936e5bd2470d37f32f83122

sssoleileraaa: > Could you clarify? This PR doesn't modify the behavior of the /user endpoint (it currently already includes the is_admin Boolean), do you think that it should?

Oh, I see, I forgot that we don't store the `is_admin` result from the `/user` endpoint so we don't have to worry about the situation I brought up where some admin acounts will have `is_admin` set and some will not. Okay, ignore previous comment.

sssoleileraaa (dismissed):
alright this looks good to me. I think it would be good followup to add the `all_info` check to the "/user" endpoint eventually for consistency. plus it's easy to get confused about how we're adding `is_admin` to the response body yet never using or storing it.

looks like lint is failing, so once that's fixed I'll approve

sssoleileraaa (approved):
lgtm
- [x] Observe that you see this endpoint listed on http://localhost:8081/api/v1
- [x] Observe that you cannot access this endpoint at http://localhost:8081/api/v1/users without an authentication token
- [x] Observe that you can access the list of users when providing a valid authentication token as part of the request (see example below)
- [x] Observe that the list of users only includes the expected fields (UUID, username, first name, last name)

conorsch (approved):
Looks great! @creviera already reviewed, but I'll add further confirmation:

- [x] Observe that you see this endpoint listed on http://localhost:8081/api/v1
- [x] Observe that you cannot access this endpoint at http://localhost:8081/api/v1/users without an authentication token
- [x] Observe that you can access the list of users when providing a valid authentication token as part of the request 
- [x] Observe that the list of users only includes the expected fields (UUID, username, first name, last name) - The the first/last fields were present, although empty, which I'd expect given the create-dev-data logic.

Not blocking, but let's look for an opportunity to test the "last_login" field later on.

conorsch:
Proceeding with merge. There's a bit more review required for #5505. Once both #5505 & #5506 are in, we'll rebase and review #5513.

PR #5513, opened 2020-09-17
API changes for listing/marking source conversation items that have been seen by journalists

### LINKED ISSUE #5475: Add API endpoints for `seen_by`
As part of adding SecureDrop Client support for highlighting sources with unseen submissions (https://github.com/freedomofpress/securedrop-client/issues/187), we need to make server-side changes to record which journalist has seen a message, file, or reply. This breaks down into two main tasks:

- issue #5474: migrating the existing `downloaded` status and business logic to `seen_by` (for messages, files, and replies);

- this issue: adding API endpoints for updating and retrieving `seen_by` status

## Technical specification

A working draft specification can be found here:
https://docs.google.com/document/d/1szPEKI1gTpRiM_YI87plKgiW39T7zT4Y4HG0uFnsJD4/edit#

## Functional requirements

**Setting status via API:** As an API user, I should be able to set the `seen_by` status for a file, message, or reply to the authenticated user.

**Retrieving status via API:** As an API user, I should be able to retrieve the `seen_by` status for any message, file or reply.

## Performance requirements

We want to keep API consumers like the SecureDrop Client responsive with a large number of sources and submissions per source:

https://github.com/freedomofpress/securedrop-client/wiki/Scalability-&-Performance#scalability-requirements

Our performance goals may impact decisions such as how many API roundtrips are required to update or retrieve information, how much data is returned, and what operations can be grouped together. As part of this issue, we should aim to obtain before/after benchmarks e.g. of metadata sync operations for a server with a large number of submissions, so we can quantify the performance impact.

rmol:
## Status

Ready for review

## Description of Changes
    
Add new API endpoint for listing or marking source conversation items that have been seen by a journalist.
    
Add utility method to mark a heterogeneous collection of Submission and Reply objects seen.
    
Add Submission.is_file and Submission.is_message to encapsulate the characterization based on filename.

Fixes #5475.

## Testing

- Run the dev server with `make dev`.

- Get an API token as the `journalist` user:
  ```
  export TOKEN=$(curl -o- -X POST -H "Content-Type: application/json" --data '{"username":"journalist","passphrase":"correct horse battery staple profanity oil chewy","one_time_code":"047889"}' http://localhost:8081/api/v1/token | jq -r .token)
  ```

- List current submissions: 
  ```
  curl -X GET -H "Content-Type: application/json" -H "Authorization: Token $TOKEN" http://localhost:8081/api/v1/submissions
  ```
  All should have a `seen_by` property, currently an empty list.

- Mark a message seen:
  ```
  curl -X POST -H "Content-Type: application/json" -H "Authorization: Token $TOKEN" --data '{"messages": ["e711d29d-d0a1-440f-80e4-6642b77ec3ea"]}' http://localhost:8081/api/v1/seen
  ```
  Replace the UUID with one from one of the messages in the response from the `/submissions` endpoint, of course.
- Mark the same message seen again. There should be no error.
- Retrieve `/submissions` again. The message's `seen_by` list should now contain your journalist account's UUID.
- Visit the source of that message in the journalist interface. The message you marked should not be listed in bold text. Clicking "Select unread" should not select that message.
- Visit the source list in the journalist interface. Select the source of the message you marked and click "Download Unread". The zip file you receive should not contain the seen message.

- Via the source interface, upload a file, then retrieve `/submissions` again, and mark the file seen:
  ```
  curl -X POST -H "Content-Type: application/json" -H "Authorization: Token $TOKEN" --data '{"files": ["4fe4b7f5-3031-4651-9092-888951681cf7"]}' http://localhost:8081/api/v1/seen
  ```
  Replace the UUID with that of the file you just uploaded, of course.
- Repeat the checks in the journalist interface that you performed for a seen message, confirming that the file you marked seen is always considered read.

- Get another API token, this time as the `dellsberg` journalist account.
- Retrieve the list of all replies:
  ```
  curl -X GET -H "Content-Type: application/json" -H "Authorization: Token $TOKEN" http://localhost:8081/api/v1/replies
  ```
- Mark one of them read: 
  ```
  curl -X POST -H "Content-Type: application/json" -H "Authorization: Token $TOKEN" --data '{"replies": ["8924a691-3b0f-45c1-bfed-b0b91e264855"]}' http://localhost:8081/api/v1/seen
  ```
- Hit `/replies` again and confirm that the `dellsberg` account UUID appears in the reply's `seen_by` list.

## Deployment

Depends on the database changes in #5505.

### DIFF
--- a/securedrop/create-dev-data.py
+++ b/securedrop/create-dev-data.py
@@ -15,7 +15,7 @@ import journalist_app
 
 from sdconfig import config
 from db import db
-from models import Journalist, Reply, Source, Submission
+from models import Journalist, Reply, SeenReply, Source, Submission
 from specialstrings import strings
 
 
@@ -138,6 +138,9 @@ def create_source_and_submissions(
             journalist = journalist_who_replied
         reply = Reply(journalist, source, fname)
         db.session.add(reply)
+        db.session.flush()
+        seen_reply = SeenReply(reply_id=reply.id, journalist_id=journalist.id)
+        db.session.add(seen_reply)
 
     db.session.commit()
 
--- a/securedrop/journalist_app/api.py
+++ b/securedrop/journalist_app/api.py
@@ -16,8 +16,7 @@ from werkzeug.exceptions import default_exceptions
 
 from db import db
 from journalist_app import utils
-from models import (Journalist, Reply, SeenFile, SeenMessage,
-                    SeenReply, Source, Submission,
+from models import (Journalist, Reply, SeenReply, Source, Submission,
                     LoginThrottledException, InvalidUsernameException,
                     BadTokenException, WrongPasswordException)
 from sdconfig import SDConfig
--- a/securedrop/journalist_app/api.py
+++ b/securedrop/journalist_app/api.py
@@ -317,44 +317,13 @@ def make_blueprint(config: SDConfig) -> Blueprint:
         return jsonify(
             {'replies': [reply.to_json() for reply in replies if reply.source]}), 200
 
-    @api.route("/seen", methods=["GET", "POST"])
+    @api.route("/seen", methods=["POST"])
     @token_required
     def seen() -> Tuple[flask.Response, int]:
         """
         Lists or marks the source conversation items that the journalist has seen.
         """
         user = _authenticate_user_from_auth_header(request)
-
-        if request.method == "GET":
-            seen_files = [
-                {
-                    "file_uuid": f.file.uuid,
-                    "journalist_uuid": f.journalist.uuid if f.journalist_id else None
-                }
-                for f in SeenFile.query.all()
-            ]
-            seen_messages = [
-                {
-                    "message_uuid": f.message.uuid,
-                    "journalist_uuid": f.journalist.uuid if f.journalist_id else None
-                }
-                for f in SeenMessage.query.all()
-            ]
-            seen_replies = [
-                {
-                    "reply_uuid": f.reply.uuid,
-                    "journalist_uuid": f.journalist.uuid if f.journalist_id else None
-                }
-                for f in SeenReply.query.all()
-            ]
-
-            return jsonify(
-                {
-                    "files": seen_files,
-                    "messages": seen_messages,
-                    "replies": seen_replies,
-                }
-            ), 200
 
         if request.method == "POST":
             if request.json is None or not isinstance(request.json, collections.abc.Mapping):
--- a/securedrop/journalist_app/api.py
+++ b/securedrop/journalist_app/api.py
@@ -1,7 +1,8 @@
+import collections.abc
 import json
 
 from datetime import datetime, timedelta
-from typing import Tuple, Callable, Any
+from typing import Tuple, Callable, Any, Set, Union
 
 import flask
 import werkzeug
@@ -15,7 +16,8 @@ from werkzeug.exceptions import default_exceptions
 
 from db import db
 from journalist_app import utils
-from models import (Journalist, Reply, Source, Submission,
+from models import (Journalist, Reply, SeenFile, SeenMessage,
+                    SeenReply, Source, Submission,
                     LoginThrottledException, InvalidUsernameException,
                     BadTokenException, WrongPasswordException)
 from sdconfig import SDConfig
@@ -69,6 +71,7 @@ def make_blueprint(config: SDConfig) -> Blueprint:
                      'all_users_url': '/api/v1/users',
                      'submissions_url': '/api/v1/submissions',
                      'replies_url': '/api/v1/replies',
+                     'seen_url': '/api/v1/seen',
                      'auth_token_url': '/api/v1/token'}
         return jsonify(endpoints), 200
 
@@ -268,6 +271,9 @@ def make_blueprint(config: SDConfig) -> Blueprint:
 
             try:
                 db.session.add(reply)
+                db.session.flush()
+                seen_reply = SeenReply(reply_id=reply.id, journalist_id=user.id)
+                db.session.add(seen_reply)
                 db.session.add(source)
                 db.session.commit()
             except IntegrityError as e:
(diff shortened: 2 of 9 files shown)


securedrop/create-dev-data.py
```diff
@@ -138,6 +138,9 @@ def create_source_and_submissions(
             journalist = journalist_who_replied
         reply = Reply(journalist, source, fname)
         db.session.add(reply)
+        db.session.flush()
+        seen_reply = SeenReply(reply_id=reply.id, journalist_id=journalist.id)
```
sssoleileraaa, line 142: nice change because if you were to do this from the JI this would happen automatically

securedrop/journalist_app/api.py
```diff
@@ -309,6 +314,57 @@ def get_all_replies():
...
 
+    @api.route('/seen', methods=['GET', 'POST'])
+    @token_required
+    def seen():
+        """
+        Lists or marks the source conversation items that the journalist has seen.
+        """
+        user = get_user_object(request)
+
+        if request.method == 'GET':
```
sssoleileraaa: We discussed in [the spec](https://docs.google.com/document/d/1szPEKI1gTpRiM_YI87plKgiW39T7zT4Y4HG0uFnsJD4)two possible implementations for this so was wondering why you went in this direction (perhaps for performance reasons as described in https://github.com/freedomofpress/securedrop/issues/5475?)

sssoleileraaa (commented):
did a first pass

sssoleileraaa:
Not sure how the API is going to look once you're done, but just as a reminder, we were also considering updating the current replies and submissions endpoints as an option:

>[Option 2] Update current endpoints

```yml
"replies": [
        {
            "filename": <filename>,
            ...
            "seen_by": [<journalist_uuid>, ]
        },
        {
            "filename": <filename>,
            ...
            "seen_by": [<journalist_uuid>, ]
        },
]

"submissions": [
        {
            "is_read": <is_read>,
            "filename": <filename>,
            ...
            "seen_by": [<journalist_uuid>, ]
            "opened_by": [<journalist_uuid>, ]
        },
        {
            "is_read": <is_read>,
            "filename": <filename>,
            ...
            "seen_by": [<journalist_uuid>, ]
            "opened_by": [<journalist_uuid>, ]
        },
]
```

sssoleileraaa:
just a thought, we could rename "is_read" to "downloaded_from_web" ~or "seen_from_web"~

rmol:
> Not sure how the API is going to look once you're done, but just as a reminder, we were also considering updating the current replies and submissions endpoints as an option

That is already present, as suggested in the test plan. Look at the `to_json` methods of [`Submission`](https://github.com/freedomofpress/securedrop/blob/078a022392fd7ce423345cdcf12fd507c38ae47b/securedrop/models.py#L307) and [`Reply`](https://github.com/freedomofpress/securedrop/blob/078a022392fd7ce423345cdcf12fd507c38ae47b/securedrop/models.py#L379) in `models.py`.

rmol:
> just a thought, we could rename "is_read" to "downloaded_from_web"

I know we've been playing pretty fast and loose with the API and versioning (at least it's mostly been additions :neutral_face: ), but `is_read` is in production. It's possible that someone is using the API for something other than the workstation, so we should at least pretend we care about API stability.

securedrop/journalist_app/api.py
rmol: Both are implemented here (though I just realized `/seen`'s `GET` is incomplete, as it's only returning the authenticated client's seen items -- will fix when I rebase to catch up to the latest version of #5505).

sssoleileraaa: okay, I see that you added seen records in the /replies and /submissions endpoints :+1:. why did you also add a /seen endpoint to get all seen records for the currently logged in journalist. We never discussed doing this since it's not in scope for the seen/unseen feature - we did discuss providing **all** seen records in a new endpoint as an alternative option to updating the /replies and /submissions endpoint. or am i misunderstanding your change? it could be useful to start a docs draft pr similar to https://github.com/freedomofpress/securedrop-docs/pull/3 to provide clarity for reviewers.

rmol: In the comment immediately above this, I mentioned that it was incorrect at that point. It has since been changed to match the spec, returning all the records. If you're reviewing locally, make sure to pull frequently. Or just wait; I should be moving this from draft to ready for review shortly.

sssoleileraaa: oops sorry about that. will wait!

securedrop/models.py
```diff
@@ -212,7 +211,23 @@ def __init__(self, source: Source, filename: str) -> None:
     def __repr__(self) -> str:
         return '<Submission %r>' % (self.filename)
 
-    def to_json(self) -> 'Dict[str, Union[str, int, bool]]':
+    @property
+    def is_file(self) -> bool:
```
sssoleileraaa, line 215: thought about doing this myself in #5505 but the pr was already so large. nice change.

securedrop/models.py
```diff
@@ -212,7 +211,23 @@ def __init__(self, source: Source, filename: str) -> None:
...
+        return self.filename.endswith("doc.gz.gpg") or self.filename.endswith("doc.zip.gpg")
+
+    @property
+    def is_message(self) -> bool:
+        return self.filename.endswith("msg.gpg")
+
+    def to_json(self) -> "Dict[str, Union[str, int, bool]]":
+        seen_by = {
+            f.journalist.uuid for f in SeenFile.query.filter(SeenFile.file_id == self.id)
+            if f.journalist
```
sssoleileraaa, line 225: i see, so this won't include the fact that there is a deleted journalist who saw a file. gonna keep reviewing and will cycle back to this. could be an issue because we still send "deleted" uuids for replies when journalists are deleted. this takes a different course.

sssoleileraaa (commented):
first pass through the code - going to start manually testing

securedrop/models.py
rmol: It does not. I didn't see any way for sensible inclusion of those records in `seen_by` here. We do have `Submission.seen`, which will reflect those, and until we start keeping deleted journalists around, we can't distinguish between the records for the submission in the seen tables anyway, so since the only purpose of those rows at this point is establishing the global seen state, I thought `Submission.seen` the right answer.

conorsch (approved):
Detailed test plan checks out, functionality matches expectations. Visual review of the diff was smooth, as well. As discussion during review indicates, we should queue up some "v2" refinements for the API.

securedrop/models.py
sssoleileraaa: There are also replies we have to worry about which don't have a global seen state, but I definitely see why you made this choice. Agreed that implementing https://github.com/freedomofpress/securedrop/issues/5467 + global user account for data migration will make this work fine.

rmol: Replies do have a global seen state: if one exists, its creator has seen it. `SeenReply` can tell us who else has seen it, but only if  `journalist_id` is not null, so again, I didn't see any point in including records where it is.

Issue #5507, opened 2020-09-17
Marking grsec kernel as default in Focal staging


kushaldas:
## Description

Currently Focal by default has a kernel with higher version (`5.4.x`) than our grsec kernel (`4.14.x`). We have to mark our grsec kernel as the default kernel for boot.

We can use steps mentioned in 
https://github.com/freedomofpress/securedrop-debian-packaging/blob/main/securedrop-workstation-grsec/debian/postinst#L29-32 for the same.

conorsch:
A few clarifications regarding the current state of Focal kernel support. 

The Focal VMs are already booting into grsecurity-patched kernels, and those kernels persist after the VMs are rebooted. This works because in install logic (Ansible playbooks), the grsec kernel is forced for the subsequent boot, and post reboot, we remove all non-grsec kernels from the host, so they can't be updated and take precedence in the boot menu. See for example:

https://github.com/freedomofpress/securedrop/blob/f81e41539bc5e5ae8b9a443ead719513622553c8/install_files/ansible-base/roles/grsecurity/tasks/from_fpf_repo_install_grsec.yml#L60-L61

That's not quite as elegant as the explicit version pinning via metapackage that we employ in the Workstation (https://github.com/freedomofpress/securedrop-debian-packaging/blob/05eedc6e2a5d9a1c53ebb6dda0a2188365063e05/securedrop-workstation-grsec/debian/postinst#L26-L32), and switching to that setup for the servers would make things a lot more predictable, and also enable us to delete a bunch of Ansible tasks. 

We'll likely want to split the metapackage by distro, however, so that we can evaluate newer kernels on Focal (#5479) without also forcing changes on Xenial. Additionally, we'll likely switch to paxctld (#4134), and test-driving that setup on Focal separately from Xenial will help with review. Recall that in #5518 we only updated the `paxctl` command in the Ansible tasks—in the metapackage, they're unchanged. That's fine for now, but it means that a Focal machine with the metapackage installed would fail on kernel upgrade, because the wrong header style would be used to set flags on the grub binaries, causing the package upgrade to abort.

Let's take a closer look at the "Ansible and molecule changes" section on the #4768 epic and try to map those out more explicitly.

eloquence:
(@emkll and @conorsch will discuss next steps and update the epic.)

conorsch:
[...]

Issue #5495, opened 2020-09-15
Focal staging paxctl command to convert grub binaries are failing


kushaldas:
## Description

When we are doing new Focal staging environment, the `paxctl` task to convert `grub` binaries is failing.

This is task in question https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/grsecurity/tasks/paxctl.yml#L23-L27

## Steps to Reproduce

- [ ] `molecule converge -s libvirt-staging-focal`

## Expected Behavior

This should happen without any error.

## Actual Behavior

Example error

```
[...]

Issue #4134, opened 2019-02-14
Use paxctld to manage all PaX flags in Ubuntu Focal


emkll:
## Description

Initially raised during review of #4114, we should consolidate the PaX flag management logic. We are using both paxctl and paxctld for management of PaX flags, it would be best to maintain a single paxctld.conf file with all flags. It will also simplify some of the install logic, and allow us to more consistenly apply these flags to binaries and persist reboots (as an added feature, it might also help address #3916).

We could also evaluate and reuse what was done as part of the SecureDrop Workstation as an example: https://github.com/freedomofpress/securedrop-debian-packaging/tree/master/securedrop-workstation-svs-disp

### UPDATE 2020-11-16

The following based on a conversation with @conorsch, as part of the current sprint commitments:

The `paxctl` header formats/commands are different in Focal from Xenial, and as a result, some existing paxctl logic from Xenial will not work in Focal. Specifically, updates to the flags for grub binaries (grub-probe, grub-mkdevicemap, grub-script-check) will not work in Focal under the current logic[1]. If grub is updated, PaX flags won't be applied, leading to cascading failures (grub update will fail, kernel update will fail, packages may be held back/broken) 

Based on our conversation, we went through various approaches and decided on the following:
1. Fork securedrop-grsec for Focal only (and leave the existing version for Xenial unchanged, because it works and we aren't supporting Xenial for much longer)
2. Add `paxctld` to the `Requires` field in the Control file. Since we are installing from scratch on Focal, it's a good opportunity to add, but also make 1. easier.
3. Add a securedrop-managed `paxctld.conf` file to the Focal-specific metapackage, to configure the pax flags as we would expect them.
4. Remove kernel postinst file from the metapackage, since use of paxctld makes use of paxctl moot

## User Stories

As a SecureDrop administrator or developer, I would like to manage all PaX flags in a single place and would like to consistently apply them across the system.

[1] https://github.com/freedomofpress/securedrop/blob/develop/install_files/securedrop-grsec/etc/kernel/postinst.d/paxctl-grub

conorsch:
> Fork securedrop-grsec for Focal only

I created a separate issue to track: #5690. In terms of paxctld support, we'll reuse the logic from https://github.com/freedomofpress/securedrop-debian-packaging/tree/cee267e7dfebd9553cdf4b02ecbe54783049121c/securedrop-workstation-grsec/debian for the most part. 

[...]

PR #5518, opened 2020-09-19
Provides first-run custom kernel support for Focal

### LINKED ISSUE #5508: Make sure that all non-grsec kernels are purged duing install on Focal
## Description

This is blocked on  #5507. We have to make sure that no non-grsec kernel is installed on the Focal system.

## Steps to Reproduce

- [ ] `molecule converge -s libvirt-staging-focal`

## Expected Behavior

The command should finish properly.
 
## Actual Behavior

The step for purging all non-grsec kernels will fail as it booted into `5.4.x` series kernel (default in Focal).

## Comments

conorsch:
## Status

Ready for review 

## Description of Changes

Updates the `paxctl` commands to use slightly different logic depending on whether the host is running Xenial or Focal. Notably this _only_ alters the Ansible logic, and ignores the same flags in the `securedrop-grsec` metapackage. Therefore I'm only saying "refs #5495" since technically there's a bit more follow-up required there (possibly switching to paxctld #4134). 

Also updating the package patterns so that non-grsec kernels are succesfully removed during first install. I've only tested in this in libvirt staging VMs running focal, but it's enough. More durable forcing of the hardened kernels is tracked in  #5507. Therefore, closes #5508. 

## Testing

You must use a Linux machine with libvirt support.

```
make build-debs-focal # if you don't already have them
molecule converge -s libvirt-staging-focal
```

Confirm that the install completes without error, and the app is functional. The testinfra tests are not yet passing for Focal, so don't expect the "verify" action to work.

## Deployment

The changes aim to preserve identical behavior under Xenial. Still, the CI checks for staging will confirm that the paxctl commands are not broken under Xenial. I don't think we need additional functional testing beyond that, outside of standard pre-release QA.

Choose one of the following:

### DIFF
--- a/install_files/ansible-base/roles/common/vars/Ubuntu_focal.yml
+++ b/install_files/ansible-base/roles/common/vars/Ubuntu_focal.yml
@@ -2,3 +2,4 @@
 securedrop_kernel_packages_to_remove:
   - linux-virtual
   - linux-generic
+  - 'linux-image-.*generic'
--- a/install_files/ansible-base/roles/grsecurity/defaults/main.yml
+++ b/install_files/ansible-base/roles/grsecurity/defaults/main.yml
@@ -11,3 +11,8 @@ grsec_sysctl_flags:
     # exist otherwise.
   - name: "vm.heap_stack_gap"
     value: "1048576"
+
+# Lookup table for how to create PaX flags, depending on OS.
+paxctl_header_type:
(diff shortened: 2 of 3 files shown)


kushaldas (approved):
This is simple and yet powerful. Works as excepted:

- [x] `molecule converge -s libvirt-staging-focal`

PR #5521, opened 2020-09-22
Fixes #5507 marks grsec kernel as default on Focal

### LINKED ISSUE #5507: Marking grsec kernel as default in Focal staging

kushaldas:
## Status

Ready for review

## Description of Changes

Fixes #5507

Marking the current version of the grsec kernel as the default
on Focal. We will have to update the version string when we releae
new kernel.

## Testing

**NOTE** If I manually install the freshly built `securedrop-grsec` metapackage on the vms, it behaves as it should. But, otherwise it installs the package from the apt repo (instead of the local one). Any tips? @conorsch @emkll ?

- [ ] `make build-debs-focal`
- [ ] `molecule converge -s libvirt-staging-focal`
- [ ] reboot both the vms
- [ ] login and check if both the vms are running `grsec` kernel
- [ ] `molecule destroy -s libvirt-staging-focal`
- [ ] `make build-debs`
- [ ] `molecule converge -s libvirt-staging-xenial`
- [ ] reboot both the vms
- [ ] login and check if both the vms are running `grsec` kernel

## Deployment

2. New installs.

Choose one of the following:

### DIFF
--- a/install_files/securedrop-grsec/DEBIAN/postinst
+++ b/install_files/securedrop-grsec/DEBIAN/postinst
@@ -15,14 +15,21 @@ set -x
 #          <conflicting-package> <version>
 # for details, see http://www.debian.org/doc/debian-policy/ or
 # the debian-policy package
+GRSEC_VERSION='4.14.188-grsec'
+
+# Sets default grub boot parameter to the kernel version specified
+# by $GRSEC_VERSION. The debian buster default kernel is 4.19, thus
+# supersedes this 4.14.x series grsecurity kernel at boot-time
+set_grub_default() {
+    GRUB_OPT="'Advanced options for Debian GNU/Linux>Debian GNU/Linux, with Linux $GRSEC_VERSION'"
+    perl -pi -e "s|^GRUB_DEFAULT=.*|GRUB_DEFAULT=$GRUB_OPT|" /etc/default/grub
+}
 
 case "$1" in
     configure)
 
-    # Replace the default GRUB boot option with 0, which defaults to the
-    # highest kernel version. Any kernel provided by apt.freedom.press must
-    # suprecede the ones provided by Ubuntu.
-    sed -i '/^GRUB_DEFAULT=/s/=.*/=0/' /etc/default/grub
+    # Force latest hardened kernel for next boot
+    set_grub_default
     # When using CONFIG_PAX_KERNEXEC, the grsecurity team recommends the kernel
     # is booted with "noefi" on the kernel command line if "CONFIG_EFI" is
     # enabled, as EFI runtime services are necessarily mapped as RWX.


emkll (changes requested):
Thanks @kushaldas went through the test plan and a experienced a couple of issues:

`GRUB_DEFAULT` value remains unchanged in `/etc/default/grub` when running `molecule converge -s libvirt-staging-xenial`

This will only work on playbook installs, it may be more prudent to apply those changes as part of the metapackage as part of the postinst, similar to how we do it for workstation kernels [1]? What do you think?

[1] https://github.com/freedomofpress/securedrop-debian-packaging/blob/main/securedrop-workstation-grsec/debian/postinst#L29-32

kushaldas:
> Thanks @kushaldas went through the test plan and a experienced a couple of issues:
> 
> `GRUB_DEFAULT` value remains unchanged in `/etc/default/grub` when running `molecule converge -s libvirt-staging-xenial`
> 
Yes, I modified this to be applied only on Focal installations. If we want, I can mark it for both Xenial and Focal.

> This will only work on playbook installs, it may be more prudent to apply those changes as part of the metapackage as part of the postinst, similar to how we do it for workstation kernels [1]? What do you think?
> 
> [1] https://github.com/freedomofpress/securedrop-debian-packaging/blob/main/securedrop-workstation-grsec/debian/postinst#L29-32

Yes, but as that would be on the Kernel metapackage itself, I did not touch it in this.

conorsch:
> This will only work on playbook installs
> Yes, I modified this to be applied only on Focal installations.

If we force a specific kernel, we must do so in the metapackage. The effect of this PR would be that a Focal instance is _permanently_ on a specific kernel version, meaning nightly updates won't change that value. The only way to change it woudl be to require an admin to run the playbook again. 

It's worth noting that right now, when using Focal VMs, the grsec kernel _is_ preferred. The logic that we have in place to force grsec on the subsequent boot, followed by removing all non-grsec kernels, is sufficient to force grsec over multiple reboots. Before Focal support is production-ready we should indeed peg a version via the metapackage, as @emkll recommends, but I don't see a pressing need to do that right now, especially since we're entering freeze for 1.6.0.

conorsch:
@kushaldas Based on our conversation today, I suggest we close this PR. Since the kernel-related config tests are currently passing, we don't need to update the metapackage logic, and deferring such changes simplifies the ongoing release QA, as well. We'd still keep #5507 open to track a more durable update, but we can do that _after_ the entirety of the Focal tests, both config & app, are passing. What do you think?

install_files/securedrop-grsec/DEBIAN/postinst
```diff
@@ -15,14 +15,21 @@ set -x
 #          <conflicting-package> <version>
 # for details, see http://www.debian.org/doc/debian-policy/ or
 # the debian-policy package
+GRSEC_VERSION='4.14.188-grsec'
+
+# Sets default grub boot parameter to the kernel version specified
+# by $GRSEC_VERSION. The debian buster default kernel is 4.19, thus
```
conorsch, line 21: Comment references "buster" & "4.19" kernel, neither of which are relevant to SecureDrop servers. The comment and the rest of the `postinst` patch are pulled from https://github.com/freedomofpress/securedrop-debian-packaging/blob/05eedc6e2a5d9a1c53ebb6dda0a2188365063e05/securedrop-workstation-grsec/debian/postinst#L26-L32 We should definitely rewrite for clarity if/when we adopt this approach.

install_files/securedrop-grsec/DEBIAN/postinst
```diff
@@ -15,14 +15,21 @@ set -x
...
+# supersedes this 4.14.x series grsecurity kernel at boot-time
+set_grub_default() {
+    GRUB_OPT="'Advanced options for Debian GNU/Linux>Debian GNU/Linux, with Linux $GRSEC_VERSION'"
```
conorsch, line 24: In the grub menu, does Ubuntu still set `Debian GNU/Linux>Debian GNU/Linux`? That is, I'd expect "Ubuntu", not "Debian". See in https://github.com/freedomofpress/securedrop/blob/f81e41539bc5e5ae8b9a443ead719513622553c8/install_files/ansible-base/roles/grsecurity/tasks/from_fpf_repo_install_grsec.yml#L61

where we're explicitly referring to the menu entry as `Advanced options for Ubuntu>`. It's possible the appropriate settings have changed between Xenial & Focal, but at the very least we should be setting the same in both places.

conorsch (commented):
As discussed above, I'm closing this issue. We're currently able to test the custom kernels in Focal VMs without this patch, which is good enough for the next steps on Focal support. I'll add some more info to #5507 tracking related concerns that we'll want to address prior to prod.

PR #5532, opened 2020-09-26
Add type annotations to crypto_utils and config

nabla-c0d3:
## Status

Ready.

## Description of Changes

To help with #4399 , this PR adds and enforces type annotations in the following files:

* crypto_utils.py
* sdconfig.py

This required significant changes to sdconfig.py which wasn't "typing-compatible", but also allowed me to simplify some of the config logic.


securedrop/crypto_util.py
```diff
@@ -215,15 +210,15 @@ def genkeypair(self, name, secret):
...
+        :param name: The source's filesystem id (their codename, salted
                          with SCRYPT_ID_PEPPER, and hashed with scrypt).
-        :param str secret: The source's codename.
+        :param secret: The source's codename.
         :returns: a :class:`GenKey <gnupg._parser.GenKey>` object, on which
                   the ``__str__()`` method may be called to return the
                   generated key's fingeprint.
 
         """
-        name = clean(name)
```
nabla-c0d3, line 226: clean() was always returning the same name.

securedrop/journalist_app/__init__.py
```diff
@@ -43,36 +43,28 @@ def create_app(config: 'SDConfig') -> Flask:
                 template_folder=config.JOURNALIST_TEMPLATES_DIR,
                 static_folder=path.join(config.SECUREDROP_ROOT, 'static'))
 
-    app.config.from_object(config.JournalistInterfaceFlaskConfig)  # type: ignore
-    app.sdconfig = config
```
nabla-c0d3, line 47: This sdconfig attribute was never used.

securedrop/journalist_app/__init__.py
```diff
@@ -43,36 +43,28 @@ def create_app(config: 'SDConfig') -> Flask:
...
     app.config.update(V2_ONION_ENABLED=v2_enabled, V3_ONION_ENABLED=v3_enabled)
 
+    # TODO: Attaching a Storage dynamically like this disables all type checking (and
+    # breaks code analysis tools) for code that uses current_app.storage; it should be refactored
     app.storage = Storage(config.STORE_DIR,
                           config.TEMP_DIR,
                           config.JOURNALIST_KEY)
 
+    # TODO: Attaching a CryptoUtil dynamically like this disables all type checking (and
+    # breaks code analysis tools) for code that uses current_app.storage; it should be refactored
```
nabla-c0d3, line 67: I think this is really hurts the effectiveness of type annotations and code analysis tools in general (especially for critical code like CryptoUtil).

securedrop/tests/conftest.py
```diff
@@ -116,44 +123,50 @@ def config(tmpdir):
...
     app = create_source_app(config)
     app.config['SERVER_NAME'] = 'localhost.localdomain'
     with app.app_context():
         db.create_all()
-        yield app
+        try:
+            yield app
+        finally:
+            db.session.rollback()
+            db.drop_all()
```
nabla-c0d3, line 152: This clears the DB after each test. Otherwise, one failing test would leave data which would then cause subsequent tests to fail (for example if the tests count how many Journalists are in the DB after an action).

securedrop/tests/test_crypto_util.py
```diff
@@ -109,24 +109,6 @@ def test_encrypt_binary_stream(source_app, config, test_source):
         assert fh.read() == plaintext
 
 
-def test_encrypt_fingerprints_not_a_list_or_tuple(source_app, test_source):
```
nabla-c0d3, line 112: Not needed anymore since encrypt() always takes a list in this PR.

securedrop/crypto_util.py
```diff
@@ -315,13 +310,11 @@ def get_pubkey(self, name):
...
         return key
 
-    def encrypt(self, plaintext, fingerprints, output=None):
+    def encrypt(self, plaintext: str, fingerprints: List[str], output: Optional[str] = None) -> str:
         # Verify the output path
         if output:
             current_app.storage.verify(output)
 
-        if not isinstance(fingerprints, (list, tuple)):
-            fingerprints = [fingerprints, ]
```
nabla-c0d3, line 324: I changed the callers to always pass a list

securedrop/journalist_app/__init__.py
rmol: Do you have a solution in mind? A subclass of `Flask`?

securedrop/journalist_app/__init__.py
rmol: :+1:

securedrop/crypto_util.py
rmol: :+1:

securedrop/crypto_util.py
rmol: :+1:

securedrop/crypto_util.py
```diff
@@ -356,18 +349,7 @@ def decrypt(self, secret, ciphertext):
...
-    >>> clean("Helloworld")
-    'Helloworld'
-    """
-    for c in s:
-        if c not in DICEWARE_SAFE_CHARS and c not in also:
-            raise CryptoException("invalid input: {0}".format(s))
-    # scrypt.hash requires input of type str. Since the wordlist is all ASCII
-    # characters, this conversion is not problematic
-    return str(s)
+def _validate_name_for_diceware(name: str) -> None:
```
rmol, line 352: Why the private naming?

securedrop/tests/test_crypto_util.py
rmol: :+1:

securedrop/tests/conftest.py
rmol: I think that's a good change, though it doesn't seem necessary for the purpose of adding type annotations here.

rmol (commented):
There's some great work here. I wish some of it had been done in separate PRs, since it didn't seem essential for adding type annotations, but I understand going ahead and refactoring while you're there.

I had a couple of questions, one a naming nit, and one about how you'd change the dynamic addition of `Storage` and `CryptoUtil` to the Flask app -- if we can decide how to address that, that would be a worthwhile addition to this PR to improve the type checking.

securedrop/crypto_util.py
nabla-c0d3: To convey the fact that it's only used in this file

securedrop/journalist_app/__init__.py
nabla-c0d3: A subclass wouldn't work because the pattern I've seen in the code base is:
```
from flask import current_app
# [...]
codename = current_app.crypto_util.genrandomid()
```

There is no way (I think?) to change the class of `flask.current_app`.

One approach I thought of is to have a global variable with the "current" CryptoUtil or Storage instance (based on the `config`) so it would be like this:

```
crypto_util = CryptoUtil.get_current()
crypto_util.genrandomid()
```

I can make that change but it's a big refactor and should probably be done in a separate PR? 

I have a poc branch where I did this with only the diceware / words lists logic (as a first step), that I took out of CryptoUtil: https://github.com/nabla-c0d3/securedrop/compare/typing-annotations-i18n-store-and-misc...nabla-c0d3:refactor-diceware-generator#

securedrop/tests/conftest.py
nabla-c0d3: One my changes broke a test, but then I had a lot of other tests failing because the DB wasn't cleared between each test, so I had to add this to get to the "truly broken" test.

securedrop/journalist_app/__init__.py
rmol: What about this?

```python
class JournalistApp(Flask):
    crypto_util = None  # type: CryptoUtil
    storage = None  # type: Storage


def create_app(config: 'SDConfig') -> JournalistApp:
    app = JournalistApp(
        __name__,
        static_folder=path.join(config.SECUREDROP_ROOT, 'static')
    )
    [...]
```

A quick run through `tests/test_journalist.py` passes, and `type(current_app._get_current_object())` returns `journalist_app.JournalistApp`. You have to use the old comment annotations before 3.6, but could that solve the annotation/static analysis problems you saw?

At any rate, like you said, that change would merit its own PR. Certainly doesn't have to be done here.

securedrop/crypto_util.py
rmol: OK. I'd count test code, but this is fine.

rmol (approved):
@nabla-c0d3 I'm going to merge this. I'd like to keep working with you on the annotation of the app object instance attributes, but let's do that in its own PR. Thanks for all the cleanup. You've found a lot of cruft while going through this.

securedrop/journalist_app/__init__.py
nabla-c0d3: I don't think it would work for the pattern I described; you can see it there for example: https://github.com/freedomofpress/securedrop/blob/develop/securedrop/journalist_app/main.py#L108

```
from flask import current_app
# [...]
current_app.crypto_util.encrypt(etc.)
```

`flask.current_app` is typed as a `_FlaskLocalProxy` and I don't see a way to change that.

I think fixing this will require changing how the code retrieves the CryptoUtil instance for the current app.

nabla-c0d3:
Thanks! Sure I'm interested in figuring out the app attributes!

securedrop/journalist_app/__init__.py
rmol: Hmm. With a `Flask` subclass in place, if I add these two lines there:

```python
        current_app.logger.error(
            "current_app is a {}".format(type(current_app))
        )
        current_app.logger.error(
            "current_app proxies a {}".format(type(current_app._get_current_object()))
        )
```
The dev server logs this:
```
[2020-09-29 15:16:14,897] ERROR in main: current_app is a <class 'werkzeug.local.LocalProxy'>
[2020-09-29 15:16:14,897] ERROR in main: current_app proxies a <class 'journalist_app.JournalistApp'>
```

I was going by the [Werkzeug docs](https://werkzeug.palletsprojects.com/en/1.0.x/local/#werkzeug.local.LocalProxy). I'm wondering how you're seeing a different proxy class, but it is supposed to be _some_ sort of proxy for an instance of `Flask` or a derivation of it.

nabla-c0d3: The type does get returned at runtime (as the actual object gets returned, after all), but I don't think this type information is available to mypy when its doing its (static-only) analysis. This is how the method is typed:

``` 
class _FlaskLocalProxy(Flask):
    def _get_current_object(self) -> Flask: ...
```

There is no way to change it to a subclass of `Flask`. Additionally:

* The code base only use current_app and does not call `_get_current_object()` (and people will forget to call it I think).
* I'm not familiar with it but _get_current_object() seems private.

Overall I think fixing this requires a new approach. I'm also not sure if the CryptoUtil or Storage instances need to be attached to the app - to me it just adds another layer of indirection/complexity compared to doing just `CryptoUtil.get_current()`. I might be missing something tho.

PR #5534, opened 2020-09-26
Add type annotations to secure_tempfile.py and models.py

nabla-c0d3:
## Status

Ready.

## Description of Changes

To help with #4399 , this PR adds type annotations in the following files:

* secure_tempfile.py
* models.py 
* upload-screenshots.py

### DIFF
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -117,28 +117,12 @@ class Source(db.Model):
         return collection
 
     @property
-    def fingerprint(self):
+    def fingerprint(self) -> 'Optional[str]':
         return current_app.crypto_util.get_fingerprint(self.filesystem_id)
 
-    @fingerprint.setter
-    def fingerprint(self, value):
-        raise NotImplementedError
-
-    @fingerprint.deleter
-    def fingerprint(self):
-        raise NotImplementedError
-
     @property
-    def public_key(self) -> str:
+    def public_key(self) -> 'Optional[str]':
         return current_app.crypto_util.get_pubkey(self.filesystem_id)
-
-    @public_key.setter
-    def public_key(self, value: str) -> None:
-        raise NotImplementedError
-
-    @public_key.deleter
-    def public_key(self) -> None:
-        raise NotImplementedError
 
     def to_json(self) -> 'Dict[str, Union[str, bool, int, str]]':
         docs_msg_count = self.documents_messages_count()
@@ -219,7 +203,7 @@ class Submission(db.Model):
     def is_message(self) -> bool:
         return self.filename.endswith("msg.gpg")
 
-    def to_json(self) -> "Dict[str, Union[str, int, bool]]":
+    def to_json(self) -> 'Dict[str, Any]':
         seen_by = {
             f.journalist.uuid for f in SeenFile.query.filter(SeenFile.file_id == self.id)
             if f.journalist
@@ -302,7 +286,7 @@ class Reply(db.Model):
     def __repr__(self) -> str:
         return '<Reply %r>' % (self.filename)
 
-    def to_json(self) -> "Dict[str, Union[str, int, bool]]":
+    def to_json(self) -> 'Dict[str, Any]':
         journalist_username = "deleted"
         journalist_first_name = ""
         journalist_last_name = ""
@@ -360,18 +344,16 @@ class InvalidUsernameException(Exception):
 class FirstOrLastNameError(Exception):
     """Generic error for names that are invalid."""
 
-    def __init__(self, msg):
-        msg = 'Invalid first or last name.'
+    def __init__(self, msg: str) -> None:
         super(FirstOrLastNameError, self).__init__(msg)
 
 
 class InvalidNameLength(FirstOrLastNameError):
     """Raised when attempting to create a Journalist with an invalid name length."""
 
-    def __init__(self, name):
-        self.name_len = len(name)
-        if self.name_len > Journalist.MAX_NAME_LEN:
-            msg = "Name too long (len={})".format(self.name_len)
+    def __init__(self, name: str) -> None:
+        name_len = len(name)
+        msg = "Name too long (len={})".format(name_len)
         super(InvalidNameLength, self).__init__(msg)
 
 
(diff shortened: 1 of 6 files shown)


securedrop/models.py
```diff
@@ -117,29 +117,13 @@ def collection(self) -> 'List[Union[Submission, Reply]]':
...
+    def fingerprint(self) -> 'Optional[str]':
         return current_app.crypto_util.get_fingerprint(self.filesystem_id)
 
-    @fingerprint.setter
-    def fingerprint(self, value):
-        raise NotImplementedError
-
-    @fingerprint.deleter
-    def fingerprint(self):
-        raise NotImplementedError
```
nabla-c0d3, line 129: Trying to set or delete a `@property` already raises an exception so this seems unnecessary.

securedrop/upload-screenshots.py
```diff
@@ -23,7 +29,7 @@
 # filename into the canonical title we give that screenshot in Weblate.
 #
 # Example conversion: "source-session_timeout.png" -> "source: session timeout"
-CANONICALIZATION_RULES = ((r"\.png$", ""), (r"-", ": "), (r"_", " "))
+CANONICALIZATION_RULES = [(r"\.png$", ""), (r"-", ": "), (r"_", " ")]
```
nabla-c0d3, line 32: A tuple with an arbitrary length can't be typed so I switched to a list.

securedrop/models.py
rmol: OK, but we should still test that `AttributeError` is raised when attempting to set or delete it.

securedrop/models.py
```diff
@@ -219,7 +203,7 @@ def is_file(self) -> bool:
     def is_message(self) -> bool:
         return self.filename.endswith("msg.gpg")
 
-    def to_json(self) -> "Dict[str, Union[str, int, bool]]":
+    def to_json(self) -> 'Dict[str, Any]':
```
rmol, line 206: Why relax this?

securedrop/models.py
```diff
@@ -302,7 +286,7 @@ def __init__(self,
     def __repr__(self) -> str:
         return '<Reply %r>' % (self.filename)
 
-    def to_json(self) -> "Dict[str, Union[str, int, bool]]":
+    def to_json(self) -> 'Dict[str, Any]':
```
rmol, line 289: Why relax this?

securedrop/request_that_secures_file_uploads.py
```diff
@@ -1,14 +1,16 @@
...
+from werkzeug.formparser import FormDataParser
 
 from secure_tempfile import SecureTemporaryFile
 
 
 class RequestThatSecuresFileUploads(wrappers.Request):
 
-    def _secure_file_stream(self, total_content_length, content_type,
-                            filename=None, content_length=None):
+    def _secure_file_stream(self, total_content_length: int, content_type: Any,
```
rmol: If I've followed the Flask and Werkzeug [source](https://github.com/pallets/werkzeug/blob/9444a66dec9c2ee1bacb5c9846669afcca692875/src/werkzeug/wrappers/base_request.py#L258) correctly, shouldn't `content_type` and `filename` be `Optional[str]`, and `content_length` be `Optional[int]`?

securedrop/tests/test_db.py
```diff
@@ -4,24 +4,10 @@
 from mock import MagicMock
 
 from .utils import db_helper
-from models import (Journalist, Submission, Reply, Source, get_one_or_else,
+from models import (Journalist, Submission, Reply, get_one_or_else,
                     LoginThrottledException)
 
 
-def test_source_public_key_setter_unimplemented(journalist_app, test_source):
```
rmol, line 11: If we're removing the explicit setter and deleter for `Source.public_key`, we should still test that it can't be set or deleted.

securedrop/upload-screenshots.py
rmol: I think you could also have used `Tuple[Tuple[str, str], ...]` for this, but this is fine.

rmol (changes requested):
Looks pretty good to me, except I had a couple of questions about `Any` usage, and would like not to lose test coverage. :slightly_smiling_face:

securedrop/models.py
nabla-c0d3: Then the test case will be just for "Is this attribute defined as a Python `@property`?" which seems super specific and not very important? I can add it back if needed.

securedrop/tests/test_db.py
nabla-c0d3: I can add it back if needed, but this seems overly specific as a test case? Would you add this same test for every field declared as a `@property` in the securedrop code base?

securedrop/models.py
nabla-c0d3: It was causing mypy errors, and `Dict[str, Any]` is, for now, the "official" way to define JSON data (https://github.com/python/typing/issues/182).

nabla-c0d3: This is because each value could be a dictionary of other values.

securedrop/models.py
nabla-c0d3: See my answer in the next comment

securedrop/request_that_secures_file_uploads.py
nabla-c0d3: Correct - I couldn't track it down when I looked at this, but I pushed the updated types to this PR just now.

securedrop/models.py
nabla-c0d3: Another way to think about it, which I mentioned in my other comment: would you add this same test for every field declared as a `@property` in the securedrop code base?

rmol: No, but `Source.public_key` was [originally](https://github.com/freedomofpress/securedrop/pull/3619/commits/7adb2544851f36a689a16e3bc63f6358ef05e466) added as a plain property, as you've done, and then it was deemed important to [explicitly prevent and test for writes](https://github.com/freedomofpress/securedrop/pull/3619/commits/1b852b0e3e09550f6e3d7309789735577caafab4). The reasoning was not captured in #3619 but the comments suggest that the change was deliberate. While characterized as a nit, I'm guessing they wanted an extra speed bump in case someone later changed `public_key` again. 

All that said, we can remove the old tests.

rmol (approved):

PR #5547, opened 2020-10-01
Add type annotations to i18n.py, store.py and misc files

nabla-c0d3:
## Status

Ready.

## Description of Changes

To help with #4399 , this PR adds type annotations in the following files:

* store.py
* i18n.py
* qa_loader.py
* Other misc files

### DIFF
--- a/securedrop/create-dev-data.py
+++ b/securedrop/create-dev-data.py
@@ -9,7 +9,6 @@ import math
 from itertools import cycle
 
 from flask import current_app
-from sqlalchemy.exc import IntegrityError
 
 os.environ["SECUREDROP_ENV"] = "dev"  # noqa
 import journalist_app
@@ -24,7 +23,7 @@ messages = cycle(strings)
 replies = cycle(strings)
 
 
-def main(staging=False):
+def main(staging: bool = False) -> None:
     app = journalist_app.create_app(config)
     with app.app_context():
         # Add two test users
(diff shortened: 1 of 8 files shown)


securedrop/i18n.py
```diff
@@ -35,18 +40,13 @@ class LocaleNotFound(Exception):
...
 
-def setup_app(config, app):
+def setup_app(config: SDConfig, app: Flask) -> None:
     global LOCALES
     global babel
 
-    translation_dirs = getattr(config, 'TRANSLATION_DIRS', None)
-
-    if translation_dirs is None:
-        translation_dirs = path.join(path.dirname(path.realpath(__file__)), 'translations')
```
nabla-c0d3, line 45: Moved to sdconfig,py

rmol:
The changes to `create-dev-data.py` on `develop` since you submitted this require a rebase and some further annotation. Sorry about that.

nabla-c0d3:
Should be ready now.

securedrop/create-dev-data.py
```diff
@@ -79,35 +80,31 @@ def main(staging=False):
...
-        user.otp_secret = otp_secret
-        db.session.add(user)
-        db.session.commit()
-        print('Test user successfully added: '
-              'username={}, password={}, otp_secret={}, is_admin={}'
-              ''.format(username, password, otp_secret, is_admin))
-        return user
-    except IntegrityError:
-        print("Test user already added")
-        db.session.rollback()
```
nabla-c0d3, line 99: During the latest rebase I did, I had to remove this `except` statement causing the function to return `None`, because the rest of the code in this file will crash when `add_test_user()` returns `None`. This was also causing type errors.

For example the code on line 184, 190 and 197 will crash if the journalist is `None`.

securedrop/i18n.py
rmol: Nice.

securedrop/tests/test_i18n.py
```diff
@@ -228,20 +229,6 @@ def test_i18n(journalist_app, config):
         verify_i18n(app)
 
 
-def test_verify_default_locale_en_us_if_not_defined_in_config(config):
```
rmol, line 231: Was going to flag the wholesale removal of a test with no replacement, but I don't think this one actually tests what it's supposed to. I'm OK with deleting it.

rmol (changes requested):
If you could address the unused imports flagged by LGTM, this is otherwise ready.

securedrop/tests/test_i18n.py
nabla-c0d3: This behavior is also now auto-validated by mypy since config.TRANSLATION_DIRS has a type of Path (ie. can't be None).

nabla-c0d3:
> If you could address the unused imports flagged by LGTM, this is otherwise ready.

I did it in and now all checks are passing

rmol (approved):

PR #5573, opened 2020-10-10
Handle FileNotFoundError on download of messages

### LINKED ISSUE #4787: unhandled FileNotFoundError when files have been manually deleted from disk
## Description

In 1.0.0, we introduced logic for an admin to check if there are files that have been deleted on disk but not had their corresponding db rows deleted. Right now in the JI, an unhandled exception will occur in the JI if a user tries to download all submissions and there are any missing files on disk. 

## Steps to Reproduce

1. Submit a file as a source
2. Delete that file on the server at the filesystem level
3. Now download the file from the journalist interface (e.g. Select All -> Download)

## Expected Behavior

All files on the server are downloaded. 

## Actual Behavior

500 occurs and `FileNotFoundError` is unhandled

## Comments

I don't think this the highest priority since this is an edge case but it might be nice to add exception handling to flash a message instructing the user to get their admin to run the cleanup tool (it's easy for an admin to miss a notification in an OSSEC alert).

DrGFreeman:
## Status

Ready for review.

## Description of Changes

Fixes #4787 

Handle the `FileNotFoundError` on download of message in the JI where the corresponding file is missing on the filesystem.
- Display a flash message to the journalist.
- Add a log entry reporting the error.
- Redirect to the page from which the download was attempted.

Also added functional test for flash message and redirect and tests for logging of the error.

## Testing

**Setup**

1. Run `make dev`.
1. Log in the Source Interface and submit two messages.
1. Log in the app server Docker container with `docker exec -it securedrop-dev-0 bash` and delete the first message (filename `1-something_something-msg.gpg`) from the source's directory in `/var/lib/securedrop/store`.

**Scenario 1 - Download via "n unread" button on `/`**
1. Log in the Journalist Interface.
1. Click the "2 unread" button of the first source.
    - [ ] Verify that a flash error message is displayed
    - [ ] Verify that the page remains unchanged (journalist homepage)
    - [ ] Verify that the source's number of unread message is unchanged (2)

**Scenario 2 - Download via "Download Unread" button on `/`**
1. Log in the Journalist Interface.
1. Check the first source's checkbox.
1. Click the "Download Unread" button.
    - [ ] Verify that a flash error message is displayed
    - [ ] Verify that the page remains unchanged (journalist homepage)
    - [ ] Verify that the source's number of unread message is unchanged (2)

**Scenario 3 - Download via "Download" button on `/`**
1. Log in the Journalist Interface.
1. Check the first source's checkbox.
1. Click the "Download" button.
    - [ ] Verify that a flash error message is displayed
    - [ ] Verify that the page remains unchanged (journalist homepage)
    - [ ] Verify that the source's number of unread message is unchanged (2)

**Scenario 4 - Download via "Download Selected" button on `/col/<filesystem_id>`**
1. Log in the Journalist Interface.
1. Click the first source's code name.
1. Click the "Select All" button
1. Click the "Download Selected" button.
    - [ ] Verify that a flash error message is displayed
    - [ ] Verify that the page remains unchanged (source collection page)
    - [ ] Verify that the first message remains unread

**Scenario 5 - Direct message download from `/col/<filesystem_id>`**
1. Log in the Journalist Interface.
1. Click the first source's code name.
1. Click first message.
    - [ ] Verify that a flash error message is displayed
    - [ ] Verify that the page remains unchanged (source collection page)
    - [ ] Verify that the first message remains unread

## Deployment

2. New installs.

Test plan above + added unit/functional tests.

Choose one of the following:

### DIFF
--- a/securedrop/journalist_app/col.py
+++ b/securedrop/journalist_app/col.py
@@ -95,13 +95,16 @@ def make_blueprint(config: SDConfig) -> Blueprint:
         if '..' in fn or fn.startswith('/'):
             abort(404)
 
-        file_ = current_app.storage.path(filesystem_id, fn)
-        if not Path(file_).is_file():
+        file = current_app.storage.path(filesystem_id, fn)
+        if not Path(file).is_file():
             flash(
-                gettext("An unexpected error occurred! Please inform your admin."),
+                gettext(
+                    "An error occured: file not found. "
+                    + "An admin can find more information in the system logs."
+                ),
                 "error"
             )
-            current_app.logger.error("File {} could not be found.".format(file_))
+            current_app.logger.error("File {} not found".format(file))
             return redirect(url_for("col.col", filesystem_id=filesystem_id))
 
         # mark as seen by the current user
--- a/securedrop/journalist_app/col.py
+++ b/securedrop/journalist_app/col.py
@@ -1,5 +1,7 @@
 # -*- coding: utf-8 -*-
 
+from pathlib import Path
+
 from flask import (
     Blueprint,
     abort,
@@ -92,6 +94,15 @@ def make_blueprint(config: SDConfig) -> Blueprint:
         """
         if '..' in fn or fn.startswith('/'):
             abort(404)
+
+        file_ = current_app.storage.path(filesystem_id, fn)
+        if not Path(file_).is_file():
+            flash(
+                gettext("An unexpected error occurred! Please inform your admin."),
+                "error"
+            )
+            current_app.logger.error("File {} could not be found.".format(file_))
+            return redirect(url_for("col.col", filesystem_id=filesystem_id))
 
         # mark as seen by the current user
         try:
--- a/securedrop/journalist_app/col.py
+++ b/securedrop/journalist_app/col.py
@@ -99,8 +99,8 @@ def make_blueprint(config: SDConfig) -> Blueprint:
         if not Path(file).is_file():
             flash(
                 gettext(
-                    "An error occured: file not found. "
-                    + "An admin can find more information in the system logs."
+                    "Your download failed because a file could not be found. An admin can find "
+                    + "more information in the system and monitoring logs."
                 ),
                 "error"
             )
--- a/securedrop/journalist_app/utils.py
+++ b/securedrop/journalist_app/utils.py
@@ -207,8 +207,8 @@ def download(zip_basename: str, submissions: List[Union[Submission, Reply]]) -> 
     except FileNotFoundError:
         flash(
             gettext(
-                "An error occured: file not found. "
-                + "An admin can find more information in the system logs."
+                "Your download failed because a file could not be found. An admin can find "
+                + "more information in the system and monitoring logs."
             ),
             "error"
         )
--- a/securedrop/journalist_app/utils.py
+++ b/securedrop/journalist_app/utils.py
@@ -191,7 +191,7 @@ def mark_seen(targets: List[Union[Submission, Reply]], user: Journalist) -> None
             raise
 
 
-def download(zip_basename: str, submissions: List[Union[Submission, Reply]]) -> flask.Response:
+def download(zip_basename: str, submissions: List[Union[Submission, Reply]]) -> werkzeug.Response:
     """Send client contents of ZIP-file *zip_basename*-<timestamp>.zip
     containing *submissions*. The ZIP-file, being a
     :class:`tempfile.NamedTemporaryFile`, is stored on disk only
--- a/securedrop/journalist_app/utils.py
+++ b/securedrop/journalist_app/utils.py
@@ -3,6 +3,7 @@ import binascii
 import datetime
 import os
 from typing import Optional, List, Union, Any
+from urllib.parse import urlparse
 
 import flask
 import werkzeug
@@ -201,7 +202,19 @@ def download(zip_basename: str, submissions: List[Union[Submission, Reply]]) -> 
     :param list submissions: A list of :class:`models.Submission`s to
                              include in the ZIP-file.
     """
-    zf = current_app.storage.get_bulk_archive(submissions, zip_directory=zip_basename)
+    try:
+        zf = current_app.storage.get_bulk_archive(submissions, zip_directory=zip_basename)
+    except FileNotFoundError as e:
+        flash(
+            gettext("An unexpected error occurred! Please inform your admin."),
+            "error"
+        )
+
+        current_app.logger.error("File {} could not be found.".format(e.filename))
+
+        referrer = urlparse(request.referrer).path
+        return redirect(referrer)
+
     attachment_filename = "{}--{}.zip".format(
         zip_basename, datetime.datetime.utcnow().strftime("%Y-%m-%d--%H-%M-%S")
     )
--- a/securedrop/journalist_app/utils.py
+++ b/securedrop/journalist_app/utils.py
@@ -212,7 +212,7 @@ def download(zip_basename: str, submissions: List[Union[Submission, Reply]]) -> 
 
         current_app.logger.error("File {} could not be found.".format(e.filename))
 
-        referrer = urlparse(request.referrer).path
+        referrer = urlparse(str(request.referrer)).path
         return redirect(referrer)
 
     attachment_filename = "{}--{}.zip".format(
--- a/securedrop/journalist_app/utils.py
+++ b/securedrop/journalist_app/utils.py
@@ -204,14 +204,14 @@ def download(zip_basename: str, submissions: List[Union[Submission, Reply]]) -> 
     """
     try:
         zf = current_app.storage.get_bulk_archive(submissions, zip_directory=zip_basename)
-    except FileNotFoundError as e:
+    except FileNotFoundError:
         flash(
-            gettext("An unexpected error occurred! Please inform your admin."),
+            gettext(
+                "An error occured: file not found. "
+                + "An admin can find more information in the system logs."
+            ),
             "error"
         )
-
-        current_app.logger.error("File {} could not be found.".format(e.filename))
-
         referrer = urlparse(str(request.referrer)).path
         return redirect(referrer)
 
--- a/securedrop/store.py
+++ b/securedrop/store.py
@@ -201,6 +201,8 @@ class Storage:
                        for i in selected_submissions])
         # The below nested for-loops are there to create a more usable
         # folder structure per #383
+        missing_files = False
+
         with zipfile.ZipFile(zip_file, 'w') as zip:
             for source in sources:
                 fname = ""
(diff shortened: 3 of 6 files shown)


eloquence:
Thanks for picking this up @DrGFreeman. Don't hesitate to comment if you're looking for feedback even while it's still in draft stage.

DrGFreeman:
Thanks for the follow-up @eloquence. I was having difficulty setting up the missing file for the functional tests but figured out a way today (ref https://github.com/freedomofpress/securedrop/pull/5573/commits/995eed0cc673969e2561047cbae8db4831cd051a).

I'm not yet done with the PR as I also want to check for proper logging of the error. UX feedback would however be welcome when you or someone has time.  I added testing steps to the PR.

DrGFreeman:
As discussed in https://github.com/freedomofpress/securedrop/issues/4787#issuecomment-705774081, I used a catch-all flash message for now. I'm open for modifications of the message as well as the error message being logged. Thx.

eloquence:
Kicked the error message around a bit with @creviera and @ninavizz.

Currently:

> "An unexpected error occurred! Please inform your admin."

(Also used in other contexts.)

Proposed option A replacement:

> "An unexpected error has occurred. Your admin can find more information in the system logs."

Proposed option B replacement:

> "An unexpected error has occurred. More information has been recorded in the system logs."

Rationale:
- Exclamation mark can be anxiety-provoking
- Present perfect tense connects the error more clearly with the user's action
- Pointing to logs (_if it is true_, which judging by the code changes is the case here) gives more actionable information

Thoughts @DrGFreeman, others?

DrGFreeman:
@eloquence, thanks for the feedback.

> Proposed option A replacement:
> 
> > "An unexpected error has occurred. Your admin can find more information in the system logs."
> 
> Proposed option B replacement:
> 
> > "An unexpected error has occurred. More information has been recorded in the system logs."

I would personally go with option A as it refers the user to their admin while hinting at the additional information in the logs. With option B, as a journalist I may not know how to access the system logs and/or think about contacting the admin.

> Rationale:
> 
>     * Exclamation mark can be anxiety-provoking
>
>     * Present perfect tense connects the error more clearly with the user's action

I fully agree.

Side note: Instead of creating a new message, I used one that was already in the codebase:

https://github.com/freedomofpress/securedrop/blob/6475ae8591a9e553041d512a20c4a41788cd339d/securedrop/journalist_app/main.py#L130-L132

https://github.com/freedomofpress/securedrop/blob/44a2e52e67d7ed3e32c5127ed1912c9cd07862f6/securedrop/journalist_app/utils.py#L58-L60

https://github.com/freedomofpress/securedrop/blob/44a2e52e67d7ed3e32c5127ed1912c9cd07862f6/securedrop/journalist_app/utils.py#L157-L159

Maybe we should consider improving these messages as well? I could include this as an additional change in this PR.

eloquence:
Thanks @DrGFreeman. Yes, I noticed that, and think we should standardize the message where it is used, assuming of course that errors are indeed logged in all those cases. I suggest we wait another day or so, in case other folks want to weigh in before we do that.

securedrop/journalist_app/utils.py
```diff
@@ -201,7 +202,19 @@ def download(zip_basename: str, submissions: List[Union[Submission, Reply]]) ->
     :param list submissions: A list of :class:`models.Submission`s to
                              include in the ZIP-file.
     """
-    zf = current_app.storage.get_bulk_archive(submissions, zip_directory=zip_basename)
+    try:
+        zf = current_app.storage.get_bulk_archive(submissions, zip_directory=zip_basename)
```
sssoleileraaa, line 206: we can create a follow-up issue to log each missing file. it would be especially helpful if there are missing files across multiple sources.

securedrop/journalist_app/col.py
```diff
@@ -91,6 +93,15 @@ def download_single_file(filesystem_id, fn):
         if '..' in fn or fn.startswith('/'):
             abort(404)
 
+        file_ = current_app.storage.path(filesystem_id, fn)
+        if not Path(file_).is_file():
```
sssoleileraaa: [nit] it's ok to use file as a variable name, see how it's no longer a builtin: https://docs.python.org/3/library/functions.html#built-in-functions

sssoleileraaa (commented):
lgtm: we no longer display a FileNotFoundError with traceback; instead a more user-friendly error is shown on the same page. one improvement would be to keep the checkboxes checked after an error rather than clearing the selection. this will help the user remember which files they tried to download. i think it's fine to create a follow-up issue if this takes too much time, especially since we might want to catch the missing files early on and handle it differently in the future. sounds like we're going to let this pr sit a bit longer to discuss error messaging, which gives you time to address my comment about logging and leaving the checkboxes checked (but again more than happy to open some follow-up issues for this just like i did for the bug found while testing your pr: https://github.com/freedomofpress/securedrop-client/issues/1173)

sssoleileraaa:
Forgot to share this lovely test plan:

**Scenario 1 - Download via "n unread" button on `/`**
1. Log in the Journalist Interface.
1. Click the "2 unread" button of the first source.
    :heavy_check_mark: Verify that a flash error message is displayed
    :heavy_check_mark: Verify that the page remains unchanged (journalist homepage)
    :heavy_check_mark: Verify that the source's number of unread message is unchanged (2)

**Scenario 2 - Download via "Download Unread" button on `/`**
1. Log in the Journalist Interface.
1. Check the first source's checkbox.
1. Click the "Download Unread" button.
    :heavy_check_mark: Verify that a flash error message is displayed
    :heavy_check_mark: Verify that the page remains unchanged (journalist homepage)
    :heavy_check_mark: Verify that the source's number of unread message is unchanged (2)

**Scenario 3 - Download via "Download" button on `/`**
1. Log in the Journalist Interface.
1. Check the first source's checkbox.
1. Click the "Download" button.
    :heavy_check_mark: Verify that a flash error message is displayed
    :heavy_check_mark: Verify that the page remains unchanged (journalist homepage)
    :heavy_check_mark: Verify that the source's number of unread message is unchanged (2)

**Scenario 4 - Download via "Download Selected" button on `/col/<filesystem_id>`**
1. Log in the Journalist Interface.
1. Click the first source's code name.
1. Click the "Select All" button
1. Click the "Download Selected" button.
    :heavy_check_mark: Verify that a flash error message is displayed
    :heavy_check_mark: Verify that the page remains unchanged (source collection page)
    :heavy_check_mark: Verify that the first message remains unread

**Scenario 5 - Direct message download from `/col/<filesystem_id>`**
1. Log in the Journalist Interface.
1. Click the first source's code name.
1. Click first message.
    :heavy_check_mark: Verify that a flash error message is displayed
    :heavy_check_mark: Verify that the page remains unchanged (source collection page)
    :heavy_check_mark: Verify that the first message remains unread

sssoleileraaa:
re proposed error message replacement options - couple of suggestions:

* i think "please contact an administrator" instead of saying "your administrator" sounds better - people might not think of people with securedrop admin privileges as _their_ admins

* the error is that a file is missing, so we could be better about catching that error specifically and communicating about it specifically

* overall, all language suggestions seem fine to me so far, especially since a) this is such an edge case and b) it would probably be better to focus on creating clear logs of which files are missing and not show database entries without files as options for download (also might be better to alert an admin ahead of time rather than waiting until user action)

securedrop/journalist_app/utils.py
DrGFreeman: I thinks it should not be too difficult to check for the existence of each requested file in the `get_bulk_archive` method and log the missing files there. We could still raise a `FileNotFoundError` exception if any file is missing at the end of the verification and keep the same handling in the `utils.download` function.

I'll experiment with that and add it to the PR if I have an easy solution. Otherwise, we can raise a follow-up issue

DrGFreeman:
@creviera, thanks for the feedback!

I'll look into updating the error messages per your suggestions over the weekend and see how I can proceed to leave the messages selected in the case where the user used "Download Selected" (test scenario 4).

securedrop/journalist_app/utils.py
DrGFreeman: Logging of each missing file implemented in ee18a239bd1b7e00fa6910c233901e619e75ddbd.

securedrop/journalist_app/col.py
```diff
@@ -93,6 +95,18 @@ def download_single_file(filesystem_id: str, fn: str) -> werkzeug.Response:
         if '..' in fn or fn.startswith('/'):
             abort(404)
 
+        file = current_app.storage.path(filesystem_id, fn)
+        if not Path(file).is_file():
+            flash(
+                gettext(
+                    "An error occured: file not found. "
```
sssoleileraaa: typo: occured -> occurred

the following message incorporates most of the feedback from you and others, so if you could update the error message to the following then i think we're golden:

> Your download failed because a file could not be found. An admin can find more information in the system and monitoring logs.

(If you're curious, I mentioned more about the issue of how we want to direct the admin to the right place here: https://github.com/freedomofpress/securedrop-docs/issues/69#issuecomment-721347115)

sssoleileraaa (commented):
this looks great! tests pass and there are clear logs of which files are missing. left one review comment.

securedrop/journalist_app/col.py
DrGFreeman: > typo: occured -> occurred

Dang! Can never get this one right...

securedrop/store.py
```diff
@@ -201,25 +201,36 @@ def get_bulk_archive(self,
...
                     else:
-                        fname = os.path.join(zip_directory, source)
-                    zip.write(filename, arcname=os.path.join(
-                        fname,
-                        "%s_%s" % (document_number,
-                                   submission.source.last_updated.date()),
-                        os.path.basename(filename)
-                    ))
-        return zip_file
+                        missing_files = True
```
nabla-c0d3, line 227: I might be missing something but why not `raise FileNotFoundError` here directly?

securedrop/store.py
```diff
@@ -201,25 +201,36 @@ def get_bulk_archive(self,
...
                 fname = ""
                 submissions = [s for s in selected_submissions
                                if s.source.journalist_designation == source]
                 for submission in submissions:
                     filename = self.path(submission.source.filesystem_id, submission.filename)
-                    document_number = submission.filename.split('-')[0]
-                    if zip_directory == submission.source.journalist_filename:
-                        fname = zip_directory
+
+                    if os.path.exists(filename):
```
nabla-c0d3, line 214: Nitpick: in the previous file you're using pathlib (which is more modern), but not in this file.

securedrop/store.py
DrGFreeman: To allow logging all the missing files if more than one are missing (see commit message https://github.com/freedomofpress/securedrop/pull/5573/commits/ee18a239bd1b7e00fa6910c233901e619e75ddbd)

securedrop/store.py
DrGFreeman: I have a preference for `pathlib` but this function uses `os.path` functions in multiple places so I stayed with that for consistency... Don't know if that's a valid reason though ;)

nabla-c0d3: If it were up to me (it's not)... I would use pathlib for all new code. But your reason for not doing so is valid too haha.

DrGFreeman:
@creviera, message updated as per your proposal in https://github.com/freedomofpress/securedrop/pull/5573/commits/38b0aa4fc5d227574a907b9c0816568d361768db.

Looks like the type linting failure in CI is related to #5616 so I suppose it will be resolved when #5616 is merged.

DrGFreeman:
@creviera & @eloquence, returning on [this discussion](https://github.com/freedomofpress/securedrop/pull/5573#issuecomment-717606885):

> Thanks @DrGFreeman. Yes, I noticed that, and think we should standardize the message where it is used, assuming of course that errors are indeed logged in all those cases. I suggest we wait another day or so, in case other folks want to weigh in before we do that.

Initially we were to use a single generic message for all these errors. Now that the message is specific to missing files, I did not update the other instances. I could update these other messages in this PR but I'd have to review their context and propose appropriate messages.

Or would you prefer to raise a follow-up issue for these other messages?

Please advise.

eloquence:
> Now that the message is specific to missing files, I did not update the other instances.

I think that's the best course of action here, we can revisit that at a later time if warranted. You are a bit behind `develop`, so I think next step is to rebase and ensure we no longer see the type annotation related linting issues.

DrGFreeman:
> I think next step is to rebase and ensure we no longer see the type annotation related linting issues.

@eloquence & @creviera, here you go, rebased and all tests passing!

sssoleileraaa (approved):
ran through the test plan again and lgtm

PR #5583, opened 2020-10-15
Replace docs-lint checkbox in PR template with docs checklist

eloquence:
## Status

Ready for review 

## Description of Changes

Since docs now live in https://github.com/freedomofpress/securedrop-docs and this make target no longer exists in core, it no longer belongs in the PR template.


emkll:
What do you think about replacing this by something like this instead:

"These changes don't need associated documentation / Link to documentation changes"

This would help ensure the docs are kept up-to-date with the changes in this repository

eloquence:
Makes sense, will do :)

eloquence:
(Ready for re-review.)

emkll (approved):
Thanks!

PR #5597, opened 2020-10-22
Add type annotations to /admin and make annotations required for all non-test code

nabla-c0d3:
## Status
Ready.

## Description of Changes

This is the final change for #4399 ; this PR:

* Adds type annotations to /admin/*.py.
* Updates mypy.ini to make type annotations required for all non-test code.

### DIFF
--- a/admin/bootstrap.py
+++ b/admin/bootstrap.py
@@ -23,6 +23,9 @@ import os
 import shutil
 import subprocess
 import sys
+from typing import Iterator
+
+from typing import List
 
 sdlog = logging.getLogger(__name__)
 
@@ -30,7 +33,7 @@ DIR = os.path.dirname(os.path.realpath(__file__))
 VENV_DIR = os.path.join(DIR, ".venv3")
 
 
-def setup_logger(verbose=False):
+def setup_logger(verbose: bool = False) -> None:
     """ Configure logging handler """
     # Set default level on parent
     sdlog.setLevel(logging.DEBUG)
@@ -42,7 +45,7 @@ def setup_logger(verbose=False):
     sdlog.addHandler(stdout)
 
 
-def run_command(command):
+def run_command(command: List[str]) -> Iterator[bytes]:
     """
     Wrapper function to display stdout for running command,
     similar to how shelling out in a Bash script displays rolling output.
@@ -53,6 +56,8 @@ def run_command(command):
     popen = subprocess.Popen(command,
                              stdout=subprocess.PIPE,
                              stderr=subprocess.STDOUT)
+    if popen.stdout is None:
+        raise EnvironmentError("Could not run command: None stdout")
     for stdout_line in iter(popen.stdout.readline, b""):
         yield stdout_line
     popen.stdout.close()
@@ -61,12 +66,12 @@ def run_command(command):
         raise subprocess.CalledProcessError(return_code, command)
 
 
-def is_tails():
+def is_tails() -> bool:
     try:
         id = subprocess.check_output('lsb_release --id --short',
                                      shell=True).decode('utf-8').strip()
     except subprocess.CalledProcessError:
-        id = None
+        return False
 
     # dirty hack to unreliably detect Tails 4.0~beta2
     if id == 'Debian':
@@ -76,7 +81,7 @@ def is_tails():
     return id == 'Tails'
 
 
-def clean_up_tails3_venv(virtualenv_dir=VENV_DIR):
+def clean_up_tails3_venv(virtualenv_dir: str = VENV_DIR) -> None:
     """
     Tails 3.x, based on debian stretch uses libpython3.5, whereas Tails 4.x is
     based on Debian Buster and uses libpython3.7. This means that the Tails 3.x
@@ -90,7 +95,7 @@ def clean_up_tails3_venv(virtualenv_dir=VENV_DIR):
             dist = subprocess.check_output('lsb_release --codename --short',
                                            shell=True).strip()
         except subprocess.CalledProcessError:
-            dist = None
+            return None
 
         # tails4 is based on buster
         if dist == b'buster':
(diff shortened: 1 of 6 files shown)


admin/securedrop_admin/__init__.py
```diff
@@ -255,175 +276,176 @@ def validate(self, document):
...
         # Hold runtime configuration before save, to support
         # referencing other responses during validation
-        self._config_in_progress = {}
+        self._config_in_progress = {}  # type: Dict
         translations = SiteConfig.Locales(
             self.args.app_path).get_translations()
-        translations = " ".join(translations)
+        translations_as_str = " ".join(translations)
+
         self.desc = [
```
nabla-c0d3, line 296: I changed all entries from a list to a tuple so they could be typed.

admin/securedrop_admin/__init__.py
```diff
@@ -255,175 +276,176 @@ def validate(self, document):
...
+            ('v2_onion_services', self.check_for_v2_onion(), bool,
              'WARNING: For security reasons, support for v2 onion services ' +
              'will be removed in February 2021. ' +
              'Do you want to enable v2 onion services?',
              SiteConfig.ValidateYesNo(),
              lambda x: x.lower() == 'yes',
-             lambda config: True],
-            ['v3_onion_services', self.check_for_v3_onion, bool,
+             lambda config: True),
+            ('v3_onion_services', self.check_for_v3_onion(), bool,
```
nabla-c0d3: I think the missing parentheses for self.check_for_v3_onion was a typo?

admin/securedrop_admin/__init__.py
nabla-c0d3: Once Python 3.7 is supported, this would be a good candidate for using a dataclass for each entry instead of a tuple.

mypy.ini
```diff
@@ -1,19 +1,8 @@
...
-disallow_untyped_defs = True
-
-[mypy-crypto_util]
-disallow_untyped_defs = True
-
-[mypy-sdconfig]
-disallow_untyped_defs = True
-
-[mypy-source_app.*]
 disallow_untyped_defs = True
```
nabla-c0d3, line 4: This is it 💪

eloquence:
(New focal CI tests are failing on this PR; I won't be surprised if this has nothing to do with your changes but would suggest another rebase for now.)

nabla-c0d3:
Looks like the tests are passing now...

admin/bootstrap.py
```diff
@@ -53,6 +56,8 @@ def run_command(command):
     popen = subprocess.Popen(command,
                              stdout=subprocess.PIPE,
                              stderr=subprocess.STDOUT)
+    if popen.stdout is None:
+        raise RuntimeError("Should never happen: None stdout")
```
rmol: If you raised `subprocess.CalledProcessError` here, the [one place](https://github.com/freedomofpress/securedrop/blob/1d50b6ecd4bc0cfe6c803630b7fb4f8d39bc3810/admin/bootstrap.py#L146) `run_command` is used would handle it with a more informative error message. If you don't like the idea of commandeering `CalledProcessError`, we could catch `RuntimeError` there too. Either way, maybe we could include the command in the exception message, to help troubleshooting in case we do use `run_command` in another context.

admin/securedrop_admin/__init__.py
```diff
@@ -522,21 +549,18 @@ def validated_input(self, prompt, default, validator, transform):
             default = " ".join(default)
         if type(default) is not str:
             default = str(default)
-        kwargs = {}
-        if validator:
-            kwargs['validator'] = validator
         value = prompt_toolkit.prompt(prompt,
                                       default=default,
-                                      **kwargs)
+                                      validator=validator)
```
rmol, line 554: :+1:

rmol (changes requested):
This looks good. I ran through a full installation on test hardware, saw no errors, and wound up with a functional site. Running `securedrop-admin verify` passed.

I particularly like the improvement to `install_pip_dependencies` in `bootstrap.py` and the corresponding changes in `test_securedrop-admin-setup.py`.

I have two requests. The first I made in a comment, about improving the error messaging from `bootstrap.py::run_command`. The second would be to ensure that an existing `securedrop/config.py` doesn't cause `make typelint` to fail. Right now if that file has been generated with `make test-config`, you get:
```
securedrop/config.py:51: error: "Type[FlaskConfig]" has no attribute "USE_X_SENDFILE"
securedrop/config.py:57: error: "Type[FlaskConfig]" has no attribute "MAX_CONTENT_LENGTH"
Found 2 errors in 1 file (checked 106 source files)
```

Those errors could be eliminated (once the config were recreated, of course) if you added the attributes to the `FlaskConfig` class in `config.py.example`:
```diff
@@ -3,7 +3,9 @@ import os
 # Default values for production, may be overridden later based on environment
 class FlaskConfig(object):
     DEBUG = False
+    MAX_CONTENT_LENGTH = 524288000
     TESTING = False
+    USE_X_SENDFILE = True
     WTF_CSRF_ENABLED = True
 
 # Use different secret keys for the two different applications so their
```

admin/bootstrap.py
nabla-c0d3: This is not really a "normal" exception that you are supposed to catch and act on. A `None` stdout means that something is really wrong (weird OS or Python?) and the server/script can't run properly ie. a crash is the goal here, but at the same it "should never happen". 
I had to add this to fix type errors in the case stdout is None (but it also makes sense), instead of a `type: ignore`.

nabla-c0d3:
I have made the change on config.py.example. For the other request, I replied but can make another change if needed.

admin/bootstrap.py
rmol: My goal here is to not have an admin told `Should never happen: None stdout` while installing or updating SecureDrop, because that will mean absolutely nothing to them, while eroding confidence in the project. "If it should never happen, why am I seeing it?" If it can really never happen, and you're just adding it to satisfy mypy, I think `type: ignore` would be better than introducing the possibility of this. If you prefer the message, I'd like to at least give them and us something that will help troubleshoot the problem. "Error running  _command_" would be OK, or catching it where `run_command` is invoked and presenting something more specific, as done in [`install_apt_dependencies`](https://github.com/freedomofpress/securedrop/blob/1d50b6ecd4bc0cfe6c803630b7fb4f8d39bc3810/admin/bootstrap.py#L152).

nabla-c0d3: I see - I changed it to an EnvironmentError with a better error message.

rmol:
I steered you wrong, my friend. Setting `USE_X_SENDFILE = True` by default on the `FlaskConfig` class in `config.py.example` led to the CI failures, because it's not set to `False` later in the environment-specific changes for `dev` or `test`.

It's safe to just change it to `False`. Sorry about that.

nabla-c0d3:
Ok no problem - I changed it to False.

rmol (approved):

conorsch (approved):
Looks great! Thank you, @nabla-c0d3, for the slew of type annotations PRs recently. We're finally expecting them everyone now. =)

Issue #5599, opened 2020-10-24
CryptoUtil code is not type-checked due to dynamic lookup/usage


nabla-c0d3:
## Description

Initial discussion at https://github.com/freedomofpress/securedrop/pull/5532/files#r495506285

The CryptoUtil class now has type annotations, but they are never actually checked because the CryptoUtil instance is dynamically attached to the Flask app, and later dynamically accessed when it is used throughout the code base:

```
    app.crypto_util = CryptoUtil()
    # [...]
    from flask import current_app
    current_app.crypto_util.do_something()  # Problem: code analysis tools do not know what current_app.crypto_util is
```

This prevents type checking and other code analysis tools from properly analyzing CryptoUtil. However this class has security-sensitive code which should be checked using all the tools available.

## Steps to Reproduce

1. Write buggy/invalid code that uses the existing pattern in the code base for accessing CryptoUtil. For example:
```
    from flask import current_app
    current_app.crypto_util.encrypt("missing arguments here")
```

2. Run mypy on the code base.
3. Mypy does not detect any problem although the code is wrong as it's missing arguments for the encrypt() method.

## Expected Behavior

The following error when running mypy:

```
error: Too few arguments for "encrypt" of "CryptoUtil"
```

## Comments

The best solution I can think of is to move-away from the following pattern:

```
    app.crypto_util = CryptoUtil()
    # [...]
    from flask import current_app
    current_app.crypto_util.do_something()  # Problem: not type-checked
```

And instead of attaching the CryptoUtil instance to the Flask app (which also significantly complicates the test suite), making it an explicit global that's independent of Flask or the Source/Journalist apps (ie. less coupling):

```
    crypto_util = CryptoUtil.get_default()
    crypto_util.do_something()  # Properly type-checked
```

nabla-c0d3:
This is now fixed 🥂

legoktm:
(assuming it was just an oversight this issue wasn't closed earlier)

PR #5600, opened 2020-10-24
Move passphrase logic from CryptoUtil to dedicated class for type-checking

nabla-c0d3:
## Status

Ready.

## Description of Changes

Rather than fixing #5599  in a single PR which would require a lot of changes, this PR extracts the logic related to generating passphrases from the CryptoUtil class in order to:

1. Provide an example of the solution I described in #5599 to move away from `current_app.crypto_util` (which prevents proper type-checking and code analysis). To keep this PR small I only changed the code for generating passphrases within CryptoUtil, instead of trying to fix all of CryptoUtil. Hence this PR is mainly about replacing `current_app.crypto_util.genrandomid()` with `PassphrasesGenerator.get_default().generate_passphrase()`.

2. Refactor and clarify all code related to passphrases. For example the passphrase are sometimes called codename, secret, randomid, etc. in the code, which can be confusing when first looking at the code base. I introduced a DicewarePassphrase type to help clarify.

3. Consolidate the logic of managing words list and passphrases (and the related tests) into a single class; most of the logic was in CryptoUtil, but not all of it as can be seen in this PR.

If the approach I took with this PR (mainly replacing `current_app.crypto_util.genrandomid()` with `PassphrasesGenerator.get_default().generate_passphrase()`) is approved, I can do the same for the rest of the CryptoUtil code.


securedrop/wordlist
```diff
@@ -1 +0,0 @@
-wordlists/en.txt
```
nabla-c0d3, line 1: This file didn't seem to be used anymore.

securedrop/journalist_app/utils.py
```diff
@@ -313,18 +309,6 @@ def col_delete(cols_selected: List[str]) -> werkzeug.Response:
     return redirect(url_for('main.index'))
 
 
-def make_password(config: SDConfig) -> str:
-    while True:
-        password = current_app.crypto_util.genrandomid(
-            7,
-            i18n.get_language(config))
-        try:
-            Journalist.check_password_acceptable(password)
```
rmol, line 335: I think `passphrases.PassphrasesGenerator.generate_passphrase` should preserve this behavior of not returning a passphrase until it passes our (admittedly minimal) quality checks.

securedrop/manage.py
```diff
@@ -177,24 +177,14 @@ def _get_yubikey_usage() -> bool:
             print('Invalid answer. Please type "y" or "n"')
 
 
-def _make_password() -> str:
```
rmol, line 180: Yeah, this logic shouldn't be repeated here. Very nice cleanup.

securedrop/crypto_util.py
```diff
@@ -135,40 +132,6 @@ def do_runtime_tests(self) -> None:
...
-                wordlist_path = self.__word_list
-
-            with io.open(wordlist_path) as f:
-                content = f.read().splitlines()
-                self.__language2words[locale] = content
-
-        return self.__language2words[locale]
-
-    def genrandomid(self,
-                    words_in_random_id: int = DEFAULT_WORDS_IN_RANDOM_ID,
```
rmol, line 170: You can also remove [`CryptoUtil.DEFAULT_WORDS_IN_RANDOM_ID`](https://github.com/freedomofpress/securedrop/blob/5c375222519521113ebb039230b594bd35a73182/securedrop/crypto_util.py#L69).

securedrop/models.py
```diff
@@ -77,10 +77,6 @@ class Source(db.Model):
     # when deletion of the source was requested
     deleted_at = Column(DateTime)
 
-    # Don't create or bother checking excessively long codenames to prevent DoS
```
rmol, line 74: I'm not entirely comfortable with removing the source-specific passphrase constraints here. It's sometimes necessary to refer to them, as in the [source login form](https://github.com/freedomofpress/securedrop/blob/5c375222519521113ebb039230b594bd35a73182/securedrop/source_app/forms.py#L14), and I don't think they should be defined in those places.

I think the `PassphrasesGenerator` is probably the right place now, provided they're configurable and exposed to code that needs them. Will discuss there.

securedrop/passphrases.py
```diff
@@ -0,0 +1,123 @@
...
+_current_generator = None  # type: Optional["PassphrasesGenerator"]
+
+
+class InvalidWordsListError(Exception):
+    pass
+
+
+class PassphrasesGenerator:
+
+    _PASSPHRASE_WORDS_COUNT = 7
```
rmol: I think these parameters should be configurable, and visible to code that needs to know them, like web forms. What do you think about adding them to the `__init__` arguments, and making them public properties?

securedrop/passphrases.py
```diff
@@ -0,0 +1,123 @@
...
+                raise InvalidWordsListError(
+                    "Passphrases under the minimum length ({}) may be generated:"
+                    " shortest word in words list for language {} is '{}' and number of words per"
+                    " passphrase is {}".format(
+                        self._MIN_PASSPHRASE_LENGTH,
+                        language,
+                        shortest_word,
+                        self._PASSPHRASE_WORDS_COUNT,
+                    )
+                )
```
rmol, line 97: I love that you're checking the word lists -- they're something I've been wanting to clean up for a while -- but I think that `generate_passphrase` is the right place to check quality. If it doesn't return until it has a good passphrase, there's no reason to throw out locale-specific lists just because they might produce one that's too short or too long, or contain words that are too short. We should be relying on (and improving :slightly_frowning_face:) our passphrase checks to ensure generated passphrases are strong enough.

securedrop/passphrases.py
```diff
@@ -0,0 +1,123 @@
...
+
+    @classmethod
+    def get_default(cls) -> "PassphrasesGenerator":
```
rmol: What do you think about moving the word list parsing into `__init__`? Right now creating other instances of `PassphrasesGenerator` would involve repeating this logic before constructing them. I'm thinking about the possibility that we'd want to use separate configurations for sources and journalists. In fact I wonder if establishing those now, during this change, would be a good idea. We could add `passphrase_generator` module attributes to `source_app` and `journalist_app`, or their `utils` modules.

securedrop/passphrases.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+_current_generator = None  # type: Optional["PassphrasesGenerator"]
+
+
+class InvalidWordsListError(Exception):
+    pass
+
+
+class PassphrasesGenerator:
```
rmol: This is a nit, but `InvalidWordsListError` and `PassphrasesGenerator` sound odd to me. I'd expect `InvalidWordListError` and `PassphraseGenerator`.

securedrop/passphrases.py
```diff
@@ -0,0 +1,123 @@
...
+                )
+
+    @classmethod
+    def get_default(cls) -> "PassphrasesGenerator":
+        global _current_generator
+        if _current_generator is None:
+            language_to_words = _parse_available_words_list(Path(config.SECUREDROP_ROOT))
+            # TODO: The french words list is problematic because it has really short words
+            # TODO: Remove this after fixing the words list
+            del language_to_words["fr"]
```
rmol: If `generate_passphrase` is pickier about what it produces, we don't have to lose French. :slightly_smiling_face:

securedrop/source_app/forms.py
```diff
@@ -5,17 +5,21 @@
...
 from wtforms.validators import InputRequired, Regexp, Length, ValidationError
 
-from models import Source, Submission
+from models import Submission
 
 
 class LoginForm(FlaskForm):
+
+    # Don't bother checking excessively long codenames to prevent DoS
+    _MAX_CODENAME_LEN = 128
```
rmol: I think this should be defined where the passphrases are generated. The `Source` model is probably not the right place any more, but the `PassphrasesGenerator` could make its parameters public.

securedrop/tests/test_passphrases.py
```diff
@@ -0,0 +1,77 @@
...
+        assert generator.available_languages == {"en"}
+
+        # When using it to generate a passphrase
+        # It succeeds
+        passphrase = generator.generate_passphrase()
+
+        # And a reasonably-secure passphrase was generated
+        assert passphrase
+        assert len(passphrase) > 16
+        assert len(passphrase.split(" ")) > 5
```
rmol: These numbers don't match the default parameters in `passphrases.py`.

securedrop/tests/test_source.py
```diff
@@ -21,7 +22,7 @@
 from .utils.db_helper import new_codename
 from .utils.instrument import InstrumentedApp
 
-overly_long_codename = 'a' * (Source.MAX_CODENAME_LEN + 1)
+overly_long_codename = 'a' * (LoginForm._MAX_CODENAME_LEN + 1)
```
rmol: Just have to point out again that I don't think `LoginForm` should be the source of truth for maximum codename length. I can be so tiresome. I apologize. :slightly_frowning_face:

securedrop/tests/test_source.py
```diff
@@ -43,32 +44,6 @@ def test_index(source_app):
         assert 'Return visit' in text
 
 
-def test_all_words_in_wordlist_validate(source_app):
```
rmol, line 46: I'm not sure we want to lose this test. Am I missing a replacement that checks the form validation the same way?

rmol (changes requested):
This is great work. I have a _lot_ of comments and questions, but that's because of what's being changed, more than how. This is already cleaner, and I'm excited about getting this merged and moving on to `CryptoUtil` itself. 

Again, just because of what it's touching, we're going to have at least one other reviewer take a look, so hold off on making any changes until they weigh in. Please do respond with any thoughts you have on my review though.

Thanks again for your contributions.

securedrop/journalist_app/utils.py
nabla-c0d3: There are 3 checks in `check_password_acceptable()`:

1. Minimum length
2. Maximum length
3. Minimum number of words

The logic implemented in `PassphrasesGenerator.generate_passphrase()` ensures that all generated passphrases are "good" (based on these 3 criteria); it can't return a "bad"/low-quality passphrase. 
The tests on `generate_passphrase()` also double check that.

securedrop/passphrases.py
nabla-c0d3: Yeah agreed, I can change that

securedrop/passphrases.py
nabla-c0d3: I like the simplicity of them not being configurable but I can also make the change. What's the use case for making them configurable? Who will configure them, and in what situation?

I think it comes down to whether they are 1) application "constants" that only developers need to update as needed (based on evolving crypto/security best practices) or 2) application "settings" that users will need to tweak in the securedrop config.py for their securedrop instance. 

I think in this case it's 1) and 2) sounds kind of dangerous security-wise, but let me know.

securedrop/passphrases.py
nabla-c0d3: For this purpose, I like the approach of having "clean"/reliable word list, instead of having to do more work at runtime ie. try several passphrase until we find a good one. This adds complexity, requires dedicated test cases (a 100% bad words list, a semi-bad words list, etc.), will lead to bugs and surprises such as passphrase generation taking longer than expected, etc.. 

Creating clean word lists seems like a quick, one-time effort. Maintaining code that can handle bad word lists will take more effort/work (over time), I think,

nabla-c0d3: Another option would be to just filter out bad words from the supplied lists when initializing the Passphrase generator - this is still somewhat simple and doesn't require "perfect" word lists.

securedrop/passphrases.py
nabla-c0d3: Just to make sure I understand, do you have an example of how the configuration would differ?

securedrop/tests/test_passphrases.py
nabla-c0d3: Yeah I will update them

securedrop/source_app/forms.py
nabla-c0d3: Ok I can make the change

securedrop/journalist_app/utils.py
rmol: Do you mean that because word lists that could produce shorter or longer passphrases are dropped in `__init__`, `generate_passphrase` doesn't need to check?

securedrop/passphrases.py
rmol: I was thinking more along the lines of 1) -- the existing code does have vestigial hints of the number of words differing for sources and journalists, though they don't right now. We're losing support for that. But you know what, that's OK. If we need to make them different, we can do that, and refactor where the word lists are loaded, when the need arises again.

If you could just make these attributes public, so they're available in code that needs to know the passphrase requirements, that will do.

securedrop/passphrases.py
rmol: All right. We can revisit this when/if we add better checks and more locale-specific lists than we have now.

securedrop/passphrases.py
rmol: Never mind. This is another thing that can wait until we have a demonstrated need.

rmol:
This will unfortunately need to be rebased before it can be thoroughly tested.

securedrop/journalist_app/utils.py
nabla-c0d3: Exactly - there is no code path where a "bad" passphrase gets generated.

securedrop/passphrases.py
nabla-c0d3: Sounds good, I'll make the attributes public

securedrop/tests/test_source.py
nabla-c0d3: To be honest, I wasn't able to figure out which corner case or code path this was really testing. What's a word that would be rejected by the login form? Is there code somewhere that will reject specific words? The only thing I could think of was non-ascii words but the passphrase generator already enforces ascii.

Since the test case runs over every word it's very slow too, so that's why I removed it. If we have a clear idea of what a "non-allowed" word would be, I can add a test and it could be a lot simpler/faster (by just directly validating the words in the word lists, instead of submitting the login form with every word in the list).

securedrop/passphrases.py
nabla-c0d3: So I ended up implementing what I described in my last comment as I think it's a good improvement over my previous approach:

* The code will now take any word list, but will ignore words that are less than 4 characters. 
* Additionally, it will raise an `InvalidWordListError` if a word list has less than 6000 "long-enough" words, to avoid the case where a word list has like 5 long-enough words so the passphrase are totally not random/secure.
* As before, it will also raise an error if passphrases that are too short may be generated.

This allowed me to re-enable the french word list.

securedrop/passphrases.py
```diff
@@ -0,0 +1,138 @@
...
+    """Find all .txt files in the wordlists folder and parse them as words lists.
+
+    This will also ignore words that are less than 4 letters.
+    """
+    language_to_words = {}
+    words_lists_folder = securedrop_root / "wordlists"
+    for words_file in words_lists_folder.glob("*.txt"):
+        language = words_file.stem
+        all_words = words_file.read_text().strip().splitlines()
+        words_that_are_long_enough = [word for word in all_words if len(word) >= 4]
```
nabla-c0d3, line 136: Is 4 letters enough?

securedrop/passphrases.py
```diff
@@ -0,0 +1,138 @@
...
+
+class PassphraseGenerator:
+
+    PASSPHRASE_WORDS_COUNT = 7
+
+    # Enforce a reasonable maximum length for passphrases to avoid DoS
+    MAX_PASSPHRASE_LENGTH = 128
+    MIN_PASSPHRASE_LENGTH = 20
+
+    _WORD_LIST_MINIMUM_SIZE = 6000  # Minimum number of words in any of the word lists
```
nabla-c0d3: Is 6000 enough? The english word list has about 7000 4+ letter words and the french one has 6300.

nabla-c0d3:
This should be ready for another review; I think I took care of all the feedback  👍

securedrop/passphrases.py
rmol: Good question. 

According to [the Diceware FAQ](https://theworld.com/%7Ereinhold/dicewarefaq.html), Diceware's entropy is calculated as ~12.92 bits per word, based on a list of 7776 (6<sup>5</sup>) words. The formula is <code>M * log2(N)</code>, where `M` is the number of words and `N` the length of the word list, so a seven-word passphrase from 7776 words provides 90.4 bits of entropy. Our current English list should produce 90.25, the French 89.95.

The length of the words is less important then their number, as each is effectively a symbol in a large alphabet. In fact the originator of Diceware intentionally included short, meaningless words to reduce the length of passphrases and make them easier to remember and type. EFF produced Diceware lists with words shorter than three eliminated, but in their own words, that was for usability: ["Note that the security of a passphrase generated using either list is identical; the differences are in usability, including memorability, not in security."](https://www.eff.org/deeplinks/2016/07/new-wordlists-random-passphrases)

So I don't think we should just eliminate short words. It's counter-intuitive, but that would actually reduce the entropy of our passphrases, to 89.41 bits for English at 7000 words, or 88.34 for French at 6300.

My take is that we should definitely have a lower limit on the length of any word list we use for codenames, but it should probably be 7500 or more, to ensure 90 bits of entropy. Perhaps the best way forward is to replace our current English list with the EFF list, not because the short words are a security problem, but because I think the EFF list will produce better machine translations to our other supported languages.

For the purposes of this PR, I think we should set `_WORD_LIST_MINIMUM_SIZE` to 7300 and eliminate the word length requirement when loading word lists. That will keep French codenames working, while ensuring 89.83 bits of entropy.

securedrop/passphrases.py
rmol: See the comment above about Diceware entropy.

securedrop/tests/test_source.py
rmol: I think it would have caught a word that somehow didn't pass the [regular expression in `securedrop/source_app/forms/LoginForm`](https://github.com/freedomofpress/securedrop/blob/3d204257c68cfeb35da8d4f04ab906f6d6413e3a/securedrop/source_app/forms.py#L20), but your new checks of the word lists would too, and I don't think it's worth 20 seconds every test run. Let's drop it.

rmol (changes requested):

nabla-c0d3:
Ok thanks for looking into the Diceware questions. I changed the minimum word list size to 7300 and the minimum size of each word to 2.

securedrop/passphrases.py
```diff
@@ -0,0 +1,138 @@
...
+        words = [
+            self._random_generator.choice(words_list) for _ in range(self.PASSPHRASE_WORDS_COUNT)
+        ]  # type: List[str]
+        return DicewarePassphrase(" ".join(words))
+
+
+def _parse_available_words_list(securedrop_root: Path) -> Dict[str, List[str]]:
+    """Find all .txt files in the wordlists folder and parse them as words lists.
+
+    This will also ignore words that are less than 4 letters.
```
emkll: based on the code on L136, shouldn't this ignore words that are less than 2 characters?

emkll (commented):
Thanks @nabla-c0d3 for these changes and @rmol for the thorough review. Changes here look good to me, one minor nit re: a no-longer-accurate comment (inline), due to changes introduced during the PR review. Happy to append a commit to this branch if you'd like.

Thanks again your many contributions to SecureDrop, @nabla-c0d3 !

securedrop/passphrases.py
nabla-c0d3: > Thanks @nabla-c0d3 for these changes and @rmol for the thorough review. Changes here look good to me, one minor nit re: a no-longer-accurate comment (inline), due to changes introduced during the PR review. Happy to append a commit to this branch if you'd like.
> 
> Thanks again your many contributions to SecureDrop, @nabla-c0d3 !

No problem! I fixed the comment just now.

nabla-c0d3:
> Thanks @nabla-c0d3 for these changes and @rmol for the thorough review. Changes here look good to me, one minor nit re: a no-longer-accurate comment (inline), due to changes introduced during the PR review. Happy to append a commit to this branch if you'd like.
> 
> Thanks again your many contributions to SecureDrop, @nabla-c0d3 !

No problem! I fixed the comment just now.

rmol (approved):

PR #5614, opened 2020-11-03
Fix test_rm.py::test_secure_delete_capability

rmol:
## Status

Ready for review

## Description of Changes

When testing what happens when the `shred` command is not found, the `PATH` environment variable shouldn't contain any location where it could ever exist, as just happened in the upgrade from Xenial to Focal. The test script also had a typo.

## Testing

CI should be green.

## Deployment

Test only.

## Checklist

Choose one of the following:


kushaldas (approved):
Code change looks good. CI is green.

PR #5616, opened 2020-11-04
Fix type hints in models, qa_loader

rmol:
## Status

Ready for review

## Description of Changes

Corrects some type hinting problems in `models.py` and `qa_loader.py`.

## Testing

CI is green. Run `make lint` if you feel like it. 

## Deployment

Dev only, 'cause type hints don't actually do anything.

Choose one of the following:

### DIFF
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -407,17 +407,20 @@ class Journalist(db.Model):
     last_name = Column(String(255), nullable=True)
     pw_salt = Column(LargeBinary(32), nullable=True)  # type: Column[Optional[bytes]]
     pw_hash = Column(LargeBinary(256), nullable=True)  # type: Column[Optional[bytes]]
-    is_admin = Column(Boolean)
+    is_admin = Column(Boolean)  # type: Column[Optional[bool]]
     session_nonce = Column(Integer, nullable=False, default=0)
 
     otp_secret = Column(String(16), default=pyotp.random_base32)
-    is_totp = Column(Boolean, default=True)
-    hotp_counter = Column(Integer, default=0)
+    is_totp = Column(Boolean, default=True)  # type: Column[Optional[bool]]
+    hotp_counter = Column(Integer, default=0)  # type: Column[Optional[int]]
     last_token = Column(String(6))
 
-    created_on = Column(DateTime, default=datetime.datetime.utcnow)
-    last_access = Column(DateTime)
-    passphrase_hash = Column(String(256))
+    created_on = Column(
+        DateTime,
+        default=datetime.datetime.utcnow
+    )  # type: Column[Optional[datetime.datetime]]
+    last_access = Column(DateTime)  # type: Column[Optional[datetime.datetime]]
+    passphrase_hash = Column(String(256))  # type: Column[Optional[str]]
     login_attempts = relationship(
         "JournalistLoginAttempt",
         backref="journalist"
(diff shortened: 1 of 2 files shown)


emkll:
(rebased on latest develop)

securedrop/models.py
```diff
@@ -634,6 +637,8 @@ def verify_token(self, token: 'Optional[str]') -> bool:
             # window is 1:30s.
             return self.totp.verify(token, valid_window=1)
         else:
+            if not self.hotp_counter:
```
emkll: The database default value for `hotp_counter` is 0, why do we manually set to 0 here? Under which circumstances is `hotp_counter` is `None`? Shouldn't we be returning `False`?

emkll (commented):
Thanks @rmol lint target now passing after rebase and addresses lint failures in `develop`. One comment/clarification inline before merging

securedrop/models.py
rmol: Excellent question, which made me dig deeper into this. Here's my understanding of this whole house of cards:

Right now, `Journalist.hotp_counter` is defined as `Column(Integer, default=0)`
  - without `nullable=False`, the database schema says `hotp_counter` can be `null`. (In fact, being SQLite, it can be 'what a wonderfully flexible database', too, but that's beside the point.)
  - the `default=0` is _not_ the database default value; it's a value that SQLAlchemy will use [in the case where no value was provided to the INSERT or UPDATE statement for that column](https://docs.sqlalchemy.org/en/14/core/defaults.html).
  - Mypy infers `hotp_counter` as `builtins.int*` -- I thought this might be because the sqlalchemy stubs that were just added were clever enough to notice the `default` clause, but I get the same result on `develop` before the stubs were added, and I've found some nullable columns without defaults (`first_name`, for instance) that are also not inferred as optional. I'm still looking, but in any case it does not match reality, because the column can be `null`. 

So: adding `# type: Column[Optional[int]]` accurately describes what we can get from our database (as long as we disregard SQLite flexibility), but then if `hotp_counter` is not explicitly assigned, mypy knows it might be `None`, so catches that that should not be passed to `range`:
```
securedrop/models.py:641: error: Argument 1 to "range" has incompatible type "Optional[int]"; expected "int"
securedrop/models.py:642: error: Unsupported operand types for + ("None" and "int")
securedrop/models.py:642: note: Left operand is of type "Optional[int]"
```

So that's why it could be `None`, at least in mypy's understanding of the world, and why I made sure it was set to the default, but you're right, I was too far down this rabbit hole and not thinking about the actual method. We should treat a null `hotp_counter` as an error and return `False` from `verify_token`.

And as much as I'd like to grumble about the bolt-on type checker necessitating appeasement code, this is turning up valid problems with our ORM and database and the illusions they create. I think many of our columns should have been created with `nullable=False`. That would take a difficult data migration to fix. In the meantime, I'd like to propagate the `Optional` annotations to them. It doesn't have to be done in this PR, as I really just wanted to fix CI, but I think it's wrong to pretend we can't get `null`/`None` from the database. If we're going to check typing, let's be as accurate as possible.

Thoughts?

emkll: Thanks for digging into this. I agree with your assessment:
1. Immediately, to unblock this PR and CI: return false if `hotp_counter` is `null`
2. Medium term propagate the `Optional` annotation
3. Long term:  Open an issue to add/migrate nullable=False database constrains (which as you say may be a complex migration for long-running instances.

emkll (approved):
lgtm

PR #5621, opened 2020-11-10
Ensure functional test assertions are evaluated

### LINKED ISSUE #5607: Assertions are not evaluated in some functional tests
## Description

Functional tests using navigation steps in which the `assert` statement is behind the `if not hasattr(self, "accept_languages"):` guard clause always pass regardless of the assertion result.

## Steps to Reproduce

Replace the `assert` statement on line 168 below with `assert False`.
https://github.com/freedomofpress/securedrop/blob/8c86e8ee717c4971a34e8555b41f6e9d82e4870e/securedrop/tests/functional/journalist_navigation_steps.py#L162-L170

Run the test in the dev container with `make test` or `securedrop/bin/dev-shell bin/run-test -v -x tests/functional/test_journalist.py::TestJournalist`

## Expected Behavior

`tests/functional/test_journalist.py::TestJournalist::test_journalist_verifies_deletion_of_one_submission_modal` is expected to fail.

## Actual Behavior

`tests/functional/test_journalist.py::TestJournalist::test_journalist_verifies_deletion_of_one_submission_modal` passes.

## Comments

The `FunctionalTest` class from which all functional test classes inherit has `accept_languages = None` as default and this attribute is not modified except for the tests in `tests/pageslayout`. The `if not hasattr(self, "accept_languages"):` guard clause therefore always evaluates to `False` and the assertion is not executed.

I believe the guard clause could be changed to `if self.accept_languages is None:` so the assertions are evaluated during the functional tests and skipped in the page layout tests.

DrGFreeman:
## Status

Ready for review.

## Description of Changes

Fixes #5607

* Ensure functional tests assertions are evaluated by changing the guard clause that prevented them from being evaluated.
* Fix tests that fail after change of guard clause.

## Testing

No specific testing beyond code review and running unit tests.

## Deployment

2. New installs.

Choose one of the following:

Choose one of the following:

### DIFF
--- a/securedrop/tests/functional/journalist_navigation_steps.py
+++ b/securedrop/tests/functional/journalist_navigation_steps.py
@@ -822,13 +822,12 @@ class JournalistNavigationStepsMixin:
     def _visit_edit_hotp_secret(self):
         self._visit_edit_secret(
             "hotp",
-            "Reset two-factor authentication for security keys like Yubikey")
+            "Reset two-factor authentication for security keys, like a YubiKey")
 
     def _visit_edit_totp_secret(self):
         self._visit_edit_secret(
             "totp",
-            "Reset two-factor authentication for mobile apps such as FreeOTP or "
-            "Google Authenticator"
+            "Reset two-factor authentication for mobile apps, such as FreeOTP"
         )
 
     def _admin_visits_add_user(self):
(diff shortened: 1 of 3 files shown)


kushaldas (approved):
- [x] The code change looks good, also contains the text changes.
- [x] CI is green

Approved.

PR #5612, opened 2020-11-03
Update cryptography dependency to 3.2.1

rmol:
## Status

Ready for review

## Description of Changes

Fixes #5611.

Updates `cryptography`. As a bonus, eliminates its transitive dependency on `asn1crypto`.

## Testing

- Run the dev server with `make dev`
- In the source interface, ensure you can submit a large file.
- Ensure that all tests pass, particularly `admin/tests/test_securedrop-admin.py::test_generate_new_v3_keys` and `securedrop/tests/test_secure_tempfile.py`.

## Deployment

The `cryptography` package is used when generating v3 onion service keys, and receiving large submissions. 

### If you made changes to the server application code:

- [ ] Linting (`make lint`) and tests (`make test`) pass in the development container

### If you made changes to `securedrop-admin`:

- [ ] Linting and tests (`make -C admin test`) pass in the admin development container

### If you made non-trivial code changes:

- [x] I have written a test plan and validated it for this PR

Choose one of the following:

- [ ] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-docs) for these changes, or will do so later
- [ ] I would appreciate help with the documentation
- [x] These changes do not require documentation

### If you added or updated a code dependency:

Choose one of the following:

- [x] I have performed a diff review and pasted the contents to [the packaging wiki](https://github.com/freedomofpress/securedrop-debian-packaging/wiki/diff-review-of-cryptography-update-from-2.6.1-to-3.2.1)
- [ ] I would like someone else to do the diff review

PR #5623, opened 2020-11-10
Ask cryptography to please keep using Xenial's OpenSSL

### LINKED ISSUE #5622: cryptography 3.2.1 no longer works with OpenSSL 1.0.2 on Xenial
## Description

The `cryptography` project [stopped supporting OpenSSL 1.0.2 in version 3.2](https://cryptography.io/en/latest/changelog.html#v3-2). The `RuntimeError` [raised now](https://github.com/pyca/cryptography/blob/0c7592c34fd58f0634f493d6ce542ab35d940b26/src/cryptography/hazmat/bindings/openssl/binding.py#L169), after our recent [update to 3.2.1](https://github.com/freedomofpress/securedrop/pull/5612), will prevent file submissions in any environment using the `securedrop-app-code` Debian package, i.e. staging or production.

This wasn't noticed because `cryptography` is installed via binary wheels in the dev container, and [the manylinux wheels are statically linked with newer, supported OpenSSL libraries](https://cryptography.io/en/latest/installation.html#static-wheels) so manual testing and CI never encountered the error.

For now, we can still use the version of OpenSSL maintained on Xenial by setting the `CRYPTOGRAPHY_ALLOW_OPENSSL_102` environment variable, but support for 1.0.2 has now [really been removed](https://github.com/pyca/cryptography/pull/5511), so this will only work until the next time we need to upgrade `cryptography`. With luck, that will be after we migrate to Focal, but if it has to happen sooner, we may have to consider adding `--only-binary=cryptography` to the `dh-virtualenv` [options](https://github.com/freedomofpress/securedrop/blob/ce2282712e0760dadc63caead8c468e22b3a3d58/install_files/securedrop-app-code/debian/rules#L33) to package a wheel linked against a usable OpenSSL version.

Separately, we should probably think about using the same pip options in development containers as we do for the production packages, to prevent this sort of difference going undetected through manual testing, local test runs, and CI. :roll_eyes: 

## Steps to Reproduce

Set up a staging environment on `develop` and try to submit a file through the source interface. You will get a server error, with the `RuntimeError` about the old OpenSSL library in `/var/log/apache2/source-errors.log`.

rmol:
## Status

Ready for review

## Description of Changes

Fixes #5622.

Sets the `CRYPTOGRAPHY_ALLOW_OPENSSL_102` environment variable in `/var/www/journalist.wsgi` and `/var/www/source.wsgi` (and the logging version of that used in staging), so that `cryptography` will deign to use Xenial's manky old OpenSSL library. 

Adds a note to the `securedrop-app-code` requirements file about what we'll have to do if we need another `cryptography` update before we're on Focal.

## Testing

- Build a staging environment from `develop`: `make build-debs; make staging`
- Check that submitting a file through the source interface results in a server error.
- Rebuild staging from this branch:
  - `git checkout -b fix-5622 origin/fix-5622`
  - `make build-debs; make staging`
- Submitting a file should now work.

## Deployment

This will prevent the `cryptography` update from #5612 from breaking file submission in all production instances with our next release. 

Choose one of the following:

### DIFF
--- a/install_files/ansible-base/roles/app-test/files/source.wsgi.logging
+++ b/install_files/ansible-base/roles/app-test/files/source.wsgi.logging
@@ -1,6 +1,9 @@
 #!/usr/bin/python
 
+import os
 import sys
+
+os.environ["CRYPTOGRAPHY_ALLOW_OPENSSL_102"] = "True"
 sys.path.insert(0,"/var/www/securedrop")
 
 import logging
--- a/install_files/securedrop-app-code/var/www/journalist.wsgi
+++ b/install_files/securedrop-app-code/var/www/journalist.wsgi
@@ -1,7 +1,9 @@
 #!/opt/venvs/securedrop-app-code/bin/python
 
+import os
 import sys
 
+os.environ["CRYPTOGRAPHY_ALLOW_OPENSSL_102"] = "True"
 sys.path.insert(0, "/var/www/securedrop")
 
 import logging
--- a/install_files/securedrop-app-code/var/www/source.wsgi
+++ b/install_files/securedrop-app-code/var/www/source.wsgi
@@ -1,7 +1,9 @@
 #!/opt/venvs/securedrop-app-code/bin/python
 
+import os
 import sys
 
+os.environ["CRYPTOGRAPHY_ALLOW_OPENSSL_102"] = "True"
 sys.path.insert(0, "/var/www/securedrop")
 
 # import logging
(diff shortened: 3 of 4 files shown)


kushaldas (approved):
The code change looks good.

# Tested with the following steps

- [x] Build a staging environment from `develop`: `make build-debs; make staging`
- [x] Check that submitting a file through the source interface results in a server error.
- [x] Rebuild staging from this branch:
  - [x] `git checkout -b fix-5622 origin/fix-5622`
  - [x] `make build-debs; make staging`
- [x] Submitting a file should now work.

Approved.

Issue #4280, opened 2019-03-20
Include Newsroom Name & Update Alt Tag


ninavizz:
## Description

Currently there is no mention of the name of a newsroom on an instance of SecureDrop. This presents two core problems for Sources, in the Source Experience.

1. The `<alt>` on the newsroom logo today, is currently `SecureDrop`. That's strange and misleading. See 0:20 and 0:30 in [this video screencast](https://drive.google.com/open?id=1WgTdWDEd-4dG16pobSlztCvfH6JgBGMm) of AlJazeera's SecureDrop loading w/in Tor. A minimum/asap fix, would be updating the static `<alt>` w/in the code to read `Newsroom Logo` (sidenote: alt tags are a big accessibility thing, on an aside from also informing users what's what as assets load).

Because `Cancel` buttons now exist everywhere (and take users to the SD .index page), removing the interactivity on the newsroom logo also seems prudent, as it's redundant and potentially misleading.

2. Longer term, the Admin interface should be updated to accept the newsroom's name (including the word _The_ for orgs such as _The Intercept_ or _The New York Times_) and that alt-tag should be dynamically generated to read `<%NewsOrgName%> Logo`

## User Research Evidence

- None (yet). 
- General Cognitive Walkthrough Observations from UXKid™
  - Accessibility audit likely to make `<alt>` recommendations
  - As a stranger to the Source experience, it looks weird seeing "SecureDrop" as an alt-tag show-up on my screen, to then be replaced by a newsroom's logo. 
  - This is a great low-hanging fruit opportunity to start managing user expectations better wrt who they're submitting things to.
  - Get the change into the Admin interface to accept news org names and the dynamic alt-tag in the Source interface, now; then in a tbd release start inserting the news or name in the Source experience after users have validated where it's most useful for them to see.

## User Stories

As a Source I want to be less confused about who I'm submitting my stuff to, and what this SecureDrop thing is.

ninavizz:
Hmmm, seems we just got report submitted from Accessibility lab thaaaat seems to validate this? Could an update to this `alt` tag pls to get included in the 0.13 release? 😃  @eloquence

ninavizz:
Bumpitty bump! Could this get a "Help Wanted" tag, and perhaps moved to "Near Term Backlog" per this AM's UX meeting? /cc @eloquence

eloquence:
[...]

PR #5629, opened 2020-11-15
Adds option to set and use an organization name

### LINKED ISSUE #1575: Add injectable organization name to source/journalist interface
As suggested by @ninavizz here https://github.com/freedomofpress/securedrop/issues/1524, source templates could be improved by including the organization name as string in various copy and flashed messages.

zenmonkeykstop:
## Status

 Ready for review

## Description of Changes

Fixes #1575.

- adds `organization_name` field to the instance configuration table
- adds option to set said field in the Admin interface
- updates SI and JI templates to use the field value in page titles, logo image alt texts, and SI metadata endpoint
- cleans up some flash message handling duplication in the Admin interface.
- removes a potential 500 error when handling the submission preferences form response.

### dev environment

- log in to the JI with an admin account, navigate to the **Admin > Instance Config** page
- clear the organization name field and submit via **SET ORGANIZATION NAME**
  - [ ] verify that an error flash message is displayed with text "This field is required"
- update the organization name to a string with more than 64 characters and submit
  - [ ] verify that an error flash message is displayed warning about the character limit 
- update the organization name to a string with HTML code (and well under 64 char limit)
  - [ ] verify that the page title shows the new organization name with html entities escaped.  
-  update the organization name to a string with no HTML and less than 64 characters.
   - [ ] verify that the page title shows the new organization name
   - [ ] inspect the page logo image and verify that its alt text contains the new organization name
   - [ ] open the Source Interface homepage and verify that its title is prefixed with the new organization name
   - [ ] inspect the SI homepage logo image and verify that its alt text contains the new organization name
   - [ ] navigate to the SI `/lookup` page and verify that its title is prefixed with the new organization name
   - [ ] navigate to the SI `/metadata` endpoint and verify that that the JSON response is valid and contains an `organization_name` key with value set to the new organization name
- for extra QA points, get creative and try to break page rendering with funky organization names!

### staging environment (for apparmor)
- run `make staging` on this branch
- log in to the JI with an admin account, navigate to the **Admin > Instance Config** page, and update the organization name to a string with no HTML and less than 64 characters.
- [ ] verify that the `preferences saved` flash message is displayed correctly.

### upgrade scenario
- run through the upgrade scenario and verify that the organization name can be set and changed post-upgrade.

## Deployment

No special considerations for deployment - there is an added db column which should be handled by migration scripts.

Choose one of the following:


sssoleileraaa (commented):
I focused on reviewing for functionality and making sure I didn't find any bugs. Also here are some followup UX improvements, which I don't believe are blocking, since they have a higher-level scope than the change in this PR:

* The error message shows up below the field and scrolls the user back to the top of the page so that you probably won't see the error message (it shows up for me when I display the page at 67% normal size). This is also how the error for uploading the logo works: "You can only upload PNG image files" shows up but you are scrolled to the top of the page, so it feels as if it has worked, especially when the error is out of scope.

* The config options seem out of order to me. For instance, updating the org name and logo seem to go together and like they belong at the top of the page so that it's ordered least-to-most technically advanced, probably: (1) Organization Name, (2) Logo, (3) Submission Preferences, (4) Alerts

* Another UX improvment that I think could, proabably should, be done in this PR is to explain what setting the org name does. For example, instead of just "Set your organization name" we could say above it something like: "The name of your organization is displayed in the page title and logo alt text for both the Source and Journalist interface."

### dev environment

- log in to the JI with an admin account, navigate to the **Admin > Instance Config** page
- clear the organization name field and submit via **SET ORGANIZATION NAME**
  - [x] verify that an error flash message is displayed with text "This field is required"
- update the organization name to a string with more than 64 characters and submit
  - [x] verify that an error flash message is displayed warning about the character limit 
- update the organization name to a string with HTML code (and well under 64 char limit)
  - [x] verify that the page title shows the new organization name with html entities escaped.  
-  update the organization name to a string with no HTML and less than 64 characters.
   - [x] verify that the page title shows the new organization name
   - [x] inspect the page logo image and verify that its alt text contains the new organization name
   - [x] open the Source Interface homepage and verify that its title is prefixed with the new organization name
   - [x] inspect the SI homepage logo image and verify that its alt text contains the new organization name
   - [x] navigate to the SI `/lookup` page and verify that its title is prefixed with the new organization name
- for extra QA points, get creative and try to break page rendering with funky organization names!
   - [x] i was unsuccessful

### staging environment (for apparmor)
- run `make staging` on this branch
- log in to the JI with an admin account, navigate to the **Admin > Instance Config** page, and update the organization name to a string with no HTML and less than 64 characters.
- [x] verify that the `preferences saved` flash message is displayed correctly.

### upgrade scenario
- run through the upgrade scenario and verify that the organization name can be set and changed post-upgrade.
   - :cyclone: IN PROGRESS

I still have the code review to do...

sssoleileraaa:
> * The error message shows up below the field and scrolls the user back to the top of the page so that you probably won't see the error message (it shows up for me when I display the page at 67% normal size). This is also how the error for uploading the logo works: "You can only upload PNG image files" shows up but you are scrolled to the top of the page, so it feels as if it has worked, especially when the error is out of scope.

This has now been addressed. Just one nit to add more space between the blurb and the text box with the organization name. Also you made me think of an enhancement to make (outside of this PR) to add the current name of the organization as placeholder text.

> * The config options seem out of order to me. For instance, updating the org name and logo seem to go together and like they belong at the top of the page so that it's ordered least-to-most technically advanced, probably: (1) Organization Name, (2) Logo, (3) Submission Preferences, (4) Alerts

This has now been addressed.

> * Another UX improvment that I think could, proabably should, be done in this PR is to explain what setting the org name does. For example, instead of just "Set your organization name" we could say above it something like: "The name of your organization is displayed in the page title and logo alt text for both the Source and Journalist interface."

This has now been addressed.

---

As of now I just need more time to finish review of code and to rerun the staging tests, so will update soon!

zenmonkeykstop:
vertical space? I thought I added a `<br>` in there already so horizontal space should not be an issue

ninavizz:
Oh my, this feels very important to call-out in the Admin UI, as needing to follow guidelines we lay forth in the docs. Also nervous about this shipping, before we've resolved how to handle this in the Source UI. I feel they should both deploy, together; as folks need to see how it will be surfaced, before entering in their information.

zenmonkeykstop:
@ninavizz as it stands, the Source Interface impact is in the page titles and logo alt text - the logo alt text is as described in #4280 - these updates would all deploy together but not have any impact unless/until an org chose to change the setting.

ninavizz:
@zenmonkeykstop Yeah, I feel that even that small a thing wd need user testing. I don't want to ship things to the Source UI, w/o being validated by users, first. And, tbh, near-term planned pipeline fixes to the SI will open up other opportunities to name the org inline with page text, that will improve user cognition of how SecureDrop works, significantly. I'm THRILLED this was done for Aaron Swartz Day, as this has been a biggy comprehension blocker in groking what this SecureDrop thing is, I feel, for a long time. Looking forward to having a synchronous dialog about this, soon!

sssoleileraaa (dismissed):
Ran through the test plan again, including the upgrade-scenario and everything works as expected. looks good! :shipit:

sssoleileraaa:
oh, i was so involved with testing this (this test plan took a long time since i had to rebuild staging and rerun through the upgrade plan multiple times) i missed the conversation between nina an zenmonkey... reading backsroll and see that we should chat more first. @emkll what do you think, is this an after-standup sort of discusssion?

eloquence:
Overall I think the UX changes here are pretty minimal and I'm optimistic we can land this with a bit more discussion. I do think the change of the page title and alt text in the Source Interface has some potentially problematic side effects. 

- The page title will change to "Org Name | Protecting Journalists and Sources". Typically, if organizations put a motto after their name, it's their own motto, and this could be seen as potentially conflating our branding with theirs. Maybe when the name is set it should simply say "Org Name | SecureDrop".
- The same potential for brand conflation applies when the SecureDrop logo is still set to the default logo. Perhaps the name should only be used here if the logo has in fact been customized? This one's a bit less of an issue.
- I've not observed any other uses of the org name yet, but they'd have to be scrutinized similarly.

Let's discuss a bit more here on the PR & synchronously. @ninavizz, I'm not sure I agree we would need user research to validate this change, but I do think we need to at least resolve some of the potential side effects.

zenmonkeykstop:
- That title change makes sense. Title is now "SecureDrop | Protecting Journalists and Sources" if the org name is unset or set to SecureDrop, and "Org Name | SecureDrop" otherwise.
- I'm less inclined to agree on the image point. Alt text is an alternative to the image, more likely to be displayed in situations where the logo isn't legible anyway, and the accessibility benefit of setting it is more tangible than the potential downside of branding confusion.
- The only other place it's used in the SI currently is in the metadata endpoint, which IMO is pretty innocuous. It's useful to be able to access it in a structured format and it's nothing you couldn't get by scraping the page itself. (though I realize now I should have added it in the test plan - will do so)

eloquence:
Chatted a bit about this w/ @emkll and @ninavizz last week. From a UX perspective this is good to land in the current form; we can always iterate further - thanks much for working on it Kev!

rmol (changes requested):
The test plan went fine. It is possible to insert control characters in the organization name, and while they do seem to be properly handled in the UI and the metadata response, they can be problematic elsewhere. Querying the `instance_config` table in a SQLite shell can result in some garbled terminal output.  I haven't spotted a way this would be more than annoying, but in addition to escaping the name supplied by the admin, maybe we could add a regular expression validator to the form. Something like `"^[-:.\w\s]+$"` covers all the current sites in our directory and some random tests of Unicode characters.

I also think we should skip the organization name in the alt text of the site logo. It's misleading if the logo is left at the default. A screenreader will read the organization name in the title of the page, so it's also not strictly necessary.
## Testing

### dev environment

- log in to the JI with an admin account, navigate to the **Admin > Instance Config** page
- clear the organization name field and submit via **SET ORGANIZATION NAME**
  - [x] verify that an error flash message is displayed with text "This field is required"
- update the organization name to a string with more than 64 characters and submit
  - [x] verify that an error flash message is displayed warning about the character limit
- update the organization name to a string with HTML code (and well under 64 char limit)
  - [x] verify that the page title shows the new organization name with html entities escaped.
-  update the organization name to a string with no HTML and less than 64 characters.
   - [x] verify that the page title shows the new organization name
   - [x] inspect the page logo image and verify that its alt text contains the new organization name
   - [x] open the Source Interface homepage and verify that its title is prefixed with the new organization name
   - [x] inspect the SI homepage logo image and verify that its alt text contains the new organization name
   - [x] navigate to the SI `/lookup` page and verify that its title is prefixed with the new organization name
   - [x] navigate to the SI `/metadata` endpoint and verify that that the JSON response is valid and contains an `organization_name` key with value set to the new organization name
- for extra QA points, get creative and try to break page rendering with funky organization names!

### staging environment (for apparmor)
- run `make staging` on this branch
- log in to the JI with an admin account, navigate to the **Admin > Instance Config** page, and update the organization name to a string with no HTML and less than 64 characters.
- [x] verify that the `preferences saved` flash message is displayed correctly.

zenmonkeykstop:
It is possible to test for the existence of a custom logo and only update the alt text then, but it seems like overkill and being way too precious about the brand. (There's nothing to stop someone just uploading the current logo as a custom logo, or (more likely) combining the SD logo with their own. Detecting that would be possible but extremely silly.) 

If you mean just leave it blank altogether, looking at w3c guidelines, specifically https://www.w3.org/WAI/tutorials/images/decision-tree/ would seem to suggest that the best thing to do (as it is an image with a link) would be to have text like "<orgname> Securedrop: Home", to allow screenreaders to describe the behavior of the link. The only situation where we would just leave it blank is in the case where it was purely decorative.

I do like the idea of improving text field sanitisation with a regex to catch non-printing chars - IMO this should be applied to other fields as well, including first/lastnames. I'm inclined to say that it should be a separate PR in that case, as otherwise the testing scope of this one will include all those fields too.

rmol:
>  the best thing to do (as it is an image with a link) would be to have text like " Securedrop: Home", to allow screenreaders to describe the behavior of the link

Yes.

> I do like the idea of improving text field sanitisation with a regex to catch non-printing chars - IMO this should be applied to other fields as well, including first/lastnames. I'm inclined to say that it should be a separate PR in that case

Yes, good call. I'll put in a separate ticket for that.

rmol (approved):

Issue #5637, opened 2020-11-17
gpg output change causing 3 errors on Focal testinfra tests


kushaldas:
## Steps to Reproduce

- [ ] `molecule test -s libvirt-staging-focal`

## Expected Behavior

- All tests should pass

## Actual Behavior

```Python
    ________________ test_gpg_key_in_keyring[ansible://app-staging] ________________
    [gw2] linux -- Python 3.7.3 /root/code/securedrop/.venv/bin/python3

    host = <testinfra.host.Host ansible://app-staging>

        @pytest.mark.skip_in_prod
        def test_gpg_key_in_keyring(host):
            """ ensure test gpg key is present in app keyring """
            with host.sudo(sdvars.securedrop_user):
                c = host.run("gpg --homedir /var/lib/securedrop/keys "
                             "--list-keys 28271441")
    >           assert "pub   4096R/28271441 2013-10-12" in c.stdout
    E           assert 'pub   4096R/28271441 2013-10-12' in 'pub   rsa4096 2013-10-12 [SC]\n      65A1B5FF195B56353CC63DFFCC40EF1228271441\nuid           [ unknown] SecureDrop Test/Development (DO NOT USE IN PROD
UCTION)\nsub   rsa4096 2013-10-12 [E]\n\n'
    E            +  where 'pub   rsa4096 2013-10-12 [SC]\n      65A1B5FF195B56353CC63DFFCC40EF1228271441\nuid           [ unknown] SecureDrop Test/Development (DO NOT USE IN PRODUCTION)\nsub   rsa4096 2013-10-12 
[E]\n\n' = CommandResult(command=b"sudo -u www-data /bin/sh -c 'gpg --homedir /var/lib/securedrop/keys --list-keys 28271441'", ex...96 2013-10-12 [E]\n\n', stderr=b"Warning: Permanently added '192.168.121.231' (E
CDSA) to the list of known hosts.\r\n").stdout

    ../testinfra/app/test_appenv.py:75: AssertionError
    _____________ test_fpf_apt_repo_fingerprint[ansible://app-staging] _____________
    [gw2] linux -- Python 3.7.3 /root/code/securedrop/.venv/bin/python3

    host = <testinfra.host.Host ansible://app-staging>

        def test_fpf_apt_repo_fingerprint(host):
            """
            Ensure the FPF apt repo has the correct fingerprint on the associated
            signing pubkey. The key changed in October 2016, so test for the
            newest fingerprint, which is installed on systems via the
[...]

Issue #5635, opened 2020-11-17
test_listening_ports failure on Focal


kushaldas:
## Description



## Steps to Reproduce

- [ ] `molecule test -s libvirt-staging-focal`

## Expected Behavior

All tests should pass

## Actual Behavior

```Python
    __________ test_listening_ports[ansible://mon-staging-ossec_service1] __________                                                                                                                      [126/1908]
    [gw1] linux -- Python 3.7.3 /root/code/securedrop/.venv/bin/python3                                                                                                                                             
                                                                                                                                                                                                                    
    host = <testinfra.host.Host ansible://mon-staging>                                                                                                                                                              
    ossec_service = {'host': '0.0.0.0', 'listening': True, 'port': 1514, 'proto': 'udp'}                                                                                                                            
                                                                                                                                                                                                                    
        @pytest.mark.skip_in_prod                                                                                                                                                                                   
        @pytest.mark.parametrize('ossec_service', [                                                                                                                                                                 
            dict(host="0.0.0.0", proto="tcp", port=22, listening=True),                                                                                                                                             
            dict(host="0.0.0.0", proto="udp", port=1514, listening=True),                                                                                                                                           
            dict(host="0.0.0.0", proto="tcp", port=1515, listening=False),                                                                                                                                          
        ])                                                                                                                                                                                                          
        def test_listening_ports(host, ossec_service):                                                                                                                                                              
            """                                                                                                                                                                                                     
            Ensure the OSSEC-related services are listening on the                                                                                                                                                  
            expected sockets. Services to check include ossec-remoted                                                                                                                                               
            and ossec-authd. Helper services such as postfix are checked                                                                                                                                            
            separately.                                                                                                                                                                                             
                                                                                                                                                                                                                    
            Note that the SSH check will fail if run against a prod host, due                                                                                                                                       
            to the SSH-over-Tor strategy. We can port the parametrized values                                                                                                                                       
            to config test YAML vars at that point.                                                                                                                                                                 
            """                                                                                                                                                                                                     
            socket = "{proto}://{host}:{port}".format(**ossec_service)                                                                                                                                              
            with host.sudo():                                                                                                                                                                                       
                # Really hacky work-around for bug found in testinfra 1.12.0                                                                                                                                        
                # https://github.com/philpep/testinfra/issues/311                                                                                                                                                   
                if "udp" in socket:                                                                                                                                                                                 
[...]

Issue #5510, opened 2020-09-17
Add Focal staging-test-with-rebase CI job


kushaldas:
## Description

We need a CI job to run against Focal which will run both testinfra tests

Just like `staging-with-rebase` it will do  `staging-with-rebase-focal`.  This can only be done after all other Focal EPIC tickets are resolved.

emkll:
closed via https://github.com/freedomofpress/securedrop/pull/5638

Issue #5509, opened 2020-09-17
Enable testinfra tests against Focal staging


kushaldas:
## Description

We currently do not have `testinfra` tests running against `libvirt-staging-focal` environment. We should enable those and fix any issues found.

emkll:
Molecule is running the testinfra tests as of #5486. Fixing the tests is in scope of #5510 and will be addressed in https://github.com/freedomofpress/securedrop/pull/5638. Closing

PR #5638, opened 2020-11-17
Add staging-with-rebase-focal to CI and fix testinfra tests

### LINKED ISSUE #5636: test_apparmor.py failure on Focal
## Steps to Reproduce

- [ ] `molecule test -s libvirt-staging-focal`

## Expected Behavior

- [ ] tests should pass

## Actual Behavior

```Python
    _________ test_apparmor_enforced[ansible://app-staging-/sbin/dhclient] _________
    [gw0] linux -- Python 3.7.3 /root/code/securedrop/.venv/bin/python3

    host = <testinfra.host.Host ansible://app-staging>
    aa_enforced = '/sbin/dhclient'

        @pytest.mark.parametrize('aa_enforced', sdvars.apparmor_enforce)
        def test_apparmor_enforced(host, aa_enforced):
            awk = ("awk '/[0-9]+ profiles.*enforce./"
                   "{flag=1;next}/^[0-9]+.*/{flag=0}flag'")
            with host.sudo():
                c = host.check_output("aa-status | {}".format(awk))
    >           assert aa_enforced in c
    E           AssertionError: assert '/sbin/dhclient' in '   /usr/bin/man\n   /usr/lib/NetworkManager/nm-dhcp-client.action\n   /usr/lib/NetworkManager/nm-dhcp-helper\n   /usr...bin/dhclient\n   lsb_release\n  
 man_filter\n   man_groff\n   nvidia_modprobe\n   nvidia_modprobe//kmod\n   system_tor'

    ../testinfra/app/test_apparmor.py:102: AssertionError

```

## Comments

Most probably the `dhclient` path has changed in `Focal`.

kushaldas:
## Status

Ready for review

## Description of Changes

Fixes #5636 #5635 #5637 #5655 #5510 

- Adds CI target to staging reabase on Focal
- Disables `systemd-resolved` on Focal and uses `/etc/resolv.conf` file.
- Updates test to find ports open for listening in mon server
- Uses single `dhclient` path for both Xenial and Focal testing
- Updates standard gpg output tests on servers for both Xenial and Focal.
- OSSEC used `sysv` script on Xenial, and `ossec.service` systemd style service file on Focal.
- Adds new test verify how `ossec` service started on Xenial and Focal.

<https://kushaldas.in/posts/story-of-debugging-exit-0.html> describes why do we need two different service files for `ossec-sever` and `ossec-agent`. 

## Testing

- [x] CI is green.
- [x] `molecule test -s libvirt-staging-focal`  should not have any test failure on `test_apparmor.py`, there will 4 other test failure due to different bugs.
- [x] Ossec logs are flowing between app and mon, and this persists reboots
- [x] Changes to the DNS configuration make sense here (we are reusing the existing resolv.conf templates)

Choose one of the following:


emkll (dismissed):
Thanks for opening @kushaldas as well as #5635, #5636 and #5637 to track specific failures.

As part of the commitment for this sprint of a timebox around https://github.com/freedomofpress/securedrop/issues/5509 , I recommend we proceed as follows, since we currently have no way of reliably tracking failure count on Focal:

1. Amend this PR to add a CircleCI task for `staging-with-rebase-focal` (it's OK if we merge with `staging-with-rebase-focal` at the end of our timebox) it seems  best we track failures/improvements over time by having CI run it instead of a developer running these locally.  
(note that the existing GCP CI base image should already have Vagrant 20.04 boxes included, so the change should be relatively straightforward) 
2. Continue addressing the test failures in this working branch (including the issues you've opened, and whichever other testinfra failures present in Focal

kushaldas:
Right now, if you login to the `mon-staging` and then do `systemctl status ossec`, you will find it `inactive(dead)`. This should be starting up.   If you manually start the service, then everything is working.

install_files/ansible-base/roles/ossec/tasks/configure_server.yml
```diff
@@ -113,3 +113,11 @@
     creates: /var/ossec/etc/sslmanager.cert
   tags:
     - ossec_auth
+
+- name: Copy the systemd service file
+  copy:
+    src: ossec@.service
+    dest: "/etc/systemd/system/ossec@.service"
+
+- name: Enable the service
```
emkll: Here you should use the ansible built-in systemd task:

```
- name : Enable ossec service to run
  systemd:
    name: "{{ item }}"
    enabled: yes
    masked: no
  with_items: "{{ some var in defaults/main.yml }}"
```

install_files/ansible-base/roles/ossec/handlers/main.yml
```diff
@@ -1,6 +1,4 @@
 ---
 # Single handler to operate on *both* OSSEC hosts, server & client.
 - name: restart ossec
-  service:
-    name: ossec
-    state: restarted
+  command: systemctl restart ossec
```
emkll: maybe `systemd` ansible built-in here as well

emkll (commented):
Thanks for the changes @kushaldas , left some inline comments. I've also appended a commit to clean up the test logic (lsof version is different, so the UDP check would always fail on Focal).

There is, however, still an issue with the ossec service on `mon`, and the service does not start due to one of the services (`ossec@maild.service`) that is failing. By running `systemctl list-units --all` , you can observe the units that are not started, and then check the status of them. The only failed service is ``ossec@maild.service``, and then by checking its status, the root cause of the failure is clear:

```
ossec-maild [dns]: ERROR: connect() failed.
ossec-maild: ERROR: DNS failure for smtpserver
ossec-maild: ERROR: No socket.
ossec-maild: ERROR: Error Sending email to 127.0.0.1 (smtp server)
```

This might be related to #5655, perhaps we should pull in the fix for that issue in this PR, what do you think?

kushaldas:
@emkll I think the PR is in a better state now, will know for sure after all CI jobs finish :)

kushaldas:
I still have the SMTP issues:

```
Nov 30 12:26:00 mon-staging env[761]: 2020/11/30 12:26:00 ossec-maild(1223): ERROR: Error Sending email to 127.0.0.1 (smtp server)
Nov 30 12:26:15 mon-staging env[927]: 2020/11/30 12:26:15 ossec-maild: DEBUG: Running OS_Sendmail()
Nov 30 12:26:15 mon-staging env[762]: 2020/11/30 12:26:15 ossec-maild [dns]: ERROR: connect() failed.
Nov 30 12:26:15 mon-staging env[927]: 2020/11/30 12:26:15 ossec-maild: ERROR: DNS failure for smtpserver
Nov 30 12:26:15 mon-staging env[927]: 2020/11/30 12:26:15 ossec-maild: ERROR: No socket.
Nov 30 12:26:20 mon-staging env[761]: 2020/11/30 12:26:20 ossec-maild(1261): ERROR: Waiting for child process. (status: 256).
Nov 30 12:26:20 mon-staging env[761]: 2020/11/30 12:26:20 ossec-maild(1223): ERROR: Error Sending email to 127.0.0.1 (smtp server)
Nov 30 12:26:20 mon-staging env[761]: 2020/11/30 12:26:20 ossec-maild(1262): ERROR: Too many errors waiting for child process(es).
Nov 30 12:26:20 mon-staging env[761]: 2020/11/30 12:26:20 ossec-maild(1223): ERROR: Error Sending email to 127.0.0.1 (smtp server)
```
@emkll any tips about which service should I look at?

kushaldas:
I am rebuilding `xenial` staging on develop to check what is happening.

kushaldas:
Wondering if https://github.com/ossec/ossec-hids/pull/1891 is involved here?

emkll:
Rebased on latest `develop` to include changes from https://github.com/freedomofpress/securedrop/pull/5658

emkll:
Good news, the box update provided in #5658 has resolved the issue tracked in #5642  , one last test is failing, the service not running/listening to the port. 

Bad news, some ossec units are not staying up, including remoted which receives the logs. While I haven't uncovered the root cause, here are some findings:

- If you `molecule converge` and run `sudo service ossec start` on mon server, the test _does_ pass, which suggests there may be an issue with the systemd units

- I've observed a log error when running `sudo systemctl start ossec@remoted` and then `systemctl status ossec@remoted`, ossec-remoted would only bind to an IPv4 address. Applying the following patch https://github.com/freedomofpress/securedrop/commit/d59cd835fb8c75656937005ea542731712b8b6d4 seems to now show the binding to IPv4 in the `systemctl status` call, but doesn't make the tests green. It seems like `ossec-remoted` it receives SIGTERM from something, and is immediately killed. As a result, the test still fails.

Could there be, perhaps, some contention with the existing init scripts that are set by default in the server `postinst` here: https://github.com/freedomofpress/securedrop/blob/fix_dhclient_path_for_testinfra/install_files/ossec-server/DEBIAN/postinst#L196-L198

kushaldas:
Here is the journalctl log from `ossec@remoted`

```
-- Logs begin at Tue 2020-12-01 09:56:35 UTC, end at Tue 2020-12-01 10:18:15 UTC. --
Dec 01 10:08:30 mon-staging systemd[1]: Starting The OSSEC HIDS remoted server...
Dec 01 10:08:31 mon-staging systemd[1]: Started The OSSEC HIDS remoted server.
Dec 01 10:08:31 mon-staging env[762]: 2020/12/01 10:08:31 ossec-remoted: INFO: Started (pid: 762).
Dec 01 10:08:31 mon-staging env[763]: 2020/12/01 10:08:31 IPv6: :: on port 1514
Dec 01 10:08:31 mon-staging env[763]: 2020/12/01 10:08:31 Socket bound for IPv6: :: on port 1514
Dec 01 10:08:31 mon-staging env[763]: 2020/12/01 10:08:31 ossec-remoted: INFO: Started (pid: 763).
Dec 01 10:08:31 mon-staging env[763]: 2020/12/01 10:08:31 ossec-remoted(1225): INFO: SIGNAL [(15)-(Terminated)] Received. Exit Cleaning...
Dec 01 10:08:31 mon-staging systemd[1]: ossec@remoted.service: Succeeded.
Dec 01 10:18:15 mon-staging systemd[1]: Starting The OSSEC HIDS remoted server...
Dec 01 10:18:15 mon-staging systemd[1]: Started The OSSEC HIDS remoted server.
Dec 01 10:18:15 mon-staging env[1136]: 2020/12/01 10:18:15 ossec-remoted: INFO: Started (pid: 1136).
Dec 01 10:18:15 mon-staging env[1137]: 2020/12/01 10:18:15 IPv6: :: on port 1514
Dec 01 10:18:15 mon-staging env[1137]: 2020/12/01 10:18:15 Socket bound for IPv6: :: on port 1514
Dec 01 10:18:15 mon-staging env[1137]: 2020/12/01 10:18:15 ossec-remoted: INFO: Started (pid: 1137).
Dec 01 10:18:15 mon-staging env[1137]: 2020/12/01 10:18:15 ossec-remoted(1225): INFO: SIGNAL [(15)-(Terminated)] Received. Exit Cleaning...
Dec 01 10:18:15 mon-staging systemd[1]: ossec@remoted.service: Succeeded.
```

kushaldas:
More strange log difference between service and the systemd:

```
root@mon-staging:/home/vagrant# systemctl status ossec@remoted
● ossec@remoted.service - The OSSEC HIDS remoted server
     Loaded: loaded (/etc/systemd/system/ossec@.service; enabled; vendor preset: enabled)
     Active: inactive (dead) since Tue 2020-12-01 10:18:15 UTC; 5min ago
    Process: 1135 ExecStartPre=/usr/bin/env /var/ossec/bin/ossec-remoted -t (code=exited, status=0/SUCCESS)
    Process: 1136 ExecStart=/usr/bin/env /var/ossec/bin/ossec-remoted -f (code=exited, status=0/SUCCESS)
   Main PID: 1136 (code=exited, status=0/SUCCESS)

Dec 01 10:18:15 mon-staging systemd[1]: Starting The OSSEC HIDS remoted server...
Dec 01 10:18:15 mon-staging systemd[1]: Started The OSSEC HIDS remoted server.
Dec 01 10:18:15 mon-staging env[1136]: 2020/12/01 10:18:15 ossec-remoted: INFO: Started (pid: 1136).
Dec 01 10:18:15 mon-staging env[1137]: 2020/12/01 10:18:15 IPv6: :: on port 1514
Dec 01 10:18:15 mon-staging env[1137]: 2020/12/01 10:18:15 Socket bound for IPv6: :: on port 1514
Dec 01 10:18:15 mon-staging env[1137]: 2020/12/01 10:18:15 ossec-remoted: INFO: Started (pid: 1137).
Dec 01 10:18:15 mon-staging env[1137]: 2020/12/01 10:18:15 ossec-remoted(1225): INFO: SIGNAL [(15)-(Terminated)] Received. Exit Cleaning...
Dec 01 10:18:15 mon-staging systemd[1]: ossec@remoted.service: Succeeded.
root@mon-staging:/home/vagrant# ps aux | grep ossec
ossecm      1896  0.0  0.2   4108  2612 ?        S    10:22   0:00 /var/ossec/bin/ossec-maild
ossecm      1897  0.0  0.2   4108  2672 ?        S    10:22   0:00 /var/ossec/bin/ossec-maild
ossec       1905  0.0  0.4   5648  4496 ?        S    10:22   0:00 /var/ossec/bin/ossec-analysisd
ossecr      1914  0.0  0.2  20268  2620 ?        Sl   10:22   0:00 /var/ossec/bin/ossec-remoted
root        1915  0.0  0.2   3572  1972 ?        S    10:22   0:00 /var/ossec/bin/ossec-logcollector
root        1920  0.0  0.0   3744   464 ?        S    10:22   0:00 /var/ossec/bin/ossec-syscheckd
ossec       1924  0.0  0.1   3808  1480 ?        S    10:22   0:00 /var/ossec/bin/ossec-monitord
ossecm      1976  0.0  0.0      0     0 ?        Z    10:23   0:00 [ossec-maild] <defunct>
root        1978  0.0  0.0   8908   668 pts/1    S+   10:23   0:00 grep --color=auto ossec

```

install_files/ansible-base/roles/ossec/tasks/configure_server.yml
```diff
@@ -113,3 +113,20 @@
...
     - ossec_auth
+
+- name: Copy the systemd service file
+  copy:
+    src: ossec.service
+    dest: "/etc/systemd/system/ossec.service"
+
+- name: Remove the old style /etc/init.d/ossec file
+  file:
+    path: "/etc/init.d/ossec"
```
emkll: These files are provided by ossec-{agent,server} packages:

```
.
├── etc
│   ├── init.d
│   │   └── ossec
│   └── ossec-init.conf

```

This means that on subsequent ossec package upgrades, this init file will be re-added, unless we do specific tasks at build-time to remove it. If these steps are strictly required for functioning, we should address this in the build logic  as to not provide the init.d file in focal (instead of doing it in ansible)

emkll (commented):
Thanks @kushaldas this is working for me, and can confirm logs are flowing between app and mon on Focal, and CI is green :tada: 

Opened #5660 and #5661 to track unrelated Focal follow-up issues .

I left a comment inline that should be worth addressing. it's unclear to me from your commit message if a part of the issue observed was some sort of init/systemd contention. Regardless, the change provided won't persist upgrades, if we don't make any changes.

We should ensure that changes to DNS (moving away from the default systemd-resolvd and back to resolv.conf) given that it will break unattended upgrades (as identified in #5655), that would be especially problematic. @conorsch what do you think?

install_files/ansible-base/roles/ossec/tasks/configure_server.yml
kushaldas: > This means that on subsequent ossec package upgrades, this init file will be re-added, unless we do specific tasks at build-time to remove it. If these steps are strictly required for functioning, we should address this in the build logic as to not provide the init.d file in focal (instead of doing it in ansible)

I think we should remove it from the server package.

conorsch: Definitely agree, let's remove the Ansible task and clean up that file at the packaging level. Be mindful of side-effects for Xenial.

install_files/ansible-base/roles/common/tasks/harden_dns.yml
```diff
@@ -6,3 +6,22 @@
...
+    - hardening
+
+- name: Create resolv.conf for Focal
+  template:
+    src: dns_base
+    dest: /etc/resolv.conf
+  when: ansible_distribution_release == 'focal'
+  tags:
+    - dns
+    - hardening
```
conorsch: No objections to disabling `systemd-resolved`, sticking with resolvconf which we've been using for a while will be fairly straightforward. Since we just copied the same `dns_base` source a few lines above, let's use that task to write the file. Sounds like on Xenial we want the /etc/resolvconf/resolve.conf.d/ path, whereas on Focal we should write it directly to `/etc/resolv.conf`.

conorsch (commented):
Took a pass through with local VMs, looking good! Agreed with the packaging logic cleanup requested by @emkll. Overall the DNS changes appear reliable, although let's keep the customizations to a minimum. For instance, it appears that on Focal we don't need /etc/resolvconf/resolv.conf.d/base at all, so let's not write to it.

emkll (commented):
There are a couple of areas that are using service instead of systemctl, all of which appear to be still working:
- Ossec registration logic, which is still working, as the service is actually being started and the app server is successfully registered https://github.com/freedomofpress/securedrop/blob/fix_dhclient_path_for_testinfra/install_files/ansible-base/roles/ossec/tasks/register.yml#L40
- https://github.com/freedomofpress/securedrop/blob/fix_dhclient_path_for_testinfra/install_files/securedrop-ossec-agent/DEBIAN/postinst#L25
- https://github.com/freedomofpress/securedrop/blob/fix_dhclient_path_for_testinfra/install_files/securedrop-ossec-server/DEBIAN/postinst#L37

- https://github.com/freedomofpress/securedrop/blob/fix_dhclient_path_for_testinfra/install_files/ossec-server/etc/init.d/ossec (if we keep split xenial/focal logic, this should be preserved for Xenial)
- https://github.com/freedomofpress/securedrop/blob/fix_dhclient_path_for_testinfra/install_files/ossec-agent/etc/init.d/ossec (if we keep split xenial/focal logic, this should be preserved for Xenial)

If we are to move to systemd for Focal (which I think is a good idea, as you've suggested), we still need to preserve init.d/service support under Xenial (unless we carefully handle the service enabling/starting under Xenial as well). This means that we will need to conditionally handle the service logic here (init in Xenial and systemd in Focal). 


It could perhaps be helpful to take a step back and map out these changes to better understand implications, but at at strict minimum, prior to merging this PR, we should ensure no changes to Xenial, in other words, the systemctl logic you have introduced for Focal support should _not_ be applied to Xenial hosts as to not break existing installs during upgrades.

install_files/ansible-base/roles/common/tasks/harden_dns.yml
conorsch: Added a commit to consolidate the logic here a bit: under Focal, the old `/etc/resolvconf/resolve.conf.d/` path is no longer written to, and the tests now inspect the correct file based on distro.

conorsch (dismissed):
Looking solid! Banged on the resolvconf settings, and the approach you took to use a flat file in `/etc/resolv.conf` is definitely the best choice, among many (https://manpages.ubuntu.com/manpages/focal/man8/systemd-resolved.8.html). Overall the diff on this PR is rather small, given how many issues it closes :slightly_smiling_face: 

I've got just one remaining request prior to merge: given the sysv -> systemd transition for ossec-server, let's also make the same update for ossec-agent, so we can reuse the same bootstrapping logic and even reuse the same tests to verify state on both machines. I took the liberty of pushing a new commit which copies over the ossec-server test to ossec-agent, which should cause CI to fail again :frowning_face: . Once that change is made in the packaging logic, and CI is green again, no further requests from me, we're good to go!

kushaldas:
I made the change to have `ossec.service` file for the `ossec-agent` package too. But, when the `securedrop-ossec-agent` package gets installed, in the post-install script, it is running to start the service `ossec`. That fails because `/var/ossec/etc/client.keys` is missing. It is same even if I try to start the service manually. 

```
root@app-staging:/home/vagrant# /var/ossec/bin/ossec-control start
Starting OSSEC HIDS v3.6.0...
Started ossec-execd...
2020/12/04 13:12:22 ossec-agentd: INFO: Using notify time: 600 and max time to reconnect: 1800
2020/12/04 13:12:22 ossec-agentd(1402): ERROR: Authentication key file '/var/ossec/etc/client.keys' not found.
2020/12/04 13:12:22 ossec-agentd(1751): ERROR: File client.keys not found or empty.
2020/12/04 13:12:22 ossec-agentd(4109): ERROR: Unable to start without auth keys. Exiting.
ossec-agentd did not start
root@app-staging:/home/vagrant# /var/ossec/bin/ossec-agentd status
2020/12/04 13:12:30 ossec-agentd: INFO: Using notify time: 600 and max time to reconnect: 1800
2020/12/04 13:12:30 ossec-agentd(1402): ERROR: Authentication key file '/var/ossec/etc/client.keys' not found.
2020/12/04 13:12:30 ossec-agentd(1751): ERROR: File client.keys not found or empty.
2020/12/04 13:12:30 ossec-agentd(4109): ERROR: Unable to start without auth keys. Exiting.
root@app-staging:/home/vagrant# /var/ossec/bin/ossec-agentd
2020/12/04 13:12:33 ossec-agentd: INFO: Using notify time: 600 and max time to reconnect: 1800
2020/12/04 13:12:33 ossec-agentd(1402): ERROR: Authentication key file '/var/ossec/etc/client.keys' not found.
2020/12/04 13:12:33 ossec-agentd(1751): ERROR: File client.keys not found or empty.
2020/12/04 13:12:33 ossec-agentd(4109): ERROR: Unable to start without auth keys. Exiting.
```

@conorsch @emkll Can you please have a look and see if you can find the reason? I could not figure out why it was being able to start the service before. Still looking.

kushaldas:
The is the service file generated and used by the `systemd-sysv-generator`

```
# Automatically generated by systemd-sysv-generator

[Unit]
Documentation=man:systemd-sysv-generator(8)
SourcePath=/etc/init.d/ossec
Description=LSB: Start daemon at boot time
After=remote-fs.target

[Service]
Type=forking
Restart=no
TimeoutSec=5min
IgnoreSIGPIPE=no
KillMode=process
GuessMainPID=no
RemainAfterExit=yes
SuccessExitStatus=5 6
ExecStart=/etc/init.d/ossec start
ExecStop=/etc/init.d/ossec stop
```

kushaldas:
Found the issue: Our `/etc/init.d/ossec` file has an "exit 0" line at the end https://github.com/freedomofpress/securedrop/blob/develop/install_files/ossec-agent/etc/init.d/ossec#L61 which is reason it worked till then. I will make full comment and  update the PR with a new .service file for only the agent on Monday morning.

molecule/testinfra/common/test_fpf_apt_repo.py
```diff
@@ -45,11 +45,7 @@ def test_fpf_apt_repo_fingerprint(host):
 
     c = host.run('apt-key finger')
 
-    fpf_gpg_pub_key_info = """/etc/apt/trusted.gpg.d/securedrop-keyring.gpg
----------------------------------------------
-pub   4096R/00F4AD77 2016-10-20 [expires: 2021-06-30]
-      Key fingerprint = 2224 5C81 E3BA EB41 38B3  6061 310F 5612 00F4 AD77
-uid                  SecureDrop Release Signing Key"""
+    fpf_gpg_pub_key_info = "2224 5C81 E3BA EB41 38B3  6061 310F 5612 00F4 AD77"
```
zenmonkeykstop, line 48: Having the full output seems more secure, especially if the testinfra tests are being used to verify a prod system.

zenmonkeykstop (approved):
- [x] Testinfra passes against focal and xenial scenarios with same number of pass/skip/xfail, no fails
- [x] OSSEC alerts flow (tested using test ossec alert in JI) in focal
- [x] OSSEC alerts continue to flow after app and mon reboot in focal

one nit re: string for GPG output in testinfra but I don't view it as a blocker.

molecule/testinfra/common/test_fpf_apt_repo.py
kushaldas: The output is different in both `Xenial` and `Focal`, that is why I took this path.

PR #5648, opened 2020-11-24
Update Tor to 0.4.4.6

emkll:
## Status

Ready for review

## Description of Changes

Towards #5624 .
Closes #5649 ?

## Testing
- [ ] Updated tor signing key is valid, only subkey was updated
- [ ] `make fetch-tor-debs` is working as expected, for both xenial and focal

## Deployment
These packages will be deployed once the deb packages are uploaded apt.freedom.press.
### If you made changes to the server application code:

- [x] Linting (`make lint`) and tests (`make test`) pass in the development container

### If you made non-trivial code changes:

- [x] I have written a test plan and validated it for this PR

Choose one of the following:

- [ ] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-docs) for these changes, or will do so later
- [ ] I would appreciate help with the documentation
- [x] These changes do not require documentation

PR #5651, opened 2020-11-25
Rename journalist key routes

### LINKED ISSUE #4421: "Journalist key" in source interface is not (typically) a journalist key
The *Source Interface* refers to "the journalist's public key", both through the `/journalist-key` and `/why-journalist-key` route names, and in the page heading on the `/why-journalist-key` page, "Why download the journalist's public key?"

This is very misleading wording, as most SecureDrops are operated by news organizations employing more than one journalist. This key is in fact the SecureDrop Submission Key for this installation, the _exact same_ key that messages/files will be encrypted to upon submission. (See #4413 for terminology standardization.)

Offering technical users the option to pre-encrypt is essentially a workaround until we establish a reasonable method for automatic client-side encryption (see #92).

To avoid messing up people's mental model of how SecureDrop works, we should rename this route, especially given that we do refer to managing individual journalist keys in parts of the documentation (for completely different reasons: to securely export files from the *Secure Viewing Station*).

DrGFreeman:
## Status

Ready for review

## Description of Changes

Fixes #4421 

Rename `why-journalist-key` and `journalist-key` routes of the Source Interface to `why-public-key` and `public-key` respectively. This aligns the route names with the language used in the associated pages where "public key" is used to designate the SecureDrop Submission Key of the instance.

Also rename the functions associated with the routes.

Maintain the `journalist-key` and redirect it to `public-key` in case it is used or linked by instance owners or sources. Add a unit test to verify the redirect.

## Testing

1. In the source interface, navigate to `/lookup`.
1. Hover over the *public key* link.
    - [ ] Verify the link points to `/public-key`.
1. Click the *public key* link.
    - [ ] Verify the key file downloads as expected.
1. Click the *learn more* link.
    - [ ] Verify the `/why-public-key` page loads.
1. Hover over the *Download* link.
    - [ ] Verify the link points to `/public-key`.
1. Click the *Download* link.
    - [ ] Verify the key file downloads as expected.
1. Using the browser address bar navigate to `/journalist-key`.
    - [ ] Verify the redirect works and the key file downloads as expected.

## Deployment

2. New installs.

Choose one of the following:

Ref.: https://github.com/freedomofpress/securedrop-docs/issues/111, https://github.com/freedomofpress/securedrop-docs/pull/112

Choose one of the following:


emkll:
Thanks @DrGFreeman for your contribution. This PR will need a rebase on latest develop to include the changes to from #5648 to get all CI tests passing.

securedrop/source_app/info.py
```diff
@@ -18,8 +18,8 @@ def tor2web_warning() -> str:
     def recommend_tor_browser() -> str:
         return render_template("use-tor-browser.html")
 
-    @view.route('/journalist-key')
```
zenmonkeykstop, line 21: Rather than removing this, it might be better to convert it to a permanent redirect, just on the off-chance it's being used/linked by instance owners or sources. Something like this:
```
diff --git a/securedrop/source_app/info.py b/securedrop/source_app/info.py
index df43164eb..4f1a83644 100644
--- a/securedrop/source_app/info.py
+++ b/securedrop/source_app/info.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 import flask
-from flask import Blueprint, render_template, send_file, current_app
+from flask import Blueprint, render_template, send_file, current_app, redirect, url_for
 
 from io import BytesIO  # noqa
 
@@ -18,6 +18,10 @@ def make_blueprint(config: SDConfig) -> Blueprint:
     def recommend_tor_browser() -> str:
         return render_template("use-tor-browser.html")
 
+    @view.route('/journalist-key')
+    def download_journalist_key() -> flask.Response:
+            return redirect(url_for('.download_public_key'), code=301)
+    
     @view.route('/public-key')
     def download_public_key() -> flask.Response:
         journalist_pubkey = current_app.crypto_util.gpg.export_keys(
```

zenmonkeykstop (changes requested):
Test plan passes, but it would be a good idea to include a redirect from the old `/journalist-key` to the new `/public-key` in case its in use anywhere. See patch in comment for simplest approach.

securedrop/source_app/info.py
DrGFreeman: Makes total sense!

DrGFreeman:
@zenmonkeykstop, thanks for the feedback. Good points that I'll consider in future PRs.

I restored the `/journalist-key` and redirected it to `/public-key` as suggested. I also added a test to verify the redirect.

Note that the default return type of `flask.redirect` is `werkzeug.wrappers.Response` instead of `flask.Response`. It can be [changed via `Response` parameter](https://flask.palletsprojects.com/en/1.1.x/api/?highlight=redirect#flask.redirect) however I don't see the purpose of changing it at this point. Let me know if you think otherwise.

I rebased the branch as requested by @emkll. I intended to do it before submitting the PR but I forgot. 😕

zenmonkeykstop (approved):
LGTM! Thanks for adding in the 301 on the old path.

PR #5665, opened 2020-12-03
Remove paths that are no longer in use

eloquence:
## Status

Ready for review

## Description of Changes
Removes paths that are no longer in use from OSSEC monitoring. As far as I can tell, the last trace of `/var/securedrop` use was removed in #4622.

## Testing/Deployment
Are there any reasons we want to continue to monitor those paths? If not, visual review and double-checking for any remaining references to those paths should be sufficient.


rmol (approved):
I can't find any remaining trace of `/var/securedrop`. I think it's safe to remove these lines.

PR #5666, opened 2020-12-09
Update and add annotations to sshd config for servers

### LINKED ISSUE #5660: Update sshd config for Focal
## Description

In the current sshd configuration, we specify a path to a host dsa key for the server, but that has been deprecated/disabled. This was uncovered by an ossec alert while reviewing https://github.com/freedomofpress/securedrop/pull/5638

The alert is as follows:

```
Dec 1 15:08:22 app-staging sshd[3302]: error: Unable to load host key: /etc/ssh/ssh_host_dsa_key
```

The server sshd config is
https://github.com/freedomofpress/securedrop/blob/1d50b6ecd4bc0cfe6c803630b7fb4f8d39bc3810/install_files/ansible-base/roles/restrict-direct-access/templates/sshd_config#L5

and removing the line should suffice. We should also review and audit the current configuration to ensure we adhere to current best practices.

emkll:
## Status

Work in progress

## Description of Changes

Fixes #5660 

- Update supported algorthms
- Disable some agent forwarding and tunnelling options
- Annotate and reorder configuration for readability

Sources:
- https://github.com/dev-sec/ansible-ssh-hardening
- https://github.com/arthepsy/ssh-audit / https://github.com/jtesta/ssh-audit/

## Testing

- [ ] provided config is good / adheres to best practices
- [ ] provided config is adequately annotated
- [ ] provided config eliminates log message described in #5660 
- [ ] on an existing Xenial install the config is applied properly when an Ansible run occurs (I have tested this)
#### New installs

The sshd configuration will be applied on new installs via Ansible

#### Existing installs

Existing installs will not be updated unattended, but applied when an Ansible run occurs

## Checklist

Choose one of the following:

### DIFF
--- a/install_files/ansible-base/roles/restrict-direct-access/templates/sshd_config
+++ b/install_files/ansible-base/roles/restrict-direct-access/templates/sshd_config
@@ -5,8 +5,6 @@ HostKey /etc/ssh/ssh_host_rsa_key
 HostKey /etc/ssh/ssh_host_ecdsa_key
 HostKey /etc/ssh/ssh_host_ed25519_key
 
-# Enforce privilege separation by creating unprivileged child process
-UsePrivilegeSeparation yes
 KeyRegenerationInterval 3600
 ServerKeyBits 4096
 
@@ -39,7 +37,7 @@ UseDNS no
 
 # Cipher selection
 
-Ciphers chacha20-poly1305@openssh.com,aes256-gcm@openssh.com,aes128-gcm@openssh.com,aes256-ctr,aes128-ctr
+Ciphers aes256-gcm@openssh.com,aes128-gcm@openssh.com,aes256-ctr,aes128-ctr
 # Don't use SHA1 for kex
 KexAlgorithms curve25519-sha256@libssh.org,diffie-hellman-group-exchange-sha256
 # Don't use SHA1 for hashing, don't use encrypt-and-MAC mode
--- a/install_files/ansible-base/roles/restrict-direct-access/templates/sshd_config
+++ b/install_files/ansible-base/roles/restrict-direct-access/templates/sshd_config
@@ -2,6 +2,7 @@ Port 22
 ListenAddress {{ ssh_listening_address }}:22
 Protocol 2
 HostKey /etc/ssh/ssh_host_rsa_key
+HostKey /etc/ssh/ssh_host_ecdsa_key
 HostKey /etc/ssh/ssh_host_ed25519_key
 
 # Enforce privilege separation by creating unprivileged child process
@@ -39,7 +40,6 @@ UseDNS no
 # Cipher selection
 
 Ciphers chacha20-poly1305@openssh.com,aes256-gcm@openssh.com,aes128-gcm@openssh.com,aes256-ctr,aes128-ctr
-HostKeyAlgorithms ssh-ed25519,rsa-sha2-256,rsa-sha2-512
 # Don't use SHA1 for kex
 KexAlgorithms curve25519-sha256@libssh.org,diffie-hellman-group-exchange-sha256
 # Don't use SHA1 for hashing, don't use encrypt-and-MAC mode
--- a/install_files/ansible-base/roles/restrict-direct-access/templates/sshd_config
+++ b/install_files/ansible-base/roles/restrict-direct-access/templates/sshd_config
@@ -2,18 +2,29 @@ Port 22
 ListenAddress {{ ssh_listening_address }}:22
 Protocol 2
 HostKey /etc/ssh/ssh_host_rsa_key
-HostKey /etc/ssh/ssh_host_dsa_key
-HostKey /etc/ssh/ssh_host_ecdsa_key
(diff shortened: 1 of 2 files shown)


zenmonkeykstop:
One minor note, would it be worth making an accompanying docs change specifically in https://docs.securedrop.org/en/stable/servers.html?highlight=ssh-keygen to use ed25519 for new installs??

zenmonkeykstop:
Tested on prod VMs  (using existing instructions to generate an RSA SSH key on the workstation):
- [x] installation completes successfully
- [ ] on subsequent playbook runs, or on attempts to connect via `ssh <serverAlias>` SSH fails with a host key mismatch, as an ECDSA host key was set up initially but an ED25519 key is now present.

Removing the `known_hosts` entries for app and mon and accepting the new host keys clears the error but it might be good to handle it more transparently.

install_files/ansible-base/roles/restrict-direct-access/templates/sshd_config
```diff
@@ -22,20 +33,34 @@ ChallengeResponseAuthentication no
...
-PrintLastLog yes
-TCPKeepAlive yes
-AcceptEnv LANG LC_*
 UsePAM no
 UseDNS no
+
+# Cipher selection
+
+Ciphers chacha20-poly1305@openssh.com,aes256-gcm@openssh.com,aes128-gcm@openssh.com,aes256-ctr,aes128-ctr
+HostKeyAlgorithms ssh-ed25519,rsa-sha2-256,rsa-sha2-512
```
zenmonkeykstop: It looks like this ordering forces a new host key to be generated when the new config is applied.  Since ECDSA is being removed altogether I don't see that there's a way around this in Xenial. It might make sense to have `tailsconfig` remove the old host key entries on the workstation if found, and alert the user that they can expect to see prompts on their first SSH/playbook run.

zenmonkeykstop (dismissed):

emkll:
> Since ECDSA is being removed altogether I don't see that there's a way around this in Xenial. It might make sense to have tailsconfig remove the old host key entries on the workstation if found, and alert the user that they can expect to see prompts on their first SSH/playbook run.

This might also be an issue in Focal, in a default configuration, the Host key is still ECDSA.


We could preserve the same host key preferences as the previous sshd_config: Given by default ssh is exposed Tor, which provides a second layer of authentication via authenticated onion services, or only exposed through a local network,  we could simply preserve ECDSA / distribution defaults for host keys.

I will look into  @zenmonkeykstop 's suggestion above, and use Ansible to
1. Delete the server fingerprint in known_hosts, and have a user re-verify the fingerprints or
2. Have Ansible edit known_hosts and add server fingerprints there

This might need to be done before the reboot action, so in the main Ansible playbook

conorsch:
> We could preserve the same host key preferences as the previous sshd_config

Preserving as-is seems the most straightforward. Certainly for Xenial, there's no pressing need to change the config. For Focal, ideally we'd have a stronger default, but the gains from v2 -> v3 onions are pretty great already, and v3 is required for Focal. So I _don't_ think we need to block on managing the host keys, as long as you're comfortable backing out on the defaults for now, @emkll. 

It does strike me as worthwhile to consider hostkey management outside the context of config updates: on first install, the same hostkey for the IPv4 address must be re-confirmed for the ssh onions, which we needn't prompt admins to configure. Still, I don't think that work needs to go into this PR. We can file a separate issue for it, and address if it proves to be a sticking point during testing the backup-and-restore flow as part of the LTS upgrade path.

emkll:
This is ready for re-review, however the docs update identified in https://github.com/freedomofpress/securedrop/pull/5666#issuecomment-744487280 is still pending

install_files/ansible-base/roles/restrict-direct-access/templates/sshd_config
```diff
@@ -22,20 +34,33 @@ ChallengeResponseAuthentication no
...
-PrintMotd no
-PrintLastLog yes
-TCPKeepAlive yes
-AcceptEnv LANG LC_*
 UsePAM no
 UseDNS no
+
+# Cipher selection
+
+Ciphers chacha20-poly1305@openssh.com,aes256-gcm@openssh.com,aes128-gcm@openssh.com,aes256-ctr,aes128-ctr
```
kushaldas: Do we really need `chacha20-poly1305@openssh.com`? 
`chacha20` is specially being used from the cheap phones which does not have good hardware. And our users will use Tails on proper hardware to ssh.

kushaldas (commented):
The current changes look nice to me, did not test it in a server yet.

kushaldas (approved):
- [x] provided config is good / adheres to best practices
- [x] provided config is adequately annotated
- [x] provided config eliminates log message described in #5660 
- [x] on an existing Xenial install the config is applied properly when an Ansible run occurs 


I tested manually with ssh over Tor on both Focal and Xenial. Works as expected. Approving. ~ Will merge after standup.~ @conorsch  do you think is it ready for merge?

conorsch (approved):

PR #5670, opened 2020-12-15
Removes deprecated 322 test from bandit

kushaldas:
## Status

Ready for review 

## Description of Changes

Removes old https://bandit.readthedocs.io/en/latest/blacklists/blacklist_calls.html?highlight=B322#b322-input check from bandit.

## Testing

- [ ] CI should be green.

## Deployment

Any special considerations for deployment? Consider both:

1. Upgrading existing production instances.
2. New installs.

### If you made changes to the server application code:

- [ ] Linting (`make lint`) and tests (`make test`) pass in the development container

### If you made changes to `securedrop-admin`:

- [ ] Linting and tests (`make -C admin test`) pass in the admin development container

### If you made changes to the system configuration:

- [ ] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

### If you made non-trivial code changes:

- [ ] I have written a test plan and validated it for this PR

Choose one of the following:

- [ ] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-docs) for these changes, or will do so later
- [ ] I would appreciate help with the documentation
- [ ] These changes do not require documentation

### If you added or updated a code dependency:

Choose one of the following:

- [ ] I have performed a diff review and pasted the contents to [the packaging wiki](https://github.com/freedomofpress/securedrop-debian-packaging/wiki)
- [ ] I would like someone else to do the diff review

PR #5671, opened 2020-12-15
Update bandit to 1.7.0

rmol:
## Status

Ready for review

## Description of Changes

The 322 test was not yet deprecated in our last pinned version (1.4.0) so removing it from the skip list in #5670 broke in local environments in which bandit wasn't upgraded (as is done each run in CI).

Also, 1.4.0 didn't yet support glob patterns in bandit's `--exclude` option, so running it locally would take forever as the `.venv` directory was scanned.

### Verify current problems

- `rm -rf .venv && make venv && . .venv/bin/activate`
- `make bandit`: It should take forever and complain about test 322.

### Verify fix
- check out this branch with `git checkout -b upgrade-bandit origin/upgrade-bandit`
- create a virtualenv with new bandit: `rm -rf .venv && make venv && . .venv/bin/activate`
- run `make bandit`: It should only check 76 files, finish quickly, and not complain about test 322.

## Deployment

dev only.

Choose one of the following:


zenmonkeykstop (approved):
Test plan passed and only deps change is in develop-requirements.txt, so this is cool to go.

PR #5669, opened 2020-12-15
Prod vms on focal

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Fixes #5668 .

Adds functionality to do a Focal production install using local deb packages and the `focal` dist on `apt-test.freedom.press`.

## Testing

- on dev box:
    - check out this branch
    - add the `bento/ubuntu-20.04` Vagrant box if not already present
    - set up prod focal base vms with `USE_FOCAL=true vagrant up --no-provision /prod/`
    - build focal debs with `make build-debs-focal`
    - set up local apt repo with `vagrant up apt-local`
- in admin workstation:
    - proceed with prod provisioning as normal as far as `./securedrop-admin sdconfig`
    - follow instructions in `install_files/ansible-base/securedrop-apt-local.yml` to add local repo on VMs
    - update `install-fpf-repo` defaults to point to apt-test+test key
    - do `./securedrop-admin install`

- [ ] installation completes successfully on Focal prod VMs
- [ ] `apt-cache policy securedrop-app-code` lists the local apt server as the source
- [ ] Source and Journalist Interface are accessible and version matches local deb version.

## Deployment

Any special considerations for deployment? Consider both:

1. Upgrading existing production instances.
2. New installs.

### If you made changes to the server application code:

- [ ] Linting (`make lint`) and tests (`make test`) pass in the development container

### If you made changes to `securedrop-admin`:

- [ ] Linting and tests (`make -C admin test`) pass in the admin development container

### If you made changes to the system configuration:

- [ ] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

### If you made non-trivial code changes:

- [ ] I have written a test plan and validated it for this PR

Choose one of the following:

- [ ] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-docs) for these changes, or will do so later
- [ ] I would appreciate help with the documentation
- [ ] These changes do not require documentation

### If you added or updated a code dependency:

Choose one of the following:

- [ ] I have performed a diff review and pasted the contents to [the packaging wiki](https://github.com/freedomofpress/securedrop-debian-packaging/wiki)
- [ ] I would like someone else to do the diff review

PR #5677, opened 2020-12-17
Disable v2 onion addresses on restore on Focal

### LINKED ISSUE #5676: Disable v2 onion addresses while restoring from backup on Focal
## Description

If the admin restores a backup which has both v2 and v3 onion addresses on to a Focal production instance, only the v3 addresses should be restored. 

This will require 3 things:

- Remove any v2 onion address configuration from `/etc/tor/torrc`
- Remove /var/lib/tor/services/source
- Remove /var/lib/tor/services/journalist

kushaldas:
## Status

Ready for review 

## Description of Changes

Fixes #5676 , towards https://github.com/freedomofpress/securedrop/issues/5731

We filter out any v2 onion address related line from /etc/tor/torrc
and also the directories from /var/lib/tor/services. This will
happen only on Focal. On Xenial, everything stays the same.

## Testing

- [ ] Checkout this branch in Tails
- [ ] Restore a v2 + v3 backup into a Xenial prod, nothing should change
- [ ] Create a Focal prod vm following steps at https://github.com/freedomofpress/securedrop/pull/5669 with both v2+v3
- [ ] Restore the same backup from tails to the Focal VM.

## Deployment

2. New installs.

Choose one of the following:

Choose one of the following:

### DIFF
--- a/install_files/ansible-base/roles/restore/files/compare_torrc.py
+++ b/install_files/ansible-base/roles/restore/files/compare_torrc.py
@@ -49,6 +49,10 @@ if __name__ == "__main__":
         print("The Tor configuration in the backup matches the server.")
         sys.exit(0)
 
+    if (3 in server_versions) and (3 in backup_versions):
+        print("V3 services detected in backup and server - proceeding with v3-only restore")
+        sys.exit(0)
+
     print(
         "The Tor configuration on the app server offers version {} services.".format(
             strset(server_versions)
--- a/install_files/ansible-base/roles/restore/files/disable_v2.py
+++ b/install_files/ansible-base/roles/restore/files/disable_v2.py
@@ -73,7 +73,6 @@ def filter_v2(filename):
                         i += 1
                         continue
 
-
         result.append(line)
         i += 1
 
--- a/install_files/ansible-base/roles/restore/files/disable_v2.py
+++ b/install_files/ansible-base/roles/restore/files/disable_v2.py
@@ -1,0 +1,90 @@
+#!/usr/bin/env python3
+# To execute on prod:
+# python3 disable_v2.py /etc/tor/torrc /etc/tor/torrc
+# To execute for testing locally:
+# python3 disable_v2.py /etc/tor/torrc /tmp/dumytorrc
+import sys
+
+
+def filter_v2(filename):
+    # Read the file
+    with open(filename) as f:
+        data = f.readlines()
+    # We will store the filtered lines to result
+    result = []
+
+    i = 0
+    while i < len(data):
+        line = data[i]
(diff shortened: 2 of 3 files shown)


eloquence:
(Marked as release blocker for 1.8.0.)

kushaldas:
Rebased today.

rmol (dismissed):
This left Xenial services untouched. On Focal it removed all v2 services but ssh. Seems like that should go as well.

kushaldas:
I started working on the updating of the PR, and now wondering about the scenario of `v2 ssh`. My update will assume that `v3 ssh` is already setup.

kushaldas:
Rebased against develop.

zenmonkeykstop:
Tried using this to restore a v2+v3 Xenial backup onto a v3-only Focal system. It fails at the compare_torrc.py step, as the configurations are obviously different, but this scenario should probably be supported.

My expectation would be that the migration states should go as follows:

|  backup \ target     | v2 Xenial |         v2+v3 Xenial          |       v3 Xenial or Focal        | 
|-------|-----------|-------------------------------|------------------------|
| v2    | Yes       | No (restore *then* migrate) | yes with --skip-tor    |
| v2+v3 | No        | Yes                           | yes but v3-only |
| v3    | No        | Yes but v3-only        | Yes                    | 

(implicit above is that there are no valid Focal targets with v2)

- The v2 backup row is covered with current code.
- The v2+v3 backup row is *not* covered, needs changes to compare_torrc.py logic and logic to apply only the v3 config.
- the v3 backup row is *not* covered, needs same changes as v2+v3

In all cases the docs need refreshing (see https://github.com/freedomofpress/securedrop-docs/pull/133 ).

## Implementation notes
- We can probably get away without the script to clean up the servers' `torrc` files, as the install playbook is run as part of migrations and it should ensure that `torrc` is correct (but does it? goood question).
- There is existing logic to skip Tor configs altogether using the `exclude` field: https://github.com/freedomofpress/securedrop/blob/1066cd00791ec612eae469f574e04464bf4cb53a/install_files/ansible-base/roles/restore/tasks/main.yml#L62 - a similar approach could be used to selectively exclude v2 configurations This could be cleaner than removing v2 files after copying them to the server.
- Changes to `compare_torrc.py` (or a different approach) would be required for the case where v2+v3 backups are applied to a v3 target - instead of erroring out it could indicate that v3 was available on the server, and allow the restore to proceed with the v3 info from the backup.


```

kushaldas:
I tested with v2/v3 and only v3 restore. Also fixed the merge conflict in CI. Now, need at least more person to make sure that I did not loose anything after merge conflict. Maybe @zenmonkeykstop

zenmonkeykstop (approved):
Conflict resolution looks good based on visual review. Approving.

PR #5683, opened 2021-01-05
Faster unit tests by re-using GPG directory

### LINKED ISSUE #4331: Initialization of fixures requiring a GPG homedir are slow
## Description

On my laptop, the import of the private/public keys into GPG takes ~2 seconds. This is done during the creation of the `config` fixture (see below).

https://github.com/freedomofpress/securedrop/blob/49400d004c404834895b33e22d257c2da1a314b5/securedrop/tests/conftest.py#L79-L102

This fixture is run nearly 100 times during the test suite. This is a lot of time.

It should be possible to import the keys one into a fixture scoped to `session`, make the `config` fixture depend on it, then copy it into place with `shutil.copytree`.

An initial attempt this was done in #4312, but it was unstable in seemed to fail CI despite consistently passing locally.

nabla-c0d3:
## Status

 Ready.

## Description of Changes

This PR fixes https://github.com/freedomofpress/securedrop/issues/4331 by updating the `config` fixture to have it re-use the same GPG directory across all tests via a `session` scope.

More specifically, the code that imports the test GPG key within the `config` fixture takes 2s to run on my laptop. This adds a 2s setup call to every single unit test that relies on the `config` fixture. A lot of tests do use this fixture (directly or indirectly). For example, the first parallel run of app-tests on CircleCI takes 22 minutes to complete for this PR, instead of 28 minutes. 

However, this PR is more about improving the developer experience when constantly running a few tests locally, while building a feature. Having a 2s delay on each test makes iterating on a feature much slower.

### DIFF
--- a/securedrop/tests/conftest.py
+++ b/securedrop/tests/conftest.py
@@ -1,13 +1,13 @@
 # -*- coding: utf-8 -*-
 
 import configparser
-from typing import Any
-from typing import Iterator
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from typing import Any, Generator
 
 import pretty_bad_protocol as gnupg
 import logging
 
-import py
 from flask import Flask
 from hypothesis import settings
 import os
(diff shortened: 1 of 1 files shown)


eloquence:
@nabla-c0d3 Rebasing should fix the safety check failure (resolved via #5686)

securedrop/tests/conftest.py
```diff
@@ -84,42 +85,53 @@ def setUpTearDown():
     _cleanup_test_securedrop_dataroot(original_config)
 
 
+@pytest.fixture(scope="session")
+def gpg_key_dir() -> Iterator[Path]:
```
kushaldas: Shouldn't this be https://docs.python.org/3/library/typing.html#typing.Generator ?

securedrop/tests/conftest.py
```diff
@@ -84,42 +85,53 @@ def setUpTearDown():
...
-    cnf.GPG_KEY_DIR = str(keys)
-    cnf.STORE_DIR = str(store)
-    cnf.TEMP_DIR = str(tmp)
-    cnf.DATABASE_FILE = str(sqlite)
-
-    # create the db file
-    subprocess.check_call(['sqlite3', cnf.DATABASE_FILE, '.databases'])
-
-    return cnf
+def config(gpg_key_dir: Path) -> Iterator[SDConfig]:
```
kushaldas: Same question here. I guess we should use `typing.Generator` here.

kushaldas (changes requested):
Rest looks good. Thank you @nabla-c0d3 for this. It will surely speedup the tests.

securedrop/tests/conftest.py
nabla-c0d3: Correct - I just fixed all instances of this.

kushaldas (approved):
Looks good. Approved.

PR #5691, opened 2021-01-06
Creates Focal-specific kernel metapackage

### LINKED ISSUE #4134: Use paxctld to manage all PaX flags in Ubuntu Focal

### LINKED ISSUE #5690: Create Focal-specific securedrop-grsec metapackage
## Description

As part of Focal support (#4768), we must create a new `securedrop-grsec` metapackage specifically for Focal. Doing so will allow us to preserve the existing Xenial behavior until its EOL (2021-04), and add new functionality for Focal, specifically:

* Using `paxctld` rather than `paxctl` #4134 
* Pinning a specific kernel version explicitly (was #5507)

Note that we don't currently support installing the locally-built `securedrop-grsec` metapackage as part of the staging environment, which complicates development and testing somewhat. Might be worthwhile to track that change as part of the new build logic, to aid in review. Otherwise, we'll have to merge logic to build a new package, then upload it to apt-test, and circle back to add tests for its config.

conorsch:
## Status

Ready for review.

## Description of Changes

Fixes #5690. Fixes #4134.

* Creates Focal-specific `securedrop-grsec` metapackage, for pinning kernel and configuring paxctld
* Installs the locally built `securedrop-grsec` in staging environment (previously we only used the package from apt-test in staging)
* Updates tests accordingly
* Tweaks build logic so we can more easily write per-distro packages

## Testing

* Visual review
* CI is passing in all scenarios
* (Optional) Run `make build-debs-focal` and inspect the `securedrop-grsec` metapackage that's created

## Deployment

Focal-only.

Choose one of the following:

Choose one of the following:


conorsch:
Note that post-merge, we should upload the new focal `securedrop-grsec` package to apt-test in https://github.com/freedomofpress/securedrop-dev-packages-lfs . Since this PR starts using the locally built metapackage in staging, CI is passing with out, but it must be there to unblock testing of Focal on hardware.

kushaldas (approved):
Tested locally against Focal target. Also did visual review. Looks good.

PR #5692, opened 2021-01-06
Refactor scrypt hashing and source app session mgmt pt 1/3

nabla-c0d3:
## Status

Ready.

## Description of Changes

This PR started with the goal of refactoring **CryptoUtil.hash_codename()** to move it out of CryptoUtil (for #5599).

However, I ran into some problems as I was trying to do this, and saw some opportunities for larger improvements. The downside is that the changes are bigger, so I split them into 3 PRs so they're a bit easier to review. While I know big changes are annoying to review, merge and ultimately deal with, I think these ones are important and worth it.

Overall I felt that a crucial concept was "missing"/invisible in the code base: a user who is currently logged into the source application. 

So the main change of these PRs is the introduction of a **SourceUser** class, which ties it all together: how a user's session is managed (when they create an account, login, etc.), how the user's crypto works (filesystem_id, gpg_secret, etc.), and the user's corresponding "Source" record in the DB. 

My overall goal is to make all this logic self-documenting so it is clear to the reader how these users (and more generally the source app) work. Currently, this logic is spread throughout the code base and also repeated a lot, which makes it hard to understand and hard to change. The main example of this is how hash_codename() is used in a lot of code locations, but in each of them it is hard to understand what is getting hashed, and why.

While I split the changes into 3 PRs, each PR brings something useful independently of the other ones:

* 1/3: This PR: Implement SourceUser and create_source_user() to consolidate source creation (More details below)
* 2/3: https://github.com/freedomofpress/securedrop/pull/5694 Remove CryptoUtil.hash_codename(), scrypt Python dependency, and refactor session management in the source app.
* 3/3: https://github.com/freedomofpress/securedrop/pull/5695 Remove usage of flask.g in the source app and fix tests.

If everything gets merged, the following issues will be completed or will see progress: #5455, #5599, #5613. I also think the source app's code base will be significantly easier to approach, understand, and maintain (at least that's the main goal).

**About this PR which is part 1/3:**

* Implement SourceUser and create_source_user() to consolidate source creation. In the current code base there is some duplicate code (in the tests, the apps, etc.) for creating a new source. This PR replaces all of them with a call to create_source_user() so there's one consistent way of creating a source.
* Fixes https://github.com/freedomofpress/securedrop/issues/5455.

Issue #5613, opened 2020-11-03
Investigate replacement of `pyotp` with equivalent functionality in `cryptography`


rmol:
## Description

The `cryptography` package contains support for [HOTP/TOTP](https://cryptography.io/en/latest/hazmat/primitives/twofactor.html). We should see if it can do everything we use `pyotp` for. Those two packages are excellent and maintained, but our update protocol is laborious enough that we should eliminate any dependencies we can.

*Edited by @eloquence to reflect removal of `scrypt` dependency, 2021-09-28*

evilaliv3:
@rmol : Actually i come to this evaluation myself within @GlobaLeaks and i just made this change.

Here you can find the code i used to reimplement TOTP by means of python-cryptography: https://github.com/globaleaks/GlobaLeaks/commit/64750f40e7d8b49739cb184fd0e4f7b662c8d0ea

rmol:
Hey, thanks @evilaliv3! That should save us some time. :slightly_smiling_face:

nabla-c0d3:
The dependency to `scrypt` was removed in #5694.

eloquence:
Thanks, retitled accordingly!

nabla-c0d3:
PR at https://github.com/freedomofpress/securedrop/pull/6679 is ready

nabla-c0d3:
PR https://github.com/freedomofpress/securedrop/pull/6679 was merged so this issue can be closed 👍

PR #5694, opened 2021-01-11
Refactor scrypt hashing and source app session mgmt pt 2/3

nabla-c0d3:
## Status

Ready.

## Description of Changes

This PR follows #5692 .

* This PR adds logic to cache the source's information as their "session" in the newly-implemented **SessionManager**, so the user's filesystem_id and gpg_secret are only generated once, on login. In the current implementation, the filesystem_id is generated on every authenticated request hitting the source app, ie. an scrypt call (=slow) is done on each authenticated request.
* More generally, the SessionManager centralizes documents session management in the source app, and hence replaces the usage of the Flask's "thread locals" **flask.g** and **flask.session**. These globals are opaque (it's hard to know what data/keys are in them, and when), untyped, and overall a bit of an antipattern; even the Flask documentation mentions that they "make large applications harder to maintain" (https://flask.palletsprojects.com/en/1.1.x/design/#micro-with-dependencies).
* This PR removes the dependency to scrypt library and switches to cryptography's implementation (for #5613).
* The new centralized session-management logic should also fix the following issues: 
  * #5716
  * #5741


securedrop/crypto_util.py
```diff
@@ -84,9 +83,6 @@ class CryptoUtil:
     SOURCE_KEY_UID_RE = re.compile(r"(Source|Autogenerated) Key <[-A-Za-z0-9+/=_]+>")
 
     def __init__(self,
-                 scrypt_params: Dict[str, int],
-                 scrypt_id_pepper: str,
-                 scrypt_gpg_pepper: str,
```
nabla-c0d3, line 89: The scrypt hashing is now 100% done outside of CryptoUtil.

securedrop/source_app/session_manager.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class SessionManager:
+    _CACHE = SimpleCache()  # A mapping of passphrase -> SourceUser
+    _KEY_IN_SESSION_COOKIE = "codename"  # The key in flask.session for the user's passphrase
+
+    @classmethod
+    def log_user_in(cls, user: SourceUser, user_passphrase: "DicewarePassphrase") -> None:
+        # Save the passphrase in the user's session cookie
+        session[cls._KEY_IN_SESSION_COOKIE] = user_passphrase
```
nabla-c0d3, line 36: This follows the same logic as the current code base, which is to put the user's passphrase in their session cookie (which is encrypted). I don't know if this is useful or not, but we could replace the passphrase in the cookie with a randomly-generated session ID.

securedrop/source_app/__init__.py
```diff
@@ -139,57 +132,17 @@ def load_instance_config() -> None:
...
 
         session['expires'] = datetime.utcnow() + \
             timedelta(minutes=getattr(config,
                                       'SESSION_EXPIRATION_MINUTES',
                                       120))
 
-        # ignore_static here because `crypto_util.hash_codename` is scrypt
-        # (very time consuming), and we don't need to waste time running if
-        # we're just serving a static resource that won't need to access
-        # these common values.
```
nabla-c0d3, line 175: This is no longer needed since I removed the "time consuming" scrypt call from the code that handles incoming requests.

securedrop/source_app/__init__.py
```diff
@@ -139,57 +132,17 @@ def load_instance_config() -> None:
...
-                            .filter_by(deleted_at=None) \
-                            .one()
-            except NoResultFound as e:
-                app.logger.error(
-                    "Found no Sources when one was expected: %s" %
-                    (e,))
-                del session['logged_in']
-                del session['codename']
-                return redirect(url_for('main.index'))
-            g.loc = app.storage.path(g.filesystem_id)
```
nabla-c0d3, line 191: All the authentication and session logic in this file is now handled by the SessionManager, which is called when the `@login_required` decorator is used in the Flask routes. This ensures that all authentication / session management logic happens in a single code location, thereby reducing complexity.

securedrop/source_app/decorators.py
```diff
@@ -1,19 +1,28 @@
...
+        try:
+            logged_in_source = SessionManager.get_logged_in_user()
+
+        except (UserSessionExpired, UserHasBeenDeleted):
+            return clear_session_and_redirect_to_logged_out_page(session)
+
+        except UserNotLoggedIn:
+            return redirect(url_for("main.login"))
+
+        return f(*args, **kwargs, logged_in_source=logged_in_source)
```
nabla-c0d3, line 25: The decorator directly passes the source user to the Flask route/code (to make the "data flow" more obvious/simple), so using flask.g or session (and checking that they are in the right "state") is no longer needed.

securedrop/source_app/decorators.py
```diff
@@ -1,19 +1,28 @@
...
     @wraps(f)
     def decorated_function(*args: Any, **kwargs: Any) -> Any:
-        if not logged_in():
-            return redirect(url_for('main.login'))
-        return f(*args, **kwargs)
+        try:
+            logged_in_source = SessionManager.get_logged_in_user()
+
+        except (UserSessionExpired, UserHasBeenDeleted):
+            return clear_session_and_redirect_to_logged_out_page(session)
```
nabla-c0d3, line 20: This way of handling expiring sessions should fix the following issues: #5741, #5716

kushaldas (approved):
Running it on staging for a few days, no issues shown.

- [x] verified #5716 is fixed
- [x] verified #5741 is fixed
- [x] On **g** https://flask.palletsprojects.com/en/1.1.x/design/#thread-locals

Include tests,  and also updates the `apparmor` entries.  Merging this now and then moving to the last PR in this set.

nabla-c0d3:
Awesome!

kushaldas:
We may have to revert this. While reviewing #5695 I noticed that sometimes using a second or third source at the same time is causing **logout due to inactivity**.  And a a few times after updating to this session manager code, I can see that the source interface is not showing the *source* passphrase on the top.

For example:
Screenshot from 2021-09-09 13-00-44

nabla-c0d3:
I can look into this.

zenmonkeykstop:
Thanks @nabla-c0d3  - if you wanna chat about it (even just for rubber-duck mode)  folks are around on SD Gitter: https://gitter.im/freedomofpress/securedrop

nabla-c0d3:
> Thanks @nabla-c0d3 - if you wanna chat about it (even just for rubber-duck mode) folks are around on SD Gitter: https://gitter.im/freedomofpress/securedrop

Thanks! It was the morning for me so I only got to this just now. Some thoughts/questions:

> Using a second or third source at the same time is causing logout due to inactivity

* I tried on the dev server and wasn't able to reproduce this. I was logged in as 4 different sources, uploading documents and such, but never got logged out on any of them.
* This brings the following question: @kushaldas were you using the dev server or a production setup? If it was production, which web server is used to serve the source interface, and does the web server use multiple processes to process requests? If that's the case the cache at SessionManager._CACHE would not be shared across processes, so users would see the "logged out" message if their request hit a different process.
* I could revert the logic to not cache anything and only use the session cookie, but this means that every request will trigger one call to scrypt to derive the filesystem_id from the passphrase in the cookie, as it was previously. 

> source interface is not showing the source passphrase on the top

 * I tried on the dev server and the passphrase is displayed right after user creation. It is not displayed on subsequent logins, which, to me, is expected; this seemed to be the intent of the code when I refactored it. It would be easy tho to update the logic to always show the passphrase after logging in (instead of just after user creation).
 *  Or is it that right after you created a source user, the passphrase still wasn't displayed? This would be unexpected to me and a bug.

That's all I have for now... let me know your thoughts @kushaldas @zenmonkeykstop . Thanks!

zenmonkeykstop:
For the first issue, I managed to get the "logged out due to inactivity" error intermittently (about half the time) in a staging env by:

1) logging in as a new source (/generate -> /lookup)
2) logging out and then going back to the home page (without clicking the Tor Browser New Identity button)
3) clicking through to /generate and getting a new codename
4) clicking thru to /lookup - instead of getting the /lookup page I get redirected to home with the "logged out" error.

This isn't precisely what Kushal saw, though, and I haven't found a way to consistently replicate that yet. (Also as STRs go, it is as the kids say, kinda sus)

Production (and staging) instances use Apache and mod_wsgi, and 2 processes are being spawned, so it's possible different requests will be served by different child processes. It might be better to back that change out as you mentioned. We could look at alternatives for subsequent releases. (The application uses Redis to cache GPG fingerprints already, but I think more work/research would be required before using it for full-on session management.)

For the second issue (no passphrase on /lookup) - my understanding is that the  scenario Kushal is describing is the passphrase not showing up on /lookup on first login. I haven't managed to replicate this yet though, and it could just be a misunderstanding, so focusing on the first issue seems like the best idea for now.

nabla-c0d3:
I went through the steps you described a few times but I haven't been able to reproduce the logout errors/bugs. My environment might be different. Is your staging environment the same as running `make dev` ?

Regardless, I've updated the next PR at #5695 to remove the local cache, so that the whole session information is stored in the session cookie, like it was previously. This should get rid of these errors; I've tried it in my dev environment and it seemed to work fine.

zenmonkeykstop:
Hey @nabla-c0d3 - we have 3 types of environments, each with their own foibles:
- dev env (`make dev`): Docker-based, single-server environment, that just runs the two wsgi apps directly, exposing the services via open Docker ports. This is the simplest to run and most useful for web development, downside is that architecture-wise it's not a good match for production.

- staging env (`make staging`): libvirt-based, multiserver environment, this is closer to production in that it includes both separate app and mon servers, proxies the apps via Apache, and makes the services available over Tor. It's more hassle to set up and requires that you build deb packages to push changes, rather than live-editing code, but it does catch the kind of issues that will only appear once the app is running against the prod-like server config.

- prod VM env (`molecule create -s libvirt-prod-focal`): this is mostly used for upgrade testing (ie. 2.0.1 to 2.0.2). It is a virtualized production environment that requires a (usually also virtualized) Tails admin workstation to set up. It will catch any issues that you'll see in staging, but can also be used to test the effects of new packaging scripts on an existing system.

Thanks for the updates (and for the knock-on updates to pr number 3 :) )!  We'll take a pass through ASAP.

PR #5696, opened 2021-01-11
Use confirmation modal when Admin deletes a user

### LINKED ISSUE #4649: Deleting a user in the Journalist Interface should trigger a confirmation dialog, even without JavaScript
Currently, deleting a user using the "trashcan" icon in the Journalist Interface is done immediately, without a confirmation dialog. As standard practice, destructive operations without undo should have a confirmation step.

# User Stories

As an admin, I want to be prompted before performing destructive operations, so I don't perform them accidentally when my hamster jumps onto my head.

prateekj117:
## Status

Work in progress

## Description of Changes

Fixes #4649 

This PR removes the logic of only JS supported confirmation modal when Admin deletes a User and replaces it with the confirmation modal which works without JS. 

## Testing

Login as Admin and try deleting a user. Try out the confirmation modal. Should also work with Safest Setting for Tor Browser (i.e, without JS).

Choose one of the following:

### DIFF
--- a/securedrop/journalist_templates/_confirmation_modal.html
+++ b/securedrop/journalist_templates/_confirmation_modal.html
@@ -10,7 +10,7 @@
     <center>
       <div class="btn-row">
         <a href="#close" id="{{ modal_data.cancel_id }}" title="{{ gettext('Cancel') }}" class="btn cancel small">{{ gettext('Cancel') }}</a>
-        <button type="submit" id="{{ modal_data.submit_id }}" name="action" value="delete" class="btn small {{ modal_data.submit_btn_type }}">{{ modal_data.submit_btn_text }}</button>
+        <button type="submit" id="{{ modal_data.submit_id }}" name="action" value="delete" {% if modal_data.modal_action is defined %} formaction="{{ modal_data.modal_action }}" {% endif %} class="btn small {{ modal_data.submit_btn_type }}">{{ modal_data.submit_btn_text }}</button>
       </div>
     </center>
   </div>
--- a/securedrop/journalist_templates/admin.html
+++ b/securedrop/journalist_templates/admin.html
@@ -38,12 +38,14 @@
         </td>
 	{% else %}
         <td>
-          <button type="submit" class="plain delete-user" formaction="/admin/delete/{{ user.id }}" data-username="{{ user.username}}">
-            <i title="{{ gettext('Delete user {username}').format(username=user.username) }}">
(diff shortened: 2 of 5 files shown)


prateekj117:
@eloquence @kushaldas @rmol For passing the specific `user_id` to the confirmation modal, I have used URL Params and added an optional attribute of `formaction` in `_confirmation_modal.html`. I am not sure if it is the safest thing or not to pass `user_id` like this. Though proper checks are implemented in the backend endpoint logic, so I went ahead with this approach. If you think there is a better way to handle this, do let me know and I will make the changes.
If you agree with this approach, let me know and I will add proper tests for the PR as well.

prateekj117:
@eloquence I actually needed a small review on this.

eloquence:
@prateekj117 My apologies, the team is currently very busy with the Ubuntu 20.04 migration, but we'll aim to have a look soon.

rmol:
@prateekj117 I think the way you're passing the user ID is fine. Production traffic is always encrypted. We're already including the user ID in the URL of the next step, the [deletion route](https://github.com/freedomofpress/securedrop/blob/919934d54ce5ba5862a577270e4ad6e897a6a64a/securedrop/journalist_app/admin.py#L266), e.g. `/admin/delete/3`. 

We could consider using the journalist UUID instead, both here and for the deletion route, but I think the utility of that change would be minimal and it's definitely outside the scope of this PR.

We could alter the list of users to work like the source list on the JI index page, with a checkbox on each row and `Edit` and `Delete` buttons at the top, so that the confirmation modals here could work the same as that page, submitting IDs in form parameters, but again, I don't see the need, certainly not in the context of this PR.

I'm curious if @emkll or @zenmonkeykstop have any thoughts, having worked on the recent source deletion changes.

This PR does need a rebase onto latest `develop`.

prateekj117:
@rmol I will rebase it with the latest `develop` and also add tests for the same. Was just waiting for the review. If this approach looks good to you, I will add the tests and finalize the PR.

prateekj117:
@rmol Corrected the tests and rebased the PR.

eloquence:
@prateekj117 Could you rebase once more to pick up the recent Xenial removal? This should get app tests passing again.

prateekj117:
@eloquence Ready to Review.

zenmonkeykstop (approved):
Test plan passes, changes look good. Thank you for sticking with this one during some long delays!

PR #5707, opened 2021-01-14
Make singular strings singular

eloquence:
(Thanks to AO for pointing this out via Weblate comments)

## Status

Ready for review

### DIFF
--- a/securedrop/journalist_app/forms.py
+++ b/securedrop/journalist_app/forms.py
@@ -28,7 +28,7 @@ def minimum_length_validation(form: FlaskForm, field: Field) -> None:
     if len(field.data) < Journalist.MIN_USERNAME_LEN:
         raise ValidationError(
             ngettext(
-                'Must be at least {num} characters long.',
+                'Must be at least {num} character long.',
                 'Must be at least {num} characters long.',
                 Journalist.MIN_USERNAME_LEN
             ).format(num=Journalist.MIN_USERNAME_LEN)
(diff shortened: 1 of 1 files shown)


securedrop/journalist_app/forms.py
```diff
@@ -28,7 +28,7 @@ def minimum_length_validation(form: FlaskForm, field: Field) -> None:
     if len(field.data) < Journalist.MIN_USERNAME_LEN:
         raise ValidationError(
             ngettext(
-                'Must be at least {num} characters long.',
+                'Must be at least {num} character long.',
```
eloquence, line 31: Could also get rid of `{num}` here and say `1` or `one`, not sure if any option is better from the standpoint of translators

rmol: I don't know. The only benefit I saw of replacing the placeholder was that `one` meets AP style rules, but we still fall afoul of them at `2`. :wink: 

Either option right now can cause a Weblate check complaint, so no benefit there. I would just go with what you have.

conorsch:
For context, see https://weblate.securedrop.org/translate/securedrop/securedrop/en/?checksum=18c0a71ef2d0d92b#comments

conorsch (approved):
Approving based on visual review. Thanks for prompt turnaround, both!

conorsch:
Fine to merge once tests pass. @creviera I suggest we proactively open a backport PR into release/1.7 so we can run CI concurrently. Agree?

PR #5710, opened 2021-01-15
SecureDrop 1.7.0-rc2

sssoleileraaa:
## Status

Ready for review

## Description of Changes

Towards https://github.com/freedomofpress/securedrop/issues/5689


emkll (approved):
thanks @creviera , lgtm. I've pushed a minor change: the file `molecule/shared/stable.ver` was modified in this branch, but should only on release.

creviera:
(noting that we'll want to mention in the release management docs that we'll need to disregard the change to stable.ver after running `update_version.sh`)

PR #5718, opened 2021-01-18
Only configure v2 ssh service when v2 services are enabled

### LINKED ISSUE #5717: Ossec alert is still being triggered after v2 services were disabled due to v2 ssh hidden service in torrc
## Description

A short summary of the issue.

## Steps to Reproduce
- Install server with v2 to v3 (with ssh over tor set to true)
- Disable v2 onion service
- [ ] Observe torrc containing a definition for ssh using v2 services
- [ ] Observe ossec alerts stating you need to upgrade to v3 onion services

## Expected Behavior

After migrating to v3 onion services, admins should not receive alerts pertaining to v2-> v3 upgrade
## Actual Behavior
Admin receives an alert:
```

OSSEC HIDS Notification.
2021 Jan 18 19:59:21

Received From: (app) 10.0.1.4->v2_service_check
Rule: 400901 fired (level 12) -> "v2 onion services are still enabled. Support for v2 onion services is deprecated and will be removed starting in February 2021. To preserve access to SecureDrop, you must migrate to v3 onion services: https://securedrop.org/v2-onion-eol"
Portion of the log(s):

ossec: output: 'v2_service_check': HiddenServiceVersion 2



 --END OF NOTIFICATION

```

## Comments

It looks like this block on the jinja template for the torrc is misplaced, and should be moved in the "if" block above it.
https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/tor-hidden-services/templates/torrc#L19

emkll:
## Status

Ready for review 

## Description of Changes

Fixes #5717 

Updates the torrc template to conditionally add the ssh config for v2 onion services

## Testing

0. Make sure the underlying issue is reproducible
- [ ] I can reproduce the issue described in #5717 

1. Install from develop (or 1.7.0-rc2) using staging or prod scenarios, with the following config:
* enable_ssh_over_tor: true
* v2_onion_services: true
* v3_onion_services: true

2. Confirm that you receive the alert:
- [ ] An email alert is received indicating v2 services are enabled
- [ ] /var/ossec/logs/alerts/alert.log contains rule 4000901 triggered

3. Check out this branch, set `v2_onion_services: false` and run `./securedrop-admin install`
4. Confirm the changes work as expected
- [ ] `/var/ossec/logs/alerts.log` does not contain a new rule 4000901 that was triggered
- [ ] `/etc/tor/torrc` does not contain occurrences of `HiddenServiceVersion 2` on both app and mon
 
## Deployment

New and existing installs will be configured using Ansible. For orgs which have already updated to disable v2, they will need to run the playbook again to ensure v2 onion services for ssh are properly disabled.

Disabling v2 and enabling v3 in a single pass might prove problematic if provisioning over Tor, but the documentation appears sufficiently clear that one should enable v2+v3 first, then disable v2: https://docs.securedrop.org/en/stable/v3_services.html#disabling-v2-onion-services

### DIFF
--- a/install_files/ansible-base/roles/tor-hidden-services/templates/torrc
+++ b/install_files/ansible-base/roles/tor-hidden-services/templates/torrc
@@ -16,7 +16,7 @@ HiddenServicePort 80 127.0.0.1:8080
 HiddenServiceAuthorizeClient stealth journalist
 {% endif %}
 
-{% if enable_ssh_over_tor %}
+{% if enable_ssh_over_tor and v2_onion_services %}
 HiddenServiceDir /var/lib/tor/services/ssh
 HiddenServiceVersion 2
 HiddenServicePort 22 127.0.0.1:22


rmol (dismissed):
:+1: Starting from `develop`, `/etc/tor/torrc` contained v2 SSH. Reran install from this branch, and only v3 services were left in the file. Legacy SSH services were no longer reachable.

I did not get the OSSEC alerts from the `develop` installation, but that's probably because the check is daily and I didn't reduce the interval.

This is good to merge once rebased on current develop to get CI green.

rmol (approved):

PR #5759, opened 2021-01-27
Restore conservative configuration

### LINKED ISSUE #5757: Error 500 on source interface and journalist interface after 1.7.0 upgrade
## Description

Initially reported by forum user jonas.franzen in https://forum.securedrop.org/t/issues-after-update-to-1-7-0/1391

Some due to a missing variable in config.py
```
Jan 27 12:48:07 applicationserver python[1136]: AttributeError: module ‘config’ has no attribute ‘SESSION_EXPIRATION_MINUTES’
```

## Steps to reproduce

TBD

## Comments
Forum user dachary reports that simply adding the value to config.py resolves

rmol:
## Status

Ready for review

## Description of Changes

Fixes #5757.

Bring back the try/except blocks needed to support long-running instance configurations that may lack newer attributes.

## Testing

- `make test-config`
- comment out various settings, e.g. `SESSION_EXPIRATION_MINUTES`
- `make dev`

The dev server should start without errors.

## Deployment

This should be restoring the conservative configuration handling used in 1.6.0 and earlier.

### DIFF
--- a/securedrop/tests/test_config.py
+++ b/securedrop/tests/test_config.py
@@ -1,0 +1,47 @@
+import importlib
+
+import config as _config
+
+
+def test_missing_config_attribute_is_handled():
+    """
+    Test handling of incomplete configurations.
+
+    Long-running SecureDrop instances might not have ever updated
+    config.py, so could be missing newer settings. This tests that
+    sdconfig.SDConfig can be initialized without error with such a
+    configuration.
+    """
+    attributes_to_test = (
+        "JournalistInterfaceFlaskConfig",
+        "SourceInterfaceFlaskConfig",
+        "DATABASE_ENGINE",
+        "DATABASE_FILE",
+        "ADJECTIVES",
+        "NOUNS",
+        "GPG_KEY_DIR",
+        "JOURNALIST_KEY",
+        "JOURNALIST_TEMPLATES_DIR",
+        "SCRYPT_GPG_PEPPER",
+        "SCRYPT_ID_PEPPER",
+        "SCRYPT_PARAMS",
+        "SECUREDROP_DATA_ROOT",
+        "SECUREDROP_ROOT",
+        "SESSION_EXPIRATION_MINUTES",
+        "SOURCE_TEMPLATES_DIR",
+        "TEMP_DIR",
+        "STORE_DIR",
+        "WORKER_PIDFILE",
+    )
+
+    try:
+        importlib.reload(_config)
+
+        for a in attributes_to_test:
+            delattr(_config, a)
+
+        from sdconfig import SDConfig
+
+        SDConfig()
+    finally:
+        importlib.reload(_config)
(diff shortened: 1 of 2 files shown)


zenmonkeykstop (changes requested):
Test plan passes for `SESSION_EXPIRATION_MINUTES` and locale config settings, fails with numerous errors for other config settings. For some of them it *should* fail (like `SCRYPT_*`, there should never be default values for those.) Testing scope might need to be narrowed.

securedrop/tests/test_config.py
```diff
@@ -0,0 +1,44 @@
...
+    )
+
+    importlib.reload(_config)
+
+    for a in attributes_to_test:
+        delattr(_config, a)
+
+    from sdconfig import SDConfig
+
+    SDConfig()
```
zenmonkeykstop: Looks like this breaks subsequent tests that rely on the removed attributes.

zenmonkeykstop (dismissed):

rmol:
I should have written that the dev server should start without errors thrown from `sdconfig.py`, and that any errors seen are raised where the configuration values are used.

securedrop/tests/test_config.py
rmol: yeah, added another reload at the end, which resolved it with a subset locally.

emkll (approved):
Tested the changes in a staging server, from a 1.6.0 instance (with `SESSION_EXPIRATION_MINUTES` removed from `config.py`. Upon upgrading to 1.7.0, observed the error at upgrade time (and 500 on source interface):

```
  File "<frozen importlib._bootstrap_external>", line 665, in exec_module
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "/var/www/securedrop/alembic/versions/3da3fcab826a_delete_orphaned_submissions.py", line 18, in <module>
    from journalist_app import create_app
  File "/var/www/securedrop/journalist_app/__init__.py", line 13, in <module>
    import i18n
  File "/var/www/securedrop/i18n.py", line 32, in <module>
    from sdconfig import SDConfig
  File "/var/www/securedrop/sdconfig.py", line 98, in <module>
    config = SDConfig()  # type: SDConfig
  File "/var/www/securedrop/sdconfig.py", line 42, in __init__
    self.SESSION_EXPIRATION_MINUTES = _config.SESSION_EXPIRATION_MINUTES  # type: int
AttributeError: module 'config' has no attribute 'SESSION_EXPIRATION_MINUTES'
 subprocess installed post-installation script returned error exit status 1
 securedrop-app-code
Cannot start "/usr/sbin/sendmail": executable not found (adjust *sendmail* variable)
"/root/dead.letter" 1197/65704
... message not sent
root@app-staging:/var/www/securedrop# apt list --installed | grep securedrop-app-code
```

After running cron-apt on a local apt server with the changes in this branch, I can confirm this resolves the issue, and the source  no longer returns 500 and accepts submissions.

Deferring to @zenmonkeykstop for final review, but good to merge for 1.7.1, from my perspective, once CI passes

Issue #5663, opened 2020-12-01
Attempt hardware SecureDrop install on Focal


emkll:
As part of the [Focal epic](https://github.com/freedomofpress/securedrop/issues/4768) , we have now all (application and infra) tests passing in CI in Focal https://github.com/freedomofpress/securedrop/pull/5638 . 

CI and staging environments are virtualized, and as such there can be differences in kernel support, but also with packages installed. This will also help us begin to document any differences in documentation during the install steps of the underlying OS.

This ticket is to track the install of SecureDrop on a hardware server running Ubuntu 20.04 focal. We should

- [x] `make build-debs focal`
- [x] deploy these packages to apt-test.freedom.press https://github.com/freedomofpress/securedrop-dev-packages-lfs/pull/77
- [ ] Provision `app` and `mon` servers on the latest Ubuntu 20.04 ISO (20.04.1 as of this writing)
- [ ] Install SecureDrop an install (using apt-test.freedom.press on the `focal` channel)
- [ ] Document any findings either here or follow up any issues.

eloquence:
(Blocked on #5638)

eloquence:
#5638 is merged; shoring up the prod VM story is another step towards this issue, see #5669.

eloquence:
Now that we have packages on apt-test, this should be unblocked. @rmol will do a first install attempt on NUC7s during the 1/6-1/20 sprint and document findings here.

eloquence:
This is still a very high priority but slipped due to competing QA requirements for 1.7.0. @rmol and @emkll will lead this investigation during the 1/21-2/3 sprint.

eloquence:
Next steps discussed in sprint planning today:
- @emkll (Mac Mini) and @rmol (NUC7) will document their findings from their initial install attempts.
- [x] @creviera installed Focal on NUC8 and NUC10
- @kushaldas has been provisionally signed up for a NUC5 install.
- @zenmonkeykstop will attempt an install on the 1U servers.

Relatedly, in support of https://github.com/freedomofpress/securedrop-docs/issues/135, @rmol and @zenmonkeykstop will start drafting an outline (can be in the wiki for now) of what the install process looks like (especially the Ubuntu 20.04 install, which is substantially different from our current docs), informed by these initial install runs.

emkll:
Preliminary findings, based on Mac Mini testing (testing in progress, will edit this comment):
[...]

PR #5771, opened 2021-02-03
Remove the cloud-init package during server installation

rmol:
## Status

Ready for review

## Description of Changes

Towards https://github.com/freedomofpress/securedrop/issues/5663

We don't want Ubuntu's `cloud-init` package on the servers, not least because it can make them take ten minutes or more to boot, apparently while waiting for `systemd-random-seed` to initialize.

## Testing

1. Prepare two fresh Ubuntu Focal servers.
2. On your admin workstation, check out this branch and run the installation:
    - `git fetch --all`
    - `git checkout -b remove-cloud-init origin/remove-cloud-init`
    - `./securedrop-admin install`
3. Observe that the installation does not fail because of timeouts waiting ten minutes for the servers to reboot.
4. Verify that `cloud-init` is no longer installed on either server.
5. Run the configuration tests:
    - `./securedrop-admin tailsconfig`
    - `./securedrop-admin verify`
    - The `common/test_system_hardening.py::test_unused_packages_are_removed` test should pass.

## Deployment

We don't have `cloud-init` on existing Xenial instances, and this change means we also won't on new Focal systems.

Choose one of the following:

### DIFF
--- a/install_files/ansible-base/roles/prepare-servers/tasks/main.yml
+++ b/install_files/ansible-base/roles/prepare-servers/tasks/main.yml
@@ -21,6 +21,14 @@
       SecureDrop cannot be installed. For details, see
       https://github.com/freedomofpress/securedrop/issues/4058
 
+- name: Remove cloud-init
+  apt:
+    name: cloud-init
+    state: absent
+    purge: yes
+  tags:
+    - apt
+
 - name: Install python and packages required by installer
   raw: apt install -y python apt-transport-https dnsutils
   register: _apt_install_prereqs_results
--- a/molecule/testinfra/common/test_system_hardening.py
+++ b/molecule/testinfra/common/test_system_hardening.py
@@ -141,6 +141,7 @@ def test_no_ecrypt_messages_in_logs(host, logfile):
 
 
 @pytest.mark.parametrize('package', [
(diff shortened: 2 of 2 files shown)


rmol:
More findings:

- I think it's the [Seed Random module of cloud-init](https://cloudinit.readthedocs.io/en/latest/topics/modules.html#seed-random) that's causing the hang.
- It is also cured by our 5.4.88-grsec kernel, presumably because of the [new boot-time entropy collection in 5.4](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/drivers/char/random.c?h=v5.4.88#n1758). 

Just trying to document for posterity, since we don't want `cloud-init` at any rate, but should we ever need it, it should be fine as long as we're on a Focal-era kernel.

emkll (approved):
Thanks @rmol this addressed the issue for my NUC, but not on Mac Mini. As you identify in https://github.com/freedomofpress/securedrop/pull/5771#issuecomment-772846485, running a 5.4 series kernel fully resolves the issue on all hardware I've tested against. Regardless, cloud-init is something we want to remove from Focal SecureDrop instances.

I do have other testinfra tests failing on a Focal install, but unrelated to this PR.

PR #5777, opened 2021-02-08
Postfix fix for focal

### LINKED ISSUE #5775: [Focal] Emails not being sent from mon server under Focal
## Description
It appears that the monitor server's postfix configuration isincompatible with Ubuntu 20.04 Focal 

Part of #4768 

## Steps to Reproduce

1. Install SecureDrop on a Focal host
2. Observe error in syslog on `mon`:

```
mon postfix/smtpd[3103]: fatal: in parameter smtpd_relay_restrictions or smtpd_recipient_restrictions, specify at least one working instance of: reject_unauth_destination, defer_unauth_destination, reject, defer, defer_if_permit or check_relay_domains
```
## Expected Behavior

Emails should be set from the mon server containing ossec alerts

## Actual Behavior

Emails are not being sent due to incompatibility in postfix configuration.

kushaldas:
## Status

Ready for review

## Description of Changes

Fixes #5775 

Adds postfix configuration for `Focal` based on http://www.postfix.org/postconf.5.html#smtpd_recipient_restrictions

## Testing

- [ ] Create Focal prod VMs 
- [ ] ssh into both and do `sudo apt install ubuntu-release-upgrader-core` because the latest Bento box does not have that.
- [ ] Apply the following the patch on your tails vm, and then do `./securedrop-admin install` to install SecureDrop on `Focal`.

```diff
diff --git a/install_files/ansible-base/roles/install-fpf-repo/defaults/main.yml b/install_files/ansible-base/roles/install-fpf-repo/defaults/main.yml
index d186993df..6354113fa 100644
--- a/install_files/ansible-base/roles/install-fpf-repo/defaults/main.yml
+++ b/install_files/ansible-base/roles/install-fpf-repo/defaults/main.yml
@@ -10,7 +10,7 @@
 #
 # For testing/QA, set this URL to another apt server. You must also update
 # the associated public key for the apt repo for testing/QA.
-apt_repo_url: https://apt.freedom.press
+apt_repo_url: https://apt-test.freedom.press
 
 # By default, install packages from the apt-repo, but under
 # staging hosts we'll prefer locally-built deb packages
@@ -20,7 +20,7 @@ install_local_packages: False
 # May be overridden in staging to install from a test/QA server,
 # the Release file for which will *not* be signed with the prod key.
 apt_repo_pubkey_files:
-  - fpf-signing-key.pub
+  - apt-test-signing-key.pub
 
 # Enabling support for xenial by default.
 apt_repo_target_distro: "{{ ansible_distribution_release }}"

```

- [ ] login to the mon server and check for any postfix error in the `/var/log/syslog` file.

## Deployment

2. New installs.

Choose one of the following:

Choose one of the following:

### DIFF
--- a/install_files/ansible-base/roles/postfix/templates/main.cf
+++ b/install_files/ansible-base/roles/postfix/templates/main.cf
@@ -62,3 +62,7 @@ maximal_queue_lifetime = 14d
 # Used to remap outbound from address in emails
 smtp_generic_maps = hash:/etc/postfix/generic
 {% endif %}
+{% if ansible_distribution_release == 'focal' %}
+# Restriction for client RCPT TO command, details at http://www.postfix.org/postconf.5.html#smtpd_recipient_restrictions
+smtpd_recipient_restrictions = reject_unauth_destination
+{% endif %}
--- a/molecule/testinfra/mon/test_postfix.py
+++ b/molecule/testinfra/mon/test_postfix.py
@@ -37,6 +37,12 @@ def test_postfix_generic_maps(host):
     """
     assert not host.file("/etc/postfix/generic").exists
     assert not host.file("/etc/postfix/main.cf").contains("^smtp_generic_maps")
+    if host.system_info.codename == "focal":
+        assert host.file("/etc/postfix/main.cf").contains(
+                         "^smtpd_recipient_restrictions = reject_unauth_destination")
+    if host.system_info.codename == "xenial":
+        assert not host.file("/etc/postfix/main.cf").contains(
+                             "^smtpd_recipient_restrictions = reject_unauth_destination")
 
 
 def test_postfix_service(host):


install_files/ansible-base/roles/postfix/templates/main.cf
```diff
@@ -62,3 +62,7 @@ maximal_queue_lifetime = 14d
 # Used to remap outbound from address in emails
 smtp_generic_maps = hash:/etc/postfix/generic
 {% endif %}
+{% if securedrop_target_distribution == "focal" %}
```
emkll: In it's current form, this condition will never be met in production context on a Focal install. This is because `securedrop_target_distribution` is mainly used for the package build logic, and defaults to `xenial` in https://github.com/freedomofpress/securedrop/blob/2de09b4c5b1bf42deaa4331ac33915b5752d98ed/install_files/ansible-base/group_vars/all/securedrop#L29 In staging context, is it overriden in https://github.com/freedomofpress/securedrop/blob/2de09b4c5b1bf42deaa4331ac33915b5752d98ed/molecule/libvirt-staging-focal/ansible-override-vars.yml#L10 This means that while CI/tests are passing (they are using staging environments), the logic as written here will not work in production VMs nor hardware; the condition tested will never be met since `securedrop_target_distribution` is undefined. Using `ansible_distribution_release` should work for all cases (staging _and_ production) here.

emkll (changes requested):
Thanks @kushaldas , it appears that while the changes proposed here work in staging environments, they will not work in production context (see inline).

Furthermore, could you please elaborate on the following directive in the PR description:
> sudo apt install ubuntu-release-upgrader-core

Does this warrant a follow up issue?

emkll (approved):
Thanks @kushaldas . As discussed in https://github.com/freedomofpress/securedrop/pull/5777#pullrequestreview-585783064 , I believe the variable you are looking for is `ansible_distribution_release`, and not `ansible_distribution`, which may explain the legitimate test failures in https://app.circleci.com/pipelines/github/freedomofpress/securedrop/1828/workflows/9c782e72-1972-40e1-8517-521ed4b1a0c7/jobs/49990 .

I have taken the liberty of force pushing this branch and rebasing on latest develop. Approving based on the diff, but please take a look and merge once CI is passing.

PR #5772, opened 2021-02-04
Bump securedrop-grsec-focal metapackage to 5.4.88

emkll:
## Status

Ready for review
## Description of Changes

Fixes #5479 and towards https://github.com/freedomofpress/securedrop/issues/4768

Updates the kernel metapackage version to 5.4.88 for Focal installs.

See https://github.com/freedomofpress/securedrop-dev-packages-lfs/pull/89 

## Testing

- [ ] Focal metapackage pulls in 5.4.88 kernel image
- [ ] Xenial metapackage pulls in 4.14.188 kernel image
## Deployment

For new and existing installs, these changes will be distributed via deb packages through apt.

### If you made changes to the server application code:

- [ ] Linting (`make lint`) and tests (`make test`) pass in the development container

### If you made changes to `securedrop-admin`:

- [ ] Linting and tests (`make -C admin test`) pass in the admin development container

### If you made changes to the system configuration:

- [ ] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

### If you added or removed a file deployed with the application:

- [ ] I have updated AppArmor rules to include the change

### If you made non-trivial code changes:

- [x] I have written a test plan and validated it for this PR

PR #5712, opened 2021-01-15
Get testinfra working on Qubes

rmol:
## Status

Ready for review

## Description of Changes

Fixes #5377. 

Adds qubes-specific configuration to get testinfra tests passing with our Qubes staging environment.

The hostnames (e.g. `app-staging` versus `sd-staging-app`) and network addresses were the culprit in a lot of failures. 

## Testing

- `make build-debs`

While that's running, adjust the hostnames in your `sd-staging-(app|mon)-base-xenial` VMs:
- change `sd-staging-(app|mon)` to `(app|mon)-staging` in `/etc/hostname` and `/etc/hosts`

Once the debs are built:
- `molecule destroy -s qubes-staging-xenial`
- `make staging`
- `molecule verify -s qubes-staging-xenial` or `make testinfra`

The staging qubes should be created and configured without error, and all tests should pass.

## Deployment

This is dev/test only.

### If you made non-trivial code changes:

- [x] I have written a test plan and validated it for this PR

Choose one of the following:

- [x] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-docs) for these changes, or will do so later (the hostname changes need to be captured in `qubes_staging.rst`)
- [ ] I would appreciate help with the documentation
- [ ] These changes do not require documentation

PR #5780, opened 2021-02-09
Fixes #5776 adds iptables-persistent dependency on Focal

### LINKED ISSUE #5776: [Focal] iptables rules are not applied on boot
## Description

Towards #4768 

The firewall logic is not working reliably in Ubuntu Focal, and firewall rules are not correctly restored on boot. This appears to be a discrepancy between Vagrant boxes and a default Ubuntu 20.04 install.

## Steps to Reproduce
In an Ubuntu Focal install (non-staging, and likely non-VM), run `sudo iptables -L` or `sudo ip6tables -L`

## Expected Behavior

iptables and ip6tables rules should be correctly applied and enforced, with default DROP policy

## Actual Behavior

iptables rules are set to overly permissive defaults:
```
Chain INPUT (policy ACCEPT)
target     prot opt source               destination         
      

Chain FORWARD (policy ACCEPT)
target     prot opt source               destination         
       

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination         
```

kushaldas:
## Status

Ready for review.

## Description of Changes

Fixes #5776 

On Ubuntu Focal, we can use iptables-persistent package, and also
uses updated rules filepath based on distribution version.

## Testing

- [ ] Install on `Focal` on Hardware, and test `iptables -L` in both the servers
- [ ] Add some new rules, example, accept everything on `INPUT` chain.
- [ ] reboot and check `ipables -L` again. Everything should go back to default.
- [ ] Install on `Xenial` on Hardware, and test `iptables -L` in both the servers
- [ ] reboot and check `ipables -L` again.

## Deployment

2. New installs.

Choose one of the following:

Choose one of the following:

### DIFF
--- a/install_files/ansible-base/roles/app/handlers/main.yml
+++ b/install_files/ansible-base/roles/app/handlers/main.yml
@@ -5,8 +5,15 @@
     name: tor
     state: restarted
 
-- name: reload iptables rules
+- name: reload iptables rules for xenial
   shell: iptables-restore < /etc/network/iptables/rules_v4
+  when:
+    - ansible_distribution_release == 'xenial'
+
+- name: reload iptables rules for focal
+  shell: iptables-restore < /etc/iptables/rules.v4
+  when:
+    - ansible_distribution_release == 'focal'
 
 ## App/securedrop section
 - name: restart apache2
--- a/install_files/ansible-base/roles/app/handlers/main.yml
+++ b/install_files/ansible-base/roles/app/handlers/main.yml
@@ -6,14 +6,7 @@
     state: restarted
 
 - name: reload iptables rules for xenial
-  shell: iptables-restore < /etc/network/iptables/rules_v4
-  when:
-    - ansible_distribution_release == 'xenial'
-
-- name: reload iptables rules for focal
-  shell: iptables-restore < /etc/iptables/rules.v4
-  when:
-    - ansible_distribution_release == 'focal'
+  shell: iptables-restore < "{{ iptables_v4_path }}"
 
 ## App/securedrop section
 - name: restart apache2
--- a/install_files/ansible-base/roles/app/handlers/main.yml
+++ b/install_files/ansible-base/roles/app/handlers/main.yml
@@ -5,9 +5,6 @@
     name: tor
     state: restarted
 
-- name: reload iptables rules for xenial
-  shell: iptables-restore < "{{ iptables_v4_path }}"
-
 ## App/securedrop section
 - name: restart apache2
   service:
--- a/install_files/ansible-base/roles/restrict-direct-access/files/load_iptables
+++ b/install_files/ansible-base/roles/restrict-direct-access/files/load_iptables
@@ -2,6 +2,8 @@
 # Description: apply the securedrop iptable rules
 if [ -f /etc/network/iptables/rules_v4 ]; then
   iptables-restore < /etc/network/iptables/rules_v4
+elif [ -f /etc/iptables/rules.v4 ]; then
+  iptables-restore < /etc/iptables/rules.v4
 else
   echo "Iptables rules file does not exist"
   exit 1
@@ -9,6 +11,8 @@ fi
 
 if [ -f /etc/network/iptables/rules_v6 ]; then
   ip6tables-restore < /etc/network/iptables/rules_v6
+elif [ -f /etc/iptables/rules.v6 ]; then
+  ip6tables-restore < /etc/iptables/rules.v6
 else
   echo "Ip6tables rules file does not exist"
   exit 1
--- a/install_files/ansible-base/roles/restrict-direct-access/files/load_iptables
+++ b/install_files/ansible-base/roles/restrict-direct-access/files/load_iptables
@@ -2,8 +2,6 @@
 # Description: apply the securedrop iptable rules
 if [ -f /etc/network/iptables/rules_v4 ]; then
   iptables-restore < /etc/network/iptables/rules_v4
-elif [ -f /etc/iptables/rules.v4 ]; then
-  iptables-restore < /etc/iptables/rules.v4
 else
   echo "Iptables rules file does not exist"
   exit 1
@@ -11,8 +9,6 @@ fi
 
 if [ -f /etc/network/iptables/rules_v6 ]; then
   ip6tables-restore < /etc/network/iptables/rules_v6
-elif [ -f /etc/iptables/rules.v6 ]; then
-  ip6tables-restore < /etc/iptables/rules.v6
 else
   echo "Ip6tables rules file does not exist"
   exit 1
--- a/install_files/ansible-base/roles/restrict-direct-access/tasks/iptables.yml
+++ b/install_files/ansible-base/roles/restrict-direct-access/tasks/iptables.yml
@@ -35,6 +35,8 @@
     owner: root
     group: root
     dest: /etc/network/iptables
+  when:
+    - ansible_distribution_release == 'xenial'
 
 - name: Determine local platform specific routing info
   set_fact:
@@ -59,7 +61,7 @@
 - name: Copy IPv4 iptables rules.
   template:
     src: rules_v4
-    dest: /etc/network/iptables/rules_v4
+    dest: "{{ '/etc/iptables/rules.v4' if ansible_distribution_release == 'focal' else  '/etc/network/iptables/rules_v4' }}"
     owner: root
     mode: "0644"
   notify: drop flag for reboot
@@ -67,6 +69,6 @@
 - name: Copy IPv6 iptables rules.
   copy:
     src: iptables_rules_v6
-    dest: /etc/network/iptables/rules_v6
+    dest: "{{ '/etc/iptables/rules.v6' if ansible_distribution_release == 'focal' else  '/etc/network/iptables/rules_v6' }}"
     owner: root
     mode: "0644"
--- a/install_files/ansible-base/roles/restrict-direct-access/tasks/iptables.yml
+++ b/install_files/ansible-base/roles/restrict-direct-access/tasks/iptables.yml
@@ -63,7 +63,7 @@
 - name: Copy IPv4 iptables rules.
   template:
     src: rules_v4
-    dest: "{{ '/etc/iptables/rules.v4' if ansible_distribution_release == 'focal' else  '/etc/network/iptables/rules_v4' }}"
+    dest: "{{ iptables_v4_path }}"
     owner: root
     mode: "0644"
   notify: drop flag for reboot
@@ -71,6 +71,6 @@
 - name: Copy IPv6 iptables rules.
   copy:
     src: iptables_rules_v6
-    dest: "{{ '/etc/iptables/rules.v6' if ansible_distribution_release == 'focal' else  '/etc/network/iptables/rules_v6' }}"
+    dest: "{{ iptables_v6_path }}"
     owner: root
     mode: "0644"
--- a/install_files/ansible-base/roles/restrict-direct-access/tasks/iptables.yml
+++ b/install_files/ansible-base/roles/restrict-direct-access/tasks/iptables.yml
@@ -27,6 +27,8 @@
     dest: /etc/network/if-up.d/load_iptables
     owner: root
     mode: "0744"
+  when:
+    - ansible_distribution_release == 'xenial'
 
 - name: Create iptables directory.
   file:
--- a/install_files/ansible-base/roles/common/vars/Ubuntu_focal.yml
+++ b/install_files/ansible-base/roles/common/vars/Ubuntu_focal.yml
@@ -9,6 +9,7 @@ resolvconf_target_filepath: /etc/resolv.conf
 securedrop_common_packages:
   - apt-transport-https
   - aptitude
+  - iptables-persistent
   - unattended-upgrades
   - ntp
   - ntpdate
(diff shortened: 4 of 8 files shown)


conorsch:
CI failure in https://app.circleci.com/pipelines/github/freedomofpress/securedrop/1827/workflows/afc450be-1937-4dae-b6ed-e5b174ef0379/jobs/49991, OSSEC service was not listening. Kicked CI to see if it's a flake.

install_files/ansible-base/roles/app/handlers/main.yml
```diff
@@ -5,8 +5,15 @@
...
-- name: reload iptables rules
+- name: reload iptables rules for xenial
   shell: iptables-restore < /etc/network/iptables/rules_v4
+  when:
+    - ansible_distribution_release == 'xenial'
+
+- name: reload iptables rules for focal
+  shell: iptables-restore < /etc/iptables/rules.v4
+  when:
+    - ansible_distribution_release == 'focal'
```
conorsch: Since the only change here appears to be the filepath for the rules, let's make that a var and import it based on distro. See https://github.com/freedomofpress/securedrop/blob/2de09b4c5b1bf42deaa4331ac33915b5752d98ed/install_files/ansible-base/roles/common/tasks/main.yml#L2 for an example

install_files/ansible-base/roles/restrict-direct-access/files/load_iptables
```diff
@@ -2,13 +2,17 @@
...
+  iptables-restore < /etc/iptables/rules.v4
 else
   echo "Iptables rules file does not exist"
   exit 1
 fi
 
 if [ -f /etc/network/iptables/rules_v6 ]; then
   ip6tables-restore < /etc/network/iptables/rules_v6
+elif [ -f /etc/iptables/rules.v6 ]; then
+  ip6tables-restore < /etc/iptables/rules.v6
```
conorsch: As above, these paths can be interpolated from a var if we use a template. However, I see Focal-only paths in here, whereas the relevant task only copies this file on Xenial hosts. So these `elif`s appear to be unreachable.

install_files/ansible-base/roles/restrict-direct-access/tasks/iptables.yml
```diff
@@ -35,6 +47,8 @@
     owner: root
     group: root
     dest: /etc/network/iptables
+  when:
+    - ansible_distribution_release == 'xenial'
```
conorsch, line 41: If the `load_iptables` script is only copied under Xenial, then its content shouldn't change to address Focal support.

conorsch (changes requested):
Left some in-line comments about reducing duplication, similar in spirit to what's discussed for dual distro support in https://github.com/freedomofpress/securedrop/pull/5772#issuecomment-775983453. 

At present I'm not comfortable merging these changes until we understand completely what the divergence is between the two distros. I'll keep investigating on root causes and report findings in https://github.com/freedomofpress/securedrop/issues/5776

rmol:
I think the divergence is explained by the fact that on Focal server, network configuration is handled by a combination of [netplan and systemd-networkd](https://ubuntu.com/server/docs/network-configuration) (search for "hook scripts"; the document lacks anchors).

I tried simply moving our `/etc/network/if-up.d/load_iptables` script to `/etc/networkd-dispatcher/routable.d/00_load_iptables` as suggested by this [conversion table](https://netplan.io/faq/#use-pre-up%2C-post-up%2C-etc.-hook-scripts) and the iptables rules were correctly started when I bounced the interface.

I have not changed the Ansible config and rebuilt staging, or tried it on hardware. There are also other hook scripts installed into /etc/network/if-up.d by other packages (`resolvconf`, `ethtool`), which I can't explain, unless they simply weren't updated for the Focal network changes, but `ethtool` is in the `utils` section. :man_shrugging:

conorsch:
> I tried simply moving our /etc/network/if-up.d/load_iptables script to /etc/networkd-dispatcher/routable.d/00_load_iptables

That's a great suggestion, and matches with the use-a-path-based-on-vars comments above. We're bumping up against reimplementing iptables-persistent functionality, so let's not rule that out yet. As mentioned in https://github.com/freedomofpress/securedrop/issues/5776#issuecomment-776016341, tying the restore logic to the ifup event should be the strictest of all possible options. That's based on my assumption that iptables-persistent will always restore whatever rules are present on system shutdown—I'll read up on that now.

kushaldas:
I noticed https://unix.stackexchange.com/questions/125833/why-isnt-the-iptables-persistent-service-saving-my-changes

conorsch:
> That's based on my assumption that iptables-persistent will always restore whatever rules are present on system shutdown

Not so. Performed some interactive testing, and it appears that iptables-persistent is behaving the way we'd want: ad-hoc rules added on the command-line are wiped after reboot, and the original ruleset is enforced on subsequent boot. With that concern out of the way, I'm comfortable switching to iptables-persistent for Focal only (cc @emkll for visibility). 

@kushaldas Right now, CI is still running on https://github.com/freedomofpress/securedrop/pull/5712, but I've approved it. Please merge it and rebase these changes on top, then take a look at the in-line comments here and try to refactor. Also, please update the test plan to include verification that ad-hoc rules are removed. In my testing, I tried setting the input policy to ACCEPT, as well as adding a global allow on TCP 443 inbound.

install_files/ansible-base/roles/restrict-direct-access/tasks/iptables.yml
kushaldas: Done.

install_files/ansible-base/roles/restrict-direct-access/files/load_iptables
kushaldas: Removed now.

kushaldas:
Made all other changes but https://github.com/freedomofpress/securedrop/pull/5712 still needs work, so did not merge it. @conorsch

conorsch:
Rebased on top of latest develop, to include changes from https://github.com/freedomofpress/securedrop/pull/5712

conorsch:
Still seeing the [same CI failure](https://app.circleci.com/pipelines/github/freedomofpress/securedrop/1841/workflows/586b70aa-3495-4a79-818a-4a3c7dde4dfe/jobs/50100):

<details>
<summary>
failing test output
</summary>

```
assert False == True
  +False
  -True
host = <testinfra.host.Host ansible://mon-staging>
ossec_service = {'host': '0.0.0.0', 'listening': True, 'port': 1514, 'proto': 'udp'}

    @pytest.mark.skip_in_prod
    @pytest.mark.parametrize('ossec_service', [
        dict(host="0.0.0.0", proto="tcp", port=22, listening=True),
        dict(host="0.0.0.0", proto="udp", port=1514, listening=True),
        dict(host="0.0.0.0", proto="tcp", port=1515, listening=False),
    ])
    def test_listening_ports(host, ossec_service):
        """
        Ensure the OSSEC-related services are listening on the
        expected sockets. Services to check include ossec-remoted
        and ossec-authd. Helper services such as postfix are checked
        separately.
    
        Note that the SSH check will fail if run against a prod host, due
        to the SSH-over-Tor strategy. We can port the parametrized values
        to config test YAML vars at that point.
        """
        socket = "{proto}://{host}:{port}".format(**ossec_service)
        with host.sudo():
>           assert (host.socket(socket).is_listening ==
                    ossec_service['listening'])
E           assert False == True
E             +False
E             -True

```

</details>

Will dig into it more today.

rmol:
This could be a red herring, but `host.socket` [supposedly requires `netstat`](https://testinfra.readthedocs.io/en/latest/modules.html#testinfra.modules.socket.Socket), and at least on my fresh staging mon server, it's not installed. Has this ever worked for any tester or service?

conorsch:
That's helpful, @rmol, but also a bit confusing, because the test is passing locally for me. It also passed in CI as recently as merge of #5712!

conorsch:
The fact that CI is reliably failing here, yet passing on https://github.com/freedomofpress/securedrop/pull/5783, indicates we clearly have a regression in the firewall logic. I'm optimistic that the changes proposed in #5783 will resolve here, too. Pushed a separate branch now to evaluate behavior in CI: https://github.com/freedomofpress/securedrop/tree/5776-iptables-persistent-for-focal-with-reconnect

conorsch:
>  I'm optimistic that the changes proposed in #5783 will resolve here, too. Pushed a separate branch now to evaluate behavior in CI

Sure enough, tests are passing: https://app.circleci.com/pipelines/github/freedomofpress/securedrop/1844/workflows/835e3251-afb0-412d-95b3-b25373c5dbc3/jobs/50137 So let's get #5783 reviewed and in, then rebase this one.

kushaldas:
And magic, all tests are passing!!!

conorsch:
Thanks, @kushaldas! Taking another look now.

install_files/ansible-base/roles/restrict-direct-access/tasks/iptables.yml
```diff
@@ -59,14 +75,14 @@
 - name: Copy IPv4 iptables rules.
   template:
     src: rules_v4
-    dest: /etc/network/iptables/rules_v4
+    dest: "{{ '/etc/iptables/rules.v4' if ansible_distribution_release == 'focal' else  '/etc/network/iptables/rules_v4' }}"
```
conorsch: Path should use the var created above

install_files/ansible-base/roles/restrict-direct-access/tasks/iptables.yml
```diff
@@ -59,14 +75,14 @@
...
     owner: root
     mode: "0644"
   notify: drop flag for reboot
 
 - name: Copy IPv6 iptables rules.
   copy:
     src: iptables_rules_v6
-    dest: /etc/network/iptables/rules_v6
+    dest: "{{ '/etc/iptables/rules.v6' if ansible_distribution_release == 'focal' else  '/etc/network/iptables/rules_v6' }}"
```
conorsch: Path should use the var created above

install_files/ansible-base/roles/restrict-direct-access/tasks/iptables.yml
```diff
@@ -21,12 +21,26 @@
   delegate_to: localhost
   delegate_facts: True
 
+- name: Install iptables-persistent package
+  apt:
+    pkg: iptables-persistent
+    state: latest
+    update_cache: yes
+    cache_valid_time: 3600
```
conorsch: We can avoid another apt run by placing the package requirement in the common role, in the focal vars.

conorsch:
Changes are solid. To test, I set `sudo iptables -P INPUT ACCEPT` as an intentional breakage of the rules. I applied this diff:

```
diff --git a/molecule/qubes-staging-focal/molecule.yml b/molecule/qubes-staging-focal/molecule.yml
index 5bb3873cf..f8f538b29 100644
--- a/molecule/qubes-staging-focal/molecule.yml
+++ b/molecule/qubes-staging-focal/molecule.yml
@@ -55,5 +55,6 @@ verifier:
   options:
     n: auto
     v: 2
+    k: iptables_rules
   env:
     SECUREDROP_TESTINFRA_TARGET_HOST: qubes-staging
```

to run only the iptables tests, and confirmed failing:

```
 >           assert iptables_expected == iptables
    E           assert '*filter\n:IN... DROP\nCOMMIT' == '*filter\n:IN... DROP\nCOMMIT'
    E               *filter
    E             - :INPUT ACCEPT
    E             + :INPUT DROP
    E               :FORWARD DROP
    E               :OUTPUT DROP
    E               :LOGNDROP -
    E               -A INPUT -p tcp -m state --state RELATED,ESTABLISHED -m comment --comment "Allow traffic back for tor" -j ACCEPT...
    E             
    E             ...Full output truncated (39 lines hidden), use '-vv' to show
    
    ../testinfra/app/test_app_network.py:54: AssertionError
```

After rebooting the host, all tests are passing again:

```
../testinfra/mon/test_mon_network.py::test_mon_iptables_rules[ansible:/mon-staging]
    ../testinfra/app/test_app_network.py::test_app_iptables_rules[ansible:/app-staging]
    [gw0] [ 50%] PASSED ../testinfra/mon/test_mon_network.py::test_mon_iptables_rules[ansible:/mon-staging]
    [gw1] [100%] PASSED ../testinfra/app/test_app_network.py::test_app_iptables_rules[ansible:/app-staging]
```

In addition to changing the policy items, I added additional rules such as all TCP 443 allowed in, and confirmed that a reboot removed the customizations. 

There are a few changes I'd like to make to clean up the formatting, which I'll append now. Very pleased with the results of testing, these changes are solid and Focal-only, as we want.

install_files/ansible-base/roles/app/handlers/main.yml
```diff
@@ -5,8 +5,8 @@
     name: tor
     state: restarted
 
-- name: reload iptables rules
-  shell: iptables-restore < /etc/network/iptables/rules_v4
+- name: reload iptables rules for xenial
+  shell: iptables-restore < "{{ iptables_v4_path }}"
```
conorsch: This handler isn't actually used anywhere, neither in this PR nor in develop. OK to remove

conorsch (dismissed):

conorsch:
Works as advertised!

conorsch (approved):
Re-approving post-rebase, CI is green, merging!

PR #5784, opened 2021-02-11
Fixes #5128 removes snappy from the default installation

kushaldas:
## Status

Ready for review

## Description of Changes

Fixes #5128 

Removes snappy and also removes the related apparmor configuration.

## Testing
- [x] `molecule converge -s libvirt-staging-focal`
- [x] `molecule login -s libvirt-staging-focal app-staging`
- [x] check `snapd` package is not present
- [x] repeat the same for mon-staging
- [x] Now repeat the steps for `Xenial` 

## Deployment

Any special considerations for deployment? Consider both:

1. Upgrading existing production instances.
2. New installs.

### If you made changes to the server application code:

- [ ] Linting (`make lint`) and tests (`make test`) pass in the development container

### If you made changes to `securedrop-admin`:

- [ ] Linting and tests (`make -C admin test`) pass in the admin development container

### If you made changes to the system configuration:

- [ ] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

### If you added or removed a file deployed with the application:

- [ ] I have updated AppArmor rules to include the change

### If you made non-trivial code changes:

- [ ] I have written a test plan and validated it for this PR

Choose one of the following:

- [ ] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-docs) for these changes, or will do so later
- [ ] I would appreciate help with the documentation
- [ ] These changes do not require documentation

### If you added or updated a code dependency:

Choose one of the following:

- [ ] I have performed a diff review and pasted the contents to [the packaging wiki](https://github.com/freedomofpress/securedrop-debian-packaging/wiki)
- [ ] I would like someone else to do the diff review

PR #5786, opened 2021-02-11
Disables LTS upgrade prompt for Focal

### LINKED ISSUE #5782: [Focal] LTS channel allowed for upgrades
## Description

The release management settings for Focal permit `do-release-upgrade` to the next LTS version. There isn't one of those yet, but tests are failing because of it:

```
        def test_release_manager_upgrade_channel(host):
            """
            Ensures that the `do-release-upgrade` command will not
            suggest upgrades from Xenial to Bionic (which is untested
            and unsupported.)
            """
            expected_channels = {
                "xenial": "never",
                "focal": "never",
            }
    
            config_path = "/etc/update-manager/release-upgrades"
            assert host.file(config_path).is_file
    
            raw_output = host.check_output("grep '^Prompt' {}".format(config_path))
            _, channel = raw_output.split("=")
    
            expected_channel = expected_channels[host.system_info.codename]
    >       assert channel == expected_channel
    E       AssertionError: assert 'lts' == 'never'
    E         - never
    E         + lts
```

To see that test fail, you'll need to use hardware or Qubes VMs (installed from ISO), since the bento boxes modify the setting https://github.com/chef/bento/blob/3dd165cf422572ab4240c8f1ce9a7e40b7960457/packer_templates/ubuntu/scripts/update.sh#L5 in a similar fashion to what we implemented for Xenial in https://github.com/freedomofpress/securedrop/pull/3546/files

## Comments

This is almost certainly a duplicate of #5781, but filing separately until that's confirmed.

conorsch:
## Status
Work in progress

## Description of Changes

We don't want the OS to nag admins, since we want to ensure the next LTS
is fully supported by SecureDrop before recommending an in-place
upgrade. Disable the prompt on Focal, same as we've done for Xenial in
the past.

Closes #5782.

## Testing
Still need to update the testinfra tests. 

## Deployment
TK

### DIFF
--- a/install_files/securedrop-config-focal/DEBIAN/postinst
+++ b/install_files/securedrop-config-focal/DEBIAN/postinst
@@ -4,8 +4,17 @@
 set -e
 set -x
 
+# Issue #5782
+# Set Prompt=never for distro upgrades
+update_release_prompt() {
+    set -e
+    upgrade_config='/etc/update-manager/release-upgrades'
+    sed -i 's/Prompt=.*/Prompt=never/' "$upgrade_config"
+}
+
 case "$1" in
     configure)
+    update_release_prompt
     # Configuration required for unattended-upgrades
     cp /opt/securedrop/20auto-upgrades /etc/apt/apt.conf.d/
     cp /opt/securedrop/50unattended-upgrades /etc/apt/apt.conf.d/


conorsch:
No test updates are required, since the tests were already failing, although not in CI. I'm going to wait a bit longer until #5784 is in, then rebase this one, which should get 100% of the Focal tests passing on Qubes ISO-based VMs.

emkll:
Rebased on latest `develop` after merging #5784

kushaldas (approved):
Worked as in intended on `Focal` in both `app` and `mon` servers.

PR #5788, opened 2021-02-12
Check for updates before most securedrop-admin commands

### LINKED ISSUE #5317: Warn or fail when running outdated playbook
It's not uncommon that admins run an outdated playbook, because they've not updated their Admin Workstation for whatever reason (e.g. due to an update failure). Running an outdated playbook can have harmful consequences, reinstating changed configurations, or even breaking functionality. For an existing install, `securedrop-admin install` should attempt to obtain the version info from the server, and fail or warn if it doesn't correspond to the version info running on the workstation.

## User Story

As an administrator, I don't want to accidentally mess up the configuration of my SecureDrop installation, so that I can avoid spending the rest of my afternoon debugging what I just did.

eloquence:
## Status

Ready for review

## Description of Changes

Fixes #5317 

Provides a new decorator, `@update_check_required`, and adds it to `securedrop-admin` subcommands which make significant modifications on the workstation or the server. If the update check detects a need to update (using the existing logic in `securedrop-admin`), it will display a message explaining how to update. A `--force` option is provided in case the user wishes to intentionally run a subcommand with a playbook version that differs from the most recent signed release tag (e.g., during QA).

## Testing

This change is best tested on a real or virtualized *Admin Workstation* in Tails. 

1. Create a backup copy of your `~/Persistent/securedrop` as warranted.
2. Fetch all updates with `git fetch --all` and check out git tag `1.7.1`
3. Apply the changes from this branch via `git cherry-pick -n 75249c2ef9dc9ad383fa02e6c1fe9a3a4117b38d && git cherry-pick -n 26e608723fed7e7a0f166caed3d2e395583a02f1` . This will stage the changes for commit -- you can verify this via `git diff --cached`.
4. Run `securedrop-admin check_for_updates`
- [ ] Observe that the command reports "All updates applied"
5. Run `securedrop-admin logs`
- [ ] Observe that securedrop-admin checks for updates before running the playbook
- [ ] Observe that securedrop-admin attempts to run the logs playbook
6. Switch to an older tag (you may need to discard re-apply the changes)
7. Run `securedrop-admin logs`
- [ ] Observe that securedrop-admin checks for updates and reports that updates are available
- [ ] Observe that securedrop-admin fails to complete due to a version mismatch
- [ ] Observe that securedrop-admin reports the checked out tag as `Current branch status`
8. Run `securedrop-admin --force logs`
- [ ] Observe that securedrop-admin now skips the update check and runs the logs playbook
9. Check out the `develop` branch and repeat steps 7 and 8 and the associated checks.
10. Bounce the network connection in Tails.
- [ ] Observe that the graphical updater runs to completion and puts you back on the 1.7.1 tag.
- [ ] (Optional) Test and observe that all `securedrop-admin` subcommands annotated via the decorator (see diff) run the update check and produce the expected help output
11. Discard the unstaged changes with `git reset --hard`.

## Deployment

Note that this will only be available to users _after_ the next workstation update.

### Docs

While this change is largely self-documenting and does not conflict with previous guidance, I intend to add a brief explanation to the documentation, in particular https://docs.securedrop.org/en/stable/admin.html#the-securedrop-admin-utility

### DIFF
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -24,6 +24,7 @@ instances.
 """
 
 import argparse
+import functools
 import ipaddress
 import logging
 import os
@@ -45,6 +46,9 @@ import yaml
 from pkg_resources import parse_version
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import x25519
+
+from typing import cast
+
 from typing import List
 
 from typing import Set
@@ -80,6 +84,11 @@ class JournalistAlertEmailException(Exception):
 
 # The type of each entry within SiteConfig.desc
 _T = TypeVar('_T', bound=Union[int, str, bool])
+
+# The function type used for the @update_check_required decorator; see
+# https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators
+_FuncT = TypeVar('_FuncT', bound=Callable[..., Any])
+
 # (var, default, type, prompt, validator, transform, condition)
 _DescEntryType = Tuple[str, _T, Type[_T], str, Optional[Validator], Optional[Callable], Callable]
 
@@ -516,7 +525,7 @@ class SiteConfig:
                 self._config_in_progress[var] = ''
                 continue
             self._config_in_progress[var] = self.user_prompt_config_one(desc,
-                                                            self.config.get(var))  # noqa: E501
+                                                                        self.config.get(var))  # noqa: E501
         return self._config_in_progress
 
     def user_prompt_config_one(
@@ -690,6 +699,45 @@ def setup_logger(verbose: bool = False) -> None:
     sdlog.addHandler(stdout)
 
 
+def update_check_required(cmd_name: str) -> Callable[[_FuncT], _FuncT]:
+    """
+    This decorator can be added to any subcommand that is part of securedrop-admin
+    via `@update_check_required("name_of_subcommand")`. It forces a check for
+    updates, and aborts if the locally installed code is out of date. It should
+    be generally added to all subcommands that make modifications on the
+    server or on the Admin Workstation.
+
+    The user can override this check by specifying the --force argument before
+    any subcommand.
+    """
+    def decorator_update_check(func: _FuncT) -> _FuncT:
+        @functools.wraps(func)
+        def wrapper(*args: Any, **kwargs: Any) -> Any:
+            cli_args = args[0]
+            if cli_args.force:
+                sdlog.info("Skipping update check because --force argument was provided.")
+                return func(*args, **kwargs)
+
+            update_status, latest_tag = check_for_updates(cli_args)
+            if update_status is True:
+                sdlog.error("You are not running the most recent signed SecureDrop release "
+                            "on this workstation.")
+                sdlog.error("Latest available version: {}".format(latest_tag))
+                sdlog.error("Running outdated or mismatched code can cause significant "
+                            "technical issues.")
+                sdlog.error("If you are certain you want to proceed, run:\n\n\t"
+                            "./securedrop-admin --force {}\n".format(cmd_name))
+                sdlog.error("To apply the latest updates, run:\n\n\t"
+                            "./securedrop-admin update\n")
+                sdlog.error("If this fails, see the latest upgrade guide on "
+                            "https://docs.securedrop.org/ for instructions.")
+                sys.exit(1)
+            return func(*args, **kwargs)
+        return cast(_FuncT, wrapper)
+    return decorator_update_check
+
+
+@update_check_required("sdconfig")
 def sdconfig(args: argparse.Namespace) -> int:
     """Configure SD site settings"""
     SiteConfig(args).load_and_update_config(validate=False)
@@ -752,8 +800,10 @@ def find_or_generate_new_torv3_keys(args: argparse.Namespace) -> int:
     return 0
 
 
+@update_check_required("install")
 def install_securedrop(args: argparse.Namespace) -> int:
     """Install/Update SecureDrop"""
+
     SiteConfig(args).load_and_update_config(prompt=False)
 
     sdlog.info("Now installing SecureDrop on remote servers.")
@@ -767,6 +817,7 @@ def install_securedrop(args: argparse.Namespace) -> int:
     )
 
 
+@update_check_required("verify")
 def verify_install(args: argparse.Namespace) -> int:
     """Run configuration tests against SecureDrop servers"""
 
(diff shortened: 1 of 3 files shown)


admin/securedrop_admin/__init__.py
```diff
@@ -1021,6 +1077,8 @@ class ArgParseFormatterCombo(argparse.ArgumentDefaultsHelpFormatter,
                         help="Increase verbosity on output")
     parser.add_argument('-d', action='store_true', default=False,
                         help="Developer mode. Not to be used in production.")
+    parser.add_argument('--force', action='store_true', required=False,
```
kushaldas, line 1108: `--force` works as it should be.

admin/securedrop_admin/__init__.py
```diff
@@ -690,6 +699,45 @@ def setup_logger(verbose: bool = False) -> None:
...
+                            "technical issues.")
+                sdlog.error("If you are certain you want to proceed, run:\n\n\t"
+                            "./securedrop-admin --force {}\n".format(cmd_name))
+                sdlog.error("To apply the latest updates, run:\n\n\t"
+                            "./securedrop-admin update\n")
+                sdlog.error("If this fails, see the latest upgrade guide on "
+                            "https://docs.securedrop.org/ for instructions.")
+                sys.exit(1)
+            return func(*args, **kwargs)
+        return cast(_FuncT, wrapper)
```
kushaldas, line 748: I never used `cast` before, learned a new thing today.

admin/securedrop_admin/__init__.py
```diff
@@ -690,6 +699,45 @@ def setup_logger(verbose: bool = False) -> None:
...
+            cli_args = args[0]
+            if cli_args.force:
+                sdlog.info("Skipping update check because --force argument was provided.")
+                return func(*args, **kwargs)
+
+            update_status, latest_tag = check_for_updates(cli_args)
+            if update_status is True:
+                sdlog.error("You are not running the most recent signed SecureDrop release "
+                            "on this workstation.")
+                sdlog.error("Latest available version: {}".format(latest_tag))
```
kushaldas, line 729: I wonder if we can also print the current branch/tag the repository is in.

kushaldas (commented):
Did multiple reboots and also `./securedrop-admin --force tailsconfig` to make sure things are okay. Rest of the change looks okay to me. Worked as intended.

I will do one more round after this comes out of draft and formally approve it.

admin/securedrop_admin/__init__.py
eloquence: It was new to me, too; I would have preferred to avoid it but sometimes satisfying `mypy` does make the code a bit more difficult to read. :/

admin/securedrop_admin/__init__.py
eloquence: I played with that as well in an earlier version. The problem is that the `current_tag` variable is only useful for version comparisons, as it produces highly confusing output e.g., on `develop`, due to its reliance on `git describe`. For current `develop`, that results in `0.6-rc2-2646-g5eaad3b04` (!), as that's the most recent reachable tag. In addition to a branch or tag, the local repo may be in `HEAD detached` state, or in any of a number of different git states (dirty repo, rebase in progress, etc.).

The first line of `git status` is generally informative, as is the starred line of `git branch`. We could collect either line (we should probably do so in `check_for_updates`) and print it out. I think that would be helpful to ease troubleshooting. 

I have a slight preference for `git branch` as its output tends to be a bit more terse. I would suggest doing this in a follow-up PR, to keep this one tightly scoped to the update check. Here's a one-liner for doing that in Python (without error handling):

```python
re.search(r"\* (.*)\n", subprocess.check_output(['git', 'branch']).decode('utf-8')).group(1)
```

eloquence:
(Rebased on latest `develop` and squashed.)

eloquence:
(Added test plan and promoted to ready.)

admin/securedrop_admin/__init__.py
zenmonkeykstop: I have no strong feelings one way or the other if the goal is just to print out the current branch/tag - but! the output from `git status` will also let you know if there are code changes in the repo (e.g. unsanctioned poking at ansible defaults), giving you a few more options:

- `you're on x branch/tag, update to <latest>`
- `you're on x branch/tag and there are changed files, resolve this then upgrade to latest`

This is probably not going to be a one-liner though.

eloquence: For now pushed to a WIP branch: https://github.com/freedomofpress/securedrop/compare/5317-check-for-updates..5317-check-for-updates-debug-info

This tries to get us the best of both worlds:
- use `git branch` for terse information that can be included in the output
- recommends using `git status` for troubleshooting

Will still need its own test and may fail CI as-is. If the approach seems sound, happy to include in this PR, or in a follow-up PR, at the reviewer's discretion.

eloquence: Since review is still pending, I'll just clean this change up a bit further and fold it into this PR now.

eloquence: Done, and test plan updated. I've refrained from any further mangling of the `git branch` output as it's possible to use Tails in languages other than English.

kushaldas (approved):
Code change looks good.

1. Create a backup copy of your `~/Persistent/securedrop` as warranted.
2. Fetch all updates with `git fetch --all` and check out git tag `1.7.1`
3. Apply the changes from this branch via `git cherry-pick -n 75249c2ef9dc9ad383fa02e6c1fe9a3a4117b38d && git cherry-pick -n 26e608723fed7e7a0f166caed3d2e395583a02f1` . This will stage the changes for commit -- you can verify this via `git diff --cached`.
4. Run `securedrop-admin check_for_updates`
- [x] Observe that the command reports "All updates applied"
5. Run `securedrop-admin logs`
- [x] Observe that securedrop-admin checks for updates before running the playbook
- [x] Observe that securedrop-admin attempts to run the logs playbook
6. Switch to an older tag (you may need to discard re-apply the changes)
7. Run `securedrop-admin logs`
- [x] Observe that securedrop-admin checks for updates and reports that updates are available
- [x] Observe that securedrop-admin fails to complete due to a version mismatch
- [x] Observe that securedrop-admin reports the checked out tag as `Current branch status`
8. Run `securedrop-admin --force logs`
- [x] Observe that securedrop-admin now skips the update check and runs the logs playbook
9. Check out the `develop` branch and repeat steps 7 and 8 and the associated checks.
10. Bounce the network connection in Tails.
- [x] Observe that the graphical updater runs to completion and puts you back on the 1.7.1 tag.
- [x] (Optional) Test and observe that all `securedrop-admin` subcommands annotated via the decorator (see diff) run the update check and produce the expected help output
11. Discard the unstaged changes with `git reset --hard`.

PR #5793, opened 2021-02-16
Remove aptitude, disable install-recommends

conorsch:
## Status

Ready for review 

## Description of Changes

We were using "aptitude" in a few places, but it's not really necessary.
We can rely on apt-get for the functionality we need. In fact, it's more
predictable in terms of server state, especially using "dist-upgrade"
and also disabling automatic installation of "recommended" packages.

## Testing
1. Visual review.
2. Is the slight reduction in community/universe packages worthwhile? I'd say so.

 
## Deployment
The disabling-install-recommends functionality is applied to both Xenial & Focal hosts. The setting is mostly useful during initial install, when "recommended" packages will be pulled in. 

The goal of this change is to ensure predictable end state on new Focal installs, with a minimum of unnecessary packages.

### DIFF
--- a/install_files/ansible-base/roles/common/tasks/apt_upgrade.yml
+++ b/install_files/ansible-base/roles/common/tasks/apt_upgrade.yml
@@ -10,21 +10,30 @@
   tags:
     - apt
 
-- name: Install aptitude.
+- name: Configure apt preferences
+  template:
+    src: 80securedrop.j2
+    dest: /etc/apt/apt.conf.d/80securedrop
+    mode: 0644
+    owner: root
+    group: root
+  tags:
+    - apt
+
+- name: Remove aptitude.
   apt:
     name: aptitude
-    state: present
+    state: absent
   tags: apt
 
 - name: Check whether tor will be upgraded.
-  command: aptitude search '~U' --display-format '%p' --disable-columns
+  shell: apt list --upgradable -a | perl -F/ -lanE 'print $F[0]' | grep -v ^Listing | sort -u
   register: tor_upgradable_result
   changed_when: false
-  # aptitude >= 0.7.6 will exit non-zero if no hits
   failed_when: false
 
 - name: Hold tor package to prevent upgrade breaking SSH connection.
-  command: aptitude hold tor
+  command: apt-mark hold tor
   register: tor_hold_package_result
   when:
     - "'tor' in tor_upgradable_result.stdout_lines"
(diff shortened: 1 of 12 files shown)


install_files/ansible-base/roles/common/tasks/apt_upgrade.yml
```diff
@@ -33,22 +42,22 @@
     # If we're connecting to an Onion URL, then it's over Tor.
     - (ansible_host|default(ansible_host)).endswith('.onion')
 
-- name: Perform safe upgrade to ensure all the packages are updated.
+- name: Perform dist upgrade to ensure all the packages are updated.
   apt:
-    upgrade: safe
+    upgrade: dist
+    install_recommends: no
```
rmol: Is `install_recommends: no` necessary here after installing the new `80securedrop.j2` with `APT::Install-Recommends "false";`?

rmol (dismissed):
This worked perfectly in staging. There were no unexpected errors in the configuration tests. I was able to submit, then download and decrypt the submission. The background worker processes were functioning.

On hardware I did run into a problem with the order of operations when setting up the FPF apt repo. Since Focal packages are only on apt-test.freedom.press at this point, I had to adjust my working copy before running the playbook, and the current order overwrites the keyring when `securedrop-keyring` is installed. I appended a commit moving the configuration of the custom key to the end of that playbook. After that the installation completed without error, and the servers again passed manual inspection. The configuration tests do have a number of expected failures, but again none of them are due to these changes.

install_files/ansible-base/roles/common/tasks/apt_upgrade.yml
conorsch: It is not, it's completely superfluous. I started by adding `install_recommends: no`, then realized it would take a lot of edits. So provided that the `80securedrop` file gets written before the upgrade, as here, then we can remove the install_recommends parameter.

conorsch:
> On hardware I did run into a problem with the order of operations when setting up the FPF apt repo. Since Focal packages are only on apt-test.freedom.press at this point, I had to adjust my working copy before running the playbook, and the current order overwrites the keyring when securedrop-keyring is installed.

I'm thinking that deserves its own issue, @rmol, simply because the whole team is about to encounter that same problem during Focal QA. Up to you on how to handle this PR. I'd recommend adding https://github.com/freedomofpress/securedrop/pull/5793#discussion_r577689688 and squashing it into the primary commit, then experimenting with https://github.com/freedomofpress/securedrop/pull/5793/commits/d279727b580725678d86e63de1da1d30e627aa37 in a separate branch. All that after #5799, of course, so we can trust CI again.

rmol:
Good call. I've removed the `install_recommends: no` and squashed into the original commit. I've got tests for an apt bootstrapping fix running, should have a PR for that tomorrow.

rmol (approved):

PR #5811, opened 2021-02-23
No longer offer v2/v3 choice on installs without v2 configs

### LINKED ISSUE #5687: Disable v2 support for Focal installs
(Part of #5731)

Per our [roadmap](https://github.com/freedomofpress/securedrop/wiki/Development-Roadmap), support for v2 onion services will be removed for fresh installs starting with SecureDrop 1.8.0 (tentative release date: 2021-02-23). Accordingly, in addition to disabling v2 onion services during restores (#5677), we should ensure that v2 onion services cannot be configured during a fresh install of SecureDrop 1.8.0 or later.

Currently, during the installation, the user is prompted whether they want to enable v2 and/or v3 onion services, with a deprecation warning for v2:
https://github.com/freedomofpress/securedrop/blob/develop/admin/securedrop_admin/__init__.py#L435-L447

Instead, starting with SecureDrop 1.8.0, we can simply enforce v3 onion services. This should _not_ preclude a v2 onion service from being restored on _Xenial_.

This is consistent with the language in our previous deprecation warning ("support for v2 onion services will be removed in February 2021").

eloquence:
## Status

Ready for initial review; integration test updates and detailed test plan pending.

## Description

Changes `securedrop-admin sdconfig` behavior with regard to v2/v3 configuration:

- Existing installs with v2 configurations should see the same prompts as before (with slightly tweaked deprecation language), and the prompts should behave the same.
- New installs or installs that never had v2 enabled should see no v2/v3 related prompts. Existing configuration values should be honored; if there are none, v3 should be enabled and v2 disabled.

The v2/v3 configuration values are still written to `site-specific` as before. In SecureDrop 2.0.0, when support for v2 is fully removed from the codebase, we can potentially avoid remove these configuration lines altogether.

## Description of Changes

Resolves #5687

Does not prevent an existing v2 configuration from being applied against a Focal server, which should be handled separately.

## Testing

Best tested on a configured Admin Workstation. Detailed test plan to come, but for now:

1) If `site-specific` is not present and no v2 onion service configuration exists, does `securedrop-admin` still write a valid configuration file with `v2_onion_services: false` and `v3_onion_services: true` present, without prompting the user about v2/v3?
2) If `site-specific` is present and v2 onion service configuration exists, does `securedrop-admin` show the v2/v3 prompts, honor the current configuration values, correctly prevent both from being disabled, and correctly write the configuration back to disk?
3) If `site-specific` contains invalid values for `v3_onion_services` or `v2_onion_services` (e.g., a nonsense string) does the install subcommand still refuse to run due to a validation error?

"v2 onion service configuration exists" in this case means that `app-source-ths` exists in `install_files/ansible-base` and contains a v2 address (which is the check performed by `check_for_v2_onion`) .

## Checklist
- [ ] I have written a test plan and validated it for this PR (still to come)
- [ ] Docs updated (still to come)

### DIFF
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -89,7 +89,23 @@ _T = TypeVar('_T', bound=Union[int, str, bool])
 # https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators
 _FuncT = TypeVar('_FuncT', bound=Callable[..., Any])
 
-# (var, default, type, prompt, validator, transform, condition)
+# Configuration description tuples drive the CLI user experience and the
+# validation logic of the  securedrop-admin tool. A tuple is in the following
+# format.
+#
+# (var, default, type, prompt, validator, transform, condition):
+#
+# var         configuration variable name (will be stored in `site-specific`)
+# default     default value (can be a callable)
+# type        configuration variable type
+# prompt      text prompt presented to the user
+# validator   input validator based on `prompt_toolkit`'s Validator class
+# transform   transformation function to run on input
+# condition   condition under which this prompt is shown, receives the
+#             in-progress configuration object as input. Used for "if this
+#             then that" branching of prompts.
+#
+# The mypy type description of the format follows.
 _DescEntryType = Tuple[str, _T, Type[_T], str, Optional[Validator], Optional[Callable], Callable]
 
 
@@ -302,6 +318,8 @@ class SiteConfig:
             self.args.app_path).get_translations()
         translations_as_str = " ".join(translations)
 
+        is_v2_legacy_config = self.check_for_v2_onion()
+
         self.desc = [
             ('ssh_users', 'sd', str,
              'Username for SSH access to the servers',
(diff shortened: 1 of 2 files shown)


admin/securedrop_admin/__init__.py
```diff
@@ -89,7 +89,23 @@ class JournalistAlertEmailException(Exception):
...
+# default     default value (can be a callable)
+# type        configuration variable type
+# prompt      text prompt presented to the user
+# validator   input validator based on `prompt_toolkit`'s Validator class
+# transform   transformation function to run on input
+# condition   condition under which this prompt is shown, receives the
+#             in-progress configuration object as input. Used for "if this
+#             then that" branching of prompts.
+#
+# The mypy type description of the format follows.
```
eloquence, line 108: Added because I found the `securedrop-admin` code very difficult to read without documentation of this configuration tuple.

admin/tests/test_integration.py
```diff
@@ -51,74 +51,6 @@
 v3_onion_services: true
 '''
 
-WHEN_BOTH_TRUE = '''app_hostname: app
```
eloquence, line 54: I'm not suggesting we should remove those integration tests altogether; I've just removed them for now to observe that all other tests still pass. If the overall approach looks good, I'll add integration tests specifically for the behavior in legacy mode, which should broadly resemble the old v2/v3 tests.

eloquence:
After sleeping over it, I'm going to suggest closing this PR, and replacing it with a smaller change to the deprecation language (similar to the language in this PR). The v2/v3 logic in `securedrop-admin` is already pretty complex, and the conditional checks here create a risk of bugs where admins don't see an option when they need it. We can fully remove the v2/v3 options in 2.0.0 when they are no longer required.

If possible, I do think we should bail early in the install playbook if the v2 option is set and the servers are running Focal, as Kev suggested.

eloquence:
Closing per above and https://github.com/freedomofpress/securedrop/issues/5687#issuecomment-784414395

PR #5826, opened 2021-02-25
Normalize translator comments to correct them in Weblate

rmol:
## Status

Ready for review

## Description of Changes

Babel sometimes mangles translator comments extracted from source code, so Weblate ends up presenting nonsense to translators. I think it's because Babel eliminates duplicate lines in comments (https://github.com/python-babel/babel/issues/417 may be relevant?), so if multi-line comments for the same string in different locations differ only slightly, pieces of them will be eliminated, and the comment displayed in Weblate will end up being a jumble of the total set of distinct lines. 

For now, eliminating those slight differences in the few comments we have will get sensible hints to translators.

It would be good to correct this before the 1.8.0 translation period begins.

## Testing

Just verify this change reduces the set of `Translator:` comments down to two:

- `Translators: Precedes a message confirming the success of an operation.`
- `Translators: Error shown when a user has not selected items to act on.`

## Deployment

No. This only affects translator comments.

Choose one of the following:

### DIFF
--- a/securedrop/journalist_app/col.py
+++ b/securedrop/journalist_app/col.py
@@ -66,8 +66,7 @@ def make_blueprint(config: SDConfig) -> Blueprint:
         flash(
             Markup(
                 "<b>{}</b> {}".format(
-                    # Translators: Here, "Success!" appears before a message
-                    # confirming the success of an operation.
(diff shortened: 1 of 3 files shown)


sssoleileraaa (approved):
lookin good

PR #5853, opened 2021-03-05
Adjust timing of daily apt tasks

### LINKED ISSUE #5851: Override unattended-upgrade timers to make updates more closely aligned with cron-apt behaviour
## Description

By default, update and upgrade unattended-upgrade actions are triggered by systemd timers with a randomized delay of up t o12 hours. This makes updates occur at erratic intervals. The timing and delay is configurable via `systemctl edit <timername>` or by adding override.conf files under `/etc/systemd/system/...` - ideally the timers should be set to trigger just before the nightly server reboot, similarly to the cron-apt behaviour in Xenial

## User Research Evidence
Zip. Nada.

## User Stories
As an admin, I'd like to ensure that reboots and updates are applied during periods with reduced demand.

rmol:
## Status

Work in progress (timing needs to honor `sdconfig` options)

## Description of Changes

Fixes #5851.

Adjusts the timing of the `apt-daily` `systemd` timer so that updates are downloaded about every three hours.

Adjusts the timing of the `apt-daily-upgrade` timer so that the updates are applied at a random time between 0400 and 0500.

## Testing

- `make build-debs-focal && make staging-focal`
- On each server, check that:
  - An `override.conf` file is present in both of:
    - `/etc/systemd/system/apt-daily.timer.d`
    - `/etc/systemd/system/apt-daily-upgrade.timer.d`
  - The `CalendarOn` and `RandomizedDelaySec` entries in those files match what's reported by:
    - `systemctl show apt-daily.timer`
    - `systemctl show apt-daily-upgrade.timer`

## Deployment

This should ensure that updates are more reliably noticed and applied every day. 

Choose one of the following:

### DIFF
--- a/install_files/securedrop-config-focal/etc/systemd/system/apt-daily-upgrade.timer.d/override.conf
+++ b/install_files/securedrop-config-focal/etc/systemd/system/apt-daily-upgrade.timer.d/override.conf
@@ -1,0 +1,4 @@
+[Timer]
+OnCalendar=
+OnCalendar=04:00
+RandomizedDelaySec=1h
--- a/install_files/securedrop-config-focal/etc/systemd/system/apt-daily.timer.d/override.conf
+++ b/install_files/securedrop-config-focal/etc/systemd/system/apt-daily.timer.d/override.conf
@@ -1,0 +1,5 @@
+[Timer]
+OnCalendar=
+OnCalendar=*-*-* 00/3:00
+RandomizedDelaySec=1h
+Persistent=true
--- a/molecule/testinfra/common/test_automatic_updates.py
+++ b/molecule/testinfra/common/test_automatic_updates.py
@@ -273,6 +273,20 @@ def test_apt_daily_services_and_timers_enabled(host, service):
             assert s.is_enabled
 
 
+def test_apt_daily_timer_schedule(host):
+    if host.system_info.codename != "xenial":
+        c = host.run("systemctl show apt-daily.timer")
+        assert "TimersCalendar={ OnCalendar=*-*-* 00/3:00:00 ;" in c.stdout
+        assert "RandomizedDelayUSec=1h" in c.stdout
+
+
+def test_apt_daily_upgrade_timer_schedule(host):
+    if host.system_info.codename != "xenial":
+        c = host.run("systemctl show apt-daily-upgrade.timer")
+        assert "TimersCalendar={ OnCalendar=*-*-* 04:00:00 ;" in c.stdout
+        assert "RandomizedDelayUSec=1h" in c.stdout
+
+
 def test_reboot_required_cron(host):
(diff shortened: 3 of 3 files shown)


zenmonkeykstop:
I'm trying this out  applied manually as part of RC3 testing (don't wanna wait overnight to see if it's applied) and it does not appear to trigger updates or upgrades. I think `/etc/apt/apt.conf.d/20auto-upgrades` would have to change as well to change the `APT::Periodic::*` values to `"always"`.

rmol:
That's surprising. All it does is change the schedule of the timers. I don't know why it would require other configuration changes.

zenmonkeykstop:
https://wiki.debian.org/UnattendedUpgrades describes the timer format in a bit of detail but it's describing a case where you want to move the daily run from 6,1800 with a randomized delay of 12 hours to 1am precisely, and what actually happens is that the timer triggers apt-daily.service, which checks the delta between now and its last runtime against the interval defined in the  APT::Periodic::* values above, and only actually does something if  said delta is bigger than the interval.

So! in order to be able to trigger updates and upgrades (which are 2 separate timers) arbitrarily, you'd need to set those APT::Periodic:: values to "always", which appears to be a keyword meaning "always run when asked", which the developers probably had to add because "0" can't mean "0 interval" as it already means "don't ever run".

conorsch:
>  you'd need to set those APT::Periodic:: values to "always", which appears to be a keyword meaning "always run when asked"

@zenmonkeykstop Where are you seeing that "always" is a permissible value? For `APT::Periodic::Unattended-Upgrade` my understanding is that the value must be an integer, where 0 is disabled, and any positive int is the frequency, in days, that the update should be run.

zenmonkeykstop:
Docs are a little sparse, it's true. The apt.conf man page will redirect you to /usr/lib//apt/apt.systemd.dailywhich documents the N for N days or 0 for off behaviour. But if you grep for "always" in said script you will find the `check_stamp` function that does the actual interval checking, with logic to short-circuit the check if the interval is set to "always".

Most `apt.systemd.daily` operations will only run if the interval since they last ran is greater than the defined interval in the `APT::Periodic::` settings, so if you want to run them arbitrarily (say for testing purposes) you need to override that behaviour.

zenmonkeykstop:
On further review/discussion - we can get the desired test behaviour (rerunning `apt.systemd.daily` on demand) by removing timestamp files in `/var/lib/apt/periodic` - so actually yes, we don't need the changes in `/etc/apt/apt.conf.d/`, just the timer settings above. That'll get us a reliable run - my preference would be (reboot time - 1hr with a 30min fuzz) for the update and (reboot time - 30min with a 30min fuzz) for the upgrade, but do what works!

rmol:
closing in favor of #5855

PR #5874, opened 2021-03-16
Replace logo selection route with per-request global variable

rmol:
## Status

Ready for review

## Description of Changes

Getting the logo URL via the `/org-logo` route's redirect meant that the logo had to be retrieved with every page load, which of course over Tor can take several seconds. Selecting the logo at the start of each request and storing its URL in the Flask globals allows the static URL to be used in the templates, with the default caching of static assets, eliminating the extra requests.

## Testing

- `git checkout -b static-logo origin/static-logo`
- If you have a custom logo in your working copy at `securedrop/static/i/custom_logo.png`, remove it.
- `make dev`
- Verify that the default logo is shown on both source and journalist interfaces, served from its static URL, not from `/org-logo`.
- Note that the logo is cached like other static assets, and not retrieved on subsequent normal page loads.
- Visit the instance configuration page in the journalist interface, upload a custom logo, and check that it's used in both source and journalist interfaces.
- Make the same checks in an environment using Apache, like staging.

## Deployment

This does introduce the possibility of OSSEC alerts from the errors logged if no site logo is available, or if the static directory is missing.

Choose one of the following:

### DIFF
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 
 from datetime import datetime, timedelta
+from pathlib import Path
+
 from flask import (Flask, session, redirect, url_for, flash, g, request,
                    render_template)
 from flask_assets import Environment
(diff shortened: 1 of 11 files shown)


kushaldas (approved):
- [x]  Logo is static when using default logo
- [x] Custom logo is static

This is true for both `source` and `journalist` interfaces.  Approved

PR #5878, opened 2021-03-24
Generalize procurement of server OS information

rmol:
## Status

Ready for review

## Description of Changes

Not all Linux distributions have `/etc/lsb-release`. Debian 10, for instance, only has `/etc/os-release`, which means the `server_os` module could be broken in a dev environment outside of containers. This change uses the `lsb_release` command to obtain the operating system version.

It also uses functions instead of module attributes for the OS release and Xenial EOL date, in an attempt to improve patching in
tests. Manipulating the module attributes in tests was causing consistent flaky (oh, the irony) retries.

## Testing

Check out this branch, `make dev`, see that the source interface is available. Check http://localhost:8080/metadata and verify the `server_os` value is correct.

Change the `server_os.get_xenial_eol_date` function to return 2020-04-30 instead, reload the source interface, and confirm that it is now offline.

## Deployment

This does change the information used to determine whether the source interface should be disabled.

Choose one of the following:


securedrop/server_os.py
```diff
@@ -1,17 +1,30 @@
...
 
-with open("/etc/lsb-release", "r") as f:
-    installed_version = f.readlines()[1].split("=")[1].strip("\n")
+
+@functools.lru_cache()
+def get_xenial_eol_date() -> date:
+    return date(2021, 4, 30)
+
+
+@functools.lru_cache()
```
zenmonkeykstop, line 13: Nice.

securedrop/server_os.py
```diff
@@ -1,17 +1,30 @@
...
+def get_os_release() -> str:
+    return subprocess.run(
+        ["lsb_release", "--release", "--short"],
```
zenmonkeykstop, line 16: An addition to apparmor rules for Apache2 is required for this to pass in staging/prod methinks.

zenmonkeykstop (changes requested):

securedrop/server_os.py
rmol: good call, probably so, I'll verify and adjust.

zenmonkeykstop (approved):

PR #5879, opened 2021-03-25
Added mokutil check to detect SecureBoot status before installing

### LINKED ISSUE #5871: Check for Secure Boot setting during install process
## Description

Currently having Secure Boot enabled breaks the install process on first reboot after the grsec kernel is in place. While this is covered in docs, and we could look at signing kernels as per freedomofpress/kernel-builder#47 , there will probably be a lot of fresh installs happening in the immediate future, and checking early (using `mokutil` or similar) in the install playbook would catch this and prevent admin confusion and delay.

## User Research Evidence
Anecdotal feedback from support and forums.

## User Stories
As an admin, I want to be alerted to misconfigurations or incompatible BIOS settings early and in a transparent manner, rather than trying to debug opaque failures.

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #5871.

Adds a raw command check in the `prepare_servers` ansible role, to see if SecureBoot is enabled. If it is, the install exits early with an informative error. The check runs `mokutil --sb-state`, and fails iff the output includes `SecureBoot enabled`:
- `mokutil` will error out if `noefi` is set in kernel boot args (as it is for the current grsec kernel) - this is ignored and the check passes
- if mokutil isn't present (it is present for UEFI installs on 16.04.6 and 20.04) the check passes (because this is probably a Legacy boot setup)
- if the response is `SecureBoot disabled`, it obviously passes :)

## Testing

On prod hardware,
 - before installing, enable SecureBoot in the app server BIOS and disable it for the mon server
 - complete the OS install and initial setup/sdconfig using this branch
 - run `./securedrop-admin --force install` entering the server password when prompted
   -  [ ] verify that the secureboot check fails for the app server and passes for the mon server, and that the install playbook exits immediately after.
 - disable SecureBoot for the app server and rerun the `install` command
   - [ ] verify that the secureboot check passes for both servers and that the installation completes without issues.
 - after installation and `tailsconfig`, run `ssh mon sudo apt remove mokutil` to remove the utility from the mon server.
 - run `./securedrop-admin --force install` again
   - [ ] verify that the app server check fails with an `EFI variables are not supported` message
   - [ ] verify that the mon server check fails with `mokutil: not found`
   - [ ] verify that the failures are ignored and the installation completes without issues.
  

## Deployment

Deployed when workstation is updated. No implications for older systems IMO

Choose one of the following:

### DIFF
--- a/install_files/ansible-base/roles/prepare-servers/tasks/main.yml
+++ b/install_files/ansible-base/roles/prepare-servers/tasks/main.yml
@@ -21,6 +21,21 @@
       SecureDrop cannot be installed. For details, see
       https://github.com/freedomofpress/securedrop/issues/4058
 
+- name: Check SecureBoot status
+  raw: 'mokutil --sb-state'
+  ignore_errors: yes
+  register: _mokutil_results
+
+- name: Verify that SecureBoot is not enabled
+  assert:
+    that:
+      - "'SecureBoot enabled' not in _mokutil_results.stdout"
+      - "'SecureBoot enabled' not in _mokutil_results.stderr"
+    fail_msg: >-
+      SecureBoot is enabled. SecureDrop cannot be installed, as it uses a
+      custom kernel that is not signed. Please disable SecureBoot on the
+      target servers and try again.
+
 - name: Install python and packages required by installer
   raw: apt install -y python3 apt-transport-https dnsutils ubuntu-release-upgrader-core
   register: _apt_install_prereqs_results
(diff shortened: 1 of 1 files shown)


kushaldas:
As I mentioned in yesterday's standup, I don't have access to the prod hardware right now, so I will not be able to review this PR.

conorsch (approved):
I haven't tested on prod server hardware, but I did test on Focal Qubes HVMs. Added a commit to make sure that `mokutil` is present. As you say, the check passes when it's not (which is appropriate), but it seems reasonable to install the tool in order to verify the system state. 

Inside Qubes HVMs, I see `EFI variables are not supported on this system`, which still passes the check.

PR #5881, opened 2021-03-26
Correct untranslated messages

### LINKED ISSUE #5755: gettext calls containing formatted strings produce untranslated text
## Description

There are a few places where we do `gettext('something about {}'.format('blah'))` instead of `gettext('something about {}').format('blah')` and the misplaced `.format` call has `gettext` looking for the formatted string in the message catalogs. When it can't be found, English appears in the user interface.

## Steps to Reproduce

- `make dev`
- visit the JI admin interface, choose a language other than English, and add a user or edit `dellsberg`. 
- try to change the first name to something longer than 100 characters. 

## Expected Behavior

The flashed error message should be in the selected language.

## Actual Behavior

It's in English.

name-too-long-english

## Comments

There are at least these occurrences:

- https://github.com/freedomofpress/securedrop/blob/7f4076b36298475da15dde826e204310888b773d/securedrop/journalist_app/admin.py#L139
- https://github.com/freedomofpress/securedrop/blob/7f4076b36298475da15dde826e204310888b773d/securedrop/journalist_app/admin.py#L239
- https://github.com/freedomofpress/securedrop/blob/7f4076b36298475da15dde826e204310888b773d/securedrop/journalist_app/admin.py#L247
- https://github.com/freedomofpress/securedrop/blob/7f4076b36298475da15dde826e204310888b773d/securedrop/journalist_app/utils.py#L368

rmol:
## Status

Ready for review

## Description of Changes

Fixes #5755.

Corrects some misplaced parentheses in gettext/format calls. Marks some strings that were being presented untranslated.
    
Fixes locale/session handling on CSRF error. We were translating the CSRF error message, then clearing the session and showing the rest of the page in the default locale, almost certainly English. We need to reevaluate how we want to handle the locale when clearing the session, but for now, this just fixes the inconsistency.

Adds localization checks to basic source/journalist interface tests. We haven't really been testing translations. In the page layout tests, the content of strings is often only checked after making sure the page is *not* translated. This has led to us not catching text which didn't get marked for translation, thus being shown in English to users who are using a different locale.
    
Includes the following changes to make it easier to test that source strings are properly translated:
- a context manager which will xfail a test for a locale if any of the message IDs passed to it are not present in the locale's message catalog
- a pytest flaky filter which prevents retries of expected failures
- turning off flaky's reporting of successful tests
- functions to obtain test locales and counts to use in testing their pluralization rules, which can be used with pytest.parametrize to easily run a test in multiple locales, including fuzzing plural strings to check that translations are complete, while only trying as many variations as the rules need.
    
Adding locale parameterization to tests increases the overall time they take, of course: test_journalist.py went from around 2.5 minutes to 3.5 minutes.

### Application string changes

- `git checkout develop`
- `make dev`

#### Journalist interface

- Log in to the journalist interface as an admin.
- Edit user "dellsberg".
- Change the locale to "Français".
- Resetting the user's password ("Réinitialiser le mot de passe") should present a translated message containing the new password in a `<code>` tag, so that it is displayed in a fixed-width typeface. (This is just checking that the formatting wasn't broken by reorganizing the translation.)
- The following changes will all flash error messages in English:
  - Set the first name ("Prénom") to something longer than 100 characters.
  - Set the family name ("Nom de famille") to something longer than 100 characters.
  - Set the username ("Nom d’utilisateur") to a single character.
  - Set the username to "deleted".
- Return to the main admin and try to add a user with username "dellsberg". The error about the username conflict will be in English.
- In your SecureDrop repository root: `chmod 555 securedrop/static/i`
- Visit the instance configuration page and upload a new site logo. The error shown because the directory is not writable will be in English.
- Now check out this branch with `git checkout -b fix-5755 origin/fix-5755`.
- The French message catalog needs to be edited to translate the new messages. Replace `securedrop/translations/fr_FR/LC_MESSAGES/messages.po` with the test translations in this [gist](https://gist.github.com/rmol/8798da8997b7645b1112c5d433197884).
- Restart the dev server.
- Repeat all the tests after the password reset test, verifying that the error messages are now in French.
- Don't forget to restore the permissions on your image directory.

#### Source interface
- Keep using the dev server.
- This is another check that reorganized translation code hasn't broken an HTML message: running `curl -sH "X-tor2web: encrypted" 'http://localhost:8080?l=fr_FR' | grep Tor2Web` should produce:
```
<strong>AVERTISSEMENT</strong>&nbsp;Il semble que vous utilisez Tor2Web, qui ne garantit pas votre anonymat. &nbsp;<a href="/tor2web-warning">Pourquoi est-ce dangereux?</a></p>
```

## Deployment

This is mostly changes to text or tests. The string changes should correct several cases of English messages being shown to people who've chosen another locale. There are a couple of changes to messages that wind up in logs; the new messages record less information.

Choose one of the following:


securedrop/journalist_app/admin.py
```diff
@@ -52,8 +52,8 @@ def manage_config() -> Union[str, werkzeug.Response]:
                 f.save(custom_logo_filepath)
                 flash(gettext("Image updated."), "logo-success")
             except Exception:
-                flash("Unable to process the image file."
-                      " Try another one.", "logo-error")
+                # Translators: This error is shown when an uploaded image cannot be used.
```
sssoleileraaa, line 55: I think this is a good comment for devs as well as translators so perhaps remove `Translators:`

securedrop/journalist_app/admin.py
```diff
@@ -131,13 +131,16 @@ def add_user() -> Union[str, werkzeug.Response]:
                 form_valid = False
             except InvalidUsernameException as e:
                 form_valid = False
-                flash('Invalid username: ' + str(e), "error")
+                # Translators: Here, "{message}" explains the problem with the username.
```
sssoleileraaa, line 134: same for ^

securedrop/journalist_app/admin.py
```diff
@@ -236,15 +241,16 @@ def edit_user(user_id: int) -> Union[str, werkzeug.Response]:
                 Journalist.check_name_acceptable(first_name)
                 user.first_name = first_name
             except FirstOrLastNameError as e:
-                flash(gettext('Name not updated: {}'.format(e)), "error")
+                # Translators: Here, "{message}" explains the problem with the name.
```
sssoleileraaa, line 244: same for ^

securedrop/journalist_app/utils.py
```diff
@@ -437,11 +437,20 @@ def set_diceware_password(user: Journalist, password: Optional[str]) -> bool:
...
+    flash(
+        Markup(
+            "<p>{message} <span><code>{password}</code></span></p>".format(
+                message=Markup.escape(
+                    gettext(
+                        "Password updated. Don't forget to save it in your KeePassX database. "
+                        "New password:"
+                    )
+                ),
+                password=Markup.escape("" if password is None else password)
```
sssoleileraaa, line 449: i don't think password can ever be None after a password update, right? did you see it None is some case?

securedrop/models.py
```diff
@@ -380,11 +379,9 @@ def __init__(self, passphrase: str) -> None:
 
     def __str__(self) -> str:
         if self.passphrase_len > Journalist.MAX_PASSWORD_LEN:
-            return "Password too long (len={})".format(self.passphrase_len)
+            return "Password is too long."
         if self.passphrase_len < Journalist.MIN_PASSWORD_LEN:
-            return "Password needs to be at least {} characters".format(
-                Journalist.MIN_PASSWORD_LEN
-            )
+            return "Password is too short."
```
sssoleileraaa, line 384: Why remove the password length information? Is it already displayed somewhere else?

securedrop/models.py
```diff
@@ -380,11 +379,9 @@ def __init__(self, passphrase: str) -> None:
 
     def __str__(self) -> str:
         if self.passphrase_len > Journalist.MAX_PASSWORD_LEN:
-            return "Password too long (len={})".format(self.passphrase_len)
+            return "Password is too long."
```
sssoleileraaa, line 382: Seems like it might be useful to display the range for the password length requirements, something like `Password needs to be between {Journalist.MAX_PASSWORD_LEN} and {Journalist.MIN_PASSWORD_LEN} characters.`.

securedrop/models.py
```diff
@@ -496,18 +493,25 @@ def set_name(self, first_name: Optional[str], last_name: Optional[str]) -> None:
     def check_username_acceptable(cls, username: str) -> None:
         if len(username) < cls.MIN_USERNAME_LEN:
             raise InvalidUsernameException(
-                        'Username "{}" must be at least {} characters long.'
-                        .format(username, cls.MIN_USERNAME_LEN))
+                ngettext(
+                    'Must be at least one character long.',
```
sssoleileraaa: Since you can update `Frist name`, `Last name`, and `Username` in the same form, it would be better to specify which field is not within the required character range.

securedrop/source_app/__init__.py
```diff
@@ -127,14 +127,21 @@ def check_tor2web() -> None:
...
-                'You appear to be using Tor2Web. '
-                'This <strong>&nbsp;does not&nbsp;</strong> '
-                'provide anonymity. '
-                '<a href="{url}">Why is this dangerous?</a>')
-                .format(url=url_for('info.tor2web_warning'))),
-                "banner-warning")
+            flash(
+                Markup(
+                    '<strong>{}</strong>&nbsp;{}&nbsp;<a href="{}">{}</a>'.format(
+                        escape(gettext("WARNING!")),
```
sssoleileraaa: the all-caps exclamation mark does make it more likely to grab my attention. not sure what the standard is or if we really need one for the first word and formatting used to capture people's attention.

sssoleileraaa (commented):
this looks great! i left some comments that i'll provide now. i still need to run through the test plan before i can approve.

securedrop/journalist_app/admin.py
rmol: The `Translators:` prefix is necessary for the comment to be carried over to the `.po` files, so it will appear in the Weblate UI.

securedrop/models.py
rmol: As far as I know, these messages aren't ever shown to a person, just logged. I started changing them assuming they were visible, and surely were being or should be translated somewhere, but I don't think that's the case. I kept the changes because the messages were inconsistent, in that the offending length wasn't being logged for passwords that were too short, and it's not useful information -- an admin isn't going to do anything about a journalist entering a password that's too long or too short. I haven't set up prod to see if these errors surface in OSSEC alerts, but if they do, recording the length would result in more unique messages, contributing to alert fatigue.

securedrop/models.py
rmol: This message is included in flashed messages that do indicate which field had the problem.

securedrop/source_app/__init__.py
rmol: Oops, it was originally all caps, but had a colon instead of an exclamation mark. That is a little dramatic. I'll revert it to tone it down.

securedrop/journalist_app/utils.py
rmol: In the signature of `set_diceware_password`, the `password` argument is typed as `Optional[str]`, because it's populated with `request.form.get`, if I'm reading it correctly. Using just `Markup.escape(password)` here, `mypy` complains:
```
securedrop/journalist_app/utils.py:449: error: Argument 1 to "escape" of "Markup" has incompatible type "Optional[str]"; expected "str"
```

securedrop/models.py
sssoleileraaa: if it's not displayed to the user and doesn't need to be translated, then i think the changes for consistency sake are good

sssoleileraaa: ...and could potentially help with alert fatigue, yes, yes, good point

sssoleileraaa:
@rmol this message "Name not updated: Name too long" should be in French, right?

Screenshot from 2021-03-31 18-22-45

rmol:
Yes it should. I left a crucial step out of the test plan. The French message catalog needs to be edited to translate the new messages, and the dev server restarted. I've updated the test plan, sorry about that.

sssoleileraaa (approved):
works great with the [gist](https://gist.github.com/rmol/8798da8997b7645b1112c5d433197884) you provided for French translations:
Screenshot from 2021-04-01 08-56-16

PR #5885, opened 2021-03-31
Use GitHub group for code ownership

### LINKED ISSUE #5865: Refresh CODEOWNERS assignments
As the dust settles on the SecureDrop 1.8.0 release, now may be a good time to refresh CODEOWNERS files where they are used, so we can sure that default reviewer suggestions are complete and up-to-date. I was only able to find them here:

https://github.com/freedomofpress/securedrop/blob/develop/.github/CODEOWNERS
https://github.com/freedomofpress/securedrop-client/blob/main/.github/CODEOWNERS

Alternatively, if we don't think these suggestions are useful, we could consider retiring their usage. See https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners for full docs.

eloquence:
## Status

Ready for review

## Description of Changes

Resolves #5865 (unless we want to immediately apply this logic across all SD repos)

## Checklist / testing

Pure GitHub-level config change: 
- Contents should be identical to https://github.com/freedomofpress/securedrop-client/blob/main/.github/CODEOWNERS
- Post-merge, we can verify that the maintainers group is tagged for review on new PRs

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,18 +1,1 @@
-# application code
-/securedrop/ @redshiftzero @kushaldas
-
-# updater GUI
-/journalist_gui/ @redshiftzero @kushaldas
-
-# admin CLI
-/admin/ @redshiftzero @kushaldas @emkll
-
-# docs
-/docs/ @zenmonkeykstop @redshiftzero @kushaldas @emkll
-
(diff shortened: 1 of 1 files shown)


sssoleileraaa (approved):
lgtm

PR #5893, opened 2021-04-08
Acknowledge PyUp sponsorship

eloquence:
## Status

Ready for review

## Description of Changes

pyup.io has agreed to sponsor a paid subscription to their security database, which we are acknowledging.


conorsch (approved):
Looks good! I've confirmed API access is working, but it's not actually running on our repos yet. @emkll Let's have a chat about how best to do that.

Issue #5869, opened 2021-03-15
Review all pending Weblate comments/suggestions


eloquence:
Prior to the next release, we should make sure we review all pending Weblate comments and suggestions. This includes the following issues that are already tracked:

- Consider using sentence case instead of title case in UI elements - #5828
- "Say replies are saved, not stored" - #5821

eloquence:
I've taken a spin through historical translator feedback, and the two remaining issues above are already tracked; closing this issue.

PR #5895, opened 2021-04-09
Wordlist removals and translation updates for next SecureDrop release

eloquence:
## Status

Ready for review. See previously: #5357, #1546; 366c657b5f79b3dae510404ea5d6fed6adb0f6ca.

## Description of Changes

- Removes potentially offensive words (alone or in combination) from the SecureDrop wordlists. Among other things, SecureDrop uses adjectives and nouns to generate journalist designations, and while we cannot avoid all potentially offensive combinations, we can exclude some words that are almost always inappropriate in journalist designations or passphrases.

- Addresses translation feedback by pappasadrian and KwadroNaut from https://weblate.securedrop.org/translate/securedrop/securedrop/en/?checksum=e3e9531df602610b

### DIFF
--- a/securedrop/dictionaries/adjectives.txt
+++ b/securedrop/dictionaries/adjectives.txt
@@ -1110,7 +1110,6 @@ casebook
 cased
 cast-iron
 cast-off
-castrated
 casual
 catalytic
 catastrophic
@@ -3396,7 +3395,6 @@ histological
 historic
 historical
 histrionic
-hit-and-run
 hitless
 hoary
 holey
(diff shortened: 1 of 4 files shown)


eloquence:
Since it's a small change, I've pulled in one low-hanging fruit string improvement as well, suggested in  https://weblate.securedrop.org/translate/securedrop/securedrop/en/?checksum=e3e9531df602610b, as part of an overall review of Weblate comments (#5869).

zenmonkeykstop (approved):
LGTM based on visual review, CI results.

PR #5888, opened 2021-04-06
Updated multiple build and test dependencies

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Updates multiple dependencies No securedrop-app-code dependency changes (with the exception of setuptools itsefl) so no diff reviews performed.

* pip 19.3.1 -> 21.1.1
* pip-tools 4.5.1 -> 6.1.0
* setuptools 46.0.0 -> 56.0.0
*  setuptools-scm 5.0.2 (no hash check) -> 6.0.1 (w/ hash check)
*  pillow 8.1.1 -> 8.2.0

## Testing

- [ ] CI is green
- [ ] No new non-test dependencies introduced.

PR #5903, opened 2021-04-16
Add hash check for setuptools-scm

EdwardPrentice:
Mandates that the hash for setuptools-scm at version 5.0.2 is correct to avoid tampering

### DIFF
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/translations.yml
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/translations.yml
@@ -4,7 +4,7 @@
   shell: >
     set -e &&
     python3 -m venv /tmp/securedrop-app-code-i18n-ve &&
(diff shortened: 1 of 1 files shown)


zenmonkeykstop:
Hi @EdwardPrentice apologies for the delayed response. This is a definite improvement but it may not get used in its current form, as the next release will be Focal-only and hence >Python 3.6, so we can update for that.

EdwardPrentice:
Sounds good! Gives an idea of how to do it at least!

conorsch:
Thank you, @EdwardPrentice, for this PR. We just merged https://github.com/freedomofpress/securedrop/pull/5888 which I believe resolves the concern you raised here—specifically https://github.com/freedomofpress/securedrop/pull/5888/files#diff-8be480111ec0858827980d7f974731ea7c28f24065f7061558ea0464beb4b4d0L7-R7

In that PR, @zenmonkeykstop opted to create a new requirements file, but your trick with the process substitution was clever! Hopefully we'll be able to pare these requirements files down in the near future. 

I'm closing this PR as the underlying issue has been resolved, but don't hesitate to re-open if you disagree. Hope to see you around soon!

PR #5905, opened 2021-04-16
Use definite article for single file error case; improve functional test output

eloquence:
## Status

Ready for review
## Description of Changes

- Distinguishes single/multiple file error case. See @deeplow's comment in https://weblate.securedrop.org/translate/securedrop/securedrop/en/?checksum=aec87110942e8803#comments
- Registers `pytest` assertion rewriting for our helper functions so test failures show helpful string comparison when an assertion fails.

### Error message behavior
On this PR's branch:
1. Spin up Docker env
2. Remove all files from disk in the container's `/var/lib/securedrop/store` directory
3. Attempt to download single file by clicking its filename directly in the Journalist Interface
4. - [ ] Observe that single file error message is displayed
5. Attempt to download single file via the checkbox interface
6. - [ ] Observe that single file error message is displayed
7. Attempt to download multiple files
8. - [ ] Observe that multiple file error message is displayed
### Improved assertion output
1. Break one of the tests like so:
```patch
diff --git a/securedrop/tests/functional/journalist_navigation_steps.py b/securedrop/tests/functional/journalist_navigation_steps.py
index 3f3813a7d..da8e6d825 100644
--- a/securedrop/tests/functional/journalist_navigation_steps.py
+++ b/securedrop/tests/functional/journalist_navigation_steps.py
@@ -1189,7 +1189,7 @@ class JournalistNavigationStepsMixin:
         if single_file:
             error_msg = (
                 "Your download failed because the file could not be found. An admin can find "
-                + "more information in the system and monitoring logs."
+                + "more information in the system and monitoring logs. XYZZY"
             )
         else:
             error_msg = (
```

2. Run it via `securedrop/bin/dev-shell bin/run-test -s -v tests/functional/test_journalist.py -k test_download_source_unread^`
3. - [ ] Observe that the assertion output includes a diff, unlike the output in [previous test failures](https://app.circleci.com/pipelines/github/freedomofpress/securedrop/2338/workflows/da8ee35e-3cd1-4c3e-8fbb-ca9c0a481e6e/jobs/53730)

### DIFF
--- a/securedrop/journalist_app/col.py
+++ b/securedrop/journalist_app/col.py
@@ -116,7 +116,7 @@ def make_blueprint(config: SDConfig) -> Blueprint:
         if not Path(file).is_file():
             flash(
                 gettext(
-                    "Your download failed because a file could not be found. An admin can find "
+                    "Your download failed because the file could not be found. An admin can find "
                     + "more information in the system and monitoring logs."
                 ),
                 "error"
--- a/securedrop/journalist_app/utils.py
+++ b/securedrop/journalist_app/utils.py
@@ -205,9 +205,12 @@ def download(
         zf = current_app.storage.get_bulk_archive(submissions, zip_directory=zip_basename)
     except FileNotFoundError:
         flash(
-            gettext(
+            ngettext(
+                "Your download failed because the file could not be found. An admin can find "
+                + "more information in the system and monitoring logs.",
                 "Your download failed because a file could not be found. An admin can find "
-                + "more information in the system and monitoring logs."
+                + "more information in the system and monitoring logs.",
+                len(submissions)
             ),
             "error"
         )
--- a/securedrop/tests/__init__.py
+++ b/securedrop/tests/__init__.py
@@ -1,8 +1,13 @@
 # -*- coding: utf-8 -*-
 from os.path import abspath, dirname, join, realpath
+import pytest
 import sys
 
 # The tests directory should be adjacent to the securedrop directory. By adding
 # the securedrop directory to sys.path here, all test modules are able to
 # directly import modules in the securedrop directory.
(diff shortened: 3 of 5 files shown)


eloquence:
(CI's probably going to fail, will let it do its thing and then update affected tests)

rmol (dismissed):
If we wanted to show the single file error message in the case where one file is selected and "Download Selected" is clicked, we could check the length of the list of things to delete [here](https://github.com/freedomofpress/securedrop/blob/826c178f3d9e59b829c42f6afe619ee23fae45bd/securedrop/journalist_app/utils.py#L207).

I'm approving as is, and will merge unless you want to try that. The test plan went fine:

1. Spin up Docker env
2. Remove all files from disk in the container's `/var/lib/securedrop/store` directory
3. Attempt to download single file
4. - [x] Observe that single file error message is displayed
5. Attempt to download multiple files
6. - [x] Observe that multiple file error message is displayed

eloquence:
That would be nicer, I'll take another look, thanks @rmol

deeplow:
Wow. It is super interesting how little one realizes as a translator the engineering effort & implications of what looked like a simple thing.

Thank you for addressing this so quickly and keep up the good work!!

eloquence:
Thanks for the suggestion @rmol. This is now done and the correct message should be shown regardless of download method.

As part of looking at functional test output, I got frustrated that the `assert` failures didn't include diffs, and added assert rewrite logic which should fix this at least for the functional test helpers.

`staging-test-*` and `translation-tests` have been balking at me, but I'm not sure yet if this is related to anything in this PR.

eloquence:
Looks like it was just flaky tests, as usual :/

rmol (approved):
Looks good. Tested single/multiple downloads, and verified the pytest assert rewriting in support functions, which is a nice enhancement.

Issue #5914, opened 2021-04-22
Upgrade Ansible to latest stable


conorsch:
## Description
The Admin Workstation currently runs Ansible v2.9.7. Soon enough, we'll need to bump to the already released 2.10.x series. That's a difficult jump, and violates expectations about semver. 

## Example output for complicated upgrade
See the following failure message about a naive upgrade from 2.9.7 -> 2.10.7

<details>
<summary>
pip error output
</summary>

```
$ ansible --version
ansible 2.9.7
  config file = /home/user/securedrop/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/.virtualenvs/sd/lib/python3.7/site-packages/ansible
  executable location = /home/user/.virtualenvs/sd/bin/ansible
  python version = 3.7.3 (default, Jul 25 2020, 13:03:44) [GCC 8.3.0]
[user@sd-dev:~/securedrop] [sd] develop+* 6h15m51s ± 
 $ pip install 'ansible>2.10,<2.11' 
Collecting ansible<2.11,>2.10
  Downloading https://files.pythonhosted.org/packages/ba/22/7b58a8ba8e43159dc5cb32d97dd50e2b70b016585dbb188e9f2b61dac1e2/ansible-2.10.7.tar.gz (29.9MB)
     |████████████████████████████████| 29.9MB 561kB/s 
    ERROR: Command errored out with exit status 1:
     command: /home/user/.virtualenvs/sd/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-312we13s/ansible/setup.py'"'"'; __file__='"'"'/tmp/pip-install-312we13s/ansible/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-install-312we13s/ansible/pip-egg-info
         cwd: /tmp/pip-install-312we13s/ansible/
    Complete output (31 lines):
    
    
                ### ERROR ###
    
                Upgrading directly from ansible-2.9 or less to ansible-2.10 or greater with pip is
                known to cause problems.  Please uninstall the old version found at:
    
[...]

PR #5911, opened 2021-04-21
Removes Xenial related parts from codebase

kushaldas:
## Status

Ready for review.

## Description of Changes

Fixes #5725 

Removes major `Xenial` changes from the codebase.

## Testing

- [ ] CI should be green
- [ ] visual inspection of the change

## Deployment

Any special considerations for deployment? Consider both:

1. Upgrading existing production instances.
2. New installs.

### If you made changes to the server application code:

- [ ] Linting (`make lint`) and tests (`make test`) pass in the development container

### If you made changes to `securedrop-admin`:

- [ ] Linting and tests (`make -C admin test`) pass in the admin development container

### If you made changes to the system configuration:

- [ ] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass

### If you added or removed a file deployed with the application:

- [ ] I have updated AppArmor rules to include the change

### If you made non-trivial code changes:

- [ ] I have written a test plan and validated it for this PR

Choose one of the following:

- [ ] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-docs) for these changes, or will do so later
- [ ] I would appreciate help with the documentation
- [ ] These changes do not require documentation

### If you added or updated a code dependency:

Choose one of the following:

- [ ] I have performed a diff review and pasted the contents to [the packaging wiki](https://github.com/freedomofpress/securedrop-debian-packaging/wiki)
- [ ] I would like someone else to do the diff review

PR #5917, opened 2021-04-30
Fix Ansible deprecation warning when building deb packages

gonzalo-bulnes:
## Status

Ready for review

## Description of Changes

Applies the recommendation to fix a deprecation warning that was printed (18 times) when `make build-debs-focal`.

```txt
[DEPRECATION WARNING]: Invoking "apt" only once while using a loop via
squash_actions is deprecated. Instead of using a loop to supply multiple items
and specifying `name: "{{ item }}"`, please use `name: '{{
build_ossec_deb_pkg_dependencies }}'` and remove the loop. This feature will be
 removed in version 2.11. Deprecation warnings can be disabled by setting
deprecation_warnings=False in ansible.cfg.
```

Other tasks in the same file already use that syntax.
I thought I'd compare the output of the command with and without the change, but the Ansible tasks weren't executed in the same order, so the diff is large. I can go through it if deemed useful, though, I kept for log files, just let me know.

Loosely related to https://github.com/freedomofpress/securedrop/issues/5914 (once more, 2.11 won't be a semver minor update!)

## Testing

1. Run `make build-debs-focal`
2. Verify whatever you'd normally expect after running it. (Sorry, that's not super helpful!)

## Deployment

Because [other][1] [tasks][2] already use this syntax, and it is suggested by the deprecation warning, I don't have any reason to think that this change would change the resulting **deb** packages. Maybe that's worth a thought anyway before deploying.

  [1]: https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/common/tasks/install_packages.yml#L3-L4
  [2]: https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/common/tasks/remove_unused_packages.yml#L3-L4

- [ ] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass :warning: _I am in the process of building staging. I can only tell that the first step, `make build-debs-focal` exits successfully._

Choose one of the following:

Choose one of the following:

### DIFF
--- a/install_files/ansible-base/roles/build-ossec-deb-pkg/tasks/main.yml
+++ b/install_files/ansible-base/roles/build-ossec-deb-pkg/tasks/main.yml
@@ -55,11 +55,10 @@
 
 - name: Install apt dependencies for building OSSEC packages.
   apt:
-    name: "{{ item }}"
+    name: "{{ build_ossec_deb_pkg_dependencies }}"
     state: present
     update_cache: yes
     cache_valid_time: 3600
-  with_items: "{{ build_ossec_deb_pkg_dependencies }}"
 
 - name: Extract OSSEC source tarball.
   unarchive:


conorsch (approved):
@gonzalo-bulnes Thank you for taking the time, this is a very pleasant change. 

To test, I rebased locally (to include the substantial changes from #5911), and confirmed that the packages built correctly and all deb tests passed. And, of course, the warning is no longer emitted. :smiley:

Issue #5920, opened 2021-05-01
xenial app tests are failing


eloquence:
Many xenial app tests are failing on new PRs, presumably because the Source Interface is disabled after April 30, so none of the expected behavior is found. See example output here: https://circleci.com/api/v1.1/project/github/freedomofpress/securedrop/53822/output/109/0?file=true&allocation-id=608cb3d9cdd8e709525e7fdf-0-build%2F30D5FC47

eloquence:
Looks like the Xenial `lint` job is also failing now, at least after repeated kicks it has kept getting stalled on the "Save Docker image layer cache" step: https://app.circleci.com/pipelines/github/freedomofpress/securedrop/2350/workflows/04e149aa-04b0-4532-a03a-1936d01dbdfc/jobs/53826

eloquence:
This was fixed via #5911.

PR #5921, opened 2021-05-01
Bump Firefox ESR and Geckodriver versions

eloquence:
## Status

Ready for review

## Description

My first guess on why app tests are failing was that there's a version discrepancy somewhere. We're running behind a bit, but the real reason for the test failures appears to be the post-April 30 Xenial EOL maintenance mode kicking in (#5920). 

Still, why not give FF and Geckodriver a bump while we're here? ESR version corresponds to the one the latest Tor Browser is based on (https://blog.torproject.org/new-release-tor-browser-10016), Geckodriver is simply latest (https://github.com/mozilla/geckodriver/releases).

## Testing

- app-tests should still be happy


emkll (approved):
Full translations tests are passing in https://app.circleci.com/pipelines/github/freedomofpress/securedrop/2369/workflows/27bf582f-c280-440b-8c57-8a1465345d6e/jobs/53926

PR #5924, opened 2021-05-04
Use actual hostname in Postfix configuration

### LINKED ISSUE #5916: Focal OSSEC alerts not delivered when SMTP server requires that the From: address match the SASL-authenticated user
## Description

OSSEC alerts are sent via email from the monitor server using postfix and an external SMTP server. By default under Xenial, the sender email address was `ossec@ossec.server`. In cases where the SMTP server requires that the email sender matches the authenticated user, an additional `site-specific` variable, `ossec_from_address` must be set - it's used to map the local address to  the variable value.

This is failing under Focal. Even with `ossec_from_address` set, emails are being rejected due to a mismatch. It appears that Focal is using `ossec@mon` as the sender instead.

## Steps to Reproduce

- Set up a prod install using an SMTP server as described above, and setting `ossec_from_address` in the `site-specific` file before running the install playbook
- observe mail log output in `/var/log/mail.log` on `mon`

## Expected Behavior
- Mail delivered as email address specified without issue

## Actual Behavior
- Mail rejected

## Comments

rmol:
## Status

Ready for review

## Description of Changes

Fixes #5916.

The legacy "ossec.server" value of the Postfix `myhostname` setting does not match our current recommended mon server hostname ("mon"), so prevents correct rewriting of the from address when sending to the SMTP relay.

This changes the `postfix_hostname` Ansible variable to use the value of `monitor_hostname`. The `postfix_from_address` still uses `ossec_from_address` if configured, but now defaults to "sasl_username@sasl_domain" if `sasl_domain` is supplied, enabling another chance for a valid rewrite of "ossec@mon". If neither `ossec_from_address` nor `sasl_domain` is specified, `postfix_from_address` will be empty, and in that case the playbook will now remove an existing `/etc/postfix/generic` database.

### Playbook on hardware or prod VMs

- Check out this branch: `git checkout -b fix-postfix-hostname origin/fix-postfix-hostname`

- Follow #5916 STR. The `generic` database should contain the proper rewrite.

- Remove `ossec_from_address` from `site-specific`, ensure that it contains a valid `sasl_domain`, and rerun `securedrop-admin install`, or to save time, activate the admin virtualenv, `cd install_files/ansible-base` and run `ansible-playbook securedrop-prod.yml -t postfix`. The `generic` database should contain a valid rewrite from `ossec@mon` to `sasl_username@sasl_domain`, and if your SMTP relay configuration is valid, mail should flow.
- Edit `site-specific` to set `sasl_domain: ''`, then apply the configuration as above. The `generic` database should be removed.

### `postinst`

- `git checkout develop && make build-debs && make staging`
- on mon server: `grep myhostname /etc/postfix/main.cf` should return `myhostname = ossec.server`
- `/etc/postfix/generic` should not exist
- back in dev env: `mv build/focal build/focal-develop`
- `git checkout fix-postfix-hostname`
- `make build-debs`
- scp the `securedrop-ossec-server` deb to the mon server
- on mon server: `sudo apt install --reinstall /path/to/securedrop-ossec-server-3.6.0+1.9.0~rc1+focal-amd64.deb`
- `grep myhostname /etc/postfix/main.cf` should now return `myhostname = mon-staging`
- `/etc/postfix/generic` should contain `ossec@mon-staging test@ossec.test`

## Deployment

This should ensure valid from addresses in outbound OSSEC alerts. 

Choose one of the following:

### DIFF
--- a/install_files/ansible-base/roles/postfix/defaults/main.yml
+++ b/install_files/ansible-base/roles/postfix/defaults/main.yml
@@ -1,8 +1,10 @@
 ---
 # Email address listed in the FROM line when sending OSSEc email alerts.
 # Some mail servers require that this match the account that authenticated
-# to send mail. Using the `ossec_from_address` for backwards-compatibility.
-postfix_from_address: "{{ ossec_from_address|default('') }}"
+# to send mail. Using the `ossec_from_address` for
+# backwards-compatibility.
+sasl_id: "{% if sasl_domain %}{{ sasl_username }}@{{ sasl_domain }}{% endif %}"
+postfix_from_address: "{{ ossec_from_address|default(sasl_id) }}"
 
 # Apt dependencies for the ossec server package
 postfix_dependencies:
@@ -11,7 +13,7 @@ postfix_dependencies:
   - mailutils
 
 # Configuration info for procmail and postfix
-postfix_hostname: ossec.server
+postfix_hostname: "{{ monitor_hostname }}"
 
 # Whether to enable Postfix for sending mail. Required in prod,
 # but unnecessary in staging contexts, where SASL authentication
--- a/install_files/ansible-base/roles/postfix/tasks/install_postfix.yml
+++ b/install_files/ansible-base/roles/postfix/tasks/install_postfix.yml
@@ -15,12 +15,23 @@
   tags:
     - postfix
 
-- name: Create mapping for outbound address.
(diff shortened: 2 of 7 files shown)


eloquence:
@rmol Looks like the `staging-test-with-rebase` failure is actually Postfix related, could you take a look?

zenmonkeykstop (approved):
Reviewed fresh/subsequent install playbook scenarios, all good.

Reviewed postint, all good.

PR #5927, opened 2021-05-05
fix: Application Server submissions checker runs 60 times a day

### LINKED ISSUE #5847: The Application Server submissions checker runs 60 times a day instead of once.
## Description

The crontab entry for the script checking for new submissions (for use in journalist alerts) (with a reboot time of 4am) looks like:

```
* 3 * * * /var/www/securedrop/manage.py were-there-submissions-today
```
This will kick off the script every minute from 3am to 4am. If the intent is to trigger it once, it should be more like

```
* 3 0 * * /var/www/securedrop/manage.py were-there-submissions-today
```

## Steps to Reproduce
Check /var/log/syslog on a production system.

## Expected Behavior
There's one log line at 3am for the script.

## Actual Behavior
There's are multiple log lines from 3am  to reboot for the script.

## Comments
The ansible task setting it up needs a "minutes" value, which should just be 0

(This does not look like an issue introduced in 1.8.0. The code in question hasn't been touched since May 2018.)

prateekj117:
## Status

Ready for review

## Description of Changes
Application Server submissions checker runs 60 times a day because the crontab entry has `*` in minutes position. So the submission checker runs from 3, every minute till 4 which is `daily_reboot_time`.  

Fixes #5847 

Add minutes value in crontab to be 0.

## Testing

Reference documentation is [here](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/cron_module.html).

- [X] Linting (`make lint`) and tests (`make test`) pass in the development container

Choose one of the following:

### DIFF
--- a/install_files/ansible-base/roles/app/tasks/setup_cron.yml
+++ b/install_files/ansible-base/roles/app/tasks/setup_cron.yml
@@ -20,6 +20,7 @@
   cron:
     name: Update the number of submissions in the past 24h
     job: "{{ securedrop_code }}/manage.py were-there-submissions-today"
+    minute: "0"
     # 0 -> 23, 1 -> 0, 2 -> 1, ... 23 -> 22
     hour: "{{ (daily_reboot_time + 23) % 24 }}"
   tags:


rmol (approved):
I tested in a production setup and confirmed that this change corrects the crontab. Thanks @prateekj117.

Issue #5933, opened 2021-05-07
Bump 2FA secret bit length from 80 bits to 160bits as recommended by RFC4226


evilaliv3:
While evaluating a patch for https://github.com/freedomofpress/securedrop/issues/5613 i noticed that pyotp seems to be still generating secrets of 80 bits length that is actually discouraged by [RFC4226](https://tools.ietf.org/html/rfc4226):

```R6 - The algorithm MUST use a strong shared secret.  The length of the shared secret MUST be at least 128 bits. This document RECOMMENDs a shared secret length of 160 bits.```

The easily improve this I suggest we could just replace any call to pyotp.random_base32() in https://github.com/freedomofpress/securedrop/blob/fd2f5f86432005bdfc3c59119da98e442b7e4475/securedrop/models.py  with a call to a simple function:

```python
def generate_otp_secret():
    return base64.b32encode(os.urandom(20))
```

zenmonkeykstop:
This works with existing data (with 16-char OTP secrets) because the underlying db ignores the string lengths imposed by `models.py`, but it would probably be better to:

- use pyotp >= 2.6.0, which enforces a 32-char minimum
- update the db model to explicitly set that field to 32 chars (more for consistency's sake than anything else)

With either changes, existing 16-char secrets will still work. Any new secrets, either new accounts or otp refreshes, will be 32 chars long.

I'm not aware of any potential attacks that would require an enforced update for existing users, so haven't really looked at what would be required for that. It would be straightforward to detect a short secret and throw up a banner or something prompting the user to update. Bumping them straight to the account screen on login until they do is also a (less-friendly) option. I also haven't looked closely at implications for HOTP auth, but the secret key there is 80 bits and it's not immediately apparent to me how that can be changed.

eloquence:
See also #5613, which proposes removal of the `pyotp` dependency.

PR #5934, opened 2021-05-07
Bump 2FA secret bit length from 80 to 160 bits as recommended by RFC4226

evilaliv3:
Implement fix for ticket https://github.com/freedomofpress/securedrop/issues/5933

### DIFF
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -35,8 +35,8 @@ if os.environ.get('SECUREDROP_ENV') == 'test':
 ARGON2_PARAMS = dict(memory_cost=2**16, rounds=4, parallelism=2)
 
 
-def generate_otp_secret() -> bytes:
-    return base64.b32encode(os.urandom(20))
+def generate_otp_secret() -> str:
+    return base64.b32encode(os.urandom(20)).decode()
 
 
 def get_one_or_else(query: Query,
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -407,7 +407,7 @@ class Journalist(db.Model):
     is_admin = Column(Boolean)  # type: Column[Optional[bool]]
     session_nonce = Column(Integer, nullable=False, default=0)
 
-    otp_secret = Column(String(16), default=generate_otp_secret)
+    otp_secret = Column(String(32), default=generate_otp_secret)
     is_totp = Column(Boolean, default=True)  # type: Column[Optional[bool]]
     hotp_counter = Column(Integer, default=0)  # type: Column[Optional[int]]
     last_token = Column(String(6))
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -35,7 +35,7 @@ if os.environ.get('SECUREDROP_ENV') == 'test':
 ARGON2_PARAMS = dict(memory_cost=2**16, rounds=4, parallelism=2)
 
 
-def generate_otp_secret():
+def generate_otp_secret() -> bytes:
     return base64.b32encode(os.urandom(20))
 
 
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -35,6 +35,10 @@ if os.environ.get('SECUREDROP_ENV') == 'test':
 ARGON2_PARAMS = dict(memory_cost=2**16, rounds=4, parallelism=2)
 
 
+def generate_otp_secret():
+    return base64.b32encode(os.urandom(20))
+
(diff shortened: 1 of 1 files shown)


eloquence:
Hi @evilaliv3, thanks for working on this! The lint failure is real due to missing type annotations - would you be comfortable adding those, or would you like us to do so?

securedrop/models.py
```diff
@@ -403,7 +407,7 @@ class Journalist(db.Model):
     is_admin = Column(Boolean)  # type: Column[Optional[bool]]
     session_nonce = Column(Integer, nullable=False, default=0)
 
-    otp_secret = Column(String(16), default=pyotp.random_base32)
```
zenmonkeykstop, line 406: latest version of pyotp (2.6.0) now defaults to a 32-char base32 string from random_base32(). This should definitely be updated but it might make sense to update the dependency and the relevant references in the codebase. Any thoughts on that approach @evilaliv3 ?

zenmonkeykstop:
(moved comment to issue: https://github.com/freedomofpress/securedrop/issues/5933#issuecomment-844473763 )

securedrop/models.py
evilaliv3: Thank you for noticing this @zenmonkeykstop , you are right the size of this string should be changed accordingly.

I actually notice to have to do this on our codebase as well and as a fun fact SQLITE seems to ignore the size limit (probably due to default allogators on Strings that always allocate more than just 16 chars)

evilaliv3:
@zenmonkeykstop  @eloquence : Either the two options are fine for me.

Among the two options I support dropping the pyotp dependency for the benefit of limiting application dependencies.

In any case I've updated the pull request to reflect your pull request review and advice! :)

securedrop/models.py
zenmonkeykstop: yup varchars are humongous, had some "fun" last night writing db migration tests for what is effectively a no-op.

install_files/ansible-base/roles/restore/tasks/perform_restore.yml
```diff
@@ -23,8 +23,10 @@
   connection: local
   become: no
   unarchive:
+    # Avoid creating an unnecessary local copy
+    remote_src: yes
     dest: "{{ torrc_check_dir.path }}/backup/"
-    src: "{{ restore_file }}"
+    src: "{{playbook_dir}}/{{ restore_file }}"
```
zenmonkeykstop: this is unrelated to the PR, I wonder if a rebase went awry somewhere. will take a look.

securedrop/models.py
```diff
@@ -35,6 +35,10 @@
 ARGON2_PARAMS = dict(memory_cost=2**16, rounds=4, parallelism=2)
 
 
+def generate_otp_secret() -> bytes:
```
zenmonkeykstop: Type checking is failing, as `bytes` are specified but a `str` is expected in the `models.py` call.

zenmonkeykstop (changes requested):

securedrop/models.py
evilaliv3: It should be fixed now

install_files/ansible-base/roles/restore/tasks/perform_restore.yml
evilaliv3: Fixed preparing the pull request for integration on current develop branch

securedrop/models.py
```diff
@@ -35,6 +35,10 @@
 ARGON2_PARAMS = dict(memory_cost=2**16, rounds=4, parallelism=2)
 
 
+def generate_otp_secret() -> str:
+    return base64.b32encode(os.urandom(20)).decode()
```
zenmonkeykstop, line 39: Since the move to Focal SD can probably use secrets (Python 3.6+) now, so you could also do something like:

```
import string
import secrets

def generate_otp_secret() -> str:
    symbols = string.ascii_uppercase + string.digits
    return ''.join(secrets.choice(symbols) for i in range(OTP_LENGTH))
```

Might be good to abstract away from os.urandom, and it achieves the goal of taking pyotp out of the requirements.

evilaliv3: I agree actually. would you like to author this ?

evilaliv3:
Closing this pull in order to open a clean one with a different implementation.

PR #5954, opened 2021-05-19
Eliminate flag-for-reply process

### LINKED ISSUE #1584: Get rid of entropy checks

rmol:
## Status

Ready for review

## Description of Changes

Fixes #1584.

This PR removes the `haveged` package, the checks of the entropy pool before generating source keypairs, and asynchronous source key generation. Source keys are now generated if they don't exist whenever a source visits `/lookup` or submits something, and the `Source.flagged` column and all associated code have been removed. The `flag` route in the journalist API is now a no-op, always returning a 200 status with a message explaining that the action is no longer needed. The `is_flagged` attribute of the JSON representation of sources returned from the API is now always false.

## Testing

- `git checkout 1584-trust-in-disorder`
- `make dev`
- Visit the source interface and create a new source.
- In the container (`docker exec -it securedrop-dev-0 bash`) run  `sqlite3 /var/lib/securedrop/db.sqlite "select id, filesystem_id  from sources order by id"` and `gpg --homedir  /var/lib/securedrop/keys/ --list-keys` to verify that the new source  has a key.
- Log in as a journalist. Navigate to any individual source page and confirm that it says "You can write a secure reply to the person who submitted these messages and/or files:".
- Get into the dev container with `docker exec -it securedrop-dev-0 bash`.
- Find the filesystem ID of the source you logged in as: `sqlite3 /var/lib/securedrop/db.sqlite "select filesystem_id from sources where journalist_designation = 'its-journalist-designation'"`
- Delete the GPG keys of the source:
  - `gpg --homedir /var/lib/securedrop/keys --yes --pinentry-mode=loopback --delete-secret-keys FILESYSTEM_ID`
  - `gpg --homedir /var/lib/securedrop/keys --yes --pinentry-mode=loopback --delete-keys FILESYSTEM_ID`
- Delete the key fingerprint cache in Redis: `redis-cli del sd/crypto-util/fingerprints`
- As the journalist, reload the source page and confirm that it says: "This source has no encryption key. An encryption key will be generated for the source the next time they log in, after which you will be able to reply to the source here."
- Submit a message as the source.
- As the journalist, reload the source page. It should again allow you to reply to the source.
- Log out of the source interface.
- Delete the source's GPG keypair and clear the Redis fingerprint cache again.
- Confirm that the journalist can no longer reply.
- Log in as the source.
- Confirm that the journalist can again reply.

## Deployment

What could go wrong? :slightly_smiling_face: 

I believe the changed logic will cover existing cases of sources without keys. Any source that was flagged for reply should get a key the next time they log in, simply because they don't have a key. The changed messaging in the journalist interface could be confusing for journalists who've had to flag sources for reply, but the effective flow is the same: you just have to wait for the source to return.

Choose one of the following:


securedrop/journalist_app/api.py
```diff
@@ -175,11 +175,7 @@ def remove_star(source_uuid: str) -> Tuple[flask.Response, int]:
     @api.route('/sources/<source_uuid>/flag', methods=['POST'])
     @token_required
     def flag(source_uuid: str) -> Tuple[flask.Response, int]:
-        source = get_or_404(Source, source_uuid,
-                            column=Source.uuid)
-        source.flagged = True
-        db.session.commit()
-        return jsonify({'message': 'Source flagged for reply'}), 200
+        return jsonify({'message': 'Sources no longer need to be flagged for reply'}), 200
```
eloquence, line 178: It may be worth marking the endpoint explicitly as deprecated, to be removed in a future release.

kushaldas: I guess we can mark it as deprecated in the documentation, then we don't have to update this PR.

kushaldas (approved):
Went through the whole test steps, and also visually inspected the code changes. 

We should mark the API deprecation in the documentation, other than this I am okay with this change.

redshiftzero (approved):
visually reviewed the diff and am relying on kushal's functional review, this LGTM!

PR #5958, opened 2021-05-20
Bump 2FA secret bit length from 80 to 160 bits as recommended by RFC4226

### LINKED ISSUE #5933: Bump 2FA secret bit length from 80 bits to 160bits as recommended by RFC4226

evilaliv3:
## Status

Ready for review 

## Description of Changes

Resolves #5933

This clean patch already includes revisions as by https://github.com/freedomofpress/securedrop/pull/5934 and is ready for merge con current devel branch.

\cc @eloquence @zenmonkeykstop 

### dev environment:

- set up a dev environment from this branch with `make dev`
- [ ]  A login attempt  with the default `journalist` account and an incorrect TOTP code fails
- [ ] The `journalist` account can successfully log in using its existing 16-char shared secret to generate a valid TOTP code
- [ ] The larger QR code and longer shared secret are displayed correctly
- [ ] The `journalist` account's TOTP credentials can successfully be updated using the QR code and an authenticator app
- [ ] The `journalist` account's TOTP credentials can successfully be updated using the text shared secret and an authenticator app

### upgrade scenario:

- set up a prod VM environment using the latest release (`2.0.2`)
- set up a JI account with TOTP (and another using HOTP, if a U2F key is available)
- [ ] verify that you can log in using the account credentials
- build packages from this branch and use the upgrade scenario to update the prod VMs using the packages
- [ ] Verify that the securedrop-app-code update completed successfully
- [ ] verify that you can still  log in using the account credentials
- [ ] Verify that you can successfully update the account's TOTP  via the JI, and that you get a 32-char shared secret
- [ ] If an account using HOTP was set up, verify that it can be updated.
- [ ] Verify that a new account can be created and that its TOTP secret is 32-chars and valid.

## Deployment

- Existing instances will have 16-char OTP secrets - these will still be valid, but any new or updated secrets will be 32 chars.

Choose one of the following:

Choose one of the following:


securedrop/models.py
```diff
@@ -566,7 +576,7 @@ def valid_password(self, passphrase: 'Optional[str]') -> bool:
         return is_valid
 
     def regenerate_totp_shared_secret(self) -> None:
-        self.otp_secret = pyotp.random_base32()
+        self.otp_secret = generate_otp_secret()
```
zenmonkeykstop: We have an issue open to eliminate the pyotp dependency altogether ( https://github.com/freedomofpress/securedrop/issues/5613 ) would you be open to expanding this PR to cover that, or to additions to the PR by our team to get rid of pytop altogether?

securedrop/models.py
```diff
@@ -403,7 +413,7 @@ class Journalist(db.Model):
     is_admin = Column(Boolean)  # type: Column[Optional[bool]]
     session_nonce = Column(Integer, nullable=False, default=0)
 
-    otp_secret = Column(String(16), default=pyotp.random_base32)
+    otp_secret = Column(String(32), default=generate_otp_secret)
```
zenmonkeykstop: Varchars in sqlite ignore the length constraint, but a db migration is a good idea regardless. Happy to add it, as it's more of a maintenance task.

securedrop/models.py
```diff
@@ -35,6 +37,14 @@
 ARGON2_PARAMS = dict(memory_cost=2**16, rounds=4, parallelism=2)
 
 
+def generate_otp_secret() -> str:
+    """
+    Generate an OTP secret of 160 bits encoded base32
+    """
+    symbols = string.ascii_uppercase + string.digits
+    return ''.join(secrets.choice(symbols) for i in range(32))
```
zenmonkeykstop: For readability it might make sense to define the secret length like `OTP_LENGTH = 32` and use that throughout.

securedrop/models.py
```diff
@@ -35,6 +37,14 @@
 ARGON2_PARAMS = dict(memory_cost=2**16, rounds=4, parallelism=2)
 
 
+def generate_otp_secret() -> str:
+    """
+    Generate an OTP secret of 160 bits encoded base32
+    """
+    symbols = string.ascii_uppercase + string.digits
```
zenmonkeykstop: D'oh, my bad - this is 36 symbols. Currently the secret keys being generated use [ascii_upper]+[2.3,4,5,6,7], which makes sense in terms of readability (don't wanna mix up 0 and O or 1 and l).

zenmonkeykstop (changes requested):

securedrop/models.py
evilaliv3: i was in favour of the simple number and a comment to not confuse further developers thinking that this could be tweaked.

securedrop/models.py
zenmonkeykstop: (This plus alembic migration woes covers off the app-test failures in CI.)

evilaliv3: Correct, this is what pyotp does ABCDEFGHIJKLMNOPQRSTUVWXYZ234567

As for the symbol choice  i consider that who speculated on this idea when defining RFC 4648 Base32  didnt really evaluate end users; in my opinion end users wont know anyway that we are not using 0 ad 1 and will confuse O with 0 and 1 with I anyway.

I would suggest to use ABCDEFGHJKLMNPQRSTUVWXYZ23456789 that does not include any of the symbols O 0 I 1

securedrop/models.py
evilaliv3: I'm not able to move this forward right know but please feel free to add other changes and i may be able to support with the review. thank you

securedrop/models.py
evilaliv3: @zenmonkeykstop: Could ABCDEFGHJKLMNPQRSTUVWXYZ23456789' work better as it does not include any of O 0 I 1 ?

evilaliv3: In relation to this change i've opened a pull on pyotp so to get to understand if any drawback could exist in this variation: https://github.com/pyauth/pyotp/pull/119

zenmonkeykstop: Probably the first thing that comes to mind is interoperability with apps - just as an example google authenticator considers 8 and 9 illegal chars in secret keys. If that's the set that pyotp uses it's probably worth sticking to it, so the only ux change is that secrets are longer.

evilaliv3: Nice catch. I was actually expecting this but did not find confirmations in the RFC about this limitation so i did not try.

Thank you @zenmonkeykstop .

I've corrected the patch accordingly.

zenmonkeykstop (dismissed):

zenmonkeykstop:
Hey @evilaliv3, after looking at what would be involved in using `cryptography` in lieu of `pyotp`, it seems simpler just to update the dependency.  This also lets us use their secret generator which now enforces a 32-char length while preserving compatibility with authenticator apps &c.

There will be some additional work on docs and the diff review for the `pyotp` update, but if you're happy with this approach I can take care of that. Feel free to let me know your thoughts!

kushaldas:
Will push a change for typelint errors.

kushaldas:
Changes pushed so far:

- Type annotation fixes
- Debian update command needed change due to new Debian stable release
- Alembic heads merge (there were 2 heads)
- Putting one type annotation fix under conditional block to fix broken tests (all of them)

### The heads situation before

```
(securedrop-app-code) root@69d52d7a527b:~/code/securedrop/securedrop# alembic history
92fba0be98e9 -> de00920916bf (head), Updates journalists.otp_secret length from 16 to 32
b060f38c0c31 -> 1ddb81fb88c2 (head), unique_index_for_instanceconfig_valid_until
92fba0be98e9 -> b060f38c0c31, drop Source.flagged
48a75abc0121 -> 92fba0be98e9 (branchpoint), Added organization_name field in instance_config table
35513370ba0d -> 48a75abc0121, add seen tables
523fff3f969c -> 35513370ba0d, add Source.deleted_at
3da3fcab826a -> 523fff3f969c, add versioned instance config
60f41bb14d98 -> 3da3fcab826a, delete orphaned submissions and replies
a9fe328b053a -> 60f41bb14d98, Add Session Nonce To Journalist
b58139cfdc8c -> a9fe328b053a, Migrations for SecureDrop's 0.14.0 release
f2833ac34bb6 -> b58139cfdc8c, add checksum columns and revoke token table
Revision ID: b58139cfdc8c
Revises: f2833ac34bb6
Create Date: 2019-04-02 10:45:05.178481
6db892e17271 -> f2833ac34bb6, add UUID column for users table
e0a525cbab83 -> 6db892e17271, add reply UUID
2d0ce3ee5bdc -> e0a525cbab83, add column to track source deletion of replies
fccf57ceef02 -> 2d0ce3ee5bdc, added passphrase_hash column to journalists table
3d91d6948753 -> fccf57ceef02, create submission uuid column
faac8092c123 -> 3d91d6948753, Create source UUID column
15ac9509fc68 -> faac8092c123, enable security pragmas
<base> -> 15ac9509fc68, init
```
### Current error status from alembic migration

Right now there will be a few migration related failures as the column size is changed in the original work on the PR.
Like: 

```
tests/test_alembic.py::test_schema_unchanged_after_up_then_downgrade[e28559912ae5] FAILED                                                                         [ 50%]                                                                      [ 60%]
tests/test_alembic.py::test_schema_unchanged_after_up_then_downgrade[1ddb81fb88c2] FAILED  
```

```
[<TracebackEntry /root/code/securedrop/securedrop/tests/test_alembic.py:159>, <TracebackEntry /root/code/securedrop/securedrop/tests/test_alembic.py:61>]
test_schema_unchanged_after_up_then_downgrade[1ddb81fb88c2] failed; it passed 0 out of the required 1 times.
	<class 'AssertionError'>
	Schema for ('table', 'journalists', 'journalists') did not match:
Left:
CREATE TABLE "journalists" (
	id INTEGER NOT NULL, 
	uuid VARCHAR(36) NOT NULL, 
	username VARCHAR(255) NOT NULL, 
	first_name VARCHAR(255), 
	last_name VARCHAR(255), 
	pw_salt BLOB, 
	pw_hash BLOB, 
	passphrase_hash VARCHAR(256), 
	is_admin BOOLEAN, 
	session_nonce INTEGER NOT NULL, 
	otp_secret VARCHAR(32), 
	is_totp BOOLEAN, 
	hotp_counter INTEGER, 
	last_token VARCHAR(6), 
	created_on DATETIME, 
	last_access DATETIME, 
	PRIMARY KEY (id), 
	CHECK (is_admin IN (0, 1)), 
	CHECK (is_totp IN (0, 1)), 
	UNIQUE (uuid), 
	UNIQUE (username)
)
Right:
CREATE TABLE journalists (
	id INTEGER NOT NULL, 
	uuid VARCHAR(36) NOT NULL, 
	username VARCHAR(255) NOT NULL, 
	first_name VARCHAR(255), 
	last_name VARCHAR(255), 
	pw_salt BLOB, 
	pw_hash BLOB, 
	passphrase_hash VARCHAR(256), 
	is_admin BOOLEAN, 
	session_nonce INTEGER NOT NULL, 
	otp_secret VARCHAR(16), 
	is_totp BOOLEAN, 
	hotp_counter INTEGER, 
	last_token VARCHAR(6), 
	created_on DATETIME, 
	last_access DATETIME, 
	PRIMARY KEY (id), 
	UNIQUE (username), 
	UNIQUE (uuid), 
	CHECK (is_admin IN (0, 1)), 
	CHECK (is_totp IN (0, 1))
)

```

conorsch (approved):
Looks great! Only ran through the dev portion of the test plan, but all that checked out just fine. The VM-based scenarios we can touch on again as part of release QA.

Thanks, @evilaliv3, for submitting this, and for your patience during deep review. Thanks also to @zenmonkeykstop and @kushaldas for giving it lots of attention. :smiley:

PR #5963, opened 2021-05-27
Adds /sources/<source_uuid>/conversation endpoint supporting DELETE

zenmonkeykstop:
## Status

Work in progress

## Description of Changes

Adds an API endpoint `/sources/<source_uuid>/conversation`,  which, when called via the DELETE method, deletes all submissions and replies while preserving the source account. All other methods just return a `405`.

## Testing

Dev container:
- log into the JI via `http://localhost:8081` in a browser
- note the journalist designation of the  second source
- copy the test script in https://gist.github.com/zenmonkeykstop/2bd2ae05e486729743497d0914a6d28e to your local workstation, install its dependencies with `pip install pyotp requests`, and run it with `python3 wipe-convo.py`
  - [ ] The response for the GET request includes the text `method not allowed`, and the application log in `make dev` output shows a `405` status code for the request
  - [ ] After the GET request, the second listed source and its associated data is still present
- Hit `Enter` when prompted by the script to run the DELETE request
  - [ ] The response for the DELETE request includes the text `source data deleted`, and the application log in `make dev` output shows a `200` status code for the request
  - [ ] After the get request, the source is still present but associated conversation data is deleted
- [ ] Sources other than the second listed source are unaffected, with their conversations intact.

## Deployment
- This will be deployed with next server code update, no postint tasks or other fixup is required.
 
Choose one of the following:

### DIFF
--- a/securedrop/journalist_app/api.py
+++ b/securedrop/journalist_app/api.py
@@ -181,6 +181,16 @@ def make_blueprint(config: SDConfig) -> Blueprint:
         db.session.commit()
         return jsonify({'message': 'Source flagged for reply'}), 200
 
+    @api.route('/sources/<source_uuid>/conversation', methods=['DELETE'])
+    @token_required
+    def source_conversation(source_uuid: str) -> Tuple[flask.Response, int]:
+        if request.method == 'DELETE':
+            source = get_or_404(Source, source_uuid, column=Source.uuid)
+            utils.delete_source_files(source.filesystem_id)
+            return jsonify({'message': 'Source data deleted'}), 200
+        else:
+            abort(405)
+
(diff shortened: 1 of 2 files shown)


kushaldas (commented):
The change is good, with proper test cases. We will also need  a corresponding PR for the docs repo in the [API](https://docs.securedrop.org/en/stable/development/journalist_api.html) section.

I will do a formal review after it is marked as "ready for review"

zenmonkeykstop:
thanks @kushaldas, test plan etc in progress.

rmol (approved):
:+1: Test plan passes, works with WIP client safe deletion changes.

PR #5974, opened 2021-06-04
Ensure only one active InstanceConfig can exist

### LINKED ISSUE #5973: In dev environment, multiple active InstanceConfig records can be created, breaking things
## Description

If the database is created with `create_all` instead of migrations, there won't be a default `InstanceConfig` record. On first request the [source](https://github.com/freedomofpress/securedrop/blob/2971b91b31ba9abf0d6372f4f4db9d796f83e1c5/securedrop/source_app/__init__.py#L136) or [journalist](https://github.com/freedomofpress/securedrop/blob/2971b91b31ba9abf0d6372f4f4db9d796f83e1c5/securedrop/journalist_app/__init__.py#L127) apps will try to [create one](https://github.com/freedomofpress/securedrop/blob/2971b91b31ba9abf0d6372f4f4db9d796f83e1c5/securedrop/models.py#L862), and if there are simultaneous requests, it's possible that more than one active (with `valid_until` null) row can be created, causing problems on subsequent requests.

Changing the dev environment to create the database the same way it's done in staging and production would ensure the presence of an `InstanceConfig` before the apps are started, but we should also try to prevent multiple active configs. A [partial index](https://docs.sqlalchemy.org/en/14/dialects/sqlite.html#partial-indexes) should do that.

## Steps to Reproduce

This should only happen in the dev environment, and it's tricky to reproduce _in vivo_; you have to submit simultaneous requests right after the dev servers are started. We saw this in the demo service when multiple scripts were testing the site.

It's easier to demonstrate the problem with the `instance_config` schema alone: in one database session, add an `InstanceConfig` instance but don't commit, then in another session, call `InstanceConfig.get_current()` to create a record with the same default values, then commit the original session. This will work and you'll wind up with two active configurations in the table.

rmol:
## Status

Ready for review
## Description of Changes

Fixes #5973.

Adds a unique partial index on `InstanceConfig.valid_until` where it is null, and in `InstanceConfig.get_current`, handles the now-possible `IntegrityError` and queries again to pick up the row added between the previous query and the insertion attempt.

## Testing

- `git checkout develop`
- `git checkout origin/there_can_be_only_one -- securedrop/tests/test_db.py`
- `make test TESTFILES="tests/test_db.py::test_only_one_active_instance_config_can_exist"` will fail because it doesn't encounter `IntegrityError`.
- `git checkout there_can_be_only_one`
- Run the test again. It should pass.

## Deployment

The migration that adds the index could conceivably fail if a site had multiple rows in `instance_config` with `valid_until` null, but every production database should have been created with migrations, which would have created a default row in `instance_config`. Any site that did somehow have multiple active configurations would not be functioning.

Choose one of the following:


zenmonkeykstop (approved):
Tsst plan passes, and new instance configs can be created via the JI without any issues. Agreed on deployment comments.

PR #5990, opened 2021-06-14
fix: updated validate role to skip journalist auth check on fresh installs

### LINKED ISSUE #5989: [2.0.0-rc1] Tails environment validation fails on first installer run
## Description

A check was added in the `validate` role that assumes that `app-journalist.auth_private` should always be present. This is not the case for initial installs.

## Steps to Reproduce

In a prod environment, complete the initial setup steps and run `./securedrop-admin install`

## Expected Behavior

Installation proceeds past the `Confirm that the Journalist auth file is present` step.

## Actual Behavior

Installation fails at the `Confirm that the Journalist auth file is present` step.

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Fixes #5989 .

Modifies Tails validation check to only check for journalist auth file if there's also a source THS file, to allow for the case of fresh installs.

## Testing

- Perform a prod fresh install and confirm that it does not fail early in the `validate` role

### DIFF
--- a/install_files/ansible-base/roles/validate/tasks/validate_tails_environment.yml
+++ b/install_files/ansible-base/roles/validate/tasks/validate_tails_environment.yml
@@ -50,6 +50,11 @@
     path: "/home/amnesia/Persistent/securedrop/install_files/ansible-base/app-journalist.auth_private"
   register: v3_journalist_auth_file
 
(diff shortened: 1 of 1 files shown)


rmol (approved):
Works.

Issue #5985, opened 2021-06-10
status-indicator `img`s need explanatory `alt`s and ARIA annotations


cfm:
## Description

[Accessibility Lab recommendation](https://docs.google.com/document/d/1jZlKco2YgOrrI4PmTkG_Q8lKuV5VxCds/edit#):

> In the case of the multiple “stars”, it is recommended that you relate the text linked to that specific star, so the user know (out of context) with just the button descriptive text, what is the action or functionality: for example: “star two-handed dimple”

1. `securedrop/journalist_templates/_source_row.html`
    1. [x] star `img` 
    2. [x] `input[type=checkbox]`
2. `securedrop/journalist_templates/col.html`
    1. [x] `input[type=checkbox]`
    2. [x] file-type indicator
    3. [x] un/read indicator

## User Research Evidence

[WCAG 1.1.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/text-equiv-all.html) per #5972.

Issue #5987, opened 2021-06-10
semantic (HTML5/ARIA) page structure in Journalist Interface


cfm:
# Description

[Accessibility Lab recommendations](https://docs.google.com/document/d/1jZlKco2YgOrrI4PmTkG_Q8lKuV5VxCds/edit#) include:

- `h1`...`hn` hierarchy
- HTML5 section elements
- ARIA roles ("landmarks"), labels, and descriptions

Our findings include:

- moving decorative `img` elements to CSS styles on semantic elements
- Keep in mind per @eloquence in <https://github.com/freedomofpress/securedrop/issues/5972#issuecomment-947317684>:
    > One thing we may want to look into for the next round of changes is how to minimize the need to re-translate strings - e.g., should we generate our own translations in cases where only HTML elements or attributes change (can we do so in the main repo and then import these into Weblate?).

## User Research Evidence

[WCAG 1.3.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html) per #5972.

## Progress

To invert the way things are usually done :-):

1. [x] **First pass:**  Refactor for semantic markup (test in developer tools and screen-readers) and keep `journalist.js` in sync
    - [x] decorative `img`/`hr`/etc. elements: comment out as `FIXME`s to move to CSS styles on semantic elements 
    - [x] linters pass
    - [x] tests pass
2. [x] **Second pass:**  Audit in validators and accessibility tools ([axe](https://www.deque.com/axe/), [WAVE](https://wave.webaim.org/)), e.g.:
    - `h1`...`hn` hierarchy
    - HTML5 section elements
    - ARIA roles ("landmarks"), labels, and descriptions
3. [x] Accessibility-focused review of draft PR before continuing further
4. [x] Review string changes for translation per <https://github.com/freedomofpress/securedrop/issues/5972#issuecomment-947317684> 
5. [x] **Third pass:**  CSS and images broken in first pass (test visually in Tor Browser)
    - [x] `FIXME`s
        - [x] Revert 6d4602da288b398dea3680dfb9faba67cf11e85a
    - [x] tests pass
6. [x] Write and run through test plan in #6165

| Path | First Pass | Second Pass | Third Pass |
| --- | --- | --- | --- |
| `securedrop/journalist_templates` | ✓ | ✓ | ✓ |
| `├──_confirmation_modal.html` | ✓ | ✓| ✓ |
[...]

Issue #5986, opened 2021-06-10
 semantic (HTML5/ARIA) page structure in Source Interface


cfm:
## Description

[Accessibility Lab recommendations](https://docs.google.com/document/d/1jZlKco2YgOrrI4PmTkG_Q8lKuV5VxCds/edit#) include:

- `h1`...`hn` hierarchy
- HTML5 section elements
- ARIA roles ("landmarks"), labels, and descriptions

Our findings include:

- moving decorative `img` elements to CSS styles on semantic elements 

## User Research Evidence

[WCAG 1.3.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html) per #5972.

## Progress

To invert the way things are usually done :-):

1. [x] **First pass:**  Refactor for semantic markup (test in developer tools and screen-readers)
    - [x] decorative `img`/`hr`/etc. elements: mark as `FIXME`s to move to CSS styles on semantic elements 
    - [x] linters pass
    - [x] tests pass
2. [x] **Second pass:**  Audit in validators and accessibility tools ([axe](https://www.deque.com/axe/), [WAVE](https://wave.webaim.org/)), e.g.:
    - `h1`...`hn` hierarchy
    - HTML5 section elements
    - ARIA roles ("landmarks"), labels, and descriptions
3. [x] **Third pass:**  CSS and images broken in first pass (test visually in browser)
    - [x] `FIXME`s
    - [x] visual review

| Path | First Pass | Second Pass | Third Pass |
| --- | --- | --- | --- |
`securedrop/source_templates` | ✓ | ✓ | ✓ |
`├── banner_warning_flashed.html` | ✓ | ✓ | ✓ |
`├── base.html` | ✓ | ✓ | ✓ |
`├── error.html` | ✓ | ✓ | ✓ |
`├── first_submission_flashed_message.html` | ✓ | ✓ | ✓ |
`├── flashed.html` | ✓ | ✓ | ✓ |
`├── footer.html` | ✓ | ✓ | ✓ |
`├── generate.html` | ✓ | ✓ | ✓ |
`├── index.html` | ✓ | ✓ | ✓ |
`├── locales.html` | ✓ | ✓ | ✓ |
`├── login.html` | ✓ | ✓ | ✓ |
`├── logout.html` | ✓ | ✓ | ✓ |
`├── lookup.html` | ✓ | ✓ | ✓ |
[...]

PR #5982, opened 2021-06-10
functional icon needs functional alt tag

KwadroNaut:
See https://github.com/freedomofpress/securedrop/issues/5743 
There are probably more, similar, issues floating around.

## Status

Ready for review

## Description of Changes

Fixes part of #5743
I'm expecting that there are more alt tags that can be improved, but it's not on my priority list.

Changes proposed in this pull request:

## Deployment

I haven't tested this, translation files need to be re-generated, I think for most l10n work it's impact is minimal.

Choose one of the following:

- [X] These changes do not require documentation

PR #5996, opened 2021-06-16
normalizes decorative and functional `img` `alt` attributes

### LINKED ISSUE #5983: decorative `img`s need empty `alt`s
## Description

[Accessibility Lab recommendation](https://docs.google.com/document/d/1jZlKco2YgOrrI4PmTkG_Q8lKuV5VxCds/edit#):

> Images that do not convey content, are decorative, or are redundant to content that is already conveyed in text must be given null alternative text (alt="") 

## User Research Evidence

[WCAG 1.1.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/text-equiv-all.html) per #5972.

## Implementation

1. [x] Make list of images that are self-evidently decorative even out of context
2. [x] Close review of `alt` attributes for images in step (1)
3. [x] Skim other `img`s for exceptions

### LINKED ISSUE #5984: functional `img`s need explanatory `alt`s
## Description

[Accessibility Lab recommendation](https://docs.google.com/document/d/1jZlKco2YgOrrI4PmTkG_Q8lKuV5VxCds/edit#):

>  All actionable images (e.g., links, buttons, controls) must have alternative text.

## User Research Evidence

[WCAG 1.1.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/text-equiv-all.html) per #5972.

**See also:**  #5982

cfm:
## Status

Ready for review

## Description of Changes

1. Resolves #5983
2. Resolves #5984 
3. Lays groundwork for #5985

It turned out to be easier to review all `alt`s application-wide, commit the changes by category or resolution, and then submit them together in one omnibus PR.  #5985 has some extra bits outstanding.

## Testing

1. Navigate through the Source and Journalist Interfaces with a screen-reader running and listen for the `alt` attributes in context.
2. Note that all `img` elements have `alt` attributes:
```sh-session
$ grep "<img" **/*.html | grep -cv "alt="
0
```
3. Note that all non-empty, non-customizable `alt` attributes are translated:
```sh-session
$ grep "<img" **/*.html | grep -v 'alt=""' | grep -v "organization_name" | grep -cv "gettext"
0
```

## Deployment

No special considerations.

- [X] Linting (`make lint`) and tests (`make test`) pass in the development container


cfm:
@KwadroNaut [asks](https://github.com/freedomofpress/securedrop/pull/5982#issuecomment-862793746) (thanks!):

> the session timeout broom icon: did you check how this is handled with Tails/Windows/... screen readers? I never recognize that icon as a broom, just know where to click ¯_(ツ)_/¯

I hear this read straight through, skipping the empty `alt=""` attributes.  I understand this to be the desired narration, since the "New Identity" label is included in the surrounding text and corresponds to the tooltip of that toolbar icon in Tor Browser itself.  Would you recommend a different approach here?

> Click the New Identity button in your Tor Browser's toolbar. This will clear your Tor Browser activity data on this device.

---

> the hover icon in the admin template: is that really the way to go with an empty alt tag when there is a tooltip otherwise?

I'd thought that the two `img`s (`{off,on}-hover`) would have their `alt`s read consecutively, in duplicate.  But this doesn't appear (sound!) to be the case based on how the screen-reader follows the CSS, so I've added `alt`s back in for both of them in the revised 88f2a87.

KwadroNaut:
> @KwadroNaut [asks](https://github.com/freedomofpress/securedrop/pull/5982#issuecomment-862793746) (thanks!):
> 
> > the session timeout broom icon: did you check how this is handled with Tails/Windows/... screen readers? I never recognize that icon as a broom, just know where to click ¯_(ツ)_/¯
> 
> I hear this read straight through, skipping the empty `alt=""` attributes. I understand this to be the desired narration, since the "New Identity" label is included in the surrounding text and corresponds to the tooltip of that toolbar icon in Tor Browser itself. Would you recommend a different approach here?
> 
How does your screenreader work when using Tor browser is my question: the corresponding tooltip in Tor Browser is the answer: it's good what you did!

securedrop/journalist_templates/admin.html
```diff
@@ -28,20 +28,20 @@ <h1>{{ gettext('Admin Interface') }}</h1>
...
-          <img src="{{ url_for('static', filename='icons/trash-disabled.png') }}" width="14" height="16" alt="{{ gettext('trash icon disabled') }}">
+          <img src="{{ url_for('static', filename='icons/trash-disabled.png') }}" width="14" height="16" alt="{{ gettext('Delete disabled') }}">
         </td>
 	{% else %}
         <td>
           <button type="submit" class="plain delete-user" formaction="/admin/delete/{{ user.id }}" data-username="{{ user.username}}">
             <i title="{{ gettext('Delete user {username}').format(username=user.username) }}">
-              <img src="{{ url_for('static', filename='icons/trash.png') }}" class="off-hover" width="14" height="16" alt="{{ gettext('trash icon') }}">
-              <img src="{{ url_for('static', filename='icons/trash-hover-red.png') }}" class="on-hover" width="14" height="16" alt="{{ gettext('trash icon') }}">
+              <img src="{{ url_for('static', filename='icons/trash.png') }}" class="off-hover" width="14" height="16" alt="{{ gettext('Delete') }}">
```
zenmonkeykstop: For clarity and consistency with "Edit user" above, should these be "Delete User"?

securedrop/journalist_templates/admin.html
```diff
@@ -28,20 +28,20 @@ <h1>{{ gettext('Admin Interface') }}</h1>
...
             <i title="{{ gettext('Edit user {username}').format(username=user.username) }}">
-              <img src="{{ url_for('static', filename='icons/edit-user.png') }}" class="icon" width="18" height="16" alt="{{ gettext('edit user icon') }}">
+              <img src="{{ url_for('static', filename='icons/edit-user.png') }}" class="icon" width="18" height="16" alt="{{ gettext('Edit user') }}">
             </i>
           </a>
         </td>
 	{% if user.id == g.user.id %}
         <td>
-          <img src="{{ url_for('static', filename='icons/trash-disabled.png') }}" width="14" height="16" alt="{{ gettext('trash icon disabled') }}">
+          <img src="{{ url_for('static', filename='icons/trash-disabled.png') }}" width="14" height="16" alt="{{ gettext('Delete disabled') }}">
```
zenmonkeykstop: As per below, I'd err on the side of making this more descriptive, like "User deletion disabled."

securedrop/journalist_templates/col.html
```diff
@@ -30,23 +30,23 @@
...
                 <input type="checkbox" name="doc_names_selected" value="{{ doc.filename }}" class="doc-check">
                 <span title="{{ gettext('Reply') }}" class="icon">
-                  <img src="{{ url_for('static', filename='icons/reply.png') }}" class="icon" width="16" height="16" alt="{{ gettext('reply icon') }}">
+                  <img src="{{ url_for('static', filename='icons/reply.png') }}" class="icon" width="16" height="16" alt="{{ gettext('Reply') }}">
                 </span>
               {% else %}
                 <input type="checkbox" name="doc_names_selected" value="{{ doc.filename }}" class="doc-check">
                 <span title="{{ gettext('Read') }}" class="icon">
-                  <img src="{{ url_for('static', filename='icons/check.png') }}" class="icon" width="16" height="12" alt="{{ gettext('check icon') }}">
+                  <img src="{{ url_for('static', filename='icons/check.png') }}" class="icon" width="16" height="12" alt="{{ gettext('Check') }}">
```
zenmonkeykstop: The checkmark here indicates that it's a reply that has been "deleted" by the source, which means that it's removed from their view in the source interface tho it persists in the journalist interface. This is interpreted as it having been read by the source, so something like "reply read" or "reply viewed" might work better.

securedrop/journalist_templates/flashed.html
```diff
@@ -4,11 +4,11 @@
     {% if category not in ["banner-warning","logo-success","logo-error","submission-preferences-success","submission-preferences-error","org-name-success","org-name-error", "testalert-success", "testalert-error", "testalert-notification"] %}
       <div class="flash {{ category }}">
         {% if category == "notification" %}
-        <img src="{{ url_for('static', filename='i/flash-notification.png') }}" height="35" width="15">
+        <img src="{{ url_for('static', filename='i/flash-notification.png') }}" alt="" height="35" width="15">
```
zenmonkeykstop: Not all messages are prefixed with "Success!," "Warning!" etc, so these symbols are not purely decorative but indicate flash message type. As such it might be worth including descriptive alt text for them.

securedrop/journalist_templates/preferences_saved_flash.html
```diff
@@ -6,11 +6,11 @@
         {% set category_status = category.split('-')|last %}
           <div class="flash {{ category_status }}">
           {% if category_status == "success" %}
-              <img src="{{ url_for('static', filename='i/flash-success.png') }}" height="15" width="20">
+              <img src="{{ url_for('static', filename='i/flash-success.png') }}" alt="" height="15" width="20">
```
zenmonkeykstop: Same as per above for other flash message icons.

securedrop/source_templates/index.html
```diff
@@ -74,7 +74,7 @@ <h2 class="welcome-text">
     reading the page. Hidden by default. #}
     <div class="bubble" id="security-setting-info" class="info">
       <ol>
-        <li>{{ gettext('Click the <img src="{icon}" alt="shield icon"> in the toolbar above').format(icon=url_for("static", filename="i/font-awesome/black/guard.svg"))  }}</li>
+        <li>{{ gettext('Click the <img src="{icon}" alt="&quot;Security Level&quot; button"> in the toolbar above').format(icon=url_for("static", filename="i/font-awesome/black/guard.svg"))  }}</li>
```
zenmonkeykstop, line 77: Does this need to be in quotes?

zenmonkeykstop (changes requested):
Some changes requested, but also I think there's some room for discussion on a couple of things that might negate said changes:
- Should icons in flash messages have descriptive alts (I lean towards yes because otherwise there's no guaranteed indicator of the message type/severity)?
- Should alt descriptions on action buttons just be a single verb like "Delete" or descriptive like "Delete Reply"? (Could go either way depending on whether context of action is clear)

eloquence:
@cfm This fully supersedes #5982, correct?

cfm:
On 21 Jun 2021, at 8:58, Erik Moeller wrote:

> @cfm This fully supersedes #5982, correct?

Yes, though not intentionally; #5982 just hadn’t received a 
maintainers’ review prior to my starting in here, both based on #5982 
and in consultation with @KwadroNaut.

cfm:
Thanks for the review in https://github.com/freedomofpress/securedrop/pull/5996#pullrequestreview-687685824, @zenmonkeykstop.

> * Should icons in flash messages have descriptive alts (I lean towards yes because otherwise there's no guaranteed indicator of the message type/severity)?

I've mulled this over, and I think I lean towards a more-pedantic interpretation.  :-)  If there's a need for an indicator of the `category` outside of the `message` text, that suggests to me that messages are missing headings that would either:

1. be independent of the existing decorative (`alt=""`) images and then styled to be invisible on CSS `@media screen` (i.e., reserved for `@media speech`); or
2. replace the existing images, which would be refactored in CSS to be *purely* decorative on the `H{n}` element (i.e., not present as an `img` element at all).

Again, I want to concede that this *is* pedantic: both approaches will look the same in a browser and sound the same in a screen-reader.  But my instinct here is that the question of these `alt` attributes actually stands in for the structural questions of #5986 and #5987.

So, I propose to:

1. go ahead and add descriptive `alt`s here (changes forthcoming), so that this pull request will provide the desired narration for now; and then
2. flag these messages (and any similar cases) for consideration in #5986 and #5987.

How does this sound to you, @zenmonkeykstop?

> * Should alt descriptions on action buttons just be a single verb like "Delete" or descriptive like "Delete Reply"? (Could go either way depending on whether context of action is clear)

Good catch; thanks.  I'd planned to do another pass on these as part of #5985, to include information about *which* submission/reply/user will be subject to the action, but will go ahead now (changes forthcoming) for the cases you've flagged here.

securedrop/source_templates/index.html
cfm: Since I can't boldface this (very small amount of :-) ["[t]ext taken directly from a user interface"](https://docs.securedrop.org/en/stable/development/documentation_guidelines.html#glossary), I thought I'd quote it instead.  But I'm not set on this if you think that `Security Level button` is clear enough in context to be worth the simplicity.

cfm:
Re: https://github.com/freedomofpress/securedrop/pull/5996#issuecomment-863508238:

> I'd thought that the two `img`s (`{off,on}-hover`) would have their `alt`s read consecutively, in duplicate. But this doesn't appear (sound!) to be the case based on how the screen-reader follows the CSS, so I've added `alt`s back in for both of them in the revised [88f2a87](https://github.com/freedomofpress/securedrop/commit/88f2a87edafc1a34f7cc7567838998713a3898a1).

I've gotten different results on this question in different browsers, so I've re-marked this as draft while I test further.

cfm:
Re: https://github.com/freedomofpress/securedrop/pull/5996#issuecomment-866287505:
 
> > I'd thought that the two `img`s (`{off,on}-hover`) would have their `alt`s read consecutively, in duplicate. But this doesn't appear (sound!) to be the case based on how the screen-reader follows the CSS, so I've added `alt`s back in for both of them in the revised [88f2a87](https://github.com/freedomofpress/securedrop/commit/88f2a87edafc1a34f7cc7567838998713a3898a1).
> 
> I've gotten different results on this question in different browsers, so I've re-marked this as draft while I test further.

The duplication (in some cases, triplication) I'm hearing turns out to come not from the multiple `img[alt]` attributes but from (e.g.):

https://github.com/freedomofpress/securedrop/blob/b486c9acfde6a1216e96b3c6ad8dd32abdc8cc11/securedrop/journalist_templates/admin.html#L41-L46

1. the duplication of the `img[alt]` attributes in (e.g.) the `i[title]` attribute (which is apparently lately [considered harmful](https://silktide.com/blog/i-thought-title-text-improved-accessibility-i-was-wrong/)); and
2. idiosyncrasies in how different screen-readers (and their configurations) infer and narrate [ARIA groups](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_group_role).

So the changes proposed here stand in scope, and this behavior is more evidence in favor of the semantic refactoring ticketed in #5986 and #5987.

securedrop/source_templates/index.html
zenmonkeykstop: That's fair, it's a cromulent alternative

zenmonkeykstop (approved):
Changes look good to me, approving.

Issue #6005, opened 2021-06-18
Unattended upgrades from 1.8.2 to 2.0.0-rc3 fail


zenmonkeykstop:
## Description

Unattended upgrades from 1.8.2 to 2.0.0-rc3 fail to complete successfully on the app server, with the `securedrop-app-code` package being held back.

## Steps to Reproduce

- Set up a prod environment (hardware or VMs) with current version 1.8.2
- install testing apt key and flip repos from `apt.freedom.press` to `apt-test.freedom.press` in apt sources lists
- run `sudo apt-get update && sudo unattended-upgrades -d`

## Expected Behavior

2.0.0-rc3 packages are installed without error (altbeit with a connection drop if ssh-over-tor is enabled)

## Actual Behavior

`securedrop-app-code` is held back as "package haveged is marked to be deleted"

## Comments

Suggestions to fix, any other relevant information.

kushaldas:
I can now verify this issue.
```
<snipped>
Checking: securedrop-app-code ([<Origin component:'main' archive:'' origin:'SecureDrop' label:'' site:'apt-test.freedom.press' isTrusted:True>])              
sanity check failed for: {'haveged=1.9.1-6ubuntu1', 'securedrop-app-code=2.0.0~rc3+focal'} : pkg haveged is marked to be deleted                              
falling back to adjusting securedrop-app-code's dependencies                                                                                                  
sanity check failed for: {'haveged=1.9.1-6ubuntu1', 'securedrop-app-code=2.0.0~rc3+focal'} : pkg haveged is marked to be deleted                              
Checking: securedrop-config ([<Origin component:'main' archive:'' origin:'SecureDrop' label:'' site:'apt-test.freedom.press' isTrusted:True>])                
Checking: securedrop-keyring ([<Origin component:'main' archive:'' origin:'SecureDrop' label:'' site:'apt-test.freedom.press' isTrusted:True>])               
Checking: securedrop-ossec-agent ([<Origin component:'main' archive:'' origin:'SecureDrop' label:'' site:'apt-test.freedom.press' isTrusted:True>])           
Checking: tor ([<Origin component:'main' archive:'' origin:'SecureDrop' label:'' site:'apt-test.freedom.press' isTrusted:True>])                              
pkgs that look like they should be upgraded: securedrop-config                                                                                                
securedrop-keyring                                                                                                                                            
securedrop-ossec-agent                                                                                                                                        
tor                                                                                                                                                           
[...]

PR #6008, opened 2021-06-21
Disables haveged service in preinst, if unit file is present

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Adds to fix for #6005 .

If haveged is present, stop/disable/mask the service - this is a compromise vs removing the package altogether.

## Testing

- Make sure you're using libvirt-based VMs (upgrade scenario does not support Qubes env)
- `molecule create -s libvirt-prod-focal`
- Boot up admin workstation and install against those prod VMs with ./securedrop-admin install
- make build-debs on this branch on host (ok to run this in parallel with step 2 to save time)
- `make upgrade-start` on host, to set up local apt repo
- Back in admin workstation, run the playbook securedrop-apt-local.yml in the ansible-base directory (make sure to source the admin venv first, see docs), `ansible-playbook --diff -vv ./securedrop-apt-local.yml`
- `ssh app` and then `sudo apt-get update && sudo unattended-upgrade -d`
  - [ ] the unattended upgrade command completes successfully
  - [ ] `apt-cache policy securedrop-app-code` shows package is upgraded
  - [ ] `systemctl status haveged` shows haveged is stopped, disabled, and masked.

## Deployment

1. Upgrades: script change has been tested as above - in cases where haveged is present it should just be disabled
2. New installs: since havaged will not be installed on fresh SD installs from 2.0.0 on, the script won't do anything.

Choose one of the following:

### DIFF
--- a/install_files/securedrop-app-code/debian/preinst
+++ b/install_files/securedrop-app-code/debian/preinst
@@ -13,6 +13,17 @@ set -e
 # for details, see http://www.debian.org/doc/debian-policy/ or
 # the debian-policy package
 
+function service_exists() {
+    # Test for the existence of systemctl services. Added to check for and
+    # disable haveged if present
+
+    local x=$1
+    if systemctl list-unit-files --type service | grep -F "${x}.service"; then
+        return 0
+    else
+        return 1
+    fi
+}
 
 function permanently_disable_swap() {
     # Swap usage is prohibited in the context of SecureDrop, due to risk of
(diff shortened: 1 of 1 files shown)


rmol (approved):
I had to test using my Qubes prod VMs, as I didn't have a virtualized admin environment on Debian. I started from 1.8.2, with `haveged` installed, enabled, and running, and set up a local apt repo instead of using the upgrade scenario's server. With that in place, `unattended-upgrades` completed the upgrade of `securedrop-app-code`, and the `haveged` service was disabled and masked according to plan.

PR #6028, opened 2021-06-28
Shellcheck filter update

rmol:
## Status

Ready for review

## Description of Changes

Skips a lot more irrelevant filesystem content when running shellcheck. Installs shellcheck in the dev container so that we can run `make lint` instead of enumerating everything but shellcheck, making it less likely that this CI job will need adjustment if we add or remove linters.

Adding a line printing the files actually passed to shellcheck showed that with this extra filtering, the only differences in the set of files checked were some third-party scripts under `admin/.tox`.

## Testing

CI is green.

## Deployment

Dev-only.

## Checklist

Choose one of the following:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,14 +72,10 @@ jobs:
       - *installenchant
 
       - run:
-          name: Run all linters but shellcheck
+          name: Run all linters
           command: |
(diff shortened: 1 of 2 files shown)


zenmonkeykstop (approved):
CI is green, lint job appears 2-3 min faster, changes look good on visual review.

PR #6021, opened 2021-06-24
refactors status-indicator `img`s and related elements for ARIA labeling

cfm:
## Status

Ready for review

## Description of Changes

Resolves #5985.

This PR builds on #5996 to make more-meaningful use of both `alt` attributes (on `img` elements) and `aria-*` annotations (on `img` and other elements) in those cases where an image indicates the status of another UI element.  These images are not strictly decorative, but nor is their UI role communicated (and narrated by screen-readers) from functional markup such as `button` elements.  Instead, this more-nuanced refactoring (a) simplifies the existing markup (and narration) and (b) establishes narratable relationships between elements.

This PR (especially cbff6b9) can be thought of as the prototype for the top-down semantic refactoring proposed in #5986 and #5987.

## Testing

With a screen-reader running:

1. **Journalist Interface → log in as a journalist:**  Observe that a source's "star" button, un/starred status, selection checkbox, and link are announced separately and logically, before the other information in the row.
2. **Journalist Interface → log in as a journalist → select a source:**  Observe that a document's name, un/read status, and type are announced together; followed by its selection checkbox; followed by the other information in the row.

## Deployment

No special considerations.

### If you made changes to the server application code:

- [x] Linting (`make lint`) and tests (`make test`) pass in the development container

Issue #6031, opened 2021-07-01
evaluate HTML5 UI elements as browser support improves


cfm:
## Description

As browser (specifically, Firefox ESR → Tor Browser) support for HTML5 UI elements improves, it may be possible to use them in place of some of SecureDrop's custom-styled widgets.  These may be (a) easier to implement and maintain, especially in the JavaScript-free Source Interface, and (b) more accessible to more users.

## User Research Evidence

Prompted by experiments during #5986.

## Examples

- [ ] This section needs to be revised to clarify which widgets have been *partially* implemented as suggested here (e.g., by #6041) and which have not.

| Interface | Template | Widget | HTML5 Element | Firefox Support | Implemented in... |
| --- | --- | --- | --- | --- | --- |
| Source | `index.html` | warning re: Tor Browser security level | [`dialog`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog)[2] |  [53+](https://caniuse.com/dialog) only if [manually enabled](https://bugzilla.mozilla.org/show_bug.cgi?id=840640)| #6041[5] |
| Source | `locales.html` | language selector | [`menu`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu)[1] |  [8+](https://caniuse.com/mdn-html_elements_menu) | |
| Source | `lookup.html` | codename reminder | [`details`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details)[3] | [49+](https://caniuse.com/details) | #6041[4] |
| Source | `lookup.html` | "delete reply" dialog | [`dialog`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog)[2] |  [53+](https://caniuse.com/dialog) only if [manually enabled](https://bugzilla.mozilla.org/show_bug.cgi?id=840640)| |

**Notes:**

1. Alternative: https://css-tricks.com/solved-with-css-dropdown-menus/ (suggested by @SaptakS)
2. Alternative: https://github.com/github/details-dialog-element (suggested by @SaptakS; requires JavaScript polyfill)
3. Suggested by @SaptakS in https://github.com/freedomofpress/securedrop/pull/6041#discussion_r682041892
4. Without explicit `open` state, which we can't add to the DOM on the script-free Source Interface (cf. whatwg/html#3567).
5. Hackily: fixed `open` state and `z-index`, with visibility controlled by: 
[...]

Issue #6003, opened 2021-06-17
all-caps button labels are misread by screen-readers


cfm:
## Description

For example:  A button labeled `ADD USER` will be announced by a screen-reader as an initialism to be spelled out: "A.D.D. user".

https://github.com/freedomofpress/securedrop/blob/012d30e8baeb8ffa7c10a56f4b43c464abf87cf7/securedrop/journalist_templates/admin.html#L6-L9

## Steps to Reproduce

For this example:

1. Navigate to the Admin Interface of the Journalist Interface.
2. Activate a screen-reader.

## Expected Behavior

The button visually labeled `ADD USER` is announced as "add user".

## Actual Behavior

The button visually labeled `ADD USER` is announced as "A.D.D. user".

## Comments

Labels capitalized explicitly in #1538 can be reverted to title case (pending #5821) and then styled with the `text-transform` property.

eloquence:
Note there are i18n arguments against `text-transform`, see https://github.com/freedomofpress/securedrop/issues/1587#issuecomment-694969536

cfm:
After further research:

1. Per https://github.com/freedomofpress/securedrop/issues/1587#issuecomment-694969536: the localization argument against `text-transform` stands.  (Thanks for referring me, @eloquence.)
2. Where testing demonstrates that screen-readers may misread a *specific* label in all caps, [we can override it with `aria-label`](https://webaim.org/discussion/mail_thread?thread=8674#post1), at the cost of adding an extra string to translate.  For example:
```html
 <button type="submit" id="add-user" aria-label="{{ gettext('Add User') }}"> 
   {{ gettext('ADD USER') }} 
 </button> 
```
3. In general, however, and in the absence of a specific misreading, [we should avoid micromanaging the accessibility tree](https://stackoverflow.com/a/53825307).

I'll close this general tracking issue and address specific instances as I encounter them in the course of #5986 and #5987.

Issue #5821, opened 2021-02-25
Consider using sentence case instead of title case in UI elements


rmol:
## Description

It was [pointed out](https://weblate.securedrop.org/translate/securedrop/securedrop/es/?checksum=2cd97277750195c7) (thanks @ekcnl) during translation for 1.8.0 that the button label [**Yes, Delete Selected Source Accounts**](https://github.com/freedomofpress/securedrop/blob/94fdaacc4ff86c3da51af4fee295535770800e1b/securedrop/journalist_templates/_sources_confirmation_final_modal.html#L16) looks odd.

We use a mix of **ALL CAPS** and **Title Case** for button labels. Title case looks all right for **Shorter Labels**, but **Can Start Looking Eccentric When Longer**.

The [Material Design guidelines](https://material.io/components/buttons), to which we've referred in some recent design discussions, suggest either all caps or sentence case. We may not want to pick one or the other exclusively: we sometimes use mixed case for safe options like **Cancel** and all caps for irrevocable actions like **DELETE**. Or perhaps we should indicate danger with other cues, as long labels would also look odd and heavy in all caps. But in any event, sentence case generally works better than title case, and I think we should consider switching.

@ninavizz @zenmonkeykstop

AO-LocLab:
It is also the case for " Yes, Delete Source Account".

eloquence:
I'm not in favor of making this change for 1.8.0. All buttons in the source list of the Journalist Interface ("Select All", "Select None", "Files and Messages" in the deletion flow, etc.) currently use title case. We should not selectively change it here, but rather agree on style guidelines, and then consistently apply them. We should reserve that discussion for 1.9.0 or later, IMO.

ninavizz:
> We use a mix of ALL CAPS and Title Case for button labels. Title case looks all right for Shorter Labels, but Can Start Looking Eccentric When Longer.

You make me giggle in the best of ways sometimes, John. :)

I totally agree with everyone on this thread. It's emerged over quite a period, that an absence of clear Style Guidelines has contributed to ongoing friction that would be really nice to alleviate by taking the time to commit to making some. 

Likewise, all the web UIs are horribly inconsistent; which I am as guilty of, as anyone, for wanting to implement best practices on spots where code is being modified—while leaving it well enough alone, elsewhere. I like the idea of doing a styleguide across a few sprints, then implementing its rules, in another sprint.

PR #6041, opened 2021-07-13
refactors Source Interface for semantic HTML5/ARIA markup

### LINKED ISSUE #5986: semantic (HTML5/ARIA) page structure in Source Interface

cfm:
## Status

**Ready for review**
- [x] Depends for merge on #6021

Stylistic changes suggested for rebasing after review and before merge:
1. [x] ["[Squash] each phase's commits into one"](https://github.com/freedomofpress/securedrop/issues/5986#issuecomment-877365101)
2. [x] Rebase from `develop`
3. [x] Lint (beyond `make lint`) for indentation of HTML in all `securedrop/source_templates/*.html`

## Description of Changes

Closes #5986 by:

1. **refactoring all `securedrop/source_templates/*.html` for semantic markup;**
2. **adding ARIA annotations where HTML5 semantics are insufficient** to make the Source Interface easily navigable by assistive tech, especially in the absence of JavaScript for changing state, polyfilling for undersupported HTML5 elements, etc.; and
3. **fixing SASS/CSS in light of (1)–(2),** especially to move decorative images out of the DOM and accessibility trees.

### Known issues

For future work:

1. #6056 (out of scope)
2. #6057 (out of scope)
3. Some pages need `h1`s (address in re: [WCAG 2.4.2](https://docs.google.com/spreadsheets/d/13lbFPR1Pm4QbcVz_7oVxMmKViqSl268CKA0ixoMyYxw/edit#gid=57993727&range=17:17))

## Testing

Click through the Source Interface and inspect in terms of:

1. ***Markup,* by inspecting and validating HTML.** This need not take place in Tor Browser; I used the [Web Developer extension](https://chrispederick.com/work/web-developer/) in Chrome.
2. ***Accessibility in assistive tech,* by listening to the interface as announced from Tor Browser by a screen-reader.**  It should sound and "feel" conspicuously more fluent.  (I also audited pages in [WebAIM's Wave Evaluation Tool](https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh) and [Deque's axe DevTools](https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd), but these will yield mixed results until the conclusion of this accessibility push.)
3. ***Visual appearance,* by comparing the interface against `develop`.**  Changes are intended to be "pixel-perfect" except where there were preexisting idiosyncrasies that more-semantic markup tended to make naturally more consistent.  Of course, suggestions for fine-tuning are welcome!

## Deployment

No special considerations.

- ~~I have written a test plan and validated it for this PR~~

Let me know if this is necessary.

    - freedomofpress/securedrop-docs#245


securedrop/source_templates/flashed.html
```diff
@@ -1,18 +1,15 @@
...
     {% if category != 'banner-warning' %}
-      <div class="flash {{ category }}">
-        {% if category == 'notification' %}
-        <img src="{{ url_for('static', filename='i/font-awesome/info-circle-black.png') }}" alt="{{ gettext('Notification') }} width="20" height="16">
-        {% elif category == 'error' %}
-        <img class="pull-left" src="{{ url_for('static', filename='i/font-awesome/exclamation-triangle-black.png') }}" alt="{{ gettext('Error') }}" width="20" height="17">
-        {% elif category == 'important' %}
-        <img src="{{ url_for('static', filename='i/bang-stop.png') }}" alt="{{ gettext('Important') }}" width="22" height="22">
-        {% endif %}
+      <li class="flash {{ category }}">
```
SaptakS: Doesn't moving all the images with alt tags to stylesheet, confuse the users using screen readers more whether the flash message is a Notification, Error or Important?

securedrop/source_templates/base.html
```diff
@@ -18,33 +18,33 @@
...
           {% else %}
 
           {% if 'logged_in' in session %}
-            <a href="{{ url_for('main.logout') }}" class="btn pull-right" id="logout">{{ gettext('LOG OUT') }}</a>
-          {% endif %}
-
-          {% if not new_user%}
-            <hr class="no-line">
+            <nav>
+              <a href="{{ url_for('main.logout') }}" class="btn pull-right" id="logout">{{ gettext('LOG OUT') }}</a>
```
SaptakS: I think not having "LOG OUT" in uppercase is a good idea from accessibility standpoint. There have been researches that dyslexic people find it more difficult to read uppercase letters. Also, a lot of screen readers will read this as "L.O.G. O.U.T.). Though I think this definitely needs design decisions and probably is beyond the scope of this PR.

Also applies for other uppercase texts.

securedrop/journalist_templates/_source_row.html
```diff
@@ -4,19 +4,19 @@
...
   <div class="designation">
     {% if source.star.starred %}
-      <button class="button-star starred" type="submit" formaction="{{ url_for('col.remove_star', filesystem_id=source.filesystem_id) }}">
-        <img src="{{ url_for('static', filename='icons/starred.png') }}" class="icon-drop icon-star-source" width="16" height="15" alt="{{ gettext('Un-star {designation}').format(designation=source.journalist_designation) }}">
+      <button class="button-star starred" type="submit" formaction="{{ url_for('col.remove_star', filesystem_id=source.filesystem_id) }}" aria-label="{{ gettext('Un-star {designation}').format(designation=source.journalist_designation) }}">
+        <img src="{{ url_for('static', filename='icons/starred.png') }}" class="icon-drop icon-star-source" width="16" height="15" alt="{{ gettext('{designation} is starred').format(designation=source.journalist_designation) }}">
       </button>
-      <input type="checkbox" name="cols_selected" value="{{ source.filesystem_id }}">
+      <input type="checkbox" name="cols_selected" value="{{ source.filesystem_id }}" aria-labelledby="starred-source-link-{{ loop_index }}">
```
SaptakS: I think it's a better idea to just use `<label for="">` tag instead of using aria-labelledby for all input elements at least.  Always preferable to use as less aria attributes as possible. https://www.w3.org/TR/aria-in-html/#firstrule

securedrop/source_templates/index.html
```diff
@@ -17,43 +17,59 @@
...
     <div class="content">
+    <main>
       <div class="grid">
         {% include 'flashed.html' %}
         {# The div `index-wrap` MUST be ordered this way so that the flexbox works on large and small screens
            See _source_index.sass for a more full understanding. #}
         <div class="index-wrap">
-          <div class="header">
-            <img src="{{ g.logo }}" alt="{{ g.organization_name }} | {{ gettext('Logo Image') }}" class="logo-image">
+          <header class="header">
```
SaptakS: the `<header>` and `<footer>` elements have no corresponding role if they are a descendant of the `<main>` tag. Is that intentional? Else maybe the main tag should start after `<header>` and end before `<footer>`

securedrop/source_templates/index.html
```diff
@@ -17,43 +17,59 @@
...
-              <h2 class="welcome-text">
+            <section class="index-column index-left" aria-labelledby="first-submission-heading">
+              <div class="index-column-top-container" aria-label="">
+              <h2 id="first-submission-heading" class="welcome-text">
                 {{ gettext('First submission') }}
               </h2>
               <p class="extended-welcome-text">{{ gettext('First time submitting to our SecureDrop? Start here.') }}</p>
               </div>
-              <div class="index-column-bottom-container">
+              <div class="index-column-bottom-container" aria-label="">
```
SaptakS: What's the purpose for a blank `aria-label` ?

securedrop/source_templates/locales.html
```diff
@@ -1,35 +1,32 @@
...
+  <h2 id="locale-menu-heading" hidden>
+    {{ gettext('Selected language')}}: {{ g.locales[g.localeinfo.id] }}
+  </h2>
+  <label id="locale-menu-trigger" for="menu-1-checkbox" class="menu__trigger" aria-hidden="true">
+    {{ g.locales[g.localeinfo.id] }}
   </label>
+  <input id="menu-1-checkbox" class="menu__checkbox visually-hidden" type="checkbox" role="button" aria-label="{{ gettext('Choose language') }}" aria-haspopup="menu" aria-controls="locales-menu" aria-pressed="true">
 
-  <ul class="menu__items">
+  <ul id="locales-menu" class="menu__items" role="menu" aria-labelledby="menu-1-checkbox">
```
SaptakS, line 12: Some day hopefully the browser coverage will be enough that we can just replace this with `<menu>`. Sadly not there yet. :)

securedrop/journalist_templates/_source_row.html
SaptakS: Also, I think in the label it should be more like "Select {{designation}} to perform actions" or something of that sorts which makes the action of the checkbox more clearly rather than just the designation. The user might not know what they are checking it for.

securedrop/source_templates/generate.html
```diff
@@ -3,35 +3,24 @@
...
-<p class="explanation">{{ gettext('This codename is what you will use in future visits to receive messages from our team in response to what you submit on the next screen.') }}</p>
+<p id="codename-explanation" class="explanation">{{ gettext('This codename is what you will use in future visits to receive messages from our team in response to what you submit on the next screen.') }}</p>
 
-<hr class="no-line">
-
-<div class="code">
-  <img class="pull-left" src="{{ url_for('static', filename='i/font-awesome/lock-black.png') }}" alt="" width="20" height="23">
-  <p id="codename" class="codename">{{ codename }}</p>
+<section class="code">
+  <label id="codename-label" for="codename" hidden>{{ gettext('Codename') }}</label>
```
SaptakS: This will be hidden even for screenreaders which I think is not the desired behaviour here? Might I suggest to add a `sr-only` class which is styled to something such that hides only in presentation. Something like this is standard:

```css
.sr-only {
  position:absolute;
  left:-10000px;
  top:auto;
  width:1px;
  height:1px;
  overflow:hidden;
}
```

securedrop/source_templates/lookup.html
```diff
@@ -1,31 +1,34 @@
...
+    {# ARIA-HIDDEN violates axe rule aria-hidden-focus, because we (a) want to
+    hide the superfluous "show"/"hide" links from screen-readers, (b) do not
+    want to remove these elements from sequential navigation with TABINDEX="-1",
+    and (c) do not have recourse in the Source Interface to scripting the
+    ARIA-HIDDEN value dynamically.  Cf. #6031. #}
+    <a id="codename-hint-show" class="show pull-right visible-codename" href="#codename-hint-visible" aria-hidden="true" tabindex="-1">{{ gettext('Show') }}</a>
     <div id="codename-hint-content" class="hidden-codename codename">
-      <a id="codename-hint-hide" class="pull-right" href="#codename-hint">{{ gettext('Hide') }}</a>
-      <p>{{ codename }}</p>
+      <a id="codename-hint-hide" class="pull-right" href="#codename-hint" aria-hidden="true">{{ gettext('Hide') }}</a>
```
SaptakS: You will need to add tabindex=-1 for this also, else this is tab navigable.

securedrop/source_templates/locales.html
```diff
@@ -1,35 +1,32 @@
...
-      <img class="menu__trigger-icon" src="/static/i/languages_arrow.png" alt="" width="12" height="18">
-    </div>
+<section class="menu" aria-labelledby="locale-menu-heading">
+  <h2 id="locale-menu-heading" hidden>
+    {{ gettext('Selected language')}}: {{ g.locales[g.localeinfo.id] }}
+  </h2>
+  <label id="locale-menu-trigger" for="menu-1-checkbox" class="menu__trigger" aria-hidden="true">
+    {{ g.locales[g.localeinfo.id] }}
   </label>
+  <input id="menu-1-checkbox" class="menu__checkbox visually-hidden" type="checkbox" role="button" aria-label="{{ gettext('Choose language') }}" aria-haspopup="menu" aria-controls="locales-menu" aria-pressed="true">
```
SaptakS: I think there are better ways to do HTML only dropdowns now compared to this checkbox hack. Firstly, you can use tabindex (or a focusable element like anchor) and :focus/:focus-within styles to design a dropdown (Refs: https://css-tricks.com/solved-with-css-dropdown-menus/). But there is an even smarter way, that is using `<details>` and `<summary` tag that has the accordion behaviour already. And an accordion is just few CSS away from a dropdown (https://css-tricks.com/using-details-for-menus-and-dialogs-is-an-interesting-idea/)

Honestly though, semantically, I think the first option makes more sense, but some have been using the second way. I understand this might be beyond the scope of this PR, but I feel this hacky solution should be changed soon.

securedrop/source_templates/lookup.html
SaptakS: Also, I would write the code using `<details>` and `<summary`, which is much more semantically nice and exactly made for this kind of purpose. https://css-tricks.com/quick-reminder-that-details-summary-is-the-easiest-way-ever-to-make-an-accordion/

The only issue would be that we can't swap the text "Show" and "Hide" that easily. So if that behaviour can be changed somehow which makes sense design wise as well (like just showing "Show/Hide"), maybe worth changing this to a simple `<details>` and `<summary>`. You can do it using the open attribute in details in CSS and adding content:after or something, but will be bit tricky.

securedrop/source_templates/lookup.html
```diff
@@ -53,65 +53,61 @@ <h2 class="headline">{{ gettext('Submit Messages') }}</h2>
...
+        <time id="timestamp-{{ reply.filename }}" class="date" title="{{ reply.date|rel_datetime_format }}" datetime="{{ reply.date }}" aria-hidden="true">{{ reply.date|rel_datetime_format(relative=True) }}</time>
         <form id="delete" class="message" method="post" action="{{ url_for('main.delete') }}" autocomplete="off">
           <input name="csrf_token" type="hidden" value="{{ csrf_token() }}">
           <input type="hidden" name="reply_filename" value="{{ reply.filename }}" autocomplete="off">
-          <a href="#confirm-delete-{{ reply.filename }}" class="delete">
-            <img class="icon off-hover" id="delete-reply-{{ reply.filename }}" src="{{ url_for('static', filename='i/trash-x-out.png') }}" alt="{{ gettext('Delete reply') }}" width="16" height="20">
-            <img class="icon on-hover" id="delete-reply-{{ reply.filename }}" src="{{ url_for('static', filename='i/trash-x-solid.png') }}" alt="{{ gettext('Delete reply') }}" width="16" height="20">
+          <a href="#confirm-delete-{{ reply.filename }}" id="delete-reply-{{ reply.filename }}" class="delete" aria-label="{{ gettext('Delete this reply') }}">
```
SaptakS: Might need to add a role="button" because technically this doesn't take to a different page and hence behaves like a button rather than a link.

SaptakS (commented):
This is amazing work @cfm !!! Love to see efforts towards making the HTML more semantic and accessible. I have left some small comments/questions that caught my eyes. Few are probably bigger changes which might need their own separate issues.

securedrop/source_templates/base.html
eloquence: I think the use of upper case in the UI is indeed a bigger design question; see also the similar #5821. If we do want to stick with all-caps UI elements, is there a way to instruct screenreaders to not read them out as abbreviations?

SaptakS: Completely agree that it's a bigger design question since uppercases are being used in other places as well, as mentioned in #5821 . If it's shown in UI as all-caps, then it's completely on the assistive technology how they will interpret that.

Technically, you can change the actual text in HTML to title case, and then use CSS `text-transform: uppercase` to convert it to uppercase, which works in some assitive techs, but for most it doesn't. The assistive techs try to be as close to what is showing in UI as possible. Now usually if it's a small word, and especially if there is an abbreviation with same characters, it will mostly read out as abbreviation for sure. But for bigger words, most probably they will read it fine. Some screen reader also have a setting where there is a change of tone to denote that they are reading out an all-caps part of the sentence. So it really depends on how the screen reader is implemented, but they try to be as close to the UI shown in text as possible.

But if you really really want absolutely the behaviour where you want all-caps in UI and title case in screen reader, one workaround is to add `aria-label="Log Out"`, though it's usually recommended to avoid doing that.

Some similar discussions that might help:
https://www.drupal.org/project/drupal/issues/2958239
https://groups.google.com/a/chromium.org/g/chromium-accessibility/c/enk1PBjEfRc/m/-npdRy3qyYcJ

cfm: Thanks for the context and research, @SaptakS!  The discussions you cite are among the sources I consulted in https://github.com/freedomofpress/securedrop/issues/6003#issuecomment-880898715, where I arrived at the same conclusion for the purpose of this PR, pending #5821.  It sounds like you would recommend erring on the side of including these `aria-label`s on all-caps elements; I'll include those changes (to this instance and elsewhere) in my revision based on your review.

securedrop/source_templates/flashed.html
cfm: This was an oversight on my part!  In c6cd5afac I propose to add the `category` label back in `aria-label`.  If you'd prefer to see this label in the DOM, we could instead do something like:

```diff
--- a/securedrop/source_templates/flashed.html
+++ b/securedrop/source_templates/flashed.html
@@ -1,10 +1,12 @@
 {% with messages = get_flashed_messages(with_categories=true) %}
 {% if messages %}
 <section id="flashed" role="alert">
+  <h2 class="visually-hidden">{{ gettext('Alerts') </h2>
   <ul>
   {% for category, message in messages %}
     {% if category != 'banner-warning' %}
-      <li class="flash {{ category }}" aria-label="{{ gettext(category) }}">
+      <li class="flash {{ category }}">
+        <h3 class="visually-hidden">{{ gettext(category) }}</h3>
         {{ message }}
       </li>
     {% endif %}
```

securedrop/source_templates/base.html
cfm: d6116a27 adds these `aria-label`s.

securedrop/journalist_templates/_source_row.html
cfm: Thanks for both of these suggestions—I agree.  Since these changes came from #6021, which focuses as much as possible just on `alt` and ARIA attributes, I'll file both of these suggestions in #5987, which I'll turn to after the current push on the Source Interface.

securedrop/source_templates/index.html
cfm: Thanks for catching this.  Refactored in adf55a15 to contain just the flexbox layout, which is also more consistent with the structure used elsewhere in `securedrop/source_templates/base.html`.

securedrop/source_templates/index.html
cfm: Thanks for catching this.  Fixed in 07cd655d.

securedrop/source_templates/locales.html
cfm: Yes!  I've opened #6031 to track this wishlist.  :-)

securedrop/source_templates/generate.html
cfm: I think I've addressed this concern in f9c8184b6, although somewhat differently than you suggest here: by labeling the containing `section` with an `h2` in place of this `label`.  This makes both narration and the document outline make more sense to me.  What do you think of this approach?

I wondered if the codename would be a good candidate for the `figure` element.  But `figure`'s semantics assume content ["that can be moved to another part of the document or to an appendix without affecting the main flow"](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure#usage_notes), which isn't the case here.

securedrop/source_templates/lookup.html
cfm: I had looked into a `details`/`summary` implementation of this widget and ruled it out—and now I can't reconstruct why.  I have a feeling I was overdoing the `aria-*` attributes in a way that interfered with my screen-reader's parsing of the native semantics and behavior.

So, I've added this properly, I think, in c8baf3d14.  As I note in the commit, I would have preferred to override the browser's native styling of the `::marker` pseudo-element, but its placement relative to the `summary` content depends on the language direction.  So this implementation uses `::{before,after}` as you suggest.  The downside is that even with `@media only screen` some screen-readers (including mine) will announce the `::{before,after}` `content` values (the `Show` and `Hide` labels) in addition to the automatic cues about the state of the `details` element.

securedrop/source_templates/locales.html
cfm: I agree—thank you for these suggestions!  I've noted the `details`-based approaches in #6031.  I do think that eliminating the checkbox hack from this menu is out of scope here, but I've opened #6048 for it.

securedrop/source_templates/lookup.html
cfm: Added in 753da7541.

cfm:
@SaptakS, thank you for your review last week!  I don't seem to be able to tag you for re-review, but I wanted to let you know that I've addressed and/or replied to all of your feedback.  Tomorrow (Wednesday) I will fix the failing tests.

cfm:
Re: tests:

1. Latest `staging-test-with-rebase` (after #6051) failed with an unrelated `debconf` error.  I'm rerunning the job in the hope that it's a flake.
2. `app-tests` continues to fail `test_{alembic,schema}_*`.  This puzzles me, because #6050 was introduced by #5974, which was not present in `develop` as of 0589168d0afe02e6ad82cf2fe72268a1e8ec7509.

I'll have to investigate (2) before merge, but since there are no database-related changes in this PR it needn't block review.  The alternative is to rebase from `develop` after review and catch #6051 to pass `app-tests` before merge.

cfm:
`app-tests` is finally green after CI shenanigans, so this is ready for final review prior to the pre-merge steps I outlined above (https://github.com/freedomofpress/securedrop/pull/6041#issue-689317434).

securedrop/source_templates/flashed.html
SaptakS: I think `aria-label` is fine.

securedrop/source_templates/generate.html
SaptakS: I think in this case, `section` with a `h2` labelling it makes sense semantically as well.

securedrop/source_templates/lookup.html
SaptakS: Yes, this semantic is a little problematic. The ideal solution would be something like:

```html
<p>Remember, you codename is:</p>
<details>
    <summary>Toggle visibility</summary>
    <output>{{ codename }} </output>
</details>
```
So, in that case no CSS hack would be needed. Or with CSS hack, instead of "Toggle visibility" I would have "Show" and then `[open] summary` I can hide, and `[open] summary:after` can have "Hide". And then also control the output's visibility to assistive tech via the `[open]` state. My reasoning kind of behind that was the codename shouldn't be announced all the time and the user using a screen reader should also be able to toggle between the functionality.

But I think this might be an entire separate discussion of UX and what the team thinks is ideal, so as long as this doesn't cause any regression in the screen-reader announcements, I think it's still much cleaner than the anchor hack. @cfm what do you think?

SaptakS (dismissed):
I am happy with the changes. I think the best way to implement the showing/hiding of codenames is a separate feature of it's own and can also be discussed as part of https://github.com/freedomofpress/securedrop/issues/6031. Might need some UX decisions and design decisions as well. So as long as that doesn't cause any regressions compared to before, I am happy with all the changes here.

securedrop/source_templates/lookup.html
cfm: Thanks, @SaptakS; I agree!  I've noted this too in #6031, including the dilemma of getting an explicit `open` state on the script-free Source Interface.

cfm:
@SaptakS, thank you for your second review and your approval!

@rmol, after your substantive review, I will (as indicated above):

> 1. ["[Squash] each phase's commits into one"](https://github.com/freedomofpress/securedrop/issues/5986#issuecomment-877365101)
> 2.  Rebase from `develop`

This will resolve the current conflict with `develop`.  (A GitHub-side merge from `develop` fails rebasing in CI; and a manual rebase now will break all the refs within this branch, which I want to keep intact *until* it's been squashed up.)

> 3. Lint (beyond `make lint`) for indentation of HTML in all `securedrop/source_templates/*.html`

securedrop/source_templates/generate.html
```diff
@@ -3,43 +3,32 @@
...
+<p id="codename-explanation" class="explanation">{{ gettext('This codename is what you will use in future visits to receive messages from our team in response to what you submit on the next screen.') }}</p>
 
-<hr class="no-line">
-
-<div class="code">
-  <img class="pull-left" src="{{ url_for('static', filename='i/font-awesome/lock-black.png') }}" alt="" width="20" height="23">
-  <p id="codename" class="codename">{{ codename }}</p>
+<section class="code" aria-labelledby="codename-heading">
+  <h2 id="codename-heading" class="visually-hidden">{{ gettext('Codename') }}</h2>
+  <output id="codename" class="codename" aria-describedby="codename-instructions codename-explanation">{{ codename }}</output>
```
rmol: In testing with Orca, the `<output>` isn't being read, or at least I can't figure out how to navigate to have it read. Perhaps it's being treated as a live region but isn't receiving any updates?

If I change it to `<code>`, it's read by default, and I can navigate to it.

This page has so many long-standing usability and accessibility problems it's hard not to dive into the rest, but I think just using a different element for the codename is enough for this PR.

securedrop/source_templates/lookup.html
```diff
@@ -1,31 +1,23 @@
...
-      <p>{{ codename }}</p>
-    </div>
-  </div>
-</div>
-<hr class="header-separator-high">
-{% else %}
-  <hr class="no-line">
+<details class="code-reminder pull-left" id="codename-hint">
+    <summary>{{ gettext('Remember, your codename is:') }}</summary>
+    <output class="codename">{{ codename }}</output>
```
rmol: Here again, I'm having trouble figuring out how to get the codename read in Orca.

securedrop/source_templates/lookup.html
```diff
@@ -53,65 +42,61 @@ <h2 class="headline">{{ gettext('Submit Messages') }}</h2>
...
+<section aria-labelledby="replies-heading">
+<h2 id="replies-heading" class="headline">{{ gettext('Read Replies') }}</h2>
 
 <div id="replies">
   {% if replies %}
-    <aside>
-      <p>{{ gettext("You have received a reply. To protect your identity in the unlikely event someone learns your codename, please delete all replies when you're done with them. This also lets us know that you are aware of our reply. You can respond by submitting new files and messages above.") }}</p>
-    </aside>
-    <hr class="no-line">
+    <output role="status">
```
rmol: Orca just really doesn't like `<output>`. I can't get this read either.

securedrop/source_templates/lookup.html
```diff
@@ -53,65 +42,61 @@ <h2 class="headline">{{ gettext('Submit Messages') }}</h2>
...
+            <span>{{ gettext('Delete') }}</span>
           </a>
-          <div id="confirm-delete-{{ reply.filename }}" class="confirm-prompt">
+          <dialog open id="confirm-delete-{{ reply.filename }}" class="confirm-prompt" aria-labelledby="delete-heading-{{ reply.filename }}">
+            <h3 id="delete-heading-{{ reply.filename }}" hidden>Delete reply from {{ reply.date|rel_datetime_format(relative=True) }}?</h3>
             <p>{{ gettext('Delete this reply?') }}
-              <a href="#delete">{{ gettext('Cancel') }}</a>
-              <button type="submit" class="danger" id="confirm-delete-reply-button-{{ reply.filename }}">{{ gettext('DELETE') }}</button>
+              <a href="#delete" role="button">{{ gettext('Cancel') }}</a>
+              <button type="submit" class="danger" id="confirm-delete-reply-button-{{ reply.filename }}" role="button" aria-label="{{ gettext('Delete') }}>{{ gettext('DELETE') }}</button>
```
rmol: The `aria-label` is missing a closing double quote, preventing replies from appearing on the page.

securedrop/source_templates/lookup.html
```diff
@@ -53,65 +42,61 @@ <h2 class="headline">{{ gettext('Submit Messages') }}</h2>
...
-    </aside>
-    <hr class="no-line">
+    <output role="status">
+      {{ gettext("You have received a reply. To protect your identity in the unlikely event someone learns your codename, please delete all replies when you're done with them. This also lets us know that you are aware of our reply. You can respond by submitting new files and messages above.") }}
+    </output>
     {% for reply in replies %}
-      <div class="reply">
-        <time class="date" title="{{ reply.date|rel_datetime_format }}" datetime="{{ reply.date }}">{{ reply.date|rel_datetime_format(relative=True) }}</time>
+      <article class="reply" aria-labelledby="timestamp-{{ reply.filename }}">
+        <time id="timestamp-{{ reply.filename }}" class="date" title="{{ reply.date|rel_datetime_format }}" datetime="{{ reply.date }}" aria-hidden="true">{{ reply.date|rel_datetime_format(relative=True) }}</time>
```
rmol: This is probably another Orca issue, but I can't get to the timestamps.

rmol (changes requested):
Maybe the `output` element is only a problem in Orca, and other screenreaders are processing its content, but in any case the semantics seem wrong to me, as from my reading it's intended for live updates, which aren't happening in the source interface. If I'm wrong about that and it's just Orca, that's still a problem, because if sources are following recommendations and using Tails, they're going to be using Orca, as far as I know.

Other than that, I found just one actual bug, on the lookup page, where an attribute with no closing quote is preventing replies from appearing.

rmol (dismissed):

rmol (changes requested):
I clicked on the wrong thing when I approved changes. :confounded:  I'm still requesting changes.

cfm:
@rmol in https://github.com/freedomofpress/securedrop/pull/6041#pullrequestreview-723661645:

> Maybe the `output` element is only a problem in Orca, and other screenreaders are processing its content, but in any case the semantics seem wrong to me, as from my reading it's intended for live updates, which aren't happening in the source interface. If I'm wrong about that and it's just Orca, that's still a problem, because if sources are following recommendations and using Tails, they're going to be using Orca, as far as I know.

Alas.  In [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119) terms, I read both the [spec](https://html.spec.whatwg.org/multipage/form-elements.html#the-output-element) and [guidelines](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/output) to say that `output` MAY contain live updates, which the browser/screen-reader SHOULD announce—rather than MUST contain such updates.  (And, indeed, if the Source Interface did use JavaScript, it's easy to imagine how `/generate`'s "Refresh" button would update such a live region.)  But if some ["browser / screen reader pairings may not announce or make available the accessible name of the output, even though its a labelable element"](https://www.scottohara.me/blog/2019/07/10/the-output-element.html), then the semantics don't get to matter very much in practice.  :-(

Ah, what to do.  The *code*name isn't really "computer `code`"; it's not `samp`le output; it's not (yet) `k`ey`b`oar`d` input.  It *is* an `output`, I think, but I'll get over it.  ;-)  Back to `code`, I guess!  Revision forthcoming tonight.

rmol:
Doesn't have to be `<code>`, that was just what I tried to keep the look. Maybe `<mark>`, since it "[represents text which is marked or highlighted for reference or notation purposes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark)"? We could just go with `<pre>` or `<span>` and call it a day. :laughing:

securedrop/source_templates/lookup.html
cfm: Nope; this was me being overzealous in my ARIA annotations because macOS VoiceOver was being overzealous in its narration.  ;-)  Simplified and streamlined, in both markup and narration, in 08cce2cf6f85c8319c57c07dea3373b8f1ccff1a.

securedrop/source_templates/generate.html
cfm: @rmol, can you confirm that Orca correctly reads these as `mark`s as of 245689f7efabbe8e5d218916e655343b96ff68f5?  If so, then this and the following threads can be resolved.

securedrop/source_templates/lookup.html
rmol: After wrapping these in `h3`s, Orca is reading the timestamps.

securedrop/source_templates/generate.html
rmol: Yes, the `mark`s  are being read.

securedrop/source_templates/lookup.html
rmol: The `p` elements are read.

rmol (dismissed):

rmol:
The changes of `output` to `mark` or `p` worked, as did the fix for the replies section. Once the merge conflicts are resolved and CI is passing, I think this is good to go. Nice work, @cfm.

cfm:
Thanks, @rmol.  CI is running now with everything squashed down to one commit per phase of the refactoring.  Feel free to pop off the 8f2b503 head if you don't like this formatting.

zenmonkeykstop (approved):
LGTM, thanks to @cfm for the work and for your patience!

PR #6058, opened 2021-08-10
Make <time datetime= into valid HTML by using a custom jinja tag

### LINKED ISSUE #6057: invalid `datetime` attributes
## Description

In `securedrop/source_templates/lookup.html`, `time` elements have invalid `datetime` attributes giving too many digits for fractional seconds.

## Steps to Reproduce

Validate the HTML of the Source Interface's `/lookup` view.

## Expected Behavior

The page validates.

## Actual Behavior

The page fails validation with errors like:

> Bad value `2021-07-13 22:47:04.564322` for attribute `datetime` on element `time`: A fraction of a second must be one, two, or three digits

## Comments

Should be simple to fix with a truncating Jinja filter.

anorthall:
## Status

Ready for review 

## Description of Changes

Fixes #6057

Change the direct call to `__str__()` into a call to `strftime()` to meet the requirement of not having more than 3 decimal places for seconds on HTML. Now, no decimal places are displayed. I decided to implement this instead of the truncate filter as you cannot truncate the datetime object without first converting it to a string.

## Testing

Open the website and validate the HTML.

## Deployment

No special measures required.

## Checklist

### DIFF
--- a/securedrop/bin/run-test
+++ b/securedrop/bin/run-test
@@ -11,6 +11,9 @@ cd "${REPOROOT}/securedrop"
 PYTHONPATH=".:${PYTHONPATH:-}"
 export PYTHONPATH
 
+TEST_RESULTS=../test-results
+JUNIT_XML=${TEST_RESULTS}/junit.xml
+
 source "${BASH_SOURCE%/*}/dev-deps"
 
 run_xvfb
@@ -25,7 +28,12 @@ maybe_create_config_py
 if [ -n "${CIRCLE_BRANCH:-}" ] ; then
     touch tests/log/firefox.log
     function finish {
-        cp tests/log/firefox.log ../test-results
+        # Rewrite the TESTCASE[file] attribute in $JUNIT_XML relative to our
+        # working directory so that CircleCI can match timings to files on the
+        # next job.
+        sed -i 's/file="/file="tests\//g' "$JUNIT_XML"
+
+        cp tests/log/firefox.log "$TEST_RESULTS"
         tmpdir=$(mktemp -d -t codecov-XXXX)
         curl -s https://codecov.io/bash > "$tmpdir"/codecov;
         VERSION="$(curl --silent "https://api.github.com/repos/codecov/codecov-bash/releases/latest" | grep '"tag_name":' |sed -E 's/.*"([^"]+)".*/\1/')"
@@ -37,7 +45,7 @@ if [ -n "${CIRCLE_BRANCH:-}" ] ; then
     trap finish EXIT
 fi
 
-mkdir -p "../test-results"
+mkdir -p "$TEST_RESULTS"
 
 : "${TEST_LOCALES:="ar en_US"}"
 export TEST_LOCALES
@@ -50,10 +58,11 @@ pytest \
     --no-success-flaky-report \
     --page-layout \
     --durations 10 \
-    --junitxml=../test-results/junit.xml \
+    --junitxml "$JUNIT_XML" \
+    --override-ini junit_family=xunit1 \
     --cov-report term-missing \
-    --cov-report html:../test-results/cov_html \
-    --cov-report xml:../test-results/cov.xml \
-    --cov-report annotate:../test-results/cov_annotate \
+    --cov-report "html:${TEST_RESULTS}/cov_html" \
+    --cov-report "xml:${TEST_RESULTS}/cov.xml" \
+    --cov-report "annotate:${TEST_RESULTS}/cov_annotate" \
     --cov=. \
     "$@"


cfm:
Thank you, @anorthall!  Just a heads-up that this will likely require a rebase after #6041, which moves around the `time` elements you fix here.

anorthall:
> Thank you, @anorthall! Just a heads-up that this will likely require a rebase after #6041, which moves around the `time` elements you fix here.

No problem.

zenmonkeykstop:
Hey, @anorthall, the fix using `strftime()` looks good - one thing it doesn't account for (and this is an existing problem, not one you added) is formatting of the time values to different locales. If you want to try to support that, the approach I'd suggest would be to:

a) use [flask_babel](https://flask-babel.tkte.ch/#datetime-functions) to do the date formatting
b) do the formatting in a custom jinja filter, to make it reusable across the application if necessary (I think there's only one other instance of `<time />`

In either case, thanks for the PR!

anorthall:
> Hey, @anorthall, the fix using `strftime()` looks good - one thing it doesn't account for (and this is an existing problem, not one you added) is formatting of the time values to different locales. 

Thanks for the tip/info - I will work on this but not until later tomorrow/Thursday at the latest.

zenmonkeykstop:
> Thanks for the tip/info - I will work on this but not until later tomorrow/Thursday at the latest.

No problem, and no rush!

anorthall:
Hopefully that latest PR will be a better solution. As you can see I added the template tag (`html_datetime_format`), and I also updated the other location that `<time>` was used, in the journalist portal.

All tested and working, and the HTML validates.

securedrop/template_filters.py
```diff
@@ -49,3 +49,15 @@ def filesizeformat(value: int) -> str:
...
+
+def html_datetime_format(dt: datetime) -> str:
+    """
+    Return a datetime string that will pass HTML validation and be
+    adjusted for the application locale
+    """
+    return dates.format_datetime(
+        dt,
+        'yyyy-MM-dd HH:mm:ss.SSS',
+        locale=get_locale()
```
cfm: Thanks for adding this.  @zenmonkeykstop and I chatted and realized that the `locale` argument here is superfluous, since however the timestamp is localized will be overridden by the explicit pattern required for the `time[datetime]` attribute.  What needs localizing is rather the `time[title]` attribute—which I'll file as a separate issue.

Even if we don't need the `locale`, however, it's nice to have this reusable in a Jinja filter.  :-)  Thanks for taking the time to refactor!

cfm (changes requested):
## Test plan

- [x] Open the website and validate the HTML.

In addition:

- [x] Specifically, there are no more `Bad value` errors on the `time[datetime]` attributes.
- [x] All instances of the pattern `%Y-%m-%d %H:%M:%S%Z` have been replaced by the `html_format_datetime` filter.

Looks good to me, @anorthall!  I've left a comment about the localization question, which I've clarified with @zenmonkeykstop; fortunately the solution is just a deletion here and another issue I'll open separately.  :-)

Given that there are a couple of iterations recorded in your branch, it would be ideal if you could squash your commits before merge.  Or, if you prefer, @zenmonkeykstop can take care of that as the maintainer who will ultimately approve and merge.

securedrop/template_filters.py
anorthall: Thanks for your time, @cfm, just to clarify- are you suggesting that you'd like to:

1. Leave the code as is and squash ready for merge, despite `locale` not being required? or
2. Remove the `locale` argument from the code but keep the custom filter and squash ready for merge

Not quite sure if you are suggesting that it's nice to have the `locale`, or nice to have the filter, without the `locale`. :)

cfm: Ah, yes: (2), please.  I meant to acknowledge that, even though the `locale` argument was the original motivation for asking you to refactor from `strftime` to a filter, it still makes sense to keep this date format in filter form along with the others.  Thanks for clarifying.  :-)

anorthall:
@cfm Changes made and squashed

I note the `updater-gui-tests` fail is nothing to do with my PR.

cfm (approved):
Thanks, @anorthall!  I confirm that `updater-gui-tests` is failing without #6069, but since it's irrelevant to your changes here I agree it's not worth rebasing for it.

Tagging @zenmonkeykstop for final approval and merge.

zenmonkeykstop:
Apologies @anorthall, a rebase gone awry reverted the changes in the PR - I'll fix it momentarily!

zenmonkeykstop:
Hey @anorthall - I've recovered your changes and resubmitted them in PR #6075 rebased against the `origin` develop (so CI should pass). Sorry for the mix-up - it looks like the rebase simply overwrote your `develop` branch with the `origin` one, the simplest path to recovery was to grab your commit from the reflog and apply them to a new (non-`develop`) branch on origin.

PR #6075, opened 2021-08-25
  Make <time datetime= into valid HTML by using a custom jinja tag

### LINKED ISSUE #6057: invalid `datetime` attributes

zenmonkeykstop:
(Resubmitting @anorthall's changes from #6058, recovered after a broken rebase)

## Status

Ready for review 

## Description of Changes

Fixes #6057

Change the direct call to `__str__()` into a call to `strftime()` to meet the requirement of not having more than 3 decimal places for seconds on HTML. Now, no decimal places are displayed. I decided to implement this instead of the truncate filter as you cannot truncate the datetime object without first converting it to a string.

## Testing

Open the website and validate the HTML.

## Deployment

No special measures required.

## Checklist

### DIFF
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -115,6 +115,8 @@ def create_app(config: 'SDConfig') -> Flask:
     app.jinja_env.filters['rel_datetime_format'] = \
         template_filters.rel_datetime_format
     app.jinja_env.filters['filesizeformat'] = template_filters.filesizeformat
+    app.jinja_env.filters['html_datetime_format'] = \
+        template_filters.html_datetime_format
 
     @app.before_first_request
     def expire_blacklisted_tokens() -> None:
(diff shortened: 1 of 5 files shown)


zenmonkeykstop (commented):
HTML validation passes when a reply is present, LGTM! Thanks @anorthall!

conorsch (approved):
Looks good, thanks for juggling the commits @zenmonkeykstop.

Issue #5336, opened 2020-06-24
Narrow branch target for staging-test-with-rebase, nightly runs


eloquence:
`staging-test-with-rebase` often takes >30 minutes or flakes out entirely, which significantly slows down normal development. Unless we can very significantly speed it up, we should consider:
- making it nightly and reinstating a build manager role for a given release cycle who is on duty to monitor/escalate that these aren't failing during dev
- continuing to running it on all `release/` branch PRs, since those really should always pass

eloquence:
Deferring this for now per sprint planning, we'll see how the job performs with an updated base image. May revisit mid-sprint if too many PRs get stalled by repeated CI fails again. If we do implement this, we agreed it would be great to add Slack alerting for nightly job fails, to raise visibility.

eloquence:
Example of recent failure, after 48m: https://app.circleci.com/pipelines/github/freedomofpress/securedrop/524/workflows/0a304b9b-e1a6-4db2-a1ed-de26425b5967/jobs/42275/steps

eloquence:
I'm seeing #5401 a lot lately, often requiring multiple kicks to resolve.

cfm:
This change is drafted in #6063.  @zenmonkeykstop, did you have any additional alerting in mind for failures of the nightly workflow?  I don't think I have access to that configuration (pending freedomofpress/infrastructure#3225).

cfm:
For this sprint, after #6063, @zenmonkeykstop will monitor nightly `staging-test-with-rebase` runs and evaluate whether additional alerting will be worthwhile.

PR #6086, opened 2021-08-28
fix: Use Circle ci branch filter and remove master-ci-branch.sh

### LINKED ISSUE #6024: Use CircleCI's built-in branch filtering feature throughout
CircleCI supports [branch filters with regular expressions](https://circleci.com/docs/2.0/configuration-reference/#jobfilters). We're currently using the script [`match-ci-branch.sh`](https://github.com/freedomofpress/securedrop/blob/develop/devops/scripts/match-ci-branch.sh) for our branch filter logic.

We should evaluate if there are still reasons to use this script, and if not, replace it with CircleCI's built-in functionality, for performance and maintainability reasons.

prateekj117:
## Status

Ready for review

## Description of Changes

Fixes #6024 

Circle CI supports [branch filters with regular expressions](https://circleci.com/docs/2.0/configuration-reference/#jobfilters). Right now, we are using both the [master-ci-branch.sh](https://github.com/freedomofpress/securedrop/blob/develop/devops/scripts/match-ci-branch.sh) as well as circle CI filters. This PR removes the usage of the script with circle CI filters throughout. 

Choose one of the following:


.circleci/config.yml
```diff
@@ -407,6 +397,18 @@ workflows:
               only:
                 - develop
     jobs:
-      - deb-tests
-      - translation-tests
-      - fetch-tor-debs
+      - deb-tests:
```
prateekj117: @eloquence I am not sure if these changes are valid as these jobs only run on develop branch weekly, and `deb-tests` and `translation-tests` wouldn't run by these filters. But I made these changes as earlier while using the `match-ci-branch.sh` script, these filters used to apply anyway, and I was not sure if we wanted a change in that behavior or not.

.circleci/config.yml
```diff
@@ -368,6 +352,8 @@ workflows:
       - staging-test-with-rebase:
           filters:
             branches:
+              ignore:
+                  - /i18n-.*/
               only:
                 - /release\/.*/
```
cfm: I appreciate your including the `ignore` list here, as both (a) a record of past behavior and (b) an indicator of what the desired behavior would be without the `only` list taking precedence.  My only concern is that this may not be obvious to folks reading this file for the first time without knowledge of this history (#5336) , who may be tempted to remove the `ignore` list in a way that's confusing down the line.  Would you mind adding a comment to this effect?

.circleci/config.yml
cfm: I agree that the `branches` specifications can be removed for jobs under the `weekly` workflow too.  Now that [CircleCI has deprecated `branches` specifications at the job level](https://circleci.com/docs/2.0/configuration-reference/#branches-deprecated), these would only be relevant if these jobs were added to the `securedrop_ci` or some other per-branch workflow.  The `weekly` configuration will be much simpler without them!

.circleci/config.yml
cfm: Nit:  Different whitespace on lines 356 versus 358.  If you like, `ignore` and `only` could each be just [a string rather than a single-item list of strings](https://circleci.com/docs/2.0/configuration-reference/#branches), though both conventions are used in this file.  :-)

.circleci/config.yml
```diff
@@ -395,7 +381,11 @@ workflows:
               only:
                 - develop
     jobs:
-      - staging-test-with-rebase
+      - staging-test-with-rebase:
+          filters:
+            branches:
+              ignore:
+                  - /i18n-.*/
```
cfm: Since the `nightly` workflow runs only on `develop`, I don't think it's necessary to ignore any branches for the `staging-test-with-rebase` job as part of that workflow.

cfm (changes requested):
Thanks, @prateekj117!  I've reviewed the diff and agree with how you've translated the `match-ci-branch.sh` calls into CircleCI's `branches` configuration.

- [x] Tests ran on this branch as expected with this CircleCI configuration.

I've left a few comments and nits for the sake of simplicity and readability, but then this looks good for a final maintainer's review!

eloquence:
@prateekj117 Just touching base, do you have time to continue your work on this?

prateekj117:
@eloquence Just caught up in some office work. Will do these changes in a day or two.

prateekj117:
@eloquence @cfm Ready for another review.

cfm (dismissed):
Thanks for revising, @prateekj117!  Looks correct and tidy to me.  The CI failure in `app-page-layout-tests` is due to #6104 and unrelated to these changes.

Approved for final maintainer's review.

conorsch (approved):
Rebased on top of latest develop. Fine to merge once CI passes!

PR #6087, opened 2021-08-29
Move CryptoUtil.display_id() out of CryptoUtil to fix type-checking 

nabla-c0d3:
## Status

Ready.

## Description of Changes

This small PR refactors `CryptoUtil.display_id()` and moves it out of `CryptoUtil` in order to:

* Enable/fix type-checking  for `display_id()` for #5599 .
* Rename display_id to journalist_designation to clarify what it is.
* Simplify the code for creating a new source.
* Make it easier to add support for journalist designations in other languages (in the future).

I tested this in the dev environment and checked that journalist designations are properly generated.

### DIFF
--- a/securedrop/crypto_util.py
+++ b/securedrop/crypto_util.py
@@ -5,9 +5,7 @@ from typing import Optional
 
 import pretty_bad_protocol as gnupg
 import os
-import io
 import re
-from random import SystemRandom
 
 from datetime import date
 from flask import current_app
@@ -19,8 +17,6 @@ from typing import Dict
 
 import rm
 
-from models import Source
-
 
 # monkey patch to work with Focal gnupg.
 # https://github.com/isislovecruft/python-gnupg/issues/250
@@ -30,15 +26,6 @@ gnupg._parsers.Verify.TRUST_LEVELS["DECRYPTION_COMPLIANCE_MODE"] = 23
 
 # to fix GPG error #78 on production
 os.environ['USERNAME'] = 'www-data'
-
-# SystemRandom sources from the system rand (e.g. urandom, CryptGenRandom, etc)
-# It supplies a CSPRNG but with an interface that supports methods like choice
-random = SystemRandom()
-
-# safe characters for every possible word in the wordlist includes capital
-# letters because codename hashes are base32-encoded with capital letters
-DICEWARE_SAFE_CHARS = (' !#%$&)(+*-1032547698;:=?@acbedgfihkjmlonqpsrutwvyxzA'
-                       'BCDEFGHIJKLMNOPQRSTUVWXYZ')
 
 
 def monkey_patch_delete_handle_status(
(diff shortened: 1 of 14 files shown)


securedrop/loaddata.py
```diff
@@ -116,7 +116,7 @@ def set_source_count(s: str) -> int:
 
 
 def add_journalist(
-    username: str = "",
+    username: str,
```
nabla-c0d3, line 119: `username` is now a required argument to simplify the logic here (and improve typing).

securedrop/crypto_util.py
```diff
@@ -84,8 +71,6 @@ class CryptoUtil:
 
     def __init__(self,
                  securedrop_root: str,
-                 nouns_file: str,
-                 adjectives_file: str,
```
nabla-c0d3, line 88: This was moved to the new `_DesignationGenerator` class.

securedrop/crypto_util.py
```diff
@@ -123,23 +102,6 @@ def do_runtime_tests(self) -> None:
...
-    def display_id(self) -> str:
-        """Generate random journalist_designation until we get an unused one"""
-
-        tries = 0
-
-        while tries < 50:
-            new_designation = ' '.join([random.choice(self.adjectives),
-                                        random.choice(self.nouns)])
-
-            collisions = Source.query.filter(Source.journalist_designation == new_designation)
```
nabla-c0d3, line 135: This was a bit problematic because it pulls a database session "out of thin air" (by doing `Source.query()`) instead of `database_session.query()`), so `display_id()` could only work when called from the correct context, which can get complicated (for example in unit tests).

securedrop/source_user.py
```diff
@@ -66,23 +67,41 @@ class SourceDesignationCollisionError(Exception):
...
     gpg_secret = scrypt_manager.derive_source_gpg_secret(source_passphrase)
 
-    # Create a designation for the source
-    try:
-        # TODO: Move display_id() out of CryptoUtil
-        journalist_designation = source_app_crypto_util.display_id()
-    except ValueError:
+    # Create a unique journalist designation for the source
+    # TODO: Add unique=True to models.Source.journalist_designation to enforce uniqueness
+    #  as the logic below has a race condition (time we check VS time when we add to the DB)
```
nabla-c0d3, line 80: I left a TODO here to not have a to do a DB migration. The race condition already existed in the previous code base. However, it may also not be a "real" problem.

securedrop/tests/test_crypto_util.py
```diff
@@ -275,16 +242,10 @@ def test_get_pubkey(source_app, test_source):
...
-    name=text(alphabet=crypto_util.DICEWARE_SAFE_CHARS),
-    secret=text(alphabet=crypto_util.DICEWARE_SAFE_CHARS),
-    message=text()
-)
+@given(message=text())
 def test_encrypt_then_decrypt_gives_same_result(
         source_app,
         test_source,
-        name,
-        secret,
```
nabla-c0d3, line 287: Not used by the test.

zenmonkeykstop (approved):
LGTM - also ran test suite against staging locally, ran `loaddata.py` adding ~1000 sources - no issues found with gnerated journalist designations.

PR #6095, opened 2021-09-08
adds diff-review decision tree to pull-request template

cfm:
## Status

Ready for review

## Description of Changes

Fixes my feeling that contributors should be given an explicit decision-tree for when dependency changes require diff reviews and when they do not.  :-)

## Testing

- [ ] List of `requirements.in` files that count as "production" is correct
- [ ] Instructions match project policy and maintainers' wishes


zenmonkeykstop (approved):
Visual review says LGTM

PR #6102, opened 2021-09-21
Tweak "Reply has been stored" message

### LINKED ISSUE #5828: 1.8.0 translation feedback: say replies are saved, not stored
## Description

From [AO](https://weblate.securedrop.org/translate/securedrop/securedrop/en/?checksum=a6f950dfa57047f3): "Your reply has been saved." might be better than "Your reply has been stored."

eloquence:
"Stored" is a very technical term, let's speak instead to the journalist<->source interaction. In context:

Screenshot from 2021-09-21 12-59-51

Resolves #5828

## Status

Ready for review

### DIFF
--- a/securedrop/journalist_app/main.py
+++ b/securedrop/journalist_app/main.py
@@ -163,7 +163,8 @@ def make_blueprint(config: SDConfig) -> Blueprint:
                     "<b>{}</b> {}".format(
                         # Translators: Precedes a message confirming the success of an operation.
                         escape(gettext("Success!")),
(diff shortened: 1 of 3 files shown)


eloquence:
(Originally I went with "The source will be able to read your reply", but after briefly discussing with @ninavizz, opted for the simpler "receive" to avoid speaking to the source's specific behavior or ability.)

zenmonkeykstop (approved):
LGTM based on visual review.

PR #6106, opened 2021-09-23
Support Debian 11 for dev env

conorsch:
## Status

Ready for review

## Description of Changes

Towards https://github.com/freedomofpress/securedrop-dev-docs/issues/9 , adding support for Debian 11 

While testing the dev env on Debian 11 Bullseye, was not able to install
the development requirements due to an error on pyenchant. Bumping the
req to a recent version resolved for me.

Marking as a draft because I haven't tested on Buster, and there may be more changes that should be appended here (as well as changes to the docs). 

## Testing
You must use a Debian 11 machine to test! Following the docs in https://docs.securedrop.org/en/latest/development/setup_development.html#install-python-requirements , make sure you see no errors (or report ones you do encounter!).

## Deployment
So far, dev-env only. 

## Checklist

I haven't performed a diff review, but this is explicitly a dev req.

### DIFF
--- a/securedrop/requirements/python3/develop-requirements.in
+++ b/securedrop/requirements/python3/develop-requirements.in
@@ -24,7 +24,7 @@ netaddr
 pip>=21.1
 pip-tools>=6.1.0
 psutil>=5.6.6
-pyenchant
+pyenchant>=3.2.1
 pylint>=2.7.0
 pynacl>=1.4.0
 py>=1.10.0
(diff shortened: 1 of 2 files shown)


conorsch:
@tesitura Please take this for a spin and see if it unblocks you!

kushaldas:
Tested on both Debian 10 and a fresh Debian 11 system. Works for me.  @conorsch +1 from me :)

conorsch:
Marking ready for review. @kushaldas tried it and liked it, and @tesitura was also able to spin up the VM environments on Bullseye, with this change in the mix. There are still some superficial docs updates required, e.g. mentioning testing Debian 11, but this seems to be all that's necessary diff-wise so far.

kushaldas (approved):
Worked as expected on both Debian 10 and Debian 11. Approved.

PR #6110, opened 2021-09-27
Repair auto-updates on Tails < 4.19 workstations 

### LINKED ISSUE #6098: Prompt users to fix upgrade issue for Tails 4.14 - 4.18 (or fix it for them?)
## Description

Tails automatic upgrades were broken between versions 4.14 - 4.18 inclusive ([Tails announcement](https://tails.boum.org/doc/upgrade/error/check/index.en.html#4.18)), and some users on older versions of Tails may not realize that they are missing auto-updates, and that there is a manual step required to fix them. 

We could:
- Make no code changes, and continue to use support messaging channels etc to remind folks of this issue
- Add text to the SecureDrop updater wizard, prompting users to update if their version of Tails is too old, or
- Perform the steps to fix automatic Tails updates ourselves, which (according to the above link) consist of 
```
torsocks curl --silent https://tails.boum.org/isrg-root-x1-cross-signed.pem \
| sudo tee --append /usr/local/etc/ssl/certs/tails.boum.org-CA.pem \
&& systemctl --user restart tails-upgrade-frontend
```  

I'm kind of in favour of the last option, and I can put in a PR for a check at the end of `securedrop_init` right before our GUI updater runs. What do others think?  [edit: filing now so we can discuss inclusion in 2.1.0]

rocodes:
## Status

Ready for review

## Description of Changes

Fixes #6098

Changes proposed in this pull request: Include check for pre-4.19 Tails versions in network hook. Attempt to repair auto-updates on Tails < 4.19 by applying the fix at https://tails.boum.org/news/version_4.18/, with the additional step of verifying the new CA file against existing files on the user's machine prior to installation.  

## Testing

**On an Admin or Journo Workstation running Tails < 4.19**
- Ensure network connectivity
- `cd ~/Persistent/securedrop` and check out this branch
- Run `./securedrop-admin --force tailsconfig` 
- [ ] `tailsconfig` completes successfully
- [ ] After about 30 seconds, the Tails updater GUI appears and prompts you to download the latest version of Tails  

**On Admin/Journo Workstation running Tails 4.19 or greater** (no op)
- Ensure network connectivity
- Check out this branch
- Run `./securedrop-admin --force tailsconfig` 
- [ ] `tailsconfig completes successfully`

(note: If you don't have an old Tails workstation and are QAing, these steps aren't recommended but are acceptable)
- Edit `/etc/os-release` to change the `TAILS_VERSION_ID` to a pre-4.19 version
- `touch /usr/local/etc/ssl/certs/tails.boum.org-CA.pem`
- Check out this branch, then run `./securedrop-admin --force tailsconfig`
- [ ] `tailsconfig` completes successfully
- [ ] `cat /usr/local/etc/ssl/certs/tails.boum.org-CA.pem` and verify that the isrg-root-x1-cross-signed cert was added to the file (https://tails.boum.org/isrg-root-x1-cross-signed.pem)  

## Deployment

2. New installs.

Choose one of the following:

### DIFF
--- a/install_files/ansible-base/roles/tails-config/files/securedrop_init.py
+++ b/install_files/ansible-base/roles/tails-config/files/securedrop_init.py
@@ -7,7 +7,8 @@ import pwd
 import sys
 import subprocess
 
-from shutil import copyfile
+import tempfile
+from shutil import copyfile, copyfileobj
 
 
 # check for root
@@ -148,3 +149,70 @@ flag_location = "/home/amnesia/Persistent/.securedrop/securedrop_update.flag"
 if b'Update needed' in output or os.path.exists(flag_location):
     # Start the SecureDrop updater GUI.
     subprocess.Popen(['python3', path_gui_updater], env=env)
+
+# Check for Tails < 4.19 and apply a fix to the auto-updater.
+# See https://tails.boum.org/news/version_4.18/
+# (Suggested removal: 2022/01)
+tails_4_min_version = 19
+needs_update = False
+tails_current_version = None
+
+with open('/etc/os-release') as file:
+    for line in file:
+        try:
+            k, v = line.strip().split("=")
+            if k == "TAILS_VERSION_ID":
+                tails_current_version = v.strip("\"").split(".")
+        except ValueError:
+            continue
+
+if tails_current_version:
+    try:
+        needs_update = (len(tails_current_version) >= 2 and
+                        int(tails_current_version[1]) < tails_4_min_version)
+
+    except (TypeError, ValueError):
+        sys.exit(0)  # Don't break tailsconfig trying to fix this
+
+    if needs_update:
+        cert_name = 'isrg-root-x1-cross-signed.pem'
+        pem_file = tempfile.NamedTemporaryFile(delete=True)
+
+        try:
+            subprocess.call(['torsocks', 'curl', '--silent',
+                             'https://tails.boum.org/' + cert_name],
+                            stdout=pem_file, env=env)
+
+            # Verify against /etc/ssl/certs/DST_Root_CA_X3.pem, which cross-signs
+            # the new LetsEncrypt cert but is expiring
+            verify_proc = subprocess.check_output(['openssl', 'verify',
+                                                   '-no_check_time', '-no-CApath',
+                                                   '-CAfile',
+                                                   '/etc/ssl/certs/DST_Root_CA_X3.pem',
+                                                   pem_file.name],
+                                                  universal_newlines=True, env=env)
+
+            if 'OK' in verify_proc:
+
+                # Updating the cert chain requires sudo privileges
+                os.setresgid(0, 0, -1)
+                os.setresuid(0, 0, -1)
+
+                with open('/usr/local/etc/ssl/certs/tails.boum.org-CA.pem', 'a') as chain:
+                    pem_file.seek(0)
+                    copyfileobj(pem_file, chain)
+
+                # As amnesia user, start updater GUI
+                os.setresgid(amnesia_gid, amnesia_gid, -1)
+                os.setresuid(amnesia_uid, amnesia_uid, -1)
+                restart_proc = subprocess.call(['systemctl', '--user', 'restart',
+                                                'tails-upgrade-frontend'], env=env)
+
+        except subprocess.CalledProcessError:
+            sys.exit(0)   # Don't break tailsconfig trying to fix this
+
+        except IOError:
+            sys.exit(0)
+
+        finally:
+            pem_file.close()


cfm:
@zenmonkeykstop asked me to take a look at this since I have some bandwidth to spare this week.  I've set up a Tails 4.16 admin workstation from which I'll have test results tomorrow.

install_files/ansible-base/roles/tails-config/files/securedrop_init.py
```diff
@@ -148,3 +150,71 @@
...
+                        (int(tails_current_version[0]) < tails_min_version[0]
+                        or int(tails_current_version[1]) < tails_min_version[1]))
+
+    except (TypeError, ValueError):
+        sys.exit('Error checking Tails version. Please visit tails.boum.org ' +
+                 'to ensure your version of Tails is up to date.')
+
+    if needs_update:
+        cert_name = 'isrg-root-x1-cross-signed.pem'
+        pem_file = tempfile.NamedTemporaryFile(delete=True)
```
cfm, line 179: Just one note on this draft to inform the test plan:  The `pem_file.name` to "Observe the presence of" under `/tmp/` will not be the value of `cert_name`.  That doesn't matter for production use, but if you want to keep this testing step you could do something like [`tempfile.NamedTemporaryFile(suffix=cert_name, ...)`](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile).

rocodes: Thanks - changing the test plan instead, since the auto-updater appearing is really the important part :)

cfm (dismissed):
Diff looks good and test plan checks out!  Here's a test from (arbitrarily) Tails 4.16:

## On an Admin or Journo Workstation running Tails < 4.19
```sh-session
amnesia@amnesia:~$ cat /etc/amnesia/version
4.16 - 20210222
28f284b2539478b0e6348396dc14232e59542728
live-build: 3.0.5+really+is+2.0.12-0.tails5
live-boot: 1:20170112
live-config: 5.20190519
```

1. Ensure network connectivity
2. `cd ~/Persistent/securedrop` and check out this branch

```sh-session
amnesia@amnesia:~/Persistent/securedrop$ git checkout rocodes/6098-old-tails-upgrade-fix
[...]
HEAD is now at 327e9df4b Include check for pre-4.19 Tails versions in network hook. Attempt to repair auto-updates on those systems
```

3. Run `./securedrop-admin --force tailsconfig`
4. [x] `tailsconfig` completes successfully
5. [x] After about 30 seconds, the Tails updater GUI appears and prompts you to download the latest version of Tails

*...upgrade ensues...*

## On Admin/Journo Workstation running Tails 4.19 or greater (no op)
```sh-session
amnesia@amnesia:~$ cat /etc/amnesia/version
4.22 - 20210906
2660dc56443bf5f29feddd5c6ae6cacaf4308837
live-build: 3.0.5+really+is+2.0.12-0.tails5
live-boot: 1:20170112
live-config: 5.20190519
```

1. Ensure network connectivity
2. Check out this branch
3. Run `./securedrop-admin --force tailsconfig`
4. [x] `tailsconfig` completes successfully

---

Approving for maintainer's final review.

install_files/ansible-base/roles/tails-config/files/securedrop_init.py
```diff
@@ -37,6 +38,7 @@
                                 'install_files/ansible-base/mon-ssh.auth_private')
 }
 path_onion_auth_dir = '/var/lib/tor/onion_auth'
+path_tails_version = '/etc/amnesia/version'
```
cfm: Nit:  As of 327e9df, this no longer appears to be used anywhere (in favor of a single string literal for [`/etc/os-release`](https://github.com/freedomofpress/securedrop/pull/6110/files#diff-ec0b152f6ccd6d8069cc40054ffadd599f8df6d8431d014632537d611d41a035R162)).

install_files/ansible-base/roles/tails-config/files/securedrop_init.py
```diff
@@ -148,3 +150,72 @@
...
+                    pem_file.seek(0)
+                    copyfileobj(pem_file, chain)
+
+                # As amnesia user, start updater GUI
+                os.setresgid(amnesia_gid, amnesia_gid, -1)
+                os.setresuid(amnesia_uid, amnesia_uid, -1)
+                restart_proc = subprocess.call(['systemctl', '--user', 'restart',
+                                                'tails-upgrade-frontend'], env=env)
+        except IOError:
+            sys.exit(0)  # Don't break tailsconfig trying to fix this
```
cfm: @eloquence points out that any of the `subprocess.call()`s here could return a `CalledProcessError` (subclass only of `Exception`) that would not be caught by `except IOError`.  Would it be worth broadening the `except` clause in the spirit of `Don't break tailsconfig trying to fix this`?   :-)

Extreme "first, do no harm" position to consider:  Is there *any* failure-mode in this logic where we want to risk crashing the network hook?  See top-level review comment for an example.

install_files/ansible-base/roles/tails-config/files/securedrop_init.py
```diff
@@ -148,3 +150,72 @@
...
+            # Verify against /etc/ssl/certs/DST_Root_CA_X3.pem, which cross-signs
+            # the new LetsEncrypt cert but is expiring
+            verify_proc = subprocess.check_output(['openssl', 'verify',
+                                                   '-no_check_time', '-no-CApath',
+                                                   '-CAfile',
+                                                   '/etc/ssl/certs/DST_Root_CA_X3.pem',
+                                                   pem_file.name],
+                                                  universal_newlines=True, env=env)
+
+            if 'OK' in verify_proc:
```
cfm, line 195: In my testing a non-`OK` output from `openssl verify` will have a nonzero return code and raise a `CalledProcessError`.  See next question about whether to catch these `CalledProcessError`s in the `except` clause.

install_files/ansible-base/roles/tails-config/files/securedrop_init.py
```diff
@@ -148,3 +150,72 @@
...
+            try:
+                k, v = line.strip().split("=")
+                if k == "TAILS_VERSION_ID":
+                    tails_current_version = v.strip("\"").split(".")
+            except ValueError:
+                continue
+
+except FileNotFoundError:
+    sys.exit('Error checking Tails version. Please visit https://tails.boum.org ' +
+             'to ensure your version of Tails is up to date.')
```
cfm: NB.  If `/etc/os-release` is missing, we'll never get here, because Ansible will error out first:

```
TASK [tails-config : Confirm host OS is Tails.] ********************************
    "assertion": "ansible_lsb.id == \"Tails\" or tails_os_string.matched",
    "changed": false,
    "evaluated_to": false,
    "msg": "SecureDrop requires Tails 4 or greater for workstation environments."
}
```

cfm (commented):
At @eloquence's urging, given the nature of this change and its deployment, I've done a little extra fuzz-testing to see if I could get this logic to break at all.  More than anything else, the case I present below is a way of asking how seriously we want to `[d]on't break tailsconfig trying to fix this`.  :-)

In addition to the intended cases, `tailsconfig` completes successfully:

- [x] with `/etc/os-release` contents pathologically from the future (Tails > 4.22)
- [x] with `/etc/os-release` contents pathetically from the past (Tails 3.x)—though it wouldn't actually work, of course!

The failure-mode I *was* able to induce involves deliberately breaking the `curl` call and seeing what happens afterward:

```patch
--- a/install_files/ansible-base/roles/tails-config/files/securedrop_init.py
+++ b/install_files/ansible-base/roles/tails-config/files/securedrop_init.py
@@ -187,7 +187,7 @@ if tails_current_version:
 
         try:
             subprocess.call(['torsocks', 'curl', '--silent',
-                             'https://tails.boum.org/' + cert_name],
+                             'https://tails.boum.org/x' + cert_name],
                             stdout=pem_file, env=env)
 
             # Verify against /etc/ssl/certs/DST_Root_CA_X3.pem, which cross-signs
```

The `curl` call "succeeds", swallowing the HTTP `404`; the `openssl` call returns exit code `2` with the error `unable to load certificate`; and the Ansible play halts.

@rocodes, I defer to your and the team's deep knowledge of Tails internals and how `securedrop-admin` both does and ought to interact with them—so I'm raising the questions in this follow-on review as discussion points rather than formally requesting changes.  Please use or disregard as you see fit!

install_files/ansible-base/roles/tails-config/files/securedrop_init.py
rocodes: Thanks, you're right and this is a big oversight - initially this was catching a CalledProcessError, but with some changes yesterday I was a bit hasty in my refactoring. I'll address this and other comments presently, thank you for the review.

install_files/ansible-base/roles/tails-config/files/securedrop_init.py
rocodes: Thank you - I have addressed this and all such in ef483e0

install_files/ansible-base/roles/tails-config/files/securedrop_init.py
rocodes: Thank you! Removed in ef483e0

rocodes:
@cfm Thanks for your thoughtful review. I've addressed your comments and also tested these changes with deliberate misconfigurations in all local filepath names, unavailable cert URL on the Tails website, etc., to ensure that `tailsconfig` runs are not disrupted.

cfm (approved):
Thanks for revising, @rocodes.  Looks great as of ef483e0.  In particular, in the [case in which (e.g.) `openssl verify` has failed](https://github.com/freedomofpress/securedrop/pull/6110#pullrequestreview-766070175):

1. [x]  `tailsconfig` completes successfully **even without**
2. ~After about 30 seconds, the Tails updater GUI appears and prompts you to download the latest version of Tails~
3. [x] **The network hook does not crash**

—as intended for the Ansible play to proceed even if this logic fails for any reason.

I'll flag this case (https://github.com/freedomofpress/securedrop/pull/6110#pullrequestreview-766070175) for inclusion in v2.1.0 QA.

Re-approving for final maintainer's review.

install_files/ansible-base/roles/tails-config/files/securedrop_init.py
```diff
@@ -148,3 +149,70 @@
 if b'Update needed' in output or os.path.exists(flag_location):
     # Start the SecureDrop updater GUI.
     subprocess.Popen(['python3', path_gui_updater], env=env)
```
cfm, line 151: @zenmonkeykstop observed this morning that the invocation of the SecureDrop updater here, *before* the new logic that follows checks for a Tails upgrade, means that we will be able to update this logic in a future SecureDrop release in the event that it breaks under a future Tails version.  (That's a mouthful, but I hope it's a clear statement of the *solution* to a risk we've worried about!)

zenmonkeykstop:
Thanks @rocodes for getting this in under the wire and @cfm for the timely review!

install_files/ansible-base/roles/tails-config/files/securedrop_init.py
rocodes: Yes, and it's also my hope that we can remove this entire section after a reasonable timeframe (I've suggested Jan 2022 in the comments)

PR #6117, opened 2021-10-04
Update Focal vagrant box in CI

conorsch:
## Status

Ready for review 
## Description of Changes

Built a new CI image for the staging tests. In this GCP image, the
Ubuntu Focal Vagrant VM has had its version bumped from
202105.25.0 -> 202107.28.0. The latest version, 202109.07.0, would not
install successfully via `vagrant box add`.

## Testing
Ideally we'd test it in CI first, but the new branch restrictions around the staging tests mean that's not a simple process anymore. I propose we merge optimistically and revert if problems show up. 

## Deployment

CI only.


conorsch:
> Ideally we'd test it in CI first, but the new branch restrictions around the staging tests mean that's not a simple process anymore

Now that #6120 is merged, I'm going to resubmit this PR with a branch named `stg-*` to use the new testing logic.

PR #6120, opened 2021-10-04
add apt upgrade to all packages prior to changes to sources lists

### LINKED ISSUE #6119: [QA] The apt-test HTTPS cert fails on older versions of Focal
## Description

Older versions of Focal installing packages from https://apt-test.freedom.press fail during apt-get update, with this error:

## Steps to Reproduce
Use this dockerfile to observe the issue.

<details>
<summary>Dockerfile
</summary>

```
# Intentionally using an old tag, to test old Focal packages
# See here for options: https://quay.io/repository/freedomofpress/sd-docker-builder-focal?tab=tags
# See related expiry info for LE HTTPS certs here:
# https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/
FROM quay.io/freedomofpress/sd-docker-builder-focal:2021_04_14

RUN apt-get update && apt-get install -y curl apt-transport-https gnupg2

# Uncomment the upgrade task below to see a working build!
# RUN apt-get upgrade -y

RUN echo "deb [arch=amd64] https://apt-test.freedom.press focal main" | tee /etc/apt/sources.list.d/fpf-apt-test-repo.list

RUN curl -s https://raw.githubusercontent.com/freedomofpress/securedrop/develop/install_files/ansible-base/roles/install-fpf-repo/files/apt-test-signing-key.pub | apt-key add -

# The following line fails, with:
#
#   9 0.711 Err:4 https://apt-test.freedom.press focal Release
#   9 0.711   Certificate verification failed: The certificate is NOT trusted.
#   The certificate chain uses expired certificate.  Could not handshake: Error in the certificate verification. [IP: 206.81.1.102 443]
#
RUN apt-get update
```
</details>

## Expected Behavior
Repo is configured, apt lists are updated, packages are installed just fine.

## Actual Behavior
After repo is configured, apt list update fails, with:

```
Certificate verification failed: The certificate is NOT trusted.
The certificate chain uses expired certificate.  
Could not handshake: Error in the certificate verification. [IP: 206.81.1.102 443]
```

## Comments

We believe the LE root cert transition is affecting us here https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/ @zenmonkeykstop first observed this issue over the weekend, motivating #6116 & #6117, and is now working on a fix that'll likely make it into 2.1.0~rc2 (#6103).

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Fixes #6119

- adds an explicit update to all base packages in the `prepare-server` role before proceeding with the SD install. This will ensure that the staging env is up-to-date, and should effectively be a no-op in production unless updated packages are made available in the time between the OS install and the securedrop install playbook run.
- modifies the branch filtering logic to run `staging-test-with-rebase` if the branch is prefixed with `release/` or `stg-`

## Testing

- [ ] CI completes successfully
- [ ] `staging-test-with-rebase` is run against this branch.

### DIFF
--- a/install_files/ansible-base/roles/prepare-servers/tasks/main.yml
+++ b/install_files/ansible-base/roles/prepare-servers/tasks/main.yml
@@ -50,10 +50,10 @@
   tags:
     - apt
 
-- name: Upgrade all packages
+- name: Ensure dist-upgrade before SecureDrop install
   apt:
-    name: "*"
-    state: latest
+    upgrade: dist
     update_cache: yes
   tags:
     - apt
+    - apt-upgrade
--- a/install_files/ansible-base/roles/prepare-servers/tasks/main.yml
+++ b/install_files/ansible-base/roles/prepare-servers/tasks/main.yml
@@ -49,3 +49,11 @@
     purge: yes
   tags:
     - apt
+
+- name: Upgrade all packages
+  apt:
+    name: "*"
+    state: latest
+    update_cache: yes
+  tags:
+    - apt
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -357,7 +357,7 @@ workflows:
             branches:
               # Ignore needs to be here explicitely as only clause introduced in PR #6086 might be removed afterwards.
               ignore: /i18n-.*/
-              only: /release\/.*/
+              only: /(stg-|release\/).*/
           requires:
             - lint
       - translation-tests:


install_files/ansible-base/roles/prepare-servers/tasks/main.yml
```diff
@@ -49,3 +49,11 @@
     purge: yes
   tags:
     - apt
+
+- name: Upgrade all packages
+  apt:
+    name: "*"
+    state: latest
+    update_cache: yes
```
conorsch, line 56: The more obvious way to do this would be 

```
apt:
  upgrade: dist
  update_cache: yes
```

which is what we do here: https://github.com/freedomofpress/securedrop/blob/c267329e5bad9c3396a5087b41275bb324fbf4f8/install_files/ansible-base/roles/common/tasks/apt_upgrade.yml#L45-L50, essentially. Mind updating to that syntax just to make it clear there's no apt magic happening with the wildcard?

zenmonkeykstop: I'm explicitly not doing a dist-upgrade because I don't want unexpected state changes in terms of which packages are installed/removed. But that may just be paranoia, I don't think there's a convincing rationale for it. Let's see how it behaves as a `dist`.

conorsch (approved):
Looks great, let's take it for a spin

conorsch:
> `staging-test-with-rebase` is run against this branch.

Visible here: https://app.circleci.com/pipelines/github/freedomofpress/securedrop/3051/workflows/fbd36025-5025-4141-b528-c1ccad8b098a/jobs/57035

PR #6160, opened 2021-11-07
Refactor encryption logic to fix type-checking and improve tests (CryptoUtil 3/3)

nabla-c0d3:
## Status

Ready.

## Description of Changes

This is the last PR for refactoring CryptoUtil and fixing type-checking, as described in #5599.

More specifically:

* It hides the low-level encryption and decryption methods (`encrypt()` and `decrypt()`) by making them private. Instead, only higher-level, public methods can be used such as `encrypt_source_file()`, `decrypt_journalist_reply()`, etc.
    * This makes it easier to understand the crypto-related code, and also to write/maintain it. Previously, it was easy to make a mistake when using `encrypt()` or `decrypt()`, for example by providing the wrong fingerprints.
* It makes the test suite for the crypto-related code a bit more comprehensive. For example:
    * It ensures that encrypted data can only be decrypted by the specific keys that were the intended recipients. This is what uncovered the issue described in #6174 .
    * It no longer imports the journalist secret key when setting up GPG for the test suite. Instead the secret key is imported (and then removed) in the specific tests that need it.


securedrop/encryption.py
```diff
@@ -0,0 +1,256 @@
...
+        # Ensure that the journalist public key has been previously imported in GPG
+        try:
+            self.get_journalist_public_key()
+        except GpgKeyNotFoundError:
+            available_keys = self._gpg.list_keys()
+            raise EnvironmentError(
+                f"The journalist public key with fingerprint {journalist_key_fingerprint}"
+                f" has not been imported into GPG. "
+                f"Keys currently configured in GPG: {available_keys}"
+            )
```
nabla-c0d3, line 119: This is a new check.

securedrop/encryption.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+class GpgDecryptError(Exception):
+    pass
+
+
+_default_encryption_mgr: Optional["EncryptionManager"] = None
+
+
+class EncryptionManager:
```
nabla-c0d3, line 70: Not sure about this name... open to changing it.

securedrop/journalist_app/__init__.py
```diff
@@ -72,16 +71,7 @@ def create_app(config: 'SDConfig') -> Flask:
...
-    app.storage = Storage(config.STORE_DIR,
-                          config.TEMP_DIR,
-                          config.JOURNALIST_KEY)
-
-    # TODO: Attaching a CryptoUtil dynamically like this disables all type checking (and
-    # breaks code analysis tools) for code that uses current_app.storage; it should be refactored
-    app.crypto_util = CryptoUtil(
-        securedrop_root=config.SECUREDROP_ROOT,
-        gpg_key_dir=config.GPG_KEY_DIR,
-    )
```
nabla-c0d3, line 84: The main goal of this PR: removing `app.crypto_util`.

securedrop/journalist_app/__init__.py
```diff
@@ -72,16 +71,7 @@ def create_app(config: 'SDConfig') -> Flask:
...
+    app.storage = Storage(config.STORE_DIR, config.TEMP_DIR)
```
nabla-c0d3, line 74: `Storage` no longer needs the `config.JOURNALIST_KEY`.

securedrop/source_app/main.py
```diff
@@ -133,9 +137,12 @@ def lookup(logged_in_source: SourceUser) -> str:
...
 
-        # Generate a keypair to encrypt replies from the journalist
-        if not current_app.crypto_util.get_fingerprint(logged_in_source.filesystem_id):
-            current_app.crypto_util.genkeypair(logged_in_source)
+        # If not done yet, generate a keypair to encrypt replies from the journalist
+        encryption_mgr = EncryptionManager.get_default()
+        try:
+            encryption_mgr.get_source_public_key(logged_in_source.filesystem_id)
+        except GpgKeyNotFoundError:
+            encryption_mgr.generate_source_key_pair(logged_in_source)
```
nabla-c0d3, line 145: Is there a reason for not generating a source's key pair on account creation? It would make this simpler.

securedrop/store.py
```diff
@@ -100,13 +103,15 @@ def __init__(self, storage_path: str, temp_dir: str, gpg_key: str) -> None:
...
-        self.__gpg_key = gpg_key
-
         # where files and directories are sent to be securely deleted
         self.__shredder_path = os.path.abspath(os.path.join(self.__storage_path, "../shredder"))
         if not os.path.exists(self.__shredder_path):
             os.makedirs(self.__shredder_path, mode=0o700)
 
+        # crash if we don't have a way to securely remove files
+        if not rm.check_secure_delete_capability():
+            raise AssertionError("Secure file deletion is not possible.")
```
nabla-c0d3, line 113: Moved from CryptoUtil to here.

securedrop/tests/test_encryption.py
```diff
@@ -0,0 +1,355 @@
...
+            for_source_user=source_user1, ciphertext_in=encrypted_reply
+        )
+        assert decrypted_reply
+        assert decrypted_reply == journalist_reply
+
+        # And source2 is NOT able to decrypt the reply
+        with pytest.raises(GpgDecryptError):
+            encryption_mgr.decrypt_journalist_reply(
+                for_source_user=source_user2, ciphertext_in=encrypted_reply
+            )
```
nabla-c0d3, line 267: Without `default-cache-ttl 0`, this part of the test would fail: source2 would be able to decrypt the reply.

securedrop/tests/conftest.py
```diff
@@ -110,32 +115,41 @@ def insecure_scrypt() -> Generator[None, None, None]:
...
-    with TemporaryDirectory() as tmp_gpg_dir_name:
-        tmp_gpg_dir = Path(tmp_gpg_dir_name)
+    tmp_gpg_dir = tmp_path_factory.mktemp("gpg_dir")
 
-        # GPG 2.1+ requires gpg-agent, see #4013
-        gpg_agent_config = tmp_gpg_dir / "gpg-agent.conf"
-        gpg_agent_config.write_text("allow-loopback-pinentry")
+    # GPG 2.1+ requires gpg-agent, see #4013
+    gpg_agent_config = tmp_gpg_dir / "gpg-agent.conf"
+    gpg_agent_config.write_text("allow-loopback-pinentry\ndefault-cache-ttl 0")
```
nabla-c0d3, line 129: Important change here: as I understand it, `default-cache-ttl 0` disables the caching of the GPG passphrases. 

This was needed because previously, the source's GPG passphrase was not required to call decryption/encryption operations if that passphrase was already in the cache.

I've put a comment below in `test_encrypt_and_decrypt_journalist_reply()` in `test_encryption.py` showing an example where a source would be able to decrypt using another source's key, thanks to passphrase caching.

We should discuss whether we want to keep `default-cache-ttl 0`, and if so how to enable it when running the app (ie. outside of the test suite).

install_files/securedrop-app-code/debian/postinst
```diff
@@ -132,10 +132,14 @@ case "$1" in
         if ! grep -qE '^allow-loopback-pinentry$' /var/lib/securedrop/keys/gpg-agent.conf; then
             echo allow-loopback-pinentry >> /var/lib/securedrop/keys/gpg-agent.conf
         fi
+        if ! grep -qE '^default-cache-ttl 0$' /var/lib/securedrop/keys/gpg-agent.conf; then
```
zenmonkeykstop: Good catch here @nabla-c0d3 - how'd you feel about this being split out into a separate PR in the interests of scope management? AFAICT nothing else in this one would require testing upgrade scenarios, and it would be good to get the bugfix in independently.

securedrop/encryption.py
```diff
@@ -0,0 +1,256 @@
...
+
+        # Ensure that the journalist public key has been previously imported in GPG
+        try:
+            self.get_journalist_public_key()
+        except GpgKeyNotFoundError:
+            available_keys = self._gpg.list_keys()
+            raise EnvironmentError(
+                f"The journalist public key with fingerprint {journalist_key_fingerprint}"
+                f" has not been imported into GPG. "
+                f"Keys currently configured in GPG: {available_keys}"
```
zenmonkeykstop: This last line will list reply keys in the logs, giving the number of active sources on the instance - said logs might be shared for troubleshooting purposes, so it might be better not to include it.

securedrop/source_app/main.py
zenmonkeykstop: Historically there were problems with system entropy being exhausted - see #1584. This was addressed initially by delaying key creation if an entropy check failed and allowing journalists to flag sources for reply (creating a reply key next time the source connected) in the rare occurrence when they didn't have a reply key already. Entropy is not a problem anymore so the flag-for-reply feature was removed altogether in #5954, but if there are still (very) old source accounts without reply keys, and the source logs in, create() won't be called, only lookup().

install_files/securedrop-app-code/debian/postinst
nabla-c0d3: Sounds good - https://github.com/freedomofpress/securedrop/pull/6174

securedrop/tests/conftest.py
```diff
@@ -121,26 +125,25 @@ def setup_gpg(gpg_dir: Path) -> str:
     journalist_public_key_path = Path(__file__).parent / "files" / "test_journalist_key.pub"
     journalist_public_key = journalist_public_key_path.read_text()
     journalist_key_fingerprint = gpg.import_keys(journalist_public_key).fingerprints[0]
-
-    # TODO(AD): Don't import the journalist secret key; will be removed in my next PR
-    journalist_secret_key_path = Path(__file__).parent / "files" / "test_journalist_key.sec"
-    journalist_secret_key = journalist_secret_key_path.read_text()
-    gpg.import_keys(journalist_secret_key)
-
```
nabla-c0d3: Important change here: the journalist's private/secret key is no longer imported by default.

Having the journalist secret key available in GPG was making the test suite behave differently compared to web app. For example, data in tests could get decrypted using the journalist secret key instead of the expected source key.

securedrop/tests/test_encryption.py
```diff
@@ -0,0 +1,355 @@
...
+
+@contextmanager
+def import_journalist_private_key(
+    encryption_mgr: EncryptionManager,
+) -> typing.Generator[None, None, None]:
+    """Import the journalist secret key so the encryption_mgr can decrypt data for the journalist.
+
+    The journalist secret key is removed at the end of this context manager in order to not impact
+    other decryption-related tests.
+    """
```
nabla-c0d3, line 331: `import_journalist_private_key()` can be used when the journalist secret key needs to be available for a specific test.

securedrop/tests/conftest.py
```diff
@@ -131,12 +133,12 @@ def setup_journalist_key_and_gpg_folder() -> Generator[Tuple[str, Path], None, N
...
         journalist_key_fingerprint = gpg.import_keys(journalist_public_key).fingerprints[0]
 
-        # TODO(AD): Don't import the journalist secret key; will be removed in my next PR
-        journalist_secret_key_path = Path(__file__).parent / "files" / "test_journalist_key.sec"
-        journalist_secret_key = journalist_secret_key_path.read_text()
-        gpg.import_keys(journalist_secret_key)
+        # Reduce source GPG key length to speed up tests at the expense of security
+        with mock.patch.object(
+            EncryptionManager, "GPG_KEY_LENGTH", PropertyMock(return_value=1024)
+        ):
```
nabla-c0d3, line 139: Moved from CryptoUtil to the test suite.

securedrop/tests/test_encryption.py
```diff
@@ -0,0 +1,355 @@
...
+        assert encrypted_message
+
+        # And the journalist is able to decrypt the message
+        with import_journalist_private_key(encryption_mgr):
+            decrypted_message = encryption_mgr._gpg.decrypt(encrypted_message).data
+        assert decrypted_message.decode() == message
+
+        # And the source or anyone else is NOT able to decrypt the message
+        # For GPG 2.1+, a non-null passphrase _must_ be passed to decrypt()
+        assert not encryption_mgr._gpg.decrypt(encrypted_message, passphrase="test 123").ok
```
nabla-c0d3, line 194: Example of a new test that ensures that only the right user/key can decrypt some data.

securedrop/tests/test_encryption.py
```diff
@@ -0,0 +1,351 @@
+import typing
```
nabla-c0d3, line 1: Coverage report:
```
---------- coverage: platform linux, python 3.8.10-final-0 -----------
Name                                   Stmts   Miss Branch BrPart  Cover
------------------------------------------------------------------------
encryption.py                            123      9     28      4    91%
```

nabla-c0d3:
I ran the tests locally and also manually tested the app (`make dev`) - didn't run into any problems.

zenmonkeykstop (approved):
- [x] no issues found in code review
- [x] tests pass in staging env
- [x] upgrade scenario is successful and tests pass against upgrade env
- [x] manual tests of affected functionality look good.

LGTM, thanks for this!

PR #6166, opened 2021-11-10
updated dev TBB version check to be more robust

### LINKED ISSUE #6163: The command 'make dev' fails to retrieve the latest Tor Browser version number
## Description

When running `make dev` for the first time, the Tor Browser download fails because an incorrect version number is retrieved.

## Steps to Reproduce

1. Clone this repo
2. Run `make dev`

**Note**: The error might not happen if you've successfully set up a dev environment before because that specific Docker layer may be cached.

This shouldn't be environment dependent, but it's confirmed on Ubuntu Focal.

## Expected Behavior

The command `make dev` succeeds. More specifically, the correct version number is stored in `TBB_VERSION`, so that the subsequent download URL is correct and doesn't lead to an `HTTP 404 Not Found` error.

## Actual Behavior

When I run `make dev` for the first time (hold that thought), I get the error:

```sh
The command '/bin/sh -c TBB_VERSION=[... pruned for brevity ...]' returned a non-zero code: 8
make: *** [Makefile:185: dev] Error 8 
```

## Comments

@creviera can confirm this behavior.

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

 - Updates the Tor Browser version check in the dev Dockerfile to return the value of the `appVersion` attribute instead of the first value matching `major.minor.patch` semantics. TBB version is currently 11.0 so the original check will return the ESR version instead, breaking `make dev`.
 - Removes the Xenial dev Dockerfile, as Xenial is no longer supported.

Fixes #6163 

## Testing

- wipe existing docker images with `docker system prune --all`
- [ ] Verify that `make dev` completes successfully
- [ ] Verify via inspection of the docker output that the TBB version is 11.0

## Deployment

Dev-only.


securedrop/dockerfiles/focal/python3/Dockerfile
```diff
@@ -33,7 +33,7 @@ ENV TOR_RELEASE_KEY_FINGERPRINT "EF6E286DDA85EA2A4BA7DE684E2C6E8793298290"
 RUN curl -s https://openpgpkey.torproject.org/.well-known/openpgpkey/torproject.org/hu/kounek7zrdx745qydx6p59t9mqjpuhdf | gpg2 --import -
 
 # Fetch latest TBB version (obtained from https://github.com/micahflee/torbrowser-launcher/blob/develop/torbrowser_launcher/common.py#L198) and install Tor Browser
-RUN TBB_VERSION=$(curl -s https://aus1.torproject.org/torbrowser/update_3/release/Linux_x86_64-gcc3/x/en-US | grep -oP '\d+\.\d+\.\d+' | head -1) && \
+RUN TBB_VERSION=$(curl -s https://aus1.torproject.org/torbrowser/update_3/release/Linux_x86_64-gcc3/x/en-US | grep -oP 'appVersion="\K[^"]*' | head -1) && \
```
gonzalo-bulnes, line 36: Likely don't need to `| head -1` @zenmonkeykstop ! Totally agree your regexp is more general 👍

gonzalo-bulnes:
I updated the description to make sure that #6163 gets closed when merging.

securedrop/dockerfiles/focal/python3/Dockerfile
conorsch: Slick use of `\K`! Leaving in the `head` because we never want a newline to slip in there, so treating it as defense.

conorsch (approved):
:+1:

PR #6168, opened 2021-11-15
Use rust v1.56.1 in app-code build process

### LINKED ISSUE #6158: Update Rust version in build container to 1.56.1 or greater
## Description

Current Rust version used in the build container while compiling the `cryptography` wheel is 1.52.1. As a result of https://blog.rust-lang.org/2021/11/01/cve-2021-42574.html this should be updated to 1.56.1 or greater.

The version is set in `molecule/builder-focal/Dockerfile` - after updating it there, a new builder image should be generated, tested locally (does `make build-debs` pass? For extra credit, do tests pass against a staging environment using the built debs?), and uploaded to the docker repo.

conorsch:
Towards #6158.

## Status

Ready for review, although still needs builder changes. 

## Description of Changes

Fixes #6158 

## Testing
Ideally, we'd test both building the debs and using them in a staging environment. I haven't done that yet, but I'm comfortable merging as-is and inspecting CI output from the nightly staging jobs to make sure there are no surprises. 

## Deployment
Logic is specifically to the build process for app code. Strictly speaking this change could have side-effects, but highly unlikely given it's a stable release.

### DIFF
--- a/molecule/builder-focal/Dockerfile
+++ b/molecule/builder-focal/Dockerfile
@@ -41,7 +41,7 @@ COPY aptpreferences.conf /etc/apt/preferences.d/ubuntu-groovy
 
 RUN apt-get update && apt-get install -y dh-virtualenv
 
-ENV RUST_VERSION 1.52.1
+ENV RUST_VERSION 1.56.1
 
 # Install Rust for building cryptography
 RUN TMPDIR=`mktemp -d` && cd ${TMPDIR} \


zenmonkeykstop (approved):
LGTM - builder update still required as a follow-up

conorsch:
@zenmonkeykstop Thanks for quick review. I was a bit too eager in opening this without also making the builder changes, but small changes are fine, too. I've amended the PR description not to close the issue, so we've still got it to track the builder updates.

PR #6169, opened 2021-11-15
Remove remnants of v2 configuration code

eloquence:
- devops clean script should remove v3, not v2, configuration files
- v2 tor_instances task is unused

## Status

Ready for review. I've tested:

- manually running the `rm` command in a checkout that was used to previously provision a staging env
- a fresh install via `securedrop-admin install` playbook run against prod VM with these changes in place
   - all expected onion services were created correctly
   - install completed without errors

### DIFF
--- a/devops/clean
+++ b/devops/clean
@@ -29,10 +29,11 @@ function remove_unwanted_files() {
     fi
 
     # Remove any Onion URL info from previous instances.
-    rm -vf install_files/ansible-base/app-source-ths \
-        install_files/ansible-base/app-ssh-aths \
-        install_files/ansible-base/app-journalist-aths \
-        install_files/ansible-base/mon-ssh-aths \
+    rm -vf install_files/ansible-base/app-journalist.auth_private \
+        install_files/ansible-base/app-sourcev3-ths \
+        install_files/ansible-base/app-ssh.auth_private \
+        install_files/ansible-base/mon-ssh.auth_private \
+        install_files/ansible-base/tor_v3_keys.json \
         build/*.deb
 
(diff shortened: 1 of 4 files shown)


gonzalo-bulnes (commented):
The removed lines make sense to me, I can see why we'd want to clean the files that are specific to onion services v3 and understand the line additions. I could find where those files are created, except for `install_files/ansible-base/tor_v3_keys.json`.

eloquence:
> I could find where those files are created, except for `install_files/ansible-base/tor_v3_keys.json`

v3 onion services use asymmetric encryption for authentication, so a keypair has to be created client-side, and only the public key portion is shared with the server. As I understand it, this is done with a `securedrop-admin` subcommand in the `tor-hidden-services` Ansible role:
https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/tor-hidden-services/tasks/configure_tor_hidden_services.yml#L49-L61

I've confirmed that the file is present after running `make staging`.

eloquence:
(While this is technically ready for review by anyone, it's a nice opportunity to use https://github.com/cfm/terraform-metal-securedrop-staging to run the `securedrop-admin` test myself in the next few days. That said, if anyone wants to give it a spin before then, feel free.)

eloquence:
I've tested these changes using https://github.com/cfm/terraform-metal-securedrop-staging with prod VMs; moving to ready.

conorsch (approved):
Found one more file to remove via local grepping: `install_files/ansible-base/roles/restrict-direct-access/templates/ths_config.j2`. Rebased and squashed. Ready to merge after CI passes (again).

PR #6174, opened 2021-11-20
Disable caching of GPG passphrases in the GPG agent (CryptoUtil 1/3)

nabla-c0d3:
## Status

Ready.

## Description of Changes

This PR disables the caching of GPG passphrases by adding `default-cache-ttl 0` to the GPG agent's config.

* This fixes a security issue where any GPG passphrase could be supplied to the GPG decryption code, as long as the corresponding decryption key's GPG passphrase was already in the GPG cache. The decryption code (more specifically the GPG binary) transparently retrieves from the cache the GPG passphrase for the right key, and then uses the key to decrypt the data. This happens regardless of what GPG passphrase was supplied by the caller.
* AFAIK this issue is currently unexploitable. Exploiting it would require another another bug in the securedrop server code that would allow a malicious source to "ask" the server to decrypt another source's files/replies. In this scenario, the malicious source would not need to provide/guess the other source's GPG passphrase: the server would transparently retrieve it from its cache and decrypt the content; what this PR fixes.
* This is part 1 of a set of 3 PRs to refactor the GPG code. 
    * Part 2: https://github.com/freedomofpress/securedrop/pull/6184
    * Part 3: https://github.com/freedomofpress/securedrop/pull/6160 . That PR contains unit tests to ensure that the correct GPG passphrase must supplied to the decryption code.

### DIFF
--- a/install_files/securedrop-app-code/debian/postinst
+++ b/install_files/securedrop-app-code/debian/postinst
@@ -132,10 +132,14 @@ case "$1" in
         if ! grep -qE '^allow-loopback-pinentry$' /var/lib/securedrop/keys/gpg-agent.conf; then
             echo allow-loopback-pinentry >> /var/lib/securedrop/keys/gpg-agent.conf
         fi
+        if ! grep -qE '^default-cache-ttl 0$' /var/lib/securedrop/keys/gpg-agent.conf; then
+            echo 'default-cache-ttl 0' >> /var/lib/securedrop/keys/gpg-agent.conf
+        fi
 
     else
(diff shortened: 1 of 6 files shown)


zenmonkeykstop (approved):
- Looks good based on visual review
- confirmed that `/var/lib/securedrop/keys/gpg-agent.conf` includes the new `default-cache-ttl 0` line on both fresh prod installs (using debs built from this branch) and in the upgrade scenario.
- Confirmed reply functionality is unaffected.

Issue #6182, opened 2021-11-21
Deprecate usage of X-XSS-Protection header


evilaliv3:
With this ticket I would like to propose the deprecation of the usage of the ```X-XSS-Protection``` header that is actually supported only by IE and Safari: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

The same protection is no implemented by our current implementation of the CSP policy that disables inline JavaScript.

Considering that we do not target any IE or Safari browser browser (and expecially that IE is dead and replaced by Edge featuring a Chome engine) we could safely remove the header that represents just an overhead.

This ticket follows the same evaluation applied to the GlobaLeaks project: https://github.com/globaleaks/GlobaLeaks/commit/2c83b156b83b47487385d047474d63302a31539d

zenmonkeykstop:
Closed by #6187

PR #6183, opened 2021-11-21
Deprecate usage of X-XSS-Protection header

evilaliv3:
This pull request implements proposal #6182

### DIFF
--- a/install_files/ansible-base/roles/app/templates/sites-available/focal/journalist.conf
+++ b/install_files/ansible-base/roles/app/templates/sites-available/focal/journalist.conf
@@ -24,8 +24,6 @@ Header onsuccess unset X-Frame-Options
 Header always set X-Frame-Options "DENY"
 Header onsuccess unset Referrer-Policy
 Header always set Referrer-Policy "no-referrer"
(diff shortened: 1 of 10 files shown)


zenmonkeykstop (changes requested):
Hi @evilaliv3  and thanks, this is a good cleanup catch. Couple of things to consider:

-  For most existing instances to pick up the change, something like a postint script will be needed, as the templates updated in the PR are only hit via the Ansible install playbook.
- There were some Xenial-specific configs updated in the PR that  have since been removed in `develop` - a rebase is probably a good idea - it will also reduce the scope of the change above a bit. 

I'm happy to cover off both points above if you want! It might also be worth combining your other header-related PRs, since the same considerations  apply to them. Just let me know.

evilaliv3:
Thank you @zenmonkeykstop 

Please feel free to combine my commits and to proceed with the ansible addition that for sure is an aspect that you control better.

Thank you so much.

zenmonkeykstop:
Closing in favour of #6187 - thanks again @evilaliv3 !

Issue #4332, opened 2019-04-10
GPG agent processes are not cleaned up during testing


heartsucker:
## Description

While running the test suite, the test suite calls many times to `gpg2` each with its own `gpg-home` directory. This causes many instances of `gpg-agent` to spawn.  These processes are not killed and by the end of a run of the test suite, there are over 80 processes that are left around. When the container exists, these processes are killed, but they take up some amount of memory and resources. This is annoying when using a dev shell and repeatedly running sections of the test suite as it will leave hundreds of processes around if they are not manually cleaned up.

## Steps to Reproduce

1. `./bin/dev-shell`
1. In the container, `./bin/run-test -vv tests/test_*.py`
1. After the tests complete, `pgrep gpg-agent | wc -l`

## Expected Behavior

The number of `gpg-agent` processes is 0 at the end of the tests.

## Actual Behavior

The number is over 80.

## Comments

The `pytest` fixture that calls `gpg2` should `yield` it's fixture object and then attempt to clean up the dangling process. An example of this is in #4312.

nabla-c0d3:
This is not happening anymore: at the end of the tests, no `gpg2` processes are running.

Might have been fixed as a side-effect of the changes in https://github.com/freedomofpress/securedrop/pull/5683 and https://github.com/freedomofpress/securedrop/pull/6184 , which reduce the number of `gpg2` processes spawned by the test suite to 1, and use `yield` in the GPG fixture.

PR #6184, opened 2021-11-22
Use the existing GPG fixture in the functional tests (CryptoUtil 2/3)

nabla-c0d3:
## Status

Ready. 

## Description of Changes

This PR updates the functional tests to use the same fixture for GPG as the other tests. It also updates the config fixture.

* This is a prerequisite for the refactoring of the GPG code in https://github.com/freedomofpress/securedrop/pull/6160 which uses the updated config fixture.
* This is a small step forward for https://github.com/freedomofpress/securedrop/issues/3836. More work is needed to refactor the functional tests and their fixtures.
* Because the GPG fixture that is now being used is session-scoped, it is only called once during the whole test suite, and only one GPG home directory is used for all the tests. This helps a little bit with https://github.com/freedomofpress/securedrop/issues/4332 (only 1 GPG process should get created instead of ~80).
* There is more work to be done with the config fixture itself, and I might take a look at it eventually. The biggest problem is that tests might either use the config from the `config()` fixture, or the config from the config.py file generated by `make test` right before the tests run.
    * This creates really tricky interactions and bugs when working on the test suite, because it's unclear which config is being used. It's also sometime unclear what config is needed for a specific test to succeed, and some tests also tend to mutate the config object (for example changing the time it takes for a session to expire).
    * This PR makes small changes to slightly improve this situation.

### DIFF
--- a/securedrop/manage.py
+++ b/securedrop/manage.py
@@ -21,20 +21,23 @@ from passphrases import PassphraseGenerator
 
 sys.path.insert(0, "/var/www/securedrop")  # noqa: E402
 
-import qrcode
-from sqlalchemy.orm.exc import NoResultFound
+import qrcode  # noqa: E402
+from sqlalchemy.orm.exc import NoResultFound  # noqa: E402
 
-os.environ['SECUREDROP_ENV'] = 'dev'  # noqa
 
-from db import db
-from models import (
+if not os.environ.get("SECUREDROP_ENV"):
+    os.environ['SECUREDROP_ENV'] = 'dev'  # noqa
+
+
+from db import db  # noqa: E402
+from models import (  # noqa: E402
     FirstOrLastNameError,
     InvalidUsernameException,
     Journalist,
 )
-from management import app_context, config
-from management.run import run
-from management.submissions import (
+from management import app_context, config  # noqa: E402
+from management.run import run  # noqa: E402
+from management.submissions import (  # noqa: E402
     add_check_db_disconnect_parser,
     add_check_fs_disconnect_parser,
     add_delete_db_disconnect_parser,
--- a/securedrop/tests/__init__.py
+++ b/securedrop/tests/__init__.py
@@ -2,6 +2,11 @@
 from os.path import abspath, dirname, join, realpath
 import pytest
 import sys
+import os
+
+# WARNING: This variable must be set before any import of the securedrop code/app
+# or most tests will fail
+os.environ["SECUREDROP_ENV"] = "test"
 
 # The tests directory should be adjacent to the securedrop directory. By adding
 # the securedrop directory to sys.path here, all test modules are able to
--- a/securedrop/tests/conftest.py
+++ b/securedrop/tests/conftest.py
@@ -11,7 +11,6 @@ from unittest import mock
 import pretty_bad_protocol as gnupg
 import logging
 
-from _pytest.tmpdir import TempPathFactory
 from flask import Flask
 from hypothesis import settings
 import os
@@ -26,9 +25,7 @@ from flask import url_for
 from pyotp import TOTP
 from typing import Dict
 
-# WARNING: This variable must be set before any import of the securedrop code/app
-# or most tests will fail
-os.environ['SECUREDROP_ENV'] = 'test'  # noqa
+import sdconfig
 
 from passphrases import PassphraseGenerator
 from source_user import _SourceScryptManager, create_source_user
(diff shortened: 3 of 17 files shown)


securedrop/tests/__init__.py
```diff
@@ -2,6 +2,11 @@
 from os.path import abspath, dirname, join, realpath
 import pytest
 import sys
+import os
+
+# WARNING: This variable must be set before any import of the securedrop code/app
+# or most tests will fail
+os.environ["SECUREDROP_ENV"] = "test"
```
nabla-c0d3, line 9: Putting this here ensures that the `SECUREDROP_ENV` will always be set before any application code is imported.

securedrop/tests/conftest.py
```diff
@@ -110,37 +107,39 @@ def insecure_scrypt() -> Generator[None, None, None]:
         yield
 
 
-def setup_gpg(gpg_dir: Path) -> str:
```
nabla-c0d3, line 113: All I did here was move the logic in `setup_gpg()` to `setup_journalist_key_and_gpg_folder()`.

securedrop/tests/conftest.py
```diff
@@ -172,7 +171,10 @@ def config(
 
         config.SUPPORTED_LOCALES = i18n.get_test_locales()
 
-        yield config
+        # Set this newly-created config as the "global" config
+        with mock.patch.object(sdconfig, "config", config):
```
nabla-c0d3, line 175: Important change here: this ensures that the application code "sees" the same config as the one created in the fixture (when the code imports the config). 

It's not perfect and more work is needed with the config fixture, but it's a step forward.

securedrop/tests/conftest.py
```diff
@@ -194,6 +196,12 @@ def alembic_config(config: SDConfig) -> str:
 
 @pytest.fixture(scope='function')
 def source_app(config: SDConfig) -> Generator[Flask, None, None]:
+    config.SOURCE_APP_FLASK_CONFIG_CLS.TESTING = True
+    config.SOURCE_APP_FLASK_CONFIG_CLS.USE_X_SENDFILE = False
+
+    # Disable CSRF checks to make writing tests easier
+    config.SOURCE_APP_FLASK_CONFIG_CLS.WTF_CSRF_ENABLED = False
```
nabla-c0d3, line 203: This was copied from the config.py created by `make test`. These settings are needed for the tests to pass.

securedrop/tests/functional/functional_test.py
```diff
@@ -202,7 +199,7 @@ def set_default_driver(self):
             logging.error("Error stopping Firefox driver: %s", e)
 
     @pytest.fixture(autouse=True)
-    def sd_servers(self):
+    def sd_servers(self, setup_journalist_key_and_gpg_folder):
```
nabla-c0d3, line 202: The "main" change in this PR.

securedrop/tests/utils/env.py
```diff
@@ -34,16 +34,15 @@ def teardown(config):
...
+
+    # Delete the DB file
     db.session.remove()
-    try:
-        shutil.rmtree(config.TEMP_DIR)
-    except OSError:
-        # Then check the directory was already deleted
-        assert not os.path.exists(config.TEMP_DIR)
-    try:
-        shutil.rmtree(config.SECUREDROP_DATA_ROOT)
```
nabla-c0d3, line 44: I am no longer deleting the `SECUREDROP_DATA_ROOT` between each test because it will delete the `GPG_KEY_DIR` too (as it's a subfolder of `SECUREDROP_DATA_ROOT`).

securedrop/tests/conftest.py
zenmonkeykstop: One change required when running `make test` locally - the SECUREDROP_ENV var needs to be set, as sdconfig uses it to set the database URI to the appropriate value, otherwise alembic tests will fail. (Letting the rest of the tests run now to see if there are any similar issues with other tests).

zenmonkeykstop: `SECUREDROP_ENV` does not seem to be set here when  `make test` is run locally .  It's used to set the database URI to the appropriate value for the environment,  so without it the alembic tests in particular will fail. (Said failure isn't happening in CI for some reason).

One solution is to add `os.environ["SECUREDROP_ENV"] = "test"` before filesystem setup, but I'm sure there are more elegant ones :)

zenmonkeykstop (changes requested):
CI looks good, but the test_alembic.py  tests fail when `make test` is run locally as the db URI is being set to the prod value instead of test's.

securedrop/manage.py
```diff
@@ -21,20 +21,23 @@
...
-from sqlalchemy.orm.exc import NoResultFound
+import qrcode  # noqa: E402
+from sqlalchemy.orm.exc import NoResultFound  # noqa: E402
 
-os.environ['SECUREDROP_ENV'] = 'dev'  # noqa
 
-from db import db
-from models import (
+if not os.environ.get("SECUREDROP_ENV"):
+    os.environ['SECUREDROP_ENV'] = 'dev'  # noqa
```
nabla-c0d3, line 29: @zenmonkeykstop This fixes the alembic tests. Thanks for explaining what the problem was.

zenmonkeykstop (dismissed):

zenmonkeykstop:
(rebased to pull in a dependency update and get  tests passing)

zenmonkeykstop (approved):
Changes look good, tests passing locally as well now!

PR #6190, opened 2021-12-03
Updated wtforms dependency from 2.1 to 3.0.0

zenmonkeykstop:
## Status

Ready for Review

## Description of Changes

- Updates wtforms dependency from 2.1 to 3.0.0
- updates tests that were failing as a result of improved type checking in wtforms 3.0
- adds `--ignores` in `make safety` target for 2 Ansible-related warnings (`42926`, `42926`) that don't impact SecureDrop.

## Testing

- [ ] CI is passing

### If you made changes to the server application code:

- [x] Linting (`make lint`) and tests (`make test`) pass in the development container

Choose one of the following:

- [ ] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-docs) for these changes, or will do so later
- [ ] I would appreciate help with the documentation
- [x] These changes do not require documentation

### If you added or updated a production code dependency:

Choose one of the following:

- [ ] I have performed a diff review and pasted the contents to [the packaging wiki](https://github.com/freedomofpress/securedrop-debian-packaging/wiki)
- [ ] I would like someone else to do the diff review

PR #6191, opened 2021-12-07
Enforce 160-bit HOTP secret length and verify OTP secret length on login.

### LINKED ISSUE #6189: HOTP secrets are not  validated correctly.
## Description
HOTP secrets are not set up correctly in all cases:

On the JI:
- case 1:  if an admin adds a user and selects the "is using a Yubikey [HOTP]" checkbox, but does not provide a HOTP secret, an error is not thrown and an account using TOTP is set up instead. 
- case 2: if an admin adds a user and selects the "is using a Yubikey [HOTP]" checkbox, and provides a HOTP secret containing only spaces, an error is not thrown and an account using HOTP is set up.

## Steps to Reproduce

- log in to the JI as an admin user and go to **Admin**
- case 1: add a user, enabling HOTP but leaving the HOTP Secret field empty
- case 2: add a user, enabling HOTP and setting the HOTP Secret field to contain  one or more spaces only

## Expected Behavior
Both cases should fail, displaying the add user form again with a message about the length requirement for HOTP secrets

## Actual Behavior
- case 1: user add succeeds, admin is redirected to TOTP setup screen
- case 2: user add succeeds, admin is redirected to HOTP PIN verification screen.

## Comments

zenmonkeykstop:
## Status

Ready for review (but #6190 should be merged first and this PR should be updated accordingly)

## Description of Changes

Fixes #6189 .

- Enforces a 160-bit (40 hex chars in Admin Interface, 32 ascii-encoded chars in DB) requirement for HOTP secrets on new user accounts and edits to existing accounts' HOTP creds
  - Requires that a 40-char HOTP secret be provided when the `is_hotp` checkbox is selected on the New User form.
  - Requires that a 40-char HOTP secret be provided on the `Change Secret` admin page. 
- Checks for a minimum OTP secret length on all journalist accounts (TOTP and HOTP) on login, with the login failing if said length is less than 80 bits (20 hex, 16 ascii-encoded)

The motivation behind said 80-bit minimum is that there may still be 80-bit OTP secrets associated with journalist accounts using TOTP in production instances, as the 160-bit TOTP requirement was recently added and not applied retroactively. The minimum should be set to 160-bit at some future date, once it's reasonable to assume that the majority of users have re-rolled their OTP secret - or once 80-bit secrets are no longer considered viable security-wise.

## Testing

(all tests can be performed in a dev environment. A Yubikey and associated HOTP secret would be best for testing, but CLI commands to generate HOTP tokens like `oathtool -c <counter> --hotp "<hex secret here>"` will also work. (in most cases below `<counter>` will be `0`, but increment this field every time you use the same hotp secret. )

### Adding new users
- Run `make dev` against this branch and log in to the JI at `localhost:8081` with an admin-level account
- On the Admin page, choose **Add User** to add a new user
- add a TOTP user (leaving `is using a Yubikey` unchecked)  -  specify a username and click **Add User**
  - [ ]  confirm that the user is created
  - [ ]  confirm that the TOTP setup page is displayed with a valid QR code and a 32-char secret
- Go back to the Admin page, and choose **Add User** to add another new user
- enter a new username and select the `is using a Yubiky` checkbox. Then click **Add User**
  - [ ] confirm that the user is not created and the error message `The "otp_secret" field is required when "is_hotp" is set.` is displayed
- enter an arbitrary number of spaces in the `HOTP Secret` field and click **Add User**
  - [ ] confirm that the user is not created and the error message `The "otp_secret" field is required when "is_hotp" is set.` is displayed
- enter a 40-char (excluding spaces) string including non-hexadecimal characters  in the `HOTP Secret` field and click **Add User**
  - [ ] confirm that the user is not created and the flash error message `Invalid HOTP secret format: please only submit letters A-F and numbers 0-9.` is displayed
- enter an incomplete OTP secret in hex format (any string of `<num>` hex chars with less than 40 non-space chars)  in the `HOTP Secret` field and click **Add User**
   - [ ] confirm that the user is not created and the error message `The  HOTP secrets are 40 characters long - you have entered <num>.` is displayed
- enter a complete OTP secret in hex format (40 hex chars with arbitrary spaces)  in the `HOTP Secret` field and click **Add User**
   - [ ] confirm that the user is  created and the  HOTP secret can be used to generate a valid confirmation code on the next page.

### Editing existing users
- As an admin user, log onto the JI, go to the Admin page, and click the edit button beside a user other than the admin user.
- Click **Reset Mobile App Credentials** and click **OK**. 
  - [ ] Confirm that the TOTP page is displayed with a valid QR code and 32-char ascii-encoded secret
  - [ ] Confirm that the secret produces a valid token using the form on this page.
- Go back to the Admin page and click the edit button beside the user again
- Click **Reset Security Key Credentials** and click **OK**.
  - [ ] Confirm that the **Change HOTP Secret** page is displayed.
  - [ ] Click **Continue** without entering anything in the field, and confirm that the page is reloaded without any change.
  - [ ] Add some spaces in the field and click **Continue** - confirm that the page reloads with a length error message 
  - [ ] Add a string of less than 40 hex digits, with arbitrary spaces, in the field and click **Continue** - confirm that the page reloads with a length error message
  - [ ] Add a string of 40 chars including non-hex chars in the field and click **Continue**  - confirm that the page reloads with an error specifying the correct string formatting.
  -  [ ] Add a valid secret of 40 hex digits with arbitrary spaces in the field and click **Continue**  - confirm that the **Enable Yubikey (OATH-HOTP)** page loads, and that the token generated by the valid secret works.

- [ ] Repeat the above checks for the logged-in user's account  and confirm that they pass.

### Logging in as a user with an invalid OTP secret
- connect to the docker container with `docker exec -it securedrop-dev-0 sqlite3 /var/lib/securedrop/db.sqlite`
- pick a HOTP user to update - `select username, otp_secret from journalists where is_hotp=1` - and change their otp secret to a valid ascii-encoded but less-than-16-char string, ie. `GARBAGE` with `update journalists set otp_secret='GARBAGE' where username='<your user here>'`
- attempt to log in as the chosen user, generating a valid OTP token at the command line with `oathtool -b --hotp "GARBAGE"`
  - [ ] Verify that the user is not logged in and that an error is displayed instructing them to contact an admin to reset their 2FA.
- log in as an admin-privilege user and update the chosen user's HOTP secret with a valid 40-char hex secret.
- [ ] log in as the chosen user using the new secret and confirm that the user can log in.
- [ ] log in as a user with an 80-bit (16-char in the DB, ie `JHCOGO7VCER3EJ4L`) secret and confirm that the user can log in
- [ ] create a new user and verify that they can log in.

## Deployment

No deployment-sensitive changes were introduced in this 

Choose one of the following:

- [x] I would appreciate help with the documentation (docs for HOTP only describe Yubikeys, which already have 160-bit HOTP secrets, so no change there, but it could be worthwhile to document the new failed login state when the account's OTP secret is invalid.)


eloquence:
(Given the recommended merge order and change of base branch on this PR, it may make sense to put it in draft mode for now.)

securedrop/journalist_app/utils.py
```diff
@@ -134,21 +142,26 @@ def validate_hotp_secret(user: Journalist, otp_secret: str) -> bool:
     :param otp_secret: the new HOTP secret
     :return: True if it validates, False if it does not
     """
+    strip_whitespace = otp_secret.replace(' ', '')
+    secret_length = len(strip_whitespace)
+
+    if secret_length != HOTP_SECRET_LENGTH:
+        flash(ngettext(
+                'HOTP secrets are 40 characters long - you have entered {num}.',
+                'HOTP secrets are 40 characters long - you have entered {num}.',
```
cfm, line 151: Shouldn't the `ngettext()` string also be `format()`'d with `HOTP_SECRET_LENGTH`, for consistency with the new logic?

cfm, line 151: Nit:  With #6197 on the brain, I wonder about adding an explicit instruction after the explanation, e.g.:

> HOTP secrets are 40 characters long - you have entered {num}. Check your HOTP secret and reenter it carefully.

cfm (changes requested):
Test plan executed via `oathtool`.  Everything checks out functionally, but I've left one box unticked to make sure the cases behave as you intend, @zenmonkeykstop, and left two other comments inline for your consideration.  :-)

### Adding new users
- Run `make dev` against this branch and log in to the JI at `localhost:8081` with an admin-level account
- On the Admin page, choose **Add User** to add a new user
- add a TOTP user (leaving `is using a Yubikey` unchecked)  -  specify a username and click **Add User**
  - [x]  confirm that the user is created
  - [x]  confirm that the TOTP setup page is displayed with a valid QR code and a 32-char secret
- Go back to the Admin page, and choose **Add User** to add another new user
- enter a new username and select the `is using a Yubiky` checkbox. Then click **Add User**
  - [x] confirm that the user is not created and the error message `The "otp_secret" field is required when "is_hotp" is set.` is displayed
- enter an arbitrary number of spaces in the `HOTP Secret` field and click **Add User**
  - [x] confirm that the user is not created and the error message `The "otp_secret" field is required when "is_hotp" is set.` is displayed
- enter a string including non-hexadecimal characters  in the `HOTP Secret` field and click **Add User**
  - [ ] confirm that the user is not created and the error message `HOTP secrets must be in hex format` <num>.` is displayed

Entering a non-hex string longer or shorter than 40 characters yields the error message:

> HOTP secrets are 40 characters long - you have entered 32.

Entering a non-hex string of exactly 40 characters yields the flashed error message:

> Invalid HOTP secret format: please only submit letters A-F and numbers 0-9.

- enter an incomplete OTP secret in hex format (any string of `<num>` hex chars with less than 40 non-space chars)  in the `HOTP Secret` field and click **Add User**
   - [x] confirm that the user is not created and the error message `The  HOTP secrets are 40 characters long - you have entered <num>.` is displayed
- enter a complete OTP secret in hex format (40 hex chars with arbitrary spaces)  in the `HOTP Secret` field and click **Add User**
   - [x] confirm that the user is  created and the  HOTP secret can be used to generate a valid confirmation code on the next page.

### Editing existing users
- As an admin user, log onto the JI, go to the Admin page, and click the edit button beside a user other than the admin user.
- Click **Reset Mobile App Credentials** and click **OK**. 
  - [x] Confirm that the TOTP page is displayed with a valid QR code and 32-char ascii-encoded secret
  - [x] Confirm that the secret produces a valid token using the form on this page.
- Go back to the Admin page and click the edit button beside the user again
- Click **Reset Security Key Credentials** and click **OK**.
  - [x] Confirm that the **Change HOTP Secret** page is displayed.
  - [x] Click **Continue** without entering anything in the field, and confirm that the page is reloaded without any change.
  - [x] Add some spaces in the field and click **Continue** - confirm that the page reloads with a length error message

> HOTP secrets are 40 characters long - you have entered 0.

  - [x] Add a string of less than 40 hex digits, with arbitrary spaces, in the field and click **Continue** - confirm that the page reloads with a length error message
  
> HOTP secrets are 40 characters long - you have entered 8.
  
  - [x] Add a string of 40 chars including non-hex chars in the field and click **Continue**  - confirm that the page reloads with an error specifying the correct string formatting.

> Invalid HOTP secret format: please only submit letters A-F and numbers 0-9.

  -  [x] Add a valid secret of 40 hex digits with arbitrary spaces in the field and click **Continue**  - confirm that the **Enable Yubikey (OATH-HOTP)** page loads, and that the token generated by the valid secret works.

- [x] Repeat the above checks for the logged-in user's account  and confirm that they pass.

### Logging in as a user with an invalid OTP secret
- connect to the docker container with `docker exec -it securedrop-dev-0 sqlite3 /var/lib/securedrop/db.sqlite`
- pick a HOTP user to update - `select username, otp_secret from journalists where is_hotp=1` - and change their otp secret to a valid ascii-encoded but less-than-16-char string, ie. `GARBAGE` with `update journalists set otp_secret='GARBAGE' where username='<your user here>'`

NB.  `s/is_hotp=1/is_totp=0`

- attempt to log in as the chosen user, generating a valid OTP token at the command line with `oathtool -b --hotp "GARBAGE"`
  - [x] Verify that the user is not logged in and that an error is displayed instructing them to contact an admin to reset their 2FA.
  

> Login failed. Your 2FA details are invalid - please contact an administrator to reset them.  

- log in as an admin-privilege user and update the chosen user's HOTP secret with a valid 40-char hex secret.
- [x] log in as the chosen user using the new secret and confirm that the user can log in.
- [x] log in as a user with an 80-bit (16-char in the DB, ie `JHCOGO7VCER3EJ4L`) secret and confirm that the user can log in
- [x] create a new user and verify that they can log in.

securedrop/journalist_app/utils.py
zenmonkeykstop: I mean, there's a good argument for it, if we envisaged that changing. Considering the limit is baked into Yubikeys (which are all we recommend for HOTP) and gives a good 160 bits of key length (which is the upper limit recommended by the relevant RFC and probably overkill from a cryptographic perspective), I don't think it will change, and as such it's probably not worth changing the string and triggering work for translators.

zenmonkeykstop: As per above, don't want to update strings without a solid reason. Also users are probably going to be cutting/pasting this out of the yubikey management app in tails anyway, so telling them to enter it manually in response to what was probably a mousing snafu seems a little punitive.

zenmonkeykstop:
TY @cfm! I'll check out that questionable case in the morning - the actual behaviour seems good to me but I'm confused by the discrepancy in the test plan. Gotta have a word with past Kev.

securedrop/journalist_app/utils.py
cfm: Okay, I'm with you (a) that `HOTP_SECRET_LENGTH` is highly unlikely to change and (b) that we want to avoid needlessly adding or complicating translation work.  Something like—

```python
        flash(ngettext(
                f'HOTP secrets are {HOTP_SECRET_LENGTH} characters long - you have entered {{num}}.',
                f'HOTP secrets are {HOTP_SECRET_LENGTH} characters long - you have entered {{num}}.',
                secret_length
            ).format(num=secret_length), "error")
```

—would, I believe, defer to `HOTP_SECRET_LENGTH` (rather than hard-coding its value of `40`) *and* still satisfy these desiderata.  But I'm happy to downgrade this suggestion from "reality check" to "nit not worth picking".  :-)

cfm: I should clarify that I'm not wedded to that particular instruction, which could easily be instead (e.g.):

> HOTP secrets are 40 characters long - you have entered {num}. Make sure you type or copy-paste the entire secret.

I think #6197 offers a good general principle (say what just happened, then say what to do next), but I hear you that this may not be the time to introduce this particular string change if it can be avoided.

cfm:
@zenmonkeykstop, happy to take another look and approve after you've conferred with past @zenmonkeykstop.  ;-)  As I look at this again this morning, is there any reason not to return the same error message in both cases, to avoid having to give one precedence over the other?  Something like:

> Invalid HOTP secret format: please enter exactly 40 characters of letters A-F and numbers 0-9.

(Another string change to weigh the value of, I know....)

zenmonkeykstop:
(Adding @lsd-cat as this came up elsewhere and they showed an interest - no need for a full review.)

securedrop/journalist_app/utils.py
```diff
@@ -111,6 +114,11 @@ def validate_user(
                 "Please wait at least {num} seconds before logging in again.",
                 period
             ).format(num=period)
+        elif isinstance(e, InvalidOTPSecretException):
```
zenmonkeykstop, line 117: @lsd-cat I'd be interested in your perspective on this addition - it's the only case where we provide a non-generic login error message, which could be problematic.

securedrop/models.py
```diff
@@ -36,6 +36,13 @@
 
 ARGON2_PARAMS = dict(memory_cost=2**16, rounds=4, parallelism=2)
 
+# Required length for hex-format HOTP secrets as input by users
+HOTP_SECRET_LENGTH = 40  # 160 bits == 40 hex digits (== 32 ascii-encoded chars in db)
+
+# Minimum length for ascii-encoded OTP secrets - by default, secrets are now 160-bit (32 chars)
+# but existing Journalist users may still have 80-bit (16-char) secrets
+OTP_SECRET_MIN_ASCII_LENGTH = 16  # 80 bits == 40 hex digits (== 16 ascii-encoded chars in db)
```
zenmonkeykstop, line 44: that's a typo, should be 20 hex digits.

zenmonkeykstop:
@cfm looking at that test case again, I think it was just not explicit that the test string should be 40 chars *and* include non-hex chars. The message in that case should be as you listed. I've updated the test plan accordingly

One annoying todo here is how those error messages are being displayed - validation errors from the form parser are displayed under the affected fields which is fine, but other errors (basically the errors on insertion into the db) are top-of-page flash messages. It would be nice to have a consistent approach, tho it's outside the scope of this PR imo.

zenmonkeykstop:
> @zenmonkeykstop, happy to take another look and approve after you've conferred with past @zenmonkeykstop. ;-) As I look at this again this morning, is there any reason not to return the same error message in both cases, to avoid having to give one precedence over the other? Something like:
> 
> > Invalid HOTP secret format: please enter exactly 40 characters of letters A-F and numbers 0-9.
> 
> (Another string change to weigh the value of, I know....)

The exact wording might need tweaking but yes, this would be a worthy change. The expected format is 40 hex chars with arbitrary spaces, so the only problem is finding a nice unambiguous way to say that.

cfm (approved):

cfm:
@zenmonkeykstop in <https://github.com/freedomofpress/securedrop/pull/6191#issuecomment-1009320295>:

> @cfm looking at that test case again, I think it was just not explicit that the test string should be 40 chars _and_ include non-hex chars. The message in that case should be as you listed. I've updated the test plan accordingly

Thanks!  Approving accordingly, but I'll let you merge at your discretion re: <https://github.com/freedomofpress/securedrop/pull/6191#issuecomment-1009386735> and <https://github.com/freedomofpress/securedrop/pull/6191#discussion_r781480384>.

securedrop/journalist_app/utils.py
lsd-cat: Seems reasonable to me. Unless the 60 seconds/5 attempts is something that admins are encouraged to change, which does not look the case, a potential attacker already knows what is the limit regardless of the error messages.

zenmonkeykstop:
D'oh - the base for this PR was still pointing at `update-wtforms-3.0` - it should have been targeted at `develop`... more :coffee:  required.

PR #6195, opened 2021-12-10
Remove "Refresh codename" feature

eloquence:
## Status
Ready for review

## Description

Removes the "Refresh" icon from the page that presents a source with a seven word codename. In practice, this feature has potential for user confusion, as it introduces complexity to a critical step of the submission process while delivering minimal end user value.

One key motivation for this feature has been the occasional generation of problematic words in codenames. We have already iterated over the wordlists several times, and will continue to do so.

## Testing

1. Log into the Source Interface
2. Verify that there no longer is an icon to refresh the codename

## Deployment

Note that this PR removes an image from the package, which is also removed from the AppArmor allow-list.

Choose one of the following:


securedrop/tests/functional/source_navigation_steps.py
```diff
@@ -47,7 +47,7 @@ def _source_clicks_submit_documents_on_homepage(self):
         assert self._is_on_generate_page()
 
     def _source_regenerates_codename(self):
-        self.safe_click_by_id("regenerate-submit")
+        self.driver.refresh()
```
eloquence, line 50: From my read, `test_refreshed_duplicate_generate_pages` is still useful even without the "refresh codename" feature (for example, it tests that the user is redirected to the lookup page when they have already logged in in a separate tab), so I have simply altered the behavior here to reload the page instead.

eloquence:
For context, @zenmonkeykstop, @ninavizz and I identified this as a potential "quick win" on the path towards improving the source user experience, prior to bigger changes such as overhauling/inverting the authentication flow and retiring the term "codename".

zenmonkeykstop (approved):
LGTM, quick wins ftw.

Issue #5761, opened 2021-01-27
legacy `config.py` files may prevent the web applications from starting or break functionality


zenmonkeykstop:
## Description

See #5757 - instances without the SESSION_EXPIRY_MINUTES attribute defined in `config.py` were failing to start after a code update that assumed all required attributes were defined was pushed. This was fixed in #5759 but other problems may lurk in older config files.

It might be worth:
- looking at historical updates to `config.py` to figure out what attributes were added and when, and making sure they're handled correctly in future code updates
- figuring out how to better manage app config (@heartsucker had a long-running PR for this that I can't find now - another option is to move more attributes into the database)
- figuring out how to update `config.py` to a valid state during an application update

eloquence:
Old PR by @heartsucker: https://github.com/freedomofpress/securedrop/pull/3850
Old, related issue: #1966

emkll:
During the 1.7.0 release postmortem, other suggestions were proposed by the team:

- Testing historical possible versions of config.py
- Separating the configuration between private (salts, secrets) configuration and public (common configuration)
- Move as much configuration as possible to the database
- Add logic to the restore script, to "update" the configuration to the update the configuration with the latest/expected values

Given the complexity of this logic, but also the state being different across instances based on when the instance was first set up (config.py is carried over via backups but is not touched by the application role on subsequent installs), we should perform a timeboxed investigation to carefully review the current implementation, and propose next steps (which may include one or more tasks included in this thread)

PR #6207, opened 2021-12-26
Refactor one functional test to showcase improved fixtures & factories

nabla-c0d3:
## Status

Ready for review

## Description of Changes

This PR refactors one specific functional tests, `test_source_session_timeout()`, to showcase some improvements I've been working on, which I could apply to all the functional tests in future PRs, if this is approved.

Through the `FunctionalTest` class and inheritance/MixIns, the functional tests are implemented as a "God object": all the functionality (fixtures, actual test code, selenium/nav code, etc.) is in one object. This is visible for example by the fact that all the methods the functional tests use are attached to `self`. While the functional tests do work well, it makes them difficult to understand and modify.

This PR split `FunctionalTest` and the related `MixIn` classes into:

* A `tor_browser_web_driver` fixture for creating a Selenium web driver.
* An improved `sd_servers` fixture for running the journalist and source apps in separate processes, that can take any `SDConfig` to be used in the apps. This allows for example spawning SD servers with very short session timeouts.
* A `SourceAppNagivator` which contains the Selenium code for interacting with the source app.

Overall, this PR helps with the following issues:

* https://github.com/freedomofpress/securedrop/issues/3836
* https://github.com/freedomofpress/securedrop/issues/5761

While the PR does add ~700 lines of code, a lot of it is duplicate code that I will remove if I apply the same changes to all the functional tests.

For reviewing it, I would suggest starting at the end / last file and class: `TestSourceAppSessionExpiration`.

### DIFF
--- a/securedrop/source_app/session_manager.py
+++ b/securedrop/source_app/session_manager.py
@@ -4,7 +4,6 @@ from typing import TYPE_CHECKING
 import sqlalchemy
 from flask import session
 
-from sdconfig import config
 from source_user import SourceUser, authenticate_source_user, InvalidPassphraseError
 
 if TYPE_CHECKING:
@@ -40,6 +39,8 @@ class SessionManager:
         db_session: sqlalchemy.orm.Session,
         supplied_passphrase: "DicewarePassphrase"
     ) -> SourceUser:
+        from sdconfig import config
+
         # Validate the passphrase; will raise an exception if it is not valid
         source_user = authenticate_source_user(
             db_session=db_session, supplied_passphrase=supplied_passphrase
--- a/securedrop/tests/functional/app_navigators.py
+++ b/securedrop/tests/functional/app_navigators.py
@@ -1,0 +1,323 @@
+from contextlib import contextmanager
+import logging
+import os
+from pathlib import Path
+from enum import Enum
+import time
+from datetime import datetime
+from os.path import abspath
+from os.path import dirname
+from os.path import expanduser
+from os.path import join
+from os.path import realpath
+from typing import Generator, Optional
+
+from selenium.webdriver.firefox.webdriver import WebDriver
+
+import tbselenium.common as cm
+from selenium import webdriver
+from selenium.common.exceptions import NoAlertPresentException
+from selenium.common.exceptions import WebDriverException
+from selenium.webdriver.common.by import By
+from selenium.webdriver.remote.remote_connection import LOGGER
+from selenium.webdriver.support import expected_conditions
+from selenium.webdriver.support.ui import WebDriverWait
+from tbselenium.tbdriver import TorBrowserDriver
+
+
+LOGFILE_PATH = abspath(join(dirname(realpath(__file__)), "../log/driver.log"))
+FIREFOX_PATH = "/usr/bin/firefox/firefox"
+
+TBB_PATH = abspath(join(expanduser("~"), ".local/tbb/tor-browser_en-US/"))
+os.environ["TBB_PATH"] = TBB_PATH
+TBBRC = join(TBB_PATH, "Browser/TorBrowser/Data/Tor/torrc")
+LOGGER.setLevel(logging.WARNING)
+
+# width & height of the browser window. If the top of screenshots is cropped,
+# increase the height of the window so the the whole page fits in the window.
+BROWSER_SIZE = (1024, 1400)
+
+
+class WebDriverTypeEnum(Enum):
+    TOR_BROWSER = 1
+    FIREFOX = 2
+
+
+_DRIVER_RETRY_COUNT = 3
+_DRIVER_RETRY_INTERNVAL = 5
+
+
+def _create_torbrowser_driver(
+    accept_languages: Optional[str] = None,
+) -> TorBrowserDriver:
+    logging.info("Creating TorBrowserDriver")
+    log_file = open(LOGFILE_PATH, "a")
+    log_file.write("\n\n[%s] Running Functional Tests\n" % str(datetime.now()))
+    log_file.flush()
+
+    # Don't use Tor when reading from localhost, and turn off private
+    # browsing. We need to turn off private browsing because we won't be
+    # able to access the browser's cookies in private browsing mode. Since
+    # we use session cookies in SD anyway (in private browsing mode all
+    # cookies are set as session cookies), this should not affect session
+    # lifetime.
+    pref_dict = {
+        "network.proxy.no_proxies_on": "127.0.0.1",
+        "browser.privatebrowsing.autostart": False,
+    }
+    if accept_languages is not None:
+        pref_dict["intl.accept_languages"] = accept_languages
+
+    Path(TBB_PATH).mkdir(parents=True, exist_ok=True)
+    torbrowser_driver = None
+    for i in range(_DRIVER_RETRY_COUNT):
+        try:
+            torbrowser_driver = TorBrowserDriver(
+                TBB_PATH,
+                tor_cfg=cm.USE_RUNNING_TOR,
+                pref_dict=pref_dict,
+                tbb_logfile_path=LOGFILE_PATH,
+            )
+            logging.info("Created Tor Browser web driver")
+            torbrowser_driver.set_window_position(0, 0)
+            torbrowser_driver.set_window_size(*BROWSER_SIZE)
+            break
+        except Exception as e:
+            logging.error("Error creating Tor Browser web driver: %s", e)
+            if i < _DRIVER_RETRY_COUNT:
+                time.sleep(_DRIVER_RETRY_INTERNVAL)
+
+    if not torbrowser_driver:
+        raise Exception("Could not create Tor Browser web driver")
+
+    return torbrowser_driver
+
+
+def _create_firefox_driver(
+    accept_languages: Optional[str] = None,
+) -> webdriver.Firefox:
+    logging.info("Creating Firefox web driver")
+
+    profile = webdriver.FirefoxProfile()
+    if accept_languages is not None:
+        profile.set_preference("intl.accept_languages", accept_languages)
+        profile.update_preferences()
+
+    firefox_driver = None
+    for i in range(_DRIVER_RETRY_COUNT):
+        try:
+            firefox_driver = webdriver.Firefox(firefox_binary=FIREFOX_PATH, firefox_profile=profile)
+            firefox_driver.set_window_position(0, 0)
+            firefox_driver.set_window_size(*BROWSER_SIZE)
+            logging.info("Created Firefox web driver")
+            break
+        except Exception as e:
+            logging.error("Error creating Firefox web driver: %s", e)
+            if i < _DRIVER_RETRY_COUNT:
+                time.sleep(_DRIVER_RETRY_INTERNVAL)
+
+    if not firefox_driver:
+        raise Exception("Could not create Firefox web driver")
+
+    return firefox_driver
+
+
+# TODO(AD): This could eventually replace the web driver code in FunctionalTest
+@contextmanager
+def get_web_driver(
+    web_driver_type: WebDriverTypeEnum = WebDriverTypeEnum.TOR_BROWSER,
+) -> Generator[WebDriver, None, None]:
+    if web_driver_type == WebDriverTypeEnum.TOR_BROWSER:
+        web_driver = _create_torbrowser_driver()
+    elif web_driver_type == WebDriverTypeEnum.FIREFOX:
+        web_driver = _create_firefox_driver()
+    else:
+        raise ValueError(f"Unexpected value {web_driver_type}")
+
+    try:
+        yield web_driver
+    finally:
+        try:
+            web_driver.quit()
+        except Exception:
+            logging.exception("Error stopping driver")
+
+
+# TODO(AD): This could eventually replace the navigation/driver code in FunctionalTest
+class _NavigationHelper:
+
+    _TIMEOUT = 10
+    _POLL_FREQUENCY = 0.1
+
+    def __init__(self, web_driver: WebDriver) -> None:
+        self.driver = web_driver
+
+    def wait_for(self, function_with_assertion, timeout=_TIMEOUT):
+        """Polling wait for an arbitrary assertion."""
+        # Thanks to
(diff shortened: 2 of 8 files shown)


securedrop/tests/functional/app_navigators.py
```diff
@@ -0,0 +1,323 @@
+from contextlib import contextmanager
```
nabla-c0d3, line 1: This file is mostly a copy&paste of `FunctionalTest` with some minor fixes for readability and potential flakyness.

securedrop/tests/functional/app_navigators.py
```diff
@@ -0,0 +1,323 @@
...
+                return True
+
+        self.driver.switch_to.alert.accept()
+        WebDriverWait(self.driver, self._TIMEOUT, self._POLL_FREQUENCY).until(
+            alert_is_not_present, "Timed out waiting for confirmation popup to disappear."
+        )
+
+
+# TODO(AD): This could eventually replace the SourceNavigationStepsMixin
+class SourceAppNagivator:
```
nabla-c0d3, line 281: This is a copy&paste of `SourceNavigationStepsMixin`.

securedrop/tests/functional/db_session.py
```diff
@@ -0,0 +1,38 @@
+from contextlib import contextmanager
```
nabla-c0d3, line 1: This file contains a helper function that I think will be useful in the future.

securedrop/tests/functional/factories.py
```diff
@@ -0,0 +1,75 @@
+from pathlib import Path
```
nabla-c0d3, line 1: This file contains two factories, which follow the API convention of the `factoryboy` library (in case this library is ever added to the unit tests dependencies). The factories can later be used in more code locations.

securedrop/tests/functional/sd_config_v2.py
```diff
@@ -0,0 +1,102 @@
+from dataclasses import dataclass, field
```
nabla-c0d3, line 1: This file contains a proposal for tweaking the SDConfig object. It is only used in the test modified in this PR right now. 
* I needed it because the existing `SDConfig` class does not have a usable constructor: I can't easily instantiate one with the config values/settings that I want. Even just importing `sdconfig.py` requires a `config.py` file to be present (which should not be a requirement when running the tests, especially when different tests use different configs).
* One improvement it brings is the fact that it's immutable (thanks to `frozen=True`). This prevents dynamically modifying the config as done sometimes in the tests. It's better to not do it because it doesn't really work in the first place (most of the app's code has already imported the `sdconfig.config`by the time it gets mutated). It's also a step toward "data-only" / static configuration files.

securedrop/tests/functional/test_source_session_timeout.py
```diff
@@ -1,17 +1,176 @@
...
+from tests.functional.sd_config_v2 import SecureDropConfig
 
-    def test_source_session_timeout(self):
-        self._source_visits_source_homepage()
-        self._source_clicks_submit_documents_on_homepage()
-        self._source_continues_to_submit_page()
-        self._source_waits_for_session_to_timeout()
-        self.driver.refresh()
-        self._source_sees_session_timeout_message()
+
```
nabla-c0d3, line 18: What follows is very similar to the existing sd_servers fixture, but with some fixes and tweaks.

securedrop/tests/functional/test_source_session_timeout.py
```diff
@@ -1,17 +1,176 @@
...
+        yield web_driver
+
+
+@pytest.fixture(scope="session")
+def firefox_web_driver() -> WebDriver:
+    with get_web_driver(web_driver_type=WebDriverTypeEnum.FIREFOX) as web_driver:
+        yield web_driver
+
+
+class TestSourceAppSessionExpiration:
```
nabla-c0d3, line 155: The actual test showing how to use the new fixture / testing code:
* It doesn't rely on everything being accessible in `self` - the functionality is split in different, smaller and easier-to-grasp classes.
* It uses the two relevant fixtures which are also decoupled from `self`.

securedrop/tests/functional/test_source_session_timeout.py
```diff
@@ -1,17 +1,176 @@
...
+
+
+def _start_source_server(port: int, config: SecureDropConfig) -> None:
+    # This function will be called in a separate Process that runs the source app
+    # Modify the sdconfig module in the app's memory so that it mirrors the supplied config
+    # Do this BEFORE importing any other module of the application so the modified config is
+    # what eventually gets imported by the app's code
+    import sdconfig
+
+    sdconfig.config = config  # type: ignore
```
nabla-c0d3, line 34: The above is the "killer" feature of the new `sd_servers_with_short_session_timeout` fixture (which can become a more general `sd_servers_v2` fixture in a future PR): any SDConfig can be supplied, and the source and journalist apps to be spawned will use that (regardless of the content of securedrop/config.py or anything else).

securedrop/tests/functional/test_source_session_timeout.py
```diff
@@ -1,17 +1,176 @@
...
+            database_uri=config_to_use.DATABASE_URI
+        ) as db_session_for_sd_servers:
+            journalist = Journalist(
+                username="journalist",
+                password="correct horse battery staple profanity oil chewy",
+                is_admin=True,
+            )
+            journalist.otp_secret = "JHCOGO7VCER3EJ4L"
+            db_session_for_sd_servers.add(journalist)
+            db_session_for_sd_servers.commit()
```
nabla-c0d3, line 98: This demonstrates the usefulness of `get_database_session()`.

securedrop/tests/functional/test_source_session_timeout.py
```diff
@@ -1,17 +1,176 @@
...
+
```
nabla-c0d3, line 99: What follows is very similar to the `sd_servers` fixture.

securedrop/tests/functional/sd_config_v2.py
nabla-c0d3: Note that this new `SecureDropConfig` works transparently as a replacement for `SDConfig`, because it has the same attributes.

nabla-c0d3:
A small example of the kind of refactoring/improvements this PR can lead to in the functional tests is available at https://github.com/freedomofpress/securedrop/pull/6215/commits/43d7cff43de7ab4ac3bcd833822cb554c9f1121c#diff-e614026a9dafb6e23598838328889476e52daee6b8da7a4e6f2c76f9b390364cL1

securedrop/tests/functional/sd_config_v2.py
legoktm: I really like the idea of using a frozen dataclass for config, it makes perfect sense. But I'm wondering if it would make more sense to just do this in SDConfig itself, without trying to maintain a parallel implementation for tests.

We could add a proper `__init__` to SDConfig and move the current one to be a classmethod `from_config()` or something. Then we would swap SDConfig into a dataclass underneath that while still supporting the pretty permissive creation style we have currently.

securedrop/tests/functional/test_source_session_timeout.py
legoktm: To make sure I understand correctly, all of this setup and fixture code is only in this file because this is the only test you've ported so far, in the end it would be shared across all functional tests and be in something like functional_test.py, right?

securedrop/tests/functional/test_source_session_timeout.py
```diff
@@ -1,17 +1,176 @@
...
+    journalist_app_base_url: str
+
+
+# Very short session expiration time
+SESSION_EXPIRATION_SECONDS = 3
+
+
+# TOOD(AD): This could eventually replace the sd_servers fixture
+@pytest.fixture(scope="session")
+def sd_servers_with_short_session_timeout(setup_journalist_key_and_gpg_folder):
```
legoktm, line 71: How would you pass in a different SESSION_EXPIRATION_SECONDS in this fixture for tests that don't need a a short session?

legoktm (commented):
I only reviewed the proposed architecture/high-level changes since you said this is still a work in progress, let me know if you want(ed) a more detailed review. Overall I like the direction this is going in.

securedrop/tests/functional/sd_config_v2.py
nabla-c0d3: The goal is for this v2 to eventually replace SDConfig. I didn't do it here because it would require a lot more changes: importing sdconfig.py requires a config.py to be present, and changing this requires a bigger refactoring (which I plan on doing eventually). 
Also based on previous PRs, touching SDConfig has the potential to break compatibility with running SD instances and should be done separately, in a small PR.

securedrop/tests/functional/test_source_session_timeout.py
nabla-c0d3: Correct. We could also run the functional tests in parallel by spawning multiple sd_servers.

securedrop/tests/functional/test_source_session_timeout.py
nabla-c0d3: In the next PRs, I've made this code more generic so that it can be used for tests that require a normal session timeout. Here it's just an initial example.

nabla-c0d3:
Thank you! I replied to your questions/comments. This PR is ready to go, and will unlock the next few PRs where I bring these changes to more functional tests.

securedrop/tests/functional/test_source_session_timeout.py
legoktm: I think it would make more sense to move these setup into conftest.py as part of this PR, as you're already doing in https://github.com/freedomofpress/securedrop/pull/6209/commits/8df04311dd254cced235449a66335a2c615f6d9a. Having it there from the beginning will be nicer since we're not moving the same bits of code around multiple times.

securedrop/source_app/session_manager.py
```diff
@@ -40,6 +39,8 @@ def log_user_in(
         db_session: sqlalchemy.orm.Session,
         supplied_passphrase: "DicewarePassphrase"
     ) -> SourceUser:
+        from sdconfig import config
```
legoktm, line 42: Please add a comment indicating why we're importing late. E.g. in source_user.py it says:
```python3
    def get_default(cls) -> "_SourceScryptManager":
        # Late import so _SourceScryptManager can be used without a config.py in the parent folder
        from sdconfig import config
```

I assume there's a similar bootstrapping reason here.

securedrop/tests/functional/app_navigators.py
```diff
@@ -0,0 +1,323 @@
+from contextlib import contextmanager
+import logging
+import os
+from pathlib import Path
+from enum import Enum
+import time
+from datetime import datetime
+from os.path import abspath
```
legoktm, line 8: nit: All the os.path imports can be merged into one line...but really it would be nice if they could be switched to use pathlib instead.

securedrop/tests/functional/app_navigators.py
```diff
@@ -0,0 +1,323 @@
...
+from selenium.webdriver.remote.remote_connection import LOGGER
+from selenium.webdriver.support import expected_conditions
+from selenium.webdriver.support.ui import WebDriverWait
+from tbselenium.tbdriver import TorBrowserDriver
+
+
+LOGFILE_PATH = abspath(join(dirname(realpath(__file__)), "../log/driver.log"))
+FIREFOX_PATH = "/usr/bin/firefox/firefox"
+
+TBB_PATH = abspath(join(expanduser("~"), ".local/tbb/tor-browser_en-US/"))
```
legoktm, line 31: nit: The join is unnecessary, `abspath(expanduser("~/.local/tbb/tor-browser_en-US/"))` works fine.

securedrop/tests/functional/app_navigators.py
```diff
@@ -0,0 +1,323 @@
...
+            if i < _DRIVER_RETRY_COUNT:
+                time.sleep(_DRIVER_RETRY_INTERNVAL)
+
+    if not firefox_driver:
+        raise Exception("Could not create Firefox web driver")
+
+    return firefox_driver
+
+
+# TODO(AD): This could eventually replace the web driver code in FunctionalTest
```
legoktm, line 125: Maybe: "This is intended to eventually replace..."

And elsewhere in this file too.

securedrop/tests/functional/app_navigators.py
```diff
@@ -0,0 +1,323 @@
...
+# TODO(AD): This could eventually replace the SourceNavigationStepsMixin
+class SourceAppNagivator:
+    def __init__(
+        self,
+        source_app_base_url: str,
+        web_driver: WebDriver,
+        accept_languages: Optional[str] = None,
+    ) -> None:
+        self._source_app_base_url = source_app_base_url
+        self._nav_helper = _NavigationHelper(web_driver)
```
legoktm, line 289: In https://github.com/freedomofpress/securedrop/pull/6209/commits/8df04311dd254cced235449a66335a2c615f6d9a you rename this to `self.nav_helper`, I think we should start out with it like that.

securedrop/tests/functional/db_session.py
legoktm: I agree this is useful, but any reason not to put it in conftest or wherever the other fixures end up?

securedrop/tests/functional/sd_config_v2.py
legoktm: That's fair, though I'm concerned about the risks of using a separate config implementation in tests vs what's actually used in the apps, and then having to maintain a parallel implementation of config.

Here's a very naive attempt at what it would take to "fix" the constructor: https://gist.github.com/legoktm/47400d8834b629ba1c452513cc0e6cc3 - then it would be straightforward to have your factory return an instance of the real SDConfig, just constructed differently.

Agreed that changes to SDConfig should be in separate PRs.

legoktm (changes requested):
Most of my feedback is that there are multiple places where in future PRs you move things around or rename them, I think it would be better if we had them at their final destinations/names in the original PR.

nabla-c0d3:
@legoktm I've opened a new PR at https://github.com/freedomofpress/securedrop/pull/6307 to replace this one and address your feedback.

nabla-c0d3:
Replaced by https://github.com/freedomofpress/securedrop/pull/6307

PR #6216, opened 2022-01-05
Support using podman for the dev environment

legoktm:
## Status

Ready for review

## Description of Changes

podman is mostly a drop-in replacement for Docker that's lighter and
better supports rootless execution. It tends to be more widely used
in the Fedora/Red Hat world, though was recently added to Debian. It
does not require running a background daemon as root, providing some
nice security benefits.

By setting a `USE_PODMAN=1` environment variable, the dev-shell and
related scripts will invoke `podman` instead of `docker.

We also need to pass `--userns=keep-id` to `podman run` so that the host
UID/GID are properly mapped in the container so files are writable.

podman doesn't allow deleting/overwriting `/dev/random` so that step is
optional now. I assume that's OK because this is just a dev environment.

This is mostly based on <https://github.com/wikimedia/fresh/commit/d1535438cfd8ccaf39d18fa2fd43350526e43401>,
which I had contributed to/reviewed.

In the future it would be nice to automatically detect if podman is
installed when docker can't be found, but that should probably wait
for more testing/usage under podman.

## Testing

On a Fedora/Debian system, install the `podman` system package. Then run `DOCKER_BUILD_VERBOSE=true USE_PODMAN=1 make dev` and see that it brings up a development environment under podman.

You can also try without the `USE_PODMAN=1` flag to verify nothing changed for normal Docker users.

## Checklist

  - I'll do this if this patch is acceptable, we can mention that podman is experimentally supported behind a `USE_PODMAN=1` flag in the dev docs as an alternative to using Docker.

### DIFF
--- a/securedrop/bin/dev-deps
+++ b/securedrop/bin/dev-deps
@@ -34,8 +34,8 @@ function run_x11vnc() {
 }
 
 function urandom() {
-    sudo rm /dev/random
-    sudo ln -s /dev/urandom /dev/random
+    sudo rm /dev/random ||:
+    sudo ln -s /dev/urandom /dev/random || echo "Unable to replace /dev/random"
 }
 
 function append_to_exit() {
--- a/securedrop/bin/dev-shell
+++ b/securedrop/bin/dev-shell
@@ -11,6 +11,18 @@ export PATH="/opt/venvs/securedrop-app-code/bin:$PATH"
 TOPLEVEL=$(git rev-parse --show-toplevel)
 BASE_OS="${BASE_OS:-focal}"
 USE_TOR="${USE_TOR:-}"
+USE_PODMAN="${USE_PODMAN:-}"
+DOCKER_RUN_ARGUMENTS="${DOCKER_RUN_ARGUMENTS:-}"
+
+# Allow opting into using podman with USE_PODMAN=1
+if  [[ -n "${USE_PODMAN}" ]]; then
+    DOCKER_BIN="podman"
+    # Make sure host UID/GID are mapped into container,
+    # see podman-run(1) manual.
+    DOCKER_RUN_ARGUMENTS="${DOCKER_RUN_ARGUMENTS} --userns=keep-id"
+else
+    DOCKER_BIN="docker"
+fi
 
 ## Get an integer offset for exposed ports, to support multiple containers
 get_port_offset() {
@@ -26,7 +38,7 @@ get_port_offset() {
 }
 
 function docker_image() {
-    docker build \
+    $DOCKER_BIN build \
            ${DOCKER_BUILD_ARGUMENTS:-} \
            --build-arg=USER_ID="$(id -u)" \
            --build-arg=USER_NAME="${USER:-root}" \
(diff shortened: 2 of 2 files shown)


securedrop/bin/dev-deps
```diff
@@ -34,8 +34,8 @@ function run_x11vnc() {
 }
 
 function urandom() {
-    sudo rm /dev/random
-    sudo ln -s /dev/urandom /dev/random
+    sudo rm /dev/random ||:
+    sudo ln -s /dev/urandom /dev/random || echo "Unable to replace /dev/random"
```
zenmonkeykstop, line 38: Haven't checked yet, but one issue with this is that /dev/random is blocking whereas /dev/urandom is not. This could lead to dodgy behaviour on, for example, generating a bunch of test sources and associated keys. It might be worth running loaddata.py with a heckton of sources and replies to see how it behaves compared to the docker version.

conorsch: @zenmonkeykstop Do we still need to clobber random with urandom? A quick spelunking shows we started doing that a very long time ago, and a lot has changed about urandom implementations since then. I'd expect our tests to use urandom just fine without any mucking about on our part.

legoktm: My understanding is that on modern Linux kernels (2020+), /dev/random is mostly the same as /dev/urandom.

See https://lwn.net/Articles/808575/ and https://www.phoronix.com/scan.php?page=news_item&px=Linux-5.6-Random-Rework - the CRNG gathers entropy and blocks at boot, and then it'll no longer block at runtime.

zenmonkeykstop: Said changes apply to 5.6+ apparently, our grsec kernel is currently at 5.4.* - though maybe they've been backported? This doesn't apply to the dev environment obviously, as it's going to be using the host kernel, but it's worth noting that 5.4.* is still pretty common across the board (e.g. in my debian10-based Qubes dev VM, I'm on 5.4.156). It's probably worth checking said behaviour.

legoktm: Ah yeah, I'm also on 5.4 because of Qubes 4.0. I did a bit of testing with loaddata.py+podman+`/dev/random` as-is using rough benchmarking:

3 sources + 3 journalists (default): ~8.7s, 2.9s per source+journalist pair
15 + 15: ~42.7s, ~2.85s/pair
90 + 90: ~254.6s, ~2.83s/pair

I figured out by the last run that the part that really takes time is the source GPG key generation, but it seems to be pretty constant time-wise.

Two other options besides soft failing:
* Get rid of this entirely, maybe in a separate PR, as unnecessary
* Wrap it in a check for `USE_PODMAN=1` so nothing changes for Docker users

ninavizz:
FYI, Macs don't support Podman (or don't support it well—I've tried)—which has been a problem for me, working on the Qubes OS stuff. Just FYI, pls to not remove Docker support with this great PR! :)

legoktm:
> FYI, Macs don't support Podman (or don't support it well—I've tried)—which has been a problem for me, working on the Qubes OS stuff. Just FYI, pls to not remove Docker support with this great PR! :)

Yep, definitely no plans to remove Docker support. It's much more ubiquitous compared to podman. And since podman aims to be CLI-compatible with Docker, it shouldn't add much more work to support both with mostly the same code.

zenmonkeykstop (approved):
Tested in Debian 11 (kernel 5.10.0), `podman` version of dev environment works. Checked generating 100 sources with a reply each via the `podman` and `docker` versions of the environment, time/source was pretty much the same at 2.45sec vs 2.37sec.

LGTM!

PR #6217, opened 2022-01-10
Upgrade Flask dependency to 2.0.2 and refactor applications to improve static type checks

### LINKED ISSUE #6154: Update Flask version to 2.0.*, along with associated requirements
## Description

Core requirements (Flask, Werkzeug, Jinja, etc) are lagging behind their most up-to-date available versions by quite a bit. This complicates the process of upgrading other requirements with shared dependencies (e.g. the `click` package).

An attempt at updating these requirements to recent versions has been started in https://github.com/freedomofpress/securedrop/tree/add-https-to-dev - this has uncovered some juicy issues:

- some application requirements are duplicated in test requirements, and as test requirements are installed *last* in the dev Docker container, application-side updates to those requirements are not reflected in dev (current workaround in branch is to swap the order in which requirements are installed)
- `Pylint` now complains with an `E0237: assgning-non-slot` error  when attributes are assigned to the flask.g object - see https://github.com/pallets/flask/issues/4020 for more. (The simplest workaround for this is to disable pylint assigning-non-slot warnings, either selectively or globally)
- There are some new `mypy` errors of note:
  - Adding attributes to the Flask object (to make them "global") throws an error like `error: "Flask" has no attribute "instance_config"` - see https://github.com/pallets/flask/issues/4167 for an explanation
  - stricter type checking is causing errors like [this one](https://app.circleci.com/pipelines/github/freedomofpress/securedrop/3162/workflows/73985030-c78d-4f4e-a125-48303e1405b9/jobs/57561?invite=true&invite=true#step-110-415) 

A more informed attempt would probably start by cleaning up the existing requirements so that the test requirements are in sync with securedrop-app-code requirements, and then address the mypy globals issue, since that is likely to have the biggest impact on application code.

Prerequisite tasks for this dependency update:
- [x] #6160
- [ ] Update flask-babel: remove refs to `app.babel_instance` (done in `add-https-to-dev` branch)
- [x] #6201
- [x] #6202
- [ ] #6203

zenmonkeykstop:
## Status

Ready for review (but test plan TK)

## Description of Changes

Fixes #6154.

This PR updates the application Flask version to 2.0.2 and updates associated dependencies. It also refactors out the app.storage and app.instance_config dynamic attributes and corrects several typing errors, in order to allow for successul static typechecking via mypy.

- Dependency updates:
  - click from 6.7 to 8.0.3
  - flask-babel from 011.2 to 2.0.0
  - flask-wtf from 0.14.2 to 1.0.0
  - Flask from 1.0.2 to 2.0.2
  - itsdangerous from 0.24 to 2.0.1
  - jinja2 from 2.11.3 to 3.0.2
  - markupsafe from 1.1.1 to 2.0.1
  - redis from 3.3.6 to 3.5.3
  - rq from 1.1.0 to 1.10.0
  - werkzeug from 0.16.0 to 2.0.2

### Application refactor:
- Flask app.storage attribute has been removed in favour of a global Storage object, accessed via a `.get_default()` class method that creates the global object if it doesn't already exist, and then returns it.
- Flask app.instance_config attribute has been removed in favour of a global InstanceConfig object, accessed similarly to Storage.
- Numerous typing fixes have been made.
- datetime variables have been updated to be timezone-aware as required by Flask 2.0.2

### Test updates:
Tests have been updated to account for the refactor above. An app_storage fixture was added, returning a function-scoped Storage object. For tests which access storage directly, that fixture was added to their args and references like `current_app.storage.whatever()` were changed to `app_storage.whatever()`. For tests that access storage indirectly (for example, via a test journalist or source application),  the `Storage.get_default()` method must be mocked to return the `app_storage fixture`. This patch has been added in the `source_app` and `journalist_app` fixtures, so it will be applied automatically in tests that use them.

### General notes:
- There a few TODOs not addressed by this PR for various reasons
  - ~The `babel_instance` Flask app dynamic attribute has not been refactored out - it's added by `flask-babel` and an upstream change is probably required~ Fixed in #6220  
  - type checking is disabled for the CSRFError error handlers due to an issue similar to: https://github.com/pallets/flask/issues/4295
  - type checking is [disabled in the generic error handler ](https://github.com/freedomofpress/securedrop/pull/6217/commits/bae4effa98a9f9133fbb215ddbc0f003e97233c5#diff-24b8e3317d6c0ff33dfe29a1c3edfbb95a6ebfac51562e105f9214c55d1e955cR91) - error handling should be updated to remove the use of Flask's internal `app.error_handler_spec` structure
  - during initial testing, it was found that test requirements would override app requirements in the test Docker environment, meaning that updates to app requirements weren't always being applied. As a quick fix, the order in which requirements were installed was switched, so that app requirements would always be applied last. A better fix would be to clean up those requirements files to either keep them in sync or to remove application dependencies from the test requirements files altogether.

#### Secure temporary files    
                                                    
- check out this branch and run  `make dev-tor` (some latency will help when observing the tempfile behaviour)                          
- in another teminal, connect to the dev env  with `docker exec -t securedrop-dev-0 bash`                           
- in the dev env, monitor /tmp with `watch -n 1 ls -l /tmp`                                       
- connect to the Source Interface(SI) (you'll find the address listed in docker output just before the applications start) and upload a file greater than 512k.                                
  - [x] confirm that 2 files with a `.aes` extension are written and then deleted as the upload completes (the first is the gzipped initial submission, second is decompressed version)
- (optional) for bonus points, retain both files by adding a `delete=False` argument to the super().__init() in the SecureTemporaryFile class and submitting another file larger than 512k
  - [x] verify that both files contents "look" random. 
  - [x] verify that the size of the first file matches the gzipped size of the submission and the size of the second matches that of the original file.
                                                                                
#### Submissions and replies                                                       
- are files being saved to the right location?                                 
  - check out this branch and run  `make dev`                       
  - in another teminal, connect to the dev env  with `docker exec -t securedrop-dev-0 bash`       
  - in  the SI, start at least 2 source sessions, upload distinct files to both    
  - [x] verify that they were stored to the correct `/var/lib/securedrop/store/<UUID>/` paths for the respective sources
- are they being attributed to the correct source in the JI?                  
  - [x] in the Journalist Interface(JI), verify that the submissions are listed under the correct source page
- are replies only visible to the intended recipient in the SI?               
  - in the JI, reply to both sources                                                     
  - [x] in SI, confirm for each source that they can see their own reply only     
- are they encrypted?                
    - [x] in the JI, download the submissions and confirm that they can be decrypted with the submission key
- are they being deleted correctly (individually, en masse, or as part of a source wipe)
  - [x] delete one source's submissions and replies via the "files and messages" only option, confirm they are no longer present
  - [x] delete an individual submission from another source, confirm that the file was removed from the store
  - [x] delete the source in its entirety, confirm that the source store directory was deleted
- are source store directories being recreated correctly?   
  - in the SI, create another source and submit some docs/messages.                  
  - delete all files and messages from a source while preserving the source itself
  - in the dev env shell, manually delete the source's store directory                    
  - [x] attempt to submit a new doc as that source and confirm that the directory is recreated and the source submission is present
                                                                                
                                                                                
### Instance config tests                            
- check out this branch and run  `make dev`, then log in to the JI                                                 
 - on the **Admin > Instance Config** page, update the instance name.
   - [x] verify that the name is saved successfully and that the JI titles change to reflect it.
   - [x] load the SI and verify that the new name is used there as well.
 - on the **Admin > Instance Config** page, disable file uploads.
   - [x] load the SI and verify that sources only have the option to submit messages, not files.
                                                                                
### End-to-end tests                                                            
-  Perform as much of the [Acceptance Tests section ](https://github.com/freedomofpress/securedrop/wiki/2.1.0-Test-Plan#application-acceptance-testing)of the `2.1.0` test plan as you have time and patience for.
    - [x] verify that acceptance tests pass.
                                                                                
### Upgrade scenario         
- Follow the [upgrade scenario docs](https://docs.securedrop.org/en/stable/development/upgrade_testing.html#upgrade-testing-using-locally-built-packages)  to verify an upgrade from 2.1.0 to locally-built packages.      
  - [ ] verify that the upgrade scenario completes successfully       
  - [ ] verify that basic functionality is available after the upgrade.                                
  
## Deployment

These application changes will be deployed with the next release containing them. There are no schema changes or data migrations, but 2 of the alembic migration scripts were modified to reference the new global Storage object - as such, attention should be paid during testing to QA scenarios involving upgrades to verify that they work as expected.

- [ ] I have written a test plan and validated it for this PR **TK**

Choose one of the following:

- `admin/requirements.in`
- `securedrop/requirements/python3/securedrop-app-code-requirements.in`

(remove) this section.

Choose one of the following:

- [x] I would like someone else to do the diff review - given the scope, looking for volunteers. So far we have:
  - [x] click from 6.7 to 8.0.3 - @conorsch
  - [x] flask-babel from 011.2 to 2.0.0 - @cfm
  - [x] flask-wtf from 0.14.2 to 1.0.0 - @zenmonkeykstop
  - [x] Flask from 1.0.2 to 2.0.2 - @zenmonkeykstop
  - [x] itsdangerous from 0.24 to 2.0.1 - @conorsch
  - [x] jinja2 from 2.11.3 to 3.0.2 - @legoktm
  - [x] redis from 3.3.6 to 3.5.3 - @legoktm
  - [x] rq from 1.1.0 to 1.10.0 - @eaon

### DIFF
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -21,7 +21,6 @@ from journalist_app.utils import (get_source, logged_in,
                                   JournalistInterfaceSessionInterface,
                                   cleanup_expired_revoked_tokens)
 from models import InstanceConfig, Journalist
-from store import Storage
 
 import typing
 # https://www.python.org/dev/peps/pep-0484/#runtime-or-type-checking
@@ -68,10 +67,6 @@ def create_app(config: 'SDConfig') -> Flask:
     app.config['SQLALCHEMY_TRACK_MODIFICATIONS'] = False
     app.config['SQLALCHEMY_DATABASE_URI'] = config.DATABASE_URI
     db.init_app(app)
-
-    # TODO: Attaching a Storage dynamically like this disables all type checking (and
-    # breaks code analysis tools) for code that uses current_app.storage; it should be refactored
-    app.storage = Storage(config.STORE_DIR, config.TEMP_DIR)
 
     @app.errorhandler(CSRFError)
     def handle_csrf_error(e: CSRFError) -> 'Response':
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -138,17 +138,18 @@ def create_app(config: 'SDConfig') -> Flask:
 
         uid = session.get('uid', None)
         if uid:
-            g.user = Journalist.query.get(uid)
+            g.user = Journalist.query.get(uid)  # pylint: disable=assigning-non-slot
 
         i18n.set_locale(config)
 
         if app.instance_config.organization_name:
-            g.organization_name = app.instance_config.organization_name
+            g.organization_name = \
+                app.instance_config.organization_name  # pylint: disable=assigning-non-slot
         else:
-            g.organization_name = gettext('SecureDrop')
+            g.organization_name = gettext('SecureDrop')  # pylint: disable=assigning-non-slot
 
         try:
-            g.logo = get_logo_url(app)
+            g.logo = get_logo_url(app)  # pylint: disable=assigning-non-slot
         except FileNotFoundError:
             app.logger.error("Site logo not found.")
 
@@ -161,8 +162,8 @@ def create_app(config: 'SDConfig') -> Flask:
         if request.method == 'POST':
             filesystem_id = request.form.get('filesystem_id')
             if filesystem_id:
-                g.filesystem_id = filesystem_id
-                g.source = get_source(filesystem_id)
+                g.filesystem_id = filesystem_id  # pylint: disable=assigning-non-slot
+                g.source = get_source(filesystem_id)  # pylint: disable=assigning-non-slot
 
         return None
 
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -68,7 +68,9 @@ def create_app(config: 'SDConfig') -> Flask:
     app.config['SQLALCHEMY_DATABASE_URI'] = config.DATABASE_URI
     db.init_app(app)
 
-    @app.errorhandler(CSRFError)
+    # TODO: enable type checking once upstream Flask fix is available. See:
+    # https://github.com/pallets/flask/issues/4295
+    @app.errorhandler(CSRFError)  # type: ignore
     def handle_csrf_error(e: CSRFError) -> 'Response':
         app.logger.error("The CSRF token is invalid.")
         session.clear()
@@ -81,9 +83,12 @@ def create_app(config: 'SDConfig') -> Flask:
     ) -> 'Tuple[Union[Response, str], Optional[int]]':
         # Workaround for no blueprint-level 404/5 error handlers, see:
         # https://github.com/pallets/flask/issues/503#issuecomment-71383286
+        # TODO: clean up API error handling such that all except 404/5s are
+        # registered in the blueprint and 404/5s are handled at the application
+        # level.
         handler = list(app.error_handler_spec['api'][error.code].values())[0]
         if request.path.startswith('/api/') and handler:
-            return handler(error)
+            return handler(error)  # type: ignore
 
         return render_template('error.html', error=error), error.code
 
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -111,10 +111,6 @@ def create_app(config: 'SDConfig') -> Flask:
         cleanup_expired_revoked_tokens()
 
     @app.before_request
-    def load_instance_config() -> None:
-        app.instance_config = InstanceConfig.get_current()
-
-    @app.before_request
     def setup_g() -> 'Optional[Response]':
         """Store commonly used values in Flask's special g object"""
         if 'expires' in session and datetime.now(timezone.utc) >= session['expires']:
@@ -142,9 +138,9 @@ def create_app(config: 'SDConfig') -> Flask:
 
         i18n.set_locale(config)
 
-        if app.instance_config.organization_name:
+        if InstanceConfig.get_default().organization_name:
             g.organization_name = \
-                app.instance_config.organization_name  # pylint: disable=assigning-non-slot
+                InstanceConfig.get_default().organization_name  # pylint: disable=assigning-non-slot
         else:
             g.organization_name = gettext('SecureDrop')  # pylint: disable=assigning-non-slot
 
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -111,6 +111,10 @@ def create_app(config: 'SDConfig') -> Flask:
         cleanup_expired_revoked_tokens()
 
     @app.before_request
+    def update_instance_config() -> None:
+        InstanceConfig.get_default(refresh=True)
+
+    @app.before_request
     def setup_g() -> 'Optional[Response]':
         """Store commonly used values in Flask's special g object"""
         if 'expires' in session and datetime.now(timezone.utc) >= session['expires']:
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 
 from flask import (Flask, session, redirect, url_for, flash, g, request,
@@ -117,7 +117,7 @@ def create_app(config: 'SDConfig') -> Flask:
     @app.before_request
     def setup_g() -> 'Optional[Response]':
         """Store commonly used values in Flask's special g object"""
-        if 'expires' in session and datetime.utcnow() >= session['expires']:
+        if 'expires' in session and datetime.now(timezone.utc) >= session['expires']:
             session.clear()
             flash(gettext('You have been logged out due to inactivity.'),
                   'error')
@@ -131,7 +131,7 @@ def create_app(config: 'SDConfig') -> Flask:
                 flash(gettext('You have been logged out due to password change'),
                       'error')
 
-        session['expires'] = datetime.utcnow() + \
+        session['expires'] = datetime.now(timezone.utc) + \
             timedelta(minutes=getattr(config,
                                       'SESSION_EXPIRATION_MINUTES',
                                       120))
--- a/securedrop/journalist_app/col.py
+++ b/securedrop/journalist_app/col.py
@@ -138,8 +138,8 @@ def make_blueprint(config: SDConfig) -> Blueprint:
                 reply = Reply.query.filter(Reply.filename == fn).one()
                 mark_seen([reply], journalist)
             elif fn.endswith("-doc.gz.gpg") or fn.endswith("doc.zip.gpg"):
-                the_file = Submission.query.filter(Submission.filename == fn).one()
-                mark_seen([the_file], journalist)
+                submitted_file = Submission.query.filter(Submission.filename == fn).one()
+                mark_seen([submitted_file], journalist)
             else:
                 message = Submission.query.filter(Submission.filename == fn).one()
                 mark_seen([message], journalist)
--- a/securedrop/journalist_app/col.py
+++ b/securedrop/journalist_app/col.py
@@ -29,6 +29,7 @@ from journalist_app.utils import (make_star_true, make_star_false, get_source,
                                   col_download_all, col_star, col_un_star,
                                   col_delete, col_delete_data, mark_seen)
 from sdconfig import SDConfig
+from store import Storage
 
 
 def make_blueprint(config: SDConfig) -> Blueprint:
@@ -118,7 +119,7 @@ def make_blueprint(config: SDConfig) -> Blueprint:
         if '..' in fn or fn.startswith('/'):
             abort(404)
 
-        file = current_app.storage.path(filesystem_id, fn)
+        file = Storage.get_default().path(filesystem_id, fn)
         if not Path(file).is_file():
             flash(
                 gettext(
@@ -145,7 +146,7 @@ def make_blueprint(config: SDConfig) -> Blueprint:
         except NoResultFound as e:
             current_app.logger.error("Could not mark {} as seen: {}".format(fn, e))
 
-        return send_file(current_app.storage.path(filesystem_id, fn),
+        return send_file(Storage.get_default().path(filesystem_id, fn),
                          mimetype="application/pgp-encrypted")
 
     return view
--- a/securedrop/journalist_app/col.py
+++ b/securedrop/journalist_app/col.py
@@ -138,8 +138,8 @@ def make_blueprint(config: SDConfig) -> Blueprint:
                 reply = Reply.query.filter(Reply.filename == fn).one()
                 mark_seen([reply], journalist)
             elif fn.endswith("-doc.gz.gpg") or fn.endswith("doc.zip.gpg"):
-                file = Submission.query.filter(Submission.filename == fn).one()
-                mark_seen([file], journalist)
+                the_file = Submission.query.filter(Submission.filename == fn).one()
+                mark_seen([the_file], journalist)
             else:
                 message = Submission.query.filter(Submission.filename == fn).one()
                 mark_seen([message], journalist)
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -895,9 +895,10 @@ class InstanceConfig(db.Model):
         return new
 
     @classmethod
-    def get_default(cls) -> "InstanceConfig":
+    def get_default(cls, refresh: Optional[bool] = False) -> "InstanceConfig":
         global _default_instance_config
-        _default_instance_config = InstanceConfig.get_current()
+        if (_default_instance_config is None) or (refresh is not None and refresh is True):
+            _default_instance_config = InstanceConfig.get_current()
         return _default_instance_config
 
     @classmethod
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -194,12 +194,11 @@ class Submission(db.Model):
     '''
     checksum = Column(String(255))
 
-    def __init__(self, source: Source, filename: str) -> None:
+    def __init__(self, source: Source, filename: str, storage: Storage) -> None:
         self.source_id = source.id
         self.filename = filename
         self.uuid = str(uuid.uuid4())
-        self.size = os.stat(Storage.get_default().path(source.filesystem_id,
-                                                       filename)).st_size
+        self.size = os.stat(storage.path(source.filesystem_id, filename)).st_size
 
     def __repr__(self) -> str:
         return '<Submission %r>' % (self.filename)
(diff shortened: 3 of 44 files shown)


securedrop/journalist_app/__init__.py
```diff
@@ -110,14 +110,10 @@ def _handle_http_exception(
     def expire_blacklisted_tokens() -> None:
         cleanup_expired_revoked_tokens()
 
-    @app.before_request
-    def load_instance_config() -> None:
-        app.instance_config = InstanceConfig.get_current()
```
legoktm, line 115: If I'm understanding this correctly, this is a behavior shift now in that InstanceConfig is just initialized once at startup rather than on each request. Is that intentional? I haven't figured out how that works with the "valid_until" field yet...

securedrop/store.py
```diff
@@ -314,9 +329,12 @@ def save_file_submission(self,
                              filesystem_id: str,
                              count: int,
                              journalist_filename: str,
-                             filename: str,
-                             stream: 'BufferedIOBase') -> str:
-        sanitized_filename = secure_filename(filename)
+                             filename: typing.Optional[str],
+                             stream: 'IO[bytes]') -> str:
+
+        sanitized_filename = secure_filename("unknown.file")
```
legoktm: I think this would be easier to read if it was in an else block below.

securedrop/journalist_app/col.py
```diff
@@ -137,15 +138,15 @@ def download_single_file(filesystem_id: str, fn: str) -> werkzeug.Response:
                 reply = Reply.query.filter(Reply.filename == fn).one()
                 mark_seen([reply], journalist)
             elif fn.endswith("-doc.gz.gpg") or fn.endswith("doc.zip.gpg"):
-                file = Submission.query.filter(Submission.filename == fn).one()
-                mark_seen([file], journalist)
+                the_file = Submission.query.filter(Submission.filename == fn).one()
```
legoktm: If we're renaming this maybe a better name would be submitted_file?

This whole block could be more DRY by not duplicating the `mark_seen` call for each if stanza...but that can be done separately.

securedrop/source_app/main.py
```diff
@@ -82,7 +84,7 @@ def create() -> werkzeug.Response:
                 create_source_user(
                     db_session=db.session,
                     source_passphrase=codename,
```
legoktm, line 86: This is also documented as taking a DicewarePassphrase, so I don't understand why mypy is happy with it but below wants you to wrap it. In any case, shouldn't the codename variable be turned into a DicewarePassphrase as soon as its pulled out of the session object?

legoktm (commented):
As I mentioned earlier, I think this could be split a bit to make easier to review:
* datetime changes: utcnow() -> now(utc)
* Introduction of and switch to Storage.get_default()
* Introduction of and switch to InstanceConfig.get_default()

zenmonkeykstop:
> As I mentioned earlier, I think this could be split a bit to make easier to review:
> 
>     * datetime changes: utcnow() -> now(utc) 
>     * Introduction of and switch to Storage.get_default()
>     * Introduction of and switch to InstanceConfig.get_default()

This makes sense - I've reordered and merged commits to simplify things and group related changes. Let me know if it's helpful!

securedrop/journalist_app/__init__.py
zenmonkeykstop: Actually the instance config is now retrieved every time InstanceConfig.get_default() is called, so it's  actually hitting the DB as much as before (or potentially more often if a single request has multiple .get_default()s) This reduces the benefit of a global object, so there's probably some opportunity for optimization there. The trick would be to detect configuration changes when they're made in the journalist interface, as we want those to be reflected immediately in the source app.

zenmonkeykstop: Updated the InstanceConfig get_default() method to be lazier - it will now only hit the db if explicitly requested or if the global is `None`. With some tweaks to `@app.before_request`-decorate functions, changes should show up immediately in both apps without extra DB overhead.

Also, re `valid_until` - it might be a bit counter-intuitive. It records the historical info about previous configs, and when they were invalidated. (TBH I'm not sure we need to store said previous configs, but it does provide an audit trail for instance_config changes.)

securedrop/store.py
zenmonkeykstop: sure, updated.

securedrop/journalist_app/col.py
zenmonkeykstop: :+1:  updated (the name, not the mark_seen() calls)

securedrop/requirements/python3/securedrop-app-code-requirements.txt
```diff
@@ -108,33 +108,36 @@ cryptography==3.4.7 \
 flask-assets==0.12 \
     --hash=sha256:6031527b89fb3509d1581d932affa5a79dd348cfffb58d0aef99a43461d47847
     # via -r requirements/python3/securedrop-app-code-requirements.in
-flask-babel==0.11.2 \
-    --hash=sha256:462a3c599b0ccf426ca1757cc612f1db383844efd346d14170da04c8c76dd521 \
-    --hash=sha256:c0d75710bd4b0fe866f9f2347de6e19208712f9cec006436b4c1c15d4cb0c939
+flask-babel==2.0.0 \
+    --hash=sha256:e6820a052a8d344e178cdd36dd4bb8aea09b4bda3d5f9fa9f008df2c7f2f5468 \
+    --hash=sha256:f9faf45cdb2e1a32ea2ec14403587d4295108f35017a7821a2b1acb8cfd9257d
     # via -r requirements/python3/securedrop-app-code-requirements.in
```
cfm, line 114: Diff reviewed in <https://github.com/freedomofpress/securedrop-debian-packaging/wiki/Flask-Babel-0.11.2-to-2.0.0>.

securedrop/journalist_app/__init__.py
```diff
@@ -131,24 +131,25 @@ def setup_g() -> 'Optional[Response]':
...
         i18n.set_locale(config)
 
-        if app.instance_config.organization_name:
-            g.organization_name = app.instance_config.organization_name
+        if InstanceConfig.get_default().organization_name:
+            g.organization_name = \
+                InstanceConfig.get_default().organization_name  # pylint: disable=assigning-non-slot
         else:
-            g.organization_name = gettext('SecureDrop')
+            g.organization_name = gettext('SecureDrop')  # pylint: disable=assigning-non-slot
```
cfm, line 149: Nit in passing:  I wondered if this could be simplified, since [`InstanceConfig.organization_name` has `default="SecureDrop"`](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/models.py#L872), but it looks like that default's not applied in the [upgrade migration](https://github.com/freedomofpress/securedrop/blob/f768a1a5aa37e790077dfd38d6e7a0299a4b3a0f/securedrop/alembic/versions/92fba0be98e9_added_organization_name_field_in_.py#L19-L24).

securedrop/requirements/python3/securedrop-app-code-requirements.txt
```diff
@@ -226,20 +265,22 @@ python-editor==1.0.3 \
...
+    #   babel
+    #   flask-babel
 qrcode==5.3 \
     --hash=sha256:4115ccee832620df16b659d4653568331015c718a754855caf5930805d76924e \
     --hash=sha256:60222a612b83231ed99e6cb36e55311227c395d0d0f62e41bb51ebbb84a9a22b
     # via -r requirements/python3/securedrop-app-code-requirements.in
-redis==3.3.6 \
-    --hash=sha256:45682ecf226c7611efe731974c4fa3390170ba045b9cdb26f0051114a5c2a68b \
-    --hash=sha256:f2609a85e5f37f489ba3b5652e1175dc3711c4d7a7818c4f657615810afd23df
+redis==3.5.3 \
```
legoktm, line 275: Reviewed: https://github.com/freedomofpress/securedrop-debian-packaging/wiki/redis-3.3.6-to-3.5.3

conorsch:
Diff reviews complete for `click` and `itsdangerous`. Regarding the latter, we've got an upcoming deprecation tracked in https://github.com/freedomofpress/securedrop/issues/6224. For the purposes of this PR, see the comments in https://github.com/pallets/itsdangerous/issues/257 about token length; I have not tried to reproduce what's reported there, but given the scope of these changes, makes sense to evaluate as part of review here.

conorsch (dismissed):
Looks like `securedrop/requirements/python3/test-requirements.txt` still needs to be updated; it contains `itsdangerous==0.24` (the old version) and also Flash v1 still.

zenmonkeykstop:
> Looks like `securedrop/requirements/python3/test-requirements.txt` still needs to be updated; it contains `itsdangerous==0.24` (the old version) and also Flash v1 still.

Yup, that's an outstanding task in the TODOs - current workaround also described there.

legoktm:
jinja notes:
* evalcontextfilter is deprecated, there's a new sample nl2br snippet: https://jinja.palletsprojects.com/en/3.0.x/api/?highlight=nl2br#custom-filters
* The with_ and autoescape extensions are deprecated (now always built-in AIUI), so maybe they can be removed from babel.cfg?
* It seems Markup should be imported from markupsafe rather than jinja now
* Do we have a regression test that verifies jinja's autoescaping is turned on? Like `assert Template("{{foo}}").render(foo="<script>") == "&lt;script>"` or something.

I reviewed all the doc and src/ changes, ended up skipping the tests/ changes due to a lack of time.

securedrop/requirements/python3/securedrop-app-code-requirements.txt
```diff
@@ -108,33 +108,36 @@ cryptography==3.4.7 \
...
-jinja2==2.11.3 \
-    --hash=sha256:03e47ad063331dd6a3f04a43eddca8a966a26ba0c5b7207a9a9e4e08f1b29419 \
-    --hash=sha256:a6d58433de0ae800347cab1fa3043cebbabe8baa9d29e668f1c768cb87a333c6
+itsdangerous==2.0.1 \
+    --hash=sha256:5174094b9637652bdb841a3029700391451bd092ba3db90600dea710ba28e97c \
+    --hash=sha256:9e724d68fc22902a1435351f84c3fb8623f303fffcc566a4cb952df8c572cff0
+    # via
+    #   flask
+    #   flask-wtf
+jinja2==3.0.2 \
```
legoktm, line 138: Reviewed: https://github.com/freedomofpress/securedrop-debian-packaging/wiki/Jinja2-2.11.3-to-3.0.2

securedrop/models.py
```diff
@@ -877,6 +879,13 @@ def copy(self) -> "InstanceConfig":
 
         return new
 
+    @classmethod
+    def get_default(cls, refresh: Optional[bool] = False) -> "InstanceConfig":
```
nabla-c0d3: I would put `refresh: bool = False` ; not sure there's a need for refresh to ever be None?

securedrop/tests/test_store.py
```diff
@@ -32,30 +32,29 @@ def create_file_in_source_dir(config, filesystem_id, filename):
     return source_directory, file_path
```
nabla-c0d3, line 32: To further simplify the tests in this file, I would recommend:

- Removing all mentions of the journalist_app, as it is out of scope for what is being tested here (only the Storage class is being tested; not the full app).
- Not using the storage fixture as I would argue that creating a Storage class should be done in the test code as that's what's being tested: it's not "setup" code. The benefit will be that the tests won't be coupled with the config fixture (which is used in a lot of tests and seems unnecessary here) and the tests will be easier to read/follow.

You can see an example of tis approach above in TestPassphrasesGenerator.

securedrop/tests/test_encryption.py
```diff
@@ -17,13 +17,14 @@ def test_get_default(self, config):
...
-    def test_generate_source_key_pair(self, setup_journalist_key_and_gpg_folder, source_app):
+    def test_generate_source_key_pair(self, setup_journalist_key_and_gpg_folder,
+                                      source_app, app_storage):
         # Given a source user
         with source_app.app_context():
             source_user = create_source_user(
                 db_session=db.session,
                 source_passphrase=PassphraseGenerator.get_default().generate_passphrase(),
-                source_app_storage=source_app.storage,
+                source_app_storage=app_storage,
```
nabla-c0d3, line 27: Not super important, but you might now be able to remove the source_app fixture from this test and the one below 🙌

securedrop/models.py
zenmonkeykstop: :heavy_check_mark: - updated.

securedrop/tests/test_store.py
zenmonkeykstop: The `journalist_app` fixture is unused in the majority of the tests alright - there are a few that do require an app context - eg. the shredder tests, but they could probably be moved to a separate file.

I'm less convinced about removing the `app_storage` fixture in all cases. Some of the tests that use it and the `config` fixture are essentially verifying that files are being created in the store in the expected location based on config parameters, and for those it would seem either tautological to just remove `config`, or duplicative to effectively recreate both fixtures in the test. But it can definitely be removed from tests not using `config` and replaced with a test-specific Storage object.

zenmonkeykstop: Actually on further reflection there are a lot more cases where the latter makes sense...

securedrop/tests/test_encryption.py
zenmonkeykstop: Looks like yes to the one below, but no to this one without independently setting up the db and creating an app and passing its context within the test.

securedrop/store.py
```diff
@@ -314,9 +329,13 @@ def save_file_submission(self,
...
-                             filename: str,
-                             stream: 'BufferedIOBase') -> str:
-        sanitized_filename = secure_filename(filename)
+                             filename: typing.Optional[str],
+                             stream: 'IO[bytes]') -> str:
+
+        if filename is not None:
+            sanitized_filename = secure_filename(filename)
+        else:
+            sanitized_filename = secure_filename("unknown.file")
```
nabla-c0d3, line 338: (Not very important) You could simplify this by switching the argument's declaration to:
```
filename: str = "unknown.file",
```
Then no if/else would be needed.

securedrop/requirements/python3/securedrop-app-code-requirements.txt
```diff
@@ -226,20 +265,22 @@ python-editor==1.0.3 \
...
+    --hash=sha256:432b788c4530cfe16d8d943a09d40ca6c16149727e4afe8c2c9d5580c59d9f24
     # via
     #   -r requirements/python3/securedrop-app-code-requirements.in
     #   rq
-rq==1.1.0 \
-    --hash=sha256:2798d26a7b850e759f23f69695a389d676a9c08f2c14f96f0d34d9648c9d5616 \
-    --hash=sha256:4f27c6a690d1bd02b9157e615d8819555b9b359c0c4ec8ff0013d160c31b40bb
+rq==1.10.0 \
+    --hash=sha256:92950a3e60863de48dd1800882939bbaf089a37497ebf9f2ecf7c9fd0a4c4a95 \
+    --hash=sha256:be09ec43fae9a75a4d26ea3cd520e5fa3ea2ea8cf481be33e6ec9416f0369cac
```
eaon, line 283: Reviewed: https://github.com/freedomofpress/securedrop-debian-packaging/wiki/rq-1.1.0-to-1.10.0

securedrop/store.py
zenmonkeykstop: I kindof like it as is, I think @legoktm's previous point about readability is a good one.

legoktm:
I got through the first half of the test plan today, should finish the rest on Monday.

conorsch (approved):
Changes look great. Try as I might, I'm unable to break it, which is grand. =) Tested via the prod VM upgrade scenario, and ran through the full test plan. Diff is quite readable, and post-reorg, the code is a bit more intuitive IMO. Great work on this one, @zenmonkeykstop—and thanks also to @legoktm for detailed review.

PR #6220, opened 2022-01-11
Stop using app.babel_instance

legoktm:
## Status

Ready for review

## Description of Changes

Using dynamically created properties on `app` hurts our ability to
statically type check these code paths, so we want to move away from
them.

Instead, we can pass the Babel instance we just created to the one place
we've been using `app.babel_instance` for basically the same effect.

This is related to the other changes @zenmonkeykstop has been working on as part of the Flask 2.0 upgrade.

## Testing

* Verify that other locales can be switched to without any regression
* Verify that setting the default to a locale there are no translations for continues to throw an exception
 
## Deployment

I do not think there are any special considerations for deployment.

## Checklist

Choose one of the following:

### DIFF
--- a/securedrop/i18n.py
+++ b/securedrop/i18n.py
@@ -101,7 +101,7 @@ class RequestLocaleInfo:
         return get_locale_identifier(parse_locale(str(self.locale)), sep="-")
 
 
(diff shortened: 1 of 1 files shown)


zenmonkeykstop (approved):
Changes look good, test plan checks out!

Issue #6192, opened 2021-12-08
Use a special `deleted` journalist user for associations with deleted users


eloquence:
As discussed in #5467 and #5503, we currently delete rows for deleted journalist users. After deletion, `journalist_id` will be `NULL` in other tables that reference the deleted user (e.g., `replies` and the `seen` tables; see the [database diagram](https://docs.securedrop.org/en/stable/_images/securedrop-database.png)). The API will return the username `deleted` and the UUID `deleted` wherever these users are referenced.

There are a few issues with this:
- It creates an avoidable inconsistency between the data's internal representation and its API representation, which makes it harder to reason about where the data returned by the API originates
- We cannot rely on the `/users` endpoint of the API to get a complete list of users that an API consumer may need to associate data with
- User management in the SecureDrop Client does not map directly to user management on the server

In #5467, we considered preserving a record for each user, while removing all data _about_ the user. This may be desirable in the long run (alongside features such as user locking: #3926), but even if we do this, we will need a migration for historical data. The solution we've settled on for now is to create a global `deleted` user. A database migration will need to a) create this user, b) associate records currently associated with `NULL` users with this new user.

This puts us in a good position to make other changes like #5467 in future, and unblocks further work on SecureDrop Client features that rely on journalist attribution.

sssoleileraaa:
I think there are two approaches, but first here are some notes (recap):

## Notes

* The endpoint `/get_all_replies` tells us which replies exist on the server (so that we can add, update, or delete reply records accordingly to match the server), and which journalists have seen each reply (so that we can add seen records accordingly- note that we decided early on that we only delete seen records when the source or conversation are deleted).

* The only reason a reply should be updated is because the reply `filename`, `size`, or `journalist_id` has changed. (We can remove the update logic around `journalist_id` if we go with Option B take below.)

* The endpoint only sends UUIDs (IDs are partial PII), which is why "deleted" is a [hard-coded string for `journalist_uuid` in the API endpoint](https://github.com/freedomofpress/securedrop/blob/c5d7766d886b6bf3c9fdc297e4bb21b960e58447/securedrop/models.py#L293)

[...]

PR #6223, opened 2022-01-12
Get rid of the need to explicitly flush the database session

legoktm:
## Status

Ready for review

## Description of Changes

From what I can tell, nearly all the calls to db.session.flush() have
been introduced so that a newly created object's autoincrement ID can be
referenced in a second object to be created, most often this is creating
a Reply, and then a corresponding SeenReply.

However, if we just pass the Reply object into SeenReply, rather than
the id (which is None if it hasn't been flushed), SQLAlchemy takes care
of it all for us, properly assigning the correct autoincrement ID using
the specified relationship.

The motivation for doing this is that JournalistLoginAttempt in loaddata
was not properly flushing and was being assigned to a
journalist_id=NULL, which will soon be an error as part of #6192. Rather
than stick another flush in to fix that case, we can just get rid of
nearly all of them.

There is still one explicit flush() left in loaddata.py's
record_source_interaction(), but that's a different pattern than these
so I've left it alone for now.

I am splitting this out of my deleted-journalist work because it ends up touching a pretty different area of code (saving replies) compared to the rest of the branch.

## Testing

* Run loaddata.py, check in the database that there are no journalist_id=NULL being inserted, nor reply_id=NULL.
  * In my deleted-journalist branch there will soon be a migration to make all those fields nullable=False, which is how I tested this, that no exceptions were thrown .
* Leave a reply, observe that it correctly is marked as seen.

## Deployment

## Checklist

### DIFF
--- a/securedrop/journalist_app/api.py
+++ b/securedrop/journalist_app/api.py
@@ -278,8 +278,7 @@ def make_blueprint(config: SDConfig) -> Blueprint:
 
             try:
                 db.session.add(reply)
-                db.session.flush()
-                seen_reply = SeenReply(reply_id=reply.id, journalist_id=user.id)
+                seen_reply = SeenReply(reply=reply, journalist=user)
                 db.session.add(seen_reply)
                 db.session.add(source)
                 db.session.commit()
--- a/securedrop/journalist_app/main.py
+++ b/securedrop/journalist_app/main.py
@@ -141,8 +141,7 @@ def make_blueprint(config: SDConfig) -> Blueprint:
         try:
             reply = Reply(g.user, g.source, filename)
             db.session.add(reply)
-            db.session.flush()
-            seen_reply = SeenReply(reply_id=reply.id, journalist_id=g.user.id)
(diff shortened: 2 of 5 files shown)


zenmonkeykstop (approved):
- [x] Ran through the upgrade scenario just because this touched models.py, no issues there.
- [ ] Run loaddata.py, check in the database that there are no journalist_id=NULL being inserted, nor reply_id=NULL. *Failed for me, but I think it's a test plan problem, not a code problem. The loaddata.py script adds and then deletes a user, so you would expect to see null journalist_ids (about a third of them) in the seen_replies table. I think this would need to be rechecked with the changes in #6225 for it to pass.*
        In my deleted-journalist branch there will soon be a migration to make all those fields nullable=False, which is how I tested this, that no exceptions were thrown .
- [x] Leave a reply, observe that it correctly is marked as seen.

Approving/merging  with the proviso about `loaddata.py` and nulls.

legoktm:
Yeah, you're right about the test plan, there's no practical way to test that part of it without #6225.

PR #6225, opened 2022-01-14
Use special "deleted" journalist for associations with deleted users

### LINKED ISSUE #6192: Use a special `deleted` journalist user for associations with deleted users

legoktm:
## Status

Ready for review. Note that this depends on #6223 to work properly.

## Description of Changes

Currently when a journalist is deleted, most referential tables are
updated with `journalist_id=NULL`, forcing all clients to accomodate
that case. Instead, we are now going to either re-associate those rows
with a special "deleted" journalist account, or delete them outright.
All journalist_id columns are now NOT NULL to enforce this.

Tables with rows migrated to "deleted":
* replies
* seen_files
* seen_messages
* seen_replies

Tables with rows that are deleted outright:
* journalist_login_attempt
* revoked_tokens

The "deleted" journalist account is a real account that exists in the
database, but cannot be logged into and has no passphase set. It is not
possible to delete it nor is it shown in the admin listing of
journalists. It is lazily created on demand using a special
DeletedJournalist subclass that bypasses username and passphrase
validation.

Because we now have a real user to reference, the api.single_reply
endpoint will return a real UUID instead of the "deleted" placeholder.

Journalist objects must now be deleted by calling the new delete()
function on them. Trying to directly `db.session.delete(journalist)`
will most likely fail with an Exception because of rows that weren't
migrated first.

The migration step looks for any existing rows in those tables with
journalist_id=NULL and either migrates them to "deleted" or deletes
them. Then all the columns are changed to be NOT NULL.

Fixes #6192.

## Testing

* Start the dev environment, which calls loaddata.py. Doing so creates a journalist ("clarkkent"), adds some replies, and then immediately deletes it
* Using the same dev environment, visit the admin page and notice that no "deleted" journalist is shown.
* Create a journalist, login as them, have them reply, then delete them (so journalist_id=NULL exists in the DB). Then apply this patch, run the upgrader and observe in the DB that a "deleted" user is created and all the replies, seen_* rows have been updated to point to it
*  Hit the API single_reply endpoint and see that you still get "deleted" for username but a real UUID

## Deployment

This has a schema change and data migration for existing installs.

## Checklist

- [x] These changes do not require documentation (correct me if I'm wrong please)

### DIFF
--- a/securedrop/journalist_app/admin.py
+++ b/securedrop/journalist_app/admin.py
@@ -30,7 +30,7 @@ def make_blueprint(config: SDConfig) -> Blueprint:
     @view.route('/', methods=('GET', 'POST'))
     @admin_required
     def index() -> str:
-        users = Journalist.query.all()
+        users = Journalist.query.filter(Journalist.username != "deleted").all()
         return render_template("admin.html", users=users)
 
     @view.route('/config', methods=('GET', 'POST'))
@@ -288,16 +288,22 @@ def make_blueprint(config: SDConfig) -> Blueprint:
             current_app.logger.error(
                 "Admin {} tried to delete itself".format(g.user.username))
             abort(403)
-        elif user:
-            db.session.delete(user)
-            db.session.commit()
-            flash(gettext("Deleted user '{user}'.").format(
-                user=user.username), "notification")
-        else:
+        elif not user:
             current_app.logger.error(
                 "Admin {} tried to delete nonexistent user with pk={}".format(
                     g.user.username, user_id))
             abort(404)
+        elif user.is_deleted_user():
+            # Do not flash because the interface does not expose this.
+            # It can only happen by manually crafting a POST request
+            current_app.logger.error(
+                "Admin {} tried to delete \"deleted\" user".format(g.user.username))
+            abort(403)
+        else:
+            user.delete()
+            db.session.commit()
+            flash(gettext("Deleted user '{user}'.").format(
+                user=user.username), "notification")
 
         return redirect(url_for('admin.index'))
 
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -30,6 +30,7 @@ from logging import Logger
 from pyotp import TOTP, HOTP
 
 from encryption import EncryptionManager, GpgKeyNotFoundError
+from passphrases import PassphraseGenerator
 from store import Storage
 
 _default_instance_config: Optional["InstanceConfig"] = None
@@ -256,7 +257,7 @@ class Reply(db.Model):
     id = Column(Integer, primary_key=True)
     uuid = Column(String(36), unique=True, nullable=False)
 
-    journalist_id = Column(Integer, ForeignKey('journalists.id'))
+    journalist_id = Column(Integer, ForeignKey('journalists.id'), nullable=False)
     journalist = relationship(
         "Journalist",
         backref=backref(
@@ -296,18 +297,8 @@ class Reply(db.Model):
         return '<Reply %r>' % (self.filename)
 
     def to_json(self) -> 'Dict[str, Any]':
-        journalist_username = "deleted"
-        journalist_first_name = ""
-        journalist_last_name = ""
-        journalist_uuid = "deleted"
-        if self.journalist:
-            journalist_username = self.journalist.username
-            journalist_first_name = self.journalist.first_name
-            journalist_last_name = self.journalist.last_name
-            journalist_uuid = self.journalist.uuid
         seen_by = [
             r.journalist.uuid for r in SeenReply.query.filter(SeenReply.reply_id == self.id)
-            if r.journalist
         ]
         json_reply = {
             'source_url': url_for('api.single_source',
@@ -317,10 +308,10 @@ class Reply(db.Model):
                                  reply_uuid=self.uuid) if self.source else None,
             'filename': self.filename,
             'size': self.size,
-            'journalist_username': journalist_username,
-            'journalist_first_name': journalist_first_name,
-            'journalist_last_name': journalist_last_name,
-            'journalist_uuid': journalist_uuid,
+            'journalist_username': self.journalist.username,
+            'journalist_first_name': self.journalist.first_name or '',
+            'journalist_last_name': self.journalist.last_name or '',
+            'journalist_uuid': self.journalist.uuid,
             'uuid': self.uuid,
             'is_deleted_by_source': self.deleted_by_source,
             'seen_by': seen_by
@@ -436,10 +427,17 @@ class Journalist(db.Model):
     )  # type: Column[Optional[datetime.datetime]]
     last_access = Column(DateTime)  # type: Column[Optional[datetime.datetime]]
     passphrase_hash = Column(String(256))  # type: Column[Optional[str]]
+
     login_attempts = relationship(
         "JournalistLoginAttempt",
-        backref="journalist"
+        backref="journalist",
+        cascade="all, delete"
     )  # type: RelationshipProperty[JournalistLoginAttempt]
+    revoked_tokens = relationship(
+        "RevokedToken",
+        backref="journalist",
+        cascade="all, delete"
+    )  # type: RelationshipProperty[RevokedToken]
 
     MIN_USERNAME_LEN = 3
     MIN_NAME_LEN = 0
@@ -713,8 +711,7 @@ class Journalist(db.Model):
         except NoResultFound:
             raise InvalidUsernameException(gettext("Invalid username"))
 
-        if user.username in Journalist.INVALID_USERNAMES and \
-                user.uuid in Journalist.INVALID_USERNAMES:
+        if user.username in Journalist.INVALID_USERNAMES:
             raise InvalidUsernameException(gettext("Invalid username"))
 
         if len(user.otp_secret) < OTP_SECRET_MIN_ASCII_LENGTH:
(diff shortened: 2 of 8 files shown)


legoktm:
Is there a way to suppress the lgtm warning? The thing it's complaining about is intentional. And codecov is claiming that the migration script isn't covered, but AFAICT the test I wrote for it is running and works. The other new thing it said wasn't covered I just added a test for.

securedrop/journalist_app/admin.py
```diff
@@ -288,16 +288,22 @@ def delete_user(user_id: int) -> werkzeug.Response:
...
-            flash(gettext("Deleted user '{user}'.").format(
-                user=user.username), "notification")
-        else:
+        elif not user:
             current_app.logger.error(
                 "Admin {} tried to delete nonexistent user with pk={}".format(
                     g.user.username, user_id))
             abort(404)
+        elif user.username == "deleted":
+            # Do not flash because the interface does not expose this.
```
nabla-c0d3, line 297: I would recommend using an explicit field on the Journalist model to identify the deleted user, such as `is_special_deleted_journalist: bool` (and it can have some kind of unique constraint).

Using the username might cause problems later: what if an instance (somehow) has a journalist named "deleted", etc.

securedrop/models.py
```diff
@@ -784,13 +790,57 @@ def to_json(self, all_info: bool = True) -> Dict[str, Any]:
 
         return json_user
 
+    def delete(self) -> None:
```
nabla-c0d3, line 811: I would recommend explicitly passing the db_session as an argument:

- It will make testing easier.
- Importing the db object/module directly creates the implicit assumption that the caller has setup the special (flask) context where this import will work. This is usually fine in the running app, but can be tricky in utility scripts and in the test suite. Making the db_session an argument makes this requirement explicit/documented, and leaves it to the caller to get the db session.

securedrop/models.py
```diff
@@ -294,18 +295,16 @@ def __repr__(self) -> str:
...
-        if self.journalist:
-            journalist_username = self.journalist.username
-            journalist_first_name = self.journalist.first_name
-            journalist_last_name = self.journalist.last_name
+        journalist_username = self.journalist.username
+        journalist_first_name = self.journalist.first_name or ""
+        journalist_last_name = self.journalist.last_name or ""
+        if journalist_username == "deleted":
+            # special-case the "deleted" user
+            journalist_uuid = "deleted"
```
nabla-c0d3: I think this will break the API contract, as it's not a valid uuid. It will require clients to implement custom parsing logic just for this field. I would recommend returning a valid uuid instead (that points to the deleted journalist).

securedrop/models.py
```diff
@@ -446,7 +452,7 @@ class Journalist(db.Model):
 
     def __init__(self,
                  username: str,
-                 password: str,
+                 password: 'Optional[str]',
```
nabla-c0d3: I think this make the Journalist class' API less clear: the reader might assume that for any journalist, the password is optional. My recommendation is to leave the password as a non-optional str, and move the logic to handle the special case of the deleted journalist somewhere else (so that the special case doesn't "pollute" the "general" API).

securedrop/journalist_app/admin.py
legoktm: SecureDrop already prevents creating a user with the name "deleted" so I don't think there's any worry with collisions. I don't see much value in having a separate column for it either, but I do like the suggestion to have a specific method for it so we're not hardcoding the username everywhere

```python
def is_deleted_user(self) -> bool:
    return self.username == "deleted"
```

securedrop/models.py
legoktm: All the other methods in Journalist don't do this, so I'd prefer to keep it consistent in the class for now. What you explained seems reasonable to me, I just think it should be done separately.

securedrop/models.py
legoktm: This is matching the current behavior of the API, which returns "deleted" for the UUID of a null journalist. I agree that it should return a real UUID instead as I hinted in my PR message, but AIUI it needs coordination with the workstation.

securedrop/models.py
legoktm: Good point. I think we can have DeletedJournalist just pass in `password='unused'` since set_password() is a no-op. And the string "unused" is too short, so if it somehow did make it into a real password field it would get rejected.

securedrop/models.py
sssoleileraaa: It makes sense to me to go ahead and use a valid UUID. Downstream, the client-side User database model has this field:
```py
class User(Base):
    ...
    def deleted(self) -> bool:
        return True if self.uuid == "deleted" else False
```

Which can be updated and released, before your changes, to:
```py
class User(Base):
    ...
    def deleted(self) -> bool:
        return True if self.username == "deleted" else False
```

The reason we didn't call this something like `is_special_deleted_journalist` was because we imagined potentially adding support to distinguish replies sent by distinct deleted journalists. But the non-null and unique username presents a challenge for the case where an org wants to distinguish replies sent by distinct deleted journalists without having to freeze usernames or create unique username replacements, e.g. `deleted_0` `deleted_1` `deleted_2` ... and so on. For now, just supporting one deleted user account that everything gets rolled up into is a big improvement.

sssoleileraaa: You can test against https://github.com/freedomofpress/securedrop-client/tree/users-endpoint if it helps.

legoktm: Done.

legoktm:
In testing the client PR, I realized I overlooked that all the `seen_*` tables have a `UNIQUE(fk_id, journalist_id)`, which means if two journalists have seen the same thing and then both are deleted, the second update would fail with a unique key violation since the "deleted" journalist has already seen that thing. The delete() method now checks to see if deleted has already seen that item in which case it deletes the duplicate row. The migration script tries to update as many rows as possible, ignoring any key conflicts, and then does a second pass to delete everything remaining.

securedrop/models.py
```diff
@@ -784,13 +781,89 @@ def to_json(self, all_info: bool = True) -> Dict[str, Any]:
...
+        # For the rest of the associated data we rely on cascading deletions
+        db.session.delete(self)
+
+
+class DeletedJournalist(Journalist):
+    """A system user that represents deleted journalists for referential integrity"""
+    def __init__(self) -> None:
+        # password field is unused (we override set_password below), plus
+        # it's too short to be valid anyway.
+        super().__init__(username='deleted', password='unused')
```
zenmonkeykstop: A valid OTP secret is generated for the `deleted` user - attempting to log in with said credentials fails with the following raceback:
```
ValueError

ValueError: Should never happen: pw_salt is none for legacy Journalist 4
Traceback (most recent call last)

    File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 2309, in __call__

    return self.wsgi_app(environ, start_response)

    File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 2295, in wsgi_app

    response = self.handle_exception(e)

    File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 1741, in handle_exception

    reraise(exc_type, exc_value, tb)

    File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/_compat.py", line 35, in reraise

    raise value

    File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 2292, in wsgi_app

    File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 1815, in full_dispatch_request

    File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 1718, in handle_user_exception

    reraise(exc_type, exc_value, tb)

    File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/_compat.py", line 35, in reraise

    raise value

    File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 1813, in full_dispatch_request

    rv = self.dispatch_request()

    File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 1799, in dispatch_request

    return self.view_functions[rule.endpoint](**req.view_args)

    File "/home/user/securedrop/securedrop/journalist_app/main.py", line 30, in login

    user = validate_user(request.form['username'],

    File "/home/user/securedrop/securedrop/journalist_app/utils.py", line 97, in validate_user

    return Journalist.login(username, password, token)

    File "/home/user/securedrop/securedrop/models.py", line 730, in login

    if not user.valid_password(password):

    File "/home/user/securedrop/securedrop/models.py", line 571, in valid_password

    raise ValueError(

    ValueError: Should never happen: pw_salt is none for legacy Journalist 4
```
Login should fail, but not this explosively! There was existing code to bail out of logins for invalid usernames, of which `deleted` is currently the only one, but it's not getting hit because it assumes that the invalid username and UUID are the same. 

I'm kindof coming around to the idea of a "login_enabled" field or something in the journalist table, which would be checked first when validating user logins. This will probably have to be added anyway in the future to support account locking, but it could simplify this PR now, and allow for the removal of that INVALID_USERNAMES array in favour of a db query.

zenmonkeykstop: Also, setting/forgetting a valid diceware password for the deleted user, and dropping the overriden set_password method, would both rule out the exception above and clear the new LGTM alert. I recognise that it's an intentional, not accidental, change, but the side effects of overriding the method seem undefined, which is a worry.

legoktm: > Login should fail, but not this explosively! There was existing code to bail out of logins for invalid usernames, of which deleted is currently the only one, but it's not getting hit because it assumes that the invalid username and UUID are the same.

Uh, I missed that UUID check somehow. Fixing that and writing a test for this as well.

It is semi-intentional on my part that it does explode, the idea was that it should be impossible to log into "deleted" on multiple levels, first being the INVALID_USERNAMES check (or login_enabled in the future), and if you somehow bypassed it (as happened in this case), you'd still get stuck because there's no salt/password/TOTP. If we are fine with just having a random diceware+TOTP values to provide that second layer, I can try to switch to that. It is likely that will add additional complexity to the database migration script since we need to go through all that logic without being able to reuse the `Journalist` class, but it seems doable. We would still need the override for check_username_acceptable, but the consequences of that seem much more predictable (compared to your concern with the override of set_password).

I think we still want to keep INVALID_USERNAMES even with a potential login_enabled because we want to prevent people from manually creating "deleted" account. So I can work on adding in login_enabled but I don't think it'll help that much here compare to the increased scope of another schema change.

zenmonkeykstop: I'm just leery of something like that throwing an undhandled exception, but the point about migration woes is well-taken. Maybe we don't even need the `login_enabled` field just yet: if you had a `Journalist.is_unlocked()` that just returned true you could override that in `DeletedJournalist` to return false without falling foul of constructor overrides. If we ever did add account locking (and said field) it would still be useful, the `Journalist` version of the method would probably just query the db.

legoktm: Creating a PassphraseGenerator in the migration step seems to work with minimal copy and paste, so I pushed changes that now set a real passphrase/totp secret for "deleted", and no longer override "set_password". I also fixed the pre-existing tests for the deleted journalist login to work with the new style of having a real UUID.

> if you had a `Journalist.can_log_in()` that just returned true you could override that in `DeletedJournalist` to return false without falling foul of constructor overrides

This wouldn't work because on the login form when we query by username, we get an instance of `Journalist` back, not `DeletedJournalist`. Now `DeletedJournalist` is only used to bypass the username validation in check_username_acceptable.

Now that we no longer need it to override password handling, we could just get rid of the class and do something like:
```
deleted = Journalist(username="placeholder", password=...)
deleted.username = "deleted"
```
to bypass the constructor validation. And `DeletedJournalist.get()` would become `Journalist.get_deleted()`. What do you think?

zenmonkeykstop: SGTM - will hold off on further review for now, lemme know when it's updated.

legoktm: @zenmonkeykstop done!

legoktm:
(Rebased to fix the merge conflict after the Flask 2.0 PR landed)

zenmonkeykstop (approved):
Ran through test plan, LGTM!
Also:
- spun up a prod VM install in 2.1.0
- added a journalist account, viewed submissions with it and deleted the account
- confirmed nulls in seen_* tables
- upgraded to packages built off this branch
- confirmed the creation of a `deleted` user with a valid UUID, and that seen_* table entries with nulls now reference this user's id
- confirmed the API reply behaviour.

PR #6231, opened 2022-01-19
Don't register trigger on Python for securedrop-app-code package

### LINKED ISSUE #6230: postinst error in `securedrop-app-code` package install
Seen on my test instance and one other:

```
Processing triggers for securedrop-app-code (2.1.0+focal) ...
+ set -o pipefail
+ SDVE=/opt/venvs/securedrop-app-code
+ SDBIN=/opt/venvs/securedrop-app-code/bin
+ case "$1" in
+ echo 'postinst called with unknown argument `triggered'\'''
postinst called with unknown argument `triggered'
+ exit 1
 installed securedrop-app-code package post-installation script subprocess returned error exit status 1
```

I see this for upgrades going back to SecureDrop 2.0.0, so it does not appear to interfere with successful upgrades.

legoktm:
## Status

Ready for review, but untested

## Description of Changes

Don't register trigger on Python for securedrop-app-code package
    
We currently register a trigger with the Python
interpreter so dh_virtualenv can fix up the interpreter's symlinks
in case something changes as recommended in their docs[1].
    
However, our packages are designed for a single Ubuntu release, which
keeps the same version of Python for its entire lifetime, so the
interpreter in the venv will always be a symlink to `/usr/bin/python3`
and the dh_virtualenv postinst code will always be a no-op because it
skips all symlinks.
    
This was noticed because our custom postinst was not handling the
"triggered" state, which is valid per deb-postinst(5). Add that in for
future proofing even though we don't expect it to be called

[1] https://dh-virtualenv.readthedocs.io/en/latest/tutorial.html?highlight=trigger#step-2-set-up-packaging-for-your-project

Fixes #6230.

## Testing

* Install an older version of Python 3 (grab an old Ubuntu container maybe?)
* Install old or currently released securedrop-app-code package
* Build new securedrop-app-code package with this patch
* Install new securedrop-app-code package
* Upgrade Python 3 package
* Observe no error message related to triggers being unrecognized.
* Observe no triggers run (no "Processing triggers for securedrop-app-code...")

Alternatively I think you can directly run the trigger with something like `dpkg-trigger --by-package python3.8 /usr/bin/python3.8`

## Deployment

This will remove the trigger behavior entirely, which should be a no-op because it was broken until now.

## Checklist

### DIFF
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/defaults/main.yml
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/defaults/main.yml
@@ -46,7 +46,3 @@ securedrop_venv: "/opt/venvs/securedrop-app-code"
 securedrop_venv_bin: "{{ securedrop_venv }}/bin"
 securedrop_python_version: "{{ '3.8' if securedrop_target_distribution == 'focal' else '3.5' }}"
 securedrop_venv_site_packages: "{{ securedrop_venv }}/lib/python{{ securedrop_python_version }}/site-packages"
-
-securedrop_app_focal_files:
-  - src: securedrop-app-code.triggers-focal
-    dest: "{{ securedrop_app_code_prep_dir }}/debian/securedrop-app-code.triggers"
--- a/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/securedrop-app-code.triggers-focal
+++ b/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/securedrop-app-code.triggers-focal
@@ -1,7 +1,0 @@
-# Register interest in Python interpreter changes; and
-# don't make the Python package dependent on the virtualenv package
-# processing (noawait)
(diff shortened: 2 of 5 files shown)


conorsch (approved):
Looks good. Thanks for the detailed comments discussing the approach!

FYI, I grepped through the repo and found some lingering references to the trigger file in the build logic, and snipped those out, squashing into your commit. The per-PR CI didn't complain, but the full nightly CI runs would have. Let's keep an eye on the nightly builds to make sure we didn't miss anything.

legoktm:
@conorsch Thanks! Looks like the nightly CI passed fine. Is there any reason we don't run the package building CI parts when those files are modified in a PR?

conorsch:
> Is there any reason we don't run the package building CI parts when those files are modified in a PR?

Last I checked, CircleCI doesn't support triggering jobs based on subpaths within a repo. We could rig up our own bash script to compare branches with `--name-status` or the like... but we've never done that. We shunted the full-install job to nightly back in... https://github.com/freedomofpress/securedrop/issues/5336

Issue #5972, opened 2021-06-04
Review and prioritize 2019 accessibility report findings


eloquence:
In 2019, with support from Internews, Accessibility Lab completed a review of the SecureDrop Journalist Interface and the SecureDrop website. While we reviewed the findings of the report, at the time, we did not have bandwidth to implement significant changes. Thanks to continued support from Internews, @cfm is working with us, and will review these findings and develop a prioritized set of improvements. This issue will serve as an epic to organize this work as it pertains to the Journalist Interface.

## Tasks [prioritized](https://docs.google.com/spreadsheets/d/13lbFPR1Pm4QbcVz_7oVxMmKViqSl268CKA0ixoMyYxw/edit#gid=57993727) so far
- [x] #5983 (WCAG 1.1.1)
- [x] #5984 (WCAG 1.1.1)
- [x] #5985 (WCAG 1.1.1)
- [x] semantic (HTML5/ARIA) page structure (WCAG 1.3.1) that passes validation (WCAG 4.1.1)
    - [x] #5986
    - [x] #5987
 - [x] ping Accessibility Lab with an update after #5986 and #5987
 - [x] add "skip to main content" link to all views (WCAG 2.4.1)
     - [x] #6237
 - [x] #6238
 - [x] add non-color differentiation to `:hover` and `:active` styles (WCAG 1.4.1, 1.4.11, 2.4.7; but see [caveat][caveat])
     - https://github.com/freedomofpress/securedrop/issues/5972#issuecomment-1024771193
 - [x] check current color scheme meets contrast requirements (WCAG 1.4.3, 1.4.11; but see [caveat][caveat])
     - https://github.com/freedomofpress/securedrop/issues/5972#issuecomment-1024771193
 - [ ] #6241 (blocked)
 - [x] #6313
 - [x] #6404
 - [ ] #6405

[caveat]: https://github.com/freedomofpress/securedrop/issues/5972#issuecomment-1006234249

## Related/open questions
- [x] #3268
- Guidelines and documentation: cf. #5743  → #6314
- Tests for a11y compliance to prevent regressions in new features, etc.  →  #6314

cfm:
[...]

PR #6237, opened 2022-01-26
adds "skip to main content" link to all pages that inherit from base templates

cfm:
## Status

Ready for review

## Description of Changes

Towards #5972, adds a "skip to main content link" to all pages that inherit from base templates.

## Testing

1. Log into the Source Interface.
2. Observe that there is no "Skip to main content link" visible.
3. Tab through the page, with or without a screen-reader running.
4. Observe that the "Skip to main content link" appears when it receives focus.
5. Click the link and continue tabbing through the `main` content.
6. Repeat steps (1)–(5) on the Journalist Interface.

## Deployment

No deployment considerations.

### DIFF
--- a/securedrop/journalist_templates/base.html
+++ b/securedrop/journalist_templates/base.html
@@ -20,6 +20,7 @@
 
   {% if g.user %}
   <nav aria-label="{{ gettext('Navigation') }}">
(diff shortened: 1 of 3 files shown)


legoktm (approved):
LGTM, verified the test plan on both SI and JI by tabbing through it (not a screen-reader).

PR #6244, opened 2022-02-02
Properly register "pagelayout" marker with pytest

legoktm:
## Status

Ready for review

## Description of Changes

This gets rid of warnings like:

```
tests/pageslayout/test_journalist.py:25
.../securedrop/securedrop/tests/pageslayout/test_journalist.py:25:
PytestUnknownMarkWarning: Unknown pytest.mark.pagelayout - is this a
typo?  You can register custom marks to avoid this warning - for
details, see https://docs.pytest.org/en/stable/mark.html
    @pytest.mark.pagelayout
```

Our `--page-layout` parameter setup now matches the example from
<https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option>.

While we're at it, run pytest with the `--strict-markers` flag so if a new
marker is added without properly registering it, it'll fail right away.

## Testing

- [ ] Run tests and see that the PytestUnknownMarkWarnings are gone
- [ ] Run `./securedrop/bin/dev-shell bin/run-test --markers` and see "@pytest.mark.pagelayout: ..." in the output
- [ ] Run `./securedrop/bin/dev-shell bin/run-test -m pagelayout` and see it only runs tests marked as pagelayout
- [ ] Comment out the newly added pytest_configure function and observe that pytest fails when collecting tests because we use `--strict-markers` and didn't register it

## Deployment

## Checklist

### DIFF
--- a/securedrop/bin/run-test
+++ b/securedrop/bin/run-test
@@ -65,4 +65,5 @@ pytest \
     --cov-report "xml:${TEST_RESULTS}/cov.xml" \
     --cov-report "annotate:${TEST_RESULTS}/cov_annotate" \
     --cov=. \
+    --strict-markers \
     "$@"
--- a/securedrop/tests/conftest.py
+++ b/securedrop/tests/conftest.py
@@ -66,6 +66,10 @@ def pytest_addoption(parser):
                      default=False, help="run page layout tests")
 
 
+def pytest_configure(config):
+    config.addinivalue_line("markers", "pagelayout: Tests which verify page layout")
+
+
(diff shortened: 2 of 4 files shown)


zenmonkeykstop (changes requested):
Fix looks good, clears the listed warning. 

While testing I noticed that `securedrop/tests/pageslayout/test_source_layout_torbrowser.py` is missing the marker,wanna just update this PR to fix that while we're here?

zenmonkeykstop (approved):
Test plan looks good, all page layout tests have marker now, LGTM!

PR #6245, opened 2022-02-03
Updates for flask and jinja deprecations

legoktm:
## Status

Ready for review

## Description of Changes

* Update deprecated/renamed parameters in `flask.send_file()`  
   * `attachment_filename` is now `download_name`
   * `add_etags` is now `etag`
   * See <https://flask.palletsprojects.com/en/2.0.x/api/#flask.send_file>.
* Update for jinja deprecations
   * `Markup` and `escape` must be imported from markupsafe directly, they
      were already internally imported from there by jinja.
   * `evalcontextfilter` was renamed to `pass_eval_context`. Apply it
      directly on the filter itself as a decorator for ease of use and
      to match the style in the jinja documentation (<https://jinja.palletsprojects.com/en/3.0.x/api/#custom-filters>).

## Testing

These changes are pretty trivial, it's easy to trace in the jinja/flask code that these are simple renames and shouldn't affect functionality.

* From the journalist UI, send a message to a source with newlines and HTML tags. View the message in the source UI and see that the newlines are converted into `<br>` and the HTML is escaped properly.
* Download multiple files from the journalist UI so it gets turned into a zip file and verify the filename on the zip is as expected, like `{source_name}--{timestamp}.zip`.

## Deployment

Not really.

## Checklist

### DIFF
--- a/securedrop/journalist_app/utils.py
+++ b/securedrop/journalist_app/utils.py
@@ -241,7 +241,7 @@ def download(
     return send_file(
         zf.name,
         mimetype="application/zip",
-        attachment_filename=attachment_filename,
+        download_name=attachment_filename,
         as_attachment=True,
     )
 
@@ -504,7 +504,7 @@ def serve_file_with_etag(db_obj: Union[Reply, Submission]) -> flask.Response:
     response = send_file(file_path,
                          mimetype="application/pgp-encrypted",
                          as_attachment=True,
-                         add_etags=False)  # Disable Flask default ETag
+                         etag=False)  # Disable Flask default ETag
 
     if not db_obj.checksum:
         add_checksum_for_file(db.session, db_obj, file_path)
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -15,7 +15,7 @@ from cryptography.hazmat.primitives.kdf import scrypt
 from flask import current_app, url_for
 from flask_babel import gettext, ngettext
 from itsdangerous import TimedJSONWebSignatureSerializer, BadData
-from jinja2 import Markup
+from markupsafe import Markup
 from passlib.hash import argon2
 from sqlalchemy import ForeignKey
 from sqlalchemy.orm import relationship, backref, Query, RelationshipProperty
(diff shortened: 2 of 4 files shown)


legoktm:
We might want to cherry-pick 75b737c4f06ab4a8a3f45585230ba9c8c8fb5b64 into a separate PR since it seems like some part of CI is running with the wrong versions of jinja2/markupsafe.

gonzalo-bulnes (commented):
- [x] The requirements in *develop* and *test* are the same.

- [ ] I would have expected `securedrop-app-code-requirements.in` to specify the same constraints: `Jinja>=3.0`, `markupsafe>=2.0`. That being said, I am not familiar with this code and I may be missing something.
      Currently they are: `Jinja2>= 3.0.0`, `markupsafe>=1.1`.

- [x] The code changes match the description

legoktm:
Thanks for the review!

>  * [ ]  I would have expected `securedrop-app-code-requirements.in` to specify the same constraints: `Jinja>=3.0`, `markupsafe>=2.0`. That being said, I am not familiar with this code and I may be missing something.
>       Currently they are: `Jinja2>= 3.0.0`, `markupsafe>=1.1`.

3.0 and 3.0.0 are semantically the same but I can switch to the latter for consistency. The `markupsafe>=1.1` is misleading because jinja 3.0 requires markupsafe>=2.0 so it gets pulled in correctly anyways. We can bump it to >=2.0 in the app-code requirements to make it more obvious what's actually being installed (it'll be a no-op).

conorsch:
> The markupsafe>=1.1 is misleading because jinja 3.0 requires markupsafe>=2.0 so it gets pulled in correctly anyways. We can bump it to >=2.0 in the app-code requirements to make it more obvious what's actually being installed (it'll be a no-op).

Strong agree. FWIW, we recently tried to capture our expectations on version constraints here: https://github.com/freedomofpress/securedrop/wiki/Dependency-specification-and-update-policies#specifying-version-constraints and what you're describing is perfectly aligned.

legoktm:
The fixes to requirements were merged in #6247, so this is now just the actual deprecation changes.

eaon (approved):
LGTM, works as expected, thanks!

PR #6247, opened 2022-02-03
Fix version constraints in requirements and dependencies used for tests

legoktm:
## Status

Ready for review

## Description of Changes

* Fix markupsafe version constraint be >=2.0 in app-code-requirements (no-op)
  Jinja 3.0 already requires markupsafe>=2.0 and in the pinned
  requirements.txt file we use markupsafe 2.0.1. Adjust the requirement in
  the requirements.in file to be >=2.0 to be less confusing and reflect
  what we're actually installing.
* Bump jinja2+markupsafe in {develop,test}-requirements to match app-code
  So tests/CI are properly using the same versions of dependencies
  as we do in production.

This was split out of #6245.

## Testing

* Verify that hashes and versions match those already in app-code-requirements.txt
* Verify CI passes

## Deployment

No special considerations

## Checklist

- [x] Diff review for these was already done


zenmonkeykstop (approved):
CI is passing, same versions confirmed in develop reqs, and the slimming down of test reqs is a good fix. LGTM!

PR #6251, opened 2022-02-08
Expand the 2.2.0 changelog

legoktm:
## Status

Ready for review

## Description of Changes

Expand the 2.2.0 changelog by reviewing <https://github.com/freedomofpress/securedrop/compare/c6a9636d0c00ffdc78fd711948e18997b81b5a5e...develop> and manual curation.

I/we intend to use the same changelog in the release announcements.

## Testing

Just read it. Please suggest alternative wording or if things should be in different sections...

## Deployment

n/a

## Checklist

n/a

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -3,21 +3,47 @@
 ## 2.2.0~rc1
 
 ### Web applications
-* Improved accessibility of Journalist Interface via semantic markup (#6165)
-* Improve handling of HOTP tokens setup (#5599)
-* Disabled caching of passphrases in gpg-agent (#
-* Added additional defensive HTTP headers to Apache config (#6187)
-* Updated dependencies: click 6.7 to 8.0.3; flask 1.0.2 to 2.0.2; flask-wtf 0.14.2 to 1.0.0; jinja2 2.11.3 to 3.0.2; markupsafe 1.1.1 to 2.0.1; redis 3.3.6 to 3.5.3; rq 1.1.0 to 1.10.0; werkzeug 0.16.0 to 2.0.2 (#6190, #6217)
+
+* Refactor and simplify code for creating a new source (#6087).
+* Timestamps will now be displayed using the language's "long" format (#6144).
+* The "Forgot your codename?" hint will only be shown on the initial login (#6130).
+* Disable caching of GPG passphrases to ensure the correct passphrase must be passed every time (#6174).
+* Set minimum length requirements on HOTP and TOTP secrets (#6191).
+* Improve security-related/defensive HTTP headers used in the source and journalist interfaces (#6187).
+* Remove "Refresh codename" feature because of potential user confusion (#6195).
+* Refactor code related to encryption (#6160).
+* Stop explicitly flushing the database session (#6223).
+* Upgrade to Flask 2.0 and associated refactoring (#6217).
+* Improve accessibility of journalist interface by using semantic HTML5/ARIA markup (#6165).
+* Information corresponding to deleted journalists will now be associated with a reserved "deleted" account internally. (#6225).
 
 ### Operations
 
-* Migrated kernel version from 5.4.136 to 5.15.18, for newer hardware support (#6242)
-* Added support for custom hostnames to `./securedrop-admin verify` (#6154)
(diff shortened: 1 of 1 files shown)


changelog.md
```diff
@@ -3,21 +3,49 @@
...
+  * wtforms from 2.1.0 to 3.0.0 (#6190)
+* Install dh-virtualenv from Ubuntu 21.10/Impish (#6206).
+* Upgrade builder to use Rust 1.58.1 (#6234).
+* Don't register trigger on Python for securedrop-app-code package (#6231).
 
 ### Development
-* Added support for onion services to dev container (#6152)
-* Refactored cryptography functions to add type-checking throughout (#5599)
+
+* Refactoring mentioned above has enabled type-checking against significantly more code.
```
zenmonkeykstop: This seems duplicative and doesn't have a distinct issue/PR associated with it, can probably drop.

changelog.md
```diff
@@ -3,21 +3,49 @@
...
+* Improve robustness of Tor Browser version check in `make dev` (#6166).
+* Remove unsupported Xenial dev Dockerfiles, templates and more. (#6166, #6185).
+* Support running the dev setup using a public onion service v3 address with `make dev-tor` (#6167).
```
zenmonkeykstop: make dev-tor sets up both services, not just the public one, could probably be amended to "Support running the dev setup using onion services with `make dev-tor`"

changelog.md
```diff
@@ -3,21 +3,49 @@
...
+* Upgrade to Flask 2.0 and associated refactoring (#6217).
+* Improve accessibility of journalist interface by using semantic HTML5/ARIA markup (#6165).
+* Information corresponding to deleted journalists will now be associated with a reserved "deleted" account internally. (#6225).
 
 ### Operations
 
-* Migrated kernel version from 5.4.136 to 5.15.18, for newer hardware support (#6242)
-* Added support for custom hostnames to `./securedrop-admin verify` (#6154)
-* Removed references to v2 onion config (#6169)
+* Update Tor to 0.4.6.9 (#6227).
```
zenmonkeykstop: I'm kindof on the fence about including Tor updates in the SD changelog anymore, IMO we should be decoupling them from the release cycle and striving just to stay on latest stable. We updated Tor a week ago so it's not really a 2.2.0 thing.

zenmonkeykstop (changes requested):
Thanks for this @legoktm, only a few small tweaks required IMO

changelog.md
legoktm: Makes sense, removed.

changelog.md
legoktm: By public I meant it's a real live onion service that anyone with Tor access could hit, not just some localhost thing, but I can see how that could be confusing so your wording is better. Fixed.

changelog.md
legoktm: There have been multiple changes this cycle that were valuable in their own right, but as a side-effect, helped with type checking, like the crypto refactoring, Flask 2.0, etc. I wanted to capture that somehow, but probably that's better for release announcement prose rather than in the changelog. Removed.

conorsch (approved):
Beautiful, @legoktm, thanks for taking the time. I caught one small typo and amended to resolve.

<details>
<summary>view diff</summary>

```diff
diff --git a/changelog.md b/changelog.md
index d4af03704..ac67d8d66 100644
--- a/changelog.md
+++ b/changelog.md
@@ -8,7 +8,7 @@
 * Timestamps will now be displayed using the language's "long" format (#6144).
 * The "Forgot your codename?" hint will only be shown on the initial login (#6130).
 * Disable caching of GPG passphrases to ensure the correct passphrase must be passed every time (#6174).
-* Set minimum length requirements on HOTP and TOTP secrets (##6191).
+* Set minimum length requirements on HOTP and TOTP secrets (#6191).
 * Improve security-related/defensive HTTP headers used in the source and journalist interfaces (#6187).
 * Remove "Refresh codename" feature because of potential user confusion (#6195).
 * Refactor code related to encryption (#6160).
```
</details>

zenmonkeykstop (approved):

PR #6256, opened 2022-02-08
Update API to return datetimes without offset field.

### LINKED ISSUE #6255: Journalist API datetime format has changed, breaking client login.
## Description

See freedomofpress/securedrop-sdk#171 - the datetime format being returned by the Journalist API includes timezone info as a result of the Flask 2.0 update - previously it dd not, and the SecureDrop Client expects naive datetimes with no timezones.

## Steps to Reproduce

Attempt to use the client with a server running 2.2.0-rc1 or latest develop

## Expected Behavior

Client works as normal

## Actual Behavior

Client crashes with error in linked issue above

## Comments

A simple fix would be to force the datetime format to not include timezones in the API. There's a larger issue here of how to handle tz-aware datetimes in the client, but that might be too much to take on during the client process, it seems better to preserve the existing behaviour.

Looks like there are two potential areas to consider: the API token expiry datetime (which is probably causing the issue above), and the source last_updated field (which is used in the client). The first is a change in securedrop/journalist_app/api.py, the second is a change in models.py, in Source.to_json(). I don't think there are any other affected areas.

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #6255 .

Sets datetime values returned by the API to the format `%Y-%m-%dT%H:%M:%S.%fZ`, excluding the timezone offset information being added after the Flask 2.0 update

## Testing

- Set up a staging or prod vm instance based off this branch, with source and journalist accounts, and configure a Securedrop Workstation install to use it:
  - [ ] Confirm that you can log in to the securedrop client
  - [ ] Confirm that you can view source conversations and reply to sources
- Add a new source
  - [ ] confirm that the new source is displayed correctly in the client's source list after a sync
- Upload new files as an existing source
  - [ ] Confirm that the source is displayed correctly in the client's source list after next sync.

Choose one of the following:

### DIFF
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -48,6 +48,9 @@ HOTP_SECRET_LENGTH = 40  # 160 bits == 40 hex digits (== 32 ascii-encoded chars 
 # but existing Journalist users may still have 80-bit (16-char) secrets
 OTP_SECRET_MIN_ASCII_LENGTH = 16  # 80 bits == 40 hex digits (== 16 ascii-encoded chars in db)
 
+# Timezone-naive datetime format expected by SecureDrop Client
+API_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
+
 
 def get_one_or_else(query: Query,
                     logger: 'Logger',
@@ -138,9 +141,9 @@ class Source(db.Model):
         docs_msg_count = self.documents_messages_count()
 
         if self.last_updated:
-            last_updated = self.last_updated.isoformat() + 'Z'
+            last_updated = self.last_updated.strftime(API_DATETIME_FORMAT)
         else:
-            last_updated = datetime.datetime.utcnow().isoformat() + 'Z'
+            last_updated = datetime.datetime.utcnow().strftime(API_DATETIME_FORMAT)
 
         if self.star and self.star.starred:
             starred = True
@@ -777,7 +780,7 @@ class Journalist(db.Model):
         if all_info is True:
             json_user['is_admin'] = self.is_admin
             if self.last_access:
-                json_user['last_login'] = self.last_access.isoformat() + 'Z'
+                json_user['last_login'] = self.last_access.strftime(API_DATETIME_FORMAT)
             else:
                 json_user['last_login'] = None
 
--- a/securedrop/tests/test_journalist_api.py
+++ b/securedrop/tests/test_journalist_api.py
@@ -2,6 +2,7 @@
 import json
 import random
 
+from datetime import datetime
 from flask import url_for
 from itsdangerous import TimedJSONWebSignatureSerializer
 from pyotp import TOTP
@@ -19,6 +20,8 @@ from models import (
 )
 
 from .utils.api_helper import get_api_headers
+
+API_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
 
 random.seed('◔ ⌣ ◔')
 
@@ -52,6 +55,15 @@ def test_valid_user_can_get_an_api_token(journalist_app, test_journo):
         assert response.status_code == 200
         assert response.json['journalist_first_name'] == test_journo['first_name']
         assert response.json['journalist_last_name'] == test_journo['last_name']
+
+        def _valid_date(date_str, date_format):
+            try:
+                if date_str != datetime.strptime(date_str, date_format).strftime(date_format):
+                    raise ValueError
+                return True
+            except ValueError:
+                return False
+        assert _valid_date(response.json['expiration'], API_DATETIME_FORMAT)
 
 
 def test_user_cannot_get_an_api_token_with_wrong_password(journalist_app,
--- a/securedrop/journalist_app/api.py
+++ b/securedrop/journalist_app/api.py
@@ -20,10 +20,9 @@ from journalist_app import utils
 from models import (Journalist, Reply, SeenReply, Source, Submission,
                     LoginThrottledException, InvalidUsernameException,
                     BadTokenException, InvalidOTPSecretException,
-                    WrongPasswordException)
+                    WrongPasswordException, API_DATETIME_FORMAT)
 from sdconfig import SDConfig
 from store import NotEncrypted, Storage
-
 
 TOKEN_EXPIRATION_MINS = 60 * 8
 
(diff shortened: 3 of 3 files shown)


conorsch (approved):
I tested these changes in a dev container on this branch, and could no longer reproduce the error reported in https://github.com/freedomofpress/securedrop-client/issues/1418. Was able to log in, download, and send replies as a journalist. As for the other testing scenarios, I suggest we get to those in the course of ongoing QA.

conorsch:
CI failed on page layout tests, specifically for HOTP secrets: https://app.circleci.com/pipelines/github/freedomofpress/securedrop/3596/workflows/c629b899-e1e8-4edb-a256-7e287563beec/jobs/59330 I'm re-running CI to confirm it's a valid error; if so, we may need to tweak the test accordingly.

eloquence:
Test may have been a flake, CI is green now.

sssoleileraaa (approved):
I confirmed that this PR unbreaks the client (see where it was breaking here: https://github.com/freedomofpress/securedrop-sdk/blob/9b70550f1dc43a052f5025265473281e05c738a1/sdclientapi/__init__.py#L251) and visual inspection of the code lgtm. I just want to give a heads-up to @gonzalo-bulnes, who is looking into updating the SDK, _before_ the server changes its API, to support a UTC+timezone format for token expiration.

gonzalo-bulnes:
I have reasons to think the API _should_ include time-zone information in its responses (even if we only ever outputs times in their UTC representation, which I'd suggest we do). I've detailed why in https://github.com/freedomofpress/securedrop/issues/6257 but the summary is: standard, unambiguous and convenient formats are available even within our constraints :slightly_smiling_face:

securedrop/tests/test_journalist_api.py
```diff
@@ -20,6 +21,8 @@
 
 from .utils.api_helper import get_api_headers
 
+API_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
```
legoktm, line 24: Could you import this from models instead of duplicating?

securedrop/tests/test_journalist_api.py
```diff
@@ -53,6 +56,15 @@ def test_valid_user_can_get_an_api_token(journalist_app, test_journo):
         assert response.json['journalist_first_name'] == test_journo['first_name']
         assert response.json['journalist_last_name'] == test_journo['last_name']
 
+        def _valid_date(date_str, date_format):
```
legoktm, line 59: Maybe I'm missing something, is there any reason this couldn't just be:
```python
def _valid_date(date_str, date_format):
    assert date_str == datetime.strptime(date_str, date_format).strftime(date_format)
```

If the assertion fails, pytest will show us the values of the left and right sides instead of something generic like "asserting False is True". And we let the ValueError bubble up since it'll fail the test anyways.

securedrop/models.py
```diff
@@ -138,9 +141,9 @@ def to_json(self) -> 'Dict[str, object]':
         docs_msg_count = self.documents_messages_count()
 
         if self.last_updated:
-            last_updated = self.last_updated.isoformat() + 'Z'
+            last_updated = self.last_updated.strftime(API_DATETIME_FORMAT)
         else:
-            last_updated = datetime.datetime.utcnow().isoformat() + 'Z'
+            last_updated = datetime.datetime.utcnow().strftime(API_DATETIME_FORMAT)
```
legoktm, line 146: For the future: While it doesn't matter since we discard the TZ information, we probably should switch this to `datetime.now(timezone.utc)` for consistency.

legoktm (commented):
A bit late review. For the future we should consider implementing a custom JSON encoder as described in <https://stackoverflow.com/questions/12122007/python-json-encoder-to-support-datetime> so we don't need to update every timestamp in the code.

securedrop/models.py
gonzalo-bulnes: Wait, I agree with you @legoktm, but we don't strip the TZ information, we do encode `Z` which stands for `+00:00` (UTC). So it does matter.
If the datetime was generated in the local time zone, this would format it incorrectly (shifting as many hours as the TZ offset from UTC).

```py
from datetime import datetime, timezone

# It's 18:10 in Australia.
correct = datetime.now(timezone.utc)
wrong = datetime.now()

API_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
correct.strftime(API_DATETIME_FORMAT)
# '2022-02-09T07:10:59.478694Z'

wrong.strftime(API_DATETIME_FORMAT)
# '2022-02-09T18:10:59.479154Z'
#                            ^-- this is wrong, because Z means UTC, and it's not 18:10 UTC
```

gonzalo-bulnes: It's in part because this is not trivial that I'd avoid `strftime` and favor `isoformat`/`fromisoformat` instead, see #6257

gonzalo-bulnes: To be clear, `correct` and `wrong` are both correct **until** we encode them using the `strftime` format we defined, it's their string representation that's correct or wrong.

legoktm: The current code is correct-ish because it's using `utcnow`:
```python
from datetime import datetime, timezone

# It's 11:45 in California.
correct = datetime.now(timezone.utc)
wrong = datetime.utcnow()

API_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
correct.strftime(API_DATETIME_FORMAT)
# '2022-02-09T07:45:00.795348Z'

wrong.strftime(API_DATETIME_FORMAT)
# '2022-02-09T07:45:26.082728Z'
```

The only distinction between `datetime.now(timezone.utc)` and `datetime.utcnow()` is what the internal tzinfo value is:
```
>>> correct.tzinfo is timezone.utc
True
>>> wrong.tzinfo is None
True
```

...which also affects how its formatted by default but not with our current strftime method.

gonzalo-bulnes: :bulb: Got it. I had missed that completely, thanks for explaining @legoktm !

securedrop/tests/test_journalist_api.py
zenmonkeykstop: I did initially but it makes the test too tautological: "the format I expect is the format the code sends" vs "the format I expect is this." If someone changes the format string in models.py, the test should break.

zenmonkeykstop:
hah - I didn't realize this had already been merged - will defer the test change above until next time we're in the code.

PR #6260, opened 2022-02-09
Use a custom JSON encoder to format datetime objects consistently

legoktm:
## Status

Ready for review

## Description of Changes

We can set `app.json_encoder` to override the default Flask JSONEncoder
used by `jsonify()` when serializing custom types. This will make
changing the timestamp format in the future easier. It also makes it
easier on developers who can now just return a datetime object and
serializing it will just do the right thing.

While we're at it, simplify the test case that verifies the timestamp
format.

This follows up e4f385fb911d8040a7589d9a3a8c96c6ed45a330.

## Testing

- [ ] CI passes
- [ ] Hit an API endpoint that returns a timestamp, like /api/v1/sources (get_all_sources), observe that timestamp format is exactly the same.

## Deployment

None, this should functionally be a no-op

## Checklist

### DIFF
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -4,7 +4,7 @@ from datetime import datetime, timedelta, timezone
 from pathlib import Path
 
 from flask import (Flask, session, redirect, url_for, flash, g, request,
-                   render_template)
+                   render_template, json)
 from flask_assets import Environment
 from flask_babel import gettext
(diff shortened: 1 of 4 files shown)


zenmonkeykstop (approved):
- [x] CI is passing
- [x] timestamps are consistent, ie `"last_updated": "2022-03-23T21:49:03.711120Z"` on this branch vs `develop`

PR #2946, opened 2018-01-27
Css: Delete confirmation with css-only modal

SaptakS:
## Status

Work in progress

## Description of Changes

Fixes #295.

Changes proposed in this pull request:
According to the discussions in #295 and #2506, I have implemented a CSS-only modal for confirmation of deletion of collections or documents. In this PR, I have replaced the js confirmation by the CSS confirmation modal for 
* multiple collection deletions
* single collection deletion
* multiple documents of a single collection deletion

## Testing

How should the reviewer test this PR?
Send a message as a whistleblower, log as a journalist mark the message and press `Delete`. A confirmation modal will appear with options to `Cancel` or `Delete`. You can click on `Cancel`, close icon or anywhere outside the modal to close the modal. On clicking `Delete`, the selection collection(s)/document(s) will get deleted.

### If you made changes to the app code:

- [ ] Unit and functional tests pass on the development VM

PR #6165, opened 2021-11-10
refactors Journalist Interface for semantic HTML5/ARIA markup

cfm:
## Status

* Accessibility changes approved by @SaptakS 
* Ready for final maintainer's review
    * See "Checklist" section for pre-merge steps

## Description of Changes

Closes #5987 by:

1. [x] refactoring all `securedrop/journalist_templates/*.html` for semantic markup;
2. [x] adding ARIA annotations where HTML5 semantics are insufficient; and
3. [x] fixing SASS/CSS in light of (1)–(2), especially to move decorative images out of the DOM and accessibility trees.

## Testing

Click through the Journalist Interface and inspect in terms of:

1. ***Markup,* by inspecting and validating HTML.** This need not take place in Tor Browser; I used the [Web Developer extension](https://chrispederick.com/work/web-developer/) in Chrome.
2. ***Accessibility in assistive tech,* by listening to the interface as announced from Tor Browser by a screen-reader.**  It should sound and "feel" conspicuously more fluent.  (I also audited pages in [WebAIM's Wave Evaluation Tool](https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh) and [Deque's axe DevTools](https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd), but these will yield mixed results until the conclusion of this accessibility push.)
3. ***Visual appearance,* by comparing the interface against <https://demo-journalist.securedrop.org>.**  Changes are intended to be "pixel-perfect" except where there were preexisting idiosyncrasies that more-semantic markup tended to make naturally more consistent (or where noted in individual commits).  Of course, suggestions for fine-tuning are welcome!

Steps (2)–(3) should be performed in Tor Browser at both "standard" and "safest" security levels.

Finally:

4. Click through the Source Interface and compare against <https://demo-source.securedrop.org> for any regressions introduced by shared Sass.

## Deployment

No special considerations.

### After approval and before merge

1. [x] Reformat `securedrop/journalist_templates/*.html`
2. [x] Update as many translations as possible per <https://github.com/freedomofpress/securedrop/issues/5972#issuecomment-947317684>; test in Weblate sandbox  →  see <https://github.com/freedomofpress/securedrop/pull/6165#issuecomment-1016857219>
3. Approver will squash-merge ~(especially fix-ups) to one commit per pass~

### If you made changes to the server application code:

[...]

PR #6261, opened 2022-02-09
Remove unused confirm_bulk_delete functionality

legoktm:
## Status

Ready for review

## Description of Changes

This page was originally used as a confirmation interstitial when deleting
files or messages, but in 1e64f3682e (#2946) it was replaced with a CSS-only
deletion confirmation on the same page, so users would never hit the
separate confirmation page.

Remove the tests that were still calling this endpoint, the HTML template
and code that was checking for the "confirm_delete" action.

I noticed this when investigating a translator comment about the "Permanently Delete Files" text and was unable to figure out how to take a screenshot of it. We probably don't need to get this into 2.2.0, but maybe we can mark these strings as not requiring translation in the meantime.

## Testing

* Verify that you still see a confirmation dialog in the JI when deleting files/messages

## Deployment

## Checklist


cfm:
I'll be happy to review this in this sprint, @legoktm.  I developed the same suspicion while working on #6165 but didn't have time to prove that this view and logic could indeed be removed.  :-)

cfm (approved):
Confirmed all deletion flows appear and function as expected:

1. [x] `/`: select one or more sources → **Delete** → **Files and Messages**
2. [x] `/`: select one or more sources → **Delete** → **Source Accounts** →**Yes, Delete Selected Source Accounts**
3. [x] `/col/<filesystem_id>`: select one or more entries → **Delete Selected** → **DELETE**

Nothing like a good all-deletion pull request.  :-)

PR #6264, opened 2022-02-10
Fixed an insufficiently-specific SSL config check

### LINKED ISSUE #6233: Consider using reprepro's update functionality instead of fetch-tor-packages
## Description

We mirror the tor and tor-geoip packages from Tor's apt repo in to the one run by FPF to make sure new Tor versions don't accidentally break SecureDrops. Currently this is done with a custom fetch-tor-packages script, which requires you configure a container/VM with Tor's sources and uses SecureApt to download them, and runs some basic tests.

reprepro, which we use to build our apt repo, has this mirroring functionality built in: https://manpages.debian.org/bullseye/reprepro/reprepro.1.en.html#conf/updates (look at [Wikimedia's config for an example](https://gerrit.wikimedia.org/r/plugins/gitiles/operations/puppet/+/2dfe83dc68dcee6c92bae798d6a59acffbb567c3/modules/aptrepo/files/updates)).

I think it might reduce complexity to use reprepro's functionality instead of having our own script that does it. This also seems like something CI could prepare for apt-test and just have maintainers verify that checksums match and any other manual testing.

### LINKED ISSUE #6263: SSLSessionTickets directive added by postint even if it already exists
## Description

The 2.1.0 release included a postint fixup to convert existing SSL-enabled instances to use TLS1.3. However, the condition for the conversion will still be triggered if the instance has already been ypgraded to 1.3, resulting in a duplicate `SSLSessionTickets off` line being added every time postint is triggered. The duplicate directive doesn't seem to affect functionality and existing instances should not be affected, but the postint condition should be updated to only pass if TLS1.3 is not already set up.

## Steps to Reproduce

- Set up an instance with HTTPS enabled based on 2.1.0 or later. 
- on the app server, run `sudo dpkg-reconfigure securedrop-app-code` repeatedly

## Expected Behavior

The apache SI config should have a single `SSLSessionTickets off` directive

## Actual Behavior

It has N+1 copies of the directive, where N is the number of times you ran `dpkg-reconfigure`

## Comments

Caused by the check for SSL setup not being specific enough  - it's a regex with value ^SSLProtocol all' when it should be more like ^SSLProtocol all -SSLv3 -TLSv1 -TLSv1.1$'

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Fixes #6263.

Updates the TLS1.3 update function in the postint to only update the SI Apache config if TLS 1.2 or lower is enabled, instead of all the time.

## Testing
on this branch:
- `make build-debs`
- set up a staging instance with HTTPS enabled 
- [ ] verify that TLS  1.3 is enabled and other SSL/TLS protocols are not.
- [ ] verify that there is a single copy of the `SSLSessionTickets off` directive
- log in to the app server and run `sudo dpkg-reconfigure securedrop-app-code
- [x] verify that the SSL configuration is unchanged
- manually edit /etc/apache2/source.conf, replacing the block
  ```
  SSLProtocol all -SSLv3 -TLSv1 -TLSv1.1 -TLSv1.2
  SSLHonorCipherOrder off                 
  SSLSessionTickets off
  SSLCompression off
  ```
  with: 
  ```
  SSLProtocol all -SSLv3 -TLSv1 -TLSv1.1
  SSLCipherSuite ECDHE-ECDSA-AES256-GCM-SHA384
  SSLHonorCipherOrder on
  SSLCompression off
  ```
- run `sudo dpkg-reconfigure securedrop-app-code` again
- [x] verify that the TLS1.3 directives have been updated.

## Deployment

- fixes a postint deployment issue

### DIFF
--- a/install_files/securedrop-app-code/debian/postinst
+++ b/install_files/securedrop-app-code/debian/postinst
@@ -108,7 +108,7 @@ remove_bytecode() {
 # Modify existing instance to use only TLS1.3 for the source.
 update_to_tls13(){
     source_conf="/etc/apache2/sites-available/source.conf"
-    if grep -qP '^SSLProtocol all' "$source_conf"; then
+    if grep -qP '^SSLProtocol all -SSLv3 -TLSv1 -TLSv1.1$' "$source_conf"; then
         sed -i '/^SSLProtocol all/c\SSLProtocol all -SSLv3 -TLSv1 -TLSv1.1 -TLSv1.2' "$source_conf"
         sed -i '/^SSLCipherSuite ECDHE-ECDSA-AES256-GCM-SHA384/d' "$source_conf"
         sed -i '/^SSLHonorCipherOrder on/c\SSLHonorCipherOrder off' "$source_conf"


legoktm (approved):
I was unable to figure out how to actually do "set up a staging instance with HTTPS enabled", so I just copied the relevant lines out of the source.conf template and then stepped through the test plan, confirming it works as expected.

PR #6266, opened 2022-02-11
Pin and verify rustup-init sha256 in builder image

legoktm:
## Status

Ready for review

## Description of Changes

As an extra defense, pin the rustup version used and verify the
hash of the downloaded rustup-init binary. Previously we were downloading
the hash from the same place we were downloading the binary, so it didn't
really offer any extra protection besides making sure the download wasn't
corrupted (which HTTPS does for us).

This does not completely protect us, as rustup-init downloads rustup
without verifying signatures, but that will hopefully be fixed
soon: <https://github.com/rust-lang/rustup/issues/2028>.

This shouldn't add a significant amount of maintenance overhead, as old
rustup versions can still be used to download newer Rust versions.

Fixes https://github.com/freedomofpress/securedrop-security/issues/70.

## Testing

- [ ] Build image, see "rustup-init: OK" in output
- [ ] Modify sha256sum in the Dockerfile, try to build, see that it fails

## Deployment

## Checklist

### DIFF
--- a/molecule/builder-focal/Dockerfile
+++ b/molecule/builder-focal/Dockerfile
@@ -42,14 +42,13 @@ COPY dh-virtualenv.pref /etc/apt/preferences.d/
 RUN apt-get update && apt-get install -y dh-virtualenv
 
 ENV RUST_VERSION 1.58.1
+ENV RUSTUP_VERSION 1.24.3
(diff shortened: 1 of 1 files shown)


zenmonkeykstop (approved):
Test plan checks out, 1.58.1 still installed, LGTM.

PR #6267, opened 2022-02-12
Use pyjwt for API tokens instead of itsdangerous's deprecated code

### LINKED ISSUE #6224: Switch from itsdangerous to another JWT library
## Description

An upcoming release of `itsdangerous`, 2.1, will drop support for JSON Web Signatures. The project maintainers recommend switching to `authlib` instead. Encountered the deprecation notice while performing a [diff review](https://github.com/freedomofpress/securedrop-debian-packaging/wiki/itsdangerous-0.24-to-2.0.1). See this [relevant GitHub issue](https://github.com/pallets/itsdangerous/issues/257) with maintainer comments.

legoktm:
## Status

Ready for review

## Description of Changes

itsdangerous deprecated TimedJSONWebSignatureSerializer in 2.0 and has
entirely removed it in 2.1, so we need to switch to something else. They
recommended authlib, except that's a pretty large general-purpose
authentication library.

pyjwt is a smallish (1k LOC) library that just provides JWT/JWS
functionality. For the HS512 algorithm, it uses the same implementation
from hashlib in the Python standard library as itsdangerous did.

Tests are provided to verify successful operation as well as different
failure modes, such as invalid tokens, incorrect secret key, wrong
algorithm, and lack of an expiry.

Tokens generated by itsdangerous are not decodable by pyjwt because
itsdangerous passed the expiry as a header while pyjwt includes the
expiry in the body/payload itself. This shouldn't be a significant issue
since these tokens already expire after 8 hours.

Fixes #6224.

## Testing

- [ ] Get a valid token: `curl -X POST -d '{"username":"journalist", "passphrase":"correct horse battery staple profanity oil chewy","one_time_code":"XXXXXX"}' -H "Content-Type: application/json" http://127.0.0.1:8081/api/v1/token`. Then try it against an API endpoint: `curl -H 'Authorization: Token {token}' http://127.0.0.1:8081/api/v1/users` and get a proper response back.
- [ ] Log in with SD Client, perform various interactions (uses API token for authentication)
- [ ] Log out of SD client, select the previously used token out of the `revoked_tokens` database table and try to use it against the API, e.g. `curl -H 'Authorization: Token {token}' http://127.0.0.1:8081/api/v1/users`. You should get an error message saying the token is invalid or expired.
- [ ] Edit line 122 of journalist_app/api.py so that the expiration is `10` (seconds). Get a valid token using the first curl command, then wait 10 seconds. Try making an API request using that token, and get a response back saying the token is invalid or expired.

## Deployment

Any special considerations for deployment? Yes-ish.

Existing API tokens will no longer be valid and need to be reissued. This shouldn't be a big deal since they only live for 8 hours anyways.

- [x] I would like someone else to do the diff review
  - I already reviewed the code myself, it would be good for someone else to do a review as well.

### DIFF
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -2,11 +2,13 @@
 import binascii
 import datetime
 import base64
+import jwt
 import os
 import pyotp
 import qrcode
 # Using svg because it doesn't require additional dependencies
 import qrcode.image.svg
+import time
 import uuid
 from io import BytesIO
 
@@ -14,7 +16,6 @@ from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.kdf import scrypt
 from flask import current_app, url_for
 from flask_babel import gettext, ngettext
-from itsdangerous import TimedJSONWebSignatureSerializer, BadData
 from jinja2 import Markup
 from passlib.hash import argon2
 from sqlalchemy import ForeignKey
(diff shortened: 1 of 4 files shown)


lsd-cat (approved):
Well structured, clean and no specific behavior differences from before!

legoktm:
We want to move forward with #6403 instead.

PR #6270, opened 2022-02-14
Reset mtime of source private keys to default creation time

legoktm:
## Status

Ready for review

## Description of Changes

We set the GPG key creation time to 2013-05-14 to hide when they
were created, revealing when a source started using SecureDrop. However
this information was being leaked via the file modification time of the
private key material in the $keydir/private-keys-v1.d/ folder.

We set the GPG key creation time to 2013-05-14 to hide when they
were created, revealing when a source started using SecureDrop. However
this information was being leaked via the file modification time of the
private key material in the $keydir/private-keys-v1.d/ folder.

While we can easily change a file's mtime to a past date, faking the
ctime really isn't possible. So instead we touch each private key when
the source app starts to mask the real creation time. Because of the
nightly restarts, this will be updated within 24 hours of source
creation.

Fixes https://github.com/freedomofpress/securedrop-security/issues/71.

## Testing

- [ ] Start dev container, look at mtime+ctime of private keys in `/var/lib/securedrop/keys/private-keys-v1.d`, wait a minute, make a whitespace-only change to `source_app/__init__.py` so the auto reloader restarts the app. Look at mtime+ctime of private keys, they should be different.
- [ ] On staging/prod, install SD, create a source. See mtime+ctime of private keys are the time of creation. Reboot the server, look again and see that mtime+ctime of private keys is now when the server was rebooted.

## Deployment

## Checklist

### DIFF
--- a/securedrop/source_app/__init__.py
+++ b/securedrop/source_app/__init__.py
@@ -1,6 +1,8 @@
 from pathlib import Path
 from typing import Optional
 
+import os
+import time
 import werkzeug
 from flask import (Flask, render_template, escape, flash, Markup, request, g, session,
                    url_for)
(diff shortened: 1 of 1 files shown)


zenmonkeykstop (approved):
Tested in both dev and staging environments, verified mtime and ctimes changes with `stat` - looking good!

PR #6280, opened 2022-02-17
Merge 2.2.0 translations

legoktm:
## Status

Ready for review

## Description of Changes

Merge all the translations made during the 2.2.0 translation cycle from the securedrop-i18n repo.

## Testing

- [ ] View SD in various languages, make sure nothing is broken
- [ ] Review the diff, make sure it looks reasonable, see guidance at https://docs.securedrop.org/en/stable/development/i18n.html#merge-translations-back-to-develop

## Deployment


conorsch (approved):
Ran through the process locally as part of review, and confirmed that even the last-minute translations are included here. Looks great! Just waiting on a local test run for the page-layout tests, will merge after.

PR #6288, opened 2022-02-18
Translate Show/Hide strings in the remember codename widget

### LINKED ISSUE #6287: Untranslated strings in "remember codename" widget
## Description

image

image

"Show" and "Hide" are injected via CSS, therefore never translated.

## Comments

Fixed easily by adding data attributes on the element and referencing those with CSS's `attr()` function

eaon:
## Status

Ready for review

## Description of Changes

Add two new translated strings "Show" and "Hide" as data attributes in the codename widget and reference them from CSS.

Fixes #6287

## Testing

Because there's currently no translations for these strings, this is currently a kind of no-op, however a manual check can be done with the browser's Developer Tools by changing the contents of the `data-show` and `data-hide` attributes. 

## Checklist

### DIFF
--- a/securedrop/sass/source.sass
+++ b/securedrop/sass/source.sass
@@ -40,7 +40,7 @@
       cursor: pointer
 
       &:after
-        content: "Show"
+        content: attr(data-show)
         display: block
         float: right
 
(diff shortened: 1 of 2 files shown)


legoktm (approved):
Tested, LGTM. This looks like the first time we've used `attr()` in CSS, but it all works.

Issue #6293, opened 2022-02-18
investigate and implement improvements to header-based tor2web detection for current active tor2web proxies


zenmonkeykstop:

legoktm:
From poking at it a bit and reviewing the tor2web source code, I came up with https://github.com/freedomofpress/securedrop/pull/6304, which relies on buggy URL mangling on the `/create` form. I've marked it as a draft because the UX is pretty bad, you have to click through like 2 forms and only then do we fire the tor2web alert. How bad would it be if the very first "Get started" button was actually a `<form>` that POST'd to `/generate`, so we could include the hidden form field and do the detection at the very first step?

zenmonkeykstop:
This is really neat - I'm trying to think of downsides to converting the Get Started button to a POSTed form but none spring to mind. Does it detect tor2web specifically tho, or any proxy that rewrites links? Both are bad, but we might need to make the language more general on the warning page in the latter case.

It would still be good to detect tor2web via headers if possible. It has fewer UX implications and it could happen on the first connection. But this is a great extra defense.

eaon:
Oh my, I *love* this! I don't think there's downsides to a `POST`ed Get Started button - it'd be bad for search engines, but messing with those is an upside in our case anyway.

> Does it detect tor2web specifically tho, or any proxy that rewrites links?

Given that it's based on a `href=` inside an attribute, I'd expect other naïve proxies that do simple substitution to trip over this as well, but realistically speaking, how many of those are configured to access onion services and would they be used for anything other than a service that's essentially the same as Tor2Web? I think it'd be enough to adapt the language on the warning to something along the lines of "If you use a Tor2Web or similar proxy"

zenmonkeykstop:
Yeah I think we're all of the same mind here, will tweak the language in #6300 a little. @legoktm I would say go for it on the change to the Get Started button.

legoktm:
The rewrites that tor2web do are pretty naive, it's entirely possible that it becomes more robust in the future (e.g. using a proper HTML parser) and we need some other detection mechanism. I think this should catch any similarly naive proxy that wants to make onion services accessible on the clearnet, because they have to hijack any *.onion URL.

PR #6300, opened 2022-02-22
Updated tor2web check to return a 403 status on success

### LINKED ISSUE #6291: update tor2web detection response, replacing warning flash message with a redirect to a static page.

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Fixes #6291 .

Changes the SI behaviour when a tor2web request is detected. Previously, the application would display a warning flash message on the page, with a link to `info/tor2web_warning` summarizing the risks of using tor2web. Now, the application will respond with the warning page directly, served with a 403 error status.

The warning page now also contains the correct onion URL for the instance, and text has been updated:

tor2web

Note that by itself this PR will not improve detection of tor2web proxies, preserving the current logic of checking for the `X-tor2web` header (which is rarely present anymore).

## Testing

- check out this branch and run `make dev-tor`
- connect to the docker container with `docker exec -it securedrop-dev-0 bash` and add a file `/var/lib/securedrop/source_v3_url` containing a single line with a validly-formatted v3 onion URL (`<56chars>.onion`, no protocol or slashes)
- [x] verify that the SI is available (via `localhost:8080` or over Tor) and that basic functionality works with no tor2web warnings or redirects
- [x] verify that the v3 url specified above is available via the SI `/metadata` endpoint
- add an X-tor2web header to your browser requests (eg. via a browser addon like [ModHeader](https://modheader.com/))
- [x] verify that any requests for non-static resources on the SI result in a 403 and the Tor2Web warning page in the screenshot above
- [x] verify that the v3 url you specified matches the one displayed on the warning page.
- [x] visit the Tor version of the dev environment via a tor2web proxy and verify that the v3 url in the page has *not* been rewritten.

## Deployment

 n/a

Choose one of the following:

### DIFF
--- a/securedrop/source_app/__init__.py
+++ b/securedrop/source_app/__init__.py
@@ -4,8 +4,7 @@ from typing import Optional
 import os
 import time
 import werkzeug
-from flask import (Flask, flash, render_template, request, g, session,
-                   url_for)
+from flask import (Flask, render_template, request, g, session, redirect, url_for)
 from flask_babel import gettext
 from flask_assets import Environment
 from flask_wtf.csrf import CSRFProtect, CSRFError
@@ -23,7 +22,7 @@ from request_that_secures_file_uploads import RequestThatSecuresFileUploads
 from sdconfig import SDConfig
 from source_app import main, info, api
 from source_app.decorators import ignore_static
-from source_app.utils import clear_session_and_redirect_to_logged_out_page, get_sourcev3_url
+from source_app.utils import clear_session_and_redirect_to_logged_out_page
 
 
 def get_logo_url(app: Flask) -> str:
@@ -110,10 +109,11 @@ def create_app(config: SDConfig) -> Flask:
 
     @app.before_request
     @ignore_static
-    def check_tor2web() -> Optional[Tuple[str, int]]:
+    def check_tor2web() -> Optional[werkzeug.Response]:
+        # TODO: expand header checking logic to catch modern tor2web proxies
         if 'X-tor2web' in request.headers:
-            flash(gettext("Tor2Web proxies do not protect your anonymity!"), "error")
-            return render_template('tor2web-warning.html', source_url=get_sourcev3_url()), 403
+            if request.path != url_for('info.tor2web_warning'):
+                return redirect(url_for('info.tor2web_warning'))
         return None
 
     @app.errorhandler(404)
--- a/securedrop/source_app/__init__.py
+++ b/securedrop/source_app/__init__.py
@@ -4,7 +4,7 @@ from typing import Optional
 import os
 import time
 import werkzeug
-from flask import (Flask, render_template, escape, flash, Markup, request, g, session,
+from flask import (Flask, flash, render_template, request, g, session,
                    url_for)
 from flask_babel import gettext
 from flask_assets import Environment
@@ -23,7 +23,7 @@ from request_that_secures_file_uploads import RequestThatSecuresFileUploads
 from sdconfig import SDConfig
 from source_app import main, info, api
 from source_app.decorators import ignore_static
-from source_app.utils import clear_session_and_redirect_to_logged_out_page
+from source_app.utils import clear_session_and_redirect_to_logged_out_page, get_sourcev3_url
 
 
 def get_logo_url(app: Flask) -> str:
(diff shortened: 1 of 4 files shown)


ninavizz:
The body and header text both need some work, imho. 

**1. Active voice.** 
    - OSS trends towards being far too passive, wanting to leave things open for users. In UI messaging, people just want to be told what "is" and what "to do." 
**2. Headers should rarely exceed 2 words.** What is the page topic, as TL;DR as possible. Not complete sentences. Speculatively, "Tor2Web Detected" seems like the most appropriate header, with an inline error message below that says as its declarative "Anonymity Risk" and then in body text after that, qualify why. Then, include a link to Tor Browser, after that.
**3. If in fact this is a warning page, an inline error-bang warning** should also display with this page. Similar to how it's styled in the new SI. See above text recommendation. 
4. Second sentence and "strongly advise" both sound too wishy-washy, and the second sentence takes too long to cut to the chase.
5. Please do not leave out private adversaries; returning to the personas, Ed Snowden is not the primary whistleblower SD targets. Speaking to "government" as a broad concept, instead of more abstractly "public or private adversaries"

legoktm (commented):
Agreed that "strongly advise" doesn't cut it here, since we're actually fully blocking access to SDs if we see Tor2Web is being used.

In the commit message and PR title, you described this as a "redirect", but it's not actually a redirect, it just overrides the current page. I think actually doing a redirect to an endpoint like `/tor2web` would be nicer if we end up implementing a proxy/URL check in JavaScript.

eloquence:
I agree with Nina and Kunal that, given that the behavior change is from warning to blocking, we'll need to update the page a bit more to make that explicit. It may also be worth making it even clearer that, at this point, the user has effectively disclosed their IP address to the Tor2web operator (and apparently Google Analytics and ad providers for many of them). In other words, depending on their behavior or threat model, it may no longer be safe for them to make a submission at all.

I realize most of the page pre-dates the PR, but I think it'd be worth workshopping the entire page content a bit in a GDoc or Figma, assuming we can actually reliably detect Tor2web gateways in one way or another.

zenmonkeykstop:
I'm down to make the changes mentioned - afterwards if we do want to workshop it I'd prefer to get that scheduled sooner rather than later, in the interest of getting this change out the door in a reasonable timeframe.

zenmonkeykstop:
@legoktm re: redirect word, that's just sloppy terminology on my part, but I do think that a 403 is appropriate given that the intent is to deny access. Will update the PR. Also, the original `/info/tor2web-warning` route is preserved, precisely so that it can be used with a Javascript check.

zenmonkeykstop:
Not down with the "public or private adversaries" bit on further reflection, as it's probably best not to overload sources with domain-specific terminology. Also, the current text does speak accurately, tho not completely, to the specific adversaries a source in this situation would have to worry about.

eloquence:
Could say "(your employer, your government, your Internet provider)" which doesn't strike me as overly unwieldy. But I think after you have a version up you're happy with, maybe we can have a quick huddle on Wednesday or Thursday to futz a bit more.

zenmonkeykstop:
OK, new version up.

- updated header as suggested by @ninavizz 
- added error flash message using existing SI flash styling. Note that the existing flash message format does not have a declarative (unless you add it as markup, which I'd prefer to avoid), so it doesn't have that part. IMO this can be revisited in the context of ongoing design updates.
- decreased wishy-washiness by 47% (took two passes as in the first pass I added "please" everywhere...)
- see previous reservations re "adversaries" etc. - updated list to include the proxy operator, who, given historical events and obvious DPI nonsense currently going on, are probably one of the more significant threats.

tor2web2

legoktm:
Technically LGTM, not giving it the green check in case people have other wording concerns. One thing I was worried about was that the proxy would try to rewrite the onion URL, but in my testing it only does so if it's in a `href`, not if it's just in plain text - so we're good on that front.

zenmonkeykstop:
tweaked the language a bit based on the discussion in #6293 - now looks like this:

tor2web3

eloquence:
I've proposed a slight reword of the page here:
https://pad.riseup.net/p/tor2web-warning

Proposed changes in a nutshell:
- Active voice for "You are not anonymous right now" warning
- Simplify some language
- Removed article "the" in front of "Tor Browser" (see Gonzalo's similar cleanup commit: https://github.com/freedomofpress/securedrop-docs/commit/0071f52a014081643c93ee19f0715d17532abc88)
- Remind them to note down the .onion address if they click "New identity", since it will disappear from view otherwise

Please feel free to futz further in the pad or challenge any of those tweaks :)

securedrop/source_app/__init__.py
```diff
@@ -128,6 +108,14 @@ def setup_g() -> Optional[werkzeug.Response]:
 
         return None
 
+    @app.before_request
+    @ignore_static
+    def check_tor2web() -> Optional[Tuple[str, int]]:
+        if 'X-tor2web' in request.headers:
+            flash(gettext("Tor2Web proxies do not protect your anonymity!"), "error")
```
legoktm: Can we refactor out these two lines, the flash and template into a separate function? Then I can reuse it in my PR rather than redirecting to `/tor2web-warning`.

securedrop/source_app/info.py
```diff
@@ -1,20 +1,23 @@
...
 
 
 def make_blueprint(config: SDConfig) -> Blueprint:
     view = Blueprint('info', __name__)
 
     @view.route('/tor2web-warning')
     def tor2web_warning() -> str:
-        return render_template("tor2web-warning.html")
+        flash(gettext("Tor2Web proxies do not protect your anonymity!"), "error")
+        return render_template("tor2web-warning.html", source_url=get_sourcev3_url())
```
legoktm: Probably this route should always return a 403?

securedrop/source_app/__init__.py
zenmonkeykstop: went a different route (:drum:) - /tor2web-warning now has 403 status and this check redirects to it.

securedrop/source_app/info.py
zenmonkeykstop: :heavy_check_mark:

zenmonkeykstop:
YA round of text updates is complete. See below:

tor2web4

eloquence:
LGTM! Thanks for patiently poking at it together.

legoktm (approved):
Confirmed test plan, LGTM. If there are still concerns with the new wording, it can be adjusted in a follow-up.

PR #6302, opened 2022-02-22
Add honeypot to Source Interface to stop very basic spambots

### LINKED ISSUE #6295: Add honeypot field to SI message form to block automated submissions
## Description

The Source Interface has limited protections in place to block bots. One simple bot-blocking measure would be to add a hidden form field on the message submission form, and reject submissions that include a value for the field.

## User Research Evidence

Spam reports from instance maintainers.

## User Stories

As a journalist, I want to save time by not having to review spam messages.

legoktm:
## Status

Ready for review

## Description of Changes

Add a hidden, empty input field to the Source Interface form that will
abort if anything is submitted in it. This will hopefully stop the most
basic of spambots that fill in spam in every form field they see and
submit the form. It is unlikely to stop any spambot specifically
designed for SecureDrop, as those will adapt as necessary.

This is inspired by and copied from MediaWiki's "SimpleAntiSpam"
functionality originally written by Ryan Schmidt[1].

[1] https://www.mediawiki.org/w/index.php?oldid=797352

Fixes #6295.

## Testing

- [ ] Create a new source and submit a dummy message in the SI, observe that it works fine
- [ ] Use the browser's inspector to disable the "display: none", write something in the antispam input field, and submit. You should get a 403 error screen.

## Deployment

No concerns.

### DIFF
--- a/securedrop/sass/source.sass
+++ b/securedrop/sass/source.sass
@@ -308,3 +308,7 @@ p#max-file-size
     background:
       image: url(/static/icons/chevron-left.png)
       size: contain
+
+/** This is the vaguely named antispam check */
+#text-field
+  display: none
--- a/securedrop/source_app/forms.py
+++ b/securedrop/source_app/forms.py
@@ -1,7 +1,8 @@
 import wtforms
+from flask import abort
 from flask_babel import lazy_gettext as gettext
 from flask_wtf import FlaskForm
-from wtforms import FileField, PasswordField, TextAreaField
+from wtforms import FileField, PasswordField, StringField, TextAreaField
 from wtforms.validators import InputRequired, Regexp, Length, ValidationError
 
 from models import Submission, InstanceConfig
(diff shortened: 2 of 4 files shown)


legoktm:
The HTML lint check is failing with:
```
securedrop/source_templates/lookup.html:61:10: Error: Separation of concerns: CSS should be defined in its own file: Move the contents of the "style" attribute to its own CSS file.
```

Putting the CSS inline is intentional, it ensures the field is always hidden, regardless of other CSS rules or if CSS fails to load for whatever reason. That said, I don't think it's a big loss to put it in the CSS and can do so if others think we should.

zenmonkeykstop:
:+1: to managing it via SASS like the rest of the CSS. Don't want to end up with a bunch of unnecessary linting exemptions

legoktm:
Latest version renamed the form field to a boring "text", while keeping the name "antispam" server-side. And moved the inline styling to CSS.

zenmonkeykstop (approved):
Test plan looks good, code changes are minimal, field name is sufficiently boring. :+1:

PR #6304, opened 2022-02-23
Use Tor2Web's URL mangling to detect it

### LINKED ISSUE #6293: investigate and implement improvements to header-based tor2web detection for current active tor2web proxies

legoktm:
## Status

Ready for review

## Description of Changes

Tor2Web mangles onion URLs to point to the proxy, but its implementation
is buggy and we can use that for detection. In this case, it rewrites
a literal `href="fake.onion"` (inside a hidden form `<input>`) to
something like: `href="//{$address}.onion.ly/fake.onion.ly"`. On form
submission, we can detect that it doesn't roundtrip properly and flag
the mangling as Tor2Web usage.

Since we want to show users the error as soon as possible, we turn the
initial "Get started" button link into a `<form>` with this special hidden
`<input>` that POSTs to `/generate`, which will check if Tor2Web mangles
the field and display the warning accordingly.

Fixes #6293.

## Testing

- [ ] Run `make dev-tor`, access over tor2web hit "Get Started". You should get the warning screen rather than seeing your codename.
- [ ] Try again but over the real onion rather than tor2web, you shouldn't see any errors.

## Deployment

## Checklist

### DIFF
--- a/securedrop/source_app/main.py
+++ b/securedrop/source_app/main.py
@@ -37,8 +37,15 @@ def make_blueprint(config: SDConfig) -> Blueprint:
     def index() -> str:
         return render_template('index.html')
 
-    @view.route('/generate', methods=('GET',))
+    @view.route('/generate', methods=('POST',))
     def generate() -> Union[str, werkzeug.Response]:
+        # Try to detect Tor2Web usage by looking to see if tor2web_check got mangled
+        tor2web_check = request.form.get('tor2web_check')
+        if tor2web_check is None:
+            # Missing form field
+            abort(403)
+        elif tor2web_check != 'href="fake.onion"':
+            return redirect(url_for('info.tor2web_warning'))
         if SessionManager.is_user_logged_in(db_session=db.session):
             flash(gettext(
                 "You were redirected because you are already logged in. "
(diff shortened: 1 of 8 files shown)


securedrop/source_app/main.py
```diff
@@ -65,6 +65,9 @@ def generate() -> Union[str, werkzeug.Response]:
 
     @view.route('/create', methods=['POST'])
     def create() -> werkzeug.Response:
+        # Try to detect Tor2Web usage by looking to see if source_url got mangled
+        if request.form['source_url'] != 'href="fake.onion"':
+            return redirect(url_for('info.tor2web_warning'))
```
zenmonkeykstop: Would it make sense to give this a 403 status as well?

legoktm: Yeah, I think it would be better to just flash/return the warning here rather than redirecting. See my comment on your PR for exposing it as a reusable function.

legoktm:
The `TestDuplicateSourceInterface.test_refreshed_duplicate_generate_pages` test is still failing for reasons I haven't been able to figure out yet. Otherwise I think it's basically ready.

zenmonkeykstop (approved):
LGTM, tested with a few tor2web gateways (.ly, .ws, darknet.to), and no adverse effects from the GET vs POST.

PR #6306, opened 2022-02-25
Add basic message filtering in the SI

### LINKED ISSUE #6297: Add  message length checks to the Source Interface
## Description

Some production instances have seen an uptick in messages containing only a few characters and no actionable content. This may be attributable to bots, or to UX failures, or a combination of both. Adding a length requirement on initial messages would help in both cases. 
This change should take account of the following:

- the requirement should be applied on initial submissions only. Arbitrarily-short messages should be considered valid in the case of a source-journalist conversation after initial contact
- the requirement should not apply to initial submissions that also include a document, as the actionable content may be in the document
- the length requirement should be displayed on the submission form for initial submissions, to avoid unpleasant surprises for sources.
- rejected messages should result in useful feedback to sources, and should not prevent them from submitting a valid message afterwards
- the minimum length should be configurable by admins in the JI instance config panel, and it should be disabled by default.

## User Research Evidence

Spam reports from instance admins.

### LINKED ISSUE #6298: Add check for codename-only messages in Source Interface
## Description

Some production instances have seen an uptick in initial messages from sources containing only a source codename and no actionable content. This may be attributable to bots, or to UX failures, or a combination of both. Planned changes to the source workflow will mitigate both cases, but in the interim a check should be added to reject codename-only initial messages.

This change should take account of the following:

• the requirement should be applied on initial submissions only. Returning sources will have had to save their codenames and use them to log in, and therefore can reasonably be expected to do the right thing
• rejected messages should result in useful feedback to sources, and should not prevent them from submitting a valid message afterwards
• the feature should be configurable by admins in the JI instance config panel, and it should be disabled by default.

## User Research Evidence

Spam reports by instance admins.

zenmonkeykstop:
## Status

 Work in progress (tests required)

## Description of Changes

Fixes #6297.
Fixes #6298.

adds options in the JI instance config, and in the SI,  to:
- set a minimum message length for source's initial messages, as per #6297 
- reject initial messages that consist of the source's codename, as per #6298

## Testing
- check out this branch and run `make dev`

### minimum message length
- in the JI, navigate to the Instance Config page via the Admin page
  - [x] verify that the Submission Preferences  section includes a "Prevent sources from sending initial messages shorter than the minimum required length" option, currently unchecked and with the length field set to 0.
- check the "Prevent sources from sending initial messages shorter than..." checkbox but do not set a length. Click **Update Submission Preferences**
  - [x] verify that the checkbox is unchecked and an error message is displayed asking you to set a length.
 - check the "Prevent sources from sending initial messages shorter than..." checkbox and set a length of between 5 and 20 chars. Click **Update Submission Preferences**
   - [x] verify that the checkbox is checked, the length you chose is set, and a success message is displayed.
- Create a new source account on the SI, navigating through to the /lookup page
  - [x] verify that a notice is displayed under the message textfield like "If you are only sending a message, it must be at least N characters long" where N is the length you chose above
  - [x] Verify that if you try to submit  too short a message, the submission fails with a flashed error like "Your first message must be at least N characters long"
  - [x] Verify that a message N characters long can be submitted successfully
  - [x] after a successful first submission, verify that the  notice under the text field is no longer displayed and you can submit message shorter than N chars
  - [x] in the JI, uncheck the "Prevent sources from sending initial messages shorter than..." checkbox,, click **Update Submission Prefs** and verify that the length is set to zero and a success message is displayed
  - [x] Create a new source account in the SI and verify that no message length limit is mentioned or set

### codename messages

- in the JI, navigate to the Instance Config page via the Admin page
  - [x] verify that the Submission Preferences  section includes a "Prevent sources from submitting their codename as an initial message." option, currently unchecked.
- check the "Prevent sources from submitting their codename as an initial message" checkbox. Click **Update Submission Preferences**
   - [x] verify that the checkbox is checked and a success message is displayed.
 - Create a new source account on the SI, navigating through to the /lookup page
   - [x] Enter the codename as a message and click **Submit** - verify that the message is not submitted and an error is flashed like "Please do not submit your codename.."
   - [x] Enter anything else as a message and click **Submit** - verify that the message was submitted correctly
   - [x]  Enter your codename again and click **Submit** - verify that the message was submitted correctly.

## Deployment

DB migration script required, but new functionality is disabled by default.

Choose one of the following:

### DIFF
--- a/securedrop/alembic/versions/55f26cd22043_remove_partial_index_on_valid_until_set_.py
+++ b/securedrop/alembic/versions/55f26cd22043_remove_partial_index_on_valid_until_set_.py
@@ -1,0 +1,57 @@
+"""remove partial_index on valid_until, set it to nullable=false, add message filter fields
+
+Revision ID: 55f26cd22043
+Revises: 2e24fc7536e8
+Create Date: 2022-03-03 22:40:36.149638
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+from datetime import datetime
+
+# revision identifiers, used by Alembic.
+revision = '55f26cd22043'
+down_revision = '2e24fc7536e8'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # remove the old partial index on valid_until, as alembic can't handle it.
+    op.execute(sa.text('DROP INDEX IF EXISTS ix_one_active_instance_config'))
+
+    # valid_until will be non-nullable after batch, so set existing nulls to
+    # the new default unix epoch datetime
+    op.execute(sa.text(
+        'UPDATE OR IGNORE instance_config SET '
+        'valid_until=:epoch WHERE valid_until IS NULL;'
+    ).bindparams(epoch=datetime.fromtimestamp(0)))
+    with op.batch_alter_table('instance_config', schema=None) as batch_op:
+        batch_op.alter_column('valid_until',
+                              existing_type=sa.DATETIME(),
+                              nullable=False)
+        batch_op.add_column(sa.Column('initial_message_min_len', sa.Integer(), nullable=True))
+        batch_op.add_column(sa.Column('reject_message_with_codename', sa.Boolean(), nullable=True))
+
+    # remove the old partial index *again* in case the batch op recreated it.
+    op.execute(sa.text('DROP INDEX IF EXISTS ix_one_active_instance_config'))
+
+
+def downgrade():
+    with op.batch_alter_table('instance_config', schema=None) as batch_op:
+        batch_op.alter_column('valid_until',
+                              existing_type=sa.DATETIME(),
+                              nullable=True)
+        batch_op.drop_column('reject_message_with_codename')
+        batch_op.drop_column('initial_message_min_len')
+
+    # valid_until is nullable again, set entries with unix epoch datetime value to NULL
+    op.execute(sa.text(
+        'UPDATE OR IGNORE instance_config SET '
+        'valid_until = NULL WHERE valid_until=:epoch;'
+            ).bindparams(epoch=datetime.fromtimestamp(0)))
+
+    # manually restore the partial index
+    op.execute(sa.text('CREATE UNIQUE INDEX IF NOT EXISTS ix_one_active_instance_config ON '
+                       'instance_config (valid_until IS NULL) WHERE valid_until IS NULL'))
--- a/securedrop/journalist_app/admin.py
+++ b/securedrop/journalist_app/admin.py
@@ -15,7 +15,7 @@ from sqlalchemy.orm.exc import NoResultFound
 from db import db
 from html import escape
 from models import (InstanceConfig, Journalist, InvalidUsernameException,
-                    FirstOrLastNameError, PasswordError)
+                    FirstOrLastNameError, PasswordError, Submission)
 from journalist_app.decorators import admin_required
 from journalist_app.utils import (commit_account_changes, set_diceware_password,
                                   validate_hotp_secret, revoke_token)
@@ -36,9 +36,18 @@ def make_blueprint(config: SDConfig) -> Blueprint:
     @view.route('/config', methods=('GET', 'POST'))
     @admin_required
     def manage_config() -> Union[str, werkzeug.Response]:
-        # The UI prompt ("prevent") is the opposite of the setting ("allow"):
+        if InstanceConfig.get_default().initial_message_min_len > 0:
+            prevent_short_messages = True
+        else:
+            prevent_short_messages = False
+
+        # The UI document upload prompt ("prevent") is the opposite of the setting ("allow")
         submission_preferences_form = SubmissionPreferencesForm(
-            prevent_document_uploads=not InstanceConfig.get_default().allow_document_uploads)
+            prevent_document_uploads=not InstanceConfig.get_default().allow_document_uploads,
+            prevent_short_messages=prevent_short_messages,
+            min_message_length=InstanceConfig.get_default().initial_message_min_len,
+            reject_codename_messages=InstanceConfig.get_default().reject_message_with_codename
+            )
         organization_name_form = OrgNameForm(
             organization_name=InstanceConfig.get_default().organization_name)
         logo_form = LogoForm()
@@ -67,6 +76,7 @@ def make_blueprint(config: SDConfig) -> Blueprint:
             return render_template("config.html",
                                    submission_preferences_form=submission_preferences_form,
                                    organization_name_form=organization_name_form,
+                                   max_len=Submission.MAX_MESSAGE_LEN,
                                    logo_form=logo_form)
 
     @view.route('/update-submission-preferences', methods=['POST'])
@@ -75,9 +85,17 @@ def make_blueprint(config: SDConfig) -> Blueprint:
         form = SubmissionPreferencesForm()
         if form.validate_on_submit():
             # The UI prompt ("prevent") is the opposite of the setting ("allow"):
+            allow_uploads = not form.prevent_document_uploads.data
+
+            if form.prevent_short_messages.data:
+                msg_length = form.min_message_length.data
+            else:
+                msg_length = 0
+
+            reject_codenames = form.reject_codename_messages.data
+
+            InstanceConfig.update_submission_prefs(allow_uploads, msg_length, reject_codenames)
             flash(gettext("Preferences saved."), "submission-preferences-success")
-            value = not bool(request.form.get('prevent_document_uploads'))
-            InstanceConfig.set_allow_document_uploads(value)
             return redirect(url_for('admin.manage_config') + "#config-preventuploads")
         else:
             for field, errors in list(form.errors.items()):
--- a/securedrop/sass/source.sass
+++ b/securedrop/sass/source.sass
@@ -259,6 +259,11 @@ p#max-file-size
   font-style: italic
   margin-bottom: 0
 
+p#min-msg-length
+  font-size: 10px
+  font-style: italic
+  margin-bottom: 12px
+
 .bubble
   width: 415px
   position: absolute
--- a/securedrop/source_app/main.py
+++ b/securedrop/source_app/main.py
@@ -7,7 +7,7 @@ from datetime import datetime, timedelta, timezone
 from typing import Union
 
 import werkzeug
-from flask import (Blueprint, render_template, flash, redirect, url_for,
+from flask import (Blueprint, render_template, flash, redirect, url_for, escape,
                    session, current_app, request, Markup, abort, g, make_response)
 from flask_babel import gettext
 
@@ -24,7 +24,7 @@ from sdconfig import SDConfig
 from source_app.decorators import login_required
 from source_app.session_manager import SessionManager
 from source_app.utils import normalize_timestamps, fit_codenames_into_cookie, \
-    clear_session_and_redirect_to_logged_out_page
+    clear_session_and_redirect_to_logged_out_page, codename_detected
 from source_app.forms import LoginForm, SubmissionForm
 from source_user import InvalidPassphraseError, create_source_user, \
     SourcePassphraseCollisionError, SourceDesignationCollisionError, SourceUser
@@ -119,6 +119,13 @@ def make_blueprint(config: SDConfig) -> Blueprint:
             source_id=logged_in_source_in_db.id, deleted_by_source=False
         ).all()
 
+        first_submission = logged_in_source_in_db.interaction_count == 0
+
+        if first_submission:
+            min_message_length = InstanceConfig.get_default().initial_message_min_len
+        else:
+            min_message_length = 0
+
         for reply in source_inbox:
             reply_path = Storage.get_default().path(
                 logged_in_source.filesystem_id,
@@ -158,6 +165,7 @@ def make_blueprint(config: SDConfig) -> Blueprint:
             is_user_logged_in=True,
             allow_document_uploads=InstanceConfig.get_default().allow_document_uploads,
             replies=replies,
+            min_len=min_message_length,
             new_user_codename=session.get('new_user_codename', None),
             form=SubmissionForm(),
         )
@@ -191,6 +199,23 @@ def make_blueprint(config: SDConfig) -> Blueprint:
         fnames = []
         logged_in_source_in_db = logged_in_source.get_db_record()
         first_submission = logged_in_source_in_db.interaction_count == 0
+
+        if first_submission:
+            min_len = InstanceConfig.get_default().initial_message_min_len
+            if (min_len > 0) and (msg and not fh) and (len(msg) < min_len):
+                flash(gettext(
+                    "Your first message must be at least {} characters long.".format(min_len)),
+                    "error")
+                return redirect(url_for('main.lookup'))
+
+            codenames_rejected = InstanceConfig.get_default().reject_message_with_codename
+            if codenames_rejected and codename_detected(msg, session['new_user_codename']):
+                flash(Markup('{}<br>{}'.format(
+                    escape(gettext("Please do not submit your codename!")),
+                    escape(gettext("Keep your codename secret, and use it to log in later"
+                                   " to check for replies."))
+                    )), "error")
+                return redirect(url_for('main.lookup'))
 
         if not os.path.exists(Storage.get_default().path(logged_in_source.filesystem_id)):
             current_app.logger.debug("Store directory not found for source '{}', creating one."
--- a/securedrop/tests/migrations/migration_55f26cd22043.py
+++ b/securedrop/tests/migrations/migration_55f26cd22043.py
@@ -1,0 +1,89 @@
+# -*- coding: utf-8 -*-
+from db import db
+from journalist_app import create_app
+import sqlalchemy
+import pytest
+import secrets
+
+from .helpers import random_bool, random_datetime, random_ascii_chars
+
+
+index_definition = (
+    "index",
+    "ix_one_active_instance_config",
+    "instance_config",
+    (
+        "CREATE UNIQUE INDEX ix_one_active_instance_config "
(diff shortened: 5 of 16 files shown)


legoktm:
Do we also want to use `<textarea minlength="...">`? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#attr-minlength

zenmonkeykstop:
> Do we also want to use `<textarea minlength="...">`? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#attr-minlength

Yep, that'd be a nice addition.

zenmonkeykstop:
Looking at that minlength addition with @eaon, it occurred to me that it would prevent initial submissions with a file attachment but no message (which are currently allowed). So, in the absence of client-side logic to allow for that case I think we can't include it except for the edge case upon an edge case of message-only configs. Length checks will be server-side only.

eloquence:
Thanks for all the hard work on this! A couple of initial UX notes:

## Organization of new preferences is potentially confusing

Screenshot from 2022-03-03 17-46-00

There are IMO a few UX issues here:
- Having the user have to set a "0" value to enable or disable a preference makes it harder for them to follow a single process for all yes/no preferences, to quickly parse the state of preferences, and to know what the previous state was when the feature is disabled.
- One way to avoid this would be to have a "Prevent initial messages below a minimum character length" preference, with the actual character length either as an inline field, or as a separate, indented sub-option.
- Because other preferences follow the "Prevent .." format, the fact that the "Minimum message length" preference does not, can make it feel like it is associated with the previous preference.
- The actual input control only needs to accommodate 4 digits, so it can be far less wide.
## Error message scrolled out of view

Can you reproduce this? If I configure a minimum length and submit a message that gets blocked, the error message is not scrolled within the viewport:
Screenshot from 2022-03-03 17-45-45

I don't see this behavior with error messages in `develop` (e.g., when attempting to send a blank message).

zenmonkeykstop:
All seems doable, had planned to restrict that length field size via CSS, could probably just add an attribute directly as well.

I'm thinking that the extra para in the SI about message lengths might be what's pushing the flash error out of viewport - for neatness sake I was already thinking of moving that into the placeholder text in the textarea instead. Will push up something tomorrow.

zenmonkeykstop:
@eloquence a revision with some UX-related changes addressing your points above is working its way thru CI. Note that it does *not*  update the placeholder text in the submission textarea as I mentioned above - instead the same design language/placement as is currently used for file size limits was applied.

eloquence:
Thanks for the changes, it's coming together nicely. I'm still noticing the issue with the error message not being scrolled into view, not sure if you've been able to reproduce / had a chance to debug that one.

The new preferences form seems to work well. I think it would be nice if it remembered the previous value of the form field even after you disable the feature, but that's definitely a nice-to-have.

Regarding the UX changes, I would suggest tweaking as follows:
Screenshot from 2022-03-07 17-14-16

Tweaks:
- No background color (the gray color can suggest a "disabled" look, and does not IMO meaningfully add to the grouping; with the new wording, it seems clearer to me that these are distinct preferences)
- Narrower input
- Changed "minimum required" to just "minimum" ("prevent" in combination with "minimum" are IMO sufficiently clear)
- Avoid repetition in the content hint, style it a bit differently from the field label

Regarding the minimum length hint, I would suggest also slightly tweaking it:

> If your first submission is message-only, it must be at least 5 characters long.

->

> If you are only sending a message, it must be at least 5 characters long.

- Avoid the awkward construction "message-only"
- Keep it simple, trying to communicate too many qualifiers means the message is more likely to be ignored as "fine print".

zenmonkeykstop:
@eloquence further tweaks applied.

securedrop/sass/source.sass
```diff
@@ -255,6 +255,11 @@ p#max-file-size
   font-style: italic
   margin-bottom: 0
 
+p#min-msg-length
+  font-size: 10px
+  font-style: italic
+  margin-bottom: 0
```
eaon: I think this margin could be set to 12px - `#max-file-size` after which this is modeled does not have a bottom margin, but it's wrapped in an element with a 12px padding all around, which isn't really suitable for the `textarea` side of things

zenmonkeykstop: as you have spoken, so has it been actioned.

eloquence:
Thanks for the additional tweaks! Two remaining issues I'm noticing:

-  The error message if you don't set the integer makes direct reference to internal field names, presumably because it's standard built-in validation logic. (`Preferences not updated. The "min_message_length" field is required when "prevent_short_messages" is set.`) Can we customize that, e.g., "To configure a minimum message length, you must set the  required number of characters." ?
- I'm still seeing the issue with the flash message scrolling out of view.

Otherwise it LGTM, and potentially mergeable for 2.3.0 (I'm still queasy about the min-length change, but I'm OK with it as an opt-in to learn more; as discussed, we may want to at least tweak language on `/generate` to accommodate the codename blocking feature).

zenmonkeykstop:
re error, It's actually a custom validator added while fixing the OTP checks - lemme just throw an optional custom message in there real quick...

I'm not seeing the same issue with the flash message and the viewport. Will try to repro.

securedrop/alembic/versions/55f26cd22043_remove_partial_index_on_valid_until_set_.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def upgrade():
+    # remove the old partial index on valid_until, as alembic can't handle it.
+    op.execute(sa.text('DROP INDEX IF EXISTS ix_one_active_instance_config'))
+
+    # valid_until will be non-nullable after batch, so set existing nulls to
+    # the new default unix epoch datetime
+    op.execute(sa.text(
+        f'UPDATE OR IGNORE instance_config SET '  # nosec
```
legoktm: We don't need fstrings here since we're not actually interpolating anything, which probably fixes the security warning too.

securedrop/alembic/versions/55f26cd22043_remove_partial_index_on_valid_until_set_.py
```diff
@@ -0,0 +1,57 @@
...
+    with op.batch_alter_table('instance_config', schema=None) as batch_op:
+        batch_op.alter_column('valid_until',
+                              existing_type=sa.DATETIME(),
+                              nullable=True)
+        batch_op.drop_column('reject_message_with_codename')
+        batch_op.drop_column('initial_message_min_len')
+
+    # valid_until is nullable again, set entries with unix epoch datetime value to NULL
+    op.execute(sa.text(
+        f'UPDATE OR IGNORE instance_config SET '  # nosec
```
legoktm: Same, no f-strings needed

securedrop/journalist_app/admin.py
```diff
@@ -36,9 +36,18 @@ def index() -> str:
     @view.route('/config', methods=('GET', 'POST'))
     @admin_required
     def manage_config() -> Union[str, werkzeug.Response]:
-        # The UI prompt ("prevent") is the opposite of the setting ("allow"):
+        # The UI document upload prompt ("prevent") is the opposite of the setting ("allow")
```
legoktm: Should this comment be moved to be next to the `prevent_document_uploads=not...` line?

legoktm: Should this comment be moved to be next to the `prevent_document_uploads=not...` line?

securedrop/journalist_app/admin.py
```diff
@@ -36,9 +36,18 @@ def index() -> str:
...
+        if InstanceConfig.get_default().initial_message_min_len > 0:
+            prevent_short_messages = True
```
legoktm, line 40: This could be one line, `prevent_short_messages = InstanceConfig.get_default().initial_message_min_len > 0`

securedrop/journalist_app/admin.py
```diff
@@ -75,9 +85,27 @@ def update_submission_preferences() -> Optional[werkzeug.Response]:
...
             # The UI prompt ("prevent") is the opposite of the setting ("allow"):
+            allow_uploads = not form.prevent_document_uploads.data
+
+            if form.prevent_short_messages.data:
+                try:
+                    msg_length = form.min_message_length.data
+
+                    if isinstance(msg_length, int):
+                        int_length = msg_length
+                    elif isinstance(msg_length, str):
```
legoktm: How do you hit this case? I wasn't able to figure out how to reach it, even when hacking the HTML to post something like "aaa".

When I did submit non-int data, wtforms filled it with None, so then the RequiredIf validator tripped and displayed an error message.

securedrop/alembic/versions/55f26cd22043_remove_partial_index_on_valid_until_set_.py
```diff
@@ -0,0 +1,57 @@
...
+    op.execute(sa.text('DROP INDEX IF EXISTS ix_one_active_instance_config'))
+
+    # valid_until will be non-nullable after batch, so set existing nulls to
+    # the new default unix epoch datetime
+    op.execute(sa.text(
+        f'UPDATE OR IGNORE instance_config SET '  # nosec
+        f'valid_until=:epoch WHERE valid_until IS NULL;'
+    ).bindparams(epoch=datetime.fromtimestamp(0)))
+    with op.batch_alter_table('instance_config', schema=None) as batch_op:
+        batch_op.alter_column('valid_until',
```
legoktm, line 31: Don't we need something to make this unique?

securedrop/tests/migrations/migration_55f26cd22043.py
```diff
@@ -0,0 +1,91 @@
...
+        self.app = create_app(config)
+
+    def load_data(self):
+        pass
+
+    def check_downgrade(self):
+        assert index_definition in get_schema(self.app)
+
+        with self.app.app_context():
+            with pytest.raises(sqlalchemy.exc.OperationalError):
```
legoktm: The `raises()` should be inside the for loop, otherwise after it raises on the first one it won't run the second query.

securedrop/tests/migrations/migration_55f26cd22043.py
```diff
@@ -0,0 +1,91 @@
...
+                for query in [
+                     "SELECT * FROM instance_config WHERE initial_message_min_len IS NOT NULL",
+                     "SELECT * FROM instance_config WHERE reject_message_with_codename IS NOT NULL"
+                     ]:
+
+                    result = db.engine.execute(sqlalchemy.text(query)).fetchall()
+                    assert len(result) == 0
```
legoktm: This seems unnecessary? Since it'll raise.

legoktm (changes requested):

securedrop/alembic/versions/55f26cd22043_remove_partial_index_on_valid_until_set_.py
zenmonkeykstop: It's already unique - the gotcha is that uniqueness is not enforced on NULL values in SQL.

securedrop/journalist_app/admin.py
zenmonkeykstop: This is just paranoia on my part, it probably is unnecessary.

securedrop/journalist_app/admin.py
zenmonkeykstop: sure.

securedrop/journalist_app/admin.py
zenmonkeykstop: Prefer to leave it as is for readability.

securedrop/tests/migrations/migration_55f26cd22043.py
zenmonkeykstop: aye, deleted.

securedrop/tests/migrations/migration_55f26cd22043.py
zenmonkeykstop: aye, moved.

securedrop/journalist_app/admin.py
zenmonkeykstop: (nuked)

securedrop/alembic/versions/55f26cd22043_remove_partial_index_on_valid_until_set_.py
zenmonkeykstop: fixed

securedrop/alembic/versions/55f26cd22043_remove_partial_index_on_valid_until_set_.py
zenmonkeykstop: fixed.

securedrop/journalist_app/admin.py
zenmonkeykstop: duplicate comment?

securedrop/alembic/versions/55f26cd22043_remove_partial_index_on_valid_until_set_.py
legoktm: Ack!

eloquence:
The viewport issue seems very dependent on the window size. At a high vertical resolution, this is the kind of worst case that I'm seeing:
flash

It's possible that attempting to anchor to the flash message is more trouble than it's worth.

cfm:
<https://github.com/freedomofpress/securedrop/pull/6306#issuecomment-1063580054>:

> The viewport issue seems very dependent on the window size. At a high vertical resolution, this is the kind of worst case that I'm seeing: [...]
> 
> It's possible that attempting to anchor to the flash message is more trouble than it's worth.

This originated in #6041 as an [accessibility feature](https://docs.google.com/document/d/1jZlKco2YgOrrI4PmTkG_Q8lKuV5VxCds/edit?disco=AAAAIplLpuY) (see especially top of page 53), so that screen-readers can provide an equivalent to `#flashed`'s visual emphasis.  As much as I wish the browser's behavior here didn't force us to choose which experience (read: class of users) to privilege—at least without the benefit of the recommended JavaScript implementation—I agree that we shouldn't let this desideratum compromise the visual UX.

legoktm:
I'll double check with codecov after CI finishes, but I think all the new code should be covered by tests now.

eaon:
Regarding `#flashed` viewport jump issue, I've not paid enough attention to this behaviour before, but I just confirmed that this happens to me on the demo instance. I opened #6333 so we can separate that problem from the PR

legoktm:
Test plan all checks out (and pointed out a gap in the tests I wrote)!

zenmonkeykstop:
(for whoever ends up merging, this can probably be rebased into a single commit first)

securedrop/source_app/main.py
```diff
@@ -192,6 +200,23 @@ def submit(logged_in_source: SourceUser) -> werkzeug.Response:
         logged_in_source_in_db = logged_in_source.get_db_record()
         first_submission = logged_in_source_in_db.interaction_count == 0
 
+        if first_submission:
+            min_len = InstanceConfig.get_default().initial_message_min_len
+            if (min_len > 0) and (msg and not fh) and (len(msg) < min_len):
+                flash(gettext(
+                    "Your first message must be at least {} characters long.".format(min_len)),
+                    "error")
+                return redirect(f"{url_for('main.lookup')}#flashed")
```
eloquence: Per #6336 I think we'll want to now redirect without the #flashed fragment.

securedrop/source_app/main.py
```diff
@@ -192,6 +200,23 @@ def submit(logged_in_source: SourceUser) -> werkzeug.Response:
...
+
+            codenames_rejected = InstanceConfig.get_default().reject_message_with_codename
+            if codenames_rejected and codename_detected(msg, session['new_user_codename']):
+                flash(Markup('{}<br>{}'.format(
+                    escape(gettext("Please do not submit your codename!")),
+                    escape(gettext("Keep your codename secret, and use it to log in later"
+                                   " to check for replies."))
+                    )), "error")
+                return redirect(f"{url_for('main.lookup')}#flashed")
```
eloquence: As above.

eloquence:
I've pushed up a branch rebased to `develop`, with the anchor tweaks above, to `add-message-filtering-redux`, in case you want to use it. (I did not want to force push to this branch in case you have local changes.)

legoktm:
I'm assuming Kev had nothing left locally, so I'm going to cherry-pick 540a693bcdce595a2a8764c631bba2dfb0c50681 over to this branch, push, then squash+rebase it all and push again.

legoktm (dismissed):
I think we're set to land this but I'd like someone else to confirm that they reviewed the test cases I wrote (so I'm not merging them without review), and double check my squash and commit message.

cfm (dismissed):
Reviewed test cases and rebase/squash at @legoktm's request.  All looks good to me except for the following nit about the commit message on the squashed 500531deea0372c332b55f911fda5988b828fb64:

> When trying to do the schema change for the InstanceConfig table, it was
> discovered that SQLAlchemy/Alembic don't support the partial index on
> the table, so it's dropped and the valid_until column is now
> nullable=False. The current/active config is represented by a
> valid_until of the unix epoch (0).

On first reading, if one isn't already familiar with this problem, it might be unclear whether the clause beginning "so it's dropped" describes (a) SQLAlchemy's behavior in not supporting the partial index *prior to* this commit or (b) the fix made *by* this commit.  Up to you, @legoktm, whether it's worth revising, but since the commit (once merged) is the immutable thing here I thought I'd mention it.  :-)

zenmonkeykstop:
> On first reading, if one isn't already familiar with this problem, it might be unclear whether the clause beginning "so it's dropped" describes (a) SQLAlchemy's behavior in not supporting the partial index _prior to_ this commit or (b) the fix made _by_ this commit. Up to you, @legoktm, whether it's worth revising, but since the commit (once merged) is the immutable thing here I thought I'd mention it. :-)

Agreed, reworded.

conorsch (approved):
Re-blessing for merge after requested changes were made. :sunglasses:

PR #6307, opened 2022-02-27
Refactor functional test fixtures in source timeout and metadata tests

nabla-c0d3:
## Status

Ready for review

## Description of Changes

This PR includes and replaces the changes in https://github.com/freedomofpress/securedrop/pull/6207, and also addresses @legoktm 's feedback, such as putting the classes and functions in their "final" location in the code base.

Besides this addition, the intent and goals of the changes haven't changed, and are described [there](https://github.com/freedomofpress/securedrop/pull/6207) in details.


securedrop/tests/functional/app_navigators.py
```diff
@@ -0,0 +1,387 @@
+from contextlib import contextmanager
```
nabla-c0d3: Most of this is existing code from the `FunctionalTest` class that I've repackaged into separate classes to improve readability.

securedrop/tests/functional/db_session.py
```diff
@@ -0,0 +1,38 @@
...
+        yield db
+
+
+@contextmanager
+def get_database_session(database_uri: str) -> Generator[Session, None, None]:
+    """Easily get a session to the DB without having to deal with Flask apps.
+
+    Can be used for tests and utility scripts that need to add data to the DB outside of the
+    context of the source & journalist Flask applications.
+    """
```
nabla-c0d3, line 32: This could also eventually be used in utility scripts (to not have to instantiate a fake Flask app just to get "raw" access to the DB).

securedrop/tests/functional/factories.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+
+
+class SecureDropConfigFactory:
+    @staticmethod
+    def create(
+        SECUREDROP_DATA_ROOT: Path,
+        SESSION_EXPIRATION_MINUTES: float = 120,
+        NOUNS: Path = DEFAULT_SECUREDROP_ROOT / "dictionaries" / "nouns.txt",
+        ADJECTIVES: Path = DEFAULT_SECUREDROP_ROOT / "dictionaries" / "adjectives.txt",
```
nabla-c0d3, line 35: These arguments allow configuring the SD config that gets generated.

securedrop/tests/functional/factories.py
```diff
@@ -0,0 +1,79 @@
+from pathlib import Path
```
nabla-c0d3, line 1: This file contains two factories, which follow the API convention of the `factoryboy` library (in case this library is ever added to the unit tests dependencies). The factories can later be used in more code locations.

securedrop/tests/functional/test_source_session_timeout.py
```diff
@@ -1,17 +1,56 @@
...
+    )
+
+    # Ensure the GPG settings match the one in the config to use, to ensure consistency
+    journalist_key_fingerprint, gpg_dir = setup_journalist_key_and_gpg_folder
+    assert Path(config_with_short_timeout.GPG_KEY_DIR) == gpg_dir
+    assert config_with_short_timeout.JOURNALIST_KEY == journalist_key_fingerprint
+
+    # Spawn the apps in separate processes
+    with spawn_sd_servers(config_to_use=config_with_short_timeout) as sd_servers_result:
+        yield sd_servers_result
```
nabla-c0d3, line 31: The above is the "killer" feature of the new fixture and factory code: the source and journalist apps to be spawned will use the supplied `config_to_use` regardless of the content of `securedrop/config.py`. 

The confusion that comes from having to deal with live SDConfig objects that don't match the content of `securedrop/config.py` has been one of the biggest challenges for me when updating the test suite. This also comes from having different parts of the code sometimes use the live object or sometimes `securedrop/config.py`. 

I think one medium term goal (started with some of the changes here) is to get rid of `securedrop/config.py` in the context of the test suite.

securedrop/tests/functional/sd_config_v2.py
```diff
@@ -0,0 +1,101 @@
...
+    # This is recommended for performance, and also resolves #369
+    USE_X_SENDFILE: bool = True
+
+
+DEFAULT_SECUREDROP_ROOT = Path(__file__).absolute().parent.parent.parent
+
+
+# TODO(AD): This mirrors the field in an SDConfig; it is intended to eventually replace it
+@dataclass(frozen=True)
+class SecureDropConfig:
```
nabla-c0d3, line 28: I have added a test (below in this PR) to ensure that SDConfig (the "v1") and SecureDropConfig (the "v2") have the same attributes.

securedrop/tests/functional/web_drivers.py
```diff
@@ -0,0 +1,140 @@
+from contextlib import contextmanager
```
nabla-c0d3, line 1: Most of this is existing code taken from `FunctionalTest`.

securedrop/tests/functional/conftest.py
```diff
@@ -0,0 +1,158 @@
...
+class SdServersFixtureResult:
+    source_app_base_url: str
+    journalist_app_base_url: str
+
+
+@contextmanager
+def spawn_sd_servers(
+    config_to_use: SecureDropConfig
+) -> Generator[SdServersFixtureResult, None, None]:
+    """Spawn the source and journalist apps as separate processes with the supplied config."""
```
nabla-c0d3, line 83: The above is the "killer" feature of the new `spawn_sd_servers` fixture and other factory code: the source and journalist apps to be spawned will use the supplied config_to_use regardless of the content of `securedrop/config.py`.

The confusion that comes from having to deal with live SDConfig objects that don't match the content of securedrop/config.py has been one of the biggest challenges for me when updating the test suite. This also comes from having different parts of the code sometimes use the live object or sometimes securedrop/config.py (it also depends on the order of imports in the code base...).

I think one medium term goal (started with some of the changes here) is to get rid of securedrop/config.py in the context of the test suite.

nabla-c0d3:
This should be ready for review again. Thanks!

securedrop/tests/functional/app_navigators.py
```diff
@@ -0,0 +1,252 @@
+import time
```
nabla-c0d3, line 1: Most of this is existing code taken from `FunctionalTest` but "repackaged" as separate, smaller/simpler classes.

securedrop/tests/functional/sd_config_v2.py
legoktm: This is great, thanks.

securedrop/tests/functional/test_source_notfound.py
```diff
@@ -1,10 +1,11 @@
-from . import source_navigation_steps
-from . import functional_test
+# TODO(AD): This test duplicates TestSourceLayout.test_notfound() - remove it?
```
legoktm: Good catch, yeah, let's remove it.

securedrop/tests/functional/test_source_session_timeout.py
```diff
@@ -1,17 +1,56 @@
...
+
+
+# TODO(AD): This test duplicates TestSourceSessionLayout - remove it?
```
legoktm: Same, let's remove it.

legoktm (changes requested):
Sorry about the delay in reviewing, this looks great! The two places you flagged as being duplicates, let's just remove them as part of this cleanup. Otherwise I think this is ready to go. Thank you for continuing to work on it!

securedrop/tests/functional/test_source_session_timeout.py
```diff
@@ -1,17 +0,0 @@
-from . import source_navigation_steps
```
nabla-c0d3, line 1: Removed because it was duplicating TestSourceSessionLayout.

securedrop/tests/functional/test_source_notfound.py
```diff
@@ -1,10 +0,0 @@
-from . import source_navigation_steps
```
nabla-c0d3, line 1: Removed because it was duplicating TestSourceLayout.test_notfound().

nabla-c0d3:
@legoktm Thanks for looking at this again; I removed the two duplicate tests.

legoktm (approved):
LGTM!

Issue #6211, opened 2022-01-04
SI Redesign: Phase One


ninavizz:
_This is a child-epic to the parent epic #6210_ 

## Introduction
This epic has been assembled to create the _neatest-possible artifact_ to push as a Feature Branch into a first phase of user testing. **Outputs from this Epic will not need immediate L10 work, nor change-management stuff with customers.** One or two customers will need to be sourced to run a branch of this as their SD instance, per the [Phase One Research Plan](https://docs.google.com/document/d/1HC-7JJwnTVmysNFV0KOxRWdf-nGEq56ntYXEiR7iu-o/edit?usp=sharing). 

## Acceptance Criteria
A functional branch created that reflects all of the changes shown in the [Frankenstein E](https://www.figma.com/proto/rHT1RE7zpNttqIGdZpPkhQ/SI_FlowInversion-deux?page-id=572%3A4726&node-id=572%3A5041&viewport=387%2C48%2C0.15&scaling=min-zoom&starting-point-node-id=572%3A5041) prototype, from which the [Phase One Research Plan](https://docs.google.com/document/d/1HC-7JJwnTVmysNFV0KOxRWdf-nGEq56ntYXEiR7iu-o/edit?usp=sharing) was drafted. 

Please go into the corresponding Figma file to look at details such as hex values, padding/positioning particulars, components, etc. If you're new to Figma with SecureDrop, so are all of us—[here's your guide](https://github.com/freedomofpress/securedrop-ux/wiki/Using-Figma)! Because pixel-perfect matching (within reason) is the goal, it is my ask that folks not try to "eyeball" things—and make the effort to look at values in the "Inspect" panel w/in Figma. 🙏🏻  

# Stuff
This section is broken-out into three distinct categories of stuff. Page layouts for the Flow Inversion will be _much_ easier to do, once the earlier of the Design System items have been tackled. Et Cetera are quick-win items neither dependent upon the broader design system, nor inverting the Generate/Submit flow. 
- [Design System](#designsystem)
- [Flow Inversion](#flowinversion)
- [Et Cetera](#etcetera) 

<a name="designsystem"></a>
## Design System
[...]

Issue #4452, opened 2019-05-22
Ruby sass gem used at build time has reached end-of-life


eloquence:
Per https://github.com/sass/ruby-sass (which is archived) and https://sass-lang.com/ruby-sass, this gem, which is [used at package build time](https://github.com/freedomofpress/securedrop/blob/develop/install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/tasks/sass.yml) for CSS generation, has reached end-of-life. We should consider migrating to a maintained codebase, like [sassc](https://github.com/sass/sassc) / `libsass`.

Note that we're also using a pinned outdated version due to #1594, but that no longer seems worth resolving at this point.

eaon:
Since I started working on the SI frontend, I looked into this a bit:

* [sassc](https://github.com/sass/sassc), while available from Debian/Ubuntu/Fedora repos, is now also deprecated but (supposedly) maintained (at time of writing the CI badges show that builds fail :shrug:)
* [DartSass](https://github.com/sass/dart-sass) is the only library/tool that is being actively worked on anymore. It requires either node.js to run the compiled-to-JS version, or Dart to build a native version. Dart itself is not available in Debian/Ubuntu (or Fedora) repos, instead they provide their own repo. They also release binaries on GitHub, but there's no signatures for those.

However, it may turn out that relying on Sass at all may not be necessary - after all, SecureDrop is an unusual webapp that has different targets and goals than most contemporary web apps. I'll look into that first, once I'm closer with #6211

legoktm:
> However, it may turn out that relying on Sass at all may not be necessary - after all, SecureDrop is an unusual webapp that has different targets and goals than most contemporary web apps. I'll look into that first, once I'm closer with https://github.com/freedomofpress/securedrop/issues/6211

@eaon I'm curious how you feel about this now that #6315 has landed. If we're still stuck with sass in the short term, I would like to switch us over to sassc, which is at least maintained in Debian, allowing us to drop all the ruby stuff from our stack.

legoktm:
Briefly discussed with @eaon, we could theoretically get rid of sass but it would require compiling everything to CSS and switching to CSS variables, etc. and I'm not ready to take that all on. I'm first going to see how difficult it would be to drop in sassc, as that would be enough to unblock my Debian packaging work.

legoktm:
[...]

PR #6315, opened 2022-03-03
Source Interface Redesign Omnibus

eaon:
## Status

Ready for review

## Description of Changes

Refactoring towards #6211 - changes in this PR are primarily visual, the vast majority of the interaction stays the same.

* **`source.sass` is dead, long live `source.sass`**
  
  A complete rewrite of `source.sass`, plus quite a bit of the templates as well. So we're separating the source interface and journalist interface (temporarily?) to not break the JI while refreshing the SI according to Nina's specs

* **Refactor locale menu**

  The menu for locale selection has a lot of unused complex crud we don't need to make it look/work the way it does. Less stuff to wrangle: good!

* **Reuse flash message template, supporting "declarative" and icon only messages**

  In the future, every flash message ought to have a declarative heading added to it, this change makes that less cumbersome to add in the future once we found the right language. Also: bring flash message categories closer to SI redesign terminology

* **Change codename widget UI, and keep it in its own template**

  The codename widget should look and work the same in both places where it is being served.

  Regarding the implementation, we're emulating a checkbox visually using `::before` pseudo-elements with a label, while the actual checkbox input state is used to show/hide the codename, necessitating that it precedes the codename and label in HTML. (Similar to the locale menu implementation)

* **Use descriptive/semantic variables for all use of colors**

  Any time a color is used, use a variable that describes the context of its use rather than the color itself. (Distant future possibilities: can help us implement dark mode or - more controversially - allow very simple white label design features beyond just replacing the logo.)

* **Switch out icons**

  Nina designed some icons for us, and changed some others. Material Design based icons live in their own folder with their own README, referencing the license, plus the SVG files have a comment with an SPDX short identifier.

  There's a lot of files which haven't been used in a long while, so this PR includes a first attempt at cleaning those up, but there's more left to do (font-awesome icons are barely used, make a good candidate for getting cut).

  The new icons are monochrome. This isn't primarily a space-saving measure, but it enables color tweaks by using them as mask-images. Similarly, there's a couple of instances in which there's "negative" icons, where a white image foreground sits on top of a CSS background-color.

  SVG files of these icons are versioned in case images need to be replaced, but SVG rendering is disabled in Tor Browser's "Safest" mode, so they're not directly used in the UI.

* **Refactor JS browser warnings**

  * Orfox no longer exists, it's now called Tor Browser for Android
  * Update Tor Browser for Android UA regex
  * Removed dead JS
  * Removed dependency on font-awesome icon
  * Removed animations - there's no other animations in the source interface and adds very little value - if we want animations back, I'd suggest implementing them as much as possible with CSS

* **Tab focus changes**

  Make sure that tab focus is properly visible and works like we want to even on input elements in which what we actually interact with is the label (codename widget & locale menu)

* **Add most recent logo - colors were tweaked!**

* **RTL support**

  While RTL support is already present in SD, some of the details weren't quite right. This should flip all relevant UI elements including the codename widget & locale menu.

* **Revise button order**

  Button order for regular actions is: `[ Escape ] [ Action ]`, where `[ Action ]` is "solid" / attention-grabbing

  For destructive actions this is reversed: `[ Destructive Action ] [ Escape ]`, and `[ Escape ]` is "solid"

### Design

- [x] The "Product Colors" logo [from Figma](https://www.figma.com/file/iEBfB1Xf3j9sq5NQlImGho/Logo?node-id=8%3A5) is used (colors "pop" more than before)

- Using the latest Tor Browser, nothing shows up distorted, overlaps when it obviously should not, or jumps unexpectedly (on hover, focus, or click)
  - Expected jump: If the codename is long enough to be broken up in two lines, the widget will expand to accommodate two lines instead of one
  - [x] Linux
  - [x] Windows
  - [ ] Mac

- Basic liquid layout works
  - Resize browser window for a viewport size smaller than 800px (Tor Browser changes viewport in 100px increments):
    - [x] Index page: "First submission" and "Return visit" on index page sit on top of each other instead of next to each other
    - [x] Everywhere: Logo/locale sits on top of the main content instead of being in a column to the left (right in rtl languages)
    - [x] "Skip to main content" and "Skip to notification" links (that only appear when tabbing through the page) show up above the logo without any content jumping

- [x] The codename widget "weeble" icon is always vertically centered, even when the codename is long enough for a line break

### UX

- [x] Button order for regular actions is: `[ Cancel ] [ Action ]`
- [x] Button order for destructive actions (replies): `[ Destructive Action ] [ Cancel ]`
- [x] When the locale menu is visible, it shows creates a margin under the menu making the shadow of the menu visible even in pages that would otherwise be too short

### Accessibility

- Use the entire workflow of the SI by keyboard only
  - [x] Links can be clicked and forms submitted by pressing return
  - [x] `[ ] Show Codename` widget checkbox can be interacted with by pressing the spacebar
  - [x] "Skip to main content" link appears on all pages (except the index page)
  - [x] "Skip to notification" link appears on pages which show a notification (except the index page)
  - [x] The logo does not jump when the "Skip to …" links appear

- [x] Items in (tab rather than mouse) focus show a 3 pixel big cyan border, text inputs a light blue border
- [x] Buttons "grow" a border on hover

_Screen reader based tests TBD_

### I18n

- Right-to-left languages (currently only Arabic)
  - [x] Flips the column order
  - [x] Codename stays left-to-right because it's currently English-only
  - [x] Shadow of codename widget "leans" to the left
  - [x] Icons (flash messages, weeble in codename widget and login form) are shown on the right
  - [x] Checkbox in the codename widget is shown on the right side of the text
  - [x] Date/time and trash cans for deletion of replies are shown on the left

- Arabic specifics
  - [x] Arabic does not have any letter spacing applied
  - [x] Codename (which uses an English word list) but has letter spacing applied

### Staging

- [ ] Verify that all templates are accessible to Apache
- [ ] Verify that all images loaded by CSS and HTML are accessible to Apache

### Demo site usage

- UI renders the same on Browsers other than Firefox/Tor Browser
  - [x] Chrome/Chromium
  - [ ] Safari
  - [ ] Edge
 
## Checklist

## Known issues

* In Tor Browser/Firefox, zoom at certain sizes (but at least 120% to 150%) slightly distorts the icon of the info box underneath "Read Replies" on `/lookup`

## Comments

The most unusual (and arguably impactful) choice I made in this change (other than to opt for refactoring in the first place) is avoiding adding to HTML whenever possible/reasonable, instead of following [BEM](https://en.bem.info/methodology/quick-start/) or a similar methodology. This intentional, as BEM duplicates a lot of information already implicitly present in HTML, while not providing us with enough benefits for our (web-tech wise rather niche) use case. Onion services bandwidth can be incredibly limited at times, which steers me away from redundant HTML especially because SecureDrop's frontend wouldn't benefit much (at all?) of moving to more discrete components. The main downside is that there are a few uses of complex CSS selectors that aren't particularly friendly to developers that don't do a lot of frontend design work (especially if/when [Sass gets removed as would be my preference](https://github.com/freedomofpress/securedrop/issues/4452#issuecomment-1036444065)), but it is far from overwhelming because the SI doesn't have that many different screens to begin with.


eloquence:
Took a quick first spin, coming together really nicely! A few quick notes from my end:
- I would prefer to keep the logo tweaks out of scope for now; we didn't discuss it at length, and we'd have to update in multiple places and ensure it's available for third party use. 
- The codename widget looks great. I'll note that it does wrap codenames a lot more frequently than the old one. When it wraps it takes up a lot of space, and might also induce more copy/paste errors.
- The codename, "show passphrase" widget and "Continue" button are currently stacked very closely together.  Figma has a bit more whitespace at the bottom; I'd also appreciate @tina-ux's eye on this.
Screenshot from 2022-04-07 11-52-51

eaon:
* Logo: :+1: you got it
* Codename widget wrapping: Yes, agreed. I was mulling over this a bit, in part it's a side effect of slightly different/more consistent spacing of the layout, but that bothered me too. Will revisit/adjust.
* "Show [Codename|Passphrase]" / Continue button spacing: totally a late-rebase oversight on my part, will take care of that too

eaon:
Feedback incorporated

securedrop/source_app/main.py
```diff
@@ -47,11 +46,10 @@ def generate() -> Union[str, werkzeug.Response]:
         elif tor2web_check != 'href="fake.onion"':
             return redirect(url_for('info.tor2web_warning'))
         if SessionManager.is_user_logged_in(db_session=db.session):
-            flash(gettext(
+            flash_msg(
```
legoktm: Did you test if this pattern works with the pybabel extractor? IIRC it scans the Python source code looking for `gettext()` function calls.

securedrop/source_templates/index.html
```diff
@@ -41,72 +39,47 @@ <h2 id="security-level-heading" hidden>{{ gettext('How to change your security l
...
-        <header class="header">
-          <h1><img src="{{ g.logo }}" alt="{{ gettext('{} logo'.format(g.organization_name)) }}" class="logo-image">
-          </h1>
-          <div id="index-locales">
-            {% include 'locales.html' %}
-          </div>
-        </header>
+    {% include 'flashed.html' %}
+    <header>
+      <h1><img src="{{ g.logo }}" alt="{{ gettext('{} logo'.format(g.organization_name)) }}"></h1>
```
legoktm: .format needs to go outside gettext()

securedrop/source_templates/lookup.html
```diff
@@ -32,76 +25,75 @@ <h2 id="submit-heading" class="headline">{{ gettext('Submit Messages') }}</h2>
...
-  {% else %}
-      {% if min_len > 0 %}
-        <p class="center" id="min-msg-length">{{ gettext('Your first message must be at least {} characters long.').format(min_len) }}</p>
-      {% endif %}
-  {% endif %}
+      <div class="message">
+        {% if min_len > 0 %}
+        {{ form.msg(**{"aria-describedby": "min-char-length"}) }}
+        {% if allow_document_uploads %}
+        <p id="min-char-length">{{ gettext('If you are only sending a message, it must be at least {} characters long.'.format(min_len)) }}</p>
```
legoktm: format needs to be outside gettext

securedrop/source_templates/lookup.html
```diff
@@ -32,76 +25,75 @@ <h2 id="submit-heading" class="headline">{{ gettext('Submit Messages') }}</h2>
...
+        {% else %}
+        <p id="min-char-length">{{ gettext('Your first message must be at least {} characters long.'.format(min_len)) }}</p>
```
legoktm: format needs to be outside gettext

securedrop/sass/_source_vars.sass
```diff
@@ -0,0 +1,166 @@
+// Colors for "local" referencing.
+// Please use descriptive variables stylesheets
+
+// Blue-ish
```
legoktm, line 4: I think this might be a pre-existing problem, but these variable names/colors are kind of useless to me because I would have no clue when to use cerulean vs cyan vs grimace blue, you get the idea.

I like what the Wikimedia Style Guide does with Red90, Red50, Red30, etc.: https://design.wikimedia.org/style-guide/visual-style_colors.html and https://design.wikimedia.org/style-guide/resources/WikimediaUI_color_palette.svg

legoktm (changes requested):
Overall testing went well, no issues with functionality. I left inline comments about possible localization issues.

It somewhat bothers me that the shape/size redacted version of the codename doesn't actually match the underlying codename. I don't know how hard it would be to implement that though.

securedrop/source_app/main.py
legoktm: Specifically see https://github.com/python-babel/babel/blob/61109c4bd4d60a79d7cb5f216a8286ddff6fb30b/babel/messages/extract.py#L31 - I think we'd need a custom extractor if we want to support more function names.

securedrop/source_templates/index.html
```diff
@@ -41,72 +39,47 @@ <h2 id="security-level-heading" hidden>{{ gettext('How to change your security l
     <p>
       {{ gettext('<a href="/" aria-label="Follow these instructions, then refresh this page">Refresh this page</a>, and you\'re done!') }}
     </p>
-
   </dialog>
-  <div id="use-tor-browser" class="warning" role="alert">
+  <div id="browser-tb" class="warning" role="alert">
     {{ gettext('<strong>It is recommended to use Tor Browser to access SecureDrop:</strong> <a class="recommend-tor" href="{tor_browser_url}">Learn how to install it</a>, or ignore this warning to continue.').format(tor_browser_url=url_for('info.recommend_tor_browser')) }}
```
zenmonkeykstop, line 44: Uses default `a` link colors instead of original white (for both visited and unvisited). This doesn't appear to be covered in design mockups, but the combo of blue and purple is pretty low-contrast, the original white color works better imo.

securedrop/sass/_source_vars.sass
zenmonkeykstop: My kneejerk reaction to this was "absolutely, and functional descriptions would be even better", but looking ahead in _source_vars.sass, the actual variables using these colors are pretty well-named, and those vars are what actually get used. (I don't know if this modifies your postion at all).

securedrop/source_templates/index.html
eaon: Hmm, what browser does this happen with for you? Both Chromium and Firefox show the link here as white and underlined for me

zenmonkeykstop: Yikes, Firefox 78esr apparently, will update the test environment  and try it again.

zenmonkeykstop: Yupp, fine in 91esr, my bad.

securedrop/source_app/main.py
eaon: You're totally correct, I was overly optimistic/absent minded about pybabel. Changed the helper function and returned to the `gettext()` calls

securedrop/sass/_source_vars.sass
eaon: Yep, the color names are partially inherited, some are taken from Figma, some are new because they haven't had a name assigned but popped up often enough (for what it's worth it might be cool to introduce them in Figma as well but I'm better at reading Figma than changing stuff on there). `source.sass` should never reference those colors directly (which is what "local referencing" is intended to convey - maybe I should expand the comment?)

My general preference would be that developers wouldn't make the (final) choice for what color is used where. Ideally @tina-ux would be involved in the process before implementation happens, but even if that's not possible, a descriptive variable name should be used in `source.sass` and the final color can be adjusted later on. Does that make sense?

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
```diff
@@ -222,22 +222,22 @@
...
   /var/www/securedrop/source_templates/footer.html r,
   /var/www/securedrop/source_templates/generate.html r,
   /var/www/securedrop/source_templates/index.html r,
   /var/www/securedrop/source_templates/locales.html r,
   /var/www/securedrop/source_templates/login.html r,
   /var/www/securedrop/source_templates/logout.html r,
   /var/www/securedrop/source_templates/lookup.html r,
-  /var/www/securedrop/source_templates/next_submission_flashed_message.html r,
   /var/www/securedrop/source_templates/notfound.html r,
   /var/www/securedrop/source_templates/session_timeout.html r,
```
zenmonkeykstop: file was deleted, not needed in apparmor profile anymore. (also journalist_templates/delete.html was removed in 2.3.0, tho strictly speaking not in scope for this PR)

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
```diff
@@ -222,22 +222,22 @@
...
   /var/www/securedrop/source_templates/tor2web-warning.html r,
-  /var/www/securedrop/source_templates/utils.html r,
   /var/www/securedrop/source_templates/use-tor-browser.html r,
+  /var/www/securedrop/source_templates/utils.html r,
   /var/www/securedrop/source_templates/why-public-key.html r,
   /var/www/securedrop/static/.webassets-cache/** rw,
   /var/www/securedrop/static/css/font-awesome.css r,
```
zenmonkeykstop: not introduced in this release, but looks like this and static/css/normalize.css has been gone for a long time!

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
```diff
@@ -263,16 +263,19 @@
   /var/www/securedrop/static/gen/journalist.js rw,
   /var/www/securedrop/static/gen/source.css rw,
```
zenmonkeykstop: cssmin was removed in #4227 so these are no longer present.

zenmonkeykstop (changes requested):
Stepped through source flow in a staging environment, looks great, notes as follows:
- no issues with apparmor denies, but comparing the profile to what's installed shows some deleted resources still listed (mostly not introduced by this PR)
- (personal pref) the explanatory reply text is pretty tiny for me on Debian 11 and TBB, I don't see the rationale in having it be 3 points smaller than other page text. 

With fixes requested by other reviewers and tweaks to the Apache apparmor profile, this looks good to go!

eaon:
> the explanatory reply text is pretty tiny for me on Debian 11 and TBB

As far as I remember, there's two parts to this, though the second one I recalled from memory so take that with a grain of salt:
* [Upcoming language changes](https://www.figma.com/file/rHT1RE7zpNttqIGdZpPkhQ/SI_FlowInversion-deux?node-id=1003%3A7430) - keeping the mental load for sources low by providing them with clearer but simpler instructions
* [If you compare it with the full-redesign mockup](https://www.figma.com/file/rHT1RE7zpNttqIGdZpPkhQ/SI_FlowInversion-deux?node-id=403%3A3974) (warning: outdated language there), the size feels more appropriate IMO - that too is (IIRC) a form of lowering the mental load but with design instead - while still highlighting good practice, we're toning down the scary stuff.

@tina-ux what do you think, as a sort of in-between stage, is it more appropriate to adapt font-sizes a tad?

As for your requested changes, I updated the AppArmor profile accordingly :slightly_smiling_face:

eaon:
Checked in with Tina and decided to increase the font size of the reply info note by 2 pixels!

zenmonkeykstop (approved):

zenmonkeykstop:
Thanks @eaon and @tina-ux, apparmor and UX changes look good!

legoktm (approved):
I only tested in dev and not staging, but everything LGTM otherwise!

PR #6321, opened 2022-03-07
Translate string found in SI replies

### LINKED ISSUE #6279: Untranslated string in SI reply interface
Just stumbled over [a visually hidden header](https://github.com/freedomofpress/securedrop/blob/4a80f70901f49dfe2f30b79d2a9bbc69cf807cec/securedrop/source_templates/lookup.html#L83) in the reply section that is missing from translations. Because it's hidden it only affects screen readers.

A quick look at the `message.pot` history makes me believe this is a new string that we missed during review.

eaon:
## Status

Ready for review

## Description of Changes

Fixes #6279.

We recently caught an untranslated string in the source interface. It involves a reference to a relative timestamp, so instead of duplicating code, I added a macro and use its output on two different occasions.

## Testing

* Either add a new submission and reply to it, or log in with a source that has pre-filled replies 
* Change language to something other than English
* Use **Web Developer Tools** to inspect the DOM
* Navigate to (in CSS notation) `.reply dialog h3`
* Inside that `h3` you should find an English *Delete reply from*, but a translated timestamp in a `time` tag

## Checklist

### DIFF
--- a/securedrop/source_templates/lookup.html
+++ b/securedrop/source_templates/lookup.html
@@ -1,4 +1,5 @@
 {% extends "base.html" %}
+{% import 'utils.html' as utils %}
 
 {% block body %}
 {% if new_user_codename %}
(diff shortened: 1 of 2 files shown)


legoktm (approved):
Test plan checks out, I like the use of the macro to keep the HTML markup out of the translation.

PR #6322, opened 2022-03-07
Center text in reply deletion confirmation prompt

eaon:
## Status

Ready for review

## Description of Changes

All text for the replies was set to align to left, but the confirm-prompt ought to be centered as it looks glitched otherwise.

**Before**
confirm-prompt-align-left

**After**
confirm-prompt-align-center

## Testing

* Either add a new submission and reply to it, or log in with a source that has pre-filled replies
* Use trash can icon to delete a reply
* Confirm that the prompt looks like the second screenshot rather than the first

## Checklist

### DIFF
--- a/securedrop/sass/source.sass
+++ b/securedrop/sass/source.sass
@@ -203,6 +203,7 @@ p#codename
     background: white
     border: 1px solid $color_grey_dark
     padding: 0
+    text-align: center
 
     p
       *


legoktm:
Definitely looks better, happy to merge on that alone. In English, and probably other languages with short text, it all is stuck onto one line - do we want to force the "DELETE" button onto a separate line unconditionally?

Screenshot 2022-03-09 at 15-43-10 SecureDrop Protecting Journalists and Sources

zenmonkeykstop (approved):
LGTM - we can defer other layout questions to design work on the SI in the next release.

Issue #6311, opened 2022-03-01
Update new "Get Codename" page in line with design recommendations


eloquence:
As part of the Source Interface design improvements (see #6210), we have broad consensus around language changes to the "codename" pages which could be made regardless of whether or not we ultimately fully reverse the password generation flow. This is ["E1" in Figma](https://www.figma.com/file/rHT1RE7zpNttqIGdZpPkhQ/SI_FlowInversion-deux?node-id=915%3A5608):
![Screenshot from 2022-02-28 16-50-53](https://user-images.githubusercontent.com/213636/156083829-9e10ef5e-5032-4f35-a931-c84a984e4352.png)

The wording changes, which were collaboratively workshopped at a couple of UX meetings, in summary:
- reduce text on the page to the essentials, clearly summarized
- avoid the potentially confusing phrase "using this codename will be the only way we have to communicate with you"
- change button to "Continue" instead of "Submit Documents" (avoid a button label that refers to a task that actually happens on the _next_ step)

This does _not_ change the word "codename" to "passphrase" yet, because a) that change needs to be carefully rolled out, given that the term has been used from the very beginning, b) we'll want to user-test the term "passphrase" before fully committing to that change, so we don't have to potentially change it twice.

The styling changes to the codename widget are the same as already implemented in @eaon's branch for #6211. The show/hide widget is also the same as implemented there (but the default state on this page would be "show", not "hide").

eloquence:
(Retitled issue since the language updates have landed now.)

eaon:
Fixed in #6315

Issue #5864, opened 2021-03-11
securedrop-admin integration tests fail during "tag is live, release is not" time window


eloquence:
Discovered during 1.8.0 release (#5794): If a tag has been pushed but the release object has not been created yet, two `securedrop-admin` tests start failing:

- `test_check_for_update_when_updates_not_needed`
- `test_update_with_duplicate_branch_and_tag`

Those tests operate against the production repo, and they rely on the latest release and the latest tag being identical. We should document this behavior in the RM guide, or make the tests more resilient. See https://github.com/freedomofpress/securedrop/pull/5862#issuecomment-797071274 for test output.

(We confirmed that this was indeed the cause by re-running CI after the release object was created.)

eloquence:
Discussed with @emkll at backlog review, we'll probably just document this behavior for now.

PR #6330, opened 2022-03-09
Rewrite codename explanation to be simpler, clearer

eloquence:
The previous language is ambiguous and overly verbose. Simpler language was workshopped in a cross-functional team workshop and is implemented in this PR, see #6311. This does not yet implement all the agreed upon design changes but gets us one step closer.

## Status

Ready for review

## Description of Changes

Towards #6311

## Testing

1. Visit the `/generate` page in the development environment
2. Confirm that there is no visual regression
3. (Optional) Confirm that there is no regression for users of assistive technologies (or review the code to confirm same)

## Deployment

No special considerations. We will need to update screenshots in the documentation, as we generally do with any minor-level release.

## Checklist

- [x] I would appreciate help with the documentation (see above)

### DIFF
--- a/securedrop/source_templates/generate.html
+++ b/securedrop/source_templates/generate.html
@@ -16,12 +16,10 @@
     aria-describedby="codename-instructions codename-explanation">{{ codename }}</mark>
 </section>
 
-<p class="explanation">
-  <ul class="explanation">
-    <li>{{ gettext('<strong>Keep it secret.</strong> Do not share it with anyone.') }}</li>
-    <li>{{ gettext('<strong>Keep it safe.</strong> There is no account recovery option.') }}</li>
-  </ul>
-</p>
+<ul class="explanation">
+  <li>{{ gettext('<strong>Keep it secret.</strong> Do not share it with anyone.') }}</li>
+  <li>{{ gettext('<strong>Keep it safe.</strong> There is no account recovery option.') }}</li>
+</ul>
 
 <form id="create-form" method="post" action="/create" autocomplete="off">
   <input name="csrf_token" type="hidden" value="{{ csrf_token() }}">
--- a/securedrop/source_templates/generate.html
+++ b/securedrop/source_templates/generate.html
@@ -1,13 +1,10 @@
 {% extends "base.html" %}
 
 {% block body %}
-<h1>{{ gettext('Welcome') }}</h1>
-
-<p id="codename-instructions" class="aside center">
-  {{ gettext('Please either write this codename down and keep it in a safe place, or memorize it.') }}</p>
+<h1>{{ gettext('Get Your Codename') }}</h1>
 
 <p id="codename-explanation" class="explanation">
-  {{ gettext('This codename is what you will use in future visits to receive messages from our team in response to what you submit on the next screen.') }}
+  {{ gettext('A <em>codename</em> in SecureDrop functions as both your username and your password:') }}
 </p>
 
 <section class="code" aria-labelledby="codename-heading">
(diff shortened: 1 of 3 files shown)


securedrop/source_templates/generate.html
```diff
@@ -17,17 +14,19 @@ <h2 id="codename-heading" class="visually-hidden">{{ gettext('Codename') }}</h2>
 </section>
 
 <p class="explanation">
-  {{ gettext('Because we do not track users of our <b>SecureDrop</b>
- service, in future visits, using this codename will be the only way we have to communicate with you should we have
- questions or are interested in additional information. Unlike passwords, there is no way to retrieve a lost codename.') }}
+  {{ gettext('You will need this codename to log into our SecureDrop later.') }}
```
eloquence: This is not exactly as workshopped, but per chat with Kev, "access" seemed to retain some ambiguity, and we do use the term "log in" on the index page, so it seemed appropriate to use it here as well.

zenmonkeykstop: Couple of potential tweaks:

1) the explanation text (first two lines) and the ordered list have different text colors - it's a little jarring. It might be better for the text to be styled consistently.
2) Looking at it in the page, it doesn't actually explicitly state that the codename displayed is the user's. It might be better to put both explanatory lines over the codename like so:

```
A codename in SecureDrop functions as both your username and your password.
You will need this codename to log into our SecureDrop later:

    [ lock] purple monkey dishwasher oil horse etc.

1. Keep it secret. Do not share it with anyone.
2. Keep it safe. There is no account recovery option.
```

zenmonkeykstop: UX tweak: move line up over codename widget?

securedrop/source_templates/generate.html
```diff
@@ -17,17 +14,19 @@ <h2 id="codename-heading" class="visually-hidden">{{ gettext('Codename') }}</h2>
...
+  <ol>
```
zenmonkeykstop: UX tweak: apply consistent styling to text (my pref would be to use the styling on the list for the explanation as well, rather than vice versa, but either would be better).

zenmonkeykstop (changes requested):

securedrop/source_templates/generate.html
eloquence: I agree. I've reorganized accordingly, and also made the list unnumbered.

securedrop/source_templates/generate.html
eloquence: I've for now used the existing dark grey styling throughout, since it's used in other places as well; that will be overhauled as part of Michael's PR.

securedrop/source_templates/generate.html
```diff
@@ -17,17 +17,18 @@ <h2 id="codename-heading" class="visually-hidden">{{ gettext('Codename') }}</h2>
...
 
 <p class="explanation">
-  {{ gettext('Because we do not track users of our <b>SecureDrop</b>
- service, in future visits, using this codename will be the only way we have to communicate with you should we have
- questions or are interested in additional information. Unlike passwords, there is no way to retrieve a lost codename.') }}
+  <ul class="explanation">
+    <li>{{ gettext('<strong>Keep it secret.</strong> Do not share it with anyone.') }}</li>
+    <li>{{ gettext('<strong>Keep it safe.</strong> There is no account recovery option.') }}</li>
+  </ul>
 </p>
```
eaon: 2 changes I'd suggest here:

* Remove the paragraph as a wrapper - `ul`s aren't allowed to be inside paragraphs, all that happens is the browser adding invisible empty paragraphs before and after the unordered list
* Optional: change the `<strong>` tags to `<b>` - see also: https://github.com/freedomofpress/securedrop/commit/700adab63a0b50c8ebef7a95bc3ee87a6dfb051e - but maybe that's just me being not that good with prioritizing text

eaon (changes requested):

securedrop/source_templates/generate.html
eloquence: Thanks @eaon. I've removed the wrapper; I've kept the `<strong>` tag as it does seem to me that we'd want a screen-reader to emphasize those portions distinctly (it's not just a styling difference), especially the unambiguous "Keep it secret".

eaon (approved):
LGTM

zenmonkeykstop (approved):

eloquence:
Test failure appears to be #5864 because 2.2.1 tag is now live but release is not published yet.

Issue #6328, opened 2022-03-09
Allow bumping kernel version without requiring a new SecureDrop release


legoktm:
## Description

It should be possible for us to upgrade the Linux kernel version without requiring a new SecureDrop release.

In https://github.com/freedomofpress/securedrop/issues/6323 we wanted to bump just the kernel for Dirty Pipe, but needed to go through a full SD release cycle for it. Besides the extra unnecessary work, it also causes more churn for SD servers, since they have to upgrade securedrop-app-code packages that are identical except the version number.

legoktm:
There are two main references to the current kernel version in this git repository.

First the actual linux package we use is named `linux-image-5.15.26-grsec-securedrop`. This is already separate and "maintained" in the kernel-builder repository.

Then we have a `securedrop-grsec` package, with a version of 5.15.26+focal, which depends on `linux-image-5.15.26-grsec-securedrop,intel-microcode`.

We use the actual kernel version in places like testinfra/common/test_grsecurity.py, which verifies the exact kernel version that is installed.

I propose we move the securedrop-grsec packaging to a separate repository, and in places like test_grsecurity.py, we only verify the major+minor version of the installed kernel, so 5.15 in this case. Bumping patch versions can be done independently, requiring just new securedrop-grsec and linux-image-* packages, while a minor or major kernel bump will require a full SD release.

zenmonkeykstop:
:+1:  from me, tis a fine plan.

conorsch:
>  with a version of 5.15.26+focal

Ideally, we'd have a version for the metapackage that's independent of the underlying kernel image. But the general outlines of the plan are great, and would indeed free us up a bit. I also support adding the kernel version to the `/metadata` endpoint so that we can still monitor for successful kernel updates across prod instances. The kernel version is already public information, tied to the app code version, so no change to information shared.

zenmonkeykstop:
It's possible for instances to downgrade to other grsec kernels, so we don't actually currently know what they're running, we're just inferring it with a reasonable degree of certainty. But overall yeah it's worth the additional information leak. We might also want to consider adding Tor versions to that if we're committing to updating that independently.

zenmonkeykstop:
(Dang now I'm wondering should we make the hypothetical /version endpoint authenticated, and let instances opt in to us monitoring it by sending us a token or something.)

legoktm:
@conorsch wrote:
> > with a version of 5.15.26+focal
> 
[...]

PR #6332, opened 2022-03-10
Expose Linux kernel and Tor version in /metadata endpoint

legoktm:
## Status

Code is ready for review, feature itself needs discussion

## Description of Changes

As discussed in #6328, we want to separate the kernel and tor version
upgrade process from SecureDrop releases. Currently we (and anyone else)
can imply the kernel + tor versions from the SecureDrop version, but
once they are separated, we'll lose that ability.

Explicitly add those version numbers to the /metadata endpoint so we can
still see/verify that the kernel or tor upgrade happened successfully
from our monitoring.

Getting the kernel version is trivial since Python has a `os.uname()`
function, while the tor version requires shelling out to `tor --version`
and processing the output. Since that's more complicated, it will return
"unknown" rather than breaking the /metadata endpoint if an exception
occurs.

## Testing

* Open the SI and go to `/metadata`, see keys for "kernel_version" and "tor_version"

## Deployment

**Yes**, this could have some security implications which are being discussed at #6328 and elsewhere.

## Checklist

### DIFF
--- a/securedrop/server_os.py
+++ b/securedrop/server_os.py
@@ -1,4 +1,5 @@
 import functools
+import re
 import subprocess
 
 FOCAL_VERSION = "20.04"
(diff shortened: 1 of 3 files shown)


securedrop/server_os.py
```diff
@@ -12,3 +13,25 @@ def get_os_release() -> str:
         stdout=subprocess.PIPE,
         universal_newlines=True
     ).stdout.strip()
+
+
+@functools.lru_cache()
+def get_tor_version() -> str:
+    try:
+        output = subprocess.run(
+            ["/usr/bin/tor", "--version"],
```
zenmonkeykstop, line 22: Is this gonna require an apparmor ruleset change? (dev doesn't have it so staging test required)

zenmonkeykstop:
Given the potential security implications and ongoing chats elsewhere, would you be ok with flipping this to draft for now?

legoktm:
We discussed this today and came to a consensus that we do not want to explicitly expose Tor/kernel versions publicly, rather we will advertise in the Journalist Interface if things are out of date. I will file a separate ticket that discusses this in more detail.

PR #6336, opened 2022-03-10
Prevent viewport jumps when there's flashed messages

### LINKED ISSUE #6333: SI: focusing on #flashed message after submitting sometimes scrolls the viewport too far, may obscure message
## Description

First identified by @eloquence in #6306, the message we want to draw attention to may get scrolled out of view when we use `#flashed` as the anchor so screen-readers would know where to put their attention to first.

## Steps to Reproduce

* Visit index with Tor Browser, JavaScript disabled
* Inspect the `html` element with the **Web Developer Tools**
* Using the layout tab, make sure that the content size of the `html` element is 1000x540 (when no scrollbar is visible)
* Add a new submission, see most of the _Success!_ flashed message disappear

## Expected Behavior

Flashed message stays in view

## Actual Behavior

flash

_Originally posted by @eloquence in https://github.com/freedomofpress/securedrop/issues/6306#issuecomment-1063580054_

## Comments

@cfm highlighted that this is probably due to changes from #6041, and IMO we don't want to leave screen-reader users to figure out themselves what's going on. The page 53 of [WCAG 2.1 Audit](https://docs.google.com/document/d/1jZlKco2YgOrrI4PmTkG_Q8lKuV5VxCds/edit) already describes an option that we can probably handle in time for 2.3.0:

We already have a _Skip to main content_ link which we could either combine with or add to with another link specifically to jump to `#flashed` if `get_flashed_messages()` isn't empty.

eaon:
## Status

Ready for review

## Description of Changes

Jumping directly to the `#flashed` anchor on redirect introduced unintended browser behaviour that pushed the messages we wanted users to focus on out of view. Since this was in part an accessibility feature, we don't want to lose the ability for screen-reader users to jump directly to what's important immediately, so this introduces a new "Skip to notification" link, which like the "Skip to main content" link is visually hidden until in focus.

Fixes #6333.

## Testing

* Visit index with Tor Browser, JavaScript disabled
* Inspect the `html` element with the **Web Developer Tools**
* Using the layout tab, make sure that the content size of the `html` element is 1000x540 (when no scrollbar is visible)
* Add a new submission
* Browser does **not** scroll to _Success!_ flashed message
* While _Success!_ or other flashed message is visible, us tab to navigate site
* _Skip to notification_ becomes visible above the logo first, _Skip to main content_ becomes visible second

## Checklist

### DIFF
--- a/securedrop/source_app/main.py
+++ b/securedrop/source_app/main.py
@@ -171,7 +171,7 @@ def make_blueprint(config: SDConfig) -> Blueprint:
             for field, errors in form.errors.items():
                 for error in errors:
                     flash(error, "error")
-            return redirect(f"{url_for('main.lookup')}#flashed")
+            return redirect(url_for('main.lookup'))
 
         msg = request.form['msg']
         fh = None
@@ -186,7 +186,7 @@ def make_blueprint(config: SDConfig) -> Blueprint:
                       "error")
             else:
                 flash(gettext("You must enter a message."), "error")
-            return redirect(f"{url_for('main.lookup')}#flashed")
+            return redirect(url_for('main.lookup'))
 
         fnames = []
         logged_in_source_in_db = logged_in_source.get_db_record()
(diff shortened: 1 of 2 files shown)


eloquence:
(Taking a first look.)

eloquence (approved):
Looks good to me functionally:
- Success and error notices display within viewport
-  "Skip" link is available when tabbing backwards and works as expected

Final code review by a maintainer pending.

cfm (approved):
With @eloquence's functional review, diff looks good to me on visual review.  Merging with thanks, @eaon!

PR #6345, opened 2022-03-16
updated message filter fields to be non-nullable, set default values to disable the filters

### LINKED ISSUE #6342: [2.3.0] `/lookup` fails to render on fresh install due to `min_length` value in template being set to None
## Description

On a fresh install, the initial_message_min_len field in instance_config is NULL, causing checks for equality with integers to fail.

## Steps to Reproduce

- Install 2.3.0-rc1 packages in production or staging
- visit the SI without updating any config settings in the JI

## Expected Behavior

- SI renders correctly

## Actual Behavior

500 errors, with errors like "TypeError: '>' not supported between instances of 'NoneType' and 'int'" in the source logs, if enabled.

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Fixes #6342.

Updates the  db model and alembic migration script to set default values for the fields added in `instance_config`

## Testing

- check out this branch
- start a dev container with `cd securedrop && .bin/dev-shell`
- set up the previous version of the DB with the following commands in said container:
  ```
  source bin/dev-deps
  maybe_create_config_py
  sudo mkdir -p /var/lib/securedrop/
  sudo chmod 0700 /var/lib/securedrop/
  sudo chown "$(id -u):$(id -g)" /var/lib/securedrop
  rm -f /var/lib/securedrop/db.sqlite
  sqlite3 /var/lib/securedrop/db.sqlite .databases > /dev/null
  alembic upgrade 2e24fc7536e8
  ```
- [x] verify that the schema looks correct with `echo '.schema instance_config' |  sqlite3 /var/lib/securedrop/db.sqlite`:
  ```
   CREATE TABLE instance_config (
	version INTEGER NOT NULL, 
	valid_until DATETIME, 
	allow_document_uploads BOOLEAN, organization_name VARCHAR(255), 
	PRIMARY KEY (version), 
	UNIQUE (valid_until), 
	CHECK (allow_document_uploads IN (0, 1))
   );
   CREATE UNIQUE INDEX ix_one_active_instance_config ON instance_config (valid_until IS NULL) WHERE valid_until IS NULL;
  ```
- upgrade the database fully with `alembic upgrade head`
- [x] verify that the schema looks correct with `echo '.schema instance_config' |  sqlite3 /var/lib/securedrop/db.sqlite`:
  ```
  CREATE TABLE IF NOT EXISTS "instance_config" (
	version INTEGER NOT NULL, 
	valid_until DATETIME NOT NULL, 
	allow_document_uploads BOOLEAN, 
	organization_name VARCHAR(255), 
	initial_message_min_len INTEGER DEFAULT '0' NOT NULL, 
	reject_message_with_codename BOOLEAN DEFAULT '0' NOT NULL, 
	PRIMARY KEY (version), 
	CHECK (allow_document_uploads IN (0, 1)), 
	CHECK (reject_message_with_codename IN (0, 1)), 
	CHECK (reject_message_with_codename IN (0, 1)), 
	UNIQUE (valid_until)
  );
  ```
- [x] verify that the default values are in fact set, with `echo 'select * from instance_config;' | sqlite3 /var/lib/securedrop/db.sqlite`

## Deployment
 - migration will be applied as part of regular package update
 - note that an upgrade from 2.3.0-rc1 to an RC with these changes (eg 2.3.0-rc2) will fail, upgrades should be tested like 2.2.1 -> 2.3.0-rc2


legoktm (approved):
This is the first time we've used `server_default`, but using it makes the migration much simpler since we can rely on SQLite to fill in missing values and make the field non-nullable. The test plan checks out, and CI also verifies each migration individually too.

Also as noted this will break upgrading from 2.3.0-rc1 to rc2, however we don't care about that upgrade path.

Issue #6346, opened 2022-03-16
Add special mypy plugin/support for SQLAlchemy


legoktm:
## Description

I did not verify this, but I suspect that #6342 could've been avoided if mypy was able to understand that `nullable=True` maps to `Optional[...]` in Python types and complained about not checking for `is not None` first.

In the past I've used https://github.com/dropbox/sqlalchemy-stubs for this purpose. SQLAlchemy also has its own mypy plugin based off the Dropbox one, https://docs.sqlalchemy.org/en/14/orm/extensions/mypy.html, but I haven't personally used it.

We should investigate one or both of these and see if it improves what mypy can detect in terms of type issues.

legoktm:
This was actually attempted in https://github.com/freedomofpress/securedrop/pull/5589, but missed the step of adding the plugin to mypy, so it didn't have much of an impact :(

When I enabled it locally (and upgraded mypy), the main thing it caught is that `Source.filesystem_id` is technically nullable, even though in practice it never is. Still have some more to clean up before I push a PR.

Issue #6226, opened 2022-01-14
Various database columns should be NOT NULL


legoktm:
## Description

There are a decent amount of columns in our models that are nullable just because the default is `nullable=True`, and not because they should be. Some will be very easy to migrate (booleans, where null is coerced to false), others might be more involved, and then there were some I wasn't 100% sure about.

## Easy (booleans)
* sources.pending
* source_stars.starred
* journalists.is_admin
* instance_config.allow_document_uploads

## Might need a data migration or checks
* submissions.source_id (fk to sources.id)
* replies.source_id (fk to sources.id)
* source_stars.source_id (fk to sources.id)
* journalists.created_on (defaults to current time)
* journalist_login_attempt.timestamp (defaults to current time) 

## Unclear
* sources.filesystem_id (seems required)
* journalist.otp_secret,is_totp,hotp_counter

## Comments

We should end up in a state where every column explicitly has `nullable=True` or `nullable=False` so it's clear the nullability of every column has been considered rather than using the default.

legoktm:
I will be making sources.filesystem_id NOT NULL as part of #6346.

PR #6350, opened 2022-03-18
Make sources.filesystem_id non-nullable

legoktm:
## Status

Ready for review

## Description of Changes

We already set this field all the time and in all places expect it to be
a string, not checking if its None first.

Discovered by the mypy work I'm doing in #6346.

Refs #6226.

## Testing

* [ ] Before applying this patch, create a source, log in with it. Then check out this branch, in the dev container run `alembic upgrade head` to apply this schema change, and try to log in with the source. Verify you can log in properly.
* [ ] CI passes

## Deployment

Any special considerations for deployment? No, just a normal schema change.

## Checklist

### DIFF
--- a/securedrop/alembic/versions/b7f98cfd6a70_make_filesystem_id_non_nullable.py
+++ b/securedrop/alembic/versions/b7f98cfd6a70_make_filesystem_id_non_nullable.py
@@ -1,0 +1,58 @@
+"""make_filesystem_id_non_nullable
+
+Revision ID: b7f98cfd6a70
+Revises: d9d36b6f4d1e
+Create Date: 2022-03-18 18:10:27.842201
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'b7f98cfd6a70'
+down_revision = 'd9d36b6f4d1e'
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Not having a filesystem_id makes the source useless, so if any of those do exist, we'll
+    # delete them first, as part of this migration.
+    # Because we can't rely on SQLAlchemy's cascade deletion, we have to do it manually.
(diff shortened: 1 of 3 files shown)


legoktm:
I tried to git blame on whether we were ever intentionally writing NULLs to filesystem_id and didn't really find anything (but also the filesystem_id column dates back to the initial SQLAlchemy commit, so it's super old!).

I am relatively confident that there aren't NULLs lying around because SD would blow up pretty quickly in the app code if filesystem_id was NULL, so moving this to ready for review.

zenmonkeykstop:
Poked around a bit as well (including manually setting a source's filesystem_id to null and seeing what happens) - returning users are authenticated by generating the expected filesystem_id  and querying the database for it. So, if a source's filesystem_id field is ever set to null, they can't log back in.

Probably what this means for this PR is that just deleting sources with a null filesystem_id value is a valid migration strategy, as their submissions are already effectively orphaned and will get swept up in the nightly jobs to detect disconnected files/entries.

legoktm:
Deleting a source is not the easiest thing in alembic because we need to clean up referential tables, but it seems doable.

legoktm:
Moving back to draft state, I pushed my work from today but there's still a bit more to do in the migration, we delete stuff out of sources, replies, submissions, but aren't then deleting stuff that refer to those replies and submissions...

zenmonkeykstop (approved):
Tested as follows:
- set up 2.3.1 prod VM environment, created multiple sources, nulled out the filesystem ID for one of them. 
- [x] Confirmed that a 500 error is thrown on the All Sources page due to the missing filesystem_id (so this unlikely scenario would have been immediately noticeable)
- used the upgrade scenario procedure to update the prod instance with debs built from this branch
- [x] confirmed that the All Sources page no longer 500s out and displays only the non-nulled sources.
- [x] confirmed that replies and message from those sources are still available for download
- [x] confirmed that the nulled source was deleted from the database
- [x] confirmed that the filesystem_id field has nullable=false applied.

LGTM!

PR #6351, opened 2022-03-18
Upgrade mypy and enable SQLAlchemy plugin

### LINKED ISSUE #6346: Add special mypy plugin/support for SQLAlchemy

### LINKED ISSUE #6358: Install securedrop-app-code dependencies in development environment via `make venv`
## Description

We should install the `securedrop-app-code-requirements.txt` dependencies when setting up the development environment via `make venv`.

The main motivation for this is to make mypy work better out of the box. Currently we use an old version of mypy that bundles possibly-outdated type stubs for key libraries we use like flask. Because new mypy (as planned in https://github.com/freedomofpress/securedrop/pull/6351) no longer bundles those stubs, it'll be impossible to use mypy locally unless you also have flask, etc. installed in your development virtualenv.

In addition this will make using editors like PyCharm much easier, because it expects all the dependencies to be installed in a single virtualenv.

Finally, this should bring the local development environment closer to what CI does, which AIUI takes a venv with `securedrop-app-code-requirements` deps installed, and then installs `develop-requirements` over it.

The only obstacle to this is that right now develop wants click==7, while sd-app-code wants click==8. This appears to be easily fixable, since our molecule version is already compatible with click 8.

legoktm:
## Status

Ready for review

## Description of Changes

* Upgrade mypy to 0.941 and sqlalchemy-stubs to 0.4. Type stubs are no
longer bundled with mypy, so we need to explicitly install various
types-{name} packages.
* Since we need to install extra packages and need other dependencies
  like Flask, move mypy into test-requirements.txt and run it via the
  dev container.
* The main issue spotted by the SQLAlchemy plugin, that
sources.filesystem_id is nullable, is being fixed in a separate PR:
#6350.
* Drop the explicit type comments now that mypy can infer them.
* Use `--explicit-package-bases` when running mypy, so it treats the
securedrop/admin directories as modules regardless of whether
`__init__.py` files exist. This is needed so the two tests/
directories don't conflict with each other.
* Add missing return types to all schema changes and the templates that
generate them.
* In the `create_source_uuid_column` migration, drop the use of
`quoted_name()`. For some reason mypy doesn't like it, but more
importantly we don't need it in SQLite and we don't use it in any
other schema change.
* Make SecureTemporaryFile.write()'s return type match its parent by
returning the number of bytes that were written.

Fixes #6346.
Fixes #6358.

## Testing

* Verify that CI passes

## Deployment

## Checklist


legoktm:
We also need to figure out a solution to https://github.com/freedomofpress/securedrop/issues/6358 otherwise `make venv && source .venv/bin/activate && make typelint` will break for developers.

zenmonkeykstop (approved):
LGTM!

PR #6353, opened 2022-03-21
Fix deduping in test_schema_unchanged_after_up_then_downgrade

### LINKED ISSUE #6352: Next schema change will fail test_schema_unchanged_after_up_then_downgrade because of instance_config weirdness
## Description

The next schema change will fail the `test_schema_unchanged_after_up_then_downgrade` test because of something wrong with the `instance_config` table:

```
E               AssertionError: Schema for ('table', 'instance_config', 'instance_config') did not match:
E               Left:
E               CREATE TABLE "instance_config" (
E               	version INTEGER NOT NULL, 
E               	valid_until DATETIME NOT NULL, 
E               	allow_document_uploads BOOLEAN, 
E               	organization_name VARCHAR(255), 
E               	initial_message_min_len INTEGER DEFAULT '0' NOT NULL, 
E               	reject_message_with_codename BOOLEAN DEFAULT '0' NOT NULL, 
E               	PRIMARY KEY (version), 
E               	CHECK (allow_document_uploads IN (0, 1)), 
E               	CHECK (reject_message_with_codename IN (0, 1)), 
E               	CHECK (reject_message_with_codename IN (0, 1)), 
E               	UNIQUE (valid_until)
E               )
E               Right:
E               CREATE TABLE "instance_config" (
E               	version INTEGER NOT NULL, 
E               	valid_until DATETIME NOT NULL, 
E               	allow_document_uploads BOOLEAN, 
E               	organization_name VARCHAR(255), 
E               	initial_message_min_len INTEGER DEFAULT '0' NOT NULL, 
E               	reject_message_with_codename BOOLEAN DEFAULT '0' NOT NULL, 
E               	PRIMARY KEY (version), 
E               	CHECK (allow_document_uploads IN (0, 1)), 
E               	CHECK (reject_message_with_codename IN (0, 1)), 
E               	CHECK (reject_message_with_codename IN (0, 1)), 
E               	UNIQUE (valid_until)
E               )

tests/test_alembic.py:63: AssertionError
```

You will notice that both of those schemas are identical.

## Steps to Reproduce

Create an entirely no-op migration, like https://gist.github.com/legoktm/c0384c339e1485c350bbc6d1b6b2da6b

## Expected Behavior

Tests pass

## Actual Behavior

See failure above.

## Comments

I ran into this while working on https://github.com/freedomofpress/securedrop/pull/6350. @zenmonkeykstop since you touched this part in https://github.com/freedomofpress/securedrop/commit/fb36b50afbfe96d988af0c19de16d424fff2e6b1#diff-1d479d01d55da852b7ea6c4f260fe28e83b88438446013fa093754299d063461, do you have any initial ideas what the problem might be? Otherwise I'll start poking at it.

legoktm:
## Status

Ready for review

## Description of Changes

The deduping added in fb36b50afbfe96 only applied to the alembic output.
But when we add another schema change on top of it, the test starts
failing because the right side isn't also being de-duped. The easiest
thing to do is dedupe that as well.

While we're at it, improve the test to use normal assertions so that
pytest can print out the exact values that aren't matching, rather than
*only* the schemas, which in this case were actually identical.

Fixes #6352.

## Testing

- [ ] CI passes
- [ ] Check out this change, verify this test passes locally (e.g. `./securedrop/bin/dev-shell bin/run-test --no-cov -v -k test_schema_unchanged_after_up_then_downgrade -vv`)
- [ ] Generate a dummy schema change that does nothing, you can grab https://gist.github.com/legoktm/c0384c339e1485c350bbc6d1b6b2da6b for this purpose. Then re-run the `test_schema_unchanged_after_up_then_downgrade` test and verify it too passes locally.

## Deployment

Any special considerations for deployment? No, this is test-only.

## Checklist

### DIFF
--- a/securedrop/tests/test_alembic.py
+++ b/securedrop/tests/test_alembic.py
@@ -4,12 +4,12 @@ import os
 import pytest
 import re
 import subprocess
+from collections import OrderedDict
 
 from alembic.config import Config as AlembicConfig
 from alembic.script import ScriptDirectory
 from os import path
 from sqlalchemy import text
-from collections import OrderedDict
 
 from db import db
 from journalist_app import create_app
@@ -55,22 +55,12 @@ def get_schema(app):
 
 
 def assert_schemas_equal(left, right):
-    for (k, v) in list(left.items()):
-        if k not in right:
-            raise AssertionError(
-                'Left contained {} but right did not'.format(k))
-        if not ddl_equal(v, right[k]):
-            raise AssertionError(
-                'Schema for {} did not match:\nLeft:\n{}\nRight:\n{}'
-                .format(k, v, right[k]))
-        right.pop(k)
-
-    if right:
-        raise AssertionError(
-            'Right had additional tables: {}'.format(list(right.keys())))
+    assert list(left) == list(right), 'Left and right do not contain same list of tables'
+    for (table, left_schema) in list(left.items()):
+        assert_ddl_equal(left_schema, right[table])
 
 
-def ddl_equal(left, right):
+def assert_ddl_equal(left, right):
     '''Check the "tokenized" DDL is equivalent because, because sometimes
         Alembic schemas append columns on the same line to the DDL comes out
         like:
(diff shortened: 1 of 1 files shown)


legoktm:
Wonderful, CI and my tests locally are giving different results. Guess we need the deduping after all :/

zenmonkeykstop (approved):
- [x] CI passes
- [x] Check out this change, verify this test passes locally (e.g. ./securedrop/bin/dev-shell bin/run-test --no-cov -v -k test_schema_unchanged_after_up_then_downgrade -vv)
- [x] Generate a dummy schema change that does nothing... Then re-run the test_schema_unchanged_after_up_then_downgrade test and verify it too passes locally.

legoktm:
Thanks! I don't think we need to backport this to 2.3.0 since it's test-only, but if we do need to do another schema change for whatever reason in 2.3.x (seems pretty unlikely), we'll need it.

PR #6359, opened 2022-03-23
Include securedrop-app-code deps into develop-requirements.txt

### LINKED ISSUE #6358: Install securedrop-app-code dependencies in development environment via `make venv`

legoktm:
## Status

Ready for review

## Description of Changes

We want the development virtualenv to include the production dependencies
as well as the development ones for the benefit of tools like mypy as well
as editors like PyCharm. Care was taken to ensure all the versions match
those used in production.

The main change to develop-requirements.txt is that click has been upgraded
to v8, because Flask v2 requires it.

We also have `make venv` downgrade setuptools before we install anything,
this is mostly for the benefit of `jsmin` which uses now-removed setuptools
functionality and requires an older version just to be installable.

Fixes #6358.

## Testing

* [ ] Verify the dependencies listed in `securedrop-app-code-requirements.txt` have the same version in `develop-requirements.txt`
* [ ] Delete or move any existing `.venv` out of the way
* [ ] Run `make venv`, activate it with `source .venv/bin/activate`, verify with `pip freeze` that it installs both development dependencies (e.g. mypy) and production ones (e.g. flask)
* [ ] Run `make lint`, observe everything passes (this also tests molecule is OK with the new click version)
* [ ] Run `make safety`, observe it outputs whatever fine (verifying safety is OK with the new click version)

## Deployment

## Checklist

  - I will open a PR to recommend the use of `make venv`, otherwise you'll run into errors when trying to install jsmin with modern Python.


legoktm:
We discussed today that we'd like these to be in test-requirements.txt instead, and start running mypy in the dev container, like we do for pytest, etc.

legoktm:
I folded this into https://github.com/freedomofpress/securedrop/pull/6351.

PR #6361, opened 2022-03-26
Refactor source functional tests codenames and collision

nabla-c0d3:
## Status

Work in progress

## Description of Changes

This PR continue the work started in #6307, by bringing the new test code and fixtures to a few more functional tests for the source app.

As before, the goals are to:

* Split the FunctionalTest and the related MixIn classes into smaller classes in order to make the tests more readable.
* Use more modern fixtures that don't rely on mutating the SDConfig object or attributes attached to the Test c;lass.
* Overall, speed up and simplify the functional tests. After this work is done (more PRs to come), it should also be possible to run the functional tests in parallel, to make them even faster.

Issue #4558, opened 2019-06-27
Fonts in screenshots generated by functional tests are broken


eloquence:
The functional test run produces screenshot artifacts in `tests/pageslayout/screenshots` that can be used for the documentation. However, the fonts in those screenshots are currently broken (see below).

A fix currently exists in this remote branch:
https://github.com/rmol/securedrop/commit/6f493c745a8e0532c7c112f2a5949627875ec0fd

![source-login](https://user-images.githubusercontent.com/213636/60239308-09320200-9862-11e9-9de7-a62804973900.png)

PR #6365, opened 2022-03-28
Refactor source functional tests for static pages and session timeout

nabla-c0d3:
## Status

Ready

## Description of Changes

This PR continue the work from https://github.com/freedomofpress/securedrop/pull/6361, by bringing the new test code and fixtures to a few more functional tests for the source app.

More specifically, this PR:

* Refactors the functional test for when a source's session expires.
* Refactors a few test checking some of the source app's static pages.
* Moves the "pagelayout" tests to a sub-folder within the functional tests, so that fixtures can be shared (via conftest.py, to help with #3836). Fixtures were already shared previously but that was done by having a second class called `FunctionalTest`  inheriting from `functional_test.FunctionalTest`. Ultimately, most of the pagelayout tests are indeed functional tests and should be treated that way. This will also avoid having duplicate tests (one in functional, one in pagelayout, testing the exact same thing).


securedrop/tests/functional/pageslayout/test_journalist_account.py
```diff
@@ -28,6 +29,9 @@ class TestJournalistLayoutAccount(
         source_navigation_steps.SourceNavigationStepsMixin,
         journalist_navigation_steps.JournalistNavigationStepsMixin):
 
+    # TODO(AD): This test fails often when run with the "ar" locale
+    #  The screenshot that gets saved is blank, which causes autocrop_btm() to fail
+    @flaky(max_runs=5)
```
nabla-c0d3, line 34: This test is also flaky on the develop branch.

securedrop/tests/functional/pageslayout/test_journalist_account.py
```diff
@@ -28,6 +29,9 @@ class TestJournalistLayoutAccount(
         source_navigation_steps.SourceNavigationStepsMixin,
         journalist_navigation_steps.JournalistNavigationStepsMixin):
 
+    # TODO(AD): This test fails often when run with the "ar" locale
+    #  The screenshot that gets saved is blank, which causes autocrop_btm() to fail
+    @flaky(max_runs=6)
```
nabla-c0d3: This test is also flaky on the develop branch.

nabla-c0d3:
Ready for review 👍

zenmonkeykstop:
Tests pass happily but (at least for me) the `en_US` pagelayout screenshots have unicode markers in place of text characters. Will update with a proper review after poking about a bit.

zenmonkeykstop:
Tests pass happily but (at least for me) the `en_US` pagelayout screenshots have unicode markers in place of text characters when run in the context of the full test suite. Will update with a proper review after poking about a bit.

nabla-c0d3:
@zenmonkeykstop Sorry I should have checked the screenshots. Are you seing the same problem as in https://github.com/freedomofpress/securedrop/issues/4558 ?

zenmonkeykstop:
Looked a bit more, it doesn't seem to be. But as it turn out, my apologies, this wasn't actually introduced by your changes, the same behavior is present in current `develop` and the last release tag as well. It's not a merge blocker - the only time the screenshots are relevant are when they're used to update the docs, and since the page layout tests are run separately for that the issue doesn't appear.

zenmonkeykstop (dismissed):
Changes make sense and tests are passing. 

There is one unintended consequence in CI - the `app-page-layout-tests` job runs the entire test suite, as CircleCI's cli tool tries to find tests matching `/tests/pageslayout/test*py`, gets no result, and then that gets interpreted as a match against everything. I've pushed a commit to fix, if it works then this is good to go.

zenmonkeykstop (approved):
Changes make sense and tests are passing.

There is one unintended consequence in CI - the app-page-layout-tests job runs the entire test suite, as CircleCI's cli tool tries to find tests matching /tests/pageslayout/test*py, gets no result, and then that gets interpreted as a match against everything. I've pushed a commit to fix, if it works then this is good to go.

(Updating to add another fix to `make safety`, required to get CI passing)

nabla-c0d3:
@zenmonkeykstop Thanks for reviewing this! Next PR is at https://github.com/freedomofpress/securedrop/pull/6379 👍

PR #6368, opened 2022-03-29
Run string formatting after gettext() call

### LINKED ISSUE #6367: anonymous string-formatting fields break translation
**This description is superseded by <https://github.com/freedomofpress/securedrop/issues/6367#issuecomment-1082391316>!**

## Description

When a `gettext()`-wrapped string uses anonymous string-formatting fields `{}`, rather than positional fields `{n}` and named fields `{foo}`, translation of that string breaks silently:

1. in Jinja templates, entirely, falling back to the English source string; and
2. in Flask views, by omitting the field(s).

## Steps to Reproduce

TK, reproduction branch not yet pushed

1. From the admin interface, enable a 20-character minimum a source's first message.
2. Start a new Source Interface session.
3. Switch to any translated locale (e.g. `de_DE`).

## Expected Behavior

TK, reproduction branch not yet pushed

### Source string
`Your first message must be at least {} characters long.`

### Translation template
[`Ihre erste Nachricht muss mindestens {} Zeichen lang sein.`](https://weblate.securedrop.org/translate/securedrop/securedrop/de/?checksum=461cee53e88af00c)

### Translated string
`Ihre erste Nachricht muss mindestens 20 Zeichen lang sein.`

## Actual Behavior

Screen Shot 2022-03-29 at 13 11 56

## Comments

For the two source strings that display this behavior, the code-level fix is as simple as `s/{}/{foo}`.  This will propagate downstream to Weblate as a source-string change requiring translation changes.  We could make the latter on our own, without involving translators, but not without further delaying the current v2.3.0 release.

legoktm:
## Status

Ready for review

## Description of Changes

Don't interpolate parameters before looking up the string in gettext,
otherwise it won't find anything and fallback to English.

Fixes #6367.

## Testing

* Rebase onto @cfm's i18n branch (which might not exist yet), enable minimum length, change language to German, verify no raw `{}` appear in the error message when inputting too short of a message, and the correct character count is shown.

## Deployment

## Checklist

### DIFF
--- a/securedrop/source_app/main.py
+++ b/securedrop/source_app/main.py
@@ -204,7 +204,7 @@ def make_blueprint(config: SDConfig) -> Blueprint:
             min_len = InstanceConfig.get_default().initial_message_min_len
             if (min_len > 0) and (msg and not fh) and (len(msg) < min_len):
                 flash(gettext(
-                    "Your first message must be at least {} characters long.".format(min_len)),
+                    "Your first message must be at least {} characters long.").format(min_len),
                     "error")
                 return redirect(url_for('main.lookup'))
 
--- a/securedrop/source_templates/lookup.html
+++ b/securedrop/source_templates/lookup.html
@@ -46,11 +46,11 @@
         {{ form.msg(class="fill-parent") }}
   {% if allow_document_uploads %}
       {% if min_len > 0 %}
-         <p class="center" id="min-msg-length">{{ gettext('If you are only sending a message, it must be at least {} characters long.'.format(min_len)) }}</p>
+         <p class="center" id="min-msg-length">{{ gettext('If you are only sending a message, it must be at least {} characters long.').format(min_len) }}</p>
       {% endif %}
   {% else %}
       {% if min_len > 0 %}
-        <p class="center" id="min-msg-length">{{ gettext('Your first message must be at least {} characters long.'.format(min_len)) }}</p>
+        <p class="center" id="min-msg-length">{{ gettext('Your first message must be at least {} characters long.').format(min_len) }}</p>
       {% endif %}
   {% endif %}
       </div>


securedrop/source_app/main.py
```diff
@@ -204,7 +204,7 @@ def submit(logged_in_source: SourceUser) -> werkzeug.Response:
             min_len = InstanceConfig.get_default().initial_message_min_len
             if (min_len > 0) and (msg and not fh) and (len(msg) < min_len):
                 flash(gettext(
-                    "Your first message must be at least {} characters long.".format(min_len)),
+                    "Your first message must be at least {} characters long.").format(min_len),
```
zenmonkeykstop, line 207: testing - confirmed that this case works correctly when rebased onto the `i18n-merge` branch.

securedrop/source_templates/lookup.html
```diff
@@ -46,11 +46,11 @@ <h2 id="submit-heading" class="headline">{{ gettext('Submit Messages') }}</h2>
         {{ form.msg(class="fill-parent") }}
   {% if allow_document_uploads %}
       {% if min_len > 0 %}
-         <p class="center" id="min-msg-length">{{ gettext('If you are only sending a message, it must be at least {} characters long.'.format(min_len)) }}</p>
+         <p class="center" id="min-msg-length">{{ gettext('If you are only sending a message, it must be at least {} characters long.').format(min_len) }}</p>
```
zenmonkeykstop, line 49: Testing: this case did not work correctly, but that's due to this string not being collected for translation, the code change is valid.

securedrop/source_app/main.py
zenmonkeykstop, line 207: testing - confirmed that this case works correctly when rebased onto the `i18n-merge` branch.

securedrop/source_templates/lookup.html
```diff
@@ -46,11 +46,11 @@ <h2 id="submit-heading" class="headline">{{ gettext('Submit Messages') }}</h2>
...
       {% endif %}
   {% else %}
       {% if min_len > 0 %}
-        <p class="center" id="min-msg-length">{{ gettext('Your first message must be at least {} characters long.'.format(min_len)) }}</p>
+        <p class="center" id="min-msg-length">{{ gettext('Your first message must be at least {} characters long.').format(min_len) }}</p>
```
zenmonkeykstop, line 53: Testing: this case works correctly (with min-length set and document uploads disabled).

zenmonkeykstop (approved):
LGTM based on visual review and testing in `make dev`.

PR #6379, opened 2022-03-30
Refactor source layout tests for tor browser

nabla-c0d3:
## Status

Ready

## Description of Changes

This PR continue the work from https://github.com/freedomofpress/securedrop/pull/6365, by bringing the new test code and fixtures to a few more functional tests for the source app.

More specifically, this PR:

* Brings the new test fixtures to a couple layout tests (for #3836).
* Combine the two layout tests into one, for a slightly faster test suite.

### DIFF
--- a/securedrop/tests/functional/pageslayout/test_source_layout_torbrowser.py
+++ b/securedrop/tests/functional/pageslayout/test_source_layout_torbrowser.py
@@ -16,32 +16,35 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 import pytest
+from tbselenium.utils import disable_js
 
-from tests.functional import journalist_navigation_steps
-from tests.functional import source_navigation_steps
-from tests.functional.functional_test import TORBROWSER
-from . import functional_test
+from tests.functional.app_navigators import SourceAppNagivator
+from tests.functional.pageslayout.functional_test import list_locales
+from tests.functional.pageslayout.screenshot_utils import save_screenshot_and_html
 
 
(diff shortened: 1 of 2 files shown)


securedrop/tests/functional/pageslayout/test_source_layout_torbrowser.py
```diff
@@ -16,32 +16,35 @@
...
+    def test_index_and_logout(self, locale, sd_servers_v2):
+        # Given a source user accessing the app from their browser
+        locale_with_commas = locale.replace("_", "-")
+        with SourceAppNagivator.using_tor_browser_web_driver(
+            source_app_base_url=sd_servers_v2.source_app_base_url,
+            accept_languages=locale_with_commas,
+        ) as navigator:
+
+            # And they have disabled JS in their browser
+            disable_js(navigator.driver)
```
nabla-c0d3, line 39: I wasn't sure why JS needs to be disabled here.

eaon: You're right, it doesn't strictly need to be disabled to work, but it does represent the ideal way of how we prefer sources to use SecureDrop at the moment. That would change once we have an instance-independent way to verify that the JS sources would execute is the same that we ship.

(PS: As someone who spent a lot of time on the source interface in recent months, thanks so much for your work!)

zenmonkeykstop (approved):
LGTM!

nabla-c0d3:
Great! Only a few more left (for now 😀). The next PR is at https://github.com/freedomofpress/securedrop/pull/6382.

PR #6382, opened 2022-04-02
Refactor functional test for submitting and retrieving a message

nabla-c0d3:
## Status

Ready.

## Description of Changes

This PR continue the work from https://github.com/freedomofpress/securedrop/pull/6379. More specifically, this PR:

* Brings the new test fixtures to the `TestSubmitAndRetrieveMessage` functional test (for #3836).
* Introduces a new fixture, `sd_servers_v2_with_clean_state` for functional tests that need to modify the app's state (for example by submitting a file).
* Introduces the `JournalistAppNavigator` class, which will eventually replace `JournalistNavigationStepsMixin` for navigating through the journalist app.

### DIFF
--- a/securedrop/tests/functional/app_navigators.py
+++ b/securedrop/tests/functional/app_navigators.py
@@ -1,7 +1,12 @@
+import gzip
+import tempfile
 import time
 from contextlib import contextmanager
-from typing import Optional, Generator
 
+from typing import Optional, Dict, Iterable, Generator
+
+import pyotp
+import requests
 from selenium.webdriver.firefox.webdriver import WebDriver
 from selenium.webdriver.remote.webelement import WebElement
 
@@ -12,6 +17,10 @@ from selenium.webdriver.support import expected_conditions
 from selenium.webdriver.support.ui import WebDriverWait
 
 from tests.functional.web_drivers import get_web_driver, WebDriverTypeEnum
+
+from encryption import EncryptionManager
+from tests.functional.tor_utils import proxies_for_url
+from tests.test_encryption import import_journalist_private_key
 
 
 # TODO(AD): This is intended to eventually replace the navigation/driver code in FunctionalTest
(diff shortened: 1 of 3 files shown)


securedrop/tests/functional/test_submit_and_retrieve_message.py
```diff
@@ -1,21 +1,46 @@
...
-        self._source_visits_source_homepage()
-        self._source_chooses_to_submit_documents()
-        self._source_continues_to_submit_page()
-        self._source_submits_a_message()
-        self._source_logs_out()
-        self.switch_to_firefox_driver()
-        self._journalist_logs_in()
-        self._journalist_checks_messages()
-        self._journalist_downloads_message()
-        self.switch_to_torbrowser_driver()
```
nabla-c0d3, line 21: This refactor also fixes the problem of the web driver used by the test being a singleton attached to `self`.

It's the reason for the calls to `self.switch_to_firefox_driver()` and `self.switch_to_torbrowser_driver()` in the code above, which modify the global state of the whole test suite ie. the web driver to use. 

If, as a developer, I forget to call `self.switch_to_torbrowser_driver()` at the end of my test, it will mess up the whole test suite.

The refactored code below does not have that problem.

securedrop/tests/functional/test_submit_and_retrieve_message.py
```diff
@@ -1,21 +1,46 @@
...
+        # When a journalist logs in
+        journ_app_nav = JournalistAppNavigator(
+            journalist_app_base_url=sd_servers_v2.journalist_app_base_url,
+            web_driver=firefox_web_driver,
+        )
+        journ_app_nav.journalist_logs_in(
+            username=sd_servers_v2.journalist_username,
+            password=sd_servers_v2.journalist_password,
+            otp_secret=sd_servers_v2.journalist_otp_secret,
+        )
```
nabla-c0d3, line 35: Arguments are passed explicitly instead of being pulled out of `self`.

securedrop/tests/functional/app_navigators.py
```diff
@@ -250,3 +258,120 @@ def source_sees_flash_message(self) -> str:
         notification = self.driver.find_element_by_css_selector(".notification")
         assert notification
         return notification
+
+
+# TODO(AD): This is intended to eventually replace the JournalistNavigationStepsMixin
+class JournalistAppNavigator:
```
nabla-c0d3, line 285: Mostly the same as `JournalistNavigationStepsMixin`.

securedrop/tests/functional/app_navigators.py
```diff
@@ -270,3 +279,120 @@ def source_sees_flash_message(self) -> str:
...
+            fp.write(raw_content)
+            fp.seek(0)
+
+            gzf = gzip.GzipFile(mode="rb", fileobj=fp)
+            content = gzf.read()
+
+        return content.decode("utf-8")
+
+    def journalist_downloads_first_message(
+        self, encryption_mgr_to_use_for_decryption: EncryptionManager
```
nabla-c0d3, line 361: The `EncryptionManager` needs to be explicitly passed, as the one used by the `sd_servers` fixtures (ie. in a separate process) might have a different configuration than the "local" / in-process one (which, in reality, has nothing to do with the functional tests).

This is something that was confusing in the functional test suite.

nabla-c0d3:
It looks like the `app-tests` CI step does not run the updated test from this PR; it only runs a subset of all the available unit tests.
I think this wasn't the case before? @zenmonkeykstop do you know how CI decides which tests to run?

zenmonkeykstop:
> It looks like the `app-tests` CI step does not run the updated test from this PR; it only runs a subset of all the available unit tests. I think this wasn't the case before? @zenmonkeykstop do you know how CI decides which tests to run?
 
the `circleci` utility gets passed a list of paths with wildcards, which it uses to find valid test files. See 
https://github.com/nabla-c0d3/securedrop/blob/898e0a87bf738395ce4fb8fc911a4d3b5ec2c659/.circleci/config.yml#L101 on your current branch. I'll take a look now to see whether that's working as expected, if so we may just need to update the list of test paths.

zenmonkeykstop:
It looks like it did run? The tests are parallelized across 3 containers at present, and I see it here: https://app.circleci.com/pipelines/github/freedomofpress/securedrop/4343/workflows/d52e7ff6-8d96-4a85-acf6-75bcefb067e6/jobs/62619/parallel-runs/2?filterBy=ALL&invite=true#step-109-367

nabla-c0d3:
> It looks like it did run? The tests are parallelized across 3 containers at present, and I see it here: https://app.circleci.com/pipelines/github/freedomofpress/securedrop/4343/workflows/d52e7ff6-8d96-4a85-acf6-75bcefb067e6/jobs/62619/parallel-runs/2?filterBy=ALL&invite=true#step-109-367

That's my bad - it was run indeed!

zenmonkeykstop (approved):
LGTM

nabla-c0d3:
👍👍👍 The next PR is at https://github.com/freedomofpress/securedrop/pull/6383.

PR #6390, opened 2022-04-07
Make `gpg` not check the trustdb as we do not use the WoT

### LINKED ISSUE #6389: Large `gpg` keyrings slow down application performance due to automatic Web of Trust checks
## Description

Automatic trustdb checks during `gpg` operations may cause those operations to run long on instances with large public keyrings, degrading application performance or in extreme cases causing page timeouts.

## Steps to Reproduce

* With a fresh install, run `loaddata.py` often enough to generate keys for 10-20k sources
* Repeatedly generate new sources manually and observe intermittent slow response times

## Comments

`gpg` provides the `--no-auto-check-trustdb` argument that skips regular trustdb checks. Ought to be the right way for us to handle this as we do not make use of the Web of Trust at all.

eaon:
## Status

Ready for review

## Description of Changes

GPG automatically checks the trustdb on the fly, an operation that can take a long time with very large keyrings as they sometimes exist on large instances. We don't use the Web of Trust at all though so we want to tell GPG to skip these checks by default.

Fixes #6389

## Testing

* Source key generation
  * During `make dev` startup, run `while true; do grep "--no-auto-check-trustdb"<<<$(ps a); sleep 1; done` and confirm that `--no-auto-check-trustdb` is actually passed to `gpg2` 
  * With a fresh install, run `loaddata.py` to generate keys for 10-20k sources
  * Repeatedly generate new sources manually and see response times stay approximately the same
* Source deletion
  * Log into journalist interface 
  * Run `while true; do grep "--no-auto-check-trustdb"<<<$(ps a); done` (no sleep as this operation is faster than generating a key)
  * Delete source account and ensure `--no-auto-check-trustdb` is in the the `grep` loop output

## Checklist

- [x] I have written a test plan ~and validated it for this PR~

### DIFF
--- a/securedrop/encryption.py
+++ b/securedrop/encryption.py
@@ -128,7 +128,9 @@ class EncryptionManager:
         # invoking pinentry-mode=loopback
         # see: https://lists.gnupg.org/pipermail/gnupg-users/2016-May/055965.html
         self._gpg_for_key_deletion = gnupg.GPG(
-            binary="gpg2", homedir=str(self._gpg_key_dir), options=["--yes"]
+            binary="gpg2",
+            homedir=str(self._gpg_key_dir),
+            options=["--yes", "--no-auto-check-trustdb"]
         )
 
         # Ensure that the journalist public key has been previously imported in GPG
--- a/securedrop/encryption.py
+++ b/securedrop/encryption.py
@@ -2,7 +2,7 @@ import typing
 from distutils.version import StrictVersion
 from io import StringIO, BytesIO
 from pathlib import Path
-from typing import Optional, Dict, List
+from typing import Optional, Dict, List, FrozenSet
 
 import pretty_bad_protocol as gnupg
 import os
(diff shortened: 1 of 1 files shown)


eaon:
I updated test plan above to reflect the new commit adding `--no-auto-check-trustdb` to the deletion invocation

zenmonkeykstop (approved):
Source key generation:

- [x]  During make dev startup, run while true; do grep "--no-auto-check-trustdb"<<<$(ps a); sleep 1; done and confirm that --no-auto-check-trustdb is actually passed to gpg2
-  With a fresh install, run loaddata.py often enough to generate keys for 10-20k sources
  - [x] Repeatedly generate new sources manually and see response times stay approximately the same (tested on dev environment with ~3k sources and prod environment with ~20k sources with diff manually applied)

Source deletion:
- Log into journalist interface
- Run while true; do grep "--no-auto-check-trustdb"<<<$(ps a); done (no sleep as this operation is faster than generating a key)
- [x] Delete source account and ensure --no-auto-check-trustdb is in the the grep loop output

LGTM, no side-effects noted during exploratory testing on prod or dev either.

PR #6398, opened 2022-04-12
Use `gpg` with `--trust-model direct`

### LINKED ISSUE #6397: Make `gpg` calls use `--trust-model` instead of `--no-auto-check-trustdb`
## Description

In #6390 we introduced the use of the `--no-auto-check-trustdb` option for `gpg` to keep it from wasting CPU cycles on checking whether keys are trusted or not. While this does the job, a side effect of this is that calling `gpg` will still check whether the trustdb is stale and inform the user by writing to stderr.

The way SecureDrop utilises `gpg` on the server though means all keys are always trusted. `gpg` has native support for alternative trust models with its `--trust-model` option, which supports [`always`](https://www.gnupg.org/documentation/manuals/gnupg/GPG-Configuration-Options.html#index-trust_002dmodel_003aalways) and [`direct`](https://www.gnupg.org/documentation/manuals/gnupg/GPG-Configuration-Options.html#index-trust_002dmodel_003adirect). Using either of would skip all WoT associated checks. My reading is that both options would be interchangeable in our case since all keys except the journalist key are generated locally, and so will therefore have ultimate trust.

This would also allow us to remove the associated monkey patch, because pretty-bad-protocol / python-gnupg supports the `--trust-model` options natively.

eaon:
## Description of Changes

`--no-auto-check-trustdb` still glances at the trustdb, even if only to check if it's stale or not, but we really truly do not care about the WoT in our usecase, which `--trust-model direct` makes explicit. Also removes python-gnupg/pretty-bad-protocol monkey patch.

Fixes #6397

## Testing

* Source key generation
  * During `make dev` startup, run `while true; do grep "--no-auto-check-trustdb"<<<$(ps a); sleep 1; done` and confirm that `--trust-model direct` is passed to `gpg2`
  * Repeatedly generate new sources on an install with 10-20k sources, and make sure response times stay approximately the same
* Source deletion
  * Log into journalist interface
  * Run `while true; do grep "--trust-model direct"<<<$(ps a); done` (no sleep as this operation is faster than generating a key)
  * Delete source account and ensure `--trust-model direct` is in the the `grep` loop output

## Checklist

- [x] I have written a test plan ~and validated it for this PR~

### DIFF
--- a/securedrop/encryption.py
+++ b/securedrop/encryption.py
@@ -2,7 +2,7 @@ import typing
 from distutils.version import StrictVersion
 from io import StringIO, BytesIO
 from pathlib import Path
-from typing import Optional, Dict, List, FrozenSet
+from typing import Optional, Dict, List
 
 import pretty_bad_protocol as gnupg
 import os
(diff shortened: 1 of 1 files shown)


zenmonkeykstop (approved):
Ran through test plan on staging env:
- [x] source creation consistently around the 4sec mark on a dataset with 10k sources (~3 sec or so for the list-keys operation, which is where the slowdown was with the previous trust model)
- [x] source deletion is also consistently snappy, < 1sec.

LGTM, thanks for following up with an even better fix!

PR #6401, opened 2022-04-13
Disable fwupd-refresh.timer, triggers OSSEC warnings

### LINKED ISSUE #6204: fwupd produces multiple alerts
Example notification:

```
OSSEC HIDS Notification.
2021 Dec 20 19:06:01

Received From: (app) 10.20.2.2->/var/log/syslog
Portion of the log(s):

Dec 20 19:06:01 app fwupd[135559]: message repeated 2 times: [ 19:06:01:0220 GLib                 g_bytes_get_data: assertion 'bytes != NULL' failed]
```

We don't currently support `fwupd` and have previously silenced its alerts: #5882, #6107

legoktm:
## Status

Ready for review

## Description of Changes

For various reasons, the timer to run `fwupdmgr refresh` ocassionally
triggers OSSEC alerts, which admins can't do anything about.

We currently don't use fwupd for firmware updates, so the daily refresh
of metadata is useless and should be safe to disable. If in the future
we do want admins to install updates with fwupd, they can run refresh
manually as part of the process.

Fixes #6204.

## Testing

* [ ] Run `make build-debs` to get a new securedrop-config deb package
* [ ] Verify on a staging instance that `systemctl is-enabled fwupd-refresh.timer` prints "enabled", and `systemctl list-timers` shows it too
* [ ] Install the newly built package, `sudo apt install ./securedrop-config[...].deb`
* [ ] Verify package installation was successful, and the `is-enabled` command from earlier now prints "disabled"
* [ ] Re-install the package again with `sudo apt install --reinstall ./securedrop-config[...].deb`, to run the postinst again, against an already-disabled state.
* [ ] Verify package re-installation was successful, and the `is-enabled` command still prints "disabled".
* [ ] Reboot, then verify fwupd-refresh is no longer printed by `systemctl list-timers`

## Deployment

Yes, this affects the postinst of the package, which means it cannot be allowed to fail under any circumstances.

## Checklist

### DIFF
--- a/install_files/securedrop-config-focal/DEBIAN/postinst
+++ b/install_files/securedrop-config-focal/DEBIAN/postinst
@@ -20,6 +20,9 @@ case "$1" in
     cp /opt/securedrop/50unattended-upgrades /etc/apt/apt.conf.d/
     cp /opt/securedrop/reboot-flag /etc/cron.d/
 
+    # Disable fwupd-refresh (#6204)
+    systemctl is-enabled fwupd-refresh.timer && systemctl disable fwupd-refresh.timer
+
     ;;
     abort-upgrade|abort-remove|abort-deconfigure)
     ;;


legoktm:
Of course, the securedrop-app-code package isn't installed on the monitor server, so probably the securedrop-config package is a better option, like Conor had originally suggested.

zenmonkeykstop (approved):
- Followed test plan in staging against both app-staging and mon-staging servers, everything checks out.
- Also did a fresh staging install with the debs built from this branch, confirmed the fwupd-refresh timer is disabled there too.

LGTM! Looking forward to a world of fewer OSSEC false positives.

PR #6403, opened 2022-04-18
Unified server side sessions for JI and API

### LINKED ISSUE #204: Flask cookies leak (server-side) session values

### LINKED ISSUE #6224: Switch from itsdangerous to another JWT library

### LINKED ISSUE #6428: Move away from flask.g.user
## Description

Previously we had identified that sticking stuff onto `flask.g` should be discouraged because it hurts type checking, pylint doesn't like it, etc.

With `g.user` specifically, it's somewhat problematic because it's now a secondary source of whether a user is logged in or not, and who that user is - which is really the responsibility of the session. In the process of developing https://github.com/freedomofpress/securedrop/pull/6403, we hit this and needed to add more checks to keep them in sync.

But now that we have a custom session class, we could easily add a `session.current_user()` that returns a `Journalist` object, and not have to worry about it falling out of sync with the session.

lsd-cat:
## Status
- [X] Design phase ended
- [X] Coding pahse ended
- [x] Test plan

## Summary of behavioral changes
***Reminder: the changes apply only to the Journalist Interface and Journalist API, NOT to the Source Interface***

| Change  |  Before |  After |  Comment |
|---|---|---|---|
| GUI Session expiration  | [120 minutes](https://github.com/freedomofpress/securedrop/blob/081b2b2d8a4a5205033a599ccc7a304c5dd443cd/securedrop/journalist_app/__init__.py#L133-L150) - Renewable unlimited times upon any request | 120 minutes - Renewable up to 5 times upon any request in the last half hour of validity  |   | 
| API Session expiration  | [8 hours](https://github.com/freedomofpress/securedrop/blob/f8aecebfc0836fe1bb4fa633ff83e539cf83a686/securedrop/journalist_app/api.py#L27) - Nonrenewable | 120 minutes - Renewable up to 5 times upon any request in the last half hour of validity  | Uses same backend and mechanism as the JI.It is renewed automatically, so the client does not need to issue any 'renew' or 'refresh' request. In the last half an hour of its validity, if any request is made, token life is extended server-side. |
| Session info storage JI | Flask signed session cookie, the payload is fully client-side, no server state | Signed cookie with a session identifier. The identifier matches a record in Resid with the actual session payload | Sessions are now stored server-side with all the consequences of this. |
| Session info storage API | The signed JWT token contains the signed payload, namely the authenticated user ID and the expiration. | The token is now the signed session identifier with the same backend as the JI. | API/JI session tokens are not exchangeable as they use different prefixes in Redis and different signing salts. |
| API token request | The current API request to issue a token is not subject to any change. | N/A | The expiration is the initial 2 hours expiration. The behavior can be easily changed, but it seems not to be used by securedrop-client anyway. | 
| Cookie signing  | Cookies are signed to ensure their integrity.  | Cookies are signed to prevent usage of non-issued server tokens (weird Redis payloads, leaked ids from Redis, etc).  | In general, this is not enforced by most server-side implementations. As it has little overhead, looks like a reasonable hardening.  |

## Additional changes to the `g` object
See https://github.com/freedomofpress/securedrop/issues/6428

## Description of Changes
This commit attempts to completely redesign the session mechanism for the Journalist Interface and the Journalist API in order to unify the two backends and behaviors, drop some dependencies, and have more control overall.

Currently, the JI relies on built-in Flask sessions for user authentication. They rely on using a session cookie with a signed payload using the app `SECRET_KEY`. These session cookies are renewable and non-revokable unless changing the `SECRET_KEY`. It is impossible to know how many sessions are there and destroy them server-side. Current logout functionality just destroys the cookie in the browser.
The API uses issues a non-renewable JWT token upon login. Upon logout, the token is placed in a database table for revocation until its expiration. For each authenticated API endpoint, a decorator is used to enforce the check of such token. Upon internal discussions, we have concluded that we are not benefiting from any of the benefits of JWT as a delegation mechanism in the project but we are keeping the burden of maintaining two session backends. Furthermore, `itsdangerous` has deprecated its JWT signing code.

The changes here move all the sessions from the client-side to the server-side. Sessions are now stored in Redis, and the auth cookie only contains a random string. The session identifier is still signed server-side and verified at each request: this is to prevent that a read primitive in Redis could allow an attacker to steal sessions; with this hardening, the attacker would also need the knowledge of the `SECRET_KEY` (or of course steal a valid cookie from a logged-in user). No major changes for the JI, while for the API the decorator is dropped, as well as the whole JWT mechanism and now behaves the same as the JI. Authentication on protected endpoints is enforced via the same approach of the JI, which is overall less error-prone than the decorator.
Both logouts now destroy sessions server-side. Before, on passphrase change, a nonce was incremented to invalidate past sessions for the affected user. That mechanism is dropped and sessions are simply deleted from the session store in that event.

Session lifetime is preset to 2 hours. Each session can be renewed a maximum of 5 times by simply using it 30 minutes before its expiration. Both values are configurable and can be tuned depending on the needs that arise from discussion or during testing.

No new dependencies are needed.
The database migration drops the `nonce` column for journalists and the `revoked_tokens` table for JWT revocation as well as the corresponding models.

## Fixes

Fixes #6428.
Fixes #6224.
Fixes #204 (only for the JI, NOT for the SI).

## Testing
For manual testing, a lot of manual logins/perform tasks/logout as well as for the API. We should test compatibility with securedrop-client and check if changes there are needed. The `token` API endpoint has not changed, as well as the authenticator header so we are not expecting big breakage.

## Deployment
The new `sessions` library has some configuration options that could be added to `config.py`. However, since it has been written ad hoc for this project, the default settings are already the proper ones.

## Source
https://github.com/fengsp/flask-session has been used as a starting point, albeit very little of it remains.

## Test Plan (draft)
Here is a quick deaft for the test plan. Will be updated this week as I go through the changes one last time and do some more manual testing.

### Migration
 * Database migration work (deletes the revoked token table)
 * Token cleanup procedures do not run/exists
~~* Reboot of the server/restart of redis logs destroy all sessions~~ Persistant by Redis default config
 * Test new config with non standard values
 * Config migration works
 
### Security
* Session ids are random
 * Sessions id are signed
##### Invalid session id (non existant/wrong signature) gets ignored
 1. Login to the JI
 2. Tamper the cookie signature
 - [x] Confirm that user is logged out

#####  Journalists can Login
 1. Login to the JI
 - [x] Confirm that the user is logged 

##### Sessions of multiple users are separate
1. Login to the JI as user X in browser A
2. Login to the JI as user Y in browser B
 - [x] Confirm that the two userss navigation does not interfere

##### Sessions are renewed in accordance with the new expiration/renewal policy
1. Login to the JI
2. Wait 2 hours
 - [x] Confirm that user is logged out automatically
 - [x] Confirm that an error message is shown
1. Login to the JI
2. Wait between 1.5h and 2h and refresh
3. Wait 1h
 - [x] Confirm that user is still logged in

#####  Logout works
1. Login to the JI as user X in browser A
2. Login to the JI as user Y in browser B
3. Logout from user X in browser A
- [x] Confirm that user A is logged out
- [x] Confirm that user B is still logged in

 #####  Deleting a user destroy all sessions of the user
1. Login to the JI as admin X in browser A
2. Login to the JI as admin/user Y in browser B
3. From admin X in browser A, delete user Y
 - [x] Confirm that user Y in browser B is logged out
 
 #####  Changing a user password destroy all sessions of the user
1. Login to the JI as admin X in browser A
2. Login to the JI as admin/user Y in browser B
3. From admin X in browser A, change password of user Y
 - [x] Confirm that user Y in browser B is logged out

1. Login to the JI as admin X
3. Change admin X password
 - [x] Confirm that admin X is logged out on password change

#####  Token are issues on succesful API login
1. Login to the `/token` endpoint
- [x] Confirm that the login is succesful
2. Use the token for an authenticated operation
- [x] Verify that the authenticated operation accepts the token
3. Send a request to the logout endpoint
- [x] Confirm that a message fo succesful logout is returned
4. Retry the operation at point 2.
- [x] Cionfirm that the operation at point 2 now requires a valid token

### Integration
***This changes should not affect the client, as the token format is exactly the same and token renewal will happen in the background transparently, however testing remains ncessary***
#####  Login and sesison lifetime
1. Login through the client
 - [x] Confirm that the client synchronizes succesfully
#####  Session lifetime
***We could need a bit of tuning here, what is going to happen is as follow:***
1. The server will issue a token valid for two hours
2. As the client constantly tries to sync, as soon as the token hit ~1.30h of lifetime the token will be renewed for 2 additional hours.
3. At the ~3.30h mark, the token will renew again, and again at ~5.30, ~7.30, ~9.30. After that 5 renewal, the token lifetime ends and the session is destroyed. (max duration time would be 12h)
It is possible, that during testing phase we want to adjust this settings (5 renewals, 2h each, triggered at 30 minutes to session end)

1. Login through the client
 - [x]  Confirm that the token gets renewed and the session last the expected time.


securedrop/tests/test_journalist_utils.py
```diff
@@ -3,38 +3,8 @@
 import pytest
 import random
 
-from models import RevokedToken
 from sqlalchemy.orm.exc import NoResultFound
 
-from journalist_app.utils import cleanup_expired_revoked_tokens
-
```
lsd-cat: The test here have been dropped as the table do not exists anymore. The file can probably be deleted.

securedrop/tests/test_journalist_api.py
```diff
@@ -1055,23 +1054,6 @@ def test_reply_download_generates_checksum(journalist_app,
     assert not mock_add_checksum.called
 
 
-def test_revoke_token(journalist_app, test_journo, journalist_api_token):
```
lsd-cat, line 1092: This test should be replaced with a similar one but leveraging the new logout mechanism.

securedrop/journalist_app/utils.py
```diff
@@ -507,40 +508,17 @@ def serve_file_with_etag(db_obj: Union[Reply, Submission]) -> flask.Response:
...
-def revoke_token(user: Journalist, auth_token: str) -> None:
-    try:
-        revoked_token = RevokedToken(token=auth_token, journalist_id=user.id)
-        db.session.add(revoked_token)
-        db.session.commit()
-    except IntegrityError as e:
-        db.session.rollback()
-        if "UNIQUE constraint failed: revoked_tokens.token" not in str(e):
-            raise e
+def logout_all() -> None:
```
lsd-cat: This is not called anywhere yet, but could be called by a button in the admin interface to logout all users!

securedrop/journalist_app/utils.py
```diff
@@ -507,40 +508,17 @@ def serve_file_with_etag(db_obj: Union[Reply, Submission]) -> flask.Response:
...
-
-    for revoked_token in revoked_tokens:
-        if Journalist.validate_token_is_not_expired_or_invalid(revoked_token.token):
-            pass  # The token has not expired, we must keep in the revoked token table.
-        else:
-            # The token is no longer valid, remove from the revoked token table.
-            db.session.delete(revoked_token)
-
-    db.session.commit()
+def logout_user(uid: int) -> None:
```
lsd-cat: This is user for logging out journalists when the passphrase is changed instead o using the nonce.

securedrop/journalist_app/admin.py
```diff
@@ -334,10 +334,8 @@ def new_password(user_id: int) -> werkzeug.Response:
             abort(404)
 
         password = request.form.get('password')
+        logout_user(user_id)
         if set_diceware_password(user, password) is not False:
-            if user.last_token is not None:
-                revoke_token(user, user.last_token)
```
lsd-cat, line 370: This was probably a bug since it was calling the JWT revocation function on journalists OTP, probably due to the common usage of 'token'.

securedrop/journalist_app/admin.py
```diff
@@ -334,10 +334,8 @@ def new_password(user_id: int) -> werkzeug.Response:
             abort(404)
 
         password = request.form.get('password')
+        logout_user(user_id)
```
lsd-cat: This should be placed two lines below!

securedrop/journalist_app/api.py
```diff
@@ -115,12 +85,10 @@ def get_token() -> Tuple[flask.Response, int]:
...
         try:
             journalist = Journalist.login(username, passphrase, one_time_code)
-            token_expiry = datetime.now(timezone.utc) + timedelta(
-                seconds=TOKEN_EXPIRATION_MINS * 60)
 
             response = jsonify({
-                'token': journalist.generate_api_token(expiration=TOKEN_EXPIRATION_MINS * 60),
-                'expiration': token_expiry,
+                'token': session.get_token(),
+                'expiration': session.get_lifetime(),
```
lsd-cat: How do we tell the client the new renew mechanism? Which expiration do we want to present here?

securedrop/journalist_app/main.py
```diff
@@ -42,16 +42,14 @@ def login() -> Union[str, werkzeug.Response]:
...
                 return redirect(url_for('main.index'))
 
         return render_template("login.html")
 
     @view.route('/logout')
     def logout() -> werkzeug.Response:
-        session.pop('uid', None)
-        session.pop('expires', None)
-        session.pop('nonce', None)
+        flash(session.destroy())
```
lsd-cat: Clearly flash here was left when debugging and should just be `session.destroy()`.

securedrop/journalist_app/__init__.py
```diff
@@ -167,8 +146,9 @@ def setup_g() -> 'Optional[Response]':
             app.logger.error("Site logo not found.")
 
         if request.path.split('/')[1] == 'api':
-            pass  # We use the @token_required decorator for the API endpoints
-        else:  # We are not using the API
+            if request.endpoint not in _insecure_api_views and not logged_in():
```
legoktm: These if statements can be joined with an `and` I think

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,182 @@
...
+        CallbackDict.__init__(self, initial, on_update)
+        self.sid = sid
+        self.token = None
+        self.is_api = False
+        self.to_destroy = False
+        self.to_regenerate = False
+        self.modified = False
+
+
+class RedisSession(ServerSideSession):
```
legoktm: Does this need to be a separate class given that we will always be using redis? Can it be merged with ServerSideSession (or maybe the other way around)?

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def _generate_sid(self) -> 'str':
+        return token_urlsafe(32)
+
+    def _get_signer(self, app: 'Flask') -> 'URLSafeTimedSerializer':
+        return URLSafeTimedSerializer(app.secret_key,
+                                      salt=self.salt)
+
+
+class RedisSessionInterface(SessionInterface):
```
legoktm: Same comment about merging with the parent.

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,182 @@
...
+        self.new = False
+        self.has_same_site_capability = hasattr(self, "get_cookie_samesite")
+
+    def open_session(self, app: 'Flask', request: 'Request') -> 'Optional[SessionMixin]':
+        '''This function is called by the flask session interface at the
+        beginning of each request.
+        '''
+        is_api = (request.path.split('/')[1] == 'api')
+
+        def new_session(is_api: 'bool' = False) -> 'SessionMixin':
```
legoktm: Why not make this a normal function on the RedisSessionInterface class? If it should be private, just prefix it with an underscore.

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,182 @@
...
+            sid = self._generate_sid()
+            session = self.session_class(sid=sid)
+            session.token = self._get_signer(app).dumps(sid)
+            session.new = True
+            session.is_api = is_api
+            return session
+
+        if not is_api:
```
legoktm: I would flip this just for easier readability, so it's `if is_api:`...

securedrop/journalist_app/utils.py
legoktm: Could this be moved into the rest of the session-related code somehow? Just so all the related logic stays together...

securedrop/alembic/versions/c5a02eb52f2d_dropped_session_nonce_from_journalist_.py
```diff
@@ -0,0 +1,41 @@
...
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table('journalists', schema=None) as batch_op:
+        batch_op.add_column(sa.Column('session_nonce', sa.INTEGER(), nullable=False))
+
+    op.create_table('revoked_tokens',
+    sa.Column('id', sa.INTEGER(), nullable=False),
```
legoktm: nit: please fix the indentation here

securedrop/alembic/versions/c5a02eb52f2d_dropped_session_nonce_from_journalist_.py
```diff
@@ -0,0 +1,41 @@
+"""dropped session_nonce from journalist table and revoked tokens table due to new session implementation
```
legoktm, line 1: This schema change needs a test, see https://docs.securedrop.org/en/latest/development/database_migrations.html#unit-testing-migrations

For the dropped table, you can try a plain `SELECT * FROM revoked_tokens` and verify it raises a SQL error (there should be some examples of that in other tests)

legoktm (dismissed):
As a general comment, you don't need to wrap types with quotes to turn them into strings, 95% just using the plain type is fine. They only need to be turned into strings if there's some recursive loading going on (like in models.py).

Other than that, I don't see any mypy failures, the one in CI is from pylint.

legoktm:
re, mypy, the biggest cause of warnings is save_session types session as a SessionMixin rather than RedisSession. Switching that gets rid of most of the errors. There are also (at least) 2 cases where a field is typed as `Optional[T]` so we need to check it's not null first. Here's my diff:
```diff
diff --git a/securedrop/journalist_app/sessions.py b/securedrop/journalist_app/sessions.py
index 3811ec687..a202cbc61 100644
--- a/securedrop/journalist_app/sessions.py
+++ b/securedrop/journalist_app/sessions.py
@@ -18,7 +18,7 @@ class ServerSideSession(CallbackDict, SessionMixin):
             self.modified = True
         CallbackDict.__init__(self, initial, on_update)
         self.sid = sid
-        self.token = None
+        self.token: Optional[str] = None
         self.is_api = False
         self.to_destroy = False
         self.to_regenerate = False
@@ -45,6 +45,8 @@ class SessionInterface(FlaskSessionInterface):
         return token_urlsafe(32)
 
     def _get_signer(self, app: 'Flask') -> 'URLSafeTimedSerializer':
+        if not app.secret_key:
+            raise RuntimeError('No secret key set')
         return URLSafeTimedSerializer(app.secret_key,
                                       salt=self.salt)
 
@@ -102,6 +104,8 @@ class RedisSessionInterface(SessionInterface):
             sid = self._get_signer(app).loads(sid)
         except BadSignature:
             return new_session(is_api)
+        if not sid:
+            return new_session(is_api)
 
         val = self.redis.get(self.key_prefix + sid)
         if val is not None:
@@ -112,7 +116,7 @@ class RedisSessionInterface(SessionInterface):
                 return new_session(is_api)
         return new_session(is_api)
 
-    def save_session(self, app: 'Flask', session: 'SessionMixin', response: 'Response') -> 'None':
+    def save_session(self, app: 'Flask', session: 'RedisSession', response: 'Response') -> 'None':
         '''This is called at the end of each request, just
         before sending the response.
         '''
```

That leaves me with:
```
securedrop/journalist_app/sessions.py:51: error: "SessionInterface" has no attribute "salt"
securedrop/journalist_app/sessions.py:81: error: Return type "Optional[SessionMixin]" of "open_session" incompatible with return type "None" in supertype "SessionInterface"
securedrop/journalist_app/sessions.py:154: error: Argument 2 to "set_cookie" of "BaseResponse" has incompatible type "Optional[str]"; expected "str"
```

The first issue goes away if the interfaces are merged, like I suggested inline. The second is because open_session() is defined in the parent as returning nothing, so we should get rid of our returns too.

The final one has to do with `response.set_cookie(app.session_cookie_name, session.token,`, where token is defined as `Optional[str]`. Should token be typed as `str` instead?

securedrop/journalist_app/__init__.py
lsd-cat: I am not sure bout this one. If I join line 148 and 149 with an `and` then the app will always return a 403 weh using the API. But I agree that this dhould structured more clearly as it is crucial to understand authentication

securedrop/journalist_app/utils.py
lsd-cat: Where would you add it? I did not put it into `sessions.py` because I kept it to be just the `1:1` session wrapper with the native Flask one, but since the native one does not have a server side logout then it might make sense. Should we try there?

lsd-cat:
Implemented most required changes, tests still in progress, thanks for the help!

 
> The second is because open_session() is defined in the parent as returning nothing, so we should get rid of our returns too.

I am not sure how to resolve this one. Flask docs specifically says that `open_session()` should return a session object or None if impossible. https://tedboy.github.io/flask/interface_api.session_interface.html

Also here in the parent https://github.com/pallets/flask/blob/a03719b01076a5bfdc2c8f4024eda7b874614bc1/src/flask/sessions.py#L299 it indeed doe not return but it has the Optional return type and it is there just as a placeholder to extend. Do we have any other option?

legoktm:
> I am not sure how to resolve this one. Flask docs specifically says that `open_session()` should return a session object or None if impossible. https://tedboy.github.io/flask/interface_api.session_interface.html

The correct docs link is https://flask.palletsprojects.com/en/2.0.x/api/#flask.sessions.SessionInterface.open_session btw, but the return type seems to still be `Optional[SessionMixin]`. I'll try to poke at it, otherwise we can use `# type: ignore` to suppress mypy.

securedrop/journalist_app/__init__.py
```diff
@@ -119,39 +118,13 @@ def _handle_http_exception(
...
     @app.before_request
     def update_instance_config() -> None:
         InstanceConfig.get_default(refresh=True)
 
     @app.before_request
     def setup_g() -> 'Optional[Response]':
         """Store commonly used values in Flask's special g object"""
-        if 'expires' in session and datetime.now(timezone.utc) >= session['expires']:
-            session.clear()
-            flash(gettext('You have been logged out due to inactivity.'),
```
eloquence: As briefly discussed with @lsd-cat today: As I understand, we're no longer able to show specific error message because we no longer have information about the session after it expires. So the current behavior of the PR is to not show any error at all. Is that correct? If so, I would suggest that we may still want to show an error message if we can't validate a session present in the cookie, just a more generic one, to give a hint to the user that the behavior is expected (and to aid with troubleshooting).

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,232 @@
...
+    def get_user(self) -> Optional[Journalist]:
+        if hasattr(self, 'user'):
+            return self.user
+
+    def get_uid(self) -> Optional[int]:
+        if hasattr(self, 'uid'):
+            return self.uid
+
+    def set_uid(self, uid) -> None:
+        self.uid = uid
```
legoktm, line 64: Don't we need to clear `self.user` if we're changing the uid?

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def logged_in(self) -> bool:
+        if hasattr(self, 'uid') and self.uid is not None:
+            return True
+        else:
+            return False
+
+    def destroy(self) -> None:
+        self.uid = None
```
legoktm, line 78: Seems like we should clear `self.user` here too

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def get_lifetime(self) -> datetime:
+        return datetime.now(timezone.utc) + timedelta(seconds=self.lifetime)
+
+    def set_user(self) -> None:
+        if hasattr(self, 'uid') and self.uid is not None:
+            self.user = Journalist.query.get(self.uid)
+
+    def get_user(self) -> Optional[Journalist]:
+        if hasattr(self, 'user'):
```
legoktm: In general I don't like dynamic properties like this, where they might not exist, it makes static typing harder. Why not have the variable exist from the beginning, but set to `None`, so it's just typed as `Optional[Journalist]` and can be returned directly? 

Same applies to `self.uid` too.

securedrop/journalist_app/utils.py
legoktm: Yeah, I think just moving it into session.py would be a good start. It's directly related to the session code and accesses the session logic directly...

securedrop/journalist_app/__init__.py
legoktm: Right. And we have no knowledge whether the session was legit and expired, or just a bogus session.

We could do something like if there's no valid session, seeing if they tried to pass a session cookie in the request, and if so, flashing a generic error, "Your session has expired or is invalid. Please log in again."

lsd-cat: We have indeed a condition in which we could assume that the session was valid, but got destroyed server-side either due to timeout, user/admin password change or account deletion. The case is when the session cookie has a valid signature for the session_id, but then the session is not in Redis. Not sure if it is better to use this condition, or the more generic has a session cookie to display the message.

lsd-cat: Implemented that case here https://github.com/freedomofpress/securedrop/pull/6403/commits/22b409217d1cc3b298cba2a0dd7d4940601939cf
The only thing is that at that point in time we do not have knowledge of the locale anymore. So either we use always english, or we try to get the request accept-language and decide based only on that.

securedrop/journalist_app/__init__.py
```diff
@@ -17,10 +16,9 @@
 
 from db import db
 from journalist_app import account, admin, api, main, col
-from journalist_app.utils import (get_source, logged_in,
-                                  JournalistInterfaceSessionInterface,
-                                  cleanup_expired_revoked_tokens)
-from models import InstanceConfig, Journalist
+from journalist_app.sessions import Session, session
```
nabla-c0d3, line 16: A lot of things are now called "session", which makes it hard to understand the code in this PR. I would suggest clarifying the names, perhaps:

* Session -> SessionMiddleware (not sure my understanding is correct here)
* session -> journalist_app_session

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,278 @@
...
+        if not session.is_api and (session.new or session.to_regenerate):
+            response.headers.add('Vary', 'Cookie')
+            response.set_cookie(app.session_cookie_name, session.token,
+                                httponly=httponly, domain=domain, path=path,
+                                secure=secure, **conditional_cookie_kwargs)  # type: ignore
+
+
+class Session(object):
+
+    def __init__(self, app: Optional[Flask] = None) -> None:
```
nabla-c0d3: I didn't see a use case in this PR for having `app=None`. Why not have the argument defined as `app: Flask` ?

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,278 @@
...
+            self.redis.set(name=self.key_prefix + session.sid, value=val,
+                           ex=expires, xx=True)
+        if not session.is_api and (session.new or session.to_regenerate):
+            response.headers.add('Vary', 'Cookie')
+            response.set_cookie(app.session_cookie_name, session.token,
+                                httponly=httponly, domain=domain, path=path,
+                                secure=secure, **conditional_cookie_kwargs)  # type: ignore
+
+
+class Session(object):
```
nabla-c0d3: nit: you don't need to inherit from `object` in Python 3.

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,278 @@
...
+
+        session_interface = SessionInterface(
+            config['SESSION_LIFETIME'], config['SESSION_RENEW_COUNT'],
+            config['SESSION_REDIS'], config['SESSION_KEY_PREFIX'],
+            config['SESSION_SIGNER_SALT'], config['SESSION_HEADER_NAME'])
+
+        return session_interface
+
+
+def logout_user(uid: int, is_current: bool = True) -> None:
```
nabla-c0d3: I didn't understand what the `is_current` argument means/does - perhaps a more detailed name is needed?

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,278 @@
...
+
+    """Uses the Redis key-value store as a session backend.
+
+    :param redis: A ``redis.Redis`` instance.
+    :param key_prefix: A prefix that is added to all Redis store keys.
+    :param salt: Allows to set the signer salt from the calling interface
+    :param header_name: if use_header, set the header name to parse
+    """
+
+    def __init__(self, lifetime: int, renew_count: int, redis: Optional[Redis],
```
nabla-c0d3: I would suggest requiring the `redis` parameter (ie. not having it Optional) for better separation of concerns.

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,278 @@
...
+            config['SESSION_REDIS'], config['SESSION_KEY_PREFIX'],
+            config['SESSION_SIGNER_SALT'], config['SESSION_HEADER_NAME'])
+
+        return session_interface
+
+
+def logout_user(uid: int, is_current: bool = True) -> None:
+    redis = Redis()
+    for key in (redis.keys(app.config['SESSION_KEY_PREFIX'] + "*") +
+                redis.keys("api_" + app.config['SESSION_KEY_PREFIX'] + "*")):
```
nabla-c0d3: This is a little bit hard to read - I would suggest generating the key first (perhaps using f-strings), and then looking for it in redis.

securedrop/tests/test_journalist_session.py
```diff
@@ -0,0 +1,305 @@
+# -*- coding: utf-8 -*-
```
nabla-c0d3, line 1: The tests don't have a lot of comments so they're a bit hard to follow/understand. I would suggest using the Given/When/Then comments style (example at https://github.com/freedomofpress/securedrop/blob/develop/securedrop/tests/test_passphrases.py)

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,278 @@
+import typing
```
nabla-c0d3, line 1: I would suggest running the [black code formatter](https://github.com/psf/black) on this file (since it's new) to facilitate a future migration to black.

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,278 @@
...
+        self.serializer = session_json_serializer
+        if redis is None:
+            redis = Redis()
+        self.redis = redis
+        self.lifetime = lifetime
+        self.renew_count = renew_count
+        self.key_prefix = key_prefix
+        self.api_key_prefix = 'api_' + key_prefix
+        self.salt = salt
+        self.api_salt = 'api_' + salt
```
nabla-c0d3: Is there a use case for the salt and prefix to be different for API sessions? If not it might be unneeded complexity.

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,278 @@
...
+                if len(split) != 2 or split[0] != 'Token':
+                    return self._new_session(is_api)
+                sid: Optional[str] = split[1]
+            else:
+                return self._new_session(is_api)
+        else:
+            sid = request.cookies.get(app.session_cookie_name)
+        if sid:
+            try:
+                sid = self._get_signer(app).loads(sid)
```
nabla-c0d3, line 153: Just to confirm: the max_age argument does not need to be supplied to the loads() call here, to reject expired signatures?

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,278 @@
...
+class SessionInterface(FlaskSessionInterface):
+
+    def _generate_sid(self) -> str:
+        return token_urlsafe(32)
+
+    def _get_signer(self, app: Flask) -> URLSafeTimedSerializer:
+        if not app.secret_key:
+            raise RuntimeError('No secret key set')
+        return URLSafeTimedSerializer(app.secret_key,
+                                      salt=self.salt)
```
nabla-c0d3: nit: the two methods above can be a `@staticmethod` as `self` is not used. It would also allow doing `SessionInterface._get_signer(app)` in the tests to re-use the implementation here.

nabla-c0d3 (commented):
I did a quick review - I don't have all the context (nor the Flask session mgmt knowledge) so it's really just a few thoughts.

securedrop/journalist_app/sessions.py
lsd-cat: In this case yes, I would say that we do not want to mix the SecureDrop Client sessions with the Jounalist Interface sessions even tough they originate from the same credentials. If we drop this, then an API token could be set in a cookie and would work there. While there are no obvious security risks that can I image now, this is still something unwanted and generally unexpected in most applications. In the future, it could also be useful to show active sessions in the admin page for instance.

securedrop/journalist_app/sessions.py
lsd-cat: In my opinion it would not add any security as the session gets destroyed server side on expiration and thus regardless of the signature the signed session identifier in the cookie does not point to anything meaningful anymore. Also there are negligible chances for session identifiers to be reused ever.

securedrop/tests/test_journalist_session.py
lsd-cat: This commit adds a description of these tests following your suggestion  https://github.com/freedomofpress/securedrop/pull/6403/commits/4c0794f6832725f6c9429451c016cf04bd170993

nabla-c0d3: It looks great! Much easier to understand what each test is about 👍

securedrop/journalist_app/sessions.py
lsd-cat: `is_current` here is used to tell the function whether the current user is being logged out, or another user is being logged out. As an example, when an admin change a journalist password, then such user is logged out, but in that case we do not want to remove the `uid` from the admin user session, logging him out. It is probably not needed though, as self logout is handed differently (via `destroy()`). I'll double check and clean it up in case!

ghost:
I'm not seeing any issues, so +1 for me.

securedrop/journalist_app/__init__.py
legoktm: That's mostly correct. I like the idea of importing the `Session` class as something like `SessionProvider`. I would like to keep the other one as `session` since it's supposed to be equivalent to `from flask import session`, so aside from the different import line, someone reading the code should be able to understand it as mostly equivalent to flask's built-in session stuff.

legoktm:
Did a quick skim through, I think we're mostly ready to move this forward?

* There are still a few unaddressed comments from @nabla-c0d3 that we should get to.
* And then we need to squash this down somehow.
* I explicitly listed some bugs that will be fixed by this as well.

zenmonkeykstop:
Does this actually close #204? That issue seems to relate only to the source interface. (Altho as an aside, I do like @garrettr's option no.2 [here](https://github.com/freedomofpress/securedrop/issues/204#issuecomment-32001071) as a way of removing the codename from SI cookies altogether without storing anything server-side. Definitely more feasible than a PAKE-type solution.)

securedrop/journalist_app/sessions.py
lsd-cat: This has been addressed in the recent commits and follows the addition of black to all the securedrop codebase.

securedrop/journalist_app/sessions.py
lsd-cat: Agreed, the change will be reflected in the next commit.

lsd-cat:
As for https://github.com/freedomofpress/securedrop/issues/204 that issue is strictly related to the source interface which is unaffected by the changes here. Although technically the same problem existed in the journalist interface, the issue was very minimal as the session contained mostly just the journalist id and I see no security benefit in hiding it in general, even though now it will be stored server side anyway.

For context, this change affects only the JI for a multitude of reasons: first it unifies the auth/session scheme for both the frontend and the API. Second, we do not want to have the code names stored server side on redis on purpose, even if ephemeral.

I also did another review of the logout mechanism following @nabla-c0d3 comments. There are a few things that needs to be changed: for example as of now if a user change its own password, only his session will be destroyed and not the ones of others logged with the same credentials. That mass deletion will happen only if the admin force change a user password through the `admin/edit endpoint`. Furthermore, since when an admin tries to change its password from the admin panel, it gets redirected anyway to the `account` password change endpoint, currently the `logout_user()` function is not invoked at all in that case. I'll try to address this in the next days and add a relevant test for this since apparently we have a wanted behavior but do not test for it.

securedrop/journalist_app/sessions.py
lsd-cat: Dropped the `is_current` parameter and the corresponding code paths as there were simpler way to achieve the same outcome.

securedrop/journalist_app/sessions.py
lsd-cat: Done in https://github.com/freedomofpress/securedrop/pull/6403/commits/a3a238fb553ec0d9954b914afd985f06b5aa126f

securedrop/journalist_app/sessions.py
lsd-cat: Agreed, implemented in https://github.com/freedomofpress/securedrop/pull/6403/commits/a3a238fb553ec0d9954b914afd985f06b5aa126f

lsd-cat:
Test plan moved to the top comment.

zenmonkeykstop:
The failing tests seem unrelated (getting errors retrieving the release pubkey from openpgp.org), rerunning.

legoktm:
We should remove the "Rebase:" prefix from the commit messages, otherwise I have no other comments!

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,283 @@
...
+                # create the new session already, so we can pass along messages and locale
+                session = self._new_session(False, initial=initial)
+        expires = self.redis.ttl(name=self.key_prefix + session.sid)
+        if session.new:
+            session["renew_count"] = self.renew_count
+            expires = self.lifetime
+        else:
+            if expires < (30 * 60) and session["renew_count"] > 0:
+                session["renew_count"] -= 1
+                expires += self.lifetime
```
zenmonkeykstop, line 196: Should this be `expires = self.lifetime`? Otherwise a session lifetime can be set up to `7200+1800` sec which is over the limit in the spec.

zenmonkeykstop:
Performed upgrade testing as follows:
- set up 2.4.2 prod vms
-  built packages off this branch and installed them on prod using the upgrade test procedure
- [x] upgrade was successful and db migration worked.

Tested session expiry using `redis-cli {keys|get|ttl|expire}`. There are a couple of issues with session lifetimes:
- sessions persist across server reboots, as there's nothing stopping the keys being written to disk (our redis.conf is pretty much the default, so that would be tunable there but require provisioning logic). This seems counter to the test plan's stated `Reboot of the server/restart of redis logs destroy all sessions`
- when a session is extended in the last 30 minutes of its lifetime, the extension is for 2hrs+the remainder. The feature description would seem to indicate that renewal should be just for 2hrs.

securedrop/journalist_app/sessions.py
lsd-cat: I think that the spec is badly worded and we should keep this behavior so that the overall duration is equal for every session regardless of when the renew happens. It would be easier to debug and maintain in the future, especially considering that the client will renew almost immediately when it is possible.

lsd-cat:
> 
> Tested session expiry using `redis-cli {keys|get|ttl|expire}`. There are a couple of issues with session lifetimes:
> 
>     * sessions persist across server reboots, as there's nothing stopping the keys being written to disk (our redis.conf is pretty much the default, so that would be tunable there but require provisioning logic). This seems counter to the test plan's stated `Reboot of the server/restart of redis logs destroy all sessions`

The non persistance in my testing environment was due to container themselves being not persistant, even though Redis in the early days used not to store snapshots by default, it indeed does now. Although the persistance of sessions is counterintuitive per se (most systems do not), it is not a major issue since the payload contains only the user id, the csrf token, the renewal count and the ttl. I'll amend the test plan to reflect this.

> 
>     * when a session is extended in the last 30 minutes of its lifetime, the extension is for 2hrs+the remainder. The feature description would seem to indicate that renewal should be just for 2hrs.

As noted above, I will also amend the description to properly reflect this behavior, as we concluded during the development phase that the currently implemented is the intended on.

securedrop/config.py.example
```diff
@@ -17,6 +17,10 @@ class SourceInterfaceFlaskConfig(FlaskConfig):
 class JournalistInterfaceFlaskConfig(FlaskConfig):
     SECRET_KEY = '{{ journalist_secret_key.stdout }}'
     SESSION_COOKIE_NAME = "js"
+    SESSION_SIGNER_SALT = "js_session"
```
zenmonkeykstop: New config.py variables should not be expected to be present on existing instances, as config.py is created at install time and rarely touched afterward.

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,283 @@
...
+            config["SESSION_SIGNER_SALT"],
+            config["SESSION_HEADER_NAME"],
+        )
+
+        return session_interface
+
+
+def logout_user(uid: int) -> None:
+    redis = Redis()
+    for key in redis.keys(app.config["SESSION_KEY_PREFIX"] + "*") + redis.keys(
```
zenmonkeykstop: This config key may not be set on existing instances that were upgraded. Calling this function causes a 500 error with the following example stack trace: 
```
[Fri Sep 23 18:35:04.907262 2022] [wsgi:error] [pid 687:tid 118508390467328] [remote 127.0.0.1:60948] E
RROR:journalist_app:Exception on /account/new-password [POST]                                          
[Fri Sep 23 18:35:04.907331 2022] [wsgi:error] [pid 687:tid 118508390467328] [remote 127.0.0.1:60948] T
raceback (most recent call last):                                                                      
[Fri Sep 23 18:35:04.907354 2022] [wsgi:error] [pid 687:tid 118508390467328] [remote 127.0.0.1:60948]  
 File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 2073, in wsgi_app
[Fri Sep 23 18:35:04.907370 2022] [wsgi:error] [pid 687:tid 118508390467328] [remote 127.0.0.1:60948]  
[Fri Sep 23 18:35:04.907384 2022] [wsgi:error] [pid 687:tid 118508390467328] [remote 127.0.0.1:60948]  
 File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 1518, in full_dis
patch_request                                                                                          
[Fri Sep 23 18:35:04.907414 2022] [wsgi:error] [pid 687:tid 118508390467328] [remote 127.0.0.1:60948]  
[Fri Sep 23 18:35:04.907428 2022] [wsgi:error] [pid 687:tid 118508390467328] [remote 127.0.0.1:60948]   File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 1516, in full_dispatch_request            
[Fri Sep 23 18:35:04.907442 2022] [wsgi:error] [pid 687:tid 118508390467328] [remote 127.0.0.1:60948]     rv = self.dispatch_request()                    
[Fri Sep 23 18:35:04.907454 2022] [wsgi:error] [pid 687:tid 118508390467328] [remote 127.0.0.1:60948]   File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 1502, in dispatch_request                 
[Fri Sep 23 18:35:04.907467 2022] [wsgi:error] [pid 687:tid 118508390467328] [remote 127.0.0.1:60948]     return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)                        
[Fri Sep 23 18:35:04.907478 2022] [wsgi:error] [pid 687:tid 118508390467328] [remote 127.0.0.1:60948]   File "/var/www/securedrop/journalist_app/account.py", line 46, in new_password                        
[Fri Sep 23 18:35:04.907486 2022] [wsgi:error] [pid 687:tid 118508390467328] [remote 127.0.0.1:60948]     logout_user(user.id)  
[Fri Sep 23 18:35:04.907498 2022] [wsgi:error] [pid 687:tid 118508390467328] [remote 127.0.0.1:60948]
 File "/var/www/securedrop/journalist_app/sessions.py", line 262, in logout_user                       
[Fri Sep 23 18:35:04.907511 2022] [wsgi:error] [pid 687:tid 118508390467328] [remote 127.0.0.1:60948] KeyError: 'SESSION_KEY_PREFIX'  
```

securedrop/journalist_app/sessions.py
```diff
@@ -0,0 +1,283 @@
...
+        found = redis.get(key)
+        if found:
+            sess = session_json_serializer.loads(found.decode("utf-8"))
+            if "uid" in sess and sess["uid"] == uid:
+                redis.delete(key)
+
+
+def logout_all() -> None:
+    redis = Redis()
+    for key in redis.keys(app.config["SESSION_KEY_PREFIX"] + "*") + redis.keys(
```
zenmonkeykstop: similarly as for logout_user(), this will fail with a 500 error

zenmonkeykstop (dismissed):

securedrop/journalist_app/sessions.py
lsd-cat: https://github.com/freedomofpress/securedrop/pull/6403/commits/64d40726c8f611b137dcfbdb76caf9e6f955bfa4 removes that function as it is never used currently.

securedrop/journalist_app/sessions.py
lsd-cat: As `logout_user` was the only function outside the class it would thus not fallback on the default values. Fix is to move it in the same context as the other functions that require `self.key_prefix` and `self.redis`. It is also cleaner since we do not reinstantiate the redis object too. https://github.com/freedomofpress/securedrop/pull/6403/commits/64d40726c8f611b137dcfbdb76caf9e6f955bfa4

zenmonkeykstop (approved):
Test plan looking good!

## Migration
- [x] Database migration work (deletes the revoked token table)
- [x] Token cleanup procedures do not run/exists
- [x] Sessions persist across reboot (as per default redis config)

## Security

- [x] Session ids are random
- [ ] Sessions id are signed **not checked**

### Invalid session id (non existant/wrong signature) gets ignored

- Login to the JI
- Tamper the cookie signature
- [x] Confirm that user is logged out

## Functional - JI
### Journalists can Login

- Login to the JI
- [x] Confirm that the user is logged

### Sessions of multiple users are separate
- Login to the JI as user X in browser A
- Login to the JI as user Y in browser B
- [x] Confirm that the two users navigation does not interfere

### Sessions are renewed in accordance with the new expiration/renewal policy
- Login to the JI
- Wait 2 hours (or use `redis-cli` to expire the session key...)
- [x] Confirm that user is logged out automatically
- [x] Confirm that an error message is shown

- Login to the JI
- Wait between 1.5h and 2h and refresh (or fake it with expires as above)
- Wait 1h (same deal)
- [x] Confirm that user is still logged in

### Logout works

- Login to the JI as user X in browser A
- Login to the JI as user Y in browser B
- Logout from user X in browser A
- [x] Confirm that user X is logged out
- [x] Confirm that user Y is still logged in

### Deleting a user destroy all sessions of the user

- Login to the JI as admin X in browser A
- Login to the JI as admin/user Y in browser B
- From admin X in browser A, delete user Y
- [x] Confirm that user Y in browser B is logged out

### Changing a user password destroy all sessions of the user 

- Login to the JI as admin X in browser A
- Login to the JI as admin/user Y in browser B
- From admin X in browser A, change password of user Y
-[x] Confirm that user Y in browser B is logged out

- Login to the JI as admin X
- Change admin X password
- [x] Confirm that admin X is logged out on password change

## API (tested via client)
### Token are issues on succesful API login
- Login to the /token endpoint
- [x] Confirm that the login is succesful
- Use the token for an authenticated operation
- [x] Verify that the authenticated operation accepts the token
- Send a request to the logout endpoint
- [x] Confirm that a message fo succesful logout is returned
- Retry the authenticated operation.
- - [x] Confirm that the operation now requires a valid token

LGTM!

PR #6407, opened 2022-04-19
Add a container for the "demo" deployment

maeve-fpf:
## Status

Ready for review

## Description of Changes

Related to https://github.com/freedomofpress/infrastructure/issues/3664

Over in infra, we'd like to deploy the demo as containers directly, instead of doing a clone and `make dev` to bring one up. We also want to get the demo "landing page" (links and TOTP generator) out of the infrastructure repo and into something public.

This PR adds two containers, one for the landing page and one that is a slimmed-down version of the existing `focal/python3` Dockerfile, with only what the demo deployment needs. There are a few script fixes for minor issues that I found running the container in a k8s Pod. Currently, I'm doing that with a read-write root filesystem, but I think I can follow up with some more changes that will remove the need to move things around with `sudo` at runtime and let us run it read-only.

The JS check for a non-onion connection has been extended to cover the possible hostnames that this will be deployed to, instead of only localhost. It's a little unwieldy, but I didn't have a better idea.

## Testing

To build the image: `docker build -t quay.io/freedomofpress/securedrop-demo-focal-py3 -f securedrop/dockerfiles/focal/python3/DemoDockerfile .`

To run: `docker run -p 8080:8080 -p 8081:8081 -it --rm securedrop-demo-focal-py3:latest`

You should see the source and journalist interfaces on ports 8080-8081, like if you ran `make dev`.

In practice this Dockerfile will be built by CI and run via the manifests in k8s-configs. For developers, this version of the container is not interesting, and `make dev` is still the way to start local containers.

The script changes should not affect normal `make dev` usage - if they break something please let me know.

## Deployment

Will be handled by infra.

## Checklist

I think none of these apply.

Choose one of the following:


gonzalo-bulnes (commented):
I had a superficial look, the changes are clear 👍 thanks @maeve-fpf for splitting the commits the way you did!

I didn't read the Dockerfiles into enough details, and I'll let the rest of the team weight in.

Question @maeve-fpf: these Dockerfiles seems to be exclusively used for "demo" purposes, and are deployed by infra... can you think of anything   you'd like us to test / review from an application point of view?

Subsidiary question: in what occasions would this setup have to change? (Just so that we make a note as a dev team and keep in mind to bring you in the loop before you notice that a deployment has failed 😉)

**Edit**: Following up on the question above: do you think it's worth including a test deployment of the containers in CI? (E.g. test that the containers start and ports are open?)

maeve-fpf:
@gonzalo-bulnes I think the most likely thing to need a change is:

- `DemoDockerfile`, if something in `Dockerfile` changes, like installation steps or the base image (this is to some extent duplicated code)
- the landing page Dockerfile, which is pretty trivial and probably will only have the base image updated for new versions/security fixes of nginx

I don't think there's anything application-specific to test, but you can go to https://staging-demo.securedrop.org/ now to use the deployment (I'll need to follow up to change the links to remove `staging-`).

As for the CI question - I think it might be more useful to add Icinga monitoring of the demo URLs once they are being continuously deployed, to test the rest of the deployment path as well, and generate alerts if they are down for any other reason.

devops/demo/landing-page/Dockerfile
```diff
@@ -0,0 +1,10 @@
+# sha256 as of 2022-02-16
+FROM nginx:mainline-alpine@sha256:050385609d832fae11b007fbbfba77d0bba12bf72bc0dca0ac03e09b1998580f
```
legoktm, line 2: Is there a reason we're using alpine here instead of a Debian/Ubuntu base like we do for everything else?

securedrop/static/js/source.js
```diff
@@ -124,9 +124,9 @@ function suggestTor() {
 
 function checkClearnet() {
   let url = new URL(location.href);
-  // Allow localhost for development
-  let localhost = ["127.0.0.1", "localhost"];
-  if (localhost.indexOf(url.hostname) === -1 && !url.host.endsWith(".onion")) {
+  // Allow localhost for development, and official demo hostnames for demo
```
legoktm: Can we do what the demo does currently, and patch in these hostnames rather than putting them in the source that goes out to all SecureDrops? I think it would be relatively straightforward to do when building the container?

securedrop/dockerfiles/focal/python3/DemoDockerfile
```diff
@@ -0,0 +1,32 @@
...
+                       gettext paxctl enchant libffi-dev sqlite3 gettext sudo \
+                       libasound2 libdbus-glib-1-2 libgtk2.0-0 libfontconfig1 libxrender1 \
+                       libcairo-gobject2 libgtk-3-0 libstartup-notification0 basez
+
+RUN gem install sass -v 3.4.23
+
+COPY . /opt/securedrop
+RUN python3 -m venv /opt/venvs/securedrop-app-code && \
+    /opt/venvs/securedrop-app-code/bin/pip3 install --no-deps --require-hashes -r /opt/securedrop/securedrop/requirements/python3/docker-requirements.txt && \
+    /opt/venvs/securedrop-app-code/bin/pip3 install --no-deps --require-hashes -r /opt/securedrop/securedrop/requirements/python3/test-requirements.txt && \
```
legoktm: What are the test requirements needed for?

legoktm (changes requested):
Yaaay!!

Assuming we didn't write jsOTP.js ourselves, it would be good to note the license/attribution for that file and the firasans-book font in the assets/ directory (or elsewhere).

re: CI, agreed that starting with icinga should be fine. If it turns out we are regularly breaking the demo site (which I think is unlikely), then it would make sense to add to part of CI checks.

devops/demo/landing-page/Dockerfile
maeve-fpf: Only that we are using an Alpine base for the other nginx containers in the k8s environment (docs and HTTPS Everywhere).

securedrop/static/js/source.js
maeve-fpf: We'd need to wrap the build process somehow, which is difficult with the way we currently build web containers in CI (it uses something declarative instead of directly running a command that runs `docker build`).

If including this check in the code served to the client could depend on something in the server environment, that'd be easy, but it's a static file :-/ .

I can definitely remove the `staging-` versions, if the concern is that those are not publicly advertised (they're going to be torn down anyway).

maeve-fpf:
I put the licenses for browser-totp and Fira in the `landing-page` directory so that they don't end up as part of the nginx image. Is that acceptable?

securedrop/static/js/source.js
maeve-fpf: (I did go ahead and do that, so there's only one entry added.)

devops/demo/landing-page/Dockerfile
legoktm: Ack, makes sense then.

securedrop/dockerfiles/focal/python3/DemoDockerfile
```diff
@@ -0,0 +1,31 @@
+# ubuntu 20.04 image from 2021-09-21
+FROM ubuntu@sha256:3555f4996aea6be945ae1532fa377c88f4b3b9e6d93531f47af5d78a7d5e3761
+
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get -y install tzdata && \
+    apt-get install -y paxctl apache2-dev coreutils devscripts vim \
```
legoktm: paxctl and gettext are listed twice.

If we want to slim it down further, we can also remove devscripts, vim, curl, wget.

securedrop/static/js/source.js
legoktm: What I had in mind was using https://github.com/freedomofpress/infrastructure/blob/70ee349d3e5f0b570b8444d181768f4151835145/roles/sd-demo/files/0001-Disable-tor2web-redirect-for-localization-demo.patch, copying it into the container, and then applying it to `/opt/securedrop/` before continuing with the build process.

Even simpler would be `sed -i 's/"localhost"\];/"localhost", "demo-source.securedrop.org"];/' /opt/securedrop/securedrop/static/js/source.js`

I'm mostly proposing this since that what Conor had done earlier. But if you think we should just hardcode it in the source, I'm okay with that too if @zenmonkeykstop is onboard too.

legoktm:
The license info LGTM. I tested the container and ran it, I think this is good to go pending the thread about the .onion/clearnet warning.

securedrop/static/js/source.js
zenmonkeykstop: My pref would be the patch route if possible, just to keep the code shipped to prod clean.

maeve-fpf: OK, I've switched it to sed in the Dockerfile.

securedrop/dockerfiles/focal/python3/DemoDockerfile
maeve-fpf: Thanks, I removed the duplication of gettext and paxctl from the original Dockerfile too.

I wasn't able to remove apache2-dev (a Python module needs it at build time) or git (some other Python code calls out to it at startup), but I removed the rest of those.

securedrop/dockerfiles/focal/python3/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN apt-get update && apt-get install -y paxctl && \
     apt-get install -y apache2-dev coreutils devscripts vim \
                        python3-pip python3-all python3-venv virtualenv libpython3.8-dev libssl-dev \
                        gnupg2 ruby redis-server git xvfb curl wget \
-                       gettext paxctl x11vnc enchant libffi-dev sqlite3 gettext sudo \
+                       x11vnc enchant libffi-dev sqlite3 gettext sudo \
```
legoktm, line 15: Am I totally missing it or is paxctl not duplicated in this one and now missing?

conorsch: `paxctl` was only present in the dockerfile to support folks who were running grsecurity-patched kernels on their development workstations. These days I believe that number is zero, so it's OK in my view to remove paxctl from the container images entirely.

legoktm: Got it, perfect then!

legoktm (approved):
LGTM!

securedrop/dockerfiles/focal/python3/Dockerfile
maeve-fpf: For this file, currently it's still in the first line of the apt-get commands, by itself.

PR #6413, opened 2022-04-27
Delegate date strings formatting to datetime.isoformat()

### LINKED ISSUE #6257: API responses should include time-zone information, be valid ISO8061/RFC339
## Description

A [recent change][bug-introduced] aimed at including time zone information caused the API to respond with date strings in an invalid format. As a workaround, the time zone information was [removed][workaround] from the API responses. 
_For the sake of focus, I'll keep the details on how the issue happened in the *comments* section below._

While that is an effective workaround, it does result in more fragile communication between the SecureDrop server and potential clients.
Instead of communicating via date strings that do not include time zone information (but are ambiguous because of that), we should resume communicating via date strings that do include time zone information (i.e. that are unambiguous) and make sure that we:
1. use a valid date format (like we used to do, and can still do within existing constraints)
2. use UTC where possible for inter-service communication (like we used to do and have no reason to stop doing)

  [bug-introduced]: https://github.com/freedomofpress/securedrop/commit/16335661105bac0621e24d40c51d671677a331c2#diff-0814e8ec8e8bcae2e28501f8d8397a3199fbff361b132552d597e7e9696e88baR118-R122
  [workaround]: https://github.com/freedomofpress/securedrop/pull/6256

## Current Behavior

The API responses contain date strings with no time zone information, e.g. `2022-02-09T05:22:12.165773`
This time could be legitimately interpreted as 5 in the morning in Australia, or 5 in the morning in London.

To make things even less reliable, default behaviors vary: for example RFC3339 specifies that time zone information shall always be provided [[Section 4.3][tz-1]] [[Section 5.6][tz-2]], while ISO8061 specifies that the [local time zone could be assumed][iso-tz]. :sun_behind_rain_cloud:  Because of that, stripping time zone information is not a reliable long-term solution.

  [tz-1]: https://datatracker.ietf.org/doc/html/rfc3339#section-4.3
  [tz-2]: https://datatracker.ietf.org/doc/html/rfc3339#section-5.6
  [iso-tz]: https://en.wikipedia.org/wiki/ISO_8601#Local_time_(unqualified)
  
**Note**: The ISO documents are behind a pay wall, please bear with the Wikipedia links! :bear:

## Expected Behavior

The API responses contain date strings in the following format `2022-02-09T05:22:12.165773+00:00`.
- :green_apple: This string is valid when interpreted by both the RFC3339 and ISO8061 standards.
- :green_apple: This string is supported by the ISO8061 parser included in the `datetime` package (which is [rather limited][doc]).

  [doc]: https://docs.python.org/3/library/datetime.html#datetime.time.fromisoformat

### Why this format specifically?

1. The date strings contained in API responses should be unambiguous. And several formats can achieve that, for example `2022-02-09T05:22:12.165773Z` or `2022-02-09T05:22:12.165773+00:00` or `2022-02-09T15:52:12.165773+10:30`. All these examples describe the exact same moment, around 5 in the morning UTC.

1. For machine to machine communication, there is no need to communicate a _preference_ for any given local time representation, and UTC could (should?) be favored. (Local times are mostly useful for people, with the noteworthy exception of people troubleshooting logs, who benefit from consistent logging! UTC is a good "neutral ground" for that.)  So `2022-02-09T05:22:12.165773Z` or `2022-02-09T05:22:12.165773+00:00` are preferred.

1. Different standards interpret a zero time offset sightly differently. In RFC3339, `+00:00` expresses a preference for UTC and `-00:00` expresses the absence of preference. The negative zero offset `-00:00` is simply not valid in the context of ISO8061. The `+00:00` notation is valid in both standards, and expresses a preference for UTC. (So far, the `+00:00` and `Z` variants would be equally good, hold that thought.)

1. Additionally, the `datetime` parser only supports a limited subset of ISO8061 strings. I particular, `2022-02-09T05:22:12.165773Z` cannot be parsed by the `datetime.fromisoformat(date_string)` function. So `2022-02-09T05:22:12.165773+00:00` is a more convenient choice.
Because using existing `datetime` methods ensures consitency and could prevent similar bugs in the future, the `+00:00`-flavored format seems like a better choice.

### Summary of usage

```python
from datetime import datetime, timezone

# Datetime object MUST be created with time zone information:
d = datetime.now(timezone.utc)
d
# datetime.datetime(2022, 2, 9, 5, 9, 19, 592689, tzinfo=datetime.timezone.utc)

# Serialization:
date_string = d.isoformat()
# '2022-02-09T05:09:49.423016+00:00'

# This date_string is good to send to other services, across the network etc.

# Deserialization:
d = datetime.fromisoformat(date_string)
d
# datetime.datetime(2022, 2, 9, 5, 9, 19, 592689, tzinfo=datetime.timezone.utc)
```

### Timeline

I was able to trace the sequence of events that lead to this situation:

1. In 2018, a UTC time started being sent in the API response. Date-times were handled without timezone information internally, and UTC was specified during serialization. (https://github.com/freedomofpress/securedrop/commit/c324abe8e8a9a28a47244147a6dc80802b93315e) On the SDK side, those were de-serialized in a consistent manner.
The hard-coding of the time-zone information very likely responded to the limitations of the `datetime` package (e.g. [reference][so]), the code did internally assumed UTC and all ambiguity was cleared in communications. :heavy_check_mark: 

2. A few months ago, time zones started being handled internally. (https://github.com/freedomofpress/securedrop/commit/16335661105bac0621e24d40c51d671677a331c2#diff-0814e8ec8e8bcae2e28501f8d8397a3199fbff361b132552d597e7e9696e88baR118-R122) A side-effect of that was that the hard-coded UTC `Z` became redundant, and caused the API to respond with invalid (ISO 8061) date strings. IMHO that was an easy side-effect to miss, and hopefully we don't hard-code time format handling in the future :slightly_smiling_face: 
At that point, the SDK couldn't handle the invalid strings. :bug: https://github.com/freedomofpress/securedrop/issues/6255 and https://github.com/freedomofpress/securedrop-sdk/issues/171 (Note: I don't think that issue suggests the best solution, because `Z` is less well supported than `+00:00` in Python context, see above.)

3. Today a workaround was applied, https://github.com/freedomofpress/securedrop/pull/6256 which brings us here.

I hope that makes sense : )

  [so]: https://stackoverflow.com/questions/19654578/python-utc-datetime-objects-iso-format-doesnt-include-z-zulu-or-zero-offset/42777551#42777551
  
### Proposed plan

- [x] Open this issue describing the situation in details.
- [ ] Open a PR to ensure that the SecureDrop API responds with valid ISO8061/RFC3339 strings, which use the `+00:00` variant to ensure we let  `datetime` do the formatting and hopefully prevent future inconsistencies.
- [x] Open a PR to ensure that the SDK supports the time-zone information (`+00:00` in practice) in the responses it gets from the API and allows the SDK to handle some of the formats (valid or invalid) we've used in the past: with time zone (`Z`-flavored, `+00:00`-flavored), with twice the info `+00:00Z`, any other? https://github.com/freedomofpress/securedrop-sdk/pull/172
- [x] Upgrade the Client with the new SDK.

### Urgency

This is not urgent, since a workaround is in place.
The SDK should be updated first, then the Client should use it (**not a release blocker**) and finally the servers could be released. The SDK will accept inputs generously, so the server release doesn't need to be coordinated with the release of the Client as long as it happens last.

CC: @creviera

gonzalo-bulnes:
## Status

Ready for review

## Description of Changes

Closes #6257 (see **Proposed plan** section in the description).

- Remove the custom date string format (that used to be the only one supported by the SecureDrop Client)
- Use ISO8061 via `datetime.isoformat()` instead (should be easier to use consistently too)

## Testing

- [ ] Visual review: do these changes make sense given the context recapped below?

Then test here:
- [ ] Pay special attention to CI: do the tests changed here still pass?

And then check for surprises downstream:
- [ ] Run `make dev` here and `make regenerate-sdk-cassettes` in `freedomofpress/securedrop-client/client`.  It should pass.
- [ ] Run `make test` in `freedomofprses/securedrop-client/client`.  It should pass.

## Deployment

Once this change has been released, the try/except logic in <https://github.com/freedomofpress/securedrop-client/blob/7f7f60e67f7606708a6c9ef519c9657cd738be9e/client/securedrop_client/sdk/timestamps.py> can be removed in favor of just `datetime.fromisoformat()`.

### DIFF
--- a/securedrop/tests/test_journalist_api.py
+++ b/securedrop/tests/test_journalist_api.py
@@ -20,8 +20,7 @@ random.seed("◔ ⌣ ◔")
 
 def assert_valid_timestamp(timestamp: str) -> None:
     """verify the timestamp is encoded in the format we want"""
-    dt_format = "%Y-%m-%dT%H:%M:%S.%fZ"
-    assert timestamp == datetime.strptime(timestamp, dt_format).strftime(dt_format)
+    assert timestamp == datetime.fromisoformat(timestamp).isoformat()
 
 
 def test_unauthenticated_user_gets_all_endpoints(journalist_app):
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -20,8 +20,6 @@ from werkzeug.exceptions import HTTPException, default_exceptions
 
 _insecure_views = ["main.login", "static"]
 _insecure_api_views = ["api.get_token", "api.get_endpoints"]
-# Timezone-naive datetime format expected by SecureDrop Client
-API_DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
 
 
 def get_logo_url(app: Flask) -> str:
@@ -67,7 +65,7 @@ def create_app(config: SecureDropConfig) -> Flask:
 
         def default(self, obj: "Any") -> "Any":
             if isinstance(obj, datetime):
-                return obj.strftime(API_DATETIME_FORMAT)
+                return obj.isoformat()
             super().default(obj)
 
     app.json_encoder = JSONEncoder
--- a/securedrop/tests/test_journalist_session.py
+++ b/securedrop/tests/test_journalist_session.py
@@ -292,7 +292,7 @@ def test_session_api_login(journalist_app, test_journo, redis):
         # Then the expiration date returned in `get_token` response also conforms to the same rules
         assert (
             datetime.now(timezone.utc)
-            < datetime.strptime(resp.json["expiration"], "%Y-%m-%dT%H:%M:%S.%f%z")
+            < datetime.fromisoformat(resp.json["expiration"])
             < (
                 datetime.now(timezone.utc)
                 + timedelta(seconds=journalist_app.config["SESSION_LIFETIME"])


securedrop/tests/test_journalist_api.py
```diff
@@ -26,7 +26,7 @@
 
 def assert_valid_timestamp(timestamp: str) -> None:
     """verify the timestamp is encoded in the format we want"""
-    dt_format = "%Y-%m-%dT%H:%M:%S.%fZ"
+    dt_format = "%Y-%m-%dT%H:%M:%S.%f+00:00"
```
legoktm: Can we use `datetime.fromisoformat()` to match securedrop-sdk? Looking at https://github.com/freedomofpress/securedrop-sdk/blob/31ea11e3f9d8c243c0fbe2a19bbdd4f75bd32bd0/sdclientapi/timestamps.py

gonzalo-bulnes: Yes, we can! I left this one as-is because I like having distinct implementations for the tests and the code under test. I found that using the format string made particularly clear the expectation in the test file. But all that being said, I'd say as a `securedrop` repo maintainer it's your call, and I don't personally have strong feelings about it.

legoktm: Maybe both then? I would like to make sure we're at least testing with what the sdk is actually using, which is now `fromisoformat()`.

gonzalo-bulnes:
@legoktm I'm trying to gather again the context around this PR. It seems like one of the API responses contains timestamps without time zone information ([specific error in CI](https://app.circleci.com/pipelines/github/freedomofpress/securedrop/4046/workflows/50a888a5-dc5e-4439-8601-7055422fbc89/jobs/62222)). I believe that's a legit test suite error and that we want to fix the API response to conform to the new convention, does that seem right to you?

cfm:
**Tl;dr:**  This change does *not* have any side effects on time-zone support.  It simply standardizes us to `datetime.isoformat()` and should be either approved or closed based whether we currently think that standardization is worthwhile.

We decided in today's backlog-pruning session that this was worth revisiting in a time-box to see if we could finalize it.  Substantively all I've done is rebase this from `develop`.  But to recap the context here:

1. `securedrop` currently sends UTC timestamps in a custom `API_DATETIME_FORMAT`.
2. With this change, `securedrop` will send UTC timestamps exclusively in `datetime.isoformat()`.
3. `securedrop-client`, specifically the SDK, will already [parse](https://github.com/freedomofpress/securedrop-client/blob/7f7f60e67f7606708a6c9ef519c9657cd738be9e/client/securedrop_client/sdk/timestamps.py) either (1) or (2).  (The test plan above confirms this.)
4. After this change has been released, `securedrop-client` can drop the trial-and-error parsing in (3).

legoktm (approved):
LGTM; I tested that the current client can handle the new ISO timestamp format just fine.

gonzalo-bulnes:
Wow. That was some time ago. 😮

PR #6422, opened 2022-05-03
Viewport shouldn't always show a scrollbar

### LINKED ISSUE #6419: [develop] Viewport is scrollable without scrollable content
## Steps to Reproduce
1. Load https://demo-source.securedrop.org/ (built from latest `develop`) at a resolution where the entire page fits in the viewport

## Expected Behavior
No vertical scrollbar is displayed and there is no scrollable content
## Actual Behavior

Vertical scrollbar displayed and I can scroll down about a third of the page. This is repeatable to different degrees on other parts of the Source Interface

## Screenshot
Scrollable "Get Codename" page:

Screenshot from 2022-05-02 15-20-47

## Comments

- Behavior does not appear to be present on 2.3.1
- Behavior does not appear to be the result of the length of the language list
- I've not tested Journalist Interface behavior

eaon:
## Description of Changes

Because of a `body::after` element used to pre-load a couple of images was visually hidden but still part of the DOM, the element forced the viewport to always show a scrollbar.

I overlooked this visual degradation that was introduced by the most recent accessibility related changes to hide the element from screen-readers but forgot to position the element off-screen.

Fixes #6419

## Testing

* Check SI index page in full screen mode, it should *not* show a scrollbar.

### DIFF
--- a/securedrop/sass/source.sass
+++ b/securedrop/sass/source.sass
@@ -1169,6 +1169,7 @@ footer p
 // Preload some images we're likely to display when moving on from the index
 body::after
   position: absolute
+  top: -999px
   width: 0
   height: 0
   visibility: hidden


zenmonkeykstop (approved):
Tested in dev, extra space has been eliminated on all SI views.

PR #6441, opened 2022-05-10
Switches Keepass database template to Argon2d KDF

sts10:
Updates `tails_files/securedop-keepass.kdbx` KeePass database template file to use [Argon2d KDF](https://en.wikipedia.org/wiki/Argon2) (key derivation function), matching the new defaults of KeePassXC.

## Status

Ready for review 

## Description of Changes

I opened the KeePass database template file and, using KeePassXC, changed the security settings to match the defaults that KeePassXC 2.6.2 uses for a brand new database.

New KeePass database template security settings, including changing KDF to Argon2d

Most significantly, I changed the KDF to Argon2d, which KeePassXC lists as the recommended KDF.

## Testing

Testers could, while running Tails, clone down this repo and attempt to unlock and Save As... the new version of the template to ensure that works. I don't think there will be any issues. Tails 5 includes KeePassXC v 2.6.2, which can use the Argon2d KDF. ([Argon2d KDF was introduced in KeePassXC version 2.3.0](https://keepassxc.org/blog/2018-02-28-2.3-released/).) Testers may also want to verify that I entered the settings correctly, matching a brand-new KeePass database made with KeePassXC. 

An assumption: Users running older versions of Tails likely won't be using this KeePass database template file, and thus are not at risk of using a version of KeePassXC older than version 2.3.0, which would prevent them from being able to unlock the template due to it using the newer KDF.  

## Deployment

We may want to instruct existing users to change the KDFs of their KeePass databases (and run a 1 second benchmark), though that is a separate issue. 

For new installs, this shouldn't affect that installation process. The process for using the KeePass template should be unaffected by these changed settings.

## Checklist

Choose one of the following:

- [X] I would appreciate help with the documentation

Choose one of the following:

- [X] I would like someone else to do the diff review

This is my first PR to SecureDrop, so please let me know if there is more work for me to do!


sts10:
As the KeePassXC folks kindly [explained to me on Twitter](https://twitter.com/KeePassXC/status/1524142110055804934), here's why a user should prefer Argon2d over AES-KDF: 

> Argon2d is a CPU and Memory hard key derivation algorithm. AES-KDF is only CPU hard. This means you can use graphics cards or ASIC processors to break AES-KDF far easier than Argon2d.

sts10:
Now realizing that if the admin's KeePass database [will never have a non-trivial password or keyfile protecting it](https://docs.securedrop.org/en/stable/set_up_admin_tails.html#create-the-admin-passphrase-database), it doesn't matter what KDF it uses. And in fact, of the KDF options the latest versions of KeePassXC offers, we should likely prefer the older KDF for increased backwards compatibility (with KeePassX and KeePassXC versions older than 2.3). Closing this PR!

gonzalo-bulnes:
Thanks for this thread @sts10! I found it very insightful 💡

> Now realizing that if the admin's KeePass database [will never have a non-trivial password or keyfile protecting it](https://docs.securedrop.org/en/stable/set_up_admin_tails.html#create-the-admin-passphrase-database), it doesn't matter what KDF it uses.

Is that because the password is the weaker link in this case? (Or because it happens to be double-encrypted as part of the persistent volume? I'm guess that might defeat some of the attacks you mentionned on AES-KDF but I may be wrong.)

sts10:
Thanks so much for this comment! KDFs are super neat! 

I closed this because I assumed that most (all?) SecureDrop admins leave their KeePass database passphrase blank for convenience (and if they have a keyfile, the keyfile is just sitting right next to the database). This assumption is based on this line [in the docs](https://docs.securedrop.org/en/stable/set_up_admin_tails.html#create-the-admin-passphrase-database):

> Because your persistent volume is encrypted, setting up this additional password is not strictly required. It provides some additional protection, e.g., if a computer is left running, at the cost of convenience.

And thinking about it, it seems like a lot to ask for an admin to have one long passphrase memorized to get into the persistent storage, and then _another_ memorized for the KeePass database within that persistent storage. 

All that said: It's my understanding that switching the KDF algorithm to Argon2d and benchmarking it to 1 second (a single button click in KeePassXC) _does_ make the KeePass database more costly to attack (each passphrase guess takes more time), especially against graphics cards or ASIC processors attacks (see comment above). _But_ if the passphrase is blank, the KDF won't really slow an attacker down as I'm assuming a blank passphrase would be the attacker's first guess. 

The trade-off here is that switching the template database's KDF from AES-KDF to Argon2d means that KeePassX and versions of KeePassXC older than 2.3 will NOT be able to open the template database. This wide compatibility of the template database is [something that the docs currently state explicitly](https://docs.securedrop.org/en/stable/set_up_admin_tails.html#create-the-admin-passphrase-database), so I figure it has come up in the past: 

> Earlier versions of Tails used KeePassX instead of KeePassXC. The provided template is compatible with both.

So I think that's the tradeoff here: **Merging this PR gives those admins who choose to set a unique passphrase on their KeePass database a little more time against attackers, but we lose the aforementioned compatibility.** My hunch is that it's not worth the tradeoff, but I welcome comments! 

A 3rd option here would be to advise, in the docs, that admins open the KeePassXC security settings themselves and switch/confirm the KDF to Argon2d (or whatever is currently recommended) and then run the 1-second benchmark. But again, that only matters if they also set a non-trivial passphrase on the KeePass database.

gonzalo-bulnes:
Your reasonning makes a lot of sense to me, and I believe the broader compatibility is valuable in the field. I'll still ping @lsd-cat and @l3th3 who may have additional insights 🙂

PR #6444, opened 2022-05-12
Updated SI index page to reflow to single vertical column on small screens.

### LINKED ISSUE #6443: Source Interface index page displays incorrectly on resize to vertical layout
## Description

The existing SI design resizes to flow the logo and buttons into a single centered vertical column without visible spacers. In the redesign, the buttons are offset to the right, and the spacer on the right of the Get Started button remains visible.

Develop:
resize-develop

vs

release/2.4.0:
resize-240

## Steps to Reproduce

Please specify your environment if that is necessary to reproduce the bug (if in
doubt, include it).

## Expected Behavior

on resize, elements are centered in a vertical column

## Actual Behavior

see screenshot

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #6443 .

CSS update to restore the source interfaces index page's vertical layout on screens <= 768px 

## Testing

- spin up a dev instance from this branch and from the 2.4.0-rc1 tag
 - [ ] Compare with screen sizes at >768px and confirm no layout differences
 - [ ] Compare with screen sizes at <=768px and confirm that the version on this branch displays all sections in a single column using the full screen width and with no separator on the `First Submission` section.

## Deployment
n/a

Choose one of the following:

### DIFF
--- a/securedrop/sass/source.sass
+++ b/securedrop/sass/source.sass
@@ -297,8 +297,24 @@ header
       padding-top: 0
       min-height: unset
 
-    p
+    &> section:not(#locales-menu):not(#flashed)
+
+      width: 100%
+      padding-top: 10px
+      min-height: unset
+
+      &:first-child
+        @include ltr
+          border-right: 0
+        @include rtl
+          border-left: 0
+
+      &#return-visit
+        padding-left: 50px
+
+    p, hr
       width: auto
+      max-width: 500px
 
     button, .btn
       width: 80%
@@ -314,6 +330,10 @@ header
 
       div
         width: auto
+
+
+
+
 
 // Logo is a simple anchor on all pages except index
 header h1, header h1 a
--- a/securedrop/sass/source.sass
+++ b/securedrop/sass/source.sass
@@ -285,22 +285,15 @@ header
   html #upload .attachment input
     font-size: 130%
 
-  html #source-index > .content > main
+  // no html selector here because we use the ltr/rtl includes here
+  #source-index > .content > main
     display: block
     width: 100%
 
-    &> section
-      border: 0
+    &> section:not(#locales-menu):not(#flashed)
       width: auto
       margin: 0
-      padding-left: 30px
       padding-top: 0
-      min-height: unset
-
-    &> section:not(#locales-menu):not(#flashed)
-
-      width: 100%
-      padding-top: 10px
       min-height: unset
 
       &:first-child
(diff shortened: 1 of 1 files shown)


securedrop/sass/source.sass
```diff
@@ -297,8 +297,24 @@ header
...
+
+      width: 100%
+      padding-top: 10px
+      min-height: unset
+
+      &:first-child
+        @include ltr
+          border-right: 0
+        @include rtl
+          border-left: 0
```
eaon: This section replicates the old (thanks to Chrome's lack of support for `lang:()`) dead rules of the selection directly above it. I think the rules from `&> section` should be incorporated/removed. Detail nit: I'd use `width: auto` just because that's a bit more forgiving in certain situations.

securedrop/sass/source.sass
```diff
@@ -297,8 +297,24 @@ header
...
+
+      &#return-visit
+        padding-left: 50px
```
eaon: This actually highlighted a lack of polish with right-to-left/left-to-right handling of the index page, but that should probably be handled in a separate ticket/PR.

securedrop/sass/source.sass
```diff
@@ -297,8 +297,24 @@ header
...
+
+    p, hr
       width: auto
+      max-width: 500px
```
eaon: There's no `hr` left in the HTML, so that can be left out.

Wondering about the `max-width`, is that just an aesthetic choice?

securedrop/sass/source.sass
```diff
@@ -315,6 +331,10 @@ header
       div
         width: auto
 
+
+
+
+
```
eaon: Is this to make navigation easier?

eaon (changes requested):

securedrop/sass/source.sass
zenmonkeykstop: Carry-over from the previous rules in the `source_index` module.

securedrop/sass/source.sass
zenmonkeykstop: It's lucky whitespace. (actually just untidiness on my part :) )

securedrop/sass/source.sass
zenmonkeykstop: if it's `width: auto`, will the sections be ordered vertically in all cases tho?

eaon: Yes - for small screens we override the `main`'s `display: flex` with `display: block`, and since there's no float rules here at all, both `section`s should just show up one after the other!

eaon (approved):
Looks good to me now, but since I contributed someone else might want to double check if it looks/works like it is expected to before merging

zenmonkeykstop:
Your changes LGTM!

PR #6460, opened 2022-05-23
Fix issues with token reuse:

lsd-cat:
## Status

Ready for review.

## Description of Changes

Fixes an internally discovered issue.
In `models.py`, fixes two conditions in which the token reuse protection fails:

1) The token is normalized after comparison with `last_token`
2) `last_token` is updated regardless of the reult of `verify_token`

The simple fix is just to change the order of statements in order to force the intended behavior, which is nornalize before everything else and update only if valid.

## Testing

A few tests are updated to reflect the new order of operations.
 * `test_totp_reuse_protections3` checks that normalization happen before everything else
 * `test_totp_reuse_protections4` checks that an ivalid token does not update `last_token`

## Deployment
To be included in 2.4.0, no particular changes to setup or configuration are require for these changes.

## Checklist

### DIFF
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -654,12 +654,6 @@ class Journalist(db.Model):
         if not token:
             return False
 
-        token = self._format_token(token)
-
-        # Store latest token to prevent OTP token reuse
-        self.last_token = token
-        db.session.commit()
-
         if self.is_totp:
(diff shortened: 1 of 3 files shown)


zenmonkeykstop (approved):
LGTM, CI is passing and TOTP and HOTP tokens work.

PR #6470, opened 2022-05-25
Added boundary check on image crop in pagelayout tests

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Fixes intermittent failure in `test_account_edit_hotp_secret` pagelayout test.

## Testing

- [x] CI is passing
- [x] `securedrop/bin/dev-shell bin/run-test -m pagelayout` completes successfully locally
- [ ] screenshots in `securedrop/tests/pagelayouts/screenshots` look correct.

### DIFF
--- a/securedrop/tests/pageslayout/functional_test.py
+++ b/securedrop/tests/pageslayout/functional_test.py
@@ -48,7 +48,7 @@ def autocrop_btm(img, bottom_padding=12):
     bg = gray.getpixel((0, btm))
 
     # Move up until the full row is not of the background luminance
-    while all([gray.getpixel((col, btm)) == bg for col in range(gray.width)]):
+    while btm > 0 and all([gray.getpixel((col, btm)) == bg for col in range(gray.width)]):
         btm -= 1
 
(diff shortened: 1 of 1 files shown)


legoktm:
The screenshots look fine except journalist-admin_edit_hotp_secret.png, which is a 1024x12 white rectangle...

journalist-admin_edit_hotp_secret

^ I attached it and because it's just white you can't see anything...

zenmonkeykstop:
:grimacing: that's what I'd expect to see now instead of the CI fail (basically when whatever causes the bug happens, the image y-index is constrained to 0 (+12) pixels instead of getting into negative values.

zenmonkeykstop:
Maybe instead of just letting it ride we should get it to default to the full height for the crop, I'll take another shot at it in a bit if this doesn't pass for you.

legoktm:
For the developer experience, I'll take the screenshot silently failing over CI constantly failing, and will approve. But I'll file a follow-up task so we remember that this isn't really fixed.

legoktm (approved):
Fixes CI failure, though underlying problem still remains.

PR #6473, opened 2022-06-01
Ignore bogus pylint "security" issue

legoktm:
## Status

Ready for review

## Description of Changes

<https://github.com/PyCQA/pylint/issues/5322> describes a crash when
using a specific extension that we don't use, and really isn't a
security issue at all.

Ideally we'd upgrade pylint anyways, but we've fallen a bit behind
and it isn't a trivial version bump.

## Testing

* [x] Run `make safety`, observe it complains about nothing

## Deployment

## Checklist


zenmonkeykstop (approved):
`make safety passing`, issue indeed looks safe to ignore.

PR #6477, opened 2022-06-16
Ignore cookiecutter CVE-2022-24065

legoktm:
## Status

Ready for review

## Description of Changes

The vulnerability is about Mercurial command injection, which would be a
problem except that we use Git.

## Testing

* [ ] `make safety` now passes.

## Deployment

## Checklist


zenmonkeykstop (approved):
`make safety` is passing, and the reasoning for ignoring this check is sound.

PR #6479, opened 2022-06-22
6111 - Add semgrep tests

ghost:
## Status

Ready for review

## Description of Changes

Add basic semgrep analysis to tests, which tests the `securedrop/` directory. Tests do not only cover python, but also html templates, as well as other languages.

Very simple for now, uses publicly available tests. Can be extended in the future.

False positives can be removed as they appear using comments at the end of the concerned line, or on a line directly above the concerned line.

For python (the first example suppresses only the specific error code, the second suppresses all):

```
# nosemgrep: python.a_specific_warning_code.given_by_semgrep
eval(something_insecure)

eval(something_equally_insecure)    # nosemgrep
```

For HTML templates:

```
<!-- nosemgrep: html.warning_code -->
<a href=mytotallyquotedfilename>link</a>
```

## Testing

Verify that the tool yields results if errors are present (it does... see commits which filter false positives)

Verify that the build process fails if errors are found (it also does)

## Deployment

Only affects the build process

## Checklist

Choose one of the following:

- [x ] These changes do not require documentation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -250,11 +250,16 @@ jobs:
             DOCKER_BUILD_ARGUMENTS="--cache-from securedrop-test-focal-py3:${fromtag:-latest}" securedrop/bin/dev-shell bash -c "pip3 install -U -q --upgrade safety && make -C .. safety"
 
       - run:
-          name: Run static security testing on source code
+          name: Run static security testing on source code with bandit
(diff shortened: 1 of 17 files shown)


zenmonkeykstop (approved):
LGTM, thanks @l3th3!

PR #6481, opened 2022-06-25
Refactor functional tests for journalist app filters

nabla-c0d3:
## Status

Ready

## Description of Changes

This PR continue the work from https://github.com/freedomofpress/securedrop/pull/6476, by bringing the new test code and fixtures to a few functional tests for the journalist app.

More specifically, this PR:

* Brings the new test fixtures to `TestJournalist` (for #3836).
  * It also moves a lot of the testing code back to `TestJournalist` instead of having it in `journalist_navigation_steps.py`, which made it hard to follow the tests as it required constantly jumping between the two files.
  * Only code that's shared across multiple tests should go in the "common" classes/files (`journalist_navigation_steps.py`, `app_navigators.py`, etc.).
* Introduces a new fixture, `sd_servers_v2_with_submitted_file`, which spawns the source and journalist apps with a pre-uploaded submission.
  * This fixture directly creates the submission in the app's DB, instead of using Selenium to upload a submission as a source (as done previously). This new approach reduces setup time by ~7 seconds per test.

PR #6482, opened 2022-06-26
Refactor functional tests for journalist app when file is deleted

nabla-c0d3:
## Status

Ready

## Description of Changes

This PR continue the work from https://github.com/freedomofpress/securedrop/pull/6481, by bringing the new test code and fixtures to a few functional tests for the journalist app.

More specifically, this PR:

* Brings the new test fixtures to `TestJournalistDelete` (for #3836).
* Introduces a new fixture, `_sd_servers_v2_with_missing_file`, which spawns the source and journalist apps with a pre-uploaded submission whose file was deleted from disk.

### DIFF
--- a/securedrop/tests/functional/app_navigators.py
+++ b/securedrop/tests/functional/app_navigators.py
@@ -312,27 +312,16 @@ class JournalistAppNavigator:
         )
 
     def journalist_logs_in(self, username: str, password: str, otp_secret: str) -> None:
+        # Submit the login form
+        self.driver.get(f"{self._journalist_app_base_url}/login")
+        self.nav_helper.safe_send_keys_by_css_selector('input[name="username"]', username)
+        self.nav_helper.safe_send_keys_by_css_selector('input[name="password"]', password)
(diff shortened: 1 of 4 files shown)


zenmonkeykstop (changes requested):
Apologies @nabla-c0d3 , I only noticed with this PR that the refactored journalist tests were taking about 150%  the time of the old versions. It looks like the culprit is https://github.com/nabla-c0d3/securedrop/blob/70c030fe7eef4f6f0e95e93747ba1329418632aa/securedrop/tests/functional/app_navigators.py#L327 - specifically, `self.driver.current_url` has a trailing slash while `self._journalist_app_base_url` does not, so the match failst and the login function goes through the loop 3 times for no reason. 

With trailing slashes stripped, it's much zippier (about 25% the time of the previous version.) Could you verify and maybe append the fix on this PR? I didn't see any other issues.

nabla-c0d3:
@zenmonkeykstop Nice catch! I will fix it 👍

zenmonkeykstop (approved):

PR #6406, opened 2022-04-19
feat: validate `SDConfig.SUPPORTED_LANGUAGES` for usable locales

cfm:
## Status

Ready for review

## Description of Changes

Closes #6366 by determining that a locale is *usable* if it is both (a) *available* in the filesystem and (b) *configured* by the administrator in `SDConfig.SUPPORTED_LANGUAGES`.  Once we've determined which configured locales are actually usable, we:

1. warn if a configured locale is not available;

2. fall back to the hard-coded `FALLBACK_LOCALE` (`en_US`) if `SDConfig.DEFAULT_LOCALE` is not usable; and

3. error out if neither the default nor the fallback locale is usable.

### Routine testing

In the development environment:

* [x] Confirm that the language menu works normally.
    * [x] `zh_{Hans,Hant}` are each (separately) listed, selectable, and loadable.

### Configured language loses support

In the development environment:
    
1. First:

    ```sh-session

   $ rm -rf securedrop/translations/de_DE  # or the language of your choice
   $ make dev
   ```
2. [x] Observe the error:

    ```sh-session

    [2022-04-19 22:14:24,290] ERROR in i18n: Configured locales {Locale('de', territory='DE')} are not in the set of usable locales {Locale('es', territory='ES'), Locale('is'), Locale('pt', territory='BR'), Locale('zh', script='Hans'), Locale('ro'), Locale('it', territory='IT'), Locale('en', territory='US'), Locale('ca'), Locale('nl'), Locale('fr', territory='FR'), Locale('el'), Locale('sv'), Locale('ar'), Locale('cs'), Locale('ru'), Locale('hi'), Locale('nb', territory='NO'), Locale('tr'), Locale('zh', script='Hant'), Locale('sk')}

    ```

3. [x] In the Source Interface, observe that **Deutsch** is no longer available in the language menu.

### Default language loses support

In the development environment:
    
1. Apply the patch:

    ```patch

    --- a/securedrop/config.py.example

    +++ b/securedrop/config.py.example

    @@ -82,7 +82,7 @@ DATABASE_ENGINE = 'sqlite'

     DATABASE_FILE = os.path.join(SECUREDROP_DATA_ROOT, 'db.sqlite')

     

     # Which of the available locales should be displayed by default ?

    -DEFAULT_LOCALE = 'en_US'

    +DEFAULT_LOCALE = 'de_DE'

     

     # How long a session is valid before it expires and logs a user out

     SESSION_EXPIRATION_MINUTES = 120

    ```

2. Then:

[...]

PR #6485, opened 2022-07-11
fix: display full locale name for all locales with shared languages

### LINKED ISSUE #6475: `map_locale_display_names()` does not set `use_display_name` for the first of multiple locales for a language
## Description

For languages like `pt` and `zh` for which we now have multiple translated locales, `map_locale_display_names()` sets `use_display_name` for only the second and following locales.  Prior to an overzealous refactoring in #6406, it would be set for all locales for a multi-locale language.

## Steps to Reproduce

Reported by @deeplow in <https://forum.securedrop.org/t/can-pt-br-become-portugues-brasil-instead-of-just-portugues/1455>.  *Test cases TK.*

## Expected Behavior

* `pt_BR` = `Português (Brasil)`
* `pt_PT` = `Português (Portugal)`

## Actual Behavior

* `pt_BR` = `Português`
* `pt_PT` = `Português (Portugal)`

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Fixes #6475 .

Updates locale config to use full locale names for locales with common languages (ie Portuguese) if both locales are enabled.

## Testing

- [ ] confirm that both Portuguese locales are listed with their full locale name including country.
- [ ] confirm that switching locales works.

## Deployment

n/a

### DIFF
--- a/securedrop/i18n.py
+++ b/securedrop/i18n.py
@@ -16,7 +16,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 import collections
-from typing import List, Set
+from typing import Dict, List, Set
 
 from babel.core import (
     Locale,
@@ -175,18 +175,21 @@ def map_locale_display_names(config: SDConfig) -> None:
     to distinguish them. For languages with more than one translation,
     like Chinese, we do need the additional detail.
     """
-    seen: Set[str] = set()
+
+    language_locale_counts = collections.defaultdict(int)  # type: Dict[str, int]
+    for l in sorted(config.SUPPORTED_LOCALES):
+        locale = RequestLocaleInfo(l)
+        language_locale_counts[locale.language] += 1
+
     locale_map = collections.OrderedDict()
     for l in sorted(config.SUPPORTED_LOCALES):
         if Locale.parse(l) not in USABLE_LOCALES:
             continue
 
         locale = RequestLocaleInfo(l)
-        if locale.language in seen:
+        if language_locale_counts[locale.language] > 1:
             # Disambiguate translations for this language.
             locale.use_display_name = True
-        else:
-            seen.add(locale.language)
 
         locale_map[str(locale)] = locale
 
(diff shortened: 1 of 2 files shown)


securedrop/i18n.py
```diff
@@ -175,18 +175,21 @@ def map_locale_display_names(config: SDConfig) -> None:
     to distinguish them. For languages with more than one translation,
     like Chinese, we do need the additional detail.
     """
-    seen: Set[str] = set()
+
+    language_locale_counts = collections.defaultdict(int)  # type: Dict[str, int]
+    for l in sorted(config.SUPPORTED_LOCALES):
+        locale = RequestLocaleInfo(l)
+        language_locale_counts[locale.language] += 1
```
cfm, line 182: The key insight here, which I [previously missed](https://github.com/freedomofpress/securedrop/pull/6406/files#r863117986), is that setting `locale.use_display_name = True` for all locales of a multiple-locale language _does_ require iterating twice over the list of locales.  It's effectively a lookahead.

In bc5e3661babc4ff8af7fb8fe18b542daeef1b984 I checked for the edge case in which a multiple-locale language has one or more unsupported locales.  I don't think this check is worth restoring here, since without it the worst case is that a locale belonging to a multiple-locale language appears overspecified in the language selector.

securedrop/tests/test_i18n.py
```diff
@@ -410,3 +410,17 @@ def test_html_attributes(journalist_app, config):
     )
     html = resp.data.decode("utf-8")
     assert '<html lang="en-US" dir="ltr">' in html
+
+
+def test_same_lang_diff_locale(journalist_app, config):
```
cfm, line 415: I would have just unit-tested the value of `LOCALES` after `map_locale_display_names()`.  But functional-testing that the expected locales appear in the response body works too!

cfm (approved):
I've left comments for the record, but this looks good to me!  Thanks, @zenmonkeykstop.

- [x] confirm that both Portuguese locales are listed with their full locale name including country.
- [x] confirm that switching locales works.

PR #6492, opened 2022-07-18
fix: compare bytes, not strings, in codename filter check.

### LINKED ISSUE #6491: Initial messages containing non-ascii characters fail if codename filtering is enabled.
## Description

Codename filtering was introduced in 2.3.0, allowing admins to block initial submissions containing only the user's codename, as they should not be shared with journalists. The filter uses the `compare_digest()` function to ensure constant-time comparison, but this fn will throw a `TypeError` if any of the strings being compared  contain Unicode.

## Steps to Reproduce

- start up `make dev` on 2.4.0
- visit the JI and enable codename filtering under Admin > Instance Config
- visit the SI, create a new source, and submit an initial message containing unicode, ie `Hallo! ö, ü, ä, or ß`

## Expected Behavior
- Message is submitted

## Actual Behavior
- 500 error, and (in dev) stack trace due to TypeError

## Comments

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #6491 .

When codename filtering is enabled, compares utf-encoded bytes of messages and codenames instead of raw strings, as valid messages may include non-ASCII chars.

## Testing

- run `make dev` from this branch
- visit the JI and enable codename filtering via Admin > Instance config
- visit the SI and create a new source.
- [x] attempt to submit the source's codename and verify that a flash error message is displayed instructing you to not do that
- [x] attempt to submit a message containing unicode, such as `Hallo! ö, ü, ä, or ß`, and verify that it is submitted successfully.

## Deployment

To be deployed as part of a regular or hotfix release, no special considerations.

Choose one of the following:

### DIFF
--- a/securedrop/source_app/utils.py
+++ b/securedrop/source_app/utils.py
@@ -23,7 +23,7 @@ def codename_detected(message: str, codename: str) -> bool:
     """
     message = message.strip()
 
-    return compare_digest(message.strip(), codename)
+    return compare_digest(message.strip().encode("utf-8"), codename.encode("utf-8"))
 
 
 def flash_msg(
--- a/securedrop/tests/test_source.py
+++ b/securedrop/tests/test_source.py
@@ -356,7 +356,10 @@ def _dummy_submission(app):
     """
     return app.post(
         url_for("main.submit"),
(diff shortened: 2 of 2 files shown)


eaon (approved):
Thanks for the quick turnaround on this!

Issue #6410, opened 2022-04-22
Host dh-virtualenv package on apt.freedom.press instead of pinning it from impish


legoktm:
## Description

dh-virtualenv isn't available in focal. In the builder we pull it from impish with:
```
# TEMPORARY: install dh-virtualenv from unstable Ubuntu release, pending focal package:
# https://github.com/spotify/dh-virtualenv/issues/298
RUN echo "deb http://archive.ubuntu.com/ubuntu/ impish universe main" > /etc/apt/sources.list.d/ubuntu-impish.list
COPY dh-virtualenv.pref /etc/apt/preferences.d/

RUN apt-get update && apt-get install -y dh-virtualenv
```

Instead of this kind of hack that needs to rely on pinning to only take this from impish, we can create a component on apt.freedom.press called "builddeps" (or backports, or ...), copy dh-virtualenv there, and add that source to the builder (with some bootstrapping of the securedrop-keyring fun).

## Motivation
* At some point in the near? future Ubuntu will archive Impish, so we'll need to update the URL or mirror it anyways. And maybe at the same time or at a different point in time, the signatures on Impish's Release files will expire, causing other issues. Would be nice to just get head of all of that.
* Also, the current setup is a bit icky :v

legoktm:
I think #6538 (Drop dh-virtualenv) is a better direction to go in.

PR #6502, opened 2022-07-28
Update builder  image to use Jammy repo for dh-virtualenv

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Updates builder to use Jammy for dh-virtualenv. This also involves pulling in python3-virtualenv explicitly before adding Jammy to the repo list, as otherwise dh-virtualenv's dependencies cannot be satisfied.

## Testing
- [x] CI is passing
- [x] `make build-debs` passes locally.
- [x] SI and JI are available on a staging (or prod vm) env provisioned with packages from this branch

### DIFF
--- a/molecule/builder-focal/Dockerfile
+++ b/molecule/builder-focal/Dockerfile
@@ -25,6 +25,7 @@ RUN apt-get -y update && apt-get upgrade -y && apt-get install -y \
         python3-pip \
         python3-setuptools \
         python3-venv \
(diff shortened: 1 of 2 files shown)


zenmonkeykstop:
This should also close https://github.com/freedomofpress/securedrop/issues/6410 - though it does not host dh_virtualenv in our own repo.

legoktm (approved):

PR #6519, opened 2022-08-16
Remove old signing key from securedrop-admin

### LINKED ISSUE #6511: Phase out old signing key in `securedrop-admin`
`securedrop-admin` currently still trusts the old signing key for releases. It's been more than a year since the [full key rotation](https://media.securedrop.org/media/documents/signing-key-transition.txt) (not to be confused with the recent expiry bump), so I think this can be safely phased out:

https://github.com/freedomofpress/securedrop/blob/2f2db872f60bbc9024e0b35b8ebcdc2f5695acb2/admin/securedrop_admin/__init__.py#L49-L55

eloquence:
## Status

Ready for review

## Description of Changes

- Ensures the previous signing key no longer validates tags verified via `securedrop-admin`. It's still valid for old  tags, but `securedrop-admin` should never have to verify those.
- Also removes the previously used UIDs.
- Updates GPG test output to current year for clarity (the fingerprint and year combinations otherwise don't make sense); it makes no difference in terms of test results. Also updates copyright statement for consistency.
- Sprinkles in some code comments :sparkles: 

Fixes #6511

## Testing

- On an Admin Workstation, apply the changes to `__init.py__` in this PR (I manually applied them to the 2.4.1 tag, but testing on this branch should work as well for this purpose) and bounce the network. Verify that you can update to the latest tag successfully.

## Deployment considerations

- This PR removes the UID variant without an email address as valid. I believe this should be safe, but would very much appreciate additional fact-checking on this, see my comment here: https://github.com/freedomofpress/securedrop/pull/6519#issuecomment-1220113645

## Checklist

- [x] I stepped through test plan above.

### DIFF
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -1019,7 +1019,7 @@ def update(args: argparse.Namespace) -> int:
         # appears on the second line of the output, and that there is a single
         # match from good_sig_text[]
         if (
-            RELEASE_KEYS[0] in gpg_lines[1]
+            any(key in gpg_lines[1] for key in RELEASE_KEYS)
             and len(good_sig_matches) == 1
             and bad_sig_text not in sig_result
         ):
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -46,11 +46,9 @@ from prompt_toolkit.validation import ValidationError, Validator
 
 sdlog = logging.getLogger(__name__)
 
-# We list two (2) pubkeys as authorized to sign SecureDrop release artifacts,
-# to provide a transition window during key rotation. On or around v2.0.0,
-# we can remove the older of the two keys and only trust the newer going forward.
+# In the event of a key rotation, this list can be added to
+# in order to support a transition period.
 RELEASE_KEYS = [
-    "22245C81E3BAEB4138B36061310F561200F4AD77",
     "2359E6538C0613E652955E6C188EDD3B7B22E6A3",
 ]
 DEFAULT_KEYSERVER = "hkps://keys.openpgp.org"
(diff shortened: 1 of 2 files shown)


admin/securedrop_admin/__init__.py
```diff
@@ -1021,7 +1019,7 @@ def update(args: argparse.Namespace) -> int:
         # appears on the second line of the output, and that there is a single
         # match from good_sig_text[]
         if (
-            (RELEASE_KEYS[0] in gpg_lines[1] or RELEASE_KEYS[1] in gpg_lines[1])
+            RELEASE_KEYS[0] in gpg_lines[1]
```
legoktm: Maybe `any(key in gpg_lines[1] for key in RELEASE_KEYS)`?

legoktm (commented):
https://github.com/freedomofpress/securedrop/blob/a17aeec4c9120cf6939b48145ebe4df1be9919a6/admin/securedrop_admin/__init__.py#L1007

Don't we also want to remove the securedrop-release-key@ from being valid here?

admin/securedrop_admin/__init__.py
eloquence: SGTM, done!

eloquence:
> Don't we also want to remove the securedrop-release-key@ from being valid here?

Oh, great catch - I think both of those UIDs can indeed be safely removed as well. Pushed a commit to do so.

eloquence:
OK, I've removed those old UIDs in e8ae44aa8827854ef17ea36354149eedf3f08f06 and updated the associated tests in e7e589e923bcefe2da6d66dc126384aaca45b4aa.

My understanding is that the history of the UID is as follows:
- Up until we migrated to keys.openpgp.org in July 2019, there was no email address associated with the release key.
- keys.openpgp.org requires an email address, which is why we added securedrop-release-key@freedom.press.
- For the key rotation in 2021, we used securedrop-release-key-2021@freedom.press to avoid clashes.

All of which is to say that for pre-2021 keys, we can expect both versions with and without the email address to exist. For the 2021 key, only the version with an email address should exist.

For that reason, I think it's indeed safe to remove variants without an email address from the output checks. 

However, please fact-check me on this (especially @zenmonkeykstop). If there are circumstances where users might encounter output for the 2021+ key without the email address, the signature check would now fail.

zenmonkeykstop (approved):
Test plan checks out, agree with the reasoning around UIDs.

PR #6525, opened 2022-08-25
Replace usage of python random module with python secrets module

evilaliv3:
The [secrets](https://docs.python.org/3/library/secrets.html#module-secrets) module is used for generating cryptographically strong random numbers suitable for managing data such as passwords, account authentication, security tokens, and related secrets.

In particular, [secrets](https://docs.python.org/3/library/secrets.html#module-secrets) should be used in preference to the default pseudo-[random](https://docs.python.org/3/library/random.html#module-random) number generator in the random module, which is designed for modelling and simulation, not security or cryptography

### DIFF
--- a/securedrop/tests/migrations/helpers.py
+++ b/securedrop/tests/migrations/helpers.py
@@ -1,16 +1,17 @@
 # -*- coding: utf-8 -*-
 
 import random
+import secrets
 import string
 from datetime import datetime
 
 
 def random_bool():
-    return bool(random.getrandbits(1))
+    return bool(secrets.getrandbits(1))
 
 
 def bool_or_none():
-    return random.choice([None, True, False])
+    return secrets.choice([None, True, False])
 
 
 def random_bytes(min, max, nullable):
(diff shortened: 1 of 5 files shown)


securedrop/tests/migrations/helpers.py
```diff
@@ -1,16 +1,17 @@
...
 
 import random
+import secrets
 import string
 from datetime import datetime
 
 
 def random_bool():
-    return bool(random.getrandbits(1))
+    return bool(secrets.getrandbits(1))
```
legoktm: This should be `secrets.randbits`

legoktm (changes requested):
See inline comment (why CI is failing).

Some of the cases you replaced are not security critical (tests or developer tooling) in which speed might be preferred over true randomness. But I think it's good to replace everything just for consistency, so any use of the random module is a code smell.

evilaliv3:
Fixed @legoktm. Thank you for your review

legoktm (approved):
I peeked a bit more and it turns out `secrets.SystemRandom` and `random.SystemRandom` are the same thing:

```
user@dev ~> python3.8
Python 3.8.13 (default, Jun 10 2022, 00:00:00) 
[GCC 12.1.1 20220507 (Red Hat 12.1.1-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import random, secrets
>>> random.SystemRandom == secrets.SystemRandom
True
```

So for production impact this is really a no-op. Thank you for the PR!

PR #6529, opened 2022-08-29
Drop Sass

### LINKED ISSUE #4452: Ruby sass gem used at build time has reached end-of-life

eaon:
## Description of Changes

Given the type of web application that SecureDrop is, and how the requirements of the JI and the SI are so disparate, Sass is not particularly useful to us. Pulling in Dart for that very limited usefulness seems silly to me, and LibSass (with its helper tool `sassc`) which is available in OS repos is [considered deprecated](https://github.com/sass/libsass#readme). So bare-metal CSS it is!

Fixes #4452

### Source Interface

As the source interface was recently refactored to also make extensive use of semantically named variables, I converted Sass variables to CSS variables before processing the `.sass` files to get the CSS in this PR. It looks a bit unusual with definitions using the `:root` selector, and references looking like `var(--button-danger-border)`, but: no pre-processing needed :+1:

### Journalist Interface

`journalist.css` is a direct conversion to a single file as there was very little of Sass' features used here except for nested selectors and having everything in small component files. As far as I can tell, there's still a lot of dead code in there, but in my view cleanup should happen in a different PR. My assumption here is that any serious work on the JI would involve a bunch of refactoring anyway.

## Testing

_Note:_ stop your development environment before checking out and testing this branch, otherwise the Sass compiler gets confused.

- Dev environment
  - [x] When running the development environment, there's no Sass compiler output interleaved with Flask
  - [x] When either `securedrop/static/css/source.css` or `journalist.css` are changed, the changes are immediately served by the development HTTP server
- Repo
  - [x] When using an established dev environment, files in `securedrop/.sass-cache` show up as untracked
  - [x] When starting from a fresh checkout, `securedrop/.sass-cache` will not be created by the dev environment
- UX/Design
  - [x] When browsing/using the JI and SI, everything looks and works as expected
- Packaging
  - [ ] Sass isn't run in preparation for packaging, instead the unminimised static files are used directly.


legoktm (changes requested):
The package build doesn't have the CSS:
```
user@sd-dev11:~/securedrop/build/focal$ dpkg -c securedrop-app-code_2.5.0~rc1+focal_amd64.deb | grep -i css
-rw-r--r-- root/root     25481 2022-07-25 11:25 ./opt/venvs/securedrop-app-code/lib/python3.8/site-packages/pip/_vendor/chardet/mbcssm.py
-rw-r--r-- root/root      6704 2021-10-05 16:24 ./opt/venvs/securedrop-app-code/lib/python3.8/site-packages/werkzeug/debug/shared/style.css
```

And the test failed too:
```
    =================================== FAILURES ===================================
    __ test_securedrop_app_code_contains_css[docker://focal-sd-dpkg-verification] __
    [gw1] linux -- Python 3.9.2 /home/user/securedrop/.venv/bin/python
    
    securedrop_app_code_contents = 'drwxr-xr-x root/root         0 2022-07-25 18:25 ./\ndrwxr-xr-x root/root         0 2022-07-25 18:25 ./etc/\ndrwxr-xr-...n3 -> /usr/bin/python3\nlrwxrwxrwx root/root         0 2022-07-25 18:25 ./opt/venvs/securedrop-app-code/lib64 -> lib\n'
    
        def test_securedrop_app_code_contains_css(securedrop_app_code_contents: str) -> None:
            """
            Ensures the `securedrop-app-code` package contains necessary css files
            """
            for css_type in ["journalist", "source"]:
    >           assert re.search(
                    r"^.*\./var/www/securedrop/static/" "css/{}.css$".format(css_type),
                    securedrop_app_code_contents,
                    re.M,
                )
    E           AssertionError: assert None
    E            +  where None = <function search at 0x77e2572b7e50>('^.*\\./var/www/securedrop/static/css/journalist.css$', 'drwxr-xr-x root/root         0 2022-07-25 18:25 ./\ndrwxr-xr-x root/root         0 2022-07-25 18:25 ./etc/\ndrwxr-xr-...n3 -> /usr/bin/python3\nlrwxrwxrwx root/root         0 2022-07-25 18:25 ./opt/venvs/securedrop-app-code/lib64 -> lib\n', re.MULTILINE)
    E            +    where <function search at 0x77e2572b7e50> = re.search
    E            +    and   '^.*\\./var/www/securedrop/static/css/journalist.css$' = <built-in method format of str object at 0x77e2560128d0>('journalist')
    E            +      where <built-in method format of str object at 0x77e2560128d0> = '^.*\\./var/www/securedrop/static/css/{}.css$'.format
    E            +    and   re.MULTILINE = re.M
    
    tests/test_securedrop_deb_package.py:307: AssertionError
```

Poking at it now...

zenmonkeykstop:
Probably need to update `securedrop/.rsync_filter`

legoktm:
Yep, that's exactly where I was at :) Testing a build with:
```
diff --git a/securedrop/.rsync-filter b/securedrop/.rsync-filter
index bb4d6f2c1..844a4fd4a 100644
--- a/securedrop/.rsync-filter
+++ b/securedrop/.rsync-filter
@@ -28,6 +28,8 @@ include source_app/*.py
 include source_templates/
 include source_templates/*.html
 include static/
+include static/css/
+include static/css/*.css
 include static/fonts/
 include static/fonts/**
 include static/i/
```

(the good news is that all these different filtering systems are going away in my PR...)

legoktm (approved):
LGTM, new package build contains CSS in the correct spot:
```
$ dpkg -c securedrop-app-code_2.5.0~rc1+focal_amd64.deb | grep -i css
-rw-r--r-- root/root     25481 2022-07-25 11:25 ./opt/venvs/securedrop-app-code/lib/python3.8/site-packages/pip/_vendor/chardet/mbcssm.py
-rw-r--r-- root/root      6704 2021-10-05 16:24 ./opt/venvs/securedrop-app-code/lib/python3.8/site-packages/werkzeug/debug/shared/style.css
drwxr-xr-x root/root         0 2022-07-25 11:25 ./var/www/securedrop/static/css/
-rw-r--r-- root/root     82300 2022-07-25 11:25 ./var/www/securedrop/static/css/journalist.css
-rw-r--r-- root/root     39458 2022-07-25 11:25 ./var/www/securedrop/static/css/source.css
```

debdiff looks good, though diffoscope is useless because it's no longer minified.

PR #6530, opened 2022-08-30
Remove usage of lsb_release

### LINKED ISSUE #6527: Remove usage of lsb_release
## Description

lsb_release is mostly obsolete these days, instead we should read from `/etc/os-release`.

It's still used in https://github.com/freedomofpress/securedrop/blob/develop/securedrop/server_os.py, as well as:
```
user@dev ~/g/f/securedrop> ack lsb_release --ignore-dir=.venv
admin/bootstrap.py
67:        id = subprocess.check_output("lsb_release --id --short", shell=True).decode("utf-8").strip()
88:            dist = subprocess.check_output("lsb_release --codename --short", shell=True).strip()

securedrop/server_os.py
10:        ["/usr/bin/lsb_release", "--release", "--short"],

install_files/ansible-base/roles/tails-config/files/securedrop_init.py
127:cmd = ["lsb_release", "--id", "--short"]

install_files/ansible-base/roles/build-securedrop-app-code-deb-pkg/files/usr.sbin.apache2
101:  /usr/bin/lsb_release rix,

install_files/ansible-base/inventory-dynamic
131:    Returns output of `lsb_release --id --short`, to check
134:    cmd = ["lsb_release", "--id", "--short"]

install_files/securedrop-app-code/debian/rules
5:SECUREDROP_BUILD_PLATFORM=$(shell lsb_release -sc)
```

* https://0pointer.de/blog/projects/os-release (announcement, addresses "why not lsb_release")
* https://www.freedesktop.org/software/systemd/man/os-release.html

jameelkaisar:
## Status

Ready for review

## Description of Changes

Fixes #6527.

- Remove usage of lsb_release

### DIFF
--- a/admin/bootstrap.py
+++ b/admin/bootstrap.py
@@ -63,17 +63,8 @@ def run_command(command: List[str]) -> Iterator[bytes]:
 
 
 def is_tails() -> bool:
-    try:
-        id = subprocess.check_output("lsb_release --id --short", shell=True).decode("utf-8").strip()
-    except subprocess.CalledProcessError:
-        return False
-
-    # dirty hack to unreliably detect Tails 4.0~beta2
-    if id == "Debian":
-        if os.uname()[1] == "amnesia":
-            id = "Tails"
-
-    return id == "Tails"
+    with open("/etc/os-release") as f:
+        return "TAILS_PRODUCT_NAME" in f.read()
 
 
 def clean_up_old_tails_venv(virtualenv_dir: str = VENV_DIR) -> None:
@@ -84,18 +75,19 @@ def clean_up_old_tails_venv(virtualenv_dir: str = VENV_DIR) -> None:
     venv, so it'll get recreated.
     """
     if is_tails():
-        try:
-            dist = subprocess.check_output("lsb_release --codename --short", shell=True).strip()
-        except subprocess.CalledProcessError:
-            return None
-
-        # Tails 5 is based on bullseye / Python 3.9
-        if dist == b"bullseye":
-            python_lib_path = os.path.join(virtualenv_dir, "lib/python3.7")
-            if os.path.exists(python_lib_path):
-                sdlog.info("Tails 4 virtualenv detected. Removing it.")
-                shutil.rmtree(virtualenv_dir)
-                sdlog.info("Tails 4 virtualenv deleted.")
+        with open("/etc/os-release") as f:
+            os_release = f.readlines()
+            for line in os_release:
+                if line.startswith("TAILS_VERSION_ID="):
+                    version = line.split("=")[1].strip().strip('"')
+                    if version.startswith("5."):
+                        # Tails 5 is based on Python 3.9
+                        python_lib_path = os.path.join(virtualenv_dir, "lib/python3.7")
+                        if os.path.exists(python_lib_path):
+                            sdlog.info("Tails 4 virtualenv detected. Removing it.")
+                            shutil.rmtree(virtualenv_dir)
+                            sdlog.info("Tails 4 virtualenv deleted.")
+                    break
 
 
 def checkenv(args: argparse.Namespace) -> None:
--- a/install_files/ansible-base/inventory-dynamic
+++ b/install_files/ansible-base/inventory-dynamic
@@ -6,7 +6,6 @@ for SSH. The THS must already be provisioned on the target host.
 import io
 import json
 import os
-import subprocess
 
 import yaml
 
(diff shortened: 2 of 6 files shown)


install_files/securedrop-app-code/debian/rules
```diff
@@ -2,7 +2,7 @@
 
 DEB_DH_INSTALL_ARGS=-X .git
 
-SECUREDROP_BUILD_PLATFORM=$(shell lsb_release -sc)
+SECUREDROP_BUILD_PLATFORM=$(shell awk -F= '$1=="VERSION_CODENAME" { print $2 ;}' /etc/os-release)
```
legoktm: Since this is Debian packaging, we already have access to the distro name. We have to add:

`include include /usr/share/dpkg/pkg-info.mk` to the top, which will create a `$DEB_DISTRIBUTION` variable for us.

admin/bootstrap.py
```diff
@@ -64,7 +64,10 @@ def run_command(command: List[str]) -> Iterator[bytes]:
 
 def is_tails() -> bool:
     try:
-        id = subprocess.check_output("lsb_release --id --short", shell=True).decode("utf-8").strip()
+        id = subprocess.check_output(
```
legoktm: Instead of shelling out to awk from Python, can we do the string parsing in Python itself? Something like splitting the lines, check if it starts with "VERSION_CODENAME", then taking the part of it after that and stripping quotes.

legoktm (changes requested):
Thanks for submitting the PR! I left comments inline

install_files/securedrop-app-code/debian/rules
legoktm: Actually, I grepped and don't see this used anywhere. We can just remove it entirely.

jameelkaisar: Ok, going to remove it now.

admin/bootstrap.py
jameelkaisar: Yes, makes sense.

admin/bootstrap.py
```diff
@@ -63,10 +63,12 @@ def run_command(command: List[str]) -> Iterator[bytes]:
 
 
 def is_tails() -> bool:
```
legoktm, line 65: I should've checked earlier because I misled you yesterday :( Tails has a non-standard `/etc/os-release`, it looks like:
```
TAILS_PRODUCT_NAME="Tails"
TAILS_VERSION_ID="5.0"
TAILS_DISTRIBUTION="unstable"
```

Since this function is just trying to detect whether we're on tails or not, it could be as simple as `return "TAILS_PRODUCT_NAME" in f.read()`.

That should work across recent Tails 4.x and 5.x, so we don't even need the "dirty hack" anymore.

admin/bootstrap.py
```diff
@@ -84,10 +86,12 @@ def clean_up_old_tails_venv(virtualenv_dir: str = VENV_DIR) -> None:
...
-            return None
+        with open("/etc/os-release") as f:
+            os_release = f.readlines()
+            for line in os_release:
+                if line.startswith("VERSION_CODENAME="):
+                    dist = line.split("=")[1].strip().strip('"')
+                    break
 
         # Tails 5 is based on bullseye / Python 3.9
         if dist == b"bullseye":
```
legoktm: Here we're trying to detect whether we're on Tails 5. So we can look for `TAILS_VERSION_ID` and see if it `.startswith('5.')` probably.

install_files/ansible-base/inventory-dynamic
```diff
@@ -131,9 +130,12 @@ def lookup_host_distro_id():
     Returns output of `lsb_release --id --short`, to check
     whether host machine is Tails.
     """
-    cmd = ["lsb_release", "--id", "--short"]
-    p = subprocess.check_output(cmd).decode("utf-8")
-    distro_id = p.rstrip()
+    with open("/etc/os-release") as f:
```
legoktm: In this case, `lookup_host_distro_id` is only ever used in the next function below, `host_is_tails`. So I think you can merge the two functions and just copy the same code from the earlier `is_tails` in admin/boostrap.py.

install_files/ansible-base/roles/tails-config/files/securedrop_init.py
```diff
@@ -124,9 +124,12 @@
     subprocess.call(["gio", "set", path_desktop + shortcut, "metadata::trusted", "true"], env=env)
 
 # in Tails 4, reload gnome-shell desktop icons extension to update with changes above
-cmd = ["lsb_release", "--id", "--short"]
-p = subprocess.check_output(cmd)
-distro_id = p.rstrip()
+with open("/etc/os-release") as f:
```
legoktm, line 127: Same thing, we just need to detect Tails here.

securedrop/server_os.py
```diff
@@ -1,14 +1,14 @@
...
 
 @functools.lru_cache()
 def get_os_release() -> str:
-    return subprocess.run(
-        ["/usr/bin/lsb_release", "--release", "--short"],
-        check=True,
-        stdout=subprocess.PIPE,
-        universal_newlines=True,
-    ).stdout.strip()
+    with open("/etc/os-release") as f:
```
legoktm, line 8: This will run on Ubuntu and looks correct to me!

legoktm (changes requested):
Unfortunately I made a mistake in assuming the format of `/etc/os-release` on Tails, see my inline comments. But overall I think it'll allow simplifying the code even more.

admin/bootstrap.py
jameelkaisar: I did the same but the `python3_buster_venv_deleted_in_bullseye` test is failing. I am not sure why the script is unable to delete the Tails 4 virtualenv.

legoktm: I think we just need to adjust the mocking, let me see...

legoktm:
OK! CI looks like it's passing (the failure seems unrelated), so now I just want to test it on a real Tails machine, squash it down to a single commit and then merge it :D

legoktm (approved):
Testing on Tails went well and I squashed it down. LGTM, will merge once CI passes! Congrats on your first contribution to SecureDrop!

jameelkaisar:
Feeling awesome on my first contribution to SecureDrop. Thanks @legoktm for guiding me along the way.

PR #6550, opened 2022-09-19
Do not escape org name twice

### LINKED ISSUE #6357: Organization name gets escaped twice
## Description

When configuring an instance's name as (for example) "Hello & World", the `&` shows up as `&amp;` wherever the organization name is used, because it is escaped when it is being set and again when it is being displayed. 

## Comments

As far as I can tell, the organization name is the only user provided input that is escaped before it is set - oversight or intentional?

eaon:
## Description of Changes

If `InstanceConfig.organization_name` is escaped before writing it to the database, we'd need to mark every use of it in the templates as `| safe`, which is more dubious than not escaping the database entry in the first place.

Fixes #6357

## Testing

* Log into JI with an admin user
* Navigate to the _Instance Configuration_ interface
- [ ] Change the instance name to something that would be escaped (for example `&`s or ümläüts) and click _Set Organization Name_
- [ ] When you get the _Preferences saved._ notification, none of the characters of the name you submitted are double-escaped
- [ ] Navigate to the SI and verify the correct name is shown in the title

## Checklist

### DIFF
--- a/securedrop/journalist_app/admin.py
+++ b/securedrop/journalist_app/admin.py
@@ -2,7 +2,6 @@
 
 import binascii
 import os
-from html import escape
 from typing import Optional, Union
 
 import werkzeug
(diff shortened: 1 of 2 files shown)


zenmonkeykstop (approved):
Test plan checks out, made a couple of attempts to  try SQL injections which were shrugged off by SQLAlchemy without fuss!

PR #6534, opened 2022-09-04
Make SDConfig immutable and free of side-effects on import

nabla-c0d3:
## Status

Work in progress

## Description of Changes

TBD

PR #6551, opened 2022-09-19
Create explicit fixture for rqworker to simplify test code

nabla-c0d3:
## Status

Ready

## Description of Changes

This PR is the first step toward implementing a read-only / immutable SDConfig, ie. preventing patterns such as `config.THING = something`, which is used in some of the tests and even in the app's code. This pattern makes it hard to know what the "actual" config is at a given time, and can cause problems when multiple code locations or tests are editing the config.

The end result can be seen at https://github.com/freedomofpress/securedrop/pull/6534 , which I am splitting into multiple PRs including that one, which does the following:

* Create an explicit fixture for setting up rqworker and update other fixtures accordingly.
* Remove `usefixtures = setUpTearDown` as an "implicit" fixture in pytest.ini.
* Replace the existing test for the journalist app's session timeout with a functional test, to avoid having to do `config.SESSION_EXPIRATION_MINUTES = -1`
* Remove dead code in securedrop/tests/utils/env.py.


securedrop/tests/functional/conftest.py
```diff
@@ -153,18 +153,26 @@ def spawn_sd_servers(
         journalist_app_base_url = f"http://127.0.0.1:{journalist_port}"
 
         # Sleep until the source and journalist web apps are up and running
```
nabla-c0d3, line 155: The changes here are for helping with test flakiness

legoktm (commented):
Looks good, though I only did a skim for now. Note that you'll have to rebase this once #6403 lands (I'll do a full review then).

nabla-c0d3:
> Looks good, though I only did a skim for now. Note that you'll have to rebase this once #6403 lands (I'll do a full review then).

I rebased it just now

legoktm (changes requested):
All looks good except `requeue_interrupted_jobs()` in `securedrop/scripts/rqrequeue` needs to be updated for the new required argument.

Surprisingly this is the first case I've encountered of LGTM actually pointing out a real issue, I wonder if pylint is not running on our scripts (will file a separate task for that).

legoktm (approved):
Very nice :)

PR #6562, opened 2022-09-24
Simplify i18n tests and remove sh.sed usage

nabla-c0d3:
## Status

Ready

## Description of Changes

* Simplify i18n-related tests:
  * De-couple the tests from the `journalist_app` and `config` fixtures, which are not actually needed by the tests.
  * Get closer to implementing an immutable SDConfig.
* Remove usage of `sh.sed` (https://github.com/freedomofpress/securedrop/issues/6547).

This PR should be reviewed and  merged after https://github.com/freedomofpress/securedrop/pull/6551.

PR #6563, opened 2022-09-25
Retrieve SDConfig via explicit call and make config immutable

### LINKED ISSUE #5761: legacy `config.py` files may prevent the web applications from starting or break functionality

nabla-c0d3:
## Status

Ready

## Description of Changes

Fixes #5761.

* Make retrieving the SDConfig explicit via a function call `SecureDropConfig.get_current()`, instead of implicit via an import.
    * The "import-to-get-the-config" pattern was complicating parts of the code, as the import could fail depending on the state the code base was in (for example whether a config.py was present or not when running specific tests).
* Make the config class immutable (via `frozen=True`) so that there's only one location where the config values are / can be set.
    * This SDConfig "v2" was already in the functional tests, and I just brought it to the rest of the code base.
* Add a test to ensure that the oldest config.py in the repo (from 2014) can be parsed and that the app will launch.
    * I an hoping this enough to prevent the problem described in https://github.com/freedomofpress/securedrop/issues/5761 , but not 100% sure? The SESSION_EXPIRATION_MINUTES key is definitely missing from this 2014 config.py (so it seems to be a valid test).
* Bring pathlib to more parts of the code.
* The new function to parse the config.py will make it easier to switch to a text-only (ie. no logic) config file (mentioned at https://github.com/freedomofpress/securedrop/issues/1966).

This PR should be reviewed and merged after https://github.com/freedomofpress/securedrop/pull/6562.


securedrop/journalist_app/__init__.py
```diff
@@ -52,11 +39,11 @@ def get_logo_url(app: Flask) -> str:
...
 
 
-def create_app(config: "SDConfig") -> Flask:
+def create_app(config: SecureDropConfig) -> Flask:
     app = Flask(
         __name__,
-        template_folder=config.JOURNALIST_TEMPLATES_DIR,
-        static_folder=path.join(config.SECUREDROP_ROOT, "static"),
+        template_folder=str(config.JOURNALIST_TEMPLATES_DIR.absolute()),
+        static_folder=str(config.STATIC_DIR.absolute()),
```
legoktm: static_folder can be a Path object, per https://flask.palletsprojects.com/en/2.0.x/api/

Note that template_folder cannot, I filed https://github.com/pallets/flask/issues/4892 for that.

securedrop/models.py
```diff
@@ -872,7 +872,7 @@ class JournalistLoginAttempt(db.Model):
 
     __tablename__ = "journalist_login_attempt"
     id = Column(Integer, primary_key=True)
-    timestamp = Column(DateTime, default=datetime.datetime.utcnow)
+    timestamp = Column(DateTime, default=datetime.datetime.utcnow, nullable=True)
```
legoktm, line 875: This is the current default, so was there a reason you made this change (mypy?)? Or is it just part of what I had suggested at the bottom of #6226?

securedrop/sdconfig.py
```diff
@@ -173,5 +118,137 @@ def DATABASE_URI(self) -> str:
...
+    # Parse the local config; as there are SD instances with very old config files
+    # the parsing logic here has to assume some values might be missing, and hence
+    # set default values for such config entries
+    final_db_engine = getattr(config_from_local_file, "DATABASE_ENGINE", "sqlite")
+    final_db_username = getattr(config_from_local_file, "DATABASE_USERNAME", None)
+    final_db_password = getattr(config_from_local_file, "DATABASE_PASSWORD", None)
+    final_db_host = getattr(config_from_local_file, "DATABASE_HOST", None)
+    final_db_name = getattr(config_from_local_file, "DATABASE_NAME", None)
+
+    final_default_locale = getattr(config_from_local_file, "DEFAULT_LOCALE", "en_US")
```
legoktm: I think this should use FALLBACK_LOCALE instead of hardcoding en_US. Same with the next line.

securedrop/sdconfig.py
```diff
@@ -1,155 +1,100 @@
+from dataclasses import dataclass
+from importlib import import_module
 from pathlib import Path
-from typing import Dict, Optional, Set, Type
-
-import config as _config
+from typing import Dict, List, Optional
 
+# TODO(AD): Should be removed or merged into SecureDropConfig or moved to i18n.py?
```
legoktm: My impression is that this is a constant rather than a configuration setting, and so it should be kept wherever it's used (see my comments below).

securedrop/sdconfig.py
```diff
@@ -1,155 +1,100 @@
...
+    SCRYPT_GPG_PEPPER: str
+    SCRYPT_ID_PEPPER: str
+    SCRYPT_PARAMS: Dict[str, int]
+
+    SECUREDROP_DATA_ROOT: Path
+
+    DATABASE_ENGINE: str
+    DATABASE_FILE: Path
+
+    # The following fields cannot be None if the DB engine is NOT sqlite
```
legoktm, line 60: As a TODO, we really should drop all of these. Pretty sure practically we don't support anything other than sqlite, so there's really no point keeping it around. I'll file an issue for it.

securedrop/sdconfig.py
```diff
@@ -1,155 +1,100 @@
...
+    DATABASE_USERNAME: Optional[str]
+    DATABASE_PASSWORD: Optional[str]
+    DATABASE_HOST: Optional[str]
+    DATABASE_NAME: Optional[str]
+
+    SECUREDROP_ROOT: Path
+    STATIC_DIR: Path
+    TRANSLATION_DIRS: Path
+    SOURCE_TEMPLATES_DIR: Path
```
legoktm, line 69: Same with dropping these paths that are really part of the application that people shouldn't be modifying...

securedrop/source_app/__init__.py
```diff
@@ -36,11 +35,11 @@ def get_logo_url(app: Flask) -> str:
...
 
 
-def create_app(config: SDConfig) -> Flask:
+def create_app(config: SecureDropConfig) -> Flask:
     app = Flask(
         __name__,
-        template_folder=config.SOURCE_TEMPLATES_DIR,
-        static_folder=path.join(config.SECUREDROP_ROOT, "static"),
+        template_folder=str(config.SOURCE_TEMPLATES_DIR.absolute()),
+        static_folder=str(config.STATIC_DIR.absolute()),
```
legoktm: Same here

legoktm (changes requested):
Awesome work! My comments are pretty minor things. Could you also expand the commit message (what you had in the PR description is good)?

And I think we can call this PR as fixing #5761.

securedrop/models.py
nabla-c0d3: Correct, it was for mypy. When I run mypy locally, it complains if `nullable=True` is not present.

securedrop/journalist_app/__init__.py
nabla-c0d3: Sounds good - I removed the call to `str()`

securedrop/sdconfig.py
nabla-c0d3: Done

securedrop/sdconfig.py
nabla-c0d3: Not supporting anything but sqlite (and making it "official") would definitely simplify some of the code 👍.

securedrop/sdconfig.py
nabla-c0d3: I 100% agree that only settings that are meant to be modified by the user should be kept here.

securedrop/sdconfig.py
nabla-c0d3: I had to leave it in sdconfig.py so that it could be used as the default value for DEFAULT_LOCALE.

securedrop/source_app/__init__.py
nabla-c0d3: Done

nabla-c0d3:
I made the requested changes and expanded the commit message 👍. Thanks for the review!

legoktm (changes requested):
Noticed that `securedrop/scripts/` needs to be updated.

nabla-c0d3:
> Noticed that `securedrop/scripts/` needs to be updated.

Good catch... I updated them now.

legoktm (approved):
Awesome! This is all really great

Issue #6566, opened 2022-09-27
`make build-debs` failing on develop during a translation step


zenmonkeykstop:
## Steps to Reproduce

- Check out current develop branch
- run `make build-debs`

## Expected Behavior
Debs get built.

## Actual Behavior
build fails with:
```
    TASK [build-securedrop-app-code-deb-pkg : Compile PO to MO.] *******************
fatal: [focal-sd-app]: FAILED! => {"changed": true, "cmd": "cp config.py.example config.py ; trap 'rm config.py' EXIT ; . /tmp/securedrop-app-code-i18n-ve/bin/activate ; /tmp/securedrop-app-code-i18n-ve/bin/python3 ./i18n_tool.py --verbose translate-messages --compile", "delta": "0:00:00.134601", "end": "2022-09-27 17:17:33.927583", "msg": "non-zero return code", "rc": 1, "start": "2022-09-27 17:17:33.792982", "stderr": "Traceback (most recent call last):\n  File \"./i18n_tool.py\", line 36, in <module>\n    class I18NTool:\n  File \"./i18n_tool.py\", line 47, in I18NTool\n    with open(I18N_CONF) as i18n_conf:\nFileNotFoundError: [Errno 2] No such file or directory: './i18n.json'", "stderr_lines": ["Traceback (most recent call last):", "  File \"./i18n_tool.py\", line 36, in <module>", "    class I18NTool:", "  File \"./i18n_tool.py\", line 47, in I18NTool", "    with open(I18N_CONF) as i18n_conf:", "FileNotFoundError: [Errno 2] No such file or directory: './i18n.json'"], "stdout": "", "stdout_lines": []}
```    

## Comments

Suggestions to fix, any other relevant information.

PR #6544, opened 2022-09-14
Reorganize Debian packaging, have debhelper do most of the work

legoktm:
## Status

Ready for review

## Diffs
These are slightly out of date now, but gives a general idea of the *expected* changes.

* [securedrop-config](https://gist.github.com/legoktm/3cff068fea0be430bc925159d7e55f0d), only metadata changes, version number now follows SecureDrop ("2.5.0")
* [securedrop-keyring](https://gist.github.com/legoktm/4bc6549379ff5b54ba0a321d053b3371), only metadata changes
* [securedrop-ossec-agent](https://gist.github.com/legoktm/9b999c483eaf6e25ef4ffd299796f64b), mypy fixes plus metadata changes
* [securedrop-ossec-server](https://gist.github.com/legoktm/b5fae9a97ae2f29e7beb497852337774), mypy fixes plus metadata changes
* securedrop-app-code:
  * [debdiff](https://gist.github.com/legoktm/768df9f184aa8260b81be65dc35ec25a), shows removal of non-reproducible RECORD files, plus the addition of setup.py.
  * [full diffoscope](https://gist.github.com/legoktm/6e13f00dad969dd0b69db4c3bc165e64) this is absurdly long because SOURCES.txt no longer contains every file in the package (it was entirely useless to begin with).
  * [diffoscope without SOURCES.txt](https://gist.github.com/legoktm/0f2ec8bc5011581bc51a961c31241812), shows removal of RECORD files, removal of debuginfo, and subtle changes to the binary `.so` files, because those are not reproducible (yet!). 

## Description of Changes
It's recommended to review this by stepping through the commits one-by-one. The first 7 commits are all cleanup.

Then the important commits:
* `Reorganize Debian packaging, have debhelper do most of the work` - moves stuff around, sets up `<package>.install` files that control what gets installed. All packaging logic removed from ansible except translations. This is step 1 on #6310.
* `Fix some reproducibility issues in securedrop-app-code` - drop `RECORD` files and debugging info. All packages should be reproducible except for the compiled Build IDs in securedrop-app-code. This is towards #308. The remaining issue is #5901.
* `Compile translations via debian/rules` - ports translation ansible logic to a bash script.

And then the remaining stuff is just fixing lint issues and other minor things as a result of the changes.

### Package building
* [x] Run `make build-debs` without this PR, save the debs somewhere ("originals")
* [x] Run `make build-debs` with this PR, with no errors
[...]

PR #6568, opened 2022-09-27
docs: distinguish build-time and runtime allow-lists

cfm:
## Status

Ready for review

## Description of Changes

Learns a lesson from #6566 by asking contributors to distinguish between two different allow-lists and the impact their proposed changes may have on them:
1. The `.rsync-filter` list governs what files are available in the build environment.
2. AppArmor rules govern what files can be accessed at runtime.

## Testing

Filing for the record and for the purpose of debate:

- [ ] This change is worthwhile documentation, even if it's soon obsoleted by #6544, which will eliminate (1) but not (2) above.

Otherwise, feel free to close this in favor of waiting for #6544! 

## Deployment

Developer-only change to GitHub template; no deployment considerations.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -34,7 +34,13 @@ Any special considerations for deployment? Consider both:
 
 - [ ] [Configuration tests](https://docs.securedrop.org/en/latest/development/testing_configuration_tests.html) pass
 
-### If you added or removed a file deployed with the application:
+### If you added or removed a file
+
+#### ...used to build the application:
+
+- [ ] I have updated `securedrop/.rsync-filter` to include the change
+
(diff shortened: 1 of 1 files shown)


legoktm:
I think this is a bit overkill to put in the PR template for everyone, given it's a very specific edgecase..that will hopefully go away soon. What bit you was that you added a new *filetype* instead of just a new file.

If we do really want to add it to the checklist, there's also `MANIFEST.in` which is used to exclude some files from the package (also going away in the same PR!).

zenmonkeykstop:
Yeah, I agree with @legoktm  on this one - I'm kindof inclined not to overburden the PR template, but rather update the developer documentation and pay more attention to CI builds that catch these errors.

cfm:
Works for me.  Thanks, both!

Issue #6483, opened 2022-06-29
reconsider gettext's fuzzy matching


cfm:
## Description

Both `securedrop` (via Babel) and `securedrop-client` (via Weblate) merge source strings equivalent to running `xgettext` *without* `--no-fuzzy-matching`.  These ["fuzzy" entries](https://www.gnu.org/software/gettext/manual/html_node/Fuzzy-Entries.html) can be confusing to translators, to translation reviewers, and (in `securedrop-client`) to reviewers of translation PRs.  We should:
1. [x] ask Localization Lab for their advice on keeping or disabling gettext's fuzzy matching; and
2. [ ] make tooling changes if so recommended.
    - [x] `securedrop`: freedomofpress/securedrop#6772
    - [x] `securedrop-client`: in Weblate

## User Research Evidence

Cross-project concern raised by @gonzalo-bulnes in freedomofpress/securedrop-client#1527.  (Filed here since the ~v2.5.0~ v2.6.0 localization cycle is our next natural point of coordination with Localization Lab, and we might as well be consistent about this between the two projects and their Weblate components.)

## User Stories

*Presumably:*

As a translation reviewer, I don't want to have to guess whether a translation came from a human translator or gettext's fuzzy matching.

As a translation-PR reviewer, I don't want to be confused by apparent errors in incoming translation diffs.

gonzalo-bulnes:
Thanks @cfm for opening this issue and writing the recap!

cfm:
Bumping to v2.6.0, so that (after #6557) we're making only one change to our localization tooling in each release.

cfm:
In today's Server hangout: confirmed for v2.6.0.

cfm:
Localization Lab confirms:

> [F]uzzy translations usually require revision and can contribute to a false sense of productivity and completion with a translation project, so I'd lean toward disabling this support.

gonzalo-bulnes:
For future reference: I flagged specific examples of inadequate `fuzzy`-matched translations in https://github.com/freedomofpress/securedrop-client/pull/1548. If anyone is looking for concrete examples, these are some.

As the Localization Lab folks said, automated translations require review, and that's what we're seeing in those examples. What I'm pointing at is that the automated attempts at translating are not good, and IMHO don't provide much, if any, value to a translator.

cfm:
This is done for `securedrop-client` in Weblate and prepared for `securedrop` in #6772, which I'll update to close this ticket upon merge.

[...]

PR #6570, opened 2022-09-28
l10n: update strings to be translated for v2.5.0

eaon:
## Description of Changes

This diff is the result of "make translate", per
<https://developers.securedrop.org/en/latest/i18n.html#update-strings-to-be-translated>.

## Testing

- [x] Visually sanity-check:
    - [x] the `messages.pot` template; and
    - [x] the `messages.po` catalog for the language of your choice.

### DIFF
--- a/securedrop/translations/de_DE/LC_MESSAGES/messages.po
+++ b/securedrop/translations/de_DE/LC_MESSAGES/messages.po
@@ -37,9 +37,6 @@ msgstr "vor {time}"
 msgid "You have been logged out due to inactivity."
 msgstr "Sie wurden wegen Inaktivität abgemeldet."
 
-msgid "You have been logged out due to password change"
-msgstr "Sie wurden wegen einer Passwortänderung abgemeldet"
-
 msgid "SecureDrop"
 msgstr "SecureDrop"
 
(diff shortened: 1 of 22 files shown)


securedrop/translations/de_DE/LC_MESSAGES/messages.po
```diff
@@ -771,11 +787,15 @@ msgstr "Danke, dass Sie uns diese Information gesendet haben. Bitte kommen Sie s
 msgid "Thanks! We received your message."
 msgstr "Danke! Wir haben Ihre Nachricht erhalten."
 
-msgid "Thanks! We received your document."
-msgstr "Danke! Wir haben Ihr Dokument erhalten."
+#, fuzzy
+#| msgid "Thanks! We received your message."
+msgid "Thanks! We received your file."
+msgstr "Danke! Wir haben Ihre Nachricht erhalten."
```
cfm, line 793: This is a good example of the concern raised in #6483.

cfm (approved):
Visually reviewed `messages.pot` and `de_DE`'s `messages.po` both in pairing with @eaon and here.  Approving, with one caveat inline, and will merge as soon as CI passes.

PR #6571, opened 2022-09-28
Update pytz to 2022.2.1

### LINKED ISSUE #6569: Update pytz to a release that uses modern source packages
## Description

Update pytz to a release that uses modern source packages: `*.tar.gz`.

## Problem

We're pinned to a 5 year old version of pytz: https://pypi.org/project/pytz/2017.3/#files which notably only offers a zip source download. This causes problems with our reproducible wheel scripts that, correctly, expect a gzipped tarball.

## Expected

We should update to a release that provides a gzipped tarball as source a download. Given this is time-based data, grabbing the most recent version should be fine I assume.

legoktm:
## Status

Ready for review

## Description of Changes

Aside from getting 5 years worth of timezone database updates, the older release is problematic because it shipped a zip source release rather than a gzipped tarball, which our reproducible wheel scripts expect.

Fixes #6569.

## Testing

* [x] Basic smoke testing of JI, where timestamps are shown

## Deployment

## Checklist

- [x] I have performed a diff review and pasted the contents: https://github.com/freedomofpress/securedrop-debian-packaging/wiki/pytz-2017.3-to-2022.2.1

### DIFF
--- a/securedrop/requirements/python3/securedrop-app-code-requirements.txt
+++ b/securedrop/requirements/python3/securedrop-app-code-requirements.txt
@@ -256,9 +256,9 @@ python-dateutil==2.7.2 \
 python-editor==1.0.3 \
     --hash=sha256:a3c066acee22a1c94f63938341d4fb374e3fdd69366ed6603d7b24bed1efc565
     # via alembic
-pytz==2017.3 \
-    --hash=sha256:c41c62827ce9cafacd6f2f7018e4f83a6f1986e87bfd000b8cfbd4ab5da95f1a \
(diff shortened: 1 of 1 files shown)


eloquence (approved):
Looks good to me. I compared the display of timestamps in the Journalist Interface across all languages and did not spot any differences whatsoever. The JI displays dates in UTC, and makes calls to `babel`'s `dates` module that do not specify a timezone. Other timezone conversions are handled via Python's `datetime` module. So I'm not sure that our use of `pytz` gets us anything right now, it's just a dependency of `babel`, but good to keep updated nonetheless.

Issue #6542, opened 2022-09-12
LGTM.com shutting down in December 2022


legoktm:
## Description

LGTM.com, which we use for some Python and JavaScript checks, is shutting down at the end of the year: https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/

Their recommendation is to move to "GitHub code scanning".

legoktm:
I tagged this with needs discussion because my personal opinion is that the checks were not useful and if we just dropped them that would be fine. I think other people liked them though? In which case we need to find a replacement.

eloquence:
https://github.com/freedomofpress/securedrop/pull/6537#issuecomment-1239541960 is an example of an LGTM report which I think is somewhat useful. It looks like GitHub's code scanning feature is also based on CodeQL and intended to fully replace LGTM. I defer to @zenmonkeykstop and our security engineers, but I would recommend that we switch over to GitHub's code scanning feature in repos that use LGTM and then re-evaluate after a while whether to expand usage or discontinue it.

legoktm:
> [#6537 (comment)](https://github.com/freedomofpress/securedrop/pull/6537#issuecomment-1239541960) is an example of an LGTM report which I think is somewhat useful.

All of those are already pylint checks:
* Unused import: https://pylint.pycqa.org/en/latest/user_guide/messages/warning/unused-import.html
* Non-callable called: https://pylint.pycqa.org/en/latest/user_guide/messages/error/not-callable.html
* Unreachable code: https://pylint.pycqa.org/en/latest/user_guide/messages/warning/unreachable.html

The only reason that `make app-lint` doesn't fail on that PR right now is we currently only run pylint over `securedrop/` and the newly added files are in a separate directory. I assume that'll be fixed before it lands.

eloquence:
That's fair; it does look like GitHub's CodeQL support is fairly configurable and there may be a narrower subset of checks that usefully complements the ones we already have. [Here's](https://codeql.github.com/docs/codeql-overview/supported-languages-and-frameworks/#python-built-in-support) an overview of the built-in checks.

legoktm:
From https://github.com/freedomofpress/securedrop/pull/6551#pullrequestreview-1127047557:
> Surprisingly this is the first case I've encountered of LGTM actually pointing out a real issue, I wonder if pylint is not running on our scripts (will file a separate task for that).

eloquence:
[...]

PR #6581, opened 2022-09-30
Have pylint run against securedrop/scripts/*

legoktm:
## Status

Ready for review

## Description of Changes

These files don't get matched by default because they don't end with a ".py" suffix, so add them explicitly.

I noticed this because LGTM caught an issue in #6551 that pylint did not.

Refs #6542.

## Testing

* [x] CI passes
* [x] (optional) Manually running `find . -name '*.py' -or -path './scripts/*'` spits out all Python files plus the 3 scripts.
* [x] (optional) Introducing a pylint error into one of the scripts causes it to fail.

## Deployment

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,7 +106,7 @@ ansible-config-lint: ## Run custom Ansible linting tasks.
 .PHONY: app-lint
 app-lint:  ## Test pylint compliance.
 	@echo "███ Linting application code..."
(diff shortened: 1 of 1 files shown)


legoktm:
Also flake8, black, isort are not running against scripts/ :| I wonder if we should just add Python extensions to these and have the package install them as extension-less files.

eaon (approved):

PR #6585, opened 2022-10-01
Remove LOGIN_HARDENING global and reduce usage of SECUREDROP_ENV

### LINKED ISSUE #3600: Alter LOGIN_HARDENING so tests more closely match production
## Description

From `securedrop/models.py`:

```python
LOGIN_HARDENING = True
# Unfortunately, the login hardening measures mess with the tests in
# non-deterministic ways.  TODO rewrite the tests so we can more
# precisely control which code paths are exercised.
if os.environ.get('SECUREDROP_ENV') == 'test':
    LOGIN_HARDENING = False
```

nabla-c0d3:
## Status

Ready.

## Description of Changes

* Removes the `LOGIN_HARDENING` global and moves logic to bypass login hardening to the test suite.
    * This fixes https://github.com/freedomofpress/securedrop/issues/3600 .
    * This ensures that the hardening cannot be mistakenly disabled on a production instance.
    * Hardening will be enabled by default in the tests, but the `login_journalist()` helper function can be used to bypass it.
* Reduces usage of the `SECUREDROP_ENV` environment variable, which, I think, should eventually be removed as it's another, different mechanism for changing the app's config.
* Merged multiple implementations of `_login_user()` into a single `login_journalist()`.


legoktm:
@nabla-c0d3 want to rebase this?

nabla-c0d3:
> @nabla-c0d3 want to rebase this?

Done 👍

securedrop/tests/test_store.py
```diff
@@ -261,6 +261,29 @@ def test_add_checksum_for_file(config, app_storage, db_model):
         assert db_obj.checksum == "sha256:" + expected_hash
 
 
+def _wait_for_redis_worker(job: Job, timeout: int = 60) -> None:
```
nabla-c0d3, line 264: I moved this function here and made it "private" since it's only used in this file.

securedrop/tests/__init__.py
```diff
@@ -1,19 +0,0 @@
...
-# or most tests will fail
-os.environ["SECUREDROP_ENV"] = "test"
-
-# The tests directory should be adjacent to the securedrop directory. By adding
-# the securedrop directory to sys.path here, all test modules are able to
-# directly import modules in the securedrop directory.
-sys.path.append(abspath(join(dirname(realpath(__file__)), "..", "securedrop")))
-
-# This ensures we get pytest's more detailed assertion output in helper functions
-pytest.register_assert_rewrite("tests.functional.journalist_navigation_steps")
```
legoktm, line 18: Do we no longer need these?

legoktm (commented):
Only one question, otherwise looks good!

securedrop/tests/__init__.py
nabla-c0d3: Correct:

* The `SECUREDROP_ENV` variable no longer needs to be set for the tests to pass (it was needed for setting LOGIN_HARDENING to False, for the tests)
* The rest of this file could have been removed sooner, in previous PRs.

legoktm (approved):
🚢 it!

PR #6608, opened 2022-10-11
Security triage

gonzalo-bulnes:
## Description of Changes

This PR silences security alerts for some vulnerabilities, as part of security triage.
Context: https://github.com/freedomofpress/fpf-www-projects/pull/290

## Testing

- [ ] Confirm that the correct alerts were silenced. You can find the _vulnerability IDs_ in the alerting channel.
- [ ] Confirm that nothing needs to be done to prevent this file from being deployed or packaged

### DIFF
--- a/project.json
+++ b/project.json
@@ -1,0 +1,5 @@
+{
+    "variables": {
+        "SAFETY_IGNORE_IDS": []
+    }
+}
--- a/project.json
+++ b/project.json
@@ -1,5 +1,11 @@
 {
     "variables": {
-        "SAFETY_IGNORE_IDS": []
+        "SAFETY_IGNORE_IDS": [
+		"42926",
+		"42923",
+		"49337",
+		"45185",
+		"51385"
+	]
     }
 }


gonzalo-bulnes:
@l3th3 The file needs to be in this repository for now. I first created `securedrop/project.json` in [**fpf-pwww-projects**](https://github.com/freedomofpress/fpf-www-projects) and it took me a while to realize that all checks were skipped because... the `git clone` is skipped if the directory already exists!

I think we can iterate on the file location, but that it would be useful to start by confirming this adds the behavior we expect :slightly_smiling_face:

gonzalo-bulnes:
@zenmonkeykstop @cfm @eaon: Just a heads up that I'm adding the file. The current form of the process of muting security alerts was pre-existing (just not used a lot). @l3th3 and I may revisit some aspects of it as part of our month :pear: pairing for Security Triage. We'l keep you updated of our thinking! :slightly_smiling_face: 

Also CC: @rocodes because this applies to other repos as well, even if not immediately!

ghost:
We can discuss this further in fpf-www-projects, but I think it would definitely be worth us putting in a quick change to centralize all of the ignored IDs, rather than having a projects.json in each individual project

ghost (approved):
Whoops, forgot to approve. Looks good to me :t-rex:

PR #6613, opened 2022-10-11
Re-enable codecov uploader with bundled PGP key

### LINKED ISSUE #6610: Codecov uploads fail when their GPG key is unavailable.
## Description

The codecov uploader hash file is verified using a GPG key downloaded from keybase.io. The key does not appear to be available elsewhere. As a result, CI jobs will fail when keybase.io is down.

## Steps to Reproduce

Run CI when keybase.io is down

## Expected Behavior

app-tests job passes and uploads results to codecov

## Actual Behavior

app-tests job fails when attempting to verify codecov hash file sig.

legoktm:
## Status

Ready for review

## Description of Changes

keybase.io seems to have issue serving the key still, so codecov has posted their key on <https://about.codecov.io/security/>, which I've copied into the repo.

Also use a keyring named "codecov.gpg" instead of an ambigiously named "trustedkeys.gpg" (I don't think we trust codecov to sign anything but their own software).

Fixes #6610.

## Testing

* [ ] key matches <https://about.codecov.io/security/>
* [ ] manually run adjusted gpg commands 

## Deployment


legoktm:
Meh, something is weird with their key, I saw it locally too. Will figure this out later, it's not super urgent.

zenmonkeykstop:
there's one of these puppies: https://unicode-table.com/en/200B/ on line 2. imports OK without it.

zenmonkeykstop (approved):
With the unicode excised from the armored key, CI is passing. Key fingerprint verified locally and gpg commands also check out. Good to go IMO if you want to flip it from draft.

Issue #6620, opened 2022-10-13
Slim down default environment spun up by "make dev" 


zenmonkeykstop:
## Description

(As suggested by @legoktm)

Slim down default environment spun up by "make dev", removing dependencies and setup required for testing. This would help to speed up dev env provisioning.

(This could be done via separate Dockerfiles, but another approach might be to have a single file with named phases, and to call docker build with --target as appropriate.)

## User Research Evidence
n/a, as suggested by @legoktm

## User Stories
As a developer, I want to use minimal time and resources when spinning up a dev environment.

legoktm:
We could also have CI publish the dev container to e.g. quay.io so people only need to download it on first run instead of "building" it.

My PR so far:
```
localhost/securedrop-full-focal-py3                       latest               ce9e5513fa26  About a minute ago  2.9 GB
localhost/securedrop-test-focal-py3                       latest               788b40289994  31 minutes ago      1.17 GB
```

zenmonkeykstop:
Closed by #6625 - I don't think there's a need to get into the publishing to quay.io bit for this round.

PR #6625, opened 2022-10-13
Use a slimmer container for `make dev`

legoktm:
## Status

Ready for review

## Description of Changes

The current dev container is pretty heavy, it pulls in a bunch of different dependencies (X11, Java, devscripts) that are extraneous if you're just trying to run the Journalist/Source Interfaces. The new slimmer container is ~1.19GB.

The full container, used by `make dev-full`, is still the same ~2.9GB.

Refs #6620.

(explicitly not fixing the issue since I think there's still some more room for improvement...)

## Testing

* [ ] `make dev` brings up JI+SI
* [ ] `make lint` uses only the slim container
  * Delete the full container (`podman rmi localhost/securedrop-test-focal-py3`)
  * Run `make lint`
  * Check `podman images` that the test image wasn't built
* [ ] `make dev-tor` brings up onion services
* [ ] `make dev-full` brings up JI+SI, outputs VNC port
* [ ] Other make targets work just fine

## Deployment

2. New installs.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -235,13 +235,13 @@ test-config: securedrop/config.py
 .PHONY: dev
 dev:  ## Run the development server in a Docker container.
 	@echo "███ Starting development server..."
-	@OFFSET_PORTS='false' DOCKER_BUILD_VERBOSE='true' $(DEVSHELL) $(SDBIN)/run
+	@OFFSET_PORTS='false' DOCKER_BUILD_VERBOSE='true' SLIM_BUILD=1 $(DEVSHELL) $(SDBIN)/run
 	@echo
 
 .PHONY: dev-tor
 dev-tor:  ## Run the development server with onion services in a Docker container.
-	@echo "███ Starting development server..."
-	@OFFSET_PORTS='false' DOCKER_BUILD_VERBOSE='true' USE_TOR='true' $(DEVSHELL) $(SDBIN)/run
+	@echo "███ Starting development server with onion services..."
+	@OFFSET_PORTS='false' DOCKER_BUILD_VERBOSE='true' USE_TOR='true' SLIM_BUILD=1 $(DEVSHELL) $(SDBIN)/run
 	@echo
 
 .PHONY: demo-landing-page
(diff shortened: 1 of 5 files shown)


legoktm:
Probably going to slip `FULL_BUILD=1` into `SLIM_BUILD=1` so anything we don't explicitly make slim will keep working.

legoktm:
One downside of this is that if anything changes in the slim container (e.g. a dependency bump), you have to rebuild the full container from scratch.

cfm:
On Fri, Oct 14, 2022 at 02:17:19PM -0700, Kunal Mehta wrote:
> One downside of this is that if anything changes in the slim container
> (e.g. a dependency bump), you have to rebuild the full container from
> scratch.

I'm okay with this compromise, mostly because it makes the layering easy
to reason about.

legoktm:
I have no useful data on how much faster this makes things, I've found that it's dependent on how fast your connection is, but during the day my apartment internet is slower, but at night it's blazing fast so it can be very situational...

Makefile
```diff
@@ -235,13 +235,19 @@ test-config: securedrop/config.py
...
 
 .PHONY: dev-tor
 dev-tor:  ## Run the development server with onion services in a Docker container.
-	@echo "███ Starting development server..."
-	@OFFSET_PORTS='false' DOCKER_BUILD_VERBOSE='true' USE_TOR='true' $(DEVSHELL) $(SDBIN)/run
+	@echo "███ Starting development server with onion services..."
+	@OFFSET_PORTS='false' DOCKER_BUILD_VERBOSE='true' USE_TOR='true' SLIM_BUILD=1 $(DEVSHELL) $(SDBIN)/run
+	@echo
+
+.PHONY: dev-full
```
zenmonkeykstop: Do we need a `dev-full` target? We do need the full build for automated testing purposes, but it doesn't seem like there are development tasks that would require it. (Unless folks are VNCing in to the container to run TBB there instead of locally, which seems unnecessary.)

zenmonkeykstop (changes requested):
Looks/works great, decent time saving (anec-data: 3min vs 7min for me) and ditto on the bandwidth front. One small nit about whether the `dev-full` target is required - if there is a valid use case for it please feel free to ignore.

Makefile
legoktm: I don't have any use case for it, can drop it and we can re-add it later if one does come up.

zenmonkeykstop (approved):

PR #6656, opened 2022-10-19
[WIP] Add security considerations to feature requests

ghost:
## Status

Work in progress

## Description of Changes

Adds a security checklist to:

- PR templates
- Feature request templates (not done)

## Testing

Try to apply it to existing PRs?

## Deployment

No particular considerations

## Checklist

Changes to the developer docs will be necessary:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -22,6 +22,21 @@ Any special considerations for deployment? Consider both:
 
 ## Checklist
 
+### Security Checklist
+
+Do any of the changes made allow a malicious user to:
+
+- [ ] pretend to be a different user? (Spoofing / Authenticity)
+- [ ] modify data used by the system? (Tampering / Integrity)
+- [ ] deny that they performed an action to change the system's state? (Repudiation / Non-repudiability)
+- [ ] extract information that should be kept secret? (Information disclosure / Confidentiality)
+- [ ] exhaust system resources such that the system is no longer functioning as intended? (Denial of Service / Availability)
+- [ ] increase their ability to work with system resources? (Elevation of Privilege / Authorization)
+
+#### If yes, please list what mechanisms are in place for mitigation and/or prevention:
+
+List any protective mechanisms here
+
 ### If you made changes to the server application code:
 
 - [ ] Linting (`make lint`) and tests (`make test`) pass in the development container


zenmonkeykstop:
Would it make sense to have a section on the questions in the PR in  https://developers.securedrop.org/en/latest/contributing.html or in the developer guide in general? My concerns with just adding it into the PR template are that a) said template is already quite long and b) it introduces security concerns late in the process, as folks will typically have already been working on changes for quite a while before submitting a PR.

ghost:
My issue with not having it directly in the PR template is that if it's just in the contribution guidelines, it will definitely be consistently ignored. I agree with keeping it short, but the whole point is to have something that gets consistently considered. If it's only in the contribution guidelines, then it really won't serve a purpose at all

rocodes:
Thank you for putting this together @l3th3 -- I like the idea of getting folks to think about the threat model at more stages of our development cycle.

I do agree with @zenmonkeykstop that I'd like to see something like this in the `Contributing` section of the docs, even though I understand that it will be mostly overlooked there. (I'd _really_ like to see us incorporate a version of this as we plan out features  -- so if we put together any "this is a feature spec template" doc, something structured that refers back to our threat model would be a great addition).

In terms of the specific changes proposed here:

- For me, the wording is a little confusing--we're asking users to "check a box if the following bad thing could happen as a result of your code changes", whereas typically and in the rest of the PR template, we use the checkbox to mostly signal positive actions that we have to make sure we accomplish before merging (eg "I have written a test plan").  I stumbled a bit over "Do any of the changes made allow a malicious user to: [...]" followed by the checkbox items.

- I don't think we can ask people to understand and accurately estimate whether the code they are introducing does any of these things (eg [very unscientific survey](https://www.zdnet.com/article/no-love-lost-between-security-specialists-and-developers/), [discussion](https://www.schneier.com/blog/archives/2019/07/software_develo.html)).

I think some other possibilities that get at the same goal could be:

- link to our public threat model, ask people to read it, and [insert small task here, such as asking them which threat(s) in the threat model are most relevant to this code, or what questions they have about our threat model after submitting this PR]

- give more concrete examples that nudge people towards thinking about security, rather than mentioning the high-level principles/terms, which may be confusing and not accessible to some contributors, especially newer ones. Instead of saying "could your code increase [an attacker's] ability to work with system resources?" (since no one will necessarily be able to pick out if that is true, and if they could, they probably would have created a different implementation), we could provide or link out to best-practices examples ("Writing a file? Introducing a new script? Make sure that....").  Same thing with the CIA/STRIDE terms. 

- (? not sure, just brainstorming) encourage behaviours such as "suggest a few tags that describe this change" (so that PRs get labelled properly). This doesn't mean that, for example, UI-tagged PRs don't get security reviews; rather, it encourages people to think about the _kind_ of change they are making (and the _kinds_ of vulnerabilities that such a change could introduce). An alternative to this would be "suggest a few types of security themes that come to mind when working on this PR" (eg: memory safety, networking, writing to disk, etc).

Lastly, I hope we can disincentivize people from the "check a box to get the PR up" (or "don't know how to check the box so can't get the PR up") - if we do keep something like this in, we should have an "I don't know" answer.  I'd have more confidence asking contributors to tag positive attributes ("This code writes files to disk", "this code makes network requests", etc) than I would asking them (us) to make assessments about what kind of vulnerabilities could be introduced by those changes (the 'unknown unknowns' issue).

PR #6657, opened 2022-10-19
Drop passlib, switch to argon2id hashes

### LINKED ISSUE #6631: Consider using argon2-cffi directly and dropping passlib
passlib was introduced in #2918, on the basis that it would give us "future proof and easily migrateable password management should we change password hashing algorithms/parameters."

Unfortunately passlib looks unmaintained (no commits in 2 years and the upstream issue tracker has [literal spam tickets](https://foss.heptapod.net/python-libs/passlib/-/issues/?sort=created_date&state=opened) sitting for weeks). We're also not using any of the migration functionality, we're just invoking argon2 hashing directly:
```python
        self.passphrase_hash = argon2.using(**ARGON2_PARAMS).hash(passphrase)
```

I don't think the migration functionality is worth the ~20k lines of Python passlib brings in (counted by tokei, minus tests). If we need migration, we'd probably implement it with something like:
```python
if hash.startswith('$argon2'):
    verify_argon2(hash, passphrase)
    self.passprase_hash = hash_passphrase(passphrase)
    self.save()
else:
    verify_modern(hash, passphrase)
```

So my preference would be to drop passlib and use argon2-cffi directly.

### LINKED ISSUE #6655: Journalist passphrases should be hashed with Argon2id instead of Argon2i
*Originally filed as https://github.com/freedomofpress/securedrop-security/issues/86*

We currently use Argon2i to encrypt journalist passphrases:
```
sqlite> select passphrase_hash from journalists;
$argon2i$v=19$m=65536,t=4,p=2$otS6V+rd29u79z6nFAJAaA$fWn/7La/sHiQAM7YPOLo5Q
$argon2i$v=19$m=65536,t=4,p=2$VAoBIMSY8957T6lVCsFYKw$Z8vGfVR/P87pNqmn0oz/yg
$argon2i$v=19$m=65536,t=4,p=2$Quj9H4PwntPau5fSOgdgjA$qzvLuOzgN86ITlpdepFiJQ
```

However, Argon2id is what's recommended these days, [this SO answer goes into depth and summarizes with](https://security.stackexchange.com/questions/193351/in-2018-what-is-the-recommended-hash-to-store-passwords-bcrypt-scrypt-argon2):
> In short: use Argon2id if you can, use Argon2d in almost every other case, consider Argon2i if you really do need memory side-channel attack resistance.

Part of the issue here is that we fell behind on upgrading passlib. The latest version of it uses Argon2id by default (our argon2-cffi version already defaults to id).

legoktm:
## Status

Ready for review

## Description of Changes

* Use argon2-cffi directly instead of through passlib.
* Switch to argon2id passphrase hashes for journalists

The individual commit messages have more details on specifics.

Fixes #6631.
Fixes #6655.

## Testing

* [x] Spin up a dev environment using develop (probably use `HEAD~2` since develop will have moved forward since I wrote this). Log in as `journalist`. Look in the database (e.g. `SELECT username, passphrase_hash FROM journalists`) and verify the hash starts with `$argon2i$...`.
* [x] Check out this branch, see the dev environment reload models.py
* [x] Log in as `journalist` again. Look in the database and verify the hash now starts with `$argon2id$...`
* [x] Spin up a fresh dev environment, verify the passphrase_hashes start with `$argon2id$...`

Note that there's a 50% chance the generated journalist uses the even more legacy pw_hash/pw_salt scrypt mechanism if you notice that passphrase_hash is sometimes null.

## Deployment

* Newly created journalists will use the argon2id hash by default.
* Existing journalists will be migrated as they log in.

## Checklist

### DIFF
--- a/securedrop/alembic/versions/2e24fc7536e8_make_journalist_id_non_nullable.py
+++ b/securedrop/alembic/versions/2e24fc7536e8_make_journalist_id_non_nullable.py
@@ -8,10 +8,10 @@ Create Date: 2022-01-12 19:31:06.186285
 import os
 import uuid
 
+import argon2
 import pyotp
 import sqlalchemy as sa
 from alembic import op
-from passlib.hash import argon2
 
 # raise the errors if we're not in production
 raise_errors = os.environ.get("SECUREDROP_ENV", "prod") != "prod"
@@ -33,7 +33,7 @@ depends_on = None
 
 def generate_passphrase_hash() -> str:
     passphrase = PassphraseGenerator.get_default().generate_passphrase()
-    return argon2.using(**ARGON2_PARAMS).hash(passphrase)
+    return argon2.PasswordHasher(**ARGON2_PARAMS).hash(passphrase)
 
 
 def create_deleted() -> int:
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -8,6 +8,7 @@ from io import BytesIO
 from logging import Logger
 from typing import Any, Callable, Dict, List, Optional, Union
 
+import argon2
 import pyotp
 import qrcode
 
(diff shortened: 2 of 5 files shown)


eaon (approved):
Yay for dependency slashing. Checks out, looks great, thanks!

PR #6666, opened 2022-11-01
Automatically change respective SELinux contexts when running containers

### LINKED ISSUE #6665: Make Docker/Podman/Moby play nice with SELinux
## Description

Currently, we do not instruct Docker/Podman/Moby to change the security context for our containers automatically. This results in machines with SELinux enabled being greeted with permission denied errors when trying to run `make dev[-tor]` or other containers of ours.

We should make sure that it _just works_ by using the `:Z` option for `--volume`.

## User Stories

I use Podman on Fedora with SELinux enabled. I have to manually change the security context for the repository so that the development environment executes successfully.

eaon:
## Description of Changes

This allows us to run containers without manual intervention on systems with SELinux enabled.

Fixes #6665

## Testing

This may be reviewed by multiple people that have different setups at hand.

- To ensure that previously manually changed security contexts do not impact testing this PR, always grab a fresh checkout of the respective branches
- When SELinux is installed and enabled:
  - [ ] Confirm that with the `develop` branch, `make dev` errors out
  - [ ] Confirm that with the `6665-containers-selinux-compat` branch, `make dev` works as expected
- When SELinux is not installed:
  - [ ] Confirm that with a fresh checkout of the `6665-containers-selinux-compat` branch, `make dev` works

## Checklist

### DIFF
--- a/admin/bin/dev-shell
+++ b/admin/bin/dev-shell
@@ -31,7 +31,7 @@ function docker_run() {
         docker run \
             --rm \
             --user "${USER:-root}" \
-            --volume "${TOPLEVEL}:/sd-root" \
+            --volume "${TOPLEVEL}:/sd-root:Z" \
             --workdir "/sd-root/admin" \
             -ti ${DOCKER_RUN_ARGUMENTS:-} securedrop-admin "$@"
     else
--- a/devops/scripts/shellcheck.sh
+++ b/devops/scripts/shellcheck.sh
@@ -16,7 +16,7 @@ function run_native_or_in_docker () {
     if [ "$(command -v shellcheck)" ]; then
         shellcheck -x --exclude="$EXCLUDE_RULES" "$@"
     else
-        $DOCKER_BIN run --rm -v "$(pwd):/sd" -w /sd \
+        $DOCKER_BIN run --rm -v "$(pwd):/sd:Z" -w /sd \
             -t docker.io/koalaman/shellcheck:v0.4.7 \
             -x --exclude=$EXCLUDE_RULES "$@"
     fi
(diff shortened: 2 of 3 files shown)


zenmonkeykstop:
Tried it out in a Qubes fedora-36 VM with SELinux enabled. Couldn't replicate the failure on either `develop` or `6665-containers-selinux-compat` - is it bare-metal-specific?

nathandyer:
I can echo @zenmonkeykstop's findings, I tested on bare metal using my personal laptop running Fedora 37 Beta with SELinux enabled (verified using `sestatus`). Even with SELinux enabled, `make dev` works without issue. Both this branch and the `develop` branch appear to have the same results for me.

legoktm:
Hmmm, I definitely saw the failure when I was sitting next to @eaon two weeks ago :/ And adding the `:Z` did fix it so it's good at least that the new branch isn't breaking anything...

legoktm:
Were you both using docker or podman?

zenmonkeykstop:
podman for me.

nathandyer:
Docker for me

eaon:
At least for regular Docker, I believe there wouldn't be any SELinux issues anyway because all the would be run as `root` anyway. Podman on the other hand defaults to running containers as the same user, and that's where the container security context becomes relevant.

Why that didn't work for @zenmonkeykstop I'm not entirely sure but my guess would also be that it's due to Qubes virtualisation / kernel.

nathandyer:
@eaon Great news: I was able to reproduce this! :tada: And even better news, this branch fixes the issue.

After running:
```
export USE_PODMAN=1
make dev
```

from the `develop` branch on my Fedora system, I get: "Permission denied"

If I then switch to this branch and re-run `make dev`, everything works perfectly.

nathandyer (approved):
Everything LGTM. Tested it out in Fedora with SELinux enabled (using podman), and in Debian without SELinux. Can confirm this fixes the issue, and with it being a small two-liner, I think it's good to go ahead and merge.

Issue #6588, opened 2022-10-03
Clean up manual sys.path additions in scripts


legoktm:
*This is a good first issue for new contributors to take on, if you have any questions, please ask on the task or in our [Gitter room](https://gitter.im/freedomofpress/securedrop)!*

## Description

Our scripts have some manual `sys.path` additions that aren't actually needed:
```
$ rg sys.path securedrop/scripts -C 1
securedrop/scripts/rqrequeue
8-
9:sys.path.insert(0, "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages")  # noqa: E402
10:sys.path.insert(0, "/var/www/securedrop")  # noqa: E402
11-

securedrop/scripts/shredder
11-
12:sys.path.insert(0, ".")  # noqa: E402
13:sys.path.insert(
14-    0, "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages"
15-)  # noqa: E402
16:sys.path.insert(0, "/var/www/securedrop")  # noqa: E402
17-

securedrop/scripts/source_deleter
11-
12:sys.path.insert(0, ".")  # noqa: E402
13:sys.path.insert(
14-    0, "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages"
15-)  # noqa: E402
16:sys.path.insert(0, "/var/www/securedrop")  # noqa: E402
17-
```

There's no point in adding `.` to the path because it's already part of the lookup path by default. And the site-packages directory is already added by the virtualenv, so it's unnecessary too. We do need to keep the `/var/www/securedrop` one though.

rahulXs:
Hi, I would like to work on this issue. Could you please assign it to me? Thanks!

legoktm:
Done, let us know if you have any questions or issues!

rahulXs:
@legoktm I've raised the PR #6589 for this issue. Please help to review.
I didn't fill the PR template checklist completely because few points (test plan, docs changes required) were not clear to me. Please let me know if any change is required. Thanks!

legoktm:
I made a mistake, it works fine in production systems but breaks in the dev environment :/
```
Traceback (most recent call last):
  File "/home/user/github/freedomofpress/securedrop/securedrop/scripts/shredder", line 14, in <module>
    import journalist_app
ModuleNotFoundError: No module named 'journalist_app'
Traceback (most recent call last):
[...]

PR #6668, opened 2022-11-02
Fix PYTHONPATH for scripts in dev environment

legoktm:
## Status

Ready for review

## Description of Changes

`shredder` and `source_deleter` need to be able to import `journalist_app`, and should have the securedrop/ folder added to PYTHONPATH, just like is already done for rqrequeue. (rqworker doesn't need it since that's not our script, it's installed into the venv.)

This better matches how the systemd services are run in production since the unit has `PYTHONPATH="/var/www/securedrop:..."` (and also the sys.path addition).

Refs #6588.

## Testing

* [x] Without this patch, run `make dev`. See the errors pasted in https://github.com/freedomofpress/securedrop/issues/6588#issuecomment-1301006264
* [x] With this patch, run `make dev`. See the following output:
  * 2022-11-02 18:03:29,196 INFO Purging deleted sources every 10 seconds.
  * 2022-11-02 18:03:29,244 INFO Clearing shredder every 60 seconds.

## Deployment

Any special considerations for deployment? No, dev-only.

## Checklist

### DIFF
--- a/securedrop/bin/run
+++ b/securedrop/bin/run
@@ -19,7 +19,7 @@ maybe_use_tor
 # run the batch processing services normally managed by systemd
 /opt/venvs/securedrop-app-code/bin/rqworker &
 PYTHONPATH="${REPOROOT}/securedrop" /opt/venvs/securedrop-app-code/bin/python "${REPOROOT}/securedrop/scripts/rqrequeue" --interval 60 &
-/opt/venvs/securedrop-app-code/bin/python "${REPOROOT}/securedrop/scripts/shredder" --interval 60 &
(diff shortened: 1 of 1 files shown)


eloquence (approved):
Nice! I'm pretty sure this used to work in the dev env, maybe a recent regression? Fix works as expected (checked test plan boxes directly in PR).

legoktm:
> Nice! I'm pretty sure this used to work in the dev env, maybe a recent regression?

Yeah, sorry for not making that clear, it's a regression from 9de62c7913431e91ae5c4c3144934f82f7909aa4.

PR #6673, opened 2022-11-08
Add Makefile helper command to return current dev OTP code

### LINKED ISSUE #6621: Add a background TOTP generator to the dev environment
## Description

The JI requires 2FA using TOTP or HOTP, this adds a little bit of extra complexity for devs when working in the dev environment in particular. Since the TOTP secrets for default accounts are known, it should be possible to add a job that periodically outputs the token values for journalist accounts to Docker's stdout, saving devs the hassle of using oathtool or an authentication app when working on the JI.

## User Research Evidence
n/a - suggested as a hackathon prep idea

## User Stories
As a dev, I don't want to have to manage TOTP secrets for accounts in the dev environment in the same was as I would "real" ones.

eaon:
## Description of Changes

With this patch, `make otp` returns the current OTP code used with the test users set up by the development environment.

It also opportunistically copies the code with `wl-copy` (Wayland), `xclip` (X11) or `pbcopy` (macOS)

Fixes #6621

## Testing

- Run `make otp`
  - [x] If `oathtool` is installed/in your `$PATH`, the command returned the code
  - [x] If `oathtool` is not installed/in your `$PATH`, you're asked to installed oathtool
- Running `make otp` on a system using Wayland
  - [ ] If `wl-copy` is installed and Wayland is used, the otp code is also in your clipboard
- Running `make otp` on X11
  - [x] If `xclip` is installed and X11 is used, the otp code is also in your clipboard
- Runing `make otp` on macOS
  - [ ] The otp code is in your clipboard

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -274,6 +274,9 @@ clean:  ## DANGER! Delete all uncommitted files, virtual machines, Onion address
 	@$(SDROOT)/devops/clean
 	@echo
 
+.PHONY: otp
+otp: ## Show (and opportunistically copy) the current development OTP (to the clipboard)
+	@$(SDROOT)/devops/scripts/otp-code.sh
 
 #########
 #
--- a/devops/scripts/otp-code.sh
+++ b/devops/scripts/otp-code.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+## Show (and opportunistically copy) the current development OTP (to the
+## clipboard)
+
+if [ -n "$(command -v oathtool)" ]; then
+    OTP=$(oathtool --totp --base32 JHCOGO7VCER3EJ4L)
+    echo "$OTP"
+
+    COPY=$(command -v wl-copy)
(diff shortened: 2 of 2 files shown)


legoktm (dismissed):
If there's no command-line clipboard thing installed the failure mode is pretty reasonable:

```
user@dev ~/g/f/securedrop> make otp
836234
make: *** [Makefile:279: otp] Error 1
```

I tested with xclip and it worked fine, assuming you've tried it with wl-clip and pbcopy.

eaon:
Well, that error wasn't supposed to happen, so I added an `exit 0` :smile:

As for Wayland, yes this works for me on Wayland. `pbcopy` / macOS compatibility @cfm could confirm (I haven't tried) since we've been fiddling with the mac dev environment today anyway.

legoktm (approved):

PR #6679, opened 2022-11-12
Remove pyotp as a dependency

nabla-c0d3:
## Status

Ready

## Description of Changes

This PR removes pyotp as a dependency, for https://github.com/freedomofpress/securedrop/issues/5613 .

- [X] I have updated AppArmor rules to include the change

## Testing

What I tested using `make dev`:

### TOTP

- [x] Can login using a valid token.
```INFO in main: 'journalist' logged in with the two-factor code 854450```
- [X] Cannot login using an invalid token.
```ERROR in utils: Login for 'journalist' failed: Token verification failed```
- [X] Cannot login using the same token twice:
```ERROR in utils: Login for 'journalist' failed: Token already used```
- [X] Can reset 2fa to TOTP and then login.

### HOTP

- [X] Can login using a valid token.
- [X] Can reset 2fa to HOTP and then login.

### CODE REVIEW

File: securedrop/tests/test_2fa.py [hunk 1]
```diff
@@ -1,165 +0,0 @@
-# -*- coding: utf-8 -*-
```
nabla-c0d3: Refactored and moved to test_two_factor_in_apps.py.

securedrop/tests/test_integration.py
```diff
@@ -675,42 +670,6 @@ def test_user_change_password(journalist_app, test_journo):
             ins.assert_redirects(resp, "/")
 
 
-def test_login_after_regenerate_hotp(journalist_app, test_journo):
```
nabla-c0d3, line 664: Moved to test_two_factor_in_apps.py.

securedrop/tests/test_journalist.py
```diff
@@ -73,22 +73,6 @@ def _login_user(app, username, password, otp_secret, success=True):
     assert (session.get_user() is not None) == success
 
 
-@pytest.mark.parametrize("otp_secret", ["", "GA", "GARBAGE", "JHCOGO7VCER3EJ4"])
-def test_user_with_invalid_otp_secret_cannot_login(journalist_app, otp_secret):
```
nabla-c0d3, line 65: Moved to test_two_factor_in_apps.py.

zenmonkeykstop:
Looking at this now. Some review notes for my own edification:

- [ ] using same crypto and defaults as current implementation?
- [ ] (related to above) generating same tokens as current implementation? (so users don't have to reroll secrets on server upgrade)
- [ ] New secrets are all 160-bit but old 80-bit secrets still work? (at some stage we should require a migration)
- [ ] TOTP and HOTP functionality works as expected?

(AlsoFYI  @nabla-c0d3, I'll rebase it to catch the conflicts below - there isn't anything serious AFAICT)

zenmonkeykstop:
Rebased to pull in a regression fix from `develop`. Will retest in a bit.

zenmonkeykstop (approved):
- [x] tested base functionality in dev env, :heavy_check_mark: 
- ran upgrade scenario against 2.5.1 prod VM setup, confirmed:
  - [x] TOTP and HOTP secrets generated in 2.5.1 work against this branch
  - [x] New TOTP and HOTP secrets can be generated and produce valid tokens
  - [x] New secrets are the correct length

Based on the above, and on visual review and comparison with pyotp@2.6.0, this looks to be a seamless replacement for the relevant functionality!

LGTM.

PR #6680, opened 2022-11-12
Switch deb-tests job to use plain Debian container

### LINKED ISSUE #6671: deb-tests failing because of enchant missing in jammy
https://circleci.com/gh/freedomofpress/securedrop/65772

It appears the `cimg/python:3.8` image switched from focal to jammy sometime in the past week. jammy only has `enchant-2` (which pylint/pyenchant is compatible with).

But we want to move away from the CircleCI images anyways, and given that I believe most of us are building packages in Debian VMs, I'll switch it to a bullseye image.

@eaon and I were also initially confused exactly how enchant gets used, it's apparently run via pylint after some git log digging. I think if we install `pylint[spelling]` instead of `pyenchant` directly, it'll be clearer.

legoktm:
## Status

Ready for review

## Description of Changes

The cimg/python:3.8 container unexpectedly switched to jammy within the past week, breaking CI since it doesn't have enchant (it's now enchant-2).

We're moving away from those images anyways, so use a plain debian:bullseye image and install the correct package.

*Update:* also includes:
- a change to the full dockerfile to update the apt cache before installing packages, to clear a CI issue
- an update of the wheel dependency for the Docker-installed version to clear safety issue 51499.

Fixes #6671.

## Testing

* [x] CI `deb-tests` passes

## Deployment

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -292,12 +292,12 @@ jobs:
 
   deb-tests:
     docker:
-      - image: cimg/python:3.8
+      - image: debian:bullseye
     environment:
       LC_ALL: C.UTF-8
       LANG: C.UTF-8
     steps:
-      - run: sudo apt-get update && sudo apt-get install -y make virtualenv enchant jq python3-dev build-essential rsync
+      - run: apt-get update && apt-get install -y git docker.io make virtualenv enchant-2 jq python3-dev build-essential rsync
       - checkout
       - setup_remote_docker
       - run:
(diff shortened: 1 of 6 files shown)


zenmonkeykstop:
This is failing because there are security updates pending for the builder image, not because of any issues with the changes in the PR, so we have a couple of options:

1.  update the builder image as well as part of this PR (annoying/time consuming but harmless)
2.  approve despite the `deb-tests` failure given that the security update check is the only error and is expected.

I'm more in favour of (1) tbh. Can add a commit if that's OK.

eaon (approved):
See ya CircleCI images

PR #6674, opened 2022-11-08
fix: determine and add `safe.directory` entry for `$REPOROOT` directly from the filesystem

cfm:
## Status

Ready for review

## Description of Changes

Fixes #6672 by:
* determining `$REPOROOT` directly from the filesystem passes; and
* explicitly adding a `safe.directory` entry to placate Git versions that are paranoid about heterogeneous ownership between host and containers.

Thanks to @eaon for pairing and @legoktm for good instincts.

## Testing

* [x] CI passes.
* `make dev` starts normally:
    * [x] on Linux
    * [x] ideally on macOS too

## Deployment

Development-only.

PR #6682, opened 2022-11-13
Fix run-test to also use git safe.directory

audiodude:
## Status

Ready for review

## Description of Changes

Adds required git config line to run-test shell script. This makes it so that the script works properly on macOS, using the same settings as was added in #6674

## Testing

Easy to test by doing simple `make test`.

## Deployment

None

## Checklist


legoktm (approved):
Saw this in action on a mac so I know it works. Thanks!

PR #6683, opened 2022-11-13
Configure hooks when running `make venv`

eaon:
## Description of Changes

Instead of copying files to the `.git` directory, it configures git to use a hook directory. Creates [parity with `securedrop-client`](https://github.com/freedomofpress/securedrop-client/blob/f0e63dd264c7152446cd7cbc842b4bdfb5c64e95/Makefile#L34-L38). Also assumes that if developers haven't added a `venv`, they probably didn't add hooks either (implying that `make venv` in the hooks would just add unnecessary on every commit)

## Testing

Wipe away your `.venv` folder and run:

- [ ] `make venv` creates a virtual environment
- [ ] `git config core.hooksPath` returns `.githooks`
- [ ] `pre-commit` is run on every subsequent attempt to commit

## Checklist

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -8,7 +8,7 @@ set -eo pipefail
 if [[ -n "$PY_FILES" ]]; then
     # set up the virtualenv if it's not already available
     if [[ ! -v VIRTUAL_ENV ]]; then
-        make venv && source .venv/bin/activate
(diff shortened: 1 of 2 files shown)


zenmonkeykstop (approved):
Test plan checks out, and this is a cleaner approach overall. LGTM

PR #6685, opened 2022-11-13
Make venv use python3.8 opportunistically, but default to python3

eaon:
Our virtual environment generally assumes that developers use Debian stable - that's not necessarily the case though, so this commit introduces an opportunistic approach to using the same Python version.

## Testing

On a machine that defaults to a Python version higher than 3.9, but has 3.9 installed: remove your `.venv` folder:

- [ ] `make venv` creates a virtual environment with Python 3.9

## Checklist

### DIFF
--- a/devops/scripts/boot-strap-venv.sh
+++ b/devops/scripts/boot-strap-venv.sh
@@ -17,7 +17,7 @@ venv_instructions() {
 }
 
 function virtualenv_bootstrap() {
-    PYTHON_VERSION=3
+    PYTHON_VERSION="3.8"
     VIRTUAL_ENV="${VIRTUAL_ENV:-}"  # Just to get around all the "set -u"
     if [ -n "$VIRTUAL_ENV" ]
     then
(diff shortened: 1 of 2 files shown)


devops/scripts/boot-strap-venv.sh
```diff
@@ -48,12 +48,12 @@ function virtualenv_bootstrap() {
...
         then
-            p=$(which "python${PYTHON_VERSION}")
-            echo "Creating Python ${PYTHON_VERSION} virtualenv in ${VENV}"
+            p=$(which "python${PYTHON_VERSION}" 2> /dev/null || which python3)
+            echo "Creating ${p} virtualenv in ${VENV}"
             virtualenv -p "${p}" "${VENV}"
         fi
 
-        "${VENV}/bin/pip" install -q -r "securedrop/requirements/python${PYTHON_VERSION}/develop-requirements.txt"
+        "${VENV}/bin/pip" install -q -r "securedrop/requirements/python${PYTHON_VERSION::1}/develop-requirements.txt"
```
legoktm: Cool trick, but I think we can just hardcode the 3 now. It's already hardcoded in other places too

legoktm:
Hmm, shouldn't we prefer 3.8 over 3.9 since that's what CI uses and what we use in production with focal? (I've been using 3.8 for that reason)

legoktm (changes requested):
See above.

legoktm:
Code looks good, but shouldn't we prefer 3.8 over 3.9?

eaon:
Makes sense, disregard the branch name now, I changed it to use `python3.8` instead!

legoktm (approved):
LGTM!

PR #6695, opened 2022-11-30
Reorganize and refactor requirements.txt files

### LINKED ISSUE #6546: Fix up requirements.txt files used for boostrapping
## Description

They're kind of split up weirdly right now. For example, `docker-requirements.in` is:
```
pip>=21.3
setuptools>=56.0.0
wheel
```

But in the Dockerfile we do:
```
    /opt/venvs/securedrop-app-code/bin/pip3 install --no-deps --require-hashes -r requirements/python3/docker-requirements.txt && \
    /opt/venvs/securedrop-app-code/bin/pip3 install --no-deps --require-hashes -r requirements/python3/test-requirements.txt && \
    /opt/venvs/securedrop-app-code/bin/pip3 install --no-deps --require-hashes -r requirements/python3/securedrop-app-code-requirements.txt 
```

It probably should be named bootstrap-requirements.txt or something.

Then we have `translation-requirements.in`, which is:
```
setuptools-scm>=6.0.1
```

Another bootstrapping thing, you'd expect to find things like `pybabel` in translation-requirements, but nope.

### LINKED ISSUE #6619: Rename securedrop-app-code-requirements.txt to just requirements.txt
As part of reproducible wheels (#5901), we're adopting the same workflow as the SDW packages. One gotcha is that the scripts (reasonably) expect the file to be named "requirements.txt".

Given that all the other files are already named with a prefix, it seems reasonable to move our real, production dependencies to plain "requirements.txt".

legoktm:
## Status

Ready for review

## Description of Changes

* Rename securedrop-app-code-requirements.txt to requirements.txt
  * All of our reproducible wheels scripts in securedrop-builder expect the
    file to be named "requirements.txt". Renaming this file seems like the
    simplest option and it matches the convention that the unprefixed file
    is the main production dependencies. 
* Fix up requirements.txt files used for boostrapping and translations
  * Rename docker-requirements to bootstrap-requirements so it's clearer
    exactly what it's doing.
  * And then have translation-requirements actually install the translation
    tools rather than more bootstrapping. This allows us to speed up
    building translations during packaging since we only need pybabel
    installed. This should shave off about 1-2 minutes from package building.
* Use SLIM_BUILD=1 when running `make update-python3-requirements`

Fixes #6619.
Fixes #6546.

## Testing

* [x] CI passes
* [ ] No dependency version/hashes changed
* [ ] `make dev` continues to work
* [ ] `make build-debs` works, deb contains same translation files (diffoscope should show translation files as being identical)
## Deployment

Any special considerations for deployment? Should be a no-op

## Checklist


cfm (approved):
Test plan checks out!  Thanks, @legoktm.

* [x] CI passes
* [x] No dependency version/hashes changed

`babel` and `setuptools` have moved between requirements lists, but their pinned versions and hashes are unchanged.
* [x] `make dev` continues to work
* [x] `make build-debs` works, deb contains same translation files (diffoscope should show translation files as being identical)

PR #6701, opened 2022-12-06
Fix the SecureDrop ASCII art

legoktm:
## Status

Ready for review

## Description of Changes

* Use a raw string literal so we don't have to escape backslashes. This makes it so the source is identical to what is printed.
* Drop noqa comment now that line isn't too long (and also it was getting printed)
* Shift nearly everything leftward one character to fix the S
* Increase the width of the top of the D by one character

## Testing

Run `make dev`, observe that it looks better now

## Deployment

Any special considerations for deployment? Dev-only

## Checklist


eloquence:
:shipit: 

Screenshot from 2022-12-06 16-22-58

eloquence (approved):

PR #6703, opened 2022-12-07
Update builder to 2022_12_07

legoktm:
## Status

Ready for review

## Description of Changes

Routine builder update

## Testing

* [ ] CI passes

### DIFF
--- a/molecule/builder-focal/image_hash
+++ b/molecule/builder-focal/image_hash
@@ -1,2 +1,2 @@
-# sha256 digest quay.io/freedomofpress/sd-docker-builder-focal:2022_11_23
-ad77b4d808e9f31d4832c0a159ed0a1e75d856a2f8c5d7f33134fd8e22786cd9
+# sha256 digest quay.io/freedomofpress/sd-docker-builder-focal:2022_12_07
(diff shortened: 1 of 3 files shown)


legoktm:
OK, I manually examined it and the issue is:
```
root@af15f49aab6b:/# apt-get -s dist-upgrade -oDir::Etc::Sourcelist=/tmp/sec.list  |grep "^Inst" |grep -i security
Inst libjbig0 (2.1-3.1ubuntu0.20.04.1 Ubuntu:20.04/focal-security [amd64])
```

But that package isn't even installed!
```
root@af15f49aab6b:/# dpkg -l | grep libjbig0
root@af15f49aab6b:/# 
```

It would be pulled in if we ran a dist-upgrade, but we're not doing that. So I think we can safely ignore the failure and move on.

zenmonkeykstop (approved):
Agreed that it's OK to proceed, said library isn't used in build (otherwise not installing it might be a problem :) )

legoktm:
The reason the dist-upgrade got confused is because our pin file (https://github.com/freedomofpress/securedrop/blob/49fa4b0347f3ad1d83b7b77cc594082ba83e33fb/molecule/builder-focal/dh-virtualenv.pref) is still set for impish, so jammy isn't actually pinned and it wants to upgrade most of the container over to some weird focal/jammy hybrid.

PR #6707, opened 2022-12-11
Make it official that only sqlite is supported as the DB engine

### LINKED ISSUE #6700: Officially only support SQLite
## Description

Theoretically we support non-SQLite databases:
```python3
        if self.DATABASE_ENGINE == "sqlite":
            db_uri = self.DATABASE_ENGINE + ":///" + self.DATABASE_FILE
        else:
            if self.DATABASE_USERNAME is None:
                raise RuntimeError("Missing DATABASE_USERNAME entry from config.py")
            if self.DATABASE_PASSWORD is None:
                raise RuntimeError("Missing DATABASE_PASSWORD entry from config.py")
            if self.DATABASE_HOST is None:
                raise RuntimeError("Missing DATABASE_HOST entry from config.py")
            if self.DATABASE_NAME is None:
                raise RuntimeError("Missing DATABASE_NAME entry from config.py")
```

But in reality, we don't, because some migrations use raw SQL, and all of the ansible and packaging logic expects SQLite. 

For simplicity, we should stop pretending we support other database backends and only care about SQLite.

(Also discussed at https://github.com/freedomofpress/securedrop/pull/6563#discussion_r1038997400)

nabla-c0d3:
## Status

Ready

## Description of Changes

This PR removes remnants of mysql support, so that the code and app only support sqlite. Fixes https://github.com/freedomofpress/securedrop/issues/6700 .


legoktm:
Did a quick skim, LGTM. I put this on the agenda for tomorrow's team meeting just to make sure there's no concerns or history that I/we are unaware of.

zenmonkeykstop (approved):
LGTM! just rebased to pull in a CI fix - once it's green this is good to go.

PR #6711, opened 2022-12-14
Updated testinfra checks for config file ownership:

### LINKED ISSUE #6710: Nightly broken due to mismatch between old and new perms on config.py
## Description

Current testinfra tests expect /var/www/securedrop/config.py to be 0o600 and www-data:www-data. it is now 0o640 and root:www-data

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #6710 .
Updates tests for /var/www/securedrop/config.py:

- file should be owned by root:www-data
- perms are 640, not 600

## Testing

- [x] CI is green including `staging-test-with-rebase` job
- [ ] make testinfra passes locally


eaon (approved):
Ran into issues with my libvirt staging env again (should have known), but there's no reason not to believe CI here so I'm sneaky and am approving/merging anyway :nerd_face:

PR #6720, opened 2023-01-10
Apply pyupgrade to Python files

### LINKED ISSUE #6699: Add pyupgrade to our toolchain
## Description

pyupgrade is a tool that automatically rewrites code to use modern syntax. https://github.com/asottile/pyupgrade has a detailed list with examples.

This would be similar to black/isort, in which we'd have a command like `make pyupgrade` that applies all the changes, and then another like `make pyupgrade-check` which verifies there are no unapplied changes. Note that black should always run after pyupgrade changes things.

## User Stories

* As a developer, I want to automatically update Python code to modern patterns/syntax so I don't have to do it manually

legoktm:
## Status

Ready for review

## Description of Changes

Generated with `pyupgrade --py38-plus **.py`, with manual adjustments for unused imports and then running black+isort.

Mostly covers:
* migration to f-strings
* removal of coding: utf-8 comments
* io.open() -> open()

Fixes #6699.

## Testing

* [ ] visual review, especially for f-strings, it should be simpler, not harder, to read
* [ ] CI passes

## Deployment

Any special considerations for deployment? Should all be no-ops.

## Checklist


securedrop/journalist_app/main.py
```diff
@@ -141,7 +140,7 @@ def reply() -> werkzeug.Response:
             return redirect(url_for("col.col", filesystem_id=g.filesystem_id))
 
         g.source.interaction_count += 1
-        filename = "{0}-{1}-reply.gpg".format(
+        filename = "{}-{}-reply.gpg".format(
```
zenmonkeykstop, line 143: Huh, looks like there's a format case it misses, I guess you'd have to repeatedly apply it until you hit a steady state.

zenmonkeykstop (approved):
Changes look consistent. Approving based on visual review.

securedrop/journalist_app/main.py
legoktm: I think this falls under "pyupgrade is intentionally timid and will not create an f-string if it would make the expression longer or if the substitution parameters are sufficiently complicated (as this can decrease readability)." (from https://github.com/asottile/pyupgrade#f-strings)

You end up with a 92-char line:
```
        filename = f"{g.source.interaction_count}-{g.source.journalist_filename}-reply.gpg"
```

I can do it manually, though it's personally borderline on whether this decreases readability or not (yes my "old people" rant is that I like 80-char line lengths still).

(Running it multiple times doesn't help, I did that enough times while setting up this PR :))

Issue #6314, opened 2022-03-02
review options for accessibility linting (versus documentation)


cfm:
**Require WCAG compliance along with other linting enforced automatically in CI.**  Here's [an overview of how to do so](https://marcysutton.github.io/a11y-and-ci/#/) and the W3C's [list of such tools](https://www.w3.org/WAI/ER/tools/?q=wcag-20-w3c-web-content-accessibility-guidelines-20&q=command-line-tool), of which [Pa11y](https://pa11y.org/) and [AccessLint](https://accesslint.com/) look especially promising.

_Originally posted by @cfm in https://github.com/freedomofpress/securedrop/issues/5743#issuecomment-861867074_

skyvine:
# Machine Testing

Following the conversation in #6725, I took a look at pa11y/pa11y-ci and how it could be used for testing in this project. It turns out that pa11y is a wrapper around puppeteer (to manage the browser instance) and [HTML CodeSniffer](http://squizlabs.github.io/HTML_CodeSniffer/) (to run tests on the page received by the browser) which provides some benefits over using them directly:

- Management of browser setup/teardown is transparent (as long as Chrome is installed)
- Configuring the browser & the tests can happen in one place (the pa11y config file)
- A variety of built-in output formats, including json for programmatic processing
- Built-in system for running user actions before running tests
- Filtering error reports
  - Ignore specific error codes
  - Control reporting of notices & warnings (neither are considered actionable by pa11y maintainers)

Unfortunately, pa11y is hardcoded to Chrome. It might be possible to convert it to firefox (see additional notes below); however, even in this case, it seems like using pa11y to manage puppeteer to manage a separate firefox instance would add complexity to the test module, with both pa11y and the existing selenium wrappers implementing similar functionality. HTML CodeSniffer is a client-side script, injecting it using selenium would result in a less complex system overall.

On the other hand, HTML CodeSniffer does not have a built-in method for reporting results (it just saves them in memory), so reporting would need to be handled separately. Reporting seems simlpe, because all we would need to do is save the results into a cookie, and I see at least one [example of cookie-reading already in the test suite](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/tests/functional/app_navigators/journalist_app_nav.py#L124). The code is minimal and readable.

<details>
<summary>Example json output of a pa11y run on the source login page</summary>

```json
> pa11y http://127.0.0.1:8080 --reporter json
[{
[...]

PR #6725, opened 2023-01-16
Pull descriptive titles from headings

### LINKED ISSUE #6313: surface view-level `h1`s (or equivalents) in page `title`s (WCAG 2.4.2)
Per #5972:

> * The page `<title>` must be accurate and informative.
> * If a page is the result of a user action or scripted change of context, the text of the `<title>` should describe the result or change of context to the user.
> * The `<title>` should be concise.
> * The page `<title>` should be unique, if possible.
> * Unique information should come first in the `<title>`.
> * The page `<title>` should match (or be very similar to) the top heading in the main content. 

A few belated notes per <https://github.com/freedomofpress/securedrop/pull/6725#pullrequestreview-1252731682>:

1. Since the SecureDrop logo is our default `h1` across views, the "equivalents" here will usually be the per-view `h2`.
2. If the SecureDrop Web apps were larger (had more views), or if titles were more complicated or personalized, it might be worth setting the page `title` automatically from these headings.  As it is, a simple fillable Jinja `{% block title %}Your Title Here{% endblock %}` is adequate, even though it requires developers to keep the title in sync manually.
3. ~[We do need sign-off from our security engineers and UX designer](https://docs.google.com/document/d/1jZlKco2YgOrrI4PmTkG_Q8lKuV5VxCds/edit?disco=AAAAIplLpt8), to whom I'll flag this question in a comment below.~  (resolved per <https://github.com/freedomofpress/securedrop/issues/6313#issuecomment-1411389401>)

skyvine:
## Status

Work in progress

## Description of Changes

Fixes #6313.

Adds a python function that grabs the first h1 statement in the html and uses that for the page title, then calls that function from the base templates. Also updated some HTML files so that an appropriate h1 heading exists. See the commit message for details.

## Testing

I used `make dev` and browsed pages with it, as far as the cilent is concerned the only change is what static text is sent over so I'm not overly concerned about browser compatibility, but let me know if I need to do more. I did create a new user and select YubiKey to test a page where the h1 statement was in an if/else block, and it created the correct title in the else case.

## Deployment

2. New installs.

I don't think so, this is just a normal python function added to the code.

N/A

N/A

N/A

Choose one of the following:

Based on the state of the html files, I think that the new requirement of all pages needing a descriptive h1 block that takes into account users who depend on screen readers would be good to add to the documentation. I haven't started looking at how this is organized yet so it might make sense once I go over there, but if you have any initial advice I would be happy to hear it.

N/A, all newly imported modules are already available in the project

### DIFF
--- a/securedrop/journalist_templates/base.html
+++ b/securedrop/journalist_templates/base.html
@@ -4,7 +4,7 @@
 <head>
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <title>{{ g.organization_name }}</title>
+  <title>{{ get_descriptive_title() }} | {{ g.organization_name }}</title>
 
   <link rel="stylesheet" href="/static/css/journalist.css">
 
--- a/securedrop/source_templates/index.html
+++ b/securedrop/source_templates/index.html
@@ -3,9 +3,9 @@
 
 <head>
   {% if g.organization_name == "SecureDrop" %}
-  <title>{{ g.organization_name }} | {{ gettext('Protecting Journalists and Sources') }}</title>
+  <title>{{ gettext('Portal') }} | {{ g.organization_name }} | {{ gettext('Protecting Journalists and Sources') }}</title>
   {% else %}
-  <title>{{ g.organization_name }} | {{ gettext('SecureDrop') }}</title>
+  <title>{{ gettext('Portal') }} | {{ g.organization_name }} | {{ gettext('SecureDrop') }}</title>
   {% endif %}
 
   <link rel="stylesheet" href="/static/css/source.css">
--- a/securedrop/utils.py
+++ b/securedrop/utils.py
@@ -1,0 +1,24 @@
+from itertools import count
+from logging import Logger
+from typing import Callable
+
+from jinja2.runtime import Context
+from jinja2.utils import pass_context
+
+
+def create_get_descriptive_title_function(logger: Logger) -> Callable[[], str]:
+    @pass_context
+    def get_descriptive_title(context: Context) -> str:
+        for block_func in context.blocks["body"]:
+            block = list(block_func(context))
+            for (possibly_h1_tag, next_index) in zip(block, count(1)):
+                if "<h1" in possibly_h1_tag:
+                    return block[next_index]
+
+        logger.error(f"No h1 block found on page {context.name}, returning empty string!")
+        return ""
+
+    # The mypy ignore is because the pass_context decorator passes in the context automatically,
+    # so the type hint is correct but mypy doesn't understand this. The noqa is because flake6
+    # throws an error when the mypy ignore specifies the error code.
+    return get_descriptive_title  # type: ignore [return-value] # noqa: F723
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -14,6 +14,7 @@ from journalist_app.sessions import Session, session
 from journalist_app.utils import get_source
 from models import InstanceConfig
 from sdconfig import SecureDropConfig
+from utils import create_get_descriptive_title_function
 from werkzeug import Response
 from werkzeug.exceptions import HTTPException, default_exceptions
 
@@ -94,6 +95,9 @@ def create_app(config: SecureDropConfig) -> Flask:
     app.jinja_env.trim_blocks = True
     app.jinja_env.lstrip_blocks = True
     app.jinja_env.globals["version"] = version.__version__
+    app.jinja_env.globals["get_descriptive_title"] = create_get_descriptive_title_function(
+        app.logger
+    )
     app.jinja_env.filters["rel_datetime_format"] = template_filters.rel_datetime_format
     app.jinja_env.filters["filesizeformat"] = template_filters.filesizeformat
     app.jinja_env.filters["html_datetime_format"] = template_filters.html_datetime_format
--- a/securedrop/source_app/__init__.py
+++ b/securedrop/source_app/__init__.py
@@ -17,6 +17,7 @@ from sdconfig import SecureDropConfig
 from source_app import api, info, main
 from source_app.decorators import ignore_static
 from source_app.utils import clear_session_and_redirect_to_logged_out_page
+from utils import create_get_descriptive_title_function
 
 
 def get_logo_url(app: Flask) -> str:
@@ -70,6 +71,9 @@ def create_app(config: SecureDropConfig) -> Flask:
     app.jinja_env.globals["version"] = version.__version__
     # Exported to source templates for being included in instructions
     app.jinja_env.globals["submission_key_fpr"] = config.JOURNALIST_KEY
+    app.jinja_env.globals["get_descriptive_title"] = create_get_descriptive_title_function(
+        app.logger
+    )
     app.jinja_env.filters["rel_datetime_format"] = template_filters.rel_datetime_format
     app.jinja_env.filters["nl2br"] = template_filters.nl2br
     app.jinja_env.filters["filesizeformat"] = template_filters.filesizeformat
--- a/securedrop/source_templates/base.html
+++ b/securedrop/source_templates/base.html
@@ -6,9 +6,9 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <meta name="robots" content="noindex,nofollow">
   {% if g.organization_name == "SecureDrop" %}
-  <title>{{ g.organization_name }} | {{ gettext('Protecting Journalists and Sources') }}</title>
+  <title>{{ get_descriptive_title() }} | {{ g.organization_name }} | {{ gettext('Protecting Journalists and Sources') }}</title>
   {% else %}
-  <title>{{ g.organization_name }} | {{ gettext('SecureDrop') }}</title>
+  <title>{{ get_descriptive_title() }} | {{ g.organization_name }} | {{ gettext('SecureDrop') }}</title>
   {% endif %}
 
   <link rel="stylesheet" href="/static/css/source.css">
--- a/securedrop/source_templates/error.html
+++ b/securedrop/source_templates/error.html
@@ -1,6 +1,6 @@
 {% extends "base.html" %}
 {% block body %}
-<h2>{{ gettext('Server error') }}</h2>
+<h1>{{ gettext('Server error') }}</h1>
 
 <p>{{ gettext('Sorry, the website encountered an error and was unable to complete your request.') }}</p>
 
--- a/securedrop/source_templates/flash_message.html
+++ b/securedrop/source_templates/flash_message.html
@@ -1,4 +1,4 @@
 {%- if declarative %}
-<h2>{{ declarative }}</h2>
+<h1>{{ declarative }}</h1>
 {% endif -%}
 <p {% if not declarative %} class="icon"{% endif %}>{{ msg_contents }}</p>
--- a/securedrop/source_templates/generate.html
+++ b/securedrop/source_templates/generate.html
@@ -2,7 +2,7 @@
 {% import 'utils.html' as utils %}
 
 {% block body %}
-<h2>{{ gettext('Get Your Codename') }}</h2>
+<h1>{{ gettext('Get Your Codename') }}</h1>
 
 <p id="codename-explanation" class="explanation">
   {{ gettext('A <em>codename</em> in SecureDrop functions as both your username and your password.') }}
--- a/securedrop/source_templates/login.html
+++ b/securedrop/source_templates/login.html
@@ -3,7 +3,7 @@
 
 {% include 'flashed.html' %}
 
-<h2>{{ gettext('Enter Codename') }}</h2>
+<h1>{{ gettext('Enter Codename') }}</h1>
 
 <form method="post" action="{{ url_for('main.login') }}" autocomplete="off">
   <input name="csrf_token" type="hidden" value="{{ csrf_token() }}">
--- a/securedrop/source_templates/logout.html
+++ b/securedrop/source_templates/logout.html
@@ -5,7 +5,7 @@
     {{ gettext('LOG IN') }}</a>
 </nav>
 <section>
-  <h2>{{ gettext('One more thing...') }}</h2>
+  <h1>{{ gettext('Logout - Additional Action Needed') }}</h1>
   <p>
     {{ gettext('Click the <img src={icon} alt="" width="16" height="16">&nbsp;<b>New Identity</b> button in your Tor Browser\'s toolbar. This will clear your Tor Browser activity data on this device.').format(icon=url_for('static', filename='i/torbroom.png'))  }}
   </p>
--- a/securedrop/source_templates/lookup.html
+++ b/securedrop/source_templates/lookup.html
@@ -10,10 +10,10 @@
 
 <section aria-labelledby="submit-heading">
   {% if allow_document_uploads %}
-  <h2 id="submit-heading">{{ gettext('Submit Files or Messages') }}</h2>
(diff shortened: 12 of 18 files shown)


skyvine:
I'm not sure why my local `make lint` didn't catch the import order error, fixed now. Waiting on `make test` to succeed again but I think it should be fine.

securedrop/utils.py
```diff
@@ -0,0 +1,21 @@
...
+from logging import Logger
+from typing import Callable
+
+from jinja2.runtime import Context
+from jinja2.utils import pass_context
+
+
+def create_get_descriptive_title_function(logger: Logger) -> Callable[[], str]:
+    @pass_context
+    def get_descriptive_title(context: Context) -> str:
```
gonzalo-bulnes, line 11: The [latest `mypy` failure](https://app.circleci.com/pipelines/github/freedomofpress/securedrop/5355/workflows/0c5e0b21-a8d6-41c5-a2a3-f5096745ce4b/jobs/66870?invite=true#step-110-451) is legit. The type hint for the return value does indeed not match the signature of the returned function. ([docs](https://mypy.readthedocs.io/en/stable/kinds_of_types.html#callable-types-and-lambdas))

```suggestion
def create_get_descriptive_title_function(logger: Logger) -> Callable[[Context], str]:
    @pass_context
    def get_descriptive_title(context: Context) -> str:
```

securedrop/journalist_templates/base.html
```diff
@@ -4,7 +4,7 @@
 <head>
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <title>{{ g.organization_name }}</title>
+  <title>{{ get_descriptive_title() }} | {{ g.organization_name }}</title>
```
gonzalo-bulnes, line 7: I _suppose_ that the `Context` argument gets passed automatically by Jinja? I don't actually know about that @saffronsnail, maybe you do!

securedrop/source_templates/index.html
```diff
@@ -3,9 +3,9 @@
 
 <head>
   {% if g.organization_name == "SecureDrop" %}
-  <title>{{ g.organization_name }} | {{ gettext('Protecting Journalists and Sources') }}</title>
+  <title>{{ gettext('Portal') }} | {{ g.organization_name }} | {{ gettext('Protecting Journalists and Sources') }}</title>
   {% else %}
-  <title>{{ g.organization_name }} | {{ gettext('SecureDrop') }}</title>
+  <title>{{ gettext('Portal') }} | {{ g.organization_name }} | {{ gettext('SecureDrop') }}</title>
```
gonzalo-bulnes, line 8: `Portal`? Can you give a little bit of context on this addition @saffronsnail? (Apologies if I'm missing any previous conversation on the topic!)

securedrop/journalist_templates/base.html
skyvine: Yeah, I wrote the type hint that way because the `pass_context` generator adds it automatically.

gonzalo-bulnes (commented):
I'll let the folks who were involved in creating https://github.com/freedomofpress/securedrop/issues/6313 add their reviews, but this change set look great to me. Thank you @saffronsnail ! :rocket:

securedrop/journalist_templates/base.html
gonzalo-bulnes: Gotcha. Does **mypy** complain if you add the type hint but don't pass the argument explicitly? If that's the case, I think it's fair to `# type: ignore [not-sure-what-the-error-name-is]` where most relevant.

securedrop/source_templates/index.html
skyvine: I'm very open to suggestions about a better title for this page. Most of the pages have a descriptive built-in header, but the landing page for the source interface has 2 sections, one for creating a new account and one for logging in to an existing account. I feel like titling the page "Log In" is not technically accurate because creating a new account is a distinct action from logging in, but I might be over-thinking it. For reference, the rendered page can be seen in the attached screenshot.
source-index

securedrop/utils.py
skyvine: As you noted below, this actually gets passed in automatically by jinja, so the type hint is correct. I just uploaded a new commit with a comment telling mypy to ignore this error and explaining the rationale.

securedrop/journalist_templates/base.html
gonzalo-bulnes: **Note** the space between `ignore` and `[...]` :wink: 

For line length: `# noqa: E501`  it needs to go _after_ the type ignore comment, e.g. `# type: ignore [has-type]  # noqa: E501`.  **Edit**: and yes, it needs the `#` sign!

securedrop/source_templates/index.html
gonzalo-bulnes: I think you've got a good point. And "Log in or sign up" doesn't really do it because "sign up" probably gives the wrong idea. "Portal" isn't bad. :+1:

securedrop/journalist_templates/base.html
skyvine: Oh, I moved the comment to just above the function before seeing this comment. If you'd prefer that the note is in the same line as the ignore directive, I can change it to use noqa.

skyvine: Oh, it looks like I'll need to do that anyway because just having the type directive makes it too long.

skyvine: I'm still leaving the comment above the function because I think it's nice for human-intended context to not be hidden past the edge of the screen, but still let me know if that should be moved for some reason.

gonzalo-bulnes:
@saffronsnail  :point_down: (the syntax is really not tolerant in this case ><')

> Note the space between `ignore` and `[...]` (https://github.com/freedomofpress/securedrop/pull/6725#discussion_r1071659123)

securedrop/journalist_templates/base.html
gonzalo-bulnes: Any syntax that work seems good to me for this, and you've got a good reason to pick the one you chose. :+1:

skyvine: Ok, so it looks like I get a syntax error if I specify both the mypy return code and the noqa return code, but if I only specify one of them it's fine for some reason. `securedrop/journalsit_app/sessions.py` only specifies the type code, but `securedrop/store.py` only specifies the noqa code, so it's not clear which one is preferred. I'm about to upload a commit that specifies type but I can switch it around if needed.

skyvine:
Also, I fixed the lint issues on my machine so I'm seeing the errors that CircleCI was throwing locally now. Future changes should be less noisy

securedrop/utils.py
```diff
@@ -0,0 +1,23 @@
...
+from logging import Logger
+from typing import Callable
+
+from jinja2.runtime import Context
+from jinja2.utils import pass_context
+
+
+# Note: return value type hint ignored because the returned function does not take in any
+# arguments, the pass_context decorator handles the Context argument automatically.
+def create_get_descriptive_title_function(logger: Logger) -> Callable[[], str]:  # type: ignore[return-value] # noqa: E501
```
gonzalo-bulnes:
```suggestion
def create_get_descriptive_title_function(logger: Logger) -> Callable[[], str]:  # type: ignore [return-value] # noqa: E501
```

This should make **mypy** happy.

securedrop/journalist_templates/base.html
gonzalo-bulnes: I just made [a suggestion](https://github.com/freedomofpress/securedrop/pull/6725#discussion_r1071677039) that should unlock you @saffronsnail. Sorry we're spread across multiple threads :face_with_spiral_eyes:

skyvine: Actually, I'm seeing different output when running flake8/mypy manually vs running `make lint`. I'll look at this more tomorrow.

securedrop/utils.py
gonzalo-bulnes: Example in action: https://github.com/freedomofpress/securedrop-client/blob/main/securedrop_client/gui/widgets.py#L673

securedrop/journalist_templates/base.html
skyvine: Aah, you know what, I didn't think to refresh the page as I was typing in updates so I didn't see some of your replies. Still getting used to GitHub. :sweat_smile:

It turns out I was putting the directive in the wrong spot: the error occurs on the line that returns the value, not on the function signature. Putting it on the correct line deals with the long line problem without any directives. Unfortunately, `flake8` reports a syntax error when the error code is specified in a `type: ignore` directive, so I told it to ignore F723. 

`make lint` passes locally now (*with* the space!), hopefully the server agrees.

skyvine:
Linter is happy now, thanks for the help wrangling it @gonzalo-bulnes!

The new upload fixes the test by adding the current contents of the header to the literal string it's checking against. I also considered changing the `==` on each line to `in`, but this seems inexact. I do not expect that the overhead of keeping this test in sync with the heading should be great, but let me know if I'm overlooking something.

On the topic of tests, that's what I'm looking at next. I changed my response to the documentation question in the PR template because I think this can actually be verified in the page layout tests. I've only just started looking at it, but it seems like these tests have access to the raw html, so I should be able to just check that the title contains the text in the first heading. This test will need to go into (almost) all of the pages, so my initial instinct is to add a base class (`CommonTests` or similar) that all tests derive from, and override this test in the one case where it does not apply (the source index). The other option would be to add a helper function to `securedrop/tests/functional/pagelayout/utils.py` and call it from every test file except for the source index.

I'm also going to add tests for the `get_descriptive_title` function to make sure it catches some edge cases (my initial implementation had issues with trailing/leading blankspace and header tags that included attributes, so I want to make sure these are checked at a minimum).

Should the tests be added in a separate commit or added to this one?

securedrop/utils.py
```diff
@@ -0,0 +1,24 @@
...
+                if "<h1" in possibly_h1_tag:
+                    return block[next_index]
+
+        logger.error(f"No h1 block found on page {context.name}, returning empty string!")
+        return ""
+
+    # The mypy ignore is because the pass_context decorator passes in the context automatically,
+    # so the type hint is correct but mypy doesn't understand this. The noqa is because flake6
+    # throws an error when the mypy ignore specifies the error code.
+    return get_descriptive_title  # type: ignore [return-value] # noqa: F723
```
gonzalo-bulnes, line 24: This looks good :+1: 

I don't quite understand what's wrong with `# type: ignore [return-value]` that requires `# noqa: F723` but I wouldn't loose sleep over it.

gonzalo-bulnes:
> Should the tests be added in a separate commit or added to this one?

Whether you add the tests to the same commit or a separate one is mostly a matter of personal preference. At least there are no strong conventions about that in our repos. With that said, I'd argue that adding the to the same commit allows to create a nicely self-contained unit, but do as you prefer.

_For further thought: if your PR contained 20+ commits, it would be way more likely that commits touch on distinct topics, and I'd suggest more strongly to keep commits as self-contained as possible for clarity of what relates to what. But in this PR that's not relevant._

gonzalo-bulnes:
I am not too familiar with this repo's test suite, so I'll let @cfm comment on your testing strategy. I like how you're thinking about it though.

skyvine:
I think I'm happy with the tests on the python code, which are in the new version of the commit.

For the tests that look at the HTML, I'm still thinking about how to proceed. Looking at the current code in `securedrop/tests/functional/pageslayout`, it seems like these are categorically different than the tests that I'm looking at. These are testing that there are specific behaviors when users take specific actions. The tests for this commit would be doing static validation. Static validation seems trivial with selenium and programmatic DOM, but that's just based on skimming the docs I haven't worked with these technologies before. In any case, I would like to propose a larger unit of work and see what y'all think. I have a considerable amount of free time over the next few months so I could work on this, although I understand that code review is not free so I don't know if it will be feasible.

I took a brief look at the accessibility lab report and it seems like a lot of the requirements they are talking about are statically checkable, and some of them are machine checkable. For example, requirement 1.3.1 states that "Headings should convey a clear and accurate structural outline of the sections of content of a web page" which does not seem machine checkable (at least, not without some super fancy AI :stuck_out_tongue_closed_eyes:) , but also states that "Headings should not skip hierarchical levels" which is machine checkable. "Most web pages should have only one \<h1\>" is a soft requirement, but when I run `make test` there are a bunch of statistics at the end for code coverage, and having a statistic for "only has one \<h1\>" might be reasonable.

My proposal is that I will create a new test module in this commit which focuses on testing accessibility requirements. As future work, I will review each of the requirements listed in the report in order to identify what is testable and how, and add tests for each of the requirements. The docstring of each test will explicitly state what is being machine-tested, and what requires human validation. If this sounds interesting to you, I can write up something more detailed, presumably in a new GitHub issue.

Let me know what y'all think!

cfm (changes requested):
@saffronsnail, thank you for working on this!  And thanks, @gonzalo-bulnes, for starting review yesterday.

@eaon and I have discussed your proposed change here in the context of our comments on the original #6313.  Broadly we have two concerns:

1. We'd envisioned just exposing a Jinja template block that could be filled in like `{% block title %}Your Title Here{% endblock %}`.  That would require developers to keep the `title` and `h1` manually in sync (most of the time), but it would also allow for intentional deviations where they make sense.  And it would keep things simple.  :-)

2. The reason we have not gone ahead and made this change in the context of other accessibility-related work (#5972) is the [need to consider its UX and even security implications](https://docs.google.com/spreadsheets/d/13lbFPR1Pm4QbcVz_7oVxMmKViqSl268CKA0ixoMyYxw/edit#gid=57993727&range=17:17).  At least the latter may be overblown, since under Tor Browser both page `title`s and URLs are ephemeral in every session; but we'd intended to discuss this explicitly.

I apologize for leaving the original #6313 as a stub that isn't more explicit on these points.  Here's what I propose:

1. I'll follow up in #6313 with an update and ask our UX designer and security engineers to weigh in.

2. If they sign off on this approach, we'd welcome another pull request from you to implement the `title` block!  But of course you have no obligation to keep working on this.  :-)

How does this sound to you?  Thanks again for your time and effort here.

cfm:
Also, re: @saffronsnail in <https://github.com/freedomofpress/securedrop/pull/6725#issuecomment-1386332236>:

> My proposal is that I will create a new test module in this commit which focuses on testing accessibility requirements. As future work, I will review each of the requirements listed in the report in order to identify what is testable and how, and add tests for each of the requirements. The docstring of each test will explicitly state what is being machine-tested, and what requires human validation. If this sounds interesting to you, I can write up something more detailed, presumably in a new GitHub issue.

If you have the time and inclination to work on this, we'd love your input in #6314, where I started to take a look at some of the accessibility-testing tools already out there.  :-)

skyvine:
Hi @cfm, thanks for the info! I think that your approach of having a title block is clearer to readers & writers than having a magic function that converts an h1 statement into a title. I'll start looking at that version of the implementation.

I had overlooked #6314. I have experience with developing test suites, but focused on testing for the presence of security vulnerabilities, not accessibility-focused testing. I'll take a look at the links listed over there, then throw in my 2 cents. =)

cfm:
On Wed, Jan 18, 2023 at 11:12:14AM -0800, Skyler Ferris wrote:
> Hi @cfm, thanks for the info! I think that your approach of having a
> title block is clearer to readers & writers than having a magic
> function that converts an h1 statement into a title. I'll start
> looking at that version of the implementation.
> 
> I had overlooked #6314. I have experience with developing test suites,
> but focused on testing for the presence of security vulnerabilities,
> not accessibility-focused testing. I'll take a look at the links
> listed over there, then throw in my 2 cents. =)

Thanks on both counts, @saffronsnail!  Note that the template-block
approach will be easiest to review as a separate pull request, so feel
free to close this one (or I can).  I'll update #6313 with these
considerations today or tomorrow.

PR #6732, opened 2023-01-20
Try GitHub Actions by migrating updater-gui-tests and rust jobs

legoktm:
## Status

Ready for review 

## Description of Changes

Try GitHub Actions by migrating updater-gui-tests and rust jobs

The syntax is fairly close, so no actual changes were needed to the commands that are run.

I did switch usage of `apt` to `apt-get` since the former isn't supposed to be used in scripts.

I intentionally chose to not migrate the branch filter over to reduce complexity given how fast and light this job is.

## Testing

* [ ] CI passes

## Deployment

## Checklist

n/a

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -247,25 +247,6 @@ jobs:
           command: |
             cd admin; make test
 
-  updater-gui-tests:
-    docker:
-      - image: debian:bullseye
-    steps:
-      - checkout
-      - run:
-          name: Install dependencies
-          command: |
-            apt update && apt-get install -y libqt5designer5 python3-venv
-            cd journalist_gui
-            python3 -m venv .venv/ && source .venv/bin/activate
-            pip install --require-hashes -r dev-requirements.txt
-      - run:
-          name: Run tests
-          command: |
-            cd journalist_gui
-            source .venv/bin/activate
-            QT_QPA_PLATFORM=offscreen python3 test_gui.py -v
-
   static-analysis-and-no-known-cves:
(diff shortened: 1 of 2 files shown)


legoktm:
One weird thing is that GH Actions will now run the job twice, once for the branch push inside the securedrop repo, and then again for the pull request. For quick jobs like this it's not an issue, but it could get annoying for longer ones.

legoktm:
https://wildwolf.name/github-actions-how-to-avoid-running-the-same-workflow-multiple-times/ seems promising.

gonzalo-bulnes:
I too have noticed the double run (without taking action on it for my repos). Additionally, note that if tou want to (re-)trigger workflows manually for whatever reason, they must be configured to trigger on the `workflow_dispatch` event.

legoktm:
> Additionally, note that if tou want to (re-)trigger workflows manually for whatever reason, they must be configured to trigger on the `workflow_dispatch` event.

When I visit https://github.com/freedomofpress/securedrop/actions/runs/3971850268/jobs/6809178862 I see a "Re-run all jobs", did you mean something different?

gonzalo-bulnes:
Nope, if tou see the buttons, that was it. For some reason I had to enable that manually in my workflows... and found that explanation in GitHub's Help pages.

legoktm:
Moving back to draft state, this needs to be updated and reconciled with #6969. (Maybe squashed into that PR?)

legoktm:
I'm redoing these commits in #6969.

PR #6738, opened 2023-02-02
Add title blocks.

### LINKED ISSUE #6313: surface view-level `h1`s (or equivalents) in page `title`s (WCAG 2.4.2)

skyvine:
## Status

Ready for review

## Description of Changes

Per the Web Content Accessibilty Guidelines 2.0, Success Criterion 2.4.2, all pages should have descriptive titles to help users orient themselves within the website, particularly for users who take advantage of screen readers while browsing the web.

This commit adds a new `tab_title` block to the base templates for the source and journalist, as well as the index page of the source (it does not derive from base). It is named this way so that it is clear that this block will NOT add a "title" to the *contents* of the page, eg a heading at the beginning of the content. The other files had descriptive titles added. In most cases, these titles are a copy-paste of a `gettext` call already in the page, but there are a few exceptions.

`journalist_templates/admin_add_user.html` currently does not use a `gettext` call for the heading that is appropriate for the title. `journalist_templates/error.html` also does not, although it is getting the name of an error from a variable, so I assume that translation is handled elsewhere in this case.

`journalist_templates/col.html`, `source_templates/index.html`, and `source_templates/login.html` did not have headings that were appropriate for page titles, so I invented new titles for these. I am not a translator, so these are not in `gettext` calls.

`source_templates/logout.html` and
`source_templates/tor2web-warning.html` had headings which could be used for titles, but the text was either modified or replaced to make sure that users who are relying soley on the tab title, without the ability to conveniently skim the contents to better understand the length and nature of the text, receive communication about the importance and urgency of the contents of these pages from the title itself.

Also update the `test_orgname_is_changed` test so that the assertions do
not have a dependency on the page title.

Fixes #6313.

## Testing

I navigated through the pages in both the source and journalist interface and made sure the titles were visible on the tab. I tested this in plain firefox, I'm not overly concerned about browser compatibility because the client is just receiving different static text due to this change, but I can go back and look at it in tor if needed.

## Deployment

I have a couple of concerns, I'm not sure if this technically qualifies as "deployment" but this seems like the best place to bring them up.

The heading that I used for the page title in `journalist_templates/admin_add_user.html` is not wrapped in a `gettext` call, while most of the others are (other exceptions are categorically different and will be explained shortly). I'm not super familiar with gettext, but IIUC this means that this text will not be translated. This doesn't seem ideal, but I don't think I'm the right person/this is the right issue within which to address that. Technically, the error page is also not in a `gettext` call, but it's referencing a variable value and I'm assuming that error messages are handled elsewhere, possibly by the browser itself since error codes are well known.

The other exceptions are all because the page did not already contain an appropriate heading, so I added a page title that seemed reasonable. These pages are `journalist_templates/col.html`, `source_templates/index.html`, `source_templates/login.html`, and `source_templates/logout.html` (see previous comments for the reason why logout has new text). Additionally, `source_templates/tor2web-warning.html` has the word "WARNING" prepended to the `gettext` call (again, see previous comments). I could dig through the `po` files to see if there is a key that seems reasonable, but I'm skeptical there will be. Assuming we want to use the new titles, is there anything I need to do to put these titles on the translations team's radar?

N/A

N/A

N/A

Discussions about tests for accessibility are taking place in #6314.

Choose one of the following:

N/A

### DIFF
--- a/securedrop/journalist_templates/base.html
+++ b/securedrop/journalist_templates/base.html
@@ -4,7 +4,7 @@
 <head>
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <title>{{ g.organization_name }}</title>
+  <title>{% block tab_title required %}{% endblock %} | {{ g.organization_name }}</title>
 
   <link rel="stylesheet" href="/static/css/journalist.css">
 
--- a/securedrop/journalist_templates/col.html
+++ b/securedrop/journalist_templates/col.html
@@ -1,4 +1,7 @@
 {% extends "base.html" %}
+
+{% block tab_title %}{{ source.journalist_designation }}{% endblock %}
+
 {% block body %}
 <div id="content" class="journalist-view-single">
 
--- a/securedrop/source_templates/base.html
+++ b/securedrop/source_templates/base.html
@@ -6,9 +6,9 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <meta name="robots" content="noindex,nofollow">
   {% if g.organization_name == "SecureDrop" %}
-  <title>{{ g.organization_name }} | {{ gettext('Protecting Journalists and Sources') }}</title>
+  <title>{% block tab_title required %}{% endblock %} | {{ g.organization_name }} | {{ gettext('Protecting Journalists and Sources') }}</title>
   {% else %}
-  <title>{{ g.organization_name }} | {{ gettext('SecureDrop') }}</title>
+  <title>{{ self.tab_title() }} | {{ g.organization_name }} | {{ gettext('SecureDrop') }}</title>
   {% endif %}
 
   <link rel="stylesheet" href="/static/css/source.css">
--- a/securedrop/source_templates/index.html
+++ b/securedrop/source_templates/index.html
@@ -3,9 +3,9 @@
 
 <head>
   {% if g.organization_name == "SecureDrop" %}
-  <title>{{ g.organization_name }} | {{ gettext('Protecting Journalists and Sources') }}</title>
+  <title>{{ gettext('Welcome') }} | {{ g.organization_name }} | {{ gettext('Protecting Journalists and Sources') }}</title>
   {% else %}
-  <title>{{ g.organization_name }} | {{ gettext('SecureDrop') }}</title>
+  <title>{{ gettext('Welcome') }} | {{ g.organization_name }} | {{ gettext('SecureDrop') }}</title>
   {% endif %}
 
   <link rel="stylesheet" href="/static/css/source.css">
--- a/securedrop/source_templates/tor2web-warning.html
+++ b/securedrop/source_templates/tor2web-warning.html
@@ -1,4 +1,7 @@
 {% extends "base.html" %}
+
+{% block tab_title %}{{ gettext('WARNING: Proxy Service Detected') }}{% endblock %}
+
 {% block body %}
 <h2>{{ gettext('Proxy Service Detected') }}</h2>
 <div class="center">
--- a/securedrop/tests/test_i18n.py
+++ b/securedrop/tests/test_i18n.py
@@ -228,7 +228,10 @@ def verify_i18n(app):
             assert 'hreflang="fr-FR"' in locales
 
         c.get("/?l=ar")
-        base = render_template("base.html")
+        # We have to render a template that inherits from "base.html" so that "tab_title" will be
+        # set.  But we're just checking that when a page is rendered in an RTL language the
+        # directionality is correct, so it doesn't matter which template we render.
+        base = render_template("error.html", error={})
         assert 'dir="rtl"' in base
 
 
--- a/securedrop/journalist_templates/account_edit_hotp_secret.html
+++ b/securedrop/journalist_templates/account_edit_hotp_secret.html
@@ -1,4 +1,7 @@
 {% extends "base.html" %}
+
+{% block tab_title %}{{ gettext('Change Secret') }}{% endblock %}
+
 {% block body %}
 <h1>{{ gettext('Change Secret') }}</h1>
 <form method="post">
--- a/securedrop/journalist_templates/account_new_two_factor.html
+++ b/securedrop/journalist_templates/account_new_two_factor.html
@@ -1,4 +1,7 @@
 {% extends "base.html" %}
+
+{% block tab_title %}{{ gettext('Enable FreeOTP') }}{% endblock %}
+
 {% block body %}
 {% if user.is_totp %}
 <h1>{{ gettext('Enable FreeOTP') }}</h1>
--- a/securedrop/journalist_templates/admin.html
+++ b/securedrop/journalist_templates/admin.html
@@ -1,4 +1,7 @@
 {% extends "base.html" %}
+
+{% block tab_title %}{{ gettext('Admin Interface') }}{% endblock %}
+
 {% block body %}
 <h1>{{ gettext('Admin Interface') }}</h1>
 
--- a/securedrop/journalist_templates/admin_add_user.html
+++ b/securedrop/journalist_templates/admin_add_user.html
@@ -1,4 +1,7 @@
 {% extends "base.html" %}
+
+{% block tab_title %}{{ gettext('Add User') }}{% endblock %}
+
 {% block body %}
 <nav class="back" aria-labelledby="back-link">
   <a id="back-link" href="/admin">{{ gettext('Back to admin interface') }}</a>
--- a/securedrop/journalist_templates/admin_edit_hotp_secret.html
+++ b/securedrop/journalist_templates/admin_edit_hotp_secret.html
@@ -1,4 +1,7 @@
 {% extends "base.html" %}
+
+{% block tab_title %}{{ gettext('Change HOTP Secret')}}{% endblock %}
+
 {% block body %}
 <h1>{{ gettext('Change HOTP Secret')}}</h1>
 <p>{{ gettext("Enter a new HOTP secret formatted as a 40-digit hexadecimal string. Spaces will be ignored:") }}</p>
--- a/securedrop/journalist_templates/admin_new_user_two_factor.html
+++ b/securedrop/journalist_templates/admin_new_user_two_factor.html
@@ -1,5 +1,7 @@
 {% extends "base.html" %}
 
+{% block tab_title %}{{ gettext('Enable FreeOTP') }}{% endblock %}
+
 {% block body %}
 {% if user.is_totp %}
 <h1>{{ gettext('Enable FreeOTP') }}</h1>
--- a/securedrop/journalist_templates/config.html
+++ b/securedrop/journalist_templates/config.html
@@ -1,4 +1,7 @@
 {% extends "base.html" %}
+
+{% block tab_title %}{{ gettext('Instance Configuration') }}{% endblock %}
+
 {% block body %}
 <nav class="back" aria-labelledby="back-link">
   <a id="back-link" href="/admin">{{ gettext('Back to admin interface') }}</a>
--- a/securedrop/journalist_templates/edit_account.html
+++ b/securedrop/journalist_templates/edit_account.html
@@ -1,4 +1,13 @@
 {% extends "base.html" %}
+
+{% block tab_title %}
+{% if user %}
+{{ gettext('Edit user "{user}"').format(user=user.username) }}
+{% else %}
+{{ gettext('Edit your account') }}
+{% endif %}
+{% endblock %}
+
 {% block body %}
 
 {% if user %}
--- a/securedrop/journalist_templates/error.html
+++ b/securedrop/journalist_templates/error.html
@@ -1,4 +1,7 @@
 {% extends "base.html" %}
+
+{% block tab_title %}{{ error.name }}{% endblock %}
+
 {% block body %}
 <h1>{{ error.code }}: {{ error.name }}</h1>
 {% if error.description %}
--- a/securedrop/journalist_templates/index.html
+++ b/securedrop/journalist_templates/index.html
@@ -1,4 +1,7 @@
 {% extends "base.html" %}
+
+{% block tab_title %}{{ gettext('All Sources') }}{% endblock %}
+
 {% block body %}
 <div id="content" class="journalist-view-all">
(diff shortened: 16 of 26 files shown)


securedrop/journalist_templates/base.html
```diff
@@ -4,7 +4,7 @@
 <head>
   <meta charset="utf-8">
   <meta name="viewport" content="width=device-width, initial-scale=1">
-  <title>{{ g.organization_name }}</title>
+  <title>{% block tab_title %}{% endblock %} | {{ g.organization_name }}</title>
```
cfm: Is it your intention that this block be [`required`] for all templates?  If not, then we should check for its absence here, to avoid rendered templates like:

```html
<title>| SecureDrop</title>
```

[`required`]: https://jinja.palletsprojects.com/en/3.0.x/templates/#required-blocks

securedrop/source_templates/base.html
```diff
@@ -6,9 +6,9 @@
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <meta name="robots" content="noindex,nofollow">
   {% if g.organization_name == "SecureDrop" %}
-  <title>{{ g.organization_name }} | {{ gettext('Protecting Journalists and Sources') }}</title>
+  <title>{% block tab_title %}{% endblock %} | {{ g.organization_name }} | {{ gettext('Protecting Journalists and Sources') }}</title>
   {% else %}
-  <title>{{ g.organization_name }} | {{ gettext('SecureDrop') }}</title>
+  <title>{{ self.tab_title() }} | {{ g.organization_name }} | {{ gettext('SecureDrop') }}</title>
```
cfm: Same question as <https://github.com/freedomofpress/securedrop/pull/6738#discussion_r1098048795>.

securedrop/source_templates/index.html
```diff
@@ -3,9 +3,9 @@
 
 <head>
   {% if g.organization_name == "SecureDrop" %}
-  <title>{{ g.organization_name }} | {{ gettext('Protecting Journalists and Sources') }}</title>
+  <title>Portal | {{ g.organization_name }} | {{ gettext('Protecting Journalists and Sources') }}</title>
   {% else %}
-  <title>{{ g.organization_name }} | {{ gettext('SecureDrop') }}</title>
+  <title>Portal | {{ g.organization_name }} | {{ gettext('SecureDrop') }}</title>
```
cfm: I recommend we keep this `title` as it is, since we don't use the term "portal" anywhere else in SecureDrop's interfaces or documentation.

securedrop/source_templates/tor2web-warning.html
```diff
@@ -1,4 +1,7 @@
 {% extends "base.html" %}
+
+{% block tab_title %}WARNING: {{ gettext('Proxy Service Detected') }}{% endblock %}
```
cfm: While we do need all user-facing strings to be [wrapped in `gettext()`] for translation, rest assured that you are not responsible for supplying or managing it!  Our release process involves a localization phase.  :-)

```suggestion
{% block tab_title %}{{ gettext('WARNING: Proxy Service Detected') }}{% endblock %}
```

[wrapped in `gettext()`]: https://developers.securedrop.org/en/latest/i18n.html

securedrop/journalist_templates/col.html
```diff
@@ -1,4 +1,7 @@
 {% extends "base.html" %}
+
+{% block tab_title %}Source Submissions{% endblock %}
```
cfm: This is a fine generalized `title`!  We could, however, use `source.journalist_designation` here, which would have the added benefit of making different sources' `col` listings navigable in Tor Browser history.  @zenmonkeykstop, do you have any strong feelings either way?

cfm (changes requested):
Nice; thank you, @saffronsnail!  I've requested some changes in this review, I believe addressing along the way all of the points you've raised here.  Let me know what you think!

securedrop/source_templates/index.html
gonzalo-bulnes: Drive-by context: https://github.com/freedomofpress/securedrop/pull/6725#discussion_r1071647718 (Not dismissing what you said @cfm, only adding previous context FYI.)

securedrop/journalist_templates/base.html
skyvine: My intention was for it to be mandatory, as my understanding of the relevant guideline is that all pages should have some sort of descriptive title. Is this a problem?

securedrop/source_templates/index.html
skyvine: As in the previous comment above, my understanding of the guideline is that we should strive to give descriptive titles to all pages. The documentation discusses this page under the heading "Making your first submission" (https://docs.securedrop.org/en/stable/source.html), but does not specifically give the page a name. Something like "Start First Submission or Log In" would be accurate, although Technique G88 also mentions that tiles should be short (https://www.w3.org/WAI/WCAG21/Techniques/general/G88). Perhaps in could be named "New/Returning Submission"?

securedrop/journalist_templates/col.html
skyvine: SGTM, I'll update the commit if we get sign-off on using it.

securedrop/source_templates/tor2web-warning.html
skyvine: Fixed. I also added gettext calls around other title blocks which were all plain text, let me know if that was incorrect. I uploaded this as a separate commit so that it was easier for you to see the updates, but I can squash everything before you accept the pull request so that the history of develop stays clean.

securedrop/source_templates/index.html
gonzalo-bulnes: A "my 2 cents"-style comment: It seems to me that "New Submission" could be a short, accurate title indeed.

I quite like "New  [...]Submission" idea and I'd argue that whether that will be the first submission you make, or you're returning to make another one... it's going to be a _new_ (if not _first_) submission in either case!

securedrop/journalist_templates/col.html
zenmonkeykstop: No issues with this that I can see.

skyvine: Added a third commit which implements this change

securedrop/source_templates/tor2web-warning.html
cfm: Thanks, @saffronsnail!

securedrop/journalist_templates/col.html
cfm: Looks good!

securedrop/journalist_templates/base.html
cfm: No, I agree!  In that case you can add the [`required`] attribute to the `tab_title` block so that a template that doesn't override it will raise a `TemplateRuntimeError`.

[`required`]: https://jinja.palletsprojects.com/en/3.0.x/templates/#required-blocks

securedrop/source_templates/index.html
cfm: Ah, thanks for pointing me to <https://github.com/freedomofpress/securedrop/pull/6725#discussion_r1071647718>, @gonzalo-bulnes.  I'd missed that discussion.

I appreciate the suggestion of "New Submission".  Here's the remaining case to consider:  A source returns to SecureDrop *not* to submit anything new but just to see if a journalist has replied to them.  I'm sure they'll be able to figure out how to proceed from the "First Submission" and "Return Visit" sections, but it's a flow not covered by a title like "New Submission".  Now, your original "Portal" is a neutral-enough umbrella term to include that flow, just at the cost of introducing a new term and concept we don't use elsewhere.  :-)

So I'm still inclined to keep this title as it is.  It's a special case, in `index.html` rather than via `base.html`, so it doesn't conflict with the `tab_title` block being required.  By definition, then, `index.html` will be the only page with this title, so we'll comply with the letter of WCAG 2.4.2.  And I believe we'll still comply with its spirit, since this title uniquely indicates the SecureDrop instance the source has reached.

However, if you feel strongly about giving a page-specific title here, @saffronsnail, what do you think of "Welcome"?  It's another neutral, non-jargon term I can think of (other than ["landing page"]!) that uniquely identifies what this page *is* rather than specifying what sources can *do* on it.

["landing page"]: https://docs.securedrop.org/en/stable/deployment/landing_page.html

skyvine: I would prefer a page-specific title. I like "Welcome" because it does indicate the purpose of the page as the introduction to the site, and it's less technical than "landing page". Updated in the most recent commit.

securedrop/journalist_templates/base.html
skyvine: Oh yes, I added it in the new commit. Thanks for the tip!

securedrop/source_templates/base.html
skyvine: Added here as well

securedrop/journalist_templates/base.html
cfm: Thanks!

securedrop/source_templates/index.html
cfm: Yes, I meant that ["landing page"] is a reserved term we already use.  :-)  "Welcome" works well here.  Thanks!

["landing page"]: https://docs.securedrop.org/en/stable/glossary.html#landing-page

cfm (changes requested):
Thanks for your changes, @saffronsnail.  Per <https://app.circleci.com/pipelines/github/freedomofpress/securedrop/548/workflows/6f414f84-bbc5-4b0b-bd9f-4fa4cb4bf654/jobs/67448/tests>, it looks like 8ffd659 breaks `test_i18n()`, because the now-required `tab_title` will be unset in:

https://github.com/freedomofpress/securedrop/blob/c2f11efb11b59041137ca73fd19b3af96da6c109/securedrop/tests/test_i18n.py#L231

On Monday I'll think about how we can fix this in the existing `test_i18n.py` test suite—unless you beat me to it.  :-)

Finally, as a heads-up:  I think this branch is a good candidate for [squashing] before merge.  You can do that before I review again, or I'm happy to do it for you; just let me know what you prefer.

[squashing]: https://developers.securedrop.org/en/latest/contributor_guidelines.html?highlight=squash#git-history

skyvine:
Commits are squashed! I also added an additional change that fixes the test, not sure if the mechanism is acceptable but it seemed like the simplest thing to do. Instead of rendering `base.html` it renders `error.html`, and passes in `error={}` so that jinja doesn't throw an undefined error. I think this is fine because we're just checking that when a page is rendered in an rtl language the directionality is correct, so it shouldn't matter *which* template we render.

If this is a problem, my first thought was to inject a `tab_title` block into the page somehow. If I could get the contents of the template this would be pretty straightforward by using `render_template_string` and just tacking a block onto the end of it, but when I looked at the flask/jinja code, it's all going straight to rendering, I don't see a convenient helper for getting the contents without simultaneously processing it. I could grab the contents directly, but if I go through flask infrastructure then I end up re-implementing their backend and keeping it in sync sounds annoying. I could grab the contents manually (once I figure out where the files are actually located on the test machine, which seems pretty doable), which is probably what I will do if we need to render exactly `base.html` for some reason.

skyvine:
OOC, have you had a chance to look at #6745? It's a WIP but I want to make sure we're in sync about the direction of the remaining work to make sure I'm spending my time efficiently. =)

securedrop/tests/test_i18n.py
```diff
@@ -228,7 +228,7 @@ def verify_i18n(app):
             assert 'hreflang="fr-FR"' in locales
 
         c.get("/?l=ar")
-        base = render_template("base.html")
+        base = render_template("error.html", error={})
```
cfm, line 234: This change is worth documenting, I think!

```suggestion
        # We have to render a template that inherits from "base.html" so that "tab_title" will be
        # set.  But we're just checking that when a page is rendered in an RTL language the
        # directionality is correct, so it doesn't matter which template we render.
        base = render_template("error.html", error={})
```

cfm (changes requested):
@saffronsnail in <https://github.com/freedomofpress/securedrop/pull/6738#issuecomment-1435793099>:

> Commits are squashed! I also added an additional change that fixes the test, not sure if the mechanism is acceptable but it seemed like the simplest thing to do. Instead of rendering `base.html` it renders `error.html`, and passes in `error={}` so that jinja doesn't throw an undefined error. I think this is fine because we're just checking that when a page is rendered in an rtl language the directionality is correct, so it shouldn't matter _which_ template we render.

This is great!  I was going to look into something similar before recommending it.  Using the `error.html` template seems fine to me, since (as you say) the content of the template is irrelevant for this test.

Two final requests—

1. I've suggested documenting this change in `test_i18n.py`.
2. Can you edit your commit (and pull request) to remove the now-obsolete references to missing `gettext` calls?

—and then I'll be ready to merge this right away this week.  :-)

skyvine:
Docstring added to commit, message updated with accurate gettext information!

cfm (approved):
Looks great, @saffronsnail.  Thanks for all your patience with this back-and-forth, and congratulations on your first contribution to SecureDrop!

skyvine:
Thank you for all the help and feedback! I am happy and proud to help improve this important project.

Issue #6199, opened 2021-12-16
Remove securedrop-keyring package from securedrop repo.


zenmonkeykstop:
## Description

The same signing key is used for Debian and Ubuntu packages, and the logic for building it is duplicated (with minor differences in layout) in this repo and in the https://github.com/freedomofpress/securedrop-debian-packaging repo.

To reduce build complexity (and because this package is rarely updated), it could be removed from the securedrop repo, adding `focal` a a target dist in https://github.com/freedomofpress/securedrop-debian-packaging instead.

## User Research Evidence
none

## User Stories
As a developer, I want the build to only include necessary artifacts.

legoktm:
The workstation and server versions of this package are identical except for minor metadata things:
```
$ debdiff securedrop-keyring_0.1.5+2.3.2+focal_amd64.deb securedrop-keyring_0.1.5+buster_all.deb
[The following lists of changes regard files as different if they have
different names, permissions or owners.]

Files in second .deb but not in first
-------------------------------------
-rw-r--r--  root/root   /usr/share/doc/securedrop-keyring/changelog.gz
-rw-r--r--  root/root   /usr/share/doc/securedrop-keyring/copyright
-rw-r--r--  root/root   DEBIAN/conffiles
-rw-r--r--  root/root   DEBIAN/md5sums
-rwxr-xr-x  root/root   DEBIAN/postinst

Control files: lines which differ (wdiff format)
------------------------------------------------
Architecture: [-amd64-] {+all+}
Homepage: [-https://securedrop.org-] {+https://github.com/freedomofpress/securedrop-debian-packaging+}
{+Installed-Size: 23+}
[-Source: securedrop-]
Version: [-0.1.5+2.3.2+focal-] {+0.1.5+buster+}
```

It's a super duper trivial package, it just installs one file, so we can just build one deb and copy it across all 3 buster/bullseye/focal components.

legoktm:
The goal here to de-duplicate the nearly-but-not-quite identical securedrop-keyring packages across server and SDW is good, but it seems more likely that SDW will move to bookworm first and start shipping the signing key in sources.list files so we'll still need the keyring package only for the server (since a move to jammy+ seems farther out)...so it should just stay here for the time being.

PR #6740, opened 2023-02-03
Update upgrade scenario to require 7 debs, not 8

zenmonkeykstop:
## Status

Ready for review 
## Description of Changes

The `securedrop-grsec` package was recently moved to the kernel build, so we now build 7 debs from this repo, not 8. This PR is a one-char update to the upgrade scenario `converge.yml` to reflect that.

## Testing
On this branch, on a system set up to support libvirt staging/prod VMs:

- build packages with: `make build-debs`
- set up a venv with: `make venv && source .venv/bin/activate`
- [ ] confirm that `make upgrade-start` completes successfully
- [ ] (optional) run through the upgrade scenario (prod VMs required) and confirm that you get and updated working system

### DIFF
--- a/molecule/upgrade/converge.yml
+++ b/molecule/upgrade/converge.yml
@@ -16,7 +16,7 @@
     - name: Ensure debs were found
       assert:
         that:
-          - "_find_debs_result.files|length >= 8"
+          - "_find_debs_result.files|length >= 7"
         msg: "No local debs found, run 'make build-debs-focal'"
 
 - name: Configure apt-server


legoktm:
Note that we do want to remove more packages in the future, e.g. https://github.com/freedomofpress/securedrop/issues/6199. 

So...is this check actually useful? If debs are missing, won't it fail anyways when it tries to install those packages? Or maybe it should be just `>=1` if we want a fast fail when debs fail to build entirely?

zenmonkeykstop:
> So...is this check actually useful? If debs are missing, won't it fail anyways when it tries to install those packages? Or maybe it should be just `>=1` if we want a fast fail when debs fail to build entirely?

Coz it's testing by running `unattended-upgrades`, if debs are missing it just won't install them and the system will be left with a mixed bunch of packages. So checking for the expected number makes sense.

legoktm (approved):
Ack, that makes sense. I think it would be better to put this check in `make build-debs` itself, if building any of the debs fails we shouldn't proceed.

But I'll think about how to best do that in my refactor, let's land this now to unbreak it.

zenmonkeykstop:
IMO it still makes sense in the upgrade scenario rather than (maybe as well as) in `make build-debs`, as there might be cases where you're not actually building the packages locally.

PR #6744, opened 2023-02-07
Upgrade to mypy 1.0.0

legoktm:
## Status

Ready for review

## Description of Changes

Upgrade to mypy 1.0.0

The one issue it flagged is that in journalist_app.create_app(), the `and handler` check is unnecessary because `handler` will always be truthy based on the known types of `app.error_handler_spec`.

Once that's removed we can refactor the handler lookup under path conditional since it's not used otherwise.

## Testing

* [ ] `make typelint` passes
* [ ] CI passes

## Deployment

## Checklist


zenmonkeykstop (approved):
CI is passing, `make typelint` is passing, lgtm. Happy 1.0 to mypy!

PR #6739, opened 2023-02-03
Update to the CircleCI config

echoesactiii:
## Status

Note: this repository will not have CircleCI Slack notifications until either this is merged, or infra-side changes are reverted. in the case the latter is needed please ping me on slack directly!

Part of https://github.com/freedomofpress/infrastructure/issues/3225

## Description of Changes

Changes proposed in this pull request: this updates the CircleCI config to the latest version, enables the Slack orb, adds a step to call Slack on a failure, and then maps that to each job in the workflow.

this repository has been using a legacy CircleCI integration that is no longer supported, and this brings it up to date with the current guidance for integrationg CircleCI and Slack.

## Testing

easiest thing to do would be to do something that would cause a CI failure and make sure the notification pops into Slack. that would have to be done in _this_ branch (or a child) unless this is otherwise merged first.

## Deployment

deploys on merge, no effect outside of circleCI

Choose one of the following:

- [x] These changes do not require documentation

PR #6745, opened 2023-02-08
Add accessibility sniffs to pageslayout tests

### LINKED ISSUE #245: suggest tools for auditing accessibility impact of HTML/SASS changes
## Describe the change

Short of CI auditing to prevent regressions of the accessibility improvements in progress in freedomofpress/securedrop#5972, we can at least suggest tools contributors can use to audit the accessibility impact of HTML/SASS changes they submit.

## How will this impact users?

Contributors will be informed of a baseline standard for evaluating the accessibility of new front-end changes and equipped with tools to help them do so.

## User Research Evidence

Prompted by freedomofpress/securedrop#5972.

## Additional context

This could also be added a checklist step in the `securedrop` repository's PR template.

### LINKED ISSUE #6314: review options for accessibility linting (versus documentation)

skyvine:
## Status

Ready for review

## Description of Changes

This commit follows the structure of the current pageslayout tests by collecting info in the `save_screenshot_and_html` function, with the intent that it will later be consumed in the appropriate context. Presumably, a new make target similar to `validate-test-html` will be added, or this target will be updated.

- Fixes #6314.
- Closes freedomofpress/securedrop-docs#245.

### Notes for Reviewers

As previously discussed with @cfm, this commit adds code in Dockerfiles to install software from from `npm` ([context](https://github.com/freedomofpress/securedrop/issues/6314#issuecomment-1409572032)). This toolchain is new to the project.

## Testing

`make test TESTFILES=tests/functional/pageslayout` will exercise all of the changes to the test suite. A summary of the results can be seen with `make accessibility-summary`, which will process raw output into a summary file and open it in `less`. Full output can be found at `tests/functional/pageslayout/accessibility-info/`. The directory structure is similar to that of `tests/functional/pageslayout/html`, but after the locale there are 2 separate directories, `errors` and `reviews`, with the corresponding output saved to files using the `test_name` parameter.

The summary file reports when the sniff output differs by locale (this is only approximate but I believe it to be reasonably accurate, see the implementation of `summarize_accessibility_results` for details) . Currently, there are no differences by locale. I exercised this code path by manually editing the text files created in the `make test` invocation to simulate differences. Just delete any line starting with MessageType in any file to see this in action.

## Deployment

N/A

N/A

N/A

N/A

Test suite question is N/A, this is a change to the test suite itself.

Choose one of the following:

Based on the link to freedomofpress/securedrop-docs#245 added by @cfm, it looks like there does need to be a doc update here?

I previously listed this as N/A because the dependency is not "production" in the sense that it does not need to be deployed to users, is that understanding accurate?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -169,6 +169,13 @@ jobs:
             DOCKER_BUILD_ARGUMENTS="--cache-from securedrop-test-focal-py3:${fromtag:-latest}" \
               make validate-test-html || true
 
+      - run:
+          name: Validate accessibility (informational)
+          command: |
+            fromtag=$(docker images |grep securedrop-test-focal-py3 |head -n1 |awk '{print $2}')
+            DOCKER_BUILD_ARGUMENTS="--cache-from securedrop-test-focal-py3:${fromtag:-latest}" \
+              make accessibility-summary || true
+
       - store_test_results:
           path: ~/project/test-results
 
--- a/Makefile
+++ b/Makefile
@@ -282,6 +282,12 @@ validate-test-html:
 	@$(DEVSHELL) html5validator tests/functional/pageslayout/html
 	@echo
 
+.PHONY: accessibility-summary
+accessibility-summary:
+	@echo "███ Processing accessibility results..."
+	@$(DEVSHELL) $(SDBIN)/summarize-accessibility-info
+	cat securedrop/tests/functional/pageslayout/accessibility-info/summary.txt
+
 .PHONY: docker-vnc
 docker-vnc:  ## Open a VNC connection to a running Docker instance.
 	@echo "███ Opening VNC connection to dev container..."
--- a/devops/docker/CIDockerfile
+++ b/devops/docker/CIDockerfile
@@ -7,8 +7,9 @@ ENV DOCKER_SHA256_x86_64 340e0b5a009ba70e1b644136b94d13824db0aeb52e09071410f35a9
 
 RUN apt-get update && \
     apt-get install -y flake8 make virtualenv ccontrol libpython2.7-dev \
-            libffi-dev libssl-dev libyaml-dev python-pip curl git &&\
-    apt-get clean
+            libffi-dev libssl-dev libyaml-dev python-pip curl git npm &&\
+    apt-get clean && \
+    npm --global install html_codesniffer@2.5.1
 
 RUN curl -L -o /tmp/docker-${DOCKER_VER}.tgz https://get.docker.com/builds/Linux/x86_64/docker-${DOCKER_VER}.tgz; \
 	echo "${DOCKER_SHA256_x86_64} /tmp/docker-${DOCKER_VER}.tgz" | sha256sum -c -; \
--- a/securedrop/dockerfiles/focal/python3/Dockerfile
+++ b/securedrop/dockerfiles/focal/python3/Dockerfile
@@ -7,7 +7,8 @@ RUN apt-get update  && apt-get install -y \
     # cached along with everything else too.
     default-jdk \
     libasound2 libdbus-glib-1-2 libgtk2.0-0 libfontconfig1 libxrender1 \
-    libcairo-gobject2 libgtk-3-0 libstartup-notification0
+    libcairo-gobject2 libgtk-3-0 libstartup-notification0 npm && \
+    npm install --global html_codesniffer@2.5.1
 
 # Current versions of the test browser software. Tor Browser is based
 # on a specific version of Firefox, noted in Help > About Tor Browser.
--- a/securedrop/tests/functional/pageslayout/accessibility.py
+++ b/securedrop/tests/functional/pageslayout/accessibility.py
@@ -1,0 +1,191 @@
+import os
+from dataclasses import dataclass
+from enum import Enum
+from pathlib import Path
+from typing import Dict, List, Union
+
+from selenium.webdriver.firefox.webdriver import WebDriver
+
+_ACCESSIBILITY_DIR = (Path(__file__).parent / "accessibility-info").absolute()
+
+_HTMLCS_RUNNER_CODE = """
+var all_messages = [];
+console.log = msg => {
+    all_messages.push(msg);
+};
+
+HTMLCS_RUNNER.run('WCAG2AAA');
+return all_messages;
+"""
+
+
+class MessageType(Enum):
+    ERROR = 1
+    WARNING = 2
+    NOTICE = 3
+
+
+@dataclass
+class Message:
+    """Contains all of the information in a message emitted by HTML CodeSniffer."""
+
+    principle_id: str
+    message: str
+    message_type: MessageType
+    responsible_html: str
+    selector: str
+    element_type: str
+
+    @staticmethod
+    def from_output(output: str) -> "Message":
+        """Parses the output of htmlcs and returns an instance containing all data.
+
+        No processing is performed for flexibility.
+
+        Example message, post-split (note: contents of index 4 contains no newlines, but I had to
+        split it to keep the linter happy):
+
+        0: [HTMLCS] Error
+        1: WCAG2AAA.Principle1.Guideline1_3.1_3_1_AAA.G141
+        2: h2
+        3: #security-level-heading
+        4: The heading structure is not logically nested. This h2 element appears to be the
+           primary document heading, so should be an h1 element.
+        5: <h2 id="security-level-heading" hidden="">...</h2>
+        """
+
+        fields = output.split("|")
+
+        if "Error" in fields[0]:
+            message_type = MessageType.ERROR
+        elif "Warning" in fields[0]:
+            message_type = MessageType.WARNING
+        elif "Notice" in fields[0]:
+            message_type = MessageType.NOTICE
+
+        return Message(
+            message_type=message_type,
+            principle_id=fields[1],
+            element_type=fields[2],
+            selector=fields[3],
+            message=fields[4],
+            responsible_html=fields[5],
+        )
+
+    def __format__(self, _spec: str) -> str:
+        newline = "\n"
+        return f"""
+{self.message_type}: {self.principle_id}
+    {self.message}
+
+    html:
+        {self.responsible_html.replace(newline, f"{newline}        ")}
+        """
+
+
+def sniff_accessibility_issues(driver: WebDriver, locale: str, test_name: str) -> None:
+    """Runs accessibility sniffs on the driver's current page.
+
+    This function is responsible for injecting HTML CodeSniffer into the current page and writing
+    the results to a file. This way, test functions can focus on the setup required to navigate to
+    a particular URL (for example, logging in to get to the messages page).
+    """
+
+    # 1. Retrieve/compute required data
+    with open(f"/usr/local/lib/node_modules/html_codesniffer/build/HTMLCS.js") as htmlcs:
+        html_codesniffer = htmlcs.read()
+
+    errors_dir = _ACCESSIBILITY_DIR / locale / "errors"
+    errors_dir.mkdir(parents=True, exist_ok=True)
+
+    reviews_dir = _ACCESSIBILITY_DIR / locale / "reviews"
+    reviews_dir.mkdir(parents=True, exist_ok=True)
+
+    # 2. Do the thing
+    raw_messages = driver.execute_script(html_codesniffer + _HTMLCS_RUNNER_CODE)
+
+    # 3. Organize the data
+    messages: Dict[str, List[Message]] = {
+        "machine-verified": [],
+        "human-reviewed": [],
+    }
+
+    for message in map(Message.from_output, raw_messages[:-1]):  # last message is effectievly EOF
+        if message.message_type == MessageType.ERROR:
+            messages["machine-verified"].append(message)
+        else:
+            messages["human-reviewed"].append(message)
+
+    # 4. Report the data
+    # Note: it is useful to create empty files when there are no results to simplify the logic for
+    #       summarizing the results, implemented in `summarize_accessibility_results`.
+    with open(errors_dir / f"{test_name}.txt", "w") as error_file:
+        for message in messages["machine-verified"]:
+            error_file.write(f"{message}")
+
+    with open(reviews_dir / f"{test_name}.txt", "w") as review_file:
+        for message in messages["human-reviewed"]:
+            review_file.write(f"{message}")
+
+
+def summarize_accessibility_results() -> None:
+    """Creates a file containing summary information about the result of accessiblity sniffing
+
+    Note: This does not automatically run as part of the test suite, use
+          `make accessibility-summary` instead.
+    """
+
+    try:
+        summary: Dict[str, Dict[str, Dict[str, Union[int, bool]]]] = {}
+
+        # since `sniff_accessibility_issues` creates empty files, all locale/type combinations will
+        # contain the same set of files; getting filenames from en_US/reviews instead of, say,
+        # ar/errors is arbitrary and sufficient
+        for out_filename in os.listdir(_ACCESSIBILITY_DIR / "en_US" / "reviews"):
+            summary[out_filename] = {
+                "reviews": {"count": 0, "locale_differs": False},
+                "errors": {"count": 0, "locale_differs": False},
+            }
+
+            # collect all of the relevant data
+            for message_type in ["reviews", "errors"]:
+                outputs: Dict[str, Dict[str, List[str]]] = {}
+
+                for locale in ["en_US", "ar"]:
+                    outputs[locale] = {}
+                    with open(
+                        _ACCESSIBILITY_DIR / locale / message_type / out_filename
+                    ) as out_file:
+                        # Only look at lines specifying the message (including the exact WCAG error
+                        # code); this is exactly correct for the count, and approximately correct
+                        # for comparing locales. If the order of the errors differs, or if there
+                        # are a different number of any kind of error, this approximation will catch
+                        # it.
+                        outputs[locale][message_type] = [
+                            line for line in out_file.readlines() if "MessageType." in line
+                        ]
+
+                summary[out_filename][message_type]["count"] = len(outputs["en_US"][message_type])
+                summary[out_filename][message_type]["locale_differs"] = (
+                    outputs["en_US"][message_type] != outputs["ar"][message_type]
+                )
+
+        # save the data to a convenient file
+        with open(_ACCESSIBILITY_DIR / "summary.txt", "w") as summary_file:
+            for name in sorted(summary.keys()):
+                summary_file.write(name + ":\n")
+                for message_type in ["errors", "reviews"]:
+                    summary_file.write(
+                        f"\t{message_type}: {summary[name][message_type]['count']}\n"
+                    )
+                    if summary[name][message_type]["locale_differs"]:
+                        summary_file.write(f"\t        NOTE: {message_type} differ by locale\n")
+
+                summary_file.write("\n")
+
+    # this should only happen if the pageslayout tests have not created the raw output files
+    except FileNotFoundError:
+        print(
+            f"ERROR: Run `make test TESTFILES={os.path.dirname(_ACCESSIBILITY_DIR)}` before "
+            "running `make accessibility-summary`"
+        )
--- a/securedrop/bin/summarize-accessibility-info
+++ b/securedrop/bin/summarize-accessibility-info
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+set -euo pipefail
+
+export PATH="/opt/venvs/securedrop-app-code/bin:$PATH"
+
+cd ..
+export REPOROOT="${REPOROOT:-$PWD}"
+git config --global --add safe.directory "$REPOROOT"
+cd "${REPOROOT}/securedrop"
+
+PYTHONPATH=".:${PYTHONPATH:-}"
+export PYTHONPATH
+
+python3 -c 'from tests.functional.pageslayout.accessibility import summarize_accessibility_results; summarize_accessibility_results()'
--- a/securedrop/tests/functional/pageslayout/.gitignore
+++ b/securedrop/tests/functional/pageslayout/.gitignore
@@ -1,2 +1,3 @@
+accessibility-info
 html
 screenshots
--- a/securedrop/tests/functional/pageslayout/test_journalist.py
+++ b/securedrop/tests/functional/pageslayout/test_journalist.py
@@ -17,7 +17,7 @@
 #
 import pytest
 from tests.functional.app_navigators.journalist_app_nav import JournalistAppNavigator
-from tests.functional.pageslayout.utils import list_locales, save_screenshot_and_html
+from tests.functional.pageslayout.utils import list_locales, save_static_data
 
 
 @pytest.mark.parametrize("locale", list_locales())
@@ -33,7 +33,7 @@ class TestJournalistLayout:
             accept_languages=locale_with_commas,
         )
         journ_app_nav.driver.get(f"{sd_servers.journalist_app_base_url}/login")
-        save_screenshot_and_html(journ_app_nav.driver, locale, "journalist-login")
+        save_static_data(journ_app_nav.driver, locale, "journalist-login")
 
         # And they log into the app and are an admin
         assert sd_servers.journalist_is_admin
@@ -42,19 +42,17 @@ class TestJournalistLayout:
             password=sd_servers.journalist_password,
             otp_secret=sd_servers.journalist_otp_secret,
         )
-        save_screenshot_and_html(journ_app_nav.driver, locale, "journalist-index_no_documents")
+        save_static_data(journ_app_nav.driver, locale, "journalist-index_no_documents")
         # The documentation uses an identical screenshot with a different name:
         # https://github.com/freedomofpress/securedrop-docs/blob/main/docs/images/manual
         # /screenshots/journalist-admin_index_no_documents.png
         # So we take the same screenshot again here
         # TODO(AD): Update the documentation to use a single screenshot
-        save_screenshot_and_html(
-            journ_app_nav.driver, locale, "journalist-admin_index_no_documents"
-        )
+        save_static_data(journ_app_nav.driver, locale, "journalist-admin_index_no_documents")
 
         # Take a screenshot of the edit account page
         journ_app_nav.journalist_visits_edit_account()
-        save_screenshot_and_html(journ_app_nav.driver, locale, "journalist-edit_account_user")
+        save_static_data(journ_app_nav.driver, locale, "journalist-edit_account_user")
 
     def test_index_entered_text(self, locale, sd_servers, firefox_web_driver):
         # Given an SD server
@@ -73,7 +71,7 @@ class TestJournalistLayout:
             otp_secret="2HGGVF5VPHWMCAYQ",
             should_submit_login_form=False,
         )
-        save_screenshot_and_html(journ_app_nav.driver, locale, "journalist-index_with_text")
+        save_static_data(journ_app_nav.driver, locale, "journalist-index_with_text")
 
     def test_index_with_submission_and_select_documents(
         self, locale, sd_servers_with_submitted_file, firefox_web_driver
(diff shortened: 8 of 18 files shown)


cfm:
@saffronsnail, thanks for checking in about this.  Since your changes here are still WIP, this is not a full review, but I'm happy to discuss your questions so far.

> I am trying to match the structure of the current pageslayout tests. Towards this goal, the sniffing function creates a set of output files at `tests/functional/pageslayout/accessibility-info`. For the pageslayout tests, data saved in similar directories is just static captures of the page, and `make validate-test-html` is used to view the output of the validator. In this commit, the data saved is the output of the tests itself. Is this the best place to save it? I could also add it to a new directory under `test-results`, which appears in the root of the repository after running `make test`.
> 
> On a similar note, the output was written under my original assumption that this would be directly displayed to the user. However, since this is being saved to a file I wonder if it would be more useful to save it in a structured format; I see that there are already some XML files in the `test-results` directory. Would it be useful to have a separate make command to display a summary of the errors/reviews generated for each page?

The page-layout tests report via pytest, which we've configured to save to `test-results/junit.xml` for [CircleCI to parse].  It doesn't seem like we'll be able to preserve that association for these tests, so I agree that `make validate-test-html` is the better precedent: save the results along with the page-layout captures; display them at the end; and (eventually) fail on errors.

[CircleCI to parse]: https://circleci.com/docs/collect-test-data/

 
> There are also a few organizational/style things. Much of the new code is in the new `tests/functional/pageslayout/accessibility.py` file. I considered adding it to the `utils.py` file in the same directory, but held off because that would triple the size of the file and there is only one point where the code needs to interoperate (more on that later). Is there any preference for this to be in a separate file or merged with utils?

I like your approach here, since the new `tests.functional.pageslayout.accessibility` module corresponds directly to the new `html_codesniffer` dependency.


> I added the call to the sniffing function inside of `save_screenshot_and_html` because it needs the same data and avoids adding a bunch of one-line changes to virtually every function currently in the pageslayout directory. I feel like this is a little misleading to the reader, because there's no indication that `save_screenshot_and_html` contains accessibility-related functionality. Thoughts on renaming this to `save_accessibility_info_screenshot_and_html`? I'm wary of making the identifier too long, so a more generic name like `save_static_data` might make sense. I could leave it as-is. Or add the calls separately, as I originally intended.

`save_static_data()` sounds like a good refactoring, in a separate commit and documented with a docstring.  :-)

skyvine:
Thanks for the feedback! I think the new version of the branch should be good to go, I've also updated the first comment to reflect the current status. One thing that I noticed when updating the function name is several comments along the lines of:

```
        # The documentation uses an identical screenshot with a different name:
        # https://github.com/freedomofpress/securedrop-docs/blob/main/docs/images/manual
        # /screenshots/journalist-code-fail_login_many.png
        # So we take the same screenshot again here
        # TODO(AD): Update the documentation to use a single screenshot
```

The function name doesn't explicitly say that it's taking a screenshot any more. I don't think this is a big problem especially with the docstring update, but just wanted to call attention to it in case it is more concerning than I assume.

Also, the summary includes the full path to the accessibility-info directory at the top as a helpful notice. I'm not sure if this is too helpful - this reveals the developer's username if the summary file becomes public. This file is only generated as a result of `make accessibility-summary`, which I would not expect to be used by any CI process. Also, I'm not sure if it actually reveals the username, the account I'm using for preparing this pull request is named "user" and it might be that the development container also uses the name "user" regardless of my account's username, and it's just a coincidence that it's the same. Let me know if this is a concern and I can look more into it, or just modify the line to print the path relative to the root of the repository rather than the absolute path.

Finally, the circle-ci doc mentions that, "[w]hen you save test data using the `store_test_results` step, CircleCI collects data from XML files and uses it to provide insights into your job." IIUC, this means that if I updated the code to write out XML files then it could actually be integrated with circle-ci. Would there be any interest in a change that implements this on your end? If so, I still think that this should be merge-able as-is and XML output can be handled in a separate PR.

skyvine:
Here's a copy of the summary output on my machine in case it is of interest:

<details>
<summary>Accessibility Summary</summary>

journalist-account_edit_hotp_secret.txt:
	errors: 1
	reviews: 60

journalist-account_new_two_factor_hotp.txt:
	errors: 1
	reviews: 61

journalist-account_new_two_factor_totp.txt:
	errors: 2
	reviews: 60

journalist-admin.txt:
	errors: 2
	reviews: 70

journalist-admin_add_user_hotp.txt:
	errors: 1
	reviews: 76

journalist-admin_add_user_totp.txt:
	errors: 1
	reviews: 76

journalist-admin_changes_logo_image.txt:
	errors: 4
	reviews: 99

journalist-admin_edit_hotp_secret.txt:
	errors: 4
	reviews: 101

journalist-admin_edit_totp_secret.txt:
	errors: 2
	reviews: 60

journalist-admin_index_no_documents.txt:
	errors: 0
	reviews: 49

journalist-admin_interface_index.txt:
	errors: 2
	reviews: 67

journalist-admin_new_user_two_factor_hotp.txt:
	errors: 1
	reviews: 60

journalist-admin_new_user_two_factor_totp.txt:
	errors: 2
	reviews: 60

journalist-admin_ossec_alert_button.txt:
	errors: 4
	reviews: 99

journalist-admin_system_config_page.txt:
	errors: 4
	reviews: 98

journalist-clicks_on_source_and_selects_documents.txt:
	errors: 8
	reviews: 107

journalist-code-fail_login.txt:
	errors: 1
	reviews: 64

journalist-code-fail_login_many.txt:
	errors: 1
	reviews: 64

journalist-code-fail_to_visit_admin.txt:
	errors: 1
	reviews: 63

journalist-col.txt:
	errors: 8
	reviews: 107

journalist-col_has_no_key.txt:
	errors: 7
	reviews: 94

journalist-col_javascript.txt:
	errors: 8
	reviews: 107

journalist-col_no_document.txt:
	errors: 4
	reviews: 84

journalist-composes_reply.txt:
	errors: 8
	reviews: 107

journalist-delete_all.txt:
	errors: 4
	reviews: 84

journalist-delete_all_confirmation.txt:
	errors: 13
	reviews: 108

journalist-delete_none.txt:
	errors: 8
	reviews: 109

journalist-delete_one.txt:
	errors: 4
	reviews: 84

journalist-delete_one_confirmation.txt:
	errors: 13
	reviews: 108

journalist-edit_account_admin.txt:
	errors: 4
	reviews: 101

journalist-edit_account_user.txt:
	errors: 4
	reviews: 97

journalist-index.txt:
	errors: 9
	reviews: 88

journalist-index_javascript.txt:
	errors: 9
	reviews: 88

journalist-index_no_documents.txt:
	errors: 0
	reviews: 49

journalist-index_with_text.txt:
	errors: 1
	reviews: 63

journalist-login.txt:
	errors: 1
	reviews: 63

source-checks_for_reply.txt:
	errors: 7
	reviews: 94

source-deletes_reply.txt:
	errors: 8
	reviews: 69

source-enter-codename-in-login.txt:
	errors: 3
	reviews: 57

source-generate.txt:
	errors: 2
	reviews: 58

source-index.txt:
	errors: 3
	reviews: 65

source-login.txt:
	errors: 3
	reviews: 57

source-logout_page.txt:
	errors: 2
	reviews: 49

source-lookup-shows-codename.txt:
	errors: 7
	reviews: 70

source-lookup.txt:
	errors: 8
	reviews: 71

source-next_submission_flashed_message.txt:
	errors: 8
	reviews: 71

source-notfound.txt:
	errors: 2
	reviews: 47

source-session_timeout.txt:
	errors: 4
	reviews: 67

source-submission_entered_text.txt:
	errors: 7
	reviews: 70

source-tor2web_warning.txt:
	errors: 1
	reviews: 50

source-use_tor_browser.txt:
	errors: 1
	reviews: 47

source-why_journalist_key.txt:
	errors: 4
	reviews: 49

</details>

Makefile
```diff
@@ -282,6 +282,12 @@ validate-test-html:
 	@$(DEVSHELL) html5validator tests/functional/pageslayout/html
 	@echo
 
+.PHONY: accessibility-summary
+accessibility-summary:
+	@echo "███ Processing accessibility results..."
+	@$(DEVSHELL) $(SDBIN)/summarize-accessibility-info
+	less securedrop/tests/functional/pageslayout/accessibility-info/summary.txt
```
cfm: I suggest leaving this up to the user via `make accessibility-summary | less`, since otherwise this path is opaque:

```suggestion
	cat securedrop/tests/functional/pageslayout/accessibility-info/summary.txt
```

cfm (changes requested):
Very nice, @saffronsnail!  I've tested both stages of this linting locally:

1. `make test TESTFILES=securedrop/tests/functional/pageslayout`
2. `make accessibility-summary`

I do want to follow up on your original idea:

> Presumably, a new make target similar to `validate-test-html` will be added, or this target will be updated.

How do you envision this linting being used for reporting in CI?  You say that you _don't_ expect `make accessibility-summary` to be used this way.  But I'm not sure what the alternative is: until we've squashed the most-recurrent errors, displaying (never mind erroring on) the per-test output won't be useful.  As you note re: CircleCI's `store_test_results` step, one option is to write a custom `test-results/junit.xml` with the per-test output, but we still won't want to consider these errors as failures yet.

skyvine:
Ah, there was a misunderstanding. I thought that `make validate-test-html` was intended for interactive use only, but I now see that it is used in `.circleci/config.yml`.  I thought that the output needed to be parseable by CircleCI (eg, XML output) in order to be used at all, but it looks like this is incorrect. I added a new commit that updates `.circleci/config.yml` to run `make accessibility-summary` immediately after `make validate-test-html`. Let me know if this would be better done in a separate target, I assume that I'd basically be copy/pasting the `app-page-layout-tests` target and replacing the make command.

skyvine:
As before, I can sqaush these commits once the complete changeset looks good.

skyvine:
Looking at the [circleci output](https://app.circleci.com/pipelines/github/freedomofpress/securedrop/5578/workflows/db802b96-746f-4653-adc0-71cf14e9c86e/jobs/67881/steps), I have a couple of questions.

First, is it expected that there is only output for 5 files? I'm assuming that the layout tests are running on a subset of files to keep it more manageable while working towards validatability, since the `validate-test-html` output also mentions only those files, but I want to confirm in case there is a problem with my commit.

Second, I don't see a way to go to the files containing the full output in circleci. I wrote the target that way with an implicit assumption that this is running locally (I haven't worked with devops tools very much :sweat_smile:), so I'm thinking it might be better to have the make target print the full output after (or, possibly, instead of) the summary. Thoughts?

skyvine:
Hey @cfm, I'm not sure if this needs more active work or not based on the way it looks in CircleCI. As previously implied, this aspect of the CL was not fully considered before. My main point of uncertainty is about the developer's workflow, in order to determine whether or not the current CircleCI output is sufficient. If developers typically run the test suite locally, and CircleCI is just used for communicating results between people, then I expect the current CL will work fine because the developer can examine the files locally. But if developers typically use CircleCI as the mechanism for running the test suite, then it is certainly not sufficient.

In case it is not sufficient, I will mark this as "Work in Progress" again since more work needs to be done. My plan is to write out json files instead of plain text files, and invert the page/error relationship in reporting (eg, "this error occurs in these locations of these pages" instead of "this page contains these errors in these locations) in order to minimize the amount of text that y'all will have to go through. I think this will also make it clearer what amount of impact fixing each error will have, when there is a common cause. But we'll see how it looks  when it's actually implemented.

Please let me know what you think whenever you have a moment!

cfm:
Thanks for following up, @saffronsnail.  From <https://github.com/freedomofpress/securedrop/pull/6745#issuecomment-1460745195>:

> First, is it expected that there is only output for 5 files? I'm assuming that the layout tests are running on a subset of files to keep it more manageable while working towards validatability, since the `validate-test-html` output also mentions only those files, but I want to confirm in case there is a problem with my commit.

Good question!  This behavior isn't caused by anything in your code.  The key lines are:

https://github.com/freedomofpress/securedrop/blob/517765021799e6d4c6c4a2d8318c012976d493ae/.circleci/config.yml#L134
https://github.com/freedomofpress/securedrop/blob/517765021799e6d4c6c4a2d8318c012976d493ae/.circleci/config.yml#L150

Together, these mean that the `app-page-layout-tests` job is `split` across `5` runners, each of which is running a subset of the tests in the test suite balanced according to their historical `timings`.  Since the validation steps run on the *output* of the page-layout tests, both the existing `make test-validate-html` and your `make accessibility-summary` jobs have access only to the output of the tests run by a given runner.

For this reason, although some developers and contributors *may* see these results only in CI and not locally, your current implementation makes sense.  While you *could* "write out json files instead of plain text files, and invert the page/error relationship in reporting (eg, "this error occurs in these locations of these pages" instead of "this page contains these errors in these locations)", in CI that summary will still be only per-runner rather than global.  And that's fine.  :-)

Does this make sense?  If you're happy with the current state of this branch, I'll plan to review it on Monday.

skyvine:
That makes sense! And in that case, I think the branch is fine as-is. Whenever you have the resources to review the CL works for me, I just wanted to make sure I knew if there was anything actionable on my end. =)

.circleci/config.yml
```diff
@@ -162,6 +162,8 @@ jobs:
             fromtag=$(docker images |grep securedrop-test-focal-py3 |head -n1 |awk '{print $2}')
             DOCKER_BUILD_ARGUMENTS="--cache-from securedrop-test-focal-py3:${fromtag:-latest}" \
               make validate-test-html || true
+            DOCKER_BUILD_ARGUMENTS="--cache-from securedrop-test-focal-py3:${fromtag:-latest}" \
+              make accessibility-summary || true
```
cfm, line 177: For clarity I would split this out into its own step, e.g.:

```suggestion
      - run:
          name: Validate accessibility (informational)
          command: |
            fromtag=$(docker images |grep securedrop-test-focal-py3 |head -n1 |awk '{print $2}')
            DOCKER_BUILD_ARGUMENTS="--cache-from securedrop-test-focal-py3:${fromtag:-latest}" \
              make accessibility-summary || true
```

securedrop/tests/functional/pageslayout/accessibility.py
```diff
@@ -0,0 +1,191 @@
...
+        # ar/errors is arbitrary and sufficient
+        for out_filename in os.listdir(_ACCESSIBILITY_DIR / "en_US" / "reviews"):
+            summary[out_filename] = {
+                "reviews": {"count": 0, "locale_differs": False},
+                "errors": {"count": 0, "locale_differs": False},
+            }
+
+            # collect all of the relevant data
+            for message_type in ["reviews", "errors"]:
+                OUTPUTS: Dict[str, Dict[str, List[str]]] = {}
```
cfm: Nit: best just to call this `outputs`, since it's neither a global nor a constant.

cfm (changes requested):
This looks great, @saffronsnail!  CI is currently failing from your fork.  That's definitely not your fault, since the actual tests are passing.  I think it's a side effect of #6739; I'm looking into it.

**Testing:**  Since CI splits test results, I've tested locally with `TESTFILES=tests/functional/pageslayout make test && make accessibility-summary`, which reported on the entire test suite.

**Code review:**  You've designed this nicely, given (a) how complex our test suite is anyway and (b) injecting `html_codesniffer` during each functional test.  I've given some thought to whether "sniff" is a confusing concept and term to introduce in addition to "lint" and "validate".  I think it's fine: it's specific to `html_codesniffer`, and it _is_ different than static linting and validation.

I've left requested two changes inline; please review them when you get a chance.  Tomorrow I'll ask the rest of the team for feedback on the introduction of `npm` and `html_codesniffer` in these Dockerfiles and get back to you.

skyvine:
Thanks @cfm ! I implemented both of the recommended changes in a single new commit. Hopefully `npm` and `html_codesniffer` don't cause any issues!

.circleci/config.yml
skyvine: Added

securedrop/tests/functional/pageslayout/accessibility.py
skyvine: Added... not entirely sure why I chose to name it that way. :/

.circleci/config.yml
```diff
@@ -163,6 +163,13 @@ jobs:
             DOCKER_BUILD_ARGUMENTS="--cache-from securedrop-test-focal-py3:${fromtag:-latest}" \
               make validate-test-html || true
 
+      - run:
+          name: Validate accessibility (infromational)
```
cfm: Typo:

```suggestion
          name: Validate accessibility (informational)
```

devops/docker/CIDockerfile
```diff
@@ -7,8 +7,9 @@ ENV DOCKER_SHA256_x86_64 340e0b5a009ba70e1b644136b94d13824db0aeb52e09071410f35a9
 
 RUN apt-get update && \
     apt-get install -y flake8 make virtualenv ccontrol libpython2.7-dev \
-            libffi-dev libssl-dev libyaml-dev python-pip curl git &&\
-    apt-get clean
+            libffi-dev libssl-dev libyaml-dev python-pip curl git npm &&\
+    apt-get clean && \
+    npm --global install html_codesniffer
```
cfm: @saffronsnail, we discussed this internally last week and reached the consensus that we are not comfortable with an `npm install --global` invocation.  Instead, please run `npm install` locally, commit the resulting `package.json` and `package-lock.json`, and then here do:

```suggestion
    npm ci
```

npm history suggests that even this measure isn't foolproof (e.g., npm/cli#2701, npm/cli#4460), but it's our bare minimum in general and should be adequate in the development and CI environments.

(We may choose to do some follow-up work to minimize the number and size of the toolchains we install in these containers, but that's out of the scope of this pull request.)

securedrop/dockerfiles/focal/python3/Dockerfile
```diff
@@ -7,7 +7,8 @@ RUN apt-get update  && apt-get install -y \
     # cached along with everything else too.
     default-jdk \
     libasound2 libdbus-glib-1-2 libgtk2.0-0 libfontconfig1 libxrender1 \
-    libcairo-gobject2 libgtk-3-0 libstartup-notification0
+    libcairo-gobject2 libgtk-3-0 libstartup-notification0 npm && \
+    npm install --global html_codesniffer
```
cfm: As above:

```suggestion
    npm ci
```

securedrop/tests/functional/pageslayout/accessibility.py
```diff
@@ -0,0 +1,191 @@
...
+def sniff_accessibility_issues(driver: WebDriver, locale: str, test_name: str) -> None:
+    """Runs accessibility sniffs on the driver's current page.
+
+    This function is responsible for injecting HTML CodeSniffer into the current page and writing
+    the results to a file. This way, test functions can focus on the setup required to navigate to
+    a particular URL (for example, logging in to get to the messages page).
+    """
+
+    # 1. Retrieve/compute required data
+    with open(f"/usr/local/lib/node_modules/html_codesniffer/build/HTMLCS.js") as htmlcs:
```
cfm, line 95: With the change from `npm install --global` to `npm ci`, this will need to be a relative rather than an absolute system path.

(Per npm/npm#3175, there doesn't seem to be a way to do `npm ci --global`.)

cfm (changes requested):
Thank for these changes, @saffronsnail.  Please see my new review here, primarily re: npm.

I'm offline for the rest of this week, so I'll revisit this next Monday.

skyvine:
I uploaded a commit that technically uses `npm ci` successfully, but I don't think it's the version that we want to push based on the level of concern you have about using the `npm` toolchain. The issue is that I don't see a way to use the command in the Dockerfile, so it's used in the Makefile, which means the command would run every time anybody runs `make test`, instead of baking it into the image. Due to this concern, which I will  explain in more detail below, I have labelled the commit as "not for production" and I have not signed it.

The primary issue with running it in the Dockerfile is that, if I'm understanding Docker correctly, we don't actually have the repository available. In `securedrop/bin/dev-shell`, the `docker run` invocation passes in `--volume "${TOPLEVEL}:${TOPLEVEL}:Z"`, which creates a bind mount to the local copy of the repository. This is why the output files from the test suite are visible outside of the Docker container. But it also means that the `package.json` and `package-lock.json` are unavailable until we call `docker run`. Additionally, this means that the files are not installed into the docker image directly, but into the developer's system (albeit in the source tree, not in some globally referenced directory). I expect that this would raise some concerns on your end.

I *could* add the html-codesniffer package itself into the repository. The full `cloc` output of the `node-modules` directory shows that it includes > 10,000 lines of javascript, but I might be able to get away with just committing the `HTMLCS.js` file without all the other stuff. I believe this is true because the test code just reads the file and sends the contents to the browser, so I don't see how it could be pulling in any of the other files in that directory. It seems to be self-contained. This would mean adding ~350 lines of javascript. I'm not sure how to prepare this commit in a way that is compliant with their license; there is a notice at the top about the software being subject to the terms of their `license.txt` file, which looks like a fairly normal BSD-style license. Presumably I could just include a copy of that in the commit as well, and make sure that both are in a separate directory. However, I am not a lawyer and I do not know if this would be sufficient, or if it would work for the SecureDrop project. Additionally, this method would increase the maintenance burden because a new copy of the built file would have to be added instead of having commits that essentially just increment a bunch of version numbers.

If adding `HTMLCS.js` directly to the repository works on your end, I'll start working on that. Otherwise, I'm open to ideas.

cfm:
Thanks for this proof-of-concept, @saffronsnail, and for walking me through your thinking.  I think we can solve this dilemma:

> The primary issue with running it in the Dockerfile is that, if I'm understanding Docker correctly, we don't actually have the repository available. In `securedrop/bin/dev-shell`, the `docker run` invocation passes in `--volume "${TOPLEVEL}:${TOPLEVEL}:Z"`, which creates a bind mount to the local copy of the repository. This is why the output files from the test suite are visible outside of the Docker container. But it also means that the `package.json` and `package-lock.json` are unavailable until we call `docker run`.

Good catch.  In the main `Dockerfile`, I'm reminded that we already resort to indirection to work around this problem when we want to verify packages signed with third-party keys, since we can't just include the keys directly in the repository.  For example:

https://github.com/freedomofpress/securedrop/blob/e2d8bbfe4602bdca77c4cae1e13949eca6e2334a/securedrop/dockerfiles/focal/python3/Dockerfile#L18-L20

We could do the same thing here, by embedding the desired `package-lock.json` in the Dockerfile to _make_ it available to `npm ci`.  But that's both overkill and not very extensible, so I think we can keep it simpler with something like a6508747ec849d589941d993cd01ee7255f633aa.


> Additionally, this means that the files are not installed into the docker image directly, but into the developer's system (albeit in the source tree, not in some globally referenced directory). I expect that this would raise some concerns on your end.

Our main concern isn't whether the files themselves are available outside the container; this is true of the Python virtual environment too.  We just want to make sure that `npm ci` is running against a `package-lock.json` with pinned package hashes, which this provides.  :-)


What do you think, @saffronsnail?  @legoktm, if you have time, I'd appreciate your sign-off too.

legoktm:
Sorry, I feel really silly now, I didn't realize that html_codesniffer has no dependencies! Locking is pretty useless then (small benefit in that it pins sha256 checksum but that's pretty negligible). Given that, I think your original `npm install --global html_codesniffer` is fine for now. (I would like to split up the docker image as a whole, it's way too big with too many things, but that's an entirely separate issue we shouldn't tackle in this PR)

skyvine:
Ok, if I understood everything correctly then I think the new state of the PR should be mergable (assuming CircleCI is happy). I went back to the version that used `npm install --global` and squashed everything down to 2 commits, the renaming and the addition of the new tests.

I did add a new line to the Dockerfiles:

```
npm view html_codesniffer 2>/dev/null | head -n 2 | cut -d'|' -f 3 | grep none
```

This line runs _before_ `npm install --global`, and will return a non-zero exit code if `html_codesniffer` has any dependencies added to it down the line. I confirmed that this causes the docker build to fail by repacing `html_codesniffer` with `inquirer`, a random package that has 15 dependencies at time of writing. For context, the output of `view | head` gives:

```

html_codesniffer@2.5.1 | BSD-3-Clause | deps: none | versions: 7
```

When there are dependencies, the word `none` will be replaced with a number. I redirect the error output of `view` into `/dev/null` because `view` gets understandably upset when the stream it's writing to is closed out from under it. If there is some other error like the package being pulled off of the repo, then I expect that will be apparent because the `view` command will not produce the expected text onto standard output.

.circleci/config.yml
skyvine: Fixed

devops/docker/CIDockerfile
skyvine: No longer applicable, per https://github.com/freedomofpress/securedrop/pull/6745#issuecomment-1503797098

securedrop/dockerfiles/focal/python3/Dockerfile
skyvine: No longer applicable, per https://github.com/freedomofpress/securedrop/pull/6745#issuecomment-1503797098

securedrop/tests/functional/pageslayout/accessibility.py
skyvine: No longer applicable, per https://github.com/freedomofpress/securedrop/pull/6745#issuecomment-1503797098

devops/docker/CIDockerfile
```diff
@@ -7,8 +7,12 @@ ENV DOCKER_SHA256_x86_64 340e0b5a009ba70e1b644136b94d13824db0aeb52e09071410f35a9
 
 RUN apt-get update && \
     apt-get install -y flake8 make virtualenv ccontrol libpython2.7-dev \
-            libffi-dev libssl-dev libyaml-dev python-pip curl git &&\
-    apt-get clean
+            libffi-dev libssl-dev libyaml-dev python-pip curl git npm &&\
+    apt-get clean && \
+    # We are safe to do `npm install --global` because html_codesniffer has no
+    # dependencies, this line makes sure that stays true.
+    npm view html_codesniffer 2>/dev/null | head -n 2 | cut -d'|' -f 3 | grep none && \
```
cfm: Nice!  It looks like you can simplify this even further by [asking for just the `dependencies` field](https://docs.npmjs.com/cli/v7/commands/npm-view#description) and checking its length.  Compare:

```sh-session
$ npm view html_codesniffer dependencies
$ npm view express dependencies | head
 
{
  accepts: '~1.3.8',
  'array-flatten': '1.1.1',
  'body-parser': '1.20.1',
  'content-disposition': '0.5.4',
  'content-type': '~1.0.4',
  cookie: '0.5.0',
  'cookie-signature': '1.0.6',
  debug: '2.6.9',
```

cfm (changes requested):
This is looking really good, @saffronsnail.  I've requested one change inline, with which I'll be ready to merge this this week.  :-)

devops/docker/CIDockerfile
cfm: @saffronsnail, I've realized in conversation with @legoktm that even this `npm view` check is overkill.  _Since_ `html_codesniffer@2.5.1` has no dependencies, it's fine just to pin `npm install` to that known-good version.  Since a ["specific name and version combination can never be reused"][unpublish], installing `html_codesniffer@2.5.1` will _never_ pull any dependencies, even if a subsequent version should add some.  (If that happens, we're back in `package-lock.json` territory as previously discussed.)

Pinning to a version we know to be good, which here means "has no dependencies", is the more-general solution to this problem, rather then explicitly checking for no dependencies.

[unpublish]: https://docs.npmjs.com/cli/v9/commands/npm-unpublish

skyvine: Fixed!

cfm (approved):
All changes made and all looks good to me, @saffronsnail.  Thanks for implementing this—and thanks very much for your patience with this long review process.  We appreciate it!

PR #6747, opened 2023-02-13
Reset user 2fa only after entering verification code

### LINKED ISSUE #5237: the JI /account/2fa page text is misleading, giving the impression that 2FA isn't reset until a successful code is input
## Description

When a JI user's TOTP 2FA is reset, the `/account/2fa` page is displayed with the message "you're almost done!" and inviting the user to "finish resetting your two-factor authentication" by updating their mobile device's app with the new secret. A quick read of the instructions would imply that 2FA is reset when the user confirms it by entering the new verification code.

However at that point 2FA has already been reset in the journalists table in the database. 

## Steps to Reproduce

- log in to JI, click on account edit link
- click **RESET MOBILE APP CREDENTIALS**

## Expected Behavior

2FA is reset when a new valid code is entered

## Actual Behavior

2FA is reset immediately, without any confirmation that the user has recorded the new secret

## Comments

To reduce the risk of the user being locked out of their account (by ignoring the CTA to update the secret, or updating the secret incorrectly) either:

1) the page should be updated to say that they *must* update with the new secret, or
2) the secret should *not* be updated in the database until the user inputs a valid verification code.

sea-kelp:
## Status

Ready for review

## Description of Changes

Fixes #5237.

* Reset user 2fa only after entering verification code

As discussed in #5237, this change stores the `otp_secret` and `is_totp` in the Flask session and only commits it to the Journalist user once a valid verification token is entered.

This is a somewhat hacky solution which stages the Journalist `otp_secret` and `is_totp` and rolls them back if the verification token entered is not valid. This is done because [many of the OTP methods](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/models.py#L578-L643) are tightly coupled with the `Journalist` user model.

One way to tackle this would be to move those methods into the new [two_factor.py](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/two_factor.py) file. Happy to take a look at that refactoring if it makes sense but I wanted to get your thoughts!

### Unit tests
Unit tests have been added to cover the following cases:
* User resets own TOTP secret
  * Secret was changed successfully
  * Secret was unchanged because of invalid verification token
* User resets own HOTP secret
  * Secret was changed successfully
  * Secret was unchanged because of invalid verification token
* Admin resets user's TOTP secret
  * Secret was changed successfully
  * Secret was unchanged because of invalid verification token
* Admin resets user's HOTP secret
  * Secret was changed successfully
  * Secret was unchanged because of invalid verification token

Each of these cases can be spot tested manually by performing the following steps:
1. Log in as user/admin
2. Reset user's TOTP/HOTP
2a. If you log out at this point, you should be able to log back in using a token generated using the old secret
3. Enter correct/incorrect verification token
4. Log out
5. Log back in using a token generated using the old/new secret depending on (3)

## Deployment

No special considerations

Choose one of the following:

- [x] These changes do not require documentation (Reviewed the [documentation](https://github.com/freedomofpress/securedrop-docs/blob/5b255edf9899c43491ec8c0dc8a72da535e47f43/docs/journalist.rst#reset-passphrase-or-two-factor-authentication-credentials) for resetting 2fa and did not see any instructions that need updating)

- `admin/requirements.in`

(remove) this section.

Choose one of the following:

### DIFF
--- a/securedrop/tests/functional/pageslayout/test_journalist_admin.py
+++ b/securedrop/tests/functional/pageslayout/test_journalist_admin.py
@@ -15,13 +15,10 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
-import logging
 import time
 from pathlib import Path
-from typing import Callable
 
 import pytest
-from selenium.common.exceptions import TimeoutException
 from selenium.webdriver import ActionChains
 from tests.functional.app_navigators.journalist_app_nav import JournalistAppNavigator
 from tests.functional.pageslayout.utils import list_locales, save_screenshot_and_html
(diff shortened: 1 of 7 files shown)


securedrop/tests/functional/pageslayout/test_journalist_admin.py
```diff
@@ -151,72 +140,36 @@ def screenshot_of_journalist_new_user_two_factor_totp() -> None:
...
         # Then it succeeds
         # Take a screenshot
         save_screenshot_and_html(journ_app_nav.driver, locale, "journalist-admin_edit_totp_secret")
 
-    @staticmethod
-    def _retry_2fa_pop_ups(
-        journ_app_nav: JournalistAppNavigator, navigation_step: Callable, button_to_click: str
-    ) -> None:
-        """Clicking on Selenium alerts can be flaky. We need to retry them if they timeout."""
-        for i in range(15):
```
sea-kelp, line 195: Removed since the warning alerts have been removed

securedrop/tests/test_journalist.py
```diff
@@ -2728,56 +2894,6 @@ def test_valid_user_invalid_first_last_name_change(config, journalist_app, test_
...
-                data=dict(otp_secret=valid_secret),
-            )
-
-            new_secret = Journalist.query.get(test_journo["id"]).otp_secret
-
-            # check that totp is different
-            assert new_secret != old_secret
-
-            # should redirect to verification page
-            ins.assert_redirects(resp, url_for("account.new_two_factor"))
```
sea-kelp, line 2778: Removed these tests since they seem to be duplicates of `test_user_resets_totp` and `test_user_resets_hotp`

Issue #6356, opened 2022-03-23
Publish `.buildinfo` files somewhere


legoktm:
Building Debian packages generates a `.buildinfo` file (see spec: https://wiki.debian.org/ReproducibleBuilds/BuildinfoFiles). These files document what the build environment was at the time of building, so that if someone wanted to reproduce the build, they'd know exactly what packages they needed. Tools like [debrebuild](https://manpages.debian.org/bullseye/devscripts/debrebuild.1.en.html) use buildinfo files to verify that packages are actually reproducible.

Debian currently has a proof-of-concept buildinfo server (<https://buildinfo.debian.net/>) but no official API, so we can really do whatever we want here.

My proposal is to create a new public Git repository, `freedomofpress/buildinfo` (or if that's too generic, `freedomofpress/debian-buildinfo`) that just contains these as an archive for now. Similar to the build-logs repository, maintainers will copy and commit any `*.buildinfo` files to this repo. CI that builds nightlies will also copy these over. We can split them into directories by month (`2022-03/foo.buildinfo`) to avoid one directory from being super giant. The idea would be that we always just keep adding to this, old buildinfo metadata would never be purged from it.

Anyone who wants to try rebuilding our packages can either clone the Git repo or download individual files via the GitHub web interface. And if one day a standardized API exists for serving buildinfo files, we can set that up on top of the Git repo, or import everything in the Git repo into that and archive the repo.

legoktm:
Kev suggested just putting these in the existing build-logs repo, I think that's probably fine.

legoktm:
Overview of the ongoing buildinfo work:
* https://github.com/freedomofpress/kernel-builder/pull/33 output buildinfo files; were committed for the first time in https://github.com/freedomofpress/build-logs/commit/b75928b8c5a0b44de2551091ebf6a771949cb92b
* https://github.com/freedomofpress/securedrop/pull/6754 will start outputting buildinfo files for server packages
* https://github.com/freedomofpress/securedrop-builder/pull/422 will start auto-publishing buildinfo for SDW nightlies
* https://github.com/freedomofpress/securedrop-builder/pull/423 script to verify new packages match their buildinfo, added to CI in https://github.com/freedomofpress/securedrop-apt-test/pull/188
* https://github.com/freedomofpress/securedrop-dev-docs/pull/40 documents buildinfo files

legoktm:
[...]

PR #6754, opened 2023-02-22
Remove molecule from the package building process

### LINKED ISSUE #6706: Stop pinning the "builder" docker image
## Description

We build our packages in a container using a custom docker image (see `molecule/builder-focal`). The way the image is currently used is that it's frozen/pinned, pushed to quay.io, and then the person building the packages will pull it down and use it, as is. But if there are any missing security updates, the build will fail after the fact and then you need to go through the hurdle of updating the builder, rebuilding packages, etc.

Also during the last builder update, I noticed that the logic to detect pending security updates itself was buggy, because it simulates a dist-upgrade and sees if any packages will be pulled in from focal-security, except we don't intend to do a dist-upgrade, so it complaining that a package not in the container was a pending security update!

The theory is that pinning the builder helps with reproducible builds, except it doesn't because the build isn't reproducible for other reasons.

## Proposal

We just build the image on the fly. At runtime we see if there are any pending updates and if so, bail. The deployer can rebuild the image locally and restart the package building. For reproducibility, we will implement #6356, but that shouldn't block this.

legoktm:
## Status

Ready for review

## Description of Changes

Split into 5 commits:
* ~~Pin dh-virtualenv from jammy, not impish~~ Cherry-picked into https://github.com/freedomofpress/securedrop/pull/6755.
* Remove not-useful builder tests (see commit message for detailed rationale)
* Split and slim down test_securedrop_deb_package.py (again, see commit message)
* Re-organize ossec packaging to use debhelper and standard layout (basically the equivalent of https://github.com/freedomofpress/securedrop/pull/6544 but for OSSEC packages)
* Use plain container for package building, not molecule
* Verify all packages except app-code are reproducible in CI (see commit message for detailed rationale)

Fixes #6706.

## Testing

* [x] `make build-debs` ~without this PR~ from `0fb3f24^`, save the debs somewhere ("OG packages")
* [x] `make build-debs && make build-debs-ossec` with this PR ("Round 1 packages")
* [x] diffoscope the old and new packages. Expectations:
   * [x] securedrop-app-code: no non-trivial diffs (compiled .so files will be different, maybe some timestamp variation that is being fixed AFAICT)
   * [x] securedrop-config: should have identical sha256 checksum
   * [x] securedrop-keyring: should have identical sha256 checksum
   * [x] securedrop-ossec-{agent,server}: should have identical sha256 checksum
   * [x] ossec-{agent,server}: honestly it's going to be a mess because all the compiled files are different and the timestamps were previously not reproducible and now they are.
      * It's probably worth just installing these on a prod setup and verifying the notifications get sent properly as far as testing goes.
* [x] Save the ossec-{agent,server} debs from "Round 1" somewhere, then run `make build-debs-ossec` again ("Round 2"). Verify that the sha256 checksums are identical from "Round 1" and "Round 2". Aka the new ossec packages are fully bit-for-bit reproducible.
* [x] CI is passing

## Deployment
My expectation is that we're not actually going to deploy the new ossec-{agent,server} debs because it's roughly the same as the existing 3.6.0 debs. So we can use diffoscope to verify confidence in the main securedrop packages that nothing meaningful is changing.

## Checklist

- [x] I have proposed documentation updates: https://github.com/freedomofpress/securedrop-dev-docs/pull/40


legoktm:
Discussed with Kev today, one thing to add either here or in documentation is what should be done with buildinfo files (see #6356) and investigate setting up a rebuilder to independently verify reproducibility.

legoktm:
mostly finished diffoscoping, in ossec-server, it's capturing the container hostname so it's not reproducible yet:
```
│ │ ├── ./var/ossec/etc/ossec.conf
│ │ │ @@ -1,13 +1,13 @@
│ │ │  <ossec_config>
│ │ │    <global>
│ │ │      <email_notification>yes</email_notification>
│ │ │      <email_to>root@localhost</email_to>
│ │ │      <smtp_server>127.0.0.1</smtp_server>
│ │ │ -    <email_from>ossecm@be0083033c5f</email_from>
│ │ │ +    <email_from>ossecm@c26ce720f917</email_from>
│ │ │    </global>
```

Also I think I unintentionally renamed the ossec.service unit to ossec-agent.service.

legoktm:
The current value is `<email_from>ossecm@focal-sd-generic-ossec-server</email_from>`, I think we can use a fixed hostname like `sd-ossec-builder` (similar to `sd-kernel-builder`).

legoktm:
Also I backed out the change to use debhelper's systemd support because it's probably best done after we move to compat 11. So the service units are installed exactly the same as before.

legoktm:
Some remaining TODOs (but ready for reviews in absence of that):
* [x] Add CI to verify all debs (except securedrop-app-code for now) are reproducible.
* [x] Document that .buildinfo files need to be published (n.b. this should cover SDW too)

builder/Dockerfile
```diff
@@ -1,10 +1,9 @@
-# ubuntu:focal-20210921
-FROM ubuntu@sha256:3555f4996aea6be945ae1532fa377c88f4b3b9e6d93531f47af5d78a7d5e3761
-
+FROM ubuntu:focal
```
cfm, line 1: This is one half of the documentation change @eloquence requested last week: removing <https://developers.securedrop.org/en/latest/dockerbuildmaint.html>.  The other half, per <https://github.com/freedomofpress/securedrop/pull/6754#issuecomment-1444097514>, is documenting the generation of `.buildinfo` files instead.

cfm (commented):
I've read through this both commit- and diff-wise and think it looks fabulous, @legoktm.  The precedent and benefit of #6544 is evident!

I'll try to run through the test plan this week, next Monday at the latest.  In the meantime: do you see freedomofpress/securedrop-engineering#20 as a prerequisite for approving (in addition to documenting the goals and principles of) this work?

legoktm:
Thanks!

> I'll try to run through the test plan this week, next Monday at the latest. In the meantime: do you see [freedomofpress/securedrop-engineering#20](https://github.com/freedomofpress/securedrop-engineering/pull/20) as a prerequisite for approving (in addition to documenting the goals and principles of) this work?

No. Our current security posture/requirement is that we build debs in a dispVM that builds/executes a container, which doesn't change with this PR (just how the container is set up and what it does), my proposal is to drop the dispVM part, which is what the dep-check stuff is for (since you don't really need to worry about out of date containers in a dispVM), but that should be mostly independent to this.

And I'll work on the documentation stuff in the next few days.

legoktm:
Documentation PR: https://github.com/freedomofpress/securedrop-dev-docs/pull/40

builder/Dockerfile
cfm: This is proposed in freedomofpress/securedrop-dev-docs#40.

cfm (dismissed):
I've run through a full set of builds according to the test plan, comparing packages built from `0fb3f24^` (as the parent of this branch, to filter out any noise from `develop`) and from this branch.  All look good.  Thanks, @legoktm, for the hints on what level of reproducibility, thus what size and kind of diff, to expect for each packge.

I've received a few OSSEC alerts *about* the new `ossec-server`:

```
Integrity checksum changed for: '/var/ossec/checksdconfig.py'
Permission change: 'rwxr-xr-x' to 'r-xr-x---'
```

```
Integrity checksum changed for: '/var/ossec/etc/ossec.conf'
Permission change: 'rw-r--r--' to 'r--r-----'
```

```
Integrity checksum changed for: '/var/ossec/etc/resolv.conf'
What changed:
1c1
< nameserver 10.139.1.1
---
> -e nameserver 10.139.1.1
```

```
Integrity checksum changed for: '/var/ossec/etc/ossec-init.conf'
What changed:
3c3
< DATE="Thu Mar 11 20:36:21 UTC 2021"
---
> DATE="Mon Apr 13 18:53:21 UTC 2020"
```

I've now reached `Max emails per hour reached.`  Assuming (a) that these diffs are expected, @legoktm, and (b) that alerts come through as usual overnight, I'll merge in the morning.

legoktm:
Hmm, the permissions changes are not intentional, let me poke at that tomorrow. (The DATE one is though)

legoktm:
The easiest way to verify the permission fixes is to use the `debdiff` tool (`devscripts` package in Debian/Fedora). Before today's commits, it looked like https://gist.github.com/legoktm/897ca3a06677385994311a96c8d98aa7 (where original-server is the deb from https://apt.freedom.press/pool/main/o/ossec.net/). It lists files as different if they have different perms/owners/names but ignores content.

After the fixperms fixes, it looks lke:
```
user@dev ~/g/f/s/b/focal> debdiff original-server.deb ossec-server_3.6.0+focal_amd64.deb
[The following lists of changes regard files as different if they have
different names, permissions or owners.]

Files in second .deb but not in first
-------------------------------------
-rw-r--r--  root/root   /usr/share/doc/ossec-server/changelog.gz
-rw-r--r--  root/root   DEBIAN/conffiles
-rw-r--r--  root/root   DEBIAN/md5sums

Files in first .deb but not in second
-------------------------------------
-rw-r--r--  root/root   /usr/share/doc/ossec-server/changelog.Debian
-rw-r--r--  root/root   /usr/share/lintian/overrides/ossec-server
-rw-r--r--  root/root   DEBIAN/templates
-rwxr-xr-x  root/root   DEBIAN/prerm

...
```

All of the remaining changes are expected and desired because they're actually files being added/removed.

cfm (approved):
Thanks, @legoktm.  I've `debdiff`ed and installed the new `ossec-{agent,server}` packages, and I'll merge tomorrow assuming (a) OSSEC alerts come through as usual overnight and (b) CircleCI comes back from the dead.

cfm (changes requested):
Oops: can you rebase and resolve the conflicts with `develop` after #6739?

cfm:
Rebased from `develop` without changes.

cfm (approved):
After <https://github.com/freedomofpress/securedrop/pull/6754#pullrequestreview-1340506894>, OSSEC is happy, CI is happy, and the conflicts with `develop` are resolved—so I'm merging.  Thanks for this, @legoktm!

PR #6785, opened 2023-05-02
feat(manage): add `remove-pending-sources` command

### LINKED ISSUE #6488: Allow for removal of "pending" source accounts
## Description

When a source account is created but before a first submission, by default a "pending" flag is set. Pending accounts are not displayed in the journalist interface.

If a user then exits their session and never returns to provide a submission, the account will remain in a pending state indefinitely, and cannot be removed via the Journalist Interface. On long-running high-volume instances this increases the source user count significantly, requiring admin intervention to fix.

One approach to allow for the removal of unused source accounts would be to change the `pending` boolean flag to a datetime value, and purge unused accounts via a cronjob or similar after a set time (say, a month). This has the disadvantage of increasing metadata (source account creation time) about prospective sources, though the field could be set to `null` on submission, meaning that said metadata would not be stored for sources that were actually ever active. This could be mitigated somewhat by giving the datetime a resolution on the order of days or weeks.

Alternatively, the fact that active source sessions without submissions would be nuked when the purge ran could just be accepted. 

Note that this would be moot if/when inverted flow changes land, as accounts would then only be created on first submission. So it might not be worth the effort if it was just a temporary measure.

## To do
1. [x] Add `manage.py` command
2. [ ] Add tests, especially for `--keep-most-recent`'s offset logic
3. [ ] Extract standalone script for testing purposes

## User Research Evidence

long-term observations of instance behavior...

cfm:
## Status

Ready for review

## Description of Changes

Closes #6488 by:
1. adding a command `manage.py remove-pending-sources --keep-most-recent=n` to remove all but the most-recent `n` `Source`s (and their keys) with `pending=True`; and
2. correcting `loaddata.add_source()` so that new `Source`s are marked `pending=False` only in `loaddata.record_source_interaction()`.

Co-authored-by: @nathandyer
- who did the heavy lifting in `manage.py` and `securedrop.management.sources`

### Against `develop`

In particular, after #6563:

- [ ] CI passes
- [ ] Interactive test cases pass as expected, e.g.:

```sh-session
$ ./loaddata.py --source-count 13  # 2 files+messages per source → pending=False
$ ./manage.py remove-pending-sources --keep-most-recent 10  # remove 0

$ ./loaddata.py --source-count 13 --files-per-source 0 --messages-per-source 0 --replies-per-source 0  # pending=True
$ ./manage.py remove-pending-sources --keep-most-recent 10  # remove 3 of 13

$ ./loaddata.py --source-count 10 --files-per-source 0 --messages-per-source 0 --replies-per-source 0  # pending=True
$ ./manage.py remove-pending-sources  # remove 0 of 10, per default "--keep-most-recent 100"
```

### Against v2.5.2

0. Visually review <https://gist.github.com/nathandyer/bb31b6c9eaed30b078230922f437c189>
    * Suggestion: diff against <https://github.com/freedomofpress/securedrop/blob/2.5.2/securedrop/manage.py
1. Upload `manage-6488.py` to `$HOME` on the Application Server
2. Patch `loaddata.py` per 2ec4613b4788b324e2dde2d35b3cf36710452c5c
4. [ ] Test cases pass as above

## Deployment

<https://gist.github.com/nathandyer/bb31b6c9eaed30b078230922f437c189> can be uploaded directly for testing purposes.  Otherwise, no deployment implications.

Neither Apache nor Tor is permitted access to `manage.py` and its supporting modules per `securedrop/debian/app-code/etc/apparmor.d`, so the new `securedrop.management.sources` does not need to be added.

### DIFF
--- a/securedrop/management/sources.py
+++ b/securedrop/management/sources.py
@@ -21,12 +21,18 @@ def remove_pending_sources (args: argparse.Namespace) -> int:
     """
     n = args.keep_most_recent
     sources = find_pending_sources(n)
+    print(f"Found {len(sources)} pending sources")
+
+    deleted = []
     for source in sources:
         try:
             EncryptionManager.get_default().delete_source_key_pair(source.filesystem_id)
         except GpgKeyNotFoundError:
             pass
         delete_pending_source(source)
+        deleted.append(source)
+
+    print(f"Deleted {len(sources)} pending sources")
     return 0
 
 def find_pending_sources(keep_most_recent: int) -> List[Source]:
--- a/securedrop/management/sources.py
+++ b/securedrop/management/sources.py
@@ -1,0 +1,52 @@
+import argparse
+import datetime
+import os
+import sys
+import time
+from argparse import _SubParsersAction
+from typing import List, Optional
+
+from db import db
+from flask.ctx import AppContext
+from management import app_context
+from encryption import EncryptionManager, GpgKeyNotFoundError
+from models import Source
+from rm import secure_delete
+
+
+def remove_pending_sources (args: argparse.Namespace) -> int:
+    """
+    Removes pending source accounts, with the option of keeping
+    the n newest source accounts.
+    """
+    n = args.keep_most_recent
+    sources = find_pending_sources(n)
+    for source in sources:
+        try:
+            EncryptionManager.get_default().delete_source_key_pair(source.filesystem_id)
+        except GpgKeyNotFoundError:
+            pass
+        delete_pending_source(source)
+    return 0
+
+def find_pending_sources(keep_most_recent: int) -> List[Source]:
+    """
+    Finds all sources that are marked as pending
+    """
+    with app_context():
+    	pending_sources = Source.query.filter_by(pending=True).order_by(Source.id.desc()).offset(keep_most_recent).all()
+
+    return pending_sources
+
+def delete_pending_source(source: Source) -> None:
+    """
+    Delete a pending source from the database
+    """
+    if source.pending:
+        with app_context():
+            try:
+                db.session.delete(source)
+                db.session.commit()
+            except Exception as exc:
+                db.session.rollback()
+                print(f"ERROR: Could not remove pending source: {exc}.")
--- a/securedrop/management/sources.py
+++ b/securedrop/management/sources.py
@@ -1,20 +1,13 @@
 import argparse
-import datetime
-import os
-import sys
-import time
-from argparse import _SubParsersAction
-from typing import List, Optional
+from typing import List
 
 from db import db
-from flask.ctx import AppContext
-from management import app_context
 from encryption import EncryptionManager, GpgKeyNotFoundError
+from management import app_context
 from models import Source
-from rm import secure_delete
 
 
-def remove_pending_sources (args: argparse.Namespace) -> int:
+def remove_pending_sources(args: argparse.Namespace) -> int:
     """
     Removes pending source accounts, with the option of keeping
     the n newest source accounts.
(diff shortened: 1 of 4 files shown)


cfm:
`static-analysis-and-no-known-cves` is failing because `Vulnerability found in ansible version 2.9.26`.

zenmonkeykstop:
> `static-analysis-and-no-known-cves` is failing because `Vulnerability found in ansible version 2.9.2`.

Was all :thinking: but looks like comment was cut off? the version in question is 2.9.26. Also have a safety check failing in `paramiko`.

zenmonkeykstop (approved):
Ran through test plan in `make dev`, looks good! (Also verified that the expected keys are actually being removed.)

This is good to go once rebased against pending security update PRs.

zenmonkeykstop:
Reviewed force-push diff, still good to go.

securedrop/management/sources.py
```diff
@@ -0,0 +1,58 @@
...
+    print(f"Found {len(sources)} pending sources")
+
+    deleted = []
+    for source in sources:
+        try:
+            EncryptionManager.get_default().delete_source_key_pair(source.filesystem_id)
+        except GpgKeyNotFoundError:
+            pass
+        delete_pending_source(source)
+        deleted.append(source)
```
legoktm, line 26: `deleted` is never used? Maybe it was supposed to be used below in the message being printed, but if we just need to know the count, we could just increment a counter instead of holding onto the entire source object?

securedrop/management/sources.py
```diff
@@ -0,0 +1,58 @@
...
+    print(f"Deleted {len(sources)} pending sources")
+    return 0
+
+
+def find_pending_sources(keep_most_recent: int) -> List[Source]:
+    """
+    Finds all sources that are marked as pending
+    """
+    with app_context():
+        pending_sources = (
```
legoktm, line 37: nit: you can return directly inside a `with` block and it does the right thing, you don't need to set it to a variable and return outside of it.

securedrop/management/sources.py
```diff
@@ -0,0 +1,58 @@
...
+        )
+
+    return pending_sources
+
+
+def delete_pending_source(source: Source) -> None:
+    """
+    Delete a pending source from the database
+    """
+    if source.pending:
```
legoktm, line 51: If we do need to check this (shouldn't the filter in the query be good enough?), shouldn't it be one level higher, before we delete the GPG key? Or maybe the key deletion should be moved into this function?

securedrop/management/sources.py
```diff
@@ -0,0 +1,58 @@
...
+        with app_context():
+            try:
+                db.session.delete(source)
+                db.session.commit()
+            except Exception as exc:
```
legoktm, line 56: What kind of exceptions are we expecting here?

(n.b. I thought Flask/sqlalchemy automatically rolled back when exceptions are thrown.)

securedrop/management/sources.py
```diff
@@ -0,0 +1,58 @@
...
+from management import app_context
+from models import Source
+
+
+def remove_pending_sources(args: argparse.Namespace) -> int:
+    """
+    Removes pending source accounts, with the option of keeping
+    the n newest source accounts.
+    """
+    n = args.keep_most_recent
```
legoktm, line 15: nit: I think we should avoid one-character variables whenever possible, in this case why not pass it directly, `find_pending_sources(args.keep_most_recent)`

legoktm (commented):
I had to peek at this code as part of rebasing my Sequoia branch, so I gave it a quick review since it was pretty new.

securedrop/management/sources.py
nathandyer: Thanks @legoktm, I suspect you are correct about it rolling back automatically. I was using the same logic/workflow that I found [within manage.py itself.](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/manage.py#LL216C1-L219C34) Based on that, should this try/except block be removed within the branch? What about in manage.py? Or are they not roughly equivalent use cases?

nathandyer:
Thanks again for the extra review @legoktm! I'll leave a couple notes inline, and have opened a PR for the changes I've already implemented: #6793

securedrop/management/sources.py
nathandyer: This was done intentionally, to make the function usable outside of the context of this specific need, should it be useful for any other tasks. I don't think the key deletion should be moved here, as you would lose this benefit. I do agree that checking that the source is actually pending is not needed based on the current use and implementation (as you noted, the query is good enough), but I figure it would be needed if it were to be called somewhere else if it ever gets reused in the future.

Issue #6774, opened 2023-03-21
building packages should automatically save build log


cfm:
> I do think that if we want to keep publishing build logs long-term it should be automated, e.g. `make build-debs` should also spit out a properly named build log that can be directly committed.

_Originally posted by @legoktm in https://github.com/freedomofpress/securedrop-dev-docs/pull/40#discussion_r1136113719_

* [x] #6787
* [x] freedomofpress/securedrop-client#1789

cfm:
The DX working group discussed the following options today:
1. documenting the use of the `script` command to capture the entire `sd-dvm` build session;
2. wrapping `script` in a do-nothing script; and
3. actually scripting this use of something like `script -c make build-debs`.

(Enough `script` scripting?)  I'll time-box a first pass on this during an extra workday I'll have floating around next week, perhaps even pairing with @legoktm in person.  :-)

legoktm:
I tried using `script` today, overall it was straightforward, the only problem is that it very literally captured all output, including escape sequences and progress things. This means tools like `less` consider it as containing binary:
```
$ less workstation.log 
"workstation.log" may be a binary file.  See it anyway?
``` 
And I can't even upload it as a gist to show you:
```
$ gh gist create workstation.log 
failed to collect files for posting: failed to upload workstation.log: binary file not supported
```

I manually copied + pasted for https://gist.github.com/legoktm/0b589ecb8852727546f998b4c7902dc5 but presumably that lost some binary information?

Compare with the manually copy-paste-out-of-terminal logs: https://github.com/freedomofpress/build-logs/commit/08b5f09f4871253876e92a12f671d00e9df58fe1

cfm:
I find some of the non-custom suggestions in <https://unix.stackexchange.com/questions/14684/removing-control-chars-including-console-codes-colours-from-script-output> reasonable.  But I would understand if we decided this is more trouble than it's worth.

cfm:
@legoktm, running with `TERM=dumb` seems much cleaner, e.g.:

```patch
--- a/Makefile
+++ b/Makefile
@@ -354,7 +354,8 @@ endif
 .PHONY: build-debs
 build-debs: ## Build and test SecureDrop Debian packages
        @echo "Building SecureDrop Debian packages..."
-       @$(SDROOT)/builder/build-debs.sh
+       @export TERM=dumb
+       @script -c $(SDROOT)/builder/build-debs.sh
        @echo
 
 .PHONY: build-debs-notest
```

[...]

PR #6787, opened 2023-05-03
build(build-debs*): log all output via `script`

cfm:
## Status

Ready for review

## Description of Changes

Towards #6774, wraps all `build-debs.sh` invocations in `script` and displays the filename of the transcript after its conclusion.

This should be easily replicable to `securedrop-builder`.

## Testing

- Run one or more of the `make build-debs*` targets
- `^C` freely
- [ ] A transcript `build/${TIMESTAMP}-${TARGET}.log` is created...
- [ ] ...and is legible

## Deployment

Developer-only; no deployment considerations.

Choose one of the following:

- I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-docs) for these changes, or will do so later

...because they are self-documenting (-prompting), but I will probably document them anyway after merge.  :-)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -357,41 +357,57 @@ endif
 #
 ###########
 
+SCRIPT_MESSAGE="You can now examine or commit the log at:"
+SCRIPT_OUTPUT_PREFIX=$(SDROOT)/build/$(shell date +%Y%M%d)
+SCRIPT_OUTPUT_EXT=log
+
 .PHONY: build-debs
+build-debs: OUT:=$(SCRIPT_OUTPUT_PREFIX)-securedrop.$(SCRIPT_OUTPUT_EXT)
 build-debs: ## Build and test SecureDrop Debian packages
 	@echo "Building SecureDrop Debian packages..."
 	@export TERM=dumb
 	@script \
 		--command $(SDROOT)/builder/build-debs.sh \
-		$(SDROOT)/build/$(shell date --iso-8601=seconds)-$@.log
+		$(OUT)
 	@echo
+	@echo "$(SCRIPT_MESSAGE)"
+	@echo "$(OUT)"
 
 .PHONY: build-debs-notest
+build-debs-notest: OUT:=$(SCRIPT_OUTPUT_PREFIX)-securedrop.$(SCRIPT_OUTPUT_EXT)
 build-debs-notest: ## Build SecureDrop Debian packages without running tests.
 	@echo "Building SecureDrop Debian packages, skipping tests..."
 	@export TERM=dumb
 	@NOTEST=1 script \
 		--command $(SDROOT)/builder/build-debs.sh \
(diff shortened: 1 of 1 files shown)


legoktm (changes requested):
On a technical level this is great. Some things that would make it even better:

1. Can we use a better filename? `2023-05-03T22:24:29-04:00-build-debs.log` is a bit...terrible (especially the colons :p). Looking through build-logs, I think something like `YYYYMMDD-securedrop-X.Y.Z.log` and similar for the OSSEC packages. I don't think we need to label it differently whether or not tests were run, since that'll be in the log itself (or not). Also, will this overwrite the file if it already exists? Or append? If it's the latter, we should probably use a complete timestamp instead of just the day.
2. Can we print the commit sha1/tag we are building from?
3. (bonus points) Can we verify said sha1/tag is signed? But in some optional way so if you're doing local test builds it just prints e.g. "WARNING: building from unsigned commit/tag" or something.

No 2 and 3 could probably be done in build-debs.sh?

With those 3 then I think building packages would be 1) clone repo 2) checkout commit/tag 3) `make build-debs` and nothing else.

Minor: curious why you made `$SCRIPT_OUTPUT_EXT` a variable instead of just hardcoding it.

cfm:
Thanks, @legoktm!

On Wed, May 03, 2023 at 07:39:15PM -0700, Kunal Mehta wrote:
> 1. Can we use a better filename?
> `2023-05-03T22:24:29-04:00-build-debs.log` is a bit...terrible
> (especially the colons :p). Looking through build-logs, I think
> something like `YYYYMMDD-securedrop-X.Y.Z.log` and similar for the
> OSSEC packages. I don't think we need to label it differently whether
> or not tests were run, since that'll be in the log itself (or not).
> 2. Can we print the commit sha1/tag we are building from?
> 3. (bonus points) Can we verify said sha1/tag is signed? But in some optional way so if you're doing local test builds it just prints e.g. "WARNING: building from unsigned commit/tag" or something.

My goal here was really just "${TIMESTAMP}-${TARGET}.${EXT}".  :-)  But
these are all good, worthwhile suggestions.  Unless you really want to
combine them, I'd prefer to simplify (1) in this pull request and defer
(2) and (3) to a separate one, since they go beyond the original #6774.
In any case, I can make these changes next week.


> Minor: curious why you made `$SCRIPT_OUTPUT_EXT` a variable instead of
> just hardcoding it.

DRYer until proven otherwise, is my instinct.  :-)

legoktm:
> Unless you really want to combine them, I'd prefer to simplify (1) in this pull request and defer (2) and (3) to a separate one

Fine with me :)

cfm:
Back to you, @legoktm, with the rest of your suggestions filed in #6792.

legoktm (approved):
Awesome, works as desired!

Issue #6789, opened 2023-05-05
Release securedrop-grsec 5.15.89-2


legoktm:
This is the first step in fixing #6762. As described there, we need to push out a version of the securedrop-grsec metapackage that depends on both the current and previous kernel versions before the 2.6.0 release.

Because this is a one-time operation, I've pushed the necessary changes to a separate kernel-builder branch: https://github.com/freedomofpress/kernel-builder/tree/server-hack-6762 - specifically https://github.com/freedomofpress/kernel-builder/commit/79847811a0a4c227442ef234fa6c610e5be01f14. The package is fully reproducible and has been independently rebuild/verified by another SD maintainer.

## Test plan
* [x] Diffoscope with the [current package](https://apt.freedom.press/pool/main/l/linux-upstream/securedrop-grsec_5.15.89-grsec-securedrop-1_amd64.deb), the full diff should be the version number and dependency list:
```
--- securedrop-grsec_5.15.89-grsec-securedrop-1_amd64.deb
+++ securedrop-grsec_5.15.89-grsec-securedrop-2_amd64.deb
├── control.tar.xz
│ ├── control.tar
│ │ ├── file list
│ │ │ @@ -1,3 +1,3 @@
│ │ │  drwxr-xr-x   0 root         (0) root         (0)        0 2011-06-29 20:23:37.000000 ./
│ │ │ --rw-r--r--   0 root         (0) root         (0)      543 2011-06-29 20:23:37.000000 ./control
│ │ │ +-rw-r--r--   0 root         (0) root         (0)      581 2011-06-29 20:23:37.000000 ./control
│ │ │  -rwxr-xr-x   0 root         (0) root         (0)     2029 2011-06-29 20:23:37.000000 ./postinst
│ │ ├── ./control
│ │ │ @@ -1,13 +1,13 @@
│ │ │  Package: securedrop-grsec
│ │ │  Source: linux-upstream
│ │ │ -Version: 5.15.89-grsec-securedrop-1
│ │ │ +Version: 5.15.89-grsec-securedrop-2
│ │ │  Architecture: amd64
│ │ │  Maintainer: SecureDrop Team <securedrop@freedom.press>
│ │ │  Installed-Size: 15
│ │ │ -Depends: linux-image-5.15.89-grsec-securedrop, intel-microcode, paxctld
│ │ │ +Depends: linux-image-5.15.89-grsec-securedrop, linux-image-5.15.81-grsec-securedrop, intel-microcode, paxctld
[...]

Issue #6762, opened 2023-03-01
unattended-upgrades does not automatically remove old kernel packages


legoktm:
## Plan to fix

1. [x] Release securedrop-grsec metapackage that depends on current and previous kernel: #6789
2. [x] Set `Unattended-Upgrade::Remove-Unused-Dependencies "true"`; via securedrop-config package, ship in 2.6.0 release: https://github.com/freedomofpress/securedrop/pull/6791
3. [x] Build next (May hopefully) kernel with correct version format (https://github.com/freedomofpress/kernel-builder/pull/33), but keep a manual dependency on 5.15.89 so we still have an old kernel installed.
4. [x] Build next next (June) kernel using the new version format, drop the manual dependency on 5.15.89. unattended-upgrades will automatically keep 3 kernel packages around, so the May kernel should not be removed.

## Description

unattended-upgrades should, by default, automatically remove old kernel packages to prevent too many from accumulating and taking up disk space (especially on /boot). But it doesn't

## Steps to Reproduce

1. Have an instance that's been running for a while through multiple kernel versions.
3. Run `sudo apt autoremove`

## Expected Behavior

* There are no kernel packages pending autoremoval because unattended-upgrades already removed them

## Actual Behavior

* You see:
```
The following packages will be REMOVED:
  linux-image-5.15.26-grsec-securedrop linux-image-5.15.57-grsec-securedrop
```

## Comments

The regex used by unattended-upgrades is https://github.com/mvo5/unattended-upgrades/blob/2.3/unattended-upgrade#L716

The simplified version is `re.compile('(^linux-.*-[1-9][0-9]*\\.[0-9]+\\.[0-9]+-[0-9]+(-.+)?$)')`. The gotcha is that this regex matches against packages with a Debian version, not a plain upstream version. So we basically need linux-image-5.15.74<b>-1</b>-grsec-securedrop instead of our current linux-image-5.15.74-grsec-securedrop.

legoktm:
Fixing the package name to include the Debian revision seems straightforward. I think we just need to throw in a "-1" in https://github.com/freedomofpress/kernel-builder/blob/main/scripts/mkdebian#L89

This wouldn't take care of existing kernel packages. We could advise admins to run `sudo apt autoremove -y` as a one-off? Or maybe don't bother if they haven't run out of disk space yet.

This should probably be a blocker for the March kernel update. We should also take the opportunity to verify that unattended-upgrades does keep the last 2 old kernels or however many we want.

[...]

PR #6791, opened 2023-05-08
Have unattended-upgrades remove unused dependencies

legoktm:
## Status

Ready for review

Towards https://github.com/freedomofpress/securedrop/issues/6762

## Description of Changes

By setting `Unattended-Upgrade::Remove-Unused-Dependencies "true";` unattended-upgrades will remove "unused" dependencies, which are packages that were pulled in as a dependency but no longer have anything depending on them and, in theory, can be safely removed.

The main motivation in the short term is to have this remove old kernel packages (#6762), but this is probably good in the long run too, since it will ensure the package set on older instances is roughly the same as fresh installs.

It's definitely possible we depend on something that is just an implicit dependency and it gets auto removed in the future, but such a bug would have already caused issues for fresh installs and should be caught by our nightly staging tests.

## Testing
On a staging/prod setup:

* [ ] Manually edit `/etc/apt/apt.conf.d/50unattended-upgrades` to set `Unattended-Upgrade::Remove-Unused-Dependencies "true";`
* [ ] Install and immediately remove a package with some dependencies, e.g. `sudo apt install -y emacs-nox && sudo apt purge -y emacs-nox`
* [ ] Run `sudo unattended-upgrades --dry-run -d`, observe that the relevant emacs packages would be removed, e.g. `marking emacs-common for removal`

## Deployment

Any special considerations for deployment? Yes! This can only be released after https://github.com/freedomofpress/securedrop/issues/6789 is released, which has now happened.

## Checklist

### DIFF
--- a/securedrop/debian/config/opt/securedrop/50unattended-upgrades
+++ b/securedrop/debian/config/opt/securedrop/50unattended-upgrades
@@ -40,7 +40,7 @@ Unattended-Upgrade::AutoFixInterruptedDpkg "true";
 
 // Do automatic removal of new unused dependencies after the upgrade
 // (equivalent to apt-get autoremove)
-//Unattended-Upgrade::Remove-Unused-Dependencies "false";
+Unattended-Upgrade::Remove-Unused-Dependencies "true";
 
 // Automatically reboot *WITHOUT CONFIRMATION*
 //  if the file /var/run/reboot-required is found after the upgrade


rocodes (approved):
## Testing
On a staging/prod setup:

* [x] Manually edit `/etc/apt/apt.conf.d/50unattended-upgrades` to set `Unattended-Upgrade::Remove-Unused-Dependencies "true";`
* [x] Install and immediately remove a package with some dependencies, e.g. `sudo apt install -y emacs-nox && sudo apt purge -y emacs-nox`
* [x] Run `sudo unattended-upgrades --dry-run -d`, observe that the relevant emacs packages would be removed, e.g. `marking emacs-common for removal`

Thanks for crystal clear plan :) :shipit:

PR #6780, opened 2023-04-12
Switch cronjobs to systemd timers

legoktm:
## Status

Ready for review

## Description of Changes

systemd timers are easier to define and offer more flexibility in running commands. For example, 9169606decd3c81c would've been a 2 line change instead of a complex and fragile bash migration.

We clear the old crontabs in the postinst and let debhelper enable the timers.

Turns out our usage of `dh_systemd_enable` and `dh_systemd_start` was wrong, because we were calling it multiple times without specifying explicit unit files, it was adding the same stanza to the postinst/postrm multiple times. We now invoke it explicitly for each unit we want enabled/started. Hopefully when we switch to compat 12/dh_installsystemd we don't need to manually list each one.

Fixes #6705.
Fixes #6748.

## Testing

* [x] Run `make build-debs`. Copy the securedrop-app-code deb into a temporary directory (so it doesn't get messy) and extract the maintainer scripts with `ar x securedrop-app-code_2.6.0\~rc1+focal_amd64.deb && tar xvf control.tar.xz`. Verify that the 2 timers and 4 pre-existing service units all have identical stanzas in the `postinst`, `postrm`, and `prerm`.
  * [x] You can also run `make build-debs` without this PR and diffoscope against the new build with this PR, but I found enough stuff got re-arranged that the diff wasn't very useful to look at.
* [ ] Set up a staging/prod instance with either 2.5.1 or develop before this PR. Then with this PR, run `make build-debs` and install the securedrop-app-code deb with `sudo apt install ./securedrop-app-code_2.6.0~rc1+focal_amd64.deb`
* [ ] Run `systemctl list-timers`, see the 2 new timers appear
   * [ ]  `securedrop-clean-tmp` should be scheduled for 00:00 UTC.
   * [ ] `securedrop-submissions-today` should be scheduled for 1 hour before the reboot. Easiest way to verify this is add the amount of hours left (e.g. mine says 9h left) with the uptime (mine is 14h), it should add up to 23h.
* [ ] Manually create a backdated temporary file to be deleted: `sudo -u www-data touch /var/lib/securedrop/tmp/delete-me -mt 2303300000`
* [ ] Run `sudo crontab -l` and `sudo crontab -u www-data -l`, both should say "no crontab for root/www-data"
* [ ] Run `systemctl status securedrop-clean-tmp.service`, verify it is not enabled and not active. Then run  `systemctl status securedrop-clean-tmp.timer`, verify it is active and enabled.
[...]

PR #6826, opened 2023-05-30
feat: add systemd timer to run "manage.py remove-pending-sources" daily

### LINKED ISSUE #6788: schedule regular run of `remove-pending-sources`
## Description

Add a ~cron job~ `systemd` timer for `manage.py remove-pending-sources`.

- [ ] Decide on interval: nightly? weekly?
- [ ] Decide on value (or change default) of `--keep-most-recent`

**Depends on:**
- #6785
- #6705
    - #6780

cfm:
## Status

Ready for review

## Description of Changes

Closes #6788.

**NB.**  Per <https://github.com/freedomofpress/securedrop/pull/6780#discussion_r1205934851>, we are trusting that well-formed systemd units will Do the Right Thing without specific test cases.

## Testing

1. Install `securedrop-app-code` built via `make build-debs` on this branch into a production or staging installation.
2. Generate more than 100 "pending" sources, e.g.: `loaddata.py --source-count 110 --files-per-source 0 --messages-per-source 0 --replies-per-source 0`
3. Wait for the nightly timers to fire.
4. [ ] Confirm that only 100 "pending" sources remain on the server.
5. ~Check that no OSSEC alert was logged when the `securedrop-remove-pending-sources` timer fired (i.e., its standard output was suppressed).~ (obsoleted by <https://github.com/freedomofpress/securedrop/pull/6826#discussion_r1218565000>)
6. **Bonus:**
    1. Mangle `manage.py remove-pending-sources` so that it will raise an exception.
    2. Repeat steps (2) and (3).
    3. Confirm that an OSSEC alert *was* logged when the `securedrop-remove-pending-sources` timer fired ~(i.e., its standard error was *not* suppressed)~, because OSSEC triggered on a log message containing the string `error`. (clarified per <https://github.com/freedomofpress/securedrop/pull/6826#discussion_r1218565000>)

## Deployment

No deployment considerations, but it's worth documenting this behavior for administrators and flagging it in release notes, as noted below.

Choose one of the following:

- These changes do not require documentation

### DIFF
--- a/securedrop/debian/app-code/lib/systemd/system/securedrop-remove-pending-sources.service
+++ b/securedrop/debian/app-code/lib/systemd/system/securedrop-remove-pending-sources.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=job to remove pending SecureDrop sources daily
+
+[Service]
+ExecStart=/var/www/securedrop/manage.py remove-pending-sources
+PrivateDevices=yes
+PrivateTmp=yes
+ProtectSystem=full
+ReadOnlyDirectories=/
+ReadWriteDirectories=/var/lib/securedrop
+User=www-data
+WorkingDirectory=/var/www/securedrop
--- a/securedrop/debian/rules
+++ b/securedrop/debian/rules
@@ -62,12 +62,13 @@ override_dh_gencontrol:
 override_dh_installinit:
 	dh_installinit --noscripts
 
-# We want to enable all systemd units except the two that are run by timers,
-# so we have to mark those two as --no-enable first, then the final
-# `dh_systemd_enable` invocation will enable the rest
+# We want to enable all systemd units except those that are run by timers, so
+# we have to mark those as --no-enable first, then the final
+# `dh_systemd_enable` invocation will enable the rest.
 override_dh_systemd_enable:
 	dh_systemd_enable --no-enable securedrop-submissions-today.service
 	dh_systemd_enable --no-enable securedrop-clean-tmp.service
+	dh_systemd_enable --no-enable securedrop-remove-pending-sources.service
 	dh_systemd_enable
 
 # This is basically the same as the enable stanza above, just whether the
@@ -75,4 +76,5 @@ override_dh_systemd_enable:
 override_dh_systemd_start:
 	dh_systemd_start --no-start securedrop-submissions-today.service
 	dh_systemd_start --no-start securedrop-clean-tmp.service
+	dh_systemd_start --no-start securedrop-remove-pending-sources.service
 	dh_systemd_start
--- a/securedrop/debian/app-code/lib/systemd/system/securedrop-remove-pending-sources.timer
+++ b/securedrop/debian/app-code/lib/systemd/system/securedrop-remove-pending-sources.timer
@@ -1,0 +1,9 @@
+[Unit]
+Description=remove pending SecureDrop sources daily
+
+[Timer]
+OnCalendar=daily
+Persistent=true
+
+[Install]
+WantedBy=timers.target


securedrop/debian/app-code/lib/systemd/system/securedrop-remove-pending-sources.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=job to remove pending SecureDrop sources daily
+
+[Service]
+ExecStart=/var/www/securedrop/manage.py remove-pending-sources >/dev/null
```
legoktm: I don't think we should be suppressing output like this. If the goal is to prevent OSSEC notifications (which the PR description alludes to), I don't think it should be necessary, other timers/services generate output but don't trigger notifications?

If it's producing output that shouldn't be logged, then I'd rather have a --quiet flag or something.

securedrop/debian/rules
```diff
@@ -68,11 +68,13 @@ override_dh_installinit:
 override_dh_systemd_enable:
```
legoktm, line 68: Unfortunately I wrote "two" in the comment above, can we tweak that to not use a specific number?

"...except those that are run by timers"?

nathandyer:
Completed first portion of test plan this afternoon:

* [x] Installed `securedrop-app-code` from this branch on a production SD install
* [x] Generated more than 100 "pending" sources
* [x] Confirmed that 110 pending sources were created on the server

Going to allow the nighty timers to fire overnight and will return to complete the remainder of the test plan tomorrow.

securedrop/debian/rules
cfm: Fixed in 6f780c6.

securedrop/debian/app-code/lib/systemd/system/securedrop-remove-pending-sources.service
cfm: You know what: we can retitle this pull request "In which @cfm writes a systemd timer exactly _as though_ it is a cron job".  :-P

The suppression was just for OSSEC, so you're right that it's superfluous as well as non-idiomatic.  Removed in 6f780c6.

cfm:
Thanks, @nathandyer!  If you have time, you should be able to rebuild from 6f780c6 and shove it into your system without any other changes or extra testing steps, mainly for <https://github.com/freedomofpress/securedrop/pull/6826#discussion_r1218515916>.  But it's such a trivial diff I wouldn't worry about it if it's not convenient for your testing.  :-)

securedrop/debian/app-code/lib/systemd/system/securedrop-remove-pending-sources.service
legoktm: Haha, lgtm, leaving the rest of the review for Nathan :)

nathandyer:
@cfm Not a problem at all to test with the latest changes from 6f780c6! 

I am running into something a bit odd, and I think it's just from me not setting things up correctly somehow. From my Admin Workstation, I cloned the SecureDrop git repo and checked out the [stg-6788-remove-pending-sources-nightly](https://github.com/freedomofpress/securedrop/tree/stg-6788-remove-pending-sources-nightly) branch. From there, I did the standard procedure, including a `./securedrop-admin install`. 

The odd thing is, when I look on my app server itself, `manage.py` does not have the `remove-pending-sources` option, and the systemd service to enable the daily run isn't present either.

Is there some other method I should be using to get the newer `securedrop-app-code` onto my server?

cfm:
Discussed
<https://github.com/freedomofpress/securedrop/pull/6826#issuecomment-1578931028>
out of band with @nathandyer and clarified the test plan to reflect both
<https://github.com/freedomofpress/securedrop/pull/6826#discussion_r1218565000>
and the need to install a securedrop-app-code package built via "make
build-debs" on this branch.

nathandyer:
Thank you @cfm! Happy to report I successfully built the Debian packages and installed them on my server. Verified that `manage.py` is now the latest version with the `remove-pending-sources` option, and that the systemd timer is installed and active.

I should have definitive results first thing in the morning.

nathandyer (approved):
Happy to report that overnight the timer fired as expected and removed 10 of the pending sources, leaving only 100 on the server.

* [x] Confirm that only 100 "pending" sources remain on the server.

Using `systemctl` I can see the timer is still active and running, and will be triggered again tonight.

This gets my approval, although since I'm not a maintainer we'll still need a final official "Approve" action.

eaon:
Thanks Nathan! Merging

Issue #6500, opened 2022-07-21
Determine plan for auditing Rust dependencies


legoktm:
As part of using Sequoia, we'll start pulling in Rust dependencies (161 crates by my current count!). The process for performing "diff reviews" of Python dependencies is [documented on the wiki](https://github.com/freedomofpress/securedrop/wiki/Dependency-specification-and-update-policies).

We may want to take advantage of some Rust tools that allow for sharing audit status like:
* https://github.com/crev-dev/cargo-crev
* https://github.com/mozilla/cargo-vet (see LWN coverage: https://lwn.net/Articles/897435/)

legoktm:
I reviewed both the cargo-crev and cargo-vet documentation, and while it seems less mature, my preference is for cargo-vet.

cargo-crev is designed for individuals doing reviews, each person gets a crev ID and hosts their own crev proof repository. While cargo-vet seems more aligned with trusting orgs for reviews.

cargo-crev is built on the web of trust, so when you trust person, you now implicitly trust all the other people they trust. I think that's fine for casual projects, but I think we're a lot more paranoid than that! I imagine we would trust the audits done by Mozilla/Firefox since if we don't trust them we're already pwned, but not necessarily every person who someone who works on Firefox trusts.

And ultimately cargo-vet seems conceptually simpler, there's no cryptography involved or keys or IDs, it just relies on git repos (so HTTPS).

legoktm:
I tried out `cargo vet` today, it seems to work pretty nicely. The main "issue" I ran into was that we are deploying to a very specific platform ("target" in Rust lingo), Linux x86_64. There are a bunch of Windows/WASM/Redox dependencies in the vet list that we don't actually use that I filtered out manually based on https://github.com/mozilla/cargo-vet/issues/63#issuecomment-1382836458. That ticket has some suggestions on how upstream can implement some form of target limiting.

legoktm:
Thinking closely about the Python diff review workflow, there are two steps missing:
* We don't verify that the version we review matches checksum of what is included in Cargo.lock. This probably needs to be fixed upstream.
* There's no opportunity to PGP sign the review. If we consider this to be a requirement, I would suggest we just mandate GPG-signing of commits that add new audits.

legoktm:
I think we're mostly set on going forward with `cargo vet`, but FWIW Google has also adopted it: https://opensource.googleblog.com/2023/05/open-sourcing-our-rust-crate-audits.html

cfm:
[...]

Issue #6399, opened 2022-04-13
Use Sequoia-PGP instead of gpg/pretty_bad_protocol


legoktm:
## Status

A branch, [`oxidize`](https://github.com/freedomofpress/securedrop/tree/oxidize), contains the current state of progress.

## Original description

We should consider using https://sequoia-pgp.org/ for SecureDrop's encryption instead of our current setup of the unmaintained pretty_bad_privacy library shelling out to gpg.

Recent issues with gpg like https://github.com/freedomofpress/securedrop/issues/6389 indicate that we really don't need most of gpg's functionality and our reliance on an unmaintained library that the author discourages people from using (as implied in the name). Using Sequoia as a library should allow us to use only what we actually need to use. We would also no longer be tied to the gpg on-disk keyring, so we could put them in the database, avoiding issues like https://github.com/freedomofpress/securedrop/pull/6270 (we'd need some migration process of course).

According to <https://sequoia-pgp.org/status/>, no external audit of the code has been done yet.

We will also want to look into whether we can reuse something like [johnnycanencrypt](https://github.com/kushaldas/johnnycanencrypt) or whether we should just roll our own [pyO3](https://pyo3.rs/) bindings.

## Subtasks

* [x] #6499
* [x] Write code (including tests for all of the below):
   * [x] #6799
   * [x] #6800
   * [x] #6802
   * [x] #6807
   * [x] **BLOCKER:** cumulative dependency/diff review, including #6500
* [x] #6801
* [x] #6806
* [x] Rust integration
   * [x] #6814
   * [x] #6816
   * [x] #6817
   * [x] #6815

legoktm:
My current thinking is that we use pyO3 to write a very slim bridge to sequoia that just implements the bare functionality of what we need.

We might want to maintain this bridge in a separate repository, build it as a wheel and treat it as a dependency just so we don't need to add the Rust toolchain and build step into the dev container, `make build-debs`, etc. It would make integration testing annoying though since you have to keep manually building and installing the wheel.

Or we just bite the bullet and integrate all the Rust stuff in properly, which, as a side-benefit lowers the bar for the next person trying to introduce Rust things :)

legoktm:
> We would also no longer be tied to the gpg on-disk keyring, so we could put them in the database

[...]

PR #6828, opened 2023-05-30
Add "redwood" Sequoia Rust/Python bridge 

### LINKED ISSUE #6814: CI runs Rust-specific lint/test jobs
We are introducing Rust code for the first time as part of the Sequoia work and need to set up CI checks. For the first step we should:

* Run `cargo clippy`, with warnings causing failures
* Run `cargo test`
* Run `maturin build`, so we know that it builds/compiles with pyO3.

### LINKED ISSUE #6816: Development environment builds Rust code (optionally with live-reload)
Once the Sequoia migration introduces Rust code, we need to pre-compile that code and install it in the venv before launching the dev flask servers. This has been done in the oxidize branch already.

It would be nice if live-reload worked so that changing a Rust file would recompile the Rust code, install the wheel, and then trigger reload of the flask dev servers. But I could easily see this being more complicated than its worth and also I personally rarely use live reload functionality when developing my own Rust projects.

**Blocked by:**
* #6784

legoktm:
## Status

Ready for review, but depends on #6832

## Description of Changes

See individual commit messages.

Refs https://github.com/freedomofpress/securedrop/issues/6814.
Fixes https://github.com/freedomofpress/securedrop/issues/6814.
Fixes https://github.com/freedomofpress/securedrop/issues/6816.

## Testing

* [x] Visual review of Rust code
* [x] CI passes
* [x] Bringing up a development environment (`make dev`) works
* [x] Bringing up a development environment a second time doesn't recompile everything from scratch

## Deployment

Any special considerations for deployment? This PR merely introduces Rust code and builds it for the development environment, it should have no impact on staging/prod installs yet.

## Checklist

- [x] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-docs) for these changes, or will do so later - will do so later once more of the Rust stuff is in place.

Since this includes Rust code, we're going to defer the review to #6500.


Makefile
```diff
@@ -119,6 +123,12 @@ html-lint:  ## Validate HTML in web application template files.
 		securedrop/source_templates/*.html securedrop/journalist_templates/*.html
 	@echo
 
+.PHONY: rust-lint
+rust-lint: ## Lint Rust code
```
cfm, line 127: Is there any reason not to include this in `lint` below?

Makefile
```diff
@@ -282,6 +292,11 @@ test:  ## Run the test suite in a Docker container.
 .PHONY: test-focal
 test-focal:  test
 
+.PHONY: rust-test
+rust-test:
```
cfm, line 297: Is there any reason not to include this in `test` above?

redwood/redwood.pyi
```diff
@@ -0,0 +1,16 @@
+# type stub for redwood module
+# see https://pyo3.rs/v0.16.4/python_typing_hints.html
+from pathlib import Path
+from typing import List
+
+def generate_source_key_pair(passphrase: str, email: str) -> (str, str, str):
+    pass
```
cfm: @legoktm, what is your opinion on `pass` versus `...`?

securedrop/dockerfiles/focal/python3/SlimDockerfile
```diff
@@ -9,16 +9,34 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get install
...
+ENV RUSTUP_INIT_SHA256 3dc5ef50861ee18657f9db2eeb7392f9c2a6c95c90ab41e45ab4ca71476b4338
+ENV RUSTUP_HOME /opt/rustup
+ENV CARGO_HOME /opt/cargo
+
+RUN TMPDIR=`mktemp -d` && cd ${TMPDIR} \
+        && curl --proto '=https' --tlsv1.2 -OO -sSf https://static.rust-lang.org/rustup/archive/${RUSTUP_VERSION}/x86_64-unknown-linux-gnu/rustup-init \
+        && echo "${RUSTUP_INIT_SHA256} *rustup-init" | sha256sum -c - \
+        && chmod +x rustup-init \
+        && ./rustup-init --default-toolchain=${RUST_VERSION} --profile minimal -y \
+        && cd && rm -rf ${TMPDIR}
```
cfm, line 29: I know this comes from freedomofpress/securedrop-security#70, but I would love an explanatory comment for posterity here and in `builder/Dockerfile` if you're willing to add one.  (I'll live with the short flags. ;-)

securedrop/bin/dev-deps
```diff
@@ -156,3 +156,15 @@ function reset_demo() {
         ./loaddata.py
     fi
 }
+
+function build_redwood() {
+    # Create a debug build of redwood and install it in the app-code virtualenv
+    # TODO: add live reload support
+    # We tell cargo to put its registry/crate cache and build cache in the target/ folder so
+    # it persists instead of being recreated from scratch each time the container starts.
+    # n.b. we can't re-use the host's caches because of permission differences
```
cfm, line 165: This is great to have documented.  Do you plan to implement the `TODO` on this branch, or would you like to defer it to a follow-up ticket?

redwood/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+    Io(#[from] io::Error),
+    #[error("Unexpected non-UTF-8 text: {0}")]
+    NotUnicode(#[from] FromUtf8Error),
+}
+
+create_exception!(redwood, RedwoodError, PyException);
+
+impl From<Error> for PyErr {
+    fn from(original: Error) -> Self {
+        // TODO: make sure we're not losing the stacktrace here
```
cfm: Flagging `TODO`.

redwood/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+        for key in cert
+            .keys()
+            .with_policy(p, None)
+            .supported()
+            .alive()
+            .revoked(false)
+        {
+            recipient_keys.push(key);
+        }
+    }
```
cfm: Could you briefly document this progression?

redwood/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    let mut sink = File::create(destination)?;
+    let message = Message::new(&mut sink);
+    let message = Armorer::new(message).build()?;
+    let message = Encryptor::for_recipients(message, recipient_keys).build()?;
+    let mut message = LiteralWriter::new(message).build()?;
+
+    io::copy(&mut plaintext, &mut message)?;
+
+    message.finalize()?;
```
cfm: This is *so* elegant, *and*: could you briefly document it?  :-)

redwood/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    Ok(())
+}
+
+/// Decrypt the given ciphertext using the specified private key that is locked
+/// using the provided passphrase. It is assumed that the plaintext is UTF-8.
```
cfm: Nit for clarity: Are "given", "specified", and "provided" interchangeable in this comment?  If so, I would suggest something like:

```suggestion
/// Given a ciphertext, private key, and passphrase, unlock the private key with
/// the passphrase, and use it to decrypt the ciphertext.  It is assumed that the
/// plaintext is UTF-8.
```

redwood/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+        assert_eq!(format!("{}", cert.fingerprint()), fingerprint);
+        println!("{}", secret_key);
+        assert!(secret_key.starts_with("-----BEGIN PGP PRIVATE KEY BLOCK-----"));
+        assert!(secret_key.contains("Comment: Source Key <foo@example.org>"));
+        let cert = Cert::from_str(&secret_key).unwrap();
+        assert_eq!(format!("{}", cert.fingerprint()), fingerprint);
+    }
+
+    #[test]
+    fn test_encryption_decryption() {
```
cfm, line 219: Beautiful.

redwood/src/decryption.rs
```diff
@@ -0,0 +1,67 @@
...
+    pub(crate) secret: &'a sequoia_openpgp::Cert,
+    pub(crate) passphrase: &'a Password,
+}
+
+impl<'a> VerificationHelper for Helper<'a> {
+    fn get_certs(
+        &mut self,
+        _ids: &[sequoia_openpgp::KeyHandle],
+    ) -> sequoia_openpgp::Result<Vec<sequoia_openpgp::Cert>> {
+        // Return public keys for signature verification here.
```
cfm: Copypasta?

redwood/src/decryption.rs
```diff
@@ -0,0 +1,67 @@
...
+        Ok(Vec::new())
+    }
+
+    fn check(
+        &mut self,
+        _structure: MessageStructure,
+    ) -> sequoia_openpgp::Result<()> {
+        // Implement your signature verification policy here.
```
cfm: Copypasta?

redwood/src/decryption.rs
```diff
@@ -0,0 +1,67 @@
...
+        D: FnMut(SymmetricAlgorithm, &SessionKey) -> bool,
+    {
+        // The encryption key is the first and only subkey.
+        let key = self
+            .secret
+            .keys()
+            .secret()
+            .with_policy(self.policy, None)
+            .next()
+            // FIXME: unwrap()
```
cfm: Flagging `FIXME`.

redwood/src/decryption.rs
```diff
@@ -0,0 +1,67 @@
...
+        // Decrypt the secret key with the specified passphrase.
+        let mut pair = key.decrypt_secret(self.passphrase)?.into_keypair()?;
+
+        pkesks[0]
+            .decrypt(&mut pair, sym_algo)
+            .map(|(algo, session_key)| decrypt(algo, &session_key));
+
+        // XXX: In production code, return the Fingerprint of the
+        // recipient's Cert here
+        Ok(None)
```
cfm: Flagging `XXX`.

cfm (changes requested):
Test plan checks out!  Please take (or leave :-) the inline comments in the spirit of an initial read-through.

I'll review #6832 tomorrow, which will let you rebase this from `develop` along with any changes you make in response to the review.

Makefile
legoktm: Two considerations:
1. We don't currently require a Rust toolchain to run `make lint/test`
2. I would estimate that 95% of commits (currently) won't touch Rust code, so running it on every lint/test seems excessive and if you don't have a hot cache, slow.

I expect both of those to degrade over time (the first is especially weak already) that it's probably worth doing eventually, it just seems like it would be a bit annoying right now.

redwood/redwood.pyi
legoktm: I didn't realize that `...` was an option, but it seems like the correct thing per https://typing.readthedocs.io/en/latest/source/stubs.html#functions-and-methods

> Using a function or method body other than the ellipsis literal is currently unspecified. Stub authors may experiment with other bodies, but it is up to individual type checkers how to interpret them:
> ```python
> def foo(): ...  # compatible
> def bar(): pass  # behavior undefined
> ```

Changed.

securedrop/bin/dev-deps
legoktm: I would prefer to defer live-reload to a different PR/ticket - I think using something like https://lib.rs/crates/cargo-watch is probably the way to go, but there are ramifications like making container builds much slower if we build from source, etc. that it needs a bit more thinking on what actually provides a better DX.

redwood/src/lib.rs
legoktm: Done, and I added function-level docs that point to https://gitlab.com/sequoia-pgp/sequoia/-/blob/main/guide/src/chapter_02.md which is basically where this pattern comes from.

redwood/src/lib.rs
legoktm: Done, hopefully that's clear!

redwood/src/lib.rs
legoktm: Yes, ty, applied.

redwood/src/decryption.rs
legoktm: Yes, fixed the comment to explain why it's not implemented.

redwood/src/decryption.rs
legoktm: Yes, fixed the comment to explain why it's not implemented.

redwood/src/decryption.rs
legoktm: Fixed by throwing an error (instead of panic!-ing) and added a comment on why it shouldn't be reachable

redwood/src/decryption.rs
legoktm: I've expanded the XXX but mostly I think it's a question for the Sequoia team on whether we need to implement the return value or not.

securedrop/dockerfiles/focal/python3/SlimDockerfile
legoktm: I've added more documentation in https://github.com/freedomofpress/securedrop/pull/6832/commits/5cec5b54290ed7bb59b09d176234b15c2b46f06a - if that looks good to you I can copy it back here.

redwood/src/lib.rs
legoktm: I think the answer is that we're not losing the stacktrace because it never had it in the first place :/ https://docs.rs/anyhow/latest/anyhow/ describes that we need to enable both the `backtrace` feature and set an environment variable.

Based on https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations we should at least use the Debug mode: `{:?}` so we're not dropping context - done.

legoktm: pyo3 now supports a built-in conversion from anyhow (https://github.com/PyO3/pyo3/blob/342e4ca678834736779e6ada4d60354be7bd4b1d/src/conversions/anyhow.rs#L114), I'm trying to see if that's enough to drop our `Error` type entirely and just go from anyhow::Error -> RedwoodError...

legoktm: OK, it doesn't seem possible, the built-in conversion is hardcoded to only throw `RuntimeError` and I think having a dedicated `RedwoodError` (in Python) is important here so I'll leave it as is.

Makefile
cfm: Makes sense.  Thanks!

securedrop/dockerfiles/focal/python3/SlimDockerfile
cfm: That's great.  Thank you!

cfm (changes requested):
Thanks for quick turnaround on those changes, @legoktm!  From my perspective this is ready to go.  However, it doesn't rebase cleanly from `develop` after #6832.  Can you rebase, and then I'll do a final review before merge?

securedrop/dockerfiles/focal/python3/SlimDockerfile
legoktm: Copied over!

legoktm:
Rebased!

cfm (approved):
This looks terrific, @legoktm.  It's a joy to see what these functions look like in Rust!  Thank you for leading this effort.

I'll merge as soon as CI is green.

legoktm:
Oh actually, I'm wondering if we need to do a diff review for maturin now? Like, it's mostly Rust but there is a bit of Python in it that I assume we should peek at?

legoktm:
But also it's kind of useless to review the Python parts without the Rust parts, so I'm also ok with deferring it to https://github.com/freedomofpress/securedrop/issues/6500, with that being a blocker for actual usage of the code.

cfm:
I agree, @legoktm.  I've also updated #6399 to emphasize that blocking status for production acceptance and release.

PR #6490, opened 2022-07-18
Move reply key generation to /submit endpoint

zenmonkeykstop:
## Status

Ready for review, but flagged as draft to block merge until `release/2.5.0` branch is created

## Description of Changes

Fixes #6489 

Reply key generation is moved in the SI from `/lookup` to `/submit`. If a key is not already present for the given source it will be created during the first submisstion.

## Testing
- Check out this branch and run `make dev`
- [ ] in another terminal, connect to the dev container with `docker exec -it securedrop-dev-0 bash` and verify that only the default dev source reply keys exist with `gpg --homedir /var/lib/securedrop/keys -k`
- Create a new source via the SI, but do not submit anything
- [ ] verify that no new key was created, with `gpg --homedir /var/lib/securedrop/keys -k`
- [ ] submit a message as the source and verify that a reply key was created
- [ ] log in to the JI and verify that the source is listed and a reply can be sent
- [ ] on the SI, verify that the reply is displayed correctly
- [ ] submit another message and verify that a new reply key was *not* created
- [ ] on the JI, verify that another reply can be submitted.
- [ ] log out of the SI, log back in as the same source, and verify that both replies can be read.

- [ ] For extra points, try to break reply key generation by doing multiple simultaneous file submissions.

## Deployment

n/a

### If you made changes to the server application code:

- [x] Linting (`make lint`) and tests (`make test`) pass in the development container

### If you made non-trivial code changes:

- [x] I have written a test plan and validated it for this PR

Choose one of the following:

- [ ] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-docs) for these changes, or will do so later
- [ ] I would appreciate help with the documentation
- [x] These changes do not require documentation

PR #6829, opened 2023-05-31
return reply-key generation to "/lookup" (revert #6490)

### LINKED ISSUE #6489: Generate reply key on first submission
Note that this change is being reverted!

## Description

When a source visits /lookup for the first time, a GPG reply key is generated. However, if the source does not submit a message, it is impossible to reply to them, as their account is flagged as `pending` and not displayed in the JI. To avoid creating keys unnecessarily, the key could be created on submission instead if it does not already exist.

## User Research Evidence

Observed behavior of sd instances

## User Stories

As an admin, I want to keep a manageable number of source accounts and reply keys

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Fixes #6489 .

This PR reverts the changes in #6490, which moved source reply key generation to after the first submission, thereby avoiding an accumulation of pending sources (and keys). Since we have an option to manually and soon automatically purge stale pending sources,  this is no longer required and can be removed to reduce scope/potential for subtle errors due to the change in key generation.

## Testing

- [ ] CI is passing
- [ ] on the SI, when a source is created, a reply key is generated before their first submission
- [ ] Replies work correctly (can be sent, read, deleted in  SI and removed in JI)

## Deployment

2. New installs.

Choose one of the following:

- `admin/requirements.in`

(remove) this section.

Choose one of the following:

### DIFF
--- a/securedrop/journalist_app/utils.py
+++ b/securedrop/journalist_app/utils.py
@@ -6,7 +6,7 @@ from typing import List, Optional, Union
 import flask
 import werkzeug
 from db import db
-from encryption import EncryptionManager, GpgKeyNotFoundError
+from encryption import EncryptionManager
 from flask import Markup, abort, current_app, escape, flash, redirect, send_file, url_for
 from flask_babel import gettext, ngettext
 from journalist_app.sessions import session
@@ -397,11 +397,8 @@ def delete_collection(filesystem_id: str) -> None:
     if os.path.exists(path):
         Storage.get_default().move_to_shredder(path)
 
-    # Delete the source's reply keypair, if it exists
-    try:
-        EncryptionManager.get_default().delete_source_key_pair(filesystem_id)
-    except GpgKeyNotFoundError:
-        pass
+    # Delete the source's reply keypair
+    EncryptionManager.get_default().delete_source_key_pair(filesystem_id)
 
     # Delete their entry in the db
     source = get_source(filesystem_id, include_deleted=True)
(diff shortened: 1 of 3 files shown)


cfm:
@zenmonkeykstop, since this is a strict reversion to known-good code and tests, do we have reason to think interactive testing is necessary?

zenmonkeykstop:
> @zenmonkeykstop, since this is a strict reversion to known-good code and tests, do we have reason to think interactive testing is necessary?

Given the nature of the change, I erred on the side of over-testing, but I think you're right, any breakage would be caught by automated testing and would definitely come up in the course of release QA. If you're comfortable with visual review of the code then so am I.

cfm (approved):
Visual review, sure, *and* we can prove that merging this pull request will be the inverse operation of merging #6490, regardless of the intervening history:

```sh-session
user@sd-dev:~/securedrop$ git log --oneline --no-show-signature revert-6490 | head -n 3
e7063e96c Revert "moved reply key generation to /submit, to avoid generating keys for pending users"
582a87c06 Revert "updated tests to verify reply key creation"
d964e5984 Merge pull request #6825 from freedomofpress/stg-6824-thats-not-a-word
user@sd-dev:~/securedrop$ git checkout -b revert-6490-cfm d964e598
Switched to a new branch 'revert-6490-cfm'
user@sd-dev:~/securedrop$ git show --oneline --no-show-signature e7909b7
e7909b77e Merge pull request #6490 from freedomofpress/delay-reply-key
user@sd-dev:~/securedrop$ git revert -m 1 e7909b7
user@sd-dev:~/securedrop$ git diff revert-6490 | wc -l
0
```

Issue #6817, opened 2023-05-17
Packaging process builds Rust code


legoktm:
Our `make build-debs` packaging process needs to build a Rust wheel and install it into the virtualenv. As part of this we need to make sure that the built wheel is fully reproducible so we aren't regressing on reproducible builds.

legoktm:
Just noting that in my dev Qubes VM with 6 CPU cores assigned, it takes a little over 4 minutes to build redwood in release mode.

legoktm:
So here's what maturin is doing under the hood, as discovered by `RUST_LOG=debug maturin build -m redwood/Cargo.toml -v -r`

```
2023-06-28T20:05:49.562717Z DEBUG maturin::project_layout: Using cargo manifest path from command line argument: "/home/user/github/freedomofpress/securedrop/redwood/Cargo.toml"
2023-06-28T20:05:49.562741Z DEBUG maturin::project_layout: Resolving cargo metadata from "/home/user/github/freedomofpress/securedrop/redwood/Cargo.toml"
2023-06-28T20:05:49.680016Z DEBUG maturin::project_layout: Found pyproject.toml at "/home/user/github/freedomofpress/securedrop/redwood/pyproject.toml"
2023-06-28T20:05:49.680486Z DEBUG maturin::project_layout: Resolving cargo metadata from "/home/user/github/freedomofpress/securedrop/redwood/Cargo.toml"
2023-06-28T20:05:49.798743Z DEBUG maturin::project_layout: Project layout resolved project_root=/home/user/github/freedomofpress/securedrop/redwood python_dir=/home/user/github/freedomofpress/securedrop/redwood rust_module=/home/user/github/freedomofpress/securedrop/redwood/redwood python_module=/home/user/github/freedomofpress/securedrop/redwood/redwood extension_name=redwood module_name=redwood
🔗 Found pyo3 bindings
2023-06-28T20:05:49.892508Z DEBUG maturin::python_interpreter: Found CPython interpreter at /home/user/github/freedomofpress/securedrop/.venv/bin/python3
🐍 Found CPython 3.8 at /home/user/github/freedomofpress/securedrop/.venv/bin/python3
📡 Using build options compatibility from pyproject.toml
2023-06-28T20:05:49.892782Z DEBUG maturin::compile: Setting PYO3_PYTHON to /home/user/github/freedomofpress/securedrop/.venv/bin/python3
[...]

PR #6832, opened 2023-05-31
Update Rust in builder to 1.69.0

legoktm:
## Status

Ready for review

## Description of Changes

First, adjust how Rust is installed in the builder container
    
Install Rust and rustup in slightly more straightforward directories and
unconditionally add it to $PATH instead of weirdly in /root/.cargo.

This setup is partially based on <https://github.com/rust-lang/rustup/issues/1085#issuecomment-296604244>,
which is described as unsupported for multiple users but our containers
are single-user.

The plan is to copy this to the development container as well.

Second, and primarily, update Rust in builder to 1.69.0
    
For the Sequoia work, we're planning to use 1.69.0 so we need to update
the builder to match. This is being done separately since it already
affects a production component (the cryptography wheel) and should be
tested independently.

This also lets us get rid of our index-warming hack since we can use the
new fast sparse protocol instead
(<https://blog.rust-lang.org/2023/03/09/Rust-1.68.0.html#cargos-sparse-protocol>).

Refs #6817.

## Testing

* [x] CI passes, including deb building and staging build
* [x] Build new debs, deploy on a staging/prod instance:
   * [x] Create new journalist, log in as them.

(AFAIK that's the main place we use cryptography on the server)

## Deployment

## Checklist

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -41,7 +41,7 @@ COPY dh-virtualenv.pref /etc/apt/preferences.d/
 
 RUN apt-get update && apt-get install -y dh-virtualenv
 
-ENV RUST_VERSION 1.58.1
+ENV RUST_VERSION 1.69.0
 ENV RUSTUP_VERSION 1.24.3
 ENV RUSTUP_INIT_SHA256 3dc5ef50861ee18657f9db2eeb7392f9c2a6c95c90ab41e45ab4ca71476b4338
 ENV RUSTUP_HOME /opt/rustup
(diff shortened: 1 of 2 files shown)


cfm (approved):
Thanks for proactively making changes from <https://github.com/freedomofpress/securedrop/pull/6828#discussion_r1218674519>.  :-)

> * [x] CI passes, including deb building and staging build
> * [x] Build new debs, deploy on a staging/prod instance:
>    * [x] Create new journalist, log in as them.
> 
> (AFAIK that's the main place we use cryptography on the server)

I [spot](https://github.com/search?q=repo%3Afreedomofpress%2Fsecuredrop+cryptography&type=code) a few more places.  But a full smoke-test of the Source and Journalist Interfaces checks out just fine!

PR #6836, opened 2023-06-06
Vendor pretty_bad_protocol

legoktm:
## Status

Ready for review

## Description of Changes

Vendor pretty_bad_protocol and apply minimal fixes to get it to pass CI. Then remove our monkey patching from EncryptionManager.

Each commit should have a description of what is happening and why.

Fixes #6807.

## Testing

How should the reviewer test this PR?

* [x] CI passes
* [x] Basic smoke test in dev environment, submit a message/file, and then decrypt it as a journalist
* [x] Basic smoke test in a staging/prod environment (to verify apparmor, etc.), submit a message/file, and then decrypt it as a journalist

## Deployment

Any special considerations for deployment? Just the standard-ish ones.

## Checklist

- [x] Linting (`make lint`) and tests (`make test`) pass in the development container
- [x] I have updated AppArmor rules to include the change
- [ ] I have written a test plan and validated it for this PR
- [x] These changes do not require documentation

PR #6883, opened 2023-06-26
testinfra: Improve printing of apparmor denials

legoktm:
## Status

Ready for review
## Description of Changes

Dumping all of syslog is huge and often what you're looking for will get truncated by pytest. Instead, look for individual lines with the DENIED substring and assert those don't exist. This will print just those lines, giving us a better chance at seeing the relevant lines.

## Testing

I think just visual review is fine? I used this when developing #6836 to get staging-via-CI to print out the right apparmor denial line.

## Deployment


zenmonkeykstop (approved):
LGTM based on visual review

PR #6884, opened 2023-06-27
Build Rust redwood wheel during packaging process

### LINKED ISSUE #6538: Drop dh-virtualenv
## Description

We currently use `dh-virtualenv` to create a virtualenv that we install our dependencies in. It has some nice features, and some downsides. I believe that we've reached the point that the downsides outweigh the features, which are simple to reimplement ourselves. This is a proposal for all the SDW components too, I just didn't want to file two identical tickets and split discussion.

### Features
* Creates virtualenv in `/opt/venvs/...`, installs dependencies and our code
* Edits the virtualenv so it can properly be relocated (artifact of Debian packaging, it's built in `./debian/securedrop-app-code` and then moved to `opt/venvs`)
* Automatically rebuild the virtualenv when the underlying Python version changes with triggers (this "feature" caused issues for us,  #6230)

### Downsides
* Not even included in the focal release, we're currently pulling it in via jammy
* inactive Debian maintainer, I've had to fix it ~~twice~~ thrice already so it stays in testing
* issues with debhelper-compat=12 - https://github.com/spotify/dh-virtualenv/issues/326, https://github.com/freedomofpress/securedrop-debian-packaging/issues/335
* the "triggers" anti-feature
* cannot use `--require-hashes`: https://github.com/freedomofpress/securedrop-debian-packaging/issues/79
* hides deprecation and error messages, see https://github.com/freedomofpress/securedrop/issues/6538#issuecomment-1430610626.
* https://github.com/freedomofpress/securedrop-builder/issues/430

## Proposal
* Create the virtualenv ourselves, with `python3 -m venv ...`
* Install the dependencies ourselves, with `pip install ...`
* Fix up the virtualenv ourselves, with a one-line sed command (n.b. in testing this, I discovered their sed command isn't fully comprehensive as mine was, so this will help with reproducibility)

I think this will also demystify exactly what is happening at build time since we can see the underlying pip/virtualenv commands.

### LINKED ISSUE #6817: Packaging process builds Rust code

legoktm:
## Status

Ready for review

## Description of Changes

* Properly exclude custom_logo.png from package and fix test (mostly unrelated, but discovered while working on this so I'm throwing it in...)
* Drop dh-virtualenv from packaging process 
* Build redwood wheel without maturin
* Build Rust redwood wheel during packaging process

Fixes https://github.com/freedomofpress/securedrop/issues/6538.
Fixes https://github.com/freedomofpress/securedrop/issues/6817.

## Testing

Largely focusing on the dh-virtualenv change.

* [x] `cp securedrop/static/i/logo.png securedrop/static/i/custom_logo.png` (this should still be excluded from the build)
* [x] Check out this PR and run `make build-debs`
* [x] Download `https://apt.freedom.press/pool/main/s/securedrop/securedrop-app-code_2.6.0%2Bfocal_amd64.deb`, then run `debdiff` between the two files, you should see https://gist.github.com/legoktm/170f170e01437ba49671306df4425912
  * One intentional change is that `/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/securedrop_app_code-2.6.0-py3.8.egg-info/` is gone, we're no longer installing an empty securedrop-app-code Python package because dh_virtualenv mandates it.

For the Rust stuff:
* [x] deb-tests CI passes, which means the redwood.so is included in the package
* [x] staging CI passes, which means the redwood wheel is installed and usable for generating a key pair
* [x] review the `make build-debs` log from earlier, verify there's no garbage output (escape codes or progress bars) caused by Rust/cargo stuff

Verify Rust reproducibility:
* [x] Copy the securedrop-app-code.deb from your first `make build-debs` to something like `first.deb`
* [x] Re-run `make build-debs`, save the securedrop-app-code deb as something like `second.deb`
* [x] For each deb, extract the files. `ar x first.deb`, then `tar xvf data.tar.xz`
* [x] For each extracted deb, run `sha256sum opt/venvs/securedrop-app-code/lib/python3.8/site-packages/redwood/redwood.cpython-38-x86_64-linux-gnu.so` - the checksum should be `152d90961bf1cf44fcbc5c0275e31c68ca6a16f1b0c42895fe9272a53e2f1685` for both.

## Deployment

Any special considerations for deployment? Just the packaging changes.

## Checklist


legoktm:
Leaving as WIP because we need to figure out whether we want to re-use upstream maturin wheels, build it from source, or what. And whether we need to audit dependencies of that too.

legoktm:
I realized the build wasn't reproducible because we weren't building with the lockfile and using `--locked`. That's now part of the test plan.

redwood/build-wheel.py
```diff
@@ -0,0 +1,58 @@
...
+    print(f"Error: Can't find libredwood.so (looked for {libredwood})")
+    sys.exit(1)
+
+so_destination = args.redwood / "redwood" / ("redwood" + sysconfig.get_config_var("EXT_SUFFIX"))
+shutil.copy2(libredwood, so_destination)
+
+print("Building wheel")
+subprocess.check_call([sys.executable, "-m", "pip", "wheel", "."])
+# Yes the name is wrong, but it doesn't matter, this is only an intermediate step anyways
+whl = Path("redwood-0.1.0-py3-none-any.whl")
```
cfm: Can you clarify this comment?  I can _guess_ which name you're referring to, how it's wrong, and why it doesn't matter; but I can't actually tell for sure from the comment.  :-)

cfm (changes requested):
This looks great, Kunal.  In addition to the test plan, I've left one request inline, plus:

- [x] `make dev` succeeds, including building libredwood via your scripting rather than maturin.
- [ ] This branch isn't fully signed.
- Noting for the record: 8f884ea0d3785cdd52cdbd55474f3a255f5a7e17 removes `build-requirements.{in,txt}`, because they were added in #6828 exclusively to pull in `maturin`.

redwood/build-wheel.py
legoktm: Done!

cfm (approved):
Thanks, @legoktm!

PR #6885, opened 2023-07-05
Use ruff to replace flake8, pylint, isort and add more checks

legoktm:
## Status

Ready for review

## Description of Changes

Refs https://github.com/freedomofpress/securedrop-tooling/issues/10

Ruff is an incredibly fast Python linter (implemented in Rust). Out of the gate it can immediately replace flake8, isort and pyupgrade, and covers most (but not all) of pylint. And unlike flake8, it has the capability to auto-fix various issues (my main frustration with ruff has been that it does in less than a second what I used to spend hours doing manually!!).

Aside from being very fast, it's also really nice to have a very easy to configure tool (pyproject.toml) with well documented rulesets and rules (https://beta.ruff.rs/docs/rules/). We never set up any flake8 plugins but looking at the ones included in ruff, we probably should have! So far they've uncovered incorrectly marked-up translatable strings (see the flake-gettext commit) as well as broken tests (flake8-pytest-style and "Fix * test" commits). And there were a number of issues that flake8 should've caught but missed.

There are probably more plugins that we can enable but those can be done as follow-ups given how large this already is.

I did disable a number of suggested changes because I thought they hampered readability (e.g. using contextlib.suppress instead of try/except blocks), I tried to document each one in pyproject.toml and am happy to discuss and re-enable/disable other things that people don't like.

## Testing

* [ ] Visual review for code changes
* [ ] CI passes
* [ ] Team 👍🏾 on use of ruff as a tool

## Deployment

## Checklist


pyproject.toml
```diff
@@ -29,6 +31,16 @@ select = [
 ignore = [
     # incompatible with sqlalchemy https://github.com/astral-sh/ruff/issues/4560
     "E712",
+    # code complexity checks that we fail
+    "PLR0912", "PLR0913", "PLR0915",
```
rocodes, line 44: (Would there be any benefit to configuring them with a different max in the pyproject.toml instead of completely ignoring them? i.e. nudge us towards better future behaviour where we work to get that number down over time?)

admin/tests/test_securedrop-admin-setup.py
```diff
@@ -122,8 +122,8 @@ def test_envsetup_cleanup(self, tmpdir, caplog):
             "subprocess.check_output", side_effect=self.side_effect_venv_bootstrap(venv)
         ):
             bootstrap.envsetup(args, venv)
-            assert not os.path.exists(venv)
-            assert "Cleaning up virtualenv" in caplog.text
+        assert not os.path.exists(venv)
+        assert "Cleaning up virtualenv" in caplog.text
```
rocodes, line 126: oof. thanks, ruff <3

rocodes (approved):
Thanks for preparing this @legoktm.  Stepped through commit by commit and also ran linting locally, looks good to me (and a couple commits were pretty humbling...lookin' at you, 6e7cf93ed639dfac99a1278e3f6fe8d630c0b1d9).

Will hold off on merge so that others can take a quick look if they want, but changes LGTM.

pyproject.toml
legoktm: I'm generally just bearish on automated code complexity checks (e.g. I trust that if you're creating a function that takes 6+ arguments, you have a good reason to do so) which is why I opted to disable entirely rather than raise the limits, but I'm not against that if other people find them useful.

rocodes: IMO those checks are most useful when wrangling different kinds of outside contributors and trying to nudge folks in a certain direction prior to giving manual review feedback, but I'm fine to leave it as is.

Issue #6579, opened 2022-09-29
Prune obsolete dependencies from test-requirements.in


legoktm:
*This is a good first issue for new contributors to take on, if you have any questions, please ask on the task or in our [Gitter room](https://gitter.im/freedomofpress/securedrop)!*

## Description

I'm pretty sure that at least pathlib2 and blinker are no longer required to run our tests.

## Steps to reproduce

* Remove those dependencies from test-requirements.in
* Run `make update-python3-requirements` to regenerate a new test-requirements.txt file
* Run unit tests, see if anything fails. If not, send a PR removing the dependencies. If they do fail, we should document how they're used since it's not obvious with grep.

ryuusama09:
hi @legoktm , i would like to contribute

zenmonkeykstop:
> hi @legoktm , i would like to contribute

Go for it, let us know if you have any questions!

ryuusama09:
> * make update-python3-requirements

this command is not working for me(it is showing make is not an internal command) , instead can i use "pip install --upgrade --force-reinstall -r test-requirements.in"
also , I cannot find  the test-requirements file

legoktm:
@ryuusama09: What operating system are you using? You'll need to install `make` and docker/podman at a minimum to make the development environment work.

>  also , I cannot find the test-requirements file

See https://github.com/freedomofpress/securedrop/tree/develop/securedrop/requirements/python3

MRZHD1:
Hello @legoktm, could I be assigned this?

eaon:
@MRZHD1 assigned to you, as before, feel free to ask questions if you have any!

MRZHD1:
Hey @eaon, sorry for the extremely late start as I was super busy, but after numerous attempts at making the dev environment work I just can't seem to. Every time I run `make dev` I run into this:
```
  ...
  adding 'EasyProcess-0.2.3.dist-info/top_level.txt'
  adding 'EasyProcess-0.2.3.dist-info/RECORD'
  removing build/bdist.linux-x86_64/wheel
  free(): invalid pointer
  ----------------------------------------
  ERROR: Failed building wheel for easyprocess
```


which causes the whole thing to fail. I've already changed the requirements and generated the new test-requirements.txt so I don't know if that's what is causing the issue or not.

Finesse-734:
Hello @legoktm , is this issue open for contribution? if yes, could I be allotted this issue?

zenmonkeykstop:
[...]

PR #6890, opened 2023-07-19
Remove boto and boto3 dev dependencies

eaon:
## Description of Changes

Removes boto and boto3 dev dependencies. The last mention of boto3 was in c9b4fda418190a3061da6203a3d006557ba0e998

## Testing

- [ ] CI is happy

### DIFF
--- a/securedrop/requirements/python3/develop-requirements.in
+++ b/securedrop/requirements/python3/develop-requirements.in
@@ -4,9 +4,6 @@ ansible>=2.9.23,<2.10.0
 argon2_cffi>=20.1.0
 bandit
 black
-# yes, we need both boto and boto3
-boto
-boto3
 cffi>=1.14.2
 cryptography>=41.0.0
 diffoscope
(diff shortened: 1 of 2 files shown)


legoktm:
I wonder if we can drop this stanza from `install_files/ansible-base/roles/tor-hidden-services/tasks/install_tor.yml` too:
```
- name: Dump Tor version to file (for reporting)
  copy:
    dest: "{{ playbook_dir }}/../../.tor_version"
    content: "{{ extract_tor_version.stdout }}"
  delegate_to: localhost
  run_once: true
  become: "no"
  when: "'amazon' in ansible_product_version"
```

legoktm (approved):
LGTM, grepped for `aws` and `amazon` as well with no hits besides what I commented.

Related: #6579.

eaon:
> I wonder if we can drop this stanza from `install_files/ansible-base/roles/tor-hidden-services/tasks/install_tor.yml` too

Looks like that's 6 years old and `.tor_version` isn't referenced anywhere either.

Issue #6799, opened 2023-05-17
New source creation uses Sequoia


legoktm:
As part of our work to migrate to Sequoia, the first step is to have new source creation generate a PGP key pair using Sequoia, and then use those keys for encryption/decryption.

This work will include:
* Writing Sequoia bindings for Python
* Converting key storage to be in the database
* Switching over new source creation to use those functions and store said keys in the database.

Most of this work already exists in the oxidize branch

legoktm:
What is coming in this PR:
* Storing journalist public key out of the GPG keyring - including a hopefully not too hacky postinst migration
* DB schema change, adding 3 columns to the source table
* Changing all the EncryptionManager.encrypt_* functions to use redwood, and sometimes the decrypt function
* Actually switching the new source creation to Sequoia

Honestly I've spent most of the time getting tests to pass, including:
* Getting rid of all `EncryptionManager._gpg` direct calls and fixing all the other `EncryptionManager` calls really.
* Encrypting the journalist test key with a passphrase because Sequoia wants that

Except some tests are failing with `redwood.RedwoodError: OpenPgp(Missing session key: No session key decrypted)` in my new `utils.decrypt_as_journalist()` function. I'm trying to get everything else passing before I dive into this one.

legoktm:
The missing session key issue seems to be with encrypting a message for multiple recipients. Here's a pure-Rust test case that fails:

```rust
    #[test]
    fn test_multiple_recipients() {
        // Generate 2 keys
        let (public_key1, secret_key1, _fingerprint1) = generate_source_key_pair(
            "correcthorsebatterystaple",
            "foo1@example.org",
        )
        .unwrap();
        let (public_key2, secret_key2, _fingerprint2) = generate_source_key_pair(
            "correcthorsebatterystaple",
            "foo2@example.org",
        )
        .unwrap();

        let tmp = NamedTempFile::new().unwrap();
        // Encrypt a message
        encrypt_message(
            vec![public_key1, public_key2],
            "Rust is great 🦀".to_string(),
            tmp.path().to_path_buf(),
        )
        .unwrap();
        let ciphertext = std::fs::read_to_string(tmp.path()).unwrap();
        // Verify ciphertext looks like an encrypted message
        assert!(ciphertext.starts_with("-----BEGIN PGP MESSAGE-----\n"));
        // Decrypt as key 1
        let plaintext = decrypt(
            ciphertext.clone().into_bytes(),
            secret_key1,
[...]

Issue #6801, opened 2023-05-17
request review by Sequoia team


cfm:
For us to have confidence in our adoption of Sequoia, we should request a code review from the Sequoia team, to confirm that our approach is sound, API usage is correct and stable, etc.

For completeness, this should take place after the implementation of:
* #6799 
* #6800 
* #6802

legoktm:
I sent an [email](https://lists.sequoia-pgp.org/hyperkitty/list/devel@lists.sequoia-pgp.org/thread/DUGVGBDCFI72LDIUZO6AXRPQGZA6YWGB/) to the Sequoia dev mailing list yesterday asking for the review and it's in progress!

PR #6891, opened 2023-07-19
redwood: Support decrypting messages to multiple recipients

legoktm:
## Status

Ready for review

## Description of Changes

As described in the commit message, the existing decryption code only supported decrypting for the first recipient of a message, not all of them. This functionality is now verified by tests.

Refs #6799.

## Testing

* [x] CI passes

Also I'm going to rebase https://github.com/freedomofpress/securedrop/tree/sequoia-initial on top of this, so those tests passing should give additional confidence. (Really this should be part of that stack but I'm pulling this out so it's easier to review)

## Deployment

## Checklist

### DIFF
--- a/redwood/src/decryption.rs
+++ b/redwood/src/decryption.rs
@@ -69,19 +69,17 @@ impl<'a> DecryptionHelper for Helper<'a> {
             // but we don't generate any messages that use it.
             if pkesk.recipient() == &KeyID::from(key.fingerprint()) {
                 // Decrypt the secret key with the specified passphrase.
-                let mut pair =
-                    key.decrypt_secret(self.passphrase)?.into_keypair()?;
+                let mut pair = key
+                    .clone()
+                    .decrypt_secret(self.passphrase)?
+                    .into_keypair()?;
                 pkesk
                     .decrypt(&mut pair, sym_algo)
                     .map(|(algo, session_key)| decrypt(algo, &session_key));
-                // XXX: The documentation says:
-                // > If the message is decrypted using a PKESK packet, then the
-                // > fingerprint of the certificate containing the encryption subkey
-                // > should be returned. This is used in conjunction with the intended
-                // > recipient subpacket (see Section 5.2.3.29 of RFC 4880bis) to
-                // > prevent Surreptitious Forwarding.
-                // Unclear if that's something we need to do.
-                return Ok(None);
+                // Return the fingerprint of the key we decrypted with, this is used in
+                // conjunction with the intended recipient subpacket (see "Intended Recipient Fingerprint"
+                // in RFC 4880bis) to prevent Surreptitious Forwarding.
+                return Ok(Some(key.fingerprint()));
             }
         }
 
--- a/redwood/src/decryption.rs
+++ b/redwood/src/decryption.rs
@@ -6,6 +6,7 @@ use sequoia_openpgp::crypto::{Password, SessionKey};
 use sequoia_openpgp::parse::stream::*;
 use sequoia_openpgp::policy::Policy;
 use sequoia_openpgp::types::SymmetricAlgorithm;
+use sequoia_openpgp::KeyID;
 
 pub(crate) struct Helper<'a> {
     pub(crate) policy: &'a dyn Policy,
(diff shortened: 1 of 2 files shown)


redwood/src/decryption.rs
```diff
@@ -58,20 +60,25 @@ impl<'a> DecryptionHelper for Helper<'a> {
             .key()
             .clone();
 
-        // Decrypt the secret key with the specified passphrase.
-        let mut pair = key.decrypt_secret(self.passphrase)?.into_keypair()?;
+        for pkesk in pkesks {
+            if pkesk.recipient() == &KeyID::from(key.fingerprint()) {
```
wiktor-k, line 70: Just a slight note that this would prevent messages encrypted using [`--throw-keyids`](https://superuser.com/a/1493135) from decryption (as then the `pkesk.recipient().is_wildcard()` is `true`).

The more elaborate example in `decrypt-with.rs` file also has the same issue: https://gitlab.com/sequoia-pgp/sequoia/blob/main/openpgp/examples/decrypt-with.rs#L100

(Which may, nor may not be important here but I'd like to point out drawbacks anyway).

redwood/src/decryption.rs
```diff
@@ -58,20 +60,25 @@ impl<'a> DecryptionHelper for Helper<'a> {
...
+                    .decrypt(&mut pair, sym_algo)
+                    .map(|(algo, session_key)| decrypt(algo, &session_key));
+                // XXX: The documentation says:
+                // > If the message is decrypted using a PKESK packet, then the
+                // > fingerprint of the certificate containing the encryption subkey
+                // > should be returned. This is used in conjunction with the intended
+                // > recipient subpacket (see Section 5.2.3.29 of RFC 4880bis) to
+                // > prevent Surreptitious Forwarding.
+                // Unclear if that's something we need to do.
+                return Ok(None);
```
wiktor-k: :thinking: It should be as easy as `return Ok(key.fingerprint());` ...

redwood/src/decryption.rs
```diff
@@ -50,6 +51,7 @@ impl<'a> DecryptionHelper for Helper<'a> {
             .keys()
             .secret()
             .with_policy(self.policy, None)
+            .for_storage_encryption()
```
wiktor-k, line 58: You may also want to insert `.supported()` here to filter out keys that are not supported by currently running backend, similarly to: https://gitlab.com/sequoia-pgp/sequoia/blob/main/openpgp/examples/decrypt-with.rs#L72

wiktor-k (commented):
A couple of minor points, I hope you don't mind...

redwood/src/decryption.rs
legoktm: TIL about `--throw-keyids`. In this case we are only decrypting messages generated by us (either through the web UI or SecureDrop Client) and we don't use that option. I'll add a comment noting this limitation.

redwood/src/decryption.rs
legoktm: Yeah, it only clicked sometime last night that this was trivial to do, added to the PR.

redwood/src/decryption.rs
legoktm: Good point. We already have that and `.alive()` in the `encrypt()` function, was just missing here. Also going to add a comment to keep them in sync going forwards.

legoktm:
Thanks @wiktor-k, appreciate the review!

wiktor-k:
> Thanks @wiktor-k, appreciate the review!

No problem! I saw you posting this in `#sequoia`. I'm happy to review more Sequoia stuff so if you have more in the future just tag me or send a link again.

Have a nice day! 👋

legoktm:
Awesome, appreciate the offer! Once we're farther down the implementation side, it's on our list (#6801) to ask for a review from Sequoia folks, will definitely ping you at that time too.

wiktor-k:
> ...to ask for a review from Sequoia folks, will definitely ping you at that time too.

Not sure if it's extra clear but for full disclosure I am working full time on and for Sequoia (see https://sequoia-pgp.org/community/ and https://gitlab.com/sequoia-pgp/sequoia/-/graphs/main).

I don't think it matters much in open source space but just to avoid confusion I'd rather spell it out explicitly :)

See you later! 👋

redwood/src/decryption.rs
```diff
@@ -58,20 +64,25 @@ impl<'a> DecryptionHelper for Helper<'a> {
...
+                let mut pair = key
+                    .clone()
+                    .decrypt_secret(self.passphrase)?
+                    .into_keypair()?;
+                pkesk
+                    .decrypt(&mut pair, sym_algo)
+                    .map(|(algo, session_key)| decrypt(algo, &session_key));
+                // Return the fingerprint of the key we decrypted with, this is used in
+                // conjunction with the intended recipient subpacket (see Section 5.2.3.29
+                // of RFC 4880bis) to prevent Surreptitious Forwarding.
```
cfm: Nit:

```suggestion
                // conjunction with the intended recipient subpacket (see "Intended Recipient Fingerprint"
                // in RFC 4880bis) to prevent Surreptitious Forwarding.
```

Just because RFC 4880bis isn't published yet, so the enumerated reference is still drifting about.  (I've seen all of 5.2.3.{29,30,36} just while trying to pin it down!)  I think it's more useful to know _what_ to look for anyway.

cfm (changes requested):
Nice.  It's great to see how easy it is to work with this implementation as we learn more.  (Thanks for the detailed commit messages, too.)  And thanks, @wiktor-k, for your volunteer review.  :-)

One nit, @legoktm, just because I'm feeling RFC-eyed these days, but then I'll approve for merge.

redwood/src/decryption.rs
legoktm: Done, thanks. This was copied out of Sequoia's docs IIRC, I'll find the link so you can also suggest this upstream :)

legoktm: Oh, Sequoia has a link so it's not ambiguous: https://gitlab.com/sequoia-pgp/sequoia/-/blob/main/openpgp/src/parse/stream.rs?ref_type=heads#L2084 - up to you on submitting it upstream then. (I think having a link for our inline docs would be overkill especially now that we're using the section header name)

cfm: Thanks!  I agree that the upstream reference to a specific version is stable and so fine as is.

cfm (approved):
Great.  Thanks, @legoktm!

Issue #6802, opened 2023-05-17
On-demand migration of sources' private keys to database-backed storage upon login


legoktm:
As part of our work to migrate to Sequoia, we will migrate sources' private keys on-demand when they log in to the database-backed storage.

As discussed in #6499, gpg stores keys in their own encrypted format, and we need the source's passphrase to unlock them, so we can't do it offline (like the public key migration). The plan is that when a source logs in, we'll see that they don't have a private key in the database, so we'll try to export their private key from gpg and then store it in the database in the armored and encrypted OpenPGP format.

legoktm:
Likely the main work here is that pretty_bad_protocol (AFAICT) doesn't already have a function to export private keys, so that will need to be done as part of this task, whether in p_b_p or just separately and directly calling GPG.

legoktm:
> Likely the main work here is that pretty_bad_protocol (AFAICT) doesn't already have a function to export private keys, so that will need to be done as part of this task, whether in p_b_p

This is implemented in #6907.

Issue #6800, opened 2023-05-17
One-time migration of sources' public keys and fingerprints from gpg to database-backed storage


legoktm:
As part of our work to migrate to Sequoia, we will perform a one-time, alembic migration of sources' public keys and fingerprints from the gpg keyring and into the new database backed storage (added in https://github.com/freedomofpress/securedrop/issues/6799).

This will iterate over each source, fetch the armored public key out of the gpg keyring, get the fingerprint, and store both in the database.

PR #6892, opened 2023-07-21
New source creation uses Sequoia

### LINKED ISSUE #6799: New source creation uses Sequoia

legoktm:
## Status

Ready for review

## Description of Changes

At a high-level, this PR includes:

* Storing journalist public key out of the GPG keyring - including a hopefully not too hacky postinst migration
* DB schema change, adding 3 columns to the source table
* Have redwood encrypt streams instead of files
* Changing all the EncryptionManager.encrypt_* functions to use redwood, and sometimes the decrypt function
* Actually switching the new source creation to Sequoia
* A bunch of fixes to tests

Each commit message should give a detailed explanation.

Fixes #6799.

## Testing

This test plan focuses on:
* key generation and storage
* GPG<-->PGP interop points
* journalist pub key migration

I'm intentionally skipping manual testing of GPG-created sources because that makes sense to test during the PRs for public key and private key migrations since the behavior will be different. For the purposes of this PR, the automated test coverage should be sufficient.

### In the dev environment
* [x] Spinning up a dev environment with `make dev` works
* [x] In the running container, `select * from sources` and see that public and private keys of the pre-generated sources are stored in the database
* [x] In the running container, `echo "select pgp_secret_key from sources limit 1" | sqlite3 /var/lib/securedrop/db.sqlite`, copy the private key to a separate file, `test.asc`, and then run `sq inspect test.asc` and observe that it says: "Secret key: Encrypted" for both the key and the encryption subkey. It should also say its a 4096-bit RSA key.
* [x] Create a new source through the SI (key generation works via web)
* [x] Log in as a source, submit a file and a message
* [x] Log into the JI and download said file + message and decrypt them using GPG
* [x] Send a reply to the source from the JI. Switch back to the SI and verify you're able to read it (it decrypts properly)
* [x] Back to the JI, download your own reply, and verify you can decrypt it using GPG (it was correctly encrypted to both the source and the journalist)  
### On a staging instance
* [x] Set up a 2.6.0 prod/staging instance
* [x] Copy `securedrop/tests/files/test_journalist_key.pub` from the `develop` branch and import it into the keyring (`sudo -u www-data gpg --homedir /var/lib/securedrop/keys --import test_journalist_key.pub`). This will make sure the non-SHA-1 version of the key is being used. 
* [x] Run `sudo -u www-data gpg --homedir /var/lib/securedrop/keys -k`, it should show the public journalist key (and possibly a few source keys)
* [x] Check out this PR and run `make build-debs`
* [x] Copy and install the securedrop-app-code deb onto the app server.
* [x] Verify `/var/lib/securedrop/journalist.pub` now exists. It should be owned by root, with group www-data, and 640 permissions (root gets read-write, www-data is read-only, world has nothing).
* [x] Copy journalist.pub to somewhere that has `sq` installed and pass it to `sq inspect` - the fingerprint should match the gpg output from earlier and what's specified in your config.py.
* [x] Create a new source and successfully submit a message
* [x] Log in as a journalist and submit a reply

### SDW integration
* [ ] Point an SDW setup at the staging/prod instance you created
* [ ] Log in as the journalist, verify you can decrypt/read both the source message and the journalist reply
* [ ] Submit a reply to the source
* [ ] Switch back to the SI, verify you can read the reply sent from SDW as the source.

## Deployment

Any special considerations for deployment? Yes, the DB migration is one way.

## Checklist


securedrop/encryption.py
```diff
@@ -34,27 +36,14 @@ class GpgDecryptError(Exception):
...
-    DEFAULT_KEY_EXPIRATION_DATE = "0"
-
     REDIS_FINGERPRINT_HASH = "sd/crypto-util/fingerprints"
     REDIS_KEY_HASH = "sd/crypto-util/keys"
 
-    SOURCE_KEY_NAME = "Source Key"
     SOURCE_KEY_UID_RE = re.compile(r"(Source|Autogenerated) Key <[-A-Za-z0-9+/=_]+>")
 
-    def __init__(self, gpg_key_dir: Path, journalist_key_fingerprint: str) -> None:
+    def __init__(self, gpg_key_dir: Path, data_root: Path) -> None:
```
nabla-c0d3: Was there a specific reason for this change ?
The downside is that the `EncryptionManager` class now has to know the layout of Securedrop's data folder; specifically where the journalist's public key is expected to be. To me, it seems like unnecessary "coupling" (between the config / folder layout and `EncryptionManager`).

securedrop/encryption.py
```diff
@@ -134,63 +102,48 @@ def get_source_key_fingerprint(self, source_filesystem_id: str) -> str:
...
-            plaintext_in=reply_as_stream,
-            ciphertext_path_out=encrypted_reply_path_out,
+            recipients=[source.public_key, self.get_journalist_public_key()],
+            plaintext=reply_in,
+            destination=encrypted_reply_path_out,
         )
 
-    def decrypt_journalist_reply(self, for_source_user: "SourceUser", ciphertext_in: bytes) -> str:
+    def decrypt_journalist_reply(
+        self, for_source_user: "SourceUser", for_source: "Source", ciphertext_in: bytes
```
nabla-c0d3: I think it's confusing to have both `for_source_user` and `for_source`. For example, what happens if they don't match? How will the reader understand without having to look at the implementation?

Why not just have `for_source_user` ? The source can then be retrieved via `for_source_user.get_db_record()`.

securedrop/encryption.py
```diff
@@ -134,63 +102,48 @@ def get_source_key_fingerprint(self, source_filesystem_id: str) -> str:
...
-            plaintext_in=file_in,
-            ciphertext_path_out=encrypted_file_path_out,
+            recipients=[self.get_journalist_public_key()],
+            plaintext=file_in,
+            destination=encrypted_file_path_out,
         )
 
     def encrypt_journalist_reply(
-        self, for_source_with_filesystem_id: str, reply_in: str, encrypted_reply_path_out: Path
+        self, source: "Source", reply_in: str, encrypted_reply_path_out: Path
```
nabla-c0d3: For consistency, I would call the argument `for_source` instead of `source`, to mirror the argument names of `decrypt_journalist_reply()` function below.

securedrop/tests/functional/app_navigators/journalist_app_nav.py
```diff
@@ -16,7 +14,8 @@
 from selenium.webdriver.support import expected_conditions
 from tests.functional.app_navigators._nav_helper import NavigationHelper
 from tests.functional.tor_utils import proxies_for_url
-from tests.test_encryption import import_journalist_private_key
+
+from ... import utils
```
nabla-c0d3: For best readability, I would recommend not mixing absolute imports and relative imports; why not just use absolute imports everywhere ?

nabla-c0d3 (commented):
I had a quick look and left a few comments, most of them pretty minor but hopefully helpful! Nice work on the new Rust encryption library.

securedrop/encryption.py
legoktm: Since Sequoia doesn't have a keyring, we need to store the journalist key somewhere that can be read from. It theoretically could be in the DB but I thought putting it in the filesystem made the provisioning and upgrading logic simpler.

Maybe we could pass in the full path to the journalist key rather than data root?

securedrop/encryption.py
legoktm: I went back and forth on this a few times so I'm glad you're raising this because I'm not really satisfied with the current implementation. I didn't want to call `get_db_record()` since it seems wrong that the EncryptionManager is making database queries - and we already have the `Source` on hand. The issue of for_source_user/for_source not matching is bad, which I hadn't really thought of.

securedrop/tests/functional/app_navigators/journalist_app_nav.py
legoktm: I would like to switch to relative imports eventually but yeah, it should stay consistent until we're ready to migrate everything.

securedrop/encryption.py
legoktm: OK, I switched it so that `EncryptionManager.__init__` takes the full path to the pub key, but the `data_root / journalist.pub` part happens in get_default(). Does that seem reasonable to you or should I push it one level higher (in SecureDropConfig??)

securedrop/encryption.py
legoktm: I'm going to switch to just taking the SourceUser and calling get_db_record() with a TODO comment about not making DB queries.

securedrop/tests/functional/app_navigators/journalist_app_nav.py
legoktm: Switched back to absolute imports, will file a task for my ideas on relative imports.

legoktm: #6902

securedrop/encryption.py
```diff
@@ -34,27 +36,14 @@ class GpgDecryptError(Exception):
...
     REDIS_KEY_HASH = "sd/crypto-util/keys"
 
-    SOURCE_KEY_NAME = "Source Key"
     SOURCE_KEY_UID_RE = re.compile(r"(Source|Autogenerated) Key <[-A-Za-z0-9+/=_]+>")
 
-    def __init__(self, gpg_key_dir: Path, journalist_key_fingerprint: str) -> None:
+    def __init__(self, gpg_key_dir: Path, journalist_pub_key: Path) -> None:
         self._gpg_key_dir = gpg_key_dir
-        self._journalist_key_fingerprint = journalist_key_fingerprint
+        self.journalist_pub_key = journalist_pub_key
```
nabla-c0d3, line 47: I think this is good 👍

nabla-c0d3: Optionally, you could check if the file exists here.

securedrop/encryption.py
```diff
@@ -71,43 +60,23 @@ def __init__(self, gpg_key_dir: Path, journalist_key_fingerprint: str) -> None:
...
-
     @classmethod
     def get_default(cls) -> "EncryptionManager":
         global _default_encryption_mgr
         if _default_encryption_mgr is None:
             config = SecureDropConfig.get_current()
             _default_encryption_mgr = cls(
                 gpg_key_dir=config.GPG_KEY_DIR,
-                journalist_key_fingerprint=config.JOURNALIST_KEY,
+                journalist_pub_key=(config.SECUREDROP_DATA_ROOT / "journalist.pub"),
```
nabla-c0d3, line 75: Looks good too

nabla-c0d3 (approved):
The changes look good. 
As for imports, absolute imports are usually more readable and straightforward than relative imports, so I would recommend using absolute imports everywhere.

securedrop/encryption.py
legoktm: Oh good point, done.

cfm:
@legoktm, would you mind rebasing and resolving conflicts from `develop` after #6884 and #6891?  I'm starting review today commit by commit, but I'd like to read the overall diff tomorrow before starting in on the test plan.

install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml
```diff
@@ -1,18 +1,24 @@
...
+    owner: "root"
+    group: "www-data"
+    mode: "0640"
   tags:
     - securedrop_config
 
 - name: Import the SecureDrop Application GPG public key to the Application Server keyring.
-  # multiline format for command module, since this is a long command
+  # TODO: We don't need this but it sets up the GPG keyring so let's leave it in place now
+  # and get rid of it later.
```
cfm: How would you feel about explicitly specifying when this can be removed?  I assume that's probably #6800 + #6802.

(Should the keyring itself be removed on upgrade as well?)

securedrop/debian/securedrop-app-code.postinst
```diff
@@ -168,6 +168,23 @@ update_apache2_headers(){
...
+# as part of the GPG -> Sequoia migration.
+export_journalist_public_key() {
+    # config.py is root-writable, so it's safe to run it as root to extract the fingerprint
+    journalist_pub="/var/lib/securedrop/journalist.pub"
+    if ! test -f $journalist_pub; then
+        # Based on sdconfig.py, this should work with very old config.py files.
+        fingerprint=$(cd /var/www/securedrop; python3 -c "import config; print(config.JOURNALIST_KEY)")
+        # Set up journalist.pub as root/www-data 640 before writing to it.
+        install --mode=640 --owner=root --group=www-data $journalist_pub
+        # FIXME: we should verify the exported key matches the fingerprint we wanted.
```
cfm: I agree that this would be a useful sanity check.

legoktm:
Done, now the parent PRs have been merged, if it would making reviewing easier, some parts of this could be split off into separate PRs - let me know.

securedrop/encryption.py
```diff
@@ -71,43 +64,23 @@ def __init__(self, gpg_key_dir: Path, journalist_key_fingerprint: str) -> None:
...
-            name_real=self.SOURCE_KEY_NAME,
-            creation_date=self.DEFAULT_KEY_CREATION_DATE.isoformat(),
-            expire_date=self.DEFAULT_KEY_EXPIRATION_DATE,
-        )
-        new_key = self._gpg.gen_key(gen_key_input)
-
-        # Store the newly-created key's fingerprint in Redis for faster lookups
-        self._save_key_fingerprint_to_redis(source_user.filesystem_id, str(new_key))
-
     def delete_source_key_pair(self, source_filesystem_id: str) -> None:
```
cfm, line 79: Nit: I would prefer a docstring like:

```suggestion
    def delete_source_key_pair(self, source_filesystem_id: str) -> None:
        """
        Try to delete the source's key from the filesystem.  If it's not found, either (a) it doesn't exist or (b) the source is Sequoia-based and has its key stored in Source.public_key, which will be deleted when the Source instance itself is deleted.
        """
```

securedrop/encryption.py
cfm, line 1: Because I try to "see through" the directional typing of messages and replies, I wonder how you would feel about a consolidation like the following.  (Even if you like it, it might be out of scope.)  References not updated in this diff:

```diff
--- a/securedrop/encryption.py
+++ b/securedrop/encryption.py
@@ -106,30 +106,32 @@ class EncryptionManager:
         self._save_key_fingerprint_to_redis(source_filesystem_id, source_key_fingerprint)
         return source_key_fingerprint
 
-    def encrypt_source_message(self, message_in: str, encrypted_message_path_out: Path) -> None:
-        redwood.encrypt_message(
-            # A submission is only encrypted for the journalist key
-            recipients=[self.get_journalist_public_key()],
-            plaintext=message_in,
-            destination=encrypted_message_path_out,
-        )
-
-    def encrypt_source_file(self, file_in: BinaryIO, encrypted_file_path_out: Path) -> None:
+    def encrypt_file(self, file_in: BinaryIO, encrypted_file_path_out: Path) -> None:
+        """A submission is only encrypted for the journalist key."""
         redwood.encrypt_stream(
-            # A submission is only encrypted for the journalist key
             recipients=[self.get_journalist_public_key()],
             plaintext=file_in,
             destination=encrypted_file_path_out,
         )
 
-    def encrypt_journalist_reply(
-        self, for_source: "Source", reply_in: str, encrypted_reply_path_out: Path
+    def encrypt_message(
+        self,
+        message_in: str,
+        encrypted_message_path_out: Path,
+        for_source: Optional["Source"] = None,
     ) -> None:
+        """
+        A submission (message) is only encrypted for the journalist key.
+        A reply is encrypted for both the journalist key and the source key.
+        """
+        recipients = self.get_journalist_public_key()
+        if for_source:
+            recipients.append(for_source.public_key)
+
         redwood.encrypt_message(
-            # A reply is encrypted for both the journalist key and the source key
-            recipients=[for_source.public_key, self.get_journalist_public_key()],
-            plaintext=reply_in,
-            destination=encrypted_reply_path_out,
+            recipients=recipients,
+            plaintext=message_in,
+            destination=encrypted_message_path_out,
         )
```

I'm tempted to argue for further consolidating `encrypt_file`, which would hide the decision between `redwood.encrypt_{message,stream}()` as an implementation detail of the `EncryptionManager`.  But that feels like pushing it.  :-)

securedrop/encryption.py
cfm: One way to solve this coupling would be with a decorator, e.g.:

```diff
--- a/securedrop/encryption.py
+++ b/securedrop/encryption.py
@@ -132,9 +132,10 @@ class EncryptionManager:
             destination=encrypted_reply_path_out,
         )
 
-    def decrypt_journalist_reply(self, for_source_user: "SourceUser", ciphertext_in: bytes) -> str:
-        # TODO: we shouldn't be making database queries here
-        for_source = for_source_user.get_db_record()
+    @with_source_from_source_user
+    def decrypt_journalist_reply(
+        self, for_source_user: "SourceUser", for_source: "Source", ciphertext_in: bytes
+    ) -> str:
         if for_source.pgp_secret_key is not None:
             return redwood.decrypt(
                 ciphertext_in,
```

That would let `with_source_from_source_user` be the glue between `decrypt_journalist_reply()`, which needs the `Source` but which we don't want to query the database, and its caller, which we don't want to be able pass an unmatched `SourceUser`/`Source` pair.

cfm (commented):
@legoktm, I've read through the implementation, and I think this is fabulous work.  I've left some comments inline: let me know what you think, but they're not blocking.  Tomorrow I will:

1. [x] step through your test plan; and
2. [x] read through the tests.

install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml
legoktm: Yeah I was thinking around the time we do the private keys migration, clarified the comment.

> (Should the keyring itself be removed on upgrade as well?)

Well the private keys will be migrated one-by-one when sources log in, so we'll just slowly empty the keyring. I think it's probably not worth trying to check if it's empty to delete it...

securedrop/debian/securedrop-app-code.postinst
legoktm: I think the reason I left this as a fixme was because to do so we'd have to use gpg :| but we can just copy the `validate-gpg-key.sh` from admin/bin/.

really we should install `sq` just as a utility/debugging tool, I'll file an issue for that...

securedrop/encryption.py
legoktm: Applied

securedrop/encryption.py
legoktm: I personally prefer to keep the functions separate just to make it harder to use wrong, or more obvious when used wrong at the cost of (some) duplication.

Having a function like `encrypt_journalist_reply()` means it just needs to be checked for correctness in the implementation, while `encrypt_message(for_source=...)` also needs to be verified at the call site so make sure the right thing is being passed in.

At a high level I see EncryptionManager providing a dedicated function for each type of PGP en/decryption we need, and redwood providing encrypt()/decrypt() primitives to support that. What do you think?

Sidenote: there's probably not a strong reason for the redwood encrypt_message vs encrypt_stream split anymore, the only benefit encrypt_message has is that it has slightly more type safety, but I'm not sure it's worth it still... 🤔

securedrop/encryption.py
legoktm: I don't think that really helps very much since in the only call site, we already have the `Source` object on hand, so really it should just be passing it through.

Maybe we just add a manual assertion that the two source objects point to the same filesystem_id?

(The underlying issue that trips me up is that SourceUser doesn't contain a Source object, instead loading it on-demand for unclear (to me) reasons: "We don't store the actual record to force a refresh". Don't think we need to solve that now though...)

securedrop/debian/securedrop-app-code.postinst
```diff
@@ -168,6 +168,23 @@ update_apache2_headers(){
...
+# Export the journalist public key out of the GPG keyring to journalist.pub
+# as part of the GPG -> Sequoia migration.
+export_journalist_public_key() {
+    # config.py is root-writable, so it's safe to run it as root to extract the fingerprint
+    journalist_pub="/var/lib/securedrop/journalist.pub"
+    if ! test -f $journalist_pub; then
+        # Based on sdconfig.py, this should work with very old config.py files.
+        fingerprint=$(cd /var/www/securedrop; python3 -c "import config; print(config.JOURNALIST_KEY)")
+        # Set up journalist.pub as root/www-data 640 before writing to it.
+        install --mode=640 --owner=root --group=www-data $journalist_pub
```
cfm:
```sh-session
vagrant@app-staging:~$ sudo apt install ./securedrop-app-code_2.7.0~rc1+focal_amd64.deb 
[...]
+ install --mode=640 --owner=root --group=www-data /var/lib/securedrop/journalist.pub
install: missing destination file operand after '/var/lib/securedrop/journalist.pub'
Try 'install --help' for more information.
 installed securedrop-app-code package post-installation script subprocess returned error exit status 1
 securedrop-app-code
```

Based on the usage statement—

```sh-session
vagrant@app-staging:~$ install --help | head
Usage: install [OPTION]... [-T] SOURCE DEST
  or:  install [OPTION]... SOURCE... DIRECTORY
  or:  install [OPTION]... -t DIRECTORY SOURCE...
  or:  install [OPTION]... -d DIRECTORY...

This install program copies files (often just compiled) into destination
locations you choose.  If you want to download and install a ready-to-use
package on a GNU/Linux system, you should instead be using a package manager
like yum(1) or apt-get(1).
```

—I'm not sure `install` provides the intended behavior of an atomic `touch && chown && chmod`.

cfm: This seems to do the trick:

```patch
--- a/securedrop/debian/securedrop-app-code.postinst
+++ b/securedrop/debian/securedrop-app-code.postinst
@@ -177,7 +177,9 @@ export_journalist_public_key() {
         # Based on sdconfig.py, this should work with very old config.py files.
         fingerprint=$(cd /var/www/securedrop; python3 -c "import config; print(config.JOURNALIST_KEY)")
         # Set up journalist.pub as root/www-data 640 before writing to it.
-        install --mode=640 --owner=root --group=www-data $journalist_pub
+        touch $journalist_pub
+        chown root:www-data $journalist_pub
+        chmod 640 $journalist_pub
         # FIXME: we should verify the exported key matches the fingerprint we wanted.
         # shellcheck disable=SC2024
         sudo -u www-data gpg2 --homedir=/var/lib/securedrop/keys --export --armor "$fingerprint" > $journalist_pub
```

securedrop/debian/securedrop-app-code.postinst
cfm, line 1: <details><summary>On first upgrade, the application is reloaded before the database is migrated.</summary>

```
[Thu Aug 10 22:00:05.489390 2023] [wsgi:error] [pid 134774:tid 115485936953088]   File "/var/www/securedro
p/journalist.py", line 17, in prime_keycache
[Thu Aug 10 22:00:05.489427 2023] [wsgi:error] [pid 134774:tid 115485936953088]     for source in Source.q
uery.filter_by(pending=False, deleted_at=None).all():
[Thu Aug 10 22:00:05.489441 2023] [wsgi:error] [pid 134774:tid 115485936953088]   File "/opt/venvs/secured
rop-app-code/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 3161, in all
[Thu Aug 10 22:00:05.489818 2023] [wsgi:error] [pid 134774:tid 115485936953088]     return list(self)
[Thu Aug 10 22:00:05.489824 2023] [wsgi:error] [pid 134774:tid 115485936953088]   File "/opt/venvs/secured
rop-app-code/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 3317, in __iter__
[Thu Aug 10 22:00:05.490206 2023] [wsgi:error] [pid 134774:tid 115485936953088]     return self._execute_a
nd_instances(context)
[Thu Aug 10 22:00:05.490212 2023] [wsgi:error] [pid 134774:tid 115485936953088]   File "/opt/venvs/secured
rop-app-code/lib/python3.8/site-packages/sqlalchemy/orm/query.py", line 3342, in _execute_and_instances
[Thu Aug 10 22:00:05.490604 2023] [wsgi:error] [pid 134774:tid 115485936953088]     result = conn.execute(
querycontext.statement, self._params)
[Thu Aug 10 22:00:05.490610 2023] [wsgi:error] [pid 134774:tid 115485936953088]   File "/opt/venvs/secured
rop-app-code/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 988, in execute
[Thu Aug 10 22:00:05.490747 2023] [wsgi:error] [pid 134774:tid 115485936953088]     return meth(self, mult
iparams, params)
[Thu Aug 10 22:00:05.490754 2023] [wsgi:error] [pid 134774:tid 115485936953088]   File "/opt/venvs/secured
rop-app-code/lib/python3.8/site-packages/sqlalchemy/sql/elements.py", line 287, in _execute_on_connection
[Thu Aug 10 22:00:05.490818 2023] [wsgi:error] [pid 134774:tid 115485936953088]     return connection._exe
cute_clauseelement(self, multiparams, params)
[Thu Aug 10 22:00:05.490824 2023] [wsgi:error] [pid 134774:tid 115485936953088]   File "/opt/venvs/secured
rop-app-code/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1101, in _execute_clauseelement
[Thu Aug 10 22:00:05.490972 2023] [wsgi:error] [pid 134774:tid 115485936953088]     ret = self._execute_co
ntext(
[Thu Aug 10 22:00:05.490978 2023] [wsgi:error] [pid 134774:tid 115485936953088]   File "/opt/venvs/secured
rop-app-code/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1247, in _execute_context
[Thu Aug 10 22:00:05.491139 2023] [wsgi:error] [pid 134774:tid 115485936953088]     self._handle_dbapi_exc
eption(
[Thu Aug 10 22:00:05.491145 2023] [wsgi:error] [pid 134774:tid 115485936953088]   File "/opt/venvs/secured
rop-app-code/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1466, in _handle_dbapi_exception
[Thu Aug 10 22:00:05.491330 2023] [wsgi:error] [pid 134774:tid 115485936953088]     util.raise_from_cause(
sqlalchemy_exception, exc_info)
[Thu Aug 10 22:00:05.491336 2023] [wsgi:error] [pid 134774:tid 115485936953088]   File "/opt/venvs/secured
rop-app-code/lib/python3.8/site-packages/sqlalchemy/util/compat.py", line 383, in raise_from_cause
[Thu Aug 10 22:00:05.491409 2023] [wsgi:error] [pid 134774:tid 115485936953088]     reraise(type(exception
), exception, tb=exc_tb, cause=cause)
[Thu Aug 10 22:00:05.491415 2023] [wsgi:error] [pid 134774:tid 115485936953088]   File "/opt/venvs/secured
rop-app-code/lib/python3.8/site-packages/sqlalchemy/util/compat.py", line 128, in reraise
[Thu Aug 10 22:00:05.491460 2023] [wsgi:error] [pid 134774:tid 115485936953088]     raise value.with_trace
back(tb)
[Thu Aug 10 22:00:05.491466 2023] [wsgi:error] [pid 134774:tid 115485936953088]   File "/opt/venvs/secured
rop-app-code/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1243, in _execute_context
[Thu Aug 10 22:00:05.491626 2023] [wsgi:error] [pid 134774:tid 115485936953088]     self.dialect.do_execut
e(
[Thu Aug 10 22:00:05.491632 2023] [wsgi:error] [pid 134774:tid 115485936953088]   File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/sqlalchemy/engine/default.py", line 552, in do_execute
[Thu Aug 10 22:00:05.491725 2023] [wsgi:error] [pid 134774:tid 115485936953088]     cursor.execute(statement, parameters)
[Thu Aug 10 22:00:05.491749 2023] [wsgi:error] [pid 134774:tid 115485936953088] sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such column: sources.pgp_public_key
```
</details>

May be fixable as simply as:

```patch
--- a/securedrop/debian/securedrop-app-code.postinst
+++ b/securedrop/debian/securedrop-app-code.postinst
@@ -276,9 +276,6 @@ case "$1" in
     update_apache2_headers /etc/apache2/sites-available/source.conf
     update_apache2_headers /etc/apache2/sites-available/journalist.conf
 
-    # Restart apache so it loads with the apparmor profiles in enforce mode.
-    service apache2 restart
-
     # remove previously dynamically-generated assets
     rm -fr /var/www/securedrop/static/gen/
     rm -fr /var/www/securedrop/static/.webassets-cache/
@@ -289,6 +286,9 @@ case "$1" in
     # Version migrations
     database_migration
 
+    # Restart apache so it loads with the apparmor profiles in enforce mode.
+    service apache2 restart
+
     ;;
```

cfm (changes requested):
Test plan checks out in the development environment.  In the staging environment, there are some upgrade glitches, detailed below and inline.  I've started to diagnose and sketch fixes for these.  ~~@legoktm, next week I'll propose fixup commits and continue with the test plan while you're away.~~

### In the dev environment
* [x] Spinning up a dev environment with `make dev` works
* [x] In the running container, `select * from sources` and see that public and private keys of the pre-generated sources are stored in the database
* [x] In the running container, `echo "select pgp_secret_key from sources limit 1" | sqlite3 /var/lib/securedrop/db.sqlite`, copy the private key to a separate file, `test.asc`, and then run `sq inspect test.asc` and observe that it says: "Secret key: Encrypted" for both the key and the encryption subkey. It should also say its a 4096-bit RSA key.
* [x] Create a new source through the SI (key generation works via web)
* [x] Log in as a source, submit a file and a message
* [x] Log into the JI and download said file + message and decrypt them using GPG
* [x] Send a reply to the source from the JI. Switch back to the SI and verify you're able to read it (it decrypts properly)
* [x] Back to the JI, download your own reply, and verify you can decrypt it using GPG (it was correctly encrypted to both the source and the journalist)  
### On a staging instance
* [x] Set up a 2.6.0 prod/staging instance
* [x] Run `sudo -u www-data gpg --homedir /var/lib/securedrop/keys -k`, it should show the public journalist key (and possibly a few source keys)
* [x] Check out this PR and run `make build-debs`
* [ ] Copy and install the securedrop-app-code deb onto the app server.

**Fails;** see inline comments.

* [x] Verify `/var/lib/securedrop/journalist.pub` now exists. It should be owned by root, with group www-data, and 640 permissions (root gets read-write, www-data is read-only, world has nothing).
* [ ] Copy journalist.pub to somewhere that has `sq` installed and pass it to `sq inspect` - the fingerprint should match the gpg output from earlier and what's specified in your config.py.

**Yes, but compare:**

```sh-session
root@sd-staging:~# sq inspect journalist.pub 
journalist.pub: OpenPGP Certificate.

    Fingerprint: 65A1B5FF195B56353CC63DFFCC40EF1228271441
                 Invalid: No binding signature at time 2023-08-10T22:03:47Z
Public-key algo: RSA (Encrypt or Sign)
Public-key size: 4096 bits
  Creation time: 2013-10-12 17:48:16 UTC

         UserID: SecureDrop Test/Development (DO NOT USE IN PRODUCTION)
                 Invalid: Policy rejected non-revocation signature (PositiveCertification) requiring second pre-image resistance
                 because: SHA1 is not considered secure since 2023-02-01T00:00:00Z

root@sd-staging:~# cd securedrop
root@sd-staging:~/securedrop# sq inspect securedrop/tests/files/test_journalist_key.pub 
securedrop/tests/files/test_journalist_key.pub: OpenPGP Certificate.

    Fingerprint: 65A1B5FF195B56353CC63DFFCC40EF1228271441
Public-key algo: RSA (Encrypt or Sign)
Public-key size: 4096 bits
  Creation time: 2013-10-12 17:48:16 UTC
      Key flags: certification, signing

         Subkey: C1C4E16BB24E4F4ABF37C3A6C3E7C4C0A2201B2A
Public-key algo: RSA (Encrypt or Sign)
Public-key size: 4096 bits
  Creation time: 2013-10-12 17:48:16 UTC
      Key flags: transport encryption, data-at-rest encryption

         UserID: SecureDrop Test/Development (DO NOT USE IN PRODUCTION)
```

* [ ] Create a new source and successfully submit a message

**No, fails in:** (see comments on `postinst`)

```
[Thu Aug 10 22:14:37.482295 2023] [wsgi:error] [pid 135388:tid 114964295517952] [remote 127.0.0.1:32846]   File "/var/www/securedrop/source_app/main.py", line 262, in submit
[Thu Aug 10 22:14:37.482302 2023] [wsgi:error] [pid 135388:tid 114964295517952] [remote 127.0.0.1:32846]     Storage.get_default().save_message_submission(
[Thu Aug 10 22:14:37.482305 2023] [wsgi:error] [pid 135388:tid 114964295517952] [remote 127.0.0.1:32846]   File "/var/www/securedrop/store.py", line 363, in save_message_submission
[Thu Aug 10 22:14:37.482306 2023] [wsgi:error] [pid 135388:tid 114964295517952] [remote 127.0.0.1:32846]     EncryptionManager.get_default().encrypt_source_message(
[Thu Aug 10 22:14:37.482308 2023] [wsgi:error] [pid 135388:tid 114964295517952] [remote 127.0.0.1:32846]   File "/var/www/securedrop/encryption.py", line 112, in encrypt_source_message
[Thu Aug 10 22:14:37.482310 2023] [wsgi:error] [pid 135388:tid 114964295517952] [remote 127.0.0.1:32846]     recipients=[self.get_journalist_public_key()],
[Thu Aug 10 22:14:37.482312 2023] [wsgi:error] [pid 135388:tid 114964295517952] [remote 127.0.0.1:32846]   File "/var/www/securedrop/encryption.py", line 92, in get_journalist_public_key
[Thu Aug 10 22:14:37.482314 2023] [wsgi:error] [pid 135388:tid 114964295517952] [remote 127.0.0.1:32846]     return self.journalist_pub_key.read_text()
[Thu Aug 10 22:14:37.482316 2023] [wsgi:error] [pid 135388:tid 114964295517952] [remote 127.0.0.1:32846]   File "/usr/lib/python3.8/pathlib.py", line 1236, in read_text
[Thu Aug 10 22:14:37.482318 2023] [wsgi:error] [pid 135388:tid 114964295517952] [remote 127.0.0.1:32846]     with self.open(mode='r', encoding=encoding, errors=errors) as f:
[Thu Aug 10 22:14:37.482320 2023] [wsgi:error] [pid 135388:tid 114964295517952] [remote 127.0.0.1:32846]   File "/usr/lib/python3.8/pathlib.py", line 1222, in open
[Thu Aug 10 22:14:37.482322 2023] [wsgi:error] [pid 135388:tid 114964295517952] [remote 127.0.0.1:32846]     return io.open(self, mode, buffering, encoding, errors, newline,
[Thu Aug 10 22:14:37.482324 2023] [wsgi:error] [pid 135388:tid 114964295517952] [remote 127.0.0.1:32846]   File "/usr/lib/python3.8/pathlib.py", line 1078, in _opener
[Thu Aug 10 22:14:37.482326 2023] [wsgi:error] [pid 135388:tid 114964295517952] [remote 127.0.0.1:32846]     return self._accessor.open(self, flags, mode)
[Thu Aug 10 22:14:37.482328 2023] [wsgi:error] [pid 135388:tid 114964295517952] [remote 127.0.0.1:32846] PermissionError: [Errno 13] Permission denied: '/var/lib/securedrop/journalist.pub'
```

* [ ] Log in as a journalist and submit a reply

### SDW integration
* [ ] Point an SDW setup at the staging/prod instance you created
* [ ] Log in as the journalist, verify you can decrypt/read both the source message and the journalist reply
* [ ] Submit a reply to the source
* [ ] Switch back to the SI, verify you can read the reply sent from SDW as the source.

securedrop/encryption.py
cfm: That's persuasive, @legoktm: scoping functions at this level to intention rather than implementation.  The more we can make that kind of layering explicit—in both code and developer documentation—the better!

securedrop/encryption.py
cfm: Right.  I think it would need to be an explicit `if...: raise` (rather than `assert`), but I'd be happy with that approach, @legoktm.

@nabla-c0d3, the `SourceUser.get_db_record()` originates in your 203a873b1ee16f6b590aa31dced9bf00eaccd632: can you refresh (as it were :-) our understanding of why this fetch is best done on demand?

securedrop/encryption.py
nabla-c0d3: > At a high level I see EncryptionManager providing a dedicated function for each type of PGP en/decryption we need, and redwood providing encrypt()/decrypt() primitives to support that. What do you think?

I agree with this. When I did a big refactoring of the crypto code, I renamed all the functions/methods to clarify what they do at a high-level in the context of SecureDrop/PGP "operations", so that it is easier to understand for the reader (changes are there https://github.com/freedomofpress/securedrop/pull/6160/files#diff-0cce30317a02d5977fd4f6c1e0ad70ffcc4fec5b31b3bca578b36e1179275e32L1 ). 

For example, I decided to split the `encrypt()` method into `encrypt_journalist_reply()` and `encrypt_source_message()`; much simpler for the reader to understand what the code does.

securedrop/encryption.py
nabla-c0d3: @cfm It was to ensure an up-to-date record is being used. Otherwise, if we just attach the `models.Source` to the `SourceUser`, the DB record can get stale if it gets updated some time later. I think SQLAlchemy might try to auto-fetch/update it but it will raise an exception if there is no DB session currently opened.

I'm not super happy with my `get_db_record()` approach tho, as it fetches data from the DB without explicitly using a DB session, which, in my opinion, is an anti-pattern (hidden "dependencies" / magical behavior, and will throw an exception if no DB session is currently open). 

So overall I think it's not "worse" to just attach the `models.Source` to the `SourceUser`, as both approaches end up having the same problem. So I would be okay with that change.

Here's an example of the problem:

```
# Open a DB session, fetch the Source record, and create a SourceUser
with get_db_session() as db_session:
  source = db_session.query(models.Source).filter_by(filesystem_id=f_id).one()
  source_user = SourceUser(source)
  print(source_user.source)  #  Will work fine
  print(source_user.source.submissions)  #  Will work fine

# DB session is now closed
print(source_user.source)  #  Will work fine
print(source_user.source.submissions)  #  Will throw an exception
```

I think we can worry about it once it's a real problem, and like I said, my `get_db_record()` approach does not address this issue either.

cfm: Thanks, @nabla-c0d3; I agree with the problem.  :-)  If we're going to have this anti-pattern either way, and it's not *enough* of a problem to be worth solving architecturally, then I'll defer to @legoktm on which anti-pattern we accept here.

securedrop/encryption.py
legoktm: I've added a one-line docstring to clarify the goal of EncryptionManager based on this thread.

securedrop/debian/securedrop-app-code.postinst
legoktm: Thanks, applied.

securedrop/debian/securedrop-app-code.postinst
legoktm: Hmmmm, this is definitely a pre-existing issue, nice catch. Applied with more detailed doc comments.

securedrop/encryption.py
legoktm: > the DB record can get stale if it gets updated some time later. I think SQLAlchemy might try to auto-fetch/update it but it will raise an exception if there is no DB session currently opened.

My understanding is that `models.Source` is just a reference to the actual, unique object in SQLAlchemy's identity map that corresponds to the row, and if something makes changes, the object will also be updated. And if there's no session, it should error.

So...my preference would be go to back to passing both source_user + source in, adding a check to verify they're the same, and creating an issue + leaving a TODO on get_db_record() to figure out what to do with it.

securedrop/debian/securedrop-app-code.postinst
legoktm: I re-read validate-gpg-key and I don't think it's usable here because it just verifies that the file contains a key with that fingerprint, not that it _only_ contains that fingerprint.

To list out the properties we want to validate:
1. The file contains a valid PGP key.
2. The file contains one PGP key.
3. The file contains a PGP key matching the specified fingerprint.

This is pretty straightforward to do with pretty_bad_protocol actually, it's like:
`gpg.import_keys(key).fingerprints == [fingerprint]`

of course that would be evil to do in this PR, so maybe we introduce a new Sequoia function for this? hmmm :thinking:

legoktm:
This is probably the second most complicated thing we're adding to the postinst (the apache header editing the most complicated IMO). Testing the whole postinst would be very difficult, I'm wondering if we should split this routine into a separate file so we can write automated tests against it.

securedrop/encryption.py
cfm: Based on this discussion, that would be my preference too, @legoktm.

nabla-c0d3: I think it will be very confusing to the reader, and hence too problematic.

My recommendation would be to instead pick the other bad solution, which is to call `get_db_record()`. It will be "bad" but will only affect the internals of the function (ie. invisible to the reader/developer). 

Also, it can later easily be replaced by something like `for_source_user.source`; that's an easier change/refactor than removing an argument (the `for_source` arg) out of every location where `decrypt_journalist_reply()` is being called.

legoktm: Yeah I buy that. Ok, flip-flopping once more :)

legoktm: I filed https://github.com/freedomofpress/securedrop/issues/6934 summarizing the `get_db_record()` part of this convo.

legoktm:
OK, postinst works in my testing:
```
+ export_journalist_public_key
+ journalist_pub=/var/lib/securedrop/journalist.pub
+ test -f /var/lib/securedrop/journalist.pub
++ cd /var/www/securedrop
++ python3 -c 'import config; print(config.JOURNALIST_KEY)'
+ fingerprint=C5B783ABAD78C934314A7C6FD08BD765AFDB4308
+ touch /var/lib/securedrop/journalist.pub
+ chown root:www-data /var/lib/securedrop/journalist.pub
+ chmod 640 /var/lib/securedrop/journalist.pub
+ sudo -u www-data gpg2 --homedir=/var/lib/securedrop/keys --export --armor C5B783ABAD78C934314A7C6FD08BD765AFDB4308
+ sudo -u www-data /var/www/securedrop/scripts/validate-pgp-key /var/lib/securedrop/journalist.pub C5B783ABAD78C934314A7C6FD08BD765AFDB4308
Success! Specified fingerprint matches pubkey file.
```

And it matches my instance's key:
```
$ sq inspect journalist.pub 
journalist.pub: OpenPGP Certificate.

    Fingerprint: C5B783ABAD78C934314A7C6FD08BD765AFDB4308
Public-key algo: RSA
Public-key size: 4096 bits
  Creation time: 2023-05-12 16:19:58 UTC
      Key flags: certification, signing

         Subkey: F4BCE6BFBF1E88A5022FB14A38CE230F7B353423
Public-key algo: RSA
Public-key size: 4096 bits
  Creation time: 2023-05-12 16:19:58 UTC
      Key flags: transport encryption, data-at-rest encryption

         UserID: SecureDrop ([UWS] SecureDrop Submission Key 2023)
```

legoktm:
The last thing (I think) in https://github.com/freedomofpress/securedrop/pull/6892#pullrequestreview-1572781681 is that the test key used on a staging instance is still a SHA-1 key. This is because the 2.6.0 branch still has it as a weak key, then it's imported into the gpg keyring, and then with 2.7.0, it gets exported but it's not re-imported from the fixed key file. I'll update the test plan to import the fixed key before doing the upgrade.

cfm (changes requested):
### In the dev environment
https://github.com/freedomofpress/securedrop/pull/6892#pullrequestreview-1572781681

### On a staging instance
* [x] Set up a 2.6.0 prod/staging instance
* [x] Copy `securedrop/tests/files/test_journalist_key.pub` from the `develop` branch and import it into the keyring (`sudo -u www-data gpg --homedir /var/lib/securedrop/keys --import test_journalist_key.pub`). This will make sure the non-SHA-1 version of the key is being used. 

```sh-session
root@app-staging:~/securedrop# sudo -u www-data gpg --homedir /var/lib/securedrop/keys --import securedrop/tests/file/test_journalist_key.pub                                                                
gpg: key CC40EF1228271441: "SecureDrop Test/Development (DO NOT USE IN PRODUCTION)" 3 new signatures
gpg: Total number processed: 1
gpg:         new signatures: 3 
```

* [x] Run `sudo -u www-data gpg --homedir /var/lib/securedrop/keys -k`, it should show the public journalist key (and possibly a few source keys)

```sh-session
root@app-staging:~/securedrop# sudo -u www-data gpg --homedir /var/lib/securedrop/keys -k /var/lib/securedrop/keys/pubring.gpg
------------------------------------
pub   rsa4096 2013-10-12 [SC]
      65A1B5FF195B56353CC63DFFCC40EF1228271441
uid           [ unknown] SecureDrop Test/Development (DO NOT USE IN PRODUCTION)
sub   rsa4096 2013-10-12 [E]  
```

* [x] Check out this PR and run `make build-debs`
* [x] Copy and install the securedrop-app-code deb onto the app server.
* [x] Verify `/var/lib/securedrop/journalist.pub` now exists. It should be owned by root, with group www-data, and 640 permissions (root gets read-write, www-data is read-only, world has nothing).

```sh-session
root@app-staging:~/securedrop# ls -al /var/lib/securedrop/journalist.pub
-rw-r----- 1 root www-data 5645 Sep  5 20:53 /var/lib/securedrop/journalist.pub
```

* [x] Copy journalist.pub to somewhere that has `sq` installed and pass it to `sq inspect` - the fingerprint should match the gpg output from earlier and what's specified in your config.py.

```sh-session
root@sd-staging:~# ssh 192.168.121.58 -l vagrant -p 22 -i /root/.vagrant.d/insecure_private_key sudo cat /var/lib/securedrop/journalist.pub | sq inspect
-: OpenPGP Certificate.

    Fingerprint: 65A1B5FF195B56353CC63DFFCC40EF1228271441
Public-key algo: RSA (Encrypt or Sign)
Public-key size: 4096 bits
  Creation time: 2013-10-12 17:48:16 UTC
      Key flags: certification, signing

         Subkey: C1C4E16BB24E4F4ABF37C3A6C3E7C4C0A2201B2A
Public-key algo: RSA (Encrypt or Sign)
Public-key size: 4096 bits
  Creation time: 2013-10-12 17:48:16 UTC
      Key flags: transport encryption, data-at-rest encryption

         UserID: SecureDrop Test/Development (DO NOT USE IN PRODUCTION)
```

* [ ] Create a new source and successfully submit a message

**Failed:**

```sh-session
[Tue Sep 05 20:59:23.752753 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686] [2023-09-05 20:59:23,750] ERROR in app: Exception on /submit [POST]
[Tue Sep 05 20:59:23.752771 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686] Traceback (most recent call last):
[Tue Sep 05 20:59:23.752774 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]   File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 2073, in wsgi_app
[Tue Sep 05 20:59:23.752777 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]     response = self.full_dispatch_request()
[Tue Sep 05 20:59:23.752779 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]   File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 1518, in full_dispatch_request
[Tue Sep 05 20:59:23.752786 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]     rv = self.handle_user_exception(e)
[Tue Sep 05 20:59:23.752788 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]   File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 1516, in full_dispatch_request
[Tue Sep 05 20:59:23.752790 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]     rv = self.dispatch_request()
[Tue Sep 05 20:59:23.752792 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]   File "/opt/venvs/securedrop-app-code/lib/python3.8/site-packages/flask/app.py", line 1502, in dispatch_request
[Tue Sep 05 20:59:23.752795 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]     return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
[Tue Sep 05 20:59:23.752797 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]   File "/var/www/securedrop/source_app/decorators.py", line 27, in decorated_function
[Tue Sep 05 20:59:23.752799 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]     return f(*args, **kwargs, logged_in_source=logged_in_source)
[Tue Sep 05 20:59:23.752801 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]   File "/var/www/securedrop/source_app/main.py", line 262, in submit
[Tue Sep 05 20:59:23.752803 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]     Storage.get_default().save_message_submission(
[Tue Sep 05 20:59:23.752804 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]   File "/var/www/securedrop/store.py", line 363, in save_message_submission
[Tue Sep 05 20:59:23.752806 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]     EncryptionManager.get_default().encrypt_source_message(
[Tue Sep 05 20:59:23.752808 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]   File "/var/www/securedrop/encryption.py", line 119, in encrypt_source_message
[Tue Sep 05 20:59:23.752810 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]     recipients=[self.get_journalist_public_key()],
[Tue Sep 05 20:59:23.752812 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]   File "/var/www/securedrop/encryption.py", line 99, in get_journalist_public_key
[Tue Sep 05 20:59:23.752814 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]     return self.journalist_pub_key.read_text()
[Tue Sep 05 20:59:23.752816 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]   File "/usr/lib/python3.8/pathlib.py", line 1236, in read_text
[Tue Sep 05 20:59:23.752818 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]     with self.open(mode='r', encoding=encoding, errors=errors) as f:
[Tue Sep 05 20:59:23.752820 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]   File "/usr/lib/python3.8/pathlib.py", line 1222, in open
[Tue Sep 05 20:59:23.752821 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]     return io.open(self, mode, buffering, encoding, errors, newline,
[Tue Sep 05 20:59:23.752823 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]   File "/usr/lib/python3.8/pathlib.py", line 1078, in _opener
[Tue Sep 05 20:59:23.752825 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686]     return self._accessor.open(self, flags, mode)
[Tue Sep 05 20:59:23.752827 2023] [wsgi:error] [pid 135127:tid 124945101768448] [remote 127.0.0.1:58686] PermissionError: [Errno 13] Permission denied: '/var/lib/securedrop/journalist.pub
```

The following permission changes have no effect:

- `chmod 660 /var/lib/securedrop/journalist.pub`
- `chown www-data:www-data /var/lib/securedrop/journalist.pub`
- `chmod 664 /var/lib/securedrop/journalist.pub`

But observe:

```sh-session
root@app-staging:/home/vagrant# sudo -u www-data python3
Python 3.8.10 (default, May 26 2023, 14:05:08) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pathlib
>>> key = pathlib.Path("/var/lib/securedrop/journalist.pub")
>>> key.read_text()[:10]
'-----BEGIN'
```

I'll continue troubleshooting this later today if I have time.

* [ ] Log in as a journalist and submit a reply

### SDW integration
* [ ] Point an SDW setup at the staging/prod instance you created
* [ ] Log in as the journalist, verify you can decrypt/read both the source message and the journalist reply
* [ ] Submit a reply to the source
* [ ] Switch back to the SI, verify you can read the reply sent from SDW as the source.

legoktm:
apparmor :(

```
Sep  6 05:53:24 app-staging kernel: [ 6802.178341] audit: type=1400 audit(1693979604.315:20): apparmor="DENIED" operation="open" profile="/usr/sbin/apache2" name="/var/lib/securedrop/journalist.pub" pid=648 comm="apache2" requested_mask="r" denied_mask="r" fsuid=33 ouid=0
```

legoktm:
Changes:
* da06ad5c99359a2a719769640ab3b96a34e275b0 now adds journalist.pub to apparmor and fixes the `test_securedrop_application_test_journalist_key` staging test
* postinst now checks config.py exists before trying to export the key (it won't exist in install scenarios)
* Rebased in such a way that the new postinst block is contained in one commit instead of being spread over 3
* New smoke test for `/public-key` route that basically just checks the apparmor permission is in place.

molecule/testinfra/app/test_smoke.py
```diff
@@ -11,14 +11,15 @@
...
 @pytest.mark.parametrize(
-    ("name", "url", "curl_flags"),
+    ("name", "url", "curl_flags", "expected"),
     [
         # We pass -L to follow the redirect from / to /login
-        ("journalist", "http://localhost:8080/", "L"),
-        ("source", "http://localhost:80/", ""),
+        ("journalist", "http://localhost:8080/", "L", "Powered by"),
+        ("source", "http://localhost:80/", "", "Powered by"),
+        ("source", "http://localhost:80/public-key", "", "-----BEGIN PGP PUBLIC KEY BLOCK-----"),
```
cfm, line 19: Nice.

securedrop/tests/conftest.py
```diff
@@ -91,14 +89,11 @@ def setup_journalist_key_and_gpg_folder() -> Generator[Tuple[str, Path], None, N
...
+        # it anyways to match a legacy prod keyring.
+        gpg.import_keys(journalist_public_key)
 
-        # Reduce source GPG key length to speed up tests at the expense of security
-        with mock.patch.object(
-            EncryptionManager, "GPG_KEY_LENGTH", PropertyMock(return_value=1024)
-        ):
-
-            yield journalist_key_fingerprint, tmp_gpg_dir
+        yield "65A1B5FF195B56353CC63DFFCC40EF1228271441", tmp_gpg_dir
```
cfm: Nit: Why hard-code this rather than print the fingerprint of the key newly imported into the keyring?

cfm (changes requested):
...of course it was AppArmor!  Thanks for fixing that so quickly, @legoktm.

I've now run through the test plan successfully!  I've left ~a couple of~ one nit~s~ inline for your consideration, plus:
* [x] 4060c6660bd37ad3f10326b943abdfacf45d1b0a is unsigned.
    * **Edit:**  Oh, this is unsigned because of my `Co-authored-by` line.  GitHub [suggests](https://docs.github.com/en/authentication/managing-commit-signature-verification/displaying-verification-statuses-for-all-of-your-commits) that we should be able to get this to "partially verified", but I won't lose sleep over it.

Otherwise I'm very happy with this.

cfm:
Discussed this morning:  If either @lsd-cat or @zenmonkeykstop has time next week for a security-focused second review, we'll do that, since I've focused so far on the functionality and the upgrade path.  Otherwise I'll approve and merge once <https://github.com/freedomofpress/securedrop/pull/6892#pullrequestreview-1614357017> is addressed.

securedrop/tests/conftest.py
legoktm: IIRC I was debugging something that wasn't working and hardcoded it to avoid one potential issue and didn't go back and undo it. Will revert it.

cfm:
Next week @lsd-cat will do a security review of:
1. the `redwood` API called by `securedrop.encryption.EncryptionManager`; and
2. `redwood`'s creation of new keys via Sequoia.

lsd-cat:
I did a limited review based on the limited capacity I had and my limited experience with both rust and sequoia. I mostly checked that the calls to the sequoia API were OK and that overall the encryption process looks good.
I have still to go over the filesystem/permissions part, but I'll get back to it this week.

cfm:
Thanks for <https://github.com/freedomofpress/securedrop/pull/6892#issuecomment-1734380460>, @lsd-cat.  We decided in today's stand-up that we'll go ahead and merge this, although further review and any follow-up feedback are of course welcome.

cfm (approved):

PR #6894, opened 2023-07-24
Remove obsolete "Extract tor version" ansible tasks

legoktm:
## Status

Ready for review

## Description of Changes

The check for >= 0.3.4.9 has been obsolete ever since the focal switch, which ships with tor 0.4.2.7. It's not that useful anyways, since apt should pick the higher version regardless. If we really wanted to ensure that our rehosted tor package is installed, we should configure an apt pin.

And then dumping to `.tor_version` is presumably unused since we no longer use AWS for SecureDrop testing (related: #6890). In any case, we'd be better served by dumping `apt show tor` or something along those lines.
## Testing

* [ ] Visual review

## Deployment

## Checklist


zenmonkeykstop (approved):
LGTM based on visual review.

PR #6912, opened 2023-08-03
Add tests for empty or malformed recipient gpg key

### LINKED ISSUE #6889: Need Sequoia/redwood test for PGP encryption if no recipients are specified
In Python we have:

```python
        # When trying to encrypt some data without providing any recipient
        # It fails and the right exception is raised
        with pytest.raises(GpgEncryptError) as exc:
            encryption_mgr._encrypt(
                using_keys_with_fingerprints=[],
                plaintext_in="test",
                ciphertext_path_out=tmp_path / "encrypt_fails",
            )
        assert "no terminal at all requested" in str(exc)
```

We should have a similar test (in Rust) that verifies the behavior if no recipients are specified to `redwood::encrypt()`.

rocodes:
## Status

Ready for review 

## Description of Changes

Fixes #6889.

Add tests for cases of missing or malformed journalist key

Visual review/CI passes

## Deployment 

n/a

- [x] Linting (`make lint`) and tests (`make test`) pass in the development container (`make rust-lint`, `make rust-test`)

Choose one of the following:

### DIFF
--- a/redwood/src/lib.rs
+++ b/redwood/src/lib.rs
@@ -244,4 +244,34 @@ mod tests {
         // Verify message is what we put in originally
         assert_eq!("Rust is great 🦀", &plaintext);
     }
+
+    #[test]
+    fn test_encryption_missing_malformed_recipient_key() {
+        // Bad fingerprints can be: empty, empty string, or malformed
+        let bad_keys: Vec<(Vec<String>, &str)> = vec![
+            (
+                vec![],
+                "OpenPGP error: Invalid operation: \
+                Neither recipients, passwords, nor session key given",
+            ),
+            (
+                vec!["".to_string()],
+                "OpenPGP error: Malformed Cert: No data",
+            ),
+            (
+                vec!["DEADBEEF0123456".to_string()],
+                "OpenPGP error: unexpected EOF",
+            ),
+        ];
+        let tmp: NamedTempFile = NamedTempFile::new().unwrap();
+        for (key, error) in bad_keys {
+            let err = encrypt_message(
+                key, // missing or malformed recipient key
+                "Look ma, no key".to_string(),
+                tmp.path().to_path_buf(),
+            )
+            .unwrap_err();
+            assert_eq!(err.to_string(), error);
+        }
+    }
 }


legoktm (commented):
Awesome! Tests and errors look good. Since most of the code is identical across the tree tests, do you want to consolidate it into one test function that iterates over the 3 cases? There's no fancy parameterization in Rust like pytest has so it'd just be iterating over a vec.

redwood/src/lib.rs
```diff
@@ -244,4 +245,23 @@ mod tests {
         // Verify message is what we put in originally
         assert_eq!("Rust is great 🦀", &plaintext);
     }
+
+    #[test]
+    fn test_encryption_missing_malformed_recipient_key() {
+        let map: HashMap<Vec<String>, &str> = HashMap::from([
+            (vec![], "OpenPGP error: Invalid operation: Neither recipients, passwords, nor session key given"), // missing key
```
rocodes: cargo fmt did not have anything to say about this very long line...

redwood/src/lib.rs
```diff
@@ -244,4 +244,34 @@ mod tests {
...
+                vec!["".to_string()],
+                "OpenPGP error: Malformed Cert: No data",
+            ),
+            (
+                vec!["DEADBEEF0123456".to_string()],
+                "OpenPGP error: unexpected EOF",
+            ),
+        ];
+        let tmp: NamedTempFile = NamedTempFile::new().unwrap();
+        for bad_key in bad_keys {
```
legoktm: One small nitpick that is hopefully useful for learning purposes: you can unpack the tuple while iterating, so this could be:
```rust
for (recipients, expected_err) in bad keys {
    ...
}
```

instead of having to refer to them as `.0` or `.1`.

legoktm (dismissed):
One tiny note if you want to change it, otherwise LGTM!

redwood/src/lib.rs
rocodes: tis indeed an improvement, thank you for all the back and forth on this, it has been super helpful :)

rocodes: (I'm going to fix it and push cause I'm a weirdo perfectionist)

legoktm (approved):
Awesome 🚀

PR #6922, opened 2023-08-16
chore: factor out minimal, specialized template for changes to production dependencies

cfm:
## Status

Ready for review

## Description of Changes

We're good at filling in the diff review for meaningful updates to production dependencies, less good at leaving good breadcrumbs for why we choose to silence certain alerts for them.  Factoring out a specialized template for the latter is an experiment to try prompting specifically for the relevant information, with the bonus of simplifying the template for everything else.

## Testing

Do we want this?  Do we like this?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE/dependencies.md
+++ b/.github/PULL_REQUEST_TEMPLATE/dependencies.md
@@ -1,0 +1,43 @@
+---
+name: Dependency change
+about: Update a production dependency or silence an alert
+---
+
+## Status
+
+Ready for review / Work in progress
+
+## Description of Changes
+
+<!-- Fill out the appropriate "###" section below and delete the others. -->
+
+### If you've added or updated a production code dependency:
+
+Production code dependencies are defined in:
+
+- `admin/requirements.in`
+- `admin/requirements-ansible.in`
+- `securedrop/requirements/python3/requirements.in`
+
+If you changed another `requirements.in` file that applies only to development
+or testing environments, then no diff review is required, and you can open
+a regular pull request.
+
+Choose one of the following:
+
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE/dependencies.md
```diff
@@ -0,0 +1,43 @@
+---
```
zenmonkeykstop, line 1: There's the potential for confusion where feature work also involves adding or updating a dependency (eg. do I need 2 prs? if so, how do they get applied so said feature is testable?) so I'd be inclined to say that splitting out to a separate template isn't merited.

.github/PULL_REQUEST_TEMPLATE/pull_request_template.md
```diff
@@ -47,20 +52,3 @@ Choose one of the following:
 - [ ] I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-docs) for these changes, or will do so later
 - [ ] I would appreciate help with the documentation
 - [ ] These changes do not require documentation
-
-### If you added or updated a production code dependency:
```
zenmonkeykstop, line 51: It might make sense to restore this section and retitle it something like "If you added/updated a reference to a production code dependency", and then have a clause like "If you silenced an alert related to a dependency, please provide a justification for doing so for review purposes:"

zenmonkeykstop (changes requested):

cfm:
We discussed this feedback yesterday and decided that it argues for closing this pull request in favor of another, smaller one (forthcoming) that just adds these considerations to `PULL_REQUEST_TEMPLATE.md`.

PR #6925, opened 2023-08-17
Ensure each certificate has at least one supported key before encrypting

### LINKED ISSUE #6895: redwood should error on unsupported encryption keys
We currently have this code in redwood:
```rust
    // For each of the recipient certificates, pull the encryption keys that
    // are compatible with by the standard policy (e.g. not SHA-1) supported by
    // Sequoia (duh), and not revoked.
    for cert in certs.iter() {
        for key in cert
            .keys()
            .with_policy(p, None)
            .supported()
            .alive()
            .revoked(false)
        {
            recipient_keys.push(key);
        }
    }
```

If you're encrypting to multiple certs (e.g. a journalist reply which is encrypted to the source key and journalist key), but one of the keys is unsupported by the default Sequoia policy (e.g. RSA 1024-bit), then it'll silently be skipped. 

So as we iterate through each cert, we should make sure the cert adds at least one key to recipient_key.

I discovered this because we historically have generated 1024-bit keys for tests for speed, except Sequoia correctly rejects them, so then messages were only being encrypted to the journalist key and therefore not able to be decrypted by the source.

rocodes:
## Status

Ready for review (ready for initial review feedback - leaving in draft mode for that reason) 

## Description of Changes

Fixes #6895 

- Before encryption, ensure each certificate contains at least one supported key (according to the specified policy, currently the [StandardPolicy](https://docs.sequoia-pgp.org/sequoia_openpgp/policy/struct.StandardPolicy.html)).

- Add test to ensure legacy (1024-bit RSA) key is rejected

## Testing

- Visual review +/ code structure review
- Tests are passing and test cases are accounted for

## Deployment

n/a

- [x] Linting (`make lint`) and tests (`make test`) pass in the development container **`make rust-lint`, `make rust-test`**

[...]

Choose one of the following:


redwood/src/lib.rs
```diff
@@ -195,11 +209,35 @@ pub fn decrypt(
...
+        R2q2qxmwQ8e7ABEBAAGItgQYAQoAIBYhBFbO0Jy+SeQ+Qtc9NeUVm4kgzl8xBQJk
+        3WY+AhsMAAoJEOUVm4kgzl8xdQYD/3ShTd2KM0XS25gMzfTQ8NtBS+jBvR0/+Eej
+        0BPhAf57pvNs8Y/uxUPO6l1+KrQek8XEnvGAkNBMD0wd2xPoQzcQqnf0iMcPhAY2
+        wCYRRONffypz3PkTM/JsFH08i2/UsntkgCLbgCEX/0Od1OFJT8VOSIOZe8EVjlGz
+        UUAkCN7u
+        =9bsJ
+        -----END PGP PUBLIC KEY BLOCK-----";
+    const BAD_KEY_FINGERPRINT: &str =
+        "56CED09CBE49E43E42D73D35E5159B8920CE5F31";
+
```
rocodes: This seems...kinda ugly, so I'm open to other ways of doing it, but I wasn't sure the best way to include an unsupported key, since the CertBuilder doesn't have an option for unsupported CipherSuites and I got a bit into the weeds looking at how to generate certs that we don't actually want to support :)

redwood/src/lib.rs
```diff
@@ -217,6 +255,51 @@ mod tests {
...
+                .with_policy(&StandardPolicy::new(), None)
+                .supported()
+                .alive()
+                .revoked(false)
+                .for_storage_encryption()
+            {
+                actual_supported_keys += 1;
+            }
+            assert_eq!(expected_supported_keys, actual_supported_keys);
+        }
```
rocodes: (It may be fine to just skip to testing `encrypt_message` as a whole, in which case this test function can be shorter n sweeter)

redwood/src/lib.rs
legoktm: I think just hardcoding the key is the right way to go, but I'd recommend just storing it in a file and then using the `include_str!()` macro (https://doc.rust-lang.org/std/macro.include_str.html) to set it as a constant. Plus that would make it simpler to share one bad SHA-1 key here and for your other admin+sq-keyring-linter PR.

redwood/src/lib.rs
```diff
@@ -131,18 +136,27 @@ fn encrypt(
...
             .supported()
             .alive()
             .revoked(false)
             .for_storage_encryption()
-        {
-            recipient_keys.push(key);
+            .collect();
+
+        // Each certificate must have at least one supported encryption key
+        match keys.len() {
```
legoktm: I think using an if/else statement here is clearer. IMO match is better when you're comparing enum types or have multiple if/elseif/elseif/else type blocks.

redwood/src/lib.rs
```diff
@@ -32,6 +34,9 @@ pub enum Error {
     Io(#[from] io::Error),
     #[error("Unexpected non-UTF-8 text: {0}")]
     NotUnicode(#[from] FromUtf8Error),
+    // Distinct from OpenPgp error because it's about our own logic
+    #[error("Encryption Error: {0}")]
+    Encryption(String),
```
legoktm: In Rust the convention is to have a different type for each possible error. So instead of an `Encryption(String)` type with a message that describes the specific issue, we could just have `NoSupportedKeys`, and put the "At least one supported key is needed per certificate" from below into the error message itself. 

(Note that I would suggest doing `NoSupportedKeys(String)` so we can include the fingerprint, see below)

redwood/src/lib.rs
```diff
@@ -131,18 +136,27 @@ fn encrypt(
 
     // For each of the recipient certificates, pull the encryption keys that
     // are compatible with by the standard policy (e.g. not SHA-1) supported by
-    // Sequoia (duh), and not revoked.
+    // Sequoia, and not revoked.
     // These filter options should be kept in sync with `Helper::decrypt()`.
     for cert in certs.iter() {
-        for key in cert
+        let keys: Vec<ValidErasedKeyAmalgamation<PublicParts>> = cert
```
legoktm: I think Rust should be able to figure out the inner type, so in theory you should just need to write `let keys: Vec<_> = cert...collect()` (Nothing wrong with spelling out the full type if you want, I just find stuff like `ValidErasedKeyAmalgamation<PublicParts>` annoying to write so I try to skip them)

redwood/src/lib.rs
```diff
@@ -131,18 +136,27 @@ fn encrypt(
...
             .for_storage_encryption()
-        {
-            recipient_keys.push(key);
+            .collect();
+
+        // Each certificate must have at least one supported encryption key
+        match keys.len() {
+            0 => {
+                return Err(Error::Encryption(
+                    "At least one supported key is needed per certificate"
```
legoktm: Can we pass the cert's fingerprint here so for messages encrypted to multiple keys we can figure out which one is at fault?

redwood/src/lib.rs
legoktm: I would actually do the opposite, *just* test `encrypt_message` correctly fails instead of trying to count the number of supported keys, which requires duplicating the supported keys logic. If we want to test the underlying iterator then I think we should be splitting it out to a separate function.

legoktm (changes requested):
Functionally looks good, comments are mostly about Rust stuff.

rocodes:
Thanks so much for all your helpful feedback @legoktm  - I think I've addressed everything in 5aff332

legoktm (approved):
Awesome!

PR #6926, opened 2023-08-17
Tests to check key creation date (2013-05-14) and expiration (none)

### LINKED ISSUE #6888: Need Sequoia/redwood test that PGP key creation uses 2013-05-14 date
For the gpg-based EncryptionManager, we have a test that verifies new keys are created using the special 2013-05-14 date:
```python
        # And the key has a hardcoded creation date to avoid leaking information about when sources
        # first created their account
        source_key_details = encryption_mgr._get_source_key_details(source_user.filesystem_id)
        assert source_key_details
        creation_date = _parse_gpg_date_string(source_key_details["date"])
        assert creation_date.date() == EncryptionManager.DEFAULT_KEY_CREATION_DATE
```

We should have a test (in Rust) that verifies the creation date we set is properly respected.

rocodes:
## Status

Ready for review / ready for initial review feedback, so draft mode

## Description of Changes

Fixes #6888 

Add tests to check key creation date and ensure keys do not expire

## Testing

- Visual review
- CI passing
 
## Deployment

n/a 

- [x] Linting (`make lint`) and tests (`make test`) pass in the development container (`make rust-lint`, `make rust-test`)

Choose one of the following:

### DIFF
--- a/redwood/src/lib.rs
+++ b/redwood/src/lib.rs
@@ -50,6 +50,11 @@ impl From<Error> for PyErr {
 
 type Result<T> = std::result::Result<T, Error>;
 
+// All reply keypairs will be "created" on the same day SecureDrop (then
+// Strongbox) was publicly released for the first time, 2013-05-14
+// https://www.newyorker.com/news/news-desk/strongbox-and-aaron-swartz
+const KEY_CREATION_SECONDS_FROM_EPOCH: u64 = 1368507600;
+
 /// A Python module implemented in Rust.
 #[pymodule]
 fn redwood(py: Python, m: &PyModule) -> PyResult<()> {
@@ -73,12 +78,11 @@ pub fn generate_source_key_pair(
         .set_cipher_suite(CipherSuite::RSA4k)
         .add_userid(format!("Source Key <{}>", email))
         .set_creation_time(
-            // All reply keypairs will be "created" on the same day SecureDrop (then
-            // Strongbox) was publicly released for the first time.
-            // https://www.newyorker.com/news/news-desk/strongbox-and-aaron-swartz
+            // All reply keypairs will be "created" on the same day, 2013-05-14
             SystemTime::UNIX_EPOCH
-                // Equivalent to 2013-05-14
-                .checked_add(Duration::from_secs(1368507600))
+                .checked_add(Duration::from_secs(
+                    KEY_CREATION_SECONDS_FROM_EPOCH,
+                ))
                 // unwrap: Safe because the value is fixed and we know it won't overflow
                 .unwrap(),
         )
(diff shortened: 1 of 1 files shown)


redwood/src/lib.rs
```diff
@@ -217,6 +217,39 @@ mod tests {
...
+    fn test_source_key_creation_expiry() {
+        let (public_key, _secret_key, _fingerprint) =
+            generate_source_key_pair(PASSPHRASE, "foo@example.org").unwrap();
+        let cert = Cert::from_str(&public_key).unwrap();
+        let standard_creation_time = SystemTime::UNIX_EPOCH
+            // Equivalent to 2013-05-14
+            .checked_add(Duration::from_secs(1368507600))
+            // unwrap: Safe because the value is fixed and we know it won't overflow
+            .unwrap();
+        assert_eq!(
```
legoktm, line 277: I would suggest factoring out the common code here and below like:
```rust
let sig = cert.primary_key().self_signatures().nth(0).unwrap();

assert_eq!(sig.signature_creation_time(), ...);
```

redwood/src/lib.rs
```diff
@@ -217,6 +217,39 @@ mod tests {
...
+                .self_signatures()
+                .nth(0)
+                .unwrap()
+                .signature_creation_time(),
+            Some(standard_creation_time)
+        );
+
+        // From the documentation:
+        // This function is called signature_validity_period and not signature_expiration_time
+        // [because] the time is actually relative to the signature's creation time, which
```
legoktm, line 284: nit: I assume `[because]` is in brackets because in the upstream docs its a link, but since we don't have the link, remove the brackets so it's slightly easier to read?

redwood/src/lib.rs
```diff
@@ -217,6 +217,39 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_source_key_creation_expiry() {
+        let (public_key, _secret_key, _fingerprint) =
+            generate_source_key_pair(PASSPHRASE, "foo@example.org").unwrap();
+        let cert = Cert::from_str(&public_key).unwrap();
+        let standard_creation_time = SystemTime::UNIX_EPOCH
+            // Equivalent to 2013-05-14
+            .checked_add(Duration::from_secs(1368507600))
```
legoktm: Since we're using the `1368507600` magic number twice now, let's refactor it into a constant?

legoktm (changes requested):
Nice! Left two suggestions.

redwood/src/lib.rs
rocodes: Oh, it's because I elided the documentation and am paraphrasing, but maybe I'm just being pedantic and annoying

legoktm (approved):
Awesome :)

PR #6929, opened 2023-08-23
chore: add the "silencing an alert" case to the checklist

cfm:
## Status

Ready for review

## Description of Changes

We're good at filling in the diff review for meaningful updates to production dependencies, less good at leaving good breadcrumbs for why we choose to silence certain alerts for them.  Let's ask for that justification in the checklist.

This is the minimized version of #6922.

## Testing

~Do we want this?~ (Yes, according to #6922.)  Do we like this?


zenmonkeykstop (approved):
LGTM!

Issue #6930, opened 2023-08-24
`make typelint` fails locally due to path interpolation


cfm:
## Description

After c636c14a6c52dae04919b8bd766141f9dd07bc67, `make typelint` works swimmingly in CI but fails locally.

## Steps to Reproduce

```sh-session
user@sd-dev:~/securedrop$ rm -rf .venv
user@sd-dev:~/securedrop$ git checkout develop
Switched to branch 'develop'
Your branch is up to date with 'origin/develop'.
user@sd-dev:~/securedrop$ make venv
user@sd-dev:~/securedrop$ make typelint
```

## Expected Behavior

Mypy magic.

## Actual Behavior

```sh-session
user@sd-dev:~/securedrop$ make typelint
███ Running mypy type checking...
./bin/run-mypy: line 10: ./bin/dev-deps: No such file or directory
make: *** [Makefile:131: typelint] Error 1
```

## Comments

I haven't yet found a fix that retains:

https://github.com/freedomofpress/securedrop/blob/960c43b47a4a0a90e261d2206fe4aa9375622e25/securedrop/bin/run-mypy#L10-L11

legoktm:
Works for me...

```
user@dev ~/g/f/securedrop> git checkout develop
Switched to branch 'develop'
Your branch is up to date with 'origin/develop'.
user@dev ~/g/f/securedrop> git status
On branch develop
Your branch is up to date with 'origin/develop'.

nothing to commit, working tree clean
user@dev ~/g/f/securedrop> rm -rf .venv
user@dev ~/g/f/securedrop> make venv
git config core.hooksPath ".githooks"
███ Preparing Python 3 virtual environment...
Creating /usr/bin/python3.8 virtualenv in .venv
created virtual environment CPython3.8.17.final.0-64 in 302ms
  creator CPython3Posix(dest=/home/user/github/freedomofpress/securedrop/.venv, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(extra_search_dir=/usr/share/python-wheels,download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/user/.local/share/virtualenv)
    added seed packages: pip==22.3.1, setuptools==66.1.1, wheel==0.38.4
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator
Make sure to run: source .venv/bin/activate

user@dev ~/g/f/securedrop> make typelint
[...]

PR #6932, opened 2023-08-24
Have ruff apply to securedrop/scripts/ too

legoktm:
## Status

Ready for review

## Description of Changes

These need explicit mentioning since they don't have a `.py` extension.

The E402 error now applies to the imports following the sys.path.append() call, so move them around.

KOG Update: 
  - added a fix for #6930 as `make lint` was failing locally
  - updated the precommit hook to stay limited to `.py` files, as ruff got indigestion trying to lint a bash script.
  - recusing myself from review as a result :(

## Testing

* [ ] Visual review
* [ ] CI passes

## Checklist

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 # shellcheck disable=SC2086
 
-PY_FILES=$(git diff --name-only --cached --diff-filter=ACMR | grep ".py$")
+PY_FILES=$(git diff --name-only --cached --diff-filter=ACMR | grep "\.py$")
 
 set -eo pipefail
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,6 +67,7 @@ ignore = [
     # nested if and with statements, too many violations for now
     "SIM102", "SIM117",
 ]
+extend-include = ["securedrop/scripts/*"]
 
 [tool.ruff.isort]
 # ruff's isort is smart enough to know these are first party, but let's treat
--- a/securedrop/bin/run-mypy
+++ b/securedrop/bin/run-mypy
@@ -7,9 +7,8 @@ REPOROOT=$(git rev-parse --show-toplevel)
 cd "${REPOROOT}"
 
 if [ "$(command -v mypy)" ]; then
-    source "${BASH_SOURCE%/*}/dev-deps"
+    source "securedrop/bin/dev-deps"
     build_redwood
-
     mypy ./securedrop ./admin --namespace-packages --explicit-package-bases "$@"
 elif [ -d "/opt/venvs/securedrop-app-code/" ]; then
     # Inside the dev container, but no mypy
--- a/securedrop/scripts/rqrequeue
+++ b/securedrop/scripts/rqrequeue
@@ -6,10 +6,10 @@ import logging
 import sys
 import time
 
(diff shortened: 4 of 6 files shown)


legoktm:
> updated the precommit hook to stay limited to `.py` files, as ruff got indigestion trying to lint a bash script.

This shouldn't be necessary, running `ruff check .` should only lint Python files already, since that's what `make ruff` does...

legoktm:
I edited the commit message to fix the issue. I kept the precommit hook commit, but I think it should be dropped or needs more detail on what went wrong.

zenmonkeykstop:
Oh - iirc it was trying to lint the `run-mypy` bash script (which was also a problem due to a bad regex in the original hook).

rocodes:
I'm also mildly confused-- reverting to [3f4353e](https://github.com/freedomofpress/securedrop/pull/6932/commits/3f4353e2f1556a718e833e8bb3ddbe8be3ad368a) (so no changes to the pre-commit hook) also works just fine for me, and `make typelint`, `make lint`, `make rust-lint` etc are all perfectly happy without that change.

zenmonkeykstop:
> I'm also mildly confused-- reverting to [3f4353e](https://github.com/freedomofpress/securedrop/pull/6932/commits/3f4353e2f1556a718e833e8bb3ddbe8be3ad368a) (so no changes to the pre-commit hook) also works just fine for me, and `make typelint`, `make lint`, `make rust-lint` etc are all perfectly happy without that change.

The error is only triggered if there's a change in `run-mypy`. Try:
- reverting the commit with the hook change
- adding a trivial change (`# i'm a comment!`) in `securedrop/bin/run-mypy`
- adding and committing it

It's happening because the hook only checks anything if there is a match for `*py` files, and ruff is trying to parse `run-mypy` as a .py file. Changing the regex to match `*.py` and running ruff on only matching files solves that. (Yes it's super edge-casey.)

legoktm:
Thanks for the details:

> and ruff is trying to parse `run-mypy` as a .py file

This is the real problem IMO because it means that even `make ruff` could trigger it. So I'll look into fixing that filtering and having the git hook continue to call plain `ruff check .`

legoktm:
@zenmonkeykstop what's the output of `ruff check . --show-files` on your machine? On mine it doesn't include `run-mypy` so ruff really shouldn't be linting it...

rocodes:
(I'm not kev, but I am chiming in to say that `run-mypy` is also not in my list of files checked by ruff per `ruff check . --show-files`)

zenmonkeykstop:
I'm so dumb - the failure comes from the `black` check  immediately beforehand in the hook, which does take a list of files. Sooo the regex still needs fixing for that case, but the resultant list of files is fed to black, not ruff.

rocodes:
(Ready for re-review)

.githooks/pre-commit
```diff
@@ -14,7 +14,7 @@ if [[ -n "$PY_FILES" ]]; then
     echo "$PY_FILES"
     # Run black against changed python files for this commit
     black --check --diff ${PY_FILES//$'\n'/ }
-    # Run ruff (against all files, it's fast enough)
+    # Run ruff against changed python files
```
legoktm: This comment should be left as it was originally

legoktm (commented):
LGTM! (can't approve my own PR)

zenmonkeykstop (approved):
also LGTM

PR #6935, opened 2023-08-25
Fix demo build

legoktm:
## Status

Ready for review

## Description of Changes

Build the redwood wheel during the container build instead of when launching the container. Copy some of the rustup logic into the DemoDockerfile and use the same base image version as the current dev environment.

This is hopefully good enough for now, long-term we should switch to server nightlies that are just installed into the image.

## Testing

* [ ] `podman build -f securedrop/dockerfiles/focal/python3/DemoDockerfile . -t sd-demo`
* [ ] `podman run --rm -it -p 8080:8080 sd-demo`
* [ ] Visit localhost:8080 in your browser, create a new source and send a message (invoking redwood encryption with no errors)

## Deployment

Any special considerations for deployment? Affects demo instance.

### DIFF
--- a/securedrop/bin/dev-deps
+++ b/securedrop/bin/dev-deps
@@ -158,13 +158,17 @@ function reset_demo() {
 }
 
 function build_redwood() {
-    # Create a debug build of redwood and install it in the app-code virtualenv
-    # TODO: add live reload support
-    # We tell cargo to put its registry/crate cache and build cache in the target/ folder so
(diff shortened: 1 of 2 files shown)


zenmonkeykstop (approved):
- created sd-demo container
- fired it up with `-p 8080:8080 -p 8081:8081` to allow for testing the whole flow (probably unnecessarily :) )
- all good, no errors.

PR #6946, opened 2023-09-21
Migrate public keys from GPG to database-backed storage

### LINKED ISSUE #6800: One-time migration of sources' public keys and fingerprints from gpg to database-backed storage

legoktm:
## Status

Ready for review, depends on ~~#6892 and~~ #6948.

## Description of Changes

* Add --gpg to loaddata.py to create GPG-backed sources (primarily for testing/dev purposes)
* Migrate public keys from GPG to database-backed storage (the main commit in this PR)

Commit messages have more detail on each.

Fixes #6800.

## Testing

* [x] Start the dev server (`make dev`) and open a shell into the container, e.g. `podman exec --user=root -it $(podman ps --filter name=securedrop --format '{{.ID}}') bash`
* [x] Run `alembic stamp 811334d7105f` (ID of the migration before this one)
* [x] Run `./loaddata.py --gpg` to add some GPG sources
* [x] Run `select filesystem_id, pgp_fingerprint from sources;` in SQLite to verify there are some sources without a fingerprint set in the database
* [x] Run `alembic upgrade head`, completes successfully. Then run the database query again to see that all fingerprints have been populated.
* [x] Obligatory CI passes & visual review

## Deployment

This is a non-destructive migration. There is some risk the migration fails, which would halt package upgrade, but the main point where we expect that to happen (`gpg.export_key()`) is wrapped in a try/except.

## Checklist


securedrop/alembic/versions/17c559a7a685_pgp_public_keys.py
```diff
@@ -0,0 +1,80 @@
...
+            ).bindparams(filesystem_id=filesystem_id)
+        ).first()
+        if result != (None, None):
+            # Either not in the database or there's already some data in the DB.
+            # In any case, skip.
+            continue
+        fingerprint = keyinfo["fingerprint"]
+        public_key = gpg.export_keys(fingerprint)
+        if not public_key.startswith("-----BEGIN PGP PUBLIC KEY BLOCK-----"):
+            # FIXME: can we have a stronger well-formedness check here?
```
legoktm: We'll eventually rebase on top of https://github.com/freedomofpress/securedrop/pull/6948 and use it here.

legoktm:
This is ready for review now, but same caveat that it depends on https://github.com/freedomofpress/securedrop/pull/6892 and https://github.com/freedomofpress/securedrop/pull/6948.

legoktm:
This is ready for review now, but same caveat that it depends on https://github.com/freedomofpress/securedrop/pull/6892 and https://github.com/freedomofpress/securedrop/pull/6948.

securedrop/alembic/versions/17c559a7a685_pgp_public_keys.py
```diff
@@ -0,0 +1,92 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    """
+    Migrate public keys from the GPG keyring into the SQLite database
+
+    We iterate over all the secret keys in the keyring, identify the
+    corresponding source, export the key and save it in the database.
```
cfm:
```suggestion
    We iterate over all the secret keys in the keyring and see if we
    can identify the corresponding Source record.  If we can, and it
    doesn't already have key material migrated, export the key and
    save it in the database.
```

I don't mean this nitpickily.  After #6948, much of this logic feels analogous to the new `is_valid_public_key()`.  The assumptions here seem to be that:
1. On the one hand, the set of secret keys may be larger than the set of `Source` records.
2. On the other hand, the set of `Source` records is much easier to query than the set of secret keys.

So, iterate over the set of secret keys, and for any given secret key be prepared to bail out it doesn't meet the current (but not currently specified!) criteria to be attached to a `Source`.

I think that approach is sound, just worth documenting here a bit more explicitly—thus the suggestion.

securedrop/alembic/versions/17c559a7a685_pgp_public_keys.py
```diff
@@ -0,0 +1,92 @@
...
+        if result != (None, None):
+            # Either not in the database or there's already some data in the DB.
+            # In any case, skip.
+            continue
+        fingerprint = keyinfo["fingerprint"]
+        try:
+            public_key = gpg.export_keys(fingerprint)
+        except:  # noqa: E722
+            traceback.print_exc()
+            continue
```
cfm: If we think this actually might fail on a production instance, then we should consider:
1. What state do we want the instance to be in?
2. What action, if any, do we want the administrator to (have to) take?

Concretely, even if we want the rest of the migration to continue, do we want the upgrade to exit nonzero and require some kind of recovery intervention before it can ultimately pass?

I suspect the answer is probably no, and I'm fine with that.  But I'm asking now rather than waiting for v2.7 QA.  :-)

cfm (changes requested):
This looks great, @legoktm.  I've taken the liberty of being a little aggressive in my testing just to make sure we're hitting the code-paths you intend.  :-)

I've left a couple of questions inline, but otherwise I'll be happy to approve and merge this once it's rebased from `develop` after #6892.

* [x] Start the dev server (`make dev`) and open a shell into the container, e.g. `podman exec --user=root -it $(podman ps --filter name=securedrop --format '{{.ID}}') bash`
* [x] Run `alembic stamp 811334d7105f` (ID of the migration before this one)
* [x] Run `./loaddata.py --gpg` to add some GPG sources
* [x] Run `select filesystem_id, pgp_fingerprint from sources;` in SQLite to verify there are some sources without a fingerprint set in the database
* [x] **My addition:** Choose one of the new sources and smoke-test a round trip: source submits; journalist replies
* [x] Run `alembic upgrade head`, completes successfully. Then run the database query again to see that all fingerprints have been populated.
* [x] **My addition, aggressive:** `gpg --homedir /var/lib/securedrop/keys --list-secret-keys`, then  `gpg --homedir /var/lib/securedrop/keys --delete-secret-key` each one
* [x] **My addition:** Choose the same source as before and smoke-test a round trip: source submits; journalist replies
* [x] Obligatory CI passes & visual review

securedrop/alembic/versions/17c559a7a685_pgp_public_keys.py
legoktm: SGTM!

Yeah, basically I was thinking of some case where the source was deleted out of the database and deletion out of the GPG keyring failed.

securedrop/alembic/versions/17c559a7a685_pgp_public_keys.py
legoktm: I changed my mind on this twice in the middle of this PR, so here's where I ended up, and what the code implements.

Our primary goal with Sequoiafication is to improve stability and reliability, especially as we look to primarily focus on SDW. This requires that the migration process also be robust, especially in the face of potential GPG errors that we're trying to eliminate going forwards.

So what happens if the migration fails with a non-zero exit code? dpkg will error out, meaning that some updates may be interrupted. It's not clear to me if it blocks further updates of other packages, but having an incompletely upgraded package just seems like a bad spot to be in because we can't assert or reason about what state the instance is in.

What if we had the migration complete with a zero exit code, but set some flag as failed for admin intervention? We would then have to ask the maintainer to run some `gpg` commands manually, inspecting the keyring for that specific source (or sources). I think this the opposite of what we want, because it means we have to spend time debugging gpg instead of forgetting about it!

So then we're left with our current implementation of doing the migration, but quietly failing on errors. This means that the migration will optimistically happen for each source, but if there are issues it'll be skipped on a case-by-case basis. The GPG fallback still exists in `Source.public_key`, so if it was a spurious failure of some kind, the source will still be operable and can receive replies, etc.

I took special care to not increase our footprint of pretty_bad_protocol (see also my attempt at tree shaking it: https://github.com/freedomofpress/securedrop/pull/6960), all the functions used are also used elsewhere:
* list_keys() - EncryptionManager._get_source_key_details()
* export_keys() - EncryptionManager._get_public_key()

This pre-existing usage informed how I chose to guard them - list_keys() is called for every source and also in startup in prime_keycache(), so if it was broken, the whole instance would be busted, and therefore I think it's safe to assume it's successful. But if export_key() was broken, you'd have to be looking at that specific source to know (prime_keycache would theoretically hit it, but it silently suppresses errors), so it's possible some very old source is in this case, but no one ever triggered it, and the migration would uncover it.

And because the usage of pbp is the same as the migration itself, there's very little to be gained (IMO) by removing the GPG fallbacks, versus the upside of the migration being buggy somehow but everything still working.

Hopefully that was not too rambley and makes sense. Also maybe there's somewhere better to document this design decision besides a CR comment?

securedrop/alembic/versions/17c559a7a685_pgp_public_keys.py
legoktm: Re-reading prime_keycache reminded me why I did it this way. And that's because it's impossible to do it any other way with our current GPG wrappers/abstractions.

If you look at prime_keycache, it does:
```python
        for source in Source.query.filter_by(pending=False, deleted_at=None).all():
            try:
                encryption_mgr.get_source_public_key(source.filesystem_id)
            except GpgKeyNotFoundError:
                pass
```

Which you think means its iterating over each source, and looking up the public key for each. But EncryptionManager._get_source_key_details() is iterating over the keyring each time before finding a matching key, (aka it's O(N²)). So we just iterate over the keyring once and matching sources / exporting those keys (O(N)).

legoktm:
Rebased and integrated is_valid_public_key() checks.

securedrop/alembic/versions/17c559a7a685_pgp_public_keys.py
cfm: Yeah, that was my intuition.  Thanks for confirming it based on the current implementation.

securedrop/alembic/versions/17c559a7a685_pgp_public_keys.py
cfm: Thanks for laying out your thinking here, @legoktm, and for implementing this so carefully.  I'm on board.  For me this is the fundamental decision:

> So then we're left with our current implementation of doing the migration, but quietly failing on errors. This means that the migration will optimistically happen for each source, but if there are issues it'll be skipped on a case-by-case basis. The GPG fallback still exists in `Source.public_key`, so if it was a spurious failure of some kind, the source will still be operable and can receive replies, etc.

I agree that this is worth highlighting and recording, so I've summarized it in <https://github.com/freedomofpress/securedrop/wiki/2.7.0-development-notes>.  Feel free to amend (or move elsewhere) as you see fit.

cfm:
Thanks for the discussion, @legoktm.  I'm happy with this, and I'll merge as soon as #6948 is in and this is rebased with it.

legoktm:
Rebased + signed.

cfm (approved):
Thanks, @legoktm!  We've resolved all our discussions from <https://github.com/freedomofpress/securedrop/pull/6946#pullrequestreview-1653571768>, so I'll merge as soon as CI passes.

PR #6948, opened 2023-09-22
redwood: Add is_valid_public_key function

### LINKED ISSUE #6947: Add Python function that verifies the given public key is valid
## Description

We have a non-trivial definition of what makes a "valid" PGP public key in the context of our Sequoia code (last expanded in 58168f78fc7ce820beb3166023bf53d4dbecd32f). We have multiple places it would be nice to verify that a public key is actually a valid public key.

The function would return the 40-character fingerprint, so something like:
```rust
pub fn is_valid_public_key(input: &str) -> Result<String> {}
```

```python
def is_valid_public_key(input: str) -> str: ...
```

In https://github.com/freedomofpress/securedrop/pull/6892/commits/d69b3485c2f359e8dcc127632feceb774fc59a5a I left a TODO:
```python
    # FIXME: we should be using redwood/Sequoia for this!!
    gpg = gnupg.GPG(
        binary="gpg2",
        homedir=tmpdir,
        options=["--pinentry-mode loopback", "--trust-model direct"],
    )
    result = gpg.import_keys(pubkey)
    if result.fingerprints == [args.fingerprint]:
        print("Success! Specified fingerprint matches pubkey file.")
        sys.exit(0)
    else:
        print(f"Failed! Key contains {result.fingerprints}, not {args.fingerprint}")
        sys.exit(1)
```

Also consider:
```
securedrop/tests/test_encryption.py
21:        assert journalist_pub_key.startswith("-----BEGIN PGP PUBLIC KEY BLOCK----")
33:        assert source_pub_key.startswith("-----BEGIN PGP PUBLIC KEY BLOCK----")

securedrop/tests/migrations/migration_17c559a7a685.py
69:            assert source[1].startswith("-----BEGIN PGP PUBLIC KEY BLOCK-----")

securedrop/tests/test_source.py
228:        assert "BEGIN PGP PUBLIC KEY BLOCK" in text
238:        assert "BEGIN PGP PUBLIC KEY BLOCK" in resp.data.decode("utf-8")

securedrop/tests/test_pretty_bad_protocol.py
25:    assert public_key.startswith("-----BEGIN PGP PUBLIC KEY BLOCK-----")

securedrop/tests/functional/test_source.py
108:        assert "BEGIN PGP PUBLIC KEY BLOCK" in response.content.decode("utf-8")

securedrop/alembic/versions/17c559a7a685_pgp_public_keys.py
57:        if not public_key.startswith("-----BEGIN PGP PUBLIC KEY BLOCK-----"):
```
(the last one in an alembic migration is from #6946)

legoktm:
## Status

Ready for review; ~~but will fail CI until #6892 lands (this will be rebased on top of that prior to merge)~~

## Description of Changes

We have multiple cases in Python code where we want to validate that a
string is valid PGP public key that Sequoia will accept. Exporting a
`is_valid_public_key` function will help make checks more robust and
simplify some tests.
    
The validation code on the Rust side is also a bit duplicated and
currently manually kept in sync with comments, so let's take the
opportunity to unify it into one file and de-duplicated with a macro
(hopefully not too magical).
    
Fixes #6947.

## Testing

* [ ] Visual review
* [ ] CI passes

## Deployment

## Checklist


redwood/src/keys.rs
```diff
@@ -0,0 +1,55 @@
...
+use sequoia_openpgp::cert::prelude::ValidErasedKeyAmalgamation;
+use sequoia_openpgp::packet::key::{PublicParts, SecretParts, UnspecifiedRole};
+use sequoia_openpgp::packet::Key;
+use sequoia_openpgp::policy::Policy;
+use sequoia_openpgp::Cert;
+
+/// We want to use the same iterators on public and secret keys but it's not
+/// really possible to do it in a function because of type differences so we
+/// use a macro instead.
+macro_rules! filter_keys {
```
legoktm, line 11: I spent a while trying to write a generic function that applied these filters over public key parts and secret key parts (distinct types) and 1) it was super complicated and 2) I couldn't actually get it to work/compile.

So I went with a macro, but I'm also okay saying macros are too magical for our collective current level of Rust and to just keep duplicating for now.

rocodes: (I vote leave it in - it is clear/readable, basically by the book, and less error-prone than manual duplication of the same logic in different places)

legoktm:
Marking as ready for review; CI will fail because SD tests are generating 1024-bit keys until #6892 lands. But at least the code can be reviewed ahead of time.

redwood/src/lib.rs
```diff
@@ -94,6 +98,19 @@ pub fn generate_source_key_pair(
     Ok((public_key, secret_key, format!("{}", cert.fingerprint())))
 }
 
+#[pyfunction]
+pub fn is_valid_public_key(input: &str) -> Result<String> {
+    let policy = &StandardPolicy::new();
```
rocodes: (Very minor point but) if the code becomes more complex or we ever stop using StandardPolicy, we may want to create one policy instance and pass it to `encrypt` and `is_valid_pubkey` to ensure there's no developer error and we're always testing and encrypting against the same policy.

legoktm: That's a good point, and since you mentioned it, I looked, `StandardPolicy::new()` is const, so we can in fact already do that! Commit incoming...

rocodes (dismissed):
LGTM, and very polite way of making improvements to my attempt at enforcing the key policy :see_no_evil:  

Waiting til https://github.com/freedomofpress/securedrop/pull/6928 lands but otherwise :rocket:

cfm (dismissed):
This is just a drive-by review after #6892 and before #6946.  Very nice addition and refactoring!  I'll defer to you two for the final rebase and merge.

rocodes (approved):

PR #6949, opened 2023-09-22
Add `cargo audit` job in CI

### LINKED ISSUE #6815: Dependency alerts are configured using RustSec/cargo audit
We should get notifications for security issues in Rust dependencies using the https://rustsec.org/ database and their cargo-audit tool.

Presumably this should be done using the same infrastructure/setup that our current Python alerts use.

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #6815.

- adds a `make rust-audit` target that installs and runs cargo-audit against the project
- adds an `audit.toml` config file with defaults+fail-on-yanked
- adds a `rust-security` job running on commits and nightlies using said target
- updates openssl and hermit-abi rust dependencies to make said job pass :)

## Testing

- [x] CI is passing
- [x] `make rust-audit` passes locally.

## Deployment

n/a

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,27 +1,12 @@
 [advisories]
 ignore = [] # advisory IDs to ignore e.g. ["RUSTSEC-2019-0001", ...]
-informational_warnings = ["unmaintained"] # warn for categories of informational advisories
-severity_threshold = "low" # CVSS severity ("none", "low", "medium", "high", "critical")
-
-# Advisory Database Configuration
-[database]
-path = "~/.cargo/advisory-db" # Path where advisory git repo will be cloned
-url = "https://github.com/RustSec/advisory-db.git" # URL to git repo
-fetch = true # Perform a `git fetch` before auditing (default: true)
-stale = false # Allow stale advisory DB (i.e. no commits for 90 days, default: false)
 
 # Output Configuration
 [output]
 deny = ["yanked", "unmaintained"]
-format = "terminal" # "terminal" (human readable report) or "json"
-quiet = false # Only print information on error
-show_tree = true # Show inverse dependency trees along with advisories (default: true)
+quiet = false
 
 # Target Configuration
 [target]
 arch = "x86_64" # Ignore advisories for CPU architectures other than this one
 os = "linux" # Ignore advisories for operating systems other than this one
-
-[yanked]
-enabled = true # Warn for yanked crates in Cargo.lock (default: true)
-update_index = true # Auto-update the crates.io index (default: true)
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -12,7 +12,7 @@ stale = false # Allow stale advisory DB (i.e. no commits for 90 days, default: f
 
 # Output Configuration
 [output]
-deny = ["unmaintained"] # exit on error if unmaintained dependencies are found
+deny = ["yanked", "unmaintained"]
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -195,6 +195,16 @@ semgrep:
 	@semgrep --exclude "securedrop/tests/" --error --strict --metrics off --max-chars-per-line 200 --verbose --config "p/r2c-security-audit" securedrop
 	@echo
 
+
+# check dependencies in Cargo.lock
+.PHONY: rust-audit
+rust-audit:
+	@echo "███ Running Rust dependency checks..."
+	@cargo install cargo-audit
+	@cargo audit
```
legoktm, line 204: I typically run this with `cargo audit -D warnings` so e.g. yanked crates also cause a failure. But now that I re-read the documentation, we can create a `audit.toml` file (see https://github.com/rustsec/rustsec/blob/main/cargo-audit/audit.toml.example) to have yanked and unmaintained crates cause failures, plus we can limit advisories to Linux x86_64 (and not get stuff about Windows, etc.).

Makefile
```diff
@@ -195,6 +195,16 @@ semgrep:
 	@semgrep --exclude "securedrop/tests/" --error --strict --metrics off --max-chars-per-line 200 --verbose --config "p/r2c-security-audit" securedrop
 	@echo
 
+
+# check dependencies in Cargo.lock
+.PHONY: rust-audit
+rust-audit:
+	@echo "███ Running Rust dependency checks..."
+	@cargo install cargo-audit
```
legoktm, line 203: I assume it is intentional that you didn't pin the version here, which I agree with.

legoktm (changes requested):
Can you update the Rust deps with just `cargo update -p openssl` for now so we're not upgrading everything in this?

Makefile
zenmonkeykstop: *pulls collar nervously*

Makefile
zenmonkeykstop: added, and bumped `hermit-abi` as a result.

.cargo/audit.toml
```diff
@@ -0,0 +1,27 @@
...
+# Advisory Database Configuration
+[database]
+path = "~/.cargo/advisory-db" # Path where advisory git repo will be cloned
+url = "https://github.com/RustSec/advisory-db.git" # URL to git repo
+fetch = true # Perform a `git fetch` before auditing (default: true)
+stale = false # Allow stale advisory DB (i.e. no commits for 90 days, default: false)
+
+# Output Configuration
+[output]
+deny = ["yanked", "unmaintained"]
```
legoktm, line 6: I think this line and the target arch+os stuff is the only thing that isn't default? we can get rid of the rest.

Oh and maybe keep the ignore option at the top as reference for people to use?

legoktm (changes requested):
Otherwise LGTM!

legoktm:
One other thing I thought about, do we want to leave one thing not updated and merge with failing CI so we can test the nightly notification? And then merge the update the next day?

zenmonkeykstop:
> One other thing I thought about, do we want to leave one thing not updated and merge with failing CI so we can test the nightly notification? And then merge the update the next day?

TBH I'm cool with merging it and waiting for a dependency to need updating. The job does fail if something is b0rked (see earlier commits).

legoktm (approved):
LGTM!

Issue #6917, opened 2023-08-14
streamline functions currently provided by `i18n_tool.py`


cfm:
- [x] Adapt `securedrop-client`'s `make extract-strings` target
    - Can this be generalized in `securedrop-tooling`?
    - Does this let us combine the `securedrop/{securedrop,desktop}` Weblate components?  (This might violate the "No side effects" constraint.)
- [x] Enforce in CI (`make check-strings`)
    - Can this be generalized in `securedrop-tooling`?
- Per <https://github.com/freedomofpress/securedrop-engineering/pull/30#issuecomment-1633171470>:
    - [x] Reimplement "Compile `.po` catalogs → `.mo` machine objects" as `make compile-translations`
- ~`make update-docs` → https://github.com/freedomofpress/securedrop-docs/blob/main/docs/includes/l10n.txt~
    - Counterpoint: https://github.com/freedomofpress/securedrop/issues/6917#issuecomment-1689091116 
    - freedomofpress/securedrop-docs#503
- ~~freedomofpress/securedrop-client#1280~~
    - Counterpoint: https://github.com/freedomofpress/securedrop/issues/6917#issuecomment-1690762840

cfm:
The `{check,extract}-strings` targets are available in `6917-c10s-l10n` as of 63be596914b09a75acf3d58955c9b44bd8dc7c4b, which is failing CI in `lint` as intended.  Remaining `i18n_tool.py` invocations to replace next week:

```sh-session
(.venv) user@sd-dev:~/securedrop$ grep -R "i18n_tool.py" * | grep -Ev "^changelog.md:"
Makefile:	@echo "SUPPORTED_LOCALES = $$(if test -f /opt/venvs/securedrop-app-code/bin/python3; then ./securedrop/i18n_tool.py list-locales --python; else DOCKER_BUILD_VERBOSE=false $(DEVSHELL) ./i18n_tool.py list-locales --python; fi)" | sed 's/\r//' >> securedrop/config.py
Makefile:	@$(DEVSHELL) $(SDROOT)/securedrop/i18n_tool.py list-translators
Makefile:	@$(DEVSHELL) $(SDROOT)/securedrop/i18n_tool.py list-translators --all
securedrop/debian/translations.sh:/tmp/securedrop-app-code-i18n-ve/bin/python3 ./i18n_tool.py --verbose translate-messages --compile
securedrop/bin/translation-test:./i18n_tool.py translate-messages --compile
[...]

PR #6953, opened 2023-09-27
ci(translation-tests): verify reproducibility of gettext machine objects

cfm:
## Status

Ready for review

## Description of Changes

* From freedomofpress/securedrop-client#1666
    * af7024750f49658e39c8ccd02047bf07ee7c49de placates ruff.
    * 6d3f4278e9604091cc0a96756e0836fff872ad3b placates typing < Python 3.9.
    * These are cross-filed in `securedrop-client` to keep these scripts in sync: freedomofpress/securedrop-client#1668
* Towards #6917

## Testing

Just like freedomofpress/securedro-client#1666:

- [ ] CI passes, especially `lint`.
- [ ] `make verify-mo` succeeds locally.
- [ ] If you change a random `install_files/ansible-base/roles/tails-config/templates/*.po` file, `make verify-mo` fails and highlights the change (because the "original" `.mo` file for that language now differs from the recompiled one).
- [ ] If you `pip3 uninstall diffoscope`, `make verify-mo` fails with a `subprocess.CalledProcessError` for return code `127`.

## Deployment

CI-only; no deployment considerations.


nathandyer (approved):
This all looks great, @cfm! No requested changes from me on the code, and the test plan checks out. I think this is all set.

- [x] CI passes, especially `lint`.
- [x] `make verify-mo` succeeds locally.
- [x] If you change a random `install_files/ansible-base/roles/tails-config/templates/*.po` file, `make verify-mo` fails and highlights the change (because the "original" `.mo` file for that language now differs from the recompiled one).
- [x] If you `pip3 uninstall diffoscope`, `make verify-mo` fails with a `subprocess.CalledProcessError` for return code `127`.

cfm:
Thanks, @nathandyer!

.circleci/config.yml
```diff
@@ -94,6 +94,13 @@ jobs:
             DOCKER_BUILD_ARGUMENTS="--cache-from securedrop-test-focal-py3:${fromtag:-latest}" securedrop/bin/dev-shell \
                   bash -c "/opt/venvs/securedrop-app-code/bin/pip3 install --require-hashes -r requirements/python3/develop-requirements.txt && make -C .. lint"
 
+      - run:
+          name: Check that translation machine objects are reproducible
```
legoktm, line 98: Why does this need to be its own step? Can it not be part of the `make lint` target?

cfm: I'd originally assumed it would be a step in `translation-tests`.  But it doesn't benefit from `translation-tests`'s parallelism, since we might as well verify all languages at once, not just supported languages one at a time.

The `lint` job seemed like the next best logical place.  But I take `make lint` itself to be analogous to spellcheck for human-modified files, not machine-generated ones.  And I don't see any benefit to running `make verify-mo` locally at all.  I resisted the temptation to move it into its own job altogether.  ;-)

Would you prefer to see this run somewhere else?  See also some extra CI bits in #6954.

legoktm: It's fast enough that the overhead of re-installing the pip dependencies takes more time than actually running the verification script. I think merging into lint is probably the right option, I'll roll that into my upcoming PR.

Philosophically, I think it doesn't matter what is being linted/tested, rather it's that "lint" is all the fast tools that just analyze code while "test" is slower stuff that does execution, etc.

cfm: Works for me!

Issue #6956, opened 2023-09-28
Test with Firefox 115 ESR / Tor Browser 13


legoktm:
Firefox 102 ESR is now EOL, so Tor Browser should be switching from 102 to 115 soonish, it's currently in an alpha state: https://blog.torproject.org/new-alpha-release-tor-browser-130a5/

And now that I check https://github.com/freedomofpress/securedrop/blob/2b25283d6de96d3e9717b9196f62dc18923c0199/securedrop/dockerfiles/focal/python3/Dockerfile#L18 we're actually testing with FF 91 ESR :/

But anyways, we should do some quick/basic human testing with TB 13.0 and then update our tests to use 115 ESR once it's released.

legoktm:
13.0a6 is out, and the [announcement](https://blog.torproject.org/new-alpha-release-tor-browser-130a6/) says:
> It is also our first release-candidate build for 13.0 stable.

So now is probably the time to test it.

legoktm:
Tor Browser 13.0 [hit stable today](https://blog.torproject.org/new-release-tor-browser-130/), so it broke CI since the download paths changed - fixing that now.

PR #6957, opened 2023-09-29
Test against Firefox 102 ESR and geckodriver to 0.33

legoktm:
## Status

Ready for review

## Description of Changes

Tor Browser 12 (currently latest) is roughly based on Firefox 102 ESR, so we should be using that to test against. Note that 102 is already EOL, but Tor Browser hasn't made the switch yet (#6956).

And then upgrade to the latest version of geckodriver while we're at it.

## Testing

* [x] CI passes

## Deployment

## Checklist

### DIFF
--- a/securedrop/dockerfiles/focal/python3/Dockerfile
+++ b/securedrop/dockerfiles/focal/python3/Dockerfile
@@ -15,8 +15,8 @@ RUN apt-get update  && apt-get install -y \
 # Current versions of the test browser software. Tor Browser is based
 # on a specific version of Firefox, noted in Help > About Tor Browser.
 # Ideally we'll keep those in sync.
-ENV FF_VERSION 91.12.0esr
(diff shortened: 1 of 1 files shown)


zenmonkeykstop:
:grimacing: Just noticed this - put in #6965, might make sense just to jump straight to FF115esr?

legoktm:
I think we mostly don't know when Tor Browser is going to switch over...not clear to me whether we're looking at days or a week or whether it's going to be weeks/months.

zenmonkeykstop:
> I think we mostly don't know when Tor Browser is going to switch over...not clear to me whether we're looking at days or a week or whether it's going to be weeks/months.

Fair enough. Let's update to the TBB 12.* version for now.

zenmonkeykstop (approved):
LGTM.

PR #6954, opened 2023-09-28
build: eliminate `i18n_tool.py`

cfm:
## Status

Ready for review

## Description of Changes

Closes #6917 by removing `i18n_tool.py` in favor of:
- `make extract-strings`, which wraps `pybabel extract`;
- `make check-{desktop-files,strings}`, which enforces `make extract-strings` in CI;
- `pybabel compile`;
- `make supported-locales`, which consumes `i18n.json` and can be further processed via `jq`;
- #6953;
- freedomofpress/securedrop-docs#503; and
- freedomofpress/securedrop-dev-docs#86.

## Testing

Because this pull request offloads much of our existing localization workflow onto Weblate, very little self-contained, local testing is possible or worthwhile.  Namely:

- Inspect packages for differences between `develop` and here, with a fresh build image each time:
    1. `(docker|podman) image prune -a --force && make build-debs` on `develop` and save them (e.g., `cp -Rp build build@develop`).
    2. `(docker|podman) image prune -a --force && make build-debs` on this branch.
    3. [x] `diffoscope build@develop/focal/securedrop-app-code*.deb build/focal/securedrop-app-code*.deb` shows only:
        - [x] diffs from this branch
        - [x] metadata changes in `.mo` files

Instead...

### Perform a mock release and deployment via `securedrop-i18n-sandbox`

A mock Localization Manager (@cfm) and Release Manager (@nathandyer) should pair their way through (approximately) the following:

1. [x] Coordinate with Infrastructure so that: freedomofpress/infrastructure#4393
    - [x] `weblate-sandbox.securedrop.org` uses...
    - [x] GitHub user @weblate-sandbox-fpf to open pull requests in...
    - [x] `freedomofpress/securedrop-i18n-sandbox`.
2. [x] **At `weblate-sandbox.securedrop.org`,** configure `securedrop/securedrop` and `securedrop-desktop` (from `freedomofpress/securedrop-i18n-sandbox`) like `securedrop/securedrop-client` (from `freedomofpress/securedrop-client-i18n-sandbox`).  **Add-ons:**
    - [x] "Update PO files to match POT"
         - [x] Keep previous msgids of translated strings
         - [x] Remove locations of translated strings
    - [x] "Customize gettext output"
        - [x] Long lines wrapping: No line wrapping
    - [x] "Squash Git commits"
        - [x] Commit squashing: Per language
        - [x] Append trailers to squash commit
    - [x] `desktop` only: Generate MO files
[...]

PR #6961, opened 2023-09-29
Replace bandit with ruff

legoktm:
## Status

Ready for review, ~~depends on #6954 (I did not fix the bandit issues in i18n_tool.py)~~

## Description of Changes

ruff has reimplemented the bandit rules[1], so we can use that as a better-integrated tool with the rest of our stack. We enable all the bandit rules and selectively disable some across the codebase and some in just tests where they don't make sense (e.g. flagging use of `assert` or using insecure crypto).

For bonus points we can get rid of the GitPython dependency, which has a history of (non-exploitable in our context) security issues.

[1] Per https://github.com/astral-sh/ruff/issues/1646 they've implemented nearly all of them, and the remaining ones aren't that important IMO.

## Testing

* [ ] CI passes
* [ ] `make check-ruff` passes locally

## Deployment

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -308,13 +308,6 @@ jobs:
                   bash -c "pip3 install -U -q --upgrade safety && make -C .. safety"
 
       - run:
-          name: Run static security testing on source code with bandit
-          command: |
-            fromtag=$(docker images |grep securedrop-test-focal-py3 |head -n1 |awk '{print $2}')
-            DOCKER_BUILD_ARGUMENTS="--cache-from securedrop-test-focal-py3:${fromtag:-latest}" securedrop/bin/dev-shell \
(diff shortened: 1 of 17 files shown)


cfm:
I'm very much in favor of (a) this specific change and (b) in general standardizing these tools, to avoid surprises like freedomofpress/securedrop-client#1668.  But you knew that already.  :-)

rocodes (dismissed):
LGTM pending [the il8n stuff](https://github.com/freedomofpress/securedrop/pull/6954) mentioned in the PR - yay ruff, thanks @legoktm :)

legoktm:
This should be ready to go now!

zenmonkeykstop (approved):
LGTM! (plays taps for bandit)
- CI checks out
- make check-ruff passes locally
- with some `#noqa`s dropped, the check alerts as expected.

PR #6969, opened 2023-10-03
Run `make lint` and other simple jobs in GitHub Actions

legoktm:
## Status

Ready for review

### lint
This is a ground-up rewrite of the "lint" job instead of porting it from
CircleCI. It will be much faster since it doesn't try to spin up a
Docker container and work through our various unnecessary caching
layers. It does literally the bare minimum to run lints.

The CircleCI lint job needs to remain for now because it's super
intertwined with everything else.

### part 2
Migrate rust, rust-security and updater-gui-tests since they're pretty simple.

## Testing

* [ ] Visual review & CI passes


legoktm:
I split out the first mypy speedup commit to a separate PR: #6971.

leee (approved):

Issue #6831, opened 2023-05-31
Why has CI slowed down?


cfm:
I thought I was imagining it, but CircleCI's "insights" confirm over the past 30 days:

| Workflow | Branch | Duration (P95) | Median duration (P50) |
| --- | --- | --- | --- |
| `securedrop_ci` | all | 52m 29s, +137% | 19m 56s, +36% |
| `securedrop_ci` | `develop` | 23m 57s, +51% | 21m 16s, +41% |

I was prepared to accept that the all-branch metric is up because we've been pushing more `stg-` branches for `staging-test-with-rebase`.  But that wouldn't account for the slowdown on `develop`.

legoktm:
deb-tests-1 is now running unconditionally and is much slower because it does 2 builds and diffoscopes them. But when I made that change, it was roughly at the same speed as the existing app-tests job, so in theory it shouldn't have caused a PR to take longer overall.

My inclination right now is that it's not worth trying to find specific causes of slow downs but just address the bigger issues of lint blocking other jobs from starting, reducing our parallelization, that we're using VMs instead of containers, etc.

PR #6971, opened 2023-10-04
Point mypy at redwood type stub directly

legoktm:
## Status

Ready for review 

## Description of Changes

Instead of the possibly slow process of compiling redwood and installing it into the virtualenv, we can just point mypy at the stub directly. (For some reason mypy wanted us to use an explicit "tuple" type.)

Notably this means you can run the lint process without needing a Rust toolchain installed.

Refs #6831.

## Testing

* [x] CI passes
* [x] clear Rust build cache with `rm -rf target` and then run `make typelint`, it should not compile Rust code.

## Deployment

## Checklist


zenmonkeykstop (approved):
LGTM!

PR #6981, opened 2023-10-10
Add initial `cargo vet` configuration

### LINKED ISSUE #6500: Determine plan for auditing Rust dependencies

legoktm:
## Status

Ready for review

## Description of Changes

cargo vet is a tool designed by Mozilla to record audits of Rust
dependencies, and it matches nicely with the philosophy of our diff
review system for Python. It can automatically present diffs and
verify everything has been checked.

### Trust

To reduce the number of audits we have to do, we trust the following
organizations:

* Bytecode Alliance (WASM/WASI)
* Google
* Internet Security Research Group (Let's Encrypt, etc.)
* Mozilla
* Zcash

We also trust a number of individual developers, because they are
members of the Rust Project and also trusted by one of the above
organizations we trust. We also trust the two Sequoia-OpenPGP team
members who release things.

All of the individual trust markers have an expiry date to remind us to
re-evalutate trustworthiness every so often.

### Exemptions

There are a number of dependencies that appear in our tree but are not
used on Linux x86_64 so we can ignore them entirely. These are marked in
config.toml with a policy stanza that has an empty criteria block. These
crates have been identified manually, in the future cargo-vet will
hopefully let us specify specific targets we care about and take care of
it automatically.

The remaining exemptions in config.toml have not been reviewed yet; we
can incrementally chip away at them.

### Signing

Unlike diff reviews, there is no PGP signing of this file. Because these
are committed into the Git repository directly, we can rely on that as a
measure of trust (unlike random wiki pages).

### CI

CI verifies that all dependencies have either been reviewed or exempted,
so there's no need for manual tracking in PR descriptions. Upstream
provides a GitHub Actions template that we use most of.

Fixes #6500.

## Testing

* [x] CI passes
* [x] follow https://mozilla.github.io/cargo-vet/install.html, then running `cargo vet` locally succeeds. `cargo vet suggest` lists all the crates that are currently exempt and not yet reviewed.
* [x] We're all on board with the list of trusted organizations + individuals

## Deployment

## Checklist

- [x] I will open a securedrop-dev-docs PR shortly

### DIFF
--- a/.github/workflows/cargo-vet.yml
+++ b/.github/workflows/cargo-vet.yml
@@ -1,0 +1,26 @@
+# Based off of https://mozilla.github.io/cargo-vet/configuring-ci.html
+
+name: cargo vet
+
+on: [push, pull_request]
+
+jobs:
+  cargo-vet:
+    name: Vet Dependencies
+    runs-on: ubuntu-latest
+    # Keep version in sync with rust-toolchain.toml
+    container: rust:1.71.1
+    env:
+      CARGO_VET_VERSION: 0.8.0
+    steps:
+    - uses: actions/checkout@v4
+    - uses: actions/cache@v2
+      with:
+        path: ${{ runner.tool_cache }}/cargo-vet
+        key: cargo-vet-bin-${{ env.CARGO_VET_VERSION }}
+    - name: Add the tool cache directory to the search path
+      run: echo "${{ runner.tool_cache }}/cargo-vet/bin" >> $GITHUB_PATH
+    - name: Ensure that the tool cache is populated with the cargo-vet binary
+      run: cargo install --root ${{ runner.tool_cache }}/cargo-vet --version ${{ env.CARGO_VET_VERSION }} cargo-vet
+    - name: Invoke cargo-vet
+      run: cargo vet --locked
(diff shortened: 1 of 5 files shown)


legoktm:
Summary from today's team meeting:
* cargo vet is good!
* The initial list of trusted organizations is fine, though it was noted that Zcash is the only one we don't already have some "dependency" on (e.g. we use Firefox/TBB, Google stuff, etc.).
* We want to lower the expiry on individuals to 6 months instead of 1 year for now as we get more familiar with Rust stuff.
* I need to write documentation on our policy and practices for this. 
* We should be able to audit most of the dependencies prior to the 2.7.0 release.

zenmonkeykstop (approved):
LGTM with expiry revisions!

PR #6983, opened 2023-10-10
Lazily intialize GPG code in EncryptionManager

legoktm:
## Status

Ready for review 

## Description of Changes

In most cases we should no longer need GPG, so lazily initialize it only when migration code or fallbacks are invoked.

Also take the opportunity to replace a TODO that was implemented.

## Testing

* [ ] CI passes
* [ ] Visual review

## Deployment

## Checklist

### DIFF
--- a/securedrop/encryption.py
+++ b/securedrop/encryption.py
@@ -52,18 +52,33 @@ class EncryptionManager:
         self._redis = Redis(decode_responses=True)
 
         # Instantiate the "main" GPG binary
-        self._gpg = gnupg.GPG(
-            binary="gpg2",
-            homedir=str(gpg_key_dir),
-            options=["--pinentry-mode loopback", "--trust-model direct"],
-        )
+        self._gpg = None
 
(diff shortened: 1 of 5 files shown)


securedrop/encryption.py
```diff
@@ -155,16 +170,16 @@ def decrypt_journalist_reply(self, for_source_user: "SourceUser", ciphertext_in:
                 secret_key=for_source.pgp_secret_key,
                 passphrase=for_source_user.gpg_secret,
             ).decode()
-        # TODO: Migrate gpg-stored secret keys to database storage for Sequoia
+        # In practice this should be uncreachable unless the Sequoia secret key migration failed
```
zenmonkeykstop, line 173: Error logging is usually disabled for the SI, but in cases where it's not (testing, troubleshooting) should we emit an error here to alert admins that there was a failed key migration earlier?

zenmonkeykstop (approved):
LGTM, tests passing locally and visual review is good.

PR #7000, opened 2023-10-12
redwood: Relax requirements for decrypting messages

### LINKED ISSUE #6991: It is (probably) okay to decrypt a message with a revoked, expired or weak key
`secret_key_from_cert` is used by `decrypt` to decrypt a message:

https://github.com/freedomofpress/securedrop/blob/8d00ba5d9d3835a22a09bcd3a338b65aee697e7d/redwood/src/keys.rs#L40-L48

But, there is no reason to not decrypt a message if it key is expired, revoked, the policy says the algorithms are no longer strong enough.  Is there?

legoktm:
## Status

Ready for review

## Description of Changes

Even if a source key is no longer valid per policy, we still want them to be able to decrypt a previously valid message for them. We can also drop the revocation/expiry filters, which were mostly theoretical in the SecureDrop context anyways.

Fixes #6991.

## Testing

## Deployment

## Checklist

### DIFF
--- a/redwood/src/keys.rs
+++ b/redwood/src/keys.rs
@@ -2,22 +2,8 @@ use crate::{Error, Result};
 use sequoia_openpgp::cert::prelude::ValidErasedKeyAmalgamation;
 use sequoia_openpgp::packet::key::{PublicParts, SecretParts, UnspecifiedRole};
 use sequoia_openpgp::packet::Key;
-use sequoia_openpgp::policy::Policy;
+use sequoia_openpgp::policy::{NullPolicy, Policy};
 use sequoia_openpgp::Cert;
-
-/// We want to use the same iterators on public and secret keys but it's not
-/// really possible to do it in a function because of type differences so we
-/// use a macro instead.
-macro_rules! filter_keys {
-    ( $keys:expr, $policy: ident ) => {
-        $keys
-            .with_policy($policy, None)
-            .supported()
-            .alive()
-            .revoked(false)
-            .for_storage_encryption()
-    };
-}
 
 /// Get public encryption keys from the specified cert, returning an error if
 /// no valid keys are found.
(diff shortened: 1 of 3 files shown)


redwood/src/keys.rs
```diff
@@ -45,7 +38,12 @@ pub(crate) fn secret_key_from_cert<'a>(
     // Pull the encryption keys that are compatible with by the standard policy
     // (e.g. not SHA-1) supported by Sequoia, and not revoked.
     // These filter options should be kept in sync with `Helper::decrypt()`.
-    let keys: Vec<_> = filter_keys!(cert.keys().secret(), policy).collect();
+    let keys: Vec<_> = cert
+        .keys()
+        .secret()
+        .with_policy(policy, None)
+        .for_storage_encryption()
```
legoktm, line 46: @nwalfield, Is there a way to use the `for_storage_encryption()` filter without also using `with_policy()`? I couldn't find one, and dropping the storage encryption part means the wrong secret key gets picked for decryption...

nwalfield: Good point.  @teythoon just raised the same point [over here](https://gitlab.com/sequoia-pgp/sequoia-octopus-librnp/-/issues/94#note_1598255766).

One approach would be to first try the standard policy but fall back to the weaker [null policy](https://docs.sequoia-pgp.org/sequoia_openpgp/policy/struct.NullPolicy.html) if the standard policy doesn't find a valid self signature.  This is important, because a lot of keys only have self signatures that rely on SHA-1, but it should still be possible to use them to decrypt messages.

legoktm: One clarification via IRC, the use of the standard policy first before falling back to the null policy is to avoid downgrade attacks.

legoktm:
(Rebased)

zenmonkeykstop (approved):
LGTM based on visual review and CI passing. Given that the policy remains in place for encryption keys this is consistent with our overall approach to the SHA-1 problem (allow decryption of old files, prevent encryption with weak keys going forward.).

PR #7014, opened 2023-10-18
Do not include config.py in built packages

### LINKED ISSUE #6973: `make build-debs` doesn't explicitly ignore config.py
If you run `make securedrop/config.py && make build-debs`, it'll fail with:

```
FAILED builder/tests/test_securedrop_deb_package.py::test_app_code_paths_missing[/var/www/securedrop/config.py] - Failed: found -rw-r--r-- root/root      3724 2023-08-18 11:56 ./var/www/securedrop/config.py
```

This just needs a fix like https://github.com/freedomofpress/securedrop/commit/62b195f71cd93f951b85d855a90b3dc70007c0ce. It's not urgent because the workaround is trivial, just delete your local config.py before packaging.

legoktm:
## Status

Ready for review

## Description of Changes

Just like the custom_logo.png, we don't want to package config.py if it exists.

The absence of config.py is already verified by
test_securedrop_deb_package.py, but now you won't have to rebuild your packages since it won't be included at all.

Fixes #6973.

## Testing

* [ ] Run `make securedrop/config.py`
* [ ] Run `make build-debs`
* [ ] Post-package build tests pass, i.e. no config.py included in the package

## Deployment

## Checklist

### DIFF
--- a/securedrop/debian/rules
+++ b/securedrop/debian/rules
@@ -49,6 +49,8 @@ override_dh_install:
 	rm upload-screenshots.py
 	# If a custom logo exists in the tree, delete it
 	rm -f static/i/custom_logo.png
+	# Same with any config.py
+	rm -f config.py
 	dh_install
 
 override_dh_strip_nondeterminism:


zenmonkeykstop (approved):
LGTM based on visual inspection, there's no situation in which we'd want to include a common config.py in a package.

zenmonkeykstop:
(IMO we can get away without backporting this into the in-flight release as it will get caught by a test anyway...)

legoktm:
Agreed, and real builds should be done from a fresh/clean git checkout so no config.py should exist.

PR #7016, opened 2023-10-18
loaddata: Allow increasing the size of submitted files

legoktm:
## Status

Ready for review

## Description of Changes

This would make it easier to get rough benchmarks of encryption steps.

This is pretty quick/rough, not sure if there's a better way we want to go about this. Note that this would need to be manually applied to a 2.6.0 install to get numbers for gpg.

## Testing

* [ ] Try `time ./loaddata.py --file-size 500000000`, etc.

## Deployment

## Checklist

### DIFF
--- a/securedrop/loaddata.py
+++ b/securedrop/loaddata.py
@@ -192,7 +192,7 @@ def submit_message(source: Source, journalist_who_saw: Optional[Journalist]) -> 
         db.session.add(seen_message)
 
 
-def submit_file(source: Source, journalist_who_saw: Optional[Journalist]) -> None:
+def submit_file(source: Source, size: int, journalist_who_saw: Optional[Journalist]) -> None:
     """
     Adds a single file submitted by a source.
     """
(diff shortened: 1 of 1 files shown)


securedrop/loaddata.py
```diff
@@ -202,7 +202,7 @@ def submit_file(source: Source, journalist_who_saw: Optional[Journalist]) -> Non
         source.interaction_count,
         source.journalist_filename,
         "memo.txt",
-        io.BytesIO(b"This is an example of a plain text file upload."),
+        io.BytesIO(b"A" * size),
```
zenmonkeykstop, line 205: We compress then encrypt, so this will compress down to pretty much nothing before further processing, making it less useful for timing encryption or file transfers. Consider adding some randomness?

zenmonkeykstop (changes requested):

securedrop/loaddata.py
legoktm: Hmm, good point. Also there's a lot of overhead from SecureTemporaryFile. Maybe just a simple benchmarking script that just benchmarks exactly what we want (GPG vs redwood) would be better. Let me try to come up with something...

legoktm:
I don't want to merge this anymore. I created a separate benchmark script, which we could or could not merge, but IMO probably not.

Issue #7022, opened 2023-10-20
Benchmark redwood vs GPG encryption and decryption performance


legoktm:
I pushed a very rudimentary benchmarking script to the `benchmark-pgp` branch: https://github.com/freedomofpress/securedrop/blob/benchmark-pgp/securedrop/benchmark.py

~~The initial results of encryption performance (on my Qubes laptop) don't look good for larger file sizes:~~ Edit: Ignore these, the script was broken for GPG. See below
```
root@35431cb69004:/home/user/github/freedomofpress/securedrop/securedrop# sudo -u www-data ./benchmark.py 
== 5KB (100 loops) ==
redwood: ['94.6 msec', '86.5 msec', '78.5 msec', '82.2 msec', '90.3 msec']
gpg:     ['621 msec', '636 msec', '678 msec', '622 msec', '641 msec']
== 50KB (100 loops) ==
redwood: ['129 msec', '121 msec', '120 msec', '120 msec', '119 msec']
gpg:     ['627 msec', '640 msec', '643 msec', '653 msec', '646 msec']
== 500KB (100 loops) ==
redwood: ['716 msec', '679 msec', '678 msec', '710 msec', '680 msec']
gpg:     ['725 msec', '686 msec', '655 msec', '642 msec', '669 msec']
== 5MB (100 loops) ==
redwood: ['6.73 sec', '7.31 sec', '7.29 sec', '7.26 sec', '7.42 sec']
gpg:     ['665 msec', '693 msec', '676 msec', '651 msec', '658 msec']
== 50MB (10 loops) ==
redwood: ['6.08 sec', '6.05 sec', '6.12 sec', '6.86 sec', '6.14 sec']
gpg:     ['79.1 msec', '74.2 msec', '83 msec', '75.2 msec', '74.9 msec']
== 500MB (10 loops) ==
redwood: ['59.4 sec', '61.2 sec', '61.1 sec', '62.9 sec', '64.9 sec']
gpg:     ['98.5 msec', '88.7 msec', '88 msec', '83.2 msec', '84 msec']
```

legoktm:
One thing I realized was that previously we were not armoring the encrypted ciphertext: 

https://github.com/freedomofpress/securedrop/blob/a3b80ff96836c04331a714f23eb46462c61f63be/securedrop/encryption.py#L225-L231

Removing the armorer pass from redwood seems to give us a 2x speedup at the 500KB+ level:
```
== 5KB (100 loops) ==
[...]

PR #7023, opened 2023-10-20
redwood: Don't armor encrypted files/messages

legoktm:
## Status

Ready for review

## Description of Changes

We didn't have GPG armor ciphertext, so we don't need redwood to either. This will result in a noticable speedup during encryption (c.f. #7022).

We do need to adjust tests now that we can no longer easily look for the "-----BEGIN PGP MESSAGE-----" string.

## Testing

* [x] CI passes
* [x] Visual review

## Deployment

Any special considerations for deployment? No, this restores the status quo of non-armored messages.

## Checklist


legoktm:
Hmm, seems like there are tests that need an armored version to POST. 🤔

securedrop/tests/test_encryption.py
```diff
@@ -138,9 +138,9 @@ def test_encrypt_source_file(self, config, tmp_path):
                 encrypted_file_path_out=encrypted_file_path,
             )
 
-        # And the output file contains the encrypted data
+        # And the output file doesn't contain the file plaintext
         encrypted_file = encrypted_file_path.read_bytes()
-        assert encrypted_file.startswith(b"-----BEGIN PGP MESSAGE-----")
+        assert file_to_encrypt_path.read_bytes() not in encrypted_file
```
cfm, line 143: I raised an eyebrow at this, since `file_to_encrypt_path.read_bytes()` could be up to 500 MB: couldn't we just check the first chunk, of whatever size?  But of course this is just a test, so we control the plaintext input as well as the ciphertext output.

cfm (approved):
Tidy change, well documented.

Issue #7025, opened 2023-10-25
When a source's reply secret key is unavailable, but there are replies encrypted with it, the source can't log in


zenmonkeykstop:
## Steps to Reproduce

- create a source in the SI and submit a file, then log out
- reply to the source from the JI
- delete the source's reply keypair
- attempt to log in as the source

## Expected Behavior
- source logs in but encrypted replies are not available

## Actual Behavior
- the source gets an error page and the following errors are thrown in source logs:
```
[Wed Oct 25 02:27:52.095767 2023] [wsgi:error] [pid 134751:tid 132485121509120] [remote 127.0.0.1:52012]   File "/var/www/securedrop/encryption.py", line 210, in decrypt_journalist_reply
[Wed Oct 25 02:27:52.095848 2023] [wsgi:error] [pid 134751:tid 132485121509120] [remote 127.0.0.1:52012]     raise GpgDecryptError(out.stderr)
[Wed Oct 25 02:27:52.095906 2023] [wsgi:error] [pid 134751:tid 132485121509120] [remote 127.0.0.1:52012] encryption.GpgDecryptError: [GNUPG:] ENC_TO 311FF89D40A5FAC9 1 0
[Wed Oct 25 02:27:52.095963 2023] [wsgi:error] [pid 134751:tid 132485121509120] [remote 127.0.0.1:52012] [GNUPG:] ENC_TO 2FB6D22BADC72E83 1 0
[Wed Oct 25 02:27:52.096057 2023] [wsgi:error] [pid 134751:tid 132485121509120] [remote 127.0.0.1:52012] [GNUPG:] KEY_CONSIDERED 7ECA271BA6894A7EBAE6414C4EE9B0B8EF362114 0
[Wed Oct 25 02:27:52.096117 2023] [wsgi:error] [pid 134751:tid 132485121509120] [remote 127.0.0.1:52012] gpg: encrypted with 4096-bit RSA key, ID 2FB6D22BADC72E83, created 2022-09-23
[Wed Oct 25 02:27:52.096183 2023] [wsgi:error] [pid 134751:tid 132485121509120] [remote 127.0.0.1:52012]       "SD test submission key"
[Wed Oct 25 02:27:52.096268 2023] [wsgi:error] [pid 134751:tid 132485121509120] [remote 127.0.0.1:52012] [GNUPG:] NO_SECKEY 2FB6D22BADC72E83
[...]

PR #7026, opened 2023-10-25
Validate an exported secret key is decryptable by Sequoia

legoktm:
## Status

Ready for review

## Description of Changes

When we export a secret key in EncryptionManager, validate the output by checking that it is decryptable by Sequoia using the given passphrase and has the expected fingerprint.

A new redwood function, is_valid_secret_key() is the sibling to the eixsting is_valid_public_key(), except that it also takes a passphrase and verifies the secret key can be unlocked using the passphrase.

If the key passes all the conditions, only then is it returned by EncryptionManager to be saved in the database, and deleted out of GPG. If, for whatever reason, GPG fails at exporting the key, or exports something Sequoia can't handle, the key will not be stored and it'll continue to fall back to using GPG for decryption.

While we're at it, the export function is now named `get_source_secret_key_from_gpg`, to highlight that this specifically just exports the key from GPG and won't work for Sequoia based sources.

Refs #7025.

## Testing

I don't have any manual steps because this is to guard against an edge case in which GPG export fails, so it's easiest to prove with mocks and test cases, but if people have suggestions, they could be added to the test plan!

## Deployment

## Checklist


zenmonkeykstop (dismissed):
LGTM based on visual review

zenmonkeykstop:
Tried using GH's conflict resolution tool, but now the rebase in CI is b0rked - that's on me. Will back that out, rebase locally, and force-push.

zenmonkeykstop (approved):
LGTM once more

PR #7035, opened 2023-10-26
Have JI validate journalist key is valid

### LINKED ISSUE #7030: determine post-upgrade failure-mode for a SHA-1-signed submission key
## Description

After #6948 (for #6399), redwood will refuse to encrypt to a submission key with a SHA-1 signature.

After #6928, `securedrop-admin sdconfig` will reject a submission key with a SHA-1 signature.  This check guarantees that new and reconfigured instances will comply with #6948.

What will happen to an instance with a SHA-1-signed signature after upgrading to v2.7.0?

## Possible approaches

| Option | Documentation changes | Code changes | Implication |
| --- | --- | --- | --- |
| Fail open, but log | optional | ✓ | Admin must monitor logs and/or OSSEC alerts. |
| Fail open, but document | ✓ | ✗ | Admin must monitor release notes or check documentation. |
| Fail closed | optional | ✓[1] | Admin can contact us for help. |

**Notes:**
1. @legoktm observes that, without a code change to handle this case, Apache will come back up after reboot even if the `postinst` script fails under `unattended-upgrades`.

legoktm:
## Status

Ready for review

## Description of Changes

During the Sequoia migration, we need to export the journalist public key from the GPG keyring into a file on disk. We also needed to validate the key was usable by Sequoia (e.g. no SHA-1 binding signatures).

Previously the plan was to validate it during the postinst and error out if it wasn't valid, but if validation fails for whatever reason, then we abort, which interrupts the postinst, so, e.g. database upgrades won't be applied. In retrospect having the validation fail at package install time is bad, because it requires even more manual work to get an instance in a working state since you need to manually apply the updates after doing a key rotation.

Now we validate the journalist key during startup of the Journalist Interface, printing and logging an error if it doesn't validate and then exiting. This should bring attention to journalists and therefore the admin that the instance needs manual attention. We will also include information about this change in the pre-release and release announcements.

Fixes #7030.

## Testing

* [x] In the dev environment:
   * [x] `git checkout 40a53c07bae7fee643bddf6db62118638ad0b421 securedrop/tests/files/test_journalist_key.pub` (get a SHA-1 signed version of the test key)
   * [x] Edit loaddata.py to make it a no-op (comment out the if name == main block)
   * [x] Run `make dev`, the JI should immediately exit with an error message about the key not being supported.
* [ ] In a staging environment: (might be OK to defer this to rc3 QA?)
   * [ ] Set up a 2.6.0 staging environment, which should use the weak test key. Download it from `/public-key` and run it through sq-keyring-linter to verify it is using a SHA-1 signature.
   * [ ] Upgrade to 2.7.0. The upgrade should succeed, but the JI should be down

## Deployment

Any special considerations for deployment? Yes, see description.

## Checklist


securedrop/journalist.py
```diff
@@ -21,10 +25,28 @@ def prime_keycache() -> None:
...
-prime_keycache()
+def validate_journalist_key() -> None:
+    """Verify the journalist PGP key is valid"""
+    encryption_mgr = EncryptionManager.get_default()
+    # First check that we can read it
+    try:
+        journalist_key = encryption_mgr.get_journalist_public_key()
+    except Exception as e:
+        print(f"ERROR: Unable to read journalist public key: {e}", file=sys.stderr)
+        app.logger.error(f"ERROR: Unable to read journalist public key: {e}")
```
cfm, line 36: Nit: Why both `print()` and `app.logger.error()` here?  Is that for `capsys`'s benefit?

cfm (approved):
The development-environment test plan looks good:

* [x] In the dev environment:
   * [x] `git checkout 40a53c07bae7fee643bddf6db62118638ad0b421 securedrop/tests/files/test_journalist_key.pub` (get a SHA-1 signed version of the test key)
   * [x] Edit loaddata.py to make it a no-op (comment out the if name == main block)
   * [x] Run `make dev`, the JI should immediately exit with an error message about the key not being supported.

```sh-session
user@sd-dev:~/securedrop$ git checkout 40a53c07bae7fee643bddf6db62118638ad0b421 securedrop/tests/files/test_journalist_key.pub
Updated 1 path from 7b9207952
user@sd-dev:~/securedrop$ make dev
[...]
 => Journalist Interface <= 
ERROR: Journalist public key is not valid: NoSupportedKeys("65A1B5FF195B56353CC63DFFCC40EF1228271441")

 => Source Interface <= 

 * Serving Flask app 'source_app' (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.

 => Journalist Interface <= 

[2023-10-27 00:21:33,390] ERROR in journalist: ERROR: Journalist public key is not valid: NoSupportedKeys("65A1B5FF195B56353CC63DFFCC40EF1228271441")

One of the development servers exited unexpectedly. See the traceback above for details.
Once you have resolved the issue, you can re-run './manage.py run' to continue developing.
```

Restoration unbreaks the Journalist Interface:


```sh-session
user@sd-dev:~/securedrop$ git reset --hard
HEAD is now at 7fd01d908 Have JI validate journalist key is valid
user@sd-dev:~/securedrop$ make dev
[...]
 => Source Interface <= 

 * Serving Flask app 'source_app' (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.

 => Journalist Interface <= 

 * Serving Flask app 'journalist_app' (lazy loading)
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.
 * Running on all addresses (0.0.0.0)
 * Running on http://127.0.0.1:8081
 * Running on http://172.17.0.2:8081
```

I agree that we can defer staging/production-environment testing to RC3 QA, but I'll leave this for @zenmonkeykstop to merge as Release Manager.

securedrop/journalist.py
legoktm: Just to scream as loudly as possible that there's an issue. I started with print because I thought it might be to early to use a logger, but no, the logger seemed fine, so then I added it in as an extra thing but figured leaving in the print could help and wouldn't be harmful.

zenmonkeykstop:
Agreed on testing, we can add the non-dev test to the RC3 test plan.

PR #7059, opened 2023-11-02
Fix Submission Key validation at startup, improve SI experience

### LINKED ISSUE #7058: [2.7.0] JI does not fail loudly when configured with a SHA-1 key
## Description

Weak GPG keys are no longer supported (https://sequoia-pgp.org/blog/2023/02/01/202302-happy-sha1-day/). The JI should fail to start when configured with one (via a potential upgrade from a SecureDrop version that did support them and that was configured with a weak key), emitting an error in its logs as it does so. Instead, it's starting successfully and then failing on all encryption operations.

## Steps to Reproduce

- install 2.6.0 with a weak (RSA 1024-bit) key
- upgrade to 2.7.0
- visit the JI and attempt to submit a reply

## Expected Behavior

- JI is unavailable, `journalist-error.log` records message about weak journalist key

## Actual Behavior

- JI is available
- replies fail with a 500 error
- SI is available (this is actually expected) and submissions fail with a 500 error (this is suboptimal)

## Comments

legoktm:
## Status

## Description of Changes

The previously implemented Submission Key validation during JI startup did not work in production because it was under the `if __name__ == "__main__":` block, which isn't executed by WSGI.

In the JI, we now run it upon import and exit if it's not valid. For the SI, we check before each request, and if it's invalid, render a new `offline.html` template as a HTTP 503.

Testinfra smoke tests cover behavior on prod instances, but since it's destructive, will only run in staging.

Fixes #7058.

## Testing

How should the reviewer test this PR? TK

## Deployment

## Checklist


securedrop/source_app/__init__.py
```diff
@@ -106,6 +107,15 @@ def check_tor2web() -> Optional[werkzeug.Response]:
             return redirect(url_for("info.tor2web_warning"))
         return None
 
+    @app.before_request
```
legoktm, line 113: I think this check is fast enough to run on every request, but maybe in the future we can plan to just move this to the index page? Or maybe just remove this and the JI check for 2.8.0?

zenmonkeykstop: I originally added this as a single check at startup and added a config value to refer to instead, just to avoid the overhead here. We could probably do it on the index page only (since you can't get far in the app without signing up or logging in from there) but per-request seems more bulletproof so long as it's simple. I wouldn't do it per-request if we're checking the key every time.

securedrop/source_templates/offline.html
```diff
@@ -0,0 +1,3 @@
+{% extends "base.html" %}
+
+{% block tab_title %}{{ gettext('Server Error') }}{% endblock %}
```
zenmonkeykstop: Apparently this should be "Server error" (note capitalization) to avoid awaking the translation gods.

securedrop/source_app/__init__.py
legoktm: Oops, I thought I was simplifying. OK, I've re-added that back in, but I'm going to try to wait for the staging test to complete before pushing again.

securedrop/source_templates/offline.html
legoktm: Done.

zenmonkeykstop:
Tested in Qubes staging via "make build-debs; make staging"

With the existing strong key:
- [x] SI comes up, can create a source and submit successfully
- [x] JI comes up.

Subbed in a weak key for the test key and observed the following:
- [x] SI comes up with Offline message
- [x] JI returns a 500 error and message re: key is logged

zenmonkeykstop (approved):
LGTM based on visual review and review of behaviour in staging env.

PR #7065, opened 2023-11-02
build: Fix cargo-vet binary caching

### LINKED ISSUE #7011: cargo-vet CI caching isn't working
The CI manifest for running cargo-vet is supposed to cache the built binary, but it doesn't, so it's rebuilding every time.

https://github.com/freedomofpress/securedrop/blob/develop/.github/workflows/cargo-vet.yml

legoktm:
## Status

Ready for review

## Description of Changes

The upstream cargo-vet CI manifest was a bit overcomplicated, trying to install to a different path and it just wasn't working.

Instead, install cargo-vet to the normal path, store it in the actions/cache, and skip installing on cache hits.

On a cache hit, it drops job runtime to less than a minute.

Fixes #7011.

## Testing

* [x] Visual review
* [x] CI passes
* [ ] Review https://github.com/freedomofpress/securedrop/actions/workflows/cargo-vet.yml and see that runtime has dropped dramatically. Look at those workflows to see that the install step is being skipped since it's fetched from cache
* [x] https://github.com/freedomofpress/securedrop/actions/caches shows a new `cargo-vet-0.8.0` cache with 6.5MB stored, while the old key `cargo-vet-bin-0.8.0` had nothing stored (45 bytes).

## Deployment

### DIFF
--- a/.github/workflows/cargo-vet.yml
+++ b/.github/workflows/cargo-vet.yml
@@ -1,4 +1,4 @@
-# Based off of https://mozilla.github.io/cargo-vet/configuring-ci.html
+# Roughly based off of https://mozilla.github.io/cargo-vet/configuring-ci.html
 
 name: cargo vet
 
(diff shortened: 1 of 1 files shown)


legoktm:
The `push` job was able to use the cache while the `pull_request` job was not - this is because they don't share a branch so they have a different cache space. https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key explains that it checks the local branch then the `develop` branch, so once the develop cache is populated properly, all the branches + PRs will be able to use it.

zenmonkeykstop (approved):
LGTM

PR #7070, opened 2023-11-06
redwood: Intern string and improve Stream test coverage

legoktm:
## Status

Ready for review

## Description of Changes

To prevent pyo3 from converting `"read"` into a PyString over and over, we can intern it once to avoid repeatedly creating new PyStrings (which I learned while auditing pyo3).

While we're at it, improve test coverage of the Stream::read_bytes() function with more integration tests that verify more error handling behavior: if read() raises an exception or returns the wrong type.

## Testing

* [x] CI passes

## Deployment

## Checklist


zenmonkeykstop (approved):
LGTM

PR #7088, opened 2023-12-07
Fix Broken Logic in bootstrap.py

### LINKED ISSUE #7084: Cannot create virtualenv with `./securedrop-admin setup` on fresh Tails 5.20 disks
## Description

When trying to step through a SecureDrop installation or provisioning a new Admin or Journliast Workstation drive, the process fails at the first step while running `./securedrop-admin setup`, with a message of "Unable to create virtualenv. Check network settings and try again."

The full output is:
```
./securedrop-admin setup -v
INFO: Checking apt dependencies are installed
INFO: Setting up virtualenv
DEBUG: b'ERROR: virtualenv cannot be found.\n'
ERROR: Unable to create virtualenv. Check network settings and try again.
DEBUG: Cleaning up virtualenv
```

## Steps to Reproduce

Important note: This appears to only be happening on USB drives that were flashed with a fresh copy of Tails 5.20. I **could not** reproduce this on my primary Admin Workstation drive, with an in-place upgrade to 5.20.

1. Download Tails 5.20 .img file
2. Verify the file, then flash it to a USB drive
3. Boot the USB and create a new Persistent Storage Volume
4. Connect to Tor and clone the SD repo
5. Run `./securedrop-admin setup` from the local checkout

## Expected Behavior

A virtual environment is created and depedencies are installed.

## Actual Behavior

An error appears alerting that the virtualenv could not be created.

nathandyer:
## Status

Ready for review

## Description of Changes

Fixes #7084 

There were a few areas of broken logic in the is_missing_dependency function, which would result in broken or unidentified behavior if the underlying apt-cache was stale or non-existent.

If, while doing a search of the apt-cache, it only located a subset of the package list, or if it were completely unable to find a package (such as in the event of a missing cache), it would still return False, because the grep operation didn't find the "Installed: (none)" string it was looking for. So in those situations, even though the dependencies were missing, it did not return True as it should have.

The way the dependencies were listed at the top, with spaces inserted instead of tabs, meant that when we split the string using a space character as the delimiter to form the command, there were dozens of empty strings, which apt-cache would attempt to search for, and would turn up the same error as a missing package, making it impossible to determine when packages were missing.

This change eliminates those extra spaces, simplifies the logic to prefer built-in string-checking in Python rather than shelling out to grep, and fixes the broken return logic so it identifies when dependencies are missing due to not being in the apt cache at all.

## Testing

**Note**: This test plan requires Tails 5.20

1. Create a clean Tails 5.20 disk, enable persistent storage, and set an administrator password
2. Clone the SecureDrop repo, staying on the develop branch for now, and run `./securedrop-admin setup`
3. Confirm that you receive a virtualenv error
4. Switch to this branch and repeat `./securedrop-admin setup`
5. Confirm that this time you are asked for an admin password, and that it is able to update the package cache and install the dependencies
6. Re-run `./securedrop-admin setup`, and confirm that you are no longer prompted for an admin password, and that it sees the dependencies as having already been installed

### DIFF
--- a/admin/bootstrap.py
+++ b/admin/bootstrap.py
@@ -30,14 +30,8 @@ DIR = os.path.dirname(os.path.realpath(__file__))
 VENV_DIR = os.path.join(DIR, ".venv3")
 
 # Space-separated list of apt dependencies
-APT_DEPENDENCIES_STR = "python3-virtualenv \
-                   python3-yaml \
-                   python3-pip \
-                   virtualenv \
-                   libffi-dev \
-                   libssl-dev \
-                   libpython3-dev \
-                   sq-keyring-linter"
+APT_DEPENDENCIES_STR = "python3-virtualenv python3-yaml python3-pip virtualenv \
+libffi-dev libssl-dev libpython3-dev sq-keyring-linter"
 
 
 def setup_logger(verbose: bool = False) -> None:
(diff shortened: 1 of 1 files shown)


legoktm (commented):
Nice, this looks much better and makes a lot of sense! I think you just need to fix some lint/formatting errors to make CI happy.

I am wondering if we're better off just doing `os.file.exists('/usr/bin/virtualenv')` or whatever instead of even bothering to look at the apt-cache...

nathandyer:
Thanks for reviewing @legoktm! I believe I have appeased the linter gods, so this is ready for another look at your convenience.

As for your note about possibbly just checking if the virtualenv exists, my gut instinct is that that may not be sufficient to ensure a consistent state, because I can imagine a situation where someone might have a virtual environment created, but have somehow uninstalled one of the dependencies along the way, which just checking for the virtual environment wouldn't catch.

admin/bootstrap.py
```diff
@@ -111,24 +105,22 @@ def is_missing_dependency() -> bool:
...
-        returncode = grep_process.returncode
+        # If any packages couldn't be found, it may point to an apt cache issue
+        if "Unable to locate package" in apt_query_result.stderr:
+            return True
 
-        # If the above command returns 0, then one or more packages are not installed.
-        return returncode == 0
+        return False
 
     except subprocess.CalledProcessError as e:
```
legoktm, line 125: This is no longer needed AFAICT, per https://docs.python.org/3.9/library/subprocess.html#subprocess.run it is only raised if check=True, which you're not passing (maybe you should be checking the return code? not sure).

legoktm (changes requested):
Sorry for the delay, overall looks good. One last comment about checking the return code of the apt-cache command

admin/bootstrap.py
nathandyer: Thanks @legoktm! Good catch! Rather than removing the exception handling, I think we _want_ to be alerted to an instance where running apt returns a non-zero exit code, so the exception is handled appropriately (as it has been in the past).

I've tacked on a commit where I pass check=True, which I think should be all that's necessary to process that return code.

If that looks good to you, we can squash the commit. And if not, let me know how you think we should proceed and I'll adjust accordingly.

zenmonkeykstop:
(rebased to pull in a dependency bump and clear static analysis CI job)

legoktm (approved):
Nice, LGTM!

PR #7110, opened 2024-01-30
fix `staging-test-with-rebase`, failing on stale Ubuntu mirrors

cfm:
## Status

Ready for review

## Description of Changes

`staging-test-with-rebase` has been failing since January 3, because an `apt-get install` had stale mirrors.  Let's `apt-get update` them.

## Testing

- [ ] CI passes, with `stg-` including `staging-test-with-rebase`.

## Deployment

The `prepare-servers` task is also run in production.  `apt-get update` at worst has no side effects and at best will also prevent this failure during a production installation.

### DIFF
--- a/install_files/ansible-base/roles/prepare-servers/tasks/main.yml
+++ b/install_files/ansible-base/roles/prepare-servers/tasks/main.yml
@@ -3,7 +3,7 @@
 # are present
 
 - name: Install python and packages required by installer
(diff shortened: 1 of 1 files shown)


zenmonkeykstop (approved):
CI is passing, lgtm - we're probably not seeing this in prod coz said update happens initially as part of the OS install.

PR #7124, opened 2024-02-29
Persist onion addresses with "make dev-tor"

### LINKED ISSUE #7115: make dev-tor: persist onion addresses
## Description

`make dev-tor` spins up the servers in containers with onion addresses. However, these are non-persistent, meaning that each time they are ran, a new address will be generated. For Workstation development, having stable onion addresses is key. Otherwise one would need to reprovision the workstation every time they change.

Ensuring that `make dev-tor` makes the secrets persistent, would help development.

## How will this impact [SecureDrop users](https://github.com/freedomofpress/securedrop-ux/wiki/Users)?
No impact. Dev quality of life impact only.

## How would this affect SecureDrop's [threat model](https://docs.securedrop.org/en/stable/threat_model/threat_model.html)?
Does not. Dev-facing.

## User Stories

As an SDW developer, I want to spin up SD servers with persistent onion addresses quickly without provisioning physical machines of VMs with the servers.

deeplow:
Makes the SecureDrop Workstation dev workflow easier by removing the need to setup standalone server VMs if one wants a persistent server address and authentication token.

The solution involves creating a docker volume. This will persist as long as the user does not manually remove it or moves to another system without porting the volume. Since this a dev environment, long term persistence should not be a concern, so this solution sufices.

Fixes #7115

## Status

Ready for review 

## Description of Changes

Fixes #7115.

## Testing
Run `make dev-tor` multiple times and ensure that the same journalist config is generated.

## Deployment

Choose one of the following:

- `admin/requirements.in`
  container)

(remove) this section.

Choose one of the following:

### DIFF
--- a/devops/clean
+++ b/devops/clean
@@ -36,6 +36,11 @@ function remove_unwanted_files() {
         install_files/ansible-base/tor_v3_keys.json \
         build/*.deb
 
+    # Remove any Onion URL from make dev-tor
+    if docker volume inspect sd-onion-services > /dev/null; then
+        docker volume remove sd-onion-services
+    fi
+
     # Remove extraneous copies of the git repos, pulled in
     # via the Molecule upgrade testing scenario.
     rm -rf molecule/upgrade/.molecule/sd-orig \
--- a/securedrop/bin/dev-deps
+++ b/securedrop/bin/dev-deps
@@ -54,16 +54,28 @@ function maybe_create_config_py() {
 function maybe_use_tor() {
     if [[ -n "${USE_TOR:-}" ]]; then
         echo "Setting up Tor..."
+        if [ ! -d "/var/lib/tor/services" ]; then
+            sudo chown -R debian-tor:debian-tor /var/lib/tor/services
+        else
+            sudo -u debian-tor mkdir -p /var/lib/tor/services
+        fi
         # append torrc lines for SI and JI
-        sudo -u debian-tor mkdir -p /var/lib/tor/services
         echo "HiddenServiceDir /var/lib/tor/services/source/" | sudo tee -a /etc/tor/torrc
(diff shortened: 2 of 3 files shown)


zenmonkeykstop:
This looks like a great addition @deeplow, with only a couple of things to address:
- lint shellcheck fails need correcting
- the `make clean` target should be updated to remove the container volume with `/var/lib/tor` in it
- Ctrl-C no longer seems to work to kill the services? unclear if that's as a result of this PR tho.

deeplow:
>    lint shellcheck fails need correcting
Done in 6efc371

>    the make clean target should be updated to remove the container volume with /var/lib/tor in it
Addressed in c1b8c31

>    Ctrl-C no longer seems to work to kill the services? unclear if that's as a result of this PR tho.

Addressed in 2e1f3e11. I wasn't keeping the `DOCKER_RUN_ARGUMENTS` and thus loosing the `-it` params.

devops/clean
```diff
@@ -36,6 +36,11 @@ function remove_unwanted_files() {
         install_files/ansible-base/tor_v3_keys.json \
         build/*.deb
 
+    # Remove any Onion URL from make dev-tor
+    if docker volume sd-onion-services; then
```
zenmonkeykstop: Should this be `docker volume inspect sd-onion-services`? This is incorrect syntax (at least on my docker) so it will never return 0, and the volume will never be nuked.

zenmonkeykstop (changes requested):
One problem with the cleanup script, looking good so far otherwise.

devops/clean
deeplow: You're right. This should be addressed in 890020f, where I also suppressed the result to `/dev/null`.

deeplow:
Rebased and squased. This is ready to merge, pending the resolution of the discussion.

zenmonkeykstop (approved):

PR #7156, opened 2024-04-30
loaddata: Fix random_datetime() edge case

legoktm:
## Status

Ready for review

## Description of Changes

Instead of generating month and day separately, generate them at the same time by randomly selecting a day in the year, and then calculating the month and date based on that.

This should avoid randomly generating invalid dates like February 30th.

Fixes #7155.

## Testing

* [x] Visual review
* [ ] CI passes

## Deployment

Any special considerations for deployment? n/a

### If you made changes to the server application code:

- [ ] Linting (`make lint`) and tests (`make test`) pass in the development container

### If you made non-trivial code changes:

- [ ] I have written a test plan and validated it for this PR

Choose one of the following:

- [ ] These changes do not require documentation

Issue #7155, opened 2024-04-30
Tests failed because random_datetime generated invalid date on April 30


legoktm:
## Description

Tests failed because it randomly generated an invalid date.

## Steps to Reproduce

Wait until it is April 30th (March probably is the best time to trigger it), run tests a number of times

## Expected Behavior

They pass

## Actual Behavior

```
Traceback (most recent call last):
  File "./loaddata.py", line 483, in <module>
    load(parse_arguments())
  File "./loaddata.py", line 392, in load
    journalists = create_default_journalists()
  File "./loaddata.py", line 292, in create_default_journalists
    default_journalist = add_journalist("journalist", is_admin=True)
  File "./loaddata.py", line 149, in add_journalist
    attempt.timestamp = random_datetime(nullable=True)
  File "./loaddata.py", line 86, in random_datetime
    return datetime.datetime(
ValueError: day is out of range for month
```

https://github.com/freedomofpress/securedrop-client/actions/runs/8899756224/job/24440145469?pr=1980

## Comments

The `random_datetime` function does the following:
```python
    now = datetime.datetime.now()
    return datetime.datetime(
        year=random.randint(2013, now.year),
        month=random.randint(1, now.month),
        day=random.randint(1, now.day),
        hour=random.randint(0, 23),
        minute=random.randint(0, 59),
        second=random.randint(0, 59),
        microsecond=random.randint(0, 1000),
    )
```

I assume it randomly selected February as the month, and then picked a day that doesn't exist in February (30th, or 29th in a non-leap year).

A safer way to randomly generate the day in a safe manner is to generate a year, then generate a number between 1 and 365 (366 if a leap year) and then calculate month/day from that.

PR #7157, opened 2024-04-30
Ignore some safety alerts

legoktm:
## Status

Ready for review

## Description of Changes

* 66700 (CVE-2024-0690) - we only use `no_log` in tor-hidden-services where it is not in a loop
* 66777 (CVE-2023-6237) - not affected, as we build against system OpenSSL
* 66704 (CVE-2024-26130) - we should not be checking PKCS#12 keys/certs
* 66710 (CVE-2023-29483) - dev only, DoS isn't an issue

## Testing

* [x] visual review
* [x] CI passes

## Deployment

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -180,6 +180,10 @@ safety:  ## Run `safety check` to check python dependencies for vulnerabilities.
 		--ignore 65193 \
 		--ignore 65510 \
 		--ignore 65511 \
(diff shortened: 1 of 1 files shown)


cfm (approved):
CI passes now that #7155 (whose #7156 is blocked on _this_) does not apply.

PR #7158, opened 2024-05-03
add version of test key with no passphrase for convenience

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Adds a copy of the test private key without a passphrase or a filename that will trigger sequioa validity checks in CI. This is just to help with SDW development where the SD dev environment is being used to provide a server setup.


micahflee (approved):
I confirmed that this is indeed the correct passphraseless key for development!

PR #7163, opened 2024-05-24
Support podman in `make dev-tor`

legoktm:
## Status

Ready for review

## Description of Changes

Unfortunately `podman volume` and `docker volume` are not identical, podman will (sensibly) error if you try to create a volume that already exists unless you pass `--ignore`, which docker doesn't support.

So we conditionally add the needed flag depending on which tool we're using.

## Testing

* [x] Visual review
* [ ] Run `make dev-tor` twice under docker, make sure it works
* [ ] Run `make dev-tor` twice under podman, make sure it works (I've done this laready)

## Deployment

Any special considerations for deployment? n/a, dev only

## Checklist

### DIFF
--- a/securedrop/bin/dev-shell
+++ b/securedrop/bin/dev-shell
@@ -118,7 +118,7 @@ function docker_run() {
 
     if [ -n "${USE_TOR:-}" ]; then
         # Mount persistent onion services
-        docker volume create sd-onion-services
+        $DOCKER_BIN volume inspect sd-onion-services -f " " || $DOCKER_BIN volume create sd-onion-services
         DOCKER_RUN_ARGUMENTS="${DOCKER_RUN_ARGUMENTS} --volume sd-onion-services:/var/lib/tor/services"
     fi
 
--- a/devops/clean
+++ b/devops/clean
@@ -5,6 +5,14 @@
 set -e
 set -u
 
+USE_PODMAN="${USE_PODMAN:-}"
+
+# Allow opting into using podman with USE_PODMAN=1
+if  [[ -n "${USE_PODMAN}" ]]; then
+    DOCKER_BIN="podman"
+else
+    DOCKER_BIN="docker"
+fi
 
 function remove_unwanted_files() {
 
@@ -37,8 +45,8 @@ function remove_unwanted_files() {
         build/*.deb
 
     # Remove any Onion URL from make dev-tor
-    if docker volume inspect sd-onion-services > /dev/null; then
-        docker volume remove sd-onion-services
+    if $DOCKER_BIN volume inspect sd-onion-services > /dev/null; then
+        $DOCKER_BIN volume remove sd-onion-services
     fi
 
     # Remove extraneous copies of the git repos, pulled in


securedrop/bin/dev-shell
```diff
@@ -116,7 +116,8 @@ function docker_run() {
 
     if [ -n "${USE_TOR:-}" ]; then
         # Mount persistent onion services
-        docker volume create sd-onion-services
+        # shellcheck disable=SC2046
+        $DOCKER_BIN volume create sd-onion-services $(if [ "$DOCKER_BIN" = "podman" ]; then echo "--ignore"; fi)
```
deeplow: Unfortunately I don't see this option on my podman version (4.3.1) nor [on the docs](https://docs.podman.io/en/v4.4/markdown/options/volume.html). Maybe we can make it idempotent like this instead:

```suggestion
        $DOCKER_BIN volume inspect sd-onion-services -f " " || $DOCKER_BIN volume create sd-onion-services
```

deeplow (changes requested):
Thanks for finding this, at the time when I added the onion persistence I guess I did not realize podman was also supported. The solution didn't work for me so I proposed an alternative.

There's also another thing that needs to be provider-agnostic: the [make clean](https://github.com/freedomofpress/securedrop/pull/7124/files#diff-f069c7d8fd97be173c9a83b9c56b6cc869029c356cfd3ae8cff162d7efbe083cR41).

securedrop/bin/dev-shell
legoktm: Ah, it's documented at https://docs.podman.io/en/v4.4/markdown/podman-volume-create.1.html#ignore but yeah, it's new in 4.4. Will switch to your idea for 4.3 compat.

legoktm:
Thanks for the suggestion, now applied and fixed `make clean` as well.

deeplow:
Thanks for the fixes. I was reviewing this but got blocked by https://github.com/freedomofpress/securedrop/issues/7174.

deeplow:
> Thanks for the fixes. I was reviewing this but got blocked by https://github.com/freedomofpress/securedrop/issues/7174.

The above got resolved. However, I am having further challenges with podman, which seems to be beyond just `make dev-tor`, since it also fails with `make dev`.

Here's some truncated part of the output or `USE_PODMAN=1 make dev`

```
[...]

Exposed services will be available on localhost at
Source interface: http://127.0.0.1:8080
Journalist interface: http://127.0.0.1:8081
************************************************************
sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
Unable to replace /dev/random
Starting to build Rust code
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.25s
Copying libredwood.so into package

[...]
Successfully installed redwood-0.1.0
WARNING: You are using pip version 21.1.1; however, version 24.1.2 is available.
You should consider upgrading via the '/opt/venvs/securedrop-app-code/bin/python3 -m pip install --upgrade pip' command.
███ Generating securedrop/config.py...

sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
make: *** [Makefile:250: dev] Error 1
```

Do you know what could be going on? I was on a disposable debian-12 VM on Qubes. Then I installed `make` and `podman` and ran  `USE_PODMAN=1 make dev`.

deeplow (approved):
LGTM. Approving based on visual inspection alone since I [can't seem to run this](https://github.com/freedomofpress/securedrop/issues/7235). This a minor dev-facing (non-default) option, so I'd say this can be fine.

However, it would be nice if someone else could test:
>   * [ ]  Run `make dev-tor` twice under docker, make sure it works
>   * [ ]  Run `make dev-tor` twice under podman, make sure it works (I've done this laready)

legoktm:
(Just a rebase to pick up the new CI jobs)

PR #7175, opened 2024-06-05
feat: enable Tor's proof-of-work defense on the Source Interface

### LINKED ISSUE #6933: Enable Tor Proof-of-Work defenses for Onion Services
See https://blog.torproject.org/introducing-proof-of-work-defense-for-onion-services/ for details on the feature itself.

https://community.torproject.org/onion-services/advanced/dos/ contains documentation on the different configuration options we need to set.

One catch is that the PoW puzzles are LGPL, so a new `--enable-gpl` flag needs to be set at build time. There are 0.4.8.4 packages on deb.torproject.org now, but they don't have the GPL bits enabled, so we might need to do our own builds. I've commented on https://forum.torproject.org/t/proof-of-work-pow-defense-for-onion-service-is-released/8887 asking if they're going to provide GPL builds (my comment is still pending moderation).

In checklist form:
* [x] Get Tor packages that have all the bits for PoW defenses: https://github.com/freedomofpress/securedrop/issues/6937
* [ ] Update onion service configuration to enable PoW for SI (the JI is secret and protected by an auth key, so presumably doesn't need protection)
* [ ] Tune PoW flags as needed

cfm:
## Status

Ready for review

## Description of Changes

Closes #6933 by:
1. adding a `securedrop-admin sdconfig` prompt for enabling Tor's proof-of-work defense on the Source Interface (i.e., the only unauthenticated onion service a SecureDrop installation hosts); and
2. making it so.

## Testing

On this branch:

1. `securedrop-admin sdconfig`, accepting the new default:

    ```sh-session
    On the Source Interface, enable Tor's proof-of-work defense against denial-of-service attacks?: yes
    ```
2. `securedrop-admin install`
3. [x] `ssh app sudo cat /etc/tor/torrc` shows `HiddenServicePoWDefensesEnabled 1`
4. [x] [`onionprobe -e <your Source Interface>.onion`](https://tpo.pages.torproject.net/onion-services/ecosystem/apps/web/onionprobe/usage/) includes[^1] output like:

    ```sh-session
    2024-06-11 22:07:39,747 INFO: Proof of Work (PoW) params found in the descriptor
    2024-06-11 22:07:39,747 INFO: PoW v1 set with effort 0, expiration 2024-06-12T05:54:47 and seed Avq2iMON/49Vevxb+q479hUReRYKGAyw1fLScS/uKdE=
    ```
5. `securedrop-admin sdconfig` again, but turn the proof-of-work defense off
6. `securedrop-admin install` again
7. [x] `ssh app sudo cat /etc/tor/torrc` does NOT show `HiddenServicePoWDefensesEnabled 1`
8. [x] [`onionprobe -e <your Source Interface>.onion`](https://tpo.pages.torproject.net/onion-services/ecosystem/apps/web/onionprobe/usage/) ~~does NOT include `PoW` output~~ includes `Proof of Work (PoW) params not found`[^2]

[^1]: You'll need [onionprobe v1.2.0](https://gitlab.torproject.org/tpo/onion-services/onionprobe/-/blob/main/docs/changelog.md?ref_type=heads#v120---2024-04-24) [directly from Tor](https://community.torproject.org/onion-services/ecosystem/apps/web/onionprobe/installation/), not from the Debian bookworm package.
[^2]: https://github.com/freedomofpress/securedrop/pull/7175#issuecomment-2163949743

## Deployment

See discussion in this thread on half-opt-out:  This change will default to enabled on subsequent `securedrop-admin {sdconfig,install}` runs, but it won't be applied automatically on server-side update.

https://github.com/freedomofpress/securedrop/pull/7175#issuecomment-2164114278

Choose one of the following:

- These changes do not require documentation

Deferred to freedomofpress/securedrop-docs#568.

### DIFF
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -376,6 +376,15 @@ class SiteConfig:
                 lambda config: True,
             ),
             (
+                "securedrop_app_pow_on_source_interface",
+                True,
+                bool,
+                "Whether Tor proof of work should be enabled on Source Interface",
+                SiteConfig.ValidateYesNo(),
+                lambda x: x.lower() == "yes",
+                lambda config: True,
+            ),
+            (
                 "securedrop_app_https_on_source_interface",
                 False,
                 bool,
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -379,7 +379,8 @@ class SiteConfig:
                 "securedrop_app_pow_on_source_interface",
                 True,
                 bool,
-                "Whether Tor proof of work should be enabled on Source Interface",
+                "Enable Tor's proof-of-work defense against denial-of-service attacks for the "
+                "Source Interface?",
                 SiteConfig.ValidateYesNo(),
                 lambda x: x.lower() == "yes",
                 lambda config: True,
@@ -388,7 +389,7 @@ class SiteConfig:
                 "securedrop_app_https_on_source_interface",
                 False,
                 bool,
-                "Whether HTTPS should be enabled on " + "Source Interface (requires EV cert)",
+                "Enable HTTPS for the Source Interface (requires EV certificate)?",
                 SiteConfig.ValidateYesNo(),
                 lambda x: x.lower() == "yes",
                 lambda config: True,
--- a/install_files/ansible-base/roles/tor-hidden-services/templates/torrc
+++ b/install_files/ansible-base/roles/tor-hidden-services/templates/torrc
@@ -6,6 +6,10 @@ RunAsDaemon 1
 HiddenServiceDir /var/lib/tor/services/sourcev3
 HiddenServicePort 80 127.0.0.1:80
 
+{% if securedrop_app_pow_on_source_interface|default(True) %}
+HiddenServicePoWDefensesEnabled 1
+{% endif %}
+
 {% if securedrop_app_https_on_source_interface|default(False) %}
 HiddenServicePort 443 127.0.0.1:443
 {% endif %}
--- a/install_files/ansible-base/roles/tor-hidden-services/templates/torrc
+++ b/install_files/ansible-base/roles/tor-hidden-services/templates/torrc
@@ -6,7 +6,7 @@ RunAsDaemon 1
 HiddenServiceDir /var/lib/tor/services/sourcev3
 HiddenServicePort 80 127.0.0.1:80
 
-{% if securedrop_app_pow_on_source_interface|default(True) %}
+{% if securedrop_app_pow_on_source_interface|default(False) %}
 HiddenServicePoWDefensesEnabled 1
 {% endif %}
 
--- a/admin/tests/test_integration.py
+++ b/admin/tests/test_integration.py
@@ -41,6 +41,7 @@ securedrop_app_https_certificate_cert_src: ''
 securedrop_app_https_certificate_chain_src: ''
 securedrop_app_https_certificate_key_src: ''
 securedrop_app_https_on_source_interface: false
+securedrop_app_pow_on_source_interface: true
 securedrop_supported_locales:
 - de_DE
 - es_ES
@@ -73,6 +74,7 @@ securedrop_app_https_certificate_cert_src: ''
 securedrop_app_https_certificate_chain_src: ''
 securedrop_app_https_certificate_key_src: ''
 securedrop_app_https_on_source_interface: false
+securedrop_app_pow_on_source_interface: true
 securedrop_supported_locales:
 - de_DE
 - es_ES
@@ -81,7 +83,7 @@ smtp_relay_port: 587
 ssh_users: sd
 """
 
-HTTPS_OUTPUT = """app_hostname: app
+HTTPS_OUTPUT_NO_POW = """app_hostname: app
 app_ip: 10.20.2.2
 daily_reboot_time: 5
 dns_server:
@@ -105,6 +107,7 @@ securedrop_app_https_certificate_cert_src: sd.crt
 securedrop_app_https_certificate_chain_src: ca.crt
 securedrop_app_https_certificate_key_src: key.asc
 securedrop_app_https_on_source_interface: true
+securedrop_app_pow_on_source_interface: false
 securedrop_supported_locales:
 - de_DE
 - es_ES
@@ -188,9 +191,16 @@ def verify_app_gpg_key_prompt(child):
     )
 
 
+def verify_tor_pow_prompt(child):
+    # We don't need child.expect()'s regex matching, but the prompt is too long
+    # to match on the whole thing.
+    child.expect_exact("Enable Tor's proof-of-work defense", timeout=2)
+
+
 def verify_https_prompt(child):
-    child.expect(
-        rb"Whether HTTPS should be enabled on Source Interface \(requires EV cert\)\:", timeout=2
+    # We don't need child.expect()'s regex matching.
+    child.expect_exact(
+        "Enable HTTPS for the Source Interface (requires EV certificate)?:", timeout=2
     )
 
 
@@ -312,6 +322,9 @@ def test_sdconfig_on_first_run():
     child.sendline("")
     verify_app_gpg_key_prompt(child)
     child.sendline("\b" * 14 + "sd_admin_test.pub")
+    verify_tor_pow_prompt(child)
+    # Default answer is yes
+    child.sendline("")
     verify_https_prompt(child)
     # Default answer is no
     child.sendline("")
@@ -375,7 +388,11 @@ def test_sdconfig_enable_journalist_alerts():
     child.sendline("")
     verify_app_gpg_key_prompt(child)
     child.sendline("\b" * 14 + "sd_admin_test.pub")
+    verify_tor_pow_prompt(child)
+    # Default answer is yes
+    child.sendline("")
     verify_https_prompt(child)
+    child.sendline("")
     # Default answer is no
     child.sendline("")
     verify_app_gpg_fingerprint_prompt(child)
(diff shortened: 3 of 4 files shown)


admin/securedrop_admin/__init__.py
```diff
@@ -375,6 +375,15 @@ def __init__(self, args: argparse.Namespace) -> None:
                 None,
                 lambda config: True,
             ),
+            (
+                "securedrop_app_pow_on_source_interface",
+                True,
+                bool,
+                "Whether Tor proof of work should be enabled on Source Interface",
```
zenmonkeykstop: nit: Said desc should probably be more explicit about the purpose, like "Enable Tor's PoW-based DoS defense on Source Interface?:"

(I know the OG wording is more consistent with the next option, but honestly I'd change it to fit this pattern not the other way around)

admin/securedrop_admin/__init__.py
```diff
@@ -375,6 +375,15 @@ def __init__(self, args: argparse.Namespace) -> None:
...
+                SiteConfig.ValidateYesNo(),
+                lambda x: x.lower() == "yes",
+                lambda config: True,
+            ),
```
zenmonkeykstop, line 387: If the user enables PoW, we could also allow them to tweak the two associated configs (with sensible default)s:
- HiddenServicePoWQueueRate (250)
- HiddenServicePoWQueueBurst (2500)

So something like:
```
            (
                "securedrop_app_tor_pow_queue_rate",
                "",
                str,
                "Set PoW Queue rate (default: 250):",
                SiteConfig.ValidateOptionalPath(self.args.ansible_path),
                None,
                lambda config: config.get("securedrop_app_pow_on_source_interface"),
            ),
```
etc

install_files/ansible-base/roles/tor-hidden-services/templates/torrc
```diff
@@ -6,6 +6,10 @@ RunAsDaemon 1
 HiddenServiceDir /var/lib/tor/services/sourcev3
 HiddenServicePort 80 127.0.0.1:80
 
+{% if securedrop_app_pow_on_source_interface|default(True) %}
+HiddenServicePoWDefensesEnabled 1
```
zenmonkeykstop, line 10: Adding in the two directives mentioned above would be good, again specifying default values just in case.

zenmonkeykstop (dismissed):
Adding in the PoW tuning directives now might save us doing another release down the line. Otherwise this is straightforward and a good first step.

rocodes:
My opinion on the questions posed:
- Opt-out (default enabled), pending upstream confirmation and assuming we've had any remaining questions answered by Tor team (re performance concerns) would be what I'd suggest. 
- Timeout configuration: if there's any possibility of a footgun, we should validate inputs if we even surface that level of configuration to the end user (re maintainability/support, I'm of half a mind to say "nope, that's buried in site-specific if you absolutely think you want to tweak things you shouldn't be tweaking", but I think that would depend on what Tor says about those values. If there are good reasons for average users to adjust these values I'd change my mind) 
- Testing/failure modes: good q / open q.

zenmonkeykstop:
Opt-out would be nice but means changing the tor config via a deb postint (likely securedrop-app-code as we only need to update app) - that's high-impact if it goes awry for users relying on Tor for access to the server.

Ideally site admins should also be monitoring Tor and adjusting those values constantly in response to a targeted DoS - I don't think we can get the monitoring setup in place in a reasonable timeframe, but giving admins the ability to tweak without futzing would at least be something.

legoktm:
I think this is a reasonable approach given the various considerations and the amount of time we want to spend on it, thanks for working on it :)

cfm:
(Marking "ready for review" just to enable notifications from GitHub to Slack.  Not *quite* ready for review.)

After reviewing Tor's [example configuration](https://community.torproject.org/onion-services/ecosystem/technology/pow/#example-configuration) and playing with a [self-contained example onion service](https://gitlab.torproject.org/tpo/onion-services/onion-pow-example), **I don't think we should expose `sdconfig` knobs for `HiddenServicePoWQueue{Rate,Burst}` at all,** for two reasons.  First, Tor will *always* be the bottleneck for a SecureDrop instance; we have no other performance or scaling knob that would suggest "lower or higher values depending on the overall capacity of your system to handle requests and with the current load of your system".[^1]  Second:

> The service starts with a default suggested effort value of 0, which keeps the PoW defenses dormant until it notices signs of overload.
> 
> The system is dynamic: as time goes by, the server regularly updates the suggested effort based on its load.[^2]

This feedback loop is going to be more responsive than any administrator's fine-tuning, and we have no precedent for exposing knobs (or footguns, per @rocodes) like these in `sdconfig` for any other SecureDrop component.

However, I'll write a test plan here and for v2.9.0 QA that includes monitoring Tor's actual performance with `HiddenServicePoWDefensesEnabled` via `MetricsPort`.

Thanks for your other feedback, @zenmonkeykstop, in <https://github.com/freedomofpress/securedrop/pull/7175#pullrequestreview-2100160381>, which I'll address now.  And thanks for reminding me that this is "one-half opt-in": it'll only take effect the next time an administrator runs a full `securedrop-admin {sdconfig,install}`.

[^1]: https://community.torproject.org/onion-services/ecosystem/technology/pow/#example-configuration
[^2]: https://community.torproject.org/onion-services/ecosystem/technology/pow/#how-does-it-work-in-simple-terms

admin/securedrop_admin/__init__.py
cfm: I don't think we either need to or should expose these knobs, per <https://github.com/freedomofpress/securedrop/pull/7175#issuecomment-2161733057>, but feel free to overrule me.  :-)

install_files/ansible-base/roles/tor-hidden-services/templates/torrc
cfm: Same as <https://github.com/freedomofpress/securedrop/pull/7175#discussion_r1635570027>.

admin/securedrop_admin/__init__.py
cfm: Done in f298448acdfc811599a846a385539b40b631d402.

cfm:
Feedback addressed.  Tests and test plan forthcoming.

admin/securedrop_admin/__init__.py
```diff
@@ -375,11 +375,21 @@ def __init__(self, args: argparse.Namespace) -> None:
                 None,
                 lambda config: True,
             ),
+            (
+                "securedrop_app_pow_on_source_interface",
+                True,
+                bool,
+                "On the Source Interface, enable Tor's proof-of-work defense against "
```
nathandyer: Would "Enable Tor's proof-of-work defense against denial-of-service attacks for the Source Interface" be more straight-forward?

admin/securedrop_admin/__init__.py
```diff
@@ -375,11 +375,21 @@ def __init__(self, args: argparse.Namespace) -> None:
...
+                SiteConfig.ValidateYesNo(),
+                lambda x: x.lower() == "yes",
+                lambda config: True,
+            ),
             (
                 "securedrop_app_https_on_source_interface",
                 False,
                 bool,
-                "Whether HTTPS should be enabled on " + "Source Interface (requires EV cert)",
+                "On the Source Interface, enable HTTPS? (Requires EV certificate.)",
```
nathandyer: Similar question as above: is "Enable HTTPS for the Source Interface (Requires EV certificate.)" clearer?

nathandyer:
Just stepped through the test plan for this branch, and am happy to report that everything worked beautifully!

Just a couple small caveats to note, regarding deviations from expected output and a possible stumbling point for any additional testers:

1. `onionprobe` is not installed by default in Tails. After `sudo apt install onionprobe` I was able to run it, but it did not work. It kept giving a `CRITICAL: Error initializing Tor` error after a 90 second timeout. I discovered this is a [bookworm-related bug](https://gitlab.torproject.org/tpo/onion-services/onionprobe/-/issues/81), and is fixed in the 1.1.2 release (the version available in Tails 6.3 is 1.0.0). 

    To work around this, I installed the 1.2.0 version available in `sid` on my daily driver machine, since the `onionprobe` step didn't strictly need to be performed from my Admin Workstation.

2. >  8 [x] [`onionprobe -e <your Source Interface>.onion`](https://tpo.pages.torproject.net/onion-services/ecosystem/apps/web/onionprobe/usage/) does NOT include `PoW` output

    is not accurate to what I saw; instead, it reported back:

    ```
    2024-06-12 17:33:39,984 INFO: Proof of Work (PoW) params not found in the descriptor
    ```

    which I do believe _is_ what we would expect (and want) to see there

admin/securedrop_admin/__init__.py
cfm: Yes, this is the correct happy medium between the original and my text.  :-)  Fixed in 40f18b915!

cfm:
Thanks for testing, @nathandyer!  I've addressed your feedback and updated the test plan with your clarification of the expected output, which I agree still passes as intended.

cfm:
Rebased from `develop` after #7179.  I've updated the `admin` test suite for the addition of the `securedrop_app_pow_on_source_interface` `site-specific` variable.  [We apparently don't verify the onion-service configuration in testinfra](https://github.com/freedomofpress/securedrop/blob/f92ecce51ed4033ec2ad5087275f9ea3f8efb795/molecule/testinfra/app/test_tor_hidden_services.py#L10-L11), so I've not added anything there.  CI `lint` is currently failing on a flaky Ubuntu mirror, but I'll kick it when CircleCI lets me.

cfm:
@zenmonkeykstop, with @nathandyer's review today and tests passing, this is now ready for your review and merge tomorrow morning.

install_files/ansible-base/roles/tor-hidden-services/templates/torrc
```diff
@@ -6,6 +6,10 @@ RunAsDaemon 1
 HiddenServiceDir /var/lib/tor/services/sourcev3
 HiddenServicePort 80 127.0.0.1:80
 
+{% if securedrop_app_pow_on_source_interface|default(True) %}
```
zenmonkeykstop: The default(True) means that admins who  run `install` on an existing system without running `sdconfig` will have PoW enabled. About the only scenario where one would do that on purpose would be to apply the existing expected state, so the default should probably be False.

zenmonkeykstop (changes requested):

install_files/ansible-base/roles/tor-hidden-services/templates/torrc
cfm: Good point, @zenmonkeykstop.  Fixed in 02a81f1421496deeea7be2d22eac99e59854747a.

zenmonkeykstop (approved):

Issue #6897, opened 2023-07-26
Running securedrop-admin tests locally is painful


legoktm:
I'm trying to debug the test failures in #6885 and it's kind of painful to debug these test failures. It builds a 6GB container (in comparison the SD container is 2-3GB), and then if you make a change and want to re-run the tests, it needs to rebuild half the container, which takes a good 5 minutes. The tests run under tox so all the dependencies are installed 3 times unnecessarily, etc. I ran out of disk space on my 6th test run.

The container is so big because it's apparently copying the entire securedrop tree into the container, including stuff like `target/` and not just the admin folder.

Ideally, this would reuse the same dev-shell stuff we have for the server parts and just parameterize the container being used, and not use tox.

legoktm:
@cfm has proposed a workaround of sorts in https://github.com/freedomofpress/securedrop/pull/7181.

PR #7181, opened 2024-06-13
docs(test): suggest a workaround for the slow admin test suite

cfm:
## Status

Ready for review

## Description of Changes

"Fixes with documentation" the fact that the admin test suite runs in a container that's rebuilt very slowly on every change.  Given how rarely we interact with it directly (as opposed to in CI), I don't think that's worth fixing, but it is worth hinting at a quality-of-life workaround.

Refs https://github.com/freedomofpress/securedrop/issues/6897.

## Testing

- [ ] Visual review.

## Deployment

Development/test-only.

### DIFF
--- a/admin/Makefile
+++ b/admin/Makefile
@@ -2,6 +2,7 @@ DEFAULT_GOAL: help
 
 .PHONY: test
 test:  ## Run tox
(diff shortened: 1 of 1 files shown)


legoktm (approved):
Cross-referenced with the issue I filed about this last year 😭 but this is better than the status quo. Thanks!

Issue #7168, opened 2024-05-24
Release Securedrop 2.9.0


zenmonkeykstop:
This is a tracking issue for the release of SecureDrop 2.9.0

Tentatively scheduled as follows:

**Pre-release announcement:** TK
**Release date:** TK (no later than 2024-07-05)

**Release manager:** @zenmonkeykstop  
**Deputy release manager:** @legoktm 
**Localization manager:** n/a (no string changes)
**Communications manager:** @nathandyer 

_SecureDrop maintainers and testers:_ As you QA 2.9.0, please report back your testing results as comments on this ticket. File GitHub issues for any problems found, tag them "QA: Release".

Test debian packages will be posted on https://apt-test.freedom.press signed with [the test key](https://gist.githubusercontent.com/conorsch/ec4008b111bc3142fca522693f3cce7e/raw/2968621e8ad92db4505a31fcc5776422d7d26729/apt-test%2520apt%2520pubkey).

# [QA Matrix for 2.9.0](https://docs.google.com/spreadsheets/d/1suShRhMPmYc9t7vGxifJ7AFP5ERwptP29JfPoXBEfMo/edit#gid=96348658)
# [Test Plan for 2.9.0](https://github.com/freedomofpress/securedrop/wiki/2.9.0-Test-Plan)

# Prepare release candidate (2.9.0~rc1)
- [x] Link to latest version of Tails, including release candidates, to test against during QA
- [x] Prepare 2.9.0~rc1 release changelog
- [x] Branch off release/2.9.0 from develop
- [x] Prepare 2.9.0-rc1
- [x] Build debs, preserving build log, and put up `2.9.0~rc1` on test apt server
- [x] Commit build log.

# Prepare release candidate (2.9.0~rc2)

- [x] Prepare 2.9.0~rc2 release changelog
- [x] Prepare 2.9.0-rc2
- [x] Build debs, preserving build log
- [x] Commit build log
- [x] Put up `2.9.0~rc2` on test apt server.

# Prepare release candidate (2.9.0~rc3)

- [x] Prepare 2.9.0~rc3 release changelog
- [x] Prepare 2.9.0-rc3
- [x] Build debs, preserving build log
- [x] Commit build log
- [ ] Put up `2.9.0~rc3` on test apt server.

After each test, please update the QA matrix and post details for Basic Server Testing, Application Acceptance Testing and release-specific testing below in comments to this ticket.

# Final release
[...]

PR #7193, opened 2024-06-27
Bump Securedrop version to 2.9.0 for release.

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Towards #7168 .

Increments version to 2.9.0, merges RC changelogs.

## Testing
- [ ] version bump successful
- [ ] changelog is accurate.


changelog.md
```diff
@@ -1,35 +1,17 @@
...
-
-### Web applications
-* Added accessibility improvements (#6536)
-
 ### Journalist API
-* Added support for HTTP range request for submissions and replies endpoints (#7160)
+* Added support for partial content requests for submissions and replies endpoints (#7160, #7189)
 
 ### Operations
+* Added opt-in Tor PoW defense for the Source Interface (#7175)
```
cfm, line 14: Per <https://github.com/freedomofpress/securedrop-docs/blob/2c7d9ab8fd36626ec4edd5b91c4590f079dcd5a0/docs/admin/deployment/tor_pow.rst?plain=1#L13-L15>, I think this is worth making more explicit:

```suggestion
* Added Tor PoW defense for the Source Interface: opt-in for existing instances; on by default for new instances (#7175)
```

Or just `Added Tor PoW defense for the Source Interface`.

cfm (changes requested):
One clarification inline for your consideration, @zenmonkeykstop.  Looks good otherwise.

changelog.md
zenmonkeykstop: Eh - that's a bit too much in the weeds for a changelog IMO. New instances are still explicitly required to make the choice in sdconfig so I'm comfortable with current language.

cfm: That's fair.  I think you and I are parsing this language differently versus the current documentation—thus the second suggestion offering less rather than more detail to avoid that confusion altogether.  But if you're happy with this as it is I'll approve for your merge.

cfm (approved):
Approving for your merge, @zenmonkeykstop.

PR #7207, opened 2024-08-01
Add `make dev-arti` as an alternative to C-Tor

legoktm:
## Status

Ready for review

## Description of Changes

Arti is Tor's RIIR effort, and while it is 1.x, the onion service support is still experimental.

Given that, it is to our (and Tor's) benefit for us to try using Arti in development modes to provide feedback. To that end, using Arti is now possible, in parallel, to the existing C-Tor support.

Running `make dev-arti` should give exactly the same experience as the existing `make dev-tor` endpoint, aside from the inital launch, which will be slower since it needs to compile Arti from scratch.

As an intentional choice, we don't version pin Arti since this is an experimental thing, so it's more beneficial we just automatically stay on the latest version and fix breakage instead of needing to constantly bump versions.

Note that `make dev-tor` is entirely independent from this, so the two commands will serve the dev server on different onion addresses.

As far as the Arti code itself, generating a config.toml file seems much cleaner than adding to `torrc` and generating `.auth` files.

## Testing

* [ ] Without this PR, run `make dev-tor`, see that it's all fine
* [ ] With this PR, run `make dev-tor`, see that it uses the same onion addresses and is all fine. Basically this PR should not adversely affect the existing dev-tor in any way.
* [ ] With this PR, run `make dev-arti`, wait a bit for it to all compile, and then it works in just about the same way.
* [ ] Stop the dev server, and then run `make dev-arti` again, verify it gives you the same onion addresses as the previous run.

## Deployment

Any special considerations for deployment? n/a, dev-only

## Checklist

While this modifies the dev container, I explicitly didn't try anything in it yet because this is just for interactive use and not for like, running tests and stuff.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -249,6 +249,13 @@ dev-tor:  ## Run the development server with onion services in a Docker containe
 	@OFFSET_PORTS='false' DOCKER_BUILD_VERBOSE='true' USE_TOR='true' SLIM_BUILD=1 $(DEVSHELL) $(SDBIN)/run
 	@echo
 
+.PHONY: dev-arti
+dev-arti:  ## Same as `dev-tor`, but using Arti instead of C-Tor
+	@echo "███ Starting development server with onion services..."
+	@echo "███ WARNING: Arti support is experimental. Good luck!"
+	@USE_ARTI='true' $(MAKE) dev
+	@echo
+
 .PHONY: demo-landing-page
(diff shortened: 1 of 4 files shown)


legoktm:
Marking this as ready for review; I've reworked it to no longer replace `make dev-tor` entirely, but rather to be an alternative for it.

eloquence:
Closing for now, pending further use of Arti in the project.

PR #7196, opened 2024-06-28
Enable deprecated docker pull functionalty to fix nightlies.

zenmonkeykstop:
## Status

RfR

## Description of Changes

Fixes #7195 .

- Enables deprecated docker pull feature to fix `staging-tests-with rebase` CI job, run both nightly and for branches prefixed with `stg-`

## Testing

- [ ] CI is passing, in particular `staging-test-with-rebase`

## Deployment
n/a - CI only.

PR #7215, opened 2024-08-13
Use upstream gcloud-sdk image instead of our oudated one

### LINKED ISSUE #7195: Nightly CircleCI staging-test-with-rebase jobs are failing
@mig5 tracked down the underlying issue to reliance on a Docker feature that has been deprecated. As a quick fix we can re-enable it. Long-term we should update the job to work with the latest version of the `gcloud-sdk` image. Longer-term we could move this job elsewhere altogether - either to another platform with support for nested virtualization or to a bare metal provider ala https://github.com/cfm/terraform-metal-securedrop-staging

legoktm:
## Status

Ready for review

## Description of Changes

We have not updated our gcloud-sdk image since 2018 and it's really not worth us maintaining this when Google provides its own image.

We now authenticate using the method documented at <https://cloud.google.com/sdk/docs/downloads-docker>, in which we invoke the container to login, which creates a volume that can be used in future calls. This means we no longer need to have a running container that we exec into; we can just create a new container each time we want to make a gcloud call.

Fixes #7195.
Fixes #7196.

## Testing

* [ ] CI, especially staging job, passes
* [ ] Visual review

## Deployment

## Checklist


devops/gce-nested/ci-env.sh
```diff
@@ -48,26 +48,25 @@ function generate_gce_creds_file() {
             fi
         fi
     fi
+
+    if ! docker inspect --format '{{ .Id }}' gcloud-config > /dev/null 2>&1; then
```
cfm, line 53:
```suggestion
    # Prime the "gcloud-config" container by authenticating from $GCE_CREDS_FILE.
    if ! docker inspect --format '{{ .Id }}' gcloud-config > /dev/null 2>&1; then
```

devops/gce-nested/ci-env.sh
```diff
@@ -48,26 +48,25 @@ function generate_gce_creds_file() {
...
+            --env="CLOUDSDK_COMPUTE_ZONE=${CLOUDSDK_COMPUTE_ZONE}" \
+            --volume "${GCE_CREDS_FILE}:/gce-svc-acct.json" \
+            --name gcloud-config \
+            gcr.io/google.com/cloudsdktool/google-cloud-cli:stable \
+            gcloud auth activate-service-account --key-file /gce-svc-acct.json
+    fi
 }
 
 # Wrapper function to communicate with the gcloud API. Ensure gcloud-sdk
 # container is running, and if so, pass all args to it.
```
cfm:
```suggestion
# Each gcloud_call() invocation just runs a fresh (note "--rm") instance of the "google-cloud-cli" container primed from "gcloud-config" above.
```

cfm (changes requested):
This is a nice improvement both operationally and conceptually.  Because it changes the conceptual model of this CI workflow, I think it's worth documenting further, so I've proposed two comments inline.

legoktm:
Thanks, accepted both.

cfm (approved):
Great.  Thanks, @legoktm.

PR #7222, opened 2024-08-15
Migrate translation-tests to GitHub Actions

legoktm:
## Status

Ready for review

## Description of Changes

Translation tests are slow because it's our normal tests just run 20+ times, once for each language. We can use a dynamic matrix to generate separate jobs pretty easily (I learned this pattern from https://github.com/suzuki-shunsuke/example-github-actions-dynamic-matrix).

This job is a little special as we only run it in specific circumstances. Two are easy to handle 1) branch filter and 2) scheduled run.

But only triggering on PRs from "weblate-fpf" is a bit harder, so we add some run-time logic in the locales job itself.

## Testing

* [x] Visual review
* [x] CI for `push` passes and runs full Translation jobs
* [x] CI for `pull request` runs locales job but skips Translation / test jobs

## Deployment


cfm (approved):
Gorgeous.

https://github.com/freedomofpress/securedrop/blob/92798a6cce212922c66138a1ab6555e8dbfaac6e/.github/workflows/translation.yml#L49-L50

Thanks, @legoktm.

PR #7224, opened 2024-08-15
Enforce GCE lifespan during instantiation

leee:
## Status

Ready for review

## Description of Changes

FPF used to have a GCE instance culler tool:
 https://github.com/freedomofpress/container-gce-instance-culler
which was packed into a container and run as a cronjob every other hour on Codefresh. Instead of trying to maintain said tool and Codefresh pipeline, define a lifespan during VM creation that GCP will enforce.

## Testing

- [ ] Visual review


legoktm (approved):
Thanks! Will wait for CI to pass before merging.

For reference, the relevant GCE docs are at <https://cloud.google.com/compute/docs/instances/limit-vm-runtime>.

PR #7233, opened 2024-09-19
Upgrade ruff, remove black

legoktm:
## Status

Ready for review

## Description of Changes

This is split into three commits for ease of review:

#### Upgrade ruff, remove black

Upgrade ruff to the latest version so we can use it for formatting and
native editor integration. While we're at it, introduce the standard
`make fix` target
(<https://github.com/freedomofpress/securedrop-tooling/issues/11>).

Actually fixing new issues will happen in follow-ups to make this change
easier to review.

Fixes #7226.

#### Apply `ruff format .`

This is all automated except test_manage.py was manually fixed to remove
extraneous commas that ruff tried to turn into tuples.

#### Fix most errors spotted by ruff

About half of these fixes were automatic, the rest I did by hand.
    
I skipped removing % formatting from pretty_bad_protocol since there
were like a hundred of them and it wasn't really worth it.

## Testing

How should the reviewer test this PR?

* [x] CI passes
* [x] visual review

## Deployment

Any special considerations for deployment? n/a

## Checklist

- [x] Linting (`make lint`) and tests (`make test`) pass in the development container
- [x] Linting and tests (`make -C admin test`) pass in the admin development container

PR #7234, opened 2024-09-23
Speed up `update-python3-dependencies`

legoktm:
## Status

Ready for review

## Description of Changes

The `make update-python3-dependencies` step is slow for two main reasons: container image layering and pip-compile.

Switch to the new uv tool, which reimplements pip-compile in a much faster way. The output is basically the same, except the sorting is smarter (e.g. pytest comes before pytest-cov) and package names are properly normalized. We can also drop the `--allow-unsafe` because uv is entirely independent of setuptools and pip-tools.

Because uv is independent of the system Python (it's pure Rust), we can run it on the host and outside the container, so no image rebuilding happens during the updates, it'll just be needed once afterwards.

uv is still quite new to the Python ecosystem, but this allows us to begin using it without any lock-in, it should be trivial to swap back to pip-tools if needed.

Overall `make update-python3-dependencies` now takes seconds to run instead of minutes \o/

Refs <https://github.com/freedomofpress/securedrop-tooling/issues/16>.

## Testing

* [ ] Visual review; no changes to dependencies outside of removal of pip-tools
* [ ] Run `make update-python3-dependencies` yourself.

## Deployment

## Checklist

  - I'll do this post-merge


legoktm:
This is a rather conservative PR, there are other places we could be swapping in `uv` for speed like in CI or in admin/, but I think we can start with this for now and expand it over time.

cfm (commented):
This looks like a great improvement.  Given the relatively privileged place `pip-compile` or `uv` occupies in our toolchain—as the producer of hashes for build-time verification—I would question the adoption of uv if it were closed-source.  (Or at least I'd want us to do something like verify *its* output with `pip-compile` in CI.)  But since it's (a) open-source, (b) format-compatible, and (c) as easily swapped out as you've swapped it in here, I don't much care how it's funded.  :-)

I'll leave this open for other feedback for twenty-four hours.  Barring objections, I'll approve and merge it tomorrow.

cfm:
Ah, @legoktm, do you mind rebasing after #7233?

cfm:
Discussed further out of band: no objections.

legoktm:
I'm gonna make one change, we don't actually need to run this in the container anymore, so it'll be even faster.

cfm (approved):
Reviewed the diff in <https://github.com/freedomofpress/securedrop/compare/b21cdf23b4a53222a74feb1417935fff23585926..37bb2cf3523726128881f47cc71633e853163b30>.  Very nice!

PR #7247, opened 2024-10-11
Remove obsolete dependencies from package building

legoktm:
## Status

Ready for review

## Description of Changes

These packages aren't available in noble, but also aren't needed when building in focal:

* dh-systemd was merged into debhelper itself
* sqlite is actually sqlite2 and not sqlite3, but regardless, package building doesn't require the sqlite command-line tool

This is a no-op for the packages themselves.

## Testing

* [ ] CI passes
* [ ] If you build without this change and build with it, the diffoscope should look something like https://gist.github.com/legoktm/7f7f40327d0ff40f481df0b975db8268, which importantly shows no systemd nor sqlite related changes.

## Deployment

## Checklist

### DIFF
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -11,7 +11,6 @@ RUN apt-get -y update && apt-get upgrade -y && apt-get install -y \
         debhelper \
         devscripts \
         dh-python \
-        dh-systemd \
         gdb \
         git \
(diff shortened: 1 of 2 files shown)


zenmonkeykstop (approved):
LGTM based on visual review, builds seem unaffected.

PR #7257, opened 2024-10-17
[2.10.1] Backport "Translations update from Weblate"

cfm:
## Status

Ready for review

## Description of Changes

Backport of #7143.

## Testing

* [x] CI is passing
* [x] base is `release/2.10.1`
* [x] Only contains changes from #7143 and #7259
    - *Except* the merge commit 6e3d7bed17537fa8f69f941f21a8eec34e8da113


cfm:
That was the right *set* of commits.  Let's try the right *sequence* of commits now....

install_files/ansible-base/roles/tails-config/templates/locale/hi/LC_MESSAGES/messages.po
```diff
@@ -48,4 +37,4 @@ msgstr "SecureDrop स्रोत अंतराफलक"
 
 #. Icon
 msgid "{{ tails_config_securedrop_dotfiles }}/securedrop_icon.png"
-msgstr ""
+msgstr "सभी क्षेत्रो मे हम एक ही तरह के आइकॉन फाइल प्रयोग करते है"
```
zenmonkeykstop: IIRC This should be a path but instead translates to "we use a single icon file in all regions"

zenmonkeykstop (changes requested):
Changes are confined to the expected subset of commits, but there's a text string in place of a filepath in the Hindi desktop shortcut strings. This should be fixed - users with Hindi setas their language in Tails will at best have a broken icon and at worst undefined behaviours :ghost:

install_files/ansible-base/roles/tails-config/templates/locale/hi/LC_MESSAGES/messages.po
zenmonkeykstop: On further examination I'm not sure we're actually using this string - the path is just hardcoded as far as I can tell. If we can confirm that then this is probably not a blocker (but we should remove unused strings from translation so as to avoid confusion and delay).

cfm: Discussed out of band: This winds up being a no-op when the desktop file is [rendered](https://github.com/freedomofpress/securedrop/blob/7e622c654888f6c3418fa1c07509ab4cb044efaf/install_files/ansible-base/roles/tails-config/templates/desktop-journalist-icon.j2#L31) from these translation catalogs, but it's exactly the kind of potentially behavior-altering string change we want to catch.  I'm going to:
1. [x] check Weblate to see why Weblate's own linting didn't catch it;
2. [x] fix this string in Weblate;
3. [x] see if we can mark it read-only to prevent this class of error in the future;
4. [x] bring the fix into `develop`; and then
5. [x] cherry-pick the fix here for final review.

cfm: Weblate is able to enforce the Python-brace formatting we use in general but not the Jinja templating used here.  I've enforced the expected string in <https://github.com/freedomofpress/securedrop/pull/7259#pullrequestreview-2375805844>, which I've cherry-picked in here.  Notice these changes pass linting without `make update-desktop-files`, which confirms that they were a no-op for the rendered desktop file.

zenmonkeykstop (approved):

zenmonkeykstop:
LGTM once CI is good, we should make a note to be sure we include the Hindi and Arabic translator in the acknowledgements, as you're now the most recent updater there :)

PR #7262, opened 2024-10-17
chore: update version to 2.10.1

cfm:
## Status

Ready for review

## Description of Changes

Towards #7239.

## Testing

- [x] Base branch is `release/2.10.1`.
- [ ] Visual review.

PR #7263, opened 2024-10-18
feat: pass default "SecureDrop X.Y.Z" commit message to `git commit` and `git tag`

cfm:
## Status

Ready for review

## Description of Changes

Tries to prevent things like <https://github.com/freedomofpress/securedrop/pull/7262#issuecomment-2420318902> by making our convention an explicit default.

## Testing

Visual review.  `update_version.sh 3.0.0` if you feel like it; just make sure you `git tag -d 3.0.0` afterwards to avoid side effects from an errant `git push --tags` in the future.

## Deployment

Developer-only; no deployment considerations.

### DIFF
--- a/update_version.sh
+++ b/update_version.sh
@@ -1,7 +1,10 @@
 #!/bin/bash
 # shellcheck disable=SC2230
 
-## Usage: ./update_version.sh <version>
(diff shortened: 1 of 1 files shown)


legoktm (approved):
Perfect

Issue #6963, opened 2023-10-02
Update to Flask 3.0.0


legoktm:
Flask 3.0.0 is out! This is likely going to trigger updates for a number of other dependencies too...some notes inline.

* [ ] Flask - https://flask.palletsprojects.com/en/3.0.x/changes/#version-3-0-0
* [ ] Flask-Babel 4.0.0
* [ ] Flask-SQLAlchemy - 3.1.1
   * [ ] SQLAlchemy - 2.x
* [ ] Flask-WTF - 1.2

legoktm:
It might make sense to jump to Flask 2.2.0 as an interim step, since that seems to introduce a lot of new functionality (JSON provider, some config attributes) where the old way is being removed in 3.0.

Flask-Babel also has a good number of changes that seem incompatible with how we use it, specifically that various lookups can now no longer be done during setup...we may need to ask upstream for some tweaks.

I've pushed a branch with how far I got by just quickly trying it: https://github.com/freedomofpress/securedrop/tree/flask-3

TkTech:
> Flask-Babel also has a good number of changes that seem incompatible with how we use it, specifically that various lookups can now no longer be done during setup...we may need to ask upstream for some tweaks.

I don't see anything from a quick skim of this project that would be fundamentally incompatible. Just switch to using Babel.init_app to pass your settings in your i18n.py and you should be good to go.

legoktm:
Hi @TkTech! Thanks for your work on Flask-Babel and for being proactive in finding this issue :) I already made a few changes along those lines (see https://github.com/freedomofpress/securedrop/commit/08d6704bc592d12c1d7306c2fd4f9e6dcf6d3796) but here's what I ran into using the `flask-3` branch of this repo:

```
$ make dev
...
Traceback (most recent call last):
  File "./manage.py", line 452, in <module>
    _run_from_commandline()
  File "./manage.py", line 442, in _run_from_commandline
    rc = args.func(args)
  File "./manage.py", line 99, in reset
    with context or app_context():
  File "/usr/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/home/user/github/freedomofpress/securedrop/securedrop/management/__init__.py", line 11, in app_context
    with journalist_app.create_app(config).app_context():
  File "/home/user/github/freedomofpress/securedrop/securedrop/journalist_app/__init__.py", line 93, in create_app
    i18n.configure(config, app)
[...]

PR #7264, opened 2024-10-18
Use sqlalchemy.LargeBinary instead of deprecated Binary

legoktm:
## Status

Ready for review

## Description of Changes

The Binary type was renamed to LargeBinary in ~2010, with the old name being an alias and deprecated. It will be removed in sqlalchemy 1.4, so let's migrate ahead of that.

See <https://github.com/sqlalchemy/sqlalchemy/issues/6263#issuecomment-819645247>.

Refs #6963.

## Testing

* [x] Visual review
* [x] CI passes

## Deployment

Any special considerations for deployment? n/a; should be a complete no-op

## Checklist


legoktm:
(The admin-tests CI jobs are failing because we're currently in a mid-release state, where a tag has been pushed but no release object exists)

zenmonkeykstop (approved):

PR #7271, opened 2024-10-23
Fix pagelayout tests under Firefox 128 and Tor Browser 14.0

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Following the release of Tor Browser 14.0 (based on FF 128.3.0esr) `pageslayout` tests involving setting Tor Browser preferences are failing. The problem is already fixed upstream (see https://github.com/webfp/tor-browser-selenium/blob/main/tbselenium/utils.py#L96) but a release with the fix isn't available yet. This PR:

- Updates the test Dockerfile to check for and use the correct Firefox version for latest stable Tor Browser
- monkeypatches `tbselenium` with the upstream fix. 

## Testing

- [ ] CI is passing - in particular pageslayouts tests.
- [ ] Visual review

### DIFF
--- a/securedrop/dockerfiles/focal/python3/Dockerfile
+++ b/securedrop/dockerfiles/focal/python3/Dockerfile
@@ -13,9 +13,9 @@ RUN apt-get update  && apt-get install -y \
     npm install --global html_codesniffer@2.5.1
 
 # Current versions of the test browser software. Tor Browser is based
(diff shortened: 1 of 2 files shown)


rocodes (approved):
- [x] CI passing
- [x] visual review- changes as seen [upstream](https://github.com/webfp/tor-browser-selenium/blob/main/tbselenium/utils.py#L96)

PR #7278, opened 2024-10-24
build: add `dev-get-id`, `dev-enter`, and `dev-load-data` targets

cfm:
## Status

Ready for review

## Description of Changes

Adds some helper Make targets working with a running `make dev[-tor]` container:
1. `dev-get-id` looks up the container ID
2. `dev-enter` starts a shell in the running container
3. `dev-load-data` runs `loaddata.py` in the running container

## Testing

With or without `export USE_PODMAN=1`, start `make dev` or `make dev-tor`, then try each of the following and confirm that they work as expected:

```sh-session
$ make dev-get-id
$ make dev-enter
$ make dev-load-data
```

## Deployment

Development-only.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,11 @@ STABLE_VER := $(shell cat molecule/shared/stable.ver)
 SDBIN := $(SDROOT)/securedrop/bin
 DEVSHELL := $(SDBIN)/dev-shell
 
+ifdef USE_PODMAN
+OCI_BIN=podman
+else
+OCI_BIN=docker
+endif
 
 ######################################
 #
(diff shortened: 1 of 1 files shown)


legoktm:
I have an alias of "sdshell" set to `podman exec --user=root -it (podman ps --filter name=securedrop --format '{{.ID}}') bash`. So +1 from me.

legoktm (changes requested):
Looks good functionally but let's add podman support too please :)

cfm:
Done in 1a2602cd73391f8a589b03a646f4ad3e3e5be1ee.

Makefile
```diff
@@ -8,6 +8,11 @@ STABLE_VER := $(shell cat molecule/shared/stable.ver)
 SDBIN := $(SDROOT)/securedrop/bin
 DEVSHELL := $(SDBIN)/dev-shell
 
+ifdef $(USE_PODMAN)
```
legoktm: This didn't work for me, per https://www.gnu.org/software/make/manual/html_node/Conditional-Syntax.html I think it needs to be the name of the variable, not the value, so:

```suggestion
ifdef USE_PODMAN
```

and that did work for me.

Makefile
```diff
@@ -249,6 +254,24 @@ dev-tor:  ## Run the development server with onion services in a Docker containe
...
 	@echo
 
+.PHONY:
+dev-get-id:  ## Get the ID of the running "make dev" or "make dev-tor" container.
+	@$(OCI_BIN) ps --format json --filter "name=securedrop-dev" --format '{{.ID}}'
+
+.PHONY:
+dev-enter:  ## Start a shell directly in the running "make dev" or "make dev-tor" container.
+	@$(OCI_BIN) exec -it \
+		$(shell make dev-get-id) \
```
legoktm: This internal `make` command is outputting `make[1]: Entering directory /home/user/github/freedomofpress/securedrop ae8314c6d60a make[1]: Leaving directory /home/user/github/freedomofpress/securedrop`

so then obviously podman fails with: `Error: no container with name or ID "make[1]:" found: no such container`.

Solution appears to be:

```suggestion
		$(shell make -s dev-get-id) \
```

Makefile
```diff
@@ -249,6 +254,24 @@ dev-tor:  ## Run the development server with onion services in a Docker containe
...
+		bash
+
+.PHONY: dev-load-data
+dev-load-data:  ## Run "loaddata.py" in the running "make dev" or "make dev-tor" container. Set $NUM_JOURNALISTS and/or $NUM_SOURCES on the command line as needed.
+	@$(OCI_BIN) exec -it \
+		-e NUM_JOURNALISTS \
+		-e NUM_SOURCES \
+		$(shell make dev-get-id) \
```
legoktm:
```suggestion
		$(shell make -s dev-get-id) \
```

Makefile
```diff
@@ -249,6 +254,24 @@ dev-tor:  ## Run the development server with onion services in a Docker containe
...
+		./loaddata.py
```
legoktm: can we add an $(ARGS) or something here? To pass e.g. `--random-file-size`.

legoktm (changes requested):

Makefile
cfm: Fixed in c15991b48a04b516b62ad61450173968a8402aaf.

Makefile
cfm: Fixed in c15991b48a04b516b62ad61450173968a8402aaf.

Makefile
cfm: Fixed in e0c1f3783984c579e55d4e7167a37e88d95890e5.

Makefile
cfm: Added in c15991b48a04b516b62ad61450173968a8402aaf.

cfm:
Whoops; sorry for the sloppiness there.  Should be all fixed up, @legoktm.

legoktm (approved):
LGTM, thanks!

Issue #7258, opened 2024-10-17
staging job is failing with dpkg lock contention


legoktm:
* https://github.com/freedomofpress/securedrop/actions/runs/11377625042/attempts/1

`failed: [mon-staging] (item=securedrop-keyring_0.2.2+2.10.0+focal_all.deb) => {"ansible_loop_var": "item", "changed": true, "cmd": ["dpkg", "-i", "/root/securedrop-keyring_0.2.2+2.10.0+focal_all.deb"], "delta": "0:00:00.007442", "end": "2024-10-17 03:28:49.153608", "item": "securedrop-keyring_0.2.2+2.10.0+focal_all.deb", "msg": "non-zero return code", "rc": 2, "start": "2024-10-17 03:28:49.146166", "stderr": "dpkg: error: dpkg frontend lock is locked by another process", "stderr_lines": ["dpkg: error: dpkg frontend lock is locked by another process"], "stdout": "", "stdout_lines": []}`

* https://github.com/freedomofpress/securedrop/actions/runs/11377625042/attempts/2

`failed: [mon-staging] (item=securedrop-ossec-server_3.6.0+2.10.0+focal_all.deb) => {"ansible_loop_var": "item", "changed": true, "cmd": ["dpkg", "-i", "/root/securedrop-ossec-server_3.6.0+2.10.0+focal_all.deb"], "delta": "0:00:00.008102", "end": "2024-10-17 15:21:57.993267", "item": "securedrop-ossec-server_3.6.0+2.10.0+focal_all.deb", "msg": "non-zero return code", "rc": 2, "start": "2024-10-17 15:21:57.985165", "stderr": "dpkg: error: dpkg frontend lock is locked by another process", "stderr_lines": ["dpkg: error: dpkg frontend lock is locked by another process"], "stdout": "", "stdout_lines": []}
[...]

PR #7299, opened 2024-10-29
Don't enable apt-daily services, just the timers

### LINKED ISSUE #7298: We shouldn't enable apt-daily service units
## Description

In the ansible playbook:
```yaml
- name: Ensure apt-daily and apt-daily-upgrade services are unmasked, started and enabled.
  systemd:
    name: "{{ item }}"
    state: started
    enabled: yes
    masked: no
  with_items:
    - 'apt-daily'
    - 'apt-daily-upgrade'
  tags:
    - apt
    - unattended-upgrades

- name: Ensure apt-daily and apt-daily-upgrade timers are started, and enabled.
  systemd:
    name: "{{ item }}"
    state: started
    enabled: yes
  with_items:
    - 'apt-daily.timer'
    - 'apt-daily-upgrade.timer'
  tags:
    - apt
    - unattended-upgrades
```

We should only be starting/enabling the *timers* not the *services* themselves.

I noticed this while trying to figure out why `./securedrop-admin install` kept reporting things as changing (the service needed to be started), even though I had just run it seconds before.

legoktm:
## Status

Ready for review

## Description of Changes

By enabling the services, it means it runs every time the machine boots, defeating the point of the timer.

Similarly, starting the service/timer means that it starts running while the playbook is still going, which might also explain the dpkg lock contention (#7258).

Ansible will now just ensure the units are unmasked and the securedrop-config postinst will disable the services if enabled.

Fixes #7298.

## Testing

* [ ] visual review
* [ ] staging CI passes

## Deployment

1. Upgrading existing production instances - handled by securedrop-config postinst
2. New installs - handled by ansible logic

## Checklist

### DIFF
--- a/install_files/ansible-base/roles/common/tasks/unattended_upgrades.yml
+++ b/install_files/ansible-base/roles/common/tasks/unattended_upgrades.yml
@@ -20,11 +20,11 @@
   # Ensure daemon-reload has happened before starting/enabling
 - meta: flush_handlers
 
-- name: Ensure apt-daily and apt-daily-upgrade services are unmasked, started and enabled.
+- name: Ensure apt-daily and apt-daily-upgrade services are unmasked
   systemd:
     name: "{{ item }}"
-    state: started
-    enabled: yes
+    # We disable the service unit and enable the timer
+    enabled: no
     masked: no
   with_items:
     - 'apt-daily'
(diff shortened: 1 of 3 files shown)


legoktm:
(This isn't on the critical path for noble, so I'll fix it up later.)

eloquence:
Ended up not being necessary and was developed for Noble migrations, closing.

PR #7301, opened 2024-10-29
Don't create separate apache files per distro version

legoktm:
## Status

Ready for review

## Description of Changes

These are already jinja2 templates, so we can add inline conditionals if needed; but let's not have to copy-paste them for each new version.

## Testing

* [ ] Visual review
* [ ] staging CI passes

## Deployment

## Checklist


zenmonkeykstop (approved):
Looks consistent. Approving.

PR #7305, opened 2024-10-29
Remove test_securedrop_application_apt_dependencies test

legoktm:
## Status

Ready for review 

## Description of Changes

We have much more confidence in our Debian packages now, so we don't need to keep a duplicated list of dependencies and check that they're installed.

This happened to be the last use of the `python_version` variable, so remove that entirely.

## Testing

* [ ] visual review
* [ ] staging CI passes

## Deployment

## Checklist

### DIFF
--- a/molecule/testinfra/app-code/test_securedrop_app_code.py
+++ b/molecule/testinfra/app-code/test_securedrop_app_code.py
@@ -3,7 +3,6 @@ import testutils
 
 securedrop_test_vars = testutils.securedrop_test_vars
 testinfra_hosts = [securedrop_test_vars.app_hostname]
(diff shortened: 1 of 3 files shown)


zenmonkeykstop:
Is there any harm in leaving this check in?

legoktm:
It needs to be updated for noble to remove the libpython3.8 check; when thinking of doing that I realized it wasn't a useful test anyways (it'll only fail if apt is seriously broken, at which point we're already hosed), just unnecessary duplication IMO.

zenmonkeykstop (approved):
LGTM based on visual review.

PR #7308, opened 2024-10-29
Remove tests checking that no apparmor profiles are complaining

legoktm:
## Status

Ready for review

## Description of Changes

Ubuntu noble now has some transmission-related profiles in complain mode. We don't really care about this, so just remove the test. Ideally distros will continue to add more apparmor profiles, even if they're complain-only to start.
## Testing

* [ ] visual review

## Deployment

### DIFF
--- a/molecule/testinfra/app/test_apparmor.py
+++ b/molecule/testinfra/app/test_apparmor.py
@@ -63,22 +63,6 @@ def test_apparmor_ensure_not_disabled(host, profile):
         assert not f.exists
 
 
-@pytest.mark.parametrize("complain_pkg", sdvars.apparmor_complain)
-def test_app_apparmor_complain(host, complain_pkg):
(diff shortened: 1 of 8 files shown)


zenmonkeykstop:
Is transmission actually being installed tho? In that case we should be removing said package, not disabling tests that get caught up on it.

legoktm:
No, it isn't. But Ubuntu is installing a bunch of profiles like transmission, discord, steam, etc. I didn't check which package though (can do tomorrow).

legoktm:
Here's all the profiles the base `apparmor` package installs: https://gist.github.com/legoktm/c15059f372eab85754829b6984237c67. Since it's in the base package, there's nothing we can really do and I don't think overriding/disabling them would be worth it.

zenmonkeykstop (approved):
LGTM based on visual review, discussion.

PR #7315, opened 2024-10-31
Remove ufw from new and existing installs

### LINKED ISSUE #7313: Remove ufw from focal instances
## Description

ufw is installed as part of the standard Ubuntu focal server installation. However, we don't use it and instead rely on iptables-persistent.

On focal this is fine, but in noble we have:
```
$ apt info ufw
Package: ufw
Version: 0.36.2-6
Priority: standard
Section: admin
Origin: Ubuntu
Maintainer: Jamie Strandboge <jdstrand@ubuntu.com>
Bugs: https://bugs.launchpad.net/ubuntu/+filebug
Installed-Size: 869 kB
Depends: iptables, ucf, python3:any, debconf (>= 0.5) | debconf-2.0
Suggests: rsyslog
Breaks: iptables-persistent, netfilter-persistent
```

Since we're not using ufw and want to keep iptables-persistent, let's just uninstall it pre-upgrade.

This cannot be done during a postinst since you can't modify packages then, so I think we can install a systemd unit that just runs `apt-get remove --yes ufw` with `ConditionPathExists=/usr/sbin/ufw`. For fresh installs ansible can also ensure the package is absent.

legoktm:
## Status

Ready for review

## Description of Changes

We don't use ufw and in noble, it conflicts with iptables-persistent, which we do want to use.

Remove it during provisioning and install a systemd timer to remove it. (We can't do it during a postinst because we're already in an apt session at that time.)

Fixes #7313.

## Testing

* [x] Visual review
* [x] staging CI passes (tests fresh install case)
* [ ] build debs and install on a focal system, wait 1 day and see that ufw was removed

## Deployment

Any special considerations for deployment? Both new installs and upgrades are handled

## Checklist

### DIFF
--- a/install_files/ansible-base/roles/prepare-servers/tasks/main.yml
+++ b/install_files/ansible-base/roles/prepare-servers/tasks/main.yml
@@ -23,9 +23,11 @@
       custom kernel that is not signed. Please disable SecureBoot on the
       target servers and try again.
 
-- name: Remove cloud-init
+- name: Remove cloud-init and ufw
   apt:
-    name: cloud-init
+    name:
+      - cloud-init
+      - ufw
     state: absent
     purge: yes
   tags:
--- a/molecule/testinfra/common/test_system_hardening.py
+++ b/molecule/testinfra/common/test_system_hardening.py
@@ -1,4 +1,5 @@
 import re
+import time
 
 import pytest
 import testutils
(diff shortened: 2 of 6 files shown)


legoktm:
Note that I haven't tested the upgrade path yet, I was planning to set it on Friday afternoon and check on it on Monday.

legoktm:
I think I should rework this to be more flexible given https://github.com/freedomofpress/securedrop/issues/7326.

legoktm:
I renamed the unit to `securedrop-remove-packages` but kept everything else the same. To remove haveged we can do:

```ini
[Unit]
Description=Remove ufw if installed
ConditionPathExists|=/usr/sbin/ufw
ConditionPathExists|=/usr/sbin/haveged

[Service]
Type=oneshot
Environment="DEBIAN_FRONTEND=noninteractive"
ExecStart=/usr/bin/apt-get purge --yes ufw haveged
User=root
```

But, I don't want to do that in this PR because I haven't had the chance to look into haveged yet and let's keep the scope down.

zenmonkeykstop (approved):
Approving based on visual review.

PR #7317, opened 2024-10-31
Don't reuse the "ssh" group for our own access control

### LINKED ISSUE #7316: Create our own ssh group for access control
## Description

We're currently using the `ssh` group created by the openssh-client package, which in noble gets renamed to `_ssh` to more clearly indicate that it's for internal usage only.

Because the group is listed in our sshd_config and firewall rules, the rename breaks SSH access.

We should create our own group, migrate all users to it and then update config to reflect the change. This should be done in focal, before the upgrade.

legoktm:
## Status

Ready for review

## Description of Changes

The "ssh" group was supposed to be an internal group that was used by openssh itself for privilege separation. It's since been renamed to "_ssh"[1] to clarify that it's internal-only and flagged that we shouldn't be using it either.

Instead, create a new "sdssh" group, add our users to it, and use it in the sshd_config.

For upgrades, we migrate users to the new group and then update sshd_config and iptables rules.

[1] https://salsa.debian.org/ssh-team/openssh/-/blob/eac38305fc5d8eb8301a106294cf6c79447bdeb3/debian/openssh-client.postinst

Fixes #7316.

## Testing

* [x] staging CI passes
* [x] install focal (either 2.10.1 or develop without this change), copy the securedrop-migrate-ssh-group script and run it on either app or mon. reboot the instance, then verify you can ssh back in and `id sdadmin` includes "sdssh" and not plain "ssh"

## Deployment

Any special considerations for deployment? Both new installs and upgrades are considered

## Checklist

### DIFF
--- a/install_files/ansible-base/roles/common/tasks/create_users.yml
+++ b/install_files/ansible-base/roles/common/tasks/create_users.yml
@@ -13,11 +13,18 @@
     - users
     - sudoers
 
+- name: Create "sdssh" group (for limiting SSH access)
+  group:
+    name: sdssh
+    state: present
+  tags:
+    - users
+
 - name: Create shell accounts for SecureDrop admins.
   user:
     name: "{{ item }}"
     shell: /bin/bash
-    groups: sudo,ssh
+    groups: sudo,sdssh
   with_items: "{{ ssh_users }}"
   tags:
     - users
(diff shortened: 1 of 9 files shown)


legoktm:
(Still work in progress, I need to actually test the upgrade path)

legoktm:
OK, I've finished testing this and it (now) works, I learned that `usermod -r` isn't supported on focal so we have to use gpasswd instead. Should be ready to go!

zenmonkeykstop (approved):
LGTM, migration script works as expected. This will require upgrade testing during its release's QA tho.

PR #7319, opened 2024-10-31
Have app-code depend on apparmor in noble

legoktm:
## Status

Ready for review

## Description of Changes

During a system upgrade, the securedrop-app-code postinst tries to enable the apparmor profiles for apache2 and tor, except it uses newer apparmor syntax, which can fail if apparmor hasn't already been upgraded.

The easiest and correct way to ensure that securedrop-app-code is upgraded after apparmor is by setting a versioned dependency. We need to do a bit of makefile magic so it only applies to noble builds though, and a test verifies that.

## Testing

* [x] CI passes, esp the build-debs jobs
* [ ] version of apparmor we set the dependency on matches what noble has (see https://launchpad.net/ubuntu/+source/apparmor)

## Deployment

## Checklist


zenmonkeykstop (approved):
LGTM based on visual review.

PR #7320, opened 2024-11-01
Don't return in test_swap_disabled, use skip instead

legoktm:
## Status

Ready for review

## Description of Changes

Returning triggers a warning:
```
 PytestReturnNotNoneWarning: Expected None, but
 testinfra/common/test_system_hardening.py::test_swap_disabled[ansible://mon-staging]
 returned True, which will be an error in a future version of pytest.
 Did you mean to use `assert` instead of `return`?
```

## Testing

* [ ] staging CI passes

## Deployment

## Checklist


zenmonkeykstop (approved):
LGTM

Issue #7323, opened 2024-11-01
Fix sysctl handling for noble/Linux 6.6


legoktm:
## Description

With noble/Linux 6.6, the grsec_lock flag now strictly enforces that it cannot be set again, even if the value is the same as the current. This means running `sysctl -p /etc/sysctl.conf` after it is set will error out.

We also set each sysctl flag twice, once for each one explicitly, and then again with `sysctl -p /etc/sysctl.conf`. I think we should switch to just doing `sysctl -p` once, globally.

* grsec_sysctl_flags:
   * kernel.grsecurity.rwxmap_logging
   * kernel.grsecurity.grsec_lock
   * vm.heap_stack_gap <-- now fails because it's after grsec_lock; note that this is already set via the securedrop-grsec package
* sysctl_flags, which is all `net.ipv4.*` stuff. setting these also fail because it's after the grsec_lock. I think we can just move them all to the kernel metapackage, and ensure it gets released before the next SD server release.

So proposed work plan:
1. [x] Move net.ipv4 flags to securedrop-grsec metapackage; postinst will remove vm.heap_stack_gap and net.ipv4 flags from sysctl.conf: https://github.com/freedomofpress/kernel-builder/pull/55
2. [x] ansible change to stop setting vm.heap_stack_gap and net.ipv4 flags for new installs: https://github.com/freedomofpress/securedrop/pull/7324
3. [x] Switch to sysctl -p once, globally (in ansible provisioning) https://github.com/freedomofpress/securedrop/pull/7353
4. [x] Ship kernel upgrade for step 1: https://github.com/freedomofpress/securedrop-apt-prod/pull/154
5. [x] Release SD server with step 2

legoktm:
The kernel update is now live, I think we can close this tomorrow once it's successfully rolled out. (No point keeping this open until 2.11.0 is out, which is the final step).

legoktm:
> I think we can close this tomorrow

Only like 2 months late, but this is fully done and released now.

PR #7324, opened 2024-11-01
Drop setting vm.heap_stack_gap and net.ipv4 sysctl flags

legoktm:
## Status

Ready for review

## Description of Changes

These are now set via the securedrop-grsec metapackage (see <https://github.com/freedomofpress/kernel-builder/pull/55>).

Refs #7323.

## Testing

* [ ] visual review; compare that same flags are moved via https://github.com/freedomofpress/kernel-builder/pull/55

## Deployment

This can only be deployed after a kernel with https://github.com/freedomofpress/kernel-builder/pull/55 is released stable. It's safe to merge ahead of time though once the kernel-builder one is approved.

## Checklist


molecule/testinfra/common/test_system_hardening.py
```diff
@@ -7,34 +7,6 @@
 testinfra_hosts = [sdvars.app_hostname, sdvars.monitor_hostname]
 
 
-@pytest.mark.parametrize(
```
zenmonkeykstop: IMO keeping the testinfra checks makes sense, in case the metapackage-set values get borked somehow.

molecule/testinfra/common/test_grsecurity.py
```diff
@@ -81,7 +81,6 @@ def test_grsecurity_kernel_is_running(host):
     [
         ("kernel.grsecurity.grsec_lock", 1),
         ("kernel.grsecurity.rwxmap_logging", 0),
-        ("vm.heap_stack_gap", 1048576),
```
zenmonkeykstop, line 84: (repeat) IMO keeping the testinfra checks makes sense, in case the metapackage-set values get borked somehow.

zenmonkeykstop (dismissed):
kernel-builder change looks good, but unless there's a compelling reason I'd keep the testinfra check.

molecule/testinfra/common/test_system_hardening.py
legoktm: OK, happy to leave it in for now but long-term I think we don't want to run into needing to constantly keep this repo updated if we change things in kernel-builder (but I doubt we'll be changing that many sysctl flags).

legoktm:
This should fail CI now because we haven't built new kernel packages yet, I will see about doing some today or tomorrow.

legoktm:
The error is: `sed -i '/^vm\\.heap_stack_gap/d' /etc/sysctld.conf", "sed: can't read /etc/sysctld.conf: No such file or directory`

which I guess is an install order issue in that we're installing the new kernel before we try setting grsec kernel flags, so there's no sysctld.conf yet. Let me guard it with an if exists check in the metapackage.

legoktm:
kernel-builder fix: https://github.com/freedomofpress/kernel-builder/pull/57

cfm:
I've merged freedomofpress/kernel-builder#57, @legoktm, and I'll be happy to re-review this once it passes with kernels rebuilt with it.  (Let me know if you'd like me to build them tomorrow.)

cfm (approved):
Looks good after 37f182724bfc685b5bb1c43fd35c405290c3ff66 and freedomofpress/securedrop-apt-test#258.  Thanks, @legoktm.

PR #7331, opened 2024-11-07
Install backup script on app server via Debian package

### LINKED ISSUE #7329: Allow running backups from the app server itself
## Description

Currently the backup script is shipped on the admin workstation (via the git repo), then when invoked, rsynced up to the app server and executed. This means it has to be invoked solely from the admin workstation.

But from what I can tell, there's no reason for that. We can just ship the backup script via the app-code package and have the workstation playbook execute that.

This will allow us to create backups directly on the app server.

legoktm:
## Status

Ready for review

#### Install backup script on app server via Debian package

This will allow the script to be invoked from the server itself.
    
Fixes #7329.

#### Give better control over where backups are stored
    
Instead of saving them in the current directory and requiring awkward workarounds to change that directory, take it as an argument so the script can be run from anywhere without requiring context changes.

#### Add testinfra check that the backup script works
    
Try creating a backup and verify it contains a few of the files we expect it to.

## Testing

* [x] staging CI passes
* [ ] create a backup from the admin workstation, it should work just as before.

## Deployment

Consider the following scenarios:

* upgraded admin workstation + upgraded app server -> works fine
* old admin workstation + upgraded app server -> works fine (continues to rsync local backup.py)
* upgraded admin workstation + old app server -> breaks, file won't exist on app server to invoke. This is OK in my opinion because your app server should be getting updates, and if not, you can downgrade your workstation.

## Checklist

### DIFF
--- a/install_files/ansible-base/roles/backup/tasks/backup.yml
+++ b/install_files/ansible-base/roles/backup/tasks/backup.yml
@@ -1,8 +1,6 @@
 ---
 - name: Run the backup script (can take a while).
-  # Setting args inline (against project style conventions) to work around
-  # Ansible bug: https://github.com/ansible/ansible/issues/9693
-  command: /usr/bin/securedrop-app-backup.py chdir=/tmp
+  command: /usr/bin/securedrop-app-backup.py --dest /tmp
   # If there are a lot of submissions, this task could take a while.
   async: 3600
   poll: 10
--- a/install_files/ansible-base/roles/backup/tasks/backup.yml
+++ b/install_files/ansible-base/roles/backup/tasks/backup.yml
@@ -1,15 +1,8 @@
 ---
-- name: Copy the backup script to the target server.
-  copy:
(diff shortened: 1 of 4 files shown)


cfm (approved):
Nice, @legoktm.  I've tested this fully, taking your list of deployment scenarios as a test plan, because I regard it as a somewhat significant change to move this script from being managed by Git+Ansible, where it's invoked, to being managed by `securedrop-app-code`.  That said, I agree that it's the _right_ change, insofar as the rest of `securedrop-app-code` is what both the backup format and the backup/restore operations ought to be versioned _with_.

* [x] upgraded admin workstation + upgraded app server -> works fine
* [x] old admin workstation + upgraded app server -> works fine (continues to rsync local backup.py)
* [x] upgraded admin workstation + old app server -> breaks, file won't exist on app server to invoke. This is OK in my opinion because your app server should be getting updates, and if not, you can downgrade your workstation.

Issue #7322, opened 2024-11-01
Check for noble readiness and tell journalists/admins


legoktm:
## Description

_Update: This is an old proposal, we're not immediately planning to do the API check, instead we'll focus on a JI banner and other things._

The SI's public `/metadata` endpoint outputs some information about a SecureDrop install that is used by us for various ecosystem observation purposes (primarily checking that auto-updates have succeeded).

With the upcoming noble migration, we are looking at doing it automatically, but there will still be some issues that cannot be fixed ahead of time (e.g. enough free disk space).

My idea is that we add a flag to the API output like `noble_readiness: true/false`. It only outputs a boolean flag, which we can use to find instances that are not ready, and reach out to their admins, who can log in and run a script that outputs more details on what's wrong.

Example checks that would trigger a false response:
* enough free diskspace on the root partition to download new packages
* migrated users out of the ssh group (see #7317)
* removed ufw firewall (#7313)
* check no other unexpected apt sources

The main argument against this is that we are exposing information about an instances internal health, we mitigate that by just outputting a boolean, and checking enough things that a potential attacker cannot figure out which thing is triggering the failure.

Also because this is running from the app server, I don't think we can handle the mon server.

zenmonkeykstop:
Overall I don't think this poses much of a security risk. The set of changes an attacker might be able to infer is smaller than it seems tho. 2/4 changes (ufw,ssh) above could be inferred by looking at version number, as they'd be enforced by a release's postint.  Other ones might be useful to an attacker looking to, say, use up disk space - but they could just put in lots of submissions. 

I'm kindof of two minds about this approach tho, just because it is a public API, this is extra non-public info (albeit collapsed into a boolean). One alternative - which is in line with what was done on previous updates - would be to consider a banner message in the JI. This can be a lot more verbose - and while we won't be able to see it, if it's sufficiently "alerty" it will enlist the help of users to get admins to contact us rather than the other way round.

The other advantage of an alert within the JI is that for folks running an instance without contact with FPF, the API change is unlikely to be noticed, while the JI banner is going be visible to anyone using the system. (Obviously, they are not mutually exclusive.)

legoktm:
I'll start working on the banner message. I'm also thinking about an OSSEC alert since that can check the mon server as well.

zenmonkeykstop:
> I'll start working on the banner message. I'm also thinking about an OSSEC alert since that can check the mon server as well.

[...]

Issue #7332, opened 2024-11-08
Create focal -> noble upgrade script


legoktm:
## Description

Part of #7211.

The workflow of the script will be:

* apply any pending updates and reboot
* run migration check to make sure we're all set to go
* stop apache2
* take a backup
* write `/etc/securedrop-upgraded-from-focal` marker file
* check disk space, again??
* raise OSSEC email_alert_level config to 15
* disable unattended upgrades
* edit sources.list to point to noble
* `apt-get update`
* `apt-get upgrade --without-new-pkgs`
* `apt-get full-upgrade`
* re-enable unattended-upgrades
* re-enable OSSEC(?)
* reboot
* some basic verification
* start apache2
* delete backup

Everything the script does should be aggressively logged and each major step should trigger an OSSEC notification.

I think we can trigger the script from a systemd timer/service. Each step should be recorded in some kind of state file so it can recover and resume if interrupted.

We should also leave behind a marker file like `/etc/securedrop-upgraded-from-focal` so if we detect some bug in the future, we can conditionally apply logic based on whether it's a fresh noble install or an upgrade.

legoktm:
One question I've been debating in my head is whether to write the script in Rust or Python. The pros for Rust are pretty clear, this is a place we want to have solid error handling and recovery. The cons are that I think it'll be a little harder to get review for and be a bit more friction for others to contribute.

My plan is to start sketching out the script in Python and then see if it can do all the error handling stuff in a not-crazy way. I expect any Python<-->Rust porting to be trivial.

legoktm:
> raise OSSEC email_alert_level config to 15

This is doable on the mon server but going to be an issue on the app server.

zenmonkeykstop:
Doesn't OSSEC have a message threshold anyway? Given this is a one-time event I think you could probably do without the alert level change.

legoktm:
Yes, after it hits some limit it stops sending messages and just queues them up...which means there's a backlog of like 1000 emails and any new alerts (e.g. using the send OSSEC test button) don't get sent. When I upgraded the mon server I got ~27 emails before it stopped sending new ones. The queue seems to be stored in memory though, because I restarted the ossec-server service and then it dropped everything and started sending my newly triggered OSSEC test alerts.

I'll put it into the nice to have bucket for now and when we do test upgrades we can see how bad the impact is and decide if we want something else.

legoktm:
[...]

PR #7334, opened 2024-11-08
Add a basic noble migration check script

legoktm:
## Status

Ready for review 

## Description of Changes

Perform a number of checks to ensure the system is ready for the noble migration. The results are written to a JSON file in /etc/ that other things like the JI and the upgrade script itself can read from.

The script is run hourly on a systemd timer but can also be run interactively for administrators who want slightly more details.

For ease of review, this is split into two commits: first set up all the configuration needed to ship Rust code in the securedrop-config package; and then second the actual check script, Rust dependencies, and systemd units.

Out of scope for this PR: OSSEC alerts, JI banner.

Refs #7322.

## Testing

* [ ] staging CI passes
* [ ] visual review

## Deployment

## Checklist

### DIFF
--- a/noble-migration/src/bin/check.rs
+++ b/noble-migration/src/bin/check.rs
@@ -1,0 +1,359 @@
+//! Check migration of a SecureDrop server from focal to noble
+//!
+//! This script is run as root on both the app and mon servers.
+//!
+//! It is typically run by a systemd service/timer, but we also
+//! support admins running it manually to get more detailed output.
+use anyhow::{bail, Context, Result};
+use rustix::process::geteuid;
+use serde::Serialize;
+use std::{
+    fs,
+    path::Path,
+    process::{self, ExitCode},
+};
+use url::{Host, Url};
+use walkdir::WalkDir;
+
+/// This file contains the state of the pre-migration checks.
+///
+/// There are four possible states:
+/// * does not exist: check script hasn't run yet
+/// * empty JSON object: script determines it isn't on focal
+/// * {"error": true}: script encountered an error
+/// * JSON object with boolean values for each check (see `State` struct)
+const STATE_PATH: &str = "/etc/securedrop-noble-migration.json";
+
+#[derive(Serialize)]
+struct State {
+    ssh: bool,
+    ufw: bool,
+    free_space: bool,
+    apt: bool,
+    systemd: bool,
+}
+
+impl State {
+    fn is_ready(&self) -> bool {
+        self.ssh && self.ufw && self.free_space && self.apt && self.systemd
+    }
+}
+
+/// Parse the OS codename from /etc/os-release
+fn os_codename() -> Result<String> {
+    let contents = fs::read_to_string("/etc/os-release")
+        .context("reading /etc/os-release failed")?;
+    for line in contents.lines() {
+        if line.starts_with("VERSION_CODENAME=") {
+            // unwrap: Safe because we know the line contains "="
+            let (_, codename) = line.split_once("=").unwrap();
+            return Ok(codename.trim().to_string());
+        }
+    }
+
+    bail!("Could not find VERSION_CODENAME in /etc/os-release")
+}
+
+/// Check that the UNIX "ssh" group has no members
+///
+/// See <https://github.com/freedomofpress/securedrop/issues/7316>.
+fn check_ssh_group() -> Result<bool> {
+    // There are no clean bindings to getgrpname in rustix,
+    // so jut shell out to getent to get group members
+    let output = process::Command::new("getent")
+        .arg("group")
+        .arg("ssh")
+        .output()
+        .context("spawning getent failed")?;
+    if output.status.code() == Some(2) {
+        println!("ssh: group does not exist");
+        return Ok(true);
+    } else if !output.status.success() {
+        bail!(
+            "running getent failed: {}",
+            String::from_utf8_lossy(&output.stderr)
+        );
+    }
+
+    let stdout = String::from_utf8(output.stdout)
+        .context("getent stdout is not utf-8")?;
+    let members = parse_getent_output(&stdout)?;
+    if members.is_empty() {
+        println!("ssh: group is empty");
+        Ok(true)
+    } else {
+        println!("ssh: group is not empty: {members:?}");
+        Ok(false)
+    }
+}
+
+/// Parse the output of `getent group ssh`, return true if empty
+fn parse_getent_output(stdout: &str) -> Result<Vec<&str>> {
+    let stdout = stdout.trim();
+    // The format looks like `ssh:x:123:member1,member2`
+    if !stdout.contains(":") {
+        bail!("unexpected output from getent: '{stdout}'");
+    }
+
+    // unwrap: safe, we know the line contains ":"
+    let (_, members) = stdout.rsplit_once(':').unwrap();
+    if members.is_empty() {
+        Ok(vec![])
+    } else {
+        Ok(members.split(',').collect())
+    }
+}
+
+/// Check that ufw was removed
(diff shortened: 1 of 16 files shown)


legoktm:
I'm still thinking of a few more checks to add, but I think this is ready for review + merge and other checks can be added incrementally.

securedrop/debian/config/usr/bin/securedrop-noble-migration-check.py
```diff
@@ -0,0 +1,137 @@
...
+        return
+
+    state = {}
+    state["ssh"] = check_ssh_group()
+    state["ufw"] = check_ufw_removed()
+    state["free_space"] = check_free_space()
+    state["apt"] = check_apt()
+    state["systemd"] = check_systemd()
+
+    state_path.write_text(json.dumps(state))
```
legoktm: We need some crash resistance here; if something fails before we dump to disk, we try our hardest to not leave the old state behind.

legoktm:
I plan to port this to Rust based on https://github.com/freedomofpress/securedrop/issues/7332#issuecomment-2474915210.

legoktm:
Now that this is in Rust, need to do a few more cargo vet reviews:

```
recommended audits for safe-to-deploy:
    Command                                     Publisher      Used By                            Audit Size
    cargo vet diff zerofrom-derive 0.1.3 0.1.4  robertbastian  zerofrom                           5 files changed, 49 insertions(+), 43 deletions(-)
    cargo vet diff idna 1.0.2 1.0.3             valenting      url and sequoia-openpgp            7 files changed, 207 insertions(+), 208 deletions(-)
    cargo vet diff url 2.5.1 2.5.3              valenting      noble-migration                    12 files changed, 386 insertions(+), 92 deletions(-)
    cargo vet inspect stable_deref_trait 1.2.0  Storyyeller    yoke and icu_provider              484 lines
    cargo vet inspect idna_adapter 1.2.0        hsivonen       idna                               612 lines
    cargo vet inspect smallvec 1.8.0            mbrubeck       idna, icu_normalizer, and 1 other  3879 lines
      NOTE: mozilla trusts Matt Brubeck (mbrubeck) - consider cargo vet trust smallvec mbrubeck

estimated audit backlog: 5960 lines
```

The main new dependency triggering reviews is `url`, which pulls in `idna`.

legoktm:
In the end we only needed one audit for zerofrom-derive 0.1.3 -> 0.1.4 (done); I hadn't configured the new crate to use the safe-to-run standard. Went a bit out of the way to upgrade cargo-vet since a newer version was available.

legoktm:
Marking this as ready for review, I went through and added a bunch of docs and tried to make the error handling a bit more robust and obvious. Staging CI will fail until the new kernels are on apt-test.

noble-migration/src/bin/check.rs
```diff
@@ -0,0 +1,259 @@
...
+    } else {
+        println!("ufw: ufw was removed");
+        true
+    }
+}
+
+const DESIRED_FREE_SPACE: u64 = 10 * 1024 * 1024 * 1024; // 10GB
+
+/// Check that there is enough free space for taking a backup
+/// and downloading all the new packages. 10GB is a bit of an
```
legoktm: I mostly made this number up. IIRC it was like 2-3GB of packages that needed to be downloaded, let's conservatively round it up to 4GB. That leaves 6GB for a backup.

This totally fails in two obvious cases:
1. Instance that has like 500MB of submissions stored and has 8GB free. They have enough space for the upgrade plus backup with enough headroom, but we'll still flag it.
2. Instance with 20GB of submissions stored and 12GB free. They don't have enough space for the backup but we don't flag it.

My current feeling is that we can make this check smarter (and also the backup script) with not too much effort, but let's do it as a follow-up so we can get this infrastructure in place.

legoktm:
On staging we get:
```
{'ssh': False, 'ufw': True, 'free_space': False, 'apt': True, 'systemd': True}
```

Free space is understandable, I'll have it ignore that one. I'm confused by ssh, that should've been resolved. Will add some debug code 🤔

legoktm:
I had a bug in my getent output parser when I ported the logic from Python to Rust. I have paid penance by adding a unit/regression test.

noble-migration/src/bin/check.rs
```diff
@@ -0,0 +1,313 @@
...
+    Ok(free_space)
+}
+
+const EXPECTED_DOMAINS: [&str; 5] = [
+    // Ubuntu domains
+    "archive.ubuntu.com",
+    "security.ubuntu.com",
+    // SecureDrop domains (including testing ones)
+    "apt.freedom.press",
+    "apt-qa.freedom.press",
```
zenmonkeykstop: IMO a test domain should error out or at least emit a warning. Prod instances should not be using them under any circumstances (especially not `apt-test` given signing key policies).

noble-migration/src/bin/check.rs
zenmonkeykstop: Any reason not to do it now and just require `4GB + get_size(/var/lib/securedrop/store) + 0.1*volume_size()`? (Allowing for the disk to get up to 90% during upgrades, but no further.)

noble-migration/src/bin/check.rs
```diff
@@ -0,0 +1,313 @@
...
+    fs::write(
+        STATE_PATH,
+        serde_json::to_string(&state).context("serializing state failed")?,
+    )
+    .context("writing state file failed")?;
+    if state.is_ready() {
+        println!("All ready for migration!");
+    } else {
+        println!(
+            "Some errors were found that will block migration.
```
zenmonkeykstop, line 296: One thing to consider would be adding a "getting ready for Noble" page in the docs, that lists steps admins should take, and linking to that here as well/instead of the support link. Giving folks the tools to solve problems themselves drives down our support overhead and also helps admins who don't want to work with FPF for whatever reason.

(Just added https://github.com/freedomofpress/securedrop-docs/issues/610)

zenmonkeykstop (commented):
added some comments. Haven't tested yet tho.

noble-migration/src/bin/check.rs
legoktm: We need to support test domains somehow for QA testing, but in replying to you I realized that we don't have the logic for them elsewhere (when switching them from focal to noble), so I'm going to back it out for now and we can do some kind of environment variable/flag to explicitly opt in to qa/test stuff for ourselves.

noble-migration/src/bin/check.rs
legoktm: > Any reason not to do it now

My initial intent was that this PR was just going to be the framework and then we could add more checks iteratively, but we're hitting end-of-sprint and also I haven't really come up with any other checks, so yeah, let me just do it now.

noble-migration/src/bin/check.rs
legoktm: Absolutely.

legoktm: I will spend some time working on the documentation and agree that we should link it here, but don't think it should hold up review/approval of this since it'll be trivial to add in afterwards but before the release.

legoktm:
Hm, and CI is failing now that I've removed the apt-test/apt-qa URLs... 🤔

noble-migration/src/bin/check.rs
legoktm: I've re-added the test domains (so CI will pass in staging) but it now prints a warning.

legoktm:
They just yanked the `url` crate, so rust-audit is failing. Let me upgrade it...

zenmonkeykstop (approved):
LGTM based on visual review and testing in staging.

PR #7335, opened 2024-11-12
Remove haveged if installed

### LINKED ISSUE #7326: haveged is still installed on long-running systems
## Description

Despite 04d62440e82340052ca29b7f031a11619551dfc9, and a test verifying that the package is missing, it is still installed on long-running systems. We should do something similar to #7315 for haveged as well.

legoktm:
## Status

Ready for review

## Description of Changes

Previously we only disabled/masked haveged instead of removing the package. Now that we have the infrastructure to remove packages (created for ufw), let's remove it properly.

Note that this will remove it from both app and mon, even though it was only ever disabled on app. That should be fine since it's not expected to be running on mon.

Fixes #7326.

## Testing

* [ ] staging CI passes
* [ ] visual review

## Deployment

Any special considerations for deployment? only applies to very old installs

## Checklist

### DIFF
--- a/securedrop/debian/config/lib/systemd/system/securedrop-remove-packages.service
+++ b/securedrop/debian/config/lib/systemd/system/securedrop-remove-packages.service
@@ -1,9 +1,10 @@
 [Unit]
-Description=Remove ufw if installed
-ConditionPathExists=/usr/sbin/ufw
+Description=Remove ufw/haveged if installed
+ConditionPathExists|=/usr/sbin/ufw
+ConditionPathExists|=/usr/sbin/haveged
 
 [Service]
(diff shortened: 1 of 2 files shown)


rocodes:
Thanks, this is great. Miniscule nit, the systemd [timer description](https://github.com/freedomofpress/securedrop/blob/4c9062b078d11c99c85923f93872cffc37548f27/securedrop/debian/config/lib/systemd/system/securedrop-remove-packages.timer#L2) should be updated since it's now for not just ufw, but I'll put in a followup PR.

rocodes (approved):
- [x] visual review
- [x] testinfra passing (CI)

PR #7337, opened 2024-11-12
Replace reboot-flag cron job with a systemd timer

### LINKED ISSUE #7013: Replace reboot-required cron with systemd timer
We have one cron job left (as root, so no OSSEC issues):
```
# The purpose of this cron is to drop the reboot-required flag every 12 hours
# to ensure the system is rebooted nightly, regardless of updates being installed
# or not.
0 */12 * * * root touch /var/run/reboot-required
```

Would be good to swap this out for a systemd timer.

legoktm:
## Status

Ready for review

## Description of Changes

This is our very last cron job, so let's get rid of it. On the plus side this makes it easier to disable when needed (e.g. during a system upgrade).

The testinfra check now actually verifies that the file is created when the service is run.

Fixes #7013.

## Testing

* [x] visual review
* [x] staging CI passes

## Deployment

Any special considerations for deployment? handles both

## Checklist

### DIFF
--- a/molecule/testinfra/common/test_automatic_updates.py
+++ b/molecule/testinfra/common/test_automatic_updates.py
@@ -1,4 +1,5 @@
 import re
+import time
 
 import pytest
 import testutils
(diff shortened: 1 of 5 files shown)


cfm (approved):
Very nice.

PR #7342, opened 2024-11-20
Absent sources.list on noble

legoktm:
## Status

Ready for review

## Description of Changes

On a fresh install, this is an empty file with a comment pointing to /etc/apt/sources.list.d/ubuntu.sources and isn't needed.

In the upgrade case it doesn't make sense for us to create the empty pointer file, so let's just ensure it's absent entirely.

## Testing

* [x] visual review
* [x] staging CI passes (no-op on focal)
* [ ] in theory you could do a fresh install of noble ubuntu server to verify the sources.list file is useless but I've already done that a bunch

## Deployment

Any special considerations for deployment? this handles the fresh install case, upgrades will be handled by the script

## Checklist

### DIFF
--- a/install_files/ansible-base/roles/common/tasks/apt_sources.yml
+++ b/install_files/ansible-base/roles/common/tasks/apt_sources.yml
@@ -20,6 +20,18 @@
   tags:
     - apt
 
+# Remove the obsolete sources.list on noble. On a fresh install,
+# this is an empty file with a comment pointing to ubuntu.sources.
+# On upgrade from focal, we're deleting this file.
+- name: Remove obsolete sources.list (noble)
+  template:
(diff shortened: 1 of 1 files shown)


legoktm:
staging is failing because the grsec package on apt-test is borked, will get it rebuilt tonight.

cfm (approved):
Thanks, @legoktm!

* [x] visual review
* [x] staging CI passes (no-op on focal)
* [ ] in theory you could do a fresh install of noble ubuntu server to verify the sources.list file is useless but I've already done that a bunch

I'm not going to do that while Noble support is still entirely WIP.  :-)  (And the logic here to exclude Focal looks sound.)

PR #7348, opened 2024-11-22
Display a banner in the JI regarding the noble migration

legoktm:
## Status

Ready for review

#### Display a banner in the JI regarding the noble migration

This is largely copied from the same functionality that was implemented during the focal migration (ecfecea).

There are two banners that can be seen:
    
OS_PAST_EOL is in effect after April 2, 2025 if the system is still
running on focal. The Source Interface automatically disables itself and
the Journalist Interface will display a banner informing journalists to
contact their administrator.

OS_NEEDS_MIGRATION_FIXES will display a notice in the Journalist
Interface if the check script has run and found issues that need
resolution. It doesn't affect the Source Interface.

The banners point at <https://securedrop.org/focal-eol>, which will be
set up as a redirect to the relevant documentation.

Both checks are done during startup, which means if the state changes
(e.g. disk space is freed up or a systemd unit fails), the banner state
will only change after the nightly reboot.

#### Disable "protocol" check from html_lint.py

This actively dissuades from using HTTPS URLs, favoring
protocol-relative ones. Even ignoring HTTPS-only URLs as a best
practice, given most onion services are hosted as HTTP sites,
they'd become HTTP links instead of HTTPS.

So let's just suppress this rule and link to the correct protocol.

Refs #7322

## Testing

* [ ] visual review
* [ ] CI passes 
* [ ] if you create `/etc/securedrop-noble-migration.json` in the dev container with a false value, the migration banner will be triggered
   * [ ] the way this file is interpreted is consistent with https://github.com/freedomofpress/securedrop/pull/7334
* [ ] if you change the `FOCAL_ENDOFLIFE` date in server_os.py  to 2024 or some other past date, the EOL banner will be triggered.

## Deployment

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ fix: ruff ## Apply automatic fixes.
 .PHONY: html-lint
 html-lint:  ## Validate HTML in web application template files.
 	@echo "███ Linting application templates..."
-	@html_lint.py --printfilename --disable=optional_tag,extra_whitespace,indentation,names,quotation \
+	@html_lint.py --printfilename --disable=optional_tag,extra_whitespace,indentation,names,quotation,protocol \
 		securedrop/source_templates/*.html securedrop/journalist_templates/*.html
 	@echo
 
--- a/securedrop/journalist_app/__init__.py
+++ b/securedrop/journalist_app/__init__.py
@@ -3,6 +3,7 @@ from pathlib import Path
 from typing import Any, Optional, Tuple, Union
 
 import i18n
+import server_os
 import template_filters
 import version
 from db import db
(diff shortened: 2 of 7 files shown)


legoktm:
Marking this as ready for review now.

zenmonkeykstop:
I think that I need to get the checker script merged first - it's not clear to me from the test plan what the JSON file should contain.

legoktm:
> I think that I need to get the checker script merged first - it's not clear to me from the test plan what the JSON file should contain.

Sorry for not making that clear, but also thanks for merging that other PR!

The format is basically:
```json
{
    "ssh": true,
    "free_space": false,
}
```

etc., where false means the check failed and the banner should be displayed. The main special case is `{"error": True}`, which also means the banner should be displayed. The JI intentionally doesn't care about the key names (aside from error) so if needed, we can trivially add more checks to the checker script and the JI won't need modifications.

zenmonkeykstop (approved):
LGTM, test plan checks out.

PR #7350, opened 2024-11-22
Switch systemd units to use Type=exec instead of Type=oneshot

### LINKED ISSUE #7349: Our systemd services are probably using Type=oneshot wrong
Based on reading https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html,

`Type=oneshot` is for services that you want to be considered "up" after they finish running, in conjunction with `RemainAfterExit`. But we don't do that, so as the doc explains, "the service will never enter "active" unit state, but will directly transition from "activating" to "deactivating" or "dead", since no process is configured that shall run continuously".

I think we want `Type=exec`, which triggers the command, moves to "running" as it runs, and then once it stops, moves to "dead".

We also have a number of services that don't specify any Type, and are defaulting to `Type=simple`. I don't think this is wrong, but exec seems like it has slightly better semantics.

Since I introduced all the oneshot ones, I'm going to fix those and leave the simple ones alone.

legoktm:
## Status

Ready for review

## Description of Changes

`oneshot` is designed to be used with RemainAfterExit, but we don't need that nor use it. After reviewing
<https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html>, `exec` appears to be the correct type since we want the service to be running once the binary has started, not once it's finished.

Practically this shouldn't make a big difference except that `systemctl start <name>` won't wait for the entire command to complete.

Fixes #7349.

## Testing

* [x] staging CI passes
* [x] visual review

## Deployment

## Checklist

### DIFF
--- a/securedrop/debian/config/lib/systemd/system/securedrop-cleanup-ossec.service
+++ b/securedrop/debian/config/lib/systemd/system/securedrop-cleanup-ossec.service
@@ -2,6 +2,6 @@
 Description=Cleanup OSSEC diff queue
 
 [Service]
-Type=oneshot
+Type=exec
 ExecStart=/usr/bin/securedrop-cleanup-ossec.py
 User=root
--- a/securedrop/debian/config/lib/systemd/system/securedrop-reboot-required.service
+++ b/securedrop/debian/config/lib/systemd/system/securedrop-reboot-required.service
@@ -2,6 +2,6 @@
 Description=Touch /var/run/reboot-required file
 
 [Service]
-Type=oneshot
+Type=exec
 ExecStart=/usr/bin/touch /var/run/reboot-required
 User=root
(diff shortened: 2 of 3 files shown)


cfm (approved):
Thanks, @legoktm.  This—

> Note that this means **systemctl start** command lines for `simple` services will report success even if the service's binary cannot be invoked successfully (for example because the selected `User=` doesn't exist, or the service binary is missing).[^1]

—makes me think that we really should be using `exec` throughout, to avoid missing services that have actually failed to execute.  Any reason not to make that change here while you're at it?  I'll go ahead and approve this in case you'd rather merge it and open a follow-up ticket.

[^1]: https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html

legoktm:
I agree that we should use `exec` but, I don't think "...to avoid missing services that have actually failed to execute." is correct.

AIUI with `simple`, systemctl start foo will always return exit code 0, even if the binary is missing. With `exec`, systemctl start will fail. But in both cases the service itself will be marked as failed, the behavioral difference is whether the service gets marked as running, or fails before that.

So I'd also like to move to `exec` but just wanted to limit this one to fixing the `oneshot` usage because that was more egregiously wrong.

PR #7353, opened 2024-11-23
Only try to apply grsec_lock once

legoktm:
## Status

Ready for review

## Description of Changes

Currently we specify both `sysctl_set: yes` and `reload: yes` when setting sysctl settings, which ends up with it being applied twice, first with `sysctl -w` (sysctl_set) and then through `sysctl -p` (reload).

With noble/Linux 6.6, setting the lock twice errors out, so just enable it once with `sysctl -p`. This is also closer to what the kernel will do normally when booting in which the whole file is loaded at once.

Refs #7323.

## Testing

* [ ] staging CI passes (this is a no-op, but tests will verify the sysctl flags are being applied)
* [ ] visual review

## Deployment

Any special considerations for deployment? only affects new installs, which is where the bug is

## Checklist

### DIFF
--- a/install_files/ansible-base/roles/grsecurity/tasks/apply_grsec_lock.yml
+++ b/install_files/ansible-base/roles/grsecurity/tasks/apply_grsec_lock.yml
@@ -22,7 +22,6 @@
   sysctl:
     name: "{{ item.name }}"
     value: "{{ item.value }}"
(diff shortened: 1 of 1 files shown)


zenmonkeykstop (approved):
LGTM

PR #7355, opened 2024-11-25
Have migrate-ssh-group handle the source group not existing (fresh install)

legoktm:
## Status

Ready for review

## Description of Changes

On a fresh install, the source group doesn't exist, so it would fail with a KeyError when trying to look up the group's membership.

Since this is only possible on a new install or post-focal upgrade, we can short circuit and skip the rest of the script.

## Testing

* [ ] try noble fresh install and not have this script fail

## Deployment

Any special considerations for deployment? This handles the fresh install case.

## Checklist


zenmonkeykstop (approved):
LGTM based on visual review.

PR #7360, opened 2024-11-28
Add CI staging job testing Ubuntu 24.04 (noble)

### LINKED ISSUE #7312: Set up noble staging job
## Description

We're pretty close to having a working noble install, so we should set up a staging job to verify it.

I naively tried setting up a noble CI staging job previously, it didn't work and really didn't even get very far. The bento/ubuntu-24.04 box appears to be a "version 2" box as described at https://vagrant-libvirt.github.io/vagrant-libvirt/boxes.html#box-formats. Running it with our old-ish vagrant version turns into "No image virtual size specified for box" (which is technically correct, it's not in the metadata.json file). Note that we can't easily upgrade to the latest version of vagrant because of the licensing change.

zenmonkeykstop:
## Status
Ready for review

## Description of Changes

Fixes #7312 

- adds libvirt-staging-noble molecule scenario
- updates staging CI to run against both Focal and Noble
- fixes most of the failing staging tests, rationales are in the commit messages

## Testing

- [ ] Visual review
- [ ] A staging(noble) job runs in addition to CI, and completes the `testinfra` tests.
- [ ] CI is green, esp. the two staging jobs


legoktm:
Remaining staging failures:
* [ ] There's no longer apparmor profiles for `sbin/dhclient` and `/usr/lib/connman/scripts/dhclient-script` because they're no longer shipped in noble
* [ ] same with `/usr/sbin/tcpdump` I think?
* [ ] `test_interface_up` is failing against the source interface
* [ ] `ip6tables -S` is exiting with a zero status code despite IPv6 being fully disabled at the kernel level. Might be a nftables change or Linux 6.6. Maybe there's a better way to verify IPv6 is fully disabled?
* [ ] paxtest outputs different results

legoktm:
I've just been pushing stuff here for now as I work through the failures, later we can figure out if we want them to get individual PRs or not.

legoktm:
An interesting quirk of ip6tables on noble; despite IPv6 being disabled the command still works, so on boot you get a wide-open ruleset of:
```
sdadmin@mon$ sudo ip6tables -S
-P INPUT ACCEPT
-P FORWARD ACCEPT
-P OUTPUT ACCEPT
```

This is because netfilter-persistent is smart and checks if IPv6 is disabled, and if so, doesn't load rules. So our "drop everything" in `install_files/ansible-base/roles/restrict-direct-access/files/iptables_rules_v6` is never even loaded. We should still keep it around as a defense-in-depth just in case IPv6 is ever actually turned on, it'll get loaded and drop everything.

legoktm:
I fixed the paxtest failure on focal because of unattended-upgrades preventing it from being installed.

The output on noble is uh, off. On app-staging, it's reliably:
```
...
Return to function (strcpy)              : paxtest: return address contains a NULL byte.
Return to function (memcpy)              : Killed
Return to function (strcpy, PIE)         : paxtest: return address contains a NULL byte.
Return to function (memcpy, PIE)         :
```
(the last line doesn't have a result)

And then on mon-staging:
```
...
Return to function (strcpy)              : paxtest: return address contains a NULL byte.
Return to function (memcpy)              : Return to function (strcpy, PIE)         : paxtest: return address contains a NULL byte.
Return to function (memcpy, PIE)         :
```
(interspersed output, missing two results)

legoktm:
@zenmonkeykstop I'm curious if you have any ideas on why the paxtest output is different across hosts and how to design a test to address it?

zenmonkeykstop:
> @zenmonkeykstop I'm curious if you have any ideas on why the paxtest output is different across hosts and how to design a test to address it?

Not off the top of my head -  assuming same kernel and paxctld setup the results should be the same. I'll look into it once we're done with 2025JanPlanfest.

legoktm:
Ack; I've set the test to be skipped on noble so we can get a green check and land this and fix it incrementally instead of holding the entire staging job back - I've flipped it out of draft mode.

The individual commit messages have rationales for all the further changes.

zenmonkeykstop (commented):
@legoktm your changes LGTM - I can't merge coz original author, will flag in channel

legoktm (approved):
Per Kev's review

PR #7363, opened 2024-12-02
Fix free space check by measuring bytes, not kilobytes

legoktm:
## Status

Ready for review

## Description of Changes

I didn't realize that df defaults to kilobytes, so our calculations were off by a factor of a whole SI prefix. We can pass `-B1` to df to calculate plain bytes instead.

## Testing

* [x] Run `df /` locally, see the output is for 1K blocks (kilobytes), then switch to `df -B1 /` and see that you get bytes (compare with updated test output)
* [x] CI passes, esp. Rust tests

## Deployment

## Checklist


cfm (approved):
Looks good!  Blocked on rebasing after #7364.

PR #7368, opened 2024-12-06
Wipe and regenerate Redis password when restoring from backups

### LINKED ISSUE #7328: backup regression regarding the redis password
## Description

I haven't verified it yet, but there's probably a regression in the backup/restore prodecure now that we set a redis password. The password is set in config.py, which we backup, but we don't backup the redis.conf, so when it is restored, that same password won't be re-added to the redis config (a different one will be randomly generated).

We're also not backing up the new rq_config.py file.

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Fixes #7328 .

Updates the ansible `restore` role to remove and regenerate the Redis password.

## Testing

- set up a production SecureDrop instance (h/w or virtual) using the current release, and add journalist accounts and sources/submissions
- take a backup of the system with `./securedrop-admin backup`
- check the current redis password, eg. with `ssh app sudo cat /etc/redis/redis.conf`
- switch to this branch on the Admin workstation
- restore from the backup file with `./securedrop-admin restore <path to backup>`
  - [x] verify that the backup completes successfully and expected accounts were restored
  - [x] verify that the Redis password has changed and that the same password is listed in `/var/www/securedrop/config.py`, `/var/www/securedrop/rq_config.py`, and `/etc/redis/redis.conf`
  - [x] verify that you can create a new source account and reply to it in the journalist interface.

Choose one of the following:

### DIFF
--- a/install_files/ansible-base/roles/restore/tasks/perform_restore.yml
+++ b/install_files/ansible-base/roles/restore/tasks/perform_restore.yml
@@ -121,7 +121,24 @@
     exclude: "var/lib/tor,etc/tor/torrc"
   when: restore_skip_tor
 
-- name: Reconfigure securedrop-app-code
+- name: Remove Redis password line from restored config.py, if it exists
+  lineinfile:
+    state: absent
+    path: /var/www/securedrop/config.py
+    regexp: "^REDIS_PASSWORD = .*$"
+
+- name: Remove /var/www/securedrop/rq_config.py if it exists
+  file:
+    state: absent
+    path: /var/www/securedrop/rq_config.py
+
+- name: Remove Redis password line from /etc/redis/redis.conf, if it exists
+  lineinfile:
+    path: /etc/redis/redis.conf
+    state: absent
+    regexp: "^requirepass .*$"
+
+- name: Reconfigure securedrop-app-code, regenerating Redis config vi postint
   command: dpkg-reconfigure securedrop-app-code
 
 - name: Reconfigure securedrop-config


install_files/ansible-base/roles/restore/tasks/perform_restore.yml
```diff
@@ -121,7 +121,24 @@
...
+    state: absent
+    path: /var/www/securedrop/rq_config.py
+
+- name: Remove Redis password line from /etc/redis/redis.conf, if it exists
+  lineinfile:
+    path: /etc/redis/redis.conf
+    state: absent
+    regexp: "^requirepass .*$"
+
+- name: Reconfigure securedrop-app-code, regenerating Redis config vi postint
```
cfm, line 141: Nit: I'd prefer to keep the "Reconfigure" task's description to just `Reconfigure securedrop-app-code` and then annotate each "Remove" task with a comment that explains this lifecycle, like so:

```yaml
- name: Remove Redis password line from restored config.py, if it exists
  # Will be regenerated during "Reconfigure securedrop-app-code" below.
```

But up to you.  If you keep this:

```suggestion
- name: Reconfigure securedrop-app-code, regenerating Redis config via postint
```

cfm (approved):
Test plan checks out!  One nit inline, but approving for merge at your discretion, @zenmonkeykstop.

install_files/ansible-base/roles/restore/tasks/perform_restore.yml
zenmonkeykstop: We'll fix it in release bump :)

PR #7377, opened 2024-12-11
Run securedrop-remove-packages hourly instead of daily

legoktm:
## Status

Ready for review 

## Description of Changes

This is an incredibly quick thing so let's run it more frequently so it is taken care of sooner and reduces the risk of the check script complaining that ufw is still present.

https://www.freedesktop.org/software/systemd/man/latest/systemd.time.html#Calendar%20Events indicates that "hourly" is a supported shorthand.

## Testing

* [ ] visual review

## Deployment

## Checklist

### DIFF
--- a/securedrop/debian/config/lib/systemd/system/securedrop-remove-packages.timer
+++ b/securedrop/debian/config/lib/systemd/system/securedrop-remove-packages.timer
@@ -2,7 +2,7 @@
 Description=Remove ufw and haveged if installed
 
 [Timer]
-OnCalendar=daily
+OnCalendar=hourly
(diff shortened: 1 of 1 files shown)


zenmonkeykstop:
Any expected impact in terms of OSSEC alerts or logging spam with the higher frequency?

legoktm:
No, it should only trigger OSSEC alerts upon package removal, which is just a one-time operation. After that the service won't even start because the ConditionPathExists= checks won't be met.

zenmonkeykstop (approved):
LGTM based on visual review

Issue #7372, opened 2024-12-09
Release SecureDrop 2.11.0


zenmonkeykstop:
This is a tracking issue for the release of SecureDrop 2.11.0

Tentatively scheduled as follows:

**Pre-release announcement:** 2024-12-10
**Release date:** 2024-12-17 (tentative)

**Release manager:** @zenmonkeykstop  
**Deputy release manager:** @legoktm 
**Localization manager:** @cfm
**Communications manager:** @nathandyer 

_SecureDrop maintainers and testers:_ As you QA 2.11.0, please report back your testing results as comments on this ticket. File GitHub issues for any problems found, tag them "QA: Release".

Test debian packages will be posted on https://apt-test.freedom.press signed with [the test key](https://gist.githubusercontent.com/conorsch/ec4008b111bc3142fca522693f3cce7e/raw/2968621e8ad92db4505a31fcc5776422d7d26729/apt-test%2520apt%2520pubkey).

# [QA Matrix for 2.11.0](https://docs.google.com/spreadsheets/d/1civB5BzOtPKvNJ1UP6Xq7v6NtPFSyzsgLEutZPxOIY0/edit?gid=349367281#gid=349367281)
# [Test Plan for 2.11.0](https://github.com/freedomofpress/securedrop/wiki/2.11.0-Test-Plan)

# Prepare release candidate (2.11.0~rc1)
- [x] Link to latest version of Tails, including release candidates, to test against during QA
- [x] Prepare 2.11.0~rc1 release changelog
- [x] Branch off release/2.11.0 from develop
- [x] Prepare 2.11.0-rc1
- [x] Build debs, preserving build log, and put up `2.11.0~rc1` on test apt server
- [x] Commit build log.

# Prepare release candidate (2.11.0~rc2)
- [x] Link to latest version of Tails, including release candidates, to test against during QA
- [x] Prepare 2.11.0~rc2 release changelog
- [x] Prepare 2.11.0-rc2
- [x] Build debs, preserving build log, and put up `2.11.0~rc2` on test apt server
- [x] Commit build log.

After each test, please update the QA matrix and post details for Basic Server Testing, Application Acceptance Testing and release-specific testing below in comments to this ticket.

# Final release
- [ ] ~Ensure builder in release branch is updated and/or update builder image~ (no longer in use)
- [x] Push signed tag 
- [x] Pre-Flight: Test updater logic in Tails (apt-qa tracks the `release` branch in the LFS repo)
[...]

PR #7383, opened 2024-12-11
SecureDrop 2.11.0~rc2

legoktm:
## Status

Ready for review 

## Description of Changes

Refs #7372.

## Testing

* [ ] visual review


changelog.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 2.11.0~rc2
+
+* Address feedback on source strings (#7370)
+* Improve noble migration check script output (#7369)
+* Run securedrop-remove-packages hourly instead of daily (#7377)
+* Allow apache2 to read /etc/securedrop-noble-migration.json in apparmor (#7378)
```
cfm: Fine for an RC changelog; I suggest clarifying for the final release, but take what's useful and leave the rest:

```suggestion
* Address translators' feedback on UI strings (#7370)
* Improve output of Ubuntu Noble pre-migration check script (#7369)
* Run `securedrop-remove-packages` hourly instead of daily (#7377)
* Allow `apache2` to read `/etc/securedrop-noble-migration.json` under AppArmor (#7378)
* Remove stray Ubuntu file `/etc/apt/apt.conf.d/zzzz-temp-installer-unattended-upgrade` if it exists (#7380)
```

cfm (changes requested):
I suggest adding #7380, along with some other non-blocking suggestions inline.

NB. This diff includes some whitespace changes further down in the file that I agree we might as well keep.

changelog.md
legoktm: Thanks, squashed your changes in.

cfm (approved):
Thanks.  Assuming we want this to be the new head of the release branch, this looks good to me after #7384.

changelog.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 2.11.0~rc2
+
+* Address translators' feedback on UI strings (#7370)
+* Improve output of Ubuntu Noble pre-migration check script (#7369)
+* Run `securedrop-remove-packages` hourly instead of daily (#7377)
```
zenmonkeykstop, line 7: When we have follow-up PRs in RCs to a new feature I tend to prefer rolling them up into one entry. For example, if we find a bug fixed in PR P2 in a new feature PR P1 in RC1, I'd just put them as "* new feature! (P1, P2)". This is partly for brevity but also to make it clear that the bugfix in P2 is not a production fix. It's good as is, but I'd likely do said rollup for the prod changelog unless there are objections.

legoktm: Yep; I thought we'd do rollup at prod changelog time.

PR #7400, opened 2024-12-20
Run `make update-python3-requirements` in the dev container (again)

### LINKED ISSUE #7245: move `make update-pip-requirements` back into development containers
After I initially reviewed #7234, @legoktm added in <https://github.com/freedomofpress/securedrop/issues/7234#issuecomment-2402639037>:

> I'm gonna make one change, we don't actually need to run this in the container anymore, so it'll be even faster.

I didn't retest with that change, and we've just discovered why I should have: without the container, the available system Python is what's used to build packages, and `apxs` is required from the `apache2-dev` package.  @legoktm thinks these are reasons to move `make update-pip-requirements` back into the development containers after all.

legoktm:
## Status

Ready for review 

## Description of Changes

Some packages, especially mod-wsgi, have dependencies on system packages and cannot universally be run on the host.

Moving it back into the container will be a bit slower, but overall it should still be noticibly faster than pre-uv.

Fixes #7245.
## Testing

* [x] running `make update-python3-requirements` works

## Deployment

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ update-admin-pip-requirements:  ## Update admin requirements.
 .PHONY: update-python3-requirements
 update-python3-requirements:  ## Update Python 3 requirements with pip-compile.
 	@echo "███ Updating Python 3 requirements files..."
-	@$(SDBIN)/update-requirements
+	@SLIM_BUILD=1 $(DEVSHELL) $(SDBIN)/update-requirements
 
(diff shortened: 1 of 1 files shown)


legoktm:
I tested this PR by creating the dependency part of https://github.com/freedomofpress/securedrop/pull/7401 using it.

cfm (approved):
Thanks, @legoktm.  It's _definitely_ still faster!

PR #7401, opened 2024-12-20
Lint our GitHub Actions workflows with zizmor

legoktm:
## Status

Ready for review

## Description of Changes

zizmor is a new tool to lint GitHub Actions workflows. For the most part our workflows are pretty low risk since we don't give it a bunch of credentials, but we can avoid issues in the future by locking them down now.

Two overall issues needed to be fixed:
* setting persist-credentials: false for actions/checkout, which we do everywhere except in the workflows that need to push.
* Don't use template expansion when we can use a normal bash variable.

While zizmor is written in Rust, it is also shipped as a prebuilt binary via PyPI, so we can set it as a pip dependency and run it as part of our normal lint CI.

Refs <https://github.com/freedomofpress/securedrop-tooling/issues/18>.

## Testing

* [ ] visual review
* [ ] CI passes

## Deployment

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,6 +25,8 @@ jobs:
       artifact_id: ${{ steps.upload.outputs.artifact-id }}
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
       - uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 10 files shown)


zenmonkeykstop (approved):
LGTM

PR #7403, opened 2025-01-06
Enable flake8-bugbear rules via ruff

legoktm:
## Status

Ready for review

#### Upgrade ruff to 0.8.6
    
Switch to f-string and remove no-longer needed override.

#### Enable flake8-bugbear rules via ruff
    
Address most of the issues found by the flake8-bugbear rules,
and suppress a few that aren't worth fixing right now. The remaining
ones will be fixed in follow-up commits.

#### Don't use finally to return a redirect
    
As described in
<https://docs.astral.sh/ruff/rules/jump-statement-in-finally/> and more
recently in <https://peps.python.org/pep-0765/>, using return in a
finally block has the potential of suppressing exceptions.

Given that in both cases we have a blanket exception handler, it should
be safe to return outside of a finally. If there are exceptions being
suppressed, we should surface them.

#### pretty_bad_protocol: Suppress warning about modifying os.environ
    
As described in
<https://docs.astral.sh/ruff/rules/assignment-to-os-environ/>, assigning
to os.environ doesn't actually clear the environment (i.e. remove keys
that are not set in the new value).

I think it is better to just explicitly suppress the warning since the
existing code works and isn't worth modifying.

#### Have migration test actually fail if value isn't missing
    
flake8-bugbear noticed that the second loop was checking the wrong
value. I noticed that the test wouldn't actually fail if the expected
exception wasn't raised, so use pytest.raises, which will correctly fail
if it doesn't work as expected.

## Testing

* [ ] visual review
* [ ] CI passes

## Deployment

## Checklist


securedrop/journalist_app/admin.py
```diff
@@ -85,10 +85,9 @@ def manage_config() -> Union[str, werkzeug.Response]:
                     gettext("Unable to process the image file. Please try another one."),
                     "logo-error",
                 )
-            finally:
-                return redirect(url_for("admin.manage_config") + "#config-logoimage")
+            return redirect(url_for("admin.manage_config") + "#config-logoimage")
         else:
-            for field, errors in list(logo_form.errors.items()):
+            for errors in list(logo_form.errors.values()):
```
cfm: Can we remove the `list()` calls while we're here, since we're iterating over `errors` anyway?  (We don't bother with with `list()` at all in `source_app.main`.)

securedrop/pretty_bad_protocol/_meta.py
```diff
@@ -296,7 +296,10 @@ def update_path(environment, path):  # type: ignore[no-untyped-def]
                 :param list path: A list of strings to update the PATH with.
                 """
                 log.debug("Updating system path...")
-                os.environ = environment
+                # This assignment is mostly to reset the monkey patch, but it doesn't
+                # actually reset the environment. Leaving it alone since it works as-is
+                # despite being wrong.
+                os.environ = environment  # noqa: B003
```
cfm: Since this is a partial vendoring anyway, intuitively I flip the burden of proof: we'd have to justify the change, not the non-change.  How about:

```suggestion
                # This assignment doesn't reset the environment, but it does reset the monkey-patch as intended.  Leaving as-is from upstream.
                os.environ = environment  # noqa: B003
```

I defer to you; I won't wordsmith you further.  :-)

cfm (dismissed):
Thanks, @legoktm.  Couple of nits for your consideration; looks great otherwise.

securedrop/pretty_bad_protocol/_meta.py
legoktm: That works for me, thanks & done.

cfm (approved):
Thanks, @legoktm.

PR #7404, opened 2025-01-06
Switch systemd services to use Type=exec instead of Type=simple

### LINKED ISSUE #7358: Switch systemd units to Type=exec instead of default Type=simple
## Description

Follow-up from https://github.com/freedomofpress/securedrop/issues/7349 and https://github.com/freedomofpress/securedrop/pull/7350#pullrequestreview-2459744086; most of our systemd units are using the default of Type=simple but we'd prefer the Type=exec semantics.

legoktm:
## Status

Ready for review

## Description of Changes

Per <https://github.com/freedomofpress/securedrop/pull/7350#pullrequestreview-2459744086>, exec is the ideal semantics we want, in which systemd verifies it can actually execute the command we want before marking the unit as running.

Aside from that very subtle difference, there should be no user-facing changes.

Fixes #7358.

## Testing

* [x] Visual review
* [x] CI passes

## Deployment

## Checklist

### DIFF
--- a/molecule/testinfra/app-code/test_securedrop_rqrequeue.py
+++ b/molecule/testinfra/app-code/test_securedrop_rqrequeue.py
@@ -17,6 +17,7 @@ def test_securedrop_rqrequeue_service(host):
             "Wants=redis-server.service",
             "",
             "[Service]",
+            "Type=exec",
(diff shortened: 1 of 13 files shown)


cfm (approved):
Thanks, @legoktm.

PR #7406, opened 2025-01-07
Script to upgrade from focal to noble

### LINKED ISSUE #7332: Create focal -> noble upgrade script

legoktm:
## Status

Ready for review

## Description of Changes

The script is split into various stages where progress is tracked on-disk. The script is able to resume where it was at any point, and needs to, given multiple reboots in the middle.

The new noble-upgrade.json file shipped in the securedrop-config package is used to control the upgrade process.

Fixes #7332.

## Testing

#### Preparation

* [x] Build focal and noble packages off this PR (`UBUNTU_VERSION=focal make build-debs` and `UBUNTU_VERSION=noble make build-debs`).
   * [x] Until https://github.com/freedomofpress/securedrop/issues/7411 is resolved, you will also need to build noble ossec packages: `UBUNTU_VERSION=noble make build-debs-ossec`
* [x] Copy the respective focal and noble folders to a accessible website 
* [x] Create a `dist` folder alongside focal and noble, run `apt-ftparchive packages ./focal > dists/focal/main/binary-amd64/Packages` and `apt-ftparchive packages ./noble > dists/noble/main/binary-amd64/Packages` (you'll need to create the intermediate directories in dist/ manually)
   * As an example, you can look at how I've set up https://legoktm.com/apt/ - but don't use these packages, they're out of date and volatile! 
* [x] Set up a 2.11.1 staging/prod install (n.b. I've only tested this on physical hardware, nothing should stop it from working on a virtualized setup though)

The next step of steps need to be applied to both app and mon:

* [x] Add your custom apt repository to the `/etc/apt/sources.list.d/apt_freedom_press.list` file with a line like `deb [trusted=yes] https://example.org/apt focal main`. The `[trusted=yes]` bypasses PGP signature checking so we don't need to also fiddle with signing the temporary packages and installing the keyring.
* [ ] Run `sudo apt-get update &&` ~~` sudo unattended-upgrade`~~`sudo apt upgrade` to upgrade to the 2.12.0-rc0 packages.
* [x] Edit `/lib/systemd/system/securedrop-noble-migration-upgrade.service` to add the line `Environment=EXTRA_APT_SOURCE="deb [trusted=yes] https://example.org/apt noble main"` (note that this says noble and not focal! also keep an eye on the quotes)
* [x] Reboot.

#### Upgrading

This should be repeated twice, once for app and then once that's done, for mon.

* [x] verify `/etc/securedrop-noble-migration-state.json` was created with `{"finished": None, "bucket": 1-5}`.
* [x] open a background window that runs `journalctl -f` - primarily to follow along the progress
* [x] edit `/usr/share/securedrop/noble-upgrade.json`, to set `app.enabled = true` and `app.bucket = 5`. (or mon if that's what you're upgrading)
* [x] Wait for the securedrop-noble-migration-upgrade systemd timer to start (no more than 3 minutes). You can also initiate it manually with `sudo systemctl start securedrop-noble-migration-upgrade`.
* [x] The server should reboot (you'll need to restart your journalctl -f window). Once it comes back, `/etc/securedrop-noble-migration-state.json` should now have finished PendingUpdates.
* [x] Wait again for the securedrop-noble-migration-upgrade systemd timer to start. You should see apt's progress in the journalctl window, it'll take a while depending on internet and hardware speed.
* [x] if this is the app migration, you should be able to verify that the SI/JI are unreachable (apache is masked)
* [x] Eventually it should reboot again. Once it's back, `/etc/securedrop-noble-migration-state.json` should be at Reboot.
* [x] Wait for the securedrop-noble-migration-upgrade systemd timer to start again, once it does it should pretty quickly reach the `Done` stage.
* [x] `cat /etc/os-release` should output noble.

#### Verification

After the upgrade:

* [ ] Running `./securedrop-admin verify` should pass.
* [x] Basic SI/JI functionality should work
* [x] OSSEC notifications are coming through like expected

#### Misc.

* [x] visual review of code
* [x] review of https://github.com/freedomofpress/securedrop/wiki/noble-upgrade-architecture
* [ ] CI passes

## Deployment

2. New installs.

Choose one of the following:

- `admin/requirements.in`
  container)

(remove) this section.

Choose one of the following:


legoktm:
I think the script is basically complete at this point, but I haven't actually tried it yet. So I need to do that, and then figure out how we're going to do CI on it. I think we should ideally be able to take the focal staging environment, upgrade it, and then re-run testinfra (noble) checks on it.

legoktm:
Fixed a number of issues found by actual test runs, currently hit:
```
# apt-get upgrade --without-new-pkgs --force-confold --force-confdef
E: Command line option --force-confold is not understood in combination with the other options
```

Will get to that tomorrow.

legoktm:
Interesting wrinkle, because the systemd unit is installed by the Debian package, it smartly wants to restart the service during package upgrade. Except that kills the apt-get process upgrading the package, which totally breaks everything.

 I'm trying to figure out how to stop that restart, but it doesn't seem like dh_systemd_start's `--no-restart-after-upgrade` is doing what it's documented to do. One option I'm trying is systemd's `RefuseManualStart`/`RefuseManualStop`.

Alternatively we could have the script fork in a way that it doesn't get killed when the service stops - similar to what unattended-upgrades does.

legoktm:
I'm thinking of getting rid of the --without-new-pkgs step and just doing a single full-upgrade step, it'll avoid a lot of the dependency constraint weirdness that keeps manifesting in weird ways. For example, I dropped the apparmor-utils dependency, which meant that Python 3.12 got pulled in later, and then app-code got upgraded too early.

legoktm:
> Alternatively we could have the script fork in a way that it doesn't get killed when the service stops - similar to what unattended-upgrades does.

I think this is the way to go, but with a slightly different variant. We should just ensure the apt-get/dpkg processes don't get killed. It's okay if the upgrade script dies, as long as apt-get keeps going. The commands are all idempotent so when the script gets restarted by the timer, it'll re-run the apt-get command it'll do nothing and not kill itself, and then keep moving on.

Plus by just keeping apt-get alive, we don't need to implement any locking, etc ourselves, because dpkg already takes care of all of that in a battle-tested manner.

legoktm:
The changes I just pushed introduce a new `check_call_nokill` that runs commands in a separate process group. Combined with `KillMode=process` in the systemd unit, the nokill commands will continue to run and live, even when the systemd unit is restarted. 

To visualize:

* systemd starts securedrop-noble-migration-upgrade.service via timer
   * upgrade script runs `apt-get full-upgrade` (in nokill mode)
      * apt-get updates packages, including securedrop-config
* securedrop-config's postinst stops the upgrade.service, killing the securedrop-noble-migration-upgrade process, but not apt-get
* systemd starts  securedrop-noble-migration-upgrade.service via timer
   * upgrade script runs `apt-get full-upgrade` (in nokill mode)
      * if apt has not yet finished from the previous invocation: it'll error on the lock and the upgrade script will die (to be resumed on the next timer invocation)
      * if apt has finished, then it'll be a quick no-op and the upgrade script will proceed.

One gotcha here is that we can no longer reliably capture stdout/stderr because if the parent is killed, it'll go nowhere. So we will need to send it to a file presumably.

legoktm:
I mostly got through a full automated migration; something is going wrong during the installation of iptables-persistent/nftables-persistent and /etc/iptables/rules.{v4,v6} are blank, so there's no firewall up, causing the integrity check to fail. Once I bypassed that, it reached the done stage. \o/

legoktm:
I successfully completed a fully-automated app migration today, so I'm marking this as ready for review. I still need to write up a more comprehensive test plan and stuff but at least the code can begin to be looked at.

legoktm:
I've written up the full test plan now.

legoktm:
Also as far as the code review goes, I'll try to split this up into more commits to simplify review. I also want to write a brief architecture document that explains how it's all supposed to work.

legoktm:
> I also want to write a brief architecture document that explains how it's all supposed to work.

https://github.com/freedomofpress/securedrop/wiki/noble-upgrade-architecture

legoktm:
> Also as far as the code review goes, I'll try to split this up into more commits to simplify review.

I thought it was going to be more, but in the end it's two commits: 1) move some logic out of the check.rs file into a new Rust lib.rs and 2) the migration script and everything else.

legoktm:
(Rebased on top of the Rust upgrade)

securedrop/debian/config/lib/systemd/system/securedrop-noble-migration-upgrade.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=Run noble migration
+
+RefuseManualStop=true
+
+[Service]
+Type=exec
+Environment=RUST_LOG=debug
+Environment=LAUNCHED_BY_SYSTEMD=1
```
cfm, line 9: `RUST_LOG` is self-explanatory, but would you mind documenting `LAUNCHED_BY_SYSTEMD`?  I had to google a bit before I realized that this is a pattern, not an environment variable used or suggested by systemd itself.

securedrop/debian/securedrop-app-code.postinst
```diff
@@ -311,8 +311,14 @@ case "$1" in
     database_migration
 
     # Restart apache now that we've updated everything, setup AppArmor
-    # and applied all migrations
-    service apache2 restart
+    # and applied all migrations. Only restart if it is not masked, which
+    # it is during the noble migration.
```
cfm: Clarification:

```suggestion
    # and applied all migrations. While the Noble migration is in progress, the service unit is masked and must not be restarted.
```

noble-migration/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+impl State {
+    pub fn is_ready(&self) -> bool {
+        self.is_ready_except_apt() && self.apt
+    }
+
+    /// For when developers inject extra APT sources for testing
+    pub fn is_ready_except_apt(&self) -> bool {
+        self.ssh && self.ufw && self.free_space && self.systemd
+    }
+}
```
cfm, line 30: I'm reading these functions as (pseudocode):

* `is_ready()`
* `is_ready(prod=false)`

I probably would have favored intensive over extensive names like these, but it's not important for a single-use tool like this where we don't expect the gap between testing and production criteria to grow.

noble-migration/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+
+    bail!("Could not find VERSION_CODENAME in /etc/os-release")
+}
+
+/// Check that the UNIX "ssh" group has no members
+///
+/// See <https://github.com/freedomofpress/securedrop/issues/7316>.
+fn check_ssh_group() -> Result<bool> {
+    // There are no clean bindings to getgrpname in rustix,
+    // so jut shell out to getent to get group members
```
cfm:
```suggestion
    // so just shell out to getent to get group members
```

noble-migration/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+        println!("ssh OK: group is empty");
+        Ok(true)
+    } else {
+        println!("ssh ERROR: group is not empty: {members:?}");
+        Ok(false)
+    }
+}
+
+/// Parse the output of `getent group ssh`, return true if empty
+fn parse_getent_output(stdout: &str) -> Result<Vec<&str>> {
```
cfm, line 81: This is really above and beyond!  I would have been perfectly happy with line-counting and verbatim output here.  :-)

noble-migration/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+    } else {
+        Ok(members.split(',').collect())
+    }
+}
+
+/// Check that ufw was removed
+///
+/// See <https://github.com/freedomofpress/securedrop/issues/7313>.
+fn check_ufw_removed() -> bool {
+    if Path::new("/usr/sbin/ufw").exists() {
```
cfm, line 101: I don't feel strongly about it, but why check for the binary rather than the package as in <https://github.com/freedomofpress/securedrop/blob/3e20adee7fd2a30f3d256833f67d958625e69f5f/molecule/testinfra/common/test_system_hardening.py#L189>?  Just what's easy to test from Rust versus under testinfra?

noble-migration/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+
+/// Estimate the size of the backup so we know how much free space we'll need.
+///
+/// We just check the size of `/var/lib/securedrop` since that's really the
+/// data that'll take up space; everything else is just config files that are
+/// negligible post-compression. We also don't estimate compression benefits.
+fn estimate_backup_size() -> Result<u64> {
+    let path = Path::new("/var/lib/securedrop");
+    if !path.exists() {
+        // mon server
```
cfm, line 118: Is it worth checking this assumption, as `upgrade.rs` does in `is_mon_server()`?

noble-migration/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+    let mut total: u64 = 0;
+    let walker = WalkDir::new(path);
+    for entry in walker {
+        let entry = entry.context("walking /var/lib/securedrop failed")?;
+        if entry.file_type().is_dir() {
+            continue;
+        }
+        let metadata = entry.metadata().context("getting metadata failed")?;
+        total += metadata.len();
+    }
```
cfm, line 130: If we're comfortable shelling out elsewhere, why not just defer to the output of `du -s /var/lib/securedrop` (as `check_free_space()` defers to `df`)?

noble-migration/src/lib.rs
```diff
@@ -0,0 +1,307 @@
...
+}
+
+/// Sizes are in bytes
+struct DfOutput {
+    total: u64,
+    free: u64,
+}
+
+fn parse_df_output(stdout: &str) -> Result<DfOutput> {
```
cfm, line 178: Can this parsing be simplified with `df --output=<fields>`, perhaps even multiple calls to `df --output=<field>` for different values of `field`?

noble-migration/src/bin/upgrade.rs
```diff
@@ -0,0 +1,670 @@
+//! Migrate a SecureDrop server from focal to noble
+//!
+//! This script should never be run directly, only via the
+//! systemd service.
+use anyhow::{bail, Context, Result};
```
cfm: Clarification:

```suggestion
//! This script must never be run directly, only via the
//! systemd service, which is enforced by checking $LAUNCHED_BY_SYSTEMD in main() below.
use anyhow::{bail, Context, Result};
```

noble-migration/src/bin/upgrade.rs
```diff
@@ -0,0 +1,670 @@
...
+        process::{CommandExt, ExitStatusExt},
+    },
+    path::Path,
+    process::{self, Command, ExitCode},
+};
+
+/// Package-provided instructions on whether auto-migrations should run
+const CONFIG_PATH: &str = "/usr/share/securedrop/noble-upgrade.json";
+/// Serialized version of `State`
+const STATE_PATH: &str = "/etc/securedrop-noble-migration-state.json";
```
cfm: Clarification:

```suggestion
/// Configuration for the migration process (installed by securedrop-config)
const CONFIG_PATH: &str = "/usr/share/securedrop/noble-upgrade.json";
/// Serialized version of `State` (left by the last run of this script)
const STATE_PATH: &str = "/etc/securedrop-noble-migration-state.json";
```

noble-migration/src/bin/upgrade.rs
```diff
@@ -0,0 +1,670 @@
...
+    RestoreIptables,
+    ReenableUnattendedUpdates,
+    ReenableOSSEC,
+    CleanApacheFolder,
+    Reboot,
+    SwitchUbuntuSources,
+    IntegrityCheck,
+    ReenableApache2,
+    RemoveBackup,
+    Done,
```
cfm: Nit: It would be nice if these names were verbose and consistent enough (e.g., all imperative) to read like a checklist of steps, without having to consult the documentation of each stage's corresponding function.

noble-migration/src/bin/upgrade.rs
```diff
@@ -0,0 +1,670 @@
...
+}
+
+/// Current state of the migration.
+#[derive(Serialize, Deserialize, Debug)]
+struct State {
+    /// most recently finished stage. This is updated
+    /// on-disk as soon as every stage is completed.
+    finished: Stage,
+    /// randomly selected number between 1 and 5, which is used
+    /// to auto-migrate instances in chunks. This number is generated once,
```
cfm: "Chunks" surprised me here; how about "batches" or "cohorts"?

noble-migration/src/bin/upgrade.rs
```diff
@@ -0,0 +1,670 @@
...
+        fs::write(
+            STATE_PATH,
+            serde_json::to_string(self).context("Failed to serialize state")?,
+        )
+        .context("Failed to write state")
+    }
+}
+
+/// State machine to invoke the next stage based on what has finished.
+fn run_next_stage(state: &mut State) -> Result<()> {
```
cfm, line 116: Very elegant.

noble-migration/src/bin/upgrade.rs
```diff
@@ -0,0 +1,670 @@
...
+    match state.finished {
+        Stage::None => {
+            pending_updates(state)?;
+            // n.b. this is unreachable because we've already rebooted
```
cfm: Clarification:

```suggestion
            // REBOOT LIKELY HERE: If there are pending updates to be applied, we won't actually reach this state.set() call until we've installed them, rebooted, and then reentered this state with no moe pending updates.
```

noble-migration/src/bin/upgrade.rs
```diff
@@ -0,0 +1,670 @@
...
+/// Step 12: Restore iptables rules
+///
+/// Now that we've upgraded, move the backups of the iptables rules we created
+/// into place and clean up the backups.
+fn restore_iptables() -> Result<()> {
+    info!("Restoring iptables...");
+    fs::copy("/etc/iptables/rules.v4.backup", "/etc/iptables/rules.v4")?;
+    fs::copy("/etc/iptables/rules.v6.backup", "/etc/iptables/rules.v6")?;
+    fs::remove_file("/etc/iptables/rules.v4.backup")?;
+    fs::remove_file("/etc/iptables/rules.v6.backup")?;
```
cfm: Are these worth preserving (timestamping) in some way just in case?  Or would we want to defer to Ansible if something went wrong here?

(This makes me wish we had the ability to run the testinfra suite locally after upgrade, without requiring an administrator to run `securedrop-admin verify` manually.)

noble-migration/src/bin/upgrade.rs
```diff
@@ -0,0 +1,670 @@
...
+/// We delete this with ansible, but the upgrade brings it back (but empty)
+/// so remove it.
+fn clean_apache_folder() -> Result<()> {
+    if is_mon_server() {
+        return Ok(());
+    }
+    info!("Cleaning /var/www/html folder...");
+    fs::remove_dir("/var/www/html")?;
+    Ok(())
+}
```
cfm: If we're duplicating this enforcement step from Ansible, can we consolidate it in `securedrop-app-code`?

noble-migration/src/bin/upgrade.rs
```diff
@@ -0,0 +1,670 @@
...
+    mon: HostUpgradeConfig,
+}
+
+#[derive(Deserialize)]
+struct HostUpgradeConfig {
+    /// whether upgrades are enabled
+    enabled: bool,
+    /// all servers <= this bucket will be upgraded, i.e. 1 upgrades bucket 1,
+    /// 2 upgrades buckets 1 and 2, etc.
+    bucket: usize,
```
cfm: Clarification:

```suggestion
    /// The `bucket` setting increases inclusively: i.e., `bucket=1` enables hosts in bucket 1 to upgrade; `bucket=2` enables hosts in buckets 1 and 2 to upgrade; etc.
    bucket: usize,
```

cfm (commented):
This is terrific work, @legoktm.  The upgrade bucketing and state machine are very elegant.

My comments here are intended (a) to clarify things and (b) to see if there are simplifications worth adopting.  If not, that's fine too: none of these suggestions are blocking, and I'll begin testing tomorrow either way.  If there's anything that you'd like to discuss, rather than just either addressing or dismissing, let's chat to save us some back-and-forth here!

securedrop/debian/config/lib/systemd/system/securedrop-noble-migration-upgrade.service
legoktm: Yeah, it's just a simple check to prevent the script from being launched outside of systemd to prevent someone from accidentally running it manually but also simple enough for developers to set if they do want to run it manually.

noble-migration/src/bin/upgrade.rs
legoktm: batches sounds good.

noble-migration/src/bin/upgrade.rs
legoktm: If you look at pending_updates, you'll see it ends with:

```rust
    state.set(Stage::PendingUpdates)?;
    check_call("systemctl", &["reboot"])?;
    // Because we've initiated the reboot, do a hard stop here to ensure that
    // we don't keep moving forward if the reboot doesn't happen instantly
    process::exit(0);
```

So the reboot is unconditional.

Instead of the comment, we could stick in `unreachable!()` which would enforce that code can't reach there, and panic if it does.

noble-migration/src/bin/upgrade.rs
legoktm: In general: I would be okay leaving them on the system somewhere, but I would prefer not to keep extra files in `/etc/iptables`. Maybe `/etc/securedrop-backup/`?

If this step did go wrong, it would be caught at the `integrity_check` step, at which point it's safe to re-run the install playbook to regenerate it.

noble-migration/src/bin/upgrade.rs
legoktm: Yes, that's a good idea.

legoktm (commented):
Thanks for the review, I'll push fixes in a bit. I didn't respond to most of the comments in lib.rs because that code isn't new, it was already shipped in check.rs and I just moved it into lib.rs so it can be reused by upgrade.rs. The only new part should be the `is_ready_except_apt` you commented on - sorry for not making this clear.

Given that, do you think the modifications you suggested, which were all reasonable, are still worth it? From one side it's good that they're getting more scrutiny now that they're being used in a different and more critical context, but also we know that the current implementations work in production, so I'm not sure changing them is worth it.

cfm:
@legoktm in <https://github.com/freedomofpress/securedrop/pull/7406#pullrequestreview-2578686092>:

> Given that, do you think the modifications you suggested, which were all reasonable, are still worth it? From one side it's good that they're getting more scrutiny now that they're being used in a different and more critical context, but also we know that the current implementations work in production, so I'm not sure changing them is worth it.

That's very fair.  Let's leave `lib.rs` well enough alone, then—I've marked remaining threads there as resolved—and I'll defer to you on what's still worth doing in `upgrade.rs`.

noble-migration/src/bin/upgrade.rs
legoktm: They are now saved as `/etc/securedrop-backup/iptables-rules.v{4,6}` and not removed once restored.

noble-migration/src/bin/upgrade.rs
legoktm: Done.

noble-migration/src/bin/upgrade.rs
legoktm: Done.

noble-migration/src/bin/upgrade.rs
legoktm: Done.

legoktm:
I believe I've addressed everything except for the nit regarding imperative naming; I'm down to do it, just trying to address the more important stuff first.

cfm (commented):
Here's testing on production VMs: success!  A few notes inline.


#### Preparation

* [x] Build focal and noble packages off this PR (`UBUNTU_VERSION=focal make build-debs` and `UBUNTU_VERSION=noble make build-debs`).
   * [x] Until https://github.com/freedomofpress/securedrop/issues/7411 is resolved, you will also need to build noble ossec packages: `UBUNTU_VERSION=noble make build-debs-ossec`
* [x] Copy the respective focal and noble folders to a accessible website 
* [x] Create a `dist` folder alongside focal and noble, run `apt-ftparchive packages ./focal > dists/focal/main/binary-amd64/Packages` and `apt-ftparchive packages ./noble > dists/noble/main/binary-amd64/Packages` (you'll need to create the intermediate directories in dist/ manually)
   * As an example, you can look at how I've set up https://legoktm.com/apt/ - but don't use these packages, they're out of date and volatile! 
* [x] Set up a 2.11.1 staging/prod install (n.b. I've only tested this on physical hardware, nothing should stop it from working on a virtualized setup though)

The next step of steps need to be applied to both app and mon:

* [x] Add your custom apt repository to the `/etc/apt/sources.list.d/apt_freedom_press.list` file with a line like `deb [trusted=yes] https://example.org/apt focal main`. The `[trusted=yes]` bypasses PGP signature checking so we don't need to also fiddle with signing the temporary packages and installing the keyring.
* [ ] Run `sudo apt-get update && sudo unattended-upgrade` to upgrade to the 2.12.0-rc0 packages.

Afterwards:

```sh-session
root@app-prod:/home/vagrant# apt-cache policy securedrop-app-code
securedrop-app-code:
  Installed: 2.11.1+focal
  Candidate: 2.12.0~rc1+focal
  Version table:
     2.12.0~rc1+focal 500
        500 https://homes.cs.washington.edu/~cfmyers/securedrop focal/main amd64 Packages
 *** 2.11.1+focal 500
        500 https://apt.freedom.press/ focal/main amd64 Packages
        100 /var/lib/dpkg/status
```

I had to use `apt upgrade`.

* [x] Edit `/lib/systemd/system/securedrop-noble-migration-upgrade.service` to add the line `Environment=EXTRA_APT_SOURCE="deb [trusted=yes] https://example.org/apt noble main"` (note that this says noble and not focal! also keep an eye on the quotes)
* [x] Reboot.


#### Upgrading

This should be repeated twice, once for app and then once that's done, for mon.

* [x] verify `/etc/securedrop-noble-migration-state.json` was created with `{"finished": None, "bucket": 1-5}`.
* [x] open a background window that runs `journalctl -f` - primarily to follow along the progress
* [x] edit `/usr/share/securedrop/noble-upgrade.json`, to set `app.enabled = true` and `app.bucket = 5`. (or mon if that's what you're upgrading)
* [x] Wait for the securedrop-noble-migration-upgrade systemd timer to start (no more than 3 minutes). You can also initiate it manually with `sudo systemctl start securedrop-noble-migration-upgrade`.
* [x] The server should reboot (you'll need to restart your journalctl -f window). Once it comes back, `/etc/securedrop-noble-migration-state.json` should now have finished PendingUpdates.
* [x] Wait again for the securedrop-noble-migration-upgrade systemd timer to start. You should see apt's progress in the journalctl window, it'll take a while depending on internet and hardware speed.

NB.  One needs to restart `journalctl -f` after systemd and journald restart with `Journal stopped`.

* [x] if this is the app migration, you should be able to verify that the SI/JI are unreachable (apache is masked)
* [x] Eventually it should reboot again. Once it's back, `/etc/securedrop-noble-migration-state.json` should be at Reboot.
* [x] Wait for the securedrop-noble-migration-upgrade systemd timer to start again, once it does it should pretty quickly reach the `Done` stage.
* [x] `cat /etc/os-release` should output noble.

#### Verification

After the upgrade:

* [ ] Running `./securedrop-admin verify` should pass.

~~This is still running (slowly, per #7428); I'll check up on it in the morning.~~

Modulo:

- https://github.com/freedomofpress/securedrop/pull/7406#issuecomment-2627877849
- https://github.com/freedomofpress/securedrop/pull/7406#issuecomment-2628014491

* [x] Basic SI/JI functionality should work
* [x] OSSEC notifications are coming through like expected

...which were suspended during the upgrade on `mon` but not on `app`, as expected.

legoktm:
> I had to use apt upgrade.

Ah yeah, my bad, unattended-upgrades won't work unless we specifically configure it for your apt repository ("Origin"). Let me update the test plan to say `apt upgrade` is fine.

cfm:
> - [ ] Running `./securedrop-admin verify` should pass.
> 
> This is still running (slowly, #7428); I'll check up on it in the morning.

With #7428 cherry-picked onto this branch, testinfra is passing up to exactly the place expected by <https://github.com/freedomofpress/securedrop/pull/7428#issuecomment-2617030976>.  :-)

cfm:
And, correcting for <https://github.com/freedomofpress/securedrop/pull/7428#issuecomment-2617030976>, testinfra passes with this patch:

```patch
--- a/molecule/testinfra/app/test_apparmor.py
+++ b/molecule/testinfra/app/test_apparmor.py
@@ -93,7 +93,7 @@ def test_aastatus_unconfined(host):
         aa_status_output = host.check_output("aa-status")
         assert unconfined_chk in aa_status_output
 
-
+@pytest.mark.skip_in_prod
 def test_aa_no_denies_in_syslog(host):
     """Ensure that there are no apparmor denials in syslog"""
     with host.sudo():
```

cfm (approved):
I won't merge this, since I expect we'll wind up doing further testing next week, but I'm happy to give initial approval based on <https://github.com/freedomofpress/securedrop/pull/7406#pullrequestreview-2585579583>.

legoktm:
> I won't merge this, since I expect we'll wind up doing further testing next week

No pressure to merge from me, but given the size of this PR and that we're starting to have dependent work on it (ansible playbook PR, soon the Rust compiler upgrade), I'd be down to land this sooner rather than later and incrementally fix anything that comes up as part of testing.

And, thank you for the review and approval @cfm!!

cfm:
@legoktm in <https://github.com/freedomofpress/securedrop/pull/7406#issuecomment-2628538740>:
 
> No pressure to merge from me, but given the size of this PR and that we're starting to have dependent work on it (ansible playbook PR, soon the Rust compiler upgrade), I'd be down to land this sooner rather than later and incrementally fix anything that comes up as part of testing.

That's persuasive.  :-)  Merging—onwards!

PR #7409, opened 2025-01-09
Use a single script to set and reset Redis password

### LINKED ISSUE #7386: [2.11.0-rc2] Redis reconfiguration can yield multiple `requirepass` directives
## Description

We now set or reset a `requirepass` directive in `/etc/redis/redis.conf` in three places, with different conditions and via different mechanisms:

1. Initial installation playbook:
    https://github.com/freedomofpress/securedrop/blob/0798935bdd43a362c08c24ee7156303b0bdd3c23/install_files/ansible-base/roles/app/tasks/initialize_securedrop_app.yml#L156-L163
2. `securedrop-app-code.postinst`:
    https://github.com/freedomofpress/securedrop/blob/0798935bdd43a362c08c24ee7156303b0bdd3c23/securedrop/debian/securedrop-app-code.postinst#L200
3. Backup-restoration playbook:
    https://github.com/freedomofpress/securedrop/blob/0798935bdd43a362c08c24ee7156303b0bdd3c23/install_files/ansible-base/roles/restore/tasks/perform_restore.yml#L135-L142

This has evolved ad hoc and can lead to inconsistent (and difficult-to-reproduce) configuration.

## Steps to Reproduce

Long-running QA instance....

## Expected Behavior

Consistent Redis configuration, including exactly one `requirepass` directive.

## Actual Behavior

I run [etckeeper](https://documentation.ubuntu.com/server/how-to/backups/install-etckeeper/) on my QA instance for forensic purposes.  On automatic upgrade to `securedrop-app-code_2.11.0~rc1`, etckeeper recorded (NB. QA values; no real secrets leaking here):

```patch

From 4f76e15a7698d6d695d5f7682c093114a01c07f0 Mon Sep 17 00:00:00 2001
From: root <root@app>
Date: Wed, 11 Dec 2024 03:15:56 +0000
Subject: [PATCH 2/2] committing changes in /etc made by "/usr/bin/python3
 /usr/bin/unattended-upgrade"

Package changes:
---
 redis/redis.conf | 1 +
 1 file changed, 1 insertion(+)

diff --git a/redis/redis.conf b/redis/redis.conf
index 3052aff..8704f5d 100644
--- a/redis/redis.conf
+++ b/redis/redis.conf
@@ -1371,3 +1371,4 @@ rdb-save-incremental-fsync yes
 # active-defrag-max-scan-fields 1000
 
 requirepass iECwDLwqmO2gDCJgMkaUMKkmtqgTih3IF951DMmO0PA=
+requirepass LeabNIFsavoNC6yQm1tjaCn+vCsFBLYFNMPxKHQzq7s=
-- 
2.25.1
```

## Comments

@legoktm and I discussed adding a "re/configure Redis" script that:
1. would be our sole entry-point for Redis configuration across Ansible playbooks and `postinst` scripts;
2. would enforce consistency across both Redis's own and consumers' configuration; and
3. wouldn't be strictly idempotent, but could be safely rerun without violating that consistency.

legoktm:
## Status

Ready for review

## Description of Changes

Code to set and reset Redis passwords has now proliferated into the postinst, dev environment, ansible provisioning and backup restore. Instead of maintaining separate code for this, write a single script that handles it all.

A new `securedrop-set-redis-auth.py` tool offers three operations:
* check: verify all the passwords are in sync (for testinfra checks)
* reset: forcibly change the password everywhere (for backup restore)
* reset-if-needed: change the password everywhere if needed (for postinst)

Fixes #7386.

## Testing

* [x] Visual review
* [ ] Staging CI passes

## Deployment

Both upgrades and fresh installs are considered.

## Checklist

### DIFF
--- a/securedrop/debian/app-code/usr/bin/securedrop-set-redis-auth.py
+++ b/securedrop/debian/app-code/usr/bin/securedrop-set-redis-auth.py
@@ -1,0 +1,134 @@
+#!/usr/bin/env python3
+
+import argparse
+import re
+import secrets
+import shutil
+import sys
+from pathlib import Path
+from typing import Optional
+
+CONFIG_PY = Path("/var/www/securedrop/config.py")
+RQ_CONFIG_PY = Path("/var/www/securedrop/rq_config.py")
+REDIS_CONF = Path("/etc/redis/redis.conf")
+
+PYTHON_RE = re.compile(r"^REDIS_PASSWORD = ['\"](.*?)['\"]$")
+REDIS_CONF_RE = re.compile(r"^requirepass (.*?)$")
+
+
+def read_python_file(path: Path) -> Optional[str]:
+    """Extract the password from a Python file"""
+    if not path.exists():
+        # rq_config.py might not exist yet
+        return None
+    contents = path.read_text()
+    # Read in reverse because we want to look for the last matching line
+    # since it'll take precedence in Python
+    for line in contents.splitlines()[::-1]:
+        match = PYTHON_RE.match(line)
+        if match:
+            return match.group(1)
+    # Nothing found
+    return None
+
+
+def write_python_file(path: Path, password: str) -> None:
+    """Set the new password in a Python file, removing any existing passwords"""
+    if path.exists():
+        lines = path.read_text().splitlines()
+    else:
+        lines = []
+    # Take the existing file, remove any matching password lines, and then add
+    # our new password at the end
+    lines = [line for line in lines if not PYTHON_RE.match(line)]
+    lines.append(f"REDIS_PASSWORD = '{password}'")
+    if path == RQ_CONFIG_PY and not path.exists():
+        # Ensure rq_config.py is created with the correct permissions
+        path.write_text("")
+        path.chmod(0o640)
+        shutil.chown(path, "root", "www-data")
+
+    path.write_text("\n".join(lines) + "\n")
+
+
+def read_redis_conf(path: Path) -> Optional[str]:
+    """Extract the password from our redis.conf file"""
+    # n.b. we assume redis.conf exists, since the package should already
+    # be installed
+    contents = path.read_text()
+    # Read in reverse because we want to look for the last matching line
+    # since redis uses the last requirepass stanza
+    for line in contents.splitlines()[::-1]:
+        match = REDIS_CONF_RE.match(line)
+        if match:
+            return match.group(1)
+    # Nothing found
+    return None
+
+
+def write_redis_conf(path: Path, password: str) -> None:
+    """Set the new password in a redis.conf file, removing any existing passwords"""
+    if not path.exists():
+        raise RuntimeError("redis.conf does not already exist")
+    lines = path.read_text().splitlines()
+    # Take the existing file, remove any matching password lines, and then add
+    # our new password at the end
+    lines = [line for line in lines if not REDIS_CONF_RE.match(line)]
+    lines.append(f"requirepass {password}")
+    path.write_text("\n".join(lines) + "\n")
+
+
+def generate_password() -> str:
+    """
+    Generate a base64-encoded, 32-byte random string
+
+    This is roughly equivalent to `head -c 32 /dev/urandom | base64`
+    """
+    return secrets.token_urlsafe(32)
+
+
+def check() -> bool:
+    config_py = read_python_file(CONFIG_PY)
+    rq_config_py = read_python_file(RQ_CONFIG_PY)
+    redis_conf = read_redis_conf(REDIS_CONF)
+    all_passwords = {config_py, rq_config_py, redis_conf}
+    # If any are None, then we don't have the password set yet
+    if None in all_passwords:
+        return False
+    # True if there's only one unique password
+    return len(all_passwords) == 1
+
+
+def reset() -> None:
+    password = generate_password()
+    write_python_file(CONFIG_PY, password)
+    write_python_file(RQ_CONFIG_PY, password)
+    write_redis_conf(REDIS_CONF, password)
+    print("Redis password has been reset; please restart redis/apache2")
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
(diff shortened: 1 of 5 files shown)


legoktm:
staging is failing because test_grsecurity_paxtest can't install paxtest because unattended-upgrades is still running. I thought I had fixed that in 6214f5390a30380f51a328f28d3b1fe24acb7ff3 but I guess it wasn't sufficient. I'll file a follow-up task for that but it shouldn't block this PR.

securedrop/debian/app-code/usr/bin/securedrop-set-redis-auth.py
```diff
@@ -0,0 +1,125 @@
...
+    This is roughly equivalent to `head -c 32 /dev/urandom | base64`
+    """
+    return secrets.token_urlsafe(32)
+
+
+def check() -> bool:
+    config_py = read_python_file(CONFIG_PY)
+    rq_config_py = read_python_file(RQ_CONFIG_PY)
+    redis_conf = read_redis_conf(REDIS_CONF)
+    return config_py is not None and config_py == rq_config_py and rq_config_py == redis_conf
```
cfm, line 90: Nit: I found the 1+2 conditionals confusing at first; how about:

```suggestion
    configs = {config_py, rq_config_py, redis_conf}
    if None in configs:
        return False
    return len(configs) == 1
```

cfm: Nit: I found the 1+2 conditionals confusing at first; how about:

```suggestion
    configs = {config_py, rq_config_py, redis_conf}
    if None in configs:
        return False
    return len(configs) == 1
```

cfm (commented):
Here's a visual review with one nit, @legoktm.  Otherwise this looks great, and I'll test it out tomorrow on my long-running instance, including:

- [x] Backup
- [x] Upgrade
- [x] Restore

securedrop/debian/app-code/usr/bin/securedrop-set-redis-auth.py
legoktm: Done!

securedrop/debian/app-code/usr/bin/securedrop-set-redis-auth.py
```diff
@@ -0,0 +1,130 @@
...
+        # Ensure rq_config.py is created with the correct permissions
+        path.write_text("")
+        path.chmod(0o640)
+        shutil.chown(path, "root", "www-data")
+
+    path.write_text("\n".join(lines) + "\n")
+
+
+def read_redis_conf(path: Path) -> Optional[str]:
+    """Extract the password from our redis.conf file"""
```
cfm, line 55: Shouldn't this also have (from `read_python_file()`):

```suggestion
    """Extract the password from our redis.conf file"""
    if not path.exists():
        # redis.conf might not exist yet
        return None
```

cfm (changes requested):
This works great, @legoktm.  One more code-level suggestion inline.

And, after reviewing the Ansible changes, one late-breaking question of approach:  Why bother with the `reset-if-needed` mode at all?  Why not *always* `reset` on every `postinst`, whether triggered by installation, upgrade, or restoration of a backup?  `check` is still useful for infrastructure tests, as you note.  But going from `{check,reset-if-needed,reset}` to just `{check,reset}` would simplify both the script and the logic around all of its invocations.

Put another way: Do we care that the Redis password is *persistent*? or do we only care that it's *consistent*?


---

```sh-session
sdadmin@app:~$ sudo /usr/bin/securedrop-set-redis-auth.py check
Yay, all three passwords are the same!
sdadmin@app:~$ sudo /usr/bin/securedrop-set-redis-auth.py reset-if-needed
All three passwords are the same; nothing changed
```

Journalist Interface login works.

```sh-session
sdadmin@app:~$ sudo /usr/bin/securedrop-set-redis-auth.py reset          
Redis password has been reset; please restart redis/apache2
sdadmin@app:~$ sudo systemctl restart redis
```

Journalist Interface returns HTTP `500`.

```sh-session
sdadmin@app:~$ sudo systemctl restart apache2
```

Journalist Interface session works again.

securedrop/debian/app-code/usr/bin/securedrop-set-redis-auth.py
legoktm: No, we should be able to safely assume that redis.conf was installed by the package, so failing loudly here if it doesn't exist is the right way to go. I can leave a comment to that effect though.

legoktm:
> But going from `{check,reset-if-needed,reset}` to just `{check,reset}` would simplify both the script and the logic around all of its invocations.

Agreed that getting rid of reset-if-needed would simplify the script, but I think it shifts the complexity, because we are now resetting the redis password incredibly more frequently, on every SD release. My current feeling is that the extra logic in the script is worth it so we aren't adding more stuff to the upgrade process in making a reset an every-time operation instead of a one-time thing.

You could easily argue this the other way, that the best way to ensure it's robust is to run it regularly (like alembic), and I would probably buy that when thinking longer-term, but I also don't want to throw more moving parts into upgrades now.

securedrop/debian/app-code/usr/bin/securedrop-set-redis-auth.py
legoktm: Probably writing a blank redis.conf with just a password would be a bad thing, and this is implicitly making sure that doesn't happen, so let me add an explicit check that it *does* exist.

cfm:
Thanks, @legoktm.  Your distinction between long-term robustness versus short-term conservatism is persuasive to me.

securedrop/debian/app-code/usr/bin/securedrop-set-redis-auth.py
legoktm: Done.

cfm (approved):
Thanks, @legoktm.  Approving pending CI, expecting unrelated failures on Focal (#7410) and on Noble.

PR #7417, opened 2025-01-17
Ensure /etc/iptables exists before writing to it

### LINKED ISSUE #7119: Reinstalling mon with existing app server triggers missing iptables-persistent dependency
## Description

During a recent SD install to a freshly provisioned server running Ubuntu Server 20.04.6, the `./securedrop-admin install` process ran into an error during the run, citing that the `/etc/iptables` directory was missing and that it couldn't continue.

To workround this, I had to do an `apt install iptables-persistent` on the server, at which point `./securedrop-admin install` was able to complete the installation without issue.

We should determine if this is a required package, and if so, add it as a dependency.

## Steps to Reproduce

1. Install a fresh copy of Ubuntu Server 20.04.6 and configure it according to the SD docs
2. Copy SSH keys from the Admin Workstation to it
3. Run `./securedrop-admin install` (along with `setup` and `sdconfig` beforehand, if necessary)

## Expected Behavior

The install happens without issue.

## Actual Behavior

An error regarding a missing `/etc/iptables` directory that stops the installation.

## Comments

Not sure if this was a one-off, and also not sure of any potential security concerns about the `iptables-persistent` package.

legoktm:
## Status

Ready for review

## Description of Changes

In the specific case of installing a fresh mon server when the app server is already configured AND you're using ssh over the local network, we'll try to write to /etc/iptables before the iptables-persistent package is installed.

This is because we end up running the restrict-direct-access role before the common role, which installs the base packages.

The easy fix is to install iptables-persistent ahead of time if we see that it's necessary.

Fixes #7119.

## Testing

* [x] setup a 2.11.1 prod install that uses ssh over localhost
* [x] on the admin workstation, cherry-pick this patch 
* [x] reinstall ubuntu fresh on just the mon server, and then run `./securedrop-admin install` (i.e. with the cherry-pick)
* [x] the install should complete successfully with no errors about /etc/iptables being missing.

## Deployment

Any special considerations for deployment? n/a, bug only affects fresh installs

## Checklist

### DIFF
--- a/install_files/ansible-base/roles/restrict-direct-access/tasks/iptables.yml
+++ b/install_files/ansible-base/roles/restrict-direct-access/tasks/iptables.yml
@@ -41,6 +41,19 @@
     admin_net_cidr: "{{ '/'.join([admin_net_info.network, admin_net_info.netmask])|ipaddr('cidr') }}"
   delegate_to: localhost
 
+- name: Check iptables-persistent was installed
+  stat:
+    path: /etc/iptables/
+  register: etc_iptables_check
+
+- name: Install iptables-persistent if needed
+  apt:
+    name: iptables-persistent
+    state: present
+    update_cache: yes
+    cache_valid_time: 3600
+  when: not etc_iptables_check.stat.exists
+
 - name: Copy IPv4 iptables rules.
   template:
     src: rules_v4


legoktm:
(This is submitted untested)

cfm:
> * [ ] reinstall ubuntu fresh on just the mon server, and then run `./securedrop-admin install` (i.e. with the cherry-pick)

After cfm/terraform-metal-securedrop-staging@06a19ed619aaec0cd73b6f4250382f4a1720796e I can use `vagrant snapshot {push,pop}` for this step, which should make this pretty quick to test in production VMs.

cfm (approved):
Thanks, @legoktm.  Testing the "install into freshly-reinstalled Monitor Server" path was slowed down by my troubleshooting #7425, but the actual fix here is straightforward and works as expected.

PR #7424, opened 2025-01-23
Set a fixed APT::Machine-ID so phased updates are consistent

### LINKED ISSUE #7422: Consistently apply phased updates across all SD instances
## Description

Ubuntu noble introduces phased updates, in which non-security updates are slowly rolled out. This means that SDs can end up in a non-homogenus state. Given that we want to keep that state, we should figure out how to keep all SDs in the same bucket.

legoktm:
## Status

Ready for review

## Description of Changes

We generally try to keep SecureDrops as homogenus as possible. APT's new phased updates can cause issues here if different instances are running different sets of packages.

As explained in <https://discourse.ubuntu.com/t/phased-updates-in-apt-in-21-04/20345>, we can override the `APT::Machine-ID` configuration setting so that all instances apply the exact same set of phased updates.

Per machine-id(5), the value is supposed to be a "hexadecimal, 32-character, lowercase ID". The hash is a random value that was selected; clever reviewers will be able to ascertain the source ;-)

Fixes #7422

## Testing

* [ ] staging CI passes - at least the `test_fixed_phasing` test.

## Deployment

Any special considerations for deployment? Only takes effect on noble, should be a no-op for focal.

## Checklist

### DIFF
--- a/securedrop/debian/config/opt/securedrop/50unattended-upgrades
+++ b/securedrop/debian/config/opt/securedrop/50unattended-upgrades
@@ -66,3 +66,7 @@ Dpkg::Options {
     "--force-confdef";
     "--force-confold";
 }
+
+// Set a fixed machine-id to ensure phased updates are consistent across all instances
+// see <https://discourse.ubuntu.com/t/phased-updates-in-apt-in-21-04/20345>.
+APT::Machine-ID "1ebf5f15850c540b3142f1584bdd496d";
--- a/molecule/testinfra/common/test_automatic_updates.py
+++ b/molecule/testinfra/common/test_automatic_updates.py
@@ -162,6 +162,13 @@ def test_unattended_upgrades_functional(host):
     assert expected_result in c.stdout
 
 
+def test_fixed_phasing(host):
+    """Verify APT's machine-id is set to a fixed value for consistent phasing"""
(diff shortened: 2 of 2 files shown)


legoktm:
the noble staging failure is unrelated (the newly added test passes).

securedrop/debian/config/opt/securedrop/50unattended-upgrades
```diff
@@ -66,3 +66,7 @@ Dpkg::Options {
     "--force-confdef";
     "--force-confold";
 }
+
+// Set a fixed machine-id to ensure phased updates are consistent across all instances
+// see <https://discourse.ubuntu.com/t/phased-updates-in-apt-in-21-04/20345>.
+APT::Machine-ID "1ebf5f15850c540b3142f1584bdd496d";
```
cfm, line 72: I confirmed via <https://utcc.utoronto.ca/~cks/space/blog/linux/Ubuntu2204ServerPhasedUpdates?showcomments> that `APT::Machine-ID` if empty is derived from `/etc/machine-id` but if set doesn't override it.

cfm (approved):
Thanks, @legoktm, for an elegant disablement of an elegant feature we just happen not to want.  :-)

PR #7426, opened 2025-01-24
fix: set up the `sdssh` group before using it in ACLs

### LINKED ISSUE #7425: servers can become inaccessible if SSH login is restricted to the `sdssh` group before it's applied to any users
## Description

Under certain (probably edge-case QA) conditions, a server may reboot (a) with SSH [restricted to the `sdssh` group](https://github.com/freedomofpress/securedrop/blob/034b834845105c726c0a998589bd0d053b1e05e6/install_files/ansible-base/roles/restrict-direct-access/templates/sshd_config#L25) before (b) it's [created](https://github.com/freedomofpress/securedrop/blob/034b834845105c726c0a998589bd0d053b1e05e6/install_files/ansible-base/roles/common/tasks/create_users.yml#L16-L21) and any users are [added](https://github.com/freedomofpress/securedrop/blob/034b834845105c726c0a998589bd0d053b1e05e6/install_files/ansible-base/roles/common/tasks/create_users.yml#L27) to it.

## Steps to Reproduce

Encountered while testing #7417.

## Expected Behavior

After [rebooting](https://github.com/freedomofpress/securedrop/blob/034b834845105c726c0a998589bd0d053b1e05e6/install_files/ansible-base/roles/grsecurity/tasks/main.yml#L16) into the grsecurity kernel, SSH connections are accepted and the installation continues.

## Actual Behavior

SSH fails with `Permission denied (publickey)`.

## Comments

This can be worked around by doing before reboot:

```sh-session
$ sudo groupadd sdssh
$ sudo usermod -a -G sdssh $(whoami)
```

Basically, the `restrict-direct-access` role depends on the `sdssh` group, which is set up by the later `common` role.

cfm:
## Status

Ready for review

## Description of Changes

Fixes #7425—I believe as non-invasively as possible.

## Testing

I've successfully installed a Monitor Server from scratch with this change, following #7417.  In addition, please consider:

- [ ] Are there reasons we *shouldn't* move the `create_users` task from the later `common` playbook to the earlier `restrict-direct-access` playbook?
- [ ] Staging CI passes.

## Deployment

Assuming that the above checks out, this change should have no deployment considerations: it changes the sequence of provisioning, not the final provisioned state.


legoktm (approved):
Thanks, this looks good. The noble staging failure is unrelated: `test_unattended_upgrades_functional`

I looked for other uses of the the SSH user since we're creating it at a different stage and the only use was in the reset-ssh-key playbook, which is an entirely separate thing.

Issue #6262, opened 2022-02-09
/var/log/syslog size over ~2.125MB makes testinfra tests hang


eaon:
## Description

`test_no_ecrypt_messages_in_logs` in `molecule/testinfra/common/test_system_hardening.py` triggered by `securedrop-admin verify` hangs indefinitely when confronted with a `/var/log/syslog` that's larger than approximately 2.125MB.

## Steps to Reproduce

* SSH into app server
* Run `sudo truncate -s 2125k /var/log/syslog`
* On Admin Workstation, run `securedrop-admin verify`

eaon:
I hit this again during 2.5.0 QA

legoktm:
Found this in the testinfra bug tracker: https://github.com/pytest-dev/pytest-testinfra/issues/141

> Yes, transferring the whole file across the network is not a good idea, you should use remote operations instead like "contains()"

legoktm:
Of course, there's a comment explaining why that doesn't work:
```
        # Not using `f.contains(<pattern>)` because that'd cause the sought
        # string to make it into syslog as a side-effect of the testinfra
        # invocation, causing subsequent test runs to report failure.
        assert error_message not in f.content_string
```

legoktm:
One approach could be to break up the pattern through string concatenation so it doesn't match in syslog.

legoktm:
I think the best approach to solve this would be to run something like `grep <pattern> /var/syslog/syslog`, and then post-process the output of that in Python, removing *our* grep command from that.

legoktm:
The ecryptfs test was fixed in https://github.com/freedomofpress/securedrop/pull/7428. But there's still test_aa_no_denies_in_syslog.

PR #7428, opened 2025-01-27
Speed up pam_ecryptfs testinfra check

legoktm:
## Status

Ready for review

## Description of Changes

After a noble migration, transmitting syslog back to the host so it can be searched can be pretty slow, so let's do it on the host. By using grep, we can escape the message we're looking for in a way that ensures our search doesn't match the error we're looking for.

Since there's already a `.` in the message, use re.escape(), which turns it into `\.`, which obfuscates our grep.

## Testing

* [x] staging CI passes (or this test doesn't fail)
* [x] visual review

## Deployment

## Checklist

### DIFF
--- a/molecule/testinfra/common/test_system_hardening.py
+++ b/molecule/testinfra/common/test_system_hardening.py
@@ -142,13 +142,11 @@ def test_no_ecrypt_messages_in_logs(host, logfile):
     Ensure pam_ecryptfs is removed from /etc/pam.d/common-auth : not only is
     no longer needed, it causes error messages (see issue #3963)
     """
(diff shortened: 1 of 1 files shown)


legoktm:
Running into the same issue with `test_aa_no_denies_in_syslog`; might patch that one tomorrow too.

cfm (approved):
Tested in <https://github.com/freedomofpress/securedrop/pull/7406#issuecomment-2627877849> and passes CI modulo #7433.  Thanks, @legoktm.

legoktm:
Note: https://github.com/freedomofpress/securedrop/issues/6262 was the old issue tracking this issue.

PR #7434, opened 2025-01-31
Fix syntax of systemd's ConditionPathExists

### LINKED ISSUE #7412: Syntax error in securedrop-remove-packages.service ConditionPathExists
## Description

https://github.com/freedomofpress/securedrop/blob/3f737a224de20b17a8b9b8be4a5f7f3f151aaead/securedrop/debian/config/lib/systemd/system/securedrop-remove-packages.service#L3-L4

Stupid typo on my part, `|=` is wrong, it's supposed to be `=|`. This shows up when you run `systemctl daemon-reload`.

There's no real adverse effect, it just ends up invoking apt more often than necessary, so I'm flagging this for the 2.12 release.

Would be good to figure out a regression test for this in the future, especially since systemd can clearly spot the error.

legoktm:
## Status

Ready for review

## Description of Changes

systemd prefers =|, not |=.

Fixes #7412.

## Testing

* [ ] visual review, cross-reference with systemd syntax if you want
* [ ] if you really really want, run `systemctl daemon-reload` on a staging/prod system and see that it doesn't emit warnings about the wrong syntax.

I initially held off because I wanted a regression test, and Ro is working on one in the workstation, so let's just get this in now and we can land the regression test later, as a non-release blocker.

## Deployment

## Checklist

### DIFF
--- a/securedrop/debian/config/lib/systemd/system/securedrop-remove-packages.service
+++ b/securedrop/debian/config/lib/systemd/system/securedrop-remove-packages.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=Remove ufw/haveged if installed
-ConditionPathExists|=/usr/sbin/ufw
-ConditionPathExists|=/usr/sbin/haveged
+ConditionPathExists=|/usr/sbin/ufw
+ConditionPathExists=|/usr/sbin/haveged
 
 [Service]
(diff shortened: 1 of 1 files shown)


rocodes (approved):
Approving based on visual review

PR #7438, opened 2025-02-03
Switch Ubuntu APT sources to use HTTPS

### LINKED ISSUE #3286: Use APT over HTTPS whenever possible
*Originally filed by @emkll in our infrastructure repo on 2017-11-13:*

Using TLS for APT would increase the difficulty of man-in-the-middle attacks at install/update time such as:

 * Injecting malicious packages should a signing key be compromised or package/signature verification silently fails.
 * Inspecting or preventing the install of certain packages on the host.

The following APT repos are used by SecureDrop:

 * http://security.ubuntu.com/ubuntu trusty-security main
 * http://security.ubuntu.com/ubuntu trusty-security main
 * http://security.ubuntu.com/ubuntu trusty-security universe
 * http://security.ubuntu.com/ubuntu trusty-security universe
 * https://apt.freedom.press trusty main
 * http://deb.torproject.org/torproject.org trusty main

*.ubuntu.com currently does **[not support HTTPS](https://bugs.launchpad.net/ubuntu/+bug/1464064)**, but some mirrors do [0]. `deb.torproject.org` does, but it will soon be mirrored by apt.freedom.press

[0]https://www.reddit.com/r/Ubuntu/comments/3q53kc/list_of_ubuntu_repository_mirrors_available_over/

legoktm:
## Status

Ready for review 

## Description of Changes

Ubuntu now supports HTTPS on their primary mirrors, including archive.ubuntu.com and security.ubuntu.com.

While APT verifies integrity using PGP signatures, there have been a few vulnerabilities in APT that would've been prevented by also layering TLS on top.

No attempt is made to update the configuration on existing instances; instead this change will be made during the noble migration.

Fixes #3286.

## Testing

* [x] staging CI passes

## Deployment

1. Upgrading existing production instances. -> taken care of during noble migration
3. New installs. -> handled in this PR via ansible

## Checklist

### DIFF
--- a/molecule/testinfra/common/test_automatic_updates.py
+++ b/molecule/testinfra/common/test_automatic_updates.py
@@ -39,10 +39,10 @@ def test_cron_apt_config(host):
 @pytest.mark.parametrize(
     "repo",
     [
-        "deb http://security.ubuntu.com/ubuntu {securedrop_target_platform}-security main",
-        "deb http://security.ubuntu.com/ubuntu {securedrop_target_platform}-security universe",
-        "deb http://archive.ubuntu.com/ubuntu/ {securedrop_target_platform}-updates main",
-        "deb http://archive.ubuntu.com/ubuntu/ {securedrop_target_platform} main",
+        "deb https://security.ubuntu.com/ubuntu {securedrop_target_platform}-security main",
+        "deb https://security.ubuntu.com/ubuntu {securedrop_target_platform}-security universe",
+        "deb https://archive.ubuntu.com/ubuntu/ {securedrop_target_platform}-updates main",
+        "deb https://archive.ubuntu.com/ubuntu/ {securedrop_target_platform} main",
     ],
 )
 def test_sources_list(host, repo):
@@ -74,13 +74,13 @@ def test_ubuntu_sources(host):
     assert f.user == "root"
     assert f.mode == 0o644
     expected = f"""\
-URIs: http://archive.ubuntu.com/ubuntu/
+URIs: https://archive.ubuntu.com/ubuntu/
 Suites: {distro} {distro}-updates
 Components: main universe restricted multiverse
 """
     assert f.contains(expected)
     expected_security = f"""\
-URIs: http://security.ubuntu.com/ubuntu/
+URIs: https://security.ubuntu.com/ubuntu/
 Suites: {distro}-security
 Components: main universe restricted multiverse
 """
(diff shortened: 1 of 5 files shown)


molecule/testinfra/common/test_automatic_updates.py
cfm, line 1: > No attempt is made to update the configuration on existing instances; instead this change will be made during the noble migration.

In that case, @legoktm, shouldn't these tests be parameterized on `securedrop_target_platform`?  Otherwise, won't an Admin Workstation updated to v2.12 expect to see `https://` URLs on a Focal-based Server running v2.12 still with `http://` URLs?

cfm (changes requested):

molecule/testinfra/common/test_automatic_updates.py
legoktm: Hmm, yes but also if someone does a focal fresh install on 2.12, then they'll also get HTTPS URLs.

But also with 2.12 we will enable noble fresh installs so maybe no one should be doing focal fresh installs at that point and we don't need to worry about it?

Or is tying this with the noble migration a bad idea and we should just have a mechanism to update the sources.list file outside of ansible installs regardless?

cfm: The case I'm concerned about is "upgrade to v2.12, but not yet Noble, and run testinfra", e.g.:

| Time | Admin Workstation tag | Testinfra checks for | Server OS version | Server SecureDrop version | URLs | ? |
| --- | --- | --- | --- | --- | --- | --- |
| 0 | 2.11.1 | `http://` | Focal | 2.11.1 | `http://` | ✓ |
| 1 | 2.12.0 | **`https://`** | Focal | 2.12.0 | **`http://`** | ✗ |

Put another way, I read the current diff as saying: "If you've upgraded to v2.12, but haven't yet upgraded to Noble, then `test_automatic_updates.py` will fail."  Am I contriving an edge case without consequence?

> Or is tying this with the noble migration a bad idea and we should just have a mechanism to update the sources.list file outside of ansible installs regardless?

I think having largely separate installation and upgrade configuration paths is a bad idea, but I know we can't fix that here!  :-)

legoktm:
Just a note that I'm not pursuing this for 2.12/noble migration, we'll do it after everyone is on noble.

legoktm:
Not actively pursuing this, can be resurrected when we do.

PR #6557, opened 2022-09-22
feat: update all languages in Weblate, but only offer supported ones in `securedrop-admin sdconfig`

cfm:
## Status

Ready for review

## Description of Changes

Towards #6156, which (with these changes) should be completed automatically as part of the v2.5.0 localization cycle.

Closes #6387 by:

1. creating a canonical list of _supported_ languages in `securedrop/i18n.json`;
2. relaxing `i18n_tool.py update-from-weblate` to pull translations for *all* locales available for translation in Weblate; thereby also
3. relaxing `i18n_tool.py translate-{messages,desktop}` to update source strings for *all* locales available for translation in Weblate; while
4. restricting `securedrop-admin sdconfig` to allow administrators to select only those languages both *present* (in the `securedrop` repository) and *supported* (in `i18n.json`).

### `securedrop-admin sdconfig`

This is the higher priority for testing, since it is an administrator-facing change.

1. Check out `6387-supported-languages`.
2. Run `securedrop-admin sdconfig` and attempt to enable a nonexistent locale (e.g., `foo`).
    - [x] The nonexistent locale is rejected.
3. Run `securedrop-admin sdconfig` and attempt to enable any two or more supported locales (e.g., `en_US de_DE ro`).
    - [x] All locales are accepted.
4. For one of the locales you enabled in step (3), delete the corresponding directory from `securedrop/translations`.  Then run `securedrop-admin sdconfig`.
    - [x] The locale no longer suggested and is rejected, because it is no longer present.
5. For another of the locales you enabled in step (3), delete the corresponding `supported_locales` entry in `securedrop/i18n.json`.  Then run `securedrop-admin sdonfig`.
    - [x] The locale is no longer suggested and is rejected, because it is no longer supported.

### `i18n_tool.py`

This is the lower priority for testing, since it's a tooling change we can tweak in place if necessary, including during the localization cycle.

- Run `securedrop/i18n_tool.py update-from-weblate`.
- [x] An `INFO` is logged to the standard output for each _unsupported_ language like:

    https://github.com/freedomofpress/securedrop/blob/efdc952d94166cc10d0837106c0def1f2808bcad/securedrop/i18n_tool.py#L327-L328

- [x] Changes to all languages, not only supported ones, are fetched, checked out, staged, and committed.

## Deployment

[...]

PR #7443, opened 2025-02-20
feat(`validate_locale_configuration`): discard and warn on configured but unsupported locale

### LINKED ISSUE #7442: test disablement of Hindi and Romanian
We will likely have to [revoke support](https://developers.securedrop.org/en/latest/supported_languages.html#revoking-support-for-a-language) for Hindi (`hi`) and Romanian (`ro`).  We implemented this case in #6156, but we've never used it in production.  I'd like to try it out to generate test cases for release QA.

cfm:
## Status

Ready for review

## Description of Changes

Closes #7442:  If `validate_locale_configuration()` doesn't check the authoritative list of supported locales in the `i18n.json` file installed by `securedrop-app-code`, then a language for which support has been revoked won't be disabled until `config.SUPPORTED_LOCALES` is updated on the next `securedrop-admin {sdconfig,install}` run.

This isn't a bug; it was simply out of scope of #6557.  Now #7442 motivates us to at least consider: **Do we *want* an unsupported language to be disabled immediately on upgrade of `securedrop-app-code`, not just on the next `securedrop-admin {sdconfig,install}`?**  If yes, then this diff handles this case.  If not, then we can just close this pull request.

## Testing

Under `make dev` with:

```patch
--- a/securedrop/bin/dev-config
+++ b/securedrop/bin/dev-config
@@ -26,7 +26,7 @@ with open('securedrop/config.py', 'w') as f:
     text = env.get_template("securedrop/config.py.example").render(ctx)
     text += '\n'
     supported_locales = subprocess.check_output(['make', '--quiet', 'supported-locales']).decode().strip()
-    text += f'SUPPORTED_LOCALES = {supported_locales}\n'
+    text += f'SUPPORTED_LOCALES = {"de_DE", "ro", "en_US"}\n'
     f.write(text)
 
 with open('securedrop/rq_config.py', 'w') as f:
--- a/securedrop/i18n.json
+++ b/securedrop/i18n.json
@@ -63,10 +63,6 @@
       "name": "Portuguese, Portugal",
       "desktop": "pt_PT"
     },
-    "ro": {
-      "name": "Romanian",
-      "desktop": "ro"
-    },
     "ru": {
       "name": "Russian",
       "desktop": "ru"
```

- [ ] Observe on startup:
    ```
    [2025-02-19 23:51:46,846] WARNING in i18n: Configured locales {Locale('ro')} are not in the set of usable locales {Locale('de', territory='DE'), Locale('en', territory='US')}
    ```
- [ ] Romanian is not offered in the locale switcher
- [ ] `/?l=ro` defaults to en_US

## Deployment

As above:

1. Do we want an unsupported language to be disabled immediately on upgrade of `securedrop-app-code`, not just on the next `securedrop-admin {sdconfig,install}`?
2. ~~Do we want an error to be logged and an OSSEC alert to be sent in this case?~~
    - Release QA should confirm that the warning (downgraded from error) does not trigger an OSSEC alert.

**NB.**  This is already covered at the implementation (as opposed to configuration) level by [`test_valid_but_unusable_locales()`](https://github.com/freedomofpress/securedrop/blob/35e4b2df2b5c3197c605884e078adc6116cffc69/securedrop/tests/test_i18n.py#L374).

Choose one of the following:

- I have opened a PR in the [docs repo](https://github.com/freedomofpress/securedrop-docs) for these changes, or will do so later

### DIFF
--- a/securedrop/i18n.py
+++ b/securedrop/i18n.py
@@ -159,7 +159,7 @@ def validate_locale_configuration(config: SecureDropConfig, babel: Babel) -> Set
 
     missing = configured - usable
     if missing:
-        babel.app.logger.error(
+        babel.app.logger.warning(
             f"Configured locales {missing} are not in the set of usable locales {usable}"
         )
 
--- a/securedrop/i18n.py
+++ b/securedrop/i18n.py
@@ -16,6 +16,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 #
 import collections
+import json
 from typing import DefaultDict, List, OrderedDict, Set
 
 from babel.core import (
@@ -28,6 +29,8 @@ from babel.core import (
 from flask import Flask, current_app, g, request, session
 from flask_babel import Babel
 from sdconfig import FALLBACK_LOCALE, SecureDropConfig
+
+I18N_CONF = "i18n.json"
 
 
 class RequestLocaleInfo:
@@ -137,12 +140,17 @@ def validate_locale_configuration(config: SecureDropConfig, babel: Babel) -> Set
     available = set(babel.list_translations())
     available.add(Locale.parse(FALLBACK_LOCALE))
 
+    # These locales are supported in the current version of securedrop-app-code.
+    with open(I18N_CONF) as i18n_conf_file:
+        i18n_conf = json.load(i18n_conf_file)
+    supported = parse_locale_set(i18n_conf["supported_locales"].keys())
+
     # These locales were configured via "securedrop-admin sdconfig", meaning
     # they were present on the Admin Workstation at "securedrop-admin" runtime.
     configured = parse_locale_set(config.SUPPORTED_LOCALES)
 
     # The intersection of these sets is the set of locales usable by Babel.
-    usable = available & configured
+    usable = available & configured & supported
 
     missing = configured - usable
     if missing:
--- a/securedrop/i18n.py
+++ b/securedrop/i18n.py
@@ -17,6 +17,7 @@
 #
 import collections
 import json
+from pathlib import Path
 from typing import DefaultDict, List, OrderedDict, Set
 
 from babel.core import (
@@ -30,7 +31,7 @@ from flask import Flask, current_app, g, request, session
 from flask_babel import Babel
 from sdconfig import FALLBACK_LOCALE, SecureDropConfig
 
-I18N_CONF = "i18n.json"
+I18N_CONF = Path(__file__).parent / "i18n.json"
 
 
 class RequestLocaleInfo:
@@ -141,9 +142,13 @@ def validate_locale_configuration(config: SecureDropConfig, babel: Babel) -> Set
     available.add(Locale.parse(FALLBACK_LOCALE))
 
     # These locales are supported in the current version of securedrop-app-code.
-    with open(I18N_CONF) as i18n_conf_file:
-        i18n_conf = json.load(i18n_conf_file)
-    supported = parse_locale_set(i18n_conf["supported_locales"].keys())
+    try:
+        with open(I18N_CONF) as i18n_conf_file:
+            i18n_conf = json.load(i18n_conf_file)
+        supported = parse_locale_set(i18n_conf["supported_locales"].keys())
+    # I18N_CONF may not be available under test.
+    except FileNotFoundError:
+        supported = available
 
     # These locales were configured via "securedrop-admin sdconfig", meaning
     # they were present on the Admin Workstation at "securedrop-admin" runtime.


cfm:
Noting for the record (from way back in 9c11950): `Configured locales [...] are not in the set of usable locales [...]` is probably a warning, not an error, but the decision then was that we *did* want this case to trigger an OSSEC alert.

cfm:
Clarified out of band with @zenmonkeykstop:  We can just close this pull request if we don't care about this disablement case—or defer to v2.13.

legoktm:
IMO the ideal behavior would be:
* any language we disable, is disabled on all SDs that have it explicitly enabled.
* if you're using `./securedrop-admin sdconfig/install`, an unsupported language is an error
* the existing server configuration is *not* an OSSEC alert, and if/when Hindi and Romanian are supported again, those servers will automatically re-enable the language without any manual interaction needed.

Would that be reasonable?

Also, I'm not sure where in your plan it was, but can we also disable hi/ro before the rc is cut so that behavior can be QA'd at the same time?

securedrop/i18n.py
```diff
@@ -29,6 +30,8 @@
 from flask_babel import Babel
 from sdconfig import FALLBACK_LOCALE, SecureDropConfig
 
+I18N_CONF = "i18n.json"
```
legoktm: Let's use an absolute path please, like `Path(__file__).parent / "i18n.json"`

securedrop/i18n.py
```diff
@@ -137,12 +140,20 @@ def validate_locale_configuration(config: SecureDropConfig, babel: Babel) -> Set
     available = set(babel.list_translations())
     available.add(Locale.parse(FALLBACK_LOCALE))
 
+    # These locales are supported in the current version of securedrop-app-code.
+    try:
+        with open(I18N_CONF) as i18n_conf_file:
+            i18n_conf = json.load(i18n_conf_file)
+        supported = parse_locale_set(i18n_conf["supported_locales"].keys())
+    except FileNotFoundError:
```
legoktm, line 150: In what case would the file not be found? If i18n.json is missing, that feels like a fatal error to me.

legoktm (dismissed):

nathandyer:
Noting from a docs perspective, I don't believe any update is necessary. We currently link out to [i18n.rst](https://github.com/freedomofpress/securedrop/blob/develop/securedrop/i18n.rst) rather than explicitly enumerating them in the documentation itself.

nathandyer:
I agree with @legoktm's ideal scenario above, and disabling a removed language on update rather than running `./securedrop-admin` feels like the most straightforward / discoverable path. By which I mean, if I were an admin and the language suddenly reverted to a different language, I would be able to see there was an upgrade and find the release notes.

If we do decide to go this route, I think adding a short blurb in our docs saying "in the event that you are using a language that is no longer supported, your SecureDrop will disable it during the upgrade process for that release" (or similar).

cfm:
Thanks both.  Replying to @legoktm in <https://github.com/freedomofpress/securedrop/pull/7443#issuecomment-2671879001>:

> IMO the ideal behavior would be:
> 
> * [x] **any language we disable, is disabled on all SDs that have it explicitly enabled.**
> * [x] if you're using `./securedrop-admin sdconfig/install`, an unsupported language is an error
> * [ ] the existing server configuration is _not_ an OSSEC alert
> * [x] and if/when Hindi and Romanian are supported again, those servers will automatically re-enable the language without any manual interaction needed.
> 
> Would that be reasonable?

Very reasonable, and I've done some slicing and checklisting here to show what we have currently (checked) and as of this pull request (checked and in bold).

Changing from error (with OSSEC alert) to warning (no OSSEC alert) should be a one-line change I'll tack on tomorrow along with responding to your review feedback.

securedrop/i18n.py
cfm: Thanks; fixed in ef98aa080.

securedrop/i18n.py
cfm: Clarified in ef98aa080.  Let me know if this doesn't address your concern.

cfm:
As of e50ead6f9:

> * [x] **any language we disable, is disabled on all SDs that have it explicitly enabled.**
> * [x] if you're using `./securedrop-admin sdconfig/install`, an unsupported language is an error
> * [x] **the existing server configuration is _not_ an OSSEC alert**
> * [x] and if/when Hindi and Romanian are supported again, those servers will automatically re-enable the language without any manual interaction needed.

zenmonkeykstop (approved):
Change and test plan LGTM - this should likely have an additional test case but that can happen via a subsequent PR.

securedrop/i18n.py
legoktm: Sorry, but why would it not be available under test? Is something deleting it?

legoktm: https://github.com/freedomofpress/securedrop/commit/2b1214c13269f03660b7a6714be61a46bbff3acb has a green check. I'll turn it into a PR, but we don't need to get it into 2.12.

If I had to guess, if it was failing earlier possibly because the path was relative, but now that you changed it to an absolute path, it works?

Issue #7433, opened 2025-01-31
noble staging failure in CI for test_all_packages_updated


legoktm:
## Description

`assert "0 upgraded, 0 newly installed, 0 to remove" in c.stdout` fails for:

```
NOTE: This is only a simulation!
      apt-get needs root privileges for real execution.
      Keep also in mind that locking is deactivated,
      so don't depend on the relevance to the real current situation!
Reading package lists...
Building dependency tree...
Reading state information...
Calculating upgrade...
The following upgrades have been deferred due to phasing:
  libnss-systemd libpam-systemd libsystemd-shared libsystemd0 libudev1 systemd
  systemd-dev systemd-resolved systemd-sysv systemd-timesyncd udev
The following packages will be upgraded:
  libdrm-common libdrm2
2 upgraded, 0 newly installed, 0 to remove and 11 not upgraded.
Inst libdrm-common [2.4.120-2build1] (2.4.122-1~ubuntu0.24.04.1 Ubuntu:24.04/noble-updates [all])
Inst libdrm2 [2.4.120-2build1] (2.4.122-1~ubuntu0.24.04.1 Ubuntu:24.04/noble-updates [amd64])
Conf libdrm-common (2.4.122-1~ubuntu0.24.04.1 Ubuntu:24.04/noble-updates [all])
Conf libdrm2 (2.4.122-1~ubuntu0.24.04.1 Ubuntu:24.04/noble-updates [amd64])
```

https://github.com/freedomofpress/securedrop/actions/runs/13065618335/job/36457392613

I have not looked to see why libdrm wasn't already upgraded.

legoktm:
Based on reading https://bugs.launchpad.net/ubuntu/+source/llvm-toolchain-19/+bug/2083020, libdrm was upgraded as a dependency of mesa. mesa is being phased, but I don't see why that would affect the installation of libdrm...

legoktm:
We can confirm libdrm is not being phased by looking at:

```
$ apt show libgbm1
Package: libgbm1
Version: 24.2.8-1ubuntu1~24.04.1
Priority: optional
Section: libs
Source: mesa
Origin: Ubuntu
Phased-Update-Percentage: 40
...
```

But `apt show libdrm2` contains no "Phased-Update-Percentage" line. So it really didn't upgrade for some reason.

legoktm:
[...]

PR #7451, opened 2025-02-27
fix: allow Apache to read `i18n.json`

### LINKED ISSUE #7450: `i18n.json` is used in app startup but not included in apache2 apparmor config.
`i18n.json` is now being used during app setup to determine which locales are valid. It needs to be included in Apparmor for apache2 (with read perms) to support this in production and staging environments. Otherwise app startup fails and interfaces return a 500 error.

## Expected behavior
App setup completes in prod

## Actual behavior
App setup fails, with file access errors for the aforementioned file.

cfm:
## Status

Ready for review

## Description of Changes

Fixes #7450 re: #7443.

## Testing

- [ ] Staging CI passes except for `test_all_packages_updated` per #7433.

Indeed, and I will follow up in #7450 on why we didn't catch this in #7443....

### DIFF
--- a/securedrop/debian/app-code/etc/apparmor.d/usr.sbin.apache2
+++ b/securedrop/debian/app-code/etc/apparmor.d/usr.sbin.apache2
@@ -173,6 +173,7 @@
   /var/www/securedrop/dictionaries/nouns.txt r,
   /var/www/securedrop/encryption.py r,
   /var/www/securedrop/execution.py r,
+  /var/www/securedrop/i18n.json r,
   /var/www/securedrop/i18n.py r,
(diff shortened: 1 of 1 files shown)


zenmonkeykstop (approved):
LGTM based on visual review. Will merge when CI is passing (looks like staging job is flaking)

cfm:
@zenmonkeykstop, I've updated the test plan to reflect the fact that `test_all_packages_updated` is expected to fail according to #7433....

zenmonkeykstop:
> @zenmonkeykstop, I've updated the test plan to reflect the fact that `test_all_packages_updated` is expected to fail according to #7433....

Good stuff -merging in that case.

PR #7463, opened 2025-03-05
Add .well-known file to set relationship with funding.json on securedrop.org

zenmonkeykstop:
## Status

Ready for review 

## Description of Changes

Adds a .well-known file to the repo to correlate it with the information to be posted at https://securedrop.org/documents/35/funding.json

## Testing

- [ ] CI is passing
- [ ] File format follows spec at https://floss.fund/funding-manifest/

## Deployment

This file lives in the repo only and is not included in packaging or deployment.

### DIFF
--- a/.well-known/funding-manifest-urls
+++ b/.well-known/funding-manifest-urls
@@ -1,0 +1,1 @@
+https://securedrop.org/documents/35/funding.json


legoktm (approved):
Seems straightforward enough, I wonder if their validator will be happy with the fact that the URL redirects to a different subdomain (media.securedrop.org). Easiest way is just to try it and see.

PR #7466, opened 2025-03-06
Remove systemd-resolved, explicitly install systemd-hwe-hwdb on noble

### LINKED ISSUE #7464: Instance upgraded from focal to noble is missing systemd-resolved
## Description

Doing an in-place upgrade to noble results in the system missing systemd-resolved. A fresh noble install will still have it.

## Steps to Reproduce

1. Set up focal 2.12.0-rc.2
2. Do a semiautomated / fully automated upgrade
3. On a server, run `systemctl status systemd-resolved` - it'll error
4. Try running `./securedrop-admin install` - it will error on the step about stopping systemd-resolved.

## Expected Behavior

Either:
1. systemd-resolved is present on upgraded systems
2. Or, systemd-resolved is removed from all noble systems

And, `./securedrop-admin install` should work post-upgrade.

## Next steps

1. diff the output of `dpkg -l` from both a fresh noble install and an upgrade and see if anything else is missing (ideally check both app and mon)
2. figure out what to do with systemd-resolved. Not sure why it's getting removed during upgrade, a missing dependency maybe? We could fix that, or just say that we don't actually need it since we stop it and uninstall it on fresh installs too. The former is probably safer.

legoktm:
## Status

Ready for review

## Description of Changes

These two packages are installed on fresh systems, but not on upgrades because they were split out of the systemd package. ~Set the dependency ourselves to make sure it's always pulled in.~ [see discussion below]

In the future once all SecureDrops are on noble, we can uninstall systemd-resolved entirely instead of merely stopping it.

Fixes #7464.

## Testing

* [x] CI passes (verifies conditional dependency is correctly applied)
* [x] staging CI passes, verifies fresh install is fine
* [x] do a focal to noble upgrade with these packages, verify systemd-resolved is not installed post-upgrade, `/etc/resolv.conf` points to your configured DNS, and e.g. `curl https://securedrop.org` works since DNS is working.
* [x] verify that `./securedrop-admin install` works post-upgrade.

## Deployment

Any special considerations for deployment? upgrade is more important.

### DIFF
--- a/builder/tests/test_securedrop_deb_package.py
+++ b/builder/tests/test_securedrop_deb_package.py
@@ -130,3 +130,22 @@ def test_apparmor_conditional():
 
     print(info)
     assert found, "Depends: line wasn't found"
+
+
+def test_systemd_conditional():
+    try:
+        path = [pkg for pkg in DEB_PATHS if pkg.name.startswith("securedrop-config")][0]
+    except IndexError:
+        raise RuntimeError("Unable to find securedrop-config package in build/ folder")
+    info = subprocess.check_output(["dpkg", "--info", path]).decode()
+    found = False
+    for line in info.splitlines():
+        if line.startswith(" Depends:"):
+            found = True
+            if UBUNTU_VERSION == "focal":
+                assert "systemd-hwe-hwdb" not in line, "focal has no systemd-hwe-hwdb dependency"
+            else:
+                assert "systemd-hwe-hwdb" in line, "noble has systemd-hwe-hwdb dependency"
+
+    print(info)
+    assert found, "Depends: line wasn't found"
--- a/install_files/ansible-base/roles/common/tasks/harden_dns.yml
+++ b/install_files/ansible-base/roles/common/tasks/harden_dns.yml
@@ -7,11 +7,24 @@
     - dns
     - hardening
 
-- name: Disable systemd-resolved
+- name: Disable systemd-resolved (focal)
   systemd:
     name: systemd-resolved
     state: stopped
     enabled: no
+  when: ansible_distribution_release == "focal"
   tags:
+    - dns
+    - hardening
+
+- name: Uninstall systemd-resolved (noble)
+  apt:
+    name:
+      - systemd-resolved
+    state: absent
+    purge: yes
+  when: ansible_distribution_release != "focal"
+  tags:
+    - apt
     - dns
     - hardening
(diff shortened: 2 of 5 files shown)


legoktm:
I performed a focal -> noble upgrade with these packages and it worked fine, confirmed that the two new systemd packages were installed but that systemd-resolved was still disabled.

Then I re-ran `./securedrop-admin --force install` against the now-noble hosts and it completed successfully.

legoktm:
Unfortunately this correctly installs systemd-resolved, but then that gets enabled and is actively running post-upgrade. So we need a postinst step to re-disable it. Will work on that tomorrow.

legoktm:
This is going even more poorly; once we install systemd-resolved, it blows away /etc/resolv.conf, so when we turn it off, there's no DNS. Since DNS is configured by the admin, it's not something we can directly recreate (we could do the backup/restore strategy, like iptables).

So I'm going to switch gears and just have us not install systemd-resolved. Unfortunately this invalidates our fresh install testing but I think it's going to work out better.

legoktm:
It worked!! I can confirm that:
* doing a focal -> noble migration leaves you with a system without systemd-resolved, and still working DNS (status quo before this PR). systemd-hwe-hwdb is pulled in.
* after that, running `./securedrop-admin --force install` (using this PR), will not fail, and the line to remove systemd-resolved is marked as "ok" and not "changed".
* And even after that DNS still works.

I didn't manually test a noble fresh install because staging CI does that for us, and it worked as expected:
```
2025-03-07T16:57:17.6852219Z TASK [common : Disable systemd-resolved (focal)] *******************************
2025-03-07T16:57:17.6997915Z skipping: [app-staging]
2025-03-07T16:57:17.7069178Z skipping: [mon-staging]
2025-03-07T16:57:17.7126125Z 
2025-03-07T16:57:17.7126715Z TASK [common : Uninstall systemd-resolved (noble)] *****************************
2025-03-07T16:57:22.0994976Z changed: [app-staging]
2025-03-07T16:57:22.1548691Z changed: [mon-staging]
```

rocodes:
Per release checkin meeting today, I will visually review, relying on CI + @legoktm 's prior testing. We will incorporate further testing as part of regular rc QA test plan.

rocodes (approved):
I've caught up with the context in https://github.com/freedomofpress/securedrop/issues/7464, the initial implementation plan, and the subsequent decision to remove systemd-resolved instead of installing everywhere, and I've reviewed the changes here- at first the systemd depends in the control file was slightly confusing given that context, but I see that it is conditionally pulling in the systemd-hwe-hwdb dependency for noble. 

Thanks for all the documentation and for resolving this so quickly @legoktm - LGTM. Will look out for the backport.

PR #7484, opened 2025-03-26
Make noble-migration playbook smarter for SSH-over-Tor handling

### LINKED ISSUE #7483: semi-auto noble upgrade playbook fails with ssh-over-tor
## Description

@eloquence reported that during his semi-auto upgrade of his long-running instance, the playbook failed at the end of the upgrade:

* https://gist.github.com/eloquence/8fb21b46f42628c1384bf9241773438a
* https://gist.github.com/eloquence/1e4604e3741ac2648f01e9ee2ad36a00

He shared the full journalctl logs with me, in looking at it, I see that he reconnected before the second reboot. So it interpreted that reconnection as the second reboot, and then when the second reboot actually happened, it wasn't expecting to get disconnected and failed. Meanwhile the upgrade proceeded happily in the background and finished successfully on its own (yay).

The log doesn't show the disconnection, so I'm not exactly sure what was the cause. Right before he reconnected is when sshd restarted (because the new package was installed), and ~20s earlier was when Tor restarted. Normally sshd is supposed to keep connections alive when it restarts, but I don't think that works when Tor restarts? So maybe it's ssh-over-tor causing the issue and we just need to expect a third disconnect/reconnect sequence in that case?

----

Then I was able to reproduce this on my 2.12.0 fresh install. I lost my ssh connection when tor was restarted by the upgrade process, and then it reconnected before the expected reboot. It's racy though, I only reconnected like, 30 seconds before the reboot, so I suspect it succeeded in other testing because it took a bit longer for tor to come back up + ansible only checks every 10 seconds, that it had already rebooted by then.

legoktm:
## Status

Ready for review 

## Description of Changes

Previously, the ansible playbook didn't actually understand the sequencing - it was just a dumb, hardcoded sequence of two reboots.

We recently learned that with SSH-over-Tor there's a race condition, you might end up with 3 disconnections instead of 2 because you'll lose SSH when Tor is upgraded. In some cases if things are slow enough, you might only reconnect after the second reboot. So we need to be able to dynamically handle 2 or 3 disconnections.

The playbook is now aware of the most recently completed state, as well as the position in the overall run.

In human-readable pseudocode, the logic roughly is:

* Get current finished state
* If finished_state == None: (i.e. not started)
  * Kick off the migration.
  * Wait for a reboot (should switch to PendingUpdates here)
  * Get current finished state
* If Reboot > finished_state >= PendingUpdates: ("Phase 2")
* Wait for a reboot (should switch to Reboot here, unless SSH-over-Tor disconnects first)
  * Get current finished state
* If SSH-over-Tor and Reboot > finished_state >= PendingUpdates: ("Phase 2.5")
  * Wait for a reboot (should switch to Reboot here)
  * Get current finished state
* If Done > finished_state >= Reboot: ("Phase 3")
  * Wait for finished state to be Done
  * Set current finished state to Done
* If finished_state != Done:
  * Fail :(

Because the logic is fully aware of the migration state, if you e.g. start the playbook, the upgrade begins, you Ctrl+C, and then restart the playbook, it'll just work because it'll drop you into the correct phase's logic instead of starting from the beginning.

Fixes #7483.

## Testing

* [ ] visual review
* [ ] CI passes
* [x] set up 2.12.0 on focal (VMs or prod is fine), check out this PR on the AW and:
   * [x] enable SSH-over-Tor, `./securedrop-admin noble_migration` completes successfully
   * [x] without SSH-over-Tor, `./securedrop-admin noble_migration` completes successfully
   * [ ] Start `./securedrop-admin noble_migration`, wait until you see "The upgrade is in progress; it may take up to 30 minutes", then Ctrl+C the playbook and restart it. It should still complete successfully even though it's one less reboot. (SSH-over-Tor doesn't matter for this scenario)

## Deployment

Any special considerations for deployment? Will be backported into a 2.12.1 AW-only release. I added a comment to upgrade.rs, which is not AW-only but it's not actually a change.

## Checklist

### DIFF
--- a/install_files/ansible-base/roles/noble-migration/tasks/main.yml
+++ b/install_files/ansible-base/roles/noble-migration/tasks/main.yml
@@ -81,7 +81,7 @@
         connect_timeout: 20
         sleep: 5
         delay: 10
-        timeout: 300
+        timeout: 600
 
     - name: Recheck host's migration state
       ansible.builtin.slurp:
@@ -132,7 +132,7 @@
         connect_timeout: 20
         sleep: 5
         delay: 10
-        timeout: 300
+        timeout: 600
 
     - name: Recheck host's migration state
       ansible.builtin.slurp:
@@ -174,7 +174,7 @@
         connect_timeout: 20
         sleep: 5
         delay: 10
-        timeout: 300
+        timeout: 600
 
     - name: Recheck host's migration state
       ansible.builtin.slurp:
--- a/install_files/ansible-base/roles/noble-migration/tasks/main.yml
+++ b/install_files/ansible-base/roles/noble-migration/tasks/main.yml
@@ -1,17 +1,50 @@
 ---
 
+- name: Initialize
+  set_fact:
+    # Keep in sync with upgrade.rs
+    stages:
+      - 'None'
+      - 'PendingUpdates'
+      - 'MigrationCheck'
+      - 'DisableApache2'
+      - 'Backup'
+      - 'BackupIptables'
+      - 'Marker'
+      - 'SuspendOSSEC'
+      - 'ChangeAptSources'
+      - 'AptGetUpdate'
+      - 'AptGetFullUpgrade'
+      - 'AptGetAutoremove'
+      - 'RestoreIptables'
+      - 'ReenableUnattendedUpdates'
+      - 'ReenableOSSEC'
+      - 'Reboot'
+      - 'SwitchUbuntuSources'
+      - 'IntegrityCheck'
+      - 'ReenableApache2'
+      - 'RemoveBackup'
+      - 'Done'
+
 - name: Check host's migration state
   ansible.builtin.slurp:
     src: /etc/securedrop-noble-migration-state.json
   register: migration_json
-  ignore_errors: yes
 
-- name: Skip migration if already done
+# Note: the variable finished_state is only for debugging and human-readable output.
+# The finished_state_index is what must be used for logic in `when` blocks.
+
+- name: Extract current state
   set_fact:
-    already_finished: "{{ not migration_json.failed and (migration_json.content | b64decode | from_json)['finished'] == 'Done' }}"
+    # slurp base64-encodes our file
+    finished_state: "{{ (migration_json.content | b64decode | from_json)['finished'] }}"
 
-- name: Perform migration
-  when: not already_finished
+- name: Extract current state (2)
+  set_fact:
+    finished_state_index: "{{ stages.index(finished_state) }}"
+
+- name: Kick off migration
+  when: finished_state_index|int == 0
   block:
     - name: Instruct upgrade to begin
       ansible.builtin.copy:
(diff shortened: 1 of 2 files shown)


legoktm:
Posting this because it's the end of the day, I only personally tested through the enable SSH-over-Tor plan. Will get through the other two tomorrow.

install_files/ansible-base/roles/noble-migration/tasks/main.yml
```diff
@@ -1,17 +1,47 @@
...
+      - 'IntegrityCheck'
+      - 'ReenableApache2'
+      - 'RemoveBackup'
+      - 'Done'
+
 - name: Check host's migration state
   ansible.builtin.slurp:
     src: /etc/securedrop-noble-migration-state.json
   register: migration_json
-  ignore_errors: yes
```
legoktm, line 7: This was a mistake and shouldn't have been there to start with - the slurp has to succeed.

install_files/ansible-base/roles/noble-migration/tasks/main.yml
```diff
@@ -50,6 +80,29 @@
...
+        finished_state_index: "{{ stages.index(finished_state) }}"
+
+    - debug:
+        var: finished_state
+    - debug:
+        var: finished_state_index
+
+- name: Phase 2 of migration
+  # After PendingUpdates (index 1) but before Reboot (index 15)
+  when: finished_state_index|int >= 1 and finished_state_index|int < 15
```
legoktm, line 107: The `|int` filter is necessary because technically the index variable is a string, so ansible won't let you run `>=` with a integer.

noble-migration/src/bin/upgrade.rs
```diff
@@ -30,6 +30,8 @@ const EXTRA_APT_SOURCE: &str = "EXTRA_APT_SOURCE";
 ///
 /// Each stage needs to be idempotent so that it can be run multiple times
 /// in case it errors/crashes.
+///
+/// Keep this in sync with the version in noble-migration/tasks/main.yml
```
legoktm, line 34: This is technically touching outside of the AW-only release, but it's only a comment so I figured it's okay.

legoktm:
Kev found a bug in the playbook that I had introduced in what I had thought was a cosmetic refactor, when I switched the first phase's block's `when` condition from `finished_state_index == 0` to `finished_state == "None"`. The latter is more human-readable, so I went with it, assuming the two were identical.

However, the [block documentation](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_blocks.html) explicitly says:
> For example, a _when_ statement is applied to the tasks within a block, not to the block itself.

So the final step in the block which set finished_state_index would get skipped because the `finished_state == "None"` condition no longer applies. This also meant that any debug statements added after setting `finished_state` would get skipped too.

I was able to track this down by using the [ansible debugger](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_debugger.html), which does get invoked if a task is skipped, and it'll tell you why it was skipped.

The simple fix was to go back to checking `finished_state_index == 0`. I also added comments to the end of each block explaining why that had to be the last statement in the block and nothing could come after it.

zenmonkeykstop:
:+1: reviewing now, expect delay coz multiple installs :).

legoktm:
Did an upgrade with SSH-over-LAN, nothing interesting to report, it went smoothly: https://gist.github.com/legoktm/79213fa1deca9ef39f19eaba5a5d1476

zenmonkeykstop (approved):
Test scenarios check out, all is well!

PR #7497, opened 2025-04-10
Verify all ethernet interfaces are UP before noble migration

### LINKED ISSUE #7495: systemd-networkd-wait-online is stricter in noble
## Description

The systemd-networkd-wait-online service is now stricter in noble, it requires all configured interfaces to be up and working.

This means that even if your system is online and functional, if there is an extra interface that's configured but non-operational, the unit will fail, which blocks the integrity checks part of the noble migration.

SD instances should only ever have one ethernet device configured, because it goes to the hardware firewall.

We should have a noble pre-flight check that parses the netplan YAML config and verifies that only one such device is configured.

legoktm:
## Status

Ready for review

## Description of Changes

In noble, systemd-networkd-wait-online is stricter and requires that all ethernet interfaces which are configured are online. Mostly emulate that check on the focal side by looking at the active interfaces and verifying they're all UP.

Fixes #7495.

## Testing

* [x] staging CI passes on focal, specifically the `test_migration_check` test (other tests might be flaky)
* [ ] 

## Deployment

## Checklist

### DIFF
--- a/molecule/testinfra/common/test_release_upgrades.py
+++ b/molecule/testinfra/common/test_release_upgrades.py
@@ -56,7 +56,7 @@ def test_migration_check(host):
         contents = json.loads(cmd.stdout)
         print(contents)
         # The script did not error out
-        if "error" in contents:
+        if "error" in contents or not all(contents.values()):
             # Run the script manually to get the error message
             cmd = host.run("securedrop-noble-migration-check")
             print(cmd.stdout)
--- a/noble-migration/src/lib.rs
+++ b/noble-migration/src/lib.rs
@@ -1,6 +1,6 @@
 //! Common code for the noble-migration that is used by check and upgrade
 use anyhow::{bail, Context, Result};
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use std::{
     fs,
     path::Path,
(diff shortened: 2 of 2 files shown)


legoktm:
I'm not sure how to test the failure case though, of having an extra ethernet interface that is configured but not plugged in. I naively tried adding a `eth0: dhcp4: true` to my netplan config and it didn't change any of the `ip` output, presumably because that's not a real device on my setup.

legoktm:
> I'm not sure how to test the failure case though, of having an extra ethernet interface that is configured but not plugged in.

I just connected over console and unplugged the ethernet cord and checked `ip -json -brief addr show`. The output was `{"ifname":"enp89s0","operstate":"DOWN", "addr_info":[]}`, which I've now updated in the test case.

zenmonkeykstop:
I can't remember if we're setting the boot parameter for old-skool interface names or not, might be worth checking on a production machine if you're matching by name.

legoktm:
> I can't remember if we're setting the boot parameter for old-skool interface names or not, might be worth checking on a production machine if you're matching by name.

The only interface name we match is `lo` for loopback, and the test cases are from my prod setup (except for the one labeled staging).

zenmonkeykstop (approved):
Seeing same staging-noble flake on apt output, good to go otherwise

PR #7502, opened 2025-04-16
Don't begin noble migration on noble

### LINKED ISSUE #7501: fully automated noble migration starts on noble
## Description

The noble migration should obviously not run if the system is already noble, but if fully automated upgrades are enabled 

## Steps to Reproduce

* Fresh install noble on 2.12.1
* Edit your bucket to be group 1
* Upgrade to 2.12.2-rc1

## Expected Behavior

The upgrade is not triggered

## Actual Behavior

The upgrade starts, finishes the "PendingUpdates" step, reboots, and then crashes at the MigrationCheck stage, which notices that it isn't focal.

## Comments

Should be straightforward to add another check, earlier. PR incoming.

legoktm:
## Status

Ready for review

## Description of Changes

Now that we're enabling fully automated upgrades, we need to explicitly check that we're not already noble before starting the migration.

The existing check in the MigrationCheck stage is too late, since by then we'll have already disabled unattended-upgrades.

Fixes #7501.

## Testing

* [ ] visual review of logic
* [x] build packages, install on a fresh install noble system and see that it does not begin the upgrade

## Deployment

### DIFF
--- a/noble-migration/src/bin/upgrade.rs
+++ b/noble-migration/src/bin/upgrade.rs
@@ -299,6 +299,7 @@ fn pending_updates(state: &mut State) -> Result<()> {
 /// we begin
 fn migration_check() -> Result<()> {
     info!("Checking pre-migration steps...");
+    // This should've been caught in should_upgrade() but let's double check
     if noble_migration::os_codename()? != "focal" {
         bail!("not a focal system");
     }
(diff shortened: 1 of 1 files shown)


legoktm:
I installed this on a fresh-installed noble system. Previously it would output "Auto-upgrades are enabled (app), but our bucket hasn't been enabled yet" or, actually start the migration. Even after I switched the bucket to 1, it now outputs "Already on a noble system; no need to upgrade."

zenmonkeykstop (approved):
LGTM based on visual review, will test against prod noble system in nxt release candidate

PR #7506, opened 2025-04-21
update `setuptools` >= 78.1.1 for Safety 76752

cfm:
## Status

Work in progress

## Description of Changes

Updates `setuptools` >= 78.1.1 for Safety 76752.  This required updating `packaging` and, given how fundamental these dependencies are, some manual tweaking instead of just running `make update-pip-requirements`.

## Testing

- [ ] CI passes.

## Deployment

No deployment considerations.

se one of the following:

Both `packaging` and `setuptools` are published by the Python Packaging Authority, so they don't require diff review, per https://github.com/freedomofpress/securedrop-dev-docs/blob/4a42bbb4eb4007edcccb317da5f19998dc3175e0/docs/dependency_updates.rst?plain=1#L52-L53.

- I would like someone else to do the diff review
- I am silencing an alert related to a production dependency, because (please explain below):


cfm:
It turns out that [`setuptools` has required Python >= 3.9 since v75.3.2](https://github.com/pypa/setuptools/commit/750a1891ec4a1c0602050e3463e9593a8c13aa14); Ubuntu Focal only gives us Python 3.8.  I'll flag this at tomorrow's team meeting.

cfm:
The team (and then @legoktm and I) discussed this today and agreed that we need to have a way to update dependencies between Python 3.8 on Ubuntu Focal and Python 3.12 on Ubuntu Noble whether or not this `setuptools` update is worth the trouble.  That's now set up in 5430b4456db0a9b2ca508f707da77fa5cc280d90 (compilation step) and 83bf78807cf0fe391ce08dbe513232db56a19971 (installation step).  If we like this, I'll replace 734641fd064807d8666032d3ab79a837bc2dec87 with a commit that fully handles the requirements files I've split from `securedrop/requirements/python3/*.in` → `securedrop/requirements/python3/{focal,noble}/*.txt`.

legoktm:
> and [83bf788](https://github.com/freedomofpress/securedrop/commit/83bf78807cf0fe391ce08dbe513232db56a19971) (installation step). If we like this, I'll replace [734641f](https://github.com/freedomofpress/securedrop/commit/734641fd064807d8666032d3ab79a837bc2dec87) with a commit that fully handles the requirements files I've split from `securedrop/requirements/python3/*.in` → `securedrop/requirements/python3/{focal,noble}/*.txt`.

Admittedly I had forgotten that doing this would require duplicating the requirements.txt files but it looks solid. I assume the symlinks are just temporary and once they're removed Git will show the focal files as being renamed and not deleted/created?

Also if you want, you could drop the "python3" part of the path, so it's just `securedrop/requirements/{focal,noble}/`?

securedrop/bin/update-requirements
```diff
@@ -6,7 +6,7 @@ set -euxo pipefail
 source /etc/os-release
 
 # https://peps.python.org/pep-0508/#environment-markers
-PYTHON_VERSION="`python3 -c 'import platform; print(\".\".join(platform.python_version_tuple()[:2]))'`"
+PYTHON_VERSION="$(python3 -c 'import platform; print(".".join(platform.python_version_tuple()[:2]))')"
```
cfm: Meant to squash this into 5430b4456db0a9b2ca508f707da77fa5cc280d90.

cfm:
Yes, exactly, @legoktm.  I'll finish that (in place of 734641fd064807d8666032d3ab79a837bc2dec87) early next week.

cfm:
As far as I can tell, `admin-tests` is failing here because of the `2.12.x` tag churn in this repository, unrelated to these changes.  So this should be ready for review.

securedrop/bin/update-requirements
```diff
@@ -2,23 +2,29 @@
...
+COMMON_ARGS="--generate-hashes --python-version ${PYTHON_VERSION} --python-platform x86_64-unknown-linux-gnu --custom-compile-command ./securedrop/bin/update-requirements"
 
 cd "$(git rev-parse --show-toplevel)/securedrop"
 
 uv pip compile ${COMMON_ARGS} \
-    --output-file requirements/python3/develop-requirements.txt \
-    requirements/python3/translation-requirements.in \
-    requirements/python3/develop-requirements.in
+    --output-file requirements/${VERSION_CODENAME}/develop-requirements.txt \
+    requirements/translation-requirements.in \
```
zenmonkeykstop, line 17: `make update-python3-requirements` fails, as .in files are still in `.../python3/`

zenmonkeykstop (dismissed):
The `update-python3-requirements` makefile target isn't working due to path changes. Others seem fine.

securedrop/bin/update-requirements
cfm: Oops; missed that in doing the refactoring @legoktm suggested.  Fixed in ff0b8acaea8ad2ea69dfab170ef3e2c02bd9c55f; diff confirms that (apart from a side effect of rebasing from `develop`) just paths changed even after rerunning `make update-python3-requirements`:

https://github.com/freedomofpress/securedrop/compare/c11b76d5f922ea61ced8f22fca309f71e2bbe992..ff0b8acaea8ad2ea69dfab170ef3e2c02bd9c55f

zenmonkeykstop (approved):
CI and makefile targets working, good to go.

securedrop/requirements/noble/translation-requirements.txt
```diff
@@ -0,0 +1,15 @@
...
+babel==2.12.1 \
+    --hash=sha256:b4246fb7677d3b98f501a39d43396d3cafdc8eadb045f4a31be01863f655c610 \
+    --hash=sha256:cc2d99999cd01d44420ae725a21c9e3711b3aadc7976d6147f622d8581963455
+    # via
+    #   -r requirements/translation-requirements.in
+    #   babelgladeextractor
+babelgladeextractor==0.7.0 \
+    --hash=sha256:bcf805e28b4bb18c8b6909a65a7cf5c7c2bcbf4ae50b164878c9682d22271798
+    # via -r requirements/translation-requirements.in
+setuptools==70.3.0 \
```
deeplow, line 12: Looks like the setuptools version was not bumped here. Was this intentional @cfm? This is the only `/noble` dependency I'm seeing as missing a version bump.

cfm: Good catch.  Let me see if I can consolidate these specifications while I'm at it....

cfm: Proposed in #7551.

PR #7512, opened 2025-04-22
chown packages to host's user after building

legoktm:
## Status

Ready for review 

## Description

The intention was always that the host user is mapped as root in the container; but differences in the various container runtimes and versions we support make that difficult.
    
For now, let's chown the packages after building so they'll always have the correct ownership.

## Testing

* [ ] Run `make build-debs`, verify newly built debs are owned by your host user

## Deployment


zenmonkeykstop (approved):
LGTM, ownership as expected.

PR #7515, opened 2025-04-24
noble-migration: Run migration check first

### LINKED ISSUE #7514: noble migration check should run before we disable unattended-updates
## Description

We currently run pending_updates() and then run migration_check(), and the initial thinking was that we should update everything to latest before checking stuff, which is reasonable, except at some point I refactored it to also disable unattended-updates during pending_updates() to take advantage of the fact that pending_updates() does a reboot.

The only problem is that we expect the migration_check() to somewhat be fallible. But we shouldn't end up in a state where it fails and then now updates are fully stopped, needing manual intervention.

## Proposed solution

I think we can simply reorder the stages, so migration_check() is the very first thing that runs, and only once we're ready to really to the upgrade, then we take care of pending_updates().

It's theoretically possible that a pending_update brings in something that causes a migration_check to fail, but at this point I think it's pretty unlikely given that the fully automated migration is only going to be kicked off immediately after pending updates are applied.

legoktm:
## Status

Ready for review

## Description of Changes

We currently run `pending_updates()` and then run `migration_check()`, and the initial thinking was that we should update everything to latest before checking stuff, which is reasonable, except at some point I refactored it to also disable unattended-updates during `pending_updates()` to take advantage of the fact that `pending_updates()` does a reboot.

The only problem is that we expect the `migration_check()` to somewhat be fallible. But we shouldn't end up in a state where it fails and then now updates are fully stopped, needing manual intervention.

So the fix is to simply reorder them, which is relatively trivial, though the manual index in the semiautomated playbook also needs updating.

It's theoretically possible that `pending_updates()` brings in something that would have caused `migration_check()` to fail, but at this point I think it's pretty unlikely given that the fully automated migration is only going to be kicked off immediately after pending updates are applied.

Fixes #7514.
## Testing

Part 1:
* [ ] Set up a focal system, and then on the app server:
* [x] Cause the migration check to fail by e.g. `sudo apt install ufw -y`
   * note: there's an hourly systemd timer that removes this, so you probably want to temporarily stop that timer `systemctl stop securedrop-remove-packages.timer` 
* [x] Move your app server into bucket 1, and then `sudo systemctl start securedrop-noble-migration-upgrade` (or wait for the timer)
* [x] Observe that 1) the state file still shows you in "None" and 2) the systemd unit unattended-upgrades is not masked
* [x] Then fix the migration check thing (uninstall ufw). Then `sudo systemctl start securedrop-noble-migration-upgrade` (or wait for the timer) and the upgrade should proceed and complete like normal.

Part 2:
* [x] verify semiautomated upgrade still works (i.e. the index changes are correct)

## Deployment

### DIFF
--- a/install_files/ansible-base/roles/noble-migration/tasks/main.yml
+++ b/install_files/ansible-base/roles/noble-migration/tasks/main.yml
@@ -5,8 +5,8 @@
     # Keep in sync with upgrade.rs
     stages:
       - 'None'
-      - 'PendingUpdates'
       - 'MigrationCheck'
+      - 'PendingUpdates'
       - 'DisableApache2'
       - 'Backup'
       - 'BackupIptables'
@@ -103,8 +103,8 @@
     # the block is in a `when` that we just made false by updating `finished_state_index`.
 
 - name: Phase 2 of migration
-  # After PendingUpdates (index 1) but before Reboot (index 15)
-  when: finished_state_index|int >= 1 and finished_state_index|int < 15
+  # After PendingUpdates (index 2) but before Reboot (index 15)
+  when: finished_state_index|int >= 2 and finished_state_index|int < 15
   block:
     # Start the systemd service manually to avoid waiting for the timer to pick it up
     - name: Resume upgrade systemd service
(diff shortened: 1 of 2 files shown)


zenmonkeykstop (approved):
Test plan checks out, changes lgtm

PR #7522, opened 2025-04-29
fix(`setup_i18n`): no need to guard with `@ignore_static`

### LINKED ISSUE #7504: 404 errors on `/static/` result in `UndefinedError`
## Description

Requesting a page under the `/static` route that does not exist results in an `UndefinedError` (development environment; not tested yet in staging/prod).

## Steps to Reproduce

1. Run `make dev-tor`
2. Visit onion service for Source Interface in Tor Browser
3. Attempt to access `<path to source interface>/static/foo`

## Expected Behavior

404 error

## Actual Behavior

Application throws `UndefinedError`

Image

cfm:
## Status

Ready for review

## Description of Changes

Trivially fixes #7504, and only slightly less trivially adds a regression test for it.

## Testing

- [x] `test_static_asset_not_found()` passes in CI.
- [x] If you're bored, `curl -I http://localhost:8080/static/foo` against `make dev` returns HTTP `404` rather than `500`.

### DIFF
--- a/securedrop/source_app/__init__.py
+++ b/securedrop/source_app/__init__.py
@@ -49,7 +49,6 @@ def create_app(config: SecureDropConfig) -> Flask:
     i18n.configure(config, app)
 
     @app.before_request
-    @ignore_static
     def setup_i18n() -> None:
         """Store i18n-related values in Flask's special g object"""
         i18n.set_locale(config)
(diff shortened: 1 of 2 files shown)


zenmonkeykstop (approved):
Change LGTM.

PR #7539, opened 2025-05-08
ignore Safety 76752 in setuptools

deeplow:
## Status

Ready for review

## Description of Changes

Suppresses vulnerability warning.

Setuptools has been bumped to the fixed version [1], but that was not possible on Ubuntu Focal due to setuptools 78.1.1 requiring Python >=3.9 and Ubuntu Focal only providing Python 3.8. In fact ealier versions of setuptools required this newer Python version but only this release justified bumping it.

Therefore, test-requirements.txt was split [2] into Ubuntu Focal and Noble and the setuptools dependency will have to be kept back on the former. This is minor given that this is a build-time dependency only and there is an ongoing migration of instances to Ubunut Noble.

The end conclusion is that this vulnerability will be silenced for the entire repository, since it has already been patched where it could be.

[1]: https://github.com/freedomofpress/securedrop/pull/7506
[2]: https://github.com/freedomofpress/securedrop/pull/7506#issuecomment-2825889614

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -194,6 +194,7 @@ safety:  ## Run `safety check` to check python dependencies for vulnerabilities.
 		--ignore 74221 \
 		--ignore 74261 \
 		--ignore 74735 \
+		--ignore 76752 \
 		--full-report -r $$req_file \
 		&& echo -e '\n' \
 		|| exit 1; \


legoktm (dismissed):
Thanks. Note that you'll also need to suppress it in https://github.com/freedomofpress/fpf-misc-resources/blob/main/projectfiles/securedrop.json (yes, it's bad that it's duplicated :/)

legoktm (commented):
Actually...I don't think this is the right file. @cfm, shouldn't the suppression go in the Makefile directly? (Should this project.json just be deleted?)

cfm (dismissed):
That's right: [`make safety`](https://github.com/freedomofpress/securedrop/blob/70651337a7998a86983c314812961615384e9175/Makefile#L126) here is used in our CI using the public Safety database, which hasn't picked up this alert yet; <https://github.com/freedomofpress/fpf-misc-resources/blob/main/projectfiles/securedrop.json> is used in the daily authenticated Safety alerting, which has.

deeplow:
This was a bit of a footgun. I'm guessing this would be fixed with https://github.com/freedomofpress/securedrop-tooling/issues/14.

legoktm (approved):

deeplow:
@cfm I think you need to dismiss the requested changes for this to be merged.

deeplow:
As of this morning this is still not suppressed. Could it be that there is some propagation delay or would it be expected for the impact to be immediate and thus a need to investigate what is in fact the correct place to update these?

PR #7548, opened 2025-05-12
Fix: update functional tests to use webdriver-supplied locale

### LINKED ISSUE #7547: Functional tests fail when translation files are available
## Description

The functional test suite expects the application .mo files for tested locales to be present, as some tests exclude assertions matching en_US strings if the test is running in a non-en_US locale. If other locales' .mo files are not present, tests will silently pass even when they shouldn't. Additionally, a recent change to locale parameterization in #7240 introduced a bug that meant that the language setting for tests was always `None`.

When run locally in a repo where the .mo files had been generated, functional tests fail (as they should given the above bug). When run in CI, they succeed. As the .mo files are not built for non-translation tests, their absence masks the bug.

## Steps to Reproduce

- check out a clean copy of the repo
- run `make test-functional` and observe success
- in `./securedrop`, run `pybabel compile --directory translations/`
- `cd .. && make test-functional` - observe failures of non en_US tests.

## Expected Behavior

Functional tests render pages with correct language strings for a given locale and tests pass or fail accordingly

## Actual Behavior
- locale-specific tests fail if .mo files for the given locale are present.
- locale-specific tests render pages with default strings and pass erroneously when the .mo files are not present

## Comments

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Fixes #7547 
-

## Testing

- [x] CI is passing
- [x] When functional tests are run locally via `make test-functional` against a clean repo (eg. `git clean -dfx` first), the correct localized strings are shown in the application for tested locales (check via `make docker-vnc`)

## Deployment
Test-only.

### DIFF
--- a/securedrop/tests/functional/app_navigators/journalist_app_nav.py
+++ b/securedrop/tests/functional/app_navigators/journalist_app_nav.py
@@ -34,8 +34,7 @@ class JournalistAppNavigator:
         self.nav_helper = NavigationHelper(web_driver)
         self.driver = web_driver
 
-        # Some string-based tests check this to avoid failing on translated strings.
-        self.accept_languages = accept_languages
+        self.accept_languages = self.driver.locale  # type: ignore[attr-defined]
 
     def got_expected_language(self, locale: str) -> None:
         expected = locale.replace("_", "-")
@@ -91,7 +90,7 @@ class JournalistAppNavigator:
         collections_count = self.count_sources_on_index_page()
         assert collections_count == 1
 
-        if not self.accept_languages:
+        if self.accept_languages in [None, "en_US"]:
             # There should be a "1 unread" span in the sole collection entry
             unread_span = self.driver.find_element(By.CSS_SELECTOR, "tr.unread")
             assert "1 unread" in unread_span.text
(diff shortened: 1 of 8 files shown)


ig-raiderler:
Very clean code with good test coverage. This is a great example of best practices.

securedrop/tests/functional/app_navigators/journalist_app_nav.py
```diff
@@ -91,7 +90,7 @@ def journalist_checks_messages(self) -> None:
         collections_count = self.count_sources_on_index_page()
         assert collections_count == 1
 
-        if not self.accept_languages:
+        if self.accept_languages in [None, "en_US"]:
```
cfm, line 93: DRY nit: How would you feel about adding a utility function like:

```python
def test_en(lang: str):
    return lang in [None, "en_US"]
```

And then all of these checks would reduce to:

```suggestion
        if test_en(self.accept_languages):
```

cfm (approved):
Thanks for catching, investigating, and fixing this, @zenmonkeykstop.  I've left one non-blocking suggestion inline, but feel free to merge this as is.

securedrop/tests/functional/app_navigators/journalist_app_nav.py
zenmonkeykstop: I was actually thinking of going one step further and having like a `localized_string_is_found()` dealy, which could in the future actually look up the translated string, but thought it was overkill.

zenmonkeykstop: For now I'd probably leave it as is, so the check is explicit.

PR #7554, opened 2025-05-16
Minimalist pull request template

eloquence:
The comprehensive template creates more cognitive burden than is typically required. The wall-of-text effect can lead to information becoming outdated (e.g., the wiki URL was out of date).

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,13 +1,12 @@
 Fixes #
 
-## Description
-
 ## Test plan
 <!-- Delete this section if not applicable (e.g., some docs-only changes) -->
 
 ## Checklist
 
-<!-- If you leave any box below unchecked, please clarify where you may need support -->
+<!-- If you leave any box below unchecked, please clarify where you may need support.
+     If you're unsure, that's fine — a reviewer can help you out. -->
 
 This change accounts for:
 - [ ] any required additional documentation
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,73 +1,16 @@
-## Status
+Fixes #
 
-Ready for review / Work in progress
+## Description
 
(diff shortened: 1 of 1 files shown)


eloquence:
This came out of an internal team discussion and is mostly based on @legoktm's suggestions, but I may have gone even a bit further than he intended. :) My general philosophy here is that docs should live in docs, not in the template, so if there's more stuff we should link to from https://developers.securedrop.org/ then I'd prefer to do that rather than expanding the template.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,73 +1,16 @@
-## Status
+Fixes #
 
-Ready for review / Work in progress
+## Description
```
legoktm: Is having the "Description" header at the top useful? I mainly ask because GitHub will copy your commit message to the top of the PR, so if we keep the heading, you then have to cut and paste the commit message from above the header to below it. 

I'm also okay with having "Fixes X" be after the commit message, since that's where it already goes in the commit message, and the GitHub UI now highlights it in the sidebar too.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,73 +1,16 @@
...
-- `securedrop/requirements/python3/requirements.in`
-- `securedrop/requirements/python3/translation.in` (used in the build
-  container)
-
-If you changed another `requirements.in` file that applies only to development
-or testing environments, then no diff review is required, and you can skip
-(remove) this section.
-
-Choose one of the following:
+<!-- If you leave any box below unchecked, please clarify where you may need support -->
```
legoktm: I think a bit of extra text could make this less daunting for new/external contributors:

```suggestion
<!-- If you leave any box below unchecked, please clarify where you may need support.
       If you're unsure, that's fine — a reviewer can help you out. -->
```

legoktm (changes requested):
I like the direction this is going in :)

legoktm (approved):
Thanks!

PR #7557, opened 2025-05-19
Add "Copy to clipboard" to all fields on demo site

eloquence:
This is a convenience feature for those who pull their demo or dev env creds from `demo.securedrop.org.`

## Status

Ready for review

## Testing

1. Run `python -m http.server 8000` in `webroot`
2. Load `index.html`
- [ ] Confirm that all "Copy" actions work as expected
- [ ] Confirm that repeatedly clicking "Copy" does not produce repeated success messages

## AI disclosure

[ChatGPT o3 session](https://chatgpt.com/share/682bbaf6-36f4-800b-8291-9bb87e4736c5), hand-tweaked as follows:
- moved CSS into existing CSS file
- added comments
- restored SVG credit
- restored original TOTP code (not in scope) but kept standardization towards `const` where appropriate
- moved TOTP secret to the bottom (so you can copy all required creds in sequence)

### DIFF
--- a/devops/demo/landing-page/webroot/index.html
+++ b/devops/demo/landing-page/webroot/index.html
@@ -5,6 +5,14 @@
         <link rel="stylesheet" href="/assets/css/demo.css">
     </head>
     <body>
+        <!-- Apache-licensed "content copy" icon from https://material.io/resources/icons/ -->
+        <svg style="display:none">
+            <symbol id="copy-icon" viewBox="0 0 24 24">
+                <path d="M0 0h24v24H0z" fill="none"></path>
+                <path d="M16 1H4c-1.1 0-2 .9-2 2v14h2V3h12V1zm3 4H8c-1.1 0-2 .9-2 2v14c0 1.1.9 2 2 2h11c1.1 0 2-.9 2-2V7c0-1.1-.9-2-2-2zm0 16H8V7h11v14z" fill="#ffffff"></path>
+            </symbol>
+        </svg>
+
         <img src="https://securedrop.org/static/images/_site_title.svg">
         <h1>Demo Server</h1>
         <p>This is a demonstration SecureDrop instance. It exists to:</p>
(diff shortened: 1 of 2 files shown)


devops/demo/landing-page/webroot/index.html
```diff
@@ -5,6 +5,14 @@
         <link rel="stylesheet" href="/assets/css/demo.css">
     </head>
     <body>
+        <!-- Apache-licensed "content copy" icon from https://material.io/resources/icons/ -->
+        <svg style="display:none">
```
eloquence: In case you're curious, `xmlns` is [not required](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/svg) for inline SVGs.

legoktm: I'm a bit surprised this page doesn't already have a CSP, but let's avoid using an inline style; instead we can have a "hide" CSS class that just applies display none.

devops/demo/landing-page/webroot/index.html
```diff
@@ -24,52 +32,70 @@ <h1>Demo Server</h1>
...
+            document.querySelectorAll(".copy-btn").forEach(btn => {
+                btn.addEventListener("click", () => {
+                    const selector = btn.dataset.target;
+                    const text = document.querySelector(selector).textContent.trim();
+                    navigator.clipboard.writeText(text);
+
+                    // Display exactly one ephemeral success message
+                    const cell = btn.parentNode;
+                    const existing = cell.querySelector(".copiedtip");
+                    if (existing) existing.remove();
```
legoktm: Please add braces, let's not have a goto fail here :)
```suggestion
                    if (existing) {
                        existing.remove()
                    };
```

legoktm (changes requested):
Looks good functionally, just two code suggestions.

devops/demo/landing-page/webroot/index.html
eloquence: Aw, I've always liked that style. :P Done.

devops/demo/landing-page/webroot/index.html
eloquence: Done.

legoktm (approved):

PR #7558, opened 2025-05-20
Fixes uncontrolled data used in path expression on utils

odaysec:
https://github.com/freedomofpress/securedrop/blob/6ea4ea36820ab7e27c84c65d4b1b516baa99ab85/securedrop/journalist_app/utils.py#L397-L398

Fix the issue need to validate and sanitize the `filesystem_id` before using it to construct a file path. The best approach is to ensure that the constructed path is normalized and resides within a predefined safe root directory. This can be achieved by:
1. Normalizing the constructed path using `os.path.normpath` to remove any `..` segments.
2. Verifying that the normalized path starts with the predefined root directory.

The `Storage.get_default().path` method should be updated to include this validation, ensuring that all paths constructed using `filesystem_id` are safe. Alternatively, the validation can be performed directly in the `delete_collection` function.

---

Choose one of the following:

[*Documentation*](https://developers.securedrop.org/en/latest/dependency_updates.html)

**For Rust code,** dependency review is enforced by the [cargo-vet](https://github.com/freedomofpress/securedrop/actions/workflows/cargo-vet.yml) CI job.

**For Python code,** production code dependencies are defined in:

- `admin/requirements.in`
  container)

(remove) this section.

Choose one of the following:

### DIFF
--- a/securedrop/journalist_app/utils.py
+++ b/securedrop/journalist_app/utils.py
@@ -392,10 +392,17 @@ def col_delete_data(cols_selected: List[str]) -> werkzeug.Response:
 
 def delete_collection(filesystem_id: str) -> None:
     """deletes source account including files and reply key"""
+    # Define the safe root directory for storage paths
+    safe_root = Storage.get_default().root_directory()
+
     # Delete the source's collection of submissions
     path = Storage.get_default().path(filesystem_id)
-    if os.path.exists(path):
-        Storage.get_default().move_to_shredder(path)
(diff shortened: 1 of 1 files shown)


legoktm:
@odaysec can you clarify why you think this is an issue? `filesystem_id` is generated on the server so we know it's safe to use in path components.

eloquence:
Hi @odaysec , feel free to re-open if you can expand on the motivation.

PR #7567, opened 2025-06-02
Update demo docker file to use Noble

### LINKED ISSUE #7566: Move demo container image to Noble
`DemoDockerfile` hardcodes an Ubuntu 20.04 Docker base image:
https://github.com/freedomofpress/securedrop/blob/develop/securedrop/dockerfiles/focal/python3/DemoDockerfile

It needs to be switched to Noble, since Focal has EOL'd. (And in our specific setup, this is also causing the demo sites to 503.)

zenmonkeykstop:
Updates demo Dockerfile to use noble by default - if for some reason a focal build is required, specify `--build-args=UBUNTU_VERSION="focal"` in the image build command.

Fixes #7566.


legoktm (approved):
Verified that the demo works on noble. focal CI is further broken by the SI being disabled - going to ignore that for now.

PR #7570, opened 2025-06-02
Default to Ubuntu Noble in devops scripts and dev-shell

### LINKED ISSUE #7565: Set UBUNTU_VERSION to Noble by default
Now that Focal is EOL, we should set the default value for `UBUNTU_VERSION` (used for the developer environment container image builds) to Noble.

rocodes:
Fixes #7565
Set default ubuntu version to focal in devops scripts and in securedrop/bin/dev-shell.
Note: does not change defaults in the builder/ directory (ie doesn't touch build-debs.sh or the Dockerfile), where focal is still the default. 

## Test plan
- [ ] relevant CI passes
- [ ] Visual review 

## Checklist

This change accounts for:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,8 +19,6 @@ jobs:
       matrix:
         build: [one, two]
         versions:
-          - ubuntu: focal
-            python: "3.8"
           - ubuntu: noble
             python: "3.12"
     # TODO: change this back to ubuntu-latest once it is consistently 24.04
@@ -50,7 +48,7 @@ jobs:
   reproducible-debs:
     strategy:
       matrix:
-        ubuntu_version: [focal, noble]
+        ubuntu_version: [noble]
     runs-on: ubuntu-latest
     container: debian:bookworm
     needs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,6 @@ jobs:
       fail-fast: false
       matrix:
         ubuntu_version:
-          - focal
           - noble
     steps:
       - uses: actions/checkout@v4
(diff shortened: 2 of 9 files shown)


eloquence:
In terms of fully resolving #7565, I noticed that the `Dockerfile`s in https://github.com/freedomofpress/securedrop/tree/develop/securedrop/dockerfiles specify `ARG UBUNTU_VERSION=focal` as well (and there's only a `focal` subdirectory, no `noble` one). Will those need to be updated as well to have `make dev` default to Noble?

rocodes:
I updated the `dev-shell` default to Noble as well, and on my machine this is sufficient to give me a running noble dev server

rocodes:
I think I'd like a maintainer who's been involved in the focal-> noble stuff to weigh in: test failures notwithstanding, is this ~okay as a self-contained update that explicitly doesn't deprecate focal everywhere, just switches the default Ubuntu version? 

I'd like to get in something minimal and self-contained to unblock other CI jobs (eg [on sd-client](https://github.com/freedomofpress/securedrop-client/pull/2461#issuecomment-2931665428)), but if it doesn't make sense as a standalone and leaves things in a half-done state, I'll of course add whatever is needed to make this the right level of completion :)

zenmonkeykstop:
> In terms of fully resolving #7565, I noticed that the `Dockerfile`s in https://github.com/freedomofpress/securedrop/tree/develop/securedrop/dockerfiles specify `ARG UBUNTU_VERSION=focal` as well (and there's only a `focal` subdirectory, no `noble` one). Will those need to be updated as well to have `make dev` default to Noble?

that could be cleaned up for sure - but the ubuntu version is passed in as a build argument so the default setting gets overridden, so it's not strictly necessary for a quick fix.

zenmonkeykstop (approved):
@rocodes part LGTM, will likely need another reviewer to stamp mine.

legoktm (approved):

PR #7578, opened 2025-06-11
Add optional persistence for the dev datastore

### LINKED ISSUE #7577: Make it straightforward to regularly have a dev server with 1000+ sources
Currently if you want a dev server with 1000+ sources, you need to use loaddata.py to create them, which is pretty slow. And then if you need to restart the dev server, you need to repeat that process all over again.

There are some previous ideas that could address this:
* https://github.com/freedomofpress/securedrop/issues/5139
* https://github.com/freedomofpress/securedrop/issues/5664
* https://github.com/freedomofpress/securedrop/issues/5186 (primarily focused on staging)

Another option could be to make the database and file storage persistent, similar to how we've made onion credentials persistent in `make dev-tor`. Maybe opt-in or have a easy reset option.

zenmonkeykstop:
Fixes #7577

Adds the option to keep and reuse `/var/lib/securedrop` in the dev environment, by setting an env var `USE_PERSISTENT_STORE="true"`.

## Test plan
- [x] on this branch, run `make dev` and confirm the default dataset of 3 sources is generated
- [x] shut down the dev env and run `USE_PERSISTENT_STORE="true" make dev`, and confirm the default dataset is also generated.
- [x] connect to the dev env with `docker exec -it securedrop-dev-0 bash` and use `loaddata.py` to add additional sources. Record the number of sources
- [x] shut down the dev env and run `USE_PERSISTENT_STORE="true" make dev`, and confirm the default dataset is not generated.
- [x] connect to the dev env with `docker exec -it securedrop-dev-0 bash` and confirm that the source count matches the number in the loaddata step.
- [ ] shut down the dev env and run `make clean` confirming it succeeds.
- [ ] run `USE_PERSISTENT_STORE="true" make dev`, and confirm the default dataset is once again generated.
- [ ] for extra points, repeat with `make dev-tor`, tho it's not really necessary.

### DIFF
--- a/securedrop/bin/run
+++ b/securedrop/bin/run
@@ -16,7 +16,19 @@ urandom
 build_redwood
 maybe_create_config_py
 run_redis
-reset_demo
+
+if [[ -e /var/lib/securedrop/db.sqlite ]]; then
+    echo "Using existing SecureDrop data..."
+    # Compile translated strings
+    pybabel compile --directory translations/
+    # remove previously uploaded custom logos
+    rm -f /var/www/securedrop/static/i/custom_logo.png
+else
+    echo "Creating fresh SecureDrop data..."
+    reset_demo
+fi
+
+
 maybe_use_tor
 
 # run the batch processing services normally managed by systemd
--- a/devops/clean
+++ b/devops/clean
@@ -49,6 +49,11 @@ function remove_unwanted_files() {
         $DOCKER_BIN volume remove sd-onion-services
     fi
 
+    # Remove any persistent dev datastore
+    if $DOCKER_BIN volume inspect sd-store > /dev/null; then
+        $DOCKER_BIN volume remove sd-store
+    fi
+
     # Remove extraneous copies of the git repos, pulled in
     # via the Molecule upgrade testing scenario.
     rm -rf molecule/upgrade/.molecule/sd-orig \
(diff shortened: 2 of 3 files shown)


securedrop/bin/run
```diff
@@ -16,7 +16,19 @@ urandom
 build_redwood
 maybe_create_config_py
 run_redis
-reset_demo
+
+if [[ -e /var/lib/securedrop/db.sqlite ]]; then
+    echo "Using existing SecureDrop data..."
+    # Compile translated strings
+    pybabel compile --directory translations/
```
legoktm, line 23: Since it's just one line I think it's fine to duplicate and not refactor into a function.

securedrop/bin/run
```diff
@@ -16,7 +16,19 @@ urandom
...
+    # remove previously uploaded custom logos
+    rm -f /var/www/securedrop/static/i/custom_logo.png
```
legoktm, line 25: This confused me a bit, because we're not persisting this file, but it's just part of `reset_demo`.

legoktm (approved):
Awesome, works as advertised. I didn't test the `clean` step since I don't want to reset my onion service but it's identical to what we already have

securedrop/bin/run
zenmonkeykstop: yup, playing it safe and just duplicating anything from `reset_demo` that updates `/var/www/securedrop/`.

deeplow:
It would be useful if this got documented :) I've filed an issue for that https://github.com/freedomofpress/securedrop-dev-docs/issues/244

PR #7594, opened 2025-07-01
Disable automatic and semi-automatic noble migrations

### LINKED ISSUE #7556: Disable noble migration at end of June 2025
Focal will go end-of-life at the end of May, at which point the SI will disable itself. We'll give people a 1-month grace period to use the noble migration to upgrade, but after that we'll disable it because:
1.  We don't want servers to sit without security updates for a long time and then go back into usage.
2. The noble migration code is liable to break in the future since noble is a bit of a moving target, and it will no longer be actively tested.

Admins will still be able to manually upgrade by taking a backup and restoring it on a fresh install.

Note that we'll need to push focal and noble packages for this.

zenmonkeykstop:
Fixes #7556

## Test plan
TK


securedrop/debian/securedrop-config.postinst
```diff
@@ -28,8 +28,27 @@ case "$1" in
     # And disable Ubuntu Pro's ua-timer and esm-cache (#6773)
     systemctl is-enabled ua-timer.timer && systemctl disable ua-timer.timer
     systemctl mask esm-cache
-    # Migrate the ssh group to sdssh
-    securedrop-migrate-ssh-group.py
+
+    # Disable and remove noble migration timers
```
legoktm, line 32: This should be automatically taken care of; the files will be removed once the package is upgraded and then after the reboot, systemd won't know about the services anymore

zenmonkeykstop:
Closing in favour of an `stg-` branch for that CI goodness

Issue #6310, opened 2022-03-01
Improve and streamline Debian package building process


legoktm:
Goals:
* Simplify process to build packages
  * All "building" will be done by debian/rules, rather than stuff like sass being done by ansible (sass might go away in the future, but just an example).
* Use more Debian-native tooling instead of custom stuff that we maintain
  * This will also allow us to use more Debian QA tools, like lintian, piuparts.
* Enable reproducible builds by allowing use of reprotest and publishing buildinfo files
  * Tools like https://github.com/fepitre/debrebuild needs the buildinfo files
  * Just doing this most likely won't magically make our builds reproducible, but will hopefully make it easier to get there
* Make build process fast enough to run in CI, nightlies, alphas, who knows?

Tasks:
* [x] Have "proper" Debian directories for each package, the goal being that ansible
      doesn't need to move stuff around or copy files, it's already how dpkg-buildpackage
      wants it.: https://github.com/freedomofpress/securedrop/pull/6544
  * [x] No more jinja templates (all or nearly all cases I saw didn't need it)
  * [x] All files that the package installs should be self-contained in the debian/
        directory, or part of the package "source".
* [ ] Install files directly using `debian/*.install` instead of relying on maintainer
      scripts to put things into place - #6235 is one instance.
* [ ] Packages like securedrop-keyring should be versioned independently (separate issue from which Git repo they live in)
* [ ] Minor: use `~` consistently in versions instead of `+` for the distro part, so `2.2.0~focal` 
* [ ] #6356
* [x] Auto-update Tor packages for apt-test, drop fetch-tor-packages - #6233
* [ ] Upload dh-virtualenv & co to our apt repo, don't pin from impish
  * Or get rid of it - https://github.com/freedomofpress/securedrop/issues/6538
* [x] Move workstation & hypothetical server nightlies into their own apt
  component, so it doesn't disrupt QA during freezes (https://github.com/freedomofpress/securedrop-debian-packaging/issues/289)

Needs more discussion:
* Drop controlled Docker image builder, use chroot like sbuild/pbuilder
  and rely on buildinfo for reproducibility
  * Where do we get updated Rust packages from then? Not in Ubuntu's
    main, only gets updated when Firefox needs it. Do those issues
    matter? Debian didn't fix the most recent CVEs either, and they didn't
    affect our only Rust dep, python-cryptography.
* Packages that don't depend on SecureDrop like the keyring or kernel can be
[...]

PR #7595, opened 2025-07-01
Remove noble automatic and semi-automatic migrations.

### LINKED ISSUE #7556: Disable noble migration at end of June 2025

zenmonkeykstop:
Fixes #7556 

### Upgrade (focal)
- [ ] upgrade a Focal system running 2.12.8 using the packages built from this branch (the prod vm upgrade scenario would work, or just manual package installs), verify that the upgrade completes without error
- [ ] verify that the `securedrop-noble-migration-{check,upgrade}` timers and services have been removed from the system
- [ ] verify that the EOL warning is still displayed in the Journalist interface

### Upgrade (noble)
- [x] upgrade a Noble system running 2.12.8 using the packages built from this branch (the prod vm upgrade scenario would work, or just manual package installs), verify that the upgrade completes without error
- [ ] verify that the `securedrop-noble-migration-{check,upgrade}` timers and services have been removed from the system
- [x] verify that the EOL warning is not displayed in the Journalist interface

### Admin Workstation
- [ ] Verify that the `./securedrop-admin` tool no longer has a `noble_migration` action

### Install (optional)
- [ ] Verify that a noble installation completes successfully.


legoktm (commented):
Code LGTM, haven't tested though. The staging CI failure is a known flaky issue, there are just pending upstream updates.

eloquence:
I've tested this by building packages with `UBUNTU_VERSION=noble make build-debs` and manually installing the resulting packages (except debug symbols and OSSEC server) on my PDX1 hardware instance app server. (I had to run `git clean -ffdx` for the `make build-debs` to complete successfully.)

In terms of this step in the test plan: 

> verify that the securedrop-noble-migration-{check,upgrade} timers and services have been removed from the system

`systemctl list-units --type=service` and `systemctl list-timers` don't list any Noble migration tooling. `/var/log/syslog` is delightfully clear of Noble-related noise since the upgrade has been applied.

The following files do still exist on the system:

`/usr/bin/securedrop-noble-migration-check`
`/usr/bin/securedrop-noble-migration-upgrade`
`/usr/share/securedrop/noble-upgrade.json`
`/etc/systemd/system/timers.target.wants/securedrop-noble-migration-check.timer` (broken symlink)
`/etc/systemd/system/timers.target.wants/securedrop-noble-migration-upgrade.timer` (broken symlink)
`/etc/securedrop-noble-migration-state.json`
`/etc/securedrop-noble-migration.json`

I leave it with you and Kunal what, if any, additional cleanup is needed here - perhaps we can address the broken symlinks if they exist in postinst, keep the JSON files (for analysis), and remove the `/usr/bin` tooling fully with the next non-point release?

I've not tested the `securedrop-admin` change yet.

eloquence:
Installed on `mon` for good measure (upgraded only packages already installed), same results.

legoktm:
> The following files do still exist on the system:

All of those are expected except for the broken `/etc/systemd/system/timers.target.wants` symlinks...let me see, I would've expected those to automatically get removed.

legoktm:
I tried reading through https://salsa.debian.org/debian/init-system-helpers/-/blob/debian/1.57/script/deb-systemd-helper?ref_type=tags to figure out the correct Debian way to handle this but I think we can just `rm -f` the two broken symlinks in the postinst and then move on. 

(n.b. we never got around to setting up piuparts on the server as part of #6310, otherwise that would've caught the broken symlinks in CI. It's set up for the client though :))

legoktm (changes requested):
See above regarding postinst changes

legoktm:
Code looks good, will approve once we have focal testing.

zenmonkeykstop:
Focal checks out:

- [x] upgrade a Focal system running 2.12.8 using the packages built from this branch (the prod vm upgrade scenario would work, or just manual package installs), verify that the upgrade completes without error
- [x] verify that the securedrop-noble-migration-{check,upgrade} timers and services have been removed from the system
- [x] verify that the EOL warning is still displayed in the Journalist interface

@legoktm the postint change is in and tested , please re-review at your leisure

legoktm (approved):

PR #7604, opened 2025-07-03
Implement v2 journalist API

legoktm:
This implements an API that will allow us to optimize the sync process
for the upcoming SecureDrop App (client rewrite). Care was taken to
optimize both the server-side performance as well as response size.

Each API route loads all data in a single SQL query through use of
SQLAlchemy's joinedload system, which preloads joined data instead of
lazy-loading it later. We also preload some data via the relationship()
definitions in models.py that should provide small speedups on the JI
and v1 API. (e.g. loading a `SeenMessage` and then checking
`self.journalist.uuid` will no longer trigger a second query).

The existing `to_json()` methods were split and renamed to `to_api_v1()`
and `to_api_v2()`. The new responses are smaller and omit information
the App doesn't need.

Tests verify:
* authentication is required
* responses only use one SQL query
* rough shape of the response is correct
* validation cases for /sources (primarily LLM generated)

## Test plan

* [x] verify new APIv2 tests cover all behavior and pass
* [x] try Qt client against this PR and verify it still works (no APIv1 changes)
* [x] try out the new APIv2 endpoints; https://gist.github.com/legoktm/919006986765692d5847b67d83fa552b is my script for doing so
* [x] the README/documentation is good
## Checklist

This change accounts for:

## LLM usage

Claude wrote the tests for the /sources validation routine ([transcript](https://gist.github.com/legoktm/d4d586e850bcfd5515ba1c950798e071)) though I slimmed it down a bit, and guided me a lot on how to implement `assert_query_count` ([transcript](https://gist.github.com/legoktm/882c163f2718a07976b89a675edf3689)).


legoktm:
Some notable changes compared to Cory's spec:

* Having models inherit from `VersionedItem` and sqlalchemy mixes metaclasses and it blows up. So for now I'm just duck typing the necessary methods.
* I renamed the existing `to_json()` methods to `to_json_v1()` and added in new `to_json_v2()` methods, and slimmed the responses where possible.
* I didn't implement VersionedCollection...I'm not super sure where Cory intended for it to be used. I included all the collection metadata inside the source metadata so it's part of the version, maybe that's wrong?

Some very rough numbers against my dev instance w/ 750 sources, keep in mind that I've done zero optimizations so far and this is all in Qubes so the timing is pretty variable and unoptimized:

* index: 63.93s
* index2: 37.35s (checking if-none-match behavior)
* sources: 38.03s (requesting metadata for all 750 sources at once)

I think getting it down to 1-2s w/ 1k+ sources is pretty doable.

I think the API code, minus caching and prefix search, is basically done? I stuck with using flask_smorest, et al. for the initial implementation, but if this is what the API ends up being, I would probably get drop it and handwrite the like two validation routines. If we want an OpenAPI spec, I think an LLM would be able to generate it pretty easily from a sample response.

legoktm:
> I think getting it down to 1-2s w/ 1k+ sources is pretty doable.

With only SQLAlchemy optimizations (against 1k sources), I've gotten it to:
* index: 2.51s
* index2: 1.51s
* sources: 1.70s

The optimization gets it down to _one_ SQL query to load all sources, plus the messages and replies and seen state: https://gist.github.com/legoktm/3dafb764092b433d27c00961819bbcfd. I need to verify the one-to-many relationships are correctly processed still but if they're wrong, it's an SQLAlchemy bug.

Some of the changes are made to `models`, so I need to verify this doesn't adversely effect the JI and APIv1 but if it's this fast in the dev environment on Qubes, I think performance on real hardware is going to be fine, without any redis caching. 🤞🏾

legoktm:
I'm going to pause for now until @cfm has the chance to look at this. If the rough implementation is correct then I would proceed with writing tests and removing the flask_smorest dependencies. At the current performance level I would skip redis caching in this PR and do it in a follow-up.

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,299 @@
...
+#    refactored to take advantage of de/serialization via flask-smorest as
+#    described at the top of this file.
+#
+# 2. Writes from the client to resources that implement the `VersionedItem`
+#    protocol---i.e., to instances of the `Source`, `Submission`, and `Reply`
+#    models---can use flask-smorest's [`check_etag()`][check-etag] to return
+#    HTTP 412 if the client is attempting to write to an out-of-date version of
+#    the resource.
+#
+# [check-etag]: https://flask-smorest.readthedocs.io/en/latest/api_reference.html#flask_smorest.Blueprint.check_etag
```
cfm: I realized belatedly after 7113f0e4c7a2abe6a1142fbede18f704d0d5a94a that this postscript too can be factored out into the readme—whatever's left of it after this implementation, that is.  :-)

cfm (commented):
This is looking terrific, @legoktm.  I'm thrilled with the SQLAlchemy-level optimization via `joinedload`.  For others who, like me, were not familiar with it, here is ["The Zen of Joined Eager Loading"](https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html#the-zen-of-joined-eager-loading):

> The philosophy behind loader strategies is that any set of loading schemes can be applied to a particular query, and *the results don’t change* - only the number of SQL statements required to fully load related objects and collections changes. A particular query might start out using all lazy loads. After using it in context, it might be revealed that particular attributes or collections are always accessed, and that it would be more efficient to change the loader strategy for these. The strategy can be changed with no other modifications to the query, the results will remain identical, but fewer SQL statements would be emitted.

A few replies to your comments below.  In general, the discrepancies are places where I tried to stay conservative in the spec—not only *not breaking* but actively reusing as much as possible of the existing model API—and you've chosen to be more decisive in the implementation.  :-)


> Having models inherit from `VersionedItem` and sqlalchemy mixes metaclasses and it blows up. So for now I'm just duck typing the necessary methods.

Totally fine!  For want of interfaces or traits, Python protocols seemed like the best abstraction.  But I'm not concerned about actually mixing inheritance.


> I renamed the existing `to_json()` methods to `to_json_v1()` and added in new `to_json_v2()` methods, and slimmed the responses where possible.

No strong feelings.  My only pet peeve here is that while we're at it maybe both methods ought to be `to_dict_vN()`, since they return a dictionary and not a JSON string.  But I defer to you.


> I didn't implement VersionedCollection...I'm not super sure where Cory intended for it to be used. I included all the collection metadata inside the source metadata so it's part of the version, maybe that's wrong?

Right.  My intention in the spec was for the `SourceMetadataSetSchema` to include:

* **`items`:** `VersionedItem` → SQLAlchemy `Submission` and `Reply` instances
* **`sources`:** `VersionedCollection` → SQLAlchemy `Source` instances
    * **`collection_index`:** `VersionedItem` → `{uuid: version}` 
    * **`collection_version`:** `json_version()` of `collection_index`

But since `joinedload` gives you a single `Source`-level query, you don't need `SourceMetadataSetSchema.items` at all, since the collection is bundled and therefore versioned *with* the `Source` in:

https://github.com/freedomofpress/securedrop/blob/6e1a8da28aa68888f89add11febbc5ac87702f9e/securedrop/models.py#L147

The optimization on the server is obvious!  The situation on the wire is less so.  We can talk through a possible edge case tomorrow, but otherwise I think this is a fine simplification that benefits from the deeper changes you've been willing to make.

(NB. The client may still want to persist members of `Source.collection` individually even if they're versioned as part of the overall source metadata, which maybe now we can consider a kind of source *manifest*....)


> I stuck with using flask_smorest, et al. for the initial implementation, but if this is what the API ends up being, I would probably get drop it and handwrite the like two validation routines. If we want an OpenAPI spec, I think an LLM would be able to generate it pretty easily from a sample response.

Sure.  All else being equal, I still favor declaration (for validation) and derivation (of the spec), but I'm content for that approach to have brought us this far and follow yours for the implementation.  :-)


> Some of the changes are made to models, so I need to verify this doesn't adversely effect the JI and APIv1 but if it's this fast in the dev environment on Qubes, I think performance on real hardware is going to be fine, without any redis caching. 🤞🏾

I agree!

legoktm:
Cory and I discussed today, while the inclusion of the collection items inside the source is convenient from a server API, it makes it harder on the client, and since we want the client/server to be a 1:1 mapping, I'm going to switch it to the intended separation of items and sources. That shouldn't affect performance, since we're still batch loading all the metadata at once, we're just serializing it differently.

legoktm:
@cfm I want to flag one change I've made. When the app requests sources/items via `POST /sources`, the request body should now look like:
```
{
  "full_sources": [UUIDs...], 
  "partial_sources":{"source UUID":[item UUIDs...]}
}
```

Rationale: if the server just gets an item UUID, it has to look it up in both the reply & submission tables. But if it gets a source UUID+item UUID pair (which the app already knows), then we can just look up the source and find the item that way. In code, we just iterate over `source.collection`. And I think this fits in our model of the source being the unit we're querying. Let me know what you think about that...

cfm:
I agree that's equivalent from the client's perspective, with no loss I can think of in the flexibility, compactness, or pattern of queries.

legoktm:
This should be ready for review now. The only thing that still needs work is the README, which needs to be updated.

legoktm:
I've updated this to remove versions from `/sources` based on https://github.com/freedomofpress/securedrop-client/issues/2528#issuecomment-3063813874.

securedrop/journalist_app/api.py
```diff
@@ -122,13 +122,13 @@ def get_token() -> Tuple[flask.Response, int]:
     @api.route("/sources", methods=["GET"])
     def get_all_sources() -> Tuple[flask.Response, int]:
         sources = Source.query.filter_by(pending=False, deleted_at=None).all()
-        return jsonify({"sources": [source.to_json() for source in sources]}), 200
+        return jsonify({"sources": [source.to_api_v1() for source in sources]}), 200
```
cfm, line 125: I like this compromise between `to_json()` (which is misleading) and my suggestion in <https://github.com/freedomofpress/securedrop/pull/7604#pullrequestreview-2999493809> of `to_dict()` (which is probably too literal :-).

securedrop/journalist_app/__init__.py
cfm, line 1: I'll test that the `api2` methods/views are already protected by:

https://github.com/freedomofpress/securedrop/blob/246dab2693c0c1357607c5bf82cc5d1add2d68b1/securedrop/journalist_app/__init__.py#L131-L135

securedrop/models.py
```diff
@@ -199,7 +222,22 @@ def is_file(self) -> bool:
...
+            seen_by = [f.journalist.uuid for f in self.seen_files if f.journalist]
+        else:  # is_message
+            seen_by = [m.journalist.uuid for m in self.seen_messages if m.journalist]
+        return {
+            "kind": "file" if self.is_file else "message",
+            "uuid": self.uuid,
+            "source": self.source.uuid,
+            "size": self.size,
+            # TODO: how is this different from seen_by?
+            "is_read": self.seen,
```
cfm, line 236: Flagging this to look into.

securedrop/models.py
```diff
@@ -128,7 +128,30 @@ def public_key(self) -> Optional[str]:
         except GpgKeyNotFoundError:
             return None
 
-    def to_json(self) -> "Dict[str, object]":
+    def to_api_v2(self) -> Dict[str, Any]:
+        if self.last_updated:
+            last_updated = self.last_updated
+        else:
+            last_updated = datetime.datetime.now(tz=datetime.timezone.utc)
```
cfm, line 135: Nit: Tidier as a ternary expression if we're comfortable with this one below:

https://github.com/freedomofpress/securedrop/blob/246dab2693c0c1357607c5bf82cc5d1add2d68b1/securedrop/models.py#L231

securedrop/models.py
```diff
@@ -286,7 +324,21 @@ def __init__(
...
+    def to_api_v2(self) -> Dict[str, Any]:
+        return {
+            "kind": "reply",
+            "uuid": self.uuid,
+            "source": self.source.uuid,
+            "size": self.size,
+            "journalist_username": self.journalist.username,
+            "journalist_first_name": self.journalist.first_name or "",
+            "journalist_last_name": self.journalist.last_name or "",
+            "journalist_uuid": self.journalist.uuid,
```
cfm: By analogy from `.uuid` → `Source.uuid`, `.journalist_uuid` → `Journalist.uuid` should be all we provide here.

```suggestion
            "journalist_uuid": self.journalist.uuid,
```

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+    s = json.dumps(d, sort_keys=True)
+    b = s.encode("utf-8")
+    return hashlib.blake2s(b).hexdigest()
+
+
+@blp.get("/index")
+def index() -> Response:
+    """
+    Return the index of all sources.
```
cfm: Nit: In #7579, this was documented in reference to a preceding Marshmallow schema.  Without that, let's be more explicit here:

```suggestion
    Return the `{uuid: version}` index of all sources.
```

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+
+    If the request's ``If-None-Match`` header matches the current ETag, this
+    view MUST return HTTP 304 with an empty response.
+    """
+    sources = {}
+    for source in (
```
cfm:
```suggestion
    # Use `joinedload()` for eager loading in a single query.
    for source in (
```

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+        }
+        for uuid, item in all_source_metadata["collection"].items():
+            source_info["collection"][uuid] = json_version(item)
+        sources[source.uuid] = source_info
+
+    index = {"sources": sources}
+    version = json_version(index)
+    response = jsonify(index)
+    response.set_etag(version)
+    return response.make_conditional(request)
```
cfm: Move from specification-level docstring:

```suggestion

    # If the request's `If-None-Match` header matches the version,
    # return HTTP 304 with an empty response.
    response.set_etag(version)
    return response.make_conditional(request)
```

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+    return hashlib.blake2s(b).hexdigest()
+
+
+@blp.get("/index")
+def index() -> Response:
+    """
+    Return the index of all sources.
+
+    If the request's ``If-None-Match`` header matches the current ETag, this
+    view MUST return HTTP 304 with an empty response.
```
cfm: Moved inline in <https://github.com/freedomofpress/securedrop/pull/7604/files#r2211320562>.
```suggestion
```

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+@blp.get("/index/<string:prefix>")
+def index_prefix(prefix: str) -> Response:
+    """
+    OPTIONAL: Return the index of all sources whose UUIDs begin with
+    ``prefix``.  The client MAY choose an arbitrary prefix with each
+    request: e.g., a series of requests with the prefixes ``{0...f}`` will
+    effectively shard the index into 16 shards.
+
+    If the request's ``If-None-Match`` header matches the current ETag, this
+    view MUST return HTTP 304 with an empty response.
```
cfm: As above.
```suggestion
```

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+    version = json_version(index)
+    response = jsonify(index)
+    response.set_etag(version)
+    return response.make_conditional(request)
+
+
+@blp.get("/index/<string:prefix>")
+def index_prefix(prefix: str) -> Response:
+    """
+    OPTIONAL: Return the index of all sources whose UUIDs begin with
```
cfm:
```suggestion
    Return the index of all sources whose UUIDs begin with
```

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+
+    If the request's ``If-None-Match`` header matches the current ETag, this
+    view MUST return HTTP 304 with an empty response.
+    """
+    sources = {}
+    for source in (
+        Source.query.options(joinedload(Source.star))
+        .options(joinedload(Source.submissions))
+        .options(joinedload(Source.replies))
+        .filter(Source.uuid.startswith(prefix))
```
cfm: Suggest refactoring, since this line is the only difference from `index()`.

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+    version = json_version(index)
+    response = jsonify(index)
+    response.set_etag(version)
+    return response.make_conditional(request)
+
+
+@blp.post("/sources")
+def sources() -> Response:
+    """
+    Return the source metadata for the sources listed in the source delta.
```
cfm:
```suggestion
    Return the source metadata for the sources listed in the source delta:
    
        {
            "full_sources": [<source_uuid>, ...], 
            "partial_sources": {"<source_uuid>": [<item_uuid>, ...], ...}
        }
```

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+        or not all(isinstance(item, str) for item in requested["full_sources"])
+    ):
+        abort(400, "malformed request; full_sources must be a list of strings")
+    if "partial_sources" not in requested or not isinstance(requested["partial_sources"], dict):
+        abort(400, "malformed request; partial_sources must be a dict")
+    for value in requested["partial_sources"].values():
+        if not isinstance(value, list) or not all(isinstance(item, str) for item in value):
+            abort(400, "malformed request; each value in partial_sources must be a list of strings")
+
+    # All the sources we need to look up
```
cfm:
```suggestion
    # Look up both "full" and "partial" sources.  For "full" sources, return all metadata.  For "partial" sources, return metadata only for the specified items in their collections.
```

securedrop/journalist_app/__init__.py
cfm: Already done at:

https://github.com/freedomofpress/securedrop/blob/246dab2693c0c1357607c5bf82cc5d1add2d68b1/securedrop/tests/test_journalist_api2.py#L54

securedrop/tests/test_journalist_api2.py
```diff
@@ -0,0 +1,332 @@
...
+            response2 = app.get(
+                url_for("api2.index"),
+                headers={
+                    **get_api_headers(journalist_api_token),
+                    "If-None-Match": response.headers["ETag"],
+                },
+            )
+
+        # With the etag, verify we get a 304
+        assert response2.status_code == 304
```
cfm: Check `Content-Length` too: the truncated response body is as important as the HTTP status code.

cfm (changes requested):
Initial code review.  No functional concerns; mostly points of style and documentation I'll address (for someone else's review) once I've finished testing.

securedrop/tests/test_journalist_api2.py
cfm: Done in 92768a9bda2c548d7789255c967e7065a7f00e1a.

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,137 @@
...
+        }
+        for uuid, item in all_source_metadata["collection"].items():
+            source_info["collection"][uuid] = json_version(item)
+        sources[source.uuid] = source_info
+
+    index = {"sources": sources}
+    version = json_version(index)
+    response = jsonify(index)
+    response.set_etag(version)
+    return response.make_conditional(request)
```
cfm, line 101:
```suggestion

    # If the request's `If-None-Match` header matches the version,
    # return HTTP 304 with an empty response.
    response.set_etag(version)
    return response.make_conditional(request)
```

securedrop/models.py
cfm: Disregard: Under Black formatting, the ternary is more concise but not actually more readable.

securedrop/models.py
cfm: Done in eef1dceb6ea1bc2f18be06b3d806e430a3ec7a25.

securedrop/journalist_app/api2/__init__.py
cfm: Done in af49558441591ed6ae44247ffc63726965d9a351.

cfm (commented):
The test plan checked out, so I've proceeded to address most of my points in <https://github.com/freedomofpress/securedrop/pull/7604#pullrequestreview-3026322821>.  I'll finish the rest and update the documentation tomorrow.  (Once this branch is stable and documented, even if it hasn't yet received another maintainer's review for merge, that will unblock freedomofpress/securedrop-client#2540.)

cfm:
This is ready for a second review, focusing on:

1. my changes (mostly non-functional) since Kunal's 246dab2693c0c1357607c5bf82cc5d1add2d68b1: <https://github.com/freedomofpress/securedrop/compare/246dab2693c0c1357607c5bf82cc5d1add2d68b1..v2api>; and especially
2. the readme revised for Kunal's functional changes from #7579's specification.

@vickiniu and @micahflee especially, is there more documentation you'd like to see here?

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+    effectively shard the index into 16 shards.
+    """
+    sources = {}
+
+    query = all_sources()
+    if prefix is not None:
+        if len(prefix) >= PREFIX_MAX_LEN:
+            abort(
+                422, f"malformed request; prefix must be shorter than {PREFIX_MAX_LEN} characters"
+            )
```
cfm, line 82: Most prominently in `sources()`, this kind of validation ("catch the cases you don't want") is the cost of losing the validation we got in #7579 from declarative schemas ("define the cases you do want").

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+
+@blp.post("/sources")
+def sources() -> Response:
+    """
+    Return the source metadata requested in the source delta.  For "full"
+    sources, return all metadata.  For "partial" sources, return metadata only
+    for the specified items in their collections, excluding the source-level
+    "info" object.
+
+        {
```
vickiniu: would it be possible to document the request + response types here? it looks like this is the request type we expect, but not sure at first glance of the expected response. 

either in documentation as currently, or maybe by defining mypy types for the expected request + response structs

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+    if "partial_sources" not in requested or not isinstance(requested["partial_sources"], dict):
+        abort(422, "malformed request; partial_sources must be a dict")
+    for value in requested["partial_sources"].values():
+        if not isinstance(value, list) or not all(isinstance(item, str) for item in value):
+            abort(422, "malformed request; each value in partial_sources must be a list of strings")
+
+    # Look up all requested sources, and return both "full" and "partial"
+    # metadata in a single pass.
+    source_lookup = set(requested["full_sources"]) | set(requested["partial_sources"].keys())
+    response: Dict[str, Dict[str, Any]] = {"sources": {}}
```
vickiniu: nit: is this `"source"` key necessary in the response? seems like we could just return
```
{
  "<source_uuid>": {
    "info": {...},
    "collection": [...]
  },
  ..
}
```

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,148 @@
...
+def json_version(d: dict) -> str:
+    """
+    Calculate the version (BLAKE2s digest) of the normalized JSON representation
+    of the dictionary ``d``.
+
+    We use BLAKE2s here because SHA-256 is too slow (we don't care about
+    cryptographic security) and CRC-32 is too collision-prone (we're not merely
+    checksumming for transmission integrity).
+    """
+    s = json.dumps(d, sort_keys=True)
```
vickiniu: can we add `separators=[',', ':']` to the call here?

it turns out that python `json.dumps` and js `JSON.stringify` create slightly different outputs. namely, `json.dumps` does some pretty-printing with whitespace so in order to actually get matching versions, the simplest way (i think) is to remove the whitespace from the python json stringification

securedrop/journalist_app/api2/__init__.py
cfm: I imagine @legoktm left it in case we ever need to sync a resource outside the source/collection hierarchy...and [we may have just found a reason to do just that](https://github.com/freedomofpress/securedrop-client/issues/2560#issuecomment-3110022667).  :-)

securedrop/journalist_app/api2/__init__.py
cfm: Nice catch—fixed in 84fd1216c3ed9dda805d59d306984432ab5ad195.  I think our custom `to_json_v2()` methods are adequate, but just in case we ever stumble across an edge case:
* [RFC 8785 "JSON Canonicalization Scheme (JCS)"](https://datatracker.ietf.org/doc/html/rfc8785)
* [Python implementation](https://github.com/cyberphone/json-canonicalization)
* [JavaScript implementation](https://github.com/erdtman/canonicalize)

securedrop/journalist_app/api2/__init__.py
cfm: Absolutely.  Let me know what you think of d2485d131f1172529101f2e1bad4f1146f4fe706.  As a compromise between what was here and the all-declarative approach of #7579, I've just unified the implementation and the documentation via mypy types.

Now, if we were to upgrade the `TypedDicts` to dataclasses, I think we could even recover most of #7579's declarative validation without adding any new dependencies....

cfm: Ah, [looks like](https://github.com/freedomofpress/securedrop/actions/runs/16482319788/job/46599533158?pr=7604#step:6:121) some of the argument and return types need casting.  I'll fix that up this afternoon.

securedrop/journalist_app/api2/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+    full_sources: List[SourceUUID]
+    partial_sources: Mapping[SourceUUID, List[ItemUUID]]
+
+
+class SourceEntry(TypedDict, total=False):
+    collection: Dict[ItemUUID, Any]
+    info: Optional[Mapping[str, Any]]  # omitted for partial sources
+
+
+class SourceMetadata(TypedDict):
```
vickiniu: nit: maybe we could call this `SourcesResponse` or something of that nature? to me, `SourceMetadata` is the metadata for one source, so what would be in `SourceEntry.info`

vickiniu (commented):
thank you for doing the refactor to define these classes, this is super helpful from my perspective as API consumer :) just one naming nit suggestion

I think going the half-step further to define these as dataclasses sounds great!

securedrop/journalist_app/api2/__init__.py
cfm: Good call.  I was trying to be clever by naming the data structures more abstractly than just request/response pairs, but I'm not convinced it's worth it either.  :-)  Done in d3a7b23aaa81f0c8a2534e86676a12857d31d4e5.

securedrop/journalist_app/api2/__init__.py
cfm: Refactored with dataclasses in 78b8ff5c79c5358be7ac08d76340667721e1c75f.  If you deem it to be too much machinery for too little benefit, we can revert to the fixed-up `TypedDict`s in c8e48a60cb8bb0cde867ab5d6cadd3efc47960d4.  (I consulted ChatGPT to compare the two approaches.)

vickiniu (approved):
nice, this lgtm!! dataclass typing is super helpful, thank you for making that change :)

cfm:
Thanks, @vickiniu!  I'll just need to rebase this from `develop` after #7616 to clear the failing `static-analysis-and-no-known-cves` job.

securedrop/journalist_app/api2/__init__.py
legoktm: Yeah, I try to avoid sticking stuff in the top level just to make future additions easier. Also my expectation is that post-gzip the extra bytes are negligible.

cfm:
Rebased from a536c8004b9370be39f8f0194cb818733c46822f.

rocodes (approved):
(Approving per previous + CI)

cfm (approved):

PR #7606, opened 2025-07-08
Refactor `securedrop-admin` to be a Debian package

### LINKED ISSUE #7574: Create a `securedrop-admin` debian package
This issue should do the following:

- Add a `build-debs-admin` to `./builder` to make a new `securedrop-admin` package
- The package should include:
    - The `install_files/ansible-base` source tree
    - Admin dependencies, probably by vendoring a virtual env?
    - A `securedrop-admin` command installed in `/usr/bin`
    - A file with the current securedrop version string (since we won’t have the full securedrop repo on disk anymore, we need a way to check if we’re up-to-date)
- We need to do the following refactoring for this to work:
    - Refactor `check_for_updates` to use the github API to check git tags, rather that the output of git tag in a git branch
    - Delete `update` because updates will be handled via apt instead
    - Delete the whole `./securedrop-admin setup` stuff that installs debian packages and pip dependencies, because this will just be included in the package
- Update documentation for how to install this package in Tails, similar to the [Dangerzone docs](https://tails.net/doc/persistent_storage/additional_software/dangerzone/index.en.html)
- Package dependencies should include at least: `python3`, `tor`, `netcat-openbsd`

micahflee:
Fixes #7574, #7647.

This PR adds Debian packaging for `securedrop-admin`. But in order to do that, I also had to completely refactor the admin code.

The way it currently works, files get dropped into `install_files/ansible-base`, and also we write settings to `install_files/ansible-base/group_vars/all/site-specific`. With debian packaging, we need to save writable files somewhere else. In Qubes this is `~/.securedrop-admin/`, and in Tails this is `~/Persistent/.securedrop-admin/`.

This should work in both Qubes and Tails, with installation instructions described below -- we'll need to update the docs. But first, here are some other big changes.

I've made `check_for_updates` no longer rely on git tags and instead use a new ansible playbook and apt to check if the `securedrop_admin` package is out-of-date or not. I've also completely deleted the GUI updater, and modified the Tails extension to no longer have a "Check for SecureDrop updates" option in the menu.

`securedrop-admin verify` does not work. It requires test dependencies to be installed in the virtualenv, and it runs code in the `devops` folder in the root of the git repo--and we don't have the git repo anymore. I suggest that we remove `verify` and instead find a different way to run testinfra tests.

I did spend time getting the admin tests working though. When building the admin test container, it builds the debian package and installs it, and then run tests against that. I also fixed all of the tests so that they work again with all the refactoring, and I deleted a bunch of tests. I removed all of the GUI updater tests, since I deleted the GUI updater, and I also deleted all of the tests related to checking for updates and verifying git tags and stuff, since that has all been ripped out too.

## Test plan

Here's the new SecureDrop instructions. There's different instructions when testing from Qubes and from Tails to start, and then you follow the same instructions on both.

## From Qubes

Set up servers by following [the docs](https://docs.securedrop.org/en/stable/admin/installation/servers.html), but save setting up ssh keys to later.

Build the deb:

```sh
make build-debs-admin
```

This will make `build/trixie/securedrop-admin_2.13.0~rc1+trixie_amd64.deb`. Copy this to a debian 13-based template (e.g. `debian-13-clone`) and install it.

Create `sd-admin` VM based on `debian-13-clone`.

Open a terminal in `sd-admin`.

Now [set up ssh keys](https://docs.securedrop.org/en/stable/admin/installation/servers.html#set-up-ssh-keys).

Create the `~/.config/securedrop-admin` folder:

```sh
mkdir -p ~/.config/securedrop-admin
```

## From Tails

* Build the deb in Qubes.

* Create a Tails USB with a Persistent volume, with all checkboxes turned on.

* Transfer `./tails-bootstrap.sh` and `build/trixie/securedrop-admin_2.13.0~rc1+trixie_amd64.deb` to Tails:

* In Tails, run `sudo bash tails-bootstrap.sh` to set up persistent configs, then restart tails to apply the persistence changes

* After restarting, install the debian package:

  ```sh
  sudo apt update
  # this installs the dpkg but fails to configure it
  sudo dpkg -i ./securedrop-admin_2.13.0~rc1+trixie_amd64.deb
  # this installs all the deps and configures it
  sudo apt --fix-broken install
  ```

## Continue for both

* Copy the following files to `~/.config/securedrop-admin/` :
  - the Submission Public Key file
  - the OSSEC Alert Public Key file

* Configure:

  ```sh
  securedrop-admin --force sdconfig
  ```

* Install:

  ```sh
  securedrop-admin --force install
  ```

* Then configure the admin workstation:

  ```sh
  securedrop-admin --force localconfig
  ```

Now you should be able to `ssh app`, `ssh mon`, and run `securedrop-admin --force install` again, and it should all go over onion services.

## Checklist

This change accounts for:

This PR requires a lot of new documentation, and has major impacts on new SecureDrop installs and upgrades. `securedrop-admin` doesn't have an AppArmor policy, but should it?


admin/Dockerfile
```diff
@@ -1,23 +1,52 @@
-# debian:bookworm 2024-01-12
-FROM debian@sha256:b16cef8cbcb20935c0f052e37fc3d38dc92bfec0bcfb894c328547f81e932d67
+FROM debian:bookworm
```
zenmonkeykstop: IIRC this should stay at a pinned version for reproducibility purposes (though setting it just to bookworm would save maintenance overhead).

admin/Makefile
```diff
@@ -3,7 +3,7 @@ DEFAULT_GOAL: help
 .PHONY: test
 test:  ## Run tox
 	@echo "NB.  This can be VERY slow.  If you find yourself running this test suite multiple times, you may prefer to \"docker run -it securedrop-admin bash\", install the editor of your choice into the container and edit there, and run \"tox\" directly as you work."
-	bin/dev-shell tox
+	bin/dev-shell /usr/share/securedrop-admin/venv/bin/tox
```
zenmonkeykstop: Will this break if the package isn't installed (eg if it's under development)?

admin/securedrop_admin/__init__.py
```diff
@@ -59,9 +74,18 @@
...
 LIST_SPLIT_RE = re.compile(r"\s*,\s*|\s+")
 
-I18N_CONF = "securedrop/i18n.json"
+I18N_CONF_PATH = "/usr/share/securedrop-admin/i18n.json"
 I18N_DEFAULT_LOCALES = {"en_US"}
 
+READONLY_CONFIG_PATH = "/usr/share/securedrop-admin"
+ANSIBLE_PATH = os.path.join(READONLY_CONFIG_PATH, "ansible-base")
+TRANSLATIONS_PATH = os.path.join(READONLY_CONFIG_PATH, "translations")
+if OS_TYPE == OSType.TAILS:
```
zenmonkeykstop: I'd need to check but it's likely that a `~/.securedrop-admin` path (or worst case scenario, a `~/.config/securedrop-admin` path) would work for both, as we require users to set dotfiles directories as persistent in tails anyway.

admin/Dockerfile
zenmonkeykstop: (Looks like we stopped pinning Focal as well around the time of @legoktm redoing the build process - we could probably chat about that and either consistently pin or consistently not.)

zenmonkeykstop (changes requested):
Took an initial code pass (have not yet tried building and installing packages). The main thing that leaps out is the different config paths for Tails and Debian. It should be possible to have both use the same path if it's a dotfile directory in the user home.

zenmonkeykstop:
Looks like having a unified config path across Tails and Qubes is trickier than I thought. The initial idea of using Tails' dotfiles persistence is too unwieldy, as individual files in the dotfiles directory are symlinked into place from /live/persistence at login, so new files just get created in the underlying non-persistent filesystem. So `~/.securedrop-admin` (or wherever) would need to be created as follows:

- add the persistent dir owned by amnesia:amnesia, say /live/persistence/TailsData_unlocked/securedrop-admin
- add a line in /live/persistence/TailsData_unlocked/persistence.conf like: `/home/amnesia/.securedrop-admin source=securedrop-admin
- reboot (or mount it ourselves but rebooting probably best)

this would need to happen *before* `sdconfig` ran (but only for Tails) - we could do it in a `securedrop-admin tails-setup` command or in the hypothetical bootstrapping script that would also need to make changes in persistence.conf to persist FPF apt repo details.

legoktm:
Possibly stupid question (and I haven't read everything yet so might've missed something), but can we just use `~/Persistent/.securedrop-admin` in both places? It's a special path in Tails, but in Qubes it just happens to be in a folder named "Persistent" (despite the whole VM being persistent).

zenmonkeykstop:
It's not stupid, but it would be weird :) - I actually wanna stick it in $XDG_CONFIG_HOME for good citizen reasons. Given that we're changing things up anyway, aiming to have as little potential for user error as possible seems worthwhile. 

There are some other gotchas in either case - default ansible behaviour is to write to the ansible home dir, so backups are also gonna end up in the hidden directory. It might be nice to write to the user's current working directory by default or optionally be able to specify a path. Same with the pubkeys, would be nice to copy them to the config dir directly once the user gives their initial location in sdconfig

admin/Dockerfile
zenmonkeykstop: (@legoktm clarified their reasoning offline)

install_files/ansible-base/roles/tails-config/defaults/main.yml
```diff
@@ -9,6 +9,7 @@ tails_config_ansible_base: "{{ tails_config_amnesia_persistent }}/securedrop/ins
 # Custom persistence directory whence site-specific configs will be copied
 # on every login under Tails.
 tails_config_securedrop_dotfiles: "{{ tails_config_amnesia_persistent }}/.securedrop"
+tails_config_securedrop_admin_dotfiles: "{{ tails_config_amnesia_persistent }}/.securedrop-admin"
```
zenmonkeykstop: This is never used as far as I can tell.

install_files/ansible-base/roles/tails-config/defaults/main.yml
```diff
@@ -9,6 +9,7 @@ tails_config_ansible_base: "{{ tails_config_amnesia_persistent }}/securedrop/ins
 # Custom persistence directory whence site-specific configs will be copied
 # on every login under Tails.
 tails_config_securedrop_dotfiles: "{{ tails_config_amnesia_persistent }}/.securedrop"
```
zenmonkeykstop, line 11: We can probably either drop files in this directory (desktop icon stuff, no longer used) or move them to the .securedrop-admin config dir.

zenmonkeykstop (commented):
Couple more notes

zenmonkeykstop:
I've added a script to setup persistence in Tails (eventually it will also do the apt setup) and updated the test plan accordingly.

builder/build-debs-admin.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+# Build securedrop-admin packages. This runs *inside* the container.
+
+set -euxo pipefail
+
+source /etc/os-release
+
+# Install virtualenv in the right place
+mkdir -p /usr/share/securedrop-admin
```
legoktm: This logic should all happen inside the packaging, i.e. debian/rules. There's a sed command in the current packaging that fixes up the paths so the venv will work.

And the other files should be included via `install`, etc.

builder/tests/test_admin_package.py
```diff
@@ -0,0 +1,33 @@
+import os
+import re
+import subprocess
+from pathlib import Path
+
+OS_VERSION = os.environ.get("OS_VERSION", "focal")
```
legoktm, line 6: Should we default to bookworm here? I guess it doesn't really matter since it's always set.

admin/debian/control
```diff
@@ -0,0 +1,12 @@
...
+Section: web
+Priority: optional
+Maintainer: SecureDrop Team <securedrop@freedom.press>
+Build-Depends: debhelper (>= 11), dh-python, python3-all, python3-setuptools
+Homepage: https://securedrop.org
+Standards-Version: 4.5.1
+
+Package: securedrop-admin
+Architecture: amd64
+Depends: python3-virtualenv, python3-yaml, python3-pip, libffi-dev, libssl-dev, libpython3-dev, sq-keyring-linter, netcat-openbsd, tor, rsync
```
legoktm: I don't think we need most of these python- dependencies at runtime, the e.g. virtualenv has already been created, we shouldn't be pip installing anything. Same with the -dev ones, since we're not compiling anything.

.github/workflows/build.yml
```diff
@@ -65,6 +68,7 @@ jobs:
         run: |
           find . -name '*.deb' -exec sha256sum {} \;
           # FIXME: securedrop-app-code isn't reproducible
+          # FIXME: securedrop-admin isn't reproducible
```
legoktm: IMO non-reproducibility is a blocker to move forward, but it's literally just the bytecode that's unreproducible: https://gist.github.com/legoktm/fc213ffb0dbe20385e50441aaf544a28

If you look in the current Debian packaging we already have an override to strip all the bytecode, we should do the same here.

legoktm (changes requested):
Just did a pass on the Debian packaging.

As a general note I do think it'll be worth extracting out some of the broader fixes from this PR and merging them separately and ahead of time, like the `UBUNTU_VERSION` -> `OS_VERSION` stuff.

micahflee:
I've hit blocks doing some of the client-side write work this week. So now that removing focal has landed, I'm going to spend some time rebasing this PR.

micahflee:
With the help of Claude Code, I have rebased develop into this branch. I'm also currently working on getting all of the Github Actions checks to pass. Rather than force pushing (in case we want to move forward in a different way), I'm pushing my changes to `admin-deb-rebased`: https://github.com/freedomofpress/securedrop/tree/admin-deb-rebased

My plan is, once I succeed in having all of the checks passing, I will do manual testing of admin tools.

zenmonkeykstop:
Got wires crossed and did a manual rebase while the Claude one was in progress. Manual looks *slightly* better so I've force-pushed it. Next up:
- double checking stuff still works under trixie (Tails 7 or Qubes/debian-13-xfce)
- resurrecting enough of the updater to support a 1-pass upgrade for git-based workstations
- more upgrade script.

admin/Makefile
zenmonkeykstop: NM - works fine post-Trixie

micahflee:
I've been working on #7647, but pushing commits into this PR, so when it's ready it will close that issue as well. I've made a lot of progress. Here are the recent changes.

# Migration scripts

The `./securedrop-admin` file in the root of the repo is a new script that only only supports the `setup` and `tailsconfig` commands. This is only meant to facilitate the migration from git-based updated to deb-based updates, and the _only_ time this script should run is when running the GUI updater.

After the migration is complete and the user reboots, there will be no more GUI updater. From that point on, the `./securedrop-admin` script from the git repo isn't executed, and instead `/usr/bin/securedrop-admin` is executed.

The migration scripts include:

- `migrate-to-deb.sh`: this is what's executed when the GUI updater runs `./securedrop-admin setup`. It validates that the user is in Tails and has an existing config. Then it installs the deb package, configures persistence for `~/.config/securedrop` and bind-mounts it, and copies all of the config files from `~/Persistent/securedrop/install-files` into `~/.config/securedrop`.
  - `migrate-install-package.sh`: executed as root by `migrate-to-deb.sh`. It installs `~/Persistent/securedrop-admin.deb` temporarily -- we still need to update this to add the FPF SD repo and install the package from there
  - `migrate-setup-persistence.sh`: executed as root by `migrate-to-deb.sh`. This sets up persistence for `~/.config/securedrop-admin` and then bind-mounts it, so that it can be used right away without requiring a reboot
- `migrate-tailsconfig.sh`: this is what's executed when the GUI updater runs `./securedrop-admin tailsconfig`. This just runs `/usr/bin/securedrop-admin localconfig`, and then pops up an alert telling the user to reboot Tails when it's done

# Changes to ansible playbooks

The GUI installer has logic to ask the user for their sudo password, and then it passes this password into the `ansible-playbook` command after the "SUDO password:" prompt.

The way `securedrop-admin` was written, there was a `@update_check_required` decorator before all of the commands that used ansible to check for updates using apt (instead of `git fetch`), which could be bypassed with `--force`. So this means that if you run `securedrop-admin localconfig` it would run two ansible playbooks:

- for checking for updates
- for configuring Tails

The GUI installer only provided the sudo password for the _first_ time it was prompted for "SUDO password", and when it was prompted a second time, it would just hang and after 2 minutes the GUI installer would timeout.

To fix this, I refactored the ansible playbooks. I made a new `update-check` role that checks for updates (which is skipped if the `skip_update_check` is true), and I added this role to the beginning of all of the playbooks. When `securedrop-admin` subprocesses to ansible, it includes `skip_update_check=true` if `--force` is passed in.

So in the end, it does the same thing, but with only a single ansible playbook getting executed instead of two separate ones, so it succeeds with the GUI installer.

# Testing

Here's how to test it:

- Set up an admin Tails USB that is configured to admin a SecureDrop server (it should have a `~/Persistent/securedrop` folder), and make sure it works to ssh to app and mon, connect to the source interface, etc.
- In Qubes, checkout this `admin-deb` branch and make a deb with `make build-debs-admin`
- Copy `build/trixie/securedrop-admin_2.13.0~rc1+trixie_amd64.deb` to the Tails persistent volume as `~/Persistent/securedrop-admin.deb` (the exact path is important, since it's hardcoded into the migration script)
- In your Tails, create `~/Persistent/changes.diff` that contains this file:

```diff
diff --git a/admin/securedrop_admin/__init__.py b/admin/securedrop_admin/__init__.py
index faac742df..4c9c9ac49 100755
--- a/admin/securedrop_admin/__init__.py
+++ b/admin/securedrop_admin/__init__.py
@@ -762,7 +762,7 @@ def update_check_required(cmd_name: str) -> Callable[[_FuncT], _FuncT]:
         @functools.wraps(func)
         def wrapper(*args: Any, **kwargs: Any) -> Any:
             cli_args = args[0]
-            if cli_args.force:
+            if True:
                 sdlog.info("Skipping update check because --force argument was provided.")
                 return func(*args, **kwargs)
 
@@ -1028,6 +1028,10 @@ def get_release_key_from_keyserver(
 
 def update(args: argparse.Namespace) -> int:
     """Verify, and apply latest SecureDrop workstation update"""
+    subprocess.call(["git", "checkout", "."], cwd=args.root)
+    subprocess.call(["git", "checkout", "admin-deb"], cwd=args.root)
+    subprocess.call(["git", "pull", "origin", "admin-deb"], cwd=args.root)
+    return 0
     sdlog.info("Applying SecureDrop updates...")
 
     update_status, latest_tag = check_for_updates(args)
diff --git a/journalist_gui/journalist_gui/SecureDropUpdater.py b/journalist_gui/journalist_gui/SecureDropUpdater.py
index 75e801803..14882ccfb 100644
--- a/journalist_gui/journalist_gui/SecureDropUpdater.py
+++ b/journalist_gui/journalist_gui/SecureDropUpdater.py
@@ -88,6 +88,12 @@ class UpdateThread(QThread):
         self.failure_reason = ""
 
     def run(self):
+        subprocess.call(["git", "checkout", "."], cwd="/home/amnesia/Persistent/securedrop")
+        subprocess.call(["git", "checkout", "admin-deb"], cwd="/home/amnesia/Persistent/securedrop")
+        subprocess.call(["git", "pull", "origin", "admin-deb"], cwd="/home/amnesia/Persistent/securedrop")
+        self.update_success = True
+        self.signal.emit({"status": True, "output": "", "failure_reason": ""})
+        return
         sdadmin_path = "/home/amnesia/Persistent/securedrop/securedrop-admin"
         update_command = [sdadmin_path, "update"]
         try:
```

- In Tails, checkout the latest git tag and apply that diff:
  ```sh
  cd ~/Persistent/securedrop
  git checkout 2.12.10
  git apply ../changes.diff
  ```
- Before proceeding, make a clone of your Tails USB, and backup the persistent volume. This is required to be able to test this more than once. After you do the migration process, your Tails will be totally changed, so you can just boot from the backup USB, use the "Back Up Persistent Storage" tool to overwrite your other Tails persistence, and then try again, switching between USBs each time you try.

When you're ready (your `~/Persistent/securedrop` must be in the 2.12.10 branch and patched, and you need a `~/Persistent/securedrop-admin.deb`), do the upgrade.

- Click SecureDrop > Check for SecureDrop Updates.
- Switch to the Detailed Update Progress, and click "Update Now"

This will use the patched GUI updater to switch to the `admin-deb` branch, to simulate switching to the latest release branch. Then it will run `/home/amnesia/Persistent/securedrop-admin` commands, which calls the new migration scripts.

It will ask for your password twice, using `pkexec` to run commands as root. The first one is installing `securedrop-admin.deb`, though this will change to instead add the repo and install the package via apt:

IMG_0058

The second time it's setting up `~/.config/securedrop-admin` persistence:

IMG_0059

Then it asks for your password again, in a slightly different way, to run the ansible playbook:

IMG_0061

Then you wait a moment for the `localconfig` playbook to finish, and it shows this popup, telling you to reboot Tails:

IMG_0062

Reboot Tails. To test it, you'll also need to manually install `securedrop-admin.deb`, but we won't need to once we install the package from our apt repo:

```sh
sudo dpkg -i ~/Persistent/securedrop-admin.deb
```

After you connect to Tor, you'll see the new GNOME extension no longer has an option to check for updates:

IMG_0063

You should be able to ssh into app and mon, connect to the journalist interface, etc. But this time, it's pulling the securedrop config from `~/config/securedrop-admin`.

You should also be able to open the console and run `securedrop-admin install`, to reinstall the server. This should ensure that you're using the latest version of `securedrop-admin` from apt before it proceeds (unless you use `--force`).

~(However, I just discovered that the journalist key and the OSSEC key can have arbitrary filenames, and the migration script isn't copying them, so I'm going to fix that too.)~ (fixed)

# Next steps: install from apt repo

Before this is done, we'll need to make the migration scripts add the FPF apt repo and install `securedrop-admin` from it. It's a chicken and egg problem though, since we won't have the `securedrop-admin` deb package made an in an apt repo until this PR is merged.

So I think the next step is to create a deb from _this branch_ and publish it to the `apt-test.freedom.press` apt repo. I don't know how to go about doing that.

Then, I can update this branch to install the test repo and then install `securedrop-admin` from it.

Once we have confirmed that that works, we can switch the migration script to add the `apt.freedom.press` repo. It's unclear exactly how this might conflict with users who also have the dangerzone `apt.freedom.press` repo installed (called `dangerzone.sources`), so maybe this should actually just follow the dangerzone instructions and keep the source `dangerzone.sources`.

micahflee:
Using the GUI updater in Tails to migrate from git-based updates to deb-based updates:

https://github.com/user-attachments/assets/d304b60e-6276-4a8d-851e-6375cbfeeeae

micahflee:
For the fresh install scenario, I think we should provide instructions for adding the securedrop APT repository rather than cloning the git repo, verifying a tag, and checking it out. If we can get Tails to bundle the securedrop source like they do with dangerzone, this will be simple. But until then, it's a little more difficult.

There are [two methods](https://gitlab.tails.boum.org/tails/tails/-/issues/17510#note_233659) for adding third-party APT repos to Tails with the PGP key.

- The more complicated method is using the `.list` format and including the key in `/etc/apt/trusted.gpg.d/`, but this requires adding extra lines to persist the key, and Tails doesn't want to document this method.
- The simpler method is using the new `.sources` format, which bundles the key in the sources file, and only requires persisting that one file. This is the recommended method.

For the simple method, we could provide instructions for the user to add the securedrop test repo by creating `/live/persistence/TailsData_unlocked/apt-sources.list.d/securedrop-test.sources` and copying and pasting this:

```
Types: deb
URIs: tor+https://apt-test.freedom.press/
Suites: trixie
Components: main
Signed-By:
  -----BEGIN PGP PUBLIC KEY BLOCK-----
  
  mQINBGYPK+sBEAC/BBuTjTZFARM45uNQqFdP5X7ZwdyPiyurOoncIPDvP5l2Qgza
  DP4+Xe/RFbCs2JyQLQW4PFUi+3kDwQIXsYDgAK2bpxdwPAQPa0g4kaY0gOu4a6O5
  EI/rw0yEsRbsDkCmxpla4iM7QghqoM5sxq7pA326QS0vr8Di0LpVO+Niy9S3LgpU
  qknW7zT+J3Lnfr9GSstd2XWetOd8pPfnXEySiOctzMbIwtp4y32xczZSLaQM1pa1
  NhNJIcb5NL6FzWcgoIaqJOCnujU0uytQBdKw4D7f9WQwFxWwLL8PMyqh0/j/zOy7
  Pl/GOTrL1txRJpj3clbTegWNs5KgDZwqk1qszJVwE5bQOMKOwRHSZfaQ+h+AQALa
  II3pDmRe3demZ2zbzmJUiDDx8a8L+C+rnoVK3znsxUo524qWR/YO3Zc1rU/DAH8y
  uBkf4SCtWKp+vzsqtHB2O0qPn7GMcS40YJlp5wPuyO8CZw5+kcPYbMhfBwDt2Z+P
  JMq34K1Lk2PUCRJkhigMptddiLhYUBKnhKFu7pTWEO963NGz/9rhZBETppjhpX6V
  7KGOFGN44290PFDQUEpwLvO1hXPVOXU+dVr9t782ahXoZV3jcN43TgLirO7DHDZ5
  lEBcRmcrjx6+1X0ePNtLwlB3R0TociNhGkJNLsNJt/PBl8BJ7C/wQdKoAQARAQAB
  tDFTZWN1cmVEcm9wIFRFU1RJTkcga2V5IDxzZWN1cmVkcm9wQGZyZWVkb20ucHJl
  c3M+iQJOBBMBCgA4FiEEgxJ/aLq7BPP+mmmqVF6UUD+rZasFAmYPK+sCGwMFCwkI
  BwIGFQoJCAsCBBYCAwECHgECF4AACgkQVF6UUD+rZaubIQ/9EMd/D3RUAxPYxvoD
  2gdByXKaEq9e8LSNul1Xfe0Jwv7cULaIdXjIZZ6kjwVLa3qDGcGZfo8tqVL1cTmm
  VcjRZlSZw4KnvGt94mDTAlZDOFcw8Os06JRyUIEMNzZknapbTgSDQUGUh98jJeFL
  3qzy3ig1K2uM+2EewD1SwxYVYC1FvM07L+Wju4Qhko/hzuZokenusO6OBny9Mv1T
  1PKzo7Z1c3zO8K7mvzyL0tVbWA8Sx8yd5aONO4IpUSojJPV/I4dSIPVkPCN3Kxb+
  l9r2AjsNgxmVrcvjg8NEfSUjh7HoSnI2mKfa6S3P/6V66zybp20x3wtsPKR816AC
  EOkdD/sw4l/LOKJXDaca04nUjl3HvY0ie6d4F7iVtKCRG3ImsWQq//9kTiZ4lUQF
  QwICpWnngMAl+WTuei25fTXQ/dzOR1Ec/HfTy1YVrnQiNq51Rsr0reAJW8gGdzeq
  R9iyno4ZZ+QQPS5sA5/gNUlljH0n+1qL1gLiT90QnTdBngZaa8QBs3wMSdZRXekh
  gXdAsXuwrkpfo+Wg25RcaiT8Kvzr3ZsKiW8JpoSuH3v2/efDlJMhg21X6XTXtBsC
  gQw36lnk69mhccMK01uiITuzAKBfqUYFUmz3kwP3s1v6FKnUnbesdkH0NrgKiqZ9
  gPJ7a1Wk7eA2TbLFgkYkMl9qW3m5Ag0EZg8r6wEQALZPLEedFUWcVZV6Gyvzby5P
  jWM4owsHwi8jkjoHLPvSVMp29DbE+WLxf0bZV5aY5VXEr8bRu4ZZqlJhGrTkzBGO
  GyNxDwfMatVNlHWM7IrG+hlNj4ZVPpbRXO+wMzh2uLE037R1kt+jQKDYLKSr7f3c
  tKAueg6v/p14ion89F9DMfEgJ3B0Tzr11BG/CHRXfkgkXQJBqmtAZzZhQ5+hmFdD
  rWK9hRaPajUed2xdr6JtCGWWFTJOmKXhNZ49KM6nf4eo4+1IJLRXuzTW7U8bcNRk
  WZ0R38AHhlfxeUvAMJUo5NTPwYHewZV863tDRIRBqCYRw2CueAY06+6GDVwC+chA
  Lc7mFgIvr1BfndK4stoAM0FAEHrPXvTKwJdgTBGC5IREET395976uYM2dsykrBb7
  9fR09fkdco3JI+pZXo3cVwkbQ4lZsqyxWqwGXLKusHd8xZfAtF1r1eYq8mONAoTx
  4iYZX3XizUAYGu0OfH+ywpQVV/+7MI6Zg85PHWFfEdNzGpKELnJdJ1vUR8hmFjCG
  D8jTDEh2ejscUjn+1749gn7ucUUl09ezzNUSORnduopbzkFVE2rjf9hE8DedMabb
  NVNILXm6or6XFaYaBFFRkRYumMrsjJb76rzENJzHwB21bwyRadedw2WN1emC+a3/
  gbYtDrjejMKhB/r7Xw7NABEBAAGJAjYEGAEKACAWIQSDEn9oursE8/6aaapUXpRQ
  P6tlqwUCZg8r6wIbDAAKCRBUXpRQP6tlqy/VEACbJymSzR7CxQ2fsryFufaejp7j
  7tDgnplB+5Jfcy5EFlnIPM3C/k7DjXssveg+L40U1dm/jKrI7b/qKOJI9noLkEpr
  8ckoo5sMLIf0vWjmjv/J06q/OkSqG9Gu5spGuLe0v1JODwy6XIc0z8EZb6Ib0XpN
  runJXiMhrLAYHE8DTJcVPXKOxWkxtDM1FfkCvEhNIbI3Scn1nOO3lIVgqxqxAbpk
  xaZfO+F7Nr5w7WBPPVXTrK933sBJ4eTBGrpaLn14Li/hA1+dWeBLmVT7T1TYyU66
  HqXaYr/sPVCgj6I0Of5/ZuM9/LqVhAOA3AAJHiwPmUlIJIRFQIXvJiFdEOM2sEsA
  fX4MdOsUp87+GuCGLCaNrkwTeFlFv1QO4SnyyuzknyBw/mKrErG0QVZGRm16w+Bl
  voaNdSeizKhpCfBeHqcCl0p5uxgmcB1Z1KhvxpfulZe5meZebs7R90r1CRzagYpZ
  tJzec10wTUb6+d26pbYJaeyEHfNRtJZY/6hxV8PBD8S8JlO/jRVlwIP65X9mKK6Q
  f1tw7zvE15eYsdUtF+aDG4CaaKYk/uJmNpaMxRGTGqkiz5P8jbQRzb0KumhfQzOF
  tzEqxSSI/f+SqNr8OEWjY0oRbCLJKWs2URzEhk5GvgS4hLCVZlvrAUXSm9Wvdmjw
  F4VrJxJS72cAX4+vPw==
  =szwx
  -----END PGP PUBLIC KEY BLOCK-----
```

I've confirmed that this does indeed work. The problem is, since the public key is bundled, and also since it's all indented, there's no easy way for the user to verify the fingerprint. We could have them verify the fingerprint manually and then use `sed` or something to build this file, but that gets really confusing with a lot of commands to run.

In short: if the user can trust https://docs.securedrop.org/ and copy and paste the `.sources` file, this is definitely the way to go. If we want the user to manually verify the PGP fingerprint, this gets complicated. Any suggestions on which we should go with?

Also, on a fresh install, there are various things that need to be done which the migration script handles, like making `~/.config/securedrop-admin/` persistent. I think this should all be handled in the `postinst` script of the debian package, taking care to ensure that it works properly in either Tails or Qubes.

zenmonkeykstop:
An alternative might be to have a installer script that's downloaded, hash-checked (and possibly signature-checked), and then run. so:

```
curl -O sd-bootstrap.sh https://get.securedrop.org/install.sh  # hostname optional for fun, command grabs but does not run the script
shasum -a 256 sd-bootstrap.sh  # compare output to string in docs or on the securedrop.org site
sudo bash sd-bootstrap.sh  # does the persistence setup and .sources install, likely needs sudo :(
```

alternatively the more belt-and suspendersy version would involve:
- grab the release key from a keyserver
- grab the script from wherever we post it
- grab a .sha256 file and a signature from a separate location we control
- verify the signature on the sha256 file
- verify the hash in the sha256 file matches the script's

There are probably other permutations of hashes and signatures etc, but the core idea would be grab an installer script rather than the whole repo, and verify it before use.

I'm more on the fence about the `postint` stuff, some of it needs to be set up for the apt repo config first. But it could probably test for Tails and check if the persistent config is there updating it if not.

legoktm:
I would prefer that we have, as a requirement, that the user (or system) performs a PGP signature check instead of matching checksums or other things.

I would propose two ways forward:

1) have the user continue to fetch the PGP key, verify fingerprint, then clone git, verify tag, run a bootstrap script from the repo that takes care of what we need and installs the package. Treat this as a temporary measure that allows us to move forward now without needing to do a lot of work to figure out a new process for bootstrapping trust.

2) long term, we ask Tails to ship our key embedded in the `.sources` file, similar to how Qubes is doing for us and how Tails is doing for Dangerzone. This would be a time-limited ask since we're phasing out Tails, but if they aren't on board (or we don't want to), we can stick with option 1.

zenmonkeykstop:
If we're confident about phasing out Tails workstations in the near-to-mid term, then I'd take @legoktm's option 1 and just run with that, given this is an issue for new installs only. Also means we can probably repurpose the new migration scripts.

micahflee:
Maybe to start, I will try to write instructions for the `.list` method, with storing the key in `/etc/apt/trusted.gpg.d/` and persisting it. This way we get 1) manual PGP key verification and 2) no need to clone the git repo. If this turns out to be cumbersome, I can re-assess and do the git repo strategy.

micahflee:
> Maybe to start, I will try to write instructions for the `.list` method, with storing the key in `/etc/apt/trusted.gpg.d/` and persisting it. This way we get 1) manual PGP key verification and 2) no need to clone the git repo. If this turns out to be cumbersome, I can re-assess and do the git repo strategy.

Okay yeah, any way I do it, it requires a lot of steps that could easily be scripted. With that in mind, I'll go with the git repo way, where we can actually just include a script.

micahflee:
Okay, I've made some changes to support fresh installs, and I've tested that it works in both Tails and Qubes. I haven't manually tested that I didn't break the migration from git-based to deb-based updating, but I don't suspect that I did.

First, I've added the `./securedrop-admin bootstrap-tails` command to the `securedrop-admin` script in the root of the repo (but not the `securedrop-admin` CLI tool). To get started on Tails, you git clone to securedrop repo, verify the git tag, check it out, and then run the bootstrap-tails command.

Here's a video of bootstrapping Tails for a fresh install:

https://github.com/user-attachments/assets/0ad3acc6-df40-4e74-9806-c6cb32f0755b

Second, I had to refactor the `securedrop-admin` CLI tool to use the `@update_check_required` decorator after all, instead of forcing an update check at the beginning of each ansible playbook. (It turns out, it only worked if your sudo password for Tails and for the remote SD servers are the same password.) The reason I had changed that around to begin with was to deal with how the GUI updater called securedrop-admin. but it shouldn't be an issue anymore because now the migration script kills the GUI updater and does its own thing.

After rebooting, here's a video of starting to configure and install the SecureDrop servers. Note that I manually install a .deb here because `securedrop-admin` in the test apt repo needs to get updated.

https://github.com/user-attachments/assets/5aa9c2db-5858-427f-89aa-29c03a36a220

I think the fresh install work is complete.

admin/configure-tails-persistence.sh
```diff
@@ -0,0 +1,139 @@
+#!/bin/bash
+# Part 2 of migration: Configure Tails persistence and bind-mount
+# This script is executed as root via pkexec
+set -e
+set -o pipefail
+
+# Environment configuration
```
zenmonkeykstop, line 7: This is kindof risky, as developer error during review or release could end up with a package defaulting to the test repo. it would be better to require manual intervention by devs to switch to test mode, possibly via an override file?

zenmonkeykstop (changes requested):
one small change here - I'll add a commit.

zenmonkeykstop:
One problem I'm hitting during testing - the additional software install, including the securedrop-admin package, occasionally fails on startup for various reasons. This is more of an upstream issue, but if it does happen `securedrop-admin` won't be available. We'll likely need at least troubleshooting/workaround docs for this.

admin/migrate-to-deb.sh
```diff
@@ -0,0 +1,124 @@
...
+
+NEW_CONFIG_DIR="$HOME/.config/securedrop-admin"
+
+# Copy site-specific config
+SITE_SPECIFIC_FILE="$OLD_CONFIG_DIR/group_vars/all/site-specific"
+if [[ -f "$SITE_SPECIFIC_FILE" ]]; then
+    cp "$SITE_SPECIFIC_FILE" "$NEW_CONFIG_DIR/"
+    echo "- Migrated: site-specific"
+
+    # Parse site-specific for GPG public key filenames and copy them
```
zenmonkeykstop: If the `site-specific` file is present, that means this is an Admin Workstation and not a Journalist Workstation (which would only have connection configs for the SI and JI) - so the submission key *and* ossec key must be present. If they're not that's a fail state.

admin/migrate-to-deb.sh
```diff
@@ -0,0 +1,124 @@
...
+        echo "! Not found (skipping): $ossec_key (OSSEC GPG public key)"
+    fi
+
+    securedrop_key=$(grep '^securedrop_app_gpg_public_key:' "$SITE_SPECIFIC_FILE" | awk '{print $2}' | tr -d "'\"")
+    if [[ -n "$securedrop_key" && "$securedrop_key" != "''" && -f "$OLD_CONFIG_DIR/$securedrop_key" ]]; then
+        cp "$OLD_CONFIG_DIR/$securedrop_key" "$NEW_CONFIG_DIR/"
+        echo "- Migrated: $securedrop_key (SecureDrop GPG public key)"
+    elif [[ -n "$securedrop_key" && "$securedrop_key" != "''" ]]; then
+        echo "! Not found (skipping): $securedrop_key (SecureDrop GPG public key)"
+    fi
```
zenmonkeykstop, line 119: There is a 3rd optional GPG pubkey used for configuring journalist alerts (this is distinct from the ossec key) - if it's configured and not present it's also a fail state. If it's not present in `site-specific` tho, that's fine.

admin/migrate-to-deb.sh
```diff
@@ -0,0 +1,124 @@
...
+done
+
+# Copy source onion address
+if [[ -f "$OLD_CONFIG_DIR/app-sourcev3-ths" ]]; then
+    cp "$OLD_CONFIG_DIR/app-sourcev3-ths" "$NEW_CONFIG_DIR/"
+    echo "- Migrated: app-sourcev3-ths"
+else
+    echo "! Not found (skipping): app-sourcev3-ths"
+fi
+
```
zenmonkeykstop: Instances with TLS certs will have 3 other files that need to be copied - the https crt, chain cert, and key files. If these values are set in `site-specific` they must be copied over.

admin/migrate-to-deb.sh
```diff
@@ -0,0 +1,124 @@
...
+# Copy Tor v3 keys
+if [[ -f "$OLD_CONFIG_DIR/tor_v3_keys.json" ]]; then
+    cp "$OLD_CONFIG_DIR/tor_v3_keys.json" "$NEW_CONFIG_DIR/"
+    echo "- Migrated: tor_v3_keys.json"
+else
+    echo "! Not found (skipping): tor_v3_keys.json"
+fi
+
+# Copy auth files
+for auth_file in app-journalist.auth_private app-ssh.auth_private mon-ssh.auth_private; do
```
zenmonkeykstop: This could be cleaner:
- for Admin Workstations with SSH-over-tor enabled *all* of these should be present.
- for Admin Workstation with SSH-over-lan enabled *only* the JI file should be present
- for Journalist Workstations *only* the JI file should be present.

zenmonkeykstop (dismissed):
The migration process currently doesn't cover all cases correctly
- for all admin workstation installs, it should fail loudly if there are missing submission or ossec keys, or if the `.auth_private` files found don't match what's expected from the config
- for installs with HTTPS enabled, HTTPS files (cert,chain,key) need to be copied over
- for Journalist Workstations, the site-specific file will not be available.

The `migrate-to-deb.sh` script needs some revisions to catch said cases.

zenmonkeykstop:
**IN PROGRESS NOTES FOR FINAL REVIEW**

# Overview

This PR flips the SecureDrop admin tooling from being deployed via git using a release tag signed with the SecureDrop signing key, to being deployed as a Debian (Trixie) package from the FPF apt repo. It uses generic locations for installation and config files, and so is suitable for use under both Tails 7 and Qubes (`debian-13-xfce` or similar).

Migration scripts are included to convert existing Tails git-based workstations to use the new packaged code, triggered by the existing GUI updater. For fresh installs, a bootstrapping script is available in the code repo, which can be verified using the same release tag mechanism as before.

# Functionality:

## builder
The `securedrop-admin` package is built via a similar process to the `securedrop-*` server packages, but using a separate configuration and changelog:
- the Debian build config lives in `admin/debian/`
- the build image is trixie-based, and created from `admin/Dockerfile`
- new makefile targets `build-debs-admin` and `build-debs-admin-notest` trigger the build, with artifacts written to `build/trixie`

As with the `securedrop-app-code` package, a Python virtualenv is built via debhelper and included in the `securedrop-admin` package. One major difference is that said virtualenv uses binary-only  Python modules from Pypi - with the removal of non-deterministic bytecode, this allows `securedrop-admin` to be reproducible.

## deployment
For production deployments, a new `securedrop-admin` package should be built as needed from a signed release tag, either as part of a server release or independently for a workstation-only release. The package should be deployed to `apt-test.freedom.press` for testing in the `trixie` distro, and then pushed to `apt.freedom.press`. 

Qubes users will pick up changes when the base template where `securedrop-admin` is installed is updated. Tails users will pick up changes on boot when `additional software` is installed (KOG note - I'm not sure that's consistently done?) or when they run a `securedrop-admin` command that includes a background update check. 
## utility changes

Previously, `securedrop-admin`: 
- ran from the root of the checked-out repo
- required supporting debian packages installed and a virtualenv be set up via a `./securedrop-admin setup` bootstrapping step
- used `~/Persistent/securedrop/install_files/ansible-base` as its location for config data,  ansible roles and playbooks, and backups.

Now, instead it uses non-Tails-specific locations - it:
- is installed to `/usr/bin`
- has a prebuilt virtualenv installed under `/usr/share/securedrop-admin/venv`
- has debian package dependencies installed via `apt`.
- expects config data in `$HOME/.config/securedrop-admin`
- installs ansible playbooks and roles to `/usr/share/securedrop-admin/ansible-base`
- writes backups to `./config/securedrop-admin`

The utility is largely unchanged otherwise apart from the following:
- the `setup` action has been removed as it's no longer needed.
- The `tailsconfig` action, run after server installation to configure the local workstation, is now an alias for `localconfig`, which checks whether the action is run under Tails or Qubes, and configures the workstation accordingly.
- the update check run before sensitive commands to ensure users had the latest version of `securedrop-admin` now runs an ansible playbook to check the installed `securedrop-admin` package version and update it, instead of checking for a more recent release tag.

## migration for existing users

Happy-path migration involves the following:

- User runs pre-migration GUI updater
- Updater verifies prod tag and checks out latest release
- the repurposed `./securedrop-admin setup` command run next by the updater runs a wrapper script that:
  - kills the GUI updater
  - configures persistence for the `~/.config/securedrop` directory and FPF apt repo
  - sets up the FPF apt repo
  - installs the `securedrop-admin` package
  - copies the existing install's config data from `~/Persistent/securedrop/...` to `~/.config/securedrop-admin`
  - runs `securedrop-admin localconfig` to set up Tails with the config info from the new location.

Manual updates are straightforward:
- Check out the release tag with the debian packaging changes
- run `./securedrop-admin setup`, which does much the same as above except for not having to kill the GUI updater.

## fresh install

A fresh install involves the following steps:

- clone the securedrop github repo, then verify and check out the release tag
- run `./securedrop-admin tails-bootstrap` - this will:
  - configure persistence for the `~/.config/securedrop` directory and FPF apt repo
  - set up the FPF apt repo
  - install the securedrop-admin apt package
  - prompt the user to allow additional packages to be installed every time
  - prompt the user to reboot.
- After reboot:
  - copy the submission, ossec, and optionally journalist-alert pubkeys and https cert info to `~/.config/securedrop`
  - run `securedrop-admin sdconfig` (note the lack of the `./` - the command is now installed in the default path)
  - run `securedrop-admin install`
  - run `securedrop-admin localconfig`

# development and testing

By default, the migration and fresh install process will use production values for the FPF ap repo (`apt.freedom.press`) and the production signing key. To override this, before starting a migration or installation:

 - create a file `~/securedrop_bootstrap.env` setting the `APT_REPO_URL` and `APT_SIGNING_KEY_FILE` to the appropriate values: for example `apt-test.freedom.press and `apt-test-signing-key.pub` for the test repo, or `apt-qa.freedom.press` and `fpf-signing-key-2021.pub` for the staging repo

Then proceed.

To test locally-built changes, build the `securedrop-admin` package and install it directly.

admin/migrate-to-deb.sh
zenmonkeykstop: (updated)

admin/migrate-to-deb.sh
zenmonkeykstop: (updated)

admin/migrate-to-deb.sh
zenmonkeykstop: (Updated)

admin/configure-tails-persistence.sh
zenmonkeykstop: (updated)

admin/migrate-to-deb.sh
zenmonkeykstop: (resolved)

builder/build-debs-admin.sh
zenmonkeykstop: virtualenv creation was moved to `debian/rules`, other files installed via `debian/install` - though the commands to move them to a working copy still live here.

builder/tests/test_admin_package.py
zenmonkeykstop: trixie now :)

admin/debian/control
zenmonkeykstop: removed.

.github/workflows/build.yml
zenmonkeykstop: package should now be reproducible

zenmonkeykstop (dismissed):
I've taken a final pass through this monster PR and written up some notes as a cheatsheet for future reviewers/ maintainers. In general:
- build is working as expected
- GUI updater and manual migrations work
- fresh installs work
- Tails-specific configuration works
- Package is installable both in Tails and generic Debian 13.

Approving-  but I've made changes too, so passing off to @legoktm for final stamp.

legoktm (dismissed):
Let's do it. I kicked off a staging build, let's make sure that passes before merging.

zenmonkeykstop:
rebased to clear conflicts - @legoktm `stg-admin-deb` should now be passing (with one consistently flaky testinfra test newly skipped), lmk if you're good to merge.

legoktm (approved):
Most recent set of changes LGTM, with the note that cffi is also a production dependency and in general we should keep the develop + prod versions in sync, but we can fix that for the next release and not block this on the diff review.

I synced stg-admin-deb with this PR so it'll show up on this PR, once it's green I think we're good to go!

Issue #7563, opened 2025-05-28
Remove all Focal-specific code from the codebase


eloquence:
Soon after the 1 month grace period (from June 1-30), we should remove all Focal-specific code from the codebase, including any Focal-specific application code, config, provisioning logic, or developer tooling, and any Python 3.8-specific code, configuration or dependencies.

mig5:
Regarding '1 month grace period', thought I would note that the Demo interfaces have just gone down, because https://demo-source.securedrop.org is throwing a 503, I believe [due to the OS being past its EOL date](https://github.com/freedomofpress/securedrop/blob/d3edb81bbbbee0b160b29a5a485c031ea1ee94e7/securedrop/server_os.py#L9-L27). (It's a Focal image)

The Dockerfiles in https://github.com/freedomofpress/securedrop/blob/develop/securedrop/dockerfiles probably need to be updated to Noble.

eloquence:
For the sprint kicking off today, as a lower priority goal, we're focusing on removing Focal-specific dependencies, with further cleanup planned after that.

PR #7616, opened 2025-07-25
chore: ignore Safety 77744 and 77745 in urllib3

cfm:
Per freedomofpress/fpf-misc-resources#66.


eloquence (approved):
LGTM, cross-ref #7563 in case we can un-ignore this at a later point.

Issue #7552, opened 2025-05-14
Simplify "seen" logic in SecureDrop Client by adding an "update_seen" boolean field in the  `sources` table


zenmonkeykstop:
## Description

SecureDrop Client has performance issues with large numbers of sources, in the 500+ range. Based on profiling results (and ignoring network I/O), updating the source list after a server sync takes about 75% of elapsed time, and within that, checking the `seen` status of sources takes over 50%. This manifests in the UX as a noticeable freeze in updates to the source list immediately following a sync.

Currently, seen status for a source is checked by looping through a collection of a source's submissions and replies, checking if each has been seen by any journalist, and if any item has not been seen, returning False, otherwise True. This happens for every source in the list.

Replies always count as seen (because the sending journalist saw them), so only messages and file submissions matter. If a boolean was set to False when a source first submits something, that value could be used, instead of looping through the collection. This would effectively eliminate a large chunk of the GUI freeze, improving user experience for instances with large numbers of sources

Server-side changes would be required to:

- add a new `update_seen` field in the database
- update it to `false` on source submissions
- update it to `true` when journalist seen_by entries are added for files and messages.
- add the field to the relevant endpoints in the journalist API

Client-side changes would be required to:
- drop the `seen` property in the db model
- add the new field in the db and sdk
- deal with possible edge cases where a source submits while their conversation is being viewed.

## How will this impact [SecureDrop users](https://github.com/freedomofpress/securedrop-ux/wiki/Users)?

- Improvement in UX for users on high-volume instances

## How would this affect SecureDrop's [threat model](https://docs.securedrop.org/en/stable/threat_model/threat_model.html)?

The value of the field could be inferred by an adversary with database access just by looking at the seen_by statuses available in the database already - so no change.

eloquence:
It may not be necessary to change the server-side schema and update logic to achieve this with reasonable performance. This SQL query (I had AI help, but tweaked it a bit) seems to run quite efficiently over hundreds of sources and appears to have the intended effect, with `is_read` set to 1 if none of the associated submissions are unseen. With `.timer on` after `loaddata.py` insertion of 350 sources, my results were:

Run Time: real 0.023 user 0.023855 sys 0.000034


```
SELECT
    src.journalist_designation,
    CASE
        WHEN  NOT EXISTS (
                 SELECT 1
                 FROM submissions  AS sub
                 WHERE sub.source_id = src.id
                   /* for this submission there is neither a seen-file nor a seen-message row */
[...]

PR #7622, opened 2025-08-04
API: Add source-level `is_seen` and `has_attachment` properties

### LINKED ISSUE #7621: v2 API should include source-level properties for unread and file attachment status
## Description

During performance optimization of https://github.com/freedomofpress/securedrop-client/pull/2559, we realized it would be valuable to have source-level properties for:

* `is_read` - are all the items in the source's collection marked as read
* `has_attachment` - is any of the items in the source's collection a file

The implementation will look something like https://github.com/freedomofpress/securedrop/issues/7552#issuecomment-3050198384 (and this issue supersedes that one).

legoktm:
- is_seen: indicates whether all submissions in source are marked as seen
- has_attachment: indicates whether source contains any file attachments

While the client can calculate these, let's have the server do it for efficiency, and leave to the client only things that it can do (i.e. operate on decrypted submissions).

This trivial to add to both API v1 and v2, so we'll do both.

Fixes #7621.

## Test plan
n.b. see LLM usage below

* [x] Review the tests and check behavior is accurate. 

## Checklist

This change accounts for:

## LLM usage

I generated this entirely using Claude Code, here's the transcript: https://gist.github.com/legoktm/7b7071708073185a06ed94a77d47d7f1

It wrote too many tests, so I had it cut down and consolidate; I also rewrote the commit message because I didn't like the one it generated. This is probably a better implementation than what I had proposed in https://github.com/freedomofpress/securedrop/issues/7552#issuecomment-3050198384 as it makes it far easier to port to API v1 IMO.

~~I have not finished comprehensively reviewing this, hence why it's in draft mode, but I wanted to publish on the record what Claude generated in case there were flaws/bugs a human needed to manually fix.~~

### DIFF
--- a/securedrop/models.py
+++ b/securedrop/models.py
@@ -111,6 +111,18 @@ class Source(db.Model):
         return collection
 
     @property
+    def is_seen(self) -> bool:
+        """Return True if all submissions in this source's collection are marked as seen."""
+        if not self.submissions:
+            return True
+        return all(submission.seen for submission in self.submissions)
+
+    @property
+    def has_attachment(self) -> bool:
+        """Return True if any submission in this source's collection is a file attachment."""
+        return any(submission.is_file for submission in self.submissions)
+
+    @property
     def fingerprint(self) -> Optional[str]:
         if self.pgp_fingerprint is not None:
             return self.pgp_fingerprint
@@ -144,6 +156,8 @@ class Source(db.Model):
                 "uuid": self.uuid,
                 "journalist_designation": self.journalist_designation,
                 "is_starred": starred,
+                "is_seen": self.is_seen,
+                "has_attachment": self.has_attachment,
                 "last_updated": last_updated,
                 "public_key": self.public_key,
                 "fingerprint": self.fingerprint,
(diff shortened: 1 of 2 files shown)


legoktm:
My only question is whether this should be called "is_read" or "is_seen"

securedrop/models.py
```diff
@@ -110,6 +110,18 @@ def collection(self) -> "List[Union[Submission, Reply]]":
         collection.sort(key=lambda x: int(x.filename.split("-")[0]))
         return collection
 
+    @property
+    def is_read(self) -> bool:
```
cfm: For consistency, both here and in the JSON representations, let's use `is_seen` (or even `has_been_seen` :-) for the roll-up of `Submission.seen`.  That follows the notion of `Seen{File,Message,Reply}` and leaves the "read" notion for `Reply.is_deleted_by_source`, per freedomofpress/securedrop-client#2581.

cfm (changes requested):
This looks good to me!  I appreciate the DRYness of the property-based approach.

I've answered <https://github.com/freedomofpress/securedrop/pull/7622#issuecomment-3155739343> inline.  While I'm in the neighborhood, I'll also diff this conceptually against freedomofpress/securedrop-client#2581.

securedrop/models.py
legoktm: Done.

cfm (approved):
Thanks, @legoktm.

Issue #7537, opened 2025-05-07
Functional tests sometimes get HTTP status code `None` on noble


eloquence:
See example test failure: `test_journalist_uses_index_delete_files_button_modal`
https://github.com/freedomofpress/securedrop/actions/runs/14890073741/job/41819713512?pr=7209#step:4:453

```
>           assert response_journalist_status_code == 200
E           assert None == 200

/home/runner/work/securedrop/securedrop/securedrop/tests/functional/conftest.py:181: AssertionError
```

nathandyer:
@legoktm says this hasn't happened since the fix was applied

PR #7623, opened 2025-08-04
Try increasing functional test startup timeout to help with Ubuntu Noble

legoktm:
Address #7537 by increasing the timeout for functional test app startup from 7.5 seconds to 30 seconds (60 attempts × 0.5 seconds), and add better error messages when apps fail to start.

Refs #7537.

## Test plan

I think we can just merge it and see if it stops happening.

## Checklist

This change accounts for:

## LLM usage

All generated with Claude Code (see https://gist.github.com/legoktm/08a96e8e7959b2c784eb39aaf600f202) but it kind of did the wrong thing so I told it to do something else.


eloquence (approved):
Let's try it!

PR #7624, opened 2025-08-06
refactor: generalize `/index` and `/metadata` to flat collections

### LINKED ISSUE #7620: `/api/v2/index`: revert to flat collections
@legoktm and I discussed this today and agreed that this makes a compelling case for reverting to the original flat schema.

---

I'm a little disappointed to see how complex the client's side of the sync algorithm, specifically `reconcileIndex()`, has to be to accommodate freedomofpress/securedrop#7604's refactoring of the index structure from—

```json
{
  "sources": {
    <uuid>: <version>,
    ...
  },
  "items": {
    <uuid>: <version>,
    ...
  }
}
```

—to—

```json
{
  "sources": {
    <uuid>: {
      "version": <version>,
      "collection": {
        <uuid>: <version>,
        ...
      }
    },
    ...
  }
}
```

With a flat structure, we would have been able to parameterize the reconciliation so it operated identically over `sources`, `items`, etc.  With the nested structure, we have to special-case them, especially for deletion (https://github.com/freedomofpress/securedrop-client/pull/2562/files#r2241229902).  I've flagged this in <https://docs.google.com/document/d/13xyJeJNJ1Gkd1555j03Ds4XhQejBtM2I5KVKvi2V_es/edit?tab=t.0#heading=h.w2k9mijowic2>.

_Originally posted by @cfm in https://github.com/freedomofpress/securedrop-client/pull/2562#pullrequestreview-3069261968_

cfm:
Closes #7620 by generalizing `/index` and `/metadata` to flat collections of sources and items.

Once the index loses the source/collection hierarchy, the client has no way to "hint" the server in what source's collection to find a given item.  There are other index structures that could provide this hint, but it's simpler just to scan the `Submission` and `Reply` tables directly (without any relationships to eager-load) for the set of item UUIDs requested by the client.

As a bonus, the dataclasses provide sufficient structural validation of incoming requests.

## Test plan

- [ ] New request/response schemas look as expected for <https://github.com/freedomofpress/securedrop-client/pull/2562#pullrequestreview-3069261968>:
    - `Index`
    - `MetadataRequest`
    - `MetadataResponse` 

## Checklist

This change accounts for:

### DIFF
--- a/securedrop/journalist_app/api2/__init__.py
+++ b/securedrop/journalist_app/api2/__init__.py
@@ -122,18 +122,24 @@ def metadata() -> Response:
 
     response = MetadataResponse()
 
-    for source in all_sources().filter(Source.uuid.in_(str(uuid) for uuid in requested.sources)):
-        response.sources[source.uuid] = source.to_api_v2()
+    if requested.sources:
+        for source in all_sources().filter(
+            Source.uuid.in_(str(uuid) for uuid in requested.sources)
+        ):
+            response.sources[source.uuid] = source.to_api_v2()
 
-    for item in Submission.query.filter(Submission.uuid.in_(str(uuid) for uuid in requested.items)):
-        response.items[item.uuid] = item.to_api_v2()
+    if requested.items:
+        for item in Submission.query.filter(
+            Submission.uuid.in_(str(uuid) for uuid in requested.items)
+        ):
+            response.items[item.uuid] = item.to_api_v2()
 
-    for item in Reply.query.filter(Reply.uuid.in_(str(uuid) for uuid in requested.items)):
-        if item.uuid in response.items:
-            # Fail if we get unlucky and hit a UUID collision between the
-            # `Submission` and `Reply` tables.  This is vanishingly unlikely,
-            # but SQLite can't enforce uniqueness between them.
-            raise MultipleResultsFound(f"found {item.uuid} in both submissions and replies")
-        response.items[item.uuid] = item.to_api_v2()
+        for item in Reply.query.filter(Reply.uuid.in_(str(uuid) for uuid in requested.items)):
+            if item.uuid in response.items:
+                # Fail if we get unlucky and hit a UUID collision between the
+                # `Submission` and `Reply` tables.  This is vanishingly unlikely,
+                # but SQLite can't enforce uniqueness between them.
+                raise MultipleResultsFound(f"found {item.uuid} in both submissions and replies")
+            response.items[item.uuid] = item.to_api_v2()
 
     return jsonify(asdict(response))
--- a/securedrop/journalist_app/api2/__init__.py
+++ b/securedrop/journalist_app/api2/__init__.py
@@ -1,9 +1,9 @@
 import hashlib
 from dataclasses import asdict, dataclass, field
-from typing import Any, Dict, List, Mapping, NewType, Optional
+from typing import Any, Dict, Mapping, NewType, Optional, Set
 
 from flask import Blueprint, abort, json, jsonify, request
-from models import Source
+from models import Reply, Source, Submission
 from sqlalchemy.inspection import inspect
 from sqlalchemy.orm import Query, joinedload
 from werkzeug.wrappers.response import Response
@@ -48,29 +48,23 @@ ItemUUID = NewType("ItemUUID", str)
 
 
 @dataclass
-class IndexSourceEntry:
-    version: Version
-    collection: Dict[ItemUUID, Version] = field(default_factory=dict)
-
-
-@dataclass
 class Index:
-    sources: Dict[SourceUUID, IndexSourceEntry] = field(default_factory=dict)
+    sources: Dict[SourceUUID, Version] = field(default_factory=dict)
+    items: Dict[ItemUUID, Version] = field(default_factory=dict)
 
 
 @blp.get("/index")
 @blp.get("/index/<string:prefix>")
 def index(prefix: Optional[str] = None) -> Response:
     """
-    By default, return the ETag-versioned ``Index`` consisting of an
-    ``IndexSourceEntry`` for each source and its collection, unless the client
-    provides the ETag of the current index.
+    By default, return the ETag-versioned ``Index`` of all sources and their
+    items, unless the client provides the ETag of the current index.
 
     Given a ``prefix``, return the sub-index of all sources whose UUIDs begin
-    with that prefix, unless the client provides the ETag of the current
-    sub-index for that prefix.  The client MAY choose an arbitrary prefix with
-    each request: e.g., a series of requests with the prefixes ``{0...f}`` will
-    effectively shard the index into 16 shards.
+    with that prefix (and their items), unless the client provides the ETag of
+    the current sub-index for that prefix.  The client MAY choose an arbitrary
+    prefix with each request: e.g., a series of requests with the prefixes
+    ``{0...f}`` will effectively shard the index into 16 shards.
     """
     index = Index()
 
@@ -85,12 +79,9 @@ def index(prefix: Optional[str] = None) -> Response:
 
     for source in query.all():
         all_source_metadata = source.to_api_v2()
-        source_entry = IndexSourceEntry(
-            version=json_version(all_source_metadata["source"]),
-        )
+        index.sources[source.uuid] = json_version(all_source_metadata)
         for uuid, item in all_source_metadata["collection"].items():
-            source_entry.collection[uuid] = json_version(item)
-        index.sources[source.uuid] = source_entry
+            index.items[uuid] = json_version(item)
 
     version = json_version(asdict(index))
     response = jsonify(asdict(index))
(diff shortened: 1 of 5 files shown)


securedrop/journalist_app/api2/__init__.py
```diff
@@ -102,71 +93,43 @@ def index(prefix: Optional[str] = None) -> Response:
...
 
-@dataclass
-class SourcesResponse:
-    sources: Dict[SourceUUID, SourceEntry] = field(default_factory=dict)
-
-
-@blp.post("/sources")
-def sources() -> Response:
+@blp.post("/metadata")
+@blp.post("/sources")  # deprecated
```
legoktm: I think we can get rid of this now

securedrop/journalist_app/api2/__init__.py
```diff
@@ -102,71 +93,43 @@ def index(prefix: Optional[str] = None) -> Response:
...
-        for uuid, item in all_source_metadata["collection"].items():
-            if want_full or uuid in partial:
-                source_entry.collection[uuid] = item
-        if want_full:
-            source_entry.info = all_source_metadata["source"]
-        response.sources[source.uuid] = source_entry
+    for source in all_sources().filter(Source.uuid.in_(str(uuid) for uuid in requested.sources)):
+        response.sources[source.uuid] = source.to_api_v2()
+
+    for item in Submission.query.filter(Submission.uuid.in_(str(uuid) for uuid in requested.items)):
```
legoktm: I think it's worth adding a comment that, while it's technically possible for there to be a UUID collision across Submission/Reply tables, we don't consider it practically possible and therefore aren't guarding against it.

securedrop/tests/test_journalist_api2.py
```diff
@@ -166,40 +167,33 @@ def test_sources(journalist_app, test_files, journalist_api_token):
         source_versions = index.json["sources"][uuid]
 
         # Get the full source
-        with assert_query_count(1):
+        with assert_query_count(3):
```
legoktm: I think we can live with O(3) instead of O(1) :)

legoktm (commented):
LGTM, just a few minor suggestions. Also would like Vicki to look at it for the client-side implications before approving.

securedrop/journalist_app/api2/__init__.py
cfm: Done in e222bd1856a167336a41880e2680672fa7e30a20.

securedrop/journalist_app/api2/__init__.py
cfm: @legoktm and I discussed this edge case further and decided to treat it like the fault it is.  That's done (and tested) in b1023f2390873080720a77a698d4fff83b08a857.

securedrop/journalist_app/api2/__init__.py
```diff
@@ -47,30 +48,24 @@ def json_version(d: Mapping) -> Version:
...
-class IndexSourceEntry:
-    version: Version
-    collection: Dict[ItemUUID, Version] = field(default_factory=dict)
-
-
 @dataclass
 class Index:
-    sources: Dict[SourceUUID, IndexSourceEntry] = field(default_factory=dict)
+    sources: Dict[SourceUUID, Version] = field(default_factory=dict)
+    items: Dict[ItemUUID, Version] = field(default_factory=dict)
```
vickiniu, line 54: by flattening the `items` here, I think we lose a bit of granularity on the sync in terms of being able to request only the items that need to be updated for a given `source` with version mismatch. the client will now only be able to request all `items` from a source, since we can't map between source:item in the index. and in that case, I think there might not be use in sending all the item versions, instead of just the source versions.

an alternative might be that the server sends its `Index` which is just `{source_id: source_version, ...}`, and then the client can respond with a request to sync that contains `{source_id: [item_version, ...], source_id: null}` where the array of item versions is the list of items that the client knows about. then the server can fetch the items to be upserted and deleted and respond to the client. wdyt?

cfm: Clarified offline:  This is absolutely the goal!  The `Item.source_uuid` → `Source` relationship is a property of the `Item` schema on either end, but the `Index` and `Metadata{Request,Response}` structures are designed to be agnostic to the underlying models and any relationships between models.  In theory, a client can request _any_ set of `sources` and _any_ set of `items` (whether or not they belong to those `sources`).  In practice, a client will indeed request `items` belonging to `sources` shown to be new or changed in the index.

securedrop/journalist_app/api2/__init__.py
```diff
@@ -102,71 +94,47 @@ def index(prefix: Optional[str] = None) -> Response:
...
-        cls = FullSourceEntry if want_full else SourceEntry
-        source_entry = cls()
-        for uuid, item in all_source_metadata["collection"].items():
-            if want_full or uuid in partial:
-                source_entry.collection[uuid] = item
-        if want_full:
-            source_entry.info = all_source_metadata["source"]
-        response.sources[source.uuid] = source_entry
+    for source in all_sources().filter(Source.uuid.in_(str(uuid) for uuid in requested.sources)):
+        response.sources[source.uuid] = source.to_api_v2()
```
vickiniu: question: now that we're flattening the `sources` + `items`, do we still need to send back `collection: ...` with the source metadata here? I think we can exclude this now so that `source.to_api_v2()` returns only the source metadata https://github.com/freedomofpress/securedrop/blob/develop/securedrop/models.py#L139-L140

securedrop/journalist_app/api2/__init__.py
legoktm: Hmm, will this mean we won't actually know when a source is ready for display if we only know which items a source has once all the items have finished downloading?

Edit: we could rectify that by having the source object include a list of item UUIDs as part of its metadata.

vickiniu: so I think right now, the source metadata does include `collection: ...` which has all the items 

if we're going to flatten the responses, I'd actually prefer that we exclude any item information from the source metadata (see: https://github.com/freedomofpress/securedrop/pull/7624#discussion_r2260489152). that does mean that we could sync a source before we sync all of the items in that source, but I think that's ultimately ok! also, I think the sync operations will end up basically executing in one big batch from the frontend's perspective, but even if not I don't think we need to couple source + item sync

vickiniu (commented):
overall, this is looking good! the interface definitely works out, was able to adapt the client sync to match :) the implementation is way simpler now, which is exciting!

the main comment I have is if we want to keep the source metadata as 
```
{
   "source": <...>,
   "collection": [...]
}
```

my vote would be to remove the `collection`/items information for simplicity and to really decouple the source + item entities.

securedrop/journalist_app/api2/__init__.py
cfm: Good point!  Done in 82b0f9ec594fc382bb9c258d767f912406f2a705.

securedrop/journalist_app/api2/__init__.py
cfm: @legoktm in <https://github.com/freedomofpress/securedrop/pull/7624#discussion_r2260615533>:

> Hmm, will this mean we won't actually know when a source is ready for display if we only know which items a source has once all the items have finished downloading?
> 
> Edit: we could rectify that by having the source object include a list of item UUIDs as part of its metadata.

You know, this is what I'd originally intended in [`VersionedCollection.collection_index`](https://github.com/freedomofpress/securedrop/blob/7113f0e4c7a2abe6a1142fbede18f704d0d5a94a/securedrop/journalist_app/api2/__init__.py#L82-L90).  But I think you're right here, @vickiniu, that we might as well go all the way and "trust" this emergent property.  See 82b0f9ec594fc382bb9c258d767f912406f2a705 for more commentary.

cfm:
Rebased from `develop` after #7622.

vickiniu (dismissed):
nice, this lgtm!!

securedrop/tests/test_journalist_api2.py
```diff
@@ -166,41 +168,48 @@ def test_sources(journalist_app, test_files, journalist_api_token):
...
-            assert json_version(item) == source_versions["collection"][item_uuid]
-
-        # Get a partial source
-        item_uuid = list(source["collection"].keys())[0]
-        with assert_query_count(1):
+        assert json_version(source) == source_versions
+
+        # Get an item
+        item_uuid = test_files["submissions"][0].uuid
+        with assert_query_count(3):
```
legoktm: CI is saying this is making 4 queries...

legoktm: Hmmm, passes for me locally.

legoktm: annoying, https://github.com/freedomofpress/securedrop/actions/runs/16812632262/job/47621526510#step:4:1166 elides the queries so you can't see what they actually are.

legoktm: I've triggered a re-run for now but since it failed across both the push and pull_request jobs I suspect it's a real failure. Still confused as to why I can't reproduce it locally.

Since this is blocking other stuff I would be okay with commenting out the assertion and leaving it as a FIXME if we can't solve it today.

securedrop/journalist_app/api2/__init__.py
```diff
@@ -102,71 +93,53 @@ def index(prefix: Optional[str] = None) -> Response:
...
+        ):
+            response.items[item.uuid] = item.to_api_v2()
+
+        for item in Reply.query.filter(Reply.uuid.in_(str(uuid) for uuid in requested.items)):
+            if item.uuid in response.items:
+                # Fail if we get unlucky and hit a UUID collision between the
+                # `Submission` and `Reply` tables.  This is vanishingly unlikely,
+                # but SQLite can't enforce uniqueness between them.
+                raise MultipleResultsFound(f"found {item.uuid} in both submissions and replies")
+            response.items[item.uuid] = item.to_api_v2()
```
cfm, line 143: Do want to eager-load these too?  (Could be a quick follow-up optimization.)

securedrop/tests/test_journalist_api2.py
cfm: 224a7840a5a94b82c2aa46e4e3ad6f7e3761d0bb should align reality with our intuitions (and `assert_query_count()`'s expectations).  But it does raise the question of <https://github.com/freedomofpress/securedrop/pull/7624/files#r2261328894>.

securedrop/journalist_app/api2/__init__.py
legoktm: Yeah, if it isn't already, let's do it as a follow-up.

securedrop/tests/test_journalist_api2.py
legoktm: Thanks, though I don't get why it was passing for me locally. In any case, I'll merge once CI passes this time :)

legoktm (approved):
Pending CI passing

securedrop/journalist_app/api2/__init__.py
cfm: Filed in #7628.

PR #7625, opened 2025-08-06
Add dependency review workflow

eloquence:
Towards https://github.com/freedomofpress/securedrop-dev/issues/14. This does not directly address the redundancy (it adds another check), but it covers the ground that our per-repo `safety` checks cover so we can remove them without regret.

If this looks good, I will prepare additional similar PRs to add the action, and to remove the per-repo `safety` checks, as appropriate.

## Test plan
- Review the description of the action itself: https://github.com/actions/dependency-review-action
- Review https://github.com/freedomofpress/securedrop/actions/runs/16791402650 as an example of a failed run (commit since dropped)

## Checklist

This change accounts for:

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,0 +1,14 @@
+name: 'Dependency Review'
+on: [pull_request]
+
+permissions:
+  contents: read
+
+jobs:
+  dependency-review:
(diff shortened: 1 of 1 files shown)


legoktm (approved):
LGTM, thanks for setting this up.

PR #7626, opened 2025-08-07
feat: include journalist accounts in sync

### LINKED ISSUE #7618: `/api/v2/metadata`: include journalist accounts
* After #7604 
* After #7620
* Towards freedomofpress/securedrop-client#2560

The v2 Journalist API's sync endpoints should return `Journalist` user accounts.  Unlike `Source` accounts and collections, these records are tiny and change rarely, but we should return them with the same versioning semantics so that we treat them as first-class objects and avoid special-casing them.

1. [x] Make the v2 `/sources` endpoint also available at `/metadata`
2. [x] Add a `Journalist.to_api_v2()` method
3. [x] `/index`: Add a `journalists` key, keyed by `Journalist.uuid`
4. [x] `/metadata`: Add a `journalists` key, keyed by `Journalist.uuid`

cfm:
- [x] Depends on #7624

Closes #7618 by returning `Journalist` accounts from the `/index` and `/metadata` endpoints.

Notably, after the slight refactoring in b3e42274f6ac0b2b27c31f267127335bf1fdce93, 7e6cfc21e27fb9f69a1621c9279d557abed88608 is representative of the marginal effort involved in adding new objects to the v2 sync protocol.

## Test plan

- [x] `curl localhost:8081/api/v2/index` includes a `{<uuid>: <version>}` dictionary of journalist accounts.
- [x] `curl -H "Content-Type: application/json" -d '{"journalists": ["<uuid>"]}' localhost:8081/api/v2/metadata` returns the expected metadata for that journalist account.

## Checklist

This change accounts for:

### DIFF
--- a/securedrop/journalist_app/api2/__init__.py
+++ b/securedrop/journalist_app/api2/__init__.py
@@ -50,35 +50,42 @@ ItemUUID = NewType("ItemUUID", str)
 
 @dataclass
 class Index:
+    # Source metadata, optionally filtered by `source_prefix`:
     sources: Dict[SourceUUID, Version] = field(default_factory=dict)
     items: Dict[ItemUUID, Version] = field(default_factory=dict)
 
+    # Non-source metadata (always returned):
+
 
 @blp.get("/index")
-@blp.get("/index/<string:prefix>")
-def index(prefix: Optional[str] = None) -> Response:
+@blp.get("/index/<string:source_prefix>")
+def index(source_prefix: Optional[str] = None) -> Response:
     """
-    By default, return the ETag-versioned ``Index`` of all sources and their
-    items, unless the client provides the ETag of the current index.
+    By default, return the ETag-versioned ``Index`` of all metadata unless the
+    client provides the ETag of the current index.
 
-    Given a ``prefix``, return the sub-index of all sources whose UUIDs begin
-    with that prefix (and their items), unless the client provides the ETag of
-    the current sub-index for that prefix.  The client MAY choose an arbitrary
-    prefix with each request: e.g., a series of requests with the prefixes
-    ``{0...f}`` will effectively shard the index into 16 shards.
+    Given a ``source_prefix``, return the sub-index of source metadata for all
+    sources whose UUIDs begin with that prefix, plus all non-source metadata,
+    unless the client provides the ETag of the current sub-index for that
+    prefix.  The client MAY choose an arbitrary prefix with each request: e.g.,
+    a series of requests with the prefixes ``{0...f}`` will effectively shard
+    the source index into 16 shards.  (Non-source metadata is not filtered by
+    the prefix and is always returned.)
     """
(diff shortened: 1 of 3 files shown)


securedrop/models.py
```diff
@@ -750,6 +750,9 @@ def to_api_v1(self, all_info: bool = True) -> Dict[str, Any]:
 
         return json_user
 
+    def to_api_v2(self, *args, **kwargs) -> Dict[str, Any]:  # type: ignore
+        return self.to_api_v1(*args, **kwargs)
```
cfm: I did this lazily on purpose, to demonstrate the case where we want to use the v1 JSON representation verbatim in the v2 API.  I welcome review feedback on whether this is a smell, especially if mypy needs to be told to ignore the passthrough `*args, **kwargs`.  :-)

cfm: However, this has the side effect of using the default `all_info=True`, which is _not_ the behavior of the v1 API:

https://github.com/freedomofpress/securedrop/blob/5cfe25bfa6e9858e399100e040fb8b47214b58aa/securedrop/journalist_app/api.py#L360

What is the value of protecting this information from authenticated journalists?

legoktm: I would prefer we just duplicate the output we want in the v2 function instead of wrapping around v1, to make it easier to understand what's happening and also make it easier for the next person who wants to modify v2 and not v1 (or vice versa).

Re: `all_info=True`, I'm unsure if we want journalists to be able to have access to e.g a list of inactive admin accounts.

I would recommend that we side-step that though and only include the information in v2 that we actually need in the app, and I don't believe we need that right now.

legoktm (changes requested):
I appreciate how clean it was to integrate this in!

cfm:
Rebased from #7624.

securedrop/models.py
cfm: Done in 343b3a13ab69dcb4beff7ed86a98447be26c0e4c.

securedrop/journalist_app/api2/__init__.py
```diff
@@ -136,4 +156,9 @@ def metadata() -> Response:
             raise MultipleResultsFound(f"found {item.uuid} in both submissions and replies")
         response.items[item.uuid] = item.to_api_v2()
 
+    for journalist in Journalist.query.filter(
```
cfm: This will need to be conditional if we keep 224a7840a5a94b82c2aa46e4e3ad6f7e3761d0bb.

legoktm: Yes please. And then if you can rebase this on develop (once the other PR passes the merge queue)

cfm: Done in 68993138fbdecbe48b4f3259ec1037ebcd479f1a.

legoktm (approved):
Test plan checks out, LGTM

PR #7627, opened 2025-08-07
Don't run safety in CI and remove duplicative ignore-list

eloquence:
Towards https://github.com/freedomofpress/securedrop-dev/issues/14

## Test plan
- Verify that the `semgrep` test in `static-analysis` still works as expected
- Verify that `safety` has been correctly removed from all requirements
- Verify that `safety` no longer runs and what remains is what we want to keep

## Checklist

This change accounts for:


Makefile
```diff
@@ -127,85 +127,16 @@ zizmor:  ## Lint GitHub Actions workflows.
 .PHONY: lint
 lint: check-ruff ansible-config-lint app-lint html-lint shellcheck typelint yamllint zizmor check-strings check-supported-locales check-desktop-files ## Runs all lint checks
 
+# This Makefile target is provided for convenience. The project relies on
+# GitHub Dependabot alerts and centralized safety checks as the primary method
+# to generate vulnerability alerts.
```
eloquence: On balance, I figured there's no harm in keeping this around for quick local checks if there's anything we might be overlooking by other methods, stuff that should no longer be ignored, etc.

.github/workflows/security.yml
```diff
@@ -19,21 +19,18 @@ jobs:
...
+  static-analysis:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
         with:
           persist-credentials: false
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.8'
+          python-version: '3.11'
```
eloquence: 3.8 is EOL :). Seems to be just as happy with 3.11.

legoktm: Let's do 3.12 since that's what all the dependencies are aiming for given noble.

Makefile
legoktm: I personally would like to just get rid of it entirely to reduce confusion and make it clear that safety isn't used at the repo level...since we already require people to install it manually I don't think having the `make safety` wrapper provides much benefit if there is a legit reason for people to run it locally.

eloquence: @legoktm In that case, do you want me to remove it from here and the pip-compiled version as well?

https://github.com/freedomofpress/securedrop/blob/6c2475402eb814971cf649a2b04d8619fec8b277/securedrop/requirements/develop-requirements.in#L35

legoktm: Yes please.

eloquence: Done.

eloquence:
(Because it renames a required check, this will require infra involvement to be mergeable once it's been approved; the required check will always remain yellow otherwise.)

legoktm (approved):
LGTM, will ask infra to rename the job.

PR #7629, opened 2025-08-08
refactor: `eager_query()` wraps SQLAlchemy queries for consistent eager loading

### LINKED ISSUE #7628: `/api/v2/*`: eager-load consistently
@legoktm's insight in #7604 is that the v2 Journalist API should prefer expensive eager queries over _apparently_ cheaper lazy queries, because SecureDrop Application Servers are sufficiently overprovisioned that we might as well load everything into RAM every chance we get.

#7624 preserves this strategy in `/index` for top-down queries of all `Source`s and their collections but violates it in `/metadata` for filtered queries out of a given `MetadataRequest`.[^1]  We should standardize eager loading across the v2 API endpoints.  This change will be invisible to consumers (except in the optimization over large datasets).  A few tests' `assert_query_count()` expectations may change, but they should become more intuitive.

[^1]: https://github.com/freedomofpress/securedrop/pull/7624#discussion_r2261328894

cfm:
- [x] Depends on #7626

Fixes #7628 by providing `eager_query(<model>)` as an extensible DRY abstraction over chaining SQLAlchemy `joinedload()` calls in `<model>.query.options()`.

## Test plan

- [ ] Visual review.
- [ ] `assert_query_count()` calls in a given test now match what you'd intuitively expect from reading through the corresponding API endpoint.

## Checklist

This change accounts for:

## LLM usage

[ChatGPT and I went back and forth on the DRYest way to set query options for eager loading.](https://chatgpt.com/share/68954813-10c0-800f-ace5-7de89e791de3)

### DIFF
--- a/securedrop/journalist_app/api2/__init__.py
+++ b/securedrop/journalist_app/api2/__init__.py
@@ -1,11 +1,12 @@
 import hashlib
 from dataclasses import asdict, dataclass, field
-from typing import Any, Dict, Mapping, NewType, Optional, Set
+from typing import Any, Dict, Mapping, NewType, Optional, Set, Tuple
 
+from db import db
 from flask import Blueprint, abort, json, jsonify, request
-from models import Journalist, Reply, Source, Submission
+from models import Journalist, Reply, SeenFile, SeenMessage, SeenReply, Source, Submission
 from sqlalchemy.inspection import inspect
-from sqlalchemy.orm import Query, joinedload
+from sqlalchemy.orm import Load, Query, configure_mappers, joinedload
 from sqlalchemy.orm.exc import MultipleResultsFound
 from werkzeug.wrappers.response import Response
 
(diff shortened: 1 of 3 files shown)


cfm:
_The_ DRYest approach would actually be to generalize eager loading at the model level, as we do for `Seen{File,Message,Reply}`, e.g.:

https://github.com/freedomofpress/securedrop/blob/6c2475402eb814971cf649a2b04d8619fec8b277/securedrop/models.py#L845-L849

But, even given the "let the server do all the work" assumption I noted from @legoktm in #7628, I imagine we'd want to profile such a change before making it application-wide.

legoktm:
> I imagine we'd want to profile such a change before making it application-wide.

I did a quick smoke test when I made those changes in the API v2 PR, but my theory/assumption has been that it's all lightweight enough that it won't matter in any negative way but may help positively in some cases.

cfm:
Got it.  I'll prepare that change this week in place of this one.

securedrop/journalist_app/api2/__init__.py
```diff
@@ -14,18 +15,57 @@
...
-    """
+# Used by `{Submission,Reply}_query_options()` when called directly rather than
+# via `Source_query_options()`.
+LOADER_BASE = {
+    Submission: Load(Submission),
+    Reply: Load(Reply),
+}
+
+
+def Submission_query_options(base: Load = LOADER_BASE[Submission]) -> Tuple:
```
legoktm: Since this is a function let's name it in all lowercase (same with the ones below)

securedrop/journalist_app/api2/__init__.py
```diff
@@ -14,18 +15,57 @@
...
+    configure_mappers()
+    return (
+        joinedload(Source.star),
+        *Submission_query_options(joinedload(Source.submissions)),
+        *Reply_query_options(joinedload(Source.replies)),
+    )
+
+
+EAGER_BUNDLES = {
+    Source: lambda: Source_query_options(),
```
legoktm: I don't think you need the lambda, I think something like this should work?

```suggestion
    Source: Source_query_options,
```

securedrop/journalist_app/api2/__init__.py
```diff
@@ -14,18 +15,57 @@
...
+    Submission: lambda: Submission_query_options(),
+    Reply: lambda: Reply_query_options(),
+}
+
+
+def eager_query(model: db.Model) -> Query:
+    """
+    Return ``model.query`` with the appropriate eager-loading options applied.
+    Falls back to plain ``model.query`` if no bundle is registered.
```
legoktm: I don't think we should have a silent fallback; if you use `eager_query()` it should actually do something or loudly complain (i.e. an exception) that it's not going to make a difference so that you know to configure it for that type. Or maybe instead of an exception we can codify it at the type-level?

legoktm (changes requested):
Overall looks good, just a few coding style things.

securedrop/journalist_app/api2/__init__.py
cfm: Fixed in bf39c6e74db1c727d6c6955795d8eb5739dc3732.

securedrop/journalist_app/api2/__init__.py
cfm: Fixed in bf39c6e74db1c727d6c6955795d8eb5739dc3732.

securedrop/journalist_app/api2/__init__.py
cfm: Done, with the caveats noted, in fb0b3e003b5e250107b730916cf180e54aac4c5c—and proven useful by 44419da1060c86a8934231528ffcce9137f8d1bc.  :-)

securedrop/journalist_app/api2/__init__.py
cfm: I realized that the `<model>_query_options()` pattern smelled like class methods, so I've refactored this all into `securedrop.models` in bcdee50f63e1bbae9c245549ddcc2e82c43a1045.  Feel free to pop this commit off if you find it overreaching.

legoktm (approved):
Thanks :)

PR #7634, opened 2025-08-08
Remove Focal-specific code and build/development tooling

### LINKED ISSUE #7563: Remove all Focal-specific code from the codebase

zenmonkeykstop:
Fixes #7563 

WIP, some outstanding tasks include:

- ~removing focal requirements and updating scripts~
- ~adding a qubes-staging-noble staging scenario and removing the focal one~ Removed qubes-staging altogether, as a straw poll of team showed it not to be used.
- cleaning up the testinfra tests, removing focal-specific exceptions
- 
(One thing to bear in mind if anyone picks this up - the upcoming admin deb packaging change uses some shared build tooling across `admin` and `securedrop` - so some logic referencing UBUNTU_VERSION  will likely need to stick around to be updated to reference OS_VERSION instead.)
 
## Test plan

TK


builder/tests/test_ossec_package.py
```diff
@@ -1,15 +1,13 @@
...
 from pathlib import Path
 
 OSSEC_VERSION = "3.6.0"
 
-UBUNTU_VERSION = os.environ.get("UBUNTU_VERSION", "focal")
 SECUREDROP_ROOT = Path(
     subprocess.check_output(["git", "rev-parse", "--show-toplevel"]).decode().strip()
 )
-BUILD_DIRECTORY = SECUREDROP_ROOT / f"build/{UBUNTU_VERSION}"
+BUILD_DIRECTORY = SECUREDROP_ROOT / "build/noble"
```
cfm, line 10: I agree with you in <https://github.com/freedomofpress/securedrop/pull/7634#issue-3305323038>, @zenmonkeykstop: I'd rather not lose parameterization even if we wind up narrowing it (always `focal`) or widening it (`UBUNTU_VERSION` → `OS_VERSION`).

zenmonkeykstop: There's a bit of nuance tho: it makes sense to keep it around for cases where we now  have to consider Debian (Tails/Qubes) packages, but not for cases where it's supporting different Ubuntu builds. I know we say this every 5 years, but Noble is likely the last Ubuntu LTS SD will support given work on SecureDrop Protocol and such.

zenmonkeykstop:
Closing in favour of an `stg-`prefixed branch.

PR #7638, opened 2025-08-12
Add CSP for demo.securedrop.org (landing page)

### LINKED ISSUE #7560: Set content-security-policy for demo.securedrop.org
## Description

```
$ curl -I 'https://demo.securedrop.org/'
HTTP/2 200 
date: Tue, 20 May 2025 22:01:59 GMT
content-type: text/html
content-length: 3997
last-modified: Wed, 27 Apr 2022 16:23:45 GMT
etag: "62696e11-f9d"
accept-ranges: bytes
strict-transport-security: max-age=63072000; includeSubDomains; preload
referrer-policy: same-origin
x-content-type-options: nosniff
x-xss-protection: 1; mode=block
x-frame-options: DENY
cf-cache-status: DYNAMIC
server: cloudflare
cf-ray: 942f1da1ca3c25dc-EWR
```

Should be straightforward to add a CSP header via https://github.com/freedomofpress/securedrop/blob/develop/devops/demo/landing-page/nginx.conf - the inline JS should be moved to a separate file so we can just have `script-src 'self'`

eloquence:
Fixes #7560

## Test plan
1. Spin up `make demo-landing-page`
2. Visit to verify TOTP cycling & clipboard features
3. `curl -I http://localhost:8000` should show:

> Content-Security-Policy: default-src 'self'; img-src 'self' https://securedrop.org; object-src 'none'; frame-ancestors 'none';

## AI disclosure

Used Claude Code to bootstrap things, then hand-tweaked CSP a bit and verified it via https://csp-evaluator.withgoogle.com/

## Checklist

This change accounts for:

### DIFF
--- a/devops/demo/landing-page/nginx.conf
+++ b/devops/demo/landing-page/nginx.conf
@@ -23,6 +23,7 @@ http {
         location / {
             root /opt/nginx/webroot;
             index index.html;
+            add_header Content-Security-Policy "default-src 'self'; img-src 'self' https://securedrop.org; object-src 'none'; frame-ancestors 'none';" always;
         }
     }
 }
--- a/devops/demo/landing-page/nginx.conf
+++ b/devops/demo/landing-page/nginx.conf
@@ -23,7 +23,7 @@ http {
         location / {
             root /opt/nginx/webroot;
             index index.html;
-            add_header Content-Security-Policy "default-src 'self'; img-src 'self' https://securedrop.org; object-src 'none'; frame-ancestors 'none';" always;
+            add_header Content-Security-Policy "default-src 'none'; script-src 'self'; style-src 'self'; font-src 'self'; img-src 'self' https://securedrop.org;" always;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


devops/demo/landing-page/nginx.conf
```diff
@@ -23,6 +23,7 @@ http {
         location / {
             root /opt/nginx/webroot;
             index index.html;
+            add_header Content-Security-Policy "default-src 'self'; img-src 'self' https://securedrop.org; object-src 'none'; frame-ancestors 'none';" always;
```
legoktm: I would suggest having default-src be 'none' and then explicitly setting script-src, style-src, font-src and img-src. That ways we don't need to set other things like object-src, and frame-ancestors to none.

legoktm (changes requested):
Works as expected in my testing, one suggestion on how we express the policy.

devops/demo/landing-page/nginx.conf
eloquence: Makes sense, done.

legoktm (approved):
Thanks!

PR #7654, opened 2025-09-08
Fix OS check ansible task in staging.

zenmonkeykstop:
Adds the necessary facts toggle in the staging playbook for the `prepare-servers` role, so that the OS check for Noble will work there too.

## Test plan
- [ ] CI is passing, in particular the staging job.


legoktm (approved):
Thanks, verified there are no other playbooks that could need updating.

Issue #7659, opened 2025-09-20
Implement Clear-Site-Data header on logout handlers


evilaliv3:
This ticket is to propose implementing the modern and standard ```Clear-Site-Data``` HTTP response header on SecureDrop logout endpoints. It ensures a complete cleanup of client-side data (cache, storage, cookies, service workers) upon logout, strengthening user privacy and security.
The header is fully supported by the Tor Browser, as support was integrated in Firefox back in 2018.

The functionality is currently also being within the Globaleaks project. For more details, see [Globaleaks Issue #4636](https://github.com/globaleaks/globaleaks-whistleblowing-software/issues/4636)

References:
- [W3C: Clear Site Data specification](https://www.w3.org/TR/clear-site-data/)
- [OWASP: Session Management Cheat Sheet](https://owasp.org/www-project-secure-headers/#clear-site-data)
- [Mozilla Developer Network: Clear-Site-Data documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Clear-Site-Data)
- [Can I Use – Clear-Site-Data Header Wildcard Support](https://caniuse.com/mdn-http_headers_clear-site-data_wildcard)

legoktm:
Fixed by #7660, will be included in the upcoming 2.13.0 release.

PR #7660, opened 2025-09-20
Implement Clear-Site-Data header on logout handlers

evilaliv3:
Implements #7659

## Checklist

This change accounts for:
- [X] any required additional documentation
- [X] any necessary AppArmor changes (added or removed application files)
- [X] any impact on new SecureDrop installs and upgrades
- [X] our [dependency update policy](https://developers.securedrop.org/en/latest/dependency_updates.html)


zenmonkeykstop (approved):
Confirmed header is present - this is a good automatic reinforcement to the logout page's advice to sources to use TBB's new identity button!

PR #7662, opened 2025-09-22
Actually use source flask config for Source Interface

### LINKED ISSUE #7553: [dev env] Source/Journalist Interface invalidate each other's session
## Description

I cannot use the Source and Journalist Interface alongside each other in the same browser without them invalidating each other's sessions.

## Steps to Reproduce

1. Fire up a dev env with `make dev`
2. Open `localhost:8081` and log in as `journalist`
3. Reload and confirm that you stay logged in
4. Load Source Interface in a second tab with `localhost:8080`
5. Switch back to the Journalist Interface tab

## Expected Behavior

I should be logged into the JI.

## Actual Behavior

I am logged out of the JI.

legoktm:
The main impact of this was that the SI used a "js" cookie name instead of "ss".

Notably this did not affect the secret key, which was correctly using the source config.

Fixes #7553.

## Test plan

* [x] On develop, log into the SI and notice that the cookie is "js"
* [x] Check out this change, log into the SI and notice that the cookie is correctly "ss"

## Checklist

This change accounts for:

### DIFF
--- a/securedrop/sdconfig.py
+++ b/securedrop/sdconfig.py
@@ -191,12 +191,12 @@ def _parse_config_from_file(config_module_name: str) -> SecureDropConfig:
     source_flask_config = config_from_local_file.SourceInterfaceFlaskConfig
     parsed_source_flask_config = SourceInterfaceConfig(
         SECRET_KEY=source_flask_config.SECRET_KEY,
-        SESSION_COOKIE_NAME=getattr(journ_flask_config, "SESSION_COOKIE_NAME", "ss"),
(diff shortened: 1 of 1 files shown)


legoktm:
I noticed this while reviewing #7660; presumably this also merits a regression test but not exactly sure where it belongs.

zenmonkeykstop (approved):
LGTM, config for SI still looks consistent in dev too.

zenmonkeykstop:
Merging despite CI fuzziness - affected checks are build-related and are failing due to security repo availability issues, so not related to the change.

PR #7653, opened 2025-09-05
Update rust flate2 version to 1.1.2

zenmonkeykstop:
Updates `flate2` to 1.1.2, to bump its unmaintained `adler` dependency to `adler2` instead.

## Test plan
- [ ] CI is passing
- [ ] dependency update is consistent with above.

PR #7664, opened 2025-09-22
[2.12.x] Backport "Update rust flate2 version to 1.1.2"

zenmonkeykstop:
## Status

Ready for review

## Description of Changes

Backport of #7653.

## Testing

* [ ] CI is passing
* [ ] base is `release/2.12.x`
* [ ] Only contains changes from #7653

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,10 +3,10 @@
 version = 4
 
 [[package]]
-name = "adler"
-version = "1.0.2"
+name = "adler2"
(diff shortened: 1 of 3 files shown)


zenmonkeykstop:
Note that `static-analysis-and-no-known-cves` will fail (coz another change needs backporting) but all else should be green. As will `focal` tests - those can be ignored.

vickiniu (approved):

PR #7669, opened 2025-09-25
Support importing of fixed datasets

legoktm:
loaddata.py generates a database with randomized data, which makes it difficult to write tests against. The new loadfixedata.py takes a dataset that's declared in a YAML file, invoke it via:

```
USE_FIXED_DATA=data.yaml make dev
```
The primary intent for this is to be used with the app.

Because it does override some data the script is not as efficient as it could be, for example create_source_user() generates a PGP keypair for the source, and then it's immediately overwritten by the keys supplied in the YAML.

Not every field is supported yet, the intent is that basically everything that would be exposed in APIv2 would be taken from the YAML.

There is a bit of logic added to dev-shell to ensure that even if the YAML file is in some other directory (e.g. the app), it still gets mounted into the container so the script can read from it.

Refs https://github.com/freedomofpress/securedrop-client/issues/2695.

## Test plan

* [x] review in tandem with https://github.com/freedomofpress/securedrop-client/issues/2695, notably that this PR supports everything we need for the app.
## Checklist

This change accounts for:

## AI usage

I dumped a normal database and then had Claude Code transform it to YAML, manually refined it down, and then had Claude take the YAML and create a script for it, based off of loaddata.py, which I then cleaned up.

### DIFF
--- a/securedrop/bin/dev-deps
+++ b/securedrop/bin/dev-deps
@@ -163,9 +163,13 @@ function reset_demo() {
 
     gpg2 --homedir /var/lib/securedrop/keys --no-default-keyring --keyring /var/lib/securedrop/keys/pubring.gpg --import /var/lib/securedrop/journalist.pub
 
+    # Check if we should use fixed data instead of random data
+    if [ -n "${USE_FIXED_DATA:-""}" ]; then
+        echo "Loading fixed data from ${FIXED_DATA_CONTAINER_PATH}..."
+        ./loadfixeddata.py --yaml-path "${FIXED_DATA_CONTAINER_PATH}"
     # Can't pass an array environment variable with "docker --env", so
     # break up the string we can pass.
-    if [ -n "${LOADDATA_ARGS:-""}" ]
+    elif [ -n "${LOADDATA_ARGS:-""}" ]
     then
         IFS=" " read -r -a loaddata_arg_array <<< "${LOADDATA_ARGS}"
         ./loaddata.py "${loaddata_arg_array[@]}"
--- a/securedrop/bin/dev-shell
+++ b/securedrop/bin/dev-shell
@@ -116,6 +116,22 @@ function docker_run() {
         DOCKER_RUN_ARGUMENTS="${DOCKER_RUN_ARGUMENTS} -it"
     fi
 
+    # Handle fixed data paths
+    if [ -n "${USE_FIXED_DATA:-}" ]; then
+        # Resolve absolute path for the YAML file
+        FIXED_DATA_PATH="$(realpath "${USE_FIXED_DATA}")"
+
+        # Get the directory containing the YAML file
+        FIXED_DATA_DIR="$(dirname "${FIXED_DATA_PATH}")"
+        FIXED_DATA_FILE="$(basename "${FIXED_DATA_PATH}")"
+
+        # Mount the directory containing the YAML file
+        DOCKER_RUN_ARGUMENTS="${DOCKER_RUN_ARGUMENTS} --volume ${FIXED_DATA_DIR}:/fixed-data:Z"
+
+        # Set environment variable for the script to use the mounted path
+        DOCKER_RUN_ARGUMENTS="${DOCKER_RUN_ARGUMENTS} -e FIXED_DATA_CONTAINER_PATH=/fixed-data/${FIXED_DATA_FILE}"
+    fi
+
     if [ -n "${USE_TOR:-}" ]; then
         # Mount persistent onion services
         $DOCKER_BIN volume inspect sd-onion-services -f " " || $DOCKER_BIN volume create sd-onion-services
@@ -141,6 +157,7 @@ function docker_run() {
            -e NUM_JOURNALISTS \
            -e NUM_SOURCES \
            -e LOADDATA_ARGS \
+           -e USE_FIXED_DATA \
            -e LC_ALL=C.UTF-8 \
            -e LANG=C.UTF-8 \
            -e TEST_LOCALES \
--- a/securedrop/loadfixeddata.py
+++ b/securedrop/loadfixeddata.py
@@ -1,0 +1,393 @@
+#!/opt/venvs/securedrop-app-code/bin/python
+
+"""
+Loads fixed test data from a YAML file into the SecureDrop database.
+"""
+
+import argparse
+import datetime
+import io
+import os
+import shutil
+from pathlib import Path
+from typing import Any, Dict, List, Tuple
+
+import journalist_app
+import yaml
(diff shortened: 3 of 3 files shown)


legoktm:
This should be complete now; it likely needs to be reviewed in the context of https://github.com/freedomofpress/securedrop-client/pull/2720, which contains a data.yaml file and verifies that what's in the file fully covers everything, since the v2 index it returns is identical on subsequent server runs.

cfm:
Per <https://github.com/freedomofpress/securedrop-client/pull/2720#issuecomment-3352932601>, I think what's missing here (and this is not a criticism, just musing about a possible follow-up task :-) is a non-Claude *export* script that takes the SQLite dump and serializes a *ciphertext* for each item, with or without its plaintext.  In that case, I would favor base-64-encoding both the ciphertexts and the keys directly into the YAML file so that it's a self-contained versioned snapshot of server state, at least according to the v2 Journalist API.

cfm (dismissed):
This looks good to me, and it works as documented in freedomofpress/securedrop-client#2720.  I'm approving here to unblock <https://github.com/freedomofpress/securedrop-client/pull/2720#discussion_r2388604772>, but I'll defer to you, @legoktm, on whether you want to address <https://github.com/freedomofpress/securedrop/pull/7669#issuecomment-3353349033> for <https://github.com/freedomofpress/securedrop-client/pull/2720#issuecomment-3352932601> here or as a follow-up.

legoktm:
I'll try to fix it now because the app tests are effectively flaky until we do. I was thinking of keeping the plaintext in the YAML and then a `loadfixedata.py --reencrypt-items` that generates the ciphertext (but not compressed) as separate files.

I was hoping that the YAML file format would be something that is easily human read and modified (otherwise I would've likely ended up with something close to a pretty-formatted SQLite dump). Admittedly the main reason I moved the PGP keys out of the YAML file in the first place was because they interfered with readability but primarily because they didn't fit in Claude's context window, so it couldn't read the entire file.

legoktm:
> I was thinking of keeping the plaintext in the YAML and then a `loadfixedata.py --reencrypt-items` that generates the ciphertext (but not compressed) as separate files.

I settled on a `--save-items` flag, which encrypts the content on the fly (as it currently does), but then saves the items alongside the data.yaml file so it can be used for future runs. There's a dependency loop when it comes to bootstrapping the dataset to the point where the easiest way to use it is to just hack `dev-deps` to add the `--save-items` flag once, and then you're set.

Using https://gist.github.com/legoktm/ec72eab15a1e032696e4984042be2300 (generated by Claude), ~~I ran the app server tests against a fresh server 20 times, all passed, which gives me reasonable enough confidence that it's no longer going to be flaky (previously it failed on the 10th try).~~ It just failed in CI, sigh. Digging into it...

legoktm:
I've run it in a loop 60+ times now locally without any flakiness, and tried twice in CI as well. So I'm going to say it's fine and ready for re-review @cfm; if we do see it flake in CI again we can add some debugging things to capture the failures better to identify things.

cfm (approved):
Thanks, @legoktm!  I confirm that with diff—

```diff
--- a/securedrop/bin/dev-deps
+++ b/securedrop/bin/dev-deps
@@ -166,7 +166,7 @@ function reset_demo() {
     # Check if we should use fixed data instead of random data
     if [ -n "${USE_FIXED_DATA:-""}" ]; then
         echo "Loading fixed data from ${FIXED_DATA_CONTAINER_PATH}..."
-        ./loadfixeddata.py --yaml-path "${FIXED_DATA_CONTAINER_PATH}"
+        ./loadfixeddata.py --save-items --yaml-path "${FIXED_DATA_CONTAINER_PATH}"
     # Can't pass an array environment variable with "docker --env", so
     # break up the string we can pass.
     elif [ -n "${LOADDATA_ARGS:-""}" ]

```

—you get a new round of ciphertexts:

```sh-session
user@sd:~/securedrop$ USE_FIXED_DATA=~/securedrop-client/app/server_tests/data/data.yaml make dev

user@sd:~/securedrop-client$ git diff --stat
 app/server_tests/data/items/127b98a7-2976-45a9-8dde-dd50602bffbb.gpg | Bin 704 -> 704 bytes
 app/server_tests/data/items/2c287eff-a397-46d0-905f-8842147d183d.gpg | Bin 739 -> 739 bytes
 app/server_tests/data/items/31eeac69-e02f-46f1-a5da-b8f912eb773d.gpg | Bin 720 -> 720 bytes
 app/server_tests/data/items/3f3d5d05-e758-4cbb-9ea0-876d7152fd7f.gpg | Bin 1215 -> 1215 bytes
 app/server_tests/data/items/40e13a88-5409-4201-9495-d06c335e203f.gpg | Bin 1276 -> 1276 bytes
 app/server_tests/data/items/48256b37-2761-4695-8a1a-282601dc3c87.gpg | Bin 658 -> 658 bytes
 app/server_tests/data/items/485396d7-114b-49d5-a910-43337f72b471.gpg | Bin 710 -> 710 bytes
 app/server_tests/data/items/6428c271-3108-4c73-9a44-9acfb7e2b388.gpg | Bin 1228 -> 1228 bytes
 app/server_tests/data/items/6ba8a2c8-aa3a-4b0e-95f9-34d3001a5043.gpg | Bin 695 -> 695 bytes
 app/server_tests/data/items/85f0d337-b077-41b6-a914-33fee7bfcaa1.gpg | Bin 699 -> 699 bytes
 app/server_tests/data/items/96325861-741a-41eb-91b5-bb1ff34f3e70.gpg | Bin 819 -> 819 bytes
 app/server_tests/data/items/a500332a-52dd-4296-a8e7-213cb575d9ff.gpg | Bin 1398 -> 1398 bytes
 app/server_tests/data/items/adca50a5-9f70-49db-a5ae-0c184b54c6ed.gpg | Bin 1309 -> 1309 bytes
 app/server_tests/data/items/b8260d37-87b2-49b3-a235-bac31c11021d.gpg | Bin 726 -> 726 bytes
 app/server_tests/data/items/cc1be744-8a71-4e52-92e7-51315b6cb643.gpg | Bin 1303 -> 1303 bytes
 app/server_tests/data/items/d6dc8fb4-582f-4cfc-b466-19e4b402b3ae.gpg | Bin 866 -> 866 bytes
 app/server_tests/data/items/d8782e98-0cbd-4142-850d-6bcefabb8237.gpg | Bin 783 -> 783 bytes
 app/server_tests/data/items/f53f43d9-41fa-42a6-88b0-6529aaacc599.gpg | Bin 687 -> 687 bytes
 18 files changed, 0 insertions(+), 0 deletions(-)
```

Issue #3918, opened 2018-11-02
API repsonse for a `Reply` has full system path for `filename`


heartsucker:
## Description

The filename of a reply to a source should only be the `$(basename $path)` and not just `$path`.

## Steps to Reproduce

Boot the dev env, reply via the API, fetch all replies from the API.

## Expected Behavior

The response does not contain `/var/lib/securedrop/store/$filesystem_id`.

## Actual Behavior

```json
$ http localhost:8081/api/v1/replies
HTTP/1.0 200 OK
Content-Length: 4438
Content-Type: application/json
Date: Fri, 02 Nov 2018 13:58:11 GMT
Server: Werkzeug/0.14.1 Python/2.7.6
Set-Cookie: js=eyJleHBpcmVzIjp7IiBkIjoiRnJpLCAwMiBOb3YgMjAxOCAxNTo1ODoxMSBHTVQifX0.Dr3pcw.p_VpbLQcvnnVzg_jf6MTp9gZ8VI; HttpOnly; Path=/
Vary: Cookie

{
    "replies": [
        ...
        {
            "filename": "/var/lib/securedrop/store/V7ND6HFB3ZDSWS647VA2HPTTTOKD6M433M6ZSZOKTQTQAU7RWTWS5YU62F4WIASHUYTDAYW3HZVNORQAN3QKZVBZRELYCV3NUVARL6Y=/5-fourteen_recitative-reply.gpg", 
            "is_deleted_by_source": false, 
            "journalist_username": "journalist", 
            "journalist_uuid": "3328e4e8-4037-437b-88d9-a0aa270090cb", 
            "reply_url": "/api/v1/sources/c81a50ea-4eeb-4422-ba51-9bc11e598b14/replies/8b97e43d-e84d-412b-b7e2-1fccfa0fda17", 
            "size": 906, 
            "source_url": "/api/v1/sources/c81a50ea-4eeb-4422-ba51-9bc11e598b14", 
            "uuid": "8b97e43d-e84d-412b-b7e2-1fccfa0fda17"
        }, 
        ...
    ]
}
```

heartsucker:
This is a critical error and will cause unrecoverable 500's in prod when attempting to view a source with a contaminated DB.

```
[...]

PR #7672, opened 2025-10-02
`/api/v2/data`: handle batched events from clients

### LINKED ISSUE #7665: `/api/v2/metadata`: accept batched events from clients
*From <https://docs.google.com/document/d/1wO2MCJI5Xm_BM98JeSBIxEPne4PWNjV9cQNegAPNZp0/edit?tab=t.0>:*

1. [x] Update documentation of `Index` → `MetadataRequest` in `api2/README.md` first
2. [x] "The server caches the snowflake IDs of processed events [...]"
    - [ ] "...with the same lifetime as the session that submitted them."
4. [x] Order by snowflake ID within a given request

## Current `MetadataRequest`

```json
{
	"sources": [...],
	"items": [...],
	"journalists": [...]
}
```

## Proposed `BatchRequest` replaces `MetadataRequest`

```json
{
	# Unchanged:
	"sources": [...],
	"items": [...],
	"journalists": [...],

	# New:
	"events": [  # or "changes"
		{"uuid": "<uuid>", "type": "source_deleted", "source_uuid": "<uuid>", "source_version": "<version>"},
		{"uuid": "<uuid>", "type": "source_starred", "source_uuid": "<uuid>", "source_version": "<version>"},
		{"uuid": "<uuid>", "type": "reply_sent", "source_uuid": "<uuid>", "journalist_uuid": "<uuid>", ...}
	]
}
```

cfm:
Closes #7665 by implementing an event-handling framework and handlers for two illustrative events:

https://github.com/freedomofpress/securedrop/blob/d6dfb8775cb974f86665a7fc7b594e0455e67575/securedrop/journalist_app/api2/types.py#L26-L28

Adding the [remaining events](https://docs.google.com/document/d/1wO2MCJI5Xm_BM98JeSBIxEPne4PWNjV9cQNegAPNZp0/edit?tab=t.0#heading=h.scvzqrmhv4e9) will be follow-up work once this framework has been reviewed and can be extended.

## Test plan

Each `EventType` has a test in `test_journalist_api2.py`, including for its own conflict and idempotence properties.  I've flagged some others for follow-up work.

## Checklist

This change accounts for:

### DIFF
--- a/securedrop/journalist_app/api.py
+++ b/securedrop/journalist_app/api.py
@@ -12,6 +12,7 @@ from journalist_app.api2.shared import save_reply
 from journalist_app.sessions import session
 from models import (
     InvalidUsernameException,
+    InvalidUUID,
     Journalist,
     LoginThrottledException,
     Reply,
@@ -229,7 +230,7 @@ def make_blueprint() -> Blueprint:
                     abort(409, "That UUID is already in use.")
             except NotEncrypted:
                 return jsonify({"message": "You must encrypt replies client side"}), 400
-            except ValueError:
+            except InvalidUUID:
                 abort(400, "reply does not have a valid UUID")
 
             return (
--- a/securedrop/journalist_app/api.py
+++ b/securedrop/journalist_app/api.py
@@ -1,29 +1,27 @@
 import collections.abc
 import json
 from datetime import datetime, timezone
-from os import path
 from typing import Set, Tuple, Union
-from uuid import UUID
 
 import flask
 import werkzeug
 from db import db
 from flask import Blueprint, abort, jsonify, request
 from journalist_app import utils
+from journalist_app.api2.shared import save_reply
 from journalist_app.sessions import session
 from models import (
     InvalidUsernameException,
     Journalist,
     LoginThrottledException,
     Reply,
-    SeenReply,
     Source,
     Submission,
     WrongPasswordException,
 )
 from sqlalchemy import Column
 from sqlalchemy.exc import IntegrityError
-from store import NotEncrypted, Storage
+from store import NotEncrypted
 from two_factor import OtpSecretInvalid, OtpTokenInvalid
 from werkzeug.exceptions import default_exceptions
 
@@ -224,43 +222,15 @@ def make_blueprint() -> Blueprint:
             if not data["reply"]:
                 abort(400, "reply should not be empty")
 
-            source.interaction_count += 1
             try:
-                filename = Storage.get_default().save_pre_encrypted_reply(
-                    source.filesystem_id,
-                    source.interaction_count,
-                    source.journalist_filename,
-                    data["reply"],
-                )
-            except NotEncrypted:
-                return jsonify({"message": "You must encrypt replies client side"}), 400
-
-            # issue #3918
-            filename = path.basename(filename)
-
-            reply = Reply(session.get_user(), source, filename, Storage.get_default())
-
-            reply_uuid = data.get("uuid", None)
-            if reply_uuid is not None:
-                # check that is is parseable
-                try:
-                    UUID(reply_uuid)
-                except ValueError:
-                    abort(400, "'uuid' was not a valid UUID")
-                reply.uuid = reply_uuid
-
-            try:
-                db.session.add(reply)
-                seen_reply = SeenReply(reply=reply, journalist=session.get_user())
-                db.session.add(seen_reply)
-                db.session.add(source)
-                db.session.commit()
+                reply = save_reply(source, data)
             except IntegrityError as e:
-                db.session.rollback()
                 if "UNIQUE constraint failed: replies.uuid" in str(e):
                     abort(409, "That UUID is already in use.")
-                else:
-                    raise e
+            except NotEncrypted:
+                return jsonify({"message": "You must encrypt replies client side"}), 400
+            except ValueError:
+                abort(400, "reply does not have a valid UUID")
 
             return (
                 jsonify(
--- a/securedrop/journalist_app/api2/README.md
+++ b/securedrop/journalist_app/api2/README.md
@@ -86,3 +86,15 @@ Server ->> Client: BatchResponse
 Note over Client: Global version uvwxyz
 end
 ```
+
+This diagram implies single-round-trip consistency. To make that expectation
+explicit:
+
+1. If the server $S$ currently has exactly one active client $C$; and
+
+2. $C$ submits a valid `BatchRequest` $BR$ with $n$ events $\{E_0, \dots,
+E_n\}$; and
+
+3. $S$ accepts $BR$ as valid and successfully processes all $E_i$; then
+
+4. $C$'s index SHOULD match $S$'s index without a subsequent synchronization.
--- a/securedrop/journalist_app/api2/README.md
+++ b/securedrop/journalist_app/api2/README.md
@@ -61,3 +61,28 @@ else Out of date
     Server ->> Client: MetadataResponse
 end
 ```
+
+### Batched events from client
+
+```mermaid
+sequenceDiagram
+participant Client
+participant Server
+
+Note over Client: Global version abcdef
+Note over Server: Global version abcdef
+
+Client ->> Client: reply_sent {id: X, uuid: Y, source: Z, ...}
+Client -->> Server: POST /api/v2/metadata<br>BatchRequest
+alt Already processed:
+Server ->> Server: look up status of event {id: X}
+Note over Server: Return status of event {id: X},<br>in addition to anything else requested.
+Server ->> Client: BatchResponse
+else
+Server ->> Server: process "reply_sent" event for reply {uuid: Y}
+Note over Server: Return new item {uuid: Y} and updated source {uuid: Z},<br>in addition to anything else requested.
+Note over Server: Global version uvwxyz
+Server ->> Client: BatchResponse
+Note over Client: Global version uvwxyz
+end
+```
--- a/securedrop/journalist_app/api2/__init__.py
+++ b/securedrop/journalist_app/api2/__init__.py
@@ -1,15 +1,14 @@
 import hashlib
-from dataclasses import asdict, dataclass, field
-from typing import (
-    Any,
-    Dict,
-    Mapping,
-    NewType,
-    Optional,
-    Set,
-)
+from dataclasses import asdict
+from typing import Mapping, Optional
 
 from flask import Blueprint, abort, json, jsonify, request
+from journalist_app.api2.types import (
+    Index,
+    MetadataRequest,
+    MetadataResponse,
+    Version,
+)
 from models import EagerQuery, Journalist, Reply, Source, Submission, eager_query
 from sqlalchemy.inspection import inspect
 from sqlalchemy.orm.exc import MultipleResultsFound
@@ -18,9 +17,6 @@ from werkzeug.wrappers.response import Response
 blp = Blueprint("api2", __name__, url_prefix="/api/v2")
 
 PREFIX_MAX_LEN = inspect(Source).columns["uuid"].type.length
-
-
-Version = NewType("Version", str)
 
 
 def json_version(d: Mapping) -> Version:
@@ -35,22 +31,6 @@ def json_version(d: Mapping) -> Version:
     s = json.dumps(d, separators=[",", ":"], sort_keys=True)
     b = s.encode("utf-8")
     return Version(hashlib.blake2s(b).hexdigest())
-
-
-# TODO: generic UUID[T] in Python 3.12
-SourceUUID = NewType("SourceUUID", str)
-ItemUUID = NewType("ItemUUID", str)
-JournalistUUID = NewType("JournalistUUID", str)
-
-
-@dataclass
-class Index:
-    # Source metadata, optionally filtered by `source_prefix`:
-    sources: Dict[SourceUUID, Version] = field(default_factory=dict)
-    items: Dict[ItemUUID, Version] = field(default_factory=dict)
-
-    # Non-source metadata (always returned):
-    journalists: Dict[JournalistUUID, Version] = field(default_factory=dict)
 
 
 @blp.get("/index")
@@ -97,26 +77,6 @@ def index(source_prefix: Optional[str] = None) -> Response:
     # return HTTP 304 with an empty response.
     response.set_etag(version)
     return response.make_conditional(request)
-
-
-@dataclass
-class MetadataRequest:
-    # Source metadata:
-    sources: Set[SourceUUID] = field(default_factory=set)
-    items: Set[ItemUUID] = field(default_factory=set)
-
-    # Non-source metadata:
-    journalists: Set[JournalistUUID] = field(default_factory=set)
-
-
-@dataclass
-class MetadataResponse:
-    # Source metadata:
-    sources: Dict[SourceUUID, Any] = field(default_factory=dict)
-    items: Dict[ItemUUID, Any] = field(default_factory=dict)
-
-    # Non-source metadata:
-    journalists: Dict[JournalistUUID, Any] = field(default_factory=dict)
 
 
 @blp.post("/metadata")
--- a/securedrop/journalist_app/api2/__init__.py
+++ b/securedrop/journalist_app/api2/__init__.py
@@ -7,6 +7,7 @@ from journalist_app.api2.events import EventHandler
 from journalist_app.api2.types import (
     BatchRequest,
     BatchResponse,
+    Event,
     Index,
     Version,
 )
@@ -110,6 +111,11 @@ def data() -> Response:
     response = BatchResponse()
 
     if requested.events:
+        try:
+            events = [Event(**d) for d in requested.events]
+        except TypeError:
+            abort(400, "invalid event")
+
         # Don't set up the EventHandler, connect to Redis, etc., unless we have
         # events to process.
         config = SecureDropConfig.get_current()
@@ -118,7 +124,7 @@ def data() -> Response:
         )
 
         # Process events in snowflake order.
-        for event in sorted(requested.events, key=lambda e: e.id):
+        for event in sorted(events, key=lambda e: e.id):
             result = handler.process(event)
             for uuid, source in result.sources.items():
                 response.sources[uuid] = source.to_api_v2() if source is not None else None
--- a/securedrop/journalist_app/api2/__init__.py
+++ b/securedrop/journalist_app/api2/__init__.py
@@ -3,10 +3,11 @@ from dataclasses import asdict
 from typing import Mapping, Optional
 
 from flask import Blueprint, abort, json, jsonify, request
+from journalist_app.api2.events import EventHandler
 from journalist_app.api2.types import (
+    BatchRequest,
+    BatchResponse,
     Index,
-    MetadataRequest,
-    MetadataResponse,
     Version,
 )
 from models import EagerQuery, Journalist, Reply, Source, Submission, eager_query
@@ -79,10 +80,11 @@ def index(source_prefix: Optional[str] = None) -> Response:
     return response.make_conditional(request)
 
 
-@blp.post("/metadata")
-def metadata() -> Response:
+@blp.post("/data")  # read-write BatchRequest
+@blp.post("/metadata")  # DEPRECATED: read-only MetadataRequest
+def data() -> Response:
     """
-    Return the ``MetadataResponse`` requested in the ``MetadataRequest``.  The
+    Return the ``BatchResponse`` requested in the ``BatchRequest``.  The
     client MAY choose an arbitrary list of objects with each request, e.g. from
     a shard retrieved from ``/index/<source_prefix>``.
 
@@ -91,11 +93,27 @@ def metadata() -> Response:
     the ``Reply`` tables for the set of all item UUIDs.
     """
     try:
-        requested = MetadataRequest(**request.json)  # type: ignore
+        requested = BatchRequest(**request.json)  # type: ignore
     except (TypeError, ValueError) as exc:
         abort(422, f"malformed request; {exc}")
 
-    response = MetadataResponse()
+    response = BatchResponse()
+
+    for event in requested.events:
+        """
+        # Case 1: already processed: If event.snowflake_id is already cached in
+        # Redis, we've already processed it; just return its status.
+        status = EventStatus(event.snowflake_id, 200)  # FIXME
+        response.events.append(status)
+        """
+
+        # Case 2: needs to be processed.
+        result = EventHandler.process(event)
+        for uuid, source in result.sources.items():
+            response.sources[uuid] = source.to_api_v2()
+        for uuid, item in result.items.items():
+            response.items[uuid] = item.to_api_v2()
+        response.events[result.event_id] = result.status
 
     if requested.sources:
         source_query: EagerQuery = eager_query("Source")
--- a/securedrop/journalist_app/api2/__init__.py
+++ b/securedrop/journalist_app/api2/__init__.py
@@ -117,8 +117,8 @@ def data() -> Response:
 
         try:
             events = [Event(**d) for d in requested.events]
-        except TypeError:
-            abort(400, "invalid event")
+        except (TypeError, ValueError) as e:
+            abort(400, f"invalid event: {e}")
 
         # Don't set up the EventHandler, connect to Redis, etc., unless we have
         # events to process.
--- a/securedrop/journalist_app/api2/__init__.py
+++ b/securedrop/journalist_app/api2/__init__.py
@@ -11,6 +11,8 @@ from journalist_app.api2.types import (
     Version,
 )
 from models import EagerQuery, Journalist, Reply, Source, Submission, eager_query
+from redis import Redis
+from sdconfig import SecureDropConfig
 from sqlalchemy.inspection import inspect
 from sqlalchemy.orm.exc import MultipleResultsFound
 from werkzeug.wrappers.response import Response
@@ -88,9 +90,10 @@ def data() -> Response:
     client MAY choose an arbitrary list of objects with each request, e.g. from
     a shard retrieved from ``/index/<source_prefix>``.
 
-    NB.  Returning sources is O(1) from the eagerly-loaded ``all_sources()``.
-    Returning items is O(2), since we have to search both the ``Submission`` and
-    the ``Reply`` tables for the set of all item UUIDs.
+    NB.  Reading sources (without any side effects from processing events) is
+    O(1) from the eagerly-loaded ``all_sources()``.  Reading items is O(2),
+    since we have to search both the ``Submission`` and the ``Reply`` tables for
+    the set of all item UUIDs.
     """
     try:
         requested = BatchRequest(**request.json)  # type: ignore
@@ -99,21 +102,19 @@ def data() -> Response:
 
     response = BatchResponse()
 
-    for event in requested.events:
-        """
-        # Case 1: already processed: If event.snowflake_id is already cached in
-        # Redis, we've already processed it; just return its status.
-        status = EventStatus(event.snowflake_id, 200)  # FIXME
-        response.events.append(status)
-        """
+    if requested.events:
+        # Don't set up the EventHandler, connect to Redis, etc., unless we have
+        # events to process.
+        config = SecureDropConfig.get_current()
+        handler = EventHandler(redis=Redis(decode_responses=True, **config.REDIS_KWARGS))
 
-        # Case 2: needs to be processed.
-        result = EventHandler.process(event)
-        for uuid, source in result.sources.items():
-            response.sources[uuid] = source.to_api_v2() if source is not None else None
-        for uuid, item in result.items.items():
-            response.items[uuid] = item.to_api_v2() if item is not None else None
-        response.events[result.event_id] = result.status
+        for event in requested.events:
+            result = handler.process(event)
+            for uuid, source in result.sources.items():
+                response.sources[uuid] = source.to_api_v2() if source is not None else None
+            for uuid, item in result.items.items():
+                response.items[uuid] = item.to_api_v2() if item is not None else None
+            response.events[result.event_id] = result.status
 
     if requested.sources:
         source_query: EagerQuery = eager_query("Source")
--- a/securedrop/journalist_app/api2/__init__.py
+++ b/securedrop/journalist_app/api2/__init__.py
@@ -90,6 +90,12 @@ def data() -> Response:
     client MAY choose an arbitrary list of objects with each request, e.g. from
     a shard retrieved from ``/index/<source_prefix>``.
 
+    The client MAY include a list of ``Event``s for the server to process over
+    arbitrary sources and items.  Ordering is guaranteed within a given
+    ``BatchRequest``.  Sources and items changed by one or more events will be
+    returned in their most-recent state in the ``BatchResponse`` whether or not
+    they were explicitly requested in the ``BatchRequest``.
+
     NB.  Reading sources (without any side effects from processing events) is
     O(1) from the eagerly-loaded ``all_sources()``.  Reading items is O(2),
     since we have to search both the ``Submission`` and the ``Reply`` tables for
@@ -108,7 +114,8 @@ def data() -> Response:
         config = SecureDropConfig.get_current()
         handler = EventHandler(redis=Redis(decode_responses=True, **config.REDIS_KWARGS))
 
-        for event in requested.events:
+        # Process events in snowflake order.
+        for event in sorted(requested.events, key=lambda e: e.id):
             result = handler.process(event)
             for uuid, source in result.sources.items():
                 response.sources[uuid] = source.to_api_v2() if source is not None else None
--- a/securedrop/journalist_app/api2/__init__.py
+++ b/securedrop/journalist_app/api2/__init__.py
@@ -21,6 +21,7 @@ from werkzeug.wrappers.response import Response
 
 blp = Blueprint("api2", __name__, url_prefix="/api/v2")
 
+EVENTS_MAX = 50
 PREFIX_MAX_LEN = inspect(Source).columns["uuid"].type.length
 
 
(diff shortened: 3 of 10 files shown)


securedrop/journalist_app/api2/__init__.py
```diff
@@ -131,11 +93,27 @@ def metadata() -> Response:
...
+        response.events.append(status)
+        """
+
+        # Case 2: needs to be processed.
+        result = EventHandler.process(event)
+        for uuid, source in result.sources.items():
+            response.sources[uuid] = source.to_api_v2()
+        for uuid, item in result.items.items():
+            response.items[uuid] = item.to_api_v2()
+        response.events[result.event_id] = result.status
```
cfm: This is the entry-point for handling events.  The `events` dictionary is just status information.  Changes are returned in the `sources` and `items` dictionaries just like reads.

securedrop/journalist_app/api2/__init__.py
```diff
@@ -131,11 +93,27 @@ def metadata() -> Response:
...
-    response = MetadataResponse()
+    response = BatchResponse()
+
+    for event in requested.events:
+        """
+        # Case 1: already processed: If event.snowflake_id is already cached in
+        # Redis, we've already processed it; just return its status.
+        status = EventStatus(event.snowflake_id, 200)  # FIXME
+        response.events.append(status)
+        """
```
cfm: Assumption: `event.id` snowflakes come from the client already parameterized by (hash of) the session token, so a Redis-backed TTL-expired cache of IDs is all we need to enforce idempotence.  Any concerns about this approach?

securedrop/journalist_app/api2/types.py
cfm, line 1: @legoktm, I'd be especially interested in your review of this type/dataclass hierarchy.  As with the original #7579 → #7604, I'm going for precision first, and we can streamline as needed.

securedrop/journalist_app/api.py
```diff
@@ -224,43 +221,15 @@ def all_source_replies(source_uuid: str) -> Tuple[flask.Response, int]:
...
-                    abort(400, "'uuid' was not a valid UUID")
-                reply.uuid = reply_uuid
-
-            try:
-                db.session.add(reply)
-                seen_reply = SeenReply(reply=reply, journalist=session.get_user())
-                db.session.add(seen_reply)
-                db.session.add(source)
-                db.session.commit()
+                reply = utils.save_reply(source, data)
```
cfm: I'd like for the v2 API event-handlers to follow this pattern, as much as possible reusing "write to database" functions factored out from the v1 API endpoint.  Any concerns about this approach?

cfm (commented):
I've flagged a few questions for feedback inline before I flesh out the rest of the event-handling machinery and the `reply_sent` handler.  (Reviewing commit by commit may give a better overview than the overall diff.)

securedrop/journalist_app/api2/__init__.py
legoktm: That seems right to me, notably our redis eviction policy is the default "noeviction", so keys should never be removed except for when the TTL expires (if redis fills up, then it'll reject writes entirely, but I don't think we need to worry about that scenario too much, it would be a pretty extreme self-DoS)

securedrop/journalist_app/api.py
legoktm: Agreed on the approach, but let's create a new api2.shared or api2.backend module for the logic instead of sticking in utils.py which is already a mess.

securedrop/journalist_app/api2/events.py
```diff
@@ -0,0 +1,55 @@
...
+            return EventResult(
+                event_id=event.id,
+                status=EventStatusCode.NotImplemented,
+            )
+
+        return handler(event)
+
+    @staticmethod
+    def handle_reply_sent(event: Event) -> EventResult:
+        source = get_or_404(Source, event.target.source_uuid, column=Source.uuid)
```
legoktm: I don't think we want to use this here since the or_404 part will immediately abort(), which we don't want, since we'd want to process other batched requests?

securedrop/journalist_app/api2/types.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+EventID = NewType("EventID", str)  # int, but opaque on the wire
+
+
+class EventType(StrEnum):
+    REPLY_SENT = auto()
+
+
+class EventStatusCode(IntEnum):
```
legoktm, line 31: Not 100% sold on using HTTP status codes...or at least would be nice if we could also send some sort of more specific error message if we have one (e.g. in EventHandler.process() we're losing the underlying exception reason)

securedrop/journalist_app/api2/events.py
```diff
@@ -0,0 +1,55 @@
...
+                raise TypeError("invalid event target")
+
+        except (TypeError, ValueError):
+            return EventResult(
+                event_id=event.id,
+                status=EventStatusCode.BadRequest,
+            )
+
+        try:
+            handler = getattr(cls, f"handle_{event.type}")
```
legoktm: I don't think this can result in an RCE as it currently stands but I'd feel more comfortable with just an explicit mapping, e.g. `handler = {"reply_sent": cls.handle_reply_sent}[event.type]`.

securedrop/journalist_app/api2/types.py
```diff
@@ -0,0 +1,97 @@
...
+
+    # Non-source metadata:
+    journalists: Set[JournalistUUID] = field(default_factory=set)
+
+    # Events submitted by the client:
+    events: List[Event] = field(default_factory=list)
+
+
+@dataclass
+class BatchResponse:
```
legoktm, line 109: How will we represent the absence of data, i.e. a source being deleted?

legoktm (commented):
I like the direction so far!

securedrop/journalist_app/api2/events.py
cfm: Yes, this counts on `EventType("invalid")` raising `ValueError: "invalid" is not a valid EventType` as an equivalent implicit guard.  Happy to do it explicitly too.

securedrop/journalist_app/api2/types.py
cfm: Do we need to?  If the client has submitted an event like `source_deleted(event_id=123456, source_uuid="abcdef")`, then it's already carried it out the deletion client-side; it just needs to know the server accepted it:

1. ```{"events": {"id": 123456, "status": 200}, ...}``` → event `123456` accepted

If the server has deleted source `abcdef`, then in the next sync the client will receive an index missing source `abcdef` and know to delete it, per freedomofpress/securedrop-client#2716.

Now, we could also return here:

2. ```{..., "sources": {"abcdef": null}}``` → delete source `abcdef`

(1) should be sufficient, but we can do (2) as well if we want to be equally explicit about the effects of all events.

legoktm: IMO the value of returning new data here is so that the state is consistent without needing to wait for the next sync to pick up the deletion (which was partly one of the issues in the old client) - so I like your suggestion of having it be `null` to indicate deletion.

securedrop/journalist_app/api2/events.py
legoktm: Right, I would like it to be explicit just to make it easier to reason about when reading :)

securedrop/journalist_app/api2/types.py
```diff
@@ -0,0 +1,97 @@
...
+    items: Dict[ItemUUID, Version] = field(default_factory=dict)
+
+    # Non-source metadata (always returned):
+    journalists: Dict[JournalistUUID, Version] = field(default_factory=dict)
+
+
+@dataclass
+class SourceTarget:
+    source_uuid: SourceUUID
+    version: Version
```
vickiniu, line 56: Is there a reason we need to send the version with the pending event request?

cfm: I'm realizing that [this was only implicit in the proposal][proposal], but the idea is that the server can choose to reject events from out-of-date clients.  The easy edge case goes like this:

1. Server is at version $v$, where source $S$ has items $`\{x, y, z\}`$
2. Client is at version $v-1$, where source $S$ has items $`\{x, y\}`$
3. Client submits a `source_conversation_deleted` event for $S$, expecting that only items $`\{x, y\}`$ will be deleted
4. Server accepts the event and deletes items $`\{x, y, z\}`$

We can choose per event/handler whether to enforce this check, but I'd like to have the capability.


[proposal]: https://docs.google.com/document/d/1wO2MCJI5Xm_BM98JeSBIxEPne4PWNjV9cQNegAPNZp0/edit?tab=t.0#heading=h.vpcprshnlfc9

securedrop/journalist_app/api.py
cfm: Good idea.  Done in ce7e5efb365deb2b119a15e5bf186cc09e2a092c.

securedrop/journalist_app/api2/events.py
cfm: Fixed in 246256c5f191736c5a39aa5b4372e4526c009f0b.

securedrop/journalist_app/api2/types.py
cfm: Added a minimal structure for this in 64ad8ca967d88e731f21731ed9966c4745c639b0.  Let me know what you think.

legoktm: looks good :)

securedrop/journalist_app/api2/types.py
cfm: Scaffolded in 5bc727cd2070498f0d4ee3f616d6104c11e0b2f5 and demonstrated for the `item_deleted` event in 4857e786f930069869036566073c037eac3e1b5c.

securedrop/journalist_app/api2/events.py
cfm: Done in 76a925c1ff89fcd58e8e2617b819ec94d75964db.

securedrop/journalist_app/api2/events.py
```diff
@@ -48,18 +55,35 @@ def process(cls, event_dict: dict) -> EventResult:
...
+        self.mark_as_processed(event, result)
+        return result
+
+    def has_processed(self, event: Event) -> bool:
+        return self._redis.get(f"{self.REDIS_EVENT_PREFIX}{event.id}") is not None
+
+    def mark_as_processed(self, event: Event, result: EventResult) -> None:
+        # FIXME: get expiration from session
+        self._redis.set(
+            f"{self.REDIS_EVENT_PREFIX}{event.id}", result.status[0], ex=3600
```
cfm: This will require adding a method to `journalist_app.sessions.SessionInterface` to return the session's _current_ TTL from Redis.

cfm (commented):
Feedback so far addressed!  582bae64cbc39b4ca1dc37d63d0647c40d97464d adds naïve idempotence.

securedrop/journalist_app/api2/__init__.py
```diff
@@ -99,43 +82,46 @@ def index(source_prefix: Optional[str] = None) -> Response:
...
+    response = BatchResponse()
+
+    if requested.events:
+        # Don't set up the EventHandler, connect to Redis, etc., unless we have
+        # events to process.
+        config = SecureDropConfig.get_current()
+        handler = EventHandler(redis=Redis(decode_responses=True, **config.REDIS_KWARGS))
+
+        # Process events in snowflake order.
+        for event in requested.events.sort(key=lambda e: e.id):
```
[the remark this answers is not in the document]
cfm: Fixed in 0c3d8ef4ffc73ead76fa568a3b02e0b191a66f3a.

securedrop/journalist_app/api2/events.py
```diff
@@ -0,0 +1,147 @@
...
+            if "source_uuid" in event.target:
+                event.target = SourceTarget(**event.target)
+            elif "item_uuid" in event.target:
+                event.target = ItemTarget(**event.target)
+            else:
+                raise TypeError("invalid event target")
+
+        except (TypeError, ValueError) as e:
+            return EventResult(
+                event_id=event.id,
```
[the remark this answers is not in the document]
cfm: Fixed in 3024d09a2d7cd6f359aadec933a6ae290c8af918.

securedrop/journalist_app/api2/__init__.py
```diff
@@ -99,43 +82,46 @@ def index(source_prefix: Optional[str] = None) -> Response:
...
+    response = BatchResponse()
+
+    if requested.events:
+        # Don't set up the EventHandler, connect to Redis, etc., unless we have
+        # events to process.
+        config = SecureDropConfig.get_current()
+        handler = EventHandler(redis=Redis(decode_responses=True, **config.REDIS_KWARGS))
+
+        # Process events in snowflake order.
+        for event in sorted(requested.events, key=lambda e: e.id):
```
cfm: This needs a test in `test_journalist_api2.py`.

securedrop/journalist_app/api2/types.py
cfm, line 1: I think our `make typelint` is [failing](https://github.com/freedomofpress/securedrop/actions/runs/18487751338/job/52674368134#step:6:104) on an old mypy version that's tripping on Python 3.11 types for `enum.StrEnum` and the native `dict`.

securedrop/journalist_app/api2/events.py
cfm: For now I've set this to one day:

https://github.com/freedomofpress/securedrop/blob/0c3d8ef4ffc73ead76fa568a3b02e0b191a66f3a/securedrop/journalist_app/api2/events.py#L34

I now question my initial assumption (in freedomofpress/securedrop-client#2595) that the idempotence period needs to be aligned to the session lifetime.  Rather, I think we want to balance:

1. caching event IDs long enough to enforce idempotence; but
2. expiring event IDs soon enough to avoid clogging up the works.  (I put this so hand-wavily because this is an authenticated API endpoint on an authenticated onion service: we're not concerned about this as a DoS vector.)

So I'm inclined to relax to the assumption that the idempotence period needs to be *greater* than the remaining session lifetime, or always longer than [8 hours](https://github.com/freedomofpress/securedrop/blob/f8aecebfc0836fe1bb4fa633ff83e539cf83a686/securedrop/journalist_app/api.py#L27).  An idempotence period of one day satisfies that requirement and is easy both to communicate and to reason about.

legoktm:
Let me first rebase this to get rid of the focal CI and then I'll start the review + look at mypy.

securedrop/journalist_app/api2/types.py
legoktm: As part of de-focalification, we left `python_version = "3.8"` in the mypy config, switching it to 3.12 makes mypy pass. I'll push a commit for it.

securedrop/journalist_app/api2/README.md
```diff
@@ -61,3 +61,29 @@ else Out of date
...
+alt Already processed:
+Server ->> Server: look up status of event {snowflake_id: X}
+Note over Server: Return status,<br>in addition to anything else requested.
+Server ->> Client: BatchResponse
+else
+Server ->> Server: process "reply_sent" event
+Note over Server: Return new item and updated source,<br>in addition to anything else requested.
+Note over Server: Global version uvwxyz
+Server ->> Client: BatchResponse
+Note over Client: Global version uvwxyz
```
legoktm, line 86: GitHub doesn't seem happy with this, there's a render error at https://github.com/freedomofpress/securedrop/blob/166e9947450c0576542cee375384ea34d2fd7ec7/securedrop/journalist_app/api2/README.md#batched-events-from-client

securedrop/journalist_app/api2/__init__.py
cfm: [Actually](https://github.com/freedomofpress/securedrop/actions/runs/18503891526/job/52727734703?pr=7672#step:4:2037), what this needs is to order before attempting deserialization into `Event`, which will obviate the path in `EventHandler.process()` to skip events so invalid as to be missing an `id` attribute:


```suggestion
        # Process events in snowflake order.
        for event in sorted(requested.events, key=lambda e: e["id"]):  # TODO: catch KeyError
```

legoktm: maybe `e.get("id")` to address the KeyError concern?

though any reason we wouldn't deserialize first, and then process/order them? I would be okay with deserialization errors short-circuiting all execution.

securedrop/journalist_app/api2/events.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+    This class is the per-request context for handling events.  To add a handler
+    for a new event `thing_done`, you must:
+
+    1. define the enum value `EventType.THING_DONE` in journalist_api2.types;
+
+    2. define the handler as a static method `handle_thing_done(event: Event)`
+       in this class
+
+    3. explicitly register `{"thing_done": cls.handle_thing_done}` inside
```
legoktm:
```suggestion
    3. explicitly register `{"thing_done": self.handle_thing_done}` inside
```

securedrop/journalist_app/api2/events.py
```diff
@@ -0,0 +1,147 @@
...
+                event_id=event.id,
+                status=(EventStatusCode.NotImplemented, f"no handler for event type: {event.type}"),
+            )
+
+        result = handler(event)
+        self.mark_as_processed(event, result)
+        return result
+
+    def has_processed(self, event: Event) -> bool:
+        return self._redis.get(f"{self.REDIS_EVENT_PREFIX}{event.id}") is not None
```
legoktm: I just want to note that we are directly putting non-validated "user" (scare quotes because they're a trusted user) input into the redis key; based on https://stackoverflow.com/a/30271837 I don't believe this is an issue (e.g. in memcached the allowed key set is far more restrictive, so in MediaWiki we'd sha1() any raw user input to make it safe)

securedrop/journalist_app/api2/events.py
```diff
@@ -0,0 +1,147 @@
...
+
+        elif submission:
+            utils.delete_file_object(submission)
+            return EventResult(
+                event_id=event.id,
+                status=(EventStatusCode.OK, None),
+                items={event.target.item_uuid: None},
+            )
+
+        elif reply:
```
legoktm: This code block is identical to the one above it with the exception of submission vs reply being deleted, right? Can we unify it? e.g. `item = submission or reply; delete_file_object(item)`?

securedrop/journalist_app/api2/shared.py
```diff
@@ -0,0 +1,50 @@
...
+def save_reply(source: Source, data: dict) -> Reply:
+    source.interaction_count += 1
+    filename = Storage.get_default().save_pre_encrypted_reply(
+        source.filesystem_id,
+        source.interaction_count,
+        source.journalist_filename,
+        data["reply"],
+    )
+
+    # issue #3918
```
legoktm: nit: Let's make this more useful instead of just pointing to https://github.com/freedomofpress/securedrop/issues/3918 since we're modifying it...
```suggestion
    # We only save the stored reply's basename and not the whole storage path
```

securedrop/journalist_app/api2/shared.py
```diff
@@ -0,0 +1,50 @@
+"""
+This module contains helper functions factored out of the v1 Journalist API
+(journalist_app.api) and shared with the v2 Journalist API (journalist_app.api2)
+"""
+
+from os import path
```
legoktm: nit: this feels a bit unusual, IMO `import os.path` is better so you can use `path` as a normal variable without overloading. (Or switch to pathlib but that's asking more of you to refactor a function you're just moving to a different file so I'm not asking for that)

Edit: this is pre-existing, even more of a nit then.

securedrop/journalist_app/api2/shared.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # issue #3918
+    filename = path.basename(filename)
+
+    reply = Reply(session.get_user(), source, filename, Storage.get_default())
+
+    reply_uuid = data.get("uuid")
+    if reply_uuid is not None:
+        # check that is is parseable
+        UUID(reply_uuid)
```
legoktm: The try/except no longer directly wraps this, I kind of think we should have an explicit InvalidUUID exception vs treating all ValueErrors as equivalent to invalid UUID.

securedrop/journalist_app/api2/types.py
```diff
@@ -0,0 +1,109 @@
...
+    Set,
+    Tuple,
+    Union,
+)
+
+Record = NewType("Record", dict[str, Any])
+Version = NewType("Version", str)
+
+
+# TODO: generic UUID[T] in Python 3.12
```
legoktm, line 17: nit: We now live in a 3.12 world!

securedrop/journalist_app/api2/events.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def mark_as_processed(self, event: Event, result: EventResult) -> None:
+        self._redis.set(
+            f"{self.REDIS_EVENT_PREFIX}{event.id}",
+            result.status[0],
+            ex=self.IDEMPOTENCE_PERIOD,
+        )
+
+    @staticmethod
+    def handle_item_deleted(event: Event) -> EventResult:
```
legoktm, line 88: Q since you said it in a different comment, will we have a `source_conversation_deleted` that is effectively bulk `item_deleted`? Or would the app decompose that action into multiple `item_deleted`? And if we have the `source_conversation_deleted` is there another use case for deleting individual items? (I can't think of one based on the current client...)

securedrop/journalist_app/api2/types.py
```diff
@@ -0,0 +1,109 @@
...
+    type: EventType
+    data: dict[str, Any] = field(default_factory=dict)
+
+
+@dataclass
+class EventResult:
+    event_id: EventID
+    status: EventStatus
+
+    # Changed (return {<uuid>: None} to indicate deletion):
```
legoktm: nit: the content in the parenthetical is important and not an aside :)

```suggestion
    # Changed sources/items, return {<uuid>: None} to indicate deletion:
```

securedrop/journalist_app/api2/events.py
legoktm: Sounds good to me.

securedrop/tests/test_journalist_api2.py
```diff
@@ -304,8 +307,186 @@ def test_api2_metadata_validation_valid_requests(
...
+            target=ItemTarget(item_uuid=submission_uuid, version=submission_version),
+            type=EventType.ITEM_DELETED,
+        )
+        response = app.post(
+            url_for("api2.data"),
+            json={"events": [asdict(event)]},
+            headers=get_api_headers(journalist_api_token),
+        )
+        assert response.json["events"][event.id] == [200, None]
+        assert response.json["items"][event.target.item_uuid] is None
```
legoktm, line 473: Let's make this a more complete integration test and verify the submission (and later reply) in question was deleted by querying the model directly?

securedrop/tests/test_journalist_api2.py
```diff
@@ -304,8 +307,186 @@ def test_api2_metadata_validation_valid_requests(
...
+            type=EventType.REPLY_SENT,
+            data=reply,
+        )
+        response = app.post(
+            url_for("api2.data"),
+            json={"events": [asdict(event)]},
+            headers=get_api_headers(journalist_api_token),
+        )
+        assert response.json["events"][event.id] == [200, None]
+        assert reply["uuid"] in response.json["items"]
```
legoktm: Can we verify the reply outside of the API code? Maybe fetch the saved ciphertext, run it through decrypt_as_journalist and verify it matches what we expect?

securedrop/journalist_app/api2/__init__.py
```diff
@@ -99,43 +82,46 @@ def index(source_prefix: Optional[str] = None) -> Response:
...
-    response = MetadataResponse()
+    response = BatchResponse()
+
+    if requested.events:
+        # Don't set up the EventHandler, connect to Redis, etc., unless we have
+        # events to process.
+        config = SecureDropConfig.get_current()
+        handler = EventHandler(redis=Redis(decode_responses=True, **config.REDIS_KWARGS))
+
+        # Process events in snowflake order.
```
legoktm, line 130: One thing we had in the previous bulk delete API PR was a fixed limit on how many events the server would process in one request. I think that would still be useful just to ensure we don't accidentally self-DoS ourselves, or send too many events at once that the server times out and can't respond. Maybe we can start at 50 events? Assume worst case 1s per event giving us 50s, which keeps us under 60s with a bit of buffer room.

Would I be correct in assuming that won't make the app implementation any more complex than just needing to grab the first 50 pending events when kicking off a write request?

legoktm (changes requested):
Overall looks good, see inline comments :)

securedrop/journalist_app/api2/types.py
vickiniu: @cfm right, but I had thought we were using the snowflake/event ID to decide which events are stale? looking at the code, I don't see the `SourceTarget.version` being read or used anywhere, so I'm wondering if we need to include, but I might be missing that logic (or it's yet to come)

securedrop/tests/test_journalist_api2.py
```diff
@@ -304,8 +307,186 @@ def test_api2_metadata_validation_valid_requests(
...
+        index = app.get(
+            url_for("api2.index"),
+            headers=get_api_headers(journalist_api_token),
+        )
+
+        assert index.status_code == 200
+        source_version = index.json["sources"][source_uuid]
+
+        reply = {
+            "journalist_uuid": test_journo["uuid"],
```
vickiniu: talking to @micahflee, it might make sense instead of sending the `journalist_uuid` with the pending event for the server to populate the `journalist_uuid` for an event based on the auth token of the request? this would make it easier for us to create events in offline mode (when we don't have the journalist UUID in the client) 

separately, I think (afaict) the event data right now is just typed as `Any`. it might be helpful to have the types for event data specified in the Python code? or in some batch API documentation, just to have more certainty that client + server match

securedrop/journalist_app/api2/types.py
cfm: Thanks!

securedrop/journalist_app/api2/events.py
cfm: This comes from <https://github.com/freedomofpress/securedrop-client/pull/2728#discussion_r2417637104>: provided by the v1 Journalist API, not consumed by the legacy client, but intended for feature parity—and here a good base case for demonstrating that `{"<item_uuid>": null}` means deletion.  :-)

securedrop/journalist_app/api2/types.py
cfm: This is also explicit here:

https://github.com/freedomofpress/securedrop/blob/6bebcfb4f2f0014d5ec836492d61081f90add4c8/securedrop/journalist_app/api2/types.py#L96-L98

securedrop/journalist_app/api2/events.py
cfm: Formalized (documented and tested) in afcfcdd618c217c6f7e951cc47a6276aceac2efe.

securedrop/journalist_app/api2/__init__.py
cfm: Done in 46dc2d4db7fcbd91b048e44d7e21c0f50fa41bc5.

securedrop/journalist_app/api2/README.md
cfm: Fixed in a0e903b8756fdf982777de7f8edd812727eb8e7b.

securedrop/journalist_app/api2/events.py
cfm: Fixed in afcfcdd618c217c6f7e951cc47a6276aceac2efe.

securedrop/journalist_app/api2/events.py
cfm: Good point.  d4bc7cf25933f067893553041d4bc9537d712576 adds some isolation using session values controlled by the server, following this pattern:

https://github.com/freedomofpress/securedrop/blob/52b2aed3866af8a2639cbb0161a0aea7fed25a54/securedrop/journalist_app/sessions.py#L156

securedrop/journalist_app/api2/events.py
cfm: Done in f6da1486341a979d7737f6caedead7bf2589bebb.

securedrop/journalist_app/api2/shared.py
cfm: Fixed in 05c240a42c959a9da7146f16e4662f274578227d.

securedrop/journalist_app/api2/shared.py
cfm: Done in 3e8de143c599307db263c64fa833a362763fffac.

securedrop/journalist_app/api2/__init__.py
cfm: Done in 444bf7a1aec247a7891262f7fcde467f6a838586.

securedrop/tests/test_journalist_api2.py
cfm: Re: `journalist_uuid`:  Turns out the `save_reply()` helper from the v1 API was doing exactly this!  I've removed the `journalist_uuid` from the event in 3014e0c7dee8129e398c73c157d84796737f1a10.

cfm: Re: typing:  I hear you!  I'll typedef `Event.data` for `reply_sent` and `item_deleted` tomorrow, either here or in a follow-up pull request depending on @legoktm's review.

securedrop/journalist_app/api2/types.py
cfm: In practice, this implementation so far uses the `Event.id` snowflake just for ordering.  `source_conversation_deleted` (to be implemented in a follow-up pull request next sprint, at this point) is the main case I have in mind for enforcing `SourceTarget.version`, because I envision that the client "MUST" be current with the server for that source, not just "SHOULD".

Otherwise, wouldn't we need to add (and update) snowflakes like `{Source,Submission,Reply}.mtime` to be able to compare them against incoming snowflake `Event.id`s?  Let's chat if I'm missing something....

cfm:
Thanks for the detailed review, @legoktm!  I've addressed all the implementation-level feedback.  Outstanding comments (from you re: tests, from @vickiniu re: typing) I'll address tomorrow, either here or in follow-up pull requests depending on your re-review.

securedrop/journalist_app/api2/events.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def __init__(self, session: Session, redis: Redis) -> None:
+        self._session = session
+        self._redis = redis
+
+    def process(self, event: Event) -> EventResult:
+        """The per-event entry-point for handling a single event."""
+
+        try:
+            event.type = EventType(event.type)  # strict enum
```
legoktm: I think we should move this validation logic to also be pre-processing, that way regardless of where we are, once we start dealing with `Event` objects we know they're already validated and using the correct internal types. Or to put it another way, we should only have one place that can raise TypeErrors.

So maybe we have a `deserialize_events_from_user_input` function that takes care of this, or integrate this into Event's `__post_init__`?

legoktm (changes requested):
One more thing regarding validation/deserialization.

securedrop/journalist_app/api2/events.py
cfm: "Only one place that can raise `TypeError`" is a great way to put it.  That's done (with notes) in c8ee0f47ad59893b92bce3cae44a44d310e11be9, and `EventHandler.process()` is much more straightforward as a result.

securedrop/tests/test_journalist_api2.py
cfm: Done in 0a35036c9a5f5522ccc3a99fceb5a285a6214a41.

securedrop/tests/test_journalist_api2.py
cfm: Done in 9e1bf857135cc0dde7be3b4c2225c10c19e65795.

legoktm (approved):
Looks great, thanks :)

legoktm:
(Will wait for CI to finish before adding to merge queue)

securedrop/journalist_app/api2/types.py
cfm: In theory this now lets us do (e.g.):

```diff
diff --git a/securedrop/journalist_app/api2/types.py b/securedrop/journalist_app/api2/types.py
index b39c1c81b..46b8e701b 100644
--- a/securedrop/journalist_app/api2/types.py
+++ b/securedrop/journalist_app/api2/types.py
@@ -1,23 +1,32 @@
+import uuid
 from dataclasses import dataclass, field
 from enum import IntEnum, StrEnum, auto
 from typing import (
     Any,
+    Generic,
     List,
+    Literal,
     NewType,
     Optional,
     Set,
     Tuple,
+    TypeVar,
     Union,
 )
 
 Record = NewType("Record", dict[str, Any])
 Version = NewType("Version", str)
 
+K = TypeVar("K")
 
-# TODO: generic UUID[T] in Python 3.12
-SourceUUID = NewType("SourceUUID", str)
-ItemUUID = NewType("ItemUUID", str)
-JournalistUUID = NewType("JournalistUUID", str)
+
+class BrandedUUID(uuid.UUID, Generic[K]):
+    __slots__ = ()
+
+
+type SourceUUID = BrandedUUID[Literal["Source"]]
+type ItemUUID = BrandedUUID[Literal["Item"]]
+type JournalistUUID = BrandedUUID[Literal["Journalist"]]
 
 
 EventID = NewType("EventID", str)  # int, but opaque on the wire
@@ -56,12 +65,18 @@ class SourceTarget:
     version: Version
 
 
+s = SourceTarget(source_uuid=uuid.uuid4(), version=Version("foo"))
+
+
 @dataclass
 class ItemTarget:
     item_uuid: ItemUUID
     version: Version
 
 
+i = ItemTarget(item_uuid=s.source_uuid, version=Version("bar"))  # should fail typelint
+
+
 @dataclass
 class Event:
     id: EventID

```

This doesn't actually get us any more practical domain separation.  But it *does* require us to update to [mypy 1.12](https://mypy.readthedocs.io/en/stable/changelog.html#mypy-1-12), which breaks `sqlmypy`.  I'll propose some more typing we can add for robustness, but I don't think this is worth doing after all, at least not currently.

securedrop/tests/test_journalist_api2.py
cfm: Proposed in #7681.

PR #7674, opened 2025-10-09
Set noble end-of-life date

legoktm:
Historically we only did this closer to the end of life, but that left very old SecureDrops that were on an outdated version still up because they never picked up the code update that shipped the EOL check.

So let's enable it now, so that regardless of whether they continue to receive further updates, they will go down in April 2029.

Note that the https://securedrop.org/noble-eol URL doesn't exist yet, presumably we'll create it as an alias later.

## Test plan

* [ ] I think visual review is fine, unless tests would be beneficial?
## Checklist

This change accounts for:

### DIFF
--- a/securedrop/journalist_templates/base.html
+++ b/securedrop/journalist_templates/base.html
@@ -20,7 +20,7 @@
   {% if session.logged_in() %}
   {% if g.show_os_past_eol_warning %}
   <div id="os-past-eol" class="alert-banner">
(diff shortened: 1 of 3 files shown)


zenmonkeykstop:
LGTM based on visual review - staging is failing on something that looks unrelated, giving it another run in case it's a flake.

legoktm:
Probably is #7675 ?

zenmonkeykstop:
The failing integration test can't be reproduced locally and isn't affected by the actual changes in the PR AFAICT. Merging regardless, but #7675 will need to be resolved or xfailed before release time.

zenmonkeykstop (approved):

PR #7681, opened 2025-10-17
feat(`Event`): typedef `Event.data` per `Event.type`

cfm:
Addresses @vickiniu's suggestion in <https://github.com/freedomofpress/securedrop/pull/7672#discussion_r2434068205>.

### DIFF
--- a/securedrop/journalist_app/api2/types.py
+++ b/securedrop/journalist_app/api2/types.py
@@ -14,7 +14,8 @@ Record = NewType("Record", dict[str, Any])
 Version = NewType("Version", str)
 
 
-# TODO: generic UUID[T] in Python 3.12
+# NB.  Ideally we'd have a generic UUID[T], but the semantics don't change
+# before mypy 1.12, which is incompatible with our use elsewhere of sqlmypy.
 ReplyUUID = NewType("ReplyUUID", str)
 SourceUUID = NewType("SourceUUID", str)
 ItemUUID = NewType("ItemUUID", str)
--- a/securedrop/journalist_app/api2/types.py
+++ b/securedrop/journalist_app/api2/types.py
@@ -3,11 +3,11 @@ from enum import IntEnum, StrEnum, auto
 from typing import (
     Any,
     List,
+    Mapping,
     NewType,
     Optional,
     Set,
     Tuple,
-    Union,
 )
 
 Record = NewType("Record", dict[str, Any])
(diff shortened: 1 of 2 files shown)


securedrop/journalist_app/api2/types.py
```diff
@@ -3,18 +3,19 @@
...
     Set,
     Tuple,
-    Union,
 )
 
 Record = NewType("Record", dict[str, Any])
 Version = NewType("Version", str)
 
 
 # TODO: generic UUID[T] in Python 3.12
```
legoktm: Can we update this per https://github.com/freedomofpress/securedrop/pull/7672#discussion_r2441422075 since 3.12 is no longer the blocker?

(Haven't reviewed the rest yet)

securedrop/journalist_app/api2/types.py
```diff
@@ -51,36 +52,71 @@ class Index:
...
             if "source_uuid" in self.target:
                 self.target = SourceTarget(**self.target)
             elif "item_uuid" in self.target:
                 self.target = ItemTarget(**self.target)
             else:
                 raise TypeError(f"invalid event target: {self.target}")
 
+        if not isinstance(self.data, EventData):
+            try:
+                self.data = EVENT_DATA_TYPES[self.type](**(self.data or {}))
```
legoktm: Can we split this into multiple lines? I think it's just trying to do too much in one line, which is:
1) looking up if there is an `EVENT_DATA_TYPES` handler
2) if self.data exists
3) creating the dataclass

maybe:
```patch
diff --git a/securedrop/journalist_app/api2/types.py b/securedrop/journalist_app/api2/types.py
index 6635d79d1..7207c2c7c 100644
--- a/securedrop/journalist_app/api2/types.py
+++ b/securedrop/journalist_app/api2/types.py
@@ -109,13 +109,13 @@ class Event:
             else:
                 raise TypeError(f"invalid event target: {self.target}")
 
-        if not isinstance(self.data, EventData):
+        if not isinstance(self.data, EventData) and self.data and self.type in EVENT_DATA_TYPES:
             try:
-                self.data = EVENT_DATA_TYPES[self.type](**(self.data or {}))
-            except KeyError:
-                self.data = None
+                self.data = EVENT_DATA_TYPES[self.type](**self.data)
             except TypeError:
                 raise TypeError(f"invalid event data for type {self.type}")
+        else:
+            self.data = None
```

legoktm (changes requested):

securedrop/journalist_app/api2/types.py
cfm: I like that.  Done in a12dbc847e611cf26eb11902a8167dde4670fc4d.

securedrop/journalist_app/api2/types.py
cfm: Revised in c3f99e494e83f6145a06105cf1f45b2803602560.

cfm:
Rebased from `develop` at 28423c26a1182c3daafb66a72cfd64bb87faecbb.

legoktm (approved):
Thanks!

legoktm:
(waiting on CI before adding to the merge queue)

PR #7683, opened 2025-10-21
Add remaining write methods to APIv2

### LINKED ISSUE #7679: API v2: implement remaining event handlers
#7672 provides handlers for only these events:
- `item_deleted`
- `reply_sent`

Once it's merged, the remaining events [expected](https://github.com/freedomofpress/securedrop-client/blob/40ad8c7c30f0e83730b9eb49b409e9fe0e911b90/app/src/types.ts#L213-L222) by freedomofpress/securedrop-client#2728 should be quick incremental work:

1. [ ] `source_deleted`
2. [ ] `source_conversation_deleted` (see <https://github.com/freedomofpress/securedrop/pull/7672#discussion_r2438045024>)
3. [ ] `source_starred`
4. [ ] `source_unstarred`
5. [ ] `source_seen`

Per <https://github.com/freedomofpress/securedrop/pull/7672#discussion_r2434068205>, implementing these handlers should include defining and validating their expected `Event.data` dataclasses.

legoktm:
Fixes #7679.

## Test plan

* [ ] Test along with associated app PR that it works as expected

## Checklist

This change accounts for:

### DIFF
--- a/securedrop/journalist_app/api2/events.py
+++ b/securedrop/journalist_app/api2/events.py
@@ -6,8 +6,9 @@ from journalist_app.api2.types import (
     EventResult,
     EventStatusCode,
     EventType,
+    ItemUUID,
 )
-from journalist_app.sessions import Session
+from journalist_app.sessions import Session, session
 from models import Reply, Source, Submission
 from redis import Redis
 from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
@@ -53,7 +54,12 @@ class EventHandler:
 
             handler = {
                 EventType.ITEM_DELETED: self.handle_item_deleted,
+                EventType.ITEM_SEEN: self.handle_item_seen,
                 EventType.REPLY_SENT: self.handle_reply_sent,
+                EventType.SOURCE_DELETED: self.handle_source_deleted,
+                EventType.SOURCE_CONVERSATION_DELETED: self.handle_source_conversation_deleted,
+                EventType.SOURCE_STARRED: self.handle_source_starred,
+                EventType.SOURCE_UNSTARRED: self.handle_source_unstarred,
             }[event.type]
         except KeyError:
             return EventResult(
@@ -86,24 +92,11 @@ class EventHandler:
 
     @staticmethod
     def handle_item_deleted(event: Event) -> EventResult:
-        submission = Submission.query.filter(
-            Submission.uuid == event.target.item_uuid
-        ).one_or_none()
-        reply = Reply.query.filter(Reply.uuid == event.target.item_uuid).one_or_none()
-
-        if submission and reply:
-            # Fail if we get unlucky and hit a UUID collision between the
-            # `Submission` and `Reply` tables.  This is vanishingly unlikely,
-            # but SQLite can't enforce uniqueness between them.
-            raise MultipleResultsFound(
-                f"found {event.target.item_uuid} in both submissions and replies"
-            )
-
-        item = submission or reply
+        item = find_item(event.target.item_uuid)
         if item is None:
             return EventResult(
                 event_id=event.id,
-                status=(EventStatusCode.NotFound, f"could not find item: {event.target.item_uuid}"),
+                status=(EventStatusCode.Gone, None),
             )
 
         utils.delete_file_object(item)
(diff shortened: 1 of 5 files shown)


securedrop/journalist_app/api2/events.py
```diff
@@ -135,3 +128,160 @@ def handle_reply_sent(event: Event) -> EventResult:
...
+                    f"could not find source: {event.target.source_uuid}",
+                ),
+            )
+
+        current_version = json_version(source.to_api_v2())
+        if event.target.version != current_version:
+            return EventResult(
+                event_id=event.id,
+                status=(
+                    EventStatusCode.BadRequest,
```
cfm: Might be a little richer semantically to use [HTTP `409 Conflict`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/409):

```suggestion
                    EventStatusCode.Conflict,
```

securedrop/journalist_app/api2/events.py
```diff
@@ -54,6 +55,11 @@ def process(self, event: Event) -> EventResult:
             handler = {
                 EventType.ITEM_DELETED: self.handle_item_deleted,
                 EventType.REPLY_SENT: self.handle_reply_sent,
+                EventType.SOURCE_DELETED: self.handle_source_deleted,
+                EventType.SOURCE_CONVERSATION_DELETED: self.handle_source_conversation_deleted,
+                EventType.SOURCE_STARRED: self.handle_source_starred,
+                EventType.SOURCE_UNSTARRED: self.handle_source_unstarred,
+                EventType.SOURCE_SEEN: self.handle_source_seen,
```
vickiniu: should this instead be `ITEM_SEEN`? since this is just for marking an item as seen?

vickiniu: I wonder if in the case of `SEEN`, the client should submit events via batch, but then when we receive a successful response from a `SEEN` item event, that doesn't necessarily mean that we can mark the entire source as seen. The source seen state we may have to update via the `sources` payload?

legoktm: Oops, yes, it should've been item_seen, fixed. The code itself already operates on the item level; it does return both the updated item and the source, just in case the updating of the item was enough to flip the source from unread to read.

vickiniu: oh nice!! I see, we return the updated `sources`/`items`, makes the client side much simpler actually :upside_down_face:

securedrop/journalist_app/api2/events.py
```diff
@@ -135,3 +128,160 @@ def handle_reply_sent(event: Event) -> EventResult:
             sources={source.uuid: source},
             items={reply.uuid: reply},
         )
+
+    @staticmethod
+    def handle_source_deleted(event: Event) -> EventResult:
+        from journalist_app.api2 import json_version  # cyclic import
```
cfm: Nit: Why not move `json_version()` to `shared.py`?  I guess it's not strictly speaking shared with the v1 API, but creating a separate `utils.py` just for this feels like overkill.  :-)

securedrop/journalist_app/api2/events.py
```diff
@@ -135,3 +128,160 @@ def handle_reply_sent(event: Event) -> EventResult:
...
+            return EventResult(
+                event_id=event.id,
+                status=(
+                    EventStatusCode.Conflict,
+                    f"outdated source: expected {current_version}, got {event.target.version}",
+                ),
+            )
+
+        # Grab the UUIDs of all the items in the collection that will be deleted
+        collection_uuids = {item.uuid: None for item in source.collection}
```
cfm: Nit for clarity:

```suggestion
        # Mark as deleted all the items in the source's collection
        deleted_items = {item.uuid: None for item in source.collection}
```

securedrop/journalist_app/api2/events.py
```diff
@@ -135,3 +128,160 @@ def handle_reply_sent(event: Event) -> EventResult:
...
```
cfm: As above in <https://github.com/freedomofpress/securedrop/pull/7683/files#r2471286987>.

cfm (dismissed):
Looks good!  A few nits for your consideration, but none blocking.

securedrop/journalist_app/api2/events.py
legoktm: Done.

securedrop/journalist_app/api2/events.py
legoktm: Done.

securedrop/journalist_app/api2/events.py
legoktm: Done.

cfm (approved):
Thanks, @legoktm.

PR #7685, opened 2025-10-22
Don't include pending and deleted sources in APIv2

legoktm:
Both should be invisible to journalists, so don't include them in the APIv2 output.

This matches the APIv1 behavior, which has the same filter.

## Test plan

* [ ] Start server and app, run sync, should see the 3 generated sources
* [ ] Create a new source, but do not submit anything. Run sync in app, the source does not appear
* [ ] Submit something, run sync again, now the source appears.

## Checklist

This change accounts for:


zenmonkeykstop (approved):
LGTM, test plan checks out and filter is as described

PR #7686, opened 2025-10-23
api2: define unimplemented event types

vickiniu:
Defines all the event types, even if they are not implemented yet, so the server doesn't 500 on batch requests from the client. Allows us to test the batch endpoint flow from the client with the handlers that are currently implemented!

## Checklist

This change accounts for:


vickiniu:
closing in favor of the real thing in https://github.com/freedomofpress/securedrop/pull/7683

PR #7688, opened 2025-10-23
Update Rust version to 1.90.0

### LINKED ISSUE #7687: Bump Rust version to 1.90.0
rust-audit is currently not building thanks to a dependency requiring 1.89.0 or later - let's get on latest stable.

zenmonkeykstop:
Fixes #7687

## Test plan

- [ ] CI is green
- [ ] Rust version updated across the board (no more 1.87.0 refs)

## Checklist

This change accounts for:

### DIFF
--- a/.github/workflows/cargo-vet.yml
+++ b/.github/workflows/cargo-vet.yml
@@ -9,7 +9,7 @@ jobs:
     name: Vet Dependencies
     runs-on: ubuntu-latest
     # Keep version in sync with rust-toolchain.toml
-    container: rust:1.87.0
+    container: rust:1.90.0
     env:
       CARGO_VET_VERSION: 0.10.0
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,7 +100,7 @@ jobs:
   rust:
     runs-on: ubuntu-latest
     # Keep version in sync with rust-toolchain.toml
-    container: rust:1.87.0
+    container: rust:1.90.0
     steps:
       - uses: actions/checkout@v5
         with:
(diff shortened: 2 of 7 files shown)


legoktm (changes requested):
Few more places that need to be updated:

```
user@dev ~/g/f/securedrop (7687-bump-rust)> rg 1\.87 .github/
.github/workflows/security.yml
13:    container: rust:1.87.0

.github/workflows/cargo-vet.yml
12:    container: rust:1.87.0

.github/workflows/ci.yml
103:    container: rust:1.87.0
```

zenmonkeykstop:
> Few more places that need to be updated:
> 
> ```
> user@dev ~/g/f/securedrop (7687-bump-rust)> rg 1\.87 .github/
> .github/workflows/security.yml
> 13:    container: rust:1.87.0
> 
> .github/workflows/cargo-vet.yml
> 12:    container: rust:1.87.0
> 
> .github/workflows/ci.yml
> 103:    container: rust:1.87.0
> ```

Is this a smell? like it seems weird not to have it defined in a single location.

legoktm:
Kind of, yes. GHA only recently introduced using YAML anchors, so in theory we could use those now (but the zizmor guy wasn't very happy with those which is why I haven't introduced it).

In a technical sense the rust-toolchain.toml file is all we need, because that will make sure we use the correct toolchain no matter what; except then if we pull a different version's docker image, then it has to download the correct version on the fly, which we'd like to a void.

legoktm:
Also, if you don't mind could you file an issue or PR for doing the same toolchain bump in the client repo? Just so we keep them in sync

legoktm (approved):
Pending CI passing

PR #7690, opened 2025-10-24
Update uv dev dependency to 0.9.5

zenmonkeykstop:
Fixes Safety alert 80587:

```
Vulnerability ID: 80507
Affected versions: <0.9.5
Affected versions of the uv package are vulnerable to Type Confusion due to inconsistent parsing of PAX header size overrides in TAR archives. The source-distribution unpacking path in uv, implemented via the astral-tokio-tar module, may ignore a PAX “size” override when the ustar header reports size 0, causing file data to be misinterpreted as additional TAR headers and producing a different extracted file set than other Python installers.
```

No diff review required as this is dev-only.


legoktm (approved):
LGTM, thanks

PR #7699, opened 2025-10-28
feat(`/data`): enforce and document typing and interop considerations for snowflake IDs

cfm:
Prompted by freedomofpress/securedrop-client@2728f88aa22c30f3f2c101f46df08100fa1605cd:

1. [x] Document typing and interop considerations for snowflake IDs
2. [x] Sort events by integer-converted ID
3. [x] Add test cases for lexicographically versus numerically sorted IDs

## Test plan

- [ ] Sanity-check with freedomofpress/securedrop-client#2769.

## Checklist

This change accounts for:

### DIFF
--- a/securedrop/tests/test_journalist_api2.py
+++ b/securedrop/tests/test_journalist_api2.py
@@ -826,3 +826,46 @@ def test_api2_idempotence_period(journalist_app):
     """
 
     assert journalist_app.config["SESSION_LIFETIME"] <= api2.events.IDEMPOTENCE_PERIOD
+
+
+def test_api2_event_ordering(journalist_app, journalist_api_token, test_files):
+    """
+    If two `item_deleted` events for the same item arrive out of order, the
+    numerically later event must observe that the item is already gone by the
+    time it's processed.
+    """
+    with journalist_app.test_client() as app:
+        index = app.get(
+            url_for("api2.index"),
+            headers=get_api_headers(journalist_api_token),
+        )
+        assert index.status_code == 200
+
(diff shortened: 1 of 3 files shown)


cfm:
Rebased from `develop` for f44b470958f55745052354ded2e4619193bb67af, which moved the v2 API's readme.

securedrop/tests/test_journalist_api2.py
```diff
@@ -826,3 +826,46 @@ def test_api2_idempotence_period(journalist_app):
...
+        item_version = index.json["items"][submission_uuid]
+
+        # Two `item_deleted` events targeting the same item:
+        e2 = Event(
+            id="2",
+            target=ItemTarget(item_uuid=submission_uuid, version=item_version),
+            type=EventType.ITEM_DELETED,
+        )
+        e1 = Event(
+            id="001",  # client sends as string; server orders as integer
```
vickiniu: nit: could we use more realistic snowflake IDs here, just to ensure we don't have precision loss issues when converting to `int` in python? 

so maybe: `3419026047977394170` and `3419026047977394171`

vickiniu (changes requested):
looks great!! small comment on changing the test case slightly but otherwise lgtm :)

securedrop/tests/test_journalist_api2.py
cfm: Sure!  Done in f9d207e63a56d7e5bb2352d76e77bcf4c9a3304c.

vickiniu (approved):
sweet!!

